PR #24455, opened 2019-10-16
Throw when failing to convert sequence inner values

### LINKED ISSUE #24406: URLSearchParams constructor interprets an array as a record
```js
const params = new URLSearchParams(["ab"]);
document.write(params);
```

Expected: An error, because the constructor requires `sequence<sequence<USVString>>`.
Actual: Instead it interprets as a record, so the result is `0=ab&length=1`. The spec says [sequences and records are distinguishable](https://heycam.github.io/webidl/#dfn-distinguishable), so this shouldn't happen.

See also #19776 which is also a distinguishablity problem.

saschanaz:
---

- [x] These changes fix #24406

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3097,7 +3097,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.12.1"
-source = "git+https://github.com/servo/rust-mozjs#0832eb778689143cd9f2f3967fac912f7096a466"
+source = "git+https://github.com/servo/rust-mozjs#c2896b9c9f8681070890bc9582378472d0139b13"
 dependencies = [
  "cc",
  "lazy_static",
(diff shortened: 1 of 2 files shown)


jdm:
Needs a `mach update-manifest`.

tests/wpt/web-platform-tests/url/urlsearchparams-constructor.any.js
```diff
@@ -147,6 +147,10 @@ test(function() {
     assert_equals(params.get('a\uD83D\uDCA9b'), 'c');
 }, 'Parse %f0%9f%92%a9');  // Unicode Character 'PILE OF POO' (U+1F4A9)
 
+test(() => {
```
jdm: I think a more appropriate place for this test might be tests/wpt/web-platform-tests/WebIDL/ecmascript-binding/sequence-conversion.html or one of the other files in that directory.

jdm:
@bors-servo r+
Thanks!

PR #23773, opened 2019-07-15
Changed performance interface's timing to an attribute

bcamp:
---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #23330 (GitHub issue number if applicable)

- [X] There are tests for these changes OR
- [ ] These changes do not require tests because ___

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/23773)

PR #24466, opened 2019-10-16
Performance interface's timing member as an attribute change

### LINKED ISSUE #23330: Performance interface's timing member should be an attribute
Per https://w3c.github.io/navigation-timing/#extensions-to-the-performance-interface we should have a `timing` attribute in our WebIDL interface, not a method.

Code: `components/script/dom/webidls/Performance.webidl`, `components/script/dom/performance.rs`
Tests: `./mach test-wpt tests/wpt/web-platform-tests/navigation-timing/`

tomasdivito:
In this patch we change the Performance Interface's `timing` member to be an attribute as the spec requires: https://w3c.github.io/navigation-timing/#extensions-to-the-performance-interface

- [x] These changes fix #23330

---

### DIFF
--- a/components/script/dom/webidls/Performance.webidl
+++ b/components/script/dom/webidls/Performance.webidl
@@ -47,7 +47,7 @@ partial interface Performance {
 // https://dvcs.w3.org/hg/webperf/raw-file/tip/specs/NavigationTiming/Overview.html#performance-timing-attribute
 [Exposed=(Window)]
 partial interface Performance {
-  PerformanceNavigationTiming timing();
+  readonly attribute PerformanceNavigationTiming timing;
 };
 // https://w3c.github.io/navigation-timing/#extensions-to-the-performance-interface
 partial interface Performance {


tomasdivito:
I have yet to check but I believe some tests are broken or at least it gave me that impression when I tried `./mach test-wpt tests/wpt/web-platform-tests/navigation-timing/`

tomasdivito:
@jdm I'm taking a look at the comments you made here: https://github.com/servo/servo/pull/23773

when you talk about the panicked test, I believe you meant this:
`./mach test-wpt tests/wpt/mozilla/tests/mozilla/window_performance.html`
instead of 
`./mach test-wpt tests/wpt/mozilla/window_performance.html` which reports "CRITICAL Unable to find any tests at the path(s)"

Having this in mind, we are currently facing the same issue? Am I right? If that's the case, the solution you proposed should be filled in another issue and worked apart from this one (I can take that, would be nice to work on) but not sure if that would be the steps to follow

jdm:
Yes, this is the same issue that I described. Given the number of additional passing tests in https://build.servo.org/builders/linux-rel-wpt/builds/13812/steps/shell__4/logs/filtered-wpt-errorsummary.log I am happy to merge these changes and file that issue separately.

jdm:
Filed https://github.com/servo/servo/issues/24468.

jdm:
I recommend running the affected test directories and using [these instructions](https://github.com/servo/servo/blob/master/tests/wpt/README.md#updating-test-expectations) to generate logs you can use to automatically update the expected test results.

jdm:
@tomasdivito Are you planning to finish this PR?

tomasdivito:
@jdm Yes, or at least when #24468 is done.

If I don't make progress this weekend I'll report it and remove the assignation. If you think this / or the other patch should have more priority feel free to un-assign me if that's the case

CYBAI:
Close in favor of #25205

PR #24475, opened 2019-10-17
Support [LegacyWindowAlias]

### LINKED ISSUE #24474: Support [LegacyWindowAlias]
Needed for [`webkitURL`](https://url.spec.whatwg.org/#webkiturl).

saschanaz:
---

- [x] These changes fix #24474 (GitHub issue number if applicable)


components/script/dom/bindings/interface.rs
```diff
@@ -260,6 +261,12 @@ pub fn create_noncallback_interface_object(
     define_name(cx, rval.handle(), name);
     define_length(cx, rval.handle(), i32::try_from(length).expect("overflow"));
     define_on_global_object(cx, global, name, rval.handle());
+
+    if is_exposed_in(global, Globals::WINDOW) {
```
nox: This attribute is forbidden in non-Window globals so I'm not sure what this test is for.

nox (commented):
Just one question, r=me otherwise. Thanks for your contribution!

components/script/dom/bindings/interface.rs
saschanaz: Interfaces with `[Exposed=(Window, Worker)]` can also have legacy window aliases, while they only actually get aliases on Window. This is to match the specified behavior.

saschanaz:
Sorry to push again, fixed a test-tidy error, interface map data generation (this caused a bug in WebKitCSSMatrix test), and also added WebKitCSSMatrix.

nox:
No worries.

@bors-servo r+

jdm:
This needs `mach update-manifest` now.

tests/wpt/mozilla/tests/mozilla/interfaces.js
```diff
@@ -45,9 +45,9 @@ function test_interfaces(interfaceNamesInGlobalScope) {
       "Uint32Array",
       "Uint8Array",
       "Uint8ClampedArray",
-      "Uint8ClampedArray",
       "WeakMap",
       "WeakSet",
+      "WebKitCSSMatrix",
```
jdm: Whoops; this belongs in interfaces.html instead, and with a lowercase K.

saschanaz: Whoooops!! Fixed it. It's `WebKitCSSMatrix` anyway though, as WebKit people somehow decided to use uppercase K here.

PR #24487, opened 2019-10-18
UIEvent does not properly initialize its view attribute 

### LINKED ISSUE #24463: UIEvent does not properly initialize its view attribute
The `view` attribute on UIEvent should be properly initialized with its global Window object. Right now, the `view` is always set to null when `UIEvent::new` is called. 

**Spec:** https://w3c.github.io/uievents/#event-type-click
**Relevant code:** `components/script/dom/uievent.rs`

aug2uag:
* `uievent.rs` initializes view for events
```
    pub fn new_inherited(window: &Window) -> UIEvent {
        UIEvent {
            event: Event::new_inherited(),
            view: MutNullableDom::new(Some(window)),
            detail: Cell::new(0),
        }
    }
```

* `compositionevent.rs`
* `focusevent.rs`
* `inputevent.rs`
* `keyboardevent.rs`
* `mouseevent.rs`
* `touchevent.rs`
* `uievent.rs`
new function signatures for events handlers to accomodate changes in `uievent.rs`

---

- [X] These changes fix #24463 

### DIFF
--- a/components/script/dom/compositionevent.rs
+++ b/components/script/dom/compositionevent.rs
@@ -8,7 +8,7 @@ use crate::dom::bindings::codegen::Bindings::CompositionEventBinding::{
 use crate::dom::bindings::codegen::Bindings::UIEventBinding::UIEventBinding::UIEventMethods;
 use crate::dom::bindings::error::Fallible;
 use crate::dom::bindings::reflector::reflect_dom_object;
-use crate::dom::bindings::root::DomRoot;
+use crate::dom::bindings::root::{DomRoot, MutNullableDom};
 use crate::dom::bindings::str::DOMString;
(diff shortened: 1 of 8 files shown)


jdm:
So I took a closer look at the code in MouseEvent::new, and I realized that we when we call InitMouseEvent, we then call InitUIEvent, and that initializes UIEvent's view member. I think my original analysis in #24463 was incorrect, and the changes to "fix" it are unnecessary, since all of the callers of MouseEvent::new already pass a view argument. I'm really sorry :(

aug2uag:
@jdm the original issue was the UIEvent constructor always yielded a `Default::default()` type, in any case was a great experience and thank you for your help

Issue #24483, opened 2019-10-17
Support performing generic operations on DOM objects during document teardown


jdm:
For problems like #23986 and #24052, we have code that currently runs in DOM objects' Drop method because that's the easiest way to deal with "something needs to be unregistered when the DOM object no longer exists". Destructors interact poorly with the GC and can operate in invalid or unexpected states, so we should add a way for DOM objects to opt-in to some kind of cleanup action when a global is being torn down (https://github.com/servo/servo/blob/58c61d3aed273f29d0b19d80b194dc384ce147f0/components/script/dom/window.rs#L1330). We could store a list of weak references and invoke a closure with the strong reference if it still exists. However, we will still want to support running the cleanup if the object is GCed before document teardown. It would be best to have a special DomCleanup type that can be stored inside an Option inside the object that requires cleanup and is responsible for tracking all important cleanup state and destroying itself when appropriate, so that there is no special Drop implementation needed for any DOM objects.

gterzian:
Could you elaborate on how you envision the proposed `DomCleanup` object? 

----

Personally I liked the pattern used in https://github.com/servo/servo/pull/23637(and also starting to be applied in https://github.com/servo/servo/pull/24123) where one would:

1. split up the DOM struct from a non-DOM implementation struct(especially relevant for objects that should be [structured serializable or transferable](https://html.spec.whatwg.org/multipage/structured-data.html), since it's much harder to do so with a singular DOM struct).
2. Store(on the `GlobalScope`) a strong ref to the non-DOM implementation.
3. Store only a weak ref to the DOM struct.  
4. Run a [periodic check](https://github.com/servo/servo/pull/23637/files#diff-1879ac6bd5d567e2aa43529d33474677R1595) as part of the script event-loop, where you clean-up the non-DOM object if the weakref to the corresponding DOM struct has been GC'ed. 
5. When all objects have been cleaned-up, clean-up any "meta" objects used to managed to rest, which could include sending a message to other components to let them know you're not managing any of a type of object anymore(see for MessagePort, where a "router" is used by each global managing ports, and the constellation also keeps a reference to that router, hence when the global is done managing ports, it throws away it's own router and let's the constellation know to do the same([here](https://github.com/servo/servo/pull/23637/files#diff-59d233642d0ce6d687484bdd009e1017R323)). 

----

The above might appear like overkill for most DOM objects, and it might actually make sense for most DOM objects that involve some communication with non-script components, including clean-ups. 

[...]

Issue #21696, opened 2018-09-12
Intermittent CRASH with "mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy"


jdm:
```
  ▶ CRASH [expected PASS] /css/CSS2/normal-flow/max-height-060.xht
  │ 
  │ VMware, Inc.
  │ softpipe
  │ 3.3 (Core Profile) Mesa 18.1.0-devel
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
  │ thread panicked while processing panic. aborting.
  │ thread panicked while processing panic. aborting.
  │ thread panicked while processing panic. aborting.
  │ thread panicked while processing panic. aborting.
  │ thread panicked while processing panic. aborting.
  │ thread panicked while processing panic. aborting.
```
There are about 1000 more entries for the panic too.

CYBAI:
Got this in #23746 at https://taskcluster-artifacts.net/HHjbrwiWRiyGmvT62miTyQ/0/public/filtered-wpt-errorsummary.log

```
  ▶ TIMEOUT [expected FAIL] /css/CSS2/csswg-issues/submitted/css2.1/abspos-non-replaced-width-margin-000.xht
  │ 
  │ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  │ VMware, Inc.
  │ softpipe
  │ 3.3 (Core Profile) Mesa 18.3.0-devel
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ 46 threads are still running after shutdown (bad).
  │ mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Resource busy
  └ thread panicked while processing panic. aborting.
```

CYBAI:
```
  ▶ TIMEOUT [expected FAIL] /css/CSS2/tables/border-collapse-offset-001.xht
  │ 
  │ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  │ 
  │ 
  │ 
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ 50 threads are still running after shutdown (bad).
  │ mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Resource busy
[...]

PR #24499, opened 2019-10-18
Media Session API

### LINKED ISSUE #24172: Media Session API
Spec: https://w3c.github.io/mediasession/

This will allow the implementation of external media controls on mobile and mixed reality platforms.

ferjm:
- [x] These changes fix #24172

This PR introduces all the pieces required to prove an end to end media session flow with Android as a test platform.


components/script/dom/mediasession.rs
```diff
@@ -0,0 +1,146 @@
...
+            Some(handler) => self
+                .action_handlers
+                .borrow_mut()
+                .insert(action.into(), handler.clone()),
+            None => self.action_handlers.borrow_mut().remove(&action.into()),
+        };
+    }
+}
+
+impl Drop for MediaSession {
```
gterzian: It might be worth revisiting this in the light of https://github.com/servo/servo/issues/24483

components/script/script_thread.rs
```diff
@@ -3871,6 +3882,17 @@ impl ScriptThread {
         }
     }
 
+    fn handle_media_session_action(
+        &self,
+        browsing_context_id: TopLevelBrowsingContextId,
+        action: MediaSessionActionType,
+    ) {
+        match self.media_sessions.borrow().get(&browsing_context_id) {
```
gterzian: Have you considered sending the `PipelineId` along in the message, and then grabbing the corresponding document via [`find_document`](https://github.com/servo/servo/blob/882dcc8408159bed5633337d11e83b6839f97e34/components/script/script_thread.rs#L459), and then only handling the action if the document is fully active?

gterzian: On second thoughts, this might be best dealt with via the concept of the [active-media-session](https://w3c.github.io/mediasession/#active-media-session), although I assume the active session will belong to a fully-active document...

components/script/script_thread.rs
```diff
@@ -693,6 +694,11 @@ pub struct ScriptThread {
 
     /// Code is running as a consequence of a user interaction
     is_user_interacting: Cell<bool>,
+
+    /// The MediaSessions known by this thread, if any.
+    /// There can only be one active MediaSession.
+    /// The constellation has the BrowsingContextId of the active MediaSession, if any.
+    media_sessions: DomRefCell<HashMap<TopLevelBrowsingContextId, Dom<MediaSession>>>,
```
gterzian: Have you considered using the `BrowsingContextId` vs the top-level for the key? The spec seem to say that each browsing-context, whether it is a top-level or a nested one, can have a session, see https://w3c.github.io/mediasession/#media-session-routing.

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -459,6 +463,12 @@ impl ServoGlue {
         self.process_event(WindowEvent::Keyboard(key_event))
     }
 
+    pub fn media_session_action(&mut self, action: i32) -> Result<(), &'static str> {
+        info!("Media session action {:?}", action);
+        let browser_id = self.get_browser_id()?;
```
gterzian: One potential issue here I can think of is that the `browser_id` will be the "active tab", whereas the "active media session" could, I think, be a tab playing media while not being the "active tab".

Have you considered not tying the media session action to a top-level at this point, and simply forwarding it "as is" to the constellation. The constellation could then keep track of which session is "active", and in which event-loop the session is found(allowing for iframe to have an active session, for example), and then forwarding the action with the `BrowsingContextId` of the active session to the event-loop where it is found(I guess via the pipeline corresponding to the BC)?

components/servo/lib.rs
```diff
@@ -762,6 +776,31 @@ where
                     self.embedder_events.push(event);
                 },
 
+                (EmbedderMsg::MediaSessionEvent(event), ShutdownState::NotShuttingDown) => {
```
gterzian: In the light of the suggestion at https://github.com/servo/servo/pull/24499/files#r336776086, it might make sense to do this in the constellation, and make the `active_media_session` a `BrowsingContextId`(not a top-level). 

It might then make sense to make the `EmbedderMsg::MediaSessionEvent` into a [`ScriptMsg`](https://github.com/servo/servo/blob/882dcc8408159bed5633337d11e83b6839f97e34/components/script_traits/script_msg.rs#L116) instead, where the constellation would run the below logic and then potentially forward the event to the embedder, potentially containing the `TopLevelBrowsingContextId` representing the "tab" where the event came from(while it might have come from an iframe nested inside of it).

components/script/dom/htmlmediaelement.rs
```diff
@@ -1719,6 +1719,14 @@ impl HTMLMediaElement {
                 if self.Controls() {
                     self.render_controls();
                 }
+
+                // Send a media session event with the obtained metadata.
+                self.send_media_session_event(MediaSessionEvent::SetMetadata(MediaMetadata {
+                    // TODO(ferjm) set url if no title.
```
Manishearth: should this also modify the metadata object on `navigator.mediasession`?

Manishearth (approved):
looks good overall!

@gterzian's suggestion for making it a global thing might be useful. Unsure how to structure that.

jdm:
The wdspec failures can be ignored, but https://community-tc.services.mozilla.com/tasks/groups/VGii-LvCT0Kgqgu54WfRZQ has a number of relevant failures.

jdm:
Oops, missing a `./mach fmt` now.

jdm:
@bors-servo retry
* #21696

Issue #22854, opened 2019-02-08
Remove uses of opts::get() from as much code as possible


jdm:
Right now the code in components/config has a bunch of code depending on it because the command line arguments for Servo are global singleton that any code can access. This makes changing the command line code for Servo frustrating, because a lot of code gets rebuilt unnecessarily. We should remove as many uses of opts::get() as possible, and replace them with values that are passed into appropriate constructors via the code in components/servo/lib.rs.

We can focus on one component at a time:
* [ ] bluetooth
* [ ] compositing
* [ ] constellation
* [ ] layout
* [ ] layout_thread
* [ ] metrics
* [ ] net
* [ ] profile
* [ ] profile_traits
* [ ] script
* [ ] style

JacksonCoder:
@highfive: assign me

peterjoel:
Couldn't `opts` just be a non-persistent `prefs` override?

jdm:
I'm not sure I understand what you're suggesting. Do you have an example?

peterjoel:
> I'm not sure I understand what you're suggesting. Do you have an example?

I am suggesting that all CLI arguments would actually set a preference, which would not be persisted. Instead of accessing `opts`, any configurable code would just access `prefs`. The actual value would be evaluated in ascending order of precedence: `prefs.json`, user preferences, CLI argument.

jdm:
Ah, I see. The advantage of keeping the command line arguments separate from preferences is that the arguments are not mutable. This is one reason why it's safe to remove access to the global arguments storage and pass the values directly to the code that uses them, allowing us to remove dependencies.

peterjoel:
@jdm It's not ideal then that _some_ opts are actually duplicated as (mutable) prefs, so you _could_ get a different value for `layout.threads`, depending on if you access it via `opts` or `prefs`.

jdm:
That's true! Anything that is duplicated should be removed from the immutable options structure and any users changed to use the preference.

jdm:
With the caveat that using the command line option should mutate the preference.

peterjoel:
Rather than replace all instances of `opts` with dependency injection, another approach is to split opts into two levels: globals (which should change infrequently) and crate-level options (which can develop alongside the features of the crate itself). For example, `bluetooth` only uses one option: `headless`, which is global. `compositing` uses some globals (`exit_after_load` maybe?), but also more specific options like `device_pixels_per_px`.

Each crate would define its own options and provide access to both its own and the global options. 

[...]

Issue #23290, opened 2019-04-29
Intermittent reftest failures due to font selection choices


jdm:
Test:
<img width="414" alt="Screen Shot 2019-04-29 at 10 48 40 AM" src="https://user-images.githubusercontent.com/27658/56904505-5f561500-6a6c-11e9-8a5e-c0fe0d56ebae.png">
Reference:
<img width="380" alt="Screen Shot 2019-04-29 at 10 48 56 AM" src="https://user-images.githubusercontent.com/27658/56904519-6715b980-6a6c-11e9-9f16-4f97ef94878d.png">

https://searchfox.org/mozilla-central/source/testing/web-platform/tests/css/CSS2/syntax/at-charset-001.xht
https://searchfox.org/mozilla-central/source/testing/web-platform/tests/css/CSS2/syntax/at-charset-001-ref.xht
https://searchfox.org/mozilla-central/source/testing/web-platform/tests/css/CSS2/syntax/support/at-charset-001.css

It's not clear to me how this can happen, but it does.

jdm:
<img width="803" alt="Screen Shot 2019-05-09 at 11 02 43 PM" src="https://user-images.githubusercontent.com/27658/57499686-95866680-72ae-11e9-87b5-49998821b68a.png">
<img width="803" alt="Screen Shot 2019-05-09 at 11 02 47 PM" src="https://user-images.githubusercontent.com/27658/57499689-97502a00-72ae-11e9-9073-1221a3de4c09.png">

jdm:
This is a relatively common intermittent failure on taskcluster mac jobs; this will only get worse as we add more of them.

jdm:
The info added to #23375 has yielded interesting output like the following:
```
"[2019-05-14T23:50:05Z INFO  gfx::font] couldn't find font for 'T' (54) in families [FontGroupFamily { font_descriptor: FontDescriptor { template_descriptor: FontTemplateDescriptor { weight: FontWeight(400.0), stretch: FontStretch(NonNegative(Percentage(1.0))), style: Normal }, variant: Normal, pt_size: 16px }, family_descriptor: FontFamilyDescriptor { name: Generic(Atom('serif' type=static)), scope: Any }, loaded: true, font: None }]"
```
I find this highly suspect!

jdm:
[...]

PR #24500, opened 2019-10-19
Remove usage of opts::get() from style.

kitlith:
Use `AtomicBool`s to create a global storage for a couple of settings used by an implementation of `Default`, suggested in https://github.com/servo/servo/issues/22854#issuecomment-540955207.

I do have one main question: I'm currently setting these statics in `Servo::new()`, would there be a better place to put this? Same goes with making these `pub`, is there a better way to go about this?

---

  - Seems to be failing in script for unknown/unrelated reasons, experiencing SIGKILL on two different machines. (May be an issue with the latest commit?) Still works fine on CI though.
- [X] These changes help with #22854

These changes do not require tests because these are cleanup changes.

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -310,6 +310,15 @@ where
         // Global configuration options, parsed from the command line.
         let opts = opts::get();
 
+        use std::sync::atomic::Ordering;
+
+        style::context::DEFAULT_DISABLE_STYLE_SHARING_CACHE
+            .store(opts.disable_share_style_cache, Ordering::Relaxed);
+        style::context::DEFAULT_DUMP_STYLE_STATISTICS
+            .store(opts.style_sharing_stats, Ordering::Relaxed);
+        style::traversal::IS_SERVO_NONINCREMENTAL_LAYOUT
+            .store(opts.nonincremental_layout, Ordering::Relaxed);
+
         if !opts.multiprocess {
             media_platform::init();
         }
--- a/components/style/context.rs
+++ b/components/style/context.rs
@@ -103,14 +103,29 @@ fn get_env_usize(name: &str) -> Option<usize> {
     })
 }
 
+/// A global variable holding the state of
+/// `StyleSystemOptions::default().disable_style_sharing_cache`.
+/// See [#22854](https://github.com/servo/servo/issues/22854).
+#[cfg(feature = "servo")]
+pub static DEFAULT_DISABLE_STYLE_SHARING_CACHE: std::sync::atomic::AtomicBool =
+    std::sync::atomic::AtomicBool::new(false);
+
+/// A global variable holding the state of
+/// `StyleSystemOptions::default().dump_style_statistics`.
+/// See [#22854](https://github.com/servo/servo/issues/22854).
+#[cfg(feature = "servo")]
+pub static DEFAULT_DUMP_STYLE_STATISTICS: std::sync::atomic::AtomicBool =
+    std::sync::atomic::AtomicBool::new(false);
+
 impl Default for StyleSystemOptions {
     #[cfg(feature = "servo")]
     fn default() -> Self {
-        use servo_config::opts;
+        use std::sync::atomic::Ordering;
 
         StyleSystemOptions {
-            disable_style_sharing_cache: opts::get().disable_share_style_cache,
-            dump_style_statistics: opts::get().style_sharing_stats,
+            disable_style_sharing_cache: DEFAULT_DISABLE_STYLE_SHARING_CACHE
+                .load(Ordering::Relaxed),
+            dump_style_statistics: DEFAULT_DUMP_STYLE_STATISTICS.load(Ordering::Relaxed),
             style_statistics_threshold: DEFAULT_STATISTICS_THRESHOLD,
         }
     }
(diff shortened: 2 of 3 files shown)


kitlith:
sigh. well the build failure is definitely from master, because that's what I was trying to build on the second machine. test-tidy should be fixed now.

components/servo/lib.rs
```diff
@@ -310,6 +310,13 @@ where
         // Global configuration options, parsed from the command line.
         let opts = opts::get();
 
+        use std::sync::atomic::Ordering;
+
+        style::context::DEFAULT_DISABLE_STYLE_SHARING_CACHE
+            .store(opts.disable_share_style_cache, Ordering::SeqCst);
```
emilio: Just `Relaxed` is fine.

components/style/context.rs
```diff
@@ -103,14 +103,28 @@ fn get_env_usize(name: &str) -> Option<usize> {
     })
 }
 
+/// A global variable holding the state of
+/// `StyleSystemOptions::default().disable_style_sharing_cache`.
+/// This is motivated by [issue #22854](https://github.com/servo/servo/issues/22854).
```
emilio: I don't think this line of the comment is useful, it only makes sense in the context of this PR.

components/style/context.rs
```diff
@@ -103,14 +103,28 @@ fn get_env_usize(name: &str) -> Option<usize> {
...
 impl Default for StyleSystemOptions {
     #[cfg(feature = "servo")]
     fn default() -> Self {
-        use servo_config::opts;
+        use std::sync::atomic::Ordering;
 
         StyleSystemOptions {
-            disable_style_sharing_cache: opts::get().disable_share_style_cache,
-            dump_style_statistics: opts::get().style_sharing_stats,
+            disable_style_sharing_cache: DEFAULT_DISABLE_STYLE_SHARING_CACHE.load(Ordering::SeqCst),
```
emilio: Same, just relaxed is fine.

emilio (commented):
This seems ok. @jdm is this the pattern that other code has been following? I guess it does accomplish the "changing opts rebuilds less" bits (assuming we remove the other usage in the crate and the dependency), but the pattern seems somewhat weird.

jdm:
This is our first time trying this pattern, since we skipped other uses of opts::get inside static functions. I'm in favour.

components/style/context.rs
kitlith: I thought I'd try to leave the issue link for motivation especially because this is an odder pattern than usual, to try and preempt "why not just use `opts::get()`" for people not familiar with the issue.

emilio: I still think it's not particularly useful. I assume we'd get rid of `opts` as a dependency of `style` after changing the other similar uses. Once that's done, if somebody wants to use `opts::get()` again they need to re-introduce it as a dependency, which makes the consequences of it much more obvious.

Not going to block this patch on removing this comment, but I don't quite see the point of it... Maybe just `See #22854` is a good compromise?

Either way this looks ok, mind squashing? Thanks!

kitlith: K, I'll do that and squash once I'm at a computer.

kitlith:
I just noticed that there's only a single `opts::get()` left in the style component:

https://github.com/servo/servo/blob/882dcc8408159bed5633337d11e83b6839f97e34/components/style/traversal.rs#L48-L54

Would you like me to apply the same pattern here and remove opts from the dependency tree of this crate?

jdm:
Yes please!

emilio:
That'd indeed be ideal. Do you plan to do this as part of this PR or a new one? (Just to know if I should submit the current PR as is :))

kitlith:
I can do that in this PR, but it might take a bit so you can merge if impatient. (I don't mind doing it in a seperate one either, gets me another PR towards octoberfest lol, but i'd prefer not to PR spam)

emilio (approved):
Two PRs doing two different things is not spam, in my book ;)

Anyhow, let's submit this. If you get to push your other changes before it lands then it will just not get merged and we'll review those too. Otherwise new PR is fine :)

kitlith:
Uh? I don't think I touched anything related to the tests that changed...

Anyway, I'll get rid of the other usage I mentioned earlier now.

kitlith:
So: we can't remove the dep on servo_config yet because style still relies on pref in a few places. This feels like something that is not as easilly replaced...

emilio:
> So: we can't remove the dep on servo_config yet because style still relies on pref in a few places. This feels like something that is not as easilly replaced...

Ah, that is a good point... :/

So there's no realistic path to get rid of the dependency for now... Is there any point in removing `opts::get()` usage then @jdm?

jdm:
There is! I want to split prefs and options into separate crates in the future, so this work is still useful.

jdm:
@bors-servo r=emilio
And to clear up the earlier mystery, the test failure was caused by #23290.

jdm:
```
error[E0308]: mismatched types
  --> components/style/traversal.rs:57:40
   |
57 | fn is_servo_nonincremental_layout() -> bool {
   |    ------------------------------      ^^^^ expected bool, found ()
   |    |
   |    implicitly returns `()` as its body has no tail or `return` expression
...
60 |     IS_SERVO_NONINCREMENTAL_LAYOUT.load(Ordering::Relaxed);
   |                                                           - help: consider removing this semicolon
   |
   = note: expected type `bool`
              found type `()`

```

kitlith:
*facepalm*

PR #24508, opened 2019-10-20
Add start_time to resource timing.

### LINKED ISSUE #21254: PerformanceResourceTiming: startTime
In components/net/http_loader.rs::http_fetch we need to set start_time for the PerformanceEntry member of PerformanceResourceTiming. This will be equal to either fetch_start or redirect_start.

Spec: https://w3c.github.io/resource-timing/#dfn-starttime

pajamapants3000:
`start_time` property added to `ResourceFetchTiming`, which enables the setting of `start_time` in the `PerformanceEntry` member of `PerformanceResourceTiming`.

Following the specification at https://w3c.github.io/resource-timing/#dfn-starttime, `start_time` is set to the value of `redirect_start` if redirection occurs and the timing allow check passes. Otherwise it has the same value as `fetch_start`.

---

- [x] These changes fix #21254

### DIFF
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -21,8 +21,8 @@ use net_traits::filemanager_thread::RelativePos;
 use net_traits::request::{CredentialsMode, Destination, Referrer, Request, RequestMode};
 use net_traits::request::{Origin, ResponseTainting, Window};
 use net_traits::response::{Response, ResponseBody, ResponseType};
-use net_traits::ResourceAttribute;
 use net_traits::{FetchTaskTarget, NetworkError, ReferrerPolicy, ResourceFetchTiming};
+use net_traits::{ResourceAttribute, ResourceTimeValue};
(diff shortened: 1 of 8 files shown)


pajamapants3000:
I question whether the timing logic belongs inside `set_attribute`. I went with that approach for consistency with existing code, but it seems like the caller already needs to know what times to set and when. Feels like the logic is being split into two places. I could change it so `set_attribute` just sets the time to whatever the caller says (at least for `start_time`) and just have the caller be entirely responsible for setting the time correctly. It already _mostly_ is.

jdm:
Yeah, I'm on the fence about it. Let's merge it the way it's written and if it turns out to be difficult to reason about in the future we can rewrite it. Thanks!

@bors-servo r+

jdm:
And thank you for writing those unit tests, too!

jdm:
@bors-servo retry
#23290

PR #24435, opened 2019-10-14
Issue 23607: first pass of changes for compatibility with Python3

marmeladema:
As much as i want to migrate entirely to Python3 (see #23607), it will require some time as changes in web-platform-tests are significant and rely on upstream fixes to be merged and synced downstream.
In the meantime, lets improve compatibility with Python3 so that later, migration will be less painful.

Build system is definitely not ready yet for Python3, but its a step in the right direction.

---

- [x ] `./mach build -d` does not report any errors
- [ x] `./mach test-tidy` does not report any errors

Issue #23607, opened 2019-06-23
Support Python 3


saschanaz:
We'll probably want to migrate to Python 3 because Python 2 will die next year.

marmeladema:
What is left to do here? I am willing to help if needed

jdm:
Ok, so pieces that I know about:
* mach does not support python 3 (https://bugzilla.mozilla.org/show_bug.cgi?id=1473498)
* mozbuild does not support python 3 (https://bugzilla.mozilla.org/show_bug.cgi?id=1473499)
* probably some more dependencies of https://bugzilla.mozilla.org/show_bug.cgi?id=1388447
* the web-platform-test test harness/webserver/etc. does not support python 3 (https://github.com/web-platform-tests/wpt/issues/10128, https://github.com/web-platform-tests/wpt/issues/8820)
* we need to import changes to the python WebIDL parser to support python 3 (https://bugzilla.mozilla.org/show_bug.cgi?id=1563937)
* we need to support running all of the non-parser python code in components/script/dom/bindings/codegen under python 3
* then our own developer tooling needs to be ported to run on python 3 (python/servo, python/tidy, tests/wpt/webgl/tools)
* then we will need to ensure that any additional python scripts we care about under etc/ run under python 3

I think that's probably the full set of dependencies.

marmeladema:
Thanks! I've updated most of the pieces i knew of to be python3 compatible but ultimately ended up fighting with `mach` itself:
```Python
Traceback (most recent call last):                                                                                                                                                                                                                             
  File "./mach", line 103, in <module>                                                                                                                                                                                                                         
    main(sys.argv)                                                                                                                                                                                                                                             
  File "./mach", line 33, in main                                                                                                                                                                                                                              
    mach = mach_bootstrap.bootstrap(topdir)                                                                                                                                                                                                                    
  File "/home/adema/code/servo/python/mach_bootstrap.py", line 288, in bootstrap                                                                                                                                                                               
    import mach.main                                                                                                                                                                                                                                           
  File "/home/adema/code/servo/python/_virtualenv/lib/python3.6/site-packages/mach/main.py", line 34, in <module>                                                                                                                                              
    from .config import ConfigSettings                                                                                                                                                                                                                         
  File "/home/adema/code/servo/python/_virtualenv/lib/python3.6/site-packages/mach/config.py", line 166                                                                                                                                                        
    raise AttributeError().__class__, exc, tb                                                                                                                                                                                                                  
```

[...]

PR #24512, opened 2019-10-20
More changes for Python3 compatibility

marmeladema:
Following #24435 for #23607 here are even more changes to be compatible with Python3.
I managed to get `./mach build` properly work with Python3 but `test-tidy` does not work yet because of a lot of problems in `web-platform-tests` which i will have to deal with at some point.

Because flake8 appears to be incompatible with the distro package in some way, i had to change to way we call flake8 (which was deprecated anyway). With this change, we should be able to update flake8 to a recent version but subprocess can be surprising on Windows platform so i'd like someone to try out those changes.

---

### DIFF
--- a/python/tidy/servo_tidy/tidy.py
+++ b/python/tidy/servo_tidy/tidy.py
@@ -9,14 +9,12 @@
 
 from __future__ import print_function
 
-import contextlib
 import fnmatch
 import imp
 import itertools
 import json
 import os
 import re
-from io import StringIO
 import subprocess
 import sys
 
@@ -327,28 +325,21 @@ def check_by_line(file_name, lines):
 
 
 def check_flake8(file_name, contents):
-    from flake8.main import check_code
-
     if not file_name.endswith(".py"):
         raise StopIteration
-
-    @contextlib.contextmanager
-    def stdout_redirect(where):
-        sys.stdout = where
-        try:
-            yield where
-        finally:
-            sys.stdout = sys.__stdout__
 
     ignore = {
         "W291",  # trailing whitespace; the standard tidy process will enforce no trailing whitespace
         "E501",  # 80 character line length; the standard tidy process will enforce line length
     }
 
-    output = StringIO()
-    with stdout_redirect(output):
-        check_code(contents, ignore=ignore)
-    for error in output.getvalue().splitlines():
+    output = ""
+    try:
+        args = ["flake8", "--ignore=" + ",".join(ignore), file_name]
+        subprocess.check_output(args)
+    except subprocess.CalledProcessError as e:
+        output = e.output
+    for error in output.splitlines():
         _, line_num, _, message = error.split(":", 3)
         yield line_num, message.strip()
 
--- a/python/tidy/servo_tidy/tidy.py
+++ b/python/tidy/servo_tidy/tidy.py
@@ -24,7 +24,7 @@ import colorama
 import toml
 import voluptuous
 import yaml
-from licenseck import OLD_MPL, MPL, APACHE, COPYRIGHT, licenses_toml, licenses_dep_toml
+from .licenseck import OLD_MPL, MPL, APACHE, COPYRIGHT, licenses_toml, licenses_dep_toml
 from six import iteritems
 topdir = os.path.abspath(os.path.dirname(sys.argv[0]))
 wpt = os.path.join(topdir, "tests", "wpt")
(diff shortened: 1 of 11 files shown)


marmeladema:
@jdm i don't know how to add you as reviewer so i thought i would just ping you :+1:

JacobAWilkins (approved):
Very nice

jdm:
r? @SimonSapin
Honestly ai think Simon has more experience reviewing python3 than I do.

marmeladema:
> r? @SimonSapin
> Honestly I think Simon has more experience reviewing python3 than I do.

Yeah i did not imply to remove @SimonSapin from reviewers, just wanted to add you

python/tidy/servo_tidy/tidy.py
```diff
@@ -327,28 +326,25 @@ def check_by_line(file_name, lines):
...
-    @contextlib.contextmanager
-    def stdout_redirect(where):
-        sys.stdout = where
-        try:
-            yield where
-        finally:
-            sys.stdout = sys.__stdout__
+    temp_file = tempfile.NamedTemporaryFile()
+    temp_file.write(contents)
+    temp_file.flush()
```
SimonSapin: Why create this temporary file rather than pass `file_name` to flake8?

SimonSapin (commented):
I’d like to also get some testing so we don’t accidentally regress this. If `./mach build` is supposed to work on Python 3 now, please change it to `python3 mach build` here, for example:

https://github.com/servo/servo/blob/5fb941d6f7de9289b5ffa683c924c2e2bba74d3a/etc/taskcluster/decision_task.py#L180

python/tidy/servo_tidy/tidy.py
marmeladema: I wondered that myself and choose to dump 
the content because I did not take time to figure out if file_name really existed on disk

marmeladema:
> I’d like to also get some testing so we don’t accidentally regress this. If `./mach build` is supposed to work on Python 3 now, please change it to `python3 mach build` here, for example:
> 
> https://github.com/servo/servo/blob/5fb941d6f7de9289b5ffa683c924c2e2bba74d3a/etc/taskcluster/decision_task.py#L180

It passes on a branch based on this one with a few more tweaks. We can either merge this one, and i'll come back in a few days with the everything sorted out to build with python3 or i add the remaining bits in this one.

Also about just changing `./mach build` to `python3 mach build`, i wonder the effect on the subsequent commands that will run with python2 like `./mach test-tidy --no-progress --self-test`

SimonSapin:
I’ll leave it up to you whether you prefer to add a commit in this PR for those few more tweaks (maybe they’re already ready) or if you prefer not to block landing this PR and make a new one later.

I don’t expect that a Python 3 command followed by a Python 2 command would cause any trouble, they’re independent processes. Finding out for sure is what CI is for!

marmeladema:
> I’ll leave it up to you whether you prefer to add a commit in this PR for those few more tweaks (maybe they’re already ready) or if you prefer not to block landing this PR and make a new one later.

Then i would prefer merging this now and come back later with the remaining bits. My other branch actually only check that you have python3, not python2 anymore (even though the code remains compatible) and will take some time to make python version detection compatible as its not ready yet.

> I don’t expect that a Python 3 command followed by a Python 2 command would cause any trouble, they’re independent processes. Finding out for sure is what CI is for!

We might have to remove `python/_virtualenv` between python3 and python2 though

python/tidy/servo_tidy/tidy.py
marmeladema: Should I change that before merging?

SimonSapin: Yes, please do.

SimonSapin:
Oh yeah, there’s virtualenv. It may be worth having it at a different location based on the Python version.

python/tidy/servo_tidy/tidy.py
marmeladema: Done!

SimonSapin:
Looks good, thanks!

@bors-servo r+

CYBAI:
@bors-servo retry

- network issue in magicleap build

marmeladema:
> @bors-servo retry
> 
>     * network issue in magicleap build

I don't think the `retry` worked?

jdm:
It does work; there is a queue at https://build.servo.org/homu/queue/servo.

PR #24519, opened 2019-10-21
Cleanup canvasrenderingcontext

### LINKED ISSUE #24496: Rename CanvasState methods to use snake_case method names instead of UpperCamelCase, and extract it to its own module
The member methods of CanvasState were largely extracted from CanvasRenderingContext2D, where they used to be members of the generated CanvasRenderContext2DMethods trait. Now that they are handwritten, it's confusing to have them keep the same naming pattern as generated code. We should rename them so GetImageData becomes get_image_data (for example).

We should also extract CanvasState into its own file in the components/script root directory.

Code: `components/script/dom/canvasrenderingcontext.rs`

hbina:
---

- [x] These changes fix #24496

- [x] These changes do not require tests because it is purely a refactoring...so whatever tests we are already using now should suffice.

### DIFF
--- a/components/script/dom/canvasrenderingcontext2d.rs
+++ b/components/script/dom/canvasrenderingcontext2d.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use crate::canvas_state::{CanvasContextState, CanvasState};
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::CanvasRenderingContext2DBinding;
 use crate::dom::bindings::codegen::Bindings::CanvasRenderingContext2DBinding::CanvasFillRule;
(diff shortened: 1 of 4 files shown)


components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -77,25 +54,25 @@ pub struct CanvasRenderingContext2D {
...
-    line_width: f64,
-    line_cap: LineCapStyle,
-    line_join: LineJoinStyle,
-    miter_limit: f64,
-    transform: Transform2D<f32>,
-    shadow_offset_x: f64,
-    shadow_offset_y: f64,
-    shadow_blur: f64,
-    shadow_color: RGBA,
+pub struct CanvasContextState {
```
jdm: We should move this into the new file instead of making all of it public.

components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -9,57 +9,34 @@ use crate::dom::bindings::codegen::Bindings::CanvasRenderingContext2DBinding::Ca
...
+use euclid::default::{Point2D, Rect, Size2D, Transform2D};
+use ipc_channel::ipc::IpcSender;
+use servo_url::ServoUrl;
 use std::{fmt, mem};
 
 #[must_root]
 #[derive(Clone, JSTraceable, MallocSizeOf)]
 #[allow(dead_code)]
-enum CanvasFillOrStrokeStyle {
+pub enum CanvasFillOrStrokeStyle {
```
jdm: Let's move this into the new file instead of making it public.

components/script/dom/canvasrenderingcontext.rs
```diff
@@ -0,0 +1,1401 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jdm: Let's move this to `components/script/canvas_state.rs` instead.

components/script/dom/canvasrenderingcontext.rs
```diff
@@ -0,0 +1,1401 @@
...
+use net_traits::image_cache::UsePlaceholder;
+use net_traits::request::CorsSettings;
+use pixels::PixelFormat;
+use profile_traits::ipc as profiled_ipc;
+use script_traits::ScriptMsg;
+use servo_url::{ImmutableOrigin, ServoUrl};
+use std::cell::Cell;
+use std::str::FromStr;
+use std::sync::Arc;
+#[must_root]
```
jdm: nit: Add a newline above this.

components/script/dom/canvasrenderingcontext.rs
```diff
@@ -0,0 +1,1401 @@
...
+#[derive(JSTraceable, MallocSizeOf)]
+pub struct CanvasState {
+    #[ignore_malloc_size_of = "Defined in ipc-channel"]
+    pub ipc_renderer: IpcSender<CanvasMsg>,
```
jdm: Rather than making all of these members public, can we add getter methods for anything that needs to be accessed from canvasrenderingcontext2d.rs and offscreenrenderingcontext2d.rs?

components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -2073,12 +713,12 @@ pub fn parse_color(string: &str) -> Result<RGBA, ()> {
 
 // Used by drawImage to determine if a source or destination rectangle is valid
 // Origin coordinates and size cannot be negative. Size has to be greater than zero
-fn is_rect_valid(rect: Rect<f64>) -> bool {
+pub fn is_rect_valid(rect: Rect<f64>) -> bool {
```
jdm: Let's move these helper functions to the new file instead of making them public.

jdm (changes requested):
Looking good so far!

components/script/dom/canvasrenderingcontext.rs
jdm: Right, I suggest we rename components/script/dom/canvasrenderingcontext.rs to components/script/canvas_state.rs.

hbina: Hmmm, how is `components/script/dom/canvasrenderingcontext2d.rs` suppose to access `canvas_state` then? there is no `mod.rs` that exposes this

jdm: components/script/lib.rs should include a `pub(crate) mod canvas_state;` line.

hbina:
I have applied the changes that you requested, @jdm but I have no idea what to name the static helper functions that I put in `some_func.rs`

components/script/dom/some_func.rs
```diff
@@ -0,0 +1,78 @@
+// FIXME :: Rename this file...
```
jdm: Since the only callers of these functions live in canvas_state.rs, we should move all of them into that file and remove this one.

components/script/lib.rs
```diff
@@ -77,6 +77,7 @@ pub mod textinput;
 mod timers;
 mod unpremultiplytable;
 mod webdriver_handlers;
+pub mod canvas_state;
```
jdm: This can be `pub(crate)`.

jdm (changes requested):
This is almost there! You can go ahead and squash all the commits into one so it's ready to merge.

hbina:
I have squashed them, see if I mess up any Git

jdm:
@bors-servo r+
Thanks!

PR #24521, opened 2019-10-22
Implement MouseEvent's x/y and offsetX/offsetY attributes

### LINKED ISSUE #24248: Missing MouseEvent attributes defined by CSSOM View
https://drafts.csswg.org/cssom-view/#extensions-to-the-mouseevent-interface

In particular, the pageX/pageY attributes are used by three.js code in http://rawcdn.githack.com/mrdoob/three.js/r105/examples/js/controls/FirstPersonControls.js . This is likely to be a source of surprising web compat issues!

This will likely have an impact on #20931.

tigleym:
Implement MouseEvent's x/y and offsetX/offsetY attributes

---

- [ ] These changes fix #24248 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/mouseevent.rs
+++ b/components/script/dom/mouseevent.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use crate::dom::bindings::codegen::Bindings::EventBinding::EventBinding::EventMethods;
 use crate::dom::bindings::codegen::Bindings::MouseEventBinding;
 use crate::dom::bindings::codegen::Bindings::MouseEventBinding::MouseEventMethods;
 use crate::dom::bindings::codegen::Bindings::UIEventBinding::UIEventMethods;
@@ -12,6 +13,7 @@ use crate::dom::bindings::root::{DomRoot, MutNullableDom};
 use crate::dom::bindings::str::DOMString;
 use crate::dom::event::{Event, EventBubbles, EventCancelable};
 use crate::dom::eventtarget::EventTarget;
+use crate::dom::node::Node;
 use crate::dom::uievent::UIEvent;
 use crate::dom::window::Window;
 use dom_struct::dom_struct;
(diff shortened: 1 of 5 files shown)


tests/wpt/metadata/css/cssom-view/mouseEvent.html.ini
```diff
@@ -1,4 +1 @@
 [mouseEvent.html]
-  [MouseEvent's x and y must be equal to clientX and clientY.]
-    expected: FAIL
-
```
atouchet: If there are no more failing tests you can delete the file.

components/script/dom/mouseevent.rs
```diff
@@ -192,6 +203,56 @@ impl MouseEventMethods for MouseEvent {
...
+        self.client_y.get()
+    }
+
+    // https://drafts.csswg.org/cssom-view/#dom-mouseevent-offsetx
+    fn OffsetX(&self) -> i32 {
+        let event = self.upcast::<Event>();
+        if event.dispatching() {
+            match event.GetTarget() {
+                Some(target) => {
+                    if let Some(element) = target.downcast::<Element>() {
```
jdm: If we downcast to Node instead then we can call `Node::client_rect`. This means we should get expected behaviour for mouse events on non-Elements like text nodes, for example.

components/script/dom/mouseevent.rs
```diff
@@ -192,6 +203,56 @@ impl MouseEventMethods for MouseEvent {
...
+        }
+    }
+
+    // https://drafts.csswg.org/cssom-view/#dom-mouseevent-offsety
+    fn OffsetY(&self) -> i32 {
+        let event = self.upcast::<Event>();
+        if event.dispatching() {
+            match event.GetTarget() {
+                Some(target) => {
+                    if let Some(element) = target.downcast::<Element>() {
```
jdm: Same as previous comment.

components/script/dom/domrect.rs
```diff
@@ -32,6 +32,14 @@ impl DOMRect {
         )
     }
 
+    pub fn left(&self) -> f64 {
+        self.rect.Left()
+    }
+
+    pub fn top(&self) -> f64 {
```
jdm: We should be able to remove these based on my other comments.

jdm (changes requested):

jdm:
This looks good, it just needs a final `./mach update-manifest` to reflect the changes to the WPT test.

tigleym:
> This looks good, it just needs a final `./mach update-manifest` to reflect the changes to the WPT test.

Sounds good! I'll get this updated today.

PR #24525, opened 2019-10-23
Issue 23009/separate angle and disable vsync

### LINKED ISSUE #23009: Separate command line option parsing into global and embedding-specific options
Right now there are global options like `--disable-vsync`, `--headless`, `--multiprocess`, and the forthcoming `--angle` that set fields in the global `Options` structure, but are only used by the Glutin embedding for desktop builds. It would be nice to remove backend-specific options from the global set of engine options. Instead of the current `from_cmdline_args` API which creates an `Options` value and returns `ArgumentParsingResult`, we could have this:
```rust
fn from_cmdline_args(mut opts: getopts::Options, args: &[String]) -> getopts::Result
```
This would allow the embedder to provide an Options object with embedder-specific option parsing, which the engine then appends global option parsing to, parses the provided argument values, updates the global options with the resulting values, then returns a result which includes the parsed options that should be handled by the embedder.

There are two parts to this work:
1. modifying the API to support accepting an existing Options object and returning the parsing result
2. extracting the embedder-specific preferences from the global argument parsing code into the caller code in `ports/servo/non_android_main`

Code: `components/config/opts.rs`, `ports/servo/non_android_main.rs`, `ports/servo/glutin_app/window.rs`, `ports/libsimpleservo/api/src/lib.rs`

Tests: This command should continue to exit without popping a window after these changes - `./mach run -d --headless -x tests/html/about-mozilla.html`

glowe:
The `--angle` and `--disable-vsync` options were declared as global options, but only used in the Glutin embedding for desktop builds. Moving them to the Glutin embedding code makes them easier to update in the future.

I modified `opts::from_cmdline_args` to accept a `getopts::Options` (as prescribed in the issue) and augmented `opts::ArgumentParsingResult` to include an `opts::Matches` and `content-process` String when appropriate. I could use some feedback on this last bit. I could have changed the function to return `opts::Matches` and have the embedding code look for the presence of `content-process`, but I felt that the approach I went with was closer to the original design.

The other aspect I'm not sure about is moving `disable-vsync` from a global debug option to a plain embedder option. This changes the command line interface for glutin, which is maybe bad. However I wasn't sure whether it was worth preserving the original behavior given the complexity of injecting debug options into `opts::from_cmdline_args`.

- [x] These changes partially fix #23009 – there are 2 more options to deal with, but I'm not sure if we should handle them yet.
- [x] These changes do not require tests because this is a refactoring and I'm hoping that the existing tests cover these changes.

r? @jdm

### DIFF
--- a/ports/libmlservo/Cargo.toml
+++ b/ports/libmlservo/Cargo.toml
@@ -25,7 +25,6 @@ simpleservo = { path = "../libsimpleservo/api", features = ["no_static_freetype"
 rust-webvr = { version = "0.16", features = ["magicleap"] }
 webxr-api = { git = "https://github.com/servo/webxr", features = ["ipc"] }
 webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "magicleap"] }
-getopts = "0.2.11"
 libc = "0.2"
 log = "0.4"
 servo-egl = "0.2"
--- a/ports/libmlservo/Cargo.toml
+++ b/ports/libmlservo/Cargo.toml
@@ -25,6 +25,7 @@ simpleservo = { path = "../libsimpleservo/api", features = ["no_static_freetype"
 rust-webvr = { version = "0.16", features = ["magicleap"] }
 webxr-api = { git = "https://github.com/servo/webxr", features = ["ipc"] }
 webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "magicleap"] }
+getopts = "0.2.11"
 libc = "0.2"
 log = "0.4"
 servo-egl = "0.2"
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -9,6 +9,7 @@ pub mod gl_glue;
 
 pub use servo::script_traits::MouseButton;
 
+use getopts::Options;
 use servo::compositing::windowing::{
     AnimationState, EmbedderCoordinates, EmbedderMethods, MouseWindowEvent, WindowEvent,
     WindowMethods,
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -205,7 +205,7 @@ pub fn init(
         gl: gl.clone(),
     });
 
-    let servo = Servo::new(embedder_callbacks, window_callbacks.clone());
+    let servo = Servo::new(embedder_callbacks, window_callbacks.clone(), None);
 
     SERVO.with(|s| {
         let mut servo_glue = ServoGlue {
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -171,7 +171,7 @@ pub fn init(
             gfx.subpixel_text_antialiasing.enabled,
             init_opts.enable_subpixel_text_antialiasing
         );
-        opts::from_cmdline_args(&args);
+        opts::from_cmdline_args(Options::new(), &args);
     }
 
     let embedder_url = init_opts.url.as_ref().and_then(|s| ServoUrl::parse(s).ok());
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2673,6 +2673,7 @@ dependencies = [
 name = "libmlservo"
 version = "0.0.1"
 dependencies = [
+ "getopts",
  "libc",
  "libservo",
  "log",
@@ -4406,6 +4407,7 @@ dependencies = [
  "cc",
  "clipboard",
  "euclid",
+ "getopts",
  "gleam 0.6.18",
  "glutin",
  "image",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2673,7 +2673,6 @@ dependencies = [
 name = "libmlservo"
 version = "0.0.1"
 dependencies = [
- "getopts",
  "libc",
  "libservo",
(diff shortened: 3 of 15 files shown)


jdm:
I like these changes, and I appreciate the care you have taken in describing your reasoning for them! I think moving the disable-vsync flag to be a non-debug option is fine in this case. I've found several other flags that are glutin-only that we can convert as well: device_pixels_per_px, clean_shutdown, use_msaa, and no_native_titlebar.

glowe:
I extracted `clean_shutdown`, `use_msaa`, and `no_native_titlebar`. The `device_pixels_per_px` is a bit more difficult, because it's used in the Constellation pipeline:

https://github.com/servo/servo/blob/dc8be8f2825b7de873bf442aeb4a5dd0ed88d4b1/components/constellation/pipeline.rs#L612

jdm:
We can file a followup to remove the use from constellation by passing it as an argument to the constructor instead, if you'd like.

ports/libmlservo/Cargo.toml
```diff
@@ -25,6 +25,7 @@ simpleservo = { path = "../libsimpleservo/api", features = ["no_static_freetype"
 rust-webvr = { version = "0.16", features = ["magicleap"] }
 webxr-api = { git = "https://github.com/servo/webxr", features = ["ipc"] }
 webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "magicleap"] }
+getopts = "0.2.11"
```
jdm: Did this belong in libsimpleservo/api/Cargo.toml instead?

jdm (changes requested):
This looks great!

ports/libmlservo/Cargo.toml
glowe: Yes, you're right. Thanks for catching!

glowe:
> We can file a followup to remove the use from constellation by passing it as an argument to the constructor instead, if you'd like.

We can definitely make that change in this PR. However, I suspect that's going to force a modification of `servo::Servo::new`, which would be a public interface change. If so, does that mean we need to bump the version of `libservo`?

jdm:
libservo isn't published anywhere so we don't need to follow any semver rules.

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -159,6 +158,7 @@ pub fn init(
     gl: Rc<dyn gl::Gl>,
     waker: Box<dyn EventLoopWaker>,
     callbacks: Box<dyn HostTrait>,
+    device_pixels_per_px: Option<f32>,
```
jdm: This will require updating https://github.com/servo/servo/blob/6a54d600431621bfd4c4d4e49e7816e95deec68e/ports/libmlservo/src/lib.rs#L193 and https://github.com/servo/servo/blob/6a54d600431621bfd4c4d4e49e7816e95deec68e/ports/libsimpleservo/jniapi/src/lib.rs#L117 and https://github.com/servo/servo/blob/6a54d600431621bfd4c4d4e49e7816e95deec68e/ports/libsimpleservo/capi/src/lib.rs#L320.

However, I just noticed https://github.com/servo/servo/blob/6a54d600431621bfd4c4d4e49e7816e95deec68e/ports/libsimpleservo/api/src/lib.rs#L49 which is the value that is consulted when device_pixels_per_px is None in [this code](https://github.com/servo/servo/blob/master/components/compositing/compositor.rs#L1083-L1091). So I think we don't actually need to add this new argument; the device_pixels_per_px value is an override that is only meaningful in the glutin port, and the other ports already have a supported API for reporting the DPI factor, so we can revert this change to the `init` method.

glowe: Thanks for catching this. I removed this change and pushed the rebased commit without this change.

jdm:
@bors-servo r+
Looks great! Thanks for tackling this work!

glowe:
> Looks great! Thanks for tackling this work!

Thank you for your guidance. It was fun!

glowe:
> 💔 Test failed - status-taskcluster

Looks like these errors are caused by me failing to update some things in `libsimpleservo`, however `mach build -d` built cleanly on my machine. I assume I didn't catch this issue, because `mach build -d` doesn't compile this crate? Is there an alternate invocation I can run that would catch these types of errors before submitting again?

CYBAI:
@glowe I guess you're using macOS? maybe it's easier to fix them and push to the PR and ask homu to try 🤔? 

```rust
error[E0433]: failed to resolve: use of undeclared type or module `Options`
   --> ports/libsimpleservo/api/src/lib.rs:173:33
    |
173 |         opts::from_cmdline_args(Options::new(), &args);
    |                                 ^^^^^^^ use of undeclared type or module `Options`

error[E0061]: this function takes 3 parameters but 2 parameters were supplied
   --> ports/libsimpleservo/api/src/lib.rs:207:17
    |
207 |     let servo = Servo::new(embedder_callbacks, window_callbacks.clone());
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 3 parameters

```

jdm:
`mach build --libsimpleservo` is what you're looking for.

glowe:
Thanks for the help again. I believe I've addressed the libsimpleservo build errors.

jdm:
@bors-servo r+
Thanks!

PR #24526, opened 2019-10-23
Pass source origin in MessageEvent.

### LINKED ISSUE #22888: Use origin of source document for MessageEvent resulting from postMessage
Right now `Window::post_message` passes `None` as the value that will initialize `MessageEvent.origin`. [The spec](https://html.spec.whatwg.org/multipage/comms.html#dom-messageevent-origin) says this should be the origin of the source document that posted the message instead. We will want to add a new origin argument to `Window::post_message`, and store the origin in `ScriptMsg::PostMessage` as well so the code path in `DissimilarOriginWindow::post_message` also works.

This should allow tests/wpt/web-platform-tests/dom/events/EventListener-incumbent-global-1.sub.html to make more progress.

Code: `components/script/dom/window.rs`, `components/script/dom/dissimilaroriginwindow.rs`, `components/script/script_thread.rs`, `components/script_traits/lib.rs`
Test: `./mach test-wpt tests/wpt/web-platform-tests/dom/events/EventListener-incumbent-global-1.sub.html`

mmiecz:
Add source origin to message event.
This PR makes `tests/wpt/web-platform-tests/dom/events/EventListener-incumbent-global-1.sub.html` to pass

---

- [X] These changes fix #22888  (GitHub issue number if applicable)

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -1562,10 +1562,17 @@ where
             FromScriptMsg::PostMessage {
                 target: browsing_context_id,
                 source: source_pipeline_id,
+                source_origin: source_origin,
                 target_origin: origin,
                 data,
             } => {
-                self.handle_post_message_msg(browsing_context_id, source_pipeline_id, origin, data);
+                self.handle_post_message_msg(
+                    browsing_context_id,
+                    source_pipeline_id,
+                    source_origin,
+                    origin,
+                    data,
+                );
             },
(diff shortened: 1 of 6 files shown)


jdm:
This needs to be rebased against a recent master. I suspect that #23637 has merged and caused conflicts.

mmiecz:
As @gterzian said here: https://github.com/servo/servo/issues/22888#issuecomment-545851455 this is already done, so I'm closing this PR.

PR #24534, opened 2019-10-24
Launch UWP app from command line

ferjm:

This PR makes it possible to launch the UWP app from the command line by executing (on any path) the `Servo` command. It requires building and launching Servo once on VS, so the command alias is properly registered.

It also allows passing command line arguments. So we can do for example `Servo --pref dom.webxr.enabled`.

Debugging is also possible when launching the app from the command line. It requires setting the `Launch Application` option (right click on `ServoApp (Universal Windows`) on the Solution Explorer view -> Properties ->  Configuration Properties -> Debugging) to `No`.

### DIFF
--- a/support/hololens/ServoApp/App.cpp
+++ b/support/hololens/ServoApp/App.cpp
@@ -32,9 +32,8 @@ App::App() {
 }
 
 void App::createRootFrame(
-    bool prelaunchActivated,
+    Frame &rootFrame, bool prelaunchActivated,
     winrt::Windows::Foundation::IInspectable const &args) {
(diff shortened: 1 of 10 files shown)


paulrouget:
Aren't we gonna lose the `dom.webxr.enabled` pref when deploying from VS?

ferjm:
Yes. I just fixed it with the last commit.

paulrouget:
Works well :) Thank you.

@bors-servo r+

jdm:
@bors-servo retry
#23290

PR #7601, opened 2015-09-11
Improve DOM interfaces' extended attributes

nox:
Some should have been `[SameObject]` instead of `[Constant]`. The rest of the changes are additional `[Constant]` and `[Pure]` extended attributes on many operations.

[Review on Reviewable](https://reviewable.io/reviews/servo/servo/7601)

PR #24544, opened 2019-10-25
Support IDL stringifier attributes

### LINKED ISSUE #7590: Support for WebIDL stringifier attributes
[Related Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=824857)

[Spec link](https://heycam.github.io/webidl/#idl-stringifiers) - Starting with "The stringifier keyword can also be placed on an attribute."

[Example in URLUtils.webidl](https://github.com/servo/servo/blob/347e9b6ef45fb8b0e5a824944e4b1c7338d15554/components/script/dom/webidls/URLUtils.webidl#L9)

saschanaz:
---

- [x] These changes fix #7590

### DIFF
--- a/components/script/dom/webidls/DOMTokenList.webidl
+++ b/components/script/dom/webidls/DOMTokenList.webidl
@@ -22,8 +22,7 @@ interface DOMTokenList {
   void replace(DOMString token, DOMString newToken);
 
   [CEReactions, Pure]
-           attribute DOMString value;
+  stringifier attribute DOMString value;
 
-  stringifier;
   iterable<DOMString?>;
 };
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -3667,6 +3667,8 @@ class CGSpecializedMethod(CGAbstractExternMethod):
 
     @staticmethod
     def makeNativeName(descriptor, method):
+        if method.underlyingAttr:
+            return CGSpecializedGetter.makeNativeName(descriptor, method.underlyingAttr)
         name = method.identifier.name
         nativeName = descriptor.binaryNameFor(name)
         if nativeName == name:
@@ -5762,7 +5764,7 @@ class CGInterfaceTrait(CGThing):
             for m in descriptor.interface.members:
                 if (m.isMethod() and not m.isStatic() and
                         not m.isMaplikeOrSetlikeOrIterableMethod() and
-                        (not m.isIdentifierLess() or m.isStringifier()) and
+                        (not m.isIdentifierLess() or (m.isStringifier() and not m.underlyingAttr)) and
                         not m.isDefaultToJSON()):
                     name = CGSpecializedMethod.makeNativeName(descriptor, m)
                     infallible = 'infallible' in descriptor.getExtendedAttributes(m)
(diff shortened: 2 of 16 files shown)


components/script/dom/webidls/DOMTokenList.webidl
```diff
@@ -21,9 +21,8 @@ interface DOMTokenList {
   [CEReactions, Throws]
   void replace(DOMString token, DOMString newToken);
 
-  [CEReactions, Pure]
-           attribute DOMString value;
+  [CEReactions]
```
saschanaz: Removed `Pure` because WebIDL.py currently does not allow it on stringifier attribute (just because gecko didn't have it on any of its stringifier attributes).

The support can be added, but not sure why servo has Pure but gecko does not. @nox, could you give me some idea?

nox (commented):
Where do the WebIDL parser and tests changes come from? Upstream? If not, you are not supposed to change those files.

saschanaz:
> Where do the WebIDL parser and tests changes come from? Upstream? If not, you are not supposed to change those files.

From https://phabricator.services.mozilla.com/D48355 😀

components/script/dom/webidls/DOMTokenList.webidl
saschanaz: Thanks for the docs, but I mean why DOMTokenList in Gecko does not have Pure.

saschanaz: In other words, should we upstream the changes in #7601 ?

nox: No clue, I don't know if that conflicts with how stringifiers are handled through SpiderMonkey. Or it could be a case of no one adding support for `[Pure]` on them in the Gecko side.

nox:
Cool. Then please make a first commit with the vendoring of those new changes (with the `update.sh` script) and then a second commit with the rest of the changes.

components/script/dom/webidls/DOMTokenList.webidl
saschanaz: There is no blocker to add `[Pure]` support on stringifier attributes, it's just that no one was bothered to add one on DOMTokenList or to do what #7601 did.

@bzbarsky, thoughts?

jdm: I'm not actually confident that Pure has any meaning in Servo's implementation. It may have been copied from upstream webidl in Gecko.

nox: > I'm not actually confident that Pure has any meaning in Servo's implementation. It may have been copied from upstream webidl in Gecko.

`WebIDL.py` translates `[Pure]` and `[Constant]` to specific `[Affects]`/`[DependsOn]` combinations, and you can actually find references to those in `CodegenRust.py`.

nox: (Submitted too fast.)

And in #7601 I went out of my way to make use of them more extensively so those weren't there by coincidence.

bzbarsky: Fwiw, I think we should add support for [Pure] in the parser here and probably add it in the Gecko IDL.

jdm (approved):

Issue #24530, opened 2019-10-23
Babylon color picker demo requires two clicks to choose a colour


jdm:
In http://paulrouget.com/test/bbjs/basic/ when I run it in Firefox I can choose a colour with one click. In Servo it requires two clicks each time.

paulrouget:
The issue was coming from a workaround I put in place to avoid https://github.com/servo/servo/issues/24363 - this is now fixed in the demo.

Issue #24363, opened 2019-10-03
Can't move camera in babylon demos


jdm:
```
[2019-10-03T19:24:15Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:234 setting getter-only property "buttons"
[2019-10-03T19:24:15Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:15Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:15Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:16Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:16Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:16Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:16Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:17Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:153:28 setting getter-only property "buttons"
[2019-10-03T19:24:26Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:26Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[2019-10-03T19:24:26Z ERROR script::dom::bindings::error] Error at https://code.jquery.com/pep/0.4.2/pep.min.js:149:48 setting getter-only property "buttons"
[...]

PR #24547, opened 2019-10-25
UWP: better mouse interaction support

paulrouget:


jdm:
I filed #24530 after observing it in a desktop glutin window - are there similar changes we need to make to that port as well?

paulrouget:
> I filed #24530 after observing it in a desktop glutin window - are there similar changes we need to make to that port as well?

Good to know. This is actually a different issue with the pointer event polyfil. Which I worked around in the latest bbjs demo update.

This PR doesn't fix this issue for desktop, but does fix the behavior for UWP builds.

paulrouget:
The bbjs demos now use a better workaround for #24363. The double click issue (#24530) doesn't exist anymore.

But still need that fix for the non-glutin port.

support/hololens/ServoApp/ServoControl/ServoControl.cpp
```diff
@@ -93,15 +96,50 @@ void ServoControl::OnSurfaceManipulationDelta(
...
+    }
+
+    if (!b.has_value() && mPressedMouseButton.has_value()) {
+      auto b = *mPressedMouseButton;
+      mPressedMouseButton = {};
+      RunOnGLThread([=] { mServo->MouseUp(x, y, b); });
+      e.Handled(true);
+    }
+
+    if (b.has_value()) {
```
jdm: Won't this also trigger when we've just run a MouseUp event in the previous branch?

jdm (changes requested):

support/hololens/ServoApp/ServoControl/ServoControl.cpp
paulrouget: It won't. I think the confusion comes from the very non descriptive variable name (`b`) which is used twice. I will change that.

Issue #23220, opened 2019-04-16
Missing gstreamer plugins on CI


jdm:
```
  │ (servo:27338): GStreamer-WARNING **: External plugin loader failed. This most likely means that the plugin loader helper binary was not found or could not be run. You might need to set the GST_PLUGIN_SCANNER environment variable if your setup is unusual. This should normally not be required though.
  │ 0:00:00.028966467 27338 0x7f2265531780 WARN      GST_PLUGIN_LOADING gstplugin.c:792:_priv_gst_plugin_load_file_for_registry: module_open failed: libmms.so.0: cannot open shared object file: No such file or directory
  │ 
  │ (servo:27338): GStreamer-WARNING **: Failed to load plugin '/home/servo/gst/lib/gstreamer-1.0/libgstmms.so': libmms.so.0: cannot open shared object file: No such file or directory
  │ 0:00:00.029185843 27338 0x7f2265531780 WARN      GST_PLUGIN_LOADING gstplugin.c:792:_priv_gst_plugin_load_file_for_registry: module_open failed: libfaad.so.2: cannot open shared object file: No such file or directory
  │ 
  │ (servo:27338): GStreamer-WARNING **: Failed to load plugin '/home/servo/gst/lib/gstreamer-1.0/libgstfaad.so': libfaad.so.2: cannot open shared object file: No such file or directory
  │ 0:00:00.029414027 27338 0x7f2265531780 WARN      GST_PLUGIN_LOADING gstplugin.c:792:_priv_gst_plugin_load_file_for_registry: module_open failed: libcairo-gobject.so.2: cannot open shared object file: No such file or directory
  │ 
  │ (servo:27338): GStreamer-WARNING **: Failed to load plugin '/home/servo/gst/lib/gstreamer-1.0/libgstcairo.so': libcairo-gobject.so.2: cannot open shared object file: No such file or directory
  │ 0:00:00.029625762 27338 0x7f2265531780 WARN      GST_PLUGIN_LOADING gstplugin.c:792:_priv_gst_plugin_load_file_for_registry: module_open failed: libmpg123.so.0: cannot open shared object file: No such file or directory
  │ 
  │ (servo:27338): GStreamer-WARNING **: Failed to load plugin '/home/servo/gst/lib/gstreamer-1.0/libgstmpg123.so': libmpg123.so.0: cannot open shared object file: No such file or directory
```

jdm:
libmms.so.0
libfaad.so.2
libcairo-gobject.so.2
[...]

PR #24552, opened 2019-10-25
mach fmt calls clang-format

### LINKED ISSUE #24031: Do something about HoloLens C++ code formating
Visual Studio can format the whole file based on the [clang-format file](https://github.com/servo/servo/blob/master/support/hololens/.clang-format).

It's not done automatically, but it, sometimes, happens (on pasting for example).

That means we get some random syntax changes in pull requests. For example: https://github.com/servo/servo/pull/24030

We should enable tidy for hololens code, and figure out the right formatting.

paulrouget:
Fix #24031

What would be the right way to also integrate this with tidy?


SimonSapin:
Does clang-format have something like `rustfmt --check`, which prints a diff and returns an error exit code if it would make any change? `./mach test-tidy` does this here:

https://github.com/servo/servo/blob/b7b1b903d33278fa866574c2a3dfd8c2be559fb0/python/servo/testing_commands.py#L335-L338

paulrouget:
> Does clang-format have something like rustfmt --check

No. clang-format is very limited. I'm gonna have to implement that manually.

Also, note to self, clang-format is not guaranteed to be installed on Mac and Linux.

paulrouget:
There's an option to get a XML output. I will use that.

paulrouget:
@SimonSapin r?

paulrouget:
We will need to install clang-format on the CI machines.

SimonSapin:
Since this requires an externally-installed tool I’d prefer we make this either opt-in or opt-out.

How fast is clang-format? If not very, can we only run it on files that have been changed [like tidy](https://github.com/servo/servo/blob/6f4c6712cba4b6a72f93d6ed776263a0243a3196/python/tidy/servo_tidy/tidy.py#L161-L165)

Dependencies for Linux CI tasks are installed here https://github.com/servo/servo/blob/master/etc/taskcluster/docker/build.dockerfile. Modifying this file in your PR should Just Work.

r? @jdm

paulrouget:
Note to self:
- make it optin or optout
- check speed
- version differences on Windows and MacOS (6 vs 9)
- update etc/taskcluster/docker/build.dockerfile and etc/taskcluster/macos/Brewfile

SimonSapin:
> update etc/taskcluster/docker/build.dockerfile and etc/taskcluster/macos/Brewfile

Do we need both? Isn’t it enough if CI runs clang-format once on all source files?

paulrouget:
I change the logic a bit:
- if `--no-cpp` is passed, clang-format is not used
- if `clang-format` is not available, or has the wrong version (not v6), it skips the check (no failure) and print a message

There's an issue where Visual Studio use clang-format 6, not 9 (the one available on MacOS and probably Linux). And the clang files are not compatible (that's why I'm checking the clang-format version).

As for the speed, now it only scan files that have changed. It's fast.

I am not changing the CI configuration for osx or linux, as I think it should only be tested on Windows, where the supported version is supposed to be installed. Also, for now, we only check HoloLens code, so it makes sense to support the Windows version. I'm not sure though how my Visual Studio version got `clang-format`. And I don't know if the CI will have it. Will see from the logs.

paulrouget:
@jdm r?

python/servo/testing_commands.py
```diff
@@ -774,6 +797,25 @@ def create_parser_create():
...
+        print(version)
+        if not version.startswith("clang-format version 6."):
+            print("clang-format: wrong version (v6 required). Skipping CPP formatting.")
+            return False, None, None
+    except OSError:
+        print("clang-format not installed. Skipping CPP formatting.")
+        return False, None, None
+    gitcmd = ['git', 'ls-files']
+    if not all_files:
+            gitcmd.append('-m')
```
jdm: nit: unindent this one level.

jdm (approved):

CYBAI:
@bors-servo retry

-  #23220

CYBAI:
@bors-servo retry

- #24622

PR #24557, opened 2019-10-26
Upgrade to rustc 1.40.0-nightly (246be7e1a 2019-10-25)

SimonSapin:


jdm:
script_plugins isn't building correctly.

SimonSapin:
This is because compiler crates used by compiler plugins have been moved out of the `rust-std` component in rustup and into a new `rustc-dev` component. I’ve pushed another commit to install that.

jdm:
The scrupt_plugins for the taskcluster tidy/unit job is still broken.

SimonSapin:
Ah, pre-review testing of PRs was yet another code path in `decision_task.py`. That’s a bit unfortunate.

Issue #21476, opened 2018-08-22
Support asynchronous WebAssembly compilation


fabricedesre:
This is a web assembly game demo.

STR:
- Open https://www.funkykarts.rocks/demo.html 

Expected:
- you have fun!

Observed:
- You are sad: `WebAssembly Promise APIs not supported in this runtime`

I guess these are the streaming apis from https://webassembly.github.io/spec/web-api/index.html

jdm:
This looks like something we need to configure properly to support: https://github.com/servo/mozjs/blob/9b44b1e870cae2d96ea41bf2aa8a7d5eaa4cdce7/mozjs/js/src/wasm/WasmJS.cpp#L2337-L2345

jdm:
https://github.com/servo/mozjs/blob/9b44b1e870cae2d96ea41bf2aa8a7d5eaa4cdce7/mozjs/js/src/jsapi.h#L4484-L4521

jdm:
Additionally:
https://github.com/servo/mozjs/blob/9b44b1e870cae2d96ea41bf2aa8a7d5eaa4cdce7/mozjs/js/src/jsapi.h#L4523-L4568
https://searchfox.org/mozilla-central/rev/3fa761ade83ed0b8ab463acb057c2cf0b104689e/dom/base/nsJSEnvironment.cpp#2660-2667
https://searchfox.org/mozilla-central/rev/3fa761ade83ed0b8ab463acb057c2cf0b104689e/dom/fetch/FetchUtil.cpp#528-611
https://searchfox.org/mozilla-central/rev/3fa761ade83ed0b8ab463acb057c2cf0b104689e/dom/fetch/FetchUtil.cpp#335-516

asajeffrey:
Now happens in the tanks wasm demo. This used to work, did the demo get upgraded to use promises at some point?
```
$ ./mach run -d https://webassembly.org/demo
ERROR 2018-11-06T16:28:06Z: offscreen_gl_context::draw_buffer: The given GLContext doesn't support requested antialising
Loading blob blob:https://webassembly.org/dd95f4ece2b9411dbf9d56db995b10b1
Loading blob blob:https://webassembly.org/c7a0ecbc96f3434bb8d339fb0c6856fc
trying binaryen method: native-wasm
asynchronously preparing wasm
[...]

PR #24563, opened 2019-10-28
Binding stream consumer to support asynchronous WebAssembly compilation

Akash-Pateria:
This PR contains changes related to binding the Stream Consumer wrapper for supporting async WebAssembly compilation in Servo. The changes are listed below:

- Added a StreamConsumer wrapper and implemented its corresponding behaviors such as consumeChunk, streamEnd, streamError, and noteResponseURLs.
- Added ReportStreamErrorCallback for reporting the CompileError/TypeError occurred during the compilation.

---
- [X] `./mach build -d` does not report any errors, however, it has a few warnings of unused functions since we are calling them as yet.
- [X] These changes are part of #21476 fix

### DIFF
--- a/components/script/script_runtime.rs
+++ b/components/script/script_runtime.rs
@@ -5,6 +5,8 @@
 //! The script runtime contains common traits and structs commonly used by the
 //! script thread, the dom, and the worker threads.
 
+#![allow(dead_code)]
+
 use crate::dom::bindings::codegen::Bindings::PromiseBinding::PromiseJobCallback;
 use crate::dom::bindings::conversions::get_dom_class;
 use crate::dom::bindings::conversions::private_from_object;
@@ -26,7 +28,10 @@ use crate::script_thread::trace_thread;
 use crate::task::TaskBox;
 use crate::task_source::{TaskSource, TaskSourceName};
 use js::glue::{CollectServoSizes, CreateJobQueue, DeleteJobQueue, JobQueueTraps, SetBuildId};
+use js::glue::{RUST_js_GetErrorMessage, StreamConsumerConsumeChunk, StreamConsumerStreamEnd};
+use js::glue::{StreamConsumerNoteResponseURLs, StreamConsumerStreamError};
 use js::jsapi::ContextOptionsRef;
+use js::jsapi::StreamConsumer as JSStreamConsumer;
 use js::jsapi::{BuildIdCharVector, DisableIncrementalGC, GCDescription, GCProgress};
 use js::jsapi::{HandleObject, Heap, JobQueue};
 use js::jsapi::{JSContext as RawJSContext, JSTracer, SetDOMCallbacks, SetGCSliceCallback};
@@ -51,6 +56,7 @@ use profile_traits::mem::{Report, ReportKind, ReportsChan};
 use servo_config::opts;
 use servo_config::pref;
 use std::cell::Cell;
+use std::ffi::CString;
 use std::fmt;
 use std::io::{stdout, Write};
 use std::ops::Deref;
@@ -775,4 +781,54 @@ impl Deref for JSContext {
     fn deref(&self) -> &Self::Target {
         &self.0
     }
+}
+
+pub struct StreamConsumer(*mut JSStreamConsumer);
+
+#[allow(unsafe_code)]
+impl StreamConsumer {
+    fn consume_chunk(&self, stream: &[u8]) -> bool {
+        unsafe {
+            let stream_ptr = stream.as_ptr();
+            return StreamConsumerConsumeChunk(self.0, stream_ptr, stream.len());
+        }
+    }
+
+    fn stream_end(&self) {
+        unsafe {
+            StreamConsumerStreamEnd(self.0);
+        }
+    }
+
+    fn stream_error(&self, error_code: usize) {
+        unsafe {
+            StreamConsumerStreamError(self.0, error_code);
+        }
+    }
+
+    fn note_response_urls(&self, maybe_url: Option<String>, maybe_source_map_url: Option<String>) {
+        unsafe {
+            let maybe_url = maybe_url.map(|url| CString::new(url).unwrap());
+            let maybe_source_map_url = maybe_source_map_url.map(|url| CString::new(url).unwrap());
+
+            let maybe_url_param = match maybe_url.as_ref() {
+                Some(url) => url.as_ptr(),
+                None => ptr::null(),
+            };
+            let maybe_source_map_url_param = match maybe_source_map_url.as_ref() {
+                Some(url) => url.as_ptr(),
+                None => ptr::null(),
+            };
+
+            StreamConsumerNoteResponseURLs(self.0, maybe_url_param, maybe_source_map_url_param);
+        }
+    }
+}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn report_stream_error_callback(_cx: *mut JSContext, error_code: usize) {
+    error!(
+        "Error initializing StreamConsumer: {:?}",
+        RUST_js_GetErrorMessage(ptr::null_mut(), error_code as u32)
+    );
 }


components/script/script_runtime.rs
```diff
@@ -776,3 +780,44 @@ impl Deref for JSContext {
...
+            StreamConsumerNoteResponseURLs(sc, maybeUrl, maybeSourceMapUrl);
+        }
+    }
+}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn ReportStreamErrorCallback(cx: *mut JSContext, errorCode: usize){
+    error!(
+        "Error initializing StreamConsumer: {:?}",
+        RUST_js_GetErrorMessage(cx as *mut c_void, errorCode as u32)
```
jdm: The first argument to this API isn't actually used (https://searchfox.org/mozilla-central/rev/74cc0f4dce444fe0757e2a6b8307d19e4d0e0212/js/src/vm/JSContext.cpp#995) so let's pass `ptr::null_mut()` instead.

components/script/script_runtime.rs
```diff
@@ -776,3 +780,44 @@ impl Deref for JSContext {
         &self.0
     }
 }
+
+pub struct StreamConsumer(*mut JSStreamConsumer);
+
+#[allow(unsafe_code)]
+impl StreamConsumer {
+    fn consumeChunk(sc: &mut JSStreamConsumer, begin: *const u8, length: usize) -> bool {
```
jdm: Let's use a safer [Rust slice](https://doc.rust-lang.org/book/ch04-03-slices.html#other-slices) here instead of separate pointer and length arguments. We'll want `&[u8]`, and we can extract the pointer and length values from that: https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_ptr

components/script/script_runtime.rs
```diff
@@ -776,3 +780,44 @@ impl Deref for JSContext {
...
+        }
+    }
+
+    fn streamEnd(sc: &mut JSStreamConsumer){
+        unsafe {
+            StreamConsumerStreamEnd(sc);
+        }
+    }
+
+    fn streamError(sc: &mut JSStreamConsumer, errorCode: usize){
```
jdm: All of these methods should accept a `&self` argument instead of `sc: &mut JSStreamConsumer`, and pass `self.0` as the argument to the JSAPI function that is being called instead. This will allow us to use the objects like `some_stream_consumer.streamError(error_code)` from Rust code.

components/script/script_runtime.rs
```diff
@@ -776,3 +780,44 @@ impl Deref for JSContext {
...
+        unsafe {
+            StreamConsumerStreamError(sc, errorCode);
+        }
+    }
+
+    fn noteResponseURLs(
+        sc: &mut JSStreamConsumer,
+        maybeUrl: *const c_char,
+        maybeSourceMapUrl: *const c_char
```
jdm: Let's use `Option<String>` arguments here. We can convert them to C strings with https://doc.rust-lang.org/stable/std/ffi/struct.CString.html, or use `ptr::null()` if there are [None](https://doc.rust-lang.org/stable/std/option/enum.Option.html) values instead of Some values present.

components/script/script_runtime.rs
```diff
@@ -776,3 +780,44 @@ impl Deref for JSContext {
...
+
+#[allow(unsafe_code)]
+impl StreamConsumer {
+    fn consumeChunk(sc: &mut JSStreamConsumer, begin: *const u8, length: usize) -> bool {
+        unsafe {
+            return StreamConsumerConsumeChunk(sc, begin, length);
+        }
+    }
+
+    fn streamEnd(sc: &mut JSStreamConsumer){
```
jdm: Let's give these methods `snake_case` names instead of `camelCase` ones.

jdm (changes requested):
Good start!

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
shubham-pampattiwar: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
Akash-Pateria: Done. Please review

components/script/script_runtime.rs
```diff
@@ -798,27 +799,46 @@ impl StreamConsumer {
...
-        maybeSourceMapUrl: *const c_char
+        maybe_url: Option<String>,
+        maybe_source_map_url: Option<String>
     ) {
         unsafe {
-            StreamConsumerNoteResponseURLs(self.0, maybeUrl, maybeSourceMapUrl);
+            let maybe_url_param: *const c_char = match maybe_url {
+                Some(maybe_url) => {
+                    let c_str_url = CString::new(maybe_url).unwrap();
+                    c_str_url.as_ptr()
```
jdm: Unfortunately this isn't safe, since we store a pointer to a value that is destroyed when we leave this match scope. We need to do the following instead:
```rust
let maybe_url = maybe_url.map(|url| CString::new(url).unwrap());
let maybe_url_param = match maybe_url.as_ref() {
    Some(url) => url.as_ptr(),
    None => ptr::null(),
};
```

components/script/script_runtime.rs
```diff
@@ -798,27 +799,46 @@ impl StreamConsumer {
...
+                },
+                None => ptr::null(),
+            };
+            let maybe_source_map_url_param: *const c_char = match maybe_source_map_url {
```
jdm: Same as above.

jdm (changes requested):

jdm:
You will also need to do the following:
* run `./mach fmt` to address code formatting issues
* squash your changes into a single commit
* rebase your changes against a recent revision from upstream (see step 7 of https://github.com/servo/servo/wiki/Github-workflow) and address merge conflicts

components/script/script_runtime.rs
Akash-Pateria: Yeah. You're right. I'll make the changes.

Would it also work if I first convert Option<String> to Option<CString>, then just do the match?

```rust
let maybe_url = maybe_url.as_ref().map(|url| CString::new(url).unwrap());
let maybe_url_param = match maybe_url {
    Some(url) => url.as_ptr(),
    None => ptr::null(),
};
```

jdm: I believe so!

Akash-Pateria:
@jdm I have rebased our changes with the latest revision from upstream.

components/script/script_runtime.rs
```diff
@@ -776,3 +781,53 @@ impl Deref for JSContext {
         &self.0
     }
 }
+
+pub struct StreamConsumer(*mut JSStreamConsumer);
```
jdm: Sorry, one last change that will be necessary - add a `#![allow(dead_code)]` at the top of the file after the license header comment. Otherwise our CI will complain because these new methods and functions are not actually called by any code.

Akash-Pateria: Done.

components/script/script_runtime.rs
```diff
@@ -1,3 +1,4 @@
+#[allow(dead_code)]
```
jdm: I said:
* `#![allow(dead_code)` (this has a different meaning than `#[allow(dead_code)]`
* place it after the license header comment

This doesn't build right now.

Akash-Pateria: I am sorry, I'd misread the comments. I have fixed that lint and updated the PR.

jdm:
Looks like there's still a `mach fmt` required, according to the CI results.

jdm:
@bors-servo r+
Thanks!

PR #24576, opened 2019-10-29
Implement srcdoc support for iframes

### LINKED ISSUE #4767: Implement srcdoc support for iframes
Explanation: https://bocoup.com/blog/third-party-javascript-development-future#iframe-srcdoc
Spec:
* https://html.spec.whatwg.org/multipage/embedded-content.html#attr-iframe-srcdoc
* https://html.spec.whatwg.org/multipage/iframe-embed-object.html#process-the-iframe-attributes

Code:
* `components/script/dom/htmliframeelement.rs`
* `components/script/script_thread.rs`

Basic test:
`./mach test-wpt tests/wpt/web-platform-tests/html/semantics/embedded-content/the-iframe-element/srcdoc_process_attributes.html`

amj23897:
This PR contains changes related to adding srcdoc attribute parsing support for iframes in Servo. The following changes have been made:

- uncomment the [srcdoc](https://github.com/servo/servo/blob/f63b404e0cbf30380c4043700861110d06e548bb/components/script/dom/webidls/HTMLIFrameElement.webidl#L10-L11) WebIDL attribute, and implement the attribute getter.
- add a field to [LoadData](https://github.com/servo/servo/blob/e6b271d32981ad95442c029bf72bd35efc88f9c5/components/script_traits/lib.rs#L137-L164) for storing the srcdoc contents when loading a srcdoc iframe.
- (partially) implemented a new `page_load_about_srcdoc` method to `script_thread.rs` which loads the special `about:srcdoc` URL [per the specification](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#process-the-iframe-attributes) and takes the srcdoc contents as an argument
- call this new method from [handle_new_layout](https://github.com/servo/servo/blob/e6b271d32981ad95442c029bf72bd35efc88f9c5/components/script/script_thread.rs#L2409-L2412) when it's detected that a srcdoc iframe is being loaded
- (partially) in [attribute_mutated](https://github.com/servo/servo/blob/e6b271d32981ad95442c029bf72bd35efc88f9c5/components/script/dom/htmliframeelement.rs#L560), ensure that changing the `srcdoc` attribute of an iframe element [follows the specification](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element:the-iframe-element-9).

---

- [x] These changes are a part of fix #4767  (GitHub issue number if applicable)


jdm:
r? @jdm
I'll answer your questions tomorrow morning! Looks like a good start.

components/script/dom/htmliframeelement.rs
```diff
@@ -228,6 +228,8 @@ impl HTMLIFrameElement {
     /// <https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes>
     fn process_the_iframe_attributes(&self, mode: ProcessingMode) {
         // TODO: srcdoc
+        // Still remaining because we are understanding the specification:
+        // https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes
```
jdm: You will want to:
* get the srcdoc attribute value
* if the srcdoc attribute value exists, create a LoadData instance that contains the attribute's value and the url about:srcdoc
* call self.navigate_or_reload_child_browsing_context with the new loaddata
* return without executing the rest of the method

components/script/dom/htmliframeelement.rs
```diff
@@ -578,6 +586,18 @@ impl VirtualMethods for HTMLIFrameElement {
...
                     }));
             },
+            &local_name!("srcdoc") => {
+                // https://html.spec.whatwg.org/multipage/#the-iframe-element:the-iframe-element-9
+                // Whenever an iframe element with a non-null nested browsing context has its
+                // srcdoc attribute set, changed, or removed, the user agent must process the
+                // iframe attributes.
+
+                // < Q1 > Not sure whether following if statement is required
+                // if self.upcast::<Node>().is_connected_with_browsing_context() {
```
jdm: Yes, this is necessary for the same reason that the code dealing with "src" modifications uses it.

components/script/script_thread.rs
```diff
@@ -2429,6 +2429,9 @@ impl ScriptThread {
         );
         if load_data.url.as_str() == "about:blank" {
             self.start_page_load_about_blank(new_load, load_data.js_eval_result);
+        } else if load_data.url.as_str() == "about:srcdoc" {
+            // < Q2 > Which parameters should be passed to the following method?
+            self.page_load_about_srcdoc(load_data);
```
jdm: This is a very similar operation to start_page_load_about_black, so I would expect `new_load` and `load_data.srcdoc`.

components/script/script_thread.rs
```diff
@@ -3828,6 +3831,20 @@ impl ScriptThread {
         context.process_response_eof(Ok(ResourceFetchTiming::new(ResourceTimingType::None)));
     }
 
+    // refer definition of pre_page_load method in "script_thread.rs"; code taken from "start_page_load_about_blank"
+    fn page_load_about_srcdoc(&self, load_data: LoadData) {
```
jdm: You will likely end up with a method that is 90% similar to start_page_load_about_blank. We will use the srcdoc string as the initial content for the parser (in process_response_chunk), and we'll use about:srcdoc instead of about:blank, but otherwise I suspect the rest can be used unchanged.

jdm (changes requested):

components/script/script_thread.rs
cagandhi: Modified this method. Please review it.

components/script/dom/htmliframeelement.rs
cagandhi: Done.

components/script/script_thread.rs
jaymodi98: Done. Please review it.

components/script/dom/htmliframeelement.rs
amj23897: Implemented the requested changes. It'd be really helpful if you could review it as tests are still failing.

components/script/dom/htmliframeelement.rs
```diff
@@ -228,6 +233,31 @@ impl HTMLIFrameElement {
...
+                LoadOrigin::Script(document.origin().immutable().clone()),
+                url,
+                pipeline_id,
+                Some(Referrer::ReferrerUrl(document.url())),
+                document.get_referrer_policy(),
+            );
+            load_data.srcdoc = self.get_srcdoc();
+            self.navigate_or_reload_child_browsing_context(
+                load_data,
+                NavigationType::Regular,
```
jdm: Ah, here's the missing piece - in `navigate_or_reload_child_browsing_context` there's a special branch for NavigationType::InitialAboutBlank which ends up calling ScriptThread::handle_new_layout. If you pass that instead of NavigationType::Regular you should end up with more meaningful test results.

jdm: The reason this matters is that Regular causes us to go to the network thread and try to make a request for about:srcdoc, which is then not a recognized URL so we get an error page. To quickly verify if this is working correctly or not, use this test page:
```html
<iframe srcdoc="<div>hi!</div><script>foo=5;</script>" onload="console.log(this.contentWindow.foo)"></iframe>
```
In the current branch, I see an iframe containing an error message. With the change I suggested, there should be an iframe with "hi!" and the terminal should show `5`.

amj23897: I have done the change. But not tested yet.

amj23897: Done. I can see the iframe and terminal shows 5. But main tests are failing. The command used: 
`/mach test-wpt tests/wpt/web-platform-tests/html/semantics/embedded-content/the-iframe-element/srcdoc_process_attributes.html`

jdm:
Note to self: https://build.servo.org/builders/linux-rel-wpt/builds/14051 contains more tests that had unexpected results.

jdm:
With respect to the failing test - it's a bit more subtle than "the test still fails". I have asked the original test author for more information about why the test is written the way it is [in this comment](https://github.com/web-platform-tests/wpt/pull/17437#issuecomment-549563144). One point of success:
```
  UNEXPECTED-OK /html/semantics/embedded-content/the-iframe-element/srcdoc_process_attributes.html
```
This means that the test used to timeout because srcdoc wasn't supported, and now it runs to completion while some of the subtests still fail. That's progress :)

The other two failing subtests in that test fail with a JS exception: "iframe.contentDocument is null". I looked into this, and we can address this by adding a check for "about:srcdoc" to [this code](https://github.com/servo/servo/blob/1d450ba1f30b970d5ae007f42139190f8d265ce1/components/script/script_thread.rs#L3179) that override's the new document's origin. This ensures that the srcdoc document that is created shares the creator document's origin; the `iframe.contentDocument is null` exception occurred because the two documents did not share the same origin, so the JS code is not allowed to inspect the cross-origin document's contents.

cagandhi:
> With respect to the failing test - it's a bit more subtle than "the test still fails". I have asked the original test author for more information about why the test is written the way it is [in this comment](https://github.com/web-platform-tests/wpt/pull/17437#issuecomment-549563144). One point of success:
> 
> ```
>   UNEXPECTED-OK /html/semantics/embedded-content/the-iframe-element/srcdoc_process_attributes.html
> ```
> 
> This means that the test used to timeout because srcdoc wasn't supported, and now it runs to completion while some of the subtests still fail. That's progress :)
> 
> The other two failing subtests in that test fail with a JS exception: "iframe.contentDocument is null". I looked into this, and we can address this by adding a check for "about:srcdoc" to [this code](https://github.com/servo/servo/blob/1d450ba1f30b970d5ae007f42139190f8d265ce1/components/script/script_thread.rs#L3179) that override's the new document's origin. This ensures that the srcdoc document that is created shares the creator document's origin; the `iframe.contentDocument is null` exception occurred because the two documents did not share the same origin, so the JS code is not allowed to inspect the cross-origin document's contents.

Added a check on the line no. `3179` in `script_thread.rs` that clones `incomplete.url` into variable `origin`. The test still fails, however, the blocks that are shown right now are: "srcdoc new src" instead of "srcdoc old srcdoc" when test result is shown.

jdm:
Yes, those remaining failures are caused by the [issue I mentioned earlier](https://github.com/web-platform-tests/wpt/pull/17437#issuecomment-549563144). We do not need to try to resolve them in this PR.

@bors-servo try=wpt

cagandhi:
> Yes, those remaining failures are caused by the [issue I mentioned earlier](https://github.com/web-platform-tests/wpt/pull/17437#issuecomment-549563144). We do not need to try to resolve them in this PR.
> 
> @bors-servo try=wpt

@jdm Should we revert back to the previous commit where we didn't override srcdoc document origin so the tests can pass since they have been modified in the referenced issue?

jdm:
I'm not sure I understand - without the changes from 8205bbd562f27002f394f22cbbf12c909a2b6ac1, we get a different set of failures (`iframe.contentDocument is null`). With those changes, this test should start passing as soon as we sync our local copy of the test with https://github.com/web-platform-tests/wpt/commit/08f5e4f43dce6826d9a045eff8d85327c95ca2f3.

jdm:
In any case, you will need to run the following commands to update the expected test results for lots of tests:
```
wget https://build.servo.org/builders/linux-rel-wpt/builds/14066/steps/test__1/logs/test-wpt.log -O wpt1
wget https://build.servo.org/builders/linux-rel-css/builds/14003/steps/test/logs/test-wpt.log -O wpt2
./mach update-wpt wpt1 wpt2
git commit -a -m "update srcdoc-using test expectations"
```

jdm:
Also `./mach fmt` is required:
```

[94m./components/script/script_thread.rs[0m:[93m3182[0m: [91mLine is longer than 120 characters[0m
```

amj23897:
> In any case, you will need to run the following commands to update the expected test results for lots of tests:
> 
> ```
> wget https://build.servo.org/builders/linux-rel-wpt/builds/14066/steps/test__1/logs/test-wpt.log -O wpt1
> wget https://build.servo.org/builders/linux-rel-css/builds/14003/steps/test/logs/test-wpt.log -O wpt2
> ./mach update-wpt wpt1 wpt2
> git commit -a -m "update srcdoc-using test expectations"
> ```

The latest commit fixes formatting issues. `./mach check` and `./mach build --dev --verbose` commands were also successfully executed. Also, we executed the mentioned commands to fetch the test files. But tests are failing - they don't seem to have been updated.

jdm:
Ack, I gave the wrong links.
```
wget https://build.servo.org/builders/linux-rel-wpt/builds/14066/steps/test__1/logs/test-wpt.log/text -O wpt1
wget https://build.servo.org/builders/linux-rel-css/builds/14003/steps/test/logs/test-wpt.log/text -O wpt2
./mach update-wpt wpt1 wpt2
git commit -a -m "update srcdoc-using test expectations"
```

amj23897:
> Ack, I gave the wrong links.
> 
> ```
> wget https://build.servo.org/builders/linux-rel-wpt/builds/14066/steps/test__1/logs/test-wpt.log/text -O wpt1
> wget https://build.servo.org/builders/linux-rel-css/builds/14003/steps/test/logs/test-wpt.log/text -O wpt2
> ./mach update-wpt wpt1 wpt2
> git commit -a -m "update srcdoc-using test expectations"
> ```

I tried the new links. Some files, mainly *.html.ini, were updated a5503e8. But the test is not passing.

jdm:
Like I said - the test will continue to fail until we sync with upstream. That has not happened yet.

components/script/script_thread.rs
```diff
@@ -3176,11 +3178,13 @@ impl ScriptThread {
             self.timer_event_chan.clone(),
         );
 
-        let origin = if final_url.as_str() == "about:blank" {
-            incomplete.origin.clone()
-        } else {
-            MutableOrigin::new(final_url.origin())
-        };
+        // srcdoc document's origin must share creator document's origin - added OR condition
```
jdm: This comment doesn't add anything over reading the code, so let's remove it.

components/script/script_thread.rs
```diff
@@ -3176,11 +3178,13 @@ impl ScriptThread {
...
+        let origin =
+            if final_url.as_str() == "about:blank" || incomplete.url.as_str() == "about:srcdoc" {
```
jdm: There should be no difference in behaviour using `final_url` here, so let's do that instead.

components/script/script_thread.rs
```diff
@@ -3828,6 +3832,27 @@ impl ScriptThread {
         context.process_response_eof(Ok(ResourceFetchTiming::new(ResourceTimingType::None)));
     }
 
+    // refer definition of pre_page_load method in "script_thread.rs"; code taken from "start_page_load_about_blank"
```
jdm: Let's add this comment instead:
`/// Synchronously parse a srcdoc document from a giving HTML string.`

components/script/script_thread.rs
```diff
@@ -3828,6 +3832,27 @@ impl ScriptThread {
...
+    fn page_load_about_srcdoc(&self, incomplete: InProgressLoad, src_doc: String) {
+        // ADD NEW CODE
```
jdm: Let's remove this comment.

components/script/script_thread.rs
```diff
@@ -3828,6 +3832,27 @@ impl ScriptThread {
...
+
+        self.incomplete_loads.borrow_mut().push(incomplete);
+
+        let url = ServoUrl::parse("about:srcdoc").unwrap();
+        let mut context = ParserContext::new(id, url.clone());
+
+        let mut meta = Metadata::default(url);
+        meta.set_content_type(Some(&mime::TEXT_HTML));
+
+        // let chunk be src_doc
```
jdm: This comment doesn't add anything over the code, so let's remove it.

components/script_traits/lib.rs
```diff
@@ -163,6 +163,11 @@ pub struct LoadData {
     pub referrer: Option<Referrer>,
     /// The referrer policy.
     pub referrer_policy: Option<ReferrerPolicy>,
+
+    /// add a field to store srcdoc contents - init step 3; since srcdoc is a DOMString which is
+    /// inherently a Rust String - https://doc.servo.org/script/dom/bindings/str/struct.DOMString.html
+    /// https://doc.rust-lang.org/rust-by-example/std/str.html#targetText=Strings,in%20Rust%3A%20String%20and%20%26str%20.&targetText=%26str%20is%20a%20slice%20(%20%26%5B,view%20into%20Vec%20.
```
jdm: Let's replace this comment with `/// The source to use instead of a network response for a srcdoc document.`

components/script/dom/htmliframeelement.rs
```diff
@@ -228,6 +233,31 @@ impl HTMLIFrameElement {
...
         // TODO: srcdoc
+        // https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes
+        if self
+            .upcast::<Element>()
+            .has_attribute(&local_name!("srcdoc"))
+        {
+            let url = ServoUrl::parse("about:srcdoc").unwrap();
+            let document = document_from_node(self);
+            let window = window_from_node(self);
+            //added pipeline_id but still tests are failing
```
jdm: Let's remove this comment.

components/script/dom/htmliframeelement.rs
```diff
@@ -107,6 +107,11 @@ impl HTMLIFrameElement {
             .unwrap_or_else(|| ServoUrl::parse("about:blank").unwrap())
     }
 
+    fn get_srcdoc(&self) -> String {
```
jdm: No need for a method, since there's only one caller. We can inline this code in the caller instead.

components/script/dom/htmliframeelement.rs
```diff
@@ -228,6 +233,31 @@ impl HTMLIFrameElement {
     /// <https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes>
     fn process_the_iframe_attributes(&self, mode: ProcessingMode) {
         // TODO: srcdoc
```
jdm: This TODO can be removed now.

components/script/dom/htmliframeelement.rs
```diff
@@ -228,6 +233,31 @@ impl HTMLIFrameElement {
...
+        // https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes
```
jdm: This same comment exists two lines earlier, so let's remove it.

jdm (changes requested):
Looking good! Please squash all the commits together when making fixes to address these review comments.

jaymodi98:
> Looking good! Please squash all the commits together when making fixes to address these review comments.

I have made all the suggested changes and pushed them in one commit. Thanks.

jdm:
Thanks for addressing the comments! I have gone ahead and squashed the commits, addressed some merge conflicts in the test results, and verified that we now completely pass srcdoc_process_attributes.html \o/

@bors-servo r+

cagandhi:
@jdm Is there some way we can help in solving the synchronous navigation issue?

jdm:
@cagandhi The actual implementation work would be doable, but I would need to do some talking with more knowledgeable people first about what the expected model should be.

cagandhi:
@jdm On behalf of all my team members @jaymodi98 @amj23897, I would like to thank you for helping us immensely with this. This was our first open-source contribution and it feels really great now that the code is merged.

jdm:
Thanks for take ng the plunge with us!

PR #24584, opened 2019-10-30
implement PerformanceResourceTiming duration

### LINKED ISSUE #21255: PerformanceResourceTiming: duration
In components/net/http_loader.rs::http_fetch we need to set duration for the PerformanceEntry member of PerformanceResourceTiming. This will be the difference between response_end and start_time

Spec: https://w3c.github.io/resource-timing/#dfn-duration

JoshMcguigan:
Implements #21255

I wanted to give this a shot, but I'm not sure this is the appropriate implementation. In particular, I'm not sure if adding duration to the webidl interface is correct, because that doesn't seem to align with the web version (linked above the interface)? 

Also, I'd like to write a test for this method, but I'm not sure where that test should go.

My apologies if this is way off target.

---

- [x] These changes fix #21255 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/performanceresourcetiming.rs
+++ b/components/script/dom/performanceresourcetiming.rs
@@ -55,7 +55,6 @@ pub struct PerformanceResourceTiming {
     decoded_body_size: u64, //size in octets
 }
 
-// TODO(#21255): duration
 // TODO(#21269): next_hop
 // TODO(#21264): worker_start
 // TODO(#21258): fetch_start
@@ -115,7 +114,7 @@ impl PerformanceResourceTiming {
                 DOMString::from(url.into_string()),
                 DOMString::from("resource"),
                 resource_timing.start_time as f64,
-                0.,
+                resource_timing.response_end as f64 - resource_timing.start_time as f64,
             ),
             initiator_type: initiator_type,
             next_hop: next_hop,
--- a/tests/wpt/metadata/resource-timing/test_resource_timing.html.ini
+++ b/tests/wpt/metadata/resource-timing/test_resource_timing.html.ini
@@ -15,9 +15,6 @@
   [PerformanceEntry has correct protocol attribute (iframe)]
     expected: FAIL
 
-  [PerformanceEntry has correct name, initiatorType, startTime, and duration (img)]
-    expected: FAIL
-
   [PerformanceEntry has correct order of timing attributes (img)]
     expected: FAIL
 
--- a/tests/wpt/metadata/resource-timing/test_resource_timing.https.html.ini
+++ b/tests/wpt/metadata/resource-timing/test_resource_timing.https.html.ini
@@ -66,9 +66,6 @@
   [PerformanceEntry has correct protocol attribute (script)]
     expected: FAIL
 
-  [PerformanceEntry has correct name, initiatorType, startTime, and duration (img)]
-    expected: FAIL
-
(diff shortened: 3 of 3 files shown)


components/script/dom/webidls/PerformanceResourceTiming.webidl
```diff
@@ -23,6 +23,7 @@ interface PerformanceResourceTiming : PerformanceEntry {
     readonly attribute DOMHighResTimeStamp requestStart;
     readonly attribute DOMHighResTimeStamp responseStart;
     readonly attribute DOMHighResTimeStamp responseEnd;
+    readonly attribute DOMHighResTimeStamp duration;
```
jdm: Thanks for pointing out the difference with the specification - duration is actually a property of PerformanceEntry, which PerformanceResourceTiming inherits from, so the specification points out that we should be ensuring that that interface's Duration method returns the difference between ResponseEnd and StartTime.

components/script/dom/performanceresourcetiming.rs
```diff
@@ -250,4 +249,9 @@ impl PerformanceResourceTimingMethods for PerformanceResourceTiming {
     fn ResponseEnd(&self) -> DOMHighResTimeStamp {
         Finite::wrap(self.response_end)
     }
+
+    // https://w3c.github.io/resource-timing/#dfn-duration
```
jdm: We will need to revert this change and ensure that we pass appropriate values to PerformanceEntry::new_inherited instead, so that PerformanceEntry::Duration and PerformanceEntry::StartTime return the expected values..

jdm (changes requested):

JoshMcguigan:
Thanks for the feedback @jdm. It seems the `PerformanceResourceTiming` struct is immutable, and `responseEnd` is set to zero in the `PerformanceResourceTiming::new_inherited` method, so this change is only needed in the `from_resource_timing` method?

jdm:
Inheritance in Servo is implemented by forcing the first field of a DOM struct Child that "inherits" from type Parent to be an instance of Parent. This allows us to cast a `*const Child` to a `*const Parent` (and similarly for &Child -> &Parent) and interact it the values as usual (and this is hidden behind a bunch of [traits](https://doc.servo.org/script/dom/bindings/inheritance/trait.Castable.html) that make this safe).

But yes, I agree that we should only need to make this change in `from_resource_timing`. Then you can run the tests in `./mach test-wpt tests/wpt/web-platform-tests/resource-timing` and see if any test results change.

jdm:
```
{
    "status": "PASS", 
    "group": "default", 
    "message": null, 
    "stack": null, 
    "subtest": "PerformanceEntry has correct name, initiatorType, startTime, and duration (img)", 
    "test": "/resource-timing/test_resource_timing.html", 
    "line": 212920, 
    "action": "test_result", 
    "expected": "FAIL"
}
{
    "status": "PASS", 
    "group": "default", 
    "message": null, 
    "stack": null, 
    "subtest": "PerformanceEntry has correct name, initiatorType, startTime, and duration (img)", 
    "test": "/resource-timing/test_resource_timing.https.html", 
    "line": 327444, 
    "action": "test_result", 
    "expected": "FAIL"
}
```
You will need to [update the expected test results](https://github.com/servo/servo/blob/master/tests/wpt/README.md#updating-test-expectations). to reflect that they're now passing \o/

JoshMcguigan:
Thanks again for the follow up. I've updated the test expectations.

PR #24590, opened 2019-10-30
Fix timing allow check

### LINKED ISSUE #24367: Timing allow check should only run for cross-origin requests
Right now we perform the check for every response we get and don't actually compare the current request's origin (or any of the requests in its redirect chain) against the origin of the original source of the request. This means that same-origin requests are treated as failing the timing allow check, which is incorrect.

shnmorimoto:
Fix #24367

---

- [x] These changes fix #24367 (GitHub issue number if applicable)

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -1462,7 +1462,15 @@ fn http_network_fetch(
             _ => false,
         });
 
-    if !req_origin_in_timing_allow && !wildcard_present {
+    let is_same_origin = request.url_list.iter().any(|url| match request.origin {
+        SpecificOrigin(ref immutable_request_origin) => {
+            url.clone().into_url().origin().ascii_serialization() ==
+                immutable_request_origin.ascii_serialization()
+        },
+        _ => false,
+    });
+
+    if !(is_same_origin || req_origin_in_timing_allow || wildcard_present) {
         context.timing.lock().unwrap().mark_timing_check_failed();
     }
 
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -1462,11 +1462,8 @@ fn http_network_fetch(
             _ => false,
         });
 
-    let is_same_origin = request.url_list.iter().any(|url| match request.origin {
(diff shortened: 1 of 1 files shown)


components/net/http_loader.rs
```diff
@@ -1462,7 +1462,15 @@ fn http_network_fetch(
             _ => false,
         });
 
-    if !req_origin_in_timing_allow && !wildcard_present {
+    let is_same_origin = request.url_list.iter().any(|url| match request.origin {
```
jdm: We should either call this `is_cross_origin` and check if any of the urls in the url list are not same-origin, or we should call it `is_same_origin` and check if _all_ of the urls in the url list are same-origin.

components/net/http_loader.rs
```diff
@@ -1462,7 +1462,15 @@ fn http_network_fetch(
...
+        SpecificOrigin(ref immutable_request_origin) => {
+            url.clone().into_url().origin().ascii_serialization() ==
+                immutable_request_origin.ascii_serialization()
```
jdm: Let's do `url.origin() == immutable_request_origin` or `url.origin() != immutable_request_origin` instead.

jdm (changes requested):
Looks good! I'll do a test run once the changes to perform the same-origin check for all URLs are present.

components/net/http_loader.rs
shnmorimoto: Thanks for the review.

I've fixed it as following.

> check if all of the urls in the url list are same-origin.

and

```rust
let is_same_origin = request.url_list.iter().all(|url| match request.origin {
```

components/net/http_loader.rs
shnmorimoto: Thanks.
I've fixed it.

```rust
url.origin() == *immutable_request_origin
```

jdm:
@bors-servo r+
Thanks!

jdm:
@bors-servo retry
* https://github.com/rust-lang/rust/issues/65536

Issue #24622, opened 2019-11-01
Intermittent mac timeout due to missing font URLs (No URL for Core Text font!)


jdm:
This affects lots of different tests, so I'm not adding any tests to the summary.
```
 ▶ TIMEOUT [expected FAIL] /css/css-fonts/font-variant-numeric-08.html
  │ 
  │ 
  │ 
  │ 
  │ No URL for Core Text font! (thread FontCacheThread, at src/libcore/option.rs:1185)
  │ stack backtrace:
  │    0: backtrace::backtrace::trace
  │    1: backtrace::capture::Backtrace::new
  │    2: servo::main::{{closure}}
  │    3: std::panicking::rust_panic_with_hook
  │    4: std::panicking::continue_panic_fmt
  │    5: rust_begin_unwind
  │    6: core::panicking::panic_fmt
  │    7: core::option::expect_failed
  │    8: gfx::platform::macos::font_template::FontTemplateData::bytes
  │    9: gfx::font_cache_thread::FontCache::run
  │   10: std::sys_common::backtrace::__rust_begin_short_backtrace
  │   11: std::panicking::try::do_call
  │   12: __rust_maybe_catch_panic
  │   13: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   14: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │   15: std::sys::unix::thread::Thread::new::thread_start
  │   16: _pthread_body
  │   17: _pthread_start
```

PR #24624, opened 2019-11-03
Fix keypress trigger condition

### LINKED ISSUE #22346: Enter and backspace don't trigger onkeypress events for text inputs
`./mach run -d 'data:text/html,<input value="hi" onkeypress="console.log(event)">'`

garasubo:
Fix #22346

keypress event should be triggered for keys representing character
values. So, we should trigger this event for enter key.
This event should not trigger for IME inputs.

TODO:
- It seems we don't handle composition events correctly. To implement this keypress condition correctly, we should fix that first. In my current implementation, onkeypress event will be trigger when the user press Enter key to send inputs in IME (onCompotionEnd).
- I don't update any tests, and I couldn't find any tests related to this change in WPT. It might be better to add some tests for it, but I don't know what is the appropriate way.

---

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -1435,7 +1435,8 @@ impl Document {
 
         // https://w3c.github.io/uievents/#keys-cancelable-keys
         if keyboard_event.state == KeyState::Down &&
-            keyboard_event.key.legacy_charcode() != 0 &&
+            is_character_value_key(&(keyboard_event.key)) &&
+            !keyboard_event.is_composing &&
             cancel_state != EventDefault::Prevented
         {
             // https://w3c.github.io/uievents/#keypress-event-order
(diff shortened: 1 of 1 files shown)


garasubo:
@SimonSapin Could you review this PR? Also I'm not sure I should update the tests or not. What do you think?

paulrouget:
> It seems we don't handle composition events correctly. To implement this keypress condition correctly, we should fix that first. In my current implementation, onkeypress event will be trigger when the user press Enter key to send inputs in IME (onCompotionEnd).

Feel free to file a followup issue for that.

> I don't update any tests, and I couldn't find any tests related to this change in WPT. It might be better to add some tests for it, but I don't know what is the appropriate way.

Let's see how happy the wpt tests are.

@bors-servo try=wpt

garasubo:
@paulrouget Thanks for reviewing. For IME issue, I opened #24724 . I'll update more after some investigation.

paulrouget:
@garasubo is this ready to land? If so, remove the "WIP" part of the PR title.

garasubo:
@paulrouget Yes. It is ready if I don't have to update the tests.

CYBAI:
@bors-servo retry

- #24622

PR #24626, opened 2019-11-03
Issue 21810/improve validation methods

### LINKED ISSUE #21810: Remaining issues with input value sanitization
- [x] We are currently missing the value sanitization algorithm for e-mail type inputs, and the corresponding spec can be found here: https://html.spec.whatwg.org/multipage/input.html#e-mail-state-(type=email), and the implementation in the code can be found here: https://github.com/servo/servo/blob/6e844f22780a1a2b78c3a1f4bd934581144cb34e/components/script/dom/htmlinputelement.rs#L1126-L1150 (#23472)

- [ ] The following validation methods: https://github.com/servo/servo/blob/df2adebefdfa3da49f173e480fa1e56450f9bda2/components/script/dom/bindings/str.rs#L273-L301 need to be rewritten, as I have communicated with WHATWG before and found out that successfully parsing the string does not mean a successful validation. We'll need to go through each of the input types in the spec and look at their validation algorithm to see how it should look like instead:
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-date-string
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-month-string
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-week-string
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-floating-point-number

- [ ] Remove the catch-all case here https://github.com/servo/servo/blob/6e844f22780a1a2b78c3a1f4bd934581144cb34e/components/script/dom/htmlinputelement.rs#L1188
so that we don't forget about implementing value sanitizations on any of the input types again.

glowe:
This is a start at addressing #21810. I'm putting these changes out early to get some feedback on the following items:

1. I added unit tests for the validation methods mentioned in #21810, because I couldn't tell whether any of the existing WPT tests covered them. Are these tests worthwhile? Are any of them unnecessary?
2. I changed the implementation for `is_valid_floating_point_number_string` so that it passed the tests. The previous version of the function wasn't restrictive enough (it allowed certain whitespace characters before the number string).
3. I changed the catch-all condition in `htmlinputelement.rs` to account for the remaining input types that don't have a value sanitization algorithm. This last change seems good to me since we won't be able to add a new input type without adding it to the case and checking the spec for an algorithm.
---

- [x] These changes fix #21810


jdm (approved):
I'm in favour of unit tests for small, self-contained routines like this.

components/script/dom/bindings/str.rs
```diff
@@ -322,10 +322,79 @@ impl DOMString {
     }
 
     /// A valid month string should be "YYYY-MM"
-    /// YYYY must be four or more digits, MM both must be two digits
+    /// YYYY must be four or more digits, where YYYY > 0
+    /// MM both must be two digits, where 1 <= MM <= 12
     /// https://html.spec.whatwg.org/multipage/#valid-month-string
     pub fn is_valid_month_string(&self) -> bool {
-        parse_month_string(&self.0).is_ok()
```
jdm: Whoops, I missed reading this commit. While in general I'm in favour of using state machines for clarity of purpose and avoiding invalid states, I find this one to be difficult to follow. I'm not sure if it's the names of the states, or the repetition between them, or something else, but my eyes keep glazing over when I try to follow the logic. Given how straightforward the specification text is, I'm not exactly sure what to recommend.

jdm: My reading of the specification is that the parsing algorithm for month strings is the same as the validation algorithm for month strings, so I don't see a good reason to reimplement this particular algorithm. Presumably there are other input types where they diverge?

glowe: > My reading of the specification is that the parsing algorithm for month strings is the same as the validation algorithm for month strings, so I don't see a good reason to reimplement this particular algorithm. Presumably there are other input types where they diverge?

Actually, the tests passed for all of the original implementations, so I judge them to be equivalent to the specifications. The original issue #21810 recommended reimplementing all of these functions, but I think that recommendation was based on the premise that the functions *might* be incorrect. 

To move things forward, how about dropping 8a96009 from this PR?

jdm (approved):

jdm:
@bors-servo r+
Thanks!

jdm:
Hmm, looks like we now fail a test that we used to pass:
```
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected \"\" but got \"2e308\"", 
    "stack": "@http://web-platform.test:8000/html/semantics/forms/the-input-element/number.html:50:7\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1931:25\ntest@http://web-platform.test:8000/resources/testharness.js:544:30\n@http://web-platform.test:8000/html/semantics/forms/the-input-element/number.html:43:5\n", 
    "subtest": "value >= Number.MAX_VALUE", 
    "test": "/html/semantics/forms/the-input-element/number.html", 
    "line": 60091, 
    "action": "test_result", 
    "expected": "PASS"
}
```
This can be reproduced with `./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms/the-input-element/number.html`.

glowe:
> This can be reproduced with ./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms/the-input-element/number.html.

Thanks for identifying the failing test for me.

We are now failing on a bounds check: `2e308` exceeds `Number.MAX_VALUE` (`1.7976931348623157e+308`). I could be reading this wrong, but the specification for a valid floating point number (https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-floating-point-number) doesn't mention a bounds. However the specification for parsing (https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-floating-point-number-values) does mention it in that the converted value must be a `IEEE 754 double-precision floating-point value`.

_Interestingly, I tried this specific test on Firefox 70.0.1 and it failed (http://wpt.live/html/semantics/forms/the-input-element/number.html). However, this test passes in the latest Chrome and Safari browsers._

Here's what I propose we do next (let me know what you think):

1.) I'll update the code to handle the bounds check, probably by delegating to the parsing code after verifying that there's no leading whitespace.
2.) I'm now wondering about the value of the unit tests I added. I didn't notice these web platform tests before and it seems like the unit tests are duplicative and potentially misleading. I'm leaning towards dropping the unit tests for this PR.
3.) The web platform tests for number inputs are incomplete for the leading whitespace check:

https://github.com/servo/servo/blob/b62d1d4f7a70292b55547f6ea9f59e19525522ec/tests/wpt/web-platform-tests/html/semantics/forms/the-input-element/number.html#L38

The specification mentions that the whitespace characters can be `TAB`, `LF`, `FF`, `CR`, or `SPACE` (https://infra.spec.whatwg.org/#ascii-whitespace). Our previous implementation only handled `SPACE`s. Would it be worth me filing an issue (and potentially a PR) to add tests for the other characters?

jdm:
Those choices seem reasonable! You can also include the WPT changes in this PR, because we automatically sync them upstream when they merge in our vendored copy.

jdm (approved):

jdm:
And FYI https://wpt.fyi/results/html/semantics/forms/the-input-element/number.html?label=master&label=experimental&aligned&q=number.html backs up your results with respect to Firefox's test results.

glowe:
> Those choices seem reasonable! You can also include the WPT changes in this PR, because we automatically sync them upstream when they merge in our vendored copy.

Thanks! I did not know that. I've enhanced the number input WPT to include those additional cases and deleted the unit test since it's redundant coverage.

jdm:
@bors-servo r+
Thanks!

Issue #24631, opened 2019-11-04
loading https://foo@bar.com rewrites the URL to https://http//foo@bar.com


paulrouget:

paulrouget:
Fixed by #24634

PR #24634, opened 2019-11-04
Do not assume UWP and Servo agree on URL validity

### LINKED ISSUE #24589: Failing to parse a URL in the UWP app causes a panic
Try to load an invalid URL in the UWP app and watch it panic. We should pop up an alert dialog or something, but it should really be ignored rather than bring down the browser.

paulrouget:
This should address #24589.

Note: to check that a URL is valid, we try to parse the url and if an exception is raised, we assume the url is not valid. Even though we catch the exception, Visual Studio might "force" the crash + break here: https://github.com/servo/servo/compare/master...paulrouget:url?expand=1#diff-4e059561062fe024a35522f60f7f14c1R116 - I'm not sure why. In VS, in the Exception Setting, we need to uncheck the Break When Thrown option to avoid that crash. It only happens for few urls and I'm unclear why.

---

- [x] These changes fix #24589 (GitHub issue number if applicable)

### DIFF
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -152,6 +152,12 @@ pub fn servo_version() -> String {
     servo::config::servo_version()
 }
 
+/// Test if a url is valid.
+pub fn is_uri_valid(url: &str) -> bool {
+    info!("load_uri: {}", url);
+    ServoUrl::parse(url).is_ok()
+}
+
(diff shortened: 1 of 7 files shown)


paulrouget:
Not ready for review. I will change how this works. We should expose some ServoURL methods to the API then we don't rely on 2 different parsers.

paulrouget:
Also fix #24631

paulrouget:
Ready for review.

@jdm r?

Issue #24454, opened 2019-10-16
WebMessaging: fully support MessageEventSource


gterzian:
This requires extending `MessageEvent` and `ExtendableMessageEvent` with `MessageEventSource`, and setting it to the right object when the event is dispatched.

`MessageEvent` only supports `WindowProxy` as source.

Spec: https://html.spec.whatwg.org/multipage/comms.html#messageeventsource

Webidls to be updated:

https://github.com/servo/servo/blob/b96b08a3d9dd7dc042de869b0a109435af776a89/components/script/dom/webidls/MessageEvent.webidl#L14

Code where event is dispatched and where the `source` should be added/updated:

MessagePort(PR in progress): https://github.com/servo/servo/pull/23637/files/a1a3c1e60ad87bb90658a3a111be52b2cda68b04#diff-59d233642d0ce6d687484bdd009e1017R444

Window, currently using an `Option<WindowProxy>` as source, instead of `MessageEventSource` : https://github.com/servo/servo/blob/b96b08a3d9dd7dc042de869b0a109435af776a89/components/script/dom/window.rs#L2373

gterzian:
When the webidl are updated, running `./mach check` will generate the `MessageEventSource` binding, which can then be imported in a way similar to https://github.com/servo/servo/blob/b96b08a3d9dd7dc042de869b0a109435af776a89/components/script/dom/messageevent.rs#L7

gterzian:
Moved the service-worker part to a separate issue, because it involves a separate `Client` interface, to https://github.com/servo/servo/issues/24659

gterzian:
Closing this, because actually `MessagePort` as a source is only used when dispatching the `connnect` event of shared-worker(step 27 of https://html.spec.whatwg.org/multipage/workers.html#worker-processing-model). See https://github.com/servo/servo/pull/24636#discussion_r342390301

I don't think there is a need for a separate issue from https://github.com/servo/servo/issues/7458

gterzian:
Setting a service worker as "source" would have to be done as part of https://github.com/servo/servo/issues/24660

PR #24636, opened 2019-11-04
Use MessageEventSource on MessageEvent IDL

### LINKED ISSUE #22617: WindowProxy interface is considered only exposed to Window environment
WindowProxy is a weird fake interface that's listed in Bindings.conf but does not have a WebIDL interface. This makes the WebIDL parser report an error if the WindowProxy interface is used in interfaces that are exposed in non-Window globals, like MessageSourceEvent.webidl.

saschanaz:
---

- [x] These changes fix #22617 

### DIFF
--- a/components/script/dom/bindings/conversions.rs
+++ b/components/script/dom/bindings/conversions.rs
@@ -45,6 +45,7 @@ use crate::dom::htmlcollection::HTMLCollection;
 use crate::dom::htmlformcontrolscollection::HTMLFormControlsCollection;
 use crate::dom::htmloptionscollection::HTMLOptionsCollection;
 use crate::dom::nodelist::NodeList;
+use crate::dom::windowproxy::WindowProxy;
 use js::conversions::latin1_to_string;
 pub use js::conversions::ConversionBehavior;
 pub use js::conversions::{ConversionResult, FromJSValConvertible, ToJSValConvertible};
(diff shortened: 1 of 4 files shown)


components/script/dom/bindings/conversions.rs
```diff
@@ -634,3 +635,24 @@ where
         Err(()) => Err(Error::JSFailed),
     }
 }
+
+/// Get a `DomRoot<T>` for a WindowProxy accessible from a `HandleValue`.
+/// Caller is responsible for throwing a JS exception if needed in case of error.
+pub fn windowproxy_from_handlevalue(
```
nox: This should be unsafe.

nox (changes requested):

components/script/dom/messageevent.rs
```diff
@@ -129,7 +151,11 @@ impl MessageEvent {
             false,
             message,
             DOMString::from(origin.unwrap_or("")),
-            source,
+            source
+                .map(|source| {
+                    WindowProxyOrMessagePortOrServiceWorker::WindowProxy(DomRoot::from_ref(source))
```
gterzian: I think you might want to instead update the `source` argument to be a `WindowProxyOrMessagePortOrServiceWorker`, and then set it at the call site, at https://github.com/servo/servo/blob/c89676be1d4b0e727b793f39cfb2d5db006ff1a8/components/script/dom/globalscope.rs#L534 and https://github.com/servo/servo/blob/c89676be1d4b0e727b793f39cfb2d5db006ff1a8/components/script/dom/dedicatedworkerglobalscope.rs#L479

Then you could also close https://github.com/servo/servo/issues/24454

gterzian (changes requested):
I think this could quite easily also close https://github.com/servo/servo/issues/24454

components/script/dom/messageevent.rs
saschanaz: [The spec](https://html.spec.whatwg.org/multipage/web-messaging.html#message-port-post-message-steps) does not set the `source` attribute, should the spec be updated?

jdm: That's a good point. I double-checked Gecko's implementation and it also appears to have a null source for message events triggered by MessagePort.postMessage, which I did not expect.

jdm: Maybe it's because for two MessagePorts in separate event loops there is no reasonable object to provide for the source argument.

jdm: https://html.spec.whatwg.org/multipage/comms.html#dom-messageevent-source suggests that the MessagePort is only used for shared worker connect events, which makes more sense to me.

gterzian: You're right, it seems to only be used for shared-workers in step 27 of https://html.spec.whatwg.org/multipage/workers.html#worker-processing-model

CYBAI:
@bors-servo retry

- cygwin oom

PR #24640, opened 2019-11-04
Remove all_files options for clang-format check

### LINKED ISSUE #24638: The "all files" option for test-tidy + clang-format doesn't make sense
It was supposed to be an optimisation. That was a bit naive. If `all_files` is false, it checks only files that have not been commited. Which doesn't make sense, we usually commit our changes before running `mach test-tidy`. We should just remove that option as we don't test many files any way.

paulrouget:
fix #24638


jdm:
@bors-servo retry
* #23290

Issue #24611, opened 2019-10-31
Builds intermittently fail with a permission error during harfbuzz builds


jdm:
I've seen this over the past couple days on nightly windows builds, and just saw this in #24603 during the PR pre-merge job.

```
Caused by:
  process didn't exit successfully: `/repo/target/debug/build/harfbuzz-sys-c7b89349823ef263/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-env-changed=HARFBUZZ_SYS_NO_PKG_CONFIG
[...]

Issue #13480, opened 2016-09-28
intermittent crash ("Connection reset by peer" from create_api) in /html/semantics/embedded-content/the-canvas-element/2d.getcontext.shared.html, /html/semantics/scripting-1/the-template-element/template-element/node-document-changes.html, /dom/traversal/TreeWalker-walking-outside-a-tree.html, /XMLHttpRequest/responseText-status.html, /_mozilla/css/line_break_nowrap.html


metajack:
```
 ▶ CRASH [expected FAIL] /css21_dev/html4/vertical-align-007.htm
  │ 
  │ VMware, Inc.
  │ Gallium 0.4 on softpipe
  │ 3.3 (Core Profile) Mesa 12.0.1
  │ called `Result::unwrap()` on an `Err` value: IoError(Error { repr: Os { code: 104, message: "Connection reset by peer" } }) (thread LayoutThread PipelineId { namespace_id: PipelineNamespaceId(0), index: PipelineIndex(0) }, at ../src/libcore/result.rs:799)
  │ stack backtrace:
  │    0:     0x7fa759924c8d - backtrace::backtrace::trace::h0e60ef08c7c34e9f
  │    1:     0x7fa759925312 - backtrace::capture::Backtrace::new::h8bf319c36d8f5d1b
  │    2:     0x7fa75846e9cd - servo::main::_{{closure}}::h911afab4cc009f33
  │    3:     0x7fa759f03745 - std::panicking::rust_panic_with_hook::hb1322e5f2588b4db
  │    4:     0x7fa759f03612 - std::panicking::begin_panic::hfbeda5aad583dc32
  │    5:     0x7fa759f03550 - std::panicking::begin_panic_fmt::h4fe9fb9d5109c4bf
  │    6:     0x7fa759f034d1 - rust_begin_unwind
  │    7:     0x7fa759f3ff5c - core::panicking::panic_fmt::h4395919ece15c671
  │    8:     0x7fa759a8964f - core::result::unwrap_failed::h3191a47cbfbe2a26
  │    9:     0x7fa759a8ad20 - webrender_traits::api::_<impl webrender_traits..RenderApiSender>::create_api::h74e2147a94b73b3e
  │   10:     0x7fa7587e10f7 - _<layout_thread..LayoutThread as layout_traits..LayoutThreadFactory>::create::_{{closure}}::hcbdf224d94fa1d3a
  │   11:     0x7fa75875c796 - std::panicking::try::do_call::h032f92c4e6f5b87e
[...]

PR #24653, opened 2019-11-04
Add consume stream callback

### LINKED ISSUE #21476: Support asynchronous WebAssembly compilation

ridhimrastogi:
Added the consume stream callback function as per the steps mentioned [here](https://webassembly.github.io/spec/web-api/index.html#compile-a-potential-webassembly-response)

---

- [x] These changes fix #21476


components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+    obj: HandleObject,
+    _mimeType: MimeType,
+    _consumer: *mut JSStreamConsumer,
+) {
+    let cx = JSContext::from_ptr(_cx);
+    let global = GlobalScope::from_context(*cx);
+    let returnValue = Promise::new(&global);
+    let mut handleobj = UndefinedValue();
+    GetProxyPrivate(*obj, &mut handleobj);
+    rooted!(in(*cx) let target = handleobj.to_object());
```
jdm: It's not clear to me why these lines are necessary. Why can't we use `obj` as the argument to root_from_handleobject?

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
 }
 
 #[allow(unsafe_code)]
-unsafe extern "C" fn report_stream_error_callback(_cx: *mut JSContext, error_code: usize) {
+unsafe extern "C" fn ConsumeStreamCallback(
```
jdm: Let's call this `consume_stream`, and add a comment like `/// <https://webassembly.github.io/spec/web-api/#compile-a-potential-webassembly-response` to be clear that we're implementing a particular specification.

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+        }
+    } else {
+        return returnValue.reject_error(Error::Type(
+            "expected Response or Promise resolving to Response".to_string(),
+        ));
+    }
+}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn ReportStreamErrorCallback(_cx: *mut JSContext, error_code: usize) {
```
jdm: Let's keep this as `report_stream_error`.

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
 }
 
 #[allow(unsafe_code)]
-unsafe extern "C" fn report_stream_error_callback(_cx: *mut JSContext, error_code: usize) {
+unsafe extern "C" fn ConsumeStreamCallback(
+    _cx: *mut RawJSContext,
+    obj: HandleObject,
+    _mimeType: MimeType,
+    _consumer: *mut JSStreamConsumer,
+) {
```
jdm: This needs to return a boolean.

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+    let cx = JSContext::from_ptr(_cx);
+    let global = GlobalScope::from_context(*cx);
+    let returnValue = Promise::new(&global);
+    let mut handleobj = UndefinedValue();
+    GetProxyPrivate(*obj, &mut handleobj);
+    rooted!(in(*cx) let target = handleobj.to_object());
+    if let Ok(unwrappedSource) = root_from_handleobject::<Response>(target.handle(), *cx) {
+        let mimetype = unwrappedSource.Headers().extract_mime_type();
+        if std::str::from_utf8(&mimetype).unwrap() != "application/wasm" {
+            return returnValue.reject_error(Error::Type(
```
jdm: Looking more closely at Firefox's implementation, it appears that we do not actually want to use Promise values in this code. Instead, we return true if there was no error, or throw an exception and return false if there was an error (contrary to what the specification says). We can throw exceptions using the [throw_dom_exception API](https://doc.servo.org/script/dom/bindings/error/fn.throw_dom_exception.html).

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+) {
+    let cx = JSContext::from_ptr(_cx);
+    let global = GlobalScope::from_context(*cx);
+    let returnValue = Promise::new(&global);
+    let mut handleobj = UndefinedValue();
+    GetProxyPrivate(*obj, &mut handleobj);
+    rooted!(in(*cx) let target = handleobj.to_object());
+    if let Ok(unwrappedSource) = root_from_handleobject::<Response>(target.handle(), *cx) {
+        let mimetype = unwrappedSource.Headers().extract_mime_type();
+        if std::str::from_utf8(&mimetype).unwrap() != "application/wasm" {
```
jdm: Instead of using from_utf8, we can do: `if mimetype == b"application/wasm" {`

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+            return returnValue.reject_error(Error::Type(
+                "Response has unsupported MIME type".to_string(),
+            ));
+        }
+        if unwrappedSource.Type().as_str() != "Basic" ||
+            unwrappedSource.Type().as_str() != "Cors" ||
+            unwrappedSource.Type().as_str() != "Default"
```
jdm: Let's do:
```rust
match unwrapped_source.Type() {
    DOMResponseType::Basic | DOMResponseType::Cors | DOMResponseType::Default => {},
    _ => {
        // error handling
    }
}
```

components/script/script_runtime.rs
```diff
@@ -826,7 +832,46 @@ impl StreamConsumer {
...
+        {
+            return returnValue.reject_error(Error::Type(
+                "Response.type must be 'basic', 'cors' or 'default'".to_string(),
+            ));
+        }
+        if !unwrappedSource.Ok() {
```
jdm: Let's add some comments like `// step 2.5` to make it more clear which parts of the spec are being implemented by the code.

jdm (changes requested):
Good start! We should add the call to InitStreamConsumerCallback as part of this pull request.

components/script/script_runtime.rs
ridhimrastogi: The consume_stream function takes a JS::HandleObject as an argument while the root_from_handle_obj expects a mozjs::rust::HandleObject.

jdm: Ah, you can use `mozjs::rust::HandleObject::from_raw(obj)` in that case.

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
ridhimrastogi: Done

components/script/script_runtime.rs
```diff
@@ -831,47 +832,67 @@ impl StreamConsumer {
...
+        //Step 2.2 Let mimeType be the result of extracting a MIME type from response’s header list.
+        let mimetype = unwrapped_source.Headers().extract_mime_type();
+
+        //Step 2.3 If mimeType is not `application/wasm`, return with a TypeError and abort these substeps.
+        if mimetype == b"application/wasm" {
+            throw_dom_exception(
+                cx,
+                &global,
+                Error::Type("Response has unsupported MIME type".to_string()),
+            );
```
jdm: In each of the cases where we throw an exception, we need to return false instead of true.

components/script/script_runtime.rs
```diff
@@ -831,47 +832,67 @@ impl StreamConsumer {
     }
 }
 
+/// Implements the steps to compile webassembly response mentioned here
+/// <https://webassembly.github.io/spec/web-api/#compile-a-potential-webassembly-response
```
jdm: nit: add a trailing > so it shows up as a link in the generated documentation.

jdm (changes requested):
Almost there!

components/script/script_runtime.rs
ridhimrastogi: done

components/script/script_runtime.rs
ridhimrastogi: Done

jdm:
If you add a call to InitStreamConsumerCallback that makes use of consume_stream and report_stream_error, then we can run the automated tests and see if any of these changes are observable!

jdm:
I did a little poking around in this branch on my own machine and discovered that if you add the following:
```
    unsafe extern "C" fn dispatch_to_event_loop(_closure: *mut c_void, dispatchable: *mut Dispatchable) -> bool {
        false
    }
    InitDispatchToEventLoop(cx, Some(dispatch_to_event_loop), ptr::null_mut());
```
Then you will start getting meaningful test results by running `./mach test-wpt tests/wpt/web-platform-tests/wasm/webapi`. In particular, a bunch of tests in contenttype.any.js will start timing out because there is an `==` instead of an `!=` in the new code :)

ridhimrastogi:
@jdm Can you give a little more detail on where to invoke the InitDispatchToEventLoop  method. On adding it to the new_rt_and_cx_with_parent() I see no change in behaviour in the test results from  ./mach test-wpt tests/wpt/web-platform-tests/wasm/webapi

jdm:
This is the only change I made to your code:
```diff
diff --git a/components/script/script_runtime.rs b/components/script/script_runtime.rs
index 364bd1d2cc..74f9459899 100644
--- a/components/script/script_runtime.rs
+++ b/components/script/script_runtime.rs
@@ -37,8 +37,8 @@ use js::glue::{CollectServoSizes, CreateJobQueue, DeleteJobQueue, JobQueueTraps,
 use js::glue::{RUST_js_GetErrorMessage, StreamConsumerConsumeChunk, StreamConsumerStreamEnd};
 use js::glue::{StreamConsumerNoteResponseURLs, StreamConsumerStreamError};
 use js::jsapi::ContextOptionsRef;
-use js::jsapi::InitConsumeStreamCallback;
-use js::jsapi::MimeType;
+use js::jsapi::{InitConsumeStreamCallback, InitDispatchToEventLoop};
+use js::jsapi::{Dispatchable, MimeType};
 use js::jsapi::StreamConsumer as JSStreamConsumer;
 use js::jsapi::{BuildIdCharVector, DisableIncrementalGC, GCDescription, GCProgress};
 use js::jsapi::{HandleObject, Heap, JobQueue};
@@ -434,6 +434,10 @@ unsafe fn new_rt_and_cx_with_parent(parent: Option<ParentRuntime>) -> Runtime {
     DisableIncrementalGC(cx);

     InitConsumeStreamCallback(cx, Some(consume_stream), Some(report_stream_error));
+    unsafe extern "C" fn dispatch_to_event_loop(_closure: *mut c_void, dispatchable: *mut Dispatchable) -> bool {
+        false
+    }
+    InitDispatchToEventLoop(cx, Some(dispatch_to_event_loop), ptr::null_mut());

     let microtask_queue = Rc::new(MicrotaskQueue::default());
     let job_queue = CreateJobQueue(
```
Now when I run `./mach test-wpt tests/wpt/web-platform-tests/wasm/webapi/` I see this output:
```
Ran 44 tests finished in 27.0 seconds.
  • 30 ran as expected. 0 tests skipped.
  • 2 tests timed out unexpectedly
  • 1 tests unexpectedly okay
  • 14 tests had unexpected subtest results
```

ridhimrastogi:
@jdm My bad, I was using `./mach check` to  find build errors all this while .so the tests were  running against the old binary. On rebuilding servo and changing the `==` to `!=` in mimetype comparision, more tests have started failing, specifically the ones checking for status. Any ideas on how to fix this and what is the expected behaviour. Are all the test cases supposed to pass with the code we have added or more code needs to be added.

jdm:
My expectation is that wit the code changes a small subset of tests that check for error cases should pass (which will be reported as an unexpected OK or PASS result, so it may look like a failure), and lots of tests that expected successful streaming to occur will fail. I recommend that you submit the latest code along with the [updated test expectations](https://github.com/servo/servo/blob/master/tests/wpt/README.md#updating-test-expectations) and we should be able to tell if any test results look out of place.

jdm:
Please update the test results for all of the wasm/ directory!

jdm:
To be clear, I'm looking for the following:
```
./mach test-wpt tests/wpt/web-platform-tests/wasm --log-raw wptlog
./mach update-wpt wptlog
```
Please squash all the commits into one, as well!

jdm (approved):
Looking good! Just need to squash the commits now before merging.

jdm:
@bors-servo retry
* https://github.com/rust-lang/rust/issues/65536

CYBAI:
@bors-servo retry

- #24611

CYBAI:
@bors-servo retry

- #24611

CYBAI:
@bors-servo retry

- #13480

PR #24655, opened 2019-11-04
Patch surfman and webxr

Manishearth:
Gets us a working hololens build without readback.

We should make sure these commits get preserved by merging them in eventually (so that bisection isn't broken)

r? @jdm


jdm:
@bors-servo retry
* #23290 #24656

PR #24661, opened 2019-11-05
Only updated history once

### LINKED ISSUE #24633: UWP: OnUrlChanged is called twice
Not sure if a UWP or servo issue

paulrouget:
We get the `HistoryChanged` event twice in many situations. This should fix it.

@cbrewster do these change make sense?

---

- [x] These changes fix #24633 (GitHub issue number if applicable)

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -3995,7 +3995,6 @@ where
                     new_context_info.is_visible,
                 );
                 self.update_activity(change.new_pipeline_id);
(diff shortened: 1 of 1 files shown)


cbrewster:
@paulrouget this makes sense to me since we always notify session history changed at the end of that function regardless of the match branch.

https://github.com/servo/servo/blob/28dbbda5bb0736b4f8f6c5c08a8bc268496f5422/components/constellation/constellation.rs#L4099

It's been a while since I worked on this stuff, so it would be good to have other eyes look at this too.

PR #24664, opened 2019-11-05
Fix loophole in messageport transfer

### LINKED ISSUE #24600: Fix MessagePort transfer bug
To investigate further, but writing it down already:

When a port is "transfer-received" into a realm, the port is made available to script by firing a `MessageEvent` containing it. 

At that point, the port is, while already managed by the global of that realm, in a ["pending"](https://github.com/servo/servo/blob/d671010e466839010ca76c1292b8dc936941a3f0/components/script/dom/globalscope.rs#L197) state because it might be transferred-out of it in the same task in which it is received. 

It's "pendingness" relates mostly to incoming messages for the port that, if any were sent while it was in transfer, would have been [buffered by the constellation](https://github.com/servo/servo/blob/d671010e466839010ca76c1292b8dc936941a3f0/components/constellation/constellation.rs#L188). 

While it is "pending", we don't ask the constellation to "complete the transfer", which essentially means sending the global the buffer of messages, if any. 

The global queues a task to ["mayde add the port"](https://github.com/servo/servo/blob/d671010e466839010ca76c1292b8dc936941a3f0/components/script/dom/globalscope.rs#L546) to a "managed" state, essentially to see if it has been re-transferred in the task firing the `MessageEvent`, or if it is still there. 

If still there, the global then [sends a message to the constellation](https://github.com/servo/servo/blob/d671010e466839010ca76c1292b8dc936941a3f0/components/script/dom/globalscope.rs#L564) asking to "complete the transfer". 

This message is non-blocking, and the answer from the constellation is received sometimes later, at which point the buffered messages will be used to ["sync-up" the port](https://github.com/servo/servo/blob/d671010e466839010ca76c1292b8dc936941a3f0/components/script/dom/globalscope.rs#L298).

So, in between sending the message to the constellation, and gettting the "complete transfer" answer, script keeps running. 

While no `MessageEvent` can be fired on the port in that time, since the constellation will actually keep buffering any messages sent until it handles the "complete transfer" request, at which point it will forward it's buffer to script, it's actually still possible for script to store the port somewhere and access it in another type of task, and transfer it out!

For example, a `setTimeout` could be used from the `onmessage` handler in which the port is initially received. If the timer fires after the "maybe add port" task, but before the "complete transfer" messages has been received by the constellation, we'll lose any messages the constellation would have buffered(because they will be received by the global who isn't managing the port by then anymore).

So, I think the simplest solution is the make the "complete transfer" call from script to the constellation sync. One could alternatively do something by sending the buffer back to the constellation if the port is gone by the time the answer from the constellation is received by the global, however that sounds a bit like asking for more troubles. 

It's probably worth writing a test for, since this isn't covered by the current suite.

gterzian:
---

- [ ] These changes fix #24600 (GitHub issue number if applicable)

### DIFF
--- a/components/script_traits/script_msg.rs
+++ b/components/script_traits/script_msg.rs
@@ -30,6 +30,7 @@ use net_traits::storage_thread::StorageType;
 use net_traits::CoreResourceMsg;
 use servo_url::ImmutableOrigin;
 use servo_url::ServoUrl;
+use std::collections::{HashMap, VecDeque};
 use std::fmt;
 use style_traits::viewport::ViewportConstraints;
 use style_traits::CSSPixel;
@@ -114,6 +115,17 @@ pub enum HistoryEntryReplacement {
 /// Messages from the script to the constellation.
 #[derive(Deserialize, Serialize)]
 pub enum ScriptMsg {
+    /// Request to complete the transfer of a set of ports to a router.
+    CompleteMessagePortTransfer(MessagePortRouterId, Vec<MessagePortId>),
+    /// The results of attempting to complete the transfer of a batch of ports.
+    MessagePortTransferResult(
+        /* The router whose transfer of ports succeeded, if any */
+        Option<MessagePortRouterId>,
+        /* The ids of ports transferred successfully */
+        Vec<MessagePortId>,
+        /* The ids, and buffers, of ports whose transfer failed */
+        HashMap<MessagePortId, VecDeque<PortMessageTask>>,
+    ),
     /// A new message-port was created or transferred, with corresponding control-sender.
     NewMessagePort(MessagePortRouterId, MessagePortId),
     /// A global has started managing message-ports
@@ -248,6 +260,8 @@ impl fmt::Debug for ScriptMsg {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         use self::ScriptMsg::*;
         let variant = match *self {
+            CompleteMessagePortTransfer(..) => "CompleteMessagePortTransfer",
+            MessagePortTransferResult(..) => "MessagePortTransferResult",
             NewMessagePortRouter(..) => "NewMessagePortRouter",
             RemoveMessagePortRouter(..) => "RemoveMessagePortRouter",
             NewMessagePort(..) => "NewMessagePort",
--- a/tests/wpt/mozilla/tests/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.js
+++ b/tests/wpt/mozilla/tests/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.js
@@ -1,0 +1,33 @@
+// META: script=/common/get-host-info.sub.js
+
+async_test(function(t) {
+  var channel1 = new MessageChannel();
+  var channel2 = new MessageChannel();
+  var host = get_host_info();
+  let iframe = document.createElement('iframe');
+  iframe.src = host.HTTP_NOTSAMESITE_ORIGIN + "/webmessaging/support/ChildWindowPostMessage.htm";
+  document.body.appendChild(iframe);
+  var TARGET = document.querySelector("iframe").contentWindow;
+  iframe.onload = t.step_func(function() {
+
+    // Send a message, expecting it to be received in the iframe.
+    channel1.port2.postMessage(1)
+
+    // First, transfer the port into the same realm.
+    channel2.port2.postMessage(0, [channel1.port1]);
+
+    channel2.port1.onmessage = t.step_func(function (evt) {
+      assert_equals(Number(evt.data), 0);
+
+      t.step_timeout(function () {
+        // Transfer the port to the iframe.
+        TARGET.postMessage("ports", "*", evt.ports);
+      }, 0);
+    });
+
+    channel1.port2.onmessage = t.step_func(function (evt) {
+      assert_equals(Number(evt.data), 1);
+      t.done();
+    });
+  });
+}, `A port transferred outside of a onmessage handler does not lose messages along the way.`);
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -186,6 +186,18 @@ enum TransferState {
     /// The port is currently in-transfer,
     /// and incoming tasks should be buffered until it becomes managed again.
     TransferInProgress(VecDeque<PortMessageTask>),
+    /// A global has requested the transfer to be completed,
+    /// it's pending a confirmation of either failure or success to complete the transfer.
+    CompletionInProgress(MessagePortRouterId),
+    /// While a completion of a transfer was in progress, the port was shipped,
+    /// hence the transfer failed to complete.
+    /// We start buffering incoming messages,
+    /// while awaiting the return of the previous buffer from the global
+    /// that failed to complete the transfer.
+    CompletionFailed(VecDeque<PortMessageTask>),
+    /// While a completion failed, another global requested to complete the transfer.
+    /// We are still buffering messages, and awaiting the return of the buffer from the global who failed.
+    CompletionRequested(MessagePortRouterId, VecDeque<PortMessageTask>),
     /// The entangled port has been removed while the port was in-transfer,
     /// the current port should be removed as well once it is managed again.
     EntangledRemoved,
@@ -1553,6 +1565,13 @@ where
         };
 
         match content {
+            FromScriptMsg::CompleteMessagePortTransfer(router_id, ports) => {
+                self.handle_complete_message_port_transfer(router_id, ports);
+            },
+            FromScriptMsg::MessagePortTransferResult(router_id, succeeded, failed) => {
+                self.handle_message_port_transfer_completed(router_id, succeeded);
+                self.handle_message_port_transfer_failed(failed);
+            },
             FromScriptMsg::RerouteMessagePort(port_id, task) => {
                 self.handle_reroute_messageport(port_id, task);
             },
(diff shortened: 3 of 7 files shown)


gterzian:
@bors-servo try=wpt 

- expect to hit `panic!("CompleteTransfer msg received in a global not managing the port.");`

gterzian:
Bingo

```
  ▶ CRASH [expected OK] /_mozilla/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.html
  │ 
  │ 
  │ 
  │ 
  │ CompleteTransfer msg received in a global not managing the port. (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at components/script/dom/globalscope.rs:304)
  │ stack backtrace:
  │    0: servo::main::{{closure}}
  │    1: std::panicking::rust_panic_with_hook
  │              at src/libstd/panicking.rs:472
  │    2: std::panicking::begin_panic
  │    3: <script::dom::globalscope::MessageListener::notify::process_complete_transfer<F> as script::task::TaskOnce>::run_once
  │    4: <T as script::task::TaskBox>::run_box
  │    5: script::script_thread::ScriptThread::handle_msg_from_script
  │    6: _ZN6script13script_thread12ScriptThread11handle_msgs17h50c924a64a0224bfE.llvm.3934872618367253061
  │    7: script::script_thread::ScriptThread::start
  │    8: profile_traits::mem::ProfilerChan::run_with_memory_reporting
  │    9: std::sys_common::backtrace::__rust_begin_short_backtrace
  │   10: _ZN3std9panicking3try7do_call17h12a91747562b248cE.llvm.10564501558317530913
  │   11: __rust_maybe_catch_panic
  │              at src/libpanic_unwind/lib.rs:80
  │   12: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   13: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/246be7e1a557b8ac8287c6842379a0db67770be6/src/liballoc/boxed.rs:942
  │              at /rustc/246be7e1a557b8ac8287c6842379a0db67770be6/src/liballoc/boxed.rs:942
  │       std::sys_common::thread::start_thread
  │              at src/libstd/sys_common/thread.rs:13
  │              at src/libstd/sys/unix/thread.rs:79
  │   15: start_thread
  │   16: clone
  │ 
  │ [2019-11-05T14:44:37Z ERROR servo] CompleteTransfer msg received in a global not managing the port.
```

gterzian:
```
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected 1 but got 2", 
    "stack": "channel2.port2.onmessage</evt.ports[0].onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_with_transfer_entangled.any.js:23:22\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\n", 
    "subtest": "An entangled port transferred to the same origin receives messages in order", 
    "test": "/webmessaging/Channel_postMessage_with_transfer_entangled.any.html", 
    "line": 146223, 
    "action": "test_result", 
    "expected": "PASS"
}
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected 1 but got 2", 
    "stack": "channel2.port2.onmessage</evt.ports[0].onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_with_transfer_entangled.any.js:23:22\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\n", 
    "subtest": "An entangled port transferred to the same origin receives messages in order", 
    "test": "/webmessaging/Channel_postMessage_with_transfer_entangled.any.worker.html", 
    "line": 178072, 
    "action": "test_result", 
    "expected": "PASS"
}
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected \"First\" but got \"Fourth\"", 
    "stack": "channel2.port2.onmessage</channel3.port1.onmessage</evt.ports[0].onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_with_transfer_incoming_messages.any.js:10:11\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\n", 
    "subtest": "When transferring a non-enabled port mutiple times,\n    incoming messages sent at various transfer steps are received in order upon enablement.", 
    "test": "/webmessaging/Channel_postMessage_with_transfer_incoming_messages.any.html", 
    "line": 182715, 
    "action": "test_result", 
    "expected": "PASS"
}
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected \"First\" but got \"Fourth\"", 
    "stack": "channel2.port2.onmessage</channel3.port1.onmessage</evt.ports[0].onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_with_transfer_incoming_messages.any.js:10:11\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\n", 
    "subtest": "When transferring a non-enabled port mutiple times,\n    incoming messages sent at various transfer steps are received in order upon enablement.", 
    "test": "/webmessaging/Channel_postMessage_with_transfer_incoming_messages.any.worker.html", 
    "line": 225664, 
    "action": "test_result", 
    "expected": "PASS"
}
```

gterzian:
```
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_true: expected true got false", 
    "stack": "c.port1.onmessage<@http://web-platform.test:8000/webmessaging/message-channels/close.html:43:9\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func_done/<@http://web-platform.test:8000/resources/testharness.js:1948:32\n@http://web-platform.test:8000/webmessaging/message-channels/close.html:46:13\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nasync_test@http://web-platform.test:8000/resources/testharness.js:576:22\n@http://web-platform.test:8000/webmessaging/message-channels/close.html:39:1\n", 
    "subtest": "Inflight messages should be delivered even when sending port is closed afterwards.", 
    "test": "/webmessaging/message-channels/close.html", 
    "line": 316901, 
    "action": "test_result", 
    "expected": "PASS"
}
{
    "status": "FAIL", 
    "group": "default", 
    "message": "assert_equals: expected 0 but got 1", 
    "stack": "iframe.onload</channel1.port1.onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_transfer_xsite_incoming_messages.window.js:13:7\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\niframe.onload</channel1.port1.onmessage<@http://web-platform.test:8000/webmessaging/Channel_postMessage_transfer_xsite_incoming_messages.window.js:16:22\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\niframe.onload<@http://web-platform.test:8000/webmessaging/Channel_postMessage_transfer_xsite_incoming_messages.window.js:23:20\nTest.prototype.step@http://web-platform.test:8000/resources/testharness.js:1908:25\nTest.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:1932:35\n", 
    "subtest": "Tasks enqueued on the port-message-queue of an enabled port,\n  are transferred along with the port, when the transfer happens in the same task\n  during which postMessage is called", 
    "test": "/webmessaging/Channel_postMessage_transfer_xsite_incoming_messages.window.html", 
    "line": 339468, 
    "action": "test_result", 
    "expected": "PASS"
}
{
    "status": "ERROR", 
    "group": "default", 
    "message": "uncaught exception: unknown (can't convert to string)", 
    "stack": "w.onmessage@http://web-platform.test:8000/webmessaging/message-channels/worker-post-after-close.html:31:7\n", 
    "subtest": null, 
    "test": "/webmessaging/message-channels/worker-post-after-close.html", 
    "line": 364321, 
    "action": "test_result", 
    "expected": "OK"
}
```

gterzian:
```
{
    "status": "CRASH", 
    "group": "default", 
    "message": null, 
    "stack": null, 
    "subtest": null, 
    "test": "/_mozilla/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.html", 
    "line": 207944, 
    "action": "test_result", 
    "expected": "OK"
}
```

Ok so sync communication with the constellation doesn't help, because it messes up the ordering of messages in the following way:

1. A port is transfer-received. A task is scheduled to "maybe add the port" and complete it's transfer with the constellation.
2. `postMessage` is called on the entangled port, that happens to be in the same realm. A task is schedule to "route" that message to the entangled port.
3. The task to "maybe add port" runs, transfer completed sync, port marked as ready to go. If the port is started, a task will be queued to dispatch message events based on it's current buffer, including the buffer received from the constellation.
4. The "route" task queued at 2 runs, since the port is not awaiting transfer anymore, the message event is dispatched.
5. Problem: the task queued at 2 runs before the buffered messages received at 3. 

The "async" behavior is different at:
3.  The task to "maybe add port" runs, transfer completed message sent to the constellation without blocking.
4. The "route" task queued at 2 runs, the port is still awaiting transfer(since we haven't received a response from the constellation yet), so the message is buffered locally.
5. The "complete transfer" message from the constellation is received, the local and constellation buffer are reconciled and a task is queued to dispatch message events. The ordering of messages is right, because 4 buffered the message.

However the problem is as initially stated, which is a crash in the new test because the port is transferred out before 5. 

So the solution is going to have to be sending the buffer back to the constellation, and additional messaging to prevent the constellation from completing the transfer(the second one), before it either receives the buffer sent in the "complete transfer" back(to ensure it's not lost), or a async confirmation that the buffer has been used to complete the transfer.

gterzian:
```
{
    "status": "TIMEOUT", 
    "group": "default", 
    "message": "Test timed out", 
    "stack": null, 
    "subtest": "Test Description: postMessage to cross-site iframe with MessagePort array containing 100 ports.", 
    "test": "/webmessaging/postMessage_MessagePorts_xsite.sub.window.html", 
    "line": 363832, 
    "action": "test_result", 
    "expected": "PASS"
}
```

Ok so the problem now, is that in between the "complete transfer" and the "transfer completed" flow, intermittently the constellation might buffer some messages, which are then lost. 

So we really need a fully complicated flow where the track the state at each step, with confirmation of transfer having been completed, or otherwise returning the buffer to the constellation(instead of just keeping it there, where it can lead to erroneous buffering).

And we probably also need to take into account that not only the port can be transferred again and the transfer completion "fails", but also that another global can start sending a request to complete the transfer as well. 

Nothing some additional variant of `TransferState` and some back and forth can't solve.

gterzian:
wpt tests seem to pass

Screen Shot 2019-11-08 at 10 54 50 PM


@jdm r?

tests/wpt/mozilla/tests/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.js
```diff
@@ -0,0 +1,33 @@
...
+        TARGET.postMessage("ports", "*", evt.ports);
+      }, 0);
+    });
+
+    channel1.port2.onmessage = t.step_func(function (evt) {
+      assert_equals(Number(evt.data), 1);
+      t.done();
+    });
+  });
+}, `A port transferred outside of a onmessage handler does not loose messages along the way.`);
```
jdm: nit: s/loose/lose/

components/script_traits/script_msg.rs
```diff
@@ -118,12 +118,12 @@ pub enum ScriptMsg {
...
-    MessagePortTransferCompleted(MessagePortRouterId, Vec<MessagePortId>),
-    /// A transfer of ports has failed.
-    /// Returns the buffer of tasks that was received for each port.
-    MessagePortTransferFailed(
+    /// The results of attempting to complete the transfer of a batch of ports.
+    MessagePortTransferResult(
         MessagePortRouterId,
+        /* The id's of ports transferred successfully */
+        Vec<MessagePortId>,
+        /* The id's, and buffers, of ports whose transfer failed */
```
jdm: nit: s/id's/ids/

tests/wpt/mozilla/tests/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.js
```diff
@@ -22,6 +22,14 @@ async_test(function(t) {
       t.step_timeout(function () {
         // Transfer the port to the iframe.
         TARGET.postMessage("ports", "*", evt.ports);
+
+        // Keep the event-loop busy for a while,
+        // which will intermittently result in the iframe
+        // starting the "complete port transfer" flow,
+        // before the window global could finish it's own.
+        for (var i = 0; i < 99999; i++) {
```
jdm: Should we use Date.now() and block for a specific amount of time instead? Alternatively, a synchronous XHR that delays the response by a certain amount of time.

jdm (changes requested):
Looks like a good set of improvements!

tests/wpt/mozilla/tests/mozilla/Channel_postMessage_with_second_transfer_in_timeout.window.js
gterzian: I think it might be better to keep it loose, because the unknown is how long it takes for the iframe to start the workflow with the constellation, which is going to vary(it's basically the IPC). So I wouldn't know exactly for how long to block the event-loop.

Additionally, the test will cover all branches in the code, albeit not within the same run, due to the intermittent nature of whether the iframe requests the completion of the transfer, or not, by the time the event-loop is unblocked(and the test shold pass either way).

gterzian: Ok I'm trying to write two tests instead, and it seems it's intermittent for another reason, maybe a bug. Looking into it...

gterzian: Ah ok I get it, the timer doesn't actually go through the task queue, it's fired directly when receiving a message from the scheduler, so sometimes it "jumps ahead" of the task queued to "maybe add the port", and then the port is transferred into the iframe before the global even had a chance to "add it" and request completion from the constellation. 

The goal of the test is for the "maybe add pending ports" task to run first, so the global requests completion, and then run the timer which transfers out the port before the global has received the buffer from the constellation, hence branching into the "transfer failed" stuff. 

But when the timer runs first, then the port is simply transferred and the global will not try to complete the transfer since the port is gone.

gterzian: Ok so I think it's actually pretty hard to write a test that will be deterministic, so there are in fact three possibilities:
1. The timer fires first, port goes straight to the iframe.
2. The "maybe add pending port" task runs first, starts the complete transfer flow, then the timer runs, port is transferred, then the global returns the buffer, then the iframe requests transfer completion.
3. Same as 2, except the iframe requests transfer completion before the other global returned the buffer. 

Maybe we can just keep one test, since it passes in all cases, despite it being rather fuzzy?

jdm: The reason I ask about using Date.now() is that blocking the event loop for a specific amount of time is easier to reason about than 9999 iterations of an operation like JSON.parse. How does a time-based loop cause different results?

gterzian: Ok, I switched to using a sync XHR.

PR #24669, opened 2019-11-06
Set response mime_type based on Content-Type

### LINKED ISSUE #24628: Response object's mime type is not updated by fetch algorithm
[This code](https://github.com/servo/servo/blob/03f223663f1c0bc7432f0f47b533fb41c0226eb1/components/script/fetch.rs#L227-L251) fills in the DOM response object's headers when an HTTP response is received, but the [mime_type](https://github.com/servo/servo/blob/8e14894faeaefad84debbfa1b5363d5eea5deda0/components/script/dom/response.rs#L38) is not updated based on the Content-Type header.

Code: `components/script/dom/response.rs`, `components/script/fetch.rs`
Tests: `./mach test-wpt tests/wpt/web-platform-tests/fetch/api`

glowe:
Set the response's `mime_type` based on its `Content-Type` header whenever `Response::set_headers` is called. This ensures that `mime_type` is always consistent with `Content-Type`. The other reason for going with this approach is it relies on `extract_mime_type`: https://github.com/servo/servo/blob/afbcbf75eaa63ff0eec8fd3858e9155eb8dbadaa/components/script/dom/headers.rs#L260-L261

Besides the consistency argument, I also expect that we'll update the `extract_mime_type` method to conform with its specification.

Another option I explored was adding a `set_mime_type` member to `Response` and calling that with the mime type from `Metadata::content_type` in `FetchResponseListener::process_response`. I decided against this since it expanded the interface of `Response` and didn't have the same benefits of the first option. 

Fixes #24628.

---

- [x] These changes fix #24628.

### DIFF
--- a/components/script/dom/response.rs
+++ b/components/script/dom/response.rs
@@ -403,6 +403,7 @@ impl Response {
             Some(hyper_headers) => hyper_headers.into_inner(),
             None => HyperHeaders::new(),
         });
+        *self.mime_type.borrow_mut() = self.Headers().extract_mime_type();
     }
 
     pub fn set_raw_status(&self, status: Option<(u16, Vec<u8>)>) {
--- a/tests/wpt/metadata/fetch/api/response/response-consume.html.ini
+++ b/tests/wpt/metadata/fetch/api/response/response-consume.html.ini
@@ -67,9 +67,6 @@
   [Consume response's body: from stream without correct urlencoded type to formData (error case)]
     expected: FAIL
 
-  [Consume response's body: from fetch to blob]
-    expected: FAIL
-
   [Consume response's body: from multipart form data blob to formData]
     expected: FAIL
 
--- a/tests/wpt/metadata/fetch/corb/script-html-via-cross-origin-blob-url.sub.html.ini
+++ b/tests/wpt/metadata/fetch/corb/script-html-via-cross-origin-blob-url.sub.html.ini
@@ -1,6 +1,5 @@
 [script-html-via-cross-origin-blob-url.sub.html]
-  [Untitled]
-    expected: FAIL
+  expected: ERROR
 
   [script-html-via-cross-origin-blob-url]
     expected: FAIL


jdm:
Sorry for the review delay, and thanks for the detailed explanation! I agree that this seems like a sensible way to fix the issue.
@bors-servo r+

jdm:
```
  ▶ ERROR [expected OK] /fetch/corb/script-html-via-cross-origin-blob-url.sub.html
  │   → expected expression, got '<'
  │ 
  └ blob:http://www1.web-platform.test:8000/814fc97067474afd80a7dda71854d8e8:1:0
```
That's interesting...

jdm:
https://searchfox.org/mozilla-central/source/testing/web-platform/tests/fetch/corb/script-html-via-cross-origin-blob-url.sub.html loads https://searchfox.org/mozilla-central/source/testing/web-platform/tests/fetch/corb/resources/subframe-that-posts-html-containing-blob-url-to-parent.html which does use fetch and Response, so we should dig into what's changed there.

glowe:
It looks like setting the mime type is allowing this test (`script-html-via-cross-origin-blob-url.sub.html`)  to progress to a point where it's revealing a bug in the CORB handling in servo. Here's the relevant excerpt:

https://github.com/servo/servo/blob/f65cb94b9e417bdaab6fd8d699bf213ae55cd740/tests/wpt/web-platform-tests/fetch/corb/script-html-via-cross-origin-blob-url.sub.html#L21-L33

`step2_processSubframeMsg` is the function that is called when `subframe-that-posts-html-containing-blob-url-to-parent.html` posts a message back to its parent. Previously, this function was failing at line 23:

```assert_equals(msg.blob_type, 'text/html');```

Now that the proper mime type is set, the test progresses but errors out later. It looks like Servo is allowing [html-correctly-labeled.html](https://github.com/servo/servo/blob/f65cb94b9e417bdaab6fd8d699bf213ae55cd740/tests/wpt/web-platform-tests/fetch/corb/resources/html-correctly-labeled.html) to be added to the document as a script element. 

https://github.com/servo/servo/blob/f65cb94b9e417bdaab6fd8d699bf213ae55cd740/tests/wpt/web-platform-tests/fetch/corb/script-html-via-cross-origin-blob-url.sub.html#L33

The trouble is that this is content actually an HTML page and not javascript, so Servo chokes on the first character of `<!DOCTYPE html>`. I actually changed the content of `html-correctly-labeled.html` to javascript and the test continued but failed with:


Screen Shot 2019-11-14 at 12 42 27 AM

Servo __should__ actually block this content from being loaded, but I suspect that we haven't implemented the appropriate CORB checks yet.

One option to move things forward would be to update `script-html-via-cross-origin-blob-url.sub.html.ini` to expect an `ERROR` rather than a `FAIL`. Another option would be to leave this PR open until we've implemented the proper CORB checks.

jdm:
Aha, thanks! Your analysis makes complete sense; let's update the expected result since we haven't implemented the CORB specification yet.

jdm:
@bors-servo retry
* #23290

PR #24672, opened 2019-11-06
Print an error if Glutin port needs to manage a new tab

paulrouget:
When a tab needs to be created because the user clicks on a link with a `target=…` attribute, the Glutin port just ignores it, creating weird behaviors.

We should at least print an error.


CYBAI:
@bors-servo retry

- #23290

PR #24677, opened 2019-11-06
Ensure all iframes are sized correctly at creation

jdm:
Following in the path of #22395, these commits serve several purposes:
* prevent layout instability early during iframe loads caused by a succession of resize events
* reduce the complexity of determining what actual DPI and initial window size values are being used at startup
* ensure that all documents have a correct initial viewport size at creation

These changes fix problems that were exposed by the changes in #24462 but are independent of that PR.

---

### DIFF
--- a/components/script/dom/htmliframeelement.rs
+++ b/components/script/dom/htmliframeelement.rs
@@ -190,6 +190,7 @@ impl HTMLIFrameElement {
                     load_data: load_data.clone(),
                     old_pipeline_id: old_pipeline_id,
                     sandbox: sandboxed,
+                    window_size,
                 };
                 global_scope
                     .script_to_constellation_chan()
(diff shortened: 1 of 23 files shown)


components/script/dom/htmliframeelement.rs
```diff
@@ -173,6 +172,13 @@ impl HTMLIFrameElement {
             replace: replace,
         };
 
+        let window_size = WindowSizeData {
+            initial_viewport: window
+                .inner_window_dimensions_query(browsing_context_id)
```
jdm: Note: this new query is subtly more correct than the previous use of (effectively) getClientBoundingRect. That query includes the margin/padding/borders that exist on the iframe, so by default that's an extra 2px on every side. The actual viewport for the window should not include these, hence the new layout query.

asajeffrey: +1

asajeffrey (approved):
LGTM

PR #24679, opened 2019-11-07
Run non-macOS tasks on Community-TC for push events

SimonSapin:
These tasks will run on *both* TC deployments, and with https://github.com/servo/saltfs/pull/983 PRs will not land if *either* deployment has any failure. Note that this does not include WPT, so intermittents should not be an issue.


SimonSapin:
@bors-servo p=1

This is one of several remaining steps that we need to complete before the Saturday Nov. 9 deadline.

jdm (approved):

PR #24689, opened 2019-11-07
Move all tasks to Community-TC

SimonSapin:

PR #24694, opened 2019-11-08
Remove use of on_unimplemented

SimonSapin:
It errors in today’s Nightly:

```rust
error[E0557]: feature has been removed
 --> components/script/lib.rs:9:12
  |
9 | #![feature(on_unimplemented)]
  |            ^^^^^^^^^^^^^^^^ feature has been removed

error[E0658]: this is an internal attribute that will never be stable
  --> components/script/dom/bindings/conversions.rs:77:1
   |
77 | #[rustc_on_unimplemented(message = "The IDL interface `{Self}` is not derived from `{T}`.")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/29642
   = help: add `#![feature(rustc_attrs)]` to the crate attributes to enable

```

### DIFF
--- a/components/script/lib.rs
+++ b/components/script/lib.rs
@@ -6,7 +6,6 @@
 #![feature(core_intrinsics)]
 #![feature(drain_filter)]
 #![feature(inner_deref)]
-#![feature(on_unimplemented)]
 #![feature(plugin)]
 #![deny(unsafe_code)]
 #![allow(non_snake_case)]
--- a/components/canvas_traits/webgl.rs
+++ b/components/canvas_traits/webgl.rs
@@ -63,7 +63,7 @@ pub enum WebGLMsg {
         WebGLVersion,
         Size2D<u32>,
         GLContextAttributes,
-        WebGLSender<Result<(WebGLCreateContextResult), String>>,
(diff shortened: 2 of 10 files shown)


SimonSapin:
(Are Homu commands ignored in a merged PR? https://github.com/servo/servo/pull/24689#issuecomment-551572184)

@bors-servo treeclosed-

components/script/lib.rs
```diff
@@ -6,7 +6,6 @@
 #![feature(core_intrinsics)]
 #![feature(drain_filter)]
 #![feature(inner_deref)]
-#![feature(on_unimplemented)]
```
jdm: If we use the rustc_attrs feature, can we keep using rustc_on_unimplemented?

SimonSapin: I haven’t tried, but the wording of the error message seems fairly strong in suggesting we shouldn’t? It would also not-prevent the use of many other attributes.

PR #24711, opened 2019-11-12
Make `./mach rustup` install the rustc-dev component

SimonSapin:


nox:
What's the difference?

SimonSapin:
This only matters if you use `rustup set profile minimal`, see https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles

`rustc-dev` is required for compiler plugins. It is part of the `default` profile but not `minimal`.

Darkspirit:
Yesterday I tried to build Servo on my Raspberry Pi (Raspbian Testing). I had to switch to `minimal` because clippy is not supported on https://rust-lang.github.io/rustup-components-history/armv7-unknown-linux-gnueabihf.html
But then I was stuck because of a sudden "can't find crate for rustc". `rustup component add rust-src` did not help, but **rustc-dev seems to be right and working. Omg, thank you! :)** It would be cool to see WebRender running with llvmpipe. ^^

CYBAI:
@bors-servo retry

- #21696

Issue #24726, opened 2019-11-13
Intermittent reftest failures on linux (blank rendering)


jdm:
I've noticed this in two different pull requests since yesterday, and I'm pretty sure we've been ignoring these intermittent failures for an indeterminate amount of time. One test result looks like:
<img width="631" alt="Screen Shot 2019-11-13 at 10 51 15 AM" src="https://user-images.githubusercontent.com/27658/68779982-8a1c4b00-0603-11ea-941a-10d05197c7ef.png">
The other test result is completely blank. This has been observed with /css/CSS2/text/text-transform-applies-to-013.xht and /css/CSS2/fonts/shand-font-002.xht so far.

jdm:
Observed in /css/CSS2/margin-padding-clear/margin-left-103.xht, so it's not just font-related:
<img width="259" alt="Screen Shot 2019-11-13 at 3 22 29 PM" src="https://user-images.githubusercontent.com/27658/68801127-6c61dc80-0629-11ea-89fb-eb3e78dcfc66.png">

jdm:
One theory: https://github.com/servo/servo/pull/24677 might have caused some layout or compositing code to believe that the window is incorrectly sized.

jdm:
Seen in https://github.com/servo/servo/commit/59eca0260c9800dab52cf314b03554a061a6d733.

jdm:
More specifically, #24700 did not merge but #24701 did which include the same changes.

jdm:
I plan to test this theory by reverting changes since https://github.com/servo/servo/commit/e5689df6b44696bf24f8757cd1f6e14d1aabf6f0 and throwing this at try-wpt many times.

jdm:
Attempting to find the regression range is challenging because:
* we migrated the linux try from buildbot to taskcluster in the window, so this requires cherry-picking the commits from #24753, #24768, #24785, and #24811 in that order
* any pull request for try needs to merge cleanly against master, so we can't just do a `git reset --hard` and cherry-pick the aforementioned commits
* using `git revert abcdef.. -m 1 -n` quickly encounters merge conflicts in #25280 for reason that I do not understand

SimonSapin:
> any pull request for try needs to merge cleanly against master

We could hack something together to run try-like CI without merging to master. Would that help?

jdm:
It would probably help a lot!

SimonSapin:
https://github.com/servo/servo/pull/25351 does that.

jdm:
[...]

PR #24741, opened 2019-11-14
Update Tetsuharu OHZEKI's entry in mailmap

tetsuharuohzeki:
I no longer use these mail addresses (the former is still reachable but I cannot access to the later one). I'll update them with reachable one by myself.


jdm:
@bors-servo r+
Thanks!

saneyuki:
Ugh, is this random failure...?

-----

@jdm  thank you for your review!

CYBAI:
@bors-servo retry

(@saneyuki yes, that's #23290)

saneyuki:
@CYBAI aha, thank you.

jdm:
@CYBAI #23290 is only for mac reftest failures. Linux ones are probably #24726.

CYBAI:
@jdm argh!! thanks for pointing that out!

PR #24744, opened 2019-11-14
[WIP] Install ragel in linux CI.

jdm:
This is a possible workaround for #24611.


SimonSapin:
Isn’t Harfbuzz used on all platforms?

jdm:
Yes, but this is just an experiment from the github editing UI to see what happens.

jdm:
I'll try to remember to do that next time.

SimonSapin:
We would also have needed to install Ragel on Windows, but it looks like https://www.colm.net/open-source/ragel/ doesn’t provide binaries.

PR #24746, opened 2019-11-14
Update surfman-chains

jdm:
This allows us to get usable results out of babylon demos on the HoloLens 2, and gets us much closer to usable results on desktop.


asajeffrey (approved):
LGTM

PR #24754, opened 2019-11-15
Add a gstreamer servosrc plugin

asajeffrey:
A first draft of a gstreamer plugin. Lots of stuff to fix, most importantly that it should write to GL memory, not do readback. But it's a start!

---

- [x] These changes do not require tests because it's a new platform

### DIFF
--- a/ports/gstplugin/resources.rs
+++ b/ports/gstplugin/resources.rs
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// This is a copy of the resource loader from the glutin port
+// TODO: move this to somewhere where it can be shared.
+// https://github.com/servo/servo/issues/24853
+
+use lazy_static::lazy_static;
+use servo::embedder_traits::resources::{self, Resource};
+use std::env;
+use std::fs;
+use std::io;
+use std::path::PathBuf;
+use std::sync::Mutex;
+
+lazy_static! {
+    static ref CMD_RESOURCE_DIR: Mutex<Option<String>> = Mutex::new(None);
+}
+
+struct ResourceReader;
+
+fn filename(file: Resource) -> &'static str {
+    match file {
+        Resource::Preferences => "prefs.json",
+        Resource::BluetoothBlocklist => "gatt_blocklist.txt",
+        Resource::DomainList => "public_domains.txt",
+        Resource::HstsPreloadList => "hsts_preload.json",
+        Resource::SSLCertificates => "certs",
+        Resource::BadCertHTML => "badcert.html",
+        Resource::NetErrorHTML => "neterror.html",
+        Resource::UserAgentCSS => "user-agent.css",
+        Resource::ServoCSS => "servo.css",
+        Resource::PresentationalHintsCSS => "presentational-hints.css",
+        Resource::QuirksModeCSS => "quirks-mode.css",
+        Resource::RippyPNG => "rippy.png",
+        Resource::MediaControlsCSS => "media-controls.css",
+        Resource::MediaControlsJS => "media-controls.js",
+    }
+}
+
+pub fn init() {
+    resources::set(Box::new(ResourceReader));
+}
+
+fn resources_dir_path() -> io::Result<PathBuf> {
+    // This needs to be called before the process is sandboxed
+    // as we only give permission to read inside the resources directory,
+    // not the permissions the "search" for the resources directory.
+    let mut dir = CMD_RESOURCE_DIR.lock().unwrap();
+    if let Some(ref path) = *dir {
+        return Ok(PathBuf::from(path));
+    }
+
+    // FIXME: Find a way to not rely on the executable being
+    // under `<servo source>[/$target_triple]/target/debug`
+    // or `<servo source>[/$target_triple]/target/release`.
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    // Follow symlink
+    path = path.canonicalize()?;
+
+    while path.pop() {
+        path.push("resources");
+        if path.is_dir() {
+            break;
+        }
+        path.pop();
+        // Check for Resources on mac when using a case sensitive filesystem.
+        path.push("Resources");
+        if path.is_dir() {
+            break;
+        }
+        path.pop();
+    }
+    *dir = Some(path.to_str().unwrap().to_owned());
+    Ok(path)
+}
+
+impl resources::ResourceReaderMethods for ResourceReader {
+    fn read(&self, file: Resource) -> Vec<u8> {
+        let file = filename(file);
+        let mut path = resources_dir_path().expect("Can't find resources directory");
+        path.push(file);
+        fs::read(path.clone()).expect(&format!("Can't read file {:?}", path))
+    }
+    fn sandbox_access_files_dirs(&self) -> Vec<PathBuf> {
+        vec![resources_dir_path().expect("Can't find resources directory")]
+    }
+    fn sandbox_access_files(&self) -> Vec<PathBuf> {
+        vec![]
+    }
+}
--- a/ports/gstplugin/servosrc.rs
+++ b/ports/gstplugin/servosrc.rs
@@ -1,0 +1,618 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use crate::logging::CATEGORY;
+
+use crossbeam_channel::Receiver;
+use crossbeam_channel::Sender;
+
+use euclid::Point2D;
+use euclid::Rect;
+use euclid::Scale;
+use euclid::Size2D;
+
+use glib::glib_bool_error;
+use glib::glib_object_impl;
+use glib::glib_object_subclass;
+use glib::object::Cast;
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::object::ObjectImplExt;
+use glib::subclass::simple::ClassStruct;
+use glib::subclass::types::ObjectSubclass;
+use gstreamer::gst_element_error;
+use gstreamer::gst_loggable_error;
+use gstreamer::subclass::element::ElementClassSubclassExt;
+use gstreamer::subclass::element::ElementImpl;
+use gstreamer::subclass::ElementInstanceStruct;
+use gstreamer::BufferRef;
+use gstreamer::Caps;
+use gstreamer::CoreError;
+use gstreamer::ErrorMessage;
+use gstreamer::FlowError;
+use gstreamer::FlowSuccess;
+use gstreamer::Format;
+use gstreamer::Fraction;
+use gstreamer::FractionRange;
+use gstreamer::IntRange;
+use gstreamer::LoggableError;
+use gstreamer::PadDirection;
+use gstreamer::PadPresence;
+use gstreamer::PadTemplate;
+use gstreamer_base::subclass::base_src::BaseSrcImpl;
+use gstreamer_base::BaseSrc;
+use gstreamer_base::BaseSrcExt;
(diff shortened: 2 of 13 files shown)


ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+        let context = device
+            .create_context(&descriptor)
+            .expect("Failed to create context");
+        let mut device = Device::Hardware(device);
+        let mut context = Context::Hardware(context);
+        let gl = Gl::gl_fns(gl::ffi_gl::Gl::load_with(|s| {
+            device.get_proc_address(&context, s)
+        }));
+
+        device.make_context_current(&context).unwrap();
```
sdroege: You probably want to make the context not current at the end of this function. It's not guaranteed (and unlikely) that this is the same function as where `fill()` is called later.

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+    fbo: GLuint,
+}
+
+impl ServoSrcGfx {
+    fn new() -> ServoSrcGfx {
+        let version = surfman::GLVersion { major: 4, minor: 3 };
+        let flags = surfman::ContextAttributeFlags::empty();
+        let attributes = surfman::ContextAttributes { version, flags };
+
+        let connection = surfman::Connection::new().expect("Failed to create connection");
```
sdroege: All this allocation of resources should ideally happen in `ElementImpl::change_state` in `gst::StateChange::NullToReady` and be deallocated again in `ReadyToNull`

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+            "The Servo web browser",
+            env!("CARGO_PKG_AUTHORS"),
+        );
+
+        let src_caps = Caps::new_simple(
+            "video/x-raw",
+            &[
+                ("format", &VideoFormat::Bgrx.to_string()),
+                ("width", &IntRange::<i32>::new(512, 1024)),
+                ("height", &IntRange::<i32>::new(512, 1024)),
```
sdroege: Why limited to these resolutions instead of `(1, i32::MAX)`?

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+                (
+                    "framerate",
+                    &FractionRange::new(Fraction::new(1, 1), Fraction::new(15, 1)),
```
sdroege: Why limited to these framerates instead of `1/MAX` to `MAX/1`?

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+}
+
+impl ObjectImpl for ServoSrc {
+    glib_object_impl!();
+
+    fn constructed(&self, obj: &glib::Object) {
+        self.parent_constructed(obj);
+        let basesrc = obj.downcast_ref::<BaseSrc>().unwrap();
+        basesrc.set_live(true);
+        basesrc.set_format(Format::Time);
```
sdroege: You probably also want `basesrc.do_timestamp(true)` here until you implement more accurate timestamping yourself

ports/gstplugin/servosrc.rs
sdroege: Or `BaseSrcImpl::start` and `::stop`

ports/gstplugin/servosrc.rs
asajeffrey: Done.

ports/gstplugin/servosrc.rs
asajeffrey: Good poimt, that was just for testing. Fixed.

ports/gstplugin/servosrc.rs
asajeffrey: Ditto.

ports/gstplugin/servosrc.rs
asajeffrey:
```
   --> ports/gstplugin/servosrc.rs:454:10
    |
454 |     basesrc.do_timestamp(true);
    |             ^^^^^^^^^^^^ method not found in `&gstreamer_base::auto::base_src::BaseSrc`
```

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,566 @@
...
+            "Servo as a gstreamer src",
+            "Filter/Effect/Converter/Video",
+            "The Servo web browser",
+            env!("CARGO_PKG_AUTHORS"),
+        );
+
+        let src_caps = Caps::new_simple(
+            "video/x-raw",
+            &[
+                ("format", &VideoFormat::Bgrx.to_string()),
```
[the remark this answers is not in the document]
asajeffrey:
```
   --> ports/gstplugin/servosrc.rs:429:47
    |
429 |                 ("format", &VideoFormat::Bgrx.to_str()),
    |                                               ^^^^^^ method not found in `gstreamer_video::auto::enums::VideoFormat`
```

asajeffrey (commented):
Thanks for the review! I implemented \the changes apart from
- The initialization of gfx has to happen in a thread-local block, since it doesn't implement `Send`.
- A couple of your suggestions don't compile, perhaps due to versioning?

asajeffrey:
Annoyingly, though I'd tested this on Mac, it turns out to produce GL errors on Linux, which I'm currently investigating.

ports/gstplugin/servosrc.rs
sdroege: Indeed, but you could start the gfx stuff (and shut down) at that point. Where is that happening currently, where would this code be triggered currently? :)

ports/gstplugin/servosrc.rs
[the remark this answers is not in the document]
sdroege: Oh indeed, that's only in git master (and will be in 0.15). Sorry :) Not too important

ports/gstplugin/servosrc.rs
sdroege: Sorry, `set_do_timestamp(true)`, not `do_timestamp(true)` :)

ports/gstplugin/servosrc.rs
asajeffrey: I could do that, but it's not obvious that the benefit (deallocating on `stop` rather than at the program end) is worth the complexity (making the thread-local an option which refers to a flag saying whether or not the plugin has been started).

ATM the TLS is only allocated when it is used, so the servo thread allocates on startup and the gstreamer threads allocate the first time they call `fill`.

sdroege: Will that work if you try to use multiple servo sources at the same time in the same process?

Doing it globally seems fine too, I assume an application that uses the servo source once is going to not be bothered by the fact that for the whole runtime of the program there will be some servo thread running.

asajeffrey:
Let's see if that builds in CI @bors-servo try=linux

asajeffrey (commented):
The only remaining big issue is pipelines that have more than one servosrc in them. I am tempted to punt on that for the moment.

asajeffrey:
@jdm: If we're prepared to live with it not working on Linux/X11 I think this is ready for review. there are some TODO items, shall I file issues for those?

ports/gstplugin/servosrc.rs
sdroege: Not really, what exactly are you looking for? :) You need to ensure that it's possible to have multiple instances at the same time and that they don't cause each other to break. If that's not possible at all by design (e.g. because exclusive hardware access), then there needs to be a mechanism implemented that the second instance gracefully fails. Ideally during the `NullToReady` state transition of the element.

asajeffrey: Well, docs about what is shared between the multiple instances of the same element would be nice. I'm guessing that the class will only be instantiated once, but each instance of the element will get it's own object?

sdroege: Yes, there's nothing shared between instances other than what you do in your code yourself. In terms of your code, using two `servosrc` would cause two separate `ServoSrc` to be created.

asajeffrey: OK, that will do for the moment, but it's a bit wasteful. Really we should be sharing servo state between the instances, and sending a `NewBrowser` message for each of them. But that can be a follow-up.

python/servo/command_base.py
```diff
@@ -678,6 +678,7 @@ def build_env(self, hosts_file_path=None, target=None, is_build=False, test_unit
             # do not get precedence
             extra_path = [libpath] + extra_path
             extra_lib = [libpath] + extra_path
+            append_to_path_env(libpath, env, "LD_LIBRARY_PATH")
```
jdm: This is already supposed to take place in the `if extra_lib` branch a few lines further down. Can we figure out why that isn't working?

jdm:
Filing issues for outstanding items seems reasonable to me.

python/servo/command_base.py
asajeffrey: Oh I see what you mean. Shouldn't the setting be `extra_lib = [libpath] + extra_lib`?

asajeffrey:
Rebased to get the new newness of surfman-chains 0.2.

asajeffrey:
Servo streaming across the local network:

Screen Shot 2019-11-21 at 5 28 06 PM

ports/gstplugin/servosrc.rs
sdroege: Agreed :)

asajeffrey: Filed https://github.com/servo/servo/issues/24840

asajeffrey (commented):
Thanks for all the help @sdroege! I created a P-gst-plugin label if you're interested to see what the outstanding issues are. https://github.com/servo/servo/labels/P-gst-plugin

sdroege:
> Thanks for all the help @sdroege!

No problem, just CC me in other issues if there's anything GStreamer-related you'd like to have reviewed or if there are any open questions :) I'll also try to follow the changes

asajeffrey:
Thanks! I'll cc you if I hit any gst roadblocks.

asajeffrey:
I should add thanks for writing the tutorials, they made this all possible!

asajeffrey:
OMG, did I succeed in fighting CI? @bors-servo try=windows

asajeffrey:
Oh, did that force-push kill the windows build? @bors-servo try=windows

asajeffrey:
Of course I'm not finished fighting CI, silly me. @bors-servo try=windows

asajeffrey:
And again... @bors-servo try=windows

asajeffrey:
Okay that really really is it for fighting CI. Really. Over to you @ferjm.

ports/gstplugin/resources.rs
```diff
@@ -0,0 +1,91 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// This is a copy of the resource loader from the glutin port
+// TODO: move this to somewhere where it can be shared.
```
ferjm: #24853

ports/gstplugin/servosrc.rs
```diff
@@ -0,0 +1,618 @@
...
+        u64::try_from(self.info.lock().ok()?.as_ref()?.size()).ok()
+    }
+
+    fn start(&self, _src: &BaseSrc) -> Result<(), ErrorMessage> {
+        info!("Starting");
+        Ok(())
+    }
+
+    fn stop(&self, _src: &BaseSrc) -> Result<(), ErrorMessage> {
+        info!("Starting");
```
ferjm: s/Starting/Stopping

ferjm (approved):
Awesome work!

ports/gstplugin/resources.rs
asajeffrey: Done.

ports/gstplugin/servosrc.rs
asajeffrey: Done.

asajeffrey:
Squashed and rebased. @bors-servo r=ferjm

jdm:
@bors-servo retry
* #24726

PR #24753, opened 2019-11-15
Use an already-installed Homebrew at /usr/local

SimonSapin:
This requires https://github.com/servo/taskcluster-config/pull/4 to be deployed.

Having the standard location helps `pkg-config` (CC https://github.com/servo/servo/pull/24688), and allows installing pre-compiled pakcages (which is much faster than compiling from source).

Issue #23655, opened 2019-06-28
Automatically retry WPT tasks with few failures?


SimonSapin:
docker-worker and generic-worker both support a `retry` configuration in task definitions:

* https://github.com/taskcluster/docker-worker/blob/e2eb847d404d7fa776a394c5845c5e98040cc13b/schemas/v1/payload.json#L190-L197
* https://github.com/taskcluster/generic-worker/blob/e31923d3322d3d31af563ae6a1f3e328b9223d5f/multiuser_windows.yml#L222-L237

It takes an list of exit codes. When the task’s command fails with one of these, the worker resolves the task with the queue as “exception” rather than “failed”. The queue will then automatically re-schedule that task again, up to a configurable number of retries (defaults to 5).

-----

In this repository, PRs often fail to land because of some intermittent WPT failure. We mark some test filenames as known intermittents whose failure we ignore, but I suspect that at least some of the source of non-determinism is weakly or not correlated to the filename, as PRs still regularly fail at first but then land after a retry (or a few).

Retrying an entire PR though homu is costly, in terms of overall cycle time. https://github.com/servo/servo/pull/23383 can help but not when another PR was merged in the meantime, which is common when homu’s queue is non-empty as homu will start on the next PR after a failure quicker than a human can type the `retry` command.

Taskcluster queue’s retry mechanism is more fine-grained: task level instead of PR level. When running fewer tests again in the second try (one out of 6 WPT chucks for example), we’re less likely to hit another random/intermittent failure.

~However, the downside is multiplying the time to reporting deterministic test failures when a PR breaks something. This could be limited by setting a low retry count like 2 or 3 for WPT tasks.~ (https://github.com/servo/servo/pull/24768 greatly reduced this time loss.)

@jdm, what do you think?

jdm:
I'm in favour. There are definitely intermittent problems like #23290 that can hit trigger on just about any retreat, which really hurts our throughput.

SimonSapin:
https://github.com/servo/servo/pull/24769 implemented auto-retry, but we didn’t land it:

> This will make PRs much more likely to land without manual retry. However it will give us a lot less opportunity to see which tests fail intermittently, file issues based on test filenames, track them, etc.

> We could imagine some more automation, such as:
> 
> * `./mach filter-intermittent` uploads its filtered summary of unexpected results somewhere
> * If and when that PR ends up landing without change, promote those previous unexpected results to a queue of possible intermittents.
> 
> 
[...]

PR #24768, opened 2019-11-18
Split WPT macOS testing into many more chunks

SimonSapin:
## Before this

Before this PR, we had roughly as many chunks as available workers. Because the the number of test files is a poor estimate for the time needed to run them, we have significant variation in the completion time between chunks when testing one given PR.

https://github.com/servo/taskcluster-config/pull/9 adds a tool to collect this data. Here are two full runs of `test_wpt` before this PR:

https://community-tc.services.mozilla.com/tasks/groups/DBt9ki9gTdWmwAk-VDorzw

```
count 1, total 0:00:32, max: 0:00:32    docker  0:00:32
count 1, total 0:59:14, max: 0:59:14    macos-disabled-mac1     0:59:14
count 6, total 4:12:16, max: 1:01:14    macos-disabled-mac1 WPT 0:40:29 0:18:55 0:46:50 0:44:38 1:01:14 0:40:10
count 1, total 0:55:19, max: 0:55:19    macos-disabled-mac9     0:55:19
count 6, total 4:25:09, max: 1:01:40    macos-disabled-mac9 WPT 0:37:58 0:37:24 0:27:18 1:01:40 0:46:17 0:54:31
```

Times for a given chunk vary between 19 minutes and 61 minutes. Assuming no `try` testing, with Homu’s serial scheduling of `r+` testing this means that that worker sits idle for 42 minutes and our limited CPU resources are under-utilized.

When there *are* `try` PRs being tested however, they compete with each other and any `r+` PR for the same workers. If we get unlucky, a 61 minute task could only *start* after some other tasks have finished, Increasing the overall time-to-merge a lot.

## This

This PR changes the number of chunks to be significantly more than the number of available workers. When one of them finishes, that worker can pick up another one instead of sitting idle.

Now the ratio of number of tasks to number of workers doesn’t matter: the differences in run time between tasks becomes somewhat of an advantage and the distribution to workers evens out on average.

The number 30 is a bit arbitrary. A higher number reduces resource under-utilization, but increases the effect of per-task overhead. The git cache added in https://github.com/servo/servo/pull/24753 reduced that overhead, though.

Another worry I had was whether this would make worse the similar problem of unequal scheduling between processes within a task, where some CPU cores sit idle while the rest processes finish their assigned work.

This turned out not to be enough of a problem to negatively affect the total machine time:

https://community-tc.services.mozilla.com/tasks/groups/VnDac92HQU6QmrpzWPCR2w
```
count 1, total 0:00:48, max: 0:00:48    docker  0:00:48
count 1, total 0:39:04, max: 0:39:04    macos-disabled-mac9     0:39:04
count 31, total 4:03:29, max: 0:15:29   macos-disabled-mac9 WPT
        0:07:26 0:08:39 0:04:21 0:07:13 0:12:47 0:10:11 0:04:01 0:03:36
        0:10:43 0:12:57 0:04:47 0:04:06 0:10:09 0:12:00 0:12:42 0:04:40
        0:04:24 0:12:20 0:12:15 0:03:03 0:07:35 0:11:35 0:07:01 0:04:16
        0:09:40 0:05:08 0:05:01 0:06:29 0:15:29 0:02:28 0:06:27
```

(4h03min is even lower than above, but seems within variation.)

## After this

https://github.com/servo/servo/issues/23655 proposes automatically restarting failed WPT tasks, in case the failure is intermittent. With the test suite split into more chunks we have fewer tests per chunk, and therefore lower probability that a given one fails. Restarting one of them also causes less repeated work.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -14,6 +14,7 @@ tasks:
         else: proj-servo
       created: {$fromNow: ''}
       deadline: {$fromNow: '1 day'}
+      priority: high
       extra:
         treeherder:
           machine: {platform: Linux}
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -645,7 +645,7 @@ def macos_wpt():
 def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
                repo_dir, chunks="all", repo_kwargs={}):
     if chunks == "all":
-        chunks = [n + 1 for n in range(total_chunks)]
+        chunks = range(total_chunks + 1)
     for this_chunk in chunks:
         task = (
             make_chunk_task("WPT chunk %s / %s" % (this_chunk, total_chunks))
@@ -662,9 +662,10 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
                 GST_DEBUG="3",
             )
         )
-        if this_chunk == chunks[-1]:
-            task.name += " + extra"
-            task.extra["treeherder"]["symbol"] += "+"
+        # `test-wpt` is piped into `cat` so that stdout is not a TTY
+        # and wptrunner does not use "interactive mode" formatting:
+        # https://github.com/servo/servo/issues/22438
+        if this_chunk == 0:
             task.with_script("""
                 ./mach test-wpt-failure
                 time ./mach test-wpt --release --binary-arg=--multiprocess \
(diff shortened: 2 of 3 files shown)


SimonSapin:
@bors-servo try=wpt-mac

https://github.com/servo/taskcluster-config/pull/10

SimonSapin:
(One WPT test has unexpected result)

This is 92 minutes from scheduling to results. After the release build task is finished, this makes better use of all available workers (which are currently 8, not just 6).

PR #24769, opened 2019-11-18
Automatically retry failing WPT tasks (once)

### LINKED ISSUE #23655: Automatically retry WPT tasks with few failures?

SimonSapin:
Fixes https://github.com/servo/servo/issues/23655

Only when `./mach filter-intermittents` finds failures that are not known intermittents.

This will make PRs much more likely to land without manual retry. However it will give us a lot less opportunity to see which tests fail intermittently, file issues based on test filenames, track them, etc.

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -660,6 +660,7 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
             .with_index_and_artifacts_expire_in(log_artifacts_expire_in)
             .with_max_run_time_minutes(90)
             .with_retry_on_exit_codes(TASKCLUSTER_RETRY_EXIT_CODE)
+            .with_max_auto_retries(1)  # This does not include the initial run
             .with_env(
                 TOTAL_CHUNKS=str(total_chunks),
                 THIS_CHUNK=str(this_chunk),
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -659,6 +659,7 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
             .with_script("tar -xzf target.tar.gz")
             .with_index_and_artifacts_expire_in(log_artifacts_expire_in)
             .with_max_run_time_minutes(90)
+            .with_retry_on_exit_codes(TASKCLUSTER_RETRY_EXIT_CODE)
             .with_env(
                 TOTAL_CHUNKS=str(total_chunks),
                 THIS_CHUNK=str(this_chunk),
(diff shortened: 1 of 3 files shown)


nox:
LGTM, will let @jdm sign it off given that's quite a change.

SimonSapin:
@jdm What do you think? You spoke in favor in https://github.com/servo/servo/issues/23655, but I don’t think we’d discussed the loss of tracking.

SimonSapin:
We could imagine some more automation, such as:

* `./mach filter-intermittent` uploads its filtered summary of unexpected results somewhere
* If and when that PR ends up landing without change, promote those previous unexpected results to a queue of possible intermittents.

However if triaging not anymore part of the expected retry process, it could grow until someone gives it attention.

jdm:
Yeah, I think my biggest concern here is losing the ability to determine a meaningful window in which an intermittent failure appeared. We could build some tooling based on the disabled intermittent failure reporting that already exists that only reports the filtered summary (like you suggested), and then make it someone's responsibility to check every day or two and file new issues.

SimonSapin:
Alright. I’m not planning to build that tooling soon, so let’s close this PR and keep #23655 open.

PR #24777, opened 2019-11-18
Enable webxr by default

jdm:
This changes two things:
* enable WebXR by default, since it's not exactly an experimental technology in Servo anymore
* enable the glwindow webxr backend by default so it's easier to test WebXR on desktop nightlies

Since the glwindow pref is only checked in the glutin desktop port, enabling it for other ports does not hurt us at all and it removes a papercut for testing WebXR content on desktop.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/24777)


Manishearth (approved):
(Wouldn't this require interfaces.html updates in the mozilla tests?)

So I'm okay with landing this, but also I'm a bit wary: this means we're releasing webxr by default, and we need to figure out how that factors in to web stability stuff. Firstly: Are we allowed to do this? Is it okay to release a browser containing currently-still-in-flux APIs? I guess webxr is very close to shipping/stabilizing so perhaps we're fine on that front.

Also, does this count as a separate shipped implementation according to the w3c? (because then servo+firefox might automatically count for various spec stability requirements). Servo is currently only "shipped" as a nightly, so perhaps none of this is relevant yet. But I don't know.

cc @larsbergstrom

Manishearth:
@bors-servo r+

from reading https://www.w3.org/2019/Process-20190301/#implementation-experience it seems like it doesn't really matter if it's enabled by default or not

larsbergstrom:
As Servo is not currently shipping in a production browser to end users, it has not been considered a separate "shipped" browser in the W3C/Khronos sense to date. Once that does happen, the specific browser context (e.g., Firefox Reality on Hololens2) would "count", but the engine would be a slightly different thing, as it is for the difference between blink or webkit implementing something vs. Chrome or Safari shipping it. e.g., last I checked WebVR is in webkit due to some awesome work by Igalia, but is not on by default in any webkit-derived browser.

CYBAI:
```
 0:09.11(B ERROR(B Manifest /repo/tests/wpt/mozilla/meta/MANIFEST.json is outdated, use |./mach update-manifest| to fix.
```

Issue #24598, opened 2019-10-31
Properly fire touch, pinch and mouse events on HoloLens and Desktop


paulrouget:
As of now we have a bit of a mix of everything.

We should:
- send touch events in hololens
- send mouse event on Desktop
- only scroll on pan in hololens (use wheel on Desktop)
- use pinch methods on hololens

paulrouget:
Fixed by #24778

PR #24778, opened 2019-11-19
UWP: Handle touch events

### LINKED ISSUE #24698: Hololens click events not working for some people
I first thought this was a platform issue, but some folks (including @pcwalton) are having trouble clicking links (etc) on Servo's 2D hololens port.

It might be related to https://github.com/servo/servo/pull/24547 . I haven't really investigated this since I can't easily replicate it myself, but I think these folks poke instead of tap, and that may trigger mouse events instead of touch events or something.

I _think_ the "tap shadow ripple" which happens when you click on servo is still visible for these folks. (Is that `panel.PointerPressed()`)

The DOM synthesizes events at some point and perhaps our events are in conflict with those.

I might let Patrick use the device with some logging at some point.

cc @paulrouget

Manishearth:
Fixes https://github.com/servo/servo/issues/24698

This means that we can start using the click emulation from https://github.com/servo/servo/blob/master/components/compositing/touch.rs , which is better at handling this than UWP. UWP requires a specific short tap to trigger OnTapped, and lots of folks seem to have trouble doing that.

I'm not 100% sure if this is correct since it doesn't quite disable the existing scroll handling (the ManipulationDelta stuff); I'm wondering if we should. Removing that code seems to make scrolling a bit jumpy (i also got some random zoom events?), which is counterintuitive -- if anything it should be jumpy when we have multiple scroll systems active! If the existing manipulation delta code is not removed, `preventDefault()` on touch events does not prevent scrolling, which isn't great.

r? @paulrouget 

cc @jdm

### DIFF
--- a/support/hololens/ServoApp/ServoControl/ServoControl.cpp
+++ b/support/hololens/ServoApp/ServoControl/ServoControl.cpp
@@ -38,9 +38,9 @@ void ServoControl::OnLoaded(IInspectable const &, RoutedEventArgs const &) {
   auto panel = Panel();
   panel.Tapped(std::bind(&ServoControl::OnSurfaceTapped, this, _1, _2));
   panel.PointerPressed(
-      std::bind(&ServoControl::OnSurfacePointerPressed, this, _1, _2));
+      std::bind(&ServoControl::OnSurfacePointerPressed, this, _1, _2, true));
   panel.PointerReleased(
-      std::bind(&ServoControl::OnSurfacePointerPressed, this, _1, _2));
+      std::bind(&ServoControl::OnSurfacePointerPressed, this, _1, _2, false));
   panel.PointerCanceled(
       std::bind(&ServoControl::OnSurfacePointerCanceled, this, _1, _2));
   panel.PointerCaptureLost(
@@ -105,6 +105,7 @@ void ServoControl::OnSurfaceManipulationDelta(
 
 void ServoControl::OnSurfaceTapped(IInspectable const &,
                                    Input::TappedRoutedEventArgs const &e) {
+
   auto coords = e.GetPosition(Panel());
   auto x = coords.X * mDPI;
   auto y = coords.Y * mDPI;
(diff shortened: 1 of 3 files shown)


support/hololens/ServoApp/DefaultUrl.h
```diff
@@ -1,3 +1,3 @@
 #pragma once
 
-#define DEFAULT_URL L"https://servo.org/hl-home/";
+#define DEFAULT_URL L"https://manishearth.net/sand/inputs.html";
```
paulrouget: Remove.

support/hololens/ServoApp/ServoControl/ServoControl.cpp
```diff
@@ -113,9 +114,10 @@ void ServoControl::OnSurfaceTapped(IInspectable const &,
 }
 
 void ServoControl::OnSurfacePointerPressed(
-    IInspectable const &, Input::PointerRoutedEventArgs const &e) {
-  if (e.Pointer().PointerDeviceType() ==
-      Windows::Devices::Input::PointerDeviceType::Mouse) {
+    IInspectable const &, Input::PointerRoutedEventArgs const &e, bool down) {
+  log("pointer pressed");
```
paulrouget: Remove. Unnecessary noise.

paulrouget:
r=me with the nits addressed. This fixes part of #24598.

paulrouget:
Before landing that, let me take a look at the scrolling problem you mentioned.

Manishearth:
Updated.

paulrouget:
After testing this, I think we need to investigate a bit more what should be done here. Scrolling is badly broken.

If the main issue you want to address here is the missed click events, maybe that's something we should be reporting the MS people.

We could disable the ManipulationDelta, but we would lose the native scroll behavior. It's, in general, something I'd like to avoid across our different ports.

paulrouget:
If found few issues with the touch events:

- 2 touch down from UWP can happen, from one finger, without the intermediate touch up. Servo interpret that as another finder, triggering a pinch. We should send a touch_cancel on PointerExited as well.
- UWP send pointer events even after sending a Lost event. Servo doesn't like that. Pinching doesn't work because of that.
- Sometimes, Servo crashes with using 2 fingers. Not sure why yet.
- touchmove doesn't trigger the CSS :hover state

I've investigating all these issues.

paulrouget:
I think I managed to fix most of the weirdness.
Closing this PR in favor of https://github.com/servo/servo/pull/25076

PR #24782, opened 2019-11-19
Prepend `-Ztimings=info`, don’t append, so it can be overridden

SimonSapin:


jdm:
@bors-servo retry
#13480

PR #24784, opened 2019-11-19
Use harfbuzz unity build branch.

jdm:
This is just verifying that the build continues to work on all platforms.


jdm:
Ignoring the expected failures around the lockfile changing, https://community-tc.services.mozilla.com/tasks/SebVW_o9QFmRhdjn1EyipQ is an android build where the build step succeeded.

PR #24418, opened 2019-10-10
[WIP] Enable WPT on Linux in Taskcluster

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/24418)

PR #24785, opened 2019-11-19
Move WPT Linux from Buildbot to Taskcluster

SimonSapin:
Closes https://github.com/servo/servo/pull/24418

### DIFF
--- a/etc/ci/buildbot_steps.yml
+++ b/etc/ci/buildbot_steps.yml
@@ -18,45 +18,6 @@ linux-rel-nogate:
   - env RUSTFLAGS= bash ./etc/ci/mutation_test.sh
   - ./etc/ci/clean_build_artifacts.sh
 
-linux-rel-wpt:
-  env:
-    CCACHE: sccache
-    RUSTC_WRAPPER: sccache
-    CC: clang-4.0
-    CXX: clang++-4.0
-  commands:
-    - ./mach clean-nightlies --keep 3 --force
-    - ./mach clean-cargo-cache --keep 3 --force
-    - ./etc/ci/clean_build_artifacts.sh
-    - ./mach build --release --with-debug-assertions -p servo
-    - ./mach test-wpt-failure
-    - ./mach test-wpt --release --processes 24 --total-chunks 2 --this-chunk 1 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log --always-succeed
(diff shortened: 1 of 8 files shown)


SimonSapin:
This run looks pretty good, I think? https://community-tc.services.mozilla.com/tasks/groups/XVj2vEiQT_K-RK9MYhBJlA

Filtered log:

```
  ▶ TIMEOUT [expected OK] /webdriver/tests/element_send_keys/file_upload.py
  │ 
  └ [2019-11-19T16:40:02Z ERROR script::dom::bindings::error] Error at :2:28 arguments[0] is undefined

  ▶ PASS [expected FAIL] /_mozilla/css/text_decoration_underline_subpx_a.html

  ▶ PASS [expected FAIL] /css/css-text/word-break/word-break-break-all-005.html

  ▶ OK [expected CRASH] /_mozilla/mozilla/fetch_cannot_overwhelm_system.window.html

jdm (approved):

SimonSapin:
A second run got the exact same filtered unexpected results, so they don’t look intermittent: https://community-tc.services.mozilla.com/tasks/groups/UhZtEvtISO-VD3qkSvNRRA

I don’t have a precise explanation for these differences with Buildbot, but we’re running Ubuntu 18.04 instead of 14.04 and Ubuntu’s gstreamer instead of https://github.com/servo/servo/wiki/How-to-generate-GStreamer-binaries-for-CI

jdm:
Yeah, I would go ahead and add linux-specific expected results for those tests.

SimonSapin:
Alright, let’s see how this looks.

@bors-servo try=wpt

SimonSapin:
Currently this runs on GCP instances that have 16 “vCPU” (which as I understand it map to 8 cores / 16 threads). We can change max capacity (number of instances running concurrently) or the machine type in https://community-tc.services.mozilla.com/worker-manager/proj-servo%2Fdocker and send a PR to [`projects.yml` in `mozilla/community-tc-config`](https://github.com/mozilla/community-tc-config/blob/878acd84f0a9e7f34c56412603a155cae787fa8c/config/projects.yml#L214-L230) in order to make that change persistent (not overwritten next time someone runs `tc-admin` in that repository).

Other knobs at disposal are the number of chunks (= number of tasks), and the number of wptrunner processes per task.

With the current config the time for release build + WPT tasks on Linux is well under that of macOS, which remains the limiting factor for cycle time.

SimonSapin:
Per your previous comments:

@bors-servo r=jdm

PR #24750, opened 2019-11-15
Use `#![register_tool]` instead of `Registry::register_attribute` 

SimonSapin:
CC https://github.com/rust-lang/rust/pull/66344, https://github.com/rust-lang/rust/issues/66079

PR #24811, opened 2019-11-21
Fix updating the GitHub Status as soon as any TC task fails

SimonSapin:
… rather than only when the entire task group is resolved. This allows Homu to more quickly be notified of a failure, and move on to the next PR in the queue sooner.

(Plus drive-by Brewfile fix.)


jdm:
Something I'm not excited about is when a failure happens during the WPT tests and it causes any unscheduled tasks to fail when they become scheduled later because the git revision is no longer available. This makes it much more tedious to find the actual WPT failures, given that there are 30 mac WPT tasks that could fail.

jdm:
For example: https://community-tc.services.mozilla.com/tasks/groups/O8DjXNy-RfOWxqilil9Ycg

SimonSapin:
That’s a good point. I’ve added a few commits that try to solve this by making most tasks not use git at all. Instead they download a tarball of the repository’s contents from a (new) artifact of the decision task.

Note that landing this alone will likely break daily tasks until https://github.com/servo/taskcluster-config/pull/14 is deployed. (But deploying it without this PR landing likely also breaks them.)

@bors-servo try

jdm:
That's clever, and should also improve the situation where try builds can stomp on each other if a new one is launched before the build step for the previous one is complete.

SimonSapin:
6 jobs failed because `./mach test-tidy` and `./etc/ci/lockfile_changed.sh` run `git diff` without a git repository. I think the former can be fixed by passing `--all`. As to the latter, I believe it’s a design goal of Cargo that the lock file is the same far all possible target and feature selections, so maybe we only need to run it on a single platform?

Or we can use a git bundle instead of a tarball and still get a git repository without relying of ephemeral GitHub branches.

jdm:
Yes, we should choose one task for the lockfile check and remove the others.

SimonSapin:
I went wit the bundle, since we’ll still want at least one task that runs `lockfile_changed.sh` and therefore git.

@bors-servo try

SimonSapin:
Alright, this is ready to land but has changed quite a bit since last review. r? @jdm

jdm (approved):

jdm:
@bors-servo retry
* #23290

SimonSapin:
@bors-servo retry
```
  ▶ FAIL [expected PASS] /css/CSS2/margin-padding-clear/padding-bottom-007.xht
  └   → /css/CSS2/margin-padding-clear/padding-bottom-007.xht 54a9df64f1476dd12020019d7cf22ac34d727bc0
/css/CSS2/margin-padding-clear/margin-top-007-ref.xht 127fc2036bb8b5977576175ca1867d34e8c9bdd9
```

54a9df64f1476dd12020019d7cf22ac34d727bc0 matches https://build.servo.org/builders/linux-rel-wpt/builds/14235/steps/shell__4/logs/filtered-wpt-errorsummary.log and https://github.com/servo/servo/pull/24750#issuecomment-554561969, so it might be the hash of a blank (white) PNG.

https://github.com/servo/servo/issues/24726 ?

PR #24816, opened 2019-11-21
Require PromiseRejectionEventInit dictionary

### LINKED ISSUE #22670: Dictionaries with required members that need rooting break codegen
The design for dictionaries that contain rooted members is that an empty dictionary value is created and then each member is set individually. This doesn't work when the WebIDL type specifies `required`, like the `FormDataEvent`:
```
error[E0599]: no function or associated item named `empty` found for type `dom::bindings::codegen::Bindings::FormDataEventBinding::FormDataEventInit` in the current scope
   --> /Users/cybai/codespace/mozilla/servo/target/debug/build/script-59d4418593395066/out/Bindings/FormDataEventBinding.rs:857:95
    |
259 | pub struct FormDataEventInit {
    | ---------------------------- function or associated item `empty` not found for this
...
857 |             crate::dom::bindings::codegen::Bindings::FormDataEventBinding::FormDataEventInit::empty()
    |             ----------------------------------------------------------------------------------^^^^^
    |             |
    |             function or associated item not found in `dom::bindings::codegen::Bindings::FormDataEventBinding::FormDataEventInit`
```
Addressing this probably means figuring out the precise issue that the current design is addressing.

saschanaz:
---

- [x] These changes fix #22670

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -6425,7 +6425,8 @@ class CGDictionary(CGThing):
         mustRoot = ""
         if self.membersNeedTracing():
             mustRoot = "#[unrooted_must_root_lint::must_root]\n"
-            derive += ["Default"]
+            if not self.hasRequiredFields(self.dictionary):
+                derive += ["Default"]
 
         return (string.Template(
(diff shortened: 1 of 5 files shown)


jdm:
@bors-servo r+
Thanks!

PR #24817, opened 2019-11-21
Unindent Document::set_activity

jdm:
I was reading this method while reviewing some code and I find the early-return version easier to read.


jdm:
@bors-servo retry
#13480

PR #24668, opened 2019-11-05
Allow build to explicitly set the media stack

ghost:
_[This is based on @asajeffrey's work in #23423, and is a rebase and continuation of that branch]_

This allows `./mach build` to override the media stack, which is currently hard-wired based on the target. To skip gstreamer, run `./mach build -d --media-stack=dummy`.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #23050 (GitHub issue number if applicable)

- [x] These changes do not require tests because it's build infra

PR #24829, opened 2019-11-21
Build and publish Linux amd64 --media-stack=dummy

ghost:
- this depends on servo/servo#24668
- servo/download.servo.org#54 depends on this

---

- [x] These changes do not require tests because it's build infra

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -107,6 +107,7 @@ def tasks(task_for):
         daily_tasks_setup()
         with_rust_nightly()
         linux_nightly()
+        linux_dummy_nightly()
         android_nightly()
         windows_nightly()
         macos_nightly()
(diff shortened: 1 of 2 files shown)


Darkspirit:
What is the purpose of storing and publishing half-functional Servo Linux builds and paying costs for it?

jdm:
I don't see a compelling reason to publish nightly builds with media disabled. For the forseeable future this is only a build configuration to help local developers.

tuncer:
To be clear, were you waiting for me to lists arguments in favor of including it in nightly builds?

jdm:
As the person requesting this change, that would help us understand your use case better, yes.

tuncer:
Missing media playback support is a wanted feature for part of the user base. In fact, Firefox and Blink browsers, as distributed by their upstreams, work without media playback already. Further, building servo takes considerable time, and its rustc nightly version is changed regularly, requiring full rebuilds from time to time.

This means that trying out a fresh nightly build, testing, reporting issues, and contributing will be easier, because one doesn't have to build their personal media-dummy servo. Also, having nightlies means that this build profile will be tested more often than it would be otherwise, as it's readily available.

tuncer:
@jdm, what do you think? It wouldn't be ideal, but not linking to the artifact, and merely including it in the build matrix, would also suffice.

jdm:
Sorry, I don't think there's a clear value for making them easily available at this point. A better longer-term solution would be to focus on enabling the media library to load dynamically at runtime and allow the application to continue running without it.

tuncer:
That would work, too. Without arguing for a nightly build (artifact), do you think it's unnecessary to include it in CI to avoid bitrot?

tuncer:
Should a ticket be filed regarding media-stack graceful fallback at runtime?

jdm:
Yes, it would be useful to have an issue on file for making the media stack less statically coupled. And yes, I think adding a CI build that verifies that this configuration doesn't break is valuable.

tuncer:
@jdm see https://github.com/servo/servo/pull/25495

tuncer:
Also see https://github.com/servo/media/issues/329

PR #24841, opened 2019-11-22
Useful scripts for interacting with WPT logs

jdm:
These are scripts I used to generate the list in https://github.com/servo/servo/issues/24828, analyze test failures for #23290, and disable slow tests.

### DIFF
--- a/etc/wpt-summarize.py
+++ b/etc/wpt-summarize.py
@@ -13,7 +13,7 @@ import sys
 import json
 
 full_search = len(sys.argv) > 3 and sys.argv[3] == '--full'
-    
+
 with open(sys.argv[1]) as f:
     data = f.readlines()
     thread = None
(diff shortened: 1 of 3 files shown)


etc/wpt-summarize.py
```diff
@@ -0,0 +1,37 @@
...
+full_search = len(sys.argv) > 3 and sys.argv[3] == '--full'
+
+with open(sys.argv[1]) as f:
+    data = f.readlines()
+    thread = None
+    for entry in data:
+        entry = json.loads(entry)
+        if thread and "thread" in entry:
+            if entry["thread"] == thread:
+                print(json.dumps(entry))
```
Manishearth: won't this skip test_starts from other threads?

Manishearth (approved):
Could each script have a comment explaining what it does?

etc/wpt-summarize.py
jdm: I'm not sure I understand the concern here. The purpose of this script is to extract the output of a single test, so we don't care about other test_start events.

Manishearth: oh! i think i hadn't fully understood that part

jdm:
Added comments.

PR #24844, opened 2019-11-22
Debugging information for mac fonts

jdm:


mrobinson:
I think this can be closed now that CoreText fonts are always loaded via their paths. That said, once we get the Mac WPT running again we can re-apply this if we notice that fonts are still flaky.

Issue #22212, opened 2018-11-17
Windows taskcluster build intermittently crashes during mozjs configure (Couldn't reserve space for cygwin's heap, Win32 error 0)


jdm:
```
error: failed to run custom build command for `mozjs_sys v0.61.2`
process didn't exit successfully: `C:\Users\task_1542424298\repo\target\debug\build\mozjs_sys-cc0d89accd431be8\build-script-build` (exit code: 101)
--- stdout
[[ /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure -ot /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure.in ]] && touch /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure || true
[[ /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/old-configure -ot /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/old-configure.in ]] && touch /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/old-configure || true
! [[ /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure.in -ot /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure ]] && touch /c/Users/task_1542424298/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.2/mozjs/js/src/configure || true
[...]

PR #24858, opened 2019-11-25
Handle u32 property indices

### LINKED ISSUE #14093: Properly handle array property name indices
We fail `/dom/collections/HTMLCollection-supported-property-indices.html` because `get_array_index_from_id` fails to see that 2147483648 and 4294967293 are valid property name indices, and thus as per the [WebIDL spec](https://heycam.github.io/webidl/#getownproperty-guts) (see `ignoreNamedProps` in that algorithm) if the indexed getter returns nothing, the named getter shouldn't even be considered. This seems to be a problem with the `Id` type being internally an `i32` for numerical indices.

```
± ./mach test-wpt --no-pause /dom/collections/HTMLCollection-supported-property-indices.html
Running 1 tests in web-platform-tests

  ▶ Unexpected subtest result in /dom/collections/HTMLCollection-supported-property-indices.html:
  │ FAIL [expected PASS] Handling of property names that look like integers around 2^32
  │   → assert_equals: expected (undefined) undefined but got (object) Element node <foo id="4294967293"></foo>
  │
  │ @http://web-platform.test:8000/dom/collections/HTMLCollection-supported-property-indices.html:63:3
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1401:20
  │ test@http://web-platform.test:8000/resources/testharness.js:501:9
  └ @http://web-platform.test:8000/dom/collections/HTMLCollection-supported-property-indices.html:46:1

  ▶ Unexpected subtest result in /dom/collections/HTMLCollection-supported-property-indices.html:
  │ FAIL [expected PASS] Handling of property names that look like integers around 2^31
  │   → assert_equals: expected (undefined) undefined but got (object) Element node <foo id="2147483648"></foo>
  │
  │ @http://web-platform.test:8000/dom/collections/HTMLCollection-supported-property-indices.html:42:3
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1401:20
  │ test@http://web-platform.test:8000/resources/testharness.js:501:9
  └ @http://web-platform.test:8000/dom/collections/HTMLCollection-supported-property-indices.html:22:1

Ran 1 tests finished in 1.0 seconds.
  • 0 ran as expected. 0 tests skipped.
  • 1 tests had unexpected subtest results
```

saschanaz:
Tried porting from Gecko, not sure how to do void JSID check yet.

---

- [x] These changes fix #14093

### DIFF
--- a/components/script/dom/bindings/utils.rs
+++ b/components/script/dom/bindings/utils.rs
@@ -16,11 +16,13 @@ use crate::dom::bindings::trace::trace_object;
 use crate::dom::messageport::MessagePort;
 use crate::dom::windowproxy;
 use crate::script_runtime::JSContext as SafeJSContext;
-use js::conversions::ToJSValConvertible;
+use js::conversions::{jsstr_to_string, ToJSValConvertible};
 use js::glue::{CallJitGetterOp, CallJitMethodOp, CallJitSetterOp, IsWrapper};
 use js::glue::{GetCrossCompartmentWrapper, JS_GetReservedSlot, WrapperNew};
 use js::glue::{UnwrapObjectDynamic, RUST_JSID_TO_INT, RUST_JSID_TO_STRING};
-use js::glue::{RUST_FUNCTION_VALUE_TO_JITINFO, RUST_JSID_IS_INT, RUST_JSID_IS_STRING};
+use js::glue::{
+    RUST_FUNCTION_VALUE_TO_JITINFO, RUST_JSID_IS_INT, RUST_JSID_IS_STRING, RUST_JSID_IS_VOID,
+};
 use js::jsapi::HandleId as RawHandleId;
 use js::jsapi::HandleObject as RawHandleObject;
 use js::jsapi::MutableHandleObject as RawMutableHandleObject;
@@ -29,7 +31,10 @@ use js::jsapi::{Heap, JSAutoRealm, JSContext, JS_FreezeObject};
 use js::jsapi::{JSJitInfo, JSObject, JSTracer, JSWrapObjectCallbacks};
 use js::jsapi::{JS_EnumerateStandardClasses, JS_GetLatin1StringCharsAndLength};
 use js::jsapi::{JS_IsExceptionPending, JS_IsGlobalObject};
-use js::jsapi::{JS_ResolveStandardClass, JS_StringHasLatin1Chars, ObjectOpResult};
+use js::jsapi::{
+    JS_ResolveStandardClass, JS_StringHasLatin1Chars, ObjectOpResult, StringIsArrayIndex1,
+    StringIsArrayIndex2,
+};
 use js::jsval::{JSVal, UndefinedValue};
 use js::rust::wrappers::JS_DeletePropertyById;
 use js::rust::wrappers::JS_ForwardGetPropertyTo;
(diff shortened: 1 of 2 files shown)


components/script/dom/bindings/utils.rs
```diff
@@ -178,28 +181,46 @@ pub unsafe fn get_property_on_prototype(
...
-    //         return -1;
-    //     }
-    //
-    //     let i = 0;
-    //     let str = AtomToLinearString(JSID_TO_ATOM(id));
-    //     return if StringIsArray(str, &mut i) != 0 { i } else { -1 }
-    // } else {
-    //     IdToInt32(cx, id);
+
+    // if !RUST_JSID_IS_VOID(raw_id) {
```
jdm: In the short term this could be `if raw_id.asBits == JSID_VOID.asBits` (based on https://doc.servo.org/mozjs_sys/jsid/constant.JSID_VOID.html). We can also add a RUST_JSID_IS_VOID API to https://github.com/servo/rust-mozjs/blob/master/src/glue.rs and https://github.com/servo/rust-mozjs/blob/master/src/jsglue.cpp.

components/script/dom/bindings/utils.rs
```diff
@@ -178,28 +181,46 @@ pub unsafe fn get_property_on_prototype(
...
+
+    let first = s.chars().next().unwrap();
+    if first.is_ascii_lowercase() {
+        return None;
+    }
+
+    let mut i: u32 = 0;
+    let is_array = if s.is_ascii() {
+        let chars = s.as_bytes();
+        StringIsArrayIndex1(chars.as_ptr() as *const i8, chars.len() as u32, &mut i)
```
jdm: c_char is a different type on ARM, so we need `as *const _` instead of hardcoding i8.

jdm (changes requested):

saschanaz:
Ooops, last time important bit 🤯

jdm:
@bors-servo retry
* #22212

saschanaz:
@bors-servo retry

Lots of intermittents, how can I see filtered log in the new CI?

jdm:
@bors-servo retry
* #22276

saschanaz:
@bors-servo retry

(Starting to feel unlucky)

jdm:
I will retry this when our CI has caught up with the existing build queue. Please do not retry in the meantime.

saschanaz:
Starting to worry a bit, but didn't find any evidence that this PR caused those failures.

jdm:
This PR is not the cause - we have several high frequency intermittent failures right now across all our CI platforms, and every time we have to do a full CI run it's very likely that we hit them. Retrying while there are builds in progress means we have to re-run all of the in-progress builds; whereas when we wait until the builds are complete, we only re-run the ones that failed.

jdm:
https://treeherder.allizom.org/#/jobs?repo=servo-auto shows the view that I am looking at to determine when it's a good time to retry.

saschanaz:
>we have several high frequency intermittent failures right now across all our CI platforms

Sounds like a good reason to filter them, what's the reason not to do so?

jdm:
We filter based on test filename, and these are failures that can (and do) affect any test that runs.

jdm:
@bors-servo retry
* #24726

PR #24860, opened 2019-11-25
Linux reftest investigations

jdm:
Investigations for #24726.


jdm:
Reproduced the reftest failure and didn't see any difference in viewport behaviour between the good and bad renderings.

jdm:
Shouldn't be necessary with #25566.

PR #24863, opened 2019-11-25
The gstreamer plugin should resize when it's presented with frames to fill

### LINKED ISSUE #24835: GStreamer plugin resizing
For some reason when the gstreamer plugin gets resized, the page isn't reflowed so any GL content that dynamically sizes its canvas is still using the old size.

asajeffrey:
This is a workaround for race conditions in the embedder.

---

- [x] These changes fix #24835
- [x] These changes do not require tests because it's fixing a bug with an embedding

### DIFF
--- a/ports/gstplugin/servosrc.rs
+++ b/ports/gstplugin/servosrc.rs
@@ -485,11 +485,6 @@ impl BaseSrcImpl for ServoSrc {
     fn set_caps(&self, _src: &BaseSrc, outcaps: &Caps) -> Result<(), LoggableError> {
         let info = VideoInfo::from_caps(outcaps)
             .ok_or_else(|| gst_loggable_error!(CATEGORY, "Failed to get video info"))?;
-        let size = Size2D::new(info.width(), info.height()).to_i32();
-        debug!("Setting servosrc buffer size to {}", size,);
-        self.sender
-            .send(ServoSrcMsg::Resize(size))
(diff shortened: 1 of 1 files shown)


asajeffrey:
OK, I think what's going on is this...

The embedder API uses push for some things (`WindowEvent`s being pushed from the embedder to servo) and pull for others (the `WindowMethods` trait implemented by the embedder), with the result that it's easy for things to get out of sync. In this case, it was the `get_coordinates` method, I think what was happening was:

- Embedder gets a `Start` message, and sends a `NewBrowser` event to Servo
- Embedder gets a `Resize` message, and resizes itself, then sends a `Resize` event to Servo
- Servo (still processing `NewBrowser`) calls `get_coordinates` and gets the new size
- When Servo processes `Resize`, there's nothing to do, since the size hasn't changed, and as a result script is never informed of the new size.

This PR is a workaround for that, by delaying sending the `Resize`.

A better fix is to revisit the embedding API to make it all push rather than mixed push/pull (or all async).

@paulrouget does my analysis sound right?

@ferjm r?

paulrouget:
I know nothing about gstplugin. But

> When Servo processes Resize, there's nothing to do, since the size hasn't changed, and as a result script is never informed of the new size.

Right. But in the first place script should have had the new size.

> A better fix is to revisit the embedding API to make it all push rather than mixed push/pull (or all async).

One day :)

ferjm (approved):

ferjm:
@bors-servo retry
- #22212

Issue #2909, opened 2014-07-23
Parse data URLs (and their MIME types) according to the Fetch Standard


SimonSapin:
Spec is at https://fetch.spec.whatwg.org/#data-urls

Status as of November 2019: https://crates.io/crates/data-url has an initial implementation, but integration with Servo is blocked on figuring out the interactions with https://crates.io/crates/mime. CC https://github.com/hyperium/mime/issues/78, https://github.com/servo/rust-url/issues/567.

----

**Initial title:** Figure out the details of data URL parsing

**Initial description:**

For example, what should happen when a base64 sequence contains a (0x20 ASCII) space? Another non-alphabet character?

Since [RFC 2397](http://tools.ietf.org/html/rfc2397) does no define error handling, this might involve reverse-engineering other implementations.

https://www.w3.org/Bugs/Public/show_bug.cgi?id=19494
http://greenbytes.de/tech/tc/datauri/

frewsxcv:
Should this be moved to rust-url?

SimonSapin:
The corresponding code is currently in `components/net/data_loader.rs` in this repository, not in rust-url. I might move that code to rust-url once all the details are figured out and I’m confident to tell the world "this is how you do it" (i.e. when this issue can be closed) but in the mean time I’m not planning to.

nox:
The spec bug trail brought me to https://github.com/whatwg/fetch/issues/234.

pshaughn:
https://infra.spec.whatwg.org/#forgiving-base64-decode might be the thing to do here?

jdm:
Indeed, step 11.2 of https://fetch.spec.whatwg.org/#data-urls uses the forgiving base64 decode algorithm, so that looks like a change we will need to make in our implementation.

pshaughn:
I am doing that now; we're already forgiving spaces, just need to add forgiveness of a few other kinds of byte.

pshaughn:
There is one remaining problem: "ab=" is invalid base64, but we're responding to this particular invalid base64 in a different way from all the other invalid base64. I don't know where this "Reached unreachable code" string is even being generated yet; it doesn't seem to be a string in Servo or in the base64 crate.

`  ▶ Unexpected subtest result in /fetch/data-urls/base64.any.worker.html:
  │ FAIL [expected PASS] data: URL base64 handling: "ab="
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code

[...]

PR #24865, opened 2019-11-25
Correct handling of whitespace in base64 URLs

pshaughn:
Whitespace handling in base64 url parsing now treats tabs, form-feeds, and linefeeds the same way it treats regular spaces

---

- [X] These changes fix (part of) #2909

### DIFF
--- a/components/net/data_loader.rs
+++ b/components/net/data_loader.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use base64;
+use data_url::forgiving_base64;
 use mime::Mime;
 use percent_encoding::percent_decode;
 use servo_url::ServoUrl;
@@ -16,6 +16,10 @@ pub enum DecodeError {
 pub type DecodeData = (Mime, Vec<u8>);
 
 pub fn decode(url: &ServoUrl) -> Result<DecodeData, DecodeError> {
+    // data_url could do all of this work for us,
+    // except that it currently (Nov 2019) parses mime types into a
+    // different Mime class than other code expects
+
     assert_eq!(url.scheme(), "data");
     // Split out content type and data.
     let parts: Vec<&str> = url[Position::BeforePath..Position::AfterQuery]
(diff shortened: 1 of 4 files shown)


components/net/data_loader.rs
```diff
@@ -44,11 +44,13 @@ pub fn decode(url: &ServoUrl) -> Result<DecodeData, DecodeError> {
...
-        // FIXME(#2909): It’s unclear what to do with non-alphabet characters,
-        // but Acid 3 apparently depends on spaces being ignored.
+        // FIXME(#2909):
+        // infra.spec.whatwg.org/#forgiving-base64-decode does a pretty good
+        // job of telling us what to do here. Most WPT tests are working;
+        // the base64 string "ab=" in particular is not.
         bytes = bytes
             .into_iter()
-            .filter(|&b| b != b' ')
+            .filter(|&b| b != b' ' && b != b'\x0C' && b != b'\n' && b != b'\r' && b != b'\t')
```
jdm: While it's logical to extend the existing code here, we should honour SimonSapin's request in https://github.com/servo/servo/issues/2909#issuecomment-558328602 to make use of the [data-url crate](https://docs.rs/data-url/0.1.0/data_url/) and replace all of this decoding code if possible.

We should be able to use `DataUrl::process(url.as_str())` and return the value of decode_to_vec() along with the url's mime_type() value. We'll need to add data-url as a new dependency in components/net/Cargo.toml as well.

jdm (changes requested):
Good find on the data url decoding problems!

components/net/data_loader.rs
pshaughn: Will do! I've never added a Cargo dependency before, but I'll have to sooner or later and I may as well start with this one!

pshaughn:
Now using data-url's version of forgiving base64 decoding, but still doing the rest of the MIME parse within data_loader.rs, as suggested in https://github.com/servo/servo/issues/2909#issuecomment-558537093. This is a full fix for #2909 as far as WPT tests are concerned, but integrating with the full data-url class eventually would be nice.

SimonSapin:
Could you squash your commits into one? Assuming https://github.com/servo/servo is your `origin` remote, that would be something like `git fetch origin && git rebase -i origin/master`. Or I can do it for you prefer.

pshaughn:
@SimonSapin : Is there a way to do that within this pull request, or should I make a new pull request with the merged commit?

pshaughn:
#24870 is the commit-squashed version of this, sorry for the mess

Issue #24706, opened 2019-11-11
Implement WebGPU in Servo


zakorgy:
One of the goals of the University of Szeged team in this year is to implement the WebGPU API (https://gpuweb.github.io/gpuweb/) in Servo.
This includes the following tasks:
- [ ] Set up a background thread where we call the WebGPU functions (WebGPU thread). Here we would use [wgpu-native](https://github.com/gfx-rs/wgpu) crate.
- [ ] Establish the asynchronous communication between this background thread and the WebGPU DOM objects in the script thread.
- [ ] Implement the WebGPU WebIDL bindings in `script\dom`, including but not limited to:
    - [ ] Implement the `GPU` interface
    - [ ] Implement `requestAdapater` and `GPUAdapter` interface
    - [ ] Implement `requestDevice` and `GPUDevice` interface
    - [ ] Implement resource creation (textures, buffers, samplers)
    - [ ] Implement pipeline creation (bind groups, pipeline layouts, shader modules, computer/render pipelines)
    - [ ] Implement command interfaces (command encoders, queues)
    - [ ] implement fences
    - [ ] Implement compute functionalities
    - [ ] Implement graphic functionalities (draw calls, swap chain)
- [ ] Pass the WebGPU CTS.

cc @jdm, @kvark

zakorgy:
Some high-level design suggestions from @gterzian from https://github.com/servo/servo/pull/24708#discussion_r346662690: 
- [ ] I wonder if you really want to always start the webgpu component on start-up, rather you might want to start-it when it's first needed.
- [ ] I wonder if you want any script to just be able to start using WebGPU whenever they want by calling methods on GPU inside script. Instead I think you want to mediate that capability via the constellation. So when script calls RequestAdapter, it should ideally go through the constellation via a ScriptMsg::RequestAdapter(sender, options), and be treated as an untrusted request for a specific capability. It ties into 1 in the sense that the constellation could then start a WebGPU instance, and respond to script with a dedicated means of communication with it(so subsequent messages don't all have to go through the constellation).
[...]

PR #24883, opened 2019-11-27
Add WebGPU identity management

imiklos:
---

cc @jdm, @kvark, @zakorgy


components/script/dom/navigator.rs
```diff
@@ -38,6 +41,8 @@ pub struct Navigator {
     permissions: MutNullableDom<Permissions>,
     mediasession: MutNullableDom<MediaSession>,
     gpu: MutNullableDom<GPU>,
+    #[ignore_malloc_size_of = "Defined in wgpu"]
+    gpu_id_hub: RefCell<Option<Identities>>,
```
kvark: I don't think we need an `Option` here. Each identity manager will be empty if not used, anyway.

imiklos: Thank you! I've updated the pull request with this change.

Issue #23766, opened 2019-07-12
Add a nightly job that tries to upgrade webrender


jdm:
Similar to the job that performs the upstream WPT sync every night, it would be nice to have a job that tries to pull down the latest webrender changes, build them, and submit a pull request. This would let us catch new test failures and API changes as they are merged upstream.

atouchet:
Could Dependabot be used for this?

jdm:
It would possibly happen automatically once we get out from the backlog of out of date dependencies.

atouchet:
Is there a limit to how many PRs Dependabot will generate per day? I've noticed that it doesn't do every single dependency all at once.

jdm:
There is a built in limit, yes.

atouchet:
Is this still desired now that Servo is using WebRender from crates.io instead of a Git version?

mrobinson:
I suspect that this is not something we'd like to do as we will always wanting to be using the version from crates.io. We should increase the velocity of WebRender rebases and upgrades though.

PR #24889, opened 2019-11-27
Add nightly task to build and run tests with latest webrender.

jdm:
Like the nightly Rust compiler upgrade task, this will help us learn sooner when Webrender introduces breaking API changes or causes regressions in our testsuite.

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -113,6 +113,7 @@ def tasks(task_for):
         update_wpt()
         magicleap_nightly()
         uwp_nightly()
+        webrender_nightly()
 
 
 # These are disabled in a "real" decision task,
@@ -529,6 +530,10 @@ def windows_nightly():
     )
 
 
+def webrender_nightly():
+    return linux_wpt_common(total_chunks=4, layout_2020=False, webrender_nightly=True)
+
+    
 def linux_nightly():
     return (
         linux_build_task("Nightly build and upload")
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
r+

etc/taskcluster/decision_task.py
```diff
@@ -529,6 +530,10 @@ def windows_nightly():
     )
 
 
+def webrender_nightly():
+    return linux_wpt_common(total_chunks=4, layout_2020=False, webrender_nightly=True)
```
SimonSapin: Nit: I’d prefer to keep the calls to `linux_wpt_common` closer to each other, so we see them all when changing for example the number of chunks.

atouchet:
Can this be used to submit a pull request like #23766 suggests?

SimonSapin (changes requested):
I get the desire for something like this PR, but I’m not a fan of:

* Adding any more daily tasks that are expected to sometimes fail without a `servo/servo` PR breaking them
* Adding a known source of intermittent failures (WPT) to daily tasks

… as long as the only feedback we have from them is IRC notifications to a few people.

I tend to take care of compiler updates, but I’m not personally planning to do the same for WebRender. Yet with this PR I’ll get pinged *every day* when WebRender’s master doesn’t build in Servo or has any impact on WPT, until Servo is updated.

SimonSapin:
Possible mitigation: replace this:

https://github.com/servo/servo/blob/a020522f37241001b01ee77dab562d50b6e7a924/etc/taskcluster/decision_task.py#L128

With some per-task configuration where someone is pinged if they personally opt-in.

jdm:
I'll make the pinging configurable like you suggest.

jdm:
@atouchet That's more effort than I'm willing to spend on this task right now, given the complexities involved in https://github.com/servo/servo/blob/master/etc/ci/update-wpt-checkout.

mrobinson:
Going to close this PR since it's against the old taskcluster config.

PR #24928, opened 2019-11-29
Start supporting images in layout 2020

nox:


components/layout_2020/flow/mod.rs
```diff
@@ -479,3 +470,104 @@ fn layout_in_flow_non_replaced_block_level<'a>(
...
+    padding_border_inline_sum: Length,
+    computed_margin_inline_start: LengthOrAuto,
+    computed_margin_inline_end: LengthOrAuto,
+    inline_size: Length,
+) -> (Length, Length) {
+    let inline_margins = containing_block.inline_size - padding_border_inline_sum - inline_size;
+    match (computed_margin_inline_start, computed_margin_inline_end) {
+        (LengthOrAuto::Auto, LengthOrAuto::Auto) => (inline_margins / 2., inline_margins / 2.),
+        (LengthOrAuto::Auto, LengthOrAuto::LengthPercentage(end)) => (inline_margins - end, end),
+        (LengthOrAuto::LengthPercentage(start), _) => (start, inline_margins - start),
```
SimonSapin: Nit: This `_` can only be `Auto`, right? Could we make it a specific pattern?

components/layout_2020/replaced.rs
```diff
@@ -2,20 +2,77 @@
...
+
+    pub fn layout<'a>(
+        &'a self,
+        layout_context: &LayoutContext,
+        style: &ServoArc<ComputedValues>,
+        containing_block: &ContainingBlock,
+    ) -> IndependentLayout {
+        let (fragments, content_block_size) = match self.kind {
+            ReplacedContentKind::Image(ref image) => {
+                // FIXME(nox): We should not assume block size is known.
```
SimonSapin: Or, on the contrary, isn’t it always known for replaced boxes and should we take `DefiniteContainingBlock` here instead?

SimonSapin (approved):

components/layout_2020/flow/mod.rs
nox: It can be `LengthOrAuto::LengthPercentage` too, which we ignore because that's over-constraining.

components/layout_2020/replaced.rs
nox: You are right, at that point we should know both sizes, I'll make it use `DefiniteContainingBlock`.

SimonSapin: Alright, r=me with that.

nox: It turns out we can't do that: this is also called from `IndependentFormattingContext::layout`, where we don't know for sure the block size.

nox: Tbh that makes me wonder if we shouldn't separate replaced content from independent formatting contexts.

SimonSapin: Yeah I considered that when reading your previous message. Or if not separate data types, at least remove `IndependentFormattingContext::layout` and make parent boxes call `IndependentFormattingContext::as_replaced` themselves.

SimonSapin: That doesn’t need to happen in this PR though. I’ll r+

SimonSapin:
One failure in layout 2020 mode, could be legit?
```
  ▶ FAIL [expected CRASH] /css/CSS2/box-display/containing-block-007.xht
  └   → /css/CSS2/box-display/containing-block-007.xht 8ad90a71ec464d4cd92c5f8f5343f88d4e6f4902
/css/CSS2/box-display/containing-block-007-ref.xht 46b4d6b0ef99e6a3b5982147134474a332c2e7a1
```

SimonSapin:
The fixed crash is due to removing on `unimplemented!`. I’ve updated WPT expectations.

@bors-servo r+

Issue #24715, opened 2019-11-12
Intermittent android build failure during fontconfig


jdm:
```
make[3]: Entering directory '/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out/src'
  GEN      fcalias.h
  GEN      stamp-fcstdint
  GEN      fcftalias.h
  GEN      fcobjshash.gperf
  GEN      fcobjshash.h
Makefile:882: recipe for target 'fcobjshash.h' failed
config.status: executing src/fcstdint.h commands
config.status: creating src/fcstdint.h : _FONTCONFIG_SRC_FCSTDINT_H
config.status: src/fcstdint.h is unchanged
make[3]: Leaving directory '/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out/src'
Makefile:562: recipe for target 'all-recursive' failed
make[2]: Leaving directory '/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out'
Makefile:445: recipe for target 'all' failed
make[1]: Leaving directory '/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out'
makefile.cargo:61: recipe for target '/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out/libfontconfig.a' failed

--- stderr
configure: WARNING: using cross tools not prefixed with host triplet
make[1]: warning: -jN forced in submake: disabling jobserver mode.
/root/.cargo/registry/src/github.com-1ecc6299db9ec823/servo-fontconfig-sys-4.0.7/missing: line 81: gperf: command not found
WARNING: 'gperf' is missing on your system.
         You might have modified some files without having the proper
         tools for further handling them.  Check the 'README' file, it
         often tells you about the needed prerequisites for installing
         this package.  You may also peek at any GNU archive site, in
         case some other package contains this missing 'gperf' program.
make[3]: *** [fcobjshash.h] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [all-recursive] Error 1
make[1]: *** [all] Error 2
make: *** [/repo/target/android/i686-linux-android/release/build/servo-fontconfig-sys-51d7b83465512d8a/out/libfontconfig.a] Error 2
thread 'main' panicked at 'assertion failed: Command::new("make").env("MAKEFLAGS",
[...]

PR #24929, opened 2019-11-29
Extend WPT update job timeout.

jdm:
https://community-tc.services.mozilla.com/tasks/HIGiBAoHQQ-iiw0q-oYX0Q was a job on a machine with no stale processes and no clear problems running the tests. It does include an upstream change that causes a bunch of CSS interpolation tests to start running, so I think we're just hitting the timeout naturally again.


SimonSapin:
@bors-servo r+

I think it would be good to separately look (again?) into moving this to Linux, where capacity is less limited and where we can have more CPU cores.

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24715

jdm:
@bors-servo retry
* #24715

PR #24977, opened 2019-11-30
Add StreamConsumer wrapper and methods to response

### LINKED ISSUE #21476: Support asynchronous WebAssembly compilation

ridhimrastogi:
Add Streamconsumer wrapper to Response

---

- [x] These changes fix #21476

### DIFF
--- a/components/script/dom/response.rs
+++ b/components/script/dom/response.rs
@@ -19,6 +19,7 @@ use crate::dom::headers::{is_obs_text, is_vchar};
 use crate::dom::headers::{Guard, Headers};
 use crate::dom::promise::Promise;
 use crate::dom::xmlhttprequest::Extractable;
+use crate::script_runtime::StreamConsumer;
 use dom_struct::dom_struct;
 use http::header::HeaderMap as HyperHeaders;
 use hyper::StatusCode;
@@ -48,6 +49,8 @@ pub struct Response {
     body: DomRefCell<NetTraitsResponseBody>,
     #[ignore_malloc_size_of = "Rc"]
     body_promise: DomRefCell<Option<(Rc<Promise>, BodyType)>>,
+    #[ignore_malloc_size_of = "StreamConsumer"]
+    stream_consumer: DomRefCell<Option<StreamConsumer>>,
 }
 
 impl Response {
@@ -64,6 +67,7 @@ impl Response {
             url_list: DomRefCell::new(vec![]),
             body: DomRefCell::new(NetTraitsResponseBody::Empty),
             body_promise: DomRefCell::new(None),
+            stream_consumer: DomRefCell::new(None),
         }
     }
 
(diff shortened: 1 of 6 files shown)


ridhimrastogi:
The commit fails as the newly added functions, set_stream_consumer and stream_chunk aren't being called yet.

components/script/dom/response.rs
```diff
@@ -441,11 +445,24 @@ impl Response {
         }
     }
 
+    pub fn set_stream_consumer(&self, sc: Option<StreamConsumer>) {
+        *self.stream_consumer.borrow_mut() = sc;
+    }
+
+    pub fn stream_chunk(&self, stream: &[u8]) {
+        if let Some(stream_consumer) = self.stream_consumer.borrow_mut().take() {
```
jdm: take() consumes the value, so we want  as_ref() instead.

components/net/tests/fetch.rs
```diff
@@ -138,6 +138,7 @@ fn test_fetch_blob() {
         fn process_response(&mut self, _: &Response) {}
         fn process_response_chunk(&mut self, chunk: Vec<u8>) {
             self.buffer.extend_from_slice(chunk.as_slice());
+            self.stream_chunk(buffer.as_slice());
```
jdm: This is a unit test for unrelated code; we should be modifying components/script/fetch.rs instead.

jdm (changes requested):
Good start!

components/script/dom/response.rs
ridhimrastogi: Done

components/net/tests/fetch.rs
ridhimrastogi: Oops, had both files open and made changes in the wrong one. Fixed it now

components/script/fetch.rs
```diff
@@ -261,6 +261,7 @@ impl FetchResponseListener for FetchContext {
 
     fn process_response_chunk(&mut self, mut chunk: Vec<u8>) {
         self.body.append(&mut chunk);
+        self.stream_chunk(chunk.as_slice());
```
jdm: Since this is a method on FetchContext and not Response, you need to use `aelf.respinse_object.root()` to obtain a `DomRoom<Response>` object that you can use to call Response methods.

jdm:
I think the only missing piece now is calling Response::set_stream_consumer from the stream consumer callback?

components/script/fetch.rs
ridhimrastogi: Done

ridhimrastogi:
> I think the only missing piece now is calling Response::set_stream_consumer from the stream consumer callback?

Done

ridhimrastogi:
@jdm I see some changes in the test results when I run  
`./mach test-wpt tests/wpt/web-platform-tests/wasm/webapi/`.
Do I need to update test expectations for these changes?

jdm:
Yep!

components/script/fetch.rs
```diff
@@ -261,6 +261,7 @@ impl FetchResponseListener for FetchContext {
 
     fn process_response_chunk(&mut self, mut chunk: Vec<u8>) {
         self.body.append(&mut chunk);
+        self.response_object.root().stream_chunk(chunk.as_slice());
```
jdm: It turns out that `self.body.append(&mut chunk);` empties the passed vector, which is why a bunch of tests that I would expect to pass are failing with these changes. If we reverse the order of these lines, then the `compileStreaming` tests from /wasm/webapi/contenttype.any.html (and likely other test files as well) should start passing.

jdm (changes requested):
Looks great! With one final change this should be ready to merge :)

components/script/fetch.rs
ridhimrastogi: I have made the changes. Now the mimeType related tests are passing. However, some of the response status tests in ```/tests/wpt/metadata/wasm/webapi/status.any.js.ini``` are failing, which I believe should have passed. Is this expected?

jdm:
I'm fine with merging the changes as-is.
@bors-servo r+

CYBAI:
on both linux and macOS builds:

```
  ▶ ERROR [expected TIMEOUT] /wasm/serialization/module/identity-not-preserved.html
  └   → Unhandled rejection: The object can not be cloned.
```

jdm:
That's #24726; I'll retry when the other pending tasks finish running.

jdm:
@bors-servo retry
* #24726

PR #25025, opened 2019-12-03
Origin offset fixes

Manishearth:
https://github.com/immersive-web/webxr/issues/567 was closed out. We were computing offset spaces of already-offset spaces incorrectly, but otherwise our math is correct. I improved our comments around this with more math, so I never have to do this math again.

Chrome's math isn't, which is why we fail some tests around this: https://bugs.chromium.org/p/chromium/issues/detail?id=1030049 . I'm planning to wait for them to fix and upstream the tests, I've already verified that we pass the corrected test.

r? @jdm


Manishearth:
```
0 matches
{
    "status": "FAIL", 
    "group": "default", 
    "message": "promise_test: Unhandled rejection with value: object \"TypeError: Network error occurred\"", 
    "stack": null, 
    "subtest": "idl_test setup", 
    "test": "/html/dom/idlharness.https.html?exclude=(Document|Window|HTML.*)", 
    "line": 31367, 
    "action": "test_result", 
    "expected": "PASS"
}
```

@bors-servo retry

Manishearth:
With https://github.com/servo/servo/pull/25086 synced in we now pass xrFrame_getPose

Issue #21067, opened 2018-06-18
Intermittent panic: "Failed to receive a response from live font cache" from get_font_instance (`/encoding/legacy-mb-korean/euc-kr/euckr-decode-windows-949.html?11001-12000`, `/encoding/legacy-mb-japanese/euc-jp/eucjp-decode-cseucpkdfmtjapanese.html?3001-4000`)


jdm:
```
  ▶ TIMEOUT [expected PASS] /css/CSS2/css1/c43-rpl-ibx-000.xht
  │ 
  │ VMware, Inc.
  │ softpipe
  │ 3.3 (Core Profile) Mesa 18.1.0-devel
  │ Failed to receive a response from live font cache (thread LayoutThread PipelineId { namespace_id: PipelineNamespaceId(0), index: PipelineIndex(1) }, at components/gfx/font_cache_thread.rs:470)
  │ stack backtrace:
  │    0:     0x7fa9f7cd94bc - backtrace::backtrace::trace::h771ec7e4a16b9188
  │    1:     0x7fa9f7cd8c62 - <backtrace::capture::Backtrace as core::default::Default>::default::ha7872455e6e477e3
  │    2:     0x7fa9f7cd8ca8 - backtrace::capture::Backtrace::new::h7a84dd53000abc29
  │    3:     0x7fa9f522e1a8 - servo::main::{{closure}}::h0e01fc13ea220a38
  │    4:     0x7fa9f7d02323 - std::panicking::rust_panic_with_hook::h7a007654e7427bd4
  │                         at libstd/panicking.rs:515
  │    5:     0x7fa9f679e476 - std::panicking::begin_panic::h3a6a8c73cf6210f7
  │    6:     0x7fa9f67bf00b - <gfx::font_cache_thread::FontCacheThread as gfx::font_context::FontSource>::get_font_instance::h8b87686ea555d885
  │    7:     0x7fa9f64ab79f - <gfx::font_context::FontContext<S>>::font::h0f315736397a85be
  │    8:     0x7fa9f649ff72 - <core::iter::FilterMap<I, F> as core::iter::iterator::Iterator>::try_fold::{{closure}}::h0b065e2497c46a57
  │    9:     0x7fa9f64a6374 - layout::text::TextRunScanner::scan_for_runs::h86641f021b472292
  │   10:     0x7fa9f5684245 - <std::thread::local::LocalKey<T>>::with::he22b905e40a1e8d3
[...]

PR #25029, opened 2019-12-03
Initial implementation of GPUDevice for WebGPU

imiklos:
Added the WebIDL bindigs for GPUDevice, GPUObjectDescriptorBase, GPUDeviceDescriptor, GPUObjectBase
Implemented the `requestDevice` function of `GPUAdapter`

---

- [X]  These changes addresses a part of #24706

cc @jdm, @kvark, @zakorgy


components/webgpu/lib.rs
```diff
@@ -77,6 +83,7 @@ struct WGPU {
     adapters: Vec<WebGPUAdapter>,
     // Track invalid adapters https://gpuweb.github.io/gpuweb/#invalid
     _invalid_adapters: Vec<WebGPUAdapter>,
+    device_to_adapter: HashMap<WebGPUDevice, WebGPUAdapter>,
```
kvark: do we need this map?
it seems like we should just carry a JS reference to `WebGPUAdapter` from the device struct

kvark (changes requested):

components/webgpu/lib.rs
imiklos: We don't use it right now. I will remove this map. The reference to the `Adapter` is already in place. I will update the pull request with this change. Thank you!

kvark (approved):

components/script/dom/gpudevice.rs
```diff
@@ -0,0 +1,91 @@
...
+    #[ignore_malloc_size_of = "mozjs"]
+    extensions: Heap<*mut JSObject>,
+    #[ignore_malloc_size_of = "mozjs"]
+    limits: Heap<*mut JSObject>,
+    label: DomRefCell<Option<DOMString>>,
+    device: WebGPUDevice,
+}
+
+impl GPUDevice {
+    pub fn new_inherited(
```
jdm: nit: no need for pub.

components/script/dom/gpuadapter.rs
```diff
@@ -60,4 +71,47 @@ impl GPUAdapterMethods for GPUAdapter {
...
+
+impl AsyncWGPUListener for GPUAdapter {
+    fn handle_response(&self, response: WebGPUResponse, promise: &Rc<Promise>) {
+        match response {
+            WebGPUResponse::RequestDevice(device_id, _descriptor) => {
+                let device = GPUDevice::new(
+                    &self.global(),
+                    &self,
+                    Heap::default(),
+                    Heap::default(),
```
jdm: Will we create non-default values based on descriptor in the future?

components/script/dom/webidls/GPUDevice.webidl
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// https://gpuweb.github.io/gpuweb/#gpudevice
+[Exposed=(Window, DedicatedWorker)/*, Serializable */]
```
jdm: This should be `Pref="dom.webgpu.enabled"`.

jdm (changes requested):

components/script/dom/gpuadapter.rs
```diff
@@ -60,4 +71,47 @@ impl GPUAdapterMethods for GPUAdapter {
     fn Extensions(&self, _cx: SafeJSContext) -> NonNull<JSObject> {
         NonNull::new(self.extensions.get()).unwrap()
     }
+
+    /// https://gpuweb.github.io/gpuweb/#dom-gpuadapter-requestdevice
+    fn RequestDevice(&self, descriptor: &GPUDeviceDescriptor, comp: InCompartment) -> Rc<Promise> {
+        let promise = Promise::new_in_current_compartment(&self.global(), comp);
+        let sender = response_async(&promise, self);
+        let id = self.global().as_window().Navigator().create_device_id();
```
CYBAI: I just noticed that `GPUAdapter` is possibly exposed to `DedicatedWorker`.

Maybe we should call these window methods inside a `if let` instead 👀?

Ex.

```rust
if let Some(window) = self.global().downcast::<Window>() {
    let id = window.Navigator().create_device_id();
    ...
}
```

components/script/dom/gpuadapter.rs
imiklos: Yes, we will deal with it in a follow up patch.

components/script/dom/gpuadapter.rs
imiklos: Thank you! I will updated the pull request with this change.

imiklos:
@jdm @CYBAI Thank you for your review! I've addressed your comments and updated the PR.

kvark:
@jdm error looks unrelated to the PR (android release font linking)

jdm:
@bors-servo retry
* #24715

jdm:
@bors-servo retry
* network issue

jdm:
@bors-servo retry
* #24726 #21067

PR #25030, opened 2019-12-03
Only start WebGPU thread if an adapter is requested

zakorgy:
This addresses 1 and 2 from https://github.com/servo/servo/issues/24706#issuecomment-555491438
We send a message to constellation instead of creating the `WebGPU` thread on the start. We send back the result to script and set the `Window`'s `web_gpu` component there.
cc @jdm @imiklos 

---

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/25030)

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -173,7 +173,7 @@ use std::sync::Arc;
 use std::thread;
 use style_traits::viewport::ViewportConstraints;
 use style_traits::CSSPixel;
-use webgpu::WebGPU;
+use webgpu::{WebGPU, WebGPURequest};
 use webvr_traits::{WebVREvent, WebVRMsg};
 
 type PendingApprovalNavigations = HashMap<PipelineId, (LoadData, HistoryEntryReplacement)>;
@@ -242,6 +242,9 @@ struct BrowsingContextGroup {
     /// share an event loop, since they can use `document.domain`
     /// to become same-origin, at which point they can share DOM objects.
     event_loops: HashMap<Host, Weak<EventLoop>>,
+
+    /// The set of all WebGPU channels in this BrowsingContextGroup.
+    webgpus: HashMap<Host, WebGPU>,
 }
 
 /// The `Constellation` itself. In the servo browser, there is one
@@ -450,10 +453,6 @@ pub struct Constellation<Message, LTF, STF> {
     /// Entry point to create and get channels to a WebGLThread.
     webgl_threads: Option<WebGLThreads>,
 
-    /// An IPC channel for the constellation to send messages to the
-    /// WebGPU threads.
-    webgpu: Option<WebGPU>,
-
     /// A channel through which messages can be sent to the webvr thread.
     webvr_chan: Option<IpcSender<WebVRMsg>>,
 
@@ -536,9 +535,6 @@ pub struct InitialConstellationState {
 
     /// Entry point to create and get channels to a WebGLThread.
     pub webgl_threads: Option<WebGLThreads>,
-
-    /// A channel to the WebGPU threads.
-    pub webgpu: Option<WebGPU>,
 
     /// A channel to the webgl thread.
     pub webvr_chan: Option<IpcSender<WebVRMsg>>,
@@ -973,7 +969,6 @@ where
                         (rng, prob)
                     }),
                     webgl_threads: state.webgl_threads,
-                    webgpu: state.webgpu,
                     webvr_chan: state.webvr_chan,
                     webxr_registry: state.webxr_registry,
                     canvas_chan,
@@ -1230,7 +1225,6 @@ where
                 .webgl_threads
                 .as_ref()
                 .map(|threads| threads.pipeline()),
-            webgpu: self.webgpu.clone(),
             webvr_chan: self.webvr_chan.clone(),
             webxr_registry: self.webxr_registry.clone(),
             player_context: self.player_context.clone(),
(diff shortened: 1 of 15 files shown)


components/constellation/constellation.rs
```diff
@@ -1945,6 +1945,12 @@ where
                     EmbedderMsg::MediaSessionEvent(event),
                 ));
             },
+            FromScriptMsg::StartWebGPU(sender) => {
+                self.webgpu = WebGPU::new();
```
gterzian: Even though you are tracking the `webgpu_channel_requested` flag inside the DOM GPU, multiple script-threads(`EventLoop` from the perspective of the constellation) could send this message, in which case you would be overwriting previously set `self.webgpu`(and running multiple gpu threads?). 

Also, even within one script-thread, multiple windows could be running on it(see `documents` in script-thread), in which case I think you could also have multiple DOM `GPU` instances, each tracking their own `webgpu_channel_requested` flag. 

So I think you need to decide:

1. Do you want to allow multiple DOM `GPU` to share a `WebGPU` at the same time? Is so, how would they be shared(one webgpu shared across the browser, one shared per origin, per agent-cluster?). Then the `StartWebGPU` would need to include information so that you can either clone an existing one, or start a new one.
2. Do you want to allow multiple parallel `WebGPU` at all? In that case, you would need a map with a key representing how an instance is shared, so that you could get it out here, clone the sender, or start a new one. 

It's not clear to me from the spec how sharing should happen, although @kvark has previously said multiple parallel applications are wanted(although it's not clear if that means sharing a single webgpu, or running multiple ones, or a combination of that).

components/script/dom/gpu.rs
```diff
@@ -40,7 +44,23 @@ impl GPU {
...
     fn wgpu_channel(&self) -> Option<WebGPU> {
-        self.global().as_window().webgpu_channel()
+        let global = self.global();
+        let window = global.as_window();
+        if window.webgpu_channel().is_none() && !self.webgpu_channel_requested.get() {
+            self.webgpu_channel_requested.set(true);
+            let (sender, receiver) = ipc::channel().unwrap();
+            let script_to_constellation_chan = global.script_to_constellation_chan();
+            if script_to_constellation_chan
+                .send(ScriptMsg::StartWebGPU(sender))
```
gterzian: So I would say you could make the `WebGPURequest::RequestAdapter` into a `ScriptMsg`, and in `RequestAdapter` below just use the `script_to_constellation_chan` to send it to the constellation. 

Then the constellation can 

1. Start a webgpu if needed,
2. forward the `RequestAdapter` call.

And webgpu can then reply on a channel included in the `RequestAdapter` message with a `webgpu_channel` which can be stored in the adapter(so that it can communicate directly with webgpu going forward). 

The main benefit is not blocking on the constellation reply inside the `wgpu_channel` call, instead making it non-blocking as the actual `RequestAdapter` call, where the reply eventually resolves via a promise. 

You cold store the `webgpu_channel` somewhere besides the adapter, however it doesn't seem to have much benefit since any method of `GPU` would seem to have to go through the actual webgpu component. So it would only make sense to store it internally on the adapter, so that methods of it could send a message directly to webgpu without having to go through the constellation.

components/constellation/constellation.rs
kvark: Yes, the choice of boundaries for separating WebGPU instances is still ours.

Here is how it works in Gecko, for the reference:
  - each content process manages multiple tabs
  - a single WebGPU remote client is established per content process
  - it spawns a separate WebGPU server on the GPU process
  - the IPC is a child of CompositorBridge and thus the server message loop is happening on a thread that this parent protocol manages (I assume it's a separate thread)

Does Servo have multiple content processes? I think we can start here by having the same approach - one WebGPU client-server pair per content process.

gterzian: Yes so a content-process would correspond to a [`EventLoop`](https://github.com/servo/servo/blob/master/components/constellation/event_loop.rs) in Servo. 

From the constellation's perspective a document corresponds to a [`Pipeline`](https://github.com/servo/servo/blob/master/components/constellation/pipeline.rs), of which multiple can run on one event-loop. 

When a pipeline is spawned, an event-loop is either created or re-used, see https://github.com/servo/servo/blob/6e3c131139c412031a4d288aa1ffd048ca0311c3/components/constellation/pipeline.rs#L219

The "re-using" of event-loops is done when a pipeline is part of the same [`BrowsingContextGroup`](https://github.com/servo/servo/blob/6e3c131139c412031a4d288aa1ffd048ca0311c3/components/constellation/constellation.rs#L233), and based on the `Host` of the url of that pipeline, see [`get_event_loop`](https://github.com/servo/servo/blob/6e3c131139c412031a4d288aa1ffd048ca0311c3/components/constellation/constellation.rs#L1024).

So, this would actually be the first time something is shared like that on an "per-event-loop" basis, and you're going to have to provide some infra for this, because currently the constellation receives message from a given pipeline(see how each `ScriptMsg` comes with a `PipelineID`, representing the pipeline from which the message originated).

I think you could do something like:

1. As suggested in the other comment, make requesting an adapater a `ScriptMsg`.
2. When you receive it:
    - get the corresponding pipeline 
    - get the top-level browsing context for that pipeline.
    - get the host for that pipeline.
3. The above info should be enough to then grab the right webgpu instance. 
4. I would suggest you store them on a `HashMap<Host, WebGPU> inside a `BrowsingContextGroup`, similarly to what is done with `EventLoop`.

zakorgy:
@gterzian @kvark thanks for the well detailed descriptions, I will update my PR according to what you described.

zakorgy:
@gterzian @kvark it took me some time but finally updated the PR to use the solution described in https://github.com/servo/servo/pull/25030#discussion_r353538261.

components/constellation/constellation.rs
```diff
@@ -1945,9 +1962,68 @@ where
...
+            None => return warn!("ScriptMsg from closed pipeline {:?}.", source_pipeline_id),
+        };
+        match self
+            .browsing_context_group_set
+            .get_mut(&browsing_context_group_id)
+        {
+            Some(browsing_context_group) => {
+                match browsing_context_group.webgpus.entry(host) {
+                    Entry::Vacant(v) => {
+                        let webgpu = v.insert(WebGPU::new().unwrap());
```
jdm: Note: using unwrap in the constellation isn't allowed (and our CI catches it), so please turn this into an error! if it fails.

gterzian (changes requested):
Ok, I think it looks very good, great work!

Before I review everything in details, there are two small changes in the overall structure I would like to suggest:

1. Do we need the `Option<WebGPU>` in `WebGPUResponse::RequestAdapter`? Could we make sending back the channel non-optional(It doesn't seem to be about conveying failure to create an adapter)?
2. I would like to try to get rid of `window.wegbu_channel` and `window.set_webgpu_channel`. Instead, you could store the channel on the `GPUAdapter` directly, when you handle `WebGPUResponse::RequestAdapter`. Then, when the adapter creates a device, it could simply clone it's channel and store it directly on the device when handling the `WebGPUResponse::RequestDevice`(and the device could clone it's own sender when creating a buffer, and so on, without having to go through the window everytime)

zakorgy:
@gterzian those changes sounds doable for me. I will start adding them.

zakorgy:
@gterzian I've addressed your requested changes, could you please take a look now?

components/webgpu/lib.rs
```diff
@@ -16,7 +16,7 @@ use smallvec::SmallVec;
 
 #[derive(Debug, Deserialize, Serialize)]
 pub enum WebGPUResponse {
-    RequestAdapter(String, WebGPUAdapter),
+    RequestAdapter(String, WebGPUAdapter, WebGPU),
```
gterzian: Could it make sense for the `WebGPU` to keep a sender to itself, so that the constellation doesn't have to send it in the request?

gterzian (changes requested):
Ok it looks good to me. I have one small comment, and I'll take another read after it has been addressed.

components/webgpu/lib.rs
zakorgy: I think yes. Updated the code according to that.

gterzian (changes requested):
Ok looks good, just a few last suggestions.

zakorgy:
@gterzian updated the PR with your proposed changes, also added a separate commit where I've addressed https://github.com/servo/servo/issues/25492

gterzian:
FIX #25492

Great job!

@bors-servo r+

zakorgy:
@gterzian thanks for the guidance and the valuable information about Servo infrastructure.

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

PR #25031, opened 2019-12-03
Implement size extremums

nox:


components/layout_2020/flow/mod.rs
```diff
@@ -486,23 +541,155 @@ fn layout_in_flow_replaced_block_level<'a>(
...
         .block
-        .maybe_percentage_relative_to(containing_block.block_size.non_auto());
+        .maybe_percentage_relative_to(cbbs)
+        .auto_is(Length::zero);
+    let max_block_size = match max_box_size.block {
+        MaxSize::LengthPercentage(max_block_size) => {
+            max_block_size.maybe_percentage_relative_to(cbbs)
+        },
+        MaxSize::None => None,
+    };
```
SimonSapin: Looks like this `match` could be a method of `MaxSize`?

components/layout_2020/flow/mod.rs
```diff
@@ -486,23 +541,155 @@ fn layout_in_flow_replaced_block_level<'a>(
...
         },
         (LengthOrAuto::LengthPercentage(inline), LengthOrAuto::Auto) => {
-            (inline, inline / intrinsic_ratio)
+            clamp(inline, inline / intrinsic_ratio)
         },
         (LengthOrAuto::Auto, LengthOrAuto::LengthPercentage(block)) => {
-            (block * intrinsic_ratio, block)
+            clamp(block * intrinsic_ratio, block)
+        },
+        (LengthOrAuto::Auto, LengthOrAuto::Auto) => {
```
SimonSapin: Note to myself: review this case tomorrow :)

components/layout_2020/flow/mod.rs
nox: I didn't want to because we actually unwrap the `NonNegativeLengthPercentage` in `StyleExt::max_box_size`, and it feels weird to me to add methods to `MaxSize<LengthPercentage>` in stylo if they are only used in layout. Maybe at some point we should consider bringing back `NonNegativeLengthPercentage` and `NonNegativeLength` in the layout code. I feel like this would be helpful avoid cases where we end up with negative lengths too anyway.

components/layout_2020/flow/mod.rs
```diff
@@ -486,23 +541,155 @@ fn layout_in_flow_replaced_block_level<'a>(
...
+                    (max_inline_size, block_size)
+                },
+                // Row 3.
+                (Violation::Below(min_inline_size), Violation::None) => {
+                    let mut block_size = min_inline_size / intrinsic_ratio;
+                    if let Some(max_block_size) = max_block_size {
+                        if block_size > max_block_size {
+                            block_size = max_block_size;
+                        }
+                    }
```
SimonSapin: Nit: why not use `.min()` here? Or even for a function that handles the `Option`

SimonSapin (approved):

components/layout_2020/flow/mod.rs
SimonSapin: Something like:

```rust
fn maybe_clamp_at_max(x: Length, max_x: Option<Length>) -> Length {
    if let Some(max_x) = max_x {
        x.min(max_x)
    } else {
        x
    }
}

File: components/layout_2020/flow/mod.rs [hunk 4]
```diff
@@ -755,11 +708,44 @@ fn solve_inline_margins_for_in_flow_block_level(
...
-    if let Some(max_size) = max_size {
-        if size > max_size {
-            size = max_size;
-        }
+fn clamp_between_extremums(size: Length, min_size: Length, max_size: Option<Length>) -> Length {
+    clamp_below_max(size, max_size).max(min_size)
+}
+
+fn clamp_below_max(size: Length, max_size: Option<Length>) -> Length {
+    max_size.filter(|max_size| size > *max_size).unwrap_or(size)
```
SimonSapin: This feels a bit clever to me, as in needs a few more seconds to follow what’s going on. What do you think of using `.min(…)`?

SimonSapin (approved):

components/layout_2020/flow/mod.rs
nox: All I needed is <s>love</s> `map` and `min` indeed.

PR #25033, opened 2019-12-03
Add support for inline-block and for computing min/max-content

SimonSapin:


components/layout_2020/flow/inline.rs
```diff
@@ -77,6 +81,119 @@ struct LinesBoxes {
...
+                                // atomic
+                                // .intrinsic_inline_sizes
+                                // .as_ref()
+                                // .expect("Accessing intrinsic size that was not requested")
+                                // .clone()
+                                todo!()
+                            };
+                            let (outer, pc) = outer_intrinsic_inline_sizes(&atomic.style, &inner);
+                            self.current_line.min_content += outer.min_content;
+                            self.current_line.max_content += outer.max_content;
```
nox: This can be yet another method on `IntrinsicSizes`.

components/layout_2020/flow/inline.rs
```diff
@@ -77,6 +81,119 @@ struct LinesBoxes {
...
 impl InlineFormattingContext {
+    // This works on an already-constructed `InlineFormattingContext`,
+    // Which would have to change if/when
+    // `BlockContainer::construct` parallelize their construction.
+    #[allow(unused)]
+    pub(super) fn intrinsic_sizes(&self, layout_context: &LayoutContext) -> IntrinsicSizes {
+        struct Computation {
+            paragraph: IntrinsicSizes,
+            current_line: IntrinsicSizes,
+            current_line_percentages: Percentage,
```
nox: I'm not sure why this is plural.

components/layout_2020/flow/inline.rs
```diff
@@ -77,6 +81,119 @@ struct LinesBoxes {
...
+                                    add_length!($x.length_component());
+                                    self.current_line_percentages += $x.percentage_component();
+                                }};
+                            }
+                            macro_rules! add {
+                                ($condition: ident, $side: ident) => {
+                                    if inline_box.$condition {
+                                        add_lengthpercentage!(padding.$side);
+                                        add_length!(border.$side);
+                                        margin.$side.non_auto().map(|x| add_lengthpercentage!(x));
```
nox: Nit: I don't like using `map` if you don't care about the return value and this is only done for side-effects, `if let Some(lp)` would be clearer IMO.

components/layout_2020/flow/inline.rs
```diff
@@ -77,6 +81,119 @@ struct LinesBoxes {
...
+                self.current_line
+                    .adjust_for_pbm_percentages(take(&mut self.current_line_percentages));
+                self.paragraph
+                    .max_content
+                    .max_assign(take(&mut self.current_line.max_content));
+            }
+        }
+        fn take<T: Zero>(x: &mut T) -> T {
+            std::mem::replace(x, T::zero())
+        }
```
nox: Don't we only call that twice?

components/layout_2020/flow/inline.rs
```diff
@@ -283,11 +400,13 @@ impl<'box_tree> PartialInlineBoxFragment<'box_tree> {
 }
 
 impl TextRun {
-    fn layout(&self, layout_context: &LayoutContext, ifc: &mut InlineFormattingContextState) {
+    fn break_and_shape(
+        &self,
+        layout_context: &LayoutContext,
+    ) -> (Au, Au, FontInstanceKey, Vec<GlyphRun>, bool) {
```
nox: Please introduce a structure for that.

components/layout_2020/intrinsic.rs
```diff
@@ -16,13 +16,26 @@ impl IntrinsicSizes {
             max_content: Length::zero(),
         }
     }
+
+    /// Relevant to outer intrinsic inline sizes, for percentages from padding and margin.
+    pub fn adjust_for_pbm_percentages(&mut self, percentages: Percentage) {
```
nox: Same comment about plural percentages.

components/layout_2020/style_ext.rs
```diff
@@ -58,6 +60,23 @@ impl ComputedValuesExt for ComputedValues {
...
+
+    fn inline_size_is_auto(&self) -> bool {
+        let position = self.get_position();
+        let size = if self.writing_mode_is_horizontal() {
+            position.width
+        } else {
+            position.height
+        };
+        matches!(size, Size::Auto)
+    }
```
nox: Why not just do `self.box_size().inline == LengthOrAuto::Auto`? At the very least, I don't think the `matches!` is necessary here.

components/layout_2020/flow/construct.rs
```diff
@@ -158,14 +158,15 @@ impl BlockContainer {
         }
 
         let mut contains_floats = builder.contains_floats;
+        let request_content_sizes = false; // FIXME
```
nox: New type, no boolean parameters.

components/layout_2020/style_ext.rs
```diff
@@ -77,6 +78,19 @@ impl ComputedValuesExt for ComputedValues {
         matches!(size, Size::Auto)
     }
 
+    fn inline_box_offsets_are_both_auto(&self) -> bool {
+        let position = self.get_position();
+        let offsets = if self.writing_mode_is_horizontal() {
+            (position.left, position.right)
+        } else {
+            (position.top, position.bottom)
+        };
```
nox: Why not just use `self.box_offsets()`?

components/layout_2020/style_ext.rs
```diff
@@ -77,6 +78,19 @@ impl ComputedValuesExt for ComputedValues {
...
+        matches!(
+            offsets,
+            (LengthPercentageOrAuto::Auto, LengthPercentageOrAuto::Auto)
+        )
```
nox: Not sure why you always use `matches!` where a simple equality is enough. :)

components/layout_2020/positioned.rs
```diff
@@ -42,6 +45,25 @@ pub(crate) enum AbsoluteBoxOffsets<NonStatic> {
...
+            contents: IndependentFormattingContext::construct(
+                context,
+                style,
+                display_inside,
+                contents,
+                request_content_sizes,
+            ),
+        }
+    }
+
```
nox: I feel like this commit would benefit from being split in two: first part with the introduction of the additional `construct` methods, second part with the introduction of the new `request_content_sizes` argument.

components/layout_thread_2020/lib.rs
```diff
@@ -1082,7 +1082,8 @@ impl LayoutThread {
             driver::traverse_dom(&traversal, token, Some(rayon_pool));
 
             let root_node = document.root_element().unwrap().as_node();
-            let box_tree = rayon_pool.install(|| BoxTreeRoot::construct(traversal.context(), root_node));
+            let box_tree =
+                rayon_pool.install(|| BoxTreeRoot::construct(traversal.context(), root_node));
```
nox: Why did that change? Oh, forgot to run fmt on a previous commit?

components/layout_2020/flow/construct.rs
```diff
@@ -157,25 +157,29 @@ impl BlockContainer {
             builder.end_ongoing_inline_formatting_context();
         }
 
-        let mut contains_floats = builder.contains_floats;
+        type Intermediate<Node> = IntermediateBlockLevelBox<Node>;
+        #[derive(Default)]
+        struct Target {
```
nox: Why `Target`?

components/layout_2020/flow/construct.rs
```diff
@@ -157,25 +157,29 @@ impl BlockContainer {
...
+        let iter = iter.mapfold_reduce_into(
+            &mut target,
+            |target, (intermediate, box_slot): (Intermediate<_>, BoxSlot<'_>)| {
+                let (block_level_box, box_contains_floats) =
+                    intermediate.finish(context, request_content_sizes);
+                target.contains_floats |= box_contains_floats;
+                box_slot.set(LayoutBox::BlockLevel(block_level_box.clone()));
+                block_level_box
+            },
+            |left, right| left.contains_floats |= right.contains_floats,
```
nox: I don't feel like the intermediate `iter` bindings and the local `Intermediate` type alias help readability, I prefer the previous version of the code.

components/layout_2020/flow/construct.rs
```diff
@@ -158,25 +162,46 @@ impl BlockContainer {
...
+                    outer_content_sizes_of_children: ContentSizes::zero(),
+                }
+            }
         }
         let mut target = Target {
             contains_floats: builder.contains_floats,
+            outer_content_sizes_of_children: ContentSizes::zero(),
         };
-        let request_content_sizes = false; // FIXME
+        let request_childrens_outer_content_sizes = request_content_sizes;
```
nox: Just pass `request_content_sizes`, no?

components/layout_2020/flow/construct.rs
```diff
@@ -158,25 +162,46 @@ impl BlockContainer {
...
                 box_slot.set(LayoutBox::BlockLevel(block_level_box.clone()));
                 block_level_box
             },
-            |left, right| left.contains_floats |= right.contains_floats,
+            |left, right| {
+                left.contains_floats |= right.contains_floats;
+                if request_content_sizes {
+                    left.outer_content_sizes_of_children
+                        .max_assign(&right.outer_content_sizes_of_children)
+                }
```
nox: I do not think this is compatible with how we handle margin collapsing.

components/layout_2020/flow/construct.rs
```diff
@@ -639,9 +677,3 @@ impl std::ops::BitOrAssign for ContainsFloats {
         }
     }
 }
-
-impl Default for ContainsFloats {
-    fn default() -> Self {
-        ContainsFloats::No
-    }
-}
```
nox: Why remove this impl?

components/layout_2020/positioned.rs
```diff
@@ -292,14 +292,23 @@ impl<'a> AbsolutelyPositionedFragment<'a> {
...
+            Err(non_replaced) => {
+                let dummy_tree_rank = 0;
+                non_replaced.layout(
+                    layout_context,
+                    &containing_block_for_children,
+                    dummy_tree_rank,
+                    &mut absolutely_positioned_fragments,
+                )
+            },
+        };
```
nox: It more and more looks like we should just have additional `Replaced` variants in various enums instead of putting them in the `IndependentFormattingContext` type.

components/layout_2020/sizing.rs
```diff
@@ -96,3 +93,20 @@ pub(crate) fn outer_inline_content_sizes_and_percentages(
...
+    available_size
+        .max(content_sizes.min_content)
+        .min(content_sizes.max_content)
+}
+
+fn expect(content_sizes: &Option<ContentSizes>) -> &ContentSizes {
+    content_sizes
+        .as_ref()
+        .expect("Accessing content size that was not requested")
+}
```
nox: What do you think of a structure with these two as a method and a single `Option<ContentSizes>` field?

components/layout_2020/flow/inline.rs
```diff
@@ -10,14 +10,14 @@ use crate::fragments::CollapsedBlockMargins;
 use crate::fragments::{AnonymousFragment, BoxFragment, Fragment, TextFragment};
 use crate::geom::flow_relative::{Rect, Sides, Vec2};
 use crate::positioned::{AbsolutelyPositionedBox, AbsolutelyPositionedFragment};
-use crate::sizing::{outer_inline_content_sizes_and_percentages, ContentSizes};
+use crate::sizing::{outer_inline_content_sizes_and_percentages, shrink_to_fit, ContentSizes};
```
nox: Nit: I prefer to call functions with paths qualified by their containing module.

components/layout_2020/flow/inline.rs
```diff
@@ -67,8 +67,9 @@ struct PartialInlineBoxFragment<'box_tree> {
     parent_nesting_level: InlineNestingLevelState<'box_tree>,
 }
 
-struct InlineFormattingContextState<'box_tree, 'cb> {
-    containing_block: &'cb ContainingBlock,
+struct InlineFormattingContextState<'box_tree, 'a> {
+    absolutely_positioned_fragments: &'a mut Vec<AbsolutelyPositionedFragment<'box_tree>>,
```
nox: Consider making this change separately from the implementation of `inline-block` so that its rationale is written in a commit message.

components/layout_2020/flow/inline.rs
```diff
@@ -393,6 +391,118 @@ impl<'box_tree> PartialInlineBoxFragment<'box_tree> {
...
+) {
+    let cbis = ifc.containing_block.inline_size;
+    let padding = atomic.style.padding().percentages_relative_to(cbis);
+    let border = atomic.style.border_width();
+    let margin = atomic
+        .style
+        .margin()
+        .percentages_relative_to(cbis)
+        .auto_is(Length::zero);
+    ifc.inline_position += padding.inline_start + border.inline_start + margin.inline_start;
```
nox: AFAICT, we always add together the three same values from `padding`, `border` and `margin`, consider introducing a local `pbm` variable as we do in the other layout functions.

components/layout_2020/flow/inline.rs
```diff
@@ -393,6 +391,118 @@ impl<'box_tree> PartialInlineBoxFragment<'box_tree> {
...
+            );
+            // FIXME is this correct?
+            let dummy_tree_rank = 0;
+            // FIXME: Do we need to call `adjust_static_positions` somewhere near here?
+            let independent_layout = non_replaced.layout(
+                layout_context,
+                &containing_block_for_children,
+                dummy_tree_rank,
+                ifc.absolutely_positioned_fragments,
+            );
```
nox: A dummy tree rank should only be used if this is an absolute containing block (i.e. a containing block in which absolutely positioned descendants should be placed relatively to).

components/layout_2020/positioned.rs
```diff
@@ -302,24 +302,35 @@ impl<'a> AbsolutelyPositionedFragment<'a> {
...
         let mut independent_layout = match self.absolutely_positioned_box.contents.as_replaced() {
-            // FIXME: implement https://drafts.csswg.org/css2/visudet.html#abs-replaced-width
-            Ok(replaced) => replaced.layout(
-                &self.absolutely_positioned_box.contents.style,
-                &containing_block_for_children,
-            ),
+            Ok(replaced) => {
+                // FIXME: implement https://drafts.csswg.org/css2/visudet.html#abs-replaced-width
+                // and https://drafts.csswg.org/css2/visudet.html#abs-replaced-height
+                let block_size = block_size.auto_is(|| Length::zero());
```
nox: Nit: `.auto_is(Length::zero)`.

nox (commented):
Lots of remarks but many of them are not that interesting.

components/layout_2020/flow/construct.rs
```diff
@@ -127,6 +127,7 @@ impl BlockContainer {
         context: &SharedStyleContext<'style>,
         block_container_style: &Arc<ComputedValues>,
         contents: NonReplacedContents<impl NodeExt<'dom>>,
+        //        intrinsic_sizes_requested: bool,
```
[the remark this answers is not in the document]
SimonSapin: It’s due to the combination of my editor’s “Comment this line” command inserting `//` at the start of the line, and then rustfmt indenting  comments to the same level as the stuff around them. But it’s uncommented in a later commit anyway.

components/layout_2020/intrinsic.rs
```diff
@@ -0,0 +1,69 @@
...
+    // Percentages for 'width' are treated as 'auto'
+    let specified = specified.map(|lp| lp.as_length());
+    // The (inner) min/max-content are only used for 'auto'
+    let mut outer = match specified.non_auto().flatten() {
+        None => get_inner_intrinsic_sizes(),
+        Some(length) => IntrinsicSizes {
+            min_content: length,
+            max_content: length,
+        }
+    };
```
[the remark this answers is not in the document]
SimonSapin: Oh yeah I totally used the wrong name here. `computed` is correct per https://drafts.csswg.org/css-cascade/#computed (though it doesn’t match the `getComputedStyle` API, but it’s the API which is named incorrectly) for the first binding.

I’ve renamed to `inline_size`, but kept the bindings so that comments apply to each step and in the expected order.

components/layout_2020/intrinsic.rs
```diff
@@ -0,0 +1,69 @@
...
+    outer.max_content += bpm_lengths;
+    // " Note that this may yield an infinite result, but undefined results
+    //   (zero divided by zero) must be treated as zero. "
+    if outer.max_content.px() == 0. {
+        // Avoid a potential `NaN`.
+        // Zero is already the result we want regardless of `denominator`.
+    } else {
+        let denominator = (1. - bpm_percentages).max(0.);
+        outer.max_content = Length::new(outer.max_content.px() / denominator);
+    }
```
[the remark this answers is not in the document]
SimonSapin: Dividing a non-zero finite value by zero creates infinity, not `NaN`. Per the bit of (proto-)spec quoted above, infinity is the expected result of this computation in this case. An infinite max-content makes shrink-to-fit use all of the available space.

components/layout_2020/flow/inline.rs
SimonSapin: Let’s do that if there would be more than one caller. (I did add the other two methods since they replace macros.)

components/layout_2020/flow/inline.rs
SimonSapin: Maybe it shouldn’t? It’s (potentially) the sum of percentages values from multiple properties on multiple boxes.

components/layout_2020/flow/inline.rs
SimonSapin: Done.

components/layout_2020/flow/inline.rs
SimonSapin: Three times. I found it made the code around the calls easier to follow (especially since it’s a bit repetitive).

components/layout_2020/flow/inline.rs
SimonSapin: Done.

components/layout_thread_2020/lib.rs
SimonSapin: Probably?

components/layout_2020/positioned.rs
SimonSapin: I agree this would have been better, but at this point it’s not trivial to go back and make an intermediate commit that compiles. Do you feel strongly about this?

components/layout_2020/flow/construct.rs
SimonSapin: That’s how it’s called in https://docs.rs/rayon_croissant/0.1.1/rayon_croissant/trait.ParallelIteratorExt.html#method.mapfold_reduce_into, and it’s a type that exists only for the purpose of calling that method.

components/layout_2020/style_ext.rs
SimonSapin: It seems like more work to also compute the block_size only to drop it. But anyway since https://doc.servo.org/style/properties/style_structs/struct.Position.html#structfield.inline_size & co exist in memory we should remove all of these methods and access those fields instead. But let’s keep that change for a future PR

components/layout_2020/style_ext.rs
SimonSapin: Out of habit from some enums that don’t implement `PartialEq`. But yeah `==` would be better.

components/layout_2020/flow/construct.rs
SimonSapin: That was an attempt at reducing rightwards drift and avoiding line breaks in the middle of a closure’s parameter list.

components/layout_2020/flow/construct.rs
SimonSapin: At some point I had a boolean parameter in `IntermediateBlockLevelBox::finish` (instead of `Option<&mut _>`) and a different name help indicate who requests what from whom, but yeah now it doesn’t matter. I’ll change it.

components/layout_2020/flow/construct.rs
SimonSapin: This only computes sizes in the inline direction, margin-collapsing only affects the block direction.

components/layout_2020/flow/construct.rs
SimonSapin: It was only used in a code that called `Default::default()` to initialize three struct fields, and I felt that having type-specific constructors like `Vec::new()` and `None` made that code easier to read.

components/layout_2020/positioned.rs
SimonSapin: Maybe. But this PR is already large :)

components/layout_2020/flow/inline.rs
SimonSapin: What change? This field is only added for the because layout code for `inline-block` needs to access it.

components/layout_2020/flow/inline.rs
SimonSapin: Done.

components/layout_2020/flow/inline.rs
SimonSapin: That’s what I suspected. But I think line breaking makes it tricky to computed the correct rank, especially when we’ll add support for going “back” through an atomic inline-level to find the last wrap opportunity after overflowing a line.

components/layout_2020/positioned.rs
SimonSapin: Done.

components/layout_2020/flow/construct.rs
SimonSapin: Done.

components/layout_2020/sizing.rs
SimonSapin: Added new enum.

components/layout_2020/flow/inline.rs
SimonSapin: These are now methods (and so don’t need to be imported).

components/layout_2020/flow/construct.rs
SimonSapin: Maybe `Accumulator`? (Might work for `rayon_croissant` too)

Issue #24762, opened 2019-11-17
Intermittent timeouts in reftests


jdm:
```
  ▶ TIMEOUT [expected PASS] /FileAPI/url/url_xmlhttprequest_img.html
  │ 
  │ 
  │ 
  └ 
  ▶ TIMEOUT [expected PASS] /css/CSS2/colors/color-002.xht
  │ 
  │ 
  │ 
  └ 
  ▶ TIMEOUT [expected PASS] /css/CSS2/colors/color-004.xht
  │ 
  │ 
  │ 
  └ 
```
I'm really not sure where to start investigating here. It would be useful to know in which file that makes up a reftest the timeout actually occurs.

jdm:
Interestingly, this exact same set of timeouts was observed in #24753 as well.

jdm:
Seen in /css/CSS2/selectors/first-letter-punctuation-077.xht.

jdm:
```
  ▶ TIMEOUT [expected PASS] /css/CSS2/margin-padding-clear/margin-left-008.xht
  │ 
  │ 
  │ 
  └ 
```

jdm:
Observed in linux as well:
```
  ▶ TIMEOUT [expected FAIL] /css/css-paint-api/style-before-pseudo.https.html
  │ 
  │ 
  │ 
  └ 
```

Manishearth:
In #25144:

```
  ▶ TIMEOUT [expected PASS] /css/CSS2/box-display/insert-inline-in-blocks-middle-002.xht
```

and 

```
  â–¶ TIMEOUT [expected PASS] /css/CSS2/linebox/border-padding-bleed-002.xht


  â–¶ TIMEOUT [expected FAIL] /css/CSS2/linebox/anonymous-inline-inherit-001.html


  â–¶ TIMEOUT [expected PASS] /css/CSS2/linebox/fractional-line-height.html


  â–¶ TIMEOUT [expected PASS] /css/CSS2/linebox/border-padding-bleed-001.xht


  â–¶ TIMEOUT [expected PASS] /css/CSS2/csswg-issues/submitted/css2.1/anonymous-boxes-001.xht


  â–¶ TIMEOUT [expected FAIL] /css/CSS2/linebox/baseline-block-with-overflow-001.html


  â–¶ TIMEOUT [expected FAIL] /css/CSS2/linebox/empty-inline-003.xht
```

Manishearth:
[...]

Issue #24501, opened 2019-10-19
Intermittent timeout in /html/semantics/forms/form-submission-0/form-double-submit.html, /html/semantics/forms/form-submission-0/form-double-submit-3.html, /html/semantics/forms/form-submission-0/form-double-submit-2.html


jdm:

pshaughn:
I think the test changes I made in #25636 might turn these deterministic.

yvt:
#15727 is another source of intermittency, as verified by repeated test runs.

|             Test            | % of runs passed | ... after the fix |
| :-------------------------- | ---------------: | ----------------: |
| `form-double-submit.html`   |               9% |              100% |
| `form-double-submit-2.html` |              15% |              100% |
| `form-double-submit-3.html` |              17% |              100% |

<details>
<summary>More</summary>

With the fix:

```
$ set num_runs 100; set num_passed 0; time for i in (seq $num_runs); ./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms/form-submission-0/form-double-submit.html --binary-arg=--cpu --headless --release; and set num_passed (math $num_passed + 1); end; echo "$num_passed of $num_runs runs passed"
________________________________________________________
Executed in   29.74 mins    fish           external
   usr time   27.68 mins   58.16 millis   27.67 mins
   sys time    2.29 mins  102.44 millis    2.29 mins

Of 100 runs, 100 passed

$ set num_runs 100; set num_passed 0; time for i in (seq $num_runs); ./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms/form-submission-0/form-double-submit-2.html --binary-arg=--cpu --headless --release; and set num_passed (math $num_passed + 1); end; echo "$num_passed of $num_runs runs passed"
________________________________________________________
Executed in   19.85 mins    fish           external
   usr time   18.27 mins   50.38 millis   18.27 mins
   sys time    1.52 mins   70.52 millis    1.52 mins

Of 100 runs, 100 passed

$ set num_runs 100; set num_passed 0; time for i in (seq $num_runs); ./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms/form-submission-0/form-double-submit-3.html --binary-arg=--cpu --headless --release; and set num_passed (math $num_passed + 1); end; echo "$num_passed of $num_runs runs passed"

________________________________________________________
Executed in   24.62 mins    fish           external
   usr time   22.62 mins   57.59 millis   22.62 mins
[...]

PR #25070, opened 2019-12-03
Named form getter

### LINKED ISSUE #16479: won't login into zoho.com because of missing named getter on HTMLFormElement
Servo reports an error in the authentication. Sometimes it freezes afterwards. Windows build.

jaymodi98:
This PR contains changes related to adding named getter in Servo for getting the list of all meaningful property names for a HTMLFormElement object, and getting the value of a specific property name. The following changes have been made:

* uncomment the [named getter](https://github.com/servo/servo/blob/f63b404e0cbf30380c4043700861110d06e548bb/components/script/dom/webidls/HTMLFormElement.webidl#L30) from HTMLFormElement.webidl
* add the missing `NamedGetter` and `SupportedPropertyNames` methods to [HTMLFormElement](https://github.com/servo/servo/blob/f63b404e0cbf30380c4043700861110d06e548bb/components/script/dom/htmlformelement.rs#L113)
* implement `SupportedPropertyNames` according to [the specification](https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names):
  * create an enum to represent the `id`, `name`, and `past` states for the sourced names
  * create a vector of `(SourcedName, DomRoot<HTMLElement>)` by iterating over `self.controls` and checking the element type and calling methods like `HTMLElement::is_listed_element`
---
- [x] These changes fix #16479 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -274,7 +274,8 @@ impl HTMLFormElementMethods for HTMLFormElement {
                 .downcast::<HTMLElement>()
                 .map_or(false, |c| c.is_listed_element())
             {
-                if child.has_attribute(&local_name!("id")) ||
+                if (child.has_attribute(&local_name!("id")) &&
+                    child.get_string_attribute(&local_name!("id")) == name) ||
                     (child.has_attribute(&local_name!("name")) &&
                         child.get_string_attribute(&local_name!("name")) == name)
                 {
@@ -286,7 +287,8 @@ impl HTMLFormElementMethods for HTMLFormElement {
         if candidates.len() == 0 {
             for child in controls.iter() {
                 if child.is::<HTMLImageElement>() {
-                    if child.has_attribute(&local_name!("id")) ||
+                    if (child.has_attribute(&local_name!("id")) &&
+                        child.get_string_attribute(&local_name!("id")) == name) ||
                         (child.has_attribute(&local_name!("name")) &&
                             child.get_string_attribute(&local_name!("name")) == name)
                     {
@@ -459,13 +461,16 @@ impl HTMLFormElementMethods for HTMLFormElement {
         // Step 6
         sourcedNamesVec.retain(|sn| !sn.name.to_string().is_empty());
 
-        // Step 7
-        // Q1. Unable to clearly understand. It seems to contradict with step 4.
-
-        // Step 8
+        // Step 7-8
         let mut namesVec: Vec<DOMString> = Vec::new();
         for elem in sourcedNamesVec.iter() {
-            namesVec.push(elem.name.clone());
+            if namesVec
+                .iter()
+                .find(|name| name.to_string() == elem.name.to_string())
+                .is_none()
+            {
+                namesVec.push(elem.name.clone());
+            }
         }
 
         return namesVec;
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -256,18 +256,18 @@ impl HTMLFormElementMethods for HTMLFormElement {
         elements.IndexedGetter(index)
     }
 
+    // https://html.spec.whatwg.org/multipage/#the-form-element%3Adetermine-the-value-of-a-named-property
     fn NamedGetter(&self, name: DOMString) -> Option<RadioNodeListOrElement> {
         // return Option::Some::<RadioNodeListOrElement>;
         unimplemented!();
     }
 
-    // https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names
+    // https://html.spec.whatwg.org/multipage/#the-form-element:supported-property-names
     fn SupportedPropertyNames(&self) -> Vec<DOMString> {
-
         enum SourcedNameSource {
             Id,
             Name,
-            Past(std::time::Duration)
+            Past(std::time::Duration),
         }
 
         struct SourcedName {
@@ -285,16 +285,23 @@ impl HTMLFormElementMethods for HTMLFormElement {
         // controls - list of form elements
         // check all listed elements first, push to sourcedNamesVec as per spec
         for child in controls.iter() {
-
-            if child.is_html_element() { // if child.is_listed()
+            if child.is_html_element() {
+                // if child.is_listed()
 
                 if child.has_attribute(&local_name!("id")) {
                     // https://learning-rust.github.io/docs/b2.structs.html
-                    let entry = SourcedName {name: child.get_string_attribute(&local_name!("id")), element: child.root_element(), source: SourcedNameSource::Id};
+                    let entry = SourcedName {
+                        name: child.get_string_attribute(&local_name!("id")),
+                        element: child.root_element(),
+                        source: SourcedNameSource::Id,
+                    };
                     sourcedNamesVec.push(entry);
-                }
-                else if child.has_attribute(&local_name!("name")) {
-                    let entry = SourcedName {name: child.get_string_attribute(&local_name!("name")), element: child.root_element(), source: SourcedNameSource::Name};
+                } else if child.has_attribute(&local_name!("name")) {
+                    let entry = SourcedName {
+                        name: child.get_string_attribute(&local_name!("name")),
+                        element: child.root_element(),
+                        source: SourcedNameSource::Name,
+                    };
                     sourcedNamesVec.push(entry);
                 }
             }
@@ -307,14 +314,20 @@ impl HTMLFormElementMethods for HTMLFormElement {
 
             // https://users.rust-lang.org/t/how-check-type-of-variable/33845/7
             if child.is::<HTMLImageElement>() {
-
                 if child.has_attribute(&local_name!("id")) {
                     // https://learning-rust.github.io/docs/b2.structs.html
-                    let entry = SourcedName {name: child.get_string_attribute(&local_name!("id")), element: child.root_element(), source: SourcedNameSource::Id};
+                    let entry = SourcedName {
+                        name: child.get_string_attribute(&local_name!("id")),
+                        element: child.root_element(),
+                        source: SourcedNameSource::Id,
+                    };
                     sourcedNamesVec.push(entry);
-                }
-                else if child.has_attribute(&local_name!("name")) {
-                    let entry = SourcedName {name: child.get_string_attribute(&local_name!("name")), element: child.root_element(), source: SourcedNameSource::Name};
+                } else if child.has_attribute(&local_name!("name")) {
+                    let entry = SourcedName {
+                        name: child.get_string_attribute(&local_name!("name")),
+                        element: child.root_element(),
+                        source: SourcedNameSource::Name,
+                    };
                     sourcedNamesVec.push(entry);
                 }
             }
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -69,6 +69,8 @@ use crate::dom::radionodelist::RadioNodeList;
 use std::collections::HashMap;
 use time::{now, Duration, Tm};
 
+// use crate::dom::bindings::codegen::Bindings::NodeBinding::NodeBinding::NodeMethods;
+
 #[derive(Clone, Copy, JSTraceable, MallocSizeOf, PartialEq)]
 pub struct GenerationId(u32);
 
@@ -334,10 +336,20 @@ impl HTMLFormElementMethods for HTMLFormElement {
     // https://html.spec.whatwg.org/multipage/#the-form-element:supported-property-names
     fn SupportedPropertyNames(&self) -> Vec<DOMString> {
         // Step 1
+        #[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
         enum SourcedNameSource {
             Id,
             Name,
             Past(Duration),
+        }
+
+        impl SourcedNameSource {
+            fn is_past(&self) -> bool {
+                match self {
+                    SourcedNameSource::Past(..) => true,
+                    _ => false,
+                }
+            }
         }
 
         struct SourcedName {
@@ -403,15 +415,30 @@ impl HTMLFormElementMethods for HTMLFormElement {
             let entry = SourcedName {
                 name: key.clone(),
                 element: DomRoot::from_ref(&*val.0),
-                source: SourcedNameSource::Past(now()-val.1), // calculate difference now()-val.1 to find age
+                source: SourcedNameSource::Past(now() - val.1), // calculate difference now()-val.1 to find age
             };
             sourcedNamesVec.push(entry);
         }
 
         // Step 5
-        // TODO need to sort as per spec. This is a partially implemented function.
-        // Kindly guide us on how to refine this function further.
-        sourcedNamesVec.sort_by(|a, b| a.name.partial_cmp(&b.name).unwrap());
+        // TODO need to sort as per spec. Kindly guide us how to sort by tree order of the element entry of each tuple.
+        // Following gives error as CompareDocumentPosition return u16 and not Ordering.
+        // sourcedNamesVec.sort_by(|a, b|
+        //     a.element.upcast::<Node>().CompareDocumentPosition(b.element.upcast::<Node>())
+        // );
+
+        // The remaining part where sorting is to be done by putting entries whose source is id first,
+        // then entries whose source is name, and finally entries whose source is past,
+        // and sorting entries with the same element and source by their age, oldest first.
+        // can be done as follows:
+
+        sourcedNamesVec.sort_by(|a, b| {
+            if a.source.is_past() && b.source.is_past() {
+                b.source.cmp(&a.source)
+            } else {
+                a.source.cmp(&b.source)
+            }
+        });
 
         // Step 6
         sourcedNamesVec.retain(|sn| !sn.name.to_string().is_empty());
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -64,6 +64,8 @@ use std::cell::Cell;
 use style::attr::AttrValue;
 use style::str::split_html_space_chars;
 
+use crate::dom::bindings::codegen::UnionTypes::RadioNodeListOrElement;
+
 #[derive(Clone, Copy, JSTraceable, MallocSizeOf, PartialEq)]
 pub struct GenerationId(u32);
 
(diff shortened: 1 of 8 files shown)


components/script/dom/htmlformelement.rs
```diff
@@ -253,6 +255,92 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+        // vector to store sourced names tuple information
+        let mut sourcedNamesVec: Vec<SourcedName> = Vec::new();
+
+        // let controls = self.controls.borrow_mut(); - line 849 → unsure of which "borrow" to use
+        let controls = self.controls.borrow(); // line 807 in this file
+
+        // controls - list of form elements
+        // check all listed elements first, push to sourcedNamesVec as per spec
+        for child in controls.iter() {
+            if child.is_html_element() {
```
cagandhi: @jdm Is the function call `is_html_element()` correct or we should use `is_listed()`? We weren't able to call the method `is_listed_element()` as mentioned by you on the project page. `./mach check` was showing error for function call on incorrect variable type.

jdm: Ah, that's HTMLElement::is_listed_element. You can do:
```
if child.downcast::<HTMLElement>().map_or(false, |c| c.is_listed_element()) {
```

cagandhi: Done.

components/script/dom/htmlformelement.rs
```diff
@@ -253,6 +255,92 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                // if child.is_listed()
+
+                if child.has_attribute(&local_name!("id")) {
+                    // https://learning-rust.github.io/docs/b2.structs.html
+                    let entry = SourcedName {
+                        name: child.get_string_attribute(&local_name!("id")),
+                        element: child.root_element(),
```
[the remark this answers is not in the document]
cagandhi: Fixed it.

amj23897:
Hi @jdm, we are actually trying to implement the `NamedGetter` function since it modifies the `past names map` which is also used in the `SupportedPropertyNames` method. 

Following the spec, this is the code snippet we have come up with:
```
fn NamedGetter(&self, name: DOMString) -> Option<RadioNodeListOrElement> {
    // unimplemented!();
    let window = window_from_node(self);
    let mut candidates: RadioNodeList = RadioNodeList::new(&window, Dom<HTMLElement>);

    let controls = self.controls.borrow();

    for child in controls.iter() {
        if child.downcast::<HTMLElement>().map_or(false, |c| c.is_listed_element()) {
            if (child.has_attribute(&local_name!("id")) || (child.has_attribute(&local_name!("name")) && child.get_string_attribute(&local_name!("name")).to_string() == "name")) {
                // candidates.node_list.append(child);
                println!("Listed Element");
            }
        }
    }

    // https://developer.mozilla.org/en-US/docs/Web/API/NodeList - candidates.node_list.length
    // https://github.com/servo/servo/blob/master/components/script/dom/nodelist.rs#L116 - candidates.node_list.Length()
    if candidates.Length() == 0 {
        for child in controls.iter() {
            if child.is::<HTMLImageElement>() {
                if (child.has_attribute(&local_name!("id")) || (child.has_attribute(&local_name!("name")) && child.get_string_attribute(&local_name!("name")).to_string() == "name")) {
                    // candidates.node_list.append(child);
                    println!("Image Element");
                }
            }
        }
    }
    return candidates;
}
```

Below are the errors which we are encountering:

1. What should we pass in for the 2nd argument of `RadioNodeList::new()`? The error message says: `error: chained comparison operators require parentheses.` What should we modify?

    From the `src` file [radionodelist.rs](https://github.com/servo/servo/blob/master/components/script/dom/radionodelist.rs#L33), it seems that the 2nd argument will be the nodelist type. We have mentioned `Dom<HTMLElement>` since it is mentioned on the project page that the hashmap for `past names map` contains elements of type `Dom<HTMLElement>`.

2. Is the return type for `NamedGetter()` correct? If it is, how should we return the `candidates` radionodelist created? The error message is: `expected enum std::option::Option, found struct dom::radionodelist::RadioNodeList. 

3. How can we add the elements to the `candidates` radoionodelist? We weren't able to find a method that does this both in `radionodelist` and `nodelist` structs.

Thanks.

jdm:
> Below are the errors which we are encountering:
> 
>     1. What should we pass in for the 2nd argument of `RadioNodeList::new()`? The error message says: `error: chained comparison operators require parentheses.` What should we modify?

[RadioNodeList::new](https://github.com/servo/servo/blob/master/components/script/dom/radionodelist.rs#L33) accepts a NodeListType argument. Searching that file for NodeListType shows that it's defined in components/script/dom/nodelist.rs. Since the specification talks about creating a RadioNodeList that contains a particular set of elements, you will want to create a vector of the desired elements, then use [RadioNodeList::new_simple_list](https://github.com/servo/servo/blob/e70397d90aef6af1f14952cd22839bb0a4a1f072/components/script/dom/radionodelist.rs#L41) to create a RadioNodeList from the contents of that vector. There are other uses of this API elsewhere is components/script that you can refer to.

>        From the `src` file [radionodelist.rs](https://github.com/servo/servo/blob/master/components/script/dom/radionodelist.rs#L33), it seems that the 2nd argument will be the nodelist type. We have mentioned `Dom<HTMLElement>` since it is mentioned on the project page that the hashmap for `past names map` contains elements of type `Dom<HTMLElement>`.
> 
>     2. Is the return type for `NamedGetter()` correct? If it is, how should we return the `candidates` radionodelist created? The error message is: `expected enum std::option::Option, found struct dom::radionodelist::RadioNodeList.

The error message is correct - the method declaration says that it should expect an `Option<RadioNodeListOrElement>` value, but `return candidates` is a `RadioNodeList` object. There are two levels or type wrapping missing here - first you need to wrap the RadioNodeList value inside the [RadioNodeListOrElement](https://doc.servo.org/script/dom/bindings/codegen/UnionTypes/enum.RadioNodeListOrElement.html) enum, and then you need to wrap that inside of an [Option](https://doc.rust-lang.org/stable/std/option/enum.Option.html) enum.

>     3. How can we add the elements to the `candidates` radionodelist? We weren't able to find a method that does this both in `radionodelist` and `nodelist` structs.

Push them into the vector, then build the RadioNodeList from the complete vector. I recommend avoiding building the RadioNodeList value until you are just about to return it, since there are various points in the spec where we create new lists or return a specific element instead of a list.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
 
     // https://html.spec.whatwg.org/multipage/#the-form-element%3Adetermine-the-value-of-a-named-property
     fn NamedGetter(&self, name: DOMString) -> Option<RadioNodeListOrElement> {
-        // return Option::Some::<RadioNodeListOrElement>;
-        unimplemented!();
+        // unimplemented!();
+        let window = window_from_node(self);
+
+        // Q2. ERROR in the following declaration - check the attached screenshot (image2)
+        let mut candidates: Vec<Dom<HTMLElement>> = Vec::new();
```
jdm: This will need to be DomRoot, and store Node instead of HTMLElement.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+            if child
+                .downcast::<HTMLElement>()
+                .map_or(false, |c| c.is_listed_element())
+            {
+                if child.has_attribute(&local_name!("id")) ||
+                    (child.has_attribute(&local_name!("name")) &&
+                        child.get_string_attribute(&local_name!("name")) == name)
+                {
+                    // println!("{:?}", child.what_is_this());
+                    candidates.push(Dom::from_ref(&*child.downcast::<HTMLElement>().unwrap()));
```
jdm: This will need to use `child.upcast::<Node>()` instead.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                    // println!("{:?}", val.what_is_this());
+                    return Some(RadioNodeListOrElement::Element(DomRoot::from_ref(
+                        &*val.upcast::<Element>(),
+                    )));
+                }
+            }
+        }
+
+        if candidates.len() > 1 {
+            let mut candidatesDomRoot: Vec<DomRoot<Node>> = Vec::new();
```
jdm: This separate vector shouldn't be necessary.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+        if candidates.len() == 0 {
+            for (key, val) in past_names_map.iter() {
+                if *key == name {
+                    // println!("{:?}", val.what_is_this());
+                    return Some(RadioNodeListOrElement::Element(DomRoot::from_ref(
+                        &*val.upcast::<Element>(),
+                    )));
+                }
+            }
+        }
```
jdm: We'll need to return None here if there is no matching entry in past_names_map.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                    }
+                }
+            }
+        }
+
+        let mut past_names_map = self.past_names_map.borrow_mut();
+
+        if candidates.len() == 0 {
+            for (key, val) in past_names_map.iter() {
+                if *key == name {
```
jdm: It's a map, so we can use `past_names_map.get(&key)` instead of iterating through every entry.

components/script/dom/htmlformelement.rs
```diff
@@ -258,8 +262,83 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+            // Q1. ERROR in the following line - check the attached screenshot (image1)
+            // Related link for potential fix: https://stackoverflow.com/a/26953620/11978763
+            // we are unsure of how to incorporate this
+            return Some(RadioNodeListOrElement::RadioNodeList(
+                RadioNodeList::new_simple_list(&window, candidatesDomRoot.iter()),
+            ));
+        }
+
+        let element_node = &candidates[0];
+        if past_names_map.contains_key(&name) {
```
jdm: We should be able to call insert whether or not there is an existing entry in past_names_map.

jdm (changes requested):
No screenshots were attached anywhere that I could find.

components/script/dom/htmlformelement.rs
amj23897: image2:
image2

components/script/dom/htmlformelement.rs
amj23897: > No screenshots were attached anywhere that I could find.

Image 1 for Q1:

image1

components/script/dom/htmlformelement.rs
jdm: Yep, DomRoot should solve that.

components/script/dom/htmlformelement.rs
jdm: I think you need to pass into_iter() instead of iter() as the second argument to RadioNodeList::new_simple_list.

components/script/dom/htmlformelement.rs
cagandhi: If we change this vector to contain a `DomRoot<Node>` element, should we need to change the `Dom<HTMLElement>` in `past names map` declaration or we will need to downcast the `Node` to an `HTMLElement` when inserting an entry in `past names map`?

jdm: It should probably be enough to store `Dom<Element>` in the past names map, and that will require downcasting the node to Element, yes.

cagandhi: Done.

components/script/dom/htmlformelement.rs
cagandhi: Fixed.

components/script/dom/htmlformelement.rs
cagandhi: Removed this vector.

components/script/dom/htmlformelement.rs
cagandhi: Handled.

components/script/dom/htmlformelement.rs
cagandhi: Fixed.

components/script/dom/htmlformelement.rs
cagandhi: Done.

cagandhi:
@jdm For step 4 of the [supported property names spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names), we need to add all entries of `past names map` into `sourced names` vector. It is mentioned that we also need to add the length of time past entry has been in the `past names map` as age. 

How will we calculate this quantity since we do not store the time when an entry was added into the `past names map`?

jdm:
Ah, right. Let's make the past names map look like this: `past_names_map: DomRefCell<HashMap<DOMString, (Dom<Element>, time::Tm)>>,` where the second field of the tuple is the time when the element is inserted into the map (using [time::now()](https://doc.servo.org/time/fn.now.html)). Since the Tm struct implements the PartialOrd trait (https://doc.servo.org/time/struct.Tm.html), that means we can compare them and use them for sorting later.

amj23897:
> Ah, right. Let's make the past names map look like this: `past_names_map: DomRefCell<HashMap<DOMString, (Dom<Element>, time::Tm)>>,` where the second field of the tuple is the time when the element is inserted into the map (using [time::now()](https://doc.servo.org/time/fn.now.html)). Since the Tm struct implements the PartialOrd trait (https://doc.servo.org/time/struct.Tm.html), that means we can compare them and use them for sorting later.

Made the changes you suggested. Added `time::now()` call whenever an entry is added to `past names map`. However, when running `./mach check`, the following error is encountered:

Screenshot from 2019-12-05 23-25-54

jdm:
Add a `unsafe_no_jsmanaged_fields!(Tm);` to components/script/dom/bindings/trace.rs, and add `Tm` to [this line](https://github.com/servo/servo/blob/250184fe60388a001b9cc2146f95ed0bf8b59435/components/script/dom/bindings/trace.rs#L148).

amj23897:
> Add a `unsafe_no_jsmanaged_fields!(Tm);` to components/script/dom/bindings/trace.rs, and add `Tm` to [this line](https://github.com/servo/servo/blob/250184fe60388a001b9cc2146f95ed0bf8b59435/components/script/dom/bindings/trace.rs#L148).

It worked! Pushed a commit. Thanks. We'll be continuing to work on implementing SupportedPropertyNames [according to the specification](https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names).

components/script/dom/htmlformelement.rs
```diff
@@ -395,6 +397,28 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                element: DomRoot::from_ref(&*val.0),
+                source: SourcedNameSource::Past(now()-val.1), // calculate difference now()-val.1 to find age
+            };
+            sourcedNamesVec.push(entry);
+        }
+
+        // Step 5
+        // TODO need to sort as per spec. This is a partially implemented function.
+        // Kindly guide us on how to refine this function further.
+        sourcedNamesVec.sort_by(|a, b| a.name.partial_cmp(&b.name).unwrap());
```
jaymodi98: @jdm As suggested in the [project page](https://github.com/servo/servo/wiki/Missing-DOM-features-project) we tried to use `Node::CompareDocumentPosition`. But since its return type is `u16` and not `std::cmp::Ordering` we were getting error while trying to implement Step 5 [SupportedPropertyNames](https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names) as follows:

sourcedNamesVec.sort_by(|a, b| a.element.upcast::<Node>().CompareDocumentPosition(b.element.upcast::<Node>()));

Kindly help us how to sort `sourcedNamesVec` as mentioned in the spec.

jdm:
@bors-servo try=wpt

In the meantime let's check how the tests are doing.

amj23897:
> You will want to use the value returned by CompareDocumetPosition and map that into appropriate Ordering values. So for a that is earlier in tree order than node b, return Ordering::Less from the sort closure.

Done.

components/script/dom/htmlformelement.rs
```diff
@@ -395,6 +397,28 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+
+        // Step 6
+        sourcedNamesVec.retain(|sn| !sn.name.to_string().is_empty());
+
+        // Step 7
+        // Q1. Unable to clearly understand. It seems to contradict with step 4.
```
amj23897: @jdm Can you kindly explain how to implement step 7? It seems to be contradicting with step 4 in the [spec](https://html.spec.whatwg.org/multipage/forms.html#the-form-element:supported-property-names).

jdm: The idea here is that the list of names returned in step 8 should not contain any duplicates, and when removing any duplicates the first instance of the name should be retained.

jdm: It should be enough to add a check in the implementation of step 8 to only push the name if `namesVec.find(elem.name.clone()).is_none()` is true.

components/script/dom/htmlformelement.rs
```diff
@@ -253,6 +262,214 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+        let mut candidates: Vec<DomRoot<Node>> = Vec::new();
+
+        let controls = self.controls.borrow();
+        // Step 1
+        for child in controls.iter() {
+            if child
+                .downcast::<HTMLElement>()
+                .map_or(false, |c| c.is_listed_element())
+            {
+                if child.has_attribute(&local_name!("id")) ||
```
jdm: The specification text is ambiguous, but we actually want to verify that the id attribute's value equals name as well.

components/script/dom/htmlformelement.rs
```diff
@@ -253,6 +262,214 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                {
+                    candidates.push(DomRoot::from_ref(&*child.upcast::<Node>()));
+                }
+            }
+        }
+        // Step 2
+        if candidates.len() == 0 {
+            for child in controls.iter() {
+                if child.is::<HTMLImageElement>() {
+                    if child.has_attribute(&local_name!("id")) ||
```
jdm: Same as above.

jdm (changes requested):

components/script/dom/htmlformelement.rs
jaymodi98: Done.

components/script/dom/htmlformelement.rs
jaymodi98: Done.

components/script/dom/htmlformelement.rs
jaymodi98: Done.

jdm:
```
  ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
  └ PASS [expected FAIL] Calling item() on the NodeList returned from the named getter should work
  ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
  └ PASS [expected FAIL] Indexed getter on the NodeList returned from the named getter should work
  ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
  └ PASS [expected FAIL] The form attribute should be taken into account for named getters (single element)
  ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
  └ PASS [expected FAIL] Input should only be a named property on the innermost form that contains it
  ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
  └ PASS [expected FAIL] Trying to set an expando that shadows a named property that gets added later
  ▶ Unexpected subtest result in /html/semantics/forms/the-fieldset-element/HTMLFieldSetElement.html:
  └ PASS [expected FAIL] The controls must root at the fieldset element
```
You should run `./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms --log-raw wpt.log` and then `./mach update-wt wpt.log`, and commit the changes. That will update the expected test results to reflect that these tests now pass as a result of these changes.

cagandhi:
> ```
>   ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
>   └ PASS [expected FAIL] Calling item() on the NodeList returned from the named getter should work
>   ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
>   └ PASS [expected FAIL] Indexed getter on the NodeList returned from the named getter should work
>   ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
>   └ PASS [expected FAIL] The form attribute should be taken into account for named getters (single element)
>   ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
>   └ PASS [expected FAIL] Input should only be a named property on the innermost form that contains it
>   ▶ Unexpected subtest result in /html/semantics/forms/the-form-element/form-nameditem.html:
>   └ PASS [expected FAIL] Trying to set an expando that shadows a named property that gets added later
>   ▶ Unexpected subtest result in /html/semantics/forms/the-fieldset-element/HTMLFieldSetElement.html:
>   └ PASS [expected FAIL] The controls must root at the fieldset element
> ```
> 
> You should run `./mach test-wpt tests/wpt/web-platform-tests/html/semantics/forms --log-raw wpt.log` and then `./mach update-wt wpt.log`, and commit the changes. That will update the expected test results to reflect that these tests now pass as a result of these changes.

Updated the test log and pushed the changes.

jdm:
I amended the latest commit with corrected test results and removed the 2mb wpt.log file.

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24762

amj23897:
Thank you for working with us on this and getting the pull request merged. 

We looked at the code diffs and it seems that certain tests expected error however they were getting timed out. Could you explain to us what these specific tests were?

jdm:
Those are tests that are known to time out intermittently in our current implementation: https://github.com/servo/servo/issues/24501 . It's unlikely that these changes actually triggered a change in behaviour.

PR #25117, opened 2019-12-05
Move WPT sync-from-upstream task to Linux

SimonSapin:
Before this PR, this task keeps one macOS workers busy for many hours every day.

### DIFF
--- a/etc/ci/update-wpt-checkout
+++ b/etc/ci/update-wpt-checkout
@@ -67,7 +67,7 @@ function cleanup() {
 # Build Servo and run the full WPT testsuite, saving the results to a log file.
 function unsafe_run_tests() {
     # Run the full testsuite and record the new test results.
-    ./mach test-wpt --release --processes 6 --log-raw "${1}" \
+    ./mach test-wpt --release --processes 12 --log-raw "${1}" \
            --always-succeed || return 1
 
     # Run the bluetooth testsuite, which uses the webdriver test harness.
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -579,28 +579,22 @@ def macos_nightly():
 
 
 def update_wpt():
-    build_task = macos_release_build_with_debug_assertions()
-    update_task = (
-        macos_task("WPT update")
-        .with_python2()
-        .with_treeherder("macOS x64", "WPT update")
+    build_task = linux_release_build_with_debug_assertions(layout_2020=False)
+    return (
+        linux_task("WPT update")
+        .with_treeherder("Linux x64", "WPT update")
+        .with_dockerfile(dockerfile_path("run"))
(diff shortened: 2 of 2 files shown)


SimonSapin:
Because this task is not used by Homu, testing it requires some manual steps:

* On top of commits destined to go into a PR, make a temporary commit that:
  - Disables irrelevant daily tasks in `etc/taskcluster/decision_task.py`
  - Disables auto-`r+` in `etc/ci/update-wpt-checkout`
  - Add `[Do not merge]` in the title of the PR created by `etc/ci/update-wpt-checkout`
* Go to https://community-tc.services.mozilla.com/hooks/project-servo/daily
* Click on one of the task IDs under "Last Fired Attempts"
* Click on "View task definition"
* Copy that JSON blob
* Go to https://community-tc.services.mozilla.com/tasks/create
* Paste the JSON blob in the editor
* Click on "Update timestamps" in the bottom-right menu. This also reformats the definition to JSON.
* Edit `GIT_REF` in the task definition to use the branch pushed previously.
* Click the "Create task" button at the bottom left
* Wait 3 to 4 hours

[The first attempt](https://community-tc.services.mozilla.com/tasks/FJxuQpwbTDyBYr9gva1wUw/runs/0) (that didn’t fail) resulted in https://github.com/servo/servo/pull/25086, which looks similar to https://github.com/servo/servo/pull/25077 which ran on macOS around the same time. Both PRs change many files, presumably because upstream WPT had many changes since the last sync.

To eliminate that source of change, https://github.com/servo/servo/pull/25115 was made by [another run](https://community-tc.services.mozilla.com/tasks/FUnhRfE3TOuaEEX2fUFOSQ/runs/0) that also disabled syncing from upstream WPT. It touches 54 `.ini` file. That number would be zero if Servo was perfectly deterministic and behaved the same on Linux v.s. macOS.

I’ve started [another run](https://community-tc.services.mozilla.com/tasks/FUnhRfE3TOuaEEX2fUFOSQ/runs/1) with the same configuration to see if the results look different. jdm says on IRC that those 54 are plausibly all due to intermittent test results, though.

SimonSapin:
Uh, it looks like https://github.com/servo/servo/pull/25115/files may have been overwritten by the actual daily task (since the branch name is based only on the date.) Anyway, https://github.com/servo/servo/pull/25122/commits/5aa458e20dbd87b981d3c321a222e5f77b58e644 also changes 53 `.ini` files and looks similar to my (vague) memory of the other diff.

If some of these are deterministic but due to differences between Linux and macOS, do we risk having macOS CI start failing after a PR like this is merged? @jdm, what do you think?

jdm:
I don't see how it's any different from the current situation - linux CI has to pass for any new sync PR when the mac machines generate them, this is just the reverse. I think this is fine.

PR #25048, opened 2019-12-03
GStreamer plugin should use GLMemory

asajeffrey:
Get the gstreamer servosrc plugin to generate frames in GLMemory rather than main memory.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #24831 
- [x] These changes do not require tests because it's an embedding perf issue

PR #25139, opened 2019-12-05
Use a glmemory bufferpool

### LINKED ISSUE #25121: GStreamer plugin wrong memory type
With #25048, the plugin sometimes fails due to receiving regular memory rather than GL memory to fill. This is caused by the plugin not setting up a buffer pool (which in GStreamer is the source's responsibility, not the sink's).

asajeffrey:
This ensures the gstreamer plugin is always using GLMemory.

---

- [x] These changes fix #25121

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4789,6 +4789,7 @@ dependencies = [
  "gstreamer-base",
  "gstreamer-gl",
  "gstreamer-gl-sys",
+ "gstreamer-sys",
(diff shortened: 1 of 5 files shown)


asajeffrey:
The first commit is #25048.

asajeffrey:
#25048 has landed, so this is ready for review. r? @ferjm

jdm:
@bors-servo retry
* #24726 #24762

PR #25164, opened 2019-12-06
Increase parallelism on Linux WPT testing

SimonSapin:
The time taken by each chunk is uneven, with WPT-1 the longest before this change at 30 ~ 45 minutes. This reduces it to 15 ~ 20 minutes.

Surprisingly, increasing the number of processes seems to also make `test_element_in_collection` in `/webdriver/tests/execute_script/cyclic.py` **unexpectedly pass**. This happened reliably in three different runs:

https://community-tc.services.mozilla.com/tasks/S9O27WJvSa6j2PSjcRcbBA/runs/2


SimonSapin:
Failures include a network error during a rustup download, and one instance of https://github.com/servo/servo/issues/24726. But WPT chuck 0, which contains `/webdriver/tests/execute_script/cyclic.py`, is green on both Linux and macOS with the conditional expectation.

r? @jdm

jdm:
@bors-servo retry
* #23290

Issue #24788, opened 2019-11-20
WebGL Error 1282 on latest three.js examples


Manishearth:
With #23545 rebased on master (https://github.com/Manishearth/servo/tree/modules), I get the following error on all of the three.js examples (the latest version). The modules seem to be loading, and I get a flash of rendering before the assertion failure

```
assertion failed: `(left == right)`
  left: `1282`,
 right: `0` (thread WebGL thread, at components/canvas/webgl_thread.rs:1006)
```

cc @jdm

jdm:
What are some URLs that reproduce this? I tried http://rawcdn.githack.com/mrdoob/three.js/master/examples/webvr_rollercoaster.html (`assertion failed: fetched_module_object.is_some()`) and http://rawcdn.githack.com/mrdoob/three.js/master/examples/webgl_animation_cloth.html (segfault inside mozjs from CompileModule).

jdm:
I should note that I've been trying to reproduce this on macOS with my patch from https://github.com/servo/servo/pull/23545#issuecomment-556063220 to fix the assertion failure and segfault and haven't seen this problem yet.

Manishearth:
I get the same errors on the rawcdn githack things, but not on threejs.org

```
[22:05:14] मanishearth@noether ~/mozilla/servo ^_^ 
$ RUST_BACKTRACE=1 ./mach run -d https://threejs.org/examples/webvr_rollercoaster.html
module script of https://threejs.org/examples/webvr_rollercoaster.html compile done
Start to load dependencies of https://threejs.org/examples/webvr_rollercoaster.html
Requested url: https://threejs.org/examples/jsm/vr/WebVR.js
Start to fetch https://threejs.org/examples/jsm/vr/WebVR.js
Requested url: https://threejs.org/examples/jsm/misc/RollerCoaster.js
Start to fetch https://threejs.org/examples/jsm/misc/RollerCoaster.js
Requested url: https://threejs.org/build/three.module.js
Start to fetch https://threejs.org/build/three.module.js
module script of https://threejs.org/examples/jsm/misc/RollerCoaster.js compile done
Start to load dependencies of https://threejs.org/examples/jsm/misc/RollerCoaster.js
Requested url: https://threejs.org/build/three.module.js
See existing descendant: https://threejs.org/build/three.module.js
Visited array of requested url https://threejs.org/build/three.module.js: it's visited true and status is Fetching
module script of https://threejs.org/examples/jsm/vr/WebVR.js compile done
[...]

PR #25188, opened 2019-12-06
Add extra debug assertions for WebGL errors

asajeffrey:
Add some extra debug assertions checking for GL errors, so we'll get a bit more help tracking down issues like #24788 

---

### DIFF
--- a/components/canvas/webgl_thread.rs
+++ b/components/canvas/webgl_thread.rs
@@ -680,10 +680,14 @@ impl WebGLThread {
             )
             .expect("Where's the GL data?");
 
+            // Ensure there are no pending GL errors from other parts of the pipeline.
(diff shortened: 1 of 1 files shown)


components/canvas/webgl_thread.rs
```diff
@@ -680,10 +680,14 @@ impl WebGLThread {
             )
             .expect("Where's the GL data?");
 
+            // Ensure there are no pending GL errors from other parts of the pipeline.
```
Manishearth: Perhaps this comment could be a part of the assert string?

Manishearth (approved):
r=me, though if we can add assertion strings to the asserts that would be nice

asajeffrey:
My experience is that assert strings are less use than the line numbers. YMMV. @bors-servo r=Manishearth

jdm:
@bors-servo retry
* #25216

PR #25204, opened 2019-12-07
Link to Gecko’s reftest analyzer

SimonSapin:
http://hoppipolla.co.uk/410/reftest-analyser-structured.xhtml seems to be gone


SimonSapin:
@jgraham, does this look right? Or is there some other version of the analyzer we should use?

SimonSapin:
Amended per IRC conversation https://mozilla.logbot.info/servo/20191209#c16806476-c16806485 to revert back to using `--log-raw` format, since that’s what our CI produces.

PR #25207, opened 2019-12-08
min/max-width/height, replaced elements

SimonSapin:


components/style/values/generics/length.rs
```diff
@@ -207,6 +207,15 @@ impl<LengthPercentage> MaxSize<LengthPercentage> {
     pub fn none() -> Self {
         MaxSize::None
     }
+
+    /// Convert
+    #[cfg(not(feature = "gecko"))]
+    pub fn to_option(self) -> Option<LengthPercentage> {
```
nox: Nit: Just name that `length`, like `Result::ok`?

components/layout_2020/sizing.rs
```diff
@@ -107,17 +107,37 @@ impl BoxContentSizes {
...
+            .max_box_size()
+            .inline
+            .to_option()
+            .and_then(|lp| lp.as_length());
+        let clamp = |l: Length| l.clamp_between_extremums(min_inline_size, max_inline_size);
+
         // Percentages for 'width' are treated as 'auto'
         let inline_size = inline_size.map(|lp| lp.as_length());
         // The (inner) min/max-content are only used for 'auto'
         let mut outer = match inline_size.non_auto().flatten() {
```
nox: Can't you use `and_then` to define `inline_size` and then remove the `flatten` call?

components/layout_2020/dom_traversal.rs
```diff
@@ -299,7 +298,7 @@ impl Drop for BoxSlot<'_> {
 pub(crate) trait NodeExt<'dom>: 'dom + Copy + LayoutNode + Send + Sync {
     fn is_element(self) -> bool;
     fn as_text(self) -> Option<String>;
-    fn as_image(self) -> Option<(Option<Arc<NetImage>>, Vec2<Length>)>;
+    fn as_image(self) -> Option<(Option<Arc<NetImage>>, Vec2<f64>)>;
```
nox: Why the type change, and why not `CSSFloat`?

components/layout_2020/replaced.rs
```diff
@@ -66,12 +93,21 @@ impl ReplacedContent {
...
+    ) -> Vec2<Length> {
         let mode = style.writing_mode();
-        // FIXME(nox): We shouldn't pretend we always have a fully known intrinsic size.
-        let intrinsic_size = self.intrinsic_size.size_to_flow_relative(mode);
-        // FIXME(nox): This can divide by zero.
-        let intrinsic_ratio = intrinsic_size.inline.px() / intrinsic_size.block.px();
+        let intrinsic_size = physical::Vec2 {
+            x: self.intrinsic_width,
+            y: self.intrinsic_height,
+        };
```
nox: Why not store the intrinsic size as a `physical::Vec2<Option<_>>`?

components/layout_2020/replaced.rs
```diff
@@ -82,25 +118,83 @@ impl ReplacedContent {
...
+            //  than 150px, and has a width not greater than the device width.”
+            physical::Vec2 {
+                x: Length::new(300.),
+                y: Length::new(150.),
+            }
+            .size_to_flow_relative(mode)
+        };
+        let clamp = |inline_size: Length, block_size: Length| Vec2 {
+            inline: inline_size.clamp_between_extremums(min_box_size.inline, max_box_size.inline),
+            block: block_size.clamp_between_extremums(min_box_size.block, max_box_size.block),
```
nox: Wondering if we should have a meethod for that, don't we do that for non-replaced things elsewhere in the code?

components/layout_2020/replaced.rs
```diff
@@ -82,25 +118,83 @@ impl ReplacedContent {
...
         };
         // https://drafts.csswg.org/css2/visudet.html#min-max-widths
         // https://drafts.csswg.org/css2/visudet.html#min-max-heights
-        let (inline_size, block_size) = match (box_size.inline, box_size.block) {
+        match (box_size.inline, box_size.block) {
```
nox: Returning `Vec2<_>` values made the code more verbose, IMO, but it's not very important.

components/layout_2020/replaced.rs
```diff
@@ -82,25 +118,83 @@ impl ReplacedContent {
...
             (LengthOrAuto::Auto, LengthOrAuto::Auto) => {
+                let inline_size =
+                    match (intrinsic_size.inline, intrinsic_size.block, intrinsic_ratio) {
+                        (Some(inline), _, _) => inline,
+                        (None, Some(block), Some(i_over_b)) => {
+                            // “used height” in CSS 2 is always gonna be the intrinsic one,
+                            // since it is available.
+                            block * i_over_b
+                        },
+                        // (Vec2 { inline: None, block: None }, Some(_)) => {
```
nox: Commented code that is out-of-date, who would have known? :D Please remove.

components/layout_2020/replaced.rs
```diff
@@ -82,25 +118,83 @@ impl ReplacedContent {
...
         // https://drafts.csswg.org/css2/visudet.html#min-max-widths
         // https://drafts.csswg.org/css2/visudet.html#min-max-heights
-        let (inline_size, block_size) = match (box_size.inline, box_size.block) {
+        match (box_size.inline, box_size.block) {
             (LengthOrAuto::LengthPercentage(inline), LengthOrAuto::LengthPercentage(block)) => {
                 clamp(inline, block)
             },
             (LengthOrAuto::LengthPercentage(inline), LengthOrAuto::Auto) => {
-                clamp(inline, inline / intrinsic_ratio)
+                let block = if let Some(i_over_b) = intrinsic_ratio {
```
nox: This name confuses me, please just name it `ratio`.

components/layout_2020/replaced.rs
```diff
@@ -63,6 +64,41 @@ impl ReplacedContent {
...
+            y: self.intrinsic_height,
+        };
+        intrinsic_size.size_to_flow_relative(style.writing_mode())
+    }
+
+    fn inline_size_over_block_size_intrinsic_ratio(
+        &self,
+        style: &ComputedValues,
+    ) -> Option<CSSFloat> {
+        self.intrinsic_ratio.map(|width_over_height| {
```
nox: Sorry, but again, I find the code more readable when that is just named `ratio`.

components/layout_2020/positioned.rs
```diff
@@ -158,7 +158,30 @@ impl<'a> AbsolutelyPositionedFragment<'a> {
...
         let cbbs = containing_block.size.block;
 
-        let box_size = style.box_size();
+        let size;
+        let replaced_used_size;
+        match self.absolutely_positioned_box.contents.as_replaced() {
+            Ok(replaced) => {
+                // https://drafts.csswg.org/css2/visudet.html#abs-replaced-width
+                // https://drafts.csswg.org/css2/visudet.html#abs-replaced-height
+                let u = replaced.used_size_as_if_inline_element(&containing_block.into(), style);
```
nox: Not sur I like `u`.

components/layout_2020/positioned.rs
```diff
@@ -235,11 +235,12 @@ impl<'a> AbsolutelyPositionedFragment<'a> {
...
-                    let mut margin_start = computed_margin_start.auto_is(Length::zero);
-                    let mut margin_end = computed_margin_end.auto_is(Length::zero);
-
-                    let size = if let LengthOrAuto::LengthPercentage(size) = size {
-                        let margins = containing_size - start - end - padding_border_sum - size;
+                    let margin_start;
+                    let margin_end;
+                    let used_size;
+                    if let LengthOrAuto::LengthPercentage(s) = size {
+                        used_size = s;
```
nox: Name `size` was just fine, why shorten it?

components/layout_2020/positioned.rs
```diff
@@ -208,7 +208,7 @@ impl<'a> AbsolutelyPositionedFragment<'a> {
             padding_border_sum: Length,
             computed_margin_start: LengthOrAuto,
             computed_margin_end: LengthOrAuto,
-            solve_margins: impl FnOnce(Length) -> (Length, Length),
+            avoid_negative_margin_start: bool,
```
nox: Please remove this change or introduce yet another new type for this boolean.

nox (commented):
Lots of nits but otherwise nothing to say.

components/style/values/generics/length.rs
SimonSapin: This is subtle but I mean this method not as "if this enum is the `LengthPercentage` variant, access its field" but rather "losslessly convert all the information in this enum". If/when we add support for e.g. `min-width: min-content` https://drafts.csswg.org/css-sizing/#min-size-properties this enum will have additional variants like it does today for Gecko, this method will need to be removed, and the callers should be modified to handle those variants separately.

components/layout_2020/sizing.rs
SimonSapin: The call two lines above is `LengthPercentageOrAuto::map`, not `Option::map`. Maybe we should have `LengthPercentageOrAuto::and_then` that return `Auto` when its closure returns `Option::None`?

components/layout_2020/dom_traversal.rs
SimonSapin: It’s not `Length` because it is measured in image pixels rather than CSS `px`.

I wanted to make it `u32` but it’s already divided by `script::dom::htmlimageelement::ImageRequest::current_pixel_density`. I’m not sure what that density means exactly or if diving there is correct.

It could be `CSSFloat`. It probably wouldn’t affect much since it’s not kept in memory for a long time.

components/layout_2020/replaced.rs
SimonSapin: Could be either. An earlier version of the comment on the field declaration talked about "all three components"

components/layout_2020/replaced.rs
SimonSapin: I grepped through the code and didn’t find another instance of computing both at the same time.

In block layout for example, we compute the used width before recurring and the used height after.

components/layout_2020/replaced.rs
SimonSapin: For some reason I thought this change helped having an early return, but it looks like it wasn’t necessary after all.

components/layout_2020/replaced.rs
SimonSapin: What do you mean, out of date? The next line is a FIXME comment about a part of the spec that is not implemented yet, and this line has the match arm that we will need to use when we implement that part. I’ll shuffle this so it doesn’t look like comment-out code:

```rust
// FIXME
//
// “If 'height' and 'width' both have computed values of 'auto'
// and the element has an intrinsic ratio but no intrinsic height or width,
// […]”
//
// In this `match` expression this would be an additional arm here:
//
// ```
// (Vec2 { inline: None, block: None }, Some(_)) => {…}
// ```
//
// “[…] then the used value of 'width' is undefined in CSS 2.
// However, it is suggested that, if the containing block's width
// does not itself depend on the replaced element's width,
// then the used value of 'width' is calculated from the constraint
// equation used for block-level, non-replaced elements in normal flow.”

components/layout_2020/replaced.rs
SimonSapin: The problem these new names try to address is that when A divided by B is called a ratio, B divided by A is also called a ratio although they’re not interchangeable. Code that uses that ratio sometimes to multiple and sometimes to divide a length, having a variable name that indicates the “direction” of the ratio helps check if those are correct by doing pseudo-dimensional analysis.

We could declare that an aspect ratio is by convention always a width divided by a height. But this code work in flow-relative dimensions and accesses a ratio field in physical dimensions. I’ve added `inline_size_over_block_size_intrinsic_ratio` to make that conversion.

I’m open to other names, but I’d prefer if they indicate the “direction” of the ratio somehow. Or maybe we could provide that indication in doc-comments and code comments?

components/layout_2020/positioned.rs
SimonSapin: Even with exactly two call sites, that both have a comment to repeat the name of the parameter?

SimonSapin:
This conflicts with https://github.com/servo/servo/pull/25203, I’ll push a rebase + changes after that one lands.

components/layout_2020/replaced.rs
nox: Fair enough, I don't mind the name.

components/layout_2020/positioned.rs
nox: IIRC the collapsing margin stuff doesn't have more callers than that, and the comment can go if you introduce a type, but as I said on IRC, all my nits are optional.

PR #25211, opened 2019-12-08
`worker_id` type changed to uuid

### LINKED ISSUE #6631: Worker ids aren't unique
The devtools-related code for notifying when new worker globals are created won't work correctly in the presence of subworkers. Each WorkerGlobalScope/Window has its own next_worker_id field, so it's easy to end up with overlapping ids for independent workers.

kunalmohan:
Fixes #6631
`worker_id` is now generate as uuid and saved as string.

---

- [X] These changes fix #6631  (GitHub issue number if applicable)

### DIFF
--- a/components/devtools_traits/Cargo.toml
+++ b/components/devtools_traits/Cargo.toml
@@ -20,3 +20,4 @@ msg = {path = "../msg"}
 serde = "1.0"
 servo_url = {path = "../url"}
 time = "0.1"
+uuid = {version = "0.8", features = ["v4", "serde"]}
--- a/components/devtools_traits/lib.rs
+++ b/components/devtools_traits/lib.rs
@@ -25,7 +25,7 @@ use msg::constellation_msg::PipelineId;
 use servo_url::ServoUrl;
 use std::net::TcpStream;
 use time::{self, Duration, Tm};
-//use uuid::Uuid;
+use uuid::Uuid;
 
 // Information would be attached to NewGlobal to be received and show in devtools.
 // Extend these fields if we need more information.
@@ -357,5 +357,5 @@ impl PreciseTime {
     }
 }
 
-#[derive(Clone, Debug, Deserialize, Eq, Hash, MallocSizeOf, PartialEq, Serialize)]
-pub struct WorkerId(pub String);
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, MallocSizeOf, PartialEq, Serialize)]
+pub struct WorkerId(pub Uuid);
--- a/components/devtools_traits/lib.rs
+++ b/components/devtools_traits/lib.rs
@@ -25,6 +25,7 @@ use msg::constellation_msg::PipelineId;
 use servo_url::ServoUrl;
 use std::net::TcpStream;
 use time::{self, Duration, Tm};
+//use uuid::Uuid;
 
 // Information would be attached to NewGlobal to be received and show in devtools.
 // Extend these fields if we need more information.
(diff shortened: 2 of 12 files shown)


components/devtools_traits/lib.rs
```diff
@@ -359,5 +360,5 @@ impl PreciseTime {
     }
 }
 
-#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, MallocSizeOf, PartialEq, Serialize)]
-pub struct WorkerId(pub u32);
+#[derive(Clone, Debug, Deserialize, Eq, Hash, MallocSizeOf, PartialEq, Serialize)]
+pub struct WorkerId(pub String);
```
paulrouget: You can keep uuid and implement malloc_size_of, or maybe use u128 (https://docs.rs/uuid/0.8.1/uuid/struct.Uuid.html#method.to_u128_le).

components/script/dom/globalscope.rs
```diff
@@ -94,7 +95,7 @@ impl Drop for AutoCloseWorker {
 pub struct GlobalScope {
     eventtarget: EventTarget,
     crypto: MutNullableDom<Crypto>,
-    next_worker_id: Cell<WorkerId>,
+    next_worker_id: WorkerId,
```
paulrouget: I don't think that field is necessary anymore. Just create a new uuid when necessary.

kunalmohan:
Ok. I'll start working on these changes.
A small query (and probably a silly one)- malloc_size_of for UUID should return 0, right?

paulrouget:
> malloc_size_of for UUID should return 0, right?

I don't know. @jdm ?

jdm:
Yes, that's fine.

kunalmohan:
I have changed `worker_id` to uuid and implemented malloc_size_of.

components/devtools_traits/Cargo.toml
```diff
@@ -20,3 +20,4 @@ msg = {path = "../msg"}
 serde = "1.0"
 servo_url = {path = "../url"}
 time = "0.1"
+uuid = {version = "0.8", features = ["v4", "serde"]}
```
paulrouget: Is `"serde"` necessary?

components/malloc_size_of/Cargo.toml
```diff
@@ -47,3 +47,4 @@ url = { version = "2.0", optional = true }
 webrender_api = { git = "https://github.com/servo/webrender", optional = true }
 xml5ever = { version = "0.16", optional = true }
 void = "1.0.2"
+uuid = {version = "0.8", features = ["v4", "serde"]}
```
paulrouget: Is `"serde"` necessary?

components/devtools_traits/Cargo.toml
kunalmohan: Since we need to implement Serialize and Deserialize traits on struct `WorkerId`, serde will be required.

components/malloc_size_of/Cargo.toml
kunalmohan: I guess I can remove serde here.

jdm: Please make this optional and enable it as part of the servo feature.

components/malloc_size_of/lib.rs
```diff
@@ -75,6 +75,7 @@ extern crate thin_slice;
 extern crate time;
 #[cfg(feature = "url")]
 extern crate url;
+extern crate uuid;
```
jdm: Please hide all uses of uuid in this file behind the servo feature.

jdm (changes requested):

components/malloc_size_of/lib.rs
kunalmohan: Done

components/malloc_size_of/Cargo.toml
kunalmohan: I have removed `serde` here and made the other changes too.

jdm:
@bors-servo retry
* #24726

CYBAI:
@bors-servo retry

-  #24762

jdm:
@bors-servo retry
* #22212

PR #25214, opened 2019-12-09
Basic IndexedDB Functionality

### LINKED ISSUE #6963: Implement IndexedDB.
(Just filing as a placeholder.)

rasviitanen:
This PR consists of basic functionality for `IndexedDB`.
The database engine used is `Rkv`, and might not be the most optimal engine to use, so I created a *pluggable* engine. This makes it easy to change for eventual benchmarks etc.
Recently, `Rkv` has raised [warnings](https://github.com/mozilla/rkv) for using it in production, and the plan seems to be to change the backend to sqlite instead. If we expect that IndexedDB is finished after the issues in `Rkv` has been fixed, we should be able to use it after all. The ergonomics of Rkv is really good, and suits IndexedDB quite well.

What works:
* Basic Open
* Basic Put
* Basic Add
* Basic Get
* Basic Remove

Some things that don't work:
* Key generators
* Key ranges
* Aborting/Reverting transactions, including upgrades
* Closing/Removing a store/db
* Injection of keys
* Proper scheduling of transactions (they are started immediately atm)
* IDBIndex
* IDBCursor
* Manually committing transactions
* Some Error handling is missing
* ...  lots of other stuff

Implementation details worth noting:
* Transactions run in a custom thread pool. `read` and `write` transactions use the same pool, which in the future probably should be separate.
* IndexedDB is disabled by default, as it has a lot of troubles

I am making this PR pre-prematurely, as it is part of a course I take in school (and it's soon due date), so I would be very happy if *reviews can be done rather quickly*. 

---

- [X] These changes fix (PARTLY) #6963

- [X] These changes do not require tests because:
There are many WebIDL tests that we can run to test the code. I was not able to run wpt tests with logging on windows, so test metadata is currently missing. I have, however, run some tests manually, and we can expect some of the tests to `PASS`.
Some examples of passing tests:
`idbobjectstore_put.htm`
`idbobjectstore_put2.htm`
`idbobjectstore_put3.htm`
`idbobjectstore_put5.htm`
`idbobjectstore_put7.htm`
`idbobjectstore_put9.htm`,
`idbobjectstore_put10.htm`
`idbobjectstore_put11.htm`
...
`idbfactory_open.htm`
`idbfactory_open2.htm`
`idbfactory_open3.htm`
`idbfactory_open4.htm`
...


jdm:
Please add a [__dir__.ini](https://github.com/servo/servo/blob/master/tests/wpt/webgl/meta/conformance2/__dir__.ini) to the indexeddb metadata directory to enable the new preference. This will make it possible for us to run the tests on CI and get the updated metadata you're looking for!

jdm:
Run `wget https://community-tc.services.mozilla.com/api/queue/v1/task/P8Qk0aigQUKtiQRajoTYzg/runs/0/artifacts/public/test-wpt.log` and then `./mach update-wpt test-wpt.log` to update the metadata automatically.

rasviitanen:
Huh, those are not the test results I was expecting. Looking at the test harness and databases that are created for each test, it looks like we must be able to create multiple databases with different names.

`dbname = (dbname ? dbname : "testdb-" + new Date().getTime() + Math.random() );`

It looks like this doesn't work at the moment, and everything essentially operates on the same database. I assume this is why e.g. `idbobjectstores_add.htm` fails.  This probably explains why I have been able to get a lot more passes when running one test at a time. I think I just forgot to add a call to `Open`  and it shouldn't be too difficult to fix. 

Hopefully, we'll get more sensical test results once it's fixed.

components/script/dom/idbrequest.rs
```diff
@@ -0,0 +1,273 @@
...
+                operation,
+            ))
+            .unwrap();
+
+        let (task_source, canceller) = window
+            .task_manager()
+            .dom_manipulation_task_source_with_canceller();
+
+        let trusted_request = Trusted::new(&*request);
+        Builder::new()
```
gterzian: For this it could be convenient to use the IPC router, as is done for example at https://github.com/servo/servo/blob/2cf9a00c9983bf23eff23d56321973d36a14f977/components/script/image_listener.rs#L33

components/script/dom/idbfactory.rs
```diff
@@ -0,0 +1,227 @@
...
+            .task_manager()
+            .dom_manipulation_task_source_with_canceller();
+
+        let this = Trusted::new(self);
+        let trusted_request = Trusted::new(&*request);
+        let name_as_string = name.to_string(); // DOMString is not Send
+        Builder::new()
+            .name("IndexedDBOpenDatabase".to_owned())
+            .spawn(move || {
+                // FIXME:(rasviitanen) This should not be queued as a task,
```
gterzian: Ah ok so I understand a bit better why you use a thread here, and I think you want to restructure this a bit. 

In general when the spec says to do something "in parallel", in this context it probably means doing it somewhere inside `net`. Then you want to setup a route using the `ROUTER` to get the result from that parallel operation over IPC, and then queue a task from the router thread to the event-loop to have that result affect the world of javascript. 

So what you're doing here is running `open_internal` on the event-loop, effectively blocking it while waiting for the IPC reply from `net`, and then you message back to this thread. This thread is indeed running "in-parallel" to the event-loop, however the blocking `create_database` task is then run on the event-loop which defeats the purpose of "in-parallel" steps. 

So everytime you do a `recv()` on an ipc channel here in `script`, I think you actually want to use the `ROUTER`, do the "in-parallel" stuff back in `net`, and then reply back to the route, and finally queue a task with the result(Unless the spec really means to make an operation blocking)

components/net/indexeddb/thread_pool.rs
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use crossbeam_channel::{self, Receiver, Sender};
+use std::thread;
+
+#[derive(Clone)]
+pub struct TransactionPool {
```
gterzian: For this you can consider using a pool from `rayon`, fetch for example uses one at https://github.com/servo/servo/blob/4d5bf653f72a4ebf02c627eda7c5fa7f3744cbe8/components/net/resource_thread.rs#L496

components/net/indexeddb/idb_thread.rs
```diff
@@ -0,0 +1,228 @@
...
+            .spawn(move || {
+                IndexedDBManager::new(port, RkvEngine::new(idb_dir.as_path())).start();
+            })
+            .expect("Thread spawning failed");
+
+        chan
+    }
+}
+
+struct IndexedDBManager<E: KvsEngine> {
```
gterzian: At a very high level, the spec reads that:

> Each origin has an associated set of databases. A database has zero or more object stores which hold the data stored in the database. http://w3c.github.io/IndexedDB/#ref-for-database

So I'm not sure how this would be best organized in terms of threading and underlying engines, and I do think you want a map somewhere of `ImmutableOrigin` -> `Database`. 

In script, to get the origin from which you're calling you can usually do `self.global().origin().immutable()`, which you could send along each message so that the manager appropriately route the request to the right database.

rasviitanen: Im working on a fix for this, this was done quite stupidly, and think I have figured out a really nice way to restructure this to allow multiple databases. The stores don't have to be namespaced once I am done with it either, so it will solve quite a bunch of issues

components/script/dom/idbfactory.rs
```diff
@@ -0,0 +1,227 @@
...
+            upgrade_notificator.send(Ok(())).unwrap();
+            result_channel.send(Err(Error::Version)).unwrap();
+        }
+    }
+}
+
+impl IDBFactoryMethods for IDBFactory {
+    // https://www.w3.org/TR/IndexedDB-2/#dom-idbfactory-open
+    #[allow(unsafe_code)]
+    fn Open(&self, name: DOMString, version: Option<u64>) -> Fallible<DomRoot<IDBOpenDBRequest>> {
```
gterzian: Ok so to be a bit clearer on what I meant. Here is how you could consider doing this:

1. let `request` be a new `IDBOpenDBRequest`.
2. create an ipc channel.
3. Setup a route with the receiver of that channel.
4. Send the sender inside a message to the indexDB component.
5. Create some sort of struct like:
```rust
struct OpenListener {
    open_request: Trusted<`IDBOpenDBRequest`>
    global: Trusted<Global>,
}

impl OpenListener {
    fn handle_upgrade_transaction_started() {
        // queue a task to:
        // 1. Set open_request’s result to connection.
        // 2. Set open_request’s transaction to transaction.
       // 3. fire a version change event 
       // and more from Step 9 of http://w3c.github.io/IndexedDB/#run-an-upgrade-transaction
    }
    fn handle_database_opened() {
        // queue a task to Set request’s done flag to true, and so on...
       // and more from http://w3c.github.io/IndexedDB/#ref-for-open-a-database%E2%91%A0
    }
}
```

And then from the "in-parallel" steps running in the `net` component, you can send the various messages over IPC back to script using the sender sent in the initial message from script, and the `OpenListener` on the `ROUTER` thread would handle messages by calling it's appropriate methods, which would always queue a task as per the spec to manipulate that various DOM objects and fire events.

components/script/dom/idbtransaction.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    // https://www.w3.org/TR/IndexedDB-2/#dom-idbtransaction-mode
+    fn Mode(&self) -> IDBTransactionMode {
+        self.mode
+    }
+
+    // https://www.w3.org/TR/IndexedDB-2/#dom-idbtransaction-error
+    fn Error(&self) -> DomRoot<DOMException> {
+        // FIXME:(rasviitanen) ???
+        // It's weird that the WebIDL specifies that this isn't returning an Option.
```
jdm: Filed https://github.com/w3c/IndexedDB/issues/312. Gecko uses ?, so we should as well.

components/script/dom/webidls/IDBFactory.webidl
```diff
@@ -0,0 +1,31 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+/*
+ * The origin of this IDL file is
+ * https://w3c.github.io/IndexedDB/#idbfactory
+ *
+ */
+
+// https://w3c.github.io/IndexedDB/#idbfactory
+partial interface mixin WindowOrWorkerGlobalScope {
+  [SameObject] readonly attribute IDBFactory indexedDB;
```
jdm: This is missing a Pref annotation, which is why the tests are changing behaviour.

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl IDBOpenDBRequest {
+    pub fn new_inherited() -> IDBOpenDBRequest {
+        IDBOpenDBRequest {
+            idbrequest: IDBRequest::new_inherited(),
+        }
+    }
+
+    pub fn new(global: &Window) -> DomRoot<IDBOpenDBRequest> {
```
jdm: This should take a &GlobalScope argument, since these can be created from worker global scopes as well.

jdm:
You can run `wget https://community.taskcluster-artifacts.net/VZJOQ2QZT02QrGOjmc4Qvw/0/public/test-wpt.log` followed by `./mach update-wpt test-wpt.log` to automatically update the metadata for the indexeddb tests.

rasviitanen:
Sorry, I forgot to push a line that caused some flaky behavior. Hopefully, it is working as it should now :)

rasviitanen:
Now it looks much more like I expected, test metadata is now updated.

components/script/dom/idbtransaction.rs
```diff
@@ -0,0 +1,322 @@
...
+            store_handles: Default::default(),
+            requests: Default::default(),
+            active: Cell::new(true),
+            finished: Cell::new(false),
+            serial_number: serial_number,
+        }
+    }
+
+    pub fn new(
+        global: &Window,
```
jdm: This should be GlobalScope.

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+            index_names: DOMStringList::new(global, Vec::new()),
+            transaction: Default::default(),
+            auto_increment: false,
+
+            db_name: db_name,
+        }
+    }
+
+    pub fn new(
+        global: &Window,
```
jdm: This should be GlobalScope.

components/script/dom/idbrequest.rs
```diff
@@ -0,0 +1,281 @@
...
+
+            result: Heap::default(),
+            error: Default::default(),
+            source: Default::default(),
+            transaction: Default::default(),
+            ready_state: Cell::new(IDBRequestReadyState::Pending),
+        }
+    }
+
+    pub fn new(global: &Window) -> DomRoot<IDBRequest> {
```
jdm: GlobalScope.

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    fn handle_delete_db(&self, result: Result<(), ()>) {
+        let open_request = self.open_request.root();
+        let global = open_request.global();
+        open_request.idbrequest.set_ready_state_done();
+
+        match result {
+            Ok(_) => {
+                let window = global.as_window();
```
jdm: No need for this; GlobalScope::get_cx exists.

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,386 @@
...
+        transaction.upgrade_db_version(version);
+
+        // Step 8
+        let this = Trusted::new(self);
+        let connection = Trusted::new(connection);
+        let trusted_transaction = Trusted::new(&*transaction);
+        global
+            .as_window()
+            .task_manager()
+            .dom_manipulation_task_source()
```
jdm: According to https://www.w3.org/TR/IndexedDB-2/#database-access-task-source we should be using a new task source, and it exist for both windows and workers (so there should be a getter on GlobalScope to avoid calling as_window).

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,386 @@
...
+            .queue(
+                task!(send_upgradeneeded_notification: move || {
+                    let this = this.root();
+                    let txn = trusted_transaction.root();
+                    let conn = connection.root();
+                    let global = this.global();
+                    let window = global.as_window();
+                    let cx = window.get_cx();
```
jdm: global.get_cx() avoids calling as_window.

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,386 @@
...
+        let open_operation = SyncOperation::OpenDatabase(
+            sender,
+            global.origin().immutable().clone(),
+            name.to_string(),
+            version,
+        );
+
+        let (task_source, canceller) = window
+            .task_manager()
+            .dom_manipulation_task_source_with_canceller();
```
jdm: Should be using the new database access task source instead.

components/script/dom/idbrequest.rs
```diff
@@ -0,0 +1,281 @@
...
+struct RequestListener {
+    request: Trusted<IDBRequest>,
+}
+
+impl RequestListener {
+    fn handle_async_request_finished(&self, result: Option<Vec<u8>>) {
+        let request = self.request.root();
+        let global = request.global();
+        let window = global.as_window();
+        let cx = window.get_cx();
```
jdm: global.get_cx() instead.

jdm (commented):
Haven't reviewed any real logic yet, just noticed a bunch of worker/window global issues that we'll want to avoid.

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,378 @@
...
+        );
+
+        let task_source = global.database_access_task_source();
+
+        let trusted_request = Trusted::new(self);
+        let name = name.to_string();
+        ROUTER.add_route(
+            receiver.to_opaque(),
+            Box::new(move |message| {
+                let global = trusted_request.root().global();
```
jdm: It is not possible to root DOM objects off of the script thread. Sinec this is only used for passing as an arugment to TaskSource::queue, you will need to use TaskSource::queue_with_canceller instead, and call GlobalScope::task_canceller when calling GlobalScope::database_access_task_source.

This the cause of a huge number of panics in the most recent try run :)

components/script/dom/idbopendbrequest.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        let delete_operation =
+            SyncOperation::DeleteDatabase(sender, global.origin().immutable().clone(), name);
+
+        let trusted_global = Trusted::new(&*global);
+        ROUTER.add_route(
+            receiver.to_opaque(),
+            Box::new(move |message| {
+                let response_listener = response_listener.clone();
+                let global = trusted_global.root();
```
jdm: Same issue here - we need to have a task canceller already instead of rooting this global.

components/script/dom/idbrequest.rs
```diff
@@ -0,0 +1,277 @@
...
+        ROUTER.add_route(
+            receiver.to_opaque(),
+            Box::new(move |message| {
+                let response_listener = response_listener.clone();
+                let _ = task_source.queue(
+                    task!(request_callback: move || {
+                        response_listener.handle_async_request_finished(
+                            message.to().expect("Could not unwrap message"));
+                    }),
+                    trusted_global.root().upcast(),
```
jdm: Same issue here.

rasviitanen: yeah, I should probably get some sleep. I'm missing some really obvious things right now. I even thought I compiled and tested the program with `mach check`... Anyway, I've fixed this now.

jdm:
```
16 unexpected results that are NOT known-intermittents:
  â–¶ OK [expected CRASH] /IndexedDB/blob-delete-objectstore-db.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/blob-delete-objectstore-db.any.worker.html:
  â”‚ FAIL [expected PASS] Deleting an object store and a database containing blobs doesn't crash.
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/blob-valid-before-commit.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/blob-valid-before-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Blobs can be read back before their records are committed.
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/idb-explicit-commit-throw.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit-throw.any.worker.html:
  â”‚ FAIL [expected PASS] Any errors in callbacks that run after an explicit commit will not stop the commit from being processed.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/blob-valid-after-deletion.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/blob-valid-after-deletion.any.worker.html:
  â”‚ FAIL [expected PASS] Blobs stay alive after their records are deleted.
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/blob-contenttype.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/blob-contenttype.any.worker.html:
  â”‚ FAIL [expected PASS] Ensure that content type round trips when reading blob data
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/idbindex_keyPath.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idbindex_keyPath.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex's keyPath attribute returns the same object.
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/structured-clone-transaction-state.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/structured-clone-transaction-state.any.worker.html:
  â”‚ FAIL [expected PASS] Transaction inactive during structured clone in IDBObjectStore.add()
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/structured-clone-transaction-state.any.worker.html:
  â”‚ FAIL [expected PASS] Transaction inactive during structured clone in IDBObjectStore.put()
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/structured-clone-transaction-state.any.worker.html:
  â”‚ FAIL [expected PASS] Transaction inactive during structured clone in IDBCursor.update()
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/idbindex_reverse_cursor.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idbindex_reverse_cursor.any.worker.html:
  â”‚ FAIL [expected PASS] Reverse cursor sees update from separate transactions.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idbindex_reverse_cursor.any.worker.html:
  â”‚ FAIL [expected PASS] Reverse cursor sees in-transaction update.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/key-generators/reading-autoincrement-indexes-cursors.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.openCursor() iterates over an index on the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.openKeyCursor() iterates over an index on the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.openCursor() iterates over an index not covering the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.openKeyCursor() iterates over an index not covering the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/key-generators/reading-autoincrement-store.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-store.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore.getAll() for an autoincrement store
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-store.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore.getAllKeys() for an autoincrement store
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-store.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore.get() for an autoincrement store
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.getAll() for an index on the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.getAllKeys() for an index on the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.get() for an index on the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.getAll() for an index not covering the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.getAllKeys() returns correct result for an index not covering the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-indexes.any.worker.html:
  â”‚ FAIL [expected PASS] IDBIndex.get() for an index not covering the autoincrement key
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/idbcursor-request.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idbcursor-request.any.worker.html:
  â”‚ FAIL [expected PASS] cursor.request from IDBObjectStore.openCursor
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code

  â–¶ Unexpected subtest result in /IndexedDB/idbcursor-request.any.worker.html:
  â”‚ FAIL [expected PASS] cursor.request from IDBObjectStore.openKeyCursor
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code

  â–¶ Unexpected subtest result in /IndexedDB/idbcursor-request.any.worker.html:
  â”‚ FAIL [expected PASS] cursor.request from IDBIndex.openCursor
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code

  â–¶ Unexpected subtest result in /IndexedDB/idbcursor-request.any.worker.html:
  â”‚ FAIL [expected PASS] cursor.request from IDBIndex.openKeyCursor
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/idb-explicit-commit.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Explicitly committed data can be read back out.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] commit() on a version change transaction does not cause errors.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] A committed transaction becomes inactive immediately.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] A committed transaction is inactive in future request callbacks.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Puts issued after commit are not fulfilled.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Calling commit on an aborted transaction throws.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Calling commit on a committed transaction throws.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Calling abort on a committed transaction throws and does not prevent persisting the data.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Calling txn.commit() when txn is inactive should throw.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Transactions with same scope should stay in program order, even if one calls commit.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Transactions that explicitly commit and have errors should abort.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/idb-explicit-commit.any.worker.html:
  â”‚ FAIL [expected PASS] Transactions that handle all errors properly should behave as expected when an explicit commit is called in an onerror handler.
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/idbobjectstore_keyPath.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/idbobjectstore_keyPath.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore's keyPath attribute returns the same object.
  â””   â†’ assert_unreached: deleteDatabase should succeed Reached unreachable code


  â–¶ OK [expected CRASH] /IndexedDB/key-generators/reading-autoincrement-store-cursors.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-store-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore.openCursor() iterates over an autoincrement store
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/key-generators/reading-autoincrement-store-cursors.any.worker.html:
  â”‚ FAIL [expected PASS] IDBObjectStore.openKeyCursor() iterates over an autoincrement store
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"


  â–¶ OK [expected CRASH] /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Committed data can be read back out: case 0
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Committed data can be read back out: case 1
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Committed data can be read back out: case 2
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Committed data can be read back out: case 3
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Committed data can be read back out: case 4
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"

  â–¶ Unexpected subtest result in /IndexedDB/transaction-relaxed-durability.tentative.any.worker.html:
  â”‚ FAIL [expected PASS] Invalid durability option throws a TypeError
  â””   â†’ assert_equals: Expected success event, but got error event instead expected "success" but got "error"
```

Finally, a much more reasonable set of results!

components/script/dom/idbdatabase.rs
```diff
@@ -0,0 +1,293 @@
...
+        // Step 9
+        let object_store = IDBObjectStore::new(
+            &self.global(),
+            self.name.clone(),
+            name.clone(),
+            Some(options),
+        );
+        object_store.set_transaction(&upgrade_transaction);
+
+        // FIXME:(rasviitanen!!!) Move this to constructor
```
jdm: File a followup issue for this?

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+#[dom_struct]
+pub struct IDBObjectStore {
+    reflector_: Reflector,
+    name: DomRefCell<DOMString>,
+    key_path: Option<KeyPath>,
+    index_names: DomRoot<DOMStringList>,
+    transaction: MutNullableDom<IDBTransaction>,
+    auto_increment: bool,
+
+    // No specification below this line
```
jdm: This is a surprising comment.

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    // https://www.w3.org/TR/IndexedDB-2/#valid-key-path
+    pub fn is_valid_key_path(key_path: &StrOrStringSequence) -> bool {
+        let is_valid = |path: &DOMString| {
+            let path = path.to_string();
+            let mut identifiers = path.split('.').into_iter();
+
+            while let Some(_identifier) = identifiers.next() {
+                // FIXME:(rasviitanen) Implement this properly
```
jdm: File a followup issue?

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+        if input.is_string() {
+            let key = structuredclone::write(cx, input, None).expect("Could not serialize key");
+            return Ok(IndexedDBKeyType::String(key.serialized.clone()));
+        }
+
+        if input.is_object() {
+            rooted!(in(*cx) let object = input.to_object());
+            unsafe {
+                let built_in_class: *mut ESClass =
+                    libc::malloc(mem::size_of::<ESClass>()) as *mut ESClass;
```
jdm: Rather than using malloc, let's do `let mut built_in_class = ESClass::Other;` and pass `&mut built_in_class` to GetBuiltInClass.

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+                    panic!("failed to allocate memory");
+                }
+
+                if !GetBuiltinClass(*cx, object.handle().into(), built_in_class) {
+                    libc::free(built_in_class as *mut libc::c_void);
+                    return Err(Error::Data);
+                }
+
+                if let ESClass::Date = *built_in_class {
+                    // FIXME:(rasviitanen)
```
jdm: File a followup issue?

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+                    unimplemented!("Dates as keys is currently unsupported");
+                }
+
+                if IsArrayBufferObject(*object) || JS_IsArrayBufferViewObject(*object) {
+                    let _is_view_object = JS_IsArrayBufferViewObject(*object);
```
jdm: This seems odd.

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+                            rooted!(in(*cx) let input_val = current_val.to_string());
+                            unsafe {
+                                let string_len = JS_GetStringLength(*input_val) as f32;
+                                string_len.to_jsval(*cx, return_val);
+                            }
+                            break;
+                        }
+
+                        if !current_val.is_object() {
+                            // FIXME:(rasviitanen) Return a proper error
```
jdm: File a followup please.

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+                } // All tokens processed
+
+                if !target_object.get().is_null() {
+                    // If this fails, we lose, and the web page sees a magical property
+                    // appear on the object :-(
+                    unsafe {
+                        let prop_name_as_utf16: Vec<u16> =
+                            target_object_prop_name.unwrap().encode_utf16().collect();
+                        let succeeded: *mut ObjectOpResult =
+                            libc::malloc(mem::size_of::<ObjectOpResult>()) as *mut ObjectOpResult;
```
jdm: Instead of malloc, use `let mut succeeded = ObjectOptResult { code_: ObjectOptResult_SpecialCodes::Uninitialized as usize };`

components/script/dom/idbobjectstore.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    // https://www.w3.org/TR/IndexedDB-2/#extract-a-key-from-a-value-using-a-key-path
+    fn extract_key(
+        cx: SafeJSContext,
+        input: HandleValue,
+        key_path: &KeyPath,
+        multi_entry: Option<bool>,
+    ) -> Result<IndexedDBKeyType, Error> {
+        // Step 1: Evaluate key path
+        // FIXME:(rasviitanen) Do this propertly
```
jdm: File a followup please.

components/script/dom/idbtransaction.rs
```diff
@@ -0,0 +1,316 @@
...
+    // Not specified in WebIDL below this line
+    store_handles: DomRefCell<HashMap<String, Dom<IDBObjectStore>>>,
+    // https://www.w3.org/TR/IndexedDB-2/#transaction-request-list
+    requests: DomRefCell<Vec<Dom<IDBRequest>>>,
+    // https://www.w3.org/TR/IndexedDB-2/#transaction-active-flag
+    active: Cell<bool>,
+    // https://www.w3.org/TR/IndexedDB-2/#transaction-finish
+    finished: Cell<bool>,
+    // An unique identifier, used to commit and revert this transaction
+    // FIXME:(rasviitanen) Replace this with a channel
```
jdm: File a followup?

jdm (changes requested):
Looking good!

components/net/indexeddb/engines/rkv.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    // Starts a transaction, processes all operations for that transaction,
+    // and commits the changes.
+    fn process_transaction(
+        &self,
+        transaction: KvsTransaction,
+    ) -> Box<dyn Future<Item = Option<Vec<u8>>, Error = RecvError> + Send> {
+        // FIXME:(rasviitanen) This piece of code is a mess,
+        // it should probably be refactored, and `expects` needs
+        // another look to assert that they are safe.
```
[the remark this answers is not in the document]
rasviitanen: I think I forgot to remove the fixme when I refactored this code, it should be OK now. Likely a side effect of middle-of-the-night changes.

components/script/dom/idbobjectstore.rs
jdm: Yeah, removing unused code is best.

jdm:
Darn - lmdb-rkv-sys contains native Windows code that isn't compatible with UWP builds. The shortest path to success here is to have a dummy IDB backend that's used when `cfg(target_vendor = "uwp")` is true that means that we don't compile the rkv crates.

jdm:
Closing in favour of #33044.

PR #25218, opened 2019-12-09
Rename the ServoSrc type to ServoGstSrc to avoid a name clash

### LINKED ISSUE #25051: Two gstreamer plugins, both called servosrc
Servo media has a plugin called `ServoSrc` which conflicts with the one in `ports/gstplugin`. https://github.com/servo/media/blob/1e527446cf57f5a8773cda1257e46957a295dd9c/backends/gstreamer/source.rs#L247-L249

asajeffrey:
There's a name clash with the type name `ServoSrc`.

---

- [x] These changes fix #25051 

### DIFF
--- a/ports/gstplugin/lib.rs
+++ b/ports/gstplugin/lib.rs
@@ -4,7 +4,7 @@
 
 use glib::subclass::types::ObjectSubclass;
 use gstreamer::gst_plugin_define;
-use servosrc::ServoSrc;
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Woohoo media

asajeffrey:
A side-effect of this PR is that the lamp demo now works:

Screen Shot 2019-12-09 at 12 07 12 PM

PR #25219, opened 2019-12-09
Update ANGLE for UWP fix.

jdm:
This fixes a WACK error where the ANGLE DLLs do not contain the App Container feature.


jdm:
@bors-servo retry
#24726

jdm:
@bors-servo retry
* #24762

PR #25223, opened 2019-12-10
Changed error type on qname, per spec

### LINKED ISSUE #25110: Passing invalid qualified name arguments to createDocument, createElementNS, and setAttributeNS throw NamespaceError instead of INVALID_CHARACTER_ERR
These throws of the wrong exception are in WPT dom/nodes/DOMImplementation-createDocument.html, Document-createElementNS.html, and attributes.html respectively; other browsers are in agreement with WPT about the exception type.

pshaughn:
As specified in https://dom.spec.whatwg.org/#validate invalid QNames when creating a namespaced document/element/attribute get an invalid character exception, not a namespace exception.

---

- [X] These changes fix #25110 


components/script/dom/bindings/xmlname.rs
```diff
@@ -10,15 +10,10 @@ use html5ever::{LocalName, Namespace, Prefix};
...
-        XMLName::InvalidXMLName => {
-            // Step 1.
-            Err(Error::InvalidCharacter)
-        },
-        XMLName::Name => {
-            // Step 2.
-            Err(Error::Namespace)
-        },
+        XMLName::InvalidXMLName => Err(Error::InvalidCharacter),
+        XMLName::Name => Err(Error::InvalidCharacter),
```
jdm: My reading of https://dom.spec.whatwg.org/#validate says that XMLName shouldn't throw anything. How do the test results look if that's the case?

Manishearth: Oh, I misread this as being "does not match Name", but that's not the case

pshaughn: Good point, maybe there's more than just the exception-type to fix here. I'll find out.

pshaughn: When I change it to let Name in instead of throwing InvalidCharacter for it, then some tests go from correctly throwing InvalidCharacter to incorrectly not throwing, like Document.implementation.createDocumentType("prefix::local", "", "") is expected to throw Invalid Character. "prefix::local" appears to match the Name production. The test seems more consistent with just "if qualifiedName does not match the QName production."

pshaughn: However I should check the codepaths for those "" parts too, maybe those are what's supposed to cause the error this time.

pshaughn: https://dom.spec.whatwg.org/#interface-documenttype nope those are totally allowed to be empty strings, the error has to be for the prefix::local part

pshaughn:
Could I get someone's opinion on what the specs say to do vs. what the tests say to do vs. what the code does? I'm not sure which part's out of sync.

Manishearth:
Sorry about the delay; been doing various travel things. I would file an issue on the spec and tag whoever has touched that area recently. Typically I've found that the tests for this kind of thing are more "correct" (in that they match what other browsers do). You should also check what other browsers do, and if it doesn't match the spec then get the spec changed.

pshaughn:
It turns out there's already an issue: https://github.com/whatwg/dom/issues/671

pshaughn:
(this push is just a rebase, not a functional update)

pshaughn:
Based on that issue thread I think it's safe to assume the spec change reflects reality and will eventually land; I'm going to rebase again and add a comment about the issue.

pshaughn:
This is old enough some test might have slid out from under it...
@bors-servo try=wpt

PR #25227, opened 2019-12-10
Remove unnecessary thread in filereader, add stream TODO

gterzian:
I think there are no benefits to queuing the various `FileReadingTask` from a thread, and reading the actual blob bytes are still done on the main-thread. We could try to move the reading off the main-thread(but note the `Blob` is not something you would want to move to another thread "as-is" since it's a `DomObject`), and I think we should just aim for implementing streams and follow the spec instead. 

---


gterzian:
@jdm r?

gterzian:
Hang-on I'll try to remove the `Arc` around the blob contents as well..

jdm:
Agreed!

@bors-servo r+

jdm:
@bors-servo retry
* network issue

jdm:
@bors-servo retry
* network issue

Issue #25175, opened 2019-12-06
Content-Type preflight safelisting gets some cases wrong.


pshaughn:
WPT cors/cors-safelisted-request-header.any.js (seen as .any.html in test results) tests this.

Case "text/plain;garbage" appears to trigger preflighting when it shouldn't, unless it's causing some other unrelated NetworkError. This should be safelisted because it consists only of CORS-safe ASCII and it is an essence match for text/plain.

All the cases with CORS-unsafe characters appear to bypass preflighting when they should trigger it, even "text/plain;garbage\u0001\u0002".

pshaughn:
Some of this is happening inside Hyper's MIME crate but I can at least add the concept of a cors-unsafe character.
@highfive assign me

pshaughn:
This is going up higher than I expected; we never even get to the safelist-checker, because the net_traits requests in this test always end up in RequestMode::CorsMode by the time we construct the DOM request. I'm looking for where that happens now.

pshaughn:
I am mistaken: we get to _a_ safelist checker. #25235 had me looking at the wrong one.

pshaughn:
I have implemented some checks for the CORS-unsafe characters, but the headers have been stored in a HeaderMap, which quietly strips out the unsafe characters on its own, so when the headers are iterated later they are not there to trigger preflight. I'll check those changes in as part of #25235 but this issue is not being solved by them.

pshaughn:
HeaderMap also seems to have a different interpretation of what it means to insert a header than WHATWG does. When I assumed it had the same interpretation and adjusted the Servo code around it, I ended up with crashes, so I've left the part that calls HeaderMap's insert unchanged and a few tests I'd have liked to fix are failing for that reason.

The way headers are normalized in the course of constructing and operating on a DOM request seems to be different in many places from how they're used in an actual http transaction, or at least from how hyper uses them. I'm not sure it makes architectural sense to use the same HeaderMap class in both places.

PR #25236, opened 2019-12-10
De-deplicate is_cors_safelisted_request_header helper functions

### LINKED ISSUE #25235: is_cors_safelisted_request_header is implemented twice, with differences
script/dom/headers.rs has one that takes name:&str and value:&[u8], which is called from the DOM Request constructor only when a request init's mode is explicitly set to no-cors. 

net/fetch/methods.rs has one that takes name:&HeaderName and value:&HeaderValue, which is called routinely as part of the "main fetch" algorithm.

Due to the different data types, unifying them isn't as simple as I'd like it to be, but it's obviously the right thing to do. Neither is currently implementing the full specification logic, possibly due to specification changes since they were written.

pshaughn:
Separate is_cors_safelisted_request_header implementations in script::dom::request and net::fetch::methods have been merged to a single implementation in net_traits::request, with additional logic for spec requirements that weren't previously there. This doesn't pass all the failing tests, but it doesn't fail any passing ones either and it reduces confusion about what's supposed to happen where.

---

- [X] These changes fix #25235 and some but not all subcases in #25175

- [X] There are tests for these changes, in that the WPT CORS tests that did already pass still do

### DIFF
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -12,15 +12,17 @@ use content_security_policy as csp;
 use crossbeam_channel::{unbounded, Receiver, Sender};
 use devtools_traits::DevtoolsControlMsg;
 use headers::{AccessControlExposeHeaders, ContentType, HeaderMapExt, Range};
-use http::header::{self, HeaderMap, HeaderName, HeaderValue};
+use http::header::{self, HeaderMap, HeaderName};
 use hyper::Method;
 use hyper::StatusCode;
 use ipc_channel::ipc::IpcReceiver;
 use mime::{self, Mime};
 use net_traits::blob_url_store::{parse_blob_url, BlobURLStoreError};
 use net_traits::filemanager_thread::RelativePos;
+use net_traits::request::{
+    is_cors_safelisted_method, is_cors_safelisted_request_header, Origin, ResponseTainting, Window,
+};
 use net_traits::request::{CredentialsMode, Destination, Referrer, Request, RequestMode};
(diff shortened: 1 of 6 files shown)


pshaughn:
It turns out this did pass a few failing tests I hadn't tried, and after seeing that I looked back at the spec, saw a rule I missed, and made it pass more of them. I think everything else that went wrong in that test run was intermittents of some sort, but I stand to be proven wrong.

jdm:
Yes, there were a bunch of known intermittent failures in the non-cors test results.

CYBAI:
```
Diff in /repo/components/net_traits/request.rs at line 554:
 ) -> bool {
     let name: &str = name.as_ref();
     let value: &[u8] = value.as_ref();
-    if value.len()>128 { return false; }
(B+    if value.len() > 128 {
(B+        return false;
(B+    }
(B     match name {
         "accept" => is_cors_safelisted_request_accept(value),
         "accept-language" | "content-language" => is_cors_safelisted_language(value),
clang-format not installed. Skipping CPP formatting.
```

pshaughn:
Ready to retry; there are still some failing cases but I think I've gotten as far as I can on this without needing changes to how we use the hyper HeaderMap itself.

jdm:
@bors-servo retry
* #23290

jdm:
@bors-servo retry
* #24762

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

PR #25240, opened 2019-12-11
Took out the special mouseover cancel case

### LINKED ISSUE #25190: Remove obsolete special-case handling of Boolean return value from mouseover event
WPT html/webappapis/scripting/events/event-handler-processing-algorithm.html handily points us at https://github.com/whatwg/html/pull/2398 which removed a Gecko-specific quirk from the spec and came with a matching WPT change. Mouseover events should now just do the same thing on true/false as other mouse events.

pshaughn:
eventtarget.rs was checking the event type to see if it was a mouseover; now it doesn't. Other special-case events like error and beforeunload go down other code paths entirely, so this one has no special cases left.

---

- [X] These changes fix #25190

### DIFF
--- a/components/script/dom/eventtarget.rs
+++ b/components/script/dom/eventtarget.rs
@@ -227,14 +227,13 @@ impl CompiledEventListener {
                             rooted!(in(*cx) let value = value);
                             let value = value.handle();
 
-                            //Step 4
-                            let should_cancel = match event.type_() {
-                                atom!("mouseover") => {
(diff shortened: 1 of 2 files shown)


SimonSapin:
Looks good, thanks!

@bors-servo r+

I’ve added a commit to your branch to add `FIXME` to the new comment.

jdm:
@bors-servo retry
#24762

Issue #24786, opened 2019-11-19
Intermittent macos build error in mozjs (clang++: cannot execute binary file) (/Library/Developer/CommandLineTools/usr/bin/make: cannot execute binary file)


jdm:
```
[...]

PR #25246, opened 2019-12-11
Block the gstreamer plugin waiting for the next frame

### LINKED ISSUE #24833: GStreamer plugin framerates
It looks like the gstreamer plugin is ignoring the negotiated framerate and running at full speed. It shouldn't.

asajeffrey:
Get the GStreamer plugin to produce frames at the requested rate rather than relying on downstream elements to perform throttling.

---

- [x] These changes fix #24833 

### DIFF
--- a/ports/gstplugin/README.md
+++ b/ports/gstplugin/README.md
@@ -24,7 +24,6 @@ To run locally:
 ```
 GST_PLUGIN_PATH=target/gstplugins \
   gst-launch-1.0 servowebsrc \
-    ! videorate \
     ! video/x-raw\(memory:GLMemory\),framerate=50/1,width=1920,height=1080,format=RGBA \
     ! glimagesink rotate-method=vertical-flip
 ```
@@ -33,7 +32,6 @@ To stream over the network:
 ```
 GST_PLUGIN_PATH=target/gstplugins \
   gst-launch-1.0 servowebsrc \
-    ! videorate \
     ! video/x-raw\(memory:GLMemory\),framerate=50/1,width=512,height=256 \
     ! glcolorconvert \
     ! gldownload \
@@ -47,7 +45,6 @@ To  save to a file:
 ```
 GST_PLUGIN_PATH=target/gstplugins \
   gst-launch-1.0 servowebsrc \
-    ! videorate \
     ! video/x-raw\(memory:GLMemory\),framerate=50/1,width=512,height=256 \
     ! glcolorconvert \
     ! gldownload \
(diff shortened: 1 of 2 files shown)


ports/gstplugin/servosrc.rs
```diff
@@ -577,6 +611,23 @@ impl BaseSrcImpl for ServoGstSrc {
     }
 
     fn create(&self, src: &BaseSrc, _offset: u64, _length: u32) -> Result<Buffer, FlowError> {
+        // We block waiting for the next frame to be needed.
+        // TODO: Once get_times is in BaseSrcImpl, we can use that instead.
+        // It's been merged but not yet published.
+        // https://github.com/servo/servo/issues/25234
+        let elapsed_micros = self.start.elapsed().as_micros() as u64;
+        let frame_duration_micros = self.frame_duration_micros.load(Ordering::SeqCst);
```
Manishearth: If this is the only thread doing reads and writes we can just have Acquire/Release everywhere here, I think

ports/gstplugin/servosrc.rs
```diff
@@ -511,6 +533,18 @@ impl BaseSrcImpl for ServoGstSrc {
...
 
+        // Save the framerate if it is set
+        let framerate = outcaps
+            .get_structure(0)
+            .and_then(|cap| cap.get::<Fraction>("framerate"));
+        if let Some(framerate) = framerate {
+            let frame_duration_micros =
+                1_000_000 * *framerate.denom() as u64 / *framerate.numer() as u64;
+            debug!("Setting frame duration to {}micros", frame_duration_micros);
+            self.frame_duration_micros
```
Manishearth: This never changes: Does it need to be an atomic?

Manishearth (approved):
approach looks fine, i feel like there's probably a builtin way to do this, and i think the atomics can be better

ports/gstplugin/servosrc.rs
Manishearth: actually if it's just one thread then relaxed probably works? unsure.

asajeffrey: All true, I am being conservative here,m I doubt that the different orderings make much difference, given that the next operations are `thread::sleep` followed by locking a `Mutex`.

ports/gstplugin/servosrc.rs
asajeffrey: Renegotiating caps can happen at any time, and the src needs to be `Sync`, so yes it needs to be an `Atomic`.

asajeffrey (commented):
Really this should be using gstreamers `get_times` API, but that's not exposed to Rust yet.

asajeffrey:
Rebased. @bors-servo r=Manishearth

jdm:
@bors-servo retry
* #24786

Issue #13152, opened 2016-09-01
Dispatch error events when event handler attributes fail to compile.


Ms2ger:
I'm not currently doing that because of double borrows in `html/webappapis/scripting/processing-model-2/runtime-error-in-body-onerror.html`.

We should also make sure to pass `tests/wpt/web-platform-tests/html/webappapis/scripting/events/invalid-uncompiled-raw-handler-compiled-late.html`; I believe that we'll need to change the timing of compilation for that.

Ms2ger:
Should also write tests that add/remove event listeners from the error event handler.

Ms2ger:
> when compiling the event handlers, we borrow eventtarget.handlers, and then we call back into script with that borrow held

jdm:
@ofekd Does this look interesting? This is in the code in `components/script/dom/eventtarget.rs`.

ofekd:
@jdm Yeah I can do this one. Though take into account I'm still not fast solving things in Servo

jdm:
That's fine! Please ask questions if anything's unclear.

ofekd:
@Ms2ger I've spent some time learning the events code, but I still miss the double borrows, and how it's all specific to `body onerror`.

jdm:
@ofekd I believe the problem is that if we dispatch an error event when compilation fails, it triggers a double borrow error. Because of this, we do not dispatch an error event right now. I haven't verified this, however.

Ms2ger:
Say we have
```html
<body onerror="f()" onclick="}">
<script>
function f() { ... }
document.body.dispatchEvent(new Event("click"));
</script>
```

When `dispatchEvent` is called, we will borrow the body element's `handlers` field (mutably), try to compile the `onclick` attribute's value, and fail. Then we'll try to dispatch an `error` event, which will try to borrow the same `handlers` field, and this panics.

I don't have a clear idea of how to fix this, but I do think the specification requires us to compile the just before calling the event handler. It is probably possible to detect with this test case:

```html
<body onerror="f()">
<script>
function f() { ... }
var body = document.body;
body.addEventListener("click", function g() { ... });
body.setAttribute("onclick", "}");
body.dispatchEvent(new Event("click"));
</script>
```

In this case, it is my understanding `g()` should be called before `f()`.

ofekd:
Hi all

I'd like to apologize --

I took this when I was relatively available and then literally days later got held up pretty bad with work. This is not what I meant when I said I'm not that fast with Servo.

[...]

PR #25249, opened 2019-12-11
Throwing errors on event compilation, without panics and at least sometimes test-passingly

### LINKED ISSUE #25200: Setting onclick to a garbage string and then evaluating it (without calling it) should fire an error event but doesn't
WPT html/webappapis/scripting/events/invalid-uncompiled-raw-handler-compiled-once.window.html  and invalid-uncompiled-raw-handler-keeps-position.window.html fail on this; they expect window.onerror to have fired and it hasn't. The onclick is showing up as null, so there was an attempt to compile the Javascript that decided it was invalid, but the invalidness didn't manifest in an error.

pshaughn:
Error events on event handler attribute string compilation were disabled a few years ago (#13152) for fear of a double-borrow panic; event propagation code has changed enough since then that enabling the events is possible and makes some WPT cases happy.

---

- [X] These changes fix #25200 and make progress on #13152 

### DIFF
--- a/components/script/dom/eventtarget.rs
+++ b/components/script/dom/eventtarget.rs
@@ -553,8 +553,8 @@ impl EventTarget {
             // Step 3.7
             unsafe {
                 let _ac = JSAutoRealm::new(*cx, self.reflector().get_jsobject().get());
-                // FIXME(#13152): dispatch error event.
-                report_pending_exception(*cx, false);
+                // FIXME(#13152): this might not be timed correctly
+                report_pending_exception(*cx, true);
             }
             return None;
         }
--- a/components/script/dom/htmlbodyelement.rs
+++ b/components/script/dom/htmlbodyelement.rs
@@ -183,48 +183,27 @@ impl VirtualMethods for HTMLBodyElement {
     }
 
     fn attribute_mutated(&self, attr: &Attr, mutation: AttributeMutation) {
-        let do_super_mutate = match (attr.local_name(), mutation) {
+        use crate::dom::macros::BODY_FRAMESET_EVENT_HANDLERS;
+        match (attr.local_name(), mutation) {
             (name, AttributeMutation::Set(_)) if name.starts_with("on") => {
-                let window = window_from_node(self);
                 // https://html.spec.whatwg.org/multipage/
                 // #event-handlers-on-elements,-document-objects,-and-window-objects:event-handlers-3
-                match name {
-                    &local_name!("onfocus") |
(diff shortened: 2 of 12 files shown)


jdm:
```
  ▶ OK [expected TIMEOUT] /html/webappapis/scripting/events/compile-event-handler-lexical-scopes.html

  ▶ ERROR [expected OK] /_mozilla/mozilla/event_handler_syntax_error.html
  │   → missing } after function body
  │ 
  │ @http://web-platform.test:8000/_mozilla/mozilla/event_handler_syntax_error.html:14:5
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1984:25
  └ @http://web-platform.test:8000/_mozilla/mozilla/event_handler_syntax_error.html:12:3
```

pshaughn:
https://github.com/servo/servo/blob/master/components/script/dom/htmlelement.rs#L728 seems to be at fault: any attribute starting with "on" gets stored as an uncompiled event handler instead of a string, even when it isn't an event handler attribute.

pshaughn:
This was sort of working, but I stopped it to await https://github.com/servo/servo/pull/25274 so we wouldn't get spurious errors, and then I think that one ran aground?

jdm:
I suspect we can fix event_handler_syntax_error.html by using assert_throws_dom judiciously to verify that an error event occurs without it reaching the test harness?

pshaughn:
It shouldn't be throwing an exception at all. onx and ony should just behave as ordinary attribute names and not do any event-listener compilation.

pshaughn:
Oh, I misread the line number and error type. Line 14 fails for an ErrorEvent, not an exception; we can get around this by putting `setup({ allow_uncaught_exception: true })` at the top of the script. Since that'll also allow literal uncaught exceptions, we can also wrap everything in a try/catch with an assert_unreached in the catch block; that way we allow error events but still catch unexpected exceptions.

pshaughn:
I adjusted that test, pulled in #25249, adjusted #25249 to concatenate "on" before all the event names when building the lookup map for handler names, and I suspect this might actually work.
@bors-servo try=wpt

pshaughn:
Weird and probably not the code; I need to add another check to that event_handler_syntax_error.html test anyway though

pshaughn:
Nope, this combination of merges isn't working, I'm getting script in an "onx" content attribute run when I fire an "x" event. My own code changes here are tiny compared to @Manishearth's; there's not really anything of mine here especially worth keeping other than the concat("on") bit.

PR #25250, opened 2019-12-11
Update surfman for UWP fix.

jdm:
This removes the use of a forbidden UWP API in the ANGLE backend when appropriate.


jdm:
@bors-servo retry
* #24726  #24762

PR #25256, opened 2019-12-12
Allow mach bootstrap to run on windows

### LINKED ISSUE #25224: Fix mach bootstrap on Windows
According to jdm on the IRC, mach bootstrap doesn't do anything useful, because every build-related command is already triggering [windows_msvc](https://github.com/servo/servo/blob/master/python/servo/bootstrap.py#L275).

Invoking bootstrap however, is done by https://github.com/servo/servo/blob/master/python/servo/bootstrap.py#L396 which in turn is called from https://github.com/servo/servo/blob/master/python/mach_bootstrap.py#L241

The following exception happens when ./mach bootstrap is invoked:
```
$ ./mach bootstrap
  File "./mach", line 103, in <module>
    main(sys.argv)
  File "./mach", line 31, in main
    sys.exit(mach_bootstrap.bootstrap_command_only(topdir))
  File "e:\Marc\Documents\Coding\servo\python\mach_bootstrap.py", line 245, in bootstrap_command_only
    bootstrap(context, force)
  File "e:\Marc\Documents\Coding\servo\python\servo\bootstrap.py", line 413, in bootstrap
    return bootstrapper(context, force=force)
  File "e:\Marc\Documents\Coding\servo\python\servo\bootstrap.py", line 278, in windows_msvc
    deps_dir = os.path.join(context.sharedir, "msvc-dependencies")
AttributeError: 'DummyContext' object has no attribute 'sharedir'
```
I think the solution to this is adding `context.sharedir = os.path.join(context.topdir, ".servo")` to the lastly linked code in mach_bootstrap.py, because other build tasks indeed _do_ download the dependencies to `.servo` relative to the topdir.

Note that this would affect windows and linux alike, but it seems this sharedir is only requested from salt and windows and not linux, which installs the packets globally using apt.

MeFisto94:
Fix mach bootstrap, so that it works under windows [and any potential other platforms using a shared dependency download dir]

---

- [X] These changes fix #25224
- [X] These changes do not require tests because mach bootstrap is non-critical tooling, which shouldn't be changed very often and thus should not break.

Question: Is there some kind of formatting I have to/can run on this file?
Also this PR is more of a suggestion, maybe someone can verify this and/or come up with a more clever version not requiring code duplication (though I guess there is hardly anything we can do about that)

Another thing to discuss is whether this shouldn't be done as suggested in #25224, completely hardcoded, without env vars or config files, but that's a more political decision and I belive that it's okay, because I also run mach bootstrap regularly (at least when I get warned by mach) to upgrade my dependencies (in my gecko workflow, that is). In this use-case it should support honoring the config.

I guess @SimonSapin is the Python Guy here?

### DIFF
--- a/python/mach_bootstrap.py
+++ b/python/mach_bootstrap.py
@@ -248,6 +248,23 @@ def bootstrap_command_only(topdir):
 
     context = DummyContext()
     context.topdir = topdir
+
+    # Taken from command_base.py#CommandBase, which we can't depend on during bootstrapping.
+    config_path = os.path.join(context.topdir, ".servobuild")
+    if os.path.exists(config_path):
+        import toml
+        with open(config_path) as f:
+            config = toml.loads(f.read())
+    else:
+        config = {}
+
+    # Handle missing/default items
+    config.setdefault("tools", {})
+    default_cache_dir = os.environ.get("SERVO_CACHE_DIR",
+                                       os.path.join(context.topdir, ".servo"))
+    config["tools"].setdefault("cache-dir", default_cache_dir)
+    context.sharedir = os.path.join(context.topdir, os.path.expanduser(config["tools"]["cache-dir"]))
+
     force = False
     if len(sys.argv) == 3 and sys.argv[2] == "-f":
         force = True


MeFisto94:
Given the Taskgroup failure, I definitely need some linting/formatting tool, does anyone know what the right tool would be?

SimonSapin:
The linting tool is `./mach test-tidy`. It checks (among other things) that Python code is formatted according to https://www.python.org/dev/peps/pep-0008/. Unfortunately we don’t have a formatting tool for Python. You’ll need to reformat manually until the tool is happy. https://stackoverflow.com/questions/15435811/what-is-pep8s-e128-continuation-line-under-indented-for-visual-indent has more details about this particular error message.

SimonSapin:
As to the PR itself, I have no idea how this bootstrap works or what it even does. I’m hoping someone else on the team knows a bit more…

asajeffrey (approved):
LGTM

asajeffrey:
@jdm, thoughts before we start the dance to try to land this?

jdm:
What is the problem we're solving, exactly? As far as I know, the .servo directory only needs to be present so that we can download the MSVC dependencies that are required for building. However, we already do that as part of `mach build` so can we make `mach bootstrap` not do that, instead?

MeFisto94:
Well it would remove an integral part of bootstrap, that is preparing the developers toolchain, so one might intentionally run bootstrap to prepare everything before compiling.
It might be an unfair comparison, but gecko also relies on this heavily. If you try to build and something is too old, it tells you to run bootstrap again.

In this regard servo's mach might be kind of odd due to the combination of bootstrapping behavior in build.

And the precise problem is actually that the MSVC Download just expects the share dir path to be present, but the bootstrapping context is `DummyContext`, so a simple context which did _not_ provide the dir path. So passing that path to me looks better than removing the functionality.

jdm:
My preference is to keep the mach bootstrap as simple as possible. Duplicating code between files that can't share it does not help that, sadly.

SimonSapin:
I wonder if there’s is confusion between `python/mach_bootstrap.py`, which bootstraps mach itself by creating a Python virtualenv and installing its Python dependencies, and the `./mach bootstrap` sub-command defined in `python/servo/bootstrap_commands.py` that installs some dependencies of building Servo.

SimonSapin:
> one might intentionally run bootstrap to prepare everything before compiling

In my mind that’s the role of `./mach fetch`, which runs rustup to make sure the appropriate Rust toolchain is installed and then runs `cargo fetch` which downloads crate dependencies. Other downloads like MSVC should probably be added there if they’re not already.

>  gecko also relies on this heavily

I get the feeling that in Gecko a lot of work has gone into `./mach bootstrap`, such that it is *the* thing to run that will try hard to figure out the situation and make everything Just Work. This is good for contributors, but building and maintaining a tool like this is not trivial. Gecko has a number of people involved in that work.

Just because we also use mach and also have a sub-command named `bootstrap` doesn’t mean that it fills exactly the same role.

MeFisto94:
That is definitely a source of confusion, yes, I was thinking bootstrap is the thing to do all the work, didn't know about fetch. 

Anyway, mach_bootstrap.py is calling into bootstrap_commands.py, see the stacktrace at the relevant issue:
```
 ./mach bootstrap
  File "./mach", line 103, in <module>
    main(sys.argv)
  File "./mach", line 31, in main
    sys.exit(mach_bootstrap.bootstrap_command_only(topdir))
  File "e:\Marc\Documents\Coding\servo\python\mach_bootstrap.py", line 245, in bootstrap_command_only
    bootstrap(context, force)
  File "e:\Marc\Documents\Coding\servo\python\servo\bootstrap.py", line 413, in bootstrap
    return bootstrapper(context, force=force)
  File "e:\Marc\Documents\Coding\servo\python\servo\bootstrap.py", line 278, in windows_msvc
    deps_dir = os.path.join(context.sharedir, "msvc-dependencies")
AttributeError: 'DummyContext' object has no attribute 'sharedir'
```
Maybe we could also fix this by making ./mach bootstrap not be a "bootstrap_command_only", but handle it like any other command and have every comment auto bootstrap? Likewise, that "bootstrap" functionality should belong into "fetch"?

asajeffrey:
@SimonSapin I think you are in a better place than me to review this PR.

mrobinson:
Closing in favor of #29954.

PR #25264, opened 2019-12-12
Initial implementation of GPUBuffer for WebGPU

imiklos:
Added WebIDL bindings for `GPUBuffer`, `GPUBufferDescriptor`, `GPUBufferUsage`.
Implemented the `createBuffer` and `createBufferMapped` functions of `GPUDevice`.

---

cc @kvark @jdm @zakorgy

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/25264)


components/script/dom/gpubuffer.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// https://gpuweb.github.io/gpuweb/#dom-gpubuffer-destroy
+    fn Destroy(&self) {
+        match *self.state.borrow() {
+            GPUBufferState::Mapped => {
+                self.Unmap();
+            },
+            _ => {},
+        };
+        if let Some(window) = self.global().downcast::<Window>() {
```
jdm: Since this method is called from the destructor, instead of getting the channel from the global we should hold on to the channel so we can use it directly.

components/webgpu/lib.rs
```diff
@@ -152,6 +166,41 @@ impl WGPU {
...
+                },
+                WebGPURequest::CreateBufferMapped(sender, device, id, descriptor) => {
+                    let global = &self.global;
+                    let mut arr_buff: *mut u8 = std::ptr::null_mut();
+
+                    let _output = gfx_select!(id =>
+                        global.device_create_buffer_mapped(device.0, &descriptor, &mut arr_buff, id));
+                    let buffer = WebGPUBuffer(id);
+
+                    if let Err(e) = sender.send((buffer, arr_buff as usize)) {
```
jdm: This sends a pointer over IPC, so this won't work with multiple processes. How do we deal with this? Do we need to copy the memory into a Vec and send that instead?

components/script/dom/gpudevice.rs
```diff
@@ -63,6 +75,74 @@ impl GPUDevice {
     }
 }
 
+impl GPUDevice {
+    fn resolve_create_buffer_mapped(
```
jdm: This should be marked unsafe, since it accepts an arbitrary integer, converts it to a pointer, and dereferences it by using it to create a slice.

components/script/dom/gpudevice.rs
```diff
@@ -63,6 +75,74 @@ impl GPUDevice {
...
+            gpu_buffer,
+            self.device,
+            GPUBufferState::Mapped,
+            descriptor.size,
+            descriptor.usage.bits(),
+            valid,
+        );
+        unsafe { buff.to_jsval(*cx, js_gpu_buffer.handle_mut()) };
+        let mut out = Vec::new();
+        out.push(js_gpu_buffer.get());
```
jdm: This can be `ObjectValue(buff.reflector().get_jsobject().get())`, and then we don't need `js_gpu_buffer`.

jdm (changes requested):

kvark (commented):
> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/servo/servo/25264)* status: all files reviewed, 8 unresolved discussions (waiting on @imiklos)

___
*[components/script/dom/gpubuffer.rs, line 32 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv0RpVEulb4QH4nHEa:-Lvv0RpVEulb4QH4nHEb:bne6637) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpubuffer.rs#L32)):*
> ```Rust
>     label: DomRefCell<Option<DOMString>>,
>     size: Cell<GPUBufferSize>,
>     usage: Cell<u32>,
> ```

usage and size shouldn't be mutable: they are set at creation and persist across the life time
___
*[components/script/dom/gpubufferusage.rs, line 9 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv0gBI3UPagEwfB-xB:-Lvv0gBI3UPagEwfB-xC:bivinc1) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpubufferusage.rs#L9)):*
> ```Rust
> 
> #[dom_struct]
> pub struct GPUBufferUsage {
> ```

does this have to be an interface?
There are no methods expected, it's just a namespace, conceptually.
In Gecko, we mark those interfaces as [concrete: false](https://searchfox.org/mozilla-central/rev/62a130ba0ac80f75175e4b65536290b52391f116/dom/bindings/Bindings.conf#1400) and don't need any classes on C++ side for them.
___
*[components/script/dom/gpudevice.rs, line 190 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv1gUA0ZP7CPZ34usc:-Lvv1gUA0ZP7CPZ34usd:b-udeyx2) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpudevice.rs#L190)):*
> ```Rust
>                     .unwrap(),
>                 None => {
>                     return GPUBuffer::new(
> ```

if the GPU process is dead and the channel is not available, we should probably not pretend we are doing anything constructive here
___
*[components/script/dom/gpudevice.rs, line 246 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv1sFSDMWZlsKH29g7:-Lvv1sFSDMWZlsKH29g8:b8165hk) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpudevice.rs#L246)):*
> ```Rust
>         };
> 
>         let (buffer, array_ptr_address) = receiver.recv().unwrap();
> ```

we can't pass around pointers between processes, as @jdm noted already

components/script/dom/gpubuffer.rs
gterzian: I think "holding on to the channel so we can use it directly" will be generally addressed in https://github.com/servo/servo/pull/25030, it might be a good idea to coordinate with @zakorgy and perhaps rebase this one off of those changes when they are completed(or do it in a follow-up requiring some refactoring).

imiklos: I've saved the channel in the `GPUBuffer` so we can send requests directly.

components/webgpu/lib.rs
imiklos: Yes, you are right. I've copied the memory into a `Vec` as you mentioned.

imiklos (commented):
> *[Reviewable](https://reviewable.io/reviews/servo/servo/25264)* status: all files reviewed, 8 unresolved discussions (waiting on @imiklos, @jdm, @kvark, and @zakorgy)

___
*[components/script/dom/gpubuffer.rs, line 32 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv0RpVEulb4QH4nHEa:-LwDzZWa6txkg6hdjtkd:bxb2fp1) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpubuffer.rs#L32)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

usage and size shouldn't be mutable: they are set at creation and persist across the life time
</blockquote></details>

Thank you! I will remove the Cell.
___
*[components/script/dom/gpubufferusage.rs, line 9 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv0gBI3UPagEwfB-xB:-LwE-CUq-YgGSaBocxz_:b-to1kym) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpubufferusage.rs#L9)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

does this have to be an interface?
There are no methods expected, it's just a namespace, conceptually.
In Gecko, we mark those interfaces as [concrete: false](https://searchfox.org/mozilla-central/rev/62a130ba0ac80f75175e4b65536290b52391f116/dom/bindings/Bindings.conf#1400) and don't need any classes on C++ side for them.
</blockquote></details>

I didn't find similar to `concrete: false` in servo. @jdm Do you have an idea about this?
___
*[components/script/dom/gpudevice.rs, line 190 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv1gUA0ZP7CPZ34usc:-LwE-ucP9j6kr0w1hfiT:b-qyrgle) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpudevice.rs#L190)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

if the GPU process is dead and the channel is not available, we should probably not pretend we are doing anything constructive here
</blockquote></details>

You are right! I've just put an `unimplemented!` macro instead of returning with a `GPUBuffer` here.
___
*[components/script/dom/gpudevice.rs, line 246 at r1](https://reviewable.io/reviews/servo/servo/25264#-Lvv1sFSDMWZlsKH29g7:-LwE0QEdDg0bwIWAgZGs:bqhfrm1) ([raw file](https://github.com/servo/servo/blob/706c2acdc537045d31b7cdc80238496164a63348/components/script/dom/gpudevice.rs#L246)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

we can't pass around pointers between processes, as @jdm noted already
</blockquote></details>

Yes, I've copied the memory into a `Vec` and sent that instead of pointers.

imiklos:
@jdm @kvark @gterzian Thank your for your review! I've addressed what you mentioned.

imiklos:
@kvark I get the following validation layer error on destruction:
(this is an unmapped buffer created with `device.createBuffer`)

> [2019-12-16T13:09:39Z ERROR relevant] Values of this type can't be dropped!
> [2019-12-16T13:09:39Z ERROR gfx_backend_vulkan] 
> VALIDATION [VUID-vkDestroyDevice-device-00378 (0)] : OBJ ERROR : For device 0x7fac702eac70, DeviceMemory object 0x3 has not been destroyed. The Vulkan spec states: All child objects created on device must have been destroyed prior to destroying device (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkDestroyDevice-device-00378)
> object info: (type: DEVICE_MEMORY, hndl: 3)
> 
> [2019-12-16T13:09:39Z ERROR relevant] Values of this type can't be dropped!

I got the same result with explicitly call the `Buffer.destroy()` or without.
The `Global.delete()` called before exit.
Am I missing something?

kvark:
>I get the following validation layer error on destruction:

This comes from `rendy-memory`, which you need to double-check if you are destroying correctly.

jdm (approved):

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

PR #25267, opened 2019-12-12
Issue warning to devtools when module scripts in use.

jdm:
This makes it easier to identify why content isn't working in environments which don't have access to the RUST_LOG output.


jdm:
@bors-servo retry
* #22212

jdm:
@bors-servo retry
* #24622

jdm:
@bors-servo retry
* #24726

PR #25272, opened 2019-12-12
DOMException now inherits its Javascript toString from Error

### LINKED ISSUE #25042: DOMException.prototype has its own toString but should only inherit from Error
WPT WebIDL/ecmascript-binding/es-exceptions/custom-bindings.any brings us this detail, "Inherits its toString() from Error.prototype". It seems a bit academic at first glance, but if someone sets a custom toString on Error as part of some framework they're going to want their customization to apply to DOMException, not be overridden by a more specific one.

pshaughn:
stringifier is out of the DOMException IDL, and Stringifier(&self) is moved into the main body of DOMException where it can continue to be called internally.

---

- [X] These changes fix #25042 


components/script/dom/domexception.rs
```diff
@@ -149,6 +149,10 @@ impl DOMException {
             DOMExceptionBinding::Wrap,
         ))
     }
+
+    pub fn Stringifier(&self) -> DOMString {
```
Manishearth: this should be lowercase (and edited in error.rs as well)

jdm:
@bors-servo retry
* #25275

PR #25273, opened 2019-12-12
 Layout `position: fixed` in the initial containing block 

SimonSapin:


components/layout_2020/positioned.rs
```diff
@@ -21,9 +21,15 @@ pub(crate) struct AbsolutelyPositionedBox {
...
+#[derive(Default)]
+pub(crate) struct PositioningContext<'box_tree> {
+    /// With `position: absolute`
+    pub abspos: Vec<CollectedAbsolutelyPositionedBox<'box_tree>>,
+}
+
 #[derive(Debug)]
-pub(crate) struct AbsolutelyPositionedFragment<'box_> {
-    absolutely_positioned_box: &'box_ AbsolutelyPositionedBox,
+pub(crate) struct CollectedAbsolutelyPositionedBox<'box_tree> {
```
nox: What about `HoistedAbsolutelyPositionedBox`? These exist only when being hoisted up the tree to reach their final absolute containing block, right?

components/layout_2020/flow/construct.rs
```diff
@@ -199,13 +191,21 @@ impl BlockContainer {
...
                 .block_level_boxes
                 .into_par_iter()
-                .mapfold_reduce_into(&mut acc, mapfold, |left, right| {
-                    left.contains_floats |= right.contains_floats;
-                    if content_sizes.requests_inline() {
-                        left.outer_content_sizes_of_children
-                            .max_assign(&right.outer_content_sizes_of_children)
-                    }
-                })
+                .mapfold_reduce_into(
```
nox: I wonder how many more closures this call will involve by 2025.

components/layout_2020/flow/mod.rs
```diff
@@ -433,80 +430,70 @@ fn layout_in_flow_non_replaced_block_level<'a>(
...
-    if this_start_margin_can_collapse_with_children.0 {
-        block_margins_collapsed_with_children
-            .start
-            .adjoin_assign(&flow_layout.collapsible_margins_in_children.start);
-        if flow_layout
-            .collapsible_margins_in_children
-            .collapsed_through
-        {
+
+    positioning_context.for_maybe_position_relative(layout_context, style, |positioning_context| {
```
nox: I understand the logic but I'm not sure I like how big this closure is and how it drifts everything even more to the right. Do you think it could be its own function, living in the module next to this one?

components/layout_2020/positioned.rs
```diff
@@ -447,7 +447,7 @@ impl<'box_tree> CollectedAbsolutelyPositionedBox<'box_tree> {
 
             let inline_start = match inline_axis.anchor {
                 Anchor::Start(start) => start + pb.inline_start + margin.inline_start,
-                Anchor::End(end) => cbbs - end - pb.inline_end - margin.inline_end - size.inline,
+                Anchor::End(end) => cbis - end - pb.inline_end - margin.inline_end - size.inline,
```
nox: EWWWWWWWW

nox (approved):
Just a couple of useless remarks and some nits.

components/layout_2020/flow/mod.rs
```diff
@@ -332,10 +311,9 @@ impl BlockLevelBox {
     }
 }
 
-#[derive(PartialEq)]
-enum BlockLevelKind {
-    SameFormattingContextBlock,
-    EstablishesAnIndependentFormattingContext,
+enum BlockLevelKind<'a> {
```
nox: Would it make more sense to name that `BlockKind`?

components/layout_2020/flow/mod.rs
```diff
@@ -424,20 +399,38 @@ fn layout_in_flow_non_replaced_block_level<'a>(
     );
 
     let this_start_margin_can_collapse_with_children = CollapsibleWithParentStartMargin(
-        block_level_kind == BlockLevelKind::SameFormattingContextBlock &&
+        matches!(block_level_kind, BlockLevelKind::SameFormattingContextBlock(_)) &&
```
nox: Maybe a boolean method for this.

nox (approved):
Nice!

components/layout_2020/flow/mod.rs
nox: Or `BlockContents`, or something with "contents" somewhere. I don't like how it has less variants than `BlockLevelBox`.

components/layout_2020/flow/mod.rs
SimonSapin: This is removed in a latter commit.

components/layout_2020/flow/mod.rs
SimonSapin: Done.

jdm:
@bors-servo retry
* #24726

PR #25274, opened 2019-12-13
Only compile applicable event handler content attributes

Manishearth:
Part of https://github.com/servo/servo/issues/13152

r? @jdm

cc @pshaugn

### DIFF
--- a/components/script/dom/htmlbodyelement.rs
+++ b/components/script/dom/htmlbodyelement.rs
@@ -183,48 +183,27 @@ impl VirtualMethods for HTMLBodyElement {
     }
 
     fn attribute_mutated(&self, attr: &Attr, mutation: AttributeMutation) {
-        let do_super_mutate = match (attr.local_name(), mutation) {
+        use crate::dom::macros::BODY_FRAMESET_EVENT_HANDLERS;
+        match (attr.local_name(), mutation) {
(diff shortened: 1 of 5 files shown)


Manishearth:
Hm, this is causing a lot of timeouts, maybe the overhead is too much there.

pshaughn:
> 
> 
> Hm, this is causing a lot of timeouts, maybe the overhead is too much there.

Timeouts can also be because t.done() was inside an event handler that never came. I'd guess that before the overhead from some hashmap lookups.

Manishearth:
It's not the lookups, the lazy_static means that we spend some time setting up the hashmap the first time someone ever sets an event handler. But yeah, we could also be just not triggering the event handlers.

Manishearth:
CLosing in favor of an approach that uses proc macros.

jdm:
@manishearth Where did the approach that uses proc macros end up?

Manishearth:
@jdm Oh, it didn't happen, I played around with it a bit but you basically have to manually generate a PHF map. I was indicating that I wanted to close this in favor of us looking into such an approach, not necessarily that such an approach was already underway. I'll file an issue.

Manishearth:
https://github.com/servo/servo/issues/25760

PR #25300, opened 2019-12-16
Mach: Improve Visual Studio detection for non-standard-path installations

### LINKED ISSUE #25225: Make vcvarsall.bat detection more robust
On my system, there are two installs of Visual Studio:
- Visual Studio 2015
- Visual Studio 2017 Community

Now it seems that I don't even have a proper Visual Studio 2015 installed, which may be some kind of source for the problem, but using 2015 probably won't work either.
Windows says I only have `VS 2015 Shell (Isolated)` and `VS 2015 Tools for Unity` installed.

The problem is, that Visual Studio 2015 contains `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat`. 
This is what mach is picking up (`VS140COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\` is set) and thus it fails to locate MSBuild (`"Can't find MSBuild.exe installation under C:\Program Files (x86)\Microsoft Visual Studio."`)

The real path should be `E:\Programme\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat`

Maybe mach could just search all paths, launch the vcvarsall.bat, try to invoke msbuild.exe and use it to query the version of VS and then use the most recent one. I have to look into the registry as well, if there is something pointing to the VS 2017 Install Directory.

Workaround Number 2 from https://github.com/servo/servo/blob/master/README.md#troubleshooting-a-windows-environment makes MSBuild.exe available, but isn't a solution for me either, because I couldn't add python to the path, because Windows seems to truncate env vars to 1024 characters [and %PATH% got much longer than that when using said prompt]

Apart from that I'd want to go the unsupported way of making mach work under mozilla-build, i.e. the regular firefox environment, because I'm all set there (python installed, git installed, several tools and configurations).
It seems it suffers the same problem as mach.bat: just being unable to locate the vcvarsall.bat file.

I understand there might be no official support for my undertaking, but maybe someone else is willing to help out or my findings can help them with their environment somehow, because given my current situation, I am unable to build servo either way.

Edit1: Do note that multiple folders within the Visual Studio Folder are on the Path by default, so it might be an idea to scan %PATH% for vcvarsall.bat

MeFisto94:
Improve locating the Visual Studio installation

---
- [X] These changes fix #25225
- [X] These changes do not require tests because testing build infrastructure is difficult

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -36,6 +36,7 @@ from .bootstrap import check_gstreamer_lib
 from mach.decorators import CommandArgument
 from mach.registrar import Registrar
 import toml
+import json
 
 from servo.packages import WINDOWS_MSVC as msvc_deps
 from servo.util import host_triple
@@ -1040,6 +1041,24 @@ install them, let us know by filing a bug!")
                 print("Clobber not needed.")
 
 
+def find_highest_msvc_version_ext():
+    def vswhere(args):
+        program_files = (os.environ.get('PROGRAMFILES(X86)') or
+                         os.environ.get('PROGRAMFILES'))
+        if not program_files:
+            return []
+        vswhere = os.path.join(program_files, 'Microsoft Visual Studio',
+                               'Installer', 'vswhere.exe')
+        if not os.path.exists(vswhere):
+            return []
+        return json.loads(check_output([vswhere, '-format', 'json'] + args).decode(errors='ignore'))
+
+    for install in vswhere(['-products', '*', '-requires', 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64',
+                            '-requires', 'Microsoft.VisualStudio.Component.Windows10SDK']):
+        version = install['installationVersion'].split('.')[0] + '.0'
+        yield (install['installationPath'], version, "Current" if version == '16.0' else version)
+
+
 def find_highest_msvc_version():
     editions = ["Enterprise", "Professional", "Community", "BuildTools"]
     prog_files = os.environ.get("ProgramFiles(x86)")
@@ -1059,8 +1078,13 @@ def find_highest_msvc_version():
             vsinstalldir = os.path.join(base_vs_path, version, edition)
             if os.path.exists(vsinstalldir):
                 return (vsinstalldir, vs_version, msbuild_version)
-    print("Can't find MSBuild.exe installation under %s." % base_vs_path)
-    sys.exit(1)
+
+    versions = sorted(find_highest_msvc_version_ext(), key=lambda tup: float(tup[1]))
+    if not versions:
+        print("Can't find MSBuild.exe installation under %s. Please set the VSINSTALLDIR and VisualStudioVersion" +
+              " environment variables" % base_vs_path)
+        sys.exit(1)
+    return versions[0]
 
 
 def get_msbuild_version(vs_version):


MeFisto94:
Explanation:
As you can see, the old code only relied on trying a fixed path to locate visual studio, but this fails for me.
This code essentially comes from https://searchfox.org/mozilla-central/rev/923eec8d2fb8078ebc7a33a9e1ce73eac01f7446/build/moz.configure/toolchain.configure#581 but has been simplified and cut down, so that it fits into servo.

I understand if you don't want to maintain/integrate these changes, but it might help others being stuck in the same situation. I've also did it so that the default behavior is kept and only replaced when the operation fails.
In theory, it would be better to only use the _ext version, because the current version depends on version names (Community, Professional) as well as requireing manual edits to it, everytime there is a new VS Version.
My code also requires vs_version_to_year to be adjusted, but maybe after thorough code review knowing the actual version isn't even important anymore (apart from paths).

There are problems (apart from the "political" decision about maintenance and whethere this code should be touched or not):
 - I forgot that get_msbuild_version uses an even stranger version format instead of following the VS Version ("Current", but then also supporting "14.0", which isn't supported by find_highest_msvc_version). I will force-push this asap, but to me that feels odd.
 - mach.bat still uses a hardcoded version and runs vcvarsall.bat to set quite a few environmental variables. How is this code then required? Or put differently: Should mach.bat just rely on this code? I can only tell that locally running in mozilla-build, everything works, I use ./mach instead of mach.bat, though I cannot test mach.bat because it WOULD find vcvars from my 2015 install, which doesn't contain MSBuild though. Testing that would help me a lot.
 - I've read somewhere that Servo requires the Graphics Workload, but I couldn't find that in the docs. Maybe someone knows this cleary? Because then we could add `Microsoft.VisualStudio.Component.Graphics.Tools` to vswhere, so that only Installations with the required workloads are considered.

saschanaz:
I think the existing detection code isn't needed if vswhere is used.

jdm:
The mozjs build failure happens when your CARGO_HOME directory is on a different drive than your servo working directory.

python/servo/command_base.py
```diff
@@ -1059,7 +1078,12 @@ def find_highest_msvc_version():
             vsinstalldir = os.path.join(base_vs_path, version, edition)
             if os.path.exists(vsinstalldir):
                 return (vsinstalldir, vs_version, msbuild_version)
-    print("Can't find MSBuild.exe installation under %s." % base_vs_path)
+
+    vers = find_highest_msvc_version_ext()
+    for version in sorted(vers, key=lambda tup: float(tup[1])):
+        return version
+    print("Can't find MSBuild.exe installation under %s. Try so specify the VSINSTALLDIR and VisualStudioVersion" +
```
jdm: Let's say "Please set the VSINSTALLDIR and VisualStudioVersion environment variables." instead.

python/servo/command_base.py
```diff
@@ -1059,7 +1078,12 @@ def find_highest_msvc_version():
             vsinstalldir = os.path.join(base_vs_path, version, edition)
             if os.path.exists(vsinstalldir):
                 return (vsinstalldir, vs_version, msbuild_version)
-    print("Can't find MSBuild.exe installation under %s." % base_vs_path)
+
+    vers = find_highest_msvc_version_ext()
+    for version in sorted(vers, key=lambda tup: float(tup[1])):
+        return version
```
jdm: Let's write this as:
```python
versions = sorted(vers, key=lambda tup: float(tup[1]))
if not versions:
    print(...)
    sys.exit(1)
return versions[0]
```

jdm (approved):

python/servo/command_base.py
MeFisto94: Should I replace vers here by the direct function call?

jdm: I don't mind the way it is.

MeFisto94:
What about the two remaining things (which could be follow ups to this PR):
- Does servo require the graphics workload?
- mach.bat?

jdm:
I am not aware of any requirement for the graphics workload. It's not clear to me what we can do differently with mach.bat.

saschanaz:
>The mozjs build failure happens when your CARGO_HOME directory is on a different drive than your servo working directory.

Is this a reply for my comment? Not sure how relevant it is...

jdm:
No, it's a reply to the part of the earlier comment that mentions a failure to build mozjs_sys.

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24762

jdm:
@bors-servo retry
* #24762

jdm:
@bors-servo retry
* #24762

jdm:
@bors-servo retry
* #23290 #24762

MeFisto94:
Speaking of mach.bat, in my opinion, we could delete everything before https://github.com/servo/servo/blob/79408fa36bae63eef07c5bbb5c92d40d3193d808/mach.bat#L49

The reason is, that this searches for visual studio installations the old/legacy way and calls vcvars.bat. 
I *suspect* this is so that MSBUILD is on PATH.
This will, however, fail early and thus not profit from the improved detection.
At the same time it's unnecessary to double-check for Visual Studio Installations.

In my experiments, having vcvars.bat excluded didn't break anything, especially since the code already locates the path to VS and uses that one (no need to have them on PATH, wouldn't work for me anyway [path limited to 1024 chars...]).

So we could do one of two things (or leave it as it is, of course):
 - Do thorough testing and exclude the vcvars.bat and see if the build breaks under any condition for anyone in their specific environment
 - Just continue with that legacy without thinking if it's required and just execute the .bat file from the python file (short after find_highest_msvc_version).

jdm:
@bors-servo retry
* network failure

PR #25266, opened 2019-12-12
WIP: Devtools prompt

paulrouget:
Depends on #25242

Ask the user before accepting devtools incoming connection.

PR #25242, opened 2019-12-11
Mechanism to allow Servo to prompt the user

paulrouget:
This blocks the embedder thread (compositor thread). Not ideal. I don't think it's too much work to only block script, I'll do that in a follow up bug.

Fix #23376 

@Manishearth we have a few new APIs. Hopefully this will cover your needs. A thing I haven't implemented yet is a way to ask the user to pick from a list. Let me know if it's something you'll need.

Issue #25119, opened 2019-12-05
Make it possible to start the devtools after startup


paulrouget:

jdm:
This has interesting implications, because every script thread has a direct channel to the devtools server, and is responsible for sending notifications about new globals. One possibility is that we could have the channels always exist and slowly fill up over time with nothing reading them; this would allow us to start the devtools server later and read the backlog of messages to end up in a reasonable state.

The other option would be to have script threads store an optional channel, and when it's time to start a server then the constellation distributes a channel to all script threads, who then send notifications about the globals that currently exist. We might need to add some kind of buffering somewhere for console.log output/JS exceptions/etc. that could be sent when connecting to a global, as well, so that developers are able to see output that occurred before the devtools server started?

jdm:
We might be able to make it easier by having the script thread always talk to the constellation instead of directly to the devtools, which can then make decisions about buffering or discarding in a single central location.

PR #25307, opened 2019-12-16
Enable devtools by default on hololens.

jdm:


jdm:
No point in doing this before #25266.

paulrouget:
Do you foresight any potential performance impact?

Otherwise, r=me.

jdm:
I would be surprised by performance impact, since the devtools thread just waits on a channel and there's no default messages that are sent every frame.

Manishearth:
A slight concern I have here is that this means that content can spam permissions prompts and lock up the browser by sending requests to localhost. Perhaps we can make this a setting or something? It's fine right now when it's just us using this.

paulrouget:
@Manishearth we can’t start the devtools after servo started. Yet (#25119). So a setting would require a restart.

No ideal solution until we get the delayed devtools startup.

I think this is an ok short term solution.

But in general, we need a way to prevent any dialog to repeatedly be prompted: https://github.com/servo/servo/issues/25706

paulrouget:
Landed as part of #25242

Issue #18194, opened 2017-08-22
Servo user agent should list user's actual operating system


atouchet:
Currently the Windows and macOS versions listed in Servo's user agent are hardcoded to Windows NT 6.1 and Intel Mac OS X 10.10 respectively. Presumably this should be changed at some point to list the user's actual OS version instead.

SimonSapin:
Why should it? Wouldn’t that only help fingerprinting?

atouchet:
To some degree it probably would but I don't see that making a major difference. Firefox lists the user's correct OS so I don't see why Servo shouldn't. I believe this was discussed when the user agent strings were first added: https://github.com/servo/servo/pull/7143#issuecomment-129859641

atouchet:
This probably applies to iOS now as well as a user agent string was added for that OS in #18205.

atouchet:
This is causing issues with https://my.panomoments.com/ which requires Windows 8 or newer to run and Servo says it's using Windows 7.

jdm:
Oh, that's really interesting. Thanks for pointing that out!

MeFisto94:
Actually, I'm thinking about implementing this.

Does servo have preferences? If so we could add one `servo.spoof-os-version` (naming arbitrary), but do note that this only helps when you're part of the majority. Windows 7 will soon reach it's EOL, then not having Windows 10 as User Agent would make it more unique again, thus this pref only helps when running a less-than-majority OS and the pref being regularly updated to the OS used *the most* among *Servo*(!) Users. Anyway, @atouchet is right, this has been talked about in the linked comment and is useful for things like forwarding to the Windows Store, when the Version is >= 8.1

I've looked into https://github.com/DarkEld3r/os_info as a crate, but it has one major "issue":
It uses https://docs.microsoft.com/de-de/windows/win32/devnotes/rtlgetversion?redirectedfrom=MSDN over https://docs.microsoft.com/de-de/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa?redirectedfrom=MSDN, which would allow for UWP Support.

Also see the Remark there:
```
RtlGetVersion is the equivalent of the GetVersionEx function in the Windows SDK. See the example in the Windows SDK that shows how to get the system version.
```

We can try to talk to the crate owner to see what we can do or just implement this one API call on our own, but I think the former method is better, as it'd also come with Mac OS X Version Detection.

jdm:
[...]

PR #25308, opened 2019-12-16
Use Windows 10 instead of 7 in user agent string

atouchet:
I tried testing https://my.panomoments.com/ in the latest nightly version of Servo but it wouldn't load because it requires Windows 8 or newer and Servo's user agent string says it's using Windows 7. This change should allow sites requiring a modern version of Windows to run. This could potentially cause issues for users running older versions of Windows however. A better long term solution would be to do #18194.

---


jdm:
@bors-servo retry
* #24762 #23290

jdm:
@bors-servo retry
* #24762

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24762

PR #25310, opened 2019-12-16
Attr is a Node, with consequences for many Node methods

### LINKED ISSUE #25124: Attr does not implement interface Node
WPT dom/nodes/Document-importNode, last test.

doc is created using document.implementation.createHTMLDocument. setAttributeNS and getAttributeNodeNS are used on it, and then the output of getAttributeNodeNS is passed to document.importNode, which doesn't believe it's a Node. Comparing Servo's IDL to https://dom.spec.whatwg.org/#interface-attr it looks like Servo has implemented Node attributes directly into attr without inheritance.

pshaughn:
Attr is now a Node, as current WHATWG specs require. I think I did some unidiomatic things to make compareDocumentPosition work and it could use a look by someone more familiar with how to write concise Rust code. I also think the new cases in compareDocumentPosition lack tests; it is possible to compare the position of two attributes, or of an attribute and an element, and I don't think any tests are exercising that functionality.

---

- [X] These changes fix #25124

- [ ] There are tests for these changes (existing cases of Node methods are well-tested, and there is a WPT test specifically for whether Attr is Node, but I'm not sure about new Node method cases)

### DIFF
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -5,7 +5,9 @@
 //! The core DOM types. Defines the basic DOM hierarchy as well as all the HTML elements.
 
 use crate::document_loader::DocumentLoader;
+use crate::dom::attr::Attr;
 use crate::dom::bindings::cell::DomRefCell;
+use crate::dom::bindings::codegen::Bindings::AttrBinding::AttrMethods;
 use crate::dom::bindings::codegen::Bindings::CharacterDataBinding::CharacterDataMethods;
 use crate::dom::bindings::codegen::Bindings::DocumentBinding::DocumentMethods;
 use crate::dom::bindings::codegen::Bindings::ElementBinding::ElementMethods;
@@ -1764,6 +1766,7 @@ impl Node {
             NodeTypeId::CharacterData(CharacterDataTypeId::ProcessingInstruction) |
             NodeTypeId::CharacterData(CharacterDataTypeId::Comment) => (),
             NodeTypeId::Document(_) => return Err(Error::HierarchyRequest),
+            NodeTypeId::Attr => unreachable!(),
         }
 
         // Step 6.
@@ -1833,6 +1836,7 @@ impl Node {
                 },
                 NodeTypeId::CharacterData(_) => (),
                 NodeTypeId::Document(_) => unreachable!(),
+                NodeTypeId::Attr => unreachable!(),
             }
         }
         Ok(())
@@ -2108,6 +2112,19 @@ impl Node {
                 );
                 DomRoot::upcast::<Node>(doctype)
             },
+            NodeTypeId::Attr => {
+                let attr = node.downcast::<Attr>().unwrap();
+                let attr = Attr::new(
+                    &document,
+                    attr.local_name().clone(),
+                    attr.value().clone(),
+                    attr.name().clone(),
+                    attr.namespace().clone(),
+                    attr.prefix().cloned(),
+                    None,
+                );
+                DomRoot::upcast::<Node>(attr)
+            },
             NodeTypeId::DocumentFragment(_) => {
                 let doc_fragment = DocumentFragment::new(&document);
                 DomRoot::upcast::<Node>(doc_fragment)
@@ -2235,6 +2252,12 @@ impl Node {
     pub fn locate_namespace(node: &Node, prefix: Option<DOMString>) -> Namespace {
         match node.type_id() {
             NodeTypeId::Element(_) => node.downcast::<Element>().unwrap().locate_namespace(prefix),
+            NodeTypeId::Attr => node
+                .downcast::<Attr>()
+                .unwrap()
+                .GetOwnerElement()
+                .as_ref()
+                .map_or(ns!(), |elem| elem.locate_namespace(prefix)),
             NodeTypeId::Document(_) => node
                 .downcast::<Document>()
                 .unwrap()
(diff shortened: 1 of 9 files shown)


components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
 
-        // FIXME(emilio): This will eventually need to handle attribute nodes.
+        // step 2
+        let node1: Option<&Node> = other;
+        let node2: Option<&Node> = Some(self);
 
-        let mut self_and_ancestors = self
+        // step 3
+        let attr1: Option<&Attr>;
+        let attr2: Option<&Attr>;
```
Manishearth: Just get rid of these and have `let`s below, leaving a `step 3: nop` comment.

Or, make this `let mut attr1 = None`.

components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
+    fn CompareDocumentPosition(&self, other: Option<&Node>) -> u16 {
         // step 1.
-        if self == other {
+        if Some(self) == other {
             return 0;
         }
 
-        // FIXME(emilio): This will eventually need to handle attribute nodes.
+        // step 2
+        let node1: Option<&Node> = other;
```
Manishearth: The type signatures are unnecessary here

components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
 
-        let mut self_and_ancestors = self
+        // step 3
+        let attr1: Option<&Attr>;
+        let attr2: Option<&Attr>;
+
+        // step 4
+        attr1 = match node1 {
+            None => None,
+            Some(n1) => match n1.downcast::<Attr>() {
```
Manishearth: This can just be `Some(n) => n.downcast::<Attr>()`

No need to number it too.

components/script/dom/node.rs
Manishearth: Yeah I think it's better if you `let mut attr1 = None` and then redo the code below to closer match the spec text's structure. So step 4 would be

```rust
if let Some(a) = node2.unwrap().downcast::<Attr>() {
    attr2 = Some(a);
    node2 = a.GetOwnerElement().map(|e| e.upcast());
}
```

Step 5 would be similar, with an additional subpart for 5.1

Manishearth: That should also fix any lifetime issues

components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
+        };
+
+        let node2 = match attr2owner {
+            None => node2,
+            Some(ref root2) => Some(&*root2.upcast()),
+        };
+
+        // Step 5.2
+        // This substep seems lacking in test coverage.
+        if node1 == node2 && attr1 != None && attr2 != None {
```
Manishearth: use `.is_some()`

Manishearth: Actually this is better written as 

```rust
match (attr1, attr2) {
  (Some(a1), Some(a2)) => {
       // actual code
   }
   _ => ()
}
```

Avoid unwraps wherever possible

components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
+                        NodeConstants::DOCUMENT_POSITION_FOLLOWING;
+                }
+            }
+            // both attrs have node2 as their owner element, so
+            // we can't have left the loop without seeing them
+            unreachable!();
+        }
+
+        // Step 6
+        if node1 == None {
```
Manishearth: `.is_none()`

Manishearth (changes requested):
Would be nice if you could add tests for the CompareDocumentPosition stuff, but not necessary.

components/script/dom/node.rs
pshaughn: I got compiler errors when I took them off the attrs, but they are unnecessary on the nodes and removed now.

components/script/dom/node.rs
pshaughn: I had lifetime errors with your suggested changes and ended up going back to holding the GetOwnerElement results in their own variables, but with most of your syntax suggestions.

components/script/dom/node.rs
```diff
@@ -2726,18 +2779,109 @@ impl NodeMethods for Node {
...
+                Some(ref e) => Some(&*e.upcast()),
+                None => None,
+            }
+        }
+
+        // Step 5.2
+        // This substep seems lacking in test coverage.
+        // We hit this when comparing two attributes that have the
+        // same owner element.
+        if !node1.is_none() && node1 == node2 {
```
Manishearth: wrap the whole thing in `if let Some(node1) = node1` so you don't have to unwrap

Manishearth: (also, for future reference, `!is_none` is `is_some()`)

pshaughn: Will do!

Manishearth:
@bors-servo r+ 

would like to also see WPT tests for all the subtleties of compareDocumentPosition, but not a big deal

pshaughn:
> would like to also see WPT tests for all the subtleties of compareDocumentPosition, but not a big deal

Me too! I filed https://github.com/web-platform-tests/wpt/issues/20881. The case is complex enough that I'm worried if I write the tests myself, I might end up writing a test that matches my own misunderstanding instead of the spec.

pshaughn:
Oh, I just looked at those test failures; I've got a lot more test metadata to update!

pshaughn:
What's still left for me to do here?

jdm:
@bors-servo retry
* #25454

jdm:
@bors-servo retry
* #25459

PR #25312, opened 2019-12-17
Do not run faster than 60 FPS

### LINKED ISSUE #25305: Useless calls from the embedder
Follow up of #25263.

The number of calls from the embedder to Servo can be absurdly high. 300K+ per seconds.

The way we let the embedder know about the running animation state and run animations is spaghetti-esque. A mix of #21855 and #20379 issues.

This is how we would expect animations to work:

1. animation happens Layout
2. let the compositor know
3. let the embedder know
4. embedder runs full speed and block on swapbuffer.

But:

Step 2, letting the compositor know, triggers the compositor proxy to wake up the embedder loop which will run a `compositor.perform_update()` call which triggers compositing.
Now goes step 3, animating is set to animating.
Step 4, compositor won't swap the buffer because nothing needs to be done, compositing has already happened in step 2.

Adding to that the fact that AnimationState is sent at each tick, and we don't end the animation properly, and there is nothing to animate, the embedder calls perform_update full speed, and block on nothing, ending up with 300K+ calls per seconds.

paulrouget:
Fix #25305

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -1478,9 +1478,9 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         true
     }
 
-    pub fn perform_updates(&mut self) -> bool {
+    pub fn perform_updates(&mut self) {
         if self.shutdown_state == ShutdownState::FinishedShuttingDown {
-            return false;
+            return;
         }
 
         // If a pinch-zoom happened recently, ask for tiles at the new resolution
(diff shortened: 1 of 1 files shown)


atouchet:
Does this limit Servo to running at 60 FPS? Would it make sense to instead use the display's refresh rate?

paulrouget:
If the monitor is cadenced at 120 FPS, it will force Servo to run at 60 FPS.

We could rely on https://docs.rs/winit/0.20.0-alpha4/winit/monitor/struct.VideoMode.html#method.refresh_rate

We should also maybe only sleep if not call to `compositor::composite` happen.

This is an improvement non the less.

atouchet:
> If the monitor is cadenced at 120 FPS, it will force Servo to run at 60 FPS.

What about displays that don't run at multiples of 60 (ex. 75, 90, 144, etc.)?

paulrouget:
Servo will always run at 60FPS max.

This is not the best approach. Open to suggestions.

atouchet:
Using that winit method instead of assuming 60 seems like a reasonable idea to me.

Manishearth:
cc @asajeffrey thoughts on this?

Manishearth (approved):
Approach seems fine by me but I'm not sure if this is the right way to do this.

asajeffrey:
Can we avoid blocking the main thread? The embedder may have things they want to do on the main thread at more than 60fps. If nothing else we should make the number 60 configurable, e.g. for devices that run at 120fps.

paulrouget:
Closing this as this starting to look more complex that I initially imagined. See #25305.

PR #25337, opened 2019-12-19
Implement mach bootstrap-gstreamer for Windows

### LINKED ISSUE #25335: Support mach bootstrap-gstreamer on Windows
Currently installing gstreamer is a bit tedious on Windows:
Multiple msi's have to be installed, a package manager (chocolatey) has to be installed or pkg-conf-lite has to be installed manually and the paths have to be adjusted.

In my case, GStreamer was suggesting me E:\gstreamer as installation location, which is where pkg-config isn't looking. In fact it only seemed to look in it's own directory (provided servo doesn't pass a custom hardcoded path to where GStreamer usually lands, I suspect C:\gstreamer).

```bash
wget https://downloads.sourceforge.net/project/pkgconfiglite/0.28-1/pkg-config-lite-0.28-1_bin-win32.zip
unzip pkg-config-lite-0.28-1_bin-win32.zip -d pkg-config
cd pkg-config/pkg-config-lite-0.28-1/ && mv * ../ && cd ../ && rm -rf pkg-config-lite-0.28-1 && cd ../
```
This is the script I used to install pkg-config into `.servo/msvc-dependencies`, which would be Pull Request one: Adding this to mach as well as properly outputting an error when os != linux (and windows, so soon: when os == mac), because right now, `mach bootstrap-gstreamer` doesn't output anything when not on Linux.

This made servo build using
`PATH=$PATH:~/.cargo/bin:$(pwd)/.servo/msvc-dependencies/pkg-config/bin PKG_CONFIG_PATH="/e/gstreamer/1.0/x86_64/lib/pkgconfig" mach build -d`

So far so good, but then I hit 
```
 Finished release [optimized] target(s) in 39m 29s
Der Befehl "editbin" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
Packaging EGL DLLs
Packaging gstreamer DLLs
Could not find GStreamer installation directory.
Packaging MSVC DLLs
DLL file `api-ms-win-crt-runtime-l1-1-0.dll` not found!
Build FAILED in 0:39:29
```
Which is something mach does, I have to look into where it tries to locate GStreamer and why and how we could do that when installing gstreamer into msvc-dependencies (Or look in the PKG_CONFIG_PATH)

MeFisto94:
Download and Install GStreamer's MSIs into .servo/msvc-dependencies

---
- [X] These changes fix #25335 
- [X] These changes do not require tests because mach commands are untested currently

### DIFF
--- a/python/servo/bootstrap.py
+++ b/python/servo/bootstrap.py
@@ -94,18 +94,41 @@ def install_salt_dependencies(context, force):
         print("Dependencies are already installed")
 
 
-def gstreamer(context, force=False):
+def gstreamer_linux(context, force=False):
     cur = os.curdir
     gstdir = os.path.join(cur, "support", "linux", "gstreamer")
     if not os.path.isdir(os.path.join(gstdir, "gst", "lib")):
         subprocess.check_call(["bash", "gstreamer.sh"], cwd=gstdir)
-        return True
-    return False
-
-
-def bootstrap_gstreamer(context, force=False):
-    if not gstreamer(context, force):
+    else:
         print("gstreamer is already set up")
+    return 0
+
+
+def gstreamer_windows(context, force=False):
+    dependencies = os.path.join(context.sharedir, "msvc-dependencies")
+    gstdir = os.path.join(dependencies, "gstreamer", "1.0")
+
+    if not os.path.exists(os.path.join(gstdir, "X86_64", "bin", "ffi-7.dll")):
+        # Please keep this up to date
+        gst_url1 = "https://gstreamer.freedesktop.org/data/pkg/windows/1.16.0/gstreamer-1.0-msvc-x86_64-1.16.0.msi"
+        gst_url2 = "https://gstreamer.freedesktop.org/data/pkg/windows/1.16.0/gstreamer-1.0-devel-msvc-x86_64-1.16.0.msi"
+
+        for url in [gst_url1, gst_url2]:
+            filename = url[url.rfind("/") + 1:]
+            filepath = os.path.join(dependencies, filename)
+            if not os.path.isfile(filepath):
+                download_file("GStreamer", url, filepath)
+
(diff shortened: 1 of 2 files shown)


MeFisto94:
Two things might require addressing here:
1.  This approach will only try to download and install GStreamer if the folder is NOT present. Bumping the GStreamer Version OR cancelling the process somewhere inbetween the installation would require a manual deletion of this folder, unless we would be checking for individual files in that folder, which is also intermittent (depending on where the installation was interrupted).
https://github.com/servo/servo/blob/b49e74b1f277587b4899fb8feea6bb50d3a41312/python/servo/command_base.py#L259 is using it like this, though.

2. This approach currently doesn't delete the msi files upon successful installation. This is an oversight, and will be force-pushed once 1 is solved. I don't know if deinstallation will work when the MSI is deleted, I will try that before that.

MeFisto94:
So all the issues should be resolved here:
1. has been implemented the same way as other approaches
2. Deleted MSI files, this breaks the "change" button (where one can change the feature set), but activates the uninstall button (which doesn't work when change is available)
3. has been changed

Remaining:
ARM64 and X86 MSIs are missing, but then I see "# We don't build gstreamer for non-x86_64 / android yet" in command_base.py, which means get_gstreamer_root could even be simplified (or gstreamer could be supported on those platforms in the future)

I tried to add detection of the sharedir (msvc_dependencies) to get_gstreamer_root, but gave up on that, as the method is called from quite a few places and passing the sharedir from everywhere complicates review for little gain (restarting the shell once)

Also note the changes to bootstrap() itself: Previously, when a given bootstrap task was unknown, the platform default one was invoked. This is however incorrect. If there'd be a command `bootstrap rust`, then I wouldn't want it to bootstrap `python`, just because the code is faulty (usually commands fail earlier when they are not registered, so typos are caught etc).

mrobinson:
This should now be fixed by #29781.

PR #25338, opened 2019-12-19
Remove Optional pipeline_id

### LINKED ISSUE #24772: Optional pipeline in WorkerScriptLoadOrigin should not be optional
All code that creates a WorkerScriptLoadOrigin passes a Some value for the pipeline, so we should remove the Option from the struct member field.

Code:
* `components/script_traits/lib.rs`
* `components/script/dom/serviceworkerregistration.rs`
* `components/script/dom/worker.rs`

No need to run tests - if it builds, it's good enough for a pull request.

warren-fisher:
All code that creates a WorkerScriptLoadOrigin passes a Some value for the pipeline, so we should remove the Option from the struct member field. 

I changed the `WorkerScriptLoadOrigin` struct in `components/script_traits/lib.rs` to have a PipelineId and not an Option<PipelineId>. In `components/script/dom/worker.rs` and `components/script/dom/serviceworkerregistration.rs` it was changed so that `WorkerScriptLoadOrigin` was instantiated with a non-Optional. In `components/script/dom/dedicatedworkerglobalscope.rs` testing for if pipeline_id is an Optional is removed. 

In `components/script/dom/serviceworkerglobalscope.rs` and `components/script/dom/dedicatedworkerglobalscope.rs` a `PipelineId` was provided as an Optional to a `RequestBuilder` and I changed it to provide a `Some(pipeline_id)` instead. I am not sure changing the [RequestBuilder struct](https://github.com/warren-fisher/servo/blob/5c3bda025144118fc8dd67a3f975d8f91fd25e78/components/net_traits/request.rs#L118-L159) at line 147 to accept a non-optional was within this issue so I left it as is.

I was able to successfully build it so I assume these changes worked, however I am not sure about the implications of the changes made in the unsafe block in `components/script/dom/dedicatedworkerglobalscope.rs`.

- [x] These changes fix #24772 (GitHub issue number if applicable)

- [x] These changes do not require tests because the issue says so

### DIFF
--- a/components/script/dom/dedicatedworkerglobalscope.rs
+++ b/components/script/dom/dedicatedworkerglobalscope.rs
@@ -320,24 +320,20 @@ impl DedicatedWorkerGlobalScope {
                     .credentials_mode(CredentialsMode::CredentialsSameOrigin)
                     .parser_metadata(ParserMetadata::NotParserInserted)
                     .use_url_credentials(true)
-                    .pipeline_id(pipeline_id)
+                    .pipeline_id(Some(pipeline_id))
                     .referrer(referrer)
                     .referrer_policy(referrer_policy)
                     .origin(origin);
 
                 let runtime = unsafe {
-                    if let Some(pipeline_id) = pipeline_id {
-                        let task_source = NetworkingTaskSource(
-                            Box::new(WorkerThreadWorkerChan {
-                                sender: own_sender.clone(),
-                                worker: worker.clone(),
-                            }),
-                            pipeline_id,
-                        );
-                        new_child_runtime(parent, Some(task_source))
-                    } else {
-                        new_child_runtime(parent, None)
-                    }
+                    let task_source = NetworkingTaskSource(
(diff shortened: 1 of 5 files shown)


components/script/dom/dedicatedworkerglobalscope.rs
```diff
@@ -320,24 +320,20 @@ impl DedicatedWorkerGlobalScope {
...
-                        new_child_runtime(parent, None)
-                    }
+                    let task_source = NetworkingTaskSource(
+                        Box::new(WorkerThreadWorkerChan {
+                            sender: own_sender.clone(),
+                            worker: worker.clone(),
+                        }),
+                        pipeline_id,
+                    );
+                    new_child_runtime(parent, Some(task_source))
```
warren-fisher: Not sure if changing this unsafe block is even ok.

jdm:
This is what I expected to see. Thanks for simplifying this code!

@bors-servo r+

PR #25343, opened 2019-12-19
Do not wait for rAF to be requested in render loop

Manishearth:
Servo-side of https://github.com/servo/webxr/pull/114

r? @jdm

### DIFF
--- a/components/script/dom/xrsession.rs
+++ b/components/script/dom/xrsession.rs
@@ -348,6 +348,8 @@ impl XRSession {
         if self.is_immersive() {
             base_layer.swap_buffers();
             self.session.borrow_mut().render_animation_frame();
+        } else {
+            self.session.borrow_mut().start_render_loop();
         }
 
         // If the canvas element is attached to the DOM, it is now dirty,
--- a/components/script/dom/xrsession.rs
+++ b/components/script/dom/xrsession.rs
@@ -41,7 +41,7 @@ use crate::dom::xrwebgllayer::XRWebGLLayer;
 use crate::task_source::TaskSource;
 use dom_struct::dom_struct;
 use euclid::{Rect, RigidTransform3D, Transform3D};
-use ipc_channel::ipc::IpcSender;
+use ipc_channel::ipc::IpcReceiver;
 use ipc_channel::router::ROUTER;
 use metrics::ToMs;
 use profile_traits::ipc;
@@ -73,8 +73,6 @@ pub struct XRSession {
     next_raf_id: Cell<i32>,
     #[ignore_malloc_size_of = "closures are hard"]
     raf_callback_list: DomRefCell<Vec<(i32, Option<Rc<XRFrameRequestCallback>>)>>,
-    #[ignore_malloc_size_of = "defined in ipc-channel"]
-    raf_sender: DomRefCell<Option<IpcSender<Frame>>>,
     input_sources: Dom<XRInputSourceArray>,
     // Any promises from calling end()
     #[ignore_malloc_size_of = "promises are hard"]
@@ -108,7 +106,6 @@ impl XRSession {
 
             next_raf_id: Cell::new(0),
             raf_callback_list: DomRefCell::new(vec![]),
-            raf_sender: DomRefCell::new(None),
             input_sources: Dom::from_ref(input_sources),
             end_promises: DomRefCell::new(vec![]),
             ended: Cell::new(false),
@@ -116,7 +113,12 @@ impl XRSession {
         }
     }
 
-    pub fn new(global: &GlobalScope, session: Session, mode: XRSessionMode) -> DomRoot<XRSession> {
+    pub fn new(
+        global: &GlobalScope,
+        session: Session,
+        mode: XRSessionMode,
+        frame_receiver: IpcReceiver<Frame>,
+    ) -> DomRoot<XRSession> {
         let ivfov = if mode == XRSessionMode::Inline {
             Some(FRAC_PI_2)
         } else {
@@ -136,7 +138,7 @@ impl XRSession {
         );
         input_sources.set_initial_inputs(&ret);
         ret.attach_event_handler();
-        ret.setup_raf_loop();
+        ret.setup_raf_loop(frame_receiver);
         ret
     }
 
@@ -153,21 +155,15 @@ impl XRSession {
         self.mode != XRSessionMode::Inline
     }
 
-    fn setup_raf_loop(&self) {
-        assert!(
-            self.raf_sender.borrow().is_none(),
-            "RAF loop already set up"
-        );
+    fn setup_raf_loop(&self, frame_receiver: IpcReceiver<Frame>) {
         let this = Trusted::new(self);
         let global = self.global();
         let window = global.as_window();
         let (task_source, canceller) = window
             .task_manager()
             .dom_manipulation_task_source_with_canceller();
-        let (sender, receiver) = ipc::channel(global.time_profiler_chan().clone()).unwrap();
-        *self.raf_sender.borrow_mut() = Some(sender);
         ROUTER.add_route(
-            receiver.to_opaque(),
+            frame_receiver.to_opaque(),
             Box::new(move |message| {
                 let this = this.clone();
                 let _ = task_source.queue_with_canceller(
@@ -179,15 +175,7 @@ impl XRSession {
             }),
         );
 
-        self.request_new_xr_frame();
-    }
-
-    /// Requests a new https://immersive-web.github.io/webxr/#xr-animation-frame
-    ///
-    /// This happens regardless of the presense of rAF callbacks
-    fn request_new_xr_frame(&self) {
-        let sender = self.raf_sender.borrow().clone().unwrap();
-        self.session.borrow_mut().request_animation_frame(sender);
+        self.session.borrow_mut().start_render_loop();
     }
 
     pub fn is_outside_raf(&self) -> bool {
@@ -361,7 +349,6 @@ impl XRSession {
             base_layer.swap_buffers();
             self.session.borrow_mut().render_animation_frame();
         }
-        self.request_new_xr_frame();
 
         // If the canvas element is attached to the DOM, it is now dirty,
         // and we need to trigger a reflow.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6450,7 +6450,7 @@ dependencies = [
 [[package]]
 name = "webxr"
 version = "0.0.1"
-source = "git+https://github.com/servo/webxr#b79870164f518d90268aaaac1fb83059b83dceda"
+source = "git+https://github.com/servo/webxr#c6abf4c60d165ffc978ad2ebd6bcddc3c21698e1"
 dependencies = [
  "bindgen",
  "euclid",
@@ -6471,7 +6471,7 @@ dependencies = [
 [[package]]
 name = "webxr-api"
 version = "0.0.1"
-source = "git+https://github.com/servo/webxr#b79870164f518d90268aaaac1fb83059b83dceda"
+source = "git+https://github.com/servo/webxr#c6abf4c60d165ffc978ad2ebd6bcddc3c21698e1"
 dependencies = [
  "euclid",
  "ipc-channel",
--- a/components/script/dom/xr.rs
+++ b/components/script/dom/xr.rs
@@ -26,14 +26,14 @@ use crate::dom::xrtest::XRTest;
 use crate::script_thread::ScriptThread;
 use crate::task_source::TaskSource;
 use dom_struct::dom_struct;
-use ipc_channel::ipc::IpcSender;
+use ipc_channel::ipc::{self as ipc_crate, IpcReceiver, IpcSender};
 use ipc_channel::router::ROUTER;
 use profile_traits::ipc;
 use std::cell::Cell;
 use std::rc::Rc;
 use webvr_traits::{WebVRDisplayData, WebVRDisplayEvent, WebVREvent, WebVRMsg};
 use webvr_traits::{WebVRGamepadData, WebVRGamepadEvent, WebVRGamepadState};
-use webxr_api::{Error as XRError, Session, SessionMode};
+use webxr_api::{Error as XRError, Frame, Session, SessionMode};
 
 #[dom_struct]
 pub struct XR {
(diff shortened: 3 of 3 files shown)


components/script/dom/xrsession.rs
```diff
@@ -153,21 +155,15 @@ impl XRSession {
...
-            self.raf_sender.borrow().is_none(),
-            "RAF loop already set up"
-        );
+    fn setup_raf_loop(&self, frame_receiver: IpcReceiver<Frame>) {
         let this = Trusted::new(self);
         let global = self.global();
         let window = global.as_window();
         let (task_source, canceller) = window
             .task_manager()
             .dom_manipulation_task_source_with_canceller();
```
jdm: To reduce the time between when we're signalled with a new frame and when we run the rAF callback, we should use a high-priority task source. The spec doesn't actually specify a particular ordering, so we should be able to get away with jumping the queue. For example, we could use a dedicated sender/receiver pair and use try_recv on it before calling select! in ScriptThread::handle_msgs.

jdm: https://github.com/jdm/servo/tree/xr-frame-prioritize

Manishearth:
The FPS values seem to slightly go up from 12-15 to 15-20.

Timing values (in microseconds, I forgot what a nanosecond was and divided wrong) between the time the openxr backend constructs the frame ("Frame") and the time we trigger callbacks in script ("Now") :

<details>

```
Now: 1720318156	 Frame: 1720311936	Delta: 6219.375
Now: 1720390996	 Frame: 1720389289	Delta: 1707.292
Now: 1720446786	 Frame: 1720442530	Delta: 4256.355
Now: 1720528122	 Frame: 1720523149	Delta: 4973.020
Now: 1720584494	 Frame: 1720575796	Delta: 8697.812
Now: 1720657133	 Frame: 1720655931	Delta: 1202.447
Now: 1720714996	 Frame: 1720713575	Delta: 1421.093
Now: 1720773981	 Frame: 1720772679	Delta: 1302.292
Now: 1720830164	 Frame: 1720828616	Delta: 1547.604
Now: 1720890760	 Frame: 1720889245	Delta: 1515.781
Now: 1720945515	 Frame: 1720943989	Delta: 1525.729
Now: 1721007255	 Frame: 1721005985	Delta: 1269.844
Now: 1721062167	 Frame: 1721059961	Delta: 2205.833
Now: 1721123669	 Frame: 1721122519	Delta: 1150.000
Now: 1721181460	 Frame: 1721178199	Delta: 3261.302
Now: 1721240545	 Frame: 1721239333	Delta: 1211.979
Now: 1721309603	 Frame: 1721303655	Delta: 5947.812
The thread 0x1a7c has exited with code 0 (0x0).
Now: 1721374806	 Frame: 1721372548	Delta: 2258.073
Now: 1721415320	 Frame: 1721413298	Delta: 2022.031
Now: 1721474244	 Frame: 1721472577	Delta: 1667.136
Now: 1721516310	 Frame: 1721515056	Delta: 1253.802
```

</details>

Seems like we're losing around 2-6ms on just the communication delay. Digging further into this

Manishearth:
More timings. `Delta` is between us getting a frame from openxr and us seeing it in `raf_callback` in script. `Delta_task` is between the script process router getting the frame message and the script thread running it. Ignore the last one.

It seems like  we're losing time in waiting for the script router to pick up the IPC message, which bolsters my initial hunch that IPC is slow for some reason.

<details>
```
Now: 30135.365	 Frame: 30133.465	Delta: 1.900	Delta_task: 0.707	Delta_rafc: 0.029
Now: 30187.074	 Frame: 30185.543	Delta: 1.530	Delta_task: 0.237	Delta_rafc: 0.015
Now: 30235.812	 Frame: 30234.467	Delta: 1.345	Delta_task: 0.171	Delta_rafc: 0.014
Now: 30287.797	 Frame: 30285.883	Delta: 1.913	Delta_task: 0.227	Delta_rafc: 0.015
Now: 30335.949	 Frame: 30334.205	Delta: 1.744	Delta_task: 0.399	Delta_rafc: 0.015
Now: 30387.547	 Frame: 30385.756	Delta: 1.789	Delta_task: 0.231	Delta_rafc: 0.016
Now: 30437.031	 Frame: 30435.457	Delta: 1.575	Delta_task: 0.188	Delta_rafc: 0.016
Now: 30487.953	 Frame: 30485.637	Delta: 2.317	Delta_task: 0.230	Delta_rafc: 0.015
Now: 30537.453	 Frame: 30535.949	Delta: 1.505	Delta_task: 0.214	Delta_rafc: 0.012
Now: 30604.316	 Frame: 30602.482	Delta: 1.835	Delta_task: 0.210	Delta_rafc: 0.015
Now: 30657.307	 Frame: 30652.455	Delta: 4.852	Delta_task: 0.705	Delta_rafc: 0.016
Now: 30737.090	 Frame: 30735.533	Delta: 1.555	Delta_task: 0.240	Delta_rafc: 0.016
Now: 30784.846	 Frame: 30783.484	Delta: 1.361	Delta_task: 0.157	Delta_rafc: 0.010
```
</details>

asajeffrey:
2-6ms is a *lot* of extra time!

Manishearth:
So `locate_views` is taking around 0.2ms, `.locate()` is taking around 0.15ms, `sync_actions` is 0.044ms, and the input `.frame()` calls are all together taking `0.6ms` (it's a total of 4 `locate()` calls, essentially).

asajeffrey:
Hmm, that all sounds like stuff that's out of our control? The only thing I can think of is decoupling the calls to input `.frame()`, since if that information is a frame late it probably won't matter so much?

Manishearth:
It might be that we're using the APIs incorrectly, too. I tried using unbounded spaces as recommended, that didn't fix it. Going for jdm's scheduling patch now to see if I can at least fix the communication overhead, which is still a nontrivial chunk.

jdm:
Perhaps we should look into calling begin_frame before render_animation_frame, to be more like the SDK sample at https://github.com/microsoft/OpenXR-SDK-VisualStudio/blob/4378fbda334809852331fccf2ffa9b4979e0684f/samples/BasicXrApp/OpenXrProgram.cpp#L609-L612.

Manishearth:
Timings with @jdm's patch. Delta is the time between frame creation and frame reception in script, delta_task is the time between the script process receiving it and the script thread receiving it. Callback time is the time taken to execute raf callbacks (which we can expect to be somewhat large)


<details>

```
Now: 33839.398	 Frame: 33837.664	Delta: 1.736	Delta_task: 0.172	Callback time: 2.842
Now: 33895.418	 Frame: 33892.207	Delta: 3.213	Delta_task: 1.973	Callback time: 2.785
Now: 33939.176	 Frame: 33937.449	Delta: 1.726	Delta_task: 0.235	Callback time: 3.556
Now: 33993.762	 Frame: 33992.102	Delta: 1.660	Delta_task: 0.149	Callback time: 2.864
Now: 34055.547	 Frame: 34053.891	Delta: 1.654	Delta_task: 0.218	Callback time: 4.446
Now: 34127.477	 Frame: 34125.398	Delta: 2.076	Delta_task: 0.879	Callback time: 3.595
Now: 34187.035	 Frame: 34185.129	Delta: 1.906	Delta_task: 0.329	Callback time: 4.338
Now: 34244.059	 Frame: 34242.164	Delta: 1.893	Delta_task: 0.146	Callback time: 3.102
```
</details>


There are also a bunch of entries with much larger deltas sprinkled throughout. Some are in a sequence:

```
Now: 34850.398	 Frame: 34842.527	Delta: 7.875	Delta_task: 0.152	Callback time: 3.159
Now: 35625.762	 Frame: 35620.098	Delta: 5.662	Delta_task: 0.152	Callback time: 3.357
```

the larger deltas are paired with slightly higher time taken in all the openxr stuff in wait_for_animation_frame, but it doesn't fully account for it. It's possible we're doing a lot of computation elsewhere that's just slowing everything down.

(Perhaps it's because we're still rendering the webpage but not displaying it?)

jdm:
Other interesting things to time include:
* time from sending SwapBuffer message in script to time to process that message in the webgl thread
* time taken to run handle_swap_buffers
* time taken to send signal that swapping is complete back to script
* time taken to complete layout after swapping buffers (https://github.com/servo/servo/blob/master/components/script/dom/window.rs#L1579)

I wonder if we could send the swap buffer message, start layout, wait for layout to complete, then join with the swap buffer completion response to avoid blocking twice sequentially...

jdm:
Mmm, some of those may not actually make sense here; I didn't realize we call [swap_buffers](https://github.com/servo/servo/blob/bac9903fbeed0a394a86c0091e727aada665433d/components/script/dom/xrwebgllayer.rs#L147) directly in immersive mode.

jdm:
That being said, we do run layout, and we do dirty the canvas, and both of those involve blocking the script thread on receiving a response from another thread, so they could contribute to the deltas.

asajeffrey:
Running layout should only block if the document is dirty or animations are running? Which hopefully pages aren't doing once they've entered immersive mode.

jdm:
We mark the canvas dirty right after we swap the base layer's buffer: https://github.com/servo/servo/blob/bac9903fbeed0a394a86c0091e727aada665433d/components/script/dom/xrsession.rs#L366-L372

Manishearth:
which makes sense for the test backend, and not for any other.

asajeffrey:
Ah, that's a holdover from the pre-surfman days.

Manishearth:
We should probably merge these since I was seeing an improvement from 12-15FPS to 15-20FPS.

Manishearth:
Removing the dirty-marking does not impact FPS visibly. Worth doing anyway.

jdm:
This is still marked as a draft, fyi.

Manishearth:
Hm? doesn't seem to be.

But i need to update this to deal with the other one landing

jdm (approved):

Manishearth:
@bors-servo retry

 font cache intermittent, plus something else I'll look at later

jdm:
Oops:
```
  ▶ TIMEOUT [expected OK] /webxr/xrSession_requestAnimationFrame_getViewerPose.https.html
  │ 
  │ 
  │ 
  └ 
  ▶ Unexpected subtest result in /webxr/xrSession_requestAnimationFrame_getViewerPose.https.html:
  │ TIMEOUT [expected PASS] XRFrame getViewerPose updates on the next frame for non-immersive sessions
  └   → Test timed out
  ▶ Unexpected subtest result in /webxr/xrSession_requestAnimationFrame_getViewerPose.https.html:
  └ NOTRUN [expected PASS] XRFrame getViewerPose updates on the next frame for immersive sessions
  ▶ TIMEOUT [expected OK] /webxr/render_state_vertical_fov_inline.https.html
  │ 
  │ 
  │ 
  └ 
  ▶ Unexpected subtest result in /webxr/render_state_vertical_fov_inline.https.html:
  │ TIMEOUT [expected PASS] inlineVerticalFieldOfView is set appropriately on inline sessions
  └   → Test timed out
  ▶ TIMEOUT [expected OK] /webxr/xrStationaryReferenceSpace_floorlevel_updates.https.html
  │ 
  │ 
  │ 
  └ 
  ▶ Unexpected subtest result in /webxr/xrStationaryReferenceSpace_floorlevel_updates.https.html:
  │ TIMEOUT [expected FAIL] 'floor-level' XRStationaryReferenceSpace updates properly when the transform changes for non-immersive sessions
  └   → Test timed out
  ▶ TIMEOUT [expected OK] /webxr/xrSession_viewer_referenceSpace.https.html
```

Manishearth:
Oh, it's possible that the synchronization of mock data no longer works with this.

components/script/dom/xrsession.rs
gterzian: I think that's a good idea, and maybe it could be further improved by not going through the router to receive the frame. 

I think it could be done through:

1. Whenever a callback is added:
    - set a `awaiting_frame` flag on the script-thread.
    - send a `MainThreadScriptMsg::WakeUp` to ensure the main loop doesn't block on the select
2. After the [microtask checkpoint](https://github.com/jdm/servo/blob/871a805e124b11661adbee5dd63f473ed3661f20/components/script/script_thread.rs#L1589):
    - remove the `awaiting_frame` flag. 
    - block on the frame receiver, and 
    - immediately run the callbacks when receiving it. 
3. This would require also running 2 if `sequential` is empty(in the case that the only event is the `MainThreadScriptMsg::WakeUp`, since it will be filtered out by the task-queue). It also requires Xr to always send a message on the frame receiver to unblock the script-thread, even if no frame can be provided(in response to which you'd have to unset the flag on the script-thread). 

So that could be a way to block directly on a `IpcReceiver` inside `handle_msgs`, with the benefit being always running the frame callbacks immediately after you receive the frame, and not going through the router.  

Then I guess the script-thread could make the [`xr_frame_` pair](https://github.com/servo/servo/compare/master...jdm:xr-frame-prioritize#diff-1879ac6bd5d567e2aa43529d33474677R711) an ipc-channel where the sender would be cloned inside `request_new_xr_frame` and send to Xr?

gterzian: Ah ok I see you also need a way to call from the script-thread into the session, so instead of an `awaiting_frame` flag, you could have an `Option<SessionAwaitingFrame>` on the script-thread, which would be set by `request_new_xr_frame` and taken before blocking on the frame receiver from within `handle_msgs`. I assume you can have only one active Xr session per script-thread?

gterzian: Like https://github.com/gterzian/servo/tree/restructure_xr_frame_handling
If it seems like a good idea I could just open a PR as a follow-up.

Manishearth: You can have multiple online sessions, but only one immersive 

We first need to figure out why this is breaking the tests. I bet it's because of a race between data getting sent and read from the mock device.

gterzian: I think the problem with the timeouts is that with the current changes, the call to [`render_animation_frame`](https://github.com/servo/servo/blob/1b5f0ed15a70aa0da88ddc5e67329474f206adc1/components/script/dom/xrsession.rs#L350) from script only occurs if the session is immersive. 

Previously, [a new frame would be requested](https://github.com/servo/servo/blob/106a0185f404c2ce5011272a9191eb71a731e855/components/script/dom/xrsession.rs#L364) at the end of running callbacks, regardless of whether the session was immersive.

Manishearth:
@bors-servo r-

oops, push didn't work

components/script/dom/xrsession.rs
Manishearth: That was indeed the issue, was poking at this earlier. Fixed now.

Opening a followup PR would be great, but we should test if it affects FPS.

PR #25355, opened 2019-12-21
Install pkg-config-lite using mach bootstrap on Windows

MeFisto94:
Support Installing pkg-config for Windows without manual interaction or chocolatey as part of the mach bootstrap process.
---
- [X] These changes don't have a tracking issue
- [X] These changes do not require tests because they are part of the tooling

### DIFF
--- a/python/servo/bootstrap.py
+++ b/python/servo/bootstrap.py
@@ -308,6 +308,26 @@ def windows_msvc(context, force=False):
         else:
             print("done")
 
+    # Until pkg-config-lite is in the depedencies, install it manually
+    dep_dir = os.path.join(deps_dir, "pkg-config")
+    if not os.path.exists(dep_dir):
+        zip_path = os.path.join(deps_dir, "pkg-config.zip")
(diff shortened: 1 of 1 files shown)


MeFisto94:
One of the present but unnecessary steps is to install pkg-config-lite on Windows.
Since I don't know how a progress would be to add them into the dependencies.zip and since this is no custom build or anything, we can directly download the related files from sourceforge to .servo.

I don't know if the comments are appropriate, if not I can change them.
Note that this does only download and unzip pkg-config and that A) mach bootstrap doesn't work at the moment (#25224) and a future PR is required to add pkg-config to the path, otherwise the user has to specify `PATH=$PATH:$(pwd)/.servo/msvc-dependencies/pkg-config/bin` before invoking mach. 
I intentionally decided to split those PRs so it's easier to work on them and review them, though.

jdm:
I recently performed a Servo build on a brand new installation of Windows and did not ever need to set up pkg-config-lite. I'm going to close this PR and remove the requirement from the build setup documentation.

Issue #24903, opened 2019-11-28
Some fetch request headers don't add properly


pshaughn:
A few WPT tests fail because trying to put headers in a request does the wrong thing in some subcases. 

fetch/api/basic/forbidden-headers: Servo stops a script from injecting an Accept-Encoding, Access-Control-Request-Headers, or Access-Control-RequestMethod header by throwing an exception, where the expected behavior is to just silently drop the header.
fetch/api/basic/request-headers-case: Nonstandard request header names are getting case-folded. I think this might be a problem with using Hyperium's implementation of HeaderMap, which is focused on holding a very "normative" form of headers and might not be appropriate for a client that exposes Web APIs.
fetch/api/basic/request-referrer: It looks like trying to set a referrer header to "about:client" sets it to null instead, instead of leaving it as the actual refererrer URL.

pshaughn:
The test cases in fetch/api/basic/request-forbidden-headers.any.js are failing not because of anything to do with those three particular header names, but because the test is trying to set "" as the header values in these cases, which Servo currently disallows but should allow if it wants to pass this and some other WPT tests. https://github.com/servo/servo/blob/bb5cd02da30fe8ef4dcdea5cca17667cd4ac6cdb/components/script/dom/headers.rs#L347-L369

pshaughn:
Servo's implementation of https://fetch.spec.whatwg.org/#dom-request successfully sets the referrer to NetTraitsReferrer::Client, but nothing else in Servo acts on that value. https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer step 3 is what's supposed to use it, but I can't find step 3 in Servo's implementation.

https://github.com/servo/servo/blob/bb5cd02da30fe8ef4dcdea5cca17667cd4ac6cdb/components/net/http_loader.rs#L210-L212 This starts at step 4, and its caller is https://github.com/servo/servo/blob/bb5cd02da30fe8ef4dcdea5cca17667cd4ac6cdb/components/net/fetch/methods.rs#L233-L257 which is written under the impression that the Client value should have been handled at an earlier point.

pshaughn:
[...]

PR #25358, opened 2019-12-21
Header values no longer have to be ASCII or UTF-8

pshaughn:
This passes some failed tests related to header validity when handling ByteStrings outside the printable ASCII range. A few failures remain because the HeaderValue class is stricter than WHATWG/WPT, disallowing various control-code bytes that the spec and tests expect to be allowed.

---

- [X] These changes fix some of the test cases described in #24903 


jdm:
```
  ▶ Unexpected subtest result in /fetch/api/response/response-init-002.html:
  └ PASS [expected FAIL] Testing empty Response Content-Type header
  ▶ Unexpected subtest result in /fetch/api/request/request-headers.html:
  └ PASS [expected FAIL] Testing empty Request Content-Type header
```

pshaughn:
A couple notes about this:
* I _think_ I've taken out all the places where a header value gets parsed as a UTF-8 string without sanity-checking, but I don't know for sure. Hyper HeaderMap does a lot of the sanity-checking for us; for most "typed" headers we can safely cast the value that comes out of the map to a string. Conversion from a header value to a DOMString for Javascript consumption happens via ByteString, which should give a weird but safe "isomorphic decode" from bytes to UTF-16 values 0-255 instead of a UTF-8 parse.
* The "Servo thinks \"{:?}\" is a valid HTTP header value but HeaderValue doesn't." codepath represents a case where different parties still haven't really converged on a spec; Hyper is currently following the RFC rather than WHATWG. In practice I don't think any legitimate website is going to send a header where this particular disagreement matters.

jdm:
@bors-servo r+
Thanks for sorting this out!

jdm:
@bors-servo retry
* #24622

Issue #25336, opened 2019-12-18
mach.bat: Remove manual Visual Studio detection and defer to command_base.py


MeFisto94:
Speaking of mach.bat, in my opinion, we could delete everything before https://github.com/servo/servo/blob/79408fa36bae63eef07c5bbb5c92d40d3193d808/mach.bat#L49

The reason is, that this searches for visual studio installations the old/legacy way and calls vcvars.bat. 
I *suspect* this is so that MSBUILD is on PATH.
This will, however, fail early and thus not profit from the improved detection.
At the same time it's unnecessary to double-check for Visual Studio Installations.

In my experiments, having vcvars.bat excluded didn't break anything, especially since the code already locates the path to VS and uses that one (no need to have them on PATH, wouldn't work for me anyway [path limited to 1024 chars...]).

So we could do one of two things (or leave it as it is, of course):
 - Do thorough testing and exclude the vcvars.bat and see if the build breaks under any condition for anyone in their specific environment
 - Just continue with that legacy without thinking if it's required and just execute the .bat file from the python file (short after find_highest_msvc_version).

_Originally posted by @MeFisto94 in https://github.com/servo/servo/pull/25300#issuecomment-566804761_

jdm:
A good first start would be to see if switching to `python mach` instead of `mach` in https://github.com/servo/servo/blob/master/etc/taskcluster/decision_task.py#L487-L496 passes CI.

atouchet:
This was done in #25365.

PR #25365, opened 2019-12-22
Launch vcvarsall.bat for the recognized VS Installation Directory from python instead of making mach.bat try that on hardcoded paths.

### LINKED ISSUE #25360: Windows: Packaging DLLs fails when WindowsSdkDir isn't set
This is related to #25336: Without vcvars.bat `WindowsSdkDir=C:\Program Files (x86)\Windows Kits\10\` isn't set for me.

This is the case because locating my VS Path fails and thus vcvarsall.bat wasn't found. Also I'm not launching mach.bat at all but the unix version from Mozilla-Build Environment.

To me there are two solutions here:
- https://stackoverflow.com/q/35119223
- Proper solving of #25336 by executing vcvars.bat from within mach.

I'll look into the second approach as that looks like the best (as in most stable and clean) solution to me, even though this specific issue would be easier to solve by looking at the registry.

Edit: Also maybe https://github.com/servo/servo/blob/master/python/servo/build_commands.py#L1003 should be at least a warning?

Edit2: editbin (build_commands.py:681) also requires vcvars, so the second option will be the way to go, without doubt.

MeFisto94:
Move the Execution of vcvars (which sets up the environment for visual studio tools) from mach.bat to python, so that ./mach works under mozilla-build and that #25300 can be used.

---
- [X] These changes fix #25360 #25336 
- [X] These changes do not require tests because changes to build infra

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -10,6 +10,7 @@
 from __future__ import print_function, unicode_literals
 
 import datetime
+import locale
 import os
 import os.path as path
 import platform
@@ -317,6 +318,16 @@ class MachCommands(CommandBase):
                 self.msvc_package_dir("gstreamer-uwp"), arch['gst_root'],
                 "lib", "pkgconfig"
             )
+
+        if 'windows' in host:
+            process = subprocess.Popen('("%s" %s > nul) && "python" -c "import os; print(repr(os.environ))"' %
+                                       (os.path.join(vs_dirs['vcdir'], "Auxiliary", "Build", "vcvarsall.bat"), "x64"),
+                                       stdout=subprocess.PIPE, shell=True)
+            stdout, _ = process.communicate()
+            exitcode = process.wait()
+            encoding = locale.getpreferredencoding()  # See https://stackoverflow.com/a/9228117
+            if exitcode == 0:
+                os.environ.update(eval(stdout.decode(encoding)))
 
         # Ensure that GStreamer libraries are accessible when linking.
         if 'windows' in target_triple:
--- a/mach.bat
+++ b/mach.bat
@@ -1,65 +1,8 @@
 @echo off
 
-setlocal
-
-if EXIST "%VCINSTALLDIR%" (
-  GOTO mach
-)
-
-pushd .
-
-IF EXIST "%ProgramFiles(x86)%" (
-  set "ProgramFiles32=%ProgramFiles(x86)%"
-) ELSE (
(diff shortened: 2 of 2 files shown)


MeFisto94:
With this being the most critical change, I encourage every windows user to just try this tooling and see what happens. We'll also see what the CI does.

In theory this only moves executing the bat to a later place and as such I can only imagine two possible failures:
 - We overwrite some environment variables (other recommendations as opposed to environ.update are welcome)
 - We do that too late and thus other code is affected (e.g. the Visual Studio Detection needed to be enhanced, as otherwise the env var already points to _some_ install of Visual Studio). It could happen that something breaks as the desired env var isn't set (this could happen silently, though, as in the case of #25360 where the presence of the SDK isn't enforced).

MeFisto94:
So this looks like at least part one (overwriting env vars) isn't really a concern. It could probably still behave odd on some subcommands, given that this is only added to build_commands, but that's maybe a good chance to look at those subcommands then and if required move the code to base_command only if really required.

nox:
r? @jdm 

I don't use Windows.

jdm:
Ah, this would have avoided the need for #25458 when ajeffrey encountered #25360.

jdm:
@bors-servo r+
Let's go ahead and merge this; that's the easiest way to shake out any latent issues.

jdm:
@bors-servo retry
Sorry for forgetting about this!

python/servo/build_commands.py
```diff
@@ -318,6 +319,16 @@ def build(self, release=False, dev=False, jobs=None, params=None, media_stack=No
...
 
+        if 'windows' in host:
+            process = subprocess.Popen('("%s" %s > nul) && "python" -c "import os; print(repr(os.environ))"' %
+                                       (os.path.join(vs_dirs['vcdir'], "Auxiliary", "Build", "vcvarsall.bat"), "x64"),
+                                       stdout=subprocess.PIPE, shell=True)
+            stdout, _ = process.communicate()
+            exitcode = process.wait()
+            encoding = locale.getpreferredencoding()  # See https://stackoverflow.com/a/9228117
+            if exitcode == 0:
+                os.environ.update(eval(stdout.decode(encoding)))
```
Manishearth, line 330: Does this actually do anything? All of the calls below use `env` with the call, so `os.environ` doesn't actually get used. Am I missing something?

I have my VS installation on `E:` and this causes the build to be unable to find msvcrt.lib (and other core libs)

PR #25367, opened 2019-12-23
apply is: to Document.createElement even when name isn't registered yet

### LINKED ISSUE #24997: 'Node.prototype.cloneNode(false) must be able to clone as a customized built-in element when it has an inconsistent "is" attribute' in WPT custom-elements/upgrading/Node-cloneNode.html
Here's the sequence of operations that fails, as I understand it:

A div instance3 is created (not yet added to DOM) with is=my-div3 when my-div3 isn't a registered name yet
The is= attribute of instance3 is updated to 'my-div2'.
Class MyDiv3 is registered as a custom element class for 'my-div3'.
The instance3 div is added to the DOM.
`instance3 instanceof MyDiv3` is unexpectedly false.

According to WPT behavior, which Firefox, Opera, and Chrome pass, the original is= from creation time should be applied and instance3 should be upgraded to MyDiv3 during this course of events; Safari also fails the test.

### LINKED ISSUE #24998: 'document.importNode() should import "undefined" custom elements successfully' in WPT custom-elements/upgrading/Document-importNode.html
This is the same kind of failure as #24997, but this time the element is never added to the DOM, and document.importNode is the point at which WPT expects it to be upgraded to the custom element class. As with #24997, Safari has its failure case in common with Servo. Unlike #24997, the is= attribute is never being modified here.

### LINKED ISSUE #25009: appendChild isn't upgrading a custom element
WPT custom-elements/Document-createElement.html creates an element with a 'my-div' is, then registers 'my-div', then appends the element, and the element has not become an instance of the registered class.

pshaughn:
The "is" option to Document.createElement should be respected even when the name hasn't been registered yet, in which case the name gets looked up again at the time the element should be upgraded. This change does that.
I'm now seeing a few test timeouts that aren't in the metadata, but I suspect they're slowness on my local configuration and not actual breakage.

---

- [X] These changes fix #25009 fix #24997 and fix #24998 

### DIFF
--- a/components/script/dom/create.rs
+++ b/components/script/dom/create.rs
@@ -189,14 +189,21 @@ fn create_html_element(
         }
     }
 
-    // Steps 7.1-7.2
+    // Steps 7.1-7.3
     let result = create_native_html_element(name.clone(), prefix, document, creator);
+    match is {
+        Some(is) => {
+            result.set_is(is);
+            result.set_custom_element_state(CustomElementState::Undefined);
+        },
+        None => {
+            if is_valid_custom_element_name(&*name.local) {
+                result.set_custom_element_state(CustomElementState::Undefined);
+            }
+        },
+    };
 
-    // Step 7.3
-    if is_valid_custom_element_name(&*name.local) || is.is_some() {
-        result.set_custom_element_state(CustomElementState::Undefined);
-    }
-
+    // Step 8
     result
 }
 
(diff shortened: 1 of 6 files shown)


jdm:
@bors-servo try=wpt
Good find!

CYBAI:
@bors-servo retry

- #24762

PR #25369, opened 2019-12-23
 Always pass InRealm to GlobalScope::from_context to avoid getting a null global

### LINKED ISSUE #25348: GlobalScope::from_context does not handle a null global object
https://github.com/servo/servo/blob/bac9903fbeed0a394a86c0091e727aada665433d/components/script/dom/globalscope.rs#L1297-L1302

It calls CurrentGlobalOrNull, but the return type is a `DomRoot<GlobalScope>`. In the short term adding an assert that the result is not null would be more safe than a segfault.

CYBAI:
---

- [x] These changes fix #25348

### DIFF
--- a/components/script/dom/bindings/error.rs
+++ b/components/script/dom/bindings/error.rs
@@ -14,6 +14,7 @@ use crate::dom::bindings::conversions::{
 use crate::dom::bindings::str::USVString;
 use crate::dom::domexception::{DOMErrorName, DOMException};
 use crate::dom::globalscope::GlobalScope;
+use crate::realms::InRealm;
 use crate::script_runtime::JSContext as SafeJSContext;
 #[cfg(feature = "js_backtrace")]
 use backtrace::Backtrace;
@@ -231,7 +232,7 @@ impl ErrorInfo {
 ///
 /// The `dispatch_event` argument is temporary and non-standard; passing false
 /// prevents dispatching the `error` event.
-pub unsafe fn report_pending_exception(cx: *mut JSContext, dispatch_event: bool) {
+pub unsafe fn report_pending_exception(cx: *mut JSContext, dispatch_event: bool, realm: InRealm) {
     if !JS_IsExceptionPending(cx) {
         return;
     }
@@ -285,7 +286,7 @@ pub unsafe fn report_pending_exception(cx: *mut JSContext, dispatch_event: bool)
     }
 
     if dispatch_event {
-        GlobalScope::from_context(cx).report_an_error(error_info, value.handle());
+        GlobalScope::from_context(cx, realm).report_an_error(error_info, value.handle());
     }
 }
 
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -38,7 +38,7 @@ use crate::dom::window::Window;
 use crate::dom::workerglobalscope::WorkerGlobalScope;
 use crate::dom::workletglobalscope::WorkletGlobalScope;
 use crate::microtask::{Microtask, MicrotaskQueue, UserMicrotask};
-use crate::realms::enter_realm;
+use crate::realms::{enter_realm, InRealm};
 use crate::script_module::ModuleTree;
 use crate::script_runtime::{CommonScriptMsg, JSContext as SafeJSContext, ScriptChan, ScriptPort};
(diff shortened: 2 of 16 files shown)


CYBAI:
@jdm I tried to make it return an option though, does this make sense to you? Or, we shouldn't get a null global in that case so should just use assert instead 👀?

CYBAI:
@bors-servo try=wpt

- Should not break anything 👀

components/script/dom/globalscope.rs
```diff
@@ -1296,9 +1296,13 @@ impl GlobalScope {
 
     /// Returns the global scope for the given JSContext
     #[allow(unsafe_code)]
-    pub unsafe fn from_context(cx: *mut JSContext) -> DomRoot<Self> {
+    pub unsafe fn from_context(cx: *mut JSContext) -> Option<DomRoot<Self>> {
```
jdm: Thinking about this further, I think we can do something more useful. `CurrentGlobalOrNull` uses the global from the current realm. This means that if we add an `InCompartment` argument to this method, we can force callers to provide proof that they have already entered a realm. Then we can remove the Option from the return value and assert that `global` is never null.

jdm (changes requested):
Thanks for making this pull request! It helped me to see all of the callers of this code and clarify how it could be improved.

components/script/dom/globalscope.rs
CYBAI: Thanks! Good point and understood! Will update!

components/script/dom/windowproxy.rs
```diff
@@ -386,7 +386,14 @@ impl WindowProxy {
                     opener_id,
                 ) {
                     Some(opener_top_id) => {
-                        let global_to_clone_from = unsafe { GlobalScope::from_context(cx) };
+                        let global_to_clone_from = unsafe {
+                            let in_compartment_proof =
+                                AlreadyInCompartment::assert_for_cx(SafeJSContext::from_ptr(cx));
```
jdm: Instead, let's add an `InCompartment` argument to `opener`, and since it's only called by Window::Opener we can modify https://github.com/servo/servo/blob/12693b51f510761d5dd58a9103596a192edb803c/components/script/dom/bindings/codegen/Bindings.conf#L60 to include that method.

components/script/dom/windowproxy.rs
```diff
@@ -983,8 +990,10 @@ pub fn new_window_proxy_handler() -> WindowProxyHandler {
 #[allow(unsafe_code)]
 unsafe fn throw_security_error(cx: *mut JSContext) -> bool {
     if !JS_IsExceptionPending(cx) {
-        let global = GlobalScope::from_context(cx);
-        throw_dom_exception(SafeJSContext::from_ptr(cx), &*global, Error::Security);
+        let safe_context = SafeJSContext::from_ptr(cx);
+        let in_compartment_proof = AlreadyInCompartment::assert_for_cx(safe_context);
```
jdm: Instead, let's add an `InCompartment` argument to `throw_security_error` and make the callers prove that they're in a realm.

components/script/dom/bindings/error.rs
```diff
@@ -285,7 +286,9 @@ pub unsafe fn report_pending_exception(cx: *mut JSContext, dispatch_event: bool)
     }
 
     if dispatch_event {
-        GlobalScope::from_context(cx).report_an_error(error_info, value.handle());
+        let in_compartment_proof = AlreadyInCompartment::assert_for_cx(SafeJSContext::from_ptr(cx));
```
jdm: Let's add an `InCompartment` argument to `report_pending_exception` instead.

jdm (changes requested):

CYBAI:
@bors-servo try=wpt

- Should not break anything 👀

jdm (approved):
This looks great! Please rebase and we'll merge :)

jdm:
```
error: unused import: `JSAutoRealm`
  --> components/script/dom/globalscope.rs:67:17
   |
67 | use js::jsapi::{JSAutoRealm, JSContext};
   |                 ^^^^^^^^^^^
   |
```

jdm:
@nors-servo r+

jdm:
@bors-servo retry
* network

jdm:
@bors-servo retry
* #24726

PR #25372, opened 2019-12-23
Implement OS Version Detection for the User Agent

### LINKED ISSUE #18194: Servo user agent should list user's actual operating system

MeFisto94:
User Agent now correctly reports Windows 7, 8.1, 10, ...

---
- [X] These changes fix #18194
- [X] These changes do not require tests, because that would involve testing servo across all windows versions, to be meaningful

### DIFF
--- a/components/config/opts.rs
+++ b/components/config/opts.rs
@@ -478,27 +478,48 @@ enum UserAgent {
 }
 
 fn default_user_agent_string(agent: UserAgent) -> &'static str {
-    #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
-    const DESKTOP_UA_STRING: &'static str =
-        "Mozilla/5.0 (X11; Linux x86_64; rv:63.0) Servo/1.0 Firefox/63.0";
-    #[cfg(all(target_os = "linux", not(target_arch = "x86_64")))]
-    const DESKTOP_UA_STRING: &'static str =
-        "Mozilla/5.0 (X11; Linux i686; rv:63.0) Servo/1.0 Firefox/63.0";
-
-    #[cfg(all(target_os = "windows", target_arch = "x86_64"))]
-    const DESKTOP_UA_STRING: &'static str =
-        "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:63.0) Servo/1.0 Firefox/63.0";
-    #[cfg(all(target_os = "windows", not(target_arch = "x86_64")))]
-    const DESKTOP_UA_STRING: &'static str =
-        "Mozilla/5.0 (Windows NT 10.0; rv:63.0) Servo/1.0 Firefox/63.0";
-
-    #[cfg(not(any(target_os = "linux", target_os = "windows")))]
-    // Neither Linux nor Windows, so maybe OS X, and if not then OS X is an okay fallback.
-    const DESKTOP_UA_STRING: &'static str =
-        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:63.0) Servo/1.0 Firefox/63.0";
-
     match agent {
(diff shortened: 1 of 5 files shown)


MeFisto94:
In order to test this changes, surf to `"https://www.whatsmyua.info` or `"https://www.whatismybrowser.com/detect/what-is-my-user-agent` (the latter shows another issue: analyseing the UA fails due to some CSRF problem).

I did only test this on Windows 10, but the crate is fairly simple and as such I don't expect any issues, apart from the following two:
- The crate still uses 0.2.8 winapi, which makes test-tidy fail:
```
./Cargo.lock:1: duplicate versions for package `winapi`
The following packages depend on version 0.2.8 from 'crates.io':
The following packages depend on version 0.3.8 from 'https://github.com/servo/winapi-rs?branch=patch-1':
```
See https://github.com/DarkEld3r/os_info/issues/111 for tracking

- The crate doesn't work on UWP, which doesn't really have a concept of a Windows Version either (https://github.com/DarkEld3r/os_info/issues/112), thus it has just been excluded in this PR.

paulrouget:
winapi duplicate versions is annoying. Do we have to use os_info? How does `os_info::get().version().version()` work?

MeFisto94:
Someone already created an issue for winapi 0.30 on os_info. We could also submit a PR.
The code itself is really [small](https://github.com/DarkEld3r/os_info/blob/master/src/windows/winapi.rs), though and we probably don't need it for Linux and most of the other supported OS anyway. Only for Mac OS X maybe.

atouchet:
The latest version of os_info is now using winapi 0.3.

components/config/opts.rs
```diff
@@ -21,6 +21,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{RwLock, RwLockReadGuard};
 use url::{self, Url};
 
+#[cfg(all(target_os = "windows", not(target_vendor = "uwp")))]
+extern crate os_info;
```
jdm: This statement should not be necessary in the 2018 edition of Rust, which components/config/Cargo.toml enables.

jdm (approved):
This seems fine to me. Please update to the latest version of os_info and we can merge this.

components/config/opts.rs
```diff
@@ -478,27 +478,48 @@ enum UserAgent {
...
-
     match agent {
-        UserAgent::Desktop => DESKTOP_UA_STRING,
+        UserAgent::Desktop => {
+            if cfg!(all(target_os = "windows", not(target_vendor = "uwp"))) &&
+                !pref!(network.user_agent.spoof)
+            {
+                match os_info::get().version().version() {
+                    os_info::VersionType::Semantic(maj, min, _patch) => {
+                        return Box::leak(
```
jdm: I just noticed that this will leak memory on every network request.

jdm (changes requested):
Thinking about this a little harder, and how to avoid the leak and compilation error on non-windows, I'm becoming more hesitant about merging this. It only affects desktop windows builds, and only impacts windows versions that are already end-od-life. I am incluned to stick with the existing implementation and avoid the complexity without a compelling motivation otherwise.

components/config/opts.rs
MeFisto94: While it doesn't change much about the reasoning, I _think_ or better put _assumed_ that this would happen only once on startup in https://github.com/MeFisto94/servo/blob/a4471368965d30c08042b73b5d475558a2e277f0/components/config/opts.rs#L567

jdm: Ah, right. My mistake.

PR #25373, opened 2019-12-23
let document.createElement[NS] accept a string for options

### LINKED ISSUE #25008: WPT asserts "document.createElementNS()'s third argument is to be ignored when it's a string", likewise createElement()'s second
This is in /custom-elements/Document-createElementNS.html. Opera, Chrome, and Safari fail, though possibly not in the same way; Servo's throwing an exception.

pshaughn:
The string actually does nothing, but spec and WPT don't want it to do anything. https://dom.spec.whatwg.org/#dom-document-createelement only cares about the options value when it's a dictionary, and the WPT test on the string case is just that it isn't throwing an exception.

---

- [X] These changes fix #25008


jdm:
@bors-servo retry
* #24726

Issue #14897, opened 2017-01-06
Integrate websocket connection creation into fetch network stack


jdm:
Right now, the DOM implementation of WebSocket sends a special message to the resource thread asking for a websocket connection to a particular URL. #14895 allows us to model the code more like the [specification](https://fetch.spec.whatwg.org/#concept-http-network-fetch) instead, and make the DOM code look like other code that initiates network connections. To start with, let's remove the special case (`WebsocketConnection`) from `CoreResourceMsg`, and make the resource thread check the request's mode and use the websocket loader if applicable (instead of the regular fetch code).

@dowoncha Would you like to do this?

dowoncha:
Yeah, I will take this. If the websocketconnection is removed from CoreResourceMsg, where would I get the WebSocketCommunicate, and WebSocketConnectData to pass to websocket_connect? Also would I be adding this check inside the thread builder in CoreResourceManager::fetch?

jdm:
Those are good questions! I propose we add the `WebSocketCommunicate` and `protocols` (from `WebSocketConnectData`) as fields of the `WebSocket` request mode; the other data should already be present in the `RequestInit` structure that we send with the Fetch message. This check can live outside of the thread builder, since websocket_loader already creates its own thread.

dowoncha:
Attempting to add fields to the WebSocket enum variant throws an error as WebSocket does not implement Copy. How can I resolve this?

jdm:
`WebSocketCommunicate` cannot be `Copy`, but it should be ok to remove the `Copy` derive from `RequestMode` and use `.clone()` everywhere that's necessary.

dowoncha:
Sorry for the constant questions, removing `Copy` derive still requires `WebSocketCommunicate` to derive
`Clone`, `HeapSizeOf`, and `PartialEq`.

jdm:
That's fine!

jdm:
More specifically, if `HeapSizeOf` causes problems you can use `#[ignore_heap_size_of` (look for other uses in the code), and if `PartialEq` causes problems you can remove `PartialEq` from `RequestMode` and modify the code that relies on it to use `match` or `if let` instead.

dowoncha:
[...]

PR #25404, opened 2019-12-30
Fix HSTS

### LINKED ISSUE #14363: Implement HSTS

### LINKED ISSUE #20120: Servo should only support HTTPS connections, zero without "S"
Hello, 

first, this looks like a great future project, next-gen webbrowser, wow. 

Second, I think it should only support HTTPS connections and not plain HTTP.

If it points to the future. 

only problem: OCSP uses plain HTTP: https://moxie.org/papers/ocsp-attack.pdf

Darkspirit:
The headers crate does not [expose](https://github.com/hyperium/headers/blob/0c42ad8cf56f9af28973c3da71616fa478781fdf/src/common/strict_transport_security.rs#L42) HSTS struct fields. At the moment, it's only usable for HSTS header encoding. An update of the headers crate would require a huge update of http, hyper, hyper_serde, net::decoder as well. Therefore I've copied the `typed_get::<StrictTransportSecurity>` decoding feature  for now, but with exposed struct fields. Let's remove this custom struct with the next hyper upgrade. I tried to prevent needless HSTS database lookups when network.enforce_tls.enabled is set.

---
- [x] These changes fix #14363, fix #20120.


components/net/hsts.rs
```diff
@@ -139,15 +142,160 @@ impl HstsList {
     }
 
     /// Step 10 of https://fetch.spec.whatwg.org/#concept-main-fetch.
```
asajeffrey: Nit: is this the right spec link?

components/net/hsts.rs
```diff
@@ -139,15 +142,160 @@ impl HstsList {
...
-    pub fn switch_known_hsts_host_domain_url_to_https(&self, url: &mut ServoUrl) {
-        if url.scheme() != "http" {
+    pub fn apply_hsts_rules(&self, url: &mut ServoUrl) {
+        if url.scheme() != "http" && url.scheme() != "ws" {
```
asajeffrey: Is there a spec link for rewriting ws as wss?

components/net/hsts.rs
```diff
@@ -139,15 +142,160 @@ impl HstsList {
...
             return;
         }
-        if url
-            .domain()
-            .map_or(false, |domain| self.is_host_secure(domain))
+
+        let mut upgrade_scheme = false;
```
asajeffrey: Nit: can we make this not `mut` by assigning a boolean expression?

components/net/hsts.rs
```diff
@@ -139,15 +142,160 @@ impl HstsList {
...
+                    }
+
+                    self.push(entry);
+                }
+            }
+        }
+    }
+}
+
+// TODO: Remove this when the headers crate exposes STS struct fields
```
asajeffrey: File an issue for this in the headers repo, and link to it here?

asajeffrey (approved):
Some nits but otherwise LGTM.

components/net/hsts.rs
Darkspirit: I did not touch this line, but it should be corrected to "Step 2.9".
> Set request’s current URL’s scheme to "https" if all of the following conditions are true:
> * request’s current URL’s scheme is "http"
> * request’s current URL’s host is a domain
>  * Matching request’s current URL’s host per Known HSTS Host Domain Name Matching results in either a superdomain match with an asserted includeSubDomains directive or a congruent match (with or without an asserted includeSubDomains directive). [HSTS]

components/net/hsts.rs
Darkspirit: https://fetch.spec.whatwg.org/#websocket-opening-handshake
> 1. Let requestURL be a copy of url, with its scheme set to "http", if url’s scheme is "ws", and to "https" otherwise.
> 
> This change of scheme is essential to integrate well with fetching. E.g., HSTS would not work without it. There is no real reason for WebSocket to have distinct schemes, it’s a legacy artefact. [HSTS]

The spec recommends to work with a http(s) url, but ws-rs expects to get a ws(s) url. I did not change this aspect.
https://github.com/servo/servo/pull/25404/commits/6dc1f49c6eee059e6bfb8bcbbc53239a7d6f18c0#diff-68b5c651915d8b239e5e3b6c84f02981L194-R252
net_url contains the `ws(s)://` url as it is needed for ws-rs. Instead of converting the upgraded url back to wss in websocket_loader.rs, I kept it clean and upgraded ws to wss.
#14897 seems to be about changing the websocket code to be tighter along the spec.

components/net/hsts.rs
Darkspirit: Should I change the other occurrences of `upgrade_scheme` to `let upgrade_scheme` instead?

components/net/hsts.rs
Darkspirit: https://github.com/hyperium/headers/issues/61

components/net/hsts.rs
asajeffrey: Can you add that link (and the discussion about ws urls) to your PR?

components/net/hsts.rs
asajeffrey: No, just initialize `upgrade_scheme` to be the "right" value, rather than initializing it to be false then updating it.

components/net/hsts.rs
asajeffrey: Thanks!

components/net/hsts.rs
Darkspirit: Added here: https://github.com/servo/servo/pull/25404/files#diff-68b5c651915d8b239e5e3b6c84f02981R202

components/net/hsts.rs
asajeffrey: Can you add a link to the hyper issue?

components/net/http_loader.rs
```diff
@@ -1564,9 +1564,10 @@ fn http_network_fetch(
 
     // Substep 2
 
-    // TODO Determine if response was retrieved over HTTPS
-    // TODO Servo needs to decide what ciphers are to be treated as "deprecated"
```
Darkspirit: (I removed this comment because deprecated ciphers have been removed from Servo to align with the deprecation process of other browsers.)

components/net/hsts.rs
Darkspirit: Done, and I will create a PR to add getter methods.

components/net/http_loader.rs
asajeffrey: 👍

asajeffrey:
OK, thanks for dealing with all my niggles! @bors-servo r+

jdm:
@bors-servo retry
* #25459

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* our old friend ipc-channel on linux

PR #25408, opened 2019-12-31
Now using eq_ignore_case on WASM mimetype, per #25317

### LINKED ISSUE #25317: Update async wasm mime type check to use case-insensitive comparison
https://github.com/servo/servo/blob/a5844d5efd9135c0448ace252508bdce20e341c3/components/script/script_runtime.rs#L917-L919

pshaughn:
I have no idea how this would be tested.

---

- [X] These changes fix #25317


jdm:
@bors-servo r+
Thanks!

PR #25410, opened 2020-01-01
Implement CustomElementRegistry.upgrade

### LINKED ISSUE #24989: Implement CustomElementRegistry.upgrade
WPT test custom-element-registry/upgrade fails on this

pshaughn:
An additional small change will be needed if whatwg/html#5126 lands, which will pass an infinite recursion test this is currently failing in custom-elements/upgrading.html (right now the test is asking for behavior that isn't in the spec).
The remaining test in custom-elements/custom-element-registry/upgrade.html is because it needs shadow DOM.

---

- [X] These changes fix #24989

### DIFF
--- a/components/script/dom/customelementregistry.rs
+++ b/components/script/dom/customelementregistry.rs
@@ -448,6 +448,17 @@ impl CustomElementRegistryMethods for CustomElementRegistry {
         // Step 6
         promise
     }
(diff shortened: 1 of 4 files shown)


jdm:
@bors-servo r+
Thanks for implementing this!

PR #25418, opened 2020-01-02
Fix some warnings in future Rust nightlies

SimonSapin:


SimonSapin:
Ah, those unnecessary `extern crate proc_macro` that cause warnings in a recent Nightly only because unnecessary more recently than our current toolchain. And actually upgrading is still blocked on https://github.com/rust-lang/rust/issues/67601. I’ve reverted that part.

PR #22264, opened 2018-11-24
[WIP] Implement live NodeList

KiChjang:
Creating this PR just to see what WPT says for now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/22264)

PR #25424, opened 2020-01-03
Make labelable element .labels a live list in tree order

### LINKED ISSUE #25391: Many out-of-spec behaviors associating input elements with labels
html/semantics/forms/the-label-element/label-attributes.sub.html tests various conditions about labels and gets several different failures; for example, moving an input element in or out of a label's descendants appears not to cause it to associate or disassociate with the label.

One failure is just that we don't have Shadow DOM but the rest look like they're legitimately about things labels should do.

pshaughn:
This is not the highest-performance solution possible but it's visibly spec-aligned in a way a faster-performing implementation would be harder to verify, and I don't expect label-getting to deal with more than a few nodes at once in practice.
I added a macro by analogy with some of the existing make_XXX_getter! macros; I will change it if it doesn't seem right.
Remaining test failures are because keygen, shadow DOM, and ElementInternals are unimplemented. Shadow DOM should already be handled by the existing code when it is implemented, and keygen should just be able to add a labels_node_list and use the macro like the other labelable elements. ElementInternals labels are slightly different and might need another NodeList case.

---

- [X] These changes fix #25391

- [X] There are tests for these changes, although there's room for more (see https://github.com/web-platform-tests/wpt/issues/21028)

### DIFF
--- a/components/script/dom/nodelist.rs
+++ b/components/script/dom/nodelist.rs
@@ -7,6 +7,7 @@ use crate::dom::bindings::codegen::Bindings::NodeListBinding;
 use crate::dom::bindings::codegen::Bindings::NodeListBinding::NodeListMethods;
 use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
 use crate::dom::bindings::root::{Dom, DomRoot, MutNullableDom};
+use crate::dom::htmlelement::HTMLElement;
 use crate::dom::node::{ChildrenMutation, Node};
 use crate::dom::window::Window;
 use dom_struct::dom_struct;
@@ -17,6 +18,7 @@ use std::cell::Cell;
 pub enum NodeListType {
     Simple(Vec<Dom<Node>>),
     Children(ChildrenList),
+    Labels(LabelsList),
 }
 
 // https://dom.spec.whatwg.org/#interface-nodelist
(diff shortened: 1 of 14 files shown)


jdm:
The only failures are three separate instances of #24726.

components/script/dom/nodelist.rs
```diff
@@ -319,3 +327,39 @@ impl ChildrenList {
...
+        }
+    }
+
+    pub fn len(&self) -> u32 {
+        self.element.labels_vec().len() as u32
+    }
+
+    pub fn item(&self, index: u32) -> Option<DomRoot<Node>> {
+        let index = index as usize;
+        let vec = self.element.labels_vec();
```
Manishearth: If the only time we use `labels_vec()` is to immediately index it, we should avoid spending the extra cost of iteration and allocation and have a `label_at(i)` function instead

Manishearth (approved):
Seems like Firefox uses a mutation observer to make the live list work efficiently. We do support them (I think), so this should be possible, but perhaps not necessary in this first pass.

components/script/dom/nodelist.rs
pshaughn: > If the only time we use `labels_vec()` is to immediately index it, we should avoid spending the extra cost of iteration and allocation and have a `label_at(i)` function instead

Yes. I realized that when I was working on an analogous case in #25428 and I should do that, assuming we are going this road and not the mutation observer one.

> Seems like Firefox uses a mutation observer to make the live list work efficiently. We do support them (I think), so this should be possible, but perhaps not necessary in this first pass.

#22264 is a separate attempt by @KiChjang at doing it that way, of which I was unaware when I worked on this. I don't see how that approach can be made to work in the case where the element and its label are in a detached subtree rather than part of a document, but I probably just don't know enough about mutation observers.

Manishearth: Yeah we should land this with label_at, and then worry about mutation observers. Worth looking closer at Firefox's implementation to see what it does.

pshaughn: Updated to separate label_at, labels_count methods.

pshaughn:
(sorry about the double-force-push, I forgot test-tidy the first time and I had tabs)

jdm:
@bors-servo retry
* #24726 #25443

PR #25428, opened 2020-01-03
make Document.getElementsByName a live collection

### LINKED ISSUE #25147: document.getElementsByName seems to be a static collection that doesn't reflect a subsequent appendChild
WPT html/dom/documents/dom-tree-accessors/document.getElementsByName/document.getElementsByName-liveness.html tests this. I also tried changing the appended child from an embed to an input in case it was a problem with embed getting, with no change.

pshaughn:
Another new case for NodeList; this and the labels live collection in #25424 are in the pipeline simultaneously, so one of them will need a merge resolution when the other one lands.

Iterating over many same-named elements is potentially slower than it has to be, since getting the nth element from the live view always starts from the start of the tree.

---

- [X] These changes fix #25147

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -3056,14 +3056,56 @@ impl Document {
         self.redirect_count.set(count)
     }
 
-    fn create_node_list<F: Fn(&Node) -> bool>(&self, callback: F) -> DomRoot<NodeList> {
+    pub fn elements_by_name_count(&self, name: &DOMString) -> u32 {
+        if name.is_empty() {
+            return 0;
+        }
+        self.count_node_list(|n| Document::is_element_in_get_by_name(n, name))
+    }
+
+    pub fn nth_element_by_name(&self, index: u32, name: &DOMString) -> Option<DomRoot<Node>> {
+        if name.is_empty() {
+            return None;
+        }
+        self.nth_in_node_list(index, |n| Document::is_element_in_get_by_name(n, name))
+    }
+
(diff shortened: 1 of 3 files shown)


asajeffrey:
`ChildrenList` uses a cursor to avoid the problem of a niave iteration over the list taking `O(n^2)` time rather than `O(n)`. Can we do the same for `ElementByNameList`?

asajeffrey:
Also, https://github.com/servo/servo/pull/22264#issuecomment-571164865 may apply to this PR too?

pshaughn:
@asajeffrey: Yes to both questions, and indeed #22264 was an attempt to do it that way. I think #22264 might have been successful for the .getElementsByName subcase and only ran into trouble in the .labels case, so I could look into using that work here.

pshaughn:
I don't think I understand everything #22264 was doing with mutation observers. I understand the part within document.rs where it's using a mutation observer message to know when to mark the list stale, but I don't follow the changes to mutationobserver.rs. I might just need a fresher look in the morning though.

jdm:
@pshaughn What's the status here? Is this still waiting on a review?

pshaughn:
I don't know how to make it fast. If slow and spec-compliant is better than fast and not-spec-compliant, then this is ready to review; if not then this should just be closed.

jdm:
I am inclined to start with slow and correct, since that gets us closer to web compatibility than no implementation at all.

jdm:
@bors-servo retry
* network, #24786

PR #25393, opened 2019-12-26
Implement HTMLOptionElement named constructor

teapotd:
This change implements `Option` named constructor for `HTMLOptionElement`.

---
- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #25380
- [X] There are tests for these changes

PR #25431, opened 2020-01-04
use create_html_element for HTMLAudioElement and HTMLImageElement

### LINKED ISSUE #25421: Use create_html_element in Audio and Image constructors
https://github.com/servo/servo/pull/25393#discussion_r362679594

warren-fisher:
Updated the Image and Audio constructors to use `create_html_element` via the Element::create method. This was done to meet these specifications of "Let (audio/image) be the result of **creating an element** given document, audio, and the HTML namespace." for [dom-image](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-image) and [dom-audio](https://html.spec.whatwg.org/multipage/media.html#dom-audio)
Not sure what _is_ is according to the [create-element guidelines](https://dom.spec.whatwg.org/#concept-create-element) so I left it as None copying from #25393. Also copied the ElementCreator and CustomElementCreationMode from #25393 as I do not know what they do.

---

- [x] These changes fix #25421 (GitHub issue number if applicable)

- [x] These changes do not require tests because it is a small swap out of the way used to generate these HTML elements. The pre-existing tests should be sufficient.

### DIFF
--- a/components/script/dom/htmlaudioelement.rs
+++ b/components/script/dom/htmlaudioelement.rs
@@ -10,12 +10,12 @@ use crate::dom::bindings::inheritance::Castable;
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::DOMString;
 use crate::dom::document::Document;
-use crate::dom::element::Element;
+use crate::dom::element::{CustomElementCreationMode, Element, ElementCreator};
 use crate::dom::htmlmediaelement::HTMLMediaElement;
 use crate::dom::node::Node;
 use crate::dom::window::Window;
 use dom_struct::dom_struct;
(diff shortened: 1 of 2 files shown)


nox:
@jdm Why does it matter whether `create_html_element` is used or not? AFAIK this will never go through the custom element code path (because the name "audio" cannot possibly correspond to a custom element), and this leads to more unwraps in the code.

jdm:
I'm not confident enough in my understanding of the spec to say definitively if that's true. My preference is for the code to clearly match the specification when that's the case, for future-proofing reasons.

jdm:
@bors-servo retry
* #25522 #22212

PR #25434, opened 2020-01-05
Put correct version number in rustup version-requirement string

pshaughn:
Just fixing a little typo that confused me when I saw it come up.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -1033,13 +1033,14 @@ install them, let us know by filing a bug!")
                 print("It looks like rustup is not installed. See instructions at "
                       "https://github.com/servo/servo/#setting-up-your-environment")
                 print()
-                return 1
+                sys.exit(1)
             raise
         version = tuple(map(int, re.match(b"rustup (\d+)\.(\d+)\.(\d+)", version_line).groups()))
-        if version < (1, 21, 0):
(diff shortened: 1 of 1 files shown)


python/servo/command_base.py
```diff
@@ -1037,7 +1037,7 @@ def ensure_rustup_version(self):
             raise
         version = tuple(map(int, re.match(b"rustup (\d+)\.(\d+)\.(\d+)", version_line).groups()))
         if version < (1, 21, 0):
-            print("rustup is at version %s.%s.%s, Servo requires 1.11.0 or more recent." % version)
+            print("rustup is at version %s.%s.%s, Servo requires 1.21.0 or more recent." % version)
```
jdm: Let's make this a variable that's used in the check as well to avoid this happening again in the future.

jdm (changes requested):

SimonSapin:
Ouch, I specifically thought of not making this mistake and then went on to do it. Sorry!

Also, `return 1` in this position is not effective anymore. (With this section of code having moved into another method.) Should this check failing call `sys.exit` or raise an exception instead?

jdm:
```
./python/servo/command_base.py:1039: E703 statement ends with a semicolon
```

pshaughn:
Sorry about that!

Issue #25429, opened 2020-01-04
"If an element listed in a form element's past names map changes form owner, then its entries must be removed from that map."


pshaughn:
This requirement is placed in an easily missable location after the steps of https://html.spec.whatwg.org/multipage/forms.html#the-form-element:radionodelist-4 and doesn't even seem to have its own anchor to direct-link to. I think it can be satisfied by having each form-associated element's set_form_owner notify the old owner that it's leaving before changing to a new owner.

https://github.com/servo/servo/blob/2488213d18d5bc9806b41fb2756a2a82dbe933a4/components/script/dom/htmlinputelement.rs#L1688-L1690
This would first check if form_owner.get is_some and is different from form, and if so call something like form_owner.get().unwrap().remove_from_past_names_map(self), and likewise for every other element class with impl FormControls.

Another idea I had would be to just check, when getting from the map, if it's still owned, but that would be observably different in a case where it stopped being owned and then started being owned again.

jdm:
Good catch!

pshaughn:
The last test in html/semantics/forms/the-form-element/form-nameditem.html covers this thoroughly, associating, renaming, and disassociating elements in many ways; the first assert we fail on is when it uses .remove() on a renamed element and expects that to remove the past names map entry for its previous name.

Issue #25036, opened 2019-12-03
"Legacy platform object" getOwnPropertyDescriptor returns undefined when it should be finding dynamic properties


pshaughn:
WebIDL/ecmascript-binding/legacy-platform-object.html tests out various things "legacy platform objects" (https://heycam.github.io/webidl/#es-legacy-platform-objects) are expected to do, most of which involving pretending to be arrays or open-ended hashmaps in one way or another. Mutliple tests fail on getOwnPropertyDescriptor, which is not seeing the dynamic properties.

HTMLSelectElement insufficiently pretending to be an array is also in https://github.com/servo/servo/issues/25003; there is possibly overlap, if the reason getOwnPropertyDescriptor isn't working is that the property isn't even there to get.

Other failures are DOMTokenList with numeric properties and DOMStringMap with named ones.

jdm:
Relevant to getOwnPropertyDescriptor for legacy objects: https://github.com/servo/servo/blob/54d88fd042d3d1d382b36bf5d53308dabc35991d/components/script/dom/bindings/codegen/CodegenRust.py#L5130-L5228

pshaughn:
I don't understand that code very well. In the named getter case, where is it actually looking up the name?

pshaughn:
https://github.com/servo/servo/blob/5c7a4db5f4e374dee287d403d90c7bc0e07ac9d0/tests/wpt/web-platform-tests/html/semantics/forms/the-form-element/form-nameditem.html#L112-L118 this test seems to be getting back a descriptor that exists, but that has .value=undefined. Could that be related to the FIXME(#11868) not setting the real value?

jdm:
That seems entirely plausible, yes.

pshaughn:
Removing the .get() gets us three asserts further down in this particular test (failing on desc.enumerable); if it doesn't break anything else, it's an improvement.

jdm:
What's the enumerable failure? The descriptor returned should be unconditionally enumerable based on the [this line](https://github.com/servo/servo/blob/54d88fd042d3d1d382b36bf5d53308dabc35991d/components/script/dom/bindings/codegen/CodegenRust.py#L5147).

pshaughn:
The test wants it to not be enumerable: https://github.com/servo/servo/blob/5c7a4db5f4e374dee287d403d90c7bc0e07ac9d0/tests/wpt/web-platform-tests/html/semantics/forms/the-form-element/form-nameditem.html#L121

[...]

Issue #21495, opened 2018-08-23
Shall we delete the comment `fixme(#3553)`?


tigercosmos:
https://github.com/servo/servo/blob/a1abdde8c62d01a144d12ae4b425d6b062bad41b/components/script/dom/htmlformelement.rs#L495-L498

#3553 is closed.
There is a merged PR in #15938.

tigercosmos:
cc @KiChjang

jdm:
I don't think that code has been updated to take advantage of the changes in #3553 yet.

tigercosmos:
I think we usually fill the issue number, which is opened for the description, into `FIXME(#<number>)`.

In this case, maybe modify this comment to avoid misleading is better.
And since the PR is already merged, it could also be: 
```
// FIXME: This is an incorrect way of getting controls owned by the 
//             form, should refactor this based on html5ever's form owner(#3553)
```

KiChjang:
I think the better solution here is to fix the underlying issue, rather than to fix the wording of the comments. @tigercosmos Do you want to make a PR for the changes?

tigercosmos:
I can have a try. I don't look much about this code. @KiChjang Do you remember why you said it is incorrect, and what do you expect it would be like?

pshaughn:
At least one thing is testably wrong relating to this. html/semantics/forms/the-form-element/form-nameditem.html expects an earlier `<input form=a name=b>` to be picked up when a `<form id=a></form>` appearing later in the page is going through its controls, and it doesn't find it.

pshaughn:
There's a FIXME(#3553) that probably really means this issue.

Issue #12978, opened 2016-08-22
Support OverrideBuiltins WebIDL annotation


jdm:
Specification: http://heycam.github.io/webidl/#OverrideBuiltins
Used in: HTMLFormElement.webidl, Document.webidl, DOMStringMap.webidl

Filing to keep track of it; I don't think it's a high priority to implement this.

pshaughn:
html/semantics/forms/the-form-element/form-nameditem.html tests this, just in case someone wants to name a radio group "addEventListener" or "ownerDocument" and then used the named getter on it.

jdm:
```html
<!doctype html>
<p id="X" data-foobar="hello" data-to-string="hello" data-constructor="hello"></p>
<script>
console.log(X.dataset.foobar, X.dataset.toString, X.dataset.constructor)
</script>
```
From #46388

jdm:
Look in https://searchfox.org/firefox-main/source/dom/bindings/Codegen.py for uses of LegacyOverrideBuiltIns; we need to copy it.

PR #25435, opened 2020-01-05
RadioNodeList liveness

### LINKED ISSUE #25415: RadioNodeList should be a live view
The concept of a live NodeList appears in several places, much like the concept of a live HTMLCollection. Currently, the only kind of live NodeList we have is for a ParentNode's children. Three other cases need a live one: the labels of a labelable element (issue #25391), the result of Document.getElementsByName (issue #25147), and RadioNodeList. [querySelectorAll and MutationRecord have NodeLists but are not meant to be live]

https://html.spec.whatwg.org/multipage/forms.html#dom-form-nameditem step 1 and https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#dom-htmlformcontrolscollection-nameditem step 4 are the two places where a RadioNodeList can be created; both specify that it's live, so RadioNodeList should always be live.

The two specifications for what the RadioNodeList exposes do not seem identical at first glance, and the implementation might need two different states to cover them both:
"all the listed [form-associated] elements, whose form owner is the form element, that have either an id attribute or a name attribute equal to name, with the exception of input elements whose type attribute is in the Image Button state, in tree order."
" a live view of the HTMLFormControlsCollection object, further filtered so that the only nodes in the RadioNodeList object are those that have either an id attribute or a name attribute equal to name. The nodes in the RadioNodeList object must be sorted in tree order."

pshaughn:
When the named getter of a form's .elements gets a list, that list is now a live view into the form, passing a WPT test that removes a node and then looks at the list again.

---

- [X] These changes fix #25415 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -13,6 +13,7 @@ use crate::dom::bindings::codegen::Bindings::HTMLFormElementBinding;
 use crate::dom::bindings::codegen::Bindings::HTMLFormElementBinding::HTMLFormElementMethods;
 use crate::dom::bindings::codegen::Bindings::HTMLInputElementBinding::HTMLInputElementMethods;
 use crate::dom::bindings::codegen::Bindings::HTMLTextAreaElementBinding::HTMLTextAreaElementMethods;
+use crate::dom::bindings::codegen::Bindings::NodeListBinding::NodeListMethods;
 use crate::dom::bindings::codegen::Bindings::WindowBinding::WindowBinding::WindowMethods;
 use crate::dom::bindings::inheritance::{Castable, ElementTypeId, HTMLElementTypeId, NodeTypeId};
 use crate::dom::bindings::refcounted::Trusted;
@@ -45,6 +46,8 @@ use crate::dom::htmltextareaelement::HTMLTextAreaElement;
 use crate::dom::node::{document_from_node, window_from_node};
 use crate::dom::node::{Node, NodeFlags, ShadowIncluding};
 use crate::dom::node::{UnbindContext, VecPreOrderInsertionHelper};
+use crate::dom::nodelist::{NodeList, RadioListMode};
+use crate::dom::radionodelist::RadioNodeList;
 use crate::dom::validitystate::ValidationFlags;
 use crate::dom::virtualmethods::VirtualMethods;
 use crate::dom::window::Window;
@@ -65,7 +68,6 @@ use style::attr::AttrValue;
 use style::str::split_html_space_chars;
 
 use crate::dom::bindings::codegen::UnionTypes::RadioNodeListOrElement;
-use crate::dom::radionodelist::RadioNodeList;
 use std::collections::HashMap;
 use time::{now, Duration, Tm};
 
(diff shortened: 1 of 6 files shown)


pshaughn:
There are still more test failures than I'd expect. It might be better to leave this one unmerged until I've looked into them more.

KiChjang:
Just making sure, is there any overlap between this PR and #22264?

pshaughn:
I was unaware of #22264, @KiChjang. I've been treating the different subcases of live NodeLists as different problems and have been working on them in this, #25424, and #25428. It looks like I've been following a very different strategy from you, and #25424 passes some tests that #22264 doesn't.

I understand that you might feel you have priority on this and that I should step away from it to let you keep working on your version. If you do feel that way, let me know and I'll back off; I didn't intend to invalidate anyone else's work.

pshaughn:
Related issues that this change from static to live doesn't solve: #25429 (an element is staying gettable too long because it's not leaving the past names map), #21495 (an element with a `form=` attribute that appears earlier on the page than the form with that ID isn't getting into form.controls), #12978 (named getter can't get an element named addEventListener), #25036 (getOwnPropertyDescriptor doesn't find an element even when the named getter does)

pshaughn:
It occurs to me that, since this is always looking at form.controls and the past names map and never at the DOM tree directly, it could be optimized with a cursor like #22264 without even needing a mutation observer. It would just need to invalidate or repair the cursor on every code path that changes form.controls or the past names map.

pshaughn:
This push fixes the one bug I had that is within the purview of the list-getting itself (I wasn't excluding input type=image on the check that had to exclude it); it does not take into account the idea of using a cursor for performance.

components/script/dom/nodelist.rs
```diff
@@ -100,20 +105,22 @@ impl NodeList {
...
         } else {
-            panic!("called as_simple_list() on a children node list")
+            panic!("called as_radio_list() on a non-radio node list")
         }
     }
 
     pub fn iter<'a>(&'a self) -> impl Iterator<Item = DomRoot<Node>> + 'a {
         let len = self.Length();
+        // There is room for optimization here in non-simple cases,
+        // as calling Item repeatedly on a live list can involve redundant work.
```
Manishearth: File a followup for this?

components/script/dom/htmlformelement.rs
```diff
@@ -265,43 +330,23 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
-                        (child.has_attribute(&local_name!("name")) &&
-                            child.get_string_attribute(&local_name!("name")) == name)
-                    {
-                        candidates.push(DomRoot::from_ref(&*child.upcast::<Node>()));
-                    }
-                }
-            }
+        if candidates_length == 0 {
+            candidates = RadioNodeList::new_images(&window, self, name.clone());
+            candidates_length = candidates.Length();
```
Manishearth: we could use a separate function that just returns if there are zero, 1, or more (returning the element if it's one)

Manishearth (approved):
We should probably have a followup filed to move everything to mutation observers

Manishearth (approved):
We should probably have a followup filed to move everything to mutation observers

Issue #11868, opened 2016-06-26
MutableHandle<PropertyDescriptor> misused by mutating fields of .get() copies.


eddyb:
[Example in `script::dom::browsingcontext`](https://github.com/servo/servo/blob/8dcb12a019885d8333f35f43e43d3f2e0c56573a/components/script/dom/browsingcontext.rs#L269).
This looks bad, and `CodegenRust.py` generates a bunch more, but I have no idea what impact it has.
Also not sure what other types this happens with, I stumbled onto it during an unrelated refactoring.

The only solution I can think of is implementing `{,Mutable}Handle::get` only for pointers and `Value`.
That is, assuming there is no need to copy the whole `PropertyDescriptor` structure out.

jdm:
To elaborate - we could create a trait in https://github.com/servo/rust-mozjs/blob/master/src/rust.rs and implement that trait for `*mut T`. We could then add that trait to the bounds for [RootedGuard::get](https://github.com/servo/rust-mozjs/blob/3de4ff3d52361a47a17e3b4fcb02c779b99d93d4/src/rust.rs#L431-L433), [Handle::get](https://github.com/servo/rust-mozjs/blob/3de4ff3d52361a47a17e3b4fcb02c779b99d93d4/src/rust.rs#L475-L479), and [MutableHandle::get](https://github.com/servo/rust-mozjs/blob/3de4ff3d52361a47a17e3b4fcb02c779b99d93d4/src/rust.rs#L511-L515) and see what needs to be changed to keep Servo building.

avadacatavra:
It looks like this causes an issue for xows, so if no one else takes it, I probably will

Issue #25237, opened 2019-12-10
Intermittent sccache errors in linux mozjs build ([wintzimpl.o] Error 254)


jdm:
```
sccache: encountered fatal error
sccache: error : corrupt deflate stream
sccache:  cause: corrupt deflate stream
make[4]: *** [wintzimpl.o] Error 254
```
From https://community-tc.services.mozilla.com/tasks/CouK7ILuSYagtmfMSS-X4w

PR #25446, opened 2020-01-06
Make getOwnPropertyDescriptor hold the correct value for indexed/named properties

### LINKED ISSUE #25036: "Legacy platform object" getOwnPropertyDescriptor returns undefined when it should be finding dynamic properties

pshaughn:
This is towards #25036 and #25415; it could have far-reaching implications so I want to test it in isolation and see the results on the full test suite.

A few lines of code had a FIXME(#11868) despite that issue being closed, and looking for the pattern that was marked that way, I found one other unmarked instance of it. It doesn't immediately crash, and maybe it will pass some tests or fail some tests in informative ways.

EDIT: After adding an overlooked extended attribute to HTMLFormElement, this works very well indeed and seems to be worth merging!

---

- [X] These changes fix #25036

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -5147,8 +5147,7 @@ class CGDOMJSProxyHandler_getOwnPropertyDescriptor(CGAbstractExternMethod):
             attrs = "JSPROP_ENUMERATE"
             if self.descriptor.operations['IndexedSetter'] is None:
                 attrs += " | JSPROP_READONLY"
-            # FIXME(#11868) Should assign to desc.value, desc.get() is a copy.
-            fillDescriptor = ("desc.get().value = result_root.get();\n"
+            fillDescriptor = ("desc.value = result_root.get();\n"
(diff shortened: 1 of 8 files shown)


pshaughn:
I haven't seen this build failure before but it looks like mozjs, not anything I changed.

jdm:
Yep, that's #25237.

jdm:
@bors-servo retry
* #24726

PR #25449, opened 2020-01-07
Implement HTMLSelectElement.add() and indexed setter

### LINKED ISSUE #25003: Some ways of adding, getting, and removing HTMLSelectElement options don't work
WPT custom-elements/reactions/HTMLSelectElement.html adds and removes `<option>` elements from a `<select>` various ways. `select[0]=option` and `select.add(option)` fail to add, and `select[0] = null` fails to remove. `select.length = 0` and `select.remove(0)` do successfully remove. (appendChild and removeChild are not tested by this test)

pshaughn:
HTMLSelectElement.add and its indexed setter just needed to forward to matching methods in HTMLOptionsCollection, which they now do. It was also necessary to change codegen slightly; it had accidentally assumed that if an indexed setter existed, a named getter or setter also would.

I expect this and #25446 will combine to pass more tests than either alone does.

---

- [X] These changes fix #25003


Manishearth:
```
  ▶ Unexpected subtest result in /_mozilla/mozilla/union.html:
  │ FAIL [expected PASS] union
  └   → The object can not be found here.
```

feel free to tweak that test (or remove it) if it's wrong

pshaughn:
That test is both right and interesting. It's making sure that, if script calls a method that takes a union type using an argument that's incorrect for that union, codegen code throws a TypeError at the script. This was working when the method didn't have the [Throws] extended attribute, and now it isn't working.

pshaughn:
I stand corrected; the TypeError is still getting thrown, and some other line is throwing a NotFoundError for reasons as yet unclear.

pshaughn:
The test only worked because add() wasn't actually doing anything; I wasn't sure which of two behaviors was the one really being tested for so now it tries for both.

tests/wpt/mozilla/meta/MANIFEST.json
```diff
@@ -14053,7 +14051,7 @@
   ],
   "css/border_black_groove.html": [
    "49e1647a6f71e320770225ad537b4fd4020bd700",
-   "reftest_node"
```
pshaughn: I don't know what this means; I just ran update-manifest as usual and it changed this!

Manishearth: probably something changed in the runner :man_shrugging:

PR #25451, opened 2020-01-07
Add accountable-refcell as optional build time feature

### LINKED ISSUE #19308: add feature for accountable refcell in domrefcell
Add a feature to use https://crates.io/crates/accountable-refcell in order to get more information for debugging panics

kunalmohan:
It can be used with the format `./mach build --features refcell_backtrace`.

---

- [X] These changes fix #19308  (GitHub issue number if applicable)

- [X] These changes do not require tests because it adds a optional build-time feature only.


components/constellation/Cargo.toml
```diff
@@ -40,6 +41,7 @@ net = {path = "../net"}
 net_traits = {path = "../net_traits"}
 profile_traits = {path = "../profile_traits"}
 script_traits = {path = "../script_traits"}
+script = { path = "../script"}
```
jdm: This shouldn't be necessary. We can do this from `components/servo/Cargo.toml` instead.

components/script/body.rs
```diff
@@ -14,6 +14,8 @@ use crate::dom::formdata::FormData;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::promise::Promise;
 use crate::script_runtime::JSContext;
+#[cfg(feature = "refcell_backtrace")]
+use accountable_refcell::Ref;
```
jdm: Rather than changing all of these files like this, let's change the uses of `std::cell::Ref` to `dom::bindings::cell::Ref` and make that type re-export `std::cell::Ref` or `accountable_refcell::Ref` as needed.

components/script/dom/response.rs
```diff
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
-
```
jdm: nit: don't delete this, please.

ports/glutin/Cargo.toml
```diff
@@ -14,6 +14,9 @@ path = "main.rs"
 test = false
 bench = false
 
+[dependencies]
+script = { path = "../../components/script"}
```
jdm: This isn't necessary.

ports/glutin/Cargo.toml
```diff
@@ -42,6 +45,7 @@ max_log_level = ["log/release_max_level_info"]
 native-bluetooth = ["libservo/native-bluetooth"]
 no-wgl = ["libservo/no-wgl"]
 profilemozjs = ["libservo/profilemozjs"]
+refcell_backtrace = ["script/refcell_backtrace"]
```
jdm: We need to use `libservo/refcell_backtrace` instead.

ports/libsimpleservo/api/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 getopts = "0.2.11"
 libservo = { path = "../../../components/servo" }
 log = "0.4"
+script = { path = "../../../components/script"}
```
jdm: Remove this.

ports/libsimpleservo/api/Cargo.toml
```diff
@@ -48,6 +49,7 @@ native-bluetooth = ["libservo/native-bluetooth"]
 no_static_freetype = ["libservo/no_static_freetype"]
 no-wgl = ["libservo/no-wgl"]
 oculusvr = ["libservo/oculusvr"]
+refcell_backtrace = ["script/refcell_backtrace"]
```
jdm: libservo/refcell_backtrace

ports/libsimpleservo/capi/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bench = false
 
 [dependencies]
 simpleservo = { path = "../api" }
+script = { path = "../../../components/script"}
```
jdm: Remove this.

ports/libsimpleservo/capi/Cargo.toml
```diff
@@ -43,6 +44,7 @@ max_log_level = ["simpleservo/max_log_level"]
 native-bluetooth = ["simpleservo/native-bluetooth"]
 no-wgl = ["simpleservo/no-wgl"]
 oculusvr = ["simpleservo/oculusvr"]
+refcell_backtrace = ["script/refcell_backtrace"]
```
jdm: simpleservo/refcell_backtrace

jdm (changes requested):
We should add a CI job so this doesn't break, too - you can replace https://github.com/servo/servo/blob/master/etc/taskcluster/decision_task.py#L209 with refcell_backtrace.

components/constellation/Cargo.toml
kunalmohan: I don't completely understand what you re suggesting. Removing this will throw an error and I don't see `../servo` in dependencies. There is `../constellation` in dependencies of  `/components/servo` but how does that help?

jdm: components/servo/Cargo.toml already has a script dependency in it that we can use to propagate the refcell_backtrace feature. We shouldn't need to modify the constellation Cargo.toml at all.

kunalmohan: Just to clear it in my mind, `components/servo` will propagate it to `components/constellation` because the latter is a dependency of former, right? :sweat_smile:

jdm: No; components/servo will propagate it to components/script.

jdm: I would not expect to see a refcell_backtrace feature in components/constellation/Cargo.toml, so there should be no errors.

kunalmohan: Okay. I just re-read the cargo feature link you had mentioned in the issue that this PR addresses and things are now much clear to me.

jdm (approved):
Looks good! Please squash and I'll merge!

kunalmohan:
Done. Thanks for guiding me through this; learnt a lot while working on this issue. :smile:

jdm:
@bors-servo retry
* #22212 #24726

Issue #24297, opened 2019-09-25
Intermittent pass in /css/css-text/white-space/control-chars-096.html, /css/css-text/white-space/control-chars-010.html, /css/css-text/white-space/control-chars-01B.html


jdm:
```
{
    "status": "PASS", 
    "group": "default", 
    "message": null, 
    "stack": null, 
    "subtest": null, 
    "test": "/css/css-text/white-space/control-chars-096.html", 
    "line": 3324, 
    "action": "test_result", 
    "expected": "FAIL"
}

```

PR #25456, opened 2020-01-07
Past names expire when form owner is reset

### LINKED ISSUE #25429: "If an element listed in a form element's past names map changes form owner, then its entries must be removed from that map."

pshaughn:
On their way out of form.controls, elements now also leave form.past_names_map, passing one WPT test. We're already scanning linearly through form.controls linearly to get the index of the control there, so additionally scanning through the past names map by value shouldn't raise any particular performance concern.

---

- [X] These changes fix #25429

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -1081,6 +1081,13 @@ impl HTMLFormElement {
             .iter()
             .position(|c| &**c == control)
             .map(|idx| controls.remove(idx));
+
+        // https://html.spec.whatwg.org/multipage#forms.html#the-form-element:past-names-map-5
+        // "If an element listed in a form element's past names map
+        // changes form owner, then its entries must be removed
+        // from that map."
+        let mut past_names_map = self.past_names_map.borrow_mut();
+        past_names_map.retain(|_k, v| v.0 != control);
     }
 }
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Thanks!

jdm:
@bors-servo retry
* #24726 #24622

jdm:
@bors-servo retry
* #24297

jdm:
@bors-servo retry
* #24726

PR #25458, opened 2020-01-07
Add documentation for WindowsSdkDir to README

asajeffrey:
This was needed for me to get servo building on Windows 10 with VS2017.

---

- [x] These changes do not require tests because it's documentation


asajeffrey:
@jdm there may be something cleverer we can do about this, but copying from the windows kit seems to be what is recommended by https://docs.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=vs-2019#local-deployment

Issue #25265, opened 2019-12-12
Run adopt as part of insert


jdm:
https://github.com/whatwg/dom/pull/754

pshaughn:
This is bigger than I thought, but I think I can still do it. Challenges here:
* The concept of a document fragments' "host" element is directly exposed in ShadowRoot, but needs to actually be stored in DocumentFragment so that template elements can also use it; this makes the layout getter for it slightly more elaborate. (@jdm got me past the tricky parts of this in IRC)
* Insert code is putting a script blocker on the owner document at the start and taking a script blocker off the owner document at the end; now that adoption is happening in mid-insert the owner document isn't actually the same document in these two places, leading to a crash.
* At least one debug assertion about nodes sharing an owner doc during the insert process is false now.

pshaughn:
^ This PR moves adopt to where the spec says to move it, and adds the host-element check, but it's not passing as many tests as it seems like it should.

pshaughn:
https://github.com/whatwg/dom/pull/819 suggests this is not going to happen, or if it does happen it will be written a totally different way.

PR #25460, opened 2020-01-07
Run adopt as part of insert 

pshaughn:
Changes of whatwg/dom#754 are reflected, but we mostly aren't passing the associated tests. In many cases this is because we don't actually have the shadow DOM functionality that makes this behavior relevant, but that's not all that's being tested and there is probably some other missing piece.

It might make sense to merge this (assuming it passes CI) and open another issue investigating the failures?

---

- [X] These changes fulfill the letter of #25265 but aren't passing all the associated tests

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -3809,9 +3809,19 @@ impl DocumentMethods for Document {
         }
 
         // Step 3.
+        if let Some(fragment) = node.downcast::<DocumentFragment>() {
+            if fragment.get_host().is_some() {
+                // Spec text doesn't actually say what to return,
+                // but IDL doesn't allow null, and
+                // returning the node itself seems to be what's expected
+                return Ok(DomRoot::from_ref(node));
+            }
+        }
+
+        // Step 4
         Node::adopt(node, self);
 
-        // Step 4.
+        // Step 5.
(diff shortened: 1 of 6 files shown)


components/script/dom/document.rs
```diff
@@ -3809,9 +3809,19 @@ impl DocumentMethods for Document {
         }
 
         // Step 3.
+        if let Some(fragment) = node.downcast::<DocumentFragment>() {
+            if fragment.get_host().is_some() {
+                // Spec text doesn't actually say what to return,
+                // but IDL doesn't allow null, and
+                // returning the node itself seems to be what's expected
```
jdm: I went to open a spec issue about this and found https://github.com/whatwg/dom/issues/813.

jdm:
Given https://github.com/whatwg/dom/issues/813#issuecomment-570131386, perhaps we should hold off on this work until it settles in spec land.

pshaughn:
That makes sense.

pshaughn:
https://github.com/whatwg/dom/pull/819 hasn't landed yet but https://github.com/whatwg/dom/issues/813 is still unresolved, so if adopt is going to be moved, it will probably be in a different enough way to require completely different code changes.

PR #25467, opened 2020-01-09
Use dom and new_inherited in (extendable)messageevent

gterzian:
Also FIX #24707

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #25464 (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #25468, opened 2020-01-09
Add generic to frozen array

gterzian:
https://github.com/servo/servo/pull/25467#discussion_r364575071

Depends on https://github.com/servo/servo/pull/25467

---

### DIFF
--- a/components/script/dom/messageevent.rs
+++ b/components/script/dom/messageevent.rs
@@ -13,7 +13,7 @@ use crate::dom::bindings::reflector::reflect_dom_object;
 use crate::dom::bindings::root::{Dom, DomRoot};
 use crate::dom::bindings::str::DOMString;
 use crate::dom::bindings::trace::RootedTraceableBox;
-use crate::dom::bindings::utils::message_ports_to_frozen_array;
+use crate::dom::bindings::utils::to_frozen_array;
 use crate::dom::event::Event;
 use crate::dom::eventtarget::EventTarget;
 use crate::dom::globalscope::GlobalScope;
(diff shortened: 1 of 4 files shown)


gterzian:
@Manishearth r?

components/script/dom/messageevent.rs
```diff
@@ -219,7 +238,13 @@ impl MessageEventMethods for MessageEvent {
 
     /// <https://html.spec.whatwg.org/multipage/#dom-messageevent-ports>
     fn Ports(&self, cx: JSContext) -> JSVal {
-        message_ports_to_frozen_array(self.ports.borrow().as_slice(), cx)
+        let ports: Vec<DomRoot<MessagePort>> = self
```
Manishearth: It's actually worth caching this in a `MutNullableDom<Heap>` or something. The spec doesn't say it's SameObject, but it probably should, and even if it doesn't we can still do it.

Manishearth (approved):

components/script/dom/messageevent.rs
Manishearth: Doesn't have to be this PR

gterzian: Ok, I will look into it.

gterzian:
@Manishearth r?

Manishearth:
```
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/messageevent-constructor.https.html:
  └ PASS [expected FAIL] ports attribute should be a FrozenArray
```

gterzian:
`assert_true(ev.ports === ev.ports, "ev.ports should return the same object")`, so the caching was  necessary afterall...

PR #25477, opened 2020-01-09
Add taskcluster retriggering script.

jdm:


etc/taskcluster/retrigger.sh
```diff
@@ -0,0 +1,26 @@
...
+# eg. `./retrigger.sh 10 THkZut0wRq-SAmDSKIQGjg LFa41aanSHyLzlodxlgA9w` will
+# cause 20 new tasks to be scheduled for the associated task group(s).
+
+export TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com/
+eval $(taskcluster signin)
+times=$1
+shift
+for task in $@
+do
+    for (( i = 1 ; i < $times; i++ ))
```
SimonSapin: Is this an off-by-one error? To iterate `$times` times, wouldn’t you need to start at `i = 0` *or* compare with `<=`?

SimonSapin:
Also tidy appears to be unhappy.

PR #25488, opened 2020-01-11
Event dispatch rewritten to align to spec, activate on clicks better

### LINKED ISSUE #22783: Integrate activation behaviour into event dispatch algorithm
`synthetic_click_activation` implements an old algorithm for activating elements via synthetic mouse click events. We should rewrite it to implement [this synthetic mouse event algorithm](https://html.spec.whatwg.org/multipage/#fire-a-synthetic-mouse-event) and integrate the activation behaviour into the event dispatch algorithm [per the spec](https://dom.spec.whatwg.org/#concept-event-dispatch). This will address the comment in `Document::dispatch_key_event` about synthetic click activiation being unspecified, since it has been specified appropriately in the past 4 years.

Code: `components/script/dom/document.rs`, `components/script/dom/eventtarget.rs`, `components/script/activation.rs`

### LINKED ISSUE #25199: .click() on an img inside a td, or a q inside a button, apparently isn't firing the td/button's onclick
This happens in html/webappapis/scripting/events/compile-event-handler-lexical-scopes.html. The fact that the event doesn't even fire causes breakage before the test can get around to actually testing the https://html.spec.whatwg.org/multipage/webappapis.html#getting-the-current-value-of-the-event-handler behavior it's intended to test.

### LINKED ISSUE #25384: Synthetic click events aren't putting checkboxes/radio buttons in the right state in some cases
In WPT html/semantics/forms/the-input-element/checkbox-click-events.html it looks like calling preventDefault in a click listener isn't stopping .click() from toggling the box (though the listener is called), and dispatching a new MouseEvent("click") when a click listener has been defined is calling the listener but not toggling the box.

pshaughn:
I went over the changes to the event dispatch spec that had accumulated over the past few years, rewriting dispatch/invoke/inner-invoke almost completely and modifying other code where it was relevant. Most of the remaining obvious deviations from spec are things that will only come up when we start handling events in shadow DOM.

I am pushing now because I want to see CI test results, but please do not approve this PR just if automated test improvements look good. I may have broken some actual UI interactions in the course of fixing synthetic events, and some manual testing is needed, including checking that manual interactions with interactive content continue to fire the events they're supposed to.

---

- [X] These changes fix #25384 and fix #22783 and fix #25199 

- [ ] There are automated tests for the synthetic-click parts of these changes, BUT the effects on real UI events need some manual testing before merging


pshaughn:
/webdriver/tests/element_send_keys/ tests broke so I have definitely mangled something in the less-synthetic input pipeline.
_mozilla/mozilla/event_dispatch_order.html also broke and bears some looking into...
other than that there are just a bunch of new successes in unexpected places, so far so good

pshaughn:
I think _mozilla/mozilla/event_dispatch_order.html is wrong for the current spec. Even though event.phase is AT_TARGET, we have to go through the listeners based on the phase that was passed to invoke, NOT event.phase, and that means all the captures first and then all the bubbles. https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke steps 3 and 4 are the decision point.

I would like another pair of eyes on it before I change the test to match my understanding.

pshaughn:
https://github.com/servo/servo/blob/c5069835d72bb7c14034c13674b6d1b500a95425/components/script/dom/event.rs#L179-L185 represented the concept of "default action", by which a DOM-observable event could cause a node to do something other than activation. 

https://dom.spec.whatwg.org/#action-versus-occurance suggests there may have been a big philosophical shift since then.

I haven't found anything in the current spec saying that a KeyboardEvent fired at a textbox should cause a character to be typed. MouseEvents named click are special and cause activation, but I haven't seen anything that says a KeyboardEvent has an effect in the absence of an actual event listener.

Our webdriver code seems to work entirely by dispatching KeyEvents via DOM, not hooking into any other layer of UI logic, and that seems to be why this PR is failing webdriver tests that expect key events to produce an outcome.

I might be drastically misunderstanding something, however, and I'd like @jdm and others to take a look at it.

jdm:
As far as I know, the webdriver key inputs go through the compositor (https://github.com/servo/servo/blob/c5069835d72bb7c14034c13674b6d1b500a95425/components/constellation/constellation.rs#L4148-L4198) the same as the UI events. They're supposed to be indistinguishable from normal user input.

pshaughn:
It looks like that leads to document.dispatch_key_event, which then fires a KeyboardEvent at whatever's focused. If I'm not missing something major in the spec, a KeyboardEvent is not in itself what a text control is supposed to be responding to. This feels weird enough that I'm leaning towards the possibility that I'm just missing it, though.

jdm:
My reading of parts of the UI Events spec like https://w3c.github.io/uievents/#event-flow-default-cancel and https://w3c.github.io/uievents/#keys-cancelable-keys suggests that the handling of editing a text field comes as part of the default action of a keydown event (https://w3c.github.io/uievents/#keydown).

pshaughn:
It seems like the UI Event spec is out of sync with the DOM and HTML specs, as nothing in the latter specifies that input elements have a default action for keydowns, nor specifies when to apply any default action that exists. I'll put the vtable.handle_event into the refactored code at the closest approximation of its old position and note the discrepancy.

pshaughn:
I can type and backspace in an <input> on an actual window (linux, LIBGL_ALWAYS_SOFTWARE=1) and that one webdriver test passes for me.
@bors-servo try=wpt

pshaughn:
I am going to guess that timeout wasn't my changes' fault.
@bors-servo try=wpt

pshaughn:
One blank reftest intermittent, and the one _mozilla test that I suspect is wrong. I feel this changes enough about UI events that it needs actual interacting with on different platforms (including ones I don't have), but it looks clean.

components/script/dom/activation.rs
```diff
@@ -21,13 +17,17 @@ pub trait Activatable {
...
-    // https://html.spec.whatwg.org/multipage/#run-pre-click-activation-steps
-    fn pre_click_activation(&self);
+    // https://dom.spec.whatwg.org/multipage/#eventtarget-legacy-pre-activation-behavior
+    fn legacy_pre_activation_behavior(&self) -> Option<InputActivationState> {
+        None
+    }
 
-    // https://html.spec.whatwg.org/multipage/#run-canceled-activation-steps
-    fn canceled_activation(&self);
+    // https://dom.spec.whatwg.org/multipage/#eventtarget-legacy-canceled-activation-behavior
```
CYBAI:
```suggestion
    // https://dom.spec.whatwg.org/#eventtarget-legacy-canceled-activation-behavior
```

components/script/dom/activation.rs
```diff
@@ -21,13 +17,17 @@ pub trait Activatable {
...
+    fn legacy_canceled_activation_behavior(&self, _state_before: Option<InputActivationState>) {}
 
-    // https://html.spec.whatwg.org/multipage/#run-post-click-activation-steps
+    // https://dom.spec.whatwg.org/multipage/#eventtarget-activation-behavior
```
CYBAI:
```suggestion
    // https://dom.spec.whatwg.org/#eventtarget-activation-behavior
```

components/script/dom/htmlinputelement.rs
```diff
@@ -2358,99 +2335,96 @@ impl Activatable for HTMLInputElement {
         }
     }
 
-    // https://html.spec.whatwg.org/multipage/#run-pre-click-activation-steps
-    #[allow(unsafe_code)]
-    fn pre_click_activation(&self) {
-        let mut cache = self.activation_state.borrow_mut();
+    // https://dom.spec.whatwg.org/multipage/#eventtarget-legacy-pre-activation-behavior
```
CYBAI:
```suggestion
    // https://dom.spec.whatwg.org/#eventtarget-legacy-pre-activation-behavior
```

components/script/dom/htmlinputelement.rs
```diff
@@ -2358,99 +2335,96 @@ impl Activatable for HTMLInputElement {
...
     }
 
-    // https://html.spec.whatwg.org/multipage/#run-canceled-activation-steps
-    fn canceled_activation(&self) {
-        let cache = self.activation_state.borrow();
-        let ty = self.input_type();
-        if cache.old_type != ty {
-            // Type changed, abandon ship
-            // https://www.w3.org/Bugs/Public/show_bug.cgi?id=27414
+    // https://dom.spec.whatwg.org/multipage/#eventtarget-legacy-canceled-activation-behavior
```
CYBAI:
```suggestion
    // https://dom.spec.whatwg.org/#eventtarget-legacy-canceled-activation-behavior
```

components/script/dom/window.rs
```diff
@@ -536,6 +536,14 @@ impl Window {
     pub fn get_event_loop_waker(&self) -> Option<Box<dyn EventLoopWaker>> {
         self.event_loop_waker.as_ref().map(|w| (*w).clone_box())
     }
+
+    // see note at https://dom/spec/whatwg.org/#event-listener-removed step 2
```
CYBAI:
```suggestion
    // see note at https://dom.spec.whatwg.org/#event-listener-removed step 2
```

CYBAI (commented):
Found some invalid DOM spec links 👀 

(Maybe it's good to have a lint rule for it?)

jdm:
@asajeffrey Review ping?

pshaughn:
Just did a non-trivial rebase (due to #25490 also affecting the legacy pre-click/post-click behaviors) and it's probably a good idea to see those tests again.
@bors-servo retry

CYBAI:
@bors-servo try=wpt

(@pshaughn as you've force pushed, you will need to call `try` again instead of `retry` here ;)

CYBAI:
Also, tidy complains about fmt 👀

```
Diff in /repo/components/script/dom/htmlinputelement.rs at line 2361:
                 });
             },
             InputType::Radio => {
-                let checked_member = radio_group_iter(self, self.radio_group_name().as_ref())
(B-                    .find(|r| r.Checked());
(B+                let checked_member =
(B+                    radio_group_iter(self, self.radio_group_name().as_ref()).find(|r| r.Checked());
(B                 let was_checked = self.Checked();
                 self.SetChecked(true);
                 return Some(InputActivationState {
Diff in /repo/components/script/dom/htmlinputelement.rs at line 2416:
                         &o,
                         self.form_owner().as_deref(),
                         self.radio_group_name().as_ref(),
-			Some(&*tree_root),
(B+                        Some(&*tree_root),
(B                     ) {
                         o.SetChecked(true);
                     } else {
clang-format not installed. Skipping CPP formatting.
```

pshaughn:
Thank you, CYBAI. I had run mach fmt but I forgot to add the change it made before I commited/pushed. Thank you also for pointing out the try/retry distinction.
@bors-servo try=wpt

CYBAI:
```
  ▶ Unexpected subtest result in /_mozilla/mozilla/event_dispatch_order.html:
  │ FAIL [expected PASS] event_dispatch_order
  │   → assert_equals: expected true but got false
  │ 
  │ handler2@http://web-platform.test:8000/_mozilla/mozilla/event_dispatch_order.html:21:5
  │ Test.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:2049:35
  │ @http://web-platform.test:8000/_mozilla/mozilla/event_dispatch_order.html:37:10
  └ @http://web-platform.test:8000/_mozilla/mozilla/event_dispatch_order.html:9:1
```

pshaughn:
event_dispatch_order is a weird one, I mentioned it before at https://github.com/servo/servo/pull/25488#issuecomment-573264871

components/script/dom/htmlinputelement.rs
```diff
@@ -2197,14 +2174,19 @@ impl VirtualMethods for HTMLInputElement {
...
+    // https://dom.spec.whatwg.org/#action-versus-occurance
     fn handle_event(&self, event: &Event) {
         if let Some(s) = self.super_type() {
             s.handle_event(event);
         }
 
         if event.type_() == atom!("click") && !event.DefaultPrevented() {
-            // TODO: Dispatch events for non activatable inputs
-            // https://html.spec.whatwg.org/multipage/#common-input-element-events
+            // WHATWG-specified ctivation behaviors are handled elsewhere;
```
jdm: nit: s/ctivation/activation/

components/script/dom/event.rs
```diff
@@ -147,49 +166,195 @@ impl Event {
...
     ) -> EventStatus {
-        assert!(!self.dispatching());
-        assert!(self.initialized());
-        assert_eq!(self.phase.get(), EventPhase::None);
-        assert!(self.GetCurrentTarget().is_none());
-
         // Step 1.
         self.dispatching.set(true);
 
+        warn!("In dispatch, type is {:?}", self.type_());
```
jdm: This should be a debug! or trace!, right? Same for other new uses in this file.

components/script/dom/event.rs
```diff
@@ -147,49 +166,195 @@ impl Event {
...
+                Some(ListenerPhase::Capturing),
+            );
+        }
+
+        // Step 5.14
+        for object in event_path.iter() {
+            if &**object == &*target {
+                self.phase.set(EventPhase::AtTarget);
+
+                self.target.set(Some(target_override));
```
jdm: Seems like we could de-duplicate these two blocks by only setting phase inside the if/else.

jdm (changes requested):
This is really readable! I only have a few requests, and the remaining suggestions from cybai's earlier review should be addressed as well.

jdm:
I agree that event_dispatch_order.html doesn't agree with the up to date specs. Go ahead and change it to match better.

components/script/dom/event.rs
pshaughn: Most of these don't even have a reason to exist now that it generally works; they're just forgotten leftovers.

components/script/dom/event.rs
pshaughn: Yes, that would be easier to read. Thank you.

pshaughn:
Changes are pushed; be sure to give this a spin on every platform before accepting it. I wouldn't be surprised if it fixed some random web compatibility issues.

jdm:
I don't think there's a big need to test this out on different platforms - the code being changed is platform-independent, and our form controls are all platform-independent as well.

@bors-servo r+

jdm:
@bors-servo retry
#24726

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #25753

jdm:
@bors-servo retry
* network issue

jdm:
@bors-servo retry
* network

jdm:
@bors-servo retry
* #24726

PR #25494, opened 2020-01-11
Make sure input value=12:30:01 doesn't turn into 12:30:1

### LINKED ISSUE #25493: datetime-local value sanitization removes leading zeroes for seconds
```
<input type="datetime-local" value="2000-01-01T12:30:00.5">
<input type="datetime-local" value="2000-01-01T12:30:01">
<script>
    let result = '';
    for (let elem of document.querySelectorAll("input")) {
        result += elem.valueAsNumber + ' ' + elem.value + '\n';
    }
    alert(result);
</script>
```

In the above example the leading zeroes for seconds are removed during sanitization, but they are required in valid datetime-local strings. This makes valueAsNumber fail, for example.

https://html.spec.whatwg.org/multipage/#valid-normalised-local-date-and-time-string
https://html.spec.whatwg.org/multipage/#valid-time-string

https://github.com/servo/servo/blob/63cb13f5fa5f52d3574eb0da883f881b52e208ff/components/script/dom/bindings/str.rs#L469-L484

pshaughn:
Fixed convert_valid_normalized_local_date_and_time_string to add the mandatory leading 0 before seconds values that needed it, added tests to see that various-length strings with zeroes in them would roundtrip properly.

---

- [X] These changes fix #25493

### DIFF
--- a/components/script/dom/bindings/str.rs
+++ b/components/script/dom/bindings/str.rs
@@ -474,7 +474,15 @@ impl DOMString {
                 "{:04}-{:02}-{:02}T{:02}:{:02}",
                 year, month, day, hour, minute
             );
+        } else if second < 10.0 {
+            // we need exactly one leading zero on the seconds,
+            // whatever their total string length might be
+            self.0 = format!(
+                "{:04}-{:02}-{:02}T{:02}:{:02}:0{}",
+                year, month, day, hour, minute, second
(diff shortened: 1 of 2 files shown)


jdm (approved):

jdm:
This can merge with a rebase.

PR #25510, opened 2020-01-13
Implement queueMicrotask

### LINKED ISSUE #21574: Implement self.queueMicrotask()
whatwg/html#2789

Spec: https://html.spec.whatwg.org/multipage/#microtask-queuing

Related Bugzilla for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1480236

Also, we'll need to update the `webidl` of `WindowOrWorkerGlobalScope` for the method

```webidl
  // microtask queuing
  void queueMicrotask(VoidFunction callback);
```

Ref: https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin:dom-queuemicrotask

pshaughn:
The VoidFunction of a user-queued microtask is identical to a PromiseJobCallback in everything but type name, so implementing this method turned out to be just a matter of attaching a WebIDL frontend to a copy-and-pasted backend. All the remaining queueMicrotask WPT failures are because we don't have shared/service workers.

---

- [X] These changes fix #21574

### DIFF
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -11,6 +11,7 @@ use crate::dom::bindings::codegen::Bindings::HistoryBinding::HistoryBinding::His
 use crate::dom::bindings::codegen::Bindings::MediaQueryListBinding::MediaQueryListBinding::MediaQueryListMethods;
 use crate::dom::bindings::codegen::Bindings::PermissionStatusBinding::PermissionState;
 use crate::dom::bindings::codegen::Bindings::RequestBinding::RequestInit;
+use crate::dom::bindings::codegen::Bindings::VoidFunctionBinding::VoidFunction;
 use crate::dom::bindings::codegen::Bindings::WindowBinding::{
     self, FrameRequestCallback, WindowMethods, WindowPostMessageOptions,
 };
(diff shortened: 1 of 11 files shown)


components/script/dom/window.rs
```diff
@@ -871,6 +872,12 @@ impl WindowMethods for Window {
         self.ClearTimeout(handle);
     }
 
+    // https://html.spec.whatwg.org/multipage#microtask_queuing
```
jdm: This link doesn't work; maybe we want https://html.spec.whatwg.org/multipage/#dom-queuemicrotask instead?

components/script/dom/workerglobalscope.rs
```diff
@@ -341,6 +342,12 @@ impl WorkerGlobalScopeMethods for WorkerGlobalScope {
         self.ClearTimeout(handle);
     }
 
+    // https://html.spec.whatwg.org/multipage#microtask_queuing
```
jdm: Same comment about the link here.

jdm (changes requested):
Implementation looks good!

components/script/dom/window.rs
pshaughn: Thanks for noticing. I had an underscore instead of a hyphen, but the dom-methodname link is what Servo usually has for methods so I'll make it that instead.

jdm:
@bors-servo retry
* #23290

jdm:
@bors-servo retry
* #25513

PR #25512, opened 2020-01-13
Render `background-image: url(…)`

SimonSapin:
… and support most `background-*` properties.

Still to do:

* `background-attachment`
* Gradients


components/layout_thread_2020/lib.rs
```diff
@@ -553,13 +563,15 @@ impl LayoutThread {
     fn build_layout_context<'a>(
         &'a self,
         guards: StylesheetGuards<'a>,
+        script_initiated_layout: bool,
```
nox: Maybe an enum, `LayoutRequestOrigin` or something?

components/layout_thread_2020/lib.rs
```diff
@@ -248,10 +260,10 @@ impl LayoutThreadFactory for LayoutThread {
...
         load_webfonts_synchronously: bool,
         window_size: WindowSizeData,
-        _dump_display_list: bool,
-        _dump_display_list_json: bool,
-        _dump_style_tree: bool,
-        _dump_rule_tree: bool,
+        dump_display_list: bool,
+        dump_display_list_json: bool,
+        dump_style_tree: bool,
+        dump_rule_tree: bool,
```
nox: Such an unethical method. 😂 No nit here though, we don't have time for that.

components/layout_2020/Cargo.toml
```diff
@@ -25,6 +25,7 @@ ipc-channel = "0.12"
 libc = "0.2"
 msg = {path = "../msg"}
 net_traits = {path = "../net_traits"}
+once_cell = "1.2.0"
```
nox: We already use `mitochondria`, which has an `OnceCell`, I think.

components/layout_2020/display_list.rs
```diff
@@ -426,7 +426,7 @@ impl<'a> BuilderForBoxFragment<'a> {
             clipping_area_origin: f32,
             clipping_area_size: f32,
             positioning_area_size: f32,
-        ) -> (f32, f32) {
+        ) -> (bool, f32, f32) {
```
nox: Struct would be nice here.

nox (commented):
Nice comments all around.

components/layout_thread_2020/lib.rs
SimonSapin: This is copied from `components/layout_thread`

components/layout_thread_2020/lib.rs
SimonSapin: Yeah it’s pretty bad, but also copied over.

components/layout_2020/Cargo.toml
SimonSapin: Switched

components/layout_2020/display_list.rs
SimonSapin: Done

SimonSapin:
@bors-servo retry

https://github.com/servo/servo/issues/25523 ?

```
  ▶ TIMEOUT [expected OK] /html/semantics/forms/the-button-element/button-submit-remove-children-jssubmit.html
  │ 
  │ 
  │ 
  └ 
  ▶ Unexpected subtest result in /html/semantics/forms/the-button-element/button-submit-remove-children-jssubmit.html:
  │ TIMEOUT [expected PASS] This test will pass if a form navigation successfully occurs when clicking a child element of a <button type=submit> element with a onclick event handler which removes the button's child and then calls form.submit().
  └   → Test timed out

Issue #25514, opened 2020-01-13
When parsing HTML as a result of assigning to srcdoc, the Javascript execution stack isn't empty.


pshaughn:
WPT custom-elements/perform-microtask-checkpoint-before-construction.html assigns some HTML to an iframe's srcdoc attribute and expects the parser to perform microtask checkpoints as it runs. This calls HTMLIFrameElement::navigate_or_reload_child_browsing_context, which does things at the wrong time and ends up seeing a non-empty Javascript stack at times when it is supposed to see an empty one, so no microtask checkpoints are performed. This is similar to #24901 and is probably best solved at the same time as that.

jdm:
We pass this test now.

Issue #25513, opened 2020-01-13
Frequent intermittent timeouts in macOS WPT tests


jdm:
```
6 unexpected results that are NOT known-intermittents:
  â–¶ TIMEOUT [expected PASS] /css/css-flexbox/flexbox_align-items-flexend.html
  â”‚ 
  â”‚ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”” 36 threads are still running after shutdown (bad).


  â–¶ TIMEOUT [expected PASS] /css/css-ui/outline-007.html
  â”‚ 
  â”‚ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”‚ 44 threads are still running after shutdown (bad).
  â”‚ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”” 19 threads are still running after shutdown (bad).


  â–¶ TIMEOUT [expected FAIL] /css/css-ui/box-sizing-012.html
  â”‚ 
  â”‚ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”” 19 threads are still running after shutdown (bad).


  â–¶ TIMEOUT [expected FAIL] /css/css-ui/box-sizing-024.html
  â”‚ 
  â”‚ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”” 40 threads are still running after shutdown (bad).


  â–¶ TIMEOUT [expected FAIL] /css/css-ui/box-sizing-008.html
  â”‚ 
[...]

PR #25515, opened 2020-01-13
Add microtask checkpoints to script elements and custom elements

### LINKED ISSUE #25016: Timing of custom element microtasks isn't the way WPT expects it to be
WPT custom-elements/microtasks-and-constructors.html and perform-microtask-checkpoint-before-construction.html test many specific expectations about when the microtasks of custom elements happen, and Servo violates many of those expectations.

The microtasks-and-constructors test is a little hard to read, since it's broken up into separate script elements to make interesting event orders happen while the document is still being parsed. My reading of it is that multiple script bodies seem to be firing in a row, where the callbacks of custom elements between those scripts are expected be firing in between them. I'm not sure I understand it, though.

pshaughn:
Servo had a microtask checkpoint at the end of running a script, but there was also supposed to be one at the end of HTML-parsing a script element before Javascript-parsing the script itself, and there were supposed to be checkpoints immediately after the call to a custom element constructor. This adds those, passing all cases of one WPT test file.

---

- [X] These changes fix #25016 except for the remaining not-really-about-microtasks case #25514


jdm:
```
  ▶ Unexpected subtest result in /dom/nodes/MutationObserver-document.html:
  └ PASS [expected FAIL] parser insertion mutations
  ▶ Unexpected subtest result in /dom/nodes/MutationObserver-document.html:
  └ PASS [expected FAIL] parser script insertion mutation
  ▶ Unexpected subtest result in /dom/nodes/MutationObserver-document.html:
  └ PASS [expected FAIL] removal of parent during parsing
```
This also changes the behaviour of a reftest, interestingly:
```
  ▶ FAIL [expected PASS] /_mozilla/mozilla/FileAPI/blob_url_upload.html
  │   → /_mozilla/mozilla/FileAPI/blob_url_upload.html 072d32c1bd0b78ac1c5af7657d6e5cedfb9ec8d6
  └   → /_mozilla/mozilla/FileAPI/blob_url_upload_ref.html 84f6d3bd2a1b363f32fc2f41c2c8dc78d0dfd881
```

pshaughn:
That mozilla blob one happens intermittently for me on this branch; I'll see if it also happens intermittently for me on master. My first guess here is that the screenshot has a chance of being taken before the file is done loading.

pshaughn:
It doesn't seem to happen on master.

Hypothesis: The new checkpoint that happens when parsing `</script>` before compiling the script body fires an ImageMicrotask, that ImageMicrotask does something that tells the document "I'm done being loaded", the document sees that html5ever is done parsing it, and the load event fires even though we're still inside the tokenizer for the final script element.

The one thing I know I didn't follow the exact spec for was that the checkpoint is supposed to happen before "Switch the insertion mode to the original insertion mode", but there wasn't a way to do that with the way Servo and html5ever are connected so I hoped it wasn't observable. I need to look at how the document decides whether it's done being loaded; maybe the insertion mode is part of the problem.

pshaughn:
Nothing is logged from document, servoparser, or htmlimageelement between entering and exiting the suspicious checkpoint; all the interesting action happens after it.

Running it with logs on, in this branch it looks there is a race between "Document got finish_load: Image([blob url])" and "Document got finish_load: PageSource([test url]), and if Image comes first the test fails. Next question I need to look at: in master, do these events always have a strict order (so I need to make sure one waits for the other), or do we pass the test regardless of order (so I need to make something else insensitive to the difference)?

pshaughn:
It looks like with my changes, the document is seeing the finish_load for the the image before the image's process_response has been reached (but after the image has had its blob URL assigned). That then means we have a race between when process_response happens and when the screenshot is taken. In master, the finish_load for the image always happens after process_response. These calls all come later in the timeline than my actual code changes, but I can infer that adding a microtask checkpoint breaks some expectation an image had about its lifecycle, and finding that expectation is the next step.

pshaughn:
https://github.com/servo/servo/blob/85ea8df00c794db3662511b255335f9252780544/components/script/dom/htmlimageelement.rs#L843-L847
This seems to be the thing that is firing at the wrong time; moving it out of the task like the comment says, nothing immediately crashes and the one reftest seems to pass consistently. I'm expecting something else to break.

pshaughn:
Just another one of those background-color intermittents and the new MutationObserver passes. I am surprised that change worked so smoothly.

jdm:
@bors-servo retry
 * #25513 #23290

PR #25519, opened 2020-01-14
Add support for WebGL2 unsigned uniform operations

mmatyas:
This adds support for the WebGL2 `uniform[1234]ui` and `uniform[1234]uiv` operations.

See: https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.8

cc @jdm @zakorgy @imiklos 

---


components/script/dom/webgl2renderingcontext.rs
```diff
@@ -1524,6 +1707,53 @@ impl WebGL2RenderingContextMethods for WebGL2RenderingContext {
...
+            }
+            let val = match val {
+                Uint32ArrayOrUnsignedLongSequence::Uint32Array(v) => v.to_vec(),
+                Uint32ArrayOrUnsignedLongSequence::UnsignedLongSequence(v) => v,
+            };
+
+            let (offset, length) =
+                self.check_uniform_vec_section(src_offset, src_length, val.len())?;
+            let val = val[offset..offset + length].to_vec();
+            if val.len() < 4 || val.len() % 4 != 0 {
```
jdm: Why don't we change check_uniform_vec_section to reduce some code duplication in these methods?
* it can take a Uint32ArrayOrUnsignedLongSequence argument and extract the vector
* it can check if offset is 0 and length is the vector length and return the original vector without copying anything
* it can take an argument for uniform size and return an error if the length doesn't match
* it can take a location argument and return an error if the location size is none and the array length doesn't match
* finally it can return the adjusted vector

components/script/dom/webglrenderingcontext.rs
```diff
@@ -1214,6 +1235,26 @@ impl WebGLRenderingContext {
...
         self.current_program.get()
     }
+
+    pub fn uniform_check_program(
+        &self,
+        program: &WebGLProgram,
+        location: &WebGLUniformLocation,
+    ) -> WebGLResult<()> {
+        self.validate_ownership(program)?;
+        //handle_potential_webgl_error!(self, self.validate_ownership(program), return NullValue());
```
jdm: Why is this commented?

jdm (changes requested):
Looking good!

components/script/dom/webglrenderingcontext.rs
mmatyas: Whoops.

components/script/dom/webgl2renderingcontext.rs
mmatyas: Ok, updated, thanks!

Issue #25498, opened 2020-01-12
alert leads to code execution on Linux


YtvwlD:
I took a quick look at the issues, but I didn't find a similar one. Also, I hope that this is the right repository to report this.

Opening this HTML file executes code on the client:

```html
<html>
<script>
alert("Hey, `id`!");
</script>
</html>
```

In the console this produces the following output which is fine:

```
ALERT: Hey, `id`!
```

But the popup shows that `id` is indeed being run:

![grafik](https://user-images.githubusercontent.com/5578100/72219536-6b74fa80-3547-11ea-97e7-d644ac24f361.png)

I reported this issue over at https://github.com/jdm/tinyfiledialogs-rs/issues/19 half a year ago but I'm not sure it is going to be fixed there.

I downloaded the nightly build from servo.org today and ran it under Ubuntu 18.04 amd64.

pshaughn:
If we want a technically standard-compliant quick fix here, we're allowed to arbitrarily truncate the string at any arbitrary point according to https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#optionally-truncate-a-simple-dialog-string, so we could truncate it before the first unsafe character.

jdm:
I think fixing it in https://github.com/jdm/tinyfiledialogs-rs/issues/19 makes sense.

pshaughn:
I'm now working on a mitigation to try to plug the known hole quickly and I'll have a PR for it tonight. There may be other hazards in tinyfiledialogs that we don't know about, as it has many different codepaths depending on which potentially dialog-displaying executables it can find on the user's system.

pshaughn:
#25525 should take the immediate severity off of this.

pshaughn:
I'd like to mention two possible ways forward from this:

**1**
I really like how direct and straightforward https://github.com/saurvs/nfd-rs compared to tinyfiledialogs. Unfortunately, part of that directness is that nfd-rs has _just_ file dialogs, not simpler things like alert boxes and input boxes which we currently also use from tinyfiledialogs.

I think it might make sense for someone with access to Windows/Linux/Mac buildchains, as a slightly ambitious but pretty reasonable project, to fork nfd-rs, clean up any dependency conflicts it might have with Servo, and add those simpler cases to it.

**2**
[...]

PR #25525, opened 2020-01-15
Mitigation for #25498

pshaughn:
This is not a complete solution:
* The alert string can get a bit mangled in some cases, but not to the point of unreadability.
* tinyfiledialogs has many codepaths that can pass strings to various different potentially-dialog-displaying executables; I do not know if some of those executables have their own unique escaping requirements.
* If some form of the same problem exists on OSX or Windows, this does not address them.

While imperfect, this is an improvement over continuing to have a known way for page authors to execute arbitrary shell script.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4799,6 +4799,7 @@ dependencies = [
  "osmesa-sys",
  "rust-webvr",
  "servo-media",
+ "shellwords",
  "sig",
  "tinyfiledialogs",
  "webxr",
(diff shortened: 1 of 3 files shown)


pshaughn:
Explanation of changes: shellwords::escape mostly solves the problem, but nonprinting control characters sometimes caused a crash, and the '<', '>' and '&' characters weren't escaping properly, so I stripped the nonprinting characters and replaced the other three with their Unicode full-width lookalikes. I did this to every string going into tinyfiledialogs that appeared to be something the page author might have control over.

jdm:
This looks like an improvement over the status quo. Thanks!
@bors-servo r+

jdm:
@bors-servo retry
* #25528

PR #25526, opened 2020-01-15
Initial implementation of GPUBindGroupLayout for WebGPU

imiklos:
Added WebIDL bindings for `GPUBindGroupLayout`, `GPUBindGroupLayoutDescriptor`, `GPUBindingType`, `GPUShaderStage` and `GPUBindGroupLayoutBinding` (Note: The servo's codegen doesn't like the name, because its already occupied).
Implemented the `createBindGroupLayout` function of `GPUDevice`.

---

cc @kvark @jdm @zakorgy


imiklos:
Added `peek-poke` and `peek-poke-derive` to the ignore list of duplicate versions.

components/script/dom/gpubindgrouplayout.rs
```diff
@@ -0,0 +1,75 @@
...
+use dom_struct::dom_struct;
+use std::cell::Cell;
+use webgpu::{WebGPU, WebGPUBindGroupLayout};
+
+#[dom_struct]
+pub struct GPUBindGroupLayout {
+    reflector_: Reflector,
+    label: DomRefCell<Option<DOMString>>,
+    bind_group_layout: WebGPUBindGroupLayout,
+    #[ignore_malloc_size_of = "mozjs"]
```
jdm: Use `"defined in webgpu"` here instead.

components/script/dom/gpubindgrouplayout.rs
```diff
@@ -0,0 +1,75 @@
...
+    bindings: Vec<GPUBindGroupLayoutBindings>,
+    #[ignore_malloc_size_of = "channels are hard"]
```
jdm: Same here - this isn't a channel, as far as I know.

components/script/dom/gpudevice.rs
```diff
@@ -229,4 +236,181 @@ impl GPUDeviceMethods for GPUDevice {
...
+        let bindings = descriptor
+            .bindings
+            .iter()
+            .map(|bind| {
+                // TODO: binding must be >= 0
+                storeBindings.insert(bind.binding);
+                let visibility = match ShaderStage::from_bits(bind.visibility) {
+                    Some(visibility) => visibility,
+                    None => {
+                        valid &= false;
```
jdm: This is the same as `valid = false`, right? If so, let's use that instead. Same comment about later uses in this method, too.

components/script/dom/gpudevice.rs
```diff
@@ -229,4 +236,181 @@ impl GPUDeviceMethods for GPUDevice {
...
+                    multisampled: bind.multisampled,
+                    texture_dimension: webgpu::wgpu::resource::TextureViewDimension::D2, // Use as default for now
+                }
+            })
+            .collect::<Vec<BindGroupLayoutBinding>>();
+
+        // bindings are unique
+        valid &= storeBindings.len() == bindings.len();
+
+        // Values does not exceeded the max limit for each ShaderStage
```
jdm: `// Ensure that values do not exceed the max limit for each ShaderStage.`

components/script/dom/gpudevice.rs
```diff
@@ -229,4 +236,181 @@ impl GPUDeviceMethods for GPUDevice {
...
+        for each in validation_map.values() {
```
jdm:
```rust
valid &= validation_map.values().all(|stage| {
    stage.max_uniform_buffers_per_shader_stage >= 0 &&
    ...
});
```

components/script/dom/gpudevice.rs
```diff
@@ -229,4 +236,181 @@ impl GPUDeviceMethods for GPUDevice {
...
+                each.max_sampled_textures_per_shader_stage < 0 ||
+                each.max_storage_textures_per_shader_stage < 0 ||
+                each.max_samplers_per_shader_stage < 0
+            {
+                valid &= false;
+                break;
+            }
+        }
+
+        // DynamicValues does not exceeded the max limit for the pipeline
```
jdm: `does not exceed`

components/script/dom/webidls/GPUBindGroupLayout.webidl
```diff
@@ -0,0 +1,34 @@
...
+[Exposed=(Window, DedicatedWorker), Serializable, Pref="dom.webgpu.enabled"]
+interface GPUBindGroupLayout {
+};
+GPUBindGroupLayout includes GPUObjectBase;
+
+dictionary GPUBindGroupLayoutDescriptor : GPUObjectDescriptorBase {
+    required sequence<GPUBindGroupLayoutBindings> bindings;
+};
+
+// Note: Servo codegen doesn't like the name `GPUBindGroupLayoutBinding` because it's already occupied
```
jdm: What's the full error message? What's the source of the duplicate name?

jdm (changes requested):
Looking good!

components/script/dom/webidls/GPUBindGroupLayout.webidl
imiklos: When the codegen generates the internal `Binding` from the WebIDL for the `GPUBindGroupLayout` interface, the internal name for this bind will be `GPUBindGroupLayoutBinding` and when the codegen tries to generate the `GPUBindGroupLayoutBinding` dictionary we got a duplicate name.
The error message: `GPUBindGroupLayoutBinding` must be defined only once in the type namespace of this module.

components/script/dom/gpudevice.rs
imiklos: You are right! Thank you!

imiklos:
@jdm Thank you! I've addressed the changes you mentioned and updated the pull request.

jdm:
I have verified that there is no way for the WebIDL dictionary name to be exposed to content, so renaming it seems like a fine solution.
@bors-servo r+

jdm:
@bors-servo retry
* #24726

Issue #25478, opened 2020-01-10
Implement window "current event"


pshaughn:
Lazily-written event listeners can be written by page authors as Javascript functions taking no arguments, trusting that the global "event" will be the right thing when they are called. https://dom.spec.whatwg.org/#interface-window-extensions is what makes this possible.
This might be easier to implement after I finish doing some cleanup for #22783; I'll be sure to leave comments for the relevant points.

pshaughn:
I think this just needs to add the field to the Window struct and IDL and implement the steps in https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke. Note that nothing exits this algorithm between step 8 and step 12, so we'll always end up putting the current event back to where it was when we came in. Multiple inner invokes can nest at step 10, which is why we have to set it back to its original value instead of null.

PR #25534, opened 2020-01-15
Put target element, form owner, and element document on scope chain of compiled events

### LINKED ISSUE #25517: Event handlers compiled from inline attributes aren't seeing properties of their current target as variables
After #25488, the first subtest in html/webappapis/scripting/events/compile-event-handler-lexical-scopes.html does what it's supposed to do far enough to find an actual problem. A more minimal case to focus on the one problem is:

```
    <div id="log"></div>
    <input value="4" onclick="document.getElementById('log').innerText+=value">
    <script>
      document.querySelector("input").click()
    </script>
```

In Firefox, this puts 4 in the div. In Servo, this logs nothing and reports an exception that value is not defined.

I need to look at specifications more to figure out what mechanism should be exposing the properties in this way and what part of it Servo is missing; prior to this test, I was unaware this behavior existed at all.

pshaughn:
Event listeners that are created from a function object just get whatever closure the function object had, but event listeners created from a string need a special closure that acts like they were defined inside a series of `with` statements. This now happens. The existing WPT test for it, html/webappapis/scripting/events/compile-event-handler-lexical-scopes.html, also relies on other behavior we don't have, so I added an easier version of the test that doesn't involve bubbling or capturing and doesn't check any IDL properties we don't have. This new test will eventually be redundant when we have everything else the upstream test expects.

---

- [X] These changes fix #25517

### DIFF
--- a/components/script/dom/eventtarget.rs
+++ b/components/script/dom/eventtarget.rs
@@ -30,6 +30,7 @@ use crate::dom::element::Element;
 use crate::dom::errorevent::ErrorEvent;
 use crate::dom::event::{Event, EventBubbles, EventCancelable, EventStatus};
 use crate::dom::globalscope::GlobalScope;
+use crate::dom::htmlformelement::FormControlElementHelpers;
 use crate::dom::node::document_from_node;
 use crate::dom::virtualmethods::VirtualMethods;
 use crate::dom::window::Window;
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo r+
Thanks for fixing this! I'm hopeful this will help our web compat story quite a bit :)

pshaughn:
I think #25478 is also needed before this will have much of a web compatibility improvement. Handlers that are just a blob of javascript with no function arguments usually want to see the event, not just the target.

pshaughn:
```  ▶ Unexpected subtest result in /WebIDL/current-realm.html:
  │ FAIL [expected PASS] createProcessingInstruction
  │   → global is undefined
  │ 
  │ isObjectFromGlobal@http://web-platform.test:8000/WebIDL/current-realm.html:16:4```

This is strange. This problem certainly relates to the scope we see inside an event handler, but I don't see any way for this code to create InternalRawUncompiledHandler and be affected by my changes, and I don't see how self could exist for the first few asserts and then be missing on createProcessingInstruction.

pshaughn:
My wild guess here is that the "cannot allocate memory" problem also caused this error; it is not reproducing for me and it didn't happen on the Mac tests.

jdm:
@bors-servo retry
I think this may just be an instance of an intermittent failure that we haven't seen before; I agree that it doesn't look like the test should be affected by the changed code.

PR #25545, opened 2020-01-17
Add WebGL2 specific version strings

mmatyas:
Updated the `VERSION` and `SHADING_LANGUAGE_VERSION` WebGL parameters to report the correct version strings when WebGL2 is in use.

See: https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.2

cc @jdm @zakorgy @imiklos 

---


jdm:
@bors-servo r+
Thanks!

Issue #25145, opened 2019-12-06
Document named getter can't find embeds and objects


pshaughn:
WPT html/dom/documents/dom-tree-accessors/nameditem-05.html and -07.html; the only passing cases are the ones that aren't supposed to find anything.

pshaughn:
-08.html's one failed case is also a logical consequence of this, since an object can't take priority over a div if the object isn't found in the first place.

pshaughn:
-02.html is iframes, likewise only passing cases where it's supposed to not find the iframe.

pshaughn:
#25548 covers iframes; I didn't put embeds and objects in there because the spec for getting embeds and objects involves testing for the "displaying fallback content" state, which doesn't exist for us

pshaughn:
Iframes work; embeds and objects can be made to work the same way, once they have nested browsing contexts to be finding.

PR #25548, opened 2020-01-17
Add SupportedPropertyNames to Document (also fix iframe getting)

### LINKED ISSUE #7273: Implement SupportedPropertyNames methods
https://github.com/servo/servo/pull/7254 introduces a method on some DOM types called `SupportedPropertyNames` that returns all the ['supported property names'](https://heycam.github.io/webidl/#idl-named-properties) as per their respective interface. That PR only implements `HTMLCollectionMethods::SupportedPropertyNames` (I didn't have enough time to complete the rest). The other ones need to be implemented.

### LINKED ISSUE #25146: Document named getter is returning something when two imgs have the same ID, but it shouldn't according to WPT and some browsers
The test4 case in WPT html/dom/documents/dom-tree-accessors/nameditem-06.html says this should return undefined; Safari is also failing this case.

pshaughn:
Existing test of named-getting an iframe now succeeds. I added a new test for Object.getOwnPropertyNames(document) based on my understanding of the spec; that test could use a second opinion.

UPDATE: This was trying to do too many things in one PR as originally submitted. It is now using #25572 as a base, and I suggest reviewing that PR before this one to avoid duplicating review effort.

---

- [X] These changes fix #7273 for all implemented named getters, fix #25146, and fix the iframe case only of #25145.


pshaughn:
This is broken right now and needs #25570 fixed as a foundation so element names don't sneak out of the atom cache.

pshaughn:
Now using #25572 as a base, which should prevent crashes. I also took into account review comments by @Manishearth from #25562 (which I plan on updating to use this as a base, if this seems solid enough.)
@bors-servo try

pshaughn:
(forgot to account for comment https://github.com/servo/servo/pull/25562#discussion_r369275246, so as of this push the new supported property names test is still broken out into a bunch of really small cases)

jdm (approved):
Great work! I found this to be quite readable, which is better than the attempt that I made at it originally!

jdm:
This can merge after a rebase and #25572 is merged.

jdm:
@bors-servo retry
* #25513

Issue #25331, opened 2019-12-18
Fix WPT test failures induced by switching 2D canvas rendering backend to Raqote


pylbrecht:
#24470 left over some failing tests, which need to be addressed.

Relevant tests: `tests/wpt/web-platform-tests/2dcontext/`

pylbrecht:
Tests using `stroke()` may require a fix on raqote's side. (https://github.com/servo/servo/pull/24470#issuecomment-560332127, https://github.com/servo/servo/pull/24470#issuecomment-561436893)

pylbrecht:
Just a quick update to show that I'm alive: I've been on vacation for the past week and will continue working on this by Monday. My fiancee is going on vacation for two weeks then, so I got some extra time on my hands!

pylbrecht:
@jrmuizel, any news on the [`stroke()` issue](https://github.com/servo/servo/pull/24470#issuecomment-560332127)?

jrmuizel:
@pylbrecht nope, I haven't had a chance to complete my research on the issue. The problem is tracked upstream here: https://github.com/jrmuizel/raqote/issues/96

pylbrecht:
Unfortunately university deadlines combined with frustrating hours in the lab ate up all the extra time I planned on using for contributing this week. I hope the debugging gods show mercy tomorrow, so I can fix some wpt tests next week.

pylbrecht:
While inspecting https://dxr.mozilla.org/servo/source/tests/wpt/web-platform-tests/2dcontext/path-objects/2d.path.rect.end.1.html, I came across something, I don't quite understand.
@jrmuizel, can you shed some light on why the image created with raqote looks like this?

expected: https://codepen.io/pylbrecht/pen/RwNJyWE

actual:
![example](https://user-images.githubusercontent.com/11317687/72283998-43b58d80-3640-11ea-8e2e-6e1b896e887a.png)
```rust
let mut dt = DrawTarget::new(300, 300);
let mut pb = PathBuilder::new();
pb.rect(0., 0., 100., 50.);
dt.fill(
    &pb.finish(),
    &Source::Solid(SolidSource::from_unpremultiplied_argb(255, 255, 0, 0)),
    &DrawOptions::new(),
    );
                                                                            
let mut pb = PathBuilder::new();
pb.rect(200., 100., 400., 1000.);
pb.line_to(-2000., -1000.);
dt.stroke(
    &pb.finish(),
[...]

PR #25551, opened 2020-01-19
Fix canvas wpt test `path-objects/2d.path.rect.zero.4.html`

pylbrecht:
After connecting the lines of a rectangle, a new subpath has to be created containing the rectangle's origin as the only point in the subpath.
(see step 4 of https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-rect)

---


jdm:
Good find!
@bors-servo r+

jdm:
@bors-servo retry
* #24762 #24622

Issue #25522, opened 2020-01-14
Intermittent timeouts in linux WPT media tests


jdm:
```
1 unexpected results that are NOT known-intermittents:
  â–¶ TIMEOUT [expected OK] /html/semantics/embedded-content/media-elements/loading-the-media-resource/autoplay-overrides-preload.html
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ 0:00:00.791281716 [332m18379[00m 0x7fbe9c0398f0 [32;01mFIXME  [00m [00;04m             default gstutils.c:3981:gst_pad_create_stream_id_internal:<appsrc0:src>[00m Creating random stream-id, consider implementing a deterministic way of creating a stream-id
  â”‚ 0:00:00.797305836 [332m18379[00m 0x55b744d66680 [32;01mFIXME  [00m [00;01m                 bin gstbin.c:4338:gst_bin_query:[00m implement duration caching in GstBin again
  â”‚ 0:00:00.802688243 [332m18379[00m 0x7fbe9402c4f0 [32;01mFIXME  [00m [00;04m             default gstutils.c:3981:gst_pad_create_stream_id_internal:<typefind:src>[00m Creating random stream-id, consider implementing a deterministic way of creating a stream-id
  â”‚ 0:00:00.804008803 [332m18379[00m 0x7fbe80003590 [32;01mFIXME  [00m [00;04m             default gstutils.c:3981:gst_pad_create_stream_id_internal:<appsrc1:src>[00m Creating random stream-id, consider implementing a deterministic way of creating a stream-id
  â”‚ 0:00:00.809664161 [332m18379[00m 0x55b744d5d4f0 [32;01mFIXME  [00m [00;04m             default gstutils.c:3981:gst_pad_create_stream_id_internal:<typefind:src>[00m Creating random stream-id, consider implementing a deterministic way of creating a stream-id
  â”‚ 0:00:00.825871998 [332m18379[00m 0x7fbe7413bad0 [32;01mFIXME  [00m [00;04m             default gstutils.c:3981:gst_pad_create_stream_id_internal:<mpegaudioparse0:src>[00m Creating random stream-id, consider implementing a deterministic way of creating a stream-id
[...]

PR #25552, opened 2020-01-19
Return the highest priority error from the descendant instead of return the very first one

### LINKED ISSUE #25436: Investigate failure of /html/semantics/scripting-1/the-script-element/module/error-type-1.html after module script landed
Before the IndexSet fix, we will get this failure intermittently. But, I can reproduce this failure consistently.

For merging #23545 first, marked it as FAIL in that PR and fix it separately.

```
  ▶ Unexpected subtest result in /html/semantics/scripting-1/the-script-element/module/error-type-1.html:
  │ FAIL [expected PASS] network error has higher priority than parse error
  │   → assert_equals: expected 3 but got 4
  │ 
  │ @http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/module/error-type-1.html:16:7
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1984:25
  └ Test.prototype.step_func_done/<@http://web-platform.test:8000/resources/testharness.js:2025:32

```

_Originally posted by @Manishearth in https://github.com/servo/servo/pull/23545#issuecomment-571051685_

CYBAI:
The test failed because we didn't return the highest priority error (which is network error in this case).

As Manish mentioned in https://github.com/servo/servo/issues/25436#issuecomment-571065323, that's because we're using the Promise.all trick to signal loads.

If we can avoid relying on Promise.all, maybe we don't need to do a complex logic like this; instead, ideally, we should always finish the module load immediately when we hit network failure so that we don't even need to do the `max()` comparison.

---
- [x] These changes fix #25436 

### DIFF
--- a/components/script/script_module.rs
+++ b/components/script/script_module.rs
@@ -633,6 +633,7 @@ impl ModuleTree {
         }
 
         // 5-6.
+        let mut errors: Vec<ModuleError> = Vec::new();
         let descendant_urls = module_tree.get_descendant_urls().borrow();
 
         for descendant_module in descendant_urls
@@ -650,13 +651,13 @@ impl ModuleTree {
                 ModuleTree::find_first_parse_error(&global, &descendant_module, discovered_urls);
 
             // 8-4.
-            if child_parse_error.is_some() {
-                return child_parse_error;
+            if let Some(child_error) = child_parse_error {
+                errors.push(child_error);
             }
         }
 
         // Step 9.
-        return None;
+        return errors.into_iter().max();
     }
 }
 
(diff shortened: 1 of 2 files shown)


CYBAI:
The failure is #25522

components/script/script_module.rs
```diff
@@ -651,12 +652,12 @@ impl ModuleTree {
 
             // 8-4.
             if child_parse_error.is_some() {
-                return child_parse_error;
+                errors.push(child_parse_error.unwrap());
```
jdm: Let's use `if let Some` instead of is_some and unwrap.

components/script/script_module.rs
```diff
@@ -651,12 +652,12 @@ impl ModuleTree {
...
             }
         }
 
         // Step 9.
-        return None;
+        return errors.iter().max().map(|err| err.clone());
```
jdm: We should be able to do `errors.into_iter().max()` instead.

jdm (changes requested):
Looks sensible!

components/script/script_module.rs
CYBAI: Thanks for the hint!!! I finally understand the difference between `into_iter()` and `iter()` 😍 That's what I want!

jdm:
@bors-servo retry
* #25558

PR #25555, opened 2020-01-19
Add default arguments so that ./mach test --all works

### LINKED ISSUE #25554: `./mach test --all` passes incorrect number of arguments to test-tidy
On Ubuntu 19.10, `./mach test --all` fails on the current master branch with:

```
Error running mach:

    ['test', '--all']

message.

TypeError: test_tidy() takes exactly 7 arguments (5 given)

  File "/home/warren/programming/servo/python/servo/testing_commands.py", line 176, in test
    Registrar.dispatch("test-%s" % suite, context=self.context, **kwargs)
  File "/home/warren/programming/servo/python/_virtualenv2.7/lib/python2.7/site-packages/mach/registrar.py", line 152, in dispatch
    return self._run_command_handler(handler, context=context, **kwargs)
  File "/home/warren/programming/servo/python/_virtualenv2.7/lib/python2.7/site-packages/mach/registrar.py", line 109, in _run_command_handler
    result = fn(**kwargs)
```

#16071 seems related. The main file in question is [python/servo/testing_commands.py](https://github.com/servo/servo/blob/master/python/servo/testing_commands.py). 

Running `./mach test-tidy --self-test` [succeeds](https://gist.github.com/warren-fisher/7d52f61574a73d959e810676225fdc76#file-gistfile1-txt) with all 29 tests passing.

Running `./mach test --all --self-test` or `./mach test tidy --self-test` fail with the same error as `./mach test --all`. Perhaps --self-test was never meant to work with either of these options, I am not sure.

warren-fisher:
Fix `./mach test --all` not getting any default values for `force_cpp` and `no_wpt` in [python/servo/testing_commands.py](https://github.com/servo/servo/blob/2a594821ba44ba2c13e9a39c6fd8c5a450bd06f4/python/servo/testing_commands.py#L319-L334). For `./mach test-tidy` the default values are obtained using the `@CommandArgument` decorator, but this does not apply for calling `test_tidy()` when using `./mach test --all` (which calls the function directly). This call is found on line 109 of `registrar.py` which should be located in something like `python/_virtualenv2.7/python2.7/site-packages/mach/registrar.py` (but not located in this git repository). 

---

- [x] These changes fix #25554 (GitHub issue number if applicable)

- [x] These changes do not require tests because the change is minimal and non breaking. Any existing code should be able to work if it uses keyword arguments because the default values will be overriden. If there is code that calls `test_tidy()` with arguments by position this would cause breakage, but I do not believe this happens anywhere in the `testing_command.py` file. `./mach test-tidy` should still work because the arguments are set using the `@CommandArgument` decorator.

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -331,7 +331,7 @@ class MachCommands(CommandBase):
     @CommandArgument('--stylo', default=False, action="store_true",
                      help="Only handle files in the stylo tree")
     @CommandArgument('--force-cpp', default=False, action="store_true", help="Force CPP check")
-    def test_tidy(self, all_files, no_wpt, no_progress, self_test, stylo, force_cpp):
+    def test_tidy(self, all_files, no_progress, self_test, stylo, force_cpp=False, no_wpt=False):
         if self_test:
             return test_tidy.do_tests()
         else:


jdm:
@bors-servo r+
Thanks for the patch!

jdm:
@bors-servo retry
* #24726 #25513 #24762

jdm:
@bors-servo retry
* #24726

PR #25560, opened 2020-01-20
Add initial scrolling support to layout_2020

mrobinson:
This adds very early scrolling support to layout_2020. Only the root element scrolls for now, but the idea is to continue to polish this.

---

### DIFF
--- a/components/layout_2020/flow/root.rs
+++ b/components/layout_2020/flow/root.rs
@@ -11,12 +11,14 @@ use crate::formatting_contexts::IndependentFormattingContext;
 use crate::fragments::Fragment;
 use crate::geom;
 use crate::geom::flow_relative::Vec2;
+use crate::geom::physical;
 use crate::positioned::AbsolutelyPositionedBox;
 use crate::positioned::PositioningContext;
 use crate::replaced::ReplacedContent;
 use crate::sizing::ContentSizesRequest;
 use crate::style_ext::{Display, DisplayGeneratingBox, DisplayInside};
 use crate::DefiniteContainingBlock;
+use euclid::Size2D;
 use gfx_traits::print_tree::PrintTree;
 use script_layout_interface::wrapper_traits::LayoutNode;
 use servo_arc::Arc;
@@ -26,7 +28,14 @@ use style::Zero;
 use style_traits::CSSPixel;
 
 pub struct BoxTreeRoot(BlockFormattingContext);
-pub struct FragmentTreeRoot(Vec<Fragment>);
+
+pub struct FragmentTreeRoot {
+    /// The children of the root of the fragment tree.
+    children: Vec<Fragment>,
+
+    /// The scrollable overflow of the root of the fragment tree.
+    scrollable_overflow: physical::Rect<Length>,
+}
 
 impl BoxTreeRoot {
     pub fn construct<'dom, Node>(context: &LayoutContext, root_element: Node) -> Self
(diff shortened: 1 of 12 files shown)


components/layout_2020/geom.rs
```diff
@@ -387,6 +396,33 @@ impl<T> physical::Rect<T> {
     }
 }
 
+impl physical::Rect<Length> {
+    pub fn union(&self, other: &Self) -> Self {
```
SimonSapin: The union of two rectangle is a polygon that is (usually) not a rectangle. This returns the axis-aligned bounding box of that union. Is `aabb` commonly used as an abbreviation in graphics code? How do you feel about renaming? If not, let’s add a doc-comment

components/layout_2020/flow/root.rs
```diff
@@ -151,15 +197,35 @@ impl FragmentTreeRoot {
...
+        for fragment in &self.children {
             fragment.print(&mut print_tree);
         }
     }
+
+    pub fn scrollable_overflow(&self) -> webrender_api::units::LayoutSize {
+        let overflow = &self.scrollable_overflow;
+        webrender_api::units::LayoutSize::from_untyped(Size2D::new(
+            overflow.size.x.px() - overflow.top_left.x.px(),
+            overflow.size.y.px() - overflow.top_left.y.px(),
```
SimonSapin: Shouldn’t these `-` be `+`? Is this computing the coordinates of the bottom right corner?

components/layout_2020/fragments.rs
```diff
@@ -90,6 +97,19 @@ impl Fragment {
...
+    pub fn scrollable_overflow(&self) -> physical::Rect<Length> {
+        match self {
+            Fragment::Box(fragment) => fragment.scrollable_overflow.clone(),
+            Fragment::Anonymous(fragment) => fragment.scrollable_overflow.clone(),
+            Fragment::Text(fragment) => fragment
+                .rect
+                .to_physical(fragment.parent_style.writing_mode, &physical::Rect::zero()),
+            Fragment::Image(fragment) => fragment
+                .rect
+                .to_physical(fragment.style.writing_mode, &physical::Rect::zero()),
```
SimonSapin: Please add a FIXME comment to these `Rect::zero` calls, and file an issue with how to grep for such comments.

components/layout_2020/geom.rs
mrobinson: I was using the nomenclature of euclid, but I'm happy to use `axis_aligned_bounding_box` to describe this concept.

mrobinson (commented):
Thanks for the review! I've uploaded a new version of the branch with the changes you've suggested.

components/layout_2020/flow/root.rs
mrobinson: Sorry. Here I mean that we are trying to return the overflow size clipped to the block-start and inline-start edges as specified. Currently this is interpreted as clipping anything to the left and top of the origin. in the latest version of the branch, I've added a comment here explaining the situation better.

mrobinson: You're right, this should be an addition here instead of a subtraction. I've fixed this in the latest version of the branch. Thanks!

SimonSapin:
It looks like GitHub doesn’t send a notification for force-pushes. When doing that please comment in the PR or ping on IRC to let me know there’s something new to look at :)

SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #25561, opened 2020-01-21
Make File objects roundtrip through FormData

### LINKED ISSUE #24939: Retrieving a File from FormData isn't getting back the same object
WPT xhr/formdata-foreach.html puts a File object into a FormData, iterates the FormData, and gets a File object back at the right point in the iteration, but the File object it gets back is not Javascript-equal-to the one it put in, failing the test.

pshaughn:
FormData.append now only creates a new File object if it needs to file-wrap a blob or set a filename, and not when appending an existing File with no filename override. The hardest part here was reading the specification in a way that matched browser behavior.

---

- [X] These changes fix #24939


jdm:
@bors-servo r+
Good catch!

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo  retry
* #25513

jdm:
@bors-servo retry
* #25513

jdm:
@bors-servo retry
* #25513

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726 #13480

PR #25569, opened 2020-01-21
Initial implementation of GPUPipelineLayout for WebGPU

imiklos:
Added WebIDL bindings for `GPUPipelineLayout`.
Implemented the createPipelineLayout function of `GPUDevice`.

---

cc @kvark @jdm @zakorgy

### DIFF
--- a/components/script/dom/gpubindgrouplayout.rs
+++ b/components/script/dom/gpubindgrouplayout.rs
@@ -42,6 +42,7 @@ impl GPUBindGroupLayout {
             valid: Cell::new(valid),
         }
     }
+
     pub fn new(
         global: &GlobalScope,
         channel: WebGPU,
@@ -59,6 +60,20 @@ impl GPUBindGroupLayout {
             global,
             GPUBindGroupLayoutBinding::Wrap,
         )
+    }
+}
+
+impl GPUBindGroupLayout {
+    pub fn is_valid(&self) -> bool {
+        self.valid.get()
+    }
+
+    pub fn id(&self) -> WebGPUBindGroupLayout {
+        self.bind_group_layout
+    }
+
+    pub fn bindings(&self) -> &[GPUBindGroupLayoutBindings] {
+        &self.bindings
     }
 }
 
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -11,6 +11,7 @@ use crate::dom::bindings::codegen::Bindings::GPUBindGroupLayoutBinding::{
 };
 use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUBufferDescriptor;
 use crate::dom::bindings::codegen::Bindings::GPUDeviceBinding::{self, GPUDeviceMethods};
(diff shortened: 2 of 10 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -410,4 +412,70 @@ impl GPUDeviceMethods for GPUDevice {
...
+                        if bind.hasDynamicOffset {
+                            max_dynamic_storage_buffers_per_pipeline_layout -= 1;
+                        };
+                    },
+                    _ => {},
+                };
+            });
+        });
+
+        valid &= descriptor.bindGroupLayouts.len() <= limits.maxBindGroups as usize &&
```
jdm: It looks like we can declare `let valid =` here, rather than declaring it as mutable earlier.

components/script/dom/gpupipelinelayout.rs
```diff
@@ -0,0 +1,68 @@
...
+    ) -> GPUPipelineLayout {
+        Self {
+            reflector_: Reflector::new(),
+            bind_group_layouts,
+            label: DomRefCell::new(None),
+            pipeline_layout,
+            valid: Cell::new(valid),
+        }
+    }
+    pub fn new(
```
jdm: nit: add a newline between these two methods.

components/webgpu/lib.rs
```diff
@@ -242,6 +248,22 @@ impl WGPU {
...
+                    let descriptor = wgpu_core::binding_model::PipelineLayoutDescriptor {
+                        bind_group_layouts: bind_group_layouts.as_ptr(),
+                        bind_group_layouts_length: bind_group_layouts.len(),
+                    };
+                    let pl_id = gfx_select!(id => global.device_create_pipeline_layout(device.0, &descriptor, id));
+                    let pipeline_layout = WebGPUPipelineLayout(pl_id);
+
+                    if let Err(e) = sender.send(pipeline_layout) {
+                        warn!(
+                            "Failed to send response to WebGPURequest::CreateBufferMapped ({})",
```
jdm: Wrong method name in this error.

components/script/dom/gpubindgrouplayout.rs
```diff
@@ -62,6 +62,20 @@ impl GPUBindGroupLayout {
...
+impl GPUBindGroupLayout {
+    pub fn is_valid(&self) -> bool {
+        self.valid.get()
+    }
+
+    pub fn id(&self) -> WebGPUBindGroupLayout {
+        self.bind_group_layout
+    }
+
+    pub fn bindings(&self) -> &Vec<GPUBindGroupLayoutBindings> {
```
jdm: Let's use `&[GPUBindGroupLayoutBindings]` instead.

jdm (changes requested):
Looks fine, generally!

components/webgpu/lib.rs
imiklos: Ooops.

imiklos:
@jdm Thank you for the review! I've addressed the changes you mentioned.

Issue #25057, opened 2019-12-03
An iframe element's name and its contained child browsing context's name interact improperly


pshaughn:
This is the iframe case of html/browsers/windows/nested-browsing-contexts/name-attribute.window.html (#25056 for other cases). iframe.setAttribute("name","meh") causes code in the iframe to see window.name as "meh". WPT expects the name that the iframe originally had to be visible as window,name, but for subsequent iframe.setAttributes by the parent to not change window.name. Firefox fails this in the same way; other browsers pass.

pshaughn:
In an unmerged branch where the window named getter is generally able to return browsing contexts (based on #21869 but with the iframe TODOs filled in), html/browsers/windows/the-window-object/window-named-properties.html fails cases "Dynamic name" and "Ghost name" and it seems like another manifestation of the same cause as the above problem.

pshaughn:
Despite the test name, html/browsers/the-window-object/named-access-on-the-window-object/window-null-names.html is actually failing for reasons relating to this. Changing it to use another character instead of the null-byte doesn't make it pass; the problem is about the iframe and its child browsing context not having the right name interaction, not about the null byte.

pshaughn:
I'm going to have to put a subset of html/browsers/windows/nested-browsing-contexts/name-attribute.window.html into the _mozilla tests so we can observe it passing or failing; in an automated run, the test harness times out on the embed and object cases so we don't get to see what happens to the iframe cases.

pshaughn:
#25572 fixed this.

PR #25572, opened 2020-01-22
Make name content attributes consistently atoms and put them in rare_data for fast access

### LINKED ISSUE #25570: Handle element names consistently
Some SetName methods are make_setter!, some are make_atomic_setter!, and HTMLIFrameElement has its own DomRefCell<DOMString> reflecting its name. This inconsistency is a contributing factor in #25057, and cleaning it up is necessary to make a #25548 that won't crash.

I had changed the Element parent class to make names atoms in #25548, reflecting them in rare_data for lookup speed, but I had neglected to consider the differences between different Element subclasses there, and now that I look at all the specific ways different elements use names, a sweep to unify them all seems to be in order.

pshaughn:
All codepaths setting the name content attribute now use an atom, which is also stored in rare_data for direct lookup by a get_name method. 

Paralleling the get_name method, I added a get_id method, which makes some internal id-lookup cases nicer.

A new test tests for a name setter on every HTML element type. In addition to its overt and upstreamable purpose of checking IDL property reflection semantics, for us this test also hits some Servo assertions that make sure the name is an atom in every case. If the test doesn't crash, even a failed test case still has the attribute as an atom rather than some other type. The failed cases are for elements that we have unimplemented or completely stubbed; I added a few missing name IDL properties to otherwise implemented elements.

---

- [X] These changes fix #25570 and make progress on #25057

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -1401,11 +1401,28 @@ impl Element {
     // https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name
     pub fn get_attribute_by_name(&self, name: DOMString) -> Option<DomRoot<Attr>> {
         let name = &self.parsed_name(name);
-        self.attrs
+        let maybe_attribute = self
+            .attrs
             .borrow()
             .iter()
             .find(|a| a.name() == name)
-            .map(|js| DomRoot::from_ref(&**js))
+            .map(|js| DomRoot::from_ref(&**js));
+
+        fn id_and_name_must_be_atoms(name: &LocalName, maybe_attr: &Option<DomRoot<Attr>>) -> bool {
+            if *name == local_name!("id") || *name == local_name!("name") {
+                match maybe_attr {
+                    None => true,
+                    Some(ref attr) => match *attr.value() {
+                        AttrValue::Atom(_) => true,
+                        _ => false,
+                    },
+                }
+            } else {
+                true
+            }
+        }
+        debug_assert!(id_and_name_must_be_atoms(name, &maybe_attribute));
+        maybe_attribute
     }
 
     pub fn set_attribute_from_parser(
@@ -1800,6 +1817,14 @@ impl Element {
         let other = other.upcast::<Element>();
         self.root_element() == other.root_element()
     }
+
+    pub fn get_id(&self) -> Option<Atom> {
+        self.id_attribute.borrow().clone()
+    }
+
+    pub fn get_name(&self) -> Option<Atom> {
+        self.rare_data().as_ref()?.name_attribute.clone()
+    }
 }
 
 impl ElementMethods for Element {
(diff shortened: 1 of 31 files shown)


pshaughn:
I am sure this will have some idlharness test result changes, and I wouldn't be surprised if filling in the name setter has had other effects.
@bors-servo try=wpt

pshaughn:
That one /css/css-transforms/animation/perspective-interpolation.html crash looks scary. Other new failures are probably because we were only "accidentally" passing before, but of course I'll have to look through them all to be sure.

pshaughn:
Oh, I was reading upwards from the bottom and missed the known-intermittents header, that one test already has an intermittency issue filed for it. Now to look at all the real new failures...

pshaughn:
Once I was reading the right part of the results, it was all new passes and completely consistent with what I changed. I notice the CI test harness does a better job of extracting results from a timed-out test than I get running on the command line; the _mozilla test I added is not actually necessary for CI purposes.

pshaughn:
I believe this is ready to roll out.
@bors-servo try

jdm:
That's #23290.

components/script/dom/element.rs
```diff
@@ -1388,11 +1388,27 @@ impl Element {
...
         let name = &self.parsed_name(name);
-        self.attrs
+        let maybe_attribute = self
+            .attrs
             .borrow()
             .iter()
             .find(|a| a.name() == name)
-            .map(|js| DomRoot::from_ref(&**js))
+            .map(|js| DomRoot::from_ref(&**js));
+        fn id_and_name_must_be_atoms(name: &LocalName, maybe_attr: &Option<DomRoot<Attr>>) -> bool {
```
jdm: We probably need a `#[cfg(debug_assertions)]` to avoid warnings about an unused function in builds without them enabled.

jdm (approved):
Thanks for doing this work! I'm impressed by how many places ended up needing to be modified.

jdm:
Sorry, I was wrong about how the debug_assert macro is treated by the compiler:
```
error[E0425]: cannot find function `id_and_name_must_be_atoms` in this scope
    --> components/script/dom/element.rs:1425:23
     |
1425 |         debug_assert!(id_and_name_must_be_atoms(name, &maybe_attribute));
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```

jdm:
This needs a `mach fmt`.

pshaughn:
The spec sometimes explicitly excludes the empty-string from being a possible name and sometimes doesn't; there aren't WPT tests on the subject so I don't know what browsers do in various name="" cases, and empty-string handling here might not be great.

pshaughn:
I don't know what that means but it sounds like something a rebase might fix.

jdm:
@bors-servo r+
Yeah, those failures are mysterious :(

PR #25574, opened 2020-01-22
Use Euclid data types for physical structs in layout_2020

mrobinson:
This removes a bit of duplication and allows layout_2020 to benefit from a
much richer set of utilities.

---

### DIFF
--- a/components/layout_2020/geom.rs
+++ b/components/layout_2020/geom.rs
@@ -12,31 +12,10 @@ use style::values::generics::length::MaxSize;
 use style::Zero;
 use style_traits::CSSPixel;
 
-pub type Point<U> = euclid::Point2D<f32, U>;
-pub type Size<U> = euclid::Size2D<f32, U>;
-pub type Rect<U> = euclid::Rect<f32, U>;
-
-pub(crate) mod physical {
-    #[derive(Clone)]
-    pub(crate) struct Vec2<T> {
-        pub x: T,
-        pub y: T,
-    }
-
-    #[derive(Clone, Debug)]
-    pub(crate) struct Rect<T> {
-        pub top_left: Vec2<T>,
-        pub size: Vec2<T>,
-    }
-
-    #[derive(Clone, Debug)]
-    pub(crate) struct Sides<T> {
-        pub top: T,
-        pub left: T,
-        pub bottom: T,
-        pub right: T,
-    }
-}
+pub type PhysicalPoint<U> = euclid::Point2D<U, CSSPixel>;
+pub type PhysicalSize<U> = euclid::Size2D<U, CSSPixel>;
+pub type PhysicalRect<U> = euclid::Rect<U, CSSPixel>;
+pub type PhysicalSides<U> = euclid::SideOffsets2D<U, CSSPixel>;
 
 pub(crate) mod flow_relative {
     #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


components/layout_2020/geom.rs
```diff
@@ -222,42 +167,33 @@ impl fmt::Debug for flow_relative::Rect<Length> {
     }
 }
 
-impl<T: Clone> flow_relative::Vec2<T> {
-    pub fn size_to_physical(&self, mode: WritingMode) -> physical::Vec2<T> {
+impl<T: Clone + Copy> flow_relative::Vec2<T> {
```
SimonSapin: Why the `Copy` bound? Please either remove it, or remove the `Clone` bound and the calls to `.clone()` and the `&` expressions taking references to fields of `self`.

components/layout_2020/geom.rs
```diff
@@ -364,79 +300,45 @@ impl<T> flow_relative::Rect<T> {
         // Will be needed for other writing modes
         // FIXME: what if the containing block has a different mode?
         // https://drafts.csswg.org/css-writing-modes/#orthogonal-flows
-        _containing_block: &physical::Rect<T>,
-    ) -> physical::Rect<T>
+        _containing_block: &PhysicalRect<T>,
+    ) -> PhysicalRect<T>
     where
-        T: Clone,
+        T: Clone + Copy,
```
SimonSapin: Same here re `Copy` v.s. `Clone`

components/layout_2020/geom.rs
```diff
@@ -364,79 +300,45 @@ impl<T> flow_relative::Rect<T> {
...
-        let bottom_corner_x = (self.top_left.x + self.size.x).max(other.top_left.x + other.size.x);
-        let bottom_corner_y = (self.top_left.y + self.size.y).max(other.top_left.y + other.size.y);
-        let size = physical::Vec2 {
-            x: bottom_corner_x - top_left.x,
-            y: bottom_corner_y - top_left.y,
-        };
-
-        Self { top_left, size }
-    }
+pub trait ToWebRender {
```
SimonSapin: Why not use the standard library `From` or `Into` trait?

components/layout_2020/replaced.rs
```diff
@@ -71,11 +71,8 @@ impl ReplacedContent {
     }
 
     fn flow_relative_intrinsic_size(&self, style: &ComputedValues) -> Vec2<Option<Length>> {
-        let intrinsic_size = physical::Vec2 {
-            x: self.intrinsic.width,
-            y: self.intrinsic.height,
-        };
-        intrinsic_size.size_to_flow_relative(style.writing_mode)
+        let intrinsic_size = PhysicalSize::new(self.intrinsic.width, self.intrinsic.height);
```
SimonSapin: Passing two arguments of the same type where order is significant instead of writing out named fields in a struct literal seems like a regression, but oh well.

components/layout_2020/geom.rs
mrobinson: Nice catch. Not sure how the `Copy` slipped in. I've fixed this.

components/layout_2020/geom.rs
mrobinson: I've fixed this as well.

components/layout_2020/geom.rs
mrobinson: The issue here is that both of the data structures are defined in separate crates. I believe that Into requires at least one type to be defined in the current crate. Perhaps I'm wrong though...

components/layout_2020/replaced.rs
mrobinson: I totally agree. This seems to be one of the sharper edges of euclid and I wish there was a way to provide named arguments or to instantiate the struct right here. It seems that it is impossible though due to it having private fields.

mrobinson (commented):
Thanks for the review. I've updated the branch with a new version addressing your comments.

SimonSapin:
Looks good, thanks!

@bors-servo r+ p=1

Prioritizing because I have an upcoming PR made on top of this one to avoid conflicts.

PR #25594, opened 2020-01-24
Background follow ups

SimonSapin:
* Fix `background-clip` + `border-radius`
* Fix `background-clip` + `background-color`
* Render linear and radial gradients
* Enable `css/css-background/` in WPT


components/layout_2020/display_list/background.rs
```diff
@@ -0,0 +1,250 @@
...
+        bounds,
+        tile_size,
+        tile_spacing,
+        repeat: result_x.repeat || result_y.repeat,
+    })
+}
+
+/// Abstract over the horizontal or vertical dimension
+/// Coordinates (0, 0) for the purpose of this function are the positioning area’s origin.
+fn layout_1d(
```
mrobinson: Perhaps it makes sense to make this a factory method for `Layout1DResult`?

components/layout_2020/display_list/background.rs
```diff
@@ -0,0 +1,250 @@
...
+    tile_size: &mut f32,
+    tile_spacing: &mut f32,
+    mut repeat: Repeat,
```
mrobinson: Any reason to have input/output parameters and a return struct here instead of simply returning these values in `Layout1DResult`?

components/layout_2020/display_list/gradient.rs
```diff
@@ -0,0 +1,336 @@
...
+                Left => -gradient_box.height,
+            };
+            let y = match vertical {
+                Top => gradient_box.width,
+                Bottom => -gradient_box.width,
+            };
+
+            // `{ x, y }` is now a vector of arbitrary length
+            // with the same direction as the gradient line.
+
```
mrobinson: Extra line here?

components/layout_2020/display_list/gradient.rs
```diff
@@ -0,0 +1,336 @@
...
+        &layer.common,
+        layer.bounds,
+        radial_gradient,
+        layer.tile_size,
+        layer.tile_spacing,
+    )
+}
+
+/// https://drafts.csswg.org/css-images-4/#color-stop-fixup
+fn stops_fixup(
```
mrobinson: Maybe name this `fixup_stops` to make the function name a verb?

components/layout_2020/display_list/mod.rs
```diff
@@ -199,38 +206,51 @@ impl<'a> BuilderForBoxFragment<'a> {
...
+                        .to_webrender(),
+                ),
+                self.border_rect,
+                builder,
+            )
+        })
+    }
+
+    fn content_edge_clip(&self, builder: &mut DisplayListBuilder) -> Option<wr::ClipId> {
+        *self.content_edge_clip_id.init_once(|| {
```
mrobinson: Nice use of init_once here and above!

components/layout_2020/display_list/mod.rs
```diff
@@ -199,38 +206,51 @@ impl<'a> BuilderForBoxFragment<'a> {
...
+        })
     }
 
     fn build(&mut self, builder: &mut DisplayListBuilder) {
         let hit_info = hit_info(&self.fragment.style, self.fragment.tag, Cursor::Default);
         if hit_info.is_some() {
             let mut common = builder.common_properties(self.border_rect);
             common.hit_info = hit_info;
-            self.with_border_edge_clip(builder, &mut common);
+            if let Some(clip_id) = self.border_edge_clip(builder) {
```
mrobinson: At some point it would be good to keep a stack of clips and spatial node ids, but for the moment I don't think it's necessary.

mrobinson (commented):
This looks pretty good to me modulo a few small doubts.

components/layout_2020/display_list/mod.rs
SimonSapin: Yes, having that stack on the call stack is the idea of:

https://github.com/servo/servo/blob/12693b51f510761d5dd58a9103596a192edb803c/components/layout_2020/display_list.rs#L63-L70

… which I copied from Layout 2013 code, but ended up not using for backgrounds because each background layer has a potentially-different `background-clip` value that only applies to one display item, so pushing/popping to the stack isn’t as useful.

components/layout_2020/display_list/background.rs
```diff
@@ -0,0 +1,250 @@
...
+use euclid::{Size2D, Vector2D};
+use style::computed_values::background_clip::single_value::T as Clip;
+use style::computed_values::background_origin::single_value::T as Origin;
+use style::values::computed::background::BackgroundSize as Size;
+use style::values::computed::{Length, LengthPercentage};
+use style::values::specified::background::BackgroundRepeat as RepeatXY;
+use style::values::specified::background::BackgroundRepeatKeyword as Repeat;
+use webrender_api::{self as wr, units};
+
+pub(super) struct Layer {
```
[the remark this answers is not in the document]
SimonSapin: Done.

components/layout_2020/display_list/background.rs
SimonSapin: I think that would be slightly nicer API design, but perhaps not worth the rightwards drift for a private function called exactly twice

components/layout_2020/display_list/background.rs
SimonSapin: I’ve done so for `tile_spacing` but kept `tile_size` since it already has a non-trivial value before this call and is only sometimes modified.

components/layout_2020/display_list/gradient.rs
SimonSapin: Done

SimonSapin:
What happened, here? https://community-tc.services.mozilla.com/tasks/groups/afsyHI-hQuC1cMli-ibEQQ was all green and d9cd4abed356bc73f92a5fc3208a3fd92a4abb46 got a green Status but it wasn’t merged. Did Homu drop it on the floor?

(After the second r+ Homu made an other merge commit 31eafadcb3ca84cfd0a1473fdc0e0ea17ebe9251 and https://community-tc.services.mozilla.com/tasks/groups/dvO_ldpRSwmaBBVTrhc06w re-used successful tasks from the previous task group, as intended.)

jdm:
I looked at all of the webhooks and github did not report having problems delivering all of them. The homu log on servo-master1 didn't show any useful information. It's a mystery!

PR #25598, opened 2020-01-24
Expire cookies on lookup

pshaughn:
Cookies were rarely, if ever, expiring; this caused tests that used max-age=0 as a cookie deletion method to have cookies they shouldn't have had.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #24911

- [X] There are tests for these changes

PR #25599, opened 2020-01-24
Enable cookies tests

### LINKED ISSUE #24869: Enable cookies WPT tests
Tests like tests/wpt/web-platform-tests/cookies/http-state/path-tests.html are testing fundamental parts of the cookie spec, so we should be running them.

pshaughn:
fix #24869

The first run of this will spit out a lot of new test failure metadata, of course.

### DIFF
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -7,6 +7,10 @@ skip: true
   skip: false
 [2dcontext]
   skip: false
+[cookies]
+  skip: false
+  [samesite]
+    skip: true
 [cors]
   skip: false
 [css]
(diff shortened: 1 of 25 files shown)


jdm:
I suspect it might make sense to skip the samesite subdirectory - it's a lot of timeouts right now, and we don't have any implementation for the new samesite cookie spec (https://web.dev/samesite-cookies-explained/), and the tests appear to use features like multiple windows that we don't support yet.

pshaughn:
Most of the failures on the first pass were because #25598 hadn't landed yet.
@bors-servo try=wpt

PR #25603, opened 2020-01-25
Update CanvasRenderingContext2D fillStyle and strokeStyle only when required

pylbrecht:
So far fill and stroke style updates have been sent to the canvas paint
thread by `SetFillStyle()` and `SetStrokeStyle()`. This resulted in
fill/stroke style updates not being considered by the canvas paint
thread between the latest call of `SetFillStyle()`/`SetStrokeStyle()` and
the drawing operation (e.g. fill or stroke).

This issue is solved by making `SetFillStyle()` and `SetStrokeStyle()`
update the local canvas state and propagating the state to the canvas
paint thread along with the drawing operations that require it.

---


pylbrecht:
Actually I think it would be better to explicitly send `fillStyle`/`strokeStyle` with the message of the corresponding drawing operation (`Canvas2dMsg::Fill(fill_style)`) instead of just updating it before each drawing operation. This way we can't forget to do the neccessary update.

But first I want to see if these changes break any other tests.

jdm:
Looks like no unexpected failures, just #24726.

PR #25610, opened 2020-01-25
do less on abort if not done

### LINKED ISSUE #24931: XMLHttpRequest.abort before send() changes readyState to 0
WPT tests xhr/abort-during-unsent, abort-during-open, and abort-event-abort all call abort() on an XHR that hasn't had send() called yet, so abort() should be a no-op, but instead readyState is changing.

pshaughn:
Abort() was resetting state when it didn't need to, and possibly also not resetting as much of it as it needed to. I'm not sure if this is a completely correct fix but it passes some WPT tests.

---

- [X] These changes fix #24931 

- [X] There are tests for these change

### DIFF
--- a/components/script/dom/xmlhttprequest.rs
+++ b/components/script/dom/xmlhttprequest.rs
@@ -758,7 +758,12 @@ impl XMLHttpRequestMethods for XMLHttpRequest {
             }
         }
         // Step 3
-        self.ready_state.set(XMLHttpRequestState::Unsent);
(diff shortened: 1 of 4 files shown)


pshaughn:
Hoping this will pass more than just the few tests.
@bors-servo try=wpt

pshaughn:
Not the wide scope I was hoping for, but it's all positive changes.

Manishearth (approved):

jdm:
@bors-servo retry
* #25513

PR #25624, opened 2020-01-27
Add trait DomObjectWrap to provide WRAP function

### LINKED ISSUE #8576: Remove the `wrap_fn` argument from `reflect_dom_object`.
It should be possible to derive it from the type. Maybe `FooBinding::Wrap` should be `<Foo as ConcreteDOMObject>::wrap`? (Name of the trait up for bikeshedding.)

0x00A5:
---

- [ ] These changes fix #8576 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2858,6 +2858,50 @@ impl PartialEq for %(name)s {
 """ % {'check': check, 'name': name}
 
 
+class CGDomObjectWrap(CGThing):
+    """
+    Class for codegen of an implementation of the DomObjectWrap trait.
+    """
+    def __init__(self, descriptor):
+        CGThing.__init__(self)
+        self.descriptor = descriptor
+
+    def define(self):
+        name = self.descriptor.concreteType
+        name = "dom::%s::%s" % (name.lower(), name)
+        return """\
+impl DomObjectWrap for %s {
+    const WRAP: unsafe fn(
+        SafeJSContext,
+        &GlobalScope,
+        Box<Self>,
+    ) -> Root<Dom<Self>> = Wrap;
+}
+""" % (name)
+
+
+class CGDomObjectIteratorWrap(CGThing):
+    """
+    Class for codegen of an implementation of the DomObjectIteratorWrap trait.
+    """
+    def __init__(self, descriptor):
+        CGThing.__init__(self)
+        self.descriptor = descriptor
+
+    def define(self):
+        assert self.descriptor.interface.isIteratorInterface()
+        name = self.descriptor.interface.iterableInterface.identifier.name
+        return """\
+impl DomObjectIteratorWrap for %s {
+    const ITER_WRAP: unsafe fn(
+        SafeJSContext,
+        &GlobalScope,
+        Box<IterableIterator<Self>>,
+    ) -> Root<Dom<IterableIterator<Self>>> = Wrap;
+}
+""" % (name)
+
+
 class CGAbstractExternMethod(CGAbstractMethod):
     """
     Abstract base class for codegen of implementation-only (no
@@ -6067,6 +6111,8 @@ def generate_imports(config, cgthings, descriptors, callbacks=None, dictionaries
         'crate::dom::bindings::namespace::create_namespace_object',
         'crate::dom::bindings::reflector::MutDomObject',
         'crate::dom::bindings::reflector::DomObject',
+        'crate::dom::bindings::reflector::DomObjectWrap',
+        'crate::dom::bindings::reflector::DomObjectIteratorWrap',
         'crate::dom::bindings::root::Dom',
         'crate::dom::bindings::root::DomRoot',
         'crate::dom::bindings::root::DomSlice',
@@ -6286,6 +6332,10 @@ class CGDescriptor(CGThing):
                 cgThings.append(CGWrapGlobalMethod(descriptor, properties))
             else:
                 cgThings.append(CGWrapMethod(descriptor))
+                if descriptor.interface.isIteratorInterface():
+                    cgThings.append(CGDomObjectIteratorWrap(descriptor))
+                else:
+                    cgThings.append(CGDomObjectWrap(descriptor))
             reexports.append('Wrap')
 
         haveUnscopables = False
@@ -7445,9 +7495,7 @@ class CGIterableMethodGenerator(CGGeneric):
             return
         CGGeneric.__init__(self, fill(
             """
-            let result = ${iterClass}::new(&*this,
-                                           IteratorType::${itrMethod},
-                                           super::${ifaceName}IteratorBinding::Wrap);
+            let result = ${iterClass}::new(&*this, IteratorType::${itrMethod});
             """,
             iterClass=iteratorNativeType(descriptor, True),
             ifaceName=descriptor.interface.identifier.name,
--- a/components/script/dom/bindings/iterable.rs
+++ b/components/script/dom/bindings/iterable.rs
@@ -9,8 +9,10 @@
 use crate::dom::bindings::codegen::Bindings::IterableIteratorBinding::IterableKeyAndValueResult;
 use crate::dom::bindings::codegen::Bindings::IterableIteratorBinding::IterableKeyOrValueResult;
 use crate::dom::bindings::error::Fallible;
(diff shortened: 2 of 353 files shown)


lyuyuan:
@jdm I have tried adding the `DomObjectWrap` trait implementation in the derive macro. But there are some interface do not provide the `Wrap` function. And `WindowProxy` doesn't have a generated binding. The build errors can be found [here](https://community-tc.services.mozilla.com/tasks/HzWrrf1vQT6aM2AjjW4uhQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FHzWrrf1vQT6aM2AjjW4uhQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L2035).

jdm:
Ok, there's an interesting mixture here:
* some interfaces (eg. GlobalScope, CSS, CharacterData) are marked `[Abstract]` in the WebIDL, meaning there is no way to create a reflector that has that particular concrete type (but interfaces that inherit from those interfaces are fine). This is why there's no Wrap method present.
* other interfaces (eg. Headers, FormData) come from WebIDL files that contain multiple types like an interface with an iterator. This means the exported names are in the form {$name}Wrap instead (ie. FormDataWrap).
* some interfaces (eg. Promise, WindowProxy) have the `[NoInterfaceObject]` marking in the WebIDL, so there is no way to create a reflector for those types, so there is no Wrap method present.

I think a reasonable choice here is:
* change the code generation to [always export the `${name}Wrap` form](https://github.com/servo/servo/blob/688d0768898343347b7b5ba140cbc6340677be90/components/script/dom/bindings/codegen/CodegenRust.py#L6174-L6176) for all interfaces, and use that form in the dom_struct plugin
* define a `invalid_wrap_function_do_not_use` function in components/script/dom/bindings/reflector.rs that panics
* for [interfaces that do not generate a Wrap method](https://github.com/servo/servo/blob/688d0768898343347b7b5ba140cbc6340677be90/components/script/dom/bindings/codegen/CodegenRust.py#L6244), change the code generation to add a `pub use crate::dom::bindings::reflector::invalid_wrap_function_do_not_use as Wrap`

lyuyuan:
> change the code generation to always export the ${name}Wrap form for all interfaces, and use that form in the dom_struct plugin

This seems to break the code in [htmlconstructor](https://github.com/servo/servo/blob/107a29121c8a0e21d03645dd07121b59b595ac93/components/script/dom/bindings/htmlconstructor.rs#L207), as it makes all HTML element bindings reexport the `GetConstructorObject` method in the form of `${name}GetConstructorObject`.

I am modifying the [`get_constructor`](https://github.com/servo/servo/blob/107a29121c8a0e21d03645dd07121b59b595ac93/components/script/dom/bindings/htmlconstructor.rs#L205) macro to adapt this change. Does it sound right?

jdm:
That sounds reasonable for now.

jdm:
This looks like a nice improvement so far! Is this PR still WIP, or is the next step to review the code?

lyuyuan:
Hi @jdm I am still working on some build errors caused by my change. I will switch it ready-for-review once I fix them. Thanks!

lyuyuan:
Hi @jdm , sorry to bother you again. How to get `Wrap` function implemented in `IterableIteratorBinding`? I have added [this](https://github.com/servo/servo/pull/25624/files#diff-60d01595cff328c165842fea9e4ccbc2R6279) in CodegenRust.py, but the iterable interface still misses it.

lyuyuan:
HI @jdm ! Build and test errors are all gone. I think this is ready for review. Thanks!

components/script/dom/window.rs
```diff
@@ -2340,8 +2340,11 @@ impl Window {
             player_context,
             event_loop_waker,
         });
-
-        unsafe { WindowBinding::Wrap(JSContext::from_ptr(runtime.cx()), win) }
+        unsafe { WindowBinding::WindowWrap(
+            JSContext::from_ptr(runtime.cx()),
+            &GlobalScope::incumbent().unwrap(),
```
lyuyuan: Hi @jdm ! To satisfy the uniform `Wrap` function signature, I now need to pass `_scope: &GlobalScope` into `Window::Wrap` as well. It is not used in the `Wrap` function, but how to provide a "dummy" scope here? Thanks.

jdm: This is an interesting question! I think the best thing to do here would be the following:
reflector.rs:
```rust
pub enum WrapFn {
    Global(unsafe fn(JSContext, Box<T>) -> DomRoot<T>),
    NonGlobal(unsafe fn(JSContext, &GlobalScope, Box<T>) -> DomRoot<T>)
}
```
Then the trait can use the WrapFn type and implementations can choose the appropriate variant.

jdm: Thinking more about this, I'm not sure my solution will make a difference. I believe the problem comes from implementing DomObjectWrap as part of the dom_struct derive. If we generate it as part of the webidl codegen process instead, I believe we can choose to avoid implementing it for globals (and possibly other types that need special wrapping behaviour, like WindowProxy) and that should avoid any need for an enum like WrapFn.

lyuyuan: Including the trait implementation in the WebIDL codegen process seems to work! Thanks for your suggestion!

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -6286,6 +6309,8 @@ def reexportedName(name):
                 cgThings.append(CGWrapGlobalMethod(descriptor, properties))
             else:
                 cgThings.append(CGWrapMethod(descriptor))
+                if not descriptor.interface.isIteratorInterface():
```
jdm: Why can we not generate a DomObjectWrap implementation for iterator interfaces?

jdm:
Overall I'm really excited about these changes!
@bors-servo try=wpt

components/script/dom/bindings/iterable.rs
```diff
@@ -119,6 +117,11 @@ impl<T: DomObject + JSTraceable + Iterable> IterableIterator<T> {
     }
 }
 
+impl<T: DomObject + DomObjectWrap + JSTraceable + Iterable> DomObjectWrap for IterableIterator<T> {
+    const WRAP: unsafe fn(JSContext, &GlobalScope, Box<Self>) -> Root<Dom<Self>> =
```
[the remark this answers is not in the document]
lyuyuan: You are right. WPT crashed when executing `IterableIterator::new` 😢.

jdm:
Indeed:
```
  ▶ CRASH [expected OK] /fetch/api/redirect/redirect-location.any.worker.html
  │ 
  │ 
  │ 
  │ 
  │ Invalid wrap function, do not use (thread WebWorker for http://web-platform.test:8000/fetch/api/redirect/redirect-location.any.worker.js, at components/script/dom/bindings/reflector.rs:121)
  │ stack backtrace:
  │              at src/libstd/panicking.rs:513
  │    4: script::dom::bindings::iterable::IterableIterator<T>::new
  │    5: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  │    6: script::dom::bindings::codegen::Bindings::HeadersBinding::HeadersBinding::entries
  │    7: CallJitMethodOp
  │    8: script::dom::bindings::utils::generic_call
```

components/script/dom/bindings/iterable.rs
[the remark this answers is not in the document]
lyuyuan: `IterableIterator::new` calls `reflect_dom_object<T, U>` which requires T implements DomObjectWrap`. This requires `IterableIterator` implements the trait.

components/script/dom/bindings/codegen/CodegenRust.py
lyuyuan: Sorry for making this mistake. For interfaces that provides Wrap function for both itself and IterableIterator<T> (FormData, Header...), I introduced another trait `DomObjectIteratorWrap` which provides a const `ITER_WRAP` that points to the iterator Wrap function. Then, `IterableIterator<T>::WRAP` could point to `T::ITER_WRAP`, therefore, implements the `DomObjectWrap` trait.

jdm:
Great solution!
@bors-servo r+

jdm:
@bors-servo retry
#25513

Issue #25530, opened 2020-01-15
Intermittent "OSError: [Errno 12] Cannot allocate memory" during linux WPT tests


jdm:
```
  ▶ ERROR [expected OK] /css/css-fonts/parsing/font-weight-valid.html
  │   → Traceback (most recent call last):
  │   →   File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/executors/base.py", line 256, in run_test
  │   →     result = self.do_test(test)
  │   →   File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/executors/executorservo.py", line 108, in do_test
  │   →     self.proc.run()
  │   →   File "/repo/python/_virtualenv2.7/lib/python2.7/site-packages/mozprocess/processhandler.py", line 785, in run
  │   →     self.proc = self.Process([self.cmd] + self.args, **args)
  │   →   File "/repo/python/_virtualenv2.7/lib/python2.7/site-packages/mozprocess/processhandler.py", line 123, in __init__
  │   →     universal_newlines, startupinfo, creationflags)
  │   →   File "/usr/lib/python2.7/subprocess.py", line 394, in __init__
  │   →     errread, errwrite)
  │   →   File "/usr/lib/python2.7/subprocess.py", line 938, in _execute_child
  │   →     self.pid = os.fork()
  └   → OSError: [Errno 12] Cannot allocate memory
  ▶ ERROR [expected OK] /referrer-policy/gen/top.meta/strict-origin-when-cross-origin/fetch/same-http.keep-origin.http.html
  │   → Traceback (most recent call last):
  │   →   File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/executors/base.py", line 256, in run_test
  │   →     result = self.do_test(test)
  │   →   File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/executors/executorservo.py", line 108, in do_test
  │   →     self.proc.run()
  │   →   File "/repo/python/_virtualenv2.7/lib/python2.7/site-packages/mozprocess/processhandler.py", line 785, in run
  │   →     self.proc = self.Process([self.cmd] + self.args, **args)
  │   →   File "/repo/python/_virtualenv2.7/lib/python2.7/site-packages/mozprocess/processhandler.py", line 123, in __init__
  │   →     universal_newlines, startupinfo, creationflags)
[...]

PR #25625, opened 2020-01-27
Make error events only pass the extra error-event arguments to global listeners

### LINKED ISSUE #25197: Dispatching an ErrorEvent to a document passes five arguments to document.onerror instead of one (it appears that only window.onerror should get five)
This happens in WPT html/webappapis/scripting/events/event-handler-processing-algorithm-error/script-element.html and document-synthetic-errorevent.html.

An error event that occurs in Javascript execution will normally hit the listener that's on window (and should be reflected to document.body), not the listener on document. As far as I can tell, the only way to hit document.onerror is to explicitly dispatch an error event directly at the document, which is what these tests do.

pshaughn:
Error event handlers were getting lineno, etc. arguments even when they weren't on globals, failing two WPT tests; this fixes that.

---

- [X] These changes fix #25197

### DIFF
--- a/components/script/dom/eventtarget.rs
+++ b/components/script/dom/eventtarget.rs
@@ -33,6 +33,7 @@ use crate::dom::htmlformelement::FormControlElementHelpers;
 use crate::dom::node::document_from_node;
 use crate::dom::virtualmethods::VirtualMethods;
 use crate::dom::window::Window;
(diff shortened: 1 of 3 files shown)


pshaughn:
#25530

jdm:
@bors-servo retry
* #25513 #24726

PR #25629, opened 2020-01-28
Implement "translate" attribute

### LINKED ISSUE #25628: Implement the global "translate" attribute
https://html.spec.whatwg.org/multipage/dom.html#translation-mode
https://html.spec.whatwg.org/multipage/dom.html#dom-translate
https://html.spec.whatwg.org/multipage/dom.html#attr-translate

WPT tests in html/dom/elements/global-attributes/the-translate-attribute*

pshaughn:
This attribute is almost a straightforward enumerated one, but the getter value inherits from parents when the content attribute is absent, even when the parents are non-HTML elements. This initial commit is using LocalName::from on a static string; once html5ever has a release with "translate" in the built-in local name list, a small change will be needed.

---

- [X] These changes fix #25628

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -514,6 +514,25 @@ impl Element {
             debug_assert!(false, "Trying to detach a non-attached shadow root");
         }
     }
+
+    // https://html.spec.whatwg.org/multipage/#translation-mode
+    pub fn is_translate_enabled(&self) -> bool {
+        // TODO change this to local_name! when html5ever updates
+        let name = &LocalName::from("translate");
+        if self.has_attribute(name) {
+            match &*self.get_string_attribute(name) {
+                "yes" | "" => return true,
+                "no" => return false,
+                _ => {},
+            }
+        }
+        if let Some(parent) = self.upcast::<Node>().GetParentNode() {
+            if let Some(elem) = parent.downcast::<Element>() {
+                return elem.is_translate_enabled();
+            }
+        }
+        true // whatwg/html#5239
+    }
 }
 
 #[allow(unsafe_code)]
--- a/components/script/dom/htmlelement.rs
+++ b/components/script/dom/htmlelement.rs
@@ -516,6 +516,23 @@ impl HTMLElementMethods for HTMLElement {
         // Step 7.
         Node::replace_all(Some(fragment.upcast()), self.upcast::<Node>());
     }
+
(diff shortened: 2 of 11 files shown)


components/script/dom/element.rs
```diff
@@ -514,6 +514,26 @@ impl Element {
...
         }
     }
+
+    // https://html.spec.whatwg.org/multipage/#translation-mode
+    pub fn is_translate_enabled(&self) -> bool {
+        // TODO compile-time intern "translate"
+        let name = &LocalName::from("translate");
+        if self.has_attribute(name) {
+            match &*self.get_string_attribute(name) {
+                "yes" => return true,
```
panaman67:
```suggestion
                "yes" | "" => return true,
```

components/script/dom/element.rs
```diff
@@ -514,6 +514,25 @@ impl Element {
...
+                "no" => return false,
+                _ => {},
+            }
+        }
+        if let Some(parent) = self.upcast::<Node>().GetParentNode() {
+            if let Some(elem) = parent.downcast::<Element>() {
+                return elem.is_translate_enabled();
+            }
+        }
+        true
```
jdm: This should be false if this node is not the document element.

jdm (changes requested):

components/script/dom/element.rs
pshaughn: document.createElement("div").translate is true in Chrome, but it is interesting that the spec doesn't actually say what to do if there's no parent and it's not the document element. https://github.com/whatwg/html/issues/5239

jdm:
```
  ▶ Unexpected subtest result in /custom-elements/reactions/HTMLElement.html:
  └ PASS [expected FAIL] translate on HTMLElement must enqueue an attributeChanged reaction when adding translate content attribute
  ▶ Unexpected subtest result in /custom-elements/reactions/HTMLElement.html:
  └ PASS [expected FAIL] translate on HTMLElement must enqueue an attributeChanged reaction when replacing an existing attribute
  ▶ Unexpected subtest result in /html/dom/idlharness.https.html?include=HTML.*:
  └ PASS [expected FAIL] HTMLElement interface: attribute translate
  ▶ Unexpected subtest result in /html/dom/idlharness.https.html?include=HTML.*:
  └ PASS [expected FAIL] HTMLElement interface: document.createElement("noscript") must inherit property "translate" with the proper type
```

Issue #25613, opened 2020-01-26
Intermittent FAIL in /cookies/prefix/__host.header.html


CYBAI:
```
  ▶ Unexpected subtest result in /cookies/prefix/__host.header.html:
  │ FAIL [expected PASS] __Host: Non-secure origin: Does not set 'Path=/;'
  └   → promise_test: Unhandled rejection with value: object "Error: Internal Server Error"
```

CYBAI:
Also observed

```
  ▶ Unexpected subtest result in /cookies/prefix/__host.document-cookie.html:
  │ FAIL [expected PASS] __Host: Non-secure origin: 'Path=/;'
  └   → promise_test: Unhandled rejection with value: object "Error: Internal Server Error"
  ▶ Unexpected subtest result in /cookies/prefix/__host.document-cookie.html:
  │ FAIL [expected PASS] __Host: Non-secure origin: 'Secure; Path=/;'
  └   → promise_test: Unhandled rejection with value: object "Error: Internal Server Error"
```

jdm:
```
  ▶ Unexpected subtest result in /cookies/prefix/__host.document-cookie.https.html:
  │ FAIL [expected PASS] __Host: Secure origin: Does not set 'Path=/;'
  └   → promise_test: Unhandled rejection with value: object "Error: Internal Server Error"
```

jdm:
Also observed in /cookies/prefix/document-cookie.non-secure.html  according to #25742.

TimvdLippe:
Closing as stale as this issue has not received any failure references for over a month

PR #25633, opened 2020-01-28
Remove unnecessary LocalName::from(constant) calls

pshaughn:
Changing calls to LocalName::from with a constant argument to use the interned local name instead doesn't really save a meaningful amount of time (when it's not in a loop or on a critical path), but it's good for codebase consistency.


jdm:
@bors-servo r+
Agreed!

jdm:
@bors-servo retry
* #25513

CYBAI:
@bors-servo retry

- #25513 
- #24726 
- #25613

PR #25637, opened 2020-01-29
Initial implementation of GPUBindGroup for WebGPU

imiklos:
Added WebIDL bindings for `GPUBindGroup`.
Implemented the `createBindGroup` function of `GPUDevice`
Renamed `GPUBindGroupBinding` to `GPUBindGroupBindings` and `GPUBufferBinding` to `GPUBufferBindings` in the WebIDL, because these names are already occupied.

---

cc @kvark @jdm @zakorgy

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -6,6 +6,7 @@
 
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::GPUAdapterBinding::GPULimits;
+use crate::dom::bindings::codegen::Bindings::GPUBindGroupBinding::GPUBindGroupDescriptor;
 use crate::dom::bindings::codegen::Bindings::GPUBindGroupLayoutBinding::{
     GPUBindGroupLayoutBindings, GPUBindGroupLayoutDescriptor, GPUBindingType,
 };
@@ -18,6 +19,7 @@ use crate::dom::bindings::str::DOMString;
 use crate::dom::eventtarget::EventTarget;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::gpuadapter::GPUAdapter;
+use crate::dom::gpubindgroup::GPUBindGroup;
 use crate::dom::gpubindgrouplayout::GPUBindGroupLayout;
 use crate::dom::gpubuffer::{GPUBuffer, GPUBufferState};
 use crate::dom::gpupipelinelayout::GPUPipelineLayout;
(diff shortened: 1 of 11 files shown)


components/script/dom/identityhub.rs
```diff
@@ -132,6 +138,20 @@ impl Identities {
         }
     }
 
+    pub fn create_bind_group_id(&mut self, backend: Backend) -> BindGroupId {
```
kvark: Why aren't we doing this simpler, like in https://searchfox.org/mozilla-central/rev/ad6234148892bc29bf08d736604ac71925138040/gfx/wgpu/wgpu-remote/src/lib.rs#277

imiklos: Yes, you're right! This could be simpler. I will remove the code duplication in another pull request.

imiklos: I've just created a PR for this: #25648

components/script/dom/gpudevice.rs
```diff
@@ -463,4 +468,64 @@ impl GPUDeviceMethods for GPUDevice {
...
+            let buffer_size = bind.resource.buffer.size();
+            let resource_size = bind.resource.size.unwrap_or(buffer_size);
+            let length = bind.resource.offset.checked_add(resource_size);
+
+            length.is_some() &&
+            buffer_size >= length.unwrap() && // check buffer OOB
+            bind.resource.offset % alignment == 0 && // check alignment
+            bind.resource.offset < buffer_size && // on Vulkan offset must be less than size of buffer
+            descriptor.layout.bindings().iter().any(|layout_bind| {
+                let usage = BufferUsage::from_bits(bind.resource.buffer.usage()).unwrap();
```
jdm: This could be moved outside of the `any` call, right?

components/script/dom/gpudevice.rs
```diff
@@ -463,4 +468,64 @@ impl GPUDeviceMethods for GPUDevice {
...
+                let ty = match layout_bind.type_ {
+                    GPUBindingType::Storage_buffer  => BufferUsage::STORAGE,
+                    // GPUBindingType::Readonly_storage_buffer  => BufferUsage::STORAGE_READ,
```
jdm: Why is this commented out?

jdm (approved):
Looks reasonable!

components/script/dom/gpudevice.rs
imiklos: The `BufferUsage::STORAGE_READ` is not in the Draft so neither in the WebIDL yet, but it is in wgpu-core. I just wanted it to be in place already, but I can remove it.

components/script/dom/gpudevice.rs
imiklos: Yes, you are right! I will move outside the `any`.

PR #25640, opened 2020-01-29
move ConstellationMsg to compositing

### LINKED ISSUE #20758: Move compositor and layout messages out of script_traits
This enum is, I think, only used by the compositor:

https://github.com/servo/servo/blob/dc17b32cebc956385084288765c9f5a78e60ee28/components/script_traits/lib.rs#L666

Why not move it to compositing? It would be more logical, remove potential circular dependencies, and maybe even save us a `#[derive(Deserialize, Serialize)]`!

Layout relevant message enums could also be moved to `layout_thread`(these still would require `#[derive(Deserialize, Serialize)]`):

https://github.com/servo/servo/blob/dc17b32cebc956385084288765c9f5a78e60ee28/components/script_traits/lib.rs#L111

and 

https://github.com/servo/servo/blob/dc17b32cebc956385084288765c9f5a78e60ee28/components/script_traits/script_msg.rs#L32

kunalmohan:
`ConstellationMsg` enum in script_traits is moved to compositing since it's not used by script_traits. The other enums mentioned in the issue, `LayoutControllerMsg` and `LayoutMsg` cannot be moved out of script_traits since script_traits depends on both of them and moving out would create cyclic dependencies.

r?@gterzian

---

- [X] These changes fix #20758  (GitHub issue number if applicable)

### DIFF
--- a/components/compositing/Cargo.toml
+++ b/components/compositing/Cargo.toml
@@ -31,6 +31,7 @@ net_traits = {path = "../net_traits"}
 num-traits = "0.2"
 pixels = {path = "../pixels", optional = true}
 profile_traits = {path = "../profile_traits"}
+rust-webvr = {version = "0.16", features = ["mock", "openvr", "vrexternal"]}
 script_traits = {path = "../script_traits"}
 servo_geometry = {path = "../geometry"}
 servo-media = {git = "https://github.com/servo/media"}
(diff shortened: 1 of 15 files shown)


components/compositing/Cargo.toml
```diff
@@ -31,6 +31,7 @@ net_traits = {path = "../net_traits"}
 num-traits = "0.2"
 pixels = {path = "../pixels", optional = true}
 profile_traits = {path = "../profile_traits"}
+rust-webvr = {version = "0.16", features = ["mock", "openvr", "vrexternal"]}
```
gterzian: is this related? There is another webvr change in another file.

kunalmohan: compositing depended on webvr for `VRServiceManager`. Moving `ConstellationMsg` to compositing made webvr depend on compositing, which created a cyclic dependency. To solve this, since `VRServiceManager` was just re-exported in webvr, I imported `VRServiceManager` directly from `rust-webvr` instead of `webvr`.

gterzian: Ok, good.

PR #25644, opened 2020-01-30
Give error and resize special getter/setters for body/frameset reflection

### LINKED ISSUE #25186: Certain event handlers aren't reflected between window and document.body
This is various failures attributed to WPT html/webappapis/scripting/events/event-handler-attributes-body-window.html and -frameset-window.html; the actual test code is a directory deeper, resources/event-handler-body.js.

The error is that setting document.body.oncopy, oncut, onpaste, onerror, or onresize doesn't cause the corresponding attributes of window to appear to change, and vice versa.

The same thing should also happen when setting on a document.createElement("body") or a document.createElement("frameset").

Most events do reflect correctly; it's just this handful that don't.

### LINKED ISSUE #25187: Setting element.onerror on a body or frameset element from a document that doesn't have a window should leave it null, not actually set it.
This is case "Ignore setting of error window event handlers on windowless {body|frameset}" in html/webappapis/scripting/events/event-handler-attributes-windowless-body.html. Many handlers are tested, only error fails.

pshaughn:
Most of the event handlers that needed to be reflected between body and window were doing so via special getter/setters in htmlelement.rs, but error and resize were missing; they are now included, passing the tests for whether these are reflected.

---

- [X] These changes fix #25186 for the case that isn't a bad test or an unimplemented event type, and fix #25187 

### DIFF
--- a/components/script/dom/htmlelement.rs
+++ b/components/script/dom/htmlelement.rs
@@ -5,6 +5,7 @@
 use crate::dom::activation::{synthetic_click_activation, ActivationSource};
 use crate::dom::attr::Attr;
 use crate::dom::bindings::codegen::Bindings::EventHandlerBinding::EventHandlerNonNull;
+use crate::dom::bindings::codegen::Bindings::EventHandlerBinding::OnErrorEventHandlerNonNull;
 use crate::dom::bindings::codegen::Bindings::HTMLElementBinding;
 use crate::dom::bindings::codegen::Bindings::HTMLElementBinding::HTMLElementMethods;
 use crate::dom::bindings::codegen::Bindings::HTMLLabelElementBinding::HTMLLabelElementMethods;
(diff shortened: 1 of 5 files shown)


jdm:
Good find!
@bors-servo r+

CYBAI:
```
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/event-handler-attributes-windowless-body.html:
  └ PASS [expected FAIL] Ignore setting of error window event handlers on windowless body
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/event-handler-attributes-windowless-body.html:
  └ PASS [expected FAIL] Ignore setting of error window event handlers on windowless frameset
```

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

PR #25645, opened 2020-01-30
Add support for overflow:scroll and overflow:hidden to layout_2020

mrobinson:
This adds clipping and interactive scrolling support, but scrolling from
script is still not functional.

---

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -9,9 +9,11 @@ use crate::replaced::IntrinsicSizes;
 use embedder_traits::Cursor;
 use euclid::{Point2D, SideOffsets2D, Size2D};
 use gfx::text::glyph::GlyphStore;
+use gfx_traits::{combine_id_with_fragment_type, FragmentType};
 use mitochondria::OnceCell;
 use net_traits::image_cache::UsePlaceholder;
 use std::sync::Arc;
+use style::computed_values::overflow_x::T as ComputedOverflow;
 use style::dom::OpaqueNode;
 use style::properties::ComputedValues;
 use style::values::computed::{BorderStyle, Length, LengthPercentage};
@@ -63,8 +65,6 @@ impl<'a> DisplayListBuilder<'a> {
         wr::CommonItemProperties::new(clip_rect, self.current_space_and_clip)
     }
 
-    // FIXME: use this for the `overflow` property or anything else that clips an entire subtree.
-    #[allow(unused)]
     fn clipping_and_scrolling_scope<R>(&mut self, f: impl FnOnce(&mut Self) -> R) -> R {
         let previous = self.current_space_and_clip;
         let result = f(self);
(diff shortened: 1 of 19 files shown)


components/layout_2020/fragments.rs
```diff
@@ -180,10 +178,33 @@ impl BoxFragment {
             border,
             margin,
             block_margins_collapsed_with_children,
-            scrollable_overflow,
+            scrollable_overflow: PhysicalRect::zero(),
```
SimonSapin: I’m not a fan of initializing here with known-incorrect data and rely on `calculate_scrollable_overflow` being called later at some point for its side effect. Why is that computation delayed? If it needs to be, what do you think of making this an `Option` and calling `unwrap` when the data is used, so we don’t accidentally use a non-initialized zero?

components/layout_2020/display_list/mod.rs
```diff
@@ -256,14 +256,49 @@ impl<'a> BuilderForBoxFragment<'a> {
...
-            .translate(self.containing_block.origin.to_vector());
-        for child in &self.fragment.children {
-            child.build_display_list(builder, &content_rect)
-        }
+
+        builder.clipping_and_scrolling_scope(|builder| {
+            let overflow_x = self.fragment.style.get_box().overflow_x;
+            let overflow_y = self.fragment.style.get_box().overflow_y;
+            let original_scroll_and_clip_info = builder.current_space_and_clip;
+            if overflow_x != ComputedOverflow::Visible || overflow_y != ComputedOverflow::Visible {
```
SimonSapin: This seems to treat all values other than `visible` the same. Landing things incrementally is fine, but what’s the plan for fixing that later?

components/layout_2020/fragments.rs
mrobinson: The issue is that the BoxFragment is created and then positioned later. Does it make sense to modify `layout_block_level_children` to create and place `BoxFragment`s at the same time?

SimonSapin: Creation is separated from placement in `layout_block_level_children` so that the former (including recursion in a subtree) can be parallelized across siblings, while the latter needs to be sequential.

What if we defined `BoxFragment::scrollable_overflow` to be in a coordinate system where `self.content_rect.start_corner` is the origin, and the parent is responsible for doing the translation?

components/layout_2020/display_list/mod.rs
mrobinson: The only difference here, I think, is with Overflow::Auto. As far as correctness goes, I think the unnecessary creation of the scroll frame doesn't affect it. Overflow::Hidden is handled below.

components/layout_2020/fragments.rs
mrobinson: I have done something like this in the latest branch. Now the final position of the scrollable overflow is calculated when it is used. This shouldn't have a large overhead. I've kept the calculation inside of the child since it knows about its position when the calculation is done.

SimonSapin:
Looks like this needs some code formatting changes, see "Details" near "checks have failed to find the CI log. You can check the same locally by running `./mach test-tidy`, and in the case of Rust code formatting fix it automatically with rustfmt by running `./mach fmt`.

(I have `./mach test-tidy --no-wpt` in `.git/hooks/pre-push` because otherwise I keep forgetting.)

Feel free to `r=SimonSapin` with that fixed.

SimonSapin (approved):

jdm:
```
  ▶ PASS [expected FAIL] /css/css-backgrounds/background-size/background-size-contain.xht
  ▶ PASS [expected FAIL] /css/css-backgrounds/background-size/background-size-cover.xht
  ▶ PASS [expected FAIL] /css/css-backgrounds/background-repeat/background-repeat-repeat-y.xht
  ▶ PASS [expected FAIL] /css/css-backgrounds/background-repeat/background-repeat-repeat-x.xht
  ▶ PASS [expected FAIL] /_mozilla/css/acid2_ref.html
```

jdm:
@bors-servo retry
* #25513

jdm:
@bors-servo retry
* #24726

CYBAI:
@bors-servo retry

- #25513

jdm:
@bors-servo retry
#24726

CYBAI:
@bors-servo retry

- #25513

PR #25654, opened 2020-01-30
fix DOMTokenList

### LINKED ISSUE #25128: Emptying a classList leaves getAttribute("class") as an empty string where WPT expects null
WPT dom/nodes/Element-classlist.html fails several classList.remove test cases this way.

pshaughn:
I fixed the bug described in 25128, and also the other cases in the same test (order of elements after replace).

---

- [X] These changes fix #25128 


components/script/dom/domtokenlist.rs
```diff
@@ -52,6 +52,16 @@ impl DOMTokenList {
             slice => Ok(Atom::from(slice)),
         }
     }
+
+    fn perform_update_steps(&self, atoms: Vec<Atom>) {
```
jdm: Add a `// https://dom.spec.whatwg.org/#concept-dtl-update` comment please.

jdm (approved):

jdm:
@bors-servo retry
* #24726

CYBAI:
@bors-servo retry

- #24726

PR #25665, opened 2020-01-31
Reduce log spam in some cases

jdm:


jdm:
I have a suspicion that this will address one common cause of macOS test timeouts.

jdm:
@bors-servo retry
* #24786

jdm:
@bors-servo retry
* #25513

PR #25667, opened 2020-01-31
Don't panic due to I/O issues in console.log.

jdm:
This attempts to address something that has shown up in zombie mac processes - a long stack showing that a panic occurred inside Console::Log and then we kept panicking over and over.


CYBAI:
Looks the timeout gets worst in this round of builds 😨

jdm:
I think I may have broken things by using write! instead of writeln!

nox:
So is the fix to just use `writeln` here?

dralley:
Would be pretty fast to find out, I would think.

jdm:
This isn't an important change any more.

Issue #25112, opened 2019-12-05
Intermittent android mozjs build failure (fatal error: 'mozilla/PlatformConditionVariable.h' file not found)


jdm:
```
        ^
In file included from /Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/jsmath.cpp:31:
In file included from /Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/vm/JSContext.h:26:
In file included from /Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/vm/Runtime.h:24:
In file included from /Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/builtin/AtomicsObject.h:13:
/Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/threading/ConditionVariable.h:12:10: fatal error: 'mozilla/PlatformConditionVariable.h' file not found
#include "mozilla/PlatformConditionVariable.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pwnorbitals:
Still reproducible ?

jdm:
Yes.

TimvdLippe:
Closing as stale as this issue has not received any failure references for over a month

Issue #22140, opened 2018-11-07
Windows mozjs build intermittently fails on taskcluster (fork: Resource temporarily unavailable)


jdm:
```

--- stderr
      0 [main] sh 6040 fork_copy: linked dll data/bss pass 0 failed, 0x6F1000..0x6F1A54, done 0, windows pid 6084, Win32 error 998
C:/Users/TASK_1~3/AppData/Local/Temp/make5444-5.sh: fork: Resource temporarily unavailable
mozmake: *** [makefile.cargo:159: maybe-configure] Error 128
thread 'main' panicked at 'assertion failed: result.success()', C:\Users\task_1541612155\.cargo\registry\src\github.com-1ecc6299db9ec823\mozjs_sys-0.61.2\build.rs:110:5
stack backtrace:
   0: <std::sync::barrier::BarrierWaitResult as core::fmt::Debug>::fmt
   1: <std::path::Iter<'a> as core::convert::AsRef<std::path::Path>>::as_ref
   2: std::panicking::take_hook
   3: std::panicking::take_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::begin_panic<str*>
             at \libstd\panicking.rs:410
   6: build_script_build::build_jsapi
             at .\build.rs:110
   7: build_script_build::main
             at .\build.rs:20
   8: std::rt::lang_start::{{closure}}<()>
             at \libstd\rt.rs:74
   9: std::panicking::update_panic_count
  10: _rust_maybe_catch_panic
  11: std::rt::lang_start_internal
  12: std::rt::lang_start<()>
             at \libstd\rt.rs:74
  13: main
  14: invoke_main
             at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:78
  15: invoke_main
             at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:78
  16: BaseThreadInitThunk
  17: RtlUserThreadStart

warning: build failed, waiting for other jobs to finish...
error: build failed
Build FAILED in 0:08:15
```

SimonSapin:
Log with context: https://tools.taskcluster.net/groups/A_t6N23DRvCxcXGKU6PwXQ/tasks/OxgSb3GZQeqX8S6DxIAvkw/runs/0/logs/public%2Flogs%2Flive.log#L957

(The metadata and logs for current Taskcluster tasks expire after a year. This is configurable when scheduling tasks.)

CYBAI:
```
--- stderr
[...]

PR #25668, opened 2020-01-31
Stop iframes from including their ancestors

pshaughn:
An iframe that included itself would just keep loading until running out of resources; the spec includes protection against that, and now Servo follows (some of) the spec about it.

---

- [X] These changes fix the same-origin case presented in #3931 but not a case in which an intermediate cross-origin iframe is hiding a same-origin ancestor further up.

- [X] There is a manual test for these changes in #3931

### DIFF
--- a/components/script/dom/htmliframeelement.rs
+++ b/components/script/dom/htmliframeelement.rs
@@ -287,7 +287,30 @@ impl HTMLIFrameElement {
 
         let url = self.get_url();
 
-        // TODO: check ancestor browsing contexts for same URL
+        // TODO(#25748):
+        // By spec, we return early if there's an ancestor browsing context
+        // "whose active document's url, ignoring fragments, is equal".
+        // However, asking about ancestor browsing contexts is more nuanced than
+        // it sounds and not implemented here.
+        // Within a single origin, we can do it by walking window proxies,
+        // and this check covers only that single-origin case, protecting
+        // against simple typo self-includes but nothing more elaborate.
+        let mut ancestor = window.GetParent();
+        while let Some(a) = ancestor {
+            if let Some(ancestor_url) = a.document().map(|d| d.url()) {
+                if ancestor_url.scheme() == url.scheme() &&
+                    ancestor_url.username() == url.username() &&
+                    ancestor_url.password() == url.password() &&
+                    ancestor_url.host() == url.host() &&
+                    ancestor_url.port() == url.port() &&
+                    ancestor_url.path() == url.path() &&
+                    ancestor_url.query() == url.query()
+                {
+                    return;
+                }
+            }
+            ancestor = a.parent().map(|p| DomRoot::from_ref(p));
+        }
 
         let creator_pipeline_id = if url.as_str() == "about:blank" {
             Some(window.upcast::<GlobalScope>().pipeline_id())


pshaughn:
I'd be surprised if there wasn't at least one WPT test that did something relating to this...
@bors-servo try=wpt

pshaughn:
I am surprised! The manual test passed for me on Linux, might be good for someone on Mac or Windows to double-check.

components/script/dom/htmliframeelement.rs
```diff
@@ -287,7 +287,24 @@ impl HTMLIFrameElement {
 
         let url = self.get_url();
 
-        // TODO: check ancestor browsing contexts for same URL
+        // return early if there's an ancestor browsing context
+        // "whose active document's url, ignoring fragments, is equal"
+        let mut ancestor = window.GetParent();
+        while let Some(a) = ancestor {
```
jdm: This implementation is currently limited to same-origin iframes; unfortunately we need to query the constellation for the full iframe ancestor chain to be certain.

pshaughn: Thanks for pointing that out. Could you give me a rough sketch of how that goes?

jdm: 1. add a new message to https://github.com/servo/servo/blob/688d0768898343347b7b5ba140cbc6340677be90/components/script_traits/script_msg.rs#L117 that contains an IpcSender that will be used to send the response, along with the desired url and the frame's BrowsingContextId
1. add a message handler to https://github.com/servo/servo/blob/688d0768898343347b7b5ba140cbc6340677be90/components/constellation/constellation.rs#L1688 that checks the `browsing_context` map, then gets the browsing context's parent pipeline id (https://github.com/servo/servo/blob/688d0768898343347b7b5ba140cbc6340677be90/components/constellation/browsingcontext.rs#L63) and uses that to get the pipeline from `pipelines`, which contains a browsing context id which corresponds with the parent. Repeat until there are no parents.
1. send the result of checking the ancestors on the provided ipc sender.
1. in the script thread code, use the global scope's `script_to_constellation_chan` to send the new message, first creating a new ipc channel to include in it
1. wait on receiving a message on the new ipc channel receiver

pshaughn: Does checking the url of each browsing context have to happen on that browsing context's own script thread?

jdm: Are you asking if the check could happen as part of navigation in the constellation instead? I don't know the answer without reading the spec, sadly.

pshaughn: What I'm not sure about is how to ask a browsing context "what's your URL" without crossing a memory boundary I'm not supposed to cross. This problem might be better tackled by someone who's gone deeper into Servo's threading systems.

pshaughn: Simplifying realization: if it's not the same script thread that asked the question, it must also not be the same URL, since within an origin we're sticking to one thread. Constellation could hand back a list of contexts, then script thread could iterate them to ask the question of the same-origin ones.

pshaughn: I can imagine the list of parents changing in mid-check due to something happening in another script thread, like if we're in a complicated tree of iframes and a cross-origin parent hands us off for adoption by another iframe that's also in that origin. The implied synchronousness of this ancestry-check operation in the spec doesn't really match the way Servo does things.

pshaughn:
Maybe it'd be worth letting this change into the code to have a guardrail against simple recursions that someone could cause with a typo, and leave the issue open for the much harder case of cross-origin ones?

components/script/dom/htmliframeelement.rs
jdm: Ick, yeah. I think I see what you're getting at. One solution might be for the constellation to reply with a list of channels to ancestor script threads, and the original script thread can then query those threads directly.

jdm:
Yeah, I'm inclined to agree as long as we file an issue about the cross-origin case and add a TODO comment that references it.

components/script/dom/htmliframeelement.rs
gterzian: Does the cross-origin situation apply here? As I understand it, here we are operating on the iframe element, the BC container, not the actual nested BC which potentially is cross-site in which case it would be running in another process. 
 
(and if the nested BC that will be loaded as a result of this navigation is cross-site, I think this implies by definition that the parent BC does not have the same url?)

gterzian: Ah you mean if the nested BC is cross-site and then loads another iframe for that is cross-site to it, but matches the url of the BC where the first iframe is contained within? Yeah that's a tough one.

gterzian: I think you might want to not completely follow the spec to the letter here. So instead of returning, you could not go ahead with the navigation when the `ScriptMsg::ScriptLoadedURLInIFrame` is handled by the constellation, using the `old_pipeline_id` field to check the url of all `Pipeline` that are ancestor to it? 

So you'd have to essentially to the same check as here, but in the constellation over at https://github.com/servo/servo/blob/89384ffec35529571be41882d77d4b0438c1f8b9/components/constellation/constellation.rs#L2940

I think it could be done in a separate PR, since this code can be kept for when all parents happen to be running in the same script-thread? This would be like an early return case, sort of independent of a later check happening in the constellation?

So I think the "local" check you perform here is the "early abort" case that matches the spec. And then in the cross-site case, you'd have to do the check a bit later when the navigation is handled by the constellation. Since the spec only returns, I think the effect of the delay will not be noticeable to script (unloading only happens later if the navigation goes through, I think).

gterzian: One last thing, if the navigation were to be disallowed in the constellation, you'd probably have to send a message to the script-thread to undo some of the changes that occur as part of https://github.com/servo/servo/blob/89384ffec35529571be41882d77d4b0438c1f8b9/components/script/dom/htmliframeelement.rs#L109 when the navigation is initiated.

jdm:
@bors-servo retry
* #25112 #22140

jdm:
@bors-servo retry
* #22212

jdm:
@bors-servo retry
* oom

PR #25669, opened 2020-01-31
Update user agent info

atouchet:
GitHub is again giving warnings in Servo due to using an outdated Firefox version in the user agent string. This updates the version numbers as well as the OS versions to the latest supported releases.

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #25672, opened 2020-02-01
Allow embedders to provide user agent string

### LINKED ISSUE #25646: Allow embedders to provide user agent string
It should be up to the embedder to decide what user agent to report in network requests, rather than deep in the engine in https://github.com/servo/servo/blob/8e0d037ee82da747fa9c13234d5c3677f930e690/components/config/opts.rs#L480 (and #25372 would have exacerbated this). We should add a method to the EmbedderMethods trait that returns the default user agent string, so that we can pass that to [the new network thread](https://github.com/servo/servo/blob/8e0d037ee82da747fa9c13234d5c3677f930e690/components/servo/lib.rs#L861) when we're [creating an instance of Servo](https://github.com/servo/servo/blob/8e0d037ee82da747fa9c13234d5c3677f930e690/components/servo/lib.rs#L324).

mediremi:
Allow embedders to provide user agent string instead of setting it in [servo/components/config/opts.rs](https://github.com/servo/servo/blob/master/components/config/opts.rs).

---

- [x] These changes fix #25646

### DIFF
--- a/components/compositing/windowing.rs
+++ b/components/compositing/windowing.rs
@@ -182,6 +182,9 @@ pub trait EmbedderMethods {
 
     /// Register services with a WebXR Registry.
     fn register_webxr(&mut self, _: &mut webxr::MainThreadRegistry) {}
+
+    /// Returns the default user agent string
+    fn default_user_agent_string(&self) -> &'static str;
 }
 
 #[derive(Clone, Copy, Debug)]
--- a/components/compositing/windowing.rs
+++ b/components/compositing/windowing.rs
@@ -187,29 +187,29 @@ pub trait EmbedderMethods {
     fn default_user_agent_string(&self) -> &'static str {
         #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
         const UA_STRING: &'static str =
-            "Mozilla/5.0 (X11; Linux x86_64; rv:63.0) Servo/1.0 Firefox/63.0";
+            "Mozilla/5.0 (X11; Linux x86_64; rv:72.0) Servo/1.0 Firefox/72.0";
         #[cfg(all(target_os = "linux", not(target_arch = "x86_64")))]
         const UA_STRING: &'static str =
-            "Mozilla/5.0 (X11; Linux i686; rv:63.0) Servo/1.0 Firefox/63.0";
+            "Mozilla/5.0 (X11; Linux i686; rv:72.0) Servo/1.0 Firefox/72.0";
 
         #[cfg(all(target_os = "windows", target_arch = "x86_64"))]
         const UA_STRING: &'static str =
-            "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:63.0) Servo/1.0 Firefox/63.0";
+            "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Servo/1.0 Firefox/72.0";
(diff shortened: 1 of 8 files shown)


atouchet:
#25669 updated Servo's user agent strings. Could you please match those changes in this PR?

components/compositing/windowing.rs
```diff
@@ -182,6 +182,48 @@ pub trait EmbedderMethods {
...
+            "Mozilla/5.0 (X11; Linux i686; rv:72.0) Servo/1.0 Firefox/72.0";
+
+        #[cfg(all(target_os = "windows", target_arch = "x86_64"))]
+        const UA_STRING: &'static str =
+            "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Servo/1.0 Firefox/72.0";
+        #[cfg(all(target_os = "windows", not(target_arch = "x86_64")))]
+        const UA_STRING: &'static str =
+            "Mozilla/5.0 (Windows NT 10.0; rv:72.0) Servo/1.0 Firefox/72.0";
+
+        #[cfg(all(target_os = "macos"))]
```
paulrouget: I don’t think the `all` is necessary here.

components/compositing/windowing.rs
```diff
@@ -182,6 +182,48 @@ pub trait EmbedderMethods {
...
+        #[cfg(not(any(
+            target_os = "linux",
+            target_os = "windows",
+            target_os = "macos",
+            target_os = "android",
+            target_os = "ios"
+        )))]
+        // Use OS X user agent as fallback
+        const UA_STRING: &'static str =
+            "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Servo/1.0 Firefox/72.0";
```
paulrouget: That’s fine because it’s what we’ve been doing until now. But maybe we should just not have a fallback. Once we support a new platform, we would let it fail at compile time.

components/config/opts.rs
```diff
@@ -911,14 +855,6 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         MULTIPROCESS.store(true, Ordering::SeqCst)
     }
 
-    let user_agent = match opt_match.opt_str("u") {
-        Some(ref ua) if ua == "ios" => default_user_agent_string(UserAgent::iOS).into(),
-        Some(ref ua) if ua == "android" => default_user_agent_string(UserAgent::Android).into(),
-        Some(ref ua) if ua == "desktop" => default_user_agent_string(UserAgent::Desktop).into(),
-        Some(ua) => ua.into(),
-        None => default_user_agent_string(DEFAULT_USER_AGENT).into(),
-    };
```
paulrouget: This still needs to be possible.

paulrouget (commented):
So the UA is defined by (sorted by priority):

- the `-u` option
- the embedder implementation
- the servo implementation

Maybe in component/servo check if the option is set (it’s also fine, even maybe better, if it happens in ports/glutin), if not, call the embedder callback. And maybe the default UA function should not live in windowing.rs but in component/servo.

It would look like this:

in components/servo,
- get UA from `-u` option (maybe via ports/glutin/main2.rs (see the arg parsing code there), not opts.rs). If None,
- get UA from embedder (windowing) function, If None,
- get the default UA

atouchet:
This was finished in #26103.

Issue #7492, opened 2015-09-01
Implement DOM selections


jdm:
I propose that we store the bounds of our selection in `Document` using a starting node, an ending node, a starting offset, and an ending offset. Additionally, we should store a vector of these to enable multiple simultaneous selections in the future, and send all of this data along with each script-initiated reflow request. When a node is part of a selection, it should have a flag set on it (c.f. [NodeFlags](http://mxr.mozilla.org/servo/source/components/script/dom/node.rs#141)); when a node with this flag is encountered, the layout code should check whether it's the starting or ending node; if it is not, we can go ahead and style the node entirely as selected. If it is, we need to deal with styling the node content in up to three separate parts (pre-selection, in selection, post-selection).

cc @aweinstock314

nox:
What if we use a DOM range for the selection boundaries?

jdm:
Seems reasonable at first glance.

jdm:
#4409 is related and I can never find it.

jdm:
@iamrohit7 If you're interested in trying this, I will do my best to advise you!

iamrohit7:
@jdm Thanks. I'll get back to you with any questions.

pshaughn:
WPT custom-elements/reactions/Selection needs this.

pshaughn:
Most failures under dom/ranges are also this.

pshaughn:
Would implement DOM selections, up to a point that satisfies non-webdriver automated tests, be a productive thing to do in itself? I imagine they're not very useful without also having a UI to select things, but on the other hand that UI might be easier to write as its own piece on top of the already-working DOM part.

jdm:
That would be a useful starting point, indeed.

pshaughn:
Here's my understanding of what https://w3c.github.io/selection-api/ tells us to do.

Section 2:

- This is mostly preamble but it points out that Document.open resets the selection when it resets the document, which isn't spelled out anywhere in the meatier sections.

Section 3:

- Selection is a new DOM object, which can have references to a range and a couple nodes.
- It has many methods that modify or test the range; these are the bulk of the new code to add, but most of them are straightforward.
- It can return a live reference to the range and be affected by subsequent changes to it.

Section 4:

- Document gets a DomRefCell<Selection>, which is populated with an empty selection when the document's created.
- Document.getSelection returns the selection if there's a browsing context for the document, or null if not.
- Window.getSelection calls Document.getSelection.
[...]

Issue #25523, opened 2020-01-14
Intermittent linux timeouts in WPT tests


jdm:
```
  ▶ TIMEOUT [expected OK] /WebCryptoAPI/getRandomValues.any.worker.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /referrer-policy/gen/worker-classic.http-rp/same-origin/xhr/same-https.swap-origin.http.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /workers/examples/fetch_tests_from_worker.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /_webgl/conformance/attribs/gl-vertex-attrib-unconsumed-out-of-bounds.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /_webgl/conformance/ogles/GL/build/build_025_to_032.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /html/interaction/focus/the-autofocus-attribute/queue-non-focusable.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /_webgl/conformance/attribs/gl-vertex-attrib-zero-issues.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /offscreen-canvas/the-offscreen-canvas/size.attributes.reflect.setidlzero.worker.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /dom/events/Event-dispatch-target-moved.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /referrer-policy/gen/top.meta/strict-origin-when-cross-origin/fetch/same-http.keep-origin.http.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /referrer-policy/gen/top.meta/strict-origin/fetch/cross-https.swap-origin.http.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected OK] /offscreen-canvas/line-styles/2d.line.miter.acute.worker.html
  │ 
  │ 
  │ 
  └ 


  ▶ TIMEOUT [expected ERROR] /WebCryptoAPI/generateKey/successes_RSA-OAEP.https.any.html?51-60
  │ 
  │ 
  │ 
[...]

PR #25674, opened 2020-02-02
Selection DOM interface (but not actual UI selections)

pshaughn:
This is work towards #7492.

I put new tests in the mozilla-specific directory rather than the main tree because I'm not sure how upstreamable they are; I'd like opinions on that point.

This adds a new exposed interface, which I understand from tests/mozilla/interfaces.html is something requiring specific approval from someone allowed to give that approval.

Things that aren't done here:
- The spec doesn't describe what selection/script-and-style-elements.html wants, and what it wants seems to require some sort of layout query.
- Actual UI interactions are not present at all; selection.rs has a few TODOs saying which methods I believe the eventual UI code should call for what actions, but there are a lot of fine points here that I don't know about.
- I haven't touched Node; you can ask if a node's in the Selection's Range the same way you'd ask about any other Range, but there's not a faster path just for selection layout.


components/script/dom/selection.rs
```diff
@@ -0,0 +1,523 @@
...
+            return;
+        }
+        let this = Trusted::new(self);
+        let window = window_from_node(&*self.document);
+        window
+            .task_manager()
+            .dom_manipulation_task_source()
+            .queue(
+                task!(selectionchange_task_steps: move || {
+                let this = this.root();
```
pshaughn: this is indented wrong, weird that tidy/fmt didn't notice

components/script/dom/selection.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    // https://w3c.github.io/selection-api/#dom-selection-selectallchildren
+    fn SelectAllChildren(&self, node: &Node) -> ErrorResult {
+        if node.is_doctype() {
+            // Range::new assumes its caller already checked this,
+            // WPT implies the check comes before the same-root check
+            return Err(Error::InvalidNodeType);
+        }
+        if !self.is_same_root(node) {
+            // Spec doesn't say to empty here, but WPT tests expect it.
```
pshaughn: This comment is in error, just noting that here to remind myself

components/script/dom/selection.rs
```diff
@@ -0,0 +1,523 @@
...
+            return false;
+        }
+    }
+
+    // https://w3c.github.io/selection-api/#dom-selection-stringifier
+    fn Stringifier(&self) -> DOMString {
+        // The spec as of Jan 31 2020 just says
+        // "See W3C bug 10583." for this method.
+        // Stringifying the range seems at least approximately right
+        // and passes the one case in the tentative WPT test.
```
pshaughn: There's two cases; we fail the other one because it cares about CSS state

components/script/dom/selection.rs
```diff
@@ -0,0 +1,523 @@
...
+        if anchor_offset > anchor_node.len() || focus_offset > focus_node.len() {
+            return Err(Error::IndexSize);
+        }
+
+        // Step 2
+        if !self.is_same_root(anchor_node) || !self.is_same_root(focus_node) {
+            return Ok(());
+        }
+
+        if anchor_node.is_doctype() || focus_node.is_doctype() {
```
pshaughn: Looking at the other methods that are doing this, this should probably be before the IndexSize check, not after the same_root one.

pshaughn:
Nothing too unexpected here... 

▶ Unexpected subtest result in /custom-elements/reactions/Selection.html:
  └ PASS [expected FAIL] deleteFromDocument on Selection must enqueue a disconnected reaction
  ▶ Unexpected subtest result in /html/browsers/the-window-object/window-properties.https.html:
  └ PASS [expected FAIL] Window method: getSelection
  ▶ Unexpected subtest result in /dom/ranges/Range-mutations-deleteData.html:
[many cases, probably the entire file like the other Range-mutations-* ones]
▶ Unexpected subtest result in /_mozilla/mozilla/interfaces.html:
  │ FAIL [expected PASS] Interfaces exposed on the window
  │   → assert_true: If this is failing: DANGER, are you sure you want to expose the new interface Selection to all webpages as a property on the global? Do not make a change to this file without review from jdm or Ms2ger for that specific change! expected true got false

pshaughn:
```
  ▶ TIMEOUT [expected OK] /2dcontext/shadows/shadowBlur_gaussian_tolerance.1.html
  │ 
  │ 
  │ 
  └ 
```
I've never seen this one before. It looks like the test is iterating over many of image pixels one at a time in Javascript, so maybe it's susceptible to timeouts if it's not getting as many CPU cycles. It isn't touching selections or ranges.

CYBAI:
@pshaughn I guess that's #25523 👀

components/script/dom/selection.rs
```diff
@@ -0,0 +1,522 @@
...
+                return Err(Error::InvalidNodeType);
+            }
+
+            if offset > node.len() {
+                // As with is_doctype, not explicit in selection spec steps
+                // implied by range-setting
+                return Err(Error::IndexSize);
+            }
+
+            // let is_old_anchor_before_or_equal;
```
jdm: Why is this here?

components/script/dom/selection.rs
```diff
@@ -0,0 +1,522 @@
...
+    // https://w3c.github.io/selection-api/#dom-selection-setbaseandextent
+    fn SetBaseAndExtent(
+        &self,
+        anchor_node: &Node,
+        anchor_offset: u32,
+        focus_node: &Node,
+        focus_offset: u32,
+    ) -> ErrorResult {
+        // Step 1
+        if anchor_node.is_doctype() || focus_node.is_doctype() {
```
jdm: Presumably this check comes from https://dom.spec.whatwg.org/#concept-range-bp-set? I'm a bit concerned that we're adding these extra checks in various places in the code; is our implementation missing them elsewhere, or is this a spec inconsistency, or are we working around tests that assume a certain order or checks that doesn't match what the spec says?

jdm (changes requested):
Great work!

components/script/dom/selection.rs
pshaughn: Tests, which Firefox and Chrome both pass in FYI, are assuming a certain order that doesn't match what the spec says.

components/script/dom/selection.rs
pshaughn: Just forgetfulness.

components/script/dom/selection.rs
pshaughn: I brought it up here: https://github.com/w3c/selection-api/issues/118

jdm:
@bors-servo retry
* #23290

jdm:
@bors-servo retry
* #23290

PR #25685, opened 2020-02-04
What if we unsilence all the "Why are errors silenced here?" errors?

pshaughn:
This might reveal interesting race conditions, or just let us clean out some unnecessary FIXMEs.

Worth noting: the dispatch_event_with_target ones were just an outdated comment; nothing there IS an error, since the events are uncancelable and the return value's always going to be NotCanceled.

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -555,8 +555,8 @@ impl Document {
                     );
                     let event = event.upcast::<Event>();
                     event.set_trusted(true);
-                    // FIXME(nox): Why are errors silenced here?
-                    let _ = window.dispatch_event_with_target_override(
+
+                    window.dispatch_event_with_target_override(
                         &event,
                     );
                 }),
@@ -1864,7 +1864,7 @@ impl Document {
             );
             let event = event.upcast::<Event>();
             event.set_trusted(true);
-            let _ = self.window.dispatch_event_with_target_override(&event);
+            self.window.dispatch_event_with_target_override(&event);
             // TODO Step 6, document visibility steps.
         }
         // Step 7
@@ -1878,7 +1878,7 @@ impl Document {
             event.set_trusted(true);
             let event_target = self.window.upcast::<EventTarget>();
             let has_listeners = event.has_listeners_for(&event_target, &atom!("unload"));
-            let _ = self.window.dispatch_event_with_target_override(&event);
+            self.window.dispatch_event_with_target_override(&event);
             self.fired_unload.set(true);
             // Step 9
             if has_listeners {
@@ -1973,8 +1973,7 @@ impl Document {
                     update_with_current_time_ms(&document.load_event_start);
 
                     debug!("About to dispatch load for {:?}", document.url());
-                    // FIXME(nox): Why are errors silenced here?
-                    let _ = window.dispatch_event_with_target_override(
+                    window.dispatch_event_with_target_override(
                         &event,
                     );
 
@@ -2017,8 +2016,7 @@ impl Document {
                         let event = event.upcast::<Event>();
                         event.set_trusted(true);
 
-                        // FIXME(nox): Why are errors silenced here?
-                        let _ = window.dispatch_event_with_target_override(
+                        window.dispatch_event_with_target_override(
                             &event,
                         );
                     }),
@@ -4223,6 +4221,7 @@ impl DocumentMethods for Document {
 
         let url = self.url();
         let (tx, rx) = profile_ipc::channel(self.global().time_profiler_chan().clone()).unwrap();
+        // FIXME: Why is send silencing errors but recv unwrapping them?
         let _ = self
             .window
             .upcast::<GlobalScope>()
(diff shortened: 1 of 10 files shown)


pshaughn:
Just a regular reftest intermittent.

components/script/dom/eventsource.rs
```diff
@@ -105,17 +105,19 @@ impl EventSourceContext {
...
-            task!(announce_the_event_source_connection: move || {
-                let event_source = event_source.root();
-                if event_source.ready_state.get() != ReadyState::Closed {
-                    event_source.ready_state.set(ReadyState::Open);
-                    event_source.upcast::<EventTarget>().fire_event(atom!("open"));
-                }
-            }),
-            &global,
-        );
+        global
```
gterzian: This is called from the IPC router thread, and I wouldn't assert here unless we add some guarantees that it is shutdown when the other threads in the process shutdown.

components/script/dom/eventsource.rs
```diff
@@ -138,40 +140,43 @@ impl EventSourceContext {
...
-                        event_source: trusted_event_source,
-                        action_sender,
-                    }
-                );
-                // FIXME(nox): Why are errors silenced here?
-                let _ = event_source.global().schedule_callback(callback, duration);
-            }),
-            &global,
-        );
+        global
```
gterzian: ditto

components/script/dom/eventsource.rs
```diff
@@ -247,16 +252,18 @@ impl EventSourceContext {
...
-        let _ = global.remote_event_task_source().queue(
-            task!(dispatch_the_event_source_event: move || {
-                let event_source = event_source.root();
-                if event_source.ready_state.get() != ReadyState::Closed {
-                    event.root().upcast::<Event>().fire(&event_source.upcast());
-                }
-            }),
-            &global,
-        );
+        global
```
gterzian: ditto

components/script/dom/window.rs
```diff
@@ -1968,17 +1968,18 @@ impl Window {
...
-                    ScriptThreadEventCategory::DomEvent,
-                    Box::new(
-                        self.task_manager
-                            .task_canceller(TaskSourceName::DOMManipulation)
-                            .wrap_task(task),
-                    ),
-                    self.pipeline_id(),
-                    TaskSourceName::DOMManipulation,
-                ));
+                self.script_chan
```
gterzian: This can actually be replaced by the syntax using `dom_manipulation_task_source`

components/script/dom/window.rs
```diff
@@ -2464,19 +2465,20 @@ impl Window {
...
-            ScriptThreadEventCategory::DomEvent,
-            Box::new(
-                self.task_manager
-                    .task_canceller(TaskSourceName::DOMManipulation)
-                    .wrap_task(task),
-            ),
-            self.pipeline_id(),
-            TaskSourceName::DOMManipulation,
-        ));
+        self.script_chan
```
gterzian: This can also use the `dom_manipulation_task_source` syntax.

gterzian (changes requested):
I think this is a good idea, with the exception of task queuing done from the IPC router thread, because currently there is no guarantee that it shutdowns before the window event-loops, or other threads, that use it(while queuing a task from a task should indeed normally work). 

I will actually open a PR in https://github.com/servo/ipc-channel/ that wil add an explicit shutdown mechanism, and then open a PR here that uses it to shutdown this IPC router thread as part of the overall shutdown workflow, and then it would be a good idea to have to those assertions on the IPC thread as well. 

FYI the calls from the IPC router thread are made by thos `_Context` structs, since they are used as part of `ROUTER.add_route`, and in the spec they will usually be identified as task queued from "in parallel" steps(but not always).

jdm:
I agree with gterzian!

gterzian:
After this https://github.com/servo/servo/pull/26036 we can actually move forward with this one!

components/script/dom/window.rs
gterzian: This comment is still relevant

components/script/dom/window.rs
gterzian: This comment is still relevant

gterzian:
Ok so now that we're shutting down the IPC router, this one is basically good to go I think. 

There are two comments that I noted are still relevant, not completely related but we might as well do it here...

components/script/dom/eventsource.rs
mrobinson: I think that after #34827 we will be able to handle this case by not queuing the task if the canceller has been canceled and just ensure that all cancellers are canceled before the `GlobalScope` is dropped.

mrobinson:
Closing this in favor of #34849.

components/script/dom/document.rs
```diff
@@ -4248,6 +4247,8 @@ impl DocumentMethods for Document {
             vec![]
         };
 
+        // FIXME: Why are we silencing errors here? Is it really okay to return
```
pshaughn, line 4250: I'm not really involved with this project anymore, and please forgive me if I'm just making unproductive noise, but I notice this observation here wasn't brought up in #34849 and I never did find out whether or not it was meaningful.

mrobinson: Good point. There are more cases where we are silencing errors where it might not make sense. This might happen if, for instance, the resource thread ended unexpectedly. We need to think more deeply about what should happen in those kind of cases.

PR #25694, opened 2020-02-05
Remove dependency of constellation on canvas

### LINKED ISSUE #24891: Constellation shouldn't have to depend on canvas crate
After #24875, constellation depends on canvas in order to start the canvas paint thread, and to send constellation-specific messages to the thread. If the message moves into canvas_traits and we start the thread in the code that creates the constellation in components/servo, we should be able to remove the dependency.

kunalmohan:
move `ConstellationCanvasMsg` to canvas_traits and start canvas paint thread
to components/servo. This, however, does not remove dependency for conditional
compilation options.

https://github.com/servo/servo/blob/5f55cd5d71df9c555fbc24777168396ddd539f28/components/constellation/Cargo.toml#L13-L15

---

- [X] These changes fix #24891  (GitHub issue number if applicable)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -534,6 +534,7 @@ dependencies = [
 name = "canvas_traits"
 version = "0.0.1"
 dependencies = [
+ "crossbeam-channel",
  "cssparser",
  "euclid",
  "ipc-channel",
@@ -806,7 +807,6 @@ dependencies = [
  "background_hang_monitor",
  "backtrace",
  "bluetooth_traits",
- "canvas",
  "canvas_traits",
  "compositing",
  "crossbeam-channel",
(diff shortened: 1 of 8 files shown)


Manishearth (changes requested):
We should actually remove the dependency and then do the feature threading in libservo instead.

kunalmohan:
I am a bit new to this, so would need a bit more information/guidance. Are you suggesting `libservo/canvas2d-azure` under features instead?

kunalmohan:
I just found that `libservo/canvas2d-azure` isn't possible since it would create a cyclic dependency.

Manishearth:
@kunalmohan no, in the libservo crate we add `canvas2d-azure = ["canvas/canvas2d-azure"]` as a feature

kunalmohan:
Oh, ok. That's already done in libservo. I guess that's what confused me a bit. I'll just remove the features section under `constellation/Cargo.toml` then.

Manishearth:
Yep

Manishearth:
@bors-servo retry

 - network issue

PR #25695, opened 2020-02-05
Let EventSource fail fast on bad schemes

### LINKED ISSUE #25692: EventSource constructor enters wrong readyState on non-same-origin urls with non-http(s) schemes
eventsource/eventsource-constructor-non-same-origin.htm expects an EventSource with an about, mailto, javascript, etc. url to end up in readyState CLOSED when it throws an error; the error is thrown but the readyState is incorrect.

Multiple cases are failing. One case currently passes by coincidence and will fail when #25687 fixes other behavior.

pshaughn:
EventSource went into an infinite reconnect loop in some cases where tests wanted it to go into a hard error state; this addresses the cases where that happens because the url isn't even http(s) and will thus definitely never result in an event stream.

https://github.com/web-platform-tests/wpt/issues/4311 suggests the tests might just be too picky here; the spec does use the word "may" on relevant behavior.

---

- [X] These changes fix #25692 


components/script/dom/eventsource.rs
```diff
@@ -472,7 +483,7 @@ impl EventSource {
                 let event_source = event_source.root();
                 if event_source.ready_state.get() != ReadyState::Closed {
                     event_source.ready_state.set(ReadyState::Closed);
-                    event_source.upcast::<EventTarget>().fire_event(atom!("error"));
+            event_source.upcast::<EventTarget>().fire_event(atom!("error"));
```
CYBAI: Heh, I'm surprised fmt didn't complain about this 🤔 looks like we should keep it with original indentation 👀?

pshaughn: This keeps happening to me in particular! My workflow is to just type stuff in emacs without thinking about indents, then run mach fmt afterwards.

nox: Are you saying that `fmt` isn't complaining about this reindented line?

nox:
Seems ok to me, r=me if fmt is happy.

@bors-servo delegate+

pshaughn:
hold on, I didn't fix the indentation yet

components/script/dom/eventsource.rs
pshaughn: Yeah, it seems to give up when deep enough in a task! macro.

pshaughn:
@bors-servo retry
#24726

pshaughn:
@bors-servo retry
more of the same

PR #25677, opened 2020-02-03
Initial implementation of GPUShaderModule

imiklos:
Added WebIDL bindings for `GPUShaderModule`.
Implemented the `createShaderModule` function of `GPUDevice`.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes addresses a part of #24706 

    
cc @kvark @jdm @zakorgy

Issue #25472, opened 2020-01-09
Try making creation of GPUBuffer async(even without a promise)


gterzian:
When a buffer is created at https://github.com/servo/servo/blob/4ae3c8bc6d8a3af51ab3e78464f93e0ea592c51c/components/script/dom/gpudevice.rs#L166

The script-thread blocks on the IPC response from the WebGPU component at https://github.com/servo/servo/blob/4ae3c8bc6d8a3af51ab3e78464f93e0ea592c51c/components/script/dom/gpudevice.rs#L191

Instead, I think you could make it non-blocking. 

First of all, why doesn't the spec use a promise? 

Secondly, even without a promise, you can still do it(I think), in the following way:

1. Use a pattern similar to [`response_async`](https://github.com/servo/servo/blob/4ae3c8bc6d8a3af51ab3e78464f93e0ea592c51c/components/script/dom/gpu.rs#L67), but without the promise(Basically just use an IPC route and a `Trusted<GPUBuffer>`). 
2. Make `GPUBuffer.buffer` optional inside a dom refcell.
3. Add a `DomRefCell<VecDeque<BufferCommands>>` to `GPUBuffer`. 

With `BufferCommand` being something like:
```rust
enum BufferCommand {
    UnMap,
    Destroy,
}
```

Then, when you create the buffer, just return it immediately without blocking on the IPC, setting the buffer to `None`. 

As long as the buffer is none, you queue any `WebGPURequest` as a corresponding `BufferCommand`.

When the buffer is received from WebGPU, you, via a task via a task queued from the IPC route(similar to `response_async`):
- Set the buffer to some.
- Drain the buffer of `BufferCommand` and send corresponding `WebGPURequest` in order to WebGPU. 

This would appear spec compliant in the light of the note at https://gpuweb.github.io/gpuweb/#buffer-destruction

> Note: This allows the user agent to reclaim the GPU memory associated with the GPUBuffer once all previously submitted operations using it are complete.

Which I read as saying that you can indeed queue requests, even if one of them is the destruction of the buffer, so long as you submit them in order at some point to WebGPU for processing. 

@kvark @zakorgy

kvark:
> First of all, why doesn't the spec use a promise?

[...]

PR #25700, opened 2020-02-06
Initial implementation of GPUComputePipeline

imiklos:
Added WebIDL bindings for `GPUComputePipeline`.
Implemented the `createComputePipeline` function of `GPUDevice`.
This pull request is based on: #25677 

---

cc @kvark @jdm @zakorgy

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -98,7 +98,10 @@ use std::sync::Arc;
 use time::{get_time, Timespec};
 use uuid::Uuid;
 use webgpu::wgpu::{
-    id::{AdapterId, BindGroupId, BindGroupLayoutId, BufferId, DeviceId, PipelineLayoutId},
+    id::{
+        AdapterId, BindGroupId, BindGroupLayoutId, BufferId, DeviceId, PipelineLayoutId,
+        ShaderModuleId,
+    },
     Backend,
 };
 
@@ -2127,6 +2130,12 @@ impl GlobalScope {
         self.gpu_id_hub
             .borrow_mut()
             .create_pipeline_layout_id(backend)
+    }
+
+    pub fn wgpu_create_shader_module_id(&self, backend: Backend) -> ShaderModuleId {
+        self.gpu_id_hub
+            .borrow_mut()
+            .create_shader_module_id(backend)
     }
 }
 
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -99,8 +99,8 @@ use time::{get_time, Timespec};
 use uuid::Uuid;
 use webgpu::wgpu::{
     id::{
-        AdapterId, BindGroupId, BindGroupLayoutId, BufferId, DeviceId, PipelineLayoutId,
-        ShaderModuleId,
+        AdapterId, BindGroupId, BindGroupLayoutId, BufferId, ComputePipelineId, DeviceId,
+        PipelineLayoutId, ShaderModuleId,
     },
     Backend,
 };
@@ -2136,6 +2136,11 @@ impl GlobalScope {
         self.gpu_id_hub
             .borrow_mut()
             .create_shader_module_id(backend)
+    }
+    pub fn wgpu_create_compute_pipeline_id(&self, backend: Backend) -> ComputePipelineId {
+        self.gpu_id_hub
+            .borrow_mut()
+            .create_compute_pipeline_id(backend)
     }
 }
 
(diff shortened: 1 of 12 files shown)


components/script/dom/globalscope.rs
```diff
@@ -2128,6 +2131,17 @@ impl GlobalScope {
...
             .create_pipeline_layout_id(backend)
     }
+
+    pub fn wgpu_create_shader_module_id(&self, backend: Backend) -> ShaderModuleId {
+        self.gpu_id_hub
+            .borrow_mut()
+            .create_shader_module_id(backend)
+    }
+    pub fn wgpu_create_compute_pipeline_id(&self, backend: Backend) -> ComputePipelineId {
+        self.gpu_id_hub
```
kvark: A bit surprised you are able to use the `RefCell` here. In Gecko implementation, multiple threads in the same content process may want to create their own resource IDs.

components/script/dom/globalscope.rs
```diff
@@ -2128,6 +2131,17 @@ impl GlobalScope {
             .borrow_mut()
             .create_pipeline_layout_id(backend)
     }
+
+    pub fn wgpu_create_shader_module_id(&self, backend: Backend) -> ShaderModuleId {
```
kvark: nit: perhaps it would be simpler to just have a function returning `&mut GpuIdHub` and let the users call the methods appropriately?

components/script/dom/gpudevice.rs
```diff
@@ -523,9 +529,74 @@ impl GPUDeviceMethods for GPUDevice {
...
+        &self,
+        descriptor: RootedTraceableBox<GPUShaderModuleDescriptor>,
+    ) -> DomRoot<GPUShaderModule> {
+        let (sender, receiver) = ipc::channel().unwrap();
+        let program: Vec<u32> = match &descriptor.code {
+            Uint32Array(program) => program.to_vec(),
+            String(program) => program
+                .as_bytes()
+                .iter()
+                .map(|c| *c as u32)
```
kvark: oh wow, how is this supposed to work? We are iterating characters of a string and casting them to `u32` words?

components/script/dom/gpudevice.rs
```diff
@@ -523,9 +529,74 @@ impl GPUDeviceMethods for GPUDevice {
...
+            .0
+            .send(WebGPURequest::CreateShaderModule(
+                sender,
+                self.device,
+                id,
+                program,
+            ))
+            .expect("Failed to create WebGPU ShaderModule");
+
+        let shader_module = receiver.recv().unwrap();
```
kvark: In all of the creation methods, we aren't expected to block on anything. Here, therefore, we can't `recv()`. Instead, we'd be creating a new ID and asking the GPU server side to create something associated with this ID, but never wait for it to do so.

components/script/dom/gpudevice.rs
```diff
@@ -523,9 +529,74 @@ impl GPUDeviceMethods for GPUDevice {
...
+    ) -> DomRoot<GPUComputePipeline> {
+        let pipeline = descriptor.parent.layout.id();
+        let program = descriptor.computeStage.module.id();
+        let entry_point = descriptor
+            .computeStage
+            .entryPoint
+            .as_bytes()
+            .iter()
+            .map(|c| *c as i8)
+            .collect::<Vec<i8>>();
```
kvark: why isn't this a `String`?

components/webgpu/lib.rs
```diff
@@ -286,6 +300,50 @@ impl WGPU {
...
+                    layout,
+                    program,
+                    entry,
+                ) => {
+                    let global = &self.global;
+                    let descriptor = wgpu_core::pipeline::ComputePipelineDescriptor {
+                        layout,
+                        compute_stage: wgpu_core::pipeline::ProgrammableStageDescriptor {
+                            module: program,
+                            entry_point: entry.as_ptr(),
```
kvark: what makes it null-terminated?

kvark (changes requested):

components/script/dom/gpudevice.rs
imiklos: I wasn't sure about this one. I've updated the #25677 pull request addressing this in it.

components/script/dom/gpudevice.rs
imiklos: Yes, We have an issue ( #25472 ) for this, we would like to address this in a separate pull request for all the call which has to be non-blocking.

components/script/dom/gpudevice.rs
imiklos: You are right! This should be a `String`.

components/webgpu/lib.rs
imiklos: I've created a `CString` from the entry_point which will be null-terminated then passed it as pointer to the descriptor.

imiklos:
@kvark Thank you for the review! I've addressed what you mentioned, I'm still not sure about the `ShaderModule` creation in case when we are getting it as `String`. I've also rebased this pull request to the #25677.

components/script/dom/globalscope.rs
jdm: @imiklos Please file an issue about this - it sounds like when we create an ID, we could have multiple threads that end up creating the same ID independently. Possible we need to share a single gpu_id_hub between all threads using Arc & Mutex?

jdm (approved):

kvark (approved):

kvark:
>I'm still not sure about the ShaderModule creation in case when we are getting it as String

We basically don't support this (yet). It depends on the shading language discussions...

jdm:
@bors-servo retry
* #24726

PR #25702, opened 2020-02-06
Initial implementation of GPUCommandEncoder

imiklos:
Added WebIDL bindings for `GPUCommandEncoder`, `GPUCommandBuffer`, `GPUComputePassEncoder` and `GPUProgrammablePassEncoder`.
Implemented the `beginComputePass`, `copyBufferToBuffer` and `finish` functions of `GPUCommandEncoder`.
Implemented the `createCommandEncoder` function of `GPUDevice`.
This pull request is based on: #25700 

---

cc @kvark @jdm @zakorgy


components/script/dom/gpucommandencoder.rs
```diff
@@ -0,0 +1,114 @@
...
+            .0
+            .send(WebGPURequest::CommandEncoderFinish(
+                sender,
+                self.encoder.0,
+                // TODO(zakorgy): We should use `_descriptor` here after it's not empty
+                // and the underlying wgpu-core struct is serializable
+            ))
+            .expect("Failed to send CopyBufferToBuffer");
+
+        let buffer = receiver.recv().unwrap();
```
kvark: similarly to object creation, this shouldn't be blocking

components/webgpu/lib.rs
```diff
@@ -81,6 +81,27 @@ pub enum WebGPURequest {
...
     DestroyBuffer(WebGPUBuffer),
+    CreateCommandEncoder(
+        IpcSender<WebGPUCommandEncoder>,
+        WebGPUDevice,
+        // TODO(zakorgy): Serialize CommandEncoderDescriptor in wgpu-core
+        // wgpu::command::CommandEncoderDescriptor,
+        wgpu::id::CommandEncoderId,
+    ),
+    CopyBufferToBuffer(
+        wgpu::id::CommandEncoderId,
```
kvark: These should have named fields. In general, prefer the named fields for enums, unless there are only 1 or 2 values with distinctly obvious meaning

kvark (changes requested):

components/script/dom/gpucommandencoder.rs
imiklos: I've responded here: https://github.com/servo/servo/pull/25700#discussion_r377018724

components/webgpu/lib.rs
imiklos: I will create another pull request addressing this by create `parameter` structs with named fields.

imiklos:
I've just rebased to #25700

jdm (approved):

kvark (approved):

kvark:
@imiklos how far is Servo (with these changes) from being able to run the compute example?

imiklos:
> @imiklos how far is Servo (with these changes) from being able to run the compute example?

We need the `GPUQueue` implementation and we also need `dispatch`, `endPass`, `setBindGroup`, `setPipeline` and `mapReadAsync` functions to be implemented. Basically we are pretty close to run the compute example.

zakorgy:
@kvark the queue implementation is already here, we just have to do a rebase : https://github.com/szeged/servo/commits/wgpu_queue . 
Also the `dispatch, endPass, setBindGroup, setPipeline` are already WIP, but we are not sure if our current approach is what we need there: https://github.com/szeged/servo/commits/szeged-wgpu
What we do is recording the compute commands in the encoder https://github.com/szeged/servo/blob/szeged-wgpu/components/script/dom/gpucomputepassencoder.rs#L29 and sending those to the WebGPU thread where we process those commands https://github.com/szeged/servo/blob/szeged-wgpu/components/webgpu/lib.rs#L447

kvark:
@zakorgy that's not the intended use, but it's still a valid one. And if you can get it to work by tomorrow (westcoast time), we'll be happy to report your progress on the face-to-face meeting that's happening in Redmond.

The intended use is via `RawPass`: you create one with `new_compute`, you encode the stuff on the client, and then finish it and turn into a byte vector that you can send over the IPC: https://github.com/gfx-rs/wgpu/blob/557ed0674c6d4cfa300d621e3fb7fc363f3d5055/wgpu-core/src/command/compute.rs#L47

zakorgy:
@kvark using `RawPass` was my first idea, but I thought  that's for c->rust communication and would be an overhead for rust->rust communication. But whit your approval I will change our logic according to that.

kvark:
Yeah, you are right, `RawPass` main value comes from a good FFI story, esp with regards to IPC.
However, we figured that having it used uniformly by all the users of `wgpu-core` would be good for consistency. Thus, both Gecko and wgpu-rs use `RawPass`, and so is advised for Servo.

jdm:
@bors-servo retry
* network

jdm:
@bors-servo retry
* network

jdm:
@bors-servo retry
* #24726 and network failure

jdm:
@bors-servo retry
* #24726 #24622

jdm:
@bors-servo retry
* #25742 #23290

PR #25675, opened 2020-02-02
Add usability checks to CanvasRenderingContext2D.drawImage()

pylbrecht:
These changes add the usability checks performed on an image to be drawn to the canvas.

References:
- https://html.spec.whatwg.org/multipage/canvas.html#drawing-images
- https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix part of #25331

- [x] There are tests for these changes

PR #25710, opened 2020-02-07
Make CanvasRenderingContext2D.createPattern() return null for incomplete images

pylbrecht:
`createPattern()` should return `null` if the passed `image` argument is not usable.

References:
- https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-createpattern
- https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument
---

### DIFF
--- a/components/script/canvas_state.rs
+++ b/components/script/canvas_state.rs
@@ -895,12 +895,14 @@ impl CanvasState {
         global: &GlobalScope,
         image: CanvasImageSource,
         mut repetition: DOMString,
-    ) -> Fallible<DomRoot<CanvasPattern>> {
+    ) -> Fallible<Option<DomRoot<CanvasPattern>>> {
         let (image_data, image_size) = match image {
             CanvasImageSource::HTMLImageElement(ref image) => {
-                // https://html.spec.whatwg.org/multipage/#img-error
-                // If the image argument is an HTMLImageElement object that is in the broken state,
-                // then throw an InvalidStateError exception
+                // https://html.spec.whatwg.org/multipage/#check-the-usability-of-the-image-argument
+                if !image.is_usable()? {
+                    return Ok(None);
+                }
+
                 image
                     .get_url()
                     .and_then(|url| {
@@ -933,13 +935,13 @@ impl CanvasState {
         }
 
         if let Ok(rep) = RepetitionStyle::from_str(&repetition) {
-            Ok(CanvasPattern::new(
+            Ok(Some(CanvasPattern::new(
                 global,
                 image_data,
                 image_size,
                 rep,
                 self.is_origin_clean(image),
-            ))
+            )))
         } else {
             Err(Error::Syntax)
         }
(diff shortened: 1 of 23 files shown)


pylbrecht:
These changes are based on https://github.com/servo/servo/pull/25675.

pshaughn:
```
  ▶ Unexpected subtest result in /2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html:
  │ FAIL [expected PASS] Canvas test: 2d.pattern.image.broken
  │   → assert_throws_dom: function "function() { ctx.createPattern(img, 'repeat'); }" did not throw
  │ 
  │ @http://web-platform.test:8000/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html:23:1
  │ _addTest/</<@http://web-platform.test:8000/2dcontext/resources/canvas-tests.js:66:15
  └ _addTest/<@http://web-platform.test:8000/2dcontext/resources/canvas-tests.js:63:11
```
Looks like this test is expecting a throw on an image it thinks is broken, and possibly getting the new null instead?

pylbrecht:
`is_usable()` somehow returns *bad*, when it shouldn't, because `create_pattern()` returns before actually trying to fetch the image data, which should fail for `broken.png` and therefore lead to the desired result. Maybe we need to incorporate fetching the image data into `is_usable()` somehow.

It would be very helpful to know how `broken.png` is broken, so we can figure out why our checks fail.
I guess today is about looking at the image's bytes and learning how PNG works.

pylbrecht:
> No one else is passing either, so probably the spec is right and the test is wrong https://wpt.fyi/results/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html?label=experimental&label=master&aligned
I've never considered this an option. It would spare me the debugging though.

pylbrecht:
I wasn't checking for a broken state in case the `current_request`'s image attribute is `None`. Hopefully the rewrite will still pass all the tests. 🤞

pylbrecht:
```
  ▶ Unexpected subtest result in /2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html:
  │ FAIL [expected PASS] Canvas test: 2d.drawImage.broken
  └   → The object is in an invalid state.
```
This is weird. It seems `createPattern()` and `drawImage()` need to be handled differently, although the spec links to the same [usability checks](https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument).

But maybe my implementation of `is_usable()` is missing something.

pylbrecht:
Since looking at other browsers' behavior for reference seems the common thing to do, I'm a little lost on how to proceed.
Should we fail this test (like most other browsers) in favor of passing the `drawImage()` equivalent? Or is implementing two different usability checks for `createPattern()` and `drawImage()` the way to go?

pylbrecht:
It seems `2d.pattern.broken.image.html` is wrong, so I will roll back to the previous changes that fail this test, but pass `2d.drawImage.broken.html`.
See https://github.com/guest271314/wpt/pull/10 for reference.

jdm:
@bors-servo r+
Looks like a good improvement!

jdm:
@bors-servo retry
* #24726 #25720

jdm:
@bors-servo retry
#24726

PR #23670, opened 2019-07-01
Implement checks on Location setters

braddunbar:
So, this is my first pull to servo. I've tried to follow all the directions, but I may have missed or forgotten some of them. If so, please let me know - I'll be happy to fix things up!

As described in #23522, this implements various checks on the Location setters. In particular, for the following components.

* protocol
* host
* hostname
* port
* pathname

I'd also like to add code for throwing a `SyntaxError` when appropriate, but I may do that separately.

I've removed the expectations for the protocol setter tests, but I haven't yet found any failing tests for the other components. Should I write tests for those? If so, should I add them here or is there an upstream that I should contribute to?

Let me know what you think!

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #23522
- [ ] There are tests for these changes

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/23670)

PR #25714, opened 2020-02-09
Add comments linking to the steps of the specs in the checks on Location setters

### LINKED ISSUE #23522: Implement various checks on the Location setters
Spec: https://html.spec.whatwg.org/multipage/history.html#location

See all attributes that say "Can be set, to navigate to the given URL.". 

Code: See all the `Set_` methods at https://github.com/servo/servo/blob/7a5ce048d2e6740c91187c12abebce132f90bfaa/components/script/dom/location.rs#L78

As well as `set_url_component` which is used by those methods: https://github.com/servo/servo/blob/7a5ce048d2e6740c91187c12abebce132f90bfaa/components/script/dom/location.rs#L45

Currently failing tests(and maybe others):
- https://github.com/servo/servo/blob/master/tests/wpt/metadata/html/browsers/history/the-location-interface/location-protocol-setter.html.ini
- https://github.com/servo/servo/blob/master/tests/wpt/metadata/html/browsers/history/the-location-interface/location-protocol-setter-non-broken.html.ini
- https://github.com/servo/servo/blob/master/tests/wpt/metadata/html/browsers/history/the-location-interface/location-protocol-setter-non-broken-weird.html.ini

And the actual tests are found in https://github.com/servo/servo/tree/master/tests/wpt/web-platform-tests/html/browsers/history/the-location-interface

A start was made at https://github.com/servo/servo/pull/23368/files#diff-45df5e6ab4c1d79fae956506be147b62R244

and https://github.com/servo/servo/pull/23368/files#diff-c1196726ff21e3563c663b049214dafeL21

Note the only gotcha is that the `ServoUrl` contains an `Arc`, so you can't manipulate it directly when you do the initial parse to then perform the check, as that will actually change the url of the window. So instead you'll need to get a copy, as is for example done at https://github.com/servo/servo/pull/23368/files#diff-45df5e6ab4c1d79fae956506be147b62R244 (or you need to perform the checks without manipulating the current url)

uxioandrade:
As stated in [https://github.com/servo/servo/pull/23670#issuecomment-512293118](https://github.com/servo/servo/pull/23670#issuecomment-512293118), I added some comments to the checks on Location setters, so that it is easier to follow the steps in specs.

---

- [x] These changes fix #23522 (GitHub issue number if applicable)

- [x] These changes do not require tests because I just added some comments


gterzian:
Thanks! Could you please also squash into one commit?

components/script/dom/location.rs
```diff
@@ -242,7 +256,11 @@ impl LocationMethods for Location {
     // https://html.spec.whatwg.org/multipage/#dom-location-protocol
     fn SetProtocol(&self, value: USVString) -> ErrorResult {
         self.check_same_origin_domain()?;
-        self.set_url_component(value, UrlHelper::SetProtocol);
+        // Step 6: If copyURL's scheme is not an HTTP(S) scheme, then terminate these steps.
+        let scheme = value.split(':').next().unwrap();
```
paulrouget: Isn’t it dangerous to have `unwrap()` here?

UxioAndrade:
Hi @gterzian should I squash all the commits or just the ones before I merged the master branch?

gterzian:
I would say just one(and I think it will still automatically list @braddunbar as co-author, which would be a good thing).

components/script/dom/location.rs
gterzian: Yes you're right, actually per Step 5 of the protocol setter spec, we should return a `Error::Syntax` here, see https://html.spec.whatwg.org/multipage/history.html#dom-location-protocol

UxioAndrade: Would my latest commit fix that?

gterzian: Yes. 

Please run `./mach fmt` to fix https://community-tc.services.mozilla.com/tasks/VcHxks51RSywXhLH0W3dXg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FVcHxks51RSywXhLH0W3dXg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

And please also squash...

components/script/dom/location.rs
```diff
@@ -216,7 +222,10 @@ impl LocationMethods for Location {
     // https://html.spec.whatwg.org/multipage/#dom-location-pathname
     fn SetPathname(&self, value: USVString) -> ErrorResult {
         self.check_same_origin_domain()?;
-        self.set_url_component(value, UrlHelper::SetPathname);
+        // Step 3: If copyURL's cannot-be-a-base-URL flag is set, terminate these steps.
```
gterzian: Actually, can we put this comment, and all others, above the `Ok(())` branch, because the comments all refer to the steps take when the condition is not met.

components/script/dom/location.rs
```diff
@@ -242,9 +256,16 @@ impl LocationMethods for Location {
...
-        Ok(())
-    }
+        // Step 6: If copyURL's scheme is not an HTTP(S) scheme, then terminate these steps.
+        if let Ok(scheme) = value.split(':').next() {
+            if scheme.eq_ignore_ascii_case("http") || scheme.eq_ignore_ascii_case("https") {
+                self.set_url_component(value, UrlHelper::SetProtocol);
+            }
+            Ok(())
+        } else {
+            Err(Error::Syntax)
```
gterzian: Here we can put a comment for Step 5, throw a "SyntaxError" DOMException.

dralley:
@UxioAndrade Are you planning on finishing up the PR?

jdm:
Closing due to inactivity.

PR #25724, opened 2020-02-11
Use rayon for file operations off the main resource thread

### LINKED ISSUE #25723: Simplify File Manager
I think the overall concurrency of the file manager can be simplified, which would provide a stronger foundation for fixes like https://github.com/servo/servo/issues/25226.

- [ ] Use the same [thread pool as fetch](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/resource_thread.rs#L432) for the various file operations [`read_file`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L85), [`select_file`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L137), and [`select_files`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L144). In progress at https://github.com/servo/servo/pull/25724
- [ ] Remove the need for synchronization around the [`FileManagerStore`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L180), and the use of atomics inside of it in [`FileStoreEntry`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L35) 

The first one should be quite easy, I think we don't even need to share the handle to the threadpool, we can instead call `rayon::spawn` and that will use the same thread-pool?  

The second one is more complicated. 

Looking at operations like https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L85

The reason we need to synchronize the  `FileManagerStore` is because in such an operation, we clone the `Arc` to it, and then move it inside a thread. The thread will then essentially call [`get_impl`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L191) to get a **clone** of the `FileImpl` contained in the `FileStoreEntry`. 

So, it would appear to me that instead of moving the entire `Arc<FileManagerStore>` into the thread,  we could simply move this clone of the `FileImpl`. 

So essentially, we can do `get_impl` on the "main thread" of the file manager, and then we do the actual potential reading of disk on the thread-pool. 

Result: only the main thread owns the `FileManagerStore`, and we can remove synchronization around it, and the use of atomics, inside of it. 

What about an operation like [`FileManagerStore.insert`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L214), currently invoked off the main thread as part of [`create_entry`](https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L366) for example in the `select_file` operation?

See https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L310

Again, I think we can do the file operation off the main thread, and then then send the `FileImpl` in a message back to the main thread, which can then insert it in the file manager store. 

The most complicated might be `fetch`, which also gets a clone of the [`FileManager`](FileManagerhttps://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/filemanager_thread.rs#L390)

https://github.com/servo/servo/blob/8e657877379eaabc0540eaeffb49f6f67879432f/components/net/resource_thread.rs#L471

and then uses it to potentially "fetch a file" at https://github.com/servo/servo/blob/e201b1688ba6350a4f5c1df73893e9032dcb6f08/components/net/fetch/methods.rs#L766

So the goal would be instead of passing the `FileManager` to an instance of fetch, we would only pass a `FileImpl`. 

I think this would require moving some fetch logic, like parts of [`scheme_fetch`](https://github.com/servo/servo/blob/e201b1688ba6350a4f5c1df73893e9032dcb6f08/components/net/fetch/methods.rs#L613), to happen "on the main thread", before the fetch is spawned on a separate thread. 

Note that if we can find a way to pass a `FileImpl` to fetch, instead of an arc to the filemanager, we could fix half of https://github.com/servo/servo/issues/25226 (the tests that are currently intermittent).

gterzian:
FIX #25723

---

### DIFF
--- a/components/net/filemanager_thread.rs
+++ b/components/net/filemanager_thread.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::fetch::methods::{CancellationListener, Data, RangeRequestBounds};
+use crate::resource_thread::CoreResourceThreadPool;
 use crossbeam_channel::Sender;
 use embedder_traits::{EmbedderMsg, EmbedderProxy, FilterPattern};
 use headers::{ContentLength, ContentType, HeaderMap, HeaderMapExt};
@@ -24,8 +25,7 @@ use std::mem;
 use std::ops::Index;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{self, AtomicBool, AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex, RwLock};
-use std::thread;
+use std::sync::{Arc, Mutex, RwLock, Weak};
 use url::Url;
 use uuid::Uuid;
 
@@ -72,13 +72,18 @@ enum FileImpl {
 pub struct FileManager {
     embedder_proxy: EmbedderProxy,
     store: Arc<FileManagerStore>,
+    thread_pool: Weak<CoreResourceThreadPool>,
 }
 
 impl FileManager {
-    pub fn new(embedder_proxy: EmbedderProxy) -> FileManager {
+    pub fn new(
+        embedder_proxy: EmbedderProxy,
+        pool_handle: Weak<CoreResourceThreadPool>,
+    ) -> FileManager {
         FileManager {
             embedder_proxy: embedder_proxy,
             store: Arc::new(FileManagerStore::new()),
+            thread_pool: pool_handle,
         }
     }
 
@@ -90,14 +95,19 @@ impl FileManager {
         origin: FileOrigin,
     ) {
         let store = self.store.clone();
-        thread::Builder::new()
-            .name("read file".to_owned())
-            .spawn(move || {
-                if let Err(e) = store.try_read_file(&sender, id, check_url_validity, origin) {
-                    let _ = sender.send(Err(FileManagerThreadError::BlobURLStoreError(e)));
-                }
+        self.thread_pool
+            .upgrade()
+            .and_then(|pool| {
+                pool.spawn(move || {
+                    if let Err(e) = store.try_read_file(&sender, id, check_url_validity, origin) {
+                        let _ = sender.send(Err(FileManagerThreadError::BlobURLStoreError(e)));
+                    }
+                });
+                Some(())
             })
-            .expect("Thread spawning failed");
+            .unwrap_or_else(|| {
+                warn!("FileManager tried to read a file after CoreResourceManager has exited.");
+            });
     }
 
     // Read a file for the Fetch implementation.
@@ -113,7 +123,7 @@ impl FileManager {
         response: &mut Response,
         range: RangeRequestBounds,
     ) -> Result<(), BlobURLStoreError> {
-        self.store.fetch_blob_buf(
+        self.fetch_blob_buf(
             done_sender,
             cancellation_listener,
             &id,
@@ -134,22 +144,36 @@ impl FileManager {
             FileManagerThreadMsg::SelectFile(filter, sender, origin, opt_test_path) => {
                 let store = self.store.clone();
                 let embedder = self.embedder_proxy.clone();
-                thread::Builder::new()
-                    .name("select file".to_owned())
-                    .spawn(move || {
-                        store.select_file(filter, sender, origin, opt_test_path, embedder);
+                self.thread_pool
+                    .upgrade()
+                    .and_then(|pool| {
+                        pool.spawn(move || {
+                            store.select_file(filter, sender, origin, opt_test_path, embedder);
+                        });
+                        Some(())
                     })
-                    .expect("Thread spawning failed");
+                    .unwrap_or_else(|| {
+                        warn!(
+                            "FileManager tried to select a file after CoreResourceManager has exited."
+                        );
+                    });
             },
             FileManagerThreadMsg::SelectFiles(filter, sender, origin, opt_test_paths) => {
                 let store = self.store.clone();
                 let embedder = self.embedder_proxy.clone();
-                thread::Builder::new()
-                    .name("select files".to_owned())
-                    .spawn(move || {
-                        store.select_files(filter, sender, origin, opt_test_paths, embedder);
+                self.thread_pool
+                    .upgrade()
+                    .and_then(|pool| {
+                        pool.spawn(move || {
+                            store.select_files(filter, sender, origin, opt_test_paths, embedder);
+                        });
+                        Some(())
                     })
-                    .expect("Thread spawning failed");
+                    .unwrap_or_else(|| {
+                        warn!(
+                            "FileManager tried to select multiple files after CoreResourceManager has exited."
+                        );
+                    });
             },
             FileManagerThreadMsg::ReadFile(sender, id, check_url_validity, origin) => {
                 self.read_file(sender, id, check_url_validity, origin);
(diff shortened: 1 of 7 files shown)


gterzian:
Ok I think it's ready for review. So I had to move some code around, so that each operation that needs a thread goes through the `FileManager`. 

Context: this is the "easy" part of https://github.com/servo/servo/issues/25723

@Manishearth r?

gterzian:
```3 unexpected results that are NOT known-intermittents:
  ▶ Unexpected subtest result in /FileAPI/url/cross-global-revoke.sub.html:
  │ FAIL [expected PASS] It is possible to revoke same-origin blob URLs from different frames.
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /FileAPI/url/cross-global-revoke.sub.html:
  │ FAIL [expected PASS] It is possible to revoke same-origin blob URLs from a different worker global.
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /FileAPI/url/url-with-xhr.any.worker.html:
  │ FAIL [expected PASS] XHR of a revoked URL should fail
  │   → assert_equals: expected 0 but got 416
  │ FAIL [expected PASS] Appending a query string should cause XHR to fail
  │   → assert_equals: expected 0 but got 416
  │ FAIL [expected PASS] Appending a path should cause XHR to fail
  │   → assert_equals: expected 0 but got 416
  │ 
  │ xhr_should_fail/result2</xhr.onreadystatechange<@http://web-platform.test:8000/FileAPI/url/url-with-xhr.any.js:27:7
  └ Test.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:2049:35
  ▶ Unexpected subtest result in /FileAPI/url/url-with-xhr.any.html:
  │ FAIL [expected PASS] XHR of a revoked URL should fail
  │   → assert_equals: expected 0 but got 416
  │ FAIL [expected PASS] Appending a query string should cause XHR to fail
  │   → assert_equals: expected 0 but got 416
  │ FAIL [expected PASS] Appending a path should cause XHR to fail
  │   → assert_equals: expected 0 but got 416
  │ 
  │ xhr_should_fail/result2</xhr.onreadystatechange<@http://web-platform.test:8000/FileAPI/url/url-with-xhr.any.js:27:7
  └ Test.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:2049:35
```

Interesting, I think we might have been getting those previously "for free" because the revoking happening "too early". Will look more into it.

gterzian:
No I think maybe it was because I wasn't returning an blob error if not file is found...

@bors-servo try=wpt

gterzian:
Needs more looking into...

gterzian:
Ok I switched the commit that needs more work into https://github.com/servo/servo/pull/25740

This one is now ready to review 

@Manishearth r?

components/net/filemanager_thread.rs
```diff
@@ -90,14 +95,14 @@ impl FileManager {
         origin: FileOrigin,
     ) {
         let store = self.store.clone();
-        thread::Builder::new()
-            .name("read file".to_owned())
-            .spawn(move || {
+        self.thread_pool.upgrade().and_then(|pool| {
```
Manishearth: what if the upgrade fails?

components/net/filemanager_thread.rs
```diff
@@ -134,22 +139,22 @@ impl FileManager {
             FileManagerThreadMsg::SelectFile(filter, sender, origin, opt_test_path) => {
                 let store = self.store.clone();
                 let embedder = self.embedder_proxy.clone();
-                thread::Builder::new()
-                    .name("select file".to_owned())
-                    .spawn(move || {
+                self.thread_pool.upgrade().and_then(|pool| {
```
Manishearth: ditto, re: handling upgrade failure

components/net/filemanager_thread.rs
```diff
@@ -171,6 +176,176 @@ impl FileManager {
...
+
+    pub fn fetch_file_in_chunks(
+        &self,
+        done_sender: Sender<Data>,
+        mut reader: BufReader<File>,
+        res_body: ServoArc<Mutex<ResponseBody>>,
+        cancellation_listener: Arc<Mutex<CancellationListener>>,
+        range: RelativePos,
+    ) {
+        self.thread_pool.upgrade().and_then(|pool| {
```
Manishearth: same here

components/net/filemanager_thread.rs
```diff
@@ -171,6 +176,176 @@ impl FileManager {
...
+                        return;
+                    }
+                    let length = {
+                        let buffer = reader.fill_buf().unwrap().to_vec();
+                        let mut buffer_len = buffer.len();
+                        if let ResponseBody::Receiving(ref mut body) = *res_body.lock().unwrap() {
+                            let offset = usize::min(
+                                {
+                                    if let Some(end) = range.end {
+                                        // HTTP Range requests are specified with closed ranges,
```
Manishearth: you can use inclusive rust ranges here if we want to

components/net/resource_thread.rs
```diff
@@ -429,10 +431,116 @@ pub struct CoreResourceManager {
...
+        OP: FnOnce() + Send + 'static,
+    {
+        {
+            let mut state = self.state.lock().unwrap();
+            match &mut *state {
+                ThreadPoolState::Active(num) => {
+                    // Increment number of active workers.
+                    *num += 1;
+                },
+                ThreadPoolState::Inactive(_) => return,
```
Manishearth: Probably should print a warning to the console when this happens? This shouldn't happen, right?

components/net/resource_thread.rs
```diff
@@ -429,10 +431,116 @@ pub struct CoreResourceManager {
...
     swmanager_chan: Option<IpcSender<CustomResponseMediator>>,
     filemanager: FileManager,
-    fetch_pool: rayon::ThreadPool,
+    thread_pool: Arc<CoreResourceThreadPool>,
     certificate_path: Option<String>,
 }
 
+/// The state of the thread-pool used by CoreResource.
+enum ThreadPoolState {
+    /// Pool is inactive, cannot spawn additional work,
```
Manishearth: Should this instead be a struct with `active_threads: u32, is_active: bool`?

Manishearth (approved):
Overall looks good!

components/net/filemanager_thread.rs
gterzian: It would mean the system has exited and a worker was still running after that, and then tried to spawn additional work(fetch for example uses the pool from work running on the pool). So I was thinking of not handling it since there seems to be no way to recover from it.

components/net/resource_thread.rs
gterzian: It can happen, because fetch runs in parallel to the main core resource thread, and could try to read a file in chunks after the main thread is trying to exit.

components/net/filemanager_thread.rs
Manishearth: Right, but we should have a `warn!` or something

components/net/filemanager_thread.rs
gterzian: I've filed an issue at https://github.com/servo/servo/issues/25751 because since this code is only moved around in this PR, I would rather not change it here. I've marked the issue "easy", perhaps you can elaborate further there?

components/net/resource_thread.rs
gterzian: I personally liked the enum, so it's your call...

components/net/filemanager_thread.rs
gterzian: Good idea.

components/net/resource_thread.rs
gterzian: As a side, this is sort of "expected", in the sense that the core resource manager is in the process of exiting, and then fetch would still try to spawn additional work. 

The "the weak cannot be upgraded" sitution noted in the other comment, would be if a worker is running even after the core resource manager has already dropped. 

Do you think we still need to warn here? I would say the return here is perhaps not worth warning about since it's a fairly normal situation.

gterzian:
@Manishearth Ok I have added a commit, to be squashed after further review.

gterzian:
Ok, squashed already actually(changes are small and easy to see I think).

components/net/filemanager_thread.rs
Manishearth: Works.

components/net/resource_thread.rs
Manishearth: Ah, if it's expected then never mind. I thought it would only happen if something went wrong in the shutdown order.

components/net/resource_thread.rs
Manishearth: I'm not a big fan of enums where both variants contain the same payload meaning the same thing.

Manishearth:
r=me when ready

components/net/resource_thread.rs
gterzian: Ok, I will change it then.

components/net/resource_thread.rs
```diff
@@ -429,10 +431,116 @@ pub struct CoreResourceManager {
...
+
+        let state = self.state.clone();
+
+        self.pool.spawn(move || {
+            {
+                let mut state = state.lock().unwrap();
+                if let ThreadPoolState::Inactive(num) = &mut *state {
+                    // Decrement number of active workers and return,
+                    // without doing any work.
+                    *num += 1;
```
gterzian: ooh, that's a bug there, should be a decrement...

gterzian:
@Manishearth Ok I switched to a struct, with methods, you're right it is clearer and less error-prone... Could you take another look?

gterzian:
Just made  a last change, now it's ready for review...

gterzian:
And now fixed a typo...

Manishearth:
@bors-servo r+

Thanks!

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #24726

Issue #25007, opened 2019-12-02
Custom element inside <template> uses registry of owner document


pshaughn:
WPT custom-elements/parser/parser-uses-registry-of-owner-document is testing this; `document.write('<template><my-custom-element></my-custom-element></template>')` is having my-custom-element interpreted in terms of the custom element registry, which it shouldn't because it's inside a template.

pshaughn:
What's supposed to be happening here is that, during parse of a template, new elements are added to a document fragment associated with the "template contents owner document", as described in https://html.spec.whatwg.org/multipage/scripting.html#appropriate-template-contents-owner-document. This document isn't supposed to share a custom element registry with the main document being parsed, and so its registry should be empty during this test, meaning we should have an element with a custom element state of "undefined" and an is value of "my-custom-element". The test then inspects this without moving it into the main document, so it should never be upgraded to the custom element constructor.

pshaughn:
My previous comment was assuming we were hitting the synchronous upgrade during parse; another possibility that needs to be distinguished is that we might be parsing it without upgrading, but then upgrading when we add the template node to the document. If that's the case, then the problem is that we're recursing into the contents of the template's fragment when we iterate over its descendants. Since templates in HTML documents seem to be mostly behaving themselves, I'm leaning towards the parser hypothesis but I haven't made any tests to distinguish the possibilities yet.

pshaughn:
Both things can go wrong, the parser and the append.

custom-elements/parser/parser-uses-registry-of-owner-document.html fails because the parser is synchronously creating the custom element when it should be in the template state.

custom-elements/reactions/Document.html "succeeds" right now because a disconnected node is never getting upgrades, but after the fix of https://github.com/servo/servo/issues/24992 it fails because the innerHTML setter is appending the customizable element and a reaction is enqueing from the append.

I don't know if I can fix the parser one but I think I can fix the append one.

pshaughn:
The parser case is a result of https://github.com/servo/servo/issues/18277, and it looks like https://github.com/servo/servo/pull/18536 was a strong but abandoned move towards solving the parser case both in servo and in html5ever.

pshaughn:
https://w3c.github.io/DOM-Parsing/#the-innerhtml-mixin directs us to https://w3c.github.io/DOM-Parsing/#dfn-fragment-parsing-algorithm and this happens _before_ the step that cares whether we're in a template. The "context element" passed to the fragment parsing algorithm is the template itself.
[...]

PR #25730, opened 2020-02-11
Upgrade nodes to custom elements on insertion, even if they're not connected to a Document

### LINKED ISSUE #24992: Using the is= property in innerHTML isn't using the custom element class
This is test case "innerHTML should instantiate a customized build-in element" in WPT custom-elements/builtin-coverage.html.

### LINKED ISSUE #25004: adding a custom element by tHead, tBody, or caption .innerHTML= doesn't fire a constructed event
WPT custom-elements/reactions/HTMLTableElement catches this, with misleading messages (it fails on the `constructed` event before even reaching the part where it expects `connectedCallback`).

Similar tests pass, in which the _table's_ .innerHTML= is being used to make a caption that already has the custom element inside it. The failures happen when the initially-empty caption (etc.) gets its _own_ innerHTML assigned to.

pshaughn:
After comparing tests to specs, I filed https://github.com/whatwg/dom/issues/833 and made this change to match what I believe the spec should be saying to reflect Firefox/Chrome/Safari shared behavior.

I also filed https://github.com/w3c/DOM-Parsing/issues/61 because this change revealed that custom elements created in a template's innerHTML setter weren't behaving themselves and the spec seemed not to reflect reality about them.

---

- [X] These changes fix #24992 and fix #25004; they are relevant to #25007 but don't fix all the subcases thereof.


pshaughn:
I just noticed the commits; I'll squash after I see these results.

Manishearth (approved):
r=me

pshaughn:
```▶ Unexpected subtest result in /custom-elements/reactions/Document.html:
  │ FAIL [expected PASS] importNode on Document must construct a new custom element when importing a custom element from a template
  │   → assert_array_equals: lengths differ, expected array [] length 0, got ["constructed"] length 1
  │ 
  │ @http://web-platform.test:8000/custom-elements/reactions/Document.html:33:5
  └ test_with_window/</<@http://web-platform.test:8000/custom-elements/resources/custom-elements-helpers.js:19:7```

I think this is a freshly exposed case of #25007; the fact that it has been exposed by this change is informative. I'm going to see if the fix for that makes sens to piggyback onto this.

pshaughn:
Fixed that, I believe; see revised top comment.

pshaughn:
It might be a good idea to see what comes out of https://github.com/whatwg/dom/issues/833 instead of immediately trying to merge this. It could be that every other browser is wrong here.

pshaughn:
These code changes were just based on my opinion of the tests; when the issues mentioned in the initial comment are addressed, the actual code changes needed will be likely to differ from my guesses.

PR #25749, opened 2020-02-12
Add new Servo Logo

malqinneh:
* Added servo-64.png (replacing servo64.png @ 64px)
* Added servo-100.png (replacing doge-tiny.png @ 100px)
* Added servo-500.png
* Added servo-500-icon.ico
* Added servo-1000.png (replacing servo.png @ 1000px)
* Added servo.svg (replacing servo.svg)

---

- [X] These changes do not require tests because they add new logo assets


paulrouget:
Can you also update support/android/apk/servoapp/src/main/res/mipmap/servo.png ?

malqinneh:
> Can you also update support/android/apk/servoapp/src/main/res/mipmap/servo.png ?

Absolutely ✅

paulrouget:
@jdm do we want to land that now?

jdm:
@bors-servo r+
Thanks @malqinneh!

jdm:
```
> /repo/support/android/apk/servoapp/src/main/res/mipmap/servo-1024.png: Error: '-' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
```

malqinneh:
> ```
> > /repo/support/android/apk/servoapp/src/main/res/mipmap/servo-1024.png: Error: '-' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore
> ```

Is this a matter of replacing `-` (dash) with `_` (underscore)?

jdm:
I expect that would work, but we should really overwrite the existing servo.png instead so that the new image gets used automatically.

malqinneh:
> I expect that would work, but we should really overwrite the existing servo.png instead so that the new image gets used automatically.

Makes sense but tried to avoid overwriting original PNG file(s) to keep them safe/available in case they are needed in the future. I suppose one can always go back in commit history for them.

atouchet:
This doesn't appear to have updated the logo on Windows. Servo.ico, servo.png, and servo64.png are still using the old logo.

malqinneh:
> This doesn't appear to have updated the logo on Windows. Servo.ico, servo.png, and servo64.png are still using the old logo.

Probably due to file name(s). In order for the new assets to simply replace the old ones, they need to have the same name as the old assets. Alternatively, we can update the line of code that references old asset names to the new names.

atouchet:
@malqinneh I'm not sure what the best approach would be but I took a quick look and it appears the relevant references are here:

https://github.com/servo/servo/blob/a9f7b132303f7c02648043df5ebdc25726bd3f88/ports/glutin/build.rs#L15

https://github.com/servo/servo/blob/8f3622af35a391b1d15ec1768aa90721ebc5396e/python/servo/build_commands.py#L722

https://github.com/servo/servo/blob/e07f0280fe9c16fbdeb34d488e00e8101f28fc9e/ports/glutin/headed_window.rs#L149

malqinneh:
@atouchet happy to update those items. I need to create a new fork of the Servo branch since the last one I started was merged, is that correct?

atouchet:
That should work. Taking another look at the files in /resources I suspect that servo.icns may be another file that needs updating but it appears to be an Apple image format and I don't have a Mac.

malqinneh:
@atouchet created a [new pull request](https://github.com/servo/servo/pull/25846). Let me know if we missed anything. Thank you very much for your help and patience 👍

atouchet:
Thanks! I will take a look at the PR.

Issue #25718, opened 2020-02-10
Add UWP smoketest to CI


jdm:
It would be nice to avoid problems like #25683 in the future. @ferjm You mentioned you were able to launch the UWP app from the command line in the past; could you explain how you did that so we can see if we can replicate it in CI?

ferjm:
https://github.com/servo/servo/pull/24534 introduced support for launching the UWP from cli with the "Servo" command (from any path). However, IIRC it required launching the app once from VS so the command gets registered.

ferjm:
Oh, according to https://blogs.windows.com/windowsdeveloper/2017/07/05/command-line-activation-universal-windows-apps/#UaF9ZyFfdF9kYhdY.97 the command alias is also registered when the user installs the app.

paulrouget:
So far I managed to open Servo, but I haven’t found a way yet to know of the launch was successful or not:

- use #25661 + a correct certificate (the PR doesn’t appear to have the right hash for `ServoApp_TemporaryKey.pfx`
- install the certificate: `certutil.exe -addstore "Root" .\ServoApp_1.0.0.0_x64_Debug.cer`
- install the package: `Add-AppxPackage -Path .\ServoApp_1.0.0.0_x64_Debug.appxbundle`
- run it: `explorer.exe "shell:appsFolder\MozillaFoundation.FirefoxReality_XXX!App"`

where XXX is the result of `(Get-AppxPackage -Name MozillaFoundation.FirefoxReality).ToString()`.

paulrouget:
There’s a much easier way to open Servo:

* start `fxr://http://example.com`

And to see if it succeeded, it looks easier to open a local http server and wait for Servo to request that page:

* start `fxr://http://localhost:45678`

paulrouget:
To close: `taskkill /f /im servoapp.exe`.

paulrouget:
To summarize, we could have a python script that:

1. install the cert with `certutil.exe`. Could this be done once for all on the CI?
2. install the app via `Add-AppxPackage`
3. start httpd server
4. launch servo: `start fxr://http://localhost:45678`
5. wait until HTTP request or timeout
6. taskkill app
7. uninstall app: `Get-AppxPackage MozillaFoundation.FirefoxReality | Remove-AppxPackage`

paulrouget:
Servo doesn’t appear to like loading localhost: https://github.com/servo/servo/issues/25741

PR #25762, opened 2020-02-14
Adjust UWP packaging for submission (bis)

paulrouget:
This is https://github.com/servo/servo/pull/25661 + a not-password-protected certificate.

It is necessary to sign the app to run a smoketest (https://github.com/servo/servo/issues/25718).

The initial certificate was password protected. This is a test certificate that we will need to be installed on the taskcluster Windows installations. Any reason for this to be a security problem?


paulrouget:
Just realized that the initial PR has more than this one. Continuing the conversation there.

PR #25763, opened 2020-02-14
Add initial support for stacking contexts to layout_2020

mrobinson:
These changes add initial support for stacking contexts, enabling some parts of z-index to work properly.

---

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -6,9 +6,13 @@ use crate::display_list::DisplayListBuilder;
 use crate::fragments::{AnonymousFragment, BoxFragment, Fragment};
 use crate::geom::{PhysicalRect, ToWebRender};
 use gfx_traits::{combine_id_with_fragment_type, FragmentType};
-use std::default::Default;
+use std::cmp::Ordering;
+use std::mem;
+use style::computed_values::float::T as ComputedFloat;
+use style::computed_values::mix_blend_mode::T as ComputedMixBlendMode;
 use style::computed_values::overflow_x::T as ComputedOverflow;
 use style::computed_values::position::T as ComputedPosition;
+use style::computed_values::transform_style::T as ComputedTransformStyle;
 use style::values::computed::Length;
 use webrender_api::units::LayoutVector2D;
 use webrender_api::{ExternalScrollId, ScrollSensitivity, SpaceAndClipInfo, SpatialId};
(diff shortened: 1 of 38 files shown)


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -0,0 +1,369 @@
...
+        }
+    }
+
+    pub(crate) fn sort_stacking_contexts(&mut self) {
+        self.stacking_contexts.sort_by(|a, b| {
+            if a.z_index != 0 || b.z_index != 0 {
+                return a.z_index.cmp(&b.z_index);
+            }
+
+            match (a.context_type, b.context_type) {
```
SimonSapin: Instead of this enum being part of the sort, what do you think of having separate `Vec`s for each categories? (I’m not saying we should do that, I haven’t thought through the advantages or downsides of each approach.)

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -0,0 +1,369 @@
...
+    // from the value specified in the style.
+    fn effective_z_index(&self) -> i32 {
+        match self.style.get_box().position {
+            ComputedPosition::Static => {},
+            _ => return self.style.get_position().z_index.integer_or(0),
+        }
+
+        if !self.style.get_box().transform.0.is_empty() {
+            return self.style.get_position().z_index.integer_or(0);
+        }
```
SimonSapin: My reading of https://drafts.csswg.org/css-transforms/#transform-rendering is that this should be removed:

> For elements whose layout is governed by the CSS box model, any value other than `none` for the `transform` property results in the creation of a stacking context. Implementations must paint the layer it creates, within its parent stacking context, at the same stacking order that would be used if it were a positioned element with `z-index: 0`. If an element with a transform is positioned, the `z-index` property applies as described in [CSS2], except that `auto` is treated as 0 since a new stacking context is always created.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -0,0 +1,369 @@
...
+    space_and_clip: SpaceAndClipInfo,
+    containing_block: PhysicalRect<Length>,
+    fragment: &'a Fragment,
+}
+
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub(crate) enum StackingContextType {
+    Real,
+    PseudoPositioned,
+    PseudoFloat,
```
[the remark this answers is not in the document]
mrobinson: Nice catch! I plan to add support for this kind of pseudo stacking context in a followup change, if that's okay with you. Adding it allows more tests to pass once support is added for sorting the fragments inside a stacking context.

components/layout_2020/display_list/stacking_context.rs
mrobinson: So it's essentially a tradeoff between sorting one slightly larger Vec of stacking contexts versus sorting three or four smaller ones.  I'm not sure which is better, but I've tried to more-or-less copy the design of the old layout system which went through a process of iteration and optimization.

components/layout_2020/display_list/stacking_context.rs
mrobinson: This is quite odd, because Gecko has a strange behavior here that Servo has been replicating. Observe how this particular file is rendered in Gecko:
```
<div style="position: absolute; top: 2px; z-index: 2; background: green; width: 100px; height: 100px;"></div>
<div style="background: blue; transform: rotate(45deg); z-index: 3; width: 100px; height: 100px;"></div>
```
In Gecko, the z-index of the transformed div is taken into account, while in Chromium and WebKit it is not.

SimonSapin: This is arguably a Gecko bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1256980

Another bit of relevant spec is https://drafts.csswg.org/css2/visuren.html#z-index with “Applies to: positioned elements”

mrobinson: Oh, thanks for finding that bug. I've pushed a new version of the branch which removes this, bringing layout_2020 in line with WebKit and Blink.

SimonSapin:
Looks good, thanks!

@bors-servo r+

mrobinson:
```
  ▶ FAIL [expected PASS] /css/CSS2/margin-padding-clear/margin-bottom-042.xht
  │   → /css/CSS2/margin-padding-clear/margin-bottom-042.xht 54a9df64f1476dd12020019d7cf22ac34d727bc0
  │   → /css/CSS2/reference/ref-no-vert-space-between.xht 361532ece0d2a0454900e9de0bb71598ea4a4295
  └   → Screenshot is solid color 0xFFFFFF for /css/CSS2/margin-padding-clear/margin-bottom-042.xht
```

Not sure what could be causing this failure on the non-Layout 2020 test run.

SimonSapin:
> Screenshot is solid color 0xFFFFFF

This is a symptom of https://github.com/servo/servo/issues/24726

@bors-servo retry

PR #25768, opened 2020-02-14
Handle access-control header wildcards

### LINKED ISSUE #24913: CORS response headers are being filtered out too aggressively
WPT fetch/api/cors/cors-expose-star expects to see the nonstandard header "test", since it asked to allow "*", but it's not seeing it.
fetch/api/cors/cors-filtering is expecting to see Content-Length and not seeing it.

pshaughn:
We were checking Access-Control-Expose-Headers for wildcards inconsistently and then discarding the result; this fixes the check and its use, passing the WPT test for having a wildcard there.

---

- [X] These changes fix #24913


jdm:
@bors-servo r+
Thanks!

PR #25770, opened 2020-02-15
supportsSession() -> isSessionSupported()

Manishearth:
r? @jdm

fixes https://github.com/servo/servo/issues/25596

PR #25773, opened 2020-02-15
Always trigger an input sources change event on session creation

Manishearth:
Fixes our behavior to match the spec, and to specifically do it in a way
that makes sense. This is also what Chromium does currently, though I'm
not sure if it does the scheduling the same way.

The spec for this may change, see https://github.com/immersive-web/webxr/issues/961.

This fix, along with https://github.com/servo/servo/pull/25770 , makes
three.js content work on servo. Instead of this fix we can also wait for
https://github.com/mrdoob/three.js/issues/18638 to land (which isn't
certain until we figure out more about
https://github.com/immersive-web/webxr/issues/961 )

Even if https://github.com/immersive-web/webxr/issues/961 decides to
choose the option that obviates this patch, we should probably keep it
for now since there's already content out in the wild relying on this
behavior.

r? @jdm @asajeffrey

### DIFF
--- a/components/script/dom/xr.rs
+++ b/components/script/dom/xr.rs
@@ -284,7 +284,6 @@ impl XR {
                 return;
             },
         };
-
         let session = XRSession::new(&self.global(), session, mode, frame_receiver);
         if mode == XRSessionMode::Inline {
             self.active_inline_sessions
@@ -294,6 +293,9 @@ impl XR {
             self.set_active_immersive_session(&session);
         }
         promise.resolve_native(&session);
+        // https://github.com/immersive-web/webxr/issues/961
+        // This must be called _after_ the promise is resolved
+        session.setup_initial_inputs();
     }
 
     pub fn get_displays(&self) -> Result<Vec<DomRoot<VRDisplay>>, ()> {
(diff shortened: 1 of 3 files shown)


jdm:
```
  ▶ Unexpected subtest result in /webxr/events_input_source_recreation.https.html:
  │ FAIL [expected PASS] Input sources are re-created when handedness or target ray mode changes
  │   → assert_equals: Added length matches expectations expected 1 but got 0
  │ 
  │ validateAdded/<@https://web-platform.test:8443/webxr/events_input_source_recreation.https.html:81:7
  │ Test.prototype.step@https://web-platform.test:8443/resources/testharness.js:2024:25
  │ validateAdded@https://web-platform.test:8443/webxr/events_input_source_recreation.https.html:79:7
  │ onInputSourcesChange/<@https://web-platform.test:8443/webxr/events_input_source_recreation.https.html:28:9
  │ Test.prototype.step@https://web-platform.test:8443/resources/testharness.js:2024:25
  └ onInputSourcesChange@https://web-platform.test:8443/webxr/events_input_source_recreation.https.html:22:7
```

Manishearth:
Oh, probably because we needlessly fire an empty inputsourceschange event at the beginning of the session when there aren't any sources

Manishearth:
Bingo. Just added an empty-vector early return.

@bors-servo r=jdm

Manishearth:
@bors-servo retry
 - https://github.com/servo/servo/issues/24622

PR #25781, opened 2020-02-17
Introduce a new type MaybeUnreflectedDom<T> (fixes #25701)

### LINKED ISSUE #25701: GC zeal completely breaks Servo, it seems
I rebuilt Servo with the following diff:

```diff
--- i/components/script/script_thread.rs
+++ w/components/script/script_thread.rs
@@ -3213,6 +3213,12 @@ impl ScriptThread {
             self.event_loop_waker.as_ref().map(|w| (*w).clone_box()),
         );
 
+        let name = unsafe {
+            let window_jsobject = window.reflector().get_jsobject();
+            std::ffi::CStr::from_ptr((*js::jsapi::JS_GetClass(window_jsobject.get())).name)
+        };
+        assert_eq!(name.to_str().unwrap(), "Window");
+
         // Initialize the browsing context for the window.
         let window_proxy = self.local_window_proxy(
             &window,
```

When running Servo without GC zeal, things just work; but when setting `js.mem.gc.zeal.level` to 14, this happens:

```
assertion failed: `(left == right)`
  left: `"Object"`,
 right: `"Window"` (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at components/script/script_thread.rs:3220)
stack backtrace:
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: servo::backtrace::print
   4: servo::main::{{closure}}
   5: std::panicking::rust_panic_with_hook
   6: rust_begin_unwind
   7: std::panicking::begin_panic_fmt
   8: script::script_thread::ScriptThread::load
   9: script::script_thread::ScriptThread::handle_page_headers_available
  10: std::thread::local::LocalKey<T>::with
  11: <script::dom::servoparser::ParserContext as net_traits::FetchResponseListener>::process_response
  12: script::script_thread::ScriptThread::handle_msg_from_constellation
  13: _ZN6script13script_thread12ScriptThread11handle_msgs17h4a00e887af7d79f8E.llvm.744559770162711036
  14: profile_traits::mem::ProfilerChan::run_with_memory_reporting
  15: _ZN3std10sys_common9backtrace28__rust_begin_short_backtrace17h1fce977ad6d2c705E.llvm.8072929449113315268
  16: _ZN3std9panicking3try7do_call17h540ae4d5baf90e5dE.llvm.1071807046001960883
  17: __rust_maybe_catch_panic
  18: core::ops::function::FnOnce::call_once{{vtable.shim}}
  19: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  20: std::sys::unix::thread::Thread::new::thread_start
  21: _pthread_body
  22: _pthread_start
```

AFAIK, this is extremely wrong.

nox:


nox:
All the failures are either the solid white intermittent issue or a silent crash that happens sometimes.

nox:
That silent crash doesn't even reproduce locally.

jdm:
Could you give me an overview of what this PR is doing? It's not clear to me how all the pieces fit together to fix #25701.

nox:
> Could you give me an overview of what this PR is doing? It's not clear to me how all the pieces fit together to fix #25701.

Sure. What it does is that it roots the boxed not-yet-reflected DOM instance with a different smart pointer type than `Dom<T>`. `MaybeUnreflectedDom<T>` traces the rooted `T` by first checking whether the reflector is set, in which case it proceeds just like `Dom<T>`, otherwise it directly traces the `T`, which will invoke the `JSTraceable` implementation of `Reflector`, which will do nothing but that's not a problem given it's not set yet.

jdm (approved):
This is a nice solution to the problem! Thanks for explaining it to me.

nox:
Not sure if this crash is my fault or not, we'll see how the other build jobs fare.

asajeffrey:
This is getting us pretty close to being able to support lazy creation of DOM reflectors.

nox:
This only works because the `T` here is always the actual concrete type of the DOM instance, if you create an `Element` and you store it as `MaybeUnreflectedDom<Node>`, you won't trace everything that is specific to `Element`.

asajeffrey:
Hmm, so we'd need support for dynamic dispatch of trace methods if we want to support subtyping of maybe-reflected DOM objects.

nox:
Tried to reproduce the issue locally and failed.

@bors-servo retry

jdm:
@bors-servo retry
Network

PR #25788, opened 2020-02-18
Add layout_2020 support for filters and mix-blend-mode

mrobinson:
---


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -64,16 +65,29 @@ pub(crate) struct StackingContext<'a> {
 
     /// All float pseudo stacking context children of this stacking context.
     float_stacking_contexts: Vec<StackingContext<'a>>,
+
+    /// CSS filters to be applied to this stacking context (including opacity).
+    filters: Vec<wr::FilterOp>,
+
+    /// The blend mode with which this stacking context blends with its backdrop.
+    mix_blend_mode: wr::MixBlendMode,
```
SimonSapin: Do these need to be kept in the stacking context tree, or could they be recomputed from the `&ComputedValues` style of the element that generates this stacking context?

mrobinson: I think they can be recomputed. I can try to organize the code in that way.

mrobinson: Okay. I've pushed a new version of this change that minimizes data members in the `StackingContext` data structure.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -414,3 +490,52 @@ impl AnonymousFragment {
         }
     }
 }
+
+pub trait ToLayout {
```
SimonSapin: Could this reuse the `geom::ToWebrender` trait instead? (And perhaps move it to some other module since it’s not geometry-specific. Perhaps `display_list/mod.rs` since that’s the module that interacts most with WebRender?)

tests/wpt/include-layout-2020.ini
```diff
@@ -19,5 +19,7 @@ skip: true
       skip: false
   [css-backgrounds]
     skip: false
+  [filter-effects]
+    skip: false
```
SimonSapin: Also add `css-color` for the `opacity` property?

SimonSapin (approved):
r+ with two changes

components/layout_2020/display_list/stacking_context.rs
mrobinson: Great idea. I've moved this to a "conversions" module that is similar to the one in layout_2013.

tests/wpt/include-layout-2020.ini
mrobinson: Okay. Done.

SimonSapin:
@bors-servo retry

* https://github.com/servo/servo/issues/24726
* `Could not HEAD 'https://jcenter.bintray.com/com/android/databinding/compilerCommon/3.1.3/compilerCommon-3.1.3.jar'. Received status code 504 from server: Gateway Time-out`

PR #25785, opened 2020-02-17
Add support for WebGL2 getFramebufferAttachmentParameter

mmatyas:
Implements the WebGL2 variant of `getFramebufferAttachmentParameter` and the related parameter checks.

See: https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.4

Depends on #25784

cc @jdm @zakorgy

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] There are tests for these changes

PR #25789, opened 2020-02-18
Add support for WebGL2 clear buffer operations

mmatyas:
Adds support for the following WebGL2 methods:

- `clearBufferfv`
- `clearBufferiv`
- `clearBufferuiv`
- `clearBufferfi`

See: https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.11

cc @jdm @zakorgy 

Note: While the code itself doesn't depend on any other PRs, some of the tests do require WebGL2 framebuffer/renderbuffer support (#25785).

---

### DIFF
--- a/components/script/dom/webgl2renderingcontext.rs
+++ b/components/script/dom/webgl2renderingcontext.rs
@@ -557,6 +557,52 @@ impl WebGL2RenderingContext {
         let retval = receiver.recv().unwrap();
         Ok(Int32Value(retval))
     }
+
+    fn clearbuffer_array_size(&self, buffer: u32, draw_buffer: i32) -> WebGLResult<usize> {
+        match buffer {
+            constants::COLOR => {
+                if draw_buffer < 0 || draw_buffer as u32 >= self.base.limits().max_draw_buffers {
+                    return Err(InvalidValue);
+                }
+                Ok(4)
+            },
+            constants::DEPTH | constants::STENCIL | constants::DEPTH_STENCIL => {
+                if draw_buffer != 0 {
+                    return Err(InvalidValue);
+                }
+                Ok(1)
+            },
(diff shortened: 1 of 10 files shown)


components/script/dom/webgl2renderingcontext.rs
```diff
@@ -3096,6 +3114,122 @@ impl WebGL2RenderingContextMethods for WebGL2RenderingContext {
...
+
+    /// https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.11
+    fn ClearBufferfv(
+        &self,
+        buffer: u32,
+        draw_buffer: i32,
+        values: Float32ArrayOrUnrestrictedFloatSequence,
+        src_offset: u32,
+    ) {
+        if ![constants::COLOR, constants::DEPTH].contains(&buffer) {
```
jdm: It looks like we could reduce some duplication with a helper like:
```rust
fn clear_buffer<V>(buffer: u32, draw_buffer: i32, valid_buffers: &[u32], src_offset: u32, array: Vec<V>, msg: fn(u32, i32, Vec<V>) -> WebGLCommand) {
    ...
}
```
And callers could extract the appropriate vector and pass `WebGLCommand::ClearBufferfv` or whatever message constructor is appropriate.

jdm (changes requested):

components/script/dom/webgl2renderingcontext.rs
mmatyas: Ok, just added this change; I was afraid it might look ugly, but turned out to be better than expected!

jdm (approved):

jdm:
This can merge after a rebase.

mmatyas:
Rebased and regenerated the tests.

PR #25799, opened 2020-02-19
Hide window

### LINKED ISSUE #25794: In immersive mode, make sure to mark the window as not visible
I haven't tested yet, but we should make sure, at least in the case of the HL, that the compositor pipeline is set to not visible, to avoid useless animation to run in the background.

paulrouget:
Stop animations when the window is hidden.

Tested with both Immersive Mode, and regular window being minimized.

---

- [x] These changes fix #25794 (GitHub issue number if applicable)

### DIFF
--- a/support/hololens/ServoApp/BrowserPage.cpp
+++ b/support/hololens/ServoApp/BrowserPage.cpp
@@ -71,6 +71,10 @@ void BrowserPage::BindServoEvents() {
                 ? Visibility::Collapsed
                 : Visibility::Visible);
       });
+  Window::Current().VisibilityChanged(
+      [=](const auto &, const VisibilityChangedEventArgs &args) {
+        servoControl().ChangeVisibility(args.Visible());
+      });
 }
 
 void BrowserPage::OnURLFocused(Windows::Foundation::IInspectable const &) {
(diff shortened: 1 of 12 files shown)


support/hololens/ServoApp/BrowserPage.cpp
```diff
@@ -71,6 +71,10 @@ void BrowserPage::BindServoEvents() {
                 ? Visibility::Collapsed
                 : Visibility::Visible);
       });
+  Window::Current().VisibilityChanged(
```
Manishearth: I assume this gets called when we enter immersive mode on the hololens as well, right? Is there an FPS improvement?

paulrouget: > I assume this gets called when we enter immersive mode on the hololens as well, right?

Yes.

> Is there an FPS improvement?

I haven't looked.

paulrouget: I don't see obvious improvements, but the emulator is probably not the best way of measure that.

jdm:
@bors-servo retry
* #24726 #25513

jdm:
@bors-servo retry
* #24726

Issue #23339, opened 2019-05-07
Improve layout trace viewer with a screenshot of each step


ferjm:
I'm not sure if this is doable, but it would be really nice to improve the layout trace viewer with a screenshot of each step or each reflow, highlighting the different boxes like in [this reflow visualization from Gecko](https://www.youtube.com/watch?v=ZTnIxIA5KGw)

![hqdefault](https://user-images.githubusercontent.com/480202/57313471-f87dcf00-70ef-11e9-8675-be907eba4209.jpg)

PR #25803, opened 2020-02-19
Layout viewer for layout 2020

ferjm:

This PR makes layout 2020 dump the box and fragment tree states into json files that can be visualized with the layout viewer tool. This tool has not much functionality other than displaying these trees and allowing to inspect each node additional data, so there is a lot of room for improvements. Some ideas for follow-ups:

- Make the tool create and display diffs between tree states.
- Actually allow creating new debug scopes during box tree and fragment tree construction. Right now there is a single scope created after constructing both trees, which is not ideal as it only allows looking at the reflow result.
-  Right now an independent JSON file is created per reflow. It would be nice to unify the data obtained on each reflow on a single JSON, so diffs between reflows can be displayed as well.
- Dump and display the DOM tree. Link boxes to DOM nodes.
- #23339


components/layout_2020/fragments.rs
```diff
@@ -342,3 +346,37 @@ impl CollapsedMargin {
...
+    }
+}
+
+impl Serialize for ImageFragment {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut serializer = serializer.serialize_struct("ImageFragment", 1)?;
+        serializer.serialize_field("rect", &self.rect)?;
+        serializer.end()
+    }
+}
```
SimonSapin: For these three impls: why are they not derived? Could they be derived with some attributes from https://serde.rs/field-attrs.html ?

components/layout_2020/layout_debug.rs
```diff
@@ -0,0 +1,110 @@
...
+/// file can then be viewed with an external tool.
+pub fn end_trace(generation: u32) {
+    let mut thread_state = STATE_KEY.with(|ref r| r.borrow_mut().take().unwrap());
+    assert_eq!(thread_state.scope_stack.len(), 1);
+    let mut root_scope = thread_state.scope_stack.pop().unwrap();
+    root_scope.post = to_value(&thread_state.fragment).unwrap();
+
+    let result = to_string(&root_scope).unwrap();
+    let mut file = File::create(format!("layout_trace-{}.json", generation)).unwrap();
+    file.write_all(result.as_bytes()).unwrap();
```
SimonSapin: Nit: this can use `std::fs::write`

etc/layout_viewer/viewer.html
```diff
@@ -100,22 +100,23 @@ <h1> Servo Layout Viewer </h1>
...
-
                 var node = {
-                    text: trace_node.class + " (" + base.id + ")",
-                    id: base.id,
-                    icon: "dummy",
-                    href: "#diff-" + base.id
+                    text: Object.keys(trace_node)[0],
+//                    id: base.id,
+//                    icon: "dummy",
+//                    href: "#diff-" + base.id
```
SimonSapin: Here and in three other places, prefer removing code rather than leaving it commented out. Git history still has it.

components/layout_2020/fragments.rs
```diff
@@ -375,8 +385,45 @@ impl Serialize for TextFragment {
...
+impl Serialize for DebugId {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_str(&format!("{:p}", &self))
+    }
+}
+
+#[cfg(debug_assertions)]
+impl Serialize for DebugId {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_u16(self.0)
```
SimonSapin: This impl can be derived with `#[serde(transparent)]`

components/layout_2020/layout_debug.rs
```diff
@@ -32,38 +32,52 @@ macro_rules! layout_debug_scope(
     )
 );
 
+#[derive(Serialize)]
+struct TreeValues {
+    pub box_tree: Value,
+    pub fragment_tree: Value,
```
SimonSapin: What do you think of keeping already-serialized `String`s here, instead of `Value`s? I don’t really know which is better, it’s just another idea.

ferjm: Since there's no clear benefit of keeping already-serialized `String`s, I'd prefer to keep the `Value`s to make the parsing on the viewer easier and to allow an easier visualization of the JSON files directly on the browser:

Strings:

Screenshot 2020-02-21 at 16 51 13

vs Objects:

Screenshot 2020-02-21 at 16 56 33

components/layout_2020/fragments.rs
ferjm: I just learnt about this. Thanks!

etc/layout_viewer/viewer.html
ferjm: Yes, this was uncommented on a later commit. Sorry about that

components/layout_2020/fragments.rs
ferjm: I think I learnt about `#[serde(skip_serializing)]` after implementing the serializers. Fixed.

components/layout_2020/layout_debug.rs
SimonSapin: I meant `String`s in JSON syntax that would be written directly to files. Your "Strings" screenshot involves double-serialization to JSON, which is not what I was suggesting. But anyway, `Value` is fine (and might be easier to combine if you want more than one in a given `.json` file?)

jdm:
@bors-servo retry
* #24726

jdm:
@bors-servo retry
* #25513

jdm:
@bors-servo retry
* #25513

PR #25817, opened 2020-02-20
Generate ::before and ::after content for layout_2020

ferjm:

This PR generates content from the `<string>` and `attr(<identifier>)` `content` property values. I'll work on other values in follow-ups.

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -9,8 +9,11 @@ use crate::replaced::ReplacedContent;
 use crate::style_ext::{Display, DisplayGeneratingBox, DisplayInside, DisplayOutside};
 use crate::wrapper::GetRawData;
 use atomic_refcell::{AtomicRefCell, AtomicRefMut};
+use html5ever::LocalName;
 use net_traits::image::base::Image as NetImage;
-use script_layout_interface::wrapper_traits::{LayoutNode, ThreadSafeLayoutNode};
+use script_layout_interface::wrapper_traits::{
+    LayoutNode, ThreadSafeLayoutElement, ThreadSafeLayoutNode,
+};
 use servo_arc::Arc as ServoArc;
 use std::marker::PhantomData as marker;
 use std::sync::Arc;
@@ -253,24 +256,23 @@ fn pseudo_element_style<'dom, Node>(
 where
     Node: NodeExt<'dom>,
 {
-    if let Some(pseudo_element) = match which {
+    match which {
         WhichPseudoElement::Before => element.to_threadsafe().get_before_pseudo(),
         WhichPseudoElement::After => element.to_threadsafe().get_after_pseudo(),
-    } {
+    }
+    .and_then(|pseudo_element| {
         let style = pseudo_element.style(context.shared_context());
         if style.ineffective_content_property() {
             None
         } else {
             Some(style)
         }
-    } else {
-        None
-    }
+    })
 }
 
 fn generate_pseudo_element_content<'dom, Node>(
     pseudo_element_style: &ComputedValues,
-    _element: Node,
+    element: Node,
     _context: &LayoutContext,
 ) -> Vec<PseudoElementContentItem>
 where
@@ -282,14 +284,25 @@ where
             for item in items.iter() {
                 match item {
                     ContentItem::String(s) => {
-                        vec.push(PseudoElementContentItem::Text(s.to_string()))
+                        vec.push(PseudoElementContentItem::Text(s.to_string()));
                     },
-                    _ => unimplemented!(),
+                    ContentItem::Attr(attr) => {
+                        let element = element
+                            .to_threadsafe()
+                            .as_element()
+                            .expect("Expected an element");
+                        let attr_val = element
+                            .get_attr(&attr.namespace_url, &LocalName::from(&*attr.attribute));
+                        vec.push(PseudoElementContentItem::Text(
+                            attr_val.map_or("".to_string(), |s| s.to_string()),
+                        ));
+                    },
+                    _ => (),
                 }
             }
             vec
         },
-        _ => vec![],
+        Content::Normal | Content::None => unreachable!(),
     }
 }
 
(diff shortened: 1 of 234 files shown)


tests/wpt/include-layout-2020.ini
```diff
@@ -13,6 +13,8 @@ skip: true
       skip: false
     [box-display]
       skip: false
+    [generated-content]
+      skip: false
     [zindex]
       skip: false
     [zorder]
```
SimonSapin: Let’s also enable `web-platform-tests/css/css-content` here, even if we don’t yet implement anything from https://drafts.csswg.org/css-content/. It’s possible that some of those tests cover some of the functionality from CSS2.

components/layout_2020/dom_traversal.rs
```diff
@@ -244,29 +246,51 @@ impl NonReplacedContents {
...
+    } {
+        let style = pseudo_element.style(context.shared_context());
+        if style.ineffective_content_property() {
+            None
+        } else {
+            Some(style)
+        }
+    } else {
+        None
+    }
```
SimonSapin: Nit: does this `if let…else` look better as `.and_then(|| {…})`?

components/layout_2020/dom_traversal.rs
```diff
@@ -244,29 +246,51 @@ impl NonReplacedContents {
...
-    unimplemented!()
+    match &pseudo_element_style.get_counters().content {
+        Content::Items(ref items) => {
+            let mut vec = vec![];
+            for item in items.iter() {
+                match item {
+                    ContentItem::String(s) => {
+                        vec.push(PseudoElementContentItem::Text(s.to_string()))
+                    },
+                    _ => unimplemented!(),
```
SimonSapin: Maybe silently ignoring unsupported values is better than panicking. I wonder if panics cause reftests to timeout. (This would be another bug that we should fix, but oh well.)

For full spec compliance and for making `@supports` meaningful we should not parse values that are not supported yet, but no need to bother if we’re adding support soon.

components/layout_2020/dom_traversal.rs
```diff
@@ -244,29 +246,51 @@ impl NonReplacedContents {
...
+                }
+            }
+            vec
+        },
+        _ => vec![],
```
SimonSapin:
```suggestion
        Content::Normal | Content::None => unreachable!(),
```

Here however this function should not be called with these values (because `pseudo_element_style` returned `Option::None`), so panicking is appropriate since it indicates a bug.

components/layout_2020/dom_traversal.rs
```diff
@@ -282,7 +285,20 @@ where
...
+                        vec.push(PseudoElementContentItem::Text(s.to_string()));
+                    },
+                    ContentItem::Attr(attr) => {
+                        let element = element
+                            .to_threadsafe()
+                            .as_element()
+                            .expect("Expected an element");
+                        let attr_val = element.get_attr(
+                            &attr.namespace_url,
+                            &LocalName::from(attr.attribute.to_string()),
```
SimonSapin:
```suggestion
                            &LocalName::from(&*attr.attribute),
```

`style::Atom` dereferences to `&str` and `LocalName` implemements `From<&str>`, so there is no need to allocate a string here.

components/style/values/generics/counters.rs
```diff
@@ -8,7 +8,6 @@
 use crate::computed_values::list_style_type::T as ListStyleType;
 #[cfg(feature = "gecko")]
 use crate::values::generics::CounterStyle;
-#[cfg(feature = "gecko")]
```
SimonSapin: This and other `cfg`s changed here should probably be something like `#[cfg(any(feature = "gecko", feature = "servo-layout-2020"))]`, since layout 2013 still doesn’t support this.

SimonSapin:
Looks good, thanks!

@bors-servo r+

jdm:
```
error: variant is never constructed: `Replaced`
  --> components/layout_2020/dom_traversal.rs:52:5
   |
52 |     Replaced(ReplacedContent),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
```

SimonSapin:
That’s… surprising. This variant is constructed here:

https://github.com/servo/servo/blob/afb1faf80f3efbdd2611adad74a125b37df25f61/components/layout_2020/dom_traversal.rs#L115

The variant constructor is passed as a function rather than being called directly, but this PR’s diff doesn’t touch that code at all

jdm:
```
  ▶ PASS [expected FAIL] /css/CSS2/generated-content/content-177.xht
  ▶ PASS [expected FAIL] /_mozilla/css/content_color.html
  ▶ PASS [expected FAIL] /_mozilla/css/pseudo_element_spacing_margin.html
  ▶ PASS [expected FAIL] /_mozilla/css/table_vertical_align_absolute_a.html
  ▶ PASS [expected FAIL] /_mozilla/css/pseudo_element_spacing_padding.html
  ▶ PASS [expected FAIL] /_mozilla/css/position_abs_pseudo_a.html
```

jdm:
@bors-servo retry
* #24622

jdm:
```
  ▶ PASS [expected FAIL] /css/css-transforms/transform-generated-002.html
  ▶ FAIL [expected PASS] /css/css-transforms/transform-input-017.html
  │   → /css/css-transforms/transform-input-017.html 9700e7c5fd889ec45912065b58f6ba72423bc3fd
  └   → /css/css-transforms/transform-input-017-ref.html 12f90dd4c1279d922153b5000ed0f94bc9dbc3d0
  ▶ FAIL [expected PASS] /css/css-transforms/transform-input-002.html
  │   → /css/css-transforms/transform-input-002.html 705964f5fdf1548f0dc113ab660f0b34cc78f98e
  └   → /css/css-transforms/transform-input-002-ref.html 83c172eb615035d7ca38eb02d3643c10f233fb0a
```

PR #25818, opened 2020-02-20
Add initial support for transforms to layout_2020

mrobinson:
---


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -476,6 +499,123 @@ impl BoxFragment {
             );
         }
     }
+
+    fn build_reference_frame_if_necessary(
+        &self,
+        builder: &mut DisplayListBuilder,
+        border_rect: &PhysicalRect<Length>,
+    ) -> bool {
```
SimonSapin: Please add a doc-comment to explain what this return value means.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -476,6 +499,123 @@ impl BoxFragment {
...
+        if !self.has_transform_or_perspective() {
+            return false;
+        }
+        let au_border_rect = f32_rect_to_au_rect(Rect::new(
```
SimonSapin: Could we avoid using `Au` and stay within `Length` or `f32` coordinates in CSS `px` unit? (To help do that replace `to_used_value` with `percentage_relative_to`.) It looks like the transformation matrix is ultimately in floating point anyway.

components/layout_2020/display_list/stacking_context.rs
mrobinson: Okay.

components/layout_2020/display_list/stacking_context.rs
mrobinson: Thanks for the review. I've pushed a new version of the branch that addresses this comment by modifying the style code to accept Length values when resolving transformations.

SimonSapin:
Looks good, thanks!

@bors-servo r+

PR #25819, opened 2020-02-20
Synchronize ANGLE linking path with VS project.

jdm:
This avoids problems when upgrading or downgrading the ANGLE NuGet package from Visual Studio. The next time the Rust code needs to be linked, it's common to forget to update the path in build_commands.py, leading to surprising linker errors. By reading the current package version from the VS project configuration, we avoid this papercut.


jdm:
@bors-servo retry
* #21067

PR #25821, opened 2020-02-21
Use new and new_inherited in messagechannel

gterzian:
Use common DOM constructor idioms in `MessageChannel`. Also, remove the falliable from the result, since that seems to match the spec better https://html.spec.whatwg.org/multipage/web-messaging.html#dom-messagechannel

---


gterzian:
@Manishearth r?

PR #25826, opened 2020-02-22
Make Background Hang Monitor Optional

### LINKED ISSUE #25647: Make background hang monitor optional
In performance-sensitive situations like immersive mode on the hololens, having the background hang monitor thread running shows up in profiles. We should make it a configuration option at engine startup, and put the channels to communicate with the thread inside Option values so we don't have unnecessary channel traffic.

kunalmohan:
This is done by wrapping all channels of communication and related objects inside Option which are configured using flag inside servo_config.

---

- [X] These changes fix #25647  (GitHub issue number if applicable)

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -282,11 +282,11 @@ pub struct Constellation<Message, LTF, STF> {
 
     /// A channel for the background hang monitor to send messages
     /// to the constellation.
-    background_hang_monitor_sender: IpcSender<HangMonitorAlert>,
+    background_hang_monitor_sender: Option<IpcSender<HangMonitorAlert>>,
 
     /// A channel for the constellation to receiver messages
     /// from the background hang monitor.
-    background_hang_monitor_receiver: Receiver<Result<HangMonitorAlert, IpcError>>,
+    background_hang_monitor_receiver: Option<Receiver<Result<HangMonitorAlert, IpcError>>>,
 
     /// An IPC channel for layout threads to send messages to the constellation.
     /// This is the layout threads' view of `layout_receiver`.
@@ -849,28 +849,42 @@ where
                     ipc_scheduler_receiver,
                 );
 
-                let (background_hang_monitor_sender, ipc_bhm_receiver) =
-                    ipc::channel().expect("ipc channel failure");
-                let background_hang_monitor_receiver =
-                    route_ipc_receiver_to_new_mpsc_receiver_preserving_errors(ipc_bhm_receiver);
+                let (background_hang_monitor_sender, background_hang_monitor_receiver) =
+                    if opts::get().background_hang_monitor {
+                        let (bhm_sender, ipc_bhm_receiver) =
+                            ipc::channel().expect("ipc channel failure");
+                        (
+                            Some(bhm_sender),
+                            Some(route_ipc_receiver_to_new_mpsc_receiver_preserving_errors(
+                                ipc_bhm_receiver,
+                            )),
+                        )
+                    } else {
+                        (None, None)
+                    };
 
                 // If we are in multiprocess mode,
                 // a dedicated per-process hang monitor will be initialized later inside the content process.
                 // See run_content_process in servo/lib.rs
-                let (background_monitor_register, sampler_chan) = if opts::multiprocess() {
-                    (None, vec![])
-                } else {
-                    let (sampling_profiler_control, sampling_profiler_port) =
-                        ipc::channel().expect("ipc channel failure");
-
-                    (
-                        Some(HangMonitorRegister::init(
-                            background_hang_monitor_sender.clone(),
-                            sampling_profiler_port,
-                        )),
-                        vec![sampling_profiler_control],
-                    )
-                };
+                let (background_monitor_register, sampler_chan) =
+                    if opts::multiprocess() || !opts::get().background_hang_monitor {
+                        (None, vec![])
+                    } else {
+                        let (sampling_profiler_control, sampling_profiler_port) =
+                            ipc::channel().expect("ipc channel failure");
+                        if let Some(bhm_sender) = background_hang_monitor_sender.clone() {
+                            (
+                                Some(HangMonitorRegister::init(
+                                    bhm_sender,
+                                    sampling_profiler_port,
+                                )),
+                                vec![sampling_profiler_control],
+                            )
+                        } else {
+                            warn!("No BHM sender found in BHM mode.");
+                            (None, vec![])
+                        }
+                    };
 
                 let (ipc_layout_sender, ipc_layout_receiver) =
                     ipc::channel().expect("ipc channel failure");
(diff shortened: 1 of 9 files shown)


components/script/script_thread.rs
```diff
@@ -1269,11 +1269,15 @@ impl ScriptThread {
 
         let task_queue = TaskQueue::new(port, chan.clone());
 
-        let background_hang_monitor = state.background_hang_monitor_register.register_component(
-            MonitoredComponentId(state.id, MonitoredComponentType::Script),
-            Duration::from_millis(1000),
-            Duration::from_millis(5000),
-        );
+        let state_id = state.id.clone();
```
gterzian: You can't do this inside the `map`?

gterzian:
Nice one, can you please squash into one commit also?

components/script/script_thread.rs
kunalmohan: No, because the value of `state` gets moved in line 1266 (`state.control_port`) and is then borrowed and used inside above closure which throws the following error-
```
error[E0382]: borrow of moved value: `state`
    --> components/script/script_thread.rs:1274:90
     |
1266 |         let control_port = ROUTER.route_ipc_receiver_to_new_crossbeam_receiver(state.control_port);
     |                                                                                ------------------ value moved here
...
1274 |         let background_hang_monitor = state.background_hang_monitor_register.clone().map(|bhm| {
     |                                                                                          ^^^^^ value borrowed here after partial move
1275 |             bhm.register_component(
1276 |                 MonitoredComponentId(state.id.clone(), MonitoredComponentType::Script),
     |                                      ----- borrow occurs due to use in closure
     |
     = note: move occurs because `state.control_port` has type `ipc_channel::ipc::IpcReceiver<script_traits::ConstellationControlMsg>`, which does not implement the `Copy` trait
```

kunalmohan:
> Nice one, can you please squash into one commit also?

Yes, I was thinking of doing that once the changes get approved.

kunalmohan:
Unit test for checking panics is currently failing. I am pretty sure that the code should not panic in runtime because that part is executed only if the `background_hang_monitor_sender` is of type `Some(x)`. What would be the best approach to avoid the above failure?

components/script/script_thread.rs
gterzian: Ok. What about if you do `&state.id.clone()`?

components/constellation/constellation.rs
```diff
@@ -849,28 +849,38 @@ where
...
+                let (background_monitor_register, sampler_chan) =
+                    if opts::multiprocess() | !opts::get().background_hang_monitor {
+                        (None, vec![])
+                    } else {
+                        let (sampling_profiler_control, sampling_profiler_port) =
+                            ipc::channel().expect("ipc channel failure");
+
+                        (
+                            Some(HangMonitorRegister::init(
+                                background_hang_monitor_sender.clone().unwrap(),
```
gterzian: To fix the panic lint, you have to remove the `unwrap` here. It should never panic, but it's just a lint on the `unwrap`. So you can return `(None, vec![])` here as well, and `warn`, if `background_hang_monitor_sender` is `None`.

components/script/script_thread.rs
kunalmohan: Nope, error-"expected struct, found reference".

gterzian: Ah Ok, I think it will work with `state.id.clone()` if you move the `let control_port =` to below `let background_hang_monitor =`(which I think it possible)

components/constellation/constellation.rs
kunalmohan: If I put a check with `is_some()` and return above if it's true but `unwrap()` would still be used inside the `if` block. Won't that be a problem?

components/script/script_thread.rs
kunalmohan: This one worked!

components/constellation/constellation.rs
gterzian: So the idea is that there is no `unwrap` in the code, even if logically it will never panic :)

I think you literally do `If let Some(bhm_sender) =`, and put the code in there, then in the `else` branch you `warn!(No BHM sender found in BHM mode.)` and return `(None, vec![])`

kunalmohan: Ah, my bad. I completely forgot about the `if let` pattern matching. I was entirely relying on the methods defined on Option enum.

components/constellation/constellation.rs
```diff
@@ -849,28 +849,38 @@ where
...
-                    (
-                        Some(HangMonitorRegister::init(
-                            background_hang_monitor_sender.clone(),
-                            sampling_profiler_port,
-                        )),
-                        vec![sampling_profiler_control],
-                    )
-                };
+                let (background_monitor_register, sampler_chan) =
+                    if opts::multiprocess() | !opts::get().background_hang_monitor {
```
CYBAI: Should this be `||` instead 👀?

```suggestion
                    if opts::multiprocess() || !opts::get().background_hang_monitor {
```

kunalmohan: Yeah, it should. Thanks for pointing it out. I'll do that now.

gterzian:
Nice one! 

@bors-servo r+

gterzian:
```
thread 'main' panicked at 'assertion failed: result.success()', /Users/worker/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/build.rs:133:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1052
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1428
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook

```

Not sure what this is, an error in the CI worker? Perhaps just worth a retry.

gterzian:
```
TIMEOUT [expected OK] /webdriver/tests/interface.html
  │ 
  │ 
  │ 
  │ 
  │ Couldn't start content, no background monitor has been initiated (thread Constellation, at /repo/components/constellation/pipeline.rs:322)
  │ stack backtrace:
  │              at src/libstd/panicking.rs:474
  │    3: rust_begin_unwind
  │              at src/libstd/panicking.rs:378
  │              at src/libcore/panicking.rs:85
  │    5: core::option::expect_failed
  │              at src/libcore/option.rs:1203
  │    6: constellation::pipeline::Pipeline::spawn
  │    7: constellation::constellation::Constellation<Message,LTF,STF>::new_pipeline
  │    8: constellation::constellation::Constellation<Message,LTF,STF>::handle_request_from_compositor
  │    9: constellation::constellation::Constellation<Message,LTF,STF>::run
  │   10: constellation::constellation::Constellation<Message,LTF,STF>::start::{{closure}}
```

this seems relevant

gterzian:
I think this needs to be handled based on whether the BHM is on or off: https://github.com/servo/servo/pull/25826/files#diff-415820df4ebfaf3e08b8fa2ea3a30beaR324

gterzian:
@kunalmohan Thanks! 

@bors-servo r+

PR #25831, opened 2020-02-24
Implement mapReadAsync function of GPUBuffer

imiklos:
Implemented the `mapReadAsync` and fixed the `unmap` functions of `GPUBuffer`.
Added `mapped` internal slot for tracking the ArrayBuffer/Promise.
Added more states to the `GPUBufferState` enum.

---

cc @kvark @jdm @zakorgy

### DIFF
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -6,17 +6,36 @@ use crate::dom::bindings::cell::{DomRefCell, Ref};
 use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::{
     self, GPUBufferMethods, GPUBufferSize,
 };
+use crate::dom::bindings::error::Error;
+use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::DOMString;
+use crate::dom::bindings::trace::RootedTraceableBox;
 use crate::dom::globalscope::GlobalScope;
+use crate::dom::gpu::{response_async, AsyncWGPUListener};
+use crate::dom::promise::Promise;
+use crate::realms::InRealm;
 use dom_struct::dom_struct;
+use js::jsapi::{Heap, JSObject};
+use js::jsval::UndefinedValue;
+use js::rust::jsapi_wrapped::{DetachArrayBuffer, IsPromiseObject, RejectPromise};
+use js::rust::MutableHandle;
+use js::typedarray::{ArrayBuffer, CreateWith};
 use std::cell::Cell;
-use webgpu::{WebGPU, WebGPUBuffer, WebGPUDevice, WebGPURequest};
+use std::ptr;
+use std::rc::Rc;
+use webgpu::{
+    wgpu::resource::BufferUsage, WebGPU, WebGPUBuffer, WebGPUDevice, WebGPURequest, WebGPUResponse,
+};
 
-#[derive(MallocSizeOf)]
+// https://gpuweb.github.io/gpuweb/#buffer-state
+#[derive(Clone, MallocSizeOf)]
 pub enum GPUBufferState {
-    Mapped,
+    MappedForReading,
+    MappedForWriting,
+    MappedPendingForReading,
+    MappedPendingForWriting,
     Unmapped,
     Destroyed,
 }
@@ -24,7 +43,7 @@ pub enum GPUBufferState {
 #[dom_struct]
 pub struct GPUBuffer {
     reflector_: Reflector,
-    #[ignore_malloc_size_of = "channels are hard"]
+    #[ignore_malloc_size_of = "defined in webgpu"]
     channel: WebGPU,
     label: DomRefCell<Option<DOMString>>,
     size: GPUBufferSize,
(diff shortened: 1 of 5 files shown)


jdm:
This needs a `mach fmt` according to CI.

components/webgpu/lib.rs
```diff
@@ -412,6 +419,46 @@ impl WGPU {
...
                 },
+                WebGPURequest::MapReadAsync(sender, buffer_id, device_id, usage, size) => {
+                    let global = &self.global;
+                    let on_read = move |status: wgpu::resource::BufferMapAsyncStatus,
+                                        ptr: *const u8| {
+                        match status {
+                            wgpu::resource::BufferMapAsyncStatus::Success => {
+                                let mut array_buffer = Vec::with_capacity(size as usize);
+                                unsafe {
+                                    array_buffer.set_len(size as usize);
```
jdm: We should call set_len after the copy completes.

jdm (changes requested):
Looking good so far!

components/script/dom/gpubuffer.rs
```diff
@@ -104,19 +127,68 @@ impl Drop for GPUBuffer {
...
+                            DetachArrayBuffer(
+                                *cx,
+                                Handle::from_marked_location(array_buffer.underlying_object()),
+                            );
+                        }
+                    },
+                    _ => {
+                        // Step 2
+                        unsafe {
+                            let handle = Handle::from_marked_location(self.mapping.get() as _);
```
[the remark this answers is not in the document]
imiklos: I think we can't use it because `self.mapping.handle()` returns a `js::rust::Handle` but `IsPromiseObject`, `RejectPromise` and `DetachArrayBuffer` accepts `js::jsapi::Handle`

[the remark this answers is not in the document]
jdm: Import IsPromiseObject and DetachArrayBuffer from js::rust::wrapped instead of jsapi. Those version accept rust::Handle. Alternativey use .into() to convert rust::Handle to jsapi::Handle.

components/script/dom/gpubuffer.rs
```diff
@@ -104,19 +127,68 @@ impl Drop for GPUBuffer {
...
+                                self.device.0,
+                                self.id(),
+                                array_buffer.to_vec(),
+                            ))
+                            .unwrap();
+                        // Step 3.2
+                        unsafe {
+                            DetachArrayBuffer(
+                                *cx,
+                                Handle::from_marked_location(array_buffer.underlying_object()),
```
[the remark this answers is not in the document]
imiklos: Same problem described above.

components/script/dom/gpubuffer.rs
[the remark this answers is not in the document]
imiklos: Thank you! Importing the right one is did the trick!

imiklos:
@jdm Thank you for the review! I've updated the pull request with the changes you mentioned.

jdm (approved):

components/webgpu/lib.rs
```diff
@@ -251,27 +259,26 @@ impl WGPU {
...
                             e
                         )
                     }
                 },
-                WebGPURequest::UnmapBuffer(buffer) => {
+                WebGPURequest::UnmapBuffer(device_id, buffer, array_buffer) => {
                     let global = &self.global;
-                    gfx_select!(buffer.0 => global.buffer_unmap(buffer.0));
+
+                    gfx_select!(buffer.0 => global.device_set_buffer_sub_data(
```
kvark: this should only happen if the buffer was mapped for writing

components/webgpu/lib.rs
```diff
@@ -412,6 +419,46 @@ impl WGPU {
                         )
                     }
                 },
+                WebGPURequest::MapReadAsync(sender, buffer_id, device_id, usage, size) => {
+                    let global = &self.global;
+                    let on_read = move |status: wgpu::resource::BufferMapAsyncStatus,
+                                        ptr: *const u8| {
+                        match status {
+                            wgpu::resource::BufferMapAsyncStatus::Success => {
+                                let mut array_buffer = Vec::with_capacity(size as usize);
```
kvark: nit: would be more idiomatic to do `slice::from_raw_parts(ptr, ...).to_vec()`

components/webgpu/lib.rs
```diff
@@ -412,6 +419,46 @@ impl WGPU {
...
+                                unsafe {
+                                    std::ptr::copy(ptr, array_buffer.as_mut_ptr(), size as usize);
+                                    array_buffer.set_len(size as usize);
+                                };
+                                if let Err(e) =
+                                    sender.send(Ok(WebGPUResponse::MapReadAsync(array_buffer)))
```
kvark: I wonder if servo's IPC allows to put stuff in shared memory without going through the regular `Vec` heap?..

jdm: We do support shared memory: https://docs.rs/ipc-channel/0.14.0/ipc_channel/ipc/struct.IpcSharedMemory.html

jdm: Probably worth filing a followup issue about investigating using that.

components/webgpu/lib.rs
imiklos: We are only call this if the `BufferState` is `MappedForWriting` and the `mapping` slot is an `ArrayBuffer`.

components/webgpu/lib.rs
imiklos: Thank you! I've updated the pull request with using `IpcSharedMemory`.

jdm:
@bors-servo retry
* network

jdm:
@bors-servo retry
* network

PR #25832, opened 2020-02-24
Fix `mach run --headless`

jdm:
This is a regression from #25350.


jdm:
@bors-servo retry
* network failure

PR #25839, opened 2020-02-24
Remove requirement to install pkg-config on Windows.

jdm:
I recently set up a new Windows laptop to build Servo and never bothered to install pkg-config. The build was successful, so we should remove it from the requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ pip install virtualenv
 ```
  If this does not work, you may need to reboot for the changed PATH settings (by the python installer) to take effect.
 
-5. Install the most recent [GStreamer](https://gstreamer.freedesktop.org/data/pkg/windows/) MSVC packages. You need to download the two `.msi` files for your platform from the [GStreamer](https://gstreamer.freedesktop.org/data/pkg/windows/) website and install them. The currently recommended version is 1.16.0. i.e.:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -228,8 +228,6 @@ pip install virtualenv
 ```
  If this does not work, you may need to reboot for the changed PATH settings (by the python installer) to take effect.
 
-4. Install `pkg-config` either via [Chocolatey](https://chocolatey.org/install#installing-chocolatey)(`choco install pkgconfiglite`) or from [here](https://sourceforge.net/projects/pkgconfiglite/).
-
 5. Install the most recent [GStreamer](https://gstreamer.freedesktop.org/data/pkg/windows/) MSVC packages. You need to download the two `.msi` files for your platform from the [GStreamer](https://gstreamer.freedesktop.org/data/pkg/windows/) website and install them. The currently recommended version is 1.16.0. i.e.:
```
atouchet: Should this be changed to step 4 then?

jdm: Numbering is automatic in markdown.

CYBAI: I just check the [rendered](https://github.com/servo/servo/blob/f70d5fcfc9d1ec807a4b54b13ae55b81362c8946/README.md) readme. Looks like we need to change the step number if it's specified.

I remember numbering can be automatic in markdown with `1.`

Ex.

```
1. First
1. Second
1. Third
```

above snippet would be 👇

1. First
1. Second
1. Third

SimonSapin: https://github.com/servo/servo/blob/f70d5fcfc9d1ec807a4b54b13ae55b81362c8946/README.md#user-content-on-windows-msvc does show 3 then 5. I’ll renumber.

jdm:
@bors-servo retry
* #24622

PR #25841, opened 2020-02-25
Remove azure canvas backend

### LINKED ISSUE #25833: Remove azure canvas backend
Now that the [raqote backend](https://github.com/servo/servo/blob/master/components/canvas/raqote_backend.rs) has reached feature parity and been enabled by default, we should remove the unused [azure backend](https://github.com/servo/servo/blob/master/components/canvas/azure_backend.rs) entirely.

dralley:
closes #25833 

Issue description: Now that the raqote backend has reached feature parity and been enabled by default, we should remove the unused azure backend entirely.

- [x] These changes fix #25833 

- [x] These changes do not require tests because PR only removes code, doesn't add new code


dralley:
Unfortunately, as I'm on Fedora 31, I don't think I can currently actually build servo.  I'll have to figure something out, maybe run Fedora 29/30 in a VM...

components/canvas/canvas_data.rs
```diff
@@ -23,6 +23,7 @@ use webrender_api::units::RectExt as RectExt_;
 /// further operations to it in device space. When it's time to
 /// draw the path, we convert it back to userspace and draw it
 /// with the correct transform applied.
+/// TODO: De-abstract now that Azure is removed?
```
dralley: I can remove these comments or change the format, I don't know what the plan is here.

python/tidy/servo_tidy_tests/rust_tidy.rs
```diff
@@ -3,12 +3,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use    app_units::Au;
-use azure::azure_hl::{ AntialiasMode, Color, ColorPattern, CompositionOp };
-use azure::azure_hl::{AntialiasMode, Color,
+use raqote::{GradientStop, Source, SolidSource};
+use raqote::{Source, SolidSource},
```
dralley: I just picked random structs, not sure if I'm still matching the intention of the tests.

components/canvas/canvas_data.rs
dralley: Is de-abstracting these things also considered part of the issue? It would be straightforwards, I think, I just didn't want to assume it was desired immediately

SimonSapin: Merging the code removing is a good step even if a now-unnecessary abstraction is still there. Leaving a TODO or FIXME comment is fine.

SimonSapin:
Looks good, thanks!

@bors-servo r+

python/tidy/servo_tidy_tests/rust_tidy.rs
dralley: @SimonSapin I just noticed -- I didn't preserve the extraneous whitespace on the first line here.

SimonSapin: Does `./mach test-tidy --self-test` succeed for you? If not adjust files until it does, make or amend commits, and push or force-push to this PR again.

components/canvas/canvas_data.rs
SimonSapin: @pcwalton Do you think we should keep this abstraction around in case we want to add a Pathfinder backend at some point?

Issue #23168, opened 2019-04-05
Intermittent failure in hang_monitor_tests


jdm:
```
     Running target/debug/deps/hang_monitor_tests-d923c72bc20a525c
running 3 tests
Enabling profiler.
Disabling profiler.
Resolving 1/10
Resolving 2/10
Resolving 3/10
Resolving 4/10
Resolving 5/10
Resolving 6/10
Resolving 7/10
Resolving 8/10
Resolving 9/10
Resolving 10/10
test test_sampler ... ok
error: process didn't exit successfully: `/home/travis/build/servo/servo/target/debug/deps/hang_monitor_tests-d923c72bc20a525c` (signal: 27)
The command "./mach test-unit" exited with 101.
```

pshaughn:
Could this have been #24984? (Do we have some way to go through test logs to see whether intermittents stopped happening?)

jdm:
Issues like this one which do not affect WPT tests only get recorded manually. As far as I know we only saw this last april and may in two different pull requests.

SimonSapin:
Occurred in a daily task: https://community-tc.services.mozilla.com/tasks/O7bkuKN6RF-LRf5vfgwBZQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FO7bkuKN6RF-LRf5vfgwBZQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L2347

lumiscosity:
I can't replicate this anymore after running the test manually ~20 times. Tested on Arch Linux x86_64, Servo built against commit 1f0f079, dev build.

PR #25862, opened 2020-02-28
Have transforms and filters be CBs for all descendants in layout_2020

mrobinson:
This is a feature that was never properly implemented in the previous
layout system. We still need to preserve their in-tree order in the
display list though.

---

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -37,7 +37,12 @@ type ItemTag = (u64, u16);
 type HitInfo = Option<ItemTag>;
 
 pub struct DisplayListBuilder<'a> {
+    /// The current SpatialId and ClipId information for this `DisplayListBuilder`.
     current_space_and_clip: wr::SpaceAndClipInfo,
+
+    /// The id of the nearest ancestor reference frame for this `DisplayListBuilder`.
+    nearest_reference_frame: wr::SpatialId,
+
     pub context: &'a LayoutContext<'a>,
     pub wr: wr::DisplayListBuilder,
 
@@ -56,6 +61,7 @@ impl<'a> DisplayListBuilder<'a> {
     ) -> Self {
         Self {
             current_space_and_clip: wr::SpaceAndClipInfo::root_scroll(pipeline_id),
+            nearest_reference_frame: wr::SpatialId::root_reference_frame(pipeline_id),
             is_contentful: false,
             context,
             wr: wr::DisplayListBuilder::new(pipeline_id, viewport_size),
@@ -68,9 +74,14 @@ impl<'a> DisplayListBuilder<'a> {
     }
 
     fn clipping_and_scrolling_scope<R>(&mut self, f: impl FnOnce(&mut Self) -> R) -> R {
-        let previous = self.current_space_and_clip;
+        let previous_space_and_clip = self.current_space_and_clip;
+        let previous_nearest_reference_frame = self.nearest_reference_frame;
+
         let result = f(self);
-        self.current_space_and_clip = previous;
+
+        self.current_space_and_clip = previous_space_and_clip;
+        self.nearest_reference_frame = previous_nearest_reference_frame;
+
         result
     }
 }
--- a/components/layout_2020/positioned.rs
+++ b/components/layout_2020/positioned.rs
@@ -25,7 +25,12 @@ pub(crate) struct AbsolutelyPositionedBox {
 
 pub(crate) struct PositioningContext<'box_tree> {
     for_nearest_positioned_ancestor: Option<Vec<HoistedAbsolutelyPositionedBox<'box_tree>>>,
-    for_initial_containing_block: Vec<HoistedAbsolutelyPositionedBox<'box_tree>>,
+
+    // For nearest `containing block for all descendants` as defined by the CSS transforms
+    // spec.
+    // https://www.w3.org/TR/css-transforms-1/#containing-block-for-all-descendants
+    for_nearest_containing_block_for_all_descendants:
+        Vec<HoistedAbsolutelyPositionedBox<'box_tree>>,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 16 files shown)


components/layout_2020/display_list/mod.rs
```diff
@@ -37,7 +37,12 @@ type ItemTag = (u64, u16);
 type HitInfo = Option<ItemTag>;
 
 pub struct DisplayListBuilder<'a> {
+    /// The current SpatialId and ClipId information for this `DisplayListBuilder`.
     current_space_and_clip: wr::SpaceAndClipInfo,
+
+    /// The id of the nearest ancestor reference frame for this `DisplayListBuilder`.
+    nearest_reference_frame: wr::SpatialId,
```
pcwalton: What is a *reference frame* defined as? A quick Google search doesn't pull up that term in a CSS spec.

components/layout_2020/positioned.rs
```diff
@@ -25,7 +25,14 @@ pub(crate) struct AbsolutelyPositionedBox {
...
 pub(crate) struct PositioningContext<'box_tree> {
     for_nearest_positioned_ancestor: Option<Vec<HoistedAbsolutelyPositionedBox<'box_tree>>>,
-    for_initial_containing_block: Vec<HoistedAbsolutelyPositionedBox<'box_tree>>,
+
+    // For nearest `containing block for all descendants` as defined by the CSS transforms
+    // spec.
+    // https://www.w3.org/TR/css-transforms-1/#containing-block-for-all-descendants
+    //
+    // We use the abbreviation `cbfad` here only because `containing block for all descendants`
+    // is so unwieldy.
```
pcwalton: 😆

components/layout_2020/positioned.rs
```diff
@@ -396,91 +476,100 @@ impl<'box_tree> HoistedAbsolutelyPositionedBox<'box_tree> {
...
-                Err(non_replaced) => {
-                    // https://drafts.csswg.org/css2/visudet.html#abs-non-replaced-width
-                    // https://drafts.csswg.org/css2/visudet.html#abs-non-replaced-height
-                    let inline_size = inline_axis.size.auto_is(|| {
-                        let available_size = match inline_axis.anchor {
-                            Anchor::Start(start) => {
-                                cbis - start - pb.inline_sum() - margin.inline_sum()
-                            },
-                            Anchor::End(end) => cbis - end - pb.inline_sum() - margin.inline_sum(),
+        let for_cbfad = for_nearest_cbfad;
```
pcwalton: This isn't changed other than the replacement of `PositioningContext::for_positioned` with `PositioningContext::create_and_layout_positioned`, right?

components/layout_2020/style_ext.rs
```diff
@@ -165,6 +173,82 @@ impl ComputedValuesExt for ComputedValues {
...
+    /// Get the effective z-index of this fragment. Z-indices only apply to positioned element
+    /// per CSS 2 9.9.1 (http://www.w3.org/TR/CSS2/visuren.html#z-index), so this value may differ
+    /// from the value specified in the style.
+    fn effective_z_index(&self) -> i32 {
+        match self.get_box().position {
+            ComputedPosition::Static => {},
+            _ => return self.get_position().z_index.integer_or(0),
+        }
+
+        0
```
pcwalton: nit: would probably be simpler and more idiomatic to simply write:

    match self.get_box().position {
        ComputedPosition::Static => 0,
        _ => self.get_position().z_index.integer_or(0),
    }

or even

    if self.get_box().position == ComputedPosition::Static {
        0
    } else {
        self.get_position().z_index.integer_or(0)
    }

components/layout_2020/style_ext.rs
```diff
@@ -165,6 +173,82 @@ impl ComputedValuesExt for ComputedValues {
             self.writing_mode,
         )
     }
+
+    /// Returns true if this style has a transform, or perspective property set.
+    fn has_transform_or_perspective(&self) -> bool {
+        !self.get_box().transform.0.is_empty() || self.get_box().perspective != Perspective::None
+    }
+
+    /// Get the effective z-index of this fragment. Z-indices only apply to positioned element
```
pcwalton: uber-nit: should be "positioned elements"

pcwalton (commented):
This looks good, modulo a few nits, but I'd like @nox to glance at it before merging.

components/layout_2020/positioned.rs
nox: Personally I think `containing_block_for_all_descendants` is way better, and not even the longest identifier in layout 2020.

nox: We have a `next_in_flow_margin_collapses_with_parent_start_margin`.

components/layout_2020/display_list/mod.rs
mrobinson: The `reference frame` is a concept from WebRender. Essentially it is the thing that does transformations and starts new coordinate systems.

components/layout_2020/positioned.rs
mrobinson: Okay. I've changed this to be `containing_block_for_all_descendants` everywhere.

components/layout_2020/positioned.rs
mrobinson: That's correct. Unfortunately this caused a reflow of the entire block.

components/layout_2020/style_ext.rs
mrobinson: Fixed!

components/layout_2020/style_ext.rs
mrobinson: Fixed this as well. Thanks for pointing it out.

mrobinson:
Okay. Thanks for the reviews. There's a new version of the branch if you'd all like to take a look.

mrobinson:
@bors-servo r=pcwalton,nox

The earlier version of the branch had added a bogus assertion. I've removed this.

mrobinson:
@bors-servo retry
  * #23168

PR #25864, opened 2020-02-28
Rename AbsolutelyPositionedBox::layout to AbsolutelyPositionedBox::to_hoisted

mrobinson:
This method doesn't actually do any layout, but converts this block to a
HoistedAbsolutelyPositionedBox which is hoisted and then laid-out with
its containing block later. This makes the code a little easier to read.

---

- [x] These changes do not require tests because it does not change behavior.


jdm:
@bors-servo r+
Looks like a reasonable improvement!

jdm:
@bors-servo retry
* #24814

PR #25867, opened 2020-02-28
Make CanvasRenderingContext2D.fillRect() draw patterns correctly

pylbrecht:

### DIFF
--- a/components/canvas/canvas_data.rs
+++ b/components/canvas/canvas_data.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::canvas_paint_thread::AntialiasMode;
+use crate::raqote_backend::Repetition;
 use canvas_traits::canvas::*;
 use cssparser::RGBA;
 use euclid::default::{Point2D, Rect, Size2D, Transform2D, Vector2D};
@@ -78,7 +79,6 @@ pub trait Backend {
     );
     fn create_drawtarget(&self, size: Size2D<u64>) -> Box<dyn GenericDrawTarget>;
     fn recreate_paint_state<'a>(&self, state: &CanvasPaintState<'a>) -> CanvasPaintState<'a>;
-    fn size_from_pattern(&self, rect: &Rect<f32>, pattern: &Pattern) -> Option<Size2D<f32>>;
 }
 
 /// A generic PathBuilder that abstracts the interface for azure's and raqote's PathBuilder.
@@ -470,12 +470,41 @@ impl<'a> CanvasData<'a> {
             return; // Paint nothing if gradient size is zero.
         }
 
-        let draw_rect = Rect::new(
-            rect.origin,
-            self.backend
-                .size_from_pattern(&rect, &self.state.fill_style)
-                .unwrap_or(rect.size),
-        );
+        let draw_rect = match &self.state.fill_style {
+            Pattern::Raqote(pattern) => match pattern {
+                crate::raqote_backend::Pattern::Surface(pattern) => {
+                    let pattern_rect = Rect::new(Point2D::origin(), pattern.size());
+                    let mut draw_rect = rect.intersection(&pattern_rect).unwrap_or(Rect::zero());
+
+                    match pattern.repetition() {
+                        Repetition::NoRepeat => {
+                            draw_rect.size.width =
+                                draw_rect.size.width.min(pattern_rect.size.width);
+                            draw_rect.size.height =
+                                draw_rect.size.height.min(pattern_rect.size.height);
+                        },
+                        Repetition::RepeatX => {
+                            draw_rect.size.width = rect.size.width;
+                            draw_rect.size.height =
+                                draw_rect.size.height.min(pattern_rect.size.height);
+                        },
+                        Repetition::RepeatY => {
+                            draw_rect.size.height = rect.size.height;
+                            draw_rect.size.width =
+                                draw_rect.size.width.min(pattern_rect.size.width);
+                        },
+                        Repetition::Repeat => {
+                            draw_rect = *rect;
+                        },
+                    }
+
+                    draw_rect
+                },
+                crate::raqote_backend::Pattern::Color(..) |
+                crate::raqote_backend::Pattern::LinearGradient(..) |
+                crate::raqote_backend::Pattern::RadialGradient(..) => *rect,
+            },
+        };
 
         if self.need_to_draw_shadow() {
             self.draw_with_shadow(&draw_rect, |new_draw_target: &mut dyn GenericDrawTarget| {
--- a/components/canvas/raqote_backend.rs
+++ b/components/canvas/raqote_backend.rs
@@ -28,31 +28,6 @@ impl Backend for RaqoteBackend {
         color.as_raqote().a != 0
     }
 
-    fn size_from_pattern(
-        &self,
-        rect: &Rect<f32>,
-        pattern: &canvas_data::Pattern,
-    ) -> Option<Size2D<f32>> {
-        match pattern {
-            canvas_data::Pattern::Raqote(Pattern::Surface(pattern)) => match pattern.repeat {
-                Repetition::RepeatX => Some(Size2D::new(
-                    rect.size.width as f32,
-                    pattern.image.height as f32,
-                )),
-                Repetition::RepeatY => Some(Size2D::new(
-                    pattern.image.width as f32,
-                    rect.size.height as f32,
-                )),
-                Repetition::Repeat => Some(rect.size),
-                Repetition::NoRepeat => Some(Size2D::new(
-                    pattern.image.width as f32,
-                    pattern.image.height as f32,
-                )),
-            },
-            _ => None,
-        }
-    }
-
     fn set_shadow_color<'a>(&mut self, color: RGBA, state: &mut CanvasPaintState<'a>) {
         state.shadow_color = Color::Raqote(color.to_raqote_style());
     }
@@ -207,6 +182,12 @@ impl<'a> SurfacePattern<'a> {
     }
     fn set_transform(&mut self, transform: Transform2D<f32>) {
         self.transform = transform;
+    }
+    pub fn size(&self) -> Size2D<f32> {
+        Size2D::new(self.image.width as f32, self.image.height as f32)
+    }
+    pub fn repetition(&self) -> &Repetition {
+        &self.repeat
     }
 }
 
--- a/tests/wpt/metadata/2dcontext/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord3.html.ini
+++ b/tests/wpt/metadata/2dcontext/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord3.html.ini
@@ -1,5 +1,0 @@
-[2d.pattern.paint.norepeat.coord3.html]
-  type: testharness
-  [Canvas test: 2d.pattern.paint.norepeat.coord3]
-    expected: FAIL
-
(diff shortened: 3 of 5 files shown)


pylbrecht:
While investigating [2d.pattern.paint.norepeat.coord3.html](https://wpt.fyi/results/2dcontext/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord3.html?label=master&label=experimental&aligned&q=pattern.paint.norepeat) it seems that we should apply the transformation to the source (= `pattern` in the test case) rather than the drawtarget.

This is still very WIP, but I need to check if these changes fails any tests before moving forward with this.

pylbrecht:
What was I thinking, this will of course break a lot of transformation tests (if not all of them). I'll think of something else.

pylbrecht:
Servo actually uses [pre_translate()](https://dxr.mozilla.org/servo/source/components/script/canvas_state.rs#1396) on the current transform. Translating that into the raqote example did not result in the behavior expected by the test:
```rust
extern crate raqote;
extern crate euclid;

use raqote::*;
use euclid::*;

fn main() {
    let mut dt = DrawTarget::new(300, 300);
    let options = DrawOptions::new();
    let red = Source::Solid(SolidSource::from_unpremultiplied_argb(255, 255, 0, 0));
    let green = Source::Solid(SolidSource::from_unpremultiplied_argb(255, 0, 255, 0));

    dt.fill_rect(0., 0., 100., 50., &green, &options);

    let img = Image {
        width: 100,
        height: 50,
        data: &vec![0xffff0000; 100*50],
    };

    let identity = Transform::identity();

    let img_src = Source::Image(img, ExtendMode::Pad, FilterMode::Nearest, identity);
    let transform = dt.get_transform();
    dt.set_transform(&transform.pre_translate(vec2(50., 25.)));
    dt.fill_rect(-50., -25., 100., 50., &img_src, &options);

    dt.fill_rect(0., 0., 50., 25., &green, &options);

    dt.write_png("example.png");
}
```

jrmuizel:
The problem here seems to be the ExtendMode::Pad which causes the image to be padded instead of having no-repeat like in the original canvas test.

jrmuizel:
You can see Firefox's handling of this here: https://searchfox.org/mozilla-central/source/dom/canvas/CanvasRenderingContext2D.cpp#2524

pylbrecht:
@jrmuizel, thanks for your input. How would I create a `raqote::Source::Image` with 'no-repeat'? `ExtendMode` only provides `::Pad` and `::Repeat`.

jrmuizel:
So you could do like Firefox does at https://searchfox.org/mozilla-central/source/dom/canvas/CanvasRenderingContext2D.cpp#2524 and manually clamp the FillRect coordinates. But this doesn't work for Fill() and Stroke() and it turns out Firefox is actually broken for those cases. See https://bugzilla.mozilla.org/show_bug.cgi?id=1618998

Instead, the best thing to do might be to push a clip that bounds the pattern and then do the fill/stroke.

pylbrecht:
Correct me if I'm wrong, but I think we're already doing this for [`fill_rect()`](https://dxr.mozilla.org/servo/source/components/canvas/canvas_data.rs#476) by using [`size_from_pattern()`](https://dxr.mozilla.org/servo/source/components/canvas/raqote_backend.rs#31).
The basic tests for repeat/no-repeat are passing. These are the only failing tests left related to "repeat":
- https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.paint.repeaty.coord1.html
- https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.paint.repeatx.coord1.html
- https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord3.html

This makes me think the issue here might be something else, but currently I'm lost. Hopefully a full night's sleep will help with new ideas.

jrmuizel:
The rects here: https://dxr.mozilla.org/servo/source/components/canvas/canvas_data.rs#476 should be intersected. The current code that uses the origin of rect and the size from the pattern doesn't make any sense to me.

pylbrecht:
Thanks, that leaves me with enough stuff to digest and investigate!

pylbrecht:
> The rects here: https://dxr.mozilla.org/servo/source/components/canvas/canvas_data.rs#476 should be intersected. The current code that uses the origin of rect and the size from the pattern doesn't make any sense to me.

Since a pattern does not have an origin (only width and height), wouldn't we want to use `rect`'s origin as a starting point and then doing the checks for applying repetition if neccessary?

pylbrecht:
I'm still struggling with this. I just can't figure out why the expectation of https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.paint.norepeat.coord3.html is a green rectangle 100x50.

```js
// fill 100x50 with green starting a (0, 0)
ctx.fillStyle = '#0f0';
ctx.fillRect(0, 0, 100, 50);

// image's size: 100x50
var img = document.getElementById('red.png');
var pattern = ctx.createPattern(img, 'no-repeat');
ctx.fillStyle = pattern;
// move origin to (50, 25)
ctx.translate(50, 25);
// fill with red.png (100x50) starting at (0, 0) because of previous translate()
ctx.fillRect(-50, -25, 100, 50);

// fill 50x25 with green starting at (50, 25) because of previous translate()
ctx.fillStyle = '#0f0';
ctx.fillRect(0, 0, 50, 25);
```
This is what I would expect from this code:
example

What am I missing here?

pylbrecht:
Alright, I got it working now.
Thanks @jrmuizel for providing that link to Firefox's implementation of this. That was very helpful!

components/canvas/raqote_backend.rs
```diff
@@ -185,7 +168,7 @@ pub struct SurfacePattern<'a> {
     image: raqote::Image<'a>,
     filter: raqote::FilterMode,
     extend: raqote::ExtendMode,
-    repeat: Repetition,
+    pub repeat: Repetition,
```
pylbrecht: I wasn't sure whether to make this field public or to write a getter. I'm open for an idiomatic suggestion.

components/canvas/canvas_data.rs
```diff
@@ -470,12 +471,44 @@ impl<'a> CanvasData<'a> {
...
+                let limitx = |draw_rect: &mut Rect<f32>, pattern_rect: &Rect<f32>| {
+                    if draw_rect.size.width > pattern_rect.size.width {
+                        draw_rect.size.width = pattern_rect.size.width;
+                    }
+                };
+                let limity = |draw_rect: &mut Rect<f32>, pattern_rect: &Rect<f32>| {
+                    if draw_rect.size.height > pattern_rect.size.height {
+                        draw_rect.size.height = pattern_rect.size.height;
+                    }
+                };
```
jdm: I think we can replace the use of these with `draw_rect.size.width = draw_rect.size.width.min(pattern_rect.size.width)` and `draw_rect.size.height = draw_rect.size.height.min(pattern_rect.size.height)`

components/canvas/raqote_backend.rs
jdm: A getter seems reasonable to me.

components/canvas/canvas_data.rs
```diff
@@ -470,12 +471,44 @@ impl<'a> CanvasData<'a> {
...
+                        limitx(&mut draw_rect, &pattern_rect);
+                    },
+                    Repetition::Repeat => {
+                        draw_rect = *rect;
+                    },
+                }
+
+                draw_rect
+            },
+            _ => *rect,
```
jdm: Do we actually need this catch-all branch any more?

jdm (changes requested):
Nice improvement!

components/canvas/canvas_data.rs
pylbrecht: Yeah, that's better! Thanks

components/canvas/canvas_data.rs
pylbrecht: Well, there are more variants in `raqote_backend::Pattern`, so rustc friendly reminds me that those aren't handled. I rewrote the whole match to be more explicit.

components/canvas/raqote_backend.rs
```diff
@@ -30,27 +30,10 @@ impl Backend for RaqoteBackend {
 
     fn size_from_pattern(
```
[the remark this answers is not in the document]
pylbrecht: Yeah, makes sense. 👍

pylbrecht:
@bors-servo retry

```
  ▶ CRASH [expected OK] /_webgl/conformance/textures/misc/texture-size-cube-maps.html
  │ 
  │ 
  │ 
  │ 
  │ DESCRIPTION: Checks issues with size of cube map textures
  │ 
  │ max cube map size: 4096
  │ 
  │ testing size: 1
  │ PASS TEXTURE_CUBE_MAP_POSITIVE_X should be red
  │ PASS TEXTURE_CUBE_MAP_NEGATIVE_X should be green
  │ PASS TEXTURE_CUBE_MAP_POSITIVE_Y should be blue
  │ PASS TEXTURE_CUBE_MAP_NEGATIVE_Y should be yellow
  │ PASS TEXTURE_CUBE_MAP_POSITIVE_Z should be cyan
  │ PASS TEXTURE_CUBE_MAP_NEGATIVE_Z should be magenta
  │ PASS 0, 0, 2, 1 should be yellow
  │ FAIL 0, 1, 2, 1 should be cyan
  │ at (0, 1) expected: 0,255,255,255 was 255,255,0,255
  │ FAIL 0, 0, 2, 1 should be cyan
  │ at (0, 0) expected: 0,255,255,255 was 0,0,255,255
  │ PASS 0, 1, 2, 1 should be blue
  │ FAIL 0, 0, 2, 1 should be magenta
  │ at (0, 0) expected: 255,0,255,255 was 255,255,0,255
  │ PASS 0, 1, 2, 1 should be yellow
  │ PASS 0, 0, 2, 1 should be blue
  │ FAIL 0, 1, 2, 1 should be magenta
  │ at (0, 1) expected: 255,0,255,255 was 0,0,255,255
  │ called `Result::unwrap()` on an `Err` value: Disconnected (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at components/script/dom/webglrenderingcontext.rs:3178)
  │ stack backtrace:
  │              at src/libstd/panicking.rs:474
  │    3: rust_begin_unwind
  │              at src/libstd/panicking.rs:378
  │              at src/libcore/panicking.rs:85
  │    5: core::option::expect_none_failed
  │              at src/libcore/option.rs:1211
  │    6: <script::dom::webglrenderingcontext::WebGLRenderingContext as script::dom::bindings::codegen::Bindings::WebGLRenderingContextBinding::WebGLRenderingContextBinding::WebGLRenderingContextMethods>::ReadPixels
  │    7: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  │    8: _ZN3std9panicking3try7do_call17h8f4b2ccdea51442cE.llvm.6715725801538248291
  │    9: __rust_maybe_catch_panic
  │              at src/libpanic_unwind/lib.rs:86
  │   10: script::dom::bindings::codegen::Bindings::WebGLRenderingContextBinding::WebGLRenderingContextBinding::readPixels
  │   11: CallJitMethodOp
  │   12: script::dom::bindings::utils::generic_call
  │   13: <unknown>
  │   14: <unknown>
  │   15: <unknown>
  │   16: <unknown>
  │   17: <unknown>
  │   18: _ZL8EnterJitP9JSContextRN2js8RunStateEPh
  │              at /root/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/jit/Jit.cpp:103
  │       _ZN2js3jit13MaybeEnterJitEP9JSContextRNS_8RunStateE
  │              at /root/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/5906588/mozjs/js/src/jit/Jit.cpp:168
  │   20: <unknown>
  │   21: <unknown>
  │ [2020-03-03T06:42:56Z ERROR servo] called `Result::unwrap()` on an `Err` value: Disconnected
```

PR #25876, opened 2020-03-02
Upgrade mime crate to use the essence function

CYBAI:
While implementing module script, the `essence` function is not supported in mime crate yet so Manish filed the issue.

So, as that issue is fixed and shipped, we can upgrade mime crate to use the `essence_str` function.

I've tried to run https://threejs.org/examples/webgl_animation_cloth.html locally. Other than a bunch of `Unimplemented canvas2d.fillText`, I can see the animated cloth.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3465,12 +3465,9 @@ dependencies = [
 
 [[package]]
 name = "mime"
-version = "0.3.13"
+version = "0.3.16"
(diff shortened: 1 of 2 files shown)


Manishearth:
@bors-servo r+

thanks!

jdm:
@bors-servo retry
* #25522

Issue #25897, opened 2020-03-04
three.js dragging demo is really slow


jdm:
Unlike the other three.js demos with #25855 applied, the dragging demo still sits at 20 fps.

jdm:
A profile shows that a hugely disproportionate amount of time is being spent on the main thread running Servo::handle_events:
* lots under IOCompositor::perform_updates
  * MainThreadRegistry::run_one_frame even though openxr is running in its own thread
* a bunch under IOCompositor::receive_messages
* a bunch under EmbedderReceiver::try_recv_embedder_msg

I suspect what we're seeing here is #25305.

paulrouget:
In immersive mode?

paulrouget:
So the animation are ticking in the main thread and nothing is blocking it?

paulrouget:
I'll see why this is happening.

paulrouget:
@jdm does #25914 help?

jdm:
#25914 did not appear to improve the framerate at all.

jdm:
That being said, the profile with #25914 applied looks very different - the main thread has disappeared completely.

jdm:
I'm seeing a bunch of action in JS helper threads that are generating JIT code for Ion. I suspect if we turn on the right flags in spidermonkey we'll see that that's where a bunch of the work is going.

paulrouget:
> the main thread has disappeared completely

Yes. We were not reseting the animating state.

jdm:
Using the steps from https://github.com/servo/servo/wiki/Diagnosing-SpiderMonkey-JIT-issues I see there are frequent ion recompilations occurring, even outside of immersive mode with a static viewport.

jdm:
The two most common bailouts show the same symptom as the bug in https://github.com/servo/servo/issues/25971#issuecomment-600790184, so we should definitely reprofile once that's fixed.

Issue #25753, opened 2020-02-13
Intermittent pass in  /css/cssom-view/elementsFromPoint-inline-vlr-ltr.html


jdm:
```
  ▶ Unexpected subtest result in /css/cssom-view/elementsFromPoint-inline-vlr-ltr.html:
  └ PASS [expected FAIL] elementsFromPoint should return all elements under a point
```

PR #25914, opened 2020-03-06
Stop embedder calls and fake rAF when window not visible

paulrouget:
This addresses 2 issues:

- a rAF loop might still be ongoing when the window is invisible if script decided that the rAF were going too fast (spurious rAF)
- a hidden window does not run the rAF loop, but the embedder would still be in animating mode

---


paulrouget:
Even though this is not fixing #25897 this is still an improvement.

jdm:
@bors-servo retry
* #25753

PR #25678, opened 2020-02-03
Update SpiderMonkey

nox:
Do not merge.

PR #25920, opened 2020-03-06
Download usable macOS SDK during bootstrap

### LINKED ISSUE #25919: Can't build Servo on macOS 10.15
From the spidermonkey upgrade, I now get this build failure:
```
--- stderr
ERROR: SDK version "10.15" is unsupported. Please downgrade to version 10.14. You may need to point to it using --with-macos-sdk=<path> in your mozconfig. Various SDK versions are available from https://github.com/phracker/MacOSX-SDKs
make: *** [maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', /Users/joshmatthews/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/aabcc9b/build.rs:181:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
We're going to need to add some build configuration to support this to mozjs, and possibly make downloading the 10.14 SDK part of mach bootstrap on macos.

jdm:
This renames the `bootstrap` mach command to `bootstrap-build` because the bootstrap name is hardcoded into mach in a way that interferes with the actual command execution (#25224). These changes also support automatically setting up a 10.14 SDK on macOS 10.15, which is needed to build spidermonkey since #25678 merged.

Depends on https://github.com/servo/mozjs/pull/233.

- [x] These changes fix #25919
- [x] These changes do not require tests because there are no 10.15 test machines

### DIFF
--- a/python/servo/bootstrap.py
+++ b/python/servo/bootstrap.py
@@ -9,6 +9,7 @@ from distutils.version import LooseVersion
 import json
 import os
 import distro
+import platform
 import shutil
 import subprocess
 import six
@@ -18,7 +19,7 @@ from subprocess import PIPE
 from zipfile import BadZipfile
 
 import servo.packages as packages
-from servo.util import extract, download_file, host_triple
+from servo.util import extract, download_file, host_triple, check_call
 
 
 def install_trusty_deps(force):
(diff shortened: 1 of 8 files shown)


python/servo/bootstrap.py
```diff
@@ -391,6 +392,40 @@ def get_linux_distribution():
...
+
+    if not os.path.isdir(deps_dir):
+        os.makedirs(deps_dir)
+
+    sdk_path = os.path.join(deps_dir, TARGET_MACOS_SDK)
+    if not os.path.isdir(sdk_path):
+        sdk_file = TARGET_MACOS_SDK + ".tar.xz"
+        archive_path = os.path.join(deps_dir, sdk_file)
+        print('Downloading SDK for macOS 10.14...')
+        sdk_url = "https://github.com/phracker/MacOSX-SDKs/releases/download/10.15/" + sdk_file
```
nox: I'm not in favour of downloading and installing things off a random GH account, tbh. Why can't those be installed from Apple stuff?

nox (commented):
Seems ok but I'm pretty concerned about this.

python/servo/bootstrap.py
jdm: My assumption is that they're not available in a convenient place. Doesn't matter with your backport, though.

Issue #22312, opened 2018-11-29
Support for WebVTT format


ferjm:
Spec: https://w3c.github.io/webvtt

ferjm:
It seems that GStreamer has a parser for this format https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-base-plugins/html/gst-plugins-base-plugins-subparse.html

iulianR:
I talked with @jdm on Matrix and he said it's OK if I try to implement VTTRegion WebIDL interface. I managed to do that (some more tests pass, but I probably missed some things) and I pushed my changes [here](https://github.com/iulianR/servo/commit/92bdfaba1e5fac0f86d49d3087a55e20e28aa4a0). Adding VTTCue seems straightforward as well and I would like to do that as well, and afterwards see what else I can do.

Before starting on VTTCue I thought I would mention here in case someone else might think about starting work on this.

jdm:
Some brief investigation into how webkit + gstreamer integrates VTT:
* https://webkit-search.igalia.com/webkit/rev/57ed6cedb51e4f83d724ad2e147927cd6376e906/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp#2990-3013
* https://webkit-search.igalia.com/webkit/rev/57ed6cedb51e4f83d724ad2e147927cd6376e906/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp#1372-1461
* https://webkit-search.igalia.com/webkit/rev/57ed6cedb51e4f83d724ad2e147927cd6376e906/Source/WebCore/platform/graphics/gstreamer/InbandTextTrackPrivateGStreamer.cpp#98-124
* https://webkit-search.igalia.com/webkit/source/Source/WebCore/html/track/WebVTTParser.cpp
* it seems like the first step is setting the `text-sink` property on the pipeline, and then setting the up the sink object to handle the `new-sample` signal as well as accept appropriate caps for the VTT data
* after that, whenever a text sample is obtained, it needs to be parsed by a custom webvtt parser
* I wonder if we could reuse https://github.com/expenses/vtt for the parsing

jdm:
To be clear, the gstreamer-related work will need to happen in https://github.com/servo/media:
* the actual gstreamer bits will be part of https://github.com/servo/media/blob/master/backends/gstreamer/player.rs (presumably)
[...]

PR #25923, opened 2020-03-07
Add VTTRegion and part of VTTCue DOM interfaces

iulianR:
Hello! In this PR I implemented the VTTRegion DOM interface and part of VTTCue (#22312). Before continuing I thought it's maybe a good idea to first request a review or even merge what I did up to this point, as I might get stuck on the next part. I have a slight idea of what needs to be done (I assume it's hooking the GStreamer parser somewhere in `GetCueAsHTML()`), but a short outline or some more instructions would help a lot.

I updated test expectations by first running:
```
./mach test-wpt tests/wpt/web-platform-tests/webvtt/api --log-raw servo.log --pref dom.webvtt.enabled=true
```
then
```
./mach update-wpt servo.log
```
Thanks!
---

### DIFF
--- a/tests/wpt/metadata/webvtt/api/VTTCue/align.html.ini
+++ b/tests/wpt/metadata/webvtt/api/VTTCue/align.html.ini
@@ -1,0 +1,5 @@
+[align.html]
+  expected: TIMEOUT
+  [VTTCue.align, parsed cue]
+    expected: TIMEOUT
+
--- a/tests/wpt/metadata/webvtt/api/VTTCue/constructor.html.ini
+++ b/tests/wpt/metadata/webvtt/api/VTTCue/constructor.html.ini
@@ -1,0 +1,14 @@
+[constructor.html]
+  expected: CRASH
+  [VTTCue(), initial values]
+    expected: FAIL
+
+  [VTTCue(), bad end time]
+    expected: FAIL
+
+  [VTTCue(), bad start time]
+    expected: FAIL
+
+  [VTTCue(), text formatting]
+    expected: FAIL
+
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -294,6 +294,9 @@ mod gen {
                     event_polling_interval: i64,
                     test: bool,
                 },
+                webvtt: {
+                    enabled: bool,
+                },
                 webxr: {
                     #[serde(default)]
                     enabled: bool,
--- a/components/script/dom/mod.rs
+++ b/components/script/dom/mod.rs
@@ -538,6 +538,7 @@ pub mod vrfieldofview;
 pub mod vrframedata;
 pub mod vrpose;
 pub mod vrstageparameters;
+pub mod vttcue;
 pub mod vttregion;
 pub mod webgl_extensions;
 pub use self::webgl_extensions::ext::*;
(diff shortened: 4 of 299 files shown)


tests/wpt/metadata/webvtt/api/VTTCue/align.html.ini
```diff
@@ -0,0 +1,5 @@
+[align.html]
```
jdm: We should add a `__dir__.ini` in tests/wpt/metadata/webvtt like https://github.com/szeged/servo/blob/mmatyas__webgl_fns_framebuf_invalidate/tests/wpt/webgl/meta/conformance2/__dir__.ini that sets the new preference, and update the test results.

tests/wpt/metadata/webvtt/api/VTTCue/constructor.html.ini
```diff
@@ -0,0 +1,14 @@
+[constructor.html]
+  expected: CRASH
```
jdm: What's the source of this CRASH right now?

jdm (changes requested):
This base implementation looks good so far!

tests/wpt/metadata/webvtt/api/VTTCue/constructor.html.ini
iulianR: It's just a "not yet implemented crash" because `getCueAsHTML()` is called and there is a `todo!()` in its implementation.

tests/wpt/metadata/webvtt/api/VTTCue/align.html.ini
iulianR: I only updated the test results expectations for `webvtt/api`. Should I do it for all 317 tests in webvtt?

tests/wpt/metadata/webvtt/api/VTTCue/constructor.html.ini
jdm: How is it being called right now if the interfaces are gated by the dom.webvtt.enabled preference, which is disabled by default?

iulianR: I passed `--pref dom.webvtt.enabled=true` to 'test-wpt` when testing the interfaces.

tests/wpt/metadata/webvtt/api/VTTCue/align.html.ini
jdm: I recommend adding `[webvtt]` and `[api]` entries to https://github.com/servo/servo/blob/master/tests/wpt/include.ini so we only run the tests that are most relevant while we're in the early stages.

iulianR: Does that .ini file has any effect if I

iulianR: Does the .ini file has any effect if I run the tests with a path? There still seems to be the same number of tests (317) that are being run.
`$ ./mach test-wpt tests/wpt/web-platform-tests/webvtt/`
```
@@ -175,6 +175,10 @@ skip: true
   skip: false
 [webvr]
   skip: false
+[webvtt]
+  skip: true
+  [api]
+    skip: false
 [webxr]
   skip: false
 [WebIDL]

```

jdm: Running with a path ignores the contents of include.ini.

iulianR: I hope I got it right. This is all magic for me now.

PR #25918, opened 2020-03-06
Don't compare vtable pointers anymore

nox:

PR #25939, opened 2020-03-10
Update to 3/4 nightly rustc.

jdm:
Now that #25918 has merged, this should be safe to use.


jdm:
@bors-servo retry
* #21067

PR #25916, opened 2020-03-06
Render HTML <canvas> elements on layout 2020

ferjm:
- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #25885
- [X] There are tests for these changes

Issue #25875, opened 2020-02-29
Intermittent pass in /css/css-text/white-space/trailing-other-space-separators-break-spaces-006.html


jdm:
```
  ▶ PASS [expected FAIL] /css/css-text/white-space/trailing-other-space-separators-break-spaces-006.html
```

PR #25940, opened 2020-03-10
Handle nonexistent images in CanvasRenderingContext2D.createPattern()

pylbrecht:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5754,7 +5754,7 @@ dependencies = [
 [[package]]
 name = "surfman-chains"
 version = "0.3.0"
-source = "git+https://github.com/asajeffrey/surfman-chains?branch=multi#80a71b1a2df71ae70c3c194d0af40b8ebf72968a"
+source = "git+https://github.com/asajeffrey/surfman-chains?branch=multi#e775b8c7807659958a4f20cf8e6eca4290f35124"
 dependencies = [
  "euclid",
  "fnv",
--- a/components/script/dom/htmlimageelement.rs
+++ b/components/script/dom/htmlimageelement.rs
@@ -167,13 +167,10 @@ impl HTMLImageElement {
     // https://html.spec.whatwg.org/multipage/#check-the-usability-of-the-image-argument
     pub fn is_usable(&self) -> Fallible<bool> {
         // If image has an intrinsic width or intrinsic height (or both) equal to zero, then return bad.
-        match &self.current_request.borrow().image {
-            Some(image) => {
-                if image.width == 0 || image.height == 0 {
-                    return Ok(false);
-                }
-            },
-            None => return Ok(false),
+        if let Some(image) = &self.current_request.borrow().image {
+            if image.width == 0 || image.height == 0 {
+                return Ok(false);
+            }
         }
 
         match self.current_request.borrow().state {
--- a/tests/wpt/metadata-layout-2020/2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html.ini
+++ b/tests/wpt/metadata-layout-2020/2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html.ini
@@ -1,0 +1,4 @@
+[2d.drawImage.broken.html]
+  [Canvas test: 2d.drawImage.broken]
+    expected: FAIL
+
--- a/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html.ini
+++ b/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html.ini
@@ -1,0 +1,4 @@
+[2d.pattern.image.broken.html]
+  [Canvas test: 2d.pattern.image.broken]
+    expected: FAIL
+
--- a/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html.ini
+++ b/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html.ini
@@ -1,4 +1,0 @@
-[2d.pattern.image.nonexistent.html]
-  [Canvas test: 2d.pattern.image.nonexistent]
-    expected: FAIL
-
--- a/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html.ini
+++ b/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html.ini
@@ -1,0 +1,4 @@
+[2d.pattern.image.zeroheight.html]
+  [Canvas test: 2d.pattern.image.zeroheight]
+    expected: FAIL
+
--- a/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html.ini
+++ b/tests/wpt/metadata-layout-2020/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html.ini
@@ -1,0 +1,4 @@
+[2d.pattern.image.zerowidth.html]
+  [Canvas test: 2d.pattern.image.zerowidth]
+    expected: FAIL
+
--- a/tests/wpt/metadata/2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html.ini
+++ b/tests/wpt/metadata/2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html.ini
@@ -1,0 +1,4 @@
+[2d.drawImage.broken.html]
+  [Canvas test: 2d.drawImage.broken]
+    expected: FAIL
+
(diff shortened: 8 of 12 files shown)


pylbrecht:
Unfortunately these changes introduce 3 test failures:
- [`2d.pattern.image.zeroheight.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html)
- [`2d.pattern.image.zerowidth.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html)
- [`2d.pattern.image.broken.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html)

The code sees the passed `image` argument as "broken" and consequently throws an `InvalidStateError` instead of returning `null` as expected by all tests above.

Although to me this seems the correct way of [handling the image argument according to the spec](https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument).
All other browsers (except Firefox) fail these tests as well:
- https://wpt.fyi/results/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html?label=master&label=experimental&aligned&q=pattern.image.zero
- https://wpt.fyi/results/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html?label=master&label=experimental&aligned&q=pattern.image.zero
- https://wpt.fyi/results/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html?label=master&label=experimental&aligned&q=pattern.image.broken

Currently it seems like just exchanging one (or more) test passes against another.
@jdm, what do you think?

jdm:
It's not clear to me why we're removing is_usable, given that it seems to implement https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument which is the first step of https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-createpattern. What is the reason that we don't throw an exception for 2d.pattern.image.nonexistent.html right now?

pylbrecht:
It's not really removed, but rather split up and moved into `create_pattern()`:

The spec states three steps:

**1. If image's current request's state is broken, then throw an "InvalidStateError" DOMException.**
  This is handled by `is_available()`, which returns `Err(Error::InvalidState)` if the current request's state is broken.
**2. If image is not fully decodable, then return bad.**
  As far as I understand this requires fetching the image data and then checking against the result, which is done [here](https://github.com/servo/servo/pull/25940/files#diff-587cd504daeade12b5ea90149c65788dR909-R917).
**3. If image has an intrinsic width or intrinsic height (or both) equal to zero, then return bad.**
  This is handled [here](https://github.com/servo/servo/pull/25940/files#diff-587cd504daeade12b5ea90149c65788dR912-R914).

I moved the check of **2.** to `create_pattern()`, because I was not sure if `HTMLImageElement.current_request.image` is the same image instance as the one fetched from the `CanvasState`'s image cache with `self.fetch_image_data()`. So I went with using the one fetched from the image cache and modified the code so that the checks are performed on that image.

pylbrecht:
Also the previous `is_usable()` did not pass [`2d.pattern.image.nonexistent.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html), which was the main motivation for changing the code.

pylbrecht:
Returning `null` instead of throwing an `InvalidStateError` if the image's current request is broken results in this:
```
▶ find tests/wpt/web-platform-tests/2dcontext -name "*pattern*image*" | xargs -o ./mach test-wpt --debug-build --headless --log-grouped -
Running 21 tests in web-platform-tests

  ▶ Unexpected subtest result in /2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html:
  └ PASS [expected FAIL] Canvas test: 2d.pattern.image.broken

  ▶ Unexpected subtest result in /2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html:
  │ FAIL [expected PASS] Canvas test: 2d.pattern.image.nonexistent
  │   → assert_throws_dom: function "function() { ctx.createPattern(img, 'repeat'); }" did not throw
  │
  │ @http://web-platform.test:8000/2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html:23:18
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1956:25
  │ _addTest/</<@http://web-platform.test:8000/2dcontext/resources/canvas-tests.js:66:15
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1956:25
  └ _addTest/<@http://web-platform.test:8000/2dcontext/resources/canvas-tests.js:63:11

  ▶ Unexpected subtest result in /2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html:
  └ PASS [expected FAIL] Canvas test: 2d.pattern.image.zerowidth

  ▶ Unexpected subtest result in /2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html:
  └ PASS [expected FAIL] Canvas test: 2d.pattern.image.zeroheight

Ran 21 tests finished in 44.6 seconds.
  • 17 ran as expected. 0 tests skipped.
  • 4 tests had unexpected subtest results
```
Either my implementation is flawed and I don't see it or these test cases are mutually exclusive.

jdm:
I'm assuming that 2d.pattern.image.nonexistent.html does not pass on master because we do not throw an exception right now. If that's the case, why does the current code on master not cause an exception to be thrown?

jdm:
Per a discussion in Matrix, I suspect we're actually working around the fact that [this branch](https://github.com/servo/servo/blob/e3c91f7c4919e96e80c5d0ef8e9acb3e4461b1c3/components/script/dom/htmlimageelement.rs#L176) is unnecessarily returning early. If we make that an `if let`, then we should hit the Broken state and throw the correct exception.

pylbrecht:
This indeed fixes [`2d.pattern.image.nonexisting.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.nonexistent.html), but again fails:
- [`2d.pattern.image.zeroheight.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html)
- [`2d.pattern.image.zerowidth.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html)
- [`2d.pattern.image.broken.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html)

The code considers the images in those failing tests as broken and consequently throws an `InvalidStateError` instead of returning `null`.

pylbrecht:
I pushed the last idea and ran CI to provide more context.

pylbrecht:
We could put a check if the image exists in [this match arm](https://dxr.mozilla.org/servo/source/components/script/dom/htmlimageelement.rs#181) and return `Ok(false)` if it does exist, `Err(Error::InvalidState)` otherwise.

pylbrecht:
In both of these tests, the image is a `.svg` file:
- [`2d.pattern.image.zeroheight.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html)
- [`2d.pattern.image.zerowidth.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html)
Could that be the reason the code interprets them as broken, because Servo has no SVG implementation yet?

jdm:
Yes, that would do it.

pylbrecht:
So we just mark these tests as expected fails and move on? Or was your comment a reply to the idea in https://github.com/servo/servo/pull/25940#issuecomment-597257917 ?

jdm:
If the only new failures are in tests that rely on svg support for meaningful results, then market by those as expected failures makes sense.

pylbrecht:
Well, there is [2d.pattern.image.broken.html](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.broken.html), which is not a `.svg` image:
```
  ▶ Unexpected subtest result in /2dcontext/drawing-images-to-the-canvas/2d.drawImage.broken.html:
  │ FAIL [expected PASS] Canvas test: 2d.drawImage.broken
  └   → The object is in an invalid state.
```
There's also this issue https://github.com/web-platform-tests/wpt/issues/21685, which makes me unsure which expectation we should fulfill: returning `null` or throwing an `InvalidStateError`

pylbrecht:
Also the passed `.svg` images in [`2d.pattern.image.zeroheight.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zeroheight.html) and [`2d.pattern.image.zerowidth.html`](https://github.com/web-platform-tests/wpt/blob/master/2dcontext/fill-and-stroke-styles/2d.pattern.image.zerowidth.html) are converted to `HTMLImageElement`s in the code. I have not checked the code yet, but I assume `.svg` images are converted to `HTMLImageElement` due to missing svg support.

pylbrecht:
@bors-servo retry

```
  ▶ PASS [expected FAIL] /css/css-text/white-space/trailing-other-space-separators-break-spaces-005.html
```
Might belong to https://github.com/servo/servo/issues/25875

jdm:
@bors-servo r+
Thanks!

jdm:
Hmm, ever since #25916 we're running the 2dcontext tests with the layout-2020 builds as well, so we get WPT failures for metadata that needs updating. Since the vast majority of those tests are testing the DOM interface rather than using reftests which would test the layout, I think we should remove https://github.com/servo/servo/blob/master/tests/wpt/include-layout-2020.ini#L8 to avoid this issue.

Issue #25476, opened 2020-01-09
Intermittent panic in webgl tests on linux ("All senders for this socket closed") ("Err` value: Disconnected") from read_pixels_into


jdm:
```
  ▶ CRASH [expected OK] /_webgl/conformance2/sync/sync-webgl-specific.html
  │ 
  │ 
  │ 
  │ 
  │ called `Result::unwrap()` on an `Err` value: Io(Custom { kind: ConnectionReset, error: "All senders for this socket closed" }) (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at src/libcore/result.rs:1188)
  │ stack backtrace:
  │    0: servo::backtrace::print
  │    1: servo::main::{{closure}}
  │    2: std::panicking::rust_panic_with_hook
  │              at src/libstd/panicking.rs:475
  │    3: rust_begin_unwind
  │              at src/libstd/panicking.rs:375
  │    4: core::panicking::panic_fmt
  │              at src/libcore/panicking.rs:84
  │    5: core::result::unwrap_failed
  │              at src/libcore/result.rs:1188
  │    6: _ZN6script3dom22webgl2renderingcontext22WebGL2RenderingContext16read_pixels_into17h0143b0f4bd78b1dfE.llvm.8397337333947771064
  │    7: <script::dom::webgl2renderingcontext::WebGL2RenderingContext as script::dom::bindings::codegen::Bindings::WebGL2RenderingContextBinding::WebGL2RenderingContextBinding::WebGL2RenderingContextMethods>::ReadPixels
  │    8: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  │    9: _ZN3std9panicking3try7do_call17h0909fbd65a236234E.llvm.12806476980974741674
  │   10: __rust_maybe_catch_panic
  │              at src/libpanic_unwind/lib.rs:78
  │   11: script::dom::bindings::codegen::Bindings::WebGL2RenderingContextBinding::WebGL2RenderingContextBinding::readPixels
  │   12: <unknown>
  │   13: <unknown>
  │ 
  │ [2020-01-09T18:54:10Z ERROR servo] called `Result::unwrap()` on an `Err` value: Io(Custom { kind: ConnectionReset, error: "All senders for this socket closed" })
  │ Pipeline failed in hard-fail mode.  Crashing!
  └ thread panicked while processing panic. aborting.
```
[...]

PR #25945, opened 2020-03-11
layout_2020: Paint hoisted positioned fragments in tree order

mrobinson:
Instead of painting hoisted position fragments in the order to which
they are hoisted, paint them in tree order and properly incorporate them
into the stacking context.

We do this by creating a placeholder fragment in the original tree position
of hoisted fragments. The ghost fragment contains an atomic id which
links back to the hoisted fragment in the containing block.

While building the stacking context, we keep track of containing blocks
and their children. When encountering a placeholder fragment we look at
the containing block's hoisted children in order to properly paint the
hoisted fragment.

One notable design modification in this change is that hoisted fragments
no longer need an AnonymousFragment as their parent. Instead they are
now direct children of the fragment that establishes their containing block.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2919,6 +2919,7 @@ dependencies = [
  "html5ever",
  "ipc-channel",
  "libc",
+ "log",
  "mitochondria",
  "msg",
  "net_traits",
--- a/components/layout_2020/Cargo.toml
+++ b/components/layout_2020/Cargo.toml
@@ -25,6 +25,7 @@ gfx_traits = {path = "../gfx_traits"}
 html5ever = "0.25"
 ipc-channel = "0.14"
 libc = "0.2"
+log = "0.4"
 msg = {path = "../msg"}
 mitochondria = "1.1.2"
 net_traits = {path = "../net_traits"}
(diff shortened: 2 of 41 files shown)


nox:
Any way you could avoid IDs and use an `Arc<_>` instead?

mrobinson:
My original idea was to use a thread-safe reference-counted type, but I believe that @SimonSapin suggested the id approach.

mrobinson:
Ah, just to clarify the main issue is that we could need to create a data structure in the Rayon task for the tree-order-parent and modify it in the containing-block-parent.

nox:
I don't understand what that means. Can't the two separate boxes point to the same content behind an `Arc<_>`?

mrobinson:
Perhaps I'm misunderstanding what you are suggesting. I could create a `Fragment` when processing the original parent and wrap it in an `Arc<_>`. Layout of this `Fragment` happens later, in the ancestor which establishes the containing block. This layout wouldn't be possible though because the hoisted `Arc<Fragment>` since it is now immutable. Adding to this situation is the fact that the ancestor establishing the containing block and the descendant parent may be laid out in different rayon tasks.

jdm:
@bors-servo retry
* #25476

jdm:
@bors-servo retry
* #25961

jdm:
@bors-servo retry
* #25961

jdm:
@bors-servo retry
* #25875 #23290

PR #25948, opened 2020-03-11
Accept only connected attributes for checkbox and radio activation behavior

### LINKED ISSUE #25906: Checkbox/radio inputs should check connectedness for input/change events
https://github.com/whatwg/html/pull/5335/

I suspect that [this](https://github.com/servo/servo/blob/ca25e18791163de5c6af1267d876c7f5bb705f61/components/script/dom/htmlinputelement.rs#L2487-L2494) is the relevant code we'll need to adjust.

NeverHappened:
Accept only connected attributes for checkbox and radio activation behavior

---

- [x] These changes fix #25906 
- [x] WPT tests were already here


jdm:
@bors-servo r+
Looks good! Thanks for fixing this!

Issue #25955, opened 2020-03-12
Update WebIDL.py for python3 support


jdm:
https://github.com/servo/servo/blob/master/components/script/dom/bindings/codegen/parser/update.sh

dralley:
I think the attached PR closes this issue. There was some discussion about other items that are missing support, but those would be tracked elsewhere.

Meta-bug https://github.com/servo/servo/issues/23607

jdm:
Indeed!

PR #25964, opened 2020-03-14
Update the WebIDL parser

nox:
It now supports Python 3.


CYBAI:
Maybe this will fix #25955 👀?

nox:
It updates WebIDL.py but it doesn't make the script crate work with Python 3. And mozjs_sys and style also don't build with Python 3.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1665,6 +1654,8 @@ def __init__(self, descriptor, name, static, unforgeable):
             (maplikeOrSetlikeOrIterable and
              maplikeOrSetlikeOrIterable.isIterable() and
              maplikeOrSetlikeOrIterable.isValueIterator())):
+            m = maplikeOrSetlikeOrIterable
```
jdm: What is the purpose of this change?

jdm:
Not super confidence-boosting that fixing the scoping issue triggered no test result changes in idlharness tests. Oh well!
@bors-servo r+

nox:
> Not super confidence-boosting that fixing the scoping issue triggered no test result changes in idlharness tests. Oh well!

AFAIK this block of code only uses object properties that are common to all descriptors of the interface for which we are generating stuff. To make it an actual bug, you need to do some different `[Pref]` attributes on different interface members and whatnot, with the last one getting different gates than the ones you want for the iterator symbols.

jdm:
@bors-servo retry
* #25513

PR #25969, opened 2020-03-16
Use matrix for TC notifications.

jdm:


SimonSapin:
Diff looks OK but it looks like the TC side of this is not fully set up yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1618384 is still open and nothing happened in the channel when I tried https://community-tc.services.mozilla.com/tasks/fLlGZ_4DQO-e88Dh_Kl2xg/runs/0 manually.

Additionally I’d like to make this configurable so we can have different nicknames mentioned in the notifications for different tasks (CC https://github.com/servo/servo/pull/24889#pullrequestreview-323864585)

And before landing we should check whether including a nickname in a message is enough to make Matrix client trigger a "you were mentioned" notification, or if we need some kind of markup in `task.extra.notify.matrixFormattedBody` (see https://docs.taskcluster.net/docs/reference/core/notify/usage)

SimonSapin:
TC now can now send to Matrix but uses "notices", which are ignored by Riot for the purpose of triggering desktop notifications. I’ve filed https://github.com/taskcluster/taskcluster/pull/2628 to add an opt out and make it send "normal" messages instead.

jdm:
We're doing ok without these so far.

PR #25745, opened 2020-02-12
make sure UWP app can run

paulrouget:
Fix #25718

This includes https://github.com/servo/servo/pull/25661 with some minor tweaks (certificate file reference was missing, and its id was wrong) and a python script.

Still need to hook that to our CI.

PR #25972, opened 2020-03-17
Disable running windows binaries on CI.

jdm:
In order to get more information out of https://github.com/servo/servo/pull/25745, we're going to temporarily disable the parts of windows CI that don't agree with the updated windows server environment.


CYBAI:
@bors-servo retry

- #21067, #25973

PR #25979, opened 2020-03-17
Implementation of doge API

ANURADHAJHA99:
We have changed the files 
- components/script/dom/doge.rs  
- components/script/dom/mod.rs  
- components/script/dom/webidls/Doge.webidl

for implementing the doge API. however it is not working.

### DIFF
--- a/components/script/dom/doge.rs
+++ b/components/script/dom/doge.rs
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ use dom::bindings::cell::DOMRefCell;
+ use dom::bindings::codegen::Bindings::DogeBinding::{DogeMethods, DogeInit, Wrap as DogeWrap};
+ use dom::bindings::error::{Error, Fallible};
+ use dom::bindings::global::GlobalRef;
+ use dom::bindings::js::Root;
+ use dom::bindings::reflector::{Reflector, reflect_dom_object};
+ use dom::bindings::str::ByteString;
+ use rand;
+ use rand::Rng;
+ 
+ #[dom_struct]
+ pub struct Doge {
+     reflector_: Reflector,
+     such_list: DOMRefCell<Vec<ByteString>>,
+ }
+ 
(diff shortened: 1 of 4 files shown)


components/script/dom/doge.rs
```diff
@@ -0,0 +1,67 @@
...
+ 
+ impl Doge {
+     pub fn new_inherited() -> Doge {
+         Doge {
+             reflector_: Reflector::new(),
+             such_list: DOMRefCell::new(vec![]),
+         }
+     }
+ 
+     pub fn new(global: GlobalRef) -> Root<Doge> {
```
jdm: A couple issues that I notice which we should update in our blog post:
* we now use `&GlobalScope` instead of `GlobalRef`
* we now use `DomRoot<Doge>` instead of `Root<Doge>`

jdm (commented):
It would help if you could describe in more detail what errors you see.

components/script/dom/webidls/Doge.webidl
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// https://jeenalee.github.io/doge-standard/
+
+typedef sequence<DOMString> DogeInit;
+
+[Constructor(optional DogeInit init),
```
jdm: We now use a different constructor syntax that looks like [this](https://github.com/servo/servo/blob/59c68e2eb7afdda17fc27e866360a1746a78bbda/components/script/dom/webidls/AnalyserNode.webidl#L18) instead. You can see how the transition happened in https://github.com/servo/servo/commit/9ce82ea1aedfd3eddad0eed9fac6718599a25ec3 .

components/script/dom/doge.rs
jdm: Actually, it looks like you're following an older revision of https://jeenalee.com/2016/10/03/implementing-doge-for-servo.html. What page were you using as a reference?

prashi23: Ok! Thanks @jdm. Now, we got it. We will resolve this and then we will write this in our blog. Just to keep a log of things that we are working on, we are writing our own team blog. :)

PR #25984, opened 2020-03-19
layout_2020: Refactor PositioningContext to be simpler and smaller

mrobinson:
Add a few helper methods which allow removing duplicate code in
PositioningContext. These methods will also be used to properly
implement hoisting in inline layout.

---

- [x] These changes do not require tests because this shouldn't change any behavior.


mrobinson:
@pcwalton @nox r?

Issue #25233, opened 2019-12-10
Python error when reporting non-intermittent failures on mac


jdm:
From https://community-tc.services.mozilla.com/tasks/B5l6bUP7Rsqe2TNtsAkKSQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FB5l6bUP7Rsqe2TNtsAkKSQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log

```
+ ./mach filter-intermittents wpt-jsonsummary.log --log-intermittents intermittents.log --log-filteredsummary filtered-wpt-errorsummary.log --tracker-api default --reporter-api default
Error running mach:
    ['filter-intermittents', 'wpt-jsonsummary.log', '--log-intermittents', 'intermittents.log', '--log-filteredsummary', 'filtered-wpt-errorsummary.log', '--tracker-api', 'default', '--reporter-api', 'default']
The error occurred in code that was called by the mach command. This is either
a bug in the called code itself or in the way that mach is calling it.
You should consider filing a bug for this issue.
If filing a bug, please include the full output of mach, including this error
message.
The details of the failure are as follows:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 57: ordinal not in range(128)
  File "/Users/worker/tasks/task_1575999096/repo/python/servo/testing_commands.py", line 580, in filter_intermittents
    format(actual_failures, description)
  File "/Users/worker/tasks/task_1575999096/repo/python/servo/testing_commands.py", line 568, in format
    file.write(formatted.encode("utf-8"))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/codecs.py", line 357, in write
    data, consumed = self.encode(object, self.errors)
```

cc @SimonSapin

SimonSapin:
Like before, the `filtered-wpt-errorsummary.log` artifacts contain the same information that the failing code would print. So we can look there until this is fixed.

mrobinson:
The code around intermittents has changed a lot so this issue likely no longer relevant.

PR #25985, opened 2020-03-19
layout_2020: Add support for hoisting positioned fragments in inline boxes

### LINKED ISSUE #25279: `display: inline; position: relative` as containing block for abspos descendants
https://drafts.csswg.org/css2/visudet.html#containing-block-details
> If the element has 'position: absolute', the containing block is established by the nearest ancestor with a 'position' of 'absolute', 'relative' or 'fixed', in the following way:
>
> In the case that the ancestor is an inline element, the containing block is **the bounding box around the padding boxes of the first and the last inline boxes generated for that element**. In CSS 2, if the inline element is split across multiple lines, the containing block is undefined. 

Even though the spec is written as though finding the containing block is an operation that involves going up the tree, as of this writing the Layout 2020 code maintains a "current containing block for `position: absolute`" while going down the box tree while creating a fragment tree.

As of https://github.com/servo/servo/pull/25273, the intent of `for_maybe_position_relative` in `components/layout_2020/positioned.rs` was that it would be called for ~every box and take care of doing that maintenance in case that box has `position: relative`.

For inline boxes however, `components/layout_2020/flow/inline.rs` contains separate `start_layout` and `finish_layout` methods (to deal with line breaking), so there isn’t a single place to easily call `for_maybe_position_relative`. Additionally the `InlineFormattingContextState` struct contains a `&mut PositioningContext` field, so we can’t easily create a new `PositioningContext` and keep it on the call stack for each nesting level, like we do in block-level layout code. Instead we should probably take/replace the `for_nearest_positioned_ancestor` Vec inside that `PositioningContext`.

https://github.com/servo/servo/pull/25273 was large enough already so I’m leaving this for later, with this issue to track.

mrobinson:
Add support for tracking containing blocks when doing inline layout.
This requires setting up a PositioningContext for inline boxes when
necessary. Instead of using the PositioningContext helper methods
and we reuse the contexts between line breaks.

Fixes #25279.

---

- [x] These changes fix #25279

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -12,7 +12,10 @@ use crate::fragments::{
     DebugId, Fragment, TextFragment,
 };
 use crate::geom::flow_relative::{Rect, Sides, Vec2};
-use crate::positioned::{relative_adjustement, AbsolutelyPositionedBox, PositioningContext};
+use crate::positioned::{
+    relative_adjustement, AbsolutelyPositionedBox, HoistedAbsolutelyPositionedBox,
+    PositioningContext,
+};
 use crate::sizing::ContentSizes;
 use crate::style_ext::{ComputedValuesExt, Display, DisplayGeneratingBox, DisplayOutside};
 use crate::ContainingBlock;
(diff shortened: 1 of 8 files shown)


nox:
Nice.

@bors-servo r+

mrobinson:
@bors-servo retry
  *  #25522

mrobinson:
@bors-servo retry
  *  #25233

PR #25992, opened 2020-03-19
Index tasks by git tree hash instead of parent commits hashes

SimonSapin:
This makes it (sligthly) easier to query the index and find those tasks from other scripts.


jdm:
@bors-servo r+
Nice simplification!

SimonSapin:
```
URL parameter 'indexPath' given as 'project.servo.linux_unit.7ebef84b0951e3ceca8081a8eaf14e82351af7d4
' must match regular expression: '/^([a-zA-Z0-9_!~*'()%-]+\.)*[a-zA-Z0-9_!~*'()%-]+$/'
```
Note the newline before the `'` that marks the end of the `indexPath` value.

Added a `.strip()` call to remove it from the `git show` output.

@bors-servo r=jdm

Issue #25564, opened 2020-01-21
Scrollable overflow calculation should take into account containing block


mrobinson:
The containing block is necessary in order to properly convert from logical geometry to physical geometry. This is currently tricky because we don't necessarily have the containing block when calculating scrollable overflow. On the other hand, it isn't clear if that is necessary since only the containing block size is used in layout 2014.

PR #26012, opened 2020-03-23
layout_2020: Use the containing block more when calculating scrolling overflow

mrobinson:
When calculating scrolling overflow calculation we cannot currently use
the actual containing block in all cases. This change increases the
amount that we do use the containing block.

---


mrobinson:
By the way, this is part of #25564.

mrobinson:
@bors-servo retry
 * #25233

PR #26025, opened 2020-03-24
Add support for DrawRangeElements in WebGL2

imiklos:
Adds initial support for the WebGL2 `DrawRangeElements` call.

I have started working on this function, but not sure how could I check for the Uniform Block Backing (https://www.khronos.org/registry/webgl/specs/latest/2.0/#ACTIVE_UNIFORM_BLOCK_BACKING).
I am looking for some advice.

---

cc @mmatyas @zakorgy @jdm


jdm:
This isn't an answer yet, but I believe Firefox's check for active uniform buffer sizes lives in [this code](https://searchfox.org/mozilla-central/source/dom/canvas/WebGLContextDraw.cpp#369-392).

jdm:
We record similar program link info [here](https://github.com/servo/servo/blob/654dd048a3fe4578883ce38616199f06cf850111/components/canvas/webgl_thread.rs#L2118) and store it in WebGLProgram in [here](https://github.com/servo/servo/blob/cf725fc3f59bc1b94e3b35ab4cfe60b3d41559cc/components/script/dom/webglprogram.rs#L157-L191).

imiklos:
@jdm Thank you! I'm working on it. I will update the pull request when I got some progress.

imiklos:
I've just updated the pull request, squashed the commits and updated the test expectations also.

components/script/dom/webgl2renderingcontext.rs
```diff
@@ -191,6 +191,48 @@ impl WebGL2RenderingContext {
         self.base.current_vao_webgl2()
     }
 
+    pub fn validate_uniform_block_for_draw(&self) {
+        if let Some(program) = self.base.current_program() {
```
jdm: Let's write this as:
```rust
let program = match self.base.current_program() {
    Some(p) => p,
    None => return,
};
```

components/script/dom/webgl2renderingcontext.rs
```diff
@@ -191,6 +191,48 @@ impl WebGL2RenderingContext {
...
+            program
+                .active_uniform_blocks()
+                .iter()
+                .for_each(|uniform_block| {
```
jdm: Let's write this as:
```rust
for uniform_block in &program.active_uniform_blocks() {
    //...
}
```

components/script/dom/webgl2renderingcontext.rs
```diff
@@ -191,6 +191,48 @@ impl WebGL2RenderingContext {
...
+                    let data_size = uniform_block.size as usize;
+                    program.active_uniforms().iter().for_each(|block| {
```
jdm: Let's write this as `for block in &program.active_uniforms() {`

components/script/dom/webgl2renderingcontext.rs
```diff
@@ -191,6 +191,48 @@ impl WebGL2RenderingContext {
...
+                        if let Some(index) = block.bind_index {
```
jdm: Let's write this as:
```rust
let index = match block.bind_index {
    Some(index) => index,
    None => continue,
};
```

components/script/dom/webgl2renderingcontext.rs
```diff
@@ -191,6 +191,48 @@ impl WebGL2RenderingContext {
...
+                            let indexed = &self.indexed_uniform_buffer_bindings[index as usize];
+                            match indexed.buffer.get() {
```
jdm:
```rust
let buffer = match indexed.buffer.get() {
    Some(buffer) => buffer,
    None => {
        self.base.webgl_error(InvalidOperation);
        return;
    },
};
```

components/script/dom/webglprogram.rs
```diff
@@ -574,6 +574,11 @@ impl WebGLProgram {
             return Err(WebGLError::InvalidValue);
         }
 
+        let mut active_uniforms = self.active_uniforms.borrow_mut();
+        if active_uniforms.len() > block_binding as usize {
+            active_uniforms[block_binding as usize].bind_index = Some(block_binding);
```
jdm: Is there any point when we should reset this to None?

jdm (changes requested):
Looking good! Let's try to reduce the rightward-drift in the validation code, though.

components/script/dom/webglprogram.rs
imiklos: I'm not sure, but I didn't find any circumstance when we should set this to None except the initial state.

imiklos:
@jdm Thank you for the review! I've addressed what you mentioned and updated the pull request.

jdm:
@bors-servo retry
* #25513

jdm:
@bors-servo retry
* #25513

Issue #11546, opened 2016-06-01
Deadlock when removing iframe from document.


emilio:
I think the fact that the iframe is not yet completely loaded might influence in the result, but not totally sure.

Testcase:

```html
<!doctype html>
<title></title>
<script>
window.addEventListener('load', function() {
  var frame = null;
  setTimeout(function f() {
    if (frame)
      document.body.removeChild(frame);
    frame = document.createElement('iframe');
    frame.src = "data:text/html,<p>test</p><script>console.log('test')<" + "/script>";
    document.body.appendChild(frame);
    setTimeout(f, 1000);
  }, 1000);
}, false);
</script>
<div>test</div>
```

emilio:
Some backtraces of affected threads:

## Constellation

Constellation is waiting on `Pipeline::exit()`.

```
(gdb) bt
#0  0x00007ffff59db98d in recvmsg () at ../sysdeps/unix/syscall-template.S:84
#1  0x0000555558a8ae62 in ipc_channel::platform::linux::{{impl}}::recv (self=0x7fffe89f6f98, fd=70, blocking_mode=Blocking)
    at /home/emilio/projects/moz/servo/.cargo/git/checkouts/ipc-channel-8c79851162410a92/master/platform/linux/mod.rs:923
#2  0x0000555558a7b7d5 in ipc_channel::platform::linux::recv (fd=70, blocking_mode=Blocking)
    at /home/emilio/projects/moz/servo/.cargo/git/checkouts/ipc-channel-8c79851162410a92/master/platform/linux/mod.rs:761
#3  0x0000555558a79840 in ipc_channel::platform::linux::{{impl}}::recv (self=0x7fffe89f7d38)
    at /home/emilio/projects/moz/servo/.cargo/git/checkouts/ipc-channel-8c79851162410a92/master/platform/linux/mod.rs:86
#4  0x0000555557c21a3f in ipc_channel::ipc::{{impl}}::recv<()> (self=0x7fffe89f7d38)
    at /home/emilio/projects/moz/servo/.cargo/git/checkouts/ipc-channel-8c79851162410a92/master/ipc.rs:73
#5  0x0000555557c214da in constellation::pipeline::{{impl}}::exit (self=0x7fffe89f7cd8) at /home/emilio/projects/moz/servo/components/constellation/pipeline.rs:326
[...]

PR #26036, opened 2020-03-25
Add ipc router shutdown

gterzian:
Explicitly shutting down IPC router as part of shutdown, for contexts see https://github.com/servo/servo/pull/25685#pullrequestreview-364604068

Note that the call to `shutdown` is idempotent, so it shouldn't matter whether the call in the constellation and the script-thread are actually on the same `ROUTER` in single-process mode...

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2748,9 +2748,9 @@ dependencies = [
 
 [[package]]
 name = "ipc-channel"
-version = "0.14.0"
+version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "55162cbe44dacbc0b4a66d1067c885a9d5975f068a432801bba0493e1ece7a51"
+checksum = "3698b8affd5656032a074a7d40b3c2a29b71971f3e1ff6042b9d40724e20d97c"
 dependencies = [
  "bincode",
  "crossbeam-channel",
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2833,6 +2833,9 @@ where
         debug!("Asking compositor to complete shutdown.");
         self.compositor_proxy
             .send(ToCompositorMsg::ShutdownComplete);
+
+        debug!("Shutting-down IPC router thread in constellation.");
+        ROUTER.shutdown();
     }
 
     fn handle_pipeline_exited(&mut self, pipeline_id: PipelineId) {
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -145,6 +145,7 @@ use script_traits::{ScriptToConstellationChan, TimerSchedulerMsg};
 use script_traits::{TouchEventType, TouchId, UntrustedNodeAddress, WheelDelta};
 use script_traits::{UpdatePipelineIdReason, WebrenderIpcSender, WindowSizeData, WindowSizeType};
 use servo_atoms::Atom;
+use servo_config::opts;
 use servo_url::{ImmutableOrigin, MutableOrigin, ServoUrl};
 use std::borrow::Cow;
 use std::cell::Cell;
@@ -2907,6 +2908,12 @@ impl ScriptThread {
         self.background_hang_monitor
             .as_ref()
             .map(|bhm| bhm.unregister());
+
+        // If we're in multiprocess mode, shut-down the IPC router for this process.
+        if opts::multiprocess() {
+            debug!("Exiting IPC router thread in script thread.");
+            ROUTER.shutdown();
+        }
 
         debug!("Exited script thread.");
     }


Manishearth (approved):
looks good so far, unsure if this is still WIP

jdm:
```
error: mismatched closing delimiter: `)`
    --> /repo/components/script/script_thread.rs:2887:60
     |
2884 |     fn handle_exit_script_thread_msg(&self) {
     |                                             - unclosed delimiter
...
2887 |         debug!"Exiting IPC router thread in script thread.");
     |                                                            ^ mismatched closing delimiter
```

gterzian:
Thanks for reviewing. I've fixed the error and removed the WIP. We could try a run, or if you want to r it already we'll get a run anyway. I think it should be ok but I actually haven't tried to locally or anything. 

I think the constellation and the script-thread would be the two places where you could have different routers if in multi-process, actually it's kind of interesting to realize the script-thread is sort of the "main thread" of a script-process...

CYBAI:
I'm just out of curiosity.

Due to seeing that Greg updated all the `ipc-channel` from `"0.14"` to `"0.14.1"`, I'm wondering if it's worth or better to just run `cargo update -p ipc-channel` to update the `ipc-channel` in lockfile to `0.14.1` so that we don't need to update all the toml files 👀?

gterzian:
```
error: failed to select a version for the requirement `ipc-channel = "^0.14.1"`
  candidate versions found which didn't match: 0.14.0, 0.13.0, 0.12.2, ...
  location searched: crates.io index
```

Does this mean we need to update the crate for `ipc-channel`?

@CYBAI thanks for thinking about this, I personally don't know.

CYBAI:
> Does this mean we need to upload the latest version for `ipc-channel`?

Just checked [ipc-channel on crates.io](https://crates.io/crates/ipc-channel) and I think yes, we will need to publish the latest version 👀
 
> @CYBAI thanks for thinking about this, I personally don't know.

Welcome! and I learned that command from @Eijebong :D

jdm:
I have published 0.14.1.

gterzian:
@Manishearth r?

gterzian:
Ok interesting, so it looks like a layout thread receiving a disconnect error on it's channel from the constellation. 

I think what might have happened previoulsy was that an IPC route would keep it alive, so that the problem was hidden.

I think the actual problem might be that the constellation, when shutting down, doesn't wait for all pipelines to shutdown first. I'll try to add some code to that effect and see if it helps.

gterzian:
Ah no that's not true, pipelines are always exited before `handle_shutdown`...

gterzian:
Ok so the test that is failing was added in the light of https://github.com/servo/servo/issues/11546

But now it's the layout thread that is receiving on a closed receiver from the constellation:

```
  │ called `Result::unwrap()` on an `Err` value: RecvError (thread LayoutThread PipelineId { namespace_id: PipelineNamespaceId(2), index: PipelineIndex(6) }, at components/layout_thread/lib.rs:730)
  │ stack backtrace:
  │              at src/libstd/panicking.rs:513
  │    3: rust_begin_unwind
  │              at src/libstd/panicking.rs:417
  │              at src/libcore/panicking.rs:111
  │    5: core::option::expect_none_failed
  │              at src/libcore/option.rs:1211
  │    6: layout_thread::LayoutThread::start
  │    7: profile_traits::mem::ProfilerChan::run_with_memory_reporting
  │    8: std::sys_common::backtrace::__rust_begin_short_backtrace
  │    9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/45ebd5808afd3df7ba842797c0fcd4447ddf30fb/src/liballoc/boxed.rs:1017
  │   11: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/45ebd5808afd3df7ba842797c0fcd4447ddf30fb/src/liballoc/boxed.rs:1017
  │       std::sys_common::thread::start_thread
  │              at src/libstd/sys_common/thread.rs:13
  │              at src/libstd/sys/unix/thread.rs:80
  │   12: start_thread
  │   13: __clone
  │ [2020-03-30T09:15:03Z ERROR servo] called `Result::unwrap()` on an `Err` value: RecvError
```

cc @asajeffrey

gterzian:
Ah I think I know what's happening: we should only shut-down the router thread when a script-thread shuts-down when running in multi-process mode, since in single-process the router might still be used by another event-loop(in this case it's the layout-thread related to another script-thread that needs to use it).

gterzian:
Ok, looks like it's fixed.

@Manishearth r?

gterzian:
Oops I think I forgot to fmt it, I guess the build will fail anyway then, no need to minus approve it...

jdm:
@bors-servo retry
* #25513

PR #26037, opened 2020-03-25
layout_2020: Use ArcRefCell in the fragment tree

mrobinson:
This will allow mutability which is useful for things like animations.

---

### DIFF
--- a/components/layout_2020/fragments.rs
+++ b/components/layout_2020/fragments.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use crate::cell::ArcRefCell;
 use crate::geom::flow_relative::{Rect, Sides};
 use crate::geom::{PhysicalPoint, PhysicalRect};
 #[cfg(debug_assertions)]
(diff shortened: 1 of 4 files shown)


components/layout_2020/fragments.rs
```diff
@@ -193,7 +228,10 @@ impl AnonymousFragment {
         AnonymousFragment {
             debug_id: DebugId::new(),
             rect,
-            children,
+            children: children
+                .into_iter()
+                .map(|fragment| ArcRefCell::new(fragment))
+                .collect(),
```
SimonSapin: Please add a commit that changes this method to take `Vec<ArcRefCell<Fragment>>` and push `ArcRefCell::new` as early as necessary in various callers to avoid the intermediate `Vec`

And similarly in:
* `BoxFragment::new`
* `flow/root.rs` for `FragmentTreeRoot::children`
* `positioned.rs` for `new_fragment.children.extend(…)`

mrobinson: Thanks for the review.

I ended up using intermediate `Vec`s here in order to avoid a lot of mutable borrows during the creation of each `Fragment`. I wasn't sure if it made sense performance-wise to do some number of mutable borrows per-fragment versus creating a temporary vector for `BoxFragment`s and `AnonymousFragment`s with children. What's your opinion on this?

mrobinson: Okay. After our discussion on Matrix, I stuck with just the original commit. Thanks for the reviews.

PR #26038, opened 2020-03-26
Use rust-mozjs exposed function to generate SourceText

CYBAI:
With the exposed functions from mozjs, we can get rid of the `get_source_text` in `script_module`.

---

- [x] These changes do not require tests because it just tries to use exposed functions from mozjs to create `SourceText`


jdm (approved):
Needs a `mach fmt`.

CYBAI:
Thanks! Totally didn't notice the fmt error! Updated!

PR #26048, opened 2020-03-27
Give a lifetime parameter to LayoutDom

nox:


jdm (approved):
These changes look reasonable. It's not clear to me why we're moving PendingRestyle to the style crate, however.

nox:
`layout_thread` accesses that data without ever naming the type, and it can't name it because it's a public type in script in a private module. I could have added to the layout reexports but that offends me even though I can't explain why it offends me. Would you prefer such a change?

jdm:
I think either the layout exports or script_layout_interface would be the logical places to me. I would expect bits present in style to be used inside style and/or the layout crate.

nox:
> I think either the layout exports or script_layout_interface would be the logical places to me. I would expect bits present in style to be used inside style and/or the layout crate.

It is used in `script_layout_interface` and it needs `style` items, but `script_layout_interface` doesn't depend on `style`, and IMO it should stay that way.

nox:
Disregard my last comment, I think I can move it to `script_layout_interface` itself.

nox:
Ok I moved it to `script_layout_interface`.

@bors-servo r=jdm

nox:
Oops

@bors-servo r=jdm

jdm:
@bors-servo retry
* #24622

PR #26043, opened 2020-03-27
Context menu API

paulrouget:
This adds an API for Servo internals to request a context menu to the embedder, along an implement for the UWP port.

Issue #23899, opened 2019-07-31
Main thread complexity


asajeffrey:
This is intended to be a meta-bug for centralizing the discussions about the complexity of our main thread.

Originally, the main thread ran the embedder (and in particular the window event loop) and webrender, and this wasn't too bad since the only source of blocking on events was the embedder. Webrender blocks on the WebGL thread, but that wasn't too bad when it was off in another thread.

Unfortunately, as Servo has been ported to different devices, we've ended up with more and more cases of things which need to be run on the main thread, including WebXR and WebGL (on devices where GL contexts which share textures need to live on the same thread). This is causing problems.

asajeffrey:
Related issues:

- https://github.com/servo/webxr/issues/34
- https://github.com/servo/servo/issues/15407
- https://github.com/servo/servo/issues/15734
- https://github.com/servo/servo/issues/18799
- https://github.com/servo/servo/issues/22914
- https://github.com/servo/servo/issues/23642
- https://github.com/servo/servo/issues/23865

There are probably others!

asajeffrey:
cc @jdm @gterzian @paulrouget @Manishearth

asajeffrey:
In particular, there is discussion starting at https://github.com/servo/servo/issues/23865#issuecomment-516200904 about whether async/await and futures 0.3 is a good fit for Servo's main thread.

gterzian:
I think the main problem is that 

https://github.com/servo/servo/blob/8ec28978cdc3e0fa66652a7a0a57911ca414254c/ports/libsimpleservo/api/src/lib.rs#L250

is running various sub-loops. 

Since each sub-loops needs to run for different reasons, "waking-up" or "animation mode" is used to achieve a regular heartbeat on each of these sub-loops. 

However, since every-time a sub-loop runs, it doesn't know exactly why, it cannot block on messages and otherwise run in a "robust" manner, instead needing to use `try_recv` or timeouts to try to "catch potentially incoming messages". 

There is also a special "animation mode", in which the main-thread doesn't block on any event as a way to run the various sub-loops without waiting for UI events or calls to `wake`. 

The irony of "animation mode" is that there is no lack of "source of events" for the main-thread, in the form of incoming messages from script. In fact, each frame requires messaging with script over IPC. 

[...]

PR #26051, opened 2020-03-27
Hook up new openxr context menu gesture to embedder context menu machinery

### LINKED ISSUE #25797: Immersive "app" menu on hololens
We should have an "app menu" on hololens, much like the system start menu. This will be where we place trusted UI, for example.

I think a good gesture to have would be a "hold palm facing you" one (suggested by our Microsoft friends).

Initially it would just have an "exit VR" button but we could add more things.

It's easy enough to design the code to detect the gesture and spawn the UI, but I'm not sure how best to write the actual UI code for this. I'm also not sure if Hololens lets you render additional 3D stuff (like UI menus) when in an openxr immersive session.

cc @paulrouget

Manishearth:
Based on https://github.com/servo/servo/pull/26043/

Fixes https://github.com/servo/servo/issues/25797, #26057

https://github.com/servo/webxr/pull/144 needs to land first

Currently when exited the Servo window is blurred, apparently we need to call `Window.Activate` on it.

r? @jdm

cc @paulrouget

### DIFF
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -20,7 +20,9 @@ use servo::compositing::windowing::{
     WindowMethods,
 };
 use servo::embedder_traits::resources::{self, Resource, ResourceReaderMethods};
-use servo::embedder_traits::{EmbedderMsg, MediaSessionEvent, PromptDefinition, PromptOrigin};
+use servo::embedder_traits::{
+    EmbedderMsg, EmbedderProxy, MediaSessionEvent, PromptDefinition, PromptOrigin,
+};
 use servo::euclid::{Point2D, Rect, Scale, Size2D, Vector2D};
 use servo::keyboard_types::{Key, KeyState, KeyboardEvent};
 use servo::msg::constellation_msg::TraversalDirection;
@@ -764,7 +766,10 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
         registry: &mut webxr::MainThreadRegistry,
         executor: WebGlExecutor,
         surface_providers: SurfaceProviders,
+        embedder_proxy: EmbedderProxy,
     ) {
+        use ipc_channel::ipc::{self, IpcReceiver};
+        use webxr::openxr;
         debug!("EmbedderMethods::register_xr");
         assert!(
             self.xr_discovery.is_none(),
@@ -772,12 +777,50 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
         );
 
         struct ProviderRegistration(SurfaceProviders);
-        impl webxr::openxr::SurfaceProviderRegistration for ProviderRegistration {
+        impl openxr::SurfaceProviderRegistration for ProviderRegistration {
             fn register(&self, id: webxr_api::SessionId, provider: servo::canvas::SurfaceProvider) {
                 self.0.lock().unwrap().insert(id, provider);
             }
-            fn clone(&self) -> Box<dyn webxr::openxr::SurfaceProviderRegistration> {
+            fn clone(&self) -> Box<dyn openxr::SurfaceProviderRegistration> {
                 Box::new(ProviderRegistration(self.0.clone()))
+            }
+        }
+
+        #[derive(Clone)]
+        struct ContextMenuCallback(EmbedderProxy);
+
+        struct ContextMenuFuture(IpcReceiver<ContextMenuResult>);
+
+        impl openxr::ContextMenuProvider for ContextMenuCallback {
+            fn open_context_menu(&self) -> Box<dyn openxr::ContextMenuFuture> {
+                let (sender, receiver) = ipc::channel().unwrap();
+                self.0.send((
+                    None,
+                    EmbedderMsg::ShowContextMenu(
+                        sender,
+                        Some("Would you like to exit the XR session?".into()),
+                        vec!["Exit".into()],
+                    ),
+                ));
+
+                Box::new(ContextMenuFuture(receiver))
+            }
+            fn clone_object(&self) -> Box<dyn openxr::ContextMenuProvider> {
+                Box::new(self.clone())
+            }
+        }
+
+        impl openxr::ContextMenuFuture for ContextMenuFuture {
+            fn poll(&self) -> openxr::ContextMenuResult {
+                if let Ok(result) = self.0.try_recv() {
+                    if let ContextMenuResult::Selected(0) = result {
+                        openxr::ContextMenuResult::ExitSession
+                    } else {
+                        openxr::ContextMenuResult::Dismissed
+                    }
+                } else {
+                    openxr::ContextMenuResult::Pending
+                }
             }
         }
 
@@ -794,6 +837,7 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
         let discovery = webxr::openxr::OpenXrDiscovery::new(
             Box::new(GlThread(executor)),
             Box::new(ProviderRegistration(surface_providers)),
+            Box::new(ContextMenuCallback(embedder_proxy)),
         );
         registry.register(discovery);
     }
@@ -804,6 +848,7 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
         registry: &mut webxr::MainThreadRegistry,
         _executor: WebGlExecutor,
         _surface_provider_registration: SurfaceProviders,
+        _embedder_proxy: EmbedderProxy,
     ) {
         debug!("EmbedderMethods::register_xr");
         if let Some(discovery) = self.xr_discovery.take() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6680,7 +6680,7 @@ dependencies = [
 [[package]]
 name = "webxr"
 version = "0.0.1"
-source = "git+https://github.com/servo/webxr#72b30f2b70ff7b40dc0ffdeefce1116b24041f3d"
+source = "git+https://github.com/servo/webxr#412f385483ad502320f94d98fe5beabf53d14de7"
 dependencies = [
  "bindgen",
  "crossbeam-channel",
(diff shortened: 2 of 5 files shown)


gterzian:
If I may make a general comment on the design: while this undoubtedly "works", could it perhaps contribute to what was already noted to be a somewhat [complex situation](https://github.com/servo/servo/issues/23899) on the main-thread? 

Even if webxr is running off the main thread, could there be a simpler way to express this new API?

I had a look at this PR, and the related one in WebXR, and if I understand it correctly this is essentially about calling `quit` on the XR device if the user chooses to? And the other side seems to be about enabling the XR device to use the embedder API to show a context menu. 

So in an ideal world, would it not be great if OpenXR could just call `embedder.show_context_menu(items)`, and inside `on_context_menu_closed`, the embedder could just call `device.quit`(if that was the chosen action? Or at least something like `device.handle_context_menu_choice(choice)`)? 

Then those API could be implemented with messaging for off-the-main-thread Xr, and on the main-thread be simply synchronous call(I think the menu is async anyway)? 

What I find hard to reason about the current design is how polling the result of the menu choice is tied to waiting on a frame inside Xr, and how if it's running on the main-thread the messaging is actually to the same event-loop(with the polling happening in a kind of sub-event-loop).

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -164,6 +169,7 @@ pub struct ServoGlue {
     browsers: Vec<BrowserId>,
     events: Vec<WindowEvent>,
     current_url: Option<ServoUrl>,
+    context_menu_sender: Option<IpcSender<ContextMenuResult>>,
```
gterzian: I *think* this could be a threaded channel. The `EmbedderProxy` itself also uses a threaded channel internally.

paulrouget:
@gterzian The API change has landed as part of #26043 

>  what I'm proposing would require a larger refactoring, so it's perhaps more suited for a follow-up.

At the moment, we're just following the current embedder API structure. But I'd be interested to know what type of changes you'd like to see in a future embedder API. Feel free to file another issue to continue the conversation.

gterzian:
> The API change has landed as part of #26043

Yes, and I'm mainly referring to the additional integration taking place here, where the result of menu choice is handled by polling the `ContextMenuFuture`, which in practice will be done inside [`wait_for_animation_frame`](https://github.com/servo/webxr/pull/144/files#diff-f3cfd7c22eb8b9fb4aeefdabdb68f7e4R709).

Manishearth:
> So in an ideal world, would it not be great if OpenXR could just call `embedder.show_context_menu(items)`, and inside `on_context_menu_closed`, the embedder could just call `device.quit`(if that was the chosen action? Or at least something like `device.handle_context_menu_choice(choice)`)?

Yes, but I don't want the webxr crate to know anything about the embedder.

> What I find somewhat harder to reason about the current design is how polling the result of the menu choice is tied to waiting on a frame inside Xr, and how if it's running on the main-thread the messaging is actually to the same event-loop(with the polling happening in a kind of sub-event-loop).

The openxr device is not running on the main thread. It's tied to waiting on a frame just because that's where all the other events are being handled, I needed to pick _some_ point on the openxr thread to poll, so I picked this one, since all the other events are being handled there, and in my eye this is basically just another device event that we're "helping".

---------------

I thought of the other way of doing this: Instead of polling, we can give the device access to the `Quitter` object (which we're not using here). However, this doesn't let the device know when the menu was closed, since it needs to start listening to the menu again, and start feeding input data back to the session. I could add more functionality to the quitter and make it into a remote-like object, but that seemed equally convoluted.

gterzian:
> Yes, but I don't want the webxr crate to know anything about the embedder.

I agree, I think the current structure wrapping the `EmbedderProxy` looks like the right one.

> The openxr device is not running on the main thread. 

Ok, thanks for pointing this out, just had a better look at the code and saw the use of `xr.spawn`...

> I needed to pick some point on the openxr thread to poll

Ok I see. 

Just a question, I now see there is actually a `SessionMsg::Quit`, so why not have the embedder send that message to be handled by open xr on it's own thread? Is it because it's hard to tie the menu call to a particular session? 

Could you not add "something wrapping a sender of a `SessionMsg::Quit`" to the `ShowContextMenu` message, so that the embedder could just call `quit` on it and it would internally send the session message? 

That seems like a potential way to separate this from the animation frame?

gterzian:
Ok so I've taken a better look at the code, and I think this is is a concrete proposal that would address my concerns with regards to tying the menu to the animation frame flow:

1. `OpenXrDiscovery` owns a `EmbedderProxy`.
2. `SessionBuilder.spawn` takes a new first argument, an `EmbedderProxy`. 
3. `DeviceAPI` has a new `set_context_menu_provider(Box< dyn ContextMenuProvider>)` method.
4. Inside `SessionThread::new`, use the `EmbedderProxy` and the `Sender<SessionMsg>` to create a  `ContextMenuProvider`, and pass it to the device by calling `set_context_menu_provider`. 

`ContextMenuProvider` looks like:

```rust
pub trait ContextMenuProvider: Send {
    /// Open a context menu, clones self and sends a message to the embedder.
    fn open_context_menu(&self);
    /// Exit the session.
    fn handle_selection(&self, result: usize);
}
```

Then, when the device calls `open_context_menu`, it doesn't need to return anything. Instead, it internally sends a `ShowContextMenu(ContextMenuProvider, Vec<String>)`, with the provider actually being a clone of the provider.

Then, when the embedder gets the result via `on_context_menu_closed`, it takes the provider and in the case of the result being `ContextMenuResult::Selected(selection)` it calls `provider.exit_session(selection)`. This then internally sends a `SessionMsg::Quit`, if `selection` is 0.

Then, when the session thread receives the `SessionMsg::Quit` message, it quits the device. 

So, not saying this is super simple either, however it does address my concern of tying the messaging into the frame loop.

gterzian:
> Yes, but I don't want the webxr crate to know anything about the embedder.

Ah yes, so then my suggestion above would have to be updated to not passing the `EmbedderProxy`, but rather a trait like:

```rust
pub trait ContextMenuProvider: Send {
    /// Called from the session thread.
    fn open_context_menu(&self, Box<dyn ContextMenuCallback>);
    /// Called from the main thread once the result have been received.
    /// Maybe separate the two in two different traits?
    fn handle_selection(&self, result: usize);
}

pub trait ContextMenuCallback {
    /// Called by the embedder,
    /// This would internally send the `SessionMsg::Quit` based on the result passed.
    fn handle_result(&self, result: ContextMenuResult)
}
```

Manishearth:
> why not have the embedder send that message to be handled by open xr on it's own thread?

the embedder doesn't have an easy way to get a handle to the openxr device's sender

> Just a question, I now see there is actually a `SessionMsg::Quit`, so why not have the embedder send that message to be handled by open xr on it's own thread? Is it because it's hard to tie the menu call to a particular session?

Yes.

We could pass the quitter down to the embedder via the ShowContextMenu trait, and add another function to it that allows for cancelling context menus. I didn't want to make the core webxr API dependent on openxr-specific things though.

Like, I considered exactly this design, especially since the quitter exists already; 

-----

Basically, I've considered this design, but I'm unconvinced that it is better -- both are kinda gnarly. I guess having one trait is better than having two. Thoughts?

jdm (approved):

gterzian:
@Manishearth Thanks for elaborating, now that I realize the XR frame loop really does need to keep polling frames while the session is running, I think my concern about polling for the result of the menu as part of the frame loop seems less relevant. I'd say this PR can also FIX #26057

Looks good to me, perhaps you could still replace the IPC channel with a threaded one if we're  indeed not crossing a process boundary?

Manishearth:
@gterzian we're not crossing a process boundary _now_, unsure if this might ever change.

Still, I'm open to the Quitter-based design as well, do you still prefer it? It might be cleaner.

gterzian:
> we're not crossing a process boundary now, unsure if this might ever change.

Ok :)

> Still, I'm open to the Quitter-based design as well, do you still prefer it? It might be cleaner.

My main concern was tying the polling of the menu result into the frame loop, since that concern doesn't apply to the Xr frame loop, I'm much more ambivalent.

Manishearth:
Cool, then in the interest of laziness I'll keep it the way it is.

jdm:
```
error[E0050]: method `register_webxr` has 4 parameters but the declaration in trait `compositing::windowing::EmbedderMethods::register_webxr` has 5
  --> ports/glutin/embedder.rs:94:9
   |
94 | /         &mut self,
95 | |         xr: &mut webxr::MainThreadRegistry,
96 | |         _executor: WebGlExecutor,
97 | |         _surface_provider_registration: SurfaceProviders
   | |________________________________________________________^ expected 5 parameters, found 4
   |
   = note: `register_webxr` from trait: `fn(&mut Self, &mut webxr_api::registry::MainThreadRegistry<surfman_chains::SwapChains<webxr_api::SwapChainId>>, crossbeam_channel::channel::Sender<std::boxed::Box<(dyn std::ops::FnOnce() + std::marker::Send + 'static)>>, std::sync::Arc<std::sync::Mutex<std::collections::HashMap<webxr_api::session::SessionId, std::boxed::Box<(dyn surfman_chains::SurfaceProvider + std::marker::Send + 'static)>>>>, embedder_traits::EmbedderProxy)`
```

CYBAI:
👀

```
error: expected one of `!`, `(`, `)`, `+`, `,`, `::`, or `<`, found `_embedder_proxy`
  --> ports/glutin/embedder.rs:98:9
   |
97 |         _surface_provider_registration: SurfaceProviders
   |                                                         -
   |                                                         |
   |                                                         expected one of 7 possible tokens
   |                                                         help: missing `,`
98 |         _embedder_proxy: EmbedderProxy,
   |         ^^^^^^^^^^^^^^^ unexpected token
error: could not compile `servo`.
```

jdm:
This will also need to be rebased over https://github.com/servo/servo/pull/26067 now.

jdm:
```
error: unused imports: `IpcReceiver`, `self`
  --> ports\libsimpleservo\api\src\lib.rs:16:24
   |
16 | use ipc_channel::ipc::{self, IpcReceiver, IpcSender};
   |                        ^^^^  ^^^^^^^^^^^
   |
```

jdm:
@bors-servo retry
* #23290

Manishearth:
@bors-servo retry
 - whitespace and linebreak failures

PR #26054, opened 2020-03-28
Fix tojson unconditionally serializes

### LINKED ISSUE #25281: ToJSON unconditionally serializes all conditionally-exposed interface members
This test will crash after this [PR](https://github.com/servo/servo/pull/25205). 

▶ CRASH [expected OK] /hr-time/idlharness.any.worker.html
  │
  │
  │
  │
  │ internal error: entered unreachable code: Are we trying to expose Performance.timing in workers? (thread WebWorker for http://web-platform.test:8000/hr-time/idlharness.any.worker.js, at components/script/dom/performan
ce.rs:394)
  │ stack backtrace:
  │    0: backtrace::backtrace::trace_unsynchronized
  │    1: <servo::backtrace::Print as core::fmt::Debug>::fmt
  │    2: core::fmt::write
  │    3: <std::io::stdio::Stdout as std::io::Write>::write_fmt
  │    4: std::io::stdio::_print
  │    5: servo::backtrace::print
  │    6: servo::main::{{closure}}
  │    7: std::panicking::rust_panic_with_hook
  │    8: rust_begin_unwind
  │    9: std::panicking::begin_panic_fmt
  │   10: <script::dom::performance::Performance as script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::PerformanceMethods>::Timing
  │   11: script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::get_timing::{{closure}}
  │   12: core::ops::function::FnOnce::call_once
  │   13: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  │   14: std::panicking::try::do_call
  │   15: __rust_maybe_catch_panic
  │   16: std::panicking::try
  │   17: std::panic::catch_unwind
  │   18: mozjs::panic::wrap_panic
  │   19: script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::get_timing
  │   20: script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::CollectJSONAttributes
  │   21: script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::toJSON::{{closure}}
  │   22: core::ops::function::FnOnce::call_once
  │   23: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  │   24: std::panicking::try::do_call
  │   25: __rust_maybe_catch_panic
  │   26: std::panicking::try
  │   27: std::panic::catch_unwind
  │   28: mozjs::panic::wrap_panic
  │   29: script::dom::bindings::codegen::Bindings::PerformanceBinding::PerformanceBinding::toJSON
  │   30: CallJitMethodOp
  │   31: script::dom::bindings::utils::generic_call
  │   32: script::dom::bindings::utils::generic_method
  │   33: _ZN2js23InternalCallOrConstructEP9JSContextRKN2JS8CallArgsENS_14MaybeConstructE
  │   34: _ZL9InterpretP9JSContextRN2js8RunStateE
  │   35: _ZN2js9RunScriptEP9JSContextRNS_8RunStateE
  │   36: _ZN2js23InternalCallOrConstructEP9JSContextRKN2JS8CallArgsENS_14MaybeConstructE
  │   37: _ZN2js4CallEP9JSContextN2JS6HandleINS2_5ValueEEES5_RKNS_13AnyInvokeArgsENS2_13MutableHandleIS4_EE
  │   38: _ZN2js3jit14InvokeFunctionEP9JSContextN2JS6HandleIP8JSObjectEEbbjPNS3_5ValueENS3_13MutableHandleIS8_EE
  │   39: _ZN2js3jit25InvokeFromInterpreterStubEP9JSContextPNS0_30InterpreterStubExitFrameLayoutE
  │
  │ [2019-12-13T14:29:19Z ERROR servo] internal error: entered unreachable code: Are we trying to expose Performance.timing in workers?

shnmorimoto:
---

- [x] These changes fix #25281

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2977,10 +2977,14 @@ class CGCollectJSONAttributesMethod(CGAbstractMethod):
                 ret += fill(
                     """
                     let conditions = ${conditions};
-                    if !conditions.iter().any(|c| c.is_satisfied(SafeJSContext::from_ptr(cx), HandleObject::from_raw(obj), HandleObject::from_raw(obj))) {
+                    if !conditions.iter().any(|c|
+                         c.is_satisfied(
+                           SafeJSContext::from_ptr(cx),
+                           HandleObject::from_raw(obj),
+                           HandleObject::from_raw(obj))) {
                       return false;
                     }
-                    rooted!(in(cx) let mut temp = UndefinedValue()); 
+                    rooted!(in(cx) let mut temp = UndefinedValue());
                     if !get_${name}(cx, obj, this, JSJitGetterCallArgs { _base: temp.handle_mut().into() }) {
                       return false;
                     }
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2976,22 +2976,24 @@ class CGCollectJSONAttributesMethod(CGAbstractMethod):
                 ret_conditions = 'vec![' + ",".join(conditions) + "]"
                 ret += fill(
                     """
+                    let incumbent_global = GlobalScope::incumbent().expect("no incumbent global");
+                    let global = incumbent_global.reflector().get_jsobject();
                     let conditions = ${conditions};
-                    if !conditions.iter().any(|c|
+                    let is_satisfied = conditions.iter().any(|c|
                          c.is_satisfied(
                            SafeJSContext::from_ptr(cx),
                            HandleObject::from_raw(obj),
-                           HandleObject::from_raw(obj))) {
-                      return false;
-                    }
-                    rooted!(in(cx) let mut temp = UndefinedValue());
-                    if !get_${name}(cx, obj, this, JSJitGetterCallArgs { _base: temp.handle_mut().into() }) {
-                      return false;
-                    }
-                    if !JS_DefineProperty(cx, result.handle().into(),
-                                          ${nameAsArray} as *const u8 as *const libc::c_char,
-                                          temp.handle(), JSPROP_ENUMERATE as u32) {
-                      return false;
+                           global));
+                    if is_satisfied {
+                      rooted!(in(cx) let mut temp = UndefinedValue());
+                      if !get_${name}(cx, obj, this, JSJitGetterCallArgs { _base: temp.handle_mut().into() }) {
+                        return false;
+                      }
+                      if !JS_DefineProperty(cx, result.handle().into(),
+                                            ${nameAsArray} as *const u8 as *const libc::c_char,
+                                            temp.handle(), JSPROP_ENUMERATE as u32) {
+                        return false;
+                      }
                     }
                     """,
                     name=name, nameAsArray=str_to_const_array(name), conditions=ret_conditions)
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2966,18 +2966,16 @@ class CGCollectJSONAttributesMethod(CGAbstractMethod):
         self.toJSONMethod = toJSONMethod
 
     def definition_body(self):
-        ret = ''
+        ret = """let incumbent_global = GlobalScope::incumbent().expect("no incumbent global");
+let global = incumbent_global.reflector().get_jsobject();\n"""
         interface = self.descriptor.interface
-
         for m in interface.members:
             if m.isAttr() and not m.isStatic() and m.type.isJSONType():
                 name = m.identifier.name
                 conditions = MemberCondition(None, None, m.exposureSet)
-                ret_conditions = 'vec![' + ",".join(conditions) + "]"
+                ret_conditions = '&[' + ", ".join(conditions) + "]"
                 ret += fill(
                     """
-                    let incumbent_global = GlobalScope::incumbent().expect("no incumbent global");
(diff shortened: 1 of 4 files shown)


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2968,11 +2968,22 @@ def __init__(self, descriptor, toJSONMethod):
...
+                ret_conditions = 'vec![' + ",".join(conditions) + "]"
                 ret += fill(
                     """
+                    let conditions = ${conditions};
+                    if !conditions.iter().any(|c|
+                         c.is_satisfied(
+                           SafeJSContext::from_ptr(cx),
+                           HandleObject::from_raw(obj),
+                           HandleObject::from_raw(obj))) {
+                      return false;
```
jdm: I believe the timeouts in the tests are caused by returning here if there condition is not satisfied, rather than only executing the following code if the condition is satisfied.

jdm (changes requested):

jdm:
Example timeout:
```
  ▶ Unexpected subtest result in /css/geometry/interfaces.html:
  │ TIMEOUT [expected PASS] Test driver
  └   → Test timed out
  ▶ Unexpected subtest result in /css/geometry/interfaces.html:
  │ TIMEOUT [expected PASS] DOMPointReadOnly interface: default toJSON operation on new DOMPointReadOnly()
  └   → Test timed out
```

components/script/dom/bindings/codegen/CodegenRust.py
shnmorimoto: Thanks for review and checking. Do you mean that we should skip like below instead of returning false?

```rust
if flag = conditions.iter().any(|c|
         c.is_satisfied(
           SafeJSContext::from_ptr(cx),
           HandleObject::from_raw(obj),
           HandleObject::from_raw(obj)));
if flag {
    rooted!(in(cx) let mut temp = UndefinedValue());
    if !get_timing(cx, obj, this, JSJitGetterCallArgs { _base: temp.handle_mut().into() }) {
    return false;
    }
    if !JS_DefineProperty(cx, result.handle().into(),
                          b"timing\0" as *const u8 as *const libc::c_char,
                        temp.handle(), JSPROP_ENUMERATE as u32) {
    return false;
    }
}
```

I've tried it. But I still got timeout.
I'll continue to look into it.

jdm: Ah, another problem is that we should be passing a global object as the third argument of is_satisfied based on https://doc.servo.org/script/dom/bindings/guard/enum.Condition.html#method.is_satisfied. We should be able to use something like:
```rust
let incumbent_global = GlobalScope::incumbent().expect("no incumbent global");
let global = incumbent_global.reflector().get_jsobject();
```
And then you could pass `global` as the third argument.

shnmorimoto: Oh, sorry. I got it.
I fixed it and confirmed if the timeout isn't occurred.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2968,22 +2968,35 @@ def __init__(self, descriptor, toJSONMethod):
...
-                    rooted!(in(cx) let mut temp = UndefinedValue());
-                    if !get_${name}(cx, obj, this, JSJitGetterCallArgs { _base: temp.handle_mut().into() }) {
-                      return false;
-                    }
-                    if !JS_DefineProperty(cx, result.handle().into(),
-                                          ${nameAsArray} as *const u8 as *const libc::c_char,
-                                          temp.handle(), JSPROP_ENUMERATE as u32) {
-                      return false;
+                    let incumbent_global = GlobalScope::incumbent().expect("no incumbent global");
+                    let global = incumbent_global.reflector().get_jsobject();
```
jdm: We can move these to the initial definition of `ret` so that we only get the global object once at the start of the method.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2968,22 +2968,35 @@ def __init__(self, descriptor, toJSONMethod):
     def definition_body(self):
         ret = ''
         interface = self.descriptor.interface
+
         for m in interface.members:
             if m.isAttr() and not m.isStatic() and m.type.isJSONType():
                 name = m.identifier.name
+                conditions = MemberCondition(None, None, m.exposureSet)
+                ret_conditions = 'vec![' + ",".join(conditions) + "]"
```
jdm: Let's use `&[` instead of `vec![` to avoid allocating a bunch of memory needlessly. Also use `", "` when joining to make the resulting code a bit easier to read.

jdm (changes requested):
Great work!

components/script/dom/bindings/codegen/CodegenRust.py
shnmorimoto: Thanks for the review. fixed 74995a5287dc0af8afe62e9b537db5141b854765

components/script/dom/bindings/codegen/CodegenRust.py
shnmorimoto: thanks! fixed 74995a5287dc0af8afe62e9b537db5141b854765

shnmorimoto:
@jdm @CYBAI Thanks for review and kind help.

PR #26060, opened 2020-03-30
implementing doge api

sresthas:
@jdm please review.
I have tested the changes and attached the screenshot below.
Screenshot from 2020-03-30 09-37-11

### DIFF
--- a/components/script/dom/doge.rs
+++ b/components/script/dom/doge.rs
@@ -1,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use crate::dom::bindings::cell::DomRefCell;
+use crate::dom::bindings::codegen::Bindings::DogeBinding::{DogeMethods, DogeInit};
+use crate::dom::bindings::error::{Error, Fallible};
+use crate::dom::globalscope::GlobalScope;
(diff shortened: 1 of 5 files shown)


components/script/dom/doge.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+        
+        Ok(doge)
+    }
+}
+
+impl DogeMethods for Doge {
+    fn Append(&self, word: ByteString) -> () {
+        *&self.such_list.borrow_mut().push(word);
```
jdm: The `*&` doesn't actually do anything useful here.

components/script/dom/doge.rs
```diff
@@ -0,0 +1,77 @@
...
+            return Ok(list[random_index].clone());
+        }
+    }
+
+    fn Remove(&self, word: ByteString) -> Fallible<()> {
+        let mut list = self.such_list.borrow_mut();
+        let mut present = 0;
+        for i in 0..list.len() {
+                if word == list[i] {
+                list.remove(i);
```
jdm: This method will panic with the following testcase, because we loop from 0 to the original list length but remove some elements while iterating:
```js
let d = new Doge(["a", "b", "c"]);
d.remove("a");
```
A safer way to write this code is to rely on the [Vec::retain](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain) API instead.

jdm:
I'm going to go ahead and close this PR, since this is only a demonstration API and not one we actually want to expose to real web content. Thanks for sharing your work, though!

PR #26065, opened 2020-03-31
Don't send empty canvases to WebRender

dralley:
If any dimension of a canvas is 0, don't try to display it as it causes
problems inside webrender.

Minimal test case available here: https://github.com/servo/servo/issues/21411#issuecomment-605226547

---

### DIFF
--- a/components/layout/display_list/builder.rs
+++ b/components/layout/display_list/builder.rs
@@ -1875,6 +1875,12 @@ impl Fragment {
                 }
             },
             SpecificFragmentInfo::Canvas(ref canvas_fragment_info) => {
+                if canvas_fragment_info.dom_width == Au(0) ||
+                    canvas_fragment_info.dom_height == Au(0)
+                {
+                    return;
+                }
+
                 let image_key = match canvas_fragment_info.source {
                     CanvasFragmentSource::WebGL(image_key) => image_key,
                     CanvasFragmentSource::Image(ref ipc_renderer) => match *ipc_renderer {
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -87,6 +87,7 @@ impl Fragment {
                     .rect
                     .to_physical(i.style.writing_mode, containing_block)
                     .translate(containing_block.origin.to_vector());
+
                 let common = builder.common_properties(rect.clone().to_webrender());
                 builder.wr.push_image(
                     &common,
--- a/components/layout_2020/replaced.rs
+++ b/components/layout_2020/replaced.rs
@@ -102,6 +102,7 @@ impl ReplacedContent {
 
         let width = (intrinsic_size_in_dots.width as CSSFloat) / dppx;
         let height = (intrinsic_size_in_dots.height as CSSFloat) / dppx;
+
         return Some(Self {
             kind,
             intrinsic: IntrinsicSizes {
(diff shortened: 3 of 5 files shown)


dralley:
Cannot run test-tidy locally because Fedora 31 is Python 3 only

cbrewster:
Since we have a minimal test case, it would be nice to add a test that makes sure we don't regress on this panic.

Since this will be a servo-specific test, I would put it here: `tests/wpt/mozilla/tests/mozilla/`
Here's a similar test that ensures we don't crash under a specific scenario: https://github.com/servo/servo/blob/029049b48630f426ac5face4d9a9c0cd5195c66c/tests/wpt/mozilla/tests/mozilla/link_rel_crash.html#L1-L15

dralley:
There's no issue specifically relating to zero-sized-canvas panic -- should I create one specifically to reference or just reference the motionmark issue?

cbrewster:
Referencing the motionmark issue would be fine since that's where most the discussion is at regarding this fix.

cbrewster:
Can you run `./mach update-manifest` and push, that will update the test manifest file.

Once that is done, this looks good to me. Thanks again!

gterzian:
I think you might want to apply the fix also to `layout_2020`.

cbrewster:
@gterzian ah good catch!
@bors-servo r-

dralley:
No idea if that is the correct fix for layout 2020, but it seems like a decent guess.

cbrewster:
Maybe it would make more sense here? https://github.com/servo/servo/blob/75ca3d8198520c1be5fa8218f8b238c7d319d78b/components/layout_2020/replaced.rs#L203-L230

It might be good to check with someone familiar with layout 2020 canvas.

cc @SimonSapin @ferjm

components/layout_2020/dom_traversal.rs
```diff
@@ -411,6 +411,9 @@ where
     fn as_canvas(self) -> Option<(CanvasInfo, PhysicalSize<f64>)> {
         let node = self.to_threadsafe();
         let canvas_data = node.canvas_data()?;
+        if canvas_data.width == 0 || canvas_data.height == 0 {
+            return None;
```
SimonSapin, line 415: This doesn’t look right: it’ll generate a non-replaced box (as for `<div></div>`), but we want to generate a replaced box, which might have different layout behavior (even if zero-size). Somewhere under `layout_2020/display_list/` would be the right place for this change.

dralley:
Alright, I've tested this patch for layout 2020 and it seems to work... still not 100% sure this is the right place but I didn't see anywhere better.. comparing equality of floats makes me cringe a bit but there is no other size available in this location.

SimonSapin:
Sorry, but this still doesn’t look right. Returning `None` from the constructor of `ReplacedContent` still cause the box to be non-replaced, instead of a zero-size replaced box. Somewhere under `layout_2020/display_list/` would be the right place for this change.

SimonSapin:
> Would it be better to just merge the fix for layout 2013 and file an issue for layout 2020?

That’s fine with me too.

----

This code generates a fragment for the replaced box for a `<canvas>` element:

https://github.com/servo/servo/blob/e66ab111a6fb5ea9fc7684a0ebeddbe6aa2134f1/components/layout_2020/replaced.rs#L203-L230

It generates a `Fragment::Image`, because at that point the canvas has already been rendered to a pixel buffer within WebRender). This image is then added to the display list here:

https://github.com/servo/servo/blob/e66ab111a6fb5ea9fc7684a0ebeddbe6aa2134f1/components/layout_2020/display_list/mod.rs#L84-L99

In that second location, it should be possible to check for `i.rect.size == Vec2::zero()` (with `use crate::geom::Vec2;` at the top of the file) to skip generating a display item for zero-size images (including rendered `<canvas>`). But if the problem this is attempting to fix is with rendering canvas to pixels in the first place, then in the first location we could check for `(self.intrinsic.width, self.intrinsic.height) == (Some(Length::zero()), Some(Length::zero()))` (with `use style::Zero;` at the top of the file) and return an empty `Vec` of zero fragments.

dralley:
@SimonSapin Thanks!  I think the more accurate description of this issue is probably the latter one, that Webrender shouldn't be rendering 0-size canvases.  The latter patch you suggested works (slightly modified), the others did not.

SimonSapin:
Looks good, thanks!

@bors-servo r=cbrewster,SimonSapin

PR #26070, opened 2020-03-31
Make LayoutNodeHelpers::text_content return a cow

nox:
What does it mean? It means that layout can process `Text` nodes' contents by just borrowing them, potentially saving us from a lot of unnecessary tempoorary allocations, which is nice.

### DIFF
--- a/components/script/dom/attr.rs
+++ b/components/script/dom/attr.rs
@@ -235,9 +235,9 @@ impl Attr {
 #[allow(unsafe_code)]
 pub trait AttrHelpersForLayout<'dom> {
     fn value(self) -> &'dom AttrValue;
-    fn value_ref_forever(self) -> &'dom str;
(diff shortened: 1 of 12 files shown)


components/script/dom/attr.rs
```diff
@@ -248,12 +248,12 @@ impl<'dom> AttrHelpersForLayout<'dom> for LayoutDom<'dom, Attr> {
...
     #[inline]
-    fn value_ref_forever(self) -> &'dom str {
+    fn as_str(self) -> &'dom str {
         &**self.value()
     }
 
     #[inline]
-    fn value_tokens(self) -> Option<&'dom [Atom]> {
+    fn as_tokens(self) -> Option<&'dom [Atom]> {
         // This transmute is used to cheat the lifetime restriction.
```
SimonSapin: Remove this obsolete comment?

SimonSapin (approved):
r+ with one nit

PR #26074, opened 2020-03-31
Avoid infinitely looping CSS transitions.

### LINKED ISSUE #20379: The `Idle` state is not always reached at the end of css transitions
When a css transition is ongoing, the embedder is notified by the https://doc.servo.org/compositing/windowing/trait.WindowMethods.html#method.set_animation_state method with state value of `Animating`, and then back with the `Idle` value once the animations are done.

I have a couple of cases where we don't go back to the `Idle` state after a transition, and instead keep hammering the embedder with `Animating` notifications. This may not be too much noticeable on desktop, but it can be pretty bad on some mobile devices (and it's wrong anyway).

Just filing for now until I manage to build a simple enough test case to reproduce.

jdm:
This change addresses the long-standing issue of CSS transitions not ending appropriately. It does not fundamentally change the way we process transitions/animations.

---
- [x] These changes fix #20379

### DIFF
--- a/components/layout/animation.rs
+++ b/components/layout/animation.rs
@@ -110,6 +110,7 @@ pub fn update_animation_state<E>(
                     .unwrap();
             }
 
+            debug!("expiring animation for {:?}", running_animation);
             expired_animations
                 .entry(*key)
                 .or_insert_with(Vec::new)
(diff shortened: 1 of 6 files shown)


jdm:
I have verified that all of the test cases from #20379 continue to transition appropriately, and the transition present in the VR button in babylon.js demos works fine with these changes.

@bors-servo try=wpt

jdm:
@bors-servo retry
* #24622 #23290

Issue #26019, opened 2020-03-23
babylon drumming demo triggers a lot of layout


jdm:
I've identified three culprits so far:
* #25893
* repeatedly checking canvasElement.clientWidth every frame in https://github.com/BabylonJS/Babylon.js/blob/18f4147d12fd3174d99cd27c3f6e8a1001ddf9cd/Playground/js/frame.js#L112
* the constellation sends notification of transition events that are processed by [this code](https://github.com/servo/servo/blob/master/components/script/script_thread.rs#L2952-L2959) - not only does it mark the node dirty, but it also performs a layout query to obtain the style information via has_css_layout_box

jdm:
My suspicion is the CSS transition events are actually a symptom of our terrible transition/animation code interacting with the Babylon.js loading animation. The easiest way to verify this would be to prevent the loading animation from running at all and see if the transition events disappear.

jdm:
I tracked down the source of the animation spam - mousing over the button to enter VR triggers an animation that makes the icon slightly grow, but that animation stops as expected. However, mousing away from the icon or clicking it leads to an endless spam of EaseOut animation processing as the animation is never fully considered ended.

jdm:
I suspect this is another instance of #20379.

jdm:
It comes from https://github.com/BabylonJS/Babylon.js/blob/63bc8d7adfcaaa179c42f0907a0a3c59fef0a57e/src/XR/webXREnterExitUI.ts#L92-L93:
`transition: transform 0.125s ease-out`
`.babylonVRicon:hover { transform: scale(1.05) }`

jdm:
With #20379 fixed, the remaining layout appears to come from calling WebGLRenderingContext.clear, which marks the canvas as dirty: https://github.com/servo/servo/blob/cf725fc3f59bc1b94e3b35ab4cfe60b3d41559cc/components/script/dom/webglrenderingcontext.rs#L2085

jdm:
When I fix https://github.com/servo/servo/issues/26019#issuecomment-606765387 and make the canvas.clientWidth calls use a cached layout value instead of triggering a layout query every frame, the resulting profiles show the only threads doing meaningful work are the script thread, the webgl thread, and the openxr thread. Success!

jdm:
All known sources of layout in this babylon demo are gone.

PR #26076, opened 2020-03-31
Reduce unnecessary layout queries in babylon.js content

jdm:
Every frame, Babylon.js compares the width property of the webgl canvas element to its clientWidth. This incurs two layout operations every frame - one to get the dimensions of the element, and at the end of the frame to construct a display list and send that to webrender.

These changes introduce the concept of cached layout values which are constructed from the result of layout queries. These cached values are automatically invalidated when a new layout operation takes place, but as long as only the query operations that stored a cached value are used and the DOM is not otherwise dirtied, the cached values will remain valid and no further layout operations will take place. 

---
- [x] These changes fix (part of) #26019
- [x] These changes do not require tests because we have no infrastructure to test whether or not reflow occurred.

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -588,16 +588,28 @@ impl Document {
         self.needs_paint.get()
     }
 
-    pub fn needs_reflow(&self) -> bool {
+    pub fn needs_reflow(&self) -> Option<ReflowTriggerCondition> {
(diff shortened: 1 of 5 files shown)


asajeffrey (approved):
Oh goody, another caching and cache invalidation mechanism! You can r=me.

PR #26077, opened 2020-03-31
webgl: Don't dirty canvas element while in immersive mode.

jdm:
There are various WebGL APIs that are supposed to trigger a frame composite at the end of the event loop when they're used. We enforce this via dirtying the canvas element and ensuring that reflow occurs for normal content. This is redundant when we're using immersive mode and incurs extra work by the layout thread and compositor that inhibits the immersive rendering performance.

---
- [x] These changes fix (part of) #26019
- [x] These changes do not require tests because we do not have infrastructure to test immersive mode on CI.

### DIFF
--- a/components/script/dom/webglrenderingcontext.rs
+++ b/components/script/dom/webglrenderingcontext.rs
@@ -502,6 +502,12 @@ impl WebGLRenderingContext {
             return;
         }
 
+        // Dirtying the canvas is unnecessary if we're actively displaying immersive
+        // XR content right now.
+        if self.global().as_window().in_immersive_xr_session() {
+            return;
+        }
+
         self.canvas
             .upcast::<Node>()
             .dirty(NodeDamage::OtherNodeDamage);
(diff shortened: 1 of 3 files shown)


components/script/dom/webglrenderingcontext.rs
```diff
@@ -502,6 +502,12 @@ impl WebGLRenderingContext {
             return;
         }
 
+        // Dirtying the canvas is unnecessary if we're actively displaying immersive
+        // XR content right now.
+        if self.global().as_window().in_immersive_xr_session() {
+            return;
```
asajeffrey: I'm a bit worried that when we come to put DOMToTexture nodes into XR sessions, we'll forget that we did this.

asajeffrey (approved):
LGTM but with a tiny bit of FUD about future interactions with DOMToTexture.

Manishearth (approved):

PR #26078, opened 2020-03-31
webgl: Avoid IPC for drawingBufferWidth/Height APIs

### LINKED ISSUE #20530: Cache drawingBufferWidth/drawingBufferHeight in WebGLRenderingContext
Those two values should be retrieved through a single `WebGLCommand` message, and we should cache (and invalidate them properly when things change of course) on the DOM side in `WebGLRenderingContext` directly.

jdm:
In the common case, there is no need to fetch the current webgl context's surface size every time these APIs are used. We only expect the values to change when the canvas is resized, or when an opaque framebuffer is bound, or the default framebuffer is bound when an opaque framebuffer was previously bound. We can store any previously determined surface size otherwise, and report it without any IPC.

---
- [x] These changes fix #20530

### DIFF
--- a/components/script/dom/webglrenderingcontext.rs
+++ b/components/script/dom/webglrenderingcontext.rs
@@ -189,6 +189,7 @@ pub struct WebGLRenderingContext {
     current_vao_webgl2: MutNullableDom<WebGLVertexArrayObject>,
     textures: Textures,
     api_type: GlType,
+    drawing_buffer_size: Cell<Option<Size2D<i32>>>,
 }
 
 impl WebGLRenderingContext {
@@ -250,6 +251,7 @@ impl WebGLRenderingContext {
                 current_vao_webgl2: Default::default(),
                 textures: Textures::new(max_combined_texture_image_units),
                 api_type: ctx_data.api_type,
+                drawing_buffer_size: Cell::new(None),
             }
         })
     }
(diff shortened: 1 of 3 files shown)


components/script/dom/webglrenderingcontext.rs
```diff
@@ -2031,6 +2041,15 @@ impl WebGLRenderingContextMethods for WebGLRenderingContext {
             return self.webgl_error(InvalidEnum);
         }
 
+        // The only time that drawingBufferWidth/drawingBufferHeight will change values is:
+        // * the canvas is resized
+        // * an opaque framebuffer is bound
+        // * the default framebuffer is bound, replacing an opaque framebuffer
```
asajeffrey: Hmm, looking at https://www.khronos.org/registry/webgl/specs/latest/1.0/#DOM-WebGLRenderingContext-drawingBufferWidth it's not obvious that attaching an opaque frame buffer should change the drawingBufferWidth/Height. Perhaps those should always be the size of the default framebuffer? So resizing should be the only thing that changes them, and we might not need any IPC to get their value?

asajeffrey (commented):
A spec question, otherwise LGTM.

jdm:
Closing this since it requires further investigation for the review comment. We can always redo the work if we start optimizing webgl perf again.

PR #26085, opened 2020-04-01
 Use a valid debug option in the example command in the Debugging guide.

utsavoza:
The debug option `dump-layer-tree` used seems maybe irrelevant to the example command, but using one from the existing debug options helps to simply copy/paste the command and see what the example is trying to demonstrate.

---

- [x] These changes do not require tests because they are doc changes exclusively.


jdm:
@bors-servo r+
Good catch!

PR #26102, opened 2020-04-03
Add support for requersting RDP access from generic-worker

### LINKED ISSUE #26082: Make it possible to start a Windows TC task that enables a RDP connection
We would not make that a default. Maybe just an option in decision_task.py or a new keyword for `try=windows`.

The idea would be to add the `rdpInfo` field as documented here: https://community-tc.services.mozilla.com/docs/reference/workers/generic-worker/multiuser-windows-payload and set the location of the json file to a private location.

It also needs a specific scope.

SimonSapin:
Fixes https://github.com/servo/servo/issues/26082

Depends on:

* [x] https://github.com/mozilla/community-tc-config/pull/252
* [x] https://github.com/servo/taskcluster-config/pull/24
* [x] https://github.com/servo/saltfs/pull/997
* [ ] https://github.com/mozilla/treeherder/pull/6265 (for relevant tasks to show up in https://treeherder.mozilla.org/#/jobs?repo=servo-try)

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -6,6 +6,7 @@
 
 import os.path
 import decisionlib
+import functools
 from decisionlib import CONFIG, SHARED
 
 
@@ -62,6 +63,10 @@ def tasks(task_for):
             "try-wpt-2020": [linux_wpt_layout_2020],
             "try-wpt-mac": [macos_wpt],
         }
+        by_branch_name["try-windows-rdp"] = [
+            functools.partial(f, rdp=True) for f in by_branch_name["try-windows"]
+        ]
+
         for function in by_branch_name.get(branch, []):
             function()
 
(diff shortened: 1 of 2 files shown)


SimonSapin:
One missing piece is deciding when to request this access. At the moment it’s only enabled for tasks started for testing a new AMI. Enabling it for *all* Windows tasks would be costly since workers are kept around for 12 hours. All `try=windows` less so, but still feels more than necessary, but maybe that’s an ok compromise. Paul mentioned adding a Homu command for this, but that requires adding support for this in Homu itself.

SimonSapin:
Based on discussion on Matrix, I’ve made this PR enable RDP for all Windows tasks in a new try branch (requires https://github.com/servo/saltfs/pull/997) to be used with `try=windows-rdp`.

SimonSapin:
Confirmed that I could connect to RDP with https://github.com/servo/taskcluster-config#mach-rdp-run-a-remote-desktop-client-for-a-windows-task for tasks created by Homu here. This is ready to land.

r? @paulrouget

paulrouget:
Thank you!

@bors-servo r+

PR #26103, opened 2020-04-03
Extract user_agent from global opts and allow embedder to provide user agent string

### LINKED ISSUE #25646: Allow embedders to provide user agent string

utsavoza:
---

- [x] These changes fix #25646
- [x] These changes do not require tests because these are refactoring changes and I'm assuming that existing tests cover these changes.

### DIFF
--- a/components/compositing/windowing.rs
+++ b/components/compositing/windowing.rs
@@ -193,6 +193,11 @@ pub trait EmbedderMethods {
         _: EmbedderProxy,
     ) {
     }
+
+    /// Returns the user agent string to report in network requests.
+    fn get_user_agent_string(&self) -> Option<String> {
+        None
+    }
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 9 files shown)


jdm:
r? @paulrouget 
Continuation of #25672.

paulrouget:
This is great!

r=me once rebased.

utsavoza:
Ok, rebased @paulrouget

jdm:
@bors-servo retry
* #26129

jdm:
@bors-servo retry
* intermittent filter server error

PR #26106, opened 2020-04-03
Check the BHM option before starting it in multi-process mode.

### LINKED ISSUE #26088: Check for Background-Hang-Monitor opts in multi-process
In single-process mode, we only start a BHM if the options has been set:

https://github.com/servo/servo/blob/af1ebe79efd799498010f196983a233aa5203d64/components/constellation/pipeline.rs#L322

But we don't perform this check before starting the BHM in multi-process mode:

https://github.com/servo/servo/blob/b3c12ad9b3404d4b9df2ccbb15b7a9259a2a0322/components/servo/lib.rs#L1019

I think we should. The same `opts` is available there as well.

qrasmont:
In multi-process mode, if the BHM option is set start with one otherwise don't.

I didn't add a test for this. However if I should I'd be happy to be pointed to where similar tests are done (meaning tests of options yielding the expected state) because I didn't find my way in all those tests.

---
- [x] These changes fix #26088 

- [x] These changes do not require tests because they are minor enough to not require one.

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -1016,8 +1016,11 @@ pub fn run_content_process(token: String) {
         create_sandbox();
     }
 
-    let background_hang_monitor_register =
-        unprivileged_content.register_with_background_hang_monitor();
+    let background_hang_monitor_register = if opts::get().background_hang_monitor {
+        unprivileged_content.register_with_background_hang_monitor()
+    } else {
+        None
+    };
 
     // send the required channels to the service worker manager
     let sw_senders = unprivileged_content.swmanager_senders();


components/servo/lib.rs
```diff
@@ -1016,23 +1016,32 @@ pub fn run_content_process(token: String) {
...
     media_platform::init();
-
-    unprivileged_content.start_all::<script_layout_interface::message::Msg,
-                                     layout_thread::LayoutThread,
-                                     script::script_thread::ScriptThread>(
-                                         true,
-                                         background_hang_monitor_register,
-                                         None,
-                                     );
+    if opts::get().background_hang_monitor {
```
gterzian, line 1025: It might be clearer to do something like:

```rust
let background_hang_monitor_register = if opts::get().background_hang_monitor {
     unprivileged_content.register_with_background_hang_monitor()
} else {
     None
};
```

gterzian (changes requested):
Thanks for the contribution, looks like the solution is right, with a small stylistic suggestion...

components/servo/lib.rs
qrasmont: Indeed!

gterzian (approved):
Looks good, and thanks for the contribution!

PR #13574, opened 2016-10-04
Implement job queue for ServiceWorkerRegistration

creativcoder:
Fixes #13424 
As of now this PR (Work in Progress), will try to implement (partially) on algorithms namely : [Schedule_Job](https://w3c.github.io/ServiceWorker/#schedule-job-algorithm), [Run_Job](https://w3c.github.io/ServiceWorker/#run-job-algorithm), [Register_Job](https://w3c.github.io/ServiceWorker/#register-algorithm).

 An issue that needs to be addressed: 
Resolving the promise value after queueing it as a `dom_manipulation_task` by wrapping it in a runnable `RunJobHandler` (which holds a `TrustedPromise`) and trying to call `resolve_native` results in a crash. [stack trace](http://paste.ofcode.org/zdjJ9npBgX6JsyWQkJjnYk)

@jdm I am not sure what's causing the crash ?

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [ ] These changes fix #__ (github issue number if applicable).

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because _____

---

This change is [Reviewable](https://reviewable.io/reviews/servo/servo/13574)

PR #26111, opened 2020-04-04
Update checking origin trustworthy align to spec

CYBAI:
While reading the [spec of ` Is origin potentially trustworthy? `](https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy), I found our second step is wrong; then, I found Josh said we didn't check it with `https` because we didn't support https in tests yet (https://github.com/servo/servo/pull/13574#discussion_r89346191).

IIRC, we've supported https wpt tests now so it might be fine to change it to align with spec.

---

### DIFF
--- a/components/script/dom/urlhelper.rs
+++ b/components/script/dom/urlhelper.rs
@@ -74,14 +74,19 @@ impl UrlHelper {
     }
     // https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy
     pub fn is_origin_trustworthy(url: &ServoUrl) -> bool {
+        // Step 1
+        if !url.origin().is_tuple() {
+            return false;
+        }
+
         // Step 3
-        if url.scheme() == "http" || url.scheme() == "wss" {
+        if url.scheme() == "https" || url.scheme() == "wss" {
             true
         // Step 4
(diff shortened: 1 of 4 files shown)


jdm:
We can avoid those failures by renaming tests/wpt/mozilla/tests/mozilla/service-workers/service-worker-registration.html to service-worker-registration.https.html and running `./mach update-manifest`.

CYBAI:
> We can avoid those failures by renaming tests/wpt/mozilla/tests/mozilla/service-workers/service-worker-registration.html to service-worker-registration.https.html and running `./mach update-manifest`.

Ah, good point! Thanks for pointing that out! Will do that!

jdm:
Needs a `./mach update-manifest`.

CYBAI:
> Needs a `./mach update-manifest`.

Updated! Thanks!

CYBAI:
@bors-servo retry

- #25513, #24622

CYBAI:
hmm, I've noticed we won't only run the failed jobs recently 🤔

CYBAI:
@bors-servo retry

- #25513

CYBAI:
@bors-servo retry

- #23290

PR #26113, opened 2020-04-04
Drive-by refactoring of some unsafe code in stylo

nox:
Some stuff I found while thinking about other things between stylo and layout.

Third commit requires Gecko changes because I removed `From<u32>` for `FontLanguageOverride` and replaced it by the unsafe `FontLanguageOverride::from_u32` method.


CYBAI:
@bors-servo retry

- #24622

PR #26116, opened 2020-04-05
Add InRealm argument to Callback trait

### LINKED ISSUE #26007: Add InRealm argument to Callback trait
Callers need to have entered a realm, so we should make the compiler check that for us.

tipowol:
- [x] These changes fix #26007


jdm:
@bors-servo r+
Thanks!

PR #26120, opened 2020-04-06
 Remove postorder traversal from layout 2020 during styling 

nox:


SimonSapin:
Diff looks good but unit tests are unhappy:

```
HTMLDivElement from 376 to 416, tests/unit/script/size_of.rs:36:1
Element from 360 to 400, tests/unit/script/size_of.rs:34:1
HTMLSpanElement from 376 to 416, tests/unit/script/size_of.rs:37:1
CharacterData from 216 to 256, tests/unit/script/size_of.rs:39:1
Node from 184 to 224, tests/unit/script/size_of.rs:33:1
Text from 216 to 256, tests/unit/script/size_of.rs:38:1
HTMLElement from 376 to 416, tests/unit/script/size_of.rs:35:1
```

Perhaps `StyleAndOpaqueLayoutData` should be boxed in `Node`’s `style_and_layout_data: UnsafeCell<Option<StyleAndOpaqueLayoutData>>` field? This would save 40 bytes whenever it is `None`.

Does this option ever stay `None` after reflow? If not, then boxing doesn’t really help. If only for non-element nodes, could this field be on `Element` instead?

nox:
It becomes `None` again for `display: none` elements and @emilio say it's significant in the wild. I just completely forgot to box it.

nox:
I fixed the units, and am assuming that "diff looks good" means that I can r=you.

@bors-servo r=SimonSapin

nox:
Actually, I can do better than that.

@bors-servo r-

jdm:
```
error[E0308]: mismatched types
   --> components/layout_thread/dom_wrapper.rs:285:9
    |
284 |     unsafe fn take_style_and_opaque_layout_data(&self) -> StyleAndOpaqueLayoutData {
    |                                                           ------------------------ expected `script_layout_interface::StyleAndOpaqueLayoutData` because of return type
285 |         self.get_jsmanaged().take_style_and_opaque_layout_data()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `script_layout_interface::StyleAndOpaqueLayoutData`, found struct `std::boxed::Box`
    |
    = note: expected struct `script_layout_interface::StyleAndOpaqueLayoutData`
               found struct `std::boxed::Box<script_layout_interface::StyleAndOpaqueLayoutData>`
For more information about this error, try `rustc --explain E0308`.
error: could not compile `layout_thread_2013`.
```

SimonSapin:
Nice work!

@bors-servo r+

nox:
@bors-servo r=SimonSapin

Oops forgot to reformat last commit.

PR #26121, opened 2020-04-06
layout_2020: Properly handle HTML element background

### LINKED ISSUE #25559: Root element background extents are wrong in layout_2020
It looks like root element background extents aren't including page margins and don't cover the page viewport.

This shows a white page:
```
<body style="background: green;"></body>
```

This shows a green page with 8 pixel white borders on the top and right:
```
<body style="background: green;">
    <div style="height: 10000px;"></div>
</body>
```

This shows a completely green page. This is what I would expect for all test cases:
```
<body style="background: green;">
    <div style="height: 10000px;"></div>
</body>
```

mrobinson:
The CSS Backgrounds and Borders specification describes the
circumstances under which the HTML element's used style for backgrounds
is determined by the BODY element. This change properly implements that
behavior, fixing a variety of tests that expect the background specified
on the BODY cover the entire page.

Fixes: #25559

---

- [x] These changes fix #25559

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -4,6 +4,7 @@
 
 use crate::context::LayoutContext;
 use crate::display_list::conversions::ToWebRender;
+use crate::dom_traversal::NodeFlags;
 use crate::fragments::{BoxFragment, Fragment, TextFragment};
 use crate::geom::{PhysicalPoint, PhysicalRect};
 use crate::replaced::IntrinsicSizes;
@@ -12,6 +13,7 @@ use euclid::{Point2D, SideOffsets2D, Size2D};
 use gfx::text::glyph::GlyphStore;
 use mitochondria::OnceCell;
 use net_traits::image_cache::UsePlaceholder;
+use servo_arc::Arc as ServoArc;
 use std::sync::Arc;
 use style::computed_values::text_decoration_style::T as ComputedTextDecorationStyle;
 use style::dom::OpaqueNode;
@@ -49,6 +51,10 @@ pub struct DisplayListBuilder<'a> {
     /// (i.e. the display list contains items of type text,
     /// image, non-white canvas or SVG). Used by metrics.
     pub is_contentful: bool,
+
+    /// The style to apply to use when painting the canvas background of the root
+    /// <html> element.
+    pub root_canvas_style: Option<ServoArc<ComputedValues>>,
 }
 
 impl<'a> DisplayListBuilder<'a> {
@@ -62,6 +68,7 @@ impl<'a> DisplayListBuilder<'a> {
             is_contentful: false,
             context,
             wr: wr::DisplayListBuilder::new(pipeline_id, viewport_size),
+            root_canvas_style: None,
         }
     }
 
@@ -78,7 +85,9 @@ impl Fragment {
         containing_block: &PhysicalRect<Length>,
     ) {
         match self {
-            Fragment::Box(b) => BuilderForBoxFragment::new(b, containing_block).build(builder),
+            Fragment::Box(fragment) => {
+                BuilderForBoxFragment::new(fragment, containing_block).build(builder);
+            },
             Fragment::AbsoluteOrFixedPositioned(_) => {},
             Fragment::Anonymous(_) => {},
             Fragment::Image(i) => {
@@ -257,6 +266,18 @@ impl<'a> BuilderForBoxFragment<'a> {
         }
     }
 
+    fn background_style(&self, builder: &DisplayListBuilder) -> ServoArc<ComputedValues> {
+        if self.fragment.flags.contains(NodeFlags::IS_HTML_ELEMENT) {
+            builder
+                .root_canvas_style
+                .as_ref()
+                .unwrap_or(&self.fragment.style)
+                .clone()
+        } else {
+            self.fragment.style.clone()
+        }
+    }
+
     fn content_rect(&self) -> &units::LayoutRect {
         self.content_rect.init_once(|| {
             self.fragment
@@ -333,19 +354,35 @@ impl<'a> BuilderForBoxFragment<'a> {
     }
 
     fn build_background(&mut self, builder: &mut DisplayListBuilder) {
+        // If this is the body element of the page and our style used on the html element
+        // then we shouldn't paint our background for our own fragment.
+        if self.fragment.flags.contains(NodeFlags::IS_BODY_ELEMENT) &&
+            builder
+                .root_canvas_style
+                .as_ref()
+                .map_or(false, |root_style| {
+                    ServoArc::ptr_eq(root_style, &self.fragment.style)
+                })
+        {
+            return;
+        }
+
         use style::values::computed::image::{Image, ImageLayer};
-        let b = self.fragment.style.get_background();
-        let background_color = self.fragment.style.resolve_color(b.background_color);
+        let background_style = self.background_style(builder).clone();
+        let background = background_style.get_background();
+        let background_color = background_style.resolve_color(background.background_color);
+
         if background_color.alpha > 0 {
             // https://drafts.csswg.org/css-backgrounds/#background-color
             // “The background color is clipped according to the background-clip
             //  value associated with the bottom-most background image layer.”
-            let layer_index = b.background_image.0.len() - 1;
-            let (_, common) = background::painting_area(self, builder, layer_index);
+            let layer_index = background.background_image.0.len() - 1;
+            let (_, common) =
+                background::painting_area(self, builder, &background_style, layer_index);
             builder.wr.push_rect(&common, rgba(background_color))
         }
         // Reverse because the property is top layer first, we want to paint bottom layer first.
-        for (index, layer) in b.background_image.0.iter().enumerate().rev() {
+        for (index, layer) in background.background_image.0.iter().enumerate().rev() {
             match layer {
                 ImageLayer::None => {},
                 ImageLayer::Image(image) => match image {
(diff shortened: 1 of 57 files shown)


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -129,6 +135,10 @@ pub(crate) struct StackingContext {
     /// The fragment that established this stacking context.
     initializing_fragment_style: Option<ServoArc<ComputedValues>>,
 
+    /// If this is the root stacking context, this contains the style for the fragment established
+    /// by the HTML element.
+    root_background_style_override: Option<ServoArc<ComputedValues>>,
```
nox, line 140: I don't feel like the name of the field matches its description very well but I have no alternative to suggest.

nox (commented):
r=me but this needs a rebase.

components/layout_2020/flow/root.rs
```diff
@@ -223,6 +226,40 @@ impl FragmentTreeRoot {
...
+                fragment
+            },
+            _ => return None,
+        };
+
+        if root_box_fragment.style.has_nontransparent_background() {
+            return Some(root_box_fragment.style.clone());
+        }
+
+        let first_child_ref = match root_box_fragment.children.get(0) {
```
SimonSapin, line 249: This and the other `.children.get(0)` call above are based on the shape of the box tree. But https://drafts.csswg.org/css-backgrounds/#body-background specifies “root element” and “child element” which are found in the DOM tree. These two trees’ shapes can be different because of `display: none`, `display: contents`, pseudo-elements, etc.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -272,7 +290,10 @@ impl StackingContext {
         while child_fragments.peek().map_or(false, |child| {
             child.section == StackingContextSection::BlockBackgroundsAndBorders
         }) {
-            child_fragments.next().unwrap().build_display_list(builder);
+            child_fragments
+                .next()
+                .unwrap()
+                .build_display_list(builder, root_background_style_override.take());
```
SimonSapin: Can this cause the canvas background to be painted above negative z-index stacking contexts? Per https://drafts.csswg.org/css2/zindex.html the canvas background should always be bottom-most.

components/layout_2020/display_list/stacking_context.rs
mrobinson: Thanks for the review! I agree with you about the name here so I've renamed this to `root_fragment_background_style`. It is only sometimes an "override."

components/layout_2020/display_list/mod.rs
```diff
@@ -254,9 +268,15 @@ impl<'a> BuilderForBoxFragment<'a> {
             border_edge_clip_id: OnceCell::new(),
             padding_edge_clip_id: OnceCell::new(),
             content_edge_clip_id: OnceCell::new(),
+            root_fragment_background_style,
         }
     }
 
+    fn background_style(&self) -> &ServoArc<ComputedValues> {
+        self.root_fragment_background_style
+            .unwrap_or(&self.fragment.style)
```
SimonSapin, line 277: If the canvas ends up taking `background-image` from a `<body>` element, what stops that image from being painted a second time when generating display lists for that element?

components/layout_2020/flow/root.rs
mrobinson: Isn't checking the node flags enough here?
  * In the case of display:none there will be no fragment marked with the HTML `NodeFlags` so this method should return `None` as expected.
  * in the case of `display:contents`, this method will not find a Fragment marked with the `BODY` `NodeFlags` so it should return the style of the HTML element. This actually isn't happening properly now (it's returning `None`), but I will fix that an upload a new version of the change.

components/layout_2020/display_list/stacking_context.rs
mrobinson: Assuming I understand your question correct, this commit shouldn't modify the that backgrounds are painted. It only changes the size of the `painting_area` and the style of the painted background.

components/layout_2020/display_list/mod.rs
SimonSapin: Since background positioning is unchanged, observing this requires alpha transparency to be involved. Test case in case there isn’t already one in WPT: http://software.hixie.ch/utilities/js/live-dom-viewer/saved/7907

tests/wpt/mozilla/tests/mozilla/scroll_root.html
```diff
@@ -7,9 +7,6 @@
         <style>
             body {
                 background: green;
-
-                /* FIXME(mrobinson): Remove this workaround when #25559 is fixed. */
-                margin: 0;
```
SimonSapin, line 12: This requires running `./mach update-manifest`

mrobinson: Thanks for pointing that out. Fixed.

components/layout_2020/display_list/mod.rs
mrobinson: Yep, it seems like this is still an issue.

components/layout_2020/display_list/stacking_context.rs
SimonSapin: No, this commit should also change the painting order. For example:

```html
<body style="border: solid; background: green">
<p style="position: absolute; z-index: -1">Text
```

My understanding of the spec is that the painting order should be: green canvas background, then text inside a stacking context with negative z-index, then border of an block-level element (that does not generate a stacking context).

But my reading of this PR is that it makes the order: text, then background (incorrectly masking text), then border. Am I reading correctly?

mrobinson: Ah, I see what you mean! I thought you were asking if this commit already modified the painting order. It doesn't change the current painting order, but it is true that the root background is currently not always placed into the proper section of the stacking context. This commit could also fix that problem though.

jdm:
```
error[E0046]: not all trait items implemented, missing: `is_body_element_for_layout`, `is_html_element_for_layout`
    --> components/layout_thread/dom_wrapper.rs:1002:1
     |
1002 | impl<'ln> ThreadSafeLayoutNode<'ln> for ServoThreadSafeLayoutNode<'ln> {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `is_body_element_for_layout`, `is_html_element_for_layout` in implementation
     |
     = help: implement the missing item: `fn is_body_element_for_layout(&self) -> bool { todo!() }`
     = help: implement the missing item: `fn is_html_element_for_layout(&self) -> bool { todo!() }`
For more information about this error, try `rustc --explain E0046`.
error: could not compile `layout_thread_2013`.
```

components/layout_2020/flow/root.rs
SimonSapin: Oh sorry, I missed this part of the spec:

>  if no boxes are generated for the element whose background would be used for the canvas (for example, if the root element has `display: none`), then the canvas background is transparent. 

This also apply if the element whose background would be used has `display: contents`. However that can’t happen on the root element because https://drafts.csswg.org/css-display-3/#transformations turns it into `display: block`. This `cfg` needs to be changed to `#[cfg(any(feature = "servo-layout-2020", feature = "gecko"))]`:

https://github.com/servo/servo/blob/3df65c02fe73a4cce9e391fa1de0b9a4d613c51b/components/style/values/specified/box.rs#L353-L359

This leaves the following "interesting" case:

No background:

* `html { display: none } body { background: green }`
* `body { display: none; background: green }`
* `body { display: contents; background: green }`

Green canvas background (including under `<body>`’s 8px margins from the UA stylesheet):

* `html { display: contents /* computes to block */ } body { background: green }`
* `html::before { content: "" } body { background: green }` (even though the box generated by `<body>` is the *second* child box of that generated by `<html>`)

I believe that to implement this spec, the checks for "is the root element" and "is the first element child of the root element" need to be based on the DOM tree through some `ThreadSafeLayoutNode` method, not based on the box tree.

But then maybe the spec doesn’t match reality and other browsers have interop on something else…

mrobinson: Okay. I'll modify the code to be entirely dependent on the DOM tree.

mrobinson:
@SimonSapin, okay I think I've implemented all your suggestions. Thanks for the review. With the changes you suggested quite a few more tests are passing!

jdm:
Needs a `./mach fmt` now.

mrobinson:
Whoops. I committed the typical error of running `./mach fmt` without committing the result. 😕 Fixed now.

mrobinson:
Okay. I think this is ready for review whenever you have a free moment, @SimonSapin.

jdm:
@SimonSapin What's the status here?

SimonSapin:
When I first started reviewing this I realized I might prefer a different approach. So I tried coding that, ran into bugs… and then got very distracted. Sorry for the delays!

Namely, this PR has display list construction code looking at a `IS_HTML_ELEMENT` flag, which is possibly buggy because it fails to also look at the `IS_BODY_ELEMENT` flag. There’s also a question of whether this flag means “is a `<body>` element” or “is the first `<body>` element of an `<html>` element that has transparent background”. But these flags should not exist at all this late in the pipeline. There is one element (or sometimes zero) that generates the background for the canvas, it seems much simpler to have imperative code that finds it once instead of having this logic weaved into the existing tree traversals.

https://github.com/servo/servo/pull/26414 is what I have so far.

PR #26117, opened 2020-04-06
Link OpenSSL statically on Unix

Darkspirit:
build.rs was removed because [Windows](https://github.com/servo/openssl-windows-build) uses OpenSSL 1.1.1 through https://github.com/servo/servo/blob/425328a64f37a795ce36516e495c386be27a6100/python/servo/packages.py#L11. Upstream openssl-src doesn't support UWP yet.

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #20783 for #26114

PR #26122, opened 2020-04-06
Support building on Void Linux

### LINKED ISSUE #26114: Void linux support for mach
I'm trying to build on void linux. Already added the packages to `./mach bootstrap`, but no luck with building. Does this have anything to do with the distro or is it another issue for cargo itself?
```sh
$ ./mach build
error: dep-info invalid
note: this is an unexpected cargo internal error
note: we would appreciate a bug report: https://github.com/rust-lang/cargo/issues/
note: cargo 1.43.0-nightly (bda50510d 2020-03-02)
Build FAILED in 0:00:07
```
I also tried with `RUST_LOG=cargo=debug`, but no difference in output. How can I get more information?

mamins1376:
Adds support for building on Void Linux.

Since this distribution uses LibreSSL, currently it's not possible to link against it. So this adds `openssl-src` feature to corresponding crates in order to activate `openssl/vendored`.
Automatic activation of this feature on Void is also included in build system.

*Note*: #26117 will activate `openssl/vendored` by default. If that's the case, I will just keep the package names in `./mach`.

---

- [X] These changes fix #26114

- [X] These changes do not require tests because AFAIK there is no way to check for required packages on each distro and they are manually checked.

### DIFF
--- a/python/servo/bootstrap.py
+++ b/python/servo/bootstrap.py
@@ -30,11 +30,11 @@ def run_as_root(command, force=False):
     if os.geteuid() != 0:
         command.insert(0, 'sudo')
     if force:
-        command += "-y"
+        command.append('-y')
     return subprocess.call(command)
 
 
-def install_linux_deps(context, pkgs_ubuntu, pkgs_fedora, force):
+def install_linux_deps(context, pkgs_ubuntu, pkgs_fedora, pkgs_void, force):
     install = False
     pkgs = []
     if context.distro in ['Ubuntu', 'Debian GNU/Linux']:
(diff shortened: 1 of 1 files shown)


Darkspirit:
Depends on #26117. At the end OpenSSL should vendored enabled for all targets. In the past, Servo vendored outdated OpenSSL versions on Windows and Android. It's still a problem that it's not updated immediately. (https://github.com/servo/servo/pull/26117#issuecomment-609689700)

python/servo/bootstrap.py
```diff
@@ -103,8 +110,20 @@ def linux(context, force=False):
...
+                 'libunwind-devel', 'MesaLib-devel', 'glib-devel', 'pkg-config',
+                 'libX11-devel', 'libXrandr-devel', 'gperf', 'bzip2-devel',
+                 'fontconfig-devel', 'cabextract', 'expat-devel', 'cmake',
+                 'cmake', 'libXcursor-devel', 'libXmu-devel', 'dbus-devel',
+                 'ncurses-devel', 'harfbuzz-devel', 'ccache', 'glu-devel',
+                 'clang', 'gstreamer1-devel', 'autoconf213',
+                 'gst-plugins-base1-devel', 'gst-plugins-bad1-devel']
+    if context.distro == "Ubuntu" and context.distro_version != "14.04":
+        pkgs_apt += ['libgstreamer1.0-dev', 'libgstreamer-plugins-base1.0-dev',
+                     'libgstreamer-plugins-bad1.0-dev']
```
paulrouget, line 123: Why is that change necessary?

mamins1376: I think it was a mistakes happens during a conflict, I'll remove that soon.

paulrouget:
Thanks!

@bors-servo r+

PR #26123, opened 2020-04-06
Add initial support for VertexAttribI4* and VertexAttribIPointer

imiklos:
Add initial support for the WebGL2 `VertexAttribI4i`, `VertexAttribI4iv`, `VertexAttribI4ui`, `VertexAttribI4uiv` and `VertexAttribIPointer` calls.

I've already updated the test expectations and noticed that I've got a crash on the `conformance2/state/gl-object-get-calls.html` test, which is related to the `GetFragDataLocation`.
The crash comes from here: https://github.com/servo/servo/blob/b4d7ec1c99259f936c9e34de157b652707d308de/components/canvas/webgl_thread.rs#L1292
The assert is invalid here because the `GetFragDataLocation` GL call could return with `-1` also. (https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glGetFragDataLocation.xhtml)

---

cc @mmatyas @zakorgy @jdm

### DIFF
--- a/components/script/dom/webgl2renderingcontext.rs
+++ b/components/script/dom/webgl2renderingcontext.rs
@@ -25,7 +25,7 @@ use crate::dom::webglprogram::WebGLProgram;
 use crate::dom::webglquery::WebGLQuery;
 use crate::dom::webglrenderbuffer::WebGLRenderbuffer;
 use crate::dom::webglrenderingcontext::{
-    uniform_get, uniform_typed, LayoutCanvasWebGLRenderingContextHelpers, Size2DExt,
+    uniform_get, uniform_typed, LayoutCanvasWebGLRenderingContextHelpers, Size2DExt, VertexAttrib,
     WebGLRenderingContext,
 };
 use crate::dom::webglsampler::{WebGLSampler, WebGLSamplerValue};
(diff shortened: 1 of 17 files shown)


components/script/dom/webgl2renderingcontext.rs
```diff
@@ -231,6 +231,62 @@ impl WebGL2RenderingContext {
...
+            let attrib = handle_potential_webgl_error!(
+                self.base,
+                vao.get_vertex_attrib(prog_attrib.location as u32)
+                    .ok_or(InvalidOperation),
+                return
+            );
+
+            let current_vertex_attrib =
+                self.base.current_vertex_attribs()[prog_attrib.location as usize];
+            let attrib_data_base_type = if attrib.enabled_as_array {
```
jdm, line 270: This check looks backward to me, and when I invert it then all the test failures in attrib-type-match.html and gl-vertex-attrib-i-render.html disappear for me.

jdm (changes requested):
Thanks for implementing this!

jdm:
I agree with you that the assert for getFragDataLocation should be removed.

components/script/dom/webgl2renderingcontext.rs
imiklos: Thank you! You are right! I've updated the test expectations but the test failures did not disappear on my side. Could it be driver related?

imiklos:
@jdm Thank you for the review! I've removed the assert from the `glGetFragDataLocation`, now on the `gl-object-get-calls.html` test I get an `Unexpected WebGL error` from the GL side _( related to the `GetActiveUniformBlockName` )_ and the test is timing out even when I use `--timeout-multiplier`.

jdm:
The GetActiveUniformBlockName panic is caused by #26134.

PR #26141, opened 2020-04-08
implement mach clean-uwp

paulrouget:
Often it's necessary to cleanup the Visual Studio artifacts only. This makes it possible.


jdm:
@bors-servo retry
* #25513

PR #26152, opened 2020-04-08
Fix missing Babylon.js content

### LINKED ISSUE #26149: All babylon content in immersive mode stopped rendering
http://servo-builds.s3.amazonaws.com/nightly/uwp/2020-04-01T09:15:23Z-ServoApp_1.0.0.0_Test.zip successfully shows the demos from http://servo.org/hl-home/wip.html in immersive mode. http://servo-builds.s3.amazonaws.com/nightly/uwp/2020-04-06T09:20:03Z-ServoApp_1.0.0.0_Test.zip does not. We regressed something, but it only affects babylon demos, not three.js demos.

jdm:
Since the surfman upgrade work already merged to master on webxr, and this fix for Babylon.js content is extremely urgent, this change puts us temporarily on a branch that doesn't include the surfman upgrade.

---
- [x] These changes fix #26149
- [x] These changes do not require tests because there are no tests for the openxr backend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3139,9 +3139,9 @@ dependencies = [
 
 [[package]]
 name = "libloading"
-version = "0.5.0"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c3ad660d7cb8c5822cd83d10897b0f1f1526792737a179e73896152f85b88c2"
+checksum = "f2b111a074963af1d37a139918ac6d49ad1d0d5e47f72fd55388619691a7d753"
 dependencies = [
  "cc",
  "winapi",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -32,3 +32,7 @@ mio = { git = "https://github.com/servo/mio.git", branch = "servo" }
 winapi = { git = "https://github.com/servo/winapi-rs", branch = "patch-1" }
 spirv_cross = { git = "https://github.com/servo/spirv_cross", branch = "wgpu-servo" }
 backtrace = { git = "https://github.com/MeFisto94/backtrace-rs", branch = "fix-strtab-freeing-crash" }
+
+[patch."https://github.com/servo/webxr"]
+webxr = { git = "https://github.com/jdm/webxr", branch = "surfman-0.1" }
+webxr-api = { git = "https://github.com/jdm/webxr", branch = "surfman-0.1" }
```
jdm, line 38: I had to use jdm instead of servo here because cargo requires a different source for a patch, and a branch from the same repository does not count.

Manishearth: Why not just `cargo update -p webxr --precise <hash>` instead? we can later mark the hash with a tag.

Manishearth (approved):

Cargo.toml
jdm: Smart.

PR #26164, opened 2020-04-10
Dirty canvas when exiting immersive sessions

### LINKED ISSUE #155: Returning to home screen from immersive mode and reactivating app shows openxr missing frame animation
STR:
1. visit a demo from https://servo.org/hl-home/ and enter immersive mode
1. return to the home screen via the home button (rather than the context menu)
1. activate the app, causing immersive mode to be re-entered immediately

Expected:
The existing immersive scene reappears.

Actual:
The spinning white balls appear.

### LINKED ISSUE #26162: Navigation doesn't work after leaving immersive mode
STR:
1. visit any demo from the https://servo.org/hl-home/ homepage
1. enter immersive mode
1. use the context menu to leave immersive mode
1. press the home button

Something broke this recently, since I have definitely successfully gone back in history in previous builds.

Manishearth:
Fixes https://github.com/servo/servo/issues/26162, fixes servo/webxr#155

We basically end up in a situation where the main thread is asleep. It's unclear to me why it doesn't wake up after attempts to interact with the screen, but we stopped dirtying the canvas during the immersive session (https://github.com/servo/servo/pull/26077) so the main thread wasn't awake initially.

This is probably not the cleanest fix -- we really should not be ending up in situations where the main thread is perma-asleep -- however, we should be dirtying the canvas after exiting immersive mode _anyway_ since the contents need to be shown to the user, so this fix is still valid, even if it's not fixing the root cause.

### DIFF
--- a/components/script/dom/xrsystem.rs
+++ b/components/script/dom/xrsystem.rs
@@ -3,13 +3,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::dom::bindings::cell::DomRefCell;
-use crate::dom::bindings::codegen::Bindings::XRRenderStateBinding::XRRenderStateMethods;
-use crate::dom::bindings::codegen::Bindings::XRSessionBinding::XRSessionMethods;
 use crate::dom::bindings::codegen::Bindings::XRSystemBinding::XRSessionInit;
 use crate::dom::bindings::codegen::Bindings::XRSystemBinding::{XRSessionMode, XRSystemMethods};
-use crate::dom::bindings::codegen::Bindings::XRWebGLLayerBinding::{
-    XRWebGLLayerMethods, XRWebGLRenderingContext,
-};
 use crate::dom::bindings::conversions::{ConversionResult, FromJSValConvertible};
 use crate::dom::bindings::error::Error;
 use crate::dom::bindings::refcounted::{Trusted, TrustedPromise};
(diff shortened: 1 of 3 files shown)


Manishearth:
I think this is also part of the fix for https://github.com/servo/webxr/issues/155 , let me check

Manishearth:
Yep, this same trick fixes the other issue.

jdm:
Unused import warning is breaking the build.

asajeffrey:
This is dirtying the canvas rather than the document, does it wake up the main thread if the canvas is detached from the DOM?

Manishearth:
Fixed.

@asajeffrey hmm. This would have already been broken then, right? That seems to be rooted in embedder events not getting consumed when the thread is asleep.

components/script/dom/xrsystem.rs
```diff
@@ -5,6 +5,7 @@
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::XRSystemBinding::XRSessionInit;
 use crate::dom::bindings::codegen::Bindings::XRSystemBinding::{XRSessionMode, XRSystemMethods};
+use crate::dom::bindings::codegen::Bindings::XRWebGLLayerBinding::XRWebGLRenderingContext;
```
jdm, line 8:
```
warning: unused import: `crate::dom::bindings::codegen::Bindings::XRWebGLLayerBinding::XRWebGLRenderingContext`
 --> components\script\dom\xrsystem.rs:8:5
  |
8 | use crate::dom::bindings::codegen::Bindings::XRWebGLLayerBinding::XRWebGLRenderingContext;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
```

jdm (commented):
I am still able to reproduce #26162 with these changes.

jdm:
Since I don't think #26179 is related to these changes, I think we should go ahead and merge them with the unused import removed.

jdm:
@bors-servo retry
* #24622

PR #26166, opened 2020-04-10
Remove DomRefCell wrapper for CanvasState

### LINKED ISSUE #23436: Remove Cell/DomRefCell member wrappers from CanvasState struct
CanvasState was introduced as part of a big refactor in #23381 that extracted common aspects of the offscreen canvas rendering context and the 2d canvas rendering context. Since the CanvasState struct is stored inside of a DomRefCell itself, that means that we can modify the methods for CanvasState to accept `&self` or `&mut self` directly, so the Cell and DomRefCell members inside of CanvasState are redundant and can be removed.

Code: `components/script/dom/canvasrenderingcontext2d.rs`

To verify that everything behaves the same after changing the code, run `./mach test-wpt tests/wpt/web-platform-tests/2dcontext/ tests/wpt/web-platform-tests/offscreen-canvas`.

utsavoza:
---

- [x] These changes fix #23436

The PR removes the DomRefCell wrapper for CanvasState from `CanvasRenderingContext2D` and `OffscreenCanvasRenderingContext2D`. The reason for keeping member wrappers in CanvasState instead of overall CanvasState wrapper itself is because I believe that removing member wrappers would make it rather difficult to refactor the internal CanvasState methods in order to cover certain edge cases. 

[For example](https://github.com/servo/servo/issues/23436#issuecomment-611089157): Drawing a canvas inside of itself would require getting and providing certain parameters for source canvas such as `canvas_id` and `is_origin_clean` that are currently accessed through source canvas context. Also, we might run into similar issue when creating patterns using canvas.

### DIFF
--- a/components/script/canvas_state.rs
+++ b/components/script/canvas_state.rs
@@ -175,14 +175,6 @@ impl CanvasState {
         &self.missing_image_urls
     }
 
-    pub fn get_state(&self) -> &DomRefCell<CanvasContextState> {
-        &self.state
-    }
-
-    pub fn get_saved_state(&self) -> &DomRefCell<Vec<CanvasContextState>> {
-        &self.saved_states
-    }
-
     pub fn get_canvas_id(&self) -> CanvasId {
         self.canvas_id.clone()
     }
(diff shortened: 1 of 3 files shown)


utsavoza:
Hey @jdm, please let me know what you think.

jdm:
@bors-servo r+
Looks good! Thanks for fixing this!

jdm:
@bors-servo retry
* #25513

PR #26170, opened 2020-04-11
Add FromJSValConvertible for enums

### LINKED ISSUE #26169: Should we be using FromJSValConvertible for enum conversions?
It seems like we don't have FromJSValConvertible implementations for enums. I can't think of a good reason for doing this, and this also means our bindings code is littered with code like

```
        let arg0: XRReferenceSpaceType = match find_enum_value(*cx, HandleValue::from_raw(args.get(0)), XRReferenceSpaceTypeValues::pairs) {
            Err(_) => { return false;
         },
            Ok((None, search)) => { throw_type_error(*cx, &format!("'{}' is not a valid enum value for enumeration 'XRReferenceSpaceType'.", search)); return false;
         },
            Ok((Some(&value), _)) => value,
        };
```

We should probably fix this.

Furthermore, this prevents us from using `sequence<EnumType>` in dictionaries in WebIDL

Manishearth:
Fixes https://github.com/servo/servo/issues/26169

This adds FromJSValConvertible implementations for webidl enums, which allows them to be used within `sequence<>`, etc. 

This also uses these implementations when converting arguments.

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -1004,17 +1004,16 @@ def getJSToNativeConversionInfo(type, descriptorProvider, failureCode=None,
                             "yet")
         enum = type.inner.identifier.name
         if invalidEnumValueFatal:
-            handleInvalidEnumValueCode = onFailureInvalidEnumValue(failureCode, 'search').define()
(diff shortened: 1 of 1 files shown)


asajeffrey (approved):
Yes please!

Issue #11681, opened 2016-06-09
Implement CanvasRenderingContext2d.fillText


sole:
Spec: https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-filltext
Code: `components/script/dom/canvasrenderingcontext2d.rs`, `components/script/dom/weibidls/CanvasRenderingContext2D.webidl`, `components/canvas/canvas_paint_thread.rs`

(I hope I identified correctly the code files - followed the style in #5411)

nox:
This was not actually implemented, we just print an error.

Manishearth:
Relevant code from Gecko

 - [overall algorithm](https://dxr.mozilla.org/mozilla-central/rev/c2593a3058afdfeaac5c990e18794ee8257afe99/dom/canvas/CanvasRenderingContext2D.cpp#4240-4493)
 - [bidi processor](https://dxr.mozilla.org/mozilla-central/rev/c2593a3058afdfeaac5c990e18794ee8257afe99/dom/canvas/CanvasRenderingContext2D.cpp#3977,3978-4237) (goes through `nsBidiPresUtils::ProcessText`)

jdm:
Relevant parts of Servo/Azure that may be useful:
* [convert_text_run_to_glyphs](https://github.com/servo/servo/blob/19495a76e49f2bfde67b80ff7d104d8594377738/components/layout/display_list/builder.rs#L2939-L2966)
* [DrawTarget::fill_glyphs](https://github.com/servo/rust-azure/blob/1ca68f302f5e914e2c63b0ec5b72af8f761f977d/src/azure_hl.rs#L828-L842)
* [AzCreateScaledFontForNativeFont(https://github.com/servo/rust-azure/blob/1ca68f302f5e914e2c63b0ec5b72af8f761f977d/src/azure-c.cpp#L551)
* some subset of [flush_clump_to_list](https://github.com/servo/servo/blob/19495a76e49f2bfde67b80ff7d104d8594377738/components/layout/text.rs#L148)

jdm:
raqote has simple APIs for drawing text based on font-kit. Now that we have a working raqote-based canvas implementation, we should try enabling raqote's `text` cargo feature and using it: https://github.com/jrmuizel/raqote/blob/846374c6d48a39c8d080389ef43eb0ce215932ff/examples/capabilities.rs#L124-L142 .

jdm:
[...]

PR #26176, opened 2020-04-13
Implement CanvasRenderingContext2D.font property

pylbrecht:
These changes implement the font property of the Canvas 2D API.

---

- [ ] These changes fix part of #25331

### DIFF
--- a/components/layout/query.rs
+++ b/components/layout/query.rs
@@ -90,6 +90,9 @@ pub struct LayoutThreadData {
 
     /// A queued response for the viewport dimensions for a given browsing context.
     pub inner_window_dimensions_response: Option<TypedSize2D<f32, CSSPixel>>,
+
+    /// A queued response for the resolved font property of a canvas element.
+    pub canvas_font_response: String,
 }
 
 pub struct LayoutRPCImpl(pub Arc<Mutex<LayoutThreadData>>);
@@ -1094,4 +1097,12 @@ fn is_last_table_cell() -> bool {
 fn is_last_table_row() -> bool {
     // FIXME(ferjm) Implement this.
     false
+}
+
+pub fn process_canvas_font_request<'dom>(
+    _node: impl LayoutNode<'dom>,
+    _font: &str,
+    _layout_root: &mut dyn Flow,
+) -> String {
+    unimplemented!()
 }
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -50,12 +50,14 @@ use layout::flow_ref::FlowRef;
 use layout::incremental::{RelayoutMode, SpecialRestyleDamage};
 use layout::layout_debug;
 use layout::parallel;
+use layout::query::{
+    process_canvas_font_request, process_offset_parent_query, process_resolved_style_request,
+};
 use layout::query::{process_client_rect_query, process_element_inner_text_query};
 use layout::query::{
     process_content_box_request, process_content_boxes_request, LayoutRPCImpl, LayoutThreadData,
 };
 use layout::query::{process_node_scroll_area_request, process_node_scroll_id_request};
-use layout::query::{process_offset_parent_query, process_resolved_style_request};
 use layout::sequential;
 use layout::traversal::{
     ComputeStackingRelativePositions, PreorderFlowTraversal, RecalcStyleAndConstructFlows,
@@ -575,6 +577,7 @@ impl LayoutThread {
                 nodes_from_point_response: vec![],
                 element_inner_text_response: String::new(),
                 inner_window_dimensions_response: None,
+                canvas_font_response: String::new(),
             })),
             webrender_image_cache: Arc::new(RwLock::new(FnvHashMap::default())),
             paint_time_metrics: paint_time_metrics,
@@ -1600,8 +1603,10 @@ impl LayoutThread {
                         .get(&browsing_context_id)
                         .cloned();
                 },
-                &QueryMsg::CanvasFontQuery(_node, ref _font) => {
-                    unimplemented!();
+                &QueryMsg::CanvasFontQuery(node, ref font) => {
+                    let node = unsafe { ServoLayoutNode::new(&node) };
+                    rw_data.canvas_font_response =
+                        process_canvas_font_request(node, font, root_flow);
                 },
             },
             ReflowGoal::Full | ReflowGoal::TickAnimations => {},
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -1255,6 +1255,9 @@ impl LayoutThread {
                         &QueryMsg::InnerWindowDimensionsQuery(_) => {
                             rw_data.inner_window_dimensions_response = None;
                         },
+                        &QueryMsg::CanvasFontQuery(..) => {
+                            unimplemented!();
+                        },
                     },
                     ReflowGoal::Full | ReflowGoal::TickAnimations => {},
                 }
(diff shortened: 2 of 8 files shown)


pylbrecht:
I just started looking into this, but I opened this PR for tracking progress. Definitely have to wrap my head around how the property's value is parsed as a CSS `<'font'>` as described [in this comment](https://github.com/servo/servo/issues/11681#issuecomment-586763530) first.

pylbrecht:
I've been a bit busy with applications and job interviews lately, but there's a good chance I get to work on this on the weekend!

pylbrecht:
I investigated a bit more and am currently facing a problem:

How should the `font` attribute be stored in the [`CanvasContextState`](https://dxr.mozilla.org/servo/source/components/script/canvas_state.rs#74)?

Currently I use [`style::properties::shorthands::font::parse_value()`](https://doc.servo.org/style/properties/shorthands/font/fn.parse_value.html) to parse the `DOMString` in [`CanvasState::set_font()`](https://github.com/servo/servo/pull/26176/files#diff-587cd504daeade12b5ea90149c65788dR1594), which gives me a [`style::properties::shorthands::font::Longhands`](https://doc.servo.org/style/properties/shorthands/font/struct.Longhands.html). This struct holds all relevant information about the font.

I tried to just add a `font` member to `CanvasContextState` of type `style::properties::shorthands::font::Longhands`, but this results in a lot of issues due to missing trait implementations, which I'm not really comfortable implementing:
```
error[E0277]: the trait bound `style::properties::shorthands::font::Longhands: std::clone::Clone` is not satisfied
   --> components/script/canvas_state.rs:100:5
    |
100 |     font: font::Longhands,
    |     ^^^^^^^^^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `style::properties::shorthands::font::Longhands`
    |
    = note: required by `std::clone::Clone::clone`
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: no method named `trace` found for reference `&style::properties::shorthands::font::Longhands` in the current scope
   --> components/script/canvas_state.rs:84:17
    |
84  | #[derive(Clone, JSTraceable, MallocSizeOf)]
    |                 ^^^^^^^^^^^ method not found in `&style::properties::shorthands::font::Longhands`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
note: `dom::bindings::trace::JSTraceable` defines an item `trace`, perhaps you need to implement it
   --> components/script/dom/bindings/trace.rs:165:1
    |
165 | pub unsafe trait JSTraceable {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `style::properties::shorthands::font::Longhands: malloc_size_of::MallocSizeOf` is not satisfied
  --> components/script/canvas_state.rs:84:30
   |
84 | #[derive(Clone, JSTraceable, MallocSizeOf)]
   |                              ^^^^^^^^^^^^ the trait `malloc_size_of::MallocSizeOf` is not implemented for `style::properties::shorthands::font::Longhands`
   |
   = note: required by `malloc_size_of::MallocSizeOf::size_of`
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
How would I work around this?

Right now I'm thinking of introducing a new `FontProperties` struct in [`canvas_state.rs`](https://dxr.mozilla.org/servo/source/components/script/canvas_state.rs) to store `Longhands`' members in `CanvasContextState`. Then I hopefully would be able to just `#[derive ...]` and `#[ignore_malloc_size_of = ...]` on `FontProperties` to get rid of the issues.

I'll try that idea the next days, but I'm open for input, if there's a better way to approach this.

jdm:
Ok, there are a few things here.
1. the MallocSizeOf and JSTraceable errors aren't difficult to get around (You can use `#[ignore_malloc_size_of = ""]` and add a `unsafe_no_jsmanaged_fields!` entry in [this code](https://github.com/servo/servo/blob/e697e6cca70bec57b5ed9512b95c2a72359d6ca1/components/script/dom/bindings/trace.rs#L198))
1. looking more closely at the gecko code, we need to not just parse the CSS value, but also resolve it and obtain computed values (like in [this code](https://searchfox.org/mozilla-central/rev/55a4faa52f72918efa51150d127aebdc057dc6cf/dom/canvas/CanvasRenderingContext2D.cpp#2272-2306)). This will help us, because the types for computed values are cloneable (eg. https://doc.servo.org/style/values/computed/font/struct.FontFamily.html).

To turn the specified parsed values into computed values, I think we need to add a new layout query which will:
1. send a message to the layout thread with the canvas element and the unparsed font value
1. when processing the layout query, parse the font value into the longhands, then use [`ToComputedValue::to_computed_value`](https://doc.servo.org/style/values/computed/trait.ToComputedValue.html) on the longhand values that we need
1. use the font cache thread to load the resolved font and obtain the bytes from the resulting font
1. make the query result be the font bytes, and store those in the CanvasState

Having the font bytes available means we can send them to the canvas rendering thread and use Font::from_bytes from font-kit to create a font that we can pass to raqote.

pylbrecht:
Oh wow, thanks a lot! I would have never figured this out by myself.
I will take a closer look into it the next days.

pylbrecht:
Aside from implementing `process_canvas_font_request()`, I don't really know how to make a `QueryMsg::CanvasFontQuery` from inside the canvas thread.

I didn't have to work across threads so far, so this is new to me.
By reading some code I'm sensing that I need some kind of channel (send/recv), but I'm not sure if there's already something in place to communicate with layout or if I have to add this by myself.

jdm:
Why is the canvas thread triggering the query? We should be able to start it in the script thread when JS invokes SetFont.

pylbrecht:
That's true, I haven't thought about this.. just to check: you mean in `CanvasRenderingContext2DMethods::SetFont()`?

How do I make a query from script to layout then?

jdm:
Take a look at how Window::GetComputedStyle is implemented, since that also performs a layout query.

jdm:
And yes, that is the SetFont that I meant.

pylbrecht:
While looking at https://searchfox.org/mozilla-central/source/dom/canvas/CanvasRenderingContext2D.cpp#2272-2282 and searching Servo's docs, I found [`Window::GetComputedValue()`](https://dxr.mozilla.org/servo/source/components/script/dom/window.rs#1090) and looking at the code of that method brought me to [`CSSStyleDeclaration::get_computed_value()`](https://dxr.mozilla.org/servo/source/components/script/dom/cssstyledeclaration.rs#237).

I want to get the `parentStyle` in `CanvasRenderingContext2D::SetFont()` by leveraging one of these methods, but without any success so far.
`CSSStyleDeclaration::get_computed_value()` is private and in case of `Window::GetComputedValue()` I don't know how to get the `Window` instance of the canvas element in `CanvasRenderingContext2D::SetFont()`.

Does this approach to get the `parentStyle` make any sense?

pylbrecht:
@jdm helped me out on [Matrix](https://chat.mozilla.org/#/room/#servo:mozilla.org), so my question from the previous comment is answered for now.

pylbrecht:
One more thing: [`Window::GetComputedStyle()`](https://doc.servo.org/script/dom/window/struct.Window.html#method.GetComputedStyle) takes a `dom::element::Element` and I don't know how to convert a `CanvasRenderingContext2D` to `Element` to pass it to the method.

pylbrecht:
Working on this feels a little frustrating and since I'm starting a full-time job next week, I will step back from contributing for now. I may come back to work on Servo issues once I accustomed myself to the new schedule, but for now I'll be focusing my efforts on the new job.

PR #26181, opened 2020-04-14
style: Refactor some animations code

mrobinson:
This change modifies the names of some methods to make it clearer what
they are doing. It also adds some clarifying comments to explain some
confusing behavior.

---

### DIFF
--- a/components/style/matching.rs
+++ b/components/style/matching.rs
@@ -12,6 +12,8 @@ use crate::context::{ElementCascadeInputs, QuirksMode, SelectorFlagsMap};
 use crate::context::{SharedStyleContext, StyleContext};
 use crate::data::ElementData;
 use crate::dom::TElement;
+#[cfg(feature = "servo")]
+use crate::dom::{OpaqueNode, TNode};
 use crate::invalidation::element::restyle_hints::RestyleHint;
 use crate::properties::longhands::display::computed_value::T as Display;
 use crate::properties::ComputedValues;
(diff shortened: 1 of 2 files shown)


components/style/matching.rs
```diff
@@ -588,46 +603,50 @@ trait PrivateMatchMethods: TElement {
...
-    //
-    // Also, this function gets the old style, for some reason I don't really
-    // get, but the functions called (mainly update_style_for_animation) expects
-    // the new style, wtf?
     #[cfg(feature = "servo")]
-    fn update_animations_for_cascade(
+    fn collect_and_update_style_for_expired_transitions(
         &self,
         context: &SharedStyleContext,
+        node: OpaqueNode,
```
emilio, line 610: `node` is always `self.as_node().opaque()`. Either assert it, or make this a static function as it doesn't use `self`?

components/style/matching.rs
```diff
@@ -588,46 +603,50 @@ trait PrivateMatchMethods: TElement {
...
+                    frame.property_animation.update(Arc::make_mut(style), 1.0);
+                    expired_transitions.push(frame.property_animation);
+                }
+            }
+        }
+    }
+
+    #[cfg(feature = "servo")]
+    fn update_style_for_animations_and_collect_running_transitions(
+        &self,
```
emilio, line 632: Same, this no longer uses `self`, could be made a standalone function.

emilio (approved):
r=me with those two cleanups.

components/style/matching.rs
mrobinson: Thanks for the review! I've converted these two to static methods, just for code organization purposes.

mrobinson:
@bors-servo retry
  * #24622
  * #25513

PR #26183, opened 2020-04-14
Use hash_map::Entry to save a map lookup

tamird:
A bit of golf noticed via https://medium.com/programming-servo/programming-servo-workers-at-your-service-db71e5943511.

r? @gterzian

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2623,32 +2623,31 @@ where
     }
 
     fn handle_register_serviceworker(&mut self, scope_things: ScopeThings, scope: ServoUrl) {
-        let origin = scope.origin();
+        // This match is equivalent to Entry.or_insert_with but allows for early return.
+        let sw_manager = match self.sw_managers.entry(scope.origin()) {
+            Entry::Occupied(entry) => entry.into_mut(),
+            Entry::Vacant(entry) => {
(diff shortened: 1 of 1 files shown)


components/constellation/constellation.rs
```diff
@@ -2623,32 +2623,30 @@ where
...
 
     fn handle_register_serviceworker(&mut self, scope_things: ScopeThings, scope: ServoUrl) {
-        let origin = scope.origin();
-
-        if let Some(mgr) = self.sw_managers.get(&origin) {
-            let _ = mgr.send(ServiceWorkerMsg::RegisterServiceWorker(scope_things, scope));
-        } else {
-            let (own_sender, receiver) = ipc::channel().expect("Failed to create IPC channel!");
+        // This match is equivalent to Entry.or_insert_with but allows for early return.
+        let _ = match self.sw_managers.entry(scope.origin()) {
```
gterzian, line 2627: Ok, how about assigning to something like `sw_sender`, and then using that to send the message?

gterzian (changes requested):
Nice catch! 

Could we perhaps still assign the result of the match and call `send` on it? I think it would be clearer than the currently proposed `let _ = match {}.send`.

To fix the formatting, you can also do `./mach fmt`.

tamird:
Thanks, done!

tamird:
At last, this is green.

gterzian:
Thanks!

@bors-servo r+

PR #26191, opened 2020-04-15
Automatically enable WebGL 2 for specific domains.

jdm:
Our WebGL 2 implementation has progressed to the point where some but not all content works well. One particular example of non-working content is Babylon.js; if the WebGL2RenderingContext interface is present, it automatically tries to use it and causes panics in Servo. However, there are demos like http://www.servoexperiments.com/webxr-particles/ that work fine today and would be great to feature on our HoloLens homepage.

This change adds what is effectively an origin trial, where we can identify content that works well, ensure it's running in a list of blessed hosts, and reap the benefits.


jdm:
@bors-servo r-
Format is unahppy for some reason.

jdm:
@bors-servo retry
* #26192

PR #26195, opened 2020-04-15
Hint to the runtime to GC after an rAF

asajeffrey:
Hints to the runtime that after an rAF is a good time to GC.

---

- [x] These changes do not require tests because it's a perf thing

### DIFF
--- a/components/script/dom/xrsession.rs
+++ b/components/script/dom/xrsession.rs
@@ -40,6 +40,11 @@ use dom_struct::dom_struct;
 use euclid::{Rect, RigidTransform3D, Transform3D};
 use ipc_channel::ipc::IpcReceiver;
 use ipc_channel::router::ROUTER;
+use js::jsapi::GCReason;
+use js::jsapi::IsIdleGCTaskNeeded;
+use js::jsapi::JS_GetRuntime;
+use js::jsapi::JS_GC;
+
 use metrics::ToMs;
 use profile_traits::ipc;
 use std::cell::Cell;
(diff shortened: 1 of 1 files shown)


components/script/dom/xrsession.rs
```diff
@@ -423,6 +428,28 @@ impl XRSession {
...
+    fn gc_hint(&self) {
+        let cx = *self.global().get_cx();
+        let rt = unsafe { JS_GetRuntime(cx) };
+
+        if unsafe { IsIdleGCTaskNeeded(rt) } {
+            #[cfg(feature = "xr-profile")]
+            let gc_start = time::precise_time_ns();
+
+            debug!("GC at the end of an XR rAF");
+            unsafe { JS_GC(cx, GCReason::API) };
```
jdm, line 445: Should we use RunIdleGCTask instead? https://searchfox.org/mozilla-central/source/js/src/jsapi.cpp#1311

asajeffrey: Ah, that function does almost exactly what we want. It only triggers a minor gc though, but I think that's fine.

asajeffrey: Done.

Manishearth (approved):

jdm:
I would prefer not to merge this until someone has evaluated the performance on a device.

asajeffrey:
Fair enough.

asajeffrey:
I filed a relevant spec issue: https://github.com/immersive-web/webxr/issues/1010

Issue #23813, opened 2019-07-19
Expose configurable logging through C API


jdm:
Right now the C API has a hardcoded logger set up on Windows that relies on OutputDebugString. We should provide some callbacks for C API users to perform their own logging instead.

I'm imagining a new member of CInitOpts which is a [function pointer](https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html) which receives a string to log. This is a callback that the embedder can provide to log the string in some preferred way - in the case of the UWP embedding, this will call OutputDebugString. The VSLogger in ports/libsimpleservo/capi/ can become a logger for all platforms, not just Windows, by storing this new function pointer as a member and invoking it every time we decide to log a message.

angelortiz1007:
Got some questions.

In C you specify a function pointer as (example):
int (*myptrfunc)(int I);

In CInitOptions in rust I specified: pub f: dyn Fn(i32) -> i32,  See below.  I have don't have "f" implemented yet, and I know the funciton signature needs to take a string.  This just a proof of concept.

```
pub struct CInitOptions {
    pub args: *const c_char,
    pub url: *const c_char,
    pub width: i32,
    pub height: i32,
    pub density: f32,
    pub vr_pointer: *mut c_void,
    pub enable_subpixel_text_antialiasing: bool,
    pub vslogger_mod_list: *const *const c_char,
    pub vslogger_mod_size: u32,
    **pub f: dyn Fn(i32) -> i32,**
}
```

However, this does not allow a clean build.  I get:
    Blocking waiting for file lock on package cache
   Compiling simpleservo_capi v0.0.1 (C:\git-mozilla\servo-angelortiz1007\ports\libsimpleservo\capi)
error[E0277]: the size for values of type `(dyn std::ops::Fn(i32) -> i32 + 'static)` cannot be known at compilation time
   --> ports\libsimpleservo\capi\src\lib.rs:267:5
    |
267 |     opts: CInitOptions,
    |     ^^^^ doesn't have a size known at compile-time
    |
    = help: within `CInitOptions`, the trait `std::marker::Sized` is not implemented for `(dyn std::ops::Fn(i32) -> i32 + 'static)`
    = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
    = note: required because it appears within the type `CInitOptions`
    = note: all local variables must have a statically known size
    = help: unsized locals are gated as an unstable feature

[...]

PR #26196, opened 2020-04-15
Dump all UWP logging output to a file

jdm:
These changes ensure that we keep getting debug output in the VS output window, but we also write each message to a file in the app's local data directory. These changes also extend the C API a bit to support more generic logging facilities, removing some of the Windows-specific nature of the VSLogger. Fixes part of #23813.

### DIFF
--- a/ports/libsimpleservo/capi/src/lib.rs
+++ b/ports/libsimpleservo/capi/src/lib.rs
@@ -27,17 +27,20 @@ use std::os::raw::{c_char, c_uint, c_void};
 use std::panic::{self, UnwindSafe};
 use std::slice;
 use std::str::FromStr;
-use std::sync::RwLock;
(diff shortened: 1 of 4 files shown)


jdm:
This is ready for review. I gated the log output on debug builds by default to avoid the risk of hurting perf by writing to a file.

Manishearth (approved):
@bors-servo r+

Manishearth:
```
 unused variable: `handler`
  --> ports/libsimpleservo/capi/src/lib.rs:70:27
   |
70 | fn redirect_stdout_stderr(handler: LogHandlerFn) -> Result<(), String> {
   |                           ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_handler`
   |
   = note: `-D unused-variables` implied by `-D warnings
```

PR #26198, opened 2020-04-16
Report error when vcvarsall fails

Manishearth:
In trying to get my Windows cross build working on my desktop, I ended up spending a lot of time trying to replicate a failure within servo's build system that ultimately turned out to be vcvarsall silently failing (I was missing a trailing slash in my `VSINSTALLDIR` env var, which vcvarsall does not handle well at all)

We should report an error when this happens.

r? @jdm

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -315,11 +315,15 @@ class MachCommands(CommandBase):
             process = subprocess.Popen('("%s" %s > nul) && "python" -c "import os; print(repr(os.environ))"' %
                                        (os.path.join(vs_dirs['vcdir'], "Auxiliary", "Build", "vcvarsall.bat"), "x64"),
                                        stdout=subprocess.PIPE, shell=True)
-            stdout, _ = process.communicate()
+            stdout, stderr = process.communicate()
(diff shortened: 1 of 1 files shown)


python/servo/build_commands.py
```diff
@@ -315,11 +315,15 @@ def build(self, release=False, dev=False, jobs=None, params=None, media_stack=No
             process = subprocess.Popen('("%s" %s > nul) && "python" -c "import os; print(repr(os.environ))"' %
                                        (os.path.join(vs_dirs['vcdir'], "Auxiliary", "Build", "vcvarsall.bat"), "x64"),
                                        stdout=subprocess.PIPE, shell=True)
-            stdout, _ = process.communicate()
+            stdout, err = process.communicate()
```
jdm, line 318: I think this should be stderr?

jdm (changes requested):

python/servo/build_commands.py
Manishearth: Oops, i'd manually copied over the patch

jdm:
@bors-servo retry
* #25513 #26203

PR #26200, opened 2020-04-16
style: Add an iterator for transition properties

mrobinson:
This simplifies the code a bit and also will allow us to more easily
make improvements to servo's animation implementation in the future.

---

### DIFF
--- a/components/style/properties/helpers/animated_properties.mako.rs
+++ b/components/style/properties/helpers/animated_properties.mako.rs
@@ -11,7 +11,7 @@
 
 #[cfg(feature = "gecko")] use crate::gecko_bindings::structs::nsCSSPropertyID;
 use itertools::{EitherOrBoth, Itertools};
-use crate::properties::{CSSWideKeyword, PropertyDeclaration};
+use crate::properties::{CSSWideKeyword, PropertyDeclaration, NonCustomPropertyIterator};
 use crate::properties::longhands;
 use crate::properties::longhands::visibility::computed_value::T as Visibility;
 use crate::properties::LonghandId;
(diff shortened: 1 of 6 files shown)


mrobinson:
@emilio, if you have some free moments do you mind looking at this change? I'm not entirely certain what the right procedure is for making changes to the Gecko-specific code. I've verified that the changes build, but I don't think I have try access any longer.

components/style/properties/helpers/animated_properties.mako.rs
```diff
@@ -883,3 +883,69 @@ impl ToAnimatedZero for AnimatedFilter {
...
+            .as_mut()
+            .and_then(|iterator| iterator.next());
+        if let Some(longhand_id) = next_longhand {
+            return Some(TransitionPropertyIteration {
+                longhand_id,
+                index: self.index_range.start,
+            });
+        }
+
+        let index = match self.index_range.next() {
```
emilio, line 931: nit: You can use `.next()?`.

components/style/properties/helpers/animated_properties.mako.rs
```diff
@@ -883,3 +883,69 @@ impl ToAnimatedZero for AnimatedFilter {
...
+            Some(index) => index,
+            None => return None,
+        };
+
+        match self.style.get_box().transition_property_at(index) {
+            TransitionProperty::Custom(..) | TransitionProperty::Unsupported(..) => self.next(),
```
emilio, line 937: May be better to make this a `loop { }` instead of recursive. I don't think it should be much harder.

components/style/properties/properties.mako.rs
```diff
@@ -3003,6 +3003,14 @@ impl ComputedValues {
         % endfor
         set
     }
+
+    /// Create a `TransitionPropertyIterator` for this styles transition properties.
+    pub fn transition_properties<'a>(
+        &'a self
+    ) -> animated_properties::TransitionPropertyIterator<'a>
+    {
```
emilio, line 3011: `{` should go to the previous line I think.

emilio (commented):
Thanks, and sorry for the lag. This looks good with a few nits to me.

I can push it to try, let me know if you want to recover access, I think we're pretty liberal reenabling accounts :)

emilio (approved):
r=me with the nits fixed.

components/style/properties/properties.mako.rs
jdm: rustfmt probably disagrees, since CI is passing.

mrobinson: It seems like rustfmt might not be processing Mako templates? In any case, I've formatted this to look like the other methods in the file.

components/style/properties/helpers/animated_properties.mako.rs
mrobinson: Oh, nice. Fixed.

components/style/properties/helpers/animated_properties.mako.rs
mrobinson: Sounds good. I've changed this too a loop and added a comment explaining what is going on.

mrobinson:
> Thanks, and sorry for the lag. This looks good with a few nits to me.
> 
> I can push it to try, let me know if you want to recover access, I think we're pretty liberal reenabling accounts :)

Thanks for the review! I've fixed your nits and will ask bors to merge this. Do you mind pushing this one to try for me? I think once I make one or two more changes to the gecko code, I might need try access again. I'd hate to ask people to do a bunch of work and then never use it.

@bors-servo r=emilio

mrobinson:
@bors-servo retry
 * #21067

mrobinson:
It turns out that this does change behavior. A new side-effect of this change is that it prevents the creation of two transitions when using the `all` shorthand (for instance for `margin-block-end` and `margin-bottom`). This is a progression, so I think it makes sense to keep it. In a followup change, I plan to make this progression more explicit.

components/style/properties/properties.mako.rs
emilio: Yeah, we don't rustfmt this file.

components/style/properties/helpers/animated_properties.mako.rs
```diff
@@ -883,3 +883,68 @@ impl ToAnimatedZero for AnimatedFilter {
...
+                return Some(TransitionPropertyIteration {
+                    longhand_id,
+                    index: self.index_range.start,
+                });
+            }
+
+            let index = self.index_range.next()?;
+            match self.style.get_box().transition_property_at(index) {
+                TransitionProperty::Longhand(longhand_id) => {
+                    self.longhand_iterator = None;
```
emilio, line 934: Would maybe be a bit cleaner to do:

```
if let Some(ref mut it) = self.longhand_iterator {
    if let Some(longhand_id) = it.next() {
        return Some(...);
    }
    self.longhand_iterator = None;
}
```

I think that should work with nll but otherwise you can move the `= None` assignment after the `if`. That should allow you to remove this assignment and the custom / unsupported property one (and it's also nicer to clean up our state asap). Wdyt?

mrobinson: Yep. All this makes sense. I've pushed a new version of the branch that makes these changes.

mrobinson:
@bors-servo retry
 * #25233

jdm:
@bors-servo retry
* #25513

PR #26201, opened 2020-04-16
Refactor style rule tree, without actually fixing its !Send behaviour yet

nox:


components/style/rule_tree/map.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+impl<K, V> MallocShallowSizeOf for Map<K, V>
+where
+    K: Eq + Hash,
+{
+    fn shallow_size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
+        match &self.inner {
+            MapInner::Map(m) => m.shallow_size_of(ops) + (**m).shallow_size_of(ops),
```
emilio, line 160: Maybe keep the `// We want to account for both the box and the hashmap.` as it's not quite obvious why the double deref is needed.

emilio: Also, fwiw, I'd rather do `match self.inner` and `Map(ref m)` than having to mint it in my head, but your call :)

components/style/rule_tree/source.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        let _ = write!(writer, "  -> {:?}", self.read(guard).declarations());
+    }
+
+    // This is totally unsafe, should be removed when we figure out the cause of
+    // bug 1607553.
+    #[allow(unsafe_code)]
+    #[cfg(feature = "gecko")]
+    pub(super) unsafe fn dump_unchecked<W: Write>(&self, writer: &mut W) {
```
emilio: This can go away, see https://phabricator.services.mozilla.com/D63286.

components/style/rule_tree/core.rs
```diff
@@ -74,14 +74,14 @@ impl MallocSizeOf for RuleTree {
     fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
         let mut n = 0;
         let mut stack = SmallVec::<[_; 32]>::new();
-        stack.push(self.root.downgrade());
+        stack.push(self.root.clone());
 
         while let Some(node) = stack.pop() {
             n += unsafe { ops.malloc_size_of(node.ptr()) };
             let children = unsafe { (*node.ptr()).children.read() };
```
emilio: Change this to `node.get()` now?

emilio (commented):
I need to attend something in a few minutes, but wanted to post what I had.

Some of my comments you already addressed in followup commits.

This looks pretty sane, the thing I don't quite get is why `UnsafeBox` instead of just `NonNull<>` (you can implement `Deref` on `StrongRuleNode` if you wanted.

This makes it very very easy to misuse the `UnsafeBox` in the `WeakRuleNode` without bumping the reference count. You did split stuff to other submodules which makes it a bit less of a concern but...

components/style/rule_tree/core.rs
nox: `get` doesn't exist anymore.

nox:
I put back the comment in the `MallocSizeOfShallow` impl for `Map<K, V>`.

> This looks pretty sane, the thing I don't quite get is why UnsafeBox instead of just NonNull<> (you can implement Deref on StrongRuleNode if you wanted.

Using `UnsafeBox<T>` means I never need to `unsafe impl` the marker traits `Send` and `Sync`.

> This makes it very very easy to misuse the UnsafeBox in the WeakRuleNode without bumping the reference count. You did split stuff to other submodules which makes it a bit less of a concern but...

`WeakRuleNode` values are never supposed to be alive if the node they point to is dead, and `UnsafeBox::clone` is unsafe.

@bors-servo r=emilio

nox:
As per the prophecy, I fumbled the rebase.

nox:
Oh I see, should have checked release mode too.

That's the problem when one defines internal data structures directly in the same crate, any method that would be public in another setup ends up `pub(crate)`, so you need to be careful to actually use *all* of them lest you get those kind of unfortunate unused warnings.

nox:
@bors-servo r-

My PR adds back some bits that were removed on the Gecko side because I fumbled that part of the rebase because I move that code around.

nox:
Ok, I removed the assertion that checked whether the children list was empty, which was removed already from the tree prior to my PR.

@bors-servo r=emilio

CYBAI:
@bors-servo retry

- #25513

jdm:
@bors-servo retry
* #25513

PR #26205, opened 2020-04-16
Don't shutdown the app when it's suspended.

### LINKED ISSUE #155: Returning to home screen from immersive mode and reactivating app shows openxr missing frame animation

### LINKED ISSUE #26162: Navigation doesn't work after leaving immersive mode

jdm:
Fixes #26162. Fixes https://github.com/servo/webxr/issues/155.

### DIFF
--- a/support/hololens/ServoApp/App.cpp
+++ b/support/hololens/ServoApp/App.cpp
@@ -103,10 +103,14 @@ void App::OnActivated(IActivatedEventArgs const &args) {
 }
 
 void App::OnSuspending(IInspectable const &, SuspendingEventArgs const &) {
-  auto content = Window::Current().Content();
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo retry
* #23290 #25513

PR #26206, opened 2020-04-16
Add VCINSTALLDIR to the 'do not run within vcvarsall' check

Manishearth:
Servo is able to run with VSINSTALLDIR set, but not VCINSTALLDIR, since cc-rs takes it to mean vcvarsall has been called.

As far as I can tell servo is able to build with a custom VS install without needing help finding it (aside from perhaps VSINSTALLDIR), since many tools use a non-env var method of finding VS.

r? @jdm


jdm:
@bors-servo retry
* #25513 #24622 #23290

jdm:
@bors-servo retry
* #23290

PR #26215, opened 2020-04-17
Fix combination of border and border-radius being drawn incorrectly

### LINKED ISSUE #20922: Elements with a combination of 'border' and 'border-radius' are drawn incorrectly
Excuse my narcism here for posting my own profile picture, but I was checking out my own [website](https://martinrogalla.com) in servo. What I noticed is that my profile picture that is supposed to be round, get's drawn incorrectly:

Google Chrome:
chrome

Servo:
servo

When I try to create a 'circular' element with the following attributes:
```css
border: 3px solid #fff;
border-radius: 100px; /* note, border-radius >= 75px for the element do be 'circular' */
width: 150px;
height: 150px;
```

To me it looks like the image gets square borders and then is rounded, instead of the other way around. I'm interested in contributing to servo and would love to pick this issue up if you think it's doable.

dralley:
fixes #20922

Manual testcase in the attached issue

- [x] These changes fix #20922 

### DIFF
--- a/components/layout/display_list/builder.rs
+++ b/components/layout/display_list/builder.rs
@@ -1733,8 +1733,18 @@ impl Fragment {
                 build_border_radius_for_inner_rect(stacking_relative_border_box, &self.style);
 
             if !radii.is_zero() {
-                let clip_id =
-                    state.add_late_clip_node(stacking_relative_border_box.to_layout(), radii);
+                // This is already calculated inside of build_border_radius_for_inner_rect(), it would be
+                // nice if it were only calculated once.
+                let border_widths = self
+                    .style
+                    .logical_border_width()
+                    .to_physical(self.style.writing_mode);
+                let clip_id = state.add_late_clip_node(
+                    stacking_relative_border_box
+                        .inner_rect(border_widths)
+                        .to_layout(),
+                    radii,
+                );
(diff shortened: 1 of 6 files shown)


dralley:
I'm open to fixing this for Layout 2020, but I haven't figured out where the controlling code is (or if it even exists yet)

jdm:
Is there a meaningful automated test that we can add to this? Is there some way of achieving the same effect as border-radius without using border-radius?

jdm:
@SimonSapin See my previous comment.

SimonSapin:
Since this is about how borders are painted in the presence of border-radius, I suppose we could have a reftests that uses nested elements with a different border-radius and background-color?

Though like with anything other than axis-aligned rectangles we risk running into false negative due to anti-aliasing, and may need to make that reftest "fuzzy".

jdm:
The new test definitely fails in layout 2020 as well:
Screen Shot 2020-04-30 at 10 12 51 PM

It would be nice to get that fixed.

jdm:
@SimonSapin Any suggestions for where to start with applying the border-radius clipping to image fragments?

SimonSapin:
I’ve filed https://github.com/servo/servo/issues/26390 about this.

PR #26218, opened 2020-04-18

emilio:
See individual commits for details.

This fixes some rebase mistakes from #26201 and such.


components/selectors/parser.rs
```diff
@@ -226,6 +234,11 @@ pub trait Parser<'i> {
         false
     }
 
+    /// Whether to parse the `:where` pseudo-class.
```
CYBAI, line 237: Just curious if this comment should also indicate `:is` pseudo-class 👀 ?

PR #26220, opened 2020-04-18
Always upgrade existing weak child references in the rule tree

nox:
Just because we didn't find a child when read-locking a node children list
doesn't mean it still won't exist while we wait to upgrade the read lock
into a write lock to create the child.


jdm:
@bors-servo retry
* #25513 #24622

jdm:
@bors-servo retry
* #25513

PR #26227, opened 2020-04-20
Make the rule tree actually threadsafe

nox:
RuleTree::gc is now a safe method that any thread can call
at any time, and StrongRuleNode values can all be dropped
whenever their owner want to, on any thread.

### DIFF
--- a/components/style/rule_tree/core.rs
+++ b/components/style/rule_tree/core.rs
@@ -6,15 +6,15 @@
 
 use crate::properties::Importance;
 use crate::shared_lock::StylesheetGuards;
-use crate::thread_state;
 use malloc_size_of::{MallocShallowSizeOf, MallocSizeOf, MallocSizeOfOps};
 use parking_lot::RwLock;
 use smallvec::SmallVec;
 use std::fmt;
 use std::hash;
 use std::io::Write;
+use std::mem;
 use std::ptr;
-use std::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
+use std::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 
 use super::map::{Entry, Map};
 use super::unsafe_box::UnsafeBox;
@@ -31,16 +31,10 @@ use super::{CascadeLevel, StyleSource};
 /// them.
 ///
 /// When the rule node refcount drops to zero, it doesn't get freed. It gets
-/// instead put into a free list, and it is potentially GC'd after a while in a
-/// single-threaded fashion.
+/// instead put into a free list, and it is potentially GC'd after a while.
 ///
 /// That way, a rule node that represents a likely-to-match-again rule (like a
 /// :hover rule) can be reused if we haven't GC'd it yet.
-///
-/// See the discussion at https://github.com/servo/servo/pull/15562 and the IRC
-/// logs at http://logs.glob.uno/?c=mozilla%23servo&s=3+Apr+2017&e=3+Apr+2017
-/// logs from http://logs.glob.uno/?c=mozilla%23servo&s=3+Apr+2017&e=3+Apr+2017#c644094
-/// to se a discussion about the different memory orderings used here.
 #[derive(Debug)]
 pub struct RuleTree {
     root: StrongRuleNode,
@@ -48,24 +42,7 @@ pub struct RuleTree {
 
 impl Drop for RuleTree {
     fn drop(&mut self) {
-        // GC the rule tree.
-        unsafe {
-            self.gc();
-        }
-
-        // After the GC, the free list should be empty.
-        debug_assert_eq!(
-            self.root.p.next_free.load(Ordering::Relaxed),
-            FREE_LIST_SENTINEL
-        );
-
-        // Remove the sentinel. This indicates that GCs will no longer occur.
-        // Any further drops of StrongRuleNodes must occur on the main thread,
-        // and will trigger synchronous dropping of the Rule nodes.
-        self.root
-            .p
-            .next_free
-            .store(ptr::null_mut(), Ordering::Relaxed);
+        unsafe { self.swap_free_list_and_gc(ptr::null_mut()) }
     }
 }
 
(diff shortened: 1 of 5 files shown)


components/style/rule_tree/core.rs
```diff
@@ -615,146 +661,61 @@ impl Drop for StrongRuleNode {
...
+    /// Must be called with items stored in a node's children list. The children
+    /// list must at least be read-locked when this is called.
+    unsafe fn upgrade(&self) -> StrongRuleNode {
         debug!("Upgrading weak node: {:p}", &*self.p);
-        self.p.refcount.fetch_add(1, Ordering::Relaxed);
-        unsafe { StrongRuleNode::from_unsafe_box(UnsafeBox::clone(&self.p)) }
+
+        if self.p.refcount.fetch_add(1, Ordering::Relaxed) == 0 {
+            atomic::fence(Ordering::Acquire);
+            while self.p.next_free.load(Ordering::Relaxed).is_null() {}
```
emilio, line 716: This deserves a comment saying that we busy-wait here, but that this only happens when we're resurrecting in the middle of a thread which is gc'ing. And probably also a comment on the swap() call on the free list.

components/style/rule_tree/core.rs
```diff
@@ -119,16 +84,18 @@ impl RuleTree {
...
     }
 
     /// This can only be called when no other threads is accessing this tree.
-    pub unsafe fn maybe_gc(&self) {
+    pub fn maybe_gc(&self) {
         #[cfg(debug_assertions)]
         self.maybe_dump_stats();
 
-        self.root.maybe_gc();
+        if self.root.p.free_count.load(Ordering::Relaxed) > RULE_TREE_GC_INTERVAL {
```
emilio, line 96: What increments the free count now? Can this ever trigger?

components/style/rule_tree/core.rs
nox: Added comments.

components/style/rule_tree/core.rs
nox: Added the incrementing back, hah.

nox:
I think it would indeed make more sense to never decrement `free_count` and to only swap it with zero, what do you think @emilio? I'm afraid the decrement in `swap_free_list_and_gc` can underflow.

components/style/rule_tree/core.rs
```diff
@@ -306,34 +358,156 @@ impl RuleNode {
...
+                // thread will have any more opportunity to resurrect the node
+                // anymore.
+                let mut children = parent.p.children.write();
+
+                this.next_free.store(ptr::null_mut(), Ordering::Relaxed);
+
+                // We decrement the counter to remove the "pretend to be
+                // on the free list" reference.
+                let old_refcount = this.refcount.fetch_sub(1, Ordering::Release);
+                debug_assert!(old_refcount != 0);
```
emilio, line 392: nit: maybe `debug_assert_ne!`?

components/style/rule_tree/core.rs
```diff
@@ -306,34 +358,156 @@ impl RuleNode {
...
+            // on success.
+            match root.next_free.compare_exchange_weak(
+                head,
+                this.as_mut_ptr(),
+                Ordering::Release,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => {
+                    // Increment the free count by one.
+                    root.free_count.fetch_add(1, Ordering::Relaxed);
```
emilio, line 483: We should probably rename `.free_count` to `.estimated_free_count` or `.approximate_free_count` something, as it can't really be relied on for other than heuristics.

components/style/rule_tree/core.rs
```diff
@@ -307,34 +358,156 @@ impl RuleNode {
...
+                debug_assert_eq!(this.refcount.load(Ordering::Relaxed), 0);
+            }
 
-    fn free_count(&self) -> &AtomicUsize {
-        debug_assert!(self.is_root());
-        &self.free_count
+            // We are going to drop this node for good this time, as per the
+            // usual refcounting protocol we need an acquire fence here before
+            // we run the destructor.
+            atomic::fence(Ordering::Acquire);
```
emilio, line 417: I think this could be `this.refcount.load(Ordering::Acquire)`, which is potentially cheaper. See the comment in servo_arc/lib.rs

components/style/rule_tree/core.rs
```diff
@@ -615,146 +675,67 @@ impl Drop for StrongRuleNode {
...
-        unsafe { StrongRuleNode::from_unsafe_box(UnsafeBox::clone(&self.p)) }
+
+        if self.p.refcount.fetch_add(1, Ordering::Relaxed) == 0 {
+            // We observed a refcount of 0, we need to wait for this node to
+            // be put on the free list. Resetting the `next_free` pointer to
+            // null is only done in `RuleNode::drop_without_free_list`, just
+            // before a release refcount decrement, so this acquire fence here
+            // makes sure that we observed the write to null before we loop
+            // until there is a non-null value.
+            atomic::fence(Ordering::Acquire);
```
emilio, line 735: Same here about `atomic::fence`.

components/style/rule_tree/core.rs
emilio: Ah, actually you made it so it's not resetted to zero but decremented, so this is alright. Thanks!

emilio (approved):
This looks good to me, thanks!

I think the atomic fences could be a bit more limited, and just be loads. Other than that and a minor nit this looks good.

components/style/rule_tree/core.rs
nox: It is useless, if they compare equal it will just show `0` twice, not giving more details to the user.

components/style/rule_tree/core.rs
nox: AFAIK it doesn't matter. https://github.com/rust-lang/rust/pull/41714#issuecomment-298996916

components/style/rule_tree/core.rs
nox: I changed it to `approximate_free_count` and I made `swap_free_list_and_gc` just swap it to 0, because what I did initially could make it underflow.

emilio: Yeah, you're right. For posterity the bad case is: you insert into the free-list successfully, and another thread steals it and completes a gc before you get a chance to increment the count. Then the other thread would decrement the count before you get the chance to increment it, potentially underflowing.

emilio:
err,

@bors-servo r+

Issue #25258, opened 2019-12-12
Intermittent timeout in CSS paint api tests (called `Option::unwrap()` on a `None` value)


jdm:
```
  â–¶ TIMEOUT [expected PASS] /css/css-paint-api/parse-input-arguments-001.https.html
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ 
  â”‚ Shutting down the Constellation after generating an output file or exit flag specified
  â”‚ called `Option::unwrap()` on a `None` value (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at /rustc/6d77e45f01079fe3d40180b3e256e414ab379f63/src/libcore/macros/mod.rs:15)
  â”‚ stack backtrace:
  â”‚    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
  â”‚    1: core::fmt::write
  â”‚    2: <std::io::stdio::Stdout as std::io::Write>::write_fmt
  â”‚    3: std::io::stdio::_print
  â”‚    4: servo::backtrace::print
  â”‚    5: servo::main::{{closure}}
  â”‚    6: std::panicking::rust_panic_with_hook
  â”‚    7: rust_begin_unwind
  â”‚    8: core::panicking::panic_fmt
  â”‚    9: core::panicking::panic
  â”‚   10: std::thread::local::LocalKey<T>::with
  â”‚   11: <script::dom::worklet::Worklet as core::ops::drop::Drop>::drop
  â”‚   12: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  â”‚   13: __rust_maybe_catch_panic
  â”‚   14: script::dom::bindings::codegen::Bindings::WorkletBinding::WorkletBinding::_finalize
  â”‚   15: _ZN2js2gc5Arena8finalizeI8JSObjectEEmPNS_6FreeOpENS0_9AllocKindEm
  â”‚   16: _ZL19FinalizeTypedArenasI8JSObjectEbPN2js6FreeOpEPPNS1_2gc5ArenaERNS4_15SortedArenaListENS4_9AllocKindERNS1_11SliceBudgetENS4_10ArenaLists14KeepArenasEnumE
  â”‚   17: _ZN2js2gc10ArenaLists18foregroundFinalizeEPNS_6FreeOpENS0_9AllocKindERNS_11SliceBudgetERNS0_15SortedArenaListE
[...]

PR #26229, opened 2020-04-20
Use python binary to run multiprocess tests.

jdm:
The python2 binary doesn't exist in the path on macOS. This change uses an existing python2 on macOS and whatever the default python version is on linux.

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -614,7 +614,7 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
         # https://github.com/servo/servo/issues/22438
         if this_chunk == 0:
             task.with_script("""
-                time python2 ./mach test-wpt --release --binary-arg=--multiprocess \
+                time python ./mach test-wpt --release --binary-arg=--multiprocess \
                     --processes $PROCESSES \
                     --log-raw test-wpt-mp.log \
                     --log-errorsummary wpt-mp-errorsummary.log \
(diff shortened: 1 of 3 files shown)


SimonSapin:
LGTM, but if this `python2` fails now how did it work before?

jdm:
It never worked. I think possibly the `| cat` on the end of the line suppresses the error so CI doesn't report a failure.

SimonSapin:
Please also add `-o pipefail` (before `-c`) at these two locations, in order to catch such failures:
https://github.com/servo/servo/blob/5f5643e7aa1b532917794cb739f8973539b45fc8/etc/taskcluster/decisionlib.py#L752
https://github.com/servo/servo/blob/5f5643e7aa1b532917794cb739f8973539b45fc8/etc/taskcluster/decisionlib.py#L688

etc/taskcluster/decision_task.py
```diff
@@ -614,7 +614,7 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
         # https://github.com/servo/servo/issues/22438
         if this_chunk == 0:
             task.with_script("""
-                time python2 ./mach test-wpt --release --binary-arg=--multiprocess \
+                time python ./mach test-wpt --release --binary-arg=--multiprocess \
```
SimonSapin, line 617: `time ./mach` (relying on shebang, like in other uses of `mach` commands) should also work, but this is fine.

SimonSapin (approved):
r+ assuming try looks good

jdm:
@bors-servo retry
* #26233 #25513

jdm:
@bors-servo retry
* #25258

Issue #25533, opened 2020-01-15
Intermittent crash during shutdown in reftests


jdm:
```
  ▶ CRASH [expected PASS] /css/CSS2/css1/c5513-ibrdr-bw-000.xht
  │ 
  │ 
  │ 
  │ 
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ 
  │ 
  │ 
  └ Shutting down the Constellation after generating an output file or exit flag specified
```

Observed on Linux.

pshaughn:
Is this always .xht?

jdm:
Unclear, but I doubt it?

asajeffrey:
No, I got this on a `.html` test (https://community-tc.services.mozilla.com/tasks/AceEISiVQ2ek7JCoEIIWEQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FAceEISiVQ2ek7JCoEIIWEQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Ffiltered-wpt-errorsummary.log)
```
0 matches
1 unexpected results that are NOT known-intermittents:
  ▶ CRASH [expected PASS] /css/css-values/vh-support-transform-translate.html
  │ 
  │ 
  │ 
  │ 
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ Stack trace for thread "LayoutThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }"
  └ stack backtrace:

```

jdm:
I suspect that segfaults from the layout thread right now are #26129.

TimvdLippe:
Closing as stale as this issue has not received any failure references for over a month

Issue #25065, opened 2019-12-03
When an clicked HTMLAnchorElement has both href and onclick, WPT and other browsers see the href, but we see the onclick


pshaughn:
This is failing in html/browsers/browsing-the-web/navigating-across-documents/006.html and 007.html (also tested by 005.html but that's failing a different way instead).

pshaughn:
The anchor element's activation behavior is firing, and follow_hyperlink goes all the way through to its debug! statement. If I'm reading the spec correctly, which I might not be, the sequence of events in 006 should be:
* onclick fires a form submit pointing at target
* anchor element activation behavior follows a hyperlink pointing at target within the same task, aborting the onclick one
* anchor element's fetch succeeds, and the page it loads fires a message event

However, the log output looks like:
* onclick starts its form submit
* anchor element activation behavior gets through follow_hyperlink and calls load_url (running single-process so #25352 shouldn't apply, AFAIK)
* the onclick element's fetch goes out to the network
* the onclick element's fetch succeeds, and the page it loads fires a message event
and the anchor element's call to load_url never even sends a net request.

pshaughn:
It still fails to activate the href even without the onclick! #25374 is a much more minimal example of an href in an iframe failing.

PR #26231, opened 2020-04-20
Implement various checks on the Location setters

### LINKED ISSUE #23522: Implement various checks on the Location setters

utsavoza:
Since there are no existing open PRs for the issue, I am raising one based on the work done by @UxioAndrade and @braddunbar in previous two PRs for the issue (#25714 and #23670) respectively.

- [x] These changes fix #23522 


components/script/dom/location.rs
```diff
@@ -210,8 +261,19 @@ impl LocationMethods for Location {
...
-        self.set_url_component(value, UrlHelper::SetPathname);
+        // Step 1: If this Location object's relevant Document is null, then return.
+        if self.window.has_document() {
+            // Step 2: If this Location object's relevant Document's origin is not
+            // same origin-domain with the entry settings object's origin, then
+            // throw a "SecurityError" DOMException.
+            self.check_same_origin_domain()?;
+            // Step 3: Let copyURL be a copy of this Location object's url.
+            let copy_url = self.get_url();
+            // Step 4: If copyURL's cannot-be-a-base-URL flag is set, terminate these steps.
```
gterzian, line 272: Are we doing step 5 currently? If not, I think we can do it using `as_mut_url().set_path`.

components/script/dom/location.rs
```diff
@@ -223,8 +285,23 @@ impl LocationMethods for Location {
...
+        // Step 1: If this Location object's relevant Document is null, then return.
+        if self.window.has_document() {
+            // Step 2: If this Location object's relevant Document's origin is not
+            // same origin-domain with the entry settings object's origin, then
+            // throw a "SecurityError" DOMException.
+            self.check_same_origin_domain()?;
+            // Step 3: Let copyURL be a copy of this Location object's url.
+            let copy_url = self.get_url();
+            // Step 4: If copyURL cannot have a username/password/port, then return.
+            // https://url.spec.whatwg.org/#cannot-have-a-username-password-port
```
gterzian, line 297: Similar re step 5, which could be done using `set_port`.

components/script/dom/location.rs
```diff
@@ -249,8 +352,14 @@ impl LocationMethods for Location {
...
     fn SetSearch(&self, value: USVString) -> ErrorResult {
-        self.check_same_origin_domain()?;
-        self.set_url_component(value, UrlHelper::SetSearch);
+        // Step 1: If this Location object's relevant Document is null, then return.
+        if self.window.has_document() {
+            // Step 2: If this Location object's relevant Document's origin is not
+            // same origin-domain with the entry settings object's origin, then
+            // throw a "SecurityError" DOMException.
+            self.check_same_origin_domain()?;
+            self.set_url_component(value, UrlHelper::SetSearch);
```
gterzian, line 361: I think we're missing step 4 and 5 here?

components/script/dom/location.rs
```diff
@@ -178,8 +212,19 @@ impl LocationMethods for Location {
...
+        if self.window.has_document() {
+            // Step 2: If this Location object's relevant Document's origin is not
+            // same origin-domain with the entry settings object's origin, then
+            // throw a "SecurityError" DOMException.
+            self.check_same_origin_domain()?;
+            // Step 3: Let copyURL be a copy of this Location object's url.
+            let copy_url = self.get_url();
+            // Step 4: If copyURL's cannot-be-a-base-URL flag is set, terminate these steps.
+            if !copy_url.cannot_be_a_base() {
+                self.set_url_component(value, UrlHelper::SetHostname);
```
gterzian, line 225: How about we add a `copy_url` argument to `set_url_component` to avoid all those double calls to `get_url`(for algorithms that need the url before calling `set_url_component`)?

gterzian (changes requested):
Great work! Thanks for figuring out all those details. I think there are a few steps missing(although you could open an issue and leave that as a follow-up if there is a reason for that), and I have a small suggestion re the `set_url_component` method. 

Also, could you please squash into one commit? I think while it's a good idea to separate unrelated work into commits, in this case it is actually related and  consolidating the work into a commit would appear clearer.

components/script/dom/location.rs
utsavoza: Oh, I assumed we were executing those steps as a part of `url::quirks::set_path` through `UrlHelper::SetPathname`, but those steps seem to correspond to dom url spec.

components/script/dom/location.rs
utsavoza: Again, I presumed those steps as a part of `UrlHelper::SetSearch`.

components/script/dom/location.rs
utsavoza: There's one caveat though when setting port directly via `as_mut_url().set_port` - we'll have to manually verify and parse the input string to `Option<u16>`. However, we can use `url::quirks::set_port` to check and set the port here instead of `UrlHelper::SetPort`.

components/script/dom/location.rs
utsavoza: I think we can actually remove the method completely - the only place we might actually need this method after adding checks would be for `SetPort` method probably (as we might not directly setting port to the url there). For the rest of the setters, we can directly call the navigate method with the appropriate arguments.

utsavoza:
> Also, could you please squash into one commit? I think while it's a good idea to separate unrelated work into commits, in this case it is actually related and consolidating the work into a commit would appear clearer.

Agreed. I'll squash into one commit once the changes are approved.

components/script/dom/location.rs
gterzian: Sounds good!

gterzian (changes requested):
Looks like a big improvement in the correctness of the implementation! Just one nit left, and then after you squash I'll approve it(after which we might still have to deal with some surprise failures on the test run).

utsavoza:
> Looks like a big improvement in the correctness of the implementation! Just one nit left, and then after you squash I'll approve it(after which we might still have to deal with some surprise failures on the test run).

Actually, I ran the wpt tests for the location interface locally. Apart from that how do I verify if we aren't breaking anything else?

gterzian:
> Actually, I ran the wpt tests for the location interface locally. Apart from that how do I verify if we aren't breaking anything else?

We have to basically run the whole suite and see, there can also be tests that unexpectedly start passing...

gterzian:
Thanks for the awesome work!

@bors-servo r+

gterzian:
This one seems relevant:

```
  ▶ Unexpected subtest result in /html/browsers/browsing-the-web/navigating-across-documents/source/navigate-child-function.html:
  │ FAIL [expected PASS] Set location from a function called from a parent
  │   → assert_equals: expected "http://web-platform.test:8000/html/browsers/browsing-the-web/navigating-across-documents/source/support/location-set.html" but got "http://web-platform.test:8000/html/browsers/browsing-the-web/navigating-across-documents/source/navigate-child-function.html"
  │ 
  │ onload</fr.onload<@http://web-platform.test:8000/html/browsers/browsing-the-web/navigating-across-documents/source/navigate-child-function.html:15:20
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2001:25
  └ Test.prototype.step_func_done/<@http://web-platform.test:8000/resources/testharness.js:2042:32
```

This one does not seem relevant(I'm not sure what issue it would match by the way):

```
  ▶ Unexpected subtest result in /_mozilla/mozilla/scroll_top_null_target.html:
  │ FAIL [expected PASS] scroll_top_null_target
  │   → assert_equals: Target should be null! expected null but got Element node <a id="test"></a>
  │ 
  │ @http://web-platform.test:8000/_mozilla/mozilla/scroll_top_null_target.html:18:23
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2001:25
  └ @http://web-platform.test:8000/_mozilla/mozilla/scroll_top_null_target.html:13:10
```

gterzian:
Ok the first one might actually a a new intermittent, perhaps you fixed something that makes it show up, and since it seems related to naviation, it's probably best done in a follow up. So please take a look, and if it doesn't appear to point out an issue with this PR, just file an issue for it.

gterzian:
this one seems defenitely relevant:

```
  ▶ Unexpected subtest result in /url/failure.html:
  │ FAIL [expected PASS] Location's href: http:/:@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:/@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:a:b@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:/a:b@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http::@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:@:www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: http:/@:www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: a should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: a/ should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ FAIL [expected PASS] Location's href: a// should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
  │ 
  │ runTests/<@http://web-platform.test:8000/url/failure.html:41:23
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2001:25
  └ runTests@http://web-platform.test:8000/url/failure.html:40:10
```

components/script/dom/location.rs
```diff
@@ -191,14 +248,20 @@ impl LocationMethods for Location {
...
+            // Note: no call to self.check_same_origin_domain()
+            // Step 2: Parse the given value relative to the entry settings object.
+            // If that failed, throw a TypeError exception.
+            let base_url = self.entry_document().url();
+            let url = match base_url.join(&value.0) {
+                Ok(url) => url,
+                Err(e) => return Err(Error::Type(format!("Couldn't parse URL: {}", e))),
+            };
+            // Step 3: Location-object-setter navigate to the resulting URL record.
+            let referrer = Referrer::ReferrerUrl(base_url.clone());
```
gterzian, line 262: Ok I think this is the problem with `/html/browsers/browsing-the-web/navigating-across-documents/source/navigate-child-function.html`, the referrer should be not from the entry document but from `self.window.Document`, at least that is what I make from the test failure(so it's not a new intermittent and not actually related much to navigation)

utsavoza:
I was able to find six relevant test failures:
1. `/url/failure.html`
2. `/html/browsers/browsing-the-web/navigating-across-documents/005.html`
3. `/html/browsers/browsing-the-web/navigating-across-documents/source/navigate-child-function.html`
4. `/html/browsers/browsing-the-web/navigating-across-documents/empty_fragment.html`
5. `/html/browsers/browsing-the-web/unloading-documents/beforeunload-synchronous.html` - `[UNEXPECTED PASS]`
6. `/fullscreen/api/document-exit-fullscreen-active-document.html` - `[UNEXPECTED PASS]`

and one NOT known-intermittent (as pointed out by @gterzian) from the logs:

```
./mach filter-intermittents wpt-jsonsummary.log --log-intermittents intermittents.log --log-filteredsummary filtered-wpt-errorsummary.log --tracker-api default --reporter-api default
  ▶ Unexpected subtest result in /_mozilla/mozilla/scroll_top_null_target.html:
  │ FAIL [expected PASS] scroll_top_null_target
  │   → assert_equals: Target should be null! expected null but got Element node <a id="test"></a>
  │
  │ @http://web-platform.test:8000/_mozilla/mozilla/scroll_top_null_target.html:18:23
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2001:25
  └ @http://web-platform.test:8000/_mozilla/mozilla/scroll_top_null_target.html:13:10
```

I will try to resolve these and post any updates here.

components/script/dom/location.rs
utsavoza: Ahh, right. The referrer url shouldn't be from entry document here - I think `let referrer = Referrer::ReferrerUrl(self.get_url());` should fix this.

jdm:
Filed #26269 for the new unrelated intermittent test result.

gterzian:
@utsavoza awesome, sounds like it's worth updating this PR with your work and we'll give the test suite another try...

utsavoza:
Hey @jdm, do we still require setting location.hash to empty string? I see that we did this earlier as a part of #11934 and whatwg/html#1318 to make this compliant with the spec. I found this a bit confusing since the spec does not list this step explicitly. 

In order to accommodate this change I believe the setter for hash would look something like this:
```rust
copy_url.as_mut_url().set_fragment(match value.0.as_str() {
    "" => Some("#"), // <- special case
    _ if value.0.starts_with('#') => Some(&value.0[1..]),
    _ => Some(&value.0),
});
```

jdm:
@utsavoza The code in location.rs matches the specification (as far as I can tell), but it's not written in a way where that is clearly the case. The specification is written so that using `location.hash = ""` in JS will cause a URL parse to occur against the current location URL (but with an empty fragment value) and the new fragment value, so the resulting URL will have no fragment if an empty string was passed. We need to be sure that we preserve that behaviour.

utsavoza:
So, I was able to fix the above mentioned failing wpt tests except for the first two.
>     1. `/url/failure.html`
The reason few tests are failing here is because in the updated implementation of `SetHref` we are now [parsing the given input value relative to the entry settings object](https://html.spec.whatwg.org/multipage/history.html#dom-location-href) as mentioned in the spec.

For example, for the given test:
```
self.test(() => {
  assert_throws_js(self[0].TypeError, () => self[0].location = test.input)
}, "Location's href: " + name)
```
The test input that is provided for example is:
```
  {
    "input": "http:/:@/www.example.com",
    "base": "about:blank",
    "failure": true
  }
```
but since we are now parsing the given input relative to entry settings object, the base url that gets picked up here is different (`http://web-platform.test:8000/url/failure.html`) from what is provided, which is now causing the test case to fail with:
```
▶ Unexpected subtest result in /url/failure.html:
  │ FAIL [expected PASS] Location's href: http:/:@/www.example.com should throw
  │   → assert_throws_js: function "() => self[0].location = test.input" did not throw
```
I am not really sure if updating the metadata for these test cases would be the correct way to go. Let me know what you think @gterzian?

<br>

>     2. `/html/browsers/browsing-the-web/navigating-across-documents/005.html`
For the second test, it seems we already have an open issue (#25065) that delineates the cause for this test failure. I have addressed this by updating the metadata for the test.

gterzian:
> but since we are now parsing the given input relative to entry settings object, the base url that gets picked up here is different

I'm wondering if the problem is instead that somehow the call to `base_url.join(&value.0)` doesn't return an error, in the case that `base_url` is "http://web-platform.test:8000/url/failure.html" and `value.0` is "http:/:@/www.example.com"?

utsavoza:
> If you were to `println!` the result of [this match](https://github.com/servo/servo/pull/26231/files#diff-45df5e6ab4c1d79fae956506be147b62R257), what do you see?

I did try to verify if this match would return an error or not. Apparently this code works:
```rust
let base_url = Url::parse("http://web-platform.test:8000/url/failure.html").unwrap();
match base_url.join("http:/:@/www.example.com") {
    Ok(url) => println!("url: {}", url), // -> prints "url: http://web-platform.test:8000/:@/www.example.com"
    Err(e) => eprintln!("error: {}", e)
};
```

gterzian:
Ok so do you mean that the test expects a parsing error, while the code currently just parses the url? It might be a quirk of the underlying `url` library, I don't know. 

I would say if it cannot be debugged successfully now it might be worth updating the metadata and filing a follow-up issue, since we're overall making progress on the test suite in this PR.

utsavoza:
> Ok so do you mean that the test expects a parsing error, while the code currently just parses the url? It might be a quirk of the underlying `url` library, I don't know.

Yes, ideally it should throw a `TypeError` but our implementation is able to parse the given input relative to base url. I am not sure though if this is a quirk of the underlying `url` lib. 

If I understand this correctly, the test case expects the href setter to fail when when trying to set the given input against the `about:blank` base url.

So I think it's the base url that gets picked up incorrectly because if the base url is `about:blank` our implementation would throw a `TypeError` as expected.

gterzian:
Ok, I think this requires some additional research and is best done in a follow-up(also there are a bunch of test failures in that test already, and they could actually be all investigated in a follow-up)I have filed this for a follow up https://github.com/servo/servo/issues/26287

One last thing, have you checked if the behavior described at https://github.com/servo/servo/pull/26231#issuecomment-617965900 is preserved with these changes?

utsavoza:
@gterzian hmm, I think I accidentally deleted my branch in my fork which caused this PR to be closed.

utsavoza:
> One last thing, have you checked if the behavior described at [#26231 (comment)](https://github.com/servo/servo/pull/26231#issuecomment-617965900) is preserved with these changes?

Yes, I had updated the PR to reflect that change.

jdm:
If there's no clear way to push the branch in a way that allows reopening the PR, go ahead and open a new one.

utsavoza:
Thanks, re-opened the PR.

utsavoza:
@gterzian, I think I have pushed all the changes that address the above listed failing wpt tests (and squashed the commits into one). Please have a look and let me know if the changes are good to go.

gterzian:
Looks good, thanks! Just a quick rebase and it should be ready to go...

gterzian (approved):

utsavoza:
Done!

gterzian:
@utsavoza Thanks for completing this, it was a lot of additional work in the end!

@bors-servo r+

gterzian:
@bors-servo retry

https://github.com/servo/servo/issues/21067
https://github.com/servo/servo/issues/26298
https://github.com/servo/servo/issues/25513
https://github.com/servo/servo/issues/26299

Plus in mac:

```
  ▶ TIMEOUT [expected OK] /_webgl/conformance2/sync/sync-webgl-specific.html
  │ 
  └ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  ▶ Unexpected subtest result in /_webgl/conformance2/sync/sync-webgl-specific.html:
  └ NOTRUN [expected PASS] Overall test
```

not sure which one that is

CYBAI:
@bors-servo retry

- #25533, #25513

utsavoza:
> @utsavoza Thanks for completing this, it was a lot of additional work in the end!

Hey @gterzian, thanks for your guidance.

Issue #26192, opened 2020-04-15
Intermittent segfault in servo_config/net/profile unit tests


jdm:
```
     Running target/debug/deps/servo_config-58794619d9a9dcda
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
     Running target/debug/deps/opts-1dd24dda002febdf
running 5 tests
test test_file_path_parsing ... ok
test test_argument_parsing_special ... ok
test test_argument_parsing ... ok
test test_parse_pref_from_command_line ... ok
error: test failed, to rerun pass '-p servo_config --test opts'
Caused by:
  process didn't exit successfully: `/Users/worker/tasks/task_1586964343/repo/target/debug/deps/opts-1dd24dda002febdf` (signal: 11, SIGSEGV: invalid memory reference)
```

jdm:
```

     Running target/debug/deps/main-1d400c975982a6d8

running 425 tests
test cookie::test_cookie_secure_prefix ... ok
test cookie::test_default_path ... ok
test cookie::fn_cookie_constructor ... ok
test cookie::test_cookie_host_prefix ... ok
test cookie::test_path_match ... ok
test cookie::test_domain_match ... ok
test cookie::test_cookie_eviction_expired ... ok
test cookie::test_secure_cookies_eviction_non_http_source ... ok
test cookie::test_secure_cookies_eviction ... ok
test cookie::test_insecure_cookies_cannot_evict_secure_cookie ... ok
test cookie_http_state::test_0001 ... ok
test cookie_http_state::test_0002 ... ok
test cookie_http_state::test_0004 ... ok
test cookie::test_sort_order ... ok
test cookie_http_state::test_0007 ... ok
test cookie_http_state::test_0008 ... ok
test cookie_http_state::test_0005 ... ok
test cookie_http_state::test_0009 ... ok
test cookie_http_state::test_0010 ... ok
test cookie_http_state::test_0011 ... ok
test cookie_http_state::test_0012 ... ok
test cookie_http_state::test_0013 ... ok
test cookie::test_cookie_eviction_all_secure_new_nonsecure ... ok
test cookie::test_cookie_eviction_all_secure_one_nonsecure ... ok
test cookie_http_state::test_0017 ... ok
test cookie_http_state::test_0018 ... ok
test cookie_http_state::test_0019 ... ok
test cookie_http_state::test_0014 ... ok
test cookie_http_state::test_0015 ... ok
[...]

PR #26238, opened 2020-04-20
Only enable RUST_BACKTRACE in non-mac build environment on CI.

jdm:
This should help work around #26192 and would explain why we're still seeing it on CI after #26217.


jdm:
@bors-servo retry
* mac webgl timeout

CYBAI:
@bors-servo retry

- #25513, #23290

CYBAI:
@bors-servo retry

- #25513

Issue #19422, opened 2017-11-29
style/gfx/metrics/script crates serialize build


jdm:
script depends on metrics, and metrics depends on gfx, and gfx depends on style (script also depends on style). If we move the bits of style that gfx depends upon into style_traits, and possibly the bits of gfx that metrics depends upon into gfx_traits, this should allow more crates to build in parallel.

tigercosmos:
The one need to move it one by one or finish it in one time?

jdm:
Moving one by one shouldn't hurt anything.

tigercosmos:
OK! I will work on this.

tigercosmos:
@jdm if some codes in `style` that `gfx` depends is also used by other crate(i.g. `layout`), shall I move it? Or just move codes only depended by `gfx`?

tigercosmos:
ummm... I think they should all be moved, otherwise `gfx` still need to wait for `style`

jdm:
Exactly. If it's moved from gfx to gfx_traits, the other crates that depend on the move items should be updated to use gfx_traits.

tigercosmos:
This issue is much harder than I though before. Actually almost everything depend on each other.

For example, If I want to move `gfx::display_list`, which is used by `metrics` and also the only one `gfx` dependency in `metrics`, since `gfx::display_list` use `gfx::text`, I will also need to move `gfx::text`, and `gfx::text` use lot of `gfx` things, which mean if I want to move `display_list`, I will move almost everything in `gfx`.

So, in this case, we can make sure move more something from `gfx` to `gfx_traits` is impossible.
In fact, I found `style` has the same problem. :(

jdm:
Try creating a trait that can replace https://dxr.mozilla.org/servo/source/components/metrics/lib.rs#306-315. That trait can live in gfx_traits, and the implementation in gfx.

tigercosmos:
I don't understand. If we move that code to `gfx_traits`.
The relationship would be `metrics` -> `gfx_traits` -> `gfx`.
Do I misunderstand?

tigercosmos:
I don't understand. If we move that code to `gfx_traits`.
The relationship would be `metrics` -> `gfx_traits` -> `gfx`.
I thought crate_traits will be independent with crate. Do I misunderstand?

jdm:
No, because metrics would depend on gfx_traits, and gfx would depend on gfx_traits.

tigercosmos:
[...]

PR #26241, opened 2020-04-21
Add a minimal alternative to `cargo graph`

SimonSapin:
I tried `cargo graph` and some of its successors, but didn’t manage to make them produce what I wanted (or in some cases make them work at all.)

This Python script reimplements similar functionality based on parsing the (JSON) output of `cargo metadata`.

Graphviz graphs can become hard to read very quickly as the number of nodes grows. Servo’s dependency graph is very large, so pruning as much as possible is important. This only shows `path` dependencies (that have their source in this repo), and can take a parameter to only show recursive dependencies of a given crate.

See https://github.com/servo/servo/issues/19422#issuecomment-617038366 for an example.

I find that `xdot` is best for visualization since it is interactive.

This script is not used by anything. I am making this PR only so that we have it somewhere in case it becomes useful again at some point.

### DIFF
--- a/etc/crates-graph.py
+++ b/etc/crates-graph.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+# Copyright 2013 The Servo Project Developers. See the COPYRIGHT
+# file at the top-level directory of this distribution.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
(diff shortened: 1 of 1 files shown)


SimonSapin:
79844191-a2b33d00-83bb-11ea-9023-c392c5a35c87

jdm:
@bors-servo retry
* #24622

PR #26276, opened 2020-04-23

emilio:
See individual commits for details, though this is a very small sync.


emilio:
@bors-servo r+

 * rustfmt

PR #26285, opened 2020-04-23
Add ElementAnimationState and PossibleElementAnimationState

mrobinson:
This refactor is preparation for implementing a specification
compliant transitions and animations processing model.

These data structures hold all the animation information about a single
node. Since adding, updating, and modifying animations for a single node
are all interdependent, it makes sense to start encapsulating animation
data and functionality into a single data structure. This also opens up
the possibility for easier concurrency in the future by more easily
allowing per-node mutexes.

---

### DIFF
--- a/components/style/animation.rs
+++ b/components/style/animation.rs
@@ -5,8 +5,7 @@
 //! CSS transitions and animations.
 
 // NOTE(emilio): This code isn't really executed in Gecko, but we don't want to
-// compile it out so that people remember it exists, thus the cfg'd Sender
-// import.
+// compile it out so that people remember it exists.
 
 use crate::bezier::Bezier;
 use crate::context::SharedStyleContext;
@@ -15,7 +14,6 @@ use crate::font_metrics::FontMetricsProvider;
 use crate::properties::animated_properties::AnimatedProperty;
 use crate::properties::longhands::animation_direction::computed_value::single_value::T as AnimationDirection;
 use crate::properties::longhands::animation_play_state::computed_value::single_value::T as AnimationPlayState;
-#[cfg(feature = "servo")]
 use crate::properties::LonghandIdSet;
 use crate::properties::{self, CascadeMode, ComputedValues, LonghandId};
 use crate::stylesheets::keyframes_rule::{KeyframesAnimation, KeyframesStep, KeyframesStepValue};
@@ -26,12 +24,8 @@ use crate::values::computed::TimingFunction;
 use crate::values::generics::box_::AnimationIterationCount;
 use crate::values::generics::easing::{StepPosition, TimingFunction as GenericTimingFunction};
 use crate::Atom;
-#[cfg(feature = "servo")]
-use crossbeam_channel::Sender;
 use servo_arc::Arc;
 use std::fmt;
-#[cfg(feature = "gecko")]
-use std::sync::mpsc::Sender;
 
 /// This structure represents a keyframes animation current iteration state.
 ///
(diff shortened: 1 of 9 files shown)


components/style/animation.rs
```diff
@@ -369,56 +365,182 @@ impl PropertyAnimation {
...
+
+    /// The animations that have been cancelled for this element. These are cleared
+    /// upon triggering a DOM event for each cancelled animation.
+    pub cancelled_animations: Vec<Animation>,
+
+    /// New animations created for this element.
+    pub new_animations: Vec<Animation>,
+}
+
+#[cfg(feature = "servo")]
```
emilio, line 387: I wouldn't bother cfg'ing out these fwiw. The less `#[cfg]`d code the better. Makes it harder for me to break servo, and makes it harder for Servo to break Gecko with unused imports and so on.

components/style/animation.rs
```diff
@@ -369,56 +365,182 @@ impl PropertyAnimation {
...
+impl ElementAnimationState {
+    /// Cancel all animations in this `ElementAnimationState`. This is typically called
+    /// when the element has been removed from the DOM.
+    pub fn cancel_all_animations(&mut self) {
+        self.cancelled_animations
+            .extend(self.finished_animations.drain(0..));
```
emilio, line 393: all across this PR, I think `0..` can just be replaced by `..`.

components/style/animation.rs
```diff
@@ -369,56 +365,182 @@ impl PropertyAnimation {
...
+                }
             }
         }
-        running_animations.push(running_animation);
+        // Otherwise just add the new running animation.
+        self.add_new_animation(new_animation);
+    }
+}
+
+#[cfg(feature = "servo")]
```
emilio, line 432: I don't get the usefulness of this as it exists, on this PR at least. This takes at least the same space as an `ElementAnimationState`, but creating an empty `ElementAnimationState` is cheap and doesn't allocate.

Am I missing something? Or is this just a footgun-preventor so that people can accidentally store empty long-lived `ElementAnimationState` objects?

emilio (commented):
Haven't looked at all the logic in this PR yet, but I'm a bit confused about `PossibleElementAnimationState`.

components/style/animation.rs
mrobinson: Okay. I've removed these.

components/style/animation.rs
mrobinson: Removed again. I've also modified this to use chain so now there is only one call to `extend(..,)`

components/style/animation.rs
mrobinson: I agree and after our conversation on Matrix I've removed this data structure. Now there is only `ElementAnimationState`.

mrobinson:
Thanks for the review! I've made the changes you've requested and also made one more modification, which is to eagerly remove empty `ElementAnimationState` entries in the layout task after processing them.

components/style/context.rs
```diff
@@ -186,25 +166,13 @@ pub struct SharedStyleContext<'a> {
...
-    pub cancelled_animations: Arc<RwLock<FxHashMap<OpaqueNode, Vec<Animation>>>>,
+    pub animation_states: Arc<RwLock<FxHashMap<OpaqueNode, ElementAnimationState>>>,
 
     /// Paint worklets
     #[cfg(feature = "servo")]
     pub registered_speculative_painters: &'a dyn RegisteredSpeculativePainters,
-
-    /// Data needed to create the thread-local style context from the shared one.
-    #[cfg(feature = "servo")]
-    pub local_context_creation_data: Mutex<ThreadLocalStyleContextCreationInfo>,
```
emilio, line 207: very nice to remove this code :)

components/style/matching.rs
```diff
@@ -437,55 +439,50 @@ trait PrivateMatchMethods: TElement {
...
     ) {
-        use crate::animation;
-
         let this_opaque = self.as_node().opaque();
-        let mut expired_transitions = vec![];
         let shared_context = context.shared;
+        let mut animation_states = shared_context.animation_states.write();
+        let mut animation_state = animation_states
+            .remove(&this_opaque)
+            .unwrap_or_else(Default::default);
```
emilio, line 447: `unwrap_or_default()`?

components/style/animation.rs
```diff
@@ -369,70 +363,168 @@ impl PropertyAnimation {
...
+        false
+    }
+
+    /// Compute before-change-style given an existing ElementAnimationState,
+    /// information from the StyleContext, and the values of the previous style
+    /// computation.
+    ///
+    /// TODO(mrobinson): This is not a correct computation of before-change-style.
+    /// For starters it's unclear why we aren't using the running transitions to
+    /// to transform this style into before-change-style.
```
emilio, line 451: nit: to to

components/style/animation.rs
```diff
@@ -369,70 +363,168 @@ impl PropertyAnimation {
...
+    pub(crate) fn cancel_transitions_with_nontransitioning_properties(
+        &mut self,
+        properties_that_transition: LonghandIdSet,
+    ) {
+        if self.running_animations.is_empty() {
+            return;
+        }
+
+        let previously_running_transitions =
+            std::mem::replace(&mut self.running_animations, Vec::new());
```
emilio, line 405: Maybe this could be marginally more efficient using `retain()` (you might need to replace the retained values with a dummy animation so you can move them to `self.cancelled_animations`, so maybe not a big deal.

components/style/animation.rs
```diff
@@ -369,70 +363,168 @@ impl PropertyAnimation {
...
+            self.finished_animations
+                .drain(..)
+                .chain(self.running_animations.drain(..))
+                .chain(self.new_animations.drain(..)),
+        );
+    }
+
+    pub(crate) fn cancel_transitions_with_nontransitioning_properties(
+        &mut self,
+        properties_that_transition: LonghandIdSet,
```
emilio, line 398: Maybe `&LonghandIdSet`?

emilio (approved):
Really nice cleanup!

jdm (approved):

components/style/matching.rs
mrobinson: Thanks! I didn't know this existed. It makes more sense here.

components/style/animation.rs
mrobinson: Fixed.

components/style/animation.rs
mrobinson: Okay. Turned this into a reference.

components/style/animation.rs
mrobinson: Is the issue here using a push for every animation. Another approach would be to rewrite `retain`, but instead of [truncating the vector](https://doc.rust-lang.org/src/alloc/vec.rs.html#1077-1097) it returns the partition index. Additionally, perhaps I could rely on `partition_in_place` though that isn't guaranteed to maintain the order. My thought was to figure out how to optimize these animation collections later, because maybe a `LinkedList` makes sense.

PR #26290, opened 2020-04-23
[WIP] Reduce number of processess for macOS WPT on CI.

jdm:
#25513 is really hitting us hard, and it go much worse recently. This is an attempt to reduce the frequency of this problem by reducing the number of concurrent processes that are running.


jdm:
My currently theory is that #25853 made this worse for us because *waves hands* GPU?

SimonSapin:
r=me for landing this whenever you think we should.

jdm:
This wasn't necessary.

PR #26312, opened 2020-04-24
Provide feedback when OpenXR can't start

### LINKED ISSUE #26213: Provide embedder feedback if openxr backend isn't set up right
Ideally we should pop up some kind of message dialog either on startup or when entering VR/AR fails. This requires some kind of coordination between the webxr backend, the DOM XR code, and the embedding API.

jdm:
- [x] These changes fix #26213
- [x] These changes do not require tests because it's testing an edge case in an environment we can't test on CI

### DIFF
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -810,7 +810,7 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
         }
 
         struct GlThread(WebGlExecutor);
-        impl webxr::openxr::GlThread for GlThread {
+        impl openxr::GlThread for GlThread {
             fn execute(&self, runnable: Box<dyn FnOnce(&surfman::Device) + Send>) {
                 let _ = self.0.send(runnable);
             }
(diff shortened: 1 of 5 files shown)


jdm:
Depends on https://github.com/servo/webxr/pull/164.

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -819,12 +819,27 @@ impl EmbedderMethods for ServoEmbedderCallbacks {
             }
         }
 
-        let discovery = webxr::openxr::OpenXrDiscovery::new(
-            Box::new(GlThread(executor)),
-            Box::new(ProviderRegistration(surface_providers)),
-            Box::new(ContextMenuCallback(embedder_proxy)),
-        );
-        registry.register(discovery);
+        if openxr::create_instance().is_ok() {
```
Manishearth, line 822: Hmm, this would be nicer if we had the prompt when you attempt to enter immersive mode, IMO. Might need another error variant

Manishearth (approved):
r=me either way

ports/libsimpleservo/api/src/lib.rs
jdm: I'll file a followup for that, since it requires a bunch more plumbing.

jdm:
@bors-servo retry
* #26315

jdm:
@bors-servo retry
#25513

Issue #11722, opened 2016-06-11
Create a central repository of Origins that can be trusted


Manishearth:
Right now, net blindly trusts requests from a script pipeline which say they are from a certain origin. This is wrong from the process isolation perspective.

Instead, a map of pipeline IDs to allowed origins should be kept in the chrome process, and origins sent from script should be checked against this.

Additionally, we may want a different origin struct, one which:

 - has the ability to support private browsing, with extra flags
 - which uses uuid instead of TLS for opaque origins

Note that for this to be totally secure we need some form of pipeline password only known to the relevant pipeline and chrome process, since right now pipeline ids can be guessed.

cc @izgzhen @asajeffrey

aneeshusa:
Does #10808 (making PipelineIds into unguessable UUIDs) help with this, or would we still need separate pipeline passwords?

Manishearth:
Unguessable is enough. I had the mental model of a separate pipeline password, that's all.

izgzhen:
"pipeline password" is something concrete or just a concept?

Manishearth:
Right now pipeline ids are incremented integers. So if manish.com's content process wishes to steal facebook.com data, it can just iterate through pipeline ids until it finds the one corresponding to facebook, and send that across.

There are two ways of fixing this. One way is to make pipeline ids unguessable; a uuid or something.

The other is to keep pipeline ids as is (for easy indexing, etc), and have an associated pipeline password for each id. Only chrome knows all the passwords, otherwise each content process only knows its password. There is a slight advantage here in that pipeline ids can still be shared and freely, e.g. if a page wishes to refer to its child iframe, just that you check the password when you rely on the pipeline id for security things.

gterzian:
This is now also relevant for messageports, which are identified by a `MessagePortId`, similar in construction to a `PipelineId`, as well as a `MessagePortRouterId`, and these ids are used by the constellation to route messages across process. 

So it will be possible for a compromised script process to guess the relevants ids and trick the constellation into forwarding fake messages to ports in other processes.

gterzian:
> Instead, a map of pipeline IDs to allowed origins should be kept in the chrome process, and origins sent from script should be checked against this.

[...]

PR #26317, opened 2020-04-25
ServiceWorker: restructure Job Queue, Register flow, to better match spec

### LINKED ISSUE #26108: ServiceWorker: restructure data structure and algorithms to better match the spec
EDIT: the second comment is more accurate, still keeping this here for a basic overview of how it currently works. We need to move the Jobqueue to the constellation or service worker manager level. 

https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/serviceworkerjob.rs#L102

Spec: https://w3c.github.io/ServiceWorker/#dfn-job-queue

The `JobQueue` used to run algorithms for service workers seems to currently be tied to having a `ScriptThread` being available. 

See for example how a job is scheduled: https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/serviceworkerjob.rs#L110

Or how a job is run:

https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/serviceworkerjob.rs#L139

So I assume this only works in the context of a window `GlobalScope`. 

Since all the service workers IDL interfaces are also available to workers, I think we should try to make the job queue also work for a worker global. 

We could try to do something similar as what is done with the [`MicrotaskQueue `](https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/microtask.rs#L28),

https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/dom/globalscope.rs#L196

 which is available on the `GlobalScope` in addition to being available on the `ScriptThread`, so that worker event-loops can also use it. 

https://github.com/servo/servo/blob/f7d3d4a44762b3196c8dcbe58b628b5a23fc7833/components/script/dom/abstractworkerglobalscope.rs#L147

Although to prevent a problem like https://github.com/servo/servo/issues/20908 we could try to just move the job queue entirely onto the global, and remove it from the script-thread, so functions like `run_job` and `schedule_job` could be made to take a `&GlobalScope` instead of a `&ScriptThread`.

gterzian:
---

- [ ] These changes fix #26108 (GitHub issue number if applicable)


gterzian:
cc @CYBAI

gterzian:
```
0 matches
  ▶ CRASH [expected OK] /_mozilla/mozilla/service-workers/service-worker-registration.https.html
  │ 
  │ assertion failed: !cx.is_null() (thread ServiceWorkerManager, at /root/.cargo/git/checkouts/rust-mozjs-8611526964119dd6/11cabde/src/rust.rs:311)
  │              at src/libstd/panicking.rs:515
  │    4: mozjs::rust::Runtime::get
  │    5: script::dom::serviceworkerglobalscope::ServiceWorkerGlobalScope::run_serviceworker_scope
  │    6: script::serviceworker_manager::ServiceWorkerManager::handle_message
  │    7: std::sys_common::backtrace::__rust_begin_short_backtrace
  │    8: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │    9: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/42abbd8878d3b67238f3611b0587c704ba94f39c/src/liballoc/boxed.rs:1008
  │   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/42abbd8878d3b67238f3611b0587c704ba94f39c/src/liballoc/boxed.rs:1008
  │              at src/libstd/sys/unix/thread.rs:87
  │   11: start_thread
  │   12: __clone
  │ [2020-04-30T08:39:35Z ERROR servo] assertion failed: !cx.is_null()
  │ Stack trace for thread "ServiceWorkerManager"
  └    1: servo::install_crash_handler::handler

```

Not sure if this is intermittent

@bors-servo try=wpt

gterzian:
Ah ok I got it, see the last commit(there is no current global where `run_serviceworker_scope` is called from)

gterzian:
Ok I think this completes the restructuring I wanted to do, and leaves some infrastructure in place to continue working on this(See the issues I've filed that link to here). 

Who wants to review it? 

@jdm it seems bors has some perverse intention to always assign you to the PRs I open, since you're still reviewing the stream PR, perhaps someone else could do this one? @Manishearth you interested?

tests/wpt/mozilla/meta/mozilla/service-workers/service-worker-registration.https.html.ini
```diff
@@ -0,0 +1,6 @@
+[service-worker-registration.https.html]
+  type: testharness
+  [Test: Asserts Active Service Worker and its Registration]
+      expected: FAIL
+  [Test: Active Service Worker ScriptURL property]
+      expected: FAIL
```
gterzian, line 6: Note, see https://github.com/servo/servo/issues/26378 for why this fails and how it will be fixed. I've written some tests making similar assertions using the "installing worker".

components/script/dom/client.rs
```diff
@@ -47,6 +47,7 @@ impl Client {
         self.active_worker.get()
     }
 
+    #[allow(dead_code)]
```
gterzian, line 50: Note: this, or a variant of it, will probably be used as part of implementing https://github.com/servo/servo/issues/26378

components/script/lib.rs
```diff
@@ -90,8 +90,6 @@ pub mod script_thread;
 #[warn(deprecated)]
 pub mod serviceworker_manager;
 #[warn(deprecated)]
-mod serviceworkerjob;
-#[warn(deprecated)]
```
gterzian, line 94: to fix: should remove the warn above, not below...

components/script/serviceworker_manager.rs
```diff
@@ -108,76 +188,194 @@ impl ServiceWorkerManager {
...
 
-    fn forward_message(&self, msg: DOMMessage, sender: &Sender<ServiceWorkerScriptMsg>) {
-        let DOMMessage { origin, data } = msg;
-        let _ = sender.send(ServiceWorkerScriptMsg::CommonWorker(
-            WorkerScriptMsg::DOMMessage { origin, data },
-        ));
+    fn handle_message_from_resource(&mut self, mediator: CustomResponseMediator) -> bool {
+        if serviceworker_enabled() {
+            if let Some(scope) = self.get_matching_scope(&mediator.load_url) {
+                if let Some(registration) = self.registrations.get_mut(&scope) {
```
CYBAI, line 194: Looks like we don't need to mutate this registration. Can we just use `get` instead 👀 ?

components/script/serviceworker_manager.rs
```diff
@@ -108,76 +188,194 @@ impl ServiceWorkerManager {
...
-                let _ = mediator.response_chan.send(None);
+    /// <https://w3c.github.io/ServiceWorker/#register-algorithm>
+    fn handle_register_job(&mut self, mut job: Job) {
+        if !job.script_url.is_origin_trustworthy() {
+            // Step 1.1
+            let _ = job
+                .client
+                .send(JobResult::RejectPromise(JobError::SecurityError));
+            return;
+        } else if job.script_url.origin() != job.referrer.origin() ||
```
CYBAI, line 249: Due to different from checking origin trustworthy and different steps, maybe it's good to have it as another standalone `if` expression?

```rust
if !job.script_url.is_origin_trustworthy() {
  ...
}

if job.script_url.origin() != job.referrer.origin() ||
  job.scope_url.origin() != job.referrer.origin()
{
  ...
}
```

components/script_traits/script_msg.rs
```diff
@@ -385,16 +386,117 @@ pub struct SWManagerSenders {
...
+}
+
+impl PartialEq for Job {
+    /// Equality criteria as described in https://w3c.github.io/ServiceWorker/#dfn-job-equivalent
+    fn eq(&self, other: &Self) -> bool {
+        let same_job = self.job_type == other.job_type;
+        if same_job {
+            match self.job_type {
+                JobType::Register | JobType::Update => {
+                    self.scope_url == other.scope_url && self.script_url == other.script_url
```
CYBAI, line 490: > For register and update jobs, their scope url, script url, worker type, and update via cache mode are the same.

From spec, looks like we will want to check `worker type`and `update via cache mode` in the future? so maybe it's good to add a TODO here 🤔 ?

CYBAI (commented):
Just had a review and found some nits 👀 but it, overall, looks quite straightforward to me!

components/script/dom/serviceworker.rs
```diff
@@ -35,31 +36,35 @@ pub struct ServiceWorker {
     script_url: DomRefCell<String>,
     scope_url: ServoUrl,
     state: Cell<ServiceWorkerState>,
-    skip_waiting: Cell<bool>,
```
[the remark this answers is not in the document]
gterzian: It didn't seem to be used anywhere anymore, but we can re-add it when we implement the part of the spec that uses the flag.

components/script/serviceworker_manager.rs
```diff
@@ -29,11 +31,112 @@ enum Message {
...
+    }
+
+    /// Send a message to the running service worker scope.
+    fn send_message(&self, msg: ServiceWorkerScriptMsg) {
+        let _ = self.sender.send(msg);
+    }
+}
+
+/// When updating a registration, which worker are we targetting?
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
gterzian: Yes, and there is one variant that isn't used yet(or actually two), so you get a warning if any variant is unused.

components/script/dom/serviceworker.rs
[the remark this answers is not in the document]
CYBAI: Indeed! This should be easy to re-add!

jdm:
r? @asajeffrey 
Alan said he would review this.

components/constellation/constellation.rs
```diff
@@ -2622,9 +2620,26 @@ where
...
+    /// which are scoped per origin.
+    fn handle_schedule_serviceworker_job(&mut self, pipeline_id: PipelineId, job: Job) {
+        let origin = job.scope_url.origin();
+
+        if self
+            .check_origin_against_pipeline(&pipeline_id, &origin)
+            .is_err()
+        {
+            return warn!(
+                "Attempt to schedule a serviceworker job from an origin not matching the origin of the job."
```
asajeffrey, line 2637: Is there a spec link for this origin check? The spec links above don't talk about origin checking.

components/constellation/constellation.rs
```diff
@@ -2635,7 +2650,8 @@ where
                     own_sender: own_sender.clone(),
                     receiver,
                 };
-                let content = ServiceWorkerUnprivilegedContent::new(sw_senders, scope.origin());
+                let content =
+                    ServiceWorkerUnprivilegedContent::new(sw_senders, job.scope_url.origin());
```
asajeffrey, line 2654: nit: `origin` rather than `job.scope_url.origin())`?

asajeffrey (approved):
OK, this looks like a big improvement! You can r=me once @CYBAI is OK with the changes.

components/constellation/constellation.rs
gterzian: Yes you're right it's not in the Spec, the idea is just to check that the origin as stored in the job message matches what the constellation has stored for the pipeline from which the message originate(it's still only limited as a check and there is a note in the function used to https://github.com/servo/servo/issues/11722)

CYBAI (approved):
Yeah! Thanks for addressing the comments! I'm excited to help remaining issues after this one merged!! 🎉

gterzian:
@bors-servo r=asajeffrey

Thanks for the reviews @asajeffrey and @CYBAI !

gterzian:
Oh wait I forgot to address https://github.com/servo/servo/pull/26317#discussion_r418883674

gterzian:
Ah no that's nonsense actually, it doesn't matter whether I remove the warn above or below...

gterzian:
```
Tests with unexpected results:
  ▶ Unexpected subtest result in /_mozilla/mozilla/service-workers/service-worker-registration.https.html:
  │ FAIL [expected PASS] Test: Installing Service Worker ScriptURL property
  │   → navigator.serviceWorker is undefined
  │ 
  │ register_sw@https://web-platform.test:8443/_mozilla/mozilla/service-workers/service-worker-registration.https.html:9:22
  │ @https://web-platform.test:8443/_mozilla/mozilla/service-workers/service-worker-registration.https.html:24:10
  │ Test.prototype.step@https://web-platform.test:8443/resources/testharness.js:1975:25
  │ promise_test/tests.promise_tests</<@https://web-platform.test:8443/resources/testharness.js:592:36
  └ promise_test/tests.promise_tests<@https://web-platform.test:8443/resources/testharness.js:591:20
  ▶ Unexpected subtest result in /_mozilla/mozilla/service-workers/service-worker-registration.https.html:
  │ FAIL [expected PASS] Test: Asserts Installing Service Worker and its Registration
  │   → navigator.serviceWorker is undefined
  │ 
  │ register_sw@https://web-platform.test:8443/_mozilla/mozilla/service-workers/service-worker-registration.https.html:9:22
  │ @https://web-platform.test:8443/_mozilla/mozilla/service-workers/service-worker-registration.https.html:38:10
  │ Test.prototype.step@https://web-platform.test:8443/resources/testharness.js:1975:25
  │ promise_test/tests.promise_tests</<@https://web-platform.test:8443/resources/testharness.js:592:36
  └ promise_test/tests.promise_tests<@https://web-platform.test:8443/resources/testharness.js:591:20
  ▶ OK [expected CRASH] /_mozilla/mozilla/fetch_cannot_overwhelm_system.window.html
  ▶ Unexpected subtest result in /_mozilla/mozilla/service-workers/service-worker-registration.https.html:
  │ FAIL [expected PASS] Test: Installing Service Worker ScriptURL property
  │   → navigator.serviceWorker is undefined
  │ 
  │ register_sw@https://web-platform.test:8443/_mozilla/mozilla/service-workers/service-worker-registration.https.html:9:22

```

Ok have to look into those, interesting it only occured on the Layout 2020 linux box...

CYBAI:
> Ok have to look into those, interesting it only occured on the Layout 2020 linux box...

Maybe you'll need to turn on the SW preference for layout-2020 👀?

gterzian:
```
  ▶ TIMEOUT [expected OK] /_mozilla/mozilla/service-workers/service-worker-registration.https.html
  │   → TestRunner hit external timeout (this may indicate a hang)
  │ 
  │ assertion failed: !cx.is_null() (thread ServiceWorker for https://web-platform.test:8443/_mozilla/mozilla/service-workers/resources/sw.js, at /Users/gregory/.cargo/git/checkouts/rust-mozjs-8611526964119dd6/11cabde/src/rust.rs:311)
  │    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
  │    1: core::fmt::write
  │    2: std::io::Write::write_fmt
  │    3: servo::backtrace::print
  │    4: servo::main::{{closure}}
  │    5: std::panicking::rust_panic_with_hook
  │    6: std::panicking::begin_panic
  │    7: mozjs::rust::Runtime::get
  │    8: std::sys_common::backtrace::__rust_begin_short_backtrace
  │    9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │   11: std::sys::unix::thread::Thread::new::thread_start
  │   12: AssociationsManager::_map
  │   13: AssociationsManager::_map
  │ assertion failed: !cx.is_null() (thread ServiceWorker for https://web-platform.test:8443/_mozilla/mozilla/service-workers/resources/sw.js, at /Users/gregory/.cargo/git/checkouts/rust-mozjs-8611526964119dd6/11cabde/src/rust.rs:311)
  │    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
  │    1: core::fmt::write
  │    2: std::io::Write::write_fmt
  │    3: servo::backtrace::print
  │    4: servo::main::{{closure}}
  │    5: std::panicking::rust_panic_with_hook
  │    6: std::panicking::begin_panic
  │    7: mozjs::rust::Runtime::get
  │    8: std::sys_common::backtrace::__rust_begin_short_backtrace
  │    9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │   11: std::sys::unix::thread::Thread::new::thread_start
  │   12: AssociationsManager::_map
  │   13: AssociationsManager::_map
  │ assertion failed: !cx.is_null() (thread ServiceWorker for https://web-platform.test:8443/_mozilla/mozilla/service-workers/resources/sw.js, at /Users/gregory/.cargo/git/checkouts/rust-mozjs-8611526964119dd6/11cabde/src/rust.rs:311)
  │    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
  │    1: core::fmt::write
  │    2: std::io::Write::write_fmt
  │    3: servo::backtrace::print
  │    4: servo::main::{{closure}}
  │    5: std::panicking::rust_panic_with_hook
  │    6: std::panicking::begin_panic
  │    7: mozjs::rust::Runtime::get
  │    8: std::sys_common::backtrace::__rust_begin_short_backtrace
  │    9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   10: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │   11: std::sys::unix::thread::Thread::new::thread_start
  │   12: AssociationsManager::_map
  │   13: AssociationsManager::_map
  │ [2020-05-12T08:14:53Z ERROR servo] assertion failed: !cx.is_null()
  │ [2020-05-12T08:14:53Z ERROR servo] assertion failed: !cx.is_null()
  └ [2020-05-12T08:14:53Z ERROR servo] assertion failed: !cx.is_null()

  [1/1] No tests running.

```

I'm also getting something else interesting, if you run the test locally and you don't close the result window, eventually it panics. This might be because I removed the timeout mechanism, which perhaps is actually necessary. 

> Maybe you'll need to turn on the SW preference for layout-2020 👀?

Thanks I'll look into it too

gterzian:
Ah Ok, re layout 2020, there is a separate metadata folder, which I hadn't updated yet, the failures are actually expected.

jdm:
Yeah, we don't distinguish between testharness tests and reftests when running the servo-specific WPT tests for layout 2020, unfortunately.

gterzian:
So the hang I'm noticing is the Js being stuck in a loop trying to shutdown, and apparently the runtime used by the SW scopes is kept alive, so looking into that now...

gterzian:
Ok so to recap, currently on master, `run_serviceworker_scope` doesn't work, it's actually never called by the tests at `mozilla/service-workers/service-worker-registration.https.html`, because the current implementation only starts the SW scope when a message is sent to the worker. 

So if you add any sort of `postMessage` call in the test, you get the below panic(on the current master:

```
pid:1391 assertion failed: !cx.is_null() (thread ServiceWorkerManager, at /Users/gregory/Projects/rust-mozjs/src/rust.rs:310)
 0:03.43 pid:1391 stack backtrace:
 0:03.84 pid:1391    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
 0:03.84 pid:1391    1: core::fmt::write
 0:03.84 pid:1391    2: std::io::Write::write_fmt
 0:03.84 pid:1391    3: servo::backtrace::print
 0:03.84 pid:1391    4: servo::main::{{closure}}
 0:03.84 pid:1391    5: std::panicking::rust_panic_with_hook
 0:03.84 pid:1391    6: std::panicking::begin_panic
 0:03.84 pid:1391    7: mozjs::rust::Runtime::get
 0:03.84 pid:1391    8: script::serviceworker_manager::ServiceWorkerManager::wakeup_serviceworker
 0:03.84 pid:1391    9: script::serviceworker_manager::ServiceWorkerManager::handle_message
 0:03.85 pid:1391   10: std::sys_common::backtrace::__rust_begin_short_backtrace
 0:03.85 pid:1391   11: core::ops::function::FnOnce::call_once{{vtable.shim}}
 0:03.85 pid:1391   12: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
 0:03.85 pid:1391   13: std::sys::unix::thread::Thread::new::thread_start
 0:03.85 pid:1391   14: _pthread_body
 0:03.85 pid:1391   15: _pthread_start
 0:03.85 pid:1391 [2020-05-16T08:53:06Z ERROR servo] assertion failed: !cx.is_null()
```

So that is partially fixed by the commit in this PR that does not use `GlobalScope::current()`.

However, I think I am running into other issues, because I assume I could just copy this over into the new setup, however "running a SW scope" wasn't working already on master, and these problems are now emerging because in the current setup the SW scope is started "earlier", when the corresponding DOM objects are first create. 

The concrete problem I am stuck with now is that while the test is passing, when closing the window you get a hang due to the loop never exiting at https://github.com/servo/servo/blob/afc4e7ec251ff4decc9915d065440a34dbb9cfb5/components/script/script_runtime.rs#L428

So it seems the SW scope is leaking the runtime it creates, and I can't tell why. 

I also tried using an option around the runtime and explicitly setting it to none when the thread exits, however that triggers what looks like a GC related panic, which I assume occurs because dropping the runtime clean-up a few things that should only be cleaned-up after the GC? For example it appeared like the call to `DeleteJobQueue` would panic. 

@asajeffrey I've looked at the ScriptThread code and I couldn't see if there was some specific clean-up of the runtime, other than the use of the `ScriptMemoryFailsafe`, which seems to set it to `None`, which I tried as noted above. 

Currently each SW scope has it's own runtime, I also tried storing the parent on the SW manager, and then creating a child for each SW scope, but that gave the exactly same problem. 

Any ideas?

jdm:
Can you share some more details about the panic that occurs when you use an Option and set it to None?

gterzian:
```
0:05.66 pid:2590    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
 0:05.66 pid:2590    1: core::fmt::write
 0:05.66 pid:2590    2: std::io::Write::write_fmt
 0:05.66 pid:2590    3: servo::backtrace::print
 0:05.66 pid:2590    4: servo::install_crash_handler::handler
 0:05.66 pid:2590    5: AssociationsManager::_map
 0:05.66 pid:2590    6: DeleteJobQueue
 0:05.66 pid:2590    7: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
 0:05.66 pid:2590    8: script::dom::bindings::codegen::Bindings::ServiceWorkerGlobalScopeBinding::ServiceWorkerGlobalScopeBinding::_finalize
 0:05.66 pid:2590    9: _ZN2js2gc5Arena8finalizeI8JSObjectEEmP8JSFreeOpNS0_9AllocKindEm
 0:05.66 pid:2590   10: _ZL19FinalizeTypedArenasI8JSObjectEbP8JSFreeOpPPN2js2gc5ArenaERNS4_15SortedArenaListENS4_9AllocKindERNS3_11SliceBudgetE
 0:05.66 pid:2590   11: _ZN2js2gc10ArenaLists18foregroundFinalizeEP8JSFreeOpNS0_9AllocKindERNS_11SliceBudgetERNS0_15SortedArenaListE
 0:05.67 pid:2590   12: _ZN2js2gc9GCRuntime17finalizeAllocKindEP8JSFreeOpRNS_11SliceBudgetE
 0:05.67 pid:2590   13: _ZN11sweepaction18SweepActionForEachI13ContainerIterIN7mozilla7EnumSetIN2js2gc9AllocKindEyEEES7_E3runERNS5_11SweepAction4ArgsE
 0:05.67 pid:2590   14: _ZN11sweepaction19SweepActionSequence3runERN2js2gc11SweepAction4ArgsE
 0:05.67 pid:2590   15: _ZN11sweepaction18SweepActionForEachIN2js2gc19SweepGroupZonesIterEP9JSRuntimeE3runERNS2_11SweepAction4ArgsE
 0:05.67 pid:2590   16: _ZN11sweepaction19SweepActionSequence3runERN2js2gc11SweepAction4ArgsE
 0:05.67 pid:2590   17: _ZN11sweepaction18SweepActionForEachIN2js2gc15SweepGroupsIterEP9JSRuntimeE3runERNS2_11SweepAction4ArgsE
 0:05.67 pid:2590   18: _ZN2js2gc9GCRuntime19performSweepActionsERNS_11SliceBudgetE
 0:05.67 pid:2590   19: _ZN2js2gc9GCRuntime16incrementalSliceERNS_11SliceBudgetERKN7mozilla5MaybeI18JSGCInvocationKindEEN2JS8GCReasonERNS0_13AutoGCSessionE
 0:05.67 pid:2590   20: _ZN2js2gc9GCRuntime7gcCycleEbNS_11SliceBudgetERKN7mozilla5MaybeI18JSGCInvocationKindEEN2JS8GCReasonE
 0:05.67 pid:2590   21: _ZN2js2gc9GCRuntime7collectEbNS_11SliceBudgetERKN7mozilla5MaybeI18JSGCInvocationKindEEN2JS8GCReasonE
 0:05.67 pid:2590   22: _ZN2js2gc9GCRuntime2gcE18JSGCInvocationKindN2JS8GCReasonE
 0:05.67 pid:2590   23: _ZN9JSRuntime14destroyRuntimeEv
 0:05.67 pid:2590   24: _ZN2js14DestroyContextEP9JSContext
 0:05.67 pid:2590   25: <mozjs::rust::Runtime as core::ops::drop::Drop>::drop
 0:05.67 pid:2590   26: _ZN4core3ptr13drop_in_place17hd7451d67d27d6bfcE.llvm.4784272764368722947
 0:05.67 pid:2590   27: script::dom::workerglobalscope::WorkerGlobalScope::clear_js_runtime
 0:05.67 pid:2590   28: std::sys_common::backtrace::__rust_begin_short_backtrace
 0:05.67 pid:2590   29: core::ops::function::FnOnce::call_once{{vtable.shim}}
 0:05.67 pid:2590   30: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
 0:05.67 pid:2590   31: std::sys::unix::thread::Thread::new::thread_start
 0:05.67 pid:2590   32: AssociationsManager::_map
 0:05.67 pid:2590   33: AssociationsManager::_map
```

The test does first print out `pid:2590 Hey Servo; lets cache something! :)`, and then the sw scope is immediately shut down(we don't run the event-loop at all).

Achieved by running `tests/wpt/mozilla/tests/mozilla/service-workers/service-worker-registration.https.html` with https://github.com/servo/servo/commit/c37cef50fc6d393e62203c27c6c7f329ea8025fb

jdm:
Given [this code](https://github.com/servo/servo/blob/0a00ea3db3c2353c78435030040fd3f57f5f4497/components/script/script_runtime.rs#L388-L405), it looks like we should have already run the Drop implementation for Runtime before the GC starts, which is triggered by dropping the actual RustRuntime field. It's not clear to me why we would end up in DropJobQueue as a result of running the _finalize function for ServiceWorkerGlobalScope, though. I'm initially suspicious that that is not an accurate top stack frame.

jdm:
Alternatively, what if we're somehow re-entering the code because the Runtime field of the worker global scope hasn't actually finished clearing by the time the GC tries to drop the WorkerGlobalScope member of ServiceWorkerGlobalScope? We could test that by doing something like:
```rust
let runtime = self.runtime.take();
drop(runtime);
```
in the clear_js_runtime method. This would ensure that if we re-enter, we only see a None value, not a Some value that is partially dropped.

gterzian:
Thanks! The manual drop seems to solve the problem(and it also solves the previous hang where the runtime would leak), so closing the window actually cleanly shuts down Servo with https://github.com/servo/servo/commit/76c45511caedda19fd3165d11ca9dc8608f6b473

gterzian:
@asajeffrey Could you please take a look at the second commit? It fixes the various issues I have encountered, which are:

1. There is no current global when `run_serviceworker_scope` is called.
2. We need to clear the js runtime otherwise it leaks even after the thread exits.
3. We're not actually running the event-loop, only executing the initial script, because we don't have a proper shutdown mechanism yet. I think it's best done in a follow-up PR, and not running an event-loop is not exactly a regression because the entire `run_serviceworker_scope` is currently not working on master.

asajeffrey:
LGTM!

gterzian:
Was missing a layout 2020 FAIL expectation

@bors-servo r=asajeffrey

PR #26321, opened 2020-04-26
Update Ubuntu base docker image for Taskcluster

SimonSapin:
There may be security or other fixes?


SimonSapin:
Multiple timeouts on macOS, but Linux tasks are green.

r? @jdm

PR #26325, opened 2020-04-26
Improve devtools experience when navigating

### LINKED ISSUE #15425: devtools can't handle new pipelines
Refreshes now give a different pipeline, and when you navigate to a new page, you also get a new pipeline. Devtools can currently only track a single pipeline (and it can't show historical data), so we can't effectively track network activity with it.

From discussion with @jdm:
We need some sort of coordination from the constellation (maybe?) to tell the devtools that pipeline 1 is navigating to pipeline 2 and the devtools should report stuff about pipeline 2 now

jdm:
The primary motivation for this work was to fix #15425, and these changes make it possible to use the devtools to meaningfully inspect multiple pages when navigating between them. Navigating through session history is not yet supported.

These changes also include improvements to the dedicated worker support, which broke at some point. We now can observe console messages in workers.

### DIFF
--- a/components/devtools/actors/worker.rs
+++ b/components/devtools/actors/worker.rs
@@ -3,14 +3,49 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
-use devtools_traits::WorkerId;
+use crate::protocol::JsonPacketStream;
+use devtools_traits::DevtoolScriptControlMsg::WantsLiveNotifications;
+use devtools_traits::{DevtoolScriptControlMsg, WorkerId};
+use ipc_channel::ipc::IpcSender;
+use msg::constellation_msg::TEST_PIPELINE_ID;
 use serde_json::{Map, Value};
+use servo_url::ServoUrl;
+use std::cell::RefCell;
 use std::net::TcpStream;
+
+#[derive(Clone, Copy)]
+#[allow(dead_code)]
+pub enum WorkerType {
+    Dedicated = 0,
+    Shared = 1,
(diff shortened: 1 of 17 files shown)


gterzian:
I would be happy to review this if @asajeffrey has other things to do, I started casually looking at this yesterday already...

In any case, I do have one suggestion: instead of using the `BrowsingContextId` as a kind of unique identifier that can survive navigation from one pipeline to another, how about we introduce a new `DevtoolsId` or similar in the `msg` crate, to be used for that purpose? 

I think that would make it easier to use in the context of the Service worker as well, since unlike the dedicated worker that one cannot easily be tied to a BC(which I guess is why this PR doesn't try to do this).

jdm:
I would rather file that as a possible improvement, since service worker support is only hypothetical right now.

gterzian (approved):
Sounds good, then as far as I can tell it looks good to me.

components/devtools/actors/worker.rs
```diff
@@ -19,11 +54,94 @@ impl Actor for WorkerActor {
...
+        Ok(match msg_type {
+            "attach" => {
+                let msg = AttachedReply {
+                    from: self.name(),
+                    type_: "attached".to_owned(),
+                    url: self.url.as_str().to_owned(),
+                };
+                self.streams.borrow_mut().push(stream.try_clone().unwrap());
+                stream.write_json_packet(&msg);
+                // FIXME: fix messages to not requir forging a pipeline for worker messages
```
gterzian, line 71: nit: typo

jdm:
@bors-servo r=gterzian
Thanks for reviewing!

jdm:
@bors-servo retry
* #25513

PR #26331, opened 2020-04-27
Restyle should reflect animations and transitions

### LINKED ISSUE #20116: Servo's transitions code stopped making sense after #15900.
In particular, what used to update the _new_ style (as expected, since that's what you're animating), started updating the old style:

```rust
       if let Some(ref mut old) = *old_values {
           self.update_animations_for_cascade(shared_context, old,
                                              possibly_expired_animations);
       }
```

Which doesn't really make sense.

mrobinson:
When doing a restyle, we should apply animations and transitions to the
new style so that it is reflected in `getComputedStyle()` and the new
style information properly cascades. This is the first part of properly
ticking animations and transitions.

This causes a couple new animations tests failures (along with many new
passes), but we currently don't have support for properly handling
animations after they have completed, so this isn't totally unexpected.

---

- [x] These changes fix #20116

### DIFF
--- a/components/style/matching.rs
+++ b/components/style/matching.rs
@@ -445,7 +445,10 @@ trait PrivateMatchMethods: TElement {
         let mut animation_state = animation_states.remove(&this_opaque).unwrap_or_default();
 
         if let Some(ref mut old_values) = *old_values {
-            animation_state.compute_before_change_style::<Self>(
+            // We convert old values into `before-change-style` here.
+            // https://drafts.csswg.org/css-transitions/#starting
+            animation_state.apply_completed_animations(old_values);
+            animation_state.apply_running_animations::<Self>(
                 shared_context,
                 old_values,
                 &context.thread_local.font_metrics_provider,
@@ -475,10 +478,20 @@ trait PrivateMatchMethods: TElement {
                 .cancel_transitions_with_nontransitioning_properties(&transitioning_properties);
         }
 
-        animation_state.finished_animations.clear();
+        animation_state.apply_running_animations::<Self>(
+            shared_context,
+            new_values,
+            &context.thread_local.font_metrics_provider,
+        );
+        animation_state.apply_new_animations::<Self>(
+            shared_context,
+            new_values,
+            &context.thread_local.font_metrics_provider,
+        );
 
-        // If the ElementAnimationState is empty, don't push it to save
-        // memory and to avoid extra processing later.
+        // If the ElementAnimationState is empty, and don't store it in order to
+        // save memory and to avoid extra processing later.
+        animation_state.finished_animations.clear();
         if !animation_state.is_empty() {
             animation_states.insert(this_opaque, animation_state);
         }
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -19,20 +19,34 @@ skip: true
     skip: false
   [CSS2]
     skip: false
+    [linebox]
+      skip:false
+      [animations]
+        skip: true
   [css-align]
     skip: false
+    [animation]
+      skip: true
   [css-animations]
     skip: false
   [css-backgrounds]
     skip: false
+    [animations]
+      skip: true
   [css-color]
     skip: false
+    [animation]
+      skip: true
   [css-conditional]
     skip: false
   [css-flexbox]
     skip: false
+    [animation]
+      skip: true
   [css-fonts]
     skip: false
+    [animations]
+      skip: true
   [css-images]
     skip: false
   [css-paint-api]
@@ -41,18 +55,28 @@ skip: true
     skip: false
   [css-text]
     skip: false
+    [animations]
+      skip: true
     [i18n]
       skip: true
   [css-text-decor]
     skip: false
   [css-transforms]
     skip: false
+    [animation]
+      skip: true
   [css-transitions]
     skip: false
+    [animations]
+      skip: true
   [css-ui]
     skip: false
+    [animation]
+      skip: true
   [css-values]
     skip: false
+    [animations]
+      skip: true
   [css-variables]
     skip: false
   [cssom]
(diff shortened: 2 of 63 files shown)


nox:
r? @jdm 

I don't know anything about animations yet and I'm taking the week off starting tomorrow.

components/style/matching.rs
```diff
@@ -445,7 +445,10 @@ trait PrivateMatchMethods: TElement {
         let mut animation_state = animation_states.remove(&this_opaque).unwrap_or_default();
 
         if let Some(ref mut old_values) = *old_values {
-            animation_state.compute_before_change_style::<Self>(
+            // We convert old values into `before-change-style` here.
+            let old_values = Arc::make_mut(old_values);
```
emilio, line 449: So, this will copy the style almost surely, right? It'd be nice to not do that if there are no animations running.

components/style/matching.rs
```diff
@@ -475,6 +478,20 @@ trait PrivateMatchMethods: TElement {
                 .cancel_transitions_with_nontransitioning_properties(&transitioning_properties);
         }
 
+        {
+            let new_values = Arc::make_mut(new_values);
```
emilio, line 482: Same, it seems this will disable all sorts of style sharing optimizations.

emilio (approved):
This looks fine but it'd be nice to avoid cloning the style unconditionally. Maybe just wrapping the relevant blocks in an `if !animation_state.is_empty()` or such would work.

jdm (approved):

mrobinson:
Thanks for the reviews! I'm about to push a new version of this change which switches back to using `Arc<ComputedValues>` everywhere so that callers of new methods don't accidentally clone styles. It will also eliminate one more use of eager `Arc::make_mut` that existed prior to this change.

components/style/matching.rs
mrobinson: Oh, nice catch! I've modified this so that the methods now take an `Arc` which will ensure that callers don't accidentally clone the style like I did in this version of the change.

components/style/matching.rs
mrobinson: Same here. I originally went with an `is_empty()` check, but ended up doing this in the methods themselves to enable more fine-grained control over when the style is cloned.

jdm:
So many new passing tests!

mrobinson:
@bors-servo r=jdm,emilio

Trying this again with updated expectations, but this might take a couple rounds.

jdm:
It might be worth using try=wpt to shake out the failures before going through a complete merge cycle.

mrobinson:
@jdm Thanks for the tip. I'll give that a shot for the next round.

mrobinson:
@bors-servo try=wpt

Trying a new version of this branch that skips various directories of tests that are now newly flaky. Hopefully this is one of the first steps in making these tests not flaky at all, but unfortunately this change causes them to pass with much more likelihood -- increasing flakiness.

mrobinson:
@bors-servo r=jdm,emilio

:crossed_fingers:

mrobinson:
@bors-servo try=wpt

Skipped a few more directories which are flaky on Mac as well.

jdm:
Note - try=wpt is linux-only; try=wpt-mac is mac-only.

jdm:
There's an ini file with a problem:
```
ParseError: Token 'group_start' doesn't equal expected type 'group_end':  line 28
  File "/repo/python/servo/testing_commands.py", line 342, in test_tidy
    tidy_failed = tidy.scan(not all_files, not no_progress, stylo=stylo, no_wpt=no_wpt)
  File "/repo/python/tidy/servo_tidy/tidy.py", line 1141, in scan
    for error in errors:
  File "/repo/python/tidy/servo_tidy/tidy.py", line 517, in check_manifest_dirs
    p = parser.parse(lines)
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 789, in parse
    return p.parse(stream)
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 540, in parse
    self.manifest()
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 564, in manifest
    self.data_block()
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 585, in data_block
    self.data_block()
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 586, in data_block
    self.eof_or_end_group()
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 591, in eof_or_end_group
    self.expect(token_types.group_end)
  File "/repo/tests/wpt/web-platform-tests/tools/wptrunner/wptrunner/wptmanifest/parser.py", line 555, in expect
    "Token '{}' doesn't equal expected type '{}'".format(self.token[0], type))
```

tests/wpt/include.ini
```diff
@@ -19,20 +19,34 @@ skip: true
     skip: false
   [CSS2]
     skip: false
+    [linebox]
+      skip:false
```
jdm, line 23: Possibly the missing space here?

mrobinson: Looks like I had an extra level of indentation in the `css-align` and `css-color` sections. Should be fixed now.

mrobinson:
@bors-servo try=wpt-mac

It looked like the tidy precheck passed without issues, so I feel a bit safer about triggering this now.

mrobinson:
@bors-servo r=jdm,emilio

The WPT failure is:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 58: ordinal not in range(128)
```

jdm:
@mrobinson For future reference, look at the filtered-wpt-errorsummary.log file instead of the live log.

mrobinson:
@jdm Thanks for the pointer. Looks like the real failure during the try was #25513.

PR #26362, opened 2020-04-30
Add support for animationend event

mrobinson:
This is triggered when an animation finishes. This is a high priority
because it allows us to start rooting nodes with animations in the
script thread.

This doesn't yet cause a lot of tests to pass because they rely on the
existence of `Document.getAnimations()` and the presence of
`animationstart` and `animationiteration` events.

---


components/script/script_thread.rs
```diff
@@ -2916,66 +2926,91 @@ impl ScriptThread {
     /// Handles firing of transition-related events.
     ///
     /// TODO(mrobinson): Add support for more events.
-    fn handle_transition_event(
+    fn handle_transition_or_animation_even(
```
jdm, line 2929: event

jdm (approved):
Looks great!

emilio (approved):
Looks sane, thanks.

components/script/dom/webidls/EventHandler.webidl
```diff
@@ -90,6 +90,11 @@ interface mixin GlobalEventHandlers {
            attribute EventHandler onwaiting;
 };
 
+// https://drafts.csswg.org/css-animations/#interface-globaleventhandlers-idl
+partial interface mixin GlobalEventHandlers {
+           attribute EventHandler onanimationend;
```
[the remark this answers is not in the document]
mrobinson: Yeah, the indentation in some of the webidls files is a bit bizarre. I think it's due to copy-and-paste effects from specification documents. :smile:  I've tried to match the indentation of the lines surrounding this one.

components/script/script_thread.rs
mrobinson: Oof. Thanks for pointing this out. :sweat_smile:

PR #26363, opened 2020-04-30
Split animation cancellation from update_style_for_animation

mrobinson:
`update_style_for_animation` previously handled both canceling defunct
animations and also updating style to reflect current animation state.
This change splits those two concerns because we want to start handling
replaced or canceled animations and finished animations in two different
places.

This is a refactor, so ideally it shouldn't change any behavior.

---

### DIFF
--- a/components/style/animation.rs
+++ b/components/style/animation.rs
@@ -9,7 +9,7 @@
 
 use crate::bezier::Bezier;
 use crate::context::SharedStyleContext;
-use crate::dom::{OpaqueNode, TElement};
+use crate::dom::{OpaqueNode, TElement, TNode};
 use crate::font_metrics::FontMetricsProvider;
 use crate::properties::animated_properties::AnimatedProperty;
 use crate::properties::longhands::animation_direction::computed_value::single_value::T as AnimationDirection;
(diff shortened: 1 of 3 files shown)


components/style/animation.rs
```diff
@@ -531,6 +531,63 @@ impl ElementAnimationState {
...
+        maybe_start_animations(element, &context, &new_style, self);
+    }
+
+    /// Update our transitions given a new style, canceling or starting new animations
+    /// when appropriate.
+    pub fn update_transitions_for_new_style(
+        &mut self,
+        context: &SharedStyleContext,
+        opaque_node: OpaqueNode,
+        before_change_style: &Option<Arc<ComputedValues>>,
```
emilio, line 572: Maybe make this `Option<&Arc<>>`? `&Option` feels a bit weird but...

emilio (approved):

components/style/animation.rs
mrobinson: That makes sense. The latest version changes this to use `Option<&Arc<>>`.

jdm:
@bors-servo retry
* #25513

PR #26369, opened 2020-04-30
Implement 'vertical-align' for layout 2020

### LINKED ISSUE #19050: Line height issue when mixing font sizes.
Line height is too big when mixing 2 different font sizes in same block.

lineheig

```
<!DOCTYPE html>
<html>
	<body>
		<div style="width: 300px;">	
			<span style="font-size: 2em;">Akcia!</span>
			Lorem ipsum dolor sit amet, consectetur adipiscing elit. Dolere malum est: in crucem qui agitur, beatus esse non potest. Sed mehercule pergrata mihi oratio tua. Non est ista, inquam, Piso, magna dissensio. Si est nihil nisi corpus, summa erunt illa: valitudo, vacuitas doloris, pulchritudo, cetera. Verba tu fingas et ea dicas, quae non sentias? Duo Reges: constructio interrete. 
		</div>
	</body>
</html>
```

### LINKED ISSUE #25882: Implement 'vertical-align'
This property determines the vertical (block direction) position of inline-level fragments.

https://drafts.csswg.org/css2/visudet.html#propdef-vertical-align
https://drafts.csswg.org/css2/visudet.html#inline-non-replaced

Currently all such fragments are **incorrectly** positioned with their top (block-start) margin edge at the top of the line:

https://github.com/servo/servo/blob/c0d9ee0ce241ee0dff86c4194588d59fe436efe6/components/layout_2020/flow/inline.rs#L378
https://github.com/servo/servo/blob/c0d9ee0ce241ee0dff86c4194588d59fe436efe6/components/layout_2020/flow/inline.rs#L470
https://github.com/servo/servo/blob/c0d9ee0ce241ee0dff86c4194588d59fe436efe6/components/layout_2020/flow/inline.rs#L705

In terms of implementation, it is likely easier (and maybe necessary) to do vertical positioning within a line *after* determining what content will be part of that line and computing the heights and baseline positions of everything within that line. This includes recurring and doing the full layout of atomic inline-level boxes such as `display: inline-block`, and adding a field for the baseline position in the relevant return type (the `layout_2020::formatting_contexts::IndependentLayout` struct).

Additionally, the `top` and `bottom` values require special handling (and likely some recursion, since nested inline boxes can have them). They create an "aligned subtree" where the content is vertically-aligned separately from anything outside, then that fragment is positioned vertically as a whole.

ferjm:
- [X] These changes fix #25882 and fix #19050

### DIFF
--- a/components/layout_2020/flow/construct.rs
+++ b/components/layout_2020/flow/construct.rs
@@ -423,7 +423,7 @@ where
         contents: Contents,
     ) -> ArcRefCell<InlineLevelBox> {
         let box_ = if display_inside == DisplayInside::Flow && !contents.is_replaced() {
-            // We found un inline box.
+            // We found an inline box.
             // Whatever happened before, all we need to do before recurring
             // is to remember this ongoing inline level box.
             self.ongoing_inline_boxes_stack.push(InlineBox {
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -26,7 +26,7 @@ use gfx::text::text_run::GlyphRun;
 use servo_arc::Arc;
 use style::dom::OpaqueNode;
 use style::properties::ComputedValues;
-use style::values::computed::{Length, LengthPercentage, Percentage};
+use style::values::computed::{Length, LengthPercentage, LineHeight, Percentage};
 use style::values::specified::text::TextAlignKeyword;
 use style::values::specified::text::TextDecorationLine;
 use style::Zero;
@@ -338,6 +338,7 @@ impl InlineFormattingContext {
                 ifc.current_nesting_level = partial.parent_nesting_level
             } else {
                 ifc.lines.finish_line(
+                    layout_context,
                     &mut ifc.current_nesting_level,
                     containing_block,
                     ifc.inline_position,
@@ -355,12 +356,13 @@ impl InlineFormattingContext {
 impl Lines {
     fn finish_line(
         &mut self,
+        layout_context: &LayoutContext,
         top_nesting_level: &mut InlineNestingLevelState,
         containing_block: &ContainingBlock,
         line_content_inline_size: Length,
     ) {
         let mut line_contents = std::mem::take(&mut top_nesting_level.fragments_so_far);
-        let line_block_size = std::mem::replace(
+        let mut line_block_size = std::mem::replace(
             &mut top_nesting_level.max_block_size_of_fragments_so_far,
             Length::zero(),
         );
@@ -373,6 +375,8 @@ impl Lines {
             .style
             .writing_mode
             .line_left_is_inline_start();
+
+        // Horizontal alignment.
         let text_align = match containing_block.style.clone_text_align() {
             TextAlignKeyword::Start => TextAlign::Start,
             TextAlignKeyword::Center => TextAlign::Center,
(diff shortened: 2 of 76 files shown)


ferjm:
This is only implementing `baseline` and `middle` alignment and it causes a lot of tests to fail. I am currently looking at avoiding all these test failures.

ferjm:
>  I am currently looking at avoiding all these test failures.

Many of them are due to the missing `top` alignment support, which is used in many test references.

ferjm:
To clarify my previous comment. All these tests are passing on master because we are aligning elements at the top by default. With these patches elements are baseline aligned by default, as the spec mandates.

tests/wpt/mozilla/meta-layout-2020/mozilla/localeCompare.html.ini
```diff
@@ -1,4 +1,2 @@
 [localeCompare.html]
   bug: https://github.com/servo/servo/issues/25802
-  [localeCompare should return the same as other browsers, even though it's implementation-dependent]
-    expected: FAIL
```
jdm, line 4: This change will need to be reverted. It has to do with locale settings on the CI machine.

ferjm:
r? @SimonSapin 

This introduces several new failures and the algorithm is not ideal, but it's a start. So any feedback will be highly appreciated. Thanks!

SimonSapin:
> This introduces several new failures

Could you sample a few and check if they were previously passing for the wrong reasons, or if this PR indeed regressed something?

ferjm:
Unfortunately, most of them are regressions. I believe the reason for many of the regressions is that relatively positioned elements are not properly aligned, but I couldn't find a fix for them yet. Any pointers would be highly appreciated.

jdm:
@SimonSapin Feedback ping?

SimonSapin:
The general approach seems to be: start with a guess for the baseline’s position and the line’s total height, then position fragments one by one, and if the guess turned out wrong adjust it and start over. There’s also a "correction growth factor" with magic value that I don’t fully understand, but its existence appears to be related to the guessing/adjusting.

Two issues with this approach are:

* Starting over makes the algorithm potentially take quadratic time, which shouldn’t be necessary
* With the "correction growth factor", results appear to be approximate. We should be able to find exact results (modulo `f32` precision)

I think both can be solved with a two pass algorithm.

During the first pass, the baseline position and the line’s overall height are unknown. Fragments are categorized into three "groups". Within each group, the vertical position of each fragment is computed relative to some reference point (on the vertical/block axis, I wanted to say reference line but that term is overloaded with "a line of text"). But the position of each group relative to other groups is not know yet. The groups are:

* Fragments with `vertical-align: top`. The reference point is the top of the line.
* Fragments with `vertical-align: bottom`. The reference point is the bottom of the line. Note that the position of the bottom of the line is not known yet. In this first pass we’re only positioning things relative to that unknown.
* Everything else. The reference point is the baseline **of the parent**.
  * Inline elements can be nested, so the baseline of the parent box/fragment is not necessarily the same as that of 
  * For example, `vertical-align: text-top` is based on “the top of the parent's content area” whose distance from the baseline is ascent + half-leading

This first pass also computes the height (block size) of every fragment.

Next, compute the respective three heights (block sizes) of the bounding boxes of each group. The height of the line is the largest of those. The "top" and "bottom" groups are positioned accordingly.

If the "everything else" group was the largest, there is only one possible position for that group within the line. If not, then the problem is under-constrained and this is the case that CSS 2 does not specify:

https://drafts.csswg.org/css2/visudet.html#line-height

> The inline-level boxes are aligned vertically according to their 'vertical-align' property. In case they are aligned 'top' or 'bottom', they must be aligned so as to minimize the line box height. If such boxes are tall enough, there are multiple solutions and CSS 2 does not define the position of the line box's baseline (i.e., the position of the strut, see below). 

Ideally it’d be nice to check what other browsers do, and if there’s interop do the same. But for now it’s fine to pick something and leave a FIXME comment.

**Then** we get to the second pass: now that we’ve determined the position each group we can go through each fragment again and set its "final" vertical position.

----

Note that the whole algorithm above needs to be **recursive**. Each fragment with `vertical-align: top` and `bottom` forms a sort of isolation boundary. The spec calls its contents an *aligned subtree* in https://drafts.csswg.org/css2/visudet.html#propdef-vertical-align. Such fragment are considered as one atomic thing by the algorithm. The same algorithm is applied recursively to each of their contents / aligned subtree.

----

There is *a lot* going on in this two-pass, recursive algorithm with three groups. It may be better to stick to a simplified version in this first PR and leave the rest for future PRs.

Namely, not implementing support for `vertical-align: top` or `vertical-align: bottom` yet. (They could either be mapped to `text-top` and `text-bottom` in the meantime, or be disabled in Stylo.) This removes the need for recursion, and reduces the three groups to just one. (Two passes are still needed though.) I suspect this would involve a lot less tricky details to get right.

mrobinson:
We landed support for vertical-align in #30902.

PR #26395, opened 2020-05-03
Move away from Promise.all way and check if we need to finish manually

### LINKED ISSUE #25442: Investigate if we can move away from Promise.all way for module script
Currently, we save a `promise` in each ModuleTree and then use it to do things asynchronously. When we're going to  fetch descendants for a module, we will merge all the descendant promises with a `Promise.all` so that we can _do something_ after all descendants fetched.

But, maybe we can do it better with just manually handling them with _tasks_ and the _Trusted_s?

Need to investigate that.

Also, not sure if we can avoid using the recursive way for checking descendants after switching to use tasks 🤔

Also see https://mozilla.logbot.info/servo/20200106#c16838760

CYBAI:
In the previous Promise.all way, we registered a promise for every
module script which means we will need to do many complex checkings like
"is this top level?" and it will make us much more difficult to understand
how the module script algorithm works.

In the new manual checking way, we will only register promises for top
level modules to notify its owner (e.g. the script element) to finish
the load. So, we can understand it much more easily and would be more
spec-aligned.

Also, I think the `Ready` and `FetchFailed` status are quite confusing
and we actually don't need them so they're removed in this patch. Then,
we will always go to `Finished` instead.

It would basically be following steps:

```
                                          +-----------------+
                                          | Failed to fetch | ----------+
+--------------+       +----------+     / +-----------------+           |
| Fetch module | ----> | Fetching | ---+                                v
+--------------+       +----------+     \ +---------+        +---------------------+
                                          | Fetched |        | Advance to Finished |
                                          +---------+        +---------------------+
                                               |                        ^
                                               v                        |
                                      +-------------------+             |
                                      | Fetch descendants | ----- if no descendants
                                      +-------------------+
                                               |
                                               V
                                     +----------------------+
                                     | Fetching Descendants |
                                     +----------------------+
```

In `Advance to Finished`, it means that module script is about to finished so it will

1. Notify all of its `ready` and `not finished` parents to finish
2. Link (instantiate) the module
3. Resolve its promise to notify owner(s) to finish

---

- [x] These changes fix #25442 
- [x] These changes do not require tests because it basically refactors the script module and should not break anything.

### DIFF
--- a/components/script/dom/htmlscriptelement.rs
+++ b/components/script/dom/htmlscriptelement.rs
@@ -870,52 +870,40 @@ impl HTMLScriptElement {
         let global = window.upcast::<GlobalScope>();
         let _aes = AutoEntryScript::new(&global);
 
-        if script.external {
-            let module_map = global.get_module_map().borrow();
+        let tree = if script.external {
+            global.get_module_map().borrow().get(&script.url).cloned()
+        } else {
+            global
+                .get_inline_module_map()
+                .borrow()
+                .get(&self.id.clone())
+                .cloned()
+        };
 
-            if let Some(module_tree) = module_map.get(&script.url) {
-                // Step 6.
-                {
-                    let module_error = module_tree.get_error().borrow();
-                    if module_error.is_some() {
-                        module_tree.report_error(&global);
-                        return;
-                    }
-                }
-
-                let module_record = module_tree.get_record().borrow();
-                if let Some(record) = &*module_record {
-                    let evaluated = module_tree.execute_module(global, record.handle());
-
-                    if let Err(exception) = evaluated {
-                        module_tree.set_error(Some(exception.clone()));
-                        module_tree.report_error(&global);
-                        return;
-                    }
+        if let Some(module_tree) = tree {
+            // Step 6.
+            {
+                let module_error = module_tree.get_rethrow_error().borrow();
+                let network_error = module_tree.get_network_error().borrow();
+                if module_error.is_some() && network_error.is_none() {
+                    module_tree.report_error(&global);
+                    return;
                 }
             }
-        } else {
-            let inline_module_map = global.get_inline_module_map().borrow();
 
-            if let Some(module_tree) = inline_module_map.get(&self.id.clone()) {
-                // Step 6.
-                {
-                    let module_error = module_tree.get_error().borrow();
-                    if module_error.is_some() {
-                        module_tree.report_error(&global);
-                        return;
-                    }
-                }
+            let record = module_tree
+                .get_record()
+                .borrow()
+                .as_ref()
+                .map(|record| record.handle());
 
-                let module_record = module_tree.get_record().borrow();
-                if let Some(record) = &*module_record {
-                    let evaluated = module_tree.execute_module(global, record.handle());
+            if let Some(record) = record {
+                let evaluated = module_tree.execute_module(global, record);
 
-                    if let Err(exception) = evaluated {
-                        module_tree.set_error(Some(exception.clone()));
-                        module_tree.report_error(&global);
-                        return;
-                    }
+                if let Err(exception) = evaluated {
+                    module_tree.set_rethrow_error(exception);
+                    module_tree.report_error(&global);
+                    return;
                 }
             }
         }
--- a/components/script/script_module.rs
+++ b/components/script/script_module.rs
@@ -12,7 +12,7 @@ use crate::dom::bindings::conversions::jsstring_to_str;
 use crate::dom::bindings::error::report_pending_exception;
 use crate::dom::bindings::error::Error;
 use crate::dom::bindings::inheritance::Castable;
-use crate::dom::bindings::refcounted::{Trusted, TrustedPromise};
+use crate::dom::bindings::refcounted::Trusted;
 use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::settings_stack::AutoIncumbentScript;
@@ -37,19 +37,19 @@ use crate::task::TaskBox;
 use crate::task_source::TaskSourceName;
 use encoding_rs::UTF_8;
 use hyper_serde::Serde;
+use indexmap::{IndexMap, IndexSet};
 use ipc_channel::ipc;
 use ipc_channel::router::ROUTER;
 use js::jsapi::Handle as RawHandle;
 use js::jsapi::HandleObject;
 use js::jsapi::HandleValue as RawHandleValue;
-use js::jsapi::{
-    CompileModule, ExceptionStackBehavior, GetModuleResolveHook, JSRuntime, SetModuleResolveHook,
-};
+use js::jsapi::{CompileModule, ExceptionStackBehavior};
+use js::jsapi::{GetModuleResolveHook, JSRuntime, SetModuleResolveHook};
 use js::jsapi::{GetRequestedModules, SetModuleMetadataHook};
-use js::jsapi::{GetWaitForAllPromise, ModuleEvaluate, ModuleInstantiate};
 use js::jsapi::{Heap, JSContext, JS_ClearPendingException, SetModulePrivate};
 use js::jsapi::{JSAutoRealm, JSObject, JSString};
 use js::jsapi::{JS_DefineProperty4, JS_NewStringCopyN, JSPROP_ENUMERATE};
+use js::jsapi::{ModuleEvaluate, ModuleInstantiate};
 use js::jsapi::{SetModuleDynamicImportHook, SetScriptPrivateReferenceHooks};
 use js::jsval::{JSVal, PrivateValue, UndefinedValue};
 use js::rust::jsapi_wrapped::{GetRequestedModuleSpecifier, JS_GetPendingException};
@@ -57,9 +57,7 @@ use js::rust::jsapi_wrapped::{JS_GetArrayLength, JS_GetElement};
 use js::rust::transform_u16_to_source_text;
 use js::rust::wrappers::JS_SetPendingException;
 use js::rust::CompileOptionsWrapper;
-use js::rust::IntoHandle;
-use js::rust::RootedObjectVectorWrapper;
-use js::rust::{Handle, HandleValue};
+use js::rust::{Handle, HandleValue, IntoHandle};
 use mime::Mime;
 use net_traits::request::{CredentialsMode, Destination, ParserMetadata};
 use net_traits::request::{Referrer, RequestBuilder, RequestMode};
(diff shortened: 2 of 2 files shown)


CYBAI:
also cc @Manishearth

CYBAI:
@bors-servo try=wpt

I also refactored `errors` for module script in 2e8381d. Does that make sense to have that commit in this PR? or I can send another one which will depend on this one.

Also, I can't reproduce the failure in previous try locally... I suspect there will be some intermittent failures 😕 so, just try again.

CYBAI:
@bors-servo try=wpt

- Added some more steps to avoid intermittent failures. Let's see if it works well!

CYBAI:
btw, is it worth writing the ASCII graph in PR description into `script_module` file as documentation comments 👀?

components/script/script_module.rs
```diff
@@ -657,6 +639,161 @@ impl ModuleTree {
...
+                    .extend(requested_urls.clone());
+
+                let parent_identities = self.parent_identities.borrow();
+
+                for parent_identity in parent_identities.iter() {
+                    match parent_identity {
+                        ModuleIdentity::ScriptId(_) => continue,
+                        ModuleIdentity::ModuleUrl(parent_url) => {
+                            if requested_urls.contains(&parent_url.clone()) {
+                                self.advance_finished_and_link(&global);
```
Manishearth, line 659: This might be quadratic if there are multiple matching URLs: we should set a boolean and run this algorithm only once outside the parent_identities loop

components/script/script_module.rs
```diff
@@ -657,6 +639,161 @@ impl ModuleTree {
...
+                                requested_urls.remove(&parent_url.clone());
+                            }
+                        },
+                    }
+                }
+
+                for requested_url in requested_urls {
+                    // https://html.spec.whatwg.org/multipage/#internal-module-script-graph-fetching-procedure
+                    let visited_urls = {
```
Manishearth, line 668: why is this sub-block necessary?

components/script/script_module.rs
```diff
@@ -657,6 +639,161 @@ impl ModuleTree {
...
+            Err(error) => {
+                self.set_error(Some(error));
+                self.advance_finished_and_link(&global);
+            },
+        }
+    }
+
+    /// https://html.spec.whatwg.org/multipage/#fetch-the-descendants-of-and-link-a-module-script
+    /// step 4-7.
+    fn advance_finished_and_link(&self, global: &GlobalScope) {
```
Manishearth, line 700: This function seems to be doing a lot more than steps 4-7, please add comments

components/script/script_module.rs
```diff
@@ -697,158 +834,64 @@ impl ModuleOwner {
...
-                } else {
-                    let module_tree = {
-                        let inline_module_map = global.get_inline_module_map().borrow();
-                        inline_module_map
-                            .get(&script.root().get_script_id())
-                            .unwrap()
-                            .clone()
+                let load = {
+                    let module_tree = match module_identity.clone() {
+                        ModuleIdentity::ModuleUrl(script_src) => {
```
Manishearth, line 831: It seems like "get the module tree for an identity" is an operation that's performed often, factor it out into a method?

Manishearth (commented):
Seems overall okay, but i'm having a hard time understanding how the resolution works: could you add more comments to the fetch modules and advance_finished_and_link functions?


also it seems like the advance_finished_and_link function, which contains a bunch of loops, is called a lot, but i'm unsure how necessary that is since it's unclear to me what the function is doing

components/script/script_module.rs
CYBAI: oh, I was thinking this will cause a double borrow panic but no, because of the clone, it won't! thanks for noticing this! will fix it!

components/script/script_module.rs
```diff
@@ -657,6 +639,161 @@ impl ModuleTree {
...
+                self.set_error(Some(error));
+                self.advance_finished_and_link(&global);
+            },
+        }
+    }
+
+    /// https://html.spec.whatwg.org/multipage/#fetch-the-descendants-of-and-link-a-module-script
+    /// step 4-7.
+    fn advance_finished_and_link(&self, global: &GlobalScope) {
+        // FIXME(cybai): try to find a better way for checking `ready` descendants
```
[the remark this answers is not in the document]
CYBAI: @Manishearth thanks for your review first!

I found most your comments are about confusion in this `advance_finished_and_link` and that's because we currently save a set of parent identities and descendant sets.

So, when we're trying to finish a module, we will need to `notify` its parent (the parent for loop) or `notify` its descendants (when it's a cyclic module) to finish.

Instead of doing so many loops, I think I'd like to try doing the chromium way. They make a Module [hold a counter](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/loader/modulescript/module_tree_linker.h;l=120) and increment when fetching more descendants and decrement when the descendants fetched.

I think using this way should make this function less scary. We might still need to loop parents to notify them to finish but at least it should be easier to understand and no need to do recursive steps. (also, we can remove this FIXME)

Please let me do more works on this 🙏 

Will try to ping you for review again when it's ready! Thanks!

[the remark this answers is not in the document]
Manishearth: This seems good! Let me know when ready!

CYBAI:
@Manishearth this is ready for review now!

@jdm I saw there are some commits in the `dynamic-module` branch about fixing borrowing; I think some of them are also fixed in this PR; maybe you will be interested in taking a look as well 👀?

components/script/script_module.rs
```diff
@@ -230,104 +225,120 @@ impl ModuleTree {
...
-        discovered_urls: &mut HashSet<ServoUrl>,
-    ) -> bool {
-        discovered_urls.insert(module_tree.url.clone());
+    pub fn remove_incomplete_fetch_url(&self, dependency: ServoUrl) {
+        self.incomplete_fetch_urls.borrow_mut().remove(&dependency);
+    }
 
-        let descendant_urls = module_tree.descendant_urls.borrow();
+    // Find circular dependencies in non-recursive way
+    pub fn find_circular_dependencies(&self, global: &GlobalScope) -> IndexSet<ServoUrl> {
```
CYBAI, line 261: This function is inspired by [this blogpost](https://breakingcode.wordpress.com/2013/03/11/an-example-dependency-resolution-algorithm-in-python/). Should I link it as a (documentation) comment here?

Manishearth: Yes please

CYBAI: Added in 17ff240! Thanks!

Manishearth:
@CYBAI did you end up figuring out how chrome avoids needing a GC?

CYBAI:
> @CYBAI did you end up figuring out how chrome avoids needing a GC?

sorry, no 😭 I do try to send a mail to Chrmoium `module-dev` team but haven't got reply yet.

components/script/dom/htmlscriptelement.rs
```diff
@@ -870,53 +870,42 @@ impl HTMLScriptElement {
...
+                return;
+            }
+        }
 
-                let module_record = module_tree.get_record().borrow();
-                if let Some(record) = &*module_record {
-                    let evaluated = module_tree.execute_module(global, record.handle());
+        let module_record = module_tree.get_record().borrow();
+        if let Some(record) = &*module_record {
+            let evaluated = module_tree.execute_module(global, record.handle());
```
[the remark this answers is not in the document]
CYBAI: Thanks for the tip! I missed this one! Fixed it in ad082f3.

Just want to confirm if I understand it correctly. So, with [borrowing the tree as ref and mapping it to the handled record](https://github.com/jdm/servo/commit/108664df20d5eaa429868ebdd62d9d29181360eb#diff-d171ba3dd3bb533163840f0c787a3095R840), we won't make the tree borrowed; instead, we will only have the raw record, is that correct 👀?

CYBAI:
@Manishearth will it be good for you to merge this and file a follow-up issue for avoiding the cyclic checking 👀?

components/script/dom/htmlscriptelement.rs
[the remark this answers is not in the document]
jdm: That is correct.

Manishearth:
The thing that makes me uneasy is that there's clearly something missing, but it's hard to find _where_ because this doesn't quite match the steps in the spec.

That said, that's not a change from what it used to be like, and this is still an improvement, so you can r=me after squashing!

We should file followups about making this closer match the spec text, and then looking into why cycles need to be explicitly collected.

CYBAI:
> The thing that makes me uneasy is that there's clearly something missing, but it's hard to find _where_ because this doesn't quite match the steps in the spec.

Yeah... I'm also confused why this doesn't match spec and result in a redundant cyclic check :( I will hope it can be fixed!

> That said, that's not a change from what it used to be like, and this is still an improvement, so you can r=me after squashing!

Thank you! I would be happy current improvement can be merged! and keep improving it! :)

> We should file followups about making this closer match the spec text, and then looking into why cycles need to be explicitly collected.

Sure! Let me file one for it!

CYBAI:
I squashed all commits into one and filed #26903 for the spec-compliant issue; hope it's clear enough 🙏 Thanks!

CYBAI:
@bors-servo retry

- intermittently webrender build failure on layout-2020 machine

CYBAI:
@bors-servo retry

```
  ▶ Unexpected subtest result in /_webgl/conformance/textures/misc/copy-tex-image-2d-formats.html:
  │ FAIL [expected PASS] WebGL test #47: should be 64,255,191,255\nat (0, 0) expected: 64,255,191,255 was 0,0,0,255
  │   → assert_true: should be 64,255,191,255
  │   → at (0, 0) expected: 64,255,191,255 was 0,0,0,255 expected true got false
  │ 
  │ reportTestResultsToHarness/<@http://web-platform.test:8000/_webgl/js/js-test-pre.js:123:24
  │ test@http://web-platform.test:8000/resources/testharness.js:535:30
  │ reportTestResultsToHarness@http://web-platform.test:8000/_webgl/js/js-test-pre.js:122:15
  │ testFailed@http://web-platform.test:8000/_webgl/js/js-test-pre.js:249:31
  │ checkCanvasRect/<@http://web-platform.test:8000/_webgl/js/webgl-test-utils.js:1269:19
  │ checkCanvasRectColor@http://web-platform.test:8000/_webgl/js/webgl-test-utils.js:1225:20
  │ checkCanvasRect@http://web-platform.test:8000/_webgl/js/webgl-test-utils.js:1257:23
  │ checkCanvas@http://web-platform.test:8000/_webgl/js/webgl-test-utils.js:1287:18
  │ testCopyTexImage2D@http://web-platform.test:8000/_webgl/conformance/textures/misc/copy-tex-image-2d-formats.html:188:7
  │ testFormats@http://web-platform.test:8000/_webgl/conformance/textures/misc/copy-tex-image-2d-formats.html:133:23
  └ @http://web-platform.test:8000/_webgl/conformance/textures/misc/copy-tex-image-2d-formats.html:99:12
```

is this same as #26862?

CYBAI:
@bors-servo retry

- #26855

PR #26398, opened 2020-05-03
Add a fast path in Element::SetInnerHTML when the value is small and trivial text

### LINKED ISSUE #25892: Calling innerHTML setter with text-only data is inefficient
babylon's FPS counter appears to be updated with with innerHTML setter, which goes to a _lot_ of extra work to end up attaching a new text node to the element. While it would be nice use the textContent setter instead, we should see if there's a way to optimize the current setter for short strings that don't contain `<`.

Eijebong:
Inspired from gecko which has a similar fast path. This makes innerHTML
more than 10 times faster for this case.

Fixes #25892

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -2449,8 +2449,6 @@ impl ElementMethods for Element {
 
     /// <https://w3c.github.io/DOM-Parsing/#widl-Element-innerHTML>
     fn SetInnerHTML(&self, value: DOMString) -> ErrorResult {
-        // Step 1.
-        let frag = self.parse_fragment(value)?;
         // Step 2.
         // https://github.com/w3c/DOM-Parsing/issues/1
         let target = if let Some(template) = self.downcast::<HTMLTemplateElement>() {
@@ -2458,6 +2456,23 @@ impl ElementMethods for Element {
         } else {
             DomRoot::from_ref(self.upcast())
         };
+
+        // Fast path for when the value is small, doesn't contain any markup and doesn't require
+        // extra work to set innerHTML.
+        if !self.node.has_weird_parser_insertion_mode() &&
+            value.len() < 100 &&
+            !value
+                .as_bytes()
+                .iter()
+                .any(|c| matches!(*c, b'&' | b'\0' | b'<' | b'\r'))
+        {
+            Node::SetTextContent(&target, Some(value));
+            return Ok(());
+        }
+
+        // Step 1.
+        let frag = self.parse_fragment(value)?;
+
         Node::replace_all(Some(frag.upcast()), &target);
         Ok(())
     }
--- a/components/script/dom/htmlframesetelement.rs
+++ b/components/script/dom/htmlframesetelement.rs
@@ -4,6 +4,7 @@
 
 use crate::dom::bindings::codegen::Bindings::HTMLFrameSetElementBinding::HTMLFrameSetElementMethods;
 use crate::dom::bindings::codegen::Bindings::WindowBinding::WindowMethods;
+use crate::dom::bindings::inheritance::Castable;
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::document::Document;
 use crate::dom::htmlelement::HTMLElement;
@@ -33,12 +34,14 @@ impl HTMLFrameSetElement {
         prefix: Option<Prefix>,
         document: &Document,
     ) -> DomRoot<HTMLFrameSetElement> {
-        Node::reflect_node(
+        let n = Node::reflect_node(
             Box::new(HTMLFrameSetElement::new_inherited(
                 local_name, prefix, document,
             )),
             document,
-        )
+        );
+        n.upcast::<Node>().set_weird_parser_insertion_mode();
+        n
     }
 }
 
--- a/components/script/dom/htmlheadelement.rs
+++ b/components/script/dom/htmlheadelement.rs
@@ -37,10 +37,13 @@ impl HTMLHeadElement {
         prefix: Option<Prefix>,
         document: &Document,
     ) -> DomRoot<HTMLHeadElement> {
-        Node::reflect_node(
+        let n = Node::reflect_node(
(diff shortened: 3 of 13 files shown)


Eijebong:
`tests/html/test_dom_performance.html`
Before:



After:

Eijebong:
```
  ▶ Unexpected subtest result in /html/syntax/parsing/html5lib_innerHTML_tests_innerHTML_1.html:
  │ FAIL [expected PASS] html5lib_innerHTML_tests_innerHTML_1.html 2555d238e04f3d2853cfbc5f6dd366f82cf0e868
  │   → assert_equals: expected "#document\n| <head>\n| <body>" but got "#document"
  │ 
  │ test_fragment@http://web-platform.test:8000/html/syntax/parsing/test.js:217:16
  │ test_next/</<@http://web-platform.test:8000/html/syntax/parsing/test.js:330:29
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1988:25
  │ test_next/<@http://web-platform.test:8000/html/syntax/parsing/test.js:324:20
  └ step_timeout/<@http://web-platform.test:8000/resources/testharness.js:937:15
```

I have **no idea** what this test is even supposed to be. There's a 18391 character long line in there with what looks like sha1sums everywhere... :confused:

Eijebong:
My guess is that gecko also blacklists `HTMLSharedElement` but there's no such thing in servo.

https://searchfox.org/mozilla-central/source/dom/html/HTMLSharedElement.h#20

The documentation here is missing so I have no idea what they even are.

jdm:
Judging by https://searchfox.org/mozilla-central/rev/b9a814e53b3b6f5cb665a78f4777868e7a16bfcd/dom/html/HTMLSharedElement.cpp#273-293, HTMLSharedElement looks like a class that is used as the implementation for a bunch of HTML elements that each have one or two custom behaviours to reduce code duplication.

jdm:
I suspect that you're missing weird parsing mode for the head element, judging by https://searchfox.org/mozilla-central/rev/b9a814e53b3b6f5cb665a78f4777868e7a16bfcd/dom/html/HTMLSharedElement.h#25-27 and the test contents being (pre-decoded) `"%23document%0A%7C%20%3Chead%3E%0A%7C%20%3Cbody%3E"`.

Eijebong:
I was actually missing it for HtmlHtmlElement. With my current patch the test seems to be ok. We'll see if something else is broken

Eijebong:
>   ▶ TIMEOUT [expected OK] /_webgl/conformance/glsl/misc/struct-nesting-of-variable-names.html

Hopefully unrelated :p

jdm:
I argue that you are _also_ missing it from HTMLHeadElement given the code that I linked :)

Eijebong:
Oh wow I somehow missed those lines when I read it... Will add for htmlheadelement

jdm:
@bors-servo r+
Thanks!

components/script/dom/element.rs
```diff
@@ -2449,15 +2449,27 @@ impl ElementMethods for Element {
...
             DomRoot::upcast(template.Content())
         } else {
             DomRoot::from_ref(self.upcast())
         };
+
+        // Fast path for when the value is small, doesn't contain any markup and doesn't require
+        // extra work to set innerHTML.
+        if !self.node.has_weird_parser_insertion_mode() &&
+            value.len() < 100 &&
+            !&['<', '&', '\0', '\r'].iter().any(|&c| value.contains(c))
```
emilio, line 2464: Would be nice to not go through `value` four times unconditionally.

Eijebong: Do you mean iterating over value and checking if current character is in `<&\0\r` ?

Eijebong: I can change it to `value.chars().any(|c| ['&', '\0', '<', '\r'].contains(&c))` if you think that's more efficient

jdm: That probably does end up generating code that's easier to optimize.

Eijebong: So I did a bit of benchmarking. What I'm proposing is faster on very small strings in release mode. However, in debug mode, perf is horrendous...

```
[eijebong@plutonium] ~   >>> rustc --test -O a.rs
[eijebong@plutonium] ~   >>> ./a --bench

running 8 tests
test current_100_length                  ... bench:          51 ns/iter (+/- 2)
test current_100_length_matching_center  ... bench:          14 ns/iter (+/- 3)
test current_10_length                   ... bench:          39 ns/iter (+/- 5)
test current_10_length_matching_center   ... bench:           7 ns/iter (+/- 0)
test proposed_100_length                 ... bench:          54 ns/iter (+/- 1)
test proposed_100_length_matching_center ... bench:          31 ns/iter (+/- 0)
test proposed_10_length                  ... bench:           5 ns/iter (+/- 0)
test proposed_10_length_matching_center  ... bench:           3 ns/iter (+/- 0)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out

[eijebong@plutonium] ~   >>> rustc --test a.rs
[eijebong@plutonium] ~   >>> ./a --bench

running 8 tests
test current_100_length                  ... bench:         267 ns/iter (+/- 17)
test current_100_length_matching_center  ... bench:         112 ns/iter (+/- 5)
test current_10_length                   ... bench:         275 ns/iter (+/- 28)
test current_10_length_matching_center   ... bench:         112 ns/iter (+/- 5)
test proposed_100_length                 ... bench:      19,971 ns/iter (+/- 1,149)
test proposed_100_length_matching_center ... bench:      10,284 ns/iter (+/- 1,097)
test proposed_10_length                  ... bench:       2,199 ns/iter (+/- 215)
test proposed_10_length_matching_center  ... bench:       1,312 ns/iter (+/- 63)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out

```

Here's the benchmark: https://gist.github.com/Eijebong/c04ed49a54c852f0e0340c5b467082a6

PR #26401, opened 2020-05-03
fixed BaseAudioContext.DecodeAudioData progress callback

### LINKED ISSUE #25840: Intermittent failure in /webaudio/the-audio-api/the-audiobuffersourcenode-interface/audiobuffersource-multi-channels.html
The failing test name contains the audio values that change each run, so it's hard to get a stables test result.

khodzha:
Gstreamer backend returns channel as single bit mask (ie 1, 2, 4, 8, 32 etc).
Progress callback was using this mask as plain channel index, thus storing decoded
audio in wrong channel.

(log2 conversion of int to int is subpar at best, but idk how to extract bit position in a clean way without a loop, any suggestions are welcome)

- [X] These changes fix #25840

### DIFF
--- a/components/script/dom/baseaudiocontext.rs
+++ b/components/script/dom/baseaudiocontext.rs
@@ -481,14 +481,17 @@ impl BaseAudioContextMethods for BaseAudioContext {
                         .unwrap()
                         .resize(channel_count as usize, Vec::new());
                 })
-                .progress(move |buffer, channel_pos| {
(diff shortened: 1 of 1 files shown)


ferjm:
Thanks!

@bors-servo r+

khodzha:
tbh maybe its better to replace plain ints with some kind of tuple struct to distinguish bitmask and channel index?

PR #26442, opened 2020-05-06
Fix the webidl for Headers

### LINKED ISSUE #26441: HeadersInit doesn't match specification
https://github.com/whatwg/fetch/commit/00fce0d96e799f5bc8bc53b30fcec56c4e0b948c removed the `Headers` value from the union, and this affects the correctness of our implementation. Code like https://github.com/servo/servo/blob/81b92b9c1a05eb526678c82e6cf28d665973ab14/tests/wpt/web-platform-tests/fetch/api/headers/headers-basic.html#L73-L75 overrides the iterator symbol of a Headers instance to return values and doesn't actually modify the contents of the DOM object. Then the Headers constructor receives the DOM object as an initializer and iterates over its internal contents, ignoring the iterator behaviour of the DOM interface. If we match the spec, the webidl codegen for sequences should do the right thing here instead.

Eijebong:
Make the HeadersInit type match the spec.

Fixes #26441

### DIFF
--- a/components/script/dom/headers.rs
+++ b/components/script/dom/headers.rs
@@ -198,20 +198,19 @@ impl HeadersMethods for Headers {
 }
 
 impl Headers {
+    pub fn copy_from_headers(&self, headers: DomRoot<Headers>) -> ErrorResult {
+        for (name, value) in headers.header_list.borrow().iter() {
+            self.Append(
+                ByteString::new(Vec::from(name.as_str())),
+                ByteString::new(Vec::from(value.as_bytes())),
+            )?;
+        }
+        Ok(())
+    }
+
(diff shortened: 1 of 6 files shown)


CYBAI:
seems more pass tests 👀 

```
  ▶ Unexpected subtest result in /fetch/range/general.any.worker.html:
  └ PASS [expected FAIL] Privileged header not allowed for guard type: request-no-cors
  ▶ Unexpected subtest result in /fetch/range/general.any.html:
  └ PASS [expected FAIL] Privileged header not allowed for guard type: request-no-cors
```

Eijebong:
There, updated expectations

PR #26446, opened 2020-05-06
Update webxr

Manishearth:
Brings in hand input

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,7 +630,7 @@ checksum = "4227269cec09f5f83ff160be12a1e9b0262dd1aa305302d5ba296c2ebd291055"
 dependencies = [
  "glob",
  "libc",
- "libloading",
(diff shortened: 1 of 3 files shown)


CYBAI:
```
Checking files for tidiness...
./Cargo.lock:1: duplicate versions for package `libloading`
	The following packages depend on version 0.5.2 from 'crates.io':
	The following packages depend on version 0.6.1 from 'crates.io':
Running the dependency licensing lint...
```

Manishearth:
Hmmm. I'm okay with allowing that, but i can also undo the dep bump in the patch (it's a proactive dep bump and not strictly required). We'll have to redo it when we switch back to openxrs master. Thoughts?

asajeffrey (approved):
Oh that's annoying re libloading. You can r=me anyway.

CYBAI:
@bors-servo retry

- #23290

PR #26452, opened 2020-05-06
Split animations and transitions into separate lists

### LINKED ISSUE #14419: Split `running_animations` into two lists, one for keyframes and one for transitions
Instead of this:

```rust
#[derive(Clone, Debug)]
pub enum Animation {
    Transition(OpaqueNode, UnsafeNode, f64, AnimationFrame, bool),
    Keyframes(OpaqueNode, Atom, KeyframesAnimationState),
}

// ...

pub fn update_animation_state(constellation_chan: &IpcSender<ConstellationMsg>,
                              script_chan: &IpcSender<ConstellationControlMsg>,
                              running_animations: &mut HashMap<OpaqueNode, Vec<Animation>>,
                              expired_animations: &mut HashMap<OpaqueNode, Vec<Animation>>,
                              new_animations_receiver: &Receiver<Animation>,
                              pipeline_id: PipelineId,
                              timer: &Timer) {
```

The data should be defined like this:

```rust
#[derive(Clone, Debug)]
pub enum Animation {
    Transition(TransitionAnimation),
    Keyframes(KeyframesAnimation),
}
#[derive(Clone, Debug)]
pub struct TransitionAnimation {
    opaque: OpaqueNode,
    node: UnsafeNode,
    start_time: f64,
    frame: AnimationFrame,
    is_expired: bool,
}
#[derive(Clone, Debug)]
pub struct KeyframesAnimation {
    opaque: OpaqueNode,
    name: Atom,
    state: KeyframesAnimationState,
}

// ...

pub fn update_animation_state(constellation_chan: &IpcSender<ConstellationMsg>,
                              script_chan: &IpcSender<ConstellationControlMsg>,
                              running_keyframes_animations: &mut HashMap<OpaqueNode, Vec<KeyframesAnimation>>,
                              running_transition_animations: &mut HashMap<OpaqueNode, Vec<TransitionAnimation>>,
                              expired_keyframes_animations: &mut HashMap<OpaqueNode, Vec<KeyframesAnimation>>,
                              expired_transition_animations: &mut HashMap<OpaqueNode, Vec<TransitionAnimation>>,
                              new_animations_receiver: &Receiver<Animation>,
                              pipeline_id: PipelineId,
                              timer: &Timer) {
```

Making this change (there are more spots that will need to be changed than the ones shown, of course) should reduce the rightward drift that comes from having to loop and then match (instead, it's just two loops). It should also save RAM, by only allocating the space we need instead of the space of the largest variant (in this case, Animation::Keyframes). The enum can still be used for sending over the animation channel.

mrobinson:
This change splits the list of animations and transitions, which are
almost always handled differently. It also renames
`ElementAnimationState` to `ElementAnimationSet` and establishes an
`AnimationState` for every transition and animation. This allows us to
stop continually reallocating lists every time a transition or animation
needs to be canceled.

Fixes #14419.

---

- [x] These changes fix #14419 

### DIFF
--- a/components/style/animation.rs
+++ b/components/style/animation.rs
@@ -26,283 +26,6 @@ use crate::Atom;
 use servo_arc::Arc;
 use std::fmt;
 
-/// This structure represents a keyframes animation current iteration state.
-///
-/// If the iteration count is infinite, there's no other state, otherwise we
-/// have to keep track the current iteration and the max iteration count.
-#[derive(Clone, Debug)]
-pub enum KeyframesIterationState {
-    /// Infinite iterations, so no need to track a state.
-    Infinite,
-    /// Current and max iterations.
-    Finite(f32, f32),
-}
-
-/// This structure represents wether an animation is actually running.
-///
-/// An animation can be running, or paused at a given time.
-#[derive(Clone, Debug)]
-pub enum KeyframesRunningState {
-    /// This animation is paused. The inner field is the percentage of progress
-    /// when it was paused, from 0 to 1.
-    Paused(f64),
-    /// This animation is actually running.
-    Running,
-}
-
-/// This structure represents the current keyframe animation state, i.e., the
-/// duration, the current and maximum iteration count, and the state (either
-/// playing or paused).
-// TODO: unify the use of f32/f64 in this file.
-#[derive(Clone)]
-pub struct KeyframesAnimationState {
-    /// The time this animation started at.
-    pub started_at: f64,
-    /// The duration of this animation.
-    pub duration: f64,
-    /// The delay of the animation.
-    pub delay: f64,
-    /// The current iteration state for the animation.
-    pub iteration_state: KeyframesIterationState,
-    /// Werther this animation is paused.
-    pub running_state: KeyframesRunningState,
-    /// The declared animation direction of this animation.
-    pub direction: AnimationDirection,
-    /// The current animation direction. This can only be `normal` or `reverse`.
-    pub current_direction: AnimationDirection,
-    /// The original cascade style, needed to compute the generated keyframes of
-    /// the animation.
-    pub cascade_style: Arc<ComputedValues>,
-}
-
-impl KeyframesAnimationState {
-    /// Given the current time, advances this animation to the next iteration,
-    /// updates times, and then toggles the direction if appropriate. Otherwise
-    /// does nothing.
-    pub fn iterate_if_necessary(&mut self, time: f64) {
-        if !self.iteration_over(time) {
-            return;
-        }
-
-        match self.running_state {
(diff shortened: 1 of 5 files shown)


components/style/animation.rs
```diff
@@ -35,47 +158,68 @@ pub enum KeyframesIterationState {
...
+
     /// The current animation direction. This can only be `normal` or `reverse`.
     pub current_direction: AnimationDirection,
+
     /// The original cascade style, needed to compute the generated keyframes of
     /// the animation.
     pub cascade_style: Arc<ComputedValues>,
+
+    /// Whether or not this animation is new and or has already been tracked
+    /// by the scrip thread.
```
jdm, line 204: nit: script

components/style/animation.rs
```diff
@@ -35,47 +158,68 @@ pub enum KeyframesIterationState {
...
+    pub is_new: bool,
```
jdm, line 205: Would be it be useful to make this part of AnimationState instead?

jdm (approved):
A bunch of the higher-level logic feels easier to follow now; thanks!

dralley:
So there's a difference between the design described in the issue and the final one, which is that now there's no "list of running animations" and "list of expired animations", just a single list of animations (and one for transitions).

Won't that incur an overhead for branch (mis)prediction and cache pollution as every animation in the animation list is visited, rather than just the active ones?

(Obviously it's all about tradeoffs, if states change frequently the cost might be higher, but I watched [this incredibly relevant talk about data-oriented CSS Animation engine design](https://www.youtube.com/watch?v=g1TsP60z2OQ&t=21m08s) recently and it seemed they had determined that maintaining separate lists was ideal for performance)

mrobinson:
@drally So ideally animations are ticking at 60FPS. In the most common case all animations and transitions are running. The animation style update for each frame currently clones the entire element style. The old version of this code would also unconditionally rebuild the lists of animations and transitions multiple times per animation tick. My thought was to make this as simple as possible now and then figure out if it makes sense to re-add a list for completed transitions (which would match the spec language) later.

dralley:
Makes sense, sorry for interrupting :)

components/style/animation.rs
mrobinson: I originally used this design, but new animations are almost always handled like running animations. My hope is that in a future change, I would totally eliminate this field when adding support for better handling of `transition-delay` and `animation-delay`.

mrobinson:
> Makes sense, sorry for interrupting :)

No worries. You raise some interesting points and I think it's worth exploring them.

components/style/animation.rs
mrobinson: I've fixed this typo. :sweat_smile:

jdm:
@bors-servo retry
* #23290

PR #26455, opened 2020-05-06
Properly set the url on the Response object when the response is an opaqueredirect

Eijebong:


jdm:
@bors-servo r+
Nice find!

jdm:
@bors-servo retry
* #26457

PR #26464, opened 2020-05-07
Move most animation processing to script

mrobinson:
This is preparation for sharing this code with layout_2020 and
implementing selective off-the-main-thread animations.

We still look for nodes not in the flow tree in the layout thread.

---


mrobinson:
@jdm r?

components/script/dom/bindings/trace.rs
```diff
@@ -586,6 +589,7 @@ unsafe_no_jsmanaged_fields!(MediaSessionActionType);
 unsafe_no_jsmanaged_fields!(MediaMetadata);
 unsafe_no_jsmanaged_fields!(WebrenderIpcSender);
 unsafe_no_jsmanaged_fields!(StreamConsumer);
+unsafe_no_jsmanaged_fields!(Animations);
```
jdm, line 592: We should derive JSTraceable on the Animations type instead.

components/script/dom/window.rs
```diff
@@ -1305,9 +1305,10 @@ impl WindowMethods for Window {
     }
 
     fn RunningAnimationCount(&self) -> u32 {
-        let (sender, receiver) = channel().unwrap();
-        let _ = self.layout_chan.send(Msg::GetRunningAnimations(sender));
-        receiver.recv().unwrap_or(0) as u32
+        self.document
+            .get()
+            .map(|d| d.animations().running_animation_count() as u32)
+            .unwrap_or(0)
```
jdm, line 1311: nit: `map_or(0, |d| ...)`

components/script/script_thread.rs
```diff
@@ -517,6 +520,8 @@ unsafe_no_jsmanaged_fields!(TaskQueue<MainThreadScriptMsg>);
 unsafe_no_jsmanaged_fields!(dyn BackgroundHangMonitorRegister);
 unsafe_no_jsmanaged_fields!(dyn BackgroundHangMonitor);
 
+unsafe_no_jsmanaged_fields!(RefCell<Vec<TransitionOrAnimationEvent>>);
```
jdm, line 523: We should derive JSTraceable on TransitionOrAnimationEvent instead.

jdm (changes requested):
Great work!

components/layout_thread/lib.rs
```diff
@@ -1413,8 +1398,13 @@ impl LayoutThread {
...
 
         // Create a layout context for use throughout the following passes.
-        let mut layout_context =
-            self.build_layout_context(guards.clone(), &map, origin, data.animation_timeline_value);
+        let mut layout_context = self.build_layout_context(
+            guards.clone(),
+            &map,
+            origin,
+            data.animation_timeline_value,
+            data.animations.clone(),
```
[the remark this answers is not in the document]
mrobinson: In this case animations is `Arc<RwLock<FxHashMap<OpaqueNode, ElementAnimationSet>>>`. This should just end up cloning the `Arc<...>` and not the entire set of animations. We are doing a lot of `Arc<...>` clones in layout_2020 as well. Does it still make sense to replace the value here?

components/script/animations.rs
```diff
@@ -0,0 +1,290 @@
...
+                    TransitionOrAnimationEventType::TransitionRun,
+                    0.,
+                );
+                transition.is_new = false;
+            }
+        }
+    }
+}
+
+impl MallocSizeOf for Animations {
```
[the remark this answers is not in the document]
mrobinson: I'm not sure this works because `MallocSizeOf` doesn't know what to do with `ServoArc`. I also cannot implement `MallocSizeOf` for `ServoArc` because it's defined outside the crate.

```
168 | impl MallocSizeOf for Arc<RwLock<FxHashMap<OpaqueNode, ElementAnimationSet>>> {
    | ^^^^^^^^^^^^^^^^^^^^^^-------------------------------------------------------
    | |                     |
    | |                     `style::servo_arc::Arc` is not defined in the current crate
    | impl doesn't use only types from inside the current crate
```

I have modified this to include the other member of the `Animations` though. Let me know if there's a better way to accomplish this.

components/script/dom/bindings/trace.rs
mrobinson: Oh yeah, that's much cleaner. Fixed.

components/script/dom/window.rs
mrobinson: Fixed.

components/script/script_thread.rs
mrobinson: Okay. I've done this an it's led to a small chain of cleanups due to implementing `JSTraceable` for `RefCell`.

components/script/script_thread.rs
```diff
@@ -826,20 +834,30 @@ impl ScriptThread {
         })
     }
 
-    pub unsafe fn note_newly_animating_nodes(
-        pipeline_id: PipelineId,
-        nodes: Vec<UntrustedNodeAddress>,
-    ) {
+    pub(crate) unsafe fn process_animations_update(mut update: AnimationsUpdate) {
```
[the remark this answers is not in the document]
mrobinson: That makes a lot of sense. I've added the comment.

components/script/animations.rs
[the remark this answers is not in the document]
jdm: Good point; Arcs always cause us problems. Let's add this comment above Animations:
`// Make sure to update the MallocSizeOf implementation when changing the contents of this struct.`

jdm (approved):

jdm:
One last comment fix and this can merge! Thanks!

mrobinson:
The failure looks like a real issue. I'll take a look tomorrow.
```
  ▶ TIMEOUT [expected OK] /dom/events/webkit-transition-end-event.html
  │ 
  ▶ Unexpected subtest result in /dom/events/webkit-transition-end-event.html:
  │ TIMEOUT [expected PASS] webkitTransitionEnd event listener should not trigger if an unprefixed listener also exists
  └   → Test timed out
  ▶ Unexpected subtest result in /dom/events/webkit-transition-end-event.html:
  └ NOTRUN [expected PASS] webkitTransitionEnd event listener should not trigger if an unprefixed event handler also exists
  ▶ Unexpected subtest result in /dom/events/webkit-transition-end-event.html:
  └ NOTRUN [expected FAIL] event types for prefixed and unprefixed transitionend event listeners should be named appropriately
  ▶ Unexpected subtest result in /dom/events/webkit-transition-end-event.html:
  └ NOTRUN [expected PASS] webkitTransitionEnd event listener is case sensitive
```

mrobinson:
@bors-servo try=wpt

Run wpt tests to see if this is a flake or a consistent failure.

mrobinson:
@bors-servo try=wpt

Run wpt tests to see if this is a flake or a consistent failure.

mrobinson:
@bors-servo r=jdm

I think this failure might be a rare flake. I'm not able to reproduce it locally at all (despite trying pretty hard). I'm going to try to pass this through bors again and see if there is something about the commit process that is causing it.

mrobinson:
@bors-servo retry
 * #24622

PR #26469, opened 2020-05-07
webaudio: Throw when setting invalid automationRate on AudioBufferSourceNode

khodzha:
The [webaudio spec](https://webaudio.github.io/web-audio-api/#AudioParam-attributes) specifies that some nodes have constraints certain constraints when setting `automationRate`. This change sets those constraints for `AudioBufferSourceNode`.

---

- [X] These changes (partially) fix https://github.com/servo/servo/issues/43407

### DIFF
--- a/components/script/dom/audioparam.rs
+++ b/components/script/dom/audioparam.rs
@@ -7,7 +7,7 @@ use std::sync::mpsc;
 
 use dom_struct::dom_struct;
 use servo_media::audio::graph::NodeId;
-use servo_media::audio::node::AudioNodeMessage;
+use servo_media::audio::node::{AudioNodeMessage, AudioNodeType};
 use servo_media::audio::param::{ParamRate, ParamType, RampKind, UserAutomationEvent};
 
 use crate::dom::baseaudiocontext::BaseAudioContext;
@@ -29,6 +29,9 @@ pub struct AudioParam {
     node: NodeId,
     #[ignore_malloc_size_of = "servo_media"]
     #[no_trace]
+    node_type: AudioNodeType,
+    #[ignore_malloc_size_of = "servo_media"]
+    #[no_trace]
     param: ParamType,
     automation_rate: Cell<AutomationRate>,
     default_value: f32,
@@ -40,6 +43,7 @@ impl AudioParam {
     pub fn new_inherited(
         context: &BaseAudioContext,
         node: NodeId,
+        node_type: AudioNodeType,
         param: ParamType,
         automation_rate: AutomationRate,
         default_value: f32,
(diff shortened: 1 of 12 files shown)


khodzha:
@Manishearth this is what I arrived at

components/script/dom/audioparam.rs
```diff
@@ -106,6 +112,22 @@ impl AudioParamMethods for AudioParam {
...
+        if automation_rate == AutomationRate::A_rate {
+            if self.node_type == AudioNodeType::AudioListenerNode {
+                // ignore
+                return;
+            }
+
+            if self.node_type == AudioNodeType::AudioBufferSourceNode &&
+                (self.param == ParamType::Detune || self.param == ParamType::PlaybackRate)
+            {
+                // TODO: this should raise InvalidState
```
Manishearth, line 124: mark the attribute as `[SetterThrows]` in webidl, make this function return `Fallible<()>`, and return `Err(Error::InvalidState)`

components/script/dom/audioparam.rs
```diff
@@ -106,6 +112,22 @@ impl AudioParamMethods for AudioParam {
 
     // https://webaudio.github.io/web-audio-api/#dom-audioparam-automationrate
     fn SetAutomationRate(&self, automation_rate: AutomationRate) {
+        if automation_rate == AutomationRate::A_rate {
+            if self.node_type == AudioNodeType::AudioListenerNode {
+                // ignore
+                return;
```
Manishearth, line 118: This is supposed to be ignored on the servo-media side, which already happens. We don't need this check.

components/script/dom/audioparam.rs
```diff
@@ -106,6 +112,22 @@ impl AudioParamMethods for AudioParam {
...
+
+            if self.node_type == AudioNodeType::AudioBufferSourceNode &&
+                (self.param == ParamType::Detune || self.param == ParamType::PlaybackRate)
+            {
+                // TODO: this should raise InvalidState
+                // but idk how to raise it
+                return;
+            }
+
+            // TODO: PannerNode should ignore AutomationRate changes
```
Manishearth, line 129: This happens in servo-media already

Manishearth:
The approach is overall correct.

khodzha:
updated with feedback

Manishearth (approved):
i'd be surprised if there weren't already tests that we currently fail

Manishearth:
@bors-servo try

if it turns out we don't have tests, yes, you should add one. otherwise update the expectations for the existing test

jdm:
The only test result changes are caused by https://github.com/servo/media/commit/628b2189dea6312e6cfd3c29fee01657c52a3560, which we haven't updated media to use yet. We should add a test for these changes.

mrobinson:
I rebased this change and verified that it causes a test to pass. WebAudio tests are currently not run so we can't see this on the CI and additionally it seems like the test was passing always due to some sort of some kind of issue with the test harness. I verified by observing the results manually though.

PR #26473, opened 2020-05-08
 Implement cross origin resource policy check 

Eijebong:
Also fixes an img load event bug that was making some test very racey.


jdm:
```
./components/net_traits/response.rs:64: derivable traits list is not in alphabetical order
	expected: Clone, Copy, Debug, Deserialize, MallocSizeOf, PartialEq, Serialize
	found: Clone, Copy, Debug, Deserialize, MallocSizeOf, Serialize, PartialEq
```

components/net/http_loader.rs
```diff
@@ -1251,7 +1252,81 @@ fn http_network_or_cache_fetch(
     // TODO: if necessary set response's range-requested flag
 
     // Step 9
-    // TODO: handle CORS not set and cross-origin blocked
+    // https://fetch.spec.whatwg.org/#cross-origin-resource-policy-header
```
jdm, line 1255: Let's link https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check instead.

components/net/http_loader.rs
```diff
@@ -1251,7 +1252,81 @@ fn http_network_or_cache_fetch(
...
+    fn cross_origin_resource_policy_check(request: &Request, response: &Response) -> bool {
```
jdm, line 1256: Let's return an Allowed/Blocked enum instead of a boolean.

components/net/http_loader.rs
```diff
@@ -1251,7 +1252,81 @@ fn http_network_or_cache_fetch(
...
+                    .unwrap_or("".into());
+                let current_url_origin_host = current_url_origin
+                    .host()
+                    .map(|h| h.to_string())
+                    .unwrap_or("".into());
+                let request_origin_reg = reg_suffix(&request_origin_host);
+                let current_url_reg = reg_suffix(&current_url_origin_host);
+
+                (request_origin.host() == current_url_origin.host() && request_origin_reg == "") ||
+                    (request_origin_reg != "" && request_origin_reg == current_url_reg)
```
jdm, line 1306: Could we extract the schemeless same site check into separate helper function? Annotating the checks with https://html.spec.whatwg.org/multipage/origin.html#schemelessly-same-site and step numbers would help me.

components/net/http_loader.rs
```diff
@@ -1251,7 +1252,81 @@ fn http_network_or_cache_fetch(
...
+            let schemeless_same_origin = if !request_origin.is_tuple() &&
+                !current_url_origin.is_tuple() &&
+                *request_origin == current_url_origin
+            {
+                true
+            } else if request_origin.is_tuple() && current_url_origin.is_tuple() {
+                let request_origin_host = request_origin
+                    .host()
+                    .map(|h| h.to_string())
+                    .unwrap_or("".into());
```
jdm, line 1297: These two `unwrap_or("".into())` could be unwrap_or_default.

components/net/http_loader.rs
jdm, line 1306: I had to read this several times closely to realize what was confusing me about it. Let's write the second check in the same order as the first: first compare the registrable domains, then check that one of them is non-null.

jdm (changes requested):
This was nice and self-contained!

components/net_traits/image_cache.rs
```diff
@@ -17,7 +17,7 @@ use std::sync::Arc;
 /// Indicating either entire image or just metadata availability
 #[derive(Clone, Debug, Deserialize, MallocSizeOf, Serialize)]
 pub enum ImageOrMetadataAvailable {
-    ImageAvailable(#[ignore_malloc_size_of = "Arc"] Arc<Image>, ServoUrl),
+    ImageAvailable(#[ignore_malloc_size_of = "Arc"] Arc<Image>, ServoUrl, bool),
```
jdm, line 20: Let's turn this into:
```rust
ImageAvailable {
    #[...]
    image: Arc<Image>,
    url: ServoUrl,
    is_placeholder: bool,
}
```

jdm (changes requested):

PR #26475, opened 2020-05-08
Initiate redirect fetching iff the request has a location

### LINKED ISSUE #25606: WPT cookies/http_state tests are slow
Running on a machine that's much slower than taskcluster, I find that I have to set --timeout-multiplier as high as 4 to get through some tests consistently. This problem is not necessarily about cookies, as the tests are doing entire HTTP roundtrips against the cookie-setter.py.

If it is actually the cookies slowing us down, it may be that the cookie jar is getting deserialized and re-serialized too often; I get the impression from the code that we are passing around the entire cookie store even when we already know which one URL we want to look up.

CYBAI:
Without this checking, we will run into a TIMEOUT request because
http_redirect_fetch will return immediately when location doesn't exist.

---

- [x] These changes partially fix #25606

### DIFF
--- a/components/constellation/network_listener.rs
+++ b/components/constellation/network_listener.rs
@@ -7,7 +7,7 @@
 //! response is received, it is forwarded to the appropriate script thread.
 
 use crossbeam_channel::Sender;
-use http::header::LOCATION;
+use http::HeaderMap;
 use ipc_channel::ipc;
 use ipc_channel::router::ROUTER;
 use msg::constellation_msg::PipelineId;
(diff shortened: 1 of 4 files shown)


CYBAI:
@bors-servo try=wpt

- not sure if we will surprisingly fix some other TIMEOUT tests, so try first!

components/constellation/network_listener.rs
```diff
@@ -132,7 +134,7 @@ impl NetworkListener {
                         //
                         // Ideally the Fetch code would handle manual redirects on its own
                         self.initiate_fetch(None);
-                    },
+                    }
```
CYBAI, line 137: hmm, I suspect this is a regression from rustfmt? 🤔

components/constellation/network_listener.rs
```diff
@@ -99,7 +99,9 @@ impl NetworkListener {
                 };
 
                 match metadata.headers {
-                    Some(ref headers) if headers.contains_key(LOCATION) => {
+                    Some(ref headers)
+                        if headers.contains_key(LOCATION) && metadata.location_url.is_some() =>
```
jdm, line 103: We should be able to remove the `headers.contains_key` check entirely. This corresponds with step 5.7.4 of https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-fetch.

jdm (approved):

components/constellation/network_listener.rs
CYBAI: oh! thanks for pointing to this part of spec! looks like we should check if `location_url` is some and validate its scheme, will change it!

PR #26477, opened 2020-05-09
Add support for getComputedStyle() to layout_2020

### LINKED ISSUE #16255: Add a fast path to get computed styles from the script thread
CC @emilio 

We have `window.GetComputedStyle`, but it would be great to have a faster path.

mrobinson:
These changes add support for `getComputedStyle()` to layout_2020.

---


components/script/dom/cssstyledeclaration.rs
```diff
@@ -235,21 +238,88 @@ impl CSSStyleDeclaration {
...
-                    return DOMString::new();
-                }
-                let addr = node.to_trusted_node_address();
-                window_from_node(node).resolved_style_query(addr, self.pseudo.clone(), property)
+            CSSStyleOwner::Element(ref element) => element.upcast::<Node>(),
+        };
+
+        if !node.is_connected() {
+            // TODO: Node should be matched against the style rules of this window.
+            // Firefox is currently the only browser to implement this.
```
emilio, line 250: This is no longer true, and spec was changed to reflect this. So we can remove this.

components/script/dom/cssstyledeclaration.rs
```diff
@@ -235,21 +238,88 @@ impl CSSStyleDeclaration {
...
+        let addr = node.to_trusted_node_address();
+        window_from_node(node).resolved_style_query(addr, self.pseudo.clone(), property)
+    }
+
+    // Try to get the resolved style for a node without having to ask the layout
+    // thread. This is possible when we can simply return the computed value or
+    // we can calculate the used value without looking at layout data structures.
+    // Some properties resovled values compute are their computed values.
+    // See: https://drafts.csswg.org/cssom/#dom-window-getcomputedstyle
+    fn try_to_get_resolved_style(&self, node: &Node, property: &PropertyId) -> Option<String> {
```
emilio, line 267: This doesn't look good if there are pending style changes. What prevents you from using an outdated style in that case?

emilio: Answering myself: `.style()` / `.style_for_pseudo` do update style.

emilio: Why is this optimization valuable then, though? Is the idea that it avoids going through the layout thread if there are no pending changes? If so it should be documented.

components/script/dom/cssstyledeclaration.rs
```diff
@@ -235,21 +238,88 @@ impl CSSStyleDeclaration {
...
+
+        // The properties that we calculate below all resolve to the computed value
+        // when the element is display:none or display:contents.
+        let display = style.get_box().display;
+        if display.is_none() || display.is_contents() {
+            return Some(computed_style());
+        }
+
+        let positioned = style.get_box().position != Position::Static;
+        match longhand_id {
```
emilio, line 306: It's a bit unfortunate to have this list of properties duplicated in multiple places :/

components/script/dom/cssstyledeclaration.rs
mrobinson: Thanks for pointing that out. I've opened #26481 for this.

mrobinson:
Thank you both for the reviews. I think the second commit here (adding a fast path for `getComputedSyle`) was not well thought-out on my part, so I'm removing it and sticking with adding support to layout 2020. The fast path only makes sense when the node isn't dirty, but the change I proposed did an unconditional reflow (:upside_down_face: ), so it was a bit pointless in the end. I'll think about that part a bit more deeply and hopefully later come up with a better change that avoids the duplication of the list of "special" properties as well.

jdm:
```
4 unexpected results that are NOT known-intermittents:
  â–¶ Unexpected subtest result in /_mozilla/mozilla/calc.html:
  â”” PASS [expected FAIL] calc(0ch + 0px + 0pt + 0pc + 0in + 0cm + 0mm + 0rem + 0em + 0ex + 0% + 0vw + 0vh + 0vmin + 0vmax)


  â–¶ Unexpected subtest result in /_mozilla/mozilla/localeCompare.html:
  â”‚ FAIL [expected PASS] localeCompare should return the same as other browsers, even though it's implementation-dependent
  â”‚   â†’ assert_equals: expected -1 but got 1
  â”‚ 
  â”‚ @http://web-platform.test:8000/_mozilla/mozilla/localeCompare.html:8:16
  â”‚ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1975:25
  â”‚ test@http://web-platform.test:8000/resources/testharness.js:535:30
  â”” @http://web-platform.test:8000/_mozilla/mozilla/localeCompare.html:7:5


  â–¶ Unexpected subtest result in /css/css-ui/box-sizing-027.html:
  â”‚ FAIL [expected PASS] Check the resolved value of 'height' when box-sizing is  border-box.
  â”‚   â†’ assert_equals: expected "100px" but got "60px"
  â”‚ 
  â”‚ @http://web-platform.test:8000/css/css-ui/box-sizing-027.html:30:17
  â”‚ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1975:25
  â”‚ test@http://web-platform.test:8000/resources/testharness.js:535:30
  â”” @http://web-platform.test:8000/css/css-ui/box-sizing-027.html:27:5

  â–¶ Unexpected subtest result in /css/css-ui/box-sizing-027.html:
  â”‚ FAIL [expected PASS] Check the resolved value of 'width' when box-sizing is  border-box.
  â”‚   â†’ assert_equals: expected "100px" but got "60px"
  â”‚ 
  â”‚ @http://web-platform.test:8000/css/css-ui/box-sizing-027.html:25:17
  â”‚ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1975:25
  â”‚ test@http://web-platform.test:8000/resources/testharness.js:535:30
  â”” @http://web-platform.test:8000/css/css-ui/box-sizing-027.html:22:5


  â–¶ Unexpected subtest result in /_mozilla/mozilla/getComputedStyle.html:
  â”” PASS [expected FAIL] Element's resolved values
```

tests/wpt/mozilla/meta-layout-2020/mozilla/localeCompare.html.ini
```diff
@@ -1,4 +1,2 @@
 [localeCompare.html]
   bug: https://github.com/servo/servo/issues/25802
-  [localeCompare should return the same as other browsers, even though it's implementation-dependent]
-    expected: FAIL
```
jdm, line 4: You will need to revert the changes to this file.

mrobinson:
@bors-servo retry
 * #25522

PR #26480, opened 2020-05-10
Prefix `on` for function name of inline events

### LINKED ISSUE #26479: window / body / frameset onerror event handler should stringify to include all its parameters
Spec: whatwg/html#5514
Tests: web-platform-tests/wpt#23467

CYBAI:
While checking what needs to be done for the spec-update, I've noticed the logic of checking `is window-reflecting element (e.g. body and frameset)` is already handled by the `is` casting function.

However, we still failed to pass the tests because we're missing `on` prefix for inline functions.

I'm not sure if this patch is good enough (or maybe at least I need to add a comment for why adding `on` prefix?).

Besides, I checked [how Gecko handles](https://searchfox.org/mozilla-central/rev/8bc4e35c9bb47c1fe3131e6155d9f482e1efef9a/dom/events/EventListenerManager.cpp#1012-1022) and looks like they also just pass the atom directly.
But, the [generated atom](https://searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h#775) is prefixed with `on` which is correct to just pass it into the `CompileFunction`.

---

- [x] These changes fix #26479 


CYBAI:
@bors-servo try=wpt

- this will need to wait for next WPT sync for the tests but I'd like to see if this breaks any other tests 👀

CYBAI:
oh, didn't notice the tests are already in our tree.

CYBAI:
also ping for review 👀 cc @jdm or @Manishearth

jdm:
@bors-servo retry
* #23290

jdm:
@bors-servo retry
* #24622

PR #26491, opened 2020-05-12
Handle cancelAnimationFrame() when called within a requestAnimationFr…

### LINKED ISSUE #26251: Handle cancelAnimationFrame() when called within a requestAnimationFrame() callback
Spec change: https://github.com/immersive-web/webxr/pull/1005#event-3257813558

Relevant code: https://github.com/servo/servo/blob/b60c70bab6e787fff594296a0996e8aeb2d8ba7e/components/script/dom/xrsession.rs#L398-L414

Currently we take `self.raf_callback_list` out and put it on the stack, and then iterate through it.

Instead, we should have a second `current_raf_callback_list` list on the object, move the vector there, and use indexed iteration. Then, in `CancelAnimationFrame`, we check both lists.

huangjiahua:
…ame() callback

In order to handle `cancelAnimationFrame()` in the callback, the `raf_callback_list` is moved to 
`current_raf_callback_list`, and each callback is cloned in the iteration of `current_raf_callback_list` to avoid "borrowed twice". 

---

- [X] These changes fix #26251 

### DIFF
--- a/components/script/dom/xrsession.rs
+++ b/components/script/dom/xrsession.rs
@@ -74,6 +74,8 @@ pub struct XRSession {
     next_raf_id: Cell<i32>,
     #[ignore_malloc_size_of = "closures are hard"]
     raf_callback_list: DomRefCell<Vec<(i32, Option<Rc<XRFrameRequestCallback>>)>>,
+    #[ignore_malloc_size_of = "closures are hard"]
+    current_raf_callback_list: DomRefCell<Vec<(i32, Option<Rc<XRFrameRequestCallback>>)>>,
     input_sources: Dom<XRInputSourceArray>,
     // Any promises from calling end()
     #[ignore_malloc_size_of = "promises are hard"]
@@ -110,6 +112,7 @@ impl XRSession {
 
             next_raf_id: Cell::new(0),
             raf_callback_list: DomRefCell::new(vec![]),
+            current_raf_callback_list: DomRefCell::new(vec![]),
             input_sources: Dom::from_ref(input_sources),
             end_promises: DomRefCell::new(vec![]),
             ended: Cell::new(false),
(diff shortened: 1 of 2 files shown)


Manishearth:
@bors-servo r+

thanks!

Manishearth:
Would be nice if we could add a test for this behavior. I might do that later.

components/script/dom/xrsession.rs
```diff
@@ -395,7 +398,10 @@ impl XRSession {
         // Step 3: XXXManishearth handle inline session
 
         // Step 4-5
-        let mut callbacks = mem::replace(&mut *self.raf_callback_list.borrow_mut(), vec![]);
+        mem::replace(
```
Manishearth, line 401: I think here we should instead just swap current and raf_callback_list, and after the for loop we should clear current

Manishearth: This basically ensures we don't hold on to RAF callbacks longer than necessary

huangjiahua: OK! Is that `drop` necessary?

Manishearth: Yes, because otherwise the guard will be borrowed for too long. You can also do this by inserting an enclosing scope.

Manishearth:
Nice, this makes us pass some tests

```
  ▶ Unexpected subtest result in /webxr/xrSession_cancelAnimationFrame.https.html:
  └ PASS [expected FAIL] XRSession requestAnimationFrame callbacks can be unregistered with cancelAnimationFrame for immersive sessions
  ▶ Unexpected subtest result in /webxr/xrSession_cancelAnimationFrame.https.html:
  └ PASS [expected FAIL] XRSession requestAnimationFrame callbacks can be unregistered with cancelAnimationFrame for non-immersive sessions
```

Manishearth:
You'll also need to update test expectations for those two listed tests. Run `./mach test-wpt /webxr/xrSession_cancelAnimationFrame.https.html --log-raw log` and then `./mach update-wpt log`.

Manishearth:
Hmm, we should not be getting a crash with that test.

Manishearth:
Tested it locally, I'm not getting a crash.

Just delete `tests/wpt/metadata/webxr/xrSession_cancelAnimationFrame.https.html.ini` and push, we can land it then. Be sure to squash your commits!

huangjiahua:
OK. It seems that I use a system with no GUI to run these tests.

Manishearth:
@bors-servo r+

servo should be able to run headless just fine, but yeah there can be weird GL issues

PR #26493, opened 2020-05-12
Switch to single WebGPU thread

kunalmohan:
Replace the current-one thread per host in a browsing context group-implementation to a single thread across the browser. Remove some garbage code in request adapter on wgpu server-side.

r?@kvark @jdm 

---

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -246,9 +246,6 @@ struct BrowsingContextGroup {
     /// share an event loop, since they can use `document.domain`
     /// to become same-origin, at which point they can share DOM objects.
     event_loops: HashMap<Host, Weak<EventLoop>>,
-
-    /// The set of all WebGPU channels in this BrowsingContextGroup.
-    webgpus: HashMap<Host, WebGPU>,
 }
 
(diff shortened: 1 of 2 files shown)


components/constellation/constellation.rs
```diff
@@ -2180,58 +2177,32 @@ where
...
+    fn handle_request_wgpu_adapter(&mut self, request: FromScriptMsg) {
+        let adapter_request = if let FromScriptMsg::RequestAdapter(sender, options, ids) = request {
+            WebGPURequest::RequestAdapter {
+                sender,
+                options,
+                ids,
+            }
+        } else {
+            return warn!("Wrong message type in handle_request_wgpu_adapter");
         };
```
jdm, line 2189: I don't understand the purpose of this change. If we only handle one message, we should make this method only accept the values we need from that message and make the caller deal with that. If we intend to accept more messages, we should accept an enum of only those valid messages.

jdm (changes requested):
These changes generally look fine to me!

kunalmohan:
Closing because this is no longer needed.

PR #26503, opened 2020-05-13
Fix an issue with the rooting of animating nodes

### LINKED ISSUE #26498: Intermittent crash in /_mozilla/mozilla/transitionend_safety.html
```
  ▶ CRASH [expected OK] /_mozilla/mozilla/transitionend_safety.html
  │ 
  └ Stack trace for thread "ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }"
```
cc @mrobinson 

This is a bit scary, since it implies that the safety being tested is no longer actually safe.

mrobinson:
Make sure nodes removed from map of rooted animating nodes are rooted on
the stack before triggering event handlers. We also make sure not to
call `from_untrusted_node_address` on nodes that aren't guaranteed to be
rooted.

Fixes #26498.

---

- [x] These changes fix #26498.

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -2976,13 +2976,10 @@ impl ScriptThread {
     ///
     /// TODO(mrobinson): Add support for more events.
     fn handle_transition_or_animation_event(&self, event: &TransitionOrAnimationEvent) {
-        let js_runtime = self.js_runtime.rt();
-        let node = unsafe { from_untrusted_node_address(js_runtime, event.node) };
-
-        // We limit the scope of the borrow here, so that we don't maintain this borrow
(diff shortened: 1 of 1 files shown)


mrobinson:
This is a flaky crash, so there isn't a solid guarantee that this fixes the issue. I tried to be as careful as possible when confirming this though. Before this change the following test:

```
./mach test-wpt   /_mozilla/mozilla/transitionend_safety.html --repeat 100
```

always resulted in at least one crash. After the change, I could not reproduce the crash at all.

jdm:
Thanks for investigating and figuring out a way to manually test it!
@bors-servo r+

Issue #26512, opened 2020-05-13
Implement WebGL2RenderingContext texImage2D/texImage3D overloads for pixel unpack buffer


jdm:
https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.6

PR #26520, opened 2020-05-14
Add support for WebGL2 TexImage2D

imiklos:
Adds initial support for one of the WebGL2 `TexImage2D` call.

---

- [x] These changes fix part of #26512 

@mmatyas @zakorgy @jdm

### DIFF
--- a/components/script/dom/webgl2renderingcontext.rs
+++ b/components/script/dom/webgl2renderingcontext.rs
@@ -18,7 +18,9 @@ use crate::dom::bindings::root::{Dom, DomRoot, LayoutDom, MutNullableDom};
 use crate::dom::bindings::str::DOMString;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::htmlcanvaselement::HTMLCanvasElement;
-use crate::dom::webgl_validations::tex_image_2d::{TexStorageValidator, TexStorageValidatorResult};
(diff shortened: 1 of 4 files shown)


components/script/dom/webgl2renderingcontext.rs
```diff
@@ -2914,6 +2916,100 @@ impl WebGL2RenderingContextMethods for WebGL2RenderingContext {
...
+        format: u32,
+        type_: u32,
+        src_data: CustomAutoRooterGuard<ArrayBufferView>,
+        src_offset: u32,
+    ) -> Fallible<()> {
+        if self.bound_pixel_unpack_buffer.get().is_some() {
+            return Ok(self.base.webgl_error(InvalidOperation));
+        }
+
+        if type_ == constants::FLOAT_32_UNSIGNED_INT_24_8_REV && src_data.len() != 0 {
```
jdm, line 2938: We shouldn't check src_data.len() at all - this method only takes a non-null argument, so we should always report an error if this type is used in this method.

jdm (changes requested):
Looks good otherwise!

components/script/dom/webgl2renderingcontext.rs
imiklos: Thank you! I've updated the pull request with this change.

PR #26524, opened 2020-05-14
Fix file extension for Linux nightly

### LINKED ISSUE #26522: Linux nightly builds are named incorrectly
From matrix: `By the way, it seems that mach is naming the Linux .tar.gz file just .gz even though it's a tar, and the download page links to .tar.gz.`

This is almost certainly a regression from https://github.com/servo/servo/pull/26207/commits/03e6787bf912a143ba097eeb58f7be16b8337d83.

camelid:
Fix file extension for the Linux nightly download. It should be published as `servo-latest.tar.gz` but was instead published as `servo-latest.gz`.

---

- [ ] `./mach build -d` does not report any errors (doesn't change Rust code)
- [X] These changes fix #26522 (GitHub issue number if applicable)

- [X] These changes do not require tests because this is an infrastructure bug fix

### DIFF
--- a/python/servo/package_commands.py
+++ b/python/servo/package_commands.py
@@ -628,8 +628,8 @@ class PackageCommands(CommandBase):
             nightly_dir = 'nightly/{}'.format(platform)
             filename = nightly_filename(package, timestamp)
             package_upload_key = '{}/{}'.format(nightly_dir, filename)
-            extension = path.splitext(path.basename(package))[1]
-            latest_upload_key = '{}/servo-latest{}'.format(nightly_dir, extension)
+            extension = path.basename(package).partition('.')[2]
+            latest_upload_key = '{}/servo-latest.{}'.format(nightly_dir, extension)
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo retry
* #25522

camelid:
Why are the tests failing?

jdm:
We have many tests that intermittently fail for various reasons. Many of those failures are restricted to a particular test and those failures can be filtered out of the logs. Some failures occur deeper in the platform and can affect a much wider set of tests, so we have to manually retry those because we can't easily filter them out of the CI results.

camelid:
> We have many tests that intermittently fail for various reasons. Many of those failures are restricted to a particular test and those failures can be filtered out of the logs. Some failures occur deeper in the platform and can affect a much wider set of tests, so we have to manually retry those because we can't easily filter them out of the CI results.

Ah, thanks for the explanation!

PR #26525, opened 2020-05-14
Share single gpu_id_hub among all threads in a process

### LINKED ISSUE #25725: WebGPU IdentityHub could create the same ID when we have multiple threads
When we create an ID, we could have multiple threads that end up creating the same ID independently. Possible we need to share a single `gpu_id_hub` between all threads using `Arc` & `Mutex`

kunalmohan:
An id_hub for a content process is created when a new script thread is created. This is then cloned whenever a new globalscope object is created. The only exception being service workers because they exist in separate processes, for them a new hub is created.
r?@jdm @kvark 

---

- [X] These changes fix #25725 (GitHub issue number if applicable)


components/script/dom/bindings/trace.rs
```diff
@@ -544,7 +544,7 @@ unsafe_no_jsmanaged_fields!(WebGLTextureId);
 unsafe_no_jsmanaged_fields!(WebGLVertexArrayId);
 unsafe_no_jsmanaged_fields!(WebGLVersion);
 unsafe_no_jsmanaged_fields!(WebGLSLVersion);
-unsafe_no_jsmanaged_fields!(Identities);
+unsafe_no_jsmanaged_fields!(Arc<Mutex<Identities>>);
```
kvark, line 547: We should probably use `parking_lot::Mutex` here. Parking lot is already used in Servo, and it's technically better than the standard mutex, and the locking doesn't require `unwrap()`.

jdm:
@bors-servo retry
* #23290

PR #26527, opened 2020-05-15
layout_2020: Fix WebRender display list dumping

mrobinson:
We need to enable display list serialization before building the display
list for it to work properly.

---

- [x] These changes do not require tests because they just fix debug output.

### DIFF
--- a/components/layout_thread_2020/lib.rs
+++ b/components/layout_thread_2020/lib.rs
@@ -1302,15 +1302,19 @@ impl LayoutThread {
             fragment_tree.scrollable_overflow(),
         );
 
+        // `dump_serialized_display_list` doesn't actually print anything. It sets up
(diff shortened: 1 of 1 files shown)


SimonSapin:
If we print a display list before creating it, won’t this print an empty list?

mrobinson:
This method's name is a little confusing. It prepares the display list for printing a text version during `finalize`. If you call it after adding display list items, they aren't added to the serialized content buffer.

https://github.com/servo/webrender/blob/master/webrender_api/src/display_list.rs#L1092

SimonSapin:
Ah I see. Please add a comment on the `dump_serialized_display_list` call to explain that it doesn’t print anything now but enables printing when items are added later. r+ with that.

mrobinson:
@bors-servo retry
 * #25522

PR #26537, opened 2020-05-15
layout_2020: Implement support for backface-visibility

mrobinson:
---


components/layout_2020/display_list/mod.rs
```diff
@@ -65,9 +66,20 @@ impl<'a> DisplayListBuilder<'a> {
...
-    fn common_properties(&self, clip_rect: units::LayoutRect) -> wr::CommonItemProperties {
-        // TODO(gw): Make use of the WR backface visibility functionality.
-        wr::CommonItemProperties::new(clip_rect, self.current_space_and_clip)
+    fn common_properties(
+        &self,
+        clip_rect: units::LayoutRect,
+        style: &ComputedValues,
+    ) -> wr::CommonItemProperties {
+        // TODO(mrobinson): We should pass the hit test information here and not create
+        // so many hit test display items.
```
SimonSapin, line 75: The problem is that we want hit testing to be always present for each fragment, and always based on the border edge of the fragment. Some fragments do not have a border or background and thus do not generate a display item other than the hit-testing one. Some have backgrounds, but with the `background-clip` property such that the rectangle is not the one appropriate for hit testing.

We can still reduce the number of display items dedicated to hit testing when there happens to be another display item with the appropriate rectangle, but it’s not as straightforward as this comment suggests. Please reword it to indicate that.

components/layout_2020/style_ext.rs
```diff
@@ -361,6 +364,14 @@ impl ComputedValuesExt for ComputedValues {
         // TODO: We need to handle CSS Contain here.
         false
     }
+
+    fn get_webrender_primitive_flags(&self) -> wr::PrimitiveFlags {
+        if self.get_box().backface_visibility == BackfaceVisiblity::Visible {
+            wr::PrimitiveFlags::default()
+        } else {
+            wr::PrimitiveFlags::empty()
+        }
```
SimonSapin, line 373: Nit: please use `match` instead of `==` so that the compiler reminds us to update this code in case a new variant is ever added to the enum. (I suspect it is very unlikely that the `backface-visibility` property will ever get a new value, but making more common a pattern that can help elsewhere seems good.)

mrobinson: This is also fewer lines of code, so it makes a lot of sense.

components/layout_2020/display_list/mod.rs
mrobinson: Sure. I'll update the comment to be a bit clearer.

mrobinson:
Thanks for the review. I've pushed a new version of the branch that addresses the issue you've identified.

SimonSapin:
Looks good, thanks!

@bors-servo r+

mrobinson:
```
  ▶ PASS [expected FAIL] /css/css-transforms/composited-under-rotateY-180deg-clip-perspective.html
```

I'll fix the expectation and submit this again.

PR #26541, opened 2020-05-15
Validate copybuffertobuffer() + some spec update

kunalmohan:
The spec update includes renaming bindings to entries, adding CommandEncoderState and renaming GPUBufferSize to GPUSize64.

r?@jdm @kvark

---

### DIFF
--- a/components/script/dom/gpucommandencoder.rs
+++ b/components/script/dom/gpucommandencoder.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::dom::bindings::cell::DomRefCell;
+use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUSize64;
 use crate::dom::bindings::codegen::Bindings::GPUCommandEncoderBinding::{
     GPUCommandBufferDescriptor, GPUCommandEncoderMethods, GPUComputePassDescriptor,
 };
@@ -16,8 +17,22 @@ use crate::dom::gpucommandbuffer::GPUCommandBuffer;
 use crate::dom::gpucomputepassencoder::GPUComputePassEncoder;
 use dom_struct::dom_struct;
 use ipc_channel::ipc;
+use std::cell::Cell;
 use std::collections::HashSet;
+use webgpu::wgpu::resource::BufferUsage;
 use webgpu::{WebGPU, WebGPUCommandEncoder, WebGPURequest};
+
+const BUFFER_COPY_ALIGN_MASK: u64 = 3;
+
+// https://gpuweb.github.io/gpuweb/#enumdef-encoder-state
+#[derive(MallocSizeOf, PartialEq)]
+#[allow(dead_code)]
+pub enum GPUCommandEncoderState {
+    Open,
+    EncodingRenderPass,
+    EncodingComputePass,
+    Closed,
+}
 
 #[dom_struct]
 pub struct GPUCommandEncoder {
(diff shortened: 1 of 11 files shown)


components/script/dom/gpucommandencoder.rs
```diff
@@ -68,18 +92,52 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
         &self,
         _descriptor: &GPUComputePassDescriptor,
     ) -> DomRoot<GPUComputePassEncoder> {
-        GPUComputePassEncoder::new(&self.global(), self.channel.clone(), self.encoder)
+        *self.state.borrow_mut() = GPUCommandEncoderState::EncodingComputePass;
```
kvark, line 95: We'd need to have a state check here (i.e. it has to be `Open`).
We could pass down the expected state to `set_state` and then turn it into invalid if the expectation doesn't match.
Not sure if you want to do it here on in the follow-up

components/script/dom/gpucommandencoder.rs
```diff
@@ -68,18 +92,52 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
...
+            None => false,
+        };
+        valid &= match BufferUsage::from_bits(destination.usage()) {
+            Some(usage) => usage.contains(BufferUsage::COPY_DST),
+            None => false,
+        };
+        valid &= (*self.state.borrow() == GPUCommandEncoderState::Open) &&
+            source.valid() &&
+            destination.valid() &
+                !(size & 3 == 0) &
```
kvark, line 127: Let's define this alignment as a constant somewhere. We'll ideally need this right in wgpu-types or wgpu-core.

components/script/dom/gpucommandencoder.rs
```diff
@@ -68,18 +92,52 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
...
+                !(source_offset & 3 == 0) &
+                !(destination_offset & 3 == 0) &
+                (source.size() >= source_offset + size) &
+                (destination.size() >= destination_offset + size);
+        if source.id().0 == destination.id().0 {
```
kvark, line 132: let's add a TODO here saying that we may need to forbid this entirely, linking to https://github.com/gpuweb/gpuweb/issues/783

components/script/dom/gpucommandencoder.rs
```diff
@@ -68,18 +92,52 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
...
+            valid &= source_offset > destination_offset + size ||
+                source_offset + size < destination_offset;
+        }
+
+        if !valid {
+            return;
```
kvark, line 138: we'd want to mark the whole command encoder as invalid if something bad happens

components/script/dom/gpucommandencoder.rs
kunalmohan: I'll do it here itself.

components/script/dom/gpucommandencoder.rs
kunalmohan: I am not completely sure of what you're expecting here. Maybe something like this-
```
const BUFFER_ALIGN:u64 = 3;
...
!(size & BUFFER_ALIGN == 0)
```
?

kvark: yes, put `const BUFFER_COPY_ALIGN_MASK: u64 = 3` somewhere at the top of the module and use it here.

components/script/dom/gpucommandencoder.rs
kunalmohan: The spec doesn't mention it yet but do we need to record a validation error here? If so I'll add a  TODO.

kvark: Yes, we definitely do need to record an error in the current error scope here.
But that relies on the error scopes being exposed and implemented, so you don't have to do it all in this PR.

kvark (approved):
Thanks!
Don't forget to click "Resolve conversation" for items you address. This helps us see what is WIP and what not.

PR #26544, opened 2020-05-16
Introduce import.meta hook for module script

### LINKED ISSUE #26542: Module metadata hook is missing
https://mixedreality.mozilla.org/area-demo-service-task/ fails immedidately with:
```
[2020-05-15T15:03:16Z ERROR script::dom::bindings::error] Error at https://mixedreality.mozilla.org/area-demo-service-task/arosg.js:3:20 Module metadata hook not set
```
This comes from https://searchfox.org/mozilla-central/rev/4166c15e2a99a23a9b38ad62c9fdfe8e5448b354/js/src/builtin/ModuleObject.cpp#1697, which is triggered by https://github.com/servo/servo/blob/b9b569aed8d9fe82d6630de732f45c4b32108ae1/components/script/script_module.rs#L1066. We should port the [equivalent code](https://searchfox.org/mozilla-central/rev/4166c15e2a99a23a9b38ad62c9fdfe8e5448b354/dom/script/ScriptLoader.cpp#937-958) from Gecko instead.

CYBAI:
---
- [x] These changes fix #26542
- [x] There are tests for these changes under `module/import-meta` folder.

### DIFF
--- a/components/script/script_module.rs
+++ b/components/script/script_module.rs
@@ -49,6 +49,7 @@ use js::jsapi::{GetRequestedModules, SetModuleMetadataHook};
 use js::jsapi::{GetWaitForAllPromise, ModuleEvaluate, ModuleInstantiate};
 use js::jsapi::{Heap, JSContext, JS_ClearPendingException, SetModulePrivate};
 use js::jsapi::{JSAutoRealm, JSObject, JSString};
+use js::jsapi::{JS_DefineProperty4, JS_NewStringCopyN, JSPROP_ENUMERATE};
 use js::jsapi::{SetModuleDynamicImportHook, SetScriptPrivateReferenceHooks};
 use js::jsval::{JSVal, PrivateValue, UndefinedValue};
 use js::rust::jsapi_wrapped::{GetRequestedModuleSpecifier, JS_GetPendingException};
@@ -1063,7 +1064,7 @@ pub unsafe fn EnsureModuleHooksInitialized(rt: *mut JSRuntime) {
     }
 
     SetModuleResolveHook(rt, Some(HostResolveImportedModule));
-    SetModuleMetadataHook(rt, None);
+    SetModuleMetadataHook(rt, Some(HostPopulateImportMeta));
     SetScriptPrivateReferenceHooks(rt, None, None);
 
     SetModuleDynamicImportHook(rt, None);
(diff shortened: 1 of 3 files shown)


components/script/script_module.rs
```diff
@@ -1118,6 +1119,35 @@ unsafe extern "C" fn HostResolveImportedModule(
...
+    rooted!(in(cx) let url_string = JS_NewStringCopyZ(cx, base_url.as_str().as_ptr() as *const i8));
+
+    // Step 3.
+    JS_DefineProperty4(
+        cx,
+        meta_object,
+        "url".as_ptr() as *const i8,
+        url_string.handle().into_handle(),
+        JSPROP_ENUMERATE.into(),
+    )
```
CYBAI, line 1148: @jdm I've followed Gecko to use those functions from mozjs but I still can't pass the import meta test which means the `import.meta` doesn't have the `url` attribute.

Do I miss something to set the property 🤔?

image

components/script/script_module.rs
```diff
@@ -1118,6 +1119,35 @@ unsafe extern "C" fn HostResolveImportedModule(
...
+    let in_realm_proof = AlreadyInRealm::assert_for_cx(SafeJSContext::from_ptr(cx));
+    let global_scope = GlobalScope::from_context(cx, InRealm::Already(&in_realm_proof));
+
+    // Step 2.
+    let base_url = match (reference_private.to_private() as *const ModuleScript).as_ref() {
+        Some(module_data) => module_data.base_url.clone(),
+        None => global_scope.api_base_url(),
+    };
+
+    rooted!(in(cx) let url_string = JS_NewStringCopyZ(cx, base_url.as_str().as_ptr() as *const i8));
```
jdm, line 1139: JS_NewStringCopyZ requires the string to be null-terminated, so we should either use CString here or JS_NewStringCopyN so we can pass the length.

jdm (changes requested):
Ensuring the resulting strings don't contain random memory should help!

CYBAI:
@jdm fixing with what you pointed out, we can pass the test now!

> Ensuring the resulting strings don't contain random memory should help!

but, I still don't understand what you mean for `contain random memory` here 🤔 
does that mean, while using `JS_NewStringCopyZ`, it's possible to return an empty string which might cause different memory in each run?

btw, how can I know when I'll need a null-terminated string 🤔 ? thanks!

jdm:
If an API expects a null-terminated string and we pass a string constant without one, whatever values are present in memory following the string contents will be included until there's a null value encountered. That's why the test failed, because we ended setting a property named something like "urlZzta3#';(84, x?&".

In general, any C aoi that accepts a chat* argument that is intended to be a name or string value should be null terminated unless there's a length argument included. In the string copy API in particular, the Z suffix stands for "zero", which means null-terminated. The N suffix stands for "length" and accepts an argument that includes an end pointer.

jdm:
I suspect this needs a fmt, but otherwise it looks fine.

CYBAI:
> If an API expects a null-terminated string and we pass a string constant without one, whatever values are present in memory following the string contents will be included until there's a null value encountered. That's why the test failed, because we ended setting a property named something like "urlZzta3#';(84, x?&".
> 
> In general, any C aoi that accepts a chat* argument that is intended to be a name or string value should be null terminated unless there's a length argument included. In the string copy API in particular, the Z suffix stands for "zero", which means null-terminated. The N suffix stands for "length" and accepts an argument that includes an end pointer.

I see! That's very clear and I understand now! Thanks! :D

> I suspect this needs a fmt, but otherwise it looks fine.

Wait me a minute! Looks like I need to upgrade my local fmt 🤔

CYBAI:
oh, it's not a fmt error from rustfmt :P

jdm:
@bors-servo retry
#24622

PR #26546, opened 2020-05-16
check http_state in determine_request_referrer

### LINKED ISSUE #14506: determine_request_referrer seems to be missing "TLS-protected" checks
But note https://github.com/w3c/webappsec-referrer-policy/issues/82

splav:
Check https status inside determine_request_referrer.
---

- [X] These changes fix #14506 (GitHub issue number if applicable)

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -166,28 +166,65 @@ pub fn set_default_accept_language(headers: &mut HeaderMap) {
 }
 
 /// <https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-state-no-referrer-when-downgrade>
-fn no_referrer_when_downgrade_header(referrer_url: ServoUrl, url: ServoUrl) -> Option<ServoUrl> {
-    if referrer_url.scheme() == "https" && url.scheme() != "https" {
+fn no_referrer_when_downgrade_header(
+    referrer_url: ServoUrl,
+    url: ServoUrl,
+    https_state: HttpsState,
+) -> Option<ServoUrl> {
+    if https_state == HttpsState::Modern && !is_origin_trustworthy(url) {
         return None;
     }
     return strip_url(referrer_url, false);
 }
 
 /// <https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-strict-origin>
-fn strict_origin(referrer_url: ServoUrl, url: ServoUrl) -> Option<ServoUrl> {
-    if referrer_url.scheme() == "https" && url.scheme() != "https" {
+fn strict_origin(
+    referrer_url: ServoUrl,
+    url: ServoUrl,
+    https_state: HttpsState,
+) -> Option<ServoUrl> {
+    if https_state == HttpsState::Modern && !is_origin_trustworthy(url) {
         return None;
     }
     strip_url(referrer_url, true)
 }
 
 /// <https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-strict-origin-when-cross-origin>
-fn strict_origin_when_cross_origin(referrer_url: ServoUrl, url: ServoUrl) -> Option<ServoUrl> {
-    if referrer_url.scheme() == "https" && url.scheme() != "https" {
+fn strict_origin_when_cross_origin(
+    referrer_url: ServoUrl,
+    url: ServoUrl,
+    https_state: HttpsState,
+) -> Option<ServoUrl> {
+    let same_origin = referrer_url.origin() == url.origin();
+    if same_origin {
+        return strip_url(referrer_url, false);
+    }
+    if https_state == HttpsState::Modern && !is_origin_trustworthy(url) {
         return None;
     }
-    let cross_origin = referrer_url.origin() != url.origin();
-    strip_url(referrer_url, cross_origin)
+    strip_url(referrer_url, true)
+}
+
+/// <https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy>
+fn is_origin_trustworthy(url: ServoUrl) -> bool {
+    match url.origin() {
+        // Step 1
+        ImmutableOrigin::Opaque(_) => false,
+        ImmutableOrigin::Tuple(_, _, _) => {
+            // Step 3
+            if url.scheme() == "https" || url.scheme() == "wss" {
+                return true;
+            }
+            // Step 4-5 TODO
+            // Step 6
+            if url.scheme() == "file" {
+                return true;
+            }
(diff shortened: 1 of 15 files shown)


splav:
Sadly, it seems there are no checks for HttpsState::Deprecated value

components/net/http_loader.rs
```diff
@@ -166,28 +166,65 @@ pub fn set_default_accept_language(headers: &mut HeaderMap) {
...
+            // Step 6
+            if url.scheme() == "file" {
+                return true;
+            }
+            // Step 7-8 TODO
+            // Step 9
+            false
+        },
+    }
 }
```
CYBAI, line 228: IIRC, we've had this function in `urlhelper` so you can just use `UrlHelper::is_origin_trustworthy`.

https://github.com/servo/servo/blob/0a00ea3db3c2353c78435030040fd3f57f5f4497/components/script/dom/urlhelper.rs#L75-L94

splav:
Thank you!

Then I think:
1. It should be moved to somewhere in the net, may be there is something
like net/url helpers? What is the best place?
2. The check for local ip is incorrect, I'll fix it using Rust std Ipv*Addr
structs methods.

---------------------------------------------------------
Александров Сергей Васильевич

вс, 17 мая 2020 г., 5:42 cybai <notifications@github.com>:

CYBAI:
> 1. It should be moved to somewhere in the net, may be there is something like net/url helpers? What is the best place?

Yes, agree! @gterzian tried to move it as a method of `ServoUrl` in #26317 ; maybe that will be a good place?

See https://github.com/servo/servo/pull/26317/files#diff-cac78be29c466ffe8160ad29af53e214R173-R191

> 2. The check for local ip is incorrect, I'll fix it using Rust std Ipv*Addr structs methods.

👍

splav:
Fixed is_origin_trustworthy + moved in to ServoUrl

jdm (changes requested):
We're missing initializing the request's https state in [Document::fetch_async](https://github.com/servo/servo/blob/619e0bceaf750ec862169a6a02cfbf82ba9be0a8/components/script/dom/document.rs#L1760).

We should also make `net_request_from_global` in component/script/fetch.rs initialize the https_state member; this might require us to:
* add a `https_state` method on GlobalScope
* make [load_whole_resource](https://github.com/servo/servo/blob/master/components/script/dom/dedicatedworkerglobalscope.rs#L378-L396) return the HTTPS state
* update the global scope's https state based on the result of the load

Finally we should add the request's HTTPS state as an argument to load_whole_resource and set the Request's https_state member before starting the request.

components/script/fetch.rs
```diff
@@ -128,6 +129,7 @@ fn request_init_from_request(request: NetTraitsRequest) -> RequestBuilder {
         parser_metadata: request.parser_metadata,
         initiator: request.initiator,
         csp_list: None,
+        https_state: HttpsState::None,
```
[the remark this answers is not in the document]
splav: Yes, missed that NetTraitsRequest means 'Request from net_traits'

splav:
I want to believe that all requested changes are fixed now.

splav:
Forgot unit tests, fixing.

components/script/dom/dedicatedworkerglobalscope.rs
```diff
@@ -395,6 +395,7 @@ impl DedicatedWorkerGlobalScope {
                     Ok((metadata, bytes)) => (metadata, bytes),
                 };
                 scope.set_url(metadata.final_url);
+                global_scope.set_https_state(metadata.https_state);
```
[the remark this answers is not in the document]
splav: Done, essentially it makes the worker global scope inherit parent global scope https state.

components/script/dom/dedicatedworkerglobalscope.rs
```diff
@@ -375,6 +375,8 @@ impl DedicatedWorkerGlobalScope {
                 let scope = global.upcast::<WorkerGlobalScope>();
                 let global_scope = global.upcast::<GlobalScope>();
 
+                global_scope.set_https_state(current_global.get_https_state());
```
jdm, line 378: I'm 99% confident that this doesn't build, since current_global can't be moved between threads. We need to get the https state value before spawning the new thread instead.

splav: Yes, fixed that. Strangely it did build. May be some caching... Also sometimes rustc crashes and builds fine after rerun.

jdm:
Huh, I'm surprised there are still no test result changes. I want to figure out what's missing before we merge this.

splav:
Ok. If there are candidates - I can investigate concrete tests. I'll try to look for appropriate tests manually though.

jdm:
I dug into it, and it turns out that there are no referrer-policy tests that start in an HTTPS context and make an HTTP request. This appears to be an oversight in the test suite generation, based on:
* https://github.com/web-platform-tests/wpt/blob/b62bd4a22154756bc021bdad08aafe9b57bb17f6/referrer-policy/spec.src.json#L130-L143 - expected results for omitting the referrer when starting in HTTPS and making an HTTP request with [no-referrer-when-downgrade](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-no-referrer-when-downgrade)
* https://github.com/web-platform-tests/wpt/blob/b62bd4a22154756bc021bdad08aafe9b57bb17f6/referrer-policy/spec.src.json#L551- all L570 - HTTPS->HTTP tests are excluded from the set of tests to generate

It looks like either the mixed-content exclusions are over-enthusiastic, or the author thought that the [mixed content](https://w3c.github.io/webappsec-mixed-content/) standard subsumes this for some reason.

I'm inclined to merge these changes and investigate extending the test suite separately.

splav:
Random thought: 
If the previous request was a modern https the it was handled correctly with checking url scheme.
Modern <=> "https" scheme; None <=> "http" scheme.
The only case not handled correctly previously I can imagine - Deprecated state from previous request.
I failed to find any uses of HttpsState::Deprecated in the code and is up to implementation what to consider deprecated. So, no fixed tests here does not seem strange.

The other thing I fixed - is_origin_trustworthy implementation. And here I do expect some test results changes.

jdm:
@bors-servo r-
Let's answer the question about is_origin_trustworthy before merging.

jdm:
Looks like the is_origin_trustworthy changes are difficult to test as part of the WPT harness, so let's merge these.
@bors-servo r+

jdm:
@bors-servo retry
* #26572

PR #26562, opened 2020-05-18
Improvements to animation keyframe computation

mrobinson:
This pull request contains two changes:

**Caching computed keyframes between animation changes**:
Instead of recomputing keyframe data for every tick of an animation,
cache the computed values when animations change. In addition to being
more efficient, this will allow us to return animation rules as property
declarations because we don't need to consult the final style to produce
them.

**Better computation of keyframe data**:
Instead of naively using `apply_declarations` to compute the style of
each keyframe, use an approach more like Gecko's where we carefully
walk through the animation keyframes and try to extract `AnimationValue`s
for each computed keyframe. In this approach we respect the order with
which the properties are set in the keyframe source and try to deal with
CSS custom properties.

---


mrobinson:
I've added a closely related change to this PR which improves the computation of keyframe data and ditches Servo's use of `apply_declarations` for this.

mrobinson:
This needs a little bit more work.

jdm:
I read through 917350369e6251b4296d658f2ce388193d000603 and it looks sensible.

mrobinson:
@jdm Thanks for looking at the first commit. That should be unchanged in the newest version of the branch. The second commit is unfortunately a bit large, which means it's probably harder to review. I wasn't sure how to split it into simpler chunks. 😕 I'm happy to walk through it with anyone or do whatever else is necessary to make it easier to review.

jdm (approved):
These changes look sensible to me.

components/style/animation.rs
```diff
@@ -172,100 +173,223 @@ pub enum KeyframesIterationState {
...
+        base_style: &Arc<ComputedValues>,
+    ) {
+        // Each keyframe declaration may optionally specify a timing function, falling
+        // back to the one defined global for the animation.
+        let guard = &context.guards.author;
+        if let Some(timing_function) = step.get_animation_timing_function(&guard) {
+            self.timing_function = Some(timing_function.to_computed_value_without_context());
+        }
+
+        match step.value {
```
emilio, line 236: Maybe `let block = match setep.value { ... };`, then de-indent the rest?

components/style/animation.rs
```diff
@@ -172,100 +173,223 @@ pub enum KeyframesIterationState {
...
+        debug_assert!(intermediate_steps.last().unwrap().start_percentage == 1.);
+
+        intermediate_steps
+    }
+
+    fn update_from_step(
+        &mut self,
+        step: &KeyframesStep,
+        context: &SharedStyleContext,
+        base_style: &Arc<ComputedValues>,
```
emilio, line 227: I think this can just be `&ComputedValues`, and same in the caller.

emilio (approved):
Yeah, nice progression :)

components/style/animation.rs
```diff
@@ -553,58 +677,46 @@ impl Animation {
...
         let mut new_style = (**style).clone();
-        let mut update_style_for_longhand = |longhand| {
-            let from = AnimationValue::from_computed_values(longhand, &prev_keyframe_style)?;
-            let to = AnimationValue::from_computed_values(longhand, &next_keyframe_style)?;
+        for (index, from) in prev_keyframe.values.iter().enumerate() {
             PropertyAnimation {
-                from,
-                to,
+                from: from.clone(),
+                to: next_keyframe.values[index].clone(),
```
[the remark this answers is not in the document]
mrobinson: Your suggestion is a lot clearer.

components/style/animation.rs
mrobinson: :+1:

components/style/animation.rs
mrobinson: Oh, nice.

mrobinson:
@bors-servo r=jdm,emilio

Thank you both for the reviews!

Issue #25508, opened 2020-01-13
Duplicate crate check no longer reports crate names


jdm:
```
Checking files for tidiness...
./Cargo.lock:1: duplicate versions for package `mach`
	The following packages depend on version 0.2.3 from 'crates.io':
	The following packages depend on version 0.3.2 from 'crates.io':
./Cargo.lock:1: duplicate versions for package `nix`
	The following packages depend on version 0.14.1 from 'crates.io':
	The following packages depend on version 0.15.0 from 'crates.io':
```
I suspect this is a regression from the python3 changes.

jdm:
https://github.com/servo/servo/blob/0131952d6d97bf904da353cec9a8fdb32307ef01/python/tidy/servo_tidy/tidy.py#L378-L399

jdm:
On the other hand, it's possible this has been broken since https://github.com/servo/servo/pull/24334, which would mean that https://github.com/servo/servo/blob/0131952d6d97bf904da353cec9a8fdb32307ef01/python/tidy/servo_tidy/tidy.py#L351-L355 is broken for some reason.

PR #26564, opened 2020-05-18
Upgrade wgpu version to "0.5.0" and add server-side code for id recycling for WebGPU

kunalmohan:
I have updated the cargo.lock to use a wgpu-core at a more recent commit where IdentityHandlerFactory was introduced.
r?@kvark

---


kunalmohan:
@kvark would you take a look at this before I proceed further?

components/webgpu/Cargo.toml
```diff
@@ -15,7 +15,8 @@ embedder_traits = {path = "../embedder_traits"}
 ipc-channel = "0.14"
 log = "0.4"
 malloc_size_of = { path = "../malloc_size_of" }
-serde = "1.0"
+serde = { version = "1.0", features = ["serde_derive"] }
 servo_config = {path = "../config"}
 smallvec = { version = "0.6", features = ["serde"] }
 wgpu-core = { version = "0.1.0", git = "https://github.com/gfx-rs/wgpu", features = ["serde"] }
+wgpu-types = { version = "0.1.0", git = "https://github.com/gfx-rs/wgpu" }
```
kvark, line 22: why are we using version 0.1.0 for wgpu? it should be 0.5

components/webgpu/identity.rs
```diff
@@ -0,0 +1,85 @@
...
+	hub::{IdentityHandler, IdentityHandlerFactory, GlobalIdentityHandlerFactory},
+	id::{
+        AdapterId, BindGroupId, BindGroupLayoutId, BufferId, CommandBufferId,
+        ComputePipelineId, DeviceId, PipelineLayoutId, ShaderModuleId, SwapChainId, RenderPipelineId, TextureId, TextureViewId, SamplerId, SurfaceId, TypedId
+    },
+};
+
+#[derive(Debug)]
+pub struct IdentityRecycler<I> {
+	fun: fn(I, IpcSender<WebGPUMsg>),
```
kvark, line 17: The current approach appears to mimic Gecko code in `gfx/wgpu_bindings` very closely. I think we can make it much more elegant in Servo, since unlike Gecko we don't have to cross the FFI boundary. More concretely, I think we can get away without function pointers here altogether. Here are more details:

Instead of having  `(self.fun)(id, self.sender.clone());` for all kind of indices, we could have this implemented specifically for each kind, i.e.
```rust
impl IdentityHandler<id::DeviceId> for IdentityRecycler<id::DeviceId> {
	type Input = id::DeviceId;
	fn process(&self, id: id::DeviceId, _backend: wgpu_types::Backend) -> id::DeviceId {
		log::debug!("process {} {:?}", self.kind, id);
		//debug_assert_eq!(id.unzip().2, backend);
		id
	}
	fn free(&self, id: id::DeviceId) {
		log::debug!("free {} {:?}", self.kind, id);
                let msg = WebGPUMsg::FreeDevice(id);
                if self.sender.send(msg).is_err() {
                    log::error!("Failed to send {:?}({:?})", msg, id);
                }
	}
}
```

That would achieve the same thing without messing with any functions. We'd obviously want to generate these `impl IdentityHandler` sections by a macro, just like you are doing in this PR for `impl_factory!`.

kvark (changes requested):
Thank you, this is one of the critical things we were missing!
Expressed some ideas on how we can rework this to be even better :)

components/webgpu/Cargo.toml
kunalmohan: You're right. Initially I was a bit worried to do that in a single step, since there are quite a lot of changes. But while working on this I realised doing it in steps would just confuse and complicate things. I'll pause on this PR for a while and work on upgrading the version first, if that's okay with you.

components/webgpu/identity.rs
kunalmohan: I'd still have to implement `IdentityHandlerFactory` for `IdentityRecyclerFactory` for each `Id` separately using `impl_factory!`, right?

kunalmohan:
It was difficult to upgrade to 0.5 without implementing the IdentityRecyclerFactory, so I have completed the task here itself. This PR now covers the following-
1. Upgrade to version wgpu-core 0.5
2. Since the spec doesn't include `MapReadAsync` anymore, I have removed it completely (instead of updating it with the version).
3. Implement the server-side part for id recycling.

kunalmohan:
The upgrade has introduced duplicate dependencies-
```
./Cargo.lock:1: duplicate versions for package `ron`
	The following packages depend on version 0.1.7 from 'crates.io':
	The following packages depend on version 0.5.1 from 'crates.io':
./Cargo.lock:1: duplicate versions for package `parking_lot_core`
	The following packages depend on version 0.6.2 from 'crates.io':
	The following packages depend on version 0.7.2 from 'crates.io':
./Cargo.lock:1: duplicate versions for package `metal`
	The following packages depend on version 0.17.0 from 'crates.io':
	The following packages depend on version 0.18.0 from 'crates.io':
./Cargo.lock:1: duplicate versions for package `parking_lot`
	The following packages depend on version 0.10.2 from 'crates.io':
	The following packages depend on version 0.9.0 from 'crates.io':
```
I tried resolving them with `./mach cargo-update ..` but wasn't successful.

kvark:
> The upgrade has introduced duplicate dependencies

Then we need a separate PR to Servo that bumps parking_lot, another one that bumps RON.

components/webgpu/identity.rs
kvark: I don't thin implementing `IdentityHandlerFactory` separately for each `I` is needed. All you need is to remove the `kind` field, and it can be implemented for all `<I>` at once. The kind is only used in the logging of `IdentityHandler` implementation, where we are going to use a macro, so we can just paste the kind string directly (from a macro argument) instead of storing it as a field.

jdm:
You can add duplicates to servo-tidy.toml if it's not straightforward to update the packages that depend on the older versions.

jdm:
And yes, the output from tidy in that case is less helpful than it could be: https://github.com/servo/servo/issues/25508

components/webgpu/identity.rs
kunalmohan: I have been trying that, but getting stuck with specifying bounds of `I` here. Is it possible to specify `I` to be from a list of Types or are Traits the only possible way of specifying bounds?

components/webgpu/identity.rs
```diff
@@ -0,0 +1,115 @@
...
+    FreeCommandBuffer(CommandBufferId),
+    FreeTexture(TextureId),
+    FreeTextureView(TextureViewId),
+    FreeSampler(SamplerId),
+    FreeSurface(SurfaceId),
+    FreeShaderModule(ShaderModuleId),
+}
+
+#[derive(Debug)]
+pub struct IdentityRecycler<I> {
```
kvark, line 37: we technically don't have to make this generic over `I` even. It's not exposed to the outside, and it's used behind the strongly typed associated types.

components/webgpu/identity.rs
```diff
@@ -0,0 +1,115 @@
...
+    "bind_group_layout",
+    WebGPUMsg::FreeBindGroupLayout
+);
+impl_identity_handler!(
+    PipelineLayoutId,
+    "pipeline_layout",
+    WebGPUMsg::FreePipelineLayout
+);
+
+impl<I: TypedId + Clone + std::fmt::Debug> IdentityHandlerFactory<I> for IdentityRecyclerFactory
```
kvark, line 102: let's not mix the generic bounds here and in the `where` section. Since we have to use the latter, let's also have a line there:
```
where
    I: TypedId + Clone + std::fmt::Debug,
    IdentityRecycler<I>: IdentityHandler<I>,
```

kvark (approved):
I think it looks good now. Have a few more minor suggestions.
The main unresolved thing is duplicated dependencies. If @jdm is fine to follow up with them, so be it:)

kunalmohan:
@jdm parking_lot is a dependency of many packages, one of which is `Surfman`. metal also is a dependency for `Surfman`. ron is required by `Webrender`. I guess upgrading them wouldn't be possible from here. So I'll proceed with the servo-tidy.toml option.

jdm:
opening pull requests can't hurt!

jdm:
but yes, no need to gate this PR on those other repositories.

kunalmohan:
I have made the required changes and squashed the commits.

components/webgpu/identity.rs
```diff
@@ -0,0 +1,116 @@
...
+    FreeCommandBuffer(CommandBufferId),
+    FreeTexture(TextureId),
+    FreeTextureView(TextureViewId),
+    FreeSampler(SamplerId),
+    FreeSurface(SurfaceId),
+    FreeShaderModule(ShaderModuleId),
+}
+
+#[derive(Debug)]
+pub struct IdentityRecycler<I: TypedId + Clone + std::fmt::Debug> {
```
kvark, line 37: we shouldn't put bounds on the structure generic arguments, unless required. IIRC, that's what Rust best practices are. Instead, the bounds are put on implementations of things.

kunalmohan: Sorry, I think I misunderstood your previous comment. I have corrected it now and removed the generic `I` from struct definition completely.

kvark (approved):

jdm:
```
thread 'rustc' panicked at 'could not find markdown in source', src/librustdoc/passes/mod.rs:411:31
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1504
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:511
  11: rust_begin_unwind
             at src/libstd/panicking.rs:419
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:111
  13: core::option::expect_failed
             at src/libcore/option.rs:1260
  14: rustdoc::passes::source_span_for_markdown_range
  15: <rustdoc::passes::check_code_block_syntax::SyntaxChecker as rustdoc::fold::DocFolder>::fold_item
  16: <alloc::vec::Vec<T> as alloc::vec::SpecExtend<T,I>>::from_iter
  17: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
  18: <std::collections::hash::map::HashMap<K,V,S> as core::iter::traits::collect::FromIterator<(K, V)>>::from_iter
  19: rustdoc::passes::check_code_block_syntax::check_code_block_syntax
  20: rustc_middle::ty::context::tls::enter_global
  21: rustc_interface::interface::run_compiler_in_existing_thread_pool
  22: rustdoc::core::run_core
  23: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  24: rustc_driver::catch_fatal_errors
error: Could not document `gfx-backend-empty`.
```

kvark:
This is https://github.com/rust-lang/rust/issues/70874, which was fixed. Sounds like we need a rustc version update in Servo before moving on?

jdm:
@kunalmohan You should be able to change the nightly version in the rust-toolchain file to any nightly that is older than 5/21/2020.

kunalmohan:
I was getting this warning at a few places after toolchain update- 
```unused return value of `std::mem::replace` that must be used```
So I have prefixed the statements with `let _ =`

PR #26594, opened 2020-05-21
Implement animationiteration event

mrobinson:
This event is triggered when an animation iterates. This change also
moves iteration out of style calculation to an "update animations" which
is the next part of having animation event handling match the HTML spec.

This change causes a few more tests to pass. Some of the other tests which
exercise this functionality require `animationstart` events.

---

### DIFF
--- a/components/script/animations.rs
+++ b/components/script/animations.rs
@@ -35,6 +35,30 @@ impl Animations {
         }
     }
 
+    pub(crate) fn update_for_new_timeline_value(
+        &mut self,
+        window: &Window,
+        now: f64,
+    ) -> AnimationsUpdate {
+        let mut update = AnimationsUpdate::new(window.pipeline_id());
+        let mut sets = self.sets.write();
+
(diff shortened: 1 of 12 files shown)


components/script/animations.rs
```diff
@@ -35,6 +35,31 @@ impl Animations {
...
+        &mut self,
+        window: &Window,
+        now: f64,
+    ) -> AnimationsUpdate {
+        let mut update = AnimationsUpdate::new(window.pipeline_id());
+        let mut sets = self.sets.write();
+
+        for set in sets.values_mut() {
+            // When necessary, iterate our running animations to the next iteration.
+            // TODO(mrobinson): We need to send animationiteration events here.
```
jdm, line 48: This Todo can be removed, right?

jdm (approved):

components/script/animations.rs
mrobinson: Yep. I've updated the branch, removing this commend.

Issue #26571, opened 2020-05-19
Compile non-inline non-module scripts off thread


jdm:
https://searchfox.org/mozilla-central/rev/61fceb7c0729773f544a9656f474e36cd636e5ea/dom/script/ScriptLoader.cpp#2191-2205
https://searchfox.org/mozilla-central/rev/61fceb7c0729773f544a9656f474e36cd636e5ea/dom/script/ScriptLoader.cpp#2035-2069

Gecko has a model where inline scripts get evaluated directly, but external scripts get compiled off the main thread, then the parser resumes execution of the compiled bytecode once compilation is finished. This allows the page to continue responding to user input when a large script has just been loaded.

gterzian:
assigned to @AbhishekSharma102

I will try to write up more details...

gterzian:
So this would have to use [`js::jsapi::CompileOffThread`](https://github.com/servo/mozjs/blob/aabcc9ba889b2755f1e4e83f28323a60415a790f/mozjs/js/public/OffThreadScriptCompilation.h#L64), and pass a [`OffThreadCompileCallback`](https://github.com/servo/mozjs/blob/aabcc9ba889b2755f1e4e83f28323a60415a790f/mozjs/js/public/OffThreadScriptCompilation.h#L41) that would be called by the JS engine with a token, and whatever we pass as `void* callbackData`. 

Since the callback is called off the main-thread, we would need to then queue a task back on the event-loop/main-thread where the original call to `CompileOffThread` happened. I would say we should use the `NetworkingTaskSource`, available at https://github.com/servo/servo/blob/49d005595964ef03019f903fedf58834408f8ae3/components/script/dom/globalscope.rs#L2158

So I'd say our `void* callbackData` should be a struct similar to https://github.com/servo/servo/blob/49d005595964ef03019f903fedf58834408f8ae3/components/script/dom/globalscope.rs#L272, that we would cast to `void* callbackData`. 

So essentially this struct should own a `Trusted<GlobalScope`, a task-source and canceller, and other data that needs to be saved somewhere in between the initial call to `CompileOffThread`, and the actual handling of the compiled script back on the main thread. 

With regards to the `OffThreadCompileCallback`, I'm not completely sure but I'm wondering if we could simply pass a `unsafe extern "C" fn`? 

[...]

Issue #24684, opened 2019-11-07
Fix improper_ctypes warnings in Rust Nightly


SimonSapin:
Steps:

* Make a branch up to date with `origin/master`
* Run `./mach rustup`, which updates the `rust-toolchain` file
* Run `./mach build`
* See many warnings like
  ```rust
  warning: `extern` fn uses type `script_runtime::JSContext`, which is not FFI-safe
     --> components/script/script_runtime.rs:829:56
      |
  829 | unsafe extern "C" fn report_stream_error_callback(_cx: *mut JSContext, error_code: usize) {
      |                                                        ^^^^^^^^^^^^^^ not FFI-safe
      |
      = note: `#[warn(improper_ctypes)]` on by default
      = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
      = note: this struct has unspecified layout
  ```
* Find a way to fix them. This may involve upgrading rust-bindgen, or making changes rust-bindgen and then upgrading to a version with those changes
* Revert the `rust-toolchain` change
* Submit a PR

We don’t need to upgrade the Rust version yet, but the next time we do those warnings would block landing the upgrade unless we change CI to allow build warnings.

Is suspect that https://github.com/rust-lang/rust/pull/65134 is what introduced these warnings.

jdm:
JSContext is indeed not FFI-safe. It should be a jsapi::JSContext instead, and https://github.com/servo/servo/pull/24653 fixes that.

SimonSapin:
The Rust team is (kindly) waiting on us to be able to upgrade before landing https://github.com/rust-lang/rust/pull/66344. I have a branch ready for that, but it requires a Nightly more recent than these new warnings.

SimonSapin:
I’ve looked into this. Unfortunately there is no accidental bug here: the lint was intentionally changed to warn against a pattern that is in pervasive use in our DOM bindings.

Specifically: when an `extern fn` function (meant to be called by C++ or JIT’ed code in SpiderMonkey) takes as parameter a raw pointer to a Rust struct that is not FFI-safe.

Our DOM objects are based on many Rust structs that are not “FFI-safe”: non-Rust code cannot relay on their exact memory layout. Making a struct FFI-safe involves using `#[repr(C)]`, but it shouldn’t be necessary for DOM objects that are only manipulated directly in Rust code, or through pointers in non-Rust code.

[...]

PR #26606, opened 2020-05-21
Added macro to create functions in PipelineNamespace

### LINKED ISSUE #26380: Write a macro to generate namespaced Ids
We have a bunch of Indexes and Ids in `components/msg/constellation_msg.rs`, see for example the `BrowsingContextIndex` and `BrowsingContextId` https://github.com/servo/servo/blob/1b4162aecbb935c8183438c3d6bc285db3932476/components/msg/constellation_msg.rs#L263

I find myself manually adding new ones regurlarly by hand, perhaps someone wants to write a macro for this? See for example the changes to this file: https://github.com/servo/servo/pull/26317/files#diff-8dc9268416cd4f037d50206e76c73cf9

I'd be great if we could just do something along the lines of `namespaced_id!(ServiceWorkerId)`.

Note that each also require implementing a method on `PipelineNamespace`, which I guess will have to be done by hand(unless someone has a better idea). 

I don't have much, if any, experience writing macros, but I'll venture ahead with a "less easy" label...

AbhishekSharma102:
Changes to constellation_msg.rs to generate namespace id using macros

- [X] These changes fix #26380 
- [X] These changes do not require tests because no functionality is changed.

### DIFF
--- a/components/msg/constellation_msg.rs
+++ b/components/msg/constellation_msg.rs
@@ -14,6 +14,45 @@ use std::num::NonZeroU32;
 use std::sync::Arc;
 use std::time::Duration;
 
+macro_rules! namespace_id_method {
+    ($func_name:ident, $func_return_data_type:ident, $self:ident, $index_name:ident) => {
+        fn $func_name(&mut $self) -> $func_return_data_type {
+            $func_return_data_type {
+                namespace_id: $self.id,
+                index: $index_name($self.next_index()),
+            }
+        }
+    };
+}
+
+macro_rules! namespace_id {
+    ($id_name:ident, $index_name:ident) => {
+        #[derive(
+            Clone, Copy, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize,
+        )]
+        pub struct $index_name(pub NonZeroU32);
+        malloc_size_of_is_0!($index_name);
+
+        #[derive(
+            Clone,
+            Copy,
+            Debug,
+            Deserialize,
+            Eq,
+            Hash,
+            MallocSizeOf,
+            Ord,
+            PartialEq,
+            PartialOrd,
+            Serialize,
+        )]
+        pub struct $id_name {
+            pub namespace_id: PipelineNamespaceId,
+            pub index: $index_name,
+        }
+    };
+}
+
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub enum TraversalDirection {
     Forward(usize),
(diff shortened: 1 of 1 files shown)


gterzian (changes requested):
Thank you, looks likes a great start!

Would it be possible to additional functionality to the macro, maybe via a separate one, that would generate the various `Id` types, something like `namespaced_id!(PipelineId)`(which would then also generate the matching `PipelineIndex`)?

Perhaps this could be done with the same macro, but I am not sure, in that case it could be for example `create_pipeline_namespace! {PipelineId, self}` that would do what the current macro does, plus would also generate the `PipelineId` and `PipelineIndex` types.

AbhishekSharma102:
Hey!

 I believe rust does not allow modification to variable names (like suffix and prefix). I was wondering if we could pass `Pipeline` to the macro, and it would append `Id` to get `PipelineId` and `Index`  to get `PipelineIndex` . But rust does not allow it. So we would need to pass all the variable names. 

We can definitely generate the Id and Index inside the macros. I am thinking of having two macros, one defined in the PR, and another one `namespace_id` which takes two args, first is `id_name` like `PipelineId` and second `index_name` like `PipelineIndex` and returns them.

Let me know if this sounds good.

gterzian:
> So we would need to pass all the variable names.

Ok, so you mean we'd have to pass `PipelineId` and `PipelineIndex`, sounds good to me, definitely better than the current situation.

> I am thinking of having two macros, one defined in the PR, and another one namespace_id which takes two args, first is id_name like PipelineId and second index_name like PipelineIndex and returns them.

Sounds good. 

> Also I believe two macros would need to be called separately. Please correct if I'm wrong.

I'm not sure, let me know what you find out!

AbhishekSharma102:
Hey, I have added the macro. Let me know if this looks good.

components/msg/constellation_msg.rs
```diff
@@ -136,68 +163,17 @@ impl PipelineNamespace {
...
-        }
-    }
-
-    fn next_blob_id(&mut self) -> BlobId {
-        BlobId {
-            namespace_id: self.id,
-            index: BlobIndex(self.next_index()),
-        }
-    }
+    create_pipeline_namespace_function! {next_pipeline_id, PipelineId, self, PipelineIndex}
```
gterzian, line 166: nit: how about we name it a bit shorter, like `namespace_id_method`?

gterzian (changes requested):
Looks great! Just one question.

components/msg/constellation_msg.rs
AbhishekSharma102: renamed it

gterzian:
Great! If you can please squash the third commit into the first, then it should be good to go!

AbhishekSharma102:
Squashed them into one.

gterzian:
Awesome, thanks!

@bors-servo r+

AbhishekSharma102:
@gterzian Can you please suggest another issue that I can take up?

gterzian:
I'm not sure, there are not many issues marked "less easy" any more, perhaps this one: https://github.com/servo/servo/issues/24684?

If you want something harder I think this one is interesting: https://github.com/servo/servo/issues/26571 (it might require some more guidance)

AbhishekSharma102:
I'd like to work on #26571 as I'm hoping there will be more learning. Could you please assign that to me? And please tell on what to read up for the same?

PR #26609, opened 2020-05-21
Improve ending of transitions

mrobinson:
For legacy reasons, transitions were marked as finished after updating the
style. According to the spec, they should be marked as finished when
animations are updated and before restyle. This change does that as well
as preventing replacement of finished transitions.

Having finished transitions survive a full restyle cycle and allowing
the replacement of finished transition could lead to issues where
animations are removed from the global list of animating transitions
too soon:

 1. A transitions finishes
 2. Restyle
 3. The transition is marked as finished and events are queued
 4. Restyle cancels finished transition and replaces it instead of
    clearing finished transition
 5. Events are sent for the incorrectly canceled transition, removing it
    completely from the list of running transitions due to the extra
    event.

---

### DIFF
--- a/components/script/animations.rs
+++ b/components/script/animations.rs
@@ -55,6 +55,8 @@ impl Animations {
                     );
                 }
             }
+
+            Self::finish_running_animations(set, now, &mut update);
         }
         update
     }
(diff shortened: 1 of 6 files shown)


mrobinson:
This causes some new failures in property inheritance tests for Layout 2020, probably due to timing changes. These tests are not expected to work correctly, since we don't do proper cascading of animating values yet. In addition, these tests are known to be flaky in other browser engines as well. I've added a new tests that specifically tests this behavior though.

PR #26611, opened 2020-05-22
Don't try to take log2 of 0 when using mono audio channels.

jdm:
This avoids a panic when loading rooms in Hubs.

### DIFF
--- a/components/script/dom/baseaudiocontext.rs
+++ b/components/script/dom/baseaudiocontext.rs
@@ -487,7 +487,11 @@ impl BaseAudioContextMethods for BaseAudioContext {
                     let channel = match channels.entry(channel_pos_mask) {
                         Entry::Occupied(entry) => *entry.get(),
                         Entry::Vacant(entry) => {
-                            let x = (channel_pos_mask as f32).log2() as usize;
+                            let x = if channel_pos_mask != 0 {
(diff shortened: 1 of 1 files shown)


Manishearth:
We also shouldn't be using log2 at all, ideally we can do `(x - 1).count_ones()`, conditional on `x.count_ones()` being one

jdm:
I now can't reproduce this panic on Hubs without this PR applied, and I haven't been able to reproduce the panic in a page using a mono audio buffer:
```html
<script>
var context = new AudioContext();
var buffer = context.createBuffer(1, 22050, 22050);
for (var channel = 0; channel < buffer.numberOfChannels; channel++) {
  // This gives us the actual ArrayBuffer that contains the data
  var nowBuffering = buffer.getChannelData(channel);
  for (var i = 0; i < buffer.length; i++) {
    // Math.random() is in [0; 1.0]
    // audio needs to be in [-1.0; 1.0]
    nowBuffering[i] = Math.random() * 2 - 1;
  }
}
var source = context.createBufferSource();
source.buffer = buffer;
source.connect(context.destination);
source.start();
</script>
```
Since I don't really want to merge changes that can't clearly show an improvement, I'm going to close this PR. We can reopen it if we find a way to trigger the problem again.

PR #26619, opened 2020-05-22
Implement client-side logic for WebGPU id recycling

kunalmohan:
1. Set up `WebGPUMsg` receiver in script thread and add respective message handler function.
2. Add `kill_xxx_id()` methods to `Identities`.

r?@kvark

---

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2209,16 +2209,13 @@ where
             Some(bc) => &bc.bc_group_id,
             None => return warn!("Browsing context not found"),
         };
-        let host = match self
-            .pipelines
-            .get(&source_pipeline_id)
-            .map(|pipeline| &pipeline.url)
-        {
-            Some(ref url) => match reg_host(&url) {
-                Some(host) => host,
-                None => return warn!("Invalid host url"),
-            },
+        let source_pipeline = match self.pipelines.get(&source_pipeline_id) {
+            Some(pipeline) => pipeline,
             None => return warn!("ScriptMsg from closed pipeline {:?}.", source_pipeline_id),
+        };
+        let host = match reg_host(&source_pipeline.url) {
+            Some(host) => host,
+            None => return warn!("Invalid host url"),
         };
         match self
             .browsing_context_group_set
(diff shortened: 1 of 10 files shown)


components/constellation/constellation.rs
```diff
@@ -2238,7 +2235,22 @@ where
                 let send = match browsing_context_group.webgpus.entry(host) {
                     Entry::Vacant(v) => v
                         .insert(match WebGPU::new() {
-                            Some(webgpu) => webgpu,
+                            Some(webgpu) => {
+                                let msg = ConstellationControlMsg::SetWebGPUPort(webgpu.1);
+                                match source_pipeline {
```
kvark, line 2240: given the previous match on the `source_pipeline`, it looks like it has to be Some() here.
I think we should refactor it to extract it right away in the beginning:
```rust
let source_pipeline = match self.pipelines.get(&source_pipeline_id) {
  Some(pipeline) => pipeline,
  None => return warn!(...),
};
```

components/script/script_thread.rs
```diff
@@ -692,8 +694,11 @@ pub struct ScriptThread {
     /// Code is running as a consequence of a user interaction
     is_user_interacting: Cell<bool>,
 
-    /// Channel to WebGPU
+    /// Identity manager for WebGPU resources
     gpu_id_hub: Arc<Mutex<Identities>>,
+
+    /// Receiver to receive commands from optional WebGPU server.
+    webgpu_port: RefCell<Option<Receiver<WebGPUMsg>>>,
```
kvark, line 701: Shouldn't this be `DomRefCell`?

kvark (changes requested):
Looks great overall, just a few notes

components/script/script_thread.rs
kunalmohan: I am not sure about this. `DomRefCell` is used for DOM objects, I don't think this is one. But again I am not sure. I'll change it if you say so and it would be great if you could help me understand the reason :)

kvark: I see. I hope that the compiler will tell us if we mess this up, so we can leave it like this (unless @jdm tells otherwise)

kvark (approved):

Issue #26502, opened 2020-05-13
Improve reliability of threading and shutdown


gterzian:
So I've had some discussions elsewhere with @matklad who pointed out a few things that can make threading, especially the shutting down of threads, more reliable, and I think we should consider adopting these practices systematically across the Servo codebase, although this can only be done in a piecemeal fashion so I'm marking this as a "meta" issue. 

For the record, full discussion was at: https://github.com/gterzian/rust_five_easy_pieces/pull/1

As a recap, we could consider:

1. Replacing exit workflows with explicit exit messages, with workflows based on handling channels closing when (all) senders have been dropped. We do this already in places, like https://github.com/servo/servo/blob/c5617efff0f77ce79a7827668cf929def2d9337a/components/background_hang_monitor/background_hang_monitor.rs#L246

The benefit of those drop-based workflows is really that it forces you to think in terms of dropping senders, which forces you to think in terms of properly shutting down everything and not leaking a channel somewhere. 

It's a challenge because I think there are places were we keep senders alive to clone then as we spawn new pipelines and so on, however I also think those workflows might be restructured, or we could separate channels better so that there is always one channel that can be reliably dropped to signal a shutdown. 

For an example of a channel kept alive, see the `control_chan` at https://github.com/servo/servo/blob/44ac62ddda3652e0f6c79bb4dc77ef389645cf37/components/script/script_thread.rs#L579

The reason it's kept alive is because it is cloned and passed to layout threads, as well as the CSS thingy I can't find right now, to send certain `ConstellationControlMsg`. 

So we could restructure this by having layout use a dedicated channel, instead of re-using the control chan from the constellation. 

Then when the the constellation exits an event-loop, the chan would drop, and the script thread could use that as a shutdown signal. 

2. Joining on all child threads. I think this could help reduce the "threads running post shutdown count".

This is somewhat complicated in the case of IPC, however if you think about it each process usually as some sort of "master thread", for example I would argue it is the `ScriptThread` for a window agent-cluster, and it would be the `ServiceWorkerManager` for a service-worker agent-cluster. 

So for example, when the constellation sends a exit IPC message, the "main thread" of the other process could ensure it joins on all child threads(like layout and web workers, or actual service workers globals), before sending the ack back to the constellation. 

3. Unwrap/expect on all channel sends. I think if we have 1 and 2, then this is feasible too.  See for a start https://github.com/servo/servo/pull/25685

Please let me know what you think.

matklad:
[...]

PR #26628, opened 2020-05-24
Improve worker shutdown

### LINKED ISSUE #25212: This runtime still has live children error in worker
To reproduce: `./mach test-wpt /FileAPI/url/url-with-fetch.any.worker.html --release`

When the suite is done, close the window manually by clicking(not control-c).

```
pid:7516 assertion failed: `(left == right)`
 0:05.35 pid:7516   left: `2`,
 0:05.35 pid:7516  right: `1`: This runtime still has live children. (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }, at /Users/gregory/.cargo/git/checkouts/rust-mozjs-8611526964119dd6/9b0d063/src/rust.rs:410)
 0:05.35 pid:7516 stack backtrace:
 0:05.41 pid:7516    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
 0:05.41 pid:7516    1: core::fmt::write
 0:05.41 pid:7516    2: <std::io::stdio::Stdout as std::io::Write>::write_fmt
 0:05.41 pid:7516    3: std::io::stdio::_print
 0:05.41 pid:7516    4: servo::backtrace::print
 0:05.41 pid:7516    5: servo::main::{{closure}}
 0:05.41 pid:7516    6: std::panicking::rust_panic_with_hook
 0:05.41 pid:7516    7: std::panicking::continue_panic_fmt
 0:05.41 pid:7516    8: std::panicking::begin_panic_fmt
 0:05.41 pid:7516    9: <mozjs::rust::Runtime as core::ops::drop::Drop>::drop
 0:05.42 pid:7516   10: _ZN4core3ptr18real_drop_in_place17h509cc1c478040de3E.llvm.14111175920642720576
 0:05.42 pid:7516   11: <alloc::rc::Rc<T> as core::ops::drop::Drop>::drop
 0:05.42 pid:7516   12: core::ptr::real_drop_in_place
 0:05.42 pid:7516   13: std::sys_common::backtrace::__rust_begin_short_backtrace
 0:05.42 pid:7516   14: _ZN3std9panicking3try7do_call17ha5548ac9f47f8e64E.llvm.9403888586713981603
 0:05.42 pid:7516   15: __rust_maybe_catch_panic
 0:05.42 pid:7516   16: core::ops::function::FnOnce::call_once{{vtable.shim}}
 0:05.42 pid:7516   17: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
 0:05.42 pid:7516   18: std::sys::unix::thread::Thread::new::thread_start
 0:05.42 pid:7516   19: _pthread_body
 0:05.42 pid:7516   20: _pthread_start
 0:05.42 pid:7516 
 0:05.42 pid:7516 [2019-12-09T06:37:27Z ERROR servo] assertion failed: `(left == right)`
 0:05.42 pid:7516   left: `2`,
 0:05.42 pid:7516  right: `1`: This runtime still has live children.
```

### LINKED ISSUE #26548: ServiceWorker: add shutdown mechanism for worker scopes
This PR https://github.com/servo/servo/pull/26317 removes the current timeout mechanism, because it didn't fit into the overall restructuring. So we need to add a new shutdown mechanism, that could be used by the `ServiceWorkerManager` itself to shut down all worker scopes when it is exiting, and it could also be used to timeout workers(although this is a separate issue, see https://github.com/servo/servo/issues/26547. 

The PR also removes the event-loop from the worker scope(by setting the `has_timed_out` to return a default `true` so that the event-loop is not run at all), so once we have a shutdown mechanism, we should turn it back on, initially simply by having `has_timed_out` return `false`.

gterzian:
FIX #26548
FIX #25212

and also  a step towards https://github.com/servo/servo/issues/26502

---

### DIFF
--- a/components/script/dom/abstractworkerglobalscope.rs
+++ b/components/script/dom/abstractworkerglobalscope.rs
@@ -82,12 +82,15 @@ impl ScriptPort for Receiver<DedicatedWorkerScriptMsg> {
 
 pub trait WorkerEventLoopMethods {
     type WorkerMsg: QueuedTaskConversion + Send;
+    type ControlMsg;
     type Event;
     fn task_queue(&self) -> &TaskQueue<Self::WorkerMsg>;
-    fn handle_event(&self, event: Self::Event);
+    fn handle_event(&self, event: Self::Event) -> bool;
     fn handle_worker_post_event(&self, worker: &TrustedWorkerAddress) -> Option<AutoWorkerReset>;
+    fn from_control_msg(&self, msg: Self::ControlMsg) -> Self::Event;
     fn from_worker_msg(&self, msg: Self::WorkerMsg) -> Self::Event;
     fn from_devtools_msg(&self, msg: DevtoolScriptControlMsg) -> Self::Event;
+    fn control_receiver(&self) -> &Receiver<Self::ControlMsg>;
 }
 
 // https://html.spec.whatwg.org/multipage/#worker-event-loop
@@ -108,6 +111,7 @@ pub fn run_worker_event_loop<T, WorkerMsg, Event>(
     };
     let task_queue = worker_scope.task_queue();
     let event = select! {
+        recv(worker_scope.control_receiver()) -> msg => worker_scope.from_control_msg(msg.unwrap()),
         recv(task_queue.select()) -> msg => {
             task_queue.take_tasks(msg.unwrap());
             worker_scope.from_worker_msg(task_queue.recv().unwrap())
@@ -136,7 +140,10 @@ pub fn run_worker_event_loop<T, WorkerMsg, Event>(
     }
     // Step 3
     for event in sequential {
-        worker_scope.handle_event(event);
+        if !worker_scope.handle_event(event) {
+            // Shutdown
+            return;
+        }
         // Step 6
         let _ar = match worker {
             Some(worker) => worker_scope.handle_worker_post_event(worker),
--- a/components/script/dom/dedicatedworkerglobalscope.rs
+++ b/components/script/dom/dedicatedworkerglobalscope.rs
@@ -86,6 +86,12 @@ impl<'a> Drop for AutoWorkerReset<'a> {
     }
 }
 
+/// Messages sent from the owning global.
+pub enum DedicatedWorkerControlMsg {
+    /// Shutdown the worker.
+    Exit,
+}
+
 pub enum DedicatedWorkerScriptMsg {
     /// Standard message from a worker.
     CommonWorker(TrustedWorkerAddress, WorkerScriptMsg),
@@ -96,6 +102,7 @@ pub enum DedicatedWorkerScriptMsg {
 pub enum MixedMessage {
     FromWorker(DedicatedWorkerScriptMsg),
     FromDevtools(DevtoolScriptControlMsg),
+    FromControl(DedicatedWorkerControlMsg),
 }
 
 impl QueuedTaskConversion for DedicatedWorkerScriptMsg {
@@ -183,23 +190,32 @@ pub struct DedicatedWorkerGlobalScope {
     #[ignore_malloc_size_of = "Arc"]
     image_cache: Arc<dyn ImageCache>,
     browsing_context: Option<BrowsingContextId>,
+    /// A receiver of control messages,
+    /// currently only used to signal shutdown.
+    #[ignore_malloc_size_of = "Channels are hard"]
+    control_receiver: Receiver<DedicatedWorkerControlMsg>,
 }
 
 impl WorkerEventLoopMethods for DedicatedWorkerGlobalScope {
     type WorkerMsg = DedicatedWorkerScriptMsg;
+    type ControlMsg = DedicatedWorkerControlMsg;
     type Event = MixedMessage;
 
     fn task_queue(&self) -> &TaskQueue<DedicatedWorkerScriptMsg> {
         &self.task_queue
     }
 
-    fn handle_event(&self, event: MixedMessage) {
-        self.handle_mixed_message(event);
+    fn handle_event(&self, event: MixedMessage) -> bool {
+        self.handle_mixed_message(event)
     }
 
     fn handle_worker_post_event(&self, worker: &TrustedWorkerAddress) -> Option<AutoWorkerReset> {
         let ar = AutoWorkerReset::new(&self, worker.clone());
         Some(ar)
+    }
+
+    fn from_control_msg(&self, msg: DedicatedWorkerControlMsg) -> MixedMessage {
+        MixedMessage::FromControl(msg)
     }
 
     fn from_worker_msg(&self, msg: DedicatedWorkerScriptMsg) -> MixedMessage {
@@ -208,6 +224,10 @@ impl WorkerEventLoopMethods for DedicatedWorkerGlobalScope {
 
     fn from_devtools_msg(&self, msg: DevtoolScriptControlMsg) -> MixedMessage {
         MixedMessage::FromDevtools(msg)
+    }
+
+    fn control_receiver(&self) -> &Receiver<DedicatedWorkerControlMsg> {
+        &self.control_receiver
     }
 }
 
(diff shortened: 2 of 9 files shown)


gterzian:
@asajeffrey r?

jdm:
@asajeffrey Review ping.

components/script/dom/abstractworkerglobalscope.rs
```diff
@@ -108,6 +110,16 @@ pub fn run_worker_event_loop<T, WorkerMsg, Event>(
...
     let task_queue = worker_scope.task_queue();
     let event = select! {
+        recv(worker_scope.control_receiver()) -> msg => {
+            // Note: this is currently only used to signal shutdown,
+            // to enable more functionality, add other methods
+            // to translate the msg into an event,
+            // and handle the event.
+
+            // Shutdown.
+            assert!(msg.is_err());
```
asajeffrey, line 120: Using an error rather than a message for signalling shutdown seems odd?

components/script/dom/dedicatedworkerglobalscope.rs
```diff
@@ -183,10 +183,15 @@ pub struct DedicatedWorkerGlobalScope {
     #[ignore_malloc_size_of = "Arc"]
     image_cache: Arc<dyn ImageCache>,
     browsing_context: Option<BrowsingContextId>,
+    /// A receiver of control messages,
+    /// currently only used to signal shutdown.
+    #[ignore_malloc_size_of = "Channels are hard"]
+    control_receiver: Receiver<()>,
```
asajeffrey, line 189: Use a `ShutdownMsg` type rather than `()`?

components/script/dom/globalscope.rs
```diff
@@ -112,15 +113,41 @@ use std::ops::Index;
...
+        // Step 1.
+        self.closing.store(true, Ordering::SeqCst);
+
+        // Drop the channel to signal shutdown.
+        drop(
+            self.control_sender
+                .take()
+                .expect("No control sender to worker thread."),
+        );
+
```
asajeffrey, line 143: Why not send a message rather than dropping the channel?

components/script/dom/globalscope.rs
```diff
@@ -112,15 +113,41 @@ use std::ops::Index;
...
+        // TODO: step 2 and 3.
+        // Step 4 is unnecessary since we don't use actual ports for dedicated workers.
+        self.join_handle
+            .take()
+            .expect("No handle to join on worker.")
+            .join()
+            .expect("Couldn't join on worker thread.");
```
asajeffrey, line 150: Should there be a panic if joining fails?

components/script/serviceworker_manager.rs
```diff
@@ -77,6 +77,25 @@ enum RegistrationUpdateTarget {
...
 
+impl Drop for ServiceWorkerRegistration {
+    /// <https://html.spec.whatwg.org/multipage/#terminate-a-worker>
+    fn drop(&mut self) {
+        // Drop the channel to signal shutdown.
+        drop(
+            self.control_sender
+                .take()
+                .expect("No control sender to worker thread."),
+        );
```
asajeffrey, line 88: Ditto.

asajeffrey (changes requested):
LGTM apart from some nits.

components/script/dom/dedicatedworkerglobalscope.rs
gterzian: The idea is that the shutdown is signalled by dropping the sender(and to avoid explicit shutdown messages), and the `()` is used as a placeholder since not other messages are used for now. Re the sthutdown, I'll answer below.

components/script/dom/globalscope.rs
gterzian: Since we're in script, I would say yes? What do you think?

components/script/dom/globalscope.rs
gterzian: I tried to write this down over at https://github.com/servo/servo/issues/26502

While we use a lot of exit messages throughout the codebase, I think @matklad had a good point that relying on dropping channels is better, because it removes all sort of opportunities for leaking things accidentally, and it also enforces some sort of recursive dropping/shutting-down of resources on designs(since otherwise if it doesn't drop it won't shut-down properly). 

So for example, if the control channel is cloned somewhere, then the shut-down will fail, and that would then force us to not clone the control channel, or otherwise ensure all senders drop. 

Then this also enforces proper separation of concerns in terms of channels, since we can't take the easy route and clone let's say a control channel and pass it to some child worker, since that would break the shutdown mechanism. 

I think a good example is how currently the script-thread clones it's `control_chan: IpcSender<ConstellationControlMsg>,` and passes it to layout threads, which means we can use it as a "drop to shutdown" channel, unless we actually introduce a new channel for layout -> script communication, which actually makes sense(it's inconsistent that layout sends `ConstellationControlMsg` in some cases). 

Please let me know what you think of this.

components/script/dom/dedicatedworkerglobalscope.rs
matklad: I also like using Never as a message type in these cases, which make it clear(I.e statically impossible) to send a message: https://github.com/rust-lang/rls/blob/17a439440e6b00b1f014a49c6cf47752ecae5bb7/rls/src/concurrency.rs#L88

It also removes the question of handling OK case on the receive side (you write an empty match: `match never {}`)

Admittedly, “wtf is Sender<Never>“ is a common reaction when people see such code for the first time. On the flip side, once they learn the pattern, `shutdown: Sender<Never>` becomes an immediately recognized code shape.

components/script/serviceworker_manager.rs
```diff
@@ -87,6 +106,11 @@ struct ServiceWorkerRegistration {
     waiting_worker: Option<ServiceWorker>,
     /// https://w3c.github.io/ServiceWorker/#dfn-installing-worker
     installing_worker: Option<ServiceWorker>,
+    /// A channel to send control message to the worker,
+    /// currently only used to signal shutdown.
+    control_sender: Option<Sender<()>>,
+    /// A handle to join on the worker thread.
+    join_handle: Option<JoinHandle<()>>,
```
matklad, line 113: There are two places in this PR where we have optional shutdown signal, optional Thread handle, and we do “take then join” in Drop. If we wrap JoinHandle into a type which joins on Drop, than we can remove options and custom code from Drop (and one custom Drop impl altogether) and replace then with `// filed(drop) order is significant` comment.

https://github.com/rust-analyzer/ra_vfs/blob/c4a2aaaf3c2d5231b11ee21a5f6ca34c3955805c/src/io.rs#L55

gterzian: Thanks, noted, I think that's something I might look into if we end-up doing this in more places. Personally I'm not sure relying on comments and field order is necessarily more robust than some additional code like with those drop implementations...

matklad: Yeah, I also prefer to write something than to rely on implicit drop order. What swayed me for VFS case is that I was able to remove the Option(which unfortunately infected code other than Drop). If OptIon is required anyway, manual impl indeed would be better.

components/script/dom/dedicatedworkerglobalscope.rs
gterzian: I think that's a good idea, added a commit that does this...

components/script/dom/globalscope.rs
asajeffrey: My personal taste would be to use shutdown messages, and reserve dropping channels for the case where the sender has panicked. This is just taste though, it'll be the same shutdown code running in either case.

components/script/dom/globalscope.rs
asajeffrey: Another matter of taste, since I think the join only fails if the worker panics? Personally I'd log an error and try to carry on.

gterzian:
@asajeffrey Ok, comments addressed, we can always discuss the various approaches further as part of https://github.com/servo/servo/issues/26502

asajeffrey (approved):
LGTM.

gterzian:
@bors-servo retry

(the css issue I don't want to look-up right now)

PR #26630, opened 2020-05-24
Add GPUSampler to WebGPU implementation

kunalmohan:
Add dom_struct and webidl for GPUSampler, implement GPUDevice.createSampler() method.

r?@kvark

---

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -16,6 +16,9 @@ use crate::dom::bindings::codegen::Bindings::GPUDeviceBinding::{
     GPUCommandEncoderDescriptor, GPUDeviceMethods,
 };
 use crate::dom::bindings::codegen::Bindings::GPUPipelineLayoutBinding::GPUPipelineLayoutDescriptor;
+use crate::dom::bindings::codegen::Bindings::GPUSamplerBinding::{
+    GPUAddressMode, GPUCompareFunction, GPUFilterMode, GPUSamplerDescriptor,
+};
 use crate::dom::bindings::codegen::Bindings::GPUShaderModuleBinding::GPUShaderModuleDescriptor;
 use crate::dom::bindings::codegen::UnionTypes::Uint32ArrayOrString::{String, Uint32Array};
 use crate::dom::bindings::reflector::{reflect_dom_object, DomObject};
@@ -32,6 +35,7 @@ use crate::dom::gpucommandencoder::GPUCommandEncoder;
 use crate::dom::gpucomputepipeline::GPUComputePipeline;
 use crate::dom::gpupipelinelayout::GPUPipelineLayout;
 use crate::dom::gpuqueue::GPUQueue;
+use crate::dom::gpusampler::GPUSampler;
 use crate::dom::gpushadermodule::GPUShaderModule;
 use crate::script_runtime::JSContext as SafeJSContext;
 use dom_struct::dom_struct;
@@ -44,11 +48,7 @@ use std::ptr::{self, NonNull};
 use webgpu::wgpu::binding_model::{
     BindGroupEntry, BindGroupLayoutEntry, BindingResource, BindingType, BufferBinding,
 };
-use webgpu::wgt::{
-    BufferDescriptor, BufferUsage, ShaderStage, TextureComponentType, TextureFormat,
-    TextureViewDimension,
-};
-use webgpu::{WebGPU, WebGPUDevice, WebGPUQueue, WebGPURequest};
+use webgpu::{wgt, WebGPU, WebGPUDevice, WebGPUQueue, WebGPURequest, WebGPUSampler};
 
 #[dom_struct]
 pub struct GPUDevice {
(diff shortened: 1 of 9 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -45,8 +49,8 @@ use webgpu::wgpu::binding_model::{
     BindGroupEntry, BindGroupLayoutEntry, BindingResource, BindingType, BufferBinding,
 };
 use webgpu::wgt::{
-    BufferDescriptor, BufferUsage, ShaderStage, TextureComponentType, TextureFormat,
-    TextureViewDimension,
+    AddressMode, BufferDescriptor, BufferUsage, CompareFunction, FilterMode, SamplerDescriptor,
```
kvark, line 52: FWIW, I generally just import `wgt` and accessing these types via `wgt::`. This would avoid confusion with similarly named types imported from other Servo parts, e.g. JS bindings.
In the end it's up to you and Servo import guidelines to make this call :)

components/script/dom/gpudevice.rs
```diff
@@ -641,4 +645,74 @@ impl GPUDeviceMethods for GPUDevice {
...
+            .0
+            .send(WebGPURequest::CreateSampler {
+                sender,
+                device_id: self.device.0,
+                sampler_id,
+                descriptor: desc,
+            })
+            .expect("Failed to create WebGPU sampler");
+
+        let sampler = receiver.recv().unwrap();
```
kvark, line 692: One of the expected outcomes from our ID management rewrite was to *not* do `recv` here. Resource creation should be fully asynchronous:
  - we create a new ID
  - we ask the GPU side to create a resource and associate it with an ID, by sending a message
  - we return to the script, no need to wait for the GPU process to send us anything back

components/webgpu/lib.rs
```diff
@@ -414,6 +422,35 @@ impl WGPU {
...
                     }
                 },
+                WebGPURequest::CreateSampler {
+                    sender,
+                    device_id,
+                    sampler_id,
+                    descriptor,
+                } => {
+                    let global = &self.global;
+                    let desc = SamplerDescriptor {
```
kvark, line 432: you should be able to do `descriptor.map_label(...)` here

kvark (changes requested):

components/script/dom/gpudevice.rs
kunalmohan: This is a good idea. I'll make the changes.

components/script/dom/gpudevice.rs
kunalmohan: Sorry, I did not realize that. I'll remove the sender and receiver and make a new `WebGPUSampler` object in the script itself.

kunalmohan: I guess we can refactor this for the existing resources too, without needing to do anything described in https://github.com/servo/servo/issues/25472. I'll open a separate PR for the for those.

kvark: Yes, that's correct. It's totally fine to do separately, just wanted to point out that this is important.

components/script/dom/gpudevice.rs
```diff
@@ -127,9 +127,9 @@ impl GPUDevice {
         } else {
             (
                 false,
-                BufferDescriptor {
+                wgt::BufferDescriptor {
                     size: 0,
-                    usage: BufferUsage::STORAGE,
+                    usage: wgt::BufferUsage::STORAGE,
```
kvark, line 132: if this is meant to be empty, we should use `wgt::BufferUsage::empty()`

components/script/dom/identityhub.rs
```diff
@@ -77,6 +79,10 @@ impl IdentityHub {
     fn create_command_encoder_id(&mut self) -> CommandEncoderId {
         self.command_encoders.alloc(self.backend)
     }
+
+    fn create_sampler_id(&mut self) -> SamplerId {
```
kvark, line 83: We shouldn't need these helper methods. The caller is `Identities::crate_sampler_id`, and it can do:
```rust
self.select(backend).samplers.alloc(backend)
```
Moreover, we may not even need `IdentityHubb::backend` field

components/webgpu/lib.rs
```diff
@@ -16,18 +16,21 @@ use malloc_size_of::{MallocSizeOf, MallocSizeOfOps};
...
     id::{
         AdapterId, BindGroupId, BindGroupLayoutId, BufferId, CommandBufferId, CommandEncoderId,
-        ComputePipelineId, DeviceId, PipelineLayoutId, QueueId, ShaderModuleId,
+        ComputePipelineId, DeviceId, PipelineLayoutId, QueueId, SamplerId, ShaderModuleId,
     },
     instance::RequestAdapterOptions,
 };
-use wgt::{BufferAddress, BufferDescriptor, CommandBufferDescriptor, DeviceDescriptor};
+use wgt::{
+    BufferAddress, BufferDescriptor, CommandBufferDescriptor, DeviceDescriptor, SamplerDescriptor,
```
kvark, line 32: Similarly, please consider using `id` as a namespace and `wgt` as a namespace.
One of the good benefits in the long run is reduced churn on these use statements: when rebasing and merging, these often come up with conflicts. Using the namespaces directly solves that.

components/webgpu/lib.rs
```diff
@@ -414,6 +422,17 @@ impl WGPU {
...
                     }
                 },
+                WebGPURequest::CreateSampler {
+                    device_id,
+                    sampler_id,
+                    descriptor,
+                } => {
+                    let global = &self.global;
+                    let desc = descriptor.map_label(move |l| {
+                        CString::new(l.as_bytes()).unwrap().as_c_str().as_ptr()
```
kvark, line 432: this looks wrong: the `CString` will be dropped before the pointer is used.
You can use something [like this](https://github.com/gfx-rs/wgpu/blob/70154373f8668b88c43fcbdee7bee8a89816eb39/player/src/main.rs#L242-L244) instead.

kvark (changes requested):

kvark (approved):
Beautiful, thank you!
I love the way it gets shaped :) :+1:

jdm:
@bors-servo retry
* #23290

Issue #26550, opened 2020-05-17
Replace uses of "static mut" with UnsafeCell


jdm:
Given https://github.com/rust-lang/rust/issues/53639, it seems safer to rely on tools that won't let us accidentally alias mutable data like UnsafeCell.

* [ ] components/background_hang_monitor/sampler_linux.rs
* [x] components/profile_traits/energy.rs
* [ ] components/script/dom/eventtarget.rs
* [ ] components/script/dom/bindings/codegen/CodegenRust.py
* [x] components/profile/heartbeats.rs

SimonSapin:
`UnsafeCell` is `!Sync`, so it can’t be used in a `static` directly. The most direct port would be to make a type that wraps `UnsafeCell` and implements `Sync`, but this transition seems to be a good time to reexamine the soundness of that `unsafe` code. It looks like some of these uses could be `AtomicPtr`.

SimonSapin:
I did the examining. Here is what I found, in increasing levels of worrisome. **TL;DR: see the parts in bold.**

## `components/script/dom/eventtarget.rs`

No mutability there, using `const` instead of `static mut` just works. (Using `static` complains that `*const c_char: !Sync`.)

## `components/script/dom/bindings/codegen/CodegenRust.py`

This generates code for "proxy handlers", pointers that are initialized once during startup in the `script::init::init` function, and then never modified again. They are read in generated bindings code. Two options here to remove `static mut`:

* Use `AtomicPtr` in `static`. This would require no unsafe code, but would theoretically have run-time cost in bindings code because previously-unsynchronized reads are now atomic. I don’t have a good sense of how hot this code is, or how big/small the overhead is for uncontended atomic operations. **Does making reads of proxy handlers atomic in DOM bindings sound acceptable?**

* To keep those reads unsynchronized, we could make a type that wraps `UnsafeCell` and implements `Sync`, to use it in a `static`. Its APIs would be `unsafe fn` and require users to ensure external synchronization. In this case, `script::init::init` must not run concurrently with bindings code. Note that we already have a code comment in that vein:

  https://github.com/servo/servo/blob/b0bf182316292b0d76f7c87e322370331ad0497f/components/servo/lib.rs#L440-L443

Either way, if somehow we fail to run initialization code before bindings code, we’ll end up giving null pointers to SpiderMonkey for proxy handlers. I don’t know if SpiderMonkey checks for null (possibly to safely crash) or if this would be UB.

[...]

PR #26641, opened 2020-05-25
Fix undefined behavior in `energymon::init`

dylni:
The layout of [`Box`](https://doc.rust-lang.org/std/boxed/struct.Box.html) is implicitly [`#[repr(Rust)]`](https://doc.rust-lang.org/nomicon/repr-rust.html), so it is undefined behavior to transmute it to another type. [`Box::into_raw`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw) can safely convert it to a pointer.

I did not run the test suite locally. I thought it would be better to let CI run than to set up the development environment for such a simple change. I did test that the file works individually with this change.

---

- [X] These changes do not require tests because the undefined behavior is not visible

### DIFF
--- a/components/profile_traits/energy.rs
+++ b/components/profile_traits/energy.rs
@@ -29,7 +29,6 @@ mod energymon {
 
     use self::energy_monitor::EnergyMonitor;
     use self::energymon::EnergyMon;
-    use std::mem;
     use std::sync::{Once, ONCE_INIT};
 
     static mut EM: Option<*mut EnergyMon> = None;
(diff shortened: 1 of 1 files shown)


SimonSapin:
Looks good, thanks!

The remaining `unsafe` code in this file still looks very suspicious to me regarding thread-safety, but we can leave that for https://github.com/servo/servo/issues/26550.

@bors-servo r+

dylni:
Nice to see the whole `static mut` being reconsidered! It's interesting how contemporaneous it is with this PR.

CYBAI:
@bors-servo retry

- #26649

hxuhack:
After replacing mem::transmute(Box::new(em)) with Box::into_raw(Box::new(em)), can we remove the surrounded unsafe marker? @dylni

SimonSapin:
If that were the case we’d have an "unnecessary `unsafe`" warning.

That `unsafe` block is still required to assign to `EM` which is a `static mut` global variable that doesn’t itself has synchronization making it thread-safe. Here synchronization is provided by `Once`.

hxuhack:
Oh, right. Thank you for the clarification!

PR #26659, opened 2020-05-26
Add support for remaining animation and transition events

### LINKED ISSUE #21564: Implement animation events

mrobinson:
This PR adds support for remaining animation and transitions events.
There are two commits here. The first is a bit more complicated: it reworks
how rooting is done for animating nodes. Instead of having the `ScriptThread`
try to track which animations are active via events (which can be inaccurate),
it just maintains roots for nodes that are actually present in the animations-
-related data structures. The second commit adds support for the new events.

Unfortunately, the existing events tests either rely on the Web Animations API
or other behavior (for example, that changing animation delay restarts
an animation). Since those two things are out-of-scope for this change,
I've forked some of the WPT tests, removed the reliance on the Web
Animations API, and added them to Servo's internal tests.

---

- [x] These changes fix #21564.

### DIFF
--- a/components/script/animations.rs
+++ b/components/script/animations.rs
@@ -78,12 +78,15 @@ impl Animations {
         let mut sets = self.sets.write();
 
         for set in sets.values_mut() {
+            self.start_pending_animations(set, now, pipeline_id);
+
             // When necessary, iterate our running animations to the next iteration.
             for animation in set.animations.iter_mut() {
                 if animation.iterate_if_necessary(now) {
                     self.add_animation_event(
                         animation,
                         TransitionOrAnimationEventType::AnimationIteration,
+                        now,
                         pipeline_id,
                     );
                 }
@@ -97,7 +100,6 @@ impl Animations {
 
     /// Processes any new animations that were discovered after reflow. Collect messages
     /// that trigger events for any animations that changed state.
-    /// TODO(mrobinson): The specification dictates that this should happen before reflow.
     pub(crate) fn do_post_reflow_update(&self, window: &Window, now: f64) {
         let pipeline_id = window.pipeline_id();
         let mut sets = self.sets.write();
@@ -140,6 +142,39 @@ impl Animations {
             .sum()
     }
 
+    /// Walk through the list of pending animations and start all of the ones that
+    /// have left the delay phase.
+    fn start_pending_animations(
+        &self,
+        set: &mut ElementAnimationSet,
+        now: f64,
+        pipeline_id: PipelineId,
+    ) {
+        for animation in set.animations.iter_mut() {
+            if animation.state == AnimationState::Pending && animation.started_at <= now {
+                animation.state = AnimationState::Running;
+                self.add_animation_event(
+                    animation,
+                    TransitionOrAnimationEventType::AnimationStart,
+                    now,
+                    pipeline_id,
+                );
+            }
+        }
+
+        for transition in set.transitions.iter_mut() {
+            if transition.state == AnimationState::Pending && transition.start_time <= now {
+                transition.state = AnimationState::Running;
+                self.add_transition_event(
+                    transition,
+                    TransitionOrAnimationEventType::TransitionStart,
+                    now,
+                    pipeline_id,
+                );
+            }
+        }
+    }
+
     /// Walk through the list of running animations and remove all of the ones that
     /// have ended.
     fn finish_running_animations(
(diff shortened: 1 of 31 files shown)


components/script/animations.rs
```diff
@@ -6,141 +6,175 @@
...
-#[derive(Clone, Debug, Default, JSTraceable)]
+#[derive(Default, JSTraceable)]
+#[unrooted_must_root_lint::must_root]
 pub(crate) struct Animations {
+    /// The map of nodes to their animation states.
     pub sets: Arc<RwLock<FxHashMap<OpaqueNode, ElementAnimationSet>>>,
-    have_running_animations: bool,
+
+    /// Whether or not we have animations that are running.
+    have_running_animations: DomRefCell<bool>,
```
jdm, line 46: This can be a Cell.

components/script/animations.rs
```diff
@@ -6,141 +6,175 @@
...
-            update.collect_newly_animating_nodes(&sets);
-
-            for set in sets.values_mut() {
-                Self::handle_canceled_animations(set, now, &mut update);
-                Self::handle_new_animations(set, &mut update);
-            }
+    pub(crate) fn do_post_reflow_update(
+        &self,
+        window: &Window,
+        js_runtime: *mut JSRuntime,
```
jdm, line 109: Let's take a &Runtime argument here, rather than a raw pointer. We could even add a getter to Window and remove the argument.

components/script/dom/document.rs
```diff
@@ -3780,10 +3780,12 @@ impl Document {
         self.animations.borrow()
     }
 
-    pub(crate) fn update_animations_post_reflow(&self) -> AnimationsUpdate {
-        self.animations
-            .borrow_mut()
-            .do_post_reflow_update(&self.window, self.current_animation_timeline_value())
+    pub(crate) fn update_animations_post_reflow(&self, js_runtime: *mut JSRuntime) {
```
jdm, line 3783: Let's add a getter to Window for the JS runtime instead so we don't need to pass around an unsafe pointer.

components/script/animations.rs
```diff
@@ -171,103 +204,195 @@ impl Animations {
...
-                num_new_animations += state
-                    .transitions
-                    .iter()
-                    .filter(|transition| transition.is_new)
-                    .count();
-
-                let node = UntrustedNodeAddress(node.0 as *const c_void);
-                std::iter::repeat(node).take(num_new_animations)
-            }));
+impl MallocSizeOf for Animations {
```
jdm, line 455: Can we derive this instead and use the `#[ignore_malloc_size_of]` annotation where necessary?

components/script/animations.rs
```diff
@@ -171,103 +204,195 @@ impl Animations {
...
-impl MallocSizeOf for Animations {
-    fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
-        self.sets.read().size_of(ops) + self.have_running_animations.size_of(ops)
+    /// Ensure that all nodes with new animations are rooted. This should be called
+    /// immediately after a restyle, to ensure that these addresses are still valid.
+    #[allow(unsafe_code)]
+    fn root_newly_animating_dom_nodes(
+        &self,
+        sets: &FxHashMap<OpaqueNode, ElementAnimationSet>,
+        js_runtime: *mut JSRuntime,
```
jdm, line 282: Can we accept a `&Runtime`argument instead so we don't pass around raw pointers? We can use [rt](https://doc.servo.org/mozjs/rust/struct.Runtime.html#method.rt) to get the right pointer from it.

jdm (changes requested):
Great work!

components/script/animations.rs
mrobinson: :+1:

components/script/animations.rs
mrobinson: Okay. This makes things quite a bit simpler, so I'll add a getter and pass around `&Window` everywhere.

components/script/dom/document.rs
mrobinson: :+1:

components/script/animations.rs
mrobinson: Sure. This means we won't measure the size of the animation sets, but perhaps it makes sense to add a new directive to `malloc_size_of_derive` to make it easier to fix this in a foolproof way.

components/script/animations.rs
mrobinson: :+1:

mrobinson:
Thanks for the review. I've pushed a new version taking into account your suggestions.

CYBAI:
@bors-servo retry

- #25533

PR #26660, opened 2020-05-26
Make WebGPU resource creation async and prevent panic during shutdown if WebGPU is enabled.

### LINKED ISSUE #25472: Try making creation of GPUBuffer async(even without a promise)

kunalmohan:
1. Make WebGPU resource creation async.
2. Remove some unused code in `WebGPURequest::RequestAdapter`.
3. Prevent panic during shutdown. Since WGPU thread is killed before script, sender and receiver in the script panic at either of the two places-
a. If a buffer is still alive, script tries to send `WebGPURequest::DestroyBuffer` to server while dropping the buffer during shutdown. 
https://github.com/servo/servo/blob/7170a69695212ba8c24a9f0c4202f0ddfe52f742/components/script/dom/gpubuffer.rs#L118-L122 https://github.com/servo/servo/blob/7170a69695212ba8c24a9f0c4202f0ddfe52f742/components/script/dom/gpubuffer.rs#L182-L186
b. Receiver in script-thread panics with `RecvError` as soon as sender on server side is dropped. https://github.com/servo/servo/blob/7170a69695212ba8c24a9f0c4202f0ddfe52f742/components/script/script_thread.rs#L1456-L1457

r?@kvark

---

- [X] These changes fix #25472 (GitHub issue number if applicable)

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -2016,6 +2016,7 @@ impl ScriptThread {
             WebGPUMsg::FreeCommandBuffer(id) => self.gpu_id_hub.lock().kill_command_buffer_id(id),
             WebGPUMsg::FreeSampler(id) => self.gpu_id_hub.lock().kill_sampler_id(id),
             WebGPUMsg::FreeShaderModule(id) => self.gpu_id_hub.lock().kill_shader_module_id(id),
+            WebGPUMsg::Exit => *self.webgpu_port.borrow_mut() = None,
             _ => {},
         }
     }
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -179,10 +179,16 @@ impl GPUBufferMethods for GPUBuffer {
             },
             _ => {},
         };
-        self.channel
+        if let Err(e) = self
+            .channel
             .0
             .send(WebGPURequest::DestroyBuffer(self.buffer.0))
-            .unwrap();
+        {
+            warn!(
+                "Failed to send WebGPURequest::DestroyBuffer({:?}) ({})",
+                self.buffer.0, e
+            );
+        };
         *self.state.borrow_mut() = GPUBufferState::Destroyed;
     }
 
--- a/components/script/dom/gpucommandencoder.rs
+++ b/components/script/dom/gpucommandencoder.rs
@@ -16,11 +16,10 @@ use crate::dom::gpubuffer::GPUBuffer;
 use crate::dom::gpucommandbuffer::GPUCommandBuffer;
 use crate::dom::gpucomputepassencoder::GPUComputePassEncoder;
 use dom_struct::dom_struct;
-use ipc_channel::ipc;
 use std::cell::Cell;
 use std::collections::HashSet;
 use webgpu::wgt::BufferUsage;
-use webgpu::{WebGPU, WebGPUCommandEncoder, WebGPURequest};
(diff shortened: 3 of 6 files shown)


kvark (approved):
Looks great!

We still have a few senders in message variants that we need to consider fro removal, for example one in CommandEncoderFinish seems unnecessary.

kunalmohan:
Right. I must have missed that. I'll remove that one here itself.

components/script/script_thread.rs
```diff
@@ -2016,6 +2016,7 @@ impl ScriptThread {
             WebGPUMsg::FreeCommandBuffer(id) => self.gpu_id_hub.lock().kill_command_buffer_id(id),
             WebGPUMsg::FreeSampler(id) => self.gpu_id_hub.lock().kill_sampler_id(id),
             WebGPUMsg::FreeShaderModule(id) => self.gpu_id_hub.lock().kill_shader_module_id(id),
+            WebGPUMsg::Exit => *self.webgpu_port.borrow_mut() = None,
```
gterzian, line 2019: How about, instead of an exit message, we do something like:

```rust
if msg.is_err() {
    *self.webgpu_port.borrow_mut() = None;
    return true;
}
FromWebGPUServer(msg.unwrap())
```

Just in case that also works, otherwise it's not a big deal, however I am trying to not add more explicit exit messages in the light of https://github.com/servo/servo/issues/26502

gterzian: Ah it's awaiting merge already, nevermind, another time...

PR #26668, opened 2020-05-27
Set `currentScript` to `null` for module scripts

CYBAI:
I misunderstood the test case when I worked on #23545. That test case is
actually not related to dynamic import; instead, the reason why it
crashes is, `currentScript` should be updated to `null`.

In spec, the step 6 of [execute-the-script-block](https://html.spec.whatwg.org/multipage/scripting.html#execute-the-script-block) only says `Assert: document's currentScript attribute is null.` but doesn't says it should be set to null. Not sure if it can be improved.

---

### DIFF
--- a/components/script/dom/htmlscriptelement.rs
+++ b/components/script/dom/htmlscriptelement.rs
@@ -757,13 +757,17 @@ impl HTMLScriptElement {
         let old_script = document.GetCurrentScript();
 
         match script.type_ {
+            ScriptType::Classic => document.set_current_script(Some(self)),
+            ScriptType::Module => document.set_current_script(None),
+        }
+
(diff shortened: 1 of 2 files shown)


CYBAI:
Also, I found [this commit](https://github.com/whatwg/html/pull/1000/commits/177d30dcbfbdbdb39938868717b6972f31966ca4) in spec PR says "They set document.currentScript to null". I don't understand why it's not specced instead 🤔 

(there's another [spec issue](https://github.com/whatwg/html/issues/1013) discussion about how module script can have a `currentScript`-like pointer)

CYBAI:
Hmm, is @highfive OOO today 🤔?

r? @Manishearth or @jdm

Issue #26661, opened 2020-05-26
Tracking: upgrade to nightly-2020-05-26


SimonSapin:
Blocked on:

* [x] Installing the `llvm-tools-preview` component https://github.com/servo/servo/issues/26645#issuecomment-633829852, https://github.com/servo/servo/pull/26662
* [x] Handling `syn::Type::Group` in `derive_common::cg::map_type_params` https://github.com/servo/servo/pull/26662
* [x] rustfmt being available https://github.com/rust-lang/rust/issues/72232
* [x] ICE in `gfx-backend-metal` https://github.com/servo/servo/issues/26648, https://github.com/rust-lang/rust/issues/72467
* [x] Upgrade parking-lot to 0.10 https://github.com/servo/servo/issues/26644, https://github.com/servo/servo/pull/26646. Work around: disable its `nightly` feature (which enables hardware lock elision) 
* [x] Exhaustiveness check regression https://github.com/rust-lang/rust/issues/72476. Work around:
  ```diff
  --- ports/glutin/headed_window.rs
  +++ ports/glutin/headed_window.rs
  @@ -543,10 +543,14 @@ impl WindowMethods for Window {
   
           #[cfg(target_os = "linux")]
           return match native_context {
  -            NativeContext::Default(NativeContext::Default(native_context)) =>
  -                PlayerGLContext::Egl(native_context.egl_context as usize),
  -            NativeContext::Default(NativeContext::Alternate(native_context)) =>
  -                PlayerGLContext::Egl(native_context.egl_context as usize),
  +            NativeContext::Default(d) => match d {
  +                NativeContext::Default(native_context) => {
  +                    PlayerGLContext::Egl(native_context.egl_context as usize)
  +                },
  +                NativeContext::Alternate(native_context) => {
  +                    PlayerGLContext::Egl(native_context.egl_context as usize)
  +                },
  +            },
               NativeContext::Alternate(_) => unimplemented!(),
           };
   
  @@ -574,10 +578,10 @@ impl WindowMethods for Window {
   
           #[cfg(target_os = "linux")]
           return match native_connection {
  -            NativeConnection::Default(NativeConnection::Default(conn)) =>
  -                NativeDisplay::Egl(conn.0 as usize),
  -            NativeConnection::Default(NativeConnection::Alternate(conn)) =>
  -                NativeDisplay::X11(conn.x11_display as usize),
  +            NativeConnection::Default(d) => match d {
[...]

PR #26698, opened 2020-05-28
Remove some unused dependency declarations

SimonSapin:
This is based on compiling with `RUSTFLAGS="-W unused_crate_dependencies"` (CC https://github.com/rust-lang/rust/pull/72342) in a recent Nightly (more so than used in the tree as of this writing, CC https://github.com/servo/servo/issues/26661 for work-arounds).

Only one crate is actually removed from the dependency graph, others are still dependended from other places.

### DIFF
--- a/components/style/Cargo.toml
+++ b/components/style/Cargo.toml
@@ -16,7 +16,8 @@ path = "lib.rs"
 doctest = false
 
 [features]
-gecko = ["style_traits/gecko", "fallible/known_system_malloc", "bindgen", "regex", "toml"]
+gecko = ["style_traits/gecko", "fallible/known_system_malloc", "bindgen", "regex", "toml",
+         "num_cpus", "thin-slice"]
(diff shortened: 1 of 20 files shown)


components/style/Cargo.toml
```diff
@@ -79,7 +80,7 @@ void = "1.0.2"
 
 [build-dependencies]
 lazy_static = "1"
-log = "0.4"
+#log = "0.4"
```
atouchet, line 83: Was this supposed to be removed?

SimonSapin: I didn’t mean to commit this. Reverted, thanks.

`log` is used unconditionally in the library crate, but only in Gecko mode in the build script. So having this in `build-dependencies` warns in Servo mode. I didn’t bother to try if Cargo supports having a dependency optional in `build-dependencies` but non-optional in `depnedencies`.

PR #26699, opened 2020-05-28
Avoid bare *mut JSObject arguments in webidl codegen

### LINKED ISSUE #13308: WebIDL: Represent optional objects as `Option<NonZero<*mut JSObject>>` in function arguments.
It's done the same way in return types, but for some reason for arguments we're still passing `*mut JSObject`, and passing the raw `null` pointer.

cc @nox

jdm:
The motivation here was fixing a bug where argument type calculation did not treat inner types of sequences correctly. This required updating some logic for when to choose to use `*mut JSObject` vs. `RootedTraceableBox<Heap<*mut JSObject>`, which then was a good reason to choose `HandleObject` when possible instead.

---
- [x] These changes fix #13308

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -677,7 +677,7 @@ def getJSToNativeConversionInfo(type, descriptorProvider, failureCode=None,
     if type.isSequence() or type.isRecord():
         innerInfo = getJSToNativeConversionInfo(innerContainerType(type),
                                                 descriptorProvider,
-                                                isMember=isMember,
+                                                isMember="Sequence",
                                                 isAutoRooted=isAutoRooted)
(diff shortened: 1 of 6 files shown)


jdm:
cc @kunalmohan 
This should allow you to update the webidl assertion like we discussed on matrix.

jdm:
The crashes come from creating Handle values from stack temporary pointers (ie. taking the result of `some_jsval_handle.to_object()` and calling `HandleObject::from_marked_location(&...)`). If we want to go through with this, we will need to create stack roots for the result of the conversion, then take a handle from those roots.

nox:
This still needs a rebase (for once that I see old stuff on which I can ping you and not the reverse… 🤡)

jdm:
It needs more than a rebase. It needs a fundamental rethink.

Issue #25209, opened 2019-12-08
Implement Blob get_stream


gterzian:
Spec: https://w3c.github.io/FileAPI/#blob-get-stream

Servo currently has a `get_bytes` method that roughly correspond to this, and is used in many places(for example in `FileReader`) in lieu of `get_stream`: https://github.com/servo/servo/blob/d92f705573ac26fb8e16abc18d7d19e1fb384715/components/script/dom/blob.rs#L141

I think we should be able to replace the use of `get_bytes` with `get_stream` everywhere. 

In at least one case, when [`stream()`](https://w3c.github.io/FileAPI/#dom-blob-stream) is called, the stream will be directly passed to the JS, in other cases, for example when the filereader [calls `get_stream`](https://w3c.github.io/FileAPI/#ref-for-blob-get-stream%E2%91%A2) directly, it would just be used from Rust. 

I would say that in both cases, we can use a Rust struct that would implement [`ReadableStreamUnderlyingSource`](https://github.com/servo/mozjs/blob/e21c05b415dfc246175ff8d5fc48b0e8c5b4e9e9/mozjs/js/public/Stream.h#L79).

gterzian:
Blob now has a `get_stream` method, 

https://github.com/servo/servo/blob/00b57b4fd95ad53859c52f9be19d94ae16bed9c2/components/script/dom/blob.rs#L92

so what is required is using it in `FileReader`, over at https://github.com/servo/servo/blob/00b57b4fd95ad53859c52f9be19d94ae16bed9c2/components/script/dom/filereader.rs#L449

and you'd have to do something similar to https://github.com/servo/servo/blob/00b57b4fd95ad53859c52f9be19d94ae16bed9c2/components/script/body.rs#L222

where you would queue tasks as data was read, the stream done, and so on, as opposed to the IPC used there.

gterzian:
I'll do this one next.

atouchet:
Did #35969 finish this?

PR #26710, opened 2020-05-29
Compile external scripts off the main thread

### LINKED ISSUE #26571: Compile non-inline non-module scripts off thread

AbhishekSharma102:
---

- [ ] These changes fix #26571  (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/bindings/trace.rs
+++ b/components/script/dom/bindings/trace.rs
@@ -77,8 +77,10 @@ use hyper::Method;
 use hyper::StatusCode;
 use indexmap::IndexMap;
 use ipc_channel::ipc::{IpcReceiver, IpcSender};
-use js::glue::{CallObjectTracer, CallStringTracer, CallValueTracer};
-use js::jsapi::{GCTraceKindToAscii, Heap, JSObject, JSString, JSTracer, JobQueue, TraceKind};
+use js::glue::{CallObjectTracer, CallScriptTracer, CallStringTracer, CallValueTracer};
+use js::jsapi::{
+    GCTraceKindToAscii, Heap, JSObject, JSScript, JSString, JSTracer, JobQueue, TraceKind,
+};
 use js::jsval::JSVal;
 use js::rust::{GCMethods, Handle, Runtime};
 use js::typedarray::TypedArray;
@@ -219,6 +221,18 @@ unsafe_no_jsmanaged_fields!(Cow<'static, str>);
 
 unsafe_no_jsmanaged_fields!(CspList);
 
+/// Trace a `JSScript`.
+pub fn trace_script(tracer: *mut JSTracer, description: &str, script: &Heap<*mut JSScript>) {
+    unsafe {
+        trace!("tracing {}", description);
+        CallScriptTracer(
+            tracer,
+            script.ptr.get() as *mut _,
+            GCTraceKindToAscii(TraceKind::Script),
+        );
+    }
+}
+
 /// Trace a `JSVal`.
 pub fn trace_jsval(tracer: *mut JSTracer, description: &str, val: &Heap<JSVal>) {
     unsafe {
@@ -325,6 +339,15 @@ unsafe impl<T: JSTraceable> JSTraceable for DomRefCell<T> {
 unsafe impl<T: JSTraceable> JSTraceable for RefCell<T> {
     unsafe fn trace(&self, trc: *mut JSTracer) {
         (*self).borrow().trace(trc)
+    }
+}
+
+unsafe impl JSTraceable for Heap<*mut JSScript> {
+    unsafe fn trace(&self, trc: *mut JSTracer) {
+        if self.get().is_null() {
+            return;
+        }
+        trace_script(trc, "heap script", self);
     }
 }
 
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -38,7 +38,7 @@ use crate::dom::file::File;
 use crate::dom::gpudevice::GPUDevice;
 use crate::dom::gpuoutofmemoryerror::GPUOutOfMemoryError;
 use crate::dom::gpuvalidationerror::GPUValidationError;
-use crate::dom::htmlscriptelement::ScriptId;
+use crate::dom::htmlscriptelement::{ScriptId, SourceCode};
 use crate::dom::identityhub::Identities;
 use crate::dom::imagebitmap::ImageBitmap;
 use crate::dom::messageevent::MessageEvent;
@@ -91,9 +91,9 @@ use js::jsval::PrivateValue;
 use js::jsval::{JSVal, UndefinedValue};
 use js::panic::maybe_resume_unwind;
 use js::rust::transform_str_to_source_text;
-use js::rust::wrappers::{JS_ExecuteScript, JS_GetScriptPrivate};
+use js::rust::wrappers::{JS_ExecuteScript, JS_ExecuteScript1, JS_GetScriptPrivate};
 use js::rust::{get_object_class, CompileOptionsWrapper, ParentRuntime, Runtime};
-use js::rust::{HandleValue, MutableHandleValue};
+use js::rust::{HandleScript, HandleValue, MutableHandleValue};
 use js::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
 use msg::constellation_msg::{
     BlobId, BroadcastChannelRouterId, MessagePortId, MessagePortRouterId, PipelineId,
@@ -2552,8 +2552,9 @@ impl GlobalScope {
         fetch_options: ScriptFetchOptions,
         script_base_url: ServoUrl,
     ) -> bool {
+        let source_code = SourceCode::Text(Rc::new(DOMString::from_string((*code).to_string())));
         self.evaluate_script_on_global_with_result(
-            code,
+            &source_code,
             "",
             rval,
             1,
@@ -2566,7 +2567,7 @@ impl GlobalScope {
     #[allow(unsafe_code)]
     pub fn evaluate_script_on_global_with_result(
         &self,
-        code: &str,
+        code: &SourceCode,
         filename: &str,
         rval: MutableHandleValue,
         line_number: u32,
@@ -2595,48 +2596,58 @@ impl GlobalScope {
                 let _aes = AutoEntryScript::new(self);
 
                 unsafe {
-                    let options = CompileOptionsWrapper::new(*cx, filename.as_ptr(), line_number);
+                    let result = match code {
+                        SourceCode::Text(text_code) => {
+                            let options =
+                                CompileOptionsWrapper::new(*cx, filename.as_ptr(), line_number);
 
-                    debug!("compiling Dom string");
-                    rooted!(in(*cx) let compiled_script =
-                        Compile1(
-                            *cx,
-                            options.ptr,
-                            &mut transform_str_to_source_text(code),
-                        )
-                    );
+                            debug!("compiling Dom string");
+                            rooted!(in(*cx) let compiled_script =
+                                Compile1(
+                                    *cx,
+                                    options.ptr,
+                                    &mut transform_str_to_source_text(text_code),
+                                )
+                            );
 
-                    if compiled_script.is_null() {
-                        debug!("error compiling Dom string");
-                        report_pending_exception(*cx, true, InRealm::Entered(&ar));
-                        return false;
-                    }
+                            if compiled_script.is_null() {
+                                debug!("error compiling Dom string");
+                                report_pending_exception(*cx, true, InRealm::Entered(&ar));
+                                return false;
+                            }
 
-                    rooted!(in(*cx) let mut script_private = UndefinedValue());
-                    JS_GetScriptPrivate(*compiled_script, script_private.handle_mut());
+                            rooted!(in(*cx) let mut script_private = UndefinedValue());
+                            JS_GetScriptPrivate(*compiled_script, script_private.handle_mut());
 
-                    // When `ScriptPrivate` for the compiled script is undefined,
-                    // we need to set it so that it can be used in dynamic import context.
-                    if script_private.is_undefined() {
-                        debug!("Set script private for {}", script_base_url);
+                            // When `ScriptPrivate` for the compiled script is undefined,
+                            // we need to set it so that it can be used in dynamic import context.
+                            if script_private.is_undefined() {
+                                debug!("Set script private for {}", script_base_url);
 
-                        let module_script_data = Rc::new(ModuleScript::new(
-                            script_base_url,
-                            fetch_options,
-                            // We can't initialize an module owner here because
-                            // the executing context of script might be different
-                            // from the dynamic import script's executing context.
-                            None,
-                        ));
+                                let module_script_data = Rc::new(ModuleScript::new(
+                                    script_base_url,
+                                    fetch_options,
+                                    // We can't initialize an module owner here because
+                                    // the executing context of script might be different
+                                    // from the dynamic import script's executing context.
+                                    None,
+                                ));
 
-                        SetScriptPrivate(
-                            *compiled_script,
-                            &PrivateValue(Rc::into_raw(module_script_data) as *const _),
-                        );
-                    }
+                                SetScriptPrivate(
+                                    *compiled_script,
+                                    &PrivateValue(Rc::into_raw(module_script_data) as *const _),
+                                );
+                            }
 
-                    debug!("evaluating Dom string");
-                    let result = JS_ExecuteScript(*cx, compiled_script.handle(), rval);
+                            debug!("evaluating Dom string");
+                            JS_ExecuteScript(*cx, compiled_script.handle(), rval)
+                        },
+                        SourceCode::Compiled(compiled_script) => {
+                            let script = compiled_script.source_code.get();
+                            let script_handle = HandleScript::new(&script);
+                            JS_ExecuteScript1(*cx, script_handle)
+                        },
+                    };
 
                     if !result {
                         debug!("error evaluating Dom string");
--- a/components/script/dom/htmlscriptelement.rs
+++ b/components/script/dom/htmlscriptelement.rs
@@ -13,6 +13,7 @@ use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::root::{Dom, DomRoot};
 use crate::dom::bindings::settings_stack::AutoEntryScript;
 use crate::dom::bindings::str::{DOMString, USVString};
+use crate::dom::bindings::trace::RootedTraceableBox;
 use crate::dom::document::Document;
 use crate::dom::element::{
     cors_setting_for_element, reflect_cross_origin_attribute, set_cross_origin_attribute,
@@ -27,15 +28,25 @@ use crate::dom::performanceresourcetiming::InitiatorType;
 use crate::dom::virtualmethods::VirtualMethods;
 use crate::fetch::create_a_potential_cors_request;
 use crate::network_listener::{self, NetworkListener, PreInvoke, ResourceTimingListener};
+use crate::realms::enter_realm;
 use crate::script_module::fetch_inline_module_script;
 use crate::script_module::{fetch_external_module_script, ModuleOwner, ScriptFetchOptions};
+use crate::task::TaskCanceller;
+use crate::task_source::dom_manipulation::DOMManipulationTaskSource;
+use crate::task_source::TaskSource;
+use crate::task_source::TaskSourceName;
 use content_security_policy as csp;
+use core::ffi::c_void;
 use dom_struct::dom_struct;
 use encoding_rs::Encoding;
 use html5ever::{LocalName, Prefix};
 use ipc_channel::ipc;
 use ipc_channel::router::ROUTER;
+use js::jsapi::{
+    CanCompileOffThread, CompileOffThread, FinishOffThreadScript, Heap, JSScript, OffThreadToken,
+};
 use js::jsval::UndefinedValue;
+use js::rust::{transform_u16_to_source_text, CompileOptionsWrapper};
 use msg::constellation_msg::PipelineId;
 use net_traits::request::{
     CorsSettings, CredentialsMode, Destination, ParserMetadata, RequestBuilder,
@@ -43,17 +54,95 @@ use net_traits::request::{
 use net_traits::{FetchMetadata, FetchResponseListener, Metadata, NetworkError};
 use net_traits::{ResourceFetchTiming, ResourceTimingType};
 use servo_atoms::Atom;
+use servo_config::pref;
 use servo_url::ImmutableOrigin;
 use servo_url::ServoUrl;
 use std::cell::Cell;
 use std::fs::{create_dir_all, read_to_string, File};
 use std::io::{Read, Seek, Write};
+use std::mem::replace;
 use std::path::PathBuf;
 use std::process::Command;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 use style::str::{StaticStringVec, HTML_SPACE_CHARACTERS};
 use uuid::Uuid;
+
+pub struct OffThreadCompilationContext {
+    script_element: Trusted<HTMLScriptElement>,
+    script_kind: ExternalScriptKind,
+    final_url: ServoUrl,
+    url: ServoUrl,
+    task_source: DOMManipulationTaskSource,
+    canceller: TaskCanceller,
+    script_text: String,
+    utf16_chars: Vec<u16>,
+    fetch_options: ScriptFetchOptions,
+}
+
+/// A wrapper to mark OffThreadToken as Send,
+/// which should be safe according to
+/// mozjs/js/public/OffThreadScriptCompilation.h
+struct OffThreadCompilationToken(*mut OffThreadToken);
+
+#[allow(unsafe_code)]
+unsafe impl Send for OffThreadCompilationToken {}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn off_thread_compilation_callback(
+    token: *mut OffThreadToken,
+    callback_data: *mut c_void,
+) {
+    let mut context = Box::from_raw(callback_data as *mut OffThreadCompilationContext);
+    let token = OffThreadCompilationToken(token);
+
+    let url = context.url.clone();
+    let final_url = context.final_url.clone();
+    let script_element = context.script_element.clone();
+    let script_kind = context.script_kind.clone();
+    let script = replace(&mut context.script_text, String::new());
+    let fetch_options = context.fetch_options.clone();
+
+    // Continue with <https://html.spec.whatwg.org/multipage/#fetch-a-classic-script>
+    let _ = context.task_source.queue_with_canceller(
+        task!(off_thread_compile_continue: move || {
+            let elem = script_element.root();
+            let global = elem.global();
+            let cx = global.get_cx();
+            let _ar = enter_realm(&*global);
+
+            rooted!(in(*cx)
+                let compiled_script = FinishOffThreadScript(*cx, token.0)
+            );
+
+            let load = if compiled_script.get().is_null() {
+                Err(NetworkError::Internal(
+                    "Off-thread compilation failed.".into(),
+                ))
+            } else {
+                let script_text = DOMString::from(script);
+                let heap = Heap::default();
+                let source_code = RootedTraceableBox::new(heap);
+                source_code.set(compiled_script.get());
+                let code = SourceCode::Compiled(CompiledSourceCode {
+                    source_code: source_code,
+                    original_text: Rc::new(script_text),
+                });
+
+                Ok(ScriptOrigin {
+                    code,
+                    url: final_url,
+                    external: true,
+                    fetch_options: fetch_options,
+                    type_: ScriptType::Classic,
+                })
+            };
+
+            finish_fetching_a_classic_script(&*elem, script_kind, url, load);
+        }),
+        &context.canceller,
+    );
+}
 
 /// An unique id for script element.
 #[derive(Clone, Copy, Debug, Eq, Hash, JSTraceable, PartialEq)]
@@ -151,9 +240,23 @@ pub enum ScriptType {
 }
 
 #[derive(JSTraceable, MallocSizeOf)]
+pub struct CompiledSourceCode {
+    #[ignore_malloc_size_of = "SM handles JS values"]
+    pub source_code: RootedTraceableBox<Heap<*mut JSScript>>,
+    #[ignore_malloc_size_of = "Rc is hard"]
+    pub original_text: Rc<DOMString>,
+}
+
+#[derive(JSTraceable)]
+pub enum SourceCode {
+    Text(Rc<DOMString>),
+    Compiled(CompiledSourceCode),
+}
+
+#[derive(JSTraceable, MallocSizeOf)]
 pub struct ScriptOrigin {
     #[ignore_malloc_size_of = "Rc is hard"]
-    text: Rc<DOMString>,
+    code: SourceCode,
     url: ServoUrl,
     external: bool,
     fetch_options: ScriptFetchOptions,
@@ -168,7 +271,7 @@ impl ScriptOrigin {
         type_: ScriptType,
     ) -> ScriptOrigin {
         ScriptOrigin {
-            text: text,
+            code: SourceCode::Text(text),
             url: url,
             external: false,
             fetch_options,
@@ -183,7 +286,7 @@ impl ScriptOrigin {
         type_: ScriptType,
     ) -> ScriptOrigin {
         ScriptOrigin {
-            text: text,
+            code: SourceCode::Text(text),
             url: url,
             external: true,
             fetch_options,
@@ -192,8 +295,35 @@ impl ScriptOrigin {
     }
 
     pub fn text(&self) -> Rc<DOMString> {
-        Rc::clone(&self.text)
+        match &self.code {
+            SourceCode::Text(text) => Rc::clone(&text),
+            SourceCode::Compiled(compiled_script) => Rc::clone(&compiled_script.original_text),
+        }
     }
+}
+
+/// Final steps of <https://html.spec.whatwg.org/multipage/#fetch-a-classic-script>
+fn finish_fetching_a_classic_script(
+    elem: &HTMLScriptElement,
+    script_kind: ExternalScriptKind,
+    url: ServoUrl,
+    load: ScriptResult,
+) {
+    // Step 11, Asynchronously complete this algorithm with script,
+    // which refers to step 26.6 "When the chosen algorithm asynchronously completes",
+    // of https://html.spec.whatwg.org/multipage/#prepare-a-script
+    let document = document_from_node(&*elem);
+
+    match script_kind {
+        ExternalScriptKind::Asap => document.asap_script_loaded(&elem, load),
+        ExternalScriptKind::AsapInOrder => document.asap_in_order_script_loaded(&elem, load),
+        ExternalScriptKind::Deferred => document.deferred_script_loaded(&elem, load),
+        ExternalScriptKind::ParsingBlocking => {
+            document.pending_parsing_blocking_script_loaded(&elem, load)
+        },
+    }
+
+    document.finish_load(LoadType::Script(url));
 }
 
 pub type ScriptResult = Result<ScriptOrigin, NetworkError>;
@@ -261,43 +391,79 @@ impl FetchResponseListener for ClassicContext {
 
     /// <https://html.spec.whatwg.org/multipage/#fetch-a-classic-script>
     /// step 4-9
+    #[allow(unsafe_code)]
     fn process_response_eof(&mut self, response: Result<ResourceFetchTiming, NetworkError>) {
-        // Step 5.
-        let load = response.and(self.status.clone()).map(|_| {
-            let metadata = self.metadata.take().unwrap();
+        let (source_text, final_url) = match (response.as_ref(), self.status.as_ref()) {
+            (Err(err), _) | (_, Err(err)) => {
+                // Step 6, response is an error.
+                finish_fetching_a_classic_script(
+                    &*self.elem.root(),
+                    self.kind.clone(),
+                    self.url.clone(),
+                    Err(err.clone()),
+                );
+                return;
+            },
+            (Ok(_), Ok(_)) => {
+                let metadata = self.metadata.take().unwrap();
 
-            // Step 6.
-            let encoding = metadata
-                .charset
-                .and_then(|encoding| Encoding::for_label(encoding.as_bytes()))
-                .unwrap_or(self.character_encoding);
+                // Step 7.
+                let encoding = metadata
+                    .charset
+                    .and_then(|encoding| Encoding::for_label(encoding.as_bytes()))
+                    .unwrap_or(self.character_encoding);
 
-            // Step 7.
-            let (source_text, _, _) = encoding.decode(&self.data);
-            ScriptOrigin::external(
+                // Step 8.
+                let (source_text, _, _) = encoding.decode(&self.data);
+                (source_text, metadata.final_url)
+            },
+        };
+
+        let elem = self.elem.root();
+        let global = elem.global();
+        let cx = global.get_cx();
+        let _ar = enter_realm(&*global);
+
+        let options =
+            unsafe { CompileOptionsWrapper::new(*cx, final_url.as_str().as_ptr() as *const i8, 1) };
+
+        let can_compile_off_thread = pref!(dom.script.asynch) &&
+            unsafe { CanCompileOffThread(*cx, options.ptr as *const _, source_text.len()) };
+
+        if can_compile_off_thread {
+            let source_string = source_text.to_string();
+            let source_text: Vec<u16> = source_text.encode_utf16().collect();
+
+            let context = Box::new(OffThreadCompilationContext {
+                script_element: self.elem.clone(),
+                script_kind: self.kind.clone(),
+                final_url,
+                url: self.url.clone(),
+                task_source: global.dom_manipulation_task_source(),
+                canceller: global.task_canceller(TaskSourceName::DOMManipulation),
+                script_text: source_string,
+                utf16_chars: source_text,
+                fetch_options: self.fetch_options.clone(),
+            });
+
+            unsafe {
+                assert!(CompileOffThread(
+                    *cx,
+                    options.ptr as *const _,
+                    &mut transform_u16_to_source_text(&context.utf16_chars) as *mut _,
+                    Some(off_thread_compilation_callback),
+                    Box::into_raw(context) as *mut c_void,
+                ));
+            }
+        } else {
+            let load = ScriptOrigin::external(
                 Rc::new(DOMString::from(source_text)),
-                metadata.final_url,
+                final_url.clone(),
                 self.fetch_options.clone(),
                 ScriptType::Classic,
-            )
-        });
-
-        // Step 9.
-        // https://html.spec.whatwg.org/multipage/#prepare-a-script
-        // Step 18.6 (When the chosen algorithm asynchronously completes).
-        let elem = self.elem.root();
-        let document = document_from_node(&*elem);
-
-        match self.kind {
-            ExternalScriptKind::Asap => document.asap_script_loaded(&elem, load),
-            ExternalScriptKind::AsapInOrder => document.asap_in_order_script_loaded(&elem, load),
-            ExternalScriptKind::Deferred => document.deferred_script_loaded(&elem, load),
-            ExternalScriptKind::ParsingBlocking => {
-                document.pending_parsing_blocking_script_loaded(&elem, load)
-            },
+            );
+            finish_fetching_a_classic_script(&*elem, self.kind.clone(), self.url.clone(), Ok(load));
         }
-
-        document.finish_load(LoadType::Script(self.url.clone()));
     }
 
     fn resource_timing_mut(&mut self) -> &mut ResourceFetchTiming {
@@ -702,7 +868,16 @@ impl HTMLScriptElement {
         // unminified content.
         let (input, output) = (tempfile::NamedTempFile::new(), tempfile::tempfile());
         if let (Ok(mut input), Ok(mut output)) = (input, output) {
-            input.write_all(script.text.as_bytes()).unwrap();
+            match &script.code {
+                SourceCode::Text(text) => {
+                    input.write_all(text.as_bytes()).unwrap();
+                },
+                SourceCode::Compiled(compiled_source_code) => {
+                    input
+                        .write_all(compiled_source_code.original_text.as_bytes())
+                        .unwrap();
+                },
+            }
             match Command::new("js-beautify")
                 .arg(input.path())
                 .stdout(output.try_clone().unwrap())
@@ -712,7 +887,7 @@ impl HTMLScriptElement {
                     let mut script_content = String::new();
                     output.seek(std::io::SeekFrom::Start(0)).unwrap();
                     output.read_to_string(&mut script_content).unwrap();
-                    script.text = Rc::new(DOMString::from(script_content));
+                    script.code = SourceCode::Text(Rc::new(DOMString::from(script_content)));
                 },
                 _ => {
                     warn!("Failed to execute js-beautify. Will store unmodified script");
@@ -763,7 +938,13 @@ impl HTMLScriptElement {
         debug!("script will be stored in {:?}", path);
 
         match File::create(&path) {
-            Ok(mut file) => file.write_all(script.text.as_bytes()).unwrap(),
+            Ok(mut file) => match &script.code {
+                SourceCode::Text(text) => file.write_all(text.as_bytes()).unwrap(),
+                SourceCode::Compiled(compiled_source_code) => {
+                    file.write_all(compiled_source_code.original_text.as_bytes())
+                        .unwrap();
+                },
+            },
             Err(why) => warn!("Could not store script {:?}", why),
         }
     }
@@ -790,7 +971,7 @@ impl HTMLScriptElement {
         match read_to_string(path.clone()) {
             Ok(local_script) => {
                 debug!("Found script stored at: {:?}", path);
-                script.text = Rc::new(DOMString::from(local_script));
+                script.code = SourceCode::Text(Rc::new(DOMString::from(local_script)));
             },
             Err(why) => warn!("Could not restore script from file {:?}", why),
         }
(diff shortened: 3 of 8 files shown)


AbhishekSharma102:
Hey @gterzian I have some questions regarding this.

1. FinishOffThread should be called on the main thread or the callback? Documentation says on the runtime's main thread. If that's the case, should we store it in the callbackData and call it in evaluate_script_on_global_with_result?

2. evaluate_script_on_global_with_result should have the CompileOffThread call or run_a_classic_script? 

3. Could you please explain more on networking_task_source?

Sorry if some questions seem trivial.

components/script/dom/globalscope.rs
```diff
@@ -2218,6 +2226,16 @@ impl GlobalScope {
         self.evaluate_script_on_global_with_result(code, "", rval, 1)
     }
 
+    unsafe extern "C" fn off_thread_compilation_callback(
+        &self,
+        token: *mut c_void,
+        callbackData: *mut c_void,
+    ) {
+        let castedCallbackData = callbackData as *mut OffThreadCompilationData;
+        let script: JSScript* = FinishOffThreadScript(*(castedCallbackData.context), token);
```
gterzian, line 2235: To answer your question: here you would use the `OffThreadCompilationData` to then queue a task, and inside that task you would do the rest such as calling `FinishOffThreadScript`. 

Also, the context can be obtained from inside the task, by calling `global.get_cx()` and then you need to de-reference it when needed, for example by doing `FinishOffThreadScript(*global.get_cx(), token);`

(and yes `context` is often used in the other structs I pointed to to refer to the `Trusted<GlobalScope>` but that is just a name to say it's the "context" for the overal task that is being queued. 

So I think this should be enough info to queue the task, and then let me know if you have other questions as to how to handle the `JsScript` further, I would say you should look into calling a method on the global with it.

components/script/dom/globalscope.rs
```diff
@@ -285,6 +285,14 @@ struct FileListener {
     task_canceller: TaskCanceller,
 }
 
+#[derive(JSTraceable, MallocSizeOf)]
```
gterzian, line 288: Since this struct will be used in the IPC router only, you don't need to derive those traits which are mostly meant for stuff inside a struct marked as `dom_struct`.

components/script/dom/globalscope.rs
```diff
@@ -285,6 +285,14 @@ struct FileListener {
...
+pub struct OffThreadCompilationData {
+    context: JSContext
+    canceller: TaskCanceller,
+    task_source: TimerTaskSource,
```
gterzian, line 292: To answer your other question, so here you should instead store the `NetworkingTaskSource`, which is available via a call to `GlobalScope.networking_task_source()`. 

Then later when you need to queue a task, you can do `OffThreadCompilationData.task_source.queue_with_canceller`, like is done for example at https://github.com/servo/servo/blob/d7d56454b07b70e3d9b2e3746ea1b2a1a1f05b7f/components/script/dom/globalscope.rs#L404

gterzian: I'd say you can define a `handle(&self, token: Token)` method on the struct, and then from the callback call `handle(token)`, and inside the method queue the task and move the token into the task closure, then inside the closure you can use the token for example to call `FinishOffThreadScript`.

components/script/dom/globalscope.rs
AbhishekSharma102: Like other structs this would also not have access to the original context, and needs to be stored. Is that correct?

components/script/dom/globalscope.rs
AbhishekSharma102: Once the `FinishOffThreadScript` has been called, we need to call `evaluate_script_on_global_with_result` with the obtained script, to call `JS_ExecuteScript`. So calling `CompileOffThread` can be called directly from `run_a_classic_script`. Is that correct?

components/script/dom/globalscope.rs
gterzian: Yes, correct, you need to store the `Trusted<GlobalScope>`(and other things if you need them.

components/script/dom/globalscope.rs
gterzian: I think we need to:

1. Make `run_a_classic_script` take ownership of the `ScriptOrigin`.
2. Store the `ScriptOrigin` on `OffThreadCompilationData`, alongside the `Trusted<Global>`
3. Call the "compile off thread" API.

Then, from the task queued:

1. call `global.evaluate_script_on_global_with_result`.

`run_a_classic_script` doesn't return a result so that should be OK. 

You probably will have to change the signature of `evaluate_script_on_global_with_resul`, making the `code` argument an enum with  a `DomString` and a `JsScript` variant. 

Currently it's a `&str`, but it looks like it could in all current cases be a `DomString` instead.

gterzian: On second thoughts, I think we should do the off-thread compilation earlier, when the data is received from the network. 

So the entry point should I think be around here: https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L268

So I think you can:

Make `ScriptOrigin` have a new `Option<JsScript>` member named `compiled_script`.

1. Store both the `Trusted<HTMLScriptElement>` and the `ScriptOrigin`, and probably also `ExternalScriptKind` onto the `OffThreadCompilationData`.
2. Call the off-thread compilation API
3. From the callback, queue a task that will:
    a. call `FinishOffThreadScript`
    b. Set `compiled_script` to to the result of that call.
    c. continue at https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L272

Then `evaluate_script_on_global_with_result` could be changed to take an `Option<JsScript>` as a new `compiled_script` argument, or maybe make the `code` into an enum, and inside you can either call `Evaluate2` or `JS_ExecuteScript`, based on what is available.

AbhishekSharma102: Just a question, can these scripts be compiled in any context? As GlobalScope is not yet defined and would not be available.

gterzian: > Just a question, can these scripts be compiled in any context?

Yes I think so.

> Also how do we ensure in this that offthreadcompilation is used only when called by run_a_classic_script, as I believe this is done for all scripts?

I think the code at https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L268 is only used for classic scripts, if not we can find another way to identify the script and either do the off-thread compilation or simply continue normally.

AbhishekSharma102: Okay. So I think I will need to create a new task_canceller for this. Is it correct?

gterzian: So I'm almost certain that the code at https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L269

is in fact only called for classic scripts, because of the name `ClassicContext` and the spec links, and I also think that is the right point to setup the `OffThreadCompilationData` and then call api to do the off-thread compilation. 

You can get the global at that point from `self.elem.global()`(since `elem` is `HTMLScriptElement`), and then you can get the JS context by doing `global.cx()`. 

Then you also need to store a `Trusted<HTMLScriptElement>` on the `OffThreadCompilationData`, meaning you don't need a `Trusted<GlobalScope>` because you can get the global from the element, and then later in the callback you queue the task that will then continue the algorithm at https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L270

(where again you can get `elem` from the `Trusted<HTMLScriptElement>` you will have available by calling `root()` on it, actually the [`ClassicContext`](https://github.com/servo/servo/blob/1a6193703134512b8d09b54a85bf58b3df82a372/components/script/dom/htmlscriptelement.rs#L186) used there to handle the network responses is very similar to the `OffThreadCompilationData` you will have to define. 

By the way, let's rename it `OffThreadCompilationContext`?

let me know if you have other questions.

gterzian: > o I think I will need to create a new task_canceller for this. Is it correct?

Yes you can call `global.task_canceller()` to get one.

AbhishekSharma102: I tried getting the globalscope object using `self.elem.global()` but wasn't able to. `HTMLScriptElement` doesn't seem to have it. Can you please confirm?

gterzian: I think you first need to get the root out of the trusted, doing `let elem = self.elem.root(); `, then `elem.global()` should work

AbhishekSharma102: I had some questions regarding OffThreadCompilation methods we are using.
1. These methods are written in C, but we do an unsafe call to them using a wrapper defined in rust?
2. For JSScript, I get the error that malloc_size_of is not implemented. Upon reading the documentation, I believe it is similar to a string in rust for the time being. So implementing that would mean something to similar done for string?
3. Another error during task queuing is OffThreadToken cannot be passed between threads safely. I think a simple clone should do the job, so that function needs to be implemented for it. Is that correct?

gterzian:
Could you please add a "FIX #{issue number}" in the opening description for the PR(that way it will be closed when this will, and it also makes it easier to go back to the issue).

Looks like it's going in the right direction!

components/script/dom/bindings/trace.rs
```diff
@@ -860,6 +860,12 @@ unsafe impl JSTraceable for StyleLocked<MediaList> {
     }
 }
 
+unsafe impl JSTraceable for JSScript {
```
gterzian, line 863: I think we're going to have to store the script in a `Heap`, and then this won't be necessary, see https://github.com/servo/rust-mozjs/blob/11cabdef2cbf0884a2cc33e3c73719646bd31ce5/src/rust.rs#L465

components/script/dom/htmlscriptelement.rs
```diff
@@ -53,6 +58,32 @@ use std::sync::{Arc, Mutex};
...
+}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn off_thread_compilation_callback(
+    token: *mut OffThreadToken,
+    callback_data: *mut c_void,
+) {
+    let off_thread_compilation_context: &mut OffThreadCompilationContext =
+        &mut *(callback_data as *mut OffThreadCompilationContext);
+    let elem = off_thread_compilation_context.script_element.root();
```
gterzian, line 76: this kind of stuff will have to be done from inside the queued task.

components/script/dom/globalscope.rs
gterzian: 1. Yes, I think that's a correct description.
2. You can ignore it, using an annoation like in https://github.com/servo/servo/blob/948dd7379618e85269227a3e37c76a091c85a966/components/script/dom/globalscope.rs#L250
3. I'm not sure about this one. I would say define a `struct OffthreadToken;`, and then in the callback cast the *mut OffThreadToken to it? Then cast it back when you call `FinishOffThreadScript`? @jdm do you have an idea how to represent `OffThreadToken* token` in Rust? It appears to be just an empty class defined at https://github.com/servo/mozjs/blob/aabcc9ba889b2755f1e4e83f28323a60415a790f/mozjs/js/src/vm/HelperThreads.h#L37

components/script/dom/htmlscriptelement.rs
```diff
@@ -53,6 +58,32 @@ use std::sync::{Arc, Mutex};
...
+    let global = elem.global();
+    let _ = off_thread_compilation_context.task_source.queue_with_canceller(
+        task! (off_thread_compile_continue: move || {
+            let global_context = global;
+            off_thread_compilation_context.script_origin.code = SourceCode::Compiled(*FinishOffThreadScript(*global_context.get_cx(), token));
```
gterzian, line 81: So I think we're going to have to do two things:

1. Root the `JsScript` as soon as it is received here, using the `rooted!` macro(you'll see various example across the codebase, just search you'll find many). 
2. store it inside a `Heap` inside the `ScriptOrigin`. For that I think you'll have to do something like:

```rust
let rooted_script = rooted! //
let heap = Heap::default();
heap.set(rooted_script.get());
let source_code = SourceCode::Compiled(heap);
```

AbhishekSharma102: What is the significance of rooted?

gterzian: https://github.com/servo/servo/blob/6b0d9afd6fdc28356ad44af0104ddd25a7b6438d/components/script/dom/bindings/root.rs#L5

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/GC_Rooting_Guide

gterzian:
Ok so I ended-up looking at this locally, and I think it's actually pretty hard to find the right structure, I ended-up finding something over at https://github.com/gterzian/servo/tree/unblock_off_thread_compilation_pr, which I think could be useful to unblock this PR? 

It took me quite a lot of compilation cycles locally to figure out, so I think it would have been impractical to do with via even more, and probably confusing, review cycles. 

@AbhishekSharma102  How about you rebase this off of that branch? There is still some work to finish, for example in `evaluate_script_on_global_with_result`, and I also suspect there will be some need for further debugging and so on. 

I think there is also an issue that is similar to this and you could use what you've learned to do that one, it is at https://github.com/servo/servo/issues/25209

AbhishekSharma102:
Hey @gterzian ! Thanks for the help. I will take a look. Sorry this is taking so long, as pre occupied with some other work.

AbhishekSharma102:
I keep facing issues in this compilation step.
`rustc --crate-name servo --edition=2018 ports/winit/main.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type bin --emit=dep-info,link -Cembed-bitcode=no -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="egl"' --cfg 'feature="layout-2013"' --cfg 'feature="max_log_level"' --cfg 'feature="media-gstreamer"' --cfg 'feature="native-bluetooth"' --cfg 'feature="webdriver"' -C metadata=f92413ac43bf4d65 -C extra-filename=-f92413ac43bf4d65 --out-dir /servo_codes/servo/target/debug/deps -C incremental=/servo_codes/servo/target/debug/incremental -L dependency=/servo_codes/servo/target/debug/deps --extern backtrace=/servo_codes/servo/target/debug/deps/libbacktrace-c2de253738f90dda.rlib --extern clipboard=/servo_codes/servo/target/debug/deps/libclipboard-8cdd848607beeec5.rlib --extern euclid=/servo_codes/servo/target/debug/deps/libeuclid-860d44d6f8bac949.rlib --extern getopts=/servo_codes/servo/target/debug/deps/libgetopts-f7c37fe6e97f0526.rlib --extern image=/servo_codes/servo/target/debug/deps/libimage-f5fa0b5fb9d8ed8d.rlib --extern keyboard_types=/servo_codes/servo/target/debug/deps/libkeyboard_types-844dac95fd59d379.rlib --extern lazy_static=/servo_codes/servo/target/debug/deps/liblazy_static-43f70af18d2bceaa.rlib --extern libc=/servo_codes/servo/target/debug/deps/liblibc-61b63b0480794d3d.rlib --extern servo=/servo_codes/servo/target/debug/deps/libservo-ce935d12301a3a40.rlib --extern log=/servo_codes/servo/target/debug/deps/liblog-1483e2d7b9f1cf22.rlib --extern servo_media=/servo_codes/servo/target/debug/deps/libservo_media-bdb2208652d515dc.rlib --extern shellwords=/servo_codes/servo/target/debug/deps/libshellwords-2c1f93ce98cb1ed4.rlib --extern sig=/servo_codes/servo/target/debug/deps/libsig-d7ed8854566421c7.rlib --extern surfman=/servo_codes/servo/target/debug/deps/libsurfman-f4f4a8eb4b1584ae.rlib --extern tinyfiledialogs=/servo_codes/servo/target/debug/deps/libtinyfiledialogs-909cab068201ce77.rlib --extern webxr=/servo_codes/servo/target/debug/deps/libwebxr-c78e8dd8736c4e40.rlib --extern winit=/servo_codes/servo/target/debug/deps/libwinit-3bb6a4b7602c9773.rlib -C link-args=-fuse-ld=gold -W unused-extern-crates -L native=/usr/lib/x86_64-linux-gnu -L native=/usr/lib/x86_64-linux-gnu -L native=/usr/lib/x86_64-linux-gnu -L native=/usr/lib/x86_64-linux-gnu -L native=/usr/lib/x86_64-linux-gnu -L native=/servo_codes/servo/target/debug/build/fontsan-a838246ab2050249/out/lib -L native=/servo_codes/servo/target/debug/build/miniz-sys-f174e987e15b97f4/out -L native=/usr/lib/x86_64-linux-gnu -L native=/servo_codes/servo/target/debug/build/jemalloc-sys-3850958afd320de4/out/build/lib -L native=/usr/lib/x86_64-linux-gnu -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/target/debug/build/mozangle-10b974c9166bcb40/out -L native=/servo_codes/servo/target/debug/build/mozjs-8aa97c24a8623cb4/out -L native=/servo_codes/servo/target/debug/build/mozjs_sys-7554fc71b8fcca7b/out/build/js/src/build -L native=/servo_codes/servo/target/debug/build/mozjs_sys-7554fc71b8fcca7b/out/build/glue -L native=/usr/lib/x86_64-linux-gnu -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/support/linux/gstreamer/gst/lib -L native=/servo_codes/servo/target/debug/build/tinyfiledialogs-4ef66bfd68031559/out`

with this note:
 note: /home/abhishek/.cargo/registry/src/github.com-1ecc6299db9ec823/gstreamer-webrtc-0.15.5/src/auto/enums.rs:274: error: undefined reference to 'gst_webrtc_data_channel_state_get_type'

Any suggestions on this?

gterzian:
It might be https://github.com/servo/servo/issues/27090, or another issue related to gstreamer, or it might warrant opening a new issue.

jdm:
You will need to run `./mach bootstrap-gstreamer` and uninstall the `gir1.2-gst-plugins-bad-1.0` package.

gterzian:
@AbhishekSharma102 Ok thanks for adding the "execute script part", which looks good(although I suspect we'll need to debug it a bit in practice). 

On that last commit, you could add me a co-author, using 2792687+gterzian@users.noreply.github.com as email and following the instructions at https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

AbhishekSharma102:
@gterzian I will do that. I am planning to squash commits into one, and will add into that. 

Regarding the squashing, unfortunately I merged from master after the first two commits, so there are several comments between the first and last two. Could you suggest on how to merge them?

gterzian:
> Regarding the squashing, unfortunately I merged from master after the first two commits, so there are several comments between the first and last two. Could you suggest on how to merge them?

I'm not sure, maybe just try squashing them all into one? 

Usually we don't merge, rather we rebase(and only if there is an actual conflict or it's otherwise necessary), see https://github.com/servo/servo/wiki/Github-workflow

gterzian:
I tried rebasing my branch off of servo/master and there were no conflicts, what problems are you encountering?

AbhishekSharma102:
Managed to resolve them. Using git too much for the first time, thus the issues. Merged all commits to one. 

Please let me know if any other changes/tests are needed.

gterzian:
Ok so this clearly broke something, actually it seems to have broken everything, so that's good because it means we just forgot something or made some other obviously mistake, better than having to deal with a few edge cases(might be the next step). 

I'll build locally also and try to help debug.

AbhishekSharma102:
> Ok so this clearly broke something, actually it seems to have broken everything, so that's good because it means we just forgot something or made some other obvious mistake, which is better than having to deal with a few edge cases(might be the next step).
> 
> I'll build locally also and try to help debug.

Any way to obtain more logs?

gterzian:
yes via the link above you can look at things like https://community-tc.services.mozilla.com/tasks/b6kS1MK3RbGRj7A4rt1BpA#artifacts

Usually you'd look at the `filtered-intermittents` one, now everything seems to have failed. There are no helpful stack-traces for some reason. I'm hoping to get something better locally .

gterzian:
```
pid:840 Stack trace for thread "ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }"
 0:05.02 pid:840    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
 0:05.02 pid:840    1: core::fmt::write
 0:05.02 pid:840    2: std::io::Write::write_fmt
 0:05.02 pid:840    3: servo::backtrace::print
 0:05.02 pid:840    4: servo::install_crash_handler::handler
 0:05.03 pid:840    5: __sigtramp
 0:05.03 pid:840    6: __ZN2JS14CompileOptionsC2EP9JSContext
 0:05.03 pid:840    7: _NewCompileOptions
 0:05.03 pid:840    8: mozjs::rust::CompileOptionsWrapper::new
 0:05.03 pid:840    9: <script::dom::htmlscriptelement::ClassicContext as net_traits::FetchResponseListener>::process_response_eof
 0:05.03 pid:840   10: <net_traits::FetchResponseMsg as net_traits::Action<T>>::process
 0:05.03 pid:840   11: <script::network_listener::ListenerTask<A,Listener> as script::task::TaskOnce>::run_once
 0:05.03 pid:840   12: <T as script::task::TaskBox>::run_box
 0:05.03 pid:840   13: script::script_thread::ScriptThread::handle_msgs
 0:05.03 pid:840   14: profile_traits::mem::ProfilerChan::run_with_memory_reporting
 0:05.03 pid:840   15: std::sys_common::backtrace::__rust_begin_short_backtrace
 0:05.03 pid:840   16: core::ops::function::FnOnce::call_once{{vtable.shim}}
 0:05.03 pid:840   17: std::sys::unix::thread::Thread::new::thread_start
 0:05.03 pid:840   18: __pthread_body
 0:05.04 pid:840   19: __pthread_start
```

Indeed getting a more useful stracktrace locally. 

The failure at `mozjs::rust::CompileOptionsWrapper::new` looks like the classic "we need to enter a realm before calling this"...

I'll give it a try locally.

gterzian:
Ok with this change: https://github.com/servo/servo/commit/7b0148f0c409d05a99532a074e0c8298901dfa39

I'm now getting:

```
pid:1563 Stack trace
 0:05.80 pid:1563    0: <servo::backtrace::Print as core::fmt::Debug>::fmt
 0:05.80 pid:1563    1: core::fmt::write
 0:05.80 pid:1563    2: std::io::Write::write_fmt
 0:05.80 pid:1563    3: servo::backtrace::print
 0:05.80 pid:1563    4: servo::install_crash_handler::handler
 0:05.80 pid:1563    5: __sigtramp
 0:05.80 pid:1563    6: script::dom::htmlscriptelement::off_thread_compilation_callback
 0:05.80 pid:1563    7: __ZN2js12HelperThread19handleParseWorkloadERNS_25AutoLockHelperThreadStateE
 0:05.81 pid:1563    8: __ZN2js12HelperThread10threadLoopEv
 0:05.81 pid:1563    9: __ZN2js6detail16ThreadTrampolineIRFvPvEJPNS_12HelperThreadEEE5StartES2_
 0:05.81 pid:1563   10: __pthread_body
 0:05.81 pid:1563   11: __pthread_start
```

jdm:
https://github.com/servo/servo/blob/7b0148f0c409d05a99532a074e0c8298901dfa39/components/script/dom/htmlscriptelement.rs#L399-L406 is a stack-owned value and you're passing a pointer to it as the argument to off_thread_compilation_callback. It needs to be a Box::into_raw instead, so that the memory lives long enough to be accessed within the callback.

AbhishekSharma102:
I tried to run it locally, but it gave crash message. Do we need a display to debug? I am running the code in a cloud VM.

gterzian:
It seems to work with those changes: https://github.com/servo/servo/commit/7633d85837df02532ade91dafae67c42705550b6 !

@jdm thanks for the pointers(no pun intendend)

@AbhishekSharma102 If you add that commit and squash it into the existing one, we can do another full run and see what happens.

gterzian:
> Do we need a display to run wpt? I am running the code in a cloud VM.

I'm not sure.

gterzian:
Actually I think I need to make one more change to avoid leaking the token, one moment please.

AbhishekSharma102:
> > Do we need a display to run wpt? I am running the code in a cloud VM.
> 
> I'm not sure.

I'll see if those pass locally (at least some).

gterzian:
Ok!

There you go: https://github.com/servo/servo/commit/286a656e4079705584e16d8d2ccae907c8f7a1e0

This one seems to work, and I don't think it leaks.

AbhishekSharma102:
> Ok!
> 
> There you go: [286a656](https://github.com/servo/servo/commit/286a656e4079705584e16d8d2ccae907c8f7a1e0)
> 
> This one seems to work, and I don't think it leaks.

Thanks a lot! Included those changes and pushed.

gterzian:
These appear relevant:

```
0 matches
2 unexpected results that are NOT known-intermittents:
  ▶ TIMEOUT [expected OK] /html/semantics/scripting-1/the-script-element/script-onerror-insertion-point-1.html
  │ 
  ▶ Unexpected subtest result in /html/semantics/scripting-1/the-script-element/script-onerror-insertion-point-1.html:
  └ NOTRUN [expected PASS] Test that the insertion point is defined in the error event of a parser-inserted script that actually started a fetch (but just had it fail).
  ▶ Unexpected subtest result in /html/semantics/scripting-1/the-script-element/load-error-events-1.html:
  │ FAIL [expected PASS] src, 404, parser-inserted, defer, no async
  │   → assert_unreached: Load event should not be fired. Reached unreachable code
  │ FAIL [expected PASS] src, 404, parser-inserted, no defer, no async
  │   → assert_unreached: Load event should not be fired. Reached unreachable code
  │ FAIL [expected PASS] src, 404, parser-inserted, no defer, async
  │   → assert_unreached: Load event should not be fired. Reached unreachable code
  │ 
  │ onLoad/<@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/resources/load-error-events-helpers.js:24:29
  │ onLoad@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/resources/load-error-events-helpers.js:16:12
  └ onload@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/load-error-events-1.html:2:7
  ▶ Unexpected subtest result in /html/semantics/scripting-1/the-script-element/load-error-events-1.html:
  │ FAIL [expected PASS] src, 404, not parser-inserted, no defer, no async, no non-blocking
  │   → assert_unreached: Load event should not be fired. Reached unreachable code
  │ 
  │ onLoad/<@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/resources/load-error-events-helpers.js:24:29
  │ onLoad@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/resources/load-error-events-helpers.js:16:12
  └ script3_dynamic_error.onload@http://web-platform.test:8000/html/semantics/scripting-1/the-script-element/load-error-events-1.html:37:44
  ▶ Unexpected subtest result in /html/semantics/scripting-1/the-script-element/load-error-events-1.html:
  │ FAIL [expected PASS] src, 404, not parser-inserted, no defer, async
```

```
0 matches
  ▶ TIMEOUT [expected OK] /html/webappapis/scripting/events/event-handler-processing-algorithm-error/script-element.html
  │ 
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/event-handler-processing-algorithm-error/script-element.html:
  │ TIMEOUT [expected PASS] error event behaves normally (return true does not cancel; one arg) on a script element, with a 404 error
  └   → Test timed out
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/event-handler-processing-algorithm-error/script-element.html:
  └ NOTRUN [expected PASS] error event behaves normally (return true does not cancel; one arg) on a script element, with a synthetic Event
  ▶ Unexpected subtest result in /html/webappapis/scripting/events/event-handler-processing-algorithm-error/script-element.html:
  └ NOTRUN [expected PASS] error event behaves normally (return true does not cancel; one arg) on a script element, with a synthetic ErrorEvent

```

Now we've reached the "debug edge conditions" phase...

gterzian:
Ok so the problem was not taking `self.status` into account when determining whether the network as an error or not, fixed at https://github.com/servo/servo/commit/8cc91623557cd67e8f45cd09b623e76d2a3f2cc3

AbhishekSharma102:
> Ok so the problem was not taking `self.status` into account when determining whether the network errored, fixed at [8cc9162](https://github.com/servo/servo/commit/8cc91623557cd67e8f45cd09b623e76d2a3f2cc3)

So earlier, in case of an error, we weren't returning, and sending incomplete script to compile off thread?

AbhishekSharma102:
> Ok so the problem was not taking `self.status` into account when determining whether the network errored, fixed at [8cc9162](https://github.com/servo/servo/commit/8cc91623557cd67e8f45cd09b623e76d2a3f2cc3)

Hey, I did the changes, and pushed. I didn't know specific permissions were there for wpt.

components/script/dom/htmlscriptelement.rs
```diff
@@ -730,7 +893,11 @@ impl HTMLScriptElement {
         debug!("script will be stored in {:?}", path);
 
         match File::create(&path) {
-            Ok(mut file) => file.write_all(script.text.as_bytes()).unwrap(),
+            Ok(mut file) => {
+                if let SourceCode::Text(text) = &script.code {
+                    file.write_all(text.as_bytes()).unwrap()
+                }
```
gterzian, line 899: Let's add a `warn!("Cannot store compiled script")` for the `else` case.

gterzian:
Getting a PASS for `/custom-elements/customized-built-in-constructor-exceptions.html`, not sure if that is an intermittent. Let's see on the next run.

gterzian:
@jdm r? (I'm co-author so don't think I should review)

gterzian:
> So earlier, in case of an error, we weren't returning, and sending incomplete script to compile off thread?

Yes

components/script/dom/htmlscriptelement.rs
```diff
@@ -669,7 +830,9 @@ impl HTMLScriptElement {
         // unminified content.
         let (input, output) = (tempfile::NamedTempFile::new(), tempfile::tempfile());
         if let (Ok(mut input), Ok(mut output)) = (input, output) {
-            input.write_all(script.text.as_bytes()).unwrap();
+            if let SourceCode::Text(text) = &script.code {
+                input.write_all(text.as_bytes()).unwrap();
+            }
```
gterzian, line 835: Here I think we should also warn in the else case(or find a way to get the text from a `JSScript`)

AbhishekSharma102: maybe we can use `JS_DecompileScript `method with uses `context` and `Handle<*mut JSScript>` and returns `*mut JSString`? I am unsure on how to get the context though.

gterzian: Good idea, give it a try I would say. You can get the global from(the equivalent of) `*self.globa().get_cx()`

AbhishekSharma102: I noticed that the wrapper for  [`DecompileScript`](https://docs.rs/mozjs/0.7.2/mozjs/rust/wrappers/fn.JS_DecompileScript.html) takes 4 args, while the [one](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS_DecompileScript) in spidermonkey project, takes 2 and [JS_DecompileScriptObject](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS_DecompileScriptObject) takes 4. Due to lack of documentation on the last one, not sure if we can call it.

gterzian: I think you can try to use this one: https://github.com/servo/mozjs/blob/aabcc9ba889b2755f1e4e83f28323a60415a790f/mozjs/js/src/jsapi.h#L1885

gterzian: We could also just store a clone of the original text alongside the compiled script, and use that when needed.

AbhishekSharma102: Yeah, that's always an option. I think it would be better, as decompile script doesnt always return the original script.

gterzian: yes.

AbhishekSharma102: So I tried to store it as DOMString, but Send is not implemented for it and cannot queue the task back. Should I implement that similar to how you did for OffThreadToken?

gterzian: I think you should store it as a String then, and later convert it back to a dom string inside the task.

components/script/dom/htmlscriptelement.rs
```diff
@@ -168,16 +260,43 @@ impl ScriptOrigin {
...
             external: true,
             type_,
         }
     }
 
     pub fn text(&self) -> DOMString {
-        self.text.clone()
+        match &self.code {
+            SourceCode::Text(text) => text.clone(),
+            SourceCode::Compiled(_) => panic!("Cannot convert a compiled script to text."),
```
AbhishekSharma102, line 273: @gterzian Shall we replace this with the stored script as well? Or is that not needed?

gterzian: Yes we can do it here as well.

AbhishekSharma102: Updated

components/script/dom/htmlscriptelement.rs
```diff
@@ -148,9 +231,22 @@ pub enum ScriptType {
     Module,
 }
 
+#[derive(JSTraceable, MallocSizeOf)]
+pub struct CompiledSourceCode {
+    #[ignore_malloc_size_of = "SM handles JS values"]
+    pub source_code: Heap<*mut JSScript>,
+    pub script: DOMString,
```
gterzian, line 238: I think it would be clearer if we call this `original_text`.

gterzian (approved):
Looks great, just one nit re the naming

gterzian:
@jdm r?

components/script/dom/htmlscriptelement.rs
AbhishekSharma102: I'll do that change

AbhishekSharma102: I just pushed the changes to this branch, but not showing here. That's weird

components/script/dom/htmlscriptelement.rs
```diff
@@ -53,6 +63,79 @@ use std::sync::{Arc, Mutex};
...
+    url: ServoUrl,
+    task_source: NetworkingTaskSource,
+    canceller: TaskCanceller,
+    script_text: String,
+}
+
+/// A wrapper to mark OffThreadToken as Send,
+/// which should be safe according to
+/// mozjs/js/public/OffThreadScriptCompilation.h
+struct OffThreadCompilationToken(Box<OffThreadToken>);
```
jdm, line 79: Why do we need a Box?

components/script/dom/htmlscriptelement.rs
```diff
@@ -53,6 +63,79 @@ use std::sync::{Arc, Mutex};
...
+    callback_data: *mut c_void,
+) {
+    let context = Box::from_raw(callback_data as *mut OffThreadCompilationContext);
+    let token = OffThreadCompilationToken(Box::from_raw(token));
+
+    let url = context.url.clone();
+    let final_url = context.final_url.clone();
+    let script_element = context.script_element.clone();
+    let script_kind = context.script_kind.clone();
+    let script = context.script_text.clone();
```
jdm, line 96: I'm concerned about cloning the string text here, because these could be very large minified scripts. What if we used `mem::replace(context.script_text, String::new())` instead?

components/script/dom/htmlscriptelement.rs
```diff
@@ -148,9 +231,22 @@ pub enum ScriptType {
     Module,
 }
 
+#[derive(JSTraceable, MallocSizeOf)]
+pub struct CompiledSourceCode {
```
jdm, line 235: We either need to add a `#[must_root]` to this struct (and store it within a `RootedTraceableBox`) since it contains a GC value, or we need to store `pub source_code: RootedTraceableBox<Heap<*mut JSScript>`.

components/script/dom/htmlscriptelement.rs
```diff
@@ -243,42 +366,85 @@ impl FetchResponseListener for ClassicContext {
...
+            (_, Err(err)) => {
+                // Step 6, response is an error.
+                finish_fetching_a_classic_script(
+                    &*self.elem.root(),
+                    self.kind.clone(),
+                    self.url.clone(),
+                    Err(err.clone()),
+                );
+                return;
+            },
```
jdm, line 391: Could we combine these as `(Err(err), _) | (_, Err(err))` and avoid the duplication?

components/script/dom/htmlscriptelement.rs
```diff
@@ -243,42 +366,85 @@ impl FetchResponseListener for ClassicContext {
...
+            unsafe { CanCompileOffThread(*cx, options.ptr as *const _, source_text.len()) };
+
+        if can_compile_off_thread {
+            let context = Box::into_raw(Box::new(OffThreadCompilationContext {
+                script_element: self.elem.clone(),
+                script_kind: self.kind.clone(),
+                final_url,
+                url: self.url.clone(),
+                task_source: global.networking_task_source(),
+                canceller: global.task_canceller(TaskSourceName::Networking),
```
jdm, line 425: Let's use the dom_task_source instead.

components/script/dom/htmlscriptelement.rs
```diff
@@ -243,42 +366,85 @@ impl FetchResponseListener for ClassicContext {
...
+                script_text: source_text.to_string(),
+            }));
 
-            // Step 7.
-            let (source_text, _, _) = encoding.decode(&self.data);
-            ScriptOrigin::external(
+            let source_text: Vec<u16> = source_text.encode_utf16().collect();
+
+            unsafe {
+                assert!(CompileOffThread(
```
jdm, line 432: If we use CompileOffThread1 we can pass a UTF8 string instead. We could even store the source text inside of an Arc and avoid having to clone the underlying source at all.

However, this could also be a followup issue.

jdm (changes requested):
I found this straightforward to read. Great work, both of you!

components/script/dom/htmlscriptelement.rs
gterzian: I got a crash at the point of using the token from inside the task that looked very similar to the crash where you pointed out the need to box the pointer, so I think this is a similar situation?

components/script/dom/htmlscriptelement.rs
gterzian: There was an issue because the `_` is not the same type in those two arms(there might be another way to make it work, but that is why I added two different match arms).

components/script/dom/htmlscriptelement.rs
gterzian: I think we are only passing this around, so then I guess using the `RootedTraceableBox` is the right option? I thougth a `Heap` would also root the value, however I guess that is not correct, and most use of `Heap` is within a dom struct which roots it...

components/script/dom/htmlscriptelement.rs
jdm: It's not clear to me how a Box actually helps anything here, since the token is just a pointer. I'd like to see what happens if we remove the Box and just have an OffThreadCompilationToken on the stack that gets used in the task in off_thread_compilation_callback.

Actually, it's not clear to me how the current code even works - we're calling Box::from_raw on the token argument, but we never created it from Box::into_raw so that's invalid.

components/script/dom/htmlscriptelement.rs
jdm: Storing the RootedTraceableBox as a member of CompiledSourceCode should be the easiest solution.

components/script/dom/htmlscriptelement.rs
jdm: Are you sure it's the _ that matters? What if you matched on `response.as_ref()`?

gterzian: I'm not sure, I'll leave it to @AbhishekSharma102 to look into.

components/script/dom/htmlscriptelement.rs
gterzian: Ok, it's worth a try to see if the boxing is unnecessary @AbhishekSharma102. By the way I think it is valid to use `from_raw` on a pointer that wasn't created via `into_raw`, see the second example at https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw

jdm: You are correct that it can be valid if it was created by a Rust allocator. The OffThreadToken is created by the spidermonkey allocator, though.

gterzian: Ok, thanks for the info.

gterzian: The docs do explain this in more detail over at https://doc.rust-lang.org/std/boxed/index.html#memory-layout (which is linked to from the `from_raw` docs)

components/script/dom/htmlscriptelement.rs
AbhishekSharma102: I think we can use this straightaway, without changing too much code. I'll try to see what can be done.

components/script/dom/htmlscriptelement.rs
AbhishekSharma102: Using `response.as_ref()` doesn't throw any errors. Have merged the two

components/script/dom/htmlscriptelement.rs
AbhishekSharma102: I am having trouble running wpt locally to test this out. Unit tests are passing without Box as well (i guess that happened last time as well, wpts failed while ut passed)

gterzian: > I am having trouble running wpt locally to test this out.

What problem are you encountering? 

If you don't know exactly which test to run, I think any test will do, since all will load scripts I believe. You could try for example `./mach test-wpt tests/wpt/web-platform-tests/html/`(maybe with the `--release` flag depending on what you have build locally). 

If you don't get a crash, I think it's worth pushing your changes and we can try here as well.

AbhishekSharma102: I keep getting timeouts, probably due to low sku.

gterzian: yes I think that's normal, but most are ok right? Or does everything time-out? 

Sounds like it's working, and maybe you can address the other comments if not already, and then push the changes and we can take it from there?

AbhishekSharma102: Apart from CompileOffThread1, I've addressed the rest.

 Will push after rebasing.

AbhishekSharma102: when i do ./mach build --dev, it tries to update a few things and demands for almost 52GB of memory, and fails.
memory allocation of 52834435104 bytes failed.

Isnt that a bit too much?

components/script/dom/htmlscriptelement.rs
```diff
@@ -48,13 +59,89 @@ use servo_url::ServoUrl;
...
+                let code = SourceCode::Compiled(CompiledSourceCode {
+                    source_code: source_code,
+                    original_text: script_text
+                });
+
+                Ok(ScriptOrigin {
+                    code,
+                    url: final_url,
+                    external: true,
+                    fetch_options: ScriptFetchOptions::default_classic_script(&global),
```
jdm, line 134: Let's store the ScriptFetchOptions in OffThreadCompilationContext and use it here instead.

components/script/dom/htmlscriptelement.rs
```diff
@@ -261,43 +388,76 @@ impl FetchResponseListener for ClassicContext {
...
+                    context as *mut _ as *mut c_void,
+                ));
+            }
+        } else {
+            let load = ScriptOrigin::external(
                 Rc::new(DOMString::from(source_text)),
-                metadata.final_url,
-                self.fetch_options.clone(),
+                final_url.clone(),
+                ScriptFetchOptions::default_classic_script(&global),
```
jdm, line 456: We should be using `self.fetch_options.clone()` here still.

jdm (changes requested):
This is very close!

components/script/dom/htmlscriptelement.rs
AbhishekSharma102: should we also store self.fetch_options in OffThreadCompilationContext?

components/script/dom/htmlscriptelement.rs
```diff
@@ -48,13 +59,89 @@ use servo_url::ServoUrl;
...
+
+            let load = if compiled_script.get().is_null() {
+                Err(NetworkError::Internal(
+                    "Off-thread compilation failed.".into(),
+                ))
+            } else {
+                let script_text = DOMString::from(script);
+                let heap = Heap::default();
+                heap.set(compiled_script.get());
+                let source_code = RootedTraceableBox::new(heap);
```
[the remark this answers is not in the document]
AbhishekSharma102: Can you please explain a bit on this or share some doc?

jdm:
```
  ▶ TIMEOUT [expected OK] /referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html
  │ 
  └ [2020-07-20T19:57:55Z ERROR script::dom::bindings::error] Error at http://web-platform.test:8000/referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html:7:17 requestViaXhr is not defined
  │ TIMEOUT [expected FAIL] Sandboxed iframe with opaque origin doesn't send referrers.
  └   → Test timed out
  ▶ OK [expected CRASH] /custom-elements/customized-built-in-constructor-exceptions.html

  ▶ ERROR [expected OK] /_webgl/conformance/more/functions/bindFramebufferLeaveNonZero.html
  │   → Tests is not defined
  │ 
  └ http://web-platform.test:8000/_webgl/conformance/more/functions/bindFramebufferLeaveNonZero.html:40:1
```
Very interesting. I'd like to take a closer look at those failures to understand what's going on.

gterzian:
> requestViaXhr is not defined

That test is loading a script via the srcdoc of an iframe. https://github.com/servo/servo/blob/1ded1ba9926f962bd84e9c949677009da06ea653/tests/wpt/web-platform-tests/referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html#L28

I'm guessing the fact that we're not blocking on compiling/executing the script when it loads exposes some sort of timing issue with running the code of an srcdoc?

gterzian:
So when an iframe has an srcdoc, we parse and execute it over at https://github.com/servo/servo/blob/1ded1ba9926f962bd84e9c949677009da06ea653/components/script/script_thread.rs#L3904

I think it's  a problem with not blocking the load event properly on that src doc while we're still compiling the script? This must have been hidden previously due to the sync execution?

gterzian:
Or maybe we should apply the off-thread compilation flow also to classic scripts with no src(as used in that srcdoc), see https://github.com/servo/servo/blob/1ded1ba9926f962bd84e9c949677009da06ea653/components/script/dom/htmlscriptelement.rs#L640

jdm:
The way I understand it is that since we only try to compile external scripts off the main thread, the parser should already be blocked and therefore we shouldn't parse the internal script before we're finished executing the external scripts. If that's not the case, we need to figure out why the parser isn't blocked.

jdm:
The parser is indeed blocked as expected. It turns out that off-thread compilation is failing for one instance of common.sub.js in the srcdoc iframe, while the other succeeds. RUST_LOG=script::dom::htmlscriptelement demonstrates this happening.

jdm:
Aha - we're violating the rule that [script source characters need to live until the callback is called](https://searchfox.org/mozilla-central/rev/3b6958c26049c1e27b2790a43154caaba9f6dd4a/js/public/OffThreadScriptCompilation.h#55-57), so we're racing with the off-thread parser as we destroy the utf-16 characters that we pass when we call CompileOffThread.

AbhishekSharma102:
> Aha - we're violating the rule that [script source characters need to live until the callback is called](https://searchfox.org/mozilla-central/rev/3b6958c26049c1e27b2790a43154caaba9f6dd4a/js/public/OffThreadScriptCompilation.h#55-57), so we're racing with the off-thread parser as we destroy the utf-16 characters that we pass when we call CompileOffThread.

I seem to have completely missed this. Sorry for this.

jdm:
@bors-servo retry
* #25513 #27342

gterzian:
@jdm thanks for catching that. You added the tracer right? So you mean previously it could be GC'ed during the off-thread compilation?

jdm:
Not GCed; we called `source_text.encode_utf16().collect()` before passing the resulting vector's pointer to CompileOffThread, but then we would drop the vector when we left that stack frame. I moved the vector into the OffThreadCompileContext to ensure that it remained alive until we dropped that context inside the off thread compilation callback function.

jdm:
I don't think I added the JSScript tracer to this PR. I guess we just haven't ever tried to store a JSScript in a Heap before?

gterzian:
> I moved the vector into the OffThreadCompileContext to ensure that it remained alive until we dropped that context inside the off thread compilation callback function.

Makes sense, thanks for the info. 

>  I guess we just haven't ever tried to store a JSScript in a Heap before?

I think so too.

There is this: https://github.com/servo/rust-mozjs/blob/0caf5549cddbb34ad16abf35fb6bfbff824a4d14/src/rust.rs#L466

However that's `Trace`, not `JsTraceable`...

AbhishekSharma102:
> I don't think I added the JSScript tracer to this PR. I guess we just haven't ever tried to store a JSScript in a Heap before?

I added the tracer as RootedTraceableBox needed it i think. So did something similar to that was done for JSString and JSObject. Hope that was correct.

Also, can you suggest a easy/medium issue that I could work on next? Should i pick up #25209 ?

gterzian:
https://github.com/servo/servo/issues/25209 is a good one I think, yes.

PR #26711, opened 2020-05-29
Support replacing external scripts with local copies from disk

### LINKED ISSUE #26456: Support loading unminified JS files from disk instead of network
Running with `--unminify-js` gives us a directory tree on disk that effectively mirrors the network, but with more readable source files. If we stored the HTTP response headers alongside the actual unminified files, we could theoretically avoid the network and read directly from disk when the files exist. This would then allow us to add debugging code to the scripts to help track down errors when we don't have access to the original source.

To accomplish this, we will need to:
* add a new command line argument that enables checking a given directory before going to the network
* store HTTP response headers as part of --unminify-js
* make the --unminify-js mode create directory trees that mirror the entire path of the URL
* only use the disk mode for GET requests
* rewrite any Content-Length and Content-Encoding headers for disk-based responses appropriately

skrzyp1:
---

- [x] These changes fix #26456(GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -119,7 +119,7 @@ use script_traits::{
 use script_traits::{TimerSchedulerMsg, WebrenderIpcSender, WindowSizeData, WindowSizeType};
 use selectors::attr::CaseSensitivity;
 use servo_geometry::{f32_rect_to_au_rect, MaxRect};
-use servo_url::{Host, ImmutableOrigin, MutableOrigin, ServoUrl};
+use servo_url::{ImmutableOrigin, MutableOrigin, ServoUrl};
 use std::borrow::Cow;
 use std::borrow::ToOwned;
 use std::cell::Cell;
@@ -127,7 +127,6 @@ use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::default::Default;
 use std::env;
-use std::fs;
 use std::io::{stderr, stdout, Write};
 use std::mem;
 use std::rc::Rc;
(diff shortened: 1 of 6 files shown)


components/script/dom/window.rs
```diff
@@ -1981,8 +1984,8 @@ impl Window {
         if let Some(&Host::Domain(ref host)) = document.url().origin().host() {
             let mut path = env::current_dir().unwrap();
             path.push("unminified-js");
-            path.push(host);
-            let _ = fs::remove_dir_all(&path);
+            //path.push(host);
+            //let _ = fs::remove_dir_all(&path);
             match fs::create_dir_all(&path) {
                 Ok(_) => {
                     *self.unminified_js_dir.borrow_mut() =
```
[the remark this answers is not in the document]
jdm: I agree. Let's remove this code.

jdm (changes requested):
This looks really good so far! Please squash all of the commits into one, then rebase that commit against master to remove any merge conflicts.

components/script/dom/window.rs
```diff
@@ -1981,8 +1984,8 @@ impl Window {
         if let Some(&Host::Domain(ref host)) = document.url().origin().host() {
             let mut path = env::current_dir().unwrap();
             path.push("unminified-js");
-            path.push(host);
-            let _ = fs::remove_dir_all(&path);
+            //path.push(host);
+            //let _ = fs::remove_dir_all(&path);
```
CYBAI, line 1988: maybe should be removed if unused 👀?

components/script/dom/window.rs
```diff
@@ -1993,7 +1996,7 @@ impl Window {
                         self.unminified_js_dir.borrow()
                     );
                 },
-                Err(_) => warn!("Could not create unminified dir for {:?}", host),
+                Err(_) => warn!("Could not create unminified dir"), // for {:?}", host),
```
CYBAI, line 1999: ditto.

```suggestion
                Err(_) => warn!("Could not create unminified dir"),
```

PR #26716, opened 2020-05-29
Add UI for bypassing SSL handshake failures

### LINKED ISSUE #26683: Make it possible to accept self-signed certs
There are two main issues at the moment:
* the embedding API embeds the contents of resources/certs into the binary (https://github.com/servo/servo/blob/748b4241b7f2b7be7dc71be4df1ad89ddec49d99/ports/libsimpleservo/api/src/lib.rs#L926)
* we need to recognize when an unsupported cert causes the connection to be terminated and offer a UI that can bypass that error if desired

jdm:
There are several parts to these changes:
1. resurrecting the network error classification code to distinguish between SSL failures and other network errors
1. adding an SSL verification callback to support verifying certs against a list that can change at runtime, rather than just at program initialization
1. exposing a privileged chrome://allowcert URI which accepts the PEM cert contents along with a secret token
1. extracting the PEM cert contents out of the network layer when a handshake failure occurs, and getting them into the HTML that is parsed when an SSL failure occurs
1. adding a button in the handshake failure page that performs an XHR to chrome://allowcert with knowledge of the secret token and the PEM cert contents, before reloading the original URL that failed

The presence of the secret token means that while the chrome://allowcert URL is currently visible to web content, they cannot make use of it to inject arbitrary certs into the verification process.

---
- [x] These changes fix #26683
- [x] These changes do not require tests because the UI requires user activation and can't clearly be automated

### DIFF
--- a/components/net/connector.rs
+++ b/components/net/connector.rs
@@ -8,8 +8,12 @@ use hyper::client::HttpConnector as HyperHttpConnector;
 use hyper::rt::Future;
 use hyper::{Body, Client};
 use hyper_openssl::HttpsConnector;
-use openssl::ssl::{SslConnector, SslConnectorBuilder, SslMethod, SslOptions};
-use openssl::x509;
+use openssl::ex_data::Index;
+use openssl::ssl::{
+    SslConnector, SslConnectorBuilder, SslContext, SslMethod, SslOptions, SslVerifyMode,
+};
+use openssl::x509::{self, X509StoreContext};
+use std::sync::{Arc, Mutex};
 use tokio::prelude::future::Executor;
 
 pub const BUF_SIZE: usize = 32768;
@@ -60,7 +64,20 @@ impl Connect for HttpConnector {
 pub type Connector = HttpsConnector<HttpConnector>;
 pub type TlsConfig = SslConnectorBuilder;
 
-pub fn create_tls_config(certs: &str, alpn: &[u8]) -> TlsConfig {
+#[derive(Clone)]
+pub(crate) struct ExtraCerts(pub Arc<Mutex<Vec<Vec<u8>>>>);
+
+impl ExtraCerts {
+    pub(crate) fn new() -> Self {
+        Self(Arc::new(Mutex::new(vec![])))
+    }
+}
+
+lazy_static! {
+    static ref INDEX: Index<SslContext, ExtraCerts> = SslContext::new_ex_index().unwrap();
+}
+
+pub(crate) fn create_tls_config(certs: &str, alpn: &[u8], extra_certs: ExtraCerts) -> TlsConfig {
     // certs include multiple certificates. We could add all of them at once,
     // but if any of them were already added, openssl would fail to insert all
     // of them.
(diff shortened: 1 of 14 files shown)


jdm:
Screen Shot 2020-05-29 at 1 35 56 PM

Here's an example of loading the default Hubs local dev server with the new UI. Pressing the button reloads the page without a handshake error.

components/net/connector.rs
```diff
@@ -30,6 +35,38 @@ const SIGNATURE_ALGORITHMS: &'static str = concat!(
...
+    pub(crate) fn new() -> Self {
+        Self {
+            certs: Arc::new(Mutex::new(HashMap::new())),
+        }
+    }
+
+    fn store(&self, host: String, cert_bytes: Vec<u8>) {
+        let mut certs = self.certs.lock().unwrap();
+        let entry = certs.entry(host).or_insert((cert_bytes, 0));
+        entry.1 += 1;
```
asajeffrey, line 53: This looks a lot like a poor dev's weak Rc. Is it?

resources/badcert.html
```diff
@@ -3,6 +3,24 @@
 <title>Certificate error</title>
 </head>
 <body>
-    <p>${reason}</p>
+  <p>${reason}</p>
+  <pre id="bytes">${bytes}</pre>
+  <button id="allow">Allow certificate temporarily</button>
```
asajeffrey, line 8: Should there also be a "decline" button?

resources/badcert.html
```diff
@@ -3,6 +3,24 @@
...
+  <script>
+    let bytes = document.getElementById('bytes').textContent;
+    let button = document.getElementById('allow');
+    if (bytes.length) {
+        button.onclick = function() {
+            let xhr = new XMLHttpRequest();
+            xhr.open('GET', 'chrome:allowcert?secret=${secret}&bytes=' + btoa(bytes));
```
asajeffrey, line 15: I'm a bit concerned that we're using a GET request for this, side-effecting GET that adds a security wormhole scares me!

asajeffrey (commented):
This all looks like a sensible approach, apart from using a side-effecting GET as the way of bypassing cert checks.

resources/badcert.html
jdm: Perhaps. What should it do, however?

resources/badcert.html
jdm: What is the security wormhole, exactly?

jdm:
Note to self: unit tests don't build now.

components/net/connector.rs
jdm: It does have the property of removing the hashmap entry when the last user goes away, so it's slightly more complex than Rc/Arc.

resources/badcert.html
jdm: Firefox does have the Get Me Out Of Here button which I think does the same thing, so I'll add that.

components/net/connector.rs
Manishearth: Yeah I could go in either direction here

resources/badcert.html
Manishearth: Yeah IMO this should be a POST. Secrets in url params is bad

Manishearth (approved):
r=me with minor changes, and we should also be passng unit tests

jdm:
Fixed the unit tests, added on for accepting self-signed certs, and added a button to go back in history on the cert error page.
@bors-servo r=Manishearth

jdm:
@bors-servo retry
#25513

PR #26720, opened 2020-05-31
Show the real commit hash for `./servo --version`, not the bundle hash

### LINKED ISSUE #26386: repo bundles make git hash in nightly versions useless
The code to create a bundle ends up creating a git commit that doesn't match anything in the master branch: https://github.com/servo/servo/blob/4e650e83284bc25b781902085df4c25b39d95dd5/etc/taskcluster/decisionlib.py#L870-L883

This means that the code that [populates](https://github.com/servo/servo/blob/174bcc443435100da31b20db5ab684f4f3c8255b/components/config/lib.rs#L24) the binary's version information uses a hash that is meaningless: https://github.com/servo/servo/blob/306e8ac5f9b1f5b30c83f6861cfc3139273274bc/python/servo/command_base.py#L736-L742

We could potentially make CI set the GIT_INFO environment variable when before building to CONFIG.git_sha, and make the avoid overwriting that variable from the mach code if it's already set.

camelid:
Show the real commit hash of the build when run on a bundle commit, rather than showing the bundle's hash.

It gets the real commit hash by extracting it from the bundle commit message, which has the form `Shallow version of commit {sha1}`, where `{sha1}` is the real commit hash.

---

- [x] `./mach build -d` does not report any errors (edits Python code, no Rust changes)
- [X] These changes fix #26386 (GitHub issue number if applicable)

- [X] These changes do not require tests because this only changes infrastructure

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -741,10 +741,10 @@ install them, let us know by filing a bug!")
 
             git_sha = None
             # Check if it's a bundle commit
-            if git_commit_subject.startswith("Shallow version of commit "):
+            if git_commit_subject.startswith(b"Shallow version of commit "):
                 # This is a bundle commit
                 # Get the SHA-1 from the bundle subject: "Shallow version of commit {sha1}"
-                git_sha = git_commit_subject.split(' ')[-1].strip()
+                git_sha = git_commit_subject.split(b' ')[-1].strip()
                 # Shorten hash
                 # NOTE: Partially verifies the hash, but it will still pass if it's, e.g., a tree
                 git_sha = subprocess.check_output([
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -734,15 +734,27 @@ install them, let us know by filing a bug!")
 
         git_info = []
         if os.path.isdir('.git') and is_build:
-            # Get the subject of the bundle commit
-            git_bundle_subject = subprocess.check_output([
+            # Get the subject of the commit
+            git_commit_subject = subprocess.check_output([
                 'git', 'show', '-s', '--format=%s', 'HEAD'
             ]).strip()
-            # Get the SHA-1 from the bundle subject: "Shallow version of commit {sha1}"
-            git_sha = git_bundle_subject.split(' ')[-1]
-            # Verify that it's a valid commit
-            # NOTE: this will pass even if `git_sha` is 'master' or another branch or ref
-            subprocess.check_call(['git', 'cat-file', 'commit', git_sha])
+
+            git_sha = None
+            # Check if it's a bundle commit
+            if git_commit_subject.startswith("Shallow version of commit "):
+                # This is a bundle commit
+                # Get the SHA-1 from the bundle subject: "Shallow version of commit {sha1}"
+                git_sha = git_commit_subject.split(' ')[-1].strip()
+                # Shorten hash
+                # NOTE: Partially verifies the hash, but it will still pass if it's, e.g., a tree
+                git_sha = subprocess.check_output([
+                    'git', 'rev-parse', '--short', git_sha
+                ])
+            else:
+                # This is a regular commit
+                git_sha = subprocess.check_output([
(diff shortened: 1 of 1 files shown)


camelid:
I can't check to see if these changes work because I don't have access to the CI, so let me know if anything seems off!

camelid:
Oops, this code will fail when you run `./mach build` on a non-bundle commit.

camelid:
Weird... it seemed to work fine on my machine. Must be a Python version incompatibility with byte strings vs regular strings or something like that.

The error:

> TypeError: startswith first arg must be bytes or a tuple of bytes, not str

SimonSapin:
Rigth, this line

https://community-tc.services.mozilla.com/tasks/HdD7jA0MSSiXnhRyEJti6A/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FHdD7jA0MSSiXnhRyEJti6A%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L462

Shows that the error happens with Python 3.

We’re in the middle of the transition https://github.com/servo/servo/issues/23607, so IIRC most of `mach` needs to support both 2.x and 3.x at the moment.

`git_commit_subject` is the result of `subprocess.check_output` and so a byte string. In 2.x those are freely interchangeable with Unicode strings as long as the contents is ASCII. `"Shallow version of commit "` is a Unicode string in 3.x. You can make it a byte string (in both) by adding a `b` prefix before the opening quote. An similarly for the `' '` string passed to `split` in the next non-comment line.

camelid:
Ah, I see. I guess I should run mach with Python 3 on my system so that I catch those breaking changes :)

camelid:
It seems that the version is only updated if the config crate is changed or you do a fresh build. Is there a way to fix that?

SimonSapin:
It looks like `mach` sets the `GIT_INFO` environment variable, and `components/config/lib.rs` uses `option_env!("GIT_INFO")` to read it. rustc knows about this, but apparently Cargo doesn’t.

We could potentially fix this by adding a build script that prints `cargo:rerun-if-env-changed=GIT_INFO`: https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-env-changedname

The problem is that this seems to disable the default dependency tracking: https://github.com/rust-lang/cargo/issues/4587 so we’d need to also print `cargo:rerun-if-changed=lib.rs` and similarly for each source file. And that code could easily get out of sync with the actual set of source files.

However there is ongoing work in rustc https://github.com/rust-lang/rust/pull/71858 to expose data about dependencies on environment variables. Cargo can eventually read that data and do the right thing.

So, given that this will be fixed upstream eventually, I think it’s not worth doing the build script work-around. In the meantime you can run `touch components/config/lib.rs` to force a rebuild.

camelid:
> In the meantime you can run touch components/config/lib.rs to force a rebuild.

Could we just have mach run `touch`? I guess the only issue is then it would rebuild that part every time. Maybe just for release builds?

SimonSapin:
I would not be in favor:

* Like the build script, this is another kind of work-around that we’d have to maintain
* The proper fix is coming to rustc + cargo
* The consequences of this bug are extremely mild: CI (including Nightly builds) is not affected since it only builds once, and for local builds made by contributors the output of `--version` is not that important in my opinion.

camelid:
@SimonSapin could you review this? Thanks!

SimonSapin:
`try` builds don’t upload build artifacts that I could test, so I hacked together a build with the same configuration as Nightly at https://community-tc.services.mozilla.com/tasks/L5hH2dsmQ7CvGRXv-G5-fA#artifacts and verified that `./servo --version` prints `Servo 0.0.1-18a73cef2` which matches commit 18a73cef2aec62627bca257583ef1661bdd1beb1 that I pushed.

In short, looks good, thanks!

@bors-servo r+

camelid:
Thank you! :)

PR #26742, opened 2020-06-01
Implement GPUTexture and GPUTextureView

kunalmohan:
This also include changes to CodegenRust.py to allow enum values starting with digits.

r?@kvark

---


components/script/dom/gpudevice.rs
```diff
@@ -636,6 +650,86 @@ impl GPUDeviceMethods for GPUDevice {
         GPUCommandEncoder::new(&self.global(), self.channel.clone(), encoder, true)
     }
 
+    /// https://gpuweb.github.io/gpuweb/#dom-gpudevice-createtexture
+    fn CreateTexture(&self, descriptor: &GPUTextureDescriptor) -> DomRoot<GPUTexture> {
+        let mut valid = true;
+        let size = match descriptor.size {
```
kvark, line 656: you'll likely need that conversion in other places too, like in copy_buffer_to_texture and such

components/script/dom/gpudevice.rs
```diff
@@ -636,6 +650,86 @@ impl GPUDeviceMethods for GPUDevice {
...
+            .send(WebGPURequest::CreateTexture {
+                device_id: self.device.0,
+                texture_id,
+                descriptor: desc,
+            })
+            .expect("Failed to create WebGPU Texture");
+
+        let texture = webgpu::WebGPUTexture(texture_id);
+
+        let texture_size = match descriptor.size {
```
kvark, line 705: you already have this match above :)

components/script/dom/gputexture.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::dom::gputextureview::GPUTextureView;
+use dom_struct::dom_struct;
+use std::cell::Cell;
+use webgpu::{wgt, WebGPU, WebGPUDevice, WebGPURequest, WebGPUTexture, WebGPUTextureView};
+
+#[dom_struct]
+pub struct GPUTexture {
+    reflector_: Reflector,
+    texture: WebGPUTexture,
+    label: DomRefCell<Option<DOMString>>,
```
kvark, line 26: would it make sense to just put the whole GPUTextureDescriptor in here instead of all the fields separately?

components/script/dom/gputexture.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn SetLabel(&self, value: Option<DOMString>) {
+        *self.label.borrow_mut() = value;
+    }
+
+    /// https://gpuweb.github.io/gpuweb/#dom-gputexture-createview
+    fn CreateView(&self, descriptor: &GPUTextureViewDescriptor) -> DomRoot<GPUTextureView> {
+        let desc = wgt::TextureViewDescriptor {
+            label: Default::default(),
+            format: match descriptor.format {
+                Some(f) => convert_texture_format(f),
```
kvark, line 122: nit: could do `convert_texture_format(descriptor.format.unwrap_or(self.format))`

components/webgpu/lib.rs
```diff
@@ -457,10 +468,37 @@ impl WGPU {
                     let _ = gfx_select!(program_id =>
                         global.device_create_shader_module(device_id, &descriptor, program_id));
                 },
+                WebGPURequest::CreateTexture {
+                    device_id,
+                    texture_id,
+                    descriptor,
+                } => {
+                    let global = &self.global;
+                    let st = CString::new(descriptor.label.as_bytes()).unwrap();
```
kvark, line 477: that should probably be changed in a way that doesn't allocate `CString` if the label is not given, but we can follow up with that

kvark (changes requested):
It looks great! I have a few really minor notes.

components/script/dom/gpudevice.rs
kunalmohan: The difference is the type that is returned by the arms. I'll try to store a `GPUTextureDescriptor` instead, as you mentioned. We won't need this in that case.

components/webgpu/lib.rs
kunalmohan: Code related to labels needs to be updated on the client-side as well- use labels provided in the descriptor and update the webidl too. So, as you suggested, I'll handle all this in a follow-up PR.

components/script/dom/gputexture.rs
kunalmohan: I don't think it would do much benefit. Instead, it would make things a little hard to access- eg the label in the descriptor and size field. We'll have to use `match` everytime we need to access texture_size (like while creating `TextureView`). Currently I am storing `GPUExtent3DDict` instead of `GPUExtent3D` which makes it easier to access. The label is stored inside parent field in the descriptor as `Option<Option<DOMString>>`.

components/script/dom/gpudevice.rs
kunalmohan: I guess I would still need this separately. Or is there a better way to do this?

components/script/dom/gpudevice.rs
```diff
@@ -636,6 +650,86 @@ impl GPUDeviceMethods for GPUDevice {
...
+            GPUExtent3D::GPUExtent3DDict(ref dict) => GPUExtent3DDict {
+                width: dict.width,
```
kvark, line 707: shouldn't this just be `dict.clone()`?

components/script/dom/gpudevice.rs
kvark: It looks like you could have it in two stages: `GPUExtent3D -> GPUExtent3DDict`, and separately `GPUExtent3DDict -> wgpu::Extent3d`

components/script/dom/gpudevice.rs
kunalmohan: `GPUExtent3DDict` does not implement `Clone`. None of the structs generated from webidl dictionary type implement `Clone`.

components/script/dom/gpudevice.rs
kunalmohan: Okay, so we'll have a separate method for each. And for `GPUExtent3D -> wgt::Extent3d` we'll have to call both of them, right?

kvark: correct, yes

components/script/dom/gpudevice.rs
kvark: bah, that's sad

components/script/dom/gpudevice.rs
kunalmohan: I am usually bad with names :P, but here's what I am thinking of naming the methods- `convert_texture_size_to_dict` for the first one and `convert_texture_size_to_wgt` for the second. I'll change them to something else if you say so. Will save me the effort to push the changes again :)

kvark: that sounds ok to me :)

kvark (approved):

Issue #26639, opened 2020-05-25
Tracking: implementing Flexbox in Layout 2020


SimonSapin:
Specification: https://drafts.csswg.org/css-flexbox/

SimonSapin:
Style system aspects are already implemented in Stylo, but some properties or values need to be "un-disabled" (possibly moved to a Flexbox-specific pref while the implementation is incomplete) either [at compile-time](https://github.com/servo/servo/commit/183f15d5aacd290aa2bd2c3a2397cd456d170161#diff-d184b34f3faa0484771dd704b0c6bb19) or [at run-time](https://github.com/servo/servo/commit/c377d9c48ee0b450841e698286e2a73daca837ec#diff-d4b2d441d08a12968d7e0ecd65c92bfb).

Integration with the rest of layout code should happen through:

* A new variant of the `IndependentFormattingContextContents` (and `NonReplacedIFCKind`) enum that represents flex containers
  https://github.com/servo/servo/blob/d0bffa5b5e158322e7b9ae42a7a7386ddc336e42/components/layout_2020/formatting_contexts.rs#L44-L50
* Some new types to represent flex containers, flex lines, flex items, etc.
* The contents of each flex item should be represented as a `IndependentFormattingContext`, similar to `InlineLevelBox::Atomic`

mrobinson:
I think we can close this. The two main missing features are visibility collapse (#20142) and writing modes. The majority of this is finished though and we can rely on more specific bugs.

PR #26755, opened 2020-06-02
Add Layout 2020 box tree support for Flexbox, behind a pref

SimonSapin:
CC https://github.com/servo/servo/issues/26639

Layout support will come in future PRs. This one generates a zero-size fragment with no content.


SimonSapin:
Failures are legit, I messed up the `cfg` for serialization of `display: inline-table`. Pushed a fixup commit for easier review, I’ll squash it before landing.

I’ve also pushed other commits for sprinkling some rayon on.

SimonSapin:
This PR adds a `FlexLevel` variant to the `LayoutBox` enum. PR https://github.com/servo/servo/pull/26741 which landed in the meantime added code that matches on `LayoutBox`. So the merge does not conflict but fails to compile.

The fix is straightforward:

<details>

```diff
    fixup! Flex items in the box tree

diff --git components/layout_2020/flow/root.rs components/layout_2020/flow/root.rs
index 61f6ca888b..79285767d3 100644
--- components/layout_2020/flow/root.rs
+++ components/layout_2020/flow/root.rs
@@ -10,6 +10,7 @@ use crate::display_list::stacking_context::{
 };
 use crate::dom_traversal::{iter_child_nodes, Contents, NodeExt};
 use crate::element_data::LayoutBox;
+use crate::flexbox::FlexLevelBox;
 use crate::flow::construct::ContainsFloats;
 use crate::flow::float::FloatBox;
 use crate::flow::inline::InlineLevelBox;
@@ -119,6 +120,7 @@ impl BoxTree {
         enum UpdatePoint {
             AbsolutelyPositionedBlockLevelBox(ArcRefCell<BlockLevelBox>),
             AbsolutelyPositionedInlineLevelBox(ArcRefCell<InlineLevelBox>),
+            AbsolutelyPositionedFlexLevelBox(ArcRefCell<FlexLevelBox>),
         }
 
         fn update_point<'dom, Node>(
@@ -188,6 +190,16 @@ impl BoxTree {
                         },
                         _ => return None,
                     },
+                    LayoutBox::FlexLevel(flex_level_box) => match &*flex_level_box.borrow() {
+                        FlexLevelBox::OutOfFlowAbsolutelyPositionedBox(_)
+                            if box_style.position.is_absolutely_positioned() =>
+                        {
+                            UpdatePoint::AbsolutelyPositionedFlexLevelBox(
+                                flex_level_box.clone(),
+                            )
+                        },
+                        _ => return None,
+                    },
                 };
             Some((primary_style.clone(), display_inside, update_point))
         }
@@ -217,6 +229,12 @@ impl BoxTree {
                                 out_of_flow_absolutely_positioned_box,
                             );
                     },
+                    UpdatePoint::AbsolutelyPositionedFlexLevelBox(flex_level_box) => {
+                        *flex_level_box.borrow_mut() =
+                            FlexLevelBox::OutOfFlowAbsolutelyPositionedBox(
+                                out_of_flow_absolutely_positioned_box,
+                            );
+                    },
                 }
                 return true;
             }
```

</details>

components/layout_2020/flexbox.rs
```diff
@@ -104,18 +118,76 @@ where
...
+                ),
+            ))
+        };
+        self.children.push(box_.clone());
+        box_slot.set(LayoutBox::FlexLevel(box_))
+    }
+}
+
+/// https://drafts.csswg.org/css-text/#white-space
+fn is_only_document_white_space(string: &str) -> bool {
```
Manishearth: is this something that should be cached?

SimonSapin: It could be, but I’m not sure that would be a win. I’d guess that white-space only text nodes are typically small (at most a couple newlines plus indentation in non-minified HTML) so this should run quickly, and in other cases we’ll typically find a non-white-space character not far from the start where `Iterator::all` will short-circuit and we’ll also return quickly.

PR #26757, opened 2020-06-02
Remove unnecessary generics from low-level script code

jdm:
This should remove one of the larger offenders from the cargo-llvm-lines output for the script crate.

CC https://github.com/servo/servo/issues/26713


jdm:
Together these changes remove ~250k lines from the output of cargo-llvm-lines.

SimonSapin:
Nice!

@bors-servo r+

nox:
Just a side note: the passed function is always a `T::derives`, which is always implemented as `proto_chain[depth] == some_interface_id`, so we could probably get rid of the passed function entirely.

PR #26758, opened 2020-06-02
Don't create empty stacking contexts in display lists

### LINKED ISSUE #26590: Crash on altspace help page (index out of bounds: the len is 0 but the index is 0)
https://help.altvr.com/hc/en-us/articles/360016344513-How-do-I-upload-my-own-Kits- crashes in the nightly build when I navigate to it from https://help.altvr.com/hc/en-us.

### LINKED ISSUE #26592: Youtube crashes (index out of bounds: the len is 0 but the index is 0)
A bunch of content on the homepage loads in the 5/20 nightly, then there's a pause before some more content appears immediately followed by a crash.

jdm:
A recent change to euclid exposed that our display lists can contain Rects that contain NaN values. These NaNs originate from creating stacking contexts with transforms that scale the horizontal or vertical dimensions to 0. WebRender isn't prepared to handle these, so we need to not produce these empty stacking contexts when building the display list.

---
- [x] These changes fix #26592 and fix #26590

### DIFF
--- a/components/layout/fragment.rs
+++ b/components/layout/fragment.rs
@@ -2730,6 +2730,12 @@ impl Fragment {
             self.style().get_box().perspective != Perspective::None
     }
 
+    /// Returns true if this fragment has a transform applied that causes it to take up no space.
+    pub fn has_non_invertible_transform(&self) -> bool {
+        self.transform_matrix(&Rect::default())
+            .map_or(false, |matrix| !matrix.is_invertible())
+    }
+
     /// Returns true if this fragment establishes a new stacking context and false otherwise.
     pub fn establishes_stacking_context(&self) -> bool {
         // Text fragments shouldn't create stacking contexts.
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -448,6 +448,14 @@ impl Fragment {
                     return;
                 }
 
+                // If this fragment has a transform applied that makes it take up no spae
+                // then we don't need to create any stacking contexts for it.
+                let has_non_invertible_transform =
+                    fragment.has_non_invertible_transform(&containing_block_info.rect.to_untyped());
+                if has_non_invertible_transform {
+                    return;
+                }
+
                 fragment.build_stacking_context_tree(
                     fragment_ref,
                     builder,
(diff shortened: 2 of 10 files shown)


jdm:
No need to review this yet. A lot of tests are failing that I need to investigate.

jdm:
Weird, some tests run fine in headed mode, but panic in headless mode.

jdm:
@mrobinson This PR is now ready for review.

components/layout/fragment.rs
```diff
@@ -2730,6 +2730,14 @@ impl Fragment {
             self.style().get_box().perspective != Perspective::None
     }
 
+    /// Returns true if this fragment has a transform applied that causes it to take up no space.
+    pub fn has_non_invertible_transform(&self) -> bool {
+        let transform = self
+            .transform_matrix(&Rect::default())
+            .unwrap_or(LayoutTransform::identity());
+        !transform.is_invertible()
```
mrobinson, line 2738: This could potentially be: 

```rust
self.transform_matrix(&Rect::default())
    .map(|matrix| transform.is_invertible())
    .unwrap_or_else(false)
```

That would avoid creating an extra transformation matrix. Testing for invertability is sometiems expensive.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -778,6 +788,18 @@ impl BoxFragment {
         })
     }
 
+    /// Returns true if the given style contains a transform that is not invertible.
+    pub(crate) fn has_non_invertible_transform(
```
mrobinson, line 792: If you made this a method that took `&self` then I think you could avoid passing the style here.

components/layout_2020/display_list/mod.rs
```diff
@@ -330,6 +330,16 @@ impl<'a> BuilderForBoxFragment<'a> {
     }
 
     fn build(&mut self, builder: &mut DisplayListBuilder) {
+        // If this box fragment has a transform applied that makes it take up no space,
```
mrobinson, line 333: I think you can omit this check here, since you will not be adding this fragment to the `StackingContextTree`.

mrobinson (approved):
Nice! I think there is one check that is redudant, but otherwise this looks correct.

jdm:
@bors-servo retry
* #26834

PR #26770, opened 2020-06-03
Make to_shmem Gecko-only

SimonSapin:
As suggested in https://github.com/servo/servo/issues/26713#issuecomment-638185219

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5203,7 +5203,6 @@ dependencies = [
  "malloc_size_of_derive",
  "serde",
  "servo_rand",
- "to_shmem",
  "url",
  "uuid",
 ]
@@ -5818,7 +5817,6 @@ dependencies = [
  "servo_arc",
  "smallbitvec",
  "smallvec 1.3.0",
- "string_cache",
  "thin-slice",
 ]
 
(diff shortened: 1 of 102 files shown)


components/style/gecko_string_cache/namespace.rs
```diff
@@ -25,17 +25,9 @@ macro_rules! ns {
 
 /// A Gecko namespace is just a wrapped atom.
 #[derive(
-    Clone,
```
jdm, line 28: I don't think we need to add feature checks to any of the style/gecko/ or style/gecko_string_cache files.

jdm (approved):

components/style/gecko_string_cache/namespace.rs
SimonSapin: Good point. Amended.

emilio:
@bors-servo r-

Can we discuss this for a second? This is going to make syncs even more painful and if I understand correctly this affects llvm IR but not generated code, right?

SimonSapin:
Correct, this is about reducing compile times. I would expect the linker to remove functions that are never called. (Maybe even without LTO? IIRC with `--gc-sections` is the default in rustc.)

I hadn’t considered the indirect impact on sync as you explained it on Matrix (mozilla-central contributors accidentally breaking the Servo configuration more often as it diverges more).

`to_shmem_slice_ptr` does show up in https://github.com/servo/servo/issues/26713#issuecomment-638112283 but with only 0.4% of IR lines in libscript. This is probably not worth the downside, so closing.

PR #26771, opened 2020-06-03
Update raqote and associated dependencies

atouchet:
---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1528,9 +1528,9 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "font-kit"
-version = "0.7.0"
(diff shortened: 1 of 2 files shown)


atouchet:
`▶ Unexpected subtest result in /html/canvas/element/path-objects/2d.path.arc.shape.1.html:
│ FAIL [expected PASS] arc() from 0 to pi does not draw anything in the wrong half
│   → assert_equals: Red channel of the pixel at (20, 48) expected 0 but got 63`

These canvas tests look related.

atouchet:
Had an issue with a change. @jdm can you re-try?

atouchet:
The test failures appear to have been due to updating lyon_geom so I reverted that bump. I think this should be good to go now.

jdm:
@bors-servo r+
It sounds like we should file an issue on lyon_geom.

PR #26792, opened 2020-06-05
Replace `static mut` with `const`, `static`+`AtomicPtr`, or `static`+`UnsafeCell`

### LINKED ISSUE #26550: Replace uses of "static mut" with UnsafeCell

SimonSapin:
Fixes https://github.com/servo/servo/issues/26550


gterzian (approved):
LGTM on the sampler changes.

PR #26793, opened 2020-06-05
Pref api

### LINKED ISSUE #26445: Ability to toggle prefs on Hololens
It would be nice to be able to set prefs (at least, from a reduced list of prefs, if not arbitrary ones) on the Hololens. Even better if we can make the prefs persist by editing an on-disk prefs.json (which is currently [hardcoded in the binary](https://github.com/servo/servo/blob/master/ports/libsimpleservo/api/src/lib.rs#L918)).

Initially it might just be worth exposing a couple high-value prefs like `dom.webxr.hands.enabled` and `dom.webgl2.enabled`.

If we're doing a limited set of prefs we can also just persist the settings for those prefs.

cc @paulrouget

paulrouget:
Needs a bit more tweaking, but this should be ready soon.

Fix #26445

Screen Shot 2020-06-03 at 14 59 57


components/config/opts.rs
```diff
@@ -640,18 +639,6 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
...
-        "pref",
-        "A preference to set to enable",
-        "dom.bluetooth.enabled",
-    );
-    opts.optmulti(
-        "",
-        "pref",
-        "A preference to set to enable",
-        "dom.webgpu.enabled",
-    );
```
paulrouget, line 654: This is moved under the winit port.

components/config/prefs.rs
```diff
@@ -58,42 +54,13 @@ pub fn pref_map() -> &'static Preferences<'static, Prefs> {
     &PREFS
 }
 
-pub(crate) fn add_user_prefs() {
-    if let Some(path) = user_prefs_path() {
-        init_user_prefs(path);
+pub fn add_user_prefs(prefs: HashMap<String, PrefValue>) {
+    if let Err(error) = PREFS.set_all(prefs.into_iter()) {
+        writeln!(&mut stderr(), "Error setting preference: {:?}", error)
+            .expect("failed printing to stderr");
```
paulrouget, line 60: Instead of reading userconfig/prefs.json, we  let the embedder decide how to read the user pref. The embedder calls add_user_prefs.

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -55,14 +58,14 @@ pub use servo::embedder_traits::EventLoopWaker;
...
-    pub url: Option<String>,
     pub coordinates: Coordinates,
     pub density: f32,
     pub xr_discovery: Option<webxr::Discovery>,
     pub enable_subpixel_text_antialiasing: bool,
     pub gl_context_pointer: Option<*const c_void>,
     pub native_display_pointer: Option<*const c_void>,
     pub native_widget: *mut c_void,
+    pub prefs: Option<HashMap<String, PrefValue>>,
 }
```
paulrouget, line 69: Now that we can pass preferences to the embedder, there's not need to also pass the URL, and the embedder can just set homepage pref.

ports/libsimpleservo/capi/src/lib.rs
```diff
@@ -532,6 +537,7 @@ pub extern "C" fn resize(width: i32, height: i32) {
 pub extern "C" fn perform_updates() {
     catch_any_panic(|| {
         debug!("perform_updates");
+        prefs::free_prefs();
```
paulrouget, line 540: Free the C-compatible structures we allocated to build the preferences list.

ports/libsimpleservo/capi/src/prefs.rs
```diff
@@ -0,0 +1,241 @@
...
+use std::cell::RefCell;
+use std::collections::{BTreeMap, HashMap};
+use std::ffi::{CStr, CString};
+use std::os::raw::{c_char, c_void};
+
+thread_local! {
+    // CPREFS keeps alive a set of CPref that are sent over to the embedder.
+    // The CPREFS are structs holding pointers to values held alive by LOCALCPREFS.
+    static CPREFS: RefCell<Vec<CPref>> = RefCell::new(Vec::new());
+    static LOCALCPREFS: RefCell<BTreeMap<String, LocalCPref>> = RefCell::new(BTreeMap::new());
```
paulrouget, line 21: Emptied in perform_udpates.

support/hololens/ServoApp/App.xaml
```diff
@@ -4,4 +4,8 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:ServoApp">
 
+    <Application.Resources>
+        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
+    </Application.Resources>
```
paulrouget, line 9: New package used to implement number textbox.

paulrouget:
The UWP app now has settings, that are mirrored with Servo's preferences. The changes persist across session.

The devtools button now opens a "toolbox" with the pref panel. We can add new panels in the future (console!).

I'm not very happy with the number of new preferences struct/types introduced. It's confusing. Also freeing the C-struct on perform_updates might not been ideal. I can look at a better lifetime management. That's something I'd like to do maybe later (this PR got bigger and bigger, really needs to land).

@jdm r?

jdm:
r? @manishearth
I'm out today.

Manishearth (approved):
I like that this uses LocalSettings to be persistent!

Manishearth:
Yeah the C-Rust-lifetime management stuff here isn't great, but I'm okay with that being fixed later.

ports/libsimpleservo/capi/src/prefs.rs
Manishearth: Leave a comment?

ports/libsimpleservo/capi/src/lib.rs
Manishearth: Add a comment?

paulrouget:
- Added comments
- Fixed tests (moved pref parsing tests under ports/)

@bors-servo try

PR #26794, opened 2020-06-05
Update XHR send to use XMLHttpRequestBodyInit

### LINKED ISSUE #26723: Spec update: use XMLHttpRequestBodyInit for XHR
This will remove the `ReadableStream` option from the optional "body init" of an XHR, so it will require updating the code at https://github.com/servo/servo/blob/1dc41c8678034f3894184e64f2d2a7822cd6e73c/components/script/dom/xmlhttprequest.rs#L550

To do once https://github.com/whatwg/xhr/issues/277 is done.

gterzian:
FIX #26723

---


gterzian:
Failure was https://github.com/servo/servo/issues/26804

gterzian:
@jdm r? Follow-up on the streams PR, I think this might be a quick r for you...

tests/wpt/metadata/MANIFEST.json
```diff
@@ -527055,42 +527055,6 @@
       }
      ]
     ],
-    "send-data-readablestream.any.js": [
```
jdm, line 527058: We should ensure there is a test that verifies the behaviour of xhr send when a ReadableStream argument is passed. This will make it clear on wpt.fyi if browsers have actually converged on the spec in the future.

jdm (changes requested):

gterzian:
@bors-servo try=wpt

@jdm I've added one test to check that `ReadableStream` is treated as a string if passed as an argument, which is what I understand should be the default behavior(see https://github.com/whatwg/xhr/issues/277#issuecomment-635800860)

tests/wpt/web-platform-tests/xhr/setrequestheader-content-type.htm
```diff
@@ -215,23 +215,11 @@
...
-        undefined,
-        "ReadableStream request with under type sends no Content-Type without setRequestHeader() call"
-      )
-      request(
-        function _ReadableStream() { return new ReadableStream() },
-        {"Content-Type": "application/xml;charset=ASCII"},
-        "application/xml;charset=ASCII",
-        "ReadableStream request keeps setRequestHeader() Content-Type and charset"
+        "text/plain;charset=UTF-8",
+        'ReadableStream request is treated as a String request."'
```
jdm, line 222: This test doesn't verify anything about what actual bytes are present in the request due to how the readablestream object is converted to a string. I think it would be wise to do that to avoid accidental incompatibilities.

gterzian: Ok, thanks for checking, I think I have now added the right test, please take another look.

gterzian:
error is `curl: (56) OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104`...

gterzian:
@jdm r?

jdm (approved):
Thanks!

PR #26805, opened 2020-06-05
layout_2020: Tag fragments with their pseudo content type

mrobinson:
This will allow us to answer queries and properly handle animations in
the future for fragments generated for pseudo content.

---

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -27,9 +27,50 @@ use style::values::generics::counters::Content;
 use style::values::generics::counters::ContentItem;
 
 #[derive(Clone, Copy, Debug)]
-pub enum WhichPseudoElement {
+pub(crate) enum WhichPseudoElement {
     Before,
     After,
+}
+
+/// A data structure used to pass and store related layout information together to
+/// avoid having to repeat the same arguments in argument lists.
+#[derive(Clone)]
+pub(crate) struct NodeAndStyleInfo<Node> {
+    pub node: Node,
+    pub pseudo_element_type: Option<WhichPseudoElement>,
+    pub style: ServoArc<ComputedValues>,
+}
+
+impl<Node> NodeAndStyleInfo<Node> {
+    fn new_with_pseudo(
+        node: Node,
+        pseudo_element_type: WhichPseudoElement,
+        style: ServoArc<ComputedValues>,
+    ) -> Self {
+        Self {
+            node,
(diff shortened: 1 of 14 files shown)


components/layout_2020/dom_traversal.rs
```diff
@@ -42,12 +42,12 @@ pub(super) enum Contents {
...
     /// Content of a `::before` or `::after` pseudo-element that is being generated.
     /// <https://drafts.csswg.org/css2/generate.html#content>
-    OfPseudoElement(Vec<PseudoElementContentItem>),
+    OfPseudoElement(WhichPseudoElement, Vec<PseudoElementContentItem>),
 }
 
 pub(super) enum NonReplacedContents {
     OfElement,
-    OfPseudoElement(Vec<PseudoElementContentItem>),
+    OfPseudoElement(WhichPseudoElement, Vec<PseudoElementContentItem>),
```
SimonSapin, line 50: Why these new fields? There are at least some cases where they are redundant since a parameter `pseudo_element_type` is passed together with a `contents` parameter.

SimonSapin (commented):
This adds another parameter to many `Something::construct` methods, each of which already has many parameters.

What do you think of adding a new `struct` that contains `impl NodeExt<'dom>` together with `Option<WhichPseudoElement>`? And while we’re at it, maybe `&Arc<ComputedValues>` and `&LayoutContext` too? (And maybe that struct itself should be passed by reference?)

mrobinson:
> What do you think of adding a new `struct` that contains `impl NodeExt<'dom>` together with `Option<WhichPseudoElement>`? And while we’re at it, maybe `&Arc<ComputedValues>` and `&LayoutContext` too? (And maybe that struct itself should be passed by reference?)

Okay. I've started passing a new structure called `NodeAndStyleInfo`. This doesn't always pass the LayoutContext, because there are times when we don't need it and also need to clone the data structure. Introducing lifetimes for TextRuns, etc seemed a bit tricky for this patch, but could be part of a future one. I'm passing the data structure by reference, though it holds a `Arc<ComputedValues>`. This shouldn't add too much more refcount traffic since I tried to take ownership of the preexisting stack values.

components/layout_2020/dom_traversal.rs
mrobinson: Hah. These aren't needed at all. I think I just got a bit overzealous here. 😅 I've fixed this.

SimonSapin:
Looks great, thank you!

@bors-servo r+

PR #26806, opened 2020-06-05
Include animations and transitions in the cascade

mrobinson:
Instead of applying animations and transitions to styled elements,
include them in the cascade. This allows them to interact properly with
things like font-size and !important rules.

---


components/style/matching.rs
```diff
@@ -493,6 +528,20 @@ trait PrivateMatchMethods: TElement {
...
+
+        // If we have modified animation or transitions, we recascade style for this node.
+        if needs_animations_update || might_need_transitions_update {
+            let mut resolver = StyleResolverForElement::new(
+                *self,
+                context,
+                RuleInclusion::All,
+                PseudoElementResolution::IfApplicable,
+            );
+            let new_primary = resolver.resolve_style_with_default_parents();
```
emilio, line 540: I'm a bit confused about this. Why does this need to `resolve_*`, then re-cascade?

mrobinson: You're right. Re-cascading here is bogus. I've fixed this in the latest version of the branch.

mrobinson:
@emilio, thanks for taking a look at this. I've addressed your comment and also have made a few changes in order to avoid re-resolving the style when animations or transitions aren't changing.

components/style/matching.rs
emilio: So... I'm still a bit unclear on why this is the right thing to do. This is not just recascading but also re-selector-matching. I'd expect this to do basically just [this](https://searchfox.org/mozilla-central/rev/9ad88f80aeedcd3cd7d7f63be07f577861727054/servo/components/style/traversal.rs#622-635).

And to be clear, I think this is fine as a stepping-stone if needed, but I don't understand why something like that wouldn't work.

mrobinson: You're right. I think I was trying this before, but not properly replacing rules. I'll need to do it a bit more carefully because `replace_rules` is written with the animations-only restyle in mind. Thanks for the feedback.

mrobinson:
I've pushed a new commit to the branch that replaces the transition and animations rules directly and then does a re-cascade.

mrobinson:
I have a pushed a new version of the branch that uses `StyleResolver::.cascade_style_and_visited_with_default_parents` instead of `cascade_styles_with_default_parents`, simplifying the second commit.

components/style/animation.rs
```diff
@@ -875,17 +864,30 @@ pub struct ElementAnimationSet {
...
     pub fn cancel_all_animations(&mut self) {
+        self.dirty = !self.animations.is_empty();
         for animation in self.animations.iter_mut() {
             animation.state = AnimationState::Canceled;
         }
+        self.cancel_active_transitions();
+    }
+
+    fn cancel_active_transitions(&mut self) {
+        self.dirty = !self.transitions.is_empty();
```
emilio, line 885: Does this dirty too much? Should it just dirty right before the `transition.state = ` assignment?

components/style/matching.rs
```diff
@@ -458,10 +465,60 @@ trait PrivateMatchMethods: TElement {
...
+        if !self.process_animations_for_style(
+            context,
+            old_values,
+            &mut new_resolved_styles.primary.style.0,
+        ) {
+            return;
+        }
+
+        // If we have modified animation or transitions, we recascade style for this node.
+        let mut rule_node = new_resolved_styles.primary.style.0.rules().clone();
```
emilio, line 481: May be worth exposing a `ResolvedElementStyles::primary_style()` or something so that this reads a bit better, but no big deal.

components/style/matching.rs
```diff
@@ -458,10 +465,60 @@ trait PrivateMatchMethods: TElement {
...
+            &mut rule_node,
+        );
+        Self::replace_single_rule_node(
+            context,
+            CascadeLevel::Animations,
+            self.animation_rule(&context.shared)
+                .as_ref()
+                .map(|a| a.borrow_arc()),
+            &mut rule_node,
+        );
```
emilio, line 497: May be worth either asserting that `rule_node` is not the primary style's rule node anymore, or check for that and bail out.

emilio (approved):
Thank you, this looks pretty sensible!

Issue #26709, opened 2020-05-29
Streams: when streaming a blob as a request body, bypass SpiderMonkey


gterzian:
Follow-up on https://github.com/servo/servo/pull/25873

In that PR, when a `Blob` is used as the `BodyInit` of a `Request`, we get a `ReadableStream` from the blob(see `impl Extractable for Blob`), then later as part of transmitting the body over the network(see `TransmitBodyConnectHandler`), we read chunks from the stream through SpiderMonkey(unless the blob has already been read into memory). 

There's a couple of problem with that:

1. It doesn't work with sync XHR(the event-loop is blocked so SM cannot read from the JS stream),
2. It's unnecessary to go though the JS stream, since there is in fact no user provided JS stream.

So we should look into building a native stream, or perhaps no stream at all, in such a case. 

The "no stream at all" might be feasible, since the data is coming from the `net` component, and that's also where it's going. Perhaps we can just pass around the file ID and then read it directly  when transmitting the body, like is done when doing a `scheme_fetch` (see `net/fetch/method.rs`)

PR #26810, opened 2020-06-07
Fix infinite stream and its missing incumbent script environment when newing a new stream

### LINKED ISSUE #26807: Searching with duckduckgo.com/html freezes Servo
URL: https://duckduckgo.com/html

Steps to reproduce:

1. Load https://duckduckgo.com/html in Servo.
2. Search something in the DuckDuckGo search bar.
3. Servo will then freeze.

This issue first occurred in the 2020-06-04 nightly build. Tested in Windows 10.

Regression range: https://github.com/servo/servo/compare/ea491b41174edd03c3ee08fe29046c33b96b0406...8536cee72cca0f8a4d9f7a14b47a8b16ed356179

CYBAI:
---
- [x] These changes fix #26807 

### DIFF
--- a/components/script/body.rs
+++ b/components/script/body.rs
@@ -55,7 +55,7 @@ pub enum BodySource {
     Null,
     /// Another Dom object as source,
     /// TODO: store the actual object
-    /// and re-exctact a stream on re-direct.
+    /// and re-extract a stream on re-direct.
     Object,
 }
 
@@ -72,6 +72,7 @@ struct TransmitBodyConnectHandler {
     bytes_sender: Option<IpcSender<Vec<u8>>>,
     control_sender: IpcSender<BodyChunkRequest>,
     in_memory: Option<Vec<u8>>,
+    in_memory_done: bool,
     source: BodySource,
 }
 
@@ -91,14 +92,22 @@ impl TransmitBodyConnectHandler {
             bytes_sender: None,
             control_sender,
             in_memory,
+            in_memory_done: false,
             source,
         }
     }
 
+    /// Reset `in_memory_done`, called when a stream is
+    /// re-extracted from the source to support a re-direct.
+    pub fn reset_in_memory_done(&mut self) {
+        self.in_memory_done = false;
+    }
+
     /// Re-extract the source to support streaming it again for a re-direct.
     /// TODO: actually re-extract the source, instead of just cloning data, to support Blob.
-    fn re_extract(&self, chunk_request_receiver: IpcReceiver<BodyChunkRequest>) {
+    fn re_extract(&mut self, chunk_request_receiver: IpcReceiver<BodyChunkRequest>) {
         let mut body_handler = self.clone();
+        body_handler.reset_in_memory_done();
 
         ROUTER.add_route(
             chunk_request_receiver.to_opaque(),
@@ -126,11 +135,20 @@ impl TransmitBodyConnectHandler {
     /// TODO: this method should be deprecated
     /// in favor of making `re_extract` actually re-extract a stream from the source.
     /// See #26686
-    fn transmit_source(&self) {
+    fn transmit_source(&mut self) {
+        if self.in_memory_done {
+            // Step 5.1.3
+            self.stop_reading();
+            return;
+        }
+
         if let BodySource::Null = self.source {
             panic!("ReadableStream(Null) sources should not re-direct.");
         }
+
         if let Some(bytes) = self.in_memory.clone() {
+            // The memoized bytes are sent so we mark it as done again
+            self.in_memory_done = true;
             let _ = self
                 .bytes_sender
                 .as_ref()
@@ -155,6 +173,12 @@ impl TransmitBodyConnectHandler {
 
     /// The entry point to <https://fetch.spec.whatwg.org/#concept-request-transmit-body>
     fn transmit_body_chunk(&mut self) {
+        if self.in_memory_done {
+            // Step 5.1.3
+            self.stop_reading();
+            return;
+        }
+
         let stream = self.stream.clone();
         let control_sender = self.control_sender.clone();
         let bytes_sender = self
@@ -165,6 +189,9 @@ impl TransmitBodyConnectHandler {
         // In case of the data being in-memory, send everything in one chunk, by-passing SpiderMonkey.
         if let Some(bytes) = self.in_memory.clone() {
             let _ = bytes_sender.send(bytes);
+            // Mark this body as `done` so that we can stop reading in the next tick,
+            // matching the behavior of the promise-based flow
+            self.in_memory_done = true;
             return;
         }
 
--- a/components/net_traits/request.rs
+++ b/components/net_traits/request.rs
@@ -117,7 +117,7 @@ pub enum ParserMetadata {
 }
 
 /// <https://fetch.spec.whatwg.org/#concept-body-source>
-#[derive(Clone, Debug, Deserialize, MallocSizeOf, Serialize)]
+#[derive(Clone, Debug, Deserialize, MallocSizeOf, PartialEq, Serialize)]
 pub enum BodySource {
     Null,
     Object,
@@ -178,10 +178,7 @@ impl RequestBody {
     }
 
     pub fn source_is_null(&self) -> bool {
-        if let BodySource::Null = self.source {
-            return true;
-        }
-        false
+        self.source == BodySource::Null
     }
 
     pub fn len(&self) -> Option<usize> {
--- a/components/script/dom/readablestream.rs
+++ b/components/script/dom/readablestream.rs
@@ -6,7 +6,7 @@ use crate::dom::bindings::conversions::{ConversionBehavior, ConversionResult};
 use crate::dom::bindings::error::Error;
 use crate::dom::bindings::reflector::{reflect_dom_object, DomObject, Reflector};
 use crate::dom::bindings::root::DomRoot;
-use crate::dom::bindings::settings_stack::AutoEntryScript;
+use crate::dom::bindings::settings_stack::{AutoEntryScript, AutoIncumbentScript};
 use crate::dom::bindings::utils::get_dictionary_property;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::promise::Promise;
@@ -118,6 +118,7 @@ impl ReadableStream {
         source: ExternalUnderlyingSource,
     ) -> DomRoot<ReadableStream> {
         let _ar = enter_realm(global);
+        let _ais = AutoIncumbentScript::new(global);
         let cx = global.get_cx();
 
         let source = Rc::new(ExternalUnderlyingSourceController::new(source));


CYBAI:
r? @gterzian or @jdm 

hmmm.... I tried several ways for an unit test but didn't figure out a workable version :/

a simplest test is like this; I confirmed it will get into the IPC router but we will even get it pass without patches in this PR... 🤔 

```js
promise_test((t) => {
  var blob = new Blob(["This is my blob content"], { type: "text/plain" });

  var formData = new FormData();
  formData.append("blob", blob, "blob.txt");

  return fetch("/fetch/api/resources/echo-content.py", {
    method: "POST",
    body: formData,
  });
}, "Fetch with POST blob");
```

CYBAI:
@bors-servo try=wpt

- Just curious will this accidentally fix some tests?

components/script/body.rs
```diff
@@ -72,6 +73,7 @@ struct TransmitBodyConnectHandler {
     bytes_sender: Option<IpcSender<Vec<u8>>>,
     control_sender: IpcSender<BodyChunkRequest>,
     in_memory: Option<Vec<u8>>,
+    in_memory_done: Cell<bool>,
```
gterzian, line 76: I think it can be just a `bool`, since you have a `&mut self` everywhere.

components/script/body.rs
```diff
@@ -155,6 +158,12 @@ impl TransmitBodyConnectHandler {
 
     /// The entry point to <https://fetch.spec.whatwg.org/#concept-request-transmit-body>
     fn transmit_body_chunk(&mut self) {
+        if self.in_memory_done.get() {
+            // Step 5.1.3
+            let _ = self.control_sender.send(BodyChunkRequest::Done);
```
gterzian, line 163: You can actually directly do `self.stop_reading()`.

gterzian (changes requested):
Thanks for the fixes! Just a few comments...

I'm not sure about the tests, actually I was surprised this wasn't caught before. Maybe you need a bit of a delay like would happen in a real network, you can try using https://web-platform-tests.org/writing-tests/server-pipes.html#trickle

components/script/body.rs
CYBAI: Indeed

CYBAI:
hmm... interesting, even I do `?pipe=trickle(100:d1:r2)`, I can still pass the test on current master :/

CYBAI:
ah!! cool! I think `100` bytes is too big, with changing it to `1` byte, I can get a TIMEOUT on current master now! Let me try to verify if it works for this PR!

CYBAI:
> ah!! cool! I think `100` bytes is too big, with changing it to `1` byte, I can get a TIMEOUT on current master now! Let me try to verify if it works for this PR!

hmm, this is not true; `1:d1:r2` will also TIMEOUT in other browsers :/ but if I try `10:d1:r2`, Servo on current master won't crash either... 🤔

CYBAI:
> 💔 Test failed - [status-taskcluster](https://community-tc.services.mozilla.com/tasks/groups/UlxEkUi9S3yiAk9pP3PBPg)

ok, I can reproduce these `TIMEOUT`s locally. Will try to investigate it 👀

gterzian:
I know what the problem is, so let me know if you want to try to find it yourself first 😄

gterzian:
Actually I'm not so sure I know what the problem is, but yeah the `TransmitBodyConnectHandler` is definitely manipulated as part of re-direct, so I'd say just add some print statements at the various places and you should be able to figure it out...

gterzian:
Yeah reading on it, it seems if you close a file descriptor, you loose all data in transit, so it reads a bit like dropping the chan might close the receiver, and remove the route, even though there is still a message in transit. 

@CYBAI you could try to add a `old_chans: Vec<IpcSender<BodyChunkRequest>>` to `RequestBody`, and then before doing https://github.com/servo/servo/blob/99ab1092ffb6e80d7e980932c5c2173c4c014a86/components/net_traits/request.rs#L176

do a `self.old_chans.push(self.chan.clone())`, which should keep the route alive while the request is being re-directed. 

I think it's the `self.stop_reading` that now happens earlier, and it results in removing the route over at https://github.com/servo/servo/blob/0645d8c36d734d77e6a783148df593deb7cfb7ee/components/net/http_loader.rs#L449
which then drops it's `IpcSender<BodyChunkRequest>`, and then the `RequestBody` still has a sender, however it is then dropped on the next `take_stream`, and then the route in `body.rs` closes before the new route for the re-direct in setup in response to receiving the `BodyChunkRequest::Extract` message. 

So previously we would not actually call the `self.stop_reading`, or do it later in the promise callback, so now it happens earlier and then senders drop too soon. 

Maybe, it's worth a try...

CYBAI:
@gterzian thanks for the tip! I will try it later!

CYBAI:
@gterzian I tried to debug with removing the `if self.in_memory_done` checking and yes, looks like we drop the IPC router too early.

I haven't tried the `old_chans` way though, I try to read the spec again.

Then, I just realized, in the [step 5.1.3 of request transmit body](https://fetch.spec.whatwg.org/#concept-request-transmit-body), it says `if read is fulfilled with an object whose done property is true`.

The `done` property means the `done` property from the readablestream. Looks like it's handled by SM and we can simply just get the value.

https://github.com/servo/servo/blob/0645d8c36d734d77e6a783148df593deb7cfb7ee/components/script/dom/readablestream.rs#L503-L505

I'm wondering, if we can know the stream is done or not by that property, maybe we don't need to add a flag like `in_memory_done` by ourselves?

But, here comes my question:

I found I don't really understand why if we have something in memory, we will just send it and return 🤔 

If we can check if done or not by the `done` property, maybe it's cleaner than adding a `old_chans`? WDYT?

https://github.com/servo/servo/blob/0645d8c36d734d77e6a783148df593deb7cfb7ee/components/script/body.rs#L165-L169

gterzian:
Yes, thanks for looking into these matters. 

So in the case where this problem happens, we have all the bytes in memory, and we don't actually go through the stream and SpiderMonkey. So we're not following the spec to the letter in that case, and I think that's ok(there will be many more cases like this, actually we should find a better way to do this, see for example https://github.com/servo/servo/issues/26709). 

So we need the `in_memory_done` flag, because we don't actually read from the stream and we will not get the `done` flag from the promise. 

> yes, looks like we drop the IPC router too early.

Then I think using this `old_chans` approach to keep the senders alive for the life of the request, is worth a try to see if it fixes it.

CYBAI:
> So we need the `in_memory_done` flag, because we don't actually read from the stream and we will not get the done flag from the promise.

Indeed! When there's no need to do chunks, it makes sense to just return the in-memory bytes!

Okay! I will give `old_chans` a try! Will update anything here again!

CYBAI:
> So previously we would not actually call the self.stop_reading, or do it later in the promise callback, so now it happens earlier and then senders drop too soon.

@gterzian I found I don' fully understand what you mean here. Do you mean, we shouldn't drop the `TransmitBodyConnectHandler` when `in_memory` is `done`? If yes, then what I don't understand is how we should drop it 🤔 

Does that mean we should drop it in this router?

https://github.com/servo/servo/blob/aaa6cea57fab894bb06bfa9b14e08cd7625ea48b/components/script/body.rs#L338

gterzian:
It's just what when we do `stop_reading`, then this will drop the sender corresponding to the route in `obtain_response`. The changes here call `stop_reading` earlier than before, so the route in `obtain_response` will drop earlier. 

So then I think this is causing the current problem, because there are two places that own a `IpcSender<BodyChunkRequest>`:

1. The route in `http_loader`,
2. The `RequestBody`. 

The sender in `RequestBody` drops [here](https://github.com/servo/servo/blob/99ab1092ffb6e80d7e980932c5c2173c4c014a86/components/net_traits/request.rs#L176)

So I think previously, during a re-direct, when `RequestBody` would drop it's sender, then the route in `obtain_response` would still have it's sender, because `stop_reading` had not been called yet(but soon would be). 

So now we're calling `stop_reading` "earlier", which means that when `RequestBody` drops it's sender, then that's actually the last sender and the route with the `TransmitBodyConnectHandler` is removed, and then we get the timeout. 

Or at least I think that is what is happening. And we can solve it by not dropping those sender upon a redirect in `RequestBody`, instead storing a clone in a `Vec<IpcSender<BodyChunkRequest>>` on `RequestBody`. (Note that the "old" chan is swapped for a new one on each re-direct, so instead of dropping the old ones we just store them in the vector to keep the routes alive).

CYBAI:
Thanks for your explanation again! I've now understood how the channels are dropped before this PR.

but sorry, I still don't get how I should fix the "drop too early issue" for this PR.
Do you mean we just add the `old_chans` for keeping the channels alive? If yes, I just tried it and it doesn't help; it seems we can't simply keep it in the vector to keep it alive.

I think we might need to find a proper place to handle the `in_memory_done` checking 🤔 ?

```rust
if self.in_memory_done {
  // Step 5.1.3
  self.stop_reading();
  return;
}
```

gterzian:
Ok I have something that is a confirmed fix locally, I actually found a way to remove the whole `read_from` flag and plug the behavior more into the spec. 

Thinking of opening a separate PR for that specific fix, then this one can be rebased off of it. What do you think?

I think the last thing we need to do here is to make sure that the behavior of `transmit_body_chunk` and `transmit_source` are similar when it comes to `in_memory_done`.

CYBAI:
> Ok I have something that is a confirmed fix locally, I actually found a way to remove the whole `read_from` flag and plug the behavior more into the spec.
> Thinking of opening a separate PR for that specific fix, then this one can be rebased off of it. What do you think?

I'm happy to learn what happened and how to fix it from your PR! 
 
> I think the last thing we need to do here is to make sure that the behavior of `transmit_body_chunk` and `transmit_source` are similar when it comes to `in_memory_done`.

okay, I will check your PR and see how to fix it!

gterzian:
> okay, I will check your PR and see how to fix it!

So this last part is something that needs to be done, I didn't do it in the other PR. What needs to be done is:

1. Reset the `in_memory_done` flag for the cloned handler as part of `re_extract`.
2. Add similar logic with the `in_memory_done` flag to `transmit_source`, as was done in `transmit_body_chunk`.

That we we can avoid infinite streams in re-direct as well.

CYBAI:
>     1. Reset the `in_memory_done` flag for the cloned handler as part of `re_extract`.
> 
>     2. Add similar logic with the `in_memory_done` flag to `transmit_source`, as was done in `transmit_body_chunk`.

With checking your PR, I can clearly understand what you mean here now :) Thanks a lot!

CYBAI:
@bors-servo try=wpt

- I tested the `tests/wpt/web-platform-tests/fetch/api/redirect/redirect-method.any.js` and it passed locally
- @gterzian THANKS A LOT! After your PR fixed the `read_from`, the patches in this PR is very straightforward to me and now I can fully understand the issue!
- Try to see if there's any other fixed / broken issues.

components/script/body.rs
```diff
@@ -127,10 +130,17 @@ impl TransmitBodyConnectHandler {
     /// in favor of making `re_extract` actually re-extract a stream from the source.
     /// See #26686
     fn transmit_source(&self) {
```
gterzian, line 132: I think we can make this take a `&mut self`, and then we would not need the cell.

components/script/body.rs
```diff
@@ -127,10 +130,17 @@ impl TransmitBodyConnectHandler {
...
+        // Reset `in_memory_done` because this means there are
+        // incoming streams and this is not done yet.
+        self.in_memory_done.set(false);
```
gterzian, line 135: We need to do this as part of https://github.com/servo/servo/blob/1d0c068ae6ec0ca33965976c78b271352861b758/components/script/body.rs#L104

because if we do it here then it will reset each time and will never be `true`.

On the other hand, here we need to call `stop_reading` and return early if it is already `true`.

components/script/body.rs
```diff
@@ -165,6 +181,9 @@ impl TransmitBodyConnectHandler {
         // In case of the data being in-memory, send everything in one chunk, by-passing SpiderMonkey.
         if let Some(bytes) = self.in_memory.clone() {
             let _ = bytes_sender.send(bytes);
+            // Mark this body as `done` so that we can correctly make it as Done in next tick
```
gterzian, line 184: nit: "so that we can stop reading in the next tick, matching the behavior of the promise-based flow."

gterzian (changes requested):
Ok great, few last comments left.

components/script/body.rs
CYBAI: ahh, yes! let me fix it! thanks! (but interestingly, current patch doesn't break any tests... 🤔)

components/script/body.rs
CYBAI: Okay!

actually I personally prefer not to use `mut` because it can help us make sure the _user_ who call this function won't mutate anything to it 👀 but I'm fine to change it! let me update!

components/script/body.rs
```diff
@@ -91,13 +92,18 @@ impl TransmitBodyConnectHandler {
...
         }
     }
 
     /// Re-extract the source to support streaming it again for a re-direct.
     /// TODO: actually re-extract the source, instead of just cloning data, to support Blob.
-    fn re_extract(&self, chunk_request_receiver: IpcReceiver<BodyChunkRequest>) {
+    fn re_extract(&mut self, chunk_request_receiver: IpcReceiver<BodyChunkRequest>) {
+        // Reset `in_memory_done` because this means there are
+        // incoming streams and this is not done yet.
+        self.in_memory_done = false;
```
gterzian, line 105: This is probably fine, since i don't think we will receive a chunk request following this, however it's strictly-speaking wrong, because what we want to do is leave `true` on the original, and only reset the clone to `false`. 

Maybe we need our own method that isn't `clone` to do this. We could also implement `Clone`, but I'm not sure if it's a good idea to have it include these side-effects.

gterzian (changes requested):
Ok, thanks for the latest fixes. There is still one point that we should address I think.

components/script/body.rs
CYBAI: Do you mean doing like this 👀? I don't have confidence that this is a good idea either though, if I understand correctly, this should be same as "leave `true` on the original route, and only reset the clone to `false`" logic 🤔?

```rust
impl Clone for TransmitBodyConnectHandler {
    fn clone(&self) -> Self {
        Self {
            stream: self.stream.clone(),
            task_source: self.task_source.clone(),
            canceller: self.canceller.clone(),
            bytes_sender: self.bytes_sender.clone(),
            control_sender: self.control_sender.clone(),
            in_memory: self.in_memory.clone(),
            source: self.source.clone(),
            in_memory_done: false,
        }
    }
}
```

I would like to cc @jdm to see his point of view of doing clone like this 🙋‍♂️

gterzian: I think you can just have a method called `reset_in_memory_done`, that would set it to false, and be called on the clone after the cloning.

CYBAI: hmm, I'm not sure if I get your point, because `RequestBody` only knows the IPC channel but doesn't know about the `in_memory_done` flag. If we'd like to add a method called `reset_in_memory_done`, does that mean we will also need a new IPC route (like `BodyChunkRequest::Reset`) to reset that flag? 🤔

gterzian: I think you can add the method to `TransmitBodyConnectHandler`, and then call it after it has been cloned.

CYBAI: ahhh, I see! because we basically need to handle the reset for redirection and we will call `Extract` for the redirection, then we can call the function right after `re_extract`.

https://github.com/servo/servo/blob/e928b3865e03f9ab7cd19ef23b2cf0aaeca3707b/components/script/body.rs#L117-L119

is that correct? (because we do the clone above it)

gterzian: https://github.com/servo/servo/blob/e928b3865e03f9ab7cd19ef23b2cf0aaeca3707b/components/script/body.rs#L118 is for nested re-redirections, the initial call is at https://github.com/servo/servo/blob/e928b3865e03f9ab7cd19ef23b2cf0aaeca3707b/components/script/body.rs#L371

CYBAI: I _might_ get your point now 🤔 so basically your point is, we should do the reset to the cloned body handler only instead of the original (a.k.a `self`) handler, is that correct?

Thanks for your patient >_<

gterzian: yes, that's it.

gterzian (changes requested):
Looks good, there's just one comment that can be made more effective I think.

components/script/body.rs
```diff
@@ -91,14 +92,22 @@ impl TransmitBodyConnectHandler {
             bytes_sender: None,
             control_sender,
             in_memory,
+            in_memory_done: false,
             source,
         }
     }
 
+    // Reset `in_memory_done` because this means there are
+    // incoming streams and this is not done yet.
```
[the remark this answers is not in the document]
CYBAI: ah, indeed! thanks!

components/script/body.rs
```diff
@@ -91,14 +92,22 @@ impl TransmitBodyConnectHandler {
             bytes_sender: None,
             control_sender,
             in_memory,
+            in_memory_done: false,
             source,
         }
     }
 
+    // Reset `in_memory_done`, called when a stream is
+    // re-extracted from the source to support a re-direct.
```
gterzian, line 101: nit: should be `///`

CYBAI: Good point! This should be a document 👍

gterzian (approved):
Great work! 

@bors-servo r+

CYBAI:
Ha, this reminds me homu doesn't support commands from review comment yet 😂

PR #26822, opened 2020-06-08
using cpp headers instead of c

### LINKED ISSUE #25986: Generate C++ headers for libsimpleservo
We [generate a C header](https://github.com/servo/servo/blob/b945de4ea910a6c6c8e296089bc3b2baf2afa248/ports/libsimpleservo/capi/build.rs#L20) when compiling libsimpleservo.

It would be better to use Cxx-type header file, at least to get class enumerators, instead of basic int-based enums. The current enum does not allow enums with the same name (`FOO:XXX` and `BAR:XXX` trigger a redefinition error).

We should create both files (simpleservo.h and simpleservo.hh). The difficult comes to the fact we  actually rely on int-to-enum transformation here: https://github.com/servo/servo/blob/b945de4ea910a6c6c8e296089bc3b2baf2afa248/support/hololens/ServoApp/ServoControl/Servo.cpp#L63 - Using an `IInspectable` in the C++ code and an `Object` in the idl code should be enough.

skrzyp1:
---

- [x] These changes fix #25986 (GitHub issue number if applicable)

### DIFF
--- a/ports/libsimpleservo/capi/src/lib.rs
+++ b/ports/libsimpleservo/capi/src/lib.rs
@@ -337,14 +337,14 @@ pub enum CContextMenuResult {
     Selected,
     // Can't use Dismissed. Already used by PromptResult. See:
     // https://github.com/servo/servo/issues/25986
-    Dismissed_,
+    Dismiss,
 }
 
 impl CContextMenuResult {
     pub fn convert(&self, idx: u32) -> ContextMenuResult {
         match self {
             CContextMenuResult::Ignored => ContextMenuResult::Ignored,
-            CContextMenuResult::Dismissed_ => ContextMenuResult::Dismissed,
+            CContextMenuResult::Dismiss => ContextMenuResult::Dismissed,
             CContextMenuResult::Selected => ContextMenuResult::Selected(idx as usize),
         }
     }
--- a/support/hololens/ServoApp/BrowserPage.cpp
+++ b/support/hololens/ServoApp/BrowserPage.cpp
@@ -59,17 +59,17 @@ void BrowserPage::BindServoEvents() {
       [=](hstring title, hstring artist, hstring album) {});
   servoControl().OnMediaSessionPlaybackStateChange(
       [=](const auto &, int state) {
-        if (state == servo::Servo::MediaSessionPlaybackState::None) {
+        if (state == static_cast<int>(servo::Servo::MediaSessionPlaybackState::None)) {
           mediaControls().Visibility(Visibility::Collapsed);
           return;
         }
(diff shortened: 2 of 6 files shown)


support/hololens/ServoApp/BrowserPage.cpp
```diff
@@ -59,17 +59,17 @@ void BrowserPage::BindServoEvents() {
       [=](hstring title, hstring artist, hstring album) {});
   servoControl().OnMediaSessionPlaybackStateChange(
       [=](const auto &, int state) {
-        if (state == servo::Servo::MediaSessionPlaybackState::None) {
+        if (state == static_cast<int>(servo::Servo::MediaSessionPlaybackState::None)) {
```
Manishearth, line 62: can we not pass in a `Servo::MediaSessionPlaybackState` instead of an `int` here?

ports/libsimpleservo/capi/src/lib.rs
```diff
@@ -337,14 +337,14 @@ pub enum CContextMenuResult {
     Selected,
     // Can't use Dismissed. Already used by PromptResult. See:
     // https://github.com/servo/servo/issues/25986
-    Dismissed_,
+    Dismiss,
```
Manishearth, line 340: will this break the C header?

skrzyp1: Break meaning change it? - yes:
```9c9
<   Dismissed_,
---
>   Dismiss,
```
I suppose this might be problem as some other projects probably use this header

support/hololens/ServoApp/BrowserPage.cpp
paulrouget: `OnMediaSessionPlaybackStateChange` is a IDL method. It's possible but not straight forward (using Inspectable objects).

ports/libsimpleservo/capi/src/lib.rs
paulrouget: Hmm… I didn't think this through, we use the C header in our Java binding… Would there be a way to use `Dismissed_` in the C header and `Dismissed` in the C++ header?

Manishearth: Break as in make it impossible to use because of the clash

support/hololens/ServoApp/BrowserPage.cpp
Manishearth: Ah, okay, seems fine then

mrobinson:
I think this change is no longer relevant as hololens support has been removed.

PR #26823, opened 2020-06-08
Update webrender

jdm:
These changes reflect changes in webrender's API that make RenderApiSender and RenderApi objects more challenging to share. This PR moves us to a model where:
* the compositor owns the main RenderApi object
* other threads that need to create transactions or manipulate fonts proxy those operations to the compositor (script/layout use IPC, while other threads use non-IPC channels)
* the webgl thread owns its own independent RenderApi

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -755,6 +755,7 @@ dependencies = [
 name = "compositing"
 version = "0.0.1"
 dependencies = [
+ "app_units",
  "canvas",
  "crossbeam-channel",
  "embedder_traits",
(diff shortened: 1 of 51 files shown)


asajeffrey:
Ouch. LGTM, though it's a pity we need to do this. @bors-servo r+

Cargo.lock
```diff
@@ -6427,7 +6423,7 @@ dependencies = [
  "num-traits",
  "plane-split",
  "rayon",
- "ron 0.1.7",
```
atouchet, line 6430: Looks like ron can be removed from servo-tidy.toml.

jdm:
Hold off on review. I need to figure out why a couple tests are suddenly failing.

jdm:
I'm going to accept a regression in functionality for mix-blend-mode and file a followup issue with what I've discovered from https://bugzilla.mozilla.org/show_bug.cgi?id=1635022.

PR #26837, opened 2020-06-09
Use the latest stable MixedReality runtime

jdm:
Per Microsoft, the preview runtime is now stable and we can depend on it without hidden environment variables.


CYBAI:
@bors-servo retry

- #25513

Issue #26665, opened 2020-05-26
Localize the UWP UI


jdm:
There are two parts to this work:
* Extract any hardcoded UI strings from XAML into strings files per https://docs.microsoft.com/en-us/windows/uwp/app-resources/localize-strings-ui-manifest
* Identify any UI strings present in Rust parts of the engine that need to be localized and choose a system for ensuring they are localized appropriately

Rather than building localization into the engine, if there are strings in Rust that should be localized, we should probably replace them with enum values that correspond with embedder strings, and modify the API to avoid string literals whenever possible.

paulrouget:
> Extract any hardcoded UI strings from XAML into strings files

#26840

> Identify any UI strings present in Rust parts of the engine that need to be localized and choose a system for ensuring they are localized appropriately

I only found strings in libsimpleservo/api. A few prompts and how the title is formatted.

Also some resources need to be translated. The API to request the resources to the embedder is already in place, but not implemented on the C side.

paulrouget:
```diff
diff --git a/ports/libsimpleservo/api/src/lib.rs b/ports/libsimpleservo/api/src/lib.rs
index 7dd1881243..7df706ba5e 100644
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -585,12 +585,14 @@ impl ServoGlue {
                     let fallback_title: String = if let Some(ref current_url) = self.current_url {
                         current_url.to_string()
                     } else {
+                        // FIXME:L10N
                         String::from("Untitled")
                     };
                     let title = match title {
                         Some(ref title) if title.len() > 0 => &**title,
                         _ => &fallback_title,
                     };
+                    // FIXME:L10N
                     let title = format!("{} - Servo", title);
                     self.callbacks.host_callbacks.on_title_changed(title);
                 },
@@ -702,9 +704,11 @@ impl ServoGlue {
                 EmbedderMsg::PromptPermission(prompt, sender) => {
                     let message = match prompt {
                         PermissionPrompt::Request(permission_name) => {
+                            // FIXME:L10N
                             format!("Do you want to grant permission for {:?}?", permission_name)
                         },
                         PermissionPrompt::Insecure(permission_name) => {
+                            // FIXME:L10N
                             format!(
                                 "The {:?} feature is only safe to use in secure context, but servo can't guarantee\n\
                                 that the current context is secure. Do you want to proceed and grant permission?",
[...]

PR #26840, opened 2020-06-10
UWP L10N

paulrouget:
Partially addresses #26665.

This does not cover Rust strings and resources files.


jdm (changes requested):
Thanks for doing this!

paulrouget:
Addressed @jdm comments.

@bors-servo r=Manishearth

PR #26852, opened 2020-06-10
Some more boring stuff to change how we compute content sizes

nox:


components/layout_2020/positioned.rs
```diff
@@ -22,7 +22,7 @@ use style::Zero;
 
 #[derive(Debug, Serialize)]
 pub(crate) struct AbsolutelyPositionedBox {
-    pub contents: IndependentFormattingContext,
```
SimonSapin, line 25: I’m not thrilled about this one (`layout_context` and `style_content` are groupings of transient data used "on the side" during layout, rather than the "object" of what we’re doing), but I don’t have a better suggestion.

components/layout_2020/style_ext.rs
```diff
@@ -335,6 +347,87 @@ impl ComputedValuesExt for ComputedValues {
         )
     }
 
+    fn outer_inline(
```
SimonSapin, line 350: This move feels conceptually wrong to me. On the method receiver type: styles don’t have content sizes, boxes do. On the module location (which is independent): this code belongs in the `sizing` module.

components/layout_2020/formatting_contexts.rs
```diff
@@ -20,44 +20,43 @@ use style::values::specified::text::TextDecorationLine;
 
 /// https://drafts.csswg.org/css-display/#independent-formatting-context
 #[derive(Debug, Serialize)]
-pub(crate) struct IndependentFormattingContext {
+pub(crate) enum IndependentFormattingContext {
```
SimonSapin, line 23: What do we gain with this change? In particular pushing common fields like `style` further "down" into enum variants.

By the way, in a sizing model as described in https://github.com/w3c/csswg-drafts/issues/5190#issuecomment-643329880 where the parent provides an available size on both axes and the child return its used size of both axis, I think we may be able to unify replace content with other kinds of independent formatting contexts and make it so that callers don’t need to handle them separately.

components/layout_2020/style_ext.rs
nox: That method living on `BoxContentSizes` look even more wrong to me given it seldomly does anything with `self` there. "There are styles, and those styles can clamp an inline size, or a content size if there are no inline size" makes *way* more sense to me and better compartmentalize the code.

nox: And given you can't split a trait in multiple parts, this needs to be in `style_ext`.

components/layout_2020/formatting_contexts.rs
nox: It lets me write a method on `NonReplacedFormattingContext` that will use `StyleExt::outer_inline_and_percentages` and compute the content size on demand. I can't do that without splitting the two kinds of contents given the computation isn't done at all for replaced content.

nox:
Fundamentally, I think it's wrong for any content-size-related code to be a method with many movable parts being passed as arguments, specifically I don't think it's good design for it to be a method on the content size type that can represent an absence of content size (in the case of explicit inline size), in a way where you can technically pass unrelated styles.

components/layout_2020/formatting_contexts.rs
nox: If I didn't do that, I would need to change `NonReplacedIFC` to also have a reference to the style field of `IndependentFormattingContext`, or have a method on `NonReplacedIFC` that takes a style argument, which seems way more confusing to me.

components/layout_2020/style_ext.rs
nox: To further explain what I mean: styles don't have content sizes, but they definitely are what owns the size extremums, and the box-sizing property, and often a non-auto inline size. When there is a non-auto inline size, the `BoxContentSizes` value isn't even looked at. That means to me that it shouldn't be a method on `BoxContentSizes`.

SimonSapin: How about making it a free function in the `sizing` module? `style_ext` already feels like a kitchen sink to me.

nox: That is totally fair, will do that, absolutely no damn clue how I didn't think of that on my own.

jdm:
@bors-servo retry
* #25522

PR #26875, opened 2020-06-11
Don't pretend we support SharedArrayBuffer.

### LINKED ISSUE #26874: Hide SharedArrayBuffer
Right now we expose the SharedArrayBuffer interface through SpiderMonkey, but we have no code that supports passing it to DOM APIs or transferring it between threads. We should probably hide it instead or exposing something that isn't actually useful.

jdm:
Fixes #26874.

### DIFF
--- a/tests/wpt/metadata/xhr/send-data-sharedarraybuffer.any.js.ini
+++ b/tests/wpt/metadata/xhr/send-data-sharedarraybuffer.any.js.ini
@@ -1,5 +1,47 @@
 [send-data-sharedarraybuffer.any.html]
-  expected: CRASH
+  [sending a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Int8Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Uint8Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Uint8ClampedArray backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Int16Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Uint16Array backed by a SharedArrayBuffer]
+    expected: FAIL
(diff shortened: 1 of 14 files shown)


gterzian:
Perhaps it's a good idea to also remove https://github.com/servo/servo/blob/33e96e95677c5c1a5d49fece4016b7acf1d7594b/components/script/script_runtime.rs#L460

So that we have a single commit turning off everything(making it easier to know what is required to turn it back on), and also maybe leaving the "can block" stuff while turning of SAB can lead to weirdness.

tests/wpt/metadata/xhr/send-data-sharedarraybuffer.any.js.ini
```diff
@@ -1,5 +1,47 @@
...
+  [sending a Int32Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Uint32Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Float32Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a Float64Array backed by a SharedArrayBuffer]
+    expected: FAIL
+  [sending a DataView backed by a SharedArrayBuffer]
+    expected: FAIL
```
gterzian, line 47: nit: do we always want to have a newline at the end of `.ini` files? If only to not have github show the warning in the diff?

gterzian (approved):
thanks for updating all those test expectations!

tests/wpt/metadata/xhr/send-data-sharedarraybuffer.any.js.ini
jdm: I'm not convinced it's a useful thing to care about.

gterzian: ok!

PR #26881, opened 2020-06-12
Fix re-extracting stream upon re-direct

gterzian:
This fixes the logic in `net` regarding re-extracting a stream from the request body as part of a re-direct. 

The problem was that the `read_from` flag was never read as `true`, because it would be set on the clone of the original request used in the http fech, and the original request would then be used for the re-direct. So the original request never had a `read_from` flag set to true. 

Also, this removes the entire `read_from` concept, and instead uses a spec-compliant hook to re-extract the stream, which doesn't require tracking whether the stream has been read from in another http fetch. 

This will unblock https://github.com/servo/servo/pull/26810. I don't expect test changes in this one, although I have verified locally(using the other branch), that this fixes the issue that has emerged as part of that PR. 

---

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -911,8 +911,8 @@ pub fn http_redirect_fetch(
     }
 
     // Step 12
-    if let Some(_) = request.body {
-        // TODO: extract request's body's source
+    if let Some(body) = request.body.as_mut() {
+        body.extract_source();
     }
 
     // Step 13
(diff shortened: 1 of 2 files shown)


jdm (approved):
Needs a rebase.

gterzian:
@jdm how wonderful that you have been assigned randomly, I was actually planning on asking you to review this... r?

CYBAI:
@gterzian thanks! this is quite simple and if I understand correctly, this means we've handled the `extract` properly instead of trying to use the `read_from` flag!

gterzian:
yes indeed.

PR #26906, opened 2020-06-13
Fix streaming request bodies

### LINKED ISSUE #26904: Make streaming upload work.
I'm surprised this test is still failing: `fetch/api/basic/request-upload.any.js`, I would have expected it to pass after https://github.com/servo/servo/pull/25873

gterzian:
FIX #26904

At this point I'm not sure if the hyper `Body::wrap_stream` integration by way of the IPC route is broken, or if perhaps there is a problem with the WPT server not handling the requests properly, as I am getting a bunch of:

```
 0:04.54 INFO STDERR: 127.0.0.1 - - [13/Jun/2020 18:22:08] code 400, message Bad request syntax ('4')
 0:04.54 INFO STDERR: 127.0.0.1 - - [13/Jun/2020 18:22:08] "4" 400 -
```

In any case, the `start_reading` call in `body.rs` had to be put into the other function, since it should not be called for each chunk(this was actually failing before). 

I've also added the chunked header in case the source is a stream.

Last thing to figure out, why is the request data not actually reaching the server, or not being handled properly by it. 

---

### DIFF
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -23,7 +23,8 @@ use net_traits::request::{
     is_cors_safelisted_method, is_cors_safelisted_request_header, Origin, ResponseTainting, Window,
 };
 use net_traits::request::{
-    BodyChunkRequest, CredentialsMode, Destination, Referrer, Request, RequestMode,
+    BodyChunkRequest, BodyChunkResponse, CredentialsMode, Destination, Referrer, Request,
+    RequestMode,
 };
 use net_traits::response::{Response, ResponseBody, ResponseType};
 use net_traits::{FetchTaskTarget, NetworkError, ReferrerPolicy, ResourceFetchTiming};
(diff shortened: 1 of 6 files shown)


gterzian:
So I have:

1. Removed the use of `Body::wrap_stream`, because it wasn't working. I'm not sure if this was a client or server-side issue, see also https://github.com/web-platform-tests/wpt/issues/24139
2. Ensure we call `start_reading` on the stream only when the route is first connected to.
3. added a mechanism to propagate read errors in script to a network error in net. 

So this makes the relevant tests pass.

@Manishearth r?

gterzian:
Ok so the problem is now that I've switched the `start_reading` operation to read on the stream, this happens multiple times in case of a re-direct, so it fails. 

However, I just need to add a conditional to this based on the source, so that it only happens when we're using an actual `ReadableStream`, and by definition in these cases there will never be a re-direct.

gterzian:
Actually I think I'll keep the chunking for when it's a stream body, what we need to do is fix the WPT server, see https://github.com/web-platform-tests/wpt/issues/24139

gterzian:
Failure looks like https://github.com/servo/servo/issues/23290

gterzian:
@Manishearth r?

So the one test is still failing, because the WPT server doesn't support chunked requests(https://github.com/web-platform-tests/wpt/issues/24139), also this isn't a feature actually implemented anywhere yet. 

Re the non-stream case, I've added some explicit buffering so that it matches the previous behavior when the body of a request was a `Vec<u8>`, note that I think hyper could be relied on to do this anyway, however I'd rather be sure and explicit.

Manishearth:
@bors-servo r+

this looks great, thanks!

PR #26909, opened 2020-06-13
Implement visibility for layout_2020

### LINKED ISSUE #26841: Implement `visibility: hidden` in Layout 2020
Relevant spec: https://drafts.csswg.org/css2/visufx.html#propdef-visibility

Relevant code: https://github.com/servo/servo/blob/master/components/layout_2020/display_list/mod.rs

Changes needed: when generating display items for a `BoxFragment` or `TextFragment`, check the value `visibility` property in its `style` or `parent_style` field respectively:

```rust
use style::properties::ComputedValues;
use style::properties::longhands::visibility::computed_value::T as Visibility;

fn display_items_are_visible(style: &ComputedValues) -> bool {
    let visible = match style.get_inherited_box().visibility {
        Visibility::Visible => true,
        Visibility::Hidden => false,

        // `visibility: collapse` is disabled at compile time
        // in `components/style/properties/longhands/inherited_box.mako.rs`
        // at the moment, so the variant doesn’t exist.
        // When it is enabled, this code should treat it like `hidden`
        // Visibility::Collapse => false,
    }
}
```

For non-visible fragments, skip emitting display items for backgrounds, borders, or text. Don’t inhibit rendering of descendant fragments: CSS inheritance already propagates the property to descendant elements, and they can override it to be visible again.

See https://github.com/servo/servo/wiki/Layout-2020#using for building with Layout 2020 enabled, to check that your changes compile.

MDeiml:
Implement the 'visibility: hidden' (and 'visibility: collapse') css properties.

---

- [X] These changes fix #26841 (GitHub issue number if applicable)

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -16,6 +16,7 @@ use net_traits::image_cache::UsePlaceholder;
 use std::sync::Arc;
 use style::computed_values::text_decoration_style::T as ComputedTextDecorationStyle;
 use style::dom::OpaqueNode;
+use style::properties::longhands::visibility::computed_value::T as Visibility;
 use style::properties::ComputedValues;
 use style::values::computed::{BorderStyle, Length, LengthPercentage};
 use style::values::specified::text::TextDecorationLine;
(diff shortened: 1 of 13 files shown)


MDeiml:
These tests need to be updated (using './mach update-wpt'), but I'm not sure how I'm supposed to do that. If I just use the log from the tests it also disables the intermittents which have nothing to do with this change. That can't be right?

SimonSapin:
Hi @MDeiml. Your patch looks good.

It also disables hit testing (which answers questions like "what elements are under this mouse cursor?") for hidden elements and I wasn’t sure that it should. Unfortunately we don’t have a specification for hit testing, so the next best thing is looking at other engines. And indeed, `visibility: hidden` seems to inhibit `:hover`: http://software.hixie.ch/utilities/js/live-dom-viewer/saved/8203

For updating expected results, `update-wpt` needs to be fed the the log files created by `test-wpt --log-raw`. On CI those are named `test-wpt.log` and published as "artifacts". When viewing a log file in the web interface, the "three dots" menu at the bottom left has a "Raw log" link. Copying those URLs for the two tasks gives:

```
curl -L \
    https://community-tc.services.mozilla.com/api/queue/v1/task/Uxg8re09Q66y5Aw4y31tnw/runs/0/artifacts/public/test-wpt.log \
    https://community-tc.services.mozilla.com/api/queue/v1/task/E_GKbQc3RcmHeidKPLdjmw/runs/0/artifacts/public/logs/live.log \
> /tmp/servo.log
./mach update-wpt --layout-2020 /tmp/servo.log
git status --short
```
```
 D tests/wpt/metadata-layout-2020/css/CSS2/positioning/position-relative-032.xht.ini
 D tests/wpt/metadata-layout-2020/css/CSS2/visufx/visibility-005.xht.ini
 D tests/wpt/metadata-layout-2020/css/CSS2/visufx/visibility-block-001.xht.ini
 M tests/wpt/metadata-layout-2020/css/css-transitions/properties-value-001.html.ini
 M tests/wpt/metadata-layout-2020/css/css-transitions/properties-value-inherit-001.html.ini
 M tests/wpt/metadata-layout-2020/css/css-transitions/properties-value-inherit-002.html.ini
 M tests/wpt/metadata-layout-2020/css/cssom-view/MediaQueryList-addListener-removeListener.html.ini
 M tests/wpt/metadata-layout-2020/css/cssom-view/elementFromPosition.html.ini
 M tests/wpt/metadata-layout-2020/css/cssom-view/elementsFromPoint-iframes.html.ini
 M tests/wpt/metadata-layout-2020/css/cssom/serialize-values.html.ini
 D tests/wpt/mozilla/meta-layout-2020/css/visibility_hidden.html.ini
?? tests/wpt/metadata-layout-2020/css/CSS2/floats/hit-test-floats-004.html.ini
?? tests/wpt/metadata-layout-2020/css/css-transitions/no-transition-from-ua-to-blocking-stylesheet.html.ini
?? tests/wpt/metadata-layout-2020/css/cssom-view/elementFromPoint-001.html.ini
```

For tests already marked as intermittent in a way it doesn’t matter because CI ignores their result for the purpose of deciding wether to merge a PR. Still, to make a PR cleaner we can look at `intermittents.log` artifacts and manually revert some of the changes made by `update-wpt`. The tooling there is not great, sorry!

… and by the time I typed this up it looks like you’ve already managed :)

@bors-servo r+

MDeiml:
Thanks for explaining it anyways

SimonSapin:
Uh it looks like editing my comment that contains the "r+" command made Homu/Bors interpret that as a new command, which is arguably a bug, and then despite commenting "no need to approve again" it made and pushed a second merge commit triggering a second CI run, which is another bug.

SimonSapin:
Ah this patch also adds `visibility: collapse` to Layout 2013 so some tests started passing there and those expectations also need updating

Issue #26569, opened 2020-05-19
Anchor elements and image elements do not include a correct referrer URL


jdm:
https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/dom/htmlanchorelement.rs#L653
https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/dom/htmlimageelement.rs#L306-L309

atouchet:
Did #26926 fix this?

jdm:
Looks like it!

PR #26926, opened 2020-06-15
Make url for "client" referrer mandatory

### LINKED ISSUE #26570: DOM fetches do not include referrer URL
It's hard to follow the logic of what happens - we call Request::Constructor to construct a request object when only a fetch URL is provided, then we make a [request](https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/dom/request.rs#L107) which has a default referrer/referrer-policy. We then initialize [another new net_traits::Request](https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/dom/request.rs#L148-L156) with the referrer/policy of the first one, and then we sometimes [default to the Client referrer](https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/dom/request.rs#L209). Finally we [initialize](https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/script/fetch.rs#L122-L123) the referrer/policy of the RequestBuilder with whatever values were eventually selected, and that's what the network stack ends up dealing with.

I suspect in most cases we end up in [this code](https://github.com/servo/servo/blob/9cbebe29077efbb8b22631f27facc0b548faf5a3/components/net/fetch/methods.rs#L239-L243) and send no Referer header at all.

MDeiml:
I added a url attribute to `Referrer::Client` so that the referrer header can be set accordingly when fetching.

`Referrer::Client` has to be kept separate from `Referrer::ReferrerUrl` as they differ in this method
https://github.com/servo/servo/blob/6b0d9afd6fdc28356ad44af0104ddd25a7b6438d/components/script/dom/request.rs#L566-L576

---

- [X] These changes fix #26570 (GitHub issue number if applicable)


MDeiml:
I know this is quite a lot of changes but most of it is just replacing `Referrer::Client` with `global_scope.get_referrer()`. Maybe it would still be best to split this up into smaller changes?

Also I didn't add tests yet because I don't know what the appropriate place to do so would be.

jdm:
Alright, now we're talking! You're going to want to run `./mach test-wpt tests/wpt/web-platform-tests/referrer-policy /html/semantics/scripting-1/the-script-element/module/ tests/wpt/web-platform-tests/fetch/api /html/semantics/embedded-content/the-img-element/image-loading-lazy-referrerpolicy-change.sub.html tests/wpt/web-platform-tests/xhr --log-raw wpt.log` and then `./mach update-wpt wpt.log` when that's complete.

MDeiml:
Some tests that previously passed no failed, do I need to check those all individually?

components/script/dom/xmlhttprequest.rs
```diff
@@ -679,28 +679,29 @@ impl XMLHttpRequestMethods for XMLHttpRequest {
...
-                    .map(|referrer_url| Referrer::ReferrerUrl(referrer_url)),
-            )
-            .referrer_policy(self.referrer_policy.clone())
-            .pipeline_id(Some(self.global().pipeline_id()));
+        let mut request = RequestBuilder::new(
+            self.request_url.borrow().clone().unwrap(),
+            self.referrer_url
+                .clone()
+                .map(|url| Referrer::ReferrerUrl(url))
+                .unwrap_or_else(|| self.global().get_referrer()),
```
jdm, line 687: Let's remove the Option around referrer_url and update [this code](https://github.com/servo/servo/blob/70af0bf14286cadc8f5b6e666131ba1c106331c1/components/script/dom/xmlhttprequest.rs#L171-L176) to use global.get_referrer directly.

components/script/script_module.rs
```diff
@@ -731,7 +733,7 @@ impl ModuleTree {
                         url.clone(),
                         visited_urls.clone(),
                         destination.clone(),
-                        Referrer::Client,
+                        referrer.clone(),
```
jdm, line 736: We can remove this argument from fetch_single_module_script and use global.get_referrer() in that code directly instead.

jdm (changes requested):
These changes look like a great improvement!

components/script/dom/xmlhttprequest.rs
MDeiml: I wasn't sure if I could change this struct since it is a dom struct and (in my understanding) can get passed to JS. Is this partially true and some fields should not be edited?

jdm: The contents of the struct are never exposed directly to JS. JS calls methods in the XMLHttpRequestMethods trait, so we can change the struct in any way we want that allows those methods to continue returning expected values.

jdm:
This PR will need to be rebased against a recent upstream revision as well.

jdm:
One more `./mach fmt` needed, and then this is ready to merge.

jdm:
One more test result needs updating:
```
  └ PASS [expected FAIL] Sandboxed iframe with tuple origin sends referrers.
  │ FAIL [expected PASS] Sandboxed iframe with opaque origin doesn't send referrers.
  │   → assert_equals: expected (undefined) undefined but got (string) "http://web-platform.test:8000/referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html"
  │ 
  │ testSandboxedIframe/</<@http://web-platform.test:8000/referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html:19:28
  └ Test.prototype.step_func/<@http://web-platform.test:8000/resources/testharness.js:2002:35
```

MDeiml:
Hm. When I run this test locally it has no unexpected result. Is there something wrong with my setup?

jdm:
It might be the difference between a debug and release build, perhaps? It might also be an intermittent result.

@bors-servo retry

jdm:
The failure appears to be consistent. I recommend updating the appropriate ini file manually.

jdm:
@bors-servo retry
* #24622

MDeiml:
Oops, you were faster

MDeiml:
This probably also solves #26569

PR #26942, opened 2020-06-16
Combine AnimationAndTransitionDeclarations and AnimationRules

mrobinson:
These two structs are very similar, so we can combine them.

---


emilio (approved):
Looks great, thank you!

PR #26952, opened 2020-06-17
[UWP] Basic console panel

### LINKED ISSUE #26850: UWP: Implement a console panel (devtools)
Now we have a toolbox, we can implement a console.

I think there's 2 different possible approaches:
- implement an embedding API set of methods to retrieve the console messages
- implement a devtools protocol client and connect through the devtools port

For a start, we probably don't need to implement a JS REPL. Just listing the logs would be good enough.

paulrouget:
This PR introduces a Devtools client module to the UWP app. And a basic console panel.

There are still bugs and missing features, that I will address in future PRs:

- ~~the output is just some text. No colors or formatting. This will be implemented as a [gridview](https://docs.microsoft.com/en-us/dotnet/framework/wpf/controls/gridview-overview) later.~~
- ~~When connecting to the devtools server, we get a prompt asking for permission. I see 2 options: we share the socket between Servo and the UWP app (no idea if that's actually possible) or we keep using the TCP transport and share a token between Servo and the UWP app that would allow to bypass the permission prompt~~
- ~~add a JS input.~~
- show HTTP errors. For now they are just ignored.
- ~~There are 3 panels now: console, preference and a random "devtool server" panel that shows the devtools port. We could just print the port number in the console instead.~~
- ~~Servo crashes if we open the devtools too early. This needs investigation~~.

Also, now that we have the foundation of a devtools client in UWP, we should remove the recently introduced preference API and rely on the preference devtools actor instead.

Fix #26850

### DIFF
--- a/support/hololens/ServoApp/BrowserPage.cpp
+++ b/support/hololens/ServoApp/BrowserPage.cpp
@@ -7,11 +7,7 @@
 #include "BrowserPage.h"
 #include "BrowserPage.g.cpp"
 #include "DefaultUrl.h"
-
-#include "winrt/Microsoft.UI.Xaml.Controls.h"
-#include "winrt/Microsoft.UI.Xaml.XamlTypeInfo.h"
-#include "winrt/Windows.UI.Text.h"
-#include "winrt/Windows.UI.Xaml.Documents.h" // For Run.Text()
+#include "Devtools/Client.h"
 
 using namespace std::placeholders;
 using namespace winrt::Windows::Foundation;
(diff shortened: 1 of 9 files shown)


jdm (approved):
Greta work! I was worried the JSON code in C++ would be complex, but the client code is actually surprisingly readable!

jdm:
```
support/hololens/ServoApp/Devtools/Client.cpp is not formatted correctly.
```

jdm:
@raananw As of today's nightly, you can now pop open a panel in Firefox Reality that shows the console messages and JS exceptions for the current page. It's the default view when clicking the wrench button.

PR #26953, opened 2020-06-17
Cancel animations for nodes which are removed from the DOM

mrobinson:
This includes nodes which are being reparented.

---

### DIFF
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -281,6 +281,11 @@ impl Node {
         }
     }
 
+    pub fn clean_up_layout_data(&self) {
+        self.owner_doc().cancel_animations_for_node(self);
+        self.style_and_layout_data.borrow_mut().take();
+    }
+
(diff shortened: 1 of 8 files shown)


components/script/dom/node.rs
```diff
@@ -281,6 +281,16 @@ impl Node {
         }
     }
 
+    #[allow(unsafe_code)]
+    pub fn clean_up_layout_data(&self) {
+        self.owner_doc().cancel_animations_for_node(self);
+        unsafe {
+            self.style_and_layout_data
+                .borrow_for_script_deallocation()
```
jdm, line 289: This should be borrow_mut, right? Then there's no need for unsafe.

jdm (changes requested):

components/script/dom/node.rs
mrobinson: That's right. I've pushed a new version of the branch that uses `borrow_mut()` as well as fixing some test expectations.

PR #26970, opened 2020-06-18
animations: Don't apply animation delay to every iteration

mrobinson:
The delay should only be applied to the animation before the first
iteration.

---


mrobinson:
I've added a Servo-specific test to this change because delays are very difficult to test without our special animation ticking API. Most of the delay tests in WPT are manual for this reason.

jdm:
@bors-servo retry
* #26979

jdm:
@bors-servo retry
* #26980

Issue #26895, opened 2020-06-12
Intermittent panic in http_loader::test_origin_set


jdm:
```
  10: std::panicking::rust_panic_with_hook
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\/src\libstd\panicking.rs:477
  11: std::panicking::begin_panic_handler
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\/src\libstd\panicking.rs:385
  12: core::panicking::panic_fmt
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\/src\libcore\panicking.rs:89
  13: core::panicking::panic
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\/src\libcore\panicking.rs:52
  14: core::option::Option<alloc::boxed::Box<net_traits::response::Response>>::unwrap<alloc::boxed::Box<net_traits::response::Response>>
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\src\libcore\macros\mod.rs:10
  15: main::http_loader::test_origin_set
             at .\tests\http_loader.rs:1437
  16: main::http_loader::test_origin_set::{{closure}}
             at .\tests\http_loader.rs:1375
```

jdm:
https://github.com/servo/servo/blob/37394a892c79d6edbef868d6451710648669cc1c/components/net/tests/http_loader.rs#L1438-L1444 is failing. A more recent rust compiler might be able to provide a more specific failing unwrap call, since https://github.com/rust-lang/rust/pull/73182 merged recently.

gterzian:
So it looks like the response is none:

```

  13: core::panicking::panic
             at /rustc/a74d1862d4d87a56244958416fd05976c58ca1a8\/src\libcore\panicking.rs:52
  14: core::option::Option<alloc::boxed::Box<net_traits::response::Response>>::unwrap<alloc::boxed::Box<net_traits::response::Response>>
```
https://community-tc.services.mozilla.com/tasks/ennyDdgxQ_2ovzskg4Nzvg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FennyDdgxQ_2ovzskg4Nzvg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive_backing.log#L2587

jdm:
```
[...]

PR #26984, opened 2020-06-19
Add GPUSampler and GPUTextureView to BindingResources and update wgpu-core

kunalmohan:
This also completes validation for `GPUBindGroup` and `GPUBindGroupLayout`.
`entry_map` is stored in `GPUBindGroupLayout` for validating `createBindGroup()`.

r?@kvark

---

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -6,11 +6,14 @@
 
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::GPUAdapterBinding::GPULimits;
-use crate::dom::bindings::codegen::Bindings::GPUBindGroupBinding::GPUBindGroupDescriptor;
+use crate::dom::bindings::codegen::Bindings::GPUBindGroupBinding::{
+    GPUBindGroupDescriptor, GPUBindGroupEntry, GPUBindingResource, GPUBufferBindings,
+};
 use crate::dom::bindings::codegen::Bindings::GPUBindGroupLayoutBinding::{
     GPUBindGroupLayoutDescriptor, GPUBindGroupLayoutEntry, GPUBindingType,
 };
 use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUBufferDescriptor;
+use crate::dom::bindings::codegen::Bindings::GPUBufferUsageBinding::GPUBufferUsageConstants;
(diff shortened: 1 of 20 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -153,6 +155,86 @@ impl GPUDevice {
             )
         }
     }
+
+    fn validate_texture_view_binding(
```
kvark, line 159: All the validation logic is common to the applications using wgpu. It needs to be implemented in `wgpu-core` instead of here. The idea is that you'd call wgpu-core API and get `Result`, from which you'd figure out how to throw an error according to the WebGPU error model.
We've started converting the methods to `Result`. As such, `create_bind_group_layout` returns one, as so do pipeline creation methods.

Therefore, I think the logic here needs to be removed.

components/script/dom/gpudevice.rs
```diff
@@ -1094,3 +1288,46 @@ fn convert_texture_size_to_wgt(size: &GPUExtent3DDict) -> wgt::Extent3d {
         depth: size.depth,
     }
 }
+
+fn get_component_from_format(format: GPUTextureFormat) -> GPUTextureComponentType {
```
kvark, line 1292: We [have this](https://github.com/gfx-rs/wgpu/blob/d7d0fd5dfe9382b596ca4f1eff12663691603ca5/wgpu-types/src/lib.rs#L1589), please use it instead :)

kvark (changes requested):
This needs to do less validation :)
Sorry for not making this clear originally!
The only responsibilities of Servo side would be:
  - managing IDs
  - managing IPC communication
  - converting between DOM WebIDL types/calls and wgpu-core types
  - converting wgpu-core errors to the WebGPU error model

There should be no need to do any validation on Servo side.
Moreover, even if you do the validation in a way it's written here - on the client side, we can't rely on it on the server side, because the content process may be compromised. Therefore, most if not all validation needs to happen on the server side, and we want it in wgpu-core, accessible to all.

kunalmohan:
I have removed validation for `GPUBindGroup`, `GPUBindGroupLayout` and `GPUPipelineLayout`. The only thing that we check now is whether a related resource is in a valid state or not, e.g. while creating BG whether the associated BGL is in a valid state or not.

kvark (approved):
Thank you!

jdm:
@bors-servo retry
* #26895 #27038 #24622

PR #27005, opened 2020-06-20
Do not raise StopIteration: PEP 479

saschanaz:
This fixes `test-tidy` Py3 compatibility, as [PEP 479](https://www.python.org/dev/peps/pep-0479/) says `return` must be used to stop iteration inside generator instead of `raise StopIteration`.

This introduces subtle behavior change where `FileList()` constructor now won't implicitly stop the caller generator when the list is empty. `wpt_lint.py` is modified to explicitly stop when empty, to match the change.

---


CYBAI:
@bors-servo retry

- #24622

PR #27008, opened 2020-06-20
Require clang-format 10

saschanaz:
LLVM 10 is here, probably time to use it.

---


jdm:
We're still only using LLVM 9: https://github.com/servo/servo/blob/master/python/servo/packages.py#L7

saschanaz:
Oops. can we use 10? Is there a blocker?

jdm:
We haven't uploaded the binaries for LLVM 10 to the s3 bucket yet.

PR #27016, opened 2020-06-22
Ensure clean shutdown of JS threads 

### LINKED ISSUE #9672: Infinite loop in Javascript will cause the servo to hang when trying to close it.
1) ./mach run http://people.mozilla.org/~nhirata/html_tp/OOMCrash.html
2) click the button
3) close servo

Expected: servo will close
Actual: servo process will hang

### LINKED ISSUE #26685: Mac OS: A busy main JS thread prevents closing browser window
In the light of https://github.com/servo/servo/pull/25873#issuecomment-635093084 it appears that if the main JS thread of a window is blocking(like it does in a sync XHR for example, but it guess any script could just keep it busy somehow), then when you click to close the window, the browser appear to be hanging. 

I think this is because the browser is waiting on some event to be handled by the JS, and I think it might be better if the browser would remain responsive, and close the window, perhaps after a short timeout if the JS appears unresponsive.

### LINKED ISSUE #26996: Win64 - Unable to close Servo window
The close button does not work

### LINKED ISSUE #27027: Use BHM to terminate hanging JS threads when closing window
Follow-up on https://github.com/servo/servo/pull/27016

Since the constellation has a channel available to the BHM(s), when exiting a pipeline fails due to a hanging `ScriptThread` because of long-running JS, we could send a message to the BHM(there is either one, or one per content-process), so that it can terminate the thread(or even the whole process), perhaps after having sampled it. 

I think this is especially important in multiprocess mode, since otherwise JS could simply keep running in a separate process after the main browser process would have exited, and there would not be a way for the user to know about it unless it would inspect running processes in some way(?). 

This would have to be done in a per-platform basis, for example for Mac it seems the `thread_terminate` mach function could be used. 

See `components/background_hang_monitor`.

gterzian:
FIX https://github.com/servo/servo/issues/26685
FIX https://github.com/servo/servo/issues/26996
FIX https://github.com/servo/servo/issues/9672
FIX #27027

---


components/constellation/constellation.rs
```diff
@@ -954,12 +955,37 @@ where
...
-        while !self.shutting_down || !self.pipelines.is_empty() {
+        while !self.shutting_down {
             // Randomly close a pipeline if --random-pipeline-closure-probability is set
             // This is for testing the hardening of the constellation.
             self.maybe_close_random_pipeline();
             self.handle_request();
         }
+
+        // Try to cleanly exit all pipelines, until a timeout hits.
+        let pipeline_exit_timeout = after(Duration::from_millis(200));
```
paulrouget, line 966: If possible, can you make that a pref?

components/constellation/constellation.rs
```diff
@@ -954,12 +955,37 @@ where
...
+        loop {
+            if self.pipelines.is_empty() {
+                break;
+            }
```
paulrouget, line 970: This loop + if could be turned into a `while`.

components/constellation/constellation.rs
```diff
@@ -954,12 +955,37 @@ where
...
+                        Ok((source_pipeline_id, FromScriptMsg::PipelineExited)) => {
+                            self.handle_pipeline_exited(source_pipeline_id);
+                        },
+                        _ => continue,
+                    }
+                }
+                recv(pipeline_exit_timeout) -> _ => {
+                    if !self.pipelines.is_empty() {
+                        warn!("Failed to exit a number of pipelines {:?}", self.pipelines.len());
+                    }
```
paulrouget, line 983: Is there any way to get more information about why it's blocked?

paulrouget: I'm afraid we will have that warning and never pay attention to it.

Maybe make it an error, or dump a stack, not sure…

gterzian: I tried sampling them, but it's a bit messy because the bhm messaging stuff is not really setup for that, I think it could potentially introduce a new hang since we don't match bhm senders to pipeline yet, so perhaps best done as a follow-up. 

Also someone can use the dedicated sampling API to do this manually in some way, the current changes are more aimed at ensuring the window actually closes when someone wants to close it. 

Also this seems to mostly happen when JS is running and blocking the script-thread from handling the exit messages, so I think what we really need is a hook to kill the script-thread, see https://github.com/servo/servo/issues/27027

gterzian:
@paulrouget r?

gterzian:
Ok actually I will do this "properly" as per #27027, and remove the timeout since that doesn't actually solve the underlying problem...

paulrouget:
This is amazing! r=me, but I'd like @jdm to also take quick look, as some changes fly above my head (in particular I don't know about the script runtime changes).

Also - how I'd like to point out that we had this code for a while that we might wan to remove: https://github.com/servo/servo/blob/afb8be50a50127258e8abe54c99ef2095af97040/ports/winit/platform/macos/mod.rs#L16-L36

gterzian:
> Also - how I'd like to point out, as we're cleaning up shutdown, that we had this code that we might wan to remove: 

Unfortunately even with this change there are still lots of threads still running after shutdown. I do think it keeps going down :) (although maybe not with this change, since the constellation would block until all pipelines where indeed shut-down).

gterzian:
I'm going to do one last thing, which is to make the script-thread check the`closing` flag before running a task, because I noticed you can still hang the thread if you queue a bunch of tasks where each tasks can potentially hang the thread, because the interrupt is called only once and the script-thread then goes back to running it's event-loop, if that event-loop has more tasks in the queue they will execute, and the thread goes back to hanging(with the BHM not trying further interrupts). 

So I think the easy way to prevent that is to check the flag in the loop running the tasks.

components/background_hang_monitor/background_hang_monitor.rs
```diff
@@ -112,6 +125,7 @@ impl BackgroundHangMonitorChan {
             sender,
             component_id: component_id,
             disconnected: Default::default(),
+            monitoring_enabled: opts::get().background_hang_monitor,
```
jdm, line 128: I would prefer to add new arguments that are provided by callers instead of adding more uses of opts::get(). One benefit is that it makes writing unit tests easier.

components/background_hang_monitor/background_hang_monitor.rs
```diff
@@ -187,6 +207,7 @@ impl BackgroundHangMonitorWorker {
             sampling_baseline: Instant::now(),
             creation: Instant::now(),
             samples: Default::default(),
+            monitoring_enabled: opts::get().background_hang_monitor,
```
jdm, line 210: Same here - let's use a new argument instead.

components/script/script_thread.rs
```diff
@@ -686,6 +693,24 @@ pub struct ScriptThread {
...
 }
 
+struct BHMExitSignal {
+    closing: Arc<AtomicBool>,
+    js_context: ContextForRequestInterrupt,
+}
+
+impl BackgroundHangMonitorExitSignal for BHMExitSignal {
+    fn signal_to_exit(&self) {
+        // Note: can we be sure the interrupt will see `true`?
```
jdm, line 703: I don't really know what this question means. If the interrupt uses the appropriate atomic operation sequencing, then it should.

jdm (approved):
This all looks pretty sensible!

components/script/script_thread.rs
```diff
@@ -1596,12 +1634,12 @@ impl ScriptThread {
...
 
-            // https://html.spec.whatwg.org/multipage/#event-loop-processing-model step 6
-            self.perform_a_microtask_checkpoint();
-
             if let Some(retval) = result {
                 return retval;
             }
+
+            // https://html.spec.whatwg.org/multipage/#event-loop-processing-model step 6
+            self.perform_a_microtask_checkpoint();
```
gterzian, line 1642: note: strictly speaking not necessary, however it does make sense to not run microtasks if we're exiting the script-thread anyway(the `return retval` above).

gterzian:
@jdm it got a little bit more complicated, please take another look(needs a squash after).

gterzian:
Ok I have broken something, will look into...

gterzian:
So all failures, on three boxes, are similar to:

```

  ▶ CRASH [expected FAIL] /css/CSS2/css1/c5501-mrgn-t-000.xht
  │ 
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ mozilla::detail::MutexImpl::mutexLock: pthread_mutex_lock failed: Invalid argument
  └ Stack trace
```

Is this a coincidence or have the changes here broken something about the reftests exit mechanism?

components/script/dom/dedicatedworkerglobalscope.rs
```diff
@@ -618,12 +618,19 @@ impl DedicatedWorkerGlobalScope {
...
+    let worker = DomRoot::downcast::<WorkerGlobalScope>(global.clone())
+        .expect("global is not a worker scope");
     assert!(worker.is::<DedicatedWorkerGlobalScope>());
 
+    let is_closing = worker.is_closing();
+    if is_closing {
+        // We are closing, ensure no script can run and potentially hang.
+        // For workers, the closing flag itself cancels all tasks.
+        global.microtask_queue().clear();
+    }
```
jdm, line 630: What if we add a check to GlobalScope::perform_a_microtask_checkpoint and ScriptThread::perform_a_microtask_checkpoint instead of adding this logic to each interrupt callback?

components/script/script_thread.rs
```diff
@@ -1580,6 +1600,24 @@ impl ScriptThread {
             let category = self.categorize_msg(&msg);
             let pipeline_id = self.message_to_pipeline(&msg);
 
+            if self.closing.load(Ordering::SeqCst) {
+                // If we're received the closed signal from the BHM, only handle exit messages.
```
jdm, line 1604: nit: s/we're/we've/

jdm (approved):

components/script/dom/dedicatedworkerglobalscope.rs
gterzian: Good idea._

gterzian:
```
  ▶ TIMEOUT [expected PASS] /_mozilla/mozilla/worklets/test_paint_worklet.html
  │ 
  │ internal error: entered unreachable code (thread <unnamed>, at components/script/dom/globalscope.rs:2733)
  │              at /rustc/a8cf3991177f30694200002cd9479ffbbe6d9a1a/src/libstd/panicking.rs:477
  │    4: script::dom::globalscope::GlobalScope::can_continue_running
  │    5: script::dom::globalscope::GlobalScope::perform_a_microtask_checkpoint
  │    6: <script::dom::bindings::settings_stack::AutoEntryScript as core::ops::drop::Drop>::drop
  │    7: profile_traits::time::profile
  │    8: script::dom::globalscope::GlobalScope::evaluate_script_on_global_with_result
  │    9: script::dom::workletglobalscope::WorkletGlobalScope::evaluate_js
  │   10: script::dom::worklet::WorkletThread::process_control
  │   11: script::dom::worklet::WorkletThread::run
  │              at /rustc/a8cf3991177f30694200002cd9479ffbbe6d9a1a/src/liballoc/boxed.rs:1076
  │       <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │              at /rustc/a8cf3991177f30694200002cd9479ffbbe6d9a1a/src/liballoc/boxed.rs:1076
  │              at /rustc/a8cf3991177f30694200002cd9479ffbbe6d9a1a/src/libstd/sys/unix/thread.rs:87
  │   15: start_thread
  │   16: __clone
  └ [2020-06-30T05:11:54Z ERROR servo] internal error: entered unreachable code
```

Ok, didn't account for worklets...

PR #27025, opened 2020-06-22
Allow embedder to bypass devtools prompt

paulrouget:
This introduces a token shared by servo and the embedder that can be used during the initial phase of the devtools connection.

Added also a commit to remove the devtools server panel. We just print the devtools port in the console now.


jdm:
@bors-servo r+
Nice solution!

jdm:
@bors-servo retry
* #23290 #27039

PR #27041, opened 2020-06-22
Report exceptions for async script executions to webdriver

### LINKED ISSUE #27031: Webdriver servo async JS script breaks some wdspec tests
In particular, tests/wpt/web-platform-tests/webdriver/tests/add_cookie/user_prompts.py uses:
```
        session.execute_async_script("""
            let dialog_type = arguments[0];
            let text = arguments[1];

            setTimeout(function() {
              if (dialog_type == 'prompt') {
                window.dialog_return_value = window[dialog_type](text, '');
              } else {
                window.dialog_return_value = window[dialog_type](text);
              }
            }, 0);
            """, args=(dialog_type, text))
```
which ends up yielding this error:
```
 0:06.21 pid:15764 [2020-06-22T14:23:42Z ERROR script::dom::bindings::error] Error at :10:65 window[dialog_type] is not a function
```
because dialog_type is a reference to the `webdriverCallback` method on Window instead of the a string like 'alert' or 'prompt'.

### LINKED ISSUE #27035: Webdriver server ignores JS script arguments
https://docs.rs/webdriver/0.40.2/webdriver/command/struct.JavascriptCommandParameters.html includes an `args` field that is ignored by both handle_execute_script and handle_async_execute_script.

### LINKED ISSUE #27036: Webdriver script execution doesn't report exceptions
https://w3c.github.io/webdriver/#execute-async-script includes steps for reporting exceptions to the client, and we ignore them instead.

jdm:
Improving the accuracy of the async script execution for our webdriver implementation allows us to run many more webdriver tests without timeouts, which should help with https://github.com/web-platform-tests/wpt/issues/24257.

Specification:
* https://w3c.github.io/webdriver/#dfn-clone-an-object
* https://w3c.github.io/webdriver/#execute-async-script

---
- [x] These changes fix #27036 and fix #27035 and fix #27031.
- [x] There are tests for these changes (but we don't run them in CI yet)

### DIFF
--- a/components/webdriver_server/lib.rs
+++ b/components/webdriver_server/lib.rs
@@ -1505,7 +1505,7 @@ impl Handler {
             .iter()
             .map(webdriver_value_to_js_argument)
             .collect();
-        args_string.push("window.webdriverCallback".to_string());
+        args_string.push("resolve".to_string());
 
         let timeout_script = if let Some(script_timeout) = self.session()?.script_timeout {
             format!("setTimeout(webdriverTimeout, {});", script_timeout)
@@ -1513,7 +1513,17 @@ impl Handler {
             "".into()
         };
         let script = format!(
-            "{} (function() {{ {}\n }})({})",
+            r#"(function() {{
+              let webdriverPromise = new Promise(function(resolve, reject) {{
+                  {}
+                  (async function() {{
+                    {}
+                  }})({})
+                    .then((v) => {{}}, (err) => reject(err))
+              }})
+              .then((v) => window.webdriverCallback(v), (r) => window.webdriverException(r))
+              .catch((r) => window.webdriverException(r));
+            }})();"#,
             timeout_script,
             func_body,
             args_string.join(", "),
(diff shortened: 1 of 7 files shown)


components/webdriver_server/lib.rs
```diff
@@ -1396,15 +1399,27 @@ impl Handler {
         parameters: &JavascriptCommandParameters,
     ) -> WebDriverResult<WebDriverResponse> {
         let func_body = &parameters.script;
-        let args_string = "window.webdriverCallback";
+        let mut args = parameters.args.as_ref().map_or(vec![], |args| {
+            args.iter().map(|a| format!("{}", a)).collect()
```
Manishearth, line 1403:
```suggestion
            args.iter().map(|a| a.to_string()).collect()
```

can also use ToString::to_string

Manishearth (approved):
seems okay

yezhizhen (commented):
If I understand correctly, [cyclic.py](https://github.com/servo/servo/blob/main/tests/wpt/tests/webdriver/tests/classic/execute_async_script/cyclic.py) tests that cyclic references correctly throw exception, so that we won't have infinite loop when deserializing and TIMEOUT?

- This PR added `seen` according to [Clone an object](https://w3c.github.io/webdriver/#dfn-clone-an-object ) to support above feature.
- `fn WebdriverException` is defined to catch throw exception from also-modified script. That's why `test_object_to_json_exception` is passing now.

I am not familiar with codegen and `can_gc`. But rest looks good!

jdm:
Yep!

PR #27044, opened 2020-06-22
First pass at implementing the Flex Layout Algorithm

SimonSapin:
CC https://github.com/servo/servo/issues/26639


SimonSapin:
```
2 unexpected results that are NOT known-intermittents:
  ▶ OK [expected CRASH] /css/css-flexbox/layout-with-inline-svg-001.html
  ▶ FAIL [expected CRASH] /css/css-flexbox/dynamic-change-simplified-layout.html
  │   → /css/css-flexbox/dynamic-change-simplified-layout.html ['d5b47219b5f8817a35f1abe46e9490c6e792a39b']
  └   → /css/reference/ref-filled-green-100px-square.xht ['25f7c572f42824d5608beb0ca129c47eec6c3272']
```

The "expected" crashes are panics in `layout::positioned::HoistedAbsolutelyPositionedBox::layout` with message `already mutably borrowed`. In `wpt/css/css-flexbox/layout-with-inline-svg-001.html` for example it’s fairly reliable to reproduce on my machine, but disappears with `STYLO_THREADS=1`.

I suspect that any test with both flexbox and abspos is potentially impacted, so marking a couple file names as knows intermittent is not gonna help much. I’m looking into this to hopefully fix it before landing.

More generally though, @nox do you have thoughts on:

* How can we reason about the duration of these atomic refcell borrows and have some confidence that we expect them not to overlap regardless of thread scheduling?
* Or, do we need to switch to `RwLock` to block instead of panicking, if overlaps are unavoidable?

SimonSapin:
Figured it out. This issue is when implementing this bit of spec:

> If the flex item has `align-self: stretch`, redo layout for its contents, treating this used size as its definite cross size so that percentage-sized children can be resolved.

“Do layout” has the side effect of collecting absolutely-positioned descendants into `PositioningContext`, so when redoing we ended up with two entries for the same abspos box. Later we lay out that box, twice in parallel leading to overlapping atomic refcell borrows. The fix is to make a new `PositioningContext` when laying out a flex item and having it part of some return value. When "re-doing", we only use one of the results.

@bors-servo try=wpt-2020

nox:
I'm a bit confused why some iterators need to be collected when they are used a single time any way, but otherwise, LGTM.

SimonSapin:
For example the module-level `layout` function returns an `impl Iterator` which lives longer than `layout`’s call stack frame, so all intermediate values need to be owned rather than borrowed. But this has all been refactored a couple times already so it’s possible that I still missed some `Vec`s that don’t need to be collected.

jdm:
@bors-servo retry
* #27062 #23290

components/layout_2020/flexbox/layout.rs
```diff
@@ -2,32 +2,1090 @@
...
+            // The spec is missing something to resolve this conflict:
+            // https://github.com/w3c/csswg-drafts/issues/5190
+            // And we’ll need to change the signature of `IndependentFormattingContext::layout`
+            // to allow the inner formatting context to “negociate” a used inline-size
+            // with the outer one somehow.
+            container_main_size
+        },
+    };
+    let fragments = flex_lines
+        .into_iter()
```
Manishearth, line 352: Can this loop be merged with the previous one?

Manishearth: I guess we're typically iterating over a small number of items so it won't matter much

components/layout_2020/flexbox/layout.rs
```diff
@@ -2,32 +2,1090 @@
...
+                        FlexBasis::Content
+                    }
+                },
+            }
+        },
+        FlexBasis::Size(Size::Auto) => {
+            // “When specified on a flex item, the `auto` keyword retrieves
+            //  the value of the main size property as the used `flex-basis`.”
+            match content_box_size.main {
+                LengthOrAuto::LengthPercentage(length) => FlexBasis::Size(length),
```
Manishearth, line 501: should percentages be resolved relative to something here?

components/layout_2020/flexbox/layout.rs
SimonSapin: This PR was getting big, so at some point I stopped refactoring and focused on getting it landed. I’m sure it can be improved still.

There’s a limitation though where the returned `impl Iterator` needs to take ownership of every closure captures, it cannot borrow stuff on the stack. Some `.collect()` calls help with that.

components/layout_2020/flexbox/layout.rs
SimonSapin: `content_box_size` here originally comes from the `ComputedValuesExt` method of the same name, which already takes care of resolving percentages (but not `auto`)

components/layout_2020/flexbox/layout.rs
```diff
@@ -2,32 +2,1090 @@
...
+                        flex_items.push((tree_rank, item));
+                        Err(())
+                    },
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let mut content_block_size_option_dance = None;
+        let fragments =
+            positioning_context.adjust_static_positions(tree_rank, |positioning_context| {
```
Manishearth, line 166: Why does this need to be done? What does it do?

components/layout_2020/flexbox/layout.rs
```diff
@@ -2,32 +2,1090 @@
...
+                    "Mixed writing modes are not supported yet"
+                );
+                // … and also the item’s inline axis.
+
+                match self.box_ {
+                    IndependentFormattingContext::Replaced(replaced) => {
+                        let pbm = replaced
+                            .style
+                            .padding_border_margin(flex_context.containing_block);
+                        let size = replaced.contents.used_size_as_if_inline_element(
```
Manishearth, line 924: Where do we do the "treat auto as fit-content"?

components/layout_2020/flexbox/layout.rs
```diff
@@ -642,13 +643,16 @@ impl FlexLine<'_> {
...
-                (cross_size, fragments)
+                if has_stretch_auto {
+                    // “If the flex item has `align-self: stretch`, redo layout for its contents,
+                    //  treating this used size as its definite cross size
+                    //  so that percentage-sized children can be resolved.”
+                    item_result = item.layout(used_main_size, flex_context, Some(cross_size));
+                }
+                flex_context
+                    .positioning_context
+                    .append(item_result.positioning_context);
```
Manishearth, line 654: why was this necessary?

Manishearth: as in, how does the positioning context help with side effects?

components/layout_2020/flexbox/layout.rs
SimonSapin: The `BoxFragment::content_rect::start_corner` position of a fragment is a point in a coordinate space whose origin is the start corner of the *parent* in the in the fragment tree. Using relative coordinates like this allows translating an entire sub-tree by modifying the coordinates of its root fragment.

When hoisting an absolutely-positioned box "up" the tree, we’re effectively moving to another coordinate space. `HoistedAbsolutelyPositionedBox::box_offsets` may contain coordinates that will contribute to the position of the eventual fragment. `adjust_static_positions` effectively converts them to another coordinate space.

components/layout_2020/flexbox/layout.rs
SimonSapin: `IndependentFormattingContext::layout` takes a `&mut PositioningContext` parameter, and has the side effect of (possibly) appending to it. To isolate this side-effect, we construct a new `PositioningContext` for each flex item, and keep it with the "result of layout". If we end up doing layout for a flex item again because of `align-self: stretch`, we throw away that `PositioningContext` and make a new one. This was added in https://github.com/servo/servo/pull/27044/commits/22b60d8b82222b542384dc8b8c2534dcfd2eda78. At some point though we want a single `PositioningContext` for the entire flex container, so we append the item-specific ones together.

I am working on a branch that removes `&mut PositioningContext` parameters everywhere and keeps hoisted boxes in respective "result of layout" types instead.

components/layout_2020/flexbox/layout.rs
SimonSapin: This code path is for replaced elements. I *think* `auto` and `fit-content` resolve to the same thing for those, though I might be wrong for some cases and it’s possible that we don’t implement everything in https://drafts.csswg.org/css-sizing/#intrinsic accurately.

PR #27059, opened 2020-06-23
UWP console: force white background

### LINKED ISSUE #27040: console.log is drawn in white text over white background
A partner says that log messages are drawn in white text, but other message types are not.

cc @paulrouget

paulrouget:
Fix #27040


jdm:
@bors-servo retry
* #26895

PR #27084, opened 2020-06-25
Implement GPUBuffer.mapAsync and update wgpu-core

kunalmohan:
This PR covers the following-
1. update wgpu-core to use serializable `RenderPass` and `ComputePass`. Implement `DispatchIndirect`.
2. Implement `GPUBuffer.mapAsync()`.

The `mapAsync()` methods has the following flow-
1. A new promise is created in `mapAsync()` and a message sent to the server-side. This returns a promise.
2. On the server-side, a `struct BufferMapInfo` object stores the relevant info required to send data to the client-side on completion of mapping. This object is stored in a `buffer_maps: HashMap<id::BufferId, BufferMapInfo<'a>>,` in `WGPU`. `buffers_maps` basically stores the data for buffers with pending mapping.
3. `buffer_map_async()` command is issued with a callback responsible to send the data back to the client.
4. All the devices are polled every time the server receives a new message.
5. The data is sent back on completion and promise resolved through `AsyncWGPUListener` (similar to how the `requestDevice()` and `RequestAdapter` are processed).
6. On receiving the data, the client sends back a `BufferMapComplete` message to the server to remove the entry from `buffer_maps` hashmap.

I intended to implement `getMappedRange()`, too, in this PR but it's quite a task in itself (implementing intersecting ArrayBuffers). This already being a 500+ line PR, I thought it best to get this model approved first (also it will probably be easier to review this and `getMappedRange` separately). `getMappedRange` would come up in the next one.

---

### DIFF
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -146,17 +146,17 @@ impl GPUBufferMethods for GPUBuffer {
                 match ArrayBuffer::from(self.mapping.get()) {
                     Ok(array_buffer) => {
                         // Step 3.2
-                        if Some(GPUMapModeConstants::READ) != self.map_mode.get() {
-                            self.channel
-                                .0
-                                .send(WebGPURequest::UnmapBuffer {
-                                    device_id: self.device.0,
-                                    buffer_id: self.id().0,
-                                    array_buffer: array_buffer.to_vec(),
-                                    mapped_at_creation: self.map_mode.get() == None,
-                                })
-                                .unwrap();
-                        }
+                        self.channel
+                            .0
+                            .send(WebGPURequest::UnmapBuffer {
+                                buffer_id: self.id().0,
+                                array_buffer: array_buffer.to_vec(),
+                                is_map_read: self.map_mode.get() == Some(GPUMapModeConstants::READ),
+                                offset: self.mapping_range.borrow().start,
+                                size: self.mapping_range.borrow().end -
+                                    self.mapping_range.borrow().start,
+                            })
+                            .unwrap();
                         // Step 3.3
                         unsafe {
                             DetachArrayBuffer(*cx, self.mapping.handle());
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -19,7 +19,6 @@ use dom_struct::dom_struct;
 use js::jsapi::{Heap, JSObject};
 use js::jsval::UndefinedValue;
 use js::rust::jsapi_wrapped::{DetachArrayBuffer, IsPromiseObject, RejectPromise};
-use js::rust::MutableHandle;
 use js::typedarray::{ArrayBuffer, CreateWith};
 use std::cell::Cell;
 use std::ops::Range;
@@ -277,12 +276,10 @@ impl AsyncWGPUListener for GPUBuffer {
     fn handle_response(&self, response: WebGPUResponse, promise: &Rc<Promise>) {
         match response {
             WebGPUResponse::BufferMapAsync(bytes) => {
+                let cx = self.global().get_cx();
+                rooted!(in(*cx) let mut array_buffer = ptr::null_mut::<JSObject>());
                 match unsafe {
-                    ArrayBuffer::create(
-                        *self.global().get_cx(),
-                        CreateWith::Slice(&bytes),
-                        MutableHandle::from_raw(self.mapping.handle_mut()),
-                    )
+                    ArrayBuffer::create(*cx, CreateWith::Slice(&bytes), array_buffer.handle_mut())
                 } {
                     Ok(_) => promise.resolve_native(&()),
                     Err(()) => {
@@ -293,6 +290,7 @@ impl AsyncWGPUListener for GPUBuffer {
                         promise.reject_error(Error::Operation);
                     },
                 }
+                self.mapping.set(array_buffer.get());
                 self.state.set(GPUBufferState::Mapped);
             },
             _ => {
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -2,8 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use crate::dom::bindings::cell::{DomRefCell, Ref};
-use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::{GPUBufferMethods, GPUSize64};
+use crate::dom::bindings::cell::DomRefCell;
+use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUBufferMethods;
+use crate::dom::bindings::codegen::Bindings::GPUMapModeBinding::GPUMapModeConstants;
 use crate::dom::bindings::error::Error;
 use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
@@ -11,22 +12,28 @@ use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::DOMString;
 use crate::dom::bindings::trace::RootedTraceableBox;
 use crate::dom::globalscope::GlobalScope;
+use crate::dom::gpu::{response_async, AsyncWGPUListener};
+use crate::dom::promise::Promise;
+use crate::realms::InRealm;
 use dom_struct::dom_struct;
 use js::jsapi::{Heap, JSObject};
 use js::jsval::UndefinedValue;
 use js::rust::jsapi_wrapped::{DetachArrayBuffer, IsPromiseObject, RejectPromise};
-use js::typedarray::ArrayBuffer;
+use js::rust::MutableHandle;
+use js::typedarray::{ArrayBuffer, CreateWith};
 use std::cell::Cell;
 use std::ptr;
-use webgpu::{WebGPU, WebGPUBuffer, WebGPUDevice, WebGPURequest};
+use std::rc::Rc;
+use webgpu::{
+    wgpu::device::HostMap, WebGPU, WebGPUBuffer, WebGPUDevice, WebGPURequest, WebGPUResponse,
+};
 
 // https://gpuweb.github.io/gpuweb/#buffer-state
-#[derive(Clone, MallocSizeOf)]
+#[derive(Clone, Copy, MallocSizeOf, PartialEq)]
 pub enum GPUBufferState {
-    MappedForReading,
-    MappedForWriting,
-    MappedPendingForReading,
-    MappedPendingForWriting,
+    Mapped,
+    MappedAtCreation,
+    MappingPending,
     Unmapped,
     Destroyed,
 }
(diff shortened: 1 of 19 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -284,28 +227,16 @@ impl GPUDeviceMethods for GPUDevice {
                 let ty = match bind.type_ {
                     GPUBindingType::Uniform_buffer => wgt::BindingType::UniformBuffer {
                         dynamic: bind.hasDynamicOffset,
-                        min_binding_size: if bind.minBufferBindingSize == 0 {
-                            None
-                        } else {
-                            Some(NonZeroU64::new(bind.minBufferBindingSize).unwrap())
-                        },
+                        min_binding_size: wgt::BufferSize::new(bind.minBufferBindingSize),
```
kvark, line 230: oh this is niiice!

components/script/dom/gpudevice.rs
```diff
@@ -433,11 +364,7 @@ impl GPUDeviceMethods for GPUDevice {
                             WebGPUBindings::Buffer(BufferBinding {
                                 buffer_id: b.buffer.id().0,
                                 offset: b.offset,
-                                size: if let Some(s) = b.size {
-                                    wgt::BufferSize(s)
-                                } else {
-                                    wgt::BufferSize::WHOLE
-                                },
+                                size: wgt::BufferSize::new(b.size.unwrap_or(0)),
```
kvark, line 367: nit: super bikesheddy, it could be `b.size.and_then(wgt::BufferSize::new)`

components/webgpu/lib.rs
```diff
@@ -296,20 +314,69 @@ impl WGPU {
...
         }
     }
 
-    fn run(mut self) {
+    fn run(&'a mut self) {
         while let Ok(msg) = self.receiver.recv() {
+            self.devices.iter().for_each(|device| {
+                let global = &self.global;
+                let device_id = device.0;
+                gfx_select!(device_id => global.device_poll(device_id, false));
```
kvark, line 334: would be nice to move it out and poll less often

components/webgpu/lib.rs
```diff
@@ -296,20 +314,69 @@ impl WGPU {
...
             webrender_document,
             external_images,
             wgpu_image_map,
         }
     }
 
-    fn run(mut self) {
+    fn run(&'a mut self) {
         while let Ok(msg) = self.receiver.recv() {
+            self.devices.iter().for_each(|device| {
```
kvark, line 331: we can use [poll_all_devices](https://github.com/gfx-rs/wgpu/blob/29e286bdc1474457c2a34155d559e30268df74ad/wgpu-core/src/device/mod.rs#L2733) instead

components/webgpu/lib.rs
```diff
@@ -296,20 +314,69 @@ impl WGPU {
...
+                    let map_info = BufferMapInfo {
+                        buffer_id,
+                        sender,
+                        global: &self.global,
+                        size: (map_range.end - map_range.start) as usize,
+                    };
+                    self.buffer_maps.insert(buffer_id, map_info);
+
+                    unsafe extern "C" fn callback(status: BufferMapAsyncStatus, userdata: *mut u8) {
+                        let info = (userdata as *mut BufferMapInfo).as_ref().unwrap();
```
kvark, line 352: This looks totally unsafe, the lifetime of `a` can't be respected if the stuff is unsafely casted from C.
Maybe just turn it into a `NonNull` pointer?

components/webgpu/lib.rs
```diff
@@ -296,20 +314,69 @@ impl WGPU {
...
+                                let data_pt = gfx_select!(info.buffer_id =>
+                                    global.buffer_get_mapped_range(info.buffer_id, 0, None));
+                                let data = slice::from_raw_parts(data_pt, info.size).to_vec();
+                                if let Err(e) =
+                                    info.sender.send(Ok(WebGPUResponse::BufferMapAsync(data)))
+                                {
+                                    warn!("Could not send BufferMapAsync Response ({})", e);
+                                }
+                            },
+                            _ => warn!("Could not map buffer({:?})", info.buffer_id),
```
kvark, line 365: we should `error!` in these cases, since we aren't really handling them gracefully

components/webgpu/lib.rs
```diff
@@ -296,20 +314,69 @@ impl WGPU {
...
+                    self.buffer_maps.insert(buffer_id, map_info);
+
+                    unsafe extern "C" fn callback(status: BufferMapAsyncStatus, userdata: *mut u8) {
+                        let info = (userdata as *mut BufferMapInfo).as_ref().unwrap();
+                        match status {
+                            BufferMapAsyncStatus::Success => {
+                                let global = info.global;
+                                let data_pt = gfx_select!(info.buffer_id =>
+                                    global.buffer_get_mapped_range(info.buffer_id, 0, None));
+                                let data = slice::from_raw_parts(data_pt, info.size).to_vec();
```
kvark, line 358: are we sending an immutable slice? That would only work for read mapping, not write mappings.

kvark (changes requested):
Very nice! Just a few notes here

components/script/dom/gpubuffer.rs
```diff
@@ -79,14 +88,21 @@ impl GPUBuffer {
         buffer: WebGPUBuffer,
         device: WebGPUDevice,
         state: GPUBufferState,
-        size: GPUSize64,
-        usage: u32,
+        size: u64,
```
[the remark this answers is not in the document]
kunalmohan: Nothing in particular. I had removed the size field initially while cleaning up the extra validation for `copyBuffertoBuffer`. Then, while implementing `mapAsync` I had to reintroduce it so I just used u64 instead of importing GPUSize64. I'll change it if it looks more appropriate.

components/script/dom/gpubuffer.rs
```diff
@@ -202,3 +269,40 @@ impl GPUBufferMethods for GPUBuffer {
         *self.label.borrow_mut() = value;
     }
 }
+
+impl AsyncWGPUListener for GPUBuffer {
+    #[allow(unsafe_code)]
+    fn handle_response(&self, response: WebGPUResponse, promise: &Rc<Promise>) {
+        match response {
+            WebGPUResponse::BufferMapAsync(bytes) => {
+                if unsafe {
```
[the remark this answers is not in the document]
kunalmohan: The return type for `ArrayBuffer::create()` is `Result<(),()>` so there is no actual error to log here. Logging a custom message could still be helpful so I'll do that.

components/webgpu/lib.rs
kunalmohan: I am a bit confused about what your concern here is. We are sending a vector to the client. If I have to change the mutability I'll have to do it on the client-side while receiving, i.e. `WebGPUResponse::BufferMapAsync(mut bytes)` in `handle_response()`? And we'll be writing back any data in the `WebGPURequest::Unmap` via `write_buffer()` or 'set_buffer_sub_data()`. How does that change things here?

components/script/dom/gpubuffer.rs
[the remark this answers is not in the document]
kvark: ah ok. In cases like this `match` is still better (it's better in most cases), since you can match on `Err(())` to explicitly show that the error is `()` and avoid people asking weird questions :)

components/script/dom/gpubuffer.rs
```diff
@@ -189,7 +207,56 @@ impl GPUBufferMethods for GPUBuffer {
...
+    }
+
+    #[allow(unsafe_code)]
+    /// https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync-offset-size
+    fn MapAsync(&self, mode: u32, offset: u64, size: u64, comp: InRealm) -> Rc<Promise> {
+        let promise = Promise::new_in_current_realm(&self.global(), comp);
+        let map_range = if size == 0 {
+            offset..self.size
+        } else {
+            offset..size
```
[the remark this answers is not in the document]
kvark: please resolve the things when you push the code that fixes them, not earlier
otherwise, it's hard to know when to verify resolution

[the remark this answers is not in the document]
kunalmohan: Sorry :sweat_smile: . I'll keep that in mind.

components/webgpu/lib.rs
kvark: Here is how I see this working, at this point at least:
  1. for a write-mappable buffer, the implementation carries around the shared memory with its data. When user asks for mapping, you make the server to notify the client when the GPU is done, and you just expose this shmem to the user to modify, followed by `set_buffer_sub_data` on the server side.
  2. for a read-mappable buffer, the server waits for GPU, copies the immutable slice out into a shmem that is given to the user.

I suppose the code we see here is for (2) only? In this case, why do we have `host_map` parameter to the message?

kunalmohan: I see. I had a different approach in my mind. I was going with sending vectors across the processes for data exchange between client and server.
I guess you are pointing towards using [`IpcSharedMemory`](https://docs.rs/ipc-channel/0.14.1/ipc_channel/ipc/struct.IpcSharedMemory.html) (similar to the previous `MapReadAsync` implementation).

kunalmohan: Sorry. I see there is a bigger concern here than the `IpcSharedMemory`. But it would probably be solved when I figure out the intersecting ArrayBuffers part.

kunalmohan:
I have addressed most of the review comments. I am still unsure about the last one- about how that would be done. I'll try to dig into it but maybe it will be better to solve that along with `get_mapped_range()` depending how the data will be exposed to the user. I can convert this to a draft and continue implementing `get_mapped_range()` here itself if that's a better idea.

kvark (approved):
We had a chat about this on "#wgpu:matrix.org", and it looks like the PR is on the right track:)

kunalmohan:
@kvark I have changed the Unmapping process to use `get_mapped_range()` to write back the data to buffer while unmapping in case of `mapAsync` as well as `mapped_at_creation`.

kvark (approved):
Looks reasonable, thank you!

kunalmohan:
```
error[E0599]: no method named `handle_mut` found for struct `dom::bindings::trace::RootedTraceableBox<mozjs_sys::jsgc::Heap<*mut js::jsapi::JSObject>>` in the current scope
    --> components/script/dom/gpubuffer.rs:284:62
     |
284  |                         MutableHandle::from_raw(self.mapping.handle_mut()),
     |                                                              ^^^^^^^^^^ method not found in `dom::bindings::trace::RootedTraceableBox<mozjs_sys::jsgc::Heap<*mut js::jsapi::JSObject>>`
     | 
    ::: components/script/dom/bindings/trace.rs:1036:1
     |
1036 | pub struct RootedTraceableBox<T: 'static + JSTraceable> {
     | ------------------------------------------------------- method `handle_mut` not found for this
error[E0599]: no method named `handle_mut` found for struct `dom::bindings::trace::RootedTraceableBox<mozjs_sys::jsgc::Heap<_>>` in the current scope
    --> components/script/dom/gpudevice.rs:189:53
     |
189  |                     MutableHandle::from_raw(mapping.handle_mut()),
     |                                                     ^^^^^^^^^^ method not found in `dom::bindings::trace::RootedTraceableBox<mozjs_sys::jsgc::Heap<_>>`
     | 
    ::: components/script/dom/bindings/trace.rs:1036:1
     |
1036 | pub struct RootedTraceableBox<T: 'static + JSTraceable> {
     | ------------------------------------------------------- method `handle_mut` not found for this
```
checks failed. I didn't get this on my system earlier, but on a rebase against master I can see it failing now. I'll rectify this and push the changes.

jdm:
@bors-servo retry
* #25522

PR #26729, opened 2020-06-01
Implement :valid :invalid pseudo classes

tipowol:
This pull request implements :valid :invalid pseudo classes. Radio button support is currently missing due to concerns about performance. I am willing to discuss how to implement it efficiently.

---
- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #10781
- [X] There are tests for these changes

Issue #10781, opened 2016-04-21
Implement the :valid and :invalid pseudo-classes


Ms2ger:
Blocks #2185, #2968.
- Spec: https://html.spec.whatwg.org/multipage/#selector-invalid.
- Code: `components/style/selector_impl.rs`, `components/layout/wrapper.rs`, `components/script/dom/element.rs`.
- Test: `html/semantics/selectors/pseudo-classes/valid-invalid.html`, `dom/nodes/Element-closest.html`.

shssoichiro:
I'd like to claim this one.

shssoichiro:
Do we have existing methods for checking the validity of an input, e.g. for form submission?

jdm:
Hmm, the problem here is that there is a student group currently working on implementing form input validation. The work to implement the pseudo-class is separate from that, but difficult to test without the actual validation present. I'd like to hold off on this until the necessary pieces are in place.

tipowol:
@highfive assign me

PR #27100, opened 2020-06-27
Implement HTMLFormElement.requestSubmit()

### LINKED ISSUE #23417: Implement <form>.requestSubmit()
whatwg/html#4597
web-platform-tests/wpt#16743

Firefox Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1552301

muodov:
This PR contains an implementation of [HTMLFormElement.requestSubmit()](https://html.spec.whatwg.org/multipage/forms.html#dom-form-requestsubmit)

This is literally my first hundred lines of Rust code, so if I crossed a few sacred lines here and there, please go easy on me :)

---

- [x] These changes fix #23417

- [x] [WPT tests](https://github.com/servo/servo/blob/master/tests/wpt/web-platform-tests/html/semantics/forms/the-form-element/form-requestsubmit.html) for these changes
There are two tests that still fail because we don't support `:invalid` CSS selector (see #10781). I verified that they pass if you change them to not use `:invalid`. Should be unlocked by #26729.

### DIFF
--- a/components/script/dom/htmlbuttonelement.rs
+++ b/components/script/dom/htmlbuttonelement.rs
@@ -77,6 +77,11 @@ impl HTMLButtonElement {
             document,
         )
     }
+
+    #[inline]
+    pub fn is_submit_button(&self) -> bool {
+        self.button_type.get() == ButtonType::Submit
+    }
 }
 
 impl HTMLButtonElementMethods for HTMLButtonElement {
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -16,6 +16,7 @@ use crate::dom::bindings::codegen::Bindings::HTMLInputElementBinding::HTMLInputE
 use crate::dom::bindings::codegen::Bindings::HTMLTextAreaElementBinding::HTMLTextAreaElementMethods;
 use crate::dom::bindings::codegen::Bindings::NodeListBinding::NodeListMethods;
 use crate::dom::bindings::codegen::Bindings::WindowBinding::WindowBinding::WindowMethods;
+use crate::dom::bindings::error::{Error, Fallible};
 use crate::dom::bindings::inheritance::{Castable, ElementTypeId, HTMLElementTypeId, NodeTypeId};
 use crate::dom::bindings::refcounted::Trusted;
 use crate::dom::bindings::reflector::DomObject;
@@ -88,6 +89,7 @@ pub struct HTMLFormElement {
     generation_id: Cell<GenerationId>,
     controls: DomRefCell<Vec<Dom<Element>>>,
     past_names_map: DomRefCell<HashMap<Atom, (Dom<Element>, Tm)>>,
+    firing_submission_events: Cell<bool>,
 }
 
 impl HTMLFormElement {
@@ -104,6 +106,7 @@ impl HTMLFormElement {
             generation_id: Cell::new(GenerationId(0)),
             controls: DomRefCell::new(Vec::new()),
             past_names_map: DomRefCell::new(HashMap::new()),
+            firing_submission_events: Cell::new(false),
         }
     }
 
@@ -241,6 +244,67 @@ impl HTMLFormElementMethods for HTMLFormElement {
     // https://html.spec.whatwg.org/multipage/#the-form-element:concept-form-submit
     fn Submit(&self) {
         self.submit(SubmittedFrom::FromForm, FormSubmitter::FormElement(self));
+    }
+
+    // https://html.spec.whatwg.org/multipage/#dom-form-requestsubmit
+    fn RequestSubmit(&self, submitter: Option<&HTMLElement>) -> Fallible<()> {
+        let submitter: FormSubmitter = match submitter {
+            Some(submitter_element) => {
+                // Step 1.1
+                let error_not_a_submit_button =
+                    Err(Error::Type("submitter must be a submit button".to_string()));
+
+                let element = match submitter_element.upcast::<Node>().type_id() {
+                    NodeTypeId::Element(ElementTypeId::HTMLElement(element)) => element,
+                    _ => {
+                        return error_not_a_submit_button;
+                    },
+                };
+
+                let submit_button = match element {
+                    HTMLElementTypeId::HTMLInputElement => FormSubmitter::InputElement(
+                        &submitter_element
+                            .downcast::<HTMLInputElement>()
+                            .expect("Failed to downcast submitter elem to HTMLInputElement."),
+                    ),
+                    HTMLElementTypeId::HTMLButtonElement => FormSubmitter::ButtonElement(
+                        &submitter_element
+                            .downcast::<HTMLButtonElement>()
+                            .expect("Failed to downcast submitter elem to HTMLButtonElement."),
+                    ),
+                    _ => {
+                        return error_not_a_submit_button;
+                    },
+                };
+
+                if !submit_button.is_submit_button() {
+                    return error_not_a_submit_button;
+                }
+
+                let submitters_owner = submit_button.form_owner();
+
+                // Step 1.2
+                let owner = match submitters_owner {
+                    Some(owner) => owner,
+                    None => {
+                        return Err(Error::NotFound);
+                    },
+                };
+
+                if *owner != *self {
+                    return Err(Error::NotFound);
+                }
+
+                submit_button
+            },
+            None => {
+                // Step 2
+                FormSubmitter::FormElement(&self)
+            },
+        };
+        // Step 3
+        self.submit(SubmittedFrom::NotFromForm, submitter);
+        Ok(())
     }
 
     // https://html.spec.whatwg.org/multipage/#dom-form-reset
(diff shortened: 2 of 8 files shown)


tests/wpt/metadata/html/semantics/forms/the-form-element/form-requestsubmit.html.ini
```diff
@@ -1,22 +1,22 @@
 [form-requestsubmit.html]
   [requestSubmit() doesn't run interactive validation reentrantly]
-    expected: FAIL
+    expected: PASS
```
atouchet, line 3: These passing tests can be removed instead of being marked as PASS.

components/script/dom/htmlformelement.rs
```diff
@@ -1236,7 +1293,8 @@ pub enum FormMethod {
 pub enum FormSubmitter<'a> {
     FormElement(&'a HTMLFormElement),
     InputElement(&'a HTMLInputElement),
-    ButtonElement(&'a HTMLButtonElement), // TODO: image submit, etc etc
+    ButtonElement(&'a HTMLButtonElement),
+    // TODO: the spec doesn't seem to put hard constraints on submitter, so it can theoretically be any element
```
gterzian, line 1297: I think instead of here, we can add a doc to the enum with `\\\ <https://html.spec.whatwg.org/multipage/#form-associated-element>`, and then add a TODO that would simply state to implement the other types of form associated elements, including custom elements.

components/script/dom/htmlformelement.rs
```diff
@@ -1332,6 +1390,29 @@ impl<'a> FormSubmitter<'a> {
                 ),
         }
     }
+
+    // https://html.spec.whatwg.org/multipage/#concept-submit-button
+    fn is_submit_button(&self) -> bool {
```
gterzian, line 1395: The spec reads to me as if only a button element could be potentially a submit button, if that is correct we should remove the `InputElement` branch.

components/script/dom/htmlformelement.rs
```diff
@@ -599,28 +646,35 @@ impl HTMLFormElement {
...
-                if f == self {
-                    None
-                } else {
-                    Some(f.upcast::<HTMLElement>())
+            self.firing_submission_events.set(true);
+
+            if !submitter.no_validate(self) {
+                if self.interactive_validation().is_err() {
+                    // TODO: Implement event handlers on all form control elements
+                    self.upcast::<EventTarget>().fire_event(atom!("invalid"));
```
gterzian, line 658: I'm seeing the even firing as part of static validation, not interactive validation. See https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constraint-validation:event-invalid

components/script/dom/htmlformelement.rs
```diff
@@ -599,28 +646,35 @@ impl HTMLFormElement {
         let base = doc.base_url();
         // TODO: Handle browsing contexts (Step 4, 5)
         // Step 6
-        if submit_method_flag == SubmittedFrom::NotFromForm && !submitter.no_validate(self) {
-            if self.interactive_validation().is_err() {
-                // TODO: Implement event handlers on all form control elements
-                self.upcast::<EventTarget>().fire_event(atom!("invalid"));
+        if submit_method_flag == SubmittedFrom::NotFromForm {
```
gterzian, line 649: Please number the various Step 6 substeps, like you've already done with 6.4 below

components/script/dom/htmlformelement.rs
```diff
@@ -630,17 +684,20 @@ impl HTMLFormElement {
             );
             let event = event.upcast::<Event>();
             event.fire(self.upcast::<EventTarget>());
+
+            self.firing_submission_events.set(false);
```
gterzian, line 688: Please number as Step 6.6

components/script/dom/htmlbuttonelement.rs
```diff
@@ -77,6 +77,11 @@ impl HTMLButtonElement {
             document,
         )
     }
+
+    #[inline]
+    pub fn button_type(&self) -> ButtonType {
```
gterzian, line 82: I would rather have a `is_submit` method on HTMLButtonElement, since the `button_type` seems to only be used for that purpose(thay way the `ButtonType` doesn't have to become public either).

gterzian (changes requested):
Looks great! Couple of comments...

components/script/dom/htmlformelement.rs
```diff
@@ -1332,6 +1390,29 @@ impl<'a> FormSubmitter<'a> {
                 ),
         }
     }
+
+    // https://html.spec.whatwg.org/multipage/#concept-submit-button
+    fn is_submit_button(&self) -> bool {
+        match *self {
+            FormSubmitter::InputElement(input_element) => {
```
gterzian, line 1397: I think here we can add a link to https://html.spec.whatwg.org/multipage/input.html#image-button-state-(type=image)

components/script/dom/htmlformelement.rs
```diff
@@ -1332,6 +1390,29 @@ impl<'a> FormSubmitter<'a> {
...
+                input_element.input_type() == InputType::Submit ||
+                    input_element.input_type() == InputType::Image
+            },
+            FormSubmitter::ButtonElement(button_element) => {
```
gterzian, line 1401: and I think this would be https://html.spec.whatwg.org/multipage/#submit-button-state-(type=submit)

components/script/dom/htmlbuttonelement.rs
muodov: Sounds reasonable, done.

components/script/dom/htmlformelement.rs
```diff
@@ -243,6 +246,50 @@ impl HTMLFormElementMethods for HTMLFormElement {
         self.submit(SubmittedFrom::FromForm, FormSubmitter::FormElement(self));
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-form-requestsubmit
+    fn RequestSubmit(&self, submitter: Option<&HTMLElement>) -> Fallible<()> {
+        let submitter: FormSubmitter = match submitter {
+            Some(submitter_element) => {
+                // Step 1.1
+                let submit_button = match submitter_element.upcast::<Node>().type_id() {
```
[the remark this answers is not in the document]
muodov: :) As a matter of fact, I was doubting whether I should do this when I was first writing this. Wasn't sure what would be more "Rust-y". I suppose it comes down to a code style really, and I don't have a strong opinion on this, so I changed it as you suggested 👍

components/script/dom/htmlformelement.rs
```diff
@@ -243,6 +246,50 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+
+                if submit_button.is_none() || !submit_button.unwrap().is_submit_button() {
+                    return Err(Error::Type("submitter must be a submit button".to_string()));
+                }
+
+                let submit_button = submit_button.unwrap();
+                let submitters_owner = submit_button.form_owner();
+
+                // Step 1.2
+                if submitters_owner.is_none() ||
```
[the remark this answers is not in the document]
muodov: Done, with a small change: `*owner != *self`

components/script/dom/htmlformelement.rs
```diff
@@ -243,6 +246,50 @@ impl HTMLFormElementMethods for HTMLFormElement {
...
+                submit_button
+            },
+            None => {
+                // Step 2
+                FormSubmitter::FormElement(&self)
+            },
+        };
+        // Step 3
+        self.submit(SubmittedFrom::NotFromForm, submitter);
+        return Ok(());
```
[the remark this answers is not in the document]
muodov: changed it. Implicit returns are hard to get used to :)

components/script/dom/htmlformelement.rs
muodov: done, for the whole function (numbers were off, I suppose the spec changed at some point)

components/script/dom/htmlformelement.rs
muodov: I didn't change this part, I merely added indentation. Looking at the code you mentioned, I got an impression that static validation handles only elements inside the form, whereas this line fires an event on the form itself. I might be wrong about this, would you mind double-checking?

components/script/dom/htmlformelement.rs
muodov: done

components/script/dom/htmlformelement.rs
muodov: ok, done

components/script/dom/htmlformelement.rs
muodov: this line handles both type=image and type=submit, I added the links

components/script/dom/htmlformelement.rs
muodov: I believe this is https://html.spec.whatwg.org/multipage/#attr-button-type-submit-state, added it

muodov (commented):
@gterzian Thanks for an in-depth review! I believe I have addressed all the comments, would you mind checking again?

components/script/dom/htmlformelement.rs
gterzian: Ok. You're right about the different event targets, and I actually don't see where in the spec the event is fired on the element as a whole, however if this isn't an actual change, let's leave it for now...

gterzian (approved):
Awesome work! Please squash into one commit and then it should be ready to go...

components/script/dom/htmlformelement.rs
gterzian: Actually I think I've found where the event should be fired on the element as a whole, and this line appears incorrect, so I've filed a follow-up at https://github.com/servo/servo/issues/27133

gterzian: Actually I think I've found where the event should be fired on the element as a whole, and this line appears incorrect, so I've filed a follow-up at https://github.com/servo/servo/issues/27133

gterzian (commented):
The event-firing does appear redundant at that line, and I think it's a good follow-up, since it's not an actual change of this PR.

muodov:
@gterzian I squashed the commits 👍 I'm happy to work on #27133, it does look like a good follow-up indeed.
Just for my understanding, is it a requirement to squash commits in PRs? The [contribution guide](https://github.com/servo/servo/blob/master/CONTRIBUTING.md#pull-request-checklist) doesn't say this, should it be updated?

gterzian:
> Just for my understanding, is it a requirement to squash commits in PRs? The contribution guide doesn't say this, should it be updated?

If the PR includes different pieces of work, then it could have multiple commits, so I wouldn't say it's a requirement to always squash. The Github workflow, linked to from the contribution guide, goes into this a bit more I believe.

gterzian:
Thanks!

@bors-servo r+

gterzian:
> I'm happy to work on #27133, it does look like a good follow-up indeed.

Sounds like a plan!

gterzian:
```
0 matches
  ▶ Unexpected subtest result in /html/dom/idlharness.https.html?include=HTML.*:
  └ PASS [expected FAIL] HTMLFormElement interface: operation requestSubmit(optional HTMLElement?)
  ▶ Unexpected subtest result in /html/dom/idlharness.https.html?include=HTML.*:
  └ PASS [expected FAIL] HTMLFormElement interface: document.createElement("form") must inherit property "requestSubmit(optional HTMLElement?)" with the proper type
  ▶ Unexpected subtest result in /html/dom/idlharness.https.html?include=HTML.*:
  └ PASS [expected FAIL] HTMLFormElement interface: calling requestSubmit(optional HTMLElement?) on document.createElement("form") with too few arguments must throw TypeError

```

Great, some more test that require updating to PASS.

PR #27106, opened 2020-06-29
[UWP] Better default url management

### LINKED ISSUE #27046: Changing DEFAULT_URL doesn't override locally-stored preference
DEFAULT_URL is only used if there's no shell.homepage preference present in the local preferences in [this code](https://github.com/servo/servo/blob/ed46f00497f4bab79717b0c6a1da08d068b5cfcf/support/hololens/ServoApp/ServoControl/Servo.cpp#L143-L145). This is really confusing when modifying DEFAULT_URL during development!

paulrouget:
Fix #27046


jdm:
@bors-servo r+
Nice solution;

Issue #27039, opened 2020-06-22
Intermittent segfault in webrender build script (signal: 11, SIGSEGV: invalid memory reference)


jdm:
```
Caused by:
  process didn't exit successfully: `/repo/target/debug/build/webrender-0c9654493a04f959/build-script-build` (signal: 11, SIGSEGV: invalid memory reference)
--- stdout
cargo:rerun-if-changed=res
cargo:rerun-if-changed=res/cs_gradient.glsl
cargo:rerun-if-changed=res/prim_shared.glsl
cargo:rerun-if-changed=res/pf_vector_cover.glsl
cargo:rerun-if-changed=res/pls_init.glsl
cargo:rerun-if-changed=res/cs_line_decoration.glsl
cargo:rerun-if-changed=res/composite.glsl
cargo:rerun-if-changed=res/ps_text_run.glsl
cargo:rerun-if-changed=res/gpu_cache.glsl
cargo:rerun-if-changed=res/base.glsl
cargo:rerun-if-changed=res/cs_blur.glsl
cargo:rerun-if-changed=res/brush_opacity.glsl
cargo:rerun-if-changed=res/ellipse.glsl
cargo:rerun-if-changed=res/cs_clip_image.glsl
cargo:rerun-if-changed=res/brush_mix_blend.glsl
cargo:rerun-if-changed=res/brush.glsl
cargo:rerun-if-changed=res/shared_other.glsl
cargo:rerun-if-changed=res/transform.glsl
cargo:rerun-if-changed=res/brush_conic_gradient.glsl
cargo:rerun-if-changed=res/brush_solid.glsl
cargo:rerun-if-changed=res/yuv.glsl
cargo:rerun-if-changed=res/ps_clear.glsl
cargo:rerun-if-changed=res/cs_clip_box_shadow.glsl
cargo:rerun-if-changed=res/pf_vector_stencil.glsl
cargo:rerun-if-changed=res/clip_shared.glsl
cargo:rerun-if-changed=res/cs_border_segment.glsl
cargo:rerun-if-changed=res/cs_svg_filter.glsl
cargo:rerun-if-changed=res/brush_radial_gradient.glsl
cargo:rerun-if-changed=res/cs_border_solid.glsl
cargo:rerun-if-changed=res/brush_linear_gradient.glsl
cargo:rerun-if-changed=res/brush_image.glsl
cargo:rerun-if-changed=res/ps_split_composite.glsl
[...]

PR #27114, opened 2020-06-29
[UWP] Key events

paulrouget:
This is the initial work required for proper keyboard events.
The text controller implementation is very basic, just enough to show the virtual keyboard when it's needed, and have basic key events.


ports/libsimpleservo/capi/src/lib.rs
```diff
@@ -717,6 +718,34 @@ pub extern "C" fn click(x: f32, y: f32) {
     });
 }
 
+#[no_mangle]
+pub extern "C" fn key_down(name: *const c_char) {
+    catch_any_panic(|| {
+        debug!("key_down");
+        let name = unsafe { CStr::from_ptr(name) };
+        let key = Key::from_str(&name.to_str().expect("Can't read string"));
```
jdm: Do we really want to panic here, or could we just warn and return instead?

ports/libsimpleservo/capi/src/lib.rs
```diff
@@ -717,6 +718,34 @@ pub extern "C" fn click(x: f32, y: f32) {
...
+        }
+    })
+}
+
+#[no_mangle]
+pub extern "C" fn key_up(name: *const c_char) {
+    catch_any_panic(|| {
+        debug!("key_up");
+        let name = unsafe { CStr::from_ptr(name) };
+        let key = Key::from_str(&name.to_str().expect("Can't read string"));
```
jdm, line 736: Same question.

support/hololens/ServoApp/ServoControl/ServoControl.cpp
```diff
@@ -89,6 +89,48 @@ void ServoControl::OnLoaded(IInspectable const &, RoutedEventArgs const &) {
...
+    e.Request().Text(L"");
+    CoreTextRange sel;
+    sel.StartCaretPosition = 0;
+    sel.EndCaretPosition = 0;
+    e.Request().Range() = sel;
+  });
+
+  mEditContext->TextUpdating([=](const auto &, const auto &e) {
+    auto keystr = *hstring2char((e.Text()));
+    RunOnGLThread([=] { mServo->KeyDown(keystr); });
```
jdm, line 108: This is scary due to #27115. We should dereference keystr inside of the closure, rather than storing a pointer to the inner string value and using that.

support/hololens/ServoApp/ServoControl/Keys.h
```diff
@@ -0,0 +1,130 @@
+using namespace winrt::Windows::System;
+
+namespace winrt {
+std::optional<const char *> KeyToString(Windows::System::VirtualKey key) {
```
jdm, line 4: What happens for letters and digits?

support/hololens/ServoApp/ServoControl/ServoControl.cpp
```diff
@@ -89,6 +89,48 @@ void ServoControl::OnLoaded(IInspectable const &, RoutedEventArgs const &) {
...
+    if (keystr.has_value()) {
+      auto keyname = *keystr;
+      RunOnGLThread([=] { mServo->KeyDown(keyname); });
+    }
+  });
+  PreviewKeyUp([=](const auto &, const auto &e) {
+    auto keystr = KeyToString(e.Key());
+    if (keystr.has_value()) {
+      auto keyname = *keystr;
+      RunOnGLThread([=] { mServo->KeyUp(keyname); });
```
jdm, line 129: Same concern about #27115, even though this is a std::optional value. We should dereference keyname inside the closure to avoid references to memory that isn't owned by the closure.

jdm (changes requested):

support/hololens/ServoApp/ServoControl/Keys.h
paulrouget: Non-printable keys are handled by the `PreviewKeyDown` event.
Printable keys are handled by the `TextUpdating` event.

paulrouget: To clarify what's going on here: we either send "raw" key events to servo. For example: `shift` key, then `a` key. Then it would be up to Servo combine those to print `A`, not just `a`. This work is already done by the embedder, so we send what needs to be printed directly.

paulrouget:
I will address `*hstring2char` issue in another PR.

Addressed your comments.

jdm:
@bors-servo retry
* #27039

PR #27126, opened 2020-06-30
Implement GPUBuffer.getMappedRange()

kunalmohan:
The new fields added to `GPUBuffer` are-
1. `mapping`- This is the buffer data. This is not directly exposed to user and is refcounted to keep track of the number of `ArrayBuffer`s that point to the content in it.
2. `mapping_range`- The range of mapped portion of the GPUBuffer.
3. `mapped_ranges`- Ranges that the various `ArrayBuffer`s expose to the user. Used for validation of `getMappedRange()`
4. `js_buffers`- Actual `ArrayBuffer`s that expose the data to the user. They are detached on unmap.
5. `map_promise`- Promise that represents the pending mapping.

This PR also changes the order of the exit of threads. WebGL thread is responsible for sending `Exit` message to WebRender and therefore should be exited after WebGPU threads.

r?@kvark @jdm

---

### DIFF
--- a/components/script/dom/gpubuffer.rs
+++ b/components/script/dom/gpubuffer.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use crate::dom::bindings::cell::{DomRefCell, RefCell};
+use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::{GPUBufferMethods, GPUSize64};
 use crate::dom::bindings::codegen::Bindings::GPUMapModeBinding::GPUMapModeConstants;
 use crate::dom::bindings::error::{Error, Fallible};
@@ -19,7 +19,7 @@ use dom_struct::dom_struct;
 use js::jsapi::DetachArrayBuffer;
 use js::jsapi::NewExternalArrayBuffer;
 use js::jsapi::{Heap, JSObject};
-use std::cell::Cell;
+use std::cell::{Cell, RefCell};
 use std::ffi::c_void;
 use std::ops::Range;
 use std::ptr::NonNull;
@@ -41,6 +41,17 @@ pub enum GPUBufferState {
     Destroyed,
 }
 
+#[derive(JSTraceable, MallocSizeOf)]
+pub struct GPUBufferMapInfo {
+    #[ignore_malloc_size_of = "Rc"]
+    pub mapping: Rc<RefCell<Vec<u8>>>,
+    pub mapping_range: Range<u64>,
+    pub mapped_ranges: Vec<Range<u64>>,
+    #[ignore_malloc_size_of = "defined in mozjs"]
+    pub js_buffers: Vec<Box<Heap<*mut JSObject>>>,
+    pub map_mode: Option<u32>,
+}
+
 #[dom_struct]
 pub struct GPUBuffer {
     reflector_: Reflector,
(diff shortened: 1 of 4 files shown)


components/script/dom/gpubuffer.rs
```diff
@@ -142,51 +153,34 @@ impl GPUBufferMethods for GPUBuffer {
...
-                            "Could not find ArrayBuffer of Mapped buffer ({:?})",
-                            self.buffer.0
-                        );
-                    },
-                };
+                if let Err(e) = self.channel.0.send(WebGPURequest::UnmapBuffer {
+                    buffer_id: self.id().0,
+                    array_buffer: self.mapping.borrow().as_ref().unwrap().clone(),
+                    is_map_read: self.map_mode.get() == Some(GPUMapModeConstants::READ),
+                    offset: self.mapping_range.borrow().as_ref().unwrap().start,
```
kvark, line 160: let's not repeat ourselves 3 times here :)

components/script/dom/gpubuffer.rs
```diff
@@ -75,8 +83,11 @@ impl GPUBuffer {
             device,
             buffer,
             mapping,
+            mapped_ranges: DomRefCell::new(None),
```
kvark, line 86: do all of these need to be separate? Can we do `DomRefCell<Option<SomeStructure>>` with all the fields in the `SomeStructure`?

components/script/dom/gpubuffer.rs
```diff
@@ -75,8 +83,11 @@ impl GPUBuffer {
...
+            js_buffers: DomRefCell::new(None),
+            map_promise: DomRefCell::new(None),
             size,
-            mapping_range: DomRefCell::new(mapping_range),
+            mapping_range: DomRefCell::new(Some(mapping_range)),
```
kvark, line 90: Why do we initialize this with `Some`?

components/script/dom/gpubuffer.rs
```diff
@@ -256,10 +255,72 @@ impl GPUBufferMethods for GPUBuffer {
...
 
+    /// https://gpuweb.github.io/gpuweb/#dom-gpubuffer-getmappedrange
+    #[allow(unsafe_code)]
+    fn GetMappedRange(
+        &self,
+        cx: JSContext,
+        offset: GPUSize64,
+        size: GPUSize64,
+    ) -> Fallible<NonNull<JSObject>> {
+        if self.mapped_ranges.borrow().is_none() {
```
kvark, line 271: It seems strange to have this lazy check here. We should know exactly when this function is valid to call (either after mapping, or in a buffer that's created with mapping), so those earlier stages need to prepare the state, and code here doesn't need to be lazy

components/script/dom/gpubuffer.rs
```diff
@@ -256,10 +255,72 @@ impl GPUBufferMethods for GPUBuffer {
...
+            act_size % RANGE_SIZE_ALIGN_MASK == 0 &&
+            offset >= self.mapping_range.borrow().as_ref().unwrap().start &&
+            offset + act_size <= self.mapping_range.borrow().as_ref().unwrap().end;
+        valid &= self
+            .mapped_ranges
+            .borrow()
+            .as_ref()
+            .unwrap()
+            .iter()
+            .all(|range| range.start > offset + act_size || range.end < offset);
```
kvark, line 292: these should be `>=` and `<=` correspondingly

components/script/dom/gpubuffer.rs
```diff
@@ -256,10 +255,72 @@ impl GPUBufferMethods for GPUBuffer {
...
+        unsafe extern "C" fn free_func(_contents: *mut c_void, free_user_data: *mut c_void) {
+            let _ = Rc::from_raw(free_user_data as _);
+        }
+
+        let array_buffer = unsafe {
+            NewExternalArrayBuffer(
+                *cx,
+                act_size as usize,
+                self.mapping.borrow_mut().as_mut().unwrap()
+                    [offset as usize..(offset + act_size) as usize]
```
kvark, line 306: we repeat `offset + act_size` quite a lot in this function

kvark (commented):
Looks reasonable overall! Noted a few small things

components/script/dom/gpubuffer.rs
kunalmohan: Sorry, it's a mistake. I'll correct it.

components/script/dom/gpubuffer.rs
```diff
@@ -255,11 +244,65 @@ impl GPUBufferMethods for GPUBuffer {
...
+            let _ = Rc::from_raw(free_user_data as _);
+        }
+
+        let array_buffer = unsafe {
+            NewExternalArrayBuffer(
+                *cx,
+                (m_end - offset) as usize,
+                m_info.mapping.borrow_mut()[offset as usize..m_end as usize].as_mut_ptr() as _,
+                Some(free_func),
+                Rc::into_raw(m_info.mapping.clone()) as _,
```
kvark, line 296: I was looking at this piece a bit. So this would keep the memory for the ArrayBuffer around until it's GC-ed.
I think it's totally fine to land in this stage. Just wanted to note that for the end game, we probably want to "detach" the array buffer on `unmap()`, so that any access to its contents would be invalid and trigger an exception. Perhaps, we could leave a comment somewhere about this TODO

kvark (approved):
Thank you, looks great!

components/script/dom/gpubuffer.rs
kunalmohan: We already detach the actual array buffers(which are the only ones exposed to the user) on unmap. We also drop the `Rc` pointer for this array buffer in during dispose `free_func` and set the `map_info` to `None` (so we drop all `Rc` pointers on unmap). What else are we missing here? Should the `contents` be set to `ptr::null()` in `free_func`?

PR #27154, opened 2020-07-02
WebGPU update presentation data asynchronously

kunalmohan:
This PR aims to make updating Webrender presentation data non-blocking.
A callback is passed to the `buffer_map_async()` fn wherein the data is read and a message sent to the server itself to write the data into the shared `PresentationData` struct object.

r?@kvark

---

### DIFF
--- a/components/webgpu/lib.rs
+++ b/components/webgpu/lib.rs
@@ -13,7 +13,7 @@ pub mod identity;
 use arrayvec::ArrayVec;
 use euclid::default::Size2D;
 use identity::{IdentityRecyclerFactory, WebGPUMsg};
-use ipc_channel::ipc::{self, IpcReceiver, IpcSender};
+use ipc_channel::ipc::{self, IpcReceiver, IpcSender, IpcSharedMemory};
 use malloc_size_of::{MallocSizeOf, MallocSizeOfOps};
 use serde::{Deserialize, Serialize};
 use servo_config::pref;
@@ -51,7 +51,7 @@ pub enum WebGPUResponse {
         queue_id: WebGPUQueue,
         _descriptor: wgt::DeviceDescriptor,
     },
-    BufferMapAsync(Vec<u8>),
+    BufferMapAsync(IpcSharedMemory),
 }
 
 pub type WebGPUResponseResult = Result<WebGPUResponse, String>;
@@ -145,7 +145,7 @@ pub enum WebGPURequest {
     },
     CreateSwapChain {
         device_id: id::DeviceId,
-        buffer_id: id::BufferId,
+        buffer_ids: ArrayVec<[id::BufferId; 5]>,
         external_id: u64,
         sender: IpcSender<webrender_api::ImageKey>,
         image_desc: webrender_api::ImageDescriptor,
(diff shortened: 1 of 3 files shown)


kunalmohan:
I am facing one problem here. Inside the `buffer_map_async()` callback, sender doesn't send the `UpdateWebRenderData` message. There's no panic or error too. The control doesn't get past that code block responsible to send that message.

(There are some debug logs here which I'll remove once this problem gets solved)

components/webgpu/lib.rs
```diff
@@ -844,8 +856,36 @@ impl<'a> WGPU<'a> {
                                 size = present_data.size;
                                 device_id = present_data.device_id;
                                 queue_id = present_data.queue_id;
-                                buffer_id = present_data.buffer_id;
                                 buffer_stride = present_data.buffer_stride;
+                                buffer_id = if present_data.available_buffer_ids.len() > 0 {
```
kvark, line 860: nit: could do `if let Some(id) = present_data.available_buffer_ids.pop() {..} else if let Some(id) == present_data.unassigned_buffer_ids.pop() {..} else {..}`

components/webgpu/lib.rs
```diff
@@ -950,6 +1003,42 @@ impl<'a> WGPU<'a> {
...
+                            txn.update_image(
+                                present_data.image_key,
+                                present_data.image_desc,
+                                present_data.image_data.clone(),
+                                &webrender_api::DirtyRect::All,
+                            );
+                            self.webrender_api
+                                .send_transaction(self.webrender_document, txn);
+                            let mut i = 0;
+                            for b_id in present_data.queued_buffer_ids.iter() {
```
kvark, line 1027: nit: could just do `present_data.queued_buffer_ids.retain(|&id| id != buffer_id);`

components/webgpu/lib.rs
```diff
@@ -893,43 +933,56 @@ impl<'a> WGPU<'a> {
...
                                 BufferMapAsyncStatus::Success => {
-                                    debug!("Buffer Mapped");
+                                    println!(
+                                        "Presentation map success {:?} {:?}",
+                                        info.buffer_id, info.sender
+                                    );
+                                    let global = &info.global;
+                                    let data_pt = gfx_select!(info.buffer_id =>
+                                        global.buffer_get_mapped_range(info.buffer_id, 0, None));
+                                    let data = slice::from_raw_parts(data_pt, info.size).to_vec();
```
kvark, line 962: we can unmap it right after this call, any reason to defer it to later?

kvark (approved):
Looks reasonable

kunalmohan:
The error was probably due to the data-carrying capacity of IpcSender. Reading the data in `UpdateWebRenderData` instead of sending it over from the async callback solved the issue.

I have also edited the buffer data exchange between content process and server to use `IpcSharedMemory` instead of `Vec` in `BufferMapAsync` and `UnmapBuffer`.

components/webgpu/lib.rs
```diff
@@ -950,6 +991,38 @@ impl<'a> WGPU<'a> {
...
+                        external_id,
+                    } => {
+                        let global = &self.global;
+                        let data_pt = gfx_select!(buffer_id =>
+                            global.buffer_get_mapped_range(buffer_id, 0, None));
+                        let data = unsafe { slice::from_raw_parts(data_pt, buffer_size) }.to_vec();
+                        if let Some(present_data) =
+                            self.wgpu_image_map.lock().unwrap().get_mut(&external_id)
+                        {
+                            present_data.data = data;
```
kvark, line 1006: This may be a bit of a problem. Does WR know about your buffer stride?
If not, we should be "re-striding" the data as we read it from the buffer into `data` vector.

kvark (changes requested):
Much nicer! One concern though

components/webgpu/lib.rs
kunalmohan: Yes WR knows about the buffer stride. I have tested this and it works fine.

kvark (approved):

jdm:
@bors-servo retry
* #24622

jdm:
@bors-servo retry
* #25522 #23290

PR #27157, opened 2020-07-03
Enable stdout logging for all UWP builds

### LINKED ISSUE #27156: Enable stdout/stderr logging by default on UWP
We currently disable logging in release builds unless the [FirefoxRealityLogStdout](https://github.com/servo/servo/blob/master/support/hololens/ServoApp/ServoControl/Servo.cpp#L226) environment variable is set. There's no good way to set such a variable on a HoloLens 2 without recompiling the app, which means it's effectively impossible to get logging out of an arbitrary user's build and they have to be provided a custom build instead. This sucks; we should log by default instead so they can always grab it from the app's local files if they encounter a problem. This will give us warn!/error! output and any println/eprintln, such as panic messages.

jdm:
Fixes #27156.

### DIFF
--- a/support/hololens/ServoApp/ServoControl/Servo.cpp
+++ b/support/hololens/ServoApp/ServoControl/Servo.cpp
@@ -221,30 +221,20 @@ Servo::Servo(hstring args, GLsizei width, GLsizei height,
 
   sServo = this; // FIXME;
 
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Getting this in before the next nightly.

PR #27168, opened 2020-07-03
Add stub isContextLost API.

### LINKED ISSUE #27116: WebGLRenderingContext. isContextLost() is undefined
The method WebGLRenderingContext. isContextLost() doesn't exist in Firefox Reality on Hololens 2.
Every Unity 3D WebGL build use this method.
I am using this workaround to make it work : https://github.com/Rufus31415/Simple-WebXR-Unity/blob/28331fc890e316e13401618a8e0da1e84bad7a39/Assets/SimpleWebXR.jspre#L6

jdm:
We don't support losing contexts yet, so it's not lying to say that the context is never lost.

---
- [x] These changes fix #27116


jdm:
@bors-servo retry
* #24622

PR #27169, opened 2020-07-03
Mach test cargo target dir

alarsyo:
---

- [X] These changes do not require tests because they're about tooling

This change allows launching tests this way:

```sh
CARGO_TARGET_DIR=other-dir ./mach test-wpt ...
```

The build and check commands already check for the presence of the `CARGO_TARGET_DIR` variable, but that wasn't the case for wpt tests. This resulted in an error about the `servo` executable not being found in `target/debug/servo`.

### DIFF
--- a/tests/wpt/run.py
+++ b/tests/wpt/run.py
@@ -4,6 +4,7 @@
 
 import multiprocessing
 import os
+from os import path
 import sys
 import mozlog
 import grouping_formatter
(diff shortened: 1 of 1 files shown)


alarsyo:
Woops, didn't create the branch from master. Let me rebase this.

jdm:
@bors-servo r+
Thanks for finding and fixing this!

tests/wpt/run.py
```diff
@@ -72,7 +73,10 @@ def set_defaults(kwargs):
         bin_name = "servo"
         if sys.platform == "win32":
             bin_name += ".exe"
-        bin_path = servo_path("target", bin_dir, bin_name)
+        if os.environ["CARGO_TARGET_DIR"]:
```
jdm, line 76: Looks like we need `os.environ.get("CARGO_TARGET_DIR")` instead.

alarsyo: My bad, didn't test without the variable :) should be fixed!

jdm:
@bors-servo retry
* #27039

Issue #24066, opened 2019-08-27
Intermittent CRASH in /html/semantics/embedded-content/the-iframe-element/iframe_sandbox_popups_nonescaping-3.html


CYBAI:
```
  ▶ CRASH [expected TIMEOUT] /html/semantics/embedded-content/the-iframe-element/iframe_sandbox_popups_nonescaping-3.html
  │ 
  │ _RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.
  │ 
  │ 
  │ 
  │ Failed to get browsing context info from constellation.: Io(Custom { kind: ConnectionReset, error: "No senders exist for this port." }) (thread ScriptThread PipelineId { namespace_id: PipelineNamespaceId(2), index: PipelineIndex(3) }, at src/libcore/result.rs:1084)
  │ stack backtrace:
  │    0: __ZN9backtrace9backtrace5trace17h6bd49c37c3e2a608E (0x1101fb94b)
  │    1: __ZN9backtrace7capture9Backtrace3new17h3ec5d87241e0727cE (0x1101fab37)
  │    2: __ZN5servo4main28_$u7b$$u7b$closure$u7d$$u7d$17h14212de9f0afcb9aE (0x10d3e790b)
  │    3: __ZN3std9panicking20rust_panic_with_hook17he24eb6798ad54c68E (0x110382131)
  │    4: __ZN3std9panicking18continue_panic_fmt17hc41eddc40bb93eb6E (0x110381b9c)
  │    5: _rust_begin_unwind (0x110381a98)
  │    6: __ZN4core9panicking9panic_fmt17h81801588e8006fa1E (0x1103a22ce)
  │    7: __ZN4core6result13unwrap_failed17h759b9cbabad88910E (0x1103a23a8)
  │    8: __ZN6script13script_thread12ScriptThread19remote_window_proxy17h3de40eb87ee4108bE (0x10e0e9da4)
  │    9: __ZN6script13script_thread12ScriptThread29handle_msg_from_constellation17h8e434629980b4985E (0x10e0daba0)
  │   10: __ZN6script13script_thread12ScriptThread11handle_msgs17h5b6b659e4ca9dbb0E.llvm.10939716038701058040 (0x10e0d8403)
  │   11: __ZN14profile_traits3mem12ProfilerChan25run_with_memory_reporting17h48b7134700a222b2E (0x10e200157)
[...]

PR #27173, opened 2020-07-04
Implement GPUQueue write methods and update wgpu-core

kunalmohan:
Implements `GPUQueue.writeBuffer` and `GPUQueue.writeTexture`.

r?@kvark

---

### DIFF
--- a/components/script/dom/gpuqueue.rs
+++ b/components/script/dom/gpuqueue.rs
@@ -3,15 +3,24 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use crate::dom::bindings::cell::DomRefCell;
+use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUSize64;
 use crate::dom::bindings::codegen::Bindings::GPUQueueBinding::GPUQueueMethods;
+use crate::dom::bindings::codegen::Bindings::GPUTextureBinding::{
+    GPUExtent3D, GPUOrigin3D, GPUTextureCopyView, GPUTextureDataLayout,
+};
+use crate::dom::bindings::error::{Error, Fallible};
 use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::DOMString;
 use crate::dom::globalscope::GlobalScope;
-use crate::dom::gpubuffer::GPUBufferState;
+use crate::dom::gpubuffer::{GPUBuffer, GPUBufferState};
 use crate::dom::gpucommandbuffer::GPUCommandBuffer;
+use crate::dom::gpudevice::{convert_texture_size_to_dict, convert_texture_size_to_wgt};
 use dom_struct::dom_struct;
-use webgpu::{WebGPU, WebGPUQueue, WebGPURequest};
+use ipc_channel::ipc::IpcSharedMemory;
+use js::rust::CustomAutoRooterGuard;
+use js::typedarray::ArrayBuffer;
+use webgpu::{wgpu::command as wgpu_com, wgt, WebGPU, WebGPUQueue, WebGPURequest};
 
 #[dom_struct]
 pub struct GPUQueue {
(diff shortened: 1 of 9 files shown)


components/script/dom/gpuqueue.rs
```diff
@@ -69,4 +78,105 @@ impl GPUQueueMethods for GPUQueue {
...
+        buffer_offset: GPUSize64,
+        data: CustomAutoRooterGuard<ArrayBuffer>,
+        data_offset: GPUSize64,
+        size: Option<GPUSize64>,
+    ) -> Fallible<()> {
+        let bytes = data.to_vec();
+        let content_size = if let Some(s) = size {
+            s
+        } else {
+            bytes.len() as GPUSize64 - data_offset
```
kvark, line 96: this should be taking the buffer size, not the bytes length

components/script/dom/gpuqueue.rs
```diff
@@ -69,4 +78,105 @@ impl GPUQueueMethods for GPUQueue {
...
+        };
+        let valid = data_offset + content_size <= bytes.len() as u64 &&
+            buffer.state() == GPUBufferState::Unmapped &&
+            content_size % 4 == 0 &&
```
kvark, line 100: please use the actual constant of `wgc::COPY_BUFFER_ALIGNMENT`

components/script/dom/gpuqueue.rs
```diff
@@ -69,4 +78,105 @@ impl GPUQueueMethods for GPUQueue {
...
+            buffer_offset % 4 == 0 &&
+            buffer.is_valid();
+
+        if !valid {
+            return Err(Error::Operation);
+        }
+
+        let final_data = IpcSharedMemory::from_bytes(
+            &bytes[data_offset as usize..(data_offset + content_size) as usize],
```
kvark, line 109: so this should throw an exception if there is not enough data in the array buffer

components/script/dom/gpuqueue.rs
```diff
@@ -69,4 +78,105 @@ impl GPUQueueMethods for GPUQueue {
...
+
+        if !valid {
+            return Err(Error::Operation);
+        }
+
+        let texture_cv = wgpu_com::TextureCopyView {
+            texture: destination.texture.id().0,
+            mip_level: destination.mipLevel,
+            origin: match destination.origin {
+                GPUOrigin3D::RangeEnforcedUnsignedLongSequence(ref v) => wgt::Origin3d {
```
kvark, line 143: there needs to be a helper function for this match, I imagine this is not the only place where it's used

kvark (commented):
General implementation is all good. There is one spot that needs to be a bit reworked.

components/script/dom/gpuqueue.rs
kunalmohan: The spec mentions it to be `data.byteLength` (https://gpuweb.github.io/gpuweb/#dom-gpuqueue-writebuffer). Should that be corrected?

components/script/dom/gpuqueue.rs
kunalmohan: A check for this is there on line no. 98
```
let valid = data_offset + content_size <= bytes.len() as u64
```
Or am I missing something?

components/script/dom/gpuqueue.rs
kunalmohan: This is the only use at the moment. I was thinking of creating the function once we come across more of its uses.

components/script/dom/gpuqueue.rs
kvark: Oh wow, that's a surprise to me :) Thank you for correcting me!

kvark (approved):
Looks like everything is resolved

components/script/dom/gpuqueue.rs
kvark: Filed https://github.com/gpuweb/gpuweb/pull/903 to clarify

kunalmohan: Should I change it to buffer size then?

CYBAI:
@bors-servo retry

- #24622,  #24066, #26895

CYBAI:
@bors-servo retry

- #24066

PR #27174, opened 2020-07-04
Do not fire `invalid` event on form elements

### LINKED ISSUE #27133: Improve spec compliance of validity check
This line https://github.com/servo/servo/pull/27100/files#diff-db2031d1f23ecf1a921fc100cfed753dR677

appears incorrect, since the spec only says to fire events on the fields that are invalid, via step 6.1 of https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#statically-validate-the-constraints, which is called into by https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#concept-form-submit

An event is fired on the form element as a whole only as part of the ["check validity steps"](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#check-validity-steps), and those are called into from other places(for example the `checkValidity` method). 

So the implementation of `checkValidaty` could also be updated to actually fire an event.

see `components/script/dom/htmlformelement.rs`.

muodov:
This is a follow-up to my [previous PR](https://github.com/servo/servo/pull/27100) suggested by @gterzian.

`invalid` event on the `<form>` element has been [recently removed](https://github.com/whatwg/html/issues/4626) from the spec.

---

- [x] These changes fix #27133

- [x] [WPT test](https://github.com/servo/servo/blob/master/tests/wpt/web-platform-tests/html/semantics/forms/form-submission-0/historical.window.js) marked as passing

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -673,8 +673,6 @@ impl HTMLFormElement {
             // Step 6.3
             if !submitter.no_validate(self) {
                 if self.interactive_validation().is_err() {
-                    // TODO: Implement event handlers on all form control elements
-                    self.upcast::<EventTarget>().fire_event(atom!("invalid"));
                     self.firing_submission_events.set(false);
                     return;
                 }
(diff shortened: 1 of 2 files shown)


gterzian:
Thanks! Re closing the issue, we might want to keep it open regarding `checkValidity` and others, depending on the outcome of the discussion.

@bors-servo r+

PR #27183, opened 2020-07-06
Add `.servobuild` configuration for the choice of media stack

SimonSapin:


SimonSapin:
```
error: could not remove 'setup' file: 'C:\Users\task_1594045570\.cargo\bin/rustup-init.exe'
error: caused by: Access is denied. (os error 5)
```

:thinking:

SimonSapin:
The same happen on all three Windows builds. I’ve cycled the workers, let’s see if this happens again.

@bors-servo retry

Issue #27161, opened 2020-07-03
Intermittent crash in pulseaudio during media tests (ssertion '(v = fcntl(fd, F_GETFD, 0)) >= 0' failed at pulsecore/core-util.c:253, function pa_make_fd_cloexec(). Aborting.)


jdm:
```
  ▶ CRASH [expected OK] /html/semantics/embedded-content/media-elements/event_timeupdate.html
  │ 
  │ error: XDG_RUNTIME_DIR not set in the environment.
  │ libEGL warning: No hardware driver found, falling back to software rendering
  │ Cannot connect to server socket err = No such file or directory
  │ Cannot connect to server request channel
  │ Assertion '(v = fcntl(fd, F_GETFD, 0)) >= 0' failed at pulsecore/core-util.c:253, function pa_make_fd_cloexec(). Aborting.
  │ Redirecting call to abort() to mozalloc_abort
  │ 
  │ Stack trace
  │ jack server is not running or cannot be started
  │ JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
  │ JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
  │    0: servo::backtrace::print
  │    1: servo::install_crash_handler::handler
  │    2: _ZL15WasmTrapHandleriP9siginfo_tPv
  │              at /root/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/fc8b315/mozjs/js/src/wasm/WasmSignalHandlers.cpp:970
  │    3: <unknown>
  │    4: _ZL22AnnotateMozCrashReasonPKc
  │              at /repo/target/release/build/mozjs_sys-495f3fdf82bc556d/out/build/dist/include/mozilla/Assertions.h:42
  │       mozalloc_abort
  │              at /root/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/fc8b315/mozjs/memory/mozalloc/mozalloc_abort.cpp:33
  │    5: abort
  │              at /root/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/fc8b315/mozjs/memory/mozalloc/mozalloc_abort.cpp:82
  │    6: pa_make_fd_cloexec
  │    7: pa_fopen_cloexec
  │    8: pa_machine_id
  │    9: pa_get_runtime_dir
  │   10: <unknown>
  │   11: <unknown>
  │   12: pa_context_connect
  │   13: <unknown>
[...]

Issue #26726, opened 2020-05-31
Intermittent TIMEOUT /xhr/send-timeout-events.htm


gterzian:
https://community-tc.services.mozilla.com/tasks/U8bhcYXfSjC-LAyssVkcwA/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FU8bhcYXfSjC-LAyssVkcwA%2Fruns%2F0%2Fartifacts%2Fpublic%2Ffiltered-wpt-errorsummary.log

```
0 matches
1 unexpected results that are NOT known-intermittents:
  ▶ TIMEOUT [expected OK] /xhr/send-timeout-events.htm
  │ 
  │ error: XDG_RUNTIME_DIR not set in the environment.
  └ libEGL warning: No hardware driver found, falling back to software rendering

```

It seems to pass locally on Mac consistently.

jdm:
We have a new intermittent issue that affects WPT tests on linux that looks like unexpected timeouts. I'm going to repurpose this issue to track that.

CYBAI:
is #26225 possibly related?

jdm:
I don't believe so; the new linux timeouts have become much more frequent in the past 1.5 weeks.

gterzian:
How about we also identify individual tests as they are noted, to make it easier to show up in search by issue, and also perhaps to allow a clearer picture of affected areas to emerge?

I saw this at 

```
  ▶ TIMEOUT [expected OK] /fetch/cross-origin-resource-policy/image-loads.html
  │ 
  │ error: XDG_RUNTIME_DIR not set in the environment.
  └ libEGL warning: No hardware driver found, falling back to software rendering
```

jdm:
```
  ▶ CRASH [expected FAIL] /css/CSS2/floats-clear/clear-applies-to-015.xht
  │ 
  │ error: XDG_RUNTIME_DIR not set in the environment.
  │ libEGL warning: No hardware driver found, falling back to software rendering
  └ Shutting down the Constellation after generating an output file or exit flag specified
```

jdm:
Another /fetch/cross-origin-resource-policy/image-loads.html.

jdm:
```
  ▶ TIMEOUT [expected OK] /css/css-flexbox/position-absolute-013.html
  │ 
  │ error: XDG_RUNTIME_DIR not set in the environment.
[...]

PR #27186, opened 2020-07-06
Update webrender to crash-backtrace branch.

jdm:
This moves us to https://github.com/servo/webrender/tree/crash-backtrace, which adds a signal handler to the webrender build script to give us more information about the crash in #27039.


CYBAI:
@bors-servo retry

- #26726

jdm:
@bors-servo retry
* #27161 #26895

PR #27192, opened 2020-07-07
Net: in the absense of a request body, assume streaming it cannot fail

### LINKED ISSUE #27185: Many warnings about request stream confirmations when loading pages
For pages that make a bunch of network requests, if you look at the output of RUST_LOG=net::http_loader or RUST_LOG=net, you end up seeing a bunch of `WARN - net::http_loader - Failed to receive confirmation request was streamed without error.` The fact that everything appears to be working fine suggests that this is not actually a case we need to warn about. @gterzian What can we do to reduce the frequency of this message?

gterzian:
FIX https://github.com/servo/servo/issues/27185

---


gterzian:
@jdm r?

PR #27198, opened 2020-07-07
Add debug output for intermittently failing test.

jdm:
This should help narrow down the cause of #26895.


jdm:
@bors-servo retry
* #24622

Issue #25167, opened 2019-12-06
Implement CSS floats


SimonSapin:
Used on https://servo.org for two-column layout.

We already have logic for disabling parallelism in a block formatting context that contains any float. ~Drawing the rest of the owl~ actual layout implementation is missing as of this writing.

* https://drafts.csswg.org/css2/visuren.html#floats
* https://drafts.csswg.org/css2/visudet.html

pcwalton:
So I'm starting to look into this. I'll add questions as I get to them (cc @nox):

* Should there be a new `Float(FloatFragment)` enum variant in `Fragment`? In the case of `AbsoluteOrFixedPositionedFragment`, I'm guessing that the `hoisted_fragment` is the link from the hypothetical box up to its real ancestor fragment? If so, it would seem that at least inline float fragments would probably want a similar link. But why does the hypothetical box of absolute positioned elements need to be in the fragment tree at all?

SimonSapin:
Could `AbsoluteOrFixedPositionedFragment` be renamed `HoistedFragment` and used for floats as well? (With an appropriate doc-comment.)

Maybe @mrobinson can say more: my understanding of why we have hoisting with this indirection at all, effectively turning a tree into a DAG, is the tension between:

* For the purpose of positioning, the "main layout phase" wants to create a fragment tree where the coordinates of each node are *relative* to that of their parent, which represents their containing block.

  Layout has a lot of data dependencies that require things to be computed in a certain order. In some cases we need to determine the size of a block container for a given fragment, then lay out its subtree, then the compute its size based on the laid out content, then position the whole subtree based on some alignment property.

  Therefore coordinates in the fragment tree can’t be absolute without a `translate` method that involves a subtree traversal to patch things up, which I presume would be expensive. And fragments need to be "near" their containing block for the next phase to be able to compute absolute coordinates.

* For the purpose of determining the painting order, we build the stacking context tree as an intermediate between the fragment tree and the display list. This is what gets sorted by `z-index`. Elements that generate a stacking context form a sort of isolation boundary such that a subtree is treated as a unit for this sorting.

[...]

PR #27216, opened 2020-07-09
Implement core float layout for layout 2020

pcwalton:
Add an implementation of the core float and clear placement logic in layout 2020, not yet wired to the rest of layout.

This commit implements an object that handles the 10 rules in CSS 2.1:

https://www.w3.org/TR/CSS2/visuren.html#float-position

The implementation strategy is that of a persistent balanced binary search tree
of float bands. Binary search trees are commonly used for implementing float
positioning; e.g. by WebKit.  Persistence enables each object that interacts
with floats to efficiently contain a snapshot of the float list at the time
that object was laid out. That way, incremental layout can invalidate and start
reflow at any point in a containing block.

This commit features extensive use of
[QuickCheck](https://github.com/BurntSushi/quickcheck) to ensure that the rules
of the CSS specification are followed.

Because this is not yet connected to layout, floats will not actually be laid
out in Web pages yet.

Note that unit tests as set up in Servo currently require types that they
access to be public. Therefore, some internal layout 2020 types that were
previously private have been made public. This is somewhat unfortunate.

---

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -2,29 +2,494 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+//! Float layout.
+//!
+//! See CSS 2.1 § 9.5.1: https://www.w3.org/TR/CSS2/visuren.html#float-position
+
 use crate::context::LayoutContext;
 use crate::dom_traversal::{Contents, NodeAndStyleInfo, NodeExt};
 use crate::formatting_contexts::IndependentFormattingContext;
+use crate::geom::flow_relative::{Rect, Vec2};
 use crate::style_ext::DisplayInside;
+use euclid::num::Zero;
+use servo_arc::Arc;
+use std::f32;
+use std::ops::Range;
+use style::values::computed::Length;
 use style::values::specified::text::TextDecorationLine;
 
+/// A floating box.
 #[derive(Debug, Serialize)]
 pub(crate) struct FloatBox {
+    /// The formatting context that makes up the content of this box.
     pub contents: IndependentFormattingContext,
 }
 
 /// Data kept during layout about the floats in a given block formatting context.
-pub(crate) struct FloatContext {
-    // TODO
+///
+/// This is a persistent data structure. Each float has its own private copy of the float context,
+/// although such copies may share portions of the `bands` tree.
+#[derive(Clone, Debug)]
+pub struct FloatContext {
+    /// A persistent AA tree of float bands.
+    ///
+    /// This tree is immutable; modification operations return the new tree, which may share nodes
+    /// with previous versions of the tree.
+    pub bands: FloatBandTree,
+    /// The logical width of this context. No floats may extend outside the width of this context
+    /// unless they are as far (logically) left or right as possible.
+    pub inline_size: Length,
+    /// The current (logically) vertical position. No new floats may be placed (logically) above
+    /// this line.
+    pub ceiling: Length,
 }
 
 impl FloatContext {
-    pub fn new() -> Self {
-        FloatContext {}
+    /// Returns a new float context representing a containing block with the given content
+    /// inline-size.
+    pub fn new(inline_size: Length) -> Self {
+        let mut bands = FloatBandTree::new();
+        bands = bands.insert(FloatBand {
+            top: Length::zero(),
+            left: None,
+            right: None,
+        });
+        bands = bands.insert(FloatBand {
+            top: Length::new(f32::INFINITY),
+            left: None,
+            right: None,
+        });
+        FloatContext {
+            bands,
+            inline_size,
+            ceiling: Length::zero(),
+        }
+    }
+
+    /// Returns the current ceiling value. No new floats may be placed (logically) above this line.
+    pub fn ceiling(&self) -> Length {
+        self.ceiling
+    }
+
+    /// (Logically) lowers the ceiling to at least `new_ceiling` units.
+    ///
+    /// If the ceiling is already logically lower (i.e. larger) than this, does nothing.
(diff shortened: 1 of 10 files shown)


pcwalton:
This is  #25167.

pcwalton:
This should be ready to go now.

r? @Manishearth

pcwalton:
@jdm I get `Package quickcheck 0.9.2 depends on blocked package rand` when I run `mach test-tidy`. How should I proceed?

pcwalton:
I added `quickcheck` to the whitelist in `servo-tidy.toml`, noting that it is only used in tests.

Manishearth:
@pcwalton while I'd love to review this I'm not familiar enough with layout 2020 to be the primary reviewer (I'm not yet done understanding the flexbox code!)

pcwalton:
@Manishearth Well, @jdm suggested that you should be the reviewer since our main layout folks are out at the moment. Do you know who else would be a good reviewer?

Manishearth:
Oh, in that case I'll go for it! Might take a while since I'm still understand things

components/layout_2020/flow/float.rs
```diff
@@ -2,29 +2,484 @@
...
+    Left = 1,
+    Right = 2,
+    Both = 3,
+}
+
+/// Internal data structure that describes an area in which floats may be placed.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct FloatBand {
+    pub top: Length,
+    pub left: Option<Length>,
```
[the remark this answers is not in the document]
Manishearth: Also what is left here anyway? Available space? Distance from left edge of box?

components/layout_2020/flow/float.rs
```diff
@@ -2,29 +2,484 @@
...
+    // replaces the existing band with the new one.
+    fn insert(&self, band: FloatBand) -> FloatBandLink {
+        let mut this = match self.0 {
+            None => return FloatBandLink(Some(Arc::new(FloatBandNode::new(band)))),
+            Some(ref this) => (**this).clone(),
+        };
+
+        if band.top < this.band.top {
+            this.left = this.left.insert(band);
+            return FloatBandLink(Some(Arc::new(this))).skew().split();
```
Manishearth, line 407: aren't skew and split inverses of each other?

components/layout_2020/flow/float.rs
```diff
@@ -2,29 +2,484 @@
...
+            },
+        };
+        let new_float_rect = Rect {
+            start_corner: new_float_origin,
+            size: new_float.size,
+        };
+
+        // Split the first band if necessary.
+        first_band.top = new_float_rect.start_corner.block;
+        self.bands = self.bands.insert(first_band);
```
Manishearth, line 132: Does insert() actually split the band? It seems to me that it just changes it

Manishearth (commented):
I think a deeper explanation of how the bands are supposed to work would help: seems like a float can exist in multiple bands?

Still not done reading everything, and I have to go through the tests which seem to give a clearer idea of how this is supposed to be used

components/layout_2020/flow/float.rs
```diff
@@ -2,29 +2,484 @@
...
+            }
+            band = next_band;
+        }
+        band.top.max(self.ceiling)
+    }
+
+    /// Places a new float and adds it to the list. Returns the start corner of its margin box.
+    pub fn add_float(&mut self, new_float: FloatInfo) -> Vec2<Length> {
+        // Find the first band this float fits in.
+        let mut first_band = self.bands.find(self.ceiling).unwrap();
```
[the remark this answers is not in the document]
pcwalton: I'd want to measure before doing that; it doesn't seem obviously a win as `find` is just O(log n).

components/layout_2020/flow/float.rs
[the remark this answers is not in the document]
pcwalton: Every band (terminology from Gecko) represents a vertical region where floats might be placed. Left and right are distances from the left/right edge of the container to the first legal horizontal position where floats might be placed. In other words, floats have to go in between the "left edge + left" and "right edge - right".

None is there to differentiate between the cases of a zero-width float on some side and no floats at all on that side. If we didn't have this, then, for example,, "clear" would not work properly with zero-width left floats, as the `left` field would be zero even if zero-width left floats are present and we would be unable to clear past them.

components/layout_2020/flow/float.rs
pcwalton: Insert and split are the same operation, because bands only store their top value. e.g. if our bands have top values of 0px, 1px, 5px before, and we're inserting a float at y = 3px, then it becomes 0px, 1px, 3px, 5px after, effectively splitting the 1px band.

components/layout_2020/flow/float.rs
pcwalton: No, because the level checks are different. Skew operates if level(T) = level(L) and split operates if level(T) = level(X).

components/layout_2020/flow/float.rs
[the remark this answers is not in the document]
Manishearth: Could you document this?

Also, what happens when a band is "full"? I don't think I see code handling that (is it supposed to be handled?)

components/layout_2020/flow/float.rs
Manishearth: Oh, I didn't notice the level checks

Manishearth (approved):
r=me

jdm:
@bors-servo retry
* #24622

jdm:
@bors-servo retry
* #24622

pcwalton:
`_RegisterApplication(), FAILED TO establish the default connection to the WindowServer, _CGSDefaultConnection() is NULL.`

jdm:
Filed #27342 about that particular brand of intermittent failure. The _RegisterApplication thing is unrelated and doesn't appear to impact tests.

jdm:
@bors-servo retry
* #26895

PR #27229, opened 2020-07-10
devtools: TabDescription actor

### LINKED ISSUE #27113: Unable to remote debug on Hololens 2
I followed these instructions: https://github.com/servo/servo/wiki/Devtools#running-firefox-reality-in-a-hololens-device-or-emulator

On my hololens (10.32.11.146) I allowed the debugger to connect.
On Firefox Desktop or Firefox Developer, I get the following error :
debug error

Firefox : 77.0.1 (64 bits)
PackageFullName: MozillaFoundation.FirefoxReality_1.1.0.0_arm64__1asrjg2sqer5r

paulrouget:
We could do better by creating the browsing context actor lazily, but that's good enough to re-enable devtools with a recent version of Firefox. fix #27113.

### DIFF
--- a/components/devtools/actors/tab.rs
+++ b/components/devtools/actors/tab.rs
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
+use crate::actors::browsing_context::{BrowsingContextActor, BrowsingContextActorMsg};
+use crate::actors::root::RootActor;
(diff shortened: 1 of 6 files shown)


components/devtools/actors/tab.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Serialize)]
+struct GetTargetReply {
+    from: String,
+    frame: BrowsingContextActorMsg,
+}
+
+pub struct TabDescriptorActor {
+    pub name: String,
+    pub title: RefCell<String>,
+    pub url: RefCell<String>,
```
jdm, line 40: Rather than duplicating this data in TabDescriptorActor and BrowsingContextActor, could we choose one of them?

jdm (changes requested):
Thanks for investigating this!

paulrouget:
@jdm that's probably a better approach.

PR #27279, opened 2020-07-15
Update hololens documentation.

jdm:
This will give us something useful to point users towards.


paulrouget:
Thank you!

@bors-servo r+

PR #27282, opened 2020-07-15
Use `Rc` instead of cloning the `DOMString`

### LINKED ISSUE #27254: ScriptOrigin::text duplicates the string contents
Since these scripts can be large minified scripts, if there's a way to borrow the result instead that would be preferable.

camelid:
I changed the `text` field of `ScriptOrigin` from a `DOMString` to an `Rc<DOMString>`. Then I updated all the related code to work with an `Rc`.

---

- [x] These changes fix #27254 (GitHub issue number if applicable)

- [x] These changes do not require tests because this doesn't introduce new code and should only need to be checked by the compiler

### DIFF
--- a/components/script/dom/htmlscriptelement.rs
+++ b/components/script/dom/htmlscriptelement.rs
@@ -651,7 +651,7 @@ impl HTMLScriptElement {
 
                     fetch_inline_module_script(
                         ModuleOwner::Window(Trusted::new(self)),
-                        Rc::clone(&text_rc),
+                        text_rc,
                         base_url.clone(),
                         self.id.clone(),
                         credentials_mode.unwrap(),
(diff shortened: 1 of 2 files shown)


camelid:
r? @jdm

This is still work-in-progress, but I wanted to get some feedback about it :)

components/script/dom/htmlscriptelement.rs
```diff
@@ -647,7 +649,7 @@ impl HTMLScriptElement {
 
                     fetch_inline_module_script(
                         ModuleOwner::Window(Trusted::new(self)),
-                        text.clone(),
+                        Rc::new(text.clone()),
```
jdm, line 652: Can we use the earlier Rc that was created instead of making a new one here?

jdm (changes requested):
Looks good!

components/script/dom/htmlscriptelement.rs
camelid: I'm guessing yes; I was just trying to get the compiler to stop yelling at me :)

I'll look into it tomorrow

camelid: Okay, I just did it!

jdm (approved):

PR #27285, opened 2020-07-15
Implement WebGPU ErrorScopes

kunalmohan:
This PR covers the following- 
1. update wgpu-core (Rustify `RenderPipelineDescriptor`)
2. Add initial implementation of `pushErrorScope` and `popErrorScope`.
3. Use existing `BGL` if an equivalent already exists.

A brief explanation of the ErrorScope Implementation-
1. We store `HashMap<u64, ErrorScopeInfo>` in `GPUDevice`. `ErrorScopeInfo` is a new struct used to store all necessary info for an ErrorScope.
2. We simultaneously store `Vec<u64>` to keep track of the ErrorScope hierarchy. `next_scope_id` is used to get the `scope_id` for a new Scope.
3. We send the `scope_id` of top-level scope to the server in the message for an operation. The server performs the operation and sends back the scope_id with the result message (only if a scope_id was sent from the client. Otherwise no message is sent.).
4. We store a `HashMap<WebGPUDevice, Dom<GPUDevice>>` in `GlobalScope` to access them from `ScriptThread`.
5. While creating a `GPUDevice` we send the `PipelineId` of the pipeline it was created in so that when the server send the response, we can access the device from the GlobalScope of that Pipeline.
6. We generate the relevant `GPUError` in `GlobalScope`(if required) can call the necessary fn in `GPUDevice`.

This implementation is pretty rough and not complete at all. Broadly the following cases aren't handled at the moment-
1. ErrorFilter is not checked when an error is captured.
2. We don't propagate an error to enclosing scope if a scope did not capture it.
3. `UncapturedErrorEvents` is not implemented.

As for Equivalent BGLs we now store `Vec<(Vec<wgt::BindGroupLayoutEntry>, Dom<GPUBindGroupLayout>)>` in `GPUDevice` and compare `Vec<wgt::BindGroupLayoutEntry>` for new BGL with the existing ones and return one if it is found. Otherwise a new one is created and an entry added to the `Vec`.

r?@kvark

---


components/script/dom/globalscope.rs
```diff
@@ -2939,6 +2948,31 @@ impl GlobalScope {
         self.gpu_id_hub.clone()
     }
 
+    pub fn add_gpu_device(&self, device: &GPUDevice) {
+        self.gpu_devices
```
kvark, line 2952: when are these cleaned up?

components/script/dom/gpudevice.rs
```diff
@@ -73,6 +90,13 @@ pub struct GPUDevice {
     label: DomRefCell<Option<DOMString>>,
     device: webgpu::WebGPUDevice,
     default_queue: Dom<GPUQueue>,
+    error_scopes: DomRefCell<HashMap<u64, ErrorScopeInfo>>,
```
kvark, line 93: can we have `type ErrorScopeId = u64` and use it here?

components/script/dom/gpudevice.rs
```diff
@@ -73,6 +90,13 @@ pub struct GPUDevice {
...
+    scope_stack: DomRefCell<Vec<u64>>,
```
kvark, line 94: don't we mostly need to access all of the scope fields? Perhaps, we could put them in a struct, like `ScopeContext` and store the whole thing under `DomRefCell`

components/script/dom/gpudevice.rs
```diff
@@ -73,6 +90,13 @@ pub struct GPUDevice {
...
+    next_scope_id: Cell<u64>,
+    #[ignore_malloc_size_of = "promises are hard"]
+    lost_promise: DomRefCell<Option<Rc<Promise>>>,
+    #[ignore_malloc_size_of = "defined in webgpu"]
+    bind_group_layouts: DomRefCell<Vec<(Vec<wgt::BindGroupLayoutEntry>, Dom<GPUBindGroupLayout>)>>,
```
kvark, line 99: should this be a hashmap instead?

components/script/dom/gpudevice.rs
```diff
@@ -708,6 +849,49 @@ impl GPUDeviceMethods for GPUDevice {
...
+        let promise = Promise::new_in_current_realm(&self.global(), comp);
+        let scope_id = if let Some(e) = self.scope_stack.borrow_mut().pop() {
+            e
+        } else {
+            promise.reject_error(Error::Operation);
+            return promise;
+        };
+        let mut err_scope;
+        {
+            err_scope = self.error_scopes.borrow_mut().remove(&scope_id).unwrap();
```
kvark, line 879: if we remove it here on the client side, what are we going to do with an error coming from the server with this "scope_id"?

components/webgpu/lib.rs
```diff
@@ -467,23 +475,66 @@ impl<'a> WGPU<'a> {
...
                             layout: bind_group_layout_id,
-                            bindings: bindings.as_slice(),
+                            entries: bindings.as_slice(),
                         };
-                        let _ = gfx_select!(bind_group_id =>
+                        let result = gfx_select!(bind_group_id =>
                             global.device_create_bind_group(device_id, &descriptor, bind_group_id));
+                        if let Some(s_id) = scope_id {
+                            let &pipeline_id = self.devices.get(&WebGPUDevice(device_id)).unwrap();
+                            let op_result;
```
kvark, line 484: let's assign this right away, i.e. `let op_result = if () {..} else {..};`

components/script/dom/gpudevice.rs
```diff
@@ -119,6 +148,37 @@ impl GPUDevice {
     pub fn id(&self) -> webgpu::WebGPUDevice {
         self.device
     }
+
+    pub fn handle_server_msg(&self, scope: u64, result: Result<(), GPUError>) {
+        let mut err_scope;
+        {
+            err_scope = self.error_scopes.borrow_mut().remove(&scope).unwrap();
```
kvark, line 155: why are we even trying so eagerly to remove the element from `self.error_scopes`?

components/script/dom/gpudevice.rs
```diff
@@ -320,6 +416,16 @@ impl GPUDeviceMethods for GPUDevice {
             }
         });
 
+        let mut scope_id = None;
+        if let Some(s_id) = self.scope_stack.borrow_mut().last() {
```
kvark, line 420: this should probably be abstracted away and become `let scope_id = self.use_current_scope();`

components/script/dom/gpudevice.rs
kunalmohan: We'll have to Implement Hash for wgt::BindGroupLayoutEntry for that.

components/script/dom/globalscope.rs
kunalmohan: These aren't. I still need to figure this out. Maybe somehow send the pipeline id with FreeDeviceId msg. Otherwise searvh for the DeviceId in all Pipeline(but that would be highly inefficient and difficult once we start support for Workers)

components/script/dom/gpudevice.rs
kunalmohan: We reinsert it(line 890) in case we have more results to be captured

components/script/dom/gpudevice.rs
kunalmohan: I figured it was less lines of code if we remove and conditionally reinsert it. Otherwise we first borrow it mutably and then make some checks a second time to see if the scope can be removed.

kvark: can we use the Entry API there?

kunalmohan: I have used flags here instead. We set the flags while in-place checking and remove only if the flags are set. Using only Entry wouldn't have made much difference.

components/webgpu/lib.rs
```diff
@@ -467,23 +475,64 @@ impl<'a> WGPU<'a> {
...
                             layout: bind_group_layout_id,
-                            bindings: bindings.as_slice(),
+                            entries: bindings.as_slice(),
                         };
-                        let _ = gfx_select!(bind_group_id =>
+                        let result = gfx_select!(bind_group_id =>
                             global.device_create_bind_group(device_id, &descriptor, bind_group_id));
+                        if let Some(s_id) = scope_id {
+                            let &pipeline_id = self.devices.get(&WebGPUDevice(device_id)).unwrap();
+                            let op_result = if let Err(e) = result {
```
kvark, line 484: this code gets repeated, can we move it to a function

components/script/dom/globalscope.rs
kunalmohan: If we want a device to be cleaned at the time `FreeDeviceId` is sent to the content process, we can send a `FreeDeviceId` to the server itself, so that `WGPU` removes the device from its hashmap and send the content process another message with the pipeline_id which could then remove the `GPUDevice` from above HashMap.

We'll have to follow this whole process only because we can't send the pipeline_id with `FreeDeviceId` from `IdentityRecycler` in `webgpu/identity.rs`. If we can figure that out somehow things would be much easier.
(But we'll still have to send a message to the server itself to remove `Device` from HashMap in `WGPU`)

components/webgpu/lib.rs
kunalmohan: `result` is of a different type for each method so we'll have to assign `op_result` separately at each instance but the `send(WebGPUOpResult)` can be extracted.

components/script/dom/gpudevice.rs
```diff
@@ -119,6 +158,53 @@ impl GPUDevice {
...
+                        }
+                    } else if err_scope.op_count == 0 {
+                        promise.resolve_native(&None::<GPUError>);
+                    }
+                }
+            }
+            err_scope.op_count == 0 && err_scope.promise.is_some()
+        } else {
+            warn!("Could not find ErrroScope with Id({})", scope);
+            return;
```
kvark, line 189: can we just `false` instead?

components/script/dom/gpudevice.rs
```diff
@@ -119,6 +158,53 @@ impl GPUDevice {
...
+        if remove {
+            let _ = context.error_scopes.remove(&scope);
+        }
+    }
+
+    fn use_current_scope(&self) -> Option<ErrorScopeId> {
+        let mut context = self.scope_context.borrow_mut();
+        let scope_id = context.scope_stack.last().map(|s_id| *s_id);
+        scope_id
+            .map(|s_id| {
```
kvark, line 200: let's use `and_then`

components/script/dom/gpudevice.rs
```diff
@@ -119,6 +158,53 @@ impl GPUDevice {
...
+            return;
+        };
+        if remove {
+            let _ = context.error_scopes.remove(&scope);
+        }
+    }
+
+    fn use_current_scope(&self) -> Option<ErrorScopeId> {
+        let mut context = self.scope_context.borrow_mut();
+        let scope_id = context.scope_stack.last().map(|s_id| *s_id);
```
kvark, line 198: we shouldn't need that `map` at the end, since we are mapping right after that anyway

components/webgpu/lib.rs
```diff
@@ -82,12 +83,14 @@ pub enum WebGPURequest {
     },
     CreateBindGroup {
         device_id: id::DeviceId,
+        scope_id: Option<u64>,
```
kvark, line 86: let's add a TODO to consider using `NonZeroU64` to reduce size of the enum by 8 bytes

components/webgpu/lib.rs
kvark: this code doesn't particularly care *what* type `result` has, as long as it's `Result<T: Display, ()>`, so a function can still help here

kvark (commented):
I think I understand the code better now, and it's almost there.
Great work! Just one last round

components/script/dom/gpudevice.rs
kunalmohan: I used `map` because contents of `context` had to be borrowed mutably in the next step. I couldn't do so if the previous borrow was alive. Therefore, I created a new `Option<u64>` via map here.

kvark: ok, can also do `.cloned()`

kvark (approved):
beautiful, thank you!

kunalmohan:
@kvark I have attempted to solve the `GPUDevice` cleanup issue. I am not completely happy with this, but this should work until we come up with a better solution.

kvark (approved):
Looks fine for now, thank you!

PR #27287, opened 2020-07-16
Make it possible to run fxr:// url in non-kiosk mode

### LINKED ISSUE #27274: Can't open devtools panel from fxr:// kiosk mode
Loading a URL with fxr:// hides the URL bar, which also hides the devtools button. We should add a separate protocol handler for kiosk mode, and allow one protocol to open the app directly to a target URL without otherwise changing the app's behaviour.

paulrouget:
Fix #27274


jdm (approved):

jdm:
@bors-servo r+
Thanks!

PR #27164, opened 2020-07-03
Replace ws-rs with async-tungstenite.

jdm:
This change moves us from ws-rs (which doesn't see a lot of maintainer activity) and its custom async implementation to tungstenite and the tokio ecosystem. This is particularly important because of #27043, which breaks SSL websockets on Windows. 

Depends on https://github.com/sdroege/async-tungstenite/pull/40/.

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #27043
- [x] There are tests for these changes

PR #27300, opened 2020-07-16
Remove explicit Host header from websocket connection.

jdm:
This was a regression from #27164, and fixing it allows hubs.mozilla.com to work again.


jdm:
@bors-servo retry
* #23290

PR #27304, opened 2020-07-17
UWP: non random port for devtools, and support for custom UWP prefs

### LINKED ISSUE #22970: prefs.json: swap prefs before building
As of now, we swap the preferences during `mach package`. libsimpleservo reads prefs.json at compile time. So it should be done before building.

### LINKED ISSUE #27267: Support non-random devtools port
A user reports that the randomized port makes the process of connecting from desktop Firefox more frustrating.

paulrouget:
packages-prefs.json is now taken into account for libsimpleservo. Making it possible to have custom pref for the UWP builds, removing some weirdness in the way we handle preferences in the hololens code.

This also adds a new set of preferences to control the devtools server startup state, and make the port choice constant across sessions.

Fix #27267 
Fix #22970


jdm (approved):
This is a nice cleanup!

paulrouget:
> What about the args variable that is passed in?

Good catch. Fixed.

paulrouget:
Hold on. I found a regression.

paulrouget:
The way the homepage pref was overridden wouldn't work because we were relying on the fact that the **UWP** pref always had a shell.homepage key, which is not the case anymore as we use the servo pref.

Fixed with: https://github.com/servo/servo/pull/27304/commits/9174b201b0b0eeedded34c9cd9868669802f6b4e#diff-831a752eac48b390b35aae339dbc1097R375

jdm:
```
error: unused import: `set_pref`
  --> ports\libsimpleservo\api\src\lib.rs:34:33
   |
34 | use servo::servo_config::{pref, set_pref};
   |                                 ^^^^^^^^
   |
```

PR #27325, opened 2020-07-20
Fire sessionavailable

### LINKED ISSUE #27272: Enable sessionavailable event when launching with fxr://
The fxr:// handler puts FxR into a kiosk-mode kind of environment that doesn't have a URL bar. It would be good to support automatically entering immersive mode as part of that.

paulrouget:
Depends on #27304
Fix #27272


jdm:
Is this still WIP? 32ddb35c69c86da3179f02dc4352405b86f14f89 looks sensible to me.

paulrouget:
@jdm r?

PR #27343, opened 2020-07-21
Update mozjs

utsavoza:
Pick up [servo/rust-mozjs#517](https://github.com/servo/rust-mozjs/pull/517) and [servo/mozjs#256](https://github.com/servo/mozjs/pull/256)


CYBAI:
@bors-servo retry

- #23290

Issue #27315, opened 2020-07-17
Intermittent crash in hang monitor tests


jdm:
```
     Running target/debug/deps/hang_monitor_tests-9410268c27e2806f
running 3 tests
test test_hang_monitoring_unregister ... ok(B
error: test failed, to rerun pass '-p background_hang_monitor --test hang_monitor_tests'
Caused by:
  process didn't exit successfully: `/repo/target/debug/deps/hang_monitor_tests-9410268c27e2806f` (signal: 27)
```

gterzian:
This might be a dupe of https://github.com/servo/servo/issues/23168

And specifically I think this relates to the stack walking somewhere at https://github.com/servo/servo/blob/9864e4ce6cb73b81f1c382af0ae20a24203e380f/components/background_hang_monitor/sampler_linux.rs#L225

lumiscosity:
There's been a lot of refactoring here, including switching to the `backtrace` crate for backtraces, and the linked line has been removed. I can't replicate this anymore after running the test manually ~20 times. Tested on Arch Linux x86_64, Servo built against commit 1f0f079, dev build.

PR #27348, opened 2020-07-21
Implement GPUCommandEncoder.copy commands

kunalmohan:
r?@kvark

---

### DIFF
--- a/components/script/dom/gpucommandencoder.rs
+++ b/components/script/dom/gpucommandencoder.rs
@@ -5,12 +5,15 @@
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::GPUBufferBinding::GPUSize64;
 use crate::dom::bindings::codegen::Bindings::GPUCommandEncoderBinding::{
-    GPUCommandBufferDescriptor, GPUCommandEncoderMethods, GPUComputePassDescriptor,
-    GPURenderPassDescriptor, GPUStencilLoadValue, GPUStoreOp,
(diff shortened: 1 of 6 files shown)


components/script/dom/gpucommandencoder.rs
```diff
@@ -275,3 +361,37 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
...
+        layout: convert_texture_data_layout(&buffer_cv.parent),
+    }
+}
+
+pub fn convert_texture_cv(texture_cv: &GPUTextureCopyView) -> wgpu_com::TextureCopyView {
+    wgpu_com::TextureCopyView {
+        texture: texture_cv.texture.id().0,
+        mip_level: texture_cv.mipLevel,
+        origin: match texture_cv.origin {
+            GPUOrigin3D::RangeEnforcedUnsignedLongSequence(ref v) => wgt::Origin3d {
```
kvark, line 377: we've got to be careful here: the user doesn't have to specify all 3 elements, it could be just `[x]`, which should mean the same as `[x, 0, 0]`, for example. Same goes for `GPUExtent3D`, but the defaults in it are `1`

kvark (approved):

components/script/dom/gpucommandencoder.rs
kunalmohan: I suppose this will be the case with `GPUColor` too. But what would be the defaults in that case?

kvark: not sure.. for color, it would make sense to default alpha to 1, but not the other components

jdm:
@bors-servo retry
* #25513 #27315

PR #27365, opened 2020-07-22
Fixes for off-thread compilation of scripts

### LINKED ISSUE #27349: Off-thread compiled scripts don't have dynamic import integration
When reading through the changes from https://github.com/servo/servo/pull/26710, I realized that only the scripts that are compiled as part of evaluation get the script private set on them. We should extend this to the off-thread compiled scripts as well.

### LINKED ISSUE #27355: Non-null terminated strings are being passed to CompileOptionsWrapper::new
https://github.com/servo/servo/blob/a24291362946f92bf155b390ea469b7be50a4c2d/components/script/dom/htmlscriptelement.rs#L428

We need to convert the rust string into a CString before passing the pointer to it.

jdm:
Fixes #27355. Fixes #27349.

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -86,14 +86,14 @@ use js::jsapi::Compile1;
 use js::jsapi::SetScriptPrivate;
 use js::jsapi::{CurrentGlobalOrNull, GetNonCCWObjectGlobal};
 use js::jsapi::{HandleObject, Heap};
-use js::jsapi::{JSContext, JSObject};
+use js::jsapi::{JSContext, JSObject, JSScript};
 use js::jsval::PrivateValue;
 use js::jsval::{JSVal, UndefinedValue};
 use js::panic::maybe_resume_unwind;
 use js::rust::transform_str_to_source_text;
-use js::rust::wrappers::{JS_ExecuteScript, JS_ExecuteScript1, JS_GetScriptPrivate};
+use js::rust::wrappers::{JS_ExecuteScript, JS_GetScriptPrivate};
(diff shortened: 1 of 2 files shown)


components/script/dom/globalscope.rs
```diff
@@ -2596,59 +2596,55 @@ impl GlobalScope {
...
+                            None,
+                        ));
+
+                        SetScriptPrivate(
+                            *compiled_script,
+                            &PrivateValue(Rc::into_raw(module_script_data) as *const _),
+                        );
+                    }
+
+                    let result = JS_ExecuteScript(*cx, compiled_script.handle(), rval);
```
CYBAI, line 2646: Because of the early return in `SourceCode::Text` branch above and also the null check for returning a network error in `off_thread_compilation_callback`, we should always get a non-null `compiled_script`.

I'm wondering, would it be worth adding an assert before executing the script 👀?

CYBAI (approved):
Thanks! This all looks good to me! Only one question!

(Btw, the patch for CString reminds me that I did a redundant `as_bytes` for transforming a string into CString in script_module 😂)

Edit: as a reviewer, I should say `r=me` 😆

components/script/dom/globalscope.rs
jdm: Sounds reasonable!

jdm:
@bors-servo retry
* #24622

jdm:
@bors-servo retry
* #26895

jdm:
@bors-servo retry
* #26895

jdm:
@bors-servo retry
* #26895

jdm:
@bors-servo retry
* #26895

PR #27368, opened 2020-07-22
Use backtrace fork with UWP-specific fix.

### LINKED ISSUE #27367: WACK regression: Module32FirstW and Module32NextW
```
API Module32FirstW in kernel32.dll is not supported for this application type. simpleservo.dll calls this API.
API Module32NextW in kernel32.dll is not supported for this application type. simpleservo.dll calls this API.
```

jdm:
Fixes #27367, which was introduced by 57f4f0a09fbb0d26448f63a3f61dc11fb4209f8f. I'd like to do a new release in the store, and I would prefer not to wait for a new version of backtrace to be released before that can happen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,8 +242,7 @@ dependencies = [
 [[package]]
 name = "backtrace"
 version = "0.3.50"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo retry
* #24622

PR #27373, opened 2020-07-22
Updated webxr

### LINKED ISSUE #27286: Hololens 2 rendering issues with Unity and WebXR - Regression?
In the current Firefox Reality in the Microsoft Hololens store, AR session works and rendered as intended.
But the nightly build version ( 13/07 ServoApp_1.1.0.0_x64_arm64 ), has some rendering issues:

- Rendering order.

- Smearing on transparent background.

I took screenshots from the emulator, but as I understood from people with the device, it's the same.

This is how it looks on the store version, you can see that the rendering order is right, and the transparent background is black (as the emulator doesn't show scene mesh by default):
image

This is how the same AR session looks in the nightly build. You can see the smearing of the numbers and other objects. You can also see some errors in rendering order, the hand model should be near the user and not behind the floating table:
image

This is how it looks when I tried to use `WebGLRenderingContext.clear` (`COLOR_BUFFER_BIT` `DEPTH_BUFFER_BIT` and with and without `STENCIL_BUFFER_BIT`) at the frame start. There's no smearing, but the hand is rendered behind the floating table:
image

As far as I know, on AR mode we shouldn't use clear at the frame start, as in Handheld and Video see through devices, it has the camera image. If we use clear at the frame start, we won't see the camera image.

asajeffrey:
Updates webxr to clear the render target every frame.

---

- [x] These changes fix #27286 (really this time)
- [x] These changes do not require tests because we don't reftest webxr

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6825,7 +6825,7 @@ dependencies = [
 [[package]]
 name = "webxr"
 version = "0.0.1"
-source = "git+https://github.com/servo/webxr#b170a86c081f762d201d3e870febc6e900ad365b"
+source = "git+https://github.com/servo/webxr#2ff286efcf8bc85fbe2fc91df1aefb1aff911781"
 dependencies = [
  "android_injected_glue",
  "bindgen",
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

asajeffrey:
@bors-servo retry
- #25513

PR #27376, opened 2020-07-23
Use the size of the containing block, not the size of the block formatting context, to place floats in layout 2020.

pcwalton:
The containing block for a float is not necessarily the same as the block
formatting context the float is in per CSS 2.1 [1]:

"For other elements, if the element’s position is relative or static, the
containing block is formed by the content edge of the nearest block container
ancestor box."

This shows up in the simplest case:

	<html>
	<body>
	<div style="float: left">Hello</div>
	</body>
	</html>

In this case, the `<html>` element is the block formatting context with inline
size equal to the width of the window, but the `<body>` element with nonzero
inline margins is the containing block for the float. The float placement must
respect the content box of the `<body>` element (i.e. floats must not overlap
the `<body>` element's margins), not that of the `<html>` element.

Because a single block formatting context may contain floats with different
containing blocks, the left and right "walls" of that containing block become
properties of individual floats at the time of placement, not properties of the
float context itself.

Additionally, this commit generalizes the float placement logic a bit to allow
the placement of arbitrary objects, not just floats. This is intended to
support inline layout and block formatting context placement.

This commit updates the `FloatContext` and associated tests only and doesn't
actually wire the context up to the rest of layout, so floats in pages still
aren't actually laid out.

[1]: https://drafts.csswg.org/css2/#containing-block-details

r? @Manishearth 

---

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -36,9 +36,6 @@ pub struct FloatContext {
     /// This tree is immutable; modification operations return the new tree, which may share nodes
     /// with previous versions of the tree.
     pub bands: FloatBandTree,
-    /// The logical width of this context. No floats may extend outside the width of this context
-    /// unless they are as far (logically) left or right as possible.
-    pub inline_size: Length,
     /// The current (logically) vertical position. No new floats may be placed (logically) above
     /// this line.
     pub ceiling: Length,
@@ -47,7 +44,7 @@ pub struct FloatContext {
 impl FloatContext {
     /// Returns a new float context representing a containing block with the given content
     /// inline-size.
-    pub fn new(inline_size: Length) -> Self {
+    pub fn new() -> Self {
         let mut bands = FloatBandTree::new();
         bands = bands.insert(FloatBand {
             top: Length::zero(),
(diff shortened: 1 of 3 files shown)


components/layout_2020/flow/float.rs
```diff
@@ -189,19 +194,20 @@ pub enum ClearSide {
...
+    /// The distance from the left edge of the block formatting context to the first legal
+    /// (logically) horizontal position where floats may be placed. If `None`, there are no floats
+    /// to the left; distinguishing between the cases of "a zero-width float is present" and "no
+    /// floats at all are present" is necessary to, for example, clear past zero-width floats.
     pub left: Option<Length>,
-    /// The distance from the right edge to the first legal (logically) horizontal position where
-    /// floats may be placed. If `None`, there are no floats to the right; distinguishing between
-    /// the cases of "a zero-width float is present" and "no floats at all are present" is
-    /// necessary to, for example, clear past zero-width floats.
+    /// The distance from the *left* edge of the block formatting context to the first legal
```
pcwalton, line 202: (The comment was incorrect before.)

jdm:
@bors-servo retry
* #24786

PR #27385, opened 2020-07-23
Add Layout 2020 support for `display: list-item`

SimonSapin:


components/layout_2020/lists.rs
```diff
@@ -0,0 +1,50 @@
...
+        UrlOrNone::Url(url) => Some(vec![
+            PseudoElementContentItem::Replaced(ReplacedContent::from_image_url(
+                info.node, context, url,
+            )?),
+            PseudoElementContentItem::Text(" ".into()),
+        ]),
+        UrlOrNone::None => None,
+    };
+    marker_image().or_else(|| {
+        Some(vec![PseudoElementContentItem::Text(
```
Manishearth, line 33: why not just inline this into the above closure?

Manishearth (approved):

components/layout_2020/lists.rs
SimonSapin: The closure uses `?` so it has two paths that can return `None`. This implements “If the `<image>` represents a valid image” in https://drafts.csswg.org/css-lists/#marker-image

jdm:
@bors-servo retry
* #24786

PR #27397, opened 2020-07-24
[return value version] Handle `inset: auto` values for absolutely positioned elements

### LINKED ISSUE #27387: <position>: auto does not work
```html
<p>Text text text text text text text </p>

<div style="width: 100px; height: 100px;  background-color: green; position: absolute;"></div>
```

Expected:
image

Actual: 
image

This is because we're resolving `auto` values for top/left/bottom/right to mean zero when we actually mean "where it was actually found"

cc @SimonSapin

Manishearth:
Fixes https://github.com/servo/servo/issues/27387

I'm not 100% sure of this approach, i don't like reconstructing the
vector there.

Also, when tried on the testcase in https://github.com/servo/servo/issues/27387, while the square no longer overlaps, it's still very close to the previous line. I think I need to handle margins.

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -166,9 +166,14 @@ fn layout_block_level_children(
     mut float_context: Option<&mut FloatContext>,
     collapsible_with_parent_start_margin: CollapsibleWithParentStartMargin,
 ) -> FlowLayout {
-    fn place_block_level_fragment(fragment: &mut Fragment, placement_state: &mut PlacementState) {
+    fn place_block_level_fragment(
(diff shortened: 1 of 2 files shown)


Manishearth:
@bors-servo try

running the entire css suite is taking forever, and also apparently my computer is fast enough to turn some of the timeouts into other errors

Manishearth:
Opened a second version of this PR in https://github.com/servo/servo/pull/27399, based on @SimonSapin's comments. It's a separate PR because the approaches are pretty different.

Manishearth:
Closing in favor of https://github.com/servo/servo/pull/27399

PR #27399, opened 2020-07-24
Handle `inset: auto` values for absolutely positioned elements

### LINKED ISSUE #27387: <position>: auto does not work

Manishearth:
Fixes #27387

This is the same as https://github.com/servo/servo/pull/27397 , but written to share the box offset data instead of bubbling the hoisted box up based on comments from @SimonSapin.

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -916,7 +916,7 @@ impl AbsoluteOrFixedPositionedFragment {
         stacking_context: &mut StackingContext,
     ) {
         let hoisted_fragment = self.hoisted_fragment.borrow();
-        let fragment_ref = match hoisted_fragment.as_ref() {
+        let fragment_ref = match hoisted_fragment.fragment.as_ref() {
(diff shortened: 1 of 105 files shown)


Manishearth:
@bors try=wpt-2020

I've already updated expectations, so i'm hoping to see if anything changed between these two PRs.

Manishearth:
It's the same tests, I just forgot to `git add` added files.

SimonSapin:
Looks good! r=me with a rebase

jdm:
@bors-servo retry
* #27414

jdm:
@bors-servo retry
* #23290

Manishearth:
@bors-servo retry

```
  ▶ FAIL [expected PASS] /css/css-color/t421-rgb-values-meaning-b.xht
```

PR #27403, opened 2020-07-25
Add more gstreamer plugins to nightly builds.

jdm:


jdm:
With the latest package from https://community-tc.services.mozilla.com/api/queue/v1/task/DT4lyLssTW-SnkeDWwKs3w/runs/0/artifacts/public/servo-tech-demo.dmg, I am able to enter a Hubs room (including the webrtc data channel) and hear sounds.

Manishearth:
@bors-servo r+

I wonder why we need the ID3 plugins? do we support mp3 metadata?

jdm:
I threw the id3 plugins in there because I saw some gstreamer warning output about not being able to demux an mp3 because of id3 stuff when testing hubs.

Manishearth:
Ah, seems fair. That's a bit surprising, but not an issue,

PR #27415, opened 2020-07-27
Place floats in layout 2020, but don't flow text around the floats yet.

pcwalton:
This commit puts floats behind the `layout.floats.enabled` pref, because of the
following issues and unimplemented features:

* Inline formatting contexts don't take floats into account, so text doesn't
  flow around the floats yet.

* Non-floated block formatting contexts don't take floats into account, so BFCs
  can overlap floats.

* The `clear` property is only respected on floats, not on other elements; i.e.
  regular boxes can't clear past floats yet.

* Block formatting contexts that contain floats don't expand vertically to
  contain all the floats. That is, floats can stick out the bottom of BFCs,
  contra spec.

* Floats don't respect negative margins yet.

* Floats are painted in the same order as regular content, contra spec.

r? @Manishearth

---


Manishearth (approved):
r=me

tests/wpt/mozilla/meta-layout-2020/mozilla/localeCompare.html.ini
```diff
@@ -1,4 +1,2 @@
 [localeCompare.html]
   bug: https://github.com/servo/servo/issues/25802
-  [localeCompare should return the same as other browsers, even though it's implementation-dependent]
-    expected: FAIL
```
jdm, line 4: This needs to be reverted.

tests/wpt/mozilla/meta-layout-2020/mozilla/navigator.html.ini
```diff
@@ -1,4 +1,4 @@
 [navigator.html]
-  [navigator.platform mac]
+  [navigator.platform linux]
```
jdm, line 2: This needs to be reverted.

tests/wpt/mozilla/meta-layout-2020/mozilla/promise.html.ini
```diff
@@ -0,0 +1,19 @@
+[promise.html]
+  [Resolve callback gets argument]
```
jdm, line 2: These failures are very surprising.

PR #27423, opened 2020-07-28
 [WIP] Drop impls for DOM types should be forbidden

### LINKED ISSUE #26488: Drop impls for DOM types should be forbidden
@pcwalton @Manishearth @asajeffrey I recall discussing this topic in Berlin, but I can't find the notes I thought I took. Attempting to reconstruct from memory:
* there exist DOM objects that have Drop impls for various reasons
* there was agreement that it should be possible to run some code in response to a GC finalizer (eg. webgl objects need to be able to unreference their graphics resources)
* the easiest way to ban Drop impls is to make the `#[dom_struct]` macro also add an empty Drop impl, which will conflict with any manual implementation
* for any code which does require a valid Drop, we can create inner structs that only hold the data required for that implementation and nothing more

Does this make sense? My memory is haziest when it comes to the last point.

nicoabie:
Drop impls for DOM types should be forbidden
---

- [X] These changes fix #26488 

- [X] These changes do not require tests because it is a refactor and the compiler will check

Status of the PR
- [X] CanvasRenderingContext2D
- [x] EventSource
- [ ] GPUBuffer
- [ ] GPUTexture
- [ ] HTMLMediaElement
- [ ] Promise
- [ ] WebGLBuffer
- [ ] WebGLFramebuffer
- [ ] WebGLProgram
- [ ] WebGLQuery
- [ ] WebGLRenderbuffer
- [ ] WebGLRenderingContext
- [ ] WebGLSampler
- [ ] WebGLShader
- [ ] WebGLSync
- [ ] WebGLTexture
- [ ] WebGLTransformFeedback
- [ ] Worklet

### DIFF
--- a/components/script/dom/canvasrenderingcontext2d.rs
+++ b/components/script/dom/canvasrenderingcontext2d.rs
@@ -31,6 +31,21 @@ use ipc_channel::ipc::IpcSender;
 use servo_url::ServoUrl;
 use std::mem;
 
+#[derive(JSTraceable, MallocSizeOf)]
+struct CanvasRenderingContext2DDroppableFields {
+    canvas_id: CanvasId,
+    #[ignore_malloc_size_of = "Defined in ipc-channel"]
+    ipc_renderer: IpcSender<CanvasMsg>,
+}
+
+impl Drop for CanvasRenderingContext2DDroppableFields {
+    fn drop(&mut self) {
+        if let Err(err) = self.ipc_renderer.send(CanvasMsg::Close(self.canvas_id)) {
+            warn!("Could not close canvas: {}", err)
+        }
+    }
+}
+
 // https://html.spec.whatwg.org/multipage/#canvasrenderingcontext2d
 #[dom_struct]
 pub struct CanvasRenderingContext2D {
@@ -39,6 +54,7 @@ pub struct CanvasRenderingContext2D {
     /// for ones created by a paint worklet, this is None.
     canvas: Option<Dom<HTMLCanvasElement>>,
     canvas_state: CanvasState,
+    droppable_fields: CanvasRenderingContext2DDroppableFields,
 }
 
 impl CanvasRenderingContext2D {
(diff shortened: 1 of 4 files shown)


components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -31,14 +31,32 @@ use ipc_channel::ipc::IpcSender;
 use servo_url::ServoUrl;
 use std::mem;
 
+#[derive(JSTraceable, MallocSizeOf)]
+#[unrooted_must_root_lint::must_root]
+struct CanvasRenderingContext2DDroppableFields {
+    canvas_state: CanvasState,
+}
```
jdm, line 38: An easier way to make this change would be to leave CanvasState as-is, and make CanvasRenderingContext2DDroppableFields only store the IpcSender that gets used in the destructor. We can call `canvas_state.get_ipc_renderer()` right after creating the `CanvasState` to obtain it.

nicoabie: Yeah, that makes sense!

nicoabie: I'm having trouble specifying the lifetime of the `ipc_renderer: &IpcSender<CanvasMsg>` because get_ipc_renderer() returns a reference. How do I tell rust to trust me and that reference will live long enough to be valid?

jdm: I suggest cloning the sender and storing that instead.

components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -31,6 +31,25 @@ use ipc_channel::ipc::IpcSender;
 use servo_url::ServoUrl;
 use std::mem;
 
+#[derive(JSTraceable, MallocSizeOf)]
+#[unrooted_must_root_lint::must_root]
```
jdm, line 35: The `must_root` attribute shouldn't be needed any more.

nicoabie: Got it, I just rebased with latest and removed the attribute but I can't figure out why it complains telling me that
```
error: Expression of type canvas_state::CanvasState must be rooted
  --> components/script/dom/canvasrenderingcontext2d.rs:66:13
   |
66 |         let canvas_state =
   |             ^^^^^^^^^^^^
   |
   = note: `#[deny(unrooted_must_root)]` on by default

```

Any hints?

jdm: Ah, that's unfortunate. I think the best thing to do here is split out the initialization of CanvasState into two parts - extract [this code](https://github.com/servo/servo/blob/3c426d707859091cae24a32db6575609c42fa646/components/script/canvas_state.rs#L151-L159) into a static method that returns a `RemoteCanvasState` struct, and make `CanvasState` accept a `RemoteCanvasState` argument. Then you can leave the `CanvasState::new` call where it was before this PR and avoid the rooting issues.

nicoabie: Thanks josh! Now I can work on the next 26 dom structs!

mrobinson:
Closing this as it is quite stale and there is a new effort to do this.

Issue #27167, opened 2020-07-03
Add tool for uploading FxR logs


jdm:
In cases where FxR encounters a panic, since we now log them by default since #27157 it would be really nice to make it easy for those logs to be uploaded to a place that FxR developers can see them. Here are some ideas I have:
* make the about:failure page contain a button that uploads stdout.txt and the current URL to somewhere
  * this requires #26523, and requires passing the extra data into the page contents
* create a separate UWP app view for a crash reporter, trigger it from [on_panic](https://github.com/servo/servo/blob/8800452e347bc62d04b5678baadb39705d45503b/support/hololens/ServoApp/ServoControl/Servo.cpp#L38)

We also need to figure out where the logs can be uploaded to. One possibility would be another flask server that runs on build.servo.org; another possibility would be gists or something similar.

jdm:
Another possibility for the UX:
* on startup, move any existing stdout.txt to stdout-last.txt
* add a new tab to the developer tools panel with a button for uploading stdout-last.txt

paulrouget:
Why is stdout empty in the case of the crash we found in #27270?

paulrouget:
It's actually not empty. The app debugger needs to be stopped for the file to be written.

paulrouget:
We could rely on the store crash reporter: https://docs.microsoft.com/en-us/appcenter/sdk/crashes/uwp

paulrouget:
I have identified 5 places where we should catch exceptions/panics:


### Handler $1:

That's the only problematic case as the app crashes afterward "sometimes":

This will catch a panic as described in #27270:
```rust
    std::panic::set_hook(Box::new(|_| {}));
```

… but I didn't find a way to keep the UWP alive.

### Handler $2:

We also have an embedder message `EmbedderMsg::Panic` that we don't handle at the moment. Comes from the constellation forwarding a pipeline panic.

UWP doesn't crash.

### Handler $3:

There's the `capi::register_panic_handler` for the code being executed from the C API.

This also triggers handler $1.

UWP doesn't crash.

### Handler $4:

There's also the non handled exceptions of the UWP app. See `UnhandledException` handler in App.cpp.
If there's an handled exception in the thread UI, it catches it. Calling `Handled(true)` will not crash the app.

### Handler $5

To catch UWP exception int the GL thread, we need to explicitly catch them:

```cpp
try {
    Loop();
} catch (const std::exception& e) {
}
```

UWP app won't stop.

paulrouget:
[...]

Issue #26523, opened 2020-05-14
Create an embedding-specific about:failure


jdm:
For the UWP application, it would be nice to load a "Ooops, something went wrong" page when panics occur. We already load `about:failure` which doesn't actually exist, so this "just" needs some plumbing from the embedder layer to provide meaningful content.

paulrouget:
Fixed by #27438.

PR #27438, opened 2020-07-29
UWP Crash reporter

### LINKED ISSUE #27435: UWP: Preference inconsistency
There's a difference between the value retrieved via `GetPref` and `GetPrefs`.

STR:
- create a bool pref set to false
- open the UWP app, and check the pref
- restart app, pref is still false

But retrieving the pref via `GetPref` gets us the proper value.

paulrouget:
This is supposed to address #27167 and #26523. Also fix #27435.

These changes are still WIP as I found a few bugs, it needs more testing and the actual code to upload is not implemented yet. But I'd like to get an early feedback.

First, panics are caught via `panic::set_hook` instead of `catch_unwind` allowing us to catch more panics.
We also now report panics reported via the `Embedder:Panic` message.
Once the panic is caught, if possible, we try to recover.
I haven't found a way to recover when the panic is caught is a non-GL thread. We need a generic way to throw from the UWP code, and even trying to add a UnhandledEvent handler doesn't appear to work.

Once a panic is caught (even if we can not recover) a crash-report file is created, including the backtrace, stdout, and the current url.

If the app did not crash at that point, or after a restart if it did, we check if the crash report file is present, and if so, we present a panel to the user to allow them to upload the report. At that point the user can also add details to the report.

Screen Shot 2020-07-29 at 12 35 44

### DIFF
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -482,6 +482,9 @@ mod gen {
                 max_length: i64,
             },
             shell: {
+                crash_reporter: {
+                    enabled: bool,
+                },
                 homepage: String,
                 keep_screen_on: {
                     enabled: bool,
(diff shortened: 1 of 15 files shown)


jdm:
This looks really great!

paulrouget:
@jdm can we land that preffed-off for now?

Next step will be to upload the crash report.

jdm (approved):

PR #27440, opened 2020-07-29
Do not lock if GL thread fails

paulrouget:
This might help sometimes when the GL thread fails.

@jdm as I was mentioning, without this, the whole browser can end up being locked (UI being unresponsive).
With this, Servo stays irresponsive, but the app is not locked.


jdm:
Ok, this is subtle enough that I'd like to spend a bit of time testing it on the device to make sure there are no surprises. Let's target it for the next release instead of the one for this week.

paulrouget:
We can skip that PR if it's not for v1.2, as it will be part of #27438.

paulrouget:
This happened in #27438

PR #27402, opened 2020-07-25
Add a copy of WebGPU CTS to wpt.

kunalmohan:
The test suite is generated from `glsl-dependent` branch in https://github.com/gpuweb/cts
We already have 6/44 tests passing and a number of subtests in other tests also pass.

r?@jdm
cc @kvark

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #27395  (GitHub issue number if applicable)

- [X] These changes do not require tests because we don't have tests for tests.

PR #27447, opened 2020-07-29
Refactor and improve GPUErrorScopes

kunalmohan:
- Remove use of equivalent BGLs
- Capture errors from more `Createxxx` operations
- Address crashes on macOS in #27402 

Improved ErrorScope model attempts to- 

1. Identify and report `OutOfMemoryError` separately.
1. Match `GPUErrorFilter` and pass on uncaptured errors to parent scope.

r?@kvark

---

### DIFF
--- a/components/script/dom/gputexture.rs
+++ b/components/script/dom/gputexture.rs
@@ -10,21 +10,21 @@ use crate::dom::bindings::codegen::Bindings::GPUTextureViewBinding::{
     GPUTextureAspect, GPUTextureViewDescriptor, GPUTextureViewDimension,
 };
 use crate::dom::bindings::reflector::{reflect_dom_object, DomObject, Reflector};
-use crate::dom::bindings::root::DomRoot;
+use crate::dom::bindings::root::{Dom, DomRoot};
 use crate::dom::bindings::str::USVString;
 use crate::dom::globalscope::GlobalScope;
-use crate::dom::gpudevice::{convert_texture_format, convert_texture_view_dimension};
+use crate::dom::gpudevice::{convert_texture_format, convert_texture_view_dimension, GPUDevice};
 use crate::dom::gputextureview::GPUTextureView;
 use dom_struct::dom_struct;
 use std::string::String;
-use webgpu::{wgt, WebGPU, WebGPUDevice, WebGPURequest, WebGPUTexture, WebGPUTextureView};
+use webgpu::{wgt, WebGPU, WebGPURequest, WebGPUTexture, WebGPUTextureView};
 
 #[dom_struct]
 pub struct GPUTexture {
     reflector_: Reflector,
     texture: WebGPUTexture,
     label: DomRefCell<Option<USVString>>,
-    device: WebGPUDevice,
+    device: Dom<GPUDevice>,
     #[ignore_malloc_size_of = "channels are hard"]
     channel: WebGPU,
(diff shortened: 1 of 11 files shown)


components/script/dom/gputexture.rs
```diff
@@ -152,31 +153,50 @@ impl GPUTextureMethods for GPUTexture {
...
             base_mip_level: descriptor.baseMipLevel,
-            level_count: if descriptor.mipLevelCount == 0 {
-                self.mip_level_count - descriptor.baseMipLevel
-            } else {
-                descriptor.mipLevelCount
-            },
+            level_count: descriptor
+                .mipLevelCount
+                .map(|count| {
+                    if count == 0 {
```
kvark, line 159: we shouldn't have the `0` check now that the `None` is used

components/webgpu/lib.rs
```diff
@@ -1088,7 +1110,7 @@ impl<'a> WGPU<'a> {
                         data,
                     } => {
                         let global = &self.global;
-                        gfx_select!(queue_id => global.queue_write_buffer(
+                        let _ = gfx_select!(queue_id => global.queue_write_buffer(
```
kvark, line 1113: TODO: handle OOM here?

kvark (changes requested):
Looks good for most parts!

components/webgpu/lib.rs
kunalmohan: Yep, I'll add it. We only report errors for `Createxxx` operations at the moment.

components/script/dom/gputexture.rs
```diff
@@ -152,31 +153,38 @@ impl GPUTextureMethods for GPUTexture {
                 GPUTextureAspect::Depth_only => wgt::TextureAspect::DepthOnly,
             },
             base_mip_level: descriptor.baseMipLevel,
-            level_count: if descriptor.mipLevelCount == 0 {
-                self.mip_level_count - descriptor.baseMipLevel
-            } else {
-                descriptor.mipLevelCount
-            },
+            level_count: descriptor.mipLevelCount.map_or(
```
kvark, line 156: Actually, you don't need to compute this. Passing `0` down to `wgpu-core` would mean the same.
Perhaps, this and the layer count computation should look like:
```rust
level_count: match descriptor.mipLevelCount {
  Some(count) if count != 0 && descriptor.baseMipLevel + count <= self.mip_level_count => NonZeroU32::new(count),
  None if descriptor.baseMipLevel < self.mip_level_count => None,
  _ => { generate an error },
},
```

kunalmohan: I didn't do this because we won't be sending a `createTextureView` message to server in last case and an error Id won't be registered.

kunalmohan: I think `descriptor.mapLevelCount.and_then(NonZeroU32::new)` would work, since the validation is done in wgpu-core

kvark: ok, so here we want to send *something* to wgpu-core, right?
`descriptor.mipLevelCount.and_then(|count| NonZero::new(if count == 0 { !0 } else {count})`?

The problem with the current code is that `0` count will turn into None and result in automatic mipmapping, for example, while the expected outcome is an error

kvark (approved):

PR #27456, opened 2020-07-30
Get the gstreamer plugin to stream Hubs rooms

asajeffrey:
This is thje changes needed to the gstreamer plugin to stream a Hubs room:
- Add the ability to specify servo prefs as a gstreamer plugin property
- Fix a deadlock caused by using  bounded channel, which Hubs hits because it takes ~30s to enter webxr.
Also add a recipe for streaming Hubs.

---

- [x] These changes do not require tests because we don't test the gstreamer plugin


jdm:
@bors-servo retry
* #24622

PR #27466, opened 2020-07-31
Do not instanciate random ServoControls

paulrouget:
That's embarrassing. We would create a new instance of Servo every time we would call `ServoControl()` (instead of `servoControl()`)… so very often.


paulrouget:
Happens when we open the devtools for example.

jdm:
@bors-servo retry
* #25522 #23290

PR #27478, opened 2020-08-01
Output to stdout instead of calling `OutputDebugString`

### LINKED ISSUE #27441: Make C++ log function use stdout
Right now https://github.com/servo/servo/blob/08366323ef4ac77aa812d44499037bab4bc60eeb/support/hololens/ServoApp/strutils.h#L21 uses OutputDebugString directly. Since we have stdout redirect to OutputDebugString (as well as log to a file) in [this code](https://github.com/servo/servo/blob/08366323ef4ac77aa812d44499037bab4bc60eeb/support/hololens/ServoApp/ServoControl/Servo.cpp#L100-L115), we should be able to use printf from that log code and get more useful output in stdout.txt as a benefit.

cc @paulrouget

camelid:
Output to stdout instead of calling `OutputDebugString` since stdout redirects to `OutputDebugString` anyway and also to a log file.

---

- [x] These changes fix #27441 (GitHub issue number if applicable)

r? @jdm

### DIFF
--- a/support/hololens/ServoApp/strutils.h
+++ b/support/hololens/ServoApp/strutils.h
@@ -18,5 +18,5 @@ std::wstring format(const std::wstring &txt, Args... args) {
 }
 
 template <typename... Args> void log(const std::wstring &txt, Args... args) {
-  OutputDebugString((format(txt, args...) + L"\r\n").c_str());
+  std::cout << (format(txt, args...) + L"\r\n").c_str();
 }
--- a/support/hololens/ServoApp/strutils.h
+++ b/support/hololens/ServoApp/strutils.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <iostream>
+
 template <typename... Args>
 std::wstring format(const std::wstring &txt, Args... args) {
   size_t size = swprintf(nullptr, 0, txt.c_str(), args...) + 1;


support/hololens/ServoApp/strutils.h
```diff
@@ -18,5 +18,5 @@ std::wstring format(const std::wstring &txt, Args... args) {
 }
 
 template <typename... Args> void log(const std::wstring &txt, Args... args) {
-  OutputDebugString((format(txt, args...) + L"\r\n").c_str());
+  std::cout << (format(txt, args...) + L"\r\n").c_str();
```
camelid, line 21: Do we want to use `cout` or would `printf` be better? I don't know C++ that well, but I know some C (and Rust of course!)

paulrouget: You'll be missing a `#include` here (iostream iirc).

… and I feel like the redirection doesn't work.

camelid: Thanks! Just added the include. It's weird how running `./mach build -d` didn't give a warning. (Oh how I miss Rust's module system...)

camelid: > … and I feel like the redirection doesn't work.

Hmm, what do you mean? I was just implementing what @jdm suggested in #27478

camelid: I just realized that `./mach build -d` might not have warned me because the header may have been included in a file that already had `#include <iostream>`...

paulrouget: `./mach build -d` doesn't build this code.

See https://github.com/servo/servo/wiki/Building-for-UWP

> > … and I feel like the redirection doesn't work.

> Hmm, what do you mean? I was just implementing what @jdm suggested in #27441

I haven't investigated, but this doesn't appear to work.

mrobinson:
Closing this since it changes code that has now been removed.

PR #27480, opened 2020-08-02
WebGPU-Report errors in Promise or void returning operations

kunalmohan:
This also updates GPUBuffer mapping to match latest spec.

r?@kvark

---


components/webgpu/lib.rs
```diff
@@ -67,6 +67,8 @@ pub enum WebGPURequest {
     BufferMapAsync {
         sender: IpcSender<WebGPUResponseResult>,
         buffer_id: id::BufferId,
+        device_id: id::DeviceId,
+        scope_id: Option<u64>,
```
kvark, line 71: If we are sending the scope ID with every operation, maybe we should move it out of the num?
Also, consider changing it to `Option<NonZeroU64>` to reduce the size

kvark (approved):

kvark (approved):

jdm:
@bors-servo retry
* network hiccough

kunalmohan:
Looks like CI doesn't want any more of my PRs to merge :laughing:

jdm:
@bors-servo retry
* #25522

PR #27207, opened 2020-07-08
Update various dependencies

jdm:
This brings us more up to date for dependencies that rely on font-related crates.

PR #27500, opened 2020-08-04
Revert webrender upgrade

### LINKED ISSUE #27492: Nightly FxR closes on startup
Installing and launching the latest UWP nightly doesn't work, and the stdout.log doesn't show anything interesting.

jdm:
These commits revert the webrender and font-related updates from #27207. This will unbreak our FxR nightly builds while we continue to investigate the underlying crash in ANGLE. Unfortunately it also returns us to the status quo of broken canvas text rendering on UWP, but there's no way to upgrade font-kit without any of the other crates because of the underlying freetype dependency changes that happened.

Fixes #27492.


jdm:
@bors-servo retry
* #24622

PR #27505, opened 2020-08-05
Do not load srcdoc iframes synchronously

### LINKED ISSUE #24718: srcdoc iframes navigate synchronously
srcdoc iframe support was added in https://github.com/servo/servo/pull/24576. We load them the same way we load about:blank - we punch a hole into ScriptThread from the HTMLIFrameElement code, create an HTML parser and seed it with the string that's available. https://searchfox.org/mozilla-central/source/testing/web-platform/tests/html/interaction/focus/the-autofocus-attribute/skip-not-fully-active.html is a test that fails in this model, because by the time the promise_test runs there is no iframe element any more because the srcdoc's script element has executed synchronously. This suggests that we should be doing a round trip to the constellation like other iframe navigations instead.

utsavoza:
- [x] These changes fix #24718

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -1090,9 +1090,12 @@ where
         let (event_loop, host) = match sandbox {
             IFrameSandboxState::IFrameSandboxed => (None, None),
             IFrameSandboxState::IFrameUnsandboxed => {
-                // If this is an about:blank load, it must share the creator's event loop.
-                // This must match the logic in the script thread when determining the proper origin.
-                if load_data.url.as_str() != "about:blank" {
+                // If this is an about:blank or about:srcdoc load, it must share the creator's
+                // event loop. This must match the logic in the script thread when determining
+                // the proper origin.
+                if load_data.url.as_str() != "about:blank" &&
+                    load_data.url.as_str() != "about:srcdoc"
+                {
                     match reg_host(&load_data.url) {
                         None => (None, None),
                         Some(host) => {
--- a/components/script/dom/htmliframeelement.rs
+++ b/components/script/dom/htmliframeelement.rs
@@ -248,7 +248,7 @@ impl HTMLIFrameElement {
             load_data.srcdoc = String::from(element.get_string_attribute(&local_name!("srcdoc")));
             self.navigate_or_reload_child_browsing_context(
                 load_data,
-                NavigationType::InitialAboutBlank,
+                NavigationType::Regular,
                 HistoryEntryReplacement::Disabled,
             );
             return;
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -1503,10 +1503,13 @@ impl ScriptThread {
                         ScriptThreadEventCategory::AttachLayout,
                         Some(pipeline_id),
                         || {
-                            // If this is an about:blank load, it must share the creator's origin.
-                            // This must match the logic in the constellation when creating a new pipeline
-                            let origin = if new_layout_info.load_data.url.as_str() != "about:blank"
-                            {
+                            // If this is an about:blank or about:srcdoc load, it must share the
+                            // creator's origin. This must match the logic in the constellation
+                            // when creating a new pipeline
+                            let not_an_about_blank_and_about_srcdoc_load =
+                                new_layout_info.load_data.url.as_str() != "about:blank" &&
+                                    new_layout_info.load_data.url.as_str() != "about:srcdoc";
+                            let origin = if not_an_about_blank_and_about_srcdoc_load {
                                 MutableOrigin::new(new_layout_info.load_data.url.origin())
                             } else if let Some(parent) =
                                 new_layout_info.parent_info.and_then(|pipeline_id| {
--- a/tests/wpt/metadata/cookies/cookie-enabled-noncookie-frame.html.ini
+++ b/tests/wpt/metadata/cookies/cookie-enabled-noncookie-frame.html.ini
@@ -1,0 +1,5 @@
+[cookie-enabled-noncookie-frame.html]
+  expected: TIMEOUT
+  [navigator.cookieEnabled behavior on frames without cookie access]
+    expected: TIMEOUT
+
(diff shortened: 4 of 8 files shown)


jdm:
There are more errors and timeouts than passes, which makes me suspicious. Could you figure out what's changed in these tests with these changes?
```
  ▶ ERROR [expected OK] /html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html
  │   → i.contentDocument is null
  │ 
  └ i.onload@http://web-platform.test:8000/html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html:21:9
  ▶ Unexpected subtest result in /html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html:
  │ TIMEOUT [expected FAIL] Sandboxed => unsandboxed popup
  └   → Test timed out
  ▶ Unexpected subtest result in /html/interaction/focus/the-autofocus-attribute/skip-not-fully-active.html:
  └ PASS [expected FAIL] Autofocus element in not-fully-active document should be skipped while flusing.

  ▶ Unexpected subtest result in /custom-elements/perform-microtask-checkpoint-before-construction.html:
  └ PASS [expected FAIL] HTML parser must perform a microtask checkpoint before constructing a custom element
  ▶ Unexpected subtest result in /custom-elements/perform-microtask-checkpoint-before-construction.html:
  └ PASS [expected FAIL] HTML parser must perform a microtask checkpoint before constructing a custom element during the adoption agency algorithm
  ▶ TIMEOUT [expected OK] /referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html
  │ 
  │ TIMEOUT [expected FAIL] Sandboxed iframe with opaque origin doesn't send referrers.
  └   → Test timed out
  │ TIMEOUT [expected PASS] Sandboxed iframe with tuple origin sends referrers.
  └   → Test timed out

  ▶ TIMEOUT [expected OK] /cookies/cookie-enabled-noncookie-frame.html
  │ 
  ▶ Unexpected subtest result in /cookies/cookie-enabled-noncookie-frame.html:
  │ TIMEOUT [expected PASS] navigator.cookieEnabled behavior on frames without cookie access
  └   → Test timed out
```

utsavoza:
@jdm, afaict for sandboxed iframes we are currently spawning a new script thread and incorrectly kicking off fetch for srcdoc page load here:

https://github.com/servo/servo/blob/90aa0a3d41bc0ff173d061f42911bd72e0dfca6a/components/script/script_thread.rs#L820
 
which results in this error:

https://github.com/servo/servo/blob/90aa0a3d41bc0ff173d061f42911bd72e0dfca6a/components/net/fetch/methods.rs#L815

and subsequently timing out the tests.

I am not sure at the moment how this might pan out, but I think we might need to handle case for `about:srcdoc` during the creation of the script thread in order to prevent the `ScriptThread::pre_page_load` step and instead call this:

https://github.com/servo/servo/blob/90aa0a3d41bc0ff173d061f42911bd72e0dfca6a/components/script/script_thread.rs#L3906

to begin page load directly.

utsavoza:
I believe the above test failures are all concerning sandboxed iframes:

```
▶ ERROR [expected OK] /html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html
│   → i.contentDocument is null
│ 
└ i.onload@http://web-platform.test:8000/html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html:21:9
▶ Unexpected subtest result in /html/browsers/sandboxing/sandbox-disallow-scripts-via-unsandboxed-popup.tentative.html:
│ TIMEOUT [expected FAIL] Sandboxed => unsandboxed popup
└   → Test timed out
```
Now that sandboxed iframes for `about:srcdoc` goes through regular navigation, the origin of `about:srcdoc` differs from the `about:blank` iframe which by default prevents the access to iframe's contentDocument because documents for both the iframes now reside in different script threads.

https://github.com/servo/servo/blob/7499b0736e0ff8f413aa8f5a8c789342b8d207bc/components/script/dom/htmliframeelement.rs#L550-L553

From what I can tell, we currently consider sandbox attribute values to only determine whether the iframe is sandboxed or not. The above test makes use of `allow-same-origin` attribute value which should basically remove the restriction for the parent to treat the iframe as coming from different origin. One way to achieve this would be to not spawn a separate script thread in
case `allow-same-origin` is present (which was the case earlier).

<br>

```
▶ TIMEOUT [expected OK] /referrer-policy/generic/sandboxed-iframe-with-opaque-origin.html
│ 
│ TIMEOUT [expected FAIL] Sandboxed iframe with opaque origin doesn't send referrers.
└   → Test timed out
│ TIMEOUT [expected PASS] Sandboxed iframe with tuple origin sends referrers.
└   → Test timed out
```
There are two possible reasons why this test can fail:
1. `about:srcdoc` not being loaded entirely for sandboxed iframes due to reasons mentioned in [previous comment](https://github.com/servo/servo/pull/27505#issuecomment-669434250).
2.  after considering the [changes suggested in previous comment](https://github.com/servo/servo/pull/27505#issuecomment-669434250), `about:srcdoc` load can further fail to resolve its document's base url correctly.

https://github.com/servo/servo/blob/7499b0736e0ff8f413aa8f5a8c789342b8d207bc/components/script/dom/htmlscriptelement.rs#L743-L751

The scripts in the test use relative paths which should be resolved correctly against the [document's browsing context's container document](https://github.com/servo/servo/blob/7499b0736e0ff8f413aa8f5a8c789342b8d207bc/components/script/dom/document.rs#L713-L721) when `allow-same-origin` attribute value is provided. However, that's not the case because the container document now resides in a separate script thread.

<br>

```
▶ TIMEOUT [expected OK] /cookies/cookie-enabled-noncookie-frame.html
│ 
▶ Unexpected subtest result in /cookies/cookie-enabled-noncookie-frame.html:
│ TIMEOUT [expected PASS] navigator.cookieEnabled behavior on frames without cookie access
└   → Test timed out
```
This failure again resulted from `about:srcdoc` not being loaded entirely for sandboxed iframes.

Additionally, I *suspect* there will some more test failures that currently timeout for sandboxed iframes once we are able to load `about:srcdoc` from a new script thread. Also, I am not sure if it's worth incorporating these changes over the current behavior until we have a more complete implementation for iframe sandbox attribute values.

jdm:
Thanks for the thorough analysis! I'm convinced that we can accept these changes and consider sandbox changes separately.

jdm:
@bors-servo retry
* #25522

PR #27506, opened 2020-08-05
Keep C++ pref value in memory longer

paulrouget:


jdm:
@bors-servo r+
This looks ok as a stop-gap solution for memory corruption. I look forward to replacing these with a pervasive solution for passing pointers to embedder-allocated memory that is deallocated when the engine no longer needs it.

PR #27534, opened 2020-08-06
Move action buttons above invalid certificate details.

### LINKED ISSUE #27520: Self-signed cert options should be near the top of the page
On the HL2, it's possible for the cert contents to be require scrolling, so the buttons to handle the error are not in view. This could be confusing for developers.

jdm:
Since the certificate details can fill the screen and require scrolling, let's move the user's choices so they're always in view when the SSL error page first loads.

---
- [x] These changes fix #27520

### DIFF
--- a/resources/badcert.html
+++ b/resources/badcert.html
@@ -4,9 +4,9 @@
 </head>
 <body>
   <p>${reason}</p>
-  <pre id="bytes">${bytes}</pre>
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo retry
* #24622

CYBAI:
@bors-servo retry

- network

PR #27536, opened 2020-08-06
Major fixes in error reporting in GPUCommandEncoder and ErrorScope Model

kunalmohan:
1. Update wgpu to use the error model on wgpu-core side. Register error Ids separately.
2. ~~Record errors only in `GPUCommandEncoder.finish()`. Errors are no longer recorded in ErrorScopes in transfer commands or while recording passes. Any errors that occur are stored on the server-side in `error_command_encoders: HashMap<id::CommandEncoderId, String>` and reported on `CommandEncoderFinish`. Note: This should be reverted when the spec gets updated.~~
3. Correct a major flaw in ErrorScope Model. If multiple operations are issued inside scope and an early operation fails, the promise resolves and script execution continues. The scope, however, was not popped until the results of all its operations were received. This meant that if the user issues another operation, it was assumed to be issued in an error scope that has already been popped by the user, which led to the failure of a number of tests. This is solved by storing a `popped` boolean to check whether `popErrorScope()` has been called on a scope or not. Operation is now issued in the lastest scope for which `popped == false`.

One of the tests used to crash, but it no longer does (All subtests under it fail now). That explains the large number of failing test expectations that have been added. Most of them fail due to the tests being outdated. I'll switch to the updated branch in the next PR.

r?@kvark

---

### DIFF
--- a/components/webgpu/lib.rs
+++ b/components/webgpu/lib.rs
@@ -744,7 +744,7 @@ impl<'a> WGPU<'a> {
                     },
                     WebGPURequest::DestroyBuffer(buffer) => {
                         let global = &self.global;
-                        gfx_select!(buffer => global.buffer_destroy(buffer, false));
+                        gfx_select!(buffer => global.buffer_drop(buffer, false));
                     },
                     WebGPURequest::DestroySwapChain {
                         external_id,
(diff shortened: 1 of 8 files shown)


kvark:
@kunalmohan reviewing the spec now, I don't think it says anything about errors being deferred to `finish()`. e.g. [copyBufferToTexture](https://gpuweb.github.io/gpuweb/#dom-gpucommandencoder-copybuffertotexture) says:
> If any of the following conditions are unsatisfied, generate a validation error and stop.

So it doesn't look like any spec changes are needed?

kunalmohan:
Ohhh...I got confused with Errorhandling.md. Should `endpass()` record any errors? cts certainly expects endPass to not record any errors. I wonder if that changed on main branch.

components/webgpu/lib.rs
```diff
@@ -82,33 +82,33 @@ pub enum WebGPURequest {
     },
     CopyBufferToBuffer {
```
kvark, line 83: nit: might be good to move them together as `Copy { command_encoder_id: id::CommandEncoderId, op: Option<CopyOperation> }` or something like this. That would allow the server side to handle the errors more uniformly.

kvark (approved):
Looks reasonable, just one suggestion.

kunalmohan:
We now record error in all transfer commands, `GPUCommandEncoder.beginxxxPass()` and `endPass()`. If `BeginPass()` has a validation error, we don't create a new `RenderPass` instead an invalid `GPURenderPassEncoder` is returned with `None` in its `render_pass` slot.

PR #27538, opened 2020-08-06
Make textures that come from webxr invalid outside an rAF

asajeffrey:
Implements "The colorTexture and depthStencilTexture objects MUST only be used during the XR animation frame of the current session and MUST be made invalid once the XR animation frame completes." from https://immersive-web.github.io/layers/#xrwebglsubimagetype

---

- [x] There are no tests for these changes because we can't test this until we have projection layers


asajeffrey:
I was going to say "tests needed" but I don't think we can test this until we have projection layers.

PR #27549, opened 2020-08-08
Update WebGPU CTS to main branch

kunalmohan:
`main` branch in https://github.com/gpuweb/cts is closer to our implementation and it makes more sense to use that.

r?@kvark

---


kunalmohan:
@bors-servo try=wpt-mac

A number of tests have been added/deleted, so we need to check again which ones crash on metal backend only.

kunalmohan:
I am worried about a large number of `TIMEOUT`s that I get, some of which intermittently `PASS/FAIL`.

kvark (approved):

kunalmohan:
@bors-servo try=wpt-mac

CI should hopefully be happy. Let's check this before approving/merging.

kunalmohan:
There are no unexpected results in the webgpu cts now. The failure comes from an unrelated test (probably intermittent).

@kvark This is ready to merge now.

PR #27558, opened 2020-08-10
Fix missing canvas text on UWP

### LINKED ISSUE #27515: CanvasRenderingContext2D.fillText doesn't draw text in UWP
Any attempt to draw text on a 2d canvas in Firefox Reality will run into problems loading fonts and silently discard the operation. This will be fixed by updating to reverting the commits in https://github.com/servo/servo/pull/27500, but that is blocked by fixing https://github.com/servo/servo/issues/27492.

jdm:
This addresses the missing fonts under UWP by updating font-kit and assorted dependencies that all depend on older incompatible versions of freetype. This also works around #27492 by using a fork of webrender that cherry-picks required dependency updates without bringing in the commit that exposes an ANGLE crash.

---
- [x] These changes fix #27515
- [x] These changes do not require tests because no UWP tests.


jdm:
@bors-servo retry
* #25522

PR #27634, opened 2020-09-12
preferences actor now returns real values.

### LINKED ISSUE #27510: devtools preference actor returns fake values
https://github.com/servo/servo/blob/90aa0a3d41bc0ff173d061f42911bd72e0dfca6a/components/devtools/actors/preference.rs#L33-L58

This should be using the preference API to retrieve the real values and sending them in the response messages.

sudarshan-reddy:
The preferences actor was previously returning mock values. This changes
picks up the pref_values from the pref_map which is global preferences
state and returns values from there.

- [X] These changes fix #27510 (GitHub issue number if applicable)

- [X] These changes do not require tests because none of the actors have tests.

### DIFF
--- a/components/devtools/actors/preference.rs
+++ b/components/devtools/actors/preference.rs
@@ -6,6 +6,8 @@ use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
 use crate::protocol::JsonPacketStream;
 use crate::StreamId;
 use serde_json::{Map, Value};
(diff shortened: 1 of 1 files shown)


sudarshan-reddy:
@jdm should be ready to review again.

jdm:
This looks good! Sorry for the long delay!
@bors-servo r+

PR #27642, opened 2020-09-15
[Do not merge] CI testing

SimonSapin:


SimonSapin:
I’ve temporarily disabled 5 out of the 9 macOS workers we have on CI, leaving 4 of them active. Let’s see how that affects cycle time with no other change.

@bors try=wpt-mac

SimonSapin:
On the last 4 PR that merged, the start-to-end time for WPT on mac (compilation not included) has been 34, 28, 29, and 29 minutes. This run with 4 workers instead of 9 took 56 minutes. It looks like chunking works predictably and the effect of removing workers is close to linear.

PR #27655, opened 2020-09-21
Fix wrong priority for XHR username/password

### LINKED ISSUE #24953: Wrong priority when user:pass@url string conflicts with XHR user and password parameters
WPT xhr/send-authentication-competing-names-passwords.htm does various combinations of specifying Basic authentication user/passwords, using url strings and explicit XHR parameters, and a few subcases end up sending the wrong thing.

ghostd:
Fix the wrong priority between the username/password from the URL and the username/password from the 'open' call

---

- [X] These changes fix #24953


jdm:
@bors-servo r+
Thanks! This indeed matches the spec at https://xhr.spec.whatwg.org/#the-open()-method.

PR #27670, opened 2020-09-25
Implements HTMLDetailsElement toggle

### LINKED ISSUE #22106: details element toggle not working
When I tried to click to toggle the following `details` element, it didn't work.

```html
<details>
    <summary>Details</summary>
    Something small enough to escape casual notice.
</details>
```

Built with latest master (4ec7dedce1) on macOS 10.13.6

(Not sure if #21336 can be resolved once this issue resolved)

ghostd:
Makes HTMLSummaryElement activatable to toggle the linked details

---

- [X] These changes fix #22106


ghostd:
The 'open' flag is updated, but the not the rendering... not sure how to debug this. These are my first days is the rust world and the servo code :)

components/script/dom/htmldetailselement.rs
```diff
@@ -49,6 +49,10 @@ impl HTMLDetailsElement {
             document,
         )
     }
+
+    pub fn toggle(&self) {
+        self.SetOpen(!self.Open());
```
jdm, line 54: Try adding `self.upcast::<Node>().dirty(NodeDamage::OtherNodeDamage)` and see if that fixes the rendering issues.

ghostd: Thanks a lot! It works!

jdm: Actually, let's move it to the attribute_mutated function so that doing things like `elem.removeAttribute('open')` or `elem.open = true` from JS also cause the rendering to update.

ghostd: Done (my manual test case works, and the toggleEvent.html WTP test is still "PASS")

components/script/dom/htmlelement.rs
```diff
@@ -783,6 +786,49 @@ impl HTMLElement {
...
+        let summary_node = self.upcast::<Node>();
+        if !summary_node.has_parent() {
+            return false;
+        }
+
+        // Step 2
+        let parent = &summary_node.GetParentNode().unwrap();
+
+        // Step 3
+        if parent.downcast::<Element>().unwrap().local_name() != &local_name!("details") {
```
jdm, line 819: This could be replaced by `parent.is::<HTMLDetailsElement>()` instead.

components/script/dom/htmlelement.rs
```diff
@@ -783,6 +786,49 @@ impl HTMLElement {
...
+            return false;
+        }
+
+        // Step 4 & 5
+        let first_summary_element = parent
+            .child_elements()
+            .find(|el| el.local_name() == &local_name!("summary"));
+        match first_summary_element {
+            Some(first_summary) => first_summary == DomRoot::from_ref(self.upcast::<Element>()),
```
jdm, line 828: I believe this could be `&*first_summary == self.upcast::<Element>())` instead.

jdm (changes requested):
Great work putting these changes together without much prior Servo experience!

jdm:
This PR will also need to be moved out of draft mode before it can be merged.

components/script/dom/htmlelement.rs
ghostd: This works, even i dont yet understand what "&*first_summary" means.

components/script/dom/htmlelement.rs
ghostd: Yep, much simpler to read, thanks :)

ghostd:
I fixed the unexpected "PASS"

jdm:
@bors-servo r+
Thanks for fixing this!

PR #27725, opened 2020-10-29
Use data_url::Mime to parse the MIME Types

### LINKED ISSUE #24923: Servo isn't strict enough about what XMLHttpRequest response XML it will accept
WPT xhr/responsexml-basic, responsexml-media-type, and responsexml-non-well-formed reveal cases where an XMLHttpRequest shouldn't be producing an XML document but is:
* an empty body
* a body that's syntactically XML, but with a Content-Type header that says it's something else
* lightly malformed XML
By WPT expectations and other browser behavior, these shouldn't parse into an XML document when they're the response to an XHR.

ghostd:
This commit follows the spectification https://fetch.spec.whatwg.org/#concept-header-extract-mime-type

This commit partially addresses #24923 (only the content type part). I'll try to fix the parser stuff in an other commit.

---

### DIFF
--- a/components/script/dom/headers.rs
+++ b/components/script/dom/headers.rs
@@ -10,8 +10,9 @@ use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::{is_token, ByteString};
 use crate::dom::globalscope::GlobalScope;
+use data_url::mime::Mime as DataUrlMime;
 use dom_struct::dom_struct;
-use http::header::{self, HeaderMap as HyperHeaders, HeaderName, HeaderValue};
+use http::header::{HeaderMap as HyperHeaders, HeaderName, HeaderValue};
 use net_traits::request::is_cors_safelisted_request_header;
 use std::cell::Cell;
 use std::str::{self, FromStr};
@@ -269,10 +270,7 @@ impl Headers {
 
     // https://fetch.spec.whatwg.org/#concept-header-extract-mime-type
     pub fn extract_mime_type(&self) -> Vec<u8> {
-        self.header_list
-            .borrow()
-            .get(header::CONTENT_TYPE)
-            .map_or(vec![], |v| v.as_bytes().to_owned())
+        extract_mime_type(&*self.header_list.borrow()).unwrap_or(vec![])
     }
 
     pub fn sort_header_list(&self) -> Vec<(String, Vec<u8>)> {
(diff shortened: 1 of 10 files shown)


components/script/dom/xmlhttprequest.rs
```diff
@@ -1585,14 +1603,20 @@ impl XMLHttpRequest {
         }
     }
 
+    /// <https://xhr.spec.whatwg.org/#response-mime-type>
+    fn response_mime_type(&self) -> Option<Mime> {
+        return extract_mime_type(&self.response_headers.borrow())
+            .map(|mime_as_bytes| String::from_utf8(mime_as_bytes).unwrap().parse().ok())
```
jdm, line 1609: Let's use `unwrap_or_default()` instead.

components/script/dom/headers.rs
```diff
@@ -469,3 +467,69 @@ pub fn is_obs_text(x: u8) -> bool {
...
+            Err(_) => continue,
+            Ok(temp_mime) => {
+                let temp_essence = format!("{}/{}", temp_mime.type_, temp_mime.subtype);
+
+                // Step 6.2
+                if temp_essence == "*/*" {
+                    continue;
+                }
+
+                let temp_charset = &temp_mime.get_parameter(&"charset".to_string());
```
jdm, line 502: The to_string shouldn't be necessary, I believe.

jdm (changes requested):
These changes look sensible! Thanks!

components/script/dom/xmlhttprequest.rs
ghostd: Fixed

components/script/dom/headers.rs
ghostd: You're right! Fixed

ghostd:
I rebased the branch.

jdm:
Looks lik ea `mach fmt` is required.

ghostd:
My bad :-/ Fixed

ghostd:
I'll investigate the regression this week.

ghostd:
I fixed the regression and rebased onto master

jdm:
```
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.worker.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;": compileStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.worker.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;": instantiateStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.worker.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;x": compileStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.worker.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;x": instantiateStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;": compileStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;": instantiateStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;x": compileStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code
  ▶ Unexpected subtest result in /wasm/webapi/contenttype.any.html:
  │ FAIL [expected PASS] Response with Content-Type "application/wasm;x": instantiateStreaming
  └   → assert_unreached: Should have rejected: undefined Reached unreachable code

  ▶ Unexpected subtest result in /xhr/overridemimetype-blob.html:
  └ PASS [expected FAIL] Use text/xml as fallback MIME type
  ▶ Unexpected subtest result in /xhr/overridemimetype-blob.html:
  └ PASS [expected FAIL] Use text/xml as fallback MIME type, 2
```

ghostd:
i'm not sure why the test `wasm/webapi/contenttype.any.js` considers the `application/wasm;` and `application/wasm;x` as invalid mime types. According to the [WebAssembly Web API](https://www.w3.org/TR/wasm-web-api-1/#ref-for-concept-header-extract-mime-type), to extract the MIME type, we should use the same algorithm than the [Fetch API](https://fetch.spec.whatwg.org/#ref-for-parse-a-mime-type%E2%91%A0). The latter points to the [parsing algorithm of MIME Sniffing](https://mimesniff.spec.whatwg.org/#parse-a-mime-type) which is implemented by [data_url](https://github.com/servo/rust-url/blob/master/data-url/src/mime.rs#L28) (purpose of this PR)

IMHO, `application/wasm;x` and `application/wasm;` and `application/wasm` are structurally equivalent: essence == 'application/wasm' and no valid parameters. To fix this, i can get the string version of the header (without parsing) before the check... but this deviates from the spec.

What do you think?

CYBAI:
I haven't dive deeply into the spec though, I just noticed the note below point 3, it says "extra parameters are not allowed, including the empty `application/wasm;`.". Maybe that's the reason why `application/wasm;` is invalid? 👀?

image

ghostd:
Thanks, i missed this, need some rest :-/
But now, how do you imagine the code since for the MIME parsing algo this is the same. I didn't see a flag meaning 'this string had additionnal data after the essence'. The parameters array is always empty (at least)

ghostd:
Ok, i'm not the only one to think it's a bit strict: https://github.com/WebAssembly/spec/issues/1138

jdm:
Good find! Want to mark these as expected failures and merge this, in that case?

ghostd:
Done

Should I open an issue to follow this "regression"?

jdm:
Yes please!
@bors-servo try=wpt

ghostd:
Rebased

ghostd:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 57: ordinal not in range(128)

  File "/Users/worker/tasks/task_1605728842/repo/python/servo/testing_commands.py", line 600, in filter_intermittents
    format(actual_failures, description)
  File "/Users/worker/tasks/task_1605728842/repo/python/servo/testing_commands.py", line 588, in format
    file.write(formatted.encode("utf-8"))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/codecs.py", line 357, in write
```
For the Mac build :-/

jdm:
Yeah, you have to look at the filtered summary instead (https://github.com/servo/servo/issues/25233).

```
  ▶ TIMEOUT [expected ERROR] /_webgl/conformance/textures/image_bitmap_from_canvas/tex-2d-luminance_alpha-luminance_alpha-unsigned_byte.html
  ▶ FAIL [expected PASS] /css/CSS2/normal-flow/width-040.xht
  │   → /css/CSS2/normal-flow/width-040.xht ['7b6d4715bf74ba79b0c5f8cd6bbc78ebba36a265']
  └   → /css/CSS2/normal-flow/max-width-006-ref.xht ['d9da3f1daeb78b8416383c9c83c9dfdcb349fd38']
```
I believe those are known intermittent failures.

@bors-servo retry

PR #27749, opened 2020-11-11
Support gzip artifacts.

jdm:


jdm:
Try jobs show it's working as expected.
@bors-servo r+ try-

SimonSapin:
This should fix errors like https://community-tc.services.mozilla.com/tasks/H7nXSQ7SRY2FSXnKRGzJwg/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FH7nXSQ7SRY2FSXnKRGzJwg%2Fruns%2F0%2Fartifacts%2Fpublic%252Flogs%252Flive.log#L55-57 `fatal: too large to be a .git file: '../repo.bundle'`

The Taskcluster team deployed yesterday a new version of docker-worker that includes https://github.com/taskcluster/taskcluster/pull/2809 which compresses artifacts and serves them with the appropriate `content-encoding` header. However `curl` does not consider that header unless we ask it to.

PR #27760, opened 2020-11-14
Upgrade CI to Ubuntu 20.04 LTS

SimonSapin:
19.10 has been at end-of-life since July.

Also switch to an un-dated Docker Hub tag. Built docker images expire after a month, and when rebuilding we install the latest versions of additional packages. We might has well pick up-to-date versions of the base system as well.

We do stay within one Ubuntu release, though.

### DIFF
--- a/etc/taskcluster/docker/base.dockerfile
+++ b/etc/taskcluster/docker/base.dockerfile
@@ -1,6 +1,4 @@
-# Dummy comment to force an image rebuild for https://github.com/servo/servo/issues/27758
-
-FROM ubuntu:eoan-20200608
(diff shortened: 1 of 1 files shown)


SimonSapin:
```
E: Unable to locate package python-pip
E: Package 'python-virtualenv' has no installation candidate
```

Packaging of Python 2 v.s. 3 seems to have changed in Ubuntu 20.04. This will need a closer look.

SimonSapin:
`mach` managed to find `virtualenv` and stuff is building, looks good enough. Probably.

@bors-servo try- r+

PR #27766, opened 2020-11-15
Embedding support for offscreen-rendering

jdm:
The current embedding API assumes it should render directly to a native widget at all times. These changes expose an option to render to an offscreen surface when starting the engine, as well as a new WebrenderSurfman API for temporarily acquiring  the surface corresponding to the front buffer so it can be used for blitting purposes.

Demonstrated to work in https://github.com/jdm/servo-embedding-example.

---

### DIFF
--- a/components/webrender_surfman/lib.rs
+++ b/components/webrender_surfman/lib.rs
@@ -159,6 +159,23 @@ impl WebrenderSurfman {
             })
     }
 
+    /// Invoke a closure with the surface associated with the current front buffer.
+    /// This can be used to create a surfman::SurfaceTexture to blit elsewhere.
+    pub fn with_front_buffer<F: FnMut(&Device, Surface) -> Surface>(&self, mut f: F) {
+        let ref mut device = self.0.device.borrow_mut();
+        let ref mut context = self.0.context.borrow_mut();
+        let surface = device
+            .unbind_surface_from_context(context)
+            .unwrap()
+            .unwrap();
+        let surface = f(device, surface);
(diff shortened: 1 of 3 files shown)


jdm:
The first two commits are from #27717 and can be ignored.

asajeffrey:
Ah, interesting. My first reaction was "didn't I already do this?" since it's needed for the gstreamer servosrc plugin, but that's because I built that using libservo not libsimpleservo.

You can r=me.

Cargo.lock
```diff
@@ -7041,10 +7041,16 @@ dependencies = [
...
+name = "weezl"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3e2bb9fc8309084dd7cd651336673844c1d47f8ef6d2091ec160b27f5c4aa277"
+
 [[package]]
 name = "wgpu-core"
 version = "0.6.0"
-source = "git+https://github.com/gfx-rs/wgpu#e72724a6e393503c73f37e86aa9317a5c62e32b8"
+source = "git+https://github.com/gfx-rs/wgpu?rev=e72724a6e393503c73f37e86aa9317a5c62e32b8#e72724a6e393503c73f37e86aa9317a5c62e32b8"
```
atouchet, line 7053: Is it possible to update these to wgpu-core 0.6.5/wgpu-types 0.6.1 instead or is there an issue with using the current published versions in Servo?

jdm: There are compilation errors using more recent git revisions. I have not attempted to investigate them at all.

Issue #27791, opened 2020-11-22
about:srcdoc in scheme_fetch unsupported


jonathanKingston:
https://github.com/servo/servo/blob/66616c65a1fec6a9ecb9260ced98c4dc7a21fef0/components/net/fetch/methods.rs#L624 doesn't support srcdoc urls.

This is part of the test fails in /secure-contexts/basic-popup-and-iframe-tests.https.html

This gets called from components/net/resource_thread.rs fetch().

The test is creating srcdoc iframes.

jdm:
It's not clear to me what the issue is here - srcdoc frames shouldn't pass through the fetch infrastructure; it's handled by [this code](https://github.com/servo/servo/blob/ce62f6f203af0f729ebebf5dbc3a0109d661fd3c/components/script/script_thread.rs#L2529-L2530) instead.

jonathanKingston:
To be a little clearer the code is printing the message from `Response::network_error(NetworkError::Internal("Unexpected scheme".into()))`

Here is a trace when this condition is met:
```
 0:07.89 pid:1884780    0: servo::backtrace::print
 0:07.89 pid:1884780              at /home/jonathan/debugging/servo/servo/ports/winit/backtrace.rs:17:5
 0:07.89 pid:1884780    1: servo::main::{{closure}}
 0:07.89 pid:1884780              at /home/jonathan/debugging/servo/servo/ports/winit/main2.rs:166:21
 0:07.98 pid:1884780    2: std::panicking::rust_panic_with_hook
 0:07.98 pid:1884780              at /rustc/6c8927b0cf80ceee19386026cf9d7fd4fd9d486f/src/libstd/panicking.rs:530:17
 0:07.98 pid:1884780    3: std::panicking::begin_panic
 0:07.98 pid:1884780              at /rustc/6c8927b0cf80ceee19386026cf9d7fd4fd9d486f/src/libstd/panicking.rs:456:5
 0:07.99 pid:1884780    4: net::fetch::methods::scheme_fetch
 0:07.99 pid:1884780              at /home/jonathan/debugging/servo/servo/components/net/fetch/methods.rs:816:13
 0:08.00 pid:1884780    5: net::fetch::methods::main_fetch::{{closure}}
 0:08.00 pid:1884780              at /home/jonathan/debugging/servo/servo/components/net/fetch/methods.rs:292:13
 0:08.06 pid:1884780    6: core::option::Option<T>::unwrap_or_else
[...]

Issue #27268, opened 2020-07-14
SecureContext restriction for WebXR is ignored


jdm:
WebXR is only supposed to be exposed to HTTPS contexts, but we don't enforce this. It would be nice to be able to opt-out of this restriction for local development, however, possibly via a pref that would be persisted.

msub2:
@servo-highfive assign me

servo-highfive:
Hey @msub2! Thanks for your interest in working on this issue. It's now assigned to you!

PR #27793, opened 2020-11-22
PoC Secure Contexts

jonathanKingston:
Proof of concept implementation for implementing secure contexts.

## Blob URLs issue
Currently blobs don't work as the origin is created into an opaque which means the `is_potentially_trustworthy` won't pass.
This code creates script threads with new origins:
https://github.com/servo/servo/blob/a3e9095faa885cebf6e17c3c161871e4d896c70b/components/script/script_thread.rs#L807

Some solutions to this:
- Add to the `ImmutableOrigin` code an `is_data` that gets passed with the origin
- Pass this `is_data` flag into the `load_data` itself instead
- Use a `Url` instead for the origin

## Srcdoc issue
https://github.com/servo/servo/issues/27791 it became clear as part of this issue that srcdoc isn't handled in frames.

## Pref gating
As suggested in https://github.com/servo/servo/issues/27268 it might be useful for local development to have a pref on the idl gating to allow `!isSecureContext` pages to access APIs.

## SpiderMonkey creation options

Firefox passes this flag, it's not needed in this PoC as we use the `GlobalScope` to be the canonical source of if the code is considered a secure context. It may make sense to continue to pass it to spider monkey for consistency.

## Member conditions

It appears in the Member condition code in components/script/dom/bindings/codegen/CodegenRust.py it only handles being gated by one of these conditions. Should this be a follow up issue, obviously this exacerbates the issue?

---

### DIFF
--- a/components/constellation/browsingcontext.rs
+++ b/components/constellation/browsingcontext.rs
@@ -24,6 +24,9 @@ pub struct NewBrowsingContextInfo {
     /// Whether this browsing context is in private browsing mode.
     pub is_private: bool,
 
+    /// Whether this browsing context inherits a secure context.
+    pub inherited_secure_context: Option<bool>,
+
     /// Whether this browsing context should be treated as visible for the
     /// purposes of scheduling and resource management.
     pub is_visible: bool,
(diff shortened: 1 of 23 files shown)


jonathanKingston:
I think we can close this now as the other patch has advanced so much.

PR #27807, opened 2020-11-24
Replace generic 'value' variable name with 'id' and 'name' to reduce confusion

twilco:
I was digging through this code trying to fix a WPT and thought these could be named a bit more clearly.  `value` is vague, and the fact it was used in both cases falsely implied they were related.

---

- [x] These changes do not require tests because there are no functional changes.


CYBAI (approved):
Thanks! LGTM!

jdm:
@bors-servo retry
* #24622

PR #27851, opened 2020-11-26
Fix some 'nosniff' tests

ghostd:
This PR implements some standard function to work with headers, hence it fixes some 'nosniff' test/

I create a PR as Draft because i have some questions:
* i added the code info the `net` module instead of `script::dom::headers` to avoid some cyclic dependencies... but i'm not sure about this choice. Is there another potential location to put this code?
* i used String object to match the spectification (it states to collect sequences of "code points", but some code use byte arrays. Should i change the code to work with bytes?

---


agniutkarsh (approved):

ghostd:
I rebased onto master.

 I think i'll needs these changes for future works.

jdm:
@bors-servo r+
Thanks!

PR #27852, opened 2020-11-27
Fix memory profile crash

### LINKED ISSUE #27836: Memory profiler crashes immediately
Running servo with `-m` on macOS yields:
```
Begin memory reports 5
|
Stack trace for thread "ScriptThread PipelineId { namespace_id: PipelineNamespaceId(1), index: PipelineIndex(1) }"
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: servo::backtrace::print
   4: servo::install_crash_handler::handler
   5: __sigtramp

```

joseph62:
Added a callback for determining which objects should be included in the memory profiling that was removed by https://github.com/servo/servo/commit/74c1e00d8163f255bb4141ff3549bbdedd7ea766#diff-86033f852232a3ea1cb2f43d2f255850fb6d091c6eeba07213152e13b7006026L219

Also added `is_platform_object_static` that did not require a JSContext to be used in the callback.

---

- [X] These changes fix #27836 (GitHub issue number if applicable)

Wasn't sure about tests for these changes. Some direction for that would be great. Thank you!


jdm:
@bors-servo r+
Nice solution for avoiding duplication!

PR #27948, opened 2020-12-14
Remove LLVM package.

### LINKED ISSUE #27545: Upgrade to llvm 10
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1657913, the latest MSVC update requires clang 10. We should upgrade our windows package soon to avoid windows developers getting stuck.

jdm:
I'm curious if we can depend on the clang binaries provided by MSVC instead of downloading a separate LLVM package. This would fix #27545.


jdm:
lld-link.exe causes some problems, unfortunately. https://github.com/rust-lang/rust/issues/71520 would help here.

PR #27952, opened 2020-12-15
Implement named access on the window object

### LINKED ISSUE #27949: Named window getter is missing
From https://html.spec.whatwg.org/multipage/window-object.html#the-window-object, we are currently missing this bit:
```
  getter object (DOMString name);
  // Since this is the global object, the IDL named getter adds a NamedPropertiesObject exotic
  // object on the prototype chain. Indeed, this does not make the global object an exotic object.
  // Indexed access is taken care of by the WindowProxy exotic object.
```
and `LegacyUnenumerableNamedProperties`

Spec:
* https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object
* https://heycam.github.io/webidl/#LegacyUnenumerableNamedProperties

https://github.com/servo/servo/pull/25562 and https://github.com/servo/servo/pull/21869 are two previous WIPs to implement this feature.

pylbrecht:
---

- [ ] These changes fix #27949

- [ ] There are tests for these changes


pylbrecht:
This is based on https://github.com/servo/servo/tree/named-getter.

I opened this PR to keep track of changes and related discussions (..and run CI at some point).

pylbrecht:
The build currently fails (which I kind of expected for a WIP):
```
error[E0609]: no field `name_map` on type `&dom::window::Window`
    --> components/script/dom/window.rs:1483:29
     |
1483 |         let name_map = self.name_map.borrow();
     |                             ^^^^^^^^ unknown field
     |
     = note: available fields are: `globalscope`, `script_chan`, `task_manager`, `navigator`, `image_cache` ... and 56 others

error[E0609]: no field `id_map` on type `&dom::window::Window`
    --> components/script/dom/window.rs:1495:27
     |
1495 |         let id_map = self.id_map.borrow();
     |                           ^^^^^^ unknown field
     |
     = note: available fields are: `globalscope`, `script_chan`, `task_manager`, `navigator`, `image_cache` ... and 56 others

For more information about this error, try `rustc --explain E0609`.
```
I'm going to have a look at the related spec(s) to figure out what `id_map` and `name_map` are supposed to be.

jdm:
I wonder if those should be referring to Document instead of Window (https://github.com/servo/servo/blob/3bf1085c72d1bf05507c4755709e9c25c6ca4e7f/components/script/dom/document.rs#L247-L248)?

pylbrecht:
I implemented getters for `id_map` and `name_map` to see if the build succeeds, but that got me into scary territory:
```
error[E0133]: use of extern static is unsafe and requires unsafe function or block
   --> components/script/window_named_properties.rs:174:11
    |
174 |     cOps: &ProxyClassOps,
    |           ^^^^^^^^^^^^^^ use of extern static
    |
    = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior

error[E0133]: use of extern static is unsafe and requires unsafe function or block
   --> components/script/window_named_properties.rs:176:10
    |
176 |     ext: &ProxyClassExtension,
    |          ^^^^^^^^^^^^^^^^^^^^ use of extern static
    |
    = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior

error[E0133]: use of extern static is unsafe and requires unsafe function or block
   --> components/script/window_named_properties.rs:177:11
    |
177 |     oOps: &ProxyObjectOps,
    |           ^^^^^^^^^^^^^^^ use of extern static
    |
    = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior

error: aborting due to 3 previous errors
```

jdm:
> I implemented getters for `id_map` and `name_map` to see if the build succeeds, but that got me into scary territory:
> 
> ```
> error[E0133]: use of extern static is unsafe and requires unsafe function or block
>    --> components/script/window_named_properties.rs:174:11
>     |
> 174 |     cOps: &ProxyClassOps,
>     |           ^^^^^^^^^^^^^^ use of extern static
>     |
>     = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior
> 
> error[E0133]: use of extern static is unsafe and requires unsafe function or block
>    --> components/script/window_named_properties.rs:176:10
>     |
> 176 |     ext: &ProxyClassExtension,
>     |          ^^^^^^^^^^^^^^^^^^^^ use of extern static
>     |
>     = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior
> 
> error[E0133]: use of extern static is unsafe and requires unsafe function or block
>    --> components/script/window_named_properties.rs:177:11
>     |
> 177 |     oOps: &ProxyObjectOps,
>     |           ^^^^^^^^^^^^^^^ use of extern static
>     |
>     = note: extern statics are not controlled by the Rust type system: invalid data, aliasing violations or data races will cause undefined behavior
> 
> error: aborting due to 3 previous errors
> ```

Luckily, all of those just require wrapping it in a `unsafe` like `cOps: unsafe { &ProxyClassOps }` (or possibly the surrounding struct value instead if that doesn't work).

pylbrecht:
That did it, thanks!

Are there any related tests I can run?

pylbrecht:
So, I started to work on this TODO:
```rust
// TODO: Handle the document-tree child browsing context name property set.
```
The [spec for named-access-on-the-window-object](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object) states:
> 1. If window's browsing context is null, then return the empty list.

`NamedGetter()` returns `Option<NonNull<JSObject>>`, but I have no idea what DOM type (`#[dom_struct]`) corresponds to "empty list" here. I skimmed through the [`script::dom` docs](https://doc.servo.org/script/dom/index.html), but I feel I'm still missing some bits and pieces.

Can you help me out?

pshaughn:
> 
> 
> So, I started to work on this TODO:
> 
> ```rust
> // TODO: Handle the document-tree child browsing context name property set.
> ```
> 
> The [spec for named-access-on-the-window-object](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object) states:
> 
> > 1. If window's browsing context is null, then return the empty list.
> 
> `NamedGetter()` returns `Option<NonNull<JSObject>>`, but I have no idea what DOM type (`#[dom_struct]`) corresponds to "empty list" here. I skimmed through the [`script::dom` docs](https://doc.servo.org/script/dom/index.html), but I feel I'm still missing some bits and pieces.
> 
> Can you help me out?

The document-tree child browsing context name property set isn't what you return to the caller! The spec's invoking the concept of the document-tree child browsing context name property set just to have a check for whether something is in it; if it's an empty list, then the thing you're checking for isn't in it. I think you might never have to actually construct the list itself at all, actually!

The actual named getter is this part: https://heycam.github.io/webidl/#dfn-determine-the-value-of-a-named-property

pylbrecht:
> The document-tree child browsing context name property set isn't what you return to the caller, so you aren't wrapping it into a JSObject. The spec's invoking the concept of the document-tree child browsing context name property set just to have a check for whether something is in it; if it's an empty list, then the thing you're checking for isn't in it. I think you might never have to actually construct the list itself at all, actually!
> 
> The actual named getter is this part: https://heycam.github.io/webidl/#dfn-determine-the-value-of-a-named-property and it's just calling up to that part for the one test; the return in that part just returns back to the determine-the-value steps, not the caller.

Oh, that makes sense. Thanks a lot for clarifying! Couldn't see the forest for the trees..

pylbrecht:
> Let childBrowsingContexts be all document-tree child browsing contexts of window's browsing context
I know how to get the window's browsing context (`document.browsing_context()`), but I am not sure how to get its children. Is there a method to get all children by specifying its parent? Or is there another way to traverse the document tree?

pshaughn:
> 
> 
> > Let childBrowsingContexts be all document-tree child browsing contexts of window's browsing context
> 
> I know how to get the window's browsing context (`document.browsing_context()`), but I am not sure how to get its children. Is there a method to get all children by specifying its parent? Or is there another way to traverse the document tree?
> 
> // edit:
> Am I on the right track with [`Constellation::all_descendant_browsing_contexts_iter()`](https://doc.servo.org/constellation/constellation/struct.Constellation.html#method.all_descendant_browsing_contexts_iter)?
> Further digging revealed `Window::send_to_constellation()`, but I couldn't find a `ScriptMsg` corresponding to `all_descendant_browsing_contexts_iter()`. Only `GetChildBrowsingContextId`, but that seems a bit tedious for getting _all_ children of a browsing context.

To be a "document-tree child browsing context", it has to be in the document tree, which means you already have a WindowProxy handy in the DOM. I think that's all you need to be able to figure out its name, and the name is really all you need here. I might me overlooking some corner case though.

pylbrecht:
Now I'm confused, but that's probably due to a lot of missing domain knowledge. So my apologies if the following statements/questions make absolutely no sense. I'm just trying to make sense of the spec in combination with your comments, @pshaughn.

What I understand by reading the [named access on the window object](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object) spec, is:
1. a window has a browsing context
2. browsing contexts may have multiple child browsing contexts

In order to get the names of all child browsing contexts, I somehow have to get all child browsing contexts in the first place, don't I?
How to actually get these child browsing contexts is what I'm currently missing.

> To be a "document-tree child browsing context", it has to be in the document tree, which means you already have a WindowProxy handy in the DOM.

I don't understand how that is helping to get the child browsing contexts of a window.
Each browsing context has a corresponding WindowProxy. WindowProxy has methods to get the top level browsing context or the parent, but that only helps traversing up the tree.

I get the sense of me missing fundamental knowledge about the relations between windows, browsing contexts, and the document tree. Maybe you can help me filling the blanks.

pshaughn:
> 
> 
> Now I'm confused, but that's probably due to a lot of missing domain knowledge. So my apologies if the following statements/questions make absolutely no sense. I'm just trying to make sense of the spec in combination with your comments, @pshaughn.
> 
> What I understand by reading the [named access on the window object](https://html.spec.whatwg.org/multipage/window-object.html#named-access-on-the-window-object) spec, is:
> 
>     1. a window has a browsing context
> 
>     2. browsing contexts may have multiple child browsing contexts
> 
> 
> In order to get the names of all child browsing contexts, I somehow have to get all child browsing contexts in the first place, don't I?
> How to actually get these child browsing contexts is what I'm currently missing.
> 
> > To be a "document-tree child browsing context", it has to be in the document tree, which means you already have a WindowProxy handy in the DOM.
> 
> I don't understand how that is helping to get the child browsing contexts of a window.
> Each browsing context has a corresponding WindowProxy. WindowProxy has methods to get the top level browsing context or the parent, but that only helps traversing up the tree.
> 
> I get the sense of me missing fundamental knowledge about the relations between windows, browsing contexts, and the document tree. Maybe you can help me fill in the blanks.

From what I recall, the only way you can get a child browsing context in servo is by having an iframe in the DOM. You don't need more distant descendant browsing contexts, just the immediate children of the current one, so you can just look at what iframes you have.

pylbrecht:
Thank you very much for elaborating! Now I have a good idea of how to approach this.
And it turned out much simpler than I imagined it to be in the beginning.

pshaughn:
> 
> 
> Thank you very much for elaborating! Now I have a good idea of how to approach this.
> And it turned out much simpler than I imagined it to be in the beginning.

The spec phrasing about this is particularly awful. I'm glad I could help.

pylbrecht:
Alright, so I feel I know what to do on an abstract level, but mapping it to code is quite a challenge for me.

So far I found two things that might be useful:
- `Document.iter_iframes()` to get all IFrames
- `HTMLIFrameElement.browsing_context_id`, which could be a gateway to the corresponding `BrowsingContext`

```rust
let child_browsing_context_ids = document.iter_iframes().map(|iframe| {
    iframe.browsing_context_id()
}).filter(|browsing_context_id| {
    browsing_context_id.is_some()
});
```

What I am missing:
- how to get a `BrowsingContext` by `BrowsingContextId`
- how to get the name of a `BrowsingContext` (and `HTMLIFrameElement` for that matter)

Are there any resources I can read up on regarding all these concepts (DOM, windows, IFrames, browsing contexts, etc.), which are easier to digest than the spec? I feel I'm missing the big picture to make sense of the details.

pshaughn:
The name of an HTMLIFrameElement is just its name content attribute, like most nameable elements. I believe the name of a browsing context, in servo, is the name field of its associated WindowProxy, so you can call https://github.com/servo/servo/blob/master/components/script/dom/htmliframeelement.rs#L555-L559 and then ask the returned WindowProxy if any. My memory might be faulty here.

jdm:
https://github.com/servo/servo/blob/a955ff280d252468e22082beabcc9ca35577f3d9/components/script/script_thread.rs#L527 is a map of BrowsingContextId to WindowProxy, which I believe answers your first question!

pylbrecht:
Thank you @pshaughn and @jdm for the hints and pointers!

pylbrecht:
I just want to leave a sign of activity here.

Our son is currently giving us a hard time during the nights, so I'm mostly recharging in my free time, which means I'm not working as actively on this as I'd like to.

In case this issue becomes urgent, please feel free to assign this to someone else, who can work on this more regularly. Otherwise I'll of course keep hacking on this!

jdm:
Not to worry! If this is still something that you'd like to keep hacking on, you're welcome to it!

pylbrecht:
I'm back with unsuccessfully trying to build my most recent changes:
```
error[E0271]: type mismatch resolving `<std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/do
m/window.rs:1447:21: 1447:70]> as std::iter::IntoIterator>::Item == dom::bindings::root::Root<dom::bindings::root::Dom<dom::htmliframeelement::HTMLIFrameEleme
nt>>`
    --> components/script/dom/window.rs:1467:41
     |
1467 |         let mut elements = iframes_iter.chain(name_iter).chain(id_iter);
     |                                         ^^^^^ expected reference, found struct `dom::bindings::root::Root`
     |
     = note: expected reference `&dom::bindings::root::Dom<dom::element::Element>`
                   found struct `dom::bindings::root::Root<dom::bindings::root::Dom<dom::htmliframeelement::HTMLIFrameElement>>`

error[E0599]: no method named `chain` found for struct `std::iter::Chain<std::iter::Filter<impl std::iter::Iterator, [closure@components/script/dom/window.rs:
1434:21: 1439:10 name:_]>, std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/dom/window.rs:1
447:21: 1447:70]>>` in the current scope
    --> components/script/dom/window.rs:1467:58
     |
1467 |           let mut elements = iframes_iter.chain(name_iter).chain(id_iter);
     |                                                            ^^^^^ method not found in `std::iter::Chain<std::iter::Filter<impl std::iter::Iterator, [clo
sure@components/script/dom/window.rs:1434:21: 1439:10 name:_]>, std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [clos
ure@components/script/dom/window.rs:1447:21: 1447:70]>>`
     |
     = note: the method `chain` exists but the following trait bounds were not satisfied:
             `<std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/dom/window.rs:1447:21: 1447
:70]> as std::iter::Iterator>::Item = dom::bindings::root::Root<dom::bindings::root::Dom<dom::htmliframeelement::HTMLIFrameElement>>`
             which is required by `std::iter::Chain<std::iter::Filter<impl std::iter::Iterator, [closure@components/script/dom/window.rs:1434:21: 1439:10 name
:_]>, std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/dom/window.rs:1447:21: 1447:70]>>: s
td::iter::Iterator`
             `std::iter::Chain<std::iter::Filter<impl std::iter::Iterator, [closure@components/script/dom/window.rs:1434:21: 1439:10 name:_]>, std::iter::Filt
er<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/dom/window.rs:1447:21: 1447:70]>>: std::iter::Iterator`
             which is required by `&mut std::iter::Chain<std::iter::Filter<impl std::iter::Iterator, [closure@components/script/dom/window.rs:1434:21: 1439:10
 name:_]>, std::iter::Filter<std::slice::Iter<'_, dom::bindings::root::Dom<dom::element::Element>>, [closure@components/script/dom/window.rs:1447:21: 1447:70]
>>: std::iter::Iterator`
```
I don't know how to turn a `dom::bindings::root::Root<dom::bindings::root::Dom<dom::htmliframeelement::HTMLIFrameElement>>` into a `&dom::bindings::root::Dom<dom::element::Element>` or if that is even the right approach for that matter.

Could you help me out?

pshaughn:
I have to admit that this is a part of Servo's architecture I've only been able to work with through trial and error and asking for help. Whenever I've hit this kind of type error, the solution usually ended up being some combination of & and * operators and a call to .upcast<Element>(), but I never came away with a complete understanding of why a particular arrangement of those worked in a particular situation.

jdm:
I'm pretty if you make iframes_iter:
```
        let iframes_iter = document
            .iter_iframes()
            .filter(|iframe| {
                if let Some(window) = iframe.GetContentWindow() {
                    return window.get_name() == name;
                }
                false
            })
            .map(|iframe| &**iframe.upcast::<Element>())
;
```
that it will fix that problem. The goal here is to convert the `Root<Dom<HTMLIFrameElement>>` to a `Root<Dom<Element>>`, then use the fact that dereferencing a `Root<T>` returns a `&T` to get us the required `&Dom<Element>` that the chain expects.

jdm:
I was wrong; it's a bit more subtle than that because dereferencing a `Root<Dom<T>>` returns `&T` instead. I'm tinkering with it locally.

jdm:
This makes the build succeed for me. I originally attempted to unify the three iterators to be over `&Element` instead of two of them being over `&Dom<Element>`, but that exposed that I couldn't do the upcast for the iframe iterator because of Rust's borrow rules. It's valid to make a new stable collection from that iterator and upcast those elements, however, which gives us our three iterators over `&Element` as original desired.

```diff
diff --git a/components/script/dom/window.rs b/components/script/dom/window.rs
index 7ef1b81dd2..bfaf604e4b 100644
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -1400,25 +1400,32 @@ impl WindowMethods for Window {
         let document = self.Document();

         // https://html.spec.whatwg.org/multipage/#document-tree-child-browsing-context-name-property-set
-        let iframes_iter = document
+        let iframes: Vec<_> = document
             .iter_iframes()
             .filter(|iframe| {
                 if let Some(window) = iframe.GetContentWindow() {
                     return window.get_name() == name;
                 }
                 false
-            });
+            })
+            .collect();
+
+        let iframes_iter = iframes
+            .iter()
+            .map(|iframe| iframe.upcast::<Element>());

-        let name = Atom::from(name);
+        let name = Atom::from(&*name);

         // Step 1.
         let elements_with_name = document.get_elements_with_name(&name);
         let name_iter = elements_with_name
             .iter()
+            .map(|element| &**element)
             .filter(|elem| is_named_element_with_name_attribute(elem));
         let elements_with_id = document.get_elements_with_id(&name);
         let id_iter = elements_with_id
             .iter()
+            .map(|element| &**element)
             .filter(|elem| is_named_element_with_id_attribute(elem));

**Comment by pylbrecht** [comment]
> I have to admit that this is a part of Servo's architecture I've only been able to work with through trial and error and asking for help.

Thanks, seeing someone else struggling with this is comforting.

@jdm, thanks for the elaborate explanation. It's not the first time I come
across casting of DOM elements and it was a mystery to me. Seeing `&**`s
now is starting to make sense.

**Comment by pylbrecht** [comment]
Since I have two weeks of vacation, I can finally spend some time on this again.
Although this branch became so stale that rebasing on latest master became the next blocker.

Is it safe to always choose the latest version of a package, if `Cargo.lock` shows conflicts?

**Comment by jdm** [comment]
I believe you can skip any Cargo.lock changes and use whatever is present on master. The only package change comes from bringing in the proxy branches of mozjs and rust-mozjs, and the only changes in those branches are https://github.com/servo/mozjs/commit/b234620eb463cda6236cce0eb1af1674ea4da234, which is already present in master mozjs.

**Comment by pylbrecht** [comment]
Great, that makes it a lot easier. Thank you!

**Comment by pylbrecht** [comment]
Oh wow, this is *a lot* of mismatched test expectations.

Can someone verify that all (or most of) these are related to the changes of this PR? And if so, please elaborate on how this is causing such a huge amount of tests to unexpectedly fail/pass.

I feel a bit overwhelmed by the mass of unexpected test results and I don't know how to handle this large amount of unexpected failures. It feels like the changes of this PR are implementing something fundamental, which a lot of other components rely on.

**Comment by pshaughn** [comment]
> 
> 
> Oh wow, these are _a lot_ of mismatched test expectations.
> 
> Can someone verify that all (or most of) these are related to the changes of this PR? And if so, please elaborate on how this is causing such a huge amount of tests to unexpectedly fail/pass.
> 
> I feel a bit overwhelmed by the mass of unexpected test results and I don't know how to handle this large amount of unexpected failures. It feels like the changes of this PR are implementing something fundamental, which a lot of other components rely on.
> 
> // edit
> Looks like 135 unexpected results.

This is probably right! Keep in mind that the window object is really the implicit Javascript global scope, not just the Javascript object named "window". Any time Javascript uses an identifier that isn't defined in the Javascript code, the named window getter checks if that identifier is an ID of a DOM object. A lot of WPT tests assume that this mechanism works and refer to DOM objects by ID directly instead of calling document.getElementById(), and those tests were seeing an "undefined" value instead of the object they were expecting to see. Usually that means the test fails, but occasionally it results in other unintended behavior, including spurious passes. If you look through the tests with changed results, the use of element IDs as Javascript identifiers should be pretty obvious in most of them.

**Comment by pylbrecht** [comment]
Thanks a lot for the elaborate response!

> Any time Javascript uses an identifier that isn't defined in the Javascript code, the named window getter checks if that identifier is an ID of a DOM object.

That really explains the insane amount of unexpected test results. I was wondering how the named window getter could affect so many things, but now I understand. I wasn't really aware what of an impact these changes will have, so I did not expect so many unexpected test results :eyes:.

I will first make all the tests in `html/browsers/the-window-object/named-access-on-the-window-object/` pass and then try the full wpt suite again. Fingers crossed that most of the unexpected failures/passes will be resolved by that.
If not, then I would start digging more into them one by one.

Makes sense?

**Comment by pshaughn** [comment]
> I will first make all the tests in `html/browsers/the-window-object/named-access-on-the-window-object/` pass and then try the full wpt suite again. Fingers crossed that most of the unexpected failures/passes will be resolved by that.
> If not, then I would start digging more into them one by one.
> 
> Makes sense?

Most of the unexpected failures/passes should logically still be there. These are mostly tests that have _never been able to run properly in Servo before_, now actually finding the DOM elements they're supposed to be testing for the first time. They'll need to have their test expectations updated, and maybe issues filed if they're now failing.

**Comment by pylbrecht** [comment]
Ah, of course, that makes sense. Thanks again for elaborating!

**Comment by pylbrecht** [comment]
While investigating test failures related to the named window getter, I got this output:
```
$ ./mach test-wpt tests/wpt/web-platform-tests/html/browsers/the-window-object/named-access-on-the-window-object --headless
Running 6 tests in web-platform-tests

  ▶ CRASH [expected OK] b'/html/browsers/the-window-object/named-access-on-the-window-object/cross-global-npo.html'
  │
  └ Assertion failure: gcMarker->tracingCompartment == comp, at /home/palbrecht/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/95387f7/mozjs/js/src/gc/Marking.cpp:243

  ▶ CRASH [expected OK] b'/html/browsers/the-window-object/named-access-on-the-window-object/named-objects.html'
  │
  └ Assertion failure: gcMarker->tracingCompartment == comp, at /home/palbrecht/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/95387f7/mozjs/js/src/gc/Marking.cpp:243

  ▶ CRASH [expected OK] b'/html/browsers/the-window-object/named-access-on-the-window-object/window-named-properties.html'

  ▶ CRASH [expected OK] b'/html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.html'
  │
  └ Assertion failure: gcMarker->tracingCompartment == comp, at /home/palbrecht/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/95387f7/mozjs/js/src/gc/Marking.cpp:243
```
Do you have any idea how I would end up failing this assertion?

To maybe give at least some context, the related cpp code looks like this:
```c++
template <typename T>
void js::CheckTracedThing(JSTracer* trc, T* thing) {
        ...
        // Check that we don't stray from the current compartment and zone without
        // using TraceCrossCompartmentEdge.
        Compartment* comp = thing->maybeCompartment();
        MOZ_ASSERT_IF(gcMarker->tracingCompartment && comp,
            gcMarker->tracingCompartment == comp);
        MOZ_ASSERT_IF(gcMarker->tracingZone,
            gcMarker->tracingZone == zone || zone->isAtomsZone());
        ...
}
```

jdm:
The most helpful information for that assertion would be the stack trace when it fails. This is deep in the JS engine GC machinery, so we'll need to figure out what invariants we are failing to maintain in our usage.

components/script/window_named_properties.rs
```diff
@@ -0,0 +1,205 @@
...
+    _proxy: HandleObject,
+    extensible: *mut bool,
+) -> bool {
+    *extensible = true;
+    true
+}
+
+#[allow(unsafe_code)]
+unsafe extern "C" fn class_name(_cx: *mut JSContext, _proxy: HandleObject) -> *const i8 {
+    &b"WindowProperties" as *const _ as *const i8
```
jdm, line 164: This string needs a `\0` final byte.

pylbrecht: Great catch! That changed the test results from `CRASH` to `TIMEOUT`. I'd call that progress.

Thanks for your effort looking through the changes!

components/script/window_named_properties.rs
```diff
@@ -0,0 +1,206 @@
...
+#[allow(unsafe_code)]
+unsafe impl Sync for SyncWrapper {}
+
+lazy_static! {
+    static ref HANDLER: SyncWrapper = {
+        let traps = ProxyTraps {
+            enter: None,
+            getOwnPropertyDescriptor: Some(get_own_property_descriptor),
+            defineProperty: Some(define_property),
+            ownPropertyKeys: None,
```
jdm, line 41: For reference, the corresponding function in Gecko is https://searchfox.org/mozilla-central/source/dom/base/WindowNamedPropertiesHandler.cpp#168.

components/script/window_named_properties.rs
```diff
@@ -113,6 +114,15 @@ unsafe extern "C" fn get_own_property_descriptor(
     return true;
 }
 
+#[allow(unsafe_code)]
+unsafe extern "C" fn own_property_keys(
+    _cx: *mut JSContext,
+    _proxy: HandleObject,
+    _props: MutableHandleIdVector,
+) -> bool {
+    false
```
pylbrecht, line 123: @jdm, not sure if that is the kind of dummy you had in mind.

jdm: Generally a false return value will be treated as an unrecoverable error by the JS engine,  so you probably want true instead.

jdm:
Poking around in lldb, I'm pretty sure the assertion failure is triggered by a window proxy brain transplant operation ([details](https://github.com/servo/servo/blob/f8105f868953be94fc1b5f069499ae9f54c3139a/components/script/dom/windowproxy.rs#L640-L646)), which looks like it could end up creating a cross-compartment proxy object and confusing the GC.

jdm:
I can trigger the same assertion failure with the following testcase:
```html
<iframe id="fm1"></iframe>
<script>document.getElementById("fm1").contentWindow</script>
```

yvt:
I got SEGV instead, which was caused by `ProxyTrap::getPrototypeIfOrdinary` being `None`. After implementing it (commit https://github.com/yvt/servo/commit/88139b8f16990be7afe20d53262da25ca1389629), all tests in `tests/wpt/web-platform-tests/html/browsers/the-window-object/named-access-on-the-window-object` but one are passing:

```
 0:09.34 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/navigated-named-objects.window.html
 0:09.34 WARNING Got command init_succeeded in state running
 0:09.38 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/named-objects.html
 0:09.38 WARNING Got command init_succeeded in state running
 0:09.43 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/window-null-names.html
 0:09.43 WARNING Got command init_succeeded in state running
 0:09.67 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/cross-global-npo.html
 0:09.67 WARNING Got command init_succeeded in state running
 0:09.71 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/window-named-properties.html
 0:09.71 WARNING Got command init_succeeded in state running
 0:13.74 TEST_END: Test OK. Subtests passed 6/6. Unexpected 0
 0:13.74 TEST_START: /html/browsers/the-window-object/named-access-on-the-window-object/prototype.html
 0:14.18 TEST_END: Test OK. Subtests passed 6/7. Unexpected 1
FAIL duplicate property names - assert_equals: expected 1 but got 0
@http://web-platform.test:8000/html/browsers/the-window-object/named-access-on-the-window-object/window-named-properties.html:64:16
Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2095:25
test@http://web-platform.test:8000/resources/testharness.js:566:30
@http://web-platform.test:8000/html/browsers/the-window-object/named-access-on-the-window-object/window-named-properties.html:61:5
```

The remaining failure is probably due to the stub implementation of `ownPropertyKeys`.

jdm:
Yeah, the assertions only exist in debug-mozjs builds.

yvt:
After some debugging, I think I found the cause of the assertion failure:  

 1. `WindowProxy::set_window` transplants `WindowProxy`'s `JSObject` from `old_js_proxy` to `new_js_proxy`. `new_js_proxy` is stored back to `WindowProxy::reflector`, and `old_js_proxy` (the CCW belonging to the old compartment) is simply thrown away.
 2. Each `Window` maintains a reference to the `WindowProxy` via `Dom<WindowProxy>`.
 3. The `<Dom<WindowProxy> as JSTraceable>::trace` method traces through `WindowProxy::reflector`, whose `JSObject` belongs to the new compartment regardless of the compartment of the `Window` being traced. When a no-longer-active `Window` is being traced, this might be different from the `Window`'s compartment, hence the assertion failure.

yvt:
Addressing the cross-compartment reference issue with `WindowProxy` (commit <https://github.com/yvt/servo/commit/8f7b0cff87f0eab921e13e6990d76e12935e8675>) does seem to fix the above assertion failures and make most of these tests pass in `debug-mozjs` builds.

yvt:
Opened #28534.

yvt:
> 💔 Test failed - [checks-github](https://github.com/servo/servo/runs/3151517184)

PR #28538 introduced new fields to `ProxyTraps`, which can be initialized with `None` to retain the old (default) behavior.

```diff
        let traps = ProxyTraps {
            enter: None,
            getOwnPropertyDescriptor: Some(get_own_property_descriptor),
            defineProperty: Some(define_property),
            ownPropertyKeys: Some(own_property_keys),
            delete_: Some(delete),
            enumerate: None,
            getPrototypeIfOrdinary: None,
+           getPrototype: None,
+           setPrototype: None,
+           setImmutablePrototype: None,
            preventExtensions: Some(prevent_extensions),
```

jtmorrisbytes:
Whats the status of this issue? Is anyone still working on this?

pylbrecht:
I am no longer working on this issue and am not planning to pick it up again.
I remember this being way over my head at that time. My apologies for not communicating that clearly.

pshaughn:
Finishing this is a lot tougher than it looks because the Window and WindowProxy are exotic objects that have to hook into Spidermonkey at a really deep level to do what they need to do across different Javascript realms, and the way the getter method connects to the Javascript prototype chain of that exotic object is itself exotic in a weird, special way. The getter method itself was never the hard part, and you should only try to take this on if you are very comfortable with Spidermonkey exotic objects.

PR #27994, opened 2020-12-28
Fix num_threads to avoid divide by zero error when running without a thread pool

teymour-aldridge:
Signed-off-by: teymour-aldridge <teymour.aldridge@icloud.com>

This updates the style thread pool to set `num_threads` to `None` rather than `0` when running without a thread pool.

---

I *think* this change might not need tests, but if it does I'm not sure exactly how to do so.


emilio (approved):
Looks great, thanks!

PR #28006, opened 2020-12-30
Fix `document.characterSet` not reflecting byte order marks.

### LINKED ISSUE #28005: `document.characterSet` doesn't take the BOM into account
See the encoding/bom-handling.html WPT test. The BOM is being taken into account when decoding, probably because `encoding_rs` handles that behind the scenes, but it doesn't have a way to communicate that to the caller.

This also used to be a bug in the specs until it was fixed with whatwg/html#5359 and whatwg/encoding#203 earlier this year.

andreubotella:
The process of decoding the network byte stream to Unicode is backed by an instance of `encoding_rs::Decoder`, which will switch the encoding it uses if it finds a BOM in the byte stream. However, this change in encoding is not communicated back to the caller and so `document.characterSet` gives the wrong result. This change fixes that.

See whatwg/html#5359 and whatwg/encoding#203 for the spec-level backing for this change.

---

- [X] These changes fix #28005 (GitHub issue number if applicable)


jdm:
Thanks for implementing this!
@bors-servo r+

components/script/dom/servoparser/mod.rs
```diff
@@ -463,6 +483,35 @@ impl ServoParser {
...
+                    //println!("Partial BOM: {:?}", partial_bom);
+                    false
+                }
+            } else {
+                //println!("partial_bom is None");
+                false
+            };
+            if set_bom_to_none {
+                *self.bom_sniff.borrow_mut() = None;
+            }
```
CYBAI, line 512: I just noticed there are some commented `println` which might be good to remove them.

Besides, as the `set_bom_to_none` only lives in this scope, maybe we can just set it in the `true` case? then, we don't need to return the boolean value. 👀

```suggestion
            let set_bom_to_none = if let Some(partial_bom) = self.bom_sniff.borrow_mut().as_mut() {
                if partial_bom.len() + chunk.len() >= 3 {
                    partial_bom.extend(chunk.iter().take(3 - partial_bom.len()).copied());

                    if let Some((encoding, _)) = Encoding::for_bom(&partial_bom) {
                        self.document.set_encoding(encoding);
                    }

                    *self.bom_sniff.borrow_mut() = None;
                } else {
                    partial_bom.extend(chunk.iter().copied());
                }
            } 
```

andreubotella: I didn't realize I hadn't cleaned up this part of the code. I'll file a PR to clean it up.

andreubotella: Now that I look into it, that would be borrowing `self.bom_sniff` inside the scope of the other borrow, which causes a panic. That's why I did that by setting a variable, so the borrow that replaced `self.bom_sniff` with `None` was outside the scope.

I'm not sure if there's a more idiomatic way to do something like this.

CYBAI: > I didn't realize I hadn't cleaned up this part of the code. I'll file a PR to fix it.

Thank you!

> That's why I did that by setting a variable, so the borrow that replaced `self.bom_sniff` with `None` was outside the scope.

Ah, I see. Then maybe let's keep it as is 🙇

PR #28093, opened 2021-01-24
Disallow duplicate taskcluster artifacts.

jdm:
Taskcluster has started complaining if the task's payload contains duplicate artifacts. These changes strip them out and add automated tests that detect them in the future.

### DIFF
--- a/etc/taskcluster/decision_task.py
+++ b/etc/taskcluster/decision_task.py
@@ -698,12 +698,13 @@ def wpt_chunks(platform, make_chunk_task, build_task, total_chunks, processes,
                     --tracker-api default \
                     --reporter-api default
             """)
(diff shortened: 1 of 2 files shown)


SimonSapin:
It doesn’t really matter here but using `list.append(x)` is preferred over `list += [x]` since it mutates in-place and keeps the memory allocation, instead of creating a new list assigning to the local variable then dropping the old one.

r+ either way

jdm:
My experience with `list.append((a, b))` is the result list had two elements appended to it. I couldn't figure out how to avoid that.

SimonSapin:
Are you sure that wasn’t `list.extend`?

jdm:
Ah, you're right.

PR #28129, opened 2021-02-01
Make unsafe

### LINKED ISSUE #16868: Clean up unsafe Document methods
All the methods taking `*mut JSRuntime`, such as `handle_mouse_event` and friends should be `unsafe`, because they take a raw pointer and do dangerous things with it.

teymour-aldridge:
---

- [ ] These changes fix #16868 (GitHub issue number if applicable) (maybe – I'm not sure)

### DIFF
--- a/components/script/animations.rs
+++ b/components/script/animations.rs
@@ -125,7 +125,7 @@ impl Animations {
     pub(crate) fn do_post_reflow_update(&self, window: &Window, now: f64) {
         let pipeline_id = window.pipeline_id();
         let mut sets = self.sets.sets.write();
-        self.root_newly_animating_dom_nodes(&sets, window);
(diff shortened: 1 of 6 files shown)


jdm (changes requested):
Thanks for doing this work! I think I see an improvement that's possible, however - if we modify the functions to accept a `&Runtime` argument instead, we can call `runtime.rt()` inside the methods and they can be called from safe code again.

teymour-aldridge:
I've just been looking at the code again, and it might be possible to actually remove the argument entirely? It seems to be passed only to `from_untrusted_node_address` which doesn't use it (as far as I can see).

jdm:
That's even better!

jdm:
@bors-servo r+
Thanks!

PR #28203, opened 2021-02-25
Update truetype dependency.

### LINKED ISSUE #28101: Migrate away from the `truetype` crate
In PR https://github.com/servo/servo/pull/27619 which updates rustc, [a test failure](https://community-tc.services.mozilla.com/tasks/DRvKYBVxTuarPa8glF1taQ/runs/0/logs/public/logs/live.log#L2293) shows that the `truetype` has [code that calls `std::mem::zeroed`](https://github.com/bodoni/truetype/blob/6da06be224c7ddb96a31db1e8c915ca87f3fc420/src/macros.rs#L64) for at least [one struct that contains a `Vec`](https://github.com/bodoni/truetype/blob/6da06be224c7ddb96a31db1e8c915ca87f3fc420/src/naming_table.rs#L25).

This is Undefined Behavior, and [newer Rust versions prevent it by panicking at runtime](https://github.com/rust-lang/rust/pull/66059).

Unfortunately this problem is still present in the latest git commit for that crate, which is from 2017. Fixing it is likely possible but not trivial. Since we only use this crate in `components/gfx/platform/windows/font.rs` it might be easier to migrate to another crate that doesn’t have this problem. There exist several that can parse Truetype files. `stb_truetype` is already in our `Cargo.lock` (though I don’t know if we actually build it in typical configs, it might be behind a feature flag of an intermediate dependency). `ttf-parser` is under active maintenance.

jdm:
Closes https://github.com/servo/servo/issues/28101

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6596,9 +6596,9 @@ checksum = "ce607aae8ab0ab3abf3a2723a9ab6f09bb8639ed83fdd888d857b8e556c868d8"
 
 [[package]]
 name = "truetype"
-version = "0.26.2"
+version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a4557422979ca99d0ffebef23d01185dde76153f09aa2840863ed60ad8e2fec"
+checksum = "063aa7cf1a7e56677e639c8f180a84d30b8fd788a684cfee2460418629925730"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


SimonSapin:
@jdm This still in "draft" state. Is something else missing?

jdm:
No, I just assumed the try run would expose a problem :)

PR #28217, opened 2021-02-26

emilio:


emilio:
@bors-servo try

 * Forgot there are two sets of WPT expectations to change now.

emilio:
@bors-servo retry

 * I think those two abspos timeouts are intermittent. I don't know how this could affect them really.

jdm:
The same test times out on all platforms in all configurations. Fascinating!

jdm:
r=me if you just want to mark it as a timeout and move on.

emilio:
I tried to debug it, but winit panics on my machine both with and without the x11 back-end, so I'll just do that for now yeah. Thanks!

emilio:
Oh, actually I can repro in headless mode, let me try to poke a bit...

emilio:
Fails also on master on my machine at least, that test takes a long long time on servo (not clear if for a good reason).

For now I'll annotate the timeout.

@bors-servo r=emilio,jdm

PR #28242, opened 2021-03-05
Readme get python2 packages

rutaihwa:
Make explicit when getting python2 packages, otherwise it will give errors.

---


jdm:
We should actually remove all python2 references from the readme. It is no longer required.

rutaihwa:
That is what I was thinking, because I could build just fine with python3.

PR #28248, opened 2021-03-07
Use swap() instead of load()/store() in get_and_reset_text_shaping_performance_counter()

### LINKED ISSUE #28245: components/gfx: possible atomicity violation in font.rs
https://github.com/servo/servo/blob/4a46a476d7c22fe22214a8c368d9d54f8a7a55d2/components/gfx/font.rs#L524-L528
For `TEXT_SHAPING_PERFORMANCE_COUNTER` ,
I wonder if the result is expected when `load()` and `store()` are interleaved by another `store()` from another thread. e.g.

|T1|T2|
|---|---|
|A=C.load()||
||A=C.load()|
||C.store(0)|
|C.store(0)|
|return A|return A|

Now we have two threads returning A instead of one returning A & the other returning 0.

delosrogers:
Replaced usage of `.load()` then `.store()` with `.swap()` to prevent possible atomicity issues.

---

- [X] These changes fix #28245

- [X] These changes do not require tests because: They do not change function of the code.

### DIFF
--- a/components/gfx/font.rs
+++ b/components/gfx/font.rs
@@ -522,8 +522,7 @@ impl RunMetrics {
 }
 
 pub fn get_and_reset_text_shaping_performance_counter() -> usize {
-    let value = TEXT_SHAPING_PERFORMANCE_COUNTER.load(Ordering::SeqCst);
-    TEXT_SHAPING_PERFORMANCE_COUNTER.store(0, Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


emilio:
This only gets called from the layout thread which is once-per-process so shouldn't really be an issue in practice. But looks good.

@bors-servo r+

PR #28269, opened 2021-03-12
update regex

jdm:


jdm:
Verified that https://github.com/rust-lang/regex/commit/e040c1b06397a254cccd3506ee80dbe042360afd is the commit that triggers the regression.

jdm:
I have not been able to reproduce the issue in mozjs or rust-mozjs, strangely (https://github.com/servo/rust-mozjs/pull/537 and https://github.com/servo/mozjs/pull/274)

BurntSushi:
This should be fixed by `regex 1.4.5`.

PR #28364, opened 2021-04-14
Use `CheckRegexSyntax` to avoid constructing a regular expression ahead of time for `<input>` elements

CYBAI:
The `JS::CheckRegExpSyntax` function is introduced into mozjs in https://bugzilla.mozilla.org/show_bug.cgi?id=1636495.

Ref: https://searchfox.org/mozilla-central/rev/3de2db87f3c9001ae478318d47a2ca3427574382/dom/base/nsContentUtils.cpp#6686-6698

So, we can use the same function to check regexp for the `pattern` attribute so that we can fix the corresponding wpt tests.

---

### DIFF
--- a/components/script/dom/htmlinputelement.rs
+++ b/components/script/dom/htmlinputelement.rs
@@ -18,7 +18,7 @@ use js::jsapi::{
     NewUCRegExpObject, ObjectIsDate, RegExpFlag_Unicode, RegExpFlags,
 };
 use js::jsval::UndefinedValue;
-use js::rust::jsapi_wrapped::{ExecuteRegExpNoStatics, ObjectIsRegExp};
+use js::rust::jsapi_wrapped::{CheckRegExpSyntax, ExecuteRegExpNoStatics, ObjectIsRegExp};
 use js::rust::{HandleObject, MutableHandleObject};
 use net_traits::blob_url_store::get_blob_origin;
 use net_traits::filemanager_thread::FileManagerThreadMsg;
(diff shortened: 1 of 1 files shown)


CYBAI:
hmm.... looks like the test is somehow not run 🤔 

image

CYBAI:
Ah, I see the problem. We don't support `:invalid` selector yet so the test will fail earlier from the `querySelectorAll`.

👇 small test case for the `:invalid` selector

```sh
→ cat test.html
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: test.html
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ <script>
   2   │   console.log(document.querySelectorAll(':invalid'));
   3   │ </script>
   4   │
   5   │
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

→ ./mach run test.html
[2021-04-14T11:31:57Z ERROR script::dom::bindings::error] Error at :0:0 SyntaxError: The string did not match the expected pattern.
```

CYBAI:
`:invalid` related issue is at #10781 and there's an open PR for it at  #26729

mrobinson (approved):
This seems like a low impact change. Before the pattern would fail during `check_js_regex_syntax`, but it's checked ahead of time using the new function.

PR #28454, opened 2021-05-30
add etc/shell.nix for nix/nixOS and instructions - fixes #10468

mkg20001:
This adds instructions for building servo on nixOS and the required shell.nix to prepare the environment

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [x] These changes fix #10468 (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because they relate to adding a new build environment, not any servo code

PR #28463, opened 2021-05-31
add NixOS support to mach and automatically re-launch mach in nix-shell

mkg20001:
This is a followup for #28454 which adds nixos support directly to mach, saving the user from typing `nix-shell` everytime

---

- [x] These changes do not require tests because they just extend mach

### DIFF
--- a/mach
+++ b/mach
@@ -110,6 +110,7 @@ if __name__ == '__main__':
         import subprocess
         from shlex import quote
         mach_dir = os.path.abspath(os.path.dirname(__file__))
+        print('NOTE: Entering nix-shell etc/shell.nix')
         try:
             # sys argv already contains the ./mach part, so we just need to pass it as-is
             subprocess.Popen(['nix-shell', mach_dir + '/etc/shell.nix', '--run', ' '.join(map(quote, sys.argv))]).wait()
--- a/mach
+++ b/mach
@@ -106,4 +106,14 @@ if __name__ == '__main__':
         orig_command_line = forking.get_command_line
         forking.get_command_line = my_get_command_line
 
-    main(sys.argv)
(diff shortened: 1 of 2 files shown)


mkg20001:
Not sure if this is the best way to do things

It makes `mach` work out of the box on nixOS systems

yvt:
I have a mixed opinion on this for the following reasons:
 - This only changes the behavior on NixOS and makes it inconsistent between platforms.
 - This makes it less obvious to the users that external dependencies are needed to compile and run Servo, making them wonder why the compiled executable won't run outside `mach`. Note that shared libraries may be GC-ed when `nix-shell` is not running, rendering the compiled executable unusable until the user realizes they need to re-link the program. I often work inside `nix-shell` and leave it open for days so that the linked shared libraries stay in place.

It'd be helpful to direct the user to `etc/shell.nix` when `mach bootstrap` is executed, though.

est31:
I think the idea to detect nix os is good. rustc has nixos detection built in as well. It does print it in the build log however, so that you know this is happening. Maybe this PR should print something as well?

mkg20001:
Would this fix the concerns?:
- add message when mach is entering nix-shell
- add message to bootstrap that says that mach will automatically use etc/shell.nix, but user can also use it directly and needs to use it when trying to run the binary

yvt:
> * add message when mach is entering nix-shell

Also check that `nix-shell` is not entered already, e.g., by detecting the existence of environment variables that are set by `shell.nix`.

mkg20001:
@yvt This is already added in this line https://github.com/servo/servo/pull/28463/files#diff-e48b94354bc9ed31688c515289d6282a76f6d93dddf9fd699f6f1d9852ca8da5R109

mkg20001:
Is this good to merge? I added the messages.

jdm:
Knowing nothing about nixOS, it sounds like the discussion has resolved so I'll merge this.

PR #28420, opened 2021-05-22
Update toolchain to nightly-2021-05-18

dominiccooney:
The nightly-2021-03-12 toolchain doesn't have RLS on Windows. This
breaks code completion in Visual Studio Code because the rust plugin
picks up the project toolchain.

RLS is available in all tier one platforms in nightly-2021-05-18 per:
https://rust-lang.github.io/rustup-components-history/

Signed-off-by: Dominic Cooney <dominic.cooney@gmail.com>

Update toolchain to nightly-2021-05-18

---

- [X ] `./mach build -d` does not report any errors
- [ X] `./mach test-tidy` does not report any errors (note, except tabs/spaces issue in an unrelated file... happy to fix in this PR)
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [X] These changes do not require tests because changing the toolchain, covered by existing tests

PR #28511, opened 2021-06-16
Update pathfinder_simd

sagudev:
Fix blocker for https://github.com/servo/servo/pull/28420

I only ran `cargo update -p pathfinder_simd`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2782,7 +2782,7 @@ dependencies = [
  "itoa",
  "log",
  "net2",
- "rustc_version",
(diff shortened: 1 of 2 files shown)


sagudev:
Unit testing is not run on linux on push?

jdm:
I suspect if you run `./mach test-tidy` you'll get errors about new duplicate package versions. Can you add the new duplicates to the list in https://github.com/servo/servo/blob/98cffa3d16baceee6feca47e5ffc431bf837369a/servo-tidy.toml#L31?

PR #28527, opened 2021-06-24
Add some additional `Debug` implemenations.

teymour-aldridge:
---

- [ ] These changes do not require tests because they only add `Debug` implemenations


jdm:
I suppose there's no particular harm to these changes, but a bunch of these types I can't imagine ever being debug printed (which reinforces why there's no annotation present).

PR #28534, opened 2021-07-02
Introduce `Untransplantable` trait and implement it on `WindowProxy`

yvt:
This PR fixes some cases of compartment mismatch that have been reported in <https://github.com/servo/servo/pull/27952#issuecomment-817338319>.

SpiderMonkey requires a `WindowProxy` object to be in the same compartment as its `Window`, so when a `WindowProxy` changes `Window`, it must be transplanted to the new `Window`'s realm. When this happens, `WindowProxy::reflector` is replaced with a new `JSObject` that belongs to the new realm, and the old `reflector` is overwritten with a CCW. (So if there's an existing reference to `reflector` somewhere else, it will now refer to the CCW.) `*Dom<T>` types always trace through `T::reflector ()`.

`WindowProxy` is referenced by `Window` through `MutNullableDom<WindowProxy>`. SpiderMonkey prohibit cross-compartment references from being formed directly. However, when the transplantation happens, `MutNullableDom<WindowProxy>` turns into a cross-compartment reference as `WindowProxy::reflector()` now returns the transplanted `JSObject`.

(This violation can be detected at runtime if Servo is built with `--debug-mozjs`. It can be observed in test cases such as those in `tests/wpt/web-platform-tests/html/browsers/the-window-object/named-access-on-the-window-object`. The minimum reproducing example I found is rather small: `<iframe src="about:blank"></iframe>`.)

This PR attempts to alleviate this problem by introducing a new trait named `Untransplantable` to express transplantability at the type level. Existing similar reference wrapper types, including but not limited to `Dom`, now require `Untransplantable` on the wrapped type. For transplantable types, this commit introduces a new family of reference wrapper types: `MutNullableTransplantableDom` and `TransplantableDom- OnceCell`. These types are essentially `(*const T, Heap<*mut JSObject>)` and always trace through `Heap<*mut JSObject>` (a reference to `T`'s reflector or CCW). Its API is similar to the corresponding non-`TransplantableDom` types except that the caller must supply a containing global scope, which is used to wrap `T::reflector()` properly before storing it to `Heap<*mut JSObject>`. Even if the target object is transplanted, `Heap<*mut JSObject>` keeps tracking the wrapper object in the original realm, maintaining the compartment invariants.

---

---
- [x] These changes do not require tests because they are already covered by existing tests, albeit this requires Servo to be built with `--debug-mozjs` for the errors to be (reliably) observable

### DIFF
--- a/components/domobject_derive/lib.rs
+++ b/components/domobject_derive/lib.rs
@@ -12,7 +12,7 @@ extern crate syn;
 use proc_macro2;
 use quote::TokenStreamExt;
 
-#[proc_macro_derive(DomObject)]
+#[proc_macro_derive(DomObject, attributes(dom_object))]
 pub fn expand_token_stream(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let input = syn::parse(input).unwrap();
     expand_dom_object(input).into()
@@ -28,6 +28,16 @@ fn expand_dom_object(input: syn::DeriveInput) -> proc_macro2::TokenStream {
     let (first_field, fields) = fields
         .split_first()
         .expect("#[derive(DomObject)] should not be applied on empty structs");
+
+    // Take additional parameters from `#[dom_object]` helper attributes (if any)
+    let mut args = Args::default();
+    for attr in input.attrs.iter() {
+        if attr.path.is_ident("dom_object") {
+            if let Err(e) = attr.parse_meta().and_then(|meta| args.parse_meta(&meta)) {
+                return e.to_compile_error().into();
+            }
+        }
+    }
 
     let first_field_name = first_field.ident.as_ref().unwrap();
     let mut field_types = vec![];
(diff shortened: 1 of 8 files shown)


yvt:
Thinking for a bit, I'm starting to feel like it might not be the best idea to have a unique compartment for each realm (which we currently do, if I understand correctly). Since JS code can move objects between similar-origin realms freely, cross-realm (cross-compartment in the current design) DOM references can appear almost everywhere, so we would have to replace practially every DOM reference with something equivalent to `*TransplantableDom*` to uphold the compartment invariants. I wonder if this compartment usage is intentional or an unintentional artifact of not specifying `JS::CompartmentSpecifier::ExistingCompartment` when doing `JS_NewGlobalObject`.

jdm:
Oh, very good point. I hadn't noticed that the realm selection had been added to realm creation options; we will likely want to copy https://searchfox.org/mozilla-central/rev/a2b327f9696ab94b80d9ebc7d233abfc04b36d7a/dom/base/nsGlobalWindowOuter.cpp#1956-1996 and parts of https://searchfox.org/mozilla-central/rev/a2b327f9696ab94b80d9ebc7d233abfc04b36d7a/dom/base/nsGlobalWindowOuter.cpp#1998-2056 for our window global creation.

jdm:
https://searchfox.org/mozilla-central/rev/a2b327f9696ab94b80d9ebc7d233abfc04b36d7a/dom/base/nsGlobalWindowOuter.cpp#1708-1714 is another interesting snippet.

jdm:
Similarly, https://searchfox.org/mozilla-central/rev/a2b327f9696ab94b80d9ebc7d233abfc04b36d7a/dom/base/nsGlobalWindowOuter.cpp#2176-2187

yvt:
Closing this in favor of #28536

PR #28536, opened 2021-07-07
Create only one compartment for each script thread (agent)

yvt:
Documents in the same [agent][1] can share and exchange JS and DOM objects freely, so putting them in separate compartments would require almost every instance of `Dom` to be capable of handling cross-compartment references.

[1]: https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-agent-formalism

---

---
- [x] These changes do not require tests because I think there's already a wide test coverage for same-origin-domain JS object passing, albeit this requires Servo to be built with `--debug-mozjs` for the errors to be (reliably) observable

### DIFF
--- a/components/script/dom/bindings/interface.rs
+++ b/components/script/dom/bindings/interface.rs
@@ -16,6 +16,10 @@ use js::glue::UncheckedUnwrapObject;
 use js::jsapi::GetWellKnownSymbol;
 use js::jsapi::HandleObject as RawHandleObject;
 use js::jsapi::{jsid, JSClass, JSClassOps};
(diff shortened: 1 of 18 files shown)


jdm:
Thinking through whether this logic makes sense: unlike gecko, which has a single JS runtime in a process for all origins, servo has one JS runtime per effective top-level domain. This means that all the compartments in a runtime should be shareable, so we don't need any checks for same-origin when choosing a compartment to share.

@bors-servo r+

jdm:
Nice, there are a bunch of new passing tests!

PR #28088, opened 2021-01-22
build: flatpak support

okias:
Development build takes about 30 minutes on i7-7500U CPU @ 2.70GHz / 3.50GHz (with turbo). At least 8 GB RAM required.

Current state:
- [x] python stuff is installed
- [x] cargo deps needs to be packages (but sadly script needs to be [patched](https://github.com/flatpak/flatpak-builder-tools/issues/181)).
- [x] fix few CMake deps to get pre-downloaded and used https://github.com/servo/servo/issues/28102 .
- [x] open Servo window. Works when based on https://github.com/servo/servo/pull/28236
- [x] wait for https://github.com/servo/servo/pull/28236 merge
- [x] fix error[E0532]: expected unit struct, unit variant or constant, found tuple variant `visit::FnKind::Closure`
- [x] create update script, to keep flatpak in sync with servo rust requirement, cargo dependencies and misc.

```
flatpak-builder --repo=repo org.servo.Servo data/org.servo.Servo.json --force-clean
flatpak build-export repo org.servo.Servo
flatpak build-bundle repo org.servo.Servo.flatpak org.servo.Servo
flatpak install --user org.servo.Servo.flatpak
```

Maintanance
Flatpak cargo dependencies update process (assuming it's in master branch):
```
git pull --rebase
./data/flatpak_updater.sh
```

Related to: https://github.com/servo/servo/issues/12102
Based on: https://github.com/servo/servo/issues/27942

---

- [x] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [x] These changes fix #12102 (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because does not change the Servo code, just add different way to build it

Issue #27942, opened 2020-12-12
Got a mostly working flatpak


BryanQuigley:
Some questions:
 * For whatever reason Ctrl-L doesn't work in the flatpak.  What's different about it?
 * Would you like a PR to commit the desktop file and metainfo.xml to this repo instead of having it live on flathubs ?
 * My current strategy is to build the flatpak from the the nightly builds- another option would be to build from the source code directly. Thoughts?

Previous Issue #12102

### Current code
Not ready for merging anywhere, but can already build a flatpak :).

org.servo.Servo.desktop:

```
[Desktop Entry]
Type=Application
Name=Servo
Comment=Rust based browser engine
Exec=servo
Icon=org.servo.Servo
Terminal=false
Categories=Network;WebBrowser;
```

org.servo.Servo.metainfo.xml
```
<?xml version="1.0" encoding="UTF-8"?>
<component type="desktop-application">
  <id>org.servo.Servo</id>
  <url type="homepage">https://servo.org/</url>
  <url type="bugtracker">https://github.com/servo/servo/issues</url>
  
  <name>Servo</name>
  <summary>Servo is a browser engine written in Rust.</summary>
  
  <metadata_license>CC-BY-SA-4.0</metadata_license>
  <project_license>MPL-2.0</project_license>
  
  <recommends>
    <control>pointing</control>
    <control>keyboard</control>
    <control>gamepad</control>
  </recommends>
  
  <description>
    <p>
      Servo’s mission is to provide an independent, modular, embeddable web engine, which allows developers to deliver content and applications using web standards.
    </p>
    <p>
    Servo is written in Rust, and shares code with Mozilla Firefox and the wider Rust ecosystem. Since its creation in 2012, Servo has contributed to W3C/WHATWG web standards by reporting specification issues and submitting new cross-browser automated tests, and core team members have co-edited new standards that have been adopted by other browsers. As a result, the Servo project helps drive the entire web platform forward while building on a platform of reusable, modular technologies that implement web standards.
    </p>
  </description>
  
   <releases>
      <release version="0.0.1-638fa8b1e" date="2020-12-12" urgency="high">
        <url>https://download.servo.org/</url>
      </release>
  </releases>

  <content_rating type="oars-1.1" />

  <launchable type="desktop-id">org.servo.Servo.desktop</launchable>
  <screenshots>
    <screenshot type="default">
      <caption>Servo viewing servo.org</caption>
      <image>https://servo.org/img/screenshot-servo.png</image>
[...]

PR #28552, opened 2021-07-20
build: flatpak support

### LINKED ISSUE #12102: Support building flatpak packages
This will make cross-distro packaging somewhat less of a headache.

I'd like to claim this issue -- just filing it here for tracking.

okias:
Development build takes about 30 minutes on i7-7500U CPU @ 2.70GHz / 3.50GHz (with turbo). At least 8 GB RAM required.

```
flatpak-builder --repo=repo org.servo.Servo support/flatpak/org.servo.Servo.json --force-clean
flatpak build-export repo org.servo.Servo
flatpak build-bundle repo org.servo.Servo.flatpak org.servo.Servo
flatpak install --user org.servo.Servo.flatpak
```

Maintanance
Flatpak cargo dependencies update process (assuming it's in master branch):
```
git pull --rebase
./support/flatpak/flatpak_updater.sh
```

Related to: https://github.com/servo/servo/issues/12102
Based on: https://github.com/servo/servo/issues/27942

---

- [x] These changes fix #12102 (GitHub issue number if applicable)

- [x] These changes do not require tests because does not change the Servo code, just add different way to build it

I managed to broke git branch, so this is successor of https://github.com/servo/servo/pull/28088

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -121,6 +121,40 @@ jobs:
           name: release-binary
           path: target.tar.gz
 
+  build-flatpak:
+    name: Build (flatpak)
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Bootstrap
+        run: |
+          python3 -m pip install --upgrade pip virtualenv
+          sudo apt update
+          sudo apt install -qy flatpak flatpak-builder
+          sudo flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
+          sudo flatpak install -y org.freedesktop.Sdk/x86_64/20.08 org.freedesktop.Platform/x86_64/20.08
+      - name: Release build
+        run: flatpak-builder --repo=repo org.servo.Servo data/org.servo.Servo.json --force-clean
+      - name: Export build
+        run: flatpak build-export repo org.servo.Servo
+      - name: Bundle build
+        run: flatpak build-bundle repo org.servo.Servo.flatpak org.servo.Servo
+      - name: Archive binary
+        uses: actions/upload-artifact@v2
+        with:
+          name: release-binary
+          path: org.servo.Servo.flatpak
+      - uses: "marvinpinto/action-automatic-releases@latest"
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: flatpak
+          prerelease: true
+          title: "Flatpak Build"
+          files: |
+            org.servo.Servo.flatpak
+
   wpt-2020:
     if: ${{ github.ref_name == 'try-wpt-2020' || inputs.layout == '2020' || inputs.layout == 'all' }}
     name: Linux WPT Tests 2020
--- a/data/org.servo.Servo.desktop
+++ b/data/org.servo.Servo.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Servo
+Comment=Rust based browser engine
+Exec=servo
+Icon=org.servo.Servo
+Terminal=false
+Categories=Network;WebBrowser;
--- a/data/org.servo.Servo.metainfo.xml
+++ b/data/org.servo.Servo.metainfo.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.servo.Servo</id>
+  <url type="homepage">https://servo.org</url>
+  <url type="bugtracker">https://github.com/servo/servo/issues</url>
+
+  <name>Servo</name>
+  <summary>Servo is a browser engine written in Rust.</summary>
+
+  <metadata_license>CC-BY-SA-4.0</metadata_license>
+  <project_license>MPL-2.0</project_license>
+
+  <recommends>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>gamepad</control>
+  </recommends>
+
+  <description>
+    <p>
+      Servo’s mission is to provide an independent, modular, embeddable web engine, which allows developers to deliver content and applications using web standards.
+    </p>
+    <p>
+    Servo is written in Rust, and shares code with Mozilla Firefox and the wider Rust ecosystem. Since its creation in 2012, Servo has contributed to W3C/WHATWG web standards by reporting specification issues and submitting new cross-browser automated tests, and core team members have co-edited new standards that have been adopted by other browsers. As a result, the Servo project helps drive the entire web platform forward while building on a platform of reusable, modular technologies that implement web standards.
+    </p>
+  </description>
+
+  <releases>
+    <release version="0.0.0" date="2021-01-26">
+    </release>
+  </releases>
+
+  <content_rating type="oars-1.1" />
+
+  <launchable type="desktop-id">org.servo.Servo.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Servo viewing servo.org</caption>
+      <image>https://servo.org/img/screenshot-servo.png</image>
+    </screenshot>
+    <screenshot>
+      <caption>Servo displaying Wikipedia</caption>
+      <image>https://servo.org/img/screenshot-wikipedia.png</image>
+    </screenshot>
+  </screenshots>
+</component>
--- a/support/flatpak/flatpak_updater.sh
+++ b/support/flatpak/flatpak_updater.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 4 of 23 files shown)


romulasry:
This would be nice to have.

.github/workflows/linux.yml
```diff
@@ -113,6 +113,40 @@ jobs:
           name: release-binary
           path: target.tar.gz
 
+  build-flatpak:
```
mrobinson, line 116: A entire extra build on the CI for every change seems too much of a price to pay for this feature. Probably we need a bigger conversation about when / how this should be built and how it should be distributed. Does it replace the existing Linux package for instance? Right now, Servo it not a usable application....it's more like a demonstration of the current state of the engine. I'm curious what the target audience of the flatpak package is?

data/cargo-sources.json
```diff
@@ -0,0 +1,7743 @@
+[
```
mrobinson, line 1: This is a really big file of hashes! Can this be auto-generated during flatpak build? Otherwise, it seems like this needs to be updated every time we bump our dependencies? In general adding a new build is also a request or a promise to maintain it. For something non-trivial, explanatory documentation about how to update support files is really useful.

data/flatpak_updater.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
mrobinson, line 1: I think that `support/flatpak` would be a more natural location for these files.

When is data/flatpak_updater.sh supposed to be run. It seems unused in this version of the change?

mrobinson (changes requested):
This seems mostly fine to me, but I think we need a bigger conversation about how this fits into the distribution of Servo. Servo is a bit unique in that it ships its "minibrowser" for people who are curious. It isn't really meant as a full-fledged application to install on your system (other than to maybe for trying to bisect failures).

I suppose the other big things missing from this change is documentation about the purpose of the different build support files and how to generate them when dependencies change. Ideally, a flatpak build would have very low maintenance burden (hopefully even generating most of the files that it needs apart from scripts).

I'm happy to chat about these bigger questions here or find me in Zulip.

.github/workflows/linux.yml
okias: For your first point, yes. I think it's too high price, and probably even nightly build is questionable to have.

I did it to be sure it's working outside my machine mostly.

I'm well aware it's more about the demonstration ( // I think I have Servo screenshot running on Linux OnePlus 6 in Flatpak somewhere from 2021).

The target audience was meant to be developers considering their web-app to be packaged with Servo, which I think could be use of it (offline or online).

data/cargo-sources.json
okias: Exactly. Everything needs to be available offline and downloaded before the build begins (like the regular distros work). `data/flatpak_updater.sh` is here to regenerate it every time.

I would run `flatpak_updater.sh` on every push to keep the flatpak files synchronized.

data/flatpak_updater.sh
okias: Make sense, moved.

.github/workflows/linux.yml
mrobinson: This should likely move to the nightly build workflow as we don't really want to be building servo twice for every run of the main workflow.

.github/workflows/linux.yml
```diff
@@ -121,6 +121,40 @@ jobs:
...
+      - name: Export build
+        run: flatpak build-export repo org.servo.Servo
+      - name: Bundle build
+        run: flatpak build-bundle repo org.servo.Servo.flatpak org.servo.Servo
+      - name: Archive binary
+        uses: actions/upload-artifact@v2
+        with:
+          name: release-binary
+          path: org.servo.Servo.flatpak
+      - uses: "marvinpinto/action-automatic-releases@latest"
```
mrobinson, line 149: Instead of making a release here (which will create a release for every commit, if I'm not wrong), this should simply use the nightly workflow.

data/org.servo.Servo.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
```
mrobinson, line 1: This should also move to `support/flatpak`.

data/org.servo.Servo.metainfo.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mrobinson, line 1: Same with this file.

support/flatpak/fontsan-deps-sources.json
```diff
@@ -0,0 +1,20 @@
+[
```
mrobinson, line 1: Instead of checking these JSON files in, flatpak-updater.sh should be integrated into the flatpak build and generate these files into the `target` directory during build.

support/flatpak/flatpak_updater.sh
```diff
@@ -0,0 +1,99 @@
...
+      -o ../../support/flatpak/python3-pluggy
+    python3 ./flatpak-pip-generator tomli \
+      -o ../../support/flatpak/python3-tomli
+    python3 ./flatpak-pip-generator trove_classifiers \
+      -o ../../support/flatpak/python3-trove_classifiers
+    python3 ./flatpak-pip-generator packaging \
+      -o ../../support/flatpak/python3-packaging
+    python3 ./flatpak-pip-generator hatch-vcs \
+      -o ../../support/flatpak/python3-hatch-vcs
+  }
```
mrobinson, line 51: It would be great to document the mapping all of these dependencies to our Python dependencies. This seems to go well beyond the Python dependencies specified in `python/requirements.txt`. This documentation is necessary so we know which of these to remove when we update our Python requirements, otherwise this will quickly get out of sync and bitrot.

support/flatpak/org.servo.Servo.template.json
```diff
@@ -0,0 +1,118 @@
...
+      "build-options" : {
+        "env": {
+          "CARGO_HOME": "/run/build/servo/cargo",
+          "CARGO_NET_OFFLINE": "true"
+        }
+      },
+      "build-commands": [
+        "mv servobuild.example .servobuild",
+        "sed -i 's|use-rustup = true|use-rustup = false|' .servobuild",
+	"sed -i -e '/need_pip_upgrade = True/d;/req_path\]/d' python/mach_bootstrap.py",
```
mrobinson, line 98: There are some small formatting errors in this file, so it would be nice to run it through a formatter.

mrobinson (changes requested):

data/org.servo.Servo.desktop
okias: These are generic Linux files with descriptions not tied to the Flatpak ecosystem (regular distributions use them). I would understand that they're not particularly useful for Servo at this moment, but in the future, they'll be needed independently from Flatpak.

I'll have a bit delayed reaction; sorry about that. Thank you for your review.

data/python3-requirements.json
```diff
@@ -0,0 +1,230 @@
...
+            "sources": [
+                {
+                    "type": "file",
+                    "url": "https://files.pythonhosted.org/packages/f0/d0/21c6449df0ca9da74859edc40208b3a57df9aca7323118c913e58d442030/colorama-0.3.7.tar.gz",
+                    "sha256": "e043c8d32527607223652021ff648fbb394d5e19cba9f1a698670b338c9d782b"
+                }
+            ]
+        },
+        {
+            "name": "python3-certifi",
```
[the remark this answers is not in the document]
orowith2os: As I mentioned further up, enabling network access is fine, and would remove the need for these.

mrobinson:
I think this could make sense one day, but `servoshell` is really a demonstration of Servo rather than a web browser, so it likely doesn't make sense to maintain a user-focused distribution mechanism for it.

yookoala:
For those who come here for a Servo nightly demo, you may find something useful here:
https://github.com/fabiand/flatpak-servo

yookoala:
> For those who come here for a Servo nightly demo, you may find something useful here: https://github.com/fabiand/flatpak-servo

Sorry. My mistake. That doesn't work either.

yookoala:
With the latest v0.0.2 release, all you need is to build and run with this flatpak manifest:

```yml
id: org.servo.Servo
runtime: org.freedesktop.Platform
runtime-version: '25.08'
sdk: org.freedesktop.Sdk
command: /app/bin/servo
finish-args:
  - --socket=wayland
  - --socket=x11
  - --share=ipc
  - --share=network
  - --device=dri
  - --socket=pulseaudio
modules:
  - name: servo
    buildsystem: simple
    build-commands:
      - install -Dm755 servo /app/bin/servo
      - cp -r resources /app/bin/
    sources:
      - type: archive
        url: https://github.com/servo/servo/releases/download/v0.0.2/servo-x86_64-linux-gnu.tar.gz
        sha256: f10a9c5e71f16511a4722c254aa2c7a5d1b920a4c20583d22a8b68070512e2c0
```

Check [Building your first Flatpak](https://docs.flatpak.org/en/latest/first-build.html) for more information.

yookoala:
Made a repo to build the manifest for building / installing flatpak:
https://github.com/yookoala/servo-flatpak/

If you're looking for solution to build and install Servo's demo browser as flatpak, please try this out.

Issue #9482, opened 2016-02-01
Focusing the native window should trigger a focus event


paulrouget:
Some notes:

In `ports/glutin/window.rs`, along events like `Event::Closed`, we should handle the `Event::Focused(focused)` event and forward it to script_thread / document.

Once there, I'm not sure exactly what should be done. There is already a focus-handling mechanism at the document level (`*_focus_transaction`), but events are not dispatched as per https://github.com/servo/servo/blob/6b1a08c051f3be4b7bc7e6200053dfa530ecd106/components/script/dom/document.rs#L584

At the compositor level, `focus_parent_pipeline` ensures that the chain of iframes are properly focused. But I don't see where iframes are unfocused.

On blur then focus, we might want to save `self.focus_pipeline_id`, unfocus all the iframes, and when focus comes back, focus the previous saved pipeline.

I'm also not sure how to deal with the script thread requesting focus while the native window is not focused.

PR #28571, opened 2021-08-14
Improve inter-document focus handling

### LINKED ISSUE #8981: window.focus/blur support
In meta-issue #8973 to get devtools.html loading, we noticed `window.focus` is not supported yet.

Should not block initial load of the tools, but something we'd eventually want for full functionality.

yvt:
This PR improves multiple aspects of Servo's focus handling.

### Implement [`Window#focus`][2], [`Window#blur`][3]

    feat(script): add `Window#{focus, blur}`
    feat: implement `Window#focus`

The behavior of `Window#focus` is implemented in `WindowProxy::focus`. This utilizes a new message type `ScriptMsg::FocusRemoteDocument` to enable focusing a document that might not reside in the same script thread.

`Window#blur` does nothing as per the specification's recommendation.

> User agents are encouraged to ignore calls to this `blur()` method entirely.

Some test cases, such as `/html/browsers/origin/cross-origin-objects/cross-origin-objects.html`, are known to depend on these methods' mere existence.

### Implement the `focus` and `blur` events for `Window`

    feat(script): support firing `focus` and `blur` events at global objects

These events are fired by [the focus update steps][4] when documents get or lose focus.

### A document shall have focus iff its container does

    feat: notify the child document when focusing a container

Previously, this only worked one way - when a document requests focus, its container and outer containers are also given focus, but not the other way. This PR makes this work both ways as required by [the focusing steps][1] from the specification. To this end, two message types are changed or added:

 - The existing `ScriptMsg::Focus` message type is extended to contain the focused container's browsing context.
 - The new `ConstellationControlMsg::FocusDocument` message type is used by the constellation to request a pipeline to focus its document.

### Support unfocusing documents

    feat(constellation): construct a focus chain explicitly before broadcasting focus messages
    fix(constellation): move focus when closing a focused browsing context
    feat: notify a document when it loses focus

When a document is removed from the top-level browsing context's focus chain by having its container or one of its outer containers robbed of its status as a document's focused area, `blur` events should be fired at the document and its focused element (if any), and the element should lose a "focused" state.

To implement this behavior, a new message type `ConstellationControlMsg::Unfocus` is added. A pipeline receiving this type of message clears its `has_focus` flag, clears its focused element (*not* spec-compliant if I understand correctly; this is for compatibility with other browsers), and fires `blur` events as necessary.

### Track per-document focus state

    feat(script): track each document's focus state, implement `Document#hasFocus` fully

Each pipeline now maintains a local flag to track whether it's included in the top-level browsing context's focus chain. The pipeline's focused element ("the designated focused area of the `Document`" in the specification's terms) is considered unfocused if this flag is cleared, per the definition of [being focused][5].

`Document#hasFocus` can now return the correct answer using this flag.

### Track system focus state

    feat: track the system focus states of top-level browsing contexts
    feat(winit): handle focus events
    feat(script): consider system focus when determining a focus chain
    feat(script): consider system focus in `Document::HasFocus`
    fix(script): send `ScriptMsg::Focus` even if the top-level BC doesn't have system focus

An embedder can now notify Servo about top-level browsing contexts' system focus states. The system focus state of a top-level browsing context is broadcasted to its fully-active pipelines by `ConstellationControlMsg::SystemFocus` messages. A pipeline's focused element is considered unfocused if it doesn't have system focus, per the definition of [being focused][5].

If the embedder does nothing, top-level browsing contexts will be considered to have system focus by default.

### Eventual consistency

    fix: add sequence numbers to focus messages, send system focus state in `change_session_history`

Some focus messages are now associated with sequence numbers. They are used to resolve the inconsistency that occurs due to bidirectional focus state synchronization and provide eventual consistency. Consider the following example:

    script                            constellation
    -----------------------------------------------------------------------
    send ActivateDocument ----------> receive ActivateDocument
                                ,---- send FocusDocument
                                |
    focus an iframe             |
    send Focus -----------------|---> receive Focus
                                |     focus the iframe's content document
    receive FocusDocument <-----'     send FocusDocument to the content pipeline --> ...
    unfocus the iframe
    focus the document
    Final state:                      Final state:
     the iframe is not focused         the iframe is focused

When the above sequence completes, from the script thread's point of view, the iframe is unfocused, but from the constellation's point of view, the iframe is still focused.

This inconsistency can be resolved by associating a sequence number to each message. Whenever a script thread initiates a focus operation, it generates and sends a brand new sequence number. The constellation attaches the last-received sequence number to each message it sends. This way, the script thread can discard out-dated incoming focus messages, and eventually, all actors converge to the consistent state which is determined based on the last focus message received by the constellation.

    script                            constellation
    -----------------------------------------------------------------------
    send ActivateDocument ----------> receive ActivateDocument
                                ,---- send FocusDocument (0)
                                |
    seq_number += 1             |
    focus an iframe             |
    send Focus (1) -------------|---> receive Focus (1)
                                |     focus the iframe's content document
    receive FocusDocument (0) <-'     send FocusDocument to the content pipeline --> ...
    ignore it because 0 < 1

    Final state:                      Final state:
     the iframe is focused             the iframe is focused

### Miscellaneous changes

    feat(script): consider system focus when determining a focus chain
    test: don't assume that `document.body.focus()` moves focus
    fix(script): don't start a focus transaction of `FocusType::Element` if the given element is not focusable

Previously, calling the `focus` method on a non-focusable element caused `document.body` to acquire focus in contrary to what the specification says. This behavior is likely a bug and therefore rectified.

    test: add `/mozilla/tests/mozilla/focus_inter_documents.html`

A new test case is added to test the various complex behaviors this PR intends to implement. All subtests pass on Firefox and Servo.

[1]: https://html.spec.whatwg.org/multipage/#focusing-steps
[2]: https://html.spec.whatwg.org/multipage/#dom-window-focus
[3]: https://html.spec.whatwg.org/multipage/#dom-window-blur
[4]: https://html.spec.whatwg.org/multipage/#focus-update-steps
[5]: https://html.spec.whatwg.org/multipage/#focused

---

## Checklist

- [x] These changes fix #8981, #9482

---


jdm:
Rebased in #36649. Apologies for not getting to this back when you opened the PR!

PR #28572, opened 2021-08-15
Allow gradient interpolation hints to influence color-stop fixup

Bryce-MW:
I am working on the [easing gradients feature for CSS](https://github.com/w3c/csswg-drafts/issues/1332). I want to eventually implement that in Servo as a demonstration since Servo is a bit easier (in my opinion) to modify than Firefox or Chromium. However, at this point, Servo does not even support gradient interpolation hints so I am working towards that support. I think that it will require modifications to [webrender](https://github.com/servo/webrender) if we don't want to use [Firefox's approximation](https://searchfox.org/mozilla-central/rev/f98dad153b59a985efd4505912588d4651033395/layout/painting/nsCSSRenderingGradients.cpp#315-391). I am happy to implement that approximation though if we want something that works now.

This pull request just makes it so that the interpolation hints influence the position of unpositioned stops. Specifically, on [my test website](https://brycemw.ca/experiments/gradients), the first and third gradients look the same prior to this fix and they look correct after this fix. It is also important to note that the white colour-stop of the first gradient should be further to the right than the white colour-stop of the second gradient. There's a bit more [discussion about this effect here](https://github.com/w3c/csswg-drafts/issues/3931).

I've implemented this in both layout_2017 and layout_2020. I had to change a few things in 2017 that I don't fully understand so I would appreciate if someone could make sure that I didn't cause a subtle bug or performance issue. Especially around line 88 and 116 of `components/layout/display_list/gradient.rs`.

---

- [X] `./mach build -d` does not report any errors (and `./mach build -d --with-layout-2020`)

- [X] There are tests for these changes (Should be covered by the reftests)

### DIFF
--- a/components/layout/display_list/gradient.rs
+++ b/components/layout/display_list/gradient.rs
@@ -83,22 +83,24 @@ fn convert_gradient_stops(
 ) -> GradientBuilder {
     // Determine the position of each stop per CSS-IMAGES § 3.4.
 
-    // Only keep the color stops, discard the color interpolation hints.
     let mut stop_items = gradient_items
         .iter()
-        .filter_map(|item| match *item {
+        .filter_map(|item| match item {
             GradientItem::SimpleColorStop(color) => Some(ColorStop {
-                color,
+                color: Some(color),
                 position: None,
             }),
             GradientItem::ComplexColorStop {
                 color,
-                ref position,
+                position,
             } => Some(ColorStop {
-                color,
+                color: Some(color),
                 position: Some(position.clone()),
             }),
-            _ => None,
+            GradientItem::InterpolationHint(position) => Some(ColorStop {
+                color: None,
+                position: Some(position.clone())
+            }),
         })
         .collect::<Vec<_>>();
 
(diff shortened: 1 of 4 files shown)


Bryce-MW:
@bors-servo try=wpt-2020

Adding this since I modified layout_2020 (https://github.com/servo/servo/wiki/Layout-2020)

mrobinson:
@emilio Would you mind taking a look at this one to see if the changes to style are something that you'd want in Gecko?

emilio (changes requested):
So the style changes are only needed so that `components/layout/display_list/gradient.rs` doesn't discard the interpolation hints. That seems like a smell.

It seems like instead `convert_gradient_stops` could work with `GradientItem` directly, and then style doesn't need to change at all.

mrobinson:
I think this PR might be abandoned? I doesn't really change test results as far as I can see and it needs to be reworked. I'm going to close for now, but @Bryce-MW please feel free to reopen if you are still working on this. Servo is more active now so it's quite likely that this can land with a little bit of TLC.

Bryce-MW:
I gave up on trying to get this merged a long time ago. I haven't been actively working on gradients for a while but if I did come back I'd likely start over rather than reworking this PR. Thanks!

PR #28573, opened 2021-08-15
Fix compiler warnings

Bryce-MW:
While working on my other pull request, I noticed that there were a few compiler warnings that are easy to fix since they simply note some deprecated syntax that can be fixed with no semantic difference. I decided to quickly fix them and submit a PR.

There was also one place where an allowance of unsafe code was not being applied to the correct thing in `components/script/dom/cssrulelist.rs`. It was being applied to a macro `unsafe_no_jsmanaged_fields!` which does contain unsafe code. My understanding (and what the compiler warning states) is that applying that annotation to a macro does nothing even if the macro contains unsafe code. So what I have done is moved this annotation to apply to the entire file. I do not know if that is even necessary since the macro definition is in another file. It would be good to know since it would be best to not add extra allowances when they aren't needed. The compiler warning is below if that helps:
```
warning: unused attribute `allow`
  --> components/script/dom/cssrulelist.rs:24:1
   |
24 | #[allow(unsafe_code)]
   | ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_attributes)]` on by default
note: the built-in attribute `allow` will be ignored, since it's applied to the macro invocation `unsafe_no_jsmanaged_fields`
  --> components/script/dom/cssrulelist.rs:25:1
   |
25 | unsafe_no_jsmanaged_fields!(RulesSource);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

---

- [ ] These changes fix #___ (GitHub issue number if applicable) (No issue number)

- [X] These changes do not require tests because they fix compiler warnings in a way that does not change semantics except as noted.

### DIFF
--- a/components/layout/flow.rs
+++ b/components/layout/flow.rs
@@ -627,7 +627,7 @@ bitflags! {
         #[doc = "virtual call is too expensive."]
         const FLOATS_RIGHT = 0b0000_0000_0000_0100_0000_0000;
         #[doc = "Text alignment. \
-
+                 \
                  NB: If you update this, update `TEXT_ALIGN_SHIFT` below."]
         const TEXT_ALIGN = 0b0000_0000_0111_1000_0000_0000;
         #[doc = "Whether this flow has a fragment with `counter-reset` or `counter-increment` \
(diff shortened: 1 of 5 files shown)


jdm:
@bors-servo r+
Thanks!

Bryce-MW:
It looks like the failing checks are failing on master as well so they are probably not caused by this change. I have no idea what is causing them but once master is fixed, I can rebase and hopefully have the checks pass here too.

jdm:
@bors-servo retry
* #28574

PR #28581, opened 2021-08-17
The condition for exposing a cross-origin setter should be `CrossOriginWritable`, not `CrossOriginReadable`

yvt:
Fixes `Location#href` being inaccessible from a cross-origin document.

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

---
- [x] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #28583, opened 2021-08-18
Implement the IDL security check

yvt:
<https://heycam.github.io/webidl/#es-operations>
<https://html.spec.whatwg.org/multipage/#integration-with-idl>

This PR implements the IDL security check, which is an important (but not only) mechanism to restrict cross-origin DOM accesses.

This implementation follows WebKit's behavior and not the specification for the following reasons:

 - Neither Gecko nor WebKit implements the specification's behavior.
 - This would require a relatively elaborate mechanism *just* to have access to the target object's `CrossOriginProperties`.

---

---


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1923,10 +1930,20 @@ def getter(attr):
                 accessor = 'get_' + self.descriptor.internalNameFor(attr.identifier.name)
                 jitinfo = "0 as *const JSJitInfo"
             else:
-                if attr.hasLenientThis():
-                    accessor = "generic_lenient_getter"
+                if attr.getExtendedAttribute("CrossOriginReadable"):
+                    assert not attr.hasLenientThis(), \
+                        "CrossOriginWritable && LenientThis: not supported"
```
yvt, line 1935: Oops, I did [it](https://github.com/servo/servo/pull/28581) again!

jdm:
This is great, and I love the CallPolicy additions. I'm going to rebase these changes and get them merged.

jdm:
After a detour to fix javascript: URL evaluation which was causing lots of tests to fail intermittently, this is now ready to merge!

jdm (approved):

delan:
hi! i’m working on the monthly update. i did a bunch of testing to understand exactly what this patch fixes and what it doesn’t fix, but i’m still not 100% sure. it seems like before this patch, cross-site DOM accesses were already correctly rejected by servo, but same-site cross-origin DOM accesses were incorrectly allowed by servo. and after this patch, we fix one kind of incorrect access, specifically those that involve binding a Window or Location method in this document with a `this` from the other document. is that correct?

```html
<!--
    1. save this file as test.html
    2. serve the directory with both `python -m http.server 8000` and `8001`
    3. go to <http://localhost:8000/test.html>
-->
<!-- case 1: same-origin -->
<!-- <iframe id=i src="http://localhost:8000/"></iframe> -->
<!-- case 2: same-site cross-origin -->
<iframe id=i src="http://localhost:8001/"></iframe>
<!-- case 3: cross-site -->
<!-- <iframe id=i src="https://servo.org/"></iframe> -->
<script>
    setTimeout(()=>{
        try {
            // for case 2...
            // servo 0.1.0: returns `"#document"`
            // servo 0.2.0: returns `"#document"`
            // firefox 150: throws DOMException
            alert(i.contentWindow.document.nodeName);
        } catch (e) {
            console.log(e);
        }
        try {
            // for case 2...
            // servo 0.1.0: returns `undefined`
            // servo 0.2.0: returns `undefined`
            // firefox 150: throws DOMException
            alert(i.contentWindow.stop());
        } catch (e) {
            console.log(e);
        }
        try {
            // for case 2...
            // servo 0.1.0: returns `undefined`
            // servo 0.2.0: throws SecurityError
            // firefox 150: throws DOMException
            const d = Object.getOwnPropertyDescriptor(window,"stop");
            alert(d.value.bind(i.contentWindow)());
        } catch (e) {
            console.log(e);
        }
    },250)
</script>
```

jdm:
That is correct. This only affects the exotic WebIDL interfaces, Window and Location. #44649 is another big missing piece here.

PR #28587, opened 2021-08-21
Make `Location` operate on the correct objects

yvt:
This PR updates the implementation of the "[`Location`-object navigate][1]" algorithm to use the correct browsing context as the source browsing context for navigation. This affects the determination of a referrer and referrer policy and the treatment of `javascript:` URLs.

This PR also fixes the derivation of a `Location` object's [relevant `Document`][2] to match the specified behavior.

[1]: https://html.spec.whatwg.org/multipage/#location-object-navigate
[2]: https://html.spec.whatwg.org/multipage/#relevant-document

---


mrobinson (approved):
A bunch of us reviewed this together, as well and it looks good to us, apart from a few nits.

I will:
 - Apply a new commit that fixes these nits.
 - Update test results that are now passing.
 - Squash all commits into a single one
 - Land this change.

@yvt thank you for the contribution and apologies for the long wait for a review.

PR #28605, opened 2021-09-27
Update the version of webrender used.

### LINKED ISSUE #28585: Work on updating the version of webrender used
Hi!

As I mentioned in one of my other PRs, I am working on improving Servo's gradient support. I want to eventually add easing gradients but for now I am working on adding gradient midpoints. To do both of those well (i.e. not an approximation), there needs to be support from webrender. It's probably best to make any additions to the current version of webrender but that means that Servo has to actually work with the current version of webrender. I've been working on this for the past few days [on my fork](https://github.com/Bryce-MW/servo/commits/webrendr-update). I just wanted to make this known so no one else  tries to do the same thing at the same time.

What I am doing first is just going through and doing the minimum needed to get everything to build on the newer versions. Later I will do the cleanup and fix bugs. I have so far introduced one bug which is where I can no longer scroll web-pages but they are interactive otherwise. I've done some debugging but not gotten anywhere so I am going to wait to work on fixing that until later.

Once I am done, I will make a PR but I am happy to do it earlier as a draft if that's better.

Bryce-MW:
So as mentioned in #28585, I have been working on updating Webrender. I have started school so I haven't had the time to contribute any further but I thought I would open a pull request with where I've gotten so far. There is one known regression which is that scrolling does not work properly. The reason for this is that Webrender removed the way that we used to do scrolling. We must now keep track of the current scroll position ourselves. I am not totally sure how to do that. Right now I pretend that the desired scroll delta is actually the desired scroll position which is obviously wrong. I'll make a comment on the line where this needs to be fixed.

In addition, a number of related repositories must be updated. I have added patches so that this pull request works on its own but ideally we would merge those changes and remove the patches.

https://github.com/asajeffrey/surfman-chains/pull/12
https://github.com/servo/surfman/pull/234
https://github.com/servo/webxr/pull/198
https://github.com/servo/media/pull/383

It's totally fine if you decide not to merge. I've ended up realizing that there is a better way to do what I want to do without having to modify Webrender (yet...).

 `./mach test-tidy` has some issues with there being a couple duplicates where there shouldn't be and no duplicates when there should.

---

- [X] These changes fix #28585 (GitHub issue number if applicable)

- [X] These changes do not require tests because it is not supposed to modify existing functionality


components/compositing/compositor.rs
```diff
@@ -1141,14 +1141,41 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+                    ScrollLocation::End => LayoutVector2D::new(1000f32, 1000f32),
+                };
+                // TODO(bryce): Figure this out. I think the pre-scrolled location in the display
+                //  list needs to be modified but I don't know how to do that yet
+                //println!("{:?}", new_point);
+                //item.point_relative_to_item +
+                //let node_address = UntrustedNodeAddress(item.tag.0 as *const c_void);
+                //let node = unsafe { from_untrusted_node_address(node_address) };
+                txn.scroll_node_with_id(
+                    new_point.to_point(),
```
Bryce-MW, line 1165: This is where the logic will need to be put in place to store the scroll position. There probably needs to be more scroll logic here is general.

yvt:
FWIW I have been attempting to continue on this work in <https://github.com/yvt/servo/tree/webrendr-update>. Here's the summary of the remaining issues for Layout 2013:

 - The latest version of WebRender removed `Transaction::scroll`, which scrolls whatever scroll frame is present at the specified coordinates, for it was only used by Servo. https://github.com/servo/servo/pull/28605#discussion_r716968717 refers to the affected piece of code. The currently present API `Transaction::scroll_node_with_id` requires specifying `ExternalScrollId` and an *absolute* scroll position, both of which are hard to come by.
 - The new version of `raqote` is causing numerous Canvas2D test failures. This is a known issue: https://github.com/servo/servo/issues/25331#issuecomment-602687397
 - `/css/cssom-view/scrolling-quirks-vs-nonquirks.html`: A change in `UnioningFragmentScrollAreaIterator`'s behavior is associated with `euclid::Rect::union`'s handling of operands with negative sizes. It's used by this code, which to be frank is not so forthcoming to explain its own machinations:
   https://github.com/servo/servo/blob/9693111c9744407730d7fc971a33800a831d575e/components/layout/query.rs#L527-L535
 - `/css/css-paint-api/geometry-border-image-002.https.html`: Judging from the rendering output, it appears that the border insets are not taken into account by some sorts of clip rectangles.
 - `bug: pictures caches don't support complex transforms (thread WRRenderBackend#0, at $HOME/.cargo/git/checkouts/webrender-c3596abe1cf4f320/5b704d1/webrender/src/picture.rs:7405)`

Bryce-MW:
Thanks for continuing this on! I knew about the first issue but the others I did not. Good luck with solving them!

jsjoeio:
Wanted to test something...ignore :)

mrobinson:
Thanks for all the work on this change and apologies for letting it sit so long. A lot of the investigation has gone into #30323. That PR has landed now.

Bryce-MW:
No problem! I definitely didn't realize how big of a change this would be when I started this. Congrats on landing #30323!

PR #28627, opened 2021-11-15
[selectors] Reduces log verbosity

sanpii:
My root problem: I use the `selectors` component trough the [scraper crate](https://crates.io/crates/scraper) and I usually display the debug log when I develop, but it’s problematic when I looking for a selector in an entire html page…

I think the trace level is more appropriate for this message.

---

- [x] These changes do not require tests because: non functional modification

### DIFF
--- a/components/selectors/matching.rs
+++ b/components/selectors/matching.rs
@@ -467,9 +467,10 @@ where
     E: Element,
     F: FnMut(&E, ElementSelectorFlags),
 {
-    debug!(
+    trace!(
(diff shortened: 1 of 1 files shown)


jdm:
I think we're unlikely to accept this change. You should be able to use `RUST_LOG=debug,selectors=warn` if you want to reduce the amount of logging you see from a particular crate like selectors (see https://docs.rs/env_logger/0.9.0/env_logger/#enabling-logging for more information).

PR #28629, opened 2021-11-18
Updated cssparser dependency to 0.29.

emirror-de:
Signed-off-by: Lewin Probst <info@emirror.de>

This commit updates the cssparser dependency to 0.29. The intention is to get the selectors crate to cssparser 0.29. This will reduce problems in other crates that belong on both, selectors and cssparser.

Would love to get a review and changes required to be able to merge this.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1192,9 +1192,9 @@ dependencies = [
 
 [[package]]
 name = "cssparser"
-version = "0.28.1"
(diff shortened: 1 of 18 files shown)


jdm (approved):
Looks reasonable!

emilio:
(Ideally we should fix all the style system changes from Gecko, but this is ok for now)

Issue #28632, opened 2021-11-26
Merge selectors crate with changes from gecko?


devongovett:
It seems the `selectors` crate here (and published to crates.io) has diverged from the [copy in gecko](https://github.com/mozilla/gecko-dev/tree/5d4998ecf441db4c2e1e1753576e01b55a14a184/servo/components/selectors). For example, `cssparser` was updated to 0.29. Should the one in gecko be the one that is published to crates.io? Should the changes be merged back here? Which version is the canonical one?

I'm happy to send a PR to update `cssparser` in the crates.io released version but would like to understand the relationship between these two versions.

emilio:
The crates.io version isn't even the servo/ version, it's the servo version with the `MallocSizeOf` stuff removed, IIRC.

I used to keep the Servo version periodically up-to-date, though I haven't done so in a while (I probably should).

cortesi:
I have a project that relies on selectors, and the lack of updates to the crates.io package is making things a bit harder than they need to be. We are also a little concerned that development momentum in servo has petered out, and that the bits we rely on might stagnate. I wonder if it's time to consider extracting selectors into a free-standing module that would be easier to move forward?

paulrouget:
@emilio can you summarise the porting process from Gecko's version to a publishable version? I can take care of the migration if that's not too painful.

mrobinson:
We are definitely using something much similar to the upstream version of selectors now and we do very consistent updates.

PR #28633, opened 2021-11-26
Update cssparser

devongovett:
Updates `cssparser` to 0.29. Mainly so the `selectors` crate has an up to date version. Commit backported from gecko: https://github.com/mozilla/gecko-dev/commit/4b3dc2089155b56de5cb84674cd54602234c95bd.

See also #28632. If the current version of `selectors` is in gecko, maybe it should be released from there rather than here?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1192,9 +1192,9 @@ dependencies = [
 
 [[package]]
 name = "cssparser"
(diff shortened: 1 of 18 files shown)


devongovett:
Oh oops, just saw #28629. Maybe that one can be landed.

PR #28634, opened 2021-11-29
Preserve URI fragment across redirects

negator:
https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2

```
If the Location value provided in a 3xx (Redirection) response does
   not have a fragment component, a user agent MUST process the
   redirection as if the value inherits the fragment component of the
   URI reference used to generate the request target (i.e., the
   redirection inherits the original reference's fragment, if any).
```

---

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -831,7 +831,7 @@ pub fn http_fetch(
         }
 
         // Substep 2-3.
-        let location = response
+        let mut location = response
             .actual_response()
             .headers
             .get(header::LOCATION)
@@ -845,13 +845,15 @@ pub fn http_fetch(
             });
 
         // Substep 4.               
-        response.actual_response_mut().location_url = location.map(|res| res.map(|mut url| {
-            let current_url = request.current_url();
-            match (current_url.fragment(), url.fragment()) {
-                (fragment @ Some(..), None) => { url.set_fragment(fragment); url },
-                _                           => url
+        if let Some(ref mut location) = location {
+            if let Ok(ref mut location) = location {
+                if location.fragment().is_none() {
+                    let current_url = request.current_url();
+                    location.set_fragment(current_url.fragment());
+                }
             }
-        }));
+        }
+        response.actual_response_mut().location_url = location;
 
         // Substep 5.
         response = match request.redirect_mode {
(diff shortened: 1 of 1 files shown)


jdm:
According to https://fetch.spec.whatwg.org/#concept-response-location-url (step 4), we should do this in https://github.com/servo/servo/blob/6fced22e4784714540a2bf31b00c5f8149cbba87/components/net/http_loader.rs#L834-L845.

negator:
Done. Thanks @jdm, that was definitely the better place to put it, and much simpler.

negator:
Would it be better to perform this logic under the `Substep 4.` section instead for clarity?

jdm:
I think that would end up being clearer, yes.

components/net/http_loader.rs
```diff
@@ -844,8 +844,14 @@ pub fn http_fetch(
                     .ok()
             });
 
-        // Substep 4.
-        response.actual_response_mut().location_url = location;
+        // Substep 4.               
+        response.actual_response_mut().location_url = location.map(|res| res.map(|mut url| {
```
jdm, line 848: I think this would be more clear as:
```rust
let mut location = ...;

// Substep 4.
if let Some(ref mut location) = location {
    if location.fragment().is_none() {
        let current_url = request.current_url();
        location.set_fragment(current_url.fragment());
    }
}
response.actual_response_mut().location_url = location;
```

jdm (changes requested):

components/net/http_loader.rs
negator: The inner most `location` is a `Result` which must be destructured (more nesting) or mapped over (may be less readable). Mutating in place is probably preferable to cloning however:

```rust
        // Substep 4.               
        if let Some(ref mut location) = location {
            if let Ok(ref mut location) = location {
                if location.fragment().is_none() {
                    let current_url = request.current_url();
                    location.set_fragment(current_url.fragment());
                }
            }            
        }
        response.actual_response_mut().location_url = location;
```

negator:
Looks like a transient error:
```
rror downloading openssl-111.3.0+1.1.1c-vs2017-2019-09-18: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. The failing URL was: https://servo-deps-2.s3.amazonaws.com/msvc-deps/openssl-111.3.0%2B1.1.1c-vs2017-2019-09-18.zip
```

Can we re-kick the test?

PR #28639, opened 2021-12-06
Non-blocking network IO

negator:
The current networking strategy uses a fixed sized thread pool to perform blocking requests on a per call basis. This is inefficient, but can be ~easily~ improved by utilizing the existing tokio runtime instead to submit async networking tasks to its executor. However, since servo is currently using an outdated version of the `hyper` http library (`0.12`) we must use the [`tokio_compat`](https://github.com/tokio-rs/tokio-compat) and [`futures_compat`](https://docs.rs/futures/0.3.1/futures/compat/index.html) libraries to integrate with the older version of [`Future` used in `hyper`](https://docs.rs/hyper/0.12.1/hyper/rt/trait.Future.html). 

~**NOTE**: This PR is just proof of concept at the moment. In addition to test failures, it appears that large javascript downloads are silently failing to stream entire payloads, and occasionally getting cutoff.~

Tests are passing.

---

- [x] These changes fix [#22813](https://github.com/servo/servo/issues/22813) (GitHub issue number if applicable)

### DIFF
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -10,8 +10,10 @@ use crate::http_loader::{determine_requests_referrer, http_fetch, HttpState};
 use crate::http_loader::{set_default_accept, set_default_accept_language};
 use crate::subresource_integrity::is_response_integrity_valid;
 use content_security_policy as csp;
-use crossbeam_channel::{unbounded, Receiver, Sender};
+use crossbeam_channel::Sender;
 use devtools_traits::DevtoolsControlMsg;
+use futures_util::compat::*;
+use futures_util::StreamExt;
 use headers::{AccessControlExposeHeaders, ContentType, HeaderMapExt, Range};
 use http::header::{self, HeaderMap, HeaderName};
 use hyper::Method;
@@ -40,6 +42,9 @@ use std::ops::Bound;
 use std::str;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex};
+use tokio2::sync::mpsc::{
+    unbounded_channel, UnboundedReceiver as TokioReceiver, UnboundedSender as TokioSender,
+};
 
 lazy_static! {
     static ref X_CONTENT_TYPE_OPTIONS: HeaderName =
@@ -48,7 +53,7 @@ lazy_static! {
 
 pub type Target<'a> = &'a mut (dyn FetchTaskTarget + Send);
 
-#[derive(Clone)]
+#[derive(Clone, Deserialize, Serialize)]
 pub enum Data {
     Payload(Vec<u8>),
     Done,
@@ -58,8 +63,8 @@ pub enum Data {
 pub struct FetchContext {
     pub state: Arc<HttpState>,
     pub user_agent: Cow<'static, str>,
-    pub devtools_chan: Option<Sender<DevtoolsControlMsg>>,
-    pub filemanager: FileManager,
+    pub devtools_chan: Option<Arc<Mutex<Sender<DevtoolsControlMsg>>>>,
+    pub filemanager: Arc<Mutex<FileManager>>,
     pub file_token: FileTokenCheck,
     pub cancellation_listener: Arc<Mutex<CancellationListener>>,
     pub timing: ServoArc<Mutex<ResourceFetchTiming>>,
@@ -93,10 +98,10 @@ impl CancellationListener {
         }
     }
 }
-pub type DoneChannel = Option<(Sender<Data>, Receiver<Data>)>;
+pub type DoneChannel = Option<(TokioSender<Data>, TokioReceiver<Data>)>;
 
 /// [Fetch](https://fetch.spec.whatwg.org#concept-fetch)
-pub fn fetch(request: &mut Request, target: Target, context: &FetchContext) {
+pub async fn fetch(request: &mut Request, target: Target<'_>, context: &FetchContext) {
     // Steps 7,4 of https://w3c.github.io/resource-timing/#processing-model
     // rev order okay since spec says they're equal - https://w3c.github.io/resource-timing/#dfn-starttime
     context
@@ -110,13 +115,13 @@ pub fn fetch(request: &mut Request, target: Target, context: &FetchContext) {
         .unwrap()
         .set_attribute(ResourceAttribute::StartTime(ResourceTimeValue::FetchStart));
 
-    fetch_with_cors_cache(request, &mut CorsCache::new(), target, context);
+    fetch_with_cors_cache(request, &mut CorsCache::new(), target, context).await;
 }
 
-pub fn fetch_with_cors_cache(
+pub async fn fetch_with_cors_cache(
     request: &mut Request,
     cache: &mut CorsCache,
-    target: Target,
+    target: Target<'_>,
     context: &FetchContext,
 ) {
     // Step 1.
@@ -150,7 +155,7 @@ pub fn fetch_with_cors_cache(
     }
 
     // Step 8.
-    main_fetch(request, cache, false, false, target, &mut None, &context);
+    main_fetch(request, cache, false, false, target, &mut None, &context).await;
 }
 
 /// https://www.w3.org/TR/CSP/#should-block-request
@@ -178,12 +183,12 @@ pub fn should_request_be_blocked_by_csp(request: &Request) -> csp::CheckResult {
 }
 
 /// [Main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch)
-pub fn main_fetch(
+pub async fn main_fetch(
     request: &mut Request,
     cache: &mut CorsCache,
     cors_flag: bool,
     recursive_flag: bool,
-    target: Target,
+    target: Target<'_>,
     done_chan: &mut DoneChannel,
     context: &FetchContext,
 ) -> Response {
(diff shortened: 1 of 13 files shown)


components/net_traits/request.rs
```diff
@@ -480,6 +482,9 @@ pub struct Request {
     pub https_state: HttpsState,
 }
 
+unsafe impl Send for Request {}
```
jdm, line 485: Why are these unsafe impls necessary?

negator:
```
error: future cannot be sent between threads safely
   --> /Users/navgattu/Documents/dev-git/servo/components/net/http_loader.rs:716:1
    |
716 | #[async_recursion]
    | ^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `net_traits::request::Request`, the trait `std::marker::Sync` is not implemented for `std::cell::Cell<()>`
note: future is not `Send` as this value is used across an await
   --> /Users/navgattu/Documents/dev-git/servo/components/net/http_loader.rs:778:40
    |
778 |                 let preflight_result = cors_preflight_fetch(&request, cache, context).await;
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ first, await occurs here, with `&request` maybe used later...
note: `&request` is later dropped here
   --> /Users/navgattu/Documents/dev-git/servo/components/net/http_loader.rs:778:92
    |
778 |                 let preflight_result = cors_preflight_fetch(&request, cache, context).await;
    |                                                             --------                       ^
    |                                                             |
    |                                                             has type `&net_traits::request::Request` which is not `Send`
    = note: required for the cast to the object type `dyn futures::Future<Output = net_traits::response::Response> + std::marker::Send`
    = note: this error originates in the attribute macro `async_recursion` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is the due to the `async_recursion` macro, which is required since `http_fetch` is a recursive function (among others).

`http_fetch` calls `http_redirect_fetch` which calls `main_fetch` which calls `http_fetch`.

jdm: I would feel more comfortable removing all uses of Cell/RefCell from those types and wrapping them in Arc and Mutex instead.

negator: `Cell`/`RefCell` seems to originate from `ipc-channel`:
https://github.com/servo/ipc-channel/blob/589a62cc92d2a686769ade6ad4fa485a9904a02b/src/platform/inprocess/mod.rs#L140

https://github.com/servo/ipc-channel/blob/589a62cc92d2a686769ade6ad4fa485a9904a02b/src/platform/inprocess/mod.rs#L68

To be clear, the suggestion would be to replace those uses?

jdm: Simply wrapping the outer type in `Arc<Mutex<...>>` will be enough to allow it to compile: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2b431b626ded9270ef9c589b198317ef

negator: Yes it will. I misunderstood your suggestion above: "removing all uses of Cell/RefCell from those types"

jdm: We should wrap the ipc-channel types locally.

components/net/fetch/methods.rs
```diff
@@ -65,6 +67,9 @@ pub struct FetchContext {
     pub timing: ServoArc<Mutex<ResourceFetchTiming>>,
 }
 
+unsafe impl Send for FetchContext {}
```
negator, line 70: I missed a couple here..

jdm:
These changes appear to cause a bunch of new timeouts in tests. I'll run them again to see if it's consistent or indicative or a more general problem that affects different tests intermittently.

@bors-servo try

negator:
Guessing there was a deadlock introduced with the mutexes.

jdm:
```
  ▶ TIMEOUT [expected OK] b'/eventsource/event-data.html'
  │ 
  └ error: XDG_RUNTIME_DIR not set in the environment.

  ▶ Unexpected subtest result in /eventsource/event-data.html:
  │ TIMEOUT [expected PASS] b'EventSource: lines and data parsing'
  └   → Test timed out

  ▶ TIMEOUT [expected OK] b'/eventsource/event-data.html'
  │ 
  └ error: XDG_RUNTIME_DIR not set in the environment.

  ▶ Unexpected subtest result in /eventsource/event-data.html:
  │ TIMEOUT [expected PASS] b'EventSource: lines and data parsing'
  └   → Test timed out

  ▶ TIMEOUT [expected OK] b'/fetch/range/general.window.html'
  │ 
  └ error: XDG_RUNTIME_DIR not set in the environment.

  ▶ Unexpected subtest result in /fetch/range/general.window.html:
  │ TIMEOUT [expected FAIL] b'Fetch with range header will be sent with Accept-Encoding: identity'
  └   → Test timed out
```
Time to see if that's a stable set of timeouts.

negator:
It appears to be an issue with file fetches:
https://github.com/negator/servo/blob/1830cc8b258c52d944f562deefe47cfe453b7f75/tests/wpt/web-platform-tests/fetch/range/general.window.js#L27

https://github.com/negator/servo/blob/1830cc8b258c52d944f562deefe47cfe453b7f75/tests/wpt/web-platform-tests/eventsource/event-data.html#L25

jdm:
Same tests failed, so hopefully it's reproducible locally as well!

jdm:
In case you're unaware, `./mach test-wpt tests/wpt/web-platform-tests/eventsource/event-data.html` and `./mach test-wpt tests/wpt/web-platform-tests/fetch/range/general.window.html` should be able to run those tests locally.

negator:
It may be fixed now. Those wpt tests pass locally.

The issue was that the body needs to be streamed as a separate parallel task, otherwise we attempt to stream the entire payload at once. Which, in the case of an eventsource, may never end.

https://github.com/servo/servo/blob/66cad685b8447abdb10b22dc81e13bafdd339572/components/net/http_loader.rs#L1869


We also need to yield to allow the body streaming future to start:
https://github.com/servo/servo/blob/66cad685b8447abdb10b22dc81e13bafdd339572/components/net/http_loader.rs#L1916

components/net/fetch/methods.rs
```diff
@@ -502,22 +508,31 @@ pub fn main_fetch(
     response
 }
 
-fn wait_for_response(response: &mut Response, target: Target, done_chan: &mut DoneChannel) {
+async fn wait_for_response(
```
jdm, line 511: This function concerns me - looping on a try_recv is often bad for performance, and it's not clear to me how much the yield_now actually improves things. I propose changing DoneChannel to use tokio channels instead of crossbeam ones, which should avoid the issues with blocking that the previous implementation encountered.

components/net/resource_thread.rs
```diff
@@ -680,44 +680,58 @@ impl CoreResourceManager {
             _ => (FileTokenCheck::NotRequired, None),
         };
 
-        self.thread_pool.spawn(move || {
```
jdm, line 683: We should remove the thread_pool member if it's no longer being used.

jdm (changes requested):
Thank you! This was much more straightforward to review than I feared that it might be.

components/net/fetch/methods.rs
negator: Gave this a shot, however there are some peculiarities with adopting tokio channels.
1.) These types are from a very old version of tokio (0.1), since we need compatibility with the older version of hyper `wrap_stream`.
2.) This required the use of `futures_compat` to wrap the Receiver with a compatibility shim, which takes ownership of the receiver object.
3.) The Receiver type doesn't support multiple consumers (no clone method), so we need to use a Mutex and swap out the object in memory when consuming the stream. This can only be done once and will panic if done again.

negator:
Seeing some issues with test-tidy that aren't clear:
```
Checking files for tidiness...

./python/mach/mach/config.py:372: E126 continuation line over-indented for hanging indent

./python/mach/mach/__init__.py:0: file is empty

./python/mach/mach/main.py:105: E302 expected 2 blank lines, found 1

./python/mach/mach/main.py:530: E741 ambiguous variable name 'l'

./python/mach/mach/main.py:534: E741 ambiguous variable name 'l'

./python/mach/mach/mixin/__init__.py:0: file is empty
```

any ideas?

jdm:
Huh, I guess that's from https://github.com/servo/servo/pull/28648 but `test-tidy` doesn't actually get run during the CI builds as part of the merge. You can ignore those.

jdm:
However it looks like there's a change to Cargo.lock that didn't get committed which will need to be included for tests to run.

negator:
Got it. I've been using a patch of `mozjs` which bumps the Xcode version requirements, and sometimes forget to commit the `Cargo.[toml|lock]` files

negator:
Hmm, same failure. Lock file is up to date.

jdm:
The tidy failures will be fixed by #28649.

jdm:
Hmm, there's a huge volume of new timeouts in tests. I suspect something about a56a349 isn't working right.

negator:
Looks like it's file download timeouts.

I think the issue is related to: https://github.com/servo/servo/pull/28639#discussion_r773298869

`thread_pool` is a weak reference. Removing it as a field in CoreResourceManager probably deallocated it before it could be used in Filemanager.

negator:
@jdm Can we ensure this gets squash merged?

jdm:
Homu only supports using a merge commit, so you will need to do the squash yourself before this is merged.

jdm (approved):

jdm:
@bors-servo r+
Thank you for digging into this!

PR #28644, opened 2021-12-15
Allow builds / tests on Ubuntu 21.10

gian-carlo-d:
This PR updates the version check in `python/servo/bootstrap.py` for Ubuntu to `> 21.10` to accommodate the latest version. It also updates the `clang-format` check in `test_commands.py` to use `find()` instead of `startswith()` to account for possible distro-specific insertions into the `clang-format --version` command. 

Ubuntu, for example, outputs `Ubuntu clang-format version 13.0.0-2`, which caused `version.startswith()` to throw an error as it looked for `clang-format` at the beginning of the string. This would cause `./mach test-tidy` to skip the CPP formatting portion.
image

These two changes should help run bootstrap and test commands on the latest Ubuntu version.
---

- [X] These changes do not require tests because: I do not believe there are any current tests for the version checks (as the version checks are in-and-of themselves a test before running the bootstrap/test commands)

### DIFF
--- a/python/servo/bootstrap.py
+++ b/python/servo/bootstrap.py
@@ -374,7 +374,7 @@ def get_linux_distribution():
             raise Exception('unsupported version of %s: %s' % (distrib, version))
         distrib, version = 'Ubuntu', base_version
     elif distrib.lower() == 'ubuntu':
-        if version > '21.04':
+        if version > '21.10':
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Thanks!

jdm:
@bors-servo retry
AWS outage.

PR #28645, opened 2021-12-17
concept-response-clone: Ensure header guard is cloned after headers

negator:
https://fetch.spec.whatwg.org/#concept-response-clone

If the header guard of the response to clone is `immutable`, then copying the headers to the new response will fail with `Guard is immutable` unless we ensure the guard is copied _after_ the headers.

https://github.com/servo/servo/blob/8650794391729c6fee34bc2644ccbb85bd8fd58d/components/script/dom/response.rs#L331-L334

https://github.com/servo/servo/blob/8650794391729c6fee34bc2644ccbb85bd8fd58d/components/script/dom/headers.rs#L202-L210

https://github.com/servo/servo/blob/8650794391729c6fee34bc2644ccbb85bd8fd58d/components/script/dom/headers.rs#L73

  

---

### DIFF
--- a/components/script/dom/response.rs
+++ b/components/script/dom/response.rs
@@ -330,8 +330,8 @@ impl ResponseMethods for Response {
 
         // Step 2
         let new_response = Response::new(&self.global());
-        new_response.Headers().set_guard(self.Headers().get_guard());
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #28651, opened 2021-12-26
Update more deprecated python imports.

mvforell:
`mach` still uses some deprecated imports that were removed in Python 3.10, and currently fails when using Python 3.10.

This is in the same line as #28649.


jdm:
@bors-servo r+
Thanks!

PR #28654, opened 2021-12-30
Allow bootstrap on Pop!_OS 21.10

wusyong:
bootstrap.py doesn't support Pop!_OS 21.x yet. Adding these lines should make the script works again.

---

- [x] These changes do not require tests because it seems there no test requires certain version check and test-tidy has passed.


jdm:
@bors-servo r+
Thanks!

PR #28663, opened 2022-01-05
Convert Web IDL void to undefined

### LINKED ISSUE #27660: WebIDL's `void` type is changing to `undefined`
As part of efforts to finally add an 'undefined' type to WebIDL (so methods/etc don't have to generalize their type to 'any'), it was decided to actually add 'undefined' as the type name, and so the previous function return signature 'void' is being removed for redundancy.

We should update our IDL parser to support the new `undefined` type and remove the `void` type. Or maybe support both types until all specifications move over to new syntax.

Related PR https://github.com/heycam/webidl/pull/906

saschanaz:
Thanks to my tool https://github.com/saschanaz/gecko-webidl 🙌

Build is failing on my current VS2022+Python 3.10 environment, but the IDL tests are passing anyway, so...

---

- [x] These changes fix #27660


jdm:
@bors-servo r+
Thanks!

PR #28669, opened 2022-01-09
Look in current directory for resources directory too

ghost:
This change scans for the resources directory both from the env::current_exe() directory, as well as the current directory, so if your exe is outside of the servo repo, it can still be found.

---


5225225:
I had a quick look around, and couldn't see any tests that would have called this. Unclear to me exactly *how* you'd test this.

PR #28702, opened 2022-01-20
Add vscode/rust-analyzer info.

teymour-aldridge:
I _think_ this configuration works.

Maybe someone else can test and confirm?


sagudev:
Hm, Rust-analyzer docs for `rust-analyzer.checkOnSave.overrideCommand` says:
> If you're changing this because you're using some tool wrapping Cargo, you might also want to change #rust-analyzer.cargo.buildScripts.overrideCommand#.

I am doing some testing on gitpod with this setting, I will report how it will go.

PR #28708, opened 2022-02-06
Run layout sequentially on script thread

jdm:
This branch is my in-progress, proof of concept refactor that merges the script and layout threads and turns off all parallelism in the layout 2013 implementation. The goal of this effort is to remove complexity from the layout engine so it's easier to reason about implementing incomplete layout features.

Current status:
* compiles
* lays out http://joshmatthews.net/
* re-lays out the page when resizing
* clicking a link does not cause the new page to be displayed


jdm:
The latest commits begin to show the benefits of building within Rust's supported model. We can now include references to DOM nodes in the reflow-related data, avoiding the need to transmute back and forth between opaque pointer values.

jdm:
Navigation works now. Iframes don't work yet (panic due to unimplemented RPC layout operations), nor do image loads triggered by CSS (need to return useful information from the new reflow method).

jdm:
Got iframes loading successfully again.

jdm:
CSS image loads work again. Need to figure out why window resizing always freezes now.

jdm:
Many canvas tests fail because of the following deadlock:
* constellation tries to decide if it's time to save a screenshot
* constellation sends a synchronous message to layout thread asking about web font loads
* meanwhile, script sends a synchronous message to constellation asking to create a new canvas context
* neither thread processes the pending messages while waiting for a reply

jdm:
Note to self: new layout traversal code is broken, because we end up walking up the entire DOM from each leaf node. We want to actually process all children for a node before processing that node, recursing up the tree after each level is complete.

jdm:
There is a behaviour change in the latest commit:
* `/css/CSS2/syntax/at-charset-quotes-001.xht` hits a panic in https://github.com/servo/servo/blob/69aa81838d8f38fdc09aa8d3d88a5878e383d3a4/components/script/dom/window.rs#L1817 because the document reports it still needs reflow after reflow completes

Reduced test case:
```html
<link type="text/css" charset="us-ascii" rel="stylesheet" href="tmp.css" />
a
```
tmp.css is an empty file. The presence of an external stylesheet and a text node is enough to trigger this assertion failure.

jdm:
Pretty consistent timeouts for:
```
  ▶ TIMEOUT [expected OK] b'/_mozilla/css/animations/transition-events.html'

  ▶ Unexpected subtest result in /_mozilla/css/animations/transition-events.html:
  │ TIMEOUT [expected PASS] b'transitionrun, transitionstart, transitionend with negative delay'
  └   → Test timed out

  ▶ TIMEOUT [expected OK] b'/css/css-transitions/events-001.html'

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  │ TIMEOUT [expected PASS] b'transition:padding changing padding-left'
  └   → Test timed out

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  └ NOTRUN [expected PASS] b'transition:padding changing padding'

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  └ NOTRUN [expected PASS] b'transition:padding changing padding but not padding-bottom'

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  └ NOTRUN [expected PASS] b'transition:padding-left changing padding-left'

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  └ NOTRUN [expected PASS] b'transition:padding-left changing padding'

  ▶ Unexpected subtest result in /css/css-transitions/events-001.html:
  └ NOTRUN [expected PASS] b'transition:padding-left changing padding but not padding-bottom'

   ▶ TIMEOUT [expected FAIL] b'/css/CSS2/margin-padding-clear/padding-left-applies-to-008.xht'
```

mrobinson:
I'm really starting to think this is a good idea. Perhaps we don't need to disable all parallelism in layout, but layout can't really do any layout when script is blocked and script is really the only thing that needs to do layout queries. Maybe we can do a limited version of this as a trial.

PR #28709, opened 2022-02-07
Pare down CI to minimal mozjs build on macOS.

### LINKED ISSUE #28704: macOS CI is broken
It started within the past day, and manifests as the mozjs build reporting that the setuptools python package is not found. We should see if there were any GitHub environment changes made.

jdm:
Attempting to isolate and fix #28704.


jdm:
Woo! Continuing in #28710.

PR #28724, opened 2022-03-06
Disable UWP packaging on CI.

jdm:
Due to #28721, we cannot merge any PRs until we regenerate the code signing certificate for UWP builds. Until that happens, let's turn off UWP packaging on PRs.


jdm:
@bors-servo retry
* #28730

PR #28736, opened 2022-03-08
Change font serialisation to use IpcBytesReceiver

### LINKED ISSUE #28731: Loading large fonts in debug builds is painful
`tests/wpt/mozilla/tests/mozilla/input_value.html` times out in a debug build on macOS because the fallback font selected that contains the glyphs used in the test is Arial Unicode MS, which requires loading and serializing 23mb of font data.

* https://github.com/servo/servo/blob/ecb4f4225acf1742671927a4132ca58de69483ff/components/gfx/font_cache_thread.rs#L528-L549
* https://github.com/servo/servo/blob/ecb4f4225acf1742671927a4132ca58de69483ff/components/gfx/font_cache_thread.rs#L126
* https://github.com/servo/servo/blob/ecb4f4225acf1742671927a4132ca58de69483ff/components/gfx/font_cache_thread.rs#L36
* https://github.com/servo/servo/blob/ecb4f4225acf1742671927a4132ca58de69483ff/components/gfx/platform/macos/font_template.rs#L46

One idea I'm toying with is creating a separate serialization format for IPC font template information (rather than sending the raw FontTemplateData type), which could make use of [IpcBytesSender/IpcBytesReceiver](https://github.com/servo/ipc-channel/blob/589a62cc92d2a686769ade6ad4fa485a9904a02b/src/ipc.rs#L136-L141), which would avoid all of the serde machinery and should improve serialization performance.

s-maurice:
use `SerializedFontTemplate` and I`pcBytesReceiver` to send font template when responding to GetFontTemplate
---

- [x] These changes fix #28731 (GitHub issue number if applicable)

### DIFF
--- a/components/gfx/font_cache_thread.rs
+++ b/components/gfx/font_cache_thread.rs
@@ -37,6 +37,25 @@ pub struct FontTemplateInfo {
     pub font_key: webrender_api::FontKey,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct SerializedFontTemplateInfo {
+    pub serialized_font_template: SerializedFontTemplate,
+    pub font_key: webrender_api::FontKey,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct SerializedFontTemplate {
+    identifier: Atom,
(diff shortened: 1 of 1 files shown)


jdm:
Thank you for fixing this

PR #28742, opened 2022-03-15
Ensure reflow damage is calculated when hypothetical box changes.

jdm:
The layout traversal code has special handling for deciding whether to traverse text nodes or not. Specifically, unless the parent of a text node is restyled in a way that requires reconstructing the text fragment, we skip processing the text nodes as a performance optimization.

The optimization causes a problem in hypothetical-box-dynamic.html, which tests that the rendering doesn't change when an absolutely-positioned span has an explicit block display removed (because absolutely-positioned elements should have hypothetical block boxes, regardless). However, the code to decide what restyling damage occurs when the block display is removed did not consider the hypothetical display, only the internal display value that treats absolute positioning as block display.

By explicitly comparing the hypothetical display as well as the internal display, we ensure the element is restyled appropriately and the layout matches after the change. 

---


CYBAI:
@bors-servo retry

- network issue and

```
   ▶ TIMEOUT [expected PASS] b'/css/css-text-decor/text-underline-offset-overline-vertical.html'
```

Issue #28744, opened 2022-03-16
html/body element background styling doesn't fill entire window


jdm:
This is the cause of the failure in tests/wpt/web-platform-tests/css/CSS2/abspos/abspos-containing-block-initial-004a.xht

wusyong:
Looks like the position of y is 0 and block_size is too large.
I tried to trace the y and current I could only trace to `root_flow.base.position` whose y is 0 somehow.
I didn't trace the block_size yet. But the debug log seems to have some hints too.
```
RUST_LOG=layout=debug ./mach run -d tests/wpt/web-platform-tests/css/CSS2/abspos/abspos-containing-block-initial-004a.xht 
[2022-03-19T13:43:53Z DEBUG layout_thread] layout: received layout request for: file:///home/wusyong/Desktop/servo/tests/wpt/web-platform-tests/css/CSS2/abspos/abspos-containing-block-initial-004a.xht
[2022-03-19T13:43:53Z DEBUG layout_thread] Number of objects in DOM: 10
[2022-03-19T13:43:53Z DEBUG layout_thread] layout: parallel? false
[2022-03-19T13:43:53Z DEBUG layout_thread] layout: processing reflow request for: <html> (0x2b4029b94150) (file:///home/wusyong/Desktop/servo/tests/wpt/web-platform-tests/css/CSS2/abspos/abspos-containing-block-initial-004a.xht) (query=Full)
[2022-03-19T13:43:53Z DEBUG layout_thread] First reflow, rebuilding user and UA rules
[2022-03-19T13:43:53Z DEBUG layout_thread] Shadow roots in document 0
[2022-03-19T13:43:53Z DEBUG layout_thread] Draining restyles: 9
[2022-03-19T13:43:53Z DEBUG layout::traversal] construct_flows_at: <head> (0x2b4029b941a0)
[2022-03-19T13:43:53Z DEBUG layout::traversal] Constructed flow for ServoThreadSafeLayoutNode { node: <head> (0x2b4029b941a0), pseudo: Normal }: 0
[2022-03-19T13:43:53Z DEBUG layout::traversal] construct_flows_at: <text node> (0x2b4029b945d8)
[2022-03-19T13:43:53Z DEBUG layout::construct] building flow for node: Display(257) None Static Some(Text)
[2022-03-19T13:43:53Z DEBUG layout::traversal] Constructed flow for ServoThreadSafeLayoutNode { node: <text node> (0x2b4029b945d8), pseudo: Normal }: 0
[...]

PR #28876, opened 2022-04-17
Fix footer shows at the top of page

### LINKED ISSUE #16410: Footer shows at the top of website
http://dinnerbooking.com/dk/da-DK

the footer of the site shows at the top of the browser window :)

wusyong:
This is my attempt to fix #16410 and #28744.
For now, #28744 could show correct shape initially, but start misplaced after resizing window.
And for #16410, it need to resize the window to see the div and it seems to always delayed a bit.
Any advice is welcome!

Update: Found that containing block size isn't set correctly initially because that function only check if it's root. It should include when it's a block formatting type. This should fix #16410! So I turned this into a PR.

---

- [x] These changes fix #16410 (GitHub issue number if applicable)

### DIFF
--- a/components/layout/block.rs
+++ b/components/layout/block.rs
@@ -797,8 +797,11 @@ impl BlockFlow {
             .base
             .flags
             .contains(FlowFlags::IS_ABSOLUTELY_POSITIONED));
-        if self.is_fixed() || self.is_root() {
-            // Initial containing block is the CB for the root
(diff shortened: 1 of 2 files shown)


wusyong:
Sorry just realized I couldn't just use viewport. Some sites in the issue don't work that way.
Since I couldn't turn back to draft, I'll close it and reopen once it's ready.

PR #28880, opened 2022-04-19
Fix absolute descendents being replaced when retreiving root

### LINKED ISSUE #16410: Footer shows at the top of website

wusyong:
The cause of #16410 is because calling `try_get_layout_root` will replace its `abs_descendants` even itself isn't absolute positioned. So I pushed them instead if the root isn't absolute. Also add a few docs and comments to help understand how block size of absolute flow is being calculated. 

---

- [x] These changes fix #16410 (GitHub issue number if applicable)

- [x] There are tests for these changes:
Reduced testcase from the issue.
```html
<style>
  html {
    min-height: 100%;
    position: relative;
  }
  div {
    position: absolute;
    bottom: 0;
    width: 100%;
    height: 25px;
    border: 1px solid black;
  }
</style>
<div>
</div>
```

### DIFF
--- a/components/layout/block.rs
+++ b/components/layout/block.rs
@@ -903,7 +903,7 @@ impl BlockFlow {
     ///   10.6.7.
     ///
     /// For absolute flows, we store the calculated content block-size for the flow. We defer the
-    /// calculation of the other values until a later traversal.
+    /// calculation of the other values until a later traversal at root flow.
     ///
     /// When `fragmentation_context` is given (not `None`), this should fit as much of the content
     /// as possible within the available block size.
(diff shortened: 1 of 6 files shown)


CYBAI:
Because these tests are related to position absolute, I think we have more PASS tests 🎉 

```
  ▶ PASS [expected FAIL] b'/css/CSS2/positioning/abspos-containing-block-005.xht'
  ▶ PASS [expected FAIL] b'/css/CSS2/positioning/abspos-containing-block-006.xht'
  ▶ PASS [expected FAIL] b'/css/filter-effects/filtered-html-is-not-container.html'
```

PR #28910, opened 2022-08-03
Fix some Clippy lints.

teymour-aldridge:

- [ ] These changes do not require tests because they do not change the behaviour of the code


jdm:
@teymour-aldridge You need to use `retry` if there have not been any code changes since the last try run.

PR #28978, opened 2022-09-15
`etc/shell.nix`: Add `AppKit` to `buildInputs` on Darwin

yvt:
System framework dependencies need to be specified explicitly on Darwin.

Fixes a linker error when building Servo inside a Nix shell on macOS.

Signed-off-by: yvt <i@yvt.jp>

---

---
- [x] These changes do not require tests because they don't affect the production code

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -20,7 +20,9 @@ clangStdenv.mkDerivation rec {
     # Build utilities
     cmake dbus gcc git pkgconfig which llvm autoconf213 perl yasm m4
     (python3.withPackages (ps: with ps; [virtualenv pip dbus]))
-  ];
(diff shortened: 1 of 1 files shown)


yvt:
> System framework dependencies need to be specified explicitly on Darwin.

Example ([`nixpkgs/pkgs/development/tools/rust/cargo-embed`](https://github.com/NixOS/nixpkgs/blob/1637945189070199494480264357738cc946c70c/pkgs/development/tools/rust/cargo-embed/default.nix#L26)):

```nix
rustPlatform.buildRustPackage rec {
  # ...
  buildInputs = [ libusb1 libftdi1 ] ++ lib.optionals stdenv.isDarwin [ AppKit ];
}
```

PR #29005, opened 2022-10-05
components/net: Fix a double-lock in image_cache

### LINKED ISSUE #29003: components/net: a double-lock bug in image_cache
https://github.com/servo/servo/blob/30525b4f5fe71a18c7d9520b6509fd1dd75464f8/components/net/image_cache.rs#L414-L416

In fn `track_image`,
https://github.com/servo/servo/blob/30525b4f5fe71a18c7d9520b6509fd1dd75464f8/components/net/image_cache.rs#L549-L555
The first lock is on L549;
Then self.add_listener is called on L555;
The second lock is on L569.
https://github.com/servo/servo/blob/30525b4f5fe71a18c7d9520b6509fd1dd75464f8/components/net/image_cache.rs#L568-L569

A possible fix is to add a NoLock version of add_listener() that takes an extra parameter `store: &mut ImageCacheStore` and requires self.store.lock() before calling it.

I found this bug with my static detector lockbud.

BurtonQin:
Add fn `add_listener_with_store` which requires `self.store.lock()` before calling.
Use this function instead of `add_listener` on L555 because `self.store.lock()` is called before.

---

- [] `./mach build -d` does not report any errors
- [] `./mach test-tidy` does not report any errors
- [X] These changes fix #29003 (GitHub issue number if applicable)

- [X] These changes do not require tests because the fix is straightforward.

### DIFF
--- a/components/net/image_cache.rs
+++ b/components/net/image_cache.rs
@@ -546,13 +546,13 @@ impl ImageCache for ImageCacheImpl {
 
         match cache_result {
             ImageCacheResult::Available(ImageOrMetadataAvailable::MetadataAvailable(_)) => {
-                let store = self.store.lock().unwrap();
-                let id = store
+                let mut store = self.store.lock().unwrap();
+                let id = *store
                     .pending_loads
                     .url_to_load_key
                     .get(&(url, origin, cors_setting))
                     .unwrap();
-                self.add_listener(*id, ImageResponder::new(sender, *id));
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Great find! How did you discover this?

BurtonQin:
Sorry, I made a mistake in configuring my local servo build. The commit cannot pass the build and I am working on it.

PR #29035, opened 2022-11-04
Don't hold back the delivery of pending animation events

yvt:
The script event loop can block for an indefinite period despite having pending animation-related events that it needs to deliver immediately.

The bug can cause a test timeout in `/css/css-transitions/events-001.html` but (to my knowledge) never manifested in `master` because any irrelevant messages received by the script thread afterward would unblock the event loop and cause it to process any pending animation-related events. We can reproduce the bug in `master` by removing [the `maybe_observe_paint_time` call][2] in `LayoutThread::compute_abs_pos_and_build_display_list`. This also explains the test failure [seen in #28708][1] because the same call is [commented out][3].

[1]: https://github.com/servo/servo/pull/28708#issuecomment-1080003778
[2]: https://github.com/servo/servo/blob/54485282792bd13fce8c51576565d8e7c00e380a/components/layout_thread/lib.rs#L1156-L1157
[3]: https://github.com/servo/servo/blob/03a41ffe327255958e2d426304a725dd1fd310a5/components/layout_thread/lib.rs#L1186-L1187

---

---

### DIFF
--- a/components/script/animations.rs
+++ b/components/script/animations.rs
@@ -40,7 +40,7 @@ pub(crate) struct Animations {
     pub sets: DocumentAnimationSet,
 
     /// Whether or not we have animations that are running.
-    have_running_animations: Cell<bool>,
+    has_running_animations: Cell<bool>,
 
     /// A list of nodes with in-progress CSS transitions or pending events.
     rooted_nodes: DomRefCell<FxHashMap<OpaqueNode, Dom<Node>>>,
@@ -53,7 +53,7 @@ impl Animations {
     pub(crate) fn new() -> Self {
         Animations {
             sets: Default::default(),
-            have_running_animations: Cell::new(false),
+            has_running_animations: Cell::new(false),
             rooted_nodes: Default::default(),
             pending_events: Default::default(),
         }
@@ -97,6 +97,7 @@ impl Animations {
         }
 
         self.unroot_unused_nodes(&sets);
+        //self.update_running_animations_presence(window);
     }
 
     /// Cancel animations for the given node, if any exist.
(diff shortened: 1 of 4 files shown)


components/script/animations.rs
```diff
@@ -138,20 +144,38 @@ impl Animations {
...
+        self.update_running_animations_presence(window, Some(have_running_animations));
     }
 
-    fn update_running_animations_presence(&self, window: &Window, new_value: bool) {
-        let have_running_animations = self.have_running_animations.get();
-        if new_value == have_running_animations {
+    fn update_running_animations_presence(
+        &self,
+        window: &Window,
+        new_have_running_animations: Option<bool>,
```
mrobinson, line 153: Accepting an Option<bool> here makes things a bit confusing, so I would just keep this function more-or-less the same, but add a new helper function:

```rust
    fn update_running_animations_presence(&self, window: &Window, new_value: bool) {
        let had_running_animations = self.has_running_animations.get();
        if new_value == had_running_animations {
            return;
        }

        self.has_running_animations.set(new_value);
        self.handle_animation_presence_or_pending_events_change(window);
    }

    fn handle_animation_presence_or_pending_events_change(&self, window: &Window) {
        let has_running_animations = self.has_running_animations.get();
        let has_pending_events = self.pending_events.borrow().is_empty();

        // Do not send the NoAnimationCallbacksPresent state until all pending
        // animation events are delivered.
        let state = match has_running_animations || has_pending_events {
            true => AnimationsPresentState::AnimationsPresent,
            false => AnimationsPresentState::NoAnimationsPresent,
        };
        window.send_to_constellation(ScriptMsg::ChangeRunningAnimationsState(state));
    }

```

components/script/animations.rs
```diff
@@ -488,6 +512,8 @@ impl Animations {
                     .fire(node.upcast());
             }
         }
+
+        self.update_running_animations_presence(window, None);
```
mrobinson, line 516: With the above change this would become:

```rust
        if self.pending_events.borrow().is_empty() {
            self.handle_animation_presence_or_pending_events_change(window);
        }
```

This requires an early return at the top if `events.is_empty()`.

components/script/animations.rs
```diff
@@ -97,6 +102,7 @@ impl Animations {
         }
 
         self.unroot_unused_nodes(&sets);
+        self.update_running_animations_presence(window, None);
```
mrobinson, line 105: I'm not totally sure I understand the purpose of this addition. Is it addressing a particular issue?

mrobinson (changes requested):
Sorry that it has taken me until now to review this change. This is a great breakdown of a very tricky issue and a nice fix! Thank you so much for tracking it down. I would squash all of these commits into one so that the test isn't broken if someone checks out an intermediate revision. If you are too busy to apply the suggestions below, let me know and I'll be happy to do it.

mrobinson:
Also, amazing commit messages! They helped a lot in understanding this change.

PR #29123, opened 2022-12-01
Replaced explicit Default implementation for Seed with derive.

mikayelgr:
This patch replaces the explicit `std::default::Default` implementation of the [Seed](https://github.com/servo/servo/blob/master/components/rand/lib.rs) struct with a derive macro. 

---

- [x] These changes do not require tests.


jdm:
@bors-servo r+
Thanks!

michaelgrigoryan25:
Of course!

Issue #29149, opened 2022-12-16
App Crashes because wayland not gud :c


Snowiiii:
wayland-client error] Attempted to dispatch unknown opcode 0 for wl_shm, aborting.
Redirecting call to abort() to mozalloc_abort

Stack trace for thread "main"
   0: servo::backtrace::print
   1: servo::crash_handler::install::handler
   2: <unknown>
   3: _ZL22AnnotateMozCrashReasonPKc
             at /home/snowii/Downloads/servo/target/release/build/mozjs_sys-37b260a6bf2b2075/out/build/dist/include/mozilla/Assertions.h:40:19
      _ZL9MOZ_CrashPKciS0_
             at /home/snowii/Downloads/servo/target/release/build/mozjs_sys-37b260a6bf2b2075/out/build/dist/include/mozilla/Assertions.h:260:3
      mozalloc_abort
             at /home/snowii/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/f87705d/mozjs/mozjs/memory/mozalloc/mozalloc_abort.cpp:35:3
   4: abort
             at /home/snowii/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/f87705d/mozjs/mozjs/memory/mozalloc/mozalloc_abort.cpp:88:3
   5: wayland_client::imp::proxy::proxy_dispatcher
   6: <unknown>
   7: wl_display_dispatch_queue_pending
   8: wl_display_roundtrip_queue
   9: scoped_tls::ScopedKey<T>::set
  10: smithay_client_toolkit::environment::Environment<E>::new
  11: winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::new
  12: winit::platform_impl::platform::EventLoop<T>::new
  13: servo::app::App::run
  14: servo::main
  15: std::sys_common::backtrace::__rust_begin_short_backtrace
  16: std::rt::lang_start::{{closure}}
  17: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/c5d82ed7a4ad94a538bb87e5016e7d5ce0bd434b/library/core/src/ops/function.rs:606:13
      std::panicking::try::do_call
             at /rustc/c5d82ed7a4ad94a538bb87e5016e7d5ce0bd434b/library/std/src/panicking.rs:483:40
      std::panicking::try
[...]

PR #29168, opened 2022-12-20
Finally Servo works with newer winit

Snowiiii:
Update Dependencies & Rust Edition 

Using Surfman cargo override for winit update 
(yay servo on linux)

---

- [X] These changes fix #[29149](https://github.com/servo/servo/issues/29149) 

- [X] These changes do not require tests because Dependency update


jdm:
Could you revert all the rust edition changes and updates to dependencies that are not required to upgrade winit? By making the changes as small as possible, it makes it easier to identify the cause of any regressions.

Snowiiii:
i put much time to Update the Dependencies so that all is working fine and Rust Edition 2021 is just the "Standart" ,CL is unhappy because of an CMake issue if i see correct
are you sure what i should only use it for winit ? if you want you can test the build yourself

jdm:
This is a change that could affects every platform differently. Theore changes the make at the same time, the harder it is to know which change is actually responsible for a problem. The best practice is to start by updating only winit, and see if anything else needs to change.

Snowiiii:
Done and works

jdm:
https://github.com/Snowiiii/servo/pull/1 has fixes for macOS.

Snowiiii:
Merged

okias:
Nice.

Snowiiii:
If i understand the Check fails becausd of CMake issue ?

jdm:
Huh, that's new:
```
   CMake Error at cmake_install.cmake:46 (file):
    file INSTALL cannot copy file
    "/home/runner/work/servo/servo/target/release/build/fontsan-28a8f43cbd285306/out/build/libots.a"
    to "/usr/local/lib/libots.a": Success.
```
We should check whether it's reprodicible in the CI for https://github.com/servo/fontsan.

jdm:
Huh. Seems like it's something about the changes in this PR in particular, since #29184 just merged successfully.

Cargo.toml
```diff
@@ -16,20 +16,7 @@ opt-level = 3
...
-#
-#     <crate> = { path = "/path/to/local/checkout" }
-#
-# Or for a git dependency:
-#
-#     [patch."https://github.com/servo/<repository>"]
-#     <crate> = { path = "/path/to/local/checkout" }
-
-# This is here to dedupe winapi since mio 0.6 is still using winapi 0.2.
-mio = { git = "https://github.com/servo/mio.git", branch = "servo-mio-0.6.22" }
```
atouchet, line 30: Is this mio fork supposed to be removed here? Isn't it still required for using a compatible version of winapi?

jdm: The newer versions of winit use a newer version of mio, so the fork can no longer be used. It was only necessary for UWP builds, and we're not shipping those in the Windows app store any more, so it's less important.

atouchet: mio 0.6 is still being used as a dependency of other crates though.

Snowiiii:
was gone a bit and was not able to code because im sick,The new fontsan patch passed the check fine

Snowiiii:
so what now.CI does not make sense if you want i can show you that its building fine locally

jdm:
```
[[patch.unused]]
name = "fontsan"
version = "0.4.0"
source = "git+https://github.com/Snowiiii/fontsan.git?branch=master#2ef5bc9ba3a58912189f8c8b1b18361a79bac40f"
```
This means the override isn't actually being used for some reason.

Snowiiii:
so @jdm what to do next i dont know what the patch is unused

atomgardner:
@Snowiiii, are you still working on this?

Snowiiii:
Hey sorry but i guess im done with Servo,Servo is very Compilcated and currently i dont have desire or Time to work on it, But it would be Very Nice if someone else Port to Winit 0.28

PR #29197, opened 2023-01-03
Cancel main build if successful build already exists for commit

### LINKED ISSUE #28326: Verify auto branch matches before starting jobs
Homu performs two pushes to auto every time a merge is attempted - one for the current master revision, immediately followed by the new merge commit. On taskcluster the decision task fails to run for the first one because the revision doesn't match any more, but for github actions all of the jobs happily go on their merry way and waste time and resources.

delan:
Homu pushes both the old and new commits for each merge, causing the main build to run twice for each commit.

This patch adds a “decision job” to the main build, which cancels the run if a successful main build already exists for the commit. The jobs that build Servo on each of the five platforms in turn block until that new job finishes.

---

- [x] These changes fix #28326

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,9 +13,31 @@ env:
   SHELL: /bin/bash
 
 jobs:
+  decision:
+    name: Decision job
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Previous successful build check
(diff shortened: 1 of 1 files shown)


jdm:
Oh, wow, this is very slick. Nice work!
@bors-servo r+

PR #29198, opened 2023-01-03
CI: remove branch check steps that don’t actually do anything

delan:
f165e17ce2fb8795c99dbb998336b7fa54b68f5f (#28778) adds a step to each job of the main build that ostensibly bails out of the job, if the context branch is not `auto` or one of the `try` branches relevant for that platform. But a step that runs `exit 0` doesn’t really have any effect.

Since the main workflow already has `branches` limited to `auto` and the `try` branches, we can safely remove those steps without worrying about the main build running on dependabot branches etc. For example:

* [dependabot/cargo/nom-7.1.2](https://github.com/servo/servo/actions?query=branch%3Adependabot%2Fcargo%2Fnom-7.1.2) only has a run for the pull request workflow
* [this test build](https://github.com/delan/servo/actions/runs/3830596812/jobs/6518672884) shows that even if the branch check fails, the build continues

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,9 +17,6 @@ jobs:
     name: Build (Windows)
     runs-on: windows-2019
     steps:
-      - name: Branch check
-        if: ${{ !contains('refs/heads/auto refs/heads/try refs/heads/try-windows', github.ref) }}
-        run: exit 0
       - uses: actions/checkout@v3
         with:
           fetch-depth: 2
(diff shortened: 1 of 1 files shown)


jdm:
Hmm, interesting. The intent behind these checks was to support things like `@bors-servo try-windows`, which pushes to a branch called `try-windows` instead of `try. It would be nice to find a solution that actually works for that.

jdm:
That being said, it doesn't need to happen in this PR, and we can remove things that are not actually making any difference.
@bors-servo r+

delan:
Thanks for the context! I have some ideas for how to solve that, but yeah, let’s do that in a second patch.

jdm:
I've deployed the saltfs changes; you shouldn't have further problems issuing homu commands.

PR #29201, opened 2023-01-04
CI: run WPT tests on macOS

### LINKED ISSUE #28284: Run WPT tests (macOS)

delan:
This patch uncomments and fixes the mac-wpt job in the main workflow, with a few changes:

* we downgrade python3 and pip3 to 3.9, because inspect.getargspec was [removed in 3.11](https://docs.python.org/3/whatsnew/3.11.html#removed)
    * we should apparently move to [inspect.getfullargspec](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec)
* we download the real Servo artifact, rather than the fake Servo for testing

We also align the rest of the job (other than `--timeout-multiplier`) to be as close as possible to linux-wpt, based on the assumption that linux-wpt is more up to date. This includes:

* setting strategy.fail-fast to false
* adding a failure() condition to the “Archive logs” step
* adding the chunk id to log filenames

There are some failing WPT tests, so for now the job is not required to succeed.

Example run: <https://github.com/delan/servo/actions/runs/3837530905>

---

- [x] These changes fix #28284 (GitHub issue number if applicable)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,87 +13,6 @@ env:
   SHELL: /bin/bash
 
 jobs:
-  build-win:
-    name: Build (Windows)
-    runs-on: windows-2019
-    steps:
-      - name: Branch check
-        if: ${{ !contains('refs/heads/auto refs/heads/try refs/heads/try-windows', github.ref) }}
-        run: exit 0
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 2
-      - name: Copy to C drive
-        run: cp D:\a C:\ -Recurse
-      - name: Bootstrap
-        working-directory: "C:\\a\\${{ github.event.repository.name }}\\${{ github.event.repository.name }}"
-        run: |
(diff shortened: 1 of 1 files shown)


delan:
oops, hit ctrl+enter by accident, but i’ve finished the description now!

andreubotella:
>       * we should apparently move to [inspect.getfullargspec](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec)

For the record, `mach` only ever seems to use the `args` property of `inspect.getargspec()`, and that is unchanged in `inspect.getfullargspec()`. In fact, in Python 3.10 [`getargspec` is implemented as a wrapper](https://github.com/python/cpython/blob/3.10/Lib/inspect.py#L1211) over `getfullargspec` that passes through a number of its properties, including `args`. So updating `mach` to support Python 3.11 should be trivial.

jdm (approved):

jdm:
Yikes.

PR #29222, opened 2023-01-10
mach filter-intermittents: add progress and --json output mode

delan:
filter-intermittents takes a while, because it needs to look up each failure against the tracker api (or GitHub api), so it would be nice to know how long we’ll need to wait when running it interactively. The output is also pretty-printed in a format that’s not the easiest to analyse.

This patch adds some progress output (on stderr, one line with \\r), plus a --json output format as follows:

[{"test": "/path/to/foo", "output": "..."}, {"test": "/path/to/bar", "output": "..."}, ...]

---

- [x] These changes do not require tests because they affect mach filter-intermittents only

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -529,6 +529,8 @@ class MachCommands(CommandBase):
                      help='Print filtered log to file')
     @CommandArgument('--log-intermittents', default=None,
                      help='Print intermittents to file')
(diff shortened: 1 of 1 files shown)


CYBAI (approved):
Thank you!

CYBAI:
as I don't have r+ permission, might need your help to r=cybai 🙏 thanks!

delan:
r=cybai

PR #29235, opened 2023-01-12
Stringify unknown JavaScript objects in global exception handlers

mrobinson:
When turning DOM exceptions into `ErrorInfo` always try to stringify
the JavaScript value, even if it's an object that isn't a `DOMException`
or native exception.  This means that exceptions that extend the `Error`
prototype are now stringified. The result is that test output for WPT
global assertion failures is more useful. For instance for the test
include-frames-from-child-same-origin-grandchild.sub.html:

Before:
```
uncaught exception: unknown (can't convert to string)
```

After:
```
uncaught exception: Error: assert_equals: expected 4 but got 3
```
---

### DIFF
--- a/components/script/dom/bindings/error.rs
+++ b/components/script/dom/bindings/error.rs
@@ -29,8 +29,7 @@ use js::jsval::UndefinedValue;
 use js::rust::wrappers::JS_ErrorFromException;
 use js::rust::wrappers::JS_GetPendingException;
 use js::rust::wrappers::JS_SetPendingException;
-use js::rust::HandleObject;
-use js::rust::MutableHandleValue;
+use js::rust::{HandleObject, HandleValue, MutableHandleValue};
 use libc::c_uint;
 use std::slice::from_raw_parts;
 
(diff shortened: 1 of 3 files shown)


jdm (approved):
Love the test!

mrobinson:
@bors-servo retry

```
   --- stderr
  pkg_config failed: `"pkg-config" "--libs" "--cflags" "dbus-1" "dbus-1 >= 1.6"` did not exit successfully: exit status: 1
  error: could not find system library 'dbus-1' required by the 'libdbus-sys' crate

  --- stderr
  Package dbus-1 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `dbus-1.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'dbus-1' found
  Package dbus-1 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `dbus-1.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'dbus-1' found

  One possible solution is to check whether packages
  'libdbus-1-dev' and 'pkg-config' are installed:
  On Ubuntu:
  sudo apt install libdbus-1-dev pkg-config
  On Fedora:
  sudo dnf install dbus-devel pkgconf-pkg-config
```

jdm:
I'm super confused by that error—we installthe dbus dev library as part of the bootstrap command.

mrobinson:
I was pretty confused as well! After some digging it seems that the apt archive was down:

From the bootstrap step:
```
 Err:40 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libnl-route-3-200 amd64 3.4.0-1
  Unable to connect to azure.archive.ubuntu.com:http: [IP: 52.154.174.208 80]
```

I'll look into making sure this causes that phase of the action to fail.

mrobinson:
Addressed CI issue in #29238.

PR #29237, opened 2023-01-13
Simplify Python code for running WPT tests

mrobinson:
Combine `run.py` and `update.py` into `servowpt.py` in order to allow them to share code. Import them directly into the mach script to avoid having to call `compile` and `exec` on the code. This makes it clearer how they are executed. In addition, move all of the setup into `setupwpt.py` to avoid differences between tests executed via mach and not. Finally, be more ambitious when detecting the build to use. If none was specified, try to use the one that exists between "release" and "debug."

---

- [x] These changes do not require tests because they only change a small script behavior.


jdm (approved):
Nice clean up.

PR #29238, opened 2023-01-13
Fail faster on CI when Linux bootstrap fails

mrobinson:
Raise an exception when dependencies fail to install. Also split the run phase of the Linux bootstrap so that either of these failing commands will cause the job to fail.

---

- [x] These changes do not require tests because they just change minor script issues.


jdm (approved):
Oh, that totally explains it.

PR #29244, opened 2023-01-16
CI: disable mac-wpt job in main workflow

delan:
In theory, the new mac-wpt jobs added in #29201 are harmless since they aren’t required for the main workflow to succeed, but in practice, the runners are so slow to assign and run that they break the build.

---


jdm:
I've canceled all other half B's inorder to get this one moving. Another PR will kill all queue progress because of the Mac builder issue discussed in zulip.

PR #29294, opened 2023-01-24
Integrate filter-intermittents into test-wpt

mrobinson:
This change integrates the `filter-intermittents` command into `test-wpt`. This is in preparation for future work on tracking intermittent failures. This change also:
- Removes `ServoJsonFormatter` and replaces it with a generic WPT log handler which tracks unexpected results.
- The intermittent filter is now controlled via environment variables and the GitHub version requires a token instead of credentials.
- Output is saved to a single file and is always text.

---

- [x] These changes do not require tests because they are changes to build infrastructure.


delan:
@bors-servo r+

Great work!

PR #29299, opened 2023-01-25
Fix upstreaming of WPT changes

mrobinson:
The GitHub Action needs access to Servo repository secrets, so switch to using the 'pull_request_target' event. Since these PRs have more complete access to the Servo repository, do not execute the version of the upstream script that comes with the PR. Instead, simply fetch the changes. To make this work, the script no longer expects the PR commit to be checked out, merely that they exist in the repository somewhere.

---

### DIFF
--- a/.github/workflows/upstream-wpt-changes.yml
+++ b/.github/workflows/upstream-wpt-changes.yml
@@ -1,6 +1,6 @@
 name: WPT export
 on:
-  pull_request:
+  pull_request_target:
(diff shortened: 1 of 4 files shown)


jdm (approved):
Nice solution!

mrobinson:
@bors-servo r=jdm

Thanks for the review!

CYBAI:
@bors-servo retry

- https://github.com/servo/servo/issues/29301

PR #29339, opened 2023-02-08
Add intermittent results for transitions related tests

mrobinson:
These tests test the behavior of many properties and due to issues in
 Servo, the results are incredibly unstable. Since the tests use large
 property lists this leads to hundreds of failed subtests every run. We
 let these tests either pass or fail so that results in the CI are
 stable. The ultimate goal here is to fix the instability in Servo so
 that these tests pass or fail consistently.
    
 This change also adds support for intermittent expectations to the
 ServoHandler. Before these kind of test results were interpreted as
 unexpected results.

---

- [x] These changes do not require tests because they just update test results.

### DIFF
--- a/tests/wpt/grouping_formatter.py
+++ b/tests/wpt/grouping_formatter.py
@@ -155,15 +155,22 @@ class ServoHandler(mozlog.reader.LogHandler):
     def test_start(self, data):
         self.running_tests[data['thread']] = data['test']
 
+    @staticmethod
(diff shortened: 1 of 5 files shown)


mrobinson:
@delan Would you mind taking a look at this change as well. I think once this lands, we can start having the CI write results to the intermittent dashboard.

delan:
Looks good!

@bors-servo r+

PR #29346, opened 2023-02-09
Aggregate unexpected results into logs

mrobinson:
This makes it easier to run `update-wpt` based on results from the bots by writing a new version of the raw log that filters out all tests with expected results.  Then aggregate this output into the filtered results archive. A future version of this could aggregate all unexpected results that were not filtered as intermittents.

---

- [x] These changes do not require tests because they are infrastructure changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -216,14 +216,16 @@ jobs:
             --release --processes $(nproc) --timeout-multiplier 2 \
             --total-chunks ${{ env.max_chunk_id }} --this-chunk ${{ matrix.chunk_id }} \
             --log-raw test-wpt.${{ matrix.chunk_id }}.log \
-            --filter-intermittents=filtered-wpt-results.${{ matrix.chunk_id }}.json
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -216,14 +216,15 @@ jobs:
...
+            --filter-intermittents=filtered-test-wpt.${{ matrix.chunk_id }}.json
       - name: Archive filtered results
         uses: actions/upload-artifact@v3
         if: ${{ always() }}
         with:
           name: wpt-filtered-results-linux
           path: |
-            filtered-wpt-results.${{ matrix.chunk_id }}.json
+            filtered-test-wpt.${{ matrix.chunk_id }}.json
+            unexpected-test-wpt.${{ matrix.chunk_id }}.log
```
delan, line 227: without reading the code, it’s not really clear where this file comes from. wdyt about lifting the `unexpected-{raw_log_file}` decision out to main.yml via a command line option?

mrobinson: This is a good point. I've plumbed this out into a command-line argument. It should be a bit clearer now.

Issue #29344, opened 2023-02-09
missing gstreamer library on MacOS


wspl:
When I build with mach.sh with `--release`, it failed with error message:
```
[Errno 2] No such file or directory: '/opt/homebrew/lib/gstreamer-1.0/libgstvideoconvert.dylib'
[Errno 2] No such file or directory: '/opt/homebrew/lib/gstreamer-1.0/libgstvideoscale.dylib'
ERROR: could not find required GStreamer DLL: /opt/homebrew/lib/gstreamer-1.0/libgstvideoconvert.dylib
ERROR: could not find required GStreamer DLL: /opt/homebrew/lib/gstreamer-1.0/libgstvideoscale.dylib
```
`libgstvideoconvert` and `libgstvideoscale` should be included in [gst-plugins-base](https://formulae.brew.sh/formula/gst-plugins-base), but it's not.

The .pkg package I downloaded from the gstreamer home page also cannot find these two libraries.

mukilan:
I'm not  familiar with gstreamer and I don't have access to Mac to test, but it looks like the 1.22 release of gst-plugins-base has introduced a new ['videoconvertscale' plugin](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/commit/d11f13f476411b828387c3c26619bc77c255affb#0f8c6b125485c13b682ebcb8fd2f7266439ac317) & removed the previously separate videoscale and videoconvert plugins.

While we investigate the exact reason the gstreamer bindings are not linking to the newer plugin,  I would recommend   downgrading to gstreamer 1.20 if its possible via homebrew.

mukilan:
This issue is happening in the [CI](https://github.com/servo/servo/actions/runs/4132942377/jobs/7142248340) environment as well.

mukilan:
Found the root cause of the issue. We hard code the plugin names [here](https://github.com/servo/servo/blob/master/python/servo/gstreamer.py#L68) and [here](https://github.com/servo/servo/blob/master/python/servo/gstreamer.py#L71), to support packaging the plugins I think.

PR #29354, opened 2023-02-12
#26488 Move worklet drop implementation into single droppable member

ben-freist:
Signed-off-by: Benjamin Freist <bfreist@soundhound.com>

Hey @jdm ,
is this issue still free? I would like to start contributing to servo.
This patch is straight from PR 28302, I thought it would be good to start with a small part, I hope that's ok. 

---

- [x] These changes do not require tests because they will be checked by the compiler.


jdm:
Yeah, I think it makes sense to do these changes piece by piece. Thanks!

jdm (approved):
We can merge this after #29344 is fixed.

PR #29359, opened 2023-02-14
Add support for the intermittent dashboard

mrobinson:
---

- [x] These changes do not require tests because this is a CI change.

### DIFF
--- a/tests/wpt/servowpt.py
+++ b/tests/wpt/servowpt.py
@@ -6,7 +6,9 @@ import dataclasses
 import grouping_formatter
 import json
 import os
+import re
 import sys
+import urllib.error
 import urllib.parse
 import urllib.request
 
@@ -14,8 +16,8 @@ import mozlog
 import mozlog.formatters
 import multiprocessing
 
-from typing import List
-from grouping_formatter import UnexpectedResult
+from typing import List, NamedTuple, Optional, Tuple, Union
+from grouping_formatter import UnexpectedResult, UnexpectedSubtestResult
 
 SCRIPT_PATH = os.path.abspath(os.path.dirname(__file__))
 SERVO_ROOT = os.path.abspath(os.path.join(SCRIPT_PATH, "..", ".."))
@@ -28,7 +30,7 @@ import update  # noqa: F401,E402
 
 TRACKER_API = "https://build.servo.org/intermittent-tracker"
 TRACKER_API_ENV_VAR = "INTERMITTENT_TRACKER_API"
-GITHUB_API_TOKEN_ENV_VAR = "INTERMITTENT_TRACKER_GITHUB_API_TOKEN"
+TRACKER_DASHBOARD_SECRET_ENV_VAR = "INTERMITTENT_TRACKER_DASHBOARD_SECRET"
 
 
 def determine_build_type(kwargs: dict, target_dir: str):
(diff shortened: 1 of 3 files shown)


tests/wpt/servowpt.py
```diff
@@ -186,41 +188,106 @@ def update_tests(**kwargs):
...
+                    'branch': context.branch_name,
+                    'build_url': context.build_url,
+                    'pull_url': context.pull_url,
+                    'attempts': attempts
+                }).encode('utf-8'),
+                headers=self.headers)
+
+            known_intermittents = dict()
+            with urllib.request.urlopen(request) as response:
+                for test in json.loads(response.read())["known"]:
```
delan, line 274: nit: i think this can be simplified to

```python
for test in json.load(response)["known"]:
```

tests/wpt/servowpt.py
```diff
@@ -186,41 +188,106 @@ def update_tests(**kwargs):
...
+            return GithubContextInformation("", "", "")
 
+        repository = github_context['repository']
+        repo_url = f"https://github.com/{repository}"
 
-class GitHubQueryFilter():
-    def __init__(self, token):
-        self.token = token
+        run_id = github_context['run_id']
+        build_url = f"{repo_url}/actions/runs/{run_id})"
```
delan, line 220: i was thinking ideally we would have the full job build url, e.g.

https://github.com/servo/servo/actions/runs/4181616898/jobs/7244211010

so you can go from the dashboard directly to the chunk that failed, and view the failed run even if someone reruns the job in place.

but i can’t see the job build id [anywhere in the context docs](https://docs.github.com/en/actions/learn-github-actions/contexts), and we rarely rerun jobs in place, and having the full job build url weakens the “all results have the same” feature anyway… so on second thought, i think this approach is the way to go :)

tests/wpt/servowpt.py
```diff
@@ -186,41 +188,106 @@ def update_tests(**kwargs):
...
-    def is_failure_intermittent(self, test_name):
-        query = urllib.parse.quote(test_name, safe='')
-        request = urllib.request.Request("%s/query.py?name=%s" % (self.url, query))
-        search = urllib.request.urlopen(request)
-        return len(json.load(search)) > 0
+    @staticmethod
+    def get_github_context_information() -> GithubContextInformation:
+        github_context = json.loads(os.environ.get("GITHUB_CONTEXT", "{}"))
+        if not github_context:
+            return GithubContextInformation("", "", "")
```
delan, line 214: wdyt about using `Optional[str]` and `None` here, rather than `""`? (the dashboard api accepts `null` for these, so no changes to the request body would be needed)

tests/wpt/servowpt.py
```diff
@@ -186,41 +188,106 @@ def update_tests(**kwargs):
...
-    def is_failure_intermittent(self, test_name):
-        url = "https://api.github.com/search/issues?q="
-        query = "repo:servo/servo+" + \
-            "label:I-intermittent+" + \
-            "type:issue+" + \
-            "state:open+" + \
-            test_name
+        commit_title = github_context["event"]["head_commit"]["message"]
+        match = re.match(r"^Auto merge of #(\d+)", commit_title)
+        pr_url = f"{repo_url}/pull/{match.group(1)}" if match else ""
```
delan, line 224: `""` → `None` (as above)

delan (commented):
Great work! These changes make sense, though I have a few minor comments.

tests/wpt/servowpt.py
mrobinson: Oh, this is a good point. I was thinking about this today. Sometimes it would be nice to know what chunk something failed in (when referring back from the [job that reports aggregated results for instance](https://github.com/servo/servo/actions/runs/4182416519/jobs/7246803223)).

tests/wpt/servowpt.py
mrobinson: Nice catch. I've fixed this in the latest version.

tests/wpt/servowpt.py
mrobinson: That makes total sense to me. I wasn't sure if the server supported that. I've made this change.

tests/wpt/servowpt.py
mrobinson: This should be done now.

mrobinson:
@delan Thanks for the review. I've updated the branch with your suggested changes.

Issue #29105, opened 2022-11-28
Please make a release of selectors which already has cssparser 0.29


federicomenaquintero:
The cssparser crate made an API break for 0.29, with the removal of the `AtRuleType` enum.  `selectors` is already updated for this.  Could we please have a release?  I'd like to update both crates in librsvg :smiley:

atouchet:
@emilio would you be able to do this? From what I can tell the version in the Servo repo has fallen behind what's being used in Firefox (#28632).

jdm:
Ah, I see why it's more complicated than that: https://github.com/servo/servo/blob/master/components/selectors/Cargo.toml#L28-L31

federicomenaquintero:
Are `to_shmem` and `to_shmem_derive` a problem for the public release?  If so, I can probably submit a PR to put them behind a feature, which Servo would enable for itself.

delan:
I’ve released:

* selectors 0.24.0 (was 0.23.0)
* servo_arc 0.2.0 (was 0.1.1)
* to_shmem 0.0.0 (empty dummy package)
* to_shmem_derive 0.0.0 (empty dummy package)

@federicomenaquintero

federicomenaquintero:
Thank you!!!  Let me try to update librsvg...

federicomenaquintero:
That worked, after the `AtRuleParser` changes.  Thanks for the release!

PR #29361, opened 2023-02-14
Add a "shmem" feature to selectors to avoid publishing that dependency

federicomenaquintero:
Hopefully as a way to help #29105, this PR adds a `shmem` feature to the selectors crate.  This is so that the crate can be released with that feature off by default, as the `ToShmem` derives are only of interest to Servo, I think.

All the places where servo's source code includes the selectors crate are made to turn on this feature.


delan:
Thanks!

@bors-servo r+

federicomenaquintero:
Thanks for the merge!

PR #29370, opened 2023-02-16
Filter unknown flaky tests when filtering known intermittents

mrobinson:
There are two kinds of flaky/intermittent tests in Servo. The traditional kind is the test that fails on the CI, but has an associated bug indicating that the test is an intermittent failure. Many of these tests have completely unstable results, for instance those where an unpredictable set of subtests fail. It's impossible to generate stable results for these, so we have traditionally simply discard these unexpected results.

Another kind of intermittent test is one that will produce an expected result when rerun (ie will flake). Some of these are also labeled with bugs, while some are not. In some cases, there is flakiness in some core Servo functionality that can lead to *any* test flaking, such as a race condition that can lead to an early screenshot for reftests. When these kinds of tests do not have associated bugs, they cause the CI to fail. In this case, it is impossible to label these tests as intermittent because it can literally be any test.

This change, reruns failed tests in order to detect unlabeled tests in the second category. Instead of blocking the CI when the second run leads to expected results, the CI will now pass, but the flake will be reported to the new flakiness dashboard. This prevents unrelated flakes from slowing down the merge queue.

---

- [x] These changes do not require tests because they are a change for CI only.

### DIFF
--- a/etc/ci/report_aggregated_expected_results.py
+++ b/etc/ci/report_aggregated_expected_results.py
@@ -97,28 +97,32 @@ class Item:
 
 def get_results() -> Optional[Item]:
     unexpected = []
-    known_intermittents = []
     for filename in sys.argv[1:]:
         with open(filename, encoding="utf-8") as file:
-            data = json.load(file)
-            unexpected += data["unexpected"]
-            known_intermittents += data["known_intermittents"]
-
+            unexpected += json.load(file)
     unexpected.sort(key=lambda result: result["path"])
-    known_intermittents.sort(key=lambda result: result["path"])
+
+    def is_flaky(result):
+        return result["flaky"]
+
+    def is_stable_and_known(result):
+        return not is_flaky(result) and result["issues"]
+
+    def is_stable_and_unexpected(result):
+        return not is_flaky(result) and not result["issues"]
+
+    def add_children(children, results, filter_func, text):
+        filtered = [Item.from_result(result) for result in
+                    filter(filter_func, results)]
+        if filtered:
+            children.append(Item(f"{text} ({len(filtered)})", "", filtered))
 
     children = []
-    if unexpected:
-        children.append(
-            Item(f"Tests producing unexpected results ({len(unexpected)})", "",
-                 [Item.from_result(result) for result in unexpected]),
-        )
-    if known_intermittents:
-        children.append(
-            Item("Unexpected results that are known to be intermittent "
-                 f"({len(known_intermittents)})", "",
-                 [Item.from_result(result) for result in known_intermittents])
-        )
+    add_children(children, unexpected, is_flaky, "Flaky unexpected result")
+    add_children(children, unexpected, is_stable_and_known,
+                 "Stable unexpected results that are known to be intermittent")
+    add_children(children, unexpected, is_stable_and_unexpected,
+                 "Stable unexpected results")
 
     run_url = get_github_run_url()
     if run_url:
--- a/tests/wpt/servowpt.py
+++ b/tests/wpt/servowpt.py
@@ -16,7 +16,7 @@ import mozlog
 import mozlog.formatters
 import multiprocessing
 
-from typing import List, NamedTuple, Optional, Tuple, Union
+from typing import List, NamedTuple, Optional, Union
 from grouping_formatter import UnexpectedResult, UnexpectedSubtestResult
 
 SCRIPT_PATH = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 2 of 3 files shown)


jdm:
I'm so excited for this change!

etc/ci/report_aggregated_expected_results.py
```diff
@@ -97,28 +97,32 @@ def to_html(self, level: int = 0) -> ElementTree.Element:
...
-    known_intermittents.sort(key=lambda result: result["path"])
+
+    def is_flaky(result):
+        return result["flaky"]
+
+    def is_stable_and_known(result):
+        return not is_flaky(result) and result["issues"]
+
+    def is_unexpected_stable(result):
+        return not is_flaky(result) and not is_stable_and_known(result)
```
delan, line 112: the logic behind this wasn’t obvious to me, since it expands to (not flaky and not (not flaky and has issues)), which is (not flaky and (flaky or no issues)), which smells contradictory (despite being correct). i wonder if this would be easier to understand if expressed as

```python
def is_stable_and_unexpected(result):
    return not is_flaky(result) and not result["issues"]
```

tests/wpt/servowpt.py
```diff
@@ -154,10 +154,23 @@ def run_tests(**kwargs):
...
+        unexpected_results = list(handler.unexpected_results)
+
+        print(80 * "=")
+        print(f"Rerunning {len(unexpected_results)} tests "
+              "with unexpected results to detect flaky tests.")
+        unexpected_results_tests = [result.path for result in unexpected_results]
+        kwargs["test_list"] = unexpected_results_tests
+        kwargs["include"] = unexpected_results_tests
+        kwargs["pause_after_test"] = False
+        wptrunner.run_tests(**kwargs)
```
delan, line 166: just to confirm, is it ok that we’re reusing the handler instance? it looks ok, since ServoHandler.suite_start happens to call reset_state, but it might be worth calling that explicitly here to avoid future bugs

tests/wpt/servowpt.py
```diff
@@ -280,49 +290,40 @@ def filter_unexpected_results(
...
+    dashboard = TrackerDashboardFilter()
+    dashboard.report_failures(unexpected_results)
+
+    def add_result(output, text, results, filter_func) -> None:
+        filtered = [str(result) for result in filter(filter_func, results)]
+        if filtered:
+            output += [f"{text} ({len(results)}): ", *filtered]
+
+    def is_unexpected_stable(result):
+        return not result.flaky and not result.issues
```
delan, line 312: if you end up renaming the other one, don’t forget this one

tests/wpt/servowpt.py
```diff
@@ -280,49 +290,40 @@ def filter_unexpected_results(
...
+
+    output = []
+    add_result(output, "Flaky unexpected results:", unexpected_results,
```
delan, line 315: nit: this section has a colon, but the other two don’t

tests/wpt/servowpt.py
```diff
@@ -154,10 +154,23 @@ def run_tests(**kwargs):
 
     # Filter intermittents if that was specified on the command-line.
     if handler.unexpected_results and filter_intermittents_output:
-        all_filtered = filter_intermittents(
-            handler.unexpected_results,
-            filter_intermittents_output
-        )
+        unexpected_results = list(handler.unexpected_results)
```
delan, line 157: suggest commenting

```python
        # Start with the unexpected results from the first run. Some results may
        # be marked as flaky after rerunning, but the list won’t change.
```

tests/wpt/servowpt.py
```diff
@@ -154,10 +154,23 @@ def run_tests(**kwargs):
...
+              "with unexpected results to detect flaky tests.")
+        unexpected_results_tests = [result.path for result in unexpected_results]
+        kwargs["test_list"] = unexpected_results_tests
+        kwargs["include"] = unexpected_results_tests
+        kwargs["pause_after_test"] = False
+        wptrunner.run_tests(**kwargs)
+
+        stable_tests = [result.path for result in handler.unexpected_results]
+        for result in unexpected_results:
+            result.flaky = result.path not in stable_tests
```
delan, line 170: suggest commenting

```python
        # Mark any tests that yielded expected results when rerun as flaky, but
        # otherwise disregard the results of the second run.
```

delan (commented):
Looks good, with some minor comments :)

etc/ci/report_aggregated_expected_results.py
mrobinson: That makes sense to me. I've done this rename.

tests/wpt/servowpt.py
mrobinson: Yes, that's correct. I've added a call to `reset_state()` here and a comment explaining it.

tests/wpt/servowpt.py
mrobinson: I've expanded your proposed comment a bit:

```
+        # Copy the list of unexpected results from the first run, so that we
+        # can access them after the tests are rerun (which changes
+        # `handler.unexpected_results`). After rerunning some tests will be
+        # marked as flaky but otherwise the contents of this original list
+        # won't change.
```

tests/wpt/servowpt.py
mrobinson: I've added a similar comment which also points to a future behavior which might be good as well.

```
+        # Use the second run to mark tests from the first run as flaky, but
+        # discard the results otherwise.
+        # TODO: It might be a good idea to send the new results to the
+        # dashboard if they were also unexpected.
```

mrobinson:
@delan Thanks for the review! I've uploaded a new version which addresses your comments.

delan (commented):
Thanks! One last thing:

etc/ci/report_aggregated_expected_results.py
```diff
@@ -97,28 +97,32 @@ def to_html(self, level: int = 0) -> ElementTree.Element:
...
-    known_intermittents.sort(key=lambda result: result["path"])
+
+    def is_flaky(result):
+        return result["flaky"]
+
+    def is_stable_and_known(result):
+        return not is_flaky(result) and result["issues"]
+
+    def is_stable_and_unexpected(result):
+        return not is_flaky(result) and not is_stable_and_known(result)
```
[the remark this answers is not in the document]
mrobinson: Oh, apologies. This should be fixed now.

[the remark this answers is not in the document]
delan: no worries!

PR #29375, opened 2023-02-17
Warning instead of exception on unsupported Ubuntu

sagudev:
For someone who is always using latest stable Ubuntu version it is pretty annoying to change the script every time, so issuing a warning instead of error would still allow users on unsupported version to continue bootstrapping.

---

- [x] These changes do not require tests because changes are in bootstrap script


jdm (approved):

sagudev:
> Results from try job ([#4204089803](https://github.com/servo/servo/actions/runs/4204089803)):
> Unexpected results that are known to be intermittent (39)

It would be more stylish to create results from `@bors-servo` instead of github-actions bot, but that is just my nitpicking.

PR #29383, opened 2023-02-19
 Do not run specialized workflow on forks

### LINKED ISSUE #29379: CI in forks
Currently if you enable CI on fork you get handful of scheduled builds and builds that are only working on main repo, that should be disabled on forks:
- [docs](https://github.com/servo/servo/blob/master/.github/workflows/docs.yml)
- [nightly-rust](https://github.com/servo/servo/blob/master/.github/workflows/nightly-rust.yml)
- [nightly](https://github.com/servo/servo/blob/master/.github/workflows/nightly.yml) (main builds creates own artifacts)
- [wpt-nightly](https://github.com/servo/servo/blob/master/.github/workflows/wpt-nightly.yml)
- [upstream-nightly-changes](https://github.com/servo/servo/blob/master/.github/workflows/upstream-wpt-changes.yml)

Most work could be done with simple if guards: `if: ${{ github.repository == "servo/servo" }}`.

As for the [main](https://github.com/servo/servo/blob/master/.github/workflows/main.yml) I am currently using `git push origin local-branch:try --force` to replicate `@bors-servo try` command. The easiest option would be to change branch on which the job is run (all branches). I believe this would be beneficial also to internal contributors (those who have permissions to run `@bors-servo try`) as they could fully test their changes in personal forks and thus having less try-fail CI runs on main repo (remember: whole servo organization shares a limited number of parallel GitHub Actions workers).

OFF-TOPIC IDEA: Make servo/servo CI try/auto runs on `@bors-servo` so all other servo org repos aren't burdened by servo/servo CI runs. The only problem here may be check reports back in main repo.

OFF-TOPIC IDEA: Provide artifacts on PRs.

sagudev:
Per #29379 to make GitHub actions more fork-friendly.

I have not decided yet on how to handle main workflow, as I started to like my `git push origin HEAD:try --force` workflow, but using CI on all forks branches would also help enforce the [rule](https://github.com/servo/servo/blob/master/CONTRIBUTING.md#pull-request-checklist):
> Commits should be as small as possible, while ensuring that each commit is correct independently (i.e., each commit should compile and pass tests).

---

- [x] These changes partially fix #29379 

- [X] These changes do not require tests because it is only CI stuff

### DIFF
--- a/.github/workflows/nightly-rust.yml
+++ b/.github/workflows/nightly-rust.yml
@@ -12,6 +12,8 @@ env:
 
 jobs:
   build-linux-with-rust-nightly:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Build (Linux) + rustc nightly
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/upstream-wpt-changes.yml
+++ b/.github/workflows/upstream-wpt-changes.yml
@@ -5,6 +5,8 @@ on:
 
 jobs:
   upstream:
+    # Run job only on servo/servo
+    if: github.repository == 'servo/servo'
     runs-on: ubuntu-latest
     steps:
       - name: Calculate PR fetch depth
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,8 @@ on:
 
 jobs:
   upload-docs:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Upload docs to GitHub Pages
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,6 +12,8 @@ env:
 
 jobs:
   upload-linux:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Upload nightly (Linux)
     runs-on: ubuntu-20.04
     steps:
@@ -33,6 +35,8 @@ jobs:
           S3_UPLOAD_CREDENTIALS: ${{ secrets.S3_UPLOAD_CREDENTIALS }}
 
   upload-mac:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Upload nightly (macOS)
     runs-on: macos-12
     steps:
@@ -74,6 +78,8 @@ jobs:
           GITHUB_HOMEBREW_TOKEN: ${{ secrets.HOMEBREW_TOKEN }}
 
   upload-win:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Upload nightly (Windows)
     runs-on: windows-2019
     steps:
--- a/.github/workflows/wpt-nightly.yml
+++ b/.github/workflows/wpt-nightly.yml
@@ -10,6 +10,8 @@ on:
 
 jobs:
   build-linux:
+    # This job is only useful when run on upstream servo.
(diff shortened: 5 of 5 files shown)


.github/workflows/nightly-rust.yml
```diff
@@ -12,6 +12,8 @@ env:
 
 jobs:
   build-linux-with-rust-nightly:
+    # Run job only on servo/servo
+    if: github.repository == 'servo/servo'
```
mukilan, line 16: Should we have some easy way to deliberately allow these jobs on forks, e.g testing changes to the workflows? Maybe an additional check on presence of variable like 'RUN_ON_FORK'.

sagudev: While testing changes you can just temporary comment out those if guards, I believe that additional variables would make whole thing just more bloated and you would still need to change the value of variable, so changes are inevitable.

mukilan: The variable can be [defined at the repo level](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository), without requiring changes to the guard. It can make things easier when creating PRs, but I have no strong opinion on this. I'll let other reviewers make a call.

sagudev: I completely forget that variables at repo level exist. But the problem is when you set it, it will apply for all workflows. This is not very hard problem, but it still can cause inconveniences.

Currently the workflows are blocked even if you manually dispatch it, so maybe the best compromise would be to allow workflows to be run when manually dispatched.

mukilan: > But the problem is when you set it, it will apply for all workflows

Will requiring the variable to specify the name of the workflow help? i.e The guard can be 
```
if: github.repository == 'servo/servo' || env.RUN_ON_FORK == github.workflow
```

sagudev: Hm, sure that could work, but I do not know if it is worth the bloat as workflow files are not changed often and someone may quickly forget about `RUN_ON_FORK` and what it does. Manual dispatch would be easier to work with it, more self-describing and what someone would expect to just-work™. The guard would be:
```
if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
```

.github/workflows/upstream-wpt-changes.yml
```diff
@@ -5,6 +5,8 @@ on:
 
 jobs:
   upstream:
+    # Run job only on servo/servo
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
```
mukilan, line 9: This workflow and test-upstream-wpt-changes are executed in the context of a PR (i.e triggered by pull_request* events), so we can't trigger them manually. The condition on 'workflow_dispatch'  is not going to be useful for these.

sagudev: Good catch. Maybe guard like: `if: github.repository == 'servo/servo' || contains('CI', github.event.pull_request.title)`.

What/How exactly is `upstream-wpt-changes` workflows doing @mrobinson?

delan: i answered in #29379 :)

mukilan (approved):
r+ from me.  @delan or @mrobinson can either of you also please review this?

.github/workflows/test-upstream-wpt-changes.yml
```diff
@@ -6,6 +6,8 @@ on:
 
 jobs:
   test:
+    # Run job only on servo/servo
+    if: github.repository == 'servo/servo'
```
mrobinson, line 10: I would say that it does make sense to run this particular action on forks, because it would allow someone to verify that the changes they made to this script actually pass CI.

mrobinson (changes requested):
This seems good to me apart from a couple suggestions.

The first is that I'm not sure how useful it is to repeat the "Run job only on servo/servo" comment as it simply repeats the logic of the line. Instead maybe say something like, "This job is only useful when run on upstream servo." or simply omit the comment.

mrobinson:
@bors-servo r+

Thank you!

mrobinson (approved):

Issue #28381, opened 2021-04-21
immeta dependency no longer maintained


atouchet:
https://github.com/netvl/immeta/commit/f2b2dd55322fb12b9f3dd9b3d6a28122650b19a2

This is a dependency of Servo's net crate. It should probably be removed, forked, or replaced with something else.

amy-behan:
I only found a single usage for immeta.

https://github.com/servo/servo/blob/master/components/net/image_cache.rs#L621-L631

It would be possible to implement this functionality in servo similar to how its done in immeta we only use the dimensions as far as i can tell. 

Jpeg example:
https://github.com/netvl/immeta/blob/master/src/formats/jpeg.rs#L108

Let me know if this sounds like a good solution.

jdm:
I don't see value in moving that functionality into servo - we already depend on the crate, and it won't disappear. Having the code inside servo implies that servo developers understand it and maintain it, which wouldn't be true. If we want to replace immeta with another crate that's not abandoned, that's fine with me.

InfRandomness:
Is this still calling for a PR ?

atouchet:
This was forked in #29421 but now Servo is relying on another Git dependency. What would be the best way to proceed here? Should we add the fork to the Servo org?

jdm:
I have forked the fork to https://github.com/servo/immeta

atouchet:
This was fixed in #30294.

PR #29421, opened 2023-02-25
Switch to immeta fork to remove arrayvec dupe

fabricedesre:
The original immeta crate is archived (https://github.com/netvl/immeta) so I forked it, updated it to newer crates and edition 2021. No public api change.

---

- [X] These changes do not require tests because just fighting dupes


atouchet:
Related: #28381

If we decide to fork the crate it would be nice to publish it so that we aren't depending on another Git dependency.

fabricedesre:
Sure, we could move it to the servo org, rename it and publish it.

PR #29423, opened 2023-02-26
General quick check for PRs and forks.

sagudev:
Pull request workflow was changed to more general Quick check workflow that is meant to be run on PR and all branches not covered by main workflow.

BONUS: I used [xvfb-run](https://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html) to enable smoketest on Linux.

---

- [x] These changes are partial fix for #29379 

- [X] These changes do not require tests because there are only CI changes

### DIFF
--- a/.github/workflows/quick-check.yml
+++ b/.github/workflows/quick-check.yml
@@ -26,6 +26,8 @@ jobs:
         run: python3 ./mach test-tidy --no-progress --all
       - name: Release build
         run: python3 ./mach build --release
+      - name: Smoketest
+        run: xvfb-run python3 ./mach smoketest
       - name: Unit tests
         run: python3 ./mach test-unit --release
       - name: Lockfile check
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Thank you! This is a nice improvement.

.github/workflows/quick-check.yml
```diff
@@ -1,14 +1,17 @@
-name: Pull request
+name: Quick check
 on:
   pull_request:
-    branches: ["**"]
+  push:
```
mukilan, line 4: I think this change is now causing the quick-check.yml workflow to execute twice for every PR, once for the push and one for the PR event. Example:

https://github.com/servo/servo/pull/29452

sagudev: Oh, yes indeed. Master branch should be excluded, as in mozjs repo.

sagudev: and all depedabot branches.

PR #29424, opened 2023-02-27
Ensure winit windows are cleaned up before TLS disappears.

jdm:
Since the winit update, closing Servo on macOS can lead to a crash. This is because winit panics while trying to clean up the window, since it requires the TLS to be available. Right now this code doesn't run until Servo's TLS is being destroyed. There's no reason to use a thread local value to store the window information, however, and moving it into the App object avoids the panic.

---
- [x] These changes do not require tests because we don't run non-headless tests on macOS.

### DIFF
--- a/ports/winit/app.rs
+++ b/ports/winit/app.rs
@@ -23,15 +23,12 @@ use std::mem;
 use std::rc::Rc;
 use webxr::glwindow::GlWindowDiscovery;
 
-thread_local! {
-    pub static WINDOWS: RefCell<HashMap<WindowId, Rc<dyn WindowPortsMethods>>> = RefCell::new(HashMap::new());
-}
-
 pub struct App {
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

mrobinson:
This seems like a good fix and a nice cleanup.

CYBAI:
build failure on windows, linux and macos 👀 

```
error[E0614]: type `std::collections::HashMap<WindowId, Rc<(dyn WindowPortsMethods + 'static)>>` cannot be dereferenced
   --> ports/winit/app.rs:196:35
    |
196 |         for (_win_id, window) in &*self.windows {
    |                                   ^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0614`.
```

jdm:
@bors-servo r+
Missed committing a local change.

PR #29443, opened 2023-02-28
Improve winit/mach segfault output

### LINKED ISSUE #29442: winit/mach segfault output can be confusing
When `mach run` runs the winit port on unix, it waits for the process to terminate and [prints a “return value”](https://github.com/servo/servo/blob/8decd02a06d39836bcc859ad465f281586eb9043/python/servo/post_build_commands.py#L172), which is positive if the process exited normally and negative if it was terminated abnormally by an uncaught signal (see [Popen.returncode](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode), [wait(2)](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html), [signal.h](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html)). But for some signals, including segfaults, our [crash handler](https://github.com/servo/servo/blob/6bd6d8db86869898ac6b9453d05d1efc017cb5e3/ports/winit/crash_handler.rs) catches them and turns them into normal exits with the signal number as the status.

This means segfaults will generally look like `Servo exited with return value 11` and not `Servo exited with return value -11`, or even completely silent if not using mach, which can be misleading if there’s no other indication that a segfault occurred, because only _uncaught_ segfaults dump core or get reported by the shell as “Segmentation fault” or get reported in syslog.

The good news is that there _is_ such an indication, because the crash handler tries to print some details before exiting, including the thread name and stack trace. At the very least, we should print “Stack trace”, even if there’s no thread name or backtrace output. You can try this yourself by running servo with the following patch:

```diff
diff --git a/ports/winit/main2.rs b/ports/winit/main2.rs
index 63560f832e..59d7bf2378 100644
--- a/ports/winit/main2.rs
+++ b/ports/winit/main2.rs
@@ -158,6 +158,7 @@ pub fn main() {
 
     let user_agent = opts_matches.opt_str("u");
 
+    unsafe { std::ptr::read::<usize>(std::ptr::null()); }
     App::run(do_not_use_native_titlebar, device_pixels_per_px, user_agent);
 
     platform::deinit(clean_shutdown)
```

```
$ ./mach build -d
$ ./mach run -d about:blank
Stack trace for thread "main"
   0: servo::backtrace::print
             at /home/delan/code/servo.master/ports/winit/backtrace.rs:17:5
   1: servo::crash_handler::install::handler
             at /home/delan/code/servo.master/ports/winit/crash_handler.rs:25:21
   2: <unknown>
   3: core::ptr::read
             at /rustc/dc1d9d50fba2f6a1ccab8748a0050cde38253f60/library/core/src/ptr/mod.rs:1157:34
   4: servo::main
             at /home/delan/code/servo.master/ports/winit/main2.rs:161:14
[...]
Servo exited with return value 11
```

The bad news is that these prints don’t work if the segfault occurs on layout or script threads. You can try this yourself by running servo with the following patch:

```diff
diff --git a/components/layout_thread/lib.rs b/components/layout_thread/lib.rs
index ac1a7fe21e..77c4c9d087 100644
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -589,6 +589,7 @@ impl LayoutThread {
             rw_data: &rw_data,
             possibly_locked_rw_data: &mut possibly_locked_rw_data,
         };
+        unsafe { std::ptr::read::<usize>(std::ptr::null()); }
         while self.handle_request(&mut rw_data) {
             // Loop indefinitely.
         }
```

```
$ ./mach build -d
$ ./mach run -d about:blank
Servo exited with return value 11
```

Playing around with the crash handler yields some interesting results:

* removing the stdout lock line doesn’t help
* flushing stdout after the prints doesn’t help
* switching to stderr _does_ fix the problem
    * but adding an unused call to std::io::stdout() breaks it again

The key difference between [std::io::stderr](https://doc.rust-lang.org/std/io/fn.stderr.html) and [std::io::stdout](https://doc.rust-lang.org/std/io/fn.stdout.html) here is that the former just returns a wrapper around a static reference, while the latter uses [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) to defer the cost of constructing its LineWriter until first called.

Debugging the second example, we can see that we segfault on the null pointer access, but we segfault again in the signal handler while trying to get a handle to stdout. This in turn calls our signal handler again, but we’ve already set BEEN_HERE_BEFORE, so we just exit without printing anything.

```
$ ./mach run -d --debug -- -y 1 about:blank
(gdb) run
Thread 104 "Layout(1,1)" received signal SIGSEGV, Segmentation fault.
0x000055555c9e1d31 in core::ptr::read<usize> (src=0x0) at /rustc/dc1d9d50fba2f6a1ccab8748a0050cde38253f60/library/core/src/ptr/mod.rs:1157

(gdb) continue
Thread 104 "Layout(1,1)" received signal SIGSEGV, Segmentation fault.
0x000055555b814278 in _rjem_mallocx (size=<error reading variable: Cannot access memory at address 0x7fff3c054408>, flags=<error reading variable: Cannot access memory at address 0x7fff3c054404>) at ../jemalloc/src/jemalloc.c:2549
2549	je_mallocx(size_t size, int flags) {

(gdb) where
#0  0x000055555b814278 in _rjem_mallocx (size=0, flags=0) at ../jemalloc/src/jemalloc.c:2549
#1  0x000055555b7e2e29 in servo_allocator::platform::{impl#0}::alloc (self=0x55555ee3c5ac, layout=...) at components/allocator/lib.rs:72
#2  0x000055555b7e2bc2 in servo_allocator::_::__rg_alloc (arg0=1024, arg1=1) at components/allocator/lib.rs:8
#3  0x0000555555fa3f16 in alloc::alloc::alloc () at library/alloc/src/alloc.rs:95
#4  alloc::alloc::Global::alloc_impl () at library/alloc/src/alloc.rs:177
#5  alloc::alloc::{impl#1}::allocate () at library/alloc/src/alloc.rs:237
#6  alloc::raw_vec::RawVec::allocate_in<u8, alloc::alloc::Global> () at library/alloc/src/raw_vec.rs:185
#7  alloc::raw_vec::RawVec::with_capacity_in<u8, alloc::alloc::Global> () at library/alloc/src/raw_vec.rs:131
#8  alloc::vec::Vec::with_capacity_in<u8, alloc::alloc::Global> () at library/alloc/src/vec/mod.rs:676
#9  alloc::vec::Vec::with_capacity<u8> () at library/alloc/src/vec/mod.rs:483
#10 std::io::buffered::bufwriter::BufWriter::with_capacity<std::io::stdio::StdoutRaw> () at library/std/src/io/buffered/bufwriter.rs:115
#11 std::io::buffered::linewriter::LineWriter::with_capacity<std::io::stdio::StdoutRaw> () at library/std/src/io/buffered/linewriter.rs:109
#12 std::io::buffered::linewriter::LineWriter::new<std::io::stdio::StdoutRaw> () at library/std/src/io/buffered/linewriter.rs:89
#13 std::io::stdio::stdout::{closure#0} () at library/std/src/io/stdio.rs:606
#14 std::sync::once_lock::{impl#0}::get_or_init::{closure#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}> () at library/std/src/sync/once_lock.rs:177
#15 std::sync::once_lock::{impl#0}::initialize::{closure#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !> ()
    at library/std/src/sync/once_lock.rs:307
#16 std::sync::once::{impl#2}::call_once_force::{closure#0}<std::sync::once_lock::{impl#0}::initialize::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !>> () at library/std/src/sync/once.rs:202
#17 std::sys_common::once::futex::Once::call<std::sync::once::{impl#2}::call_once_force::{closure_env#0}<std::sync::once_lock::{impl#0}::initialize::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffer--Type <RET> for more, q to quit, c to continue without paging--
ed::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !>>> ()
    at library/std/src/sys_common/once/futex.rs:113
#18 0x0000555555fa290e in std::sync::once::Once::call_once_force<std::sync::once_lock::{impl#0}::initialize::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !>> () at library/std/src/sync/once.rs:202
#19 std::sync::once_lock::OnceLock::initialize<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !> ()
    at library/std/src/sync/once_lock.rs:306
#20 0x000055555d555efa in std::sync::once_lock::OnceLock::get_or_try_init<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::sync::once_lock::{impl#0}::get_or_init::{closure_env#0}<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}>, !> () at library/std/src/sync/once_lock.rs:225
#21 std::sync::once_lock::OnceLock::get_or_init<std::sync::remutex::ReentrantMutex<core::cell::RefCell<std::io::buffered::linewriter::LineWriter<std::io::stdio::StdoutRaw>>>, std::io::stdio::stdout::{closure_env#0}> () at library/std/src/sync/once_lock.rs:177
#22 std::io::stdio::stdout () at library/std/src/io/stdio.rs:606
#23 0x0000555556177aa4 in servo::crash_handler::install::handler (sig=11) at ports/winit/crash_handler.rs:18
#24 0x00005555596cce6d in WasmTrapHandler (signum=11, info=0x7fff3c1085b0, context=<optimized out>)
    at /home/delan/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/d2a3526/mozjs/mozjs/js/src/wasm/WasmSignalHandlers.cpp:790
#25 <signal handler called>
#26 0x000055555c9e1d31 in core::ptr::read<usize> (src=0x0)
    at /rustc/dc1d9d50fba2f6a1ccab8748a0050cde38253f60/library/core/src/ptr/mod.rs:1157
#27 0x00005555568c195f in layout_thread::LayoutThread::start (self=...) at components/layout_thread/lib.rs:592
[...]
```

We can fix these problems by making the handler use std::io::stderr(), or if we really need to print to stdout, calling std::io::stdout() at the top of main() in [main2.rs](https://github.com/servo/servo/blob/0048bf82daa5ec985256b6be92e4302ae5ce235e/ports/winit/main2.rs). We should also consider reraising the signal to terminate abnormally, rather than exiting normally.

delan:
This patch improves the output that you see when the winit port segfaults, especially when the segfault happens on a layout or script thread, by making two changes to our crash handler and one change to mach.

* we make the crash handler use stderr instead of stdout, because [std::io::stdout](https://doc.rust-lang.org/std/io/fn.stdout.html) allocates when first called, which will often cause a second segfault
* we make the crash handler reraise the signal to terminate abnormally, allowing mach to distinguish it from exiting normally with the same status code as a signal number
* we make mach print different messages for whether the process terminated by signal or exited normally

You can try this yourself by running servo with the following patch:

```diff
diff --git a/components/layout_thread/lib.rs b/components/layout_thread/lib.rs
index ac1a7fe21e..77c4c9d087 100644
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -589,6 +589,7 @@ impl LayoutThread {
             rw_data: &rw_data,
             possibly_locked_rw_data: &mut possibly_locked_rw_data,
         };
+        unsafe { std::ptr::read::<usize>(std::ptr::null()); }
         while self.handle_request(&mut rw_data) {
             // Loop indefinitely.
         }
```

Before:

```
$ ./mach run -d about:blank
[should print “Stack trace for "Layout(1,1)"” here, but segfaults]
Servo exited with return value 11
```

After:

```
$ ./mach run -d about:blank
Caught signal 11 in thread "Layout(1,1)"
[should ideally print backtrace here, but segfaults]
Servo was terminated by signal 11
```

---

- [x] These changes fix #29442

- [x] These changes do not require tests because it would be impractical to test them

(r? @jdm, @SimonSapin based on git log for crash handler)

### DIFF
--- a/ports/winit/crash_handler.rs
+++ b/ports/winit/crash_handler.rs
@@ -8,24 +8,46 @@ pub fn install() {}
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 pub fn install() {
     use crate::backtrace;
-    use libc::_exit;
     use sig::ffi::Sig;
     use std::{io::Write, sync::atomic, thread};
 
     extern "C" fn handler(sig: i32) {
+        // Only print crash message and backtrace the first time, to avoid
+        // infinite recursion if the printing causes another signal.
         static BEEN_HERE_BEFORE: atomic::AtomicBool = atomic::AtomicBool::new(false);
         if !BEEN_HERE_BEFORE.swap(true, atomic::Ordering::SeqCst) {
-            let stdout = std::io::stdout();
-            let mut stdout = stdout.lock();
-            let _ = write!(&mut stdout, "Stack trace");
+            // stderr is unbuffered, so we won’t lose output if we crash later
+            // in this handler, and the std::io::stderr() call never allocates.
+            // std::io::stdout() allocates the first time it’s called, which in
+            // practice will often segfault (see below).
+            let stderr = std::io::stderr();
+            let mut stderr = stderr.lock();
+            let _ = write!(&mut stderr, "Caught signal {sig}");
             if let Some(name) = thread::current().name() {
-                let _ = write!(&mut stdout, " for thread \"{}\"", name);
+                let _ = write!(&mut stderr, " in thread \"{}\"", name);
             }
-            let _ = write!(&mut stdout, "\n");
-            let _ = backtrace::print(&mut stdout);
+            let _ = write!(&mut stderr, "\n");
+
+            // This call always allocates, which in practice will segfault if
+            // we’re handling a non-main-thread (e.g. layout) segfault. Strictly
+            // speaking in POSIX terms, this is also undefined behaviour.
+            let _ = backtrace::print(&mut stderr);
         }
+
+        // Outside the BEEN_HERE_BEFORE check, we must only call functions we
+        // know to be “async-signal-safe”, which includes sigaction(), raise(),
+        // and _exit(), but generally doesn’t include anything that allocates.
+        // https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_04_03_03
         unsafe {
-            _exit(sig);
+            // Reset the signal to the default action, and reraise the signal.
+            // Unlike libc::_exit(sig), which terminates the process normally,
+            // this terminates abnormally just like an uncaught signal, allowing
+            // mach (or your shell) to distinguish it from an ordinary exit, and
+            // allows your kernel to make a core dump if configured to do so.
+            let mut action: libc::sigaction = std::mem::zeroed();
+            action.sa_sigaction = libc::SIG_DFL;
+            libc::sigaction(sig, &action, std::ptr::null_mut());
+            libc::raise(sig);
         }
     }
 
--- a/python/servo/post_build_commands.py
+++ b/python/servo/post_build_commands.py
@@ -169,7 +169,10 @@ class PostBuildCommands(CommandBase):
         try:
             check_call(args, env=env)
         except subprocess.CalledProcessError as e:
-            print("Servo exited with return value %d" % e.returncode)
+            if e.returncode < 0:
+                print(f"Servo was terminated by signal {-e.returncode}")
+            else:
+                print(f"Servo exited with non-zero status {e.returncode}")
             return e.returncode
         except OSError as e:
             if e.errno == 2:


jdm (approved):
Nice! Thanks for the analysis of what's going wrong here.

jdm:
Whoops, that's just habit at this point.

PR #29465, opened 2023-03-05
Implement Headers.prototype.getSetCookie

### LINKED ISSUE #29388: Implement `Headers.prototype.getSetCookie`
Spec PR: https://github.com/whatwg/fetch/pull/1346
Test PR: https://github.com/web-platform-tests/wpt/pull/31442

yoseio:
---

- [x] These changes fix #29388 

### DIFF
--- a/components/net/fetch/headers.rs
+++ b/components/net/fetch/headers.rs
@@ -26,49 +26,53 @@ fn get_header_value_as_list(name: &str, headers: &HeaderMap) -> Option<Vec<Strin
         return c != '\u{0022}' && c != '\u{002C}';
     }
 
+    // https://fetch.spec.whatwg.org/#header-value-get-decode-and-split
     // Step 1
     let initial_value = get_value_from_header_list(name, headers);
 
     if let Some(input) = initial_value {
-        // Step 4
-        let mut position = input.chars().peekable();
+        // Step 1
+        let input = input.into_iter().map(|u| char::from(u)).collect::<String>();
 
-        // Step 5
+        // Step 2
+        let mut position = s.chars().peekable();
+
+        // Step 3
         let mut values: Vec<String> = vec![];
 
-        // Step 6
+        // Step 4
         let mut value = String::new();
 
-        // Step 7
+        // Step 5
         while position.peek().is_some() {
-            // Step 7.1
+            // Step 5.1
             value += &*collect_sequence(&mut position, char_is_not_quote_or_comma);
 
-            // Step 7.2
+            // Step 5.2
             if let Some(&ch) = position.peek() {
                 if ch == '\u{0022}' {
-                    // Step 7.2.1.1
+                    // Step 5.2.1.1
                     value += &*collect_http_quoted_string(&mut position, false);
 
-                    // Step 7.2.1.2
+                    // Step 5.2.1.2
                     if position.peek().is_some() {
                         continue;
                     }
                 } else {
                     // ch == '\u{002C}'
 
-                    // Step 7.2.2.2
+                    // Step 5.2.2.2
                     position.next();
                 }
             }
 
-            // Step 7.3
+            // Step 5.3
             value = value.trim_matches(HTTP_TAB_OR_SPACE).to_string();
 
-            // Step 7.4
+            // Step 5.4
             values.push(value);
 
-            // Step 7.5
+            // Step 5.5
             value = String::new();
         }
 
--- a/components/net/fetch/headers.rs
+++ b/components/net/fetch/headers.rs
@@ -157,7 +157,7 @@ fn collect_http_quoted_string(position: &mut Peekable<Chars>, extract_value: boo
 }
 
 /// <https://fetch.spec.whatwg.org/#concept-header-list-get>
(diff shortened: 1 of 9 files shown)


tests/wpt/metadata/fetch/api/headers/header-setcookie.any.js.ini
```diff
@@ -23,33 +23,15 @@
...
-  [Headers.prototype.getSetCookie with one header created from an object]
-    expected: FAIL
-
   [Headers.prototype.getSetCookie with multiple headers]
     expected: FAIL
 
-  [Headers.prototype.getSetCookie with an empty header]
-    expected: FAIL
-
   [Headers.prototype.getSetCookie with two equal headers]
```
mrobinson, line 29: Is there a reason these tests are still failing? Ideally, this would be annotated by a bug or made to function properly in this change.

2shiori17: I think it's because the [sort-and-combine function](https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine) is not implemented properly and I'm currently fixing it.

mrobinson: @2shiori17 Thanks for the insight here. Do you mind opening a bug for that, if it doesn't exist and adding a "bug: " annotation to the ini file? Here's [an example](https://searchfox.org/mozilla-central/source/testing/web-platform/meta/css/css-contain/content-visibility/content-visibility-064.html.ini) of how to do that.

2shiori17: whatwg/fetch#1346, mentioned in #29388, includes not only the addition of the getSetCookie function but also changes to the sort-and-combine function, so I would like to fix that in this pull request.

mrobinson: Okay. I'll wait to review this change until it's ready then.

components/script/dom/headers.rs
```diff
@@ -273,17 +270,25 @@ impl Headers {
...
-            header_vec.push(name_value);
+
+        for name in borrowed_header_list.keys() {
+            let name = name.as_str();
+            if name == "set-cookie" {
+                for value in borrowed_header_list.get_all(name).iter() {
+                    header_vec.push((name.to_owned(), value.as_bytes().to_vec()));
+                }
+            } else {
+                if let Some(value) = get_value_from_header_list(name, &borrowed_header_list) {
```
mrobinson, line 285: Could this be `else if` here and combined with the control statement above?

```rust
} else if let Some(value) = get_value_from_header_list(name, &borrowed_header_list) {
 ...
}
```

components/script/dom/headers.rs
```diff
@@ -63,57 +64,44 @@ impl Headers {
 impl HeadersMethods for Headers {
     // https://fetch.spec.whatwg.org/#concept-headers-append
     fn Append(&self, name: ByteString, value: ByteString) -> ErrorResult {
-        // Step 1
```
mrobinson, line 66: I think the "Step 1" etc comments can be quite helpful. It's unclear to me how some of the code here fits in with the steps described in https://fetch.spec.whatwg.org/#concept-headers-append.

mrobinson (changes requested):
Great work! I have a couple questions about the code here, but this is looking nice already.

components/net/fetch/headers.rs
```diff
@@ -157,7 +157,7 @@ fn collect_http_quoted_string(position: &mut Peekable<Chars>, extract_value: boo
 }
 
 /// <https://fetch.spec.whatwg.org/#concept-header-list-get>
-fn get_value_from_header_list(name: &str, headers: &HeaderMap) -> Option<String> {
+pub fn get_value_from_header_list(name: &str, headers: &HeaderMap) -> Option<String> {
```
jdm, line 160: We should move this into the poorly-named `net_traits` instead if we need to call it from other crates.

mrobinson (approved):
Thanks for the update. This looks good to me, but perhaps @jdm would like to take a quick look before we land this.

components/script/dom/headers.rs
```diff
@@ -65,55 +67,61 @@ impl HeadersMethods for Headers {
...
                 warn!(
                     "Servo thinks \"{:?}\" is a valid HTTP header value but HeaderValue doesn't.",
-                    combined_value
+                    valid_value
                 );
             },
         };
+
+        // Step 5
+        if self.guard.get() != Guard::RequestNoCors {
```
jdm, line 121: This step looks backwards - the spec says `If headers’s [guard](https://whatpr.org/fetch/1346.html#concept-headers-guard) is "request-no-cors", then [remove privileged no-CORS request-headers](https://whatpr.org/fetch/1346.html#concept-headers-remove-privileged-no-cors-request-headers) from headers.`

jdm (changes requested):
Looks good except for step 5!

2shiori17:
@jdm 
Could you please review this?

2shiori17:
@jdm
Could you please merge this PR?

PR #29490, opened 2023-03-13
Move hit testing information out of WebRender

mrobinson:
Store hit testing information in a data structure that sits alongside the display list in the compositor. This will allow the compositor to store more information per-node. The data structure also takes care of de-duplicating information between successive display list entries. In the future, the data structure can be even more aggressive in producing smaller side hit testing lists, if necessary.

---

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -28,16 +28,16 @@ use msg::constellation_msg::{
 };
 use net_traits::image::base::Image;
 use net_traits::image_cache::CorsStatus;
-use num_traits::FromPrimitive;
 #[cfg(feature = "gl")]
 use pixels::PixelFormat;
 use profile_traits::time::{self as profile_time, profile, ProfilerCategory};
+use script_traits::compositor::HitTestInfo;
 use script_traits::CompositorEvent::{MouseButtonEvent, MouseMoveEvent, TouchEvent, WheelEvent};
-use script_traits::{AnimationState, AnimationTickType, LayoutControlMsg};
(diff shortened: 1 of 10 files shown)


components/compositing/compositor.rs
```diff
@@ -99,6 +102,15 @@ impl FrameTreeId {
     }
 }
 
+#[derive(Debug)]
+struct CompositorHitTestResult {
+    pipeline_id: PipelineId,
+    point_in_viewport: euclid::default::Point2D<f32>,
+    point_relative_to_item: euclid::default::Point2D<f32>,
+    node: UntrustedNodeAddress,
+    cursor: Option<u16>,
```
mukilan, line 111: Do we need the raw u16 cursor value here to keep this generic (and not the result of Cursor::from_u8)? I'm not very familiar with this code, so it is not obvious to me.

mukilan (approved):

components/compositing/compositor.rs
mrobinson: You're totally correct. I've updated this change so that `Cursor` is used everywhere instead of converting the value to `u16`. This simplifies the code a bit.

mrobinson:
@bors-servo r=mukilan

Thanks for the review!

mrobinson:
@bors-servo r=mukilan

Fixing build for Layout 2013.

mrobinson:
@bors-servo r=mukilan

I had to make sure that the Compositor hit testing API also used the new hit testing list as well. After that it seems the tests are working now.

PR #29504, opened 2023-03-15
Add support for `<iframe>` in Layout 2020

mrobinson:
---

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -3106,13 +3106,18 @@ where
     }
 
     fn handle_iframe_size_msg(&mut self, iframe_sizes: Vec<IFrameSizeMsg>) {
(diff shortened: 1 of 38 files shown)


delan (approved):
Great work! The impl looks good, though it might be worth investigating why some of the siblings of the newly passing tests that also involve iframes are still failing:

* /css/CSS2/floats-clear/float-replaced-height-005.xht
* /css/CSS2/normal-flow/block-replaced-height-005.xht
* /css/CSS2/normal-flow/inline-block-replaced-height-005.xht
* /css/CSS2/normal-flow/inline-replaced-height-005.xht
* /css/CSS2/positioning/absolute-replaced-height-{005,007,012,014,019,021,026,028,033,035}.xht

delan:
Nice, thanks for looking into that!

@bors-servo r+

PR #29528, opened 2023-03-21
Run test-tidy when landing pull requests

mrobinson:
Currently `test-tidy` only runs during the quick check. This can mean that when a change is approved, the quick check can be skipped and a change can land without `test-tidy` ever being run on it. This change adds a `test-tidy` run to the main CI run to avoid this happening.

---


sagudev:
What if bors just fails if Quickcheck is failing, this way we would not do double-checking.

mrobinson:
Oh, in addition, running `test-tidy` before landing will also verify that no tidy errors are introduced in the process of merging, though admittedly that seems really unlikely.

sagudev:
Good point, but landing should still be prevented if Quickcheck fails.

The only difference between quickcheck and linux-build are smoketest (10s) and Unit tests (cca. 10 min) both present only in quickcheck while linux-build has packaging and uploading. I wonder if we could unify them.

jdm (approved):

PR #29431, opened 2023-02-27
Add try command to mach & try build partitioned

sagudev:
Adds `./mach try` command that enables anybody to easily test their changes without opening PR and requesting try from bors-servo, by force pushing HEAD to appropriate branch. Command accepts branches names to select only partial runs of CI (same like bors try command). So if you only want to test mac build (that would be `@bors-servo try=mac`) you run `./mach try mac`. If no job is specified, try branch is used.

As partitioned CI jobs were not working after migration to GitHub Actions I remade them by using if guards.
Also WPT jobs were failing due to empty `INTERMITTENT_TRACKER_DASHBOARD_SECRET` on my fork, so I added additional check to prevent failed run.

And that concludes my work on #29379 🎉 

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #29379 

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because it's CI

PR #29534, opened 2023-03-22
Abstract platform-specific workflows into separate files

sagudev:
Changes:
- split main workflow into multiple modular files & merged linux and quickcheck
- easily allow testing both `layout-2013` and `layout-2020` (currently layout-2020 is build on both quick-check and main workflows, but no actual tests are run)
- workflow_dispatch on any try run

Future work:
- sccache for caching
  - linux (release, layout-2013): 30min -> 15min
  - mac: 1h 30min -> 50min
  - windows: not working yet: 
```console
sccache: encountered fatal error
sccache: error: Invalid timestamp field in entry header ("-1          ")
sccache: caused by: Invalid timestamp field in entry header ("-1          ")
error: could not compile `openssl-sys`

Caused by:
  process didn't exit successfully: `sccache rustc --crate-name openssl_sys --edition=2018 C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\openssl-sys-0.9.81\src\lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C embed-bitcode=no -C metadata=15e5a7651b4bbdb6 -C extra-filename=-15e5a7651b4bbdb6 --out-dir C:\a\servo\servo\target\release\deps -C linker=lld-link.exe -L dependency=C:\a\servo\servo\target\release\deps --extern libc=C:\a\servo\servo\target\release\deps\liblibc-1fe4b5632b2d1333.rmeta --cap-lints allow -W unused-extern-crates -L native=C:\a\servo\servo\.servo\msvc-dependencies\openssl\111.3.0+1.1.1c-vs2017-2019-09-18\x64-windows\lib -l static=libssl -l static=libcrypto -l dylib=gdi32 -l dylib=user32 -l dylib=crypt32 -l dylib=ws2_32 -l dylib=advapi32 --cfg const_fn --cfg openssl --cfg "osslconf=\"OPENSSL_NO_COMP\"" --cfg "osslconf=\"OPENSSL_NO_SSL3_METHOD\"" --cfg ossl101 --cfg ossl102 --cfg ossl102f --cfg ossl102h --cfg ossl110 --cfg ossl110f --cfg ossl110g --cfg ossl110h --cfg ossl111 --cfg ossl111b --cfg ossl111c` (exit code: 0xfffffffe)
warning: build failed, waiting for other jobs to finish...
```
- report WPT test results also using GitHub checks (so we get those sexy results even when there is no associated PR) [example](https://github.com/sagudev/servo/runs/12193772513)
- run `WPT export` only when PR changes files in `tests/wpt/**`

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,170 @@
+name: Linux workflow
+on:
+  workflow_call:
+    inputs:
+      layout:
+        required: true
+        type: string
+      wpt:
+        required: true
+        type: boolean
+      unit-tests:
+        required: false
+        default: false
+        type: boolean
+  workflow_dispatch:
+    inputs:
+      layout:
+        required: true
+        type: choice
+        options: ["2013", "2020"]
+      wpt:
+        default: true
+        required: true
+        type: boolean
+      unit-tests:
+        required: false
+        default: false
+        type: boolean
+  push:
+    branches: ["try-linux", "try-wpt"]
+
+env:
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  LAYOUT: "${{ contains(inputs.layout, '2020') && 'layout-2020' || 'layout-2013' }}"
+
+jobs:
+  build-linux:
+    if: github.repository == 'servo/servo' || github.event_name == 'push'
+    name: Build
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: Bootstrap Python
+        run: python3 -m pip install --upgrade pip virtualenv
+      - name: Bootstrap dependencies
+        run: sudo apt update && python3 ./mach bootstrap
+      - name: Tidy
+        run: python3 ./mach test-tidy --no-progress --all
+      - name: Release build
+        run: python3 ./mach build --release --with-${{ env.LAYOUT }}
+      - name: Smoketest
+        run: xvfb-run python3 ./mach smoketest
+      - name: Unit tests
+        if: ${{ inputs.unit-tests }}
+        run: python3 ./mach test-unit --release --with-${{ env.LAYOUT }}
+      - name: Lockfile check
+        run: ./etc/ci/lockfile_changed.sh
+      - name: Package binary
+        run: tar -czf target.tar.gz target/release/servo resources
+      - name: Archive binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.LAYOUT }}-release-binary
+          path: target.tar.gz
+
+  linux-wpt:
+    if: ${{ github.ref_name == 'try-wpt' || inputs.wpt }}
+    name: Linux WPT Tests
+    runs-on: ubuntu-20.04
+    needs: ["build-linux"]
+    env:
+      max_chunk_id: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        chunk_id: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.LAYOUT }}-release-binary
+          path: release-binary
+      - name: unPackage binary
+        run: tar -xzf release-binary/target.tar.gz
+      - name: Prep test environment
+        run: |
+          python3 -m pip install --upgrade pip virtualenv
+          sudo apt update
+          sudo apt install -qy --no-install-recommends libgl1 libssl1.1 libdbus-1-3 libxcb-xfixes0-dev libxcb-shape0-dev libunwind8 libegl1-mesa
+          wget http://mirrors.kernel.org/ubuntu/pool/main/libf/libffi/libffi6_3.2.1-8_amd64.deb
+          sudo apt install ./libffi6_3.2.1-8_amd64.deb
+          python3 ./mach bootstrap-gstreamer
+      - name: Run tests
+        run: |
+          python3 ./mach test-wpt \
+            --release --processes $(nproc) --timeout-multiplier 2 \
+            --total-chunks ${{ env.max_chunk_id }} --this-chunk ${{ matrix.chunk_id }} \
+            --log-raw test-wpt.${{ matrix.chunk_id }}.log \
+            --log-raw-unexpected unexpected-test-wpt.${{ matrix.chunk_id }}.log \
+            --filter-intermittents filtered-test-wpt.${{ matrix.chunk_id }}.json
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+          INTERMITTENT_TRACKER_DASHBOARD_SECRET: ${{ secrets.INTERMITTENT_TRACKER_DASHBOARD_SECRET }}
+      - name: Archive filtered results
+        uses: actions/upload-artifact@v3
+        if: ${{ always() }}
+        with:
+          name: wpt-filtered-results-linux-${{ env.LAYOUT }}
+          path: |
+            filtered-test-wpt.${{ matrix.chunk_id }}.json
+            unexpected-test-wpt.${{ matrix.chunk_id }}.log
+      - name: Archive logs
+        uses: actions/upload-artifact@v3
+        if: ${{ failure() }}
+        with:
+          name: wpt-logs-linux-${{ env.LAYOUT }}
+          path: |
+            test-wpt.${{ matrix.chunk_id }}.log
+            filtered-wpt-results.${{ matrix.chunk_id }}.json
+
+  report-test-results:
+    name: Reporting test results
+    runs-on: ubuntu-latest
+    if: "!cancelled() && success('build-linux') && ${{ github.ref_name == 'try-wpt' || inputs.wpt }}"
+    needs:
+      - "linux-wpt"
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - uses: actions/download-artifact@v3
+        with:
+          name: wpt-filtered-results-linux-${{ env.LAYOUT }}
+          path: wpt-filtered-results-linux
+      - name: Create aggregated unexpected results
+        run: cat wpt-filtered-results-linux/*.log > unexpected-test-wpt.log
+      - name: Archive aggregate results
+        uses: actions/upload-artifact@v3
+        with:
+          name: wpt-filtered-results-linux-${{ env.LAYOUT }}
+          path: |
+            unexpected-test-wpt.log
+      - name: Comment on PR with results
+        run: etc/ci/report_aggregated_expected_results.py wpt-filtered-results-linux/*.json
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  build_result:
+    name: homu build finished
+    runs-on: ubuntu-latest
+    if: always()
+    # needs all build to detect cancellation
+    needs:
+      - "build-linux"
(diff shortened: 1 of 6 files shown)


.github/workflows/upstream-wpt-changes.yml
```diff
@@ -2,6 +2,8 @@ name: WPT export
 on:
   pull_request_target:
     types: ['opened', 'synchronize', 'reopened', 'edited', 'closed']
+    paths:
+      - 'tests/wpt/**'
```
mrobinson, line 6: This is looking really cool. One comment on this line though: I think we cannot add this here. When the PR previously changed files in tests/wpt, but then doesn't it still needs to run in order to close the upstream PR. That's the reason I didn't add the 'paths' directive originally.

sagudev: Is that edge case happening frequently?

Reason for this change is that `WPT export` is run on many times for nothing. Maybe this edge case could be handled by bors or by additional scheduled workflow that would check for "stale" WPT export PRs.

mrobinson: I think this adds a lot of complication when the proper functioning of the script requires that it always run. In general, it's very fast in comparison to the quick check and was before running for every PR, just on another server as a webhook. One thing we could do is to consider integrating it into the quick check script so that it doesn't show up at the bottom of the PR and also fixing the bug where it does not work well when encountering a merge commit.

sagudev: If we integrate it into quick-check, don't we then need to run quick-check for all those PR events?

Anyway if it isn't that simple I would just remove it and leave it to the future, as there is enough other stuff in this PR already.

sagudev:
@mrobinson what is the status with running WPT test on layout-2020? Do we need separate trackers for or how it goes.

Currently I am testing modular workflows with inputs. The goal is to make testing layout-2020 vs 2013 just matter of [workflow flags](https://github.com/sagudev/servo/blob/b1fbeef4963d894b763dc99522618221fd558d17/.github/workflows/main.yml#L42-L56).

sagudev:
Okay, so I enabled layout2020 checks in PR and in main check. Both have unit test disabled as it has the following failures:
```console
 Compiling layout_2020 v0.0.1 (/home/runner/work/servo/servo/components/layout_2020)
error[E0404]: expected trait, found struct `Gen`
  --> components/layout_2020/tests/floats.rs:62:12
   |
62 |         G: Gen,
   |            ^^^ not a trait

error[E0404]: expected trait, found struct `Gen`
  --> components/layout_2020/tests/floats.rs:86:12
   |
86 |         G: Gen,
   |            ^^^ not a trait

error[E0404]: expected trait, found struct `Gen`
   --> components/layout_2020/tests/floats.rs:349:12
    |
349 |         G: Gen,
    |            ^^^ not a trait
```

sagudev:
@mrobinson Are Intermittent errors dependent on layout and should both layouts have separate report_agregate_results handling?

There is still some space left for unification (nightly-rust, nightly, wpt-nightly) but that would require some additional work and I would rather put it till this lands as this PR is big enough already.

mrobinson (commented):
Thanks for working on this. A couple thoughts:
- I think it might be good to split this change up into multiple pull requests, as there are a few unrelated things going on here. For instance, turning on sccache seems like an interesting idea, but if something goes wrong and we need to revert it, we'll end up reverting the split of the platform-specific workflows as well. Maybe this change can simply handle splitting the workflows?
- I think it's useful that the test results show up in a GitHub comment. It makes them easier to find and the idea is that we can gradually work on reducing the amount of tests that are flaky because the results are right there.
- By splitting the Layout 2013 and Layout 2020 build jobs, we duplicate more effort whereas before Layout 2020 added about 7 minutes to the total build and used one less builder. I suppose there is a tradeoff there, as total job time might be less now, but what was your thought process here?

sagudev:
>    I think it might be good to split this change up into multiple pull requests, as there are a few unrelated things going on here. For instance, turning on sccache seems like an interesting idea, but if something goes wrong and we need to revert it, we'll end up reverting the split of the platform-specific workflows as well. Maybe this change can simply handle splitting the workflows?

That was actually my initial plan, but it was useful to have cache while testing. Will remove sccache and make it in new PR. Should WPT github chacks also be split into separate PR?

>  I think it's useful that the test results show up in a GitHub comment. It makes them easier to find and the idea is that we can gradually work on reducing the amount of tests that are flaky because the results are right there.

I am actually just adding WPT to github checks on top of github comments. What I was asking was if different layout should have different comments and WPT checks or do we only show layout 2013 results as we are now.

> By splitting the Layout 2013 and Layout 2020 build jobs, we duplicate more effort whereas before Layout 2020 added about 7 minutes to the total build and used one less builder. I suppose there is a tradeoff there, as total job time might be less now, but what was your thought process here?

It just felt more natural to have separate builds for them. The faster the linux built finishes the faster WPT checks starts, so WPT tests profit from building per layout in separate builds. Also I believe 7 min is only accounting for building as we currently do not run unit-tests for layout-2020, but in near future we certainly will, and then separate builds will make more sense in quick-check to.

sagudev:
>    I think it might be good to split this change up into multiple pull requests, as there are a few unrelated things going on here. For instance, turning on sccache seems like an interesting idea, but if something goes wrong and we need to revert it, we'll end up reverting the split of the platform-specific workflows as well. Maybe this change can simply handle splitting the workflows?

That was actually my initial plan, but it was useful to have cache while testing. Will remove sccache and make it in new PR. Should WPT github chacks also be split into separate PR?

>  I think it's useful that the test results show up in a GitHub comment. It makes them easier to find and the idea is that we can gradually work on reducing the amount of tests that are flaky because the results are right there.

I am actually just adding WPT to github checks on top of github comments. What I was asking was if different layout should have different comments and WPT checks or do we only show layout 2013 results as we are now.

> By splitting the Layout 2013 and Layout 2020 build jobs, we duplicate more effort whereas before Layout 2020 added about 7 minutes to the total build and used one less builder. I suppose there is a tradeoff there, as total job time might be less now, but what was your thought process here?

It just felt more natural to have separate builds for them. The faster the linux built finishes the faster WPT checks starts, so WPT tests profit from building per layout in separate builds. Also I believe 7 min is only accounting for building as we currently do not run unit-tests for layout-2020, but in near future we certainly will, and then separate builds will make more sense in quick-check to.

mrobinson:
> That was actually my initial plan, but it was useful to have cache while testing. Will remove sccache and make it in new PR. Should WPT github chacks also be split into separate PR?

Yes, that would be really useful if it's not too much work. Thank you!

> I am actually just adding WPT to github checks on top of github comments. What I was asking was if different layout should have different comments and WPT checks or do we only show layout 2013 results as we are now.

Oh, thanks for the clarification. I think that's fine then. For some background, we are actually in a very interesting moment in regards to Layout 2013 and Layout 2020. We are in the process of choosing one of these layout engines to focus on. One possibility, is that the other layout engine will be removed from the tree in order to focus development efforts.  It's unclear to me if we will turn on tests for Layout 2020 before or after this moment, and obviously it depends on what the result of the choice is. Really, this just needs a wider discussion to figure out what the best path forward is and what order to do things in.

> It just felt more natural to have separate builds for them. The faster the linux built finishes the faster WPT checks starts, so WPT tests profit from building per layout in separate builds. Also I believe 7 min is only accounting for building as we currently do not run unit-tests for layout-2020, but in near future we certainly will, and then separate builds will make more sense in quick-check to.

Hopefully, the context above sheds some light here as well. That said, unit tests might be a good place to start with this process and then WPT tests can follow. That could avoid overwhelming the builders right away. Right now dedicated 40 builders to a single WPT run might be too much to ask though.

sagudev:
Also what is good with this modular approach is that, that any test (ex. wpt on layout-2013) can be easily switched on or off, which is very useful in this times.

> That could avoid overwhelming the builders right away. Right now dedicated 40 builders to a single WPT run might be too much to ask though.

We could run layout-2020 WPT test only on PRs that change any layout-2020 files, this should be pretty trivial change in main.yml after this PR.

mrobinson (changes requested):
Thank you for splitting this up! I really like how this avoids so much duplication in the workflow files. I just have a few nits, but I think @mukilan, @delan, or @jdm should also look at this change since they have done work on these files as well and may spot something that I missed.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,169 @@
...
+        with:
+          name: wpt-logs-linux
+          path: |
+            test-wpt.${{ matrix.chunk_id }}.log
+            filtered-wpt-results.${{ matrix.chunk_id }}.json
+
+  report-test-results:
+    name: Reporting test results
+    runs-on: ubuntu-latest
+    if: "!cancelled() && success('build-linux') && ${{ github.ref_name == 'try-wpt' || contains(inputs.called, 'called') }}"
```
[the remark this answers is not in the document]
sagudev: Old relics, should be checking `inputs.wpt`.

.github/workflows/main.yml
```diff
@@ -1,17 +1,12 @@
 name: Merge queue and try
 
 on:
-  # Triggers the workflow on push events but only for the master branch
   push:
-    branches: [ "auto", "try", "try-linux", "try-mac", "try-windows", "try-wpt"]
```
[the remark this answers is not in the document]
sagudev: Yes, we have them. Look at #29431 where I add them.

sagudev:
I fixed handling of testing different layouts in wpt test (using different artifact names) and WPT-nightly is now fully using abstract workflows.

delan:
Sorry, I haven’t had a chance to look at this yet, but I will review this tomorrow.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,193 @@
...
+      - name: Bootstrap dependencies
+        run: sudo apt update && python3 ./mach bootstrap
+      - name: Tidy
+        run: python3 ./mach test-tidy --no-progress --all
+      - name: Release build
+        run: python3 ./mach build --release --with-${{ env.LAYOUT }}
+      - name: Smoketest
+        run: xvfb-run python3 ./mach smoketest
+      - name: Unit tests
+        if: ${{ inputs.unit-tests }}
```
delan, line 58: wow, good catch! it looks like we weren’t running these for linux :o

delan (approved):
Great work, thank you very much!

I have one minor concern, which is that previously we were able to run two or three concurrent builds without contention (up to 22/60 jobs per build), and now with layout 2020 we may only be able to run one without contention (up to 42/60 jobs per build).

But trying to mitigate that in advance by reducing the number of chunks can have negative side effects (slower builds and maybe even new failures), so I think we should deal with that separately, assuming it even causes any problems.

sagudev:
I will wait a day or two if we would need to revert and/or fix the patch then I will send new PR for sccache and WPT check output.

> I have one minor concern, which is that previously we were able to run two or three concurrent builds without contention (up to 22/60 jobs per build), and now with layout 2020 we may only be able to run one without contention (up to 42/60 jobs per build).

As for concurrent builds, I do not really see the need for them, due to bors and it's landing. `bors try` commands can be fully replaced with `mach try` (see https://github.com/servo/servo/pull/29431) or manually dispatching any workflow (more specific runs are also possible due to this PR), which allows contributors to fully test changes in their own forks (using their build job limits). 

`bors try` may have been important when CI was running on TC, but now everybody can run same CI workflows within their own account. Ideally we could just get their results of their run instead of running our own jobs (with checks in place that testing was done on same commit SHA).

sagudev:
I fixed if guard so agregated reporting is working again.

delan:
@bors-servo r+

> As for concurrent builds, I do not really see the need for them, due to bors and it's landing. `bors try` commands can be fully replaced with `mach try` (see #29431) or manually dispatching any workflow (more specific runs are also possible due to this PR), which allows contributors to fully test changes in their own forks (using their build job limits).

There are also the nightly builds, which can run concurrently with auto/try builds, but hopefully there won’t be too much delay caused by contention with those.

PR #29576, opened 2023-04-03
WPT aggregated results also as github check

sagudev:
Report WPT aggregated results also using GitHub check (so we get those sexy results even when there is no associated PR).

[example](https://github.com/sagudev/servo/runs/12193772513)

Last missing component coming from #29534.


jdm:
@bors-servo r+
Exciting!

PR #29584, opened 2023-04-05
Remove unnecessary steps from "text/plain encoding algorithm"

### LINKED ISSUE #25221: Some steps for "text/plain encoding algorithm, given an entry list" have been moved elsewhere in WHATWG; Servo code is performing them in their old position.
https://github.com/whatwg/html/pull/3645 moved the definition of where some form data steps occur, relating to applying a charset to a form data element and to populating the special _charset_ form input.

https://github.com/servo/servo/blob/86fae5894e3b52b2a21afbe6dc468e7a167fd71c/components/script/dom/htmlformelement.rs#L296 is coded and commented to follow a 6-step process taking a charset, where the current state of https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#text%2Fplain-encoding-algorithm has a 3-step process not taking a charset, with the charset steps happening elsewhere.

The current code already appears to do the right thing in the relevant WPT cases, so this could be a purely cosmetic refactor if there's not some other corner case that cares about when these steps happen.

ohno418:
For "text/plain encoding algorithm", the specification [1] has changed [2] and "_charset_" is not now handled here. It's supposed to be handled in "construct the form data set" algorithm, and we've already implemented that [3]. So we now have extra steps for "text/plain encoding" algorithm.

Remove no longer necessary steps from text/plain encoding algorithm so that it meets the specification.

[1]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#text/plain-encoding-algorithm
[2]: https://github.com/whatwg/html/pull/3645
[3]: https://github.com/servo/servo/pull/25217

---

- [x] These changes fix #25221 (GitHub issue number if applicable)

- [x] These changes do not require tests because this patch doesn't expect actual behavior, just removing extra steps.

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -659,20 +659,15 @@ impl HTMLFormElement {
         let mut result = String::new();
 
         // Step 2
-        let encoding = self.pick_encoding();
-
-        // Step 3
-        let charset = encoding.name();
-
-        for entry in form_data.iter_mut() {
-            // Step 4, 5
-            let value = entry.replace_value(charset);
-
-            // Step 6
-            result.push_str(&*format!("{}={}\r\n", entry.name, value));
+        for entry in form_data.iter() {
+            let value = match &entry.value {
+                FormDatumValue::File(f) => f.name(),
+                FormDatumValue::String(s) => s,
+            };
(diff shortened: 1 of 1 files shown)


jdm (approved):

jdm:
@bors-servo r+
Thanks!

PR #29596, opened 2023-04-06
Package in platform workflows & nightly switch to them

sagudev:
Packaging takes only 1min per OS which I think it is reasonable tradeoff.

Nightly builds are also getting speed up due to sccache.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -50,7 +50,6 @@ env:
 
 jobs:
   build-linux:
-    if: github.repository == 'servo/servo' || github.event_name == 'push'
     name: Build
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/linux.yml
```diff
@@ -65,6 +73,18 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.PACKAGE }}
+          path: target/release/servo-tech-demo.tar.gz
+      - name: Upload
+        if: ${{ inputs.upload }}
+        run: python3 ./mach upload-nightly ${{ env.PACKAGE }} --secret-from-environment
+        env:
+          S3_UPLOAD_CREDENTIALS: ${{ secrets.S3_UPLOAD_CREDENTIALS }}
       - name: Package binary
```
mukilan, line 88: Do we still need this step and the next one (Archive binary)?

sagudev: I am not sure, but quick inspection of [python/servo/package_commands.py](https://github.com/servo/servo/blob/master/python/servo/package_commands.py) shows that command does not just zip files, thus I imagined that they are different. I do not know if we could/should run WPT tests on packaged builds vs. raw builds so I did not change it. @jdm may know what was rationale behind current approach.

mukilan: Thanks for the explanation! I guess we can keep both artifacts for now and decide later if switching tests to the packaged artifacts makes sense.

PR #29610, opened 2023-04-10
Use notify-py to send notifications 

mrobinson:
- Use notify-py to send notifications, but use a custom notifier on
  Linux since transient (ie non-sticky) notifications are not supported.
- Add an icon to the notification.
- Don't send a notification after doing `./mach check` because that
  can trigger notifications when `rust-analyzer` is working.

---

- [x] These changes do not require tests because they just change build notifications.


mukilan (approved):

mrobinson:
@bors-servo r=mukilan

Thanks for the review!

PR #29618, opened 2023-04-11
Split quick check from pull request and rename WPT export actions

### LINKED ISSUE #29603: Four quick check jobs are triggered for every pull request update
I believe what is happening is that because `quick_check.yml` triggers on:

```
  pull_request:
  push:
    branches:
      ["**", "!master", "!auto", "!try", "!try-linux", "!try-mac", "!try-windows", "!try-wpt", "!dependabot/**"]
```

one set of quick check actions is triggered for the push to the pull request branch and another is triggered for the change to the pull request. Can this simply be the following?

```
 pull_request:
   types: [opened, synchronize]
```
cc @sagudev

mrobinson:
Split out the quick check GitHub action (which is really only meant for forks). This prevents the quick check from running twice for PRs from branches on the upstream repository. Also rename the WPT export action file to make it clearer that it runs for pull requests. This change also means we no longer trigger builds when simply editing pull request body or title.

---

- [x] These changes fix #29603
- [x] These changes do not require tests because this is just a reorganization of the workflow files.

### DIFF
--- a/.github/workflows/pull-request-wpt-export.yml
+++ b/.github/workflows/pull-request-wpt-export.yml
@@ -1,4 +1,4 @@
-name: WPT export
+name: Pull request (WPT export)
 on:
   pull_request_target:
     types: ['opened', 'synchronize', 'reopened', 'edited', 'closed']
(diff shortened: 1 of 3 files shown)


jdm:
Nice!
@bors-servo r+

PR #29621, opened 2023-04-12
Upload nightly builds to Github Releases

mukilan:
This change extends the `mach upload-nightly` command to publish the nightly builds for all platforms as GH Release assets.

The GH releases are made on a separate repository so that we can persist older nightly builds without having to accumulate git tags for them.

Example releases on a fork can be seen here: https://github.com/mukilan/servo-nightly-builds/releases/
Assets of the *latest* nightly releases will available at a stable url e.g for linux
`https://github.com/mukilan/servo-nightly-builds/releases/latest/download/servo-latest.tar.gz`

Some design tradeoffs in this approach are:
1. To allow the 'latest' link from servo.org to remain stable, the release assets are named 'servo-latest.{ext}' instead of containing the release tag/date.
2. The release is created as draft and published atomically when all platforms have been built successfully. This allows us to link to the 'latest' alias from servo.org while gauranteeing that it contains builds for all platforms. The other option here would be to have code in servo.org UI that uses GH API to find the most recent release with a successful build for a given platform.
3. The tags in the nightly repo are all based on the same commit that has no relation to servo code base.

---

- [x] These changes do not require tests because it extends the nightly CI job

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -11,42 +11,105 @@ env:
   SHELL: /bin/bash
 
 jobs:
+  create-draft-release:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
+    name: Create Draft GH Release
+    runs-on: ubuntu-20.04
+    steps:
+      - id: create-release
+        run: |
+          NIGHTLY_TAG=$(date "+%F")
+          RELEASE_URL=$(gh release create "${NIGHTLY_TAG}" \
+            --draft \
+            --title "${NIGHTLY_TAG}" \
+            --notes 'Nightly builds based on servo/servo@${{ github.sha }}' \
+            --repo ${NIGHTLY_REPO})
+          TEMP_TAG=$(basename "$RELEASE_URL")
+          RELEASE_ID=$( \
+            gh api -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "/repos/${NIGHTLY_REPO}/releases/tags/${TEMP_TAG}" \
+            | jq '.id' \
+          )
+          echo "RELEASE_ID=${RELEASE_ID}" >> ${GITHUB_OUTPUT}
+        env:
+          GITHUB_TOKEN: ${{ secrets.NIGHTLY_REPO_TOKEN }}
+          NIGHTLY_REPO: ${{ github.repository_owner }}/servo-nightly-builds
+    outputs:
+      release-id: ${{ steps.create-release.outputs.RELEASE_ID }}
+
+  publish-nightly-release:
+    # This job is only useful when run on upstream servo.
+    if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
+    name: Publish GH Release for nightly
+    runs-on: ubuntu-20.04
+    steps:
+      - run: |
+          gh api \
+            --method PATCH \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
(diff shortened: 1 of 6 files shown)


python/servo/package_commands.py
```diff
@@ -674,6 +677,22 @@ def upload_to_s3(platform, package, timestamp):
                 }
             )
 
+            if github_release_id:
```
mrobinson, line 680: This seems fairly independent of `upload_to_s3` so perhaps create a separate `upload_to_github_release` function and use an early return if there is no if github_release_id.

python/servo/package_commands.py
```diff
@@ -674,6 +677,22 @@ def upload_to_s3(platform, package, timestamp):
...
+                from github import Github
```
mrobinson, line 681: I would move this to the top of the Python script. Typically, these inner use statements are only used when the package isn't always available.

python/servo/package_commands.py
```diff
@@ -674,6 +677,22 @@ def upload_to_s3(platform, package, timestamp):
...
+                g = Github(os.environ['NIGHTLY_REPO_TOKEN'])
+                nightly_repo = g.get_repo(os.environ['NIGHTLY_REPO'])
+                release = nightly_repo.get_release(int(github_release_id))
```
mrobinson, line 684: I'm fairly sure the `int(...)` cast here can be avoided by specifying `type=int` in the `@CommandArgument` decorator.

python/servo/package_commands.py
```diff
@@ -674,6 +677,22 @@ def upload_to_s3(platform, package, timestamp):
...
+                if '2020' in platform:
+                    asset_name = f'servo-latest-layout2020.{extension}'
```
mrobinson, line 686: Nit: `layout-2020` for the sake of consistency and I would use `nightly` instead of `latest`.

python/servo/package_commands.py
```diff
@@ -674,6 +677,22 @@ def upload_to_s3(platform, package, timestamp):
...
+                else:
+                    asset_name = f'servo-latest.{extension}'
+                sha_name = f'{asset_name}.sha256'
```
mrobinson, line 689: I'd avoid the temporary here, just so that the code below is easier to read.

mrobinson (commented):
This looks great to me! I have a few small nits below.

python/servo/package_commands.py
mukilan: Done in 61fb795

python/servo/package_commands.py
mukilan: Done in 61fb795

python/servo/package_commands.py
mukilan: Makes sense. Addressed in 61fb795. I've also extracted the package hash computation to be outside of upload_to_s3.

python/servo/package_commands.py
mukilan: Changed `layout2020` to `layout-2020` in 61fb795. 

As discussed over Matrix, I've retained the 'latest' suffix for 2020 to be consistent with existing naming convention for nightly builds, but this can be changed in the future if needed since only our internal WPT dashboard uses 2020 builds today.

.github/workflows/nightly.yml
```diff
@@ -11,42 +11,101 @@ env:
   SHELL: /bin/bash
 
 jobs:
+  create-draft-release:
+    name: Create Draft GH Release
```
mrobinson, line 15: These two new jobs should likely be guarded by the same if statement as the `upload-*` jobs:

```
if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
````

mrobinson (approved):
Great! Let's give this a shot. One small comment below:

PR #29623, opened 2023-04-12
Import WPT tests every week instead of every day

mrobinson:
We often have multiple imports stacked up which makes it more confusing to land each one. In addition, giving a week to do the import will give us more margin to triage failing tests and hopefully to be able to use intermittent expectations for flaky tests.

---

- [x] These changes do not require tests because this just change the frequency of a scheduled action.


CYBAI (approved):
Thank you!

PR #29627, opened 2023-04-13
Respect MIME type from headers instead of caching it

CYBAI:
Based on the spec,

> A [Request](https://fetch.spec.whatwg.org/#request) object’s [MIME type](https://fetch.spec.whatwg.org/#concept-body-mime-type) is to return the result of [extracting a MIME type](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type) from its [request](https://fetch.spec.whatwg.org/#concept-request-request)’s [header list](https://fetch.spec.whatwg.org/#concept-request-header-list).

request and response should always return the MIME type from their header list. As we're currently caching the mime type directly, this PR will help to remove the caching field and always retrieve from headers instead.

---


components/script/dom/request.rs
```diff
@@ -444,8 +442,6 @@ impl Request {
         r.request.borrow_mut().body = input_body;
 
         // Step 41
```
mukilan, line 444: Should we remove this dangling comment and update the other step comments?

CYBAI: ah, thanks for catching it. removing the dangling comments were missed to commit. but yes, let me update other step comments as well. 🙏

CYBAI: Just updated them in da35d43.

components/script/dom/response.rs
```diff
@@ -100,34 +99,29 @@ impl Response {
...
 
-        // Step 5
+        // Step 4
         *r.raw_status.borrow_mut() = Some((init.status, init.statusText.clone().into()));
 
-        // Step 6
+        // Step 5
         if let Some(ref headers_member) = init.headers {
-            // Step 6.1
-            r.Headers().empty_header_list();
```
CYBAI, line 115: this is mentioned as a no-op and removed in https://github.com/whatwg/fetch/commit/90cd540dc2a25df48b85a361f49901ac0835957b so I just removed it along with the step comments update commit.

components/script/dom/request.rs
```diff
@@ -81,69 +79,71 @@ impl Request {
         // Step 2
         let mut fallback_mode: Option<NetTraitsRequestMode> = None;
 
-        // Step 3
+        // FIXME(cybai): As the spec changed in https://github.com/whatwg/fetch/pull/1153,
+        //               we will need to change the default value of credentials for
+        //               NetTraitsRequest and then remove fallback here.
         let mut fallback_credentials: Option<NetTraitsRequestCredentials> = None;
```
CYBAI, line 85: I will file a follow-up issue when this PR is merged.

mukilan (approved):

mukilan:
@bors-servo r+

Thank you !

CYBAI:
ah, @mukilan sorry, I thought I forgot to remove `empty_header_list` in previous commit but I was wrong... could you please help me approve again? 🙏

mukilan:
@bors-servo r+

no worries!

PR #29631, opened 2023-04-14
Fix calculation of block axis for abspos elements in Layout 2020

mrobinson:
The calculation of the block axis positioning of absolutely positioned elements was using the inline size of the containing block instead of the block size.

---


mrobinson:
@delan I have been looking into how insets are calculated for positioned elements and this intersects a bit.

mrego (approved):

delan:
Thanks for following up on this!

PR #29639, opened 2023-04-17
Reorganize Servo's WPT Python scripts

mrobinson:
This change moves all of Servo's WPT Python support scripts into one directory as they were previously scattered throughout the directory structure. This should allow more code reuse and make it easier to understand how everything fits together.

The changes:

- `tests/wpt/update` → `python/wpt/importer`
- `etc/ci/upstream-wpt-changes/wptupstreamer` → `python/wpt/exporter`
- `etc/ci/upstream-wpt-changes/test.py` → `python/wpt/test.py`
- `etc/ci/upstream-wpt-changes/tests` → `python/wpt/tests`
- `tests/wpt/servowpt.py` → - `python/wpt/update.py` - `python/wpt/run.py`
- `tests/wpt/manifestupdate.py` → `python/wpt/manifestupdate.py`

This change also removes
 - The ability to run the `update-wpt` and `test-wpt` commands without using `mach`. These didn't work very well, because it was difficult to get all of the wptrunner and mach dependencies installed outside of the Python virtualenv. It's simpler if they are always run through `mach`.
- The old WPT change sync / upstreaming script that was no longer used. This allows this code to pass `test-tidy` now.

---


python/wpt/importer/__init__.py
```diff
@@ -0,0 +1,57 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+
+from .tree import GitTree, GeckoCommit
```
mukilan, line 8: Just curious if there is any particular reason for retaining the 'gecko' terminology in this file (I see few more references to gecko in doc strings below).

mrobinson: That's a good question. My thought was to only move the scripts now and then gradually start cleaning them up later. There are also some things that could be shared between the importer and exporter, once these two scripts are in the same module.

mukilan (approved):
Nice refactoring! Thank you for the walkthrough, @mrobinson !

mrobinson:
Thank you for the review!

PR #29641, opened 2023-04-18
Throw HierarchyRequest error for attr node in pre-insertion

CYBAI:
Based on Step 4 of https://dom.spec.whatwg.org/#concept-node-ensure-pre-insertion-validity, we should throw `HierarchyRequest` error when the node is not a [DocumentFragment](https://dom.spec.whatwg.org/#documentfragment), [DocumentType](https://dom.spec.whatwg.org/#documenttype), [Element](https://dom.spec.whatwg.org/#element), or [CharacterData](https://dom.spec.whatwg.org/#characterdata) [node](https://dom.spec.whatwg.org/#concept-node).

Thus, Attr is not one of them; we should throw an error in this case.

(I'm guessing it was a kind of typo in https://github.com/servo/servo/pull/25310 because we need to handle Attr with `unreachable` in line 1946)

So, with doing so, we can fix the CRASH in `/dom/attributes-are-nodes.html`.

---

### DIFF
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -1872,8 +1872,7 @@ impl Node {
             NodeTypeId::Element(_) |
             NodeTypeId::CharacterData(CharacterDataTypeId::ProcessingInstruction) |
             NodeTypeId::CharacterData(CharacterDataTypeId::Comment) => (),
-            NodeTypeId::Document(_) => return Err(Error::HierarchyRequest),
-            NodeTypeId::Attr => unreachable!(),
+            NodeTypeId::Document(_) | NodeTypeId::Attr => return Err(Error::HierarchyRequest),
         }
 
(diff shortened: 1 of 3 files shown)


CYBAI:
> This looks good to me. Any idea why the test is still failing here?

I think the remaining failure is caused by missing handle for `Attr` in the `ReplaceChild` function. Let me fix it along with this PR!

https://github.com/servo/servo/blob/9acb9cc5cf21d14709355a3c75af7202e9301bd5/components/script/dom/node.rs#L2623-L2633

CYBAI:
the test from `WPT tests (18)` failed with network issue so I tried to rerun it from the GitHub Actions UI.

mrobinson:
Thank you for fixing the test!

mrobinson (approved):

PR #29644, opened 2023-04-19
Rename `ComputedValuesExt::establishes_containing_block`

mrobinson:
This renames the helper method to be a bit more accurate. For elements with static, relative, and sticky positioning, their containing block is always formed by their nearest block container ancestor. This method is really dealing with style that means an element will establish a containing block for absolutely positioned descendants.

---

### DIFF
--- a/components/layout_2020/style_ext.rs
+++ b/components/layout_2020/style_ext.rs
@@ -113,7 +113,7 @@ pub(crate) trait ComputedValuesExt {
     fn has_transform_or_perspective(&self) -> bool;
     fn effective_z_index(&self) -> i32;
     fn establishes_stacking_context(&self) -> bool;
-    fn establishes_containing_block(&self) -> bool;
+    fn establishes_containing_block_for_absolute_descendants(&self) -> bool;
     fn establishes_containing_block_for_all_descendants(&self) -> bool;
     fn background_is_transparent(&self) -> bool;
     fn get_webrender_primitive_flags(&self) -> wr::PrimitiveFlags;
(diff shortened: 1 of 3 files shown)


components/layout_2020/style_ext.rs
```diff
@@ -395,7 +395,11 @@ impl ComputedValuesExt for ComputedValues {
         !self.get_position().z_index.is_auto()
     }
 
-    fn establishes_containing_block(&self) -> bool {
+    /// Returns true if this style establishes a containing block for absolute
+    /// descendants. This will also return true if it establishes a containing
+    /// block for all descendants, which include those which are both
+    /// fixed and absolutely positioned.
```
delan, line 401: i think these can be made a bit clearer. how about

```rust
    /// Returns true if this style establishes a containing block for absolute
    /// descendants (position: absolute or fixed). This will also return true if
    /// it establishes a containing block for all descendants, because “all
    /// descendants” includes absolute descendants.
```

and in the all descendants method

```rust
    /// Returns true if this style establishes a containing block for all descendants,
    /// not just absolute descendants (position: absolute or fixed).
```

delan (commented):
This looks helpful, though I have a suggestion:

components/layout_2020/style_ext.rs
mrobinson: Thanks for the review! I have adjusted these a bit because `establishes_containing_block_for_absolute_descendants()` doesn't imply that a containing block is establishes for `position: fixed` descendants. 

The new comments are:

```
    /// Returns true if this style establishes a containing block for absolute
    /// descendants (`position: absolute`). If this style happens to establish a
    /// containing block for “all descendants” (ie including `position: fixed`
    /// descendants) this method will return true, but a true return value does
    /// not imply that the style establishes a containing block for all descendants.
    /// Use `establishes_containing_block_for_all_descendants()` instead.
    fn establishes_containing_block_for_absolute_descendants(&self) -> bool {
```

```
    /// Returns true if this style establishes a containing block for
    /// all descendants, including fixed descendants (`position: fixed`).
    /// Note that this also implies that it establishes a containing block
    /// for absolute descendants (`position: absolute`).
     fn establishes_containing_block_for_all_descendants(&self) -> bool {
```

CYBAI:
The CI failed with https://github.com/servo/servo/issues/29653.

PR #29646, opened 2023-04-19
Fix checking for empty set in getElementsByClassName

CYBAI:
Because empty set with `.all(predicate)` will always return `true`, it will result in always filtering wrong elements when the classes is empty set.

Thus, we return earlier with `always_empty` HTMLCollection in the empty classes case so that we can avoid filtering on empty sets.

With adding the checking, we can fix the failures in `/dom/nodes/getElementsByClassName-empty-set.html`.

---


mrobinson:
Nice fix! Could `HTMLCollection::by_atomic_class_name` simply have an early return with `HTMLCollection::always_empty(window, root)` if the `classes` argument is an empty vector? That would avoid a bit of extra work

CYBAI:
> Nice fix! Could `HTMLCollection::by_atomic_class_name` simply have an early return with `HTMLCollection::always_empty(window, root)` if the `classes` argument is an empty vector? That would avoid a bit of extra work

Thanks for the review! While fixing it, I've considered early return with `always_empty` but didn't do that because I thought it will result in never filtering for the live collection. However, I just changed my mind because when `classes` is an empty vector, it would never filter elements. Let me update the patch! 🙏

PR #29651, opened 2023-04-20
Do not use the CB writing mode when resolving box properties

mrobinson:
Layout 2020 does a lot of work to pass the containing block writing mode around in order to try to resolve logical box properties. This seems incorrect though, as logical box model properties use the element's writing mode to resolve to physical properties [1] [2]. This is handled correctly by the style component, but Layout 2020 does a lot of work to try to override them.

This change removes that code and also tries to use more style component data structures in place of Layout 2020's custom logical geometry primitives. In particular it removes the confusingly overloaded type aliases `LengthOrAuto` and `LengthPercentageOrAuto` from Layout 2020. These aliases were very similar to the ones from the style component, but had a different definition.

Since Layout 2020 does not have proper support for writing modes, this change should not modify behavior.

1. https://github.com/w3c/csswg-drafts/issues/3013
2. https://drafts.csswg.org/css-logical/#box

---


mrobinson:
After discussing with @Loirooriol, it seems that I was mistaken here. While the logical values should be resolved based on their element's writing mode, these physical values need to them be converted back to inline and block directions in the coordinate space of the containing block. There's some work from this PR that can be salvaged, particularly removing the duped `LengthOrAuto` and `LengthPercentageOrAuto` overloads and using more logical geometry types from the style system, but we can handle those little bit little in future PRs.

PR #29652, opened 2023-04-21
Fixed bug where ./mach test-wpt was unable to run on windows due to servowpt…

switchpiggy:
….py not correctly setting default binary path

Changed the default_binary_path variable to account for the binary_name change to "servo.exe" on win32 properly. Before, the target_dir was being changed after the default path variable was set. This would result in the following error when ./mach test-wpt was run:

"Binary path C:\Code\servo\target\release\servo does not exist"

---

- [x] These changes do not require tests because the validity of the change can be tested by running ./mach test-wpt.


mrobinson (approved):
Thank you for this fix.

PR #29655, opened 2023-04-22
Fix text-decoration for block containers in layout-2020

### LINKED ISSUE #29654: text-decoration only affects 1st IFC in layout-2020
While working on `text-indent` I noticed `end_ongoing_inline_formatting_context` has https://github.com/servo/servo/blob/9acb9cc5cf21d14709355a3c75af7202e9301bd5/components/layout_2020/flow/construct.rs#L709-L713

That ends the ongoing IFC and creates a new one with the default parameters. That is, with `TextDecorationLine::NONE`.

Indeed, just run

```html
<!DOCTYPE HTML>
<main style="text-decoration: line-through">foo<div>bar</div>baz</main>
```

The 2nd IFC in `<main>` (`baz`) gets no text decoration:
layout-2020

Works fine in layout-2013:
layout-2013

Loirooriol:
It was only applied to the 1st inline formatting context of a block container. Other IFCs were created with the Default trait, implying TextDecorationLine::NONE.

---

- [X] These changes fix #29654

### DIFF
--- a/components/layout_2020/flow/construct.rs
+++ b/components/layout_2020/flow/construct.rs
@@ -712,10 +712,13 @@ where
                 )
         });
 
+        let mut ifc = InlineFormattingContext::new(
+            self.ongoing_inline_formatting_context.text_decoration_line,
+            /* has_first_formatted_line = */ false,
+        );
+        std::mem::swap(&mut self.ongoing_inline_formatting_context, &mut ifc);
         let kind = BlockLevelCreator::SameFormattingContextBlock(
-            IntermediateBlockContainer::InlineFormattingContext(std::mem::take(
-                &mut self.ongoing_inline_formatting_context,
-            )),
+            IntermediateBlockContainer::InlineFormattingContext(ifc),
         );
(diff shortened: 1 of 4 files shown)


sagudev:
> @bors-servo try=wpt-2020

try-wpt-2020 runner is currently not implemented in workflows yet, but I am already working on it. In the mean time you can manually trigger workflow using workflow-dispatch.

Loirooriol:
Thanks, it seems I managed to trigger it manually: https://github.com/servo/servo/actions/runs/4785470043

sagudev:
While fixing it I found out that we are not actally passing `--layout-2020` to wpt test, so the results aren't really relevant. Fixes are on their way.

Loirooriol:
These unexpected results are unrelated to my PR. I guess that layout-2020 tests should run when landing patches to ensure the expectations are always up-to-date.

mrobinson (approved):

Loirooriol:
Not sure why it failed, I didn't get conflicts when rebasing locally. Let's try again.

Issue #29659, opened 2023-04-24
Scrolling from script does not work consistently when triggered in <script> tag


mrobinson:
When scrolling is triggered from script during page load from a <script> tag, it doesn't work consistently. There seems to be a timing issue where script-based scrolling only works after the page has fully loaded. Theoretically, scrolling from script should cause a layout flush so there shouldn't be any timing issues. This likely causes many WPT tests to fail.

PR #29660, opened 2023-04-24
Rework CB management during stacking context tree construction

mrobinson:
Manage containing blocks and WebRender `SpaceAndClip` during stacking context tree construction using the `ContainingBlockInfo` data structure. This will allow us to reuse this data structure whenever we traverse the fragment tree. In addition, `StackingContextBuilder` is no longer necessary at all. This change also fixes some bugs where fixed position fragments were not placed in the correct spatial node. Unfortunately, these fixes are difficult to test because of #29659.

---

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -5,9 +5,7 @@
 use crate::cell::ArcRefCell;
 use crate::display_list::conversions::ToWebRender;
 use crate::display_list::DisplayListBuilder;
-use crate::fragments::{
-    AbsoluteOrFixedPositionedFragment, AnonymousFragment, BoxFragment, Fragment,
-};
+use crate::fragments::{AnonymousFragment, BoxFragment, Fragment};
 use crate::geom::PhysicalRect;
 use crate::style_ext::ComputedValuesExt;
 use euclid::default::Rect;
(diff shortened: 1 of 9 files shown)


mrobinson:
I should be able to write a test for this once #29661 lands.

mrobinson:
These failed tests were fixed by #29671.

mrobinson:
I've updated the PR with the following changes:
- There is now an assertion in `BoxFragment::build_stacking_context_tree_maybe_creating_reference_frame` verifying that all fragments that establish a reference frame also establish a containing block for all descendants. This is to help show that the temporarily bogus properties on `ContainingBlockInfo` will be overwritten before they are read.
- There is a comment when starting the stacking context tree construction explaining why `ContainingBlockInfo` is initialized with 3 containing blocks instead of using one of the `new_` functions.
- I have renamed a variable to make it consistent with another in the same function.

delan:
Great work, and thanks again for walking me through it!

@bors-servo r+

mrobinson:
> -  FAIL [expected PASS] /css/css-transforms/transform-containing-block-and-scrolling-area-for-fixed.html

Odd, this test works locally. I wonder if this is an issue with flakiness in scrolling.

mrobinson:
> Odd, this test works locally. I wonder if this is an issue with flakiness in scrolling.

Oh, obviously the new test will not pass in the old layout system so needs proper expectations!

mrobinson:
@bors-servo r=delan

- Rebased.

PR #29661, opened 2023-04-24
Fix various issues with scrolling

### LINKED ISSUE #3078: Regression in Servo /css/css-backgrounds tests
Caused by #2992, according to https://github.com/servo/servo/pull/21725#issuecomment-422282806
cc @nical 
Presumably, the following tests fail:
* /css/css-backgrounds/background-attachment-local/attachment-local-positioning-2.html
* /css/css-backgrounds/background-attachment-local/attachment-scroll-positioning-1.html
* /css/css-backgrounds/scroll-positioned-multiple-background-images.html

@pyfisch is this correct and complete?

### LINKED ISSUE #29659: Scrolling from script does not work consistently when triggered in <script> tag

mrobinson:
This change fixes three issues with scrolling:

 - In order for scrolls from script to properly scroll nodes in WebRender that might require a layout, Servo should first flush layout and send the display list to WebRender.
 - `overflow: scroll` elements were not scrollable from script. This change also refactors the detection of a potentially scrollable body in quirks mode in order to make it match the specification more closely.
- Enable script scrolls in Layout 2020 by adding an implementation of the scroll area query, which is necessary to ensure that an element does support scrolling.

Each of these changes is a separate commit here with corresponding test expectation updates.

Fixes #29659.
Fixes servo/webrender#3078.

---

- [x] These changes fix #29659 and servo/webrender#3078.


mukilan (approved):

mrobinson:
@bors-servo r=mukilan

Thank you for the review!

PR #29662, opened 2023-04-24
try-wpt-2020 runner & run wpt test on selected layout

sagudev:
wpt-2020 try runner was not enabled as observed in #29655 and even if it was it would have not been working right (as observed in workflow dispatch).

There may be additional changes needed for WPT importer/report_agregator as quick look over it shows flags are not passed over to them.

A lot of layout2020 tests are still falling.

Tests:
- [Quick check](https://github.com/sagudev/servo/actions/runs/4786225230)
- [wpt](https://github.com/sagudev/servo/actions/runs/4786228332)
- [wpt-2020](https://github.com/sagudev/servo/actions/runs/4786229751)

### DIFF
--- a/python/wpt/__init__.py
+++ b/python/wpt/__init__.py
@@ -33,6 +33,8 @@ def create_parser():
                         help="Pass preferences to servo")
     parser.add_argument('--layout-2020', '--with-layout-2020', default=False,
                         action="store_true", help="Use expected results for the 2020 layout engine")
+    parser.add_argument('--layout-2013', '--with-layout-2013', default=True,
+                        action="store_true", help="Use expected results for the 2013 layout engine")
     parser.add_argument('--log-servojson', action="append", type=mozlog.commandline.log_file,
                         help="Servo's JSON logger of unexpected results")
     parser.add_argument('--always-succeed', default=False, action="store_true",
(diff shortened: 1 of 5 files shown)


python/wpt/__init__.py
```diff
@@ -33,6 +33,8 @@ def create_parser():
                         help="Pass preferences to servo")
     parser.add_argument('--layout-2020', '--with-layout-2020', default=False,
                         action="store_true", help="Use expected results for the 2020 layout engine")
+    parser.add_argument('--layout-2013', '--with-layout-2013', default=False,
```
mrobinson, line 36: I know that this is just here to avoid an error about unknown arguments, but maybe the default should be true just to avoid confusion.

mrobinson (commented):
Thank you for this fix! Just one small comment.

python/wpt/__init__.py
sagudev: If it is default true, then we can get in weird situation with both layout2020 and 2013 being selected (well 2020 would be selected). Also that's how it is done here: https://github.com/servo/servo/blob/master/python/servo/command_base.py#L849.

mrobinson:
@bors-servo r+

Thanks for this fix!

mrobinson:
> This WPT error happened multiple times aslo in my testing: https://github.com/sagudev/servo/runs/12997535584

I think this is a flaky result that made it into the WPT update.

PR #29669, opened 2023-04-26
Fixes for running WPT tests in GitHub Actions

sagudev:
- [report aggregated WPT results based on layout](https://github.com/servo/servo/commit/5ad1f241e26d7a7a62bd41e01448dde79a6b6ea1)
- [Run wpt only on try-wpt,try-wpt-2020 branches](https://github.com/servo/servo/commit/0b30baf9a4e78618f04b03de9b97365cb137815a) fixes a bug introduced in #29662

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -203,7 +203,7 @@ jobs:
           path: |
             unexpected-test-wpt.log
       - name: Comment on PR with results
-        run: etc/ci/report_aggregated_expected_results.py wpt-filtered-results-linux/*.json
+        run: etc/ci/report_aggregated_expected_results.py --tag="linux-wpt-${{ env.LAYOUT }}" wpt-filtered-results-linux/*.json
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -102,7 +102,7 @@ jobs:
           path: target.tar.gz
 
   linux-wpt:
-    if: ${{ contains(github.ref_name, 'wpt') || inputs.wpt }}
+    if: ${{ github.ref_name == 'try-wpt' || github.ref_name == 'try-wpt-2020' || inputs.wpt }}
     name: Linux WPT Tests
     runs-on: ubuntu-20.04
     needs: ["build-linux"]
@@ -183,7 +183,7 @@ jobs:
   report-test-results:
     name: Reporting test results
     runs-on: ubuntu-latest
-    if: ${{ always() && !cancelled() && success('build-linux') && (contains(github.ref_name, 'wpt') || inputs.wpt == 'test') }}
+    if: ${{ always() && !cancelled() && success('build-linux') && (github.ref_name == 'try-wpt' || github.ref_name == 'try-wpt-2020' || inputs.wpt == 'test') }}
     needs:
       - "linux-wpt"
     steps:
--- a/etc/ci/report_aggregated_expected_results.py
+++ b/etc/ci/report_aggregated_expected_results.py
@@ -17,7 +17,7 @@ import json
 import os
 import re
 import subprocess
-import sys
+import argparse
 import textwrap
 import xml.etree.ElementTree as ElementTree
 
@@ -99,9 +99,9 @@ class Item:
         return result
 
 
-def get_results() -> Optional[Item]:
+def get_results(filenames: list[str], tag: str = "") -> Optional[Item]:
     unexpected = []
-    for filename in sys.argv[1:]:
+    for filename in filenames:
         with open(filename, encoding="utf-8") as file:
             unexpected += json.load(file)
     unexpected.sort(key=lambda result: result["path"])
@@ -129,10 +129,13 @@ def get_results() -> Optional[Item]:
                  "Stable unexpected results")
 
     run_url = get_github_run_url()
+    text = "Test results"
+    if tag:
+        text += f" for {tag}"
+    text += " from try job"
     if run_url:
-        text = f"Results from try job ({run_url}):"
-    else:
-        text = "Results from try job:"
+        text += f" ({run_url})"
+    text += ":"
     return Item(text, "", children) if children else None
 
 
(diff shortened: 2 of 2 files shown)


etc/ci/report_aggregated_expected_results.py
```diff
@@ -177,18 +177,19 @@ def create_check_run(body: str):
...
         'started_at': datetime.utcnow().replace(microsecond=0).isoformat() + "Z",
         'conclusion': conclusion,
         'completed_at': datetime.utcnow().replace(microsecond=0).isoformat() + "Z",
         'output': {
-            'title': 'Aggregated WPT report',
+            'title': f'Aggregated {wpt} report',
             'summary': body,
             # 'text': '',
-            'images': [{'alt': 'WPT logo', 'image_url': 'https://avatars.githubusercontent.com/u/37226233'}]
+            # 'images': [{'alt': 'WPT logo', 'image_url': 'https://avatars.githubusercontent.com/u/37226233'}]
```
mrobinson, line 192: Why are you commenting out the image here?

etc/ci/report_aggregated_expected_results.py
```diff
@@ -205,7 +206,14 @@ def create_check_run(body: str):
 
 
 def main():
-    results = get_results()
+    parser = argparse.ArgumentParser()
```
mrobinson, line 209: Instead of adding `--layout-2020` and `--layout-2013` arguments here, maybe we could add a `--tag` argument which is a string type. It could have a default value which would allow using the script more easily from the command-line.

```python
parser.add_argument("--tag", default="", action="store_true",
                                        help="A string tag used to distinguish the results.")
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -99,9 +99,9 @@ def to_html(self, level: int = 0) -> ElementTree.Element:
         return result
 
 
-def get_results() -> Optional[Item]:
+def get_results(free_args: list[str], layout: str = "") -> Optional[Item]:
```
mrobinson, line 102:
```suggestion
def get_results(filenames: list[str], tag: str = "") -> Optional[Item]:
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -99,9 +99,9 @@ def to_html(self, level: int = 0) -> ElementTree.Element:
...
     unexpected = []
-    for filename in sys.argv[1:]:
+    for filename in free_args:
```
mrobinson, line 104:
```suggestion
    for filename in filenames:
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -130,9 +130,9 @@ def add_children(children, results, filter_func, text):
 
     run_url = get_github_run_url()
     if run_url:
-        text = f"Results from try job ({run_url}):"
+        text = f"Results for linux-wpt{layout} from try job ({run_url}):"
     else:
-        text = "Results from try job:"
+        text = f"Results for linux-wpt{layout} from try job:"
```
mrobinson, line 135: I suggest this block should be:
```python
text = "Test results"
if tag:
    text += f" for {tag}"
if run_url:
    text += f" ({run_url})"
text += ":"
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -158,7 +158,7 @@ def get_pr_number() -> Optional[str]:
     return match.group(1) if match else None
 
 
-def create_check_run(body: str):
+def create_check_run(body: str, layout2020: bool):
```
mrobinson, line 161:
```suggestion
def create_check_run(body: str, tag: str):
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -177,18 +177,19 @@ def create_check_run(body: str):
...
-        'name': 'WPT',
+        'name': wpt,
         'head_sha': github_context["sha"],
         'status': 'completed',
         'started_at': datetime.utcnow().replace(microsecond=0).isoformat() + "Z",
         'conclusion': conclusion,
         'completed_at': datetime.utcnow().replace(microsecond=0).isoformat() + "Z",
         'output': {
-            'title': 'Aggregated WPT report',
+            'title': f'Aggregated {wpt} report',
```
mrobinson, line 189:
```suggestion
            'title': f'Aggregated {tag} report',
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -205,7 +206,14 @@ def create_check_run(body: str):
...
 def main():
-    results = get_results()
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--layout-2020", "--with-layout-2020", default=False, action="store_true",
+                        help="Report results for the 2020 layout engine")
+    parser.add_argument("--layout-2013", "--with-layout-2013", default=True, action="store_true",
+                        help="Report results for the 2013 layout engine")
+    args, free_args = parser.parse_known_args()
+    layout2020: bool = vars(args)['layout_2020']
+    results = get_results(free_args, '-2020' if layout2020 else '')
```
mrobinson, line 216:
```suggestion
    results = get_results(free_args, args.tag)
```

.github/workflows/linux.yml
```diff
@@ -203,7 +203,7 @@ jobs:
           path: |
             unexpected-test-wpt.log
       - name: Comment on PR with results
-        run: etc/ci/report_aggregated_expected_results.py wpt-filtered-results-linux/*.json
+        run: etc/ci/report_aggregated_expected_results.py --with-${{ env.LAYOUT }} wpt-filtered-results-linux/*.json
```
mrobinson, line 206:
```suggestion
        run: etc/ci/report_aggregated_expected_results.py --tag="${{ env.LAYOUT }}" wpt-filtered-results-linux/*.json
```

mrobinson (changes requested):
Thanks for this fix.

etc/ci/report_aggregated_expected_results.py
sagudev: I used `--layout-2013` and `--layout-2020` to be inlined with other mach tools and make it easier to implelemt it in GitHub actions, but tag would be more futureproof for mac/win wpt runs we could have in the future.

mrobinson (approved):

PR #29673, opened 2023-04-27
Add auto wpt option to workflow

sagudev:
Introduces auto wpt option in Linux workflow, which runs wpt test for layout-engine based on file changes.

So now we have 3 wpt options:
- `test` unconditionally runs wpt tests
- `auto` runs wpt tests only if there are changes in layout folders (does are determined by filters based on selected layout)
- `sync` unconditionally runs wpt tests and sync them

In main workflow layout2020 is in `test` mode (run on every change), layout2013 is in `auto` mode (wpt test are run only if changes are present in layout2013 folders), other `try-*` builds would still be working as expected.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,7 +29,7 @@ on:
         default: "test"
         required: false
         type: choice
-        options: ["test", "sync"]
+        options: ["test", "sync", "auto"]
       unit-tests:
         required: false
         default: false
@@ -55,6 +55,12 @@ jobs:
   build-linux:
     name: Build
     runs-on: ubuntu-20.04
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      layout2020: ${{ steps.filter.outputs.layout2020 }}
+      layout2013: ${{ steps.filter.outputs.layout2013 }}
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -100,9 +106,28 @@ jobs:
...
+          filters: |
+            layout2020:
+              - 'components/layout/**'
+              - 'components/layout_2020/**'
+              - 'components/layout_thread_2020/**'
+              - '**-layout-2020/**'
+            layout2013:
+              - 'components/layout/**'
+              - 'components/layout_2013/**'
+              - 'components/layout_thread_2013/**'
```
sagudev, line 123: This filter wildcards is the most crucial part of the PR.

mrobinson: Hrm. This is a bit suspicious because there is no `layout_2013` and `layout_thread_2013` directories. The `layout` and `layout_thread` directories correspond to Layout 2013.

sagudev: Good catch.

mrobinson:
Thank you for doing this work. We've talked this and we feel like since we are doing a lot more effort on Layout 2020, we would like to always run the WPT suite in the main workflow for Layout 2020. See #29746.

sagudev:
The only change that was needed was setting layout2013 to auto and layout2020 to test, if we still want to have any test for layout2013. I also updated PR description.

mrobinson:
Apologies, I was not super clear in my last comment. For now, we'd like to try running WPT for both layout 2013 and layout 2020 in the main workflow.

sagudev:
I will keep this PR open, so when it is decided  not to run layout2013 wpt tests anymore, then this can be merged, to keep layout2013 in shape until, it gets completely removed.

mrobinson:
I've been thinking a bit more about this and it's a really tricky change. Here are some examples of why:

 - Changes to `components/script` and really any directory in components can affect WPT test results.
 - Changes to Python code in `python` can alter the way that the test harness runs and affect WPT results.
 - Changes to any of the third-party dependencies in `third_party` or Cargo.lock can affect layout, painting, anything.
 - Changes to any of the tests or metadata can affect whether the WPT tests pass.
 - Changes to the workflow files in `.github` can break WPT tests (as I have done on multiple occasion :sweat_smile:)

I think the number directories that we can change without wanting to run WPT tests are actually quite small. I do not know if it's possible to detect this as Servo currently is. Perhaps we could skip certain platforms if we only modify platform-specific code...but I'm not certain even that is possible.

sagudev:
Auto option wasn't meant to be used on layout2020, it is meant to provide best-effort CI runs for layout2013, when we want to deprecate layout2013 even further, kicking it out of main CI, but not removing from tree.

mrobinson:
This would first require switching the main bulk of tests over to be run via layout 2020.

Is the idea that this would run as the main workflow or that people would use a try=auto option?

sagudev:
This was planned to be run as part of main workflow. try=wpt would still run layout2013 no matter what. Given the recent changes we could also make parameter for normal try build to force layout2013 wpt test.

mrobinson:
> This was planned to be run as part of main workflow. try=wpt would still run layout2013 no matter what. Given the recent changes we could also make parameter for normal try build to force layout2013 wpt test.

I think we must always run all WPT tests when landing changes, but perhaps try could skip jobs.

sagudev:
Closing this, as we already stopped running layout 2013 by default (WPT imports keep expectations somehow updated) and main has diverged very much from when this was written (complete CI revamp for try runs).

Issue #29675, opened 2023-04-27
Text with font-size:0 should have 0 width


Loirooriol:
```html
<!DOCTYPE html>
[<span style="font-size: 0px">This line has a zero sized font.</span>]
```

It should be rendered as `[]`, but it looks like `[    ]`.

Both in layout-2013 and layout-2020.

Loirooriol:
This was oddly fixed by #32278. Should add a test, though.

PR #29676, opened 2023-04-27
Round text decoration thickness to >= 1 device pixels

### LINKED ISSUE #29668: text-decoration sometimes not painted
For example:

```html
<!DOCTYPE html>
<div style="text-decoration: underline">lorem</div>
<div style="text-decoration: underline">lorem</div>
```

Debugging I can see that the underline thickness is 0.7px, so I guess that depending on how it aligns withing the grid of pixels it may end up painted or not.

Maybe the thickness should be floored at 1px or something?

Loirooriol:
Previously the thickness coming from font metrics could be something like 0.7px, so with 1dppx it would be painted as either 1 or 0 device pixels.

Enforcing at least 1 device pixel ensures that the decoration will be visible, and rounding to an integral amount of device pixels ensures that the thickness won't vary depending on the position.

The specification requires this behavior when text-decoration-thickness is set to a length or percentage. It's not clear if it should also happen by default, but this seems to match other browsers (except for WebKit rounding up instead of to the nearest integer).

The test text-decoration-thickness-from-zero-sized-font.html is now failing because of #29675.

---

- [X] These changes fix #29668 (GitHub issue number if applicable)

- [X] These changes do not require tests because the exact appearance of text decorations is UA-dependent, so testing doesn't seem much feasible.

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -212,6 +212,11 @@ impl Fragment {
 
         let color = fragment.parent_style.clone_color();
         let font_metrics = &fragment.font_metrics;
+        let dppx = builder.context.style_context.device_pixel_ratio().get();
+        let round_to_nearest_device_pixel = |value: Length| -> Length {
+            // Round to the nearest integer device pixel, ensuring at least one device pixel.
+            Length::new((value.px() * dppx).round().max(1.0) / dppx)
+        };
 
         // Underline.
         if fragment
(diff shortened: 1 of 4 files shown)


components/layout_2020/display_list/mod.rs
```diff
@@ -212,6 +212,11 @@ impl Fragment {
 
         let color = fragment.parent_style.clone_color();
         let font_metrics = &fragment.font_metrics;
+        let dppx = builder.context.style_context.device_pixel_ratio().get();
+        let adjust_thickness = |thickness: Length| -> Length {
```
mrobinson, line 216: I think it might be better to call this `round_to_nearest_device_pixel` here and above.

mrobinson (approved):
Looks good to me with a minor nit.

components/layout_2020/display_list/mod.rs
Loirooriol: Done, also renamed the parameter to a more generic `value`.

mrobinson (approved):
Thank you!

PR #29677, opened 2023-04-27
Better implement getComputedStyle() for positioned insets 

mrobinson:
The specification dictates quite quite idiosyncratic return values when
querying insets of positioned elements via `getComputedStyle()`. These
depend on whether or not the element's size was overconstrained. This
change adds a better implementation of that in preparation for returning
proper values for `position: sticky`.

There are two changes here:
- The containing block used in `FragmentTree::find` is properly computed using the `ContainingBlockManager` data structure. This data structure is now updated during traversal through the fragment tree in `Fragment::find`.
- We now take into account whether or not a absolute elements are overconstrained when calculating their resolved insets. For absolutely positioned elements, this happens during absolute positioning. For relatively positioned elements, we can determine this when a `BoxFragment` is constructed.

---


mrobinson:
@delan Okay, this one should be rebased and ready to review when you have a moment. Thank you!

delan:
Thanks for the thorough description and comments!

@bors-servo r+

mrobinson:
@bors-servo r=delan

Looks like I mistakenly applied one of the expectation updates to the layout 2013 directory.

PR #29680, opened 2023-04-28
Refactor scrolling on the window object

### LINKED ISSUE #21321: Intermittent failure in /_mozilla/mozilla/scrollBy.html
```
  ▶ Unexpected subtest result in /_mozilla/mozilla/scrollBy.html:
  │ FAIL [expected PASS] Ensure that the window.scrollBy function affects scroll position as expected
  │   → assert_equals: expected Element node <html><head><meta charset="utf-8">
<title>Ensure that the... but got Element node <a id="link" href="http://mozilla.org">This is some link ...
  │ 
  │ @http://web-platform.test:8000/_mozilla/mozilla/scrollBy.html:21:9
  │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:1538:20
  │ test@http://web-platform.test:8000/resources/testharness.js:545:9
  └ @http://web-platform.test:8000/_mozilla/mozilla/scrollBy.html:10:5

```

mrobinson:
Refactor the scrolling and scrollable area calculation on the window object, to make it better match the specification. This has some mild changes to behavior, but in general things work the same as they did before. This is mainly preparation for properly handling viewport propagation of the `overflow` property.
    
There is one new pass in Layout 2020 regarding `position: sticky`, but this isn't a big deal because there is no support for `position: sticky` in Layout 2020 yet. In addition, `tests/wpt/mozilla/tests/mozilla/scrollTo.html` is updated to fix a flakiness timing issue. It tries to scroll the page before the entire page is loaded.
    
---

- [x] These changes fix #21321.

### DIFF
--- a/components/layout_2020/fragment_tree/fragment.rs
+++ b/components/layout_2020/fragment_tree/fragment.rs
@@ -167,6 +167,15 @@ impl Fragment {
         }
     }
 
+    pub fn scrolling_area(&self, containing_block: &PhysicalRect<Length>) -> PhysicalRect<Length> {
+        match self {
+            Fragment::Box(fragment) | Fragment::Float(fragment) => fragment
+                .scrollable_overflow(containing_block)
+                .translate(containing_block.origin.to_vector()),
+            _ => self.scrollable_overflow(containing_block),
+        }
+    }
+
     pub fn scrollable_overflow(
         &self,
         containing_block: &PhysicalRect<Length>,
--- a/components/layout/query.rs
+++ b/components/layout/query.rs
@@ -76,7 +76,7 @@ pub struct LayoutThreadData {
     pub scroll_id_response: Option<ExternalScrollId>,
 
     /// A queued response for the scroll {top, left, width, height} of a node in pixels.
-    pub scroll_area_response: Rect<i32>,
+    pub scrolling_area_response: Rect<i32>,
 
     /// A queued response for the resolved style property of an element.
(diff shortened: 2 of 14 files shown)


mrobinson:
cc @atbrakhi

mrobinson:
This needs a bit more work. In particular, we need to properly handle overflow: scroll on the `<body>` element. It should be propagated to the root `<html>` element. I've started that work here: #29699.

components/layout_2020/fragment_tree/fragment.rs
```diff
@@ -178,6 +178,15 @@ impl Fragment {
         }
     }
 
+    pub fn scrolling_area(&self, containing_block: &PhysicalRect<Length>) -> PhysicalRect<Length> {
+        match self {
+            Fragment::Box(fragment) | Fragment::Float(fragment) => fragment
+                .scrollable_overflow(&containing_block)
```
mukilan, line 184: Do we need to re-borrow `containing_block` here? `fn scrollable_overflow` takes a `&PhysicalRect`, so I think we should be able to pass `containing_block` directly?

mukilan (approved):
Nice!

components/layout_2020/fragment_tree/fragment.rs
mrobinson: Nope. You're right. I've removed this.

mrobinson:
Thanks for the review!

PR #29686, opened 2023-04-29
Normalize corner radii that add up to > 100%

### LINKED ISSUE #29685: inner_radii needs to normalize radii to avoid overlapping curves
`inner_radii` is very naive:

https://github.com/servo/servo/blob/4339b3bab494e477b502f82137bd360a963ec2be/components/layout_2020/display_list/mod.rs#L669-L682

For starters, this can easily produce negative radii, which seems a code smell, though webrender seems to handle it well.

But then the main problem are overlapping radii: https://drafts.csswg.org/css-backgrounds-3/#corner-overlap

> Corner curves must not overlap: When the sum of any two adjacent border radii exceeds the size of the border box, UAs must proportionally reduce the used values of all border radii until none of them overlap.

For example:

```html
<!DOCTYPE html>
<div style="
  width: 100px;
  height: 100px;
  border: 10px solid blue;
  border-radius: 100px 50px 50px 50px / 300px 50px 50px 300px;
  background: linear-gradient(cyan, cyan) padding-box, red border-box;
"></div>
```

The border area is 120x120, so `border-radius` behaves like `20px 10px 10px 10px / 60px 10px 10px 60px`.
Then when we shrink by 10px, the padding should use `10px 0 0 0 / 50px 0 0 50px`.
However, since `inner_radii` doesn't normalize, it produces `90px 40px 40px 40px / 290px 40px 40px 290px`.
When normalized to fit into 100x100, that becomes `50px calc(200px/9) calc(200px/9) calc(200px/9) / calc(1450px/9) calc(200px/9) calc(200px/9) calc(1450px/9)`, which is considerably different:

There should be no red.

Loirooriol:
WebRender already seems to be doing this normalization, but this is needed by inner_radii in order to properly compute the reduced radii when background-clip is content-box or padding-box.

This will also be needed for expanding the radii for box-shadow or outline.

Test: css/css-backgrounds/background-rounded-image-clip-002.html

---

- [X] These changes fix #29685 (GitHub issue number if applicable)


mrobinson (approved):

components/layout_2020/display_list/mod.rs
```diff
@@ -322,12 +322,25 @@ impl<'a> BuilderForBoxFragment<'a> {
                 )
             };
             let b = fragment.style.get_border();
-            wr::BorderRadius {
+            let mut r = wr::BorderRadius {
```
mrobinson, line 325: Let's call this `radius`.

components/layout_2020/display_list/mod.rs
```diff
@@ -322,12 +322,25 @@ impl<'a> BuilderForBoxFragment<'a> {
...
                 top_left: corner(&b.border_top_left_radius),
                 top_right: corner(&b.border_top_right_radius),
                 bottom_right: corner(&b.border_bottom_right_radius),
                 bottom_left: corner(&b.border_bottom_left_radius),
+            };
+            // Normalize radii that add up to > 100%.
```
mrobinson, line 331: Maybe paste the specification text so it's clear where this calculation comes from:

> Let f = min(Li/Si), where i ∈ {top, right, bottom, left}, Si is the sum of the two corresponding radii of the corners on side i, and Ltop = Lbottom = the width of the box, and Lleft = Lright = the height of the box. If f < 1, then all corner radii are reduced by multiplying them by f.

mrobinson (approved):

mrobinson:
Nice fix!

Loirooriol:
@bors-servo try=wpt-2020
Let's see if I can reduce the amount of allowed bad pixels in the test.

Issue #27212, opened 2020-07-08
Implement console.table


jdm:
This is used by cables.gl content.

jdm:
A really simple implementation that can go in a userscript:
```
console.table = (data) => {
	for (const datum of data) {
		console.log(datum.join('\t'));
	}
}
```

switchpiggy:
If this issue is still valid and an implementation is still needed, I can work on this.

PR #29689, opened 2023-04-30
Added basic console.table() implementation

switchpiggy:
Implemented a basic version of console.table() that supports arrays of primitives as input. The full functionality as described in the spec as mentioned in #27212 will also be implemented later.

---

- [x] These changes do not require tests because this is only an initial implementation, and tests will be added later if needed.

### DIFF
--- a/components/script/dom/console.rs
+++ b/components/script/dom/console.rs
@@ -73,23 +73,25 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
     })
 }
 
-///Generates matrix of table elements, and then converts matrix into a properly formatted DOMString
-///TODO: Implement generate_table for objects and arrays of objects
+/// Generates matrix of table elements, and then converts matrix into a properly formatted DOMString
+// TODO: Implement generate_table for objects and arrays of objects
 #[allow(unsafe_code)]
-unsafe fn generate_table(cx: *mut JSContext, tabular_data: &HandleValue) -> Option<DOMString> {
-    if !tabular_data.get().is_object() || !is_array_like(cx, *tabular_data) {
+fn generate_table(cx: *mut JSContext, tabular_data: &HandleValue) -> Option<DOMString> {
+    let arr_like: bool = unsafe { is_array_like(cx, *tabular_data) };
+    if !tabular_data.get().is_object() || !arr_like {
         return None;
     }
 
-    let vec: Vec<DOMString> =
+    let vec: Vec<DOMString> = unsafe {
         Vec::<DOMString>::from_jsval(cx, *tabular_data, StringificationBehavior::Empty)
             .expect("Unable to convert Array object to Vec<DOMString>")
             .get_success_value()
             .unwrap_or(&Vec::new())
-            .to_vec();
+            .to_vec()
+    };
 
     let mut table_matrix: Vec<Vec<DOMString>> = Vec::new();
-    //Only two columns because we don't expect objects as input yet
+    // Only two columns because we don't expect objects as input yet
     table_matrix.push(vec!["(index)".into(), "Values".into()]);
     for (index, word) in vec.into_iter().enumerate() {
         table_matrix.push(vec![index.to_string().into(), word]);
@@ -134,8 +136,7 @@ impl Console {
     ) {
         let cx_ptr: *mut JSContext = *cx as *mut JSContext;
 
-        #[allow(unsafe_code)]
-        let table: Option<DOMString> = unsafe { generate_table(cx_ptr, &tabular_data) };
+        let table: Option<DOMString> = generate_table(cx_ptr, &tabular_data);
 
         match table {
             Some(table_string) => console_messages(global, &[table_string], LogLevel::Log),
--- a/components/script/dom/console.rs
+++ b/components/script/dom/console.rs
@@ -8,7 +8,9 @@ use crate::dom::globalscope::GlobalScope;
 use crate::dom::workerglobalscope::WorkerGlobalScope;
 use devtools_traits::{ConsoleMessage, LogLevel, ScriptToDevtoolsControlMsg};
 use js::rust::describe_scripted_caller;
-use std::io;
+use std::{io};
+use std::any::Any;
+use std::option::Option;
 
 // https://developer.mozilla.org/en-US/docs/Web/API/Console
 pub struct Console(());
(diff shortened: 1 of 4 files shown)


components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
 
+    //https://console.spec.whatwg.org/#table
+    #[allow(unsafe_code)]
+    pub fn Table(
+        cx: SafeJSContext,
+        global: &GlobalScope,
+        tabular_data: HandleValue,
+        properties: Option<Vec<DOMString>>,
+    ) {
+        unsafe {
```
mrobinson, line 139: This `unsafe` block is very wide. Try to move it down to exactly the problematic code and then let's see if we can eliminate it entirely.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+
 #[allow(non_snake_case)]
 impl Console {
     // https://developer.mozilla.org/en-US/docs/Web/API/Console/log
     pub fn Log(global: &GlobalScope, messages: Vec<DOMString>) {
         console_messages(global, &messages, LogLevel::Log)
     }
 
+    //https://console.spec.whatwg.org/#table
+    #[allow(unsafe_code)]
```
mrobinson, line 132: This should move to an interior function.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+    tabular_data: &HandleValue,
+    _properties: &Option<Vec<DOMString>>,
+) -> Option<Vec<Vec<DOMString>>> {
+    let cx_ptr = *(cx.deref()) as *mut JSContext;
+
+    if !tabular_data.get().is_object() {
+        return None;
+    }
+
+    if is_array_like(cx_ptr, *tabular_data) {
```
mrobinson, line 89: I think this might be better as an early return.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
     })
 }
 
+//generate matrix of table elements
```
mrobinson, line 76: Documentation of functions in rust starts with three "/" and should be written with complete sentences. This allows the documentation to show up in IDEs and also be generated properly on the documentation site.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+#[allow(unsafe_code)]
+unsafe fn generate_table(
+    cx: &SafeJSContext,
+    tabular_data: &HandleValue,
+    _properties: &Option<Vec<DOMString>>,
```
mrobinson, line 81: If this is unused, simply remove it.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+}
+
 #[allow(non_snake_case)]
 impl Console {
     // https://developer.mozilla.org/en-US/docs/Web/API/Console/log
     pub fn Log(global: &GlobalScope, messages: Vec<DOMString>) {
         console_messages(global, &messages, LogLevel::Log)
     }
 
+    //https://console.spec.whatwg.org/#table
```
mrobinson, line 131: Nit: missing a space after "//".

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+        }
+
+        return Some(res);
+    }
+
+    None
+}
+
+//transform matrix of table elements into DOMString for logging
+fn generate_table_as_string(table: &[Vec<DOMString>]) -> DOMString {
```
mrobinson, line 110: I think it makes more sense to put this into the body of `generate_table` and return an Option<DOMString>`.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+
+    if is_array_like(cx_ptr, *tabular_data) {
+        let vec: Vec<DOMString> =
+            Vec::<DOMString>::from_jsval(cx_ptr, *tabular_data, StringificationBehavior::Empty)
+                .expect("Error when converting Array object to Vec<DOMString>")
+                .get_success_value()
+                .unwrap_or(&Vec::new())
+                .to_vec();
+
+        let mut res: Vec<Vec<DOMString>> = Vec::new();
```
mrobinson, line 97: Let's use full words as variable names when it isn't too difficult, so `res` should be `result`.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+    #[allow(unsafe_code)]
+    pub fn Table(
+        cx: SafeJSContext,
+        global: &GlobalScope,
+        tabular_data: HandleValue,
+        properties: Option<Vec<DOMString>>,
+    ) {
+        unsafe {
+            match generate_table(&cx, &tabular_data, &properties) {
+                Some(TableMatrix) => {
```
mrobinson, line 141: Variable names in Servo should be `snake_case`, so `TableMatrix` should be `table_matrix`.

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+                .expect("Error when converting Array object to Vec<DOMString>")
+                .get_success_value()
+                .unwrap_or(&Vec::new())
+                .to_vec();
+
+        let mut res: Vec<Vec<DOMString>> = Vec::new();
+        res.push(vec!["(index)".into(), "Values".into()]);
+        for (index, word) in vec.into_iter().enumerate() {
+            res.push(vec![index.to_string().into(), word]);
+        }
```
mrobinson, line 101: Does this implementation handle objects?

components/script/dom/console.rs
```diff
@@ -65,13 +73,80 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
...
+    None
+}
+
+//transform matrix of table elements into DOMString for logging
+fn generate_table_as_string(table: &[Vec<DOMString>]) -> DOMString {
+    let mut res: DOMString = DOMString::new();
+    for row in table.iter() {
+        for col in row.iter() {
+            res.push_str(col as &str);
+            res.push_str("\t\t");
```
mrobinson, line 115: If we implement this, it deserves an implementation that knows how to properly align the columns of the table. Two tabs are pretty brittle, so maybe investigate if there is a more resilient way to do this based on the actual strings that we are processing.

mrobinson (changes requested):
Thank you for your contribution. This is a nice start. It needs a bit of work before it's ready to merge.

components/script/dom/console.rs
switchpiggy: The implementation doesn't handle objects. It just prints [object Object] when given an object, but I don't know how to retrieve the attributes of a JSObject and the documentation isn't helping much. Do you have any suggestions?

components/script/dom/console.rs
switchpiggy: That's a good idea. I'll see if there's a good way to align the columns.

components/script/dom/console.rs
switchpiggy: The problematic part is calling the generate_table() function, which is an unsafe function because it calls Vec<T>::from_jsval() and is_array_like(). I'm not very experienced with the servo documentation, so there could be a way to implement generate_table() safely which I'm not aware of.

components/script/dom/console.rs
switchpiggy: I've implemented a simple way to align the columns by setting each column's width to be the maximum length of all of its rows, plus some extra space.

components/script/dom/console.rs
mrobinson: At the very least here, I think you should reduce the unsafe block to enclose only the unsafe operations instead of marking the entire function as unsafe.

components/script/dom/console.rs
```diff
@@ -65,13 +73,76 @@ fn console_message(global: &GlobalScope, message: DOMString, level: LogLevel) {
     })
 }
 
+///Generates matrix of table elements, and then converts matrix into a properly formatted DOMString
```
mrobinson, line 76: When making a comment in rust you should put a space after the slashes. In this case, I don't think the rustdoc will work without the space.

components/script/dom/console.rs
switchpiggy: Okay, sure! I've reduced the unsafe block to only the call to generate_table().

switchpiggy: Actually, I've removed the unsafe blocked around generate_table() and just restricted the unsafe blocks to the unsafe operations inside the function, like you said.

components/script/dom/console.rs
switchpiggy: Oh ok sure! Thanks for the suggestion, I'm not really familiar with rustdoc so that really helps. I'll make sure to fix the comment.

yerke:
Hi @switchpiggy. Thanks for your work on this PR. Are you interested in continuing to work on it, or is it ok if somebody else finishes it?

mrobinson:
I think this one has gotten stale unfortunately, so I will close it.

PR #29699, opened 2023-05-02
Detect body elements during layout

mrobinson:
During layout it is often useful, for various specification reasons, to know if an element is the `<body>` element of an `<html>` element root. There are a couple places where a brittle heuristic is used to detect `<body>` elements. This information is going to be even more important to properly handle `<html>` elements that inherit their overflow property from their `<body>` children.

Implementing this properly requires updating the DOM wrapper interface. This check does reach up to the parent of thread-safe nodes, but this is essentially the same kind of operation that `parent_style()` does, so is ostensibly safe.

This change should not change any behavior and is just a preparation step for properly handle `<body>` overflow.

---

- [x] These changes do not require tests because it does not change behavior.

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -5,6 +5,7 @@
 use crate::cell::ArcRefCell;
 use crate::context::LayoutContext;
 use crate::element_data::{LayoutBox, LayoutDataForElement};
+use crate::fragment_tree::{BaseFragmentInfo, FragmentFlags, Tag};
 use crate::geom::PhysicalSize;
 use crate::replaced::{CanvasInfo, CanvasSource, ReplacedContent};
 use crate::style_ext::{Display, DisplayGeneratingBox, DisplayInside, DisplayOutside};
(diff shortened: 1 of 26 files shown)


components/layout_2020/fragment_tree/base.rs
```diff
@@ -0,0 +1,117 @@
...
+/// a particular layout object.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub(crate) struct Tag {
+    pub(crate) node: OpaqueNode,
+    pub(crate) pseudo: Option<PseudoElement>,
+}
+
+impl Tag {
+    /// Create a new Tag for a non-pseudo element. This is mainly used for
+    ///  matching existing tags, since it does not accept an `info` argument.
```
mukilan, line 93: nit: Additional space at the beginning can be removed.

components/layout_2020/fragment_tree/base.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub(crate) fn new(node: OpaqueNode) -> Self {
+        Tag { node, pseudo: None }
+    }
+
+    /// Create a new Tag for a pseudo element. This is mainly used for
+    ///  matching existing tags, since it does not accept an `info` argument.
```
mukilan, line 99: nit: Additional space at the beginning can be removed.

components/layout_2020/dom_traversal.rs
```diff
@@ -74,6 +75,31 @@ impl<Node: Clone> NodeAndStyleInfo<Node> {
     }
 }
 
+impl<'dom, Node> Into<BaseFragmentInfo> for &NodeAndStyleInfo<Node>
```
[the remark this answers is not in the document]
mrobinson: Great point I've converted this and the other `Into` implementation into `From`.

components/layout_2020/fragment_tree/base.rs
mrobinson: Fixed!

components/layout_2020/fragment_tree/base.rs
mrobinson: Fixed!

mukilan (approved):

PR #29703, opened 2023-05-03
Update step numbers in URL constructor

ohno418:
Two commits have been made ahead of the current implementation of
Servo's URL constructor:

- Align with a more modern IDL definition style [1]
- Add URL.canParse() [2]

Since these commits don't alter the actual behavior, this commit only
updates the step numbers and adds brief descriptions for each step.

No behavior change is expected with this commit.

[1]: https://github.com/whatwg/url/commit/ea3b75d333cbce91f73676ef2d21bcadb12b4419
[2]: https://github.com/whatwg/url/commit/ae3c28b84e3e7122c2807401c26b8a63cb2ab445

---

- [x] These changes do not require tests because no behavior has changed.

### DIFF
--- a/components/script/dom/url.rs
+++ b/components/script/dom/url.rs
@@ -77,38 +77,38 @@ impl URL {
         url: USVString,
         base: Option<USVString>,
     ) -> Fallible<DomRoot<URL>> {
+        // Step 1. Parse url with base.
         let parsed_base = match base {
-            None => {
-                // Step 1.
-                None
-            },
-            Some(base) =>
-            // Step 2.1.
-            {
+            None => None,
+            Some(base) => {
                 match ServoUrl::parse(&base.0) {
                     Ok(base) => Some(base),
                     Err(error) => {
-                        // Step 2.2.
+                        // Step 2. Throw a TypeError if URL parsing fails.
                         return Err(Error::Type(format!("could not parse base: {}", error)));
                     },
                 }
             },
         };
-        // Step 3.
         let parsed_url = match ServoUrl::parse_with_base(parsed_base.as_ref(), &url.0) {
             Ok(url) => url,
             Err(error) => {
-                // Step 4.
+                // Step 2. Throw a TypeError if URL parsing fails.
                 return Err(Error::Type(format!("could not parse URL: {}", error)));
             },
         };
-        // Step 5: Skip (see step 8 below).
-        // Steps 6-7.
-        let result = URL::new(global, parsed_url);
-        // Step 8: Instead of construcing a new `URLSearchParams` object here, construct it
-        //         on-demand inside `URL::SearchParams`.
-        // Step 9.
-        Ok(result)
+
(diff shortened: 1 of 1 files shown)


components/script/dom/url.rs
```diff
@@ -77,38 +77,34 @@ impl URL {
...
+                    // Step 2
+                    return Err(Error::Type(format!("could not parse URL: {}", error)));
+                },
+            }
         };
-        // Step 5: Skip (see step 8 below).
-        // Steps 6-7.
-        let result = URL::new(global, parsed_url);
-        // Step 8: Instead of construcing a new `URLSearchParams` object here, construct it
-        //         on-demand inside `URL::SearchParams`.
```
ohno418, line 109: I don't know exactly why this happened, but this PR doesn't change the behavior around `URLSearchParams`, continuing to ignore the related steps.

mrobinson: Is this change just updating the step numbers?

ohno418: Yes. The corresponding spec change related to this PR can be found here: https://github.com/whatwg/url/commit/ae3c28b8

(I probably should have mentioned this somewhere, such as in commit message.)

mrobinson: I'm not sure the code changes are an improvement here, because they increase the nesting level of the method. Would it be possible to avoid changing the code if there is no behavior changes and also to paste a small description of each step in quotes. With a small amount of text, at least it will still be clear what each step does.

ohno418: Okay, that makes sense. I'll make the change.

ohno418:
As I mentioned in the commit message, the current implementation status of the URL constructor appears to correspond to the state before the following commit of the spec:

https://github.com/whatwg/url/commit/ea3b75d333cbce91f73676ef2d21bcadb12b4419

components/script/dom/url.rs
```diff
@@ -77,38 +77,38 @@ impl URL {
...
-        Ok(result)
+
+        // Skip the steps below.
+        // Instead of construcing a new `URLSearchParams` object here, construct it
+        // on-demand inside `URL::SearchParams`.
+        //
+        // Step 3. Let query be parsedURL’s query.
+        // Step 5. Set this’s query object to a new URLSearchParams object.
+        // Step 6. Initialize this’s query object with query.
+        // Step 7. Set this’s query object’s URL object to this.
```
ohno418, line 108: This is a bit confusing. Step 5 and 8, which were previously skipped, seem to correspond these four steps.

Step 8 is split into some parts here:
https://github.com/whatwg/url/commit/ea3b75d333cbce91f73676ef2d21bcadb12b4419#diff-29243b3b9b716b55c6a61970b0c4864f464b139d397fb961a05bb6e1e2b97cabL2840-L2842

Here's the current spec:
https://url.spec.whatwg.org/#constructors

PR #29706, opened 2023-05-03
Fix scrolling on root element

mrobinson:
eca0acf4598c173c71765b961bd2079c0bb48cd2 uncovered a bug in the way that the scrolling area of `window` was calculated and broke scrolling on the root element. This change does two things in order to fix that:

1. Does a partial revert of eca0acf4598c173c71765b961bd2079c0bb48cd2 in order to get scrolling from script working again on the window object.
2. Has the compositor always generate a frame for scrolls starting from script and waits for them. This is speculative fix for flakiness in root scrolling tests on CI.

The changes to the compositor expose some hidden failures, which is why some tests are marked as failing now.

---

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -490,7 +490,8 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
                     self.ready_to_save_state,
                     ReadyState::WaitingForConstellationReply
                 );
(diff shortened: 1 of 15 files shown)


mrobinson:
I think this change is ready to review now.

jdm:
@bors-servo r+
Nice find on the compositor scrolling issue!

Issue #29679, opened 2023-04-28
Make names of `Sender<>` and `Receiver<>` variables more coherent


mrobinson:
Sometimes these variables are called port, sometimes ipc, sometimes they have the name of the thread on the other end and sometimes not. My proposal is that we standardize on an obvious name for all of them:

- `layout_sender` ← Always sends to the layout thread
- `layout_receiver` ← Always receives from the layout thread
- `script_sender` ← Always sends to the script thread
- `script_receiver` ← Always receives from the script thread
- `constellation_sender` ← Always sends to the constellation
- `constellation_receiver` ← Always receives from the constellation thread
- etc

mrobinson:
I believe there are probably more places in the code where we can make these changes, so I'll reopen this issue.

PR #29710, opened 2023-05-04
Make names of Sender<> and Receiver<> variables more coherent

atbrakhi:
Sometimes these variables are called port, sometimes ipc, sometimes they have the name of the thread on the other end and sometimes not.  This PR tries to standardize on an obvious name for all of them:

Common pattern to note in this PR:
- If it is a `IpcSender`, then naming is followed by `ipc_sender`, such as:  `bluetooth_ipc_sender`,  `canvas_ipc_sender` etc..
- If variable is `crossbeam_channel` sender/receiver then naming is followed by `sender`/`receiver` such as:  `devtools_sender`, `canvas_sender` etc...

---

- [x] These changes fix [#29679](https://github.com/servo/servo/issues/29679)

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -282,7 +282,7 @@ pub struct Constellation<Message, LTF, STF, SWF> {
     /// to facilitate installing pipeline namespaces in threads
     /// via a per-process installer.
     namespace_receiver: Receiver<Result<PipelineNamespaceRequest, IpcError>>,
-    namespace_sender: IpcSender<PipelineNamespaceRequest>,
+    namespace_ipc_sender: IpcSender<PipelineNamespaceRequest>,
 
     /// An IPC channel for script threads to send messages to the constellation.
     /// This is the script threads' view of `script_receiver`.
@@ -358,11 +358,11 @@ pub struct Constellation<Message, LTF, STF, SWF> {
 
     /// A channel for the constellation to send messages to the
     /// devtools thread.
-    devtools_chan: Option<Sender<DevtoolsControlMsg>>,
+    devtools_sender: Option<Sender<DevtoolsControlMsg>>,
 
     /// An IPC channel for the constellation to send messages to the
     /// bluetooth thread.
-    bluetooth_thread: IpcSender<BluetoothRequest>,
+    bluetooth_ipc_sender: IpcSender<BluetoothRequest>,
 
     /// A map of origin to sender to a Service worker manager.
     sw_managers: HashMap<ImmutableOrigin, IpcSender<ServiceWorkerMsg>>,
@@ -370,7 +370,7 @@ pub struct Constellation<Message, LTF, STF, SWF> {
     /// An IPC channel for Service Worker Manager threads to send
     /// messages to the constellation.  This is the SW Manager thread's
     /// view of `swmanager_receiver`.
-    swmanager_sender: IpcSender<SWManagerMsg>,
+    swmanager_ipc_sender: IpcSender<SWManagerMsg>,
 
     /// A channel for the constellation to receive messages from the
     /// Service Worker Manager thread. This is the constellation's view of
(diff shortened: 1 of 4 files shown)


components/constellation/constellation.rs
```diff
@@ -723,10 +723,8 @@ fn log_entry(record: &Record) -> Option<LogEntry> {
 /// The number of warnings to include in each crash report.
 const WARNINGS_BUFFER_SIZE: usize = 32;
 
-/// Route an ipc receiver to an mpsc receiver, preserving any errors.
-/// This is the same as `route_ipc_receiver_to_new_mpsc_receiver`,
-/// but does not panic on deserializtion errors.
-fn route_ipc_receiver_to_new_mpsc_receiver_preserving_errors<T>(
+/// Route an ipc receiver to an crossbeam receiver, preserving any errors.
+fn route_ipc_receiver_to_new_crossbeam_receiver_preserving_errors<T>(
```
mrobinson, line 727: In think you could also remove `mpsc_` from the name of the sender and receiver in this function as well.

components/constellation/constellation.rs
```diff
@@ -1456,11 +1468,11 @@ where
             FromCompositorMsg::Exit => {
                 self.handle_exit();
             },
-            FromCompositorMsg::GetBrowsingContext(pipeline_id, resp_chan) => {
-                self.handle_get_browsing_context(pipeline_id, resp_chan);
+            FromCompositorMsg::GetBrowsingContext(pipeline_id, browsing_context_ipc_sender) => {
```
mrobinson, line 1471: `response_sender` is a actually a pretty good name here, because it describes what the channel does -- handle the response to this query.

components/constellation/constellation.rs
```diff
@@ -1456,11 +1468,11 @@ where
...
+                self.handle_get_browsing_context(pipeline_id, browsing_context_ipc_sender);
             },
-            FromCompositorMsg::GetPipeline(browsing_context_id, resp_chan) => {
-                self.handle_get_pipeline(browsing_context_id, resp_chan);
+            FromCompositorMsg::GetPipeline(browsing_context_id, pipeline_ipc_sender) => {
```
mrobinson, line 1474: Same comment here.

components/constellation/constellation.rs
```diff
@@ -1746,11 +1759,11 @@ where
...
             },
-            FromScriptMsg::ScriptNewIFrame(load_info, layout_sender) => {
-                self.handle_script_new_iframe(load_info, layout_sender);
+            FromScriptMsg::ScriptNewIFrame(load_info, layout_ipc_sender) => {
+                self.handle_script_new_iframe(load_info, layout_ipc_sender);
             },
-            FromScriptMsg::ScriptNewAuxiliary(load_info, layout_sender) => {
-                self.handle_script_new_auxiliary(load_info, layout_sender);
+            FromScriptMsg::ScriptNewAuxiliary(load_info, layout_ipc_sender) => {
+                self.handle_script_new_auxiliary(load_info, layout_ipc_sender);
```
mrobinson, line 1766: Ditto.

mrobinson (changes requested):
This is a great start. I think in cases where a sender is merely used to send a response, we shouldn't use the name of the data type that is sent over it, but describe its purpose ie `response_sender`. I understand that this is an annoying change to make so I'm happy to do that if it's too tedious.

atbrakhi:
> This is a great start. I think in cases where a sender is merely used to send a response, we shouldn't use the name of the data type that is sent over it, but describe its purpose ie `response_sender`. I understand that this is an annoying change to make so I'm happy to do that if it's too tedious.

Thanks for the review. I have tried to accumulate these changes in recent commit.

mrobinson:
@bors-servo r+

Thank you. I've made a couple small change and squashed the commits.

Issue #29704, opened 2023-05-03
clientWidth and clientHeight special cases not implemented


delan:
Given a viewport of `w` by `h` excluding scrollbars, the following should print `w h 0 0` (with doctype) or `0 0 w h` (without doctype), but it currently prints `0 0 0 0`, because [the special cases](https://w3c.github.io/csswg-drafts/cssom-view/#dom-element-clientwidth) are not yet implemented:

```html
<!doctype html>
<meta charset="utf-8"><body>
<style>html, body { width: 0; height: 0; }</style>
<script>
    console.log(
        document.documentElement.clientWidth,
        document.documentElement.clientHeight,
        document.body.clientWidth,
        document.body.clientHeight,
    );
</script>
```

delan:
Thanks for the patch @switchpiggy! Since the original test case now passes too, I think this issue is resolved.

PR #29711, opened 2023-05-05
Implemented special cases for Element.clientWidth() and clientHeight()

switchpiggy:
Implemented the special cases for Element.clientWidth() and Element.clientHeight() as outlined in issue #29704.

---

- [x] These changes do not require tests because tests already exist.

[Specification text](https://w3c.github.io/csswg-drafts/cssom-view/#dom-element-clientwidth):
> If the element is the [root element](https://w3c.github.io/csswg-drafts/css-display-4/#root-element) and the element’s [node document](https://dom.spec.whatwg.org/#concept-node-document) is not in [quirks mode](https://dom.spec.whatwg.org/#concept-document-quirks), or if the element is [the body element](https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2) and the element’s node document is in quirks mode, return the [viewport](https://www.w3.org/TR/CSS21/visuren.html#x1) width excluding the size of a rendered scroll bar (if any).

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -81,13 +81,15 @@ use crate::dom::text::Text;
 use crate::dom::validation::Validatable;
 use crate::dom::virtualmethods::{vtable_for, VirtualMethods};
 use crate::dom::window::ReflowReason;
-use crate::dom::window::Window;
 use crate::script_thread::ScriptThread;
 use crate::stylesheet_loader::StylesheetOwner;
 use crate::task::TaskOnce;
+use app_units::Au;
 use devtools_traits::AttrInfo;
 use dom_struct::dom_struct;
+use euclid::default::Point2D;
 use euclid::default::Rect;
+use euclid::default::Size2D;
 use html5ever::serialize;
 use html5ever::serialize::SerializeOpts;
 use html5ever::serialize::TraversalScope;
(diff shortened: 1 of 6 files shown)


components/script/dom/element.rs
```diff
@@ -2501,11 +2502,32 @@ impl ElementMethods for Element {
...
     // https://drafts.csswg.org/cssom-view/#dom-element-clientwidth
     fn ClientWidth(&self) -> i32 {
+        let owner_doc: &Document = &self.node.owner_doc();
+        if self.local_name().to_string() == "body" ||
+            matches!(owner_doc.quirks_mode(), QuirksMode::Quirks)
+        {
+            let window: &Window = owner_doc.window();
+            return window.window_size().initial_viewport.round().to_i32().width;
+        }
+
```
mrobinson, line 2512: Thanks for working on this.

The specification text here is:

> If the element is the [root element](https://w3c.github.io/csswg-drafts/css-display-4/#root-element) and the element’s [node document](https://dom.spec.whatwg.org/#concept-node-document) is not in [quirks mode](https://dom.spec.whatwg.org/#concept-document-quirks), or if the element is [the body element](https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2) and the element’s node document is in quirks mode, return the [viewport](https://www.w3.org/TR/CSS21/visuren.html#x1) width excluding the size of a rendered scroll bar (if any).

It seems like this code always returns the initial viewport width when the document is in quirks mode.

Look around in this file for how the body element and root element are detected. Let's use the same approach everywhere.

Let's also move this check into `client_rect` where it adjusts the width and height of the returned rectangle so it can be shared.

mrobinson (changes requested):

components/script/dom/element.rs
switchpiggy: Thanks for the suggestions! I'll implement your feedback as soon as possible.

switchpiggy: Also, is the initial viewport width always the same as the viewport width in general, or are there some differences between the two?

mrobinson: Yes, I believe that's the case. Do you mind also pasting the spec text and linking to it above your check? Thanks!

switchpiggy: Yeah sure!

switchpiggy: I also noticed that in the Window class constructor, the current_viewport size is initialized to zero. Is this the correct behavior or is it supposed to be initialized to the dimensions of window_size?

components/script/dom/element.rs
```diff
@@ -3292,7 +3295,30 @@ impl Element {
...
+            rect = Rect::<i32>::new(
+                Point2D::<i32>::new(
+                    viewport.origin.x.to_nearest_px(),
+                    viewport.origin.y.to_nearest_px(),
+                ),
+                Size2D::<i32>::new(
+                    viewport.size.width.to_nearest_px(),
+                    viewport.size.height.to_nearest_px(),
+                ),
+            );
```
mrobinson, line 3319: If you look at how `current_viewport` is used in window.rs, you can see it is tracking the current scrollport position in the document. I don't think that's the value you want here. Instead you can simply modify the width and height here:

```rust
rect.size = Size2D::<i32>::new(
...using self.initial_viewport...
);
```

mrobinson:
Thanks for the changes. This is looking pretty good. There is a compilation error, which is why this is failing the CI checks, but if you fix that we can see if this changes any test results.

mrobinson:
@bors-servo r+

Great. Thank you!

PR #29716, opened 2023-05-05
Eliminate duplicate Layout DOM wrappers

### LINKED ISSUE #29691: Look into moving the layout DOM wrappers back into the script crate
The layout / style DOM wrappers are in the layout_thread crate because they currently need to know how to initialize and clear the `LayoutData` which is stored on the DOM nodes themselves. This has a few issues:

1. Details of the script thread leak into layout whereas the `script_layout_interface` could provide a safe interface between them.
2. The wrappers must be duplicated in both Layout 2020 and Layout 2013.
3. Because of the tricky design, there is a proliferation of traits on these wrappers which makes the code difficult to understand.

An idea here is to make all of the wrappers generics on LayoutData.

mrobinson:
There are duplicate sets of Layout DOM wrappers: one for Layout 2013 and one for Layout 2020. As part of cleaning up and simplifying the wrappers, this change parameterizes them on the specific layout data they contain. This allows them to be shared again. In addition, various small cleanups are included.

Fixes #29691.

---

- [x] These changes fix #29691


jdm (approved):
This looks like a reasonable way to reduce duplication!

mrobinson:
@bors-servo r=jdm

Thanks for the review!

jdm:
@bors-servo retry
* #29724

PR #29718, opened 2023-05-05
Snap border and outline widths at computed-value time

### LINKED ISSUE #29696: outline-width should be snapped at computed-value time
css/css-ui/parsing/outline-width-computed.html fails because `outline-width: 2.5px` computes as-is instead of being snapped to `2px` like other browsers do.

Chromium aligned with Firefox and WebKit in https://chromium-review.googlesource.com/c/chromium/src/+/3717365

Loirooriol:
According to the spec:
https://www.w3.org/TR/css-values-4/#snap-a-length-as-a-border-width

Tests:
- /css/css-backgrounds/border-width-pixel-snapping-001-a.html
- /css/css-backgrounds/border-width-small-values-001-a.html
- /css/css-backgrounds/border-width-small-values-001-b.html
- /css/css-borders/subpixel-border-width.tentative.html
- /css/css-borders/subpixel-borders-with-child-border-box-sizing.html
- /css/css-borders/subpixel-borders-with-child.html
- /css/css-outline/outline-width-rounding.tentative.html
- /css/css-outline/subpixel-outline-width.tentative.html
- /css/css-ui/parsing/outline-width-computed.html

---

- [X] These changes fix #29696 (GitHub issue number if applicable)

### DIFF
--- a/components/style/values/specified/border.rs
+++ b/components/style/values/specified/border.rs
@@ -151,13 +151,19 @@ impl ToComputedValue for BorderSideWidth {
         // We choose the pixel length of the keyword values the same as both spec and gecko.
         // Spec: https://drafts.csswg.org/css-backgrounds-3/#line-width
         // Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1312155#c0
-        match *self {
+        let width = match *self {
             BorderSideWidth::Thin => NonNegativeLength::from_px(1.).to_computed_value(context),
             BorderSideWidth::Medium => NonNegativeLength::from_px(3.).to_computed_value(context),
             BorderSideWidth::Thick => NonNegativeLength::from_px(5.).to_computed_value(context),
             BorderSideWidth::Length(ref length) => length.to_computed_value(context),
+        };
+        if width.px() == 0.0 {
+            width
+        } else {
+            // https://www.w3.org/TR/css-values-4/#snap-a-length-as-a-border-width
+            let dppx = context.builder.device.device_pixel_ratio().get();
+            Self::ComputedValue::new((width.px() * dppx).floor().max(1.0) / dppx)
         }
-        .into()
     }
 
     #[inline]
--- a/tests/wpt/metadata-layout-2020/css/css-backgrounds/border-width-pixel-snapping-001-a.html.ini
+++ b/tests/wpt/metadata-layout-2020/css/css-backgrounds/border-width-pixel-snapping-001-a.html.ini
@@ -1,2 +1,0 @@
-[border-width-pixel-snapping-001-a.html]
-  expected: FAIL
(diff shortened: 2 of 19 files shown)


mrobinson:
I think it might be a good idea to coordinate this with @emilio so that we don't introduce something incompatible with the upstream `style` crate in the Gecko repository.

Loirooriol:
Sure. Gecko doesn't need this because it uses `mActualOutlineWidth`, defined as

```
    <% impl_non_negative_length("outline_width", "mActualOutlineWidth",
                                inherit_from="mOutlineWidth",
                                round_to_pixels=True) %>
```

where `round_to_pixels` uses `round_border_to_device_pixels` to do something similar to my patch.

So to avoid diverging, not sure if Gecko can remove `mActualOutlineWidth` and just make it be the computed value, or if Servo should add something like `mActualOutlineWidth`.

emilio:
Can we do a proper sync of the Gecko style crate before landing these kinds of changes? This patch looks fine, and Gecko would benefit from this kind of patch as well.

FWIW servo should probably be doing something like `mActualOutlineWidth`, see [this comment](https://searchfox.org/mozilla-central/rev/4e6970cd336f1b642c0be6c9b697b4db5f7b6aeb/servo/components/style/properties/gecko.mako.rs#656-684) as for why it's useful.

components/style/values/specified/length.rs
```diff
@@ -32,20 +32,18 @@ pub use super::image::Image;
...
+/// Number of pixels per centimeter
+pub const PX_PER_CM: CSSFloat = PX_PER_IN / 2.54;
+/// Number of pixels per millimeter
+pub const PX_PER_MM: CSSFloat = PX_PER_IN / 25.4;
+/// Number of pixels per quarter
+pub const PX_PER_Q: CSSFloat = PX_PER_MM / 4.;
+/// Number of pixels per point
+pub const PX_PER_PT: CSSFloat = PX_PER_IN / 72.;
+/// Number of pixels per pica
+pub const PX_PER_PC: CSSFloat = PX_PER_PT * 12.;
```
emilio, line 46: In particular this change is already upstream in Gecko...

Loirooriol:
All tests pass after #30770, except for `/css/css-outline/outline-width-rounding.tentative.html`

mrobinson:
> All tests pass after #30770, except for `/css/css-outline/outline-width-rounding.tentative.html`

What are your thoughts on this change after the upgrade?

Loirooriol:
I would need to check why that test is not passing like the others

Loirooriol:
OK, so it's actually failing on Firefox too.
The reason is that 2.999px is 1st converted into Au, which doesn't have that much precision, so it becomes 3px. And then the 3px is snapped, so assuming 1dppx, it remains as 3px.
I think this behavior is fine, Blink would also turn 2.9999999px into 3px. I don't think the precision of the length (before being snapped) is specified.
So it's probably better to change the test to use 2.99px or 2.9px, then Firefox and Servo will pass.
Alternatively I guess that it could be possible to convert into Au in a more manual way than `Au::from_f32_px`, but this should be done upstream in Firefox.
So I guess I can close this.

Loirooriol:
https://github.com/web-platform-tests/wpt/pull/43428

PR #29719, opened 2023-05-05
layout 2020: Implement justify-content in flexbox

stshine:
Align the items along the main-axis per justify-content.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -22,9 +22,11 @@ use style::properties::longhands::align_self::computed_value::T as AlignSelf;
 use style::properties::longhands::box_sizing::computed_value::T as BoxSizing;
 use style::properties::longhands::flex_direction::computed_value::T as FlexDirection;
 use style::properties::longhands::flex_wrap::computed_value::T as FlexWrap;
+use style::properties::longhands::justify_content::computed_value::T as JustifyContent;
 use style::values::computed::length::Size;
 use style::values::computed::Length;
 use style::values::generics::flex::GenericFlexBasis as FlexBasis;
+use style::values::CSSFloat;
 use style::Zero;
 
 // FIMXE: “Flex items […] `z-index` values other than `auto` create a stacking context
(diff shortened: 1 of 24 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -1033,15 +1069,16 @@ impl<'items> FlexLine<'items> {
     fn resolve_auto_main_margins(
         &self,
         remaining_free_space: Length,
-    ) -> impl Iterator<Item = (Length, Length)> + '_ {
+    ) -> (impl Iterator<Item = (Length, Length)> + '_, bool) {
+        let auto_margins_count = self
```
Loirooriol, line 1073: If `remaining_free_space` is zero, treating `auto_margins_count` as zero should have no effect and would avoid iterating the items. But I guess it's fine either way.

Loirooriol (approved):
Thanks!

components/layout_2020/flexbox/layout.rs
stshine: Right. According to spec, free space should only be distributed if the remaining free space is positive.

components/layout_2020/flexbox/layout.rs
```diff
@@ -1033,7 +1069,7 @@ impl<'items> FlexLine<'items> {
     fn resolve_auto_main_margins(
         &self,
         remaining_free_space: Length,
-    ) -> impl Iterator<Item = (Length, Length)> + '_ {
+    ) -> (impl Iterator<Item = (Length, Length)> + '_, bool) {
```
Loirooriol, line 1072: Maybe you could use

```rs
    fn resolve_auto_main_margins(
        &self,
        remaining_free_space: &mut Length,
    ) -> impl Iterator<Item = (Length, Length)> + '_ {
```

and just set `*remaining_free_space = Length::zero();` is there are auto margins.

stshine: I'm not against this, but I wonder if this does not conform to the style of current code, which is quite functional.

Loirooriol: It's fine as-is, but then I guess you should update the comment above the function to mention the bool.

stshine: Okay it is done.

Loirooriol:
OK I get it, it started passing since #29719. So it's unrelated to this but I guess it's fine.

stshine:
Thanks!

PR #29731, opened 2023-05-11
Consume official GStreamer binaries on MacOS

### LINKED ISSUE #29653: MacOS builds are broken due to missing Gstreamer plugins.
The [MacOS merge queue build](https://github.com/servo/servo/actions/runs/4761313063/jobs/8462444910) has failed with the following error:

```
[Errno 2] No such file or directory: '/usr/local/lib/libnice.dylib'
[Errno 2] No such file or directory: '/usr/local/lib/libnice.10.dylib'
[Errno 2] No such file or directory: '/usr/local/lib/gstreamer-1.0/libgstnice.dylib'
[Errno 2] No such file or directory: '/usr/local/lib/gstreamer-1.0/libgstwebrtc.dylib'
ERROR: could not find required GStreamer DLL: /usr/local/lib/libnice.dylib
ERROR: could not find required GStreamer DLL: /usr/local/lib/libnice.10.dylib
ERROR: could not find required GStreamer DLL: /usr/local/lib/gstreamer-1.0/libgstnice.dylib
ERROR: could not find required GStreamer DLL: /usr/local/lib/gstreamer-1.0/libgstwebrtc.dylib
```

Based on initial investigation, it looks like Homebrew's Gstreamer package was [recently updated](https://github.com/carlocab/homebrew-core/blob/5e81cb108b3ce0a8c1266fc95767c37e7b04e893/Formula/gstreamer.rb#L133) to disable building of libnice dependency which seems to be required for webrtc plugins. This might explain the above error.

Based on this [Github discussion in Hombrew](https://github.com/Homebrew/homebrew-core/pull/125996#discussion_r1166789425), it looks like this was deliberate and the recommendation is to use the officials binaries from GStreamer if webrtc is required.

mukilan:
This PR re-enables support for the gstreamer mediastack in macOS by consuming the official binary '.pkg' files from gstreamer.freedesktop.org

To maintain symmetry with other platforms, the '.pkg' files are uploaded to servo-build-deps and fetched from there using the new script 'etc/install_macos_gstreamer.sh'.

Unlike the Homebrew version, the official GStreamer is distributed as a 'relocatable' framework i.e the dylibs all have @rpath-relative install names and also link to other dylibs using @rpath relative path. To address this difference the 'servo' binary needs to be patched with 'install_name_tool' to add an LC_RPATH command that sets the relative paths that the dynamic linker should search when trying to satify dependencies. In Servo's case, this will be a path relative to the 'servo' binary itself i.e '@executable_path/lib/'

The additional 'lib' is due to a flaw in the gstreamer packaging where the install names of some of the dylibs have the prefix '@rpath/lib' and some of them just have '@rpath'.

This PR also fixes a couple of issues present in the `mach build` process on MacOS:
1. `mach build` process was not copying transitive dependencies of servo binary but only the first level dylibs
2. `mach build` process didn't patch the links to dylibs in servo binary (and dependencies). This meant though (some) dylibs were copied to local path, the binary still loaded the dylibs from system GStreamer installation i.e homebrew instead of the copieds dylibs

The build and runtime dependencies in etc/homebrew/Brewfile and etc/homebrew/Brewfile-build have also been removed in This PR.

---

- [x] These changes fix #29653 me

- [X] These changes do not require tests because they broken build and package process.


mrobinson (approved):
I think we can keep refining how we package GStreamer, but we need to focus first on fixing the build.

PR #29733, opened 2023-05-12
Workaround bug in gstreamer binaries

### LINKED ISSUE #29732: MacOS builds are broken due to different homebrew prefix
Even after running `bash etc/install_macos_gstreamer.sh`, It gives the same error showing that it is looking for the lzma dependency

image

We have come to this conclusion but need to investigate more:
- Currently Dependency exists on following path: `ls -l /usr/local/opt/xz/lib`
- But on MacOS(13.2.1) M2, Homebrew prefix is different then `/usr/local` and dependency exists on `ls -l /opt/homebrew/Cellar/xz/5.4.3/lib/` path.

Since homebrew prefix is different on latest MacOS, it is unable to find it at that path even after installing xz that is resulting in build failure.

mukilan:
The official gstreamer .pkg distribution should contain 'relocatable' dylibs, but as discovered in #29732, some dylibs have absolute links to liblzma.5.dylib ("/usr/local/opt/xz/lib/liblzma.5.dylib")

Since /opt/homebrew is the default install location on Apple Silicion, this will cause the packaging & build steps to fail, even if 'xz' package is installed via homebrew.

This is a temporary fix until upstream fixes the bug and makes the package truly 'relocatable'. Users
with non-default homebrew prefixes will still have the issue.

Thanks to @atbrakhi  for helping me debug and test the patch on her mac.

---

- [x] These changes fix #29732 (GitHub issue number if applicable)

- [x] These changes do not require tests because they fix the broken build process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ NOTE: run these steps after you've cloned the project locally.
 ``` sh
 cd servo 
 bash etc/install_macos_gstreamer.sh
+brew bundle install --file=etc/homebrew/Brewfile
 pip install virtualenv
 ```
 
--- a/etc/homebrew/Brewfile
+++ b/etc/homebrew/Brewfile
@@ -1,0 +1,3 @@
+# Runtime dependencies
+
+brew "xz"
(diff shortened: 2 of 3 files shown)


atbrakhi (approved):
Thank you for fixing it.

mrobinson (approved):

PR #29734, opened 2023-05-12
Make the cascade data cache generic

Loirooriol:
This is a backport of https://phabricator.services.mozilla.com/D107265, by Emilio Cobos Álvarez.

---

- [X] These changes do not require tests because there should be no change in behavior.


Loirooriol:
I'm including this in #29748 since going one by one is not practical.

PR #29740, opened 2023-05-15
Refactor Layout 2020 layout for a compositor-side scroll tree

mrobinson:
This change refactors how layout is done in Layout 2020 in preparation for a compositor-side scroll tree:

1. Now the SpatialId and ClipId of each fragment is stored separately. This will allow storing a scroll node id instead of only the handle to the WebRender spatial node.
2. Separate out stacking context tree construction and display list building. This change will make it possible to eventually build the stacking context tree without the full display list if we find that necessary. For instance, this might be useful to cache containing block boundaries.
3. Add a `DisplayList` struct that stores both the WebRender display list builder and the compositor info. This exposes the API to the layout thread for display list building.

In addition, this change adds a lot of missing documentation.  This should not change behavior.

---


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -164,7 +257,7 @@ impl StackingContext {
 
     fn push_webrender_stacking_context_if_necessary<'a>(
```
mukilan, line 258: Do we still need the lifetime parameter `<'a>` here? It appears to be unused.

components/layout_2020/display_list/mod.rs
```diff
@@ -45,38 +48,99 @@ pub struct WebRenderImageInfo {
...
+    iframe_sizes: FnvHashMap<BrowsingContextId, Size2D<f32, CSSPixel>>,
+
+    /// Contentful paint i.e. whether the display list contains items of type
+    /// text, image, non-white canvas or SVG). Used by metrics.
+    /// See https://w3c.github.io/paint-timing/#first-contentful-paint.
+    is_contentful: bool,
+}
+
+impl DisplayList {
+    pub fn build<'a>(
```
mukilan, line 121: Is the explicit lifetime parameter <'a> required here?

mukilan (approved):

mrobinson:
@bors-servo r=mukilan

Thanks for the review! I've gone ahead and removed the extra lifetime parameters.

PR #29744, opened 2023-05-15
Extend RequestBuilder to include ResponseTainting

### LINKED ISSUE #29741: CORS preflight check requests must have 'response tainting' set to 'cors'
This issue was discovered while debugging the failure in #29637 

Stable unexpected results (1)

    OK /xhr/access-control-preflight-request-must-not-contain-cookie.htm
        FAIL [expected PASS] subtest: Preflight request must not contain any cookie header assert_unreached: Unexpected error. Reached unreachable code

The test failure is due to a bug in the [cors_preflight_fetch](https://github.com/servo/servo/blob/17e1582ed4756fbcf2510946dfda00b616abc02c/components/net/http_loader.rs#L1962).

Step 1 in the [fetch spec](https://fetch.spec.whatwg.org/#cors-preflight-fetch-0), the request for a preflight-check needs 'response tainting' set to 'cors':

```
1. Let preflight be a new request whose method is `OPTIONS` ... and response tainting is "cors"
```

Since our cors_preflight_fetch implementation doesn't set [response-tainting](https://fetch.spec.whatwg.org/#concept-request-response-tainting), it defaults to 'basic', causing `credentials_flag` to become true [here](https://github.com/servo/servo/blob/17e1582ed4756fbcf2510946dfda00b616abc02c/components/net/http_loader.rs#L1132), which causes the Cookie headers from previous response [to be appended ](https://github.com/servo/servo/blob/17e1582ed4756fbcf2510946dfda00b616abc02c/components/net/http_loader.rs#L1242) to the preflight request. The failing test is intended to catch this scenario i.e it expects preflight requests to not have cookies set.

To fix the issue, I think we just need to extend [RequestBuilder](https://github.com/servo/servo/blob/17e1582ed4756fbcf2510946dfda00b616abc02c/components/net_traits/request.rs#L258) to allow creating requests with given response tainting and use that in cors_preflight_check.

iverks:
Extended RequestBuilder to include ResponseTainting and used said functionality to set response_tainting to ::CorsTainting in `cors_preflight_fetch()`.

---

- [x] These changes fix #29741 (GitHub issue number if applicable)

I could not get servo to compile before these changes unfortunately. 

- [x] I have not looked at how to test this

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -1978,6 +1978,7 @@ async fn cors_preflight_fetch(
         .destination(request.destination.clone())
         .referrer_policy(request.referrer_policy)
         .mode(RequestMode::CorsMode)
+        .response_tainting(ResponseTainting::CorsTainting)
         .build();
 
     // Step 2
--- a/components/net_traits/request.rs
+++ b/components/net_traits/request.rs
@@ -93,7 +93,7 @@ pub enum RedirectMode {
 }
 
 /// [Response tainting](https://fetch.spec.whatwg.org/#concept-request-response-tainting)
-#[derive(Clone, Copy, MallocSizeOf, PartialEq)]
+#[derive(Clone, Copy, Debug, Deserialize, MallocSizeOf, PartialEq, Serialize)]
 pub enum ResponseTainting {
     Basic,
     CorsTainting,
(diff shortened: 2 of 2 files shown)


iverks:
I think I'll relieve myself of this pr until i have a working setup of the repo

mukilan:
@iverks Thank you for the PR. 

I think the build failure can be fixed by adding 'response_tainting' field to the struct initializer in '[request_init_from_request](https://github.com/servo/servo/blob/c5d31c3ab6ad4c633869ef4c4d32cd01c480661f/components/script/fetch.rs#L100)' function.

Feel free to ask in Zulip or raise a GH issue if you are not able to get servo to build.

PR #29748, opened 2023-05-17
Backport several style changes from Gecko

Loirooriol:
---


mrobinson (approved):
Great work, @Loirooriol! Could you squash these into a single commit and update all the comments to say something like:

"Update the `style` and `selectors` crates to gecko *hash* (*date*)"

Gecko uses Mercurial, so maybe just link to the last commit you included if a hash isn't appropriate. This will make it easier to continue the update process from where you left off, if someone else needs to take over.

components/style/servo/media_queries.rs
```diff
@@ -202,6 +203,12 @@ impl Device {
     pub fn safe_area_insets(&self) -> SideOffsets2D<f32, CSSPixel> {
         SideOffsets2D::zero()
     }
+
+    /// Returns true if the given MIME type is supported
+    pub fn is_supported_mime_type(&self, mime_type: &str) -> bool {
+        // TODO: being a valid MIME type doesn't imply that it's supported.
+        mime_type.parse::<Mime>().is_ok()
```
mrobinson, line 210: This should probably try to do something correct even if it results in code duplication. Maybe use the list from the networking crate, with a comment that it should be kept in sync:

- "image/x-icon"
- mime::IMAGE_BMP
- mime::IMAGE_GIF
- "image/webp"
- mime::IMAGE_PNG
- mime::IMAGE_JPEG

mrobinson:
This looks good to me. I would disable style for the crates that fail `test-tidy` and then make one small change that I noted above (actually look at the mime type of the image). Other than that, I think you can land this when you think it is ready. Thank you!

Loirooriol:
- Updated test expectations
 - Ignored files in `servo-tidy.toml` to avoid unnecessary divergences
 - Didn't squash so that bisections will be easier if something breaks
 - Updated `is_supported_mime_type`  (I tried a `match` but it didn't compile):
    ```rs
       match mime_type.parse::<Mime>() {
            Ok(m) => {
                // Keep this in sync with 'image_classifer' from
                // components/net/mime_classifier.rs
                m == mime::IMAGE_BMP
                    || m == mime::IMAGE_GIF
                    || m == mime::IMAGE_PNG
                    || m == mime::IMAGE_JPEG
                    || m == "image/x-icon"
                    || m == "image/webp"
            }
            _ => false,
        }
    ```

Loirooriol:
@bors-servo try=wpt

wpt-2020 was fine https://github.com/servo/servo/actions/runs/5026662065
Not sure what happened with 2013 https://github.com/servo/servo/actions/runs/5026716509

PR #29750, opened 2023-05-17
layout_2020: fix target main size not get clamped in flexbox

stshine:
When Resolving flexible lengths, the clamp result of item's target main size is not actually saved when it should.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -943,6 +943,8 @@ impl FlexLine<'_> {
                 //  it’s a min violation.”
                 for (item_and_target_main_size, frozen) in items() {
                     if violation(item_and_target_main_size) > Length::zero() {
+                        let (item, target_main_size) = item_and_target_main_size;
+                        target_main_size.set(item.content_min_size.main);
                         frozen_count.set(frozen_count.get() + 1);
                         frozen.set(true);
                     }
@@ -954,6 +956,11 @@ impl FlexLine<'_> {
                 //  it’s a max violation.”
                 for (item_and_target_main_size, frozen) in items() {
                     if violation(item_and_target_main_size) < Length::zero() {
+                        let (item, target_main_size) = item_and_target_main_size;
+                        let Some(max_size) = item.content_max_size.main else {
+                            unreachable!()
+                        };
+                        target_main_size.set(max_size);
                         frozen_count.set(frozen_count.get() + 1);
                         frozen.set(true);
                     }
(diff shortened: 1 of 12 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -943,6 +943,9 @@ impl FlexLine<'_> {
                 //  it’s a min violation.”
                 for (item_and_target_main_size, frozen) in items() {
                     if violation(item_and_target_main_size) > Length::zero() {
+                        item_and_target_main_size
+                            .1
+                            .set(item_and_target_main_size.0.content_min_size.main);
```
Loirooriol, line 948: I'm not sure this is the proper logic, shouldn't the clamping happen unconditionally regardless of `total_violation`?

Maybe `total_violation` is positive because most items have a big min size, but there is an item with a negative violation that needs to be clamped to the max size?

stshine: In practice this won't happen, beacause all items will either grow or shrink in one direction, and items with negative violation is inflexible and has already been frozen in a previous iteration.

stshine: Oh, it is a separate step in line 840 that size all inflexible items so they won't participate in the later loop.

Loirooriol: OK I get it, non-frozen items with a violation with a different sign than `total_violation` don't need to clamp their target main size because they will stay non-frozen, and thus in the next step we will compute the free space using their base size instead of their target main size, and then we will overwrite their target main size.

Loirooriol (approved):

components/layout_2020/flexbox/layout.rs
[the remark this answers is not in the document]
stshine: Sure thing. I'll remove the `unwrap()` also.

stshine:
Update according to review comments, thanks.
BTW, I plan to make a refinement to the current flex resolving algorithm, which feels a bit ... clunky.

components/layout_2020/flexbox/layout.rs
```diff
@@ -954,6 +956,11 @@ impl FlexLine<'_> {
                 //  it’s a max violation.”
                 for (item_and_target_main_size, frozen) in items() {
                     if violation(item_and_target_main_size) < Length::zero() {
+                        let (item, target_main_size) = item_and_target_main_size;
+                        let Some(max_size) = item.content_max_size.main else {
+                            unreachable!("")
```
Loirooriol, line 961: I cant' find other uses of `unreachable!("")`. You should probably use `unreachable!()` or provide an actual message.

But I don't really see the advantage over the previous `unwrap()`.

stshine: Because using `unwrap()` is unholy :)

stshine: Updated.

PR #29754, opened 2023-05-18
Start organizing platform-specific Python code

mrobinson:
This starts to split platform-specific Python code into its own module, which should help to tidy up our mach commands and make things more reusable. This is step one toward fixing #25335.

---

- [x] These changes do not require tests because they just change mach scripts.


python/mach_bootstrap.py
```diff
@@ -244,14 +244,15 @@ def bootstrap_command_only(topdir):
...
 
-    from servo.bootstrap import bootstrap
+    # We are not set up yet, so we always use the default cache directory
+    # for the initial bootstrap.
+    # TODO(mrobinson): Why not just run the bootstrap command in this case?
+    try:
+        servo.platform.get().bootstrap(
+            servo.util.get_default_cache_dir(topdir), '-f' in sys.argv)
+    except NotImplementedError as exception:
+        print(exception)
```
mukilan, line 254: Are we intentionally not returning non-zero exit code from the exception handler?

python/servo/platform/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import platform
+
+from .linux import Linux
```
mukilan, line 12: The builds for mac and windows are failing because importing linux will cause 'distro' to be imported, which I believe is not installed on those platforms.

I guess we can move these imports into their respective if clauses in the `get()` function.

python/servo/platform/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    elif cpu_type == "arm":
+        cpu_type = "arm"
+    elif cpu_type == "aarch64":
+        cpu_type = "aarch64"
+    else:
+        cpu_type = "unknown"
+    return f"{cpu_type}-{os_type}"
+
+
+def get():
```
mukilan, line 50: We are not returning MacOS instance for darwin platforms.  We will need that so the correct `_platform_is_gstreamer_installed`  is called on macOS.

mukilan (approved):
The bootstrap logic looks very clean now!  Nice refactoring!

python/mach_bootstrap.py
mrobinson: Oh nice catch. I've fixed this.

python/servo/platform/__init__.py
mrobinson: Yep, that's definitely the issue. I've moved to making the imports conditional in `__init__.py` instead as I think that might make the platform-specific code a little less brittle.

PR #29755, opened 2023-05-18
layout_2020: Implement automatic minimum size of flex items

stshine:
Implement the algorithm described in
https://drafts.csswg.org/css-flexbox/#min-size-auto.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -444,7 +444,77 @@ impl<'a> FlexItem<'a> {
         let max_size = box_style.content_max_box_size(containing_block, &pbm);
         let min_size = box_style.content_min_box_size(containing_block, &pbm);
 
-        let min_size = min_size.auto_is(|| automatic_min_size(box_));
+        // https://drafts.csswg.org/css-flexbox/#min-size-auto
+        let automatic_min_size = || {
+            // FIXME(stshine): Consider more situations when auto min size is not needed.
+            if box_style.get_box().overflow_x.is_scrollable() {
+                return Length::zero();
+            }
+
+            if cross_axis_is_item_block_axis {
+                let specified_size_suggestion = content_box_size.inline;
+
+                let transferred_size_suggestion = match box_ {
+                    IndependentFormattingContext::NonReplaced(_) => None,
+                    IndependentFormattingContext::Replaced(ref bfc) => {
+                        match (
+                            bfc.contents
+                                .inline_size_over_block_size_intrinsic_ratio(box_style),
+                            content_box_size.block,
+                        ) {
+                            (Some(ratio), LengthOrAuto::LengthPercentage(block_size)) => {
+                                let block_size = block_size.clamp_between_extremums(
+                                    min_size.block.auto_is(|| Length::zero()),
+                                    max_size.block,
+                                );
+                                Some(block_size * ratio)
+                            },
+                            _ => None,
+                        }
+                    },
+                };
+
+                let inline_content_size = box_
+                    .inline_content_sizes(&flex_context.layout_context)
+                    .min_content;
+                let content_size_suggestion = match box_ {
+                    IndependentFormattingContext::NonReplaced(_) => inline_content_size,
+                    IndependentFormattingContext::Replaced(ref replaced) => {
+                        if let Some(ratio) = replaced
+                            .contents
+                            .inline_size_over_block_size_intrinsic_ratio(box_style)
+                        {
+                            inline_content_size.clamp_between_extremums(
+                                min_size.block.auto_is(|| Length::zero()) * ratio,
+                                max_size.block.map(|l| l * ratio),
+                            )
+                        } else {
+                            inline_content_size
+                        }
+                    },
+                };
+
+                let result = match specified_size_suggestion {
+                    LengthOrAuto::LengthPercentage(l) => l.min(content_size_suggestion),
+                    LengthOrAuto::Auto => {
+                        if let Some(l) = transferred_size_suggestion {
+                            l.min(content_size_suggestion)
+                        } else {
+                            content_size_suggestion
+                        }
+                    },
+                };
+                result.clamp_below_max(max_size.inline)
+            } else {
+                // FIXME(stshine): Implement this when main axis is item's block axis.
+                Length::zero()
+            }
+        };
+
+        let min_size = Vec2 {
+            inline: min_size.inline.auto_is(automatic_min_size),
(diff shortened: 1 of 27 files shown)


components/layout_2020/style_ext.rs
```diff
@@ -494,3 +494,30 @@ fn size_to_length(size: &Size) -> LengthPercentageOrAuto {
...
+    let min_block_size = style
+        .min_box_size(containing_block.style.writing_mode)
+        .block
+        .maybe_percentage_relative_to(containing_block.block_size.non_auto())
+        .auto_is(|| Length::zero());
+
+    match style.get_position().box_sizing {
+        BoxSizing::ContentBox => (min_block_size / ratio, max_block_size.map(|b| b / ratio)),
+        BoxSizing::BorderBox => (
+            min_block_size / ratio - pbm.padding_border_sums.inline,
```
Loirooriol, line 519: Is `ratio` for the inner sizes, the outer sizes, or depends on `box-sizing`?

 - If it's for the inner sizes, this case seems wrong since `min_block_size` is an outer size.
 - If it's for the outer sizes, it's the same problem for the `content-box` case
 - If it depends on `box-sizing`, then `transferred_size_suggestion` needs to check the `box-sizing`

components/layout_2020/flexbox/layout.rs
```diff
@@ -444,7 +445,79 @@ impl<'a> FlexItem<'a> {
         let max_size = box_style.content_max_box_size(containing_block, &pbm);
         let min_size = box_style.content_min_box_size(containing_block, &pbm);
 
-        let min_size = min_size.auto_is(|| automatic_min_size(box_));
+        // https://drafts.csswg.org/css-flexbox/#min-size-auto
+        let automatic_min_size = || {
+            // FIXME(stshine): Consider more situations when auto min size is not needed.
+            if box_style.get_box().overflow_x != Overflow::Visible {
```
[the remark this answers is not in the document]
stshine: Fine.

components/layout_2020/flexbox/layout.rs
```diff
@@ -444,7 +445,79 @@ impl<'a> FlexItem<'a> {
...
+                match specified_size_suggestion {
+                    LengthOrAuto::LengthPercentage(l) => l.min(content_size_suggestion),
+                    LengthOrAuto::Auto => {
+                        if let Some(l) = transferred_size_suggestion {
+                            l.min(content_size_suggestion)
+                        } else {
+                            content_size_suggestion
+                        }
+                    },
+                }
```
Loirooriol, line 510: I think you are missing this:

> In all cases, the size is clamped by the [maximum](https://drafts.csswg.org/css-sizing-3/#max-width) [main size](https://drafts.csswg.org/css-flexbox-1/#main-size) if it’s [definite](https://drafts.csswg.org/css-flexbox-1/#definite).

stshine: Yeah I already had this fixed :)

components/layout_2020/flexbox/layout.rs
```diff
@@ -444,7 +445,81 @@ impl<'a> FlexItem<'a> {
...
+                    .inline_content_sizes(&flex_context.layout_context)
+                    .min_content;
+                let content_size_suggestion = match box_ {
+                    IndependentFormattingContext::NonReplaced(_) => inline_content_size,
+                    IndependentFormattingContext::Replaced(ref replaced) => {
+                        if let Some(ratio) = replaced
+                            .contents
+                            .inline_size_over_block_size_intrinsic_ratio(box_style)
+                        {
+                            let (min_inline_size, max_inline_size) = transferred_inline_min_max(
```
Loirooriol, line 488: This uses the computed min/max block sizes, so it needs to check `box-sizing`. Why not simply use `min_size.block` and `max_size.block` like for the transferred size suggestion?

components/layout_2020/style_ext.rs
```diff
@@ -494,3 +494,29 @@ fn size_to_length(size: &Size) -> LengthPercentageOrAuto {
...
+        .min_box_size(containing_block.style.writing_mode)
+        .block
+        .maybe_percentage_relative_to(containing_block.block_size.non_auto())
+        .auto_is(|| Length::zero());
+
+    match style.get_position().box_sizing {
+        BoxSizing::ContentBox => (min_block_size * ratio, max_block_size.map(|b| b * ratio)),
+        BoxSizing::BorderBox => (
+            min_block_size * ratio - pbm.padding_border_sums.inline,
+            max_block_size.map(|b| b * ratio - pbm.padding_border_sums.inline),
```
Loirooriol, line 519: It the  ratio refers to the inner size, shouldn't this be more like

```rs
(min_block_size - pbm.padding_border_sums.block) * ratio,
max_block_size.map(|b| (b - pbm.padding_border_sums.block) * ratio),
```

components/layout_2020/flexbox/layout.rs
stshine: Because they are for content size, which was truncated by border padding in block direction when `box-sizing: border-box`, thus will give incorrect result.

components/layout_2020/style_ext.rs
stshine: imo max_block_size comes from style so that's what should be transferred, right?

Loirooriol: I'm checking

```html
<!DOCTYPE html>
<div style="display: flex">
  <img src="image" style="box-sizing: border-box; min-height:300px; border: solid 100px">
</div>
<script>
addEventListener("load", () => document.body.append(document.querySelector("img").getBoundingClientRect().width);
</script>
```

where image is 200px wide and 100px tall (ratio of 2).

Your code takes `min-height:300px`, multiplies it by 2, and subtracts the horizontal border. So the inner width is `300*2 - 200 = 400`, and the outer width is 600.

But you can see that in other browsers, the inner width is 200, and the outer width is 400. That's because the inner height is 100, so they preserve the content ratio of 2.

components/layout_2020/flexbox/layout.rs
Loirooriol: I think we precisely want that truncated value.

Loirooriol (changes requested):

stshine:
yeah I think you are right. the rationale behind the converted min and max size is to keep a reasonable cross size.

Loirooriol (approved):
Thanks!

stshine:
Commits revised. Thanks!

stshine:
Aaah, wrote the wrong file

CYBAI:
might need r+ again because of new commit pushed.

PR #27539, opened 2020-08-07
[Draft] Lay out floats and handle clearance in layout 2020, but don't flow text around them yet.

pcwalton:
---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #29757, opened 2023-05-18
Lay out floats and handle clearance in layout 2020, but don't flow text around them yet

Loirooriol:
This is a crude rebase of #27539

---


mrobinson:
Things pending to do for this PR:
 - Try to eliminate the `Float` and `HoistedFloat` fragment types.
 - Use CSS 2 Appendix E guidelines to paint floating fragments in the proper order.
 - See about reducing the `SequentialLayoutState` back into `FloatContext`, which more accurately describes it. It's unclear to us what else it will be used for in the future.

mrobinson:
While turning this feature on globally does lead to a lot of progressions, there are a couple regressions and timeouts related to flexbox. We need to investigate these issues further, but in the meantime I think it's a good idea to land this as is and keep iterating on floats behind the flag until it doesn't cause unrelated regressions.

mrobinson:
@Loirooriol and I worked on rebasing this change together live and also on the new commits. We feel like we've reviewed each other's work sufficiently here.

I'll wait for @Loirooriol to issue the final command to @bors-servo though so that he can verify the final changes to the unit tests.

Loirooriol:
@bors-servo r+

I think that's a good basis, further refinement can be left for follow-ups.

stshine:
Oh, affected by #29799 .

PR #29763, opened 2023-05-20
Let the Python `Linux` platform manage prepackaged gstreamer

mrobinson:
This change makes it so that the new Python Linux class handles installing and setting gstreamer environment variables for running and build. The distribution is now installed in
"target/dependencies/gstreamer". This change necessitated a few other cleanups such as:

1. Creating and moving existing environment path append helpers to `util.py`.
2. Combining the `set_run_env` and `build_dev` functions and moving some outside code into them so that it can be shared. Now code that used to call `set_run_env` calls `build_dev` and then `os.environ.update(...)`.
3. Adding typing information to some more functions in util.

---


mrobinson:
I've been thinking about this change, and I believe it makes sense to enable `./mach bootstrap-gstreamer` for all non-cross-compilation platforms in one go. I'll open a new PR with those changes.

PR #29767, opened 2023-05-21
Fix slowness in macos copy_dependencies

### LINKED ISSUE #29764: copy_dependencies makes macOS build slow and confusing
The changes to copy all dependencies and rewrite them to relative paths causes my build to take an extra minute beyond what the build claims it took (`Finished dev [unoptimized + debuginfo] target(s) in 2m 49s`), and it displays no output while that happens.

mukilan:
The `copy_dependencies` logic is invoking `change_non_system_libraries_path` with absolute paths to gstreamer *plugin dylibs*.
`change_non_system_libraries_path` skips editing links in servo binary to relocatable dylibs, but the since the plugins have absolute paths, they are treated as 'non-relocable' and it will try to edit the *non-existent* link (since plugins are loaded dynamically) in servo bin . These unnecessary calls to change_link_name is the cause of the slowness identified in #29764 

This PR fixes the issue by ensuring plugin dylibs are not passed to the first call to change_non_system_libraries_path that patches the links in the main servo binary.

With this fix, the time taken by copy_dependencies is reduced to ~3 seconds on my system.

---

- [x] These changes fix #29764 

- [x] These changes do not require tests because they fix slowness in build step

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -557,8 +557,6 @@ class MachCommands(CommandBase):
             features=features, **kwargs
         )
 
-        elapsed = time() - build_start
-
         # Do some additional things if the build succeeded
         if status == 0:
             if android and not no_package:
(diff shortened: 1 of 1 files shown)


mukilan:
I think I can just move the call to change_non_system_libraries_path up to skip the plugins. I'll test it out and update the PR.

mrobinson (approved):
This seems like a good workaround until we can move this into the packaging step.

mukilan:
@bors-servo r=mrobinson

Thanks for the review!

PR #29772, opened 2023-05-22
Backport several style changes from Gecko (2)

Loirooriol:
This continues #29748.

---


Loirooriol:
These are the extra changes that I had to do: ee58cc036 e9d8b8f9f a613212c2 dc37faa1c ac6c24544 658f1f20f 7f9799b59 226259a33 01f8e03ab 3f77b81f7 bc519b818 c99aae5e4 6b59b4261 64a947b65 d6d1ca1fc 5c008a6a2

In particular, I'm not sure about:
 - In a613212c2 I didn't see a way to use `gecko-profiler` conditionally, so I'm removing it in servo
 - In 226259a33 I'm doing a partial revert of c3b207d, which unified font-family storage for servo and gecko. The reason being that servo needs `FontFamilyList` to have `Hash` trait, and the compiler was complaining when using a `crate::ArcSlice<SingleFontFamily>` member.

mrobinson (approved):
Great work. Hopefully over time we can reduce the divergence from the upstream `style` crate. Maybe once we sync we can see what can be synced back upstream.

Loirooriol:
Thanks! I think that my changes should be fine to upstream, except for the removal of `gecko-profiler = { path = "../../../tools/profiler/rust-api" }` from `components/style/Cargo.toml`.

PR #29780, opened 2023-05-23
Set call and construct hooks for DOM bindings.

### LINKED ISSUE #29770: Cannot extend WebIDL classes
Minimal example:
```js
class Klass extends EventTarget {
        meth() {
            return true
        }
}
let a = new Klass();
a.meth(); // Fails with a.meth is not a function
```
In Firefox it runs as expected.

Live testcase is available here: https://sagudev.github.io/test-cases/extendEventTarget

jdm:
This does not appear to fix #29770, but I'm curious if it causes any test behaviour difference.


mrobinson:
It seems like this causes many tests to fail. Perhaps it would be okay to close this for now and take it up later?

jdm:
Indeed.

PR #29781, opened 2023-05-23
Implement `bootstrap-gstreamer` for all platforms

### LINKED ISSUE #25335: Support mach bootstrap-gstreamer on Windows

mrobinson:
This change makes it so that the Platform classes can now handle installing GStreamer dependencies and properly setting up the environment including when cross-compiling. For Windows and Linux, prepackaged GStreamer is now installed into `target/dependencies/gstreamer` when not installed system-wide. In addition this change:

1. The Environment path append helper is moved to `util.py` and a new `prepend` version is added.
2. `set_run_env` and `build_dev` functions are combined and include more code from callers so environment setup is more often in the same place. Now code that used to call `set_run_env` calls `build_dev` and then `os.environ.update(...)`. We can further refine this with the `is_build` argument later.
4. Python typing information is added in many places.

---

- [x] These changes fix #25335
- [x] These changes do not require tests because they are to the build scripts.


python/servo/platform/macos.py
```diff
@@ -9,28 +9,76 @@
...
+        env = os.environ.copy()
+        self.set_gstreamer_environment_variables_if_necessary(
+            env, cross_compilation_target, check_installation=False
+        )
+        gst_lib_dir = subprocess.check_output(
+            ["pkg-config", "--variable=libdir", "gstreamer-1.0"], env=env
+        )
+        if not gst_lib_dir.startswith(bytes(GSTREAMER_ROOT, "utf-8")):
+            print("GStreamer is installed, but not the official packages.\n"
+                  "Run `./mach bootstrap` or install packages from "
```
mukilan, line 52: Should this be `./mach bootstrap-gstreamer`?

mukilan (approved):
Great work!

mrobinson:
@bors-servo r=mukilan

Thanks for the review! I've addressed the comments above.

mrobinson:
@bors-servo retry

This failed due to network issues on the GitHub runner.

PR #29784, opened 2023-05-24
Clean up and document the `send_display_list` interface

mrobinson:
This moves more members to the CompositorDisplayListInfo struct, which now holds all miscellaneous, non-WebRender data when sending display lists. It also documents what each things sent with a display list does.

---

### DIFF
--- a/components/gfx_traits/lib.rs
+++ b/components/gfx_traits/lib.rs
@@ -28,6 +28,12 @@ impl Epoch {
     }
 }
 
+impl Into<webrender_api::Epoch> for Epoch {
+    fn into(self) -> webrender_api::Epoch {
+        webrender_api::Epoch(self.0)
+    }
+}
+
 /// A unique ID for every stacking context.
 #[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, MallocSizeOf, PartialEq, Serialize)]
 pub struct StackingContextId(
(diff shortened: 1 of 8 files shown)


components/gfx_traits/lib.rs
```diff
@@ -28,6 +28,12 @@ impl Epoch {
     }
 }
 
+impl Into<webrender_api::Epoch> for Epoch {
+    fn into(self) -> webrender_api::Epoch {
+        webrender_api::Epoch(self.0)
+    }
+}
+
```
atbrakhi, line 36: I think we just had one comment about using `From` instead of `Into`. So, I am approving this PR as looks good to me.

atbrakhi (approved):

components/gfx_traits/lib.rs
mukilan: Looks like [rust 1.41 has relaxed the coherence rules](https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits) so that `From<Type>` is no longer considered a foreign trait if `Type` is owned by current trait, allowing us to implement `From<Type>` for foreign types like `webrender_api::Epoch`.

I'll raise a PR to make this change.

mrobinson: Oh, wow! This is news to me. Thank you.

PR #29787, opened 2023-05-24
Update README and HACKING_QUICKSTART to build Layout 2020

mrego:
Given that current work is around Layout 2020, this patch modifies the documentation so people pass `--with-layout-2020` when following the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ To build Servo in development mode.
 This is useful for development, but the resulting binary is very slow:
 
 ``` sh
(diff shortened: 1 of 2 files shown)


stshine:
IMHO we should make layout 2020 the default instead.

mrego:
Let's follow the discussion at https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Enabling.20Layout.202020.20by.20default

PR #29792, opened 2023-05-24
Fix layout 2020 crashes

mukilan:
This PR contains (temporary) fixes for two issues discovered in layout-2020.

1. servo crashing due to 'already mutably borrowed' error when loading 'https://mozdevs.github.io/servo-experiments/experiments/bookshelf/'
2. Dev builds crashing when loading servo.org, but not in release mode.

For the first issue, when rendering a page containing an iframe, layout 2020 creates parallel 'Layout' threads which share workers in the stylo thread pool. Because of the way the 'StyleSharingCache' structures are designed using TLS for storage of LRU cache, this leads to a double borrow of the cache when both layout threads run concurrently.

More details about the issue can be found here: https://gist.github.com/mukilan/ed57eb61b83237a05fbf6360ec5e33b0

This PR is a workaround until we find a more elegant/optimal design that also can work for gecko. The fix for now is simply to not allow multiple layouts in parallel.

For the second issue, it seems like the style pool threads overrun their allocated stack space more quickly in debug mode than in release mode. Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1376883 it seems the stack size was chosen after several experiments. It is possible the results of those experiments are no longer valid.

The temporary workaround is to increase the stack size of the worker threads to avoid layout 2020 builds from crashing when servo.org is loaded, until we can confirm the theory and implement a more robust solution.

---

- [x] These changes do not require tests because they fix crashing behaviour

### DIFF
--- a/components/layout_thread_2020/lib.rs
+++ b/components/layout_thread_2020/lib.rs
@@ -517,6 +517,7 @@ impl LayoutThread {
         animation_timeline_value: f64,
         animations: &DocumentAnimationSet,
         stylesheets_changed: bool,
+        use_rayon: bool,
     ) -> LayoutContext<'a> {
         let traversal_flags = match stylesheets_changed {
             true => TraversalFlags::ForCSSRuleChanges,
@@ -541,7 +542,7 @@ impl LayoutThread {
             font_cache_thread: Mutex::new(self.font_cache_thread.clone()),
             webrender_image_cache: self.webrender_image_cache.clone(),
             pending_images: Mutex::new(vec![]),
-            use_rayon: STYLE_THREAD_POOL.pool().is_some(),
+            use_rayon,
         }
     }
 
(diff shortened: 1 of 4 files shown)


components/layout_thread_2020/lib.rs
```diff
@@ -1012,7 +1012,8 @@ impl LayoutThread {
             RecalcStyle::pre_traverse(dirty_root, shared)
         };
 
-        let rayon_pool = STYLE_THREAD_POOL.pool();
+        let rayon_pool = STYLE_THREAD_POOL.lock().unwrap();
```
mrobinson, line 1015: Instead of locking and unlocking the pool above in `build_layout_context`, it might be better to only grab the mutex above and pass a new `use_rayon: bool` argument to `build_layout_context`.

components/style/parallel.rs
```diff
@@ -32,7 +32,7 @@ use rayon;
 use smallvec::SmallVec;
 
 /// The minimum stack size for a thread in the styling pool, in kilobytes.
-pub const STYLE_THREAD_STACK_SIZE_KB: usize = 256;
+pub const STYLE_THREAD_STACK_SIZE_KB: usize = 512;
```
mrobinson, line 35: Can we use #cfg(...) here to have different behavior for both Servo and Gecko that that this could be upstreamed one day?

mrobinson (approved):
Looks good apart from a couple small requests:

mukilan:
Thank you for the review! I've addressed the two comments in 17238b5

PR #29794, opened 2023-05-24
Clean up cross-compilation and features

mrobinson:
Integrate cross-compilation and media-stack handling into the
`build_like_command_arguments` decorator. This removes a lot of
repetition in the code and standardizes how targets are selected for all
similar commands.

Now cross compilation targets, feature flags, and helper variables are
stored in the CommandBase instance. This also avoids having to
continuously pass these arguments down to functions called by the
commands.

---


jdm:
@bors-servo r+
This is a nice cleanup!

PR #29799, opened 2023-05-26
layout_2020: Specify the used cross size when replaced item is stretched

### LINKED ISSUE #29788: Images in flexbox are not rendered properly
For example, Make a flex container with 200x200 dimensions, having a image of 100x100 dimensions inside:

```html
<!DOCTYPE html>
<html>
  <head>
    <style>
     .container {
         display: flex;
         width: 200px;
         height: 200px;
     }
    </style>
  </head>
  <body>
    <div class="container">
      <img alt="" src="support/100x100-green.png"/>
    </div>
    Hello
  </body>
</html>
```

The image is rendered as 100px square, while it should be a 200px square.

stshine:
According to spec, if the flex item has [align-self: stretch](https://drafts.csswg.org/css-flexbox/#propdef-align-self), redo layout for its contents, treating this used size as its definite cross size so that percentage-sized children can be resolved.

---

- [x] These changes fix #29788  (GitHub issue number if applicable)

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -1003,10 +1003,17 @@ impl<'a> FlexItem<'a> {
                         let pbm = replaced
                             .style
                             .padding_border_margin(flex_context.containing_block);
+                        let box_size = used_cross_size_override.map(|size| Vec2 {
+                            inline: replaced
+                                .style
+                                .content_box_size(flex_context.containing_block, &pbm)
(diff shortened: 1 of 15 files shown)


stshine:
These failed tests are using column flexbox, so they are just false-positives.
This pull request is ready for review.

Loirooriol (approved):
Seems reasonable

stshine:
Test expectations updated. Thanks!###

PR #29800, opened 2023-05-27
Only check the GStreamer installation when building

### LINKED ISSUE #29797: Runing mach try fails with missing GStreamer.
There is no need for GStreamer in mach try. There may be other commands effected by this.
```console
>mach try windows
Error running mach:

    ['try', 'windows']

message.

FileNotFoundError: GStreamer libraries not found (>= version 1.16).Please see installation instructions in README.md

  File "C:\Users\user\servo\python\servo\devenv_commands.py", line 327, in try_jobs
    res = call(["git", "push", remote, "--force", f"HEAD:{b}"], env=self.build_env())
  File "C:\Users\user\servo\python\servo\command_base.py", line 524, in build_env
    servo.platform.get().set_gstreamer_environment_variables_if_necessary(env, cross_compilation_target=target)
  File "C:\Users\user\servo\python\servo\platform\base.py", line 64, in set_gstreamer_environment_variables_if_necessary
    raise FileNotFoundError(
```

mrobinson:
Instead of always checking whether GStreamer is installed in mach's `build_env`, only do this when actually building. Also, use the instance variable to find features and look for the "media-gstreamer" feature instead of looking for !"media-dummy."

Fixes #29797.

---

- [x] These changes fix #29797
- [x] These changes do not require tests because they are changes to the build scripts.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -522,13 +522,14 @@ class CommandBase(object):
             'vcdir': vcinstalldir,
         }
 
-    def build_env(self, hosts_file_path=None, is_build=False, test_unit=False, features=None):
+    def build_env(self, hosts_file_path=None, is_build=False, test_unit=False):
         """Return an extended environment dictionary."""
         env = os.environ.copy()
 
-        if not features or "media-dummy" not in features:
+        if "media-dummy" not in self.features:
             servo.platform.get().set_gstreamer_environment_variables_if_necessary(
-                env, cross_compilation_target=self.cross_compile_target)
+                env, cross_compilation_target=self.cross_compile_target,
+                check_installation=is_build)
 
         if sys.platform == "win32" and type(env['PATH']) == six.text_type:
             # On win32, the virtualenv's activate_this.py script sometimes ends up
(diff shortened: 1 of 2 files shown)


python/servo/command_base.py
```diff
@@ -522,13 +522,14 @@ def vs_dirs(self):
             'vcdir': vcinstalldir,
         }
 
-    def build_env(self, hosts_file_path=None, is_build=False, test_unit=False, features=None):
+    def build_env(self, hosts_file_path=None, is_build=False, test_unit=False):
         """Return an extended environment dictionary."""
         env = os.environ.copy()
 
-        if not features or "media-dummy" not in features:
+        if "media-gstreamer" in self.features:
```
mukilan, line 529: I think the WPT tests are crashing because the GStreamer env is not setup, since 'mach test-wpt' is not a 'build_like_command' (so self.features == []) and we no longer call `set_run_env` from [_test_wpt](https://github.com/servo/servo/blame/ae9843dcce0abf61818aaf4ee8668ac4299c28e9/python/servo/testing_commands.py#L386)

mrobinson: Oh, you're right. We need to set these environment variables even if `self.features` hasn't been initialized.

PR #29803, opened 2023-05-28
layout_2020: Only count for content size for height of non-replaced inline elements

stshine:
Accorinding to https://drafts.csswg.org/css2/#inline-non-replaced, The vertical padding, border and margin of an inline, non-replaced box start at the top and bottom of the content area, and has nothing to do with the line-height. But only the line-height is used when calculating the height of the line box.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -461,7 +461,7 @@ impl InlineBox {
             margin.inline_start = Length::zero();
         }
         let mut start_corner = Vec2 {
-            block: padding.block_start + border.block_start + margin.block_start,
+            block: Length::zero(),
             inline: ifc.inline_position - ifc.current_nesting_level.inline_start,
         };
         if style.clone_position().is_relative() {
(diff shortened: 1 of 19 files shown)


stshine:
Although the block direction margins do not take effects, they still show up in firefox devtools.
But still I'm not sure if this is the right fix.

Loirooriol (changes requested):
I don't think this should special-case margins. https://drafts.csswg.org/css2/#inline-non-replaced says

> The vertical **padding, border and margin** of an inline, non-replaced box start at the top and bottom of the content area, and has nothing to do with the [line-height](https://drafts.csswg.org/css2/#propdef-line-height). But only the line-height is used when calculating the height of the line box.

If I take `/css/CSS2/margin-padding-clear/margin-inline-002.xht` replacing the margin with padding, then it's still wrong:

- Servo:
 
- Firefox:
  

So can you try removing the vertical padding and borders too, and see if more tests pass or something breaks?

stshine:
More tests passed :)

Loirooriol (approved):
This case is still wrong:

```html
<!DOCTYPE html>
<style>
div {
    border: blue solid 3px;
    margin-top: 50px;
}
span {
    background: orange;
    padding-top: 20px;
    padding-bottom: 20px;
    height: 100px;
    line-height: 100px;
}
</style>
<div><span>Filler Text</span></div>
```

But this patch is an improvement so I guess it's fine as-is.

stshine:
that looks like is related to line height, let's deal with that later.
Thanks!

PR #29805, opened 2023-05-29
Support extending DOM classes in JS

### LINKED ISSUE #29770: Cannot extend WebIDL classes

jdm:
Adds support for determining the correct prototype as part of a `new` call for non-HTML constructors. Support for HTML constructors [was added](https://github.com/servo/servo/blob/master/components/script/dom/bindings/htmlconstructor.rs#L116) as part of the custom element work, but that wasn't enough for extending other DOM objects like `EventTarget`.

This work is based on Gecko's code in https://searchfox.org/mozilla-central/rev/2d678a843ceab81e43f7ffb83212197dc10e944a/dom/bindings/BindingUtils.cpp#3667, and is split into two parts:
1) the actualy new implementation (demonstrating that extending `EventTarget` works as expected)
2) plumbing the new prototype through all of the rest of the DOM object construction code.

This ends up being more complex than the way it's done in Gecko because they have lazy DOM reflectors, so their native DOM objects can be constructed without any knowledge of a JS prototype. Servo's reflectors are eager, however, so we need to propagate prototype information into individual constructors. As a result, this was a tedious set of changes to make.

---
- [x] These changes fix #29770


components/script/dom/bindings/interface.rs
```diff
@@ -592,3 +602,111 @@ pub fn define_dom_interface(
...
+    proto_id: PrototypeList::ID,
+    creator: unsafe fn(SafeJSContext, HandleObject, *mut ProtoOrIfaceArray),
+    mut desired_proto: MutableHandleObject,
+) -> Result<(), ()> {
+    unsafe {
+        // This basically implements
+        // https://heycam.github.io/webidl/#internally-create-a-new-object-implementing-the-interface
+        // step 3.
+
+        //assert!(args.is_constructing())
```
jdm, line 629: We don't expose `is_constructing` on the rust CallArgs object at this point.

components/script/dom/bindings/utils.rs
```diff
@@ -112,6 +115,7 @@ unsafe impl Sync for DOMClass {}
 
 /// The JSClass used for DOM object reflectors.
 #[derive(Copy)]
+#[repr(C)]
```
jdm, line 118: This is a tiny addition with a big impact. Adding the `depth` field to `DOMClass` caused rustc to reorder the fields of `DOMJSClass`, which broke lots of code that assumed that casting a `JSClass` pointer to `DOMJSClass` would yield a valid object.

sagudev:
Some docs should probably be updated too. Namely https://doc.servo.org/script/dom/index.html#construction

sagudev:
Hm, that's weird. We get [errors in building release](https://github.com/servo/servo/actions/runs/5108277838/jobs/9181939619#step:8:1085) step but error did not stopped execution of next step. Maybe returning codes are wrong.

CYBAI:
> but error did not stopped execution of next step. Maybe returning codes are wrong.

```
Failed in 0:15:20
Warning:  Could not generate notification: No module named 'dbus'
```

I suspect if the returning code is impacted by the last warning 🤔

jdm:
Bleah, I forgot that I relied on a change to mozjs that I haven't pushed up yet.

components/script/dom/bindings/utils.rs
```diff
@@ -637,3 +637,7 @@ impl AsCCharPtrPtr for [u8] {
         self as *const [u8] as *const c_char
     }
 }
+
+pub unsafe fn callargs_is_constructing(args: &CallArgs) -> bool {
```
jdm, line 641: This will be moved to a method on CallArgs in mozjs in the future. It's equivalent to https://searchfox.org/mozilla-central/source/js/public/CallArgs.h#163

components/script/dom/bindings/htmlconstructor.rs
```diff
@@ -225,10 +225,18 @@ unsafe fn html_constructor(
         None => {
             // Step 8.1
             let name = QualName::new(None, ns!(html), definition.local_name.clone());
+            // Any prototype used to create these elements will be overwritten before returning
```
mrobinson, line 228: Now that we have `get_desired_proto` could we replace step 6 above (and the workaround) with a call to that. A related change was made in Gecko here: https://phabricator.services.mozilla.com/D28518

components/script/dom/bindings/reflector.rs
```diff
@@ -23,7 +23,16 @@ where
     U: DerivedFrom<GlobalScope>,
 {
     let global_scope = global.upcast();
-    unsafe { T::WRAP(GlobalScope::get_cx(), global_scope, obj) }
+    unsafe { T::WRAP(GlobalScope::get_cx(), global_scope, None, obj) }
+}
+
+pub fn reflect_dom_object2<T, U>(obj: Box<T>, global: &U, proto: Option<HandleObject>) -> DomRoot<T>
```
mrobinson, line 29: Nit: Would it be possible to name this `reflect_dom_object_with_proto` so that it follows a similar pattern with `reflect_node_with_proto`?

Another thing to consider is to remove the `with_proto` variants entirely and always pass the prototype, but it seems like the large amount of noisy changes needed for that would be better suited for a possible followup.

mrobinson (approved):
This looks good. We took a look at today and only had a couple comments:

components/script/dom/bindings/htmlconstructor.rs
jdm: I'll file a follow-up issue for this.

PR #29807, opened 2023-05-29
Propagate status on `build_like_command_arguments`

### LINKED ISSUE #29806: mach build does not fail
Hm, that's weird. We get [errors in building release](https://github.com/servo/servo/actions/runs/5108277838/jobs/9181939619#step:8:1085) step but error did not stopped execution of next step. Maybe returning codes are wrong.

_Originally posted by @sagudev in https://github.com/servo/servo/issues/29805#issuecomment-1566511072_
          
Relevant code: https://github.com/servo/servo/blob/7d50362c72b86eaface602cfa52827c84b8f9bdd/python/servo/build_commands.py#L515-L520

sagudev:
As observed in #29805 

---

- [x] These changes fix #29806 

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -789,7 +789,7 @@ class CommandBase(object):
 
             self.features = kwargs.get("features", None) or []
             self.configure_media_stack(kwargs['media_stack'])
-            original_function(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


mrobinson (approved):
Thank you for this fix! It looks like this bug was hiding another issue which I will try to fix now so we can land this.

PR #29810, opened 2023-05-29
css `inset` should work on Layout 2020

### LINKED ISSUE #29705: ‘transition-property’ should expand shorthands
Toggling the checkbox should transition ‘left’ but does not in Layout 2020 (mach build --with-layout-2020):

```html
<style>
    input {
        position: relative;
        left: 0;
        transition: inset 0.5s;
    }
    input:checked {
        left: 1em;
    }
</style>
<input type="checkbox">
```

https://drafts.csswg.org/css-transitions/#transition-property-property

> For the keyword [all](https://w3c.github.io/csswg-drafts/css-transitions/#valdef-transition-property-all), or if one of the identifiers listed is a shorthand property, implementations must start transitions for all its longhand sub-properties (or, for all, all properties), using the duration, delay, and timing function at the index corresponding to the shorthand.

MendyBerger:
`top`, `bottom`, `left`, `right`, are already implemented in layout-2020, so adding the shorthand can be enabled.

---

- [X] These changes fix #29705 (GitHub issue number if applicable)

I'm don't know if I need tests for this, how do I determine that?


mrobinson:
It's very odd that this didn't change any test results. 🤔

jdm:
There are two other inset shorthands in that file. Should they be updated as well?

MendyBerger:
@jdm don't think they're implemented in layout-2020, but not sure

Loirooriol:
@MendyBerger The logical `inset-*` longhands are enabled in layout 2020, so there should be no problem enabling the shorthands, since they just expand into the longhands. Also, the logical longhands just share a computed value with the physical ones, so this is basically a style thing, layout shouldn't need anything special.

@mrobinson Enabling the 3 shorthands should improve `css/css-logical/logical-box-inset.html`, but it's currently skipped :(

MendyBerger:
Thanks @Loirooriol, I've updated the PR.

Just out of curiosity, how did you know that it's implemented? Doing a global search didn't return any results, so I first assumed it's not implemented. But then a manual test confirmed that you're right.

Loirooriol (commented):
Please update the test expectations, `css-logical` is no longer skipped.

These are improving:
 - /css/css-logical/logical-box-inset.html
 - /css/css-logical/parsing/inset-block-inline-valid.html
 - /css/css-logical/parsing/inset-block-inline-shorthand.html
 - /css/css-logical/parsing/inset-shorthand.html
 - /css/css-logical/parsing/inset-valid.html

And this crashes :(
 - /css/css-logical/animation-002.html

It's hitting https://github.com/servo/servo/blob/13e1aeb8cf0c970728a9565c38cdfbd8b8bcae1b/components/style/properties/helpers.mako.rs#L486

Maybe there is some problem with pending-substitution values:

```html
<!DOCTYPE html>
<style>
:root {--200px: 200px; --300px: 300px}
@keyframes anim{
  from { inset-block: var(--200px) }
  to { inset-block: var(--300px) }
}
</style>
<div style="animation: anim 10s -5s paused linear; width: 0px; height: 0px;"></div>
```

Can you investigate if there is an easy fix? Or otherwise go back to just enabling `inset` I guess, since it doesn't seem to have this problem.

MendyBerger:
Note: the `block`/`inline` issue isn't specific to `inset`, it happens to any `block`/`inline`, e.g. `margin-block` or `padding-inline`.

The problem seems to be that `cascade_property()` doesn't select the correct branch of the match, since the `block`/`inline` isn't being replaced properly. I'm assuming that [to_phys()](https://github.com/servo/servo/blob/master/components/style/properties/data.py#L223) or something similar isn't called in a place that it should be.

Any thoughts appreciated. This is tough but I'll continue to investigate.

MendyBerger:
@Loirooriol does it make sense to merge like this even with this issue since it's not specific to inset-inline/block? It also happens to padding-inline/block and margin-inline/block.
I've spent upwards of 30 hours investigating this issue but still can't pin down where things are going wrong... I'd be willing to continue if someone can help me out here, but otherwise I don't think I can justify spending so much time on this anymore :(

MendyBerger:
I've opened a separate issue for the inline/block crash issue https://github.com/servo/servo/issues/29891

Loirooriol:
Adding more crashes isn't great, but since it's already crashing for the other logical shorthands then I guess landing is fine.

Loirooriol:
@bors-servo try=wpt-2020

So I guess let's just run the bot and then please update test expectations.

PR #29813, opened 2023-05-30
layout_2020: Check blocks for whether they will establish a formatting context

stshine:
According to spec
https://drafts.csswg.org/css-overflow-3/#overflow-control, If the computed value of overflow on a block box is neither visible nor clip nor a combination thereof, it establishes an independent formatting context for its contents.

---

### DIFF
--- a/components/layout_2020/flow/construct.rs
+++ b/components/layout_2020/flow/construct.rs
@@ -11,7 +11,7 @@ use crate::flow::inline::{InlineBox, InlineFormattingContext, InlineLevelBox, Te
 use crate::flow::{BlockContainer, BlockFormattingContext, BlockLevelBox};
 use crate::formatting_contexts::IndependentFormattingContext;
 use crate::positioned::AbsolutelyPositionedBox;
-use crate::style_ext::{DisplayGeneratingBox, DisplayInside, DisplayOutside};
+use crate::style_ext::{ComputedValuesExt, DisplayGeneratingBox, DisplayInside, DisplayOutside};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use rayon_croissant::ParallelIteratorExt;
 use servo_arc::Arc;
@@ -591,7 +591,9 @@ where
 
         let kind = match contents.try_into() {
             Ok(contents) => match display_inside {
-                DisplayInside::Flow { is_list_item } => {
+                DisplayInside::Flow { is_list_item }
+                    if !info.style.establishes_block_formatting_context() =>
+                {
                     BlockLevelCreator::SameFormattingContextBlock(
                         IntermediateBlockContainer::Deferred {
                             contents,
--- a/components/layout_2020/style_ext.rs
+++ b/components/layout_2020/style_ext.rs
@@ -11,6 +11,7 @@ use style::computed_values::transform_style::T as ComputedTransformStyle;
 use style::logical_geometry::WritingMode;
 use style::properties::longhands::backface_visibility::computed_value::T as BackfaceVisiblity;
 use style::properties::longhands::box_sizing::computed_value::T as BoxSizing;
+use style::properties::longhands::column_span::computed_value::T as ColumnSpan;
 use style::properties::ComputedValues;
 use style::values::computed::image::Image as ComputedImageLayer;
 use style::values::computed::{Length, LengthPercentage};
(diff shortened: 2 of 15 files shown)


components/layout_2020/flow/construct.rs
```diff
@@ -591,7 +591,9 @@ where
 
         let kind = match contents.try_into() {
             Ok(contents) => match display_inside {
-                DisplayInside::Flow { is_list_item } => {
+                DisplayInside::Flow { is_list_item }
+                    if !info.style.get_box().overflow_x.is_scrollable() =>
```
mrobinson, line 595: I think this should be a new function on the `StyleExt` crate, because there are [quite a few properties](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Block_formatting_context) that trigger the creation of a new formatting context and we should try to handle them all. In addition, I think this should check both `overflow_x` and `overflow_y`.

mrobinson (changes requested):
Thanks for doing this. We recently noticed this problem working on floats. A couple comments below:

components/layout_2020/flow/construct.rs
Loirooriol: I agree with turning this into its own function.

Checking `overflow-y` isn't probably necessary since elements can't be a scroll container in one axis but not in the other. The style adjuster takes care of changing the other property.

However, not all elements with `overflow: scroll` end up being scroll containers (e.g. replaced elements). So maybe it would be better to check layout rather than the computed style. Not sure if it matters in practice, though.

stshine: I'm not very sure what the function should have since establishing formatting context is quite a complex topic and this check only happens when the element is a block, or, have `DisplayOutside::Block` and `DisplayInside::Flow`. Some is handled in other places(floats, abs-pos elements, `<html>`), Some is other kind of formatting context(Flexbox), and Some are not actually implemented(tables, multicol, grid).

mrobinson: > I'm not very sure what the function should have since establishing formatting context is quite a complex topic and this check only happens when the element is a block, or, have `DisplayOutside::Block` and `DisplayInside::Flow`. Some is handled in other places(floats, abs-pos elements, `<html>`), Some is other kind of formatting context(Flexbox), and Some are not actually implemented(tables, multicol, grid).

It's true that not all situations would be useful to put in this function, but offhand this could also check whether the contents are `display: flow-root` when we add support for that as well as `column-span: all`. In short, I think there will need to be more situations in the not too distant future.

stshine: > However, not all elements with `overflow: scroll` end up being scroll containers (e.g. replaced elements). So maybe it would be better to check layout rather than the computed style. Not sure if it matters in practice, though.

Since spec says bfc is established when `overflow` is not `visible` or `clip`, IMO this is OK.

Loirooriol: Ah, that's right, it checks the compute value.

stshine:
Updated according to comments.

stshine:
Updated according to comments.

components/layout_2020/style_ext.rs
```diff
@@ -365,6 +367,25 @@ impl ComputedValuesExt for ComputedValues {
         }
     }
 
+    /// Return true if this style is a normal block and establishes
+    /// a new blockformatting context.
+    fn block_establish_formatting_context(&self) -> bool {
```
mrobinson, line 372:
```suggestion
    /// a new block formatting context.
    fn establishes_block_formatting_context(&self) -> bool {
```

Let's keep the name consistent with the other methods here.
Nit: `blockformatting` →`block formatting`

mrobinson (approved):
Thank you! This looks great, other than 2 very small naming nits.

stshine:
oh wait a minute, haven't updated test expectation yet. investigating the failed tests :)

mrobinson:
> oh wait a minute, haven't updated test expectation yet. investigating the failed tests :)

Whoops sorry. I'll cancel the job.

stshine:
Looks like the tests failed because the following not implemented:

UAs must apply the [overflow](https://drafts.csswg.org/css2/#propdef-overflow) property set on the root element to the viewport. When the root element is an HTML "HTML" element or an XHTML "html" element, and that element has an HTML "BODY" element or an XHTML "body" element as a child, user agents must instead apply the overflow property from the first such child element to the viewport, if the value on the root element is [visible](https://drafts.csswg.org/css2/#valdef-overflow-visible). The visible value when used for the viewport must be interpreted as [auto](https://drafts.csswg.org/css2/#valdef-overflow-auto). The element from which the value is propagated must have a used value for overflow of visible.

Let's deal with that later.

mrobinson:
A fix for the above issue is on my todo list.

PR #29814, opened 2023-05-30
Stop skipping css-logical tests in layout-2020

Loirooriol:
---


mrobinson (approved):

Loirooriol:
I expected less failures TBH, since this should mostly be a style thing.

Loirooriol:
Ah, that explains it. Most of the logical longhands seem enabled, but if writing modes aren't supported, they will not be mapped as the tests expect.

PR #29827, opened 2023-06-01
Layout 2020: Move all Fragment code to the `fragment_tree` directory

mrobinson:
This is a simple code organization change with no behavior change with the idea of making Layout 2020 easier to understand by new folks to the project. The idea is that we will have a cleaner separation between the different parts of layout ie one directory for the fragment tree and one (currently multiple) directory for the box tree.

---


mrobinson:
This rebased and ready to review again. This looks like a big change, but is really just shuffling code around with no changes other than updating use statements. Here's what the new directory structure inside the `layout_2020` crate looks like:

```
├── fragment_tree
│   ├── base_fragment.rs
│   ├── box_fragment.rs
│   ├── containing_block.rs
│   ├── fragment.rs
│   ├── fragment_tree.rs
│   ├── hoisted_shared_fragment.rs
│   └── mod.rs
```

Most of this code is moved from `fragment.rs` in the root of the crate, which had the implementation of every fragment type. In addition, the `FragmentTree` data structure is moved from `flow/root.rs` to `fragment_tree/fragment_tree`. All of the crate-public data structures are exported from `crate::fragment_tree` directly, so the source file that they are in shouldn't affect how they are accessed.

atbrakhi (approved):

mrobinson:
@bors-servo r=atbrakhi

Thanks for the review!

jdm:
Should the contents of fragment_tree/fragment_tree.rs be re-exported in fragment_tree/mod.rs? It looks weird to write `crate::fragment_tree::fragment_tree::FragmentTree`.

mrobinson:
> Should the contents of fragment_tree/fragment_tree.rs be re-exported in fragment_tree/mod.rs? It looks weird to write `crate::fragment_tree::fragment_tree::FragmentTree`.

Huh. I think they are or at least that is what I was trying to do. For example, in the `display_list` crate the important looks like this:

```rust
use crate::fragment_tree::{BoxFragment, Fragment, FragmentTree, Tag, TextFragment};
```

ie `use crate::fragment_tree::FragmentTree`.

mrobinson:
Oh, it seems that some code was lost in merge conflict resolution. I'll fix this now.

Issue #29819, opened 2023-05-31
Layout 2020 is vulnerable to floating point precision issues


mrobinson:
While debugging floating point test failures, I noticed that large layout values (`f32`) start to introduce pretty terrible floating point precision errors. This was addressed in Layout 2013 by the use of `app_unit`, I believe and WebKit uses `LayoutUnit`. We should explore a similar solution for Layout 2020.

dvtkrlbs:
Related to this the Ladybird browser engine from the SerenityOS project jsut did a status of engine video and the developer talks about this issue. They temporarily mitigate some issues by using 64 bit floating numbers but apperently the absolute solution is to use fixed point decimals for these kind of values. We might temporarily mititage part of this issue by using f64. Youtube link with a timestamp https://youtu.be/h72mpiYgFtA

stshine:
What is the reasoning of Layout 2020 using floating point instead of Au?

stshine:
Found this in IRC log.

```
2019-11-19T22:39:10 #servo <nox> pcwalton: Can you give me a 2 minutes master class on why Au exists, btw?
2019-11-19T22:39:37 #servo <pcwalton> nox: IIRC the app_units crate has a big comment with a history explainer
2019-11-19T22:39:45 #servo <nox> Oh, should have looked at it first.
2019-11-19T22:39:59 #servo <pcwalton> I'm kind of lukewarm on app units these days; floats may be better
2019-11-19T22:40:20 #servo <nox> pcwalton: Oh?
2019-11-19T22:40:27 #servo <pcwalton> nox: https://bugzilla.mozilla.org/show_bug.cgi?id=177805 apparently
2019-11-19T22:40:29 #servo <firebot> Bug 177805 — FIXED, sharparrow1⊙yc — Fix the use of units in Gecko
2019-11-19T22:40:56 #servo <pcwalton> nox: yeah, mainly because people do things like make SVGs that are 1 unit by 1 unit and then scale them up with a transform
2019-11-19T22:41:26 #servo <pcwalton> like if your SVG uses an internal coordinate system from 0.0..1.0, which is a perfectly reasonable thing to do
2019-11-19T22:41:47 #servo <pcwalton> then use a transform to scale them to the proper pixel size
2019-11-19T22:41:49 #servo <pcwalton> this breaks with app units
2019-11-19T22:42:00 #servo <pcwalton> because 1/60 of a unit is not precise enough in most cases
[...]

PR #29832, opened 2023-06-02
Fix `mach test-wpt` to make crash tests work

mukilan:
The first commit fixes two issues related to crash tests:
1. test-wpt is unable to find existing crash tests even when called with --test-types=crashtests. The fix here is to add crashtests to the default test suite types to python/wpt/run.py
2. When running in headless mode, crashes in style threads don't cause servo to crash because the logic in constellation.rs currently calls handle_panic only when the top-level browsing context id is some value. Since style pool threads are shared, they always generate Panic messages with None as top-level browsing context id.

The second commit addresses issues with generating and capturing backtrace and stderr for crash tests.

---

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -1484,11 +1484,10 @@ where
             // Panic a top level browsing context.
             FromCompositorMsg::SendError(top_level_browsing_context_id, error) => {
                 debug!("constellation got SendError message");
-                if let Some(id) = top_level_browsing_context_id {
-                    self.handle_panic(id, error, None);
-                } else {
+                if top_level_browsing_context_id.is_none() {
                     warn!("constellation got a SendError message without top level id");
                 }
+                self.handle_panic(top_level_browsing_context_id, error, None);
             },
             // Send frame tree to WebRender. Make it visible.
             FromCompositorMsg::SelectBrowser(top_level_browsing_context_id) => {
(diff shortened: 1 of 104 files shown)


mrobinson:
Hrm. It seems like this is causing more timeouts. I wonder what could be preventing the process from exiting in these tests...

jdm (approved):

mrobinson:
Looks like the two failures are:

   - PASS [expected CRASH] /custom-elements/when-defined-reentry-crash.html (2013 / 2020)
   - PASS [expected CRASH] /streams/readable-streams/crashtests/strategy-worker-terminate.html (2020)

The first was recently fixed, so I think we can remove that expectation. The second must be a flaky crash. For that I guess we simply need to remove the expectation and open an I-Intermittent bug to track it.

mukilan:
Thanks @mrobinson! I've removed the expectations for those tests and repurposed #30124 to track the intermittent CRASH

mrobinson:
I suspect that this failure in the float unit tests is unrelated to this change. The float unit tests use quickcheck to essentially "fuzz" the float layout algorithm. Sometimes the values passed stretch the limits of our floating point layout in Layout 2020 (see #29819). To confirm, I'm going to send this back to the MQ to see if this behavior is flaky.

mrobinson (approved):

PR #29839, opened 2023-06-04
layout_2020: Allow end margin to collapse with children if height is set to 0

stshine:
Previously, end margin was allowed to collapse only when block size is not set.

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -482,7 +482,7 @@ fn layout_in_flow_non_replaced_block_level(
     let end_margin_can_collapse_with_children = block_is_same_formatting_context &&
         pbm.padding.block_end == Length::zero() &&
         pbm.border.block_end == Length::zero() &&
-        block_size == LengthOrAuto::Auto &&
+        block_size.auto_is(|| Length::zero()) == Length::zero() &&
         min_box_size.block == Length::zero();
 
     let mut clearance = Length::zero();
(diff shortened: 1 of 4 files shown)


stshine:
Ah oh.


@bors-servo retry

Loirooriol:
The patch broke this testcase:

```html
<!DOCTYPE html>
<div style="height: 0px">
  <div style="margin-bottom: 50px">a</div>
</div>
b
```

"a" and "b" are supposed to overlap instead of being separated by a collapsed margin.

https://drafts.csswg.org/css2/#adjoining-margins says:

> bottom margin of a last in-flow child and bottom margin of its parent if the parent has [auto](https://drafts.csswg.org/css2/#valdef-height-auto) computed height

It does not include `height: 0`. The relevant part for that is

> top and bottom margins of a box that does not establish a new block formatting context and that has zero computed [min-height](https://drafts.csswg.org/css2/#propdef-min-height), zero or [auto](https://drafts.csswg.org/css2/#valdef-height-auto) computed [height](https://drafts.csswg.org/css2/#propdef-height), and no in-flow children

Loirooriol:
Also, for the 2nd quote the spec checks the computed value of `height` without taking `max-height` into account, so

```html
<!DOCTYPE html>
a
<div style="margin-bottom: 50px"></div>
<div style="margin-bottom: 50px; height: 1px; max-height: 0"></div>
b
```

should separate "a" and "b" by 100px instead of 50px. That's what Firefox does (but Blink collapses :thinking:)

stshine:
whoops, let me try to fix this.

Loirooriol:
Turns out nobody is following the spec in all cases: https://github.com/w3c/csswg-drafts/issues/8919

But still following the spec may be simplest, since it just checks the computed style.

Loirooriol:
I opened https://github.com/servo/servo/pull/29856

PR #29850, opened 2023-06-07
CSS viewport removal

### LINKED ISSUE #26349: Remove @viewport.
There's a fairly complex setup to support `@viewport` rules but they were removed from the spec, see https://github.com/w3c/csswg-drafts/issues/4766.

So we should remove all that code, except for the fact that Servo uses a dummy `@viewport` rule for `<meta name="viewport">`, so that should be rewritten first.

emilio:
It was removed from the spec and it's disabled everywhere.

This also removes the meta viewport support (which was implemented on top), but that also had a single test and is disabled everywhere, so I'm not too concerned, it can be implemented again if / when needed.

- [x] These changes fix #26349


Loirooriol (commented):
Thanks! But I'm concerned that this does several changes in `components/style` that diverge from firefox, and will probably will it more difficult to update the style.

So maybe it should wait until servo has caught up with firefox, or maybe you can land this in firefox instead?

emilio:
> Thanks! But I'm concerned that this does several changes in `components/style` that diverge from firefox, and will probably will it more difficult to update the style.
> 
> So maybe it should wait until servo has caught up with firefox, or maybe you can land this in firefox instead?

Sure, my plan is landing it on Firefox too. Happy to do it first, if we haven't done it earlier is because I thought servo was using it a lot more.

emilio:
[bug 1837273](https://bugzil.la/1837273) fwiw.

mrobinson:
Sorry for letting this sit so long. I pushed a new version of this branch that makes the changes to Servo, but leaves the style / style_traits mostly intact. The idea here is that we can pick up the removal when we sync with Gecko's style. Thanks for this cleanup!

PR #12754, opened 2016-08-06
Fix unexpected freeze of flex item

stshine:
Fix the currently logic that a item will freeze if it should
grow(shrink) and its basesize is less(more) than its min(max) size. This
also fix the divide by zero error when an item should shrink but it has
zero length and zero min size.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] There are tests for these changes

---

This change is [Reviewable](https://reviewable.io/reviews/servo/servo/12754)

PR #29854, opened 2023-06-08
layout_2020: Fix divide by zero error when resolving flexible lengths

### LINKED ISSUE #29852: Infinite loop with flexbox in layout-2020
This causes an infinite loop:

```html
<div style="display: flex; width: 0px">
  <div style="flex: 1 1 0px; margin: 1px"></div>
</div>
```

The reason is that the spec is not avoiding a `0 / 0` division, so servo ends up with NaN lengths and the algorithm doesn't advance as expected. See https://github.com/w3c/csswg-drafts/issues/8928

stshine:
---

- [x] These changes fix #29852 (GitHub issue number if applicable)

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -905,9 +905,13 @@ impl FlexLine<'_> {
         for ((item, target_main_size), frozen) in items() {
             let is_inflexible = flex_factor(item) == 0. ||
                 if grow {
-                    item.flex_base_size > item.hypothetical_main_size
+                    item.flex_base_size > item.hypothetical_main_size ||
+                        item.content_max_size
+                            .main
+                            .map_or(false, |l| l == item.flex_base_size)
                 } else {
-                    item.flex_base_size < item.hypothetical_main_size
+                    item.flex_base_size < item.hypothetical_main_size ||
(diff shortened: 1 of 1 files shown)


stshine:
A previous work can be seen at #12754 .

stshine:
WHAT?

@bors-servo retry

stshine:
seems it works, but the code looks ugly. Maybe we should remove `hypothetical_main_size`.

Loirooriol:
Yeah, your 1st patch would freeze items too aggressively.
The current version may be fine, I saw that layout-2013 does something similar: https://github.com/servo/servo/blob/ac09fdf0c72844e6ddd4e7463f8917b0abbdf4a5/components/layout/flex.rs#L278-L281

But this doesn't match the spec so I will need to think whether it's equivalent.

stshine:
The layout 2013 one is also done by me :)

Loirooriol:
This doesn't solve the problem for

```html
<div style="display: flex; width: 0px">
  <div style="flex: 1 1e-40 1e-40px; margin: 1px"></div>
</div>
```
`1e-40px` is greater than 0px so it's not affected by the change, but then `1e-40 * 1e-40px` ends up becoming zero.
#29863 seems a more robust solution.

PR #29855, opened 2023-06-08
Set `id_hash` as default wpt chunker

### LINKED ISSUE #29844: Webgpu CTS are not chunked properly.
When using `./mach test-wpt --release _webgl --total-chunks 5 --this-chunk X` only chunk 3 (all `_webgpu/webgpu/cts.https.html` tests) and chunk 5 (all other - `_webgpu/webgpu/webgpu/web_platform/reftests`) are working. Others do not.

I do not know how exactly chunking should work, but I believe it is not working due to `cts.https.html` being single endpoint for all test, which can be changed in vendoring script. EDIT: No it cannot be yet. Waiting for https://github.com/gpuweb/cts/pull/1637.

sagudev:
This PR sets [new `id_hash` chunker](https://github.com/web-platform-tests/wpt/pull/40020) as default chunker when running wpt tests via mach.

Fixes #29844, and as a bonus flattens wpt test times for chunks (for layout-2013: [10 min, 30 min] -> [20 min, 25 min])

Test run available [here](https://github.com/sagudev/servo/actions/runs/5207427141).

### DIFF
--- a/python/wpt/run.py
+++ b/python/wpt/run.py
@@ -73,6 +73,9 @@ def run_tests(**kwargs):
     set_if_none(
         kwargs, "host_cert_path", os.path.join(CERTS_PATH, "web-platform.test.pem")
     )
+    # Set `id_hash` as the default chunk, as this better distributes testing across different
+    # chunks and leads to more consistent timing on GitHub Actions.
+    set_if_none(kwargs, "chunk_type", "id_hash")
 
     kwargs["user_stylesheets"].append(os.path.join(SERVO_ROOT, "resources", "ahem.css"))
 


python/wpt/run.py
```diff
@@ -73,6 +73,8 @@ def run_tests(**kwargs):
     set_if_none(
         kwargs, "host_cert_path", os.path.join(CERTS_PATH, "web-platform.test.pem")
     )
+    # set id_hash as default chunker
```
mrobinson, line 76: I think we should leave a more detailed comment here so we remember what this is useful for.

Here's a proposal:

```
# Set `id_hash` as the default chunk, as this better distributes testing across different 
# chunks and leads to more consistent timing on GitHub Actions.
```

mrobinson (approved):
This looks like a really nice change. My only suggestion here is to improve the comment a bit.

mrobinson (approved):

PR #29856, opened 2023-06-08
Improve margin collapse in layout-2020

### LINKED ISSUE #29858: Bottom margin shouldn't collapse with last child with height:0
#29839 broke this testcase:

```html
<!DOCTYPE html>
<div style="height: 0px">
  <div style="margin-bottom: 50px">a</div>
</div>
b
```

"a" and "b" are supposed to overlap instead of being separated by a collapsed margin.

https://drafts.csswg.org/css2/#adjoining-margins says:

> bottom margin of a last in-flow child and bottom margin of its parent if the parent has [auto](https://drafts.csswg.org/css2/#valdef-height-auto) computed height

It does not include `height: 0`.

Loirooriol:
According to https://drafts.csswg.org/css2/#collapsing-margins, bottom
margins should only collapse with the last child if `height` is `auto`.
Also, the note mentions `min-height: 0`, but the normative text doesn't
have such requirement, so I'm dropping it, matching WebKit.

The previous logic is moved into the case of collapsing the top and
bottom margins of the same element, since this can happen either with
`height: auto` or `height: 0`, and requires `min-height: 0`.

---

- [X] These changes fix #29858

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -483,8 +483,7 @@ fn layout_in_flow_non_replaced_block_level(
     let end_margin_can_collapse_with_children = block_is_same_formatting_context &&
         pbm.padding.block_end == Length::zero() &&
         pbm.border.block_end == Length::zero() &&
-        block_size.auto_is(|| Length::zero()) == Length::zero() &&
-        min_box_size.block == Length::zero();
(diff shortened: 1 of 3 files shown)


mrobinson (approved):
@Loirooriol walked me through this change and it looks good to me.

PR #29863, opened 2023-06-08
Fix infinite loop in flexbox algorithm

### LINKED ISSUE #29852: Infinite loop with flexbox in layout-2020

nicoburns:
Only apply step 5c of "resolve flexible lengths" if sum of scaled flexible shrink factors > 0
Probably fixes #29852 (but speculative as I can't get mach to run).

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -976,10 +976,12 @@ impl FlexLine<'_> {
                     let scaled_shrink_factors_sum: Length = unfrozen_items()
                         .map(|(item, _)| scaled_shrink_factor(item))
                         .sum();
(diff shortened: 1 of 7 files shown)


nicoburns:
That looks like it's working to me (TIMEOUT's turning into OK/FAIL seems like what I would expect if an infinite loop bug is fixed).

Loirooriol (approved):
Thanks, I agree this seems the most reasonable solution

PR #29865, opened 2023-06-09
windows workflow use C drive as target directory

sagudev:
Currently we clone to drive D (default drive in GitHub Actions), then copy to drive C, then build from there (due to low disk space on D drive). In this PR instead of copying to C drive we only set `CARGO_TARGET_DIR` to C drive, the build is reading source files from D and writing target files to C drive. This gets us approx. -20min.

Try-windows build available [here](https://github.com/sagudev/servo/actions/runs/5219706997/jobs/9421863750).

What do we lose? Well copying step was good benchmarker for GitHub runner machines, fast windows machines spend ~5min in copying step while slow machines ~8min.

For the future: I do believe that D drive may be faster for writing, so switching roles could benefit us, but it will require slow (like copying step) clone.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -40,7 +40,7 @@ env:
   LAYOUT: "${{ contains(inputs.layout, '2020') && 'layout-2020' || 'layout-2013' }}"
   PACKAGE: "${{ contains(inputs.layout, '2020') && 'windows-msvc-layout2020' || 'windows-msvc' }}"
   CCACHE: "ccache"
-  CARGO_INCREMENTAL: 0
+  CARGO_TARGET_DIR: C:\\a\\servo\\servo\\target
 
 jobs:
   build-win:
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -40,7 +40,7 @@ env:
   LAYOUT: "${{ contains(inputs.layout, '2020') && 'layout-2020' || 'layout-2013' }}"
   PACKAGE: "${{ contains(inputs.layout, '2020') && 'windows-msvc-layout2020' || 'windows-msvc' }}"
   CCACHE: "ccache"
-  CARGO_INCREMENTAL: 0
```
mrobinson, line 43: Why are you removing `CARGO_INCREMENTAL` here?

sagudev: This was actually added while testing sccache, but as it is not working for windows builds it was removed except this line.

sagudev: But this change does not benefit performance, I had a commit to remove it before this somewhere.

mrobinson:
This seems like a nice speedup.

jdm:
Fantastic! I've always been annoyee by the copying requirement, so this is a great workaround.

PR #29866, opened 2023-06-09
Run unit tests on the main Linux workflow

mrobinson:
Not running unit tests on the main Linux workflow makes it faster, but also allows merging changes that break the unit tests. This change fixes that.

---

- [x] These changes do not require tests because they are CI changes.


atbrakhi (approved):

mrobinson:
Thanks for the review!

PR #29870, opened 2023-06-14
Layout 2020: Correct rendering of floated root

mrobinson:
Fix two issues around floating a root element:

1. In the StackingContext code handle the case where a root element is a Float fragment and not a Box fragment. This fixes a debug assertion failure in the css/CSS2/float/float-root.html test.
2. When initializing the SequentialLayoutState, use the containing block width as the maximum inline float placement position instead of infinity. This fixes the rendering of css/CSS2/float/float-root.html.

Note that css/CSS2/float/float-root.html was passing before, because both the test and reference were subject to the same bug. This fixes a couple other tests as well.

---

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -429,11 +429,12 @@ impl StackingContext {
         };
 
         let fragment = first_stacking_context_fragment.fragment.borrow();
-        let box_fragment = if let Fragment::Box(box_fragment) = &*fragment {
-            box_fragment
-        } else {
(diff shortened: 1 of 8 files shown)


mrobinson:
> * FAIL [expected PASS] /_mozilla/css/abs_float_pref_width.html
> * FAIL [expected PASS] /_mozilla/css/absolute_block_format_ctx.html

It seems these two tests are failing because we do not yet implement preferred width calculation properly when floats are involved.

Loirooriol (approved):

Loirooriol:
@bors-servo try=wpt-2020

#29887 should have avoided the failures.

PR #29878, opened 2023-06-15
Fix servo crate name confliction on documentation

### LINKED ISSUE #32412: Docs are not built for libservo
[The docs for libservo](https://doc.servo.org/servo/) are shadowed by the docs for servoshell’s bin crate, because they both have the crate name “servo”, and cargo currently has no way of disambiguating this (rust-lang/cargo#6313). We could:

- Rename libservo from `servo` to `libservo` (#29878)
- Rename servoshell(bin) from `servo` to `servoshell`, accepting that the binary name will change
  - servoshell is probably a well-enough-known name at this point
  - This is not unusual among our peers (e.g. [content_shell](https://chromium.googlesource.com/chromium/src/+/028b6a9755c6428424d2a43323c23604aa4c9218/docs/testing/web_tests_in_content_shell.md#As-a-simple-browser), [minibrowser](https://trac.webkit.org/wiki/BuildingGtk?version=145#Runningfromgit))
- Rename servoshell(bin) from `servo` to `servoshell`, while keeping the binary name as `servo`
  - This either requires a post-build step in mach, or RUSTC_BOOTSTRAP=1 + [different-binary-name](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#different-binary-name)
- Disable doc output for servoshell(bin) somehow?
  - No docs would be lost by doing this, because servoshell(bin) is just a stub ([#31439](https://github.com/servo/servo/pull/31439))

wusyong:
---

- [x] These changes fix #32412

- [x] There are tests for these changes OR it should be able to build all ports

The name of winit binary is conflicted with servo crate.
We can't read [documentation of servo library](https://doc.servo.org/servo/index.html) crate at the moment.
It's actually the winit binary documentation.

### DIFF
--- a/components/servo/Cargo.toml
+++ b/components/servo/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 publish = false
 
 [lib]
-name = "servo"
+name = "libservo"
 path = "lib.rs"
 crate-type = ["rlib"]
 
--- a/ports/gstplugin/Cargo.toml
+++ b/ports/gstplugin/Cargo.toml
@@ -25,7 +25,7 @@ gstreamer-gl-sys = { version = "0.8", features = ["wayland"] }
 gstreamer-sys = "0.8"
 gstreamer-video = "0.15"
 lazy_static = { workspace = true }
-libservo = { path = "../../components/servo" }
+servo = { package = "libservo", path = "../../components/servo" }
 log = { workspace = true }
 servo-media = { git = "https://github.com/servo/media" }
 sparkle = { workspace = true }
--- a/ports/libsimpleservo/api/Cargo.toml
+++ b/ports/libsimpleservo/api/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 [dependencies]
 getopts = { workspace = true }
 ipc-channel = { workspace = true }
-libservo = { path = "../../../components/servo" }
+servo = { package = "libservo", path = "../../../components/servo" }
 log = { workspace = true }
 servo-media = { git = "https://github.com/servo/media" }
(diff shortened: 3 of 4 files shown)


mrobinson:
Hrm. I have been considering this change for a while, but I'm not sure if it's the right one. Renaming the library because of a documentation issue seems like the wrong incentive structure. Is there a way to fix the documentation without the rename. Perhaps we could move the runner/winit port documentation out of rustdoc and simply disable documentation for that crate?

wusyong:
The tracking issue of this bug is [cargo/#6313](https://github.com/rust-lang/cargo/issues/6313). I didn't follow through all conversations, but the beginning of the thread said this will be considered a hard error in the future.

Adding `doc = false` to the bin will just make `mach build` generate nothing. Another way I could think of is `cargo doc -p libservo` with features you specified and add this to `mach`. But this still won't solve the winit binary name is conflicted with servo library.

mrobinson:
An alternate approach here might be to rename the runner project, but have it produce the a binary with the name "servo." It's my understanding that the binary artifact itself will not conflict, but the documentation might.

wusyong:
Yeah, I tried this before. Artifacts will not conflict, but doc will. Carge doc will still warn about this if there are crates with same names despite the fact that they might have different package names.

wusyong:
I think another workaround could be adding another command argument like `--lib` instead.
I can look into it if you think this is better.
But we'll need to verify other mach commands because they all go through `run_cargo_build_like_command` function.

mrobinson:
I suspect that if changing the name of the crate is on the only way to fix this, it makes more sense to change the name of the crate of the binary rather than the crate of the library.

PR #29880, opened 2023-06-15
Don't pass float stacking containers up to parent stacking contexts

mrobinson:
Don't pass up float stacking containers to parent stacking contexts

Instead of passing up stacking containers created by floated content,
keep them in their original parent stacking containers. This is in in
line with specification text for stacking containers:

> To paint a stacking container, given a box root and a canvas canvas:
>
>   1.  Paint a stacking context given root and canvas, treating root as
>       if it created a new stacking context, but omitting any positioned
>       descendants or descendants that actually create a stacking context
>       (letting the parent stacking context paint them, instead).

---

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -200,6 +200,11 @@ pub(crate) enum StackingContextType {
     PseudoAtomicInline,
 }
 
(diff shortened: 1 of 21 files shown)


Loirooriol (approved):

mrobinson:
Thanks for the review.

PR #29881, opened 2023-06-16
Remove JS stream usage

### LINKED ISSUE #29088: Rewrite readable stream implementation without spidermonkey builtin object
The next spidermonkey upgrade will bring in https://bugzilla.mozilla.org/show_bug.cgi?id=1795914, which removes the JS implementation. We will need to port https://bugzilla.mozilla.org/show_bug.cgi?id=1730589 and subsequent changes from https://searchfox.org/mozilla-central/source/dom/streams/ReadableStream.cpp before we can upgrade.

jdm:
Rewrite readable stream code to avoid use of JS streams. Fixes #29088.


jdm:
Lots of those failures are expected since the current DOM ReadableStream interface is not standards-compliant at all.

jdm:
Some timeouts (like `/url/urlencoded-parser.any.html`) are caused by a problem with ReadableStream::new_from_bytes. When it's called, there readable stream doesn't have a reader yet so the notification doesn't go anywhere.

components/script/dom/readablestream.rs
```diff
@@ -89,16 +112,18 @@ impl ReadableStream {
...
         realm: InRealm,
     ) -> Result<DomRoot<ReadableStream>, ()> {
-        if !IsReadableStream(obj) {
-            return Err(());
-        }
+        /*if !IsReadableStream(obj) {
+                return Err(());
+        }*/
+
+        root_from_object(obj, *cx)
```
andreubotella, line 119: I'm not sure I fully grasp the behaviors resulting from this change.

As far as I understand, before this change, there was the SpiderMonkey native `ReadableStream`s, and then there was the Servo `ReadableStream` "interface", which was only a Rust-level wrapper over native JS streams, and was never exposed to JS – and this method made a Rust `DomRoot<ReadableStream>` from the JS object. But this PR exposes objects of the Rust-level interface to JS, while `globalThis.ReadableStream` is still the SpiderMonkey implementation. So `obj` could be either a SpiderMonkey or a Servo `ReadableStream`.

The documentation for `root_from_object` says it returns `Err` if `obj` is not the right type of object. But in my tests it seems like neither `new Response(new ReadableStream())` or `new Response(new File([], "").stream())` throw. So I'm not sure what's going on. How are the fields of the Rust `ReadableStream` struct populated when it's being created from a SpiderMonkey stream?

jdm: Great question! With these changes, we shouldn't be creating any JS stream objects; they will only be created from the native ReadableStream::new_from_bytes constructor. The only caller of ReadableStream::from_js is in generated code (https://github.com/servo/servo/blob/ffccc5c88b8d47efc777b0bc9d2620bed4bc84d7/components/script/dom/bindings/codegen/CodegenRust.py#L923), and that should only be called if we're (for example) passing a readable stream object into a DOM method that takes a union (eg. the Request/Response constructor).

jdm: Sorry, I'm not actually sure that I really answered what you were asking. I'm actually starting to doubt myself.

jdm: Before these changes, we actually had our Servo interface for ReadableStream that was exposed all the time, and passing `new ReadableStream()` to a function like Response's constructor called `ReadableStream::from_js` which would create a new Servo object and store the JS stream in a field. However, the actual generated code for the conversion looks like this:
```
        if value.get().is_object() {
            match ReadableStreamOrXMLHttpRequestBodyInit::TryConvertToReadableStream(SafeJSContext::from_ptr(cx), value) {
                Err(_) => return Err(()),
                Ok(Some(value)) => return Ok(ConversionResult::Success(ReadableStreamOrXMLHttpRequestBodyInit::ReadableStream(value))),
                Ok(None) => (),
            }
            match ReadableStreamOrXMLHttpRequestBodyInit::TryConvertToBlob(SafeJSContext::from_ptr(cx), value) {
                Err(_) => return Err(()),
                Ok(Some(value)) => return Ok(ConversionResult::Success(ReadableStreamOrXMLHttpRequestBodyInit::Blob(value))),
                Ok(None) => (),
            }
```
And `TryConvertToReadableStream` looks like this:
```
            match ReadableStream::from_js(cx, value.get().to_object(), InRealm::Already(&in_realm_proof)) {
                Ok(val) => val,
                Err(()) => {
                return Ok(None);
                }
            }
```
So passing a JS readable stream object into the function just means it is silently initialized with an empty body instead.

andreubotella: That doesn't seem to be what's going on. `new Response(new ReadableStream()).text()` should never resolve, since the passed stream never closes, and that is the behavior you see in Servo before this PR. With this PR, however, it resolves immediately with the string `"[object ReadableStream]"`.

Since `new ReadableStream()` creates a SpiderMonkey stream, which is not an instance of Servo's `_ReadableStream` WebIDL interface, `ReadableStream::from_js` on such an object fails. JS streams are, however, convertible to strings, and that is the variant of the `BodyInit` that it goes with.

But since the whole point of this PR is to get JS streams closer to being gone, maybe living with this bug is fine until everything is moved off of JS streams.

andreubotella:
The `is_disturbed` field of `ReadableStream` isn't used. It should be set to true on `read_a_chunk` and checked in the `is_disturbed` method.

jdm:
Ok, I think the issue you pointed out (the ReadableStream object is converted to a `[object ReadableStream]` string) explains the majority of the remaining errors, with the rest being caused by Servo's ReadableStream interface missing the `getReader()` API and the tests failing any `stream instanceof ReadableStream` test.

jdm:
@bors-servo try=wpt

This should fail any test that uses the ReadableStream constructor or tries to call `getReader()` on a ReadableStream object.

Cargo.toml
```diff
@@ -107,3 +107,7 @@ immeta = { git = "https://github.com/fabricedesre/immeta.git" }
 [patch."https://github.com/servo/webrender"]
 webrender = { git = "https://github.com/jdm/webrender", branch = "crash-backtrace" }
 webrender_api = { git = "https://github.com/jdm/webrender", branch = "crash-backtrace" }
+
+[patch."https://github.com/servo/mozjs"]
+mozjs_sys = { git = "https://github.com/jdm/mozjs", branch = "no-streams" }
+mozjs = { git = "https://github.com/jdm/mozjs", branch = "no-streams" }
```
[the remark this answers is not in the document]
sagudev: It's probably better to disable directly in mozjs as streams as a whole will be gone in next SM bump anyway.

components/script/dom/bindings/codegen/parser/WebIDL.py
sagudev, line 1: It may be better to not change this file as it is upstreamed from mozilla. You can achieve same thing using `use_builtin_readable_stream`. See https://github.com/sagudev/servo/commit/8f7196818dd076e80c8cb227f124f10f3bb0c6ac#diff-27c57f586c93e0626e7e988aeb22181aec12a417825fd07c922250538de52cc3

jdm: Yeah, I just applied the inverse of the existing patch at https://github.com/servo/servo/commit/42d8269ac97a231e6a24677af2e2e60197210990 .

sagudev: Hm, I completely forget that I wrote it:sweat_smile:
It did look weird, why would support for streams be still left in.

mrobinson:
Sorry for causing the merge conflict on this change. I'm happy to rebase it and bring it back to life.

jdm:
Don't worry, I'll take care of it.

mrobinson:
@jdm Should this one be closed  for now since we have updated SpiderMonkey by reverting the streams usage?

jdm:
Nope! I still think we want to proceed with this work.

andreubotella:
It seems like it'd be better if, now that there's a temporary solution for upgrading SM, this PR could be reworked to add non-SM streams as a compile-time feature, so it could be developed over time, perhaps by multiple contributors, without regressions.

jdm:
That's a fair point! My instinct is that doing all that work in one PR is going to be pretty overwhelming, though. I'm open to ideas if how to split up the work in useful ways!

components/script/dom/readablestream.rs
```diff
@@ -390,10 +423,13 @@ pub enum ExternalUnderlyingSource {
 }
 
 #[derive(JSTraceable, MallocSizeOf)]
+#[unrooted_must_root_lint::must_root]
 struct ExternalUnderlyingSourceController {
     /// Loosely matches the underlying queue,
     /// <https://streams.spec.whatwg.org/#internal-queues>
     buffer: RefCell<Vec<u8>>,
+    #[ignore_malloc_size_of = "Rc is hard"]
+    pending_reader: RefCell<Option<Rc<Promise>>>,
```
gterzian, line 432: I would rather keep those on the dom struct, in some ways `ExternalUnderlyingSourceController` doesn't really belong here at all because it's the "native" pure Rust part, and should eventually perhaps be separated out into `shared/script`(similar to Blob and Messageport).

components/script/dom/readablestream.rs
```diff
@@ -427,73 +464,142 @@ impl ExternalUnderlyingSourceController {
         None
     }
 
+    fn add_reader(&self, cx: SafeJSContext, promise: Rc<Promise>) {
+        assert!(self.pending_reader.borrow().is_none());
+        *self.pending_reader.borrow_mut() = Some(promise.clone());
+        let buflen = self.buffer.borrow().len();
+        self.maybe_signal_available_bytes(cx, &promise.global(), buflen);
```
gterzian, line 471: The problem with this approach is that it will only wake up the reader in reponse to `enqueue_native`, and in practice the queuing could also be done by a JS underlying source([example](https://streams.spec.whatwg.org/#ref-for-push-source%E2%91%A1)).

That was the part that SM dealt with internally, I think.

gterzian (commented):
Took a quick look, couple of comments.

jdm:
I am not actively working on it.

jdm:
Closing this since I'm not working on it.

PR #29882, opened 2023-06-16
Fix windows nightly build failure

mukilan:
PR #29865 changed the target directory for windows to be different from the source folder. However, the 'upload-nightly' command assumes the packages are stored in the default cargo target directory (./target), so the nightly jobs for windows were failing to find the packages.

This PR fixes the 'upload-nightly' command to account for non-default target directories specified via CARGO_TARGET_DIR.

---
- [x] These changes do not require tests because they fix a bug in packaging logic.


mrobinson (approved):
Thank you for this fix.

mukilan:
@bors-servo r=mrobinson

Thanks for the review!

PR #29887, opened 2023-06-17
Handle floats in BlockContainer::inline_content_sizes

### LINKED ISSUE #29874: Layout 2020: Implement preferred widths for floats
Floats need to be able to give their preferred widths and influence the preferred widths of their containers.

Loirooriol:
Typically, block-level contents are stacked vertically, so this was just taking the maximum size among all contents. However, floats can be stacked horizontally, so we need to sum their sizes.

---

- [X] These changes fix #29874

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -25,6 +25,8 @@ use crate::ContainingBlock;
 use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use rayon_croissant::ParallelIteratorExt;
 use servo_arc::Arc;
+use style::computed_values::clear::T as Clear;
+use style::computed_values::float::T as Float;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
 use style::values::computed::{Length, LengthOrAuto};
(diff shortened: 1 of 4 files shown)


Loirooriol:
The tests changes are kinda by chance since `width: max/min-content` aren't supported yet. Should I add more tests, or does it suffice that with this patch, #29870 will no longer have regressions?

components/layout_2020/flow/mod.rs
```diff
@@ -143,21 +145,91 @@ impl BlockContainer {
...
-                })
-                .reduce(ContentSizes::max)
-                .unwrap_or_else(ContentSizes::zero),
+            Self::BlockLevelBoxes(boxes) => {
+                let get_box_info = |box_: &ArcRefCell<BlockLevelBox>| {
+                    let size = box_
+                        .borrow_mut()
+                        .inline_content_sizes(layout_context, writing_mode);
+                    if let BlockLevelBox::OutOfFlowFloatBox(ref float_box) = *box_.borrow_mut() {
+                        let style_box = match &float_box.contents {
```
mrobinson, line 154: I think this could be turned into a function `IndependentFormattingContext::style()` which returns a reference to the style.

components/layout_2020/flow/mod.rs
```diff
@@ -143,21 +145,91 @@ impl BlockContainer {
...
+                            },
+                        };
+                        (size, style_box.float, style_box.clear)
+                    } else {
+                        // The element may in fact have clearance, but the logic below ignores it,
+                        // so don't bother retrieving it from the style.
+                        (size, Float::None, Clear::None)
+                    }
+                };
+                struct AccumulatedData {
```
mrobinson, line 169: Thanks for documenting the struct members! It might be nice to add a quick line for the struct itself as well.

components/layout_2020/flow/mod.rs
```diff
@@ -143,21 +145,91 @@ impl BlockContainer {
...
+                    /// The maximum size seen so far, not including trailing floats.
+                    max_size: ContentSizes,
+                    /// The size of the trailing floats with 'float: left'.
+                    left_floats: ContentSizes,
+                    /// The size of the trailing floats with 'float: right'.
+                    right_floats: ContentSizes,
+                }
+                impl AccumulatedData {
+                    fn max_size_including_floats(&self) -> ContentSizes {
```
mrobinson, line 178: Maybe this should be called something like `max_size_including_uncleared_floats`?

mrobinson (approved):
This is a really elegant solution. I only have a few small nits, but otherwise I think it looks good.

There's enough going on here that this should probably move to a free function like `calculate_inline_content_size_for_block_level_boxes` which gives a short rustdoc description of the approach taken along the lines of:

Find the maximum inline size of all the block level boxes and the summation of floated boxes, respecting the `clear` property of their siblings.

components/layout_2020/flow/mod.rs
Loirooriol: Actually `IndependentFormattingContext::style()` already existed :sweat_smile:

PR #29901, opened 2023-06-21
Default nightly builds to layout 2020

mukilan:
This PR switches the filenames used by nightly builds so that the default `servo-latest.{ext}` packages use the 2020 engine and the 'servo-latest-legacy-layout.{ext}` packages use the 2013 engine.

Since the platform workflows are reused by  the main, quick-check, PR and WPT import flows and since the 2013/2020 flags are also controlled by try branches, I've kept the changes isolated to  `nightly.yml` and `package_commands.py` and have avoided changing the input parameters. If we want to deault all the workflows to 2020, I can decline this PR and raise a new one.

---

- [x] These changes do not require tests because they modify packaging logic.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -59,11 +59,11 @@ jobs:
     needs:
       - create-draft-release
       - upload-linux
-      - upload-linux-2020
+      - upload-linux-2013
       - upload-win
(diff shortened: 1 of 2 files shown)


mrobinson (approved):
This seems like a good first step.

PR #29902, opened 2023-06-21
Fix interaction of clearance with margin collapse

### LINKED ISSUE #29896: Layout 2020: clearance should prevent margin collapse
https://drafts.csswg.org/css2/#adjoining-margins says that two margins are not adjoining if there is some clearance between them

```html
<!DOCTYPE html>
<section style="line-height: 20px; width: 100px">
  aaaa
  <div style="float: left; height: 70px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="clear: both"></div>
  </div>
  bbbb
  <div style="float: left; height: 71px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="clear: both"></div>
  </div>
  cccc
  <div style="float: left; height: 70px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="clear: both"></div>
    <div style="margin: 50px 0"></div>
  </div>
  dddd
  <div style="float: left; height: 71px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="clear: both"></div>
    <div style="margin: 50px 0"></div>
  </div>
  eeee
  <div style="float: left; height: 70px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="margin: 50px 0"></div>
    <div style="clear: both"></div>
  </div>
  ffff
  <div style="float: left; height: 71px; width: 10px; background: cyan;"></div>
  <div style="margin: 50px 0; outline: solid">
    <div style="margin: 50px 0"></div>
    <div style="clear: both"></div>
  </div>
  gggg
</section>
```

| Servo 2013 | Servo 2020 |Gecko  | Blink | WebKit |
| ------------- | ------------- | ------------- | ------------- | ------------- |
|  |  |  |   |  |

I think Blink and WebKit are right.

Loirooriol:
In #29897 I did the simple naive thing, but it wasn't entirely correct. This patch tries to address the problems. In particular:

 - Clearance should prevent margins from collapsing through if it happens between them, as opposed to on the element that owns them.
 - The margins of an element with clearance can still collapse through, and collapse with other siblings as normal, but the resulting margin can't collapse with the bottom margin of the parent.

---

- [X] These changes fix #29896

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -306,10 +306,11 @@ fn layout_block_level_children_in_parallel(
         })
         .collect();
 
+    let (content_block_size, collapsible_margins_in_children) = placement_state.finish();
     FlowLayout {
         fragments,
-        content_block_size: placement_state.current_block_direction_position,
-        collapsible_margins_in_children: placement_state.collapsible_margins_in_children(),
+        content_block_size,
+        collapsible_margins_in_children,
     }
 }
 
@@ -349,10 +350,11 @@ fn layout_block_level_children_sequentially(
         })
         .collect();
 
+    let (content_block_size, collapsible_margins_in_children) = placement_state.finish();
     FlowLayout {
         fragments,
-        content_block_size: placement_state.current_block_direction_position,
-        collapsible_margins_in_children: placement_state.collapsible_margins_in_children(),
+        content_block_size,
+        collapsible_margins_in_children,
     }
 }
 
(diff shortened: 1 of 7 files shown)


Loirooriol:
@bots-servo try=wpt-2020

mrobinson (approved):
This looks good to me though maybe consider the two changes suggested when we looked at this together:
- have `PlacementState::finish()` consume the `PlacementState` and return the values needed for `FlowLayout` to avoid misuse.
- add a comment / spec link some of the new code in `PlacementState` since it's a bit tricky.

PR #29903, opened 2023-06-21
Simplify the build instructions in README.md

mrobinson:
- Add a chocolatey configuration that installs all dependencies and split
  the instructions into a normal and manual installation section. In
  addition fix a warning print statement for the Windows build that would
  trigger a Python exception.
- Simplify the build instructions in the README and move all manual
  setup details to: https://github.com/servo/servo/wiki/Building

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,253 +10,45 @@ for help getting started.
 
 Visit the [Servo Project page](https://servo.org/) for news and guides.
 
-## Setting up your environment
+## Build Setup
 
-### Rustup.rs
-
-Building servo requires [rustup](https://rustup.rs/), version 1.8.0 or more recent.
-If you have an older version, run `rustup self update`.
-
-To install on Windows, download and run [`rustup-init.exe`](https://win.rustup.rs/)
-then follow the onscreen instructions.
-
-To install on other systems, run:
-
-```sh
-curl https://sh.rustup.rs -sSf | sh
-```
-
-This will also download the current stable version of Rust, which Servo won’t use.
-To skip that step, run instead:
-
-```
-curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain none
-```
-
-See also [Other installation methods](
-https://rust-lang.github.io/rustup/installation/other.html)
-
-### Other dependencies
-
-Please select your operating system:
 * [macOS](#macos)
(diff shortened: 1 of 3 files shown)


mrobinson:
I've expanded the scope of this a bit to a general clean up of the build instructions in the `README.md` file. The contents pertaining to manual setup of the build and to very old (> 5 years) software are now in https://github.com/servo/servo/wiki/Building.

Finally, information about `CentOS` is removed since the distribution is now defunct.

mrobinson:
Rendered contents can be seen here: https://github.com/mrobinson/servo/tree/improve-windows-build-instructions

README.md
```diff
@@ -381,3 +173,6 @@ There are lots of mach commands you can use. You can list them with `./mach
 
 
 The generated documentation can be found on https://doc.servo.org/servo/index.html
+
+[manual-build]: https://github.com/servo/servo/wiki/Building#manual-build-setup
+[windows-tips]: https://github.com/servo/servo/wiki/Building#troubleshooting-the-windows-build
```
mrego, line 178: I don't find how to leave a comment on an unmodified line, but some lines earlier we have:
```sh
./mach build --dev
./mach run tests/html/about-mozilla.html
```

It'd be nice to add `--dev` on `mach run` too.

mrego:
Building for Android is mentioned in the building section but not in the setup. The setup actually links to https://github.com/servo/servo/wiki/Building-for-Android. Should we just do the same for building or why we only have it in the building section?

README.md
mrobinson: The rest of the README should be cleaned up as well, but I wanted to focus on the build instructions for the first pass.

mrobinson:
> Building for Android is mentioned in the building section but not in the setup. The setup actually links to https://github.com/servo/servo/wiki/Building-for-Android. Should we just do the same for building or why we only have it in the building section?

My hope is to later do another pass and clean up the Building section as well.

README.md
mrego: Ok, this looked so small that we could just do it, but we can do it later too.

mrego (approved):
Thanks for simplifying and updating the README. Looking forward for further updates.

PR #29910, opened 2023-06-22
Fix positioning of statically positioned absolute child of inline box

### LINKED ISSUE #29892: Layout 2020: Absolutes with static insets not positioned properly inside inline blocks
Test case:
```html
<div>
    hello
    <div style="display: inline-block; width: 100px; height: 100px; background: blueviolet;">
        <div style="position: absolute; width: 10px; height: 10px; background: cornflowerblue;"></div>
    </div>
    hello
</div>
```

Layout 2020:
image

Firefox:
image

mrobinson:
---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -659,13 +659,20 @@ fn layout_atomic(
                 containing_block_for_children.style.writing_mode,
                 "Mixed writing modes are not supported yet"
             );
-            // FIXME: Do we need to adjust the static position of the hoisted fragments in the positioning
-            // context somewhere near here?
+
+            let collects_for_nearest_positioned_ancestor = ifc
+                .positioning_context
+                .collects_for_nearest_positioned_ancestor();
+            let mut child_positioning_context =
+                PositioningContext::new_for_subtree(collects_for_nearest_positioned_ancestor);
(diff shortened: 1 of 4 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -659,13 +659,20 @@ fn layout_atomic(
                 containing_block_for_children.style.writing_mode,
                 "Mixed writing modes are not supported yet"
             );
-            // FIXME: Do we need to adjust the static position of the hoisted fragments in the positioning
-            // context somewhere near here?
+
+            let collects_for_nearest_positioned_ancestor = ifc
+                .positioning_context
+                .collects_for_nearest_positioned_ancestor();
+            let mut positioning_context =
```
Loirooriol, line 666: Nit: `layout_block_level_children_in_parallel` and `layout_block_level_children_sequentially` use the name `child_positioning_context`, which may be slightly clearer than `positioning_context`

Loirooriol (approved):
Thanks!

components/layout_2020/flow/inline.rs
mrobinson: Good point. I'll make this change.

mrobinson:
@bors-servo r=Loirooriol

Thanks for the review!

PR #29912, opened 2023-06-22
Don't explicitly disable gold on Windows

mrobinson:
This seems to be working fine now on CI, so I think we can remove this
workaround.

---

- [x] These changes do not require tests because they are just build changes.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -628,8 +628,7 @@ class CommandBase(object):
         if self.config["build"]["rustflags"]:
             env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " " + self.config["build"]["rustflags"]
 
-        # Don't run the gold linker if on Windows https://github.com/servo/servo/issues/9499
-        if self.config["tools"]["rustc-with-gold"] and sys.platform != "win32":
+        if self.config["tools"]["rustc-with-gold"]:
             if subprocess.call(['which', 'ld.gold'], stdout=PIPE, stderr=PIPE) == 0:
                 env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " -C link-args=-fuse-ld=gold"
 


sagudev:
There is no `which` command on windows:

```console
FileNotFoundError: [WinError 2] The system cannot find the file specified

  File "C:\Users\user\servo\python\servo\command_base.py", line 762, in configuration_decorator
  File "C:\Users\user\servo\python\servo\testing_commands.py", line 266, in test_unit
    env=self.build_env(test_unit=True),
  File "C:\Users\user\servo\python\servo\command_base.py", line 632, in build_env
    if subprocess.call(['which', 'ld.gold'], stdout=PIPE, stderr=PIPE) == 0:
  File "C:\Program Files\Python310\lib\subprocess.py", line 345, in call
    with Popen(*popenargs, **kwargs) as p:
  File "C:\Program Files\Python310\lib\subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Program Files\Python310\lib\subprocess.py", line 1438, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
```

sagudev:
We could use [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) which is available from Python 3.3.

sagudev:
Also isn't gold only for ELF files (meaning that it does not work on windows & mac) and the only reason that current script works correctly (if I understand things correctly) is because it detects if gold is present which on unsupported systems is not.

mrobinson:
> Also isn't gold only for ELF files (meaning that it does not work on windows & mac) and the only reason that current script works correctly (if I understand things correctly) is because it detects if gold is present which on unsupported systems is not.

Hrm. Maybe this should really be using the lld linker, but it seems this isn't well supported on Mac + Rust yet. Perhaps we simply can't get away with a non-platform-specific set of linker arguments. :(

PR #29918, opened 2023-06-22
Use mozjs tracing infrastructure

sagudev:
Servos tracing infrastructure was mainly copied to mozjs in https://github.com/servo/mozjs/pull/352. Now this PR tries to switch to it.

Due to moving `JSTraceable` trait to mozjs we can no longer implement trait directly on foreign types (of script crate) due to Rust's orphan rules. This PR resolves the problem with building on top of https://github.com/servo/servo/pull/29926 and introducing `CustomTraceable` trait where custom trace implementations on foreign types are necessary with limitation that they only work inside of Dom objects (those types that use `derive(JSTraceable)`). The final code now looks like this:

```rust
#[derive(JSTraceable)]
struct S {
  js_managed: JSManagedType,

  #[no_trace] // this field will be skipped in tracing
              // SAFETY: Trait shenanigans assures that the type is not JSTraceable
  non_js: NonJSManagedType,

  #[custom_trace] // Extern type implements CustomTraceable that is from servo => no problem with orphan rules
  extern_composed_type: Extern<JSManagedType>,
}
```

Needs https://github.com/servo/mozjs/pull/359.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3704,7 +3704,7 @@ checksum = "903970ae2f248d7275214cf8f387f8ba0c4ea7e3d87a320e85493db60ce28616"
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/servo/mozjs#98b678283d1b0c6d263302ca2aff770aece0cb8e"
+source = "git+https://github.com/servo/mozjs#64711ec2e6dc4595df691bffc7f1e5052ab86c8d"
 dependencies = [
  "cc",
  "lazy_static",
@@ -3717,7 +3717,7 @@ dependencies = [
 [[package]]
 name = "mozjs_sys"
 version = "0.68.2"
-source = "git+https://github.com/servo/mozjs#98b678283d1b0c6d263302ca2aff770aece0cb8e"
+source = "git+https://github.com/servo/mozjs#64711ec2e6dc4595df691bffc7f1e5052ab86c8d"
 dependencies = [
  "bindgen",
  "cc",
--- a/components/jstraceable_derive/lib.rs
+++ b/components/jstraceable_derive/lib.rs
@@ -7,7 +7,7 @@ extern crate syn;
 #[macro_use]
 extern crate synstructure;
 
-decl_derive!([JSTraceable, attributes(no_trace)] =>
+decl_derive!([JSTraceable, attributes(no_trace, custom_trace)] =>
 /// Implements `JSTraceable` on structs and enums
 ///
 /// Example:
@@ -17,6 +17,8 @@ decl_derive!([JSTraceable, attributes(no_trace)] =>
 ///   js_managed: JSManagedType,
 ///   #[no_trace]
 ///   non_js: NonJSManagedType,
+///   #[custom_trace] // Extern type implements CustomTraceable that is in servo => no problem with orphan rules
+///   extern_managed_type: Extern<JSManagedType>,
 /// }
 /// ```
 ///
@@ -30,12 +32,16 @@ decl_derive!([JSTraceable, attributes(no_trace)] =>
 ///             S {
 ///                 js_managed: ref __binding_0,
 ///                 non_js: ref __binding_1,
+///                 extern_managed_type: ref __binding_2,
 ///             } => {
 ///                 {
 ///                     __binding_0.trace(tracer);
 ///                 }
 ///                 {
 ///                     // __binding_1 is not traceable so we do not need to trace it
+///                 }
+///                 {
+///                     <crate::dom::bindings::trace::CustomTraceable>::trace(__binding_2, tracer);
 ///                 }
 ///             },
 ///         }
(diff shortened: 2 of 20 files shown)


sagudev:
cc @jdm

jdm:
Instead of creating wrapper types, is it possible to do this?

```rust
trait EmpyTraceable {}

impl <T: EmptyTraceable> Traceable for T {
  ...
}
```
And then make no_jsmanaged_impl! implement the EmptyTraceable trait for the type?

sagudev:
As you noted it wouldn't work:
```
error[E0210]: type parameter `T` must be used as the type parameter for some local type (e.g., `MyStruct<T>`)
   --> components/script/dom/bindings/trace.rs:186:14
    |
186 | unsafe impl <T: EmptyTraceable> JSTraceable for T {
    |              ^ type parameter `T` must be used as the type parameter for some local type
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local
    = note: only traits defined in the current crate can be implemented for a type parameter
```

Here are options that I did considered (only last two remained feasible):

1. Use seperate `JSTraceable`:

```rust
pub unsafe trait JSTraceable {
    /// Trace `self`.
    unsafe fn trace(&self, trc: *mut JSTracer);
}

unsafe impl<T: Traceable> JSTraceable for T {
    unsafe fn trace(&self, trc: *mut JSTracer) {
        self.trace(trc)
    }
}
```
but we cannot store all JSTraceables in RootedVec, because JStraceable acts as superclass to Traceable (double hooks???)
2. adding servo feature that would use JSTraceable from 1. inside mozjs (introduces cyclic depencies between servo and mozjs)
3. Use `pub use js::gc::Traceable as JSTraceable;` and where problem happens we use wrappers (this PR)
4. like 3. but solve most problems with [`no_trace`](https://github.com/servo/servo/pull/29926)

jdm:
I can see how in theory the RootedVec issue from option 1 could be a problem. I would be interested in learning if it's actually a problem in practice, though, if we implement the mozjs Traceable trait for types defined in script?

sagudev:
> I can see how in theory the RootedVec issue from option 1 could be a problem. I would be interested in learning if it's actually a problem in practice, though, if we implement the mozjs Traceable trait for types defined in script?

`Traceable` is the one from mozjs, `JSTraceable` is defined in mozjs.

Even if we discard double hooking problem, I believe it would still be problematic. Let's look at: https://github.com/servo/servo/blob/9829931c279bbfd54a777830d8c430c03771108f/components/script/dom/bluetooth.rs#L644 which has `Dom<BluetoothDevice>` as inner type. Now `Dom` would need to impl `Traceable` and depend upon `BluetoothDevice` implementing `Traceable`, which depend upon `EventTarget` implementing `Traceable`, which depend upon `Atom` implementing `Traceable`, but `Atom` is external type and we are back to our initial problem.

jdm:
Thank you, that's a good example. After thinking about it some more I'm also convinced about the danger of the two different trace entry points.

sagudev:
Here is how it could look like after building some extensions on top of #29926:
```rust
#[derive(JSTraceable)]
struct S {
  js_managed: JSManagedType,
  #[no_trace] // this field will bi skipped in tracing
  // SAFETY: Trait shenanigans assures that there is no JSTraceable
  non_js: NonJSManagedType,
  #[custom_trace] // Extern type implements CustomTraceable that is from servo so no problem with orphan rules
  // SAFETY: Trait shenanigans assures that here is no JSTraceable (as the decision between two (different) would be as their impl, unsafe; to prevent abuse of CustomTraceable instead of next option)
  extern_composed_type: Extern<JSManagedType>,
  #[custom_trace=hashmap_trace_value] // type does not need to be JSTraceable, but this function must be able to trace it
  // unsafe fn hashmap_trace_value<N, M: JSTraceable>(s: HashMap<N, M>, trc: *mut JSTracer)
  // SAFETY: compiler plugin (hard) could assume safety
  composed_type: HashMap<NonJSManagedType, JSManagedType>,
}
```

Although the last one would require additional logic in compiler plugin and proc macro, which is already complex and hard.

sagudev:
Successful try run: https://github.com/sagudev/servo/actions/runs/5806627530

I will marked it as ready when https://github.com/servo/mozjs/pull/359 lands.

components/script/dom/htmlmediaelement.rs
```diff
@@ -296,6 +296,13 @@ impl From<MediaStreamOrBlob> for SrcObject {
     }
 }
 
+/*#[derive(JSTraceable)]
+pub struct Ss {
+    in_flight_play_promises_queue: ErrorResult,
+    al: Box<[Rc<Promise>]>,
+    all: (Box<[Rc<Promise>]>, ErrorResult),
+}*/
+
```
sagudev, line 305: This needs to be removed

sagudev:
It's ready now. Also includes fix for #30043.

As both commits are standalone I would suggest merge without squash if possible.

jdm (approved):

jdm:
Stupendous!

PR #29923, opened 2023-06-24
Combine all script tests into `test-scripts`

mrobinson:
Remove:
    - tidy self test support from `./mach test`
    - `./mach test-idl`

Adds a `./mach test-scripts` command that is responsible for running all
Python script tests. Run this during the CI to catch regressions in
changes to scripts. The WebIDL tests are still *very* slow and there are
from Gecko, so only run them when "-a" is passed meaning all tests.

---


python/servo/testing_commands.py
```diff
@@ -80,25 +78,12 @@ def __init__(self, context):
     @CommandArgument('params', default=None, nargs="...",
                      help="Optionally select test based on "
                           "test file directory")
-    @CommandArgument('--render-mode', '-rm', default=DEFAULT_RENDER_MODE,
```
mrobinson, line 83: I removed some of these arguments as well that were not used.

jdm (approved):

mrobinson:
This failed because I tried to call mach incorrectly on the Windows CI. Upon thinking about this a bit, I have removed the `test-script` call from the Window CI entirely. I'm not sure if all the tests pass on Windows and it's probably better not to risk it.

PR #29927, opened 2023-06-25
Increase BFC root height to contain floats

### LINKED ISSUE #29823: Layout 2020: Block formatting context with floats should be big enough to contain their floating content
This is a followup to #29757. When a block formatting context has floating content, it should be big enough to contain all of it's floating content, which normally isn't counted in content size as it's not true in flow content. This could probably be implemented as finding the clearance for one more "clear: both" object at the end of the block formatting context layout.

cc @Loirooriol

Loirooriol:
---

- [X] These changes fix #29823

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -96,10 +96,18 @@ impl BlockFormattingContext {
                 .collapsed_through
         );
 
+        // The content height of a BFC root should include any float participating in that BFC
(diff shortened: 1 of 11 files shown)


Loirooriol:
@bors-servo try=wpt-2020

The flex test is still failing locally...

Loirooriol:
Nevermind, it's passing now, previously I must have done something wrong.

mrobinson (approved):
Very nice.

PR #29931, opened 2023-06-26
Update uuid

mrobinson:
This is required to use the new version of gleam and update ipc-channels
without duplicating the dependency in the future.

---


jdm (approved):
Looks like there's one more change needed in unit test code.

mrobinson:
@bors-servo r=jdm

Thanks for the review!

PR #29934, opened 2023-06-26
Create a top-level "third_party" directory

mrobinson:
This directory now contains third_party software that is vendored into
the Servo source tree. The idea is that it would eventually hold
webrender and other crates from mozilla-central as well with a standard
patch management approach for each.

---


mrobinson:
@emilio: Followup to the conversation in the TSC meeting today. This is the proposal for managing vendored dependencies in Servo. There is a top-level `third_party` directory; in this PR it only contains pre-existing vendored dependencies. Eventually this would have a directory for `style`, `selectors`, and `webrender` though.

atbrakhi (approved):

mrobinson:
@bors-servo retry
 - Network issues

PR #29939, opened 2023-06-27
Properly position floats when subsequent boxes collapse margins with containing block

### LINKED ISSUE #29915: Layout 2020: Floats can overlap since #29899
```html
<!DOCTYPE html>
  <div>
    <div style="margin-top: 50px; height: 25px; background: green"></div>
    <div style="float: left; width: 50px; height: 50px; background: magenta"></div>
  </div>
</div>
```

Before #29899, it behaved like this:

It was wrong because the cyan float was higher than its containing block (#29886).

Now it appears at the correct position, but the magenta float is not aware of the change and then they overlap:

The correct behavior would be:

mrobinson:
Margins should be able to collapse through floats when collapsing with
parent blocks (the containing block). To properly place floats in this
situation, we need to look at these subsequent floats to find out how
much of the margin will collapse with the parent.

This initial implementation is very basic and the second step would be
to cache this in order to avoid having to constantly recalculate it.

Fixes #29915.

---

- [x] These changes fix #29915.

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -10,9 +10,7 @@ use crate::context::LayoutContext;
 use crate::dom::NodeExt;
 use crate::dom_traversal::{Contents, NodeAndStyleInfo};
 use crate::formatting_contexts::IndependentFormattingContext;
-use crate::fragment_tree::{
-    BoxFragment, CollapsedBlockMargins, CollapsedMargin, FloatFragment, Fragment,
-};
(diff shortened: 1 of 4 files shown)


Loirooriol:
@bors-servo try=wpt-2020

[040a2d85b..e23aeb452](https://github.com/servo/servo/compare/040a2d85b..e23aeb452) Rebase
[e23aeb452..206bac772](https://github.com/servo/servo/compare/e23aeb452..206bac772) Fix

mrobinson:
@bors-servo r+

@Loirooriol and I worked on this together live, reviewing each other's work as we went along.

PR #29940, opened 2023-06-27
Make Layout 2013 style a superset of Layout 2020 style

mrobinson:
This is the first step toward removing the compile-time difference in
style between Layout 2020 and Layout 2013. This change:

- Adds `attr()` CSS content support to legacy layout
- Adds parsing support for some unsupported features such as `display: flow-root` and `display: contents`.

---


Loirooriol (approved):
In general claiming to support things that aren't really supported is not a good thing, but I guess that's fine since Servo is still very experimental and layout 2013 is no longer the default.

PR #29945, opened 2023-06-28
Properly position floats when subsequent boxes collapse margins with containing block (2)

### LINKED ISSUE #29944: Layout 2020: Floats can still overlap due to margin collapse
#29915 used a lookahead to try to find the margin that will collapse with the containing block, but it only takes into account the siblings of the float and their descendants. Uncles aren't iterated, so this testcase still fails:

```html
<!DOCTYPE html>
  <div>
    <div>
    </div>
    <div style="margin-top: 50px; height: 25px; background: green"></div>
    <div style="float: left; width: 50px; height: 50px; background: magenta"></div>
  </div>
</div>
```

Loirooriol:
PR #29939 tried to address this but missed various cases.

---

- [X] These changes fix #29944

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -62,7 +62,7 @@ pub(crate) enum BlockLevelBox {
 }
 
 impl BlockLevelBox {
-    fn find_block_margin_collapsing_with_parent_for_floats(
+    fn find_block_margin_collapsing_with_parent(
         &self,
         collected_margin: &mut CollapsedMargin,
         containing_block_writing_mode: WritingMode,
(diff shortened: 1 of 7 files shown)


mrobinson (approved):
Looks like there is a test failing, but this is clearly a progression. Great fix.

Loirooriol:
@bors-servo r=mrobinson

Yes, I forgot to `git add` the test expectation. Thanks for the review!

Loirooriol:
@bors-servo retry

Unit tests seem fine locally...

Loirooriol:
BTW, if I apply a patch to fix clearance, `adjoining-float-nested-forced-clearance-002.html` fails because there is a BFC root that is not avoiding floats (we still have to implement that). I'm adding some colors to see what's going on:



The yellow BFC should avoid the magenta and cyan floats and then force the green container to grow taller, like in Blink/WebKit:



Interestingly, Firefox is also failing:

PR #29946, opened 2023-06-29
Layout 2020: implement clearance as Option<Length>

### LINKED ISSUE #29904: Clearance can be negative
https://drafts.csswg.org/css2/#clearance says

> Note: The clearance can be negative or zero

I need to check the spec in detail, but I suspect that this is one of these cases:

```html
<!DOCTYPE html>
text
<div style="outline: solid; width: 100px">
  <div style="float: left; width: 25px; height: 25px; background: cyan"></div>
  <div style="clear: both; height: 50px; margin-top: 100px; background: magenta"></div>
</div>
```

All Firefox, Blink and WebKit do something like this:

But the margin separates the cyan and magenta boxes in both layout 2013 and 2020.

Also, not sure if having a clearance of zero might be different than having no clearance in some corner cases.

Loirooriol:
Clearance was implemented as a Length, where zero meant no clearance. However, having a clearance of 0px should be different than having no clearance, since the former can still prevent margin collapse.

This patch keeps the existing behavior, so it won't be possible to get a clearance of Some(Length::zero()), but it prepares the terrain for a follow-up to fix calculate_clearance to return the proper thing.

---

- [X] These changes do not require tests because there should be no change in behavior

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -183,7 +183,7 @@ impl BlockFormattingContext {
         // The content height of a BFC root should include any float participating in that BFC
         // (https://drafts.csswg.org/css2/#root-height), we implement this by imagining there is
         // an element with `clear: both` after the actual contents.
-        let clearance = sequential_layout_state.map_or(Length::zero(), |sequential_layout_state| {
+        let clearance = sequential_layout_state.and_then(|sequential_layout_state| {
             sequential_layout_state.calculate_clearance(ClearSide::Both)
         });
 
@@ -191,7 +191,7 @@ impl BlockFormattingContext {
             fragments: flow_layout.fragments,
             content_block_size: flow_layout.content_block_size +
                 flow_layout.collapsible_margins_in_children.end.solve() +
-                clearance,
+                clearance.unwrap_or_else(Length::zero),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -183,15 +183,15 @@ impl BlockFormattingContext {
...
-        let clearance = sequential_layout_state.map_or(Length::zero(), |sequential_layout_state| {
+        let clearance = sequential_layout_state.and_then(|sequential_layout_state| {
             sequential_layout_state.calculate_clearance(ClearSide::Both)
         });
 
         IndependentLayout {
             fragments: flow_layout.fragments,
             content_block_size: flow_layout.content_block_size +
                 flow_layout.collapsible_margins_in_children.end.solve() +
-                clearance,
```
mrobinson, line 194: It's slightly more efficient to do `clearance.unwrap_or_else(Length::zero)` here.

components/layout_2020/flow/mod.rs
```diff
@@ -776,7 +778,9 @@ fn layout_in_flow_non_replaced_block_level(
 
     let content_rect = Rect {
         start_corner: Vec2 {
-            block: pbm.padding.block_start + pbm.border.block_start + clearance,
+            block: pbm.padding.block_start +
+                pbm.border.block_start +
+                clearance.unwrap_or(Length::zero()),
```
mrobinson, line 783: I think all of the `unwrap_or` calls in this change could be `unwrap_or_else`.

components/layout_2020/fragment_tree/box_fragment.rs
```diff
@@ -31,7 +31,7 @@ pub(crate) struct BoxFragment {
     pub border: Sides<Length>,
     pub margin: Sides<Length>,
 
-    pub clearance: Length,
+    pub clearance: Option<Length>,
```
mrobinson, line 34: Now might be a good time to add documentation for this member

mrobinson (approved):
Looks good! I have a few nits, but this is definitely something that we need.

Issue #29815, opened 2023-05-30
Investigate windows unit-tests


sagudev:
Unit-tests step in windows is currently running ~1h (like normal build); while linux and mac are runnig only ~15 min. This was caused by switching to resolver 2 in https://github.com/servo/servo/commit/6a8ba164395574e31179d8ccb79fd903bb75ddb5. I suspect there may be wrong features selected on mozangle; which could cause a rebuild.

dvtkrlbs:
It also seems linke the mozangle build is really poorly parallilized or it spends some time by doing nothing. Looking at the resource usage it spends up to 5 minutes on the mozangle step with non noticable cpu usage.

sagudev:
A while back I tried to separate windows workflow for build and unit-test so that they could run in parallel, but look like unit-testing depends upon building (only on windows), which is a bug in my opinion.

Relevant code is here: https://github.com/servo/servo/blob/master/python/servo/command_base.py#L619

My plan is to remove this dirty hack so unit-test could be run standalone and then split the workflow.

mrobinson:
The unit test step in Windows is now something like 15 minutes in comparison to 10 minutes for Linux. There's improvements to be made, but I think this can be closed.

PR #29947, opened 2023-06-29
 Update mozangle, cc, and cmake

### LINKED ISSUE #29674: mozangle does not build on Fedora 38
Feel free to ask me any follow-up questions!

mrobinson:
This also moves some environment variable configuration to the shared
`build_env()` method, because previously clang was only being chosen for
running `./mach build` and not `./mach test-unit` which was leading to
rebuilds and thus build failures when running `test-unit`. I guess the
cmake crate does not expect the compiler to change between subsequent
runs.

Fixes #29674

---

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -496,7 +496,7 @@ class CommandBase(object):
             'vcdir': vcinstalldir,
         }
 
-    def build_env(self, is_build=False, test_unit=False):
+    def build_env(self, is_build=False):
         """Return an extended environment dictionary."""
         env = os.environ.copy()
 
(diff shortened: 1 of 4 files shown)


mrobinson:
Going to try this update again and see what we can do to get it landed.

jdm (approved):
Oof, good find on the difference between unit tests and normal builds.

python/servo/command_base.py
```diff
@@ -585,6 +589,11 @@ def build_env(self, is_build=False, test_unit=False):
         if os.path.isdir(tools):
             env["PATH"] = "%s%s%s" % (tools, os.pathsep, env["PATH"])
 
+        if "msvc" in servo.platform.host_triple():
+            # on MSVC, we need some DLLs in the path. They were copied
+            # in to the servo.exe build dir, so just point PATH to that.
+            util.prepend_paths_to_env(env, "PATH", path.dirname(self.get_binary_path(False, False)))
+
```
sagudev, line 596: This one is actually dirty hack for unit-tests, which I would like to remove so unit-tests could run independently of `./mach build`. See https://github.com/servo/servo/issues/29815#issuecomment-1602973927

PR #29949, opened 2023-06-29
Layout 2020: Properly calculate clearance

### LINKED ISSUE #29885: Layout 2020: Element with clearance shouldn't be pushed downwards by collapsed margin from child
```html
<div style="float: left; width: 25px; height: 25px; background: cyan"></div>
<div style="clear: both; outline: solid blue">
  <div style="margin: 25px 0; width: 50px; height: 50px; background: magenta"></div>
</div>
<div style="float: left; width: 25px; height: 25px; background: cyan"></div>
<div style="margin: 25px 0; clear: both; outline: solid blue">
  <div style="margin: 25px 0; width: 50px; height: 50px; background: magenta"></div>
</div>
```

The magenta box has some margin-top that collapses with its parent. In both cases this margin should be overlapping the float, however on the 1st case it pushes the cleared element downwards. Adding explicit margin to it (2nd case) avoids the problem.

### LINKED ISSUE #29919: Layout 2020: incorrect clearance when there are margins collapsing
```html
<!DOCTYPE html>
  <div style="margin: 25px 0; background: yellow">
    <div style="float: left; width: 25px; height: 25px; background: magenta"></div>
    <div style="clear: both; margin-top: 25px; height: 25px; background: cyan"></div>
  </div>
</div>
```

| 2013 (good) | 2020 (bad) |
| - | - |
|  |  |

First we determining the hypothetical position of the element’s top border edge, behaving as `clear: none`:

This hypothetical position of the element’s top border edge is not past the relevant float, so clearance is introduced, preventing margin collapse. So, assuming a clearance of 0px,

There are two options for calculating the clearance. The 1st one is the amount necessary to place the border edge of the block even with the bottom outer edge of the lowest float that is to be cleared. That is, 0px.

The 2nd option would be flooring that 0px of the 1st option by the amount necessary to place the top border edge of the block at its hypothetical position, -25px. But `max(0px, -25px) = 0px`.

So regardless of the option, clearance should be 0px (which is different than having no clearance). But layout 2020 thinks it's 25px.

Loirooriol:
calculate_clearance() was not taking into account that adding clearance prevents top margin from collapsing with earlier margins.

---

- [X] These changes fix #29885 and #29919

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -821,37 +821,73 @@ impl SequentialLayoutState {
         self.current_margin = CollapsedMargin::zero();
     }
 
-    /// Returns the amount of clearance that a block with the given `clear` value at the current
-    /// `bfc_relative_block_position` (with top margin included in `current_margin` if applicable)
-    /// needs to have.
+    /// Returns the amount of clearance (if any) that a block with the given `clear` value
+    /// needs to have at `current_block_position_including_margins()`.
+    /// `block_start_margin` is the top margin of the block, after collapsing (if possible)
+    /// with the margin of its contents. This must not be included in `current_margin`,
+    /// since adding clearance will prevent `current_margin` and `block_start_margin`
+    /// from collapsing together.
     ///
     /// https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#flow-control
-    pub(crate) fn calculate_clearance(&self, clear_side: ClearSide) -> Option<Length> {
+    pub(crate) fn calculate_clearance(
+        &self,
+        clear_side: ClearSide,
+        block_start_margin: &CollapsedMargin,
+    ) -> Option<Length> {
         if clear_side == ClearSide::None {
             return None;
         }
 
-        let hypothetical_block_position = self.current_block_position_including_margins();
+        // Calculate the hypothetical position where the element's top border edge
+        // would have been if the element's `clear` property had been `none`.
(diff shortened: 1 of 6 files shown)


Loirooriol:
@bors-servo try=wpt-2020

The patch is on top of #29945

Loirooriol:
#29904 still not fixed :(

components/layout_2020/flow/float.rs
```diff
@@ -821,37 +821,71 @@ impl SequentialLayoutState {
...
+    /// The top margin must not be included in `current_margin`, pass it as `next_margin` instead.
+    /// The reason is that the presence of clearance must prevent `current_margin` and `next_margin`
+    /// from collapsing.
     ///
     /// https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#flow-control
-    pub(crate) fn calculate_clearance(&self, clear_side: ClearSide) -> Option<Length> {
+    pub(crate) fn calculate_clearance(
+        &self,
+        clear_side: ClearSide,
+        next_margin: &CollapsedMargin,
```
mrobinson, line 834: Naming nit: How about `block_start_margin_of_clearing_element`? I know it's a mouthful, but `next_margin` is a bit ambiguous.

mrobinson (approved):
Looks good to me! Thanks for fixing this.

components/layout_2020/flow/float.rs
Loirooriol: Renamed to `block_start_margin`

PR #29952, opened 2023-07-01
Resolve percentages in find_block_margin_collapsing_with_parent

Loirooriol:
Padding percentages weren't being resolved, this function would think
that margins wouldn't collapse even if the percentage resolved to 0px.
And margin percentages were always resolved to 0px.

So now the function takes a ContainingBlock parameter that is used to
resolve these percentages properly.

---


mrobinson (approved):
Nice.

PR #29954, opened 2023-07-01
Windows bootstrap support

### LINKED ISSUE #25224: Fix mach bootstrap on Windows

mrobinson:
---

- [x] These changes fix #25224
- [x] These changes do not require tests because they are just support script changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,12 @@ manually, try the [manual build setup][manual-build].
 
  - Download and run [`rustup-init.exe`](https://win.rustup.rs/) then follow the onscreen instructions.
  - Install [chocolatey](https://chocolatey.org/)
- - Run `choco install support\windows\chocolatey.config`
-   *This will install CMake, Git, LLVM, Ninja, NuGet, Python and the Visual Studio 2019 Build Tools.*
- - Run `mach bootstrap-gstreamer`
+ - Run `mach bootstrap`
+  - *This will install CMake, Git, Ninja, NuGet, Python and the Visual Studio 2019 Build Tools
+     via choco in an Administrator console. It can take quite a while.*
+  - *If you already have Visual Studio 2019 installed, this may not install all necessary components.
+     Please follow the Visual Studio 2019 installation instructions in the [manual setup][manual-build].*
+- Run `refreshenv`
 
 See also [Windows Troubleshooting Tips][windows-tips].
 
--- a/python/mach_bootstrap.py
+++ b/python/mach_bootstrap.py
@@ -241,13 +241,8 @@ def bootstrap_command_only(topdir):
     import servo.platform
     import servo.util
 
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -43,9 +43,12 @@ manually, try the [manual build setup][manual-build].
 
  - Download and run [`rustup-init.exe`](https://win.rustup.rs/) then follow the onscreen instructions.
  - Install [chocolatey](https://chocolatey.org/)
- - Run `choco install support\windows\chocolatey.config`
-   *This will install CMake, Git, LLVM, Ninja, NuGet, Python and the Visual Studio 2019 Build Tools.*
- - Run `mach bootstrap-gstreamer`
+ - Run `mach bootstrap`
+  - *This will install CMake, Git, LLVM, Ninja, NuGet, Python and the Visual Studio 2019 Build Tools
```
mukilan, line 47: Will LLVM be installed via choco as this comment suggests? I see we have removed it in choco.config

mukilan (approved):
Nice!

README.md
mrobinson: Oh, nice catch. I've fixed this now.

mrobinson:
@bors-servo retry
 - Looks like #23168 or #27315

PR #29955, opened 2023-07-01
Set the Windows subsystem using a rust directive

mrobinson:
This is one less thing that mach has to do now.

---


jdm (approved):
Huh, neat!

PR #29956, opened 2023-07-01
Update mozangle

mrobinson:
This should allow servo to take advantage of faster compilation times on Windows.

---


mrobinson:
Hrm. It seems this cut maybe 10 to 20 minutes off the build, so there's more to do here to get the Windows build times down.

PR #29965, opened 2023-07-04
Use the WebRender clip chain API

mrobinson:
The old clipping API has been [removed from WebRender][removed], so this switches
both legacy and new layout over to use the clip chain API in preparation
for the WebRender upgrade.

[removed]: https://hg.mozilla.org/mozilla-central/rev/dc8ed2eca5091a75725ddccb3ea7d3359b9cf6ef

---


delan (approved):
This makes sense, thanks for the patch! I would recommend including a link to [the upstream patch](https://hg.mozilla.org/mozilla-central/rev/dc8ed2eca5091a75725ddccb3ea7d3359b9cf6ef) in the description for future readers.

mrobinson:
Thanks for the review. I've added a link to the upstream change in the description.

PR #29976, opened 2023-07-06
winit: initial minibrowser

### LINKED ISSUE #29930: Servo embedding: Create a mini-browser
Current status

We use winit to create this window

**Mini browser**

Minimal requirement:

^ This window was created using egui

**Option 1: Winit + egui**
Steps:
- Create a window using winit
- Use [egui_winit](https://github.com/emilk/egui/tree/master/crates/egui-winit) to create a url bar and to render the page inside the window we created using winit. We might need to do something similar to this [egui_winit example](https://github.com/hasenbanck/egui_example)

note: I think it might not make sense to go with this option, as why would I use winit to create a window when I can create that same window using egui.

**Option 2: Egui**
Steps:
- Create window using [egui](https://github.com/emilk/egui)
- Create url bar using egui/eframe 
- Integrate servo *engine* into egui to render pages

Question: Can we achieve option 2 without changing embedding api too much?

I am still exploring the options to execute this task. Opinions and suggestions are welcomed :)

delan:
This patch introduces a new minibrowser mode, enabled by running the winit port with --minibrowser.

The minibrowser has a location bar and a go button. The location bar updates to reflect the URL of the top-level browsing context, unless the user has started editing the location without clicking go.

It has been tested on Linux with X11, but should also work with Wayland and on Windows and macOS.

image

Some changes to the compositor were necessary to allow Servo and our minibrowser code to cooperatively draw to the same OpenGL framebuffer. Normally the compositor “presents” each new frame (swapping buffers) as soon as WebRender has drawn to it, but we need to make sure that both WebRender and the minibrowser are [repainted on every frame](https://gamedev.stackexchange.com/q/7320). Failing to do so leads to strange glitches where new frames get overwritten with stale framebuffer contents.

We add an external­_present mode that instead notifies the embedder (via the constellation) that a new frame is ready, allowing the embedder to draw other things to the framebuffer. In external­_present mode, the compositor does not run again until the embedder calls present, to ensure consistency of WebRender output, and IO­Compositor­::clear­_background no longer clears the whole framebuffer, only the viewport area.

Some optimisations were also necessary to make this performant:

* we add a Browser field that tracks the current top-level URL as a string, so we can avoid repeated serialisation
* we vendor and modify EguiGlow to retain its shapes, so we can repaint the minibrowser without an egui update
* when egui needs an update, we defer it through a winit redraw request, so we can coalesce updates over events

---

- [x] These changes fix #29930
- [ ] These changes partially involve #30049

- [x] These changes do not require tests because it would be impractical to test the minibrowser directly, but we may be able to unify the codepaths so the minibrowser is indirectly tested with WPT


ports/winit/app.rs
```diff
@@ -62,6 +71,83 @@ impl App {
...
+        let gl = unsafe {
+            glow::Context::from_loader_function(|s| {
+                webrender_surfman.get_proc_address(s)
+            })
+        };
+
+        // glow needs to set framebuffer as a target
+        unsafe {
+            gl.bind_framebuffer(glow::FRAMEBUFFER, framebuffer_object);
+        };
```
delan: when i run this now, the fbo number is zero, which apparently tells [glBindFramebuffer](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBindFramebuffer.xhtml) to break any existing binding. i wonder if we really need this?

delan: removed in e9f8c3c0bbbdc6166f9006dc69d88a11103e2b2d

ports/winit/app.rs
```diff
@@ -95,9 +97,9 @@ impl App {
                         );
                     });
 
-                    // Get the toolbar size
-                    // TODO somehow pass this to the headed Window::get_coordinates
-                    dbg!(ctx.used_rect());
+                    unsafe {
+                        TOOLBAR_HEIGHT = ctx.used_rect().height();
+                    }
```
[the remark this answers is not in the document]
atbrakhi: removed in [a8e1e5b](https://github.com/servo/servo/pull/29976/commits/a8e1e5be3ed5810dc0cbc0215d1d96b3e92ab612). thanks

ports/winit/app.rs
```diff
@@ -114,7 +202,24 @@ impl App {
             }
 
             // Handle the event
-            app.queue_embedder_events_for_winit_event(e);
+            let response = match e {
+                winit::event::Event::WindowEvent { ref event, .. } => {
+                    if let Some(mut minibrowser) = app.minibrowser() {
+                        minibrowser.context.on_event(&event)
```
wusyong, line 208: There are a few event surges that will return `repaint: true` and make the response of the browser pretty laggy.
These include `WindowEvent::CursorMoved` `WindowEvent::MouseWheel` and `WindowEvent::AxisMotion`.
Not sure what's the proper way to avoid, but I think one workaround could be detecting the height since this is the only egui part it has right now.

delan: thanks! i noticed that too, but i wasn’t sure how to fix it yet ^^;;

delan: fixed in 62ec187c2d8f89957f64c10a2cd1320ef47ef145 by using winit redraw requests to coalesce minibrowser updates.

ffc971de935aa90979213db8adc8472251e82ecc also tried filtering out CursorMoved events below the toolbar, but this made it impossible to unfocus the location bar by clicking the page contents, so i abandoned that.

ports/winit/Cargo.toml
```diff
@@ -30,12 +30,13 @@ ProductName = "Servo"
 
 [features]
 debugmozjs = ["libservo/debugmozjs"]
-default = ["webdriver", "max_log_level"]
+default = ["webdriver", "max_log_level", "minibrowser"]
```
mrobinson, line 33: I wonder if we could do away with this feature entirely. I think that the winit port doesn't use conditional compilation, so removing default features would break the build.

ports/winit/app.rs
```diff
@@ -29,6 +34,62 @@ pub struct App {
     event_queue: RefCell<Vec<EmbedderEvent>>,
     suspended: Cell<bool>,
     windows: HashMap<WindowId, Rc<dyn WindowPortsMethods>>,
+    minibrowser: Option<RefCell<Minibrowser>>,
+}
+
+struct Minibrowser {
```
mrobinson, line 40: Perhaps this could move to a separate file just for cleanliness? Maybe `minibrowser.rs`.

ports/winit/app.rs
```diff
@@ -60,51 +121,107 @@ impl App {
...
+            // Set up egui context for minibrowser ui
+            // Adapted from https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui_glow/examples/pure_glow.rs
+            app.minibrowser = Some(Minibrowser {
+                context: egui_glue::EguiGlow::new(events_loop.as_winit(), Arc::new(gl), None),
+                event_queue: RefCell::new(vec![]),
+                toolbar_height: 0f32.into(),
+                location: RefCell::new(ServoUrl::into_string(get_default_url())),
+                location_dirty: false.into(),
+            }.into());
+        }
```
mrobinson, line 153: This method is pretty long. Perhaps this could move into a helper?

ports/winit/app.rs
```diff
@@ -60,51 +121,107 @@ impl App {
...
+                    gl::GlesFns::load_with(|s| webrender_surfman.get_proc_address(s))
+                },
+            };
+            let gl = unsafe {
+                glow::Context::from_loader_function(|s| {
+                    webrender_surfman.get_proc_address(s)
+                })
+            };
+            webrender_surfman.make_gl_context_current().unwrap();
+            debug_assert_eq!(webrender_gl.get_error(), gleam::gl::NO_ERROR,);
```
mrobinson, line 142: Looks like a trailing comma snuck in after `gleam::gl::NO_ERROR`.

ports/winit/browser.rs
```diff
@@ -31,6 +31,8 @@ use tinyfiledialogs::{self, MessageBoxIcon, OkCancel, YesNo};
 
 pub struct Browser<Window: WindowPortsMethods + ?Sized> {
     current_url: Option<ServoUrl>,
+    current_url_taken: bool,
```
mrobinson, line 34: Hrm. I think maybe it would be a little simpler to just store a stringified version of the URL here when it changes. Then you could only do the conversion when the URL actually change and pass and expose a reference to a getter.

ports/winit/app.rs
```diff
@@ -183,14 +337,45 @@ impl App {
...
+
+    /// Updates the location field when the [Browser] says it has changed, unless the user has
+    /// started editing it without clicking Go.
+    fn update_location_in_toolbar(&self, minibrowser: &mut RefMut<Minibrowser>) -> bool {
+        if !minibrowser.location_dirty.get() {
+            if let Some(current_url) = self.browser.borrow_mut().take_top_level_url_change() {
+                minibrowser.location = RefCell::new(ServoUrl::into_string(current_url));
+                return true;
+            }
+        }
```
mrobinson, line 366: If you take my suggestion below, I think this could be a function on the Minibrowser struct itself.

ports/winit/app.rs
```diff
@@ -183,14 +337,45 @@ impl App {
...
+                    let location = minibrowser.location.borrow();
+                    let Ok(url) = ServoUrl::parse(&location) else {
+                        warn!("failed to parse location");
+                        break;
+                    };
+                    self.event_queue.borrow_mut().push(EmbedderEvent::LoadUrl(browser_id, url));
+                },
+            }
+        }
+    }
```
mrobinson, line 356: Could this be a function on the Minibrowser struct that takes a mutable reference to the event queue? That would further isolate the minibrowser code in its own code unit.

ports/winit/browser.rs
```diff
@@ -437,6 +456,7 @@ where
                 },
                 EmbedderMsg::HistoryChanged(urls, current) => {
                     self.current_url = Some(urls[current].clone());
+                    self.current_url_taken = false;
```
mrobinson, line 459: Maybe just store the stringified version here.

mrobinson (changes requested):
Thanks for walking through this with us. I'm excited about this functionality! I just have a few suggestions.

ports/winit/Cargo.toml
delan: done in 00d61872126774efdf70f91a26074ae1332c1a2e

delan: done in 00d61872126774efdf70f91a26074ae1332c1a2e

ports/winit/app.rs
delan: done in 06e0d2987e42a1fd542a34697be3f7ca37ecdc62

ports/winit/app.rs
delan: fixed in ebdb00783024faa76abea961aced11a746b95ac4

delan: fixed in ebdb00783024faa76abea961aced11a746b95ac4

ports/winit/app.rs
delan: done in 06e0d2987e42a1fd542a34697be3f7ca37ecdc62

ports/winit/app.rs
delan: done in 06e0d2987e42a1fd542a34697be3f7ca37ecdc62

ports/winit/browser.rs
delan: great idea, that makes a lot more sense! done in b01b10c591801be4736ccd501cfa292c9a361bc7

ports/winit/browser.rs
delan: done in b01b10c591801be4736ccd501cfa292c9a361bc7

mrobinson (approved):
Thanks for these updates. This looks great!

PR #29977, opened 2023-07-06
Place replaced and non-auto inline size independent FCs next to floats

mrobinson:
The CSS2 specification says that replaced content and independent
formatting contexts should be placed next to floats. This change adds
support for that, but punts on support for independent formatting
contexts that have an auto inline size. Support for auto inline sizes
requires a much more complex layout algorithm.

---

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -17,6 +17,7 @@ use crate::style_ext::{ComputedValuesExt, DisplayInside};
 use crate::ContainingBlock;
 use euclid::num::Zero;
 use servo_arc::Arc;
+use std::collections::VecDeque;
 use std::fmt::{Debug, Formatter, Result as FmtResult};
 use std::ops::Range;
 use std::{f32, mem};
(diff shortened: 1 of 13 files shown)


mrobinson:
@Loirooriol I've gone ahead and posted this for review. There are a few changes from our work together:

 - I've split the function for placing independent formatting contexts into two (one sequential and the other which calls into it). I've also moved these to be on the NonReplacedFormattingContext struct as that means passing less arguments.
 - I've added documentation.

mrobinson:
The failure in `/css/CSS2/css1/c414-flt-fit-006.xht` looks pretty bad to be honest. The floats are moving up past the first paragraph.

mrobinson:
I pushed a couple fixes here:
 - Properly clear past independent formatting contexts with auto width.
 - Never place replaced/independent formatting contexts above their placement ceiling.

components/layout_2020/flow/float.rs
```diff
@@ -70,6 +71,112 @@ impl ContainingBlockPositionInfo {
...
+        }
+    }
+
+    fn top_of_placement_for_current_bands(&self) -> Length {
+        self.ceiling.max(self.current_bands.front().unwrap().top)
+    }
+
+    fn current_bands_height(&self) -> Length {
+        assert!(self.current_bands.len() > 0);
+        self.current_bands.back().unwrap().top - self.top_of_placement_for_current_bands()
```
Loirooriol, line 110: If flooring by `self.ceiling` makes a difference, then this height could be negative?

mrobinson: Hrm...yeah, I guess we don't want to find the first band that this object fits in, but the first band that extends beyond the ceiling. This might need a bit of work.

mrobinson: On second thought, I think this is correct. There are two options here:

1. The ceiling is in the middle of the first band in which case the the usable height of our band selection goes from [ceiling...end of bands].
2. The first band starts lower than the ceiling in which case the usable height of our band selection goes from [start of bands...end of bands].

Loirooriol: Mm, yeah I guess that's right, the problem is in `calculate_viable_inline_space`.

components/layout_2020/flow/float.rs
```diff
@@ -70,6 +71,112 @@ impl ContainingBlockPositionInfo {
...
+                .find_next(self.current_bands.back().unwrap().top)
+                .unwrap();
+            self.current_bands.push_back(next_band);
+        }
+    }
+
+    fn calculate_viable_inline_space(&self) -> (Length, Length) {
+        let mut max_inline_start = self.float_context.containing_block_info.inline_start;
+        let mut min_inline_end = self.float_context.containing_block_info.inline_end;
+        for band in self.current_bands.iter() {
```
Loirooriol, line 128: I think this should avoid the last band. The last band is only there to know the height of the bands, but we shouldn't actually consider it.

```html
<!DOCTYPE html>
<div style="display: flow-root; border: solid; width: 200px">
  <div style="float: left; height: 100px; width: 150px; background: cyan"></div>
  <div style="float: left; height: 100px; width: 175px; background: lime"></div>
  <div style="display: flow-root; background: magenta; width: 50px; height: 50px"></div>
</div>
```

Even though the magenta element would fit next to the cyan float, due to iterating here we think that we can only fit 25px, so the magenta element goes below the floats. This doesn't match other browsers.

components/layout_2020/flow/mod.rs
```diff
@@ -768,82 +766,209 @@ fn layout_in_flow_non_replaced_block_level_same_formatting_context(
...
+            pbm,
+            min_box_size,
+            max_box_size,
+            margin,
+        } = solve_containing_block_padding_border_and_margin_for_in_flow_box(
+            containing_block,
+            &self.style,
+        );
+
+        let clearance = sequential_layout_state.calculate_clearance_and_adjoin_margin(
```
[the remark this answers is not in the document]
mrobinson: Okay. I think I've fixed this by calculating a hypothetical `bfc_relative_block_position` for the SequentialLayoutState as if there was no extra adjustment from floats and using that for all calculations. Then at the end, I only adjoin margins if there was no effective clearance.

components/layout_2020/flow/float.rs
mrobinson: Oh, nice catch! I've fixed this and a few more tests are now passing.

components/layout_2020/flow/float.rs
```diff
@@ -70,6 +71,114 @@ impl ContainingBlockPositionInfo {
...
+            self.current_bands.push_back(next_band);
+        }
+    }
+
+    fn calculate_viable_inline_space(&self) -> (Length, Length) {
+        let mut max_inline_start = self.float_context.containing_block_info.inline_start;
+        let mut min_inline_end = self.float_context.containing_block_info.inline_end;
+        assert!(!self.current_bands.is_empty());
+
+        for band in self.current_bands.iter().take(self.current_bands.len() - 1) {
```
Loirooriol, line 130: There is a problem here if we only have one band:

```html
<!DOCTYPE html>
<div style="display: flow-root; border: solid; width: 100px;">
    <div style="display: flow-root; width: 50px; outline: solid red"></div>
</div>
```

I think we should enforce having at least 2 bands (except if the 1st one is already infinite?), or here not ignore the last band if there is a single one.

mrobinson: Oh, good point. I've pushed a new commit which should handle this case.

components/layout_2020/flow/float.rs
```diff
@@ -70,6 +71,123 @@ impl ContainingBlockPositionInfo {
...
+
+            self.current_bands.push_back(next_band);
+        }
+        true
+    }
+
+    fn calculate_viable_inline_space(&self) -> (Length, Length) {
+        let mut max_inline_start = self.float_context.containing_block_info.inline_start;
+        let mut min_inline_end = self.float_context.containing_block_info.inline_end;
+        assert!(self.current_bands.len() > 1);
```
Loirooriol, line 136: This is failing for

```html
<!DOCTYPE html>
<div style="display: flow-root; border: solid; width: 100px;">    
    <div style="display: flow-root; width: 150px; outline: solid red"></div>
</div>
```

Loirooriol:
I think it's easier to reason about separating `next_band` from `current_bands`, WDYT?

mrobinson:
This looks good to me, if it looks good to you!

Loirooriol (approved):
Maybe the clearance logic would look better with some refactoring, but it can be done in a follow-up

PR #29950, opened 2023-06-30
Make Layout 2020/2013 a runtime switch

mrobinson:
This change  makes the choice between Layout 2020 and Layout
2013 a runtime switch. The main part of the patch is removing all layout
related compile-time flags from the style crate. This makes the style crate a
lot simpler, but now Layout 2020 parses a good deal of features that it doesn't
support (table, vertical-align, etc). The good news is that we plan to add
these features to Layout 2020. 

- Compile a single style for both Layout 2020 and Layout 2013
- Remove layout_2020/layout_2013 features
- Update GitHub workflows for single compilation
- Update test results

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #29942
- [x] There are tests for these changes

PR #29980, opened 2023-07-06
Split the WPT workflow from the Linux one

mrobinson:
This allows running tests for only one layout system at a time. This was
originally part of #29950, but was lost in the shuffle.

---

- [x] These changes do not require tests because they improve the CI.


mrobinson:
@bors-servo r+

This change is originally from @sagudev, but was lost somehow when landing #29950.

PR #29988, opened 2023-07-10
Add .envrc to gitignore

schrottkatze:
Added a `.envrc` file for automatically loading the nix shell via direnv for developer convenience. 

---

- [x] These changes do not require tests because they update the .gitnore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ layout_trace*
 
 # Package managers
 etc/homebrew/Brewfile.lock.json
+
+# direnv
+.envrc


mrobinson:
My doubt here is that this envrc is nix specific but my understanding is that envrc itself is not. Also is it possible that people working on Servo would want to create their own envrc?

obsidianical:
I've seen direnv so far only on nix projects due to that integration with the simple `use nix`. This is probably the simplest possible to get a good dev experience on nix because when you `cd` into the servo repo it auto loads the nix shell, and unless someone made their own nix files somewhere else (which I almost did because my config doesn't have python outside of environments where it's needed) I don't think that'd be needed. 

Maybe servo could switch to nix flakes some day and that'd require a change to the envrc again, or as you said that direnv itself isn't nix specific and it's probably possible to just use nix if nix is present and otherwise do whatever other setup steps are needed, but as I said: I didn't see much direnv adoption outside of nix circles.

mrobinson:
Let's simply add .env to .gitignore for now and add instructions for people using nix.

obsidianical:
Then I'll do that in a few hours. Should I only mention nix and write the shell command or also the command to set up direnv (and put the respective file into gitignore then)?

mrobinson:
For the documentation, feel free to include anything that you think could be useful to nix users. Nix documentation is currently here: https://github.com/servo/servo/wiki/Building#nixos Just let me know what you'd like to write there and I am happy to add it. Thank you!

obsidianical:
I'd just suggest adding a subsection for NixOS at the build setup section in the readme that shows the simple `nix-shell etc/shell.nix`, because it explains stuff for other distros individually and then you have to go to the wiki for nix, which i think is quite intuitive and not straightforward

mrobinson:
> I'd just suggest adding a subsection for NixOS at the build setup section in the readme that shows the simple nix-shell etc/shell.nix, because it explains stuff for other distros individually and then you have to go to the wiki for nix, which i think is quite intuitive and not straightforward

Special instructions are printed for nix when when running mach bootstrap:

```python
            print('NixOS does not need bootstrap, it will automatically enter a nix-shell')
            print('Just run ./mach build')
            print('')
            print('You will need to run a nix-shell if you are trying '
                  'to run any of the built binaries')
            print('To enter the nix-shell manually use:')
            print('  $ nix-shell etc/shell.nix')
```

obsidianical:
yes, if you have python installed. I don't have python installed because if I need python I set up a shell with it, but the error if you don't have it installed is pretty confusing:

```
./mach: line 9: exec: python: cannot execute: Is a directory
```

mrobinson:
What is the full command that nix users need to run in order to run mach bootstrap?

mrobinson:
I've gone ahead an updated this to simply add .envrc to .gitignore. I think this is a good first step. It's not clear to me that .envrc is only used by folks using nix.

mrobinson (approved):

sagudev:
@mrobinson I believe `CRASH [expected OK] /fetch/local-network-access/nested-worker.https.window.html` is another Intermittent as I got it several times too.

mrobinson:
> @mrobinson I believe `CRASH [expected OK] /fetch/local-network-access/nested-worker.https.window.html` is another Intermittent as I got it several times too.

I opened https://github.com/servo/servo/issues/30005, so hopefully intermittent filtering will catch this test.

mukilan:
@mrobinson is the title of the issue correct? is it supposed to be **nested-worker**'.https.window.html as mentioned in the comment?

PR #29997, opened 2023-07-13
De-indent code in find_block_margin_collapsing_with_parent

Loirooriol:


mrobinson (approved):

mrobinson:
> Why is it now required to sign commits?

Theoretically it has been required since the move to LF, as described in https://github.com/servo/servo/blob/master/CONTRIBUTING.md. I've just made it so that the merge queue enforces this now. We might need to turn this of for updates to the WPT tests though.

PR #29999, opened 2023-07-16
Fix WPT sync and simplify the update scripts

mrobinson:
It seems there are more changes necessary to fix WPT imports.

Much of the code used to import WPT tests from upstream has been moved to the WPT repository itself, so rework the WPT import code to simply reuse the upstream version and delete our downstream version. In addition, simplify the workflows by merging the entire process into mach and also directly into the GitHub workflow. This should finally fix WPT imports after combining compilation of layout and legacy layout.

---

- [x] `./mach test-tidy` does not report any errorrs
- [x] These changes do not require tests because they are CI changes.


mrobinson:
The import triggered from this change is !29998, but that should wait to land until this one does since it includes this change as the first commit.

mukilan (approved):

mrobinson:
Thank you for the review!

PR #30000, opened 2023-07-16
Stop using webrender_api::get_scroll_node_state

mrobinson:
This API has been removed in the latest version of WebRender [^1] and we can
simply get this information from the compositor-side scroll tree. In
addition, this change limits the amount of data sent to the pipeline
that actually changed and gives the function in the compositor a better
name.

[^1]: https://hg.mozilla.org/mozilla-central/rev/0addf0795838c43cdd53a7534ecb96dfd7070d56

---


mukilan (approved):
Thanks for walking through the changes!

PR #30001, opened 2023-07-17
Mark skipped CI workflows as successful

mrobinson:
When the decision job decides that a workflow should be skipped because
an identical one has already run, that workflow should be marked as
successful and not unsuccessful.

---

- [x] These changes do not require tests because they fix CI behavior.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,39 +17,61 @@ jobs:
   decision:
     name: Decision
     runs-on: ubuntu-20.04
+    outputs:
+      skipped: ${{ steps.skipDecision.outputs.result }}
+      platforms: ${{ steps.platformDecision.outputs.result }}
     steps:
-      # If an identical build exists that suceeded, skip this workflow run. We don't do
-      # this check for pull_request and merge_group events, out of caution.
-      - name: Skip previous identical builds
-        if: ${{ github.event_name != 'pull_request' && github.event_name != 'merge_group' }}
+      - name: Skip Decision
+        id: skipDecision
         uses: actions/github-script@v6
         with:
+          result-encoding: string
           script: |
+            // Never skip workflow runs for pull requests or merge groups, which might
+            // need to actually run / retry WPT tests.
+            if (context.eventName == "pull_request" || context.eventName == "merge_group") {
+              return "run";
+            }
+            // Skip the run if an identical run already exists. This helps to avoid running
+            // the workflow over and over again for the same commit hash.
             if ((await github.rest.actions.listWorkflowRuns({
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -17,53 +17,76 @@ jobs:
...
-        if: ${{ github.event_name != 'pull_request' && github.event_name != 'merge_group' }}
+      - name: Skip Decision
+        id: skipDecision
         uses: actions/github-script@v6
         with:
+          result-encoding: string
           script: |
+            // Never skip workflow runs for pull requests or merge groups, which might
+            // need to actually run / retry WPT tests.
+            if (context.eventName == "pull_request" || github.event_name == "merge_group") {
```
mukilan, line 32: Can `context.eventName` be replaced with `github.event_name`  as well (similar to merge_group clause)?

.github/workflows/main.yml
```diff
@@ -17,53 +17,76 @@ jobs:
...
+      - name: Platform Decision
+        id: platformDecision
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
+          script: |
+            if ("${{ steps.skipDecision.outputs.result }}" == "skip") {
+              return "none";
+            }
+            if (context.eventName == "push" || context.eventName == "merge_group") {
```
mukilan, line 57: I'm not able to find documentation for context.eventName? Is it an alias for 'github.event_name'?

.github/workflows/main.yml
mrobinson: No, I don't think we can. The name is a bit different for scripts inside `github-script` actions.

.github/workflows/main.yml
mrobinson: This comes from the [github-script documentation(https://github.com/actions/github-script#actionsgithub-script). I think it has a different way of working than GitHub Action code that's outside github-script actions.

.github/workflows/main.yml
mukilan: Hmm....isn't the other part using `github.event_name`  of this OR clause also inside github-script action? Or do you mean value of 'pull_request' is not available outside of github-script?

.github/workflows/main.yml
mukilan: Thanks for the link!

mukilan (approved):

.github/workflows/main.yml
mrobinson: Oh, you're totally right. This is a typo that should definitely be fixed. :man_facepalming:

mukilan (approved):

PR #30010, opened 2023-07-19
Totally ignore abspos children for intrinsic sizing

### LINKED ISSUE #30009: calculate_inline_content_size_for_block_level_boxes should ignore abspos
```html
<!DOCTYPE html>
<div style="border:solid; float: left">
    <div style="float: left; width: 100px; height: 100px; background: cyan"></div>
    <div style="position: absolute; width: 100px; height: 100px"></div>
    <div style="float: left; width: 100px; height: 100px; background: magenta"></div>
</div>
```

`calculate_inline_content_size_for_block_level_boxes` treats the abspos as a 0x0 inflow, so it thinks that the max-content size is 100px. It should be 200px.

Layout 2020 (bad):

Layout 2013 (good):

Loirooriol:
calculate_inline_content_size_for_block_level_boxes was relying on inline_content_sizes to get the size of each block-level box child. For absolutely positioned boxes, this was 0x0.

That was no-op before #29887, but then it prevented adding the sizes of a sequence of floats. Abspos should just be ignored instead of treated as 0x0.

This patch removes inline_content_sizes, moves the logic into calculate_inline_content_size_for_block_level_boxes (the only caller), and handles abspos correctly.

---

- [X] These changes fix #30009

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -223,16 +223,30 @@ fn calculate_inline_content_size_for_block_level_boxes(
     writing_mode: WritingMode,
 ) -> ContentSizes {
     let get_box_info = |box_: &ArcRefCell<BlockLevelBox>| {
-        let size = box_
-            .borrow_mut()
-            .inline_content_sizes(layout_context, writing_mode);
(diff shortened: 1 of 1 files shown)


Loirooriol:
Not sure if I should add a test, the spec says "CSS 2 does not define the exact algorithm".

mrobinson (approved):

PR #30012, opened 2023-07-19
Fix intrinsic sizing of block container containing a BFC root after floats

### LINKED ISSUE #30011: calculate_inline_content_size_for_block_level_boxes should handle independent formatting contexts after floats
```html
<!DOCTYPE html>
<div style="border:solid; float: left">
    <div style="float: left; width: 100px; height: 100px; background: cyan"></div>
    <div style="display: flow-root; width: 100px; height: 100px; background: magenta"></div>
</div>
```

A block that establishes an independent formatting context is placed next to previous floats, so we should add their sizes when computing the intrinsic size of the parent block container.

**Firefox (good)**

**Layout 2013 (bad)**

**Layout 2020 (bad)**

Loirooriol:
A block that establishes an independent formatting context is placed
next to previous floats, so we should add their sizes when computing
the intrinsic contribution of the parent block container.

---

- [X] These changes fix #30011
- [X] These changes do not require tests because the spec says "CSS 2 does not define the exact algorithm".

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -260,14 +260,14 @@ fn calculate_inline_content_size_for_block_level_boxes(
                 let size = sizing::outer_inline(&style, writing_mode, || {
                     contents.inline_content_sizes(layout_context, style.writing_mode)
                 });
-                // The element may in fact have clearance, but the logic below ignores it,
-                // so don't bother retrieving it from the style.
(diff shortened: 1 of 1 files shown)


Loirooriol:
I had a full testcase at home that I can check when I get back, but I think this matches what other browser do.

mrobinson (approved):

PR #30014, opened 2023-07-19
Make "@bors-servo try" a GitHub Action

mrobinson:
This is the last piece of the puzzle to turning off bors. This makes
functionality provided by bors to understand "@bors-servo try" a GitHub
Action. For now the syntax is more or less the same, but we can modify
it in the future and even add support for custom configuration options
(more specific build combinations or even passing compiler flags).

The big difference between this and what bors does is that there is no
merge commit. GitHub simply runs tests on the version of the branch that
is on a pull request. There is always the risk that tests might start
failing when a branch is rebased, but this offers a bit more control
because you can easily rebase from the PR and the merge queue will check
this as well.

---

- [x] These changes do not require tests because they are just a CI update.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,80 +11,106 @@ on:
     branches: ["**"]
   merge_group:
     types: [checks_requested]
+  workflow_call:
+    inputs:
+      platform:
+        required: true
+        type: string
+      layout:
+        required: true
+        type: string
+      unit-tests:
+        required: true
+        type: boolean
   workflow_dispatch:
+    inputs:
+      platform:
+        required: false
+        type: choice
+        options: ["none", "linux", "windows", "macos", "all", "sync"]
+      layout:
+        required: false
+        type: choice
+        options: ["none", "2013", "2020", "all"]
+      unit-tests:
+        required: false
+        type: boolean
 
 jobs:
   decision:
     name: Decision
     runs-on: ubuntu-20.04
     outputs:
-      skipped: ${{ steps.skipDecision.outputs.result }}
-      platforms: ${{ steps.platformDecision.outputs.result }}
+      configuration: ${{ steps.configuration.outputs.result }}
     steps:
-      - name: Skip Decision
-        id: skipDecision
+      - name: Configuration
+        id: configuration
         uses: actions/github-script@v6
         with:
-          result-encoding: string
           script: |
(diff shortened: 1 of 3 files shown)


sagudev:
I like the possibility of extensibility, testing only subset of wpt would be useful for webgpu especially.

There are few potential problems I see:
- GitHub API rate limits, idk if `GITHUB_TOKEN` has same rate limits as PAT tokens. github-script offers [retries](https://github.com/actions/github-script#retries), which would be nice for more guaranteed work.
- `mach try` stops working, maybe keep `try-*` branches for now until we migrate it to workflow dispatch or something.
- Historically there was more people with try permission than review/write permission. Idk if this is still relevant, probably not.
- Clogging GitHub actions runners queue, I think that's the main reason such workflows are rarely used in the wild, and for servo this is even more problematic due to WPT runs. Potential solutions could be simple stateless bot, that could be hosted in Cloudflare Workers for **free** running off Webhooks.

mrobinson:
Thanks for taking a look at this:

>     * GitHub API rate limits, idk if `GITHUB_TOKEN` has same rate limits as PAT tokens. github-script offers [retries](https://github.com/actions/github-script#retries), which would be nice for more guaranteed work.

Is this due to the workflow running on every comment? I think 1000 requests an hour should be enough for us, but I would really like to move away from triggering these try runs on comments. That should fix the problem for good.

>     * `mach try` stops working, maybe keep `try-*` branches for now until we migrate it to workflow dispatch or something.

This is a really excellent point. I've restored these workflows. Eventually processing for these should move into the try.yml file, but for now it's fine where they are, I think.

>     * Historically there was more people with try permission than review/write permission. Idk if this is still relevant, probably not.

We actually have total flexibility here! I'm using write permissions as the proxy for try access now, but we could make this based on any group of users in the GitHub project user list.

>     * Clogging GitHub actions runners queue, I think that's the main reason such workflows are rarely used in the wild, and for servo this is even more problematic due to WPT runs. Potential solutions could be simple stateless bot, that could be hosted in Cloudflare Workers for **free** running off Webhooks.

Yeah, this is another reason I'd like to switch away from comment processing as the way to trigger these runs. Perhaps everything could eventually work via `./mach try`. The first step is to move it to GitHub Actions though.

mukilan (approved):

PR #30023, opened 2023-07-25
Fix syntax error in try.yaml

mukilan:
Use the context object and template strings instead of GitHub Workflow string substitutions which can introduce linebreaks.

---

### DIFF
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -22,8 +22,8 @@ jobs:
                   body
                 })
             }
-
-            let tokens = "${{ github.event.comment.body }}".split(" ");
+            let comment = context.payload.comment; 
+            let tokens = comment.body.split(" ");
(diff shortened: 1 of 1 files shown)


.github/workflows/try.yml
```diff
@@ -22,8 +22,8 @@ jobs:
                   body
                 })
             }
-
-            let tokens = "${{ github.event.comment.body }}".split(" ");
+            let comment = context.payload.comment; 
+            let tokens = comment.body.split(" ");
```
mrobinson, line 26: We should split on `/+s/` instead of `" "` here as well in order to properly handle newlines in the comment.

mrobinson (approved):
Looks good to me with one change before landing.

mukilan:
I've approved the newer PR.  I'll close this one in favor of that.

PR #30025, opened 2023-07-25
RFC: Switch to rustls and webpki-roots

### LINKED ISSUE #7888: Remove the dependency on openssl
Seems like we only use it for SSL through hyper. I feel like it should be possible to do better here. Maybe @briansmith has ideas.

### LINKED ISSUE #13749: Switch to mozilla-ca-certs crate instead of manually extracting certs from mozilla-central
This was created by @jld and looks like it might be a nice improvement if we can integrate it with openssl: https://crates.io/crates/mozilla-ca-certs

### LINKED ISSUE #18730: Segfault in openssl
Hello.
I managed to get a segfault in servo in its use of OpenSSL.
I made a [small change](https://github.com/antoyo/servoshell/commit/569e3cdd64b5985b85c5baf0e21d9f30e4d36e8b) to my [gtk binding](https://github.com/antoyo/servoshell) and it now causes a segfault.
Here's the backtrace from gdb:

```
#0  0x00007ffff5cf436f in ERR_put_error () at /usr/lib/libcrypto.so.1.1
#1  0x00007ffff5d772f1 in X509_STORE_add_cert () at /usr/lib/libcrypto.so.1.1
#2  0x00007ffff5d7330d in X509_load_cert_crl_file () at /usr/lib/libcrypto.so.1.1
#3  0x00007ffff5d73473 in  () at /usr/lib/libcrypto.so.1.1
#4  0x00007ffff5d761a2 in X509_STORE_load_locations () at /usr/lib/libcrypto.so.1.1
#5  0x0000555555cc66a5 in openssl::ssl::SslContextBuilder::set_ca_file<&std::path::PathBuf> (self=<optimized out>, file=<optimized out>)
    at ~/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.12/src/ssl/mod.rs:735
#6  0x0000555555cc66a5 in net::connector::create_ssl_client (ca_file=0x7fffe23fc398)
    at ~/.cargo/git/checkouts/servo-12bfca331f128de3/bb998db/components/net/connector.rs:57
#7  0x0000555555cee199 in net::resource_thread::create_http_states (config_dir=...)
    at ~/.cargo/git/checkouts/servo-12bfca331f128de3/bb998db/components/net/resource_thread.rs:117
#8  0x0000555555c6e327 in net::resource_thread::ResourceChannelManager::start (public_receiver=..., private_receiver=..., self=<optimized out>)
    at ~/.cargo/git/checkouts/servo-12bfca331f128de3/bb998db/components/net/resource_thread.rs:129
#9  0x0000555555c6e327 in net::resource_thread::new_core_resource_thread::{{closure}} ()
    at ~/.cargo/git/checkouts/servo-12bfca331f128de3/bb998db/components/net/resource_thread.rs:80
#10 0x0000555555c6e327 in std::sys_common::backtrace::__rust_begin_short_backtrace<closure,()> (f=...)
    at /checkout/src/libstd/sys_common/backtrace.rs:136
#11 0x0000555555c732a0 in std::thread::{{impl}}::spawn::{{closure}}::{{closure}}<closure,()> ()
    at /checkout/src/libstd/thread/mod.rs:394
#12 0x0000555555c732a0 in std::panic::{{impl}}::call_once<(),closure> (self=..., _args=<optimized out>)
    at /checkout/src/libstd/panic.rs:296
#13 0x0000555555c732a0 in std::panicking::try::do_call<std::panic::AssertUnwindSafe<closure>,()> (data=<optimized out>) at /checkout/src/libstd/panicking.rs:480
#14 0x0000555557a06e4d in panic_unwind::__rust_maybe_catch_panic () at /checkout/src/libpanic_unwind/lib.rs:99
#15 0x0000555555ca64bf in std::panicking::try<(),std::panic::AssertUnwindSafe<closure>> (f=...)
    at /checkout/src/libstd/panicking.rs:459
#16 0x0000555555ca64bf in std::panic::catch_unwind<std::panic::AssertUnwindSafe<closure>,()> (f=...)
    at /checkout/src/libstd/panic.rs:361
#17 0x0000555555ca64bf in std::thread::{{impl}}::spawn::{{closure}}<closure,()> ()
    at /checkout/src/libstd/thread/mod.rs:393
#18 0x0000555555ca64bf in alloc::boxed::{{impl}}::call_box<(),closure> (self=0x7fffe9c70090, args=<optimized out>)
    at /checkout/src/liballoc/boxed.rs:728
#19 0x00005555579feb9c in alloc::boxed::{{impl}}::call_once<(),()> () at /checkout/src/liballoc/boxed.rs:738
#20 0x00005555579feb9c in std::sys_common::thread::start_thread () at /checkout/src/libstd/sys_common/thread.rs:24
#21 0x00005555579feb9c in std::sys::imp::thread::{{impl}}::new::thread_start ()
    at /checkout/src/libstd/sys/unix/thread.rs:90
#22 0x00007ffff3b6408a in start_thread () at /usr/lib/libpthread.so.0
#23 0x00007ffff48751bf in clone () at /usr/lib/libc.so.6
```

If you want to try, the code is in the `segfault` branch and you must compile it with:

```
cargo build --release --features force-gtk
```

(This might be an issue with the `openssl` binding as well.)

Thanks to fix this issue.

### LINKED ISSUE #18796: Rewrite the openssl for android build code in Rust, in the relevant build.rs
When using Servo in a third party software, on Android, we have to copy some shell scripts and makefiles to compile servo as a dependency. See https://github.com/paulrouget/servofocus/blob/565eeea6ba7a7a74aa3fead29c77e3983c31f654/rust/android/build.sh#L34

Ideally, this would be done automatically in the openssl crate `build.rs`.

### LINKED ISSUE #20783: Link OpenSSL statically in Linux nightly builds
Our downloadable nightly builds don't work on all Linux distros because they are dynamically linked to a specific version of OpenSSL that is not available everywhere.  We might be able to fix this by setting the `OPENSSL_STATIC` environment variable at build time on the nightly builders:

https://github.com/sfackler/rust-openssl/blob/9e5dcb03f2e429e9a8bda2ade64b6d75058907b8/openssl-sys/build.rs#L506

### LINKED ISSUE #26835: Revoked SSL cert doesn't trigger cert error page
https://revoked.badssl.com/

### LINKED ISSUE #29291: Servo fails to launch on Linux Mint 21.1 with "error while loading shared libraries: libssl.so.1.1"
Operating system: Linux Mint 21.1 Cinnamon
Steps to reproduce:
1)Download and extract servo-latest.tar.gz.
2)Open the servo directory in the terminal.
3)Run `./servo`.
Expected results: Servo launches
Actual results: Nothing happens, and the terminal outputs
`./servo 
./servo: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

mrobinson:
This change replaces OpenSSL with rustls and also the manually curated CA certs file with webpki-roots (effectively the same thing, but as a crate).

Generally speaking the design of the network stack is the same. Changes:

- Code around certificate overrides needed to be refactored to work with rustls so the various thread-safe list of certificates are refactored into `CertificateErrorOverrideManager`
- hyper-rustls takes care of setting ALPN protocols for HTTP requests, so for WebSockets this is moved to the WebSocket code.
- The default safe set of cypher suites is chosen, which seem to correspond to the "Modern" configuration from Mozilla's documentation. [^1] This can be adjusted later.
- Instead of passing a string of PEM CA certificates around, an enum is used that includes parsed Certificates (or the default which reads them from webpki-roots).
- Code for starting up an SSL server for testing is cleaned up a little, due to the fact that the certificates need to be overriden explicitly now. This is due to the fact that the `webpki` crate is more stringent with self-signed certificates than OpenSSL (CA certificates cannot used as end-entity certificates). [^2]

[^1]: https://wiki.mozilla.org/Security/Server_Side_TLS
[^2]: https://github.com/briansmith/webpki/issues/114

Fixes #7888.
Fixes #13749.
Fixes #26835.
Fixes #29291.
Fixes #28903.
Fixes #18730.
Closes #20783.
Closes #18796.

---

- [x] These changes fix various GitHub issues listed above.


Cargo.toml
```diff
@@ -73,11 +77,15 @@ surfman-chains-api = "0.2"
...
 tokio = "1"
+tokio-rustls = "0.24"
+tungstenite = "0.19"
 unicode-bidi = "0.3.4"
 unicode-script = "0.5"
 url = "2.0"
 uuid = { version = "1.3.4", features = ["v4"] }
 webdriver = "0.48.0"
+webpki-roots = "0.23"
+webpki = "0.22"
```
djc, line 88: Instead of using webpki 0.22 I'd suggest using rustls-webpki 0.101 here to match the version that rustls 0.21.5+ uses.

components/net/resource_thread.rs
```diff
@@ -39,19 +38,38 @@ use net_traits::{ResourceThreads, WebSocketDomAction};
...
 use std::thread;
 use std::time::Duration;
 
+/// Load a file with CA certificate and produce a RootCertStore with the results.
+fn load_root_cert_store_from_file(file_path: String) -> io::Result<RootCertStore> {
+    let mut root_cert_store = RootCertStore::empty();
+
+    let mut pem = BufReader::new(File::open(file_path)?);
+    let certs = rustls_pemfile::certs(&mut pem)?;
+    let trust_anchors = certs.iter().map(|cert| {
```
djc, line 61: I'd recommend the `RootCertStore::add_parsable_certificates()` here, letting rustls handle the certificate parsing for you (maybe you can avoid the manual webpki dependency this way?).

components/net/connector.rs
```diff
@@ -244,14 +154,93 @@ where
...
+                        )
+                    },
+                ));
+                root_cert_store
+            },
+            CACertificates::Override(root_cert_store) => root_cert_store,
+        };
+
+        Self {
+            webpki_verifier: WebPkiVerifier::new(root_cert_store, None),
```
mrobinson, line 187: Add a link to the code showing that this is the default verifier for rustls.

Cargo.toml
```diff
@@ -73,11 +77,15 @@ surfman-chains-api = "0.2"
 thin-slice = "0.1.0"
 time = "0.1.41"
 tokio = "1"
+tokio-rustls = "0.24"
+tungstenite = "0.19"
```
djc, line 81: I suppose this is using older tungstenite to match async-tungstenite, and using older webpki-roots to match async-tungstenite/tungstenite? Have you looked into using tokio-tungstenite instead of async-tungstenite? It seems to keep up better with tungstenite releases at least (though both use an old version of webpki-roots for now).

components/net/connector.rs
mrobinson: I've added a link to the place this is set up in rustls codebase.

Cargo.toml
mrobinson: This direct dependency is now removed.

components/net/resource_thread.rs
mrobinson: Thanks for the tip. I'm using this now and have removed the depedency on `webpki`.

Cargo.toml
mrobinson: This is a good idea, though perhaps we should tackle upgrading tungstenite and switching to async-tunstenite in a followup change as it's a bit unrelated.

djc: Hmm, it looks like tokio-tungstenite also provides a `Connector` API that lets you provide your own `rustls::ClientConfig`. Using that would have the benefit that you no longer depend on the rustls choices/dependency versions that tokio-tungstenite comes with, so might be a good way forward (unfortunately the tokio-tungstenite Cargo features don't expose this option in a public way right now).

components/config/opts.rs
```diff
@@ -116,6 +116,9 @@ pub struct Opts {
     /// Path to SSL certificates.
     pub certificate_path: Option<String>,
 
+    /// Path to SSL certificates.
```
mukilan, line 119: Comment is incorrect.

mukilan (approved):
Great work!

mrobinson:
Thanks for the reviews!

PR #30029, opened 2023-07-26
More tweaks for the try GitHub Action

mrobinson:
- Don't trigger try jobs when comments are deleted or edited.
- Don't report success for cancelled jobs.

---

- [x] These changes do not require tests because they change CI behavior.

### DIFF
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -1,5 +1,8 @@
-on: issue_comment
 name: Try
+
+on:
+  issue_comment:
+    types: [created]
 
 jobs:
   parse-comment:
@@ -87,7 +90,7 @@ jobs:
     if: ${{ always() && fromJson(needs.parse-comment.outputs.configuration).try}}
     steps:
       - name: Success
-        if: ${{ !contains(join(needs.*.result, ','), 'failure') }}
+        if: ${{ !contains(join(needs.*.result, ','), 'failure') && !contains(join(needs.*.result, ','), 'cancelled') }}
         uses: actions/github-script@v6
         with:
           script: |


.github/workflows/try.yml
```diff
@@ -87,7 +90,7 @@ jobs:
     if: ${{ always() && fromJson(needs.parse-comment.outputs.configuration).try}}
     steps:
       - name: Success
-        if: ${{ !contains(join(needs.*.result, ','), 'failure') }}
+        if: ${{ !contains(join(needs.*.result, ','), 'failure') && !contains(join(needs.*.result, ','), 'cancelled') }}
```
mukilan, line 93: [`contains`](https://docs.github.com/en/actions/learn-github-actions/expressions#contains) seems to work with arrays directly. Do we still need the `join` here to convert results to a single string?

Loirooriol: Presumably the array may contain a string that contains `cancelled` but is longer than that?

Loirooriol (approved):
LGTM if the assumption in the comment is correct

.github/workflows/try.yml
mrobinson: Huh. This is a good question. I would like to try calling join on the array directly, but I'm copying the same approach we use on other workflows here. Maybe I can try adjusting all workflows to use join on the array directly in a follow up change so that it's the same everywhere.

mukilan (approved):

.github/workflows/try.yml
Loirooriol: Ah, good point, but as Martin says it may be better to adjust them all in a follow-up

mrobinson: > Presumably the array may contain a string that contains `cancelled` but is longer than that?

I think isn't expected to happen but maybe we can do this cleanup in all workflows in a follow up:  https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

mukilan: Agreed, we can have a separate PR for this cleanup.

Loirooriol:
Not sure if this is the cause, but apparently posting a comment in https://github.com/servo/servo/issues/30037 triggered a try job: https://github.com/servo/servo/actions/runs/5679156680

mrobinson:
> Not sure if this is the cause, but apparently posting a comment in https://github.com/servo/servo/issues/30037 triggered a try job: https://github.com/servo/servo/actions/runs/5679156680

It shouldn't fail, but it's expected to trigger the try GitHub Action. For non-PR comments it should exit early. This is basically how this functionality has always worked, just via a webhook before. This is due to a limitation of GitHub. There's no way to create an Action only for pull request comments.

Loirooriol:
In the past I wasn't getting emails like `Run failed: Try - master` just by posting a comment.

mrobinson:
> In the past I wasn't getting emails like `Run failed: Try - master` just by posting a comment.
> 
> Edit: nevermind, doesn't seem to happen more, probably a hiccup

Oh, odd! I think we were hitting some sort of internal error with GitHub there.

PR #30030, opened 2023-07-26
Embed git hash using vergen instead of mach

mrobinson:
Embed the git hash into the servo binary using vergen instead of using
custom Python code in mach. The benefit here is ones less difference
between a normal cargo run and building via mach in addition to removing
a bunch of code.

---

- [x] These changes do not require tests because they just change the way that the version is generated.

### DIFF
--- a/ports/libsimpleservo/api/Cargo.toml
+++ b/ports/libsimpleservo/api/Cargo.toml
@@ -31,6 +31,7 @@ libloading = "0.5"
 [build-dependencies]
 gl_generator = "0.14"
 serde_json = { workspace = true }
+vergen = { version = "8.0.0", features = [ "git", "gitcl" ]}
 
 [features]
 debugmozjs = ["libservo/debugmozjs"]
--- a/ports/libsimpleservo/api/src/lib.rs
+++ b/ports/libsimpleservo/api/src/lib.rs
@@ -185,12 +185,11 @@ pub struct ServoGlue {
 }
 
 pub fn servo_version() -> String {
-    let cargo_version = env!("CARGO_PKG_VERSION");
-    let git_info = option_env!("GIT_INFO");
-    match git_info {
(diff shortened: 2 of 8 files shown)


ports/libsimpleservo/api/Cargo.toml
```diff
@@ -31,6 +31,7 @@ libloading = "0.5"
 [build-dependencies]
 gl_generator = "0.14"
 serde_json = { workspace = true }
+vergen = "3"
```
mukilan, line 34: Any reason we are not using the latest version on crates.io, which is 8.2?

ports/libsimpleservo/api/src/lib.rs
```diff
@@ -185,12 +185,11 @@ pub struct ServoGlue {
 }
 
 pub fn servo_version() -> String {
-    let cargo_version = env!("CARGO_PKG_VERSION");
-    let git_info = option_env!("GIT_INFO");
-    match git_info {
-        Some(info) => format!("Servo {}{}", cargo_version, info),
-        None => format!("Servo {}", cargo_version),
```
mukilan, line 192: Was this GIT_INFO optional to support building servo from source tarballs? Are we removing support for that with this patch?

ports/libsimpleservo/api/Cargo.toml
mrobinson: Huh. Google Search took me to a very old version. I'm using the latest version now.

ports/libsimpleservo/api/src/lib.rs
mrobinson: This is a good. I'm cannot think of any time we build from source tarballs, but I've updated the code to handle this case explicitly. Now the version is `nogit`.

mukilan (approved):

PR #30031, opened 2023-07-26
Remove old code for out-of-date CA stores

mrobinson:
This code was written to handle both Python 2 (which we no longer
support) and old Windows CI machines that did not have up-to-date CA
stores. I think we can remove this now.

---

- [x] These changes do not require tests because they do not really change behavior.

### DIFF
--- a/python/servo/util.py
+++ b/python/servo/util.py
@@ -25,29 +25,8 @@ import six
 from io import BufferedIOBase, BytesIO
 from socket import error as socket_error
 
-try:
-    from ssl import HAS_SNI
-except ImportError:
-    HAS_SNI = False
-
 SCRIPT_PATH = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


python/servo/util.py
```diff
@@ -25,29 +25,8 @@
...
-    if HAS_SNI_AND_RECENT_PYTHON:
-        return "https://static.rust-lang.org/dist"
-
-    return "https://static-rust-lang-org.s3.amazonaws.com/dist"
-
-
-def get_urlopen_kwargs():
-    # The cafile parameter was added in 2.7.9
-    if HAS_SNI_AND_RECENT_PYTHON:
-        import certifi
```
mukilan, line 48: Should we remove `certifi` from [requirements.txt](https://github.com/servo/servo/blob/master/python/requirements.txt#L31)?

mukilan (approved):

PR #30032, opened 2023-07-26
Run unit tests with both layout 2013 and layout 2020

Loirooriol:
Since #29950, unit tests were only running with the legacy layout, and there was no way to run them for layout 2020.

This patch makes './mach test-unit' run unit tests for both.

Also doing some changes so that the layout 2020 floats.rs tests compile.

---

- [X] These changes do not require tests because this is about running tests

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -287,7 +287,7 @@ impl FloatContext {
     }
 
     /// Places a new float and adds it to the list. Returns the start corner of its margin box.
-    pub(crate) fn add_float(&mut self, new_float: &PlacementInfo) -> Vec2<Length> {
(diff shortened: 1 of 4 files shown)


mrobinson (commented):
Since we can now run both sets of tests with the same binary, why not just remove the option entirely and always run all tests? That's what I was shooting for in the original change.

Loirooriol:
I can do that, but worried that running both tests could be slow

mrobinson (approved):
Thanks!

mrobinson:
> I can do that, but worried that running both tests could be slow

I would say that the unit tests are a lot faster than other test suites. A lot of the time used in running them on CI is actually rebuilding different crates with different configurations for testing. The layout suites are some of the faster ones too.

PR #30034, opened 2023-07-27
Floor child sizes in calculate_inline_content_size_for_block_level_boxes

Loirooriol:
Calculating the max-content size of a block container may add the outer max-content sizes of multiple children. The problem is that the outer size may be negative (due to margins), producing an incorrect result.

In particular, it could happen that the max-content size was 50-25=25, but the min-content size would just take the maximum and be 50, which doesn't make sense.

Therefore, this patch floors the size of all children by 0. This seems to match Blink. Firefox and WebKit don't floor in some cases, but then the result seems suboptimal to me. Note that there is no spec for this, see https://github.com/w3c/csswg-drafts/issues/9120 for details.

---

- [X] These changes do not require tests because there is no spec (see https://github.com/w3c/csswg-drafts/issues/9120)

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -250,7 +250,8 @@ fn calculate_inline_content_size_for_block_level_boxes(
             BlockLevelBox::OutOfFlowFloatBox(ref mut float_box) => {
                 let size = float_box
                     .contents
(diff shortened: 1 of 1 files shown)


mrobinson (approved):
This seems fine to me with the idea of once the behavior is specified, tests come via WPT.

PR #30035, opened 2023-07-27
Remove ClearSide enum

Loirooriol:
Just use Clear instead, they have the same values.

---


Loirooriol:
Not sure if I'm missing the point, and there is some reason for this to exist.

mrobinson (approved):
Nice cleanup!

Loirooriol:
I just realized that `clear: inline-start` and `clear: inline-end` are currently resolved to physical at computed-value time, but they should stay as-is (because they should resolve using the writing mode of the containing block, not the element itself).

So once that is fixed, layout may just want to care about the used value instead of the computed one, and then `ClearSide` could make sense.

But well, I don't think Firefox intends to fix the style anytime soon.

PR #30038, opened 2023-07-27
Fix interaction of margins and clearance for PlacementAmongFloats

### LINKED ISSUE #30037: Moving BFC roots to avoid floats doesn't handle margin correctly
```html
<!DOCTYPE html>
<div style="width: 150px; border: solid">
  <div style="height: 10px; background: blue; margin-bottom: 50px"></div>
  <div style="float: left; width: 100px; height: 100px; background: magenta"></div>
  <div style="margin-top: 50px; display: flow-root; width: 100px; height: 100px; background: cyan"></div>
</div>
```

| Expected | Actual |
| - | - |
|  |  |

Only the border box must avoid the float, the margin can overlap it. https://drafts.csswg.org/css2/#floats

> The **border box** of a table, a block-level replaced element, or an element in the normal flow that establishes a new [block formatting context](https://drafts.csswg.org/css2/#block-formatting) (such as an element with [overflow](https://drafts.csswg.org/css2/#propdef-overflow) other than [visible](https://drafts.csswg.org/css2/#valdef-overflow-visible)) must not overlap the margin box of any floats in the same block formatting context as the element itself.

Loirooriol:
Consumers of PlacementAmongFloats weren't handling margins properly.
They were assuming that they would either get a positive adjustment,
or zero for no-op.

However, just like the regular clearance triggered by 'clear', the
clearance added onto blocks that establish an independent FC can be
zero or negative, and the effect is different than having no clearance.

---

- [X] These changes fix #30037 

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -905,15 +905,24 @@ impl SequentialLayoutState {
         self.current_margin = CollapsedMargin::zero();
     }
 
-    /// Returns the amount of clearance (if any) that a block with the given `clear` value
-    /// needs to have at `current_block_position_including_margins()`.
-    /// `block_start_margin` is the top margin of the block, after collapsing (if possible)
-    /// with the margin of its contents. This must not be included in `current_margin`,
-    /// since adding clearance will prevent `current_margin` and `block_start_margin`
-    /// from collapsing together.
-    ///
-    /// https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#flow-control
-    pub(crate) fn calculate_clearance(
+    /// Computes the position of the block-start border edge of an element
+    /// with the provided `block_start_margin`, assuming no clearance.
+    fn position_without_clearance(&self, block_start_margin: &CollapsedMargin) -> Length {
+        // Adjoin `current_margin` and `block_start_margin` since there is no clearance.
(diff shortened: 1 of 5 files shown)


components/layout_2020/flow/float.rs
```diff
@@ -939,23 +947,60 @@ impl SequentialLayoutState {
...
-        // but other browser don't seem to do that.
-        Some(clear_position - position_with_zero_clearance)
+    /// Returns the amount of clearance (if any) that a block with the given `clear` value
+    /// needs to have at `current_block_position_including_margins()`.
+    /// `block_start_margin` is the top margin of the block, after collapsing (if possible)
+    /// with the margin of its contents. This must not be included in `current_margin`,
+    /// since adding clearance will prevent `current_margin` and `block_start_margin`
+    /// from collapsing together.
+    ///
+    /// https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#flow-control
```
mrobinson, line 963: One thing I learned recently is that a raw link in a rustdoc comment is an error. You can fix this by writing:

`<https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#flow-control>`

instead.

mrobinson (approved):
Nice cleanup! Looks good to me apart from a couple nits.

PR #30039, opened 2023-07-27
Don't skip the "Process Comment" step for non-PR try workflows

mrobinson:
For some reason skipping this step is causing internal GitHub errors,
causing jobs to falsely fail. I think we are hitting a bug in GitHub
Actions here.

---

- [x] These changes do not require tests because they fix an issue with CI.


mrobinson:
It seems like these errors may have just been a GitHub hiccup, so I'm closing this for now.

PR #30041, opened 2023-07-28
Remove Gecko-specific Python bootstrap logic

mrobinson:
Servo is no longer completely vendored into Gecko. Instead parts of
Gecko are vendored into Servo. This change removes Python mach bootstrap
code that was written to accommodate the previous situation. It's no
longer necessary.

---

- [x] These changes do not require tests because this just removes unused code from build scripts.

### DIFF
--- a/python/mach_bootstrap.py
+++ b/python/mach_bootstrap.py
@@ -13,6 +13,12 @@ from distutils.spawn import find_executable
 from subprocess import Popen
 from tempfile import TemporaryFile
 
+SCRIPT_PATH = os.path.abspath(os.path.dirname(__file__))
+TOP_DIR = os.path.abspath(os.path.join(SCRIPT_PATH, ".."))
+WPT_PATH = os.path.join(TOP_DIR, "tests", "wpt")
+WPT_RUNNER_PATH = os.path.join(WPT_PATH, "tests", "tools", "wptrunner")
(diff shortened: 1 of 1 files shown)


python/mach_bootstrap.py
```diff
@@ -184,8 +167,8 @@ def _activate_virtualenv(topdir, is_firefox):
     # and it will check for conflicts.
     requirements_paths = [
         os.path.join("python", "requirements.txt"),
-        wptrunner_path(is_firefox, topdir, "requirements.txt",),
-        wptrunner_path(is_firefox, topdir, "requirements_firefox.txt"),
+        os.path.join(WPT_RUNNER_PATH, "requirements.txt",),
+        os.path.join(WPT_RUNNER_PATH, "requirements_firefox.txt"),
```
mukilan, line 171: Is `requirements_firefox.txt` still needed?

mukilan (approved):

python/mach_bootstrap.py
mrobinson: Great question. I've kicked off a try job, attempting to remove that. It's hard to say which requirements inside of the Servo WPT executor might be accidentally relying on Gecko's list of packages.

mrobinson:
Looks like we can safely remove this. Sending this one to the merge queue. Thanks!

PR #30047, opened 2023-07-31
Remove homebrew packaging

mrobinson:
This hasn't been updated since 2017 and homebrew installation is also
provided via a cask which downloads the latest version from the website
[^1]. I think this code is basically unused.

[^1]: https://github.com/Homebrew/homebrew-cask/blob/9e944ae828ccde27340413ff2ac4de603243f5e0/Casks/servo.rb

---

- [x] These changes do not require tests because they remove an unused packaging feature.

### DIFF
--- a/python/servo/package_commands.py
+++ b/python/servo/package_commands.py
@@ -20,7 +20,6 @@ import os.path as path
 import shutil
 import subprocess
 import sys
-import tempfile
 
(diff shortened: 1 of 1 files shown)


mukilan (approved):
I guess we can also archive [this](https://github.com/servo/homebrew-servo) repo.

PR #30048, opened 2023-07-31
Switch default layout option to `--layout-2020` in some mach commands

### LINKED ISSUE #30036: Make layout-2020 meta files the default for `./mach test-wpt`
Currently, `./mach test-wpt` defaults to using a meta file in `tests/wpt/meta-legacy-layout` (layout-2013). To use the layout-2020 meta file, we must specify the `--layout-2020` option.

Given that Servo is now built with layout 2020, it seems more natural for the layout-2020 meta files to be used by default.

ohno418:
Currently, `./mach test-wpt` family of commands (`test-wpt`, `test-wpt-android`, and `test-wpt-failure`) and `./mach update-wpt` default to using the legacy-layout option `--layout-2013` unless `--layout-2020` is specified.

Given that we are now using layout-2020 by default, this change updates these commands to default to using the `--layout-2020` option instead of `--layout-2013`.

---

This fixes https://github.com/servo/servo/issues/30036.
(I was considering addressing [this comment on the issue](https://github.com/servo/servo/issues/30036#issuecomment-1656087431) in this pull request as well. But to keep the focus of the PR clear, I have decided to handle that in a separate pull request.)

---

- [x] These changes fix #30036

### DIFF
--- a/python/wpt/update.py
+++ b/python/wpt/update.py
@@ -126,8 +126,8 @@ def run_update(**kwargs) -> bool:
 
 def create_parser(**_kwargs):
     parser = wptcommandline.create_parser_update()
-    parser.add_argument("--layout-2020", "--with-layout-2020", default=False, action="store_true",
+    parser.add_argument("--layout-2020", "--with-layout-2020", default=True, action="store_true",
                         help="Use expected results for the 2020 layout engine")
-    parser.add_argument("--layout-2013", "--with-layout-2013", default=True, action="store_true",
+    parser.add_argument("--layout-2013", "--with-layout-2013", default=False, action="store_true",
                         help="Use expected results for the 2013 layout engine")
     return parser
--- a/python/wpt/__init__.py
+++ b/python/wpt/__init__.py
@@ -33,9 +33,9 @@ def create_parser():
                         help="Run under chaos mode in rr until a failure is captured")
     parser.add_argument('--pref', default=[], action="append", dest="prefs",
                         help="Pass preferences to servo")
-    parser.add_argument('--layout-2020', '--with-layout-2020', default=False,
+    parser.add_argument('--layout-2020', '--with-layout-2020', default=True,
                         action="store_true", help="Use expected results for the 2020 layout engine")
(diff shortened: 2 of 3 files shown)


python/wpt/update.py
```diff
@@ -126,6 +126,8 @@ def run_update(**kwargs) -> bool:
 
 def create_parser(**_kwargs):
     parser = wptcommandline.create_parser_update()
+    # TODO: These two default values should be `True` and `False` to align with
+    #       `./mach test-wpt` command.
```
mrobinson, line 130: I really think we should change all scripts at once and not do this in a piecemeal way. It's very confusing when different scripts have different behaviors.

mrobinson (changes requested):
Thanks for this change. This needs a bit of work before it's ready, I think.

python/wpt/update.py
ohno418: Hm, that makes sense.

Are you referring to all scripts that have both the "--layout-2013" and "--layout-2020" options?
Since `./mach` appears to have many commands, I'm not sure I can handle all of them. (Or are there not so many commands to change?) But I'd be glad to take a look.

mrobinson: Yep, I'm referring to all commands that have those two options. A way to find them all is to grep the `python` directory for these strings. I think there are maybe 3 or 4 in total.

ohno418: It seems that only `test-wpt(-xxx)` and `update-wpt` commands are related here, if I understand correctly.

mrobinson (approved):
Thank you! This looks good to me.

mrobinson:
It looks like all Layout 2013 tests are failing, suggesting that more work needs to be done here. I think the problem is that the check for `layout_2020` on line 106 of `python/wpt/run.py` should be changed to check for `layout_2013`. The `layout_2020` key is always going to be true now, because it defaults to true.

ohno418:
@mrobinson 
Thanks for your suggested change. Done. Can you take a look?

ohno418:
Ah, it seems like you've already handled this...

PR #30053, opened 2023-08-01
Remove lzma path workaround for MacOS

### LINKED ISSUE #29736: Revert ad-hoc fix for gstreamer dylib issue.
The workaround introduced by #29733 should be reverted once upstream fixes the bug identified in the linked ticket:

https://gitlab.freedesktop.org/gstreamer/cerbero/-/issues/428

I'm creating this issue to keep track of this change on our end.

mrobinson:
Now that the new version of GStreamer fixes this issue, we can remove
the workarounds for this problem as well as all of the homebrew
bootstrapping logic.

Closes #29736  

---


python/servo/platform/macos.py
```diff
@@ -15,9 +15,9 @@
 from .. import util
 from .base import Base
 
-URL_BASE = "https://github.com/servo/servo-build-deps/releases/download/macOS"
-GSTREAMER_URL = f"{URL_BASE}/gstreamer-1.0-1.22.2-universal.pkg"
-GSTREAMER_DEVEL_URL = f"{URL_BASE}/gstreamer-1.0-devel-1.22.2-universal.pkg"
+URL_BASE = "https://github.com/servo/servo-build-deps/releases/download/macOS-2"
```
mukilan, line 18: Did you have to create a new tagged release because you were not able to edit the existing release and add the newer gstreamer packages? If so, I can just add it to the existing release as I'm able to edit them. I'll also look into how to give everyone access to edit the existing release.

mrobinson: I didn't edit the release, because I thought there was a limit on the size of the release, but it appears this isn't the case. I can use the same release in a new version of this change.

mrobinson: Okay. I've updated this change to use the same release.

mukilan (approved):
Thanks for fixing this!

mrobinson:
Looks like this failure is a GitHub infra flake:

`Error: Unable to resolve action `actions/checkout@v3`, unable to find version `v3`. Unable to resolve action `actions/download-artifact@v3`, unable to find version `v3`. Unable to resolve action `actions/upload-artifact@v3`, unable to find version `v3``

I'm sending this change back to the merge queue.

PR #30054, opened 2023-08-01
Set the MacOS rpath in the build script

mrobinson:
Use the build script to set the rpath in MacOS instead of mach. This is
another step toward allowing building servo without mach.

---

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -238,12 +238,6 @@ class MachCommands(CommandBase):
                     if not package_gstreamer_dylibs(self.cross_compile_target, servo_path):
                         return 1
 
-                    # On Mac we use the relocatable dylibs from offical gstreamer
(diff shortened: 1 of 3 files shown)


python/servo/build_commands.py
```diff
@@ -404,10 +398,6 @@ def change_link_name(binary, old, new):
     install_name_tool(binary, '-change', old, f"@executable_path/{new}")
 
 
-def add_rpath_to_binary(binary, relative_path):
-    install_name_tool(binary, "-add_rpath", relative_path)
-
-
 def change_rpath_in_binary(binary, old, new):
```
mukilan, line 401: I believe change_rpath_in_binary can also be deleted as it is not used anywhere after the hombrew logic was removed.

mrobinson: Good point! I've gone ahead and removed this function as well.

jdm (approved):

PR #30056, opened 2023-08-01
Make rustup a requirement and switch to `rust-toolchain.toml`

mrobinson:
This change makes rustup a requirement for building Servo with `./mach`
and switches to the newer `rust-toolchain.toml` format. The goal here is
to make mach builds more similar to non-mach builds.

- The new format allows listing the required components, removing some of
  the complexity from our mach scripts.
- This means we must raise the required version of rustup to 1.23. The
  current version is 1.26.
- We no longer wrap every call to cargo and rustc in "rustup run" calls
  as both cargo and rustc will take care of installing and using all
  necessary components specified in `rust-toolchain.toml` when run
  inside the project directory.

---


jdm (approved):

jdm:
Nice improvement!

sagudev:
This causes docs build to fail: https://github.com/servo/servo/actions/runs/5729143739/job/15525263105

mrobinson:
> This causes docs build to fail: https://github.com/servo/servo/actions/runs/5729143739/job/15525263105

Fix posted here: https://github.com/servo/servo/pull/30058

PR #30057, opened 2023-08-01
Handle BFC roots with auto width next to floats

Loirooriol:
The specification says that independent formatting contexts and independent block formatting contexts should be placed next to floats. The code currently handles this situation, except for the case where an independent formatting context has auto width. This PR adds support for this. This case is a big trickier, because we must try to place the object and then see if the block size calculated from layout fits at the speculated position among the floats. If it doesn't, we must try a new position and lay out again until we find an appropriate spot.

---

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -160,6 +160,12 @@ impl<'a> PlacementAmongFloats<'a> {
         return (max_inline_start, min_inline_end);
     }
 
+    /// Find the total inline size provided by the current set of bands under consideration.
+    fn calculate_viable_inline_size(&self) -> Length {
+        let (inline_start, inline_end) = self.calculate_inline_start_and_end();
+        inline_end - inline_start
+    }
+
     fn try_place_once(&mut self) -> Option<Rect<Length>> {
         assert!(!self.current_bands.is_empty());
         self.accumulate_enough_bands_for_block_size();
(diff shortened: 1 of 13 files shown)


mrobinson:
@Loirooriol I've pushed a new commit that adds some more documentation and adjust style and variable names a bit. The biggest change here is that `content_inline_size` and `content_block_size` are now just `content_size` outside the central if-else block in `mod.rs`.

mrobinson (approved):

mrobinson:
If the change looks good to you, feel free to merge or I can make adjustments to my changes.

Loirooriol:
Sorry, forgot `./mach fmt`

mrobinson:
Looks like this is failing some tests:

```
    FAIL [expected PASS] /_mozilla/css/block_formatting_context_a.html
    FAIL [expected PASS] /_mozilla/css/block_formatting_context_complex_a.html
    FAIL [expected PASS] /_mozilla/css/block_formatting_context_margin_inout_a.html
    FAIL [expected PASS] /_mozilla/css/block_formatting_context_negative_margins_a.html
    FAIL [expected PASS] /_mozilla/css/block_formatting_context_relative_a.html
    PASS [expected FAIL] /css/css-flexbox/flex-aspect-ratio-img-row-004.html
    FAIL [expected PASS] /css/css-flexbox/flex-direction-modify.html
```

Loirooriol:
Yeah I guess we should add some basic handling of the inline margins to reduce the failures.

Loirooriol:
Squashed and rebased

Loirooriol:
These failures are actually because the patch fixes the test, but the reference remains broken due to inline contents not flowing around floats:
- `/_mozilla/css/block_formatting_context_a.html`
- `/_mozilla/css/block_formatting_context_complex_a.html`
- `/_mozilla/css/block_formatting_context_relative_a.html`

These failures are due to margins, addressed by #30072
- `/_mozilla/css/block_formatting_context_margin_inout_a.html`
- `/_mozilla/css/block_formatting_context_negative_margins_a.html`
- `/css/css-flexbox/flex-direction-modify.html`

Loirooriol:
@bors-servo try=wpt-2020

Rebased on top of #30073 and updated expectations (minus the margin ones).

Loirooriol:
Actually, just let these margins test fail, since #30072 will fix them right after.

PR #30058, opened 2023-08-02
Fix the docs build

mrobinson:
Type inference was incorrectly inferring that our `check_output()`
helper was returning `str` when in reality, it returns `bytes`. This
fixes the callers that were no longer decoding those bytes and fixes the
type annotation on the function.

---

- [x] These changes do not require tests because they fix build scripts.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -141,7 +141,7 @@ def call(*args, **kwargs):
     return subprocess.call(*args, shell=sys.platform == 'win32', **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


mukilan (approved):

mrobinson:
Looks like this failure is an unrelated flake:

```
    OK /fetch/private-network-access/worker-blob-fetch.tentative.window.html
        PASS [expected FAIL] subtest: treat-as-public to public: success. 
```

PR #30060, opened 2023-08-02
Simplify solve_containing_block_padding_border_and_margin_for_in_flow_box()

Loirooriol:
Just use clamp_between_extremums() to resolve the inline size, and then only call solve_inline_margins_for_in_flow_block_level() once.

There should be no change in behavior.

---

- [X] These changes do not require tests because there should be no change in behavior


mrobinson (approved):
Nice cleanup!

PR #30061, opened 2023-08-02
Make fewer PositioningContexts when descending

mrobinson:
When descending and we have the option, don't create new
PositioningContexts just to update the static position of laid out
abspos descendants. Instead, use the new PositioningContextLength to
only update the newly added hoisted abspos boxes.

---

### DIFF
--- a/components/layout_2020/positioned.rs
+++ b/components/layout_2020/positioned.rs
@@ -158,13 +158,16 @@ impl PositioningContext {
     /// with the hoisted fragment so that it can be laid out properly at the containing
     /// block.
     ///
-    /// This function is used to update that static position at every level of the
-    /// fragment tree as the hoisted fragments move back up to their containing blocks.
-    /// Once an ancestor fragment is laid out, this function can be used to aggregate its
-    /// offset on the way back up.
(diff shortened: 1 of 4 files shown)


components/layout_2020/positioned.rs
```diff
@@ -175,20 +190,22 @@ impl PositioningContext {
     pub(crate) fn adjust_static_position_of_hoisted_fragments(
         &mut self,
         parent_fragment: &Fragment,
+        index: PositioningContextLength,
```
Loirooriol, line 193: Nit: The comment above says

> This function is used to update that static position at every level of the fragment tree

But this can now be limited with `index`, so the comment may need some update.

Loirooriol (approved):
LGTM with nit.

But this is on top of #30057, which ended up having more failures than expected. So maybe you should land this 1st, only taking the necessary parts from #30057

components/layout_2020/positioned.rs
mrobinson: Okay. I've updated this comment.

mrobinson:
Thanks for the review! I've gone ahead and taken the bits of #30057 that were relevant to this change.

Issue #30052, opened 2023-08-01
Worker.terminate() doesn't force worker threads to exit


mukilan:
With the following setup:
index.html:
```
<html>
<script>
  setInterval(() => console.log('Main thread'), 100);
  let worker = new Worker('./worker.js');
  setTimeout(() => worker.terminate(), 500);
</script>
</html>
```
worker.js
```
let i = 0
while (true) {
   i++
   if (i % 5000 === 0) {
      console.log("Worker count", i);
   }
}
```

After a few seconds, we should expect 'Worker count' logs to stops and only repeated 'Main thread' messages to be printed to console, as the main script terminates the worker after 500ms.

Actual behavior in Servo: we see the 'Worker count' messages continue to be printed indefinitely.

Possible cause:
In the [implementation of `Worker.terminate`](https://github.com/servo/servo/blob/master/components/script/dom/worker.rs#L256), the JS_RequestInterruptCallback is called with context returned by `GlobalScope::get_cx()`.
The latter, I think, just returns the context for the current thread, which will be the main script's thread where `worker.terminate()` is called and not the context associated with the Worker thread's JS engine. So Worker.terminate interrupts the main script context, instead of interrupting the Worker context, which is required to make Worker's JS engine exit out of JS loops once terminate() is called.

mukilan:
Closed via #30066

PR #30066, opened 2023-08-03
Fix race condition and other issues related to Worker destruction.

### LINKED ISSUE #30022: Race condition in Worker destruction logic leading to use-after-free of JSContext
This issue was discovered while [debugging crashes](https://github.com/servo/servo/pull/29925#issuecomment-1630108352) in CI that surfaced when updating ipc-channel and other dependencies. The full list of worker related tests that CRASH were tracked in #29990

The root cause of the crash seems to be a race condition in Servo between the logic in main Script Thread shutdown and the destruction of worker resources (JSContext and JSRuntime specifically) in the Worker thread.

In the main Script thread, when the thread is shutting down, current code performs the following sequence of events:
- Call [clear_js_runtime](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/script_thread.rs#L2953) on the  Window scope
   - Call [remove_web_messaging_and_dedicated_worker_infra](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/window.rs#L1652) on the window (as global scope)
       -  [Iterate over the  `self.list_auto_close_worker` ](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/globalscope.rs#L941) member and drop each AutoCloseWorker element (which represents shared data to communicate with a worker)
         - In the 'Drop' implementation of the AutoCloseWoker struct:
           - [Set the atomic 'closing' boolean](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/globalscope.rs#L149) to true to signal to the worker that it is being shutdown.
           - [Send the Exit message](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/globalscope.rs#L153) to the worker's control channel to make it exit out of event handling logic
           - [Call request_interrupt](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/globalscope.rs#L159) on the Worker's JSContext pointer, which internally calls the JS_RequestInterruptCallback SM API
              - *In SpiderMonkey C++ land*, JS_RequestInterruptCallback [calls the cx->requestInterrupt() member](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/jsapi.cpp#L2904) of the passed in JSContext `cx` which will:
                - [Set the interruptBits](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/vm/Runtime.cpp#L442) and jitStackLimit member variables of the JSContext class to signal the VM/JIT loops 
                -  [Call wasm::InterruptRunningCode ](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/vm/Runtime.cpp#L454)with JSContext passed in as parameter `cx` which will
                   - [Retrieve the member cx->runtime_](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/wasm/WasmRealm.cpp#L139) to get the JSRuntime which contains the necessary WASM state, including mutexes that need to be locked.

In the tests that are crashing we see that is the very last step that is causing the segmentation fault. In particular, when cx->runtime_ is obtained we get the value `0x3b3b3b3b3b3b3b3b` instead of a valid `JSRuntime*`. Subsequent field access, such as the WASM pthread_mutex are then based on this invalid pointer (which is confirmed by the address  we see in the [backtrace](https://gist.github.com/mukilan/ae55cae1bc376f7132ca362de61602cc#file-gistfile1-txt-L50))

Based on the above, and [runs in CI](https://github.com/mukilan/servo/actions/runs/5635117915/job/15265853539) with [logs added to SM ](https://github.com/mukilan/servo/commit/5027cccb6c8a32be3b527b16ea3c02c9bbb77f86) to record the values of JSContext* and its associated JSRuntime*, it seems like cx->runtime is valid when the context is created, but then becomes invalid when Wasm::InterruptRunningCode runs.

The current theory is that the following interleaved sequence of events lead to JSContext* being used after it was already destroyed:

 1. [Main Script thread]  [Drop implementation of AutoCloseWorker sets the Atomic 'closing' flag](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/globalscope.rs#L149). The thread is unscheduled by OS just after this.
 2. [Worker thread] [The closing flag is checked, the worker exits the event loop](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/dedicatedworkerglobalscope.rs#L471) and [calls `clear_js_runtime` ](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/dedicatedworkerglobalscope.rs#L479) on the CX associated with it's worker scope.
    2.1 The` clear_js_runtime` [eventually](https://github.com/servo/servo/blob/605bf5233490cbd92aa5d3db78ee4182a7a70d34/components/script/dom/workerglobalscope.rs#L175) [calls](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/rust-mozjs/src/rust.rs#L417) [`DestroyContext`](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/vm/JSContext.cpp#L201) which destroys the JSContext and JSRuntime instances and also [**poisons them**](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/vm/JSContext.cpp#L226), by [memset-ing the contents of their memory to be the poison value `0x3b`](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/public/Utility.h#L564)
3. [Main Script thread] is scheduled back, runs the remainining logic which includes calling `request_interrupt` on the `JSContext*` associated with each worker. Since that pointer now points to freed memory overwritten with 0x3b[, the value of cx->runtime_ is `0x3b3b3b3b3b3b3b3b`](https://github.com/servo/mozjs/blob/8c656b49a0ef1c9078aba3145b3e6b2660e44572/mozjs/mozjs/js/src/wasm/WasmRealm.cpp#L139), explaining the crash in the last step.

I've [confirmed](https://github.com/mukilan/servo/actions/runs/5637913509/job/15271736762) that the crashes are resolved after commenting out the clear_js_runtime step in Worker (step 2 above).

I'm not entirely sure what the ideal solution must be. It seems like the destruction of Worker's JSContext* and JSRuntime* needs to be performed by the owning Script thread rather than the worker's themselves, but I'm not sure if SpiderMonkey design allows this.

mukilan:
This PR has three fixes - the first commit contains fix for #30022 and #30052. The second commit fixes another issue in Worker.importScripts that was uncovered after #30052 was fixed.

### Fix for #30022 (Race condition in worker destruction leading to use-after-free of JSContext) -

During shutdown, the main script thread calls JS_RequestInterruptCallback(cx) for each worker thread it owns where cx is the JSContext* created for that worker. Although JS_RequestInterruptCallback is safe to call from threads other than the worker thread, it is possible the JSContext* has already been destroyed
    
For example, as noted in #30022, since the main thread sets the worker's `closing` flag to true to signal termination before it calls JS_RequestInterruptCallback, we can have a race condition where the worker exits its event loop when `closing` flags is set and then it (worker thread) destroys its own JSContext and JSRuntime. When the main thread resumes, it will call JS_RequestInterruptCallback for the worker's context, leading ta a use-after-free bug.
    
This patch solves this issue by improving the existing `ContextForRequestInterrupt` abstraction used for sharing the Worker's associated JSContext* with the parent script thread. Instead of simply wrapping a plain `*mut JSContext`, we now wrap the `*mut JSContext` in a nullable mutex i.e Mutex<Option<*mut JSContext>>
   
The mutex lock needs to be held by the parent thread when it calls JS_RequestInterruptCallback. Similary, before the worker destroys its JSContext, it locks and  sets the Option to None, signaling that the JSContext can no longer be used for interrupting the worker.

### Fix for #30052 - Woker.terminate doesn't force worker to exit.   
This patch also fixes the issue in #30052 by enforcing the use of ContextForRequestInterrupt abstraction which ensures the correct JSContext is used by the main thread when Worker.terminate is called.

Note: There maybe more optimizations possible here that we can explore in the future, like re-using the ContextForRequestInterrupt already stored in Worker's global scope as part of [list_auto_close_worker](https://github.com/servo/servo/blob/3fea90a231a94338d67712398fe3d2ba9d402211/components/script/dom/globalscope.rs#L267) instead of storing it again in Worker DOM struct.

### Fix for Worker.importScripts
After the fix for #30052,
the WPT test `/workers/Worker-terminate-forever-during-evaluation.html` started to crash, because when evaluation doesn't succeed, `importScripts` always returns `Error::JSFailed` code to the caller, which indicates that there is a Dom/JS exception to be thrown. However, this is not true when the script is terminated, which causes the generated binding layer for `importScript` to fail the assertion that there is a pending exception. This patch makes `importScripts` work similar to the [logic that evaluates the top-level script](https://github.com/servo/servo/blob/3fea90a231a94338d67712398fe3d2ba9d402211/components/script/dom/workerglobalscope.rs#L434) of the Worker - it simply prints 'evaluate_script failed - (terminated)' if the worker is terminating

---

- [x] These changes fix #30022, #30052
- [x] These changes do not require tests because the fix for race condition is covered by the existing tests that surfaced the issue in CI. The fix for Worker.terminate does have an existing test `workers/Worker-terminate-forever.html` but this doesn't currently catch the issue as it doesn't recognize the case where the "worker is terminated because it was interrupted when terminate() is called" as distinct from the case where the "main script's shutting down forces worker to be interrupted and closed" eventually. I'm not sure what is the best way to express this case using the WPT test harness api.

### DIFF
--- a/components/script/dom/dedicatedworkerglobalscope.rs
+++ b/components/script/dom/dedicatedworkerglobalscope.rs
@@ -378,7 +378,8 @@ impl DedicatedWorkerGlobalScope {
                     new_child_runtime(parent, Some(task_source))
                 };
 
-                let _ = context_sender.send(ContextForRequestInterrupt::new(runtime.cx()));
+                let context_for_interrupt = ContextForRequestInterrupt::new(runtime.cx());
+                let _ = context_sender.send(context_for_interrupt.clone());
 
                 let (devtools_mpsc_chan, devtools_mpsc_port) = unbounded();
                 ROUTER.route_ipc_receiver_to_crossbeam_sender(
@@ -476,7 +477,8 @@ impl DedicatedWorkerGlobalScope {
                         parent_sender,
                         CommonScriptMsg::CollectReports,
                     );
-                scope.clear_js_runtime();
+
+                scope.clear_js_runtime(context_for_interrupt);
             })
             .expect("Thread spawning failed")
     }
--- a/components/script/dom/serviceworkerglobalscope.rs
+++ b/components/script/dom/serviceworkerglobalscope.rs
@@ -306,7 +306,8 @@ impl ServiceWorkerGlobalScope {
             .spawn(move || {
                 thread_state::initialize(ThreadState::SCRIPT | ThreadState::IN_WORKER);
                 let runtime = new_rt_and_cx(None);
-                let _ = context_sender.send(ContextForRequestInterrupt::new(runtime.cx()));
+                let context_for_interrupt = ContextForRequestInterrupt::new(runtime.cx());
+                let _ = context_sender.send(context_for_interrupt.clone());
 
                 let roots = RootCollection::new();
                 let _stack_roots = ThreadLocalStackRoots::new(&roots);
@@ -396,7 +397,8 @@ impl ServiceWorkerGlobalScope {
                         scope.script_chan(),
                         CommonScriptMsg::CollectReports,
                     );
-                scope.clear_js_runtime();
+
+                scope.clear_js_runtime(context_for_interrupt);
             })
             .expect("Thread spawning failed")
     }
--- a/components/script/dom/worker.rs
+++ b/components/script/dom/worker.rs
@@ -4,6 +4,7 @@
 
 use crate::dom::abstractworker::SimpleWorkerErrorHandler;
 use crate::dom::abstractworker::WorkerScriptMsg;
+use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::MessagePortBinding::PostMessageOptions;
 use crate::dom::bindings::codegen::Bindings::WorkerBinding::{WorkerMethods, WorkerOptions};
 use crate::dom::bindings::error::{Error, ErrorResult, Fallible};
@@ -23,13 +24,13 @@ use crate::dom::messageevent::MessageEvent;
 use crate::dom::window::Window;
 use crate::dom::workerglobalscope::prepare_workerscope_init;
 use crate::realms::enter_realm;
-use crate::script_runtime::JSContext;
+use crate::script_runtime::{ContextForRequestInterrupt, JSContext};
 use crate::task::TaskOnce;
 use crossbeam_channel::{unbounded, Sender};
 use devtools_traits::{DevtoolsPageInfo, ScriptToDevtoolsControlMsg, WorkerId};
 use dom_struct::dom_struct;
 use ipc_channel::ipc;
-use js::jsapi::{Heap, JSObject, JS_RequestInterruptCallback};
+use js::jsapi::{Heap, JSObject};
 use js::jsval::UndefinedValue;
 use js::rust::{CustomAutoRooter, CustomAutoRooterGuard, HandleObject, HandleValue};
 use script_traits::{StructuredSerializedData, WorkerScriptLoadOrigin};
@@ -51,6 +52,8 @@ pub struct Worker {
     #[ignore_malloc_size_of = "Arc"]
     closing: Arc<AtomicBool>,
     terminated: Cell<bool>,
+    #[ignore_malloc_size_of = "Arc"]
+    context_for_interrupt: DomRefCell<Option<ContextForRequestInterrupt>>,
 }
 
 impl Worker {
@@ -60,6 +63,7 @@ impl Worker {
             sender: sender,
             closing: closing,
             terminated: Cell::new(false),
+            context_for_interrupt: Default::default(),
         }
     }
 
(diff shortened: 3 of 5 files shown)


jdm (approved):
Wonderful! Thank you for tracking this down.

PR #30072, opened 2023-08-04
Handle inline margins when avoiding floats

### LINKED ISSUE #30050: Inline margins of a BFC root should be allowed to overlap floats
Similarly to #30037, but in the inline axis, margins of a replaced element or BFC roots should be able to overlap floats

```html
<!DOCTYPE html>
<style>
.float { float: left; width: 50px; height: 250px; background: cyan; }
.float ~ div { display: flow-root; width: 25px; height: 25px; }
</style>
<div style="display: flow-root; width: 100px; border: solid">
  <div class="float"></div>

  <div style="background: magenta; margin-left: 0px; margin-right: auto"></div>
  <div style="background: magenta; margin-left: 25px; margin-right: auto"></div>
  <div style="background: magenta; margin-left: 50px; margin-right: auto"></div>
  <div style="background: magenta; margin-left: 75px; margin-right: auto"></div>

  <div style="background: orange; margin-left: auto; margin-right: 0px"></div>
  <div style="background: orange; margin-left: auto; margin-right: 25px"></div>
  <div style="background: orange; margin-left: auto; margin-right: 50px"></div>
  <div style="background: orange; margin-left: auto; margin-right: 75px"></div>

  <div style="background: lime; margin-left: auto; margin-right: auto"></div>
</div>
```

| Servo | Blink | Firefox | WebKit |
| - | - | - | - |
|  |  |  |  |

Loirooriol:
---

- [X] These changes fix #30050

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -13,7 +13,7 @@ use crate::formatting_contexts::IndependentFormattingContext;
 use crate::fragment_tree::{BoxFragment, CollapsedBlockMargins, CollapsedMargin, FloatFragment};
 use crate::geom::flow_relative::{Rect, Vec2};
 use crate::positioned::PositioningContext;
-use crate::style_ext::{ComputedValuesExt, DisplayInside};
+use crate::style_ext::{ComputedValuesExt, DisplayInside, PaddingBorderMargin};
 use crate::ContainingBlock;
 use euclid::num::Zero;
 use servo_arc::Arc;
@@ -85,6 +85,11 @@ pub(crate) struct PlacementAmongFloats<'a> {
     /// The minimum position in the block direction for the placement. Objects should not
     /// be placed before this point.
     ceiling: Length,
+    /// The inline position where the object would be if there were no floats. The object
+    /// can be placed after it due to floats, but not before it.
+    min_inline_start: Length,
+    /// The maximum inline position that the object can attain when avoiding floats.
+    max_inline_end: Length,
 }
 
 impl<'a> PlacementAmongFloats<'a> {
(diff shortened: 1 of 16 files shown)


Loirooriol:
This is the correct job: https://github.com/servo/servo/actions/runs/5762557519

mrobinson:
After the fixes this morning, the job linked above  ([#5762472065](https://github.com/servo/servo/actions/runs/5762472065)) should test commit 7ad9411a2e52aabf20ac8919f055b66d6d09648a (which I think is the right one). You can see the checkout action checking out the proper commit here: https://github.com/servo/servo/actions/runs/5762472065/job/15622296584

mrobinson:
Here's the link to the check out hash: https://github.com/servo/servo/actions/runs/5762472065/job/15622296584#step:3:589

Loirooriol:
I think I disagree with `new-fc-beside-float-with-margin.html`.
Adopting that behavior would imply either inconsistencies with the `width: auto` case, or severely nonsensical behaviors.
So I guess I will just let it fail, and add new tests instead.

mrobinson:
> I think I disagree with `new-fc-beside-float-with-margin.html`. Adopting that behavior would imply either inconsistencies with the `width: auto` case, or severely nonsensical behaviors. So I guess I will just let it fail, and add new tests instead.

Does it make sense to open a WPT or spec bug for this test?

mrobinson:
Thanks for the background. Either of those things seem fine to me.

Loirooriol:
I could add more tests of cases that seem broken in other browsers, but 7 passes seem enough for now.

components/layout_2020/flow/float.rs
```diff
@@ -81,29 +81,42 @@ pub(crate) struct PlacementAmongFloats<'a> {
...
+    min_inline_start: Length,
+    /// The maximum inline position that the object can attain when avoiding floats.
+    max_inline_end: Length,
 }
 
 impl<'a> PlacementAmongFloats<'a> {
     pub(crate) fn new(
         float_context: &'a FloatContext,
         ceiling: Length,
-        object_size: Vec2<Length>,
```
mrobinson, line 94: I think it makes sense to simply clone the input to this function instead of passing a reference.

mrobinson (approved):
Great change. Looks good to me with a small suggestion proposed below. I think it's also a good idea to open specification issue about the behavior of margins when placing an independent formatting context / replaced element next to a float.

components/layout_2020/flow/float.rs
Loirooriol: Done

Loirooriol:
Forgot to check the new tests in layout-2013

Loirooriol:
I filed https://github.com/w3c/csswg-drafts/issues/9174

Turns out Servo was already differing from other browsers even without floats: in over-constrained cases, `getComputedStyle` provides the used value (according to the spec), other browsers just absolutize the computed value.

Also, including the adjustment to avoid floats into the used value of margins means that we have round-tripping. Other browsers don't round-trip `auto` margins, which seems quite bad.

PR #30077, opened 2023-08-07
Pass `--layout-2013` flag in 'WPT import' job

mukilan:
PR #30048 switched `./mach update-wpt` to use 2020 layout engine by default.  Since the WPT import job was not passing any flags when updating 2013 expectations, it was not using the correct metadata files, leading to failures when landing the recent [wpt sync PR](https://github.com/servo/servo/pull/30075)

---

- [x] These changes do not require tests because they just fix bug in WPT import job


mrobinson (approved):
Thank you for this fix!

PR #30079, opened 2023-08-08
mach: fix exit status on NixOS

delan:
mach always exits with status code 0 on NixOS, which makes it impossible to check if the build failed programmatically. This patch fixes that.

---

- [x] These changes do not require tests because they affect the interactive tooling only


mukilan (approved):
Nice catch!

PR #30085, opened 2023-08-09
Resolve cyclic margin and padding percentages against zero

### LINKED ISSUE #30084: Intrinsic sizing algorithm treats percentages nonsensically
While working on #30083 I noticed that the intrinsic sizing algorithm treats percentages in margins or paddings in a strange way.

From https://drafts.csswg.org/css-sizing-3/#min-percentage-contribution

> For the [min size properties](https://drafts.csswg.org/css-sizing-3/#min-size-properties), as well as for **[margins](https://drafts.csswg.org/css-box-4/#margin) and [paddings](https://drafts.csswg.org/css-box-4/#padding)** (and [gutters](https://drafts.csswg.org/css-align-3/#gutter)), a [cyclic percentage](https://drafts.csswg.org/css-sizing-3/#cyclic-percentage-size) is **resolved against zero** for determining [intrinsic size contributions](https://drafts.csswg.org/css-sizing-3/#intrinsic-size-contribution).

Instead, `outer_inline_and_percentages` decomposes `<length-percentage>` into its `<length>` and `<percentage>` components, and adds them separately. So it returns a `ContentSizes` generated with the lengths, and a `Percentage` which is the sum of the percentages in margins and paddings.

Then, `adjust_for_pbm_percentages` does this:

```rs
let denominator = (1. - percentages.0).max(0.);
self.max_content = Length::new(self.max_content.px() / denominator);
```

So for example:

```html
<!DOCTYPE html>
<p id="log"></p>
<div id="target" style="float: left; outline: solid">
  <div style="overflow: hidden; width: 100px; height: 100px; margin: 25%; background: cyan"></div>
</div>
<script>
log.textContent = target.offsetWidth + " x " + target.offsetHeight;
</script>
```

| Servo | Expected |
| - | - |
|  |  |

Servo sizes to `100px / (1 - 0.25 - 0.25) = 200px`, and I guess it looks nicer, but it should just be `100px`.

Loirooriol:
From https://drafts.csswg.org/css-sizing-3/#min-percentage-contribution

> For the min size properties, as well as for margins and paddings
> (and gutters), a cyclic percentage is resolved against zero
> for determining intrinsic size contributions.

---

- [X] These changes fix #30084

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -27,7 +27,7 @@ use servo_arc::Arc;
 use style::computed_values::white_space::T as WhiteSpace;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
-use style::values::computed::{Length, LengthPercentage, Percentage};
+use style::values::computed::Length;
 use style::values::generics::text::LineHeight;
 use style::values::specified::text::TextAlignKeyword;
 use style::values::specified::text::TextDecorationLine;
@@ -251,7 +251,6 @@ impl InlineFormattingContext {
             containing_block_writing_mode: WritingMode,
             paragraph: ContentSizes,
             current_line: ContentSizes,
-            current_line_percentages: Percentage,
             /// Size for whitepsace pending to be added to this line.
             pending_whitespace: Length,
             /// Whether or not this IFC has seen any non-whitespace content.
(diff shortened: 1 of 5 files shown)


Loirooriol:
The flexbox test fails because `writing-mode: vertical-lr` is not supported yet.

components/layout_2020/flow/inline.rs
```diff
@@ -274,11 +273,12 @@ impl InlineFormattingContext {
                             macro_rules! add {
                                 ($condition: ident, $side: ident) => {
                                     if inline_box.$condition {
-                                        self.add_lengthpercentage(padding.$side);
-                                        self.add_length(border.$side);
+                                        let zero = Length::zero();
```
mrobinson, line 276: I think it makes sense to link and quote the specification here.

components/layout_2020/sizing.rs
```diff
@@ -73,34 +60,21 @@ pub(crate) fn outer_inline(
...
     let margin = style.margin(containing_block_writing_mode);
 
-    let mut pbm_percentages = Percentage::zero();
-    let mut decompose = |x: &LengthPercentage| {
-        pbm_percentages += x.to_percentage().unwrap_or_else(Zero::zero);
-        x.to_length().unwrap_or_else(Zero::zero)
-    };
-    let pb_lengths =
-        border.inline_sum() + decompose(padding.inline_start) + decompose(padding.inline_end);
+    // Percentages for paddings and margins are treated as zero
```
mrobinson, line 67: Ditto.

mrobinson (approved):

Issue #27457, opened 2020-07-30
Border width is inconsistent at device pixel ratio 1.5


SamRodri:
In Windows 10 with display scaling set to 150%, this HTML file renders different from Firefox:

```html
<!doctype html>
<html lang="fr">
<head>
	<meta charset="utf-8">
	<title></title>
	<style>
		#border {
			border: 1px solid red;
			width: 50px;
			height: 50px;
			margin: 50px;
		}
	</style>
</head>
<body>
	<p id="px_ratio">device pixel ratio is</p>
	<div id="border"></div>
	<script>
		document.getElementById("px_ratio").innerText += " " + window.devicePixelRatio;
	</script>
</body>
</html>
```

<img width="439" alt="border_error" src="https://user-images.githubusercontent.com/449925/88966694-1ac6b400-d283-11ea-8de0-e529d7d86288.png">

This Webrender snippet also causes the problem:

```rust
let border_side = BorderSide {
    color: ColorF::new(1.0, 0.0, 0.0, 1.0),
    style: BorderStyle::Solid,
};
let border_widths = LayoutSideOffsets::new_all_same(1.0);
let border_details = BorderDetails::Normal(NormalBorder {
    top: border_side,
    right: border_side,
    bottom: border_side,
    left: border_side,
    radius: BorderRadius::zero(),
    do_aa: true,
});
let bounds = LayoutRect::new(LayoutPoint::new(50.0, 50.0), LayoutSize::new(50.0, 50.0));
builder.push_border(
    &CommonItemProperties::new(
        bounds,
        SpaceAndClipInfo { spatial_id, clip_id: ClipId::root(pipeline_id) },
    ),
    bounds,
    border_widths,
    border_details,
);  
```

SimonSapin:
The key word is "pixel snapping".

I can also reproduce with device pixel ratio 1, if the specified border is `1.5px`. What matters is the border width not being an integer of *device pixels* after applying CSS transforms, user zoom, DPR, etc.

What happens is that when painting axis aligned lines for solid borders, we typically don’t want the blurry rendering that sub-pixel positioning with anti-aliasing would give. So we round the width to an integer number of device pixels and/or "snap" edges to the device pixel grid. Depending on the exact code path, 1.5 can end up rounded to 1 or 2.

[...]

PR #30086, opened 2023-08-10
Apply pixel rounding to border-width values

shanehandley:
I was looking into some WPT test failures related to [css-borders](https://wpt.fyi/results/css/css-borders?run_id=5199197094019072) and found that CSS border values were not being rounded according to the rules defined in the spec for CSS lengths.

https://drafts.csswg.org/css-values-4/#snap-a-length-as-a-border-width

It seems necessary to make this adjustment both when drawing the borders, as well as calculating the dimensions of the box fragment.

This change is naive (so marked as draft) as I don't have a good understanding of the codebase, and it does not consider the device pixel ratio (https://github.com/servo/servo/issues/27457) but several WPT tests seem to pass with this adjustment.

Happy to rework this with direction from someone more familiar with the code, or close in favour of an alternative implementation. Cheers.

---

- [x] These changes do not require tests because: The implementation of this fix is likely to change.

### DIFF
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -675,10 +675,10 @@ impl<'a> BuilderForBoxFragment<'a> {
     fn build_border(&mut self, builder: &mut DisplayListBuilder) {
         let border = self.fragment.style.get_border();
         let widths = SideOffsets2D::new(
-            border.border_top_width.px(),
-            border.border_right_width.px(),
-            border.border_bottom_width.px(),
-            border.border_left_width.px(),
+            border.border_top_width.0.round_to_pixels().px(),
+            border.border_right_width.0.round_to_pixels().px(),
+            border.border_bottom_width.0.round_to_pixels().px(),
+            border.border_left_width.0.round_to_pixels().px(),
         );
         if widths == SideOffsets2D::zero() {
             return;
--- a/components/layout_2020/style_ext.rs
+++ b/components/layout_2020/style_ext.rs
@@ -313,10 +313,10 @@ impl ComputedValuesExt for ComputedValues {
         let border = self.get_border();
         flow_relative::Sides::from_physical(
             &PhysicalSides::new(
-                border.border_top_width.0,
-                border.border_right_width.0,
-                border.border_bottom_width.0,
-                border.border_left_width.0,
+                border.border_top_width.0.round_to_pixels(),
+                border.border_right_width.0.round_to_pixels(),
+                border.border_bottom_width.0.round_to_pixels(),
+                border.border_left_width.0.round_to_pixels(),
             ),
             containing_block_writing_mode,
         )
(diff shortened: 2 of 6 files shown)


shanehandley:
Looks to have been addressed recently, but is unmerged.

components/style/values/computed/length.rs
```diff
@@ -304,6 +304,22 @@ impl CSSPixelLength {
...
+    /// device pixels.
+    pub fn round_to_pixels(self) -> Self {
+        if self.0 <= 0.0 {
+            CSSPixelLength::new(0.)
+        } else if self.0 < 1.0 {
+            CSSPixelLength::new(self.0.ceil())
+        } else {
+            CSSPixelLength::new(self.0.floor())
+        }
+    }
```
mrobinson, line 322: This isn't clear at all from the code or documentation, but `components/style` is "vendored" from upstream gecko. We are in the process of syncing the two versions. We generally shouldn't be changing code here unless it is Servo-specific.

You could expose this as a trait in `style_ext.rs` instead, I think.

components/layout_2020/display_list/mod.rs
```diff
@@ -675,10 +675,10 @@ impl<'a> BuilderForBoxFragment<'a> {
...
         let border = self.fragment.style.get_border();
         let widths = SideOffsets2D::new(
-            border.border_top_width.px(),
-            border.border_right_width.px(),
-            border.border_bottom_width.px(),
-            border.border_left_width.px(),
+            border.border_top_width.0.round_to_pixels().px(),
+            border.border_right_width.0.round_to_pixels().px(),
+            border.border_bottom_width.0.round_to_pixels().px(),
+            border.border_left_width.0.round_to_pixels().px(),
```
mrobinson, line 681: I think this might be incorrect as the "pixels" used during layout are not device pixels. A hidpi scale is applied in WebRender. This scaling should move back into layout pretty soon, I think which will make it easier to round to device pixels there.

components/layout_2020/style_ext.rs
```diff
@@ -313,10 +313,10 @@ impl ComputedValuesExt for ComputedValues {
...
         flow_relative::Sides::from_physical(
             &PhysicalSides::new(
-                border.border_top_width.0,
-                border.border_right_width.0,
-                border.border_bottom_width.0,
-                border.border_left_width.0,
+                border.border_top_width.0.round_to_pixels(),
+                border.border_right_width.0.round_to_pixels(),
+                border.border_bottom_width.0.round_to_pixels(),
+                border.border_left_width.0.round_to_pixels(),
```
mrobinson, line 319: This code has the same issue, I think. I do wonder why you need to round in both places though...shouldn't rounding in one be enough?

mrobinson (changes requested):
Thanks for this contribution. There are some issues to be sorted out which I've detailed below.

shanehandley:
Hey @mrobinson thanks for taking a look 👍 I closed this one as I saw that there is a pending PR here which addresses this issue, and is the author knows the codebase much better:

https://github.com/servo/servo/pull/29718

Thank you for the comments - I'm working on getting my head around the codebase, but the servo/gecko overlap does catch me out 👍

PR #30088, opened 2023-08-10
Fix viewport origin of webrender frame

wusyong:
If we try to provide [viewport origin](https://github.com/servo/servo/blob/master/ports/winit/headed_window.rs#L529) other that (0, 0), the origin of the document viewport will still be fixed to (0, 0) initially and it's still a bit off if we resize the window. This PR tries to fix the origin by `set_document_view` when initializing webrender document and passing correct viewport origin.

The flipped viewport is only useful when drawing directly like [clearing background](https://github.com/servo/servo/blob/master/components/compositing/compositor.rs#L1742) through gl api directly.
Webredner itself will detect this and proving correct origin when drawing frames.

---

### DIFF
--- a/components/compositing/windowing.rs
+++ b/components/compositing/windowing.rs
@@ -200,13 +200,13 @@ pub struct EmbedderCoordinates {
 }
 
 impl EmbedderCoordinates {
-    // Get the viewport for webrender usage.
+    /// Get the unflipped viewport rectangle for use with the WebRender API.
     pub fn get_viewport(&self) -> DeviceIntRect {
         DeviceIntRect::from_untyped(&self.viewport.to_untyped())
     }
 
-    // Get the flipped viewport. This is used when we need to draw directly
-    // since origin of webrender is usually bottom left.
+    /// Get the flipped viewport rectangle. This should be used when drawing directly
+    /// to the framebuffer with OpenGL commands.
     pub fn get_flipped_viewport(&self) -> DeviceIntRect {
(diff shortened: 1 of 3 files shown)


components/compositing/windowing.rs
```diff
@@ -200,6 +200,13 @@ pub struct EmbedderCoordinates {
 }
 
 impl EmbedderCoordinates {
+    // Get the viewport for webrender usage.
```
mrobinson, line 203: Rustdoc comments should start with three `/` so this should be something like:

```suggestion
    /// Get the unflipped viewport rectangle for use with the WebRender API.
```

components/compositing/windowing.rs
```diff
@@ -200,6 +200,13 @@ pub struct EmbedderCoordinates {
...
+    pub fn get_viewport(&self) -> DeviceIntRect {
+        DeviceIntRect::from_untyped(&self.viewport.to_untyped())
+    }
+
+    // Get the flipped viewport. This is used when we need to draw directly
+    // since origin of webrender is usually bottom left.
```
mrobinson, line 209:
```suggestion
    /// Get the flipped viewport rectangle. This should be used when drawing directly
    /// to the framebuffer with OpenGL commands.
```

mrobinson (changes requested):

mrobinson (approved):

mrobinson:
Thanks for this fix!

PR #30091, opened 2023-08-11
Make the `--release`/`--dev` more consistent and less surprising

mrobinson:
There were some issues with the way that the `--release` and `--dev` arguments were handled in mach commands.

 - Not all commands accepted them in the same way. For instance `./mach test-wpt` didn't really accept them at all.
 - If you did not pass either of them, mach would try to guess which build you meant. This guess was often quite surprising as it wasn't printed and it depended on the state of the your target directory, which is difficult to remember.
 - The `dev` profile is colloquially called a "debug" profile and some commands accepted `-d` or `--debug...` like arguments, but `--debug` with `./mach run` meant run in a debugger. It was easy to mix this up.

This change:

 - Centralizes where build type argument processing happens. Now it the same shared decorator in CommandBase.
 - Uses a `BuildType` enum instead of passing around two different booleans. This reduces the error checking for situations where both are true.
 - Be much less clever about guessing what build to use. Now if you don't specify a build type, `--dev` is chosen. I think this behavior matches cargo.
 - Makes it so that `./mach test-wpt` accepts the exact same arguments and has the same behavior as other commands. In addition, the suite correct for `test-wpt` is removed. There are only two suites now and it's quite unlikely that people will confuse WPT tests for rust unit tests.

---

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -35,21 +35,13 @@ from mach.registrar import Registrar
 import servo.platform
 import servo.util
 
-from servo.command_base import CommandBase, call, check_call
+from servo.command_base import BuildType, CommandBase, call, check_call
 from servo.gstreamer import windows_dlls, windows_plugins, macos_plugins
 
 
 @CommandProvider
 class MachCommands(CommandBase):
-    @Command('build',
-             description='Build Servo',
-             category='build')
-    @CommandArgument('--release', '-r',
-                     action='store_true',
-                     help='Build in release mode')
-    @CommandArgument('--dev', '-d',
-                     action='store_true',
-                     help='Build in development mode')
+    @Command('build', description='Build Servo', category='build')
     @CommandArgument('--jobs', '-j',
                      default=None,
                      help='Number of jobs to run in parallel')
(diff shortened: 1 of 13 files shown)


python/servo/build_commands.py
```diff
@@ -64,41 +56,14 @@ class MachCommands(CommandBase):
...
         has_media_stack = "media-gstreamer" in self.features
 
-        release_path = path.join(self.target_path, "release", "servo")
-        dev_path = path.join(self.target_path, "debug", "servo")
-
-        release_exists = path.exists(release_path)
-        dev_exists = path.exists(dev_path)
-
-        if not (release or dev):
-            if self.config["build"]["mode"] == "dev":
```
mukilan, line 80: I believe `self.config["build"]["mode"]` is what the user has specified in their `.servobuild` file. Wouldn't removing this code also break support for specifying build mode via `.servobuild`?

mrobinson: Hrm. That's a good point. I did end up removing this functionality, but probably this should be removed more thoroughly if I'm removing it here.

My thought is that we should remove this setting entirely. One day hopefully, when you pass --dev or --release to cargo this will work like passing the same option to mach. Instead of configuring how mach builds, you will configure how cargo builds.

mrobinson: I've gone ahead and removed the build mode configuration from the `servobuild.example` file.

mukilan: FWIW, I was thinking we should retain support until we remove the concept of `.servobuild` completely. My main concern is that it seems counterintuitive to support `.servobuild` but not allow complete build configuration to be specified via the file, but I guess .servobuild is not used that much in practice.

mukilan (approved):

mrobinson:
> warning Try run ([#5853547785](https://github.com/servo/servo/actions/runs/5853547785)) failed.

Of course this fails because try runs the workflow definitions from the master branch.

python/servo/build_commands.py
mrobinson: I've gone ahead and restored the `.servobuild` functionality for selecting the default build type. This is now printed a well.

PR #30093, opened 2023-08-11
Remove unused import

Loirooriol:
---


mrobinson (approved):
I think I asked someone to remove this change recently from an unrelated PR because we are pulling changes from style. You are managing pulling those changes though, so I think it's okay.

Loirooriol:
This is outside of `components/style/`, so I think it's fine.

PR #29804, opened 2023-05-28
bump base64 from 0.10 to 0.21

dvtkrlbs:
---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes partly fix #26013 
- [x] These changes do not require tests because it bumps version of a dependency. From running the tests there does not seem to be any regressions (there are a lot of crashed tests but I don't think they are relevant to this change)

PR #30094, opened 2023-08-11
Tackle half of canvas.toDataURL's FIXMEs

EnnuiL:
I was considering updating the image-rs crate (again) due to WebP support being further fixed with 0.24.7, but then I started to think about what sorts of outdated code would need to be updated; I ended falling on an unrelated rabbit hole that led to this PR happening!

About the TODOs:
- As far as I know, no, we should not use `URL_SAFE` instead of `STANDARD` here; testing on Chrome, Firefox and GNOME Web have shown that they all generate a data URL that uses `STANDARD`'s Base64 alphabet, and well, attempting to use `URL_SAFE` anyway results into a data URL that no other browser will accept
- After falling into the rabbit hole that was that PR link, I've decided to take advantage of the fact that we're on 2023 and that #29804 was a thing that happened in order to use a streaming encoder for Base64; something to note here though is that `EncoderStringWriter` does perform UTF-8 validation checks while the past attempt to tackle this TODO forgoes that, which might have been more performant?

While I attempted to locally run the tests here, I have lost myself in the process so I instead attempted to test it through web pages that did use `canvas.toDataURL`, and yep, I could confirm that nothing broke (at least in an obvious manner) in the process; It did appear that no test results have been changed during my attempt to run the tests though

---

### DIFF
--- a/components/script/dom/htmlcanvaselement.rs
+++ b/components/script/dom/htmlcanvaselement.rs
@@ -31,7 +31,6 @@ use crate::dom::virtualmethods::VirtualMethods;
 use crate::dom::webgl2renderingcontext::WebGL2RenderingContext;
 use crate::dom::webglrenderingcontext::WebGLRenderingContext;
 use crate::script_runtime::JSContext;
-use base64::Engine;
 use canvas_traits::canvas::{CanvasId, CanvasMsg, FromScriptMsg};
 use canvas_traits::webgl::{GLContextAttributes, WebGLVersion};
 use dom_struct::dom_struct;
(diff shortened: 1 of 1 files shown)


mrobinson:
Look like the failing jobs are infrastructure flakes, so I've rerun them.

components/script/dom/htmlcanvaselement.rs
```diff
@@ -425,16 +424,17 @@ impl HTMLCanvasElementMethods for HTMLCanvasElement {
...
         // FIXME(nox): https://github.com/image-rs/image-png/issues/87
-        PngEncoder::new(&mut png)
+        PngEncoder::new(&mut encoder)
             .write_image(&file, self.Width(), self.Height(), ColorType::Rgba8)
             .unwrap();
-        let mut url = "data:image/png;base64,".to_owned();
-        // FIXME(nox): Should this use base64::URL_SAFE?
-        // FIXME(nox): https://github.com/marshallpierce/rust-base64/pull/56
-        base64::engine::general_purpose::STANDARD.encode_string(&png, &mut url);
+        encoder.into_inner();
```
jdm, line 437: What's the purpose of this into_inner call if the return value is ignored?

EnnuiL: Its purpose is to poke the encoder into finishing encoding everything while at the same time dropping the `&mut url` reference that the encoder had borrowed; So yeah, removing this call isn't possible, and plus, it can't be directly given to `USVString` either without converting the reference into an owned value, which I believe would effectively duplicate the `url` variable?

Do note that the way the encoding is done matches the one described on `base64`'s documentation, so I believe this is how it's intended to be used

jdm (approved):

mrobinson:
Re-adding to merge queue due to #30111.

PR #30096, opened 2023-08-14
Combine all try workflows

mrobinson:
There are currently two ways to run try. One is to push to the `try` or
`try-*` branches and the other is to trigger a workflow via GitHub
comment. This change combines these methods into one workflow. In
addition, WPT results are reported together rather than separately and
filtered results for all WPT tests are bundled together in the same
artifact.

---

- [x] These changes do not require tests because they change GitHub workflows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ on:
     # Run the entire pipeline for 'master' even though the merge queue already runs checks
     # for every change. This just offers an extra layer of testing and covers the case of
     # random force pushes.
-    branches: ["master", "try"]
+    branches: ["master"]
   pull_request:
     types: ['opened', 'synchronize']
     branches: ["**"]
@@ -16,7 +16,10 @@ on:
       platform:
         required: true
         type: string
-      layout:
+      linux-wpt-layout:
+        required: true
+        type: string
+      mac-wpt-layout:
         required: true
         type: string
       unit-tests:
@@ -28,7 +31,11 @@ on:
         required: false
         type: choice
         options: ["none", "linux", "windows", "macos", "all", "sync"]
-      layout:
+      linux-wpt-layout:
+        required: false
+        type: choice
+        options: ["none", "2013", "2020", "all"]
+      mac-wpt-layout:
         required: false
         type: choice
         options: ["none", "2013", "2020", "all"]
@@ -48,11 +55,17 @@ jobs:
         uses: actions/github-script@v6
         with:
           script: |
-            // Never skip workflow runs for pull requests, merge groups, or manually triggered
-            // workfows / try jobs, which might need to actually run / retry WPT tests.
-            if (!['issue_comment', 'merge_group', 'pull_request', 'workflow_run', 'workflow_call'].includes(context.eventName)) {
-              // Skip the run if an identical run already exists. This helps to avoid running
-              // the workflow over and over again for the same commit hash.
+            // Skip this workflow if an identical run already exists. Note that we don't
+            // always want to skip workflows. In particular we don't skip duplicate
+            // workflows if:
+            //   1. This is a try job (triggered via an `issue_comment` event or a push to
+            //      a non-master branch)
+            //   2. This is a merge queue event (`merge_queue`)
+            //   3. This event was triggered by a pull request update (`pull_request`)
+            //   4. This event was triggered manually `workflow_run` / `workflow_call`
+            let isPushToMaster = context.eventName == "push" && process.env.GITHUB_REF_NAME == "master";
+            let isTryRun = context.eventName == 'issue_comment' || (context.eventName == 'push' && !isPushToMaster);
+            if (!isTryRun && !['merge_group', 'pull_request', 'workflow_run', 'workflow_call'].includes(context.eventName)) {
               if ((await github.rest.actions.listWorkflowRuns({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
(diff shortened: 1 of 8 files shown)


mrobinson:
cc @sagudev

sagudev (commented):
I love the possibilities this brings.

LGTM, although I am not very keen of reporting WPT results together:
1. If there are a lot of unexpected results we can hit message size limit even quicker (I did hit it once or twice before using single layout).
2. When layout2013 gets removed reporting WPT results would make less sense in main workflow

.github/workflows/main.yml
```diff
@@ -109,9 +120,54 @@ jobs:
...
+          done
+      - name: Archive aggregate results
+        uses: actions/upload-artifact@v3
+        with:
+          name: wpt-filtered-results
+          path: |
+            unexpected-test-wpt-*.log
+      - name: Comment on PR with results
+        run: |
+          etc/ci/report_aggregated_expected_results.py wpt-filtered-results-linux
```
sagudev, line 165: Would mac reporting still work?

mrobinson:
>     1. If there are a lot of unexpected results we can hit message size limit even quicker (I did hit it once or twice before using single layout).

I wonder if we could detect this situation and give a link to the full output?

>     2. When layout2013 gets removed reporting WPT results would make less sense in main workflow

This is true, but hopefully we will be adding Windows and Mac results as a regular feature at some point -- though this requires serious infrastructure investment and CI work.

sagudev:
> LGTM, although I am not very keen of reporting WPT results together:
> 
>     1. If there are a lot of unexpected results we can hit message size limit even quicker (I did hit it once or twice before using single layout).
> 
>     2. When layout2013 gets removed reporting WPT results would make less sense in main workflow

Silly me, I forget the main selling point:

3. Having things separated gives us results faster as we do not have to wait for both layout to finished testing (and the same would go for future OS split).

mrobinson:
>     3. Having things separated gives us results faster as we do not have to wait for both layout to finished testing (and the same would go for future OS split).

Yes, this is the main downside, but perhaps a future version of this script can create a comment and update it. Currently, tests will always complete at around the same time.

sagudev:
> Currently, tests will always complete at around the same time.

Maybe on servo org, but when running in personal forks with 20 concurrent job limit on free tier difference is much bigger as only one layout can be tested at a time.

mrobinson:
> > Currently, tests will always complete at around the same time.
> 
> Maybe on servo org, but when running in personal forks with 20 concurrent job limit on free tier difference is much bigger as only one layout can be tested at a time.

Can we do a second phase here where we turn this process into a reusable workflow that updates the GitHub check / comment? Combined output will be quite useful for upstream.

sagudev:
> Can we do a second phase here where we turn this process into a reusable workflow that updates the GitHub check / comment? Combined output will be quite useful for upstream.

Making wpt reporting into reusable workflow sounds good. Although I am not sure if we can mutate GitHub check.

.github/workflows/main.yml
```diff
@@ -50,7 +56,9 @@ jobs:
           script: |
             // Never skip workflow runs for pull requests, merge groups, or manually triggered
             // workfows / try jobs, which might need to actually run / retry WPT tests.
-            if (!['issue_comment', 'merge_group', 'pull_request', 'workflow_run', 'workflow_call'].includes(context.eventName)) {
+            let isPushToMaster = context.eventName == "push" && process.env.GITHUB_REF_NAME == "master";
+            let isTryRun = context.eventName == 'issue_comment' || (context.eventName == 'push' && !isPushToMaster);
+            if (!isTryRun && !['merge_group', 'pull_request', 'workflow_run', 'workflow_call'].includes(context.eventName)) {
               // Skip the run if an identical run already exists. This helps to avoid running
               // the workflow over and over again for the same commit hash.
               if ((await github.rest.actions.listWorkflowRuns({
```
mukilan, line 64: I'm having a hard time figuring out when this if arm will be taken. Which events are being considered for de-deduplication here?

.github/workflows/main.yml
```diff
@@ -109,9 +121,55 @@ jobs:
...
+  report-test-results:
+    name: Report WPT Results
+    runs-on: ubuntu-latest
+    needs:
+      - "decision"
+      - "build-win"
+      - "build-mac"
+      - "build-linux"
+    if: ${{
+      always() && !cancelled() &&
```
mukilan, line 136: is `always()` needed here?

.github/workflows/main.yml
mrobinson: Sure I can add a comment here. How does something like this sound?

```
            // Skip this workflow if an identical run already exists. Note that we don't always want to skip workflows. In
            // particular we don't skip duplicate workflows if:
            //   1. This is a try job (triggered via an `issue_comment` event or a push to a non-master branch)
            //   2. This is a merge queue event (`merge_queue`)
            //   3. This event was triggered by a pull request update (`pull_request`)
            //   4. This event was triggered manually `workflow_run` / `workflow_call`
            let isPushToMaster = context.eventName == "push" && process.env.GITHUB_REF_NAME == "master";
            let isTryRun = context.eventName == 'issue_comment' || (context.eventName == 'push' && !isPushToMaster);
            if (!isTryRun && !['merge_group', 'pull_request', 'workflow_run', 'workflow_call'].includes(context.eventName)) {
```

Honestly, we can likely just remove this code which prevents duplicate workflow runs now. It was more important when bors was triggering try runs for commits that we were not testing.

.github/workflows/main.yml
mrobinson: Yes, I think it is. Without always, I think that dependent jobs failing or being cancelled can cause this job not to run. It's a weird quirk of GitHub Actions.

.github/workflows/main.yml
mukilan: > Honestly, we can likely just remove this code which prevents duplicate workflow runs now. It was more important when bors was triggering try runs for commits that we were not testing.

That was my suspicion as well, but I wasn't able to follow the negated clauses to figure out which events do actually trigger the skip-workflow logic here. I'm fine with removing the logic if you think that is safe to do so.

For the comment, the new comment is similar to the existing one above the if statement.  I guess I was looking an example of an event where the skip-workflow logic is still executed.

mrobinson:
I've gone ahead and updated the branch wrt the comments above.

.github/workflows/main.yml
mukilan: Cool. Thanks for the background!

mukilan (approved):

.github/workflows/main.yml
mrobinson: That's a good point. I believe that this mainly skips jobs for pushes to master. I agree this is a bit of a weird edge case and confusing. I'll will post a followup that just removes this for now.

mrobinson:
Thanks for the review.

PR #29848, opened 2023-06-07
Backport several style changes from Gecko (4)

Loirooriol:
This continues https://github.com/servo/servo/pull/29816.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #30099, opened 2023-08-15
Backport several style changes from Gecko (5)

Loirooriol:
This continues https://github.com/servo/servo/pull/29848.

---


components/script/dom/element.rs
```diff
@@ -3307,6 +3303,16 @@ impl<'a> SelectorsElement for DomRoot<Element> {
     fn is_html_slot_element(&self) -> bool {
         self.is_html_element() && self.local_name() == &local_name!("slot")
     }
+
+    fn set_selector_flags(&self, flags: ElementSelectorFlags) {
+        use std::ops::Deref;
```
mrobinson, line 3308: It makes more sense to have to this at the top-level of the file, I think.

components/selectors/matching.rs
```diff
@@ -328,8 +328,7 @@ where
         debug_assert_eq!(next_sequence, Combinator::PseudoElement);
     }
 
-    let result =
-        matches_complex_selector_internal(iter, element, context, Rightmost::Yes);
+    let result = matches_complex_selector_internal(iter, element, context, Rightmost::Yes);
```
mrobinson, line 331: Instead of fixing the style in the selectors crate, I think we should add an exception and fix the style upstream.

components/style/macros.rs
```diff
@@ -128,11 +128,3 @@ macro_rules! local_name {
...
     };
 }
-
-/// Asserts the size of a type at compile time.
-macro_rules! size_of_test {
-    ($t: ty, $expected_size: expr) => {
-        #[cfg(target_pointer_width = "64")]
-        const_assert_eq!(std::mem::size_of::<$t>(), $expected_size);
-    };
-}
```
mrobinson: Let's keep this one around to avoid differing too much from upstream.

servo-tidy.toml
```diff
@@ -88,20 +88,27 @@ files = [
   "./components/style/counter_style/mod.rs",
   "./components/style/properties/declaration_block.rs",
   "./components/style/properties/helpers.mako.rs",
+  "./components/style/queries/feature_expression.rs",
```
mrobinson, line 91: At this point, it might make more sense to simply ignore all of `components/style` and the rest of the crates that we vendor from Gecko.

mrobinson (approved):
Looks good. Just have some minor nits.

components/style/macros.rs
Loirooriol: OK, so I will keep this in the style crate, also add a copy into its own crate to be used by layout, msg, selectors and style_traits, and then move it upstream.

Loirooriol:
These are the additions that I had to add to compile: a1c8341ee, a9afa00f8, 01e1801d2, 312ace68a, e2dc8c83a, 81a5f3412, 913edb302, 6faef7f41, a1bb844e5, 90bc74de1, 8325a5632, 3edc2959a, 42d82e846, 2d40b8ffd, 938483142, 01426783b, b9e40d780, e3e51239d, 4cfcfffd2, 2c7418498.

For 42d82e846 I don't have much idea if this unsafe code is actually safe or makes sense:
```rs
    fn set_selector_flags(&self, flags: ElementSelectorFlags) {
        #[allow(unsafe_code)]
        unsafe {
            Dom::from_ref(self.deref())
                .to_layout()
                .insert_selector_flags(flags);
        }
    }
```

but it can always be addressed in a follow-up if it's problematic.

Loirooriol:
Ah, the commits got merged together, which will be bad for bisects. I will revert and reland as a rebase

PR #30100, opened 2023-08-15
Remove gold config option and use lld on Linux

mrobinson:
There are a few motivations for this change:

 1. lld is demonstrably faster than gold, but is really only stable on
    Linux at the moment. There's a good chance that it will be ready for
    all platforms soon though.
 2. Most people do not have gold installed on MacOS and Windows. You'd
    have to do this manually through homebrew. I think it's a safe
    assumption that this probably won't be slowing things down much on
    those platforms.
 3. We need to remove all configuration of the build that happens while
    running `./mach build` if we ever hope to make `cargo build`
    equivalent to the mach build. This unlocks static configuration of
    the rustflags. One of the big blockers for proper `cargo build`
    support.

---

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -509,9 +509,11 @@ class CommandBase(object):
         if self.config["build"]["rustflags"]:
             env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " " + self.config["build"]["rustflags"]
 
-        if self.config["tools"]["rustc-with-gold"]:
-            if shutil.which('ld.gold'):
-                env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " -C link-args=-fuse-ld=gold"
+        # Turn on rust's version of lld if we are on x86 Linux.
+        # TODO(mrobinson): Gradually turn this on for more platforms, when support stabilizes.
+        # See https://github.com/rust-lang/rust/issues/39915
+        if not self.cross_compile_target and effective_target == "x86_64-unknown-linux-gnu":
+            env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " -Zgcc-ld=lld"
 
         if not (self.config["build"]["ccache"] == ""):
             env['CCACHE'] = self.config["build"]["ccache"]
--- a/servobuild.example
+++ b/servobuild.example
@@ -4,9 +4,6 @@
 
 # Tool options
 [tools]
-# If rustc-with-gold is true, will try to find and use gold linker with rustc.
-# Defaults to true
-rustc-with-gold = true
 
 # If uncommented, this command is used instead of the platform’s default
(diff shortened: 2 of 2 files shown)


sagudev:
IIRC official Firefox windows build are made using lld.

Don't we use lld on windows? https://github.com/servo/servo/blob/a7bd9f0d43f4de13c958eaea53709e8fc0d162c8/.cargo/config#L25-L26

mrobinson:
> Don't we use lld on windows?
> 
> https://github.com/servo/servo/blob/a7bd9f0d43f4de13c958eaea53709e8fc0d162c8/.cargo/config#L25-L26

Oh, great point. The above option will try to use rust's internal version of lld, so I think it makes sense to make it Linux only for now. Hopefully the second step of this is to move all rustflags to .cargo/config.

python/servo/command_base.py
```diff
@@ -531,6 +533,7 @@ def build_env(self, is_build=False):
         # Work around https://github.com/servo/servo/issues/24446
         # Argument-less str.split normalizes leading, trailing, and double spaces
         env['RUSTFLAGS'] = " ".join(env['RUSTFLAGS'].split())
+        print(env['RUSTFLAGS'])
```
mukilan, line 536: Was this a temporary print statement added for debugging? If not, can we print this with a suffix like `RUSTFLAGS=`

python/servo/command_base.py
```diff
@@ -509,9 +509,11 @@ def build_env(self, is_build=False):
...
             env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " " + self.config["build"]["rustflags"]
 
-        if self.config["tools"]["rustc-with-gold"]:
-            if shutil.which('ld.gold'):
-                env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " -C link-args=-fuse-ld=gold"
+        # Turn on rust's version of lld if we are on x86 Linux.
+        # TODO(mrobinson): Gradually turn this on for more platforms, when support stabilizes.
+        # See https://github.com/rust-lang/rust/issues/39915
+        if not self.cross_compile_target and effective_target == "x86_64-unknown-linux-gnu":
+            env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " -Zgcc-ld=lld"
```
mukilan, line 516: What is the difference betweenthe unstable `-Zgcc-ld` flag here and the original  `-C link-args=-fuse-ld=lld`?

python/servo/command_base.py
mrobinson: :innocent: 

Yeah, this was left over. I've removed it now.

mukilan (approved):

python/servo/command_base.py
mukilan: Thanks for the explanation!

Issue #24607, opened 2019-10-31
Remove use of wake_after_send in webgl code


jdm:
#24482 removes the use of main-thread webgl. The wake_after_send helpers in webglrenderingcontext.rs were added to ensure the main thread wakes up and processes webgl messages, but that is no longer necessary.

atbrakhi:
fixed in #30101

PR #30101, opened 2023-08-15
Remove `EventLoopWaker` from Constellation

### LINKED ISSUE #24609: Clean up event loop waker in Constellation that is always None
#24482 removes the need for a non-None waker, since it removes the webgl main thread support that required it.

atbrakhi:
It's always `None`.

- [x] These changes fix #24609  #24607 
- [x] These changes do not require tests because they remove dead code.

### DIFF
--- a/components/script/dom/webglrenderingcontext.rs
+++ b/components/script/dom/webglrenderingcontext.rs
@@ -63,7 +63,6 @@ use canvas_traits::webgl::{
     YAxisTreatment,
 };
 use dom_struct::dom_struct;
-use embedder_traits::EventLoopWaker;
 use euclid::default::{Point2D, Rect, Size2D};
 use ipc_channel::ipc::{self, IpcSharedMemory};
 use js::jsapi::{JSContext, JSObject, Type};
@@ -236,7 +235,7 @@ impl WebGLRenderingContext {
             let max_vertex_attribs = ctx_data.limits.max_vertex_attribs as usize;
             Self {
                 reflector_: Reflector::new(),
-                webgl_sender: WebGLMessageSender::new(ctx_data.sender, None),
+                webgl_sender: WebGLMessageSender::new(ctx_data.sender),
                 webrender_image: ctx_data.image_key,
                 webgl_version,
                 glsl_version: ctx_data.glsl_version,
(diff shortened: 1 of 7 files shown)


mrobinson (approved):
Great work.

components/script/script_thread.rs
```diff
@@ -1416,7 +1416,7 @@ impl ScriptThread {
             replace_surrogates: opts.debug.replace_surrogates,
             user_agent,
             player_context: state.player_context,
-            event_loop_waker: state.event_loop_waker,
```
mrobinson, line 1419: Oh, one question here. Is this ever not-None?

atbrakhi: It shouldn't be not-None. In ScriptThreadFactory when we create a ScriptThread the event_loop_waker is already None(in pipeline.rs `STF::create()`). And I also don't see this function being called elsewhere.

mrobinson: Could you also remove the `event_loop_waker` member from `ScriptThread` in that case as well?

atbrakhi: Removed in [339aed9](https://github.com/servo/servo/pull/30101/commits/339aed92cb6fa15a59bae9d47e6c32d53f95bc06), thanks.

components/script/script_thread.rs
```diff
@@ -3333,7 +3329,7 @@ impl ScriptThread {
             self.replace_surrogates,
             self.user_agent.clone(),
             self.player_context.clone(),
-            self.event_loop_waker.as_ref().map(|w| (*w).clone_box()),
+            None,
```
mrobinson, line 3332: Hrm. ☺

If we are always passing `None` here can we also remove this argument...or is it ever non-`None`?

atbrakhi: hmm, here it is always None and Window::new() is being called just in script_thread. 

To remove `event_loop_waker` from `Window`, I need to be a bit careful, as there are a few pub functions that are using Window's event_loop_waker and are being called in different places especially in webgl code that i am not familiar with(such as `webgl_chan` and `get_event_loop_waker`).

So far I haven't encountered a single palce where `get_event_loop_waker()` is not None. I will keep us updated here.

components/script/dom/webglrenderingcontext.rs
```diff
@@ -236,10 +236,7 @@ impl WebGLRenderingContext {
             let max_vertex_attribs = ctx_data.limits.max_vertex_attribs as usize;
             Self {
                 reflector_: Reflector::new(),
-                webgl_sender: WebGLMessageSender::new(
-                    ctx_data.sender,
-                    window.get_event_loop_waker(),
-                ),
+                webgl_sender: WebGLMessageSender::new(ctx_data.sender, None),
```
mrobinson, line 239: Sorry for this constant back and forth, but I think we need to chase this None all the way. Can you replace the `webgl_sender: WebGLMessageSender` with a simple `webgl_sender: WebGLChan` and remove all of the code around waking up the event loop, since it seems to be completely unused?

r=me with that change. Sorry again!

atbrakhi: oh no worries. I am aware. Actually I was writing, why I left it as None. See my comment here: https://github.com/servo/servo/pull/30101#discussion_r1294723047

Please let me know if you would prefer I do that changes in this same PR?

mrobinson (approved):

mrobinson:
Thanks!

PR #23163, opened 2019-04-04
Upgrade to SpiderMonkey 66

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/23163)

PR #30102, opened 2023-08-15
Replace "-mmacosx-version-min" with MACOSX_DEPLOYMENT_TARGET

mrobinson:
It's unclear if the compiler flag was doing anything, but I've verified
(with otool) that the environment variable does affect the minimum
version of the MacOS set in the binary. We could examine later if this
is still necessary.

This was added in #23163 when switching CI from gcc to clang.

---

- [x] These changes do not require tests because they just adjust build configuration.


mukilan (approved):

mrobinson:
@mukilan Thanks for the review!

PR #30105, opened 2023-08-16
Cmd or Ctrl+L should focus on location bar in minibrowser mode

atbrakhi:
Adding keyboard shortcut cmd+L & ctrl+L that focus the location bar when in minibrowser mode

- `Modifiers::COMMAND` of egui is : Cmd key on Mac, elsewhere: Ctrl key.
- Added a condition around servo's `CMD_OR_CTRL+L` shortcut, as we still want that to work when not in minibrowser mode.

- [x] These changes are part of #30049

### DIFF
--- a/ports/winit/minibrowser.rs
+++ b/ports/winit/minibrowser.rs
@@ -4,7 +4,7 @@
 
  use std::{cell::{RefCell, Cell}, sync::Arc};
 
-use egui::TopBottomPanel;
+use egui::{TopBottomPanel, Modifiers, Key};
 use servo::{servo_url::ServoUrl, compositing::windowing::EmbedderEvent};
 use servo::webrender_surfman::WebrenderSurfman;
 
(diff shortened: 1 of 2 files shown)


ports/winit/minibrowser.rs
```diff
@@ -51,16 +51,21 @@ impl Minibrowser {
                     ui.available_size(),
                     egui::Layout::right_to_left(egui::Align::Center),
                     |ui| {
+                        let response = ui.add_sized(
+                            ui.available_size(),
+                            egui::TextEdit::singleline(&mut *location.borrow_mut()),
+                        );
```
delan, line 57: can we rename this to something like “location_field”, to make it clear which widget we’re later checking and focusing? also to keep the go button on the right, we need to call this after the ui.button("go") block.

delan (changes requested):

ports/winit/minibrowser.rs
atbrakhi: Fixed in [3b4c213](https://github.com/servo/servo/pull/30105/commits/3b4c213364d7474377fdeaf3707c87ed79574e04)

delan (approved):
Great work!

PR #30112, opened 2023-08-17
Format toml files

sagudev:
Formatted with [taplo](https://taplo.tamasfe.dev/) which is also used in Even Better TOML vscode extension.

---


rust-toolchain.toml
```diff
@@ -2,10 +2,10 @@
 channel = "nightly-2023-02-01"
 components = [
     # https://github.com/rust-lang/rust/issues/72594#issuecomment-633779564
-    "llvm-tools-preview",
+    "llvm-tools",
     # For components/script_plugins, https://github.com/rust-lang/rust/pull/67469
     "rustc-dev",
     "rust-docs",
-    "rustfmt-preview",
+    "rustfmt",
```
sagudev, line 9: This wasn't from formatted but resolves warnings from extension.

mrobinson:
Some of these changes seem great such as normalizing spaces after "{". Others are a bit questionable, like limiting the line length which has the potential to introduce a lot of churn. Would it be possible to regenerate this formatting with the following changes?

 1. With no line length limit
 2. Not trying to offset or line up comments in the file
 3. Checking in the configuration file for the tool so that the results can be reproduced

rust-toolchain.toml
mrobinson: Hrm. I'm a bit nervous about changing these. Why would formatting TOML files require us to use different rustup components?

mrobinson (changes requested):
Great stuff. The only thing that I'm confused about is the change to the rustup components. I think we can omit that here.

rust-toolchain.toml
sagudev: When components where first introduced the got -preview prefix, but now they aren't using prefixes anymore as they are stable. Aliases are still in cargo to not make be breaking change.

mrobinson (approved):

rust-toolchain.toml
mrobinson: Thanks for the background. I'd be happy to review that as a separate PR.

PR #30123, opened 2023-08-18
Use system lld on NixOS instead of rust-lld

mukilan:
The -Zgcc-ld=lld flag makes rust use the rust-lld
linker that is distributed as part of rust toolchain. 
However, this flag doesn't work on nixos correctly as
  1) rust-lld needs to be patched to have the correct rpath
     to find libz.so
  2) the bin/gcc-ld/ld.lld wrapper which calls rust-lld also
     needs to be patched to use the correct dynamic loader
  3) rust-lld doesn't respect NIX_LDFLAGS which contains
     the additional search path derived from buildInputs.
     The system linkers on nixos are wrapped so that
     NIX_LDFLAGS is added as the rpath to the final binary.

---

- [x] These changes do not require tests because they fix compilation issues.

### DIFF
--- a/python/servo/platform/linux.py
+++ b/python/servo/platform/linux.py
@@ -127,6 +127,17 @@ class Linux(Base):
         installed_something |= self._platform_bootstrap_gstreamer(force)
         return installed_something
 
+    def linker_flag(self) -> str:
+        # the rust-lld binary downloaded by rustup
+        # doesn't respect NIX_LDFLAGS and also needs
+        # other patches to work correctly. Use system
+        # version of lld for now. See
+        # https://github.com/NixOS/nixpkgs/issues/220717
+        if self.distro.lower() == 'nixos':
+            return '-C link-arg=-fuse-ld=lld'
+        else:
+            return '-Zgcc-ld=lld'
+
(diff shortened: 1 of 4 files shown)


python/servo/platform/linux.py
```diff
@@ -127,6 +127,16 @@ def _platform_bootstrap(self, force: bool) -> bool:
         installed_something |= self._platform_bootstrap_gstreamer(force)
         return installed_something
 
+    def linker_flag(self) -> str:
+        # the rust-lld binary downloaded by rustup
+        # doesn't respect NIX_LDFLAGS and also needs
+        # other patches to work correctly. Use system
+        # version of lld for now
```
delan, line 134: consider mentioning NixOS/nixpkgs#220717 and filing a NixOS or rustup issue about programs linked with rustup’s lld?

delan (approved):
Thanks! I can confirm this fixes the breakage for me.

python/servo/platform/linux.py
mukilan: Good idea! I've added link to the issue in 4588f8a . I'll create a issue in NixOS for the LDFLAGS stuff.

mukilan:
Thanks for the review!

PR #30125, opened 2023-08-18
[NFC] compositing: extract types into new compositing_traits crate

delan:
This patch extracts the following compositing types into a new compositing_traits crate, eliminating the dependency on compositing in the constellation and webdriver_server components.

* CompositorProxy
* Msg and its inner types:
    * SendableFrameTree
    * CompositionPipeline
    * WebrenderMsg
    * WebrenderFontMsg
    * WebrenderCanvasMsg

We also move the remaining types in compositing’s compositing_thread.rs to lib.rs. That module was confusingly named, because there isn’t really a compositing thread, since the compositor runs in the embedder thread.

---


components/compositing/Cargo.toml
```diff
@@ -13,10 +13,11 @@ path = "lib.rs"
 
 [features]
 default = []
-gl = ["gleam", "pixels"]
```
mrobinson, line 16: I think you can also revert this change now too.

components/compositing/lib.rs
```diff
@@ -7,145 +7,43 @@
...
-    CompositorEvent, ConstellationControlMsg, LayoutControlMsg, MediaSessionActionType,
-};
-use servo_url::ServoUrl;
-use std::collections::HashMap;
-use std::fmt;
-use std::time::Duration;
+
+use compositing_traits::CompositorProxy;
+use compositing_traits::CompositorReceiver;
+use compositing_traits::ConstellationMsg;
```
mrobinson, line 15: Since these lines are changing, do you mind using the more modern style of having these on the same line?

components/compositing_traits/lib.rs
```diff
@@ -16,6 +21,8 @@ use msg::constellation_msg::{PipelineId, TopLevelBrowsingContextId};
 use net_traits::image::base::Image;
 use profile_traits::mem;
 use profile_traits::time;
+use script_traits::ConstellationControlMsg;
+use script_traits::LayoutControlMsg;
 use script_traits::{AnimationState, EventResult, MouseButton, MouseEventType};
```
mrobinson, line 26: Since this file is new, these could all be collected on the the same line as well. We are gradually moving to this style, I think. Eventually we should probably configure rustfmt to enforce it.

components/constellation/constellation.rs
```diff
@@ -106,10 +106,10 @@ use bluetooth_traits::BluetoothRequest;
...
 use canvas_traits::ConstellationCanvasMsg;
-use compositing::compositor_thread::CompositorProxy;
-use compositing::compositor_thread::Msg as ToCompositorMsg;
-use compositing::compositor_thread::WebrenderMsg;
-use compositing::{ConstellationMsg as FromCompositorMsg, SendableFrameTree};
+use compositing_traits::CompositorProxy;
+use compositing_traits::Msg as ToCompositorMsg;
+use compositing_traits::WebrenderMsg;
+use compositing_traits::{ConstellationMsg as FromCompositorMsg, SendableFrameTree};
 use crossbeam_channel::{after, never, unbounded, Receiver, Sender};
```
mrobinson, line 113: Ditto here.

mrobinson (changes requested):
Looks great! Just a couple suggestions below and in addition:

1. Can you rename `Msg` to `CompositorMsg` as we discussed? The name `Msg` has always confused the heck out of me.
2. Since the whole file is moving, lets remove the reference to the "painting thread" (RIP).
3. Is this the right time to do the other change we talked about?
  - Rename `WebrenderMsg` to `ForwardedMsg`
  - Rename all of the types wrapped in that enum in a consistent way such as:
     - `WebrenderFontMsg` → `FontCacheToCompositorMsg`.
     - `WebrenderCanvasMsg` → `CanvasToCompositorMsg`.
   - Move `script_traits::WebrenderMsg` and `net_traits::WebrenderImageMsg` to the new `compositor_traits` crate and also name them appropriately.

components/compositing/Cargo.toml
delan: done in 36dd2a70836b5b7253c7cb782ca41a755fcf7abf

components/compositing/lib.rs
delan: done in 96eec79caf5fb6dbfab8adaea382b78ca2bb8076

components/compositing_traits/lib.rs
delan: done in 96eec79caf5fb6dbfab8adaea382b78ca2bb8076

components/constellation/constellation.rs
delan: done in 96eec79caf5fb6dbfab8adaea382b78ca2bb8076

delan:
Thanks! I was originally planning on doing all of those in a separate patch, but I’ve done the first two in c08d04d3a4f719fde3f35ff5d56f70e0e1266960.

Let’s do the third part separately I think :)

mrobinson:
> Thanks! I was originally planning on doing all of those in a separate patch, but I’ve done the first two in [c08d04d](https://github.com/servo/servo/commit/c08d04d3a4f719fde3f35ff5d56f70e0e1266960).

That sounds fine to me. Thanks for clarifying.

mrobinson (approved):

mrobinson:
Feel free to toss this onto the 'ol merge queue whenever you are ready.

mrobinson:
The first followup for this is here: #30496.

PR #30126, opened 2023-08-18
Make return key equivalent to go button when location field is focused

atbrakhi:
Adds Return key that is equivalent to go button when location field is focused in Minibrowser mode.

- [x] These changes part of #30049


delan:
This is a good start, though there are a couple parts of the behaviour I think we can improve:

* the location sometimes reverts to the old value while loading, which may be confusing
* we should also accept numpad enter, not just the normal return key

atbrakhi:
>    the location sometimes reverts to the old value while loading, which may be confusing

I will look into this part separately, as it is not the result of changes in this PR. I have added it as a TODO in https://github.com/servo/servo/issues/30049

>    we should also accept numpad enter, not just the normal return key

unfortunately egui does not have support for [numpad enter](https://github.com/emilk/egui/issues/1987).  [Docs here](https://docs.rs/egui/latest/egui/enum.Key.html)  :(
I will add it as TODO in meta issue, so that we can check back later.

delan (approved):
Thanks, and good catch, I didn’t realise the location bug was pre-existing.

PR #30128, opened 2023-08-18
Enforce formatting of TOML files

### LINKED ISSUE #26728: Enforcing formatting of TOML files
Currently having consistent formatting is not enforced for TOML files and as a result they have tended to become a bit disorganized. I have manually cleaned up a number of them recently to have mostly consistent formatting but things still aren't quite perfect. Would there be any interest in enforcing formatting standards for these files or would that just be more of a nuisance?

sagudev:
Formatting is enforced by CI rather than requiring users to install taplo as I suggested in https://github.com/servo/servo/issues/26728#issuecomment-1682709808

---

- [x] These changes fix #26728

- [x] These changes do not require tests because they are helper scripts


mrobinson (changes requested):
I think this will make things tricky for developers because bootstrap isn't installing taplo, so the changes will fail on CI but not locally with test-tidy. Can we install a toml formater via pip or cargo dependencies that does the right thing?

python/servo/testing_commands.py
```diff
@@ -278,7 +278,14 @@ def test_tidy(self, all_files, no_progress, stylo, no_wpt=False):
         if rustfmt_failed:
             print("Run `./mach fmt` to fix the formatting")
 
-        return tidy_failed or manifest_dirty or rustfmt_failed
+        taplo_failed = 0
+        taplo = shutil.which("taplo")
+        if (taplo is not None):
+            taplo_failed = call([taplo, "fmt", "--check"], env={'RUST_LOG': 'error'})
+        else:
+            print("Warning: taplo is not installed. TOML formatting check skipped.")
```
sagudev, line 286: Actually we only issue a warning if taplo is not installed.

sagudev:
> Can we install a toml formater via pip or cargo dependencies that does the right thing

Is there any place where I can put `cargo install taplo-cli --locked` without doing separately for every platform?

mrobinson:
> > Can we install a toml formater via pip or cargo dependencies that does the right thing
> 
> Is there any place where I can put `cargo install taplo-cli --locked` without doing separately for every platform?

Could this be made a dev dependency of the winit port?

sagudev:
> > > Can we install a toml formater via pip or cargo dependencies that does the right thing
> > 
> > 
> > Is there any place where I can put `cargo install taplo-cli --locked` without doing separately for every platform?
> 
> Could this be made a dev dependency of the winit port?

Could we run cli from dev-dependencies? I think that is not possible: https://github.com/rust-lang/cargo/issues/2267

mrobinson:
It seems a tool like [toml-sort](https://pypi.org/project/toml-sort/) could be installed by pip and then run unconditionally. What do you think?

sagudev:
> It seems a tool like [toml-sort](https://pypi.org/project/toml-sort/) could be installed by pip and then run unconditionally. What do you think?

Hm, it certainly is more convenient to install via mach. I will see if I can make config for it.

mrobinson:
Great. That sounds like a plan. I can help out with the integration into "./mach fmt" as well.

sagudev:
After reading [toml-sort](https://pypi.org/project/toml-sort/) usage manual I am not so sure it was meant to be used this way. In taplo you can just do `taplo fmt` and it would automatically search for config and formatted all files. In toml-sort you need to do `toml-sort --in-place **.toml` if you have bash glob enabled. There is also no way to really exclude some toml files. While we could write wrappers in mach that would do all this stuff, I do not think that is the way to go.

I think Taplo is more oriented towards users like us, that want to enforce formatting in their project and is easier to be used as standalone, while toml-sort is just utility that works for files being piped rather than on projects.

mrobinson:
Hrm. Maybe we can build taplo into the target directory or install it for the user with cargo during bootstrapping?

sagudev:
I am working on user  installation with cargo during bootstrapping. 

Wow how `mach` has changed since the last time I touched it (well bootstrapping to be exact).

.cargo/config.toml
```diff
@@ -41,4 +41,4 @@ linker = "lld-link.exe"
 MACOSX_DEPLOYMENT_TARGET = "10.10"
 
 [build]
-rustdocflags = ["--document-private-items"]
```
mrobinson, line 44: I think this change was unintentional.

python/servo/platform/base.py
```diff
@@ -99,6 +100,17 @@ def bootstrap(self, force: bool):
         if not self._platform_bootstrap(force):
             print("Dependencies were already installed!")
 
+    def taplo_install(self, force: bool) -> bool:
+        if not force:
+            if shutil.which("taplo") is not None:
+                return False
+
```
mrobinson, line 107:
```suggestion
    def install_taplo(self, force: bool) -> bool:
        if not force and  shutil.which("taplo"):
            return False

```

python/servo/platform/macos.py
```diff
@@ -54,8 +54,10 @@ def is_gstreamer_installed(self, cross_compilation_target: Optional[str]) -> boo
             return False
         return True
 
-    def _platform_bootstrap(self, _force: bool) -> bool:
-        return self._platform_bootstrap_gstreamer(False)
+    def _platform_bootstrap(self, force: bool) -> bool:
+        installed_something = self._platform_bootstrap_gstreamer(False)
+        installed_something |= self.taplo_install(force)
+        return installed_something
```
mrobinson, line 60: Instead of explicitly calling `taplo_install(...)` here, let's call `taplo_install(...)` in the version of `bootstrap()` in `base.py`.

python/servo/testing_commands.py
```diff
@@ -57,6 +57,22 @@
 TEST_SUITES_BY_PREFIX = {path: k for k, v in TEST_SUITES.items() if "paths" in v for path in v["paths"]}
 
 
+def taplo(write: bool = False) -> int:
```
mrobinson, line 60: A person reading this file might not know what taplo is, so I would suggest giving this function a more descriptive name. How about `format_toml_files()`? In addition I would name the argument `check` instead of write and reverse its meaning -- just to match the argument we pass to taplo.

python/servo/testing_commands.py
```diff
@@ -57,6 +57,22 @@
...
+    taplo = shutil.which("taplo")
+    if (taplo is not None):
```
mrobinson, line 62: I would use an early return here.

```python
if not shutil.which("taplo"):
    print("Taplo is not installed.")
    print("It should be installed using `./mach bootstrap`, \
              but it can be installed manually using `cargo install taplo-cli --locked`")
    return 1
```

python/servo/testing_commands.py
```diff
@@ -370,9 +388,12 @@ def update_manifest(self, **kwargs):
         return wpt.manifestupdate.update(check_clean=False)
 
     @Command('fmt',
-             description='Format the Rust and CPP source files with rustfmt',
+             description='Format the Rust and TOML files',
```
mrobinson, line 391:
```suggestion
             description='Format Rust and TOML files',
```

python/servo/testing_commands.py
```diff
@@ -370,9 +388,12 @@ def update_manifest(self, **kwargs):
...
              category='testing')
     def format_code(self):
+        taplo_res = taplo(True)
```
mrobinson, line 394:
```suggestion
        result = format_toml_files(check=False)
```

mrobinson (changes requested):
Thanks so much for iterating on this. This is looking really good. I have a suggestion for a few minor changes.

.cargo/config.toml
sagudev: No config.toml was not formatted, so I needed to format it for CI to work.

.github/workflows/linux.yml
```diff
@@ -71,6 +71,11 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.9'
+      - name: Install taplo
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: taplo-cli
+          locked: true
```
mrobinson, line 78: Will this be installed by `./mach bootstrap` below now?

.github/workflows/mac.yml
```diff
@@ -60,6 +60,11 @@ jobs:
           fetch-depth: 2
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.3
+      - name: Install taplo
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: taplo-cli
+          locked: true
```
mrobinson, line 67: Ditto.

.github/workflows/windows.yml
```diff
@@ -61,6 +61,11 @@ jobs:
           echo "C:\\wix" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - name: ccache
         uses: hendrikmuhs/ccache-action@v1.2
+      - name: Install taplo
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: taplo-cli
+          locked: true
```
mrobinson, line 68: Obviously it won't be installed here because bootstrap isn't run for Windows.

python/servo/platform/base.py
```diff
@@ -96,9 +97,21 @@ def is_gstreamer_installed(self, cross_compilation_target: Optional[str]) -> boo
         )
 
     def bootstrap(self, force: bool):
-        if not self._platform_bootstrap(force):
+        installed_something = self._platform_bootstrap(force)
+        installed_something |= self.taplo_install(force)
+        if not installed_something:
             print("Dependencies were already installed!")
 
+    def taplo_install(self, force: bool) -> bool:
```
mrobinson, line 105: Let's call this `install_taplo` so that it's a verb phrase.

mrobinson (approved):
Looks good! Just a couple little nits before landing. Thanks!

.github/workflows/linux.yml
sagudev: It would be, but this way it's faster as this action will cache it. Using cargo install takes 4-5min.

.github/workflows/windows.yml
sagudev: > bootstrap isn't run for Windows

That's a surprise.

mrobinson (approved):
Thank you!

PR #30130, opened 2023-08-18
Layout floats as children of their inline ancestors

mrobinson:
When layout was split into two phases, floats were laid out as direct
children of the inline formatting context. This meant that they were
positioned properly, but not properly made children of their inline
ancestors' stacking contexts. This change maintains the proper
positioning of floats, but positions them relatively to their inline
ancestors.

The big change here is that `text-align` needs to be taken into account
before actually laying out LineItems. This has the added benefit of
setting inline layout for the implementation of `text-align: full`. Now
all line items are laid out at the real final position and we can adjust
the `start_corner` property of float `BoxFragments` when their ancestors
are laid out.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -121,7 +121,7 @@ struct InlineFormattingContextState<'box_tree, 'a, 'b> {
     fragments: Vec<Fragment>,
 
     /// The position of where the next line will start.
-    next_line_start_position: Vec2<Length>,
+    current_line_start_position: Vec2<Length>,
 
     /// The current inline position in the line being laid out into [`LineItems`] in this
     /// [`InlineFormattingContext`] independent of the depth in the nesting level.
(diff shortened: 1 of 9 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -174,29 +174,78 @@ impl<'box_tree, 'a, 'b> InlineFormattingContextState<'box_tree, 'a, 'b> {
...
+            self.fragments
+                .push(Fragment::Anonymous(AnonymousFragment::new(
+                    Rect { start_corner, size },
+                    fragments,
+                    containing_block.style.writing_mode,
+                )));
+        }
+    }
+
+    /// Given the amount of whitespace trimmed from the start of the line and taking into consideration
```
mukilan, line 242: The comment here says 'amount of whitespace trimmed from the *start of the line*' but in code, the value for whitespace_trimmed is set by the 'trim_whitespace_at_end` method. Is the comment here wrong or am I misunderstanding it?

components/layout_2020/flow/inline.rs
```diff
@@ -235,48 +284,21 @@ impl<'box_tree, 'a, 'b> InlineFormattingContextState<'box_tree, 'a, 'b> {
...
-                    Rect { start_corner, size },
-                    fragments,
-                    containing_block.style.writing_mode,
-                )));
+        let line_start_position_from_indent = self.current_line_start_position.inline;
+        let real_line_length_without_indent = self.current_inline_position - whitespace_trimmed;
+        match text_align {
+            TextAlign::Start => line_start_position_from_indent,
+            TextAlign::End => {
+                containing_block.inline_size - real_line_length_without_indent +
```
mukilan, line 293: Is the calculation here correct?
For [this test html](http://software.hixie.ch/utilities/js/live-dom-viewer/saved/11941), both master and legacy-layout have the same output as Firefox and Chrome, but this PR doesn't match those.

components/layout_2020/flow/inline.rs
mrobinson: Oh, that's a good point. I've updated this to smiply say "the amount of whitespace trimmed from the line."

components/layout_2020/flow/inline.rs
mrobinson: The calculation was incorrect! I have updated this in a way that I think corrects the case you mentioned and also have put a specification link and quote explaining why it's correct now.

mrobinson:
@mukilan Thanks for the review. I think this is ready to look at again, when you have a moment.

components/layout_2020/flow/inline.rs
mukilan: The `text-align: end` case  doesn't seem to work similar to the other layouts & master for [this case](http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=11943).

IIUC, the spec says 'text-indent' is treated as margin at the 'start'  edge of the linebox and 'text-align: end' will align the contents to the 'end' edge of the box, so the margin cannot not influence the end edge unless the margin exceeds the available width.

mukilan:
Thanks @mrobinson for updating the PR with the spec and comments! I've found one more case where 'text-end' doesn't work as other browsers. Please let me know if that is not in scope for this PR and needs to fixed separately as part of 'text-indent' stuff.

mrobinson:
> Thanks @mrobinson for updating the PR with the spec and comments! I've found one more case where 'text-end' doesn't work as other browsers. Please let me know if that is not in scope for this PR and needs to fixed separately as part of 'text-indent' stuff.

Thanks for pointing out this issue. I believe I've fixed it and have also added two WPT tests corresponding to the issues your raised above.

mukilan (approved):

mrobinson:
It seems that Layout 2013 does not pass this test...it looks like due to line breaking issues. All pipelines are passing in upstream WPT issue. I've updated the result and will send it back to the MQ.

PR #30140, opened 2023-08-19
Get LLVM from chocolatey

mrobinson:
LLVM is the largest package that we get from servo-build-deps, so installing it via chocolatey should reduce the amount of data that we transfer from that source. In addition, it's one less dependency that we have to manage.

It also seems that installing LLVM to the default location with choco means that we no longer have to set the LIBCLANG_PATH environment variable for bindgen.

---

- [x] These changes do not require tests because they are just a build dependency change.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -468,12 +468,9 @@ class CommandBase(object):
 
         effective_target = self.cross_compile_target or servo.platform.host_triple()
         if "msvc" in effective_target:
-            util.append_paths_to_env(env, "PATH", path.join(self.msvc_package_dir("llvm"), "bin"))
-
-            # These two environment variables are necessary for building mozjs. It would
-            # be nice to be able to set them in the cargo configuration somehow, but they
-            # are platform-dependent.
-            env["LIBCLANG_PATH"] = path.join(self.msvc_package_dir("llvm"), "lib")
+            # This environment variable is necessary for building mozjs. It would
+            # be nice to be able to set it in the cargo configuration somehow, but it
+            # is platform-dependent.
             env["MOZILLA_BUILD"] = path.join(self.msvc_package_dir("moztools"))
 
             # Always build harfbuzz from source
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -55,6 +55,8 @@ jobs:
           Start-BitsTransfer -Source https://github.com/wixtoolset/wix3/releases/download/wix3111rtm/wix311-binaries.zip -Destination C:\\wix311-binaries.zip
           Expand-Archive C:\\wix311-binaries.zip -DestinationPath C:\\wix
           echo "C:\\wix" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Remove msys from PATH # see https://github.com/actions/runner-images/issues/2208
+        run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
       - name: ccache
         uses: hendrikmuhs/ccache-action@v1.2
       - name: Bootstrap
--- a/python/servo/platform/windows.py
+++ b/python/servo/platform/windows.py
@@ -19,7 +19,6 @@ from .base import Base
 
 DEPS_URL = "https://github.com/servo/servo-build-deps/releases/download/msvc-deps/"
 DEPENDENCIES = {
-    "llvm": "15.0.5",
     "moztools": "3.2",
 }
 
--- a/support/windows/chocolatey.config
+++ b/support/windows/chocolatey.config
@@ -2,6 +2,7 @@
 <packages>
   <package id="cmake" version="3.26.4" installArguments="ADD_CMAKE_TO_PATH=System"/>
   <package id="git"/>
+  <package id="llvm"/>
   <package id="ninja"/>
   <package id="python"/>
   <package id="python3-virtualenv"/>


mukilan:
Nice! I wonder if alternatively we can convert the steps [here](https://github.com/servo/servo/wiki/Upgrading-the-windows-LLVM-binaries) into a script and use it to populate the Github Action cache on cache miss, that way we don't have to transfer the bandwidth  burden to chocolatey when we run more jobs ? The job can also upload the zip as an artifact/release for non-CI users.

python/servo/command_base.py
```diff
@@ -468,12 +468,9 @@ def build_env(self, is_build=False):
 
         effective_target = self.cross_compile_target or servo.platform.host_triple()
         if "msvc" in effective_target:
-            util.append_paths_to_env(env, "PATH", path.join(self.msvc_package_dir("llvm"), "bin"))
-
-            # These two environment variables are necessary for building mozjs. It would
+            # Thisenvironment variables is necessary for building mozjs. It would
             # be nice to be able to set them in the cargo configuration somehow, but they
             # are platform-dependent.
-            env["LIBCLANG_PATH"] = path.join(self.msvc_package_dir("llvm"), "lib")
```
mukilan, line 476: Seems like one of the failure is due to missing libclang
```
2023-08-19T09:03:22.8204856Z   thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed"', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\bindgen-0.62.0\./lib.rs:2402:31
```

mukilan:
Another error that I see in the logs is about the python site-packages, which I think means we need at least python 3.9?

```
2023-08-19T09:03:21.9765065Z error: failed to run custom build command for `mozjs_sys v0.68.2 (https://github.com/servo/mozjs#82168709)`
2023-08-19T09:03:21.9765554Z 
2023-08-19T09:03:21.9765793Z Caused by:
2023-08-19T09:03:21.9766498Z   process didn't exit successfully: `C:\\a\\servo\\servo\\target\release\build\mozjs_sys-e6bad8eef5d799e9\build-script-build` (exit code: 101)
2023-08-19T09:03:21.9767150Z   --- stdout
2023-08-19T09:03:21.9767674Z   cargo:outdir=C:\\a\\servo\\servo\\target\release\build\mozjs_sys-5705162ba5a59385\out\build
2023-08-19T09:03:21.9768898Z   [[ 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure -ot 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure.in ]] && touch 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure || true
2023-08-19T09:03:21.9770801Z   [[ 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure -ot 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure.in ]] && touch 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure || true
2023-08-19T09:03:21.9772449Z   ! [[ 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure.in -ot 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure ]] && touch 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure || true
2023-08-19T09:03:21.9774100Z   ! [[ 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure.in -ot 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure ]] && touch 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/old-configure || true
2023-08-19T09:03:21.9775345Z   if [[ 'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure -nt config.status ]] ; then \
2023-08-19T09:03:21.9775972Z     MOZ_TOOLS="/" \
2023-08-19T09:03:21.9776418Z     CC="clang-cl.exe" CFLAGS="-MD" \
2023-08-19T09:03:21.9777069Z     CPP="" CPPFLAGS="-IC:/Users/runneradmin/.cargo/registry/src/github.com-1ecc6299db9ec823/encoding_c_mem-0.2.6/include " \
2023-08-19T09:03:21.9777849Z     CXX="clang-cl.exe" CXXFLAGS="-MD" \
2023-08-19T09:03:21.9778284Z     AS="" AR="" \
2023-08-19T09:03:21.9778648Z     STLPORT_LIBS="" \
2023-08-19T09:03:21.9779766Z     'C:\Users\runneradmin\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\8216870\mozjs\mozjs'/js/src/configure --disable-jemalloc --disable-js-shell --disable-tests --disable-export-js --disable-shared-js --build-backends=RecursiveMake --enable-js-streams --with-ccache=ccache --target=x86_64-pc-mingw32 --host=x86_64-pc-mingw32 || (cat config.log && exit 1) ; \
2023-08-19T09:03:21.9780802Z   fi
2023-08-19T09:03:21.9781331Z   Using Python 3.7.9 from c:\hostedtoolcache\windows\Python\3.7.9\x64\python3.exe
2023-08-19T09:03:21.9782079Z   Site not up-to-date reason: "C:\a\servo\servo\target\release\build\mozjs_sys-5705162ba5a59385\out\build\_virtualenvs\build" does not exist
```

python/servo/command_base.py
mukilan: This [patch](https://github.com/mukilan/servo/commit/ae7fcc1308a1e426b9f18996f8e66ec87318cbb2) fixed the build for me. I've triggered another build to see if setup-python is still required.

sagudev:
I wonder if this is really needed as in mozjs repo we use LLVM from runners: https://github.com/servo/mozjs/blob/071ccca2f78b587553d2ba78a1b09377b96232cd/.github/workflows/rust.yml#L91C8

mukilan:
@sagudev Seems like you are right, we can use the LLVM from runners, but that only seems to work after removing msys64 from path variable.

Build with LLVM from runner + drop msys from PATH - https://github.com/mukilan/servo/actions/runs/5916564620
Build with just LLVM from runner - https://github.com/mukilan/servo/actions/runs/5916564620

I'm not sure why the msys path is there or why mozjs works without that logic. One difference I notice is that mozjs uses windows-latest image whereas servo uses windows-2019.

mrobinson:
> I wonder if this is really needed as in mozjs repo we use LLVM from runners: https://github.com/servo/mozjs/blob/071ccca2f78b587553d2ba78a1b09377b96232cd/.github/workflows/rust.yml#L91C8

This should work fine on the GitHub runner, but that still won't work for `./mach bootstrap`. We could perhaps be a bit better in `./mach bootstrap` in order to try not to install LLVM if it's already installed.

python/servo/command_base.py
mrobinson: It's so odd that Python needs to be installed, because it should be installed by chocolatey. I wonder if the choco install if failing for some reason.

mrobinson:
@bors-servo try=windows

Kicking off a try job to see if we can get better debugging output.

python/servo/command_base.py
mukilan: Actually, python 3.10 was not [needed](https://github.com/mukilan/servo/actions/runs/5915957966). I'm not sure why the try run failed in this PR with the error
```
2023-08-19T09:03:21.9782079Z   Site not up-to-date reason: "C:\a\servo\servo\target\release\build\mozjs_sys-5705162ba5a59385\out\build\_virtualenvs\build" does not exist
```

sagudev:
Also could we use winget instead of choco as it's already installed on windows systems

mrobinson:
> Also could we use winget instead of choco as it's already installed on windows systems

Using winget would be ideal, but there is no version for Windows Server (or it's very new) which is why it's not installed or supported on the Github Actions runners: https://github.com/actions/runner-images/issues/910. The other issue with winget is that the package repository is currently smaller so it's trickier to find all the dependencies we need.

I would prefer not to add another mechanism to automatically install dependencies, because that would be one more thing we need to support without having CI to catch breakage.

The general philosophy I think we should have is that the bots should run more-or-less what the build instructions say. This way  we can detect when things break and actually test what we are asking people to do when building servo on their own machines. This is why I think we should avoid using the pre-installed packages on the GitHub Runners if possible, but do cache downloads when we can.

mukilan:
> This way we can detect when things break and actually test what we are asking people to do when building servo on their own machines. This is why I think we should avoid using the pre-installed packages on the GitHub Runners if possible, but do cache downloads when we can.

Can we not acheive this by ensuring all nightly builds use the bootstrapped dependencies but the main & try workflows (which run extremely frequently

mukilan:
Sorry, accidentally closed the PR!

sagudev:
> > Also could we use winget instead of choco as it's already installed on windows systems
> 
> Using winget would be ideal, but there is no version for Windows Server (or it's very new) which is why it's not installed or supported on the Github Actions runners: [actions/runner-images#910](https://github.com/actions/runner-images/issues/910).

That's crazy!

Looking at our choco file it looks like we install python even though we are already running in python for `mach bootstrap`

mrobinson:
Ah, it seems the issue with this change is that the GitHub CI doesn't run `./mach bootstrap` now in any case. If this is the status quo already anyway, I guess I don't see the harm in just making it so that the pre-installed version of LLVM is used and then tackling the issue of CI==developer experience later.

mukilan (approved):

PR #30142, opened 2023-08-19
Cache .deb for libffi instead of always downloading

mukilan:
Connections to mirrors.kernel.org have recently been timing out frequently ([latest](https://github.com/servo/servo/actions/runs/5910609625/job/16032728310) failure). I've done the same in our [internal WPT dashboard ](https://github.com/servo/internal-wpt-dashboard/commit/9ef9aa7f53e757e96b1e15121ba7bac611b784c1) previously and it seems to work well so far. Maybe we can also host this in servo-build-deps as well, in addition to caching.

---

- [x] These changes do not require tests because they modify CI configuration.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -39,6 +39,16 @@ jobs:
         with:
           name: release-binary
           path: release-binary
+      - name: Cache libffi
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: ./libffi6_3.2.1-8_amd64.deb
+          key: cache-libffi
+      - name: Download libffi
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          wget http://mirrors.kernel.org/ubuntu/pool/main/libf/libffi/libffi6_3.2.1-8_amd64.deb
       - name: unPackage binary
         run: tar -xzf release-binary/target.tar.gz
       - name: Prep test environment
@@ -46,7 +56,6 @@ jobs:
           python3 -m pip install --upgrade pip virtualenv
           sudo apt update
           sudo apt install -qy --no-install-recommends libgl1 libssl1.1 libdbus-1-3 libxcb-xfixes0-dev libxcb-shape0-dev libunwind8 libegl1-mesa
-          wget http://mirrors.kernel.org/ubuntu/pool/main/libf/libffi/libffi6_3.2.1-8_amd64.deb
           sudo apt install ./libffi6_3.2.1-8_amd64.deb
           python3 ./mach bootstrap-gstreamer
       - name: Sync from upstream WPT


jdm (approved):

.github/workflows/linux-wpt.yml
```diff
@@ -39,14 +39,23 @@ jobs:
         with:
           name: release-binary
           path: release-binary
+      - name: Cache libffi
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: ./libffi6_3.2.1-8_amd64.deb
+          key: cache-libffi
```
mrobinson, line 47: It'd probably be a good idea to encode the version into this key so that new versions are a cache miss.

mrobinson:
I wonder how feasible it is to use the system version of GStreamer to build and then cache all of the apt packages?

.github/workflows/linux-wpt.yml
mukilan: Good point. This was a quick fix. Ideally the key & package url/name will be derived from a single source (e.g variable defintion, or from file) so we can updating one place will update both key and package url.

In the long term, I guess we'll need to know what causes the dependency on ffi6 so that we can upgrade that to use the system libffi7.

mrobinson: I think that this dependency on an old version of libffi is caused by the very old version of GStreamer that we ship with our binaries.

mukilan: That makes sense. Should we consider shipping newer GStreamer on Linux? I believe we ship 1.22.3 on MacOS.

Is it possible the support for a minimum of 1.16 of Gstreamer required for some particular compatiblity reason, like supporting Servo on Android? Or we just simply never updated them?

mukilan:
> I wonder how feasible it is to use the system version of GStreamer to build and then cache all of the apt packages?

Not sure about the second part ` then cache all of the apt packages` - that seems like we'll be building a poor emulation of docker :D

Also, once we ugrade to newer ubuntu for runners we would probably also need to fetch deb packages when there are security updates to them so there will be some logic to invalidate caches. Maybe we just expire them periodically, but we should also keep the different runners for a build run in sync.

Given this issue with Linux GStreamer and the issue with Windows LLVM, I wonder if we chose to use boostrapped dependencies in CI to deliberately test them. If we end up moving to system versions on all CI builds (which makes a lot of sense), should we have nightly "Boostrapped build" jobs just to ensure Servo builds with our pre-packaged dependencies as well?

mrobinson:
> > I wonder how feasible it is to use the system version of GStreamer to build and then cache all of the apt packages?
> 
> Not sure about the second part ` then cache all of the apt packages` - that seems like we'll be building a poor emulation of docker :D

Oh yeah, it's true that custom runner images might be a better bet for us in the future rather than caching everything!

I was really thinking integrating with a github action like this https://github.com/marketplace/actions/cache-apt-packages. I definitely think we should also consider a cache for cargo dependencies as well: https://github.com/marketplace/actions/rust-cache.

PR #30148, opened 2023-08-21
Handle URL without scheme

atbrakhi:
Currently: 
- When we execute `./mach run --release servo.org`, it displays the error msg `Page could not load...`
- In minibrowser mode, typing `servo.org` in location bar results in no action.

This patch fixes both issues by:
 - Checking if the URL path corresponds to a file when executed via the command line.
 - Sanitizing the URL when in minibrowser mode.


ports/winit/app.rs
```diff
@@ -346,12 +346,27 @@ fn get_default_url() -> ServoUrl {
...
+    let mut cmdline_url = opts::get().url.clone();
     let pref_url = {
         let homepage_url = pref!(shell.homepage);
         parse_url_or_filename(&cwd, &homepage_url).ok()
     };
     let blank_url = ServoUrl::parse("about:blank").ok();
 
+    // Check if a command-line URL is provided
+    if let Some(url) = cmdline_url.clone() {
+        let url_path = url.path();
```
mrobinson, line 358: Maybe you should check if the URL is relative (scheme and host are empty) before going down the path of checking the filesystem.

ports/winit/app.rs
```diff
@@ -346,12 +346,27 @@ fn get_default_url() -> ServoUrl {
...
+
+        // Check if the URL path corresponds to a file
+        if fs::metadata(url_path).is_ok() && fs::metadata(url_path).unwrap().is_file() {
```
mrobinson, line 361: Could this be something like:

```rust
fs::metadata(url_path).map(|metadata| metadata.is_file()).unwrap_or(false)
```

ports/winit/app.rs
```diff
@@ -346,12 +346,27 @@ fn get_default_url() -> ServoUrl {
...
+            return cmdline_url.unwrap();
+        }
+
+        // Extract domain from URL path
+        let domain = url_path.split('/').last().unwrap_or("");
```
mrobinson, line 366: Maybe we can convert this into an actual URL object in order to get the domain? Looking for slashes and such ourselves without actually trying to parse a URL seems a little brittle.

ports/winit/minibrowser.rs
```diff
@@ -90,11 +94,9 @@ impl Minibrowser {
...
                     let browser_id = browser.browser_id().unwrap();
                     let location = self.location.borrow();
-                    let Ok(url) = ServoUrl::parse(&location) else {
-                        warn!("failed to parse location");
-                        break;
-                    };
-                    app_event_queue.push(EmbedderEvent::LoadUrl(browser_id, url));
+                    if let Some(url) = sanitize_url(&location.clone()) {
+                        app_event_queue.push(EmbedderEvent::LoadUrl(browser_id, url));
+                    }
```
mrobinson, line 99: I think the warning is pretty useful here, so I would restore it.

mrobinson (changes requested):

delan (changes requested):
This is a tricky area of logic that is also easy to test, so I really think we should add some unit tests to make sure all of the inputs behave as expected. For example:

* `data:text/html,a`
    * should resolve to itself in the location bar ✔
    * should resolve to itself on the command line ✘
        * we search for `html,a`
* `/dev/null` (on unix at least)
    * should resolve to `file:///dev/null` in the location bar ✘
        * we yield `https://dev/null`
    * should resolve to `file:///dev/null` on the command line ✘
        * we search for `null`
* `resources/public_domains.txt`
    * should resolve to `http://resources/public_domains.txt` in the location bar ✔
    * should resolve to `file:///path/to/servo/resources/public_domains.txt` on the command line ✔
* `dragonfruit` (no dots!)
    * should resolve to `http://dragonfruit` **or** a search for `dragonfruit` in the location bar ✔
    * should resolve to `file:///path/to/servo/dragonfruit` on the command line if file exists ✔
    * should resolve to `http://dragonfruit` on the command line otherwise ✘
        * we search for `dragonfruit`
* `README.md` (file definitely exists, has dots, also a known tld)
    * should resolve to `http://README.md` in the location bar ✔
    * should resolve to `file:///path/to/servo/README.md` on the command line ✔
* `nic.md` (has dots, also a known tld)
    * should resolve to `http://nic.md` in the location bar ✔
    * should resolve to `file:///path/to/servo/nic.md` on the command line if file exists ✔
    * should resolve to `http://nic.md` on the command line otherwise ✔
* `foo.txt` (has dots, not a known tld, but we should treat it the same)
    * should resolve to `http://foo.txt` in the location bar ✔
    * should resolve to `file:///path/to/servo/foo.txt` on the command line if file exists ✔
    * should resolve to `http://foo.txt` on the command line otherwise ✔
* `nic.md/ro` (has dots, also a known tld)
    * should resolve to `http://nic.md/ro` in the location bar ✔
    * should resolve to `file:///path/to/servo/nic.md/ro` on the command line if file exists ✔
    * should resolve to `http://nic.md/ro` on the command line otherwise ✘
        * we search for `ro`
* `foo.txt/ro` (has dots, not a known tld, but we should treat it the same)
    * should resolve to `http://foo.txt/ro` **or** a search for `foo.txt/ro` in the location bar ✔
    * should resolve to `file:///path/to/servo/foo.txt/ro` on the command line if file exists ✔
    * should resolve to `http://foo.txt/ro` on the command line otherwise ✘
        * we search for `ro`

Note that to make ./mach test-unit work, we will need to rename the ports/winit package to something like “servo­_winit­_example” (see #29878), and make a “servo­_winit­_example­_tests” package in ports/winit/tests.

I think we’ll have an easier time reaching the desired behaviour if we clean up the various pieces of user-input-to-url logic, because it’s currently spread across three functions that are difficult to reason about. In particular:

* parse­_url­_or­_filename in opts.rs should be inlined into get_default_url in app.rs, Opts.url should be Option&lt;String>, and the tests should be moved accordingly
* sanitize_url should be renamed to location_bar_input_to_url

mrobinson:
> Note that to make ./mach test-unit work, we will need to rename the ports/winit package to something like “servo­_winit­_example” (see #29878), and make a “servo­_winit­_example­_tests” package in ports/winit/tests.

How would you all feel about "minibrowser" or some servo-specific name such as servoshell?

atbrakhi:
Thanks @delan @mrobinson I like the idea of adding unit tests. While fixing Martin's comments, I also discovered another bug, I am sure there are more. Unit tests will help catching and fixing those early.

I am thinking it might be good idea to do it in separate PRs. Separate once would be:
- Rename the package
- Refactor/ clean up the various pieces of user-input-to-url logic
- Add unit tests (that will be in this PR with required changes for handling url).

wdyt?

delan:
In this case, I can’t imagine how we would reasonably fix the bugs without refactoring

delan:
The current patch regresses cases like `data:text/html,a`, and I can’t imagine how we would fix those regressions without refactoring the logic so we can reason about it, or write unit tests without renaming the package so they can be run. I guess we could do the refactor + rename + tests in one patch and rebase this patch onto that, but it’s unclear to me if that’s worth the effort.

atbrakhi:
@delan and I discussed in chat and we decided to not merge this PR until the below is done:

1. Rename the package
2. Clean up the various pieces of user-input-to-url logic without any functional changes.

I will create separate PRs for both. Meanwhile I will mark this PR as draft. I will add unit test in this PR.

ports/servoshell/test.rs
```diff
@@ -85,3 +89,97 @@ fn test_argument_parsing_special() {
...
 }
+
+#[test]
+fn url_should_resolve_in_commad_line() {
+    embedder_traits::resources::set_for_tests();
+    let input = "resources/public_domains.txt";
+    let cwd = "../../";
+    env::set_current_dir(&cwd).expect("Failed to set current directory");
+
+    let result = get_default_url(Some(input.to_string()));
```
delan, line 100: nit: to_owned(), or ideally remove these and change the parameter to `Option<&str>`

delan (changes requested):
Thanks for the updates! This is a huge improvement.

We should isolate the tests from the real filesystem. This will make them more reliable, and allow us to fully test inputs that depend on whether some file exists. For example:

```rust
pub fn get_default_url(
    url_opt: Option<&str>,
    cwd: impl AsRef<Path>,
    exists: impl FnOnce(&str) -> bool,
) -> ServoUrl {}

// in production
get_default_url(todo!(), current_dir().unwrap(), |path| fs::metadata(path).is_some());

// in tests
get_default_url(todo!(), "/path/to", |_| true);
get_default_url(todo!(), "/path/to", |_| false);
```

I also think that adding a test helper function would make it easier to see the relationship between the test inputs and expected values. For example:

```rust
fn test(input: &str, location: &str, cmdline_if_exists: &str, cmdline_otherwise: &str) {
    // .map(|url| url.to_string()) omitted for clarity
    assert_eq!(location_bar_input_to_url(input), location);
    assert_eq!(get_default_url(Some(input), "/path/to", |_| true), cmdline_if_exists);
    assert_eq!(get_default_url(Some(input), "/path/to", |_| false), cmdline_otherwise);
}

test("data:text/html,a", "data:text/html,a", "data:text/html,a", "data:text/html,a");
test("nic.md/ro", "http://nic.md/ro", "file:///path/to/ro", "http://nic.md/ro");
```

Once these are addressed, we should be able to do a final review of the test cases and land!

atbrakhi:
Thanks for the review delan. I will go through review comments and trigger another CI run for Linux specific.

> Some of the tests fail on my machine, even with --test-threads 1. Are you sure they’re being run on CI?

This is really weird. It passes locally for me and on [CI](https://github.com/servo/servo/actions/runs/6151082932/job/16690425204#step:12:830) as well.

delan:
Sorry, you’re right, the test failures were caused by breakage on my end. I’ll edit that part out.

ports/servoshell/parser.rs
```diff
@@ -26,30 +26,54 @@ pub fn get_default_url(url_opt: Option<String>) -> ServoUrl {
...
+                .ok()
+        });
+
+    if let Some(url) = cmdline_url.clone() {
+        if url.scheme() == "file" && url.host().is_none() {
+            let url_path = url.path();
+
+            // Check if the URL path corresponds to a file
+            if fs::metadata(url_path)
+                .map(|metadata| metadata.is_file())
```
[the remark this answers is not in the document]
atbrakhi: Correct. I had the option of either checking for symlinks | directories | file, or just check `fs::metadata(path).is_some()` as you suggested above. Fixed in [e801366](https://github.com/servo/servo/pull/30148/commits/e801366322dbf5f8b317af36231f70bfab41d9ec)

ports/servoshell/test.rs
atbrakhi: went with Option<&str> in [e801366](https://github.com/servo/servo/pull/30148/commits/e801366322dbf5f8b317af36231f70bfab41d9ec) :)

delan (approved):
Great work, let’s merge!

mrobinson (approved):

mrobinson:
This seems to be causing build failures on Mac and Linux and the Windows smoketest to timeout: https://github.com/servo/servo/actions/runs/6703381852

atbrakhi:
`set_for_tests()` in `embedder_traits::resources` was removed in an earlier PR, which was causing the failure. It seems be fine now. I will let it run on CI and, once it passes all the checks, I will add it to the MQ

mrobinson:
I'll send it to the merge queue. The fix for `set_for_tests()` is great, but I want to double-check that issues are gone on Windows. Unfortunately builds that don't finish have a tendency to clog up the merge queue.

atbrakhi:
Glad we double checked I just looked at windows failure earlier, I think it is related to file paths. Current run should fail on window, It is because of `FAKE_CWD` 

In linux/Mac it should be: `file:///fake/cwd/filename`
On windows it is: `file:///C:/fake/cwd/filename` I think

atbrakhi:
This is the failure:
```
[2023-10-31T18:34:37Z ERROR egui_glow] GL error, at C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
[2023-10-31T18:34:40Z ERROR egui_glow] GL error, at C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
[2023-10-31T18:34:40Z ERROR egui_glow] GL error, at C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
WARNING: no real random source present!
WARN: compileToBinary(259): 
C:\fakepath(110,66-118): warning X3571: pow(f, e) will not work for negative f, use abs(f) or conditionally handle negative values if you expect them

Servo exited with non-zero status 3221225786
```

`egui_glow` is causing this error on windows. It seems to be reported by some other user on [egui repository](https://github.com/emilk/egui/issues/3131). I am investigating it. I am not sure why build is only failing for this PR and not others.

PR #30162, opened 2023-08-22
ci: use ubuntu 22.04 on linux runners

mukilan:
---

- [x] These changes do not require tests because they modify CI configuration and WPT test results should be good enough.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -21,7 +21,7 @@ env:
 jobs:
   linux-wpt:
     name: WPT ${{ inputs.layout }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       max_chunk_id: 20
     strategy:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -81,6 +81,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.9'
+      - name: Set LIBCLANG_PATH env # needed for bindgen in mozangle
+        run: echo "LIBCLANG_PATH=/usr/lib/llvm-14/lib" >> $GITHUB_ENV
       - name: Install taplo
         uses: baptiste0928/cargo-install@v2
         with:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -62,7 +62,7 @@ env:
 jobs:
   build:
     name: Linux Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
         if: github.event_name != 'issue_comment' && github.event_name != 'pull_request_target'
--- a/tests/wpt/meta-legacy-layout/css/css-fonts/infinite-size-crash.html.ini
+++ b/tests/wpt/meta-legacy-layout/css/css-fonts/infinite-size-crash.html.ini
@@ -1,0 +1,2 @@
+[infinite-size-crash.html]
+  expected: CRASH
--- a/tests/wpt/mozilla/meta-legacy-layout/css/inline_block_baseline_a.html.ini
+++ b/tests/wpt/mozilla/meta-legacy-layout/css/inline_block_baseline_a.html.ini
@@ -1,4 +1,2 @@
 [inline_block_baseline_a.html]
-  expected:
-    if os == "linux": FAIL
-  fuzzy: /_mozilla/css/inline_block_baseline_ref.html:49;2097
+  expected: [PASS, FAIL]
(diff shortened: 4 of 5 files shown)


mukilan:
@rego tested the servo binaries produced by the try job and they seem to work fine. I'm marking this PR as ready for review. Thanks for testing, @rego!

mrobinson (approved):

mrobinson:
Huh. It looks like the three crashing font related tests have to do with not having a font of last resort fallback.

mukilan:
Yeah, looks like the first try job only passed because it uses the master version of workflow definitions
```
Current runner version: '2.308.0'
Operating System
  Ubuntu
  20.04.6
  LTS
```

I guess I'll just run this in my fork by pushing to master and see what causes the font fallback issue.

mrobinson (approved):
Can you open a new bug with the stack trace about the font fallback issue? We should fix that one soonish, but it's unrelated to this change.

mrobinson:
Perhaps we can try to land this again now that failure to match fonts no longer causes a crash?

mukilan:
I was able to resolve the build failure by adding the following step to the linux workflow:
```
      - name: Set LIBCLANG_PATH env # needed for bindgen in mozangle
        run: echo "LIBCLANG_PATH=/usr/lib/llvm-14/lib" >> $GITHUB_ENV
```

Not sure why this is needed though. I'm able to compile servo on my Ubuntu 22.04 VM without issues. The above env variable is also present in mozangle's CI.

mrobinson:
Perhaps this is because the builder has a non-deb version of LLVM installed? @sagudev

sagudev:
> Perhaps this is because the builder has a non-deb version of LLVM installed? @sagudev

I knew at the time why this is needed but I do not remember it anymore.

mukilan:
I did some analysis on the unexpected test results

1. `PASS [expected FAIL] /_mozilla/css/inline_block_baseline_a.html`
 It looks like the rendered image in ubuntu 20.04 has a very subtle difference when compared to the reftest. I've uploaded the raw log [here](https://gist.github.com/mukilan/c6896bdb89e972a407c69bbe2d51aabf) which uploaded to the [analyzer](https://hg.mozilla.org/mozilla-central/raw-file/tip/layout/tools/reftest/reftest-analyzer-structured.xhtml) shows that the red div's right and bottom edges doesn't have a sharp transition from red to white/green. Maybe this is an anti-aliasing artifact and expectations can be updated now?


2. `OK /_webgl/conformance/glsl/misc/shader-varying-packing-restrictions.html`
```
        FAIL [expected PASS] subtest: WebGL test https://github.com/servo/servo/issues/4: [unexpected link status] shaders with one more than the maximum number of varyings of float should fail assert_true: [unexpected link status] shaders with one more than the maximum number of varyings of float should fail expected true got false
        FAIL [expected PASS] subtest: WebGL test https://github.com/servo/servo/pull/10: [unexpected link status] shaders with one more than the maximum number of varyings of vec2 should fail assert_true: [unexpected link status] shaders with one more than the maximum number of varyings of vec2 should fail expected true got false
        FAIL [expected PASS] subtest: WebGL test https://github.com/servo/servo/issues/22: [unexpected link status] shaders with one more than the maximum number of varyings of vec4 should fail assert_true: [unexpected link status] shaders with one more than the maximum number of varyings of vec4 should fail expected true got false
```

It looks like the unexpected cases are the ones that check that shader link fails when it uses more than maximum possible variables, but checking this case is flaky/dependenant on the driver  according to the comment [here in the test itself](https://github.com/servo/servo/blob/faf928f3a86a704b213dc6bb821fcb2f3f352285/tests/wpt/webgl/tests/conformance/glsl/misc/shader-varying-packing-restrictions.html#L168), so I believe we can update expectations for the failures.

 3. `CRASH [expected PASS] /css/css-fonts/infinite-size-crash.html`
 
 This seems similar to the earlier large font size issues discovered here. It passes on Ubuntu 20.04 but crashes on 22.04 when we unwrap [here](https://github.com/servo/servo/blob/master/components/layout/text.rs#L547). I am not sure how to handle the font matching failure in the callers to move to Result/Option.

mrobinson:
Updating the results for the first two cases sounds good to me. I will take a look at the third now.

mrobinson:
Looking at the third change, I think we should just mark it as crashing. The fix for this requires more work than I originally though and since legacy layout is in maintenance mode we likely shouldn't spend too much time on it. I'd say go ahead and update the expectation and land this one. Thanks!

mukilan:
I've added back the `expected: FAIL` for ` /_mozilla/css/inline_block_baseline_a.html` as it is now failing in CI again. I haven't investigated the reason for this change in behavior. Seems like this odd intermittent behavior is only on legacy layout, so I'm not creating an intermittent issue. Let me know if this is still needed.

sagudev:
> I've added back the `expected: FAIL` for ` /_mozilla/css/inline_block_baseline_a.html` as it is now failing in CI again. I haven't investigated the reason for this change in behavior. Seems like this odd intermittent behavior is only on legacy layout, so I'm not creating an intermittent issue. Let me know if this is still needed.

This now PASSes. Maybe we should just use `[PASS, FAIL]` in expectation file for layout-2013.

mukilan:
Good idea! I've updated the expectation to `[PASS, FAIL]`

PR #30190, opened 2023-08-24
Disable WebRender debugger and upgrade webxr / media / rand

mrobinson:
Now that our dependencies are using a newer version of rand we can
upgrade. This change also disables the WebRender debugger feature
because it has been removed in later versions of WebRender and allows
deduplicating the rand dependency as well as quite a few others.

---

- [x] These changes do not require tests because they just upgrade dependencies / change features.


mukilan (approved):

mukilan:
Nice!

PR #30191, opened 2023-08-24
Auto approve and merge Dependabot patch upgrade PRs

mukilan:
Adapted from the [Github documentation][doc]. This version approves all patch upgrades to any dependency. Not sure if we have any dependency that must always be reviewed manually.

[doc]: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

- [x] These changes do not require tests because only new CI workflow is being added.

### DIFF
--- a/.github/workflows/dependabot-pr.yml
+++ b/.github/workflows/dependabot-pr.yml
@@ -1,0 +1,25 @@
+name: Approve & merge successful dependabot patch upgrade PRs
+on: pull_request
+
+permissions:
+  content: write
+  pull-requests: write
+
+jobs:
+  dependabot:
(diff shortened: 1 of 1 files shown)


mrobinson:
This seems like a reasonable change to me.

Pinging @jdm as well who has much more experience managing dependency upgrades. Recently, we approved pretty much every one that didn't cause the build to fail without too much trouble. What do you think?

If there are dependencies that we want to hold back we can always add them to the dependabot configuration.

jdm:
Yeah, I have no problem with this! Less manual effort, same result.

mrobinson (approved):
Let's give this a shot!

PR #30206, opened 2023-08-25
Fix type attribute for module script not in case insensitive issue

### LINKED ISSUE #30205: Handle type attribute for module script in case-insensitive
Currently, we always check the `type` attribute in case-sensitive way for `"module"`.

https://github.com/servo/servo/blob/66dd68a210566f3604195ea306ebc350aa3984d5/components/script/dom/htmlscriptelement.rs#L1189

However, based on Step 10 of https://html.spec.whatwg.org/multipage/#prepare-a-script, browsers should always check the attribute in case-insensitive way. Also, in Step 8 of https://html.spec.whatwg.org/multipage/#prepare-a-script, it mentions that the type string should be trimmed with leading and trailing ASCII whitespaces.

Thus, we need to fix the checking in the above line; then, we should be able to fix the failing tests in https://github.com/servo/servo/blob/master/tests/wpt/tests/html/semantics/scripting-1/the-script-element/module/type.html

ss77995ss:
* Use `to_ascii_lowercase()` and `trim_matches(HTML_SPACE_CHARACTERS)` to check type string

---

- [x] These changes fix #30205  (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/htmlscriptelement.rs
+++ b/components/script/dom/htmlscriptelement.rs
@@ -1186,7 +1186,9 @@ impl HTMLScriptElement {
             (Some(ref ty), _) => {
                 debug!("script type={}", &***ty);
 
-                if &***ty == String::from("module") {
+                if ty.to_ascii_lowercase().trim_matches(HTML_SPACE_CHARACTERS) ==
+                    String::from("module")
+                {
                     return Some(ScriptType::Module);
                 }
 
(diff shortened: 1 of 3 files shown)


mrobinson (changes requested):
Great. I suspect you will also need to remove `tests/wpt/meta-legacy-layout/html/semantics/scripting-1/the-script-element/module/type.html.ini`.

mrobinson (approved):

mrobinson:
Thanks!

PR #30207, opened 2023-08-25
Use std::cell::OnceCell and remove mitochondria dependency

mrobinson:
`OnceCell` is now part of the standard library and we'll be able to use
it once we upgrade rust. For now we can use the version that's shipped
behind a feature flag in rust. This removes a dependency on one crate.

---


mukilan (approved):
Nice!

PR #30208, opened 2023-08-25
Enable grouping of patch updates in Dependabot

mukilan:
This feature was released recently. This could potentially cut down on the CI hours spent on merging dependabot PRs.

More info here:
https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#managing-dependabot-pull-requests-for-grouped-version-updates-with-comment-commands

---

- [x] These changes do not require tests because they only modify dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  groups:
+    patch-updates:
+      update-types:
+       - patch
   allow:
   - dependency-type: direct
   - dependency-type: indirect


mrobinson:
I've been thinking about this and I think that one way we can make things a bit smoother here is to limit the number of dependency updates that are grouped together. Is it true that groups are made of the "per-day" dependency updates? Can this change also limit the number of dependency updates in a single day to <= 5 in order to make it more likely that each batch will not fail due to tidy or build issues?

mukilan:
>  Is it true that groups are made of the "per-day" dependency updates?

I'm not sure I follow the intent of the question. I believe the group is constructed whenever dependabot checks for updates based on the scheduled configured in dependabot.yml, so it can be per-day, per-week or even per-month.  In our case, it is per-day due to our configuration. Not sure if that answers your question.

> Can this change also limit the number of dependency updates in a single day to <= 5 in order to make it more likely that each batch will not fail due to tidy or build issues?

I'm not sure that we can configure limits on number of updates in each group to construct batches of updates. GH only supports grouping based on name pattern, dependaecy type (prod or dev), update type (minor, major or patch). All updates matching the criteria for a group will be included in that group's PR. I don't see a way to configure how many such matches can be allowed in a group. 

Is there a disadvantage to merging this change as is and updating it or reverting it, based on experience? This is a new feature and based on the docs made available so far, it seems like we should be able to overcome the tidy/build issues by ignoring specific dependencies using the dependabot PR commands. But the exact behavior is still not clear to me, so I propose we try it out and see how that goes. WDYT?

Also, just an FYI, the auto-approved dependabot PRs don't see to trigger the main workflow, even though they are add added to the queue. I'm looking into this and will manually add them to the queue as I debug the issue.

sagudev:
IIRC default GITHUB_TOKEN provide in GH runners is not allowed to trigger new workflows. So it probably just needs different token to approve/queue this changes (`@bors-servo` acc might be great for that).

mukilan:
> IIRC default GITHUB_TOKEN provide in GH runners is not allowed to trigger new workflows. So it probably just needs different token to approve/queue this changes (`@bors-servo` acc might be great for that).

Oh, seems like that is the [probably the issue](https://github.com/orgs/community/discussions/27028). Thanks for the pointer! I'll update the workflow to use a bot token.

mrobinson:
> I'm not sure I follow the intent of the question. I believe the group is constructed whenever dependabot checks for updates based on the scheduled configured in dependabot.yml, so it can be per-day, per-week or even per-month. In our case, it is per-day due to our configuration. Not sure if that answers your question.

Ah, it seems I was a bit confused. There is an `open-pull-request-limit` which I thought was a limit of dependencies to update per-day, but it seems this isn't the case.

mrobinson (approved):

Issue #30150, opened 2023-08-21
Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate


mrobinson:
We have a lot of uses of a very old version of the `time` crate. We should gradually replace all of these uses with calls to `std::time` or if that isn't sufficient with `chrono` or the updated version of the `time` crate. We already depend on those crates.

Some crates should not be modified such as:

 - `components/style`
 - `components/selectors`
 - `third_party/webrender`

These are vendored from Gecko and we need to make the fixes upstream.

aadi58002:
I would like to work on this issue

mrobinson:
> I would like to work on this issue

Okay. I recommend that you try to a small portion of the code to update and do it in pieces.

aadi58002:
Hello there,
I would like know what your preferred way of doing this will be.
There are a lot of instances of the (Tm)[https://docs.rs/time/0.1.45/time/struct.Tm.html] struct which are used to compare time through out the program.It has both nanoseconds and seconds comparison which are taking place.I am using `Local::now().timestamp()` and `Local::now().timestamp_nanos()` from chrono to replace them.  I am not sure how we can clear show where the time used is in seconds or nanoseconds.

My suggestion is to move every time relates value from a i64 to a (Duration)[https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/std/time/struct.Duration.html].

aadi58002:
Hello there,
I would like to ask you about your preferred alternative to Tm struct from time.

It is frequently used for comparing time for various task ( like expiration of cookies ).
I currently replaced them with u64 which represent nanoseconds but in some place they use seconds for calculations. In struct definations it is not clear if a u64/i64 represents a seconds or nanosecond or millisecond. 

I am thinking of creating a Duration instance from std::time which we can get return from any range (seconds or nanoseconds or milliseconds) and they can be directly compared as well.

What do you think?

jdm:
Duration seems like a reasonable choice if it works.

mrobinson:
I agree. If we can use `Duration` that is much less ambiguous than using a number to hold seconds/milliseconds/nanoseconds.

michaelgrigoryan25:
Is there any work around this? If not I can help.

mrobinson:
> Is there any work around this? If not, I can help.

Any help would be greatly appreciated. Thanks!

vijayksingh:
[...]

PR #30220, opened 2023-08-26
Moving third_party/webrender to chrono and std::time::Duration

aadi58002:
---

- [X] These changes fixes part of #30150

- [X] These changes do not require tests because these changes remove a old library (time) with a better maintained library (chorno)


aadi58002:
It took sometime to move things over. Originally i tried to move every thing as once ( later found out it was not a good idea but it was fun ). 

If you have any feedback related to the changes i have made it is very welcome. I think i have named changed things which take time as consideration to duration if i have changed some field which i should not have let me know we can discuss on it.

Also i noticed when i ran `./mach cargo-clippy` it suggested many many changes. I am not sure if that is intended or not.

sagudev:
Webrender is sourced from mozilla and servo is only it's downstream user (that's why it is in folder named `third_party`). So we need to maintain all our [changes to webrender as separate patches](https://github.com/servo/servo/tree/master/third_party/webrender/patches). IMO it's not worth maintaining patch.

[A quick look over all webrender usages of time create](https://searchfox.org/mozilla-central/search?q=time%3A%3A&path=gfx%2Fwr%2Fwebrender&case=false&regexp=false) reveals it's only used for `precise_time_ns` [to create Durations](https://searchfox.org/mozilla-central/search?q=precise_time_ns&path=gfx%2Fwr%2Fwebrender&case=false&regexp=false) or for debugging messages, which can all be replaced with `std::time::Instant`, but this should be proposed to upstream first.

aadi58002:
> Webrender is sourced from mozilla and servo is only it's downstream user (that's why it is in folder named `third_party`). So we need to maintain all our [changes to webrender as separate patches](https://github.com/servo/servo/tree/master/third_party/webrender/patches). IMO it's not worth maintaining patch.
> 
> [A quick look over all webrender usages of time create](https://searchfox.org/mozilla-central/search?q=time%3A%3A&path=gfx%2Fwr%2Fwebrender&case=false&regexp=false) reveals it's only used for `precise_time_ns` [to create Durations](https://searchfox.org/mozilla-central/search?q=precise_time_ns&path=gfx%2Fwr%2Fwebrender&case=false&regexp=false) or for debugging messages, which can all be replaced with `std::time::Instant`, but this should be proposed to upstream first.

Thanks for the info about the third party folder being mozilla downstream. Didn't knew that.
I start working on one of the crates in the components folder.I might create a bug report of this over on mozilla bug tracker.For the time being I keep this PR open to see what the others think about what to do with this PR.

mrobinson:
I agree with @sagudev that the contents of the 'third_party' directory are not the best place to do this work for the reasons mentioned. There's a chance that the webrender upstream will accept your work though, so maybe look into contributing it to Gecko.

PR #30222, opened 2023-08-28
Use @servo-bot token for approving dependabot PR

mukilan:
The default GITHUB_TOKEN is created for  the 'github-bot' user
and has limitations. Specifically, events generated by  github-bot
cannot trigger additional workflows.

This PR uses fine-grained PAT generated for @servo-bot account
with the permissions scoped to servo/servo repo and grants the
'contents: write' and 'pull_request: write' permissions.

---

- [x] These changes do not require tests because they only modify CI authentcation tokens

### DIFF
--- a/.github/workflows/dependabot-pr.yml
+++ b/.github/workflows/dependabot-pr.yml
@@ -2,8 +2,7 @@ name: Approve & merge successful dependabot patch upgrade PRs
 on: pull_request
 
 permissions:
-  contents: write
-  pull-requests: write
+  pull-requests: read
 
 jobs:
   dependabot:
@@ -13,8 +12,6 @@ jobs:
       - name: Dependabot metadata
         id: metadata
         uses: dependabot/fetch-metadata@v1
-        with:
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Approve the PR & enable auto-merge
         if: ${{ steps.metadata.outputs.update-type == 'version-update:semver-patch' }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-pr.yml
```diff
@@ -13,13 +12,11 @@ jobs:
...
-          github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Approve the PR & enable auto-merge
         if: ${{ steps.metadata.outputs.update-type == 'version-update:semver-patch' }}
         run: |
           gh pr review --approve "$PR_URL"
           gh pr merge --auto --merge "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          GH_TOKEN: ${{secrets.SERVO_BOT_PR_TOKEN}}
```
mrobinson, line 22: If this is a new token, would it be possible to rename it to SERVO_DEPENDABOT_TOKEN, so that we know what it is used for in the future when looking at project settings? We have a few other "bots" as well.

mrobinson (approved):
Thanks for the fix!

.github/workflows/dependabot-pr.yml
mrobinson: Oh, I see now that this token is for a new, generically named servo-bot user. Do you have more plans for this bot as well?

mukilan: Yep, I'm trying to switch the nightly build release process to also use the new @servo-bot user, but that will use a different token, scoped to servo/servo-nightly-builds repo.

The token variable is named that way so that we can probably use it in other contexts where the bot can review, merge or otherwise interact with PRs in any way, not necessarily dependabot PRs, but I don't have any such specific use case in mind right now. Also, FYI, the fine-grained token itself is named 'Approve dependabot PRs' and the description of the token also documents the intended usage. This is just the name of the token variable in this specific repo. Let me know if you want to change this.

mrobinson: That makes sense. If this token is used only to approve dependabot PRS for now let's just name it SERVO_DEPENDABOT_TOKEN until we expand the usage.

PR #30236, opened 2023-08-29
Fix typo in build.rs link directive for EGL

mukilan:
Due to this incorrect link directive, libsimpleservo.so ends up with
undefined references that cause the JNI `System.loadLibrary` to fail.

---

- [x] These changes do not require tests because they fix broken Android build

### DIFF
--- a/ports/libsimpleservo/api/build.rs
+++ b/ports/libsimpleservo/api/build.rs
@@ -39,14 +39,14 @@ fn main() {
             .write_bindings(gl_generator::StaticStructGenerator, &mut file)
             .unwrap();
 
-        // Historically, Android builds have succeeded with rust-link-lib=EGL.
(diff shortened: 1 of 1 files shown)


atbrakhi (approved):

ports/libsimpleservo/api/build.rs
```diff
@@ -46,7 +46,7 @@ fn main() {
         if target.contains("windows") {
             println!("cargo:rustc-link-lib=libEGL");
         } else {
-            println!("cargo:rust-link-lib=EGL");
+            println!("cargo:rustc-link-lib=EGL");
```
mrobinson, line 49: Thanks for the fix! You could also fix the comment above which refers to `rust-link-lib` as well.

mukilan: Ah, I didn't notice that. Fixed now.

Issue #30189, opened 2023-08-24
Running tests in `ports/winit/prefs` leads to `libservo` compilation failure.


atbrakhi:
Running any of the tests in `ports/winit/prefs.rs` results in compilation failure. Error:
 
```
error[E0433]: failed to resolve: use of undeclared crate or module `media_platform`
   --> components/servo/lib.rs:271:13
    |
271 |             media_platform::init();
    |             ^^^^^^^^^^^^^^ use of undeclared crate or module `media_platform`

error[E0433]: failed to resolve: use of undeclared crate or module `media_platform`
    --> components/servo/lib.rs:1033:13
     |
1033 |             media_platform::init();
     |             ^^^^^^^^^^^^^^ use of undeclared crate or module `media_platform`

For more information about this error, try `rustc --explain E0433`.
warning: `libservo` (lib) generated 1 warning
error: could not compile `libservo` due to 2 previous errors; 1 warning emitted
```

While passing `features=["media-gstreamer"]`  with  `libservo = { path = "../../components/servo", features=["media-gstreamer"] }` in `cargo.toml` fixes the compilation error, I am not sure if this is the best approach.

jdm:
How are you running the tests?

atbrakhi:
I used vscode `Run test` for running the test.

I am also able to see this failure if I go to my branch `rename_winit_package`(see [PR](https://github.com/servo/servo/pull/30163)) and add "servoshell" to [this](https://github.com/servo/servo/blob/acf7fdb90c65f6d00ec147d2f12d5dae95eeff10/python/servo/testing_commands.py#L199) in addition to the changes already made

sagudev:
You must choose one of media backends (dummy or GStreamer): https://github.com/servo/servo/blob/70f00889860ff12feb7ffec45cf383094c1cac88/components/servo/lib.rs#L174

It might make more sense to make media-dummy default and use GStreamer only if requested so we would use something like:
`#[cfg(not(feature = "media-gstreamer"))]`

But that would require to always depend on `servo-media-dummy` which IMO is the way to go as that would also help mach-less cargo commands.

sagudev:
> It might make more sense to make media-dummy default and use GStreamer only if requested so we would use something like: `#[cfg(not(feature = "media-gstreamer"))]`
> 
> But that would require to always depend on `servo-media-dummy` which IMO is the way to go as that would also help mach-less cargo commands.

I did quick PR for this: https://github.com/servo/servo/pull/30240

[...]

PR #30238, opened 2023-08-29
Move user input logic into servoshell

atbrakhi:
We have user input logic spread across different functions and files. This PR brings all the logic and tests to `servoshell`. See comment https://github.com/servo/servo/pull/30148#pullrequestreview-1588405617

---


ports/servoshell/Cargo.toml
```diff
@@ -57,7 +57,7 @@ glow = "0.12.2"
 keyboard-types = { workspace = true }
 lazy_static = { workspace = true }
 libc = { workspace = true }
-libservo = { path = "../../components/servo" }
+libservo = { path = "../../components/servo", features=["media-gstreamer"] }
```
atbrakhi, line 60: This is because of https://github.com/servo/servo/issues/30189

ports/servoshell/tests/mod.rs
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#[cfg(test)]
+mod servoshell_test;
```
mrobinson, line 6: I think you can just include the contents of `servoshell_test.rs` here.

ports/servoshell/Cargo.toml
mrobinson: Hrm...I think this will need to be addressed before this can land. I'm pretty sure we don't want to unconditionally enable this feature.

components/config/opts.rs
```diff
@@ -25,7 +25,7 @@ pub struct Opts {
     pub is_running_problem_test: bool,
 
     /// The initial URL to load.
-    pub url: Option<ServoUrl>,
+    pub url: Option<String>,
```
mrobinson, line 28: It feels like if we are moving the parsing out of opt.rs, we should also move the command-line argument out of `opt.rs`. `servoshell` has a few command-line arguments...perhaps this can move there.

mrobinson (changes requested):

components/config/opts.rs
atbrakhi: I am not sure if [7214d3a](https://github.com/servo/servo/pull/30238/commits/7214d3a19dd2c87cfd5f969f1db5b83bff30fe61) is what you meant. But I realized we are only using opt url in this function so I have moved it all inside `get_default_url()`

ports/servoshell/tests/mod.rs
atbrakhi: done in [fab229d](https://github.com/servo/servo/pull/30238/commits/fab229d33f3b9003c8351cfae11a5c22187a2254)

ports/servoshell/Cargo.toml
atbrakhi: Removed this change from this PR, as issue was resolved in another PR

atbrakhi:
There are still a few things that needs to be resolved in this PR. Once they are resolved, I will mark it as ready for review.

atbrakhi:
New changes:
- Taking inspiration from `mbrubeck` suggestion [here](https://users.rust-lang.org/t/unit-tests-for-bin-crate/54403) on writing unit test in bin crate, I have moved the functions we are trying to test in `lib.rs`. It made it easier to call the functions in tests and other files. I am all ears if you think this might not be the best
- Moved command-line argument out of `opt.rs`

ports/servoshell/Cargo.toml
```diff
@@ -13,6 +13,10 @@ name = "servo"
 path = "main.rs"
 bench = false
 
+[lib]
+name = "servoshell"
+path = "lib.rs"
+
```
mrobinson, line 19: I don't think it's strictly necessary to move the functions to a lib crate. It probably doesn't make sense to do this unless it is necessary (to write doctests as in the link you gave) or if we are going to be moving a lot of the logic from `servoshell` into a library.

mrobinson (changes requested):

atbrakhi:
@mrobinson thanks for the review again. I have updated the PR with the suggestions.

The only new thing in last commit is that I have moved `sanitize_url()` from `browser.rs` to `parser.rs`, so that we can have all the url parsing logic at one place.

ports/servoshell/parser.rs
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use std::env;
+use std::path::Path;
+
+use log::warn;
```
mrobinson, line 8: Please remove the blank line here.

ports/servoshell/app.rs
```diff
@@ -8,23 +8,22 @@ use crate::browser::Browser;
 use crate::embedder::EmbedderCallbacks;
 use crate::events_loop::{EventsLoop, WakerEvent};
 use crate::minibrowser::Minibrowser;
+use crate::parser::get_default_url;
 use crate::window_trait::WindowPortsMethods;
 use crate::{headed_window, headless_window};
+
```
mrobinson, line 14: Please remove this blank line in the middle of the use statements.

mrobinson (approved):
Thanks for the fixes! Just a couple nitpicks, but otherwise this looks good to me.

mrobinson (approved):

PR #30239, opened 2023-08-29
Let WPT commands take the --legacy-layout argument

mrobinson:
This also prints a better message when starting tests.

---

- [x] These changes do not require tests because they just change command-line arguments.

### DIFF
--- a/.github/workflows/mac-wpt.yml
+++ b/.github/workflows/mac-wpt.yml
@@ -10,6 +10,7 @@ on:
 env:
   RUST_BACKTRACE: 1
   SHELL: /bin/bash
+  WPT_COMMAND_LINE_ARG: "${{ inputs.layout == 'layout-2013' && '--legacy-layout' || '' }}"
 
 jobs:
   mac-wpt:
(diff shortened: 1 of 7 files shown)


python/wpt/update.py
```diff
@@ -126,8 +128,7 @@ def run_update(**kwargs) -> bool:
 
 def create_parser(**_kwargs):
     parser = wptcommandline.create_parser_update()
```
mukilan, line 130: Should we also update the argument in the [WPT import script](https://github.com/servo/servo/blob/master/.github/workflows/scheduled-wpt-import.yml#L55)

mukilan (approved):

mrobinson:
I think this will fail try because it changes how workflows are run. :/

.github/workflows/mac-wpt.yml
```diff
@@ -43,9 +43,12 @@ jobs:
           python3 ./mach bootstrap
       - name: Smoketest
         run: python3 ./mach smoketest --release
+      - name: WPT legacy layout argument
+        if: inputs.layout == 'layout-2013'
+        run: echo "WPT_COMMAND_LINE_ARG=--legacy-layout" >> $GITHUB_ENV
```
sagudev, line 48: this could be reduced to something like:
```yaml
ENV:
    WPT_COMMAND_LINE_ARG: ${{ inputs.layout == 'layout-2013' && "--legacy-layout" || "" }}
```

which should be faster

PR #30240, opened 2023-08-29
Always make dummy available, but use it only if no gstreamer

sagudev:
Per https://github.com/servo/servo/issues/30189#issuecomment-1697627591

---


mrobinson (changes requested):
I really like this change!

I think you'll also need to change how `configure_media_stack` works in `command_base.py`.

python/servo/command_base.py
```diff
@@ -461,7 +461,7 @@ def build_env(self, is_build=False):
         """Return an extended environment dictionary."""
         env = os.environ.copy()
 
-        if "media-dummy" not in self.features:
+        if "media-gstreamer" in self.features:
```
sagudev, line 464: This worked before the change, but now it doesn't because when running test we simply do not provide any features.

sagudev: I think unconditionally running `set_gstreamer_environment_variables_if_necessary` would be best.

sagudev:
try build: https://github.com/sagudev/servo/actions/runs/6020602591

mrobinson (approved):

PR #30246, opened 2023-08-30
Convert /_mozilla/css/iframe/multiple_external.html into a WPT crashtest

mrobinson:
This test was made before the use of WPT tests, much less WPT
crashtests, but it was originally made to test a crash. Initially, it
had an empty result, but text was added. The new reference for the test
relied on the bad float layout of legacy layout so was always incorrect.

This change converts the test into a WPT crashtest.

---


tests/wpt/tests/html/semantics/embedded-content/the-iframe-element/multiple-iframes-with-allow-scripts-crash.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
+<html>
+    <body>
```
mukilan, line 3: Just to confirm my understanding, the styles in the original crash test didn't contribute to the cause of the crash, so now they are removed? i.e we are not simply deleting this test because this new test without the CSS is still valuable as a crash test?

mukilan (approved):

tests/wpt/tests/html/semantics/embedded-content/the-iframe-element/multiple-iframes-with-allow-scripts-crash.html
mrobinson: I think it's very hard to say now if the styles contributed to the crash, since this test was added so long ago. I am happy to re-add them though.

mrobinson:
Thanks for the review! I've restored more of the original test case. The reason behind it is lost to git history, but this is probably enough for now.

PR #30274, opened 2023-09-01
Use arboard in servoshell instead of rust-clipboard

mrobinson:
rust-clipboard is unmaintained, which means that it pulls in very old
dependencies (including a version xcb with 3 critical security
vulnerabilities). In addition, we already depend on arboard. This
removes four crates from our dependency graph.

---

### DIFF
--- a/servo-tidy.toml
+++ b/servo-tidy.toml
@@ -20,6 +20,7 @@ rand = [
 [ignore]
 # Ignored packages with duplicated versions
 packages = [
+    "ahash",
     "arrayvec",
     "base64",
     "cfg-if",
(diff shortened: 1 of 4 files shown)


servo-tidy.toml
```diff
@@ -82,7 +82,6 @@ packages = [
     "uuid",
 
     # winit port minibrowser (servo/servo#30049)
```
delan, line 84: this comment can now be removed. the ignore for ahash was actually added in #29795, and unrelated to the minibrowser.

delan (commented):
Thanks for this, one minor issue but otherwise this looks good.

servo-tidy.toml
mrobinson: Thanks for the review, @delan. I've pushed a new version of this change the removes the comment and moves `ahash` into the main body of the exclusions.

delan (approved):
Looks good to me!

mrobinson:
Thanks for the review!

PR #30280, opened 2023-09-01
Fix PlacementAmongFloats to avoid missing some bands

### LINKED ISSUE #30276: BFC root may be placed below some float even if it would fit next to it
```html
<!DOCTYPE html>
<div style="display: flow-root; width: 200px; border: solid">
  <div style="float: left; width: 150px; height: 100px; background: cyan"></div>
  <div style="clear: left; float: left; width: 100px; height: 100px; background: orange"></div>
  <div style="display: flow-root; width: 100px; background: magenta; height: 100px; background: magenta"></div>
</div>
```

| Actual | Expected |
| - | - |
|  |  |

Loirooriol:
PlacementAmongFloats would stop iterating when current_bands would be empty, even if next_band wasn't at infinity.

Then the BFC root or replaced block was placed after all the floats, even if it could fit next to some of them.

This patch moves the next_band into current_bands so that the loop keeps considering bands.

---

- [X] These changes fix #30276

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -200,14 +200,29 @@ impl<'a> PlacementAmongFloats<'a> {
         })
     }
 
+    /// Checks if we either have bands or we have gone past all of them.
+    /// This is an invariant that should hold, otherwise we are in a broken state.
+    fn has_bands_or_at_end(&self) -> bool {
+        !self.current_bands.is_empty() || self.next_band.top.px().is_infinite()
+    }
+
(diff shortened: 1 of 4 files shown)


mrobinson (approved):

Loirooriol:
@bors-servo try=wpt-2020

I forgot to remove one of the `self.current_bands.pop_front();`

Loirooriol:
There was already a test, but I guess I can keep mine since `c414-flt-fit-001.xht` uses an inline-level replaced element and mine uses a block-level BFC root.

Loirooriol:
My test fails on legacy (`c414-flt-fit-001.xht` doesn't)

PR #30289, opened 2023-09-04
Refactor inline layout nesting

mrobinson:
This makes the nesting architecture of inline layout a lot simpler. Now
PartialInlineBoxFragment and InlineNestingLevelState are replaced by
two structs:

 - InlineContainerState
 - InlineBoxContainerState

InlineContainerState holds state for the root of the inline formatting
context, while InlineBoxContainerState holds state for inline boxes.
InlineBoxContainerState has an InlineContainerState as the first
element, thus "extends" it.

Now the inline iterators are stack variables directly in the `layout()`
method. They are no longer stored in the state. This avoids the weird
nesting of state and instead relies on a normal vector to hold the
stack of state.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -87,33 +87,6 @@ pub(crate) struct TextRun {
     pub has_uncollapsible_content: bool,
 }
 
-struct InlineNestingLevelState<'box_tree> {
-    remaining_boxes: InlineBoxChildIter<'box_tree>,
-    line_items_so_far: Vec<LineItem>,
-    /// Whether or not we have processed any content (an atomic element or text) for
-    /// this inline box on the current line OR any previous line.
-    has_content: bool,
-    /// Indicates whether this nesting level have text decorations in effect.
-    /// From https://drafts.csswg.org/css-text-decor/#line-decoration
-    // "When specified on or propagated to a block container that establishes
-    //  an IFC..."
-    text_decoration_line: TextDecorationLine,
-}
-
-struct PartialInlineBoxFragment<'box_tree> {
-    base_fragment_info: BaseFragmentInfo,
-    style: Arc<ComputedValues>,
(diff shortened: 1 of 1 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -205,11 +215,17 @@ struct InlineFormattingContextState<'box_tree, 'a, 'b> {
     /// common ancestor is used.
     white_space: WhiteSpace,
 
-    partial_inline_boxes_stack: Vec<PartialInlineBoxFragment<'box_tree>>,
-    current_nesting_level: InlineNestingLevelState<'box_tree>,
+    /// The [`InlineContainerState`] for the container formed by the root of the
+    /// [`InlineFormattingContext`].
+    root_nesting_level: InlineContainerState,
+
+    /// A stack of [`InlineBoxState`] that is used to produce [`LineItem`]s either when we
```
mukilan, line 222: This needs to be updated to `InlineContainerState`

components/layout_2020/flow/inline.rs
```diff
@@ -739,13 +814,18 @@ impl InlineFormattingContext {
...
-                match &mut *child.borrow_mut() {
-                    InlineLevelBox::InlineBox(inline) => {
-                        let partial =
-                            PartialInlineBoxFragment::new(inline, child.clone(), &mut ifc);
-                        ifc.partial_inline_boxes_stack.push(partial);
+            match iterator.next() {
+                Some(child) => match &mut *child.borrow_mut() {
+                    InlineLevelBox::InlineBox(inline_box) => {
+                        ifc.start_inline_box(inline_box);
+                        let parent = std::mem::replace(
```
mukilan, line 824: We can eliminate  the `mem::replace` with

```
parent_iterators.push(iterator);
iterator = InlineBoxChildIter::from_inline_level_box(child.clone());
```

mukilan (approved):

PR #30290, opened 2023-09-04
constellation: crash to a new “sad tab” error page

### LINKED ISSUE #17187: about:failure doesn't exist any more
This disappeared along with the pre-fetch network stack; now basic_fetch only checks for about:blank. This explains why we sometimes end up seeing "Unexpected scheme" error pages.

delan:
Servo historically recovered from crashes by navigating forwards to an about:failure page, but we never got around to reimplementing about:failure in the fetch-based network stack, so it turned into an “Unexpected scheme” error.

That approach was less than ideal though, because navigating forwards trashes your forward history. Even if we were to navigate with replacement, the location bar would show “about:failure” instead of the original URL, and we would lose the original LoadData containing the request body and headers and so on. In both cases, it would be tricky to make the reload command work properly or add a reload button to the error page.

This patch replaces our use of about:failure with an error page that behaves similarly to a network error.

image

When the constellation detects a panic, we create a replacement pipeline as before, but we now reload the page in place with the same LoadData except with .crash set to the crash details. This gets plumbed through RequestBuilder and Request to the fetch code, which converts it to a NetworkError::Crash, which the HTML parser renders as a new kind of error page. The user can then view the reason and backtrace if any, and click a button to reload the page.

To test this manually, go to `data:text/html,<script>document.write()</script>` after applying the patch below:

```diff
diff --git a/components/script/dom/document.rs b/components/script/dom/document.rs
index c3d83be696..bbaab2182d 100644
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -5127,6 +5127,7 @@ impl DocumentMethods for Document {
 
     // https://html.spec.whatwg.org/multipage/#dom-document-write
     fn Write(&self, text: Vec<DOMString>) -> ErrorResult {
+        panic!();
         if !self.is_html_document() {
             // Step 1.
             return Err(Error::InvalidState);
```

---

- [x] These changes fix #17187

    - **feedback on how we can test this would be appreciated :)**

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2782,7 +2782,7 @@ where
 
         self.embedder_proxy.send((
             Some(top_level_browsing_context_id),
-            EmbedderMsg::Panic(reason, backtrace),
+            EmbedderMsg::Panic(reason.clone(), backtrace.clone()),
         ));
 
         let browsing_context = match self.browsing_contexts.get(&browsing_context_id) {
@@ -2810,7 +2810,7 @@ where
             Some(pipeline) => pipeline.load_data.clone(),
             None => return warn!("failed pipeline is missing"),
         };
-        if old_load_data.crash {
+        if old_load_data.crash.is_some() {
             return error!("crash page crashed");
         }
 
@@ -2818,7 +2818,12 @@ where
 
         let new_pipeline_id = PipelineId::new();
         let mut new_load_data = old_load_data.clone();
-        new_load_data.crash = true;
+        new_load_data.crash = Some(if let Some(backtrace) = backtrace {
+            format!("{}\n{}", reason, backtrace)
+        } else {
+            reason
+        });
+
         let sandbox = IFrameSandboxState::IFrameSandboxed;
         let is_private = false;
         self.new_pipeline(
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2797,7 +2797,6 @@ where
             Some(p) => p,
             None => return warn!("failed pipeline is missing"),
         };
-        let pipeline_url = pipeline.url.clone();
         let opener = pipeline.opener;
 
         self.close_browsing_context_children(
@@ -2806,23 +2805,20 @@ where
             ExitPipelineMode::Force,
         );
 
-        let failure_url = ServoUrl::parse("about:failure").expect("infallible");
-
-        if pipeline_url == failure_url {
-            return error!("about:failure failed");
+        let old_pipeline_id = pipeline_id;
+        let old_load_data = match self.pipelines.get(&pipeline_id) {
+            Some(pipeline) => pipeline.load_data.clone(),
+            None => return warn!("failed pipeline is missing"),
+        };
+        if old_load_data.crash {
+            return error!("crash page crashed");
         }
 
-        warn!("creating replacement pipeline for about:failure");
+        warn!("creating replacement pipeline for crash page");
 
         let new_pipeline_id = PipelineId::new();
-        let load_data = LoadData::new(
-            LoadOrigin::Constellation,
-            failure_url,
-            None,
-            Referrer::NoReferrer,
-            None,
-            None,
-        );
+        let mut new_load_data = old_load_data.clone();
+        new_load_data.crash = true;
         let sandbox = IFrameSandboxState::IFrameSandboxed;
         let is_private = false;
         self.new_pipeline(
@@ -2832,7 +2828,7 @@ where
             None,
             opener,
             window_size,
-            load_data,
+            new_load_data,
             sandbox,
             is_private,
             is_visible,
@@ -2841,10 +2837,12 @@ where
             top_level_browsing_context_id,
             browsing_context_id,
             new_pipeline_id,
-            replace: None,
+            // pipeline already closed, so Yes avoids harmless but useless close
+            replace: Some(NeedsToReload::Yes(old_pipeline_id, old_load_data)),
             new_browsing_context_info: None,
             window_size,
         });
+
     }
 
     fn handle_log_entry(
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2842,7 +2842,6 @@ where
             new_browsing_context_info: None,
             window_size,
         });
-
     }
 
     fn handle_log_entry(
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2844,7 +2844,8 @@ where
             top_level_browsing_context_id,
             browsing_context_id,
             new_pipeline_id,
-            // pipeline already closed, so Yes avoids harmless but useless close
+            // Pipeline already closed by close_browsing_context_children, so we can pass Yes here
+            // to avoid closing again in handle_activate_document_msg (though it would be harmless)
             replace: Some(NeedsToReload::Yes(old_pipeline_id, old_load_data)),
             new_browsing_context_info: None,
             window_size,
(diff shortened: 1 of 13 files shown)


resources/crash.html
```diff
@@ -0,0 +1,9 @@
+<html>
+<body>
+  <!-- TODO not this -->
+  <p>uh oh! servo has made a big fucky wucky! a widdle fucko boingo!</p>
```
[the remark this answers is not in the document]
jdm: to be clear I can't stop laughing at this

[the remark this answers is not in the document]
sagudev: I propose confused doge/servo + some text nobody will ever read:
confused_servo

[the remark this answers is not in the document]
delan: ^w^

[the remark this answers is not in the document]
delan: in the interest of keeping the defaults unopinionated, 99e41cc09afedcea67ae1773c1e80221adb41616 replaces this with just “Servo crashed!” while adding the reason and backtrace if any

jdm:
For testing, t might be possible to have an iframe that calls window.trap() (https://github.com/servo/servo/blob/c623ab991f804d3fd1c50e0276e8353eb0bd3eb8/components/script/dom/window.rs#L1112), and them the outer window could check that there's a cross-origin frame?

components/constellation/constellation.rs
```diff
@@ -2841,7 +2842,8 @@ where
             top_level_browsing_context_id,
             browsing_context_id,
             new_pipeline_id,
-            replace: None,
+            // pipeline already closed, so Yes avoids harmless but useless close
```
mrobinson, line 2845: Would it be possible to expand this comment a bit? Why is the pipeline already closed and where is the close avoided?

components/constellation/constellation.rs
```diff
@@ -2806,23 +2805,25 @@ where
...
-            None,
-            None,
-        );
+        let mut new_load_data = old_load_data.clone();
+        new_load_data.crash = Some(if let Some(backtrace) = backtrace {
+            format!("{}\n{}", reason, backtrace)
+        } else {
+            reason
+        });
+
```
mrobinson, line 2826: I think this could be slightly more idiomatically:

```rust
new_load_data.crash = backtrace.map(|backtrace|
    format!("{}\n{}", reason, backtrace)
).unwrap_or(reason);
```

components/net_traits/request.rs
```diff
@@ -254,6 +254,7 @@ pub struct RequestBuilder {
     pub initiator: Initiator,
     pub https_state: HttpsState,
     pub response_tainting: ResponseTainting,
+    pub crash: Option<String>,
```
mrobinson, line 257: It might be nice to put a little docstring here explaining what's up with this new member. Maybe something along the lines of:

```rust
/// When this member is non-None it means that the LoadData is meant to trigger an error page showing details about a crash.
```

mrobinson (changes requested):
Thanks for this change! This looks fine to me though I have a couple nit-picky comments.

components/constellation/constellation.rs
delan: done in db246afdd505aef117a00a9060d30846d9705fff

components/constellation/constellation.rs
delan: done in c1a6558ca29daeb81020934a828e6cf0014f99f8

components/net_traits/request.rs
delan: done in c1a6558ca29daeb81020934a828e6cf0014f99f8

delan:
> For testing, it might be possible to have an iframe that calls window.trap() ([…]), and then the outer window could check that there's a cross-origin frame?

This is interesting! std::intrinsics::trap and std::intrinsics::abort just kill the whole process, unless we are in multiprocess mode, but we can add a window.panic() function gated behind dom.servo_helpers. I’m not sure if the crash error page will count as cross-origin though, looking into it.

components/net_traits/lib.rs
```diff
@@ -758,8 +758,10 @@ pub enum NetworkError {
     /// Could be any of the internal errors, like unsupported scheme, connection errors, etc.
     Internal(String),
     LoadCancelled,
-    /// SSL validation error that has to be handled in the HTML parser
+    /// SSL validation error, to be converted to Resource::BadCertHTML in the HTML parser
     SslValidation(String, Vec<u8>),
+    /// Crash error, to be converted to Resource::Crash in the HTML parser
```
mrobinson, line 763: Nit: These docstrings are missing periods. ONly mentioned because it shows up on the doc page (if that's still working).

components/net_traits/request.rs
```diff
@@ -254,6 +254,8 @@ pub struct RequestBuilder {
     pub initiator: Initiator,
     pub https_state: HttpsState,
     pub response_tainting: ResponseTainting,
+    /// Servo internal: if crash details are present, trigger a crash error page with these details
```
mrobinson, line 257: Ditto.

components/net_traits/request.rs
```diff
@@ -488,6 +497,8 @@ pub struct Request {
     #[ignore_malloc_size_of = "Defined in rust-content-security-policy"]
     pub csp_list: Option<CspList>,
     pub https_state: HttpsState,
+    /// Servo internal: if crash details are present, trigger a crash error page with these details
```
mrobinson, line 500: Ditto.

components/script_traits/lib.rs
```diff
@@ -184,6 +184,9 @@ pub struct LoadData {
     pub srcdoc: String,
     /// The inherited context is Secure, None if not inherited
     pub inherited_secure_context: Option<bool>,
+
+    /// Servo internal: if crash details are present, trigger a crash error page with these details
```
mrobinson, line 188: Ditto.

resources/crash.html
```diff
@@ -0,0 +1,8 @@
+<p>Servo crashed!</p>
+
+<!-- NOTE: unlike in Firefox and Chrome, this reloads POST as GET -->
+<!-- see whatwg/html#6600 + whatwg/html#3215 -->
+<button onclick="location.reload()">Reload</button>
+
+<pre><plaintext>
+${details}
```
mrobinson, line 8: It looks like the [`<plaintext>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/plaintext) element is deprecated? Can this simply be:

```suggestion
<pre>${details}</pre>
```

mrobinson (approved):
Thanks! It sounds like testing this is tricky for now.  Just a couple tiny comments:

resources/crash.html
delan: not quite, because if the crash details contain `<` or `&` then they will be interpreted as html. &lt;plaintext> solves that problem by permanently moving the parser to the PLAINTEXT state, and i figured that would be simpler than escaping the output.

mrobinson: Hrm. Okay. I guess this is okay for now, but we should really consider adding some HTML escape functionality because it seems like `<plaintext>` is on the way out.

components/net_traits/lib.rs
delan: fixed in c85c102311a8914f8a34f0fb6932be6b2f163382

components/net_traits/request.rs
delan: fixed in c85c102311a8914f8a34f0fb6932be6b2f163382

components/net_traits/request.rs
delan: fixed in c85c102311a8914f8a34f0fb6932be6b2f163382

components/script_traits/lib.rs
delan: fixed in c85c102311a8914f8a34f0fb6932be6b2f163382

PR #30309, opened 2023-09-07
minibrowser: fix spurious update when location has not changed

delan:
In the minibrowser, we try to avoid egui updates unless absolutely necessary, since they’re expensive and we can easily end up falling behind a storm of input events. We reuse egui’s old display list unless one of the following happens:

1. we are rendering the minibrowser for the first time
2. we get a RedrawRequested event, after one or more Window::request_redraw calls
3. the current top-level URL has changed

While reworking #29976, we introduced a bug in that third case, making it behave as if the URL has changed on every cycle of the event loop.

This patch fixes that bug and adds trace logs for each update, showing the reason and time since last update.

A side effect of this is that our [TWGL Tunnel](https://demo.servo.org/experiments/twgl-tunnel/) demo, which would lock up *unpredictably* in minibrowser mode due to another bug we’re still investigating, now locks up *reliably*.

## To test this manually…

1. Compare a debug build with and without the following **patch that reverts the fix**

```diff
diff --git a/ports/servoshell/minibrowser.rs b/ports/servoshell/minibrowser.rs
index 870fef713a..04c26099a1 100644
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -117,6 +117,7 @@ impl Minibrowser {
                     self.location = RefCell::new(location.to_owned());
                     return true;
                 }
+                return true;
             }
         }
 
```

2. `$ export RUST_LOG=warn,servo::minibrowser=trace`

3. `$ ./mach run -d --minibrowser https://servo.org` and move your mouse around **inside the window**
    * when reverted, this yields updates due to `(RedrawRequested)` and `(update_location_in_toolbar)`
    * when fixed, this yields updates due to `(RedrawRequested)` only

4. `$ ./mach run -d --minibrowser https://servo.org` and move your mouse around **outside the window**
    * when reverted, this yields updates due to `(update_location_in_toolbar)`
    * when fixed, this yields no updates

5. `$ ./mach run -d --minibrowser https://demo.servo.org/experiments/twgl-tunnel/` and do nothing
    * when reverted, this yields constant updates due to `(update_location_in_toolbar)`
    * when fixed, this yields no updates, even if you move your mouse (inside or outside)

---

- [x] These changes are related to #30049

- [x] These changes do not require tests because the minibrowser is not currently testable

### DIFF
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -112,15 +112,17 @@ impl Minibrowser {
     /// Updates the location field from the given [Browser], unless the user has started editing it
     /// without clicking Go, returning true iff the location has changed (needing an egui update).
     pub fn update_location_in_toolbar(&mut self, browser: &mut Browser<dyn WindowPortsMethods>) -> bool {
-        if !self.location_dirty.get() {
-            if let Some(location) = browser.current_url_string() {
-                if location != self.location.get_mut() {
-                    self.location = RefCell::new(location.to_owned());
-                    return true;
-                }
-            }
+        // User edited without clicking Go?
+        if self.location_dirty.get() {
+            return false;
         }
 
-        false
+        match browser.current_url_string() {
+            Some(location) if location != self.location.get_mut() => {
+                self.location = RefCell::new(location.to_owned());
+                true
+            },
+            _ => false,
+        }
     }
 }
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
- use std::{cell::{RefCell, Cell}, sync::Arc};
+ use std::{cell::{RefCell, Cell}, sync::Arc, time::Instant};
 
 use egui::{TopBottomPanel, Modifiers, Key};
 use servo::{servo_url::ServoUrl, compositing::windowing::EmbedderEvent};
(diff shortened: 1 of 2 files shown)


ports/servoshell/minibrowser.rs
```diff
@@ -104,13 +109,15 @@ impl Minibrowser {
...
+    /// without clicking Go, returning true iff the location has changed (needing an egui update).
     pub fn update_location_in_toolbar(&mut self, browser: &mut Browser<dyn WindowPortsMethods>) -> bool {
         if !self.location_dirty.get() {
             if let Some(location) = browser.current_url_string() {
-                self.location = RefCell::new(location.to_owned());
-                return true;
+                if location != self.location.get_mut() {
+                    self.location = RefCell::new(location.to_owned());
+                    return true;
+                }
```
mrobinson, line 120: This becomes a bit deeply nested now. How do you feel about an early return in addition to a match statement?

```rust
// Comment explaining when this can happen, as it's still a little unclear
// to me and this is the opposite of what one might expect. A dirty variable
// usually triggers an update. Is it when the location is being edited?
if self.location_dirty.get() {
    return false;
}

match browser.current_url_string() {
    Some(location) if location != self.location.get_mut() {
        self.location = RefCell::new(location.to_owned());
        true
    }
   false
}
```

mrobinson (approved):
Nice fix and I appreciate the new tracing!

ports/servoshell/minibrowser.rs
delan: done in b6fcf51e006f1801bf8b9aed8d31ab09a662c6b7

PR #29906, opened 2023-06-21
Set sensible defaults for cargo without mach

sagudev:
Implemented what was talked about on [zulip](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/sensible.20default.20feature.20on.20winit.20port).

It's not very nice change but it is convenient to use.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #30310, opened 2023-09-07
Servoshell as default member

sagudev:
Small part of #29906 

---

- [x] These changes do not require tests because it only changes the default build target.


mrobinson:
This change means that when running `cargo build` the default build target is `servoshell.` This will make it a lot easier to build servo without `./mach`.

mrobinson (approved):

PR #30315, opened 2023-09-07
Layout 2013: Don't use WebRender border image outset support

mrobinson:
The border image outset support in WebRender is going to be removed and
even in versions of WebRender where it still exists, it fails to render
properly.

A border image is a type of border composed of slices of images. The
"outset" of this kind of border is a property in CSS that makes the
border boundaries expand. Previously, the value was passed to WebRender
which would expand the border by this amount and render the images into
the expanded rectangle.

Since this is going to be removed, we handle this property outside of
WebRender. The change is simply to expand the border area by the outset
before calculating the rest of the border values.

This is necessary for the WebRender upgrade.

---

### DIFF
--- a/components/layout/display_list/builder.rs
+++ b/components/layout/display_list/builder.rs
@@ -1185,19 +1185,20 @@ impl Fragment {
         let border_style_struct = style.get_border();
         let border_image_outset =
             border::image_outset(border_style_struct.border_image_outset, border_width);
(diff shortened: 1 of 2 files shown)


mrobinson:
It looks like this fixes one border image test. Things often fail due to pixel snapping and other issues, so I guess it isn't so unusual that there is at least one difference.

jdm (approved):

Issue #24710, opened 2019-11-11
Intermittent crash in /html/canvas/offscreen/the-offscreen-canvas/size.large.html, /html/canvas/offscreen/the-offscreen-canvas/size.large.worker.html


jdm:
```
  ▶ CRASH [expected OK] /offscreen-canvas/the-offscreen-canvas/size.large.html
  │ 
  │ 
  │ 
  │ 
  │ Stack trace for thread "CanvasThread"
  │ stack backtrace:
  │    0: backtrace::backtrace::trace
  │    1: backtrace::capture::Backtrace::new
  │    2: servo::install_crash_handler::handler
  │    3: _sigtramp
  │    4: AzCreateDrawTarget
  │    5: azure::azure_hl::DrawTarget::new
  │    6: <canvas::azure_backend::AzureBackend as canvas::canvas_data::Backend>::create_drawtarget
  │    7: canvas::canvas_data::CanvasData::recreate
  │    8: _ZN3std10sys_common9backtrace28__rust_begin_short_backtrace17hb22351f2903007feE.llvm.14423657731009663445
  │    9: __rust_maybe_catch_panic
  │   10: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   11: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  │   12: std::sys::unix::thread::Thread::new::thread_start
  │   13: _pthread_body
  │   14: _pthread_start
  └ 
```

jdm:
```
  ▶ CRASH [expected OK] /offscreen-canvas/the-offscreen-canvas/size.large.worker.html
  │ 
  │ 
  │ 
  │ 
  │ Stack trace for thread "CanvasThread"
  └ stack backtrace:
```

sagudev:
Fixed in https://github.com/servo/servo/pull/36445

PR #30321, opened 2023-09-08
Layout 2013: Don't make gradient display items for zero-sized gradients

mrobinson:
Before WebRender would ignore these, but newer version of WebRender have
issues with them. This change simply prevents legacy layout from
creating display items for these types of gradients. This is already the
behavior of non-legacy layout.

---

- [x] These changes do not require tests because this does not change visible behavior.


jdm (approved):

mrobinson:
Looks like the failure is #24710.

PR #30322, opened 2023-09-08
servoshell: fix lockups while animating

### LINKED ISSUE #30312: servoshell: minibrowser can cause lockups while animating
When there are active animations and the minibrowser is enabled, servoshell becomes unresponsive to input events and window resize events. Reproducible with:

* <https://demo.servo.org/experiments/twgl-tunnel/>
* `data:text/html,<span id=a>0</span><script>requestAnimationFrame(function f(){a.textContent=1-a.textContent;requestAnimationFrame(f)});</script>`
* `data:text/html,<span id=a></span><script>requestAnimationFrame(function f(){a.textContent=1;requestAnimationFrame(f)});</script>`

delan:
With the minibrowser enabled, we respond to input events by calling request_redraw, which defers the egui update to a RedrawRequested event. But once we’re there, we run the compositor _asynchronously_ by dispatching Servo a Refresh event, which defers painting (both compositor and egui) and presenting to some later tick of the event loop.

Per rust-windowing/winit#2900, this seems to be incorrect:

> applications should redraw any time they get a RedrawRequested callback (and they shouldn't defer the redraw outside of the winit callback)

When there are active animations (CSS animations, CSS transitions, and sometimes rAF), servoshell pumps its event loop as quickly as possible, without waiting for a waker. For reasons that are unclear to me, our misuse of RedrawRequested interacts with active animations to create a situation where:

* the event loop is flooded with a sequence of waker events (Recomposite → Webrender → ReadyToPresent)
* every second tick of the event loop requires a present, which blocks for the remaining time until vsync
* we never reach MainEventsCleared, RedrawEventsCleared, RedrawRequested
* we never get a chance to handle any input events or window resize events

This patch fixes those lockups by moving all painting and presenting to the RedrawRequested event handler, with two exceptions. When the compositor paints a frame of its own volition (e.g. due to WebRender), we request a redraw but suppress the recomposite step, since it already happened. When we are handling a resize event, we continue to call Servo::repaint_synchronously as before, but now we also update and paint the minibrowser and present the frame.

We also add trace logs for the flow of EmbedderEvent, EmbedderMsg, and winit events, trace logs for the major RedrawRequested and PumpRequest branches, and a debug message when we recomposite without presenting.

To test this manually, run the repros in #30312 and check that you can interact with the minibrowser ui, click on the page ([TWGL Tunnel](https://demo.servo.org/experiments/twgl-tunnel/) only), and resize the window, for at least 30 seconds.

---

- [x] These changes fix #30312

- [x] These changes do not require tests because the minibrowser is not currently testable


ports/servoshell/app.rs
```diff
@@ -146,17 +165,26 @@ impl App {
...
                     if let Some(mut minibrowser) = app.minibrowser() {
                         minibrowser.update(window.winit_window().unwrap(), "RedrawRequested");
-
-                        // Tell Servo to repaint, which will in turn allow us to repaint the
-                        // minibrowser and present a complete frame without partial updates.
-                        app.event_queue.borrow_mut().push(EmbedderEvent::Refresh);
+                        minibrowser.paint(window.winit_window().unwrap());
+                    }
+                    if external_present {
+                        app.servo.as_mut().unwrap().present();
```
jdm, line 183: Can we remove the `if external_present` condition, since that was just `let external_present = app.minibrowser.is_some();`?

ports/servoshell/app.rs
```diff
@@ -216,12 +244,32 @@ impl App {
...
+                    // TODO(servo#30049) can we replace this with the simpler Servo::recomposite?
+                    app.servo.as_mut().unwrap().repaint_synchronously();
+
                     if let Some(mut minibrowser) = app.minibrowser() {
+                        minibrowser.update(window.winit_window().unwrap(), "PumpResult::Resize");
                         minibrowser.paint(window.winit_window().unwrap());
                     }
-                    app.servo.as_mut().unwrap().present();
-                },
+                    if external_present {
```
jdm, line 269: Same question here.

jdm (approved):

delan:
9d5fd40a3e59322237c0e2ea6ec18dfcd74364c9 disables the needs_recomposite optimisation for now, since it causes problems on @atbrakhi’s machine that need deeper investigation. This shouldn’t affect the correctness of the patch.

PR #30323, opened 2023-09-08
Upgrade WebRender to e491e1ae637b2eed1e7195855d88357e5eb3ddf9

mrobinson:
This change upgrades WebRender to the upstream version as of May 25, 2021.
Upgrading to even newer versions requires even more adaptation of Servo to
adapt to APIs removed upstream.

I have made one concession in this upgrade which is that I have backported the
hit testing APIs that Servo relies on. In general, these are low impact
WebRender changes and this gives us margin to no longer rely on them.

In addition to the changes required in Servo, some tests start to fail. These
are likely due to timing and pixel alignment changes introduced in WebRender.
It is difficult to observe these failures locally and definitely not when using
Servo interactively. The upgrade also exposes some bad behavior that we have
around sticky frames. Hopefully the implementation of sticky in the new layout
system will fix this.

---


mrobinson:
@bors-servo try

Ah, there is one more thing I forgot to mention about these changes. The adaptation of Servo to the new API includes a bit more conversion to using ClipChains and ClipChainId consistently rather than CilpId. This isn't strictly necessary, but it will be necessary to deal with future version of WebRender.

mrobinson:
This is a draft because it depends on #30321 and also because I'd like to push it through the CI a bit.

mrobinson:
Okay. This one is ready for review.

components/layout_thread/lib.rs
```diff
@@ -109,7 +109,7 @@ use style::traversal_flags::TraversalFlags;
 use style_traits::CSSPixel;
 use style_traits::DevicePixel;
 use style_traits::SpeculativePainter;
-use webrender_api::{units, ColorF, HitTestFlags, ScrollClamping};
+use webrender_api::{units, ColorF, HitTestFlags /*, HitTestFlags*/};
```
mukilan, line 112: Can the commented item be removed?

mrobinson: Thanks for pointing this out. I have removed the commented out use statement.

mukilan (approved):

PR #30330, opened 2023-09-11
Update version of `nix`, and `ron`

fabricedesre:
A pot-pourri of crate updates to reduce duplicates.

---

- [X] These changes do not require tests because no feature change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4752,13 +4752,14 @@ dependencies = [
 
 [[package]]
 name = "ron"
-version = "0.8.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "300a51053b1cb55c80b7a9fde4120726ddf25ca241a1cbb926626f62fb136bff"
+checksum = "b91f7eff05f748767f183df4320a63d6936e9c6107d97c9e6bdd9784f4289c94"
 dependencies = [
- "base64 0.13.0",
- "bitflags 1.3.2",
+ "base64 0.21.3",
+ "bitflags 2.4.0",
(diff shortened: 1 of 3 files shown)


jdm (approved):
Thanks!

sagudev:
> Quotes, if present, are stripped from cookie values when parsing.

from https://github.com/SergioBenitez/cookie-rs/blob/master/CHANGELOG.md#general-changes-and-fixes-2

And that might not be according to specs https://github.com/abarth/http-state/blob/155e45c6168e6de2eb727193010fd12e3c789185/tests/data/parser/value0003-test

fabricedesre:
What do we want to do for the `cookie-rs` update?
- Wait to see if the maintainer answers in https://github.com/SergioBenitez/cookie-rs/issues/150 and/or https://github.com/SergioBenitez/cookie-rs/pull/141#issuecomment-1705504978
- Fork `cookie-rs` to the servo org and revert https://github.com/SergioBenitez/cookie-rs/commit/b62b4ed0c79258e33e41259b9512cb2f706ee486
- Not update cookie-rs and live with the dupe.

cc @mrobinson

mrobinson:
It would be great to unblock this, so my suggestion is to include everything here, but the cookie upgrade. I think that cookie will only be upgradeable once we resolve the situation with quoted cookie values.

fabricedesre:
> It would be great to unblock this, so my suggestion is to include everything here, but the cookie upgrade. I think that cookie will only be upgradeable once we resolve the situation with quoted cookie values.

done in latest forced push.

atouchet:
cookie 0.18 is now out.

PR #30338, opened 2023-09-11
Enable minibrowser by default but also keep the option to disable it

atbrakhi:
We want to enable minibrowser by default but also want to keep an option to disable it from command line

---

- [ ] These changes are part of #30049

### DIFF
--- a/components/config/opts.rs
+++ b/components/config/opts.rs
@@ -552,7 +552,7 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         "Directory root with unminified scripts",
         "",
     );
-    opts.optflag("", "no_minibrowser", "Open minibrowser");
+    opts.optflag("", "no-minibrowser", "Open minibrowser");
 
     let opt_match = match opts.parse(args) {
         Ok(m) => m,
@@ -768,7 +768,7 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         unminify_js: opt_match.opt_present("unminify-js"),
         local_script_source: opt_match.opt_str("local-script-source"),
         print_pwm: opt_match.opt_present("print-pwm"),
-        no_minibrowser: !opt_match.opt_present("no_minibrowser"),
+        no_minibrowser: !opt_match.opt_present("no-minibrowser"),
     };
 
     set_options(opts);
--- a/components/config/opts.rs
+++ b/components/config/opts.rs
@@ -127,8 +127,8 @@ pub struct Opts {
     /// Print Progressive Web Metrics to console.
     pub print_pwm: bool,
 
-    /// True to disable minibrowser
-    pub no_minibrowser: bool,
+    /// True to enable minibrowser
+    pub minibrowser: bool,
 }
 
 fn print_usage(app: &str, opts: &Options) {
@@ -413,7 +413,7 @@ pub fn default_opts() -> Opts {
         unminify_js: false,
         local_script_source: None,
         print_pwm: false,
-        no_minibrowser: true,
+        minibrowser: true,
     }
 }
 
@@ -768,7 +768,7 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         unminify_js: opt_match.opt_present("unminify-js"),
         local_script_source: opt_match.opt_str("local-script-source"),
         print_pwm: opt_match.opt_present("print-pwm"),
-        no_minibrowser: !opt_match.opt_present("no-minibrowser"),
+        minibrowser: !opt_match.opt_present("no-minibrowser"),
     };
 
     set_options(opts);
--- a/components/config/opts.rs
+++ b/components/config/opts.rs
@@ -128,7 +128,7 @@ pub struct Opts {
     pub print_pwm: bool,
 
     /// True to disable minibrowser
-    pub without_minibrowser: bool,
+    pub no_minibrowser: bool,
 }
 
 fn print_usage(app: &str, opts: &Options) {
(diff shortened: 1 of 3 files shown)


atbrakhi:
Draft because I want to see if it impacts wpt tests

@bors-servo try=wpt

components/config/opts.rs
```diff
@@ -552,7 +552,7 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         "Directory root with unminified scripts",
         "",
     );
-    opts.optflag("", "minibrowser", "Open minibrowser");
+    opts.optflag("", "without_minibrowser", "Open minibrowser");
```
delan, line 555: can this be no-minibrowser? with a dash, and “no” for simplicity

components/config/opts.rs
```diff
@@ -413,7 +413,7 @@ pub fn default_opts() -> Opts {
         unminify_js: false,
         local_script_source: None,
         print_pwm: false,
-        minibrowser: false,
+        without_minibrowser: false,
```
delan, line 416: can this be `minibrowser: true,` to avoid the double negative? then on line 771, we can invert the result of the opt_present call

components/config/opts.rs
atbrakhi: We can not rename it to `no-minibrowser` but we can to `no_minibrowser`. We get errors from rust-analyzer if we try to use dash. Also, I think for consistency reason it would be better to follow underscore as all other candidates do that in opts.rs

Renamed to `no_minibrowser` in [f23d5f6](https://github.com/servo/servo/pull/30338/commits/f23d5f6c550eb2ff1590bcd5a385c769762e3608)

components/config/opts.rs
atbrakhi: hmm, done in [f23d5f6](https://github.com/servo/servo/pull/30338/commits/f23d5f6c550eb2ff1590bcd5a385c769762e3608) but we couldn't escape double negative completely! See browser.rs.

components/config/opts.rs
mrobinson: Very strange. What error does `rust-analyzer` give? This is just a string and there are many other options with dashes.

atbrakhi: I guess i just need to restart my computer to fix the rust analyzer complains. Anyway renamed to `no-minibrowser` in [c38c20f](https://github.com/servo/servo/pull/30338/commits/c38c20f394573452f8dc9cfaf687a0ea97e18852)

With consistency I meant in Struct `Opts`. But you are right, we have many options with dash in the same file,.

components/config/opts.rs
```diff
@@ -127,8 +127,8 @@ pub struct Opts {
     /// Print Progressive Web Metrics to console.
     pub print_pwm: bool,
 
-    /// True to enable minibrowser
-    pub minibrowser: bool,
+    /// True to disable minibrowser
+    pub no_minibrowser: bool,
```
delan, line 131:
```suggestion
    /// True to enable minibrowser
    pub minibrowser: bool,
```

components/config/opts.rs
```diff
@@ -413,7 +413,7 @@ pub fn default_opts() -> Opts {
         unminify_js: false,
         local_script_source: None,
         print_pwm: false,
-        minibrowser: false,
+        no_minibrowser: true,
```
delan, line 416:
```suggestion
        minibrowser: true,
```

components/config/opts.rs
```diff
@@ -768,7 +768,7 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         unminify_js: opt_match.opt_present("unminify-js"),
         local_script_source: opt_match.opt_str("local-script-source"),
         print_pwm: opt_match.opt_present("print-pwm"),
-        minibrowser: opt_match.opt_present("minibrowser"),
+        no_minibrowser: !opt_match.opt_present("no-minibrowser"),
```
delan, line 771:
```suggestion
        minibrowser: !opt_match.opt_present("no-minibrowser"),
```

ports/servoshell/app.rs
```diff
@@ -73,7 +73,7 @@ impl App {
             minibrowser: None,
         };
 
-        if opts::get().minibrowser && window.winit_window().is_some() {
+        if opts::get().no_minibrowser && window.winit_window().is_some() {
```
delan, line 76:
```suggestion
        if opts::get().minibrowser && window.winit_window().is_some() {
```

ports/servoshell/browser.rs
```diff
@@ -112,7 +112,7 @@ where
                 }
             })
             .shortcut(CMD_OR_CONTROL, 'L', || {
-                if !opts::get().minibrowser {
+                if !opts::get().no_minibrowser {
```
delan, line 115:
```suggestion
                if !opts::get().minibrowser {
```

components/config/opts.rs
delan: only the option should use a dash, the field would have to be no_minibrowser

delan: (or just minibrowser if we’re inverting the result of opt_present, see my suggested changes)

atbrakhi: thanks for the clarification delan. I think i misunderstood earlier. keeping the `minibrowser` as it was before and as you suggested

components/config/opts.rs
atbrakhi: we kept the original name `minibrowser` in [4f70fb5](https://github.com/servo/servo/pull/30338/commits/4f70fb5e338edbe2cae80cc7d5b66370b73ec483).  With other things, it also helped us avoid double negative

delan (approved):

PR #30340, opened 2023-09-12
[NFC] servoshell: fix rust-analyzer and rustfmt breakage

delan:
servoshell uses include!() to replace the whole program with an error message when built for Android, but this makes the non-Android code completely invisible to rustfmt and partially breaks rust-analyzer. For example, Go to References fails to find the caller of App::run in main2.rs.

This patch replaces include!() with cfg_if!{} in ports/servoshell/{main,main2}.rs. All other changes are the result of rustfmt now being able to reformat the code in this package.

---


atbrakhi (approved):
Nice.

PR #30342, opened 2023-09-12
[NFC] servoshell: clarify that device_pixels_per_px is an override

delan:
App and the Window types have parameters and fields called device_pixels_per_px, which sound like they contain the device pixel ratio, but they actually only contain the override given in --device-pixel-ratio (if any). The actual device pixel ratio is calculated in Window::servo_hidpi_factor.

This patch clarifies that this is the case by renaming those to device_pixel_ratio_override.

---

### DIFF
--- a/ports/servoshell/app.rs
+++ b/ports/servoshell/app.rs
@@ -47,7 +47,7 @@ enum PumpResult {
 impl App {
     pub fn run(
         no_native_titlebar: bool,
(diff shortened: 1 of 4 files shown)


atbrakhi (approved):

mrobinson (approved):

delan:
(force push to rebase over #30341)

PR #30343, opened 2023-09-12
minibrowser: implement HiDPI support

### LINKED ISSUE #30341: HiDPI support

delan:
This patch makes the minibrowser behave correctly under HiDPI and mixed-dpi environments, by plumbing the servoshell Window’s scale factor into egui during context creation and on ScaleFactorChanged events.

* 694349f9cb001951624f00c9fcf12cfab9b3bace exposes our calculated scale factor as Window­Ports­Methods­::hidpi­_factor
* bc5ff50fc42f29fa953ff0025071feeaacd3df8d plumbs that into egui during context creation
* a9e9c9c3ead95b5d831375a0d80d95586491e0da multiplies that by the scale factor in Window­::get­_coordinates, which affects where Servo and WebRender will put the viewport, while marking the toolbar_height fields as DeviceIndependentPixel to prevent future regressions
* 69138c0b248c4dd77a9ad4279a3fd20c49939014 multiplies toolbar_height by the scale factor in the CursorMoved handler, which affects the relative coordinates of input events, while using euclid’s Size2D and Length types to prevent future regressions
* b90e315c995a4b3bfa4f4d448788aea22b014174 intercepts Scale­Factor­Changed events so we can set egui’s scale factor manually, because we don’t always want to use the device scale factor

To test this manually:

1. Run Servo on a HiDPI monitor with `RUST_LOG=warn,servo::app=info`
2. Go to `data:text/html,<body style=margin:0>BODY` and check that:
    1. text in the toolbar is at a reasonable size and not too small
    2. text in the page is at the size you would expect for unstyled text
    3. none of the word “BODY” is cut off by the toolbar
3. If you have a mixed-dpi setup, move the window to your other monitor and:
    1. check the same things as above
    2. check that the logs contain [INFO] window scale factor changed to &lt;new dpi>, setting scale factor to &lt;new dpi>
4. Repeat with --device-pixel-ratio 1 and check that:
    1. text in the toolbar and page is too small on your HiDPI monitor
    2. those log messages always end with “setting scale factor to 1”

Test results:

* ☑️ X11 with Xft.dpi: {96,192} (note that X11 doesn’t seem to support mixed-dpi)
* ☑️ Wayland with wlr-randr --output DP-2 --scale 1 --output DP-3 --scale 2
* **macOS not yet tested, please help!**
* ☑️ Windows build 19045 with one 100% monitor and one 200% monitor

---

- [x] These changes fix #30341

- [x] These changes do not require tests because the minibrowser is not currently testable, but euclid’s type safety is now used where possible

### DIFF
--- a/ports/servoshell/app.rs
+++ b/ports/servoshell/app.rs
@@ -10,13 +10,14 @@ use std::rc::Rc;
 use std::time::Instant;
 
 use gleam::gl;
-use log::{trace, warn};
+use log::{info, trace, warn};
 use servo::compositing::windowing::EmbedderEvent;
 use servo::config::opts;
 use servo::servo_config::pref;
 use servo::Servo;
 use surfman::GLApi;
 use webxr::glwindow::GlWindowDiscovery;
+use winit::event::WindowEvent;
 use winit::event_loop::EventLoopWindowTarget;
 use winit::window::WindowId;
 
@@ -95,9 +96,8 @@ impl App {
             webrender_surfman.make_gl_context_current().unwrap();
             debug_assert_eq!(webrender_gl.get_error(), gleam::gl::NO_ERROR);
 
-            // Set up egui context for minibrowser ui
-            // Adapted from https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui_glow/examples/pure_glow.rs
-            app.minibrowser = Some(Minibrowser::new(&webrender_surfman, &events_loop).into());
+            app.minibrowser =
+                Some(Minibrowser::new(&webrender_surfman, &events_loop, window.as_ref()).into());
         }
 
         if let Some(mut minibrowser) = app.minibrowser() {
@@ -119,15 +119,23 @@ impl App {
         let t_start = Instant::now();
         let mut t = t_start;
         let ev_waker = events_loop.create_event_loop_waker();
-        events_loop.run_forever(move |e, w, control_flow| {
+        events_loop.run_forever(move |event, w, control_flow| {
             let now = Instant::now();
-            match e {
-                // Uncomment to filter out logging of DeviceEvent, which can be very noisy.
+            match event {
+                // Uncomment to filter out logging of common events, which can be very noisy.
                 // winit::event::Event::DeviceEvent { .. } => {},
-                _ => trace!("@{:?} (+{:?}) {:?}", now - t_start, now - t, e),
+                // winit::event::Event::WindowEvent {
+                //     event: WindowEvent::CursorMoved { .. },
+                //     ..
+                // } => {},
+                // winit::event::Event::MainEventsCleared => {},
+                // winit::event::Event::RedrawEventsCleared => {},
+                // winit::event::Event::UserEvent(..) => {},
+                // winit::event::Event::NewEvents(..) => {},
+                _ => trace!("@{:?} (+{:?}) {:?}", now - t_start, now - t, event),
             }
             t = now;
-            match e {
+            match event {
                 winit::event::Event::NewEvents(winit::event::StartCause::Init) => {
                     let surfman = window.webrender_surfman();
 
@@ -182,7 +190,7 @@ impl App {
                 return;
             }
 
-            if let winit::event::Event::RedrawRequested(_) = e {
+            if let winit::event::Event::RedrawRequested(_) = event {
                 // We need to redraw the window for some reason.
                 trace!("RedrawRequested");
 
@@ -207,20 +215,45 @@ impl App {
             // Handle the event
             let mut consumed = false;
             if let Some(mut minibrowser) = app.minibrowser() {
-                if let winit::event::Event::WindowEvent { ref event, .. } = e {
-                    let response = minibrowser.context.on_event(&event);
-                    if response.repaint {
-                        // Request a winit redraw event, so we can recomposite, update and paint the
-                        // minibrowser, and present the new frame.
-                        window.winit_window().unwrap().request_redraw();
-                    }
+                match event {
+                    winit::event::Event::WindowEvent {
+                        event: WindowEvent::ScaleFactorChanged { scale_factor, .. },
+                        ..
+                    } => {
+                        // Intercept any ScaleFactorChanged events away from EguiGlow::on_event, so
+                        // we can use our own logic for calculating the scale factor and set egui’s
+                        // scale factor to that value manually.
+                        let effective_scale_factor = window.hidpi_factor().get();
+                        info!(
+                            "window scale factor changed to {}, setting scale factor to {}",
+                            scale_factor, effective_scale_factor
+                        );
+                        minibrowser
+                            .context
+                            .egui_ctx
+                            .set_pixels_per_point(effective_scale_factor);
 
-                    // TODO how do we handle the tab key? (see doc for consumed)
-                    consumed = response.consumed;
+                        // Request a winit redraw event, so we can recomposite, update and paint
+                        // the minibrowser, and present the new frame.
+                        window.winit_window().unwrap().request_redraw();
+                    },
+                    winit::event::Event::WindowEvent { ref event, .. } => {
+                        let response = minibrowser.context.on_event(&event);
+                        if response.repaint {
+                            // Request a winit redraw event, so we can recomposite, update and paint
+                            // the minibrowser, and present the new frame.
+                            window.winit_window().unwrap().request_redraw();
+                        }
+
+                        // TODO how do we handle the tab key? (see doc for consumed)
+                        // Note that servo doesn’t yet support tabbing through links and inputs
+                        consumed = response.consumed;
+                    },
+                    _ => {},
                 }
             }
             if !consumed {
-                app.queue_embedder_events_for_winit_event(e);
+                app.queue_embedder_events_for_winit_event(event);
             }
 
             let animating = app.is_animating();
--- a/ports/servoshell/headless_window.rs
+++ b/ports/servoshell/headless_window.rs
@@ -7,7 +7,7 @@
 use std::cell::Cell;
 use std::rc::Rc;
 
-use euclid::{Point2D, Rotation3D, Scale, Size2D, UnknownUnit, Vector3D};
+use euclid::{Length, Point2D, Rotation3D, Scale, Size2D, UnknownUnit, Vector3D};
 use servo::compositing::windowing::{
     AnimationState, EmbedderCoordinates, EmbedderEvent, WindowMethods,
 };
@@ -52,13 +52,6 @@ impl Window {
 
         Rc::new(window)
     }
-
-    fn servo_hidpi_factor(&self) -> Scale<f32, DeviceIndependentPixel, DevicePixel> {
-        match self.device_pixel_ratio_override {
-            Some(override_value) => Scale::new(override_value),
-            _ => Scale::new(1.0),
-        }
-    }
 }
 
 impl WindowPortsMethods for Window {
@@ -74,6 +67,16 @@ impl WindowPortsMethods for Window {
         unsafe { winit::window::WindowId::dummy() }
     }
 
+    fn device_hidpi_factor(&self) -> Scale<f32, DeviceIndependentPixel, DevicePixel> {
+        Scale::new(1.0)
+    }
+
+    fn device_pixel_ratio_override(
+        &self,
+    ) -> Option<Scale<f32, DeviceIndependentPixel, DevicePixel>> {
+        self.device_pixel_ratio_override.map(Scale::new)
+    }
+
     fn page_height(&self) -> f32 {
         let height = self
             .webrender_surfman
@@ -81,7 +84,7 @@ impl WindowPortsMethods for Window {
             .unwrap_or(None)
             .map(|info| info.size.height)
             .unwrap_or(0);
-        let dpr = self.servo_hidpi_factor();
+        let dpr = self.hidpi_factor();
         height as f32 * dpr.get()
     }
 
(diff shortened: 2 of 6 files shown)


ports/servoshell/app.rs
```diff
@@ -208,16 +216,36 @@ impl App {
...
-                    let response = minibrowser.context.on_event(&event);
-                    if response.repaint {
-                        // Request a winit redraw event, so we can recomposite, update and paint the
-                        // minibrowser, and present the new frame.
-                        window.winit_window().unwrap().request_redraw();
-                    }
-
-                    // TODO how do we handle the tab key? (see doc for consumed)
-                    consumed = response.consumed;
+                match e {
```
mrobinson, line 219:
```suggestion
                match event {
```

ports/servoshell/app.rs
```diff
@@ -208,16 +216,36 @@ impl App {
...
+                        // Catch any ScaleFactorChanged events and set egui’s scale factor manually,
+                        // because we have our own logic for calculating the scale factor to use.
+                        let effective_scale_factor = window.hidpi_factor().get();
+                        info!(
+                            "window scale factor changed to {}, setting scale factor to {}",
+                            scale_factor, effective_scale_factor
+                        );
+                        minibrowser
+                            .context
+                            .egui_ctx
```
mrobinson, line 233: Is there a reason that this isn't part of `minibrowser.context.on_event(...)`?  Is it just that this kind of event doesn't return a response? Perhaps returning an Option<...> in that case makes sense.

ports/servoshell/headless_window.rs
```diff
@@ -74,14 +67,21 @@ impl WindowPortsMethods for Window {
         unsafe { winit::window::WindowId::dummy() }
     }
 
+    fn hidpi_factor(&self) -> Scale<f32, DeviceIndependentPixel, DevicePixel> {
+        match self.device_pixel_ratio_override {
+            Some(override_value) => Scale::new(override_value),
+            _ => Scale::new(1.0),
+        }
+    }
+
```
mrobinson, line 76: It seems like the implemetnation of this method is very similar for headed and headless windows. Could this be implemented as a [default method on the trait](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) that uses two non-default methods: `device_pixel_ratio_override()` and `device_pixel_ratio()` (which would return 1.0) for headless windows)?

mrobinson (changes requested):
This looks great. Just a couple comments below.

ports/servoshell/app.rs
delan: done in 04a3def60440c7d9ec32d3f109ef8ab40da9f57c

ports/servoshell/headless_window.rs
delan: headed Window also has a case for `opts::get().output_file` that headless Window lacks, so i think this would be a bit of a leaky/premature abstraction. what do you think?

ports/servoshell/app.rs
mrobinson: Ah, I had forgotten that `egui_glow.rs` is essentially vendored code.

ports/servoshell/headless_window.rs
mrobinson: I think it's fine to move the `opts::get().output_file` case to the default method. It doesn't seem like this behavior should differ between the two types of windows. My suspicion is that the only reason it's only for headed windows is that headless windows default to 1.0 anyway. Moving it to the default method will make it much clearer what happens when you specify both options though.

mrobinson (approved):
Thanks!

sagudev:
let's run try build so I do not need to build it myself to test windows

@bors-servo try=all

sagudev:
I tested on my windows and something is working, but there is no logging and on mixed-dpi setup there is no change (blurred on hi-dpi monitor).

delan:
> I tested on my windows and something is working, but there is no logging and on mixed-dpi setup there is no change (blurred on hi-dpi monitor).

Thanks for testing this! It looks like there’s no logging in the Windows try artifacts, because our Windows [release builds use the Windows subsystem](https://github.com/servo/servo/blob/988e05a68b48c9e744bf49459faf41a1bd9b81d7/ports/servoshell/main.rs#L17-L21). With a debug build and `RUST_LOG=warn,servo::app` and moving the window from a 1x monitor to a 2x monitor, it looks like winit isn’t giving us a ScaleFactorChanged event for some reason, only Moved and Resized:

```
[2023-09-14T10:02:42Z TRACE servo::app] @12.6279403s (+28.6µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: ModifiersChanged(SHIFT) }
[2023-09-14T10:02:42Z TRACE servo::app] @12.6281527s (+212.4µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: KeyboardInput { device_id: DeviceId(DeviceId(0)), input: KeyboardInput { scancode: 42, state: Pressed, virtual_keycode: Some(LShift), modifiers: SHIFT }, is_synthetic: false } }
[2023-09-14T10:02:42Z TRACE servo::app] @12.6288537s (+7.6µs) RedrawRequested(WindowId(WindowId(2689246)))
[2023-09-14T10:02:42Z TRACE servo::app] RedrawRequested
[2023-09-14T10:02:42Z TRACE servo::app] need_recomposite
[2023-09-14T10:02:42Z ERROR egui_glow] GL error, at C:\Users\delan\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
[2023-09-14T10:02:42Z TRACE servo::app] PumpResult::ReadyToPresent
[2023-09-14T10:02:42Z TRACE servo::app] @12.6356575s (+23.2µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: ModifiersChanged(SHIFT | LOGO) }
[2023-09-14T10:02:42Z TRACE servo::app] @12.6357864s (+128.9µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: KeyboardInput { device_id: DeviceId(DeviceId(0)), input: KeyboardInput { scancode: 57435, state: Pressed, virtual_keycode: Some(LWin), modifiers: SHIFT | LOGO }, is_synthetic: false } }
[2023-09-14T10:02:42Z TRACE servo::app] @12.6359375s (+4.2µs) RedrawRequested(WindowId(WindowId(2689246)))
[2023-09-14T10:02:42Z TRACE servo::app] RedrawRequested
[2023-09-14T10:02:42Z TRACE servo::app] need_recomposite
[2023-09-14T10:02:42Z ERROR egui_glow] GL error, at C:\Users\delan\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
[2023-09-14T10:02:42Z TRACE servo::app] PumpResult::ReadyToPresent
[2023-09-14T10:02:43Z TRACE servo::app] @13.0215462s (+13.7µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: Moved(PhysicalPosition { x: 2567, y: 28 }) }
[2023-09-14T10:02:43Z TRACE servo::app] @13.0216566s (+110.4µs) WindowEvent { window_id: WindowId(WindowId(2689246)), event: Resized(PhysicalSize { width: 277, height: 740 }) }
[2023-09-14T10:02:43Z TRACE servo::app] PumpResult::Resize
[2023-09-14T10:02:43Z ERROR egui_glow] GL error, at C:\Users\delan\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
[2023-09-14T10:02:43Z TRACE servo::app] PumpResult::ReadyToPresent
[2023-09-14T10:02:43Z TRACE servo::app] @13.0684782s (+2.7606ms) RedrawRequested(WindowId(WindowId(2689246)))
[2023-09-14T10:02:43Z TRACE servo::app] RedrawRequested
[2023-09-14T10:02:43Z TRACE servo::app] need_recomposite
[2023-09-14T10:02:43Z ERROR egui_glow] GL error, at C:\Users\delan\.cargo\registry\src\github.com-1ecc6299db9ec823\egui_glow-0.22.0\src\painter.rs:284 (FRAMEBUFFER_SRGB): GL_INVALID_ENUM (0x500). Please file a bug at https://github.com/emilk/egui/issues
```

delan:
winit’s source [suggests](https://github.com/rust-windowing/winit/blob/v0.28.6/src/platform_impl/windows/event_loop.rs#L2072-L2276) that it delivers ScaleFactorChanged in response to WM_DPICHANGED, but even with a modified version of [Raymond Chen’s scratch program](https://devblogs.microsoft.com/oldnewthing/20030723-00/?p=43073), it looks like Windows isn’t giving us WM_DPICHANGED when I move the window between monitors with different dpi, because I don’t get a MessageBox:

```c++
// https://stackoverflow.com/q/27613278
#pragma comment (lib, "user32.lib")
#pragma comment (lib, "ole32.lib")
#pragma comment (lib, "comctl32.lib")

// https://gist.github.com/nanoant/dde75805132561140ed5e38f4048f5c1#file-win32directwrite-cpp-L146
#define HANDLE_WM_DPICHANGED(hwnd, wParam, lParam, fn) ((fn)((hwnd), LOWORD(wParam), HIWORD(wParam), *((RECT*)(lParam))), 0L)

#define STRICT
#include <windows.h>
#include <windowsx.h>
#include <ole2.h>
#include <commctrl.h>
#include <shlwapi.h>
HINSTANCE g_hinst;
BOOL
OnCreate(HWND hwnd, LPCREATESTRUCT lpcs)
{
    return TRUE;
}
void
OnDestroy(HWND hwnd)
{
    PostQuitMessage(0);
}
void
PaintContent(HWND hwnd, PAINTSTRUCT *pps)
{
}
void
OnPaint(HWND hwnd)
{
    PAINTSTRUCT ps;
    BeginPaint(hwnd, &ps);
    PaintContent(hwnd, &ps);
    EndPaint(hwnd, &ps);
}
void
OnPrintClient(HWND hwnd, HDC hdc)
{
    PAINTSTRUCT ps;
    ps.hdc = hdc;
    GetClientRect(hwnd, &ps.rcPaint);
    PaintContent(hwnd, &ps);
}
void
OnDpiChanged(HWND hwnd, WORD x, WORD y, RECT suggested)
{
    MessageBoxW(NULL, L"", L"WM_DPICHANGED", 0);
}
/*
 *  Window procedure
 */
LRESULT CALLBACK
WndProc(HWND hwnd, UINT uiMsg, WPARAM wParam, LPARAM lParam)
{
    switch (uiMsg) {
    HANDLE_MSG(hwnd, WM_CREATE, OnCreate);
    HANDLE_MSG(hwnd, WM_SIZE, OnSize);
    HANDLE_MSG(hwnd, WM_DESTROY, OnDestroy);
    HANDLE_MSG(hwnd, WM_PAINT, OnPaint);
    HANDLE_MSG(hwnd, WM_DPICHANGED, OnDpiChanged);
    case WM_PRINTCLIENT: OnPrintClient(hwnd, (HDC)wParam); return 0;
    }
    return DefWindowProc(hwnd, uiMsg, wParam, lParam);
}
BOOL
InitApp(void)
{
    WNDCLASS wc;
    wc.style = 0;
    wc.lpfnWndProc = WndProc;
    wc.cbClsExtra = 0;
    wc.cbWndExtra = 0;
    wc.hInstance = g_hinst;
    wc.hIcon = NULL;
    wc.hCursor = LoadCursor(NULL, IDC_ARROW);
    wc.hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
    wc.lpszMenuName = NULL;
    wc.lpszClassName = TEXT("Scratch");
    if (!RegisterClass(&wc)) return FALSE;
    InitCommonControls();               /* In case we use a common control */
    return TRUE;
}
int WINAPI WinMain(HINSTANCE hinst, HINSTANCE hinstPrev,
                   LPSTR lpCmdLine, int nShowCmd)
{
    MSG msg;
    HWND hwnd;
    g_hinst = hinst;
    if (!InitApp()) return 0;
    if (SUCCEEDED(CoInitialize(NULL))) {/* In case we use COM */
        hwnd = CreateWindow(
            TEXT("Scratch"),                /* Class Name */
            TEXT("Scratch"),                /* Title */
            WS_OVERLAPPEDWINDOW,            /* Style */
            CW_USEDEFAULT, CW_USEDEFAULT,   /* Position */
            CW_USEDEFAULT, CW_USEDEFAULT,   /* Size */
            NULL,                           /* Parent */
            NULL,                           /* No menu */
            hinst,                          /* Instance */
            0);                             /* No special parameters */
        ShowWindow(hwnd, nShowCmd);
        while (GetMessage(&msg, NULL, 0, 0)) {
            TranslateMessage(&msg);
            DispatchMessage(&msg);
        }
        CoUninitialize();
    }
    return 0;
}
```

sagudev:
Maybe we need to tell windows that we are DPI aware application.
https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-dpichanged#remarks

delan:
That did the trick! With d621a90e194bd02ddc9a169bb20ceb50f10b4927:

```
[2023-09-14T10:51:50Z INFO  servo::app] window scale factor changed to 2, setting scale factor to 2
[2023-09-14T10:51:52Z INFO  servo::app] window scale factor changed to 1, setting scale factor to 1
```

delan:
(force push to rebase over #30344)

PR #30344, opened 2023-09-12
[NFC] servoshell: fix euclid units associated with winit geometry

delan:
The headed Window::get_coordinates multiplies screen sizes, window sizes, and window positions by a fake dip-to-dp scale factor to work around the fact that they were incorrectly marked as DeviceIndependentPixel, when winit’s PhysicalSize and PhysicalPosition types should actually be in DevicePixel.

This patch fixes those units and adds helper functions for converting winit types to euclid types, making our geometry calculations simpler and eliminating almost all of the untyped operations.

---

### DIFF
--- a/ports/servoshell/headed_window.rs
+++ b/ports/servoshell/headed_window.rs
@@ -527,9 +527,9 @@ fn winit_position_to_euclid_point<T>(position: PhysicalPosition<T>) -> Point2D<T
 
 impl WindowMethods for Window {
     fn get_coordinates(&self) -> EmbedderCoordinates {
-        let win_size = winit_size_to_euclid_size(self.winit_window.outer_size()).to_i32();
-        let win_origin = self.winit_window.outer_position().unwrap_or_default();
-        let win_origin = winit_position_to_euclid_point(win_origin).to_i32();
+        let window_size = winit_size_to_euclid_size(self.winit_window.outer_size()).to_i32();
+        let window_origin = self.winit_window.outer_position().unwrap_or_default();
+        let window_origin = winit_position_to_euclid_point(window_origin).to_i32();
         let inner_size = winit_size_to_euclid_size(self.winit_window.inner_size()).to_f32();
 
         // Subtract the minibrowser toolbar height if any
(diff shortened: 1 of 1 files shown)


ports/servoshell/headed_window.rs
```diff
@@ -522,33 +517,32 @@ impl WindowPortsMethods for Window {
...
-            .outer_position()
-            .unwrap_or(PhysicalPosition::new(0, 0));
-        let win_size = (Size2D::new(width as f32, height as f32) * dpr).to_i32();
-        let win_origin = (Point2D::new(x as f32, y as f32) * dpr).to_i32();
-        let screen = (self.screen_size.to_f32() * dpr).to_i32();
-
-        let PhysicalSize { width, height } = self.winit_window.inner_size();
+        let win_size = winit_size_to_euclid_size(self.winit_window.outer_size()).to_i32();
+        let win_origin = self.winit_window.outer_position().unwrap_or_default();
+        let win_origin = winit_position_to_euclid_point(win_origin).to_i32();
```
mrobinson, line 532:
```suggestion
        let window_size = winit_size_to_euclid_size(self.winit_window.outer_size()).to_i32();
        let window_origin = self.winit_window.outer_position().unwrap_or_default();
        let window_origin = winit_position_to_euclid_point(win_origin).to_i32();
```

I know this `win_` predates your change, but maybe it's a good moment to clean this up.

ports/servoshell/headed_window.rs
```diff
@@ -522,33 +517,32 @@ impl WindowPortsMethods for Window {
     }
 }
 
+fn winit_size_to_euclid_size<T>(size: PhysicalSize<T>) -> Size2D<T, DevicePixel> {
+    Size2D::new(size.width, size.height)
+}
+
+fn winit_position_to_euclid_point<T>(position: PhysicalPosition<T>) -> Point2D<T, DevicePixel> {
+    Point2D::new(position.x, position.y)
+}
```
mrobinson, line 526: This would be a little bit more idiomatic as an implementation of `From` or `Into`. At the moment, I cannot recall the current rules about implementing these traits on foreign types.

mrobinson (approved):
Nice cleanup. I have a couple nits below, but otherwise this is great.

ports/servoshell/headed_window.rs
delan: both types are foreign, so sadly we can’t do that in servoshell. (we could however do that in euclid, if we added an optional winit feature or something)

ports/servoshell/headed_window.rs
delan: done in 8ba5fbb6da2e6f1d92b2459e33beb6a78a8e453c

PR #30351, opened 2023-09-13
Improve intrinsic sizing and white-space handling around forced line breaks

### LINKED ISSUE #30350: Inline elements with <br> have bigger width than expected
See this simple example:
```html
<div style="border: solid thick magenta; float: left;">Just some text <br> in two lines</div>
<div>floatting</div>
```

Screenshot where the floated element is too wide

The width of the floated element should be smaller, there are some extra space that is not needed.

mrobinson:
This change fixes a few issues:

 - Intrinsic sizing should take into account forced line breaks.
 - Non-preserved white-space should never trigger a line break. This type of white-space can hang off the end of the line and be trimmed later. Hanging white-space is already ignored for the purposes of placing lines among floats.
 - Line breaks should immediately trigger new positioning among floats.

These changes seem a little unrelated, but fixing one of these issues triggers failures due to the other ones. Fixing all of these together leads to test progressions. 

Fixes #30350.

---

- [x] These changes fix #30350.

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -707,19 +707,13 @@ impl InlineFormattingContext {
                                         self.pending_whitespace + advance;
                                     self.pending_whitespace = Length::zero();
                                 } else {
-                                    let white_space =
-                                        text_run.parent_style.get_inherited_text().white_space;
-                                    if white_space.preserve_newlines() {
-                                        let last_byte = text_run
-                                            .text
-                                            .as_bytes()
-                                            .get(run.range.end().to_usize() - 1);
-                                        if last_byte == Some(&b'\n') {
-                                            self.had_non_whitespace_content_yet = true;
-                                            self.forced_line_break();
-                                            self.current_line = ContentSizes::zero();
-                                            continue;
-                                        }
+                                    // If this run is a forced line break, we *must* break the line
+                                    // and start measuring from the inline origin once more.
+                                    if text_run.glyph_run_ends_with_preserved_newline(run) {
+                                        self.had_non_whitespace_content_yet = true;
+                                        self.forced_line_break();
+                                        self.current_line = ContentSizes::zero();
+                                        continue;
                                     }
 
                                     // Discard any leading whitespace in the IFC. This will always be trimmed.
@@ -1122,12 +1116,13 @@ impl IndependentFormattingContext {
         if ifc.new_potential_line_size_causes_line_break(&new_potential_line_size) {
             ifc.finish_current_line_and_reset(layout_context);
 
-            let new_potential_line_size = LogicalVec2 {
+            // Calling `new_potential_line_size_causes_line_break()` here triggers the
+            // new line to be positioned among floats. This should never ask for a line
+            // break because it is the first content on the line.
+            let will_break = ifc.new_potential_line_size_causes_line_break(&LogicalVec2 {
                 inline: size.inline,
                 block: size.block,
-            };
-            let will_break =
-                ifc.new_potential_line_size_causes_line_break(&new_potential_line_size);
+            });
             assert!(!will_break);
         }
 
@@ -1228,6 +1223,23 @@ impl TextRun {
         })
     }
 
+    fn glyph_run_ends_with_preserved_newline(&self, run: &GlyphRun) -> bool {
+        if !run.glyph_store.is_whitespace() {
+            return false;
+        }
+        if !self
+            .parent_style
+            .get_inherited_text()
+            .white_space
+            .preserve_newlines()
+        {
+            return false;
+        }
+
+        let last_byte = self.text.as_bytes().get(run.range.end().to_usize() - 1);
+        last_byte == Some(&b'\n')
+    }
+
     fn layout_into_line_items(
         &self,
         layout_context: &LayoutContext,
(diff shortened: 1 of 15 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -707,6 +707,22 @@ impl InlineFormattingContext {
...
+                                    if white_space.preserve_newlines() {
+                                        let last_byte = text_run
+                                            .text
+                                            .as_bytes()
+                                            .get(run.range.end().to_usize() - 1);
+                                        if last_byte == Some(&b'\n') {
+                                            self.had_non_whitespace_content_yet = true;
+                                            self.current_line.max_content += advance;
+                                            self.forced_line_break();
+                                            self.current_line.max_content = ContentSizes::zero();
```
Loirooriol, line 721: I get an error:

```
    |
721 | ...                   self.current_line.max_content = ContentSizes::zero();
    |                       -----------------------------   ^^^^^^^^^^^^^^^^^^^^ expected `CSSPixelLength`, found `ContentSizes`
    |                       |
    |                       expected due to the type of this binding
```

components/layout_2020/flow/inline.rs
```diff
@@ -707,6 +707,22 @@ impl InlineFormattingContext {
...
+                                    let white_space =
+                                        text_run.parent_style.get_inherited_text().white_space;
+                                    if white_space.preserve_newlines() {
+                                        let last_byte = text_run
+                                            .text
+                                            .as_bytes()
+                                            .get(run.range.end().to_usize() - 1);
+                                        if last_byte == Some(&b'\n') {
+                                            self.had_non_whitespace_content_yet = true;
+                                            self.current_line.max_content += advance;
```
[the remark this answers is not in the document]
mrobinson: I think it isn't necessary though we also need to fix the issue regarding soft line break opportunities before forced breaks. I'll open a bug for that issue soon.

mrobinson:
It seems that fixing this issues tickles another issue we have related to trailing non-preserved spacing before `<br>` elements. I'm trying a new approach here where non-preserved spaces *never* trigger line-breaking. These types of spaces are allowed to hang. They will later be trimmed by white-space handling in `LineItem` layout.

Still I'm worried that this work-around might lead to situations where the starting inline margins, borders, and padding will be on a different line than their starting content. That could be a larger issue and perhaps we should solve that when tackling the similar issue we have with ending inline margins, borders, and padding.

mrobinson:
I need to look at the failure in 'css/CSS2/css1/c5525-fltmrgn-000.xht' but I'm feeling pretty good about this.

Loirooriol:
This is the reduction of the regression:

```html
<div style="font: 15px/1 Ahem; width: 75px; border: solid black; color: magenta">
  x x
</div>
```

| Bad | Good |
| - | - |
|  |  |

components/layout_2020/flow/inline.rs
```diff
@@ -1293,14 +1301,28 @@ impl TextRun {
...
+            // "Then, the entire block is rendered. Inlines are laid out, taking bidi
+            // reordering into account, and wrapping as specified by the text-wrap
+            // property. As each line is laid out,
+            //
+            // > 1. A sequence of collapsible spaces at the beginning of a line is removed."
+            //
+            // This prevents whitespace from being added to the beginning of a line. We could
+            // trim it later, but we don't want it to come into play when determining line
+            // width.
+            if is_whitespace && !white_space.preserve_spaces() && !ifc.current_line.has_content {
```
Loirooriol, line 1314: Could reuse `is_non_preserved_whitespace`

mrobinson:
Okay. I think this is ready for review. Thanks for the feedback!

components/layout_2020/flow/inline.rs
```diff
@@ -1314,13 +1318,56 @@ impl TextRun {
...
-                !ifc.current_line.has_content
-            {
+            let is_non_preserved_whitespace =
+                run.glyph_store.is_whitespace() && !white_space.preserve_spaces();
+            if is_non_preserved_whitespace && !ifc.current_line.has_content {
                 continue;
             }
 
+            let new_advance_from_glyph_run = Length::from(run.glyph_store.total_advance());
+            let new_total_advance = new_advance_from_glyph_run +
```
Loirooriol, line 1328: I wouldn't cal this "advance" if it includes `ifc.current_line.inline_position`. Maybe `new_inline_position`, `inline_end_position` or something.

Loirooriol (approved):

components/layout_2020/flow/inline.rs
```diff
@@ -1205,6 +1223,23 @@ impl TextRun {
         })
     }
 
+    fn glyph_run_ends_with_preserved_newline(&self, run: &GlyphRun) -> bool {
+        if !run.glyph_store.is_whitespace() {
+            return false;
+        }
```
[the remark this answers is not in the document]
mrobinson: I've renamed this to `glyph_run_is_whitespace_ending_with_preserved_newline` to clarify that it doesn't apply to the `TextRun` itself.

components/layout_2020/flow/inline.rs
mrobinson: I've done the following renames:
 - `new_total_advance` -> `new_potential_inline_end_position`
 - `advance_from_text_run` -> `text_run_inline_size`
 - `new_advance_from_glyph_run` -> `advance_from_glyph_run`

mrobinson:
Thanks for the review!

PR #30355, opened 2023-09-14
Cargo.toml cleanup

atouchet:
Mostly fixing some ordering mistakes. The only notable change is removing thin-slice which appears to have been removed from the project in 5b62f66.

---

- [x] These changes do not require tests because they simply clean up a configuration file.


sagudev:
Given that we already have taplo config for formating TOML files, it would be nice if we could do it with taplo if possible. There is `reorder_arrays` that will work on all keys.

mrobinson (approved):

PR #30375, opened 2023-09-15
Fix main frame scrolling in quirks mode after WebRender upgrade

### LINKED ISSUE #30368: WebRender upgrade caused regression in Acid2 Test
It seems this patch https://github.com/servo/servo/pull/30323 caused a regression in Acid2: https://wpt.live/acid/acid2/test.html (it works in https://github.com/servo/servo/tree/revert-webrender)

Screenshot of Servo trying to run the test

Nothing happens when you click *Take The Acid2 Test*, not even browsing directly http://acid2.acidtests.org/#top

mrobinson:
Main frame scrolling after the WebRender upgrade was broken because the
main frame no longer has content size content bounds and quirks mode
leads to a body with zero height.

Fixes #30368.

---

- [x] These changes fix #30368.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -664,7 +664,7 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
                 self.waiting_for_results_of_scroll = true;
 
                 let mut txn = Transaction::new();
-                txn.scroll_node_with_id(point, scroll_id, ScrollClamping::ToContentBounds);
+                txn.scroll_node_with_id(point, scroll_id, ScrollClamping::NoClamping);
                 txn.generate_frame(0);
                 self.webrender_api
                     .send_transaction(self.webrender_document, txn);
--- a/tests/wpt/meta-legacy-layout/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/sticky-content.html.ini
+++ b/tests/wpt/meta-legacy-layout/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/sticky-content.html.ini
@@ -1,0 +1,3 @@
+[sticky-content.html]
+  expected: FAIL
+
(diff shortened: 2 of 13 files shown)


mrobinson:
Still a draft, because I suspect this will change test results and I'd like to update them before this lands.

mrobinson:
@bors-servo try

The new failures are tests that started failing, probably as false passes, after the WebRender upgrade.

components/compositing/compositor.rs
```diff
@@ -664,8 +664,8 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
                 self.waiting_for_results_of_scroll = true;
 
                 let mut txn = Transaction::new();
-                txn.scroll_node_with_id(point, scroll_id, ScrollClamping::ToContentBounds);
-                txn.generate_frame(0);
+                txn.scroll_node_with_id(point, scroll_id, ScrollClamping::NoClamping);
+                txn.generate_frame(200);
```
mukilan, line 668: Is this argument to `generate_frame` arbitrary? If not, maybe we can comment why this value was chosen?

mrobinson: Sorry, this argument is currently unused. I modified it in the process of debugging, but I'll revert this change.

mrobinson: This should be removed now. Thanks for pointing it out.

tests/wpt/meta-legacy-layout/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/sticky-content.html.ini
```diff
@@ -0,0 +1,4 @@
+tests/wpt/meta-legacy-layout/html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/sticky-content.html.ini
```
mukilan, line 1: I think we have a syntax error - the path here needs to be wrapped in `[` `]` to form a valid key. This is the reason the latest try run has failed.

mrobinson: Whoops. I've fixed this. Thanks.

mukilan (approved):

PR #30378, opened 2023-09-16
use newer wix in CI

sagudev:
Here is changelog: https://github.com/wixtoolset/wix3/releases/tag/wix3112rtm

---

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -55,12 +55,15 @@ jobs:
           fetch-depth: 2
       - name: ccache
         uses: hendrikmuhs/ccache-action@v1.2
+      - name: Install wixtoolset
+        run: |
+          choco install wixtoolset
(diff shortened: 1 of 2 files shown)


mrobinson (commented):
I wonder if we can start pulling this in via choco or winget? I think these both allow specifying a version.

sagudev:
> wonder if we can start pulling this in via choco

Well it would be nice if this was part of bootstrap.

sagudev:
It's available via choco: https://community.chocolatey.org/packages/wixtoolset

mrobinson (commented):
I think the addition of wix to the choco configuration is great. The only issue is that bootstrap isn't run for the CI. You could still install wix via choco though, so that the CI is more similar to developer machines.

sagudev:
I forgot that try builds via bors are useless as they use master files.

sagudev:
Try run on windows: https://github.com/sagudev/servo/actions/runs/6207064721/job/16851930976#step:13:1 package step is run successfully; fails on upload as it's on my fork.

mrobinson (approved):
Thanks for this!

Issue #30144, opened 2023-08-21
Allow testing webgpu (but not by default)


sagudev:
Currently webgpu is not tested by main workflow.

We can either:
1. test webgpu by default (not recommended as webgpu is still not finalized spec & it's slow)
2. add wpt-string to runners so we can trigger wpt runs only on specific tests (this would require reviewers to manually trigger run with _webgpu string)

While we could plug second option into workflow, I suggest we start encoding info in commits like its done on mozilla try.

cc @mrobinson

sagudev:
Here is quick JSON schema sketch: 
```jsonc
{
    ["linux", "mac", "windows"]: {
            "wpt": { // default: null; so no wpt run
                    layout: ["none", "2013", "2020", "all"], // default: "all"
                    test-list: string, // default: ""
             }
            "unit-tests": boolean, // default: false
            // always: "upload": true
    }
}
```

mrobinson:
This seems like a reasonable approach, though we should take a holistic look at how we trigger try runs. Maybe this is a conversation to start in Zulip.

sagudev:
This was done as we have `T-webgpu` label now.

PR #30383, opened 2023-09-18
Allow triggering try workfows using labels

mrobinson:
This change adds and alternate method for triggering try changes.
Instead of comments, changes are triggered via applying labels to pull
requests. The action will remove the label from the request and start
the requested jobs.

This will require creating at least a few labels:

 - T-full
 - T-linux-wpt-2013
 - T-linux-wpt-2020
 - T-macos
 - T-windows

More labels can be added as we support more configurations.

The good thing about this change is that try jobs against the actual
branch in the pull request instead of the master branch. This means
that changes to CI can be tested (unlike for comment processing).

One bit caveat with this change is that when adding multiple labels, a
CI job is triggered for each. Only one real build will run for each
label, but whether or more try jobs is triggered is a race condition.
The first CI job to successfully remove the label will actually trigger
the job. If the same job removes two compatible labels, then they can
share a build (for instance two types of WPT linux jobs). If not there
will be two. Note that this is at least as efficient as the current
behavior.

---

- [x] These changes do not require tests because they add a new CI configuration.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -55,15 +55,15 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
-        if: github.event_name != 'issue_comment'
+        if: github.event_name != 'issue_comment' && github.event_name != 'pull_request_target'
         with:
           fetch-depth: 2
       # This is necessary to checkout the pull request if this run was triggered
       # via an `issue_comment` action on a pull request.
       - uses: actions/checkout@v3
-        if: github.event_name == 'issue_comment'
+        if: github.event_name == 'issue_comment' || github.event_name == 'pull_request_target'
         with:
-          ref: refs/pull/${{ github.event.issue.number }}/head
+          ref: refs/pull/${{ github.event.issue.number || github.event.number }}/head
           fetch-depth: 2
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.3
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -48,15 +48,15 @@ jobs:
     runs-on: macos-13
     steps:
       - uses: actions/checkout@v3
-        if: github.event_name != 'issue_comment'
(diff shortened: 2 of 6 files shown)


sagudev:
That gives me idea for https://github.com/servo/servo/issues/30144 we could detect [`A-content/webgpu`](https://github.com/servo/servo/issues?q=label%3AA-content%2Fwebgpu+) label and always run _webgpu suite and normal suite on PR with this label.

mrobinson:
It seems there are some permissions issues that prevent actions for PRs from other forks from modifying the labels applied to a PR. Adding the permissions key to the action doesn't seem to be helping. I'll need to investigate this further.

sagudev:
IIRC apart from adding permission key to workflow file you also need to change repo settings so that actions get r+w permission. I noticed this when working on https://github.com/servo/servo-build-deps/pull/3

mukilan:
I think in order to modify labels of the PR from forks, we'll need to use the workflow trigger 'pull_request_target' instead of 'pull_request'

From the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target):

```
This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.
```

I guess this also means we won't be able to use this to test CI changes in a PR.

mrobinson:
> I guess this also means we won't be able to use this to test CI changes in a PR.

Unfortunately that's what I feared. Even if we cannot test CI changes I guess this offers some other benefits over the comment path.

mrobinson:
Hrm. I think the right thing to do here is to create a fine-grained access token that only allows changing the labels on pull requests. The alternative is to build and run code from PRs with full permissions.

mrobinson:
This is ready to review. I'm happy to walk through this with someone. I am testing this on my own git repository because with the `pull_request_target` action the workflow files are taken from the main branch.

mukilan (approved):

PR #30386, opened 2023-09-19
Implement recreate_context for WebGPU canvas and fix handle_wgpu_msg crash

sagudev:
This makes wgpu examples work.

~~Depends on https://github.com/servo/servo/pull/30406~~. DONE

### Current state of webgpu examples

Works on reload:
- [x] https://mdn.github.io/dom-examples/webgpu-compute-demo/
- [x] https://mdn.github.io/dom-examples/webgpu-render-demo/ (works partially: only blue background; same is in FF so it's probably wgpu problem)
- [x] [Conway's Game of Life](https://sagudev.github.io/briefcase/first-webgpu/9-endgame.html) fixed for naga from https://codelabs.developers.google.com/your-first-webgpu-app

Does not work on reload (but works on freshly opened servo; see #30417):
- [x] https://wgpu.rs/examples-gpu/?example=cube
- [x] https://wgpu.rs/examples-gpu/?example=hello-compute
- [x] https://wgpu.rs/examples-gpu/?example=hello-triangle
- [x] https://wgpu.rs/examples-gpu/?example=mipmap
- [x] https://wgpu.rs/examples-gpu/?example=shadow

---

- [x] There are tests for these changes in examples and webgpu-cts

### DIFF
--- a/components/script/dom/gpucanvascontext.rs
+++ b/components/script/dom/gpucanvascontext.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use std::cell::Cell;
+use std::cell::{Cell, RefCell};
 
 use arrayvec::ArrayVec;
 use dom_struct::dom_struct;
@@ -103,14 +103,17 @@ pub struct GPUCanvasContext {
     #[ignore_malloc_size_of = "channels are hard"]
     #[no_trace]
     channel: WebGPU,
-    /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas>
     canvas: HTMLCanvasElementOrOffscreenCanvas,
+    #[ignore_malloc_size_of = "manual writing is hard"]
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configuration-slot>
+    config: RefCell<Option<GPUCanvasConfiguration>>,
     // TODO: can we have wgpu surface that is hw accelerated inside wr ...
     #[ignore_malloc_size_of = "Defined in webrender"]
     #[no_trace]
     webrender_image: Cell<Option<webrender_api::ImageKey>>,
     context_id: WebGPUContextId,
-    /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot>
     texture: MutNullableDom<GPUTexture>,
 }
 
@@ -128,6 +131,7 @@ impl GPUCanvasContext {
             webrender_image: Cell::new(None),
             context_id: WebGPUContextId(external_id.0),
             texture: MutNullableDom::default(),
+            config: RefCell::default(),
         }
     }
 
@@ -199,6 +203,12 @@ impl GPUCanvasContext {
             HTMLCanvasElementOrOffscreenCanvas::OffscreenCanvas(canvas) => canvas.get_size(),
         }
     }
+
+    pub fn recreate(&self) {
+        if let Some(config) = self.config.take() {
+            self.Configure(&config)
+        }
+    }
 }
 
 impl LayoutCanvasRenderingContextHelpers for LayoutDom<'_, GPUCanvasContext> {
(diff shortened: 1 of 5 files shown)


mrobinson:
> Maybe I am doing something wrong or we do not have those media query implemented.

If those tests use iframes to trigger media query changes, note that all iframe media query tests are incredibly flaky and brittle because of Servo's loose implementation of iframe loading order.

components/script/dom/gpucanvascontext.rs
```diff
@@ -105,6 +105,9 @@ pub struct GPUCanvasContext {
     channel: WebGPU,
     /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas
     canvas: HTMLCanvasElementOrOffscreenCanvas,
+    #[ignore_malloc_size_of = "manual writing is hard"]
+    /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configuration-slot
```
mrobinson, line 109: To prevent a rustdoc warning: 

```suggestion
    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configuration-slot>
```

sagudev:
When based on #30406 there is no error on wgpu examples. But when refreshed examples do not work (also non-ui wgpu examples), maybe that's has something to do with wasm loading as pure JS examples work normally on refresh too.

sagudev:
TODO: How is nop resize (old size = new size) handled.

sagudev:
Resize canvas still crash, but that is not problem with `recreate_context`.

sagudev:
Try run: https://github.com/sagudev/servo/actions/runs/6339107135

components/script/script_thread.rs
```diff
@@ -2102,6 +2102,7 @@ impl ScriptThread {
                 result,
             } => {
                 let global = self.documents.borrow().find_global(pipeline_id).unwrap();
+                let _ac = enter_realm(&*global);
```
mrobinson, line 2105: I'm not sure that `ac` is adding anything here...

```suggestion
                let _ = enter_realm(&*global);
```

components/script/dom/gpucanvascontext.rs
```diff
@@ -103,14 +103,17 @@ pub struct GPUCanvasContext {
     #[ignore_malloc_size_of = "channels are hard"]
     #[no_trace]
     channel: WebGPU,
-    /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-canvas>
     canvas: HTMLCanvasElementOrOffscreenCanvas,
+    #[ignore_malloc_size_of = "manual writing is hard"]
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-configuration-slot>
+    config: RefCell<Option<GPUCanvasConfiguration>>,
```
mrobinson, line 110: Is the configuration ever read other than to know whether to reconfigure the context? Why not just store a bool?

components/script/dom/gpucanvascontext.rs
```diff
@@ -103,14 +103,17 @@ pub struct GPUCanvasContext {
...
     // TODO: can we have wgpu surface that is hw accelerated inside wr ...
     #[ignore_malloc_size_of = "Defined in webrender"]
     #[no_trace]
     webrender_image: Cell<Option<webrender_api::ImageKey>>,
     context_id: WebGPUContextId,
-    /// https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpucanvascontext-currenttexture-slot>
```
mrobinson, line 116: :+1:

components/script/dom/gpucanvascontext.rs
```diff
@@ -199,6 +203,12 @@ impl GPUCanvasContext {
             HTMLCanvasElementOrOffscreenCanvas::OffscreenCanvas(canvas) => canvas.get_size(),
         }
     }
+
+    pub fn recreate(&self) {
+        if let Some(config) = self.config.take() {
+            self.Configure(&config)
+        }
```
mrobinson, line 210: Why isn't the call `Unconfigure()` here instead of in `Configure()`?

mrobinson (changes requested):

components/script/script_thread.rs
sagudev: `let _` is not same as `let _x` see here: https://github.com/rust-lang/rust/issues/40096

`let _` does drop immediately (resolves must use warning), while `let _x` drops at the end. Also `let _ac` is also used in other servo code.

components/script/dom/gpucanvascontext.rs
sagudev: That's how it was done in mozilla, but I will recheck specs tomorrow.

sagudev:
I will need to rethink this, so I spitted off crash fix.

PR #30387, opened 2023-09-19
Update to syn 2 where possible

sagudev:
I also decoupled mozilla vendored crates from workspace syn version.

---

- [x] There are tests for these changes (compiler is test)

### DIFF
--- a/components/config_plugins/lib.rs
+++ b/components/config_plugins/lib.rs
@@ -13,10 +13,7 @@ use proc_macro2::{Span, TokenStream};
 use quote::*;
 use syn::parse::Result;
 use syn::spanned::Spanned;
-use syn::{
-    parse_macro_input, Attribute, Ident, Lit, LitStr, Meta, MetaList, MetaNameValue, NestedMeta,
-    Path,
-};
+use syn::{parse_macro_input, Attribute, Ident, LitStr, Path};
 
 mod parse;
 use parse::*;
@@ -199,23 +196,32 @@ impl Field {
 }
 
 fn attr_to_pref_name(attr: &Attribute) -> Option<LitStr> {
-    attr.parse_meta().ok().and_then(|meta| {
-        if let Meta::List(MetaList { path, nested, .. }) = meta {
-            if path.is_ident("serde") {
-                if let Some(NestedMeta::Meta(Meta::NameValue(MetaNameValue {
-                    ref path,
-                    lit: Lit::Str(val),
-                    ..
-                }))) = nested.iter().next()
-                {
-                    if path.is_ident("rename") {
-                        return Some(val.clone());
-                    }
-                }
+    if attr.path().is_ident("serde") {
+        let mut res = None;
+        // if this fails result is still None
+        let _ = attr.parse_nested_meta(|meta| {
+            if meta.path.is_ident("rename") {
+                res = Some(meta.value()?.parse()?);
             }
-        }
+            Ok(())
+        });
+        res
+    } else {
         None
-    })
+    }
+    /*let mut res = None;
+    attr.parse_nested_meta(|meta| {
+        if meta.path.is_ident("serde") {
+            meta.parse_nested_meta(|meta| {
+                if meta.path.is_ident("rename") {
+                    res = Some(meta.value()?.parse()?);
+                }
+                Ok(())
+            })?;
+        }
+        Ok(())
+    }).unwrap();
+    return res;*/
 }
 
 fn err<S: Spanned>(s: S, msg: &str) -> syn::Error {
--- a/components/config_plugins/lib.rs
+++ b/components/config_plugins/lib.rs
@@ -197,31 +197,18 @@ impl Field {
 
 fn attr_to_pref_name(attr: &Attribute) -> Option<LitStr> {
     if attr.path().is_ident("serde") {
-        let mut res = None;
-        // if this fails result is still None
+        // If `parse_nested_meta()` fails, `result` will remain None.
+        let mut result = None;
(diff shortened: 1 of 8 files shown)


components/config_plugins/lib.rs
```diff
@@ -199,23 +196,32 @@ impl Field {
...
+            meta.parse_nested_meta(|meta| {
+                if meta.path.is_ident("rename") {
+                    res = Some(meta.value()?.parse()?);
+                }
+                Ok(())
+            })?;
+        }
+        Ok(())
+    }).unwrap();
+    return res;*/
```
mrobinson, line 224: Is there a reason this code is commented out but still here?

components/jstraceable_derive/lib.rs
```diff
@@ -149,21 +149,14 @@ fn js_traceable_derive(s: synstructure::Structure) -> proc_macro2::TokenStream {
...
-                    // if reason provided we can skip JSTraceable check
-                    if path.is_ident("no_trace") {
-                        return None;
-                    }
-                },
+            if attr.path().is_ident("no_trace") {
+                // if reason is not provided we do JSTraceable check
+                if attr.meta.require_name_value().is_err() {
+                    asserts.extend(assert_not_impl_traceable(&binding.ast().ty));
+                }
```
mrobinson, line 156:
```suggestion
            if attr.path().is_ident("no_trace") {
                // Do a `JSTraceable` check unless a reason is provided with `no_check`.
                if attr.meta.require_name_value().is_err() {
                    asserts.extend(assert_not_impl_traceable(&binding.ast().ty));
                }
```

It's a bit odd to trigger an error here. Could this be:

```rust
if attr.meta != Meta::NameValue() {
 ...
}
```
?

I think avoiding the error makes things a bit clearer.

mrobinson (changes requested):
Great stuff. Hopefully we can update the Mozilla crates soon.

components/config_plugins/lib.rs
```diff
@@ -199,23 +196,32 @@ impl Field {
...
-                }))) = nested.iter().next()
-                {
-                    if path.is_ident("rename") {
-                        return Some(val.clone());
-                    }
-                }
+    if attr.path().is_ident("serde") {
+        let mut res = None;
+        // if this fails result is still None
+        let _ = attr.parse_nested_meta(|meta| {
```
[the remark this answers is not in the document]
sagudev: Resut typi is marekd with must_use in std, so yes it is there is a warning without let _ binding.

components/config_plugins/lib.rs
sagudev: That should not be here.

components/jstraceable_derive/lib.rs
sagudev: matches! macro should do the job.

components/jstraceable_derive/lib.rs
```diff
@@ -149,21 +149,15 @@ fn js_traceable_derive(s: synstructure::Structure) -> proc_macro2::TokenStream {
...
-                    }
-                },
-                syn::Meta::NameValue(syn::MetaNameValue { ref path, .. }) => {
-                    // if reason provided we can skip JSTraceable check
-                    if path.is_ident("no_trace") {
-                        return None;
-                    }
-                },
+            if attr.path().is_ident("no_trace") {
+                // if no reason is provided to `no_trace`
```
mrobinson, line 153: Do you mind included the slightly more verbose comment here. I had to do a little bit of hunting to understand this code.

```rust
 // Do a `JSTraceable` check unless a reason is provided with `no_check`.
```

components/jstraceable_derive/lib.rs
```diff
@@ -149,21 +149,15 @@ fn js_traceable_derive(s: synstructure::Structure) -> proc_macro2::TokenStream {
...
+                if !matches!(attr.meta, syn::Meta::NameValue(_)) {
+                    // assert that binded type does not implement traceable
```
mrobinson, line 155:
```suggestion
                    // Assert that the bound type does not implement traceable.
```

components/jstraceable_derive/lib.rs
```diff
@@ -149,21 +149,15 @@ fn js_traceable_derive(s: synstructure::Structure) -> proc_macro2::TokenStream {
...
-                },
-                syn::Meta::NameValue(syn::MetaNameValue { ref path, .. }) => {
-                    // if reason provided we can skip JSTraceable check
-                    if path.is_ident("no_trace") {
-                        return None;
-                    }
-                },
+            if attr.path().is_ident("no_trace") {
+                // if no reason is provided to `no_trace`
+                if !matches!(attr.meta, syn::Meta::NameValue(_)) {
```
mrobinson, line 154: Nice. `matches!()` works nicely here.

mrobinson (approved):
Just a couple english nits below, so feel free to update and send this to the merge queue. Thanks!

components/jstraceable_derive/lib.rs
sagudev: Actually I spit this comment in two (se next commented line) so it makes more sense as a whole then separate. There is no `no_check` what I wanted to say is that you can put `#[no_trace]` and this will make sure that the type of the field is not JSTraceble, or if you put `#[no_trace="I know this ..."]` then we will not do such check.

components/jstraceable_derive/lib.rs
sagudev: I am not sure this is called bound type. I might be wrong.

To be clear this is type of field in struct or wrapped type of enum variant.

PR #30390, opened 2023-09-20
Fix webgl regression after WebRender upgrade

### LINKED ISSUE #30367: WebRender upgrade caused a regression in WebGL example TWGL
It seems this patch https://github.com/servo/servo/pull/30323 caused a regression in WebGL example TWGL doesn't run: https://demo.servo.org/experiments/twgl-tunnel/ (again it works revering the change)

The test gets stopped after showing the first image.

WebGL example TWGL stopped

If you resize the window you get some new frames.

WebGL example TWGL with some new image on a smaller window

mukilan:
Sending WebRender an explicit ResourceUpdate message after every buffer swap invalidates the tile caches that depend on the image resource created for the webgl canvas.

The previous WebRender code that Servo was using might have worked with WebGL only accidentally since picture caching was disabled by [default][1] and so the tiles were not cached between frames.

Fixes #30367

[1]: https://github.com/servo/servo/blob/c385b3c9737c17d59cb02e520c3b68b232cb6497/third_party/webrender/webrender/src/renderer.rs#L7073

---

- [x] These changes fix #30367 (GitHub issue number if applicable)
- [x] These changes might require tests to catch future regression but this PR doesn't include any. Maybe those can be a separate effort since we want to fix the WebGL demos quickly and look into a proper WebGL test suite later.

### DIFF
--- a/components/canvas/webgl_thread.rs
+++ b/components/canvas/webgl_thread.rs
@@ -684,25 +684,13 @@ impl WebGLThread {
 
         // Reset framebuffer bindings as appropriate.
         framebuffer_rebinding_info.apply(&self.device, &data.ctx, &*data.gl);
+        debug_assert_eq!(data.gl.get_error(), gl::NO_ERROR);
 
-        // Update WR image if needed.
-        let info = self.cached_context_info.get_mut(&context_id).unwrap();
         let has_alpha = data
             .state
             .requested_flags
             .contains(ContextAttributeFlags::ALPHA);
-        let image_buffer_kind = current_wr_image_buffer_kind(&self.device);
-        Self::update_wr_external_image(
-            &mut self.webrender_api,
-            self.webrender_doc,
-            size.to_i32(),
-            has_alpha,
-            context_id,
-            info.image_key,
-            image_buffer_kind,
-        );
-
-        debug_assert_eq!(data.gl.get_error(), gl::NO_ERROR);
+        self.update_wr_image_for_context(context_id, size.to_i32(), has_alpha);
 
         Ok(())
     }
@@ -818,6 +806,7 @@ impl WebGLThread {
             debug_assert_eq!(data.gl.get_error(), gl::NO_ERROR);
 
             let SurfaceInfo {
+                size,
                 framebuffer_object,
                 id,
                 ..
@@ -830,6 +819,12 @@ impl WebGLThread {
                 "... rebound framebuffer {}, new back buffer surface is {:?}",
                 framebuffer_object, id
             );
+
+            let has_alpha = data
+                .state
+                .requested_flags
+                .contains(ContextAttributeFlags::ALPHA);
+            self.update_wr_image_for_context(context_id, size, has_alpha);
         }
 
         #[allow(unused)]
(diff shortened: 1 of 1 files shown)


delan (approved):
Great find! I believe this is exactly what we were missing, because the WebGPU counterpart, Web­GPU­Request­::Swap­Chain­Present, kicks off a Update­Web­Render­Data event that in turn sends an Update­Image transaction. WebGL only did so when the context was resized.

components/canvas/webgl_thread.rs
```diff
@@ -830,6 +831,25 @@ impl WebGLThread {
                 "... rebound framebuffer {}, new back buffer surface is {:?}",
                 framebuffer_object, id
             );
+
+            // Tell Webrender to invalidate the caches for this image
```
mrobinson, line 835: Sorry for this huge nit:

```suggestion
            // Tell WebRender to invalidate the caches for this image.
```

components/canvas/webgl_thread.rs
```diff
@@ -830,6 +831,25 @@ impl WebGLThread {
...
+                &mut self.webrender_api,
+                self.webrender_doc,
+                size.to_i32(),
+                has_alpha,
+                context_id,
+                info.image_key,
+                image_buffer_kind,
+            );
+
+            debug_assert_eq!(data.gl.get_error(), gl::NO_ERROR);
```
mrobinson, line 852: Is this new assertion related?

mrobinson (approved):
Great find @mukilan!

components/canvas/webgl_thread.rs
mukilan: It was a lot of work, but I've addressed this huge nit in [ec5c9df](https://github.com/servo/servo/pull/30390/commits/ec5c9df530355d1d406a497d9e4c9b538213f18f) :rofl:

components/canvas/webgl_thread.rs
mukilan: No, you are correct. There are no new OpenGL calls before this, so this assertion is not needed. Removed in [ec5c9df](https://github.com/servo/servo/pull/30390/commits/ec5c9df530355d1d406a497d9e4c9b538213f18f)

mukilan:
Thanks for the review, @mrobinson and @delan! I've addressed the review comments in [ec5c9df](https://github.com/servo/servo/pull/30390/commits/ec5c9df530355d1d406a497d9e4c9b538213f18f). I also took the opportunity to refactor the logic shared between resize and swap flows.

mrobinson (approved):
Thanks so much for this!

components/canvas/webgl_thread.rs
```diff
@@ -908,22 +903,23 @@ impl WebGLThread {
...
+        &mut self,
         context_id: WebGLContextId,
-        image_key: ImageKey,
-        image_buffer_kind: ImageBufferKind,
+        size: Size2D<i32>,
+        has_alpha: bool,
     ) {
-        let descriptor = Self::image_descriptor(size, alpha);
-        let data = Self::external_image_data(context_id, image_buffer_kind);
+        let info = self.cached_context_info.get_mut(&context_id).unwrap();
```
mrobinson, line 914: Hrm. Does this need to be `get_mut(...)` or can it simply be `get(...)`?

mrobinson (approved):
Thanks for this fix!

components/canvas/webgl_thread.rs
mukilan: Good catch! We can just use 'get'. Fixed in  bce77ef

PR #30394, opened 2023-09-20
Update dependencies held back from dependabot

mrobinson:
The dependabot configuration contained some number of ignored versions
that date from when GitHub dependabot support was added. These versions
fall into three categories:

 - Versions that are older or the same version as the version we use.
 - Versions that are newer than the versions we use.
 - Versions for crates we no longer depend on.

This change removes all ignored versions and updates dependencies in
the second category.

---


tests/wpt/meta-legacy-layout/encoding/textdecoder-labels.any.js.ini
mukilan, line 1: Are these new passes because encoding_rs in now upgraded and has some bugfix relevant to these tests?

mukilan (approved):
Nice!

tests/wpt/meta-legacy-layout/encoding/textdecoder-labels.any.js.ini
mrobinson: I suspect that's the case here, but I'm not certain.

PR #30406, opened 2023-09-22
Media query resolution

sagudev:
We have all data needed for this already present, so it was just matter of exposing it.

You can manually test it with https://sagudev.github.io/briefcase/mediaq-resolution.html and zooming on site.

---

### DIFF
--- a/components/style/servo/media_queries.rs
+++ b/components/style/servo/media_queries.rs
@@ -12,6 +12,7 @@ use crate::media_queries::MediaType;
 use crate::properties::ComputedValues;
 use crate::values::computed::CSSPixelLength;
 use crate::values::computed::Context;
+use crate::values::computed::Resolution;
 use crate::values::specified::font::FONT_MEDIUM_PX;
 use crate::values::specified::ViewportVariant;
 use crate::values::KeyframesName;
@@ -252,9 +253,19 @@ fn eval_scan(_: &Context, _: Option<Scan>) -> bool {
     false
 }
 
+/// https://drafts.csswg.org/mediaqueries-4/#resolution
+fn eval_resolution(context: &Context) -> Resolution {
+    Resolution::from_dppx(context.device().device_pixel_ratio.0)
+}
+
+/// https://compat.spec.whatwg.org/#css-media-queries-webkit-device-pixel-ratio
+fn eval_device_pixel_ratio(context: &Context) -> f32 {
+    eval_resolution(context).dppx()
+}
+
 lazy_static! {
     /// A list with all the media features that Servo supports.
-    pub static ref MEDIA_FEATURES: [QueryFeatureDescription; 2] = [
+    pub static ref MEDIA_FEATURES: [QueryFeatureDescription; 5] = [
         feature!(
             atom!("width"),
             AllowsRanges::Yes,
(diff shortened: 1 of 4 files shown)


jdm (approved):
Nice!

components/style/servo/media_queries.rs
```diff
@@ -267,5 +278,23 @@ lazy_static! {
...
+            FeatureFlags::empty(),
+        ),
+        feature!(
+            atom!("device-pixel-ratio"),
+            AllowsRanges::Yes,
+            Evaluator::Float(eval_device_pixel_ratio),
+            FeatureFlags::WEBKIT_PREFIX,
+        ),
+        feature!(
+            atom!("-moz-device-pixel-ratio"),
```
Loirooriol, line 294: The compat spec has ` -webkit-device-pixel-ratio` and not `-moz-device-pixel-ratio`, so maybe it would be better to add the former instead? Not sure

components/style/servo/media_queries.rs
```diff
@@ -252,9 +253,19 @@ fn eval_scan(_: &Context, _: Option<Scan>) -> bool {
...
 }
 
+/// https://drafts.csswg.org/mediaqueries-4/#resolution
+fn eval_resolution(context: &Context) -> Resolution {
+    Resolution::from_dppx(context.device().device_pixel_ratio.0)
+}
+
+/// https://compat.spec.whatwg.org/#css-media-queries-webkit-device-pixel-ratio
+fn eval_device_pixel_ratio(context: &Context) -> f32 {
+    eval_resolution(context).dppx()
```
Loirooriol, line 263: Seems a bit strange to make `eval_resolution` wrap the `f32` into a `Resolution` just to immediately unwrap it. Might make more sense to swap this and make `eval_resolution` call `eval_device_pixel_ratio`:

```rs
/// https://drafts.csswg.org/mediaqueries-4/#resolution
fn eval_resolution(context: &Context) -> Resolution {
    Resolution::from_dppx(eval_device_pixel_ratio(context))
}

/// https://compat.spec.whatwg.org/#css-media-queries-webkit-device-pixel-ratio
fn eval_device_pixel_ratio(context: &Context) -> f32 {
    context.device().device_pixel_ratio.0
}
```

components/style/servo/media_queries.rs
sagudev: I copied that from mozilla: https://searchfox.org/mozilla-central/source/servo/components/style/gecko/media_features.rs#751 and webkit version is also added if I understand `WEBKIT_PREFIX` correctly (that one is also copied from mozilla).

components/style/servo/media_queries.rs
sagudev: I was thinking about this, but the way it's written now aligns better with spec:
> [-webkit-device-pixel-ratio](https://compat.spec.whatwg.org/#descdef-media--webkit-device-pixel-ratio) must be treated as an alias of the [resolution](https://drafts.csswg.org/mediaqueries-4/#descdef-media-resolution) range type [media feature](https://drafts.csswg.org/mediaqueries-4/#media-feature), with its value interpreted as a [dppx](https://drafts.csswg.org/css-values-3/#dppx) unit.

and mozilla's implemention: https://searchfox.org/mozilla-central/source/servo/components/style/gecko/media_features.rs#66

Loirooriol (approved):
OK, fine, thanks

Said-Barokah:
- - > > - **@~~@pointlessone ~~ **.

PR #30407, opened 2023-09-22
fix WebGPU runtime dependencies on linux

delan:
WebGPU on Linux relies on libvulkan.so.1, so this patch adds vulkan-loader or equivalent to:

* README.md § Runtime dependencies
* LD_LIBRARY_PATH in etc/shell.nix
* APT_PKGS in linux.py, and [wiki § Debian](https://github.com/servo/servo/wiki/Building#debian-based-distributions)
* DNF_PKGS in linux.py, and [wiki § Fedora](https://github.com/servo/servo/wiki/Building#fedora)
* XBPS_PKGS in linux.py, and [wiki § Void Linux](https://github.com/servo/servo/wiki/Building#void-linux)
* [wiki § Arch](https://github.com/servo/servo/wiki/Building#arch)
* [wiki § openSUSE](https://github.com/servo/servo/wiki/Building#opensuse)
* [wiki § Gentoo](https://github.com/servo/servo/wiki/Building#gentoo)

I’ve also updated [the wiki page](https://github.com/servo/servo/wiki/Building) to include other changes that were made in APT_PKGS + DNF_PKGS + XBPS_PKGS but not yet synced to the wiki. The syncing process is tedious to do by hand, so I’ve also added instructions to make future syncing as easy as possible. **[Click here](https://github.com/servo/servo/wiki/Building/_compare/891bfc880baa8ed16d6f982a624b1d297fc764b4...4f5f974bf1008b0136ac52d43136dd729fbdd5e9) to see my wiki changes.**

It may be worth replacing all of these with Nix someday, since Nix can run on any Linux distro.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,11 @@ Run Servo with the command:
 
 * `GStreamer` >=1.16
 * `gst-plugins-bad` >=1.16
-
+* `libXcursor`
+* `libXrandr`
+* `libXi`
+* `libxkbcommon`
+* `vulkan-loader`
 
 ## Developing
 
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -44,8 +44,15 @@ clangStdenv.mkDerivation rec {
   # Enable colored cargo and rustc output
   TERMINFO = "${ncurses.out}/share/terminfo";
 
-  # Fix missing libraries errors (those libraries aren't linked against, so we need to dynamically supply them)
-  LD_LIBRARY_PATH = lib.makeLibraryPath [ xorg.libXcursor xorg.libXrandr xorg.libXi libxkbcommon ];
+  # Provide libraries that aren’t linked against but somehow required
+  LD_LIBRARY_PATH = lib.makeLibraryPath [
+    # Fixes missing library errors
+    xorg.libXcursor xorg.libXrandr xorg.libXi libxkbcommon
+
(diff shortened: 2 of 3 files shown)


mukilan (approved):
Nice! Specifying instructions for updating the wiki inline with the code is a great idea!

sagudev:
I remember when I was working on https://github.com/servo/servo/pull/29795 I didn't want to add this to bootstrap, as webgpu support is optional, but it does make sense to have it in for more seamless experience. I am just not 100% for ubuntu packages, given that I needed to add [this line to CI](https://github.com/servo/servo/commit/71e0372ac18709da66e581cf4a777a7cc5d4cb92#diff-9690642e8f4bd356a155cf09f0c12642697f30f033dc38bc5a7f078154ae38e8R58) in wgpu bump. Did you test ubuntu packages with CI?

delan:
I’ve only tested NixOS, but I guess it depends on whether anyone would need to install those manually outside of CI. Does Ubuntu desktop install them by default?

sagudev:
> I’ve only tested NixOS, but I guess it depends on whether anyone would need to install those manually outside of CI, unless we wanna add them just to be safe. Does Ubuntu desktop install them by default?

Idk, but I do know I have to install something on my machine, but I do not recall what.

PR #30415, opened 2023-09-22
Dedupe ron and base64

fabricedesre:
`webdriver` was the last dependency preventing the base64 upgrade. The 0.49 version has a few new apis that we don't implement yet.

For `ron`, upstream WebRender already switched so this is future proof.

---

- [X] These changes do not require tests because they simply update dependencies.

### DIFF
--- a/components/webdriver_server/capabilities.rs
+++ b/components/webdriver_server/capabilities.rs
@@ -81,6 +81,41 @@ impl BrowserCapabilities for ServoCapabilities {
     ) -> Result<bool, WebDriverError> {
         todo!()
     }
+
+    fn webauthn_virtual_authenticators(
+        &mut self,
+        _: &serde_json::Map<std::string::String, Value>,
+    ) -> Result<bool, WebDriverError> {
+        todo!()
+    }
+
+    fn webauthn_extension_uvm(
+        &mut self,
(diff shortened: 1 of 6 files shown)


components/webdriver_server/capabilities.rs
```diff
@@ -81,6 +81,41 @@ impl BrowserCapabilities for ServoCapabilities {
...
+        _: &serde_json::Map<std::string::String, Value>,
+    ) -> Result<bool, WebDriverError> {
+        todo!()
+    }
+
+    fn webauthn_extension_prf(
+        &mut self,
+        _: &serde_json::Map<std::string::String, Value>,
+    ) -> Result<bool, WebDriverError> {
+        todo!()
```
mrobinson, line 103: Won't this todo cause any test using these functions to panic?

mrobinson (approved):

components/webdriver_server/capabilities.rs
fabricedesre: We already have `fn web_socket_url()` that has a similar implementation and that doesn't seem to be causing issues so far. So I guess they are not exercised by current tests and we'll have to mark them as failing if that happens.

PR #30420, opened 2023-09-25
GStreamer default media backend

sagudev:
Try run: https://github.com/sagudev/servo/actions/runs/6289052004

---

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -52,6 +52,18 @@ class BuildType(Enum):
     RELEASE = 2
 
 
+class MediaStack(Enum):
+    """ Media stack """
+    DUMMY = 1
+    GSTREAMER = 2
+
+    def is_dummy(self) -> bool:
+        return self == MediaStack.DUMMY
+
+    def is_gstreamer(self) -> bool:
+        return self == MediaStack.GSTREAMER
+
+
 @contextlib.contextmanager
 def cd(new_path):
     """Context manager for changing the current working directory"""
@@ -211,6 +223,7 @@ class CommandBase(object):
     def __init__(self, context):
         self.context = context
         self.features = []
+        self.media_stack: Optional[MediaStack] = None
         self.cross_compile_target = None
         self.is_android_build = False
         self.target_path = util.get_target_dir()
@@ -833,9 +846,10 @@ class CommandBase(object):
            platform and the value of the '--media-stack' command-line argument.
            The chosen media stack is written into the `features` instance variable."""
         if not media_stack:
+            if self.media_stack is not None:
+                return  # already configured
             if self.config["build"]["media-stack"] != "auto":
-                media_stack = self.config["build"]["media-stack"]
-                assert media_stack
+                media_stack = self.config["build"]["media-stack"].lower()
             elif (
                 not self.cross_compile_target
                 or ("armv7" in self.cross_compile_target and self.is_android_build)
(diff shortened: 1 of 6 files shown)


python/servo/command_base.py
```diff
@@ -844,8 +858,15 @@ def configure_media_stack(self, media_stack: Optional[str]):
...
-            self.features += ["media-" + media_stack]
+
+        assert media_stack
+        media_stack = media_stack.lower()
+        if media_stack == "gstreamer":
+            self.media_stack = MediaStack.GSTREAMER
+        elif media_stack == "dummy":
+            self.media_stack = MediaStack.DUMMY
+        else:
+            raise RuntimeError(f"Wrong media stack {media_stack}")
```
mrobinson, line 869: Just a small nit here to make this error a bit more comprehensible:

```suggestion
            raise RuntimeError(f"Got unknown media stack {media_stack}")
```

I would also move this up to the parsing of the configuration value so that we avoid converting the serialized configuration values to a string first. Essentially:

```python
    if self.config["build"]["media-stack"] != "auto":
        ...convert from string...
    elif (
                not self.cross_compile_target
                or ("armv7" in self.cross_compile_target and self.is_android_build)
                or "x86_64" in self.cross_compile_target
            ):
                media_stack = MediaStack.GSTREAMER
            else:
                media_stack = MediaStack.DUMMY
```

python/servo/command_base.py
```diff
@@ -833,9 +846,10 @@ def configure_media_stack(self, media_stack: Optional[str]):
            platform and the value of the '--media-stack' command-line argument.
            The chosen media stack is written into the `features` instance variable."""
         if not media_stack:
+            if self.media_stack is not None:
+                return  # already configured
```
mrobinson, line 850:
```suggestion
                return  # In this case, the value is already configured.
```

mrobinson (changes requested):
Great work. I really like the use of an enum to store the configuration value. Just a couple changes requested. We should also run this by the community to make sure the change is acceptable to everyone.

python/servo/command_base.py
```diff
@@ -858,6 +879,7 @@ def run_cargo_build_like_command(
         env = env or self.build_env()
 
         args = []
+        port = None
```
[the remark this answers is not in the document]
sagudev: This is from experimenting.

python/servo/command_base.py
sagudev: I extracted this into static method, so it should be nicer now.

python/servo/command_base.py
```diff
@@ -888,6 +912,10 @@ def run_cargo_build_like_command(
                 features.append("webgl-backtrace")
             if self.config["build"]["dom-backtrace"]:
                 features.append("dom-backtrace")
+
+            if self.media_stack.is_dummy() and "--no-default-features" not in cargo_args:
+                args += ["--no-default-features"]
```
mrobinson, line 917: Hrm. There is an issue here and it's that using the dummy media stack disables the "max_log_level", "native-bluetooth", and "webdriver" features. Perhaps we should make these features non-configurable? I'm not sure what the right thing to do here is.

sagudev: Ah, that's why I wanted ports so I could detect which features to enable. Webdriver should probably be hidden behind pref.

We could also move gstreamer enabling into cargo per target, then mach would only need to handle cross and forced dummy cases.
inspiration: https://stackoverflow.com/questions/57972355/enable-a-cargo-feature-by-default-when-the-target-arch-is-wasm

theguy000:
so... update?

mrobinson:
I think @mukilan is now working on this issue and can hopefully revive this change using Cargo featuresets. Sorry for letting it sit so long, but hopefully there will be movement soon.

Issue #2782, opened 2014-07-07
Implement HTMLKeygenElement


brunoabinader:
Spec:
www.whatwg.org/html/#the-keygen-element

CC @Ms2ger

nox:
https://html.spec.whatwg.org/multipage/#the-keygen-element

frewsxcv:
Relevant: https://github.com/whatwg/html/issues/67 https://github.com/whatwg/html/pull/26

KiChjang:
Keygen is deprecated, do we still want to do this?

frewsxcv:
As per the spec:

> This feature is in the process of being removed from the Web platform. (This is a long process that takes many years.) Using the keygen element at this time is highly discouraged.

Doesn't seem worth our time to do anything related to it

jdm:
Agreed.

PR #30429, opened 2023-09-26
Treat the `keygen` element as obsolete

EnnuiL:
oh my god lmao it was so funny to suddenly see `HTMLKeygenElement` with no context at all while swimming at servo code
*ahem*

I've dug deeper, and it turns out it was an ancient form of a crypto API whose support was being worked on in the early days of Servo (with issue #2782 tracking it) until it was deprecated. After 7 years? It appears that it has been completely removed from everything, with it being completely absent on [CanIUse](https://caniuse.com/) while [Firefox's source code](https://searchfox.org) has barely given anything;

So, I went ahead, grabbed the HTML specification, and did my best to scrub out the silly `keygen` element; I did try to run tests in order to see if anything (especially the massive wall of `FAIL`s on `wpt/tests/html/dom/reflection-forms.html` with Layout 2013) changed, but it appears that nothing would budge; I did try running a bigger set of tests, but that did crash the IDE despite the computer taking it somewhat fine

I'm not completely sure if this PR indeed did the trick, so yeah, reviews are appreciated

---

### DIFF
--- a/components/script/dom/create.rs
+++ b/components/script/dom/create.rs
@@ -308,6 +308,8 @@ pub fn create_native_html_element(
         // https://html.spec.whatwg.org/multipage/#other-elements,-attributes-and-apis:isindex-2
         local_name!("isindex") => make!(HTMLUnknownElement),
         local_name!("kbd") => make!(HTMLElement),
+        // https://html.spec.whatwg.org/multipage/#keygen
+        local_name!("keygen") => make!(HTMLUnknownElement),
         local_name!("label") => make!(HTMLLabelElement),
         local_name!("legend") => make!(HTMLLegendElement),
         local_name!("li") => make!(HTMLLIElement),
--- a/resources/iso-8859-8.css
+++ b/resources/iso-8859-8.css
@@ -18,6 +18,6 @@ tbody, tfoot, tr, td, th, dir, dd, dl, dt, menu, ol, ul, li, [dir=ltr i],
   unicode-bidi: bidi-override;
 }
 input:not([type=submit i]):not([type=reset i]):not([type=button i]),
-textarea, keygen {
+textarea {
   unicode-bidi: normal;
 }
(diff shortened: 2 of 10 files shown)


resources/iso-8859-8.css
EnnuiL, line 1: While double-checking whenever I was going on the right path, I noticed that this CSS file is slightly outdated; the relevant spec includes `search` on the bigger block;

The situation involving `user-agent.css` might be more worrying though

components/script/dom/create.rs
```diff
@@ -308,6 +308,8 @@ pub fn create_native_html_element(
         // https://html.spec.whatwg.org/multipage/#other-elements,-attributes-and-apis:isindex-2
         local_name!("isindex") => make!(HTMLUnknownElement),
         local_name!("kbd") => make!(HTMLElement),
+        // https://html.spec.whatwg.org/multipage/#keygen
```
EnnuiL, line 311: While I tried to make the URL look like the others, it appears that the scheme had changed too much in order to allow that; even some of the old URLs aren't redirecting at all

jdm (approved):
This looks reasonable! Let's see if any tests complain.

EnnuiL:
I took a look at the tests situation, and oh god, it's weirder than expected;

Looking at `/_mozilla/mozilla/htmlfieldsetelement_elements.html`? It appears that it expects `keygen` to clearly be both a `HTMLUnknownElement` (it somehow passes this check) and have it count as a list element..? I'd have to look closer at the spec and the code, but the test is apparently wrong at one exceptation while the element managed to fool another? I suspect void tag shenanigans but I can't look that deep rn

Not helping matters is `/html/infrastructure/common-dom-interfaces/collections/htmlformcontrolscollection.html`, which has been passing despite the intention of making sure the `keygen` element doesn't count; Maybe an extra check that directly checks `coll1.elements` would work?

I'd dig further but I'm currently not home, but yeah, I'll fix the tests once I am

EnnuiL:
Just fixed up the Servo test; Turns out the reason the test situation isn't really a situation at all;

The Servo test was accounting for this weird obsoleted/WIP status and so, `keygen` was simultaneously `HTMLUnknownElement` (for reasons I still have to comprehend) and a listed element at the same time; The reason? The `fieldset`'s filter only checks if an element is a listed element; Solution? Do not treat the element as a listed element

Now for the WPT test? The relevant tests were passing because the returned `elements` collection filters out `keygen`s thanks to `HTMLKeygenElement` not really existing, making it invisible to the test in the end

jdm (approved):
Looks good; thanks!

PR #30430, opened 2023-09-26
Refactor the `translatable` attribute lightly

EnnuiL:
This PR is pretty similar to #30094: a random piece of code is cleared off its TODOs;

This time around, however, I went ahead and fixed how the `translate` attribute's value was parsed, making the `match` ASCII case-insensitive (as specified) and fixing 2 `FAIL`s in the process

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [x] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #30435, opened 2023-09-27
Make the `dir` attribute's getter follow the spec

EnnuiL:
In the same vein as #30430, I went ahead and tackled the tiny little issue of massive walls of text prompted by the way Servo handled `dir` attributes; So I made it closer to spec, and well? the results are a bit more disappointing than expected at first, but then I realized that thankfully there was a standard way of creating getters for enumerated attributes *and* that the setter doesn't need normalization, making this effectively a 1-line PR that has massive impact on test results

Hopefully this will be the last small PR for now; Next time I'll want to focus on something bigger that isn't in the vein of "literally change one line in order to fix not-really-impactful stuff", but I'll need more time for that

PS: GitHub's diff is being wonky with the test results; they were all deletions but the diff thinks they're changes

---


jdm (approved):
Good lord, that's an impressive set of test results improvements. Thanks!

PR #30446, opened 2023-09-29
Implement pinch zoom with a display list

mrobinson:
In the most recent version of WebRender, the pinch zoom APIs have been
removed. We can recreate this functionality by creating a root display
list with a transform.

---

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -16,7 +16,7 @@ use compositing_traits::{
 };
 use crossbeam_channel::Sender;
 use embedder_traits::Cursor;
-use euclid::{Point2D, Rect, Scale, Vector2D};
+use euclid::{Point2D, Rect, Scale, Transform3D, Vector2D};
 use fnv::{FnvHashMap, FnvHashSet};
 use gfx_traits::{Epoch, FontData};
 #[cfg(feature = "gl")]
@@ -45,11 +45,14 @@ use time::{now, precise_time_ns, precise_time_s};
 use webrender;
 use webrender::{CaptureBits, RenderApi, Transaction};
 use webrender_api::units::{
-    DeviceIntPoint, DeviceIntSize, DevicePoint, LayoutPoint, LayoutVector2D, WorldPoint,
+    DeviceIntPoint, DeviceIntSize, DevicePoint, LayoutPoint, LayoutRect, LayoutSize,
+    LayoutVector2D, WorldPoint,
 };
 use webrender_api::{
-    self, BuiltDisplayList, DirtyRect, DocumentId, Epoch as WebRenderEpoch, ExternalScrollId,
-    HitTestFlags, PipelineId as WebRenderPipelineId, ScrollClamping, ScrollLocation, ZoomFactor,
+    self, BuiltDisplayList, ClipId, DirtyRect, DocumentId, Epoch as WebRenderEpoch,
+    ExternalScrollId, HitTestFlags, PipelineId as WebRenderPipelineId, PropertyBinding,
+    ReferenceFrameKind, ScrollClamping, ScrollLocation, SpaceAndClipInfo, SpatialId,
+    TransformStyle, ZoomFactor,
 };
 use webrender_surfman::WebrenderSurfman;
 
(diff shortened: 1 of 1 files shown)


components/compositing/compositor.rs
```diff
@@ -862,20 +867,69 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+
+        let mut builder = webrender_api::DisplayListBuilder::new(root_pipeline);
+        let viewport_size = LayoutSize::new(
+            self.embedder_coordinates.get_viewport().width() as f32,
+            self.embedder_coordinates.get_viewport().height() as f32,
+        );
+        let zoom_factor = self.pinch_zoom_level();
+        let viewport_rect = LayoutRect::new(LayoutPoint::zero(), viewport_size);
+        let zoom_reference_frame = builder.push_reference_frame(
+            LayoutPoint::zero(),
```
mukilan, line 890: This not strictly related to this PR, but something I noticed when working on the gesture support for Android -  I understand the current webrender's pinch-zoom api doesn't accept the zoom origin. Would this new implementation allow pinch-zooming from a given point or will it still be from top-left corner?

components/compositing/compositor.rs
```diff
@@ -128,8 +131,10 @@ pub struct IOCompositor<Window: WindowMethods + ?Sized> {
     /// The port on which we receive messages.
     port: CompositorReceiver,
 
-    /// The root pipeline.
-    root_pipeline: RootPipeline,
+    /// The root content pipeline ie the pipeline which contains the main frame
+    /// to display. In the WebRender scene, this will be the one child of another
```
mukilan, line 135: Is this supposed to read ` the *only* child of` instead of  `the one child of`

components/compositing/compositor.rs
```diff
@@ -862,20 +867,69 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+            &SpaceAndClipInfo {
+                spatial_id: zoom_reference_frame,
+                clip_id: ClipId::root(root_pipeline),
+            },
+            root_content_pipeline.to_webrender(),
+            true,
+        );
+        let built_display_list = builder.finalize();
+        transaction.set_display_list(
+            WebRenderEpoch(0),
```
mukilan, line 912: Why do we use zero value for epoch here?

components/compositing/compositor.rs
```diff
@@ -862,20 +867,69 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+    /// Create a WebRender display list for the root pipeline in the given transaction.
+    /// This display list contains a full-screen iframe with the root content pipeline
+    /// with any pinch zoom transform applied to it.
+    fn set_root_display_list(&self, transaction: &mut Transaction) {
+        let root_content_pipeline = match self.root_content_pipeline.id {
+            Some(id) => id,
+            None => return,
+        };
+
+        let root_pipeline = WebRenderPipelineId(0, 0);
```
mukilan, line 879: Is the `(0, 0)` pipelineId guaranteed to unused by normal pipelines?

components/compositing/compositor.rs
mrobinson: This should be fixable in a followup change, but requires answering some questions about how to differentiate between scrolling and zoom in on content.

components/compositing/compositor.rs
mrobinson: I've updated this comment.

components/compositing/compositor.rs
mrobinson: The epoch value doesn't really matter here as WebRender continues to generate frames and update the scene no matter what we pass. I've added a comment explaining this.

components/compositing/compositor.rs
mrobinson: Yes, though it is true that is also what WebRender uses for it's `dummy` pipeline. I've changed this to `(0, 1)` and added a comment explaining why it shouldn't conflict.

mukilan (approved):
Nice!

mrobinson:
Thanks for the review!

mrobinson:
Looks like there were some failures due to pixel snapping. I have updated this change to only use the intermediate display list if the zoom isn't 1.0.

Issue #30364, opened 2023-09-15
Support more interfaces per webidl


sagudev:
Currently codegen needs to have every interface in separate file. This can be very annoying when trying to do diff on spec because all webgpu stuff are split across multiple files ~40. In mozilla WebGPU idl is stored in [one file](https://searchfox.org/mozilla-central/source/dom/webidl/WebGPU.webidl) so it should be possible.

I do not know if there is any specific reason for this limit maybe @jdm knows?

jdm:
I believe the biggest limitation is the interaction with rust's module system. The generated code needs to exist in a module path that is consistent so other code (both generated and handwritten) can refer to it.

sagudev:
I merged all webgpu files into one to start testing and got first problem. WebGPU specs defines `GPUBuffer` and `GPUBufferBinding` and our codegen generates mod named `GPUBufferBinding` for `GPUBuffer`.

So to solve this namespace problem I would add the following new handling:

if there are multiple interfaces in same webidl file, every intrerace would be in separate submod inside of webidl named mod, so when merged WebGPU, GPUBuffer will reside in webgpu::gpubuffer::GPUBuffer

sagudev:
After toying around with it, I think it would be much more reasonable to just change *Binding postfix to something less collision like `_Binding`. That's how it was done in ff: https://bugzilla.mozilla.org/show_bug.cgi?id=1470325

PR #30447, opened 2023-09-29
Use Foo_Binding instead of FooBinding for namespace modules

sagudev:
As noted in https://github.com/servo/servo/issues/30364#issuecomment-1740936152 that is how it was done in mozilla https://bugzilla.mozilla.org/show_bug.cgi?id=1470325

I recommend reviewing per commit (only first commit really does something)

---

- [x] These changes makes #30364 easier

- [x] There are tests for these changes in Rustc :)


jdm (approved):
This seems like a reasonable solution!

PR #30448, opened 2023-09-29
Codegen support multiple interfaces in single webidl file & WebGPU single webidl

### LINKED ISSUE #30364: Support more interfaces per webidl

sagudev:
for demonstration all WebGPU stuff is merged together (ordering is modeled after [upstream](https://github.com/w3c/webref/blob/curated/ed/idl/webgpu.idl) for easier future sync)

~~depends on #30447~~ DONE

Review per commit (first commit is the most important one).

---

- [x] These changes fix #30364 


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/6360711406/job/17274992629

jdm (approved):

PR #13693, opened 2016-10-11
Correct the unicode codes used for tree printing

mrobinson:
---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] These changes fix #__ (github issue number if applicable).

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because the tree printing code is untested and this is just a correction of a build fix.

These were converted from inline UTF-8 to escape sequences, but the
sequences should be in hexadecimal instead of decimal.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/servo/13693)

PR #13649, opened 2016-10-08
Update to Rust 1.14.0-nightly (19ac57926 2016-10-08)

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/13649)

PR #30458, opened 2023-10-02
Fix painting order of atomic inline stacking containers

### LINKED ISSUE #30428: Positioned elements should be painted in front of others
```html
<!DOCTYPE html>
<style>
.wrapper { display: inline-block; width: 100px; height: 100px; border: 5px solid }
</style>
<div class="wrapper">
  <div style="height: 50px; position: relative; left: 50px; background: cyan"></div>
</div>
<div class="wrapper"></div>
```

| 2020 (bad) | 2013 (good) |
| - | - |
|  |  |

See https://drafts.csswg.org/css-position-4/#paint-a-stacking-context, the back-to-front order is

 1. Positioned with negative z-index
 2. Non-positioned
 3. Positioned with `z-index: auto` or `z-index: 0`
 3. Positioned with positive `z-index: 0`

delan:
When we [add children to a stacking context](https://github.com/servo/servo/blob/d74d51190a523d20775bf448eee408935d346e8b/components/layout_2020/display_list/stacking_context.rs#L287-L294), [sort the stacking context](https://github.com/servo/servo/blob/d74d51190a523d20775bf448eee408935d346e8b/components/layout_2020/display_list/stacking_context.rs#L302-L321), and [build a display list for it](https://github.com/servo/servo/blob/d74d51190a523d20775bf448eee408935d346e8b/components/layout_2020/display_list/stacking_context.rs#L500-L559), we partition our children into fragments, float stacking containers, and all other stacking containers and stacking contexts. Then we partition the latter into those with negative ‘z-index’ and those with non-negative (or auto) ‘z-index’.

This is incorrect per [css-position-4 #painting-order](https://drafts.csswg.org/css-position-4/#painting-order) and w3c/csswg-drafts#2717. Atomic inline stacking containers should never be stolen by the parent stacking context, and they should be painted together with fragments in tree order.

This patch fixes that by:

* converting Stacking­Context­Fragment to an enum that can also represent “paint inline stacking container #i”
* pushing that new variant to the fragments vec whenever we create a new atomic inline stacking container
* splitting Stacking­Context­.stacking­_contexts into real­_stacking­_contexts­_and­_positioned­_stacking­_containers and atomic­_inline­_stacking­_containers

We also add warnings for the assertions in build­_canvas­_background­_display­_list when debug assertions are not enabled, and add debug logging to show the stacking context tree in painting order under `-Z dump-stacking-context-tree`.

---

Here are the results and debug logs for the original test cases:

<details><summary>test case in <a href="https://github.com/servo/servo/issues/30428#issue-1912263167">#30428 (description)</a></summary>

image

```
┌ Stacking context tree
│  ├─ RealStackingContext z=0
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ AtomicInlineStackingContainer #0
│  │  │  └─ OwnBackgroundsAndBorders
│  │  ├─ Foreground
│  │  ├─ AtomicInlineStackingContainer #1
│  │  │  └─ OwnBackgroundsAndBorders
│  │  ├─ PositionedStackingContainer
│  │  │  └─ OwnBackgroundsAndBorders
```
</details>

<details><summary>test case in <a href="https://github.com/servo/servo/issues/30428#issuecomment-1734463636">#30428 (comment)</a></summary>

image

```
┌ Stacking context tree
│  ├─ RealStackingContext z=0
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ AtomicInlineStackingContainer #0
│  │  │  └─ OwnBackgroundsAndBorders
│  │  ├─ PositionedStackingContainer
│  │  │  └─ OwnBackgroundsAndBorders
```
</details>

<details><summary>test case in <a href="https://github.com/servo/servo/pull/30458#issuecomment-1744057255">#30458 (comment)</a></summary>

image

```
┌ Stacking context tree
│  ├─ RealStackingContext z=0
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ AtomicInlineStackingContainer #0
│  │  │  ├─ OwnBackgroundsAndBorders
│  │  │  ├─ DescendantBackgroundsAndBorders
│  │  │  └─ Foreground
│  │  ├─ Foreground
│  │  ├─ Foreground
│  │  └─ Foreground
```
</details>

<details><summary>test case in #30474 (description)</summary>

```
┌ Stacking context tree
│  ├─ RealStackingContext z=0
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ DescendantBackgroundsAndBorders
│  │  ├─ RealStackingContext z=0
│  │  │  ├─ OwnBackgroundsAndBorders
│  │  │  ├─ RealStackingContext z=0
│  │  │  │  └─ OwnBackgroundsAndBorders
│  │  │  ├─ RealStackingContext z=0
│  │  │  │  └─ OwnBackgroundsAndBorders
│  │  ├─ RealStackingContext z=0
│  │  │  ├─ OwnBackgroundsAndBorders
│  │  │  ├─ RealStackingContext z=0
│  │  │  │  └─ OwnBackgroundsAndBorders
│  │  │  ├─ RealStackingContext z=0
│  │  │  │  └─ OwnBackgroundsAndBorders
```
</details>

- [x] These changes fix #30428

    - PASS [expected FAIL] /css/CSS2/floats/overflow-scroll-float-paint-order.html
    - PASS [expected FAIL] /css/CSS2/normal-flow/inline-block-zorder-003.xht
    - PASS [expected FAIL] /css/CSS2/normal-flow/inline-table-zorder-003.xht
    - FAIL [expected PASS] /css/css-transforms/preserve3d-and-flattening-z-order-005.html<br>#30474
    - PASS [expected FAIL] /html/rendering/replaced-elements/the-option-element/select-multiple-covered-by-abspos.html

### DIFF
--- a/components/gfx_traits/print_tree.rs
+++ b/components/gfx_traits/print_tree.rs
@@ -84,6 +84,6 @@ impl PrintTree {
 
 impl Drop for PrintTree {
     fn drop(&mut self) {
-        self.flush_queued_item("\u{9492}\u{9472}");
+        self.flush_queued_item("\u{2514}\u{2500}");
     }
 }
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -118,9 +118,6 @@ pub(crate) struct DisplayListBuilder<'a> {
     /// text, image, non-white canvas or SVG). Used by metrics.
     /// See https://w3c.github.io/paint-timing/#first-contentful-paint.
     is_contentful: bool,
-
-    #[cfg(debug_assertions)]
-    debug_print_indent: usize,
 }
 
 impl DisplayList {
@@ -138,9 +135,6 @@ impl DisplayList {
             context,
             display_list: self,
             iframe_sizes: FnvHashMap::default(),
-
-            #[cfg(debug_assertions)]
-            debug_print_indent: 0,
         };
         fragment_tree.build_display_list(&mut builder, root_stacking_context);
         (builder.iframe_sizes, builder.is_contentful)
--- a/components/layout_2020/display_list/mod.rs
+++ b/components/layout_2020/display_list/mod.rs
@@ -118,6 +118,9 @@ pub(crate) struct DisplayListBuilder<'a> {
     /// text, image, non-white canvas or SVG). Used by metrics.
     /// See https://w3c.github.io/paint-timing/#first-contentful-paint.
     is_contentful: bool,
+
+    #[cfg(debug_assertions)]
+    debug_print_indent: usize,
 }
 
 impl DisplayList {
@@ -135,6 +138,9 @@ impl DisplayList {
             context,
             display_list: self,
             iframe_sizes: FnvHashMap::default(),
+
+            #[cfg(debug_assertions)]
+            debug_print_indent: 0,
         };
         fragment_tree.build_display_list(&mut builder, root_stacking_context);
         (builder.iframe_sizes, builder.is_contentful)
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -7,6 +7,7 @@ use std::mem;
 
 use euclid::default::Rect;
 use gfx_traits::print_tree::PrintTree;
+use log::warn;
 use script_traits::compositor::{ScrollTreeNodeId, ScrollableNodeInfo};
 use servo_arc::Arc as ServoArc;
 use servo_config::opts::DebugOptions;
@@ -519,22 +520,33 @@ impl StackingContext {
                 })
         });
 
+        macro_rules! debug_panic {
+            ($msg: expr) => {
+                if cfg!(debug_assertions) {
+                    panic!($msg);
+                } else {
+                    warn!($msg);
+                    return;
+                }
+            };
+        }
+
         let first_stacking_context_fragment = if let Some(first) = first_if_any {
             first
         } else {
             // This should only happen if the root element has `display: none`
-            panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
+            debug_panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
         };
 
         let StackingContextContent::Fragment { fragment, scroll_node_id, containing_block, .. }
             = first_stacking_context_fragment else {
-                panic!("Expected a fragment, not a stacking container");
+                debug_panic!("Expected a fragment, not a stacking container");
             };
         let fragment = fragment.borrow();
         let box_fragment = match &*fragment {
             Fragment::Box(box_fragment) | Fragment::Float(box_fragment) => box_fragment,
             _ => {
-                panic!("Expected a box-generated fragment");
+                debug_panic!("Expected a box-generated fragment");
             },
         };
 
@@ -676,12 +688,17 @@ impl StackingContext {
         }
     }
 
-    /// Print the stacking context tree, or panics if [Self::debug_print_items] is None.
+    /// Print the stacking context tree.
     pub fn debug_print(&self) {
+        if self.debug_print_items.is_none() {
+            warn!("failed to print stacking context tree: debug_print_items was None");
+            return;
+        }
         let mut tree = PrintTree::new("Stacking context tree".to_owned());
         self.debug_print_with_tree(&mut tree);
     }
 
+    /// Print a subtree with the given [PrintTree], or panic if [Self::debug_print_items] is None.
     fn debug_print_with_tree(&self, tree: &mut PrintTree) {
         match self.context_type {
             StackingContextType::RealStackingContext => {
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -669,7 +669,7 @@ impl StackingContext {
     /// Store the fact that something was painted, if [Self::debug_print_items] is not None.
     ///
     /// This is used to help reconstruct the original painting order in [Self::debug_print] without
-    /// duplicating our painting order logic, which could fall out of sync with the real logic.
+    /// duplicating our painting order logic, since that could fall out of sync with the real logic.
     fn debug_push_print_item(&self, field: DebugPrintField, index: usize) {
         if let Some(items) = self.debug_print_items.as_ref() {
             items.borrow_mut().push(DebugPrintItem { field, index });
@@ -703,7 +703,7 @@ impl StackingContext {
                         tree.add_item(format!("{:?}", section));
                     },
                     StackingContextContent::AtomicInlineStackingContainer { index } => {
-                        tree.new_level(format!("AtomicInlineStackingContainer {}", index));
+                        tree.new_level(format!("AtomicInlineStackingContainer #{}", index));
                         self.atomic_inline_stacking_containers[index].debug_print_with_tree(tree);
                         tree.end_level();
                     },
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -210,14 +210,14 @@ pub(crate) enum StackingContextContent {
     /// An index into [StackingContext::atomic_inline_stacking_containers].
     ///
     /// There is no section field, because these are always in [StackingContextSection::Foreground].
-    InlineStackingContainer { index: usize },
+    AtomicInlineStackingContainer { index: usize },
 }
 
 impl StackingContextContent {
     fn section(&self) -> StackingContextSection {
         match self {
             Self::Fragment { section, .. } => *section,
-            Self::InlineStackingContainer { .. } => StackingContextSection::Foreground,
+            Self::AtomicInlineStackingContainer { .. } => StackingContextSection::Foreground,
         }
     }
 
@@ -246,7 +246,7 @@ impl StackingContextContent {
                     .borrow()
                     .build_display_list(builder, containing_block, *section);
             },
-            Self::InlineStackingContainer { index } => {
+            Self::AtomicInlineStackingContainer { index } => {
                 #[cfg(debug_assertions)]
                 {
                     debug!("{}InlineStackingContainer {}", indent, *index);
@@ -887,7 +887,7 @@ impl BoxFragment {
         if context_type == StackingContextType::AtomicInlineStackingContainer {
             // Push a dummy fragment that indicates when the new stacking context should be painted.
             parent_stacking_context.contents.push(
-                StackingContextContent::InlineStackingContainer {
+                StackingContextContent::AtomicInlineStackingContainer {
                     index: parent_stacking_context
                         .atomic_inline_stacking_containers
                         .len(),
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -487,36 +487,22 @@ impl StackingContext {
                 })
         });
 
-        macro_rules! debug_panic {
-            ($msg: expr) => {
-                if cfg!(debug_assertions) {
-                    panic!($msg)
-                } else {
-                    // ::log::error!($msg)
-                    panic!($msg)
-                }
-            };
-        }
-
         let first_stacking_context_fragment = if let Some(first) = first_if_any {
             first
         } else {
             // This should only happen if the root element has `display: none`
-            debug_panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
-            // return;
+            panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
         };
 
         let StackingContextFragment::Fragment { fragment, scroll_node_id, containing_block, .. }
             = first_stacking_context_fragment else {
-                debug_panic!("Expected a fragment, not a stacking container");
-                // return;
+                panic!("Expected a fragment, not a stacking container");
             };
         let fragment = fragment.borrow();
         let box_fragment = match &*fragment {
             Fragment::Box(box_fragment) | Fragment::Float(box_fragment) => box_fragment,
             _ => {
-                debug_panic!("Expected a box-generated fragment");
-                // return;
+                panic!("Expected a box-generated fragment");
             },
         };
 
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -207,9 +207,9 @@ pub(crate) enum StackingContextContent {
         fragment: ArcRefCell<Fragment>,
     },
 
-    /// An index into [StackingContext::inline_stacking_containers].
+    /// An index into [StackingContext::atomic_inline_stacking_containers].
     ///
-    /// There is no section field, because these are always in [StackingContextSection::Content].
+    /// There is no section field, because these are always in [StackingContextSection::Foreground].
     InlineStackingContainer { index: usize },
 }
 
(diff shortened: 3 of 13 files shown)


Loirooriol (commented):
The problem is that for

```
<div style="width: 100px; height: 100px; background: red; position: relative;">
  <div style="display: inline-block; width: 100px; height: 100px; background: green;"></div>
</div>
```

this paints the inline-block behind its parent due to `(SCT::PseudoAtomicInline, _) => Ordering::Less`.

delan:
It looks like we only prevent float stacking containers from being stolen by the parent stacking context, when we should be preventing all stacking containers from being stolen. Only real stacking contexts should be stolen.

I also noticed that we fail to interleave atomic inline stacking containers with fragments:

```html
<div style="display: inline-block; width: 1em; height: 1em;">
<div style="background: red; width: 5em; height: 1em;">— FAIL</div>
</div> <span style="background: white;">PASS</span>
```

image image

delan:
bbd178bb4f53e4e92a4e57ab7302819cfbb329f1 attempts to fix both of those bugs, let’s see how that goes.

@bors-servo try=wpt-2020

delan:
f76bf3c84ec14739d02fff2c8c95c053bc13f0b9 fixes interleaving with fragments and prevents atomic inline stacking containers from being stolen, but it erroneously prevents positioned stacking containers from being stolen. 29a5342d714ca71ff8da90b284f9131212053967 fixes that.

> It looks like we only prevent float stacking containers from being stolen by the parent stacking context, when we should be preventing all stacking containers from being stolen. Only real stacking contexts should be stolen.

This was incorrect. Both real stacking contexts and positioned stacking containers should be stolen, while float stacking containers and atomic inline stacking containers should be left in the new stacking container.

components/layout_2020/display_list/mod.rs
```diff
@@ -118,6 +118,9 @@ pub(crate) struct DisplayListBuilder<'a> {
     /// text, image, non-white canvas or SVG). Used by metrics.
     /// See https://w3c.github.io/paint-timing/#first-contentful-paint.
     is_contentful: bool,
+
+    #[cfg(debug_assertions)]
+    debug_print_indent: usize,
```
mrobinson, line 123: I think it would be better to use the `PrintTree` infrastructure for this.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -192,25 +193,70 @@ impl DisplayList {
     }
 }
 
-pub(crate) struct StackingContextFragment {
-    scroll_node_id: ScrollTreeNodeId,
-    clip_chain_id: wr::ClipChainId,
-    section: StackingContextSection,
-    containing_block: PhysicalRect<Length>,
-    fragment: ArcRefCell<Fragment>,
+pub(crate) enum StackingContextFragment {
```
mrobinson, line 196: Maybe this should be renamed to something like `StackingContextMember`. I think it would be good to put a little documentation here about why InlineStackingContexts need to be stored in the same data structure. I assume they need to maintain their order.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -192,25 +193,70 @@ impl DisplayList {
...
+                scroll_node_id,
+                clip_chain_id,
+                section,
+                containing_block,
+                fragment,
+            } => {
+                #[cfg(debug_assertions)]
+                {
+                    debug!("{}{:?}", indent, section);
+                }
```
mrobinson, line 236: Let's remove the debug statements from this code and have a separate set of debug functions that are called when a certain log level is activated or some other mechanism. I'm not sure that `debug_assertions` is the right trigger since these aren't really assertions.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -237,24 +283,28 @@ pub struct StackingContext {
...
     /// > To paint a stacking container, given a box root and a canvas canvas:
     /// >  1. Paint a stacking context given root and canvas, treating root as
     /// >     if it created a new stacking context, but omitting any positioned
     /// >     descendants or descendants that actually create a stacking context
     /// >     (letting the parent stacking context paint them, instead).
-    ///
-    /// Note that all stacking containers / pseudo stacking contexts are passed up
-    /// to parent stacking contexts, except in the case of floats.
-    float_stacking_contexts: Vec<StackingContext>,
+    stealable_children: Vec<StackingContext>,
```
mrobinson, line 295: Perhaps a better name is `block_level_stacking_contexts` or something along those lines? It's more in line with the names of the other collections of stacking containers and stacking contexts.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -237,24 +283,28 @@ pub struct StackingContext {
...
+    /// Float stacking containers.
+    /// These are separate from [Self::stealable_children] because they should
+    /// never be stolen by the parent stacking context.
+    float_stacking_containers: Vec<StackingContext>,
+
+    /// Atomic inline stacking containers.
+    /// These are separate from [Self::stealable_children] because they should
+    /// never be stolen by the parent stacking context, and separate from
+    /// [Self::float_stacking_containers] so that [StackingContextFragment] can
+    /// index into this vec to paint them in fragment order.
```
mrobinson, line 306: Might be good to put a pointer to the relevant part of the specification here.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -237,24 +283,28 @@ pub struct StackingContext {
...
-    ///
-    /// Note that all stacking containers / pseudo stacking contexts are passed up
-    /// to parent stacking contexts, except in the case of floats.
-    float_stacking_contexts: Vec<StackingContext>,
+    stealable_children: Vec<StackingContext>,
+
+    /// Float stacking containers.
+    /// These are separate from [Self::stealable_children] because they should
+    /// never be stolen by the parent stacking context.
+    float_stacking_containers: Vec<StackingContext>,
```
mrobinson, line 300: It's great that we can start using the "stacking container" terminology here.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -279,18 +330,21 @@ impl StackingContext {
...
 
     /// Add a child stacking context to this stacking context.
     fn add_stacking_context(&mut self, stacking_context: StackingContext) {
-        if stacking_context.context_type == StackingContextType::PseudoFloat {
-            self.float_stacking_contexts.push(stacking_context);
-        } else {
-            self.stacking_contexts.push(stacking_context);
+        match stacking_context.context_type {
+            StackingContextType::Real => &mut self.stealable_children,
+            StackingContextType::PseudoPositioned => &mut self.stealable_children,
```
mrobinson, line 343: Maybe this is a good moment to standardize on "stacking container" and rename these to be something like "PositionedStackingContainer" or whatever works for you.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -500,57 +549,90 @@ impl StackingContext {
...
+            child_fragments
+                .next()
+                .unwrap()
+                .build_display_list(builder, &self.inline_stacking_containers);
         }
 
-        // Step 3: Positioned descendants with negative z-indices
-        let mut child_stacking_contexts = self.stacking_contexts.iter().peekable();
-        while child_stacking_contexts
+        // Step 3: Stacking contexts with negative ‘z-index’
```
mrobinson, line 582: I may be wrong, but I think the only stacking contexts with negative effective z-indices are those that are positioned. Not sure if this deserves more clarification.

mrobinson (changes requested):
Great stuff. The main thing that could be changed here is specifically add `PrintTree` support to the StackingContextTree and separate this out fully from the process of painting the tree. This will keep the code a bit cleaner and make it more generally reusable; there might be other situations where we want to print the tree.

components/layout_2020/display_list/stacking_context.rs
delan: i think you’re right for now, since flex and grid aren’t yet implemented, but ‘z-index’ applies to [flex items](https://drafts.csswg.org/css-flexbox/#painting) and [grid items](https://drafts.csswg.org/css-grid/#z-order) even when ‘position’ is ‘static’. (that said, it’s not clear to me whether “behaves exactly as if ‘position’ were ‘relative’” is limited to ‘z-index’, or if the used value becomes ‘relative’ in every sense…)

components/layout_2020/display_list/stacking_context.rs
delan: i’m not sure i agree with this one, because positioned stacking containers are not stacking contexts. would stealable­_stacking­_conts work for you?

components/layout_2020/display_list/stacking_context.rs
delan: that was the approach i used initially, with a Stacking­Context­.debug­_print method that prints the whole tree, but i decided interleaving it with the real logic would be better. having a separate method means we need to duplicate the painting order logic, so it can fall out of sync with the real logic over time and become misleading.

debug­_assertions is the same trigger that [log](https://crates.io/crates/log) uses to [compile out log levels](https://docs.rs/crate/log/0.4.20/source/src/lib.rs) in release builds, but i don’t feel too strongly about it. do you know a better trigger we can use?

components/layout_2020/display_list/mod.rs
delan: oo i wasn’t aware of that, looking into it :)

components/layout_2020/display_list/stacking_context.rs
mrobinson: I think we should avoid the "stealable" terminology for members in order to stick closer to the specification language. Also they aren't stealable if `self` isn't a stacking container. How about `stacking_contexts_and_positioned_stacking_containers`. It's a long, but it's much more obvious what it holds and bits are basically free for us.

components/layout_2020/display_list/stacking_context.rs
delan: done in 8e60b20ca11c2a3d755e21931661c14d108f9261

components/layout_2020/display_list/stacking_context.rs
delan: done in 8e60b20ca11c2a3d755e21931661c14d108f9261

components/layout_2020/display_list/stacking_context.rs
delan: done in 8e60b20ca11c2a3d755e21931661c14d108f9261

components/layout_2020/display_list/stacking_context.rs
delan: thanks!

components/layout_2020/display_list/stacking_context.rs
delan: done in 8e60b20ca11c2a3d755e21931661c14d108f9261

delan:
8e60b20ca11c2a3d755e21931661c14d108f9261 and eba315de04c7351c985234dba317af4b453bbba7 rework the naming and docs as follows:

* Stacking­Context­Section variants
    * Block­Backgrounds­And­Borders → Descendant­Backgrounds­And­Borders, since the distinction is not actually “block” vs “not block” but rather “root fragment” vs “descendant fragment”
    * Backgrounds­And­Borders → Own­Backgrounds­And­Borders, for the same reason and because “root” is too easy to confuse for the root _element_ in step 3 of [#paint-a-stacking-context](https://drafts.csswg.org/css-position-4/#paint-a-stacking-context)
    * Content → Foreground, to avoid confusion with Stacking­Context­Content below
* Stacking­Context­Fragment → Stacking­Context­Content variants
    * Inline­Stacking­Container → Atomic­Inline­Stacking­Container
* Stacking­Context­Type variants
    * Real → Real­Stacking­Context
    * Pseudo­Positioned → Positioned­Stacking­Container
    * Pseudo­Float → Float­Stacking­Container
    * Pseudo­Atomic­Inline → Atomic­Inline­Stacking­Container
* Stacking­Context fields
    * stealable­_children → real­_stacking­_contexts­_and­_positioned­_stacking­_containers
    * inline­_stacking­_containers → atomic­_inline­_stacking­_containers
* “stacking context” refers to a stacking context or stacking container
* “real stacking context” refers to a stacking context only
* removed all remaining usage of the word “pseudo”

I considered referring to things that can be either a stacking context or stacking container as a “stacking cont” (and renaming symbols accordingly), but the result looked too eccentric. The approach I went with still invents terminology (“real”), but I think that’s unavoidable with the current spec.

components/layout_2020/display_list/mod.rs
delan: done in ff39a86828490e6f4056810850d792cc1760cd28

components/layout_2020/display_list/stacking_context.rs
delan: done in ff39a86828490e6f4056810850d792cc1760cd28

components/gfx_traits/print_tree.rs
```diff
@@ -84,6 +84,6 @@ impl PrintTree {
 
 impl Drop for PrintTree {
     fn drop(&mut self) {
-        self.flush_queued_item("\u{9492}\u{9472}");
+        self.flush_queued_item("\u{2514}\u{2500}");
```
delan, line 87: (see #13649, #13693)

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -500,63 +581,154 @@ impl StackingContext {
...
+    ///
+    /// This is used to help reconstruct the original painting order in [Self::debug_print] without
+    /// duplicating our painting order logic, since that could fall out of sync with the real logic.
+    fn debug_push_print_item(&self, field: DebugPrintField, index: usize) {
+        if let Some(items) = self.debug_print_items.as_ref() {
+            items.borrow_mut().push(DebugPrintItem { field, index });
+        }
+    }
+
+    /// Print the stacking context tree, or panics if [Self::debug_print_items] is None.
```
mrobinson, line 679: Looks like this doesn't panic. I think that's a better behavior actually, but perhaps the comment should be corrected.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -420,38 +509,32 @@ impl StackingContext {
...
-            };
-        }
-
         let first_stacking_context_fragment = if let Some(first) = first_if_any {
             first
         } else {
             // This should only happen if the root element has `display: none`
-            debug_panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
-            return;
+            panic!("`CanvasBackground::for_root_element` should have returned `style: None`");
```
mrobinson, line 526: I think this should fail gracefully instead of panic. I believe we need to start cleaning up as many panics as possible.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -420,38 +509,32 @@ impl StackingContext {
...
+            = first_stacking_context_fragment else {
+                panic!("Expected a fragment, not a stacking container");
+            };
+        let fragment = fragment.borrow();
         let box_fragment = match &*fragment {
             Fragment::Box(box_fragment) | Fragment::Float(box_fragment) => box_fragment,
             _ => {
-                debug_panic!("Expected a box-generated fragment");
-                return;
+                panic!("Expected a box-generated fragment");
```
mrobinson, line 537: Ditto. Failing gracefully is almost certainly better for users here.

mrobinson (approved):
Thanks for the iterations on this. This is looking good to me. There is only one thing that I think should be changed before landing and that is avoiding a panic when having issues painting the root canvas background.

components/layout_2020/display_list/stacking_context.rs
delan: done in 43692821f9bccde5eda1b5651e21c535bfaebe9d

components/layout_2020/display_list/stacking_context.rs
delan: done in 43692821f9bccde5eda1b5651e21c535bfaebe9d

components/layout_2020/display_list/stacking_context.rs
delan: done in /files/e2df884f1c60ec28793f8a9a2a1f5d37e6dba028#diff-82520938a024f875c789f7314111aaeb3f11a47fd9b956a3a90fef8399ddaea1

delan: done in 43692821f9bccde5eda1b5651e21c535bfaebe9d

PR #30459, opened 2023-10-02
Fix pinch zoom and enable it for TouchpadMagnify events

mrobinson:
Pinch zoom was broken because pinch zoom events were triggered at -1, -1
in the compositor. This change:

1. Differentiates between magnify and scroll events in the compositor to
   remove the question of where to trigger them.
2. Converts winit TouchpadMagnify events into pinch zoom events so that
   this works properly on MacOS.

---

- [x] These changes do not require tests because it is very difficult to test interaction in Servo currently.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -248,15 +248,22 @@ pub struct IOCompositor<Window: WindowMethods + ?Sized> {
 }
 
 #[derive(Clone, Copy)]
-struct ScrollZoomEvent {
-    /// Change the pinch zoom level by this factor
-    magnification: f32,
+struct ScrollEvent {
     /// Scroll by this offset, or to Start or End
     scroll_location: ScrollLocation,
     /// Apply changes to the frame at this location
     cursor: DeviceIntPoint,
     /// The number of OS events that have been coalesced together into this one event.
     event_count: u32,
+}
+
+#[derive(Clone, Copy)]
+enum ScrollZoomEvent {
+    /// An pinch zoom event that magnifies the view by the given factor.
+    PinchZoom(f32),
+    /// A scroll event that scrolls the scroll node at the given location by the
+    /// given amount.
+    Scroll(ScrollEvent),
 }
 
 #[derive(Debug, PartialEq)]
@@ -375,7 +382,7 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
             shutdown_state: ShutdownState::NotShuttingDown,
             page_zoom: Scale::new(1.0),
             viewport_zoom: PinchZoomFactor::new(1.0),
-            min_viewport_zoom: None,
+            min_viewport_zoom: Some(PinchZoomFactor::new(1.0)),
             max_viewport_zoom: None,
             zoom_action: false,
             zoom_time: 0f64,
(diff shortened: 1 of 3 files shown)


components/compositing/compositor.rs
```diff
@@ -375,8 +378,6 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
             shutdown_state: ShutdownState::NotShuttingDown,
             page_zoom: Scale::new(1.0),
             viewport_zoom: PinchZoomFactor::new(1.0),
-            min_viewport_zoom: None,
-            max_viewport_zoom: None,
```
mukilan, line 379: I am using the max_viewport_zoom in the android port to clamp the maximium value of MAX_ZOOM (currently used for page zoom). Is it not useful to have this clamping in compositor, or is that something we expect the layers above to handle (in this case the android event handling logic in Java)

components/compositing/compositor.rs
```diff
@@ -1219,14 +1220,16 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
                 let cursor = Point2D::new(-1, -1); // Make sure this hits the base layer.
-                self.pending_scroll_zoom_events.push(ScrollZoomEvent {
-                    magnification: magnification,
-                    scroll_location: ScrollLocation::Delta(LayoutVector2D::from_untyped(
-                        scroll_delta.to_untyped(),
-                    )),
-                    cursor: cursor,
-                    event_count: 1,
-                });
+                self.pending_scroll_zoom_events
```
mukilan, line 1223: Do the order of the scroll & pinch zoom events matter here? If so, can we add a comment?

components/compositing/compositor.rs
```diff
@@ -1280,107 +1283,122 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+            self.send_scroll_positions_to_layout_for_pipeline(&pipeline_id);
             self.waiting_for_results_of_scroll = true
         }
+
+        transaction.generate_frame(0);
+        self.webrender_api
+            .send_transaction(self.webrender_document, transaction);
+    }
+
+    /// Given the result of a hit test and a ScrollLocation try to scroll the node at the result,
```
mukilan, line 1371: This method doesn't accept the hit test result directly, but accepts the cursor position and performs a hit test. Not sure if this matters, just pointing it out.

components/servo/lib.rs
```diff
@@ -540,16 +540,16 @@ where
                     .on_scroll_event(scroll_location, cursor, phase);
             },
 
-            EmbedderEvent::Zoom(magnification) => {
-                self.compositor.on_zoom_window_event(magnification);
+            EmbedderEvent::Zoom(zoom) => {
```
mukilan, line 543: Several places in this compositor.rs still use `magnification` as the name of the zoom factor. Maybe this can be left as magnification to be consistent?

components/servo/lib.rs
```diff
@@ -540,16 +540,16 @@ where
...
+                self.compositor.on_zoom_window_event(zoom);
             },
 
             EmbedderEvent::ResetZoom => {
                 self.compositor.on_zoom_reset_window_event();
             },
 
-            EmbedderEvent::PinchZoom(magnification) => {
-                self.compositor.on_pinch_zoom_window_event(magnification);
+            EmbedderEvent::PinchZoom(zoom) => {
```
mukilan, line 551: Same comment as above

components/compositing/compositor.rs
mrobinson: No, it probably make sense to have this in the compositor.

components/compositing/compositor.rs
```diff
@@ -1280,107 +1283,122 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-                            ScrollClamping::NoClamping,
-                        );
-                    },
-                    None => {},
-                }
-            }
-            self.send_scroll_positions_to_layout_for_pipeline(&result.pipeline_id);
+        let mut transaction = Transaction::new();
+        if zoom_changed {
+            self.set_root_display_list(&mut transaction);
```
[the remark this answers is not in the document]
mrobinson: We will use `set_root_display_list` for both cases, but in the case that there is no pinch zoom, the root pipeline will be the root content pipeline. I've changed this to `set_root_content_pipeline_handling_pinch_zoom`.

components/compositing/compositor.rs
mrobinson: Thanks for pointing this out. The documentation was totally wrong. I've updated it.

components/servo/lib.rs
mrobinson: Sure. Though I'm not sure if long-term `magnification` is the right term.

components/compositing/compositor.rs
mrobinson: I've restored these members, but set the default min zoom to 1.0.

mrobinson:
Thanks for the review. I've updated the changes according to your comments.

mukilan (approved):

PR #30477, opened 2023-10-03
Update core-foundation in simpleservo

### LINKED ISSUE #30319: Update core-foundation version in simpleservo
The last remaining use of core-foundation 0.6 is in simpleservo. Updating this to the latest version would get rid of the duplicate dependency.

mrobinson:
---

- [x] These changes fix #30319.
- [x] These changes do not require tests because they just update a dependency.

### DIFF
--- a/ports/libsimpleservo/api/Cargo.toml
+++ b/ports/libsimpleservo/api/Cargo.toml
@@ -20,7 +20,7 @@ webxr-api = { git = "https://github.com/servo/webxr", features = ["ipc"] }
 libc = { workspace = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-core-foundation = "0.6"
+core-foundation = "0.9"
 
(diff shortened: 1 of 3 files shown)


ports/libsimpleservo/api/Cargo.toml
```diff
@@ -20,7 +20,7 @@ webxr-api = { git = "https://github.com/servo/webxr", features = ["ipc"] }
 libc = { workspace = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
-core-foundation = "0.6"
+core-foundation = "0.7"
```
atouchet, line 23: Why is this only being updated to 0.7 and not 0.9?

mrobinson: Good question. I've updated this change to upgrade it to 0.9. This depends on #30481 though.

mukilan (approved):

PR #30481, opened 2023-10-03
Fix libsimpleservo build

mrobinson:
It seems the build has been broken on desktop platforms. This is a
preliminary fix to getting this build on the bots.

This is mainly missing includes, but `package-prefs.json` is no longer
processed because that was only used for the UWP port (now removed.

---

- [x] These changes do not require tests because it's just a build fix.

### DIFF
--- a/ports/libsimpleservo/api/build.rs
+++ b/ports/libsimpleservo/api/build.rs
@@ -62,13 +62,9 @@ fn main() {
             .unwrap();
     }
 
-    // Merge prefs.json and package-prefs.json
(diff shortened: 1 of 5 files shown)


jdm:
Do we no longer build libsimpleservo on CI? Can we address that?

mrobinson:
Yeah, it's possible to get simpleservo building on CI again, though right now switching between the two builds rebuilds many packages for some reason. This should probably be fixed before enabling this on CI to avoid lengthening builds unnecessarily.

jdm (approved):

PR #30487, opened 2023-10-04
Add taplo to etc/shell.nix

mukilan:
This addresses the 'Could not find `taplo` ...' message
when running `./mach test-tidy` on NixOS.

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -22,6 +22,7 @@ clangStdenv.mkDerivation rec {
     gst_all_1.gst-plugins-bad
 
     rustup
+    taplo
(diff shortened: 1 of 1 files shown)


wusyong:
Just a note that macOS also has this warning when running `./mach test-tidy` despite `./mach bootstrap` has already runned.

mrobinson (approved):

delan:
I ran into this too, but ended up fixing it by running mach bootstrap, which runs cargo install taplo-cli --locked even on NixOS. I like this change, the more we can get from nixpkgs the better.

delan (approved):

PR #30497, opened 2023-10-04
Bootstrap pkg-config and cmake on MacOS

### LINKED ISSUE #30354: Update Build Setup for Intel Mac
On a semi-fresh latest Intel Mac, here are the things I needed to do for `./mach bootstrap` to succeed:

1. Make a `pip` executable under my local `bin` path, to get around the system only having `pip3` now, and symlinking to `pip3` won't work, either.

Here's the pass-through `pip` command:
```
#!/bin/bash
pip3 "$@"
```

Looks like `mach` needs to be updated to use `pip3` on MacOS, instead of `pip`.

2. Install `pkg-config`.

I just needed `brew install pkg-config`

This is probably a missing dep from `mach`?

Anyways, hope this is helpful for something else trying to do the same, and maybe help patch up `mach`.

mrobinson:
These need to be installed in order to build so we can install them via
Homebrew. Do this by simply restoring the Homebrew bootstrapping logic
we had in place previously.

Fixes #30354.

---

- [x] These changes do not require tests because they just update the boostrap scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,5 +58,8 @@ Sessionx.vim
 # Layout debugger trace files
 layout_trace*
 
+# Package managers
+support/macos/Brewfile.lock.json
+
 # direnv
 .envrc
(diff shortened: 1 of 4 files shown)


jdm (approved):

mrobinson:
> Wrong issue link?

Oh yep. That should be fixed now.

PR #30505, opened 2023-10-05
chore(selector): update cssparser@0.33

j-mendez:
Update `cssparser` to latest `0.33`.

The build seems to fail locally for me on master not sure if the crate changes implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -644,7 +644,7 @@ dependencies = [
  "byteorder",
  "canvas_traits",
  "crossbeam-channel",
- "cssparser",
+ "cssparser 0.29.6",
  "euclid",
  "fnv",
  "font-kit",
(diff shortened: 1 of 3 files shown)


mrobinson:
`style` is a vendored crate so we wait until the dependency is updated upstream before changing this.

mrobinson:
FWIW, I think it'd be good to make it more obvious that this crate is vendored.

j-mendez:
> `style` is a vendored crate so we wait until the dependency is updated upstream before changing this.

I accidentally messed up the first commit and named it `selector`.  Ok that makes sense on the merging for upstream. Is the same the case for the `selector` crate?

mrobinson:
> > `style` is a vendored crate so we wait until the dependency is updated upstream before changing this.
> 
> I accidentally messed up the first commit and named it `selector`. Ok that makes sense on the merging for upstream. Is the same the case for the `selector` crate?

Yes, both `selectors` and `style` have upstreams in the Gecko repository.

mrobinson:
Closing for this reason. Sorry for the confusion.

PR #30507, opened 2023-10-05
Move `*_traits` and other shared types to `shared` directory

mrobinson:
This is the start of the organization of types that are in their own
crates in order to break dependency cycles between other crates. The
idea here is that putting these packages into their own directory is the
first step toward cleaning them up. They have grown organically and it
is difficult to explain to new folks where to put new shared types. Many
of these crates contain more than traits or don't contain traits at all.

Notably, `style_traits` isn't touched because it is vendored from
Gecko. Eventually this will move to `third_party`.

---


mrobinson:
I am fine to give this subdirectory a name other than `components/shared`. My main goal here is to organize these into a subdirectory and eventually give them understandable names and a better organization.

jdm:
script_traits isn't vendored. Did you mean something else?

mrobinson:
> script_traits isn't vendored. Did you mean something else?

Whoops. I meant 'style_traits' here.

components/canvas/Cargo.toml
```diff
@@ -38,12 +38,12 @@ servo_arc = { path = "../servo_arc" }
 servo_config = { path = "../config" }
 sparkle = { workspace = true }
 style = { path = "../style" }
-style_traits = { path = "../style_traits" }
+style_traits = { workspace = true }
```
mukilan, line 41: This now references the  'style_traits'  in workspace which has the 'servo' feature enabled. Does this have any negative implications?  My understanding is features are `unioned` together during crate compilation, so compile time shouldn't be impacted. But wouldn't this allow `canvas` to now accidentally refer to implementation details exposed using `servo` feature ?

Should we specify `features = [ "servo" ]` wherever it is needed (`components/servo/Cargo.toml`)  and not enabled it in the workspace?

components/shared/compositing/Cargo.toml
```diff
@@ -11,19 +11,19 @@ name = "compositing_traits"
...
-profile_traits = { path = "../profile_traits" }
-script_traits = { path = "../script_traits" }
-servo_url = { path = "../url" }
-style_traits = { path = "../style_traits" }
+msg = { workspace = true }
+net_traits = { workspace = true }
+profile_traits = { workspace = true }
+script_traits = { workspace = true }
+servo_url = { path = "../../url" }
+style_traits = { workspace = true }
```
mukilan, line 27: Same comment as above i.e accidentally importing types with `feature = servo` enabled in workspace.

mukilan (approved):
I just have one comment about the enabling of features in workspace, but otherwise looks good to me!

components/canvas/Cargo.toml
mrobinson: You're correct on all the details, yet I don't think it's an issue that canvas accesses parts of `style_traits` that use the `servo` feature. In general, all of servo should have access to the `servo` version of style and it's better that it's more widely available to all crates now.

mrobinson:
Thanks for the review!

components/canvas/Cargo.toml
mukilan: Hmm...of the 12 (not sure if I miscounted) consumers of `style_traits` only 2 need the `servo` feature, so it seemed to me that the other crates were designed to be independent of those features. But I don't have much insight into what `style_trait` represents or how it ought to be consumed, so I'll defer to your judgement.

mrobinson: `servo` isn't really an optional feature for `style` and `style_traits`. You really need to use either `servo` or `gecko`. In this case, it doesn't make as much sense to support compilation without one of the two.

PR #30517, opened 2023-10-08
Fix filter clipping caused by `overflow: hidden`

EnnuiL:
This is an attempt to fix filter clipping which well, had limited success;
I suspect that feeding the clip chain id to the stacking context is only one piece of the puzzle (although maybe an important one), considering that while it does fix clipping caused by `overflow: hidden`, clipping caused by `clip` still doesn't affect CSS filters;

Please do note that once again, I wasn't able to test everything

---

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -425,8 +425,6 @@ impl StackingContext {
             None => return false,
         };
 
-        let clip_id: Option<wr::ClipId> = self.clip_chain_id.map(|id| wr::ClipId::ClipChain(id));
-
         // WebRender only uses the stacking context to apply certain effects. If we don't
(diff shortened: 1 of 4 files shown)


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -271,6 +271,9 @@ pub struct StackingContext {
     /// things like preserve-3d.
     spatial_id: wr::SpatialId,
 
+    /// The clip chain id of this fragment if it has one. Used for filter clipping.
```
mrobinson, line 274:
```suggestion
    /// The clip chain id of this stacking context if it has one. Used for filter clipping.
```

mrobinson (approved):

mrobinson:
Thanks for this fix!

PR #30518, opened 2023-10-09
Fix the location url that reverts to the old value while loading

atbrakhi:
Changing two things in this PR
1. When we start servo we always have a default_url, see `fn get_default_url()`, right now in the minibrowser the default is set to String::Default(). Changing it to default_url in minibrowser as well
2. To avoid flickering of the url in the location bar, I have introduce `PumpResult::HistoryChanged`, now we only update the url when history has changed ie. `HistoryChanged`

----
- [x] These changes is part of #30049

### DIFF
--- a/ports/servoshell/app.rs
+++ b/ports/servoshell/app.rs
@@ -51,6 +51,7 @@ enum PumpResult {
     /// period, but has to call [`Servo::present`] to perform page flip and tell Servo compositor
     /// to continue rendering.
     Resize,
+    HistoryChanged
 }
 
 impl App {
@@ -271,16 +272,10 @@ impl App {
             }
 
             // Consume and handle any events from the Minibrowser.
-            if let Some(mut minibrowser) = app.minibrowser() {
+            if let Some(minibrowser) = app.minibrowser() {
                 let browser = &mut app.browser.borrow_mut();
                 let app_event_queue = &mut app.event_queue.borrow_mut();
                 minibrowser.queue_embedder_events_for_minibrowser_events(browser, app_event_queue);
-                if minibrowser.update_location_in_toolbar(browser) {
-                    // Update the minibrowser immediately. While we could update by requesting a
-                    // redraw, doing so would delay the location update by two frames.
-                    minibrowser
-                        .update(window.winit_window().unwrap(), "update_location_in_toolbar");
-                }
             }
 
             match app.handle_events() {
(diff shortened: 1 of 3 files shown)


ports/servoshell/minibrowser.rs
```diff
@@ -153,7 +154,7 @@ impl Minibrowser {
 
         match browser.current_url_string() {
             Some(location) if location != self.location.get_mut() => {
-                self.location = RefCell::new(location.to_owned());
+                self.location = RefCell::new(self.location.get_mut().to_string());
```
delan, line 157: this change means the location never updates when you click a link :(

delan (changes requested):

ports/servoshell/minibrowser.rs
atbrakhi: Ahh, how did I miss that?! 😅 I'll take a step back and rethink how we can resolve this without causing any issues. I'm going to convert this PR to a draft for now. Thanks!

delan: no worries! btw don’t be afraid to change how the underlying events work or add a new event, if their behaviour is inadequate. and be aware of the edge case of navigating to http 204 :p

```js
// node test.js, then go to http://localhost:8000
const http = require("node:http");
const server = http.createServer((req, res) => {
    res.statusCode = 204;
    res.end();
});
server.listen(8000);
```

atbrakhi: @delan as we discussed #30696 
I think this PR is ready for review now.

ports/servoshell/browser.rs
```diff
@@ -277,9 +277,13 @@ where
         self.event_queue.push(event);
     }
 
-    /// Returns true iff the caller needs to manually present a new frame.
-    pub fn handle_servo_events(&mut self, events: Vec<(Option<BrowserId>, EmbedderMsg)>) -> bool {
+    /// Returns true if the caller needs to manually present a new frame.
+    pub fn handle_servo_events(
+        &mut self,
+        events: Vec<(Option<BrowserId>, EmbedderMsg)>,
+    ) -> (bool, bool) {
```
delan, line 284: can we make this more self-documenting with a struct { need_present, history_changed }?

ports/servoshell/app.rs
```diff
@@ -396,9 +414,17 @@ impl App {
...
         loop {
             // Consume and handle those embedder messages.
-            need_present |= browser.handle_servo_events(embedder_messages);
+            let (new_need_present, new_history_changed) =
+                browser.handle_servo_events(embedder_messages);
+            need_present |= new_need_present;
+            history_changed |= new_history_changed;
+
+            if history_changed {
+                return Some(PumpResult::HistoryChanged);
```
delan, line 426: i checked the values returned by handle_servo_events, and sometimes both flags are true. when that happens, we won’t present the frame, which could make the compositor hang. what do you think about reworking PumpResult to allow for this? i think the cases should be something like

- Shutdown
- Continue { history_changed: bool, Present } where Present can be
    - Immediate
    - Deferred
    - None

delan (commented):
This is coming along well!

ports/servoshell/app.rs
atbrakhi: done in [7b74ee5](https://github.com/servo/servo/pull/30518/commits/7b74ee5350cd67d25ccc8489d7864c9fbd0b1f9e)

delan (approved):
Looks good, with one suggestion :D

PR #30519, opened 2023-10-09
Improve line box block size calculation

### LINKED ISSUE #30559: Ending line with space can shorten height of line box
```html
<!DOCTYPE html>
Should be a square:
<div style="background: green; color: transparent; font-size: 10px; width: 10em; line-height: 10">
  <span style="line-height: 5">X</span>
</div>
Should be a square:
<div style="background: green; color: transparent; font-size: 10px; width: 10em; line-height: 10">
  <span style="line-height: 5">X </span>
</div>
```

| Actual | Expected |
| - | - |
|  |  |

mrobinson:
Improve the calculation of the block size of line boxes and all their
component elements. Even empty spans can increase the size of the line
based on their font-size. Elements that have a line-height should
increase the block size of the line, but that setting should not effect
their own size.

In addition to the new passes there are some new failures

Failing because a progression exposes the real issue these tests are
testing:

- css/css-color/t32-opacity-offscreen-multiple-boxes-1-c.xht
- css/css-color/t32-opacity-offscreen-multiple-boxes-2-c.xht

Likely failing because of vertical-align and another sizing issue:

- css/css-transforms/perspective-untransformable-no-stacking-context.html

Failing because a progression reveals another failure:

 - html/rendering/non-replaced-elements/hidden-elements.html

---

- [x] This change fixes #30559.

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -108,14 +108,11 @@ struct LineUnderConstruction {
     /// is trimmed from the end.
     trailing_whitespace_advance: Length,
 
-    /// The currently calculated block size of this line, taking into account all inline
-    /// content already laid out into [`LineItem`]s. Later content may increase the block
-    /// size.
-    block_size: Length,
+    /// The maximum block size of all boxes that ended and are in progress in this line.
+    max_block_size: Length,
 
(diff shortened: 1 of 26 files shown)


mukilan (approved):
Nice!

PR #30539, opened 2023-10-11
Allow raising FloatContext ceiling after processing box with overflow

### LINKED ISSUE #30304: Floats are pushed down by overflowing block content
```html
<!DOCTYPE html>
<div style="height: 50px; font-size: 100px; background: blue">line content</div>
<div style="float: left; height: 50px; width: 50px; font-size: 100px; background: pink"></div>
```

Servo:
image

Firefox:
image

It seems that the IFC layout is pushing down the float ceiling based on the line height, but the containing block size should actually limit how far the ceiling moves down.

cc @Loirooriol

mrobinson:
When a box has overflow, any floats placed in that box will lower the float ceiling into the overflow. If no float is placed in the box though, the ceiling should be the block position where the overflow starts. We already know where this is, because we might be passing a negative value for the new block position after processing a box (content_size - actual_size would be negative). This negative value never raises the ceiling though since a maximum is used.

In the case that there is overflow, this change allows raising the ceiling, but never passed the lowest float. This necessitates keeping two values for the ceiling: one for floats and one for non-floats.

Fixes #30304.

---

- [x] These changes fix #30304.


mrobinson:
There are no tests attached to this change, because this will fix 2 failing tests in #30519.

mrobinson:
Okay. I think this is ready for review.

Loirooriol (approved):

Issue #30535, opened 2023-10-11
Tackle 2D canvas performance issues with Cookie Clicker


EnnuiL:
Currently, Servo struggles a lot with rendering Cookie Clicker's canvases, especially when there are actions going on, and so far, I have explored three areas of interest:

- [ ] `ctx.fillText` and 2D canvas text rendering (lag caused by text particles)
  - I suspect that it would be interesting to replace some of its logic with more standard ones; maybe using HarfBuzz could help? Who knows; I just know that by effectively unimplementing the feature, it ended exposing itself as a bottleneck for rendering, and by disabling the "Numbers" option, this same lag is alleviated
- [ ] `ctx.drawImage` (lag caused by an excess of images, easily caused by cookie particles)
  - This affects anything involving drawing an image to the 2D canvas, and after digging a lot? I could determine the culprit being the way the image data is being fed to `canvas::canvas_data::draw_image`
    
    Seeing how WebGL had handled a similar situation, I decided to try something and so I replaced the `ByteBuf` with the image pixels' `IpcSharedMemory` in a really messy way; the lag ends being greatly reduced on Steam's browser version, although the online browser version locks itself once a language is chosen
    
    I suspect that a proper solution would be replacing the transferring of image data with the transferring of `ImageKey`s somehow; I still have to dig further into how the other areas handle this scenario
- [ ] The 2D canvas backend (currently raqote)
  - Unfortunately, Pathfinder appears to be a dead project now; however, software rendering doesn't need to be *that* bad;
  
    On the process of thinking about potential optimization, I remembered that tiny_skia existed as an alternative and it did promise faster performance compared to raqote; but promises are promises, so I decided to mess around with a prototype of a tiny_skia backend, and it works! for Cookie Clicker and Cookie Clicker only! It currently fails a lot of WPT tests, the whole code is a mess, and the canvas code will need to be adapted for it (and the absence of the azure backend), but it does seem promising!
    
    I'm still not on a point where I can say "tiny_skia is definitely better than raqote", but considering the slight performance gains that I got despite the bottlenecks and considering that tiny_skia appears to be the more used/better-maintained crate nowadays, I find it something to be considered

For testing against potential bottlenecks, the Steam version's copy of the browser version (the Steam additions are a separate patch on top of it) proved to be indispensable for quickly iterating through solutions

Maybe these points should be split into separate issues once more progress is made

EnnuiL:
Status update:

[...]

PR #30544, opened 2023-10-14
Use `IpcSharedMemory` for `Canvas2dMsg::DrawImage`

EnnuiL:
This replaces the `ByteBuf` used in order to transmit image data to the 2D canvas thread with the image's `IpcSharedMemory`, in a manner similar to how image data is transmitted in WebGL.

The result is a speed-up on `ctx.drawImage` operations which manifests most obviously on Cookie Clicker's early stages (see #30535); There are, however, other bottlenecks, such as text rendering (alleviated by disabling the "Numbers" option) and having the multiple building canvases visible (alleviated by selecting another menu).

This, however, might make an inconsistent freeze caused by selecting a language happen more frequently, but I couldn't determine whenever this is an aggravation of a pre-existing issue or just plain bad luck. Please do double-check my stuff

---
- [x] These changes partially address #30535

### DIFF
--- a/components/canvas_traits/canvas.rs
+++ b/components/canvas_traits/canvas.rs
@@ -41,7 +41,7 @@ pub struct CanvasImageData {
 pub enum Canvas2dMsg {
     Arc(Point2D<f32>, f32, f32, f32, bool),
     ArcTo(Point2D<f32>, Point2D<f32>, f32),
-    DrawImage(Size2D<f64>, Rect<f64>, Rect<f64>, bool, IpcSharedMemory),
+    DrawImage(IpcSharedMemory, Size2D<f64>, Rect<f64>, Rect<f64>, bool),
     DrawEmptyImage(Size2D<f64>, Rect<f64>, Rect<f64>),
     DrawImageInOther(CanvasId, Size2D<f64>, Rect<f64>, Rect<f64>, bool),
     BeginPath,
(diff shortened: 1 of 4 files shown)


components/canvas_traits/canvas.rs
```diff
@@ -41,7 +41,8 @@ pub struct CanvasImageData {
 pub enum Canvas2dMsg {
     Arc(Point2D<f32>, f32, f32, f32, bool),
     ArcTo(Point2D<f32>, Point2D<f32>, f32),
-    DrawImage(Option<ByteBuf>, Size2D<f64>, Rect<f64>, Rect<f64>, bool),
+    DrawImage(Size2D<f64>, Rect<f64>, Rect<f64>, bool, IpcSharedMemory),
+    DrawEmptyImage(Size2D<f64>, Rect<f64>, Rect<f64>),
```
jdm, line 45: Would it make more sense to use `Option<IpcSharedMemory>` here instead of adding a new variant?

components/script/canvas_state.rs
```diff
@@ -272,7 +271,7 @@ impl CanvasState {
 
         let image_size = Size2D::new(img.width, img.height);
         let image_data = match img.format {
-            PixelFormat::BGRA8 => img.bytes.to_vec(),
+            PixelFormat::BGRA8 => img.bytes.clone(),
```
jdm, line 274: Why do we need to clone this?

jdm (approved):

jdm:
I also wonder if we should have a threshold here—using shared memory buffers for small images could be a performance loss.

PR #30546, opened 2023-10-14
Bump mozangle to 0.5.0

sagudev:
Companion PR to https://github.com/servo/mozangle/pull/69

---

### DIFF
--- a/components/script/dom/webglshader.rs
+++ b/components/script/dom/webglshader.rs
@@ -189,7 +189,7 @@ impl WebGLShader {
             options.set_unfoldShortCircuit(1);
             // Work around that Mac drivers handle struct scopes incorrectly.
             options.set_regenerateStructNames(1);
-            // mac intel
+            // TODO: Only apply this workaround to Intel hardware
             // Work around that Intel drivers on Mac OSX handle for-loop incorrectly.
             options.set_addAndTrueToLoopCondition(1);
             options.set_rewriteTexelFetchOffsetToTexelFetch(1);
--- a/components/script/dom/webglshader.rs
+++ b/components/script/dom/webglshader.rs
@@ -12,7 +12,7 @@ use canvas_traits::webgl::{
     WebGLShaderId, WebGLVersion,
 };
 use dom_struct::dom_struct;
-use mozangle::shaders::{ffi, BuiltInResources, Output, ShaderValidator};
+use mozangle::shaders::{BuiltInResources, CompileOptions, Output, ShaderValidator};
 
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::inheritance::Castable;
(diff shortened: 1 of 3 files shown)


sagudev:
try build: https://github.com/sagudev/servo/actions/runs/6519064140

mrobinson:
The try build looks good, so maybe this can be marked as "ready" and head to the merge queue?

sagudev:
> The try build looks good, so maybe this can be marked as "ready" and head to the merge queue?

This depends on last "good" https://github.com/servo/mozangle/pull/69 and it's still in the works.

With last good version, I mean version that compiles and works, but compiles things multiple times (and thus it's slow). But I am working on solution to better simulate moz.build files, so we build each angle module only once.

sagudev:
try build (still running): https://github.com/sagudev/servo/actions/runs/6548931708

sagudev:
Manual test on windows was ok.

mrobinson:
> @mrobinson we need new mozangle release published for this.

Sorry for the delay. The new version (0.5.0) of mozangle is now published.

components/script/dom/webglshader.rs
```diff
@@ -284,44 +170,36 @@ impl WebGLShader {
...
+
+            // Work around https://bugs.webkit.org/show_bug.cgi?id=124684,
+            // https://chromium.googlesource.com/angle/angle/+/5e70cf9d0b1bb
+            options.set_unfoldShortCircuit(1);
+            // Work around that Mac drivers handle struct scopes incorrectly.
+            options.set_regenerateStructNames(1);
+            // mac intel
+            // Work around that Intel drivers on Mac OSX handle for-loop incorrectly.
+            options.set_addAndTrueToLoopCondition(1);
+            options.set_rewriteTexelFetchOffsetToTexelFetch(1);
```
mrobinson, line 196: It looks like this isn't correct for Mac with Apple Silicon. Can we check the GPU vendor here?

components/script/dom/webglshader.rs
```diff
@@ -243,7 +128,8 @@ impl WebGLShader {
             EXT_frag_depth: ext.is_enabled::<EXTFragDepth>() as c_int,
 
             FragmentPrecisionHigh: 1,
-            ..default_validator()
+            //..default_validator()
```
mrobinson, line 131: It seems that this code should be removed?

mrobinson (changes requested):
Looks good, but just a few comments.

components/script/dom/webglshader.rs
sagudev: I do not know how (mozilla has special utils for this but we do not have them). Also this might be out of scope of this PR, as old code has same thing on all macs, I just refreshed options for new mozangle.

components/script/dom/webglshader.rs
```diff
@@ -284,44 +169,36 @@ impl WebGLShader {
...
+
+        if cfg!(target_os = "macos") {
+            options.set_removeInvariantAndCentroidForESSL3(1);
+
+            // Work around https://bugs.webkit.org/show_bug.cgi?id=124684,
+            // https://chromium.googlesource.com/angle/angle/+/5e70cf9d0b1bb
+            options.set_unfoldShortCircuit(1);
+            // Work around that Mac drivers handle struct scopes incorrectly.
+            options.set_regenerateStructNames(1);
+            // mac intel
```
mrobinson, line 192: I would change this to `TODO: Only apply this to workaround to Intel hardware` then.

mrobinson (approved):

components/script/dom/webglshader.rs
```diff
@@ -284,44 +169,36 @@ impl WebGLShader {
...
+            // TODO: Only apply this workaround to Intel hardware
+            // Work around that Intel drivers on Mac OSX handle for-loop incorrectly.
+            options.set_addAndTrueToLoopCondition(1);
+            options.set_rewriteTexelFetchOffsetToTexelFetch(1);
+        } else {
+            // We want to do this everywhere, but to do this on Mac, we need
+            // to do it only on Mac OSX > 10.6 as this causes the shader
+            // compiler in 10.6 to crash
+            options.set_clampIndirectArrayBounds(1);
+        }
```
mrobinson, line 201: I'm pretty sure we only support MacOS 10.10 and up, so this if statement can likely be removed in a followup.

sagudev:
@mrobinson land when you are ready

mrobinson:
Okay. I've sent this to the MQ. Thanks for all the work here @sagudev!

PR #30548, opened 2023-10-16
Add documentation to a few pref configs

wusyong:
Initial attempt to add documentation to `pref` types. Only the last field of the config can be documented due to proc-macro. (For example, `shell.homepage`'s `homepage` field) I'm only confident of adding these doc comments for now. We can add more later if there are people who understand what other prefs mean. 

---

- [x] These changes do not require tests because this is about adding doc comments. But the `./mach doc` command should pass.

### DIFF
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -511,7 +511,7 @@ mod gen {
                     #[serde(rename = "shell.native-titlebar.enabled")]
                     enabled: bool,
                 },
-                /// URL string of the search engine webpage like google and duckduckgo.
+                /// URL string of the search engine page (for example <https://google.com> or and <https://duckduckgo.com>.
                 searchpage: String,
             },
             webgl: {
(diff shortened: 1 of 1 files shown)


components/config/prefs.rs
```diff
@@ -490,22 +492,26 @@ mod gen {
...
                     enabled: bool,
                 },
                 #[serde(rename = "shell.native-orientation")]
                 native_orientation: String,
                 native_titlebar: {
+                    /// Enable native window's titlebar and decorations.
                     #[serde(rename = "shell.native-titlebar.enabled")]
                     enabled: bool,
                 },
+                /// URL string of the search engine webpage like google and duckduckgo.
```
mrobinson, line 514:
```suggestion
                /// URL string of the search engine page (for example <https://google.com> or and <https://duckduckgo.com>.
```

mrobinson (approved):

Issue #30577, opened 2023-10-18
Debug assertion failures in components/style/logical_geometry.rs:416


delan:
(#30567, #30509)

```
>>> /home/runner/work/servo/servo/components/style/logical_geometry.rs:416
assertion failed: `(left == right)`
  left: `(empty)`,
  left: `INLINE_REVERSED | RTL`,
 right: `(empty)` (thread Layout(1,1), at /home/runner/work/servo/servo/components/style/logical_geometry.rs:416)
 right: `INLINE_REVERSED | RTL` (thread Layout(1,1), at /home/runner/work/servo/servo/components/style/logical_geometry.rs:416)
```

* layout 2013
    * /css/css-logical/logical-values-float-clear-3.html
    * /css/css-logical/logical-values-float-clear-4.html
    * /css/css-position/multicol/static-position/vlr-ltr-rtl-in-multicol.tentative.html
    * /css/css-position/multicol/static-position/vlr-rtl-ltr-in-multicol.tentative.html
    * /css/css-position/multicol/static-position/vrl-ltr-rtl-in-multicol.tentative.html
    * /css/css-position/multicol/static-position/vrl-rtl-ltr-in-multicol.tentative.html
    * /css/css-position/multicol/vlr-ltr-rtl-in-multicols.tentative.html
    * /css/css-position/multicol/vlr-rtl-ltr-in-multicols.tentative.html
    * /css/css-position/multicol/vrl-ltr-rtl-in-multicols.tentative.html
    * /css/css-position/multicol/vrl-rtl-ltr-in-multicols.tentative.html
    * /css/css-position/static-position/htb-ltr-rtl.tentative.html
    * /css/css-position/static-position/htb-rtl-ltr.tentative.html
    * /css/css-position/static-position/vlr-ltr-rtl.tentative.html
    * /css/css-position/static-position/vlr-rtl-ltr.tentative.html
    * /css/css-position/static-position/vrl-ltr-rtl.tentative.html
    * /css/css-position/static-position/vrl-rtl-ltr.tentative.html

delan:
stack 1

1. DebugWritingMode::check_debug (components/style/logical_geometry.rs:416
2. Add::add for LogicalSize&lt;T> (components/style/logical_geometry.rs:564)
3. Floats::translate (components/layout/floats.rs:166)

* /css/css-logical/logical-values-float-clear-3.html
* /css/css-logical/logical-values-float-clear-4.html

stack 2

[...]

Issue #30573, opened 2023-10-18
Debug assertion failures in components/layout/flow.rs:1379


delan:
(#30567, #30509)

```
% ( set -- 2013/filtered-test-wpt.*.json; for i; do < $i jq -r '.[] | select(.stack != null) | .stack' | rg -o --pcre2 '(?<=, at )[^()]+(?=[)]$)'; done | sort -u | while read -r site; do echo ">>> $site"; for i; do < $i jq -r '.[] | select(.stack != null) | select(.stack | contains("'"$site"'")) | .stack'; done | rg ', at |^assertion failed:|^  left:|^ right:' | sort -u; done ) 
>>> components/layout/flow.rs:1379
assertion failed: !descendant_link.has_reached_containing_block (thread Layout(1,1), at components/layout/flow.rs:1379)
```

* layout 2013
    * /css/CSS2/positioning/abspos-float-with-inline-container.html

delan:
layout 2013 has been removed, so i think we can close this now.

Issue #30572, opened 2023-10-18
Debug assertion failures in components/layout/block.rs:1765


delan:
(#30567, #30509)

```
% ( set -- 2013/filtered-test-wpt.*.json; for i; do < $i jq -r '.[] | select(.stack != null) | .stack' | rg -o --pcre2 '(?<=, at )[^()]+(?=[)]$)'; done | sort -u | while read -r site; do echo ">>> $site"; for i; do < $i jq -r '.[] | select(.stack != null) | select(.stack | contains("'"$site"'")) | .stack'; done | rg ', at |^assertion failed:|^  left:|^ right:' | sort -u; done ) 
>>> components/layout/block.rs:1765
assertion failed: !self.is_inline_flex_item() (thread Layout(1,1), at components/layout/block.rs:1765)
```

* layout 2013
    * /html/rendering/non-replaced-elements/the-fieldset-and-legend-elements/fieldset-flexbox.html

delan:
layout 2013 has been removed, so i think we can close this now.

Issue #30571, opened 2023-10-18
Debug assertion failures in components/constellation/constellation.rs:3171


delan:
(#30567, #30509)

```
% ( set -- 2020/filtered-test-wpt.*.json; for i; do < $i jq -r '.[] | select(.stack != null) | .stack' | rg -o --pcre2 '(?<=, at )[^()]+(?=[)]$)'; done | sort -u | while read -r site; do echo ">>> $site"; for i; do < $i jq -r '.[] | select(.stack != null) | select(.stack | contains("'"$site"'")) | .stack'; done | rg ', at |^assertion failed:|^  left:|^ right:' | sort -u; done )
>>> /home/runner/work/servo/servo/components/constellation/constellation.rs:3171
assertion failed: `(left == right)`
  left: `200.0x100.0`,
  left: `200.0x150.0`,
  left: `200.0x200.0`,
  left: `250.0x250.0`,
  left: `256.0x256.0`,
  left: `300.0x100.0`,
  left: `300.0x150.0`,
  left: `300.0x300.0`,
  left: `300.0x400.0`,
  left: `400.0x250.0`,
  left: `400.0x400.0`,
  left: `500.0x300.0`,
  left: `800.0x600.0`,
 right: `0.0x0.0` (thread Constellation, at /home/runner/work/servo/servo/components/constellation/constellation.rs:3171)
```

```
% ( set -- 2013/filtered-test-wpt.*.json; for i; do < $i jq -r '.[] | select(.stack != null) | .stack' | rg -o --pcre2 '(?<=, at )[^()]+(?=[)]$)'; done | sort -u | while read -r site; do echo ">>> $site"; for i; do < $i jq -r '.[] | select(.stack != null) | select(.stack | contains("'"$site"'")) | .stack'; done | rg ', at |^assertion failed:|^  left:|^ right:' | sort -u; done )
>>> /home/runner/work/servo/servo/components/constellation/constellation.rs:3171
assertion failed: `(left == right)`
  left: `300.0x150.0`,
 right: `0.0x0.0` (thread Constellation, at /home/runner/work/servo/servo/components/constellation/constellation.rs:3171)
```

* layout 2020
    * /cookies/schemeful-same-site/schemeful-iframe-subresource.tentative.html
    * /css/CSS2/linebox/iframe-in-block-in-inline.html
    * /css/css-backgrounds/background-image-shared-stylesheet.html
    * /css/css-fonts/font-size-adjust-reload.html
[...]

PR #30578, opened 2023-10-18
Temporarily convert existing cfg(debug_assertions) crashes to warnings

delan:
#30509 enables debug assertions in local dev builds and on CI, but existing debug assertions mean that hundreds of tests would now crash, and @mrobinson and I believe that updating their expectations accordingly risks worsening our test coverage until the underlying bugs are fixed.

This patch temporarily converts the offending assertions to warnings.

---

- [x] These changes are related to #30568
- [x] These changes are related to #30569
- [x] These changes are related to #30570
- [x] These changes are related to #30571
- [x] These changes are related to #30572
- [x] These changes are related to #30573
- [x] These changes are related to #30577

### DIFF
--- a/components/style/logical_geometry.rs
+++ b/components/style/logical_geometry.rs
@@ -570,6 +570,22 @@ impl<T: Add<T, Output = T>> Add for LogicalSize<T> {
     }
 }
 
+// TODO(servo#30577) remove this once underlying bugs are fixed
+impl<T: Add<T, Output = T>> LogicalSize<T> {
+    #[inline]
+    pub fn add_or_warn(self, other: LogicalSize<T>) -> LogicalSize<T> {
+        #[cfg(debug_assertions)]
+        if !(self.debug_writing_mode.mode == other.debug_writing_mode.mode) {
+            log::warn!("debug assertion failed! self.debug_writing_mode.mode == other.debug_writing_mode.mode");
+        }
+        LogicalSize {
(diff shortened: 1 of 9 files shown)


components/style/logical_geometry.rs
```diff
@@ -570,6 +570,22 @@ impl<T: Add<T, Output = T>> Add for LogicalSize<T> {
...
+            log::warn!("debug assertion failed! self.debug_writing_mode.mode == other.debug_writing_mode.mode");
+        }
+        LogicalSize {
+            debug_writing_mode: self.debug_writing_mode,
+            inline: self.inline + other.inline,
+            block: self.block + other.block,
+        }
+    }
+}
+
```
mrobinson, line 588: Do you mind moving this out of `style` and into the `layout` crate? It's not impossible that we'll remove legacy layout before fixing this one (since legacy layout is a low priority) and we eventually want to push our changes to `style` upstream.

mrobinson (approved):

components/style/logical_geometry.rs
delan: hmm, we can’t avoid crashing without some kind of modification to style, because debug_writing_mode is a private field. is there a way we can mark this as not to be upstreamed?

mrobinson: Oh, in that case, I guess it is fine to put into `style` for now.

PR #30579, opened 2023-10-18
Anonymous boxes that wrap inlines should not inherit overflow

### LINKED ISSUE #30562: `overflow` can make float inside inline disappear
```html
<!DOCTYPE html>
<p>Test passes if there is a filled green square and <strong>no red</strong>.</p>
<div style="overflow: hidden; width: 100px; height: 100px; background: red">
  <div style="height: 50px; background: green"></div>
  <span>
    <div style="float: left; width: 100px; height: 50px; background: green"></div>
  </span>
</div>
```

| Actual | Expected |
| - | - |
|  |  |

Note it works fine when forcing a BFC with `display: flow-root` instead of `overflow: hidden`, so it seems a problem specific to the latter.

mrobinson:
In legacy layout, anonymous text wrappers were inheriting the `overflow`
and `text-overflow` properties. This results in the creation of extra
clipping for these anonymous wrappers which could clip away floats. We
will likely implement `text-overflow` differently in non-legacy layout.

This change marks all legacy layout pseudo elements as "legacy" and also
adds a new pseudo element for non-legacy layout that does not inherit
`overflow`.

Fixes #30562.

---

- [x] These changes fix #30562.

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -228,7 +228,7 @@ fn traverse_pseudo_element_contents<'dom, Node>(
                         .stylist
                         .style_for_anonymous::<Node::ConcreteElement>(
                             &context.shared_context().guards,
-                            &PseudoElement::ServoText,
+                            &PseudoElement::ServoAnonymousBox,
                             &info.style,
                         )
                 });
--- a/components/layout/construct.rs
+++ b/components/layout/construct.rs
@@ -805,7 +805,7 @@ where
                 .stylist
                 .style_for_anonymous::<ConcreteThreadSafeLayoutNode::ConcreteElement>(
                     &context.guards,
-                    &PseudoElement::ServoText,
+                    &PseudoElement::ServoLegacyText,
                     &style,
                 );
             if node_is_input_or_text_area {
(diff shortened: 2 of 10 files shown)


components/layout_2020/dom_traversal.rs
```diff
@@ -228,7 +228,7 @@ fn traverse_pseudo_element_contents<'dom, Node>(
                         .stylist
                         .style_for_anonymous::<Node::ConcreteElement>(
                             &context.shared_context().guards,
-                            &PseudoElement::ServoText,
+                            &PseudoElement::ServoAnonymousBlock,
```
Loirooriol, line 231: Actually this is not a block, there is even an assert for `display: inline`.
Maybe it should be renamed to `ServoAnonymousBox`?

mrobinson: Okay. I've gone ahead and renamed this `PseudoElement` variant.

Loirooriol (approved):
So position-sticky-hyperlink.html is now failing but that's expected because sticky positioning is not implemented.

Before this patch it was falsely passing because neither the test nor the reference were showing the link due to propagating the overflow.

PR #30580, opened 2023-10-18
Add a heuristic to detect collapse through in IFCs

mrobinson:
When inline formatting contexts collapse through, margins before and
after can collapse. Previously the IFC code was assuming that margins
never collapses through them. This change adds a heuristic to detect the
most common cases of collapse through. Note that there are still
situations where do not know if an IFC will collapse through until we
actually try to lay it out. A proper implementation would be able to
rewind and relayout to fix float placement.

---

### DIFF
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -167,7 +167,7 @@ jobs:
     name: Results
     needs: ["parse-comment", "run-try"]
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


Loirooriol (approved):

mrobinson:
I'm starting to think that we should put this aside for now...as it's turning out to be more complicated than I expected and I don't really have time to tackle it right now.

PR #30581, opened 2023-10-18
Don't panic when no font is found for a TextRun

mrobinson:
Instead of panicking when no found is found for a TextRun, instead print
a warning. This prevents panics on pages with very large font sizes.

---

### DIFF
--- a/components/layout/text.rs
+++ b/components/layout/text.rs
@@ -14,7 +14,7 @@ use gfx::font::{self, FontMetrics, FontRef, RunMetrics, ShapingFlags, ShapingOpt
 use gfx::text::glyph::ByteIndex;
 use gfx::text::text_run::TextRun;
 use gfx::text::util::{self, CompressionMode};
-use log::debug;
+use log::{debug, warn};
 use range::Range;
 use servo_atoms::Atom;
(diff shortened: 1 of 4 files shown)


mrobinson:
@mukilan I think this will unblock #30162.

mukilan (approved):

mukilan:
Thanks @mrobinson ! For #30162 , this change will prevent the panics but I also like to know why the tests panicked at all  - I have forgotten where I left the investigation, but I think the fallback fonts from DejaVu family are available on both 20.04 and 22.04, so the tests shouldn't have failed.

mrobinson:
> Thanks @mrobinson ! For #30162 , this change will prevent the panics but I also like to know why the tests panicked at all - I have forgotten where I left the investigation, but I think the fallback fonts from DejaVu family are available on both 20.04 and 22.04, so the tests shouldn't have failed.

All of these tests try to create text with a very large font size. The font matching code isn't matching because the size is too big. I'm still looking into that. I think maybe other browser limit the maximum font size.

mrobinson:
The latest version issues a warning in two more places, but is otherwise the same.

PR #30598, opened 2023-10-21
Fix servoshell can't respond when there's no minibrowser

wusyong:
When servoshell launch with `no-minibrowser`, it can become unresponsive and the cursor will start spinning in both debug and release build. And in the rest of the case, the input seems pretty laggy.
This can be tested by running some WebGL example like: https://madebyevan.com/webgl-water/
I found this is because no `ReadyToPresent` will be sent in `no-minibrowser` status and hence compositor can keep rendering which makes the event loop hang.
I'm not sure if this PR is the right way to resolve it, but I tried to send `ReadyToPresent` event no matter `external_presetn` is true or false.

---

- [x] These changes do not require tests but both debug and release build shouldn't be laggy or hang.


delan:
My initial impression is that servoshell worked fine\* without external present prior to adding the minibrowser (#29976), and we later added external present for the minibrowser only (#30322), so I wonder if there was a deeper regression?

\* for the most part; see #23899

delan:
I was wrong! a7bd9f0d43f4de13c958eaea53709e8fc0d162c8, the commit before #29976, is also affected:

https://github.com/servo/servo/assets/465303/6a6ba3a3-2be2-48ea-8a63-6fe4938f42c3

components/compositing/compositor.rs
```diff
@@ -1856,16 +1856,12 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-                warn!("Sending event to constellation failed ({:?}).", e);
-            }
-        } else {
-            self.present();
+        self.waiting_on_present = true;
+        let msg = ConstellationMsg::ReadyToPresent(
+            self.root_content_pipeline.top_level_browsing_context_id,
+        );
+        if let Err(e) = self.constellation_chan.send(msg) {
+            warn!("Sending event to constellation failed ({:?}).", e);
```
delan, line 1864: this effectively makes the `external_present` flag do nothing. can we revert this and set the flag unconditionally in servoshell, or remove the flag entirely?

components/compositing/compositor.rs
```diff
@@ -1856,16 +1856,12 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         };
 
         // Perform the page flip. This will likely block for a while.
```
delan, line 1858: (if we remove the flag, we should update this comment)

ports/servoshell/app.rs
```diff
@@ -112,10 +112,6 @@ impl App {
         // frame, set this to false, so we can avoid an unnecessary recomposite.
         let mut need_recomposite = true;
 
-        // If we have a minibrowser, ask the compositor to notify us when a new frame
-        // is ready to present, so that we can paint the minibrowser then present.
```
delan, line 116: this explanation is important for understanding the interaction between servo and the minibrowser when drawing and presenting frames. can we update it, instead of removing it?

delan (commented):
Thanks for looking into this!

components/compositing/compositor.rs
wusyong: I removed `external_present` and updated this comment.

ports/servoshell/app.rs
wusyong: I tried to move this explanation to doc comment in all variants of `PumpResult`. wdyt?

ports/servoshell/app.rs
```diff
@@ -40,8 +40,13 @@ pub struct App {
 
 /// Action to be taken by the caller of [`App::handle_events`].
 enum PumpResult {
+    /// The caller should shutdown Servo and its related contexnt.
```
delan, line 43: nit: “shut down”, “context”

ports/servoshell/app.rs
```diff
@@ -40,8 +40,13 @@ pub struct App {
...
     Shutdown,
+    /// A new frame is ready to present. The caller can paint other things themselves during this
+    /// period, but has to call [`Servo::present`] to tell Servo compositor continue rendering.
```
delan, line 46: suggest “to perform the page flip and tell the Servo compositor to continue rendering”, because these are two separate things that are currently coupled but we may change later

ports/servoshell/app.rs
```diff
@@ -40,8 +40,13 @@ pub struct App {
...
     ReadyToPresent,
+    /// The size has changed. The caller can paint other things themselves during this
+    /// period, but has to call [`Servo::present`] to tell Servo compositor continue rendering.
```
delan, line 49: suggest “to perform the page flip and tell the Servo compositor to continue rendering”, because these are two separate things that are currently coupled but we may change later

components/compositing/compositor.rs
```diff
@@ -1855,17 +1850,14 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-            let msg = ConstellationMsg::ReadyToPresent(
-                self.root_content_pipeline.top_level_browsing_context_id,
-            );
-            if let Err(e) = self.constellation_chan.send(msg) {
-                warn!("Sending event to constellation failed ({:?}).", e);
-            }
-        } else {
-            self.present();
+        // Perform the page flip. Nottify embedder that servo is ready to present.
+        // Embedder should call `present` to tell compositor to continue rendering.
```
delan, line 1854: remove “perform the page flip”, because we no longer do it here

ports/servoshell/app.rs
delan: looks good!

delan (approved):
Sorry for the delay! Let’s land this :D

mrobinson:
This seems to be the failure:

```
alled `Option::unwrap()` on a `None` value (thread main, at ports/servoshell/app.rs:294)
   1: core::fmt::write
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/core/src/fmt/mod.rs:1254:17
   2: std::io::Write::write_fmt
   3: servo::backtrace::print
   4: servo::main2::main::{{closure}}
   5: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/alloc/src/boxed.rs:2002:9
      std::panicking::rust_panic_with_hook
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:696:13
   6: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:581:13
   7: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/sys_common/backtrace.rs:150:18
   8: rust_begin_unwind
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:579:5
   9: core::panicking::panic_fmt
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/core/src/panicking.rs:64:14
  10: core::panicking::panic
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/core/src/panicking.rs:114:5
  11: servo::events_loop::EventsLoop::run_forever
  12: servo::app::App::run
  13: servo::main2::main
  14: std::sys_common::backtrace::__rust_begin_short_backtrace
  15: std::rt::lang_start::{{closure}}
  16: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/core/src/ops/function.rs:287:13
      std::panicking::try::do_call
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:487:40
      std::panicking::try
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:451:19
      std::panic::catch_unwind
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panic.rs:140:14
      std::rt::lang_start_internal::{{closure}}
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/rt.rs:148:48
      std::panicking::try::do_call
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:487:40
      std::panicking::try
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panicking.rs:451:19
      std::panic::catch_unwind
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/panic.rs:140:14
      std::rt::lang_start_internal
             at /rustc/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/std/src/rt.rs:148:20
  17: _main
  18: <unknown>
```

wusyong:
@mrobinson The crash is caused by the unwrap when handling `ReadyToPresent` because it will always send the event now, even if it's headless mode. Fixed it with `if let` handling.

wusyong:
Looks like there are some unexpected wpt fails. I'll test them locally later.

wusyong:
@delan Does test-wpt run in headless mode? I'm guessing control flow of event loop might be the cause if there's no interaction.

wusyong:
I tested on my fork action and seems like it can pass with latest commit fix:
https://github.com/wusyong/servo/actions/runs/6806696811

PR #30608, opened 2023-10-24
Make LineItems a token stream on the root

mrobinson:
Flattening the LineItem tree into a token stream will allow for handling
the case where an unbreakable line segment spans multiple inline boxes
which might have different hierarchies. This change also fixes the
handling of the second anonymous fragment of a block-in-inline-split.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use std::cell::OnceCell;
+use std::vec::IntoIter;
 
 use app_units::Au;
 use atomic_refcell::AtomicRef;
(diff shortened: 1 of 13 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -982,80 +992,45 @@ impl InlineFormattingContext {
...
     }
 
     fn layout_into_line_item(
         &mut self,
         layout_context: &LayoutContext,
-        inline_position: &mut Length,
-        at_end_of_inline_element: bool,
+        is_first_fragment: bool,
     ) -> InlineBoxLineItem {
         // If we are finishing in order to fragment this InlineBox into multiple lines, do
```
mukilan, line 1024: Is this comment still relevant, now that we've removed `at_end_of_inline_element`?

mukilan (approved):

mrobinson:
Thanks for the review!

PR #30613, opened 2023-10-24
Replace the time crate with std::time in components/compositing

### LINKED ISSUE #30150: Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate

reidswan:
Replaces the `time` crate in the `components/compositing` package with `std::time`

---

- [X] These changes contribute to #30150

### DIFF
--- a/components/profile_traits/time.rs
+++ b/components/profile_traits/time.rs
@@ -2,11 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
 use ipc_channel::ipc::IpcSender;
 use log::warn;
 use serde::{Deserialize, Serialize};
 use servo_config::opts;
-use time::precise_time_ns;
 
 #[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
 pub struct TimerMetadata {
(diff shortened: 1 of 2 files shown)


components/profile_traits/time.rs
```diff
@@ -160,3 +167,9 @@ pub fn send_profile_data(
 ) {
     profiler_chan.send(ProfilerMsg::Time((category, meta), (start_time, end_time)));
 }
+
+pub(crate) fn elapsed_since_epoch() -> Duration {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+}
```
mrobinson, line 175: I would just call `SystemTime::now()` instead of exporting this new function. Perhaps it makes sense to have a series of helpers that make dealing with time easier, but I'm not sure the `profile_traits` crate is the right place for this.

mrobinson (changes requested):
Great start! One comment below, but this is looking good.

components/profile_traits/time.rs
reidswan: Thanks for the feedback! I have gotten rid of the offending function. 

When you say "just call `SystemTime::now()`", do you mean the whole `SystemTime::now().duration_since(UNIX_EPOCH)...`? Because I can't see a way to get nanoseconds from the SystemTime directly, even through the UNIX_EPOCH is just a zero value for this particular struct. Am I missing something?

mrobinson:
Nicely done. I don't think this needs new tests as the behavior shouldn't change.

mrobinson (approved):

Issue #30599, opened 2023-10-23
Add cli option to start servo in fullscreen mode


Mattia98:
Just like the title says. 
It would be extremely helpful for kiosk applications or similar to be able to start servo in fullscreen mode. An argument like `--fullscreen` would suffice.
Usage case could be for display-only signage displays, where servo is started in fullscreen mode to display a certain webpage. Or perhaps for some kiosk application where a user is able to use the touchscreen to navigate a simple webui.
In any case, it would be useful to have a `--fullscreen` option.

Quackdoc:
One kiosk application use case I look forward to eventually using is TV apps, example being youtube TV with the correct user agent

PR #30614, opened 2023-10-25
servoshell: add fullscreen option

MarybethGasman:
These changes add a cli option to start servo in fullscreen mode. As requested in #30599.
It is currently in draft form as we need to have further discussions to determine whether we truly need it.
```sh
servo --fullscreen
```
No tests for now.

### DIFF
--- a/ports/servoshell/headed_window.rs
+++ b/ports/servoshell/headed_window.rs
@@ -92,6 +92,7 @@ impl Window {
         // unstyled content is white and chrome often has a transparent background). See issue
         // #9996.
         let visible = opts.output_file.is_none() && !no_native_titlebar;
+        let fullscreen = opts.fullscreen;
 
         let win_size: DeviceIntSize = (win_size.to_f32() * window_creation_scale_factor()).to_i32();
         let width = win_size.to_untyped().width;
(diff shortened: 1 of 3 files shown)


delan (changes requested):
Thanks for the patch!

I think fullscreen generally means not having any window decorations, not just maximised. We should also return whether we started in fullscreen in the Window.fullscreen of our return value.

You can use WindowBuilder::with_fullscreen or winit Window::set_fullscreen for this, or even better, call .set_fullscreen(true) on our headed Window before returning it.

MarybethGasman:
Appreciate your advice.
I call set_fullscreen on our headed Window before returning it.

It works as expected when launching.
image

When I refresh page, the web content seems not displayed in fullscreen mode.
image

And I can't exit fullscreen mode or servo by Esc.

The Esc problem is because fullscreen is a whatwg web api.
It will do nothing if there is no fullscreen element in the document.
So I change the code to always Send EmbedderMsg first.

The web content not in fullscreen problem is solved by call .with_maximized on the window builder.
I have no idea why.
Am I missing something?

ports/servoshell/headed_window.rs
```diff
@@ -139,7 +142,7 @@ impl Window {
             .expect("Failed to create WR surfman");
 
         debug!("Created window {:?}", winit_window.id());
-        Window {
+        let window = Window {
```
Taym95, line 145: why this is needed?

mrobinson:
It seems like this PR is abandoned. For proper fullscreen support we need a bit more as well. For instance, there needs to be a way for content to request fullscreen and the UI to respond properly to this.

PR #30509, opened 2023-10-06
Enable debug assertions for all builds other than official releases

delan:
Servo should ideally never panic (see the discussion in #30458), but we still need a way to detect invalid states when hacking on Servo or testing on CI. Unfortunately our release profile has debug assertions disabled.

This patch enables debug assertions in the release profile, while adding a “production” profile that yields a release build without debug assertions, which we use when building nightly releases. This patch also:

* updates HACKING_QUICKSTART to document our profiles
* extends mach to accept a --profile argument like Cargo
* eliminates the std_test_override hack
* hardens #30537 such that we only try the current directory and its ancestors when we aren’t in the production profile

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #30567

PR #30627, opened 2023-10-27
Support running custom wpt tests in wpt runs (and add suport for webgpu runs)

### LINKED ISSUE #30144: Allow testing webgpu (but not by default)

sagudev:
If PR has webgpu label, webgpu subset of wpt test will automatically be run on try/merge.

---

- [x] These changes fix #30144

I manually tested `./mach try (wpt|webgpu|_)`

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -8,6 +8,8 @@ on:
       wpt-sync:
         required: false
         type: boolean
+      # tests to be run
+      tests:
         required: false
         type: string
       layout:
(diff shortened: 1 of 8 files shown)


sagudev:
Last commit is actually fixup for https://github.com/servo/servo/pull/30509 as `./mach try wpt` was failing (only main/try workflow was working)

sagudev:
This was mostly done in https://github.com/servo/servo/pull/30912, webgpu try target is coming in https://github.com/servo/servo/pull/30920.

PR #30630, opened 2023-10-27
Update rust version to "nightly-2023-03-18"

mrobinson:
This is necessary to update the network stack.

---


mukilan (approved):

mrobinson:
This apparently needs a new version of `objc2` which means a new version of `winit`, which means a new version of `egui-winit` which is not released yet. :/

mrobinson:
Okay. The compilation failure in `objc2` was apparently due to a bug introduced in the nightly of 2023-03-20, so I've modified this change to use 2023-03-18 which is still in the 1.70 series.

sagudev:
This is: https://github.com/servo/servo/pull/30658#issuecomment-1787849458; I'm working on it.

Issue #30636, opened 2023-10-27
Investigate other 2D canvas renderers in a post-Pathfinder world


EnnuiL:
Currently, Servo uses [raqote](https://crates.io/crates/raqote) as the backend for 2D canvas rendering. This was supposed to be a temporary measure, a light upgrade over the past `rust_azure` status quo, until a true upgrade to Pathfinder could be done. Unfortunately, however, Pathfinder is effectively a dead and unmaintained project, and we still use raqote as a backend, which may be concerning due to reasons mentioned later;

There are two possible futures to worry about:
### Software rendering

We currently use raqote, however, it might not be the best library for our purposes; From what I've observed, [tiny_skia](https://crates.io/crates/tiny-skia) has quickly risen up as a main 2D software renderer, with it replacing raqote on most libraries (usage on Iced, resvg and sctk-adwaita being notable examples) while claiming a massive performance boost over it;
This has been the route I've been experimenting with while attempting to tackle #30535, and my prototype has shown moderate success so far, with few rough edges and few missing features (which hopefully should be trivial to implement) while handling a late-game Cookie Clicker save relatively better than raqote

This might be a reasonable short-term solution, however, despite being relatively better-maintained and enjoying a higher adoption, tiny_skia does consider itself a "best solution until someone makes an actually good 2D software renderer" sort of thing, and if we do consider migrating to it, we should be aware that we will likely need to switch backends once more;

An important feature absence, however, is of text rendering, which would mean something would have to be done on our side; This would allow to improve significantly over raqote's system (which should be considered as completely broken), but at the cost of dealing with text rendering, and all of it is such a nightmare that a revamp of its canvas implementation warrants its own issue 

### Hardware rendering

Pathfinder is dead, which is really disappointing, but considering recent huge advancements like the rise of `wgpu-rs` for GPU-accelerated rendering purposes? It might have been for the best to let fresher solutions pop up. Well, at least if those popped up in the first place;

One project, however, did seem highly promising as soon as I glanced on it: Linebender's [Vello](https://github.com/linebender/vello), which aims to do what Pathfinder pretty much did, and it might be something to experiment with; I might not be a rendering girl, but I like hooking up stuff together, so I might experiment with a Vello backend later

[...]

PR #30637, opened 2023-10-27
Clean up a bit of the canvas backend abstractions

EnnuiL:
This cleans up a bit of the Azure that was left behind after its removal. While I would have done a more drastic clean-up in order to reflect a scenario where there is a raqote and a tiny_skia backend, I have kept in mind possible alternatives routes (see #30636), so I've went for a lighter approach that may be intensified once progress on that area is made

---


jdm (approved):
Thanks! I am watching your canvas explorations with interest!

PR #30639, opened 2023-10-28
script: Start replacing `time` with `std::time` and `chrono` 

augustebaum:
This change is intended to be reviewed commit-by-commit.

- [x] These changes contribute to #30150

### DIFF
--- a/components/script/animation_timeline.rs
+++ b/components/script/animation_timeline.rs
@@ -7,8 +7,8 @@
 //! A timeline module, used to specify an `AnimationTimeline` which determines
 //! the time used for synchronizing animations in the script thread.
 
+use chrono::Utc;
 use jstraceable_derive::JSTraceable;
-use time;
 
 /// A `AnimationTimeline` which is used to synchronize animations during the script
 /// event loop.
@@ -22,7 +22,7 @@ impl AnimationTimeline {
     #[inline]
     pub fn new() -> Self {
         Self {
-            current_value: time::precise_time_s(),
+            current_value: Utc::now().timestamp() as f64,
         }
     }
 
@@ -39,7 +39,7 @@ impl AnimationTimeline {
 
     /// Updates the value of the `AnimationTimeline` to the current clock time.
     pub fn update(&mut self) {
-        self.current_value = time::precise_time_s();
+        self.current_value = Utc::now().timestamp() as f64;
     }
 
     /// Increments the current value of the timeline by a specific number of seconds.
--- a/components/script/animation_timeline.rs
+++ b/components/script/animation_timeline.rs
@@ -7,7 +7,8 @@
 //! A timeline module, used to specify an `AnimationTimeline` which determines
 //! the time used for synchronizing animations in the script thread.
 
-use chrono::Utc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use jstraceable_derive::JSTraceable;
 
 /// A `AnimationTimeline` which is used to synchronize animations during the script
@@ -22,7 +23,10 @@ impl AnimationTimeline {
     #[inline]
     pub fn new() -> Self {
         Self {
-            current_value: Utc::now().timestamp() as f64,
+            current_value: SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_secs_f64(),
         }
     }
 
@@ -39,7 +43,10 @@ impl AnimationTimeline {
 
     /// Updates the value of the `AnimationTimeline` to the current clock time.
     pub fn update(&mut self) {
-        self.current_value = Utc::now().timestamp() as f64;
+        self.current_value = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs_f64();
     }
 
     /// Increments the current value of the timeline by a specific number of seconds.
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -11,9 +11,9 @@ use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread::JoinHandle;
+use std::time::Instant;
 use std::{mem, ptr};
 
-use chrono::Utc;
 use content_security_policy::CspList;
 use crossbeam_channel::Sender;
 use devtools_traits::{PageError, ScriptToDevtoolsControlMsg};
@@ -204,7 +204,7 @@ pub struct GlobalScope {
     devtools_wants_updates: Cell<bool>,
 
     /// Timers (milliseconds) used by the Console API.
-    console_timers: DomRefCell<HashMap<DOMString, u64>>,
+    console_timers: DomRefCell<HashMap<DOMString, Instant>>,
 
     /// module map is used when importing JavaScript modules
     /// https://html.spec.whatwg.org/multipage/#concept-settings-object-module-map
@@ -2262,7 +2262,7 @@ impl GlobalScope {
         }
         match timers.entry(label) {
             Entry::Vacant(entry) => {
-                entry.insert(Utc::now().timestamp_millis() as u64);
+                entry.insert(Instant::now());
                 Ok(())
             },
             Entry::Occupied(_) => Err(()),
@@ -2274,7 +2274,7 @@ impl GlobalScope {
             .borrow_mut()
             .remove(label)
             .ok_or(())
-            .map(|start| Utc::now().timestamp_millis() as u64 - start)
+            .map(|start| (Instant::now() - start).as_millis() as u64)
     }
 
     /// Get an `&IpcSender<ScriptToDevtoolsControlMsg>` to send messages
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -11,9 +11,9 @@ use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread::JoinHandle;
-use std::time::SystemTime;
 use std::{mem, ptr};
 
+use chrono::Utc;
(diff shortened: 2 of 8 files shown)


components/script/dom/globalscope.rs
```diff
@@ -731,6 +731,13 @@ impl FileListener {
     }
 }
 
+fn now() -> u64 {
```
mrobinson, line 734: I think it would be useful to indicate the units in the function name so something like `get_timestamp_in_ms`.

components/script/dom/htmlmediaelement.rs
```diff
@@ -301,6 +301,10 @@ impl From<MediaStreamOrBlob> for SrcObject {
     }
 }
 
+fn now() -> SystemTime {
+    SystemTime::now()
+}
+
```
mrobinson, line 307: I think in this case you should omit the helper function entirely as it just calls a single function. It's more useful for the reader to know what this does.

components/script/script_runtime.rs
```diff
@@ -761,22 +760,26 @@ unsafe extern "C" fn gc_slice_callback(
...
         GCProgress::GC_SLICE_BEGIN => GC_SLICE_START.with(|start| {
-            start.set(Some(now()));
+            start.set(Some(SystemTime::now()));
             println!("GC slice began");
         }),
         GCProgress::GC_SLICE_END => GC_SLICE_START.with(|start| {
-            let dur = now() - start.get().unwrap();
+            let dur = SystemTime::now()
+                .duration_since(start.get().unwrap())
+                .unwrap();
```
mrobinson, line 773: I think we should try to not panic here, so please provide defaults if we cannot unwrap successfully.

components/script/script_runtime.rs
```diff
@@ -761,22 +760,26 @@ unsafe extern "C" fn gc_slice_callback(
...
             start.set(None);
-            println!("GC slice ended: duration={}", dur);
+            println!("GC slice ended: duration={:?}", dur);
         }),
         GCProgress::GC_CYCLE_END => GC_CYCLE_START.with(|start| {
-            let dur = now() - start.get().unwrap();
+            let dur = SystemTime::now()
+                .duration_since(start.get().unwrap())
+                .unwrap();
```
mrobinson, line 780: The same comment here.

components/script/script_thread.rs
```diff
@@ -214,9 +214,9 @@ struct InProgressLoad {
     /// The origin for the document
     #[no_trace]
     origin: MutableOrigin,
-    /// Timestamp reporting the time when the browser started this load.
+    /// Timestamp reporting the time in milliseconds when the browser started this load.
```
mrobinson, line 217: Great comment update!

components/script/script_thread.rs
```diff
@@ -258,7 +261,7 @@ impl InProgressLoad {
             is_visible: true,
             url: url,
             origin: origin,
-            navigation_start: (current_time.sec * 1000 + current_time.nsec as i64 / 1000000) as u64,
+            navigation_start: navigation_start,
```
mrobinson, line 264:
```suggestion
            navigation_start,
```

components/script/script_thread.rs
```diff
@@ -4059,11 +4068,8 @@ impl Drop for ScriptThread {
...
-    let tm = Timespec::new(tm.as_secs() as i64, 0);
-    let tm = at_utc(tm);
-    tm.to_local()
-        .strftime("%m/%d/%Y %H:%M:%S")
-        .unwrap()
-        .to_string()
+    let tm: Duration = tm.duration_since(SystemTime::UNIX_EPOCH).unwrap();
+    let dt = NaiveDateTime::from_timestamp_opt(tm.as_secs() as i64, tm.subsec_nanos()).unwrap();
+    let dt: DateTime<Local> = Local.from_utc_datetime(&dt);
+    dt.format("%m/%d/%Y %H:%M:%S").to_string()
```
mrobinson, line 4074: Maybe we can improve the name of these variables while we are here. Maybe this?

- `tm` –> `current_time`
- `dt` –> `utc_time`
- `dt` –> `local_time`

Please also properly handle when the unwrap would fail here.

mrobinson (changes requested):
Nicely done! I have a few comments below, but generally speaking this looks great.

components/script/script_thread.rs
augustebaum: Yeah I've been putting `unwrap`s all over the place because I have no visibility on the overall functionality. Would you have any idea how to deal with this specific one?

augustebaum:
I'm nearly done --- only have `script/dom/htmlformelement.rs` left, but I'm getting a gnarly compilation error. See [this chat entry](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/.2E.2E.2E.20Another.20compilation.20error.20involving.20.60dom_struct.60/near/399136903). Would appreciate any support!

components/script/script_runtime.rs
augustebaum: Changed the last `unwrap` to `unwrap_or_default`, for lack of a smarter solution.

components/script/script_runtime.rs
augustebaum: See [previous conversation](https://github.com/servo/servo/pull/30639#discussion_r1375412348).

components/script/script_thread.rs
augustebaum: I contemplated doing this but it seems this would be better dealt with in a big `fmt` change.

components/script/script_thread.rs
augustebaum: See [previous comment](https://github.com/servo/servo/pull/30639#discussion_r1375412348).

components/script/script_thread.rs
sagudev: We could enforce it with [use_field_init_shorthand](https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=use_field_init_shorthand#use_field_init_shorthand).

mrobinson (approved):

mrobinson:
Looks like this is a flake in the float tests...frustrating and they should probably be switched away from a fuzzing approach to a traditional test framework. I'm resending this to the MQ.

augustebaum:
@mrobinson Thanks for the review! Please consider resolving the above conversations as I'm interested in your opinion on these.

mrobinson:
It seems that some animation and performance tests are failing, likely because the new code is returning different results that the old code. One way to debug this is to compare the two times (via the old APIs and the new ones) via print statements to see where they are different.

Failures:
- https://github.com/servo/servo/runs/18185432560
- https://github.com/servo/servo/runs/18185642035

mrobinson (changes requested):

components/script/animation_timeline.rs
```diff
@@ -22,7 +23,10 @@ impl AnimationTimeline {
     #[inline]
     pub fn new() -> Self {
         Self {
-            current_value: time::precise_time_s(),
+            current_value: SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_secs() as f64,
```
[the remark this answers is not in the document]
augustebaum: Quite right, sorry I missed this.

components/script/animation_timeline.rs
```diff
@@ -39,7 +43,10 @@ impl AnimationTimeline {
 
     /// Updates the value of the `AnimationTimeline` to the current clock time.
     pub fn update(&mut self) {
-        self.current_value = time::precise_time_s();
+        self.current_value = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs() as f64;
```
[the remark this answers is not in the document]
augustebaum: ditto

augustebaum:
Some of these test failures might be linked to the fact that [`std::time::SystemTime` is not necessarily monotonic](https://doc.rust-lang.org/std/time/struct.SystemTime.html). Will need to investigate this...

sagudev:
Where possible use https://doc.rust-lang.org/std/time/struct.Instant.html

mrobinson:
The new set of results looks a lot better! I think this is getting quite close. I would also rebase this as at least one of these failures should be fixed in the latest from the main branch.

augustebaum:
I'm reworking the changes and there is something I find strange.

The `console_timers` attribute takes `u64` as values:
https://github.com/servo/servo/blob/f2270715861d577389fe6349a32ac483199d09f2/components/script/dom/globalscope.rs#L207

The first time they're used it's to put in a timestamp:
https://github.com/servo/servo/blob/f2270715861d577389fe6349a32ac483199d09f2/components/script/dom/globalscope.rs#L2272

whereas the second time it's to put in the elapsed time since the timestamp:
https://github.com/servo/servo/blob/f2270715861d577389fe6349a32ac483199d09f2/components/script/dom/globalscope.rs#L2284

Ideally `console_timers` would hold `Duration`s, but that doesn't fit with its first usage where `Instant` would be more fitting. I'll keep using `u64` but this inconsistency might bite someone down the line and should be dealt with.

augustebaum:
Kind of stuck here, in part because of https://github.com/servo/servo/issues/30714 but also just out of my depth. Is there any way to get support on this?

augustebaum:
Pinging just in case. Let me know if there is anything I can try.

mrobinson:
I think a good approach for debugging this is:

1. Confirm that at least one of the failing tests fails locally (maybe try to pick the simplest one).
2. Figure out what part of your change you need to revert to get it passing again.
3. Focus on that small part of your change to see what might be going wrong.

mrobinson:
Looks like there is at least one more issue to address.

mrobinson:
I had to make some changes to get this working namely, fixing some precision issues in the animation timeline as well as not making the changes for `navigation_start_precise`. I think that in the former case, we will need to make these changes across the entire source code all at once.

mrobinson (approved):

PR #30642, opened 2023-10-29
Fix the try build

mrobinson:
This was broken by the recent changes which added a production build.

---

- [x] These changes do not require tests because they just fix a CI issue.


sagudev (approved):
I noticed and fixed the same thing in https://github.com/servo/servo/pull/30627 although I resorted to ` == 'true'` instead of ` || false` because it was more explicit.

mrobinson:
Thanks of the review.

PR #30645, opened 2023-10-30
mach: use `importlib` module instead of `imp`

mukilan:
`imp` module has been deprecated since python 3.4
and has been removed in 3.12. The recommended alternative
is to use the `importlib` module that was introduced in
python 3.1

This is required to fix the CI failures in macos builds
since GitHub runner images for macos-13 now use python 3.12

### DIFF
--- a/python/mach/mach/main.py
+++ b/python/mach/mach/main.py
@@ -10,7 +10,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import argparse
 import codecs
 import errno
-import imp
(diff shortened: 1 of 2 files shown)


sagudev (approved):

mukilan:
Thanks for the review! Looks like there is one more use of `imp` module in `python/mach/mach/main.py`. I'll update this PR with the fix for that as well.

mrobinson (approved):

PR #30658, opened 2023-10-31
Replace environmental cargo_profile___ with workflow expressions

sagudev:
Instead of using those variables from os env context, where we need to be careful about which shell it's used or if it's even used it's better to use workflow expressions (`${{ expr }}`) to facilitate workflow preprocesor before anything is run.

---

- [x] These changes should fix mac pckg upload https://github.com/servo/servo/actions/runs/6702300916/job/18211086369#step:15:19
- [x] These changes do not require tests because they just fix a CI issue.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -71,7 +71,7 @@ jobs:
         if: ${{ inputs.wpt != 'sync' }}
         run: |
           python3 ./mach test-wpt $WPT_COMMAND_LINE_ARG \
-            ${cargo_profile_option} --processes $(nproc) --timeout-multiplier 2 \
+            ${{ env.cargo_profile_option }} --processes $(nproc) --timeout-multiplier 2 \
             --total-chunks ${{ env.max_chunk_id }} --this-chunk ${{ matrix.chunk_id }} \
             --log-raw test-wpt.${{ matrix.chunk_id }}.log \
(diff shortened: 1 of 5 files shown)


sagudev:
Try build: https://github.com/sagudev/servo/actions/runs/6703104912

delan (approved):
Thanks! I was wondering how to do this, thinking “surely there’s a better way to do this on windows than `($env: -split ' ')`…”

mrobinson (approved):

mrobinson:
It seems like we are going to have to make some changes in mozjs in order to support the new version of Python on the Mac builders.

sagudev:
I will port this to mozjs: https://bugzilla.mozilla.org/show_bug.cgi?id=1857492

sagudev:
Rebased.

mukilan:
Seems like we are seeing intermittent unit test failures in Windows:

Original merge queue [attempt](https://github.com/servo/servo/actions/runs/6727080709/job/18284370397):
```
test resource_thread::test_exit ... ok
thread panicked while panicking. aborting.
error: test failed, to rerun pass `-p net --test main`

Caused by:
  process didn't exit successfully: `C:\\a\\servo\\servo\\target\release\deps\main-235ff7879513c48a.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

The second [manual retry attempt](https://github.com/servo/servo/actions/runs/6727080709/job/18287222685) triggered by me passed though.

PR #30669, opened 2023-11-01
Update mozjs (includes fix for py3.12)

sagudev:
Includes https://github.com/servo/mozjs/pull/426

Fixes mac failures.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes do not require tests because they just fix the build.

PR #30670, opened 2023-11-01
Use py3.11 on mac runners

sagudev:
Temp fix because #30669 is not working fully yet.

---

- [x] These changes do not require tests because it hard code python version for CI


mrobinson (approved):
I guess we need to hit this with the big hammer.

PR #30682, opened 2023-11-02
Unit tests are becoming more flaky, so retry them

sagudev:
Especially net test on mac and now also on windows, but I've also seen flaky unit tests on linux.

Hopefully this would make mq faster as there should be less faulty fails.

Temporary solution for https://github.com/servo/servo/issues/30683

---

- [x] These changes do not require tests because there are CI resilient changes

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -102,6 +102,7 @@ jobs:
         if: ${{ inputs.unit-tests || github.ref_name == 'try-linux' }}
         uses: nick-fields/retry@v2
         with:
+          timeout_minutes: 20
           max_attempts: 2 # https://github.com/servo/servo/issues/30683
           command: python ./mach test-unit ${{ env.cargo_profile_option }}
       - name: Rename build timing
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -100,7 +100,10 @@ jobs:
         run: ./mach test-scripts
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-linux' }}
-        run: python3 ./mach test-unit ${{ env.cargo_profile_option }}
+        uses: nick-fields/retry@v2
+        with:
+          max_attempts: 2 # https://github.com/servo/servo/issues/30683
+          command: python ./mach test-unit ${{ env.cargo_profile_option }}
       - name: Rename build timing
         run: cp -r target/cargo-timings target/cargo-timings-linux
       - name: Archive build timing
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -93,8 +93,11 @@ jobs:
         run: ./mach test-scripts
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-mac' }}
-        timeout-minutes: 30 # https://github.com/servo/servo/issues/30275
-        run: python3 ./mach test-unit ${{ env.cargo_profile_option }}
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 20 # https://github.com/servo/servo/issues/30275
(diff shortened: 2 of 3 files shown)


.github/workflows/mac.yml
```diff
@@ -93,8 +93,11 @@ jobs:
         run: ./mach test-scripts
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-mac' }}
-        timeout-minutes: 30 # https://github.com/servo/servo/issues/30275
-        run: python3 ./mach test-unit ${cargo_profile_option}
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 20 # https://github.com/servo/servo/issues/30275
```
mrobinson, line 98: Very nice.

.github/workflows/linux.yml
```diff
@@ -100,7 +100,10 @@ jobs:
         run: ./mach test-scripts
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-linux' }}
-        run: python3 ./mach test-unit ${cargo_profile_option}
+        uses: nick-fields/retry@v2
+        with:
+          max_attempts: 2
+          command: python ./mach test-unit ${cargo_profile_option}
```
mrobinson, line 106: I can you put a link to https://github.com/servo/servo/issues/30683 here?

.github/workflows/mac.yml
```diff
@@ -93,8 +93,11 @@ jobs:
         run: ./mach test-scripts
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-mac' }}
-        timeout-minutes: 30 # https://github.com/servo/servo/issues/30275
-        run: python3 ./mach test-unit ${cargo_profile_option}
+        uses: nick-fields/retry@v2
```
mrobinson, line 96: Ditto.

.github/workflows/windows.yml
```diff
@@ -85,7 +85,10 @@ jobs:
         run: python mach smoketest --angle ($env:cargo_profile_option -split ' ')
       - name: Unit tests
         if: ${{ inputs.unit-tests || github.ref_name == 'try-windows' }}
-        run: python mach test-unit ($env:cargo_profile_option -split ' ')
+        uses: nick-fields/retry@v2
```
mrobinson, line 88: Please add the link here as well.

mrobinson (approved):
This needs a rebase, but otherwise this looks good. Let's give this a try!

mrobinson (approved):

sagudev:
Windows crashed three times, this might depend on which runner we are getting (maybe another runner image update).

mrobinson:
> Windows crashed three times, this might depend on which runner we are getting (maybe another runner image update).

This is so strange. I wonder if the runner is in an unusual state when it starts.

sagudev:
In second and third try unit test are not recompiled so same binaries are run. Maybe we are hitting some nondeterministic compilation problem.

mukilan:
`STATUS_STACK_BUFFER_OVERRUN` [seems to be related](https://stackoverflow.com/questions/76871022/github-build-server-fails-with-status-stack-buffer-overrun) to high memory usage. Out tests don't print 'memory allocation failed' though, so it might be [unrelated](https://devblogs.microsoft.com/oldnewthing/20190108-00/?p=100655)

Is it worth trying to run windows tests sequentially with `--test-thread=1` as suggested?

sagudev:
I am try running `--test-thread=1` in my fork.

mrobinson:
Looking at the try runs on your fork @sagudev, this looks like it's working well. I'm going to try to send this to the merge queue.

sagudev:
Picture of 5 successful Windows CI runs

OFF-TOPIC: Run times are so random.

PR #30687, opened 2023-11-03
Use runner-provided Python on Linux (currently version 3.10)

sagudev:
---

- [x] These changes do not require tests because it's CI


mrobinson (approved):

sagudev:
IIRC runner images uses python version provided by official ubuntu package, so there should not be any python changes until we change image (that's also why we needed py3.9 on Ubuntu 20.04 runners).

sagudev:
This needs to land after #30688

PR #30694, opened 2023-11-05
Matrix in main workflow & try string v3

### LINKED ISSUE #30667: Use fail fast on mq for faster queue pipeline.
I think [fail fast](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) can only be used within matrix, so this would require additional boogie woogie.

sagudev:
Rewrite of CI logic to allow more powerful try builds, lays groundwork for some future extensions

This PR removes some calling endpoints (try branches) for basic workflows (like linux and mac) and instead provides unified entry point for all try runs (only `try` branch remains operative). This is achieved using empty commits on try branch that have configuration encoded within. Same try string parser is shared between try comments and `mach try` invocations. Labels are still operational and are calling presets (that are also available via try strings).

Another important addition is usage of matrix which allow as to limit number of concurrent runs and fails-fast option.

---

- [x] These changes fix #30667
- [x] These changes do have test in try_parser_test.py

### DIFF
--- a/etc/try_parser.py
+++ b/etc/try_parser.py
@@ -1,0 +1,317 @@
+#!/usr/bin/env python
+
+# Copyright 2023 The Servo Project Developers. See the COPYRIGHT
+# file at the top-level directory of this distribution.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import json
+
+from enum import Enum
+
+
+# lexer
+class TokenType(Enum):
+    Illegal = 0
+    Eof = 1
+    #
+    Comma = 2
+    SemiColon = 3
+    Assign = 4
+    # ()
+    Lparen = 5
+    Rparen = 6
+    #
+    String = 7
+    # Keyword
+    FailFast = 8
+    Full = 9
+    Try = 10  # nop token
+
+    def __str__(self) -> str:
+        return self.name
+
+    def __repr__(self) -> str:
+        return self.name
+
+
+class Token:
+    m_type: TokenType
+    m_lit: str
+
+    def __init__(self, typ: TokenType, lit: str):
+        self.m_type: TokenType = typ
+        self.m_lit: str = lit
+
+    def is_seperator(self) -> bool:
+        return self.m_type in [TokenType.Comma, TokenType.SemiColon]
+
+    def __str__(self) -> str:
+        return str(self.m_type) + ": '" + self.m_lit + "'"
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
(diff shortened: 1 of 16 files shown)


sagudev:
@mrobinson is there any reason why main is used on PRs instead of quick-check workflow?

.github/workflows/quick-check.yml
```diff
@@ -6,6 +6,9 @@ on:
   push:
     branches-ignore:
       ["master", "auto", "try", "try-*", "dependabot/**"]
+  pull_request:
+    types: ['opened', 'synchronize']
+    branches: ["**"]
```
mrobinson, line 11: I believe that for the merge queue to work properly the pull_request checks need to be the same as the ones that run for the merge queue (ie run all tests before merging). That's why the pull_request is now part of the main workflow.

sagudev: Having last step in separate workflow with same name seems to work in my test repo. Also results jobs from non-main workflows are useless as they were only used by bors to determine success or failure.

sagudev: But quick-check doesn't get run here so I reverted this and removing redundant result jobs should be followup PR.

sagudev:
Most stuff is done so it might be good time for rough review @mrobinson.

I would also like to unify production input with more general profile input that would also enable debug builds. For this it would be nice if mach could have `--production` that would expand to `--profile production`.

etc/try_parser.py
```diff
@@ -0,0 +1,324 @@
+#!/usr/bin/env python
```
mrobinson, line 1: I'm concerned about this huge custom parser for try strings, which is one reason I haven't really had a lot of time to review this change. Maybe we can chat live about ways of simplifying this change in order to make it more maintainable.

sagudev: One option would be to write parser with ply that we have in third_party for webidl parser, but I wanted standalone python script so it could also be run on edge services (ex. cloudflare workers).

Separate lexer could be merged in parser given that it's pretty simple syntax.

mrobinson:
Apologies again for taking so long to review this one. Having to check out all of Servo in order to do the try decision means that there will be more delay before triggering try jobs. I wonder if we can simply parse the last commit message in `test-wpt` or add a new mach command that runs tests based on the commit message. Perhaps we can discuss this in Zulip.

In the meantime, I've pulled out some of this change to fix pushing to try at #30903.

sagudev:
> Apologies again for taking so long to review this one. Having to check out all of Servo in order to do the try decision means that there will be more delay before triggering try jobs. I wonder if we can simply parse the last commit message in `test-wpt` or add a new mach command that runs tests based on the commit message. Perhaps we can discuss this in Zulip.
> 
> In the meantime, I've pulled out some of this change to fix pushing to try at #30903.

Actually try workflow is only used when dispatching from labels and and on workflow_dispatch. It only parses try string or label. Instead of cloning whole repo, we can replace this with only downloading try_parser file.

sagudev:
New rebased version: https://github.com/servo/servo/pull/30920

PR #30716, opened 2023-11-10
Retry smoketest on mac in CI

sagudev:
Smoketest also fails frequently. Same mitigation as in https://github.com/servo/servo/pull/30682

---

- [x] These changes do not require tests because it's CI

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -83,7 +83,11 @@ jobs:
         run: |
           python3 ./mach build --${{ inputs.profile }}
       - name: Smoketest
-        run: python3 ./mach smoketest --${{ inputs.profile }}
+        uses: nick-fields/retry@v2
+        with: # See https://github.com/servo/servo/issues/30757
+          timeout_minutes: 5
+          max_attempts: 2
+          command: python3 ./mach smoketest --${{ inputs.profile }}
       - name: Script tests
         run: ./mach test-scripts
       - name: Unit tests


mrobinson:
This seems okay, though I worry that it might hide flaky failures in the smoketest (the most basic execution test). That said, it's difficult to detect when things fail due to infrastructure issues and when they fail due to issues with Servo. Do we normally see these flaky failures in the smoketest on Linux dealing with xvfb?

sagudev:
Smoketest on mac fails due to mutex being poisoned (when exiting) IIRC. I do not recall what failure we get on linux but they do happen occasionally. Windows is in that view the most stable.

Maybe setting retry to 2 would allow as to catch real flakes?

mrobinson:
> Smoketest on mac fails due to mutex being poisoned (when exiting) IIRC. 

Hrm. Is this a real bug that is triggered when we exit?

> Windows is in that view the most stable.

Let's not turn this for Windows then.

sagudev:
Hm, I am pretty sure there used to be different error but this one has been observed for last few mac fails on smoketest: https://github.com/servo/servo/actions/runs/6919006059/job/18821866378#step:12:41

.github/workflows/windows.yml
```diff
@@ -82,7 +82,11 @@ jobs:
       - name: Copy resources
         run: cp D:\a\servo\servo\resources C:\a\servo\servo -Recurse
       - name: Smoketest
-        run: python mach smoketest --angle --${{ inputs.profile }}
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: python mach smoketest --angle --${{ inputs.profile }}
```
mrobinson, line 89: Let's not retry on Windows for now.

.github/workflows/mac.yml
```diff
@@ -83,7 +83,11 @@ jobs:
         run: |
           python3 ./mach build --${{ inputs.profile }}
       - name: Smoketest
-        run: python3 ./mach smoketest --${{ inputs.profile }}
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 5
+          max_attempts: 3
+          command: python3 ./mach smoketest --${{ inputs.profile }}
```
mrobinson, line 90: Can you please open a bug for this failure and the one for Linux and leave the bug URL in the comments?

mrobinson (approved):
Okay. Let's give this a shot.

sagudev:
I only add retry functionality for mac as it happens really frequently, linux is harder to reproduce, so I would like to wait for at least one more failed run (again) so we can put it in issue.

mrobinson (approved):
Thank you!

PR #30740, opened 2023-11-16
Stop sending " " to linebreaker for replaced content

### LINKED ISSUE #30703: Panic loading some wikipedia content
Tested on commit 1db357456a . I'm sure this used to work but did not get a chance to bisect.

Steps to reproduce:
- Launch `./servo https://en.wikipedia.org/wiki/Freediving`

Expected:
- A nice wikipedia article.

Observed:
- A very unexpected panic:
```
index out of bounds: the len is 2 but the index is 2 (thread Style#0, at /home/capyloon/.cargo/registry/src/index.crates.io-6f17d22bba15001f/xi-unicode-0.1.0/src/lib.rs:43)
[2023-11-06T22:18:47Z ERROR servo::main2] index out of bounds: the len is 2 but the index is 2
Layout thread disconnected: "SendError(..)" (thread Script(1,1), at components/script/dom/window.rs:1912)
[2023-11-06T22:18:47Z ERROR servo::main2] Layout thread disconnected: "SendError(..)"
called `Result::unwrap()` on an `Err` value: PoisonError { .. } (thread Layout(1,2), at components/layout_thread_2020/lib.rs:962)
[2023-11-06T22:18:47Z ERROR servo::main2] called `Result::unwrap()` on an `Err` value: PoisonError { .. }
[2023-11-06T22:18:47Z ERROR constellation::constellation] crash page crashed
condition was Some(DirtyDescendants) (thread Script(1,2), at components/script/dom/window.rs:2001)
[2023-11-06T22:18:47Z ERROR servo::main2] condition was Some(DirtyDescendants)
[2023-11-06T22:18:47Z ERROR constellation::constellation] crash page crashed

```

mrobinson:
We previously sent a " " to the linebreaker in order to ensure that the
next text had a soft wrap opportunity at the start. Calling `next(" ")`
without waiting until the returned index was 1, violated some
invariants of linebreaker ultimately causing a panic.

Instead of using the linebreaker for this, simply keep a flag in the
IFC layout state, which avoids the problem entirely. This also fixes
some issues with linebreaking before character classes that should
not offer soft-wrap opportunities.

Fixes #30703.

---

- [x] These changes fix #30703.

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -21,7 +21,7 @@ use style::values::generics::text::LineHeight;
 use style::values::specified::text::{TextAlignKeyword, TextDecorationLine};
 use style::Zero;
 use webrender_api::FontInstanceKey;
-use xi_unicode::LineBreakLeafIter;
+use xi_unicode::{linebreak_property, LineBreakLeafIter};
 
 use super::float::PlacementAmongFloats;
 use super::CollapsibleWithParentStartMargin;
@@ -43,6 +43,12 @@ use crate::style_ext::{
     ComputedValuesExt, Display, DisplayGeneratingBox, DisplayOutside, PaddingBorderMargin,
 };
 use crate::ContainingBlock;
+
+// These constants are the xi-unicode line breaking classes that are defined in
+// `table.rs`. Unfortunately, they are only identified by number.
+const XI_LINE_BREAKING_CLASS_GL: u8 = 12;
+const XI_LINE_BREAKING_CLASS_WJ: u8 = 30;
+const XI_LINE_BREAKING_CLASS_ZWJ: u8 = 40;
 
 #[derive(Debug, Serialize)]
 pub(crate) struct InlineFormattingContext {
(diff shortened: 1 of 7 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -364,6 +364,11 @@ struct InlineFormattingContextState<'a, 'b> {
     /// The line breaking state for this inline formatting context.
     linebreaker: Option<LineBreakLeafIter>,
 
+    /// Whether or not a soft wrap opportunity is queued. Soft wrap opportunities are
+    /// queued after replaced content and they are processed when the next text content
+    /// is encountered.
+    queued_soft_wrap_opportunity: bool,
```
mukilan, line 370: Should this be called `deferred` soft wrap opportunity to be symmetric with `defer_forced_line_break`?

mukilan (approved):

components/layout_2020/flow/inline.rs
mrobinson: Sure, I've renamed this to `have_deferred_soft_wrap_opportunity`.

mrobinson:
It turns out that the previous approach I took here had more behavioral changes than I expected. I've adapted this to take a more conservative tactic, which is to simply properly use the linebreaker instead of improperly.

mrobinson:
@mukilan Would you be able to give this one more review?

mukilan (approved):

mrobinson:
@mukilan I've updated this to fix some hidden bugs with linebreaks after atomics. This also opens up the possibility of fixing even more linebreaking + atomic issues. Do you mind giving this another review?

mukilan (approved):

PR #30743, opened 2023-11-16
Remove Servo `window.trap()` extension

mrobinson:
This Servo-specific extension is unused by any code in the repository
and removing it allows us to remove use of nightly only reliance on
rust intrinsics. This is a step toward supporting stable rust.

---

- [x] These changes do not require tests because this simply removes an unused Servo-only feature.

### DIFF
--- a/components/script/dom/webidls/Window.webidl
+++ b/components/script/dom/webidls/Window.webidl
@@ -141,8 +141,6 @@ partial interface Window {
   [Pref="dom.servo_helpers.enabled"]
   undefined gc();
   [Pref="dom.servo_helpers.enabled"]
-  undefined trap();
-  [Pref="dom.servo_helpers.enabled"]
   undefined js_backtrace();
 };
 
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -1103,11 +1103,6 @@ impl WindowMethods for Window {
     }
 
     #[allow(unsafe_code)]
-    fn Trap(&self) {
-        unsafe { ::std::intrinsics::breakpoint() }
-    }
-
-    #[allow(unsafe_code)]
     fn Js_backtrace(&self) {
         unsafe {
             capture_stack!(in(*self.get_cx()) let stack);
--- a/components/script/lib.rs
+++ b/components/script/lib.rs
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-#![feature(core_intrinsics)]
 #![feature(drain_filter)]
 #![feature(once_cell)]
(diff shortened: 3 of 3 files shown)


mukilan:
I think the `trap()` was being used for manual testing/debugging - [example](https://github.com/servo/servo/pull/30290#issuecomment-1706467131). I don't have personal experience in using trap() for debugging, but do we not foresee this being useful for such situations?

sagudev:
Maybe we should make it work when compiled on nightly, on stable it would be nop.
https://stackoverflow.com/questions/59542378/conditional-compilation-for-nightly-vs-stable-rust-or-compiler-version

mrobinson:
> I think the `trap()` was being used for manual testing/debugging - [example](https://github.com/servo/servo/pull/30290#issuecomment-1706467131). I don't have personal experience in using trap() for debugging, but do we not foresee this being useful for such situations?

This feature might be useful in some cases, but we would also like to move to a stable version of Rust. You'll need to modify Servo anyway to use a different version than the one we specify, so adding a line of code to trigger a breakpoint isn't much more work.


> Maybe we should make it work when compiled on nightly, on stable it would be nop.
https://stackoverflow.com/questions/59542378/conditional-compilation-for-nightly-vs-stable-rust-or-compiler-version

I did consider this, but we specify the version of Rust we use in the rust-toolchain file. If you change that, adding a line to trigger the breakpoint is only one more line of code. Thus, I'm not sure how useful it is to maintain this feature that's only used in some very particular debugging cases.

mukilan:
> You'll need to modify Servo anyway to use a different version than the one we specify, so adding a line of code to trigger a breakpoint isn't much more work.

With `window.trap()` we can trigger a breakpoint from JS code without having to recompile servo at all, which I think would be much more flexible for debugging. After all, there must be a reason why this was added!

That said, I'm not opposed to removing it for now. We can find a way to bring it back again if needed.

mukilan (approved):

mrobinson:
Thanks for the review!

> With `window.trap()` we can trigger a breakpoint from JS code without having to recompile servo at all, which I think would be much more flexible for debugging. After all, there must be a reason why this was added!

Just to clarify, eventually rust-toolchain.toml will likely specify a stable version of Rust. In that case, in order to set the breakpoint with `window.trap()` a recompilation would be necessary.

sagudev:
> Thanks for the review!
> 
> > With `window.trap()` we can trigger a breakpoint from JS code without having to recompile servo at all, which I think would be much more flexible for debugging. After all, there must be a reason why this was added!
> 
> Just to clarify, eventually rust-toolchain.toml will likely specify a stable version of Rust. In that case, in order to set the breakpoint with `window.trap()` a recompilation would be necessary.

Will this pin us to specific rust stable version or just enforce MSRV?

PR #30750, opened 2023-11-18
dependencies: Upgrade media / GStreamer / GLib

mrobinson:
- Upgrade the version of GStreamer for Windows
    
This upgrades the Windows build to use the most recent version of
GStreamer. This is necessary to upgrade our GStreamer dependency.
    
- Stop shipping GStreamer binaries on Linux
    
The binary bundle of GStreamer that we package is not used to compile --
only to run layout tests. It's too old for the APIs that we are using
(as evidenced by needed 1.18 for WebRTC) and nowadays Linux
distributions carry a new version so it's unecessary for our build
machines. No longer using this binary bundle will allow us to upgrade
our GStreamer dependency -- which now has stricter checks that we
are using at least version 1.18.
    
- Upgrade media to use newer versions of GStreamer / GLib dependencies

---


mrobinson:
This change now depends on ﻿﻿#30831.

mrobinson:
This is ready to review.

delan (approved):

mrobinson:
This depends on #30973 an #30972.

mrobinson:
I'm rolling those changes up into this one as it appears this is all necessary together to unblock the bots.

mrobinson:
Updates:
 - Adjusted the apt packages installed on the CI.
 - Updated test results (some errors replaced by timeouts)
 - Updated the list of bundled DLLs for Windows.

PR #30751, opened 2023-11-18
Parse canvas colors as <color>

Loirooriol:
This will allow supporting color-mix() and the new color functions like in the CSS 'color' property (once that functionality is enabled).

This is according to the HTML spec:
https://html.spec.whatwg.org/multipage/infrastructure.html#parsed-as-a-css-color-value

---

- [X] These changes do not require tests because the new color functions are currently disabled.


Loirooriol:
Doing this as part of #30752

PR #30752, opened 2023-11-18
Enable new color functions from CSS Color 4

Loirooriol:
I will need to do most of the work anyways during the style updates,
so by enabling this it will be easier to detect mistakes.

Also, canvas colors are now parsed as `<color>`, precisely to support
these new features. This is according to the HTML spec:
https://html.spec.whatwg.org/multipage/infrastructure.html#parsed-as-a-css-color-value

---


mrobinson (approved):
Nice!

mrobinson:
Instead of landing this as two changes, I think it makes sense to land it all together so that the earlier commits are contextualized by the newly passing tests.

Loirooriol:
OK I will merge with #30751, but I have realized it will become simpler if #30754 lands first.

Loirooriol:
I have unified into a single commit, with some simplifications thanks to #30754.

mrobinson (approved):

PR #30762, opened 2023-11-21
Update Servo::handle_events to take IntoIterator

wusyong:
Small change on servo API I would like to see.
The existing usage with Vec is still the same, and it can now even accept other types like array and Drain.

---

- [x] These changes do not require tests but servoshell build needs to be successful.


mrobinson (approved):

wusyong:
uh, I have no idea what's the reason for that build step failed. I also tried to run the `macos_package_smoketest.sh` but everything seems fine.

sagudev:
I forget we have two smoketest on mac. I will add retry to second one too.

Issue #30593, opened 2023-10-20
[meta] embedding


delan:
* [ ] Adapt [wry](https://crates.io/crates/wry) to Rust libsimpleservo as proof of concept
* [ ] Build a Rust wrapper around C libsimpleservo
* [ ] Migrate from Rust libsimpleservo to the new Rust wrapper

fabricedesre:
What's the benefit of going Rust -> C -> Rust ?

mrobinson:
> What's the benefit of going Rust -> C -> Rust ?

The main benefit of this approach is the ability to ship a shared object.

fabricedesre:
But wry is already pure Rust so it doesn't need that, right?

mrobinson:
> But wry is already pure Rust so it doesn't need that, right?

Since Servo takes so long to build, being able to ship a shared object version of Servo will make it easier for embedders to use it.

wusyong:
We are trying to update our entry method to use [raw window](https://github.com/tauri-apps/rfcs/pull/12) handle recently.
Wry itself doesn't handle other things but wrap around webview instance only. I just need to make sure `Servo` and other related types can work. I'll try the approach similar to `servoshell` first and then explore multiwindow next.

A shared object isn't that urgent IMHO. Defining a C API that is ergonomic enough is another challenge as well. I tried to deal with CEF's C API before, and the experience didn't go that well. What I think is better are shared objects for huge C++ dependencies like `mozangle` and `mozjs`. These are always the first blocker in fresh environment. Maybe we could just download and link to built shared libraries if users only want to compile servo itself. Another benefit is developers can dive into servo codebase and hack it directly. I found I often need to refactor some servo code too.

sagudev:
Here is interesting blog post about dylib: https://robert.kra.hn/posts/2022-09-09-speeding-up-incremental-rust-compilation-with-dylibs/
I think something like that would be the best way for mozjs. mozangle is a bit more complicated on windows as we need it's produced dlls (libEGL.dll & libGLESv2.dll).

delan:
@sagudev interesting! I didn’t realise it was possible to build a lib crate as a dylib unless we used the C ABI. I wonder if it will break monomorphisation or break if [type layout changes between runs](https://nullderef.com/blog/plugin-start/#_abi_unstability_its_much_worse_than_it_seems).

sagudev:
[...]

PR #30767, opened 2023-11-22
Add initial support for offscreen rendering

delan:
This patch makes it possible to direct the output of the compositor to an offscreen OpenGL framebuffer object, rather than the OpenGL backbuffer or an image in IpcSharedMemory or a png file.

We also make servoshell use this iff the minibrowser is enabled, and rework CompositeTarget to allow the embedder to specify the target and output file if any when initialising Servo.

CompositeTarget before:

- Window draws to the OpenGL backbuffer only
- WindowAndPng draws to an uncompressed image in shared memory only
    - the name was misleading; it’s neither WindowAnd nor Png
- PngFile draws to IOCompositor.output_file only (which must be Some) and exits

CompositeTarget after:

- Window draws to the OpenGL backbuffer only
- Fbo draws to an offscreen OpenGL framebuffer object (IOCompositor::output_framebuffer_id)
- SharedMemory draws to an uncompressed image in shared memory only
- PngFile(String) draws to a PNG file at the given path only and exits

Manually verified working on:
- [x] Windows
- [x] macOS
- [x] Linux (X11)
- [x] Linux (Wayland) — no worse than before
    - #30826
    - #30827

---

- [x] These changes are related to #30593


components/compositing/compositor.rs
```diff
@@ -1656,14 +1659,13 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         }
     }
 
-    /// Composite either to the screen or to a png image or both.
+    /// Composite to the given target if any, or the current target otherwise.
     /// Returns Ok if composition was performed or Err if it was not possible to composite
-    /// for some reason. If CompositeTarget is Window or Png no image data is returned;
-    /// in the latter case the image is written directly to a file. If CompositeTarget
-    /// is WindowAndPng Ok(Some(png::Image)) is returned.
+    /// for some reason. If read_back is true, reads back and returns the output frame as
```
wusyong, line 1664: Could this explain more about what `read_back` means? I saw `CompositeTarget::SharedMemory` will return image value. Maybe we say which CompositeTarget will return the value?

components/compositing/compositor.rs
```diff
@@ -339,33 +341,31 @@ impl PipelineDetails {
...
-enum CompositeTarget {
-    /// Normal composition to a window
+#[derive(Clone, Debug, PartialEq)]
+pub enum CompositeTarget {
+    /// Draw directly to a window.
     Window,
 
-    /// Compose as normal, but also return a PNG of the composed output
-    WindowAndPng,
+    /// Draw to an offscreen OpenGL framebuffer object (IOCompositor::output_framebuffer_id).
```
wusyong, line 349:
```suggestion
    /// Draw to an offscreen OpenGL framebuffer object [IOCompositor::output_framebuffer_id].
```
I believe `cargo doc` is able to create the link.

components/compositing/compositor.rs
delan: good catch, this became outdated in b422d3c9a07fcd43047ae06ad650f9d0951eb9ea

delan: done in 1f07d7bd76edf21bfb54842da7b96e14b6023e37

components/compositing/compositor.rs
delan: done in 1f07d7bd76edf21bfb54842da7b96e14b6023e37

wusyong:
`CompositeTarget::Window` is fine on macOS, but `CompositeTarget::Fbo` will show black screen.
I couldn't figure out the root cause yet. Maybe we have to assign the frame buffer to surfman's context as well.
Screenshot 2023-11-29 at 2 41 56 PM

wusyong:
After investigating a bit, I found there's already a swap buffer in [webrender_surfman](https://github.com/servo/servo/blob/master/components/webrender_surfman/lib.rs#L28C8-L28C8). But it doesn't support the surface that is attached to a window. And because of the design of surfman, we can't create a framebuffer ourselves and expect it to display on every platform. The frame buffer must be created from the `Surface` from the start.

So the good thing is offscreen rendering is already there, headless window is already using it. We can retrieve the frame buffer ID from its webrender_surfman.

The bad thing is this PR might not be needed, but some changes are great IMHO. I agree original variants of `CompositeTarget` are pretty misleading. Maybe we keep these three variants (Window, SharedMemory, PNGFile) in this PR.
And if we want window to also have swap buffer, we probably need to refactor both webrender_surfman and surfman's `SwapChain` to support surface type that is native widget.

ports/servoshell/minibrowser.rs
```diff
@@ -76,39 +117,91 @@ impl Minibrowser {
...
+                                    painter.gl().scissor(x, y, width, height);
+                                    painter.gl().enable(gl::SCISSOR_TEST);
+                                    painter.gl().clear(gl::COLOR_BUFFER_BIT);
+                                    painter.gl().disable(gl::SCISSOR_TEST);
+
+                                    let fbo = NativeFramebuffer(NonZeroU32::new(fbo).unwrap());
+                                    painter
+                                        .gl()
+                                        .bind_framebuffer(gl::READ_FRAMEBUFFER, Some(fbo));
+                                    painter.gl().bind_framebuffer(gl::DRAW_FRAMEBUFFER, None);
```
wusyong, line 185: Spending days trying to figure out why Fbo variant can't display on macOS. And it turns out CGL decided to give the default framebuffer ID with "1" :joy: 
So getting the framebuffer ID of the widget surface should fix it.

delan: fixed in 44ecfb6a48f931b2c087ac2e29698794fe4bbe58

delan:
> ⚠️ Try run ([#7112225827](https://github.com/servo/servo/actions/runs/7112225827)) failed.

#30828

components/compositing/gl.rs
```diff
@@ -2,125 +2,149 @@
...
+    pub fn bind(&self) {
+        trace!("Binding fbo {}", self.framebuffer_id());
+        self.gl
+            .bind_framebuffer(gl::FRAMEBUFFER, self.framebuffer_id());
     }
 
-    RgbImage::from_raw(width as u32, height as u32, pixels).expect("Flipping image failed!")
+    pub fn unbind(&self) {
+        trace!("Unbinding fbo {}", self.framebuffer_id());
+        self.gl.bind_framebuffer(gl::FRAMEBUFFER, 0);
```
wusyong, line 99: This should also bind back to the framebuffer object of webrender surface.

atbrakhi (approved):
Nice work. I tested locally and right now the `fn unbind` has the incorrect ID, which breaks the minibrowser. I passed the correct `framebuffer_id()` `(1)` locally and it works fine. I guess @wusyong's comment says the same. Other than that, everything looks nice to me.

image

components/compositing/gl.rs
delan: fixed in 59979f391cf58845b983c1bd2210dcf5c02b5714, though i decided to bind the widget surface buffer in Minibrowser.paint() rather in RenderTargetInfo.unbind()

components/compositing/compositor.rs
```diff
@@ -1781,77 +1810,82 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-                                    FramebufferUintLength::new(width),
-                                    FramebufferUintLength::new(height),
-                                );
-                                let dynamic_image = DynamicImage::ImageRgb8(img);
-                                if let Err(e) = dynamic_image.write_to(&mut file, ImageFormat::Png)
-                                {
-                                    error!("Failed to save {} ({}).", path, e);
-                                }
+        cfg_if! {
+            if #[cfg(feature = "gl")] {
```
mrobinson, line 1829: I honestly suspect that the `gl` feature might be a vestige of a time before WebRender.

components/compositing/gl.rs
```diff
@@ -2,125 +2,149 @@
...
-    // flip image vertically (texture is upside down)
-    let orig_pixels = pixels.clone();
-    let stride = width * 3;
-    for y in 0..height {
-        let dst_start = y * stride;
-        let src_start = (height - y - 1) * stride;
-        let src_slice = &orig_pixels[src_start..src_start + stride];
-        (&mut pixels[dst_start..dst_start + stride]).clone_from_slice(&src_slice[..stride]);
+    pub fn bind(&self) {
+        trace!("Binding fbo {}", self.framebuffer_id());
```
mrobinson, line 92: Sorry for this nit:

```suggestion
        trace!("Binding FBO {}", self.framebuffer_id());
```

components/compositing/gl.rs
```diff
@@ -2,125 +2,149 @@
...
+            .bind_framebuffer(gl::FRAMEBUFFER, self.framebuffer_id());
     }
 
-    RgbImage::from_raw(width as u32, height as u32, pixels).expect("Flipping image failed!")
+    pub fn unbind(&self) {
+        trace!("Unbinding fbo {}", self.framebuffer_id());
+        self.gl.bind_framebuffer(gl::FRAMEBUFFER, 0);
+    }
+
+    pub fn read(
```
mrobinson, line 102: I think a name like `read_to_image` describes a bit better what this does.

components/compositing/gl.rs
```diff
@@ -2,125 +2,149 @@
...
+            (&mut pixels[dst_start..dst_start + stride]).clone_from_slice(&src_slice[..stride]);
+        }
+
+        RgbImage::from_raw(width as u32, height as u32, pixels).expect("Flipping image failed!")
+    }
+}
+
+impl Drop for RenderTargetInfo {
+    fn drop(&mut self) {
+        trace!("Dropping fbo {}", self.framebuffer_id());
```
mrobinson, line 144:
```suggestion
        trace!("Dropping FBO {}", self.framebuffer_id());
```

ports/servoshell/minibrowser.rs
```diff
@@ -56,18 +64,62 @@ impl Minibrowser {
...
+        };
+        result
+    }
+
+    /// Return true iff the given position is in the Servo browser rect.
+    fn is_in_browser_rect(&self, position: Point2D<f32, DeviceIndependentPixel>) -> bool {
+        position.y < self.toolbar_height.get()
+    }
+
     /// Update the minibrowser, but don’t paint.
```
mrobinson, line 116: It would be good to have a bit of documentation for the optional framebuffer id. When is it passed and what does it do if it is passed?

ports/servoshell/minibrowser.rs
```diff
@@ -56,18 +64,62 @@ impl Minibrowser {
...
+                    .map_or(false, |p| self.is_in_browser_rect(p))
+            },
+            winit::event::WindowEvent::MouseWheel { .. } |
+            winit::event::WindowEvent::MouseInput { .. } => self
+                .last_mouse_position
+                .map_or(false, |p| self.is_in_browser_rect(p)),
+            _ => true,
+        };
+        result
+    }
```
mrobinson, line 109: Hrm. Out of curiosity, how does this relate to offscreen rendering?

ports/servoshell/minibrowser.rs
```diff
@@ -78,62 +130,127 @@ impl Minibrowser {
...
+                                    painter.gl().clear(gl::COLOR_BUFFER_BIT);
+                                    painter.gl().disable(gl::SCISSOR_TEST);
+
+                                    let servo_fbo =
+                                        NonZeroU32::new(servo_fbo).map(NativeFramebuffer);
+                                    let our_fbo = NonZeroU32::new(our_fbo).map(NativeFramebuffer);
+                                    painter
+                                        .gl()
+                                        .bind_framebuffer(gl::READ_FRAMEBUFFER, servo_fbo);
+                                    painter.gl().bind_framebuffer(gl::DRAW_FRAMEBUFFER, our_fbo);
```
mrobinson, line 220: So am I understanding correctly that this unconditionally adds one more framebuffer blit?

ports/servoshell/minibrowser.rs
```diff
@@ -78,62 +130,127 @@ impl Minibrowser {
...
+
+            CentralPanel::default()
+                .frame(Frame::none())
+                .show(ctx, |ui| {
+                    let min = ui.cursor().min;
+                    let size = ui.available_size();
+                    let rect = egui::Rect::from_min_size(min, size);
+                    ui.allocate_space(size);
+
+                    if let Some(servo_fbo) = servo_framebuffer_id {
```
mrobinson, line 197: It think an early return might help reduce some of the nesting here:

```rust
let servo_fbo = match servo_framebuffer_id {
    Some(fbo) => fbo,
    None => return,
};
```

ports/servoshell/minibrowser.rs
```diff
@@ -56,18 +64,62 @@ impl Minibrowser {
             .egui_ctx
             .set_pixels_per_point(window.hidpi_factor().get());
 
+        let widget_surface_fbo = match webrender_surfman.context_surface_info() {
+            Ok(Some(info)) => info.framebuffer_object,
+            Ok(None) => panic!("Failed to get widget surface info from surfman!"),
+            Err(error) => panic!(
+                "Failed to get widget surface info from surfman! {:?}",
+                error
+            ),
```
mrobinson, line 73:
```suggestion
                "Failed to get widget surface info from surfman! {error:?}",
            ),
```

ports/servoshell/minibrowser.rs
```diff
@@ -78,62 +130,127 @@ impl Minibrowser {
...
+                });
 
-            toolbar_height.set(Length::new(ctx.used_rect().height()));
             *last_update = now;
         });
     }
 
     /// Paint the minibrowser, as of the last update.
     pub fn paint(&mut self, window: &winit::window::Window) {
+        let our_fbo = NonZeroU32::new(self.widget_surface_fbo).map(NativeFramebuffer);
```
mrobinson, line 246: Just for clarity:

```suggestion
        let widget_fbo = NonZeroU32::new(self.widget_surface_fbo).map(NativeFramebuffer);
```

ports/servoshell/minibrowser.rs
```diff
@@ -18,14 +23,17 @@ use servo::webrender_surfman::WebrenderSurfman;
...
+use crate::geometry::winit_position_to_euclid_point;
 use crate::parser::location_bar_input_to_url;
 use crate::window_trait::WindowPortsMethods;
 
 pub struct Minibrowser {
     pub context: EguiGlow,
     pub event_queue: RefCell<Vec<MinibrowserEvent>>,
-    pub toolbar_height: Cell<Length<f32, DeviceIndependentPixel>>,
+    pub toolbar_height: Length<f32, DeviceIndependentPixel>,
+    widget_surface_fbo: gl::GLuint,
```
mrobinson, line 34: Why not store a `NativeFramebuffer` here to avoid having to `map(...)` it every time?

ports/servoshell/minibrowser.rs
```diff
@@ -78,62 +130,127 @@ impl Minibrowser {
             context,
             event_queue,
             toolbar_height,
+            widget_surface_fbo,
             last_update,
             location,
             location_dirty,
+            ..
         } = self;
+        let our_fbo = *widget_surface_fbo;
```
mrobinson, line 139:
```suggestion
        let widget_fbo = *widget_surface_fbo;
```

ports/servoshell/app.rs
```diff
@@ -122,8 +123,8 @@ impl App {
         }
 
         if let Some(mut minibrowser) = app.minibrowser() {
-            minibrowser.update(window.winit_window().unwrap(), "init");
-            window.set_toolbar_height(minibrowser.toolbar_height.get());
+            minibrowser.update(window.winit_window().unwrap(), None, "init");
```
mrobinson, line 126: Why is the framebuffer None here? Might be nice to have a little comment explaining why.

mrobinson (dismissed):
Nice work. I mainly have some questions and minor suggestions, but this looks great.

mrobinson:
@delan Sorry, one more super nit. Would you mind making the PR title a short sentence? It would be nice to have a standard form for commit messages like the one for Zulip: https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html.

components/compositing/compositor.rs
```diff
@@ -339,33 +349,31 @@ impl PipelineDetails {
...
+    /// Draw to a PNG file on disk, then exit the browser (for reftests).
+    PngFile(Rc<String>),
 }
 
 impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
     fn new(
         window: Rc<Window>,
         state: InitialCompositorState,
-        output_file: Option<String>,
+        composite_target: CompositeTarget,
```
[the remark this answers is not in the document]
delan: thanks ^w^

components/compositing/compositor.rs
```diff
@@ -1656,14 +1676,13 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+    /// Composite to the given target if any, or the current target otherwise.
+    /// Returns Ok if composition was performed or Err if it was not possible to composite for some
+    /// reason. When the target is [CompositeTarget::SharedMemory], the image is read back from the
+    /// GPU and returned as Ok(Some(png::Image)), otherwise we return Ok(None).
     fn composite_specific_target(
         &mut self,
-        target: CompositeTarget,
+        target_override: Option<CompositeTarget>,
         rect: Option<Rect<f32, CSSPixel>>,
     ) -> Result<Option<Image>, UnableToComposite> {
```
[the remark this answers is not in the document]
delan: done with `composite` itself in dca42a937f1302ece951c10e01638bc26113fcf6

components/compositing/compositor.rs
```diff
@@ -1713,25 +1726,41 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-            #[cfg(feature = "gl")]
-            CompositeTarget::WindowAndPng | CompositeTarget::PngFile => gl::initialize_png(
-                &*self.webrender_gl,
-                FramebufferUintLength::new(size.width),
-                FramebufferUintLength::new(size.height),
-            ),
-            #[cfg(not(feature = "gl"))]
-            _ => (),
+        cfg_if! {
+            if #[cfg(feature = "gl")] {
```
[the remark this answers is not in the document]
delan: removed in ae432a9284c75b81cd14e03fccc3164c2da64e8f

components/compositing/compositor.rs
delan: yeah, i think you’re right; removed in ae432a9284c75b81cd14e03fccc3164c2da64e8f

components/compositing/gl.rs
delan: done in 34a2a466d6f3a870b80e0c83e68a1dcfe9be158d

components/compositing/gl.rs
delan: done in dbca1a9484401fac346d5ce2ab62753a3ecbbaab

components/compositing/compositor.rs
```diff
@@ -1680,23 +1699,17 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-        };
+        let target = target_override.unwrap_or_else(|| self.composite_target.clone());
+        let wait_for_stable_image = matches!(
+            target,
+            CompositeTarget::SharedMemory | CompositeTarget::PngFile(_)
+        ) || self.exit_after_load;
+        let needs_fbo = matches!(
+            target,
+            CompositeTarget::SharedMemory | CompositeTarget::PngFile(_) | CompositeTarget::Fbo
+        );
```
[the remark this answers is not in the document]
delan: i wanted to put this next to `wait_for_stable_image` to make their similar but different conditions easier to see — notice how we wait for stable image if we are exiting after load, while we need an FBO when the composite target is in Fbo mode, but both are true if we are in SharedMemory or PngFile mode.

components/compositing/compositor.rs
```diff
@@ -1713,25 +1726,41 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+                        self.current_render_target = Some(RenderTargetInfo::new(
+                            self.webrender_gl.clone(),
+                            FramebufferUintLength::new(size.width),
+                            FramebufferUintLength::new(size.height),
+                        ));
+                    }
+                    self.current_render_target
+                        .as_ref()
+                        .expect("Initialised above")
+                        .bind();
```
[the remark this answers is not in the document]
delan: done in 67cc7cc71a66a0a31f78cae9435714c596c8446e

components/compositing/compositor.rs
```diff
@@ -225,7 +228,14 @@ pub struct IOCompositor<Window: WindowMethods + ?Sized> {
     /// Current cursor position.
     cursor_pos: DevicePoint,
 
-    output_file: Option<String>,
+    /// Offscreen framebuffer object to render to.
+    current_render_target: Option<gl::RenderTargetInfo>,
+
+    /// Offscreen framebuffer object that our last frame was rendered to.
+    last_render_target: Option<gl::RenderTargetInfo>,
```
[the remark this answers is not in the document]
delan: done in 67cc7cc71a66a0a31f78cae9435714c596c8446e

components/compositing/gl.rs
delan: renamed to `read_back_from_gpu` (dbac481ae9c71311621194382125b0a1c37ab820) to make it explicit that this is a readback

ports/servoshell/minibrowser.rs
delan: done in 2f45134c281815b2141619e605c33db18b37f92a

components/compositing/compositor.rs
```diff
@@ -1713,25 +1726,41 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
-            CompositeTarget::WindowAndPng | CompositeTarget::PngFile => gl::initialize_png(
-                &*self.webrender_gl,
-                FramebufferUintLength::new(size.width),
-                FramebufferUintLength::new(size.height),
-            ),
-            #[cfg(not(feature = "gl"))]
-            _ => (),
+        cfg_if! {
+            if #[cfg(feature = "gl")] {
+                if needs_fbo {
```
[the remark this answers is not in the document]
delan: done in f86a49cc51d580ece0316d685b868399ba9e13d3

ports/servoshell/minibrowser.rs
delan: we use offscreen rendering in servoshell by blitting the webview to an egui [CentralPanel](https://docs.rs/egui/0.22.0/egui/containers/panel/struct.CentralPanel.html) when the minibrowser is enabled. but adding the CentralPanel makes egui report events that land in the webview area as consumed. clarified in f299776edded5fc6abd5cd8b1347b92942d58a4d

ports/servoshell/minibrowser.rs
delan: yes, though only when servoshell is in a headed window with the minibrowser enabled.

ports/servoshell/minibrowser.rs
delan: done in a7d0b2e1f3dcfdfc68c222895b671c334f311c86

ports/servoshell/minibrowser.rs
delan: done in f96a28523112620bdee12fd2b88399e33f47ba0a, though i had to bump servoshell to rust 2021

ports/servoshell/minibrowser.rs
delan: done in f6f576b20cbdafa4765c66ac400894be392caa91

ports/servoshell/minibrowser.rs
delan: done in f6f576b20cbdafa4765c66ac400894be392caa91

ports/servoshell/minibrowser.rs
delan: done in f6f576b20cbdafa4765c66ac400894be392caa91

ports/servoshell/app.rs
delan: done in 30944bd054323bbcf42e106dcd9256ebef030542

components/compositing/compositor.rs
```diff
@@ -1871,6 +1905,12 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         Ok(rv)
     }
 
+    pub fn output_framebuffer_id(&self) -> Option<gleam::gl::GLuint> {
+        self.last_render_target
+            .as_ref()
+            .map(|info| info.framebuffer_id())
+    }
+
```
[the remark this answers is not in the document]
delan: done in c0a1998bd40fcffd27711eb7c068b4963a2ce16d

components/compositing/compositor.rs
```diff
@@ -1781,77 +1810,82 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+                let rv = match target {
+                    CompositeTarget::Window => None,
+                    CompositeTarget::Fbo => {
+                        self.current_render_target.as_ref().expect("Guaranteed by needs_fbo").unbind();
+                        if self.invalidate_last_render_target {
+                            // Do not reuse the last render target as the new current render target.
+                            self.last_render_target = None;
+                            self.invalidate_last_render_target = false;
+                        }
+                        swap(&mut self.current_render_target, &mut self.last_render_target);
```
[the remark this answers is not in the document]
delan: yeah, i would imagine so! can we land this as is, given it adapts a known-working approach that we’ve tested on all the desktop platforms, then migrate it to surfman chains separately?

delan:
> Would you mind making the PR title a short sentence?

No worries, done!

delan:
Landing since Martin is on leave, but happy to address any further issues if and when they come up.

components/compositing/compositor.rs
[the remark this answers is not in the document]
mrobinson: Yes, sure.

PR #30784, opened 2023-11-27
Fix typing issues in flex layout

holmes-software:
This is a simple PR to address incorrect type usage in the flex layout implementation. The change is correcting a misuse of `Result`. In this case, `Result` is being used in a way that is more appropriate for an `Option`.

In addition to that, there is a small amount of renaming and more verbose comments.

---

- [X] These changes do not require tests because changes are small corrections to existing code

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -158,14 +158,18 @@ impl FlexContainer {
         // Absolutely-positioned children of the flex container may be interleaved
         // with flex items. We need to preserve their relative order for correct painting order,
         // which is the order of `Fragment`s in this function’s return value.
-        let original_order_with_absolutely_positioned = self
+        //
+        // Example:
+        // absolutely_positioned_items_with_original_order = [Some(item), Some(item), None, Some(item), None]
+        // flex_items                                      =                         [item,             item]
+        let absolutely_positioned_items_with_original_order = self
             .children
             .iter()
             .map(|arcrefcell| {
                 let borrowed = arcrefcell.borrow_mut();
                 match &*borrowed {
                     FlexLevelBox::OutOfFlowAbsolutelyPositionedBox(absolutely_positioned) => {
-                        Ok(absolutely_positioned.clone())
+                        Some(absolutely_positioned.clone())
                     },
                     FlexLevelBox::FlexItem(_) => {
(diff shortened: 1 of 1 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -158,22 +158,26 @@ impl FlexContainer {
...
+                        Some(absolutely_positioned.clone())
                     },
                     FlexLevelBox::FlexItem(_) => {
                         let item = AtomicRefMut::map(borrowed, |child| match child {
                             FlexLevelBox::FlexItem(item) => item,
                             _ => unreachable!(),
                         });
                         flex_items.push(item);
-                        Err(())
+                        None
```
mrobinson, line 180: If we are cleaning up this code, I think it might better for this code to stop using `Result` and `Option` to track these variables and use an appropriately named enum in order to be as clear as possible. Perhaps something like this:

```rust
enum FlexContent {
    AbsolutelyPositionedBox(...)
    FlexItemPlaceholder
}
```

joshua-holmes: I love that idea

joshua-holmes: I played around with other ideas of holding all flex items in that enum you suggested, instead of just the absolutely positioned ones, just to see if I could take the simplification a step further. However, I found that to be more confusing. Having a vector of enums with data and placeholders, then another vector with items to fill the placeholders is the simplest way I can think of (this is the current implementation). I think having a dedicated enum for this, like we have it in this PR, will make the code a bit friendlier to read.

Ready for another review!

mrobinson (approved):

components/layout_2020/flexbox/layout.rs
```diff
@@ -158,39 +166,230 @@ impl FlexContainer {
...
+        let flex_item_boxes = flex_items.iter_mut().map(|child| &mut **child);
+
+        // FIXME: get actual min/max cross size for the flex container.
+        // We have access to style for the flex container in `containing_block.style`,
+        // but resolving percentages there requires access
+        // to the flex container’s own containing block which we don’t have.
+        // For now, use incorrect values instead of panicking:
+        let container_min_cross_size = Length::zero();
+        let container_max_cross_size = None;
+
```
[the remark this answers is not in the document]
joshua-holmes: I agree

PR #30796, opened 2023-11-29
Fix build script error when you using Python version > 3.9.

seasonmac:
Fix build script error when you using Python version > 3.9, because starting
with Python 3.9, the util module was moved from importlib to a separate
top-level module called importlib.util. Therefore, if you are using Python 3.9
or later and you receive an AttributeError stating that module 'importlib' has
no attribute 'util', it is likely due to a compatibility issue with your code. You
can fix this issue by updating your code to import importlib.util directly

---

- [x] These changes do not need tests because they fix a mach issue.

### DIFF
--- a/python/mach/mach/main.py
+++ b/python/mach/mach/main.py
@@ -10,7 +10,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import argparse
 import codecs
 import errno
-import importlib
+import importlib.util
 import logging
 import os
 import sys


sagudev:
Could you paste your whole error message, because I think I was getting same error.

seasonmac:
> Could you paste your whole error message, because I think I was getting same error.
seasonmac@MBP servo % ./mach build --dev
  File "/workspace/github/servo/./mach", line 46, in <module>
    main(sys.argv)
  File "/workspace/github/servo/./mach", line 27, in main
    mach = mach_bootstrap.bootstrap(topdir)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/github/servo/python/mach_bootstrap.py", line 269, in bootstrap
    mach.load_commands_from_file(os.path.join(topdir, path), module_name)
  File "/workspace/github/servo/python/mach/mach/main.py", line 262, in load_commands_from_file
    spec = importlib.util.spec_from_file_location(module_name, path)
           ^^^^^^^^^^^^^^
AttributeError: module 'importlib' has no attribute 'util'

python/mach/mach/main.py
```diff
@@ -11,6 +11,9 @@
 import codecs
 import errno
 import importlib
+from sys import version_info
+if version_info.major == 3 and version_info.minor > 9 :
+    import importlib.util
```
mrobinson, line 16: Based on https://discuss.python.org/t/python3-11-importlib-no-longer-exposes-util/25641/9 it seems that the correct thing to do here is to unconditionally import `importlib.util`. That should make this fix a bit simpler.

mrobinson (changes requested):
Thanks for this fix. I've seen this error too. I have one request before we can land it.

python/mach/mach/main.py
seasonmac: yes, you are right, just import  importlib.util can fix this issue,

seasonmac:
@mrobinson has updated.

mrobinson (approved):

PR #30803, opened 2023-11-30
Integrate hyper_serde into Servo source

mrobinson:
hyper_serde changes very infrequently and typically only in order to
upgrade Servo dependencies. crates.io lists hyper_serde as having no
dependents, and its integration will not prevent releasing new versions.
Integrating it will both reduce the number of repositories we have to
maintain as well as making it easier to upgrade Servo dependencies.

---

### DIFF
--- a/components/hyper_serde/Cargo.toml
+++ b/components/hyper_serde/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "hyper_serde"
+version = "0.13.2"
+edition = "2018"
+authors = ["The Servo Project Developers"]
+description = "Serde support for hyper types."
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/servo/servo"
+documentation = "https://docs.rs/hyper_serde"
+categories = ["encoding", "web-programming"]
+keywords = ["serde", "serialization", "hyper", "cookie", "mime"]
+
+[lib]
+path = "lib.rs"
+doctest = false
+test = false
+
+[dependencies]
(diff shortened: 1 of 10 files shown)


components/hyper_serde/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "hyper_serde"
+version = "0.13.2"
+edition = "2018"
+authors = ["Anthony Ramine <n.oxyde@gmail.com>"]
+description = "Serde support for hyper types."
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/servo/hyper_serde"
```
sagudev, line 8: This should probably change right?

mukilan:
I'm wondering this crate should be in `support/` instead of `components/`. Seems like `components/` should only contain the crates providing browser-engine specific functionality, but this doesn't seem to be followed strictly e.g crates like `range` and `rand` are also in `components/` while `rust-task-info` is in `support/`.

sagudev:
So components should have all necessary crates for compilation, while support optional supporting stuff (crown, some scripts). It would still be nice to have folder inside components that have non-browser crates (servo_arc, hyper, rand, ...).

components/hyper_serde/Cargo.toml
mrobinson: Oh yep. This is fixed in the latest version of the branch.

sagudev (approved):

PR #30807, opened 2023-12-01
Add basic support for `text-align: justify`

mrobinson:
---


components/layout_2020/flow/inline.rs
```diff
@@ -657,12 +698,30 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
...
                 TextAlign::End => (available_space - line_length).max(text_indent),
                 TextAlign::Center => (available_space - line_length + text_indent) / 2.,
-            }
+            };
+
+        // Calculate the justification adjustment. This is simply the remaining space on the line,
+        // dividided by the number of justficiation opportunities that we recorded when building
+        // the line.
+        let num_justification_opportunities = self.current_line.justification_opportunities as f32;
+        let text_align = self.containing_block.style.clone_text_align();
```
Loirooriol, line 712: Already defined above

components/layout_2020/flow/inline.rs
```diff
@@ -596,12 +639,13 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
 
     /// Given the amount of whitespace trimmed from the line and taking into consideration
     /// the `text-align` property, calculate where the line under construction starts in
-    /// the inline axis.
-    fn calculate_inline_start_for_current_line(
+    /// the inline axis as well as the adjustment needed for every justification opportunity
+    /// to account for `text-align: justify`.
+    fn calculate_current_line_inline_start_and_justification_adjustment(
         &self,
         containing_block: &ContainingBlock,
```
Loirooriol, line 646: Unrelated to this PR, but why have this parameter if it's just `self.containing_block`?

mrobinson:
> ```
> <!DOCTYPE html>
> <link rel="stylesheet" type="text/css" href="https://wpt.live/fonts/ahem.css">
> <div style="font: 10px/1 Ahem; text-align: justify; width: 120px; border: dotted">XXXX&nbsp;XXXX XXXX XXXX XXXX XXXX</div>
> ```

This is a really good point. It seems that this is a pre-existing bug in the sense that it also exists in Layout 2013 and will require changes to the layout-independent Glyph representation. Would it be okay to fix that in a followup?

components/layout_2020/flow/inline.rs
mrobinson: Okay. I've fixed this.

components/layout_2020/flow/inline.rs
mrobinson: Great question. I've gone ahead and fixed this. I'm sure it was just an oversight.

mrobinson:
> > ```
> > <!DOCTYPE html>
> > <link rel="stylesheet" type="text/css" href="https://wpt.live/fonts/ahem.css">
> > <div style="font: 10px/1 Ahem; text-align: justify; width: 120px; border: dotted">XXXX&nbsp;XXXX XXXX XXXX XXXX XXXX</div>
> > ```
> 
> This is a really good point. It seems that this is a pre-existing bug in the sense that it also exists in Layout 2013 and will require changes to the layout-independent Glyph representation. Would it be okay to fix that in a followup?

I've posted the followup for this here: #30866. If it pans out it will fix the issue for both layout systems.

Loirooriol (approved):

PR #30823, opened 2023-12-05
Re-use the TextMetrics data structure in the Layout 2020 fragment tree

mrobinson:
This data structure has all of the metrics needed to render a font and
is in `Au`. We'll need more of these metrics for implementing
`vertical-align` and its use doesn't increase the size of the Fragment
tree (as the BoxFragment is still larger). In addition, this will be
helpful when switching layout to `Au`.

---


atbrakhi (approved):
Looks great. thanks

Loirooriol (approved):

PR #30824, opened 2023-12-05
Generalize LengthPercentageOrAuto impl to Generic Type

atbrakhi:
This PR uses generic type in LengthPercentageOrAuto Implementation. While diverging `components/style` is not very ideal, it is smallish change and is useful in Au conversion, specifically when using auto_is()

---

- [x] These changes are part of adapting `app_units`

### DIFF
--- a/components/style/values/generics/length.rs
+++ b/components/style/values/generics/length.rs
@@ -68,13 +68,13 @@ impl<LengthPercentage> LengthPercentageOrAuto<LengthPercentage> {
     }
 }
 
-impl<U> LengthPercentageOrAuto<U>
+impl<T> LengthPercentageOrAuto<T>
 where
-    U: Clone,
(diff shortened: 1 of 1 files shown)


components/style/values/generics/length.rs
```diff
@@ -68,13 +68,13 @@ impl<LengthPercentage> LengthPercentageOrAuto<LengthPercentage> {
     }
 }
 
-impl<LengthPercentage> LengthPercentageOrAuto<LengthPercentage>
+impl<U> LengthPercentageOrAuto<U>
```
Loirooriol, line 71: Not much familiar with these conventions, but I think it may be a bit strange to use `U` before `T`?
Maybe use `T` here and rename the existing one to `U`?

Loirooriol (approved):

PR #30830, opened 2023-12-06
Remove thinlto servobuild.config option

mrobinson:
Rust uses a version of ThinLTO (thin Local LTO) by default now [^1]. 
This can be tweaked by adjusting rust compiler flags, which is
probably a better way of controlling this than a custom servo
configuration considering:

1. We want to remove the custom servo configuration eventually.
2. The -Z option that this configuration currently uses is unsupported
   by stable rust.

[^1]: https://blog.rust-lang.org/inside-rust/2020/06/29/lto-improvements.html

---

- [x] These changes do not require tests because they just remove a configuration option.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -302,7 +302,6 @@ class CommandBase(object):
         self.config["build"].setdefault("ccache", "")
         self.config["build"].setdefault("rustflags", "")
         self.config["build"].setdefault("incremental", None)
-        self.config["build"].setdefault("thinlto", False)
         self.config["build"].setdefault("webgl-backtrace", False)
         self.config["build"].setdefault("dom-backtrace", False)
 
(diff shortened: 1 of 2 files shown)


sagudev:
Reading https://doc.rust-lang.org/cargo/reference/profiles.html#lto, rustc by default only enables Thin local LTO which is different than just ThinLTO, but this -Z option is replaced with cargo profile option (also available via -C options).

mrobinson:
> Reading https://doc.rust-lang.org/cargo/reference/profiles.html#lto, rustc by default only enables Thin local LTO which is different than just ThinLTO, but this -Z option is replaced with cargo profile option (also available via -C options).

Yep, that's right. The idea is that this will be configurable via making a profile for Servo rather than by `./mach` changing the rustc flags eventually.

mrobinson:
I've clarified that the version of LTO used by default now is (thin local LTO),

sagudev (approved):

PR #30831, opened 2023-12-06
Compile Servo with the latest version of rust stable

mrobinson:
Compile Servo with the latest version of rust stable

This completes the transition to compiling Servo with rust stable. Some
nightly-only features are still used when compiling the `script` and
`crown` crates, as well as for some style unit tests. These will likely
break with newer compiler versions, but `crown` can be disabled for them
conditionally. This is just the first step.

This has some caveats:

1. We need to disable setting up the special linker on Linux. The -Z
   option isn't supported with stable rust so using this is out --
   meanwhile we can't be sure that lld is installed on most systems.
2. `cargo fmt` still uses some unstable options, so we need to rely on
   the unstable toolchain just for running `fmt`. The idea is to fix this
   gradually.

---

### DIFF
--- a/python/servo/platform/linux.py
+++ b/python/servo/platform/linux.py
@@ -150,17 +150,6 @@ class Linux(Base):
         installed_something |= self._platform_bootstrap_gstreamer(force)
         return installed_something
 
-    def linker_flag(self) -> str:
-        # the rust-lld binary downloaded by rustup
-        # doesn't respect NIX_LDFLAGS and also needs
-        # other patches to work correctly. Use system
-        # version of lld for now. See
-        # https://github.com/NixOS/nixpkgs/issues/220717
-        if self.distro.lower() == 'nixos':
-            return '-C link-arg=-fuse-ld=lld'
-        else:
-            return '-Zgcc-ld=lld'
-
     def install_non_gstreamer_dependencies(self, force: bool) -> bool:
         install = False
         pkgs = []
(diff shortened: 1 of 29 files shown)


python/servo/platform/linux.py
```diff
@@ -150,17 +150,6 @@ def _platform_bootstrap(self, force: bool) -> bool:
...
         return installed_something
 
-    def linker_flag(self) -> str:
-        # the rust-lld binary downloaded by rustup
-        # doesn't respect NIX_LDFLAGS and also needs
-        # other patches to work correctly. Use system
-        # version of lld for now. See
-        # https://github.com/NixOS/nixpkgs/issues/220717
-        if self.distro.lower() == 'nixos':
-            return '-C link-arg=-fuse-ld=lld'
```
sagudev, line 160: Can we just have this instead (if we detect that lld is installed)?

mrobinson: In the end, we'd like to make it so that `./mach build` and `cargo build` are compatible. I think the easiest way toward this is removing all of these non-Cargo-compatible ways of setting rust compiler flags. Rust is also gradually moving toward using its own version of ld to link which should speed up linking when it's enabled. (see https://github.com/rust-lang/rust/issues/39915)

mrobinson: Oh, I think it's also interesting to note this doesn't seem to affect compile times on the CI.

sagudev:
It would also be nice to have tracking issue for nightly features that we still use.

Could we hide all things that need RUSTC_BOOTSTRAP behind feature so embededers would not need to set it. (except for crown as embededers do not need it anyway).

sagudev (approved):

PR #30648, opened 2023-10-30
Multiple concurrent top-level browsing contexts

delan:
This patch improves support for multiple top-level browsing contexts aka “webviews”.

We add several messages that allow the embedder to show, hide, move, and resize webviews in the compositor, focus and unfocus webviews in the constellation, and notify the embedder when painting order and focus changes.

These changes can be used to create a multiple-document interface, like we do in #30785.

Parts of this patch have been broken out into:

- [x] #30767
- [x] #30840
- [x] #30841
- [x] #30842
- [x] #31417

<details><summary>Original todo list</summary>

- [x] painting
    - [x] egui then servo, naïve
    - [x] ~~egui then servo or servo then egui, with [stencil](https://learnopengl.com/Advanced-OpenGL/Stencil-testing)~~ — may need egui and webrender modifications
    - [x] interleave egui and servo
    - future: separate webrender documents and partial present?
- [ ] rework:
    - [x] IOCompositor.root_content_pipeline
    - [x] Constellation.active_browser_id
    - [x] servoshell Browser.browser_id
    - [x] {EmbedderEvent,ConstellationMsg}::ChangeBrowserVisibility
    - [ ] use FramebufferPixel in EmbedderCoordinates for flipped coordinates
- [x] fix navigating tlbcs other than the tlbc most recently opened
- [x] independent tlbc:
    - [x] show, hide, raise to top
    - [x] focus, unfocus
    - [x] viewport size
    - [x] viewport location
- [x] closing a tlbc:
    - [x] update IOCompositor.root_content_pipelines
    - [x] update IOCompositor.pipeline_details
    - [x] focus next-most-recently-focused browser
- [x] on click (servoshell):
    - [x] focus tlbc
    - [x] raise tlbc to top
- [x] notify the embedder:
    - [x] when a tlbc is focused
    - [x] when a tlbc is unfocused
    - [x] when the painting order changes
- [x] figure out what to do about handle_is_ready_to_save_image
- [x] ~~fix closing initial tlbc after navigation (other tlbcs are ok)~~
    - this is because is_script_closable is false for same-origin popups
    - same behaviour on main branch
- [x] fix compositor target under --no-minibrowser
</details>

The current design of this feature, after being split and reviewed, is as follows.

## limitations
- desktop-style page zoom is still global
- browsers with transparent or translucent content will misbehave
- sending ShowWebView and HideWebView for one webview at a time can be inefficient
[...]

PR #30840, opened 2023-12-08
Add multiview feature flag in compositing and constellation

delan:
This patch adds a Cargo feature for the new multiple webviews work that is disabled by default.

---

### DIFF
--- a/components/compositing/Cargo.toml
+++ b/components/compositing/Cargo.toml
@@ -14,6 +14,7 @@ path = "lib.rs"
 [features]
 default = []
 gl = ["gleam", "pixels"]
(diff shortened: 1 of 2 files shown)


mrobinson:
Is there a reason that multiview should be behind a flag? Rendering to multiple web views is a pretty basic feature of a web engine. Is the issue that this work isn't complete yet?

delan:
Yeah, in #30648 there have been some existing events with behaviour that we want to change, but we either need to change all in one go to avoid breakage, or the changes have bugs we still need to iron out.

atbrakhi (approved):

PR #30841, opened 2023-12-08
constellation: notify embedder when events are hit-tested to browsers

delan:
This patch adds a new EventDelivered event (embedder ← constellation) that notifies the embedder when a mouse or touch event was hit-tested and delivered to a pipeline in some top-level browsing context.

This will allow the embedder to, for example, focus and raise a browser when clicked.

---

- [x] These changes do not require tests because the behaviour changes are straightforward

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2913,17 +2913,7 @@ where
             },
             Some(pipeline) => pipeline,
         };
-        let event_variant = match event {
-            CompositorEvent::ResizeEvent(..) => CompositorEventVariant::ResizeEvent,
-            MouseButtonEvent(..) => CompositorEventVariant::MouseButtonEvent,
-            MouseMoveEvent(..) => CompositorEventVariant::MouseMoveEvent,
-            CompositorEvent::TouchEvent(..) => CompositorEventVariant::TouchEvent,
-            CompositorEvent::WheelEvent(..) => CompositorEventVariant::WheelEvent,
-            CompositorEvent::KeyboardEvent(..) => CompositorEventVariant::KeyboardEvent,
-            CompositorEvent::CompositionEvent(..) => CompositorEventVariant::CompositionEvent,
-            CompositorEvent::IMEDismissedEvent => CompositorEventVariant::IMEDismissedEvent,
-        };
-        let msg = EmbedderMsg::EventDelivered(event_variant);
+        let msg = EmbedderMsg::EventDelivered((&event).into());
         self.embedder_proxy
             .send((Some(pipeline.top_level_browsing_context_id), msg));
 
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -113,7 +113,8 @@ use devtools_traits::{
     ScriptToDevtoolsControlMsg,
 };
 use embedder_traits::{
-    Cursor, EmbedderMsg, EmbedderProxy, MediaSessionEvent, MediaSessionPlaybackState,
+    CompositorEventVariant, Cursor, EmbedderMsg, EmbedderProxy, MediaSessionEvent,
+    MediaSessionPlaybackState,
 };
 use euclid::default::Size2D as UntypedSize2D;
 use euclid::Size2D;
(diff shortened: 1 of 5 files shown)


components/constellation/constellation.rs
```diff
@@ -2904,15 +2906,29 @@ where
...
+            MouseMoveEvent(..) => CompositorEventVariant::MouseMoveEvent,
+            CompositorEvent::TouchEvent(..) => CompositorEventVariant::TouchEvent,
+            CompositorEvent::WheelEvent(..) => CompositorEventVariant::WheelEvent,
+            CompositorEvent::KeyboardEvent(..) => CompositorEventVariant::KeyboardEvent,
+            CompositorEvent::CompositionEvent(..) => CompositorEventVariant::CompositionEvent,
+            CompositorEvent::IMEDismissedEvent => CompositorEventVariant::IMEDismissedEvent,
+        };
+        let msg = EmbedderMsg::EventDelivered(event_variant);
+        self.embedder_proxy
+            .send((Some(pipeline.top_level_browsing_context_id), msg));
```
mrobinson, line 2928: I would avoid the temporary `msg` here.

ports/servoshell/browser.rs
```diff
@@ -545,6 +545,23 @@ where
...
+                        CompositorEventVariant::TouchEvent |
+                        CompositorEventVariant::WheelEvent |
+                        CompositorEventVariant::KeyboardEvent |
+                        CompositorEventVariant::CompositionEvent |
+                        CompositorEventVariant::IMEDismissedEvent,
+                    ) => {},
+                    (Some(browser_id), CompositorEventVariant::MouseButtonEvent) => {
+                        // TODO Focus browser and/or raise to top if needed.
+                        trace!("{}: Got a mouse button event", browser_id);
+                    },
```
mrobinson, line 563: This will make this a bit shorter and easier to read, I think:

```suggestion
                    (Some(browser_id), CompositorEventVariant::MouseButtonEvent) => {
                        // TODO Focus browser and/or raise to top if needed.
                        trace!("{}: Got a mouse button event", browser_id);
                    },
                    (_, _) => {},
```

mrobinson (approved):
Seems fine to me! I have a few small suggestions before landing:

atbrakhi (approved):
Currently, it will cause the `libsimpleservo` build to fail. We should add `EmbedderMsg::EventDelivered` in `handle_servo_events` of libsimpleservo as well.

Apart from that, it looks good to me

delan:
> Currently, it will cause the `libsimpleservo` build to fail. We should add `EmbedderMsg::EventDelivered` in `handle_servo_events` of libsimpleservo as well.

Done in 89b2da67582e830789084fa8e7684c6e8d60a6a6, though it still fails to compile on Linux due to #30865.

components/shared/embedder/lib.rs
```diff
@@ -208,6 +208,21 @@ pub enum EmbedderMsg {
...
+#[derive(Debug, Deserialize, Serialize)]
+pub enum CompositorEventVariant {
+    ResizeEvent,
+    MouseButtonEvent,
+    MouseMoveEvent,
+    TouchEvent,
+    WheelEvent,
+    KeyboardEvent,
+    CompositionEvent,
+    IMEDismissedEvent,
```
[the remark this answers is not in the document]
delan: done in f3a770ecc5f23d70c4e31209515c82fa0e06c1da, though i had to impl From&lt;&amp;CompositorEvent> and there was only one location needing that kind of conversion

components/constellation/constellation.rs
delan: done in 5f9b48d37f82f63e9da7feca448a26b1493028dc

delan: done in 5f9b48d37f82f63e9da7feca448a26b1493028dc

ports/servoshell/browser.rs
delan: done in 318c02ec839185d148db93cf3bc0b4ee4be9a37e

PR #30842, opened 2023-12-08
constellation: focusing and closing webviews

delan:
This patch improves support in the constellation for multiple top-level browsing contexts aka “webviews”, or in some older code, “browsers”.

We also fix two existing bugs in the logic for closing webviews: one where browsing context groups always leak, and one where browsing context groups get removed even when non-empty.

## constellation
- WebviewManager (new) manages webview-related storage and focus order
    - it allows the constellation to store arbitrary data for each webview id
    - it keeps track of which webview is focused, or whether no webview is focused
    - it can refocus the last-focused webview when a webview is removed
- we now send WebviewClosed directly when a webview is closed rather than relying on script to do it, so that the event gets sent even if the webview was not closed by script

## servoshell
- BrowserManager (was Browser) keeps a copy of the webview focus and creation order
- but for now, we always focus and show the newest webview only

## messages
embedder →\* constellation
- NewWebview (was NewBrowser) behaves as before
- CloseWebview (was CloseBrowser) behaves as before
- FocusWebview (new) makes a webview focused for keyboard events
- BlurWebview (new) makes no webview focused for keyboard events
- WebviewVisibilityChanged (was ChangeBrowserVisibility) behaves as before

embedder ← constellation
- AllowOpeningWebview (was AllowOpeningBrowser) behaves as before
- WebviewOpened (was BrowserCreated) notifies embedder that a webview was opened
- WebviewClosed (was CloseBrowser) notifies embedder that a webview was closed
- WebviewFocused (new) notifies embedder that a webview gained focus for keyboard events
- WebviewBlurred (new) notifies embedder that all webviews lost focus for keyboard events

---


components/compositing/windowing.rs
```diff
@@ -101,8 +101,8 @@ pub enum EmbedderEvent {
     /// Sent when the user triggers a media action through the UA exposed media UI
     /// (play, pause, seek, etc.).
     MediaSessionAction(MediaSessionActionType),
-    /// Set browser visibility. A hidden browser will not tick the animations.
-    ChangeBrowserVisibility(TopLevelBrowsingContextId, bool),
+    /// The visibility of the native window has changed.
+    WindowVisibility(bool),
```
mrobinson, line 105: I'm curious why this now is a property of the window, rather than the "Browser". Theoretically, an application that supports tabs may want to adjust the visibility of each WebView individually.

components/constellation/browser.rs
```diff
@@ -0,0 +1,262 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use std::collections::{HashMap, HashSet};
+
+use msg::constellation_msg::TopLevelBrowsingContextId;
+
+#[derive(Debug)]
+pub struct BrowserManager<Browser> {
```
mrobinson, line 10: The "Browser" terminology is genuinely confusing! Can we switch to something like "WebView" which is much clearer about what these objects are? Obviously there's a lot of code that uses "Browser," but we can fix that iteratively.

components/constellation/browser.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::num::NonZeroU32;
+
+    use msg::constellation_msg::{
+        BrowsingContextId, BrowsingContextIndex, PipelineNamespace, PipelineNamespaceId,
+        TopLevelBrowsingContextId,
```
mrobinson, line 146: Nice tests!

components/constellation/constellation.rs
```diff
@@ -1483,9 +1476,26 @@ where
...
+                    EmbedderMsg::BrowserFocused(top_level_browsing_context_id),
+                ));
+                if !cfg!(feature = "multiview") {
+                    self.update_frame_tree_if_focused(top_level_browsing_context_id);
+                }
+            },
+            FromCompositorMsg::UnfocusBrowser => {
+                self.browsers.unfocus();
+                self.embedder_proxy
+                    .send((None, EmbedderMsg::BrowserUnfocused));
```
mrobinson, line 1498: Sorry, this is a serious nit, but how would you feel about "Blur" terminology to match the web?

components/constellation/constellation.rs
```diff
@@ -2996,22 +3013,44 @@ where
...
-        self.browsing_context_group_set
-            .remove(&browsing_context.bc_group_id);
+        self.browsers.remove(top_level_browsing_context_id);
+        // TODO Send the compositor a RemoveBrowser event.
+        self.embedder_proxy.send((
+            Some(top_level_browsing_context_id),
+            EmbedderMsg::BrowserClosed(top_level_browsing_context_id),
+        ));
+
+        if let Some(browsing_context) = browsing_context {
```
mrobinson, line 3032: Let's stick with the early return here as it decreases the indent level making this easier to read. This a bit more preferred in Servo style, I think.

components/constellation/constellation.rs
```diff
@@ -2996,22 +3013,44 @@ where
...
+            // https://html.spec.whatwg.org/multipage/#bcg-remove
+            let bc_group_id = browsing_context.bc_group_id;
+            if let Some(bc_group) = self.browsing_context_group_set.get_mut(&bc_group_id) {
```
mrobinson, line 3035: Ditto. This will also make it easier to see where warnings are needed to be printed.

components/constellation/constellation.rs
```diff
@@ -5346,17 +5393,10 @@ where
...
-            // type system.
-            .or_insert_with(|| Browser {
-                focused_browsing_context_id: BrowsingContextId::from(top_level_id),
-                session_history: JointSessionHistory::new(),
-            })
-            .session_history
+        self.browsers
+            .get_mut(top_level_id)
+            .map(|browser| &mut browser.session_history)
+            .expect("Unknown top-level browsing context")
```
mrobinson, line 5399: Nice cleanup!

components/constellation/constellation.rs
```diff
@@ -5388,29 +5428,26 @@ where
             })
     }
 
-    /// Re-send the frame tree to the compositor.
-    fn update_frame_tree_if_active(
+    /// Send the frame tree for the given browser to the compositor.
+    fn update_frame_tree_if_focused(
```
mrobinson, line 5432: This name is much clearer, but I think it also needs to reflect that it will cause the the TopLevelBrowsingContext to be focused if it isn't already.

components/constellation/constellation.rs
```diff
@@ -5388,29 +5428,26 @@ where
...
-        {
-            self.send_frame_tree(top_level_browsing_context_id);
+        // Only send the frame tree if the given browser is focused, or if no
+        // browser is currently focused, make it focused.
+        if let Some(focused_browser_id) = self.browsers.focused_browser().map(|(id, _)| id) {
+            if top_level_browsing_context_id != focused_browser_id {
+                return;
+            }
+        } else {
+            self.browsers.focus(top_level_browsing_context_id);
```
mrobinson, line 5443: Is this what we'd like to do? For instance, imagine a web view that is integrated into some other UI system (this actually happens a lot). This will cause the web view to focus meaning an input element might start to appear active -- in addition to whatever external UI element has focus. Shouldn't the embedder completely control when the web view has focus?

components/script/dom/window.rs
```diff
@@ -788,13 +788,11 @@ impl WindowMethods for Window {
                     let window = this.root();
                     let document = window.Document();
                     // https://html.spec.whatwg.org/multipage/#closing-browsing-contexts
-                    // Step 1, prompt to unload.
+                    // Step 3, prompt to unload.
```
mrobinson, line 791: It's probably worth comment on what happened to step 1 and 2 here. :)

components/servo/lib.rs
```diff
@@ -636,11 +636,11 @@ where
                 }
             },
 
-            EmbedderEvent::SelectBrowser(top_level_browsing_context_id) => {
-                let msg = ConstellationMsg::SelectBrowser(top_level_browsing_context_id);
+            EmbedderEvent::FocusBrowser(top_level_browsing_context_id) => {
```
mrobinson, line 639: "Focus" is much clearer here!

components/shared/compositing/constellation_msg.rs
```diff
@@ -102,15 +104,16 @@ impl fmt::Debug for ConstellationMsg {
...
             SendError(..) => "SendError",
-            SelectBrowser(..) => "SelectBrowser",
             ForwardEvent(..) => "ForwardEvent",
             SetCursor(..) => "SetCursor",
             EnableProfiler(..) => "EnableProfiler",
             DisableProfiler => "DisableProfiler",
             ExitFullScreen(..) => "ExitFullScreen",
             MediaSessionAction(..) => "MediaSessionAction",
-            ChangeBrowserVisibility(..) => "ChangeBrowserVisibility",
+            WindowVisibility(..) => "WindowVisibility",
```
mrobinson, line 116: I think this is a bit funky. The embedder should have a concept of a window, but I'm not sure that Constellation should (at least for now). Instead, the embedder should send a visibility event for each web view when their window is hidden. The reason I think this is that maybe the embedder wants to move a web view to a different window. This should totally be possible without destroying it and recreating the web view completely. If the Constellation is managing associating web views to windows, then it will require complex bookkeeping in both the embedder and the Constellation to track movement of web views to different windows.

mrobinson (commented):
This looks great to me. I have some comments, some of them quite trivial, but I think the most important conversation is the one about whether the Constellation should have a concept of a window.

components/constellation/browser.rs
mrobinson: An alternative here is ServoView if we want to match the GeckoView terminology. I'm not super attached to any name, as much as opposed to 'Browser' due to the overloaded meaning.

components/constellation/constellation.rs
delan: done in 4aea74fae2e6ae328104ce45fc63336941366cd7

components/constellation/browser.rs
delan: done in 4aea74fae2e6ae328104ce45fc63336941366cd7

components/compositing/windowing.rs
delan: iirc the idea was that this would make things easier for the embedder under this intermediate model of multiple webviews but only one window, even though we would probably need to change it once we do multiple windows. but letting the embedder handle it would work too, while hugely reducing complexity.

delan: fixed in 9e307cfcefed26ecb65399ce12a7e2a7ef38e5d7

components/shared/compositing/constellation_msg.rs
delan: i agree. fixed in 9e307cfcefed26ecb65399ce12a7e2a7ef38e5d7

components/constellation/constellation.rs
delan: done in 3c910aa12c10955e72021e4ba3bb276c0cc84f59

components/constellation/constellation.rs
delan: done in 3c910aa12c10955e72021e4ba3bb276c0cc84f59

components/constellation/browser.rs
delan: thanks!

components/constellation/constellation.rs
delan: no longer needed as of 2afa67966ac13eb312cbf0fa616cbfe7c3790f4f

components/constellation/constellation.rs
delan: yeah good point, i think i did this to preserve existing behaviour since i didn’t understand the session history and frame tree stuff well enough, but i agree. fixed in 2afa67966ac13eb312cbf0fa616cbfe7c3790f4f

components/script/dom/window.rs
delan: done in f3a32f480a8afeaa1731712df8ee879a9822d04f

components/constellation/constellation.rs
```diff
@@ -31,7 +31,7 @@
 //! to a forest whose roots are top-level browsing context.  The logical
 //! relationship between these types is:
 //!
-//! ```
+//! ```text
```
mrobinson, line 34: :heart:

components/constellation/constellation.rs
```diff
@@ -223,15 +224,15 @@ struct WebrenderWGPU {
     wgpu_image_map: Arc<Mutex<HashMap<u64, webgpu::PresentationData>>>,
 }
 
-/// Servo supports tabs (referred to as browsers), so `Constellation` needs to
-/// store browser specific data for bookkeeping.
-struct Browser {
-    /// The currently focused browsing context in this browser for key events.
+/// Servo supports multiple top-level browsing contexts or “webviews”, so `Constellation` needs to
+/// store webview-specific data for bookkeeping.
+struct Webview {
```
mrobinson, line 229: I think the consensus might be WebView (two words, but together) based on the [Wikipedia article](https://en.wikipedia.org/wiki/WebView) (with does also use "Webview", but more often "WebView") and just consistency with WebKit. If it's not too much of a hassle, could you change it to the camel case version?

```suggestion
struct WebView {
```

components/constellation/constellation.rs
```diff
@@ -3000,22 +3010,46 @@ where
         &mut self,
         top_level_browsing_context_id: TopLevelBrowsingContextId,
     ) {
+        debug!("{}: Closing", top_level_browsing_context_id);
```
mrobinson, line 3013:
```suggestion
        debug!("{top_level_browsing_context_id}: Closing");
```

components/constellation/constellation.rs
```diff
@@ -3000,22 +3010,46 @@ where
...
+            return;
+        };
+        if !bc_group
+            .top_level_browsing_context_set
+            .remove(&top_level_browsing_context_id)
+        {
+            warn!(
+                "{}: Top-level browsing context not found in {}",
+                top_level_browsing_context_id, bc_group_id
+            );
```
mrobinson, line 3045:
```suggestion
            warn!(
                "{top_level_browsing_context_id}: Top-level browsing context not found in {bc_group_id}",
            );
```

components/constellation/constellation.rs
```diff
@@ -3000,22 +3010,46 @@ where
...
+        }
+        if bc_group.top_level_browsing_context_set.is_empty() {
+            self.browsing_context_group_set
+                .remove(&browsing_context.bc_group_id);
+        }
+
+        debug!("{}: Closed", top_level_browsing_context_id);
```
mrobinson, line 3052:
```suggestion
        debug!("{top_level_browsing_context_id}: Closed");
```

components/constellation/constellation.rs
```diff
@@ -4438,13 +4480,13 @@ where
             Some(browsing_context) => browsing_context.pipeline_id,
             None => {
                 return warn!(
-                    "{}: Got visibility change event after closure",
+                    "{}: Tried to notify visibility after closure",
                     browsing_context_id
```
mrobinson, line 4484:
```suggestion
                    "{browsing_context_id}: Tried to notify visibility after closure",
```

components/constellation/constellation.rs
```diff
@@ -4438,13 +4480,13 @@ where
...
                 );
             },
         };
         match self.pipelines.get(&pipeline_id) {
-            None => return warn!("{}: Got visibility change event after closure", pipeline_id),
+            None => return warn!("{}: Tried to notify visibility after closure", pipeline_id),
```
mrobinson, line 4489:
```suggestion
            None => return warn!("{pipeline_id}: Tried to notify visibility after closure"),
```

components/constellation/constellation.rs
```diff
@@ -5157,7 +5195,10 @@ where
 
         let browsing_context = match self.browsing_contexts.remove(&browsing_context_id) {
             Some(ctx) => ctx,
-            None => return warn!("{}: Closing twice", browsing_context_id),
+            None => {
+                warn!("{}: Closing twice", browsing_context_id);
```
mrobinson, line 5199:
```suggestion
                warn!("{browsing_context_id}: Closing twice");
```

components/constellation/constellation.rs
```diff
@@ -5168,12 +5209,13 @@ where
         if let Some(parent_pipeline_id) = browsing_context.parent_pipeline_id {
             match self.pipelines.get_mut(&parent_pipeline_id) {
                 None => {
-                    return warn!("{}: Child closed after parent", parent_pipeline_id);
+                    warn!("{}: Child closed after parent", parent_pipeline_id);
```
mrobinson, line 5212:
```suggestion
                    warn!("{parent_pipeline_id}: Child closed after parent");
```

mrobinson (approved):
Thanks! Just a couple comments below, but otherwise this looks good.

PR #30685, opened 2023-11-02
Update mozjs and setuptools

mrobinson:
This should fix issues with Python 3.12 on Mac builders.

Co-authored-by: Samson <16504129+sagudev@users.noreply.github.com>

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes do not require tests because they fix the build on Python 3.12 builders.

PR #30848, opened 2023-12-10
Use os version in taplo cache-key & setuptools 65 in py3.8

### LINKED ISSUE #30835: No new nightly builds?
https://github.com/servo/servo-nightly-builds

Latest build on GitHub and servo.org is from two days ago. Have nightly versions stopped being built?

sagudev:
It works: https://github.com/servo/servo/actions/runs/7167362715

---

- [x] These changes fix #30835 
- [x] There are tests for these changes (test run)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -86,7 +86,7 @@ jobs:
         with:
           crate: taplo-cli
           locked: true
-          hash: ${{ inputs.upload && '20.04' || '22.04' }}
+          cache-key: ${{ inputs.upload && '20.04' || '22.04' }}
       - name: Bootstrap Python
         run: python3 -m pip install --upgrade pip
       - name: Bootstrap dependencies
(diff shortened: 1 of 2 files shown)


mrobinson:
I wonder a bit if this is hiding the issue though? Why are things broken with Python 3.8 and the version of setuptools we are using. I notice that `requirements.txt` has this:

```
setuptools == 68.2.2; python_version >= "3.8"
setuptools == 65.5.1; python_version < "3.8"
```

Looks like we did this here: https://github.com/servo/servo/pull/30685

sagudev:
> I wonder a bit if this is hiding the issue though? Why are things broken with Python 3.8 and the version of setuptools we are using. I notice that `requirements.txt` has this:
> 
> ```
> setuptools == 68.2.2; python_version >= "3.8"
> setuptools == 65.5.1; python_version < "3.8"
> ```
> 
> Looks like we did this here: #30685

We probably need to move `=` down.

mrobinson (approved):
Thanks for this fix!

PR #30867, opened 2023-12-12
Fix two more warnings in WebRender

### LINKED ISSUE #30861: "variable does not need to be mutable" warnings in webrender
When doing `./mach build`:

```
warning: variable does not need to be mutable
  --> third_party/webrender/webrender_api/src/display_item_cache.rs:61:13
   |
61 |         let mut entry = &mut self.entries[key as usize];
   |             ----^^^^^
   |             |
   |             help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default

warning: variable does not need to be mutable
  --> third_party/webrender/webrender_api/src/display_item_cache.rs:67:13
   |
67 |         let mut entry = &mut self.entries[key as usize];
   |             ----^^^^^
   |             |
   |             help: remove this `mut`
```

Firefox addressed this in https://bugzilla.mozilla.org/show_bug.cgi?id=1841919

mrobinson:
These appeared after switching to the latest Rust stable release.

---

- [x] These changes fix #30861.
- [x] These changes do not require tests because they fix warnings.


delan (approved):
Neat! I didn’t know we had a system for patching WebRender :)

PR #30868, opened 2023-12-12
Add support for table fixups

mrobinson:
This adds support for fixing up tables so that internal table elements
that are not properly parented in the DOM have the correct box tree
structure according to the CSS Table specification [1]. Note that this
only comes into play when building the DOM via script, as HTML 5 has its
own table fixups that mean that the box tree construction fixups here
are not necessary.

There are no tests for this change. In general, it's hard to write tests
against the shape of the box tree, because it depends on the DOM. We
plan to test this via WPT tests once layout is complete.

1. https://drafts.csswg.org/css-tables/#table-internal-element

---

- [x] These changes do not require tests (see commit message)

### DIFF
--- a/components/layout_2020/flow/construct.rs
+++ b/components/layout_2020/flow/construct.rs
@@ -11,6 +11,7 @@ use style::computed_values::white_space::T as WhiteSpace;
 use style::properties::longhands::list_style_position::computed_value::T as ListStylePosition;
 use style::properties::ComputedValues;
 use style::selector_parser::PseudoElement;
+use style::str::char_is_whitespace;
 use style::values::specified::text::TextDecorationLine;
 
 use crate::cell::ArcRefCell;
@@ -23,9 +24,10 @@ use crate::flow::{BlockContainer, BlockFormattingContext, BlockLevelBox};
 use crate::formatting_contexts::IndependentFormattingContext;
 use crate::positioned::AbsolutelyPositionedBox;
 use crate::style_ext::{ComputedValuesExt, DisplayGeneratingBox, DisplayInside, DisplayOutside};
+use crate::table::{AnonymousTableContent, Table};
 
 impl BlockFormattingContext {
-    pub fn construct<'dom, Node>(
+    pub(crate) fn construct<'dom, Node>(
         context: &LayoutContext,
         info: &NodeAndStyleInfo<Node>,
         contents: NonReplacedContents,
@@ -35,15 +37,17 @@ impl BlockFormattingContext {
     where
         Node: NodeExt<'dom>,
     {
-        let contents = BlockContainer::construct(
+        Self::from_block_container(BlockContainer::construct(
             context,
             info,
             contents,
             propagated_text_decoration_line,
             is_list_item,
-        );
-        let contains_floats = contents.contains_floats();
+        ))
+    }
 
+    pub(crate) fn from_block_container(contents: BlockContainer) -> Self {
+        let contains_floats = contents.contains_floats();
         Self {
             contents,
             contains_floats,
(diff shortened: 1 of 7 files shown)


components/layout_2020/flow/construct.rs
```diff
@@ -214,34 +214,120 @@ impl BlockContainer {
...
+        // Text decorations are not propagated to atomic inline-level descendants.
+        let inline_table = !self.ongoing_inline_boxes_stack.is_empty();
+        let propagated_text_decoration_line = if inline_table {
+            TextDecorationLine::NONE
+        } else {
+            self.ongoing_inline_formatting_context.text_decoration_line
+        };
+
+        let contents: Vec<AnonymousTableContent<'dom, Node>> =
+            self.anonymous_table_content.drain(..).collect();
```
Loirooriol, line 291: Would `let contents = std::mem::take(&mut self.anonymous_table_content);` be more performant?
Ditto in `components/layout_2020/table/construct.rs`

components/layout_2020/flow/construct.rs
```diff
@@ -214,34 +214,120 @@ impl BlockContainer {
...
+        let last_text = match contents.last() {
+            Some(AnonymousTableContent::Text(info, text)) => Some((info.clone(), text.clone())),
```
Loirooriol, line 293: Instead of cloning, maybe just `pop()` the last item from `contents`?

Loirooriol (approved):

components/layout_2020/flow/construct.rs
mrobinson: I don't think we can use `std::mem::take` here since that does a move, but maybe we could do `std::mem::replace`. I'm not sure if it's more performant, but I think it's cleaner, so I'll make that change.

components/layout_2020/flow/construct.rs
mrobinson: `pop()` here would be a little awkward, because if it's not text, we'd either have to push it back on or process it specially below (after matching again). I think this might be the cleanest way to do it after investigating. Cloning the text in this case shouldn't be too expensive, because it's behind a `Cow` smart-pointer.

components/layout_2020/flow/construct.rs
Loirooriol: `std::mem::take` compiles fine for me, and it seems preferred: https://github.com/rust-lang/rust-clippy/issues/10818

components/layout_2020/flow/construct.rs
Loirooriol: It also seems weird to handle the same node in 2 places, even if it ends up being fine because `construct_anonymous` ignores texts.

So even with the current approach, it could be good to pop.

components/layout_2020/flow/construct.rs
Loirooriol: Didn't address this case in `flow/construct.rs`, just one in `table/construct.rs`.

There is another `drain(..)` in `table/construct.rs` but without the `collect()`, so it may be better as-is.

mrobinson: > `std::mem::take` compiles fine for me, and it seems preferred: [rust-lang/rust-clippy#10818](https://github.com/rust-lang/rust-clippy/issues/10818)

> Didn't address this case in flow/construct.rs, just one in table/construct.rs.
> There is another drain(..) in table/construct.rs but without the collect(), so it may be better as-is.

Huh...I tried this, but for some reason it didn't work for me. Thanks for the fixup.

mrobinson:
Looks like there was a visibility issue for `TableBuilder` used in the unit tests. I've corrected that. Not sure how it was introduced.

sagudev:
OFF-TOPIC: try runs are not working: https://github.com/servo/servo/actions/runs/7299382998/job/19892100531

PR #30894, opened 2023-12-19
Use App units in flow layout

### LINKED ISSUE #29819: Layout 2020 is vulnerable to floating point precision issues

atbrakhi:
---

- [x] These changes are part of #29819

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -133,11 +133,14 @@ impl<'a> PlacementAmongFloats<'a> {
 
     /// The height of the bands under consideration.
     fn current_bands_height(&self) -> Au {
-        if let Some(top) = self.top_of_bands() {
-            self.next_band.top - top
+        if self.next_band.top == MAX_AU {
+            // Treat MAX_AU as infinity.
+            MAX_AU
         } else {
-            assert!(self.next_band.top == MAX_AU);
-            self.next_band.top
+            let top = self
+                .top_of_bands()
+                .expect("Should have bands before reaching the end");
+            self.next_band.top - top
         }
     }
 
@@ -191,15 +194,14 @@ impl<'a> PlacementAmongFloats<'a> {
         if available_inline_size < self.object_size.inline {
             return None;
         }
-        let top = self.top_of_bands().unwrap();
         Some(LogicalRect {
             start_corner: LogicalVec2 {
                 inline: inline_start,
-                block: top,
+                block: self.top_of_bands().unwrap(),
             },
             size: LogicalVec2 {
                 inline: available_inline_size,
-                block: self.next_band.top - top,
+                block: self.current_bands_height(),
             },
         })
     }
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -8,15 +8,16 @@
 
 use std::collections::VecDeque;
 use std::fmt::{Debug, Formatter, Result as FmtResult};
+use std::mem;
 use std::ops::Range;
-use std::{f32, mem};
 
+use app_units::{Au, MAX_AU, MIN_AU};
 use euclid::num::Zero;
 use serde::Serialize;
 use servo_arc::Arc;
 use style::computed_values::float::T as FloatProperty;
 use style::properties::ComputedValues;
-use style::values::computed::{CSSPixelLength, Clear, Length};
+use style::values::computed::{Clear, Length};
 use style::values::specified::text::TextDecorationLine;
 
 use crate::context::LayoutContext;
@@ -24,7 +25,7 @@ use crate::dom::NodeExt;
 use crate::dom_traversal::{Contents, NodeAndStyleInfo};
 use crate::formatting_contexts::IndependentFormattingContext;
 use crate::fragment_tree::{BoxFragment, CollapsedBlockMargins, CollapsedMargin, FloatFragment};
-use crate::geom::{LogicalRect, LogicalVec2};
+use crate::geom::{LogicalRect, LogicalSides, LogicalVec2};
 use crate::positioned::PositioningContext;
 use crate::style_ext::{ComputedValuesExt, DisplayInside, PaddingBorderMargin};
 use crate::ContainingBlock;
@@ -47,7 +48,7 @@ pub struct ContainingBlockPositionInfo {
     /// containing block, excluding uncollapsed block start margins. Note that
     /// this does not include uncollapsed block start margins because we don't
     /// know the value of collapsed margins until we lay out children.
-    pub(crate) block_start: Length,
+    pub(crate) block_start: Au,
     /// Any uncollapsed block start margins that we have collected between the
     /// block start of the float containing independent block formatting context
     /// and this containing block, including for this containing block.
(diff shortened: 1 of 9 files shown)


components/layout_2020/flow/float.rs
```diff
@@ -84,66 +86,66 @@ pub(crate) struct PlacementAmongFloats<'a> {
...
-        ceiling: Length,
-        object_size: LogicalVec2<Length>,
+        ceiling: Au,
+        object_size: LogicalVec2<Au>,
         pbm: &PaddingBorderMargin,
     ) -> Self {
-        assert!(!ceiling.px().is_infinite());
-        let mut current_band = float_context.bands.find(ceiling).unwrap();
-        current_band.top = ceiling;
+        assert!(ceiling.to_f32_px() != INFINITY || ceiling.to_f32_px() != NEG_INFINITY);
```
Loirooriol, line 107: Use `&&` instead of `||`?

Anyways, this makes me realize that most of these `is_infinite()` were actually meant to just test for positive infinity.

components/layout_2020/flow/float.rs
```diff
@@ -84,66 +86,66 @@ pub(crate) struct PlacementAmongFloats<'a> {
...
     }
 
     /// The height of the bands under consideration.
-    fn current_bands_height(&self) -> Length {
+    fn current_bands_height(&self) -> Au {
         if let Some(top) = self.top_of_bands() {
             self.next_band.top - top
         } else {
-            assert!(self.next_band.top.px().is_infinite());
+            assert!(self.next_band.top != Au(i32::MAX) || self.next_band.top != Au(i32::MIN));
```
Loirooriol, line 140: Use `==` instead of `!=`.

components/layout_2020/flow/inline.rs
```diff
@@ -2407,3 +2410,27 @@ fn char_prevents_soft_wrap_opportunity_when_before_or_after_atomic(character: ch
...
 }
+
+fn to_length(value: &LogicalVec2<CSSPixelLength>) -> LogicalVec2<Au> {
+    LogicalVec2 {
+        inline: value.inline.into(),
+        block: value.block.into(),
+    }
+}
+
+fn to_au_rect(value: LogicalRect<Au>) -> LogicalRect<CSSPixelLength> {
```
mrobinson, line 2421: Could these functions be moved into `geom.rs` and implemented as either `From` or `Into` trait implementations? That way they could be shared.

components/layout_2020/flow/inline.rs
```diff
@@ -2407,3 +2410,27 @@ fn char_prevents_soft_wrap_opportunity_when_before_or_after_atomic(character: ch
...
+    LogicalRect {
+        start_corner: LogicalVec2 {
+            inline: value.start_corner.inline.into(),
+            block: value.start_corner.block.into(),
+        },
+        // If this float has a negative margin, we should only consider its non-negative
+        // block size contribution when determing where to place it. When the margin is
+        // so negative that it's placed completely above the current float ceiling, then
+        // we should position it as if it had zero block size.
```
mrobinson, line 2430: I'm not sure this comment belongs here? It doesn't really have anything to do with conversion of units.

components/layout_2020/tests/floats.rs
```diff
@@ -547,9 +573,9 @@ fn check_floats_rule_3(placement: &FloatPlacement) {
             // Where the top of `b` should probably be 32px per Rule 3, but unless this distinction
             // is made the top of `b` could legally be 0px.
             if this_float.origin.block >= other_float.rect().max_block_position() ||
-                (this_float.info.size.block == Length::zero() &&
+                (this_float.info.size.block == Au::zero() &&
                     this_float.rect().max_block_position() < other_float.origin.block) ||
-                (this_float.info.size.block > Length::zero() &&
+                (this_float.info.size.block > Au::zero() &&
```
mrobinson, line 578: I think you can do `is_zero()` instead of comparing to zero here.

mrobinson (changes requested):

atbrakhi:
@mrobinson @Loirooriol thanks for the review. I have two unit test failures in `floats` because of these changes. I am investigating them. I will update the PR with your suggestions as soon as I resolve them.

components/layout_2020/flow/inline.rs
atbrakhi: right, i have removed it.

components/layout_2020/tests/floats.rs
atbrakhi: it seems we cannot directly use `is_zero()` here. While size is of type `Au`, block here is of a generic type `T`.

atbrakhi:
not sure what's wrong. At first glance, the majority of the failures don't seem to be related to the changes in the PR.  Most are either `Crash from webaudio` or `Timeouts in semantics`.

components/layout_2020/flow/float.rs
```diff
@@ -84,66 +86,66 @@ pub(crate) struct PlacementAmongFloats<'a> {
...
-        current_band.top = ceiling;
+        assert!(ceiling.to_f32_px() != INFINITY);
+        let mut current_band = float_context.bands.find(ceiling.into()).unwrap();
+        current_band.top = ceiling.into();
         let current_bands = VecDeque::from([current_band]);
-        let next_band = float_context.bands.find_next(ceiling).unwrap();
+        let next_band = float_context.bands.find_next(ceiling.into()).unwrap();
         let min_inline_start = float_context.containing_block_info.inline_start +
-            pbm.margin.inline_start.auto_is(Length::zero);
+            pbm.margin.inline_start.auto_is(|| Au::zero().into()).into();
```
Loirooriol, line 113: I would keep `Length::zero` for now instead of `|| Au::zero().into()`

components/layout_2020/flow/float.rs
```diff
@@ -84,66 +86,66 @@ pub(crate) struct PlacementAmongFloats<'a> {
...
-        object_size: LogicalVec2<Length>,
+        ceiling: Au,
+        object_size: LogicalVec2<Au>,
         pbm: &PaddingBorderMargin,
     ) -> Self {
-        assert!(!ceiling.px().is_infinite());
-        let mut current_band = float_context.bands.find(ceiling).unwrap();
-        current_band.top = ceiling;
+        assert!(ceiling.to_f32_px() != INFINITY);
+        let mut current_band = float_context.bands.find(ceiling.into()).unwrap();
```
Loirooriol, line 108: I don't think the `into()` is needed.

mrobinson:
> not sure what's wrong. At first glance, the majority of the failures don't seem to be related to the changes in the PR. Most are either `Crash from webaudio` or `Timeouts in semantics`.

I think the large amount of crashes is due to running an old version of Servo against the current CI configuration. The CI now installs GStreamer via apt which I think requires the same version of Servo to run properly.

components/layout_2020/flow/float.rs
atbrakhi: why would you prefer to keep `Length::zero` for now? I am sure you have a reasoning behind it, so i have changed it to use `Length::zero` but i would also like to understand the reason.

Loirooriol: It seems strange to create `Au:zero()` just to immediately coerce it into `Length::zero()`.
Also, to reduce the diff in this patch. And once PaddingBorderMargin switches to Au, with `|| Au::zero().into()` the `into()` would become unnecessary but that can be hard to detect. With `Length::zero` the compiler will complain.

atbrakhi:
So, out of these four WPT tests that are failing:
- I guess I will update the test expectation for `/css/CSS2/css1/c414-flt-fit-004.xht` and `/html/obsolete/requirements-for-implementations/the-marquee-element-0/marquee-min-intrinsic-size.html`
- `/css/CSS2/fonts/font-weight-applies-to-005.xht` is passing for me locally. I'm not sure why it is failing here.

Regarding `/css/css-flexbox/flexbox_item-clear.html`,  @Loirooriol pointed out that it might be ok to disable this test temporarily as using `Au` can slightly change the sizes and if flow and flex layout aren't changed simultaneously, the test will fail.
We don't have flexbox enabled by default, so I am hoping there will not be any major impact.

So, I might go ahead and update the expectations! @mrobinson wdyt?

mrobinson:
> So, I might go ahead and update the expectations! @mrobinson wdyt?

Sure this seems reasonable. Thanks!

Loirooriol (approved):

mrobinson (approved):

PR #30903, opened 2023-12-20
Fix the try build when pushing to branches

mrobinson:
Make it so that all try builds go through try.yml and pass
workflow_call arguments as expected to subsequent workflows.

Co-authored-by: Samson <16504129+sagudev@users.noreply.github.com>

---

- [x] These changes do not require tests because they fix a CI issue.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -32,8 +32,6 @@ on:
         required: false
         default: false
         type: boolean
-  push:
-    branches: ["try-windows"]
 
 env:
   RUST_BACKTRACE: 1
@@ -89,7 +87,7 @@ jobs:
       - name: Smoketest
         run: python mach smoketest --angle --${{ inputs.profile }}
       - name: Unit tests
-        if: ${{ inputs.unit-tests || github.ref_name == 'try-windows' }}
+        if: ${{ inputs.unit-tests }}
         uses: nick-fields/retry@v2
         with:
           timeout_minutes: 30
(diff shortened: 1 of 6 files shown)


sagudev:
I will do some testing on my fork but generally looks good.

.github/workflows/windows.yml
sagudev, line 1: This is still run on push to try-windows branch.

mrobinson: Oh! Nice catch. I've removed this as well as one more stray reference to `ref_name` in `linux-wpt.yml`.

sagudev:
try-wpt-mac does not work (probably because try labels also do not support them), but given this was never really used I think we can wait for #30694 to bring it back.

sagudev (approved):

mrobinson:
> try-wpt-mac does not work (probably because try labels also do not support them), but given this was never really used I think we can wait for #30694 to bring it back.

That's true and a bit unfortunate. I will try to do a followup that uses more of your work in #30694 to allow configuring which platforms run tests.

mrobinson:
Thanks very much for the review.

sagudev:
> > try-wpt-mac does not work (probably because try labels also do not support them), but given this was never really used I think we can wait for #30694 to bring it back.
> 
> That's true and a bit unfortunate. I will try to do a followup that uses more of your work in #30694 to allow configuring which platforms run tests.

I plan to split some stuff from #30694 to separate PR after this one land. Mostly wpt -> wpt-sync boolean & wpt to be used from custom wpt test string (to test only specific tests) and flake8 config.

PR #30920, opened 2023-12-23
More powerful try strings & matrix in workflows

sagudev:
Rebased version of https://github.com/servo/servo/pull/30694.

This PR removes try branches and instead provides unified entry point for all try runs (one single `try` branch). This is achieved using empty commits on try branch that have configuration encoded within commit message. This allows creating more complex try runs. Labels are still operational and are calling presets in try_parser.

Another important addition is usage of matrix which allows us to limit number of concurrent runs and fails-fast option.

---

- [x] These changes fix https://github.com/servo/servo/issues/30667
- [x] There are tests for try_parser

### DIFF
--- a/.github/workflows/dispatch-os.yml
+++ b/.github/workflows/dispatch-os.yml
@@ -1,0 +1,48 @@
+name: Dispatch OS
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+      profile:
+        required: true
+        type: string
+      wpt-tests-to-run:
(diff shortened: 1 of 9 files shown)


sagudev:
rebased!

sagudev:
last few pushes are only rebases

sagudev:
Closed in favour of https://github.com/servo/servo/pull/31141. In case we really need more customizable try strings we can revisit advance try_string parser from this PR.

PR #30924, opened 2023-12-23
bhm: Fix a warning in the Linux sampler

### LINKED ISSUE #30911: Warning in background_hang_monitor
When compiling, I get

```
warning: calls to `std::mem::drop` with a reference instead of an owned value does nothing
   --> components/background_hang_monitor/sampler_linux.rs:262:9
    |
262 |         drop(shared_state);
    |         ^^^^^------------^
    |              |
    |              argument has type `&sampler_linux::SharedState`
    |
    = note: use `let _ = ...` to ignore the expression or result
    = note: `#[warn(dropping_references)]` on by default
```

mrobinson:
Use curly braces instead of an explicit drop to control the end of a
borrow.

Fix #30911.

---

### DIFF
--- a/components/background_hang_monitor/sampler_linux.rs
+++ b/components/background_hang_monitor/sampler_linux.rs
@@ -207,59 +207,59 @@ impl Sampler for LinuxSampler {
 
         // Safety: non-exclusive reference only
         // since sampled threads are accessing this concurrently
-        let shared_state = unsafe { &*SHARED_STATE.0.get() };
-        shared_state
-            .msg2
-            .as_ref()
-            .unwrap()
-            .wait_through_intr()
-            .expect("msg2 failed");
+        let result;
+        {
+            let shared_state = unsafe { &*SHARED_STATE.0.get() };
+            shared_state
+                .msg2
+                .as_ref()
+                .unwrap()
+                .wait_through_intr()
+                .expect("msg2 failed");
 
-        let context = CONTEXT.load(Ordering::SeqCst);
-        let mut cursor = mem::MaybeUninit::uninit();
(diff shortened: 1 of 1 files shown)


components/background_hang_monitor/sampler_linux.rs
```diff
@@ -207,59 +207,59 @@ impl Sampler for LinuxSampler {
...
-        // wait for thread to continue.
-        shared_state
-            .msg4
-            .as_ref()
-            .unwrap()
-            .wait_through_intr()
-            .expect("msg4 failed");
-
-        // No-op, but marks the end of the shared borrow
-        drop(shared_state);
```
gterzian, line 262: I think the simpler change is to just remove this line, since we know it's a no-op?

gterzian (approved):
LGTM, with one suggestion.

components/background_hang_monitor/sampler_linux.rs
mrobinson: Won't that cause a deadlock since `clear_shared_state()` also does a `SHARED_STATE.0.get()`?

gterzian: I guess not, because it's an [UnsafeCell](https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html)( I tried in the playground you can call `get` multiple times without deadlock), but, better be safe than sorry and you're version show the intent of "dropping the shared borrow"... Ok merge it as is?

mrobinson: Okay. That makes sense. I'll go ahead and merge.

PR #30927, opened 2023-12-24
Replace time with std::time in `components/devtools`

### LINKED ISSUE #30150: Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate

Taym95:
---

- [x] These changes fix #30150 

### DIFF
--- a/components/devtools/Cargo.toml
+++ b/components/devtools/Cargo.toml
@@ -11,6 +11,7 @@ name = "devtools"
 path = "lib.rs"
 
 [dependencies]
+chrono = "0.4"
 crossbeam-channel = { workspace = true }
 devtools_traits = { workspace = true }
 embedder_traits = { workspace = true }
@@ -24,5 +25,4 @@ serde_json = { workspace = true }
 servo_config = { path = "../config" }
 servo_rand = { path = "../rand" }
 servo_url = { path = "../url" }
-time = { workspace = true }
 uuid = { workspace = true }
--- a/components/devtools/actors/network_event.rs
+++ b/components/devtools/actors/network_event.rs
@@ -7,13 +7,14 @@
 //! Handles interaction with the remote web console on network events (HTTP requests, responses) in Servo.
 
 use std::net::TcpStream;
+use std::time::{SystemTime, UNIX_EPOCH};
 
+use chrono::{Local, LocalResult, TimeZone};
 use devtools_traits::{HttpRequest as DevtoolsHttpRequest, HttpResponse as DevtoolsHttpResponse};
 use headers::{ContentType, Cookie, HeaderMapExt};
 use http::{header, HeaderMap, Method, StatusCode};
 use serde::Serialize;
 use serde_json::{Map, Value};
-use time::Tm;
 
 use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
 use crate::protocol::JsonPacketStream;
@@ -24,7 +25,7 @@ struct HttpRequest {
     method: Method,
     headers: HeaderMap,
     body: Option<Vec<u8>>,
-    startedDateTime: Tm,
+    startedDateTime: SystemTime,
     timeStamp: i64,
     connect_time: u64,
     send_time: u64,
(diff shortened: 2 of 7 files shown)


components/devtools/Cargo.toml
```diff
@@ -11,6 +11,7 @@ name = "devtools"
 path = "lib.rs"
 
 [dependencies]
+chrono = "0.4"
```
mrobinson, line 14: Is it possible to use `std::time` instead of introducing a new dependency in this crate on the chrono crate?

Taym95: I tried, the only missing function is `to_rfc3339 `, can not found it in time and it exists in chrono

mrobinson: Okay. Would it be possible to use `chrono` only for this call to `to_rfc3339`? That way we limit our use of third-party libraries to the minimum and can replace or standardize in the future more easily.

mrobinson: I would only keep `chrono` for `components/devtools/actors/network_event.rs` which is where you need `to_rfc3339` and use `std::time` structs in all other places. I would use something like `Instant` all over the place and only convert to a `chrono` type right before you need the string version of the timestamp.

Taym95: Thanks for suggestion, I used it.

components/devtools/Cargo.toml
```diff
@@ -11,6 +11,7 @@ name = "devtools"
 path = "lib.rs"
 
 [dependencies]
```
mrobinson, line 13: Please remove `time` from the dependencies as well. It looks like there is one more use in `network_event.rs` before that's possible.

components/devtools/actors/network_event.rs
```diff
@@ -370,7 +371,18 @@ impl NetworkEventActor {
...
+                "{}",
+                (Local::now() -
+                    Duration::from_secs(
+                        self.request
+                            .startedDateTime
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap_or_default()
+                            .as_secs()
+                    ))
+                .to_rfc3339()
```
mrobinson, line 384: This doesn't seem like the right way to calculate the local time. You are creating a Local from the start time. You are creating a local time here and subtracting the timestamp of the start time. I think that you should look into `Local::timestamp_opt` in order to create the `Local`. In addition, I think that millisecond accuracy would be a better fit here.

mrobinson (changes requested):
This is looking good. I have just a few issues with the current patch.

components/devtools/actors/network_event.rs
Taym95: Indeed that was not right, thanks for proposing `timestamp_opt`, I used it.

components/devtools/Cargo.toml
Taym95: removed it

components/devtools/actors/network_event.rs
```diff
@@ -366,11 +370,24 @@ impl NetworkEventActor {
...
     pub fn event_actor(&self) -> EventActor {
         // TODO: Send the correct values for startedDateTime, isXHR, private
+        let duration = self
+            .request
+            .startedDateTime
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default();
+
+        let started_datetime_rfc3339 =
+            match Local.timestamp_opt(duration.as_secs() as i64, duration.as_nanos() as u32) {
```
mrobinson, line 380: I think you are now including the duration twice when calculating the local time. Nanoseconds should only include the portion of the time that isn't represented by the seconds part.

Taym95: I see, I can use `timestamp_millis_opt` and provide `millis`, as you mentioned that millisecond accuracy would be a better fit here."

mrobinson (approved):

mrobinson:
Thanks! This looks good now.

PR #30940, opened 2023-12-26
Fix bootstrap for python 3.8

sagudev:
It's crazy, but partial revert of https://github.com/servo/servo/pull/30848 makes nightly builds work again.

If same error would be thrown again in the future and the fix would be to reapply patch we should simply update python and end this pain.

Try CI run: https://github.com/servo/servo/actions/runs/7327178383

---


mrobinson (approved):
It's worth a shot.

PR #30941, opened 2023-12-26
tidy: A few small improvements and fixes

### LINKED ISSUE #30002: `./mach test-tidy` could be faster
`./mach test-tidy` unconditionally runs the WPT manifest lint even when no relevant files have changed. I think this could be skipped by:

1. Moving the lint into the tidy python module
2. Checking the files changed by using the default constructor for `FileList`
3. Only run the WPT manifest lint when files changed include:
  - Any file in `tests/wpt/tests`
  - `tests/wpt/**/MANIFEST.json`

A further refinement would be to run the WPT manifest lint in parallel with the rustfmt check, ensuring that their output is printed separately.

mrobinson:
1. Make the tidy output easier to follow
2. Integrate the WPT manifest cleanliness step into tidy
   itself and don't run it if nothing has changed in the WPT
   directory.
3. Fix an issue where Python test requirements were not installed,
   which could cause issues with some modules not being found.

Fixes #30002.

---

- [x] These changes do not require tests because they just update a tool.

### DIFF
--- a/python/tidy/tidy.py
+++ b/python/tidy/tidy.py
@@ -7,39 +7,34 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import configparser
 import fnmatch
 import glob
+import io
 import itertools
 import json
 import os
 import re
 import subprocess
 import sys
+from typing import List
 
 import colorama
 import toml
+
+import wpt.manifestupdate
 
 from .licenseck import OLD_MPL, MPL, APACHE, COPYRIGHT, licenses_toml, licenses_dep_toml
 
 TOPDIR = os.path.abspath(os.path.dirname(sys.argv[0]))
 WPT_PATH = os.path.join(".", "tests", "wpt")
-SUITES = ["tests", os.path.join("mozilla", "tests")]
-
-
-def wpt_path(*args):
-    return os.path.join(WPT_PATH, *args)
-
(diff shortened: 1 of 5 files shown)


python/tidy/tidy.py
```diff
@@ -986,27 +1015,6 @@ def check_dep_license_errors(filenames, progress=True):
...
-        lint = lint_cls(self.path, self.only_changed_files,
-                        self.exclude_dirs, self.progress,
-                        stylo=self.stylo, no_wpt=self.no_wpt)
-        for error in lint.run():
-            if type(error) is not tuple or (type(error) is tuple and len(error) != 3):
-                yield (self.path, 1, "errors should be a tuple of (path, line, reason)")
-                return
-            yield error
-
     def get_files(self, path, **kwargs):
```
mukilan, line 1018: This method and the below `run` method are members of the `LintRunner` class and can be deleted as well.

mukilan (approved):
Besides the two orphan methods mentioned in the previous comment, this looks good to me. Thanks!

python/tidy/tidy.py
mrobinson: Nice catch! I've removed these.

mrobinson:
Thanks so much for the quick review, @mukilan!

PR #30963, opened 2024-01-01
Unfork and upgrade jemallocator

### LINKED ISSUE #20645: Update jemalloc-sys and unfork jemallocator
In https://github.com/servo/servo/pull/20641 we duplicated ~70 lines from the `jemallocator` crate in order to be able to keep using an order `jemalloc-sys`, because the new one doesn’t build on our Android infra:

`/home/servo/.cargo/registry/src/github.com-1ecc6299db9ec823/jemalloc-sys-0.1.5/jemalloc/src/background_thread.c:83:2: error: unknown type name 'cpu_set_t'`

Upstream issue: https://github.com/jemalloc/jemalloc/issues/1175

mrobinson:
This dependency was forked in ##20641 in order to fix the Android build.
Years have gone by and it's quite likely that many things have changed
in the Android toolchain and these dependencies. We can sort out this
issue when getting the Android build working -- or if all else fails,
disable jemalloc for Android. In the meantime, unfork the dependency and
upgrade it.

Fixes #20645.

---

- [x] These changes fix #20645.

### DIFF
--- a/components/profile/Cargo.toml
+++ b/components/profile/Cargo.toml
@@ -25,4 +25,4 @@ regex = { workspace = true }
 
 [target.'cfg(not(target_os = "windows"))'.dependencies]
 libc = { workspace = true }
-servo_allocator = { path = "../allocator" }
+jemalloc-sys = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1887,12 +1887,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fs_extra"
-version = "1.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
-
-[[package]]
 name = "futf"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


components/profile/Cargo.toml
```diff
@@ -25,4 +25,4 @@ regex = { workspace = true }
 
 [target.'cfg(not(target_os = "windows"))'.dependencies]
 libc = { workspace = true }
-servo_allocator = { path = "../allocator" }
```
sagudev, line 28: Why?

mrobinson: In this case the code is trying to get stats from jemalloc, so it didn't seem to make sense to abstract away the interface into `servo_allocator` and instead get the stats directly from jemalloc itself.

sagudev: But if we ever (and we will likely try) to replace allocator it's easier to have things abstracted.

jdm (approved):

components/profile/Cargo.toml
mrobinson: It feels like in this case, the allocator should expose a higher-level function, since `mallctl` is jemalloc specific. Perhaps we can remove the abstraction now and correct it later when we add an alternative allocator?

sagudev: Sounds good!

sagudev (approved):

PR #30972, opened 2024-01-02
dependencies: Stop shipping GStreamer binaries on Linux

mrobinson:
The binary bundle of GStreamer that we package is not used to compile --
only to run layout tests. It's too old for the APIs that we are using
(as evidenced by needed 1.18 for WebRTC) and nowadays Linux
distributions carry a new version so it's unecessary for our build
machines. No longer using this binary bundle will allow us to upgrade
our GStreamer dependency -- which now has stricter checks that we
are using at least version 1.18.

---

- [x] These changes do not require tests because they just adjust the build configuration.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -48,25 +48,14 @@ jobs:
         with:
           name: release-binary
           path: release-binary
-      - name: Cache libffi
-        id: cache
-        uses: actions/cache@v3
-        with:
-          path: ./libffi6_3.2.1-8_amd64.deb
-          key: cache-libffi
-      - name: Download libffi
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: |
-          wget http://mirrors.kernel.org/ubuntu/pool/main/libf/libffi/libffi6_3.2.1-8_amd64.deb
       - name: unPackage binary
         run: tar -xzf release-binary/target.tar.gz
-      - name: Prep test environment
+      - name: Bootstrap dependencies
         run: |
           python3 -m pip install --upgrade pip
           sudo apt update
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-wpt.yml
```diff
@@ -48,25 +48,13 @@ jobs:
...
-        if: steps.cache.outputs.cache-hit != 'true'
-        run: |
-          wget http://mirrors.kernel.org/ubuntu/pool/main/libf/libffi/libffi6_3.2.1-8_amd64.deb
       - name: unPackage binary
         run: tar -xzf release-binary/target.tar.gz
       - name: Prep test environment
         run: |
           python3 -m pip install --upgrade pip
           sudo apt update
-          sudo apt install -qy --no-install-recommends libgl1 libssl1.1 libdbus-1-3 libxcb-xfixes0-dev libxcb-shape0-dev libunwind8 libgl1-mesa-dri mesa-vulkan-drivers libegl1-mesa
```
sagudev, line 67: I am not 100% sure that all things for webgpu are installed via `mach bootstrap`. It would be nice if you could dispatch workflow with production profile, disabled unit-tests and wpt string: `_webgpu`, to confirm that everything is still working.

mrobinson: If there are more requirements for WebGPU these should move to `./mach bootstrap`.

mrobinson: Unless, I guess, those requirements are hardware-dependent...is that the case here?

mrobinson: I've updated the branch to install the one deb that wasn't installed by `./mach bootstrap`.

mukilan (approved):

mrobinson:
Ultimately this will be landed as part of #30750.

PR #30973, opened 2024-01-02
dependencies: Download Windows GStreamer dependencies from GitHub

mrobinson:
This works around issues where the server is rejecting our requests and
should be more reliable.

---

- [x] These changes do not require tests because this change just updates the build configuration.

### DIFF
--- a/python/servo/platform/windows.py
+++ b/python/servo/platform/windows.py
@@ -22,7 +22,7 @@ DEPENDENCIES = {
     "moztools": "4.0",
 }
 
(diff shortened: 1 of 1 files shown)


mukilan (approved):

mrobinson:
I'm having to split this up into multiple changes due to the upgrade being more complicated than I thought. I'm going to land the switch to using GitHub releases for the dependencies here and move the upgrade changes to a different PR. This should unblock the infrastructure.

PR #30974, opened 2024-01-02
Integrate the `devices` respository

mrobinson:
Despite the name of this dependency, it only handles bluetooth.  Integrating it, allows
changes to be tested more consistently. In addition, it's likely that new bluetooth
libraries will allow removing the majority of the platform-specific code
in this directory.

This is based on the version of this dependency from:
https://github.com/servo/devices/pull/34

---


components/bluetooth/blurmac/Cargo.toml
mukilan, line 1: Is `components/bluetooth/blurmac` a duplicate of `third_party/blurmac`?

mrobinson: 😇 

Whoops. Looks like I managed to check in some code that shouldn't have been part of this PR. I've corrected that. Thanks for catching the mistake.

mukilan (approved):

PR #30987, opened 2024-01-03
Fix crown on NixOS

### LINKED ISSUE #30813: `./mach build` broken on NixOS after moving to `crown`
` ./mach build` seem broken on NixOS after #30508 landed
```
~/dev/servo ‹master› 
$ ./mach build -r
NOTE: Entering nix-shell etc/shell.nix
error: could not execute process `crown -vV` (never executed)

Caused by:
  No such file or directory (os error 2)
Failed in 0:00:00
```

Since NixOS doesn't execute bootstrap commands like other linux distros, crown is not installed by mach.
Adding installation step to `etc/shell.nix` under shellHook also fails because nix considers the path to be impure
i.e 
in etc/shell.nix
```
  shellHook = ''
    CARGO_BUILD_RUSTC=rustc cargo install --path support/crown
  '';
```

results in 
```
  = note: impure path `/home/mukilan/dev/servo/target/release/deps/crown-3d1116308c10f300' used in link
          clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```

I suppose we'll need to import crown as a derivation somehow. Note: installing crown manually using the same command above from inside the nix-shell environment does work and allows subsequent `./mach build` to also work.

delan:
This patch adds crown to the Nix shell environment (etc/shell.nix), and updates the hacking quickstart guide with additional steps to get that version of crown working with rust-analyzer in vscode.

To test this patch, uninstall crown from ~/.cargo/bin, then update your .vscode/settings.json accordingly.

The filterlock approach is a bit more complicated, but it makes the shell start faster and use less disk space:

```
buildRustPackage with allowBuiltinFetchGit

$ time nix-shell etc/shell.nix --run crown
17.16s user 7.66s system 28% cpu 1:27.45 total

$ sudo nix-collect-garbage -d
note: currently hard linking saves 1115.24 MiB
2015 store paths deleted, 4125.29 MiB freed
```

```
buildRustPackage with filterlock

$ time nix-shell etc/shell.nix --run crown
1.73s user 3.27s system 12% cpu 41.262 total

$ sudo nix-collect-garbage -d
note: currently hard linking saves 1115.24 MiB
316 store paths deleted, 2981.63 MiB freed
```

---

- [x] These changes fix #30813

- [x] These changes do not require tests because the build would fail on NixOS without them


etc/shell.nix
```diff
@@ -3,10 +3,15 @@
 
 with import <nixpkgs> {};
 let
-    pinnedSha = "6adf48f53d819a7b6e15672817fa1e78e5f4e84f";
     pinnedNixpkgs = import (builtins.fetchTarball {
-        url = "https://github.com/NixOS/nixpkgs/archive/${pinnedSha}.tar.gz";
+      url = "https://github.com/NixOS/nixpkgs/archive/6adf48f53d819a7b6e15672817fa1e78e5f4e84f.tar.gz";
     }) {};
+
+    # Keep this in sync with rust-toolchain.toml!
```
mukilan, line 10: Since rust-toolchain.toml supports comments, maybe it will be helpful to leave a comment about updating shell.nix there as well.

support/filterlock/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+
+[package]
+name = "filterlock"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+serde = { version = "1.0.194", features = ["derive"] }
+toml = { version = "0.8.8", features = ["preserve_order"] }
```
mrobinson, line 10: Would it be possible to describe what this tool does and explain that it's only used for nix here as well as adding a license.

etc/shell.nix
```diff
@@ -33,6 +38,63 @@ clangStdenv.mkDerivation rec {
...
+      name = "crown";
+      src = ../support/crown;
+      stdenv = nixpkgs_rust_1_74.clangStdenv;
+      buildInputs = [ nixpkgs_rust_1_74.llvmPackages.libllvm ];
+      doCheck = false;
+      cargoLock = {
+        lockFileContents = builtins.readFile filteredLockFile;
+
+        # Needed when filteredLockFile = ../Cargo.lock.
+        # allowBuiltinFetchGit = true;
```
mukilan, line 81: I am not sure I understand in which scenario we will want this to be uncommented. Maybe the comment above can be expanded a bit?

support/filterlock/src/main.rs
```diff
@@ -0,0 +1,70 @@
+use std::{fs::File, env::{args_os, args}, io::Read, collections::BTreeSet};
```
mrobinson, line 1: Missing a license here, I think, as well.

mrobinson (changes requested):
Drive-by review for a couple small things.

mukilan (approved):
Great work!

etc/shell.nix
delan: done in afd1254c3cf56cd0534e1aebebc6ee79c4499d6d

etc/shell.nix
delan: done in dca7c89526e3c66b09e4a72b0fddb0990706214b

support/filterlock/Cargo.toml
delan: done in bd49d8f39ca58db6f618ed59fa810b99df392bc6 + 6f76a10fca2619f66182077031bc3430a1a4ba18

support/filterlock/src/main.rs
delan: done in 6f76a10fca2619f66182077031bc3430a1a4ba18

mrobinson (approved):

Issue #30915, opened 2023-12-22
webgpureport.org crashes servo


sagudev:
### Discussed in https://github.com/servo/servo/discussions/30914

<div type='discussions-op-text'>

<sup>Originally posted by **beaufortfrancois** December 22, 2023</sup>
When running Servo 0.0.1-5e8f70f with `--pref dom.webgpu.enabled`, it fails to render https://webgpureport.org/

![image](https://github.com/servo/servo/assets/634478/bcdfd461-5a7b-47bb-92b5-14ce6b5f496c)

Without the pref, it loads and simply says WebGPU is not supported.

![image](https://github.com/servo/servo/assets/634478/0c564e57-905c-4d58-9055-fa5d7467319b)

Is this a known issue?
</div>

sagudev:
`await navigator.gpu.requestAdapter({ forceFallbackAdapter: true })` makes it work.

Crash line is:
https://github.com/servo/servo/blob/1f0f50b22bb43e965febf31f5a436113f2958725/components/script/script_thread.rs#L1498

beaufortfrancois:
I'm using macOS. I wonder if it crashes if the system doesn't have a fallback adapter, which is the case on my macBook machine.

sagudev:
> I'm using macOS. I wonder if it crashes if the system doesn't have a fallback adapter, which is the case on my macBook machine.

But forcing fallback adapter helps? Could you try `wgpu-info` (you may need to install it with `cargo install wgpu-info`)

beaufortfrancois:
It's actually **removing** `await navigator.gpu.requestAdapter({ forceFallbackAdapter: true })` that makes it work.

> `await navigator.gpu.requestAdapter({ forceFallbackAdapter: true })` makes it work.

beaufortfrancois:
> Could you try `wgpu-info` (you may need to install it with `cargo install wgpu-info`)

```
 ~/.cargo/bin/wgpu-info 
Adapter 0:
	   Backend: Metal
	      Name: "Apple M1 Pro"
	  VendorID: 0
	  DeviceID: 0
	      Type: IntegratedGpu
	    Driver: ""
	DriverInfo: ""
	 Compliant: true
	Features:
		                                           DEPTH_CLIP_CONTROL: true
		                                              TIMESTAMP_QUERY: true
		                                      INDIRECT_FIRST_INSTANCE: true
		                                                   SHADER_F16: true
		                                           BGRA8UNORM_STORAGE: true
[...]

PR #30989, opened 2024-01-03
WebGPU: always send exit message to script before exiting

gterzian:
When WebGPU exits, including when it internally encounters an error, always first send the `Exit` message to script before exiting. 

---

- [x] These changes fix https://github.com/servo/servo/issues/30915

### DIFF
--- a/components/webgpu/lib.rs
+++ b/components/webgpu/lib.rs
@@ -831,13 +831,10 @@ impl<'a> WGPU<'a> {
                         gfx_select!(texture => global.texture_drop(texture, true));
                     },
                     WebGPURequest::Exit(sender) => {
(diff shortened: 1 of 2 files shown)


sagudev (commented):
LGTM, but I would like to wait for webgpu run: https://github.com/sagudev/servo/actions/runs/7398389520

sagudev:
New run for updated expectations: https://github.com/sagudev/servo/actions/runs/7398819825

sagudev (approved):
Given that https://github.com/servo/servo/issues/30999, I think the real changes are captured in updates expectations (mostly CRASH -> TIMEOUT).

gterzian:
Thanks for the review and test update! @sagudev

Issue #30889, opened 2023-12-19
WebIDL impl: Replace use of NonNull<JSObject> 


gterzian:
Part of https://github.com/servo/servo/issues/30862

Where appropriate, we should remove the direct use of unsafe `NonNull<JSObject>`,  and replace it with use of the appropriate safe wrapper found in [`/mozjs/src/typedarray.rs`](https://github.com/servo/mozjs/blob/de842d84a07eabac4e1bfe72a954b74ee0244b82/mozjs/src/typedarray.rs).

Example: AudioBuffer's [`getChannelData`](https://webaudio.github.io/web-audio-api/#dom-audiobuffer-getchanneldata)

WedIDL says to [return a `Float32Array`](https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/webidls/AudioBuffer.webidl#L22C12), but implementation returns a `NonNull<JSObject>`:

https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/audiobuffer.rs#L265

The return value of the WebIDL method is backed by a [`Heap<*mut JSObject>`](https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/audiobuffer.rs#L34) on the dom struct. 

Solution: `use js::typedarray::Float32Array` instead of `js::jsapi::JSObject`.

Bonus: wrap the use of `js::typedarray` inside something at `dom/bindings/`. (maybe see `bindings/iterable.rs` for an example)

------------------------------

How to do it?

1. Update codegen(look also at [third_party/webidl](https://github.com/servo/mozjs/blob/de842d84a07eabac4e1bfe72a954b74ee0244b82/mozjs/src/typedarray.rs))

codegen starts at(probably needs other updates): https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/bindings/codegen/CodegenRust.py#L1463 

2. Make sure that the right `_Methods` are generated. 
3. Update the Rust code, where the struct would the typed array directly, and implement the WebIDL methods returning the typed array. 
[...]

PR #30990, opened 2024-01-03
WebIDL: use FLoat32Array 

gterzian:
First part of https://github.com/servo/servo/issues/30889: applying to `Float32Array`. 

---

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -119,7 +119,8 @@ builtinNames = {
     IDLType.Tags.unrestricted_float: 'f32',
     IDLType.Tags.float: 'Finite<f32>',
     IDLType.Tags.unrestricted_double: 'f64',
-    IDLType.Tags.double: 'Finite<f64>'
+    IDLType.Tags.double: 'Finite<f64>',
+    IDLBuiltinType.Tags.float32array: 'Float32Array'
 }
 
 numericTags = [
@@ -1465,6 +1466,11 @@ def getRetvalDeclarationForType(returnType, descriptorProvider):
         # Nothing to declare
         return CGGeneric("()")
     if returnType.isPrimitive() and returnType.tag() in builtinNames:
+        result = CGGeneric(builtinNames[returnType.tag()])
+        if returnType.nullable():
+            result = CGWrapper(result, pre="Option<", post=">")
+        return result
+    if returnType.isTypedArray() and returnType.tag() in builtinNames and descriptorProvider.interface.identifier.name == 'AudioBuffer':
         result = CGGeneric(builtinNames[returnType.tag()])
         if returnType.nullable():
             result = CGWrapper(result, pre="Option<", post=">")
@@ -6494,6 +6500,7 @@ def generate_imports(config, cgthings, descriptors, callbacks=None, dictionaries
         'js::rust::define_properties',
         'js::rust::get_object_class',
         'js::typedarray',
+        'js::typedarray::Float32Array',
         'crate::dom',
         'crate::dom::bindings',
         'crate::dom::bindings::codegen::InterfaceObjectMap',
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -120,7 +120,7 @@ builtinNames = {
     IDLType.Tags.float: 'Finite<f32>',
     IDLType.Tags.unrestricted_double: 'f64',
     IDLType.Tags.double: 'Finite<f64>',
-    IDLBuiltinType.Tags.float32array: 'Float32Array'
+    IDLType.Tags.float32array: 'Float32Array'
 }
 
 numericTags = [
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -1462,21 +1462,21 @@ def getConversionConfigForType(type, isEnforceRange, isClamp, treatNullAs):
 
 TODO_SWITCH_TO_FLOAT_32 = ['DOMMatrixReadOnly', 'XRView', 'XRRigidTransform', 'XRRay', 'GamepadPose']
 
+def builtin_return_type(returnType):
+    result = CGGeneric(builtinNames[returnType.tag()])
+    if returnType.nullable():
+        result = CGWrapper(result, pre="Option<", post=">")
+    return result
+
 # Returns a CGThing containing the type of the return value.
 def getRetvalDeclarationForType(returnType, descriptorProvider):
     if returnType is None or returnType.isUndefined():
         # Nothing to declare
         return CGGeneric("()")
     if returnType.isPrimitive() and returnType.tag() in builtinNames:
-        result = CGGeneric(builtinNames[returnType.tag()])
-        if returnType.nullable():
-            result = CGWrapper(result, pre="Option<", post=">")
-        return result
+        return builtin_return_type(returnType)
     if returnType.isTypedArray() and returnType.tag() in builtinNames and (not descriptorProvider.interface.identifier.name in TODO_SWITCH_TO_FLOAT_32):
-        result = CGGeneric(builtinNames[returnType.tag()])
-        if returnType.nullable():
-            result = CGWrapper(result, pre="Option<", post=">")
-        return result
+        return builtin_return_type(returnType)
     if returnType.isDOMString():
         result = CGGeneric("DOMString")
         if returnType.nullable():
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -1460,7 +1460,7 @@ def getConversionConfigForType(type, isEnforceRange, isClamp, treatNullAs):
     return "()"
 
 
-TODO_SWITCH_TO_FLOAT_32 = ['DOMMatrixReadOnly', 'XRView', 'XRRigidTransform', 'XRRay', 'GamepadPose']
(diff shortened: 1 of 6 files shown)


gterzian:
@jdm Here is a prototype(that compiles), could you please tell me what you think?

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1469,6 +1470,11 @@ def getRetvalDeclarationForType(returnType, descriptorProvider):
         if returnType.nullable():
             result = CGWrapper(result, pre="Option<", post=">")
         return result
+    if returnType.isTypedArray() and returnType.tag() in builtinNames and descriptorProvider.interface.identifier.name == 'AudioBuffer':
+        result = CGGeneric(builtinNames[returnType.tag()])
+        if returnType.nullable():
+            result = CGWrapper(result, pre="Option<", post=">")
+        return result
```
mrobinson, line 1477: A couple comments here:
 1. Is the body of this `if` identical to the one above? Can it be merged?
 2. What's the explanation for the special-case for `AudioBuffer`? If this is necessary could it be moved to a separate array of names that need this with some documentation on the array?

gterzian: 1. Which one do you mean? 
2. Yes that's for now so that this only affects audio buffer and I can get a first try to compile quickly. I'm thinking to add a list of affected WebIDL interfaces, so that the work can be broken into various PR's and issues(which could be marked as easy), where each one would tick off an interface.

mrobinson: It looks like the code starting at line 1469 is identical to the code starting at line 1474?

gterzian: The one above is [isPrimitive](https://github.com/servo/servo/blob/28adcf9f6b43e8051f80356ee8583b09359bf62f/third_party/WebIDL/WebIDL.py#L3795), which covers the ones listed up to `double` at https://github.com/servo/servo/blob/28adcf9f6b43e8051f80356ee8583b09359bf62f/third_party/WebIDL/WebIDL.py#L3588

mrobinson: The body of the `isPrimitive()` check is:

```
        result = CGGeneric(builtinNames[returnType.tag()])
        if returnType.nullable():
            result = CGWrapper(result, pre="Option<", post=">")
        return result
```

The body of the new if condition you've added is:

```
        result = CGGeneric(builtinNames[returnType.tag()])
        if returnType.nullable():
            result = CGWrapper(result, pre="Option<", post=">")
        return result
```

These two are the same. It seems like this duplication could be avoided.

gterzian: Ok I see, thanks for pointing it out.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct Float32ArrayOnHeap {
+    internal: Heap<*mut JSObject>,
+}
+
+impl Float32ArrayOnHeap {
+    pub fn default() -> Self {
+        Float32ArrayOnHeap {
+            internal: Heap::default(),
+        }
+    }
```
jdm, line 24: This could be replaced with a `derive(Heap)`, right?

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use js::jsapi::{Heap, JSObject, JS_GetArrayBufferViewBuffer};
+use js::rust::wrappers::DetachArrayBuffer;
+use js::rust::CustomAutoRooterGuard;
+use js::typedarray::{CreateWith, Float32Array};
+
+use crate::script_runtime::JSContext;
+
+#[derive(JSTraceable)]
+pub struct Float32ArrayOnHeap {
```
jdm, line 15: nit: perhaps `HeapFloat32Array` instead?

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        if res.is_err() {
+            return Err(());
+        }
+
+        self.internal.set(*array);
+        Ok(())
+    }
+
+    #[allow(unsafe_code)]
```
jdm, line 41: We should probably just have a single `#![allow(unsafe_code)]` at the top of the file.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+        let mut array = array.unwrap();
+        unsafe {
+            let slice = (*array).as_mut_slice();
+            let (_, dest) = slice.split_at_mut(start);
+            dest[0..end].copy_from_slice(&source.as_slice()[0..end])
+        }
+        Ok(())
+    }
+
+    pub fn is_set(&self) -> bool {
```
jdm, line 105: nit: `is_initialized`, perhaps?

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+        source: CustomAutoRooterGuard<Float32Array>,
+        start: usize,
+        end: usize,
+    ) -> Result<(), ()> {
+        assert!(self.is_set());
+        typedarray!(in(*cx) let mut array: Float32Array = self.internal.get());
+        if array.is_err() {
+            return Err(());
+        }
+        let mut array = array.unwrap();
```
jdm, line 96: `let Ok(mut array) = array else { return Err(()) };`

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+        dest: &mut [f32],
+        start: usize,
+        end: usize,
+    ) -> Result<(), ()> {
+        assert!(self.is_set());
+        typedarray!(in(*cx) let array: Float32Array = self.internal.get());
+        if array.is_err() {
+            return Err(());
+        }
+        let array = array.unwrap();
```
jdm, line 75: `let Ok(mut array) = array else { return Err(()) };`

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+        unsafe {
+            let slice = (*array).as_slice();
+            dest.clone_from_slice(&slice[start..end]);
```
jdm, line 78: This can be copy_from_slice, right?

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,112 @@
...
+        Ok(())
+    }
+
+    #[allow(unsafe_code)]
+    pub fn copy_data_from(
+        &self,
+        cx: JSContext,
+        source: CustomAutoRooterGuard<Float32Array>,
+        start: usize,
+        end: usize,
```
jdm, line 89: These arguments surprised me, because start is only used on the destination array, not on the source array.

jdm (approved):
The actual AudioBuffer code feels more clear with these changes, so I'm a fan!

components/script/dom/bindings/typedarrays.rs
gterzian: I think `end` could be removed if `AudioBuffer::length` was passed to the typed array on initialization instead of just calling `default`, but since the length is a constraint that is part of the web API, perhaps it should be left as is? Is that what you mean?

jdm: My confusion is that as an API for interacting with two typed arrays, there's no indication that start is only used for the destination. If there were dest_start and length arguments, that would be more clear.

gterzian: Ok I see, I'll rename those as such then.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -119,7 +119,8 @@ def wrapInNativeContainerType(type, inner):
     IDLType.Tags.unrestricted_float: 'f32',
     IDLType.Tags.float: 'Finite<f32>',
     IDLType.Tags.unrestricted_double: 'f64',
-    IDLType.Tags.double: 'Finite<f64>'
+    IDLType.Tags.double: 'Finite<f64>',
+    IDLBuiltinType.Tags.float32array: 'Float32Array'
```
gterzian: fixed this to use `IDLType`

gterzian:
@jdm Could you please review the last commit also? I realized the dom matrix used the array without a backing heap, so added a util func and also used it iniside `HeapFloat32Array`. Not 100% sure of the rooting safety. 

By the way I am aware of https://github.com/servo/servo/blob/355567186400d0531657d0eedda7625de9d0e93a/components/script/dom/bindings/utils.rs#L137, but it assumes a heap and I think it should remove once `HeapFloat32Array` is used everywhere else.

jdm (approved):
Makes sense and should be safely rooted!

components/script/dom/bindings/typedarrays.rs
```diff
@@ -0,0 +1,105 @@
...
+    if res.is_err() {
+        Err(())
+    } else {
+        Float32Array::from(dest.get())
+    }
+}
+
+#[derive(Default, JSTraceable)]
+pub struct HeapFloat32Array {
+    internal: Heap<*mut JSObject>,
```
jdm, line 31: I just noticed that the existing audio buffer code is unsafe, because storing Heapin a Vec means the values can be moved in memory when the vec resizes. Let's make this a `Box<Heap<...>>` instead so it's impossible to misuse?

jdm: https://doc.servo.org/mozjs_sys/jsgc/struct.Heap.html#method.boxed

gterzian: Good catch!

PR #31002, opened 2024-01-04
Fix failed request for adapter when not available

### LINKED ISSUE #30895: WebGPU: fail request for adapter when not available.
Depends on https://github.com/servo/servo/pull/30888

See the changes at https://github.com/servo/servo/pull/30888/files/4bec9ef169f51c91f616b854e231484bbd983366#diff-95aa2eb890fe794918afdadfb6a89e026ec80222ceb8e5a21768d528e9ce036cR2066

This will simply leave the async handler in script waiting forever, instead we should send a message back and properly handle the case where WebGPU is not available(re "properly", look at the [spec](https://gpuweb.github.io/gpuweb/#dom-gpu-requestadapter) .

See https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/gpu.rs#L108

Taym95:
---

- [x] These changes fix #30895

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -2063,7 +2063,11 @@ where
         };
         match request {
             FromScriptMsg::RequestAdapter(response_sender, options, ids) => match webgpu_chan {
-                None => warn!("Failed to send request adapter message, missing WebGPU channel"),
+                None => {
+                    if let Err(e) = response_sender.send(None) {
+                        return warn!("Failed to send request adapter message: {}", e);
+                    }
+                },
                 Some(webgpu_chan) => {
                     let adapter_request = WebGPURequest::RequestAdapter {
                         sender: response_sender,
--- a/components/script/dom/gpu.rs
+++ b/components/script/dom/gpu.rs
@@ -45,7 +45,7 @@ impl GPU {
 }
 
 pub trait AsyncWGPUListener {
-    fn handle_response(&self, response: WebGPUResponseResult, promise: &Rc<Promise>);
+    fn handle_response(&self, response: Option<WebGPUResponseResult>, promise: &Rc<Promise>);
 }
 
 struct WGPUResponse<T: AsyncWGPUListener + DomObject> {
@@ -55,7 +55,7 @@ struct WGPUResponse<T: AsyncWGPUListener + DomObject> {
 
 impl<T: AsyncWGPUListener + DomObject> WGPUResponse<T> {
     #[allow(crown::unrooted_must_root)]
-    fn response(self, response: WebGPUResponseResult) {
+    fn response(self, response: Option<WebGPUResponseResult>) {
         let promise = self.trusted.root();
         self.receiver.root().handle_response(response, &promise);
     }
@@ -64,13 +64,13 @@ impl<T: AsyncWGPUListener + DomObject> WGPUResponse<T> {
 pub fn response_async<T: AsyncWGPUListener + DomObject + 'static>(
     promise: &Rc<Promise>,
     receiver: &T,
-) -> IpcSender<WebGPUResponseResult> {
+) -> IpcSender<Option<WebGPUResponseResult>> {
     let (action_sender, action_receiver) = ipc::channel().unwrap();
     let task_source = receiver.global().dom_manipulation_task_source();
     let canceller = receiver
         .global()
         .task_canceller(TaskSourceName::DOMManipulation);
-    let mut trusted = Some(TrustedPromise::new(promise.clone()));
+    let mut trusted: Option<TrustedPromise> = Some(TrustedPromise::new(promise.clone()));
     let trusted_receiver = Trusted::new(receiver);
     ROUTER.add_route(
         action_receiver.to_opaque(),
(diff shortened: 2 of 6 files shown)


components/script/dom/gpu.rs
```diff
@@ -139,39 +139,41 @@ impl GPUMethods for GPU {
...
+                        Heap::default(),
+                        features,
+                        limits,
+                        adapter_info,
+                        adapter_id,
+                    );
+                    promise.resolve_native(&adapter);
+                },
+                Err(e) => {
+                    unreachable!("Failed to get a response from the constellation ({})", e)
```
gterzian, line 169: So I was wrong about this being unreachable, because here we are handling the "inner" `WebGPUResponseResult`, not the receiving of the message containing this result, which is done at https://github.com/servo/servo/blob/c219204084b8f8a747ebd37ec75472c1b7e97411/components/script/dom/gpu.rs#L77. 

Conclusion: this branch should remain ```               
 warn!("Could not get GPUAdapter ({:?})", e);
                promise.resolve_native(&None::<GPUAdapter>);```

and it's the `None` branch that should ```warn!("Couldn't get a response, because WebGPU is disabled ({})", e); promise.resolve_native(&None::<GPUAdapter>);```

components/webgpu/lib.rs
```diff
@@ -460,8 +460,8 @@ impl<'a> WGPU<'a> {
                         };
                         let global = &self.global;
                         let result = gfx_select!(buffer_id => global.buffer_map_async(buffer_id, map_range, operation));
-                        if let Err(ref e) = result {
-                            if let Err(w) = sender.send(Err(format!("{:?}", e))) {
```
gterzian, line 464: All errors in this file should be `Some(Err`, because the `None` case is only handled at the constellation level when WebGPU is not available, while here we are in the WebGPU backend.

components/script/dom/gpuadapter.rs
```diff
@@ -237,35 +237,37 @@ impl GPUAdapterMethods for GPUAdapter {
...
+                },
+                Err(e) => {
+                    unreachable!("Failed to get a response from the constellation ({})", e)
+                },
+                _ => {
+                    warn!("GPUAdapter received wrong WebGPUResponse");
+                    promise.reject_error(Error::Operation);
+                },
             },
+            None => promise.resolve_native(&None::<GPUDevice>),
```
gterzian, line 270: Here the `None` should be `unreachable`, because it's only the `RequestAdapter` message that can return `None`, all other request imply an initial adapter was acquired, therefore WebGPU is available and `None` should never be subsequently returned.

components/script/dom/gpubuffer.rs
```diff
@@ -346,27 +346,29 @@ impl GPUBufferMethods for GPUBuffer {
...
+                        .borrow_mut()
+                        .as_mut()
+                        .unwrap()
+                        .mapping
+                        .borrow_mut() = bytes.to_vec();
+                    promise.resolve_native(&());
+                    self.state.set(GPUBufferState::Mapped);
+                },
+                Err(e) => {
+                    unreachable!("Failed to get a response from the constellation ({})", e)
```
gterzian, line 364: This error should be left as before, because this is the internal error from webgpu.

components/script/dom/gpubuffer.rs
```diff
@@ -346,27 +346,29 @@ impl GPUBufferMethods for GPUBuffer {
...
+                },
+                _ => {
+                    warn!("GPUBuffer received wrong WebGPUResponse");
+                    promise.reject_error(Error::Operation);
+                },
             },
+            None => promise.resolve_native(&None::<GPUBuffer>),
```
gterzian, line 371: Unreachable(as above).

components/script/dom/gpuadapter.rs
```diff
@@ -237,35 +237,37 @@ impl GPUAdapterMethods for GPUAdapter {
...
+                        descriptor.limits,
+                        device_id,
+                        queue_id,
+                        descriptor.label.unwrap_or_default(),
+                    );
+                    self.global().add_gpu_device(&device);
+                    promise.resolve_native(&device);
+                },
+                Err(e) => {
+                    unreachable!("Failed to get a response from the constellation ({})", e)
```
gterzian, line 263: error should be left as before, because it's the internal error from webgpu.

gterzian (changes requested):
Thanks for your contribution!

The messaging is actually quite complicated, and it's only the initial `RequestAdapter` that can return `None` if webgpu is unavailable, but because of the way the ipc sender is passed through to webgpu, all others still need to return an option...

components/script/dom/gpu.rs
Taym95: Got it, thanks for explaining, I addressed other comments.

components/constellation/constellation.rs
```diff
@@ -2063,7 +2063,11 @@ where
         };
         match request {
             FromScriptMsg::RequestAdapter(response_sender, options, ids) => match webgpu_chan {
-                None => warn!("Failed to send request adapter message, missing WebGPU channel"),
+                None => {
+                    if let Err(e) = response_sender.send(None) {
+                        return warn!("Failed to send request adapter message: {}", e);
```
sagudev, line 2068: Does warn return `()`? If so can we keep return and warn in separate statements.

Taym95: Yes it returns `()`, I see that same approach is used in this file e.g:

https://github.com/servo/servo/blob/main/components/constellation/constellation.rs#L2027

sagudev: Then I guess it's alright.

gterzian (changes requested):
LGTM, with one nit. 

I think we may have to also update tests expectations, I  assume some `TIMEOUT` will change to `FAIL` or `PASS`. @sagudev other than pointing to https://github.com/servo/servo/blob/main/tests/wpt/README.md#updating-test-expectations, are there other instructions on how to do it? I forgot how :)

components/script/dom/gpu.rs
```diff
@@ -139,39 +139,45 @@ impl GPUMethods for GPU {
...
+                        adapter_info,
+                        adapter_id,
+                    );
+                    promise.resolve_native(&adapter);
+                },
+                Err(e) => {
+                    warn!("Could not get GPUAdapter ({:?})", e);
+                    promise.resolve_native(&None::<GPUAdapter>);
+                },
+                _ => {
```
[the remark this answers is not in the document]
Taym95: Indeed it is unreachable, I will remove `promise.reject_error` and `warn` and use `unreachable`

sagudev:
try run: https://github.com/sagudev/servo/actions/runs/7445674351

sagudev:
> TIMEOUT [expected OK] /html/browsers/browsing-the-web/navigating-across-documents/about-srcdoc-navigation-blocked.window.html

Is very flaky lately, but it's unrelated to this PR.

sagudev (changes requested):
There is a lot of `CRASH [expected OK]` with `Failed to get a response for BufferMapAsync (thread Script(1,1), at components/script/dom/gpubuffer.rs:369)`

gterzian:
> There is a lot of CRASH [expected OK] with Failed to get a response for BufferMapAsync (thread Script(1,1), at components/script/dom/gpubuffer.rs:369)

I think there you need to send the option: https://github.com/servo/servo/blob/2d859ac93f888da182e8d4c9df38417a038a023a/components/webgpu/lib.rs#L445

It compiles only because of the casting here https://github.com/servo/servo/blob/2d859ac93f888da182e8d4c9df38417a038a023a/components/webgpu/lib.rs#L426

which should be changed to an option as well.

sagudev:
new try run for webgpu: https://github.com/sagudev/servo/actions/runs/7457368386

gterzian (approved):
LGTM. @sagudev I think we're good on the tests, right?

sagudev (approved):
Webgpu try run looks good.

PR #31003, opened 2024-01-04
bootstrap: Print more output when installing dependencies

mrobinson:
It's often the case (especially with the taplo installation and on
Windows) that bootstrap is doing lots of stuff in the background for a
long amount of time. Without output it's hard to tell what exactly is
going on. This change adds more output to this process as well as
removing some Python 2.x era code.

---

- [x] These changes do not require tests because this just modifies bootstrap.

### DIFF
--- a/python/mach_bootstrap.py
+++ b/python/mach_bootstrap.py
@@ -5,11 +5,8 @@
 import os
 import platform
 import site
-import shutil
+import subprocess
 import sys
-
-from subprocess import Popen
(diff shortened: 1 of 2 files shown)


mukilan (approved):

mrobinson:
The Windows CI does not seem to like utf-8 in the output, so I've removed the emoji. The change is now less fun, but actually runs on Windows GitHub actions.

PR #31015, opened 2024-01-06
Fix installation instructions for Python on Debian-like distributions in README

nnabeyang:
This PR fixes the installation of Python on Debian-like distributions. Specifically, it appears that the installation of python3-env is missing, so I have added it.

---

- [x] These changes do not require tests because they adjust build instructions


mrobinson (approved):
Thanks!

PR #31020, opened 2024-01-07
Replace time with std::time in components/metrics & components/shared

### LINKED ISSUE #30150: Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate

Taym95:
Replace time with std::time in components/metrics & components/shared.

---

- [x] These changes are related to #30150

### DIFF
--- a/components/malloc_size_of/lib.rs
+++ b/components/malloc_size_of/lib.rs
@@ -892,6 +892,10 @@ impl MallocSizeOf for xml5ever::QualName {
 malloc_size_of_is_0!(time::Duration);
 #[cfg(feature = "servo")]
 malloc_size_of_is_0!(time::Tm);
+#[cfg(feature = "servo")]
+malloc_size_of_is_0!(std::time::Duration);
+#[cfg(feature = "servo")]
+malloc_size_of_is_0!(std::time::SystemTime);
 
 #[cfg(feature = "servo")]
 impl<T> MallocSizeOf for hyper_serde::Serde<T>
--- a/components/metrics/lib.rs
+++ b/components/metrics/lib.rs
@@ -5,6 +5,7 @@
 use std::cell::{Cell, RefCell};
 use std::cmp::Ordering;
 use std::collections::HashMap;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use gfx_traits::Epoch;
 use ipc_channel::ipc::IpcSender;
@@ -15,7 +16,6 @@ use profile_traits::time::{send_profile_data, ProfilerCategory, ProfilerChan, Ti
 use script_traits::{ConstellationControlMsg, LayoutMsg, ProgressiveWebMetricType};
 use servo_config::opts;
 use servo_url::ServoUrl;
-use time::precise_time_ns;
 
 pub trait ProfilerMetadataFactory {
     fn new_metadata(&self) -> Option<TimerMetadata>;
@@ -32,8 +32,8 @@ pub trait ProgressiveWebMetric {
 /// TODO make this configurable
 /// maximum task time is 50ms (in ns)
 pub const MAX_TASK_NS: u64 = 50000000;
-/// 10 second window (in ns)
-const INTERACTIVE_WINDOW_SECONDS_IN_NS: u64 = 10000000000;
+/// 10 second window
+const INTERACTIVE_WINDOW_SECONDS: Duration = Duration::from_secs(10);
 
 pub trait ToMs<T> {
     fn to_ms(&self) -> T;
@@ -63,7 +63,10 @@ fn set_metric<U: ProgressiveWebMetric>(
     };
     let now = match metric_time {
         Some(time) => time,
-        None => precise_time_ns(),
+        None => SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos() as u64,
     };
     let time = now - navigation_start;
     attr.set(Some(time));
@@ -113,13 +116,13 @@ pub struct InteractiveMetrics {
 
 #[derive(Clone, Copy, Debug, MallocSizeOf)]
 pub struct InteractiveWindow {
-    start: u64,
+    start: SystemTime,
 }
 
 impl InteractiveWindow {
     pub fn new() -> InteractiveWindow {
         InteractiveWindow {
-            start: precise_time_ns(),
+            start: SystemTime::now(),
         }
     }
 
@@ -128,16 +131,22 @@ impl InteractiveWindow {
     //   restart: there was a task > 50ms
     //   not all documents are interactive
     pub fn start_window(&mut self) {
-        self.start = precise_time_ns();
+        self.start = SystemTime::now();
     }
 
     /// check if 10s has elapsed since start
     pub fn needs_check(&self) -> bool {
-        precise_time_ns() - self.start >= INTERACTIVE_WINDOW_SECONDS_IN_NS
+        SystemTime::now()
+            .duration_since(self.start)
+            .unwrap_or_default() >=
+            INTERACTIVE_WINDOW_SECONDS
     }
 
     pub fn get_start(&self) -> u64 {
         self.start
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos() as u64
     }
 }
 
@@ -161,7 +170,12 @@ impl InteractiveMetrics {
 
     pub fn set_dom_content_loaded(&self) {
         if self.dom_content_loaded.get().is_none() {
-            self.dom_content_loaded.set(Some(precise_time_ns()));
+            self.dom_content_loaded.set(Some(
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_nanos() as u64,
+            ));
         }
     }
 
--- a/components/script/dom/performanceresourcetiming.rs
+++ b/components/script/dom/performanceresourcetiming.rs
@@ -113,7 +113,7 @@ impl PerformanceResourceTiming {
                 DOMString::from(url.into_string()),
                 DOMString::from("resource"),
                 resource_timing.start_time as f64,
-                resource_timing.response_end as f64 - resource_timing.start_time as f64,
+                (resource_timing.response_end - resource_timing.start_time) as f64,
             ),
             initiator_type: initiator_type,
             next_hop: next_hop,
--- a/components/shared/net/lib.rs
+++ b/components/shared/net/lib.rs
@@ -4,6 +4,8 @@
 
 #![deny(unsafe_code)]
 
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use cookie::Cookie;
 use headers::{ContentType, HeaderMapExt, ReferrerPolicy as ReferrerPolicyHeader};
 use http::{Error as HttpError, HeaderMap, StatusCode};
@@ -18,11 +20,11 @@ use malloc_size_of::malloc_size_of_is_0;
 use malloc_size_of_derive::MallocSizeOf;
 use mime::Mime;
 use msg::constellation_msg::HistoryStateId;
+use num_traits::Zero;
 use rustls::Certificate;
 use serde::{Deserialize, Serialize};
 use servo_rand::RngCore;
 use servo_url::{ImmutableOrigin, ServoUrl};
-use time::precise_time_ns;
 use webrender_api::{ImageData, ImageDescriptor, ImageKey};
 
 use crate::filemanager_thread::FileManagerThreadMsg;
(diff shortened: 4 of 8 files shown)


components/metrics/lib.rs
```diff
@@ -128,12 +134,20 @@ impl InteractiveWindow {
...
     }
 
     /// check if 10s has elapsed since start
     pub fn needs_check(&self) -> bool {
-        precise_time_ns() - self.start >= INTERACTIVE_WINDOW_SECONDS_IN_NS
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos() as u64 -
+            self.start >=
```
mrobinson, line 149: I think you could store `self.start` as a `SystemTime` and also `INTERACTIVE_WINDOW_DURATION` as a `Duration` here and then simply do:

```rust
SystemTime::now().duration_since(self.start).unwrap_or_default() >= INTERACTIVE_WINDOW_DURATION
```

which would simplify a lot of the code in this file.

components/shared/net/lib.rs
```diff
@@ -595,10 +596,25 @@ impl ResourceFetchTiming {
             return;
         }
         match attribute {
-            ResourceAttribute::DomainLookupStart => self.domain_lookup_start = precise_time_ns(),
+            ResourceAttribute::DomainLookupStart => {
+                self.domain_lookup_start = SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_nanos() as u64
+            },
```
mrobinson, line 604: Instead of repeating this code so many times, maybe just the current time before this match statement to simplify things? If it makes sense, perhaps store all these values as `Duration` as well?

mrobinson (changes requested):
Nice work. Just a few comments.

components/metrics/lib.rs
Taym95: I needed this is much better, thanks

components/metrics/lib.rs
```diff
@@ -33,7 +33,7 @@ pub trait ProgressiveWebMetric {
 /// maximum task time is 50ms (in ns)
 pub const MAX_TASK_NS: u64 = 50000000;
 /// 10 second window (in ns)
-const INTERACTIVE_WINDOW_SECONDS_IN_NS: u64 = 10000000000;
+const INTERACTIVE_WINDOW_SECONDS_IN_NS: Duration = Duration::from_nanos(10000000000);
```
mrobinson, line 36: I think the below will make this a bit more readable:

```suggestion
const INTERACTIVE_WINDOW_DURATION: Duration = Duration::from_secs(10);
```

components/script/dom/performanceresourcetiming.rs
```diff
@@ -112,24 +112,24 @@ impl PerformanceResourceTiming {
             entry: PerformanceEntry::new_inherited(
                 DOMString::from(url.into_string()),
                 DOMString::from("resource"),
-                resource_timing.start_time as f64,
-                resource_timing.response_end as f64 - resource_timing.start_time as f64,
+                resource_timing.start_time.as_nanos() as f64,
+                resource_timing.response_end.as_nanos() as f64 - resource_timing.start_time.as_nanos() as f64,
```
mrobinson, line 116: Can this avoid a bit of repetition?

```suggestion
                (resource_timing.response_end - resource_timing.start_time).as_nanos() as f64,
```

components/shared/net/lib.rs
```diff
@@ -502,21 +503,21 @@ pub struct ResourceCorsData {
...
+    pub response_start: Duration,
+    pub fetch_start: Duration,
+    pub response_end: Duration,
+    pub redirect_start: Duration,
+    pub redirect_end: Duration,
     pub connect_start: u64,
     pub connect_end: u64,
-    pub start_time: u64,
+    pub start_time: Duration,
 }
```
mrobinson, line 521: Nice!

components/shared/net/lib.rs
```diff
@@ -594,54 +595,69 @@ impl ResourceFetchTiming {
         if !self.timing_check_passed && !should_attribute_always_be_updated {
             return;
         }
+        let now  = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default();
```
mrobinson, line 600: You'll need to run `./mach fmt` on this change, I'm fairly sure.

components/shared/net/lib.rs
```diff
@@ -594,54 +595,69 @@ impl ResourceFetchTiming {
...
+            },
+            ResourceAttribute::ResponseStart => {
+                self.response_start = now
+            },
             ResourceAttribute::RedirectStart(val) => match val {
-                RedirectStartValue::Zero => self.redirect_start = 0,
+                RedirectStartValue::Zero => self.redirect_start = Duration::ZERO,
                 RedirectStartValue::FetchStart => {
-                    if self.redirect_start == 0 {
+                    if self.redirect_start == Duration::ZERO {
```
mrobinson, line 615: This can be:

```suggestion
                    if self.redirect_start.is_zero() {
```

components/shared/net/lib.rs
```diff
@@ -594,54 +595,69 @@ impl ResourceFetchTiming {
...
+                self.secure_connection_start = now
+            },
+            ResourceAttribute::ResponseEnd => {
+                self.response_end = now
             },
-            ResourceAttribute::ResponseEnd => self.response_end = precise_time_ns(),
             ResourceAttribute::StartTime(val) => match val {
                 ResourceTimeValue::RedirectStart
-                    if self.redirect_start == 0 || !self.timing_check_passed => {},
+                    if self.redirect_start == Duration::ZERO || !self.timing_check_passed => {},
```
mrobinson, line 637:
```suggestion
                    if self.redirect_start.is_zero() || !self.timing_check_passed => {},
```

components/style/context.rs
```diff
@@ -394,7 +394,10 @@ impl TraversalStatistics {
             dependency_selectors: stylist.num_invalidations() as u32,
             declarations: stylist.num_declarations() as u32,
             stylist_rebuilds: stylist.num_rebuilds() as u32,
-            traversal_time_ms: (time::precise_time_s() - start) * 1000.0,
+            traversal_time_ms: (SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs_f64() - start) * 1000.0,
```
mrobinson, line 400: Oh, sorry that I didn't notice this before, but please remove all changes from `style`. We are downstream from style which is in the Gecko source tree.

mrobinson (changes requested):
One more set of comments. This is very close. Thank you!

mrobinson (approved):

mrobinson:
It looks like this needs some fixes in unit tests. You can verify that these fail by running `./mach test-unit`. See https://github.com/servo/servo/actions/runs/7449081749/job/20264895970.

mrobinson (changes requested):
Requesting changes pending some unit test fixes. These should be pretty straight-forward, but feel free to comment if you need any help. Thanks!

Taym95:
> Requesting changes pending some unit test fixes. These should be pretty straight-forward, but feel free to comment if you need any help. Thanks!

Should be fixed now!

mrobinson (approved):

mrobinson:
Hrm. Seems that there is now an error when running WPT tests

```
│ overflow when subtracting durations (thread Script(1,1), at library/core/src/time.rs:930)
  │    1: servo::main2::main::{{closure}}
  │    2: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/alloc/src/boxed.rs:2021:9
  │       std::panicking::rust_panic_with_hook
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:735:13
  │    3: std::panicking::begin_panic_handler::{{closure}}
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:609:13
  │    4: std::sys_common::backtrace::__rust_end_short_backtrace
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/sys_common/backtrace.rs:170:18
  │    5: rust_begin_unwind
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
  │    6: core::panicking::panic_fmt
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
  │    7: core::panicking::panic_display
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:168:5
  │       core::panicking::panic_str
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:152:5
  │       core::option::expect_failed
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/option.rs:1988:5
  │    8: core::option::Option<T>::expect
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/option.rs:898:21
  │       <core::time::Duration as core::ops::arith::Sub>::sub
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/time.rs:930:31
  │    9: script::network_listener::submit_timing_data
  │   10: script::network_listener::submit_timing
  │   11: <script::network_listener::ListenerTask<A,Listener> as script::task::TaskOnce>::run_once
  │   12: <T as script::task::TaskBox>::run_box
  │   13: script::script_thread::ScriptThread::handle_msg_from_script
  │   14: script::script_thread::ScriptThread::handle_msgs
  │   15: script::script_thread::ScriptThread::start
  │   16: profile_traits::mem::ProfilerChan::run_with_memory_reporting
  │   17: std::sys_common::backtrace::__rust_begin_short_backtrace
  │   18: core::ops::function::FnOnce::call_once{{vtable.shim}}
  │   19: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/alloc/src/boxed.rs:2007:9
  │       <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/alloc/src/boxed.rs:2007:9
  │              at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/sys/unix/thread.rs:108:17
  │   20: <unknown>
  │   21: <unknown>
```

components/shared/net/lib.rs
```diff
@@ -594,54 +595,57 @@ impl ResourceFetchTiming {
...
+    fn get_time_value(&self, time: ResourceTimeValue) -> Duration {
         match time {
-            ResourceTimeValue::Zero => 0,
-            ResourceTimeValue::Now => precise_time_ns(),
+            ResourceTimeValue::Zero => Duration::ZERO,
+            ResourceTimeValue::Now => SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default(),
             ResourceTimeValue::FetchStart => self.fetch_start,
             ResourceTimeValue::RedirectStart => self.redirect_start,
```
mrobinson, line 638: It might make more sense to have this return a `Instant` which is guaranteed to be monotonic.

components/shared/net/lib.rs
```diff
@@ -567,18 +568,18 @@ impl ResourceFetchTiming {
...
+            response_start: Duration::ZERO,
+            fetch_start: Duration::ZERO,
+            redirect_start: Duration::ZERO,
+            redirect_end: Duration::ZERO,
             connect_start: 0,
             connect_end: 0,
-            response_end: 0,
-            start_time: 0,
+            response_end: Duration::ZERO,
+            start_time: Duration::ZERO,
```
mrobinson, line 582: Instant might be a better type for these as these a theoretically times.

components/script/dom/performanceresourcetiming.rs
```diff
@@ -112,24 +112,24 @@ impl PerformanceResourceTiming {
             entry: PerformanceEntry::new_inherited(
                 DOMString::from(url.into_string()),
                 DOMString::from("resource"),
-                resource_timing.start_time as f64,
-                resource_timing.response_end as f64 - resource_timing.start_time as f64,
+                resource_timing.start_time.as_nanos() as f64,
+                (resource_timing.response_end - resource_timing.start_time).as_nanos() as f64,
```
mrobinson, line 116: It seems we may have been covering up a bug here where `response_end` was either 0 or less than `start_time`. Perhaps it isn't being set? Perhaps this change can either address the bug or do a `checked_duration_since` here?

components/shared/net/lib.rs
Taym95: I also thought so, should it be `Instant ` or `SystemTime ` ?

mrobinson: From the documentation it seems that `Instant` is guaranteed to be monotonic, but `SystemTime` is not. Since we are subtracting in some places to find a `Duration`, I guess `Instant` might be a better fit. That said, I'm not sure if it's possible to properly create the initial value for `Instant`. It might require  bit of experimentation. Perhaps the unix epoch is a good initial value or using an `Option` or `OnceCell`.

components/script/dom/performanceresourcetiming.rs
mrobinson: If this issue is too time-consuming, another option is to give up on storing `Instant` and `SystemTime` now and just store nanoseconds. Apologies, I didn't realize there was this annoying bug here.

Taym95: Yeah that also can be an option, I am giving it a try

Taym95: @mrobinson I used nanoseconds but I will investigate this bug further in another PR

components/malloc_size_of/lib.rs
```diff
@@ -892,6 +892,10 @@ impl MallocSizeOf for xml5ever::QualName {
 malloc_size_of_is_0!(time::Duration);
 #[cfg(feature = "servo")]
 malloc_size_of_is_0!(time::Tm);
+#[cfg(feature = "servo")]
+malloc_size_of_is_0!(std::time::Duration);
+#[cfg(feature = "servo")]
+malloc_size_of_is_0!(std::time::SystemTime);
```
sagudev, line 898: I think servo feature is meant for external crates.

mrobinson (approved):

components/malloc_size_of/lib.rs
Taym95: I tried to use the same approach as `Tm` and `time::Duration `

PR #31026, opened 2024-01-08
bootstrap: Improve pip dependency resolution

### LINKED ISSUE #10611: Make `mach_bootstrap.py` use a dependency resolver
https://github.com/servo/servo/blob/bbdaf2262058fde6a938ccb47266e61b388d9727/python/mach_bootstrap.py#L184-L189

There are a variety of dependency resolvers suggested in the linked issue, https://github.com/pypa/pip/issues/988 

Steps: 
- [ ] Read https://github.com/pypa/pip/issues/988 and find a dependency resolver that looks good
    (for a value of "good" that includes works as expected, doesn't bring in too many extra dependencies of its own, actively maintained, has other projects are using it)
- [ ] Comment on this issue to discuss the proposed solution and address any concerns that it raises
- [ ] Refactor `mach_bootstrap.py` to use the selected resolver, and PR

mrobinson:
We can now use the "new" pip resolver which should prevent the
installation of conflicting packages. Also, take this opportunity to
make bootstrap more resilient. Hash all dependencies to detect
situations where a newer marker file has been installed, but for an
older branch. This should ensure that dependencies are up to date even
when switching back and forth between older and new branches.

This also updates some dependencies to be the same as the ones used for
WPT tests, which is an issue caught be the resolver.

Fixes #10611.

---

- [x] These changes fix #10611.
- [x] These changes do not require tests because they fix an issue with bootstrap.

### DIFF
--- a/python/mach_bootstrap.py
+++ b/python/mach_bootstrap.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import hashlib
 import os
 import platform
 import site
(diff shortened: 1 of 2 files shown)


mukilan (approved):

mrobinson:
Looks like this should also upgrade the version of pip used in the venv before trying to use the new dependency resolver. I've added a line to do this. We'll see if Windows old Python deals with that okay.

PR #31038, opened 2024-01-08
build: Combine and simplify GStreamer shared object lists

mrobinson:
Add documentation and simplify the way that GStreamer shared objects
lists are stored. In addition, move the list of extra GStreamer DLL
dependencies to to the `gstreamer.py` file.

The conditional plugin logic is no longer required as we have
already increased our GStreamer version requirements.

---


jdm (approved):
Nice simplification!

mrobinson:
There were issues with GitHub infrastructure, so sending this back to the MQ.

PR #31056, opened 2024-01-10
build: Simplify GStreamer shared object copying

mrobinson:
Add a new mach command to run only the post-build steps of copying
GStreamer shared objects. This allows more easily making improvements to
this step.

In addition make a few platform-specific changes:
 1. Simplify the way that dylib dependencies are chased and remove the rpath
    adjustment code. It was unused with recent GStreamer.
 2. Use exactly the same code when packaging dylibs for `./mach package`
    as for the post build action.
 3. For Windows only call vcvarsall before copying dependencies. This
    isn't necessary for the build.

---

- [x] These changes do not require tests because this shouldn't change behavior.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -332,28 +332,31 @@ class CommandBase(object):
         apk_name = "servoapp.apk"
         return path.join(base_path, build_type.directory_name(), apk_name)
 
-    def get_binary_path(self, build_type: BuildType, target=None, android=False, simpleservo=False):
+    def get_binary_path(self, build_type: BuildType, target=None, android=None):
+        if not android:
+            android = self.is_android_build
+
         base_path = util.get_target_dir()
-        if android:
+        if self.is_android_build:
             base_path = path.join(base_path, "android", self.config["android"]["target"])
-            simpleservo = True
         elif target:
             base_path = path.join(base_path, target)
 
-        binary_name = f"servo{servo.platform.get().executable_suffix()}"
-        if simpleservo:
-            if sys.platform == "win32":
-                binary_name = "simpleservo.dll"
-            elif sys.platform == "darwin":
-                binary_name = "libsimpleservo.dylib"
-            else:
-                binary_name = "libsimpleservo.so"
+        platform = servo.platform.get()
+        minibrowser_path = path.join(base_path, build_type.directory_name(),
+                                     f"servo{platform.executable_suffix()}")
+        simpleservo_path = path.join(base_path, build_type.directory_name(),
+                                     platform.shared_library_name("simpleservo"))
 
-        binary_path = path.join(base_path, build_type.directory_name(), binary_name)
-
-        if not path.exists(binary_path):
+        simpleservo_mtime = path.getmtime(simpleservo_path) if path.exists(simpleservo_path) else None
+        minibrowser_mtime = path.getmtime(minibrowser_path) if path.exists(minibrowser_path) else None
+        if not simpleservo_mtime and not minibrowser_mtime:
             raise BuildNotFound('No Servo binary found. Perhaps you forgot to run `./mach build`?')
-        return binary_path
+        if not minibrowser_mtime:
+            return simpleservo_path
+        if not simpleservo_mtime:
+            return minibrowser_path
+        return simpleservo_path if simpleservo_mtime > minibrowser_mtime else minibrowser_mtime
 
     def detach_volume(self, mounted_volume):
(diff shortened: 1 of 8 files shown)


python/servo/command_base.py
```diff
@@ -332,28 +332,31 @@ def get_apk_path(self, build_type: BuildType):
...
+        simpleservo_mtime = path.getmtime(simpleservo_path) if path.exists(simpleservo_path) else None
+        minibrowser_mtime = path.getmtime(minibrowser_path) if path.exists(minibrowser_path) else None
+        if not simpleservo_mtime and not minibrowser_mtime:
             raise BuildNotFound('No Servo binary found. Perhaps you forgot to run `./mach build`?')
-        return binary_path
+        if not minibrowser_mtime:
+            return simpleservo_path
+        if not simpleservo_mtime:
+            return minibrowser_path
+        return simpleservo_path if simpleservo_mtime > minibrowser_mtime else minibrowser_mtime
```
mukilan, line 359: Shouldn't the `else` clause return `minibrowser_path` instead of `minibrowser_mtime`?

Also, is there a way to avoid this heuristic of relying on modification times and have the method consider only input parameters?

mrobinson:
I'm going to rethink this a bit and hopefully post something later. There are complications I haven't considered with unit tests. They need to link against GStreamer properly and I need to do some investigation about how to make this happen consistently.

sagudev:
Why is using GStreamer so hard in servo. According to https://docs.rs/gstreamer/latest/gstreamer/ one time install in bootstrap should be all we need.

Maybe having dummy crate that would have build script, that would set and link gstreamer could solve some problems?.

mrobinson:
> Why is using GStreamer so hard in servo. According to https://docs.rs/gstreamer/latest/gstreamer/ one time install in bootstrap should be all we need.
> 
> Maybe having dummy crate that would have build script, that would set and link gstreamer could solve some problems?.

I have been pondering this question for a bit! I think there are a few complications depending on the platform:

## MacOS

 - You need to set up the environment beforehand. Some environment variables we need in order to compile against the GStreamer crates: `PATH`, `PKG_CONFIG_PATH`, `DYLD_LIBRARY_PATH` (maybe not necessary?), `GST_PLUGIN_SCANNER`, `GST_PLUGIN_SYSTEM_PATH`.
 - Once the build finishes, things should run as is -- but the binary isn't ready to distribute because it relies on the environment or the rpath set during build (not sure which of these is relevant) to find the dylibs and run.
 - We copy the relevant bits of GStreamer into a directory called `lib` which is [defined via an rpath](https://github.com/servo/servo/pull/30054). Which bits of GStreamer do we copy though? The whole distribution is very large, so we use otool to trace dependencies down and copy them recursively.
 - How does this work for unit tests, which are their own binary and might be built in separate directories? I have no idea.

## Windows
- We have to set up a similar environment variable beforehand for Windows.
- Windows finds DLLs by looking alongside the final executable (and sometimes in the PATH variable?).
- We must copy three kinds of DLLs:
  - MSVC and Windows SDK DLLs - This is why we have complex code to track down where MSVC is installed.
  - The EGL and GLESv2 DLLs built as part of mozangle. There doesn't seem to be a way for mozangle to compile these into the final directory automatically. I don't know what mechanism (like rpath) that rustc / cargo use to allow adding DLL search directories. Could we compile EGL / GLESv2 into the final binary statically? This seems like it would be an improvement!
- We don't have code to recursively chase DLL dependencies like the MacOS otool code. We could theoretically write this with dumpbin, but right now the list of GStreamer DLL dependencies is managed manually.

This is all a lot of work to manage. It would be nice to find simpler ways of dealing with this.

Ideas:
 - Try just setting the PATH for GStreamer instead of all the variables we set. Maybe this is enough!
 - Perhaps one thing we could do is turn off media support for the time being. It's unclear to me if it still works and no one is maintaining it.

sagudev:
or maybe using gstreamer as static lib: https://blogs.igalia.com/scerveau/discover-gstreamer-full-static-mode/

mrobinson:
> or maybe using gstreamer as static lib: https://blogs.igalia.com/scerveau/discover-gstreamer-full-static-mode/

Yes, that's an option, but I'm not sure how well that works on platforms other than Linux.

sagudev:
> Could we compile EGL / GLESv2 into the final binary statically?

We could (we have enabled build_dlls feature of mozangle that forces dynamic instead of static), but some other crates also expects to have EGL/GLES symbols available IIRC.

PR #31063, opened 2024-01-11
Fix underflow in PerformanceResourceTiming API

### LINKED ISSUE #31062: servoshell crashes on servo.org on main
92196d985dceb0ca708b097e2a847b255d8387c8
- broken for @mukilan on NixOS 23.05 with clang 11 and clang 14
- broken for me on Ubuntu 23.10 (no Nix) with clang 14

image

delan:
[servo.org](https://servo.org/) crashes in debug builds after #31020, because we take the difference of two u64 values before converting that to f64. This patch fixes that by converting the values to f64 before subtracting.

---

- [x] These changes fix #31062

- [x] These changes require tests, but our CI does not support testing debug builds yet

### DIFF
--- a/components/script/dom/performanceresourcetiming.rs
+++ b/components/script/dom/performanceresourcetiming.rs
@@ -113,7 +113,7 @@ impl PerformanceResourceTiming {
                 DOMString::from(url.into_string()),
                 DOMString::from("resource"),
                 resource_timing.start_time as f64,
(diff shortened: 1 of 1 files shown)


atbrakhi (approved):

mrobinson (approved):
Thanks. I should have caught this in the review. :/

PR #31073, opened 2024-01-12
fix conditional logic that enables native bluetooth

mukilan:
PR #30974 integrated `servo/devices` repo into servo codebase. `servo/devices` exposed the `bluetooth`
feature to conditionally compile native bluetooth support for the target platform. In servo, this feature is indirectly enabled via the `native-bluetooth` feature exposed by `components/bluetooth`. When `servo/devices` was integrated to servo, the conditional code was not updated to use the `native-bluetooth` feature directly.

---

- [x] These changes do not require tests because they re-enable code that used to be compiled before #30974. Existing tests should hopefully catch integration issues missed by #30974 which upgrades the dependencies.

### DIFF
--- a/components/bluetooth/adapter.rs
+++ b/components/bluetooth/adapter.rs
@@ -5,17 +5,17 @@
 use std::error::Error;
 use std::sync::Arc;
 
-#[cfg(all(target_os = "android", feature = "bluetooth"))]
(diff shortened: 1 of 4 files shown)


mrobinson (approved):
Thanks for this fix!

PR #31076, opened 2024-01-12
Use FLoat32Array in XRRigidTransform

### LINKED ISSUE #31047: WebIDL: use FLoat32Array in XRRigidTransform
Depends on https://github.com/servo/servo/pull/30990

How to do it:

  1.  In [CodegenRust.py](https://github.com/servo/servo/blob/main/components/script/dom/bindings/codegen/CodegenRust.py) remove XRRigidTransform from the list in todo_switch_float_32.
  2. Replace [`matrix: Heap<*mut JSObject>`](https://github.com/servo/servo/blob/355567186400d0531657d0eedda7625de9d0e93a/components/script/dom/xrrigidtransform.rs#L34C5-L34C32) with `HeapFloat32Array`. 
  3. Replace [`Matrix`](https://github.com/servo/servo/blob/355567186400d0531657d0eedda7625de9d0e93a/components/script/dom/xrrigidtransform.rs#L137) to return a `Float32Array`, by using `set_data` and doing the same as in [AudioBuffer::GetChannelData](https://github.com/servo/servo/pull/30990/files#diff-88b03ac9bcc2650883607d41bc374d7f3976567cfbee702cc124f14ce2fbc79bR231)
  
Any questions, ask here or in https://servo.zulipchat.com/

Taym95:
---

- [x] These changes fix #31047

- [x] These changes do not require tests because doesn't change behavior

### DIFF
--- a/components/script/dom/xrrigidtransform.rs
+++ b/components/script/dom/xrrigidtransform.rs
@@ -2,19 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use std::ptr::NonNull;
-
 use dom_struct::dom_struct;
 use euclid::{RigidTransform3D, Rotation3D, Vector3D};
-use js::jsapi::{Heap, JSObject};
 use js::rust::HandleObject;
+use js::typedarray::Float32Array;
 
 use crate::dom::bindings::codegen::Bindings::DOMPointBinding::DOMPointInit;
 use crate::dom::bindings::codegen::Bindings::XRRigidTransformBinding::XRRigidTransformMethods;
 use crate::dom::bindings::error::{Error, Fallible};
 use crate::dom::bindings::reflector::{reflect_dom_object_with_proto, DomObject, Reflector};
 use crate::dom::bindings::root::{DomRoot, MutNullableDom};
-use crate::dom::bindings::utils::create_typed_array;
+use crate::dom::bindings::typedarrays::HeapFloat32Array;
 use crate::dom::dompointreadonly::DOMPointReadOnly;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::window::Window;
@@ -31,7 +29,7 @@ pub struct XRRigidTransform {
     transform: ApiRigidTransform,
     inverse: MutNullableDom<XRRigidTransform>,
     #[ignore_malloc_size_of = "defined in mozjs"]
-    matrix: Heap<*mut JSObject>,
+    matrix: HeapFloat32Array,
 }
 
 impl XRRigidTransform {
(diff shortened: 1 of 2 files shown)


components/script/dom/xrrigidtransform.rs
```diff
@@ -134,15 +134,11 @@ impl XRRigidTransformMethods for XRRigidTransform {
...
-        if self.matrix.get().is_null() {
-            let cx = GlobalScope::get_cx();
-            // According to the spec all matrices are column-major,
-            // however euclid uses row vectors so we use .to_array()
-            let arr = self.transform.to_transform().to_array();
-            create_typed_array(cx, &arr, &self.matrix);
+    fn Matrix(&self, _cx: JSContext) -> Float32Array {
+        match self.matrix.get_internal() {
+            Ok(matrix) => matrix,
+            Err(_) => todo!(),
```
Taym95, line 140: @gterzian what should we return in this case?

gterzian: What you want to do is: in the `!self.matrix.is_initialized` branch, use `set_data` to initialize the float 32 array to the data of `self.transform`, and only then call `self.matrix.get_internal().expect`

gterzian (changes requested):
Thanks for the contribution! Clarifying things a bit...

components/script/dom/xrrigidtransform.rs
```diff
@@ -134,15 +132,16 @@ impl XRRigidTransformMethods for XRRigidTransform {
...
-            let cx = GlobalScope::get_cx();
-            // According to the spec all matrices are column-major,
-            // however euclid uses row vectors so we use .to_array()
-            let arr = self.transform.to_transform().to_array();
-            create_typed_array(cx, &arr, &self.matrix);
+    fn Matrix(&self, _cx: JSContext) -> Float32Array {
+        if !self.matrix.is_initialized() {
+            self.matrix
+                .set_data(_cx, &self.transform.to_transform().to_array())
+                .expect("failed to set projection matrix")
```
gterzian, line 139: "Failed to set on data on transform's internal matrix."

components/script/dom/xrrigidtransform.rs
```diff
@@ -134,15 +132,16 @@ impl XRRigidTransformMethods for XRRigidTransform {
...
         }
-        NonNull::new(self.matrix.get()).unwrap()
+
+        self.matrix
+            .get_internal()
+            .expect("failed to get projection matrix")
```
gterzian, line 144: "Failed to get transform's internal matrix."

gterzian (approved):
LGTM, with a couple of nits

gterzian (approved):
LGTM, thanks for the contrib, we can merge this one after https://github.com/servo/servo/pull/31077 has, which will probably require fixing a merge conflict first...

PR #31077, opened 2024-01-12
use FLoat32Array in XRView

### LINKED ISSUE #31046: WebIDL: use FLoat32Array in XRView
Follow-up to https://github.com/servo/servo/pull/30990

How to do it:

1. In [`CodegenRust.py`](https://github.com/servo/servo/blob/main/components/script/dom/bindings/codegen/CodegenRust.py) remove `XRView` from the list in `todo_switch_float_32`.
2. Replace [`proj: Heap<*mut JSObject>`](https://github.com/servo/servo/blob/355567186400d0531657d0eedda7625de9d0e93a/components/script/dom/xrview.rs#L28) with `HeapFloat32Array`
3. Change [`ProjectionMatrix`](https://github.com/servo/servo/blob/355567186400d0531657d0eedda7625de9d0e93a/components/script/dom/xrview.rs#L93) to return a `Float32Array` by doing the same as in [`AudioBuffer::GetChannelData`](https://github.com/servo/servo/pull/30990/files#diff-88b03ac9bcc2650883607d41bc374d7f3976567cfbee702cc124f14ce2fbc79bR231)

Any questions, ask here or in https://servo.zulipchat.com/

tannal:
---

- [ ] These changes fix #31046 (GitHub issue number if applicable)

- [x] These changes do not require tests because it doesn't change the bahavior.

### DIFF
--- a/components/script/dom/xrview.rs
+++ b/components/script/dom/xrview.rs
@@ -93,7 +93,9 @@ impl XRViewMethods for XRView {
             let cx = GlobalScope::get_cx();
             // row_major since euclid uses row vectors
             let proj = self.view.projection.to_array();
-            let _ = self.proj.set_data(cx, &proj);
+            if self.proj.set_data(cx, &proj).is_err() {
+                    // error handling
+             }
         }
 
         match self.proj.get_internal() {
--- a/components/script/dom/xrview.rs
+++ b/components/script/dom/xrview.rs
@@ -95,11 +95,11 @@ impl XRViewMethods for XRView {
             let proj = self.view.projection.to_array();
             self.proj
                 .set_data(cx, &proj)
-                .expect("failed to set projection matrix")
+                .expect("Failed to set projection matrix.")
         }
         self.proj
             .get_internal()
-            .expect("failed to get projection matrix")
+            .expect("Failed to get projection matrix.")
     }
 
     /// https://immersive-web.github.io/webxr/#dom-xrview-transform
--- a/components/script/dom/xrview.rs
+++ b/components/script/dom/xrview.rs
@@ -2,17 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use std::ptr::NonNull;
-
 use dom_struct::dom_struct;
 use euclid::RigidTransform3D;
-use js::jsapi::{Heap, JSObject};
+use js::typedarray::Float32Array;
 use webxr_api::{ApiSpace, View};
 
 use crate::dom::bindings::codegen::Bindings::XRViewBinding::{XREye, XRViewMethods};
 use crate::dom::bindings::reflector::{reflect_dom_object, Reflector};
 use crate::dom::bindings::root::{Dom, DomRoot};
-use crate::dom::bindings::utils::create_typed_array;
+use crate::dom::bindings::typedarrays::HeapFloat32Array;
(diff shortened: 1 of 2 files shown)


components/script/dom/xrview.rs
```diff
@@ -90,14 +88,18 @@ impl XRViewMethods for XRView {
...
             let cx = GlobalScope::get_cx();
             // row_major since euclid uses row vectors
             let proj = self.view.projection.to_array();
-            create_typed_array(cx, &proj, &self.proj);
+            let _ = self.proj.set_data(cx, &proj);
+        }
+
+        match self.proj.get_internal() {
+            Ok(array) => array,
+            Err(_) => unreachable!(),
```
Taym95, line 101: are we should this is unreachable ?

components/script/dom/xrview.rs
```diff
@@ -90,14 +88,18 @@ impl XRViewMethods for XRView {
...
     /// https://immersive-web.github.io/webxr/#dom-xrview-projectionmatrix
-    fn ProjectionMatrix(&self, _cx: JSContext) -> NonNull<JSObject> {
-        if self.proj.get().is_null() {
+    fn ProjectionMatrix(&self, _cx: JSContext) -> Float32Array {
+        if !self.proj.is_initialized() {
             let cx = GlobalScope::get_cx();
             // row_major since euclid uses row vectors
             let proj = self.view.projection.to_array();
-            create_typed_array(cx, &proj, &self.proj);
+            let _ = self.proj.set_data(cx, &proj);
```
Taym95, line 96: set_data returns `Result<(), ()>` we should handle err:

```suggestion
            if self.proj.set_data(cx, &proj).is_err() {
                    // error handling
             }
```

tannal: Yes, we should handle this.

components/script/dom/xrview.rs
tannal: AFAIK, If we handle the error from set_data (initialization fails), then L101 is unreachable.

gterzian: Since the [WebIDL](https://immersive-web.github.io/webxr/#ref-for-dom-xrview-projectionmatrix%E2%91%A1) and the [backing algorithm in the spec](https://immersive-web.github.io/webxr/#xrview-obtain-the-projection-matrix)do not allow for failure, there is no way we can handle failure, so one could simply `expect` on both errors.

gterzian (changes requested):
Thanks for the contribution! Clarifying the error handling. also relevant for cc @Taym95's PR

gterzian (approved):
LGTM, with a couple of nits

components/script/dom/xrview.rs
```diff
@@ -90,14 +88,18 @@ impl XRViewMethods for XRView {
...
-        if self.proj.get().is_null() {
+    fn ProjectionMatrix(&self, _cx: JSContext) -> Float32Array {
+        if !self.proj.is_initialized() {
             let cx = GlobalScope::get_cx();
             // row_major since euclid uses row vectors
             let proj = self.view.projection.to_array();
-            create_typed_array(cx, &proj, &self.proj);
+            self.proj
+                .set_data(cx, &proj)
+                .expect("failed to set projection matrix")
```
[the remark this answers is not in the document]
tannal: Sounds great.

gterzian (approved):
LGMT, thanks for your contrib. I will merge this one first, and I assume [the other ](https://github.com/servo/servo/pull/31076) will have to resolve a conflict firs...

PR #31078, opened 2024-01-12
build: Only set PATH for GStreamer

mrobinson:
The documentation [^1][^2] claims that this is the only environment variable that
needs to be set before build.

[^1]: https://docs.rs/gstreamer/latest/gstreamer/#gstreamer-binaries-1
[^2]:  https://docs.rs/gstreamer/latest/gstreamer/#gstreamer-binaries

---

- [x] These changes do not require tests because this shouldn't change behavior.

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -498,9 +498,13 @@ class CommandBase(object):
         """Return an extended environment dictionary."""
         env = os.environ.copy()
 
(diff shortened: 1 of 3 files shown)


mrobinson:
Try run here: https://github.com/servo/servo/actions/runs/7503665184

mukilan (approved):

PR #31079, opened 2024-01-12
Replace time with std::time in components/net

### LINKED ISSUE #30150: Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate

Taym95:
---

- [x] These changes fix #30150

### DIFF
--- a/components/devtools/actors/network_event.rs
+++ b/components/devtools/actors/network_event.rs
@@ -9,7 +9,7 @@
 use std::net::TcpStream;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use chrono::{Local, LocalResult, TimeZone};
+use chrono::{DateTime, Local};
 use devtools_traits::{HttpRequest as DevtoolsHttpRequest, HttpResponse as DevtoolsHttpResponse};
 use headers::{ContentType, Cookie, HeaderMapExt};
 use http::{header, HeaderMap, Method, StatusCode};
@@ -371,23 +371,13 @@ impl NetworkEventActor {
     pub fn event_actor(&self) -> EventActor {
         // TODO: Send the correct values for startedDateTime, isXHR, private
 
-        let started_datetime_rfc3339 = match Local.timestamp_millis_opt(
-            self.request
-                .startedDateTime
-                .duration_since(UNIX_EPOCH)
-                .unwrap_or_default()
-                .as_millis() as i64,
-        ) {
-            LocalResult::None => "".to_owned(),
-            LocalResult::Single(dateTime) => format!("{}", dateTime.to_rfc3339()),
-            LocalResult::Ambiguous(dateTime, _) => format!("{}", dateTime.to_rfc3339()),
-        };
+        let date_time: DateTime<Local> = self.request.startedDateTime.clone().into();
 
         EventActor {
             actor: self.name(),
             url: self.request.url.clone(),
             method: format!("{}", self.request.method),
-            startedDateTime: started_datetime_rfc3339,
+            startedDateTime: date_time.to_rfc3339(),
             timeStamp: self.request.timeStamp,
             isXHR: self.is_xhr,
             private: false,
--- a/components/hyper_serde/lib.rs
+++ b/components/hyper_serde/lib.rs
@@ -20,7 +20,7 @@
 //! * `hyper::Method`
 //! * `hyper::Uri`
 //! * `mime::Mime`
-//! * `time::Tm`
+//! * `std::time::SystemTime`
 //!
 //! # How do I use a data type with a `HeaderMap` member with Serde?
 //!
@@ -66,8 +66,10 @@
 
 use std::ops::{Deref, DerefMut};
 use std::str::FromStr;
+use std::time::SystemTime;
 use std::{cmp, fmt, str};
 
+use chrono::{DateTime, SecondsFormat, Utc};
 use cookie::Cookie;
 use headers::ContentType;
 use http::HeaderMap;
@@ -78,7 +80,6 @@ use serde::de::{self, Error, MapAccess, SeqAccess, Visitor};
 use serde::ser::{SerializeMap, SerializeSeq};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_bytes::{ByteBuf, Bytes};
-use time::{strptime, Tm};
 
 /// Deserialises a `T` value with a given deserializer.
 ///
@@ -604,15 +605,15 @@ impl<'de> Visitor<'de> for StatusVisitor {
     }
 }
 
-impl<'de> Deserialize<'de> for De<Tm> {
+impl<'de> Deserialize<'de> for De<SystemTime> {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: Deserializer<'de>,
     {
-        struct TmVisitor;
+        struct SystemTimeVisitor;
 
-        impl<'de> Visitor<'de> for TmVisitor {
-            type Value = De<Tm>;
+        impl<'de> Visitor<'de> for SystemTimeVisitor {
+            type Value = De<SystemTime>;
 
             fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                 write!(formatter, "a date and time according to RFC 3339")
@@ -622,22 +623,23 @@ impl<'de> Deserialize<'de> for De<Tm> {
             where
                 E: de::Error,
             {
-                strptime(v, "%Y-%m-%dT%H:%M:%SZ")
-                    .map(De::new)
+                DateTime::parse_from_rfc3339(v)
+                    .map(|t| De::new(SystemTime::from(t)))
                     .map_err(|e| E::custom(e.to_string()))
             }
         }
 
-        deserializer.deserialize_string(TmVisitor)
+        deserializer.deserialize_string(SystemTimeVisitor)
     }
 }
 
-impl<'a> Serialize for Ser<'a, Tm> {
+impl<'a> Serialize for Ser<'a, SystemTime> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        serializer.serialize_str(&self.v.rfc3339().to_string())
+        let date_time: DateTime<Utc> = self.v.clone().into();
+        serializer.serialize_str(&date_time.to_rfc3339_opts(SecondsFormat::Secs, true))
     }
 }
 
--- a/components/hyper_serde/tests/tokens.rs
+++ b/components/hyper_serde/tests/tokens.rs
@@ -15,7 +15,6 @@ use http::StatusCode;
 use hyper::{Method, Uri};
 use hyper_serde::{De, Ser};
 use serde_test::{assert_de_tokens, assert_ser_tokens, Token};
-use time::Duration;
 
 #[test]
 fn test_content_type() {
@@ -32,7 +31,7 @@ fn test_cookie() {
     // string with a bunch of indices in it which apparently is different from the exact same
     // cookie but parsed as a bunch of strings.
     let cookie: Cookie = Cookie::build("Hello", "World!")
-        .max_age(Duration::seconds(42))
+        .max_age(time::Duration::seconds(42))
         .domain("servo.org")
         .path("/")
         .secure(true)
@@ -112,14 +111,18 @@ fn test_raw_status() {
 }
 
 #[test]
-fn test_tm() {
-    use time::strptime;
+fn test_system_time_serialization() {
+    use std::time::SystemTime;
 
-    let time = strptime("2017-02-22T12:03:31Z", "%Y-%m-%dT%H:%M:%SZ").unwrap();
-    let tokens = &[Token::Str("2017-02-22T12:03:31Z")];
+    use chrono::{NaiveDateTime, TimeZone, Utc};
 
-    assert_ser_tokens(&Ser::new(&time), tokens);
+    let time = SystemTime::from(Utc.from_utc_datetime(
+        &NaiveDateTime::parse_from_str("2023-01-15T12:53:31Z", "%Y-%m-%dT%H:%M:%SZ").unwrap(),
+    ));
+    let tokens = &[Token::Str("2023-01-15T12:53:31Z")];
+
     assert_de_tokens(&De::new(time), tokens);
+    assert_ser_tokens(&Ser::new(&time), tokens);
 }
 
 #[test]
--- a/components/net/cookie.rs
+++ b/components/net/cookie.rs
@@ -7,13 +7,13 @@
 
 use std::borrow::ToOwned;
 use std::net::{Ipv4Addr, Ipv6Addr};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use hyper_serde::Serde;
 use net_traits::pub_domains::is_pub_domain;
 use net_traits::CookieSource;
 use serde::{Deserialize, Serialize};
 use servo_url::ServoUrl;
-use time::{at, now, Duration, Tm};
 
 /// A stored cookie that wraps the definition in cookie-rs. This is used to implement
 /// various behaviours defined in the spec that rely on an associated request URL,
@@ -31,13 +31,13 @@ pub struct Cookie {
         deserialize_with = "hyper_serde::deserialize",
         serialize_with = "hyper_serde::serialize"
     )]
-    pub creation_time: Tm,
+    pub creation_time: SystemTime,
     #[serde(
         deserialize_with = "hyper_serde::deserialize",
         serialize_with = "hyper_serde::serialize"
     )]
-    pub last_access: Tm,
-    pub expiry_time: Option<Serde<Tm>>,
+    pub last_access: SystemTime,
+    pub expiry_time: Option<Serde<SystemTime>>,
 }
 
 impl Cookie {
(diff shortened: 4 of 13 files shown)


components/hyper_serde/lib.rs
```diff
@@ -20,7 +20,7 @@
 //! * `hyper::Method`
 //! * `hyper::Uri`
 //! * `mime::Mime`
-//! * `time::Tm`
+//! * `std::time::SystemTime`
```
Taym95, line 23: I am fixing unit test

components/net/cookie_storage.rs
```diff
@@ -229,14 +229,19 @@ fn reg_host<'a>(url: &'a str) -> String {
 
 fn is_cookie_expired(cookie: &Cookie) -> bool {
     match cookie.expiry_time {
-        Some(ref t) => t.to_timespec() <= time::get_time(),
+        Some(ref t) => {
+            t.duration_since(UNIX_EPOCH).unwrap_or_default() <=
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+        },
```
mrobinson, line 237: Can't these just be compared directly with`t < SystemTime.now()`?

components/net/cookie_storage.rs
```diff
@@ -229,14 +229,19 @@ fn reg_host<'a>(url: &'a str) -> String {
...
         None => false,
     }
 }
 
 fn evict_one_cookie(is_secure_cookie: bool, cookies: &mut Vec<Cookie>) -> bool {
     // Remove non-secure cookie with oldest access time
-    let oldest_accessed: Option<(usize, Tm)> = get_oldest_accessed(false, cookies);
+    let oldest_accessed: Option<(usize, SystemTime)> = get_oldest_accessed(false, cookies);
```
mrobinson, line 244: I think you can omit the type here and below. The compiler should be able to figure this out, right?

```suggestion
    let oldest_accessed = get_oldest_accessed(false, cookies);
```

components/net/fetch/cors_cache.rs
```diff
@@ -48,7 +49,7 @@ pub struct CorsCacheEntry {
     pub max_age: u32,
```
mrobinson, line 49: I think making `max_age` a `Duration` here will greatly simplify the code and make the units clearer.

components/net/fetch/cors_cache.rs
```diff
@@ -111,10 +112,20 @@ impl CorsCache {
...
         let new_buf: Vec<CorsCacheEntry> = buf
             .into_iter()
-            .filter(|e| now.sec < e.created.sec + e.max_age as i64)
+            .filter(|e| {
+                now.as_secs() <
+                    (e.created
+                        .duration_since(UNIX_EPOCH)
+                        .unwrap_or_default()
+                        .as_secs() as u64) +
+                        e.max_age as u64
```
mrobinson, line 127: You would also be able to do simply `now < e.create + e.max_age` was a `Duration`, I think.

components/net/hsts.rs
```diff
@@ -36,15 +37,25 @@ impl HstsEntry {
...
     pub fn is_expired(&self) -> bool {
         match (self.max_age, self.timestamp) {
             (Some(max_age), Some(timestamp)) => {
-                (time::get_time().sec as u64) - timestamp >= max_age
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_secs() -
+                    timestamp >=
+                    max_age
```
mrobinson, line 58: See above.

components/net/http_loader.rs
```diff
@@ -122,7 +122,11 @@ impl HttpState {
 }
 
 fn precise_time_ms() -> u64 {
-    time::precise_time_ns() / (1000 * 1000)
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_nanos() as u64 /
+        (1000 * 1000)
```
mrobinson, line 129: Hrm...would it be possible to just do `as_millis()` here and avoid the math?

mrobinson (changes requested):
Great stuff. I have a few comments, but in theory this looks pretty good.

components/hyper_serde/lib.rs
```diff
@@ -622,22 +623,40 @@ impl<'de> Deserialize<'de> for De<Tm> {
...
 }
 
-impl<'a> Serialize for Ser<'a, Tm> {
+impl<'a> Serialize for Ser<'a, SystemTime> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        serializer.serialize_str(&self.v.rfc3339().to_string())
+        let rfc3339 = match Utc.timestamp_millis_opt(
```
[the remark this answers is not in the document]
Taym95: I tried it but `SystemTime ` does not implement `std::convert::Into`

components/net/cookie_storage.rs
Taym95: less than or equal is not implement for SystemTime

components/net/cookie_storage.rs
Taym95: Ah yes indeed

components/net/cookie_storage.rs
mrobinson: It seems like it is though? https://doc.rust-lang.org/std/time/struct.SystemTime.html#impl-PartialOrd-for-SystemTime

components/hyper_serde/lib.rs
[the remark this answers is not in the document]
mrobinson: It seems that this should be possible but you might need to clone first: https://stackoverflow.com/a/64148190

components/net/cookie_storage.rs
Taym95: Ah sorry serde misleaded me, should be `t.0 <= SystemTime::now()`

components/devtools/actors/network_event.rs
```diff
@@ -371,23 +371,13 @@ impl NetworkEventActor {
...
-                .startedDateTime
-                .duration_since(UNIX_EPOCH)
-                .unwrap_or_default()
-                .as_millis() as i64,
-        ) {
-            LocalResult::None => "".to_owned(),
-            LocalResult::Single(dateTime) => format!("{}", dateTime.to_rfc3339()),
-            LocalResult::Ambiguous(dateTime, _) => format!("{}", dateTime.to_rfc3339()),
-        };
+        let date_time: DateTime<Local> = self.request.startedDateTime.clone().into();
```
Taym95, line 374: @mrobinson I used `clone().into()` and removed `match ` I added in previous PR

components/net/tests/cookie.rs
```diff
@@ -398,7 +398,7 @@ fn test_cookie_eviction_all_secure_one_nonsecure() {
     let mut vec = Vec::new();
     for i in 1..5 {
         let st = format!(
-            "extra{}=bar; Secure; expires=Sun, 18-Apr-2026 21:06:29 GMT",
+            "extra{}=bar; Secure; expires=Sun, 18-Apr-2026 21:06:29:01 GMT",
```
Taym95, line 401: @mrobinson  I needed yo add this to make unit test pass

components/hyper_serde/tests/tokens.rs
```diff
@@ -32,7 +34,6 @@ fn test_cookie() {
     // string with a bunch of indices in it which apparently is different from the exact same
     // cookie but parsed as a bunch of strings.
     let cookie: Cookie = Cookie::build("Hello", "World!")
-        .max_age(Duration::seconds(42))
```
Taym95, line 35: @mrobinson Cookie crate uses `time` crate and  max_age takes `time:: Duration`, I removed it and adjusted the test, what do you think?

mrobinson: Is this to avoid the conflict in versions? I think it's okay to depend on the new version of `time` (0.3, I think) in order to use cookie in this way. We are already pulling it in if it's a dependency.

components/net/cookie.rs
```diff
@@ -135,18 +136,18 @@ impl Cookie {
...
     }
 
     pub fn touch(&mut self) {
-        self.last_access = now();
+        self.last_access = SystemTime::now();
     }
 
     pub fn set_expiry_time_negative(&mut self) {
-        self.expiry_time = Some(Serde(now() - Duration::seconds(1)));
+        self.expiry_time = Some(Serde(SystemTime::now() - Duration::from_nanos(1)));
```
mrobinson, line 150: I think this should be `from_secs` to match behavior?

components/net/cookie.rs
```diff
@@ -62,11 +62,12 @@ impl Cookie {
...
             (Some(max_age), _) => (
                 true,
-                Some(at(
-                    now().to_timespec() + Duration::seconds(max_age.num_seconds())
-                )),
+                Some(SystemTime::now() + Duration::from_nanos(max_age.num_seconds() as u64)),
+            ),
+            (_, Some(expires)) => (
+                true,
+                Some(SystemTime::UNIX_EPOCH + Duration::from_nanos(expires.tm_sec as u64)),
```
mrobinson, line 69: Should this use `Duration::from_secs` in both of these cases?

components/net/http_cache.rs
```diff
@@ -211,18 +209,19 @@ fn get_response_expiry(response: &Response) -> Duration {
...
+            let desired = t.duration_since(UNIX_EPOCH).unwrap_or_default();
+            let current = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default();
 
             if desired > current {
                 return desired - current;
             } else {
-                return Duration::seconds(0i64);
+                return Duration::from_secs(0u64);
```
mrobinson, line 220: Perhaps this could use `checked_sub` and just `unwrap_or_else` to return 0?

components/net/http_cache.rs
```diff
@@ -280,24 +278,22 @@ fn get_expiry_adjustment_from_request_headers(request: &Request, expires: Durati
     };
 
     if let Some(max_age) = directive.max_stale() {
-        return expires + Duration::from_std(max_age).unwrap();
+        return expires + max_age;
```
mrobinson, line 281: Nice cleanup!

components/net/http_cache.rs
```diff
@@ -813,7 +816,7 @@ impl HttpCache {
         let entry_key = CacheKey::from_servo_url(url);
         if let Some(cached_resources) = self.entries.get_mut(&entry_key) {
             for cached_resource in cached_resources.iter_mut() {
-                cached_resource.data.expires = Duration::seconds(0i64);
+                cached_resource.data.expires = Duration::from_secs(0u64);
```
mrobinson, line 819: Could this be Duration::ZERO?

components/net/tests/cookie.rs
mrobinson: Hrm...

components/net/http_cache.rs
```diff
@@ -268,7 +266,7 @@ fn get_response_expiry(response: &Response) -> Duration {
         }
     }
     // Requires validation upon first use as default.
-    Duration::seconds(0i64)
+    Duration::from_secs(0u64)
```
mrobinson, line 269: Duration::ZERO?

mrobinson (changes requested):
Nice work. This is really coming along. I have a few comments below. I would try to change all of the instances of `Duration::from_secs(0u64)` to `Duration::ZERO` as well.

components/net/cookie.rs
Taym95: I think this why I unit test did not pass, I will change it and check

components/hyper_serde/tests/tokens.rs
mrobinson: If you need to import more than one version of `Duration` into the same code units you can do something like:

```rust
use std::time::Duration as StdDuration;
```

components/net/tests/cookie.rs
Taym95: Yes, it's weird, but if I don't add seconds, none of these cookies will appear in the test result, and only one will be added: `foo=bar`

mrobinson: This is very odd though because this isn't seconds but something else:

21:06:29:01 is 21 hours, 6 minutes, 29 seconds and 1...millisecond? I'm not sure what is going on here.

components/hyper_serde/lib.rs
```diff
@@ -604,15 +605,15 @@ impl<'de> Visitor<'de> for StatusVisitor {
     }
 }
 
-impl<'de> Deserialize<'de> for De<Tm> {
+impl<'de> Deserialize<'de> for De<SystemTime> {
```
mrobinson, line 608: It looks like serde already supports serialization of `SystemTime`. Maybe removing this code and relying on the implementation from serde will fix the issue you are having with the tests?

https://github.com/serde-rs/serde/pull/949

components/net/tests/cookie.rs
Taym95: Sorry I meant `millisecond`, yes apparently I need to add `millisecond`, I will debug it further

components/hyper_serde/lib.rs
```diff
@@ -622,22 +623,26 @@ impl<'de> Deserialize<'de> for De<Tm> {
             where
                 E: de::Error,
             {
-                strptime(v, "%Y-%m-%dT%H:%M:%SZ")
-                    .map(De::new)
+                NaiveDateTime::parse_from_str(v, "%Y-%m-%dT%H:%M:%SZ")
+                    .map(|t| {
+                        let date_time = Utc.from_utc_datetime(&t);
+                        De::new(SystemTime::from(date_time))
+                    })
```
mrobinson, line 630: Even if you don't use the upstream serde deserializer, couldn't this use https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.parse_from_rfc3339?

components/net/tests/cookie.rs
Taym95: found the issue `is_cookie_expired` is returning `true` but should be `false`

components/hyper_serde/lib.rs
Taym95: I used `parse_from_rfc3339`

components/net/http_cache.rs
```diff
@@ -196,33 +195,26 @@ fn get_response_expiry(response: &Response) -> Duration {
...
+            let current = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default();
 
-            if desired > current {
-                return desired - current;
-            } else {
-                return Duration::seconds(0i64);
-            }
+            desired.checked_sub(current).unwrap_or(Duration::ZERO);
```
mrobinson, line 214: Looks like another `return` was lost.

components/net/http_cache.rs
```diff
@@ -196,33 +195,26 @@ fn get_response_expiry(response: &Response) -> Duration {
...
+            return Duration::ZERO;
         } else {
-            if let Some(secs) = directives.max_age().or(directives.s_max_age()) {
-                let max_age = Duration::from_std(secs).unwrap();
-                if max_age < age {
-                    return Duration::seconds(0i64);
-                }
-                return max_age - age;
+            if let Some(max_age) = directives.max_age().or(directives.s_max_age()) {
+                max_age.checked_sub(age).unwrap_or(Duration::ZERO);
```
mrobinson, line 201: I think a `return` was lost here.

components/net/http_cache.rs
```diff
@@ -280,24 +271,19 @@ fn get_expiry_adjustment_from_request_headers(request: &Request, expires: Durati
...
         }
         return expires - max_age;
     }
     if let Some(min_fresh) = directive.min_fresh() {
-        let min_fresh = Duration::from_std(min_fresh).unwrap();
-        if expires < min_fresh {
-            return Duration::min_value();
-        }
-        return expires - min_fresh;
+        expires.checked_sub(min_fresh).unwrap_or(Duration::ZERO);
```
mrobinson, line 283: This should `return` as well.

mrobinson (changes requested):
Looks great, just a few `return` statements were lost.

components/net/http_cache.rs
Taym95: Thanks, probably that why `request-cache-force-cache.any.worker.html` failed

mrobinson (approved):

mrobinson:
@Taym95 Thanks for you patience here! I'll try to land this now.

mrobinson:
Looks like there's a failure in the Mac unit tests:

```
---- cookie::test_sort_order stdout ----
thread 'cookie::test_sort_order' panicked at components/net/tests/cookie.rs:198:5:
  left: Equal
 right: Less
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:315:17
   4: main::cookie::test_sort_order
   6: core::ops::function::FnOnce::call_once
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5
```

sagudev:
This was reverted in [8e5f288](https://github.com/servo/servo/commit/8e5f28839cde6b9ee5cd7cb4f8c27ff0ae10a86c)

PR #31083, opened 2024-01-12
layout: Switch `IndependentLayout` to use `Au` instead of `Length`

atbrakhi:
---

- [x] These changes are part of #29819

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -243,7 +243,8 @@ impl BlockFormattingContext {
             fragments: flow_layout.fragments,
             content_block_size: (flow_layout.content_block_size +
                 flow_layout.collapsible_margins_in_children.end.solve() +
-                clearance.unwrap_or_else(Length::zero)).into(),
+                clearance.unwrap_or_else(Length::zero))
+            .into(),
             last_inflow_baseline_offset: flow_layout.last_inflow_baseline_offset.map(|t| t.into()),
         }
     }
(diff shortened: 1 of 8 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -1504,3 +1515,10 @@ impl PlacementState {
         )
     }
 }
+
+fn clamp_au(value: Au, min: Au, max: Option<Au>) -> Au {
+    match max {
+        Some(max_value) => value.min(max_value).max(min),
+        None => value.max(min),
+    }
+}
```
mrobinson, line 1524: Can you add this as a trait on `Au` to `style_ext.rs` and also call it `fn clamp_between_extremums(...)`. I think that's a slightly cleaner change as it will prevent code from having to change all over.

components/layout_2020/table/mod.rs
```diff
@@ -39,7 +39,7 @@ impl Table {
     ) -> IndependentLayout {
         IndependentLayout {
             fragments: Vec::new(),
-            content_block_size: Length::new(0.),
+            content_block_size: Au(0),
```
mrobinson, line 42: I would use `Au::zero()` here for clarity, if possible.

mrobinson (changes requested):
Looks pretty good to me, though I have a couple small suggestions.

components/layout_2020/flow/mod.rs
atbrakhi: thanks for the review. done in [523e7be](https://github.com/servo/servo/pull/31083/commits/523e7bef67e5e182746a471f524f9fff93e3c0ab)

mrobinson (approved):
Nice work!

PR #31086, opened 2024-01-15
Preliminary Android build support

mukilan:
The built APK still doesn't run on device and it crashes on loading.
This is happening only after the rust compiler toolchain update. I
am still investigating this, but it is proving to be challenging as
the generated stack trace doesn't show function names even
in debug build and after ensuring Gradle doesn't strip the symbols.

The goal with this PR is to land the changes with just the CI build
passing.

Other issues that need to be resolved in future PRs
1. Enable GStreamer support - this is disabled for now
2. Enable bluetooth support - blurdroid is unmaintained and doesn't
    compile now.
3. Revisit issues involving dependencies - jemalloc-sys, freetype and
    fontconfig
4. Revisit compositing logic & event handling. Apps can lose underlying
    GL Surface when they are sent to background. I have added logic to
    pause/resume the compositor when this happens so that we bind/unbind.
    Scrolling is also now 'janky' after the move from WebRender based zoom
    to Servo handling the zoom with root pipeline.

---

- [x] These changes do not require tests because these changes are for preliminary support. CI sanity check of android is disabled for now as apk crashes and sanity test code is also incorrect.


ports/libsimpleservo/jniapi/build.rs
```diff
@@ -3,31 +3,20 @@
...
         env::var("OUT_DIR").expect("Cargo should have set the OUT_DIR environment variable");
 
-    println!("cargo:rustc-link-lib=static=android_native_app_glue");
+    // FIXME: we need this workaround since jemalloc-sys still links
+    // to libgcc instead of libunwind, but Android NDK 23c and above
+    // don't have libgcc. Alternatively, we could disable jemalloc
+    // for android in servo_allocator
+    let mut libgcc = File::create(out_dir.clone() + "/libgcc.a").unwrap();
+    libgcc.write_all(b"INPUT(-lunwind)").unwrap();
+
```
mrobinson, line 20: I think it would simplify the build to just disable jemalloc on Android.

ports/libsimpleservo/jniapi/src/lib.rs
```diff
@@ -5,28 +5,44 @@
...
 use std::os::raw::{c_char, c_int, c_void};
-use std::ptr::{null, null_mut};
 use std::sync::Arc;
 use std::thread;
 
-use android_logger::{self, Filter};
-use gstreamer::debug_set_threshold_from_string;
+use android_logger::{self, FilterBuilder, Config};
+// FIXME: GStreamer support
+// use gstreamer::debug_set_threshold_from_string;
```
mrobinson, line 13: Instead of commenting out the code, please remove it.

ports/libsimpleservo/jniapi/src/lib.rs
```diff
@@ -79,26 +96,31 @@ pub fn Java_org_mozilla_servoview_JNIServo_init(
...
             }
         }
 
-        if let Some(gst_debug_str) = gst_debug_str {
-            debug_set_threshold_from_string(&gst_debug_str, true);
-        }
+        // FIXME: GStreamer support
+        // if let Some(gst_debug_str) = gst_debug_str {
+        //   debug_set_threshold_from_string(&gst_debug_str, true);
+        // }
```
mrobinson, line 112: Please just remove this code instead of commenting it out.

python/servo/build_commands.py
```diff
@@ -123,7 +123,9 @@ def build(self, build_type: BuildType, jobs=None, params=None, no_package=False,
             for key in env:
                 print((key, env[key]))
 
-        self.download_and_build_android_dependencies_if_needed(env)
+        # FIXME: disabled until GStreamer support is enabled
+        # self.download_and_build_android_dependencies_if_needed(env)
+
```
mrobinson, line 128: Let's just remove this for now as well as `download_and_build_android_dependencies_if_needed()`.

python/servo/command_base.py
```diff
@@ -618,48 +605,31 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
...
 
-        host_cc = env.get('HOST_CC') or shutil.which(["clang"]) or util.whichget_exec_path(["gcc"])
-        host_cxx = env.get('HOST_CXX') or util.whichget_exec_path(["clang++"]) or util.whichget_exec_path(["g++"])
-
-        llvm_toolchain = path.join(env['ANDROID_NDK'], "toolchains", "llvm", "prebuilt", host)
-        gcc_toolchain = path.join(env['ANDROID_NDK'], "toolchains",
-                                  android_toolchain_prefix + "-4.9", "prebuilt", host)
-        gcc_libs = path.join(gcc_toolchain, "lib", "gcc", android_toolchain_name, "4.9.x")
+        host_cc = env.get('HOST_CC') or shutil.which("clang") or shutil.which(["gcc"])
+        host_cxx = env.get('HOST_CXX') or shutil.which("clang++") or shutil.which("g++")
```
mrobinson, line 609: I think we should remove the `gcc` and `g++` fallbacks here, as I don't know if Servo supports gcc any longer.

python/servo/command_base.py
```diff
@@ -671,52 +641,28 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
...
-            "--sysroot=" + env['ANDROID_SYSROOT'],
-            "-I" + arch_include,
-        ])
-        env["NDK_ANDROID_VERSION"] = android_api
+        env['CFLAGS'] = "--target=" + android_toolchain_name
+        env['CXXFLAGS'] = "--target=" + android_toolchain_name
+
+        # needed by mozjs
+        env['ANDROID_API_LEVEL'] = android_api
+        env["ANDROID_NDK_HOME"] = env["ANDROID_NDK_ROOT"]
```
mrobinson, line 649: Hrm. mozjs should probably switch to using the same variable as Servo or vice versa? Which is more common? It looks like both are used?

python/servo/command_base.py
```diff
@@ -671,52 +641,28 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
...
-        env['CPPFLAGS'] = ' '.join([
-            "--target=" + self.cross_compile_target,
-            "--sysroot=" + env['ANDROID_SYSROOT'],
-            "-I" + arch_include,
-        ])
-        env["NDK_ANDROID_VERSION"] = android_api
+        env['CFLAGS'] = "--target=" + android_toolchain_name
+        env['CXXFLAGS'] = "--target=" + android_toolchain_name
+
+        # needed by mozjs
```
mrobinson, line 647:
```suggestion
        # These two variables are needed for the mozjs compilation.
```

python/servo/command_base.py
```diff
@@ -671,52 +641,28 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
...
+        env['ANDROID_API_LEVEL'] = android_api
+        env["ANDROID_NDK_HOME"] = env["ANDROID_NDK_ROOT"]
+
+        # The two vars set below are passed by our custom
+        # support/android/toolchain.cmake to the NDK's cmake toolchain file
```
mrobinson, line 652:
```suggestion
        # support/android/toolchain.cmake to the NDK's CMake toolchain file.
```

python/servo/command_base.py
```diff
@@ -964,21 +906,18 @@ def run_cargo_build_like_command(
         return call(["cargo", command] + args + cargo_args, env=env, verbose=verbose)
 
     def android_support_dir(self):
-        return path.join(self.context.topdir, "support", "android")
-
-    def android_aar_dir(self):
-        return path.join(self.context.topdir, "target", "android", "aar")
+        return
```
mrobinson, line 909: If this is unused, please just remove it.

servobuild.example
```diff
@@ -56,8 +56,8 @@ media-stack = "auto"
...
 # Android information
 [android]
-# Defaults to the value of $ANDROID_SDK, $ANDROID_NDK, $ANDROID_TOOLCHAIN, $ANDROID_PLATFORM respectively
+# Defaults to the value of $ANDROID_SDK_ROOT, $ANDROID_NDK_ROOT, $ANDROID_TOOLCHAIN, $ANDROID_PLATFORM respectively
 #sdk = "/opt/android-sdk"
 #ndk = "/opt/android-ndk"
-#toolchain = "/opt/android-toolchain"
-#platform = "android-18"
+#toolchain = "/opt/android-toolchain" FIXME: still needed?
+#platform = "android-18" FIXME: still needed?
```
mrobinson, line 63: If these aren't read, let's just remove them.

support/android/apk/gradlew
```diff
@@ -150,11 +154,19 @@ if $cygwin ; then
...
-JVM_OPTS[${#JVM_OPTS[*]}]="-Dorg.gradle.appname=$APP_BASE_NAME"
+APP_ARGS=$(save "$@")
+
+# Collect all arguments for the java command, following the shell quoting and substitution rules
+eval set -- $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS "\"-Dorg.gradle.appname=$APP_BASE_NAME\"" -classpath "\"$CLASSPATH\"" org.gradle.wrapper.GradleWrapperMain "$APP_ARGS"
+
+# by default we should be in the correct project dir, but when run from Finder on Mac, the cwd is wrong
+if [ "$(uname)" = "Darwin" ] && [ "$HOME" = "$PWD" ]; then
+  cd "$(dirname "$0")"
+fi
```
mrobinson, line 170: Is this code all from Gradle?

support/android/apk/jni/Android.mk
```diff
@@ -16,13 +16,13 @@
...
-include $(CLEAR_VARS)
-LOCAL_PATH:= $(SERVO_TARGET_DIR)/../../gstreamer/gst-build-$(APP_ABI)
-LOCAL_MODULE := gstreamer
-LOCAL_SRC_FILES := libgstreamer_android.so
-include $(PREBUILT_SHARED_LIBRARY)
+# include $(CLEAR_VARS)
+# LOCAL_PATH:= $(SERVO_TARGET_DIR)/../../gstreamer/gst-build-$(APP_ABI)
+# LOCAL_MODULE := gstreamer
+# LOCAL_SRC_FILES := libgstreamer_android.so
+# include $(PREBUILT_SHARED_LIBRARY)
```
mrobinson, line 28: Let's just remove this for now.

support/android/apk/jni/Application.mk
```diff
@@ -1,5 +1,6 @@
 NDK_TOOLCHAIN_VERSION := clang
-APP_MODULES := c++_shared servojni gstreamer
-APP_PLATFORM := android-21
-APP_STL:= c++_shared
-APP_ABI:= armeabi-v7a x86
+APP_MODULES := c++_shared servojni
+#APP_MODULES := c++_shared servojni #gstreamer
```
mrobinson, line 3: Ditto.

support/android/apk/servoapp/src/main/java/org/mozilla/servo/MainActivity.java
```diff
@@ -78,8 +78,8 @@ protected void onCreate(Bundle savedInstanceState) {
         Intent intent = getIntent();
         String args = intent.getStringExtra("servoargs");
         String log = intent.getStringExtra("servolog");
-        String gstdebug = intent.getStringExtra("gstdebug");
-        mServoView.setServoArgs(args, log, gstdebug);
+        // TODO String gstdebug = intent.getStringExtra("gstdebug");
```
mrobinson, line 81: Ditto.

support/android/apk/servoview/build.gradle
```diff
@@ -145,34 +108,51 @@ android {
...
     }
 
-
-    // Call our custom NDK Build task using flavor parameters
+    // Call our custom NDK Build task using flavor parameters.
+    // This step is needed because the Android Gradle Plugin system
+    // integration with native C/C++ shared objects (based on the
+    // `android.externalNativeBuild` dsl object) assumes that we
+    // actually execute compiler commands to produce the shared
+    // objects. We already have a libsimpleservo.so produce by rustc.
```
mrobinson, line 116:
```suggestion
    // objects. We already have a libsimpleservo.so produced by rustc.
```

support/android/apk/servoview/build.gradle
```diff
@@ -145,34 +108,51 @@ android {
...
+    //
+    // FIXME(mukilan): According to the AGP docs, we should
+    // not be relying on task names used by the plugin system
+    // to 'hook' into the build process, but instead we should
+    // use the `androidComponents.beforeVariants` and similar
+    // officially supported extension points.
     tasks.all {
         compileTask ->
-            Pattern pattern = Pattern.compile(/^compile[A-Z][\w\d]+([A-Z][\w\d]+)(Debug|Release)Ndk/)
+            // match task `mergeBasicArmv7DebugJniLibFolders`
```
mrobinson, line 131:
```suggestion
            // This matches task `mergeBasicArmv7DebugJniLibFolders`.
```

support/android/apk/servoview/build.gradle
```diff
@@ -145,34 +108,51 @@ android {
...
-                        'NDK_LIBS_OUT=' + getJniLibsPath(debug, arch),
-                        'NDK_OUT=' + getTargetDir(debug, arch) + '/apk/obj',
-                        'NDK_DEBUG=' + (debug ? '1' : '0'),
-                        'APP_ABI=' + getNDKAbi(arch),
-                        'SERVO_TARGET_DIR=' + getTargetDir(debug, arch)
+                commandLine getNdkDir() + "/ndk-build",
+                    'APP_BUILD_SCRIPT=../jni/Android.mk',
+                    'NDK_APPLICATION_MK=../jni/Application.mk',
+                    'NDK_LIBS_OUT=' + getJniLibsPath(debug, arch),
+                    //'NDK_OUT=' + getJniLibsPath(debug, arch),
```
mrobinson, line 146: Please remove.

support/android/apk/servoview/src/main/java/org/mozilla/servoview/JNIServo.java
```diff
@@ -6,21 +6,22 @@
...
 /**
  * Maps /ports/libsimpleservo API
  */
 @SuppressWarnings("JniMissingFunction")
 public class JNIServo {
     JNIServo() {
         System.loadLibrary("c++_shared");
-        System.loadLibrary("gstreamer_android");
+        // FIXME: GStreamer support
+        // System.loadLibrary("gstreamer_android");
```
mrobinson, line 18: Please just remove this.

support/android/apk/servoview/src/main/java/org/mozilla/servoview/Servo.java
```diff
@@ -30,18 +31,20 @@ public Servo(
...
 
         mRunCallback.inGLThread(() -> {
-            mJNI.init(activity, options, mServoCallbacks);
+            mJNI.init(activity, options, mServoCallbacks, surface);
         });
 
         try {
-          GStreamer.init((Context) activity);
+          // FIXME: GStreamer support
+          // GStreamer.init((Context) activity);
```
mrobinson, line 47: Ditto.

support/android/apk/servoview/src/main/java/org/mozilla/servoview/ServoView.java
```diff
@@ -154,51 +126,9 @@ public void animationStateChanged(boolean animating) {
...
-
-    private void initGestures(Context context) {
-        mGestureDetector = new GestureDetector(context, this);
-        mScaleGestureDetector = new ScaleGestureDetector(context, this);
-        mScroller = new OverScroller(context);
-    }
 
     private void startLooping() {
-      // In case we were already drawing.
+      //In case we were already drawing.
```
mrobinson, line 131: I think this change is unintentional.

third_party/webrender/webrender/Cargo.toml
```diff
@@ -9,7 +9,7 @@ build = "build.rs"
 edition = "2018"
 
 [features]
-default = ["freetype-lib"]
+default = ["freetype-lib", "no_static_freetype"]
```
mrobinson, line 12: This feature should probably not be added unconditionally, but added via a platform-specific dependency in the compositor's Cargo.toml.

.github/workflows/android.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: python3 -m pip install --upgrade pip virtualenv
+      - name: Bootstrap dependencies
+        run: sudo apt update && python3 ./mach bootstrap
+      - name: Build (arch ${{ matrix.arch }} profile ${{ inputs.profile }})
+        run: |
+          export JAVA_HOME="${JAVA_HOME_17_X64}"
+          python3 ./mach build --android --target ${{ matrix.arch }} --${{ inputs.profile }}
+      # TODO: fix test-android-startup
+      # - name: Script tests
+      #   run: ./mach test-android-startup
```
mrobinson, line 62: Would it be possible to open a bug for this issue and link here?

```suggestion
      # TODO: Android currently crashes on startup, so this is disabled currently.
      # https://github.com/servo/servo/issue...
      # - name: Script tests
      #   run: ./mach test-android-startup
```

.github/workflows/android.yml
```diff
@@ -0,0 +1,91 @@
...
+        with:
+          crate: taplo-cli
+          locked: true
+      - name: Bootstrap Python
+        run: python3 -m pip install --upgrade pip virtualenv
+      - name: Bootstrap dependencies
+        run: sudo apt update && python3 ./mach bootstrap
+      - name: Build (arch ${{ matrix.arch }} profile ${{ inputs.profile }})
+        run: |
+          export JAVA_HOME="${JAVA_HOME_17_X64}"
```
mrobinson, line 58: Don't we need to install a specific version of the SDK and NDK here?

README.md
```diff
@@ -116,6 +116,27 @@ though of course it doesn’t produce a binary you can run.
 
 ### Building for Android target
 
+Prerequisites:
+Servo's build system assumes that both the Android SDK & NDK are
```
mrobinson, line 120: It'd be great to have a small set of Android-specific instructions here like we have for other platforms. I'm happy to write this.

components/compositing/compositor.rs
```diff
@@ -501,6 +501,26 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+        debug!("Pausing compositor");
+        if let Err(e) = self.webrender_surfman.unbind_native_surface_from_context() {
+            warn!("Unbinding native surface from context failed ({:?})", e);
+        }
+    }
+
+    #[allow(unsafe_code)]
+    pub fn resume(&mut self, native_widget: *mut c_void, coords: DeviceIntSize) {
+        debug!("Resuming compositor: {native_widget:?}");
+        // Reinitialize webrender surfman and bind to context
```
mrobinson, line 514: Can you expand this comment into a full rustdoc for this function? I'm not entirely sure why this new functionality is required in the compositor.

components/compositing/compositor.rs
```diff
@@ -501,6 +501,26 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         self.shutdown_state = ShutdownState::FinishedShuttingDown;
     }
 
+    pub fn pause(&mut self) {
```
mrobinson, line 504: It might be nice to have some rustdoc here on this function.

components/compositing/windowing.rs
```diff
@@ -105,6 +106,10 @@ pub enum EmbedderEvent {
     ChangeBrowserVisibility(TopLevelBrowsingContextId, bool),
     /// Virtual keyboard was dismissed
     IMEDismissed,
+    /// Pause the compositor when underlying native surface is lost
```
mrobinson, line 109:
```suggestion
    /// Pause the compositor when underlying native surface is lost.
```

components/compositing/windowing.rs
```diff
@@ -105,6 +106,10 @@ pub enum EmbedderEvent {
...
+    PauseCompositor,
+    /// Resume the compositor when underlying native surface is regained
```
mrobinson, line 111:
```suggestion
    /// Resume the compositor when underlying native surface is regained.
```

components/compositing/windowing.rs
```diff
@@ -105,6 +106,10 @@ pub enum EmbedderEvent {
     ChangeBrowserVisibility(TopLevelBrowsingContextId, bool),
     /// Virtual keyboard was dismissed
     IMEDismissed,
+    /// Pause the compositor when underlying native surface is lost
+    PauseCompositor,
```
mrobinson, line 110: It might be worth renaming these to make it clearer what these are handling. Perhaps something like:

```
InvalidateNativeSurface,
```

components/compositing/windowing.rs
```diff
@@ -105,6 +106,10 @@ pub enum EmbedderEvent {
...
+    /// Resume the compositor when underlying native surface is regained
+    ResumeCompositor(*mut c_void, DeviceIntSize)
```
mrobinson, line 112: This could be something like:

```
ReplaceNativeSurface(*mut c_void, DeviceIntSize)
```

components/compositing/compositor.rs
```diff
@@ -501,6 +501,26 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+        // Reinitialize webrender surfman and bind to context
+        let connection = self.webrender_surfman.connection();
+        let native_widget = unsafe {
+            connection.create_native_widget_from_ptr(native_widget, coords.to_untyped())
+        };
+        if let Err(e) = self.webrender_surfman.bind_native_surface_to_context(native_widget) {
+            warn!("Binding native surface to context failed ({:?})", e);
+        }
+    }
+
```
mrobinson, line 523: Hrm...so normally the embedder is responsible for binding the native widget to the context, right? Is there a way to rework this so that this always happens in the embedder instead of making this happen in WebRender. Perhaps we could add a method to unset `webrender_surfman` in the compositor and then reset it from the embedder?

ports/libsimpleservo/jniapi/Cargo.toml
```diff
@@ -14,21 +14,27 @@ test = false
 bench = false
 
 [dependencies]
-android_injected_glue = "0.2"
-android_logger = "0.10"
-gstreamer = { workspace = true }
+android_logger = "0.13"
+# gstreamer = { workspace = true }
```
mrobinson, line 18: Please remove this.

ports/libsimpleservo/jniapi/Cargo.toml
```diff
@@ -14,21 +14,27 @@ test = false
...
 log = { workspace = true }
 serde_json = { workspace = true }
-simpleservo = { path = "../api" }
+
+# TODO: enable native-bluetooth feature for android
+# and remove explicit feature list here
+[dependencies.simpleservo]
+path = "../api"
+default-features = false
+features = ["max_log_level", "webdriver"]
```
mrobinson, line 29: I think it makes sense to keep using the one line dependency format, which we use pretty consistently in Servo.

ports/libsimpleservo/jniapi/Cargo.toml
```diff
@@ -14,21 +14,27 @@ test = false
...
+android_logger = "0.13"
+# gstreamer = { workspace = true }
 jni = "0.18.0"
 libc = { workspace = true }
 log = { workspace = true }
 serde_json = { workspace = true }
-simpleservo = { path = "../api" }
+
+# TODO: enable native-bluetooth feature for android
+# and remove explicit feature list here
```
mrobinson, line 25:
```suggestion
# TODO: Once the native-bluetooth feature works for Android, remove the explicit feature list here.
```

ports/libsimpleservo/jniapi/Cargo.toml
```diff
@@ -14,21 +14,27 @@ test = false
...
+default-features = false
+features = ["max_log_level", "webdriver"]
 
 [build-dependencies]
 cc = "1.0"
 
 [features]
 debugmozjs = ["simpleservo/debugmozjs"]
-default = ["max_log_level", "native-bluetooth", "webdriver"]
+# TODO: enable native-bluetooth feature for android
```
mrobinson, line 36: I think it's fine to remove this comment.

mrobinson (changes requested):
Great work! I'm so glad to see this happening. I have a bunch of comments below, but this is already looking very nice.

ports/libsimpleservo/jniapi/build.rs
mukilan: The comment is outdated. I did try disabling jemalloc but that requires providing an implementation of [this](https://github.com/servo/servo/blob/main/components/allocator/lib.rs#L39) method. I couldn't find a way to do that with the standard allocator.

python/servo/command_base.py
mukilan: `ANDROID_NDK_ROOT` is  the official one now, so mozjs should be changed. Can that land separately?

support/android/apk/gradlew
mukilan: Yep, this generated by Gradle to bootstrap itself on any system.

.github/workflows/android.yml
mukilan: We are relying on the SDK and NDK preinstalled in the [runner image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#environment-variables-2)

mrobinson: Let's use actions like "nttld/setup-ndk" that we use for mozjs to ensure that the proper version is installed. If the runner image changes, this image this might break one day in the future.

python/servo/command_base.py
mrobinson: Sure. I can make the fix in mozjs.

ports/libsimpleservo/jniapi/build.rs
mrobinson: Could you use `malloc_usable_size`? https://docs.rs/libc/latest/libc/fn.malloc_usable_size.html

mukilan: I did look into that as well, but the documentation [warns](https://doc.rust-lang.org/std/alloc/struct.System.html) against directly using the backend allocator methods, but I can switch to `malloc_usable_size` we think it should be safe in this specific case.

components/compositing/compositor.rs
mukilan: I'm not sure if I fully understand the suggestion. 

> instead of making this happen in WebRender.

Do you mean in WebRenderSurfman ?

Today, AFAICT, the embedder doesn't  bind the widget directly, but it simply does `WebrenderSurfman::create` during initialization and then holds on to that instance for the lifetime. Are you suggesting that we move away from that model and make embedder recreate `WebrenderSurfman` on demand? Today  `WebrenderSurfman::create` also creates other resources like handle to the device, opengl context and swapchain - are we fine with recreating all these other resources as well?

Also, today the compositor  holds on the same WebrenderSurfman instance that the embedder creates (The data is wrapped in Rc and WindowMethods trait allows libservo to create a new handle from an open window). If we reset webrender_surfman in embedder, wouldn't we also need to reset it from compositor?

components/compositing/windowing.rs
mukilan: FWIW, I wasn't sure about the correctness of the current way of handling this and suspected that eventually this will include more work such as pausing the script thread, animation etc in addition to binding/unbinding surface. But I'll change it to the suggested names for now and we can rename it later.

components/compositing/compositor.rs
mrobinson: I see. Thanks for the explanation! I think this approach is okay for now! We probably need to take a holistic look at how all these surfaces are managed at some point in the future.

README.md
mukilan: I am happy to write this.  Also, there is a link at the top of this README to servo wiki page with more [detailed instructions](https://github.com/servo/servo/wiki/Building-for-Android), but this is out of date. Do we want to add Android-specific instructions here or keep them in the wiki?

components/compositing/compositor.rs
mukilan: Added rustdoc comment on the renamed method `replaced_native_surface`

.github/workflows/android.yml
```diff
@@ -0,0 +1,107 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: android-${{ matrix.arch }}-${{ inputs.profile }}
+          path: target/android/${{ matrix.arch }}/${{ inputs.profile }}/servoapp.apk
+
+  result:
+    name: Result
+    runs-on: ubuntu-latest
+    if: always()
+    # needs all build to detect cancellation
```
mrobinson, line 97:
```suggestion
    # `needs: 'build'` is necessary to detect cancellation.
```

components/compositing/compositor.rs
```diff
@@ -501,6 +501,33 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
         self.shutdown_state = ShutdownState::FinishedShuttingDown;
     }
 
+    /// The underlying native surface can be lost during servo's lifetime.
+    /// On android, for example, this happens when the app is sent to background.
```
mrobinson, line 505:
```suggestion
    /// On Android, for example, this happens when the app is sent to background.
```

components/compositing/compositor.rs
```diff
@@ -501,6 +501,33 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+    /// We need to unbind the surface so that we don't try to use it again.
+    pub fn invalidate_native_surface(&mut self) {
+        debug!("Invalidating native surface in compositor");
+        if let Err(e) = self.webrender_surfman.unbind_native_surface_from_context() {
+            warn!("Unbinding native surface from context failed ({:?})", e);
+        }
+    }
+
+    /// On android, this function will be called when the app moves to foreground
```
mrobinson, line 514:
```suggestion
    /// On Android, this function will be called when the app moves to foreground
```

components/compositing/windowing.rs
```diff
@@ -105,6 +106,14 @@ pub enum EmbedderEvent {
     ChangeBrowserVisibility(TopLevelBrowsingContextId, bool),
     /// Virtual keyboard was dismissed
     IMEDismissed,
+    /// Sent on platforms like android where the native widget surface can be
```
mrobinson, line 109:
```suggestion
    /// Sent on platforms like Android where the native widget surface can be
```

components/compositing/windowing.rs
```diff
@@ -105,6 +106,14 @@ pub enum EmbedderEvent {
...
+    /// automatically destroyed by the system, for example when the app
+    /// is sent to background.
+    InvalidateNativeSurface,
+    /// Sent on platforms like android where system recreates a new surface for
```
mrobinson, line 113:
```suggestion
    /// Sent on platforms like Android where system recreates a new surface for
```

components/servo/lib.rs
```diff
@@ -1091,6 +1098,9 @@ fn default_user_agent_string_for(agent: UserAgent) -> &'static str {
     const DESKTOP_UA_STRING: &'static str =
         "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Servo/1.0 Firefox/111.0";
 
+    #[cfg(target_os = "android")]
```
mrobinson, line 1101: It might be a good moment to clean this up and just conditionally define a `UA_STRING` and always return it.

ports/libsimpleservo/jniapi/build.rs
mrobinson: Yeah, let's just use the backend allocator methods and disable jemaloc. I think it's fine in this case.

python/servo/command_base.py
```diff
@@ -671,52 +636,28 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
...
-        env["NDK_ANDROID_VERSION"] = android_api
+        env['CFLAGS'] = "--target=" + android_toolchain_name
+        env['CXXFLAGS'] = "--target=" + android_toolchain_name
+
+        # These two variables are needed for the mozjs compilation.
+        env['ANDROID_API_LEVEL'] = android_api
+        env["ANDROID_NDK_HOME"] = env["ANDROID_NDK_ROOT"]
+
+        # The two variables set below are passed by our custom
+        # support/android/toolchain.cmake to the NDK's Cmake toolchain file
```
mrobinson, line 647:
```suggestion
        # support/android/toolchain.cmake to the NDK's CMake toolchain file
```

mrobinson (approved):
Nice! So happy to see this finally landing. Just a couple small things before landing and also let's go ahead and disable jemalloc for Android.

components/servo/lib.rs
mukilan: Hmm, the existing code is written so that the user can switch user agent to ios/android specific ones even on desktop at runtime using [this](https://github.com/servo/servo/blob/main/ports/servoshell/main2.rs#L38) flag, probably to allow testing.

I think implementing what you suggested would require removing this functionality from libservo (maybe we can do it to ports/servoshell). Can this be tackled in a separate PR?

mrobinson: Yes, sure!

mukilan:
Thanks for the review!

PR #31088, opened 2024-01-15
deps: Raise the Python requirement to 3.10

mrobinson:
This will allow us to use the pip's new dependency resolver.

---

- [x] These changes do not require tests because this doesn't change behavior.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,7 +10,7 @@ jobs:
     # This job is only useful when run on upstream servo.
     if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Upload docs to GitHub Pages
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly-rust.yml
sagudev, line 1: Do we need this workflow at all?

.github/workflows/docs.yml
```diff
@@ -15,6 +15,9 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 2
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
sagudev, line 20: Can we just change 20.04 to 22.04?

.github/workflows/nightly-rust.yml
mrobinson: I don't think we do! I'll remove this workflow in a separate PR.

sagudev (approved):

PR #31090, opened 2024-01-15
crown: Add a cargo config.toml file

mrobinson:
This makes it more foolproof to install crown from inside the Servo
directory, because the root Servo config.toml overrides the rustc to use
crown (an obvious circular dependency).

---


sagudev (approved):
I am somewhat surprised that this works, but I am glad it does.

mrobinson:
Thanks for the review! I think this works because Cargo looks in every directory for the configuration file: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

PR #31091, opened 2024-01-15
ci: Remove the nightly-rust workflow

mrobinson:
This workflows tested Servo with the most recent nightly version of
Servo. Now that Servo compiles against a stable version of Rust, this
isn't so interesting. We are hoping to eliminate all use of unstable
features in the future.

---


sagudev (approved):
I think there is still something left-overs left in mach for update-nightly command, but we could just change it into update-rust for update to newer stable version of rust (and then this could be run monthly in CI).

PR #31092, opened 2024-01-15
build: Clean up post-build copy of Windows DLLs

mrobinson:
- No longer use vcvarsall.bat at all. Instead find the Windows SDK
  directory by looking in the registry.
- Split logic for copying Windows dependencies into its own function and
  do some minor clean up, such as collecting all MSVC functionality into
  visual_studio.py.
- Remove support for Visual Studio 2015 and Visual Studio 2017.

This is a preparatory change in order to support Visual Studio 2022.

---

- [x] These changes do not require tests because they are a build adjustment.

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -119,7 +119,8 @@ class MachCommands(CommandBase):
                     return rv
 
             if sys.platform == "win32":
-                copy_windows_dlls_to_build_directory(built_binary, target_triple)
+                if not copy_windows_dlls_to_build_directory(built_binary, target_triple):
+                    status = 1
 
             elif sys.platform == "darwin":
                 servo_path = self.get_binary_path(
@@ -362,7 +363,7 @@ def package_gstreamer_dylibs(cross_compilation_target, servo_bin):
     return True
 
 
-def copy_windows_dlls_to_build_directory(servo_binary: str, target_triple: str):
+def copy_windows_dlls_to_build_directory(servo_binary: str, target_triple: str) -> bool:
     servo_exe_dir = os.path.dirname(servo_binary)
     assert os.path.exists(servo_exe_dir)
 
@@ -389,6 +390,8 @@ def copy_windows_dlls_to_build_directory(servo_binary: str, target_triple: str):
     print(" • Copying MSVC DLLs to binary directory...")
     if not package_msvc_dlls(servo_exe_dir, target_triple):
         return False
+
+    return True
 
 
 def package_gstreamer_dlls(servo_exe_dir: str, target: str):
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -8,9 +8,9 @@
 # except according to those terms.
 
 import datetime
-import locale
 import os
 import os.path as path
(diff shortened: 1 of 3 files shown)


mrobinson:
This needs a little work. I'll address the failures tomorrow.

mrobinson:
Okay. This is ready to review!

python/servo/visual_studio.py
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+
+def find_windows_sdk_installation_path(vs_platform: str) -> str:
+    # This module must be imported here, because other platforms also
+    # load this file and the module is platform-specific.
+    import winreg
+
+    """Try to find the Windows SDK installation path using the Windows registry.
+    Raises and Exception if the path cannot be found in the registry."""
```
mukilan, line 87: `Raises an exception`?

python/servo/visual_studio.py
```diff
@@ -0,0 +1,144 @@
...
+
+def find_highest_msvc_version():
+    editions = ["Enterprise", "Professional", "Community", "BuildTools"]
+    prog_files = os.environ.get("ProgramFiles(x86)")
+    base_vs_path = os.path.join(prog_files, "Microsoft Visual Studio")
+
+    # TODO(mrobinson): Add support for Visual Studio 2022.
+    vs_versions = ["2019"]
+    versions = {
+        ("2019", "vs"): "16.0",
```
mukilan, line 46: Is the second component needed in the key tuple? It seems be hardcoded in line 51?

mrobinson: I don't think this is necessary. This was probably left over from some dead code. I'm happy to remove it.

python/servo/build_commands.py
```diff
@@ -423,7 +362,36 @@ def package_gstreamer_dylibs(cross_compilation_target, servo_bin):
...
+        except StopIteration:
+            print(f"WARNING: could not find {dll_name}")
+
+    print(" • Copying ANGLE DLLs to binary directory...")
+    find_and_copy_built_dll("libEGL.dll")
+    find_and_copy_built_dll("libGLESv2.dll")
+
+    print(" • Copying GStreamer DLLs to binary directory...")
+    if not package_gstreamer_dlls(servo_exe_dir, target_triple):
+        return False
```
mukilan, line 387: We return False here, but the caller doesn't seem to check this return value. Also the type of this function doesn't specify that we are returning a boolean.

python/servo/visual_studio.py
mrobinson: Thanks! I've fixed this.

python/servo/visual_studio.py
mrobinson: I've removed this as well as cleaned up a bit some of the surrounding code, returning a `dataclass` from the caller, which should help ensure type safety.

python/servo/build_commands.py
mrobinson: I've marked this method as returning a boolean and properly handle the result in the caller now.

mukilan (approved):

PR #31095, opened 2024-01-15
ci: Set LLVM_PATH when building docs

sagudev:
This is needed on 22.04 runners (that we started using in https://github.com/servo/servo/pull/31088).

Fixes docs building.

---

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,6 +20,8 @@ jobs:
           python3 -m pip install --upgrade pip
           sudo apt update
           python3 ./mach bootstrap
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -20,6 +20,8 @@ jobs:
           python3 -m pip install --upgrade pip
           sudo apt update
           python3 ./mach bootstrap
+      - name: Set LIBCLANG_PATH env # needed for bindgen in mozangle
```
mrobinson, line 23:
```suggestion
      - name: Set LIBCLANG_PATH # This is needed for bindgen in mozangle.
```

mrobinson (approved):

PR #31108, opened 2024-01-17
Implement ResizeObserver infra, support content box calculations. 

### LINKED ISSUE #31006: Implement ResizeObserver
Needed for https://github.com/servo/servo/issues/30822

Spec: https://drafts.csswg.org/resize-observer/#resize-observer-interface

One would have to add a new WebIDL(see `components/script/dom/webidl`), and expose the interface on `Window`. 

And then implement it in a new `components/script/dom/resizeobserver.rs` file. The WebIDL interface methods will be generated automatically and can be imported as `use crate::dom::bindings::codegen::Bindings::<interface name>`

For the callback part, `components/script/dom/eventtarget.rs` is probably a good place to look for inspiration.

gterzian:
Implement the basic infra for the [ResizeObserver](https://drafts.csswg.org/resize-observer/) Web API, and support size calculations for content boxes. 

---

- [ ] These changes fix #31006 (GitHub issue number if applicable)


gterzian:
[Those two](https://github.com/servo/servo/pull/31108#issuecomment-1911589318) appear to be intermittent, how do I mark them as such again?

I have filed an issue for tests that are still failing: https://github.com/servo/servo/issues/31182

The base case appears to work, there is some intermittency which appears related to the delivery of rafs.

gterzian:
@mrobinson @jdm @emilio  This one is ready for review. See https://github.com/servo/servo/issues/31182 for the currently failing tests(usually it's one sub-test failing in the test)

Are the [tests failure](https://github.com/servo/servo/actions/runs/7695647188/job/20968986306#step:5:16) related to the switch of the main branch? 

> python: can't open file '/home/runner/work/servo/servo/./python/servo/try_parser.py': [Errno 2] No such file or directory

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,260 @@
...
+    ResizeObservationDepth(depth)
+}
+
+/// <https://drafts.csswg.org/resize-observer/#calculate-box-size>
+fn calculate_box_size(target: &Element, observed_box: &ResizeObserverBoxOptions) -> Rect<Au> {
+    // TODO: batch into one layout query for list of elems?
+    match observed_box {
+        ResizeObserverBoxOptions::Content_box |
+        ResizeObserverBoxOptions::Device_pixel_content_box => {
+            target.upcast::<Node>().bounding_content_box_or_zero()
```
Loirooriol, line 255: Device_pixel_content_box should snap to device pixels.
Also, `bounding_content_box_or_zero` is the wrong choice if there are multiple fragments, we don't want the bounding rect. As per the current spec it should just be the 1st fragment, with the idea of extending to an array with all fragments in the future.

gterzian: Thanks! I've added your comment to https://github.com/servo/servo/issues/31182 because I think some of these might best be addressed as follow-ups, so that this PR can focus more on the overal mechanism and get the event-loop integration right. 

Would you happen to have links to the code where one could get the right layout queries or transform the resulting data into the right format?

gterzian:
I'm looking a bit more into this raf issue that makes a sub test fail, I guess it would be nice to have the entire `event-loop.html` test pass. Currently failing is one where it expects to have received a raf but it hasn't....

Feel free to review in the meantime...

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,260 @@
...
+/// https://drafts.csswg.org/resize-observer/#resizeobservation
+#[derive(JSTraceable, MallocSizeOf)]
+struct ResizeObservation {
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-target>
+    /// Note: `target` is kept out of here, to avoid having to root the `ResizeObservation`.
+
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-observedbox>
+    observed_box: ResizeObserverBoxOptions,
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-lastreportedsizes>
+    last_reported_sizes: DomRefCell<VecDeque<ResizeObserverSizeImpl>>,
```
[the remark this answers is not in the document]
gterzian: Yeah nothing... I was thinking of making this just an option, because only the last one is ever read. What do you think?

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,260 @@
...
+                // TODO: border box.
+                &[],
+                &[&*observer_size],
+                &[&*observer_size],
+            );
+            entries.push(entry);
+            observation
+                .last_reported_sizes
+                .borrow_mut()
+                .push_back(size_impl);
```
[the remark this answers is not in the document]
gterzian: I'm using the equivalent that isn't a dom object, so that `ResizeObservation` doesn't have to be a dom object either(which would with our codegen require exposing it to JS).

gterzian:
Regarding the `event-loop.html` test: it does not receive a raf when it expects to. The tests [acknowledges](https://github.com/servo/servo/blob/9b6c473695e14c1a37dd70325657519b901e7efc/tests/wpt/tests/resize-observer/eventloop.html#L91) this potential issue, and tries to work around it, but I guess in our case this is not enough. I tried moving around the raf code in the script-thread and the test would pass, but I don't think we should look too much into this now. [test0](https://github.com/servo/servo/blob/9b6c473695e14c1a37dd70325657519b901e7efc/tests/wpt/tests/resize-observer/eventloop.html#L45) and [test2](https://github.com/servo/servo/blob/9b6c473695e14c1a37dd70325657519b901e7efc/tests/wpt/tests/resize-observer/eventloop.html#L196) pass, with the exeption of the raf count subtest. test1 seems to test shadow dom things, and it times out.

components/script/dom/resizeobserver.rs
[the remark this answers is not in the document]
emilio: An option is fine as long as we only report one size at a time. you really want to set this to the array we last reported, rather than keeping appending.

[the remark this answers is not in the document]
gterzian: Ah ok, now I understand: it's the array for multiple fragments for a single reported size, where currently one fragment is used. I thought it was more of a stack of reported sizes... thanks!

gterzian:
@Loirooriol @emilio Thanks to the both of your for your initial comments, which have been addressed. 

Could I get a comprehensive review please(also perhaps from @mrobinson @jdm if you have time)? 

I think the current mechanism works, with two seperate areas that needs to be addressed:
1. Various size calculations are off, as pointed out by @Loirooriol 
2. We fail tests relying on proper integration with the raf callbacks. 

Both points are addressed in https://github.com/servo/servo/issues/31182

Regarding https://github.com/servo/servo/issues/31242, on which 2 above depends: I think merging this initial work is an excellent way to get feedback while trying to fix the event-loop in a follow-up, see the [currently failing test](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/tests/wpt/tests/resize-observer/eventloop.html#L83).

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+
+/// <https://drafts.csswg.org/resize-observer/#calculate-depth-for-node>
+fn calculate_depth_for_node(target: &Element) -> ResizeObservationDepth {
+    let node = target.upcast::<Node>();
+    let depth = node.ancestors().count();
+    ResizeObservationDepth(depth)
+}
+
+/// <https://drafts.csswg.org/resize-observer/#calculate-box-size>
+fn calculate_box_size(target: &Element, observed_box: &ResizeObserverBoxOptions) -> Rect<Au> {
```
gterzian, line 252: @mrobinson this is the entry point for the layout calcs.

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobserver-resizeobserver>
+    #[allow(non_snake_case)]
+    pub fn Constructor(
+        window: &Window,
+        proto: Option<HandleObject>,
+        callback: Rc<ResizeObserverCallback>,
+    ) -> DomRoot<ResizeObserver> {
+        let rooted_observer = ResizeObserver::new(window, proto, callback);
+        let document = window.Document();
+        document.add_resize_observer(&rooted_observer);
```
Loirooriol, line 78: This is per spec, but has memory problems: https://bugzilla.mozilla.org/show_bug.cgi?id=1596992

I don't remember how Blink managed to avoid that, but I think you should either do the same or disobey the spec like Gecko and WebKit do: https://github.com/w3c/csswg-drafts/issues/4518

gterzian: Thanks for pointing it out, there is a note over at https://github.com/servo/servo/pull/31108/files#diff-b6bdaf97b920500ab94dea9827cda04ff8136b0f23bd0c515d9195c6be31a820R439

and I'll add the info you mention to it...

gterzian:
@mrobinson Ping for the review items we discussed...

mrobinson:
> @mrobinson Ping for the review items we discussed...

I've just posted #31696 which should allow using the correct bounding rectangle for this API.

mrobinson (changes requested):
I think this just needs to use the new API for getting the boundaries of the first box for an element.

components/script/dom/document.rs
```diff
@@ -2893,6 +2901,61 @@ impl Document {
     pub fn name_map(&self) -> Ref<HashMapTracedValues<Atom, Vec<Dom<Element>>>> {
         self.name_map.borrow()
     }
+
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobserver-resizeobserver>
+    pub fn add_resize_observer(&self, resize_observer: &ResizeObserver) {
```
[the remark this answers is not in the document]
gterzian: Why? It seems unnecessary to me since the whole `dom` module is not public.

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+        shallowest_target_depth: &mut ResizeObservationDepth,
+    ) {
+        let mut entries: Vec<DomRoot<ResizeObserverEntry>> = Default::default();
+        for (observation, target) in self.observation_targets.borrow_mut().iter_mut() {
+            if !matches!(observation.state, ObservationState::Active) {
+                continue;
+            }
+            // #create-and-populate-a-resizeobserverentry
+            let box_size = calculate_box_size(target, &observation.observed_box);
+            // TODO: writing-mode aware.
```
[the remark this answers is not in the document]
gterzian: Yep, cleaned-up these.

components/script/script_thread.rs
```diff
@@ -1971,6 +1971,28 @@ impl ScriptThread {
...
+        // Run the resize observation steps.
+        for (_, document) in self.documents.borrow().iter() {
+            if !document.is_fully_active() {
+                continue;
+            }
+
+            let _realm = enter_realm(&*document);
+
+            let mut depth = Default::default();
+            loop {
```
jdm: Can we use a while here instead?

tests/wpt/meta/resize-observer/observe.html.ini
```diff
@@ -0,0 +1,4 @@
+[observe.html]
+  expected: ERROR
```
jdm, line 2: Do we know what the error is here?

tests/wpt/meta/resize-observer/notify.html.ini
```diff
@@ -0,0 +1,7 @@
+[notify.html]
+  expected: ERROR
```
jdm, line 2: Do we know what the error is here?

tests/wpt/meta/resize-observer/observe.html.ini
gterzian: Yes, it is a missing API. See https://github.com/servo/servo/issues/31182

components/script/dom/document.rs
```diff
@@ -2920,6 +2931,61 @@ impl Document {
...
+                &mut has_active_resize_observations,
+            );
+        }
+        has_active_resize_observations
+    }
+
+    /// <https://drafts.csswg.org/resize-observer/#broadcast-active-resize-observations>
+    pub fn broadcast_active_resize_observations(&self) -> ResizeObservationDepth {
+        let mut shallowest = ResizeObservationDepth::max();
+        // Breaking potential re-borrow cycle.
```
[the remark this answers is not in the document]
gterzian: When we call the observer callback, new observers can be registered, which would could a double-borrow(will add a more detailed comment).

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+                &*window,
+                target,
+                &*content_rect,
+                // TODO: border box.
+                &[],
+                &[&*observer_size],
+                &[&*observer_size],
+            );
+            entries.push(entry);
+            *observation.last_reported_sizes.borrow_mut() = vec![size_impl];
```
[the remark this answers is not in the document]
gterzian: Will do.

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+            observation.state = ObservationState::Done;
+            let target_depth = calculate_depth_for_node(target);
+            if target_depth < *shallowest_target_depth {
+                *shallowest_target_depth = target_depth;
+            }
+        }
+        let _ = self
```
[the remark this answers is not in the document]
gterzian: I've taken another look at all uses of `Call__` in the codebase, and we ignore all return values.

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobserver-disconnect>
+    fn Disconnect(&self) {
+        self.observation_targets.borrow_mut().clear();
+    }
+}
+
+/// State machine equivalent of active and skipped observations.
+#[derive(Default, JSTraceable, MallocSizeOf)]
```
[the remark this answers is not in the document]
gterzian: ok

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,262 @@
...
+/// https://drafts.csswg.org/resize-observer/#resizeobservation
+#[derive(JSTraceable, MallocSizeOf)]
+struct ResizeObservation {
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-target>
+    /// Note: `target` is kept out of here, to avoid having to root the `ResizeObservation`.
+
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-observedbox>
+    observed_box: ResizeObserverBoxOptions,
+    /// <https://drafts.csswg.org/resize-observer/#dom-resizeobservation-lastreportedsizes>
+    last_reported_sizes: RefCell<Vec<ResizeObserverSizeImpl>>,
```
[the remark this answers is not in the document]
gterzian: I guess not, will look into.

components/script/script_thread.rs
gterzian: Will do.

gterzian:
@jdm @Loirooriol @emilio @mrobinson As mentioned above, I would like to ask for a final review and try to merge this one, so that this would give us the basic infra(now integrated with "update the rendering" in the event-loop), and that remaining layout calculations, as well as a few other points noted in https://github.com/servo/servo/issues/31182, can be left as follow-ups.

jdm (approved):
Looks great! Thanks!

gterzian:
@mrobinson I think the PR needs an approval from you as well. I have addressed your suggestions at the time, but I haven't added the additional layout calculations because I think it's a good follow-up(see comment above).

components/script/dom/resizeobserver.rs
mrobinson: It should be possible to use the new query API that's been exposed to implement this correct versus in a followup

gterzian: > use the new query API

Do you mean `content_boxes`? If so, I have added a commit that uses that one.

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+        ResizeObserverBoxOptions::Content_box => {
+            // Note: only taking first fragment,
+            // but the spec will expand to cover all fragments.
+            target
+                .upcast::<Node>()
+                .content_boxes()
+                .pop()
+                .unwrap_or_else(|| Rect::zero())
+        },
+        // TODO#31182): add support for border box, and device pixel size, calculations.
```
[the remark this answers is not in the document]
gterzian: Thank you.

components/script/dom/document.rs
```diff
@@ -456,6 +457,15 @@ pub struct Document {
...
     #[ignore_malloc_size_of = "AnimationTickType contains data from an outside crate"]
     pending_animation_ticks: DomRefCell<AnimationTickType>,
+    /// <https://drafts.csswg.org/resize-observer/#dom-document-resizeobservers-slot>
+    ///
+    /// Note: we are storing, but never removing, resize observers.
+    /// The lifetime of resize observers is specified at
+    /// https://drafts.csswg.org/resize-observer/#lifetime.
+    /// But implementing it comes with known problems:
+    /// - https://bugzilla.mozilla.org/show_bug.cgi?id=1596992
+    /// - https://github.com/w3c/csswg-drafts/issues/4518.
```
mrobinson, line 467: URLs need to be wrapped in rustdoc comments or it will cause a warning:

```suggestion
    /// <https://drafts.csswg.org/resize-observer/#lifetime>.
    /// But implementing it comes with known problems:
    /// - <https://bugzilla.mozilla.org/show_bug.cgi?id=1596992>
    /// - <https://github.com/w3c/csswg-drafts/issues/4518>
```

components/script/dom/node.rs
```diff
@@ -2110,6 +2110,11 @@ impl Node {
             MutationObserver::queue_a_mutation_record(parent, mutation);
         }
         node.owner_doc().remove_script_and_layout_blocker();
+
+        // Note a rendering opportunity
+        let window = window_from_node(parent);
+        let pipeline_id = window.pipeline_id();
+        ScriptThread::note_rendering_opportunity(pipeline_id);
```
mrobinson, line 2117:
```suggestion
        ScriptThread::note_rendering_opportunity(window_from_node(parent).pipeline_id());
```

components/script/dom/node.rs
```diff
@@ -2233,6 +2238,11 @@ impl Node {
             MutationObserver::queue_a_mutation_record(parent, mutation);
         }
         parent.owner_doc().remove_script_and_layout_blocker();
+
+        // Note a rendering opportunity
+        let window = window_from_node(parent);
+        let pipeline_id = window.pipeline_id();
+        ScriptThread::note_rendering_opportunity(pipeline_id);
```
mrobinson, line 2245:
```suggestion
        ScriptThread::note_rendering_opportunity(window_from_node(parent).pipeline_id());
```

components/script/script_thread.rs
```diff
@@ -1967,6 +1984,11 @@ impl ScriptThread {
             for document in docs.iter() {
                 let _realm = enter_realm(&**document);
                 document.maybe_queue_document_completion();
+
+                // Document load is a rendering opportunity.
+                let window = document.window();
+                let pipeline_id = window.pipeline_id();
+                ScriptThread::note_rendering_opportunity(pipeline_id);
```
mrobinson, line 1991:
```suggestion
               // Document load is a rendering opportunity.
                ScriptThread::note_rendering_opportunity(document.window().pipeline_id());
```

components/script/dom/document.rs
```diff
@@ -2920,6 +2930,63 @@ impl Document {
...
+        }
+        shallowest
+    }
+
+    /// <https://drafts.csswg.org/resize-observer/#has-skipped-observations-h>
+    pub(crate) fn has_skipped_resize_observations(&self) -> bool {
+        self.resize_observers
+            .borrow()
+            .iter()
+            .any(|obs| obs.has_skipped_resize_observations())
```
mrobinson, line 2979: obs isn't really a well known abbreviation, so I would avoid it:

```suggestion
            .any(|observer| observer.has_skipped_resize_observations())
```

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+            .callback
+            .Call__(entries, self, ExceptionHandling::Report);
+    }
+
+    /// <https://drafts.csswg.org/resize-observer/#has-skipped-observations-h>
+    pub fn has_skipped_resize_observations(&self) -> bool {
+        self.observation_targets
+            .borrow()
+            .iter()
+            .any(|(obs, _)| obs.state == ObservationState::Skipped)
```
mrobinson, line 164:
```suggestion
            .any(|(observeration, _)| observation.state == ObservationState::Skipped)
```

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+                target,
+                &*content_rect,
+                &[],
+                &[&*observer_size],
+                &[],
+            );
+            entries.push(entry);
+
+            // Note: this is safe because an observation is
+            // initialized with one reported size(zero).
```
mrobinson, line 144:
```suggestion
            // initialized with one reported size (zero).
```

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+            observation.last_reported_sizes[0] = size_impl;
+            observation.state = ObservationState::Done;
+            let target_depth = calculate_depth_for_node(target);
+            if target_depth < *shallowest_target_depth {
+                *shallowest_target_depth = target_depth;
+            }
+        }
+        let _ = self
+            .callback
+            .Call__(entries, self, ExceptionHandling::Report);
```
mrobinson, line 156: Should this only happen when the size changes? I'm a bit unclear on the mechanism here because is_active() checks to see if the size has changed, but it seems this code both updates the size and then calls the resize observer callbacks. I don't see how the code avoids that if the size changes.

It also seems like `is_active()` is calculating the size but not caching it anywhere...meaning it updates here as well. It's not a cheap calculating right now, unfortunately, so that seems a little wasteful. Perhaps I misunderstand this code though...?

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        self.observation_targets
+            .borrow_mut()
+            .push((resize_observation, Dom::from_ref(target)));
+
+        // Note: noting a rendering opportunity here is necessary
+        // to make /resize-observer/iframe-same-origin.html PASS.
+        let window = window_from_node(target);
+        let pipeline_id = window.pipeline_id();
+        ScriptThread::note_rendering_opportunity(pipeline_id);
```
mrobinson, line 190:
```suggestion
        ScriptThread::note_rendering_opportunity( window_from_node(target).pipeline_id());
```

components/script/dom/resizeobserver.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn is_active(&self, target: &Element) -> bool {
+        let last_reported_size = self.last_reported_sizes[0];
+        let box_size = calculate_box_size(target, &self.observed_box);
+        let width = box_size.width().to_f64_px();
+        let height = box_size.height().to_f64_px();
+        (
+            last_reported_size.inline_size(),
+            last_reported_size.block_size(),
+        ) != (width, height)
+    }
```
mrobinson, line 252: I think the follow is shorter and clearer:
```suggestion
        box_size.width().to_f64_px() != last_reported_size.inline_size() ||
            box_size.height().to_f64_px() != last_reported_size.block_size()
```

mrobinson (changes requested):
Just a couple nits, but I don't see any big issues with this change. Hopefully we can land it in the next couple days. Sorry for the big delays!

components/script/dom/resizeobserver.rs
gterzian: > Should this only happen when the size changes?

This only happens for active observations, so that means size has changed plus observation depth is right. The code here does not update the size.

> It's not a cheap calculating right now, unfortunately, so that seems a little wasteful.

I have added a caching mechanism, see the last commit.

mrobinson (approved):

mrobinson:
@gterzian Looks good, though please squash your changes before landing or else the commit message will contain a lot of noise. Thanks you!

gterzian:
@Loirooriol @jdm @mrobinson thank you for your reviews.

PR #31115, opened 2024-01-18
wpt: Create a base class for Servo process executors

mrobinson:
This change creates a base class for the Servo process executors, to
handle shared functionality. The only thing that hasn't moved there yet
is the actual process execution, which can happen in a followup change.

The main motivation behind this change is consistently handling
`on_evironment_change` which is used to handle changes to the `prefs`
value stored in `__dir__.ini`. Inherited `prefs` (thos in `__dir__.ini`)
aren't passed when creating tests for reference HTML. This change takes
a similar approach to Gecko, which just listens to
`on_environment_change` to note when these prefs change.

---

- [x] There are tests for these changes, but will be active in a later change that addresses HTML table support.


mukilan (approved):
Nice!

PR #31117, opened 2024-01-18
layout: Convert all inline iteration to a new `foreach` function

mrobinson:
Instead of a tricky stack of enum iterators expose a `foreach()`
function on InlineFormattingContext, which takes a `FnMut`. This
prevents callers wanting to iterate from keeping a stack of iterators
and will potentially allow a future version of this function to avoid
borrowing the `ArcRefCell<...>` of inline boxes for every iteration
(presumably using something like `OwnedRef`).

Convert `inline_content_sizes` to use this new `foreach()` function and
move the `Computation` out of the function body to
`ContentSizesComputation`. This reduces the stack depth of inline size
computation, because `foreach()` is iterative and not recursive.

This is a preliminary change to removing the second round of text shaping
during layout, because shaping will use this new iterator.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -1333,6 +1333,11 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
     }
 }
 
+enum InlineFormattingContextIterItem<'a> {
+    Item(&'a mut InlineLevelBox),
+    EndInlineBox,
+}
+
 impl InlineFormattingContext {
     pub(super) fn new(
         text_decoration_line: TextDecorationLine,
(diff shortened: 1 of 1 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -2389,3 +2246,161 @@ fn inline_container_needs_strut(
...
+                    .margin(self.containing_block_writing_mode)
+                    .percentages_relative_to(zero)
+                    .auto_is(Length::zero);
+
+                let pbm = &(&margin + &padding) + &border;
+                if inline_box.is_first_fragment {
+                    self.add_length(pbm.inline_start);
+                }
+                if inline_box.is_last_fragment {
+                    self.ending_inline_pbm_stack.push(pbm.inline_end);
```
Loirooriol, line 2293: I don't really understand `ending_inline_pbm_stack`, why not just use `add_length` here?

mrobinson: This is so the length is added after the contents of the inline box, which might adjust the state of content sizes calculation. This is needed now that the calculation is not recursive any longer. The ending length is applied when the "EndInlineBox" item is processed.

mrobinson: This is so the length is added after the contents of the inline box, which might adjust the state of content sizes calculation. This is needed now that the calculation is not recursive any longer. The ending length is applied when the "EndInlineBox" item is processed.

Loirooriol (commented):
Lots of tests are failing

Loirooriol (approved):

PR #31123, opened 2024-01-18
Implement Event.composedPath

### LINKED ISSUE #31007: Implement Event.composedPath
Needed for https://github.com/servo/servo/issues/30822

See https://dom.spec.whatwg.org/#ref-for-dom-event-composedpath%E2%91%A0

Method should be added to the [WebIDL](https://github.com/servo/servo/blob/c219204084b8f8a747ebd37ec75472c1b7e97411/components/script/dom/webidls/Event.webidl), and the generated method implemented on [`Event`](https://github.com/servo/servo/blob/c219204084b8f8a747ebd37ec75472c1b7e97411/components/script/dom/event.rs)

For a starting point, look at https://github.com/servo/servo/blob/c219204084b8f8a747ebd37ec75472c1b7e97411/components/script/dom/event.rs#L147, and see https://dom.spec.whatwg.org/#event-path-invocation-target

Taym95:
---

- [x] These changes fix #31007

### DIFF
--- a/components/script/dom/event.rs
+++ b/components/script/dom/event.rs
@@ -433,6 +433,15 @@ impl EventMethods for Event {
         self.current_target.get()
     }
 
+    // https://dom.spec.whatwg.org/#dom-event-composedpath
+    fn ComposedPath(&self) -> Vec<DomRoot<EventTarget>> {
+        if let Some(target) = self.target.get() {
+            self.construct_event_path(&target)
+        } else {
+            vec![]
+        }
+    }
+
     // https://dom.spec.whatwg.org/#dom-event-defaultprevented
     fn DefaultPrevented(&self) -> bool {
         self.canceled.get() == EventDefault::Prevented
--- a/components/script/dom/event.rs
+++ b/components/script/dom/event.rs
@@ -141,7 +141,7 @@ impl Event {
         self.cancelable.set(cancelable);
     }
 
-    // https://dom.spec.whatwg.org/#event-path
+    /// <https://dom.spec.whatwg.org/#event-path>
     // TODO: shadow roots put special flags in the path,
     // and it will stop just being a list of bare EventTargets
     fn construct_event_path(&self, target: &EventTarget) -> Vec<DomRoot<EventTarget>> {
@@ -178,7 +178,7 @@ impl Event {
         event_path
     }
 
-    // https://dom.spec.whatwg.org/#concept-event-dispatch
+    /// <https://dom.spec.whatwg.org/#concept-event-dispatch>
     pub fn dispatch(
         &self,
         target: &EventTarget,
@@ -400,7 +400,7 @@ impl Event {
         self.trusted.set(trusted);
     }
 
-    // https://html.spec.whatwg.org/multipage/#fire-a-simple-event
+    /// <https://html.spec.whatwg.org/multipage/#fire-a-simple-event>
     pub fn fire(&self, target: &EventTarget) -> EventStatus {
         self.set_trusted(true);
         target.dispatch_event(self)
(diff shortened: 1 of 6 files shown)


Taym95:
@gterzian I test it on https://www.w3schools.com/jsreF/tryit.asp?filename=tryjsref_event_composedpath

seems to give same result as Chrome and Firefox, do you think this enough or I should do more?

components/script/dom/event.rs
```diff
@@ -433,6 +433,15 @@ impl EventMethods for Event {
         self.current_target.get()
     }
 
+    // https://dom.spec.whatwg.org/#dom-event-composedpath
```
gterzian, line 436: nit: `///`

components/script/dom/event.rs
```diff
@@ -433,6 +433,15 @@ impl EventMethods for Event {
...
+    fn ComposedPath(&self) -> Vec<DomRoot<EventTarget>> {
+        if let Some(target) = self.target.get() {
+            self.construct_event_path(&target)
```
gterzian, line 439: There is a TODO on that method for the shadow root stuff, it would be good to file an issue for it, and it would be a good follow-up if you are interested.

tests/wpt/mozilla/meta/mozilla/service-workers/service-worker-registration.https.html.ini
```diff
@@ -1,34 +1,12 @@
 [service-worker-registration.https.html]
-  [Test: Asserts Installing Service Worker and its Registration]
```
gterzian, line 2: The updates here are unexpected, did you notice they were consistently passing locally?

gterzian (changes requested):
Great work, thanks!

Couple of comments, and regarding the tests:

1. I would expect this one to not error anymore: https://github.com/servo/servo/blob/8dcf3dd0b0fee0658fea43645db266be32eb2c11/tests/wpt/meta/dom/events/relatedTarget.window.js.ini#L2
2. There are some tests for this in the `shadow-dom` folder, which is currently skipped. It makes sense to unskip the folder by way of changes to [`include.ini`](https://github.com/servo/servo/blob/8dcf3dd0b0fee0658fea43645db266be32eb2c11/tests/wpt/include.ini) to see if any tests are passing.

components/script/dom/event.rs
Taym95: should I use `///` for all other functions to keep it consistent ?

components/script/dom/event.rs
Taym95: I will create issue and follow up

components/script/dom/event.rs
mrobinson, line 436:
```suggestion
    /// <https://dom.spec.whatwg.org/#dom-event-composedpath>
```

Three slashes means that this comment is interpreted as rustdoc and will be included in the generated documentation, so it's always good to do that when documenting functions. Please also use angle brackets around URLs. We need to start doing that because without them there are many clippy and rustdoc errors. Thanks!

tests/wpt/mozilla/meta/mozilla/service-workers/service-worker-registration.https.html.ini
Taym95: Indeed, it is failing. I was misled by CI. I will revert this

components/script/dom/event.rs
Taym95: I used documentation comments for all function in this file

Taym95:
> Great work, thanks!
> 
> Couple of comments, and regarding the tests:
> 
> 1. I would expect this one to not error anymore: https://github.com/servo/servo/blob/8dcf3dd0b0fee0658fea43645db266be32eb2c11/tests/wpt/meta/dom/events/relatedTarget.window.js.ini#L2
> 2. There are some tests for this in the `shadow-dom` folder, which is currently skipped. It makes sense to unskip the folder by way of changes to [`include.ini`](https://github.com/servo/servo/blob/8dcf3dd0b0fee0658fea43645db266be32eb2c11/tests/wpt/include.ini) to see if any tests are passing.

Thanks @gterzian 

For 1 I am still getting:

```
Error at http://web-platform.test:8000/dom/events/relatedTarget.window.js:5:21 host.attachShadow is not a function
```

for 2 maybe I am doing something wrong but I don't see `[shadow-dom]` skipped in `include.ini`

components/script/dom/event.rs
gterzian: If they are unrelated, it would be better to do it separately. Not very important.

gterzian (approved):
LGTM, I would say we leave the shadom-dom tests alone for now, because it looks like we are missing support for most of it so it is not worth it to unskip those tests. By the way they are skipped by default, so unskipping them means adding `skip: false` to the `.ini` file...

PR #31135, opened 2024-01-19
Layout: use `Au` in `ContentSizes` 

atbrakhi:
---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -2359,7 +2359,8 @@ impl<'a> ContentSizesComputation<'a> {
                     self.containing_block_writing_mode,
                 );
 
-                self.current_line.min_content += (self.pending_whitespace + outer.min_content.into()).into();
+                self.current_line.min_content +=
+                    (self.pending_whitespace + outer.min_content.into()).into();
                 self.current_line.max_content += outer.max_content;
                 self.pending_whitespace = Length::zero();
                 self.had_non_whitespace_content_yet = true;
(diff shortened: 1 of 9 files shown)


atbrakhi:
marking as draft until results can be updated.

components/layout_2020/flow/inline.rs
```diff
@@ -1467,21 +1467,26 @@ impl InlineFormattingContext {
...
 
             fn line_break_opportunity(&mut self) {
-                self.paragraph
-                    .min_content
-                    .max_assign(take(&mut self.current_line.min_content));
+                let max_min_content = std::cmp::max(
+                    self.paragraph.min_content,
+                    take(&mut self.current_line.min_content),
+                );
+                self.paragraph.min_content = max_min_content;
```
Taym95, line 1479: is `max_min_content` needed?

```suggestion
                self.paragraph.min_content = std::cmp::max(
                    self.paragraph.min_content,
                    take(&mut self.current_line.min_content),
                );
```

atbrakhi: nice catch. I have updated it and also `self.paragraph.max_content`. Thanks!

components/layout_2020/sizing.rs
```diff
@@ -100,28 +101,32 @@ pub(crate) fn outer_inline(
...
+                max_content: border_box_size.into(),
             }
         },
         None => get_content_size().map(|content_box_size| {
             match box_sizing {
                 // Clamp to 'min-width' and 'max-width', which are sizing the…
-                BoxSizing::ContentBox => clamp(content_box_size) + pb_lengths,
-                BoxSizing::BorderBox => clamp(content_box_size + pb_lengths),
+                BoxSizing::ContentBox => {
+                    (clamp(content_box_size.into()) + pb_lengths.into()).into()
```
Taym95, line 124: maybe you can remove extra `.into()` ?

```suggestion
                    clamp(content_box_size.into()) + pb_lengths.into()
```

components/layout_2020/table/layout.rs
```diff
@@ -450,7 +456,7 @@ impl Table {
         };
 
         let sizes = cell.inline_content_sizes(layout_context, writing_mode);
-        sizes.map(|size| size / cell.colspan as f32)
+        sizes.map(|size| Au::from_f32_px(Au::to_f32_px(size) / cell.colspan as f32))
```
mrobinson, line 459: I think you can avoid the conversion in this case since size is already `Au` and we are just scaling by `colspan`.

atbrakhi: I am not sure if we can completely avoid conversion here. `size` is of type `Au`, while `cell.colspan` is a `usize`. Conversion seems necessary to ensure type compatibility for the division operation; otherwise, the compiler complains. However, I think it's better to perform the conversion on `cell.colspan` instead of `size`.

If you agree, I will go ahead and do that

atbrakhi: I see, updated in [23aef62](https://github.com/servo/servo/pull/31135/commits/23aef62e1323912904edce349aaae0e3e2149267). thanks :)

mrobinson (approved):

PR #31137, opened 2024-01-19
Layout: Clean up `Au` conversion helper functions in the flexbox 

atbrakhi:
---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -2405,6 +2405,7 @@ impl<'a> ContentSizesComputation<'a> {
             had_non_whitespace_content_yet: false,
             linebreaker: None,
             ending_inline_pbm_stack: Vec::new(),
-        }.traverse(inline_formatting_context)
+        }
+        .traverse(inline_formatting_context)
     }
 }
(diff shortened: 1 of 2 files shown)


atbrakhi:
marking as draft until results can be updated.

components/layout_2020/flow/inline.rs
```diff
@@ -2405,6 +2405,7 @@ impl<'a> ContentSizesComputation<'a> {
             had_non_whitespace_content_yet: false,
             linebreaker: None,
             ending_inline_pbm_stack: Vec::new(),
-        }.traverse(inline_formatting_context)
+        }
+        .traverse(inline_formatting_context)
```
mrobinson, line 2409: I think this change is accidental.

mrobinson (approved):
Looks good. Feel free to land this after removing the unintentional change. Thanks!

components/layout_2020/flow/inline.rs
atbrakhi: This is `./mach fmt` doing. I noticed this in another change as well. Would you like me to revert it? But mach fmt will suggest it again.

mrobinson: Huh! Sorry. Nope it's fine to leave it as-is.

PR #31141, opened 2024-01-20
Matrix in CI and `mach try` with presets

sagudev:
Splitted from https://github.com/servo/servo/pull/30920

This PR replaces try branches with single `try` branch that accepts configuration (that is encoded in last line of commit in JSON). `try_parser.py` is there just to have preset -> JSON in one place (try label = preset).

---

### DIFF
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -29,6 +29,9 @@ jobs:
     outputs:
       configuration: ${{ steps.configuration.outputs.result }}
     steps:
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
       - uses: actions/checkout@v3
         with:
           fetch-depth: 1
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -29,6 +29,17 @@ jobs:
     outputs:
       configuration: ${{ steps.configuration.outputs.result }}
     steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          sparse-checkout: |
+            python/servo/try_parser.py
+          sparse-checkout-cone-mode: false
+      - name: Generate full
+        id: full
+        run: |
+          out=$(python ./python/servo/try_parser.py)
+          echo "config<<EOF"$'\n'"$out"$'\n'EOF >> $GITHUB_OUTPUT
       - name: Configuration
         id: configuration
         uses: actions/github-script@v6
@@ -52,35 +63,7 @@ jobs:
             }
 
             // If we reach here we are likely doing a full run.
-            configuration = {
-              "fail_fast": false,
-              "matrix": [
-                {
-                  "workflow": "linux",
-                  "name": "Linux",
-                  "wpt_layout": "all",
-                  "profile": "release",
-                  "unit_tests": true,
-                  "wpt_tests_to_run": "",
-                },
-                {
-                  "workflow": "windows",
-                  "name": "Windows",
-                  "wpt_layout": "none", // not used
-                  "profile": "release",
-                  "unit_tests": true,
-                  "wpt_tests_to_run": "",
-                },
-                {
-                  "workflow": "mac",
-                  "name": "MacOS",
-                  "wpt_layout": "none",
-                  "profile": "release",
-                  "unit_tests": true,
-                  "wpt_tests_to_run": "",
-                }
-              ],
-            };
+            configuration = ${{ steps.full.outputs.config }};
 
             // Process `workflow_dispatch` provided configuration overrides.
             if (context.eventName == "workflow_dispatch") {
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   decision:
-    name: Decision
+    name: Generate matrix
     runs-on: ubuntu-20.04
     outputs:
       configuration: ${{ steps.configuration.outputs.result }}
@@ -78,7 +78,7 @@ jobs:
               let profile = '${{ inputs.profile }}';
               for (const platform of configuration.matrix) {
                 platform.profile = profile;
-                platfrom.unit_tests = unit_tests;
+                platform.unit_tests = unit_tests;
               }
             }
 
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -24,7 +24,7 @@ on:
 
 jobs:
   decision:
-    name: Generate matrix
+    name: Generate Try Configuration
     runs-on: ubuntu-20.04
     outputs:
       configuration: ${{ steps.configuration.outputs.result }}
@@ -38,11 +38,14 @@ jobs:
           sparse-checkout: |
             python/servo/try_parser.py
           sparse-checkout-cone-mode: false
-      - name: Generate full
-        id: full
+      - name: Get Full Configuration
+        id: full_config
         run: |
-          out=$(python ./python/servo/try_parser.py)
-          echo "config<<EOF"$'\n'"$out"$'\n'EOF >> $GITHUB_OUTPUT
+          {
+            echo 'config<<EOF'
+            python ./python/servo/try_parser.py
+            echo EOF
+           } >> $GITHUB_OUTPUT
       - name: Configuration
         id: configuration
         uses: actions/github-script@v6
@@ -66,7 +69,7 @@ jobs:
             }
 
             // If we reach here we are likely doing a full run.
-            configuration = ${{ steps.full.outputs.config }};
+            configuration = ${{ steps.full_config.outputs.config }};
 
             // Process `workflow_dispatch` provided configuration overrides.
             if (context.eventName == "workflow_dispatch") {
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -1,203 +1,134 @@
 name: Try
 
 on:
-  pull_request_target:
-    types: [labeled]
   push:
-    branches: ["try", "try-linux", "try-mac", "try-wpt-mac", "try-wpt-mac-2020", "try-wpt", "try-wpt-2020", "try-windows"]
+    branches: ["try"]
+  workflow_dispatch:
+    inputs:
+      profile:
+        required: false
+        default: "release"
+        type: choice
+        options: ["release", "debug", "production"]
+      wpt-tests-to-run:
+        default: ""
+        required: false
+        type: string
+      wpt-layout:
+        required: false
+        type: choice
+        options: ["none", "2013", "2020", "all"]
+      unit-tests:
+        required: false
+        type: boolean
 
 jobs:
-  parse-comment:
-    name: Trigger Try
-    runs-on: ubuntu-latest
-    concurrency:
-      group: try-${{ github.event.number }}
+  decision:
+    name: Decision
+    runs-on: ubuntu-20.04
     outputs:
       configuration: ${{ steps.configuration.outputs.result }}
     steps:
-      - uses: actions/github-script@v6
+      - name: Configuration
         id: configuration
+        uses: actions/github-script@v6
         with:
           script: |
-            function makeComment(body) {
-                console.log(body);
+            // When triggered via a push try the `try` branch, search the last commit for a configuration object.
+            if (${{ github.ref_name == 'try' }}) {
+              let commit_msg = context.payload.head_commit.message;
+              try {
+                var config = JSON.parse(commit_msg.split('\n').slice(-1));
 
-                if (context.eventName != 'pull_request_target')
-                  return;
-
-                github.rest.issues.createComment({
-                  issue_number: context.issue.number,
-                  owner: context.repo.owner,
-                  repo: context.repo.repo,
-                  body
-                })
-            }
-
-            function combineWPTLayoutOptions(layout, newLayout) {
-                let has2013 = layout == "2013" || layout == "all";
-                let has2020 = layout == "2020" || layout == "all";
-                let adding2013 = newLayout == "2013";
-                let adding2020 = newLayout == "2020";
-
-                if ((adding2020 && has2020) || (adding2013 && has2013)) {
-                  return layout;
+                if (config && typeof config === "object") {
+                  console.log("Using try commit configuration: " + JSON.stringify(config));
+                  return config;
                 }
-                if (adding2020) {
-                  return has2013 ? "all" : "2020";
-                }
-                if (adding2013) {
-                  return has2020 ? "all" : "2013";
-                }
-                return layout;
-            }
-
-            function addPlatformToConfiguration(platform, configuration) {
-              if (!configuration.platforms.includes(platform)) {
-                configuration.platforms.push(platform);
+              }
+              catch (exception) {
+                console.log("Could not parse try configuration from commit message: " + exception);
+                console.log("Triggering full try run.");
               }
             }
 
-            function updateConfigurationFromString(tryString, configuration) {
-                if (tryString.includes("full")) {
-                  configuration.platforms = ["linux", "macos", "windows"];
-                  configuration.unit_tests = true;
-                  configuration.wpt_layout = "all";
-                  return configuration;
+            // If we reach here we are likely doing a full run.
+            configuration = {
+              "fail_fast": false,
+              "matrix": [
+                {
+                  "workflow": "linux",
+                  "name": "Linux",
+                  "wpt_layout": "all",
+                  "profile": "release",
+                  "unit_tests": true,
+                  "wpt_tests_to_run": "",
+                },
+                {
+                  "workflow": "windows",
+                  "name": "Windows",
+                  "wpt_layout": "none", // not used
+                  "profile": "release",
+                  "unit_tests": true,
+                  "wpt_tests_to_run": "",
+                },
+                {
+                  "workflow": "mac",
+                  "name": "MacOS",
+                  "wpt_layout": "none",
+                  "profile": "release",
+                  "unit_tests": true,
+                  "wpt_tests_to_run": "",
                 }
-
-                if (tryString.includes("linux")) {
-                  addPlatformToConfiguration("linux", configuration);
-                  configuration.unit_tests = true;
-                } else if (tryString.includes("mac")) {
-                  addPlatformToConfiguration("macos", configuration);
-                  configuration.unit_tests = true;
-                } else if (tryString.includes("win")) {
-                  addPlatformToConfiguration("windows", configuration);
-                  configuration.unit_tests = true;
-                }
-
-                if (tryString.includes("wpt")) {
-                  addPlatformToConfiguration("linux", configuration);
-                  if (tryString.includes("2020")) {
-                    configuration.wpt_layout = combineWPTLayoutOptions(configuration.wpt_layout, "2020");
-                  } else {
-                    configuration.wpt_layout = combineWPTLayoutOptions(configuration.wpt_layout, "2013");
-                  }
-                }
-            }
-
-            let configuration = {
-              platforms: [],
-              wpt_layout: "none",
-              unit_tests: false,
-              profile: "release",
-              wpt_tests_to_run: "",
+              ],
             };
 
-            if (context.eventName == 'pull_request_target') {
-              for (const label of context.payload.pull_request.labels) {
-                if (!label.name.startsWith("T-")) {
-                  continue;
-                }
+            // Process `workflow_dispatch` provided configuration overrides.
+            if (context.eventName == "workflow_dispatch") {
+              // WPT-related overrides only affect Linux currently, as tests don't run by default on other platforms.
+              configuration.matrix[0].wpt_layout = "${{ inputs.wpt-layout }}" || "none";
+              configuration.matrix[0].wpt_tests_to_run = "${{ inputs.wpt-tests-to-run }}" || "";
 
-                // Try to remove the label. If that fails, it's likely that another
-                // workflow has already processed it or a user has removed it.
-                try {
-                  await github.rest.issues.removeLabel({
-                    owner: context.repo.owner,
-                    repo: context.repo.repo,
-                    issue_number: context.issue.number,
-                    name: label.name,
-                  });
-                } catch (exception) {
-                  console.log("Assuming '" + label.name + "' is already removed: " + exception);
-                  continue;
-                }
-
-                console.log("Found label: " + label.name);
-                updateConfigurationFromString(label.name, configuration);
-              }
-            } else {
-              let ref_name = "${{ github.ref_name || 'empty' }}";
-              if (ref_name == "try") {
-                updateConfigurationFromString("full", configuration);
-              } else {
-                updateConfigurationFromString(ref_name, configuration);
+              let unit_tests = Boolean(${{ inputs.unit-tests }});
+              let profile = '${{ inputs.profile }}';
+              for (const platform of configuration.matrix) {
+                platform.profile = profile;
+                platfrom.unit_tests = unit_tests;
               }
             }
 
-            console.log(JSON.stringify(configuration));
-
-            if (configuration.platforms.length == 0) {
-                return { platforms: [] };
-            }
-
-            let username = context.payload.sender.login;
-            let result = await github.rest.repos.getCollaboratorPermissionLevel({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              username
-            });
-            if (!result.data.user.permissions.push) {
-              makeComment('🔒 User @' + username + ' does not have permission to trigger try jobs.');
-              return { platforms: [] };
-            }
-
-            const url = context.serverUrl +
-              "/" + context.repo.owner +
-              "/" + context.repo.repo +
-              "/actions/runs/" + context.runId;
-            const formattedURL = "[#" + context.runId + "](" + url + ")";
-            let platformsString = configuration.platforms.toString();
-            makeComment("🔨 Triggering try run (" + formattedURL + ") with platforms=" +
-                        platformsString + " and layout=" + configuration.wpt_layout);
+            console.log("Using configuration: " + JSON.stringify(configuration));
             return configuration;
 
-  run-try:
-    name: Run Try
-    needs: ["parse-comment"]
-    if: ${{ fromJson(needs.parse-comment.outputs.configuration).platforms[0] != null }}
-    uses: ./.github/workflows/main.yml
+  build:
+    needs: ["decision"]
+    name: ${{ matrix.name }}
+    strategy:
+      fail-fast: ${{ fromJson(needs.decision.outputs.configuration).fail_fast }}
+      matrix:
+        include: ${{ fromJson(needs.decision.outputs.configuration).matrix }}
+    # We need to use `dipatch-workflow.yml` because workflows do not support using: ${}
+    uses: ./.github/workflows/dispatch-workflow.yml
+    secrets: inherit
     with:
-      configuration: ${{ needs.parse-comment.outputs.configuration }}
+      workflow: ${{ matrix.workflow }}
+      wpt-layout: ${{ matrix.wpt_layout }}
+      profile: ${{ matrix.profile }}
+      unit-tests: ${{ matrix.unit_tests }}
+      wpt-tests-to-run: ${{ matrix.wpt_tests_to_run }}
 
-  results:
-    name: Results
-    needs: ["parse-comment", "run-try"]
+  build-result:
+    name: Result
     runs-on: ubuntu-latest
-    if: ${{ always() && fromJson(needs.parse-comment.outputs.configuration).platforms[0] != null }}
+    if: always()
+    # `needs: "build"` is necessary to detect cancellation.
+    needs:
+      - "decision"
+      - "build"
+
     steps:
-      - name: Success
-        if: ${{ github.event_name == 'pull_request_target' && !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
-        uses: actions/github-script@v6
-        with:
-          script: |
-              const url = context.serverUrl +
-                "/" + context.repo.owner +
-                "/" + context.repo.repo +
-                "/actions/runs/" + context.runId;
-              const formattedURL = "[#" + context.runId + "](" + url + ")";
-              github.rest.issues.createComment({
-                issue_number: context.issue.number,
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                body: "✨ Try run (" + formattedURL + ") " + "succeeded.",
-              });
-      - name: Failure
-        if: ${{ github.event_name == 'pull_request_target' && contains(needs.*.result, 'failure') }}
-        uses: actions/github-script@v6
-        with:
-          script: |
-              const url = context.serverUrl +
-                "/" + context.repo.owner +
-                "/" + context.repo.repo +
-                "/actions/runs/" + context.runId;
-              const formattedURL = "[#" + context.runId + "](" + url + ")";
-              github.rest.issues.createComment({
-                issue_number: context.issue.number,
-                owner: context.repo.owner,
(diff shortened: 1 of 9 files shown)


.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,172 @@
...
+      # This is necessary to checkout the pull request if this run was triggered
+      # via an `issue_comment` action on a pull request.
+      - uses: actions/checkout@v3
+        if: github.event_name == 'issue_comment' || github.event_name == 'pull_request_target'
+        with:
+          ref: refs/pull/${{ github.event.issue.number || github.event.number }}/head
+          fetch-depth: 2
+          sparse-checkout: |
+            etc/try_parser.py
+            third_party/ply
```
mrobinson, line 106: I don't think ply is necessary any longer.

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self, name: str, os: OS = OS.linux, layout: Layout = Layout.none,
+                 profile: str = "release", unit_test: bool = True, wpt: str = ""):
+        self.os: OS = os
+        self.name: str = name
+        self.wpt_layout: Layout = layout
+        self.profile: str = profile
+        self.unit_tests: bool = unit_test
+        self.wpt_tests_to_run: str = wpt
+
```
mrobinson, line 45: This should probably be a `dataclass`, which will simplify a lot of this code.

.github/workflows/try_labels.yml
sagudev: You are right.

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    none = 'none'
+    layout2013 = '2013'
+    layout2020 = '2020'
+    all = 'all'
+
+
+class OS(str, Enum):
+    linux = 'linux'
+    mac = 'mac'
+    win = 'windows'
```
mrobinson, line 27: I don't think you need multiple superclasses for this and let's avoid shortening the names.

```suggestion
class OS(Enum):
    LINUX = 'linux'
    MACOS = 'mac'
    WINDOWS = 'windows'
```

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
...
+import json
+from enum import Enum
+import sys
+
+
+class Layout(str, Enum):
+    none = 'none'
+    layout2013 = '2013'
+    layout2020 = '2020'
+    all = 'all'
```
mrobinson, line 21: This could be a [`Flag`](https://docs.python.org/3/library/enum.html#enum.Flag) type to avoid the ambiguity?

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
...
+        self.unit_tests: bool = unit_test
+        self.wpt_tests_to_run: str = wpt
+
+
+def preset(s: str) -> JobConfig | None:
+    s = s.lower()
+
+    if s == "linux":
+        return JobConfig("Linux", OS.linux)
+    elif s == "mac":
```
mrobinson, line 52: I think it makes sense to accept "mac" as well as "macos" as you've done with "win" and "windows" below.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,172 @@
...
+            const url = context.serverUrl +
+              "/" + context.repo.owner +
+              "/" + context.repo.repo +
+              "/actions/runs/" + context.runId;
+            const formattedURL = "[#" + context.runId + "](" + url + ")";
+            makeComment("🔨 Triggering try run (" + formattedURL + ")");
+
+            return try_string;
+      - uses: actions/checkout@v3
+        if: github.event_name != 'issue_comment' && github.event_name != 'pull_request_target'
```
mrobinson, line 90: I think that the `issue_comment` bit here can be removed as we don't support triggering jobs via issues any longer.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,172 @@
...
+        with:
+          fetch-depth: 2
+          sparse-checkout: |
+            etc/try_parser.py
+            third_party/ply
+          sparse-checkout-cone-mode: false
+      # This is necessary to checkout the pull request if this run was triggered
+      # via an `issue_comment` action on a pull request.
+      - uses: actions/checkout@v3
+        if: github.event_name == 'issue_comment' || github.event_name == 'pull_request_target'
```
mrobinson, line 100: Ditto.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,172 @@
...
+
+  run-try:
+    if: ${{ needs.parse-comment.outputs.try_string }}
+    needs: ["parse-comment"]
+    name: ${{ matrix.name }}
+    strategy:
+      fail-fast: ${{ fromJson(needs.parse-comment.outputs.configuration).fail_fast }}
+      matrix:
+        include: ${{ fromJson(needs.parse-comment.outputs.configuration).matrix }}
+    # we need dipatch os because workflows do not support uses: ${}
```
mrobinson, line 123:
```suggestion
    # We need to use `dipatch-os.yml` because workflows do not support using: ${}.
```

.github/workflows/try.yml
```diff
@@ -1,203 +1,135 @@
...
-                let adding2013 = newLayout == "2013";
-                let adding2020 = newLayout == "2020";
-
-                if ((adding2020 && has2020) || (adding2013 && has2013)) {
-                  return layout;
-                }
-                if (adding2020) {
-                  return has2013 ? "all" : "2020";
+            // try runs can have config in last commit msg
+            if (${{ github.ref_name == 'try' }}) {
```
mrobinson, line 42: This is always true now, right?

.github/workflows/dispatch-os.yml
```diff
@@ -0,0 +1,48 @@
...
+  win:
+    if: ${{ inputs.os == 'windows' }}
+    uses: ./.github/workflows/windows.yml
+    secrets: inherit
+    with:
+      profile: ${{ inputs.profile }}
+      unit-tests: ${{ inputs.unit-tests }}
+
+  mac:
+    if: ${{ inputs.os == 'mac' }}
```
mrobinson, line 31: Can we rename this to `macos` since we are making changes?

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
...
+        return json.dumps(self, cls=Encoder)
+
+
+def main():
+    conf = Config(" ".join(sys.argv[1:]))
+    print(conf.toJSON())
+
+
+if __name__ == "__main__":
+    main()
```
mrobinson, line 113: Thanks! This is a lot more maintainable. Would it be possible to add some tests here and run them as part of `test-scripts`?

etc/try_parser.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
+
```
mrobinson, line 2: Let's put this in the `python` directory so that it's integrated with the other scripts.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,172 @@
...
+            etc/try_parser.py
+            third_party/ply
+          sparse-checkout-cone-mode: false
+      # This is necessary to checkout the pull request if this run was triggered
+      # via an `issue_comment` action on a pull request.
+      - uses: actions/checkout@v3
+        if: github.event_name == 'issue_comment' || github.event_name == 'pull_request_target'
+        with:
+          ref: refs/pull/${{ github.event.issue.number || github.event.number }}/head
+          fetch-depth: 2
```
mrobinson, line 103: A `fetch-depth` of 1 should be enough, right?

mrobinson (commented):
I know this is just a draft, but this is looking really nice.

etc/try_parser.py
sagudev: I need to learn more about python.

.github/workflows/try.yml
sagudev: Not on workflow_dispatch.

.github/workflows/try_labels.yml
sagudev: I think so.

etc/try_parser.py
sagudev: I forgot to port test from base PR.

sagudev:
The only reason this is still draft is because I want to retest some usages (mainly labels runs)

docs/HACKING_QUICKSTART.md
```diff
@@ -384,6 +384,12 @@ If you need to create a new test file, it should be located in `tests/wpt/mozill
 ./mach test-wpt --manifest-update
 ```
 
+### Running tests via GitHub Actions
+
+Running whole WPT suite is resource intensive and tedious task. Instead of using your computer resources, you can enable Workflows in personal fork and then send your patches to try branch. To make things easier you can use `./mach try` that will automatically send git `HEAD` (patches that are committed in current checkout) to try branch and run full CI build, using same config as it's used for landing things in servo/servo.
+
+For advance try runs see: [try guide](./try.md)
+
```
mrobinson, line 392: Let's manage this new documentation on the wiki.

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
+        '--remote', '-r', default="origin",
+        help='git remote to use for try pushes')
+    @CommandArgument(
+        'try_string', default=None, nargs='...',
+        help="Try string")
+    def try_jobs(self, remote="origin", try_string=None):
+        # if jobs is empty we just push last commit to try and it will do full run
+        if try_string:
+            try_string = " ".join(try_string)
+            sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "etc"))
```
mrobinson, line 289: This is no longer necessary.

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
+            from python.try_parser import Config
```
mrobinson, line 290: Please place the new file within the `servo` directory and do `import servo.try_parser` at the top of the file.

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
+            conf = Config(try_string)
+            res = call(["git", "commit", "--allow-empty", "-m", f"{try_string}\n\n{conf.toJSON()}"],
+                       env=self.build_env())
```
mrobinson, line 293: The full build environment isn't necessary here:

```suggestion
            config = Config(try_string)
            result = call(["git", "commit", "--allow-empty", "-m", f"{try_string}\n\n{conf.toJSON()}"])
```

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
             if res != 0:
                 return res
-        return 0
+
+        if "servo/servo" in subprocess.check_output(["git", "config", "--get", f"remote.{remote}.url"]).decode():
+            print("WARNING: You are triggering try build in upstream repo!")
+
```
mrobinson, line 299: Instead of this warning, I think it might be nice to unconditionally print where the try build is triggered so that the user can click on a URL to go to the actions.

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
+        # if jobs is empty we just push last commit to try and it will do full run
+        if try_string:
+            try_string = " ".join(try_string)
+            sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "etc"))
+            from python.try_parser import Config
+            conf = Config(try_string)
+            res = call(["git", "commit", "--allow-empty", "-m", f"{try_string}\n\n{conf.toJSON()}"],
+                       env=self.build_env())
             if res != 0:
                 return res
```
mrobinson, line 295:
```suggestion
            if result != 0:
                return result
```

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
-        return 0
+
+        if "servo/servo" in subprocess.check_output(["git", "config", "--get", f"remote.{remote}.url"]).decode():
+            print("WARNING: You are triggering try build in upstream repo!")
+
+        res = call(["git", "push", remote, "--force", "HEAD:try"], env=self.build_env())
+
+        if try_string:
+            # we need to delete last commit
+            res += call(["git", "reset", "--soft", "HEAD~1"], env=self.build_env())
```
mrobinson, line 304: I'm not sure this is necessary. I think it's fine to leave the last commit in the try branch.

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,32 @@ def ndk_gdb(self, release, target):
...
+            res = call(["git", "commit", "--allow-empty", "-m", f"{try_string}\n\n{conf.toJSON()}"],
+                       env=self.build_env())
             if res != 0:
                 return res
-        return 0
+
+        if "servo/servo" in subprocess.check_output(["git", "config", "--get", f"remote.{remote}.url"]).decode():
+            print("WARNING: You are triggering try build in upstream repo!")
+
+        res = call(["git", "push", remote, "--force", "HEAD:try"], env=self.build_env())
```
mrobinson, line 300:
```suggestion
        result = call(["git", "push", remote, "--force", "HEAD:try"])
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class Config(object):
+    def __init__(self, s: str | None = None):
+        self.fail_fast: bool = False
+        self.matrix: list[JobConfig] = list()
+        if s is not None:
+            self.parse(s)
+
+    def parse(self, s: str):
```
mrobinson, line 101:
```suggestion
    def parse(self, input: str):
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+        s = s.lower().strip()
+
+        if not s:
+            s = "full"
+
+        words: list[str] = s.split(" ")
+
+        for m in words:
```
mrobinson, line 109:
```suggestion
        for word in words:
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+            # handle keywords
+            if m in ["fail-fast", "failfast", "fail_fast"]:
+                self.fail_fast = True
+                continue  # skip over keyword
+            if m == "full":
+                words.extend(["linux", "macos", "windows"])
+                continue  # skip over keyword
+
+            # handle presets
+            p = preset(m)
```
mrobinson, line 119:
```suggestion
            preset = handle_preset(m)
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+    def parse(self, s: str):
+        s = s.lower().strip()
+
+        if not s:
+            s = "full"
+
+        words: list[str] = s.split(" ")
+
+        for m in words:
+            # handle keywords
```
mrobinson, line 110:
```suggestion
            # Handle keywords.
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+                self.fail_fast = True
+                continue  # skip over keyword
+            if m == "full":
+                words.extend(["linux", "macos", "windows"])
+                continue  # skip over keyword
+
+            # handle presets
+            p = preset(m)
+            if p is None:
+                print(f"Ignoring wrong preset {m}")
```
mrobinson, line 121:
```suggestion
                print(f"Ignoring unknown preset {word}")
```

python/try_parser.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def test_empty(self):
+        self.assertEqual(Config("").toJSON(),
+                         '{"fail_fast": false, "matrix": [\
+{"name": "Linux", "os": "linux", "wpt_layout": "none", "profile": "release", \
+"unit_tests": true, "wpt_tests_to_run": ""}, \
+{"name": "MacOS", "os": "macos", "wpt_layout": "none", "profile": "release", \
+"unit_tests": true, "wpt_tests_to_run": ""}, \
+{"name": "Windows", "os": "windows", "wpt_layout": "none", "profile": "release", \
+"unit_tests": true, "wpt_tests_to_run": ""}]}')
```
mrobinson, line 157: It would be great to add a test that simulates two presets one after another, which is what the labels path triggers.

mrobinson (changes requested):

python/servo/devenv_commands.py
sagudev: This is necessary as we add commit on top of current checkout (HEAD) then we push this to remote try branch, so now we need to remove try commit from current checkout, like it was never there.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,174 @@
...
+              makeComment('🔒 User @' + username + ' does not have permission to trigger try jobs.');
+              return "";
+            }
+
+            const url = context.serverUrl +
+              "/" + context.repo.owner +
+              "/" + context.repo.repo +
+              "/actions/runs/" + context.runId;
+            const formattedURL = "[#" + context.runId + "](" + url + ")";
+            makeComment("🔨 Triggering try run (" + formattedURL + ")");
```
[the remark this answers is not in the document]
sagudev: We can do this later after we verify that parser actually spit out valid JSON. Empty try string has check few lines up, so this is unreachable if empty try string.

python/servo/devenv_commands.py
mrobinson: Oh, I get it now. Thanks for the clarification.

python/servo/devenv_commands.py
sagudev: Last time it didn't find git without this, but it work now (at least on my windows machine).

.github/workflows/try_labels.yml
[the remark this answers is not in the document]
sagudev: Should it print JSON code or labels?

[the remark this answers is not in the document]
mrobinson: Good question. Maybe we can print something like just the "os" keys in the configuration "matrix." What do you think?

.github/workflows/try.yml
```diff
@@ -1,203 +1,138 @@
...
+            if (context.eventName == "workflow_dispatch") {
+              // WPT-related overrides only affect Linux currently, as tests don't run by default on other platforms.
+              configuration.matrix[0].wpt_layout = "${{ inputs.wpt-layout }}" || "none";
+              configuration.matrix[0].wpt_tests_to_run = "${{ inputs.wpt-tests-to-run }}" || "";
+
+              let unit_tests = Boolean(${{ inputs.unit-tests }});
+              let profile = '${{ inputs.profile }}';
+              for (const platform of configuration.matrix) {
+                platform.profile = profile;
+                platfrom.unit_tests = unit_tests;
```
mrobinson, line 99: Is the `platform` parameter of the workflow dispatch being honored here?

.github/workflows/try_labels.yml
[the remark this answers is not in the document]
sagudev: How about name from configuration matrix. That's also where job's names come from (Why didn't I come up with this sooner?).

sagudev:
I am working on rebase with android support. I also plan to rename matrix config field from os to workflow as it will be more accurate.

.github/workflows/try.yml
sagudev: there should be no "platform" here, so I removed it from workflow_dispatch.

python/servo/try_parser.py
```diff
@@ -0,0 +1,170 @@
...
+                words.extend(["linux", "macos", "windows", "android"])
+                continue  # skip over keyword
+
+            preset = handle_preset(word)
+            if preset is None:
+                print(f"Ignoring unknown preset {word}")
+            else:
+                self.matrix.append(preset)
+
+    def toJSON(self) -> str:
```
mrobinson, line 127:
```suggestion
    def to_json(self) -> str:
```

python/servo/try_parser.py
```diff
@@ -0,0 +1,170 @@
...
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import json
+from enum import Enum, Flag, auto
+import sys
+from dataclasses import dataclass
```
mrobinson, line 15:
```suggestion
import json
import sys

from dataclasses import dataclass
from enum import Enum, Flag, auto
```

mrobinson (commented):
Is this ready for another review?

sagudev:
This is generally ready for review, but I need to retest some common use-cases to be sure it works right, hence draft.

sagudev:
I think it's ready!

.github/workflows/try.yml
```diff
@@ -1,203 +1,120 @@
...
+        with:
+          fetch-depth: 1
+          sparse-checkout: |
+            python/servo/try_parser.py
+          sparse-checkout-cone-mode: false
+      - name: Generate full
+        id: full
+        run: |
+          out=$(python ./python/servo/try_parser.py)
+          echo "config<<EOF"$'\n'"$out"$'\n'EOF >> $GITHUB_OUTPUT
```
mrobinson, line 45: It's a little unfortunate that this is necessary. How long does this step take? Do you have an example of a run where this takes place that I can look at?

```suggestion
      - name: Get Full Configuration
        id: full_config
        run: |
          {
               echo 'config<<EOF'
               python ./python/servo/try_parser.py
               echo EOF
           } >> $GITHUB_OUTPUT
```
1. Use title-case consistent in step name.
2. Use a [more readable syntax](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-a-multiline-string) for the multi-line string.


Does this depend on the type of quotes that `try_parser.py` produces?

.github/workflows/try.yml
```diff
@@ -1,203 +1,120 @@
...
+        type: boolean
 
 jobs:
-  parse-comment:
-    name: Trigger Try
-    runs-on: ubuntu-latest
-    concurrency:
-      group: try-${{ github.event.number }}
+  decision:
+    name: Generate matrix
```
mrobinson, line 27:
```suggestion
    name: Generate Try Configuration
```

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - uses: actions/checkout@v3
+        if: github.event_name != 'pull_request_target'
+        with:
+          fetch-depth: 1
+          sparse-checkout: |
+            python/servo/try_parser.py
+          sparse-checkout-cone-mode: false
```
mrobinson, line 94: I think you can remove this entirely.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+      # This is necessary to checkout the pull request if this run was triggered
+      # via an `issue_comment` action on a pull request.
+      - uses: actions/checkout@v3
+        if: github.event_name == 'pull_request_target'
```
mrobinson, line 98: This if can be removed.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+          fetch-depth: 1
+          sparse-checkout: |
+            python/servo/try_parser.py
+          sparse-checkout-cone-mode: false
+      - name: Parse Try string
+        if: ${{ steps.try_string.outputs.result }}
+        id: configuration
+        run: |
+          out=$(python ./python/servo/try_parser.py ${{ steps.try_string.outputs.result }})
+          echo "config<<EOF"$'\n'"$out"$'\n'EOF >> $GITHUB_OUTPUT
```
mrobinson, line 110: Can you use the multi-line syntax from above?

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+      # via an `issue_comment` action on a pull request.
+      - uses: actions/checkout@v3
+        if: github.event_name == 'pull_request_target'
+        with:
+          ref: refs/pull/${{ github.event.number }}/head
+          fetch-depth: 1
+          sparse-checkout: |
+            python/servo/try_parser.py
+          sparse-checkout-cone-mode: false
+      - name: Parse Try string
```
mrobinson, line 105:
```suggestion
      - name: Parse Label
```

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+        if: ${{ steps.try_string.outputs.result }}
+        id: configuration
+        run: |
+          out=$(python ./python/servo/try_parser.py ${{ steps.try_string.outputs.result }})
+          echo "config<<EOF"$'\n'"$out"$'\n'EOF >> $GITHUB_OUTPUT
+      - name: Report triggered run
```
mrobinson, line 111:
```suggestion
      - name: Comment Run Start
```

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+      wpt-tests-to-run: ${{ matrix.wpt_tests_to_run }}
+
+  results:
+    name: Results
+    needs: ["parse-comment", "run-try"]
+    runs-on: ubuntu-latest
+    if: ${{ always() && needs.parse-comment.outputs.try_string }}
+    steps:
+      - name: Success
+        if: ${{ github.event_name == 'pull_request_target' && !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
```
mrobinson, line 166:
```suggestion
        if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
```

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+        id: try_string
+        with:
+          result-encoding: string
+          script: |
+            function makeComment(body) {
+                console.log(body);
+
+                if (context.eventName != 'pull_request_target')
+                  return;
+
```
mrobinson, line 28: This can be removed.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+                  issue_number: context.issue.number,
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  body
+                })
+            }
+
+            let try_string = "${{ inputs.try_string }}" || "";
+
+            if (context.eventName == 'pull_request_target') {
```
mrobinson, line 39: This if statement is no longer necessary.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+        with:
+          result-encoding: string
+          script: |
+            let config = ${{ steps.configuration.outputs.config }};
+            function makeComment(body) {
+                console.log(body);
+
+                if (context.eventName != 'pull_request_target')
+                  return;
+
```
mrobinson, line 123: This can be removed.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+            const url = context.serverUrl +
+              "/" + context.repo.owner +
+              "/" + context.repo.repo +
+              "/actions/runs/" + context.runId;
+            const formattedURL = "[#" + context.runId + "](" + url + ")";
+            makeComment("🔨 Triggering try run (" + formattedURL + ") for "
+             + config.matrix.map(m => m.name).join(", "));
+
+
```
mrobinson, line 140: Extra newline here.

.github/workflows/try_labels.yml
```diff
@@ -0,0 +1,198 @@
...
+                "/actions/runs/" + context.runId;
+              const formattedURL = "[#" + context.runId + "](" + url + ")";
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: "✨ Try run (" + formattedURL + ") " + "succeeded.",
+              });
+      - name: Failure
+        if: ${{ github.event_name == 'pull_request_target' && contains(needs.*.result, 'failure') }}
```
mrobinson, line 182:
```suggestion
        if: ${{ contains(needs.*.result, 'failure') }}
```

python/servo/devenv_commands.py
```diff
@@ -284,38 +274,35 @@ def ndk_gdb(self, release, target):
...
+            print("WARNING: You are triggering try build in upstream repo!")
+
+        result = call(["git", "push", remote, "--force", "HEAD:try"])
+        if result != 0:
+            return result
+
+        git_remote = git_remote.replace(".git", "/actions")
+        print(f"You can find triggered workflow here: {git_remote}")
+
+        # we need to delete last commit
```
mrobinson, line 306:
```suggestion
        # Remove the last commit which only contains the try configuration.
```

python/servo/try_parser.py
```diff
@@ -0,0 +1,171 @@
...
+    print(conf.to_json())
+
+
+if __name__ == "__main__":
+    main()
+
+
+import unittest # noqa
+import logging # noqa
+
```
mrobinson, line 143: Please put these at the top of the file.

python/servo/try_parser.py
```diff
@@ -0,0 +1,171 @@
...
+"unit_tests": true, "wpt_tests_to_run": ""}, \
+{"name": "MacOS", "workflow": "macos", "wpt_layout": "none", "profile": "release", \
+"unit_tests": true, "wpt_tests_to_run": ""}, \
+{"name": "Windows", "workflow": "windows", "wpt_layout": "none", "profile": "release", \
+"unit_tests": true, "wpt_tests_to_run": ""}, \
+{"name": "Android", "workflow": "android", "wpt_layout": "none", "profile": "release", \
+"unit_tests": false, "wpt_tests_to_run": ""}]}')
+
+    def test_full(self):
+        self.assertEqual(Config("linux-wpt macos windows android").to_json(), Config("").to_json())
```
mrobinson, line 165: Right now these tests depend entirely on the order in which the keys are produced, which makes them pretty brittle. Please either sort the keys or compare them as dicts.

mrobinson (approved):
Thanks for all your work here! This looks good, but I have some comments that should be addressed before landing.

python/servo/try_parser.py
```diff
@@ -0,0 +1,171 @@
...
+    def default(self, o):
+        if isinstance(o, (Config, JobConfig)):
+            return o.__dict__
+        if isinstance(o, Layout):
+            return o.to_string()
+        return json.JSONEncoder.default(self, o)
+
+
+class Config(object):
+    def __init__(self, s: str | None = None):
```
mrobinson, line 99:
```suggestion
    def __init__(self, s: Optional[str] = None):
```

.github/workflows/try.yml
sagudev: Times are negligible: https://github.com/sagudev/servo/actions/runs/7653133561/job/20854351846 and the only reason for this is so we do not need to hardcode full config in script.

sagudev: There where some problems with quotes, that's why I made as is (possibly due to how GITHUB_OUTPUT is interpreted by runners). I can try your suggestion.

python/servo/try_parser.py
sagudev: | is preferred: https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint, but I changed it anyway.

.github/workflows/try.yml
sagudev: It works.

python/servo/try_parser.py
sagudev: I compare them as dict.

mrobinson (approved):

PR #31146, opened 2024-01-22
layout: Shape text only once

mrobinson:
Shape text during InlineFormattingContext construction rather than doing
it twice during fragment tree construction. This is a step on the way
toward proper font fallback.

This also moves all `TextRun` related code into `text_run.rs` to try to
trim down the size of `inline.rs`.

---

- [x] These changes do not require tests because this should only have performance impacts.


Loirooriol (approved):
Nice, but I think this will need a rebase due to #31169

PR #31148, opened 2024-01-22
build: Add support for Visual Studio 2022 and VC143 DLLs

mrobinson:
This change adds supports for Visual Studio 2022 and the VC143 (current) version of the Visual Studio CRT. In addition, it reworks the way that Visual Studio is found, returning all installations in a generator, separately finding it via `vswhere.exe`, searching paths, and via environment variables.

All of these installations are searched for the DLLs with highest priority given to the highest version of MS Visual Studio installed. The hope is that this makes the process more robust and properly handles having multiple versions installed, but only one with the correct runtime DLLs.

In addition, build documentation is updated to reflect that choco is doing much less. The reason for this change to the choco configuration is that now rustup is responsible for installing Visual Studio 2022 or the user installs the version that they'd like.

---

- [x] These changes do not require tests because this is just a build configuration change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ manually, try the [manual build setup][manual-build].
  - Install [Python 3.11](https://www.python.org/downloads/windows/)
  - Run `mach bootstrap`
     - *This will install CMake, Git, Ninja, via choco in an
+    + *This will install CMake, Git, and Ninja via choco in an
        Administrator console. Allow the scripts to run and once
        the operation finishes, close the new console.*
 - Run `refreshenv`
--- a/README.md
+++ b/README.md
@@ -41,14 +41,20 @@ manually, try the [manual build setup][manual-build].
 
 ### Windows
 
- - Download and run [`rustup-init.exe`](https://win.rustup.rs/) then follow the onscreen instructions.
+ - Download and run [`rustup-init.exe`](https://win.rustup.rs/)
+  - Make sure to select *Quick install via the Visual Studio Community
+    installer* or otherwise install Visual Studio 2022.
+ - In the *Visual Studio Installer* ensure the following components are installed for Visual Studio 2022:
+    - **Windows 10 SDK (10.0.19041.0)** (`Microsoft.VisualStudio.Component.Windows10SDK.19041`)
+    - **MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)** (`Microsoft.VisualStudio.Component.VC.Tools.x86.x64`)
+    - **C++ ATL for latest v143 build tools (x86 & x64)** (`Microsoft.VisualStudio.Component.VC.ATL`)
+    - **C++ MFC for latest v143 build tools (x86 & x64)** (`Microsoft.VisualStudio.Component.VC.ATLMFC`)
  - Install [chocolatey](https://chocolatey.org/)
- - Install [Python 3.11](https://apps.microsoft.com/detail/9NRWMJP3717K?hl=en-US&gl=US)
+ - Install [Python 3.11](https://www.python.org/downloads/windows/)
  - Run `mach bootstrap`
-  - *This will install CMake, Git, Ninja, and the Visual Studio 2019 Build Tools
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -41,14 +41,20 @@ manually, try the [manual build setup][manual-build].
...
+    - **C++ MFC for latest v143 build tools (x86 & x64)** (`Microsoft.VisualStudio.Component.VC.ATLMFC`)
  - Install [chocolatey](https://chocolatey.org/)
- - Install [Python 3.11](https://apps.microsoft.com/detail/9NRWMJP3717K?hl=en-US&gl=US)
+ - Install [Python 3.11](https://www.python.org/downloads/windows/)
  - Run `mach bootstrap`
-  - *This will install CMake, Git, Ninja, and the Visual Studio 2019 Build Tools
-     via choco in an Administrator console. It can take quite a while.*
-  - *If you already have Visual Studio 2019 installed, this may not install all necessary components.
-     Please follow the Visual Studio 2019 installation instructions in the [manual setup][manual-build].*
+    - *This will install CMake, Git, Ninja, via choco in an
```
mukilan, line 55:
```suggestion
    - *This will install CMake, Git, Ninja, via choco in an
    + *This will install CMake, Git, and Ninja via choco in an
```

python/servo/build_commands.py
```diff
@@ -433,29 +433,37 @@ def package_gstreamer_dlls(servo_exe_dir: str, target: str):
...
 
-    for msvc_dll in missing:
-        print(f"Could not find DLL dependency: {msvc_dll}")
-    return not missing
+    for msvc_redist_dir in find_msvc_redist_dirs(vs_platform):
+        if copy_file(os.path.join(msvc_redist_dir, "msvcp140.dll")) and \
+           copy_file(os.path.join(msvc_redist_dir, "vcruntime140.dll")):
+            break
+
+    if "aarch64" not in target != "aarch64":
```
mukilan, line 460: Is this conditional correct?

python/servo/visual_studio.py
```diff
@@ -12,80 +12,141 @@
...
     return VisualStudioInstallation(
         version_number=version_number,
-        installation_path=vsinstalldir,
+        installation_path=installation_path,
         vc_install_path=vc_install_path,
     )
 
 
-def find_windows_sdk_installation_path(vs_platform: str) -> str:
+def find_msvc_installations() -> VisualStudioInstallation:
```
mukilan, line 105: Doesn't the return type need to be wrapped in `Generator` type?

mukilan: Also, does it need to be a generator? it looks like the only call site is where we convert it into a list.

python/servo/visual_studio.py
```diff
@@ -12,80 +12,141 @@
...
+                for path in [path1, path2]:
+                    path = os.path.join(redist_path, path)
+                    if os.path.isdir(path):
+                        yield path
+                    else:
+                        tried.append(path)
+
+    print("Couldn't locate MSVC redistributable directory. Tried:", file=sys.stderr)
+    for path in tried:
+        print(f"  * {path}")
```
mukilan, line 145: do we not need to redirect this to `sys.stderr` as well?

python/servo/visual_studio.py
mrobinson: That said, maybe the code is a little cleaner if it just returns a list. I'll go ahead and do that.

python/servo/build_commands.py
mrobinson: I think it isn't. Additionally, I suspect this is left over from the UWP build which didn't use the Windows SDK in the same way. I'll just go ahead and remove it.

python/servo/visual_studio.py
mrobinson: Yep, I think so. I'll fix that.

mrobinson:
@mukilan Thanks for the review! I've updated the branch according to your comments.

mukilan (approved):

Issue #31064, opened 2024-01-11
WebIDL: use TypedArray


gterzian:
As as follow-up to https://github.com/servo/servo/pull/30990, we should do the same for other typed arrays. 

The list is found at https://github.com/servo/servo/blob/92196d985dceb0ca708b097e2a847b255d8387c8/third_party/WebIDL/WebIDL.py#L3634

See the various arrays which currently default to `IDLType.Tags.interface`, and which currently all fallback to using a `JSObject` in [codegen](https://github.com/servo/servo/blob/340a86c37584079e6387fe7dec8b8bbf3c722411/components/script/dom/bindings/codegen/CodegenRust.py). Each should be changed to use it's own type, like was done in the linked-to PR for `FLoat32Array`. Note that not all may be currently used by Servo, to find out if they are, search for it in `*.webidl` files. 

Each could be done independently, although it might make sense to investigate making [`components/script/dom/bindings/typedarrays.rs`](https://github.com/servo/servo/blob/340a86c37584079e6387fe7dec8b8bbf3c722411/components/script/dom/bindings/typedarrays.rs) generic first. 

If the typed array is currently stored in a `Heap`, then a wrapper like [`HeapFloat32Array`](https://github.com/servo/servo/pull/30990/files#diff-8f10f5d3caee3cce7befe080a3333a5f9402a651884287734f41e786e4323181R30) would be useful, otherwise the unsafe code can simply be hidden behind utilities, such as [`create_float32_array`](https://github.com/servo/servo/pull/30990/files#diff-8f10f5d3caee3cce7befe080a3333a5f9402a651884287734f41e786e4323181R16)

The actual typed array JS types are found at https://github.com/servo/mozjs/blob/7184f658b7e2de8a8d288ff39001364513115782/mozjs/src/typedarray.rs

part of https://github.com/servo/servo/issues/30862

Taym95:
@gterzian @mrobinson this issue can be closed now.

gterzian:
@Taym95 thanks a lot for the contrib!

PR #31164, opened 2024-01-24
Use Int8array, int16array, uint16array, int32array & uint32array in WebIDL

Taym95:
part #31064

---

- [x] These changes do not require tests because there is no behavior changes

There types are not used we can migrate them, in next PR I will make `HeapFloat32Array` generic and convert ArrayBufferView, ArrayBuffer, Uint8ClampedArray and Float64Array.


Taym95:
@gterzian can we merge this one?

gterzian (approved):
Thanks! This way if any of these were to be used in a web idl in the future(will not happen anytime soon I think), the user would have to use to soon to follow generic typed array.

PR #31167, opened 2024-01-24
Make HeapFloat32Array generic

Taym95:
part #31064

---

### DIFF
--- a/components/script/dom/bindings/typedarrays.rs
+++ b/components/script/dom/bindings/typedarrays.rs
@@ -10,8 +10,7 @@ use js::jsapi::{Heap, JSObject, JS_GetArrayBufferViewBuffer};
 use js::rust::wrappers::DetachArrayBuffer;
 use js::rust::{CustomAutoRooterGuard, MutableHandleObject};
 use js::typedarray::{
-    CreateWith, Float32Array, JSObjectStorage, TypedArray, TypedArrayElement,
-    TypedArrayElementCreator,
+    CreateWith, Float32, Float32Array, TypedArray, TypedArrayElement, TypedArrayElementCreator,
 };
 
 use crate::script_runtime::JSContext;
@@ -21,15 +20,17 @@ pub struct HeapFloat32Array {
     internal: Box<Heap<*mut JSObject>>,
 }
 
-impl HeapFloat32Array {
-    pub fn set_data(&self, cx: JSContext, data: &[f32]) -> Result<(), ()> {
+impl HeapTypedArray for HeapFloat32Array {
+    type ArrayType = Float32;
+
+    fn set_data(&self, cx: JSContext, data: &[f32]) -> Result<(), ()> {
         rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
         let _: Float32Array = create_typed_array(cx, data, array.handle_mut())?;
         self.internal.set(*array);
         Ok(())
     }
 
-    pub fn acquire_data(&self, cx: JSContext) -> Result<Vec<f32>, ()> {
+    fn acquire_data(&self, cx: JSContext) -> Result<Vec<f32>, ()> {
         assert!(self.is_initialized());
         typedarray!(in(*cx) let array: Float32Array = self.internal.get());
         let data = if let Ok(array) = array {
@@ -50,7 +51,7 @@ impl HeapFloat32Array {
         data
     }
 
-    pub fn copy_data_to(
+    fn copy_data_to(
         &self,
         cx: JSContext,
         dest: &mut [f32],
@@ -67,7 +68,7 @@ impl HeapFloat32Array {
         Ok(())
     }
 
-    pub fn copy_data_from(
+    fn copy_data_from(
         &self,
         cx: JSContext,
         source: CustomAutoRooterGuard<Float32Array>,
@@ -85,15 +86,15 @@ impl HeapFloat32Array {
         Ok(())
     }
 
-    pub fn is_initialized(&self) -> bool {
+    fn is_initialized(&self) -> bool {
         !self.internal.get().is_null()
     }
 
-    pub fn get_internal(&self) -> Result<Float32Array, ()> {
+    fn get_internal(&self) -> Result<Float32Array, ()> {
         Float32Array::from(self.internal.get())
     }
 
-    pub fn internal_to_option(&self) -> Option<Float32Array> {
+    fn internal_to_option(&self) -> Option<Float32Array> {
         if self.is_initialized() {
             Some(self.get_internal().expect("Failed to get internal."))
         } else {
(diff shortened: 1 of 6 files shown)


components/script/dom/bindings/typedarrays.rs
```diff
@@ -9,30 +9,36 @@ use std::ptr;
...
-    CreateWith, Float32Array, JSObjectStorage, TypedArray, TypedArrayElement,
-    TypedArrayElementCreator,
-};
+use js::typedarray::{CreateWith, TypedArray, TypedArrayElement, TypedArrayElementCreator};
 
 use crate::script_runtime::JSContext;
 
 #[derive(Default, JSTraceable)]
-pub struct HeapFloat32Array {
+pub struct HeapTypedArray {
```
Taym95, line 17: @gterzian please review, I tried to make it HeapTypedArray<T> but since T is not used here I added `PhantomData ` but I got errors because of `JSTraceable `

sagudev: on phantom field you need to add `#[no_trace]`

gterzian (changes requested):
Great work, thanks! I think it's a very good idea to add a phantom type, and make it no trace as described at https://github.com/servo/servo/pull/31167/files#r1465417408

Can this phantom type also be used as the generic parameter to the various methods? 

Perhaps an alternative would be a trait `HeapTypedArray`, with a `type ArrayType;`, and then have specific structs implement the trait but only by setting the `ArrayType`(with all methods being defaults). So for example `HeapFloat32Array` would be left as is, but only implement `HeapTypedArray` and set the type to `Float32`, or something along those lines.

Taym95:
> Great work, thanks! I think it's a very good idea to add a phantom type, and make it no trace as described at https://github.com/servo/servo/pull/31167/files#r1465417408
> 
> Can this phantom type also be used as the generic parameter to the various methods?
> 
> Perhaps an alternative would be a trait `HeapTypedArray`, with a `type ArrayType;`, and then have specific structs implement the trait but only by setting the `ArrayType`(with all methods being defaults). So for example `HeapFloat32Array` would be left as is, but only implement `HeapTypedArray` and set the type to `Float32`, or something along those lines.

I added `HeapTypedArray ` trait and implemented the trait for `HeapFloat32Array`

sagudev:
I think using traits will require code duplication for each type, so if possible I would peruse generic approach using phantom type and as far as I know same generic can also be required in methods.

gterzian:
> I added HeapTypedArray trait and implemented the trait for HeapFloat32Array

Thank you!

Is it possible to make all the methods defaults on the trait, using `Self::ArrayType`, so that the actual implementation of the trait is only used to set `ArrayType` and is otherwise empty(If the `typedarray` macro doesn't like it, I would drop the macro)? If not, I agree with @sagudev that using a phantom type is better.

Taym95:
> I think using traits will require code duplication for each type, so if possible I would peruse generic approach using phantom type and as far as I know same generic can also be required in methods.

I added all the methods defaults on the trait to avoid  code duplication for each type, could not use `PhantomData<T>` with `#[no_trace]` because of:

`cannot find type `T` in this scope
not found in this scope`

sagudev:
> could not use `PhantomData<T>` with `#[no_trace]` because of:
> 
> `cannot find type T in this scope not found in this scope`

Do you have WIP branch of it, I would like to see what macro expansion does.

Manually implementing JSTraceable trait should work.

Taym95:
> > could not use `PhantomData<T>` with `#[no_trace]` because of:
> > `cannot find type T in this scope not found in this scope`
> 
> Do you have WIP branch of it, I would like to see what macro expansion does.
> 
> Manually implementing JSTraceable trait should work.

somehow macro not handling generic type, branch:

https://github.com/Taym95/servo/tree/make-HeapFloat32Array-generic-phantom

sagudev:
> somehow macro not handling generic type

Yes, indeed.

For such cases manual implementation is needed:
```rust
#[allow(unsafe_code)]
unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
    #[inline]
    unsafe fn trace(&self, tracer: *mut js::jsapi::JSTracer) {
        self.internal.trace(tracer);
    }
}
```

sagudev:
Also `TypedArrayElementCreator + TypedArrayElement` constraints on generic might be more sensible directly on struct IMO.

Taym95:
> > somehow macro not handling generic type
> 
> Yes, indeed.
> 
> For such cases manual implementation is needed:
> 
> ```rust
> #[allow(unsafe_code)]
> unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
>     #[inline]
>     unsafe fn trace(&self, tracer: *mut js::jsapi::JSTracer) {
>         self.internal.trace(tracer);
>     }
> }
> ```

Thanks @sagudev I used generic type and last commit. and kept old trait commit

components/script/dom/bindings/typedarrays.rs
```diff
@@ -4,40 +4,59 @@
...
-#[derive(Default, JSTraceable)]
-pub struct HeapFloat32Array {
+pub struct HeapTypedArray<T> {
     internal: Box<Heap<*mut JSObject>>,
+    phantom: PhantomData<T>,
 }
 
-impl HeapFloat32Array {
-    pub fn set_data(&self, cx: JSContext, data: &[f32]) -> Result<(), ()> {
+#[allow(unsafe_code)]
```
gterzian, line 22: This is unnecessary because the entire module allows unsafe code(see top of file).

components/script/dom/bindings/typedarrays.rs
```diff
@@ -4,40 +4,59 @@
...
+        typedarray!(in(*cx) let array: TypedArray = self.internal.get());
+        let data = if let Ok(array) =
+            array as Result<CustomAutoRooterGuard<'_, TypedArray<T, *mut JSObject>>, &mut ()>
+        {
             let data = array.to_vec();
             let mut is_shared = false;
             unsafe {
                 rooted!(in (*cx) let view_buffer =
-                    JS_GetArrayBufferViewBuffer(*cx, self.internal.handle(), &mut is_shared));
+                     JS_GetArrayBufferViewBuffer(*cx, self.internal.handle(), &mut is_shared));
```
gterzian, line 59: is this an fmt change?

components/script/dom/bindings/typedarrays.rs
```diff
@@ -103,16 +127,15 @@ impl HeapFloat32Array {
...
+pub fn create_typed_array<T>(
     cx: JSContext,
     data: &[T::Element],
     dest: MutableHandleObject,
-) -> Result<TypedArray<T, S>, ()>
+) -> Result<TypedArray<T, *mut JSObject>, ()>
 where
-    T: TypedArrayElementCreator + TypedArrayElement,
-    S: JSObjectStorage,
+    T: TypedArrayElement + TypedArrayElementCreator,
```
gterzian, line 136: Why no leave the generic `JSObjectStorage` bound and use `S` instead of `JSObject`?

gterzian (changes requested):
great work, thanks! Just one nit and some questions.

components/script/dom/bindings/typedarrays.rs
Taym95: no I will fix it

components/script/dom/bindings/typedarrays.rs
Taym95: to avoid adding generic `S` to create_typed_array when I call it from `set_data` or `TextEncoder`

gterzian (approved):
LGTM, thanks! And thanks for your help @sagudev

PR #31168, opened 2024-01-24
Replace time with std::time in components/gfx

### LINKED ISSUE #30150: Replace uses of `time` (`precise_time_ns`) with `std::time`, `chrono`, or the new `time` crate

devjos:
- [x] These changes are part of #30150

- [x] These changes do not require tests because no change in functionality

### DIFF
--- a/components/gfx/font.rs
+++ b/components/gfx/font.rs
@@ -8,6 +8,7 @@ use std::collections::HashMap;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use std::time::Instant;
 use std::{iter, str};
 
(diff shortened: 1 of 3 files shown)


components/gfx/font.rs
```diff
@@ -253,7 +254,7 @@ impl Font {
             .borrow_mut()
             .entry(lookup_key)
             .or_insert_with(|| {
-                let start_time = time::precise_time_ns();
```
Taym95, line 256: you can remove `time = { workspace = true }` now from `components/gfx/Cargo.toml`

mrobinson (approved):

PR #31169, opened 2024-01-25
Lint layout_2020 with clippy

Loirooriol:
`cargo clippy --fix -p layout_2020 --allow-dirty --broken-code`

---

- [X] These changes do not require tests because it's just linting

### DIFF
--- a/components/layout_2020/context.rs
+++ b/components/layout_2020/context.rs
@@ -110,7 +110,7 @@ impl<'a> LayoutContext<'a> {
             .read()
             .get(&(url.clone(), use_placeholder))
         {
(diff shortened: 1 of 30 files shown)


Loirooriol:
I guess the old title could be interpreted as happening automatically from now on. But I meant that clippy still has more complaints, this only contains the fixes that can be applied automatically.

These are the details of what I did:

1. Temporarily comment out `rustc = "crown"` in `.cargo/config.toml`
2. Run `cargo clippy --fix -p layout_2020 --allow-dirty --broken-code`
3. Run `./mach fmt`
4. Move the new `impl Default` to a better place

mrobinson (approved):

PR #31181, opened 2024-01-26
Update Servo::get_events to return drain type

wusyong:
Another improvement I hope it can see. Servo created a new vector every time when embedder calls `get_events.`
Using drain type could prevent this allocation.
I also remove the type alias for TopLevelBrowsingContextId because Servoshell and I are all aliases with other names.
It's better to keep the original name and re-export it.

---

- [x] These changes do not require tests but servoshell should be built successfully.


mrobinson (approved):

mrobinson:
@wusyong Looks like this breaks the Android build. Likely you need to make some changes in libsimpleservo.

PR #31184, opened 2024-01-26
gfx: Rename `WebrenderSurfman` to `RenderingContext` and move to `gfx`

mrobinson:
This is a small cleanup that moves and renames this class. The rename is
simply because we are exposing a lot about the details of Servo's
rendering in the API and it makes sense to start thinking about
abstracting that away a bit.

This also moves the struct to `gfx`, which does have an effect on
Servo's dependency graph. This adds a new dependency on gfx to
`compositing`, but `compositing` had a transitive dependency on
gfx before through `canvas`.

---

- [x] These changes do not require tests because they don't change behavior.


jdm (approved):
This seems reasonable!

wusyong:
Happy to see the change as well!
I'm wondering what will happen to [WindowMethods](https://github.com/servo/servo/blob/main/components/compositing/windowing.rs#L166).
I think no matter what it's better to have to trait to customize. Especially like `get_coordinates`, when the window implementation might differ from various sources. It can be gtk window, raw window, winit window, or even headless.

mrobinson:
> Happy to see the change as well! I'm wondering what will happen to [WindowMethods](https://github.com/servo/servo/blob/main/components/compositing/windowing.rs#L166). I think no matter what it's better to have to trait to customize. Especially like `get_coordinates`, when the window implementation might differ from various sources. It can be gtk window, raw window, winit window, or even headless.

I think the GL setup aspects of `WindowMethods` are used mainly for setting up GL video rendering. I've tried to move this into libservo here: https://github.com/servo/servo/pull/31209

Hopefully this reduces the amount of code duplicate between wry and simpleservo.

PR #31186, opened 2024-01-26
Replace time with std::time in components/style

devjos:
- [x] These changes are part of #30150

- [x] These changes do not require tests because no change in functionality


mrobinson:
Sorry, let's not touch style as we vendor that from Gecko upstream. I'll update the bug to reflect this.

PR #31187, opened 2024-01-26
layout: Round `clientTop`, etc queries to pixels properly

mrobinson:
Instead of just flooring all pixels in `clientLeft`, `clientTop`, `clientWidth`
and `clientHeight` queries, we should round the rectangle.

---

### DIFF
--- a/components/layout_2020/fragment_tree/fragment_tree.rs
+++ b/components/layout_2020/fragment_tree/fragment_tree.rs
@@ -156,18 +156,20 @@ impl FragmentTree {
                 return Some(Rect::zero());
             }
 
-            let padding_rect = padding_rect.to_physical(style.writing_mode, containing_block);
             let border = style.get_border();
-            Some(Rect::new(
-                Point2D::new(
-                    border.border_left_width.to_px(),
-                    border.border_top_width.to_px(),
-                ),
-                Size2D::new(
-                    padding_rect.size.width.px() as i32,
-                    padding_rect.size.height.px() as i32,
-                ),
-            ))
+            let padding_rect = padding_rect.to_physical(style.writing_mode, containing_block);
+            Some(
+                Rect::new(
+                    Point2D::new(
+                        border.border_left_width.to_f32_px(),
+                        border.border_top_width.to_f32_px(),
+                    ),
+                    Size2D::new(padding_rect.size.width.px(), padding_rect.size.height.px()),
+                )
+                .round()
+                .to_i32()
+                .to_untyped(),
+            )
         })
         .unwrap_or_else(Rect::zero)
     }
--- a/components/layout_2020/query.rs
+++ b/components/layout_2020/query.rs
@@ -211,9 +211,12 @@ pub fn process_node_scroll_area_request(
     };
 
     Rect::new(
-        Point2D::new(rect.origin.x.px() as i32, rect.origin.y.px() as i32),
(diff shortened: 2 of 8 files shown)


mrobinson:
Looks like there is an open spec issue about this here: https://github.com/w3c/csswg-drafts/issues/9866

Perhaps it would be better to modify the query interface to accept f32 and then do the rounding as a final step in the script crate. I will consider this.

Loirooriol:
I think this may be better as

```diff
--- a/components/layout_2020/fragment_tree/fragment_tree.rs
+++ b/components/layout_2020/fragment_tree/fragment_tree.rs
@@ -164,8 +164,8 @@ impl FragmentTree {
                     border.border_top_width.to_px(),
                 ),
                 Size2D::new(
-                    padding_rect.size.width.px() as i32,
-                    padding_rect.size.height.px() as i32,
+                    padding_rect.size.width.px().round() as i32,
+                    padding_rect.size.height.px().round() as i32,
                 ),
             ))
         })
```

Because here browsers say `clientTop` is `5` instead of rounding to `6`:

```html
<!DOCTYPE html>
<div id="target" style="border-top: 5.9px solid; width: 9.5px"></div>
<script>document.writeln(target.clientTop);</script>
```

mrobinson:
> I think this may be better as
> 
> 
> 
> ```diff
> 
> --- a/components/layout_2020/fragment_tree/fragment_tree.rs
> 
> +++ b/components/layout_2020/fragment_tree/fragment_tree.rs
> 
> @@ -164,8 +164,8 @@ impl FragmentTree {
> 
>                      border.border_top_width.to_px(),
> 
>                  ),
> 
>                  Size2D::new(
> 
> -                    padding_rect.size.width.px() as i32,
> 
> -                    padding_rect.size.height.px() as i32,
> 
> +                    padding_rect.size.width.px().round() as i32,
> 
> +                    padding_rect.size.height.px().round() as i32,
> 
>                  ),
> 
>              ))
> 
>          })
> 
> ```
> 
> 
> 
> Because here browsers say `clientTop` is `5` instead of rounding to `6`:
> 
> 
> 
> ```html
> 
> <!DOCTYPE html>
> 
> <div id="target" style="border-top: 5.9px solid; width: 9.5px"></div>
> 
> <script>document.writeln(target.clientTop);</script>
> 
> ```

Will this do the right thing for clientTop if the top is positioned at a pixel offset greater than 0.5?

Loirooriol:
Mm, no, it's broken, for `border-top: 5px solid` it says 4 :facepalm: Is there another bug here?

In fact, the computed value of the border widths should already be an integer multiple of a device pixel, so this is all quite weird...

Loirooriol:
Actually, the problem was that Servo wasn't using a device pixel ratio of 1, even if it should be 1 on my system.

This seems to match other browsers, then. I have added the analogous fix for scrollWidth and scrollHeight.

I have also added a big test, in its crude form it was passing in servo (except for scrollTop and scrollLeft since scrolling isn't supported). But when converting it into a testharness it stopped working because of #31195

Note that the title is incorrect, `getClientRects` and `getBoundingClientRect` do not round.

mrobinson:
These additions look goog to me @Loirooriol. Thanks for the great test.

Loirooriol (approved):

mrobinson:
Looks like the new test needs failing expectations for layout 2013.

PR #31189, opened 2024-01-26
webidlg: Handle `Float64Array` as a `TypedArray` rather than a raw `JSObject`

Taym95:
part #31064

depends on #31167 

---

### DIFF
--- a/components/script/dom/dommatrixreadonly.rs
+++ b/components/script/dom/dommatrixreadonly.rs
@@ -691,13 +691,7 @@ impl DOMMatrixReadOnlyMethods for DOMMatrixReadOnly {
 
     // https://drafts.fxtf.org/geometry-1/#dom-dommatrixreadonly-tofloat64array
     fn ToFloat64Array(&self, cx: JSContext) -> Float64Array {
-        let vec: Vec<f64> = self
-            .matrix
-            .borrow()
-            .to_array()
-            .iter()
-            .map(|&x| x as f64)
-            .collect();
+        let vec: Vec<f64> = self.matrix.borrow().to_array().to_vec();
 
         rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
         create_typed_array(cx, &vec, array.handle_mut())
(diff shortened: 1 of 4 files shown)


Taym95:
@gterzian this is also ready for review

components/script/dom/dommatrixreadonly.rs
```diff
@@ -691,14 +690,18 @@ impl DOMMatrixReadOnlyMethods for DOMMatrixReadOnly {
...
-            NonNull::new_unchecked(array.get())
-        }
+    fn ToFloat64Array(&self, cx: JSContext) -> Float64Array {
+        let vec: Vec<f64> = self
+            .matrix
+            .borrow()
+            .to_array()
+            .iter()
+            .map(|&x| x as f64)
+            .collect();
```
gterzian, line 700: Can we not just use `to_array`, since the matrix is typed to `f64`?

gterzian (approved):
LGTM, with one question.

components/script/dom/dommatrixreadonly.rs
Taym95: Ah yes we can

components/script/dom/dommatrixreadonly.rs
```diff
@@ -691,14 +690,12 @@ impl DOMMatrixReadOnlyMethods for DOMMatrixReadOnly {
...
-    #[allow(unsafe_code)]
-    fn ToFloat64Array(&self, cx: JSContext) -> NonNull<JSObject> {
-        let arr = self.matrix.borrow().to_array();
-        unsafe {
-            rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
-            let _ = Float64Array::create(*cx, CreateWith::Slice(&arr), array.handle_mut()).unwrap();
-            NonNull::new_unchecked(array.get())
-        }
+    fn ToFloat64Array(&self, cx: JSContext) -> Float64Array {
+        let vec: Vec<f64> = self.matrix.borrow().to_array().to_vec();
```
gterzian, line 694: Ok, and what about just `create_typed_array(cx, &self.matrix.borrow().to_array(), ...` to avoid allocating a new vec?

Taym95: Indeed not needed

gterzian: Great, it would be nice to have a look at the previously merged code to see if we missed similar opportunities...

PR #31200, opened 2024-01-27
Implement non-XR Gamepad discovery and input

### LINKED ISSUE #31179: Implement non-XR gamepad discovery and input
Recently I've been doing some work on Servo's WebXR crate, but while doing that I noticed that the Gamepad API wasn't supported for regular controllers and thought it would be a fun project to take on!

Spec: https://www.w3.org/TR/gamepad/

There is some existing code in Servo already for Gamepads, but it was added for use with WebVR/WebXR and as such doesn't support discovery/communication with regular game controllers. My plan is to utilize [GilRs (Game Input Library for Rust)](https://gitlab.com/gilrs-project/gilrs) as the backend for gamepad discovery/communication. From my understanding this will also involve creating a new task source for communication with Constellation and handling polling and such on the embedder's event loop.

msub2:
Implement discovery and input updating for non-XR controllers for use with the [Gamepad API](https://www.w3.org/TR/gamepad/).

---

- [x] These changes fix #31179


components/script/script_thread.rs
```diff
@@ -3669,6 +3669,14 @@ impl ScriptThread {
                 };
                 document.dispatch_composition_event(composition_event);
             },
+
+            GamepadEvent(gamepad_event) => {
+                let document = match self.documents.borrow().find_document(pipeline_id) {
+                    Some(document) => document,
```
Taym95, line 3675: maybe you can call dispatch_gamepad_event here and drop `let document` ?

```suggestion
                    Some(document) => document.dispatch_gamepad_event(gamepad_event),
```

msub2:
Okay, it's not 100% perfect yet but I think ready for a more complete review now. Some caveats of the current implementation:

- Anything related to haptic actuators has not been implemented yet. The spec updated pretty recently on this and it'll require some WebIDL work so I'm saving it for a future PR.
- Anything related to gamepad permissions is also TODO, I will also tackle that in a future PR.
- Right now gamepad input seems to not get updated until a window event happens i.e. mouse moving. Mentioned this on Zulip, might just be where I'm handling events in servoshell right now.
- A gamepad that is already plugged in upon starting servoshell will not be detected, the Embedder connection event seems to be getting dropped in constellation from not having a browsing context to reach

msub2:
I've added spec comments where applicable. Here's the full flow for reference:

# Embedder

In `ports/servoshell/app.rs`, we call `handle_gamepad_events` in the `handle_events` function. `handle_gamepad_events` is a function in `ports/servoshell/webview.rs` that handles pumping messages from GilRs, sending controller connect/disconnect events, and updating axis/button values (after mapping to standard gamepad in the spec) via EmbedderEvents, 

# Constellation
Constellation receives these EmbedderEvents and handles them in `handle_gamepad_msg` in `components/constellation/constellation.rs`.

# Script thread

Constellation sends the gamepad events to the script thread, where it then gets dispatched to the document via `document.dispatch_gamepad_event`. Here we finally handle the specific gamepad event (connect, disconnect, update) via queueing a task on the newly created gamepad task source.

components/script/dom/gamepad.rs
```diff
@@ -66,6 +67,57 @@ impl Gamepad {
...
+            GamepadButton::new(global, false, false), // Left stick pressed button
+            GamepadButton::new(global, false, false), // Right stick pressed button
+            GamepadButton::new(global, false, false), // Top button in left cluster
+            GamepadButton::new(global, false, false), // Bottom button in left cluster
+            GamepadButton::new(global, false, false), // Left button in left cluster
+            GamepadButton::new(global, false, false), // Right button in left cluster
+            GamepadButton::new(global, false, false), // Center button in center cluster
+        ];
+        rooted_vec!(let buttons <- standard_buttons.iter().map(|button| DomRoot::from_ref(&**button)));
+        let button_list = GamepadButtonList::new(global, buttons.r());
```
gterzian, line 103: It would be good to:
1. move the #dfn-initializing-buttons logic to a separate method, like you did with init axes
2. Add comments about TODOs

components/script/dom/gamepad.rs
```diff
@@ -143,10 +194,44 @@ impl Gamepad {
...
+
+    #[allow(unsafe_code)]
+    pub fn update_axis(&self, axis_index: usize, value: f32) {
+        let mut axis_vec = self.axes.get_internal().unwrap();
+        unsafe {
+            axis_vec.as_mut_slice()[axis_index] = value as f64;
+        }
+    }
+
+    pub fn update_button(&self, button_index: usize, pressed: bool, touched: bool, value: f32) {
```
gterzian, line 232: For this and other methods, instead of adding a comment at the call site, it would be better to have a name that matches the spec(here `map_and_normalize_buttons`), and add a doc comment with the link <https://w3c.github.io/gamepad/#dfn-map-and-normalize-buttons>.

A comment at the call site then becomes unnecessary.

components/script/dom/gamepad.rs
```diff
@@ -66,6 +67,57 @@ impl Gamepad {
...
+            GamepadButton::new(global, false, false), // Right button in center cluster
+            GamepadButton::new(global, false, false), // Left stick pressed button
+            GamepadButton::new(global, false, false), // Right stick pressed button
+            GamepadButton::new(global, false, false), // Top button in left cluster
+            GamepadButton::new(global, false, false), // Bottom button in left cluster
+            GamepadButton::new(global, false, false), // Left button in left cluster
+            GamepadButton::new(global, false, false), // Right button in left cluster
+            GamepadButton::new(global, false, false), // Center button in center cluster
+        ];
+        rooted_vec!(let buttons <- standard_buttons.iter().map(|button| DomRoot::from_ref(&**button)));
```
gterzian, line 102: I think you can just do `vec![GamepadButton::new...]`, since the number is not arbitrary(see https://github.com/servo/servo/blob/f6b81a97f39a157347adc13d312e2ee5fad881d3/components/script/dom/bindings/trace.rs#L483)

gterzian (changes requested):
Great start, thanks! In general, instead of adding comments for steps, I prefer to try to break things down into method mapping to a link in the spec(because the step numbers usually change). Which steps are taken inside the method is usually clear enough that no comment is necessary. Comments for TODO's are useful on the other hand. 

This is more like a first pass, with some suggested structural changes.

components/script/dom/gamepad.rs
```diff
@@ -143,10 +194,44 @@ impl Gamepad {
...
+
     pub fn notify_event(&self, event_type: GamepadEventType) {
         let event = GamepadEvent::new_with_type(&self.global(), event_type, &self);
         event
             .upcast::<Event>()
             .fire(self.global().as_window().upcast::<EventTarget>());
     }
+
+    // https://www.w3.org/TR/gamepad/#dfn-initializing-axes
+    // https://www.w3.org/TR/gamepad/#fingerprinting-mitigation
```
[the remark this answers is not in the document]
msub2: Done, will keep that consistent going forward

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                            let gamepad = Gamepad::new(&global, index.0 as u32, name);
+                            // 3.2 Let navigator be gamepad's relevant global object's Navigator object.
+                            // 3.3 Set navigator.[[gamepads]][gamepad.index] to gamepad.
+                            let gamepad_list = window.Navigator().GetGamepads();
+                            let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
+                            gamepad_list.add_if_not_exists(&gamepad_arr);
+
+                            // 3.4 If navigator.[[hasGamepadGesture]] is true:
+                            // 3.4.1 Set gamepad.[[exposed]] to true.
+                            // 3.4.2 If document is not null and is fully active, then fire an event named
```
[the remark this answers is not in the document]
msub2: Done

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                            // 3.1 Let gamepad be a new Gamepad representing the gamepad.
+                            let gamepad = Gamepad::new(&global, index.0 as u32, name);
+                            // 3.2 Let navigator be gamepad's relevant global object's Navigator object.
+                            // 3.3 Set navigator.[[gamepads]][gamepad.index] to gamepad.
+                            let gamepad_list = window.Navigator().GetGamepads();
+                            let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
+                            gamepad_list.add_if_not_exists(&gamepad_arr);
+
+                            // 3.4 If navigator.[[hasGamepadGesture]] is true:
+                            // 3.4.1 Set gamepad.[[exposed]] to true.
```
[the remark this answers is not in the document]
msub2: Yes, they are marked as such now.

components/script/dom/gamepad.rs
msub2: Done

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
         event.fire(target);
     }
 
+    pub fn dispatch_gamepad_event(&self, gamepad_event: script_traits::GamepadEvent) {
```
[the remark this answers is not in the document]
msub2: Done

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                            let window = document.window();
+
+                            // 2.1 Set gamepad.[[connected]] to false.
+                            // 2.2 Let document be gamepad's relevant global object's associated Document;
+                            //     otherwise null.
+                            // 2.3 If gamepad.[[exposed]] is true and document is not null and is fully active,
+                            //     then fire an event named gamepaddisconnected at gamepad's relevant global object
+                            //     using GamepadEvent with its gamepad attribute initialized to gamepad.
+                            // 2.4 Let navigator be gamepad's relevant global object's Navigator object.
+                            // 2.5 Set navigator.[[gamepads]][gamepad.index] to null.
```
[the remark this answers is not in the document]
msub2: Yes, marked them as such now

components/script/script_thread.rs
```diff
@@ -3669,6 +3678,14 @@ impl ScriptThread {
                 };
                 document.dispatch_composition_event(composition_event);
             },
+
+            GamepadEvent(gamepad_event) => {
+                let document = match self.documents.borrow().find_document(pipeline_id) {
```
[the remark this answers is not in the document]
msub2: Done

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
     }
 
+    pub fn dispatch_gamepad_event(&self, gamepad_event: script_traits::GamepadEvent) {
+        let global = self.window.upcast::<GlobalScope>();
+        match gamepad_event {
+            script_traits::GamepadEvent::Connected(index, name) => {
+                // TODO: 2. If document is not null and is not allowed to use the "gamepad" permission,
+                //          then abort these steps.
+                // 3. Queue a task on the gamepad task source to perform the following steps:
+                let document = Trusted::new(self);
```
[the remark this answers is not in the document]
msub2: Done

components/script/dom/gamepad.rs
msub2: Done

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                            let gamepad_list = window.Navigator().GetGamepads();
+                            if gamepad_list.Length() > 0 {
+                                gamepad_list.remove_gamepad(index.0);
+                            }
+                        }),
+                        &global,
+                    )
+                    .unwrap();
+            },
+            // https://www.w3.org/TR/gamepad/#receiving-inputs
```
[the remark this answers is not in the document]
msub2: Already fixed locally 👍

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                // 2. Queue a task on the gamepad task source to update gamepad state for gamepad.
+                self.window().task_manager().gamepad_task_source().queue(
+                    task!(gamepad_updated: move || {
+                        let document = document.root();
+                        let window = document.window();
+                        let gamepad_list = window.Navigator().GetGamepads();
+                        if let Some(gamepad) = gamepad_list.IndexedGetter(index.0 as u32) {
+                            // 1. Let now be the current high resolution time.
+                            let current_time = time::get_time();
+                            let now = (current_time.sec * 1000 + current_time.nsec as i64 / 1000000) as f64;
```
Taym95, line 1945: we are trying to remove `time` create, can you use std::time? 

cc @mrobinson

msub2: Yes, done.

components/script/dom/document.rs
```diff
@@ -1862,6 +1864,108 @@ impl Document {
...
+                            let window = document.window();
+                            let global = window.upcast::<GlobalScope>();
+
+                            // 3.1 Let gamepad be a new Gamepad representing the gamepad.
+                            let gamepad = Gamepad::new(&global, index.0 as u32, name);
+                            // 3.2 Let navigator be gamepad's relevant global object's Navigator object.
+                            // 3.3 Set navigator.[[gamepads]][gamepad.index] to gamepad.
+                            let gamepad_list = window.Navigator().GetGamepads();
+                            let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
+                            gamepad_list.add_if_not_exists(&gamepad_arr);
```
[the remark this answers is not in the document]
msub2: Move which part? I'm already calling a method on gamepad list to add the gamepad that was constructed earlier in the task

components/script/dom/gamepad.rs
msub2: When I tried this originally it complained about the slice type, the reason I used rooted_vec was from jdm pointing out I could get the DomSlice from it

ports/servoshell/app.rs
```diff
@@ -437,6 +437,7 @@ impl App {
         }
 
         // Catch some keyboard events, and push the rest onto the WebViewManager event queue.
+        webviews.handle_gamepad_events();
```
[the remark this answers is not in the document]
msub2: I've gone ahead and made this change

components/constellation/constellation.rs
```diff
@@ -5466,4 +5469,39 @@ where
...
+                    None => {
+                        return debug!("{}: Got key event after closure", pipeline_id);
+                    },
+                };
+                if let Err(e) = result {
+                    self.handle_send_error(pipeline_id, e);
+                }
+            },
+            None => {
+                warn!("No focused browsing context! TODO: handle this?");
```
gterzian, line 5503: Let's write "No focused webview to handle gamepad event".

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3104,109 @@ impl GlobalScope {
             .handle_server_msg(scope, result);
     }
 
+    pub fn handle_gamepad_event(&self, gamepad_event: script_traits::GamepadEvent) {
+        match gamepad_event {
+            script_traits::GamepadEvent::Connected(index, name, bounds) => {
```
gterzian, line 3109: Let's import `GamepadEvent`.

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3104,109 @@ impl GlobalScope {
...
+    pub fn handle_gamepad_disconnect(&self, index: usize) {
+        let this = Trusted::new(&*self);
+        self.gamepad_task_source()
+            .queue(
+                task!(gamepad_disconnected: move || {
+                    let global = this.root();
+                    if let Some(window) = global.downcast::<Window>() {
+                        let gamepad_list = window.Navigator().GetGamepads();
+                        if gamepad_list.Length() > 0 {
+                            gamepad_list.remove_gamepad(index);
```
gterzian, line 3169: let's do:

1. if let some gamepad = gamepad_list.Item
2. update_connected(false)
3. remove_gamepad(which just removes it)

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3104,109 @@ impl GlobalScope {
...
+                        }
+                    }
+                }),
+                &self,
+            )
+            .unwrap();
+    }
+
+    /// <https://www.w3.org/TR/gamepad/#receiving-inputs>
+    pub fn handle_gamepad_update(&self, index: usize, update_type: GamepadUpdateType) {
```
gterzian, line 3179: Let's call this receive_new_gamepad_buttons_or_axis.

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3104,109 @@ impl GlobalScope {
...
+        let this = Trusted::new(&*self);
+        self.gamepad_task_source()
+            .queue(
+                task!(gamepad_updated: move || {
```
gterzian, line 3183: Here we can add a comment with https://w3c.github.io/gamepad/#dfn-update-gamepad-state, and use `update_gamepad_state` as the task name.

ports/servoshell/webview.rs
```diff
@@ -113,6 +124,81 @@ where
         }
     }
 
+    pub fn handle_gamepad_events(&mut self) {
+        if let Some(ref mut gilrs) = self.gamepad {
+            while let Some(event) = gilrs.next_event() {
+                let gamepad = gilrs.gamepad(event.id);
+                let name = gamepad.name();
+                let index = GamepadIndex(event.id.into());
```
gterzian, line 132: `gamepad.into()`?

ports/servoshell/webview.rs
```diff
@@ -113,6 +124,81 @@ where
         }
     }
 
+    pub fn handle_gamepad_events(&mut self) {
```
gterzian, line 127: I think inside the method we can add various links to concepts in https://w3c.github.io/gamepad/#remapping

components/shared/script/lib.rs
```diff
@@ -1327,3 +1330,38 @@ impl SerializedImageData {
...
+    /// An existing gamepad has been disconnected
+    Disconnected(GamepadIndex),
+    /// An existing gamepad has been updated
+    Updated(GamepadIndex, GamepadUpdateType),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+/// The type of Gamepad input being updated
+pub enum GamepadUpdateType {
+    /// Axis index and input value
```
gterzian, line 1363: https://w3c.github.io/gamepad/#dfn-represents-a-standard-gamepad-axis

components/shared/script/lib.rs
```diff
@@ -1327,3 +1330,38 @@ impl SerializedImageData {
...
+    Axis(usize, f64),
+    /// Button index and input value
+    Button(usize, f64),
```
gterzian, line 1366: https://w3c.github.io/gamepad/#dfn-represents-a-standard-gamepad-button

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
 
+    pub fn update_timestamp(&self, timestamp: f64) {
+        self.timestamp.set(timestamp);
+    }
+
     pub fn notify_event(&self, event_type: GamepadEventType) {
         let event = GamepadEvent::new_with_type(&self.global(), event_type, &self);
         event
             .upcast::<Event>()
             .fire(self.global().as_window().upcast::<EventTarget>());
```
gterzian, line 204: I believe we also need to update according to https://w3c.github.io/gamepad/#extensions-to-the-windoweventhandlers-interface-mixin

See `dom/bindings/webidl/EventHandler.webidl`

components/shared/script/lib.rs
```diff
@@ -1327,3 +1330,38 @@ impl SerializedImageData {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+/// The type of Gamepad event
+pub enum GamepadEvent {
+    /// A new gamepad has been connected
+    Connected(GamepadIndex, String, GamepadInputBounds),
+    /// An existing gamepad has been disconnected
+    Disconnected(GamepadIndex),
+    /// An existing gamepad has been updated
+    Updated(GamepadIndex, GamepadUpdateType),
```
gterzian, line 1357: https://w3c.github.io/gamepad/#receiving-inputs

components/shared/script/lib.rs
```diff
@@ -1327,3 +1330,38 @@ impl SerializedImageData {
...
+    pub axis_bounds: (f64, f64),
+    /// Minimum and maximum button values
+    pub button_bounds: (f64, f64),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+/// The type of Gamepad event
+pub enum GamepadEvent {
+    /// A new gamepad has been connected
+    Connected(GamepadIndex, String, GamepadInputBounds),
```
gterzian, line 1353: https://w3c.github.io/gamepad/#event-gamepadconnected

components/shared/script/lib.rs
```diff
@@ -1327,3 +1330,38 @@ impl SerializedImageData {
...
+    /// An existing gamepad has been disconnected
+    Disconnected(GamepadIndex),
```
gterzian, line 1355: https://w3c.github.io/gamepad/#event-gamepaddisconnected

(It never hurst to add links, even if they are also added at the point of event-handling already)

components/script/task_source/mod.rs
```diff
@@ -38,6 +39,7 @@ pub enum TaskSourceName {
     MediaElement,
     Websocket,
     Timer,
+    Gamepad,
```
gterzian, line 42: https://w3c.github.io/gamepad/#dfn-gamepad-task-source

(even if it wasn't done for the others...)

components/script/script_thread.rs
```diff
@@ -3669,6 +3678,15 @@ impl ScriptThread {
                 };
                 document.dispatch_composition_event(composition_event);
             },
+
+            GamepadEvent(gamepad_event) => {
+                let window = match self.documents.borrow().find_window(pipeline_id) {
+                    None => return,
```
gterzian, line 3684: Let's do as the others `None => return warn!("Message sent to closed pipeline {}.", pipeline_id),`

components/script/dom/globalscope.rs
```diff
@@ -2517,6 +2521,15 @@ impl GlobalScope {
         unreachable!();
     }
 
+    /// `TaskSource` to send messages to the gamepad task source of
+    /// this global scope.
+    pub fn gamepad_task_source(&self) -> GamepadTaskSource {
```
gterzian, line 2526: https://w3c.github.io/gamepad/#dfn-gamepad-task-source

gterzian (changes requested):
Thanks for the changes, it is much easier to follow-along now. 

I have first a few more "structural" comments, and after they have been addressed I will review the details of the spec algorithms.

ports/servoshell/app.rs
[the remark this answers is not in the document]
wusyong: Hey sorry for the late reply. Yeah I think it's better to match all event variant into same method.

Btw I have some thoughts about some input events like keyboard and mouse should be more direct.
Right now it has to be passed from embedder to constellation and the to script thread.
In general, it should be fine. But for gaming usage, it usually requires more precision checks. Even a few nanoseconds lost may cause the trigger to miss the previous tick frame.

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3105,112 @@ impl GlobalScope {
...
+                let global = this.root();
+                let gamepad = Gamepad::new(&global, index as u32, name, axis_bounds, button_bounds);
+
+                // TODO: 3.4 If navigator.[[hasGamepadGesture]] is true:
+                // TODO: 3.4.1 Set gamepad.[[exposed]] to true.
+                if let Some(window) = global.downcast::<Window>() {
+                    let gamepad_list = window.Navigator().GetGamepads();
+                    let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
+                    gamepad_list.add_if_not_exists(&gamepad_arr);
+
```
gterzian, line 3149: The two TODOs belong rather here.

components/script/dom/gamepadlist.rs
```diff
@@ -46,6 +44,10 @@ impl GamepadList {
             }
         }
     }
+
+    pub fn remove_gamepad(&self, index: usize) {
+        self.list.borrow_mut().remove(index);
```
gterzian, line 49: Let's also  do step 6: "While navigator.[[[gamepads]]](https://www.w3.org/TR/gamepad/#dfn-gamepads) [is not empty](https://infra.spec.whatwg.org/#list-is-empty) and the last [item](https://infra.spec.whatwg.org/#list-item) of navigator.[[[gamepads]]](https://www.w3.org/TR/gamepad/#dfn-gamepads) is null, [remove](https://infra.spec.whatwg.org/#list-remove) the last [item](https://infra.spec.whatwg.org/#list-item) of navigator.[[[gamepads]]](https://www.w3.org/TR/gamepad/#dfn-gamepads)."

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3105,112 @@ impl GlobalScope {
...
+    pub fn handle_gamepad_disconnect(&self, index: usize) {
+        let this = Trusted::new(&*self);
+        self.gamepad_task_source()
+            .queue(
+                task!(gamepad_disconnected: move || {
+                    let global = this.root();
+                    if let Some(window) = global.downcast::<Window>() {
+                        let gamepad_list = window.Navigator().GetGamepads();
+                        if let Some(gamepad) = gamepad_list.Item(index as u32) {
+                            gamepad.update_connected(false);
```
gterzian, line 3170: TODO if gamepad exposed.

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3105,112 @@ impl GlobalScope {
...
+
+        // <https://w3c.github.io/gamepad/#dfn-update-gamepad-state>
+        self.gamepad_task_source()
+            .queue(
+                task!(update_gamepad_state: move || {
+                    let global = this.root();
+                    if let Some(window) = global.downcast::<Window>() {
+                        let gamepad_list = window.Navigator().GetGamepads();
+                        if let Some(gamepad) = gamepad_list.IndexedGetter(index as u32) {
+                            let current_time = SystemTime::now()
```
gterzian, line 3192: Let's use `global.performance().Now()`

components/script/dom/globalscope.rs
```diff
@@ -3091,6 +3105,112 @@ impl GlobalScope {
...
+                            gamepad.update_timestamp(current_time);
+
+                            match update_type {
+                                GamepadUpdateType::Axis(index, value) => {
+                                    gamepad.map_and_normalize_axes(index, value);
+                                },
+                                GamepadUpdateType::Button(index, value) => {
+                                    gamepad.map_and_normalize_buttons(index, value);
+                                }
+                            };
```
gterzian, line 3205: TODO for step 6?

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
+        ];
+        rooted_vec!(let buttons <- standard_buttons.iter().map(|button| DomRoot::from_ref(&**button)));
+        GamepadButtonList::new(global, buttons.r())
+    }
+
+    #[allow(unsafe_code)]
+    /// <https://www.w3.org/TR/gamepad/#dfn-map-and-normalize-axes>
+    pub fn map_and_normalize_axes(&self, axis_index: usize, value: f64) {
+        let normalized_value: f64 =
+            2.0 * (value - self.axis_bounds.0) / (self.axis_bounds.1 - self.axis_bounds.0) - 1.0;
```
gterzian, line 251: Is it possible to use the f64 methods that avoid panics?

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
     pub fn notify_event(&self, event_type: GamepadEventType) {
         let event = GamepadEvent::new_with_type(&self.global(), event_type, &self);
         event
             .upcast::<Event>()
             .fire(self.global().as_window().upcast::<EventTarget>());
     }
+
+    /// Initialize the number of axes in the "standard" gamepad mapping.
+    /// <https://www.w3.org/TR/gamepad/#dfn-initializing-axes>
+    pub fn init_axes(&self) {
```
gterzian, line 209: This can be private.

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
+        rooted_vec!(let buttons <- standard_buttons.iter().map(|button| DomRoot::from_ref(&**button)));
+        GamepadButtonList::new(global, buttons.r())
+    }
+
+    #[allow(unsafe_code)]
+    /// <https://www.w3.org/TR/gamepad/#dfn-map-and-normalize-axes>
+    pub fn map_and_normalize_axes(&self, axis_index: usize, value: f64) {
+        let normalized_value: f64 =
+            2.0 * (value - self.axis_bounds.0) / (self.axis_bounds.1 - self.axis_bounds.0) - 1.0;
+        let mut axis_vec = self.axes.get_internal().unwrap();
```
gterzian, line 252: let's use `internal_to_option` and warn that axes haven't been initiazed if none.

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
+    }
+
+    /// <https://www.w3.org/TR/gamepad/#dfn-map-and-normalize-buttons>
+    pub fn map_and_normalize_buttons(&self, button_index: usize, value: f64) {
+        let normalized_value: f64 =
+            (value - self.button_bounds.0) / (self.button_bounds.1 - self.button_bounds.0);
+        let pressed = normalized_value == 1.0;
+        let touched = normalized_value > 0.0;
+        if let Some(button) = self.buttons.IndexedGetter(button_index as u32) {
+            button.update(pressed, touched, normalized_value);
```
gterzian, line 265: TODO button pressed and touched.

components/script/dom/gamepad.rs
```diff
@@ -143,10 +193,76 @@ impl Gamepad {
...
+        let mut axis_vec = self.axes.get_internal().unwrap();
+        unsafe {
+            axis_vec.as_mut_slice()[axis_index] = normalized_value;
+        }
+    }
+
+    /// <https://www.w3.org/TR/gamepad/#dfn-map-and-normalize-buttons>
+    pub fn map_and_normalize_buttons(&self, button_index: usize, value: f64) {
+        let normalized_value: f64 =
+            (value - self.button_bounds.0) / (self.button_bounds.1 - self.button_bounds.0);
```
gterzian, line 261: same here re under- and overflow, can we avoid them?

gterzian (changes requested):
Thanks for the work, looks like very good progress. Couple of changes on the detailed algorithms. 

@wusyong Could you please take a look at the embedding layer? 

We also need to start looking at WPT updates, see https://github.com/servo/servo/tree/main/tests/wpt#updating-test-expectations

components/script/dom/gamepad.rs
gterzian: Ok!

components/script/dom/document.rs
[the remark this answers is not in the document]
gterzian: We could create the gamepad and add it to the list in `gamepad_list.add_if_not_exists` and return it, but it doesn't really matter.

msub2:
In addition to the WPT changes there's one notable remaining issue that still needs fixing, which is that button inputs don't seem to register on their very first input, but respond without issue afterwards. Going to try and track that down later today

gterzian:
> In addition to the WPT changes there's one notable remaining issue that still needs fixing, which is that button inputs don't seem to register on their very first input, but respond without issue afterwards. Going to try and track that down later today

Interesting. Is that after the connected event has been fired? I can imagine some input coming-in before that and being ignored.

msub2:
> Interesting. Is that after the connected event has been fired? I can imagine some input coming-in before that and being ignored.

It is, that's what confuses me about it. Axes values seem to update immediately so I'll need to track down where it seems to be getting lost

components/constellation/constellation.rs
```diff
@@ -5466,4 +5469,39 @@ where
             error!("Got a media session action but no active media session is registered");
         }
     }
+
+    fn handle_gamepad_msg(&mut self, event: GamepadEvent) {
```
wusyong, line 5473: Let's add doc comment to the method even it's evident in this case. People who read the code may want to know who is handling this message.

ports/servoshell/webview.rs
```diff
@@ -113,6 +124,85 @@ where
...
```
wusyong, line 127: ditto

msub2:
I've figured out the button issue! Turns out it stems from how I'm handling events from GilRS. There are ButtonPressed, ButtonReleased, and ButtonChanged events. I've been listening for ButtonChanged, but the *very* first time you press a button it only emits the ButtonPressed event, I assume because there's no prior state to identify a change with. Fixing this shouldn't be too difficult now, just need to add some checks on the embedder side

gterzian (approved):
Ok to script all LGTM, would be good to:

1. Update test expectations so that it can pass the merge queue. 
2. File an issue re TODO'(one issue can just give a sense of what needs to be addressed later). 

The frozen array stuff can be done in a follow-up I would say(unless it makes new tests fails, but I don't think it does). 

@wusyong All ok on the embedding side?

@sagudev I will be away next week, could you help to get this to merge?

sagudev:
> @sagudev Wu Yu Wei mentioned on Zulip already that it looks good

I see.

sagudev:
Some more passing test on legacy-layout:
```
   OK /gamepad/idlharness.https.window.html
        PASS [expected FAIL] subtest: HTMLBodyElement interface: attribute ongamepadconnected
        PASS [expected FAIL] subtest: HTMLBodyElement interface: attribute ongamepaddisconnected
        PASS [expected FAIL] subtest: Window interface: attribute ongamepadconnected
        PASS [expected FAIL] subtest: Window interface: attribute ongamepaddisconnected
        PASS [expected FAIL] subtest: HTMLFrameSetElement interface: attribute ongamepadconnected
        PASS [expected FAIL] subtest: HTMLFrameSetElement interface: attribute ongamepaddisconnected
```

sagudev:
I think this is wrong:

gterzian:
@msub2 thanks for the contrib!

@sagudev @wusyong thanks for your help!

PR #18161, opened 2017-08-20
Add support for legacy table attribute cellpadding

adamncasey:
Handle the parent table element's `cellpadding` attribute value during table cell layout.

When `synthesize_presentational_hints_for_legacy_attributes` is called with an `HTMLTableCellElement`, look along a fixed path for the parent table element using `LayoutJS<Node>::parent_node_ref()`.

As touched on in #12609, another implementation worth evaluating would be for the `HTMLTableElementMethods::InsertRow` method to pass down the table's cellpadding value, and then for the `HTMLTableRowElementMethods::InsertCell` to pass it to each Cell. This would remove the need for the parent table lookup for every cell during layout, at the expense of extra memory per cell. This value needs to be mutable at the table level, as the property can be updated. Let me know if this method seems more appropriate. I haven't gone down this route so far because I wasn't clear on which of Rust's reference wrappers would be best suited for this.

Hacker News rendering improvement:
cellpadding

There are a few issues running the tests on Windows right now - no headless mode implemented, and seems a lot of the test expectations aren't right. I'm relying on the CI here.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #12609

- [X] There are tests for these changes

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/18161)

PR #31201, opened 2024-01-27
Add support for cellpadding attribute

### LINKED ISSUE #12609: Table cellpadding attribute
Although fairly legacy, it probably makes sense for servo to support the `cellpadding` attribute of `<table>`.

```
<html>
<head>
    <title>Table cellpadding attribute + padding CSS</title>
    <style type="text/css">
        table { background-color: red; }
        div { width:10px;height:10px; }
    </style>
</head>
<body>
    <table>
        <tr>
            <td style="padding:5px"><div></div></td>
        </tr>
    </table>
    <table cellpadding="5">
        <tr>
            <td><div></div></td>
        </tr>
    </table>
</body>
</html>
```

The above example renders as a single column of red in Firefox, Chrome & Edge.

https://html.spec.whatwg.org/multipage/rendering.html#tables-2 describes how cellpadding should be interpreted.

Related to #11821 in the sense that this prevents HN's top bar from being rendered correctly.

I'm hoping to give this a go by following how `cellspacing` was implemented in https://github.com/servo/servo/pull/4417 . Some of that has been moved around quite a bit since though.

Loirooriol:
---

- [X] These changes fix #12609

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -989,6 +989,32 @@ impl<'dom> LayoutElementHelpers<'dom> for LayoutDom<'dom, Element> {
                 PropertyDeclaration::BorderRightWidth(width_value),
             ));
         }
+
+        if let Some(cellpadding) = self
+            .downcast::<HTMLTableCellElement>()
+            .and_then(|this| this.get_table())
+            .and_then(|table| table.get_cellpadding())
+        {
+            let cellpadding = NonNegative(specified::LengthPercentage::Length(
+                specified::NoCalcLength::from_px(cellpadding as f32),
+            ));
+            hints.push(from_declaration(
+                shared_lock,
+                PropertyDeclaration::PaddingTop(cellpadding.clone()),
+            ));
+            hints.push(from_declaration(
+                shared_lock,
+                PropertyDeclaration::PaddingLeft(cellpadding.clone()),
+            ));
+            hints.push(from_declaration(
+                shared_lock,
+                PropertyDeclaration::PaddingBottom(cellpadding.clone()),
+            ));
+            hints.push(from_declaration(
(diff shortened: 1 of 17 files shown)


mrobinson (approved):

mrobinson:
@Loirooriol it might also be worth looking at the comments on an earlier iteration of this change: https://github.com/servo/servo/pull/18161

Loirooriol:
Looking at that would have saved me some time :facepalm:

Some notes:
 - My patch doesn't handle dynamic changes, but it's the same for other attributes like `cellspacing` and `bgcolor`, so I was thinking of doing that in a follow-up.
 - Regarding passing the value down in `HTMLTableElementMethods::InsertRow` and `HTMLTableRowElementMethods::InsertCell`, I think these are only invoked from scripts, not from the HTML parser. And passing it down during layout wouldn't follow the HTML spec, since AFAIK we are supposed to use the semantic table instead of the element that establishes the table formatting context.
 - Regarding the actual algorithm for finding the table:
   - Gecko uses the grandparent if it's a `<table>`, or the great-grandparent if it's a `<table>`. I think this doesn't follow the HTML spec because it doesn't check that the parent is a `<tr>`, even though https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-processing-rows is only invoked for `<tr>`.
   - WebKit uses the nearest `<table>` ancestor, which doesn't match the spec.
   - Blink is like WebKit, but iterating in the flat tree
   - So or Servo I did it like Gecko but checking the intermediate ancestors. And the iteration is in the flat tree, because when I tried `parent_node_ref` the compiler said it's private, and it's marked as unsafe code, so I thought that `composed_parent_node_ref` wouldn't be a big deal since Blink uses the flat tree too.

PR #31209, opened 2024-01-29
libservo: Handle GL video decoding setup internally

mrobinson:
Instead of making the client set up GL video decoding, have this done
inside libservo. The details necessary for decoding are fetched via
Surfman now. This also removes the setup for the context on Android --
which has no GStreamer support now anyway. In the future, this could be
enabled, but should likely be done using Surfman, instead of passing on
all these details manually.

---

### DIFF
--- a/ports/servoshell/app.rs
+++ b/ports/servoshell/app.rs
@@ -14,7 +14,7 @@ use gleam::gl;
 use log::{info, trace, warn};
 use servo::compositing::windowing::EmbedderEvent;
 use servo::compositing::CompositeTarget;
-use servo::config::opts;
+use servo::config::{opts, set_pref};
 use servo::servo_config::pref;
 use servo::Servo;
 use surfman::GLApi;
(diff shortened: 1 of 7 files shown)


wusyong:
I really love this change which simplifies the WindowMethod trait!
I want to take this chance to document media prefs as well. But I don't fully understand their usage.
There are currently two configs in prefs for media:
- media.glvideo.enable
- media.testing.enable

Do you know what exactly they are used for?

jdm:
If media.glvideo.enabled is true, the media code will attempt to use hardware acceleration for video playback: https://github.com/servo/servo/blob/9b6c473695e14c1a37dd70325657519b901e7efc/ports/servoshell/headed_window.rs#L589

ports/servoshell/app.rs
```diff
@@ -67,6 +67,8 @@ impl App {
 
         // Implements window methods, used by compositor.
         let window = if opts::get().headless {
+            // GL video rendering is not supported on headless windows.
+            set_pref!(media.glvideo.enabled, false);
```
mukilan, line 71: Does this mean currently we have GL video rendering enabled for headless windows?

mukilan (approved):

ports/servoshell/app.rs
mrobinson: Nope, this is disabled, but only implicitly because the previous `WindowMethod` trait method implementations for headless windows did not return information useful for GL rendering.

mrobinson:
Thanks for the review!

PR #31213, opened 2024-01-29
Remove deprecated 'remove' function

### LINKED ISSUE #31211: Warning at script_module.rs:283
When compiling, I get

```
warning: use of deprecated method `indexmap::IndexSet::<T, S>::remove`: `remove` disrupts the set order -- use `swap_remove` or `shift_remove` for explicit behavior.
   --> components/script/script_module.rs:283:49
    |
283 |         self.incomplete_fetch_urls.borrow_mut().remove(&dependency);
    |                                                 ^^^^^^
    |
note: the lint level is defined here
   --> components/script/lib.rs:62:8
    |
62  | #[warn(deprecated)]
    |        ^^^^^^^^^^
```

Taym95:
---

- [x] These changes fix #31211

### DIFF
--- a/components/script/script_module.rs
+++ b/components/script/script_module.rs
@@ -280,7 +280,9 @@ impl ModuleTree {
     }
 
     pub fn remove_incomplete_fetch_url(&self, dependency: ServoUrl) {
-        self.incomplete_fetch_urls.borrow_mut().remove(&dependency);
+        self.incomplete_fetch_urls
+            .borrow_mut()
+            .shift_remove(&dependency);
     }
 
     /// recursively checks if all of the transitive descendants are


components/script/script_module.rs
```diff
@@ -280,7 +280,7 @@ impl ModuleTree {
     }
 
     pub fn remove_incomplete_fetch_url(&self, dependency: ServoUrl) {
-        self.incomplete_fetch_urls.borrow_mut().remove(&dependency);
+        self.incomplete_fetch_urls.borrow_mut().swap_remove(&dependency);
```
Taym95, line 283: from doc:
/// This is equivalent to [`.swap_remove(value)`][Self::swap_remove], replacing this
/// value's position with the last element, and it is deprecated in favor of calling that
/// explicitly.

CYBAI: We're not relying on the order of `incomplete_fetch_urls` though, as we're using `IndexSet`, I wonder maybe it's better to move to `shift_remove` (in case that folks may expect the order is preserved)? 

```suggestion
        self.incomplete_fetch_urls.borrow_mut().shift_remove(&dependency);
```

CYBAI: Thank you for the update! As the first wpt try is triggered by previous commit, let me trigger them again.

CYBAI (approved):
WPT 2020 pass and WPT 2013 failed with an intermittent network error from artifact upload. Thank you for the PR! :shipit:

PR #31219, opened 2024-01-29
Allow using cached client_rect() for paint-only reflow

### LINKED ISSUE #31195: clientWidth may not force layout when style has changed
```html
<!DOCTYPE html>
<div id="target"></div>
<div id="log" style="white-space: pre"></div>
<script>
target.style.cssText = "width: 5px";
log.append(target.clientWidth + "\n");

target.style.cssText = "width: 15px";
log.append(target.clientWidth + "\n");
</script>
```

Expected:
```
5
15
```

Actual:
```
5
5
```

Loirooriol:
PR #31210 avoided the cache for all kinds of reflow, but it's actually fine to use it for ReflowTriggerCondition::PaintPostponed.

---

- [X] These changes do not require tests because there is no change in behavior, just better perf

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -94,7 +94,9 @@ use crate::dom::create::create_element;
 use crate::dom::customelementregistry::{
     CallbackReaction, CustomElementDefinition, CustomElementReaction, CustomElementState,
 };
-use crate::dom::document::{determine_policy_for_token, Document, LayoutDocumentHelpers};
+use crate::dom::document::{
+    determine_policy_for_token, Document, LayoutDocumentHelpers, ReflowTriggerCondition,
+};
 use crate::dom::documentfragment::DocumentFragment;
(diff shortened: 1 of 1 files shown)


components/script/dom/element.rs
```diff
@@ -3380,7 +3382,9 @@ impl Element {
             .and_then(|data| data.client_rect.as_ref())
             .and_then(|rect| rect.get().ok())
         {
-            if doc.needs_reflow().is_none() {
+            if doc.needs_reflow().map_or(true, |condition| {
+                condition == ReflowTriggerCondition::PaintPostponed
+            }) {
```
mrobinson, line 3387: It might be a bit more readable to use `matches!` here.

```suggestion
            if matches!(doc.needs_reflow(), None | Some(ReflowTriggerCondition::PaintPostponed)) {
```

mrobinson (approved):

PR #31221, opened 2024-01-29
Lint layout_2013 with clippy

Loirooriol:
This was done with:
1. `CARGO_BUILD_RUSTC=rustc cargo clippy --fix -p layout_2013 --broken-code`
2. `./mach fmt`
3. Some manual cosmetic adjustments

---

- [X] These changes do not require tests because it's just linting


mrobinson (approved):
I was only able to do a quick review, as this was quite a large change. It all feels pretty mechanical though, so I think if tests pass we should land it.

PR #31226, opened 2024-01-30
WebIDL: Use `Uint8ClampedArray` instead of raw `JSObject` in bindings

Taym95:
part #31064

---


components/script/dom/imagedata.rs
```diff
@@ -131,7 +134,7 @@ impl ImageData {
         width: u32,
         height: u32,
     ) -> Fallible<DomRoot<Self>> {
-        unsafe { Self::new_without_jsobject(global, proto, width, height) }
+        unsafe { Self::new_without_data(global, proto, width, height) }
```
Taym95, line 137: @gterzian can you review this when you got time, still in progress

components/script/dom/imagedata.rs
```diff
@@ -111,15 +111,18 @@ impl ImageData {
...
         });
 
         let len = width * height * 4;
         let cx = GlobalScope::get_cx();
+
         rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
-        Uint8ClampedArray::create(*cx, CreateWith::Length(len as usize), array.handle_mut())
-            .unwrap();
-        (*imagedata).data.set(array.get());
+        let typed_array =
```
gterzian, line 121: I think this belongs in a new `new_initialized` method of the heaped array, and you should create it above before `imagedata`.

gterzian: And you can make `new_initialized` take an optional `JSObject`, so that it can be used with `new_with_js_object`(see below).

components/script/dom/imagedata.rs
```diff
@@ -144,23 +147,23 @@ impl ImageData {
         width: u32,
         opt_height: Option<u32>,
     ) -> Fallible<DomRoot<Self>> {
-        Self::new_with_jsobject(global, proto, width, opt_height, jsobject)
+        // Self::new_with_data(global, proto, width, opt_height, jsobject)
+        Self::new_without_data(global, proto, width, opt_height.unwrap())
```
gterzian, line 151: Here you should keep `new_with_jsobject`, and create a new `new_initialized` method on `HeapTypedArray` that takes an optional jsobject, in this case some.

gterzian (changes requested):
Thanks! This one is another rather complicated one, so I suggest a few changes.

PR #31230, opened 2024-01-30
layout: Collect both start and end baselines for fragments

mrobinson:
This change starts collecting the starting baseline set for fragments,
which is necessary for some layout modes (flex and tables, namely) as
well as being important for the implementation of `align-items`. In
addition, it converts baseline measurement to use `Au` everywhere.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -34,7 +34,7 @@ use crate::cell::ArcRefCell;
 use crate::context::LayoutContext;
 use crate::flow::float::{FloatBox, SequentialLayoutState};
 use crate::flow::FlowLayout;
-use crate::formatting_contexts::IndependentFormattingContext;
+use crate::formatting_contexts::{Baselines, IndependentFormattingContext};
 use crate::fragment_tree::{
     AnonymousFragment, BaseFragmentInfo, BoxFragment, CollapsedBlockMargins, CollapsedMargin,
     Fragment,
@@ -246,9 +246,9 @@ impl LineBlockSizes {
         let height_from_ascent_and_descent = self
             .baseline_relative_size_for_line_height
             .as_ref()
-            .map(|size| Length::from((size.ascent + size.descent).abs()))
-            .unwrap_or_else(Length::zero);
-        self.line_height.max(height_from_ascent_and_descent)
+            .map(|size| (size.ascent + size.descent).abs())
+            .unwrap_or_else(Au::zero);
+        self.line_height.max(height_from_ascent_and_descent.into())
     }
 
     fn max(&self, other: &LineBlockSizes) -> LineBlockSizes {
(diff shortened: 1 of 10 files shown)


mrobinson:
@Loirooriol I think this is ready to review. I've also added you as a co-author. :pray:

components/layout_2020/flow/mod.rs
```diff
@@ -1416,9 +1413,14 @@ impl PlacementState {
 
         match fragment {
             Fragment::Box(box_fragment) => {
```
Loirooriol, line 1415: While at it, this could be `if let Fragment::Box(box_fragment) = fragment {`

components/layout_2020/flow/inline.rs
```diff
@@ -760,7 +759,12 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
             return;
         }
 
-        self.last_baseline_offset = Some(baseline_offset + block_start_position.into());
+        let baseline = baseline_offset + block_start_position;
+        if self.baselines.first.is_none() {
+            self.baselines.first = Some(baseline);
+        }
```
Loirooriol, line 765: Could be shorter with `self.baselines.first.get_or_insert(baseline);`, but either is fine.

components/layout_2020/flow/mod.rs
```diff
@@ -1414,14 +1411,15 @@ impl PlacementState {
...
-            _ => {},
+        if let Fragment::Box(box_fragment) = fragment {
+            let box_block_offset = box_fragment.content_rect.start_corner.block.into();
+            match (self.inflow_baselines.first, box_fragment.baselines.first) {
+                (None, Some(first)) => self.inflow_baselines.first = Some(first + box_block_offset),
+                _ => {},
+            }
+            if let Some(last) = box_fragment.baselines.last {
+                self.inflow_baselines.last = Some(last + box_block_offset);
+            }
```
Loirooriol, line 1422: Possibly clearer like this?

```suggestion
            let Baselines { first, last } = box_fragment.baselines;
            if self.inflow_baselines.first.is_none() {
                self.inflow_baselines.first = first.map(|first| first + box_block_offset);
            }
            self.inflow_baselines.last = last.map(|last| last + box_block_offset);
```

components/layout_2020/fragment_tree/box_fragment.rs
```diff
@@ -44,10 +45,10 @@ pub(crate) struct BoxFragment {
     /// <https://drafts.csswg.org/css2/#clearance>
     pub clearance: Option<Length>,
 
-    /// When this box contains an inline formatting context, this tracks the baseline
-    /// offset of the last inflow line. This offset is used to propagate baselines to
-    /// ancestors of `display: inline-block` ancestors.
-    pub last_baseline_offset: Option<Length>,
+    /// When this [`BoxFragment`] is for content that has a baseline, this tracks
+    /// the first and last baselines of that content. This is used to propogate baselines
```
Loirooriol, line 49:
```suggestion
    /// the first and last baselines of that content. This is used to propagate baselines
```

components/layout_2020/fragment_tree/box_fragment.rs
```diff
@@ -126,10 +124,14 @@ impl BoxFragment {
...
+        // This applies even if there is no baseline set.
+        let mut baselines = Baselines::default();
         if style.establishes_scroll_container() {
-            last_inflow_baseline_offset = Some(
-                content_rect.size.block + padding.block_end + border.block_end + margin.block_end,
-            );
+            baselines.last = Some(
+                (content_rect.size.block + padding.block_end + border.block_end + margin.block_end)
+                    .into(),
+            )
```
Loirooriol, line 134: I think it would be better to move this into `with_baselines()`.

Loirooriol (approved):

components/layout_2020/flow/mod.rs
mrobinson: Hrm. The problem is that we do not want to set last to `None` if last is already set.

components/layout_2020/fragment_tree/box_fragment.rs
mrobinson: We aren't sure that `with_baselines()` will be called and we need to unconditionally set the baseline to the end position. I could move this to `with_baselines()` and call that always in the constructor, but it's a bit odd I think.

components/layout_2020/fragment_tree/box_fragment.rs
mrobinson: Fixed!

components/layout_2020/flow/inline.rs
mrobinson: Sure. I've made this change.

components/layout_2020/fragment_tree/box_fragment.rs
mrobinson: I've improved this comment so it's a bit clearer why it's happening here instead of in `with_baselines()`.

PR #31223, opened 2024-01-30
Document media configs from prefs

wusyong:
Tried to add the missing doc comments for media prefs as mentioned in [zulip discussion](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Running.20servo.20with.20Audio/near/418574223).

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because these are doc comments.

PR #31231, opened 2024-01-31
Make Android build optional on Nix

syvb:
I want to build Servo on NixOS, without installing the non-free Android SDK. This PR makes it so Android support is only built if the `buildAndroid` argument is true (e.g. by running `nix-shell --arg buildAndroid true etc/shell.nix`)

I've verified that you can still build/run non-Android Servo with this change, but I haven't tested on Android (it should still work though, since the derivation is unaffected by this change when `buildAndroid` is true).

---

- [ ] `./mach test-tidy` does not report any errors - it reports a diff in `components/config/prefs.rs`, which this PR didn't change (I think #31223 caused this by adding a trailing space?)

- [X] These changes do not require tests because the Nix configuration doesn't have tests

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -537,6 +537,9 @@ class CommandBase(object):
             if os.path.isdir(default):
                 env.setdefault(f"ANDROID_{kind.upper()}_ROOT", default)
 
+        if "IN_NIX_SHELL" in env and ("ANDROID_NDK_ROOT" not in env or "ANDROID_SDK_ROOT" not in env):
+            print("Please set SERVO_ANDROID_BUILD=1 when starting the Nix shell to include the Android SDK/NDK.")
+            sys.exit(1)
         if "ANDROID_NDK_ROOT" not in env:
             print("Please set the ANDROID_NDK_ROOT environment variable.")
             sys.exit(1)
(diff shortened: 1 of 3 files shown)


mukilan:
Hi @syvb, thank you for the contribution. This PR looks good, but it will break `./mach build --android` when run outside the `nix-shell`. We have logic to [enter](https://github.com/servo/servo/blob/b2f73723f881889650304107676c110efb5ec5cd/mach#L42) the nix-shell env for all mach commands on Nix, so ideally we should pass `buildAndroid` true there. However, this is handled very early on in the mach where we haven't parsed all the arguments. I was thinking about having an environment variable like `SERVO_ANDROID_BUILD` that we check in that code to decide if `buildAndroid` needs to be set. What do you think?

python/servo/command_base.py
```diff
@@ -537,6 +537,9 @@ def build_android_env_if_needed(self, env: Dict[str, str]):
             if os.path.isdir(default):
                 env.setdefault(f"ANDROID_{kind.upper()}_ROOT", default)
 
+        if "IN_NIX_SHELL" and ("ANDROID_NDK_ROOT" not in env or "ANDROID_SDK_ROOT" not in env):
```
mukilan, line 540: Shouldn't this be `if "IN_NIX_SHELL" in env ` ?

syvb: Yes; I fixed it.

mukilan (approved):
Thank you!

PR #31234, opened 2024-01-31
Add support for mac arm (m1) runners

sagudev:
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

---

### DIFF
--- a/.github/workflows/dispatch-workflow.yml
+++ b/.github/workflows/dispatch-workflow.yml
@@ -39,6 +39,17 @@ jobs:
       unit-tests: ${{ inputs.unit-tests }}
       wpt-tests-to-run: ${{ inputs.wpt-tests-to-run }}
 
+  macos-arm:
+    if: ${{ inputs.workflow == 'macos-arm' }}
+    uses: ./.github/workflows/mac.yml
+    secrets: inherit
+    with:
+      arm: true
+      profile: ${{ inputs.profile }}
+      wpt-layout: ${{ inputs.wpt-layout }}
+      unit-tests: ${{ inputs.unit-tests }}
+      wpt-tests-to-run: ${{ inputs.wpt-tests-to-run }}
+
   linux:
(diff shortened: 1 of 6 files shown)


sagudev:
It would be nice if any mac owner could verify that builds work: https://github.com/sagudev/servo/actions/runs/7721819161

mrobinson:
Question about this: what kind of binary is this building? Isn't it more interesting to build a universal binary for now?

sagudev:
Currently separate binary per arch, but I expect that we can just do post step that would just join two binaries (no idea if this is even possible, i am not really familiar with macs). As far as I know universal needs to be compiled twice (every time for separate target) and doing natively rather than cross should be better.

mrobinson:
I'm fairly sure that the binaries we produce right now run just fine on ARM Macs. While a native binary might be nice, is it worth maintaining an entirely new build and then having a complicated step to merge the two binaries? I'm not sure I see the issue with cross-compilation either. What's the concern?

sagudev:
AFAIK we currently build in CI is for x86 arch and works only due to Rosetta (x86 emulation) on arm macs.

mrobinson:
It feels like the right thing to do is not to have two builds, but just have proper support for building "fat binaries" on MacOS.

sagudev:
Even if we build "fat binaries" we would still need to compile twice, so having it seperate will allow as to parallelize this better.

Also joining is indeed possible: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

kklem0:
> It would be nice if any mac owner could verify that builds work: https://github.com/sagudev/servo/actions/runs/7721819161

It runs well.
Closing the browser with the close icon on top left freezes the browser, closing through menu works though and I don't think it's the build issue.

sagudev:
Resurrected in https://github.com/servo/servo/pull/38460

PR #31240, opened 2024-01-31
ci: Upgrade some GitHub actions to NodeJS 20 actions

mrobinson:
See
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

Using version 4 of all of these actions should remove many warnings from
the GitHub CI console. `upload-artifact` is not upgraded yet because we
depend on a now-removed behavior of modifying existing artifacts.

---

- [x] These changes do not require tests because these are CI changes.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -34,18 +34,18 @@ jobs:
       matrix:
         chunk_id: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: github.event_name != 'pull_request_target'
         with:
           fetch-depth: 2
       # This is necessary to checkout the pull request if this run was triggered via a
       # `pull_request_target` event.
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: github.event_name == 'pull_request_target'
         with:
           ref: refs/pull/${{ github.event.number }}/head
           fetch-depth: 2
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
           name: release-binary
           path: release-binary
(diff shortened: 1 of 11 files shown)


.github/workflows/linux-wpt.yml
```diff
@@ -82,22 +82,22 @@ jobs:
...
+        uses: actions/upload-artifact@v4
         if: ${{ always() && !inputs.wpt-sync-from-upstream }}
         with:
           name: wpt-filtered-results-linux-${{ inputs.wpt-layout }}
           path: |
             filtered-test-wpt.${{ matrix.chunk_id }}.json
             unexpected-test-wpt.${{ matrix.chunk_id }}.log
       - name: Archive logs
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
mukilan, line 93: It seems upload-artifact v4  has a [breaking change](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact) in case of many jobs uploading to same artifact name. Is this something that affects this workflow, since we upload the chunked logs to same artifact?

sagudev: My thought exactly, but I have plan for much needed artifact refactor, currently if you run two linux jobs with different profile they overwrite packages (and wrong ones are then used in WPT runs). So I think this PR should only update all checkout actions, then I can handle artifacts refactor in separate PR.

mrobinson: Oh, dang. Nice catch! I have removed the upgrade of `upload-artifact`.

It seems it will be tricky to fix this issue as I'm unsure if there is a way to avoid uploading 40 artifacts for a full build, which seems like too many.

.github/workflows/mac-wpt.yml
```diff
@@ -30,18 +30,18 @@ jobs:
...
           fetch-depth: 2
       # This is necessary to checkout the pull request if this run was triggered via a
       # `pull_request_target` event.
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         if: github.event_name == 'pull_request_target'
         with:
           ref: refs/pull/${{ github.event.number }}/head
           fetch-depth: 2
-      - uses: actions/download-artifact@v3
```
sagudev, line 44: mac workflow still uses upload-artifact@v3 and you are not allowed to mix v3/v4 for upload download as they are incompatible.

mrobinson:
Okay. I think I'll close this for the moment until the mess with artifacts can be worked out.

PR #31241, opened 2024-02-01
Make console methods take `any` instead of `string`

### LINKED ISSUE #29460: console.log() throws TypeError when passed a symbol
```
$ ./mach run -d 'data:text/html,<script>console.log(Symbol())</script>'
[2023-03-03T12:20:46Z ERROR script::dom::bindings::error] Error at data:text/html,<script>console.log(Symbol())</script>:1:9 can't convert symbol to string
```

testharness.js [currently relies on being able to print symbols](https://github.com/servo/servo/blob/18015e69e4f7f155ac40a1729b4913c9ced75df9/tests/wpt/web-platform-tests/resources/testharness.js#L1431), and this issue means that when an assertion failure involves a symbol, the console TypeError obscures the real error. It’s unclear which part of the [console spec](https://console.spec.whatwg.org) this violates, but intuitively I would agree that the console printing methods should never fail.

syvb:
Match the [Console spec](https://console.spec.whatwg.org/) by having the console methods take `any` instead of a string. This fixes #29460, where `console.log(Symbol())` throws because Symbols can't be implicitly converted to a string.

I used `JS_ValueToSource` to convert all JS values to strings. This change results in better logging for objects (`({a:{b:{c:5, d:new Number(3), e:new Boolean(false), g:4324324423432423423423n, h:3.333, i:(void 0)}}})` instead of `[object Object]`).

---

- [X] These changes fix #29460


components/script/dom/console.rs
```diff
@@ -55,14 +58,39 @@ where
...
 }
 
-fn console_messages(global: &GlobalScope, messages: &[DOMString], level: LogLevel) {
-    console_message(global, DOMString::from(messages.join(" ")), level)
+#[allow(unsafe_code)]
+fn stringify_message(message: HandleValue) -> DOMString {
+    let cx = *GlobalScope::get_cx();
+    unsafe {
+        let js_string = if message.is_string() {
+            // directly print string
```
mrobinson, line 66: I don't think this comment is necessary:

```suggestion
```

components/script/dom/console.rs
```diff
@@ -55,14 +58,39 @@ where
...
+            message.to_string()
+        } else if message.is_undefined() {
+            // better value than "(void 0)" from JS_ValueToSource
```
mrobinson, line 69:
```suggestion
            // This produces a better value than "(void 0)" from JS_ValueToSource.
```

components/script/dom/console.rs
```diff
@@ -55,14 +58,39 @@ where
...
+            return DOMString::from("undefined");
+        } else {
+            jsapi::JS_ValueToSource(cx, message.into())
+        };
```
mrobinson, line 73: The values produced are a bit odd (for instance the way booleans) are represented. Could this do `JSON.stringify()` instead?

mrobinson (changes requested):
Nice work! I'm not sure about the new `Console` test. Someone with more experience there will need to review that. I have a few other comments though:

components/script/dom/console.rs
syvb: I moved some of the logic for producing the values into Servo, which improved the output in a few cases (e.g. for `undefined` in an object).

Note that Booleans are represented as `true`/`false` with `JS_ValueToSource` - the `new Boolean(...)` display is only used for the [Boolean primitive wrapper objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean#return_value), which are essentially never used.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                    return DOMString::from("...");
+                }
+                let value_bits = value.asBits_;
+                if parents.contains(&value_bits) {
+                    return DOMString::from("[circular]");
+                }
+                parents.push(value_bits);
+                if value.is_undefined() {
+                    // This produces a better value than "(void 0)" from JS_ValueToSource.
+                    DOMString::from("undefined")
```
mrobinson, line 98: Please use an early return.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                mut parents: Vec<u64>,
+            ) -> DOMString {
+                if parents.len() >= MAX_LOG_DEPTH {
+                    return DOMString::from("...");
+                }
+                let value_bits = value.asBits_;
+                if parents.contains(&value_bits) {
+                    return DOMString::from("[circular]");
+                }
+                parents.push(value_bits);
```
mrobinson, line 95: If this is `undefined` does it still make sense to push the value bits onto the parents?

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+fn stringify_message(message: HandleValue) -> DOMString {
+    let cx = *GlobalScope::get_cx();
+    unsafe {
+        if message.is_string() {
+            jsstring_to_str(cx, message.to_string())
+        } else {
+            unsafe fn stringify_inner(
+                cx: *mut jsapi::JSContext,
+                value: HandleValue,
+                mut parents: Vec<u64>,
```
mrobinson, line 86: A pushing and popping approach with a mutable vector would probably be a bit more efficient.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                            props.push(format!("{}: {}", key, value_string,));
+                        } else {
+                            props.push(value_string.to_string());
+                        }
+                    }
+                    if object_class == ESClass::Array {
+                        DOMString::from(format!("[{}]", itertools::join(props, ", ")))
+                    } else {
+                        DOMString::from(format!("{{{}}}", itertools::join(props, ", ")))
+                    }
```
mrobinson, line 173: This complex enough that perhaps there should be a `stringify_object_from_handle_value` function.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                        if explicit_keys {
+                            let key = if id.is_string() || id.is_symbol() || id.is_int() {
+                                rooted!(in(cx) let mut key_value = UndefinedValue());
+                                let raw_id: jsapi::HandleId = id.handle().into();
+                                if !JS_IdToValue(cx, *raw_id.ptr, key_value.handle_mut()) {
+                                    return DOMString::from("/* invalid */");
+                                }
+                                value_to_source(cx, key_value.handle())
+                            } else {
+                                panic!("Key with impossible type")
```
mrobinson, line 162: Instead of panicking, please just return an error string like you do with "/* invalid */" I don't think we want to stop the entire script thread for this.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                        if !JS_GetPropertyById(cx, obj.handle(), id.handle(), property.handle_mut())
+                        {
+                            return DOMString::from("/* invalid */");
+                        }
+
+                        if !explicit_keys {
+                            if id.is_int() {
+                                if let Ok(id_int) = usize::try_from(id.to_int()) {
+                                    explicit_keys = props.len() != id_int;
+                                } else {
```
mrobinson, line 145: Can you just do `to_int()` here without first checking whether this is an integer?

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+
+#[allow(unsafe_code)]
+unsafe fn value_to_source(cx: *mut jsapi::JSContext, value: HandleValue) -> DOMString {
+    rooted!(in(cx) let mut js_string = std::ptr::null_mut::<jsapi::JSString>());
+    js_string.set(JS_ValueToSource(cx, value));
+    return jsstring_to_str(cx, *js_string);
+}
+
+#[allow(unsafe_code)]
+fn stringify_message(message: HandleValue) -> DOMString {
```
mrobinson, line 77: How about `stringify_handle_value`.

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
     f()
 }
 
-fn console_messages(global: &GlobalScope, messages: &[DOMString], level: LogLevel) {
-    console_message(global, DOMString::from(messages.join(" ")), level)
+const MAX_LOG_DEPTH: usize = 10;
+
+#[allow(unsafe_code)]
+unsafe fn value_to_source(cx: *mut jsapi::JSContext, value: HandleValue) -> DOMString {
```
mrobinson, line 70: `handle_value_to_string` ?

components/script/dom/console.rs
```diff
@@ -55,14 +64,138 @@ where
...
+                } else {
+                    return value_to_source(cx, value);
+                }
+            }
+            stringify_inner(cx, message.into(), Vec::new())
+        }
+    }
+}
+
+fn stringify_messages(messages: Vec<HandleValue>) -> DOMString {
```
mrobinson, line 183: `stringify_handle_values`?

mrobinson (changes requested):
Looks good. I have a few changes to make this just a bit easier to read.

components/script/dom/console.rs
syvb: `to_int` asserts that it's called on an integer - skipping the check would cause panics on non-integer keys (the `usize::try_from` call is returning an Option, not `to_int`).

components/script/dom/console.rs
syvb: I moved the push call to be right before `stringify_object_from_handle_value` (which is just an optimization, the value bits of non-objects were never being read).

components/script/dom/console.rs
syvb: I tried making it use a mutable vector, but it made the stringifiation and memory allocation logic quite a bit more complicated - I don't think the extra complexity is worth some slight performance gains.

mrobinson (approved):

PR #31247, opened 2024-02-02
layout: Properly size `inline-table` elements

mrobinson:
`width` and `min-width` as well as `height` and `min-height` work
differently for tables. Tables don't shrink beyond their content sizes.
This change fixes that.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -1884,8 +1884,13 @@ impl IndependentFormattingContext {
                 // https://drafts.csswg.org/css2/visudet.html#min-max-widths
                 // In this case “applying the rules above again” with a non-auto inline-size
                 // always results in that size.
+                let min_inline_size = min_box_size.inline.max(
+                    non_replaced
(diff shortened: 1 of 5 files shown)


components/layout_2020/flow/line.rs
```diff
@@ -405,6 +405,9 @@ impl InlineBoxLineItem {
             content_rect.start_corner += &relative_adjustement(&style, state.ifc_containing_block);
         }
 
+        // NB: There is no need to set a baseline offset for this BoxFragment, because the
+        // baselines of this InlineFormattingContext is what will propagate to `display:
+        // inline-block` ancestors.
```
Taym95, line 410: should we use `///` doc comments here?

mrobinson: I don't think so. rustdoc comments are only processed when they document rust functions, modules, and globally scoped variables

mrobinson:
@Loirooriol and I are working on a better version of this change that will be uploaded soon as a PR.

PR #31254, opened 2024-02-03
layout: Implement support for font fallback

mrobinson:
This adds an initial implementation of font fallback, on part with the
one used in legacy layout. This causes more tests to fail than pass, but
is a progression. There are several reasons for this.

1. Font matching is done per unicode character rather than per-grapheme.
     We should eventually do per-grapheme fallback or take the approach of
     Chromium, which is to try to shape everything and then fall back when
     there are missing glyphs.
3. Shaping is not done across element boundaries. This is step 2 of proper
     font fallback and shaping, but it causes many tests to fail.
2. Now that fall back is happening, we detect more failures, in particular with 
    white space collapse and Arabic / Persian handling.

The idea is that these changes can be made in a followup.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -8,7 +8,6 @@ use std::mem;
 use app_units::Au;
 use gfx::font::FontMetrics;
 use gfx::text::glyph::GlyphStore;
-use log::warn;
 use serde::Serialize;
 use servo_arc::Arc;
 use style::computed_values::white_space::T as WhiteSpace;
@@ -28,7 +27,7 @@ use super::line::{
     layout_line_items, AbsolutelyPositionedLineItem, AtomicLineItem, FloatLineItem,
     InlineBoxLineItem, LineItem, LineItemLayoutState, LineMetrics, TextRunLineItem,
 };
-use super::text_run::TextRun;
+use super::text_run::{add_or_get_font, get_font_for_first_font_for_style, TextRun};
 use super::CollapsibleWithParentStartMargin;
 use crate::cell::ArcRefCell;
 use crate::context::LayoutContext;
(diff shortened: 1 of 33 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -1423,21 +1461,31 @@ impl InlineFormattingContext {
         };
 
         let style = containing_block.style;
+
+        // It's unfortunate that it isn't possible to get this during IFC text processing, but in
+        // that situation the style of the cotnaining block is unknown.
```
mukilan, line 1466:
```suggestion
        // that situation the style of the containing block is unknown.
```

components/layout_2020/flow/inline.rs
```diff
@@ -1530,26 +1578,41 @@ impl InlineFormattingContext {
...
+                    {
+                        inline_box.default_font_index =
+                            Some(add_or_get_font(&font, &mut ifc_fonts));
+                    }
+                },
+                _ => {},
+            });
         });
+
+        let _ = std::mem::replace(&mut self.font_metrics, ifc_fonts);
```
mukilan, line 1602: Can this not be `self.font_metrics = ifc_fonts` ?

components/layout_2020/flow/text_run.rs
```diff
@@ -30,30 +38,155 @@ pub(crate) struct TextRun {
...
+    prevent_soft_wrap_opportunity_at_end: bool,
+}
+
+// There are two reasons why we might want to break at the start:
+//
+//  1. The line breaker told us that a break was necessary between to separate
+//     instances of sending text to it.
+//  2. We are following replaced content ie `have_deferred_soft_wrap_opportunity`.
+//
+// In both cases, we don't want to do this if the first character
```
mukilan, line 67: Sentence doesn't seem complete
> we don't want to do this if the first character

mukilan (commented):
I've taken an initial look at this PR and noted down some minor issues. However, there are a few concepts that are new to me, so it might be better if we can go through it together.

components/layout_2020/flow/text_run.rs
```diff
@@ -30,30 +38,163 @@ pub(crate) struct TextRun {
...
+    /// Whether or not to prevent a soft wrap opportunity at the end of this [`TextRun`].
+    /// This depends on the whether the last character in the run prevents a soft wrap
+    /// opportunity.
+    prevent_soft_wrap_opportunity_at_end: bool,
+}
+
+// There are two reasons why we might want to break at the start:
+//
+//  1. The line breaker told us that a break was necessary between to separate
+//     instances of sending text to it.
```
atbrakhi, line 64: did you mean to say `between two separate instances`?

components/layout_2020/flow/text_run.rs
```diff
@@ -211,3 +395,51 @@ fn char_prevents_soft_wrap_opportunity_when_before_or_after_atomic(character: ch
...
+pub(super) fn get_font_for_first_font_for_style(
+    style: &ComputedValues,
+    font_context: &mut FontContext<FontCacheThread>,
+) -> Option<FontRef> {
+    let font = font_context
+        .font_group(style.clone_font())
+        .borrow_mut()
+        .first(font_context);
+    if font.is_none() {
+        warn!("Could not find find for style: {:?}", style.clone_font());
```
atbrakhi, line 442:
```suggestion
        warn!("Could not find font for style: {:?}", style.clone_font());
```

atbrakhi (approved):
Left some small nits. Looks good to me. Thank you

mrobinson:
@mukilan @atbrakhi Thanks for the reviews.

PR #31255, opened 2024-02-03
Initial overview of webxr initialization

jdm:
Every time I answer technical questions about this code I have to go and re-read a bunch of files to remind myself how they fit together. This is an attempt to capture that knowledge in a form that's easier to review.

### DIFF
--- a/docs/components/webxr.md
+++ b/docs/components/webxr.md
@@ -1,0 +1,78 @@
+# How WebXR works in Servo
+
+## Terminology
+
+There are three main components involved in Servo's WebXR implementation:
+1. the script thread (runs all JS for a page)
+2. the WebGL thread (maintains WebGL canvas data and invokes GL operations corresponding to [WebGL APIs](https://registry.khronos.org/webgl/specs/latest/1.0/))
+3. the compositor (AKA the main thread)
+
+Additionally, there are a number of WebXR-specific concepts:
+* the [discovery object](https://doc.servo.org/webxr_api/trait.DiscoveryAPI.html) (ie. how Servo discovers if a device can provide a WebXR session)
+* the [WebXR registry](https://doc.servo.org/webxr_api/struct.MainThreadRegistry.html) (the compositor's interface to WebXR)
+* the [layer manager](https://doc.servo.org/webxr_api/layer/trait.LayerManagerAPI.html) (manages WebXR layers for a given session and frame operations on those layers)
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

matlu:
Thanks for the efforts @jdm 
Can we get this added to the Servo Wiki as well. I believe WebXR & OpenXR will be a popular topic going forward?
@mrobinson Maybe coordiate the Wiki update with the fixes landed from the contributors.

mrobinson:
> Thanks for the efforts @jdm Can we get this added to the Servo Wiki as well. I believe WebXR & OpenXR will be a popular topic going forward? @mrobinson Maybe coordiate the Wiki update with the fixes landed from the contributors.

Good point, @matlu. I've added a link at https://github.com/servo/servo/wiki/Design#webxr to the in-tree documentation, but it would be good to have only a single place for documentation like this.

matlu:
> https://github.com/servo/servo/wiki/Design#webxr

Great first step @mrobinson  but at it is under "Design" you would not know about it without opening up design.
Great if we the main sub headings of Design could be visible in the main Wiki page e.g. like:

Design
   - Architecture
   - Description
   - Script
   - Layout
   - Compositor
   - Implementation Strategy
   - Features
      - WebXR
    - Challenges

mrobinson:
Hi @matlu, I've done a revamp of the front page of the wiki so that the design bits are exploded out and WebXR is listed there now. I think this will make the section a lot more discoverable. It's true that there are many obsolete pages. I wonder if we can come up with a strategy for archiving them somehow.

matlu:
> Hi @matlu, I've done a revamp of the front page of the wiki so that the design bits are exploded out and WebXR is listed there now. I think this will make the section a lot more discoverable. It's true that there are many obsolete pages. I wonder if we can come up with a strategy for archiving them somehow.

Great job Martin. Yes it looks a lot better. As mention left to do is to figure out how the "Search box" for this Github entry works.

As we have discovered when searching for WebXR it will not find our nice new section and instead find an obsolete 6 year old wiki post.

image

matlu:
Maybe we can raise a topic under general on Zulip Chat and once there is some feedback/recommendations create a a new issue.
@mrobinson 

**What is needed I believe is:**
1. Find a few to archieve old and obosolete Wiki entries (maybe they can be moved to a "Legacy Wiki"? 
Looks like there are 370(!) pages?

2. Run a reindexing of https://github.com/servo/servo.wiki.git
(I think this should be done before 1 so at least the new entries will be included in search results).

3. Once obsolete pages have been moved to "Leagacy Wiki/Archived", run a new reindex to avoid getting search results for obsolete stuff.

PR #31258, opened 2024-02-05
canvas2d: Implement `.reset()`

syvb:
Implements [`.reset()`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/reset) for `CanvasRenderingContext2D`, which resets the rendering context to its default state. There is already logic for resetting the rendering context for when the width/height is changed, so the implementation is similar.

---

- [X] There are tests for these changes (already in WPT)

### DIFF
--- a/components/script/dom/canvasrenderingcontext2d.rs
+++ b/components/script/dom/canvasrenderingcontext2d.rs
@@ -181,7 +181,7 @@ impl CanvasRenderingContext2DMethods for CanvasRenderingContext2D {
         self.canvas_state.restore()
     }
 
-    // https://html.spec.whatwg.org/multipage/#dom-context-2d-reset
+    /// <https://html.spec.whatwg.org/multipage/#dom-context-2d-reset>
     fn Reset(&self) {
         self.canvas_state.reset()
     }
--- a/components/script/dom/canvasrenderingcontext2d.rs
+++ b/components/script/dom/canvasrenderingcontext2d.rs
@@ -74,7 +74,7 @@ impl CanvasRenderingContext2D {
         self.canvas_state
             .get_ipc_renderer()
             .send(CanvasMsg::Recreate(
-                size.to_u64(),
+                Some(size.to_u64()),
                 self.canvas_state.get_canvas_id(),
             ))
             .unwrap();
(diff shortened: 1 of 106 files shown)


components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -181,6 +181,11 @@ impl CanvasRenderingContext2DMethods for CanvasRenderingContext2D {
         self.canvas_state.restore()
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-context-2d-reset
+    fn Reset(&self) {
```
mrobinson, line 185:
```suggestion
    /// <https://html.spec.whatwg.org/multipage/#dom-context-2d-reset>
    fn Reset(&self) {
```

components/script/dom/offscreencanvasrenderingcontext2d.rs
```diff
@@ -213,6 +213,11 @@ impl OffscreenCanvasRenderingContext2DMethods for OffscreenCanvasRenderingContex
         self.canvas_state.restore()
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-context-2d-reset
```
mrobinson, line 216:
```suggestion
    /// <https://html.spec.whatwg.org/multipage/#dom-context-2d-reset>
```

components/script/dom/canvasrenderingcontext2d.rs
mrobinson: This should likely be converted to a rustdoc comment as it's documenting a function. When doing that though, it is important not to leave raw URLs in the comment.

components/script/dom/paintrenderingcontext2d.rs
```diff
@@ -95,6 +95,11 @@ impl PaintRenderingContext2DMethods for PaintRenderingContext2D {
         self.context.Restore()
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-context-2d-reset
```
mrobinson, line 98:
```suggestion
    /// <https://html.spec.whatwg.org/multipage/#dom-context-2d-reset>
```

mrobinson (approved):
What a great change! I just have a few comments about the rustdoc. Note that most of the pre-existing comments of this form should be converted as well.

PR #31259, opened 2024-02-05
Convert etc/profilicate.py to Python 3

syvb:
Converts  `etc/profilicate.py` (a script for processing raw profiler samples) from Python 2 to Python 3, to make it easier to run on modern systems. I used `2to3`, then removed unnecessary double brackets and updated the shebang.

---

- [X] These changes do not require tests because it is a testing script

### DIFF
--- a/etc/profilicate.py
+++ b/etc/profilicate.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright 2018 The Servo Project Developers. See the COPYRIGHT
 # file at the top-level directory of this distribution.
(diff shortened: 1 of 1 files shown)


jdm (approved):

etc/profilicate.py
```diff
@@ -63,7 +63,7 @@ def contents(self):
 
 tid = 0
 threads = []
-for (name, raw_samples) in sorted(thread_data.iteritems(), key=lambda x: thread_order[x[0]]):
+for (name, raw_samples) in sorted(iter(thread_data.items()), key=lambda x: thread_order[x[0]]):
```
mukilan, line 66: I think the `iter()` is is not needed in this case, since `sorted` can consume iterables directly and `items()` returns view objects that are iterable.

mukilan (approved):

PR #31268, opened 2024-02-05
script: implement navigator.hardwareConcurrency

syvb:
Implements [`navigator.hardwareConcurrency`](https://html.spec.whatwg.org/multipage/#dom-navigator-hardwareconcurrency), which returns the number of logical processors available.

I used the `num_cpus` library (which is already used by Servo in other places), and cache the result since it can be slow to query that information on some operating systems.

---

### DIFF
--- a/components/script/Cargo.toml
+++ b/components/script/Cargo.toml
@@ -75,6 +75,7 @@ mime = { workspace = true }
 mime_guess = { workspace = true }
 msg = { workspace = true }
 net_traits = { workspace = true }
+num_cpus = { workspace = true }
 num-traits = { workspace = true }
 parking_lot = { workspace = true }
 percent-encoding = { workspace = true }
--- a/components/script/dom/navigator.rs
+++ b/components/script/dom/navigator.rs
@@ -2,8 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::convert::TryInto;
+
 use dom_struct::dom_struct;
 use js::jsval::JSVal;
+use lazy_static::lazy_static;
 
 use crate::dom::bindings::codegen::Bindings::NavigatorBinding::NavigatorMethods;
 use crate::dom::bindings::reflector::{reflect_dom_object, DomObject, Reflector};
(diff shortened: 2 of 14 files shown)


components/script/dom/webidls/Navigator.webidl
```diff
@@ -70,3 +71,8 @@ partial interface Navigator {
 partial interface Navigator {
     [Pref="dom.gamepad.enabled"] GamepadList getGamepads();
 };
+
+// https://html.spec.whatwg.org/multipage/#navigatorconcurrenthardware
```
Taym95, line 75:
```suggestion
/// <https://html.spec.whatwg.org/multipage/#navigatorconcurrenthardware>
```

components/script/Cargo.toml
```diff
@@ -75,6 +75,7 @@ mime = { workspace = true }
 mime_guess = { workspace = true }
 msg = { workspace = true }
 net_traits = { workspace = true }
+num_cpus = "1.1.0"
```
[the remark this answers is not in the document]
syvb: Do you mean I should create an issue in `num_cpus`? I don't see why this PR would warrant creating an issue.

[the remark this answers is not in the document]
Taym95: no I meant on this repo

components/script/dom/navigator.rs
```diff
@@ -24,6 +27,13 @@ use crate::dom::window::Window;
 use crate::dom::xrsystem::XRSystem;
 use crate::script_runtime::JSContext;
 
+pub(super) fn hardware_concurrency() -> u64 {
```
mrobinson, line 30: Why is this function public? Could it simply be moved into the body of `HardwareConcurrency(&self)`? Since the code is not repeated, I'm not sure an extra function buys us much unless I'm misunderstanding.

components/script/Cargo.toml
mrobinson, line 78: I'm not sure this needs a bug as in the Servo project we don't necessarily create a bug for every feature or issue.

Now that `num_cpus` is used in `config` and `script` can you make it a workspace dependency so that it's easier to keep it up to date? Please don't modify `style` though as that is a vendored version of Mozilla's style and will probably soon move out of the source tree.

mrobinson (changes requested):
Nice! Just a few comments.

components/script/dom/navigator.rs
syvb: It's also called in `components/script/dom/workernavigator.rs` - I split it out to another function to avoid duplicating the logic in the `Navigator` and `WorkerNavigator`.

components/script/Cargo.toml
syvb: I converted it to a workspace dependency.

mrobinson (approved):

PR #31269, opened 2024-02-06
Implement HTMLImageElement decode

### LINKED ISSUE #31183: Implement HTMLImageElement decode
WebIDL is at https://html.spec.whatwg.org/multipage/embedded-content.html#htmlimageelement

Spec is at https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-decode

How to do it:

1. Add the `decode` member to our [WebIDL](https://github.com/servo/servo/blob/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/script/dom/webidls/HTMLImageElement.webidl)
2. Implement the required method on our [dom struct](https://github.com/servo/servo/tree/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/script/dom)

Doing 2 will require integrating the `decode` concept with the [`ImageContext`](https://github.com/servo/servo/blob/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/script/dom/htmlimageelement.rs#L213) and the [`ImageCache`](https://github.com/servo/servo/blob/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/net/image_cache.rs#L420C6-L420C16)

 Look at the async [decoding](https://github.com/servo/servo/blob/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/net/image_cache.rs#L42) and [handling of the data](https://github.com/servo/servo/blob/d68c7e7881b5c92d0b03c1b43990da26f3771615/components/net/image_cache.rs#L407) happening in the image cache, and find a way to hook it up to a promise in script. 

See also https://html.spec.whatwg.org/multipage/images.html#img-decoding-process

Taym95:
---

- [x] These changes fix #31183


components/script/dom/htmlimageelement.rs
```diff
@@ -1610,6 +1673,22 @@ impl HTMLImageElementMethods for HTMLImageElement {
         }
     }
 
+    /// <https://html.spec.whatwg.org/multipage/#dom-img-decode>
+    fn Decode(&self) -> Rc<Promise> {
+        // Step 1
+        let promise = Promise::new(&self.global());
+
+        // Step 2
+        let task = ImageElementMicrotask::DecodeTask {
```
Taym95, line 1682: @KiChjang can you please review this PR, it is still a draft

mrobinson:
@Taym95 Do you mind taking a look at the failing tests? It seems that this change introduces some failure and timeouts.

Taym95:
> @Taym95 Do you mind taking a look at the failing tests? It seems that this change introduces some failure and timeouts.

I pushed another commit after last `Try run`, maybe we should trigger it again?

components/script/dom/htmlimageelement.rs
```diff
@@ -1366,13 +1422,20 @@ impl MicrotaskRunnable for ImageElementMicrotask {
             } => {
                 elem.react_to_environment_changes_sync_steps(*generation);
             },
+            &ImageElementMicrotask::DecodeTask {
+                ref elem,
+                ref promise,
+            } => {
+                elem.decode_image_data_sync(promise);
```
gterzian, line 1429: So, here we need to break this into two parts:

1. Step 2.1, which run immediately in the microtask.
2. Step 2.2, which runs "in parallel". 

For 2, we need to integrate with the extisting async worflow with the image cache that is setup within [`react_to_environment_changes_sync_steps`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L1024) and which continues [here](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L492). 

I think this is best done by simply adding some state to the image element, so that a pending decoding can be noted and later continued. See for inspiration what is done with [`pending_request`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L166). 

This is actually simpler than I thought, because you don't need to do anything special with the image cache: just keep the decode promise, and then inside [`process_image_response_for_environment_change`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L492C8-L492C53), do the steps in 2.2(except the actual decoding, which happens in the image cache).

gterzian (changes requested):
Great start! Some pointers for the next step. Note that this is actually simpler than what I wrote in the initial issue...

mrobinson:
@Taym95 Any updates on this one? Is there some way that I can help you get this ready for review again?

Taym95:
> @Taym95 Any updates on this one? Is there some way that I can help you get this ready for review again?

Sorry, @mrobinson, I got busy in previous weeks. I will continue working on this this week.

components/script/dom/htmlimageelement.rs
Taym95: > So, here we need to break this into two parts:
> 
> 1. Step 2.1, which run immediately in the microtask.
> 2. Step 2.2, which runs "in parallel".
> 
> For 2, we need to integrate with the extisting async worflow with the image cache that is setup within [`react_to_environment_changes_sync_steps`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L1024) and which continues [here](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L492).
> 
> I think this is best done by simply adding some state to the image element, so that a pending decoding can be noted and later continued. See for inspiration what is done with [`pending_request`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L166).
> 
> This is actually simpler than I thought, because you don't need to do anything special with the image cache: just keep the decode promise, and then inside [`process_image_response_for_environment_change`](https://github.com/servo/servo/blob/e76770202c059202889bdb3555e9efb817b6f68e/components/script/dom/htmlimageelement.rs#L492C8-L492C53), do the steps in 2.2(except the actual decoding, which happens in the image cache).

@gterzian I moved decode to `process_image_response_for_environment_change`, please review again, thanks!

Taym95:
> @Taym95 Any updates on this one? Is there some way that I can help you get this ready for review again?

@mrobinson Can you run `wpt` on this PR? I ran it on my fork, and a lot of tests are passing now. Thanks!

gterzian (changes requested):
The goal of the `decode` API, is to do the decoding in parallel, and resolve the promise when it is completed, see the [example use case](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element:in-parallel-3). 

So, to do this we should:

1. At the end of the `ImageElementMicrotask::DecodeTask` micro task, check for `State::CompletelyAvailable`, and if so, resolve the promise. 
2. Otherwise, resolve the promise as part of [`handle_loaded_image`](https://github.com/servo/servo/blob/df75bf7a6db8a7865134aa526067dc339be455fb/components/script/dom/htmlimageelement.rs#L422). Sorry `process_image_response_for_environment_change` is the wrong place, I just found out after an extensive review of the current structure. 

Further notes:

- I think we need a `image_decode_promise: DomRefCell<Vec<Rc<Promise>>>`, because there can be multiple calls to `decode`, and each one should resove. 
- There is some weird stuff inside the image store, but I will file a separate issue. https://github.com/servo/servo/issues/32441

components/net/image_cache.rs
```diff
@@ -570,6 +570,25 @@ impl ImageCache for ImageCacheImpl {
...
     }
 
+    fn decode(
+        &self,
+        url: ServoUrl,
+        origin: ImmutableOrigin,
+        cors_setting: Option<CorsSettings>,
+    ) -> Option<Image> {
+        let mut store = self.store.lock().unwrap();
+
```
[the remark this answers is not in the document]
Taym95: Thanks for review @gterzian, what confused me is where to actually decode the img, I added this decode function to decode the img, but if I resolve the promise in `handle_loaded_image` I don't need this decode function because image in `handle_loaded_image` is already decoded, right?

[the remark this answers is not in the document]
gterzian: Yes you are correct.

components/script/dom/htmlimageelement.rs
```diff
@@ -1366,13 +1400,35 @@ impl MicrotaskRunnable for ImageElementMicrotask {
             } => {
                 elem.react_to_environment_changes_sync_steps(*generation);
             },
+            ImageElementMicrotask::DecodeTask {
```
gterzian, line 1403: for this one, let's follow the pattern of the other tasks and put the handler into a method on `elem`.

Taym95:
Thanks, @gterzian, for your review and guidance. I have a clear idea about decoding now. I updated the code, and more tests are passing, but some are still timing out. I am checking them.

components/script/dom/htmlimageelement.rs
```diff
@@ -186,8 +192,14 @@ impl HTMLImageElement {
 
         match self.current_request.borrow().state {
             // If image's current request's state is broken, then throw an "InvalidStateError" DOMException.
-            State::Broken => Err(Error::InvalidState),
-            State::CompletelyAvailable => Ok(true),
+            State::Broken => {
```
gterzian, line 195: I don't think we should do either here, because this is a method that does not transition the state of `current_request`. For example, if `Decode` where to be called, then the earliest point for the promise to resolve would be in the microtask, whereas with these changes an intermediate call to `is_usable` would resolve the promise.

components/script/dom/htmlimageelement.rs
```diff
@@ -414,6 +426,7 @@ impl HTMLImageElement {
 
     // Steps common to when an image has been loaded.
     fn handle_loaded_image(&self, image: Arc<Image>, url: ServoUrl) {
+        self.resolve_image_decode_promises();
```
gterzian, line 429: Let's do this at the end of the function, since otherwise the promise handler code will not see the updates happening below.

components/script/dom/htmlimageelement.rs
```diff
@@ -517,6 +530,10 @@ impl HTMLImageElement {
 
     /// <https://html.spec.whatwg.org/multipage/#abort-the-image-request>
     fn abort_request(&self, state: State, request: ImageRequestPhase) {
+        if matches!(state, State::Broken | State::Unavailable) {
```
gterzian, line 533: I don't think it should include `State::Unavailable`, see "This [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element) element's [current request](https://html.spec.whatwg.org/multipage/images.html#current-request)'s [state](https://html.spec.whatwg.org/multipage/images.html#img-req-state) becomes [broken](https://html.spec.whatwg.org/multipage/images.html#img-error)"

components/script/dom/htmlimageelement.rs
```diff
@@ -832,7 +849,14 @@ impl HTMLImageElement {
                         self.image_request.set(ImageRequestPhase::Pending);
                         self.init_image_request(&mut pending_request, url, src);
                     },
-                    (_, State::Broken) | (_, State::Unavailable) => {
+                    (_, State::Broken) => {
+                        self.reject_image_decode_promises();
```
gterzian, line 853: I think this one also does not belong here, because the request is already in the broken state.

components/script/dom/htmlimageelement.rs
```diff
@@ -1562,7 +1626,14 @@ impl HTMLImageElementMethods for HTMLImageElement {
         let request = self.current_request.borrow();
         let request_state = request.state;
         match request_state {
-            State::CompletelyAvailable | State::Broken => true,
+            State::CompletelyAvailable => {
```
gterzian, line 1629: Same here, this is not the point where the request becomes broken or completely available.

components/script/dom/htmlimageelement.rs
gterzian: Also, like in the other place, let's reject at the end of the function, so that the handler sees the updates.

gterzian (changes requested):
Couple of more comments, note we should only resolve/reject when the request's state changes, and also we should do this only after the state has changed(so usually bottom of function, not top before the mutations).

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1170,43 @@ impl HTMLImageElement {
...
+        } else if matches!(
+            self.current_request.borrow().state,
+            State::CompletelyAvailable
+        ) {
+            self.resolve_image_decode_promises();
+        } else if !elem.has_attribute(&local_name!("src")) &&
+            !elem.has_attribute(&local_name!("srcset"))
+        {
+            self.reject_image_decode_promises();
+        }
```
Taym95, line 1191: @gterzian, I need to add this condition(src, srcset) to make other tests pass, but I am not sure if this is the correct place. Also, two remaining SVG tests are failing. Does Servo Image support SVG?

Taym95: I will remove this one because does not align with spec

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1170,38 @@ impl HTMLImageElement {
...
+        } else if matches!(
+            self.current_request.borrow().state,
+            State::CompletelyAvailable
+        ) {
+            self.resolve_image_decode_promises();
+        }
+    }
+
+    fn resolve_image_decode_promises(&self) {
+        for promise in self.image_decode_promises.borrow().iter() {
```
gterzian, line 1190: should drain the vec.

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1170,38 @@ impl HTMLImageElement {
...
+            promise.resolve_native(&());
+        }
+    }
+
+    fn reject_image_decode_promises(&self) {
+        let document = document_from_node(self);
+        for promise in self.image_decode_promises.borrow().iter() {
```
gterzian, line 1197: should drain the vec.

gterzian (changes requested):
We should drain the vec of promises when resolving or rejecting them.

gterzian:
> TIMEOUT [expected PASS] subtest: HTMLImageElement.prototype.decode(), basic tests. Image without src/srcset fails decode.

The problem here is that if no src is set, the `update_the_image_data` step will not run, so the promise will never resolve. 

I think we should solve this in the microtask handler, by checking for `self.select_image_source().is_none()` and then rejecting with an error. 

-----

Another thing is that I just saw that `abort_request` sometimes also sets the request to completely available, so we should resolve in that case I think.

Taym95:
> > TIMEOUT [expected PASS] subtest: HTMLImageElement.prototype.decode(), basic tests. Image without src/srcset fails decode.
> 
> The problem here is that if no src is set, the `update_the_image_data` step will not run, so the promise will never resolve(we are currently piggybacking on that worflow to deal with trying to fetch the image).
> 
> But, I think we should run `update_the_image_data` also when the element is created, per the spec:
> 
> > the user agent must synchronously [update the image data](https://html.spec.whatwg.org/multipage/images.html#update-the-image-data) of the [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element) element, with the restart animation flag set if so stated, whenever that element is created or has experienced [relevant mutations](https://html.spec.whatwg.org/multipage/images.html#relevant-mutations)
> 
> source: https://html.spec.whatwg.org/multipage/#when-to-obtain-images
> 
> So I would say let's run `update_the_image_data` when the element is created at
> 
> https://github.com/servo/servo/blob/3a892faedeb71766ab80ddd55382de189ebb1611/components/script/dom/htmlimageelement.rs#L1313
> 
> Another thing is that I just saw that `abort_request` sometimes also sets the request to completely available, so we should resolve in that case I think.

Thanks, I addressed these remarks

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1172,40 @@ impl HTMLImageElement {
...
+        ) {
+            self.resolve_image_decode_promises();
+        }
+    }
+
+    fn resolve_image_decode_promises(&self) {
+        for promise in self.image_decode_promises.borrow().iter() {
+            promise.resolve_native(&());
+        }
+        self.image_decode_promises.borrow_mut().clear();
```
gterzian, line 1195: would it not be better to just do a `drain(...)` above?

components/script/dom/htmlimageelement.rs
```diff
@@ -1345,6 +1397,9 @@ pub enum ImageElementMicrotask {
         elem: DomRoot<HTMLImageElement>,
         generation: u32,
     },
+    DecodeTask {
+        elem: DomRoot<HTMLImageElement>,
```
gterzian, line 1401: One last thing I thought about: there is currently a race condition between when the microtask runs, and the "in parallel" stuff described at 2.2.  For example if the request becomes broken, we will reject all promises, including the one for which the micro task hasn't run yet. 

I'm not sure this is actually a problem, but we may as well be consitent with the spec and:

1. In `Decode`, not store the promise on `pending_`, but rather store it here in the task. 
2. change `react_to_decode_image_sync_steps` as described above.

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1172,40 @@ impl HTMLImageElement {
...
     }
 
+    // Step 2 for <https://html.spec.whatwg.org/multipage/#dom-img-decode>
+    fn react_to_decode_image_sync_steps(&self) {
+        let document = document_from_node(self);
+        // Step 2.1 of <https://html.spec.whatwg.org/multipage/#dom-img-decode>
+        if !document.is_fully_active() ||
+            matches!(self.current_request.borrow().state, State::Broken)
+        {
+            self.reject_image_decode_promises();
```
gterzian, line 1182: here, as per the comment below, let's only reject the promise associated with the task, and remove the resolve logic below. Instead, if the promise is not rejected, we should store it on `pending_`, and resolve/reject it later as the request state changes. 

I'm going to run the wpt tests before this change, and after, just to see if this makes a difference.

gterzian (changes requested):
Ok, one last thing I'd like to check.

components/script/dom/htmlimageelement.rs
Taym95: We use `drain(..)` from the doc I saw:

```
  /// // A full range clears the vector, like `clear()` does
 /// v.drain(..);
```

components/script/dom/htmlimageelement.rs
Taym95: I removed the resolve logic but it broke lot of tests.

components/script/dom/htmlimageelement.rs
Taym95: I moved storing promise in task and tests are okey

components/script/dom/htmlimageelement.rs
gterzian: Ok, I will just run the tests once to see, but probably we should then keep it. I guess its a kind of race condition in the spec, between the microtask and the parallel steps.

gterzian (changes requested):
Ok, one comment and lets undo the last commit please.

gterzian:
Note also there is a discussion ongoing about the potential race condition [I mentioned](https://github.com/servo/servo/pull/31269#discussion_r1634662446), at https://github.com/whatwg/html/issues/4217.

components/script/dom/htmlimageelement.rs
```diff
@@ -1159,6 +1172,47 @@ impl HTMLImageElement {
...
+        let document = document_from_node(self);
+        // Step 2.1 of <https://html.spec.whatwg.org/multipage/#dom-img-decode>
+        if !document.is_fully_active() ||
+            matches!(self.current_request.borrow().state, State::Broken)
+        {
+            promise.reject_native(&DOMException::new(
+                &document.global(),
+                DOMErrorName::EncodingError,
+            ));
+        } else if matches!(
```
gterzian, line 1186: Here I would add  a comment that this does not follow the spec, but it is discussed at https://github.com/whatwg/html/issues/4217

gterzian (approved):
LGTM

gterzian:
Thank you, great work! @Taym95

Issue #31134, opened 2024-01-19
Android: apk crashes during startup.


mukilan:
Logs:

```
01-19 18:22:28.275  3931  3961 D simpleservo: thread 'Script(1,1)' panicked at /home/mukilan/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/7184f65/mozjs/src/gc/root.rs:150:28:
01-19 18:22:28.275  3931  3961 D simpleservo: misaligned pointer dereference: address must be a multiple of 0x8 but is 0xa85f2d8c
```

Enabling backtrace doesn't help as the names of functions are not printed.

rajveermalviya:
I am seeing similar crash on Apple silicon, should JIT be disabled temporarily for it too?

```shell
> ./mach run -r

UNSUPPORTED (log once): POSSIBLE ISSUE: unit 1 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
misaligned pointer dereference: address must be a multiple of 0x8 but is 0x4b4b4b4b4b4b4b4b (thread Script(1,1), at /Users/rajveer/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/d1f1519/mozjs/src/rust.rs:809)
   0: <servo::backtrace::Print as core::fmt::Debug>::fmt
   1: core::fmt::rt::Argument::fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/fmt/rt.rs:138:9
      core::fmt::write
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/fmt/mod.rs:1114:21
   2: std::io::Write::write_fmt
   3: servo::backtrace::print
   4: servo::main2::main::{{closure}}
   5: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/alloc/src/boxed.rs:2021:9
      std::panicking::rust_panic_with_hook
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:735:13
   6: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:609:13
   7: std::sys_common::backtrace::__rust_end_short_backtrace
[...]

PR #31270, opened 2024-02-06
android: disable JIT in SM to workaround #31134

mukilan:
The crash when loading servo.org happens in the JIT code emitted by SM's CacheIRCompiler to invoke the VM function `ProxyGetPropertyByValue`.

To disable this code path, it is not sufficient to disable just the baseline JIT (which exposed in servo under the pref `js.baseline.enabled`) but the baseline interpreter which is controlled by a different flag in SM needs to be disabled as well.

This PR disables uses `js.baseline.enabled` pref in Servo to control both the baseline JIT and the interpreter together.

- [x] There are no tests for these changes since the `./mach test-android-startup` command still needs more work to make it run on GitHub CI.


components/script/script_runtime.rs
```diff
@@ -530,6 +530,11 @@ unsafe fn new_rt_and_cx_with_parent(
 
     // Enable or disable the JITs.
     let cx_opts = &mut *ContextOptionsRef(cx);
+    JS_SetGlobalJitCompilerOption(
+        cx,
+        JSJitCompilerOption::JSJITCOMPILER_BASELINE_INTERPRETER_ENABLE,
+        pref!(js.baseline.enabled) as u32,
```
mukilan, line 536: I've made  `js.baseline.enabled` control both the baseline JIT and baseline interpreters. Let me know if we should have separate prefs for both for e.g `js.baseline.jit.enabled` and `js.baseline.interpreter.enabled`

mrobinson: I think it might make sense to have a separate pref for this, as it seems the baseline JIT and baseline interpreter are two different things: https://hacks.mozilla.org/2019/08/the-baseline-interpreter-a-faster-js-interpreter-in-firefox-70/

ports/jniapi/src/lib.rs
```diff
@@ -845,13 +846,20 @@ fn get_options(
     };
 
     let native_window = unsafe { ANativeWindow_fromSurface(env.get_native_interface(), surface) };
+
+    // FIXME: enable JIT compilation on Android after the
+    // startup crash issue (#31134) is fixed.
```
mrobinson, line 851: This could probably be a single line.

mrobinson (approved):
Looks good, though I think before landing you could add a new pref for the baseline interpreter.

components/script/script_runtime.rs
mukilan: Sure, do we want to keep the current `js.baseline.enabled` or change it to `js.baseline.jit.enabled`?

mrobinson: Oh, let's change it. It was probably created before the baseline interpreter existed. How about this:

- `js.baseline-jit.enabled`
- js.baseline-interpreter.enabled`

mukilan: I've used underscores instead of dashes since that is consistent with the existing nested option `eager_unsafe_compilation` under `js.baseline` and also avoids the `serde(rename)` attributes in config/prefs.js since we'll anyway need underscores in rust syntax.

PR #31277, opened 2024-02-06
layout: Respond to shorthand property requests with real values

### LINKED ISSUE #22838: Shorthand properties of computed CSSStyleDeclaration don't return correct values
```html
<div id="test" style="overflow: hidden"></div>
<script>onload = function() {
  var div = document.getElementById('test');
  console.log(getComputedStyle(div).overflow)
  console.log(getComputedStyle(div).overflowX);
};
</script>
```
In Firefox, this shows `hidden` twice. In Servo this shows an empty string, followed by `hidden`.

sebsebmc:
Instead of responding with an empty string for shorthand properties, use the mapped longhand properties to respond to computed style requests.

---

- [X] These changes fix #22838 

- [X] There are tests for these changes, besides the manual testcase for this there are a number of WPT tests that rely on this behavior to pass such as `/css/css-flexbox/parsing/flex-computed.html`, `/css/css-flexbox/parsing/flex-flow-computed.html` and `/css/css-animations/computed-style-animation-parsing.html`


sebsebmc:
Can the maintainer approving the workflows on this also kick off a WPT run?

components/layout_2020/query.rs
```diff
@@ -258,9 +260,22 @@ pub fn process_resolved_style_request<'dom>(
...
     let longhand_id = match *property {
         PropertyId::LonghandAlias(id, _) | PropertyId::Longhand(id) => id,
-        // Firefox returns blank strings for the computed value of shorthands,
-        // so this should be web-compatible.
-        PropertyId::ShorthandAlias(..) | PropertyId::Shorthand(_) => return String::new(),
+        PropertyId::ShorthandAlias(id, _) | PropertyId::Shorthand(id) => {
+            use style::values::resolved::Context;
+            let mut block = PropertyDeclarationBlock::new();
+            let mut dest = String::new();
+            for longhands in id.longhands() {
```
Taym95, line 267:
```suggestion
            for longhand in id.longhands() {
```

sebsebmc:
Ive found that there is a generated type LonghandsToSerialize that seem to include the correct logic for only serializing properties of shorthands that are allowed in the shorthands https://github.com/servo/servo/blob/5facf436f6835aa53a500ae99168b8a00ed4802c/components/style/properties/shorthands/font.mako.rs#L219-L225  but as a Rust beginner it doesn't seem to me that I can access this member of shorthands because each shorthand type implements a different LonghandToSerialize type so there would need to be a trait made somewhere?

There are also some WPT tests that use shorthands that are failing because the serialization is expected to be returned in a shorter form than what this solution currently generates. Its not clear to me that shorter forms can be generated without knowing which longhand properties were set and which are default?

tests/wpt/meta/css/cssom/cssom-getPropertyValue-common-checks.html.ini
```diff
@@ -16,3 +16,6 @@
 
   [Can't serialize shorthand set to 'initial' when some longhand is missing]
     expected: FAIL
+
+  [Can't serialize shorthand set to initial value when some longhand is missing]
+    expected: FAIL
```
Loirooriol, line 21: This shouldn't be regressing, but it's not your fault, you will need to rebase after #31299

components/layout_2020/query.rs
```diff
@@ -379,9 +394,22 @@ pub fn process_resolved_style_request_for_unstyled_node<'dom>(
...
+                    style.computed_or_resolved_declaration(longhands, Some(&Context { style })),
+                    Importance::Normal,
+                );
+            }
+            let css = block.shorthand_to_css(id, &mut dest);
+            match css {
+                Ok(_) => return dest.to_owned(),
+                Err(_) => return String::new(),
+            }
+        },
```
Loirooriol, line 412: Instead of repeating the same code, I would move it into a helper function

components/layout_2020/query.rs
```diff
@@ -392,6 +394,26 @@ pub fn process_resolved_style_request_for_unstyled_node<'dom>(
...
+        block.push(
+            style.computed_or_resolved_declaration(longhands, Some(&Context { style })),
+            Importance::Normal,
+        );
+    }
+    let css = block.shorthand_to_css(id, &mut dest);
+    match css {
+        Ok(_) => return dest.to_owned(),
+        Err(_) => return String::new(),
+    }
```
Loirooriol, line 414: You can omit `return`:

```suggestion
    match block.shorthand_to_css(id, &mut dest) {
        Ok(_) => dest.to_owned(),
        Err(_) => String::new(),
    }
```

Loirooriol (approved):
You need to update the test expectations.

So if I get it right, the regressions are:

 - `/html/rendering/non-replaced-elements/the-hr-element-0/hr.html`, addressed by #31297
 - `/css/css-fonts/font-shorthand-serialization-prevention.html` because servo doesn't support `all-small-caps`
 - `/css/css-backgrounds/animations/background-position-interpolation.html`

It would be good to land #31297 first and investigate what's happening with `background-position-interpolation.html`, but otherwise this looks good.

sebsebmc:
`/css/css-fonts/font-shorthand-serialization-prevention.html` does not fail because servo doesn't support the `all-small-caps`. It fails because shorthands are not supposed to serialize if the longhand values present in the shorthand are not settable via the shorthand. `all-small-caps` is one of those, and so the shorthand should not serialize (hence the name of the test `font-shorthand-serialization-prevention`)

Loirooriol:
No, the test fails because Servo doesn't support `font-variant: all-small-caps` nor `font-variant-caps: all-small-caps`, as you can see via `CSS.supports`.

Therefore, the assignment fails, and the longhands can be represented by a shorthand value, so the shorthand serializes.

Loirooriol:
`background-position-interpolation.html` is also failing on Firefox: https://wpt.fyi/results/css/css-backgrounds/animations/background-position-interpolation.html

PR #31290, opened 2024-02-08
mach: Make `./mach try` a little friendlier

mrobinson:
1. Move `./mach try` to `testing_commands.py which is a bit more
   consistent.
2. Make `./mach try` print out the remote name always and properly
   form the URL for ssh repositories.
3. Print out the try configuration matrix to make it more obvious
   what is being triggered remotely.
4. Better error handling. Print and error and exit if the remote isn't
   on GitHub and also clean up properly if something fails after making
   the temporary commit.

---

- [x] These changes do not require tests because they just change a build environment command.

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -15,6 +15,7 @@ import os
 import os.path as path
 import shutil
 import subprocess
+import textwrap
 
 import wpt
 import wpt.manifestupdate
@@ -27,6 +28,7 @@ from mach.decorators import (
     Command,
 )
 
+import servo.try_parser
 import tidy
 
 from servo.command_base import BuildType, CommandBase, call, check_call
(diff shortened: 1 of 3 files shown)


python/servo/testing_commands.py
```diff
@@ -747,3 +749,51 @@ def smoketest(self, build_type: BuildType, params):
...
+        remote_url = subprocess.check_output(["git", "config", "--get", f"remote.{remote}.url"]).decode().strip()
+        if "github.com" not in remote_url:
+            print(f"The remote provided ({remote_url}) isn't a GitHub remote.")
+            return 1
+
+        try_string = " ".join(try_string)
+        config = servo.try_parser.Config(try_string)
+        print(f"Trying on {remote} ({remote_url}) with following configuration.")
+        print()
+        print(textwrap.indent(config.to_json(indent=2), prefix="  | "))
```
sagudev, line 767: Indent is good but `|` is bothering me, if one would want to copy config.

sagudev (commented):
If we are doing this could we also make git commands quiet using `--quiet`.

python/servo/testing_commands.py
```diff
@@ -747,3 +749,51 @@ def smoketest(self, build_type: BuildType, params):
         # of panics won't cause timeouts on CI.
         return self.context.commands.dispatch('run', self.context, build_type=build_type,
                                               params=params + ['-f', 'tests/html/close-on-load.html'])
+
+    @Command('try', description='Runs try jobs by force pushing to try branch', category='devenv')
+    @CommandArgument('--remote', '-r', default="origin", help='A git remote to run the try job on')
+    @CommandArgument('try_string', default=None, nargs='...',
```
sagudev, line 755:
```suggestion
    @CommandArgument('try_string', default=["full"], nargs='...',
```

python/servo/testing_commands.py
```diff
@@ -747,3 +749,51 @@ def smoketest(self, build_type: BuildType, params):
...
+                     help="A list of try strings specifying what kind of job to run.")
+    def try_command(self, remote="origin", try_string=None):
```
sagudev, line 757:
```suggestion
    def try_command(self, remote="origin", try_string=["full"]):
```

python/servo/testing_commands.py
sagudev: This is needed for `mach try` without empty string to run full try run.

python/servo/testing_commands.py
```diff
@@ -747,3 +749,51 @@ def smoketest(self, build_type: BuildType, params):
...
+
+        # From here on out, we need to always clean up the commit we added to the branch.
+        try:
+            result = call(["git", "push", "--quiet", remote, "--force", "HEAD:try"])
+            if result != 0:
+                return result
+
+            # TODO: This is a pretty naive approach to turning a GitHub remote URL (either SSH or HTTPS)
+            # into a URL to the Actions page. It might be better to create this action with the `gh`
+            # tool and get the real URL.
```
sagudev, line 786: This would mean that we would depend on gh being installed and authorized and given that this is not really crucial for `mach try` I think naive approach is good enought.

python/servo/testing_commands.py
```diff
@@ -747,3 +749,51 @@ def smoketest(self, build_type: BuildType, params):
...
+                actions_url = actions_url.replace(':', '/')
+                actions_url = actions_url.replace("git@", "")
+                actions_url = f"https://{actions_url}"
+            print(f"Actions available at: {actions_url}")
+
+        finally:
+            # Remove the last commit which only contains the try configuration.
+            result = call(["git", "reset", "--quiet", "--soft", "HEAD~1"])
+            if result != 0:
+                print("Could not clean up try commit. Sorry! Please try to reset to HEAD~1.")
```
sagudev, line 798: > Please try to reset to HEAD~1.

Not sure everybody knows what `HEAD~1` refers to.

python/servo/testing_commands.py
mrobinson: I've moved this and the other default to `@CommandArgument`.

python/servo/testing_commands.py
mrobinson: Fair point

python/servo/testing_commands.py
mrobinson: Okay. I've modified this to "the previous commit."

sagudev (changes requested):
Something is wrong: https://github.com/sagudev/servo/actions/runs/7828494013/job/21358490263#step:5:54

I ran `mach try webgpu full`

mrobinson:
> Something is wrong: https://github.com/sagudev/servo/actions/runs/7828494013/job/21358490263#step:5:54
> 
> I ran `mach try webgpu full`

It seems there was a bug when the commit message was more than a single line. I think I've fixed that now.

mrobinson:
Here's an example of the same test you ran: https://github.com/mrobinson/servo/actions/runs/7828643293/job/21358950457

sagudev (approved):
Thanks for polishing `mach try`

mrobinson:
> Thanks for polishing `mach try`

Sure thing. Thanks for implementing it originally. It's really useful!

PR #31292, opened 2024-02-08
layout: Use `BoxFragment` border widths for display list generation

mrobinson:
Instead of using the border widths from the style, use the ones recorded by the `BoxFragment`. This is necessary because inline layout can override these border widths during fragmentation. For instance, when a box is split across two lines only one fragment should have an inline start border.

---

### DIFF
--- a/components/layout_2020/flow/line.rs
+++ b/components/layout_2020/flow/line.rs
@@ -302,11 +302,6 @@ impl InlineBoxLineItem {
             border.inline_start = Length::zero();
             margin.inline_start = Length::zero();
         }
-        if !self.is_last_fragment_of_ib_split {
-            padding.inline_end = Length::zero();
-            border.inline_end = Length::zero();
-            margin.inline_end = Length::zero();
-        }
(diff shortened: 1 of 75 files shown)


Loirooriol (approved):

components/layout_2020/flow/line.rs
```diff
@@ -302,11 +302,6 @@ impl InlineBoxLineItem {
             border.inline_start = Length::zero();
             margin.inline_start = Length::zero();
         }
-        if !self.is_last_fragment_of_ib_split {
-            padding.inline_end = Length::zero();
-            border.inline_end = Length::zero();
-            margin.inline_end = Length::zero();
-        }
```
Loirooriol, line 309: I'm not sure why this can be removed, but nothing is failing, so...

mrobinson: I neglected to mention this in the commit message, but it's a related cleanup. The borders are also zeroed out at the bottom of the same function so these lines weren't really doing anything.

PR #31297, opened 2024-02-08
<hr> elements are expected to have a default overflow:hidden

sebsebmc:
See https://github.com/whatwg/html/pull/2724

This adds the required styling to the UA stylesheet. 

---

I'm not sure why the WPT test /html/rendering/non-replaced-elements/the-hr-element-0/hr.html currently passes, but while fixing #22838 the test started failing. Once both these land the test should be passing again.

### DIFF
--- a/resources/user-agent.css
+++ b/resources/user-agent.css
@@ -266,7 +266,7 @@ input[type="submit"], select, button {
 
 input, textarea, select, button { display: inline-block; }
 
-hr { color: gray; border-style: inset; border-width: 1px; margin: 0.5em auto; }
+hr { color: gray; border-style: inset; border-width: 1px; margin: 0.5em auto; overflow: hidden; }
 
 
 fieldset {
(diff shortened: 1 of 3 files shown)


mrobinson (changes requested):
This is a nice change, but while fixing this would it be possible to update `<hr>` to use the full set of style defined at https://html.spec.whatwg.org/multipage/rendering.html#the-hr-element-2:

```css
hr {
  color: gray;
  border-style: inset;
  border-width: 1px;
  margin-block-start: 0.5em;
  margin-inline-end: auto;
  margin-block-end: 0.5em;
  margin-inline-start: auto;
  overflow: hidden;
}
```

mrobinson (approved):

Loirooriol:
The test is also passing on legacy layout, you need to remove `tests/wpt/meta-legacy-layout/html/rendering/non-replaced-elements/the-hr-element-0/setting-overflow-visible.html.ini`

PR #31298, opened 2024-02-09
BHM: Support aarch64 for mac sampler

### LINKED ISSUE #30953: BHM: update Mac sampler to support aarch64
Current implementation seems to assume x86

https://github.com/servo/servo/blob/7305c59304a776dfcb197f91c02b1d28fe8a29ff/components/background_hang_monitor/sampler_mac.rs#L72

rajveermalviya:
Replaced unmaintained `mach` crate with `mach2`, and added support for getting arm registers.

---

- [x] These changes fix #30953 (GitHub issue number if applicable)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,7 @@ dependencies = [
  "lazy_static",
  "libc",
  "log",
- "mach",
+ "mach2",
  "msg",
  "nix 0.25.1",
  "serde_json",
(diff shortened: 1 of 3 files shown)


rajveermalviya:
The only other crate in the dependency graph which still uses `mach` is `surfman`, so I guess that will need a patch too.

mrobinson (approved):

PR #31310, opened 2024-02-11
Cargo.toml: explictly specify num_cpus version

syvb:
This fixes a warning from Cargo about relying on deprecated behavior:
```
warning: /home/loops/src/servo/servo/components/script/Cargo.toml: dependency (num_cpus) specified without providing a local path, Git repository, version, or workspace dependency to use. This will be considered an error in future versions
warning: /home/loops/src/servo/servo/components/config/Cargo.toml: dependency (num_cpus) specified without providing a local path, Git repository, version, or workspace dependency to use. This will be considered an error in future versions
```

---

- [X] These changes do not require tests because there isn't a way to test this, since Cargo doesn't support promoting warnings from Cargo itself into errors

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ mime_guess = "2.0.3"
 mozangle = "0.5.1"
 msg = { path = "components/shared/msg" }
 net_traits = { path = "components/shared/net" }
-num_cpus = { workspace = true }
+num_cpus = "1.1.0"
 num-traits = "0.2"
 parking_lot = "0.12"
 percent-encoding = "2.3"


Cargo.toml
```diff
@@ -71,7 +71,7 @@ mime_guess = "2.0.3"
 mozangle = "0.5.1"
 msg = { path = "components/shared/msg" }
 net_traits = { path = "components/shared/net" }
-num_cpus = { workspace = true }
+num_cpus = { version = "1.1.0" }
```
atouchet, line 74: This can just be:

`num_cpus = "1.1.0"`

syvb: I updated it.

mrobinson (approved):
Thanks for this fix.

PR #31314, opened 2024-02-11
Fix list-style serialization

### LINKED ISSUE #31307: Broken serialization for computed `list-style`
```html
<!DOCTYPE html>
<pre>
<script>
function compute(value) {
  var element = document.body;
  var style = document.body.style.listStyle = value;
  document.writeln(getComputedStyle(document.body).listStyle);
}
compute("outside none disc"); // FAIL: '', expected 'outside' or 'disc'
compute("outside none square"); // FAIL: ' square', expected 'square'
compute("outside url(about:invalid) disc"); // FAIL: ' url("about:invalid")', expected 'url("about:invalid")'
compute("outside url(about:invalid) square"); // FAIL: ' url("about:invalid") square', expected 'url("about:invalid") square'
compute("inside none disc"); // FAIL: '', expected 'inside'
compute("inside none square"); // FAIL: ' square', expected 'inside square'
compute("inside url(about:invalid) disc"); // FAIL: ' url("about:invalid")', expected 'inside url("about:invalid")'
compute("inside url(about:invalid) square"); // FAIL: ' url("about:invalid") square', expected 'inside url("about:invalid") square'
</script>
```

Firefox is correct. Blink and WebKit ignore the shortest serialization principle, but at least they are sound and don't add weird leading spaces.

IIRC, the `list-style` properties are implemented quite different than Gecko (due to a lack of `@counter-style` and such) which probably explains the difference.

Loirooriol:
In layout2020, `list-style-position` is disabled behind a pref, so the list_style_position field is an Option.

The serialization of the `list-style` shorthand wasn't correctly handling the case of it being None.

---

- [X] These changes fix #31307

### DIFF
--- a/components/style/properties/shorthands/list.mako.rs
+++ b/components/style/properties/shorthands/list.mako.rs
@@ -113,14 +113,14 @@
             #[cfg(feature = "gecko")]
             let position_is_initial = self.list_style_position == &ListStylePosition::Outside;
             #[cfg(feature = "servo")]
-            let position_is_initial = self.list_style_position == Some(&ListStylePosition::Outside);
+            let position_is_initial = matches!(self.list_style_position, None | Some(&ListStylePosition::Outside));
             if !position_is_initial {
                 self.list_style_position.to_css(dest)?;
                 have_one_non_initial_value = true;
             }
             if self.list_style_image != &ListStyleImage::None {
                 if have_one_non_initial_value {
-                   dest.write_char(' ')?;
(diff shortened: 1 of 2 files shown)


Loirooriol:
We should also consider not skipping WPT in css-lists, I guess it would have detected 3a23598080289afb8ccb2dcf096b739b1aa44136 breaking something.

mrobinson (approved):

PR #31317, opened 2024-02-11
WebIDL: Use Uint8ClampedArray instead of raw JSObject in bindings

Taym95:
Part #31064, we would like to remove the direct use of `unsafe NonNull<JSObject>`, and replace it with use of the appropriate safe wrapper found in [typedarray.rs](https://github.com/servo/mozjs/blob/de842d84a07eabac4e1bfe72a954b74ee0244b82/mozjs/src/typedarray.rs), in this PR we are replacing `unsafe NonNull<JSObject>` with `Uint8ClampedArray `, according to [spec](https://github.com/servo/servo/blob/c94d5842db11330ff32b0dd2b8ce036c53b410bb/components/script/dom/webidls/CanvasRenderingContext2D.webidl#L262) 

---


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -128,6 +128,7 @@ def wrapInNativeContainerType(type, inner):
     IDLType.Tags.uint32array: 'Uint32Array',
     IDLType.Tags.float32array: 'Float32Array',
     IDLType.Tags.float64array: 'Float64Array',
+    IDLType.Tags.uint8clampedarray: 'Uint8ClampedArray',
```
Taym95, line 131: @gterzian this is PR ready for review

gterzian (changes requested):
Thanks for the contrib again! Let's change the `new_initialized` method a bit.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -31,6 +31,16 @@ where
     T: TypedArrayElement + TypedArrayElementCreator,
     T::Element: Clone + Copy,
 {
+    pub fn new_initialized(js_object: Option<*mut JSObject>) -> HeapTypedArray<T> {
```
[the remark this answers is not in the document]
Taym95: This is much better, I updated the code, thanks

components/script/dom/imagedata.rs
```diff
@@ -180,18 +176,20 @@ impl ImageData {
...
     }
 
-    // https://html.spec.whatwg.org/multipage/#dom-imagedata-data
-    fn Data(&self, _: JSContext) -> NonNull<JSObject> {
-        NonNull::new(self.data.get()).expect("got a null pointer")
+    /// <https://html.spec.whatwg.org/multipage/#dom-imagedata-data>
+    fn Data(&self, _: JSContext) -> Uint8ClampedArray {
+        self.data
+            .get_internal()
+            .expect("Failed to get Data from ImageData.")
```
gterzian, line 193: Let's return `JsError` as we did in the other PR.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -26,6 +26,35 @@ unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
...
+            phantom: PhantomData::default(),
+        },
+        HeapTypedArrayInit::Info { len, cx } => {
+            rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
+            let typed_array =
+                create_typed_array_with_length::<T>(cx, len as usize, array.handle_mut())
+                    .expect("Creating TypedArray from length should never fail");
+            let heap_typed_array = HeapTypedArray::<T>::default();
+            heap_typed_array
+                .set_data(cx, unsafe { typed_array.as_slice() })
```
gterzian, line 46: unsafe is unnecessary, because the entire module allows it.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -26,6 +26,35 @@ unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
...
+            internal: Heap::boxed(js_object),
+            phantom: PhantomData::default(),
+        },
+        HeapTypedArrayInit::Info { len, cx } => {
+            rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
+            let typed_array =
+                create_typed_array_with_length::<T>(cx, len as usize, array.handle_mut())
+                    .expect("Creating TypedArray from length should never fail");
+            let heap_typed_array = HeapTypedArray::<T>::default();
+            heap_typed_array
```
gterzian, line 45: I think you can simply do `heap_typed_array.internal.set(*typed_array)`

components/script/dom/imagedata.rs
```diff
@@ -64,8 +64,9 @@ impl ImageData {
     ) -> Fallible<DomRoot<ImageData>> {
         // checking jsobject type
         let cx = GlobalScope::get_cx();
-        typedarray!(in(*cx) let array_res: Uint8ClampedArray = jsobject);
-        let array = array_res.map_err(|_| {
+        let heap_typed_array =
+            new_initialized_heap_typed_array(HeapTypedArrayInit::Object(jsobject));
+        let array = heap_typed_array.acquire_data(cx).map_err(|_| {
```
gterzian, line 69: `acquire_data` detaches the buffer, so instead we want to use `get_internal`.

gterzian (changes requested):
Thanks for the contrib! Couple of changes suggested...

components/script/dom/imagedata.rs
Taym95: should we do it for other changes we merged, we have lot of `.get_internal().expect` ?

gterzian: I think so, but it depends on the spec. It would be good to do this in a separate PR so each cases can be investigated separately(it may require updating test expectations also).

Taym95: I think we only return error when function returns `Fallible` but not in this case

components/script/dom/bindings/typedarrays.rs
Taym95: I still get `this operation is unsafe and requires an unsafe function or block` when I remove it

Taym95: I thought it's better to keep `unsafe` here instead of using it in `new_with_jsobject`

components/script/dom/imagedata.rs
gterzian: yes! Although we can check the spec to see if the function should be fallible...

components/script/dom/bindings/typedarrays.rs
gterzian: ah yes sorry, you need the unsafe block(but there is no need to mark the method as allowing unsafe, which is what I had in mind somehow).

components/script/dom/bindings/typedarrays.rs
Taym95: Yes `heap_typed_array.internal.set(*array)` should be okey

components/script/dom/imagedata.rs
Taym95: Indeed thanks, fixed

components/script/dom/imagedata.rs
gterzian: The spec doesn't say it direclty, for wording like "initialize the [data](https://html.spec.whatwg.org/multipage/canvas.html#dom-imagedata-data) attribute of imageData to a new [Uint8ClampedArray](https://webidl.spec.whatwg.org/#idl-Uint8ClampedArray) object" implies using the Web IDL and JS operations which are in fact fallible. 

Let's return JsError.

components/script/dom/bindings/typedarrays.rs
```diff
@@ -31,6 +31,32 @@ unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
     }
 }
 
+pub fn new_initialized_heap_typed_array<T>(init: HeapTypedArrayInit) -> HeapTypedArray<T>
```
gterzian, line 34: Let's return a result, and return the error from `create_typed_array_with_length`. (creating the buffer is falling, as we saw in the other PR).

gterzian (changes requested):
Just a few last changes!

components/script/dom/bindings/typedarrays.rs
Taym95: done

components/script/dom/imagedata.rs
Taym95: done

components/script/dom/bindings/typedarrays.rs
```diff
@@ -31,6 +31,40 @@ unsafe impl<T> crate::dom::bindings::trace::JSTraceable for HeapTypedArray<T> {
...
 
+pub fn new_initialized_heap_typed_array<T>(
+    init: HeapTypedArrayInit,
+) -> Result<HeapTypedArray<T>, ()>
+where
+    T: TypedArrayElement + TypedArrayElementCreator,
+    T::Element: Clone + Copy,
+{
+    match init {
+        HeapTypedArrayInit::Object(js_object) => Ok(HeapTypedArray {
```
gterzian, line 42: nit:

let heap_typed_array = match init { early return for errors}
Ok( heap_typed_array)

gterzian (approved):
Awesome! LGTM with just one nit...

components/script/dom/bindings/typedarrays.rs
Taym95: Done

gterzian: What I meant was the below, I think it's simpler because there is only one OK
```rust
    let heap_typed_array = match init {
        HeapTypedArrayInit::Object(js_object) => HeapTypedArray {
            internal: Heap::boxed(js_object),
            phantom: PhantomData::default(),
        },
        HeapTypedArrayInit::Info { len, cx } => {
            rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
            let typed_array_result =
                create_typed_array_with_length::<T>(cx, len as usize, array.handle_mut());
            if typed_array_result.is_err() {
                return Err(());
            }
            let heap_typed_array = HeapTypedArray::<T>::default();
            heap_typed_array.internal.set(*array);
            heap_typed_array
        },
    };
    Ok(heap_typed_array)
```

gterzian (approved):
Great work, thanks!

gterzian:
I'll add it to the queue, although it may require a rebase.

PR #31323, opened 2024-02-12
Add `libegl1-mesa-dev` to runtime deps

sagudev:
---

- [x] These changes do not require tests because it's only docs fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ Run Servo with the command:
 * `libXi`
 * `libxkbcommon`
 * `vulkan-loader`
+* `libegl1-mesa-dev`
 
 ## Developing
 


mukilan:
I've been investigating this issue. I've confirmed using runs on my [fork](https://github.com/mukilan/servo/actions/runs/7871214265) that it is related to the [GStreamer](url) upgrade in #30750. However, I've not been able to figure out how this PR introduced the dependency.

mrobinson (approved):

mukilan:
Looks like the issue is with disabling the media stack on Ubuntu 20.04, rather than the upgrade itself i.e the issue is reproducible if we build servo with --media-stack=dummy on a commit older than 7fa4ea9.

The panic that we see happens in surfman's logic for creating a new display connection, which in turn relies on loading symbols dynamically from `libEGL.so` using `dlopen` and [dlsym](https://github.com/servo/surfman/blob/main/surfman/src/platform/generic/egl/device.rs#L68).

With media stack enabled, servo binary gets linked to `/lib/x86_64-linux-gnu/libEGL.so.1` at compile time and this versioned .so is installed on 20.04 by default. My guess is the dlopen/dlsym calls are satisified because the dynamic linker has already loaded `libEGL.so.1`.

With media stack disabled, there is no link to libEGL.so.1 in the servo binary. However, installing the `libegl1-mesa-dev` package creates a symbolic link from `libEGL.so` (unversioned) to `libEGL.so.1`, which in turn allows `dlsym` to file the symbols needed by Surfman's EGL binding loader.

PR #31342, opened 2024-02-13
Fix rooting of external array buffer pointer

### LINKED ISSUE #31339: Fix rooting of external array buffer pointer
I think there is one thing that we, and the previous code, missed, which is immediate rooting of [the pointer returned by `NewExternalArrayBuffer` ](https://github.com/servo/servo/pull/31202/files#diff-8f10f5d3caee3cce7befe080a3333a5f9402a651884287734f41e786e4323181R186). 

GC can happen, even though it appears like it couldn't, for example if it occurs in response to a timer in another SpiderMonkey GC thread, see second part of https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/esr78/docs/GC%20Rooting%20Guide.md#performance-tweaking

See gecko: https://searchfox.org/mozilla-central/rev/9bb5d5f55da6cc7163dd93825c25609b769822f9/dom/webgpu/Buffer.cpp#291

We should fix this like: `rooted!(in(cx) let object = NewExternalArrayBuffer(..))`.

Taym95:
`rooted!` is needed to ensure memory safety and prevent potential garbage collection issues.
doc: https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/esr78/docs/GC%20Rooting%20Guide.md#performance-tweaking
 
Gecko example : https://searchfox.org/mozilla-central/rev/9bb5d5f55da6cc7163dd93825c25609b769822f9/dom/webgpu/Buffer.cpp#291

example:  

---

- [x] These changes fix #31339

### DIFF
--- a/components/script/dom/bindings/typedarrays.rs
+++ b/components/script/dom/bindings/typedarrays.rs
@@ -183,16 +183,18 @@ where
         let mapping_slice_ptr =
             mapping.lock().unwrap().borrow_mut()[offset as usize..m_end as usize].as_mut_ptr();
 
-        let array_buffer = NewExternalArrayBuffer(
+        // rooted! is needed to ensure memory safety and prevent potential garbage collection issues.
+        // https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/esr78/docs/GC%20Rooting%20Guide.md#performance-tweaking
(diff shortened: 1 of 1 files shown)


gterzian (approved):
Awesome! Can we just get a PR message that includes the problem, and then states the solution(like it currently does)? Something like "this fixes..."

PR #31345, opened 2024-02-13
Do not adjust margins in over-constrained cases

Loirooriol:
This deviates from css2, but it's mandated by css-align, and matches what other browsers do when no margin is 'auto'.

When some margin is 'auto', this should keep the proper round-tripping behavior that Gecko and WebKit lack, and Blink recently adopted.

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -907,6 +907,7 @@ impl NonReplacedFormattingContext {
 
         let margin_inline_start;
         let margin_inline_end;
+        let effective_margin_inline_start;
         let (margin_block_start, margin_block_end) =
             solve_block_margins_for_in_flow_block_level(&pbm);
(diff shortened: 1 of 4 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -1284,93 +1295,118 @@ fn solve_containing_block_padding_border_and_margin_for_in_flow_box<'a>(
...
+            (start, end)
+        },
+    }
+}
+
+/// Resolves 'auto' margins of an in-flow block-level box in the inline axis
+/// similarly to |solve_inline_margins_for_in_flow_block_level|. However,
+/// they align within the provided rect (instead of the containing block),
+/// to avoid overlapping floats.
+/// Additionally to the used the margins, it also returns what we will call
```
mrobinson, line 1343:
```suggestion
/// In addition to the used margins, it also returns what we will call
```

mrobinson (approved):

PR #31346, opened 2024-02-13
script: Do not run layout in a thread

### LINKED ISSUE #31215: Run layout in the script thread
Came-up during the TSC meeting. Since script (almost?) always blocks on layout queries, we may as well remove the thread and run script + layout on the same thread. But...? 

See other issues:

https://github.com/servo/servo/issues/14885
https://github.com/servo/servo/issues/22507

mrobinson:
Instead of spawning a thread for layout that almost always runs
synchronously with script, simply run layout in the script thread.

This is a resurrection of #28708, taking just the bits that remove the
layout thread. It's a complex change and thus is just a first step
toward cleaning up the interface between script and layout. Messages are
still passed from script to layout via a `process()` method and script
proxies some messages to layout from other threads as well.

Big changes:

1. Layout is created in the script thread on Document load, thus every
   live document is guaranteed to have a layout. This isn't completely
   hidden in the interface, but we can safely `unwrap()` on a Document's
   layout.
2. Layout configuration is abstracted away into a LayoutConfig struct
   and the LayoutFactory is a struct passed around by the Constellation.
   This is to avoid having to monomorphize the entire script thread
   for each layout.
3. Instead of having the Constellation block on the layout thread to
   figure out the current epoch and whether there are pending web fonts
   loading, updates are sent synchronously to the Constellation when
   rendering to a screenshot. This practically only used by the WPT.

A couple tests start to fail, which is probably inevitable since removing
the layout thread has introduced timing changes in "exit after load" and
screenshot behavior.

Co-authored-by: Josh Matthews <josh@joshmatthews.net>

---

- [x] These changes fix #31215.


emilio:
@mrobinson do we have data on how often script blocks on layout? It seems that modulo layout queries it shouldn't?

mrobinson:
> @mrobinson do we have data on how often script blocks on layout? It seems that modulo layout queries it shouldn't?

All layout is done synchronously with script. Previously, animations could run independently in the layout thread, but this was a bit of a hack since they didn't update style. Nowadays they also run synchronously on the script thread, though there should eventually be an animation thread. Essentially the only time that layout now runs independently of script is when responding to requests from the Constellation about whether or not it is time to output a screenshot.

Cargo.lock
```diff
@@ -4831,6 +4814,7 @@ dependencies = [
  "euclid",
  "fnv",
  "fxhash",
+ "gfx",
```
mukilan, line 4817: script compile time?

components/constellation/constellation.rs
```diff
@@ -302,11 +297,16 @@ pub struct Constellation<Message, LTF, STF, SWF> {
     /// from the background hang monitor.
     background_hang_monitor_receiver: Receiver<Result<HangMonitorAlert, IpcError>>,
 
-    /// An IPC channel for layout threads to send messages to the constellation.
-    /// This is the layout threads' view of `layout_receiver`.
+    /// A factory for creating layouts. This allows customizing the kind
+    /// of layout created for a [`Constellation`] and preventing a circular dependency
```
mukilan, line 301: what is the circular dependency being prevented? creation? or import?

components/layout_thread/lib.rs
```diff
@@ -768,49 +621,7 @@ impl LayoutThread {
...
-                    // Just ignore these messages at this point.
-                },
-                _ => panic!("layout: unexpected message received after `PrepareToExitMsg`"),
-            }
-        }
-    }
-
-    /// Shuts down the layout thread now. If there are any DOM nodes left, layout will now (safely)
-    /// crash.
+    /// Shuts down layout now. If there are any DOM nodes left, layout will now (safely) crash.
```
mukilan, line 624: Would it make sense to say `script thread will now safely crash` here?

components/layout_thread_2020/lib.rs
```diff
@@ -186,89 +160,28 @@ pub struct LayoutThread {
...
-                        time_profiler_chan,
-                        mem_profiler_chan.clone(),
-                        webrender_api_sender,
-                        paint_time_metrics,
-                        busy,
-                        window_size,
-                    );
-
-                    let reporter_name = format!("layout-reporter-{}", id);
-                    mem_profiler_chan.run_with_memory_reporting(
```
mukilan, line 261: Will we still be able to profile memory and time with sufficient granularity (e.g script work vs layout work) after this change?

components/script/dom/document.rs
```diff
@@ -2092,7 +2093,7 @@ impl Document {
             },
             LoadType::PageSource(_) => {
                 if self.has_browsing_context && self.is_fully_active() {
-                    // Note: if the document is not fully active, the layout thread will have exited already.
+                    // Note: if the document is not fully active, layout will have exited already.
```
mukilan, line 2096: Is this comment relevant? That is, what does it mean for layout to have exited in this new architecture?

components/script/dom/window.rs
```diff
@@ -1676,7 +1659,7 @@ impl Window {
         // objects removed from the tree that haven't been collected
         // yet). There should not be any such DOM nodes with layout
         // data, but if there are, then when they are dropped, they
-        // will attempt to send a message to the closed layout thread.
+        // will attempt to send a message to layout.
         // This causes memory safety issues, because the DOM node uses
         // the layout channel from its window, and the window has
```
mukilan, line 1664: Is this comment still relevant?

components/script/dom/window.rs
```diff
@@ -1803,18 +1786,14 @@ impl Window {
         ScriptThread::handle_tick_all_animations_for_testing(pipeline_id);
     }
 
-    /// Reflows the page unconditionally if possible and not suppressed. This
-    /// method will wait for the layout thread to complete (but see the `TODO`
-    /// below). If there is no window size yet, the page is presumed invisible
-    /// and no reflow is performed. If reflow is suppressed, no reflow will be
-    /// performed for ForDisplay goals.
-    ///
-    /// TODO(pcwalton): Only wait for style recalc, since we have
```
mukilan, line 1812: Just for my understanding, is this TODO referring to an optimization that was possible with separate layout threads, but no longer possible after this change?

components/script/dom/document.rs
mrobinson: In this case it means that the Constellation has sent an `ExitNow` message to layout. I'm not sure if it's still a relevant situation, but I think that this change is a preliminary step toward removing the `ExitNow` message and perhaps simplifying this bit of code. The thing is that this requires a bit of experimentation and testing to make sure we don't unintentionally break anything, so I think it makes sense to do this as a followup (along with many other cleanups this unlocks).

components/script/dom/window.rs
mrobinson: I think this is still relevant, because DOM nodes still access the layout channel by first talking to the `Window`. I hope that we can clean this up and make this impossible in a later cleanup, but it's all very delicate and I don't want to rework too much in one change.

components/script/dom/window.rs
mrobinson: Yes, unfortunately this was the unmet promise of the layout thread, but it was never implemented as far as I could tell. The thing is that the specifications lay down some pretty strict ordering requirements regarding the Document event loop and layout and furthermore even among different Documents that share the same origin and iframe tree structure. This is why all iframe tests are incredibly flaky in Servo.

components/layout_thread_2020/lib.rs
mrobinson: I think we lose some granularity there, but:

1. The memory profile infrastructure is fairly abandoned right now.
2. This brings layout on par with script. If we need to break down object allocation per pipeline it would be better to do it in both layout and script in the same way.

components/layout_thread/lib.rs
```diff
@@ -561,87 +481,43 @@ impl LayoutThread {
...
-        // Notify the background-hang-monitor we are waiting for an event.
-        self.background_hang_monitor.notify_wait();
-
-        let request = select! {
-            recv(self.pipeline_port) -> msg => Request::FromPipeline(msg.unwrap()),
-            recv(self.port) -> msg => Request::FromScript(msg.unwrap()),
-            recv(self.font_cache_receiver) -> msg => { msg.unwrap(); Request::FromFontCache }
+    fn handle_request<'a, 'b>(&mut self, request: Request) {
+        let rw_data = self.rw_data.clone();
+        let mut possibly_locked_rw_data = Some(rw_data.lock().unwrap());
```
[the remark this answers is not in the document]
mrobinson: Oh, good observation. I think there will never be contention for this lock and an `Arc` clone is pretty cheap. That said, this is one of the things that we can completely remove after making this change. `rw_data` should be able to go away completely.

mrobinson (commented):
Thanks for the thoughtful review @mukilan. I've tried to answer you comments below and have pushed a new commit which tries to clarify some of the code comments you pointed out.

components/script/script_thread.rs
```diff
@@ -2918,60 +2902,32 @@ impl ScriptThread {
 
     /// Handles a request to exit a pipeline and shut down layout.
     fn handle_exit_pipeline_msg(&self, id: PipelineId, discard_bc: DiscardBrowsingContext) {
-        debug!("Exiting pipeline {}.", id);
+        debug!("{id}: Starting pipeling exit.");
```
mukilan, line 2905:
```suggestion
        debug!("{id}: Starting pipeline exit.");
```

components/script/script_thread.rs
```diff
@@ -2918,60 +2902,32 @@ impl ScriptThread {
...
+        let document = self.documents.borrow_mut().remove(id);
+        if let Some(document) = document {
             if let Some(parser) = document.get_current_parser() {
                 parser.abort();
             }
-        }
-
-        // We should never have a pipeline that's still an incomplete load,
-        // but also has a Document.
-        debug_assert!(idx.is_none() || document.is_none());
```
mukilan, line 2944: It seems like this assertion could be retained as it could be helpful in checking the invariant, even though `idx` is no longer needed to get the layout for incomplete loads.

components/script/script_thread.rs
```diff
@@ -857,12 +853,26 @@ impl ScriptThreadFactory for ScriptThread {
...
             .expect("Thread spawning failed");
-
-        (sender, receiver)
     }
 }
 
 impl ScriptThread {
+    pub fn with_layout<'a, T>(
+        pipeline_id: PipelineId,
+        call: Box<dyn FnOnce(&mut dyn Layout) -> T + 'a>,
```
mukilan, line 862: Do we need `call` to be boxed? Can it not be `impl FnOnce()`?

components/script/script_thread.rs
mrobinson: I've re-added this assertion.

mrobinson (commented):
Thanks for the review. I think I've addressed all the review comments.

mukilan (approved):

PR #31347, opened 2024-02-13
ci: Merge similar try jobs when possible

mrobinson:
This comes up a lot when triggering wpt-2013 and wpt-2020. Instead of
merging the jobs and triggering both layouts, the try parser will
trigger two separate jobs. Running two of the same builds at once seems
to cause the CI to fail one of them [1]. This fixes that issue.

1. An example of this: https://github.com/servo/servo/actions/runs/7892269495

---

### DIFF
--- a/python/servo/try_parser.py
+++ b/python/servo/try_parser.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 
 import json
 import sys
-from typing import Optional
+from typing import ClassVar, List, Optional
 import unittest
 import logging
 
@@ -56,13 +56,17 @@ class JobConfig(object):
     profile: str = "release"
     unit_tests: bool = False
     wpt_tests_to_run: str = ""
+    # These are the fields that must match in between two JobConfigs for them to be able to be
+    # merged. If you modify any of the fields above, make sure to update this line as well.
+    merge_compatibility_fields: ClassVar[List[str]] = ['workflow', 'profile', 'wpt_tests_to_run']
 
     def merge(self, other: JobConfig) -> bool:
         """Try to merge another job with this job. Returns True if merging is successful
            or False if not. If merging is successful this job will be modified."""
-        if self.workflow != other.workflow or self.profile != other.profile or \
-                self.wpt_tests_to_run != other.wpt_tests_to_run:
-            return False
+        for field in self.merge_compatibility_fields:
+            if getattr(self, field) != getattr(other, field):
+                return False
+
         self.wpt_layout |= other.wpt_layout
         self.unit_tests |= other.unit_tests
         return True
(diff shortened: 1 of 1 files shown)


sagudev:
When triggering WPT for both layouts there is `wpt` preset, so just creating `T-wpt` label should run wpt on both layouts. That being said is still think it's good we do merging if possible.

The error you se are due to how artifacts are shared, that I hope will be solved with artifacts-v4 that I am working on.

python/servo/try_parser.py
```diff
@@ -68,15 +80,15 @@ def handle_preset(s: str) -> Optional[JobConfig]:
     elif s in ["wpt", "linux-wpt"]:
         return JobConfig("Linux WPT", Workflow.LINUX, unit_tests=True, wpt_layout=Layout.all())
     elif s in ["wpt-2013", "linux-wpt-2013"]:
-        return JobConfig("Linux WPT legacy-layout", Workflow.LINUX, wpt_layout=Layout.layout2013)
+        return JobConfig("Linux WPT", Workflow.LINUX, wpt_layout=Layout.layout2013)
```
sagudev, line 83: Can we just keep different names here and ignore name field when merging.

mrobinson: The problem is that the merged job will keep the name of the first job, which will be incorrect. I would like to make a later change that generates the job name lazily, but for now this is the simplest approach.

python/servo/try_parser.py
```diff
@@ -55,6 +57,16 @@ class JobConfig(object):
     unit_tests: bool = False
     wpt_tests_to_run: str = ""
 
+    def merge(self, other: JobConfig) -> bool:
+        """Try to merge another job with this job. Returns True if merging is successful
+           or False if not. If merging is successful this job will be modified."""
+        if self.workflow != other.workflow or self.profile != other.profile or \
+                self.wpt_tests_to_run != other.wpt_tests_to_run:
+            return False
```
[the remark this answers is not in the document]
mrobinson: That's an interesting idea, but it seems a bit heavy-weight and people adding new fields to JobConfig will still have to remember to update this function. Instead I've created a new class variable which holds the fields that cannot differ for a merge to happen. This is near the fields and I've added doc to remind people to update the list.

In addition, I've added more tests, which should also help prevent mistakes in the future.

python/servo/try_parser.py
sagudev: Sounds good to me!

python/servo/try_parser.py
```diff
@@ -192,6 +214,43 @@ def test_empty(self):
...
+                                  'wpt_layout': 'all',
+                                  'wpt_tests_to_run': ''
+                              }]
+                              })
+
+        # These two jobs should not be merged to one that turns on unit tests.
+        a = JobConfig("Linux", Workflow.LINUX, unit_tests=True)
+        b = JobConfig("Linux", Workflow.LINUX, unit_tests=False)
+        self.assertTrue(a.merge(b))
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
```
sagudev, line 234: I think the comment is wrong here as jobs are merged to one, fight?

python/servo/try_parser.py
```diff
@@ -192,6 +214,43 @@ def test_empty(self):
...
+
+        # These two jobs should not be mergable to one that turns on unit tests.
+        a = JobConfig("Linux", Workflow.LINUX, unit_tests=True)
+        b = JobConfig("Linux", Workflow.MACOS, unit_tests=True)
+        self.assertFalse(a.merge(b))
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
```
sagudev, line 240: This comment is wrong to as it compares different workflows.

python/servo/try_parser.py
```diff
@@ -192,6 +214,43 @@ def test_empty(self):
...
+
+        # These two jobs should not be merged to one that turns on unit tests.
+        a = JobConfig("Linux", Workflow.LINUX, unit_tests=True)
+        b = JobConfig("Linux", Workflow.LINUX, unit_tests=False)
+        self.assertTrue(a.merge(b))
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
+
+        # These two jobs should not be mergable to one that turns on unit tests.
+        a = JobConfig("Linux", Workflow.LINUX, unit_tests=True)
+        b = JobConfig("Linux", Workflow.MACOS, unit_tests=True)
```
sagudev, line 238:
```suggestion
        b = JobConfig("Mac", Workflow.MACOS, unit_tests=True)
```

python/servo/try_parser.py
```diff
@@ -192,6 +214,43 @@ def test_empty(self):
...
+        self.assertFalse(a.merge(b))
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
+
+        # Ditto.
```
sagudev, line 242: No comment is better then ditto comment IMO.

python/servo/try_parser.py
mrobinson: I think my comment was written in a pretty confusing way. I've removed all these comments and converted them to assert messages, which is more useful, I think.

python/servo/try_parser.py
```diff
@@ -192,6 +214,40 @@ def test_empty(self):
...
+        b = JobConfig("Mac", Workflow.MACOS, unit_tests=True)
+        self.assertFalse(a.merge(b), "Should not merge jobs with different workflows.")
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
+
+        a = JobConfig("Linux", Workflow.LINUX, unit_tests=True)
+        b = JobConfig("Linux", Workflow.LINUX, unit_tests=True, profile="production")
+        self.assertFalse(a.merge(b), "Should not merge jobs with different profiles.")
+        self.assertEqual(a, JobConfig("Linux", Workflow.LINUX, unit_tests=True))
+
+        # Ditto.
```
sagudev, line 245:
```suggestion
```

sagudev (approved):
One suggestion remaining but otherwise looks good.

mrobinson (commented):
Thanks for the review, @sagudev.

PR #31349, opened 2024-02-14
style: Reduce Servo's diff with upstream `to_shmem`

mrobinson:
This is part of getting Servo using an version of style closer to
upstream. This change reverts some changes we made to `to_shmem` and
`to_shmem_derive` in order to reduce our diff with upstream stylo.

---

- [x] These changes do not require tests because this does not change behavior.


delan (approved):
Looks good! `allowed_duplication_types` was in the [original upstream version](https://phabricator.services.mozilla.com/D17187) of to_shmem but then removed in [D147669](https://phabricator.services.mozilla.com/D147669), so I suppose that part of the patch is effectively a cherry-pick.

PR #31351, opened 2024-02-14
style: Add a `static_prefs` implementation

mrobinson:
This will eventually be part of the stylo crate and reduces the diff
between our verson of style and upstream's.

---


Loirooriol (approved):
Looks great!

PR #31357, opened 2024-02-15
ci: Switch to version 4 of GitHub artifact actions

mrobinson:
This switches to version 4 of the GitHub artifact actions, which
requires producing a single artifact per job and adding merge steps. In
addition, the names of artifacts are standardized:

- Build: `<profile>-binary-<platform>`
- Full WPT results (only on failure): `wpt-full-logs-<platform>-<layout>`
- Filtered WPT results (only on failure): `wpt-filtered-logs-<platform>-<layout>`

---

- [x] These changes do not require tests because they just change the CI behavior.

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -80,11 +80,13 @@ jobs:
           GITHUB_CONTEXT: ${{ toJson(github) }}
       - name: Archive results (filtered)
         uses: actions/upload-artifact@v4
+        if: ${{ always() }}
         with:
           name: wpt-filtered-logs-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}
           path: wpt-filtered-logs/*/
       - name: Archive results (full)
         uses: actions/upload-artifact@v4
+        if: ${{ always() }}
         with:
           name: wpt-full-logs-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}
           path: wpt-full-logs/*/
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -20,8 +20,10 @@ on:
 env:
   RUST_BACKTRACE: 1
   SHELL: /bin/bash
-  WPT_COMMAND_LINE_ARG: "${{ inputs.wpt-layout == 'layout-2013' && '--legacy-layout' || '' }}"
   GST_PLUGIN_FEATURE_RANK: pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
+  INTERMITTENT_TRACKER_DASHBOARD_SECRET: ${{ !inputs.wpt-sync-from-upstream && secrets.INTERMITTENT_TRACKER_DASHBOARD_SECRET || '' }}
+  WPT_LAYOUT_ARG: "${{ inputs.wpt-layout == 'layout-2013' && '--legacy-layout' || '' }}"
+  WPT_ALWAYS_SUCCEED_ARG: "${{ inputs.wpt-sync-from-upstream && '--always-succeed' || '' }}"
 
 jobs:
   linux-wpt:
@@ -45,12 +47,12 @@ jobs:
         with:
           ref: refs/pull/${{ github.event.number }}/head
           fetch-depth: 2
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
-          name: release-binary
-          path: release-binary
+          name: ${{ inputs.profile }}-binary-linux
+          path: ${{ inputs.profile }}-binary-linux
       - name: unPackage binary
-        run: tar -xzf release-binary/target.tar.gz
+        run: tar -xzf ${{ inputs.profile }}-binary-linux/target.tar.gz
       - name: Bootstrap dependencies
         run: |
           python3 -m pip install --upgrade pip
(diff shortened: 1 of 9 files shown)


.github/workflows/linux-wpt.yml
```diff
@@ -62,75 +64,66 @@ jobs:
...
-      - "linux-wpt"
+    if: ${{ always() }}
+    needs: linux-wpt
     steps:
+      - name: Merge logs (filtered)
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: wpt-filtered-logs-linux-${{ inputs.wpt-layout }}
+          pattern: wpt-filtered-logs-linux-${{ inputs.wpt-layout }}-*
+          delete-merged: true
```
sagudev, line 104: That's was the issue I was having on my experiments, but `delete-merged` solves it.

.github/workflows/linux-wpt.yml
```diff
@@ -62,75 +64,66 @@ jobs:
...
-          path: |
-            test-wpt.${{ matrix.chunk_id }}.log
-      - name: Archive logs
-        uses: actions/upload-artifact@v3
-        if: ${{ inputs.wpt-sync-from-upstream }}
+          name: wpt-filtered-logs-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}
+          path: wpt-filtered-logs/*/
+      - name: Archive results (full)
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() || inputs.wpt-sync-from-upstream }}
```
sagudev, line 88: Could we make full logs always available as I occasionally need them.

mrobinson: Yes, sure. I've made this change.

.github/workflows/linux-wpt.yml
```diff
@@ -62,75 +64,64 @@ jobs:
...
+            $WPT_ALWAYS_SUCCEED_ARG \
+            ${{ inputs.wpt-tests-to-run }} \
             --${{ inputs.profile }} --processes $(nproc) --timeout-multiplier 2 \
             --total-chunks ${{ env.max_chunk_id }} --this-chunk ${{ matrix.chunk_id }} \
-            --log-raw test-wpt.${{ matrix.chunk_id }}.log \
-            --log-raw-unexpected unexpected-test-wpt.${{ matrix.chunk_id }}.log \
-            --filter-intermittents filtered-test-wpt.${{ matrix.chunk_id }}.json
+            --log-raw wpt-full-logs/linux-${{ inputs.wpt-layout }}/${{ matrix.chunk_id }}.log \
+            --log-raw-unexpected wpt-filtered-logs/linux-${{ inputs.wpt-layout }}/${{ matrix.chunk_id }}.log \
+            --filter-intermittents wpt-filtered-logs/linux-${{ inputs.wpt-layout }}/${{ matrix.chunk_id }}.json
```
sagudev, line 78: How does sync handle those?

.github/workflows/linux-wpt.yml
```diff
@@ -62,75 +64,64 @@ jobs:
...
-          name: wpt-logs-linux-${{ inputs.wpt-layout }}
-          path: |
-            test-wpt.${{ matrix.chunk_id }}.log
-      - name: Archive logs
-        uses: actions/upload-artifact@v3
-        if: ${{ inputs.wpt-sync-from-upstream }}
+          name: wpt-filtered-logs-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}
+          path: wpt-filtered-logs/*/
+      - name: Archive results (full)
+        uses: actions/upload-artifact@v4
```
sagudev, line 87: I think we need `if: ${{ always() }}` here.

.github/workflows/linux-wpt.yml
```diff
@@ -62,75 +64,64 @@ jobs:
...
-        with:
-          name: wpt-filtered-results-linux-${{ inputs.wpt-layout }}
-          path: |
-            filtered-test-wpt.${{ matrix.chunk_id }}.json
-            unexpected-test-wpt.${{ matrix.chunk_id }}.log
-      - name: Archive logs
-        uses: actions/upload-artifact@v3
-        if: ${{ failure() && !inputs.wpt-sync-from-upstream }}
+      - name: Archive results (filtered)
+        uses: actions/upload-artifact@v4
```
sagudev, line 82: We also need `if: ${{ always() }}` here.

.github/workflows/linux-wpt.yml
mrobinson: Sync uses the full logs which are produced unconditionally.

.github/workflows/linux-wpt.yml
mrobinson: Thanks for pointing that out. This is done now.

mrobinson:
Examples of runs:
 - A commit to main: https://github.com/mrobinson/servo/actions/runs/7927614317
 - A WPT sync (only fails because the lack of a token to post the PR): https://github.com/mrobinson/servo/actions/runs/7930382648

sagudev:
[this](https://github.com/mrobinson/servo/actions/runs/7930382648/job/21654575960#step:8:27) does look like real error:
```
 FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/servo/servo/wpt-full-logs-linux-layout-2013/test-wpt.*.log'
```

mrobinson:
> [this](https://github.com/mrobinson/servo/actions/runs/7930382648/job/21654575960#step:8:27) does look like real error:
> 
> ```
>  FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/servo/servo/wpt-full-logs-linux-layout-2013/test-wpt.*.log'
> ```

Thanks for pointing that out. Here's a run with the updated version of the branch:

https://github.com/mrobinson/servo/actions/runs/7941029162/job/21683631506

.github/workflows/linux-wpt.yml
sagudev: There is a lot of [`Conflicting metadata values for ...`](https://github.com/mrobinson/servo/actions/runs/7941029162/job/21683631506#step:8:12) that didn't happened before because some tests are run/logged multiple times for filtering. While I do not think this is blocker, solving this problem would be nice followup (maybe using filtered results for wpt-update would reduce flakes in WPT import PRs).

mrobinson: Hrm. This happens when a test flakes (it's run twice, but has different results each time) and is the expected behavior during try runs. Now that filtering intermittents is turned on unconditionally, tests that fail are run twice and have their results appended to the log each time. I think this will make the generated results more reliable -- because tests that flake won't have their results updated by the flaky test run during the WPT import.

On the other hand, if a new test flakes (much less likely, but possible, no results will be generated for it). I guess in those cases, we will need to manually generate results to have the WPT import land. We have to do this anyway and since the amount of flaky tests is bigger than the intersection of flaky tests and new tests, this should hopefully make imports easier.

mrobinson: If this looks good, to you @sagudev, maybe we can land it today and see how the WPT import goes. If there is trouble, we can either back it out or fix the problem once we see how it goes. What do you think?

sagudev (approved):
I did some try runs and all seems to work.

.github/workflows/linux-wpt.yml
sagudev: > I think this will make the generated results more reliable -- because tests that flake won't have their results updated by the flaky test run during the WPT import

Will it really though? Because we currently feed full not filtered logs into wpt-update or am I missing something?

mrobinson: The way filtering mode works is that the full logs have data from all runs, but json files are produced in parallel that contain information about what tests are filtered by the intermittent dashboard. This json file doesn't affect the full log though.

sagudev:
Reports are empty now: https://github.com/servo/servo/actions/runs/7950107885/job/21702601400 due to failure in reporting: https://github.com/servo/servo/actions/runs/7950107885/job/21702596836#step:6:349

mrobinson:
> Reports are empty now: https://github.com/servo/servo/actions/runs/7950107885/job/21702601400 due to failure in reporting: https://github.com/servo/servo/actions/runs/7950107885/job/21702596836#step:6:349

Looks like the issue is that the test runner cannot connect to the intermittent dashboard for some reason:

https://github.com/servo/servo/actions/runs/7950107885/job/21702405700

Indeed it looks like the tracker is down: https://build.servo.org/intermittent-tracker/

mrobinson:
Thanks to @mukilan's great work, the intermittent dashboard should be active again. We're looking at how to make sure it doesn't go down again.

PR #31361, opened 2024-02-16
script: implement AbortController

syvb:
Adds basic support for `AbortController`/`AbortSignal`, and implements the `signal` parameter to `addEventListener` to support cancelling event listeners with `AbortSignal`.

This PR doesn't implement  `AbortSignal`  dependent signals or fetch cancellation.

---


tests/wpt/mozilla/tests/mozilla/interfaces.html
```diff
@@ -11,6 +11,8 @@
 
 // IMPORTANT: Do not change the list below without review from a DOM peer!
 test_interfaces([
+  "AbortController",
+  "AbortSignal",
```
syvb, line 15: This was failing with
> If this is failing: DANGER, are you sure you want to expose the new interface AbortController to all webpages as a property on the global? Do not make a change to this file without review from jdm or Ms2ger for that specific change!
@jdm @Ms2ger

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2268,7 +2268,8 @@ def removeWrapperAndNullableTypes(types):
                 parentName = descriptor.getParentName()
                 while parentName:
                     descriptor = descriptorProvider.getDescriptor(parentName)
-                    extras += [descriptor.path, descriptor.bindingPath]
+                    if current_name != descriptor.ifaceName:
+                        extras += [descriptor.path, descriptor.bindingPath]
```
syvb, line 2272: This fixes a bug in binding generation - `AbortSignal` both implements `EventTarget` and is part of `AddEventListenerOptions` in `EventTarget`. This resulted in the binding generation trying to make `EventTarget` import itself.

jdm:
Sorry about the delay here. I've started reviewing these changes and I'll try to finish it up later today!

jdm (approved):
I got sick in the meantime, so apologies for the additional delay. This looks good!

jdm (approved):

mrobinson:
It looks like this has modified the behavior of some tests that use the BFCache?

syvb:
I just updated the expectations so that the tests should pass now.

bors try

syvb:
I rebased the PR on `main`, it should pass tests now.

syvb:
That failure is flakey, I created an issue for it: https://github.com/servo/servo/issues/32213

sagudev:
@syvb, do you plan to continue hacking this, given that we revert this due to flakiness? Do you need any help?

syvb:
@sagudev I think the flakiness issue stems from the fact that I didn't implement AbortSignal for every DOM interface that accepts it. So e.g. passing an AbortSignal to `fetch` results in the test timing out or failing because `fetch` ignores the `signal` passed to it, instead of immediately failing when the code tried to create an AbortController. Fixing all of the test flakes requires implementing support for `AbortSignal`s in every DOM interface that accepts them.

I'm not very familiar with Servo's network stack and how request aborting could be integrated into that, which would be necessary to get AbortController working without any test flakes.

sagudev:
> AbortSignal to fetch results in the test timing out or failing because fetch ignores the signal passed to it, instead of immediately failing when the code tried to create an AbortController instead of immediately failing when the code tried to create an AbortController

This sounds like we could solve by fixing the tests, although I am not sure what would wpt feel about that.

> Fixing all of the test flakes requires implementing support for AbortSignals in every DOM interface that accepts them.

What if we just implement dummy abort controller intefaces?

> I'm not very familiar with Servo's network stack and how request aborting could be integrated into that, which would be necessary to get AbortController working without any test flakes.

There is https://github.com/servo/servo/blob/db4cba4d6d0f1a26009967d17ffdf30157a81b5e/components/script/fetch.rs#L55-L60
but I'm not very familiar with Servo's network stack either. cc @gterzian

gterzian:
> There is 

Yes that would be a place to start looking if you want to cancel fetches.

gterzian:
Was this one reverted after the merge?

PR #31365, opened 2024-02-16
selectors: Revert as many changes from upstream as possible

mrobinson:
This is part of the preparation for splitting stylo into a separate
crate. We have made various changes to selectors includings:

 1. Bumping the rust edition
 2. Fixing typos and updating links

In addition to reverting those changes, this PR pulls in some changes to
selectors we seem to have missed in the process of updates.

---

### DIFF
--- a/components/selectors/Cargo.toml
+++ b/components/selectors/Cargo.toml
@@ -9,7 +9,6 @@ readme = "README.md"
 keywords = ["css", "selectors"]
 license = "MPL-2.0"
 build = "build.rs"
-edition = "2018"
 
 [lib]
 name = "selectors"
--- a/components/selectors/README.md
+++ b/components/selectors/README.md
@@ -1,11 +1,13 @@
 rust-selectors
 ==============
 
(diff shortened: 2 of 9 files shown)


delan:
(retrying due to #31364)

mrobinson:
I think we'll need to keep the changes from https://github.com/servo/servo/commit/29f617056418eae88d396de82ddaf744766e1774 judging by the test results. I've added those back.

mrobinson:
I think this is ready for review. The one failure is due to the new test expectation that has now landed in main.

nicoburns:
Are the plans to re-apply (any of) these changes upstream?

mrobinson:
> Are the plans to re-apply (any of) these changes upstream?

We're trying to only revert changes that don't make sense to push upstream.  In this case most of the changes are due to rust edition updates that are better managed according to the needs of Gecko. The rest we hope to upstream once we fully sync.

delan (approved):

PR #31374, opened 2024-02-18
Check for XML and XMLS namespace  during 'locating a namespace'

KiChjang:
---


components/script/dom/element.rs
```diff
@@ -1190,23 +1190,35 @@ impl Element {
 
     // Element branch of https://dom.spec.whatwg.org/#locate-a-namespace
     pub fn locate_namespace(&self, prefix: Option<DOMString>) -> Namespace {
-        let prefix = prefix.map(String::from).map(LocalName::from);
+        let ns_prefix = prefix.clone().map(|s| Prefix::from(&*s));
+
+        // Step 1.
```
mrobinson, line 1195: Would it be possible to copy the text from the specification here. The step numbers often change and the spec language will give a clue as to the new numbers.

components/script/dom/element.rs
```diff
@@ -1190,23 +1190,35 @@ impl Element {
...
+        if ns_prefix == Some(namespace_prefix!("xml")) {
+            return ns!(xml);
+        }
+
+        // Step 2.
```
mrobinson, line 1200: The same here, please and at all the other numbers.

components/script/dom/node.rs
```diff
@@ -2399,7 +2399,7 @@ impl Node {
         }
     }
 
-    // https://dom.spec.whatwg.org/#locate-a-namespace
+    /// <https://dom.spec.whatwg.org/#locate-a-namespace>
```
mrobinson, line 2402: Thanks for fixing this!

components/script/dom/element.rs
```diff
@@ -1190,23 +1190,35 @@ impl Element {
 
     // Element branch of https://dom.spec.whatwg.org/#locate-a-namespace
     pub fn locate_namespace(&self, prefix: Option<DOMString>) -> Namespace {
-        let prefix = prefix.map(String::from).map(LocalName::from);
+        let ns_prefix = prefix.clone().map(|s| Prefix::from(&*s));
```
mrobinson, line 1193: I would go with `namespace_prefix` as `ns` is a bit ambigous.

```suggestion
        let namespace_prefix = prefix.clone().map(|s| Prefix::from(&*s));
```

mrobinson (approved):
Seems good to me. Do you mind making a couple minor changes before landing. Thank you!

mrobinson (approved):

PR #31375, opened 2024-02-18
WedIDL: bring dom/bindings/typedarray further in line with spec

### LINKED ISSUE #31319: WedIDL: bring dom/bindings/typedarray further in line with spec
As we are refactoring our use of JS typed arrays(for example in https://github.com/servo/servo/pull/31202), I realized what we are doing--moving away from raw JSObject pointers to "typed stuff", and wrapping operations into structs that hide the use of `jsapi`-- we are actually simply following [the spec](https://webidl.spec.whatwg.org/); there are also things we can do to come even closer to it:

1. Rename [`typedarrays.rs`](https://github.com/servo/servo/blob/c367309a8fc004e47a6d8c0528b82e9d60a5716a/components/script/dom/bindings/typedarrays.rs) into `buffer_source_types.rs`(or similar), so that is matches [`idl-buffer-source-types`](https://webidl.spec.whatwg.org/#idl-buffer-source-types).
2. Wrap [the internal JSObject](https://github.com/servo/servo/blob/c367309a8fc004e47a6d8c0528b82e9d60a5716a/components/script/dom/bindings/typedarrays.rs#L18) into an enum whose variants would match that of [`BufferSource`](https://webidl.spec.whatwg.org/#BufferSource). 
3. Add links to the various operations, for example [`create_typed_array`](https://github.com/servo/servo/blob/c367309a8fc004e47a6d8c0528b82e9d60a5716a/components/script/dom/bindings/typedarrays.rs#L129) matches https://webidl.spec.whatwg.org/#arraybufferview-create (and we should rename the function to reflect that).
4. And so on. I think if you look at the spec it is quite clear what we can rename and link to to better match it.

Taym95:
As we are refactoring our use of JS typed arrays(for example in https://github.com/servo/servo/pull/31202), we realized what we are doing--moving away from raw JSObject pointers to "typed stuff", and wrapping operations into structs that hide the use of jsapi-- we are actually simply following [the spec](https://webidl.spec.whatwg.org/).

In this PR I renamed `typedarray` to `buffer_source` and added doc links to align with spec.

- [x] These changes fix #31319

- [x] These changes do not require tests because there is not behavior changes.


components/script/dom/bindings/buffer_source_types.rs
```diff
@@ -0,0 +1,441 @@
...
+    BigInt64Array(Box<Heap<*mut JSObject>>),
+    BigUint64Array(Box<Heap<*mut JSObject>>),
+    Float32Array(Box<Heap<*mut JSObject>>),
+    Float64Array(Box<Heap<*mut JSObject>>),
+    DataView(Box<Heap<*mut JSObject>>),
+    ArrayBuffer(Box<Heap<*mut JSObject>>),
+    Default(Box<Heap<*mut JSObject>>),
+}
+
+pub struct HeapBufferSourceTypes<T> {
```
mrobinson, line 41: What does the `Types` here refer to? Maybe this should just be called `HeapBufferSource`?

mrobinson (commented):
This looks good to me, but I have a question about what one name means.

components/script/dom/bindings/buffer_source_types.rs
Taym95: I tried to follow spec blindly, indeed `Types` in not needed here

mrobinson: I would rename this to `HeapBufferSource` and this file to `buffersource` then to match the naming scheme we use for the bindings currently.

mrobinson (changes requested):

mrobinson (approved):

PR #31376, opened 2024-02-19
install libudev-dev required by gamepad API

### LINKED ISSUE #31373: Seems that the download link for Linux is missing
Hi all,
trying to download servo for Linux at
[Linux (X64)](https://download.servo.org/nightly/linux/servo-latest.tar.gz)
I receive a file not found error

There is some problem with downloads?

mukilan:
Ubuntu 20.04 does not have libudev-dev causing build failure when compiling libudev-sys, which is in-turn needed by gilrs-core. Similarly, nix build also needs the udev C library.

On Ubuntu 22.04, we don't see the build failures since the build dependency 'libgstreamer-plugins-base-1.0-dev' transitively pulls in libudev-dev.

Fixes #31373

Note: I'm not sure if other distros require similar fixes - this patch is just for Debian and Nix.

- [x] These changes fix #31373 partially 
- [x] These changes do not require tests because they fix build errors in CI job.

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -77,6 +77,8 @@ stdenv.mkDerivation (androidEnvironment // rec {
     taplo
     llvmPackages.bintools # provides lld
 
+    udev # Needed by libudev-sys for GamePad API.
+
     # Build utilities
     cmake dbus gcc git pkg-config which llvm perl yasm m4
     (python3.withPackages (ps: with ps; [virtualenv pip dbus]))
--- a/python/servo/platform/linux.py
+++ b/python/servo/platform/linux.py
@@ -35,7 +35,7 @@ APT_PKGS = [
     'libgstrtspserver-1.0-dev',
     'gstreamer1.0-tools',
     'libges-1.0-dev',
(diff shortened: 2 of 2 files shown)


etc/shell.nix
```diff
@@ -77,6 +77,8 @@ stdenv.mkDerivation (androidEnvironment // rec {
     taplo
     llvmPackages.bintools # provides lld
 
+    udev # needed by libudev-sys for GamePad API
```
mrobinson, line 80:
```suggestion
    udev # Needed by libudev-sys for GamePad API.
```

mrobinson (approved):
I wonder if we should guard the gamepad support by a feature and make it optional and turned off by default.

mukilan:
> I wonder if we should guard the gamepad support by a feature and make it optional and turned off by default.

I was wondering the same. Perhaps @gterzian and @msub2 have some thoughts?

msub2:
> I wonder if we should guard the gamepad support by a feature and make it optional and turned off by default.

I defer to the folks that have worked on Servo longer than I have. While I would enjoy it being part of the default build I would understand if doing that makes things easier to handle.

gterzian:
It could be an idea to hide gilrs behind a trait, and have a default no-op implement.

PR #31385, opened 2024-02-20
Gamepad: Align closer to spec and implement missing slots

msub2:
This follows up from the initial Gamepad implementation and addresses some TODOs around `gamepad.[[exposed]]` and `navigator.[[hasGamepadGesture]]`, as well as some missing spec steps. It also properly gates gamepad handling behind the `dom.gamepad.enabled` pref, as previously the `gamepadconnected` and `gamepaddisconnected` events would fire regardless.

This also allows one more WPT test to pass, `gamepad/not-fully-active`.

---


components/script/dom/globalscope.rs
```diff
@@ -3142,15 +3143,16 @@ impl GlobalScope {
...
+                    let gamepad_list = window.Navigator().gamepads();
                     let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
                     gamepad_list.add_if_not_exists(&gamepad_arr);
 
-                    // TODO: 3.4 If navigator.[[hasGamepadGesture]] is true:
-                    // TODO: 3.4.1 Set gamepad.[[exposed]] to true.
-
                     if window.Document().is_fully_active() {
-                        gamepad.update_connected(true);
+                        gamepad.update_connected(true, has_gesture);
```
Taym95, line 3155: here is `has_gesture ` and when you call function you use `should_notify ` I would expect prams name to be the same

components/script/dom/globalscope.rs
```diff
@@ -3211,8 +3214,26 @@ impl GlobalScope {
...
+                                        let new_gamepad = Trusted::new(&*gamepad);
+                                        if window.Document().is_fully_active() {
+                                            window.task_manager().gamepad_task_source().queue(
+                                                task!(update_gamepad_connect: move || {
+                                                    let gamepad = new_gamepad.root();
+                                                    gamepad.notify_event(GamepadEventType::Connected);
+                                                }),
+                                                window.upcast(),
+                                            )
+                                            .unwrap();
```
Taym95, line 3232: maybe you should check if gamepad_task_source is not `None`

components/script/dom/navigator.rs
```diff
@@ -171,11 +190,16 @@ impl NavigatorMethods for Navigator {
 
     // https://www.w3.org/TR/gamepad/#navigator-interface-extension
     fn GetGamepads(&self) -> DomRoot<GamepadList> {
-        let root = self
-            .gamepads
-            .or_init(|| GamepadList::new(&self.global(), &[]));
+        let global = self.global();
```
Taym95, line 193: I would expect some wpt tests for these changes

components/script/dom/gamepad.rs
```diff
@@ -265,4 +269,12 @@ impl Gamepad {
             warn!("Button bounds difference is either 0 or non-finite!");
         }
     }
+
+    pub fn exposed(&self) -> bool {
```
Taym95, line 273: I think we should add doc link when we add public API, not link for `exposed ` ?

components/script/dom/navigator.rs
msub2: The existing WPT that covers this fails on an assert that getGamepads() returns an array. Currently it returns a GamepadList, and attempting to change it to a frozen array of gamepads is non-trivial due to `to_frozen_array` seemingly being incompatible with the Gamepad object.

components/script/dom/gamepad.rs
msub2: Added

components/script/dom/navigator.rs
msub2: This is now covered by the aforementioned WPT test (`gamepad/not_fully_active.https.html`)

components/script/dom/navigator.rs
```diff
@@ -63,6 +67,7 @@ impl Navigator {
             permissions: Default::default(),
             mediasession: Default::default(),
             gpu: Default::default(),
+            has_gamepad_gesture: Cell::new(false),
```
gterzian, line 70: /// <https://w3c.github.io/gamepad/#dfn-hasgamepadgesture>

components/script/dom/navigator.rs
```diff
@@ -47,6 +50,7 @@ pub struct Navigator {
     permissions: MutNullableDom<Permissions>,
```
gterzian, line 50: Above /// <https://w3c.github.io/gamepad/#dfn-gamepads>

components/script/dom/navigator.rs
```diff
@@ -170,13 +190,18 @@ impl NavigatorMethods for Navigator {
     }
 
     // https://www.w3.org/TR/gamepad/#navigator-interface-extension
-    fn GetGamepads(&self) -> DomRoot<GamepadList> {
-        let root = self
-            .gamepads
-            .or_init(|| GamepadList::new(&self.global(), &[]));
+    fn GetGamepads(&self) -> Vec<Option<DomRoot<Gamepad>>> {
```
gterzian, line 193: https://w3c.github.io/gamepad/#dom-navigator-getgamepads

components/script/dom/navigator.rs
```diff
@@ -170,13 +190,18 @@ impl NavigatorMethods for Navigator {
...
+        let global = self.global();
+        let window = global.as_window();
+        let doc = window.Document();
+
+        if !doc.is_fully_active() || !self.has_gamepad_gesture.get() {
+            return Vec::new();
+        }
+
```
gterzian, line 201: TODO for permissions

components/script/dom/gamepadlist.rs
```diff
@@ -48,6 +48,15 @@ impl GamepadList {
     pub fn remove_gamepad(&self, index: usize) {
         self.list.borrow_mut().remove(index);
     }
+
+    pub fn list(&self) -> Vec<Option<DomRoot<Gamepad>>> {
+        self.list
+            .borrow()
+            .clone()
```
gterzian, line 55: No need to clone since you can build the list from references.

components/script/dom/gamepadlist.rs
```diff
@@ -48,6 +48,15 @@ impl GamepadList {
     pub fn remove_gamepad(&self, index: usize) {
         self.list.borrow_mut().remove(index);
     }
+
+    pub fn list(&self) -> Vec<Option<DomRoot<Gamepad>>> {
```
gterzian, line 52: I would say that in a follow-up we can remove gamepadlist, since a `RefCell<Vec<Dom<Gamepad>>>` on the navigator would seem to do the trick.

components/script/dom/globalscope.rs
```diff
@@ -3142,15 +3142,16 @@ impl GlobalScope {
...
+                        gamepad.set_exposed(true);
+                    }
+                    let gamepad_list = window.Navigator().gamepads();
                     let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
                     gamepad_list.add_if_not_exists(&gamepad_arr);
 
-                    // TODO: 3.4 If navigator.[[hasGamepadGesture]] is true:
-                    // TODO: 3.4.1 Set gamepad.[[exposed]] to true.
-
                     if window.Document().is_fully_active() {
```
gterzian, line 3153: this should be inside the  `has_gesture` block.

components/script/dom/globalscope.rs
gterzian: agree, arg name should be `has_gesture`.

components/script/dom/globalscope.rs
```diff
@@ -3211,8 +3213,26 @@ impl GlobalScope {
                                 }
                             };
 
-                            // TODO: 6. If navigator.[[hasGamepadGesture]] is false
-                            //          and gamepad contains a gamepad user gesture:
+                            if !window.Navigator().has_gamepad_gesture() {
+                                window.Navigator().set_has_gamepad_gesture(true);
```
gterzian, line 3217: We should only set it to true according to https://w3c.github.io/gamepad/#dfn-gamepad-user-gesture. Is this dealt with at the embedding layer? It might relate to what you experienced initially as the pad not responding to a first input.

components/script/dom/globalscope.rs
gterzian: This is about the queuing of the task, but we should expect with a clear message.

gterzian (changes requested):
Great start! Couple of comments...

components/script/dom/globalscope.rs
```diff
@@ -3211,8 +3213,26 @@ impl GlobalScope {
...
+                                        let new_gamepad = Trusted::new(&*gamepad);
+                                        if window.Document().is_fully_active() {
+                                            window.task_manager().gamepad_task_source().queue(
+                                                task!(update_gamepad_connect: move || {
+                                                    let gamepad = new_gamepad.root();
+                                                    gamepad.notify_event(GamepadEventType::Connected);
+                                                }),
+                                                window.upcast(),
+                                            )
+                                            .unwrap();
```
gterzian, line 3231: Let's also use `queue_with_canceller` for all tasks(here and others), and replace `unwrap` with an appropriate `expect`.

components/script/dom/globalscope.rs
msub2: I suppose it's a bit of a simplification as it is now, I'm treating any input as a gamepad user gesture but I should technically be making it more rigorous and checking for axis threshold too. This is unrelated to my initial issue with it not responding to first input though, that was caused by GilRs not emitting a changed event when a button doesn't have prior state (e.g. pressed for the very first time).

components/script/dom/gamepadlist.rs
msub2: Agreed, I'll probably do that after this and before GamepadHapticActuator

components/script/dom/globalscope.rs
```diff
@@ -3193,32 +3194,54 @@ impl GlobalScope {
...
                     let global = this.root();
                     if let Some(window) = global.downcast::<Window>() {
-                        let gamepad_list = window.Navigator().GetGamepads();
-                        if let Some(gamepad) = gamepad_list.IndexedGetter(index as u32) {
+                        let gamepad_list = window.Navigator().gamepads();
+                        if let Some(gamepad) = gamepad_list.Item(index as u32) {
                             let current_time = global.performance().Now();
                             gamepad.update_timestamp(*current_time);
 
+                            let mut contains_user_gesture = false;
```
gterzian, line 3206: let's put this in a separate function taking the update type as argument and with a doclink to https://w3c.github.io/gamepad/#dfn-gamepad-user-gesture

gterzian (approved):
LGTM with one nit.

components/script/dom/globalscope.rs
```diff
@@ -3244,6 +3240,18 @@ impl GlobalScope {
             .expect("Failed to queue update gamepad state task.");
     }
 
+    /// <https://www.w3.org/TR/gamepad/#dfn-gamepad-user-gesture>
+    pub fn contains_user_gesture(&self, update_type: GamepadUpdateType) -> bool {
```
gterzian, line 3244: Is clippy not complaining about not using `self`? 

I think this should be a function(not method), inside `gamepad`. Because now it reads as if it's the global that "contains a user gesture" or not. 

also that way you only need to import that function, and the constants are used only in `gamepad`.

gterzian (changes requested):
Ok thanks for the refactor, just one last thing and then good to go!

components/script/dom/gamepad.rs
```diff
@@ -263,4 +271,26 @@ impl Gamepad {
...
+pub fn contains_user_gesture(update_type: GamepadUpdateType) -> bool {
+    match update_type {
+        GamepadUpdateType::Axis(_, value) => {
+            return value.abs() > AXIS_TILT_THRESHOLD;
+        },
+        GamepadUpdateType::Button(_, value) => {
+            return value > BUTTON_PRESS_THRESHOLD;
+        },
+    };
 }
```
gterzian, line 296: Let's just add a newline here.

gterzian (approved):
LGTM, one nit.

components/script/dom/gamepad.rs
msub2: A second one? It ends at 297

gterzian: Ah ok, should be fine then...

components/script/dom/gamepad.rs
```diff
@@ -23,9 +24,12 @@ use crate::dom::gamepadpose::GamepadPose;
 use crate::dom::globalscope::GlobalScope;
 use crate::script_runtime::JSContext;
 
+// This value is for determining when to consider a gamepad as having a user gesture
+// from an axis tilt. This matches the threshold in Chromium.
+pub const AXIS_TILT_THRESHOLD: f64 = 0.5;
```
gterzian, line 29: One more thing: this can be private to the module.

gterzian (approved):
Thank you!

PR #31388, opened 2024-02-20
Support <div align="..."> and <center> in inline layout

Loirooriol:
As per HTML \[1\], `<div align="...">` and `<center>` should behave as if they had the text-align property set to the corresponding value.

Servo implements that as internal `text-align` values because there should the extra effect of aligning block descendants, but that part has not been implemented yet. This patch only adds support for inline layout.

\[1\]: https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3

---


mrobinson:
> Should I add a more specific test?

I think if tests start passing when the full feature is implemented, it's good enough. Thanks!

mrobinson (approved):

Issue #30862, opened 2023-12-12
In components/script, remove unsafe JS related code


gterzian:
In [script](https://github.com/servo/servo/tree/48a95b2471c590a0ee1b85710fbb5a37ee8215c5/components/script), you can find `unsafe` blocks peppered around, most(all?) of which relate to integration with the [JS engine](https://github.com/servo/mozjs/). 

Integration takes place either via the two crates of [`mozjs`](https://github.com/servo/mozjs), one of which is used as `js::rust` and is safe, the other is used as `js::jsapi` and is lower-level and unsafe. 

There are two things we should do:

1. Remove direct use of `js::jsapi` in `components/script`, outside of that found in `dom/bindings`.
2. Remove direct use of `js::jsapi`, by expending the collection of safe wrappers found in `js::rust`, and forbid unsafe code in `components/script`. 
3. Remove direct use of `js::rust`, and replace it with our own Servo-specific abstraction all to be found in `dom/bindings`. 

The goal of 1 and 2 is to remove unsafe code, either by expanding safe wrappers in `js::rust` or in `dom/bindings` while the goal of 3 is to make the use of the JS engine more modular. While 1 and 2 can be done in parallel, 3 should only be done after 1 and 2 are finished(I think). 

Here is one example:

In [`AudioBuffer`](https://github.com/servo/servo/blob/48a95b2471c590a0ee1b85710fbb5a37ee8215c5/components/script/dom/audiobuffer.rs#L45C12-L45C23), we [use](https://github.com/servo/servo/blob/48a95b2471c590a0ee1b85710fbb5a37ee8215c5/components/script/dom/audiobuffer.rs#L9) `js::jsapi::Heap` directly, because there is no safe or higher-level equivalent defined in `js::rust`(see for example safe wrappers found in [`mozjs/src/gc/root.rs`](https://github.com/servo/mozjs/blob/01d777e14e17b4a7994f5f8b7ca262e56bb6faf8/mozjs/src/gc/root.rs#L85)

An unsafe call of [`Heap::handle`](https://github.com/servo/mozjs/blob/01d777e14e17b4a7994f5f8b7ca262e56bb6faf8/mozjs-sys/src/jsgc.rs#L301) is found [here](https://github.com/servo/servo/blob/48a95b2471c590a0ee1b85710fbb5a37ee8215c5/components/script/dom/audiobuffer.rs#L212), it is necessary because of the use of the unsafe `JS_GetArrayBufferViewBuffer`.

[...]

PR #31398, opened 2024-02-21
WebIDL Fix ImageData constructor to take a Uint8ClampedArray instead of js_object

### LINKED ISSUE #31320: WebIDL: Fix ImageData constructor to take a Uint8ClampedArray instead of object
https://github.com/servo/servo/blob/c367309a8fc004e47a6d8c0528b82e9d60a5716a/components/script/dom/webidls/CanvasRenderingContext2D.webidl#L257

This may require fiddling with `components/script/dom/bindings/codegen/CodegenRust.py`.

Taym95:
Fix ImageData constructor to take a Uint8ClampedArray instead of js_object

---

- [X] These changes fix #31320

- [x] These changes do not require tests because there is not behavior change.

### DIFF
--- a/components/script/dom/bindings/buffer_source.rs
+++ b/components/script/dom/bindings/buffer_source.rs
@@ -37,7 +37,7 @@ use js::rust::{
 #[cfg(feature = "webgpu")]
 use js::typedarray::{ArrayBuffer, HeapArrayBuffer};
 use js::typedarray::{
-    ArrayBufferU8, ArrayBufferView, ArrayBufferViewU8, CreateWith, TypedArray, TypedArrayElement,
+    ArrayBufferU8, ArrayBufferViewU8, CreateWith, TypedArray, TypedArrayElement,
     TypedArrayElementCreator,
 };
 
@@ -63,36 +63,25 @@ pub(crate) enum BufferSource {
     ArrayBuffer(Box<Heap<*mut JSObject>>),
 }
 
-pub(crate) fn new_initialized_heap_buffer_source<T>(
-    init: HeapTypedArrayInit,
+pub(crate) fn create_heap_buffer_source_with_length<T>(
+    cx: JSContext,
+    len: u32,
     can_gc: CanGc,
-) -> Result<HeapBufferSource<T>, ()>
+) -> Fallible<HeapBufferSource<T>>
 where
     T: TypedArrayElement + TypedArrayElementCreator,
     T::Element: Clone + Copy,
 {
-    let heap_buffer_source = match init {
-        HeapTypedArrayInit::Buffer(buffer_source) => HeapBufferSource {
-            buffer_source,
-            phantom: PhantomData,
-        },
-        HeapTypedArrayInit::Info { len, cx } => {
-            rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
-            let typed_array_result =
-                create_buffer_source_with_length::<T>(cx, len as usize, array.handle_mut(), can_gc);
-            if typed_array_result.is_err() {
-                return Err(());
-            }
+    rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
+    let typed_array_result =
+        create_buffer_source_with_length::<T>(cx, len as usize, array.handle_mut(), can_gc);
+    if typed_array_result.is_err() {
+        return Err(Error::JSFailed);
+    }
 
-            HeapBufferSource::<T>::new(BufferSource::ArrayBufferView(Heap::boxed(*array.handle())))
-        },
-    };
-    Ok(heap_buffer_source)
-}
-
-pub(crate) enum HeapTypedArrayInit {
-    Buffer(BufferSource),
-    Info { len: u32, cx: JSContext },
+    Ok(HeapBufferSource::<T>::new(BufferSource::ArrayBufferView(
+        Heap::boxed(*array.handle()),
+    )))
 }
 
 pub(crate) struct HeapBufferSource<T> {
--- a/components/script/dom/bindings/buffer_source.rs
+++ b/components/script/dom/bindings/buffer_source.rs
@@ -131,11 +131,11 @@ where
     }
 
     pub(crate) fn from_view(
-        chunk: CustomAutoRooterGuard<ArrayBufferView>,
-    ) -> HeapBufferSource<ArrayBufferViewU8> {
-        HeapBufferSource::<ArrayBufferViewU8>::new(BufferSource::ArrayBufferView(Heap::boxed(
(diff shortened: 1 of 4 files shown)


Taym95:
@gterzian Do you have any idea what to do with the PR? After this change, tests are failing for all browsers, including Servo?

mrobinson:
@Taym95 What's the status of this PR?

gterzian:
@Taym95 What is the failing test? If all browsers are failing, it may be because the test itself is not conformant.

jdm:
https://wpt.fyi/results/html/semantics/embedded-content/the-canvas-element/imagedata.html?label=master&label=experimental&aligned&q=imagedata.html

components/script/dom/imagedata.rs
```diff
@@ -33,42 +34,54 @@ pub struct ImageData {
...
     ) -> Fallible<DomRoot<ImageData>> {
-        let heap_typed_array = match new_initialized_heap_typed_array::<ClampedU8>(
-            HeapTypedArrayInit::Object(jsobject),
-        ) {
-            Ok(heap_typed_array) => heap_typed_array,
-            Err(_) => return Err(Error::JSFailed),
-        };
+        let cx = GlobalScope::get_cx();
+        let heap_typed_array =
+            match new_initialized_heap_typed_array::<ClampedU8>(HeapTypedArrayInit::Data {
```
gterzian, line 78: I think this needs a rebase, because the api was updated.

components/script/dom/imagedata.rs
```diff
@@ -33,42 +34,54 @@ pub struct ImageData {
...
+                data: unsafe { data.as_slice() },
+                cx: cx,
+            }) {
+                Ok(heap_typed_array) => heap_typed_array,
+                Err(_) => return Err(Error::JSFailed),
```
gterzian, line 83: Either we are failing here where previously we wouldn't. Maybe try returning a type error here?

Or we are not producing a type error below at https://github.com/servo/servo/pull/31398/files#diff-74ee6b1aa263f228fb8648bb921e80b79654506e8b39570b764649a02f881402R89, don't know why. Maybe `get_buffer` is too liberal in returning an object even when it's of the wrong type?

gterzian (commented):
I think we should find a way to make the test pass.

components/script_bindings/webidls/CanvasRenderingContext2D.webidl
```diff
@@ -254,7 +254,7 @@ interface CanvasPattern {
  Serializable]
 interface ImageData {
   [Throws] constructor(unsigned long sw, unsigned long sh/*, optional ImageDataSettings settings = {}*/);
-  [Throws] constructor(/* Uint8ClampedArray */ object data, unsigned long sw, optional unsigned long sh
+  [Throws] constructor(Uint8ClampedArray data, unsigned long sw, optional unsigned long sh
```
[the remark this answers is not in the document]
jdm: The specification is dense, but I think the test as written does not match the specification. From https://webidl.spec.whatwg.org/#js-overloads we can determine which overloaded constructor to call. The distinguishing arg is 0, so we start at step 12. 12.7 checks whether there is an overload with a typed array argument that matches the kind of the typed array that was passed in. There is not, so we continue to step 12.16, which checks if there is an overload with a number at the argument index (ignoring the actual value passed in at this point). There is, so that is the overload that is selected. We only passed this test previously because we reached step 12.11 and the argument was specified as an object, so we selected the overload with the typed array argument instead.

In short: the test is wrong, it is correct that we (and every other browser) fail it.

mrobinson:
I don't even see a backtrace on this crash, which suggests it's a segfault and not a Rust panic:

```
  ▶ CRASH [expected OK] /css/compositing/canvas-composite-modes.html
  │ 
  └ Caught signal 11 in thread "Script(1,1)"
```

components/script/dom/bindings/buffer_source.rs
```diff
@@ -63,36 +63,25 @@ pub(crate) enum BufferSource {
...
-            HeapBufferSource::<T>::new(BufferSource::ArrayBufferView(Heap::boxed(*array.handle())))
-        },
-    };
-    Ok(heap_buffer_source)
-}
+    rooted!(in (*cx) let mut array = ptr::null_mut::<JSObject>());
+    let typed_array_result =
+        create_buffer_source_with_length::<T>(cx, len as usize, array.handle_mut(), can_gc);
+    if typed_array_result.is_err() {
+        return Err(Error::JSFailed);
```
gterzian, line 79: Are you sure about the `Error::JSFailed`, previously it was just a `()`. Also the error is mapped to `JSFailed` a thttps://github.com/servo/servo/pull/31398/files#diff-74ee6b1aa263f228fb8648bb921e80b79654506e8b39570b764649a02f881402R61

components/script/dom/imagedata.rs
```diff
@@ -55,31 +55,37 @@ impl ImageData {
...
+
                 let data = CreateWith::Slice(&d[..]);
                 Uint8ClampedArray::create(*cx, data, js_object.handle_mut()).unwrap();
-                Self::new_with_jsobject(global, None, width, Some(height), js_object.get(), can_gc)
+                Self::new_with_data(
+                    global,
+                    None,
+                    width,
+                    Some(height),
+                    CustomAutoRooterGuard::new(*cx, &mut CustomAutoRooter::new(typed_array)),
```
gterzian, line 70: I think you want to assign this to a variable and then call the function. May be the source of the crash actually.

gterzian (changes requested):
looks good overall, but we should find out where that crash is coming from.

components/script/dom/imagedata.rs
```diff
@@ -55,31 +55,32 @@ impl ImageData {
...
+
+                let typed_array =
+                    create_buffer_source::<ClampedU8>(cx, &d[..], js_object.handle_mut(), can_gc)
+                        .map_err(|_| Error::JSFailed)?;
+
                 let data = CreateWith::Slice(&d[..]);
                 Uint8ClampedArray::create(*cx, data, js_object.handle_mut()).unwrap();
-                Self::new_with_jsobject(global, None, width, Some(height), js_object.get(), can_gc)
+                let mut rooted_typed_array = CustomAutoRooter::new(typed_array);
+                let data = CustomAutoRooterGuard::new(*cx, &mut rooted_typed_array);
```
jdm, line 66: Use the `auto_root!` macro for this instead:
```rust
auto_root!(in(*cx) let data = typed_array);
```

gterzian (approved):
LGTM. Did we ascertain what caused the crash?

Also, let's track this as part of https://github.com/servo/servo/issues/30862

PR #31418, opened 2024-02-23
layout: Place statically positioned absolutes in IFCs at their hypothetical box position

### LINKED ISSUE #19879: Icons rotate incorrectly
URL: http://hn.premii.com/#

The `bubble` icons don't rotate correctly in Servo.

image

Note: Servo is built with bc7dd644f6

Tested in `macOS 10.12.6`

mrobinson:
Absolutes need to be placed at their hypothetical position as if the
`position` value was static. This position differs based on the value they
had before blockification. The code for placing absolutes was taking
into account the original display for the inline value, but not for the
block value. A static `display: block` box would placed at a new block
position past the end of the linebox.

---

- [x] These changes fix #19879.

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -89,6 +89,7 @@ where
     }
 }
 
+#[derive(Debug)]
 pub(super) enum Contents {
     /// Refers to a DOM subtree, plus `::before` and `::after` pseudo-elements.
     OfElement,
@@ -107,6 +108,7 @@ pub(super) enum NonReplacedContents {
     OfPseudoElement(Vec<PseudoElementContentItem>),
 }
 
+#[derive(Debug)]
 pub(super) enum PseudoElementContentItem {
     Text(String),
     Replaced(ReplacedContent),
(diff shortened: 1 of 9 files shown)


components/layout_2020/dom_traversal.rs
```diff
@@ -89,6 +89,7 @@ where
     }
 }
 
+#[derive(Debug)]
```
mrobinson, line 92: I've also added `Debug` support to some classes which should help to debug these issues in the future.

components/layout_2020/flow/line.rs
```diff
@@ -523,31 +522,34 @@ impl AbsolutelyPositionedLineItem {
...
-            },
-            Display::Contents => {
-                panic!("display:contents does not generate an abspos box")
-            },
-            Display::None => {
-                panic!("display:none does not generate an abspos box")
-            },
-        };
+            } else {
+                // After the bottom of the line at the start of the inline formatting context.
```
Loirooriol, line 546: I was thinking whether it could appear at a different position, e.g. if there is a float and the abspos would establish an independent formatting context even with `position: static`, or if it has `clear: both`.
But other browsers don't seem to do this, so looks good.

Loirooriol (approved):

PR #31419, opened 2024-02-23
layout: Add line height from preserved segment breaks in quirks mode

mrobinson:
In quirks mode, preserved segment breaks should add line height to
lines. This matches the behavior of WebKit and Blink, but not Gecko.

This also handles the special-case of `<br>` elements, which are
implemented with preserved segment breaks via `white-space: pre-line`.
This is an implementation detail though because `<br>` has a special
behavior if the line isn't empty -- it doesn't add any line height in
this case.

---

### DIFF
--- a/tests/wpt/tests/quirks/line-height-preserved-segment-break.html
+++ b/tests/wpt/tests/quirks/line-height-preserved-segment-break.html
@@ -1,0 +1,24 @@
+<!-- quirks mode -->
+<link rel="author" title="Martin Robinson" href="mrobinson@igalia.com">
+<link rel="help" href="https://quirks.spec.whatwg.org/#the-line-height-calculation-quirk">
+<link rel="help" href="https://drafts.csswg.org/css-text/#valdef-white-space-pre-line">
+<link rel="match" href="../css/reference/ref-filled-green-100px-square.xht">
+
+<style>
+    .container {
+        width: 100px;
+        background: green;
+        white-space: pre-line;
+    }
+
+    .inline-box {
+        line-height: 100px;
+    }
+</style>
+
+<p>Test passes if there is a filled green square and <strong>no red</strong>.</p>
+<div style="width: 100px; height: 100px; background: red;">
+    <!-- The preserved segment break in the span should mean that "The line height calculation
+    quirk" does not apply. -->
+  <div class="container"><span class="inline-box">&#xA;</span></div>
+</div>
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -4,7 +4,7 @@
 
 use std::borrow::Cow;
 
-use html5ever::LocalName;
+use html5ever::{local_name, LocalName};
 use script_layout_interface::wrapper_traits::{ThreadSafeLayoutElement, ThreadSafeLayoutNode};
 use servo_arc::Arc as ServoArc;
 use style::properties::ComputedValues;
@@ -78,6 +78,9 @@ where
         let flags = match threadsafe_node.as_element() {
             Some(element) if element.is_body_element_of_html_element_root() => {
                 FragmentFlags::IS_BODY_ELEMENT_OF_HTML_ELEMENT_ROOT
+            },
+            Some(element) if element.get_local_name() == &local_name!("br") => {
+                FragmentFlags::IS_BR_ELEMENT
             },
             _ => FragmentFlags::empty(),
         };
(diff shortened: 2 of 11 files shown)


mrobinson:
This needs a bit more work. In quirks mode forced line breaks do not add line-height if the line is otherwise empty, but `<br>` elements do! I think it is finally the moment to add "magic" behavior for `<br>`.

Loirooriol (changes requested):
The spec says

> It either contains no text or it contains only collapsed whitespace.

So a forced line break doesn't fulfill the condition, right? I'm testing:

```html
a
<div style="white-space: pre">&#xa;&#xa;&#xa;</div>
b
<div style="white-space: pre"><span>&#xa;&#xa;&#xa;</span></div>
c
<div><span style="white-space: pre">&#xa;&#xa;&#xa;</span></div>
d
```

I think Blink and WebKit are correct: all letters are separated.
Firefox seems to have a bug: c and d are together.
And Servo is worse: all letters are together.

So I think:
 - We shouldn't add `<br>` magic here.
 - We align Servo with Blink/WebKit.
 - We should add a WPT that fails in Firefox.

mrobinson:
There is still going to be some special behavior though, because `<br>` doesn't always add its line-height to the line. See how only the `<br>` on the empty line adds the full font size to line height in the following quirks mode examples:

```html
<!-- No br on empty line, so lines render as if font-size: 20px -->
<div style="font-size: 100px; border: solid;">
    <span style="font-size: 20px">Hello</span><br>
    <span style="font-size: 20px">Hello2</span><br>
</div>

<!-- br on empty line, so middle line renders as if font-size: 100px -->
<div style="font-size: 100px; border: solid;">
    <span style="font-size: 20px">Hello</span><br>
    <br>
    <span style="font-size: 20px">Hello2</span><br>
</div>

<!-- The same as the first example, this renders differently than with br in all browsers -->
<div style="font-size: 100px; border: solid; white-space: pre-line"><span style="font-size: 20px">Hello</span><br>
    <span style="font-size: 20px">Hello2</span><br>
</div>

```

mrobinson:
Okay. I've modified the change to always add the strut size for preserved segment breaks, except in the case for `<br>` which shouldn't do this when the line isn't empty. I've also added a test to verify this for all browsers. This test fails only in Gecko.

mrobinson:
> Possibly may be left for a follow-up, but note this differs from Gecko/Blink/WebKit:
> 
> ```
> a<div style="font-size: 0">x<span style="font-size: 16px"> </span>x</div>b
> ```

Oh yeah, that's an issue. I think it's covered by this todo from `inline.rs`:

```rust
            // TODO(mrobinson): This isn't quite right for situations where collapsible white space
            // ultimately does not collapse because it is between two other pieces of content.
```

tests/wpt/tests/quirks/line-height-preserved-segment-break.html
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    .inline-box {
+        line-height: 100px;
+    }
+</style>
+
+<!-- This div is just to make sure that red shows up if the test div doesn't
+     display properly. -->
+<div style="position: absolute; z-index: -100; width: 100px; height: 100px; background: red;"></div>
```
Loirooriol, line 22: Fine either way, but you could turn this into a wrapper to avoid `position: absolute; z-index: -100`:

```html
<div style="width: 100px; height: 100px; background: red;">
  <div class="container"><span class="inline-box">&#xA;</span></div>
</div>
```

I also think the HTML looks slightly visually clearer with `&#xA;` rather than starting one line with `<div>` and next line with `</span>`. But again, not a big deal.

tests/wpt/tests/quirks/line-height-preserved-segment-break.html
```diff
@@ -0,0 +1,27 @@
+<!-- quirks mode -->
+<link rel="author" title="Martin Robinson" href="mrobinson@igalia.com">
+<link rel="help" href="https://quirks.spec.whatwg.org/#the-line-height-calculation-quirk">
+<link rel="help" href="https://drafts.csswg.org/css-text/#valdef-white-space-pre-line">
+<link rel="match" href="line-height-preserved-segment-break-ref.html">
```
Loirooriol, line 5: Maybe use `/css/reference/ref-filled-green-100px-square.xht`? Or do you want the ref to be inside `/quirks/`?

tests/wpt/tests/quirks/line-height-preserved-segment-break.html
```diff
@@ -0,0 +1,27 @@
...
+<p>Test passes if there is a filled green square and <strong>no red</strong>.</p>
+
+<style>
```
Loirooriol, line 8: You can move the `<style>` above so that it goes into the `<head>` instead of the `<body>`, which is invalid HTML.

Loirooriol (approved):

mrobinson:
Thanks for the review, @Loirooriol. I've made the changes you suggested.

PR #31348, opened 2024-02-14
Bury failed nightly releases as prereleases

delan:
On our [nightly builds page](https://github.com/servo/servo-nightly-builds/releases), GitHub lists drafts before normal releases, and normal releases before prereleases. But when our release build fails, we leave the release for that day as a draft, which clogs up the list for anyone who can see drafts.

This patch makes the build publish the release as a prerelease on failure, which buries them under successful releases while preserving the downloads for any platforms that succeeded.

I’ve also [added a script](https://github.com/servo/servo-nightly-builds/commit/5044258532f9113d02a3b38ec64c5a975ee445e8) to the nightly builds repo that cleans up existing failed releases:

```
$ tools/bury-old-drafts.sh
Warning: ignoring release 141927042, which is only 691 seconds old
Page 1 has 100 releases; found 56 drafts so far
Page 2 has 100 releases; found 56 drafts so far
Page 3 has 100 releases; found 56 drafts so far
Page 4 has 14 releases; found 56 drafts so far
PATCH /repos/servo/servo-nightly-builds/releases/140026516 -F prerelease=true -F draft=false
PATCH /repos/servo/servo-nightly-builds/releases/136085696 -F prerelease=true -F draft=false
PATCH /repos/servo/servo-nightly-builds/releases/136047173 -F prerelease=true -F draft=false
[...]
PATCH /repos/servo/servo-nightly-builds/releases/135079699 -F prerelease=true -F draft=false
gh: Validation Failed (HTTP 422)
{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "Release",
      "code": "already_exists",
      "field": "tag_name"
    }
  ],
  "documentation_url": "https://docs.github.com/rest/releases/releases#update-a-release"
}
Delete release? [y/N] y
DELETE /repos/servo/servo-nightly-builds/releases/135079699
[...]
```

<details><summary>Screenshot of releases page sorting behaviour</summary>

image</details>

<details><summary>Screenshot of nightly builds page before cleanup</summary>

image</details>

<details><summary>Screenshot of nightly builds page after cleanup</summary>

(I have no idea why there are three old builds at the top when logged in)

image</details>

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] ~~These changes fix #___ (GitHub issue number if applicable)~~

- [ ] There are tests for these changes OR
- [x] These changes do not require tests because they only affect the nightly release build on CI

PR #31420, opened 2024-02-23
ci: Fix nightly release action

### LINKED ISSUE #31373: Seems that the download link for Linux is missing

mukilan:
As observed in #31373, the changes introduced by #31348 don't
seem to work as expected. The workflow [always executes](https://github.com/servo/servo/actions/runs/7954978996/job/21714145504) the step
corresponding to `if: success()` despite the failure in one of the
prerequisite jobs specified under 'needs'.

This seems to be a quirk of how GH Actions work. Checking the
status of the prerequisite jobs using `contains(needs.*.result)` 
seems to work correctly as we would expect.

- [x] These changes fix #31373
- [x] These changes do not require tests because they fix a bug in CI configuration.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,7 +15,7 @@ jobs:
     # This job is only useful when run on upstream servo.
     if: github.repository == 'servo/servo' || github.event_name == 'workflow_dispatch'
     name: Create Draft GH Release
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

delan (approved):
Great catch, thanks!

PR #31421, opened 2024-02-23
layout: Add initial support for row height distribution

mrobinson:
This change adds a version of row height distribution that follows the
distribution algorithm used for tables in Blink's LayoutNG. This is just
an intermediate step toward implementing a distribution algorithm for
both rows and columns more similar to Layout NG.
    
The CSS Table 3 specification is often wrong with regard to web
compatibility, which is why we have abandoned it in favor of the Layout
NG algorithm for row height distribution.
    
---


mrobinson:
I'm converting this one back into a draft. I think we should simply implement the LayoutNG distribution algorithm outright to avoid the code churn.

components/layout_2020/table/layout.rs
```diff
@@ -918,88 +1007,356 @@ impl<'a> TableLayout<'a> {
...
+
+            // Otherwise, if we only have empty rows or if all the non-empty rows are constrained,
+            // then grow the empty rows.
+            if only_have_empty_rows || non_empty_rows_all_constrained {
+                // If there are both unconstrained and constrained empty rows, only increase the
+                // size of the unconstrained ones, otherwise increase the size of all empty rows.
+                let mut rows_to_grow = &empty_rows;
+                let unconstrained_empty_rows: Vec<usize> = rows_to_grow
+                    .iter()
+                    .map(|track| *track)
```
Loirooriol, line 1289:
```suggestion
                    .copied()
```

components/layout_2020/table/layout.rs
```diff
@@ -918,88 +1007,356 @@ impl<'a> TableLayout<'a> {
...
+        };
+
+        // If we have a table height (not during rowspan distribution), first distribute to rows
+        // that have percentage sizes proportionally to the size missing to reach the percentage
+        // of table height required.
+        if let Some(percentage_resolution_size) = percentage_resolution_size {
+            let get_percent_block_size_deficit = |row_index: usize, track_size: Au| {
+                let size_needed_for_percent = percentage_resolution_size
+                    .scale_by(self.rows[row_index].percent.0 as f32 / 100.);
+                size_needed_for_percent - track_size.max(Au::zero())
```
Loirooriol, line 1208: Should this be?

```suggestion
                (size_needed_for_percent - track_size).max(Au::zero())
```

Or possibly leave it to the callers, since one is already checking `> Au::zero()` and doesn't need `.max(Au::zero())`

components/layout_2020/table/layout.rs
```diff
@@ -918,88 +1007,356 @@ impl<'a> TableLayout<'a> {
...
+            if percent_distributable_block_size > Au::zero() {
+                for track_index in track_range.clone() {
+                    let row_deficit =
+                        get_percent_block_size_deficit(track_index, track_sizes[track_index]);
+                    if row_deficit > Au::zero() {
+                        let ratio =
+                            row_deficit.to_f32_px() / percent_block_size_deficit.to_f32_px();
+                        let size = percent_distributable_block_size.scale_by(ratio);
+                        track_sizes[track_index] += size;
+                        excess_size -= size;
```
Loirooriol, line 1224: Can this make `excess_size` negative?

components/layout_2020/table/layout.rs
```diff
@@ -918,88 +1007,356 @@ impl<'a> TableLayout<'a> {
...
+        containing_block: &ContainingBlock,
+    ) {
+        // The table content height is the maximum of the computed table height from style and the
+        // sum of computed row heights from row layout plus size from borders and spacing.
+        let table_height_from_style = self
+            .table
+            .style
+            .content_box_size(containing_block, &self.pbm)
+            .auto_is(Length::zero)
+            .block
```
Loirooriol, line 1330: No need to resolve both dimensions:

```suggestion
            .block
            .auto_is(Length::zero)
```

Also this is not taking min/max-height into account, but we can leave it for the follow-up.

components/layout_2020/table/layout.rs
```diff
@@ -306,11 +399,11 @@ impl<'a> TableLayout<'a> {
         for column_index in 0..self.table.size.width {
             let column_measure = &mut column_measures[column_index];
             let final_intrinsic_percentage_width = column_measure
-                .percentage_width
+                .percentage
                 .0
                 .min(100. - total_intrinsic_percentage_width);
```
Loirooriol, line 404: Just realized that `Percentage` is actually 1-based, but we can fix them in a follow-up.

Loirooriol (approved):

components/layout_2020/table/layout.rs
mrobinson: I don't think so. We calculate the ratio via (row_deficit / total_deficit), but this ratio is multiplied by total_deficit.min(excess_size), so we always take a ratio of excess size.

mrobinson (commented):
Thanks for the review @Loirooriol. I think I've addressed all the issues you raised.

components/layout_2020/table/layout.rs
mrobinson: I fixed this one. I think we do need this in case the sum of the percentages of the rows is greater than 100%.

PR #31423, opened 2024-02-23
Support <div align="..."> and <center> in block layout

Loirooriol:
As per HTML \[1\], `<div align="...">` and `<center>` should align descendants to the left/center/right. This is similar to having 'auto' margins \[2\], but without changing their used values \[3\].

\[1\]: https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3
\[2\]: https://html.spec.whatwg.org/multipage/rendering.html#align-descendants
\[3\]: https://github.com/whatwg/html/issues/10149

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -13,7 +13,7 @@ use style::computed_values::float::T as Float;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
 use style::values::computed::{Length, LengthOrAuto};
-use style::values::specified::Display;
+use style::values::specified::{Display, TextAlignKeyword};
 use style::Zero;
 
 use crate::cell::ArcRefCell;
@@ -627,6 +627,7 @@ fn layout_in_flow_non_replaced_block_level_same_formatting_context(
         min_box_size,
         max_box_size,
         margin,
+        effective_margin_inline_start,
     } = solve_containing_block_padding_border_and_margin_for_in_flow_box(containing_block, style);
 
     let computed_block_size = style.content_block_size();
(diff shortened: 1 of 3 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -1324,25 +1366,27 @@ fn solve_inline_margins_for_in_flow_block_level(
         },
         (LengthOrAuto::LengthPercentage(start), LengthOrAuto::Auto) => (start, free_space - start),
         (LengthOrAuto::LengthPercentage(start), LengthOrAuto::LengthPercentage(end)) => {
+            let free_space = Length::zero().max(free_space - start - end);
+            justification = justify_self_alignment(containing_block, free_space);
```
mrobinson, line 1370: Does setting a margin on one side of a block take precedence over `<center>`? It might be worth leaving a comment to that effect here.

mrobinson (approved):
Nice! This is really not a lot of code in the end. I'm a bit surprised that there are only two tests passing with support for `<center>` added.

components/layout_2020/flow/mod.rs
Loirooriol: It doesn't take precedence, both work. When not avoiding floats, this centers the border box within the margin-shrunken containing block (equivalent to centering the margin box within the containing block). This matches other browsers and the spec. But of course, `auto` margins take all the free space, so no centering is possible.

When avoiding floats, this centers the border box within the intersection of the margin-shrunken containing block and the float-shrunken containing block. This matches Firefox. The spec is wrong (https://github.com/w3c/csswg-drafts/issues/9963), WebKit is broken, and Blink is nonsensical.

Loirooriol: I have added some comments.

mrobinson:
Failure here is #31428.

Issue #27190, opened 2020-07-06
Support enabling sanitizers easily


jdm:
Based on https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/sanitizer.html, I have successfully compiled Servo by adding `-Zsanitizer=address` to the `rustflags` field in my .servobuild, and using the following diff:
```diff
diff --git a/python/servo/build_commands.py b/python/servo/build_commands.py
index 58ef42e68a..f2c9b0dc4a 100644
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -645,6 +645,9 @@ class MachCommands(CommandBase):
         # given on the command line can override it
         opts = ["-Ztimings=info"] + opts

+        if "-Zsanitizer" in env.get("RUSTFLAGS", ""):
+            opts += ["-Zbuild-std", "--target", target or host]
+
         if very_verbose:
             print(["Calling", "cargo", "build"] + opts)
             for key in env:
```
If this turns out to be useful, it might be nice to add .servobuild support for enabling a sanitizer.

jdm:
There's a subtlety that took me forever to discover - since the --target flag is required, the binary ends up in target/[target]/servo, so `./mach run` ends up running the wrong binary.

ede1998:
@jdm Just stumbled over this issue. Maybe this can be resolved with #31429? Or would you also be interested to have other sanitizers easily available?

jdm:
It would be interesting to see if UBsan works in servo, but we can track that in another issue.

PR #27474, opened 2020-07-31
[WIP] Add address sanitizer support

jdm:
These changes allow building Servo with ASAN enabled using `./mach build --with-asan`, then running it with `./mach run --target x86_64-apple-darwin` (or `--target x86_64-unknown-linux-gnu` as appropriate) or `./mach test-wpt --target ...`.

CI tasks for running tests with ASAN builds are also included, mostly to check if any interesting surprises appear.

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #27190
- [x] There are tests for these changes

PR #31429, opened 2024-02-26
Add `--with-asan` option

sagudev:
Based on https://github.com/servo/servo/pull/27474, but if finding binary fails in /target/{profile}
 it fallback to /target/{target}/{profile}/servo or you can provide `--with-asan` to commands for direct binary finding.

---

- [x] These changes fix https://github.com/servo/servo/issues/27190

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -37,6 +37,9 @@ import servo.visual_studio
 from servo.command_base import BuildType, CommandBase, call, check_call
 from servo.gstreamer import windows_dlls, windows_plugins, macos_plugins
 
+SUPPORTED_ASAN_TARGETS = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu",
+                          "x86_64-apple-darwin", "x86_64-unknown-linux-gnu"]
+
 
 @CommandProvider
 class MachCommands(CommandBase):
@@ -57,7 +60,7 @@ class MachCommands(CommandBase):
                      help="Command-line arguments to be passed through to Cargo")
     @CommandBase.common_command_arguments(build_configuration=True, build_type=True)
     def build(self, build_type: BuildType, jobs=None, params=None, no_package=False,
-              verbose=False, very_verbose=False, **kwargs):
+              verbose=False, very_verbose=False, with_asan=False, **kwargs):
         opts = params or []
 
         if build_type.is_release():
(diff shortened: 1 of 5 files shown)


mrobinson (changes requested):

python/servo/platform/base.py
```diff
@@ -78,6 +82,15 @@ def install_crown(self, force: bool) -> bool:
...
 
+    def install_nightly_rust(self, force: bool) -> bool:
+        print(" * Installing respective nightly rust version...")
+        status = subprocess.call(["rustup", "install", NIGHTLY_RUST, "--profile", "minimal"])
+        status += subprocess.call(["rustup", "component", "add", "rustfmt", "rust-src", "--toolchain", NIGHTLY_RUST])
+        if status != 0:
+            raise EnvironmentError(f"Installation of rust-{NIGHTLY_RUST}")
+
+        return True
+
```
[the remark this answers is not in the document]
sagudev: Well until we do it's nice to have only one nightly for all stuff, but in the future we could add flag `--nightly-rust` that would explicitly request installing proper nightly rust.

python/servo/build_commands.py
```diff
@@ -78,10 +80,29 @@ def build(self, build_type: BuildType, jobs=None, params=None, no_package=False,
...
+            opts += ["-Zbuild-std"]
+            kwargs["target"] = target_triple
+            # sanitizers are only available on nightly
+            kwargs["toolchain"] = f"+{NIGHTLY_RUST}"
+            # do not use crown (clashes with different rust version)
+            env["RUSTC"] = "rustc"
+            # env.setdefault("CFLAGS", "")
+            # env.setdefault("CXXFLAGS", "")
+            # env["CFLAGS"] += " -fsanitize=address"
+            # env["CXXFLAGS"] += " -fsanitize=address"
```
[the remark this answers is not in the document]
sagudev: I would like to keep this for possible future extension, but I added comment about it.

sagudev:
Now that https://github.com/servo/servo/pull/31441 landed I plan to add `mach bootstrap-nightly` for ASan needs.

sagudev:
@mrobinson could we push this forward?

jdm (approved):
This looks good to me!

sagudev:
Copy trick does not work on mac (due to otool and gstreamer), so I guess we need to do this in right, but more complicated way.

python/servo/platform/base.py
```diff
@@ -12,6 +12,9 @@
 import subprocess
 from typing import Optional
 
+# Nightly version that corresponds to stable specified in rust-toolchain.toml
+NIGHTLY_RUST = "nightly-2023-09-29"
```
jschwe, line 16: Can't we just (ab)use `RUSTC_BOOTSTRAP` instead of installing another toolchain?

jschwe: I've been using RUSTC_BOOTSTRAP with unstable rust flags like -Zbuild-std

sagudev: You are right. It works 🎉 !

Thanks for the notice.

mrobinson (approved):

PR #31430, opened 2024-02-26
Fix percentages in the padding of a table-cell

Loirooriol:
---


mrobinson (approved):
I suspect you'll need a new failing expectation for legacy layout, but otherwise this looks good.

PR #31435, opened 2024-02-27
layout: Turn on synthetic small-caps for layout 2020

mrobinson:
Synthetic small caps is supported by the font subsystem, but this is
disabled in Layout 2020. We can turn this on to bring support to parity
with the old layout system.

In addition to turning on synthetic small-caps this change also improves
the way that they work. Before, synthetic small caps meant that every
character was a small version of capitalized character. After this
change, capital letters are larger than small caps versions of small
letters -- matching other browsers and the common expectation of how
small caps works.

---


Loirooriol (approved):
Looks reasonable

PR #30770, opened 2023-11-22
Backport several style changes from Gecko (11)

Loirooriol:
This continues https://github.com/servo/servo/pull/30748.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] There are tests for these changes

`/_mozilla/gfx-rs-gecko/356774-1.html` is now failing in legacy, but previously it was already wrong because both the test and the reference were supposed to show a border but they didn't. Now they do show a border, but the contents are shifted by 1px depending on border-collapse.

PR #31437, opened 2024-02-28
Update Stylo to 2023-07-23

Loirooriol:
This continues #30770.

---

`opacity-valid.html` fails because of https://bugzilla.mozilla.org/show_bug.cgi?id=1851464, should pass when a future update imports the fix.


Loirooriol:
We still need to figure out what to do with the branches in Stylo, but this PR is basically done.

mrobinson (approved):
I've reviewed these changes as well as the new commits on top of the upstream branch 2023-07-23 and this all looks good to me. Thanks!

PR #31441, opened 2024-02-28
mach: Do not use unstable rust for `rustfmt`

mrobinson:
We can use stable rust if we pass the unstable configuration as
command-line arguments to rustfmt itself. This prevents needing to
install an unstable rust toolchain.

The one downside here is that it doesn't seem that "ignore" is
supported so we have to start formatting the files in "third_party."
This shouldn't be a huge issue because we don't plan to check much more
rust code into those directories.

---

- [x] These changes do not require tests because they just adjust the `rustfmt` call.

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -39,6 +39,15 @@ PROJECT_TOPLEVEL_PATH = os.path.abspath(os.path.join(SCRIPT_PATH, "..", ".."))
 WEB_PLATFORM_TESTS_PATH = os.path.join("tests", "wpt", "tests")
 SERVO_TESTS_PATH = os.path.join("tests", "wpt", "mozilla", "tests")
 
+# Servo depends on several `rustfmt` options that are unstable. These are still
+# supported by stable `rustfmt` if they are passed as these command-line arguments.
+UNSTABLE_RUSTFMT_ARGUMENTS = [
+    "--config", "unstable_features=true",
+    "--config", "binop_separator=Back",
+    "--config", "imports_granularity=Module",
+    "--config", "group_imports=StdExternalCrate",
+]
+
 
 def format_toml_files_with_taplo(check_only: bool = True) -> int:
     taplo = shutil.which("taplo")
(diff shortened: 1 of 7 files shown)


python/servo/testing_commands.py
```diff
@@ -39,6 +39,16 @@
 WEB_PLATFORM_TESTS_PATH = os.path.join("tests", "wpt", "tests")
 SERVO_TESTS_PATH = os.path.join("tests", "wpt", "mozilla", "tests")
 
+# Servo depends on several `rustfmt` that are unstable. These are still supported by
+# stable `rustfmt if they are passed as command-line arguments. These are the
+# arguments we sue.
```
mukilan, line 44:
```suggestion
# arguments we use
```

python/servo/testing_commands.py
```diff
@@ -39,6 +39,16 @@
 WEB_PLATFORM_TESTS_PATH = os.path.join("tests", "wpt", "tests")
 SERVO_TESTS_PATH = os.path.join("tests", "wpt", "mozilla", "tests")
 
+# Servo depends on several `rustfmt` that are unstable. These are still supported by
+# stable `rustfmt if they are passed as command-line arguments. These are the
```
mukilan, line 43:
```suggestion
# stable `rustfmt` if they are passed as command-line arguments. These are the
```

python/servo/testing_commands.py
```diff
@@ -39,6 +39,16 @@
 WEB_PLATFORM_TESTS_PATH = os.path.join("tests", "wpt", "tests")
 SERVO_TESTS_PATH = os.path.join("tests", "wpt", "mozilla", "tests")
 
+# Servo depends on several `rustfmt` that are unstable. These are still supported by
```
mukilan, line 42:
```suggestion
# Servo depends on several `rustfmt` arguments that are unstable. These are still supported by
```

mrobinson:
@mukilan Thanks for the review. I've updated the PR fixing the typos.

mukilan (approved):

PR #31443, opened 2024-02-28
Add `use-system-allocator` to not use jemalloc

### LINKED ISSUE #31442: Add ability to use system allocator (on systems that use jemalloc by default)
When building ASAN build, AddressSanitizer also replaces default system alloc so it can generate smarter reports but using jemalloc overrides it.

sagudev:
~~Add `jemalloc` feature that enables jemallocator on supported platforms ( not(windows, android) ). Features are additive and unified so this is the most natural way to do this without pulling jemallocator when it's not used.~~

Add `use-system-allocator` feature so we can disable jemallocator.

---

- [x] These changes fix #31442 


mrobinson (commented):
Instead of adding more default features, which makes life complicated -- can this be a non-default use-system-allocator feature that is only exposed on the allocator crate? Cargo allows enabling features of dependencies via the command-line: https://doc.rust-lang.org/cargo/reference/features.html#command-line-feature-options

sagudev:
> Instead of adding more default features, which makes life complicated -- can this be a non-default use-system-allocator feature that is only exposed on the allocator crate? Cargo allows enabling features of dependencies via the command-line: https://doc.rust-lang.org/cargo/reference/features.html#command-line-feature-options

The problem is that features are additive, which means that if we do something like `use-system-allocator` we still pull jemallocator (so we cannot solve https://github.com/servo/servo/issues/310590).

mrobinson:
It is fine to pull in the allocator as long as you make this code conditional on the feature, right?

```rust
#[global_allocator]
static ALLOC: Allocator = Allocator;
```

mrobinson (approved):
Thanks for the updates. This looks good to me.

PR #31452, opened 2024-02-29
mach: Speed up `taplo` runs on MacOS

### LINKED ISSUE #31440: `taplo` is very slow on MacOS
Taplo seems to have an issue on MacOS with processing speed. This is what I see when I run `./mach fmt`

```shell
$ time ./mach fmt

real	0m19.056s
user	0m2.091s
sys	0m7.640s
```

If I comment out taplo formatting, I see:

```shell
$ time ./mach fmt

real	0m1.622s
user	0m1.320s
sys	0m0.145s
```

mrobinson:
`taplo` is the TOML formatter that we use. It seems to be using a very
slow directory walking mechanism on MacOS. This change works around that
issue by explicitly listing globs of TOML files to format and check.

Before:
```text
real	0m17.632s
user	0m2.164s
sys	0m6.916s
```

After:
```text
real	0m1.519s
user	0m1.391s
sys	0m0.112s
```

---

- [ ] These changes fix #31440.
- [x] These changes do not require tests because they merely make `./mach fmt` etc faster.

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -48,6 +48,18 @@ UNSTABLE_RUSTFMT_ARGUMENTS = [
     "--config", "group_imports=StdExternalCrate",
 ]
 
+# Listing these globs manually is a work-around for very slow `taplo` invocation
+# on MacOS machines. If `taplo` runs fast without the globs on MacOS, this
+# can be removed.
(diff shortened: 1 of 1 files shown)


sagudev (approved):
We should also report issue to taplo.

PR #31453, opened 2024-02-29
layout: Add support for box-shadow to Layout 2020

mrobinson:
A couple tests fail because they rely on `text-shadow` support which is also
missing.

---


nicoburns:
Why does box-shadow support live in the layout crate(s) (AFAIK shadows don't affect layout)? Is this just a case of there not being a better place to put this code?

mrobinson:
> Why does box-shadow support live in the layout crate(s) (AFAIK shadows don't affect layout)? Is this just a case of there not being a better place to put this code?

Box shadow is part of CSS (https://drafts.csswg.org/css-backgrounds/#box-shadow), so depends on both computed values and style. In addition, since it's part of CSS it also has a place in the stacking order of CSS content (https://www.w3.org/TR/CSS22/zindex.html). Layout is the component of the web engine that combines style and DOM to produce display lists. There's really no reasonable way to add these shadows apart from during this process.

Loirooriol (approved):
Looks good, but note this doesn't follow https://drafts.csswg.org/css-backgrounds-3/#shadow-layers

> If the affected box has multiple fragments, the shadows are applied as specified in [`box-decoration-break`](https://drafts.csswg.org/css-break-4/#propdef-box-decoration-break).

The initial value is `slice`, but this seems to behave as `clone`.

```html
<!DOCTYPE html>
<div style="width: 370px; border: solid cyan; font: 12px / 30px monospace">
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris <span style="box-shadow: 0px 0px 0px 5px">libero sem, facilisis nec purus pretium, tempor hendrerit mauris. Sed dignissim commodo ex, sit amet fermentum est interdum at. Nam venenatis lobortis libero</span>, vel iaculis dolor sodales in.
</div>
```

- Servo:
  
- Blink:
  


But this can be addressed in a follow-up.

PR #31455, opened 2024-02-29
Fix size of tables in flow layout

Loirooriol:
The contents of a table can make it bigger than what we would expect from its 'width', 'min-width', 'height' and ' min-height' properties. Also, 'width: auto' doesn't stretch it to fill the containing block.

We had to refactor the resolution of margins to happen after layout, otherwise 'auto' margins wouldn't align correctly.

---


components/layout_2020/formatting_contexts.rs
```diff
@@ -73,6 +73,11 @@ pub(crate) struct IndependentLayout {
     /// <https://drafts.csswg.org/css2/visudet.html#root-height>
     pub content_block_size: Au,
 
+    /// The contents of a table may force it to become wider than what we would expect
+    /// from 'width' and 'min-width'. This is the resulting inline content size,
+    /// or None for non-table layouts.
+    pub content_inline_size_for_table: Option<Au>,
```
mrobinson, line 79: It might be worth adding this text from the specification here:

```
    /// > See <https://drafts.csswg.org/css-tables/#layout>:
    /// > Unlike other block-level boxes, tables do not fill their containing block by default. 
    /// > When their [width](https://drafts.csswg.org/css-sizing-3/#propdef-width) computes
    /// > to auto, they behave as if they had fit-content specified instead. This is different from 
    /// > most block-level boxes, which behave as if they had stretch instead.
```

components/layout_2020/formatting_contexts.rs
```diff
@@ -187,18 +192,26 @@ impl NonReplacedFormattingContext {
         &self,
         layout_context: &LayoutContext,
         positioning_context: &mut PositioningContext,
+        containing_block_for_children: &ContainingBlock,
         containing_block: &ContainingBlock,
```
mrobinson, line 196: Just for the sake of clarity, perhaps this can be:

```suggestion
        containing_block_from_parent: &ContainingBlock,
```

components/layout_2020/table/layout.rs
```diff
@@ -667,12 +688,12 @@ impl<'a> TableLayout<'a> {
         // > * If the table-root has 'width: auto', the used width is the greater of min(GRIDMAX,
         // >   the table’s containing block width), the used min-width of the table.
         let used_width_of_table = match content_box_size.inline {
```
mrobinson, line 690: I think the comment above is inaccurate now? Let's just remove the part about what to do if the table has a computed inline size and write something like:

```rust
// This diverges from the specification a little:
// If table-root has a computed value for inline size: Use the max of the minimum inline content size
// of the grid items and the inline size specified on the table style. The inline size specified on table
// style is already expressed via the content block size for children.
```

components/layout_2020/table/layout.rs
```diff
@@ -1327,27 +1348,33 @@ impl<'a> TableLayout<'a> {
...
+        let style = &self.table.style;
+        let table_height_from_style =
+            match style.content_box_size(containing_block, &self.pbm).block {
+                LengthPercentage(_) => containing_block_for_children.block_size,
+                Auto => style
+                    .content_min_box_size(containing_block, &self.pbm)
+                    .block
+                    .map(Au::from),
+            }
+            .auto_is(Au::zero);
```
mrobinson, line 1365:
```suggestion
        // When there is a specified length for block size, `containing_block_for children` will have
        // this value, properly clamping for minimum and maximum block size.
        let table_height_from_style =
            match style.content_box_size(containing_block, &self.pbm).block {
                LengthPercentage(_) => containing_block_for_children.block_size,
                Auto => style
                    .content_min_box_size(containing_block, &self.pbm)
                    .block
                    .map(Au::from),
            }
            .auto_is(Au::zero);
```

components/layout_2020/flow/mod.rs
```diff
@@ -1240,15 +1281,15 @@ struct ContainingBlockPaddingBorderAndMargin<'a> {
     effective_margin_inline_start: Length,
 }
 
-/// Given the style for in in flow box and its containing block, determine the containing
-/// block for its children and the margin it should use.
+/// Given the style for in in-flow box and its containing block, determine the containing
```
mrobinson, line 1284:
```suggestion
/// Given the style for an in-flow box and its containing block, determine the containing
```

components/layout_2020/flow/mod.rs
```diff
@@ -1013,19 +1034,35 @@ impl NonReplacedFormattingContext {
...
+                } else {
+                    content_size = LogicalVec2 {
+                        block: block_size.auto_is(|| {
+                            layout
+                                .content_block_size
+                                .clamp_between_extremums(
+                                    min_box_size.block.into(),
+                                    max_box_size.block.map(|t| t.into()),
+                                )
+                                .into()
```
mrobinson, line 1061: Can this be:

```suggestion
                                .content_block_size
                                .into()
                                .clamp_between_extremums(
                                    min_box_size.block,
                                    max_box_size.block,
                                )
```

components/layout_2020/flow/mod.rs
```diff
@@ -1013,19 +1034,35 @@ impl NonReplacedFormattingContext {
...
-                    }),
-                };
+
+                if let Some(inline_size) = layout.content_inline_size_for_table {
+                    let outer_inline_size = inline_size + pbm.padding_border_sums.inline;
+                    placement.set_inline_size(outer_inline_size, &pbm);
+                    if outer_inline_size > placement_rect.size.inline {
+                        positioning_context.truncate(&positioning_context_length);
+                        continue;
+                    }
```
mrobinson, line 1046:
```suggestion
                    // If this is a table, it's impossible to know the inline size it will take up until after
                    // trying to place it. If the table doesn't fit into this positioning rectangle due to
                    // incompatibility in the inline axis, then retry at another location.
                    if outer_inline_size > placement_rect.size.inline {
                        positioning_context.truncate(&positioning_context_length);
                        continue;
                    }
```

components/layout_2020/flow/mod.rs
```diff
@@ -1013,19 +1034,35 @@ impl NonReplacedFormattingContext {
...
-                                min_box_size.block.into(),
-                                max_box_size.block.map(|t| t.into()),
-                            )
-                            .into()
-                    }),
-                };
+
+                if let Some(inline_size) = layout.content_inline_size_for_table {
+                    let outer_inline_size = inline_size + pbm.padding_border_sums.inline;
+                    placement.set_inline_size(outer_inline_size, &pbm);
```
mrobinson, line 1042: Can this be moved to inside the if-statement right below?

mrobinson (approved):

components/layout_2020/table/layout.rs
Loirooriol: But there is always a computed value for inline size...

I have refactored this function and tweaked the comments, please check.

components/layout_2020/table/layout.rs
Loirooriol: Tweaked the working a bit because it's not just when specifying lengths, also with percentages.

components/layout_2020/flow/mod.rs
Loirooriol: It could be moved. But setting it unconditionally is an optimization: even if the condition is false, we may need more iterations, and then we ensure that we will find a rect that is big enough. So this avoids potential useless layouts.

components/layout_2020/flow/mod.rs
Loirooriol: I have shortened the lines and added the explanation for always calling `set_inline_size`.

components/layout_2020/flow/mod.rs
Loirooriol: Nice! But "type annotations needed", so

```suggestion
                            Length::from(layout.content_block_size)
                                .clamp_between_extremums(min_box_size.block, max_box_size.block)
```

Loirooriol:
Funny, GRIDMIN can be greater than GRIDMAX, we should work on that on a follow-up.

components/layout_2020/flow/mod.rs
mrobinson: Ah, right. I had forgotten that tables also need to fit among floats in the block direction! :facepalm:

mrobinson: The new comment is great. Thanks!

mrobinson (approved):
Okay. This looks good to me. I made a couple small changes (naming and test expectations update). I'll send this to the merge queue.

PR #31468, opened 2024-03-01
script: Implement `<meta http-equiv="refresh">`

### LINKED ISSUE #25082: Implement http-equiv refresh
WPT html/webappapis/dynamic-markup-insertion/opening-the-input-stream/abort-refresh-immediate (some cases) and abort-refresh-multisecond-meta expect `<meta http-equiv="refresh"` to do its thing as in https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-refresh . Currently, http-equiv is commented out at the IDL level. 

#25068 is the same missing behavior, but as triggered by the never-formally-standardized Refresh response header instead of the http-equiv. (As I understand it, once upon a time http-equiv literally just meant "behave as though you saw this response header" at first, then diverged as header semantics changed out from under it; its legal values still mostly do the same thing as headers of the same name.)

syvb:
Implements support for  `<meta http-equiv="refresh" content="…; url=…">` meta tags.

---

- [x] These changes fix #25082


components/script/dom/document.rs
```diff
@@ -233,6 +234,17 @@ enum FocusTransaction {
     InTransaction(Option<Dom<Element>>),
 }
 
+/// Information about a declarative refresh
+#[derive(JSTraceable, MallocSizeOf)]
+pub enum DeclarativeRefresh {
+    PendingLoad {
+        #[no_trace]
+        url: DomRefCell<ServoUrl>,
```
jdm, line 242: If we store the enum inside of a refcell, why do we need one here as well?

components/script/dom/htmlmetaelement.rs
```diff
@@ -2,29 +2,57 @@
...
 
 #[dom_struct]
 pub struct HTMLMetaElement {
     htmlelement: HTMLElement,
 }
 
+#[derive(JSTraceable, MallocSizeOf)]
+pub struct RefreshRedirectDue {
+    #[no_trace]
+    pub url: DomRefCell<ServoUrl>,
```
jdm, line 42: Why is this refcell necessary?

components/script/dom/htmlmetaelement.rs
```diff
@@ -81,6 +111,97 @@ impl HTMLMetaElement {
...
+            captures
+        } else {
+            return;
+        };
+        let time = if let Some(time_string) = captures.name("time") {
+            u64::from_str(&String::from_utf8_lossy(time_string.as_bytes())).unwrap_or(0)
+        } else {
+            0
+        };
+        let captured_url = captures.name("url1").or_else(|| {
```
jdm, line 171: I wonder if this would be easier to read if we used or instead or or_else?

components/script/dom/htmlmetaelement.rs
```diff
@@ -81,6 +111,97 @@ impl HTMLMetaElement {
...
+        // 12-13
+        if document.completely_loaded() {
+            // TODO: handle active sandboxing flag
+            let window = window_from_node(self);
+            window.upcast::<GlobalScope>().schedule_callback(
+                OneshotTimerCallback::RefreshRedirectDue(RefreshRedirectDue {
+                    window: window.clone(),
+                    url: DomRefCell::new(url_record),
+                }),
+                MsDuration::new(dbg!(time.saturating_mul(1000))),
```
jdm, line 195: Is this dbg still necessary?

jdm (changes requested):
This looks good! Just a few questions about some possible simplifications.

components/script/dom/htmlmetaelement.rs
syvb: I accidentally left that in from debugging, removed.

components/script/dom/htmlmetaelement.rs
syvb: I simplified it to use `.or`, although I still don't like the wait rustfmt formats it.

components/script/dom/document.rs
syvb: It's not needed (nor is the one below), I removed it. (my earlier implementation did require a `DomRefCell` there, and I forgot to check if it was still needed when changing it)

jdm (approved):

PR #31481, opened 2024-03-02
Add CSSOM support for CSS layers

### LINKED ISSUE #31479: Some css-cascade tests crash
These WPT are crashing both in legacy and 2020:
- /css/css-cascade/all-prop-revert-layer.html
- /css/css-cascade/idlharness.html
- /css/css-cascade/layer-cssom-order-reverse-at-property.html
- /css/css-cascade/layer-cssom-order-reverse.html
- /css/css-cascade/layer-statement-before-import.html
- /css/css-cascade/parsing/layer.html

Loirooriol:
Instead of just crashing.

---

- [X] These changes fix #31479


jdm (approved):
Makes sense!

PR #31486, opened 2024-03-04
dependencies: Upgrade to WebRender 0.64

mrobinson:
This brings the version of WebRender used in Servo up-to-date with Gecko
upstream. The big change here is that HiDPI is no longer handled via
WebRender. Instead this happens via a scale applied to the root layer in
the compositor. In addition to this change, various changes are made to
Servo to adapt to the new WebRender API.

Along with a wide variety of new test passes there are also some new failures:

- This version of WebRender exposes a bug where stacking context
   contents are clipped before they are filtered by stacking
    contexts. In reality the contents of stacking contexts should only
    be clipped by clips from inside the stacking context and then the
    stacking context should clip everythig post-filter.
    stacking context contents
    - `/css/filter-effects/drop-shadow-clipped-001.html`
    - `/css/filter-effects/will-change-blur-filter-under-clip.html`
- It seems that the way we pass scroll offsets causes some sticky
   tests to fail, likely because the sticky nodes are only being being
   updated due to sticky offsets on subsequent frames. I'm not sure
   why this is happening exactly, but it's only an issues in tests.
    - `/css/css-position/sticky/position-sticky-escape-scroller-004.html`
    - `/css/css-position/sticky/position-sticky-scroll-with-clip-and-abspos.html`
- This version introduces some new pixel-level differences:
    - `/_mozilla/css/pixel_snapping_position_a.html`
    - `/css/css-transforms/transform-table-009.html`
    - `/css/css-transforms/transform-table-011.html`
    - `/css/css-transforms/transform-table-010.html`

---


mrobinson:
These failures are due to compilation against an old version of Stylo.

components/compositing/compositor.rs
```diff
@@ -944,13 +982,13 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+    fn set_root_content_pipeline_handling_device_scaling(&self, transaction: &mut Transaction) {
         let root_content_pipeline = match self.root_content_pipeline.id {
             Some(id) => id.to_webrender(),
             None => return,
         };
 
-        let zoom_factor = self.pinch_zoom_level();
+        let zoom_factor = self.device_pixels_per_page_pixel().0;
         if zoom_factor == 1.0 {
             transaction.set_root_pipeline(root_content_pipeline);
```
mukilan, line 993: Will this path ever be taken on HiDPI devices, given that zoom_factor now includes the HiDPI factor?

components/compositing/compositor.rs
```diff
@@ -1465,15 +1489,17 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
     /// scrolling to the applicable scroll node under that point. If a scroll was
```
mukilan, line 1489: GitHub is not allowing me to add the comment on the exact line, but the documentation above this line needs to be updated to use `DevicePoint` instead of `WorldPoint`.

```
    /// Perform a hit test at the given [`WorldPoint`]
 ```

File: components/compositing/compositor.rs [hunk 3]
```diff
@@ -705,25 +688,46 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
...
+                    return;
+                }
+
                 let mut txn = Transaction::new();
-                txn.scroll_node_with_id(point, scroll_id, ScrollClamping::NoClamping);
-                self.generate_frame(&mut txn);
+                txn.set_scroll_offsets(
+                    external_scroll_id,
+                    vec![SampledScrollOffset {
+                        offset,
```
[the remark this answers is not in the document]
mrobinson: The scroll tree and WebRender have opposite interpretations of the scroll offset unfortunately. I'd like to fix that in a followup, because it is pretty confusing indeed.

mrobinson:
@mukilan Thanks for the review! I think I have responded to all comments.

mukilan (approved):

PR #31497, opened 2024-03-04
Update WPT manifest

Loirooriol:
In #31481 I forgot to run `./mach update-manifest`

---


mrobinson:
I'm surprised that this was able to pass `./mach test-tidy` since theoretically that should check that the manifest is valid.

mrobinson (approved):

Loirooriol:
Yeah it's strange that https://github.com/servo/servo/actions/runs/8130302646/ and https://github.com/servo/servo/actions/runs/8130657242 didn't complain.

PR #31499, opened 2024-03-04
Add basic support for enumerating computed styles

Loirooriol:
The old code was so broken, it was enumerating the style attribute instead of the computed styles.

So this patch implements a basic enumeration, but not including custom properties whose computed value is not the guaranteed-invalid value.

---


components/script/dom/cssstyledeclaration.rs
```diff
@@ -342,9 +345,44 @@ impl CSSStyleDeclaration {
     }
 }
 
+lazy_static! {
+    static ref ENABLED_LONGHAND_PROPERTIES: Vec<LonghandId> = {
+        // The 'all' shorthand contains all the enabled longhands, with 2 exceptions:
+        // 'direction' and 'unicode-bidi'
+        let mut v: Vec<LonghandId> = ShorthandId::All.longhands().collect();
```
mrobinson, line 352:
```suggestion
        let mut enabled_properties: Vec<LonghandId> = ShorthandId::All.longhands().collect();
```

components/script/dom/cssstyledeclaration.rs
```diff
@@ -342,9 +345,44 @@ impl CSSStyleDeclaration {
     }
 }
 
+lazy_static! {
+    static ref ENABLED_LONGHAND_PROPERTIES: Vec<LonghandId> = {
+        // The 'all' shorthand contains all the enabled longhands, with 2 exceptions:
+        // 'direction' and 'unicode-bidi'
```
mrobinson, line 351: Hrm. I'm a bit confused what is going on here is the exception that `direction` and `unicode-bidi` are added unconditionally? If we maybe this could be:

```suggestion
        // The 'all' shorthand contains all the enabled longhands, with 2 exceptions:
        // 'direction' and 'unicode-bidi' are always added.
```

mrobinson (approved):
Looks good. Thanks!

components/script/dom/cssstyledeclaration.rs
Loirooriol: https://drafts.csswg.org/css-cascade-4/#all-shorthand

> The [all](https://drafts.csswg.org/css-cascade-4/#propdef-all) property is a [shorthand](https://drafts.csswg.org/css-cascade-4/#shorthand-property) that resets all CSS properties except [direction](https://drafts.csswg.org/css-writing-modes-3/#propdef-direction) and [unicode-bidi](https://drafts.csswg.org/css-writing-modes-3/#propdef-unicode-bidi).

Here I'm using the `all` shorthand to get a list of the longhand properties, but these 2 are missing, so they need to be added (if they are enabled, like in 2013, they are disabled in 2020).

mrobinson: Makes sense. "Exceptions" sort of implies that the entries are removed, so maybe something like this:

```
       // The 'all' shorthand contains all the enabled longhands with 2 exceptions:
       // 'direction' and 'unicode-bidi', so these must be added afterward.
```

components/script/dom/cssstyledeclaration.rs
Loirooriol: Used `enabled_longhands` because this doesn't include shorthands.

Issue #31500, opened 2024-03-04
Fix as many clippy problems as possible


mrobinson:
Steps: 

1. Get Servo building and running locally following project documentation at http://github.com/servo/servo.
2. Run `./mach cargo-clippy`
3. Fix an issue by understanding what the error means and correcting the issue. For instance
```
warning: redundant field names in struct initialization
   --> components/remutex/lib.rs:168:13
    |
168 |             data: data,
    |             ^^^^^^^^^^ help: replace it with: `data`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names
    = note: `#[warn(clippy::redundant_field_names)]` on by default
```
    In this case `data: data` can just be replaced with `data`.
4. Open an pull request with your fix, filling out the template, and giving it a full name like the following "clippy: Fix a redundant field warning in `components/remutex/lib.rs`. 
5. Include a link to this issue in your pull request description body.

<h3>As this issue is large enough to be worked on in parallel, please don't ask to assign it to yourself. If you want to post fixes for a particular section of code, just check the PRs linked here to make sure that you are fixing different issues.</h3>

***NOTE: The warnings for [missing safety documentation](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc) for unsafe code is already handled in PR #33748. Please ignore those.***

servo-highfive:
Hi! If you have any questions regarding this issue, feel free to make a comment here, or ask it in [Zulip](https://servo.zulipchat.com/).

If you intend to work on this issue, then add `@servo-highfive: assign me` to your comment, and I'll assign this to you. :smile:

pranayrauni:
Hi I want to work on this issue. 

@servo-highfive: assign me

servo-highfive:
Hey @pranayrauni! Thanks for your interest in working on this issue. It's now assigned to you!

mnaibei:
Hi @mrobinson, I am an Outreachy applicant and I would like to undertake this project.

Payne680:
@mrobinson Please I will love to work on this task but can you give some resources on how to successfully run the project locally ?

Loirooriol:
BTW, some of the clippy warnings can be fixed automatically, e.g.

```
CARGO_BUILD_RUSTC=rustc cargo clippy --fix -p layout_2020 --broken-code
```

will apply some fixes automatically in the `layout_2020` crate (and its tests). But some warnings need to be addressed manually.

[...]

PR #31501, opened 2024-03-04
clippy: Fix several warnings in components/devtools

eerii:
Fixes several clippy warnings in the `components/devtools` directory. It is part of #31500.

---

- [x] These changes are part of #31500.

### DIFF
--- a/components/devtools/actor.rs
+++ b/components/devtools/actor.rs
@@ -5,7 +5,7 @@
 use std::any::Any;
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
-use std::mem::replace;
+use std::mem;
 use std::net::TcpStream;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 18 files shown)


mrobinson:
@eerii Nice work. This change looks excellent. One thing you might need to do is to run `./mach fmt` before uploading it.

eerii:
> @eerii Nice work. This change looks excellent. One thing you might need to do is to run `./mach fmt` before uploading it.

Thanks! I forgot about it last time.

- Now that there was a redundant match removed from `components/devtools/actors/emulation.rs`, `msg_type` is unused. Should I remove it and fix the function invocations or leave it and just mark it with an underscore?
- For clippy warnings about functions with too many arguments, should I add `#[allow(clippy::too_many_arguments)]` or leave the warning?

mrobinson:
>     * Now that there was a redundant match removed from `components/devtools/actors/emulation.rs`, `msg_type` is unused. Should I remove it and fix the function invocations or leave it and just mark it with an underscore?

I think you should leave the argument as that function is implementing a trait. Simply prefix it with an underscore like the other unused arguments.

>     * For clippy warnings about functions with too many arguments, should I add `#[allow(clippy::too_many_arguments)]` or leave the warning?

Yes, I think that's a reasonable approach.

Thank you!

eerii:
Thank you! I will do that and make sure everything works, and then I'll update the pr

mrobinson (approved):
Nice work!

PR #31469, opened 2024-03-01
Enable more css tests

Loirooriol:
In particular, these directories are no longer skipped:
 - CSS1 (has no tests)
 - css-box (except margin-trim)
 - css-cascade
 - css-display (except run-in)
 - css-easing
 - css-env
 - css-lists
 - css-namespaces
 - css-overflow
 - css-pseudo
 - css-sizing (except aspect-ratio and contain-intrinsic-size)
 - css-syntax
 - reference
 - support (has no tests)

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes do not require tests because there is no change in behavior

PR #31502, opened 2024-03-05
Enable `css-text` explicitly in include.ini

mukilan:
The WPT test runner has some strange logic for determining the key's value for a node like `[css]` or `[css-text]`. In this logic, if the node doesn't have an explicit value for a key (here `skip`), then the implicit root node's setting i.e key/value pair at the top of the file that is not nested under a heading, is used as the default fallback value [1]. Only when the implicit root node doesn't have an explicit value set does the logic starts looking at the current node's parent [2].

In our case, in `include.ini` the default value for `skip` is `true` as that is the first line in the file.
Since `[css-text]` doesn't have `skip` set explicitly, the default value of `true` is used even though the parent's value is `false`.

[1]: https://github.com/servo/servo/blob/2bafcf9f182bf907c4a4391f931a7364bd7b804f/tests/wpt/tests/tools/wptrunner/wptrunner/wptmanifest/backends/conditional.py#L265
[2]: https://github.com/servo/servo/blob/2bafcf9f182bf907c4a4391f931a7364bd7b804f/tests/wpt/tests/tools/wptrunner/wptrunner/manifestinclude.py#L59

- [x] These changes do not require tests because they only enable previously disabled tests.

### DIFF
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -94,6 +94,7 @@ skip: true
   [css-speech]
     skip: true
   [css-text]
+    skip: false
     [i18n]
       skip: true
   [css-typed-om]
--- a/tests/wpt/meta-legacy-layout/css/css-text/letter-spacing/letter-spacing-percent-001.html.ini
+++ b/tests/wpt/meta-legacy-layout/css/css-text/letter-spacing/letter-spacing-percent-001.html.ini
@@ -1,0 +1,2 @@
+[letter-spacing-percent-001.html]
+  expected: FAIL
--- a/tests/wpt/meta-legacy-layout/css/css-text/parsing/letter-spacing-valid.html.ini
+++ b/tests/wpt/meta-legacy-layout/css/css-text/parsing/letter-spacing-valid.html.ini
@@ -1,0 +1,12 @@
+[letter-spacing-valid.html]
+  [e.style['letter-spacing'\] = "120%" should set the property value]
+    expected: FAIL
+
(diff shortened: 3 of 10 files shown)


mrego (approved):
Nice catch.

Loirooriol (commented):
Ah, this means I broke it in #31469. So `css-box`, `css-display` and `css-sizing` are also broken...
It's fine to land as-is to fix the wpt.servo.org problem, but I think we should fix the logic at some point to just use the parent value.

Loirooriol:
Seems some test regressed (both legacy and 2020) while css-text was skipped :(

- /css/css-text/letter-spacing/letter-spacing-percent-001.html
-  /css/css-text/parsing/letter-spacing-computed.html
- /css/css-text/parsing/letter-spacing-valid.html
- /css/css-text/word-spacing/word-spacing-002.html
- /css/css-text/word-spacing/word-spacing-percent-001.html

Loirooriol:
OK, they aren't regressions, they are just new tests from #31482

mukilan:
Sorry, just seeing the comments here. Thanks @Loirooriol for updating the expectations. 

>  It's fine to land as-is to fix the wpt.servo.org problem, but I think we should fix the logic at some point to just use the parent value.

I agree we should fix the logic. However, it looks like this logic does double duty and is used for both `include.ini manifests` and the usual expectation manifest with conditional expressions, so it wasn't immediately evident to me if this is behaviour is by design or just an unintended side effect (that suggests the two manifests need to be handled separately or at least needs some conditional logic to disable the fallback).

PR #31505, opened 2024-03-05
script: Start rework to better match the specification HTML event loop

### LINKED ISSUE #31242: Fix HTML event-loop
Our current implementation of the [HTML event-loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model) is in need of fixing. For this issue I am referring only to the window event-loop, which requires interacting with the rendering. Worker loops seem ok for now(we're not delivering rafs to workers yet).

--------------

Quick recap on how it is supposed to work:

1. The event-loop runs a task sequentially. 
2. It runs a micro-task checkpoint(so one after each task is run)
3. It  reports  long tasks.
4. It runs the idle callback logic.

In parallel(which means anywhere else in Servo, like the compositor), we must wait for for a [rendering opportunity](https://html.spec.whatwg.org/multipage/#event-loop-processing-model:rendering-opportunity) for that event-loop, and then queue a task that runs the [update the rendering steps](https://html.spec.whatwg.org/multipage/#update-the-rendering) 

The task queued is then appended to a task queue, and eventually handled at 1 above. Note that this tasks runs a lot of steps, but atomically within the same task, and in a specific order. 

For example, step 11 is [update animations and send events](https://drafts.csswg.org/web-animations-1/#update-animations-and-send-events), step 14 is [update animations and send events](https://drafts.csswg.org/web-animations-1/#update-animations-and-send-events), and step 15 runs the resize observer logic. 

Running these "update the rendering" steps in that way allows for a certain level of determinism, for example if one requests an animation frame callback from within the last notification callback of a resize observer, one can be certain that by the time the next resize notification is callback, the animation frame callback will have been invoked first. This [resize observer test](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/tests/wpt/tests/resize-observer/eventloop.html) is an example of relying on that ordering. 

--------------

Quick recap on how it current works:
1. We [select](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1490) on various channels, including one representing the various task queues, but also other channels representing messages from the constellation/compositor, the image cache, webgpu, and others. 
2. We handle certain type of messages [first](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1519). 
7. We then [check for additional messages](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1605)
8. We then [`update_animations_and_send_events`](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1627)
9. We then process events collected at 1 and 3 above [sequentially](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1631)
10. We then perform a [micro task checkpoint](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1681)
11. We then [`maybe_queue_document_completion`](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1689)
12. Finally, we [issue any batched reflows](https://github.com/servo/servo/blob/e15262c5a03f78f480282900d6828be19774ced3/components/script/script_thread.rs#L1699)

The [ResizeObserver PR](https://github.com/servo/servo/pull/31108) adds a step before the reflows, that runs the resize observer notification steps. 

--------------

The problem: our current implementation does not follow the ordering of the spec, hence it is a source of timing issues. 

Solution: restructure the event-loop to only run the 4 steps it is supposed to run, and restructure the handling of all other messages as happening via a tasks running the "update the rendering" steps.

gterzian:
Bring our HTML event-loop implementation in line with the spec: add an "update the rendering" task. 

---

- [x] These changes are part of #31242.
- [x] These changes do not require tests. Although they cause two new failures, these changes are written in a way to affect as few tests as possible. This is the start of a large refactor and many tests are quite sensitive to the timing of our current design which doesn't match the specification.

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -3228,8 +3228,7 @@ impl Document {
         }
     }
 
-    /// Note a pending animation tick,
-    /// to be processed at the next `update_the_rendering` task.
+    /// Note a pending animation tick, to be processed at the next `update_the_rendering` task.
     pub fn note_pending_animation_tick(&self, tick_type: AnimationTickType) {
         self.pending_animation_ticks.borrow_mut().extend(tick_type);
     }
@@ -3245,8 +3244,7 @@ impl Document {
         }
     }
 
-    /// Note a pending compositor event,
-    /// to be processed at the next `update_the_rendering` task.
+    /// Note a pending compositor event, to be processed at the next `update_the_rendering` task.
     pub fn note_pending_compositor_event(&self, event: CompositorEvent) {
         let mut pending_compositor_events = self.pending_compositor_events.borrow_mut();
         if matches!(event, CompositorEvent::MouseMoveEvent { .. }) {
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -1228,7 +1228,7 @@ impl Document {
     }
 
     #[allow(unsafe_code)]
-    pub unsafe fn handle_mouse_event(
+    pub unsafe fn handle_mouse_button_event(
         &self,
         button: MouseButton,
         client_point: Point2D<f32>,
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -3231,8 +3231,7 @@ impl Document {
     /// Note a pending animation tick,
     /// to be processed at the next `update_the_rendering` task.
     pub fn note_pending_animation_tick(&self, tick_type: AnimationTickType) {
-        let mut pending_animation_ticks = self.pending_animation_ticks.borrow_mut();
-        pending_animation_ticks.extend(tick_type);
+        self.pending_animation_ticks.borrow_mut().extend(tick_type);
     }
 
     /// As part of a `update_the_rendering` task, tick all pending animations.
@@ -3249,27 +3248,27 @@ impl Document {
     /// Note a pending compositor event,
     /// to be processed at the next `update_the_rendering` task.
     pub fn note_pending_compositor_event(&self, event: CompositorEvent) {
-        match event {
-            CompositorEvent::MouseMoveEvent { .. } => {
-                let mut index = self.mouse_move_event_index.borrow_mut();
-                if let Some(index) = *index {
-                    // Replace the pending mouse move event.
-                    let mut pending_events = self.pending_compositor_events.borrow_mut();
-                    let mouse_move_event = pending_events.get_mut(index).expect("index is valid");
-                    *mouse_move_event = event;
-                } else {
-                    self.pending_compositor_events.borrow_mut().push_back(event);
-                    *index = Some(self.pending_compositor_events.borrow().len() - 1);
-                }
-            },
-            _ => {
-                self.pending_compositor_events.borrow_mut().push_back(event);
-            },
+        let mut pending_compositor_events = self.pending_compositor_events.borrow_mut();
+        if matches!(event, CompositorEvent::MouseMoveEvent { .. }) {
+            // First try to replace any existing mouse move event.
+            if let Some(mouse_move_event) = self
+                .mouse_move_event_index
+                .borrow()
+                .and_then(|index| pending_compositor_events.get_mut(index))
+            {
+                *mouse_move_event = event;
+                return;
+            }
+
+            *self.mouse_move_event_index.borrow_mut() =
+                Some(self.pending_compositor_events.borrow().len());
         }
+
+        self.pending_compositor_events.borrow_mut().push_back(event);
     }
 
-    /// Get pending compositor events, for processig within an `update_the_rendering` task.
-    pub fn get_pending_compositor_events(&self) -> VecDeque<CompositorEvent> {
+    /// Get pending compositor events, for processing within an `update_the_rendering` task.
+    pub fn take_pending_compositor_events(&self) -> VecDeque<CompositorEvent> {
         // Reset the mouse event index.
         *self.mouse_move_event_index.borrow_mut() = None;
         mem::take(&mut *self.pending_compositor_events.borrow_mut())
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -450,7 +450,7 @@ pub struct Document {
     /// Pending composition events, to be handled at the next rendering opportunity.
     #[no_trace]
     #[ignore_malloc_size_of = "CompositorEvent contains data from outside crates"]
-    pending_compositor_events: DomRefCell<VecDeque<CompositorEvent>>,
+    pending_compositor_events: DomRefCell<Vec<CompositorEvent>>,
     /// The index of the last mouse move event in the pending compositor events queue.
     mouse_move_event_index: DomRefCell<Option<usize>>,
     /// Pending animation ticks, to be handled at the next rendering opportunity.
@@ -3264,11 +3264,11 @@ impl Document {
                 Some(self.pending_compositor_events.borrow().len());
         }
 
-        self.pending_compositor_events.borrow_mut().push_back(event);
+        self.pending_compositor_events.borrow_mut().push(event);
     }
 
     /// Get pending compositor events, for processing within an `update_the_rendering` task.
-    pub fn take_pending_compositor_events(&self) -> VecDeque<CompositorEvent> {
+    pub fn take_pending_compositor_events(&self) -> Vec<CompositorEvent> {
         // Reset the mouse event index.
         *self.mouse_move_event_index.borrow_mut() = None;
         mem::take(&mut *self.pending_compositor_events.borrow_mut())
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -46,8 +46,9 @@ use profile_traits::ipc as profile_ipc;
 use profile_traits::time::{TimerMetadata, TimerMetadataFrameType, TimerMetadataReflowType};
 use script_layout_interface::{PendingRestyle, ReflowGoal, TrustedNodeAddress};
 use script_traits::{
-    AnimationState, DocumentActivity, MouseButton, MouseEventType, MsDuration, ScriptMsg,
-    TouchEventType, TouchId, UntrustedNodeAddress, WheelDelta,
+    AnimationState, AnimationTickType, CompositorEvent, DocumentActivity, MouseButton,
+    MouseEventType, MsDuration, ScriptMsg, TouchEventType, TouchId, UntrustedNodeAddress,
+    WheelDelta,
 };
 use servo_arc::Arc;
 use servo_atoms::Atom;
@@ -446,6 +447,16 @@ pub struct Document {
     dirty_root: MutNullableDom<Element>,
     /// <https://html.spec.whatwg.org/multipage/#will-declaratively-refresh>
     declarative_refresh: DomRefCell<Option<DeclarativeRefresh>>,
+    /// Pending composition events, to be handled at the next rendering opportunity.
+    #[no_trace]
+    #[ignore_malloc_size_of = "CompositorEvent contains data from outside crates"]
+    pending_compositor_events: DomRefCell<VecDeque<CompositorEvent>>,
+    /// The index of the last mouse move event in the pending compositor events queue.
+    mouse_move_event_index: DomRefCell<Option<usize>>,
+    /// Pending animation ticks, to be handled at the next rendering opportunity.
+    #[no_trace]
+    #[ignore_malloc_size_of = "AnimationTickType contains data from an outside crate"]
+    pending_animation_ticks: DomRefCell<AnimationTickType>,
 }
 
 #[derive(JSTraceable, MallocSizeOf)]
@@ -3211,7 +3222,57 @@ impl Document {
             animations: DomRefCell::new(Animations::new()),
             dirty_root: Default::default(),
             declarative_refresh: Default::default(),
+            pending_animation_ticks: Default::default(),
+            pending_compositor_events: Default::default(),
+            mouse_move_event_index: Default::default(),
         }
+    }
+
+    /// Note a pending animation tick,
+    /// to be processed at the next `update_the_rendering` task.
+    pub fn note_pending_animation_tick(&self, tick_type: AnimationTickType) {
+        let mut pending_animation_ticks = self.pending_animation_ticks.borrow_mut();
+        pending_animation_ticks.extend(tick_type);
+    }
+
+    /// As part of a `update_the_rendering` task, tick all pending animations.
+    pub fn tick_all_animations(&self) {
+        let tick_type = mem::take(&mut *self.pending_animation_ticks.borrow_mut());
+        if tick_type.contains(AnimationTickType::REQUEST_ANIMATION_FRAME) {
+            self.run_the_animation_frame_callbacks();
+        }
+        if tick_type.contains(AnimationTickType::CSS_ANIMATIONS_AND_TRANSITIONS) {
+            self.maybe_mark_animating_nodes_as_dirty();
+        }
+    }
+
+    /// Note a pending compositor event,
+    /// to be processed at the next `update_the_rendering` task.
+    pub fn note_pending_compositor_event(&self, event: CompositorEvent) {
+        match event {
+            CompositorEvent::MouseMoveEvent { .. } => {
+                let mut index = self.mouse_move_event_index.borrow_mut();
+                if let Some(index) = *index {
+                    // Replace the pending mouse move event.
+                    let mut pending_events = self.pending_compositor_events.borrow_mut();
+                    let mouse_move_event = pending_events.get_mut(index).expect("index is valid");
+                    *mouse_move_event = event;
+                } else {
+                    self.pending_compositor_events.borrow_mut().push_back(event);
+                    *index = Some(self.pending_compositor_events.borrow().len() - 1);
+                }
+            },
+            _ => {
+                self.pending_compositor_events.borrow_mut().push_back(event);
+            },
+        }
+    }
+
+    /// Get pending compositor events, for processig within an `update_the_rendering` task.
+    pub fn get_pending_compositor_events(&self) -> VecDeque<CompositorEvent> {
+        // Reset the mouse event index.
+        *self.mouse_move_event_index.borrow_mut() = None;
+        mem::take(&mut *self.pending_compositor_events.borrow_mut())
     }
 
     pub fn set_csp_list(&self, csp_list: Option<CspList>) {
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -1724,14 +1724,15 @@ impl ScriptThread {
             // TODO: Perform pending transition operations from
             // https://drafts.csswg.org/css-view-transitions/#perform-pending-transition-operations.
 
-            // TODO(#31021: Run the update intersection observations steps from
+            // TODO(#31021): Run the update intersection observations steps from
             // https://w3c.github.io/IntersectionObserver/#run-the-update-intersection-observations-steps
 
             // TODO: Mark paint timing from https://w3c.github.io/paint-timing.
 
-            // TODO: Update the rendering: consolidate all reflow calls into one here?(#31871)
+            // TODO(#31871): Update the rendering: consolidate all reflow calls into one here?
 
-            // TODO: process top layer removals(https://drafts.csswg.org/css-position-4/#process-top-layer-removals)
+            // TODO: Process top layer removals according to
+            // https://drafts.csswg.org/css-position-4/#process-top-layer-removals.
         }
     }
 
@@ -1773,7 +1774,7 @@ impl ScriptThread {
         );
     }
 
-    /// Handle incoming control messages.
+    /// Handle incoming messages from other tasks and the task queue.
     fn handle_msgs(&self) -> bool {
         use self::MixedMessage::{
             FromConstellation, FromDevtools, FromImageCache, FromScript, FromWebGPUServer,
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -514,7 +514,7 @@ unsafe_no_jsmanaged_fields!(TaskQueue<MainThreadScriptMsg>);
 #[allow(crown::unrooted_must_root)]
 pub struct ScriptThread {
     /// <https://html.spec.whatwg.org/multipage/#last-render-opportunity-time>
-    last_render_opportunity_time: DomRefCell<u64>,
+    last_render_opportunity_time: DomRefCell<Option<Instant>>,
     /// Used to batch rendering opportunities
     has_queued_update_the_rendering_task: DomRefCell<bool>,
     /// The documents for pipelines managed by this thread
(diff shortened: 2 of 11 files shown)


components/script/task_manager.rs
```diff
@@ -59,6 +60,9 @@ pub struct TaskManager {
     #[ignore_malloc_size_of = "task sources are hard"]
     remote_event_task_source: RemoteEventTaskSource,
     #[ignore_malloc_size_of = "task sources are hard"]
+    rendering_task_source: RenderingTaskSource,
+    #[ignore_malloc_size_of = "task sources are hard"]
+    #[ignore_malloc_size_of = "task sources are hard"]
```
wusyong, line 65: Is this line duplicated?

components/script/task_source/mod.rs
```diff
@@ -36,6 +37,8 @@ pub enum TaskSourceName {
     PortMessage,
     UserInteraction,
     RemoteEvent,
+    /// https://html.spec.whatwg.org/multipage/#rendering-task-source
```
mrobinson, line 40:
```suggestion
    /// <https://html.spec.whatwg.org/multipage/#rendering-task-source>
```

gterzian:
So, "fixing" the html loop will require some heavy refactoring, and sometimes additional implementation, for each step that is done as part of the new "update the rendering" task. Currently those steps are run in an ad hoc way by handling messages from the constellation/compositor/layout immediately when they are received. In each case, we have to review what the steps are doing exactly, and perform a kind of batching when handling incoming messages, and only running the steps for all batched items inside the single "update the rendering" task. 

I have done this for:

1. The resize steps. 
2. Compositor events(focusing steps, plus other events which need to be broken down further). 

I think this shows the overall structure, and I have added TODOs for the other items. Doing this, and reviewing, in a single PR is not doable. 

Please let me know what you think. @mrobinson @jdm @wusyong (and anyone else interested). 

I'd say we can see how this affects the test suite, and try to merge it as is, with follow-ups for each TODO (See https://github.com/servo/servo/pull/31505/files#diff-2c5948584719622eef023c56cdf90e8278971f2f2eda26e7fd0485a24c32be3cR1675)

gterzian:
@jdm @mrobinson + anyone else interested:

I am going to debug the `Stable unexpected results (10)`, and I would appreciate feedback on the `Flaky unexpected result (19)`. Most of them relate to css. I assume the tests are influenced by the timing of reflows and events firing? Meaning that, by the time some event fires, for example a `resize` event, the reflow must have already happened for the test to pass? 

I think this is one open issue, for which I also opened https://github.com/servo/servo/issues/31871. Perhaps one way to address this at first is to do one reflow at the start of the "update the rendering task", and one at the end. Then later we can refine this and remove reflows, while maintaining test outcomes. Ideally we should have a single reflow at the end. But I think some of the steps fire events but expect the UI to already have been updated. 

Example is the [resize steps](https://drafts.csswg.org/cssom-view/#document-run-the-resize-steps), which reads: "If (...) had its width or height changed **since the last time these steps were run**, fire an event (...)" emphasis is mine, and I think this means we would need to reflow after having updated sizes, but before firing those events as part of the "update the rendering task". In other words, in general we would need to:

1. Logically update the UI(window size and so on).
2. Reflow to actually update the UI. 
3. Fire events. 

But 3 should happen only inside the "update the rendering task", so that it is properly ordered with respect to other stuff, like animation frame callbacks or resize observer callbacks.

mrobinson:
Looks like there are a lot of timeouts in MediaQuery tests and also another test using `load` (/css/css-values/viewport-units-after-font-load.html)  events. Could it be that there's an issue with `load` events?

gterzian:
> Looks like there are a lot of timeouts in MediaQuery tests and also another test using load (/css/css-values/viewport-units-after-font-load.html) events. Could it be that there's an issue with load events?

@mrobinson

Looking at one specific test, `/css/cssom-view/MediaQueryList-addListener-handleEvent.html`, the problem is that the iframes whose [width are changed to trigger the handler](https://github.com/servo/servo/blob/072b892706e9f39a2cb3559c610509cae00ab5cc/tests/wpt/tests/css/cssom-view/resources/matchMedia.js#L40) are not considered "fully active" by the script thread for "long enough" for the update the rendering task to run. In other words, the task is queued, then later it is throttled because the iframe is not "fully active" anymore, and thus the rendering is not updated. 

This is a new problem, because we are consolidating all logic into a task, whereas previously the logic would run when the compositor event was received(so as part of the script-thread message handling loop, but not part of the task queue which represents HTML event-loop tasks). 

I guess I can try to run "update the rendering" tasks even when the pipeline is considered not fully active, at least as a start.

gterzian:
Notes comparing wpt-run with my local testing(Mac OS):

Stable unexpected results (1):
- FAIL [expected PASS] /html/rendering/replaced-elements/embedded-content/tall-cross-domain-iframe-in-scrolled.sub.html => I get a consistent local `PASS`. 

Stable unexpected results that are known to be intermittent (22):
- `/html/semantics/embedded-content/the-iframe-element`:
     - TIMEOUT [expected OK] /html/semantics/embedded-content/the-iframe-element/iframe_sandbox_popups_nonescaping-2.html
     - OK /html/semantics/embedded-content/the-iframe-element/iframe_sandbox_popups_nonescaping-3.html with FAIL [expected NOTRUN]
     - 
- `/webmessaging/with-ports/`: same improvement in results locally
- `/html/canvas/element/manual/`: same improvement in results locally
- `/html/browsers/browsing-the-web/navigating-across-documents/initial-empty-document/load-pageshow-events-iframe-contentWindow.html` and `/html/browsers/browsing-the-web/navigating-across-documents/initial-empty-document/load-pageshow-events-window-open.html` FAIL for expected PASS. 
- '/html/browsers/history/the-history-interface/traverse_the_history_5.html' OK with FAIL(but intermittent).
- `/fetch/metadata/generated/`:  same improvement in results locally.
- `/css/cssom-view/MediaQueryList-addListener-removeListener.html`: same improvement in results locally
- `/css/css-flexbox/abspos/position-absolute-013.html`:  same improvement in results locally
- `/_mozilla/css/iframe/hide_and_show.html` also FAIL locally.

gterzian:
[`/_mozilla/css/animations/faster-reversing-of-transitions.html`](https://github.com/servo/servo/blob/df457c43c8f78d18e4e6fbc19910e35f82249b63/tests/wpt/mozilla/tests/css/animations/faster-reversing-of-transitions.html#L4) depends in the last test on ordering of ticking animations and sending animation events(which need to happen first). In the previous implementation this happened because at https://github.com/servo/servo/blob/1ed6b96684697cb2816cb7e09158de2e357bc496/components/script/script_thread.rs#L1636 we would always first send animation events, and then any animation frame is handled below with the other constellation msgs.  So reverting the last change and keeping `update animations and send events` inside the update the rendering task to avoid this FAIL(but it does introduce a new timeout noted in https://github.com/servo/servo/issues/31993)

gterzian:
This is now ready for review. Who is interested? @jdm @mrobinson @emilio

mrobinson:
> This is now ready for review. Who is interested? @jdm @mrobinson @emilio

I'm happy to review this...though I may need to walk through this with you in a video call? Is it possible to clean up the commit history first though? There are 46 commits here. If it's too much trouble to make separate commits, a single one is fine for me too.

gterzian:
@mrobinson Ok thank you, let's schedule a call via zulip. 

I'll squash it all into one, because there was a lot of back and forth and some earlier commits do not reflect what happened eventually...

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+
+                // TODO: if the focused area of doc is not a focusable area,
+                // then run the focusing steps for doc's viewport.
+                // (#31870)
+
+                // TODO: perform pending transition operations(https://drafts.csswg.org/css-view-transitions/).
+
+                // TODO: run the update intersection observations steps(#31021).
+
+                // TODO: mark paint timing(https://w3c.github.io/paint-timing).
```
mrobinson, line 1707:
```suggestion
                // TODO: Perform pending transition operations from
                // https://drafts.csswg.org/css-view-transitions/#perform-pending-transition-operations.

                // TODO(#31021: Run the update intersection observations steps from
                // https://w3c.github.io/IntersectionObserver/#run-the-update-intersection-observations-steps

                // TODO: Mark paint timing from https://w3c.github.io/paint-timing.
```

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+        if self.can_continue_running_inner() {
+            // TODO: Filter non-renderable documents.
+            // TODO: Unnecessary rendering.
+            let pipeline_and_docs: Vec<(PipelineId, DomRoot<Document>)> = self
+                .documents
+                .borrow()
+                .iter()
+                .map(|(id, document)| (id, DomRoot::from_ref(&*document)))
+                .collect();
+            for (pipeline_id, document) in pipeline_and_docs {
```
mrobinson, line 1661: This runs all steps for for every document, but the specification says has "for *doc* in *docs*" for every step. It might be worth putting a comment here that each step needs to iterate through the list of documents.

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+
+                // TODO: mark paint timing(https://w3c.github.io/paint-timing).
 
-        for (id, document) in self.documents.borrow().iter() {
-            // Only process a resize if layout is idle.
-            if let Some((size, size_type)) = document.window().steal_resize_event() {
-                resizes.push((id, size, size_type));
+                // TODO: Update the rendering: consolidate all reflow calls into one here?(#31871)
+
+                // TODO: process top layer removals(https://drafts.csswg.org/css-position-4/#process-top-layer-removals)
```
mrobinson, line 1711:
```suggestion
                // TODO(#31871): Update the rendering of each document.
                // Should all reflow calls be consolidated here into a single one?

                // TODO: Process top layer removals as described at
                // https://drafts.csswg.org/css-position-4/#process-top-layer-removals.
```

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+                )
+            },
+        };
+        let window = document.window();
+        let task_manager = window.task_manager();
+        let rendering_task_source = task_manager.rendering_task_source();
+        let canceller = task_manager.task_canceller(TaskSourceName::Rendering);
+
+        // Batch rendering ops.
+        if !*self.has_queued_update_the_rendering_task.borrow() {
```
mrobinson, line 1738: I would use an early return here.

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+                // TODO: process top layer removals(https://drafts.csswg.org/css-position-4/#process-top-layer-removals)
             }
         }
+    }
 
-        for (id, size, size_type) in resizes {
-            self.handle_event(id, ResizeEvent(size, size_type));
+    /// <https://html.spec.whatwg.org/multipage/#event-loop-processing-model:rendering-opportunity>
+    fn rendering_opportunity(&self, pipeline_id: PipelineId) {
+        *self.last_render_opportunity_time.borrow_mut() = time::precise_time_ns();
```
mrobinson, line 1718: Avoid new usage of the deprecated version of `time`, which we are trying to remove. Instead try to use `Instant` here from the Rust standard library.

See the alert about `time` here: https://github.com/servo/servo/security/dependabot

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+        // an opportunity in this script-thread.
+        let document = match self.documents.borrow().find_document(pipeline_id) {
+            Some(document) => document,
+            None => {
+                return warn!(
+                    "Trying to update the rendering for closed pipeline {}.",
+                    pipeline_id
+                )
+            },
+        };
```
mrobinson, line 1731: Again, I would use the pattern described above and avoid `return` with statements that return void as it introduces clippy warnings.

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+        let window = document.window();
+        let task_manager = window.task_manager();
+        let rendering_task_source = task_manager.rendering_task_source();
+        let canceller = task_manager.task_canceller(TaskSourceName::Rendering);
+
+        // Batch rendering ops.
```
mrobinson, line 1737:
```suggestion
```

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+                            script_thread.update_the_rendering();
+                        }
+                    })
+                }),
+                &canceller,
+            );
         }
+    }
+
+    /// Handle incoming control messages.
```
mrobinson, line 1758: Nit: I would say IPC here for a little more clarity.

```suggestion
    /// Handle incoming IPC messages.
```

components/script/script_thread.rs
```diff
@@ -1691,6 +1963,7 @@ impl ScriptThread {
             }
 
             // https://html.spec.whatwg.org/multipage/#event-loop-processing-model step 6
+            // TODO: only run after running a task(#32003).
```
mrobinson, line 1966:
```suggestion
            // TODO(#32003): A microtask checkpoint is only supposed to be performed after running a task.
```

components/script/script_thread.rs
```diff
@@ -2453,10 +2724,14 @@ impl ScriptThread {
         }
     }
 
-    fn handle_resize(&self, id: PipelineId, size: WindowSizeData, size_type: WindowSizeType) {
+    /// Batch windo resize operations into a single "update the rendering" task,
+    /// or, if a load is in progress,
+    /// set the window size directly.
```
mrobinson, line 2729: Nit: The linebreak is a bit odd here and there's a typo "windo".

```suggestion
    /// Batch window resize operations into a single "update the rendering" task,
    /// or, if a load is in progress, set the window size directly.
```

components/script/script_thread.rs
```diff
@@ -2453,10 +2724,14 @@ impl ScriptThread {
...
+    fn handle_resize_msg(&self, id: PipelineId, size: WindowSizeData, size_type: WindowSizeType) {
```
mrobinson, line 2730: This file uses both "msg" and "message" in function names, so let's standardize on "message":

```suggestion
    fn handle_resize_message(&self, id: PipelineId, size: WindowSizeData, size_type: WindowSizeType) {
```

components/script/script_thread.rs
```diff
@@ -2453,10 +2724,14 @@ impl ScriptThread {
...
         let window = self.documents.borrow().find_window(id);
         if let Some(ref window) = window {
-            window.set_resize_event(size, size_type);
+            self.rendering_opportunity(id);
```
mrobinson, line 2733: This isn't in the specification right? It should probably have a todo along the lines of:

```rust
// TODO: This is not in the specification, but tests fail without explicitly triggering a rendering opportunity
// here. This should be removed.
```

components/script/script_thread.rs
```diff
@@ -2872,6 +3147,10 @@ impl ScriptThread {
         RemoteEventTaskSource(self.remote_event_task_sender.clone(), pipeline_id)
     }
 
+    pub fn rendering_task_source(&self, pipeline_id: PipelineId) -> RenderingTaskSource {
```
mrobinson, line 3150: I think this doesn't need to be public and definitely not public outside the crate.

components/script/script_thread.rs
```diff
@@ -2982,23 +3261,21 @@ impl ScriptThread {
     pub fn handle_tick_all_animations_for_testing(id: PipelineId) {
         SCRIPT_THREAD_ROOT.with(|root| {
             let script_thread = unsafe { &*root.get().unwrap() };
-            script_thread
-                .handle_tick_all_animations(id, AnimationTickType::CSS_ANIMATIONS_AND_TRANSITIONS);
+            let document = match script_thread.documents.borrow().find_document(id) {
+                Some(document) => document,
+                None => return warn!("Animation tick for tests for closed pipeline {}.", id),
+            };
```
mrobinson, line 3267:
```suggestion
            let Some(document) = match script_thread.documents.borrow().find_document(id) else {
                warn!("Animation tick for tests for closed pipeline {id}.");
                return;
            };
```

components/script/dom/document.rs
```diff
@@ -3211,9 +3222,59 @@ impl Document {
...
+        if tick_type.contains(AnimationTickType::REQUEST_ANIMATION_FRAME) {
+            self.run_the_animation_frame_callbacks();
+        }
+        if tick_type.contains(AnimationTickType::CSS_ANIMATIONS_AND_TRANSITIONS) {
+            self.maybe_mark_animating_nodes_as_dirty();
+        }
+    }
+
+    /// Note a pending compositor event,
+    /// to be processed at the next `update_the_rendering` task.
```
mrobinson, line 3250:
```suggestion
    /// Note a pending compositor event, to be processed at the next `update_the_rendering` task.
```

components/script/dom/document.rs
```diff
@@ -3211,9 +3222,59 @@ impl Document {
...
             dirty_root: Default::default(),
             declarative_refresh: Default::default(),
+            pending_animation_ticks: Default::default(),
+            pending_compositor_events: Default::default(),
+            mouse_move_event_index: Default::default(),
+        }
+    }
+
+    /// Note a pending animation tick,
+    /// to be processed at the next `update_the_rendering` task.
```
mrobinson, line 3232:
```suggestion
    /// Note a pending animation tick, to be processed at the next `update_the_rendering` task.
```

components/script/script_thread.rs
```diff
@@ -2982,23 +3261,21 @@ impl ScriptThread {
...
             Some(document) => document,
             None => return warn!("Message sent to closed pipeline {}.", id),
         };
-        if tick_type.contains(AnimationTickType::REQUEST_ANIMATION_FRAME) {
-            document.run_the_animation_frame_callbacks();
-        }
-        if tick_type.contains(AnimationTickType::CSS_ANIMATIONS_AND_TRANSITIONS) {
-            document.maybe_mark_animating_nodes_as_dirty();
-        }
+        document.tick_all_animations();
```
mrobinson, line 3278: This function can be removed entirely now and the code that calls it can simply call `document.tick_all_animations()` which will make it impossible to not find the document.

components/script/script_thread.rs
```diff
@@ -3470,186 +3748,15 @@ impl ScriptThread {
...
-            Some(win) => win,
-            None => {
-                return warn!(
-                    "Compositor event sent to a pipeline that already exited {}.",
-                    pipeline_id
-                )
-            },
+        let document = match self.documents.borrow().find_document(pipeline_id) {
+            Some(document) => document,
+            None => return warn!("Compositor event sent to closed pipeline {}.", pipeline_id),
```
mrobinson, line 3756:
```suggestion
        let Some(document) = match self.documents.borrow().find_document(pipeline_id) else {
             warn!("Compositor event sent to closed pipeline {pipeline_id}.");\
             return;
        };
```

components/script/task_queue.rs
```diff
@@ -186,8 +192,9 @@ impl<T: QueuedTaskConversion> TaskQueue<T> {
         self.msg_queue.borrow_mut().pop_front().ok_or(())
     }
 
-    /// Same as recv.
+    /// Retry to take tasks, and then recv.
     pub fn try_recv(&self) -> Result<T, ()> {
+        self.take_tasks(T::wake_up_msg());
         self.recv()
     }
```
mrobinson, line 199:
```suggestion
    /// Take all tasks again and then run `recv()`.
    pub(crate) fn take_tasks_and_recv(&self) -> Result<T, ()> {
        self.take_tasks(T::wake_up_msg());
        self.recv()
    }
```

mrobinson (changes requested):
This is a nice change! I'm happy to see the event loop becoming more standards compliant. I have some suggestions (mainly nits), but in general this is great.

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+        if document.window().Closed() {
+            return warn!(
+                "Compositor event sent to a pipeline with a closed window {}.",
+                pipeline_id
+            );
+        }
+        let window = document.window();
+        ScriptThread::set_user_interacting(true);
+        let _realm = enter_realm(document.window());
+        for event in document.get_pending_compositor_events().into_iter() {
```
mrobinson, line 1512:
```suggestion
        for event in document.get_pending_compositor_events().into_iter() {
```

If you are just calling `into_iter()` on this then maybe `Vec` is a better choice than `VecDeque`?

components/script/task_queue.rs
mrobinson: I know we had a conversation about this where I was a bit worried that this might allow starving the rest of the script thread, but as you say, you wrote this code and I trust you with this one.

components/script/script_thread.rs
gterzian: The task queue is not IPC.

components/script/script_thread.rs
gterzian: The spec doesn't specify when exactly a rendering opportunity should be noted.

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,292 @@ impl ScriptThread {
...
+                            } else {
+                                EventResult::DefaultPrevented
+                            };
+                            let message = ScriptMsg::TouchEventProcessed(result);
+                            self.script_sender.send((pipeline_id, message)).unwrap();
+                        },
+                        _ => {
+                            // TODO: Calling preventDefault on a touchup event should prevent clicks.
+                        },
+                    }
```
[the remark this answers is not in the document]
gterzian: I'm not sure about this one given the TODOs.

gterzian:
@mrobinson Thanks for the review, I've addressed all comments(or commented).

components/script/script_thread.rs
```diff
@@ -1471,28 +1476,310 @@ impl ScriptThread {
...
+
+            // TODO(#31006): Implement the resize observer steps.
+
+            // TODO(#31870): Implement step 17: if the focused area of doc is not a focusable area,
+            // then run the focusing steps for document's viewport.
+
+            // TODO: Perform pending transition operations from
+            // https://drafts.csswg.org/css-view-transitions/#perform-pending-transition-operations.
+
+            // TODO(#31021: Run the update intersection observations steps from
```
mrobinson, line 1727: Nit. This is probably my fault, sorry. I'll fix this one.

```suggestion
            // TODO(#31021): Run the update intersection observations steps from
```

mrobinson (approved):

mrobinson:
Apologies for the big delay here. I was on PTO. I will send this to the MQ now.

PR #31506, opened 2024-03-05
Fix intrinsic sizing of tables

Loirooriol:
---


components/layout_2020/table/layout.rs
```diff
@@ -689,17 +669,26 @@ impl<'a> TableLayout<'a> {
...
             },
         };
 
+        let border_spacing = self.table.border_spacing();
+        let inline_border_spacing = if self.table.size.width > 0 {
+            border_spacing.inline * (self.table.size.width as i32 + 1)
+        } else {
+            Au::zero()
+        };
+
```
mrobinson, line 691: Perhaps this could be moved to a helper method that properly returns the total inline and block border spacing in order to put all this logic into the same place.

mrobinson (approved):
Looks great!

PR #31512, opened 2024-03-05
Fix several clippy warnings in components/bluetooth

mnaibei:
Fixes several clippy warnings in the components/bluetooth directory. It is part of https://github.com/servo/servo/issues/31500.

---

### DIFF
--- a/components/bluetooth/lib.rs
+++ b/components/bluetooth/lib.rs
@@ -120,7 +120,7 @@ fn matches_filter(device: &BluetoothDevice, filter: &BluetoothScanfilter) -> boo
     if !filter.get_services().is_empty() {
         if let Ok(device_uuids) = device.get_uuids() {
             for service in filter.get_services() {
-                if device_uuids.iter().find(|x| x == &service).is_none() {
+                if device_uuids.iter().any(|x| &x == &service) {
                     return false;
                 }
             }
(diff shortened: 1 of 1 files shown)


components/bluetooth/lib.rs
```diff
@@ -120,20 +120,20 @@ fn matches_filter(device: &BluetoothDevice, filter: &BluetoothScanfilter) -> boo
     if !filter.get_services().is_empty() {
         if let Ok(device_uuids) = device.get_uuids() {
             for service in filter.get_services() {
-                if device_uuids.iter().find(|x| x == &service).is_none() {
+                if device_uuids.iter().any(|x| &x == &service) {
```
mrobinson, line 123: I think this reverses the meaning of this if-statement. In the previous version it was checking if there were no results from the find and in this version it is true if there are any.

mrobinson (changes requested):

components/hyper_serde/lib.rs
```diff
@@ -138,7 +138,7 @@ impl<T> De<T> {
     /// Returns a new `De` wrapper
     #[inline(always)]
     pub fn new(v: T) -> Self {
-        De { v: v }
+        De { v }
```
mrobinson, line 141: It looks like you have also included your changes from your other PR. Please remove these changes from the PR.

components/bluetooth/lib.rs
mnaibei: @mrobinson I have updated this, please have a look

mrobinson (approved):
Thanks! This looks great!

PR #31513, opened 2024-03-05
Fix the measures of a table cell

Loirooriol:
The spec doesn't make much sense, this seems to match Gecko and Blink.

---


mrobinson (approved):
Seems pretty reasonable to me!

PR #31508, opened 2024-03-05
Fix several clippy warnings in components/hyper_serde

mnaibei:
Fixes several clippy warnings in the components/hyper_serde directory. It is part of https://github.com/servo/servo/issues/31500.

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes fix #31500

- [x] These changes do not require tests because they do not modify functionality, they only fix lint errors.

PR #31514, opened 2024-03-05
clippy: Fix more warnings

eerii:
Followup to #31501, fixes more clippy warnings. It specifically avoids the ones fixed in #31508, #31512 and #31549 to avoid conflicts.

Until now, it fixes most warnings until `net` is compiled, **excluding**:
- `result_unit_err`: use a custom `Error` type instead of `Result<_, ()>`
- `missing_safety_doc`: unsafe function's docs miss `# Safety` section

It also hopefully removes all clippy **errors** until `script` is compiled, so more warnings can be caught.

The next steps to tackle would be:
- Fix the remaining `net` warnings
- Create proper `Error` types for the first warnings
- Write safety documentation (probably someone that is familiar with the unsafe code in the project)
- Work with `canvas`, `webdriver_server` and a few small others before tackling the big ones, which are `script`, `layout` and `layout2020`. `script` alone has 27358 warnings and 433 errors, so it will be the biggest undertaking

---

### DIFF
--- a/components/gfx/text/glyph.rs
+++ b/components/gfx/text/glyph.rs
@@ -28,7 +28,7 @@ pub struct GlyphEntry {
 
 impl GlyphEntry {
     fn new(value: u32) -> GlyphEntry {
-        GlyphEntry { value: value }
+        GlyphEntry { value }
     }
 
     fn initial() -> GlyphEntry {
@@ -157,9 +157,9 @@ struct DetailedGlyph {
 impl DetailedGlyph {
     fn new(id: GlyphId, advance: Au, offset: Point2D<Au>) -> DetailedGlyph {
         DetailedGlyph {
-            id: id,
-            advance: advance,
-            offset: offset,
+            id,
+            advance,
+            offset,
         }
     }
 }
@@ -210,7 +210,7 @@ impl<'a> DetailedGlyphStore {
 
     fn add_detailed_glyphs_for_entry(&mut self, entry_offset: ByteIndex, glyphs: &[DetailedGlyph]) {
         let entry = DetailedGlyphRecord {
-            entry_offset: entry_offset,
+            entry_offset,
             detail_offset: self.detail_buffer.len(),
         };
 
@@ -245,7 +245,7 @@ impl<'a> DetailedGlyphStore {
         assert!(self.lookup_is_sorted);
 
         let key = DetailedGlyphRecord {
-            entry_offset: entry_offset,
+            entry_offset,
             detail_offset: 0, // unused
         };
 
@@ -268,7 +268,7 @@ impl<'a> DetailedGlyphStore {
         assert!(self.lookup_is_sorted);
 
         let key = DetailedGlyphRecord {
-            entry_offset: entry_offset,
+            entry_offset,
             detail_offset: 0, // unused
         };
 
@@ -329,11 +329,11 @@ impl GlyphData {
         ligature_start: bool,
     ) -> GlyphData {
         GlyphData {
-            id: id,
-            advance: advance,
+            id,
+            advance,
             offset: offset.unwrap_or(Point2D::zero()),
-            cluster_start: cluster_start,
-            ligature_start: ligature_start,
+            cluster_start,
+            ligature_start,
         }
     }
 }
@@ -455,8 +455,8 @@ impl<'a> GlyphStore {
             total_advance: Au(0),
             total_word_separators: 0,
             has_detailed_glyphs: false,
-            is_whitespace: is_whitespace,
-            is_rtl: is_rtl,
+            is_whitespace,
+            is_rtl,
         }
     }
 
--- a/components/gfx/text/glyph.rs
+++ b/components/gfx/text/glyph.rs
@@ -40,7 +40,7 @@ impl GlyphEntry {
         assert!(is_simple_glyph_id(id));
         assert!(is_simple_advance(advance));
 
-        let id_mask = id as u32;
+        let id_mask = id;
         let Au(advance) = advance;
         let advance_mask = (advance as u32) << GLYPH_ADVANCE_SHIFT;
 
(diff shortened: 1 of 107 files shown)


mnaibei:
Hi @eerii., I wanted to do some more clippy fixes today and I was just wondering should I wait for you to merge this pr first?

eerii:
Hi @mnaibei, there's plenty of warnings left! You can look into the file change list for this pr to target another files, or you could clone this, do your changes and then cherry pick them. I'd suggest looking into the files from `components/net` that are not yet changed as they seem to have the most warnings at the moment. I'll make sure to check your fork before I get back to this in the afternoon to work on different parts. Thanks! :D

components/gfx/text/glyph.rs
```diff
@@ -174,7 +174,7 @@ struct DetailedGlyphRecord {
 
 impl PartialOrd for DetailedGlyphRecord {
     fn partial_cmp(&self, other: &DetailedGlyphRecord) -> Option<Ordering> {
-        self.entry_offset.partial_cmp(&other.entry_offset)
+        Some(self.cmp(other))
```
mrobinson, line 177: Won't this also compare `details_offset`? The previous code was only comparing `entry_offset`.

components/media/lib.rs
```diff
@@ -20,12 +20,14 @@ pub use crate::media_channel::glplayer_channel;
 use crate::media_channel::{GLPlayerChan, GLPlayerPipeline, GLPlayerReceiver, GLPlayerSender};
 use crate::media_thread::GLPlayerThread;
 
+type GLPlayerSenderType = GLPlayerSender<(u32, Size2D<i32>, usize)>;
+
```
mrobinson, line 24: I thin in this case the original is a bit better as it's hard to understand what the sender sends now.

components/net/http_cache.rs
```diff
@@ -246,24 +244,22 @@ fn get_response_expiry(response: &Response) -> Duration {
...
-            result
+            }
         } else {
             max_heuristic
         };
         if is_cacheable_by_default(*code) {
             // Status codes that are cacheable by default can use heuristics to determine freshness.
             return heuristic_freshness;
-        } else {
-            // Other status codes can only use heuristic freshness if the public cache directive is present.
```
mrobinson, line 262: This comment is now gone. Please re-add it.

components/net/http_cache.rs
```diff
@@ -762,9 +762,9 @@ impl HttpCache {
         done_chan: &mut DoneChannel,
     ) -> Option<Response> {
         assert_eq!(response.status.map(|s| s.0), Some(StatusCode::NOT_MODIFIED));
-        let entry_key = CacheKey::new(&request);
+        let entry_key = CacheKey::new(request);
         if let Some(cached_resources) = self.entries.get_mut(&entry_key) {
-            for cached_resource in cached_resources.iter_mut() {
+            if let Some(cached_resource) = cached_resources.iter_mut().next() {
```
mrobinson, line 767: This used to iterate all of the cached resources, but this version only iterates the first.

components/net/mime_classifier.rs
```diff
@@ -436,8 +438,8 @@ impl BinaryOrPlaintextClassifier {
         } else if data.iter().any(|&x| {
             x <= 0x08u8 ||
                 x == 0x0Bu8 ||
-                (x >= 0x0Eu8 && x <= 0x1Au8) ||
-                (x >= 0x1Cu8 && x <= 0x1Fu8)
+                (0x0Eu8..=0x1Au8).contains(&x) ||
+                (0x1Cu8..=0x1Fu8).contains(&x)
```
mrobinson, line 442: There is now an off by one error in the `contains(...)` call as Rust ranges are exclusive on the end value.

components/shared/net/request.rs
```diff
@@ -208,7 +208,11 @@ impl RequestBody {
     }
 
     pub fn len(&self) -> Option<usize> {
-        self.total_bytes.clone()
+        self.total_bytes
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.total_bytes == Some(0)
```
mrobinson, line 215: I don't know about this `is_empty`. What should the return value be when `total_bytes` is None?

mrobinson (changes requested):
Nice work.

This change has gotten too big though. Please limit the the number of files changed in these kind of PRs to something like 10. If you don't mind split the changes here across several pull requests.

components/gfx/text/glyph.rs
eerii: I think it uses the `Ord` implementation bellow that only compares `entry_offset`, right? Maybe I can move it down so it is clear that it is using `Ord`.

components/net/http_cache.rs
eerii: The warning in this case was that this for only ran once, because it ends with a return, so there may be some logic missing if the intention was to iterate all `cached_resources`.

components/net/mime_classifier.rs
eerii: I think the `start..=end` syntax represents an [inclusive range](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html), but I may be misinterpreting it.

components/shared/net/request.rs
eerii: You are right, I think it makes sense that it returns false if it is None or if it is Some(0), what do you think?

components/net/mime_classifier.rs
mrobinson: Oh, fascinating. I didn't know about that syntax!

components/net/http_cache.rs
mrobinson: Thanks. I missed this in my review.

components/shared/net/request.rs
mrobinson: Hrm. I think maybe if the function will ultimately be unused it might better to simply not add it, since it won't have any test coverage and I'm not sure what it should mean in this case. What do you think?

eerii: I think that makes sense, I can simply add an allow so clippy doesn't trigger the warning with an explanation comment, and if in the future it is needed then it can be implemented.

eerii: Actually I just saw that #31551 is specifically for this file so its probably best handled there, and I'll just keep this out when I make the new prs

eerii:
All split into modules, I'm going to close this.

PR #31465, opened 2024-02-29
Update gstreamer to 0.22

sagudev:
Companion to https://github.com/servo/media/pull/417

This should hopefully fix doc builds

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #31515, opened 2024-03-06
Fix build failure on NixOS

syvb:
Fixes a build error when building on NixOS. This build error was introduced by #31465, which changed filterlock without also updating the `vendorTarball` hash in `etc/shell.nix`. (filterlock is not a part of the main Servo workspace (it has its own lockfile and target directory), since it has to be executed before any other Rust code can be compiled on NixOS)

```
$ nix-shell etc/shell.nix 
...
Running phase: buildPhase
error: failed to select a version for the requirement `toml = "^0.8.9"`
candidate versions found which didn't match: 0.8.8
location searched: directory source `/build/filterlock/vendor` (which is replacing registry `crates-io`)
required by package `filterlock v0.1.0 (/build/filterlock)`
perhaps a crate was updated and forgotten to be re-vendored?
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
```
---

- [X] These changes do not require tests because none of the Nix configuration is tested in CI

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -96,7 +96,7 @@ stdenv.mkDerivation (androidEnvironment // rec {
     (let
       vendorTarball = rustPlatform.fetchCargoTarball {
         src = ../support/filterlock;
-        hash = "sha256-/kJNDtmv2uI7Qlmpi3DMWSw88rzEJSbroO0/QrgQrSc=";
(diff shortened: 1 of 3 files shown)


delan (approved):
Thanks!

PR #31521, opened 2024-03-06
Move convert_* functions into gpuconvert.rs

### LINKED ISSUE #31104: Create convert mechanisem for WGPU <=> webgpu
Currently `convert_*` functions are spread across multiple gpu*.rs files. It would be nice if we have unified mechanism for this and all conversions in same file.

eerii:
Centralizes all conversion functions between wgpu/webgpu bindings in one module, `components/script/dom/gpuconvert.rs`.

---
- [x] These changes partially fix #31104
- [x] These changes do not require tests because they don't introduce any new functionality

### DIFF
--- a/components/script/dom/gpucommandencoder.rs
+++ b/components/script/dom/gpucommandencoder.rs
@@ -10,16 +10,12 @@ use dom_struct::dom_struct;
 use webgpu::wgpu::command as wgpu_com;
 use webgpu::{self, wgt, WebGPU, WebGPURequest};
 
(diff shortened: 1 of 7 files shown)


sagudev (approved):
Great stuff, thanks for all the work.


This is only part of the issue (to put all convert files into one place), but I still haven't designed mechanism (macro+traits) for two way conversion.

eerii:
Oh that makes a ton of sense, two way conversion would be way nicer! Thanks for reviewing and changing the description to match :D

PR #31523, opened 2024-03-06
compositor: Improve the way we wait for frames

mrobinson:
In the newest version of WebRender it will be harder to make the
distinction between frame queued for scrolling and other kinds of
pending frames. This change makes it so that we queue frames for both
kinds of changes the same way and keeps a counting of pending frames.
This is conceptually a lot simpler.

In addition, do queue a composite even when recomposite isn't necessary
for a WebRender frame when there are active requestAnimationFrame
callbacks. Doing a composite is what triggers the callbacks to actually
run in the script thread! I believe this was a bug, but the WebRender
upgrade made it much more obvious.

These changes are in preparation for the WebRender upgrade.

---

- [x] These changes do not require tests because there is no reliable behavior change, but we notice less flakes when running tests.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -13,9 +13,8 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use canvas::canvas_paint_thread::ImageUpdate;
 use compositing_traits::{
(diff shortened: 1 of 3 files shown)


mrobinson (commented):
Remove spurious println

atbrakhi (approved):
Thank you for walking me through it. It looks good

PR #31535, opened 2024-03-06
layout: Add documentation about table layout

mrobinson:

---

### DIFF
--- a/components/layout_2020/table/mod.rs
+++ b/components/layout_2020/table/mod.rs
@@ -2,12 +2,67 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-//! HTML Tables (╯°□°)╯︵ ┻━┻.
+//! # HTML Tables (╯°□°)╯︵ ┻━┻
 //!
-//! See <https://html.spec.whatwg.org/multipage/table-processing-model> and
-//! <https://drafts.csswg.org/css-tables>.  This is heavily based on the latter specification, but
-//! note that it is still an Editor's Draft, so there is no guarantee that what is implemented here
-//! matches other browsers or the current specification.
+//! This implementation is based on the [table section of the HTML 5 Specification][1],
+//! the draft [CSS Table Module Level! 3][2] and the [LayoutNG implementation of tables][3] in Blink.
+//! In general, the draft specification differs greatly from what other browsers do, so we
+//! generally follow LayoutNG when in question.
+//!
+//! [1]: https://html.spec.whatwg.org/multipage/#tables
+//! [2]: https://drafts.csswg.org/css-tables
+//! [3]: https://source.chromium.org/chromium/chromium/src/third_party/+/main:blink/renderer/core/layout/table
+//!
+//! Table layout is divided into two phases:
+//!
+//! 1. Box Tree Construction
+//! 2. Fragment Tree Construction
+//!
+//! ## Box Tree Construction
+//!
+//! During box tree construction, table layout (`construct.rs`) will traverse the DOM and construct
+//! the basic box tree representation of a table, using the structs defined in this file ([`Table`],
+//! [`TableTrackGroup`], [`TableTrack`], etc). When processing the DOM, elements are handled
+//! differently depending on their `display` value. For instance, an element with `display:
+//! table-cell` is treated as a table cell. HTML table elements like `<table>` and `<td>` are
+//! assigned the corresponding display value from the user agent stylesheet.
+//!
+//! Every [`Table`] holds an array of [`TableSlot`]. A [`TableSlot`] represents either a cell, a cell
+//! location occupied by a cell originating from another place in the table, or is empty. In
+//! addition, when there are table model errors, a slot may spanned by more than one cell.
+//!
+//! During processing, the box tree construction agorithm will also fix up table structure, for
+//! instance, creating anonymous rows for lone table cells and putting non-table content into
+//! anonymous cells. In addition, flow layout will collect table elements outside of tables and create
+//! anonymous tables for them.
+//!
+//! After processing, box tree construction does a fix up pass on tables, converting rowspan=0 into
+//! definite rowspan values and making sure that rowspan and celspan values are not larger than the
+//! table itself. Finally, row groups may be reordered to enforce the fact that the first `<thead>`
+//! comes before `<tbody>` which comes before the first `<tfoot>`.
+//!
+//! ## Fragment Tree Construction
+//!
+//! Fragment tree construction involves calculating the size and positioning of all table elements,
+//! given their style, content, and cell and row spans. This happens both during intrinsic inline
+//! size computation as well as layout into Fragments. In both of these cases, measurement and
+//! layout is done by [`layout::TableLayout`], though for intrinsic size computation only a partial
+//! layout is done.
+//!
+//! In general, we follow the following steps when laying out table content:
+//!
+//! 1. Compute track constrainedness and has originating cells
+//! 2. Compute cell measures
+//! 3. Compute column measures
+//! 4. Compute intrinsic inline sizes for columns and the table
+//! 5. Compute the final table inline size
+//! 6. Distribute size to columns
+//! 7. Do first pass cell layout
+//! 8. Do row layout
+//! 9. Compute table height and final row sizes
+//! 10. Create fragments for table elements (columns, column groups, rows, row groups, cells)
+//!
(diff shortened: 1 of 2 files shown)


components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
-//! <https://drafts.csswg.org/css-tables>.  This is heavily based on the latter specification, but
-//! note that it is still an Editor's Draft, so there is no guarantee that what is implemented here
-//! matches other browsers or the current specification.
+//! This implementation is based on the [table section of the HTML 5
+//! Specification](https://html.spec.whatwg.org/multipage/#tables), the draft [CSS Table Module Level
+//! 3](https://drafts.csswg.org/css-tables) and the [LayoutNG implementation of
+//! tables](https://source.chromium.org/chromium/chromium/src/third_party/+/main:blink/renderer/core/layout/table)
+//! in Blink. In general, the draft specification differs greatly from what other browsers do, so we
+//! generally follow LayoutNG when in question.
+//!
```
Loirooriol, line 13: The line breaking looks weird due to the links. This seems to work:

```suggestion
//! This implementation is based on the [table section of the HTML 5 Specification][1],
//! the draft [CSS Table Module Level! 3][2] and the [LayoutNG implementation of tables][3] in Blink.
//! In general, the draft specification differs greatly from what other browsers do, so we
//! generally follow LayoutNG when in question.
//!
//! [1]: https://html.spec.whatwg.org/multipage/#tables
//! [2]: https://drafts.csswg.org/css-tables
//! [3]: https://source.chromium.org/chromium/chromium/src/third_party/+/main:blink/renderer/core/layout/table
//!
```

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
+//! 2. Fragment Tree Construction
+//!
+//! ## Box Tree Construction
+//!
+//! During box tree construction, table layout (`construct.rs`) will traverse the DOM and construct
+//! the basic box tree representation of a table, using the structs defined in this file ([`Table`],
+//! [`TableTrackGroup`], [`TableTrack`], etc). When processing the DOM, elements are handled
+//! differently depending on their `display` value. For instance, an element with `display:
+//! table-cell` is treated as a table cell. These display values are automatically set for HTML
+//! elements by the user agent stylesheet.
```
Loirooriol, line 26: This can be understood as all HTML elements receiving some display in the UA sheet, but e.g. `<span>` just gets the initial `display: inline` without needing the UA stylesheet.

Maybe something like:

> Tabular HTML elements like `<table>` or `<td>` get assigned the corresponding display value in the user agent stylesheet.

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
+//!
+//! Inside each table is an array of [`TableSlot`]. A table slot represents either a cell, a cell
```
Loirooriol, line 28: This doesn't sound grammatically right to me.

"Inside each table there is an array" or "Each table has an array`.

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
+//! location occupied by a cell originating from another place in the table, or is empty. In
+//! addition, when there are table model errors, a slot may spanned by more than one cell. In those
+//! cases, the top-left-most cell wins.
```
Loirooriol, line 31: It's not much clear what "top-left-most cell wins" means.

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
+//!
+//! During processing, the box tree construction agorithm will also fix up table structure, for
+//! instance, creating anonymous rows for lone table cells and putting other anonymous content into
+//! cells. In addition, flow layout will collect table elements outside of tables and create
```
Loirooriol, line 35: The content that is put into cells isn't typically anonymous.

```suggestion
//! instance, creating anonymous rows for lone table cells and putting non-tabular content into
//! anonymous cells. In addition, flow layout will collect table elements outside of tables and create
```

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,66 @@
...
+//! After processing, box tree construction does a fix up pass on tables, converting rowspan=0 into
+//! definite rowspan values and making sure that rowspan and celspan values are not larger than the
+//! table itself. Finally, row groups may be reordered to enforce the fact that the first `<thead>`
+//! comes before `<tbody>` which comes before the first `<tfoot>`.
+//!
+//! ## Fragment Tree Construction
+//!
+//! Fragment tree construction involves calculating the size and positioning of all table elements,
+//! given their style, intrinsic sizes, contents, and cell and row spans. This happens both during
+//! layout into Fragments as well as intrinsic inline size computation. In both of these cases,
```
Loirooriol, line 47: intrinsic sizes depend on the contents, so we may omit the former, especially since it appears on the next line.

And intrinsic sizing happens first, so maybe mention it first

```suggestion
//! given their style, contents, and cell and row spans. This happens both during
//! intrinsic inline size computation as well as layout into Fragments. In both of these cases,
```

components/layout_2020/table/mod.rs
mrobinson: This is grammatically correct actually. :D I've modified this to be a bit less awkward though. Now it says: 

```md
 Every [`Table`] holds an array of [`TableSlot`].
```

components/layout_2020/table/mod.rs
mrobinson: I've changed this to: 

```rust
//! addition, when there are table model errors, a slot may spanned by more than one cell. In those
//! cases, the cell that comes first in DOM order will cover the slot.
```

mrobinson:
@Loirooriol Thanks for the review. I've updated the change according to your comments.

mrobinson:
One change I've made to your suggestions is to use phrases like "table elements" versus "tabular elements" as I think "tabular" suggests more of a "tableness" versus `<table>` and friends. Obviously "table elements" is a *little* bit ambiguous because it could mean `<table>` or that and also `<tr>`, `<td>`, etc, but I think it's obvious from context.

components/layout_2020/table/mod.rs
Loirooriol: OK that sounds better, thanks.

components/layout_2020/table/mod.rs
Loirooriol: But isn't the slot covered by all the cells? Like

```html
<table border="1">
<tr>
  <td>aaaa</td>
  <td rowspan="2" style="background: rgba(255, 0, 255, 0.5)">bbb</td>
</tr><tr>
  <td colspan="2" style="background: rgba(0, 255, 255, 0.5)">ccc ccc ccc</td>
</tr>
</table>
```

components/layout_2020/table/mod.rs
```diff
@@ -2,12 +2,68 @@
...
+//!
+//! After processing, box tree construction does a fix up pass on tables, converting rowspan=0 into
+//! definite rowspan values and making sure that rowspan and celspan values are not larger than the
+//! table itself. Finally, row groups may be reordered to enforce the fact that the first `<thead>`
+//! comes before `<tbody>` which comes before the first `<tfoot>`.
+//!
+//! ## Fragment Tree Construction
+//!
+//! Fragment tree construction involves calculating the size and positioning of all table elements,
+//! given their style, content, and cell and row spans. This happens both during
```
Loirooriol, line 48: This line now looks short,  I would adjust the line wrapping

Loirooriol (approved):

components/layout_2020/table/mod.rs
mrobinson: Huh. Interesting. From https://html.spec.whatwg.org/multipage/tables.html#table-model-error:

"If any of the slots involved already had a [cell](https://html.spec.whatwg.org/multipage/tables.html#concept-cell) covering them, then this is a [table model error](https://html.spec.whatwg.org/multipage/tables.html#table-model-error). Those slots now have two cells overlapping."

I'm not sure how to interpret this. It looks like our layout differs from Firefox in that case.

mrobinson: I've just removed this sentence for now.

Loirooriol: The difference from Firefox is that our intrinsic sizing is broken for `colspan="2"` so the columns are too narrow, but the overlapping seems the same.

mrobinson: Sorry, I also noticed that the baseline was different, but you're right that it isn't related to the overlapping.

PR #31538, opened 2024-03-07
ci: Fix packaging step on Windows

mrobinson:
WiX doesn't seem to be on the path any longer, so it must be added
manually as a temporary workaround.

---

- [x] These changes do not require tests because they fix the CI.


delan (approved):

mrobinson:
I'm going to skip the merge queue for this change because it is blocking landing anything else and this is on top of incredibly the flakiness we are experience in requestAnimationFrame tests.

PR #31549, opened 2024-03-07
Fix clippy warnings in `components/rand`

sandeepB3:
---

- [x] These changes fix #31500
- [x] These changes do not require tests because they do not modify functionality, it only adds a `Default` implementation for `ServoRng`

### DIFF
--- a/components/devtools/lib.rs
+++ b/components/devtools/lib.rs
@@ -144,7 +144,7 @@ fn run_server(
     });
 
     // A token shared with the embedder to bypass permission prompt.
-    let token = format!("{:X}", servo_rand::ServoRng::new().next_u32());
+    let token = format!("{:X}", servo_rand::ServoRng::default().next_u32());
 
     let port = bound.as_ref().map(|(_, port)| *port).ok_or(());
     embedder.send((None, EmbedderMsg::OnDevtoolsStarted(port, token.clone())));
--- a/components/rand/lib.rs
+++ b/components/rand/lib.rs
@@ -96,17 +96,11 @@ impl ServoRng {
 }
 
 impl Default for ServoRng {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl ServoRng {
     /// Create an auto-reseeding instance of `ServoRng`.
     ///
     /// This uses the shared `OsRng`, so avoids consuming
     /// a file descriptor.
-    pub fn new() -> ServoRng {
+    fn default() -> Self {
         trace!("Creating new ServoRng.");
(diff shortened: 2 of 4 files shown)


mrobinson (changes requested):
This adds a default implementation for `ServoRng`, but the implementation is not used. It should be used somewhere and the clippy warning should likely give you a hint as to where. Perhaps it is about replacing `new()` with `default()` in which case you should find all the callers of `new()` and have them use `default()`.

sandeepB3:
@mrobinson, I have made the required changes by replacing new() with default() in line 109, but is it applicable in line 165 as well ? As we have a Default implementation for ServoReseeder in line 147

mrobinson:
> @mrobinson, I have made the required changes by replacing new() with default() in line 109, but is it applicable in line 165 as well? As we have a Default implementation for ServoReseeder in line 147.

You can determine if this it's adequate by moving the contents of the `new()` function into `default()`, removing `new()` and then trying to recompile.

sandeepB3:
> > @mrobinson, I have made the required changes by replacing new() with default() in line 109, but is it applicable in line 165 as well? As we have a Default implementation for ServoReseeder in line 147.
> 
> You can determine if this it's adequate by moving the contents of the `new()` function into `default()`, removing `new()` and then trying to recompile.

Yes understood, so wherever `ServoRng::new()` is implemented is to be replaced with `ServoRng::default()`, I have made the required changes and both `./mach build -d` and `./mach test-tidy`, does not report any error.

Screenshot 2024-03-07 at 7 13 28 PM

mrobinson:
> > > @mrobinson, I have made the required changes by replacing new() with default() in line 109, but is it applicable in line 165 as well? As we have a Default implementation for ServoReseeder in line 147.
> > 
> > 
> > You can determine if this it's adequate by moving the contents of the `new()` function into `default()`, removing `new()` and then trying to recompile.
> 
> Yes understood, so wherever `ServoRng::new()` is implemented, it is to be replaced with `ServoRng::default()`, I have made the required changes and both `./mach build -d` and `./mach test-tidy`, does not report any error.
> Screenshot 2024-03-07 at 7 13 28 PM

Thanks. I think you'll still need to move the contents of the `new()` method into the `default()` method and then completely remove the `new()` method though.

sandeepB3:
@mrobinson, I have made the required changes, so that now calling ServoRng::default() will now perform the same operations that were originally in the new() method. Subsequently I have replaced ServoRng::new() with ServoRng::default(), wherever `./mach build -d` test failed. Do let me know if there would be any other modifications.

mrobinson (approved):
Thanks!

PR #31561, opened 2024-03-07
compositor: Limit animation tick interval when in WPT mode

mrobinson:
When exiting after load or dumping a screenshot, limit the speed that
the compositor sends tick events to the constellation (and on to the
script thread). Previously, these tick messages were sent as fast as the
winit event loop (that's really fast!), until the document was ready to
for its screenshot. rAf is meant to run at more or less the same
frequency as the refresh rate of the screen.

On my system, the massive amount of ticks was overwhelming the router
thread, preventing fonts from loading properly. This change is an attempt
to make tests with requestAnimationFrame less flaky.

---

- [x] These changes do not require tests because this is an attempt to reduce the number of flaky failures.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -9,7 +9,7 @@ use std::fs::{create_dir_all, File};
 use std::io::Write;
 use std::num::NonZeroU32;
 use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


jdm (approved):
Makes sense!

PR #31564, opened 2024-03-07
clippy: fix warnings in components/net

eerii:
Split from #31514, fixes some clippy warnings in `components/net`.

---


components/net/http_loader.rs
```diff
@@ -101,13 +99,19 @@ pub struct HttpState {
     pub override_manager: CertificateErrorOverrideManager,
 }
 
+impl Default for HttpState {
+    fn default() -> Self {
+        Self::new()
```
mrobinson, line 104: Can the `new()` method be removed now?

components/net/http_cache.rs
```diff
@@ -586,6 +582,12 @@ fn handle_range_request(
     None
 }
 
+impl Default for HttpCache {
+    fn default() -> Self {
+        Self::new()
```
mrobinson, line 587: Can the `new()` method be removed now?

components/net/http_loader.rs
```diff
@@ -1917,11 +1921,11 @@ async fn http_network_fetch(
...
-    if let Some(encoding) = response.headers.typed_get::<ContentEncoding>() {
-        if encoding.contains("gzip") {
-        } else if encoding.contains("compress") {
-        }
-    };
+    // if let Some(encoding) = response.headers.typed_get::<ContentEncoding>() {
+    //     if encoding.contains("gzip") {
+    //     } else if encoding.contains("compress") {
+    //     }
+    // };
```
mrobinson, line 1928: It looks like the bug is resolved so I would just remove all of Step 12 and this code as well.

mrobinson (changes requested):
Nice! Just a couple comments below:

mrobinson (approved):

components/net/image_cache.rs
```diff
@@ -224,7 +218,7 @@ impl ImageBytes {
                 ImageBytes::InProgress(ref mut bytes) => bytes,
                 ImageBytes::Complete(_) => panic!("attempted modification of complete image bytes"),
             };
-            mem::replace(own_bytes, vec![])
+            std::mem::take(own_bytes)
```
Loirooriol, line 221: Either `mem::take` or stop importing `std::mem`, but this is causing a compile warning (#31607).

eerii: Sorry about that, I'll fix it right away.

eerii: Should be fixed with #31608

PR #31568, opened 2024-03-07
clippy: fix warnings in various modules in components

eerii:
Split from #31514, fixes multiple small clippy warnings in multiple modules from `components`:

- `allocator`
- `background_hang_monitor`
- `dom_struct`
- `domobject_derive`
- `hyper_serde`
- `layout`
- `media`
- `pixels`
- `profile`
- `range`
- `remutex`
- `url`

---

### DIFF
--- a/components/allocator/lib.rs
+++ b/components/allocator/lib.rs
@@ -13,9 +13,7 @@ pub use crate::platform::*;
 mod platform {
     use std::os::raw::c_void;
 
-    use jemallocator;
-
-    pub use self::jemallocator::Jemalloc as Allocator;
+    pub use jemallocator::Jemalloc as Allocator;
 
     /// Get the size of a heap block.
     pub unsafe extern "C" fn usable_size(ptr: *const c_void) -> usize {
--- a/components/background_hang_monitor/background_hang_monitor.rs
+++ b/components/background_hang_monitor/background_hang_monitor.rs
@@ -99,7 +99,7 @@ impl BackgroundHangMonitorRegister for HangMonitorRegister {
             target_os = "linux",
             not(any(target_arch = "arm", target_arch = "aarch64"))
         ))]
-        let sampler = crate::sampler_linux::LinuxSampler::new();
+        let sampler = crate::sampler_linux::LinuxSampler::new_boxed();
         #[cfg(any(
             target_os = "android",
             all(target_os = "linux", any(target_arch = "arm", target_arch = "aarch64"))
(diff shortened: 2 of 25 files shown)


mrobinson (approved):

mrobinson:
Looks like this introduced a build error:

```
error[E0599]: no function or associated item named `new` found for struct `sampler::DummySampler` in the current scope
   --> components/background_hang_monitor/background_hang_monitor.rs:107:53
    |
107 |         let sampler = crate::sampler::DummySampler::new();
    |                                                     ^^^ function or associated item not found in `DummySampler`
    |
   ::: components/background_hang_monitor/sampler.rs:16:1
    |
16  | pub struct DummySampler;
    | ----------------------- function or associated item `new` not found for this struct
    |
```

sagudev:
> Looks like this introduced a build error:
> 
> ```
> error[E0599]: no function or associated item named `new` found for struct `sampler::DummySampler` in the current scope
>    --> components/background_hang_monitor/background_hang_monitor.rs:107:53
>     |
> 107 |         let sampler = crate::sampler::DummySampler::new();
>     |                                                     ^^^ function or associated item not found in `DummySampler`
>     |
>    ::: components/background_hang_monitor/sampler.rs:16:1
>     |
> 16  | pub struct DummySampler;
>     | ----------------------- function or associated item `new` not found for this struct
>     |
> ```

Clippy might neede to be run with `--all-features` and maybe even `--all-features`

eerii:
Yes, I missed the ones from other targets since they didn't show up when I built. I used the search now so there shouldn't be any more to change. Sorry about that!

eerii:
I am so sorry about that, I thought it was one of the ones that changed to default, but in this case the warning was saying that `new` usually returns `Self`, and this returns a `Box`, that's why I changed it to `new_boxed`. I missed that on the last commit so it failed again as `default` was not defined. Now all samplers (including the Windows and MacOS ones that I previously didn't see because of the clippy target) should be using `new_boxed` and the error should be gone. Sorry :(

mrobinson:
> I am so sorry about that, I thought it was one of the ones that changed to default, but in this case the warning was saying that `new` usually returns `Self`, and this returns a `Box`, that's why I changed it to `new_boxed`. I missed that on the last commit so it failed again as `default` was not defined. Now all samplers (including the Windows and MacOS ones that I previously didn't see because of the clippy target) should be using `new_boxed` and the error should be gone. Sorry :(

No worries. Thanks for the fix!

PR #31569, opened 2024-03-07
Extract generated finalizers into generic helper functions.

jdm:
This is an experiment to demonstrate one way to reduce the complexity of CodegenRust.py. By extracting the inline code generated for certain DOM binding methods that shouldn't be performance sensitive, we trade off the performance impact of a function call against code that is more easily read and modified.

Model for https://github.com/servo/servo/issues/31570.

---


mukilan (approved):

sagudev:
> we trade off the performance impact of a function call against code that is more easily read and modified.

We can hope it gets inlined or we can use `#[inline(always)]`.

PR #31574, opened 2024-03-08
Fix broken wiki link in README for Android instructions

msub2:
The existing link doesn't direct to an active wiki page and just lands you back at the index.


mukilan (approved):
Thank you for this fix!

PR #31580, opened 2024-03-08
Add rust-lldb to mozdebug

sagudev:
Supports using `rust-lldb` with  `--debugger rust-lldb`.

---


sagudev:
Should this be done as a patch, given that this is in third_party?

mrobinson:
> Should this be done as a patch, given that this is in third_party?

Yes, you should create a patch for this. Thanks!

third_party/mozdebug/Add_rust-lldb_debugger.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/mozdebug/mozdebug.py b/mozdebug/mozdebug.py
```
mrobinson, line 1: Please place this in a `patches` directory and prefix it by the order that the patch should go in, so that we know what order to apply them when upgrading. So this would move to:

`third_party/mozdebug/patches/01-Add_rust-lldb_debugger.patch`

Thanks!

mrobinson (approved):

mrobinson (approved):
Thanks!

PR #31585, opened 2024-03-08
Make image decoding thread pool the size of the system's CPUs

eerii:
Followup to #31517. Initially it just created 16 threads for the image decoding, which could be too many for CPUs with less cores.

This change makes it so it tries to get the number of CPUs, or an estimate of the amount of parallelism the program should use, see [`std::thread::available_parallelism`](https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html). It usually matches, but even in the cases it doesn't it is still a good metric for how many threads to create (and it is platform independent).

I also added a fallback so if there is an error in the calculation it defaults to 4, which seems a reasonable compromise, but let me know if another default is preferred.

I tested it on my machine (`Intel i7-4770HQ - 8 cores`), and it correctly returns 8.

---
- [x] These changes do not require tests because they don't change functionality, only the implementation

### DIFF
--- a/components/net/image_cache.rs
+++ b/components/net/image_cache.rs
@@ -4,8 +4,9 @@
 
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::HashMap;
-use std::mem;
+use std::num::NonZeroUsize;
 use std::sync::{Arc, Mutex};
+use std::{mem, thread};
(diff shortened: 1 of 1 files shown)


mrobinson (approved):
Thanks for the followup. This looks great.

PR #31586, opened 2024-03-08
layout: Do not inherit node and fragment flags in anonymous boxes

mrobinson:
This doesn't really have observable behavior right now, as much as I
tried to trigger some kind of bug. On the other hand, it's just wrong
and is very obvious when you dump the Fragment tree. If you create a
`display: table-cell` that is a child of the `<body>` all parts of the
anonymous table are flagged as if they are the `<body>` element.

---

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -5,6 +5,7 @@
 use std::borrow::Cow;
 
 use html5ever::{local_name, LocalName};
+use log::warn;
 use script_layout_interface::wrapper_traits::{ThreadSafeLayoutElement, ThreadSafeLayoutNode};
 use servo_arc::Arc as ServoArc;
 use style::properties::ComputedValues;
@@ -27,7 +28,7 @@ pub(crate) enum WhichPseudoElement {
 /// avoid having to repeat the same arguments in argument lists.
 #[derive(Clone)]
 pub(crate) struct NodeAndStyleInfo<Node> {
-    pub node: Node,
+    pub node: Option<Node>,
     pub pseudo_element_type: Option<WhichPseudoElement>,
     pub style: ServoArc<ComputedValues>,
 }
@@ -39,7 +40,7 @@ impl<Node> NodeAndStyleInfo<Node> {
         style: ServoArc<ComputedValues>,
     ) -> Self {
         Self {
-            node,
+            node: Some(node),
             pseudo_element_type: Some(pseudo_element_type),
             style,
         }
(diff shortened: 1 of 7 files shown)


components/layout_2020/dom_traversal.rs
```diff
@@ -39,22 +40,30 @@ impl<Node> NodeAndStyleInfo<Node> {
...
 impl<Node: Clone> NodeAndStyleInfo<Node> {
+    pub(crate) fn new_anonymous(&self, style: ServoArc<ComputedValues>) -> Self {
+        Self {
+            node: None,
+            pseudo_element_type: self.pseudo_element_type,
+            style,
+        }
+    }
+
     pub(crate) fn new_replacing_style(&self, style: ServoArc<ComputedValues>) -> Self {
```
Loirooriol, line 67: Should all callers of this use `new_anonymous`?

mrobinson: I chose to keep it for a couple callers which expect things to be inherited or which are creating list markers. There's a larger refactor we should do to clean this up and we need to implement the `::marker` pseudo, but these are pretty intrusive things that should be done in a larger project.

components/layout_2020/table/construct.rs
```diff
@@ -965,9 +963,15 @@ where
...
-                    let colspan = (node.get_colspan().unwrap_or(1) as usize).min(1000);
+                    let (rowspan, colspan) = info
+                        .node
+                        .map(|node| {
+                            let node = node.to_threadsafe();
+                            let rowspan = (node.get_rowspan().unwrap_or(1) as usize).min(65534);
+                            let colspan = (node.get_colspan().unwrap_or(1) as usize).min(1000);
+                            (rowspan, colspan)
+                        })
+                        .unwrap_or((1, 1));
```
Loirooriol, line 974:
```suggestion
                    let (rowspan, colspan) = info.node.map_or((1, 1), |node| {
                        let node = node.to_threadsafe();
                        let rowspan = node.get_rowspan().unwrap_or(1).min(65534) as usize;
                        let colspan = node.get_colspan().unwrap_or(1).min(1000) as usize;
                        (rowspan, colspan)
                    });
```

components/layout_2020/table/mod.rs
```diff
@@ -143,7 +143,10 @@ impl TableSlotCell {
 
     /// Get the node id of this cell's [`BaseFragmentInfo`]. This is used for unit tests.
     pub fn node_id(&self) -> usize {
-        self.base_fragment_info.tag.node.0
+        self.base_fragment_info
+            .tag
+            .map(|tag| tag.node.0)
+            .unwrap_or(0)
```
Loirooriol, line 149:
```suggestion
        self.base_fragment_info.tag.map_or(0, |tag| tag.node.0)
```

Loirooriol (approved):

PR #31595, opened 2024-03-09
wpt: Start running the WOFF2 tests

mrobinson:
This is the first step implementing WOFF2 font support in Servo.

---

### DIFF
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -116,8 +116,6 @@ skip: true
     skip: true
   [printing]
     skip: true
-  [WOFF2]
-    skip: true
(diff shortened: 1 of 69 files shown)


tests/wpt/include.ini
```diff
@@ -117,7 +117,7 @@ skip: true
   [printing]
     skip: true
   [WOFF2]
-    skip: true
+    skip: false
```
Loirooriol, line 120: CSS tests are enabled by default, so you can just remove

```
  [WOFF2]
    skip: false
```

Loirooriol (approved):

PR #31596, opened 2024-03-09
Align table cell measures in the block axis to Gecko/Blink/WebKit

Loirooriol:
---


mrobinson (approved):
Nice! This is conceptually a lot simpler too.

PR #31601, opened 2024-03-10
font_cache: Handle filtering `@font-face` rules in Servo

mrobinson:
Instead of letting Stylo filter `@font-face` rules, handle this
filtering in Servo. It doesn't make sense that Stylo knows about what
fonts Servo supports. This also cleans up a bit the way that this is
handled, giving an entire stylesheet of rules to the font cache to
process instead of letting each layout thread walk the rules. This
brings more of the font-related code into the FontCacheThread itself.

This is the first step toward adding WOFF2 support and fixing various
web font related bugs.

---


components/gfx/font_cache_thread.rs
```diff
@@ -624,3 +665,31 @@ impl fmt::Display for LowercaseString {
         self.inner.fmt(f)
     }
 }
+
+fn is_supported_web_font_source(source: &&Source) -> bool {
```
jdm, line 669: The double reference here is unusual!

jdm (approved):
Looking good!

mrobinson:
@jdm Thanks for the review.

The crashes in the test results are due to https://github.com/servo/servo/pull/31597 not landing yet (as it removes a couple `unimplemented()` calls for conic gradients). There are also a few tests that likely need to be rebased.

mrobinson:
So in the end this does change behavior a bit. Now we parse the font language override. No one is using Servo currently to a large extent and not really for the open web, so parsing this value but not using it is fairly low impact. Additionally, we would like to support this value one day.

PR #31602, opened 2024-03-10
Use libc::malloc_size on apple

sagudev:
As discovered in https://github.com/servo/servo/issues/31397#issuecomment-1987155922 Apple does not offer `malloc_usable_size`. Same is also used in Firefox: https://searchfox.org/mozilla-central/source/memory/mozalloc/mozalloc.cpp#128

---

### DIFF
--- a/components/allocator/lib.rs
+++ b/components/allocator/lib.rs
@@ -39,7 +39,10 @@ mod platform {
         #[cfg(target_os = "linux")]
         return libc::malloc_usable_size(ptr as *mut _);
 
-        #[cfg(not(target_os = "linux"))]
+        #[cfg(target_vendor = "apple")]
+        return libc::malloc_size(ptr);
+
+        #[cfg(not(any(target_os = "linux", target_vendor = "apple")))]
         return libc::malloc_usable_size(ptr);
     }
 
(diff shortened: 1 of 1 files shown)


sagudev:
Can somebody test this on mac using `--features servo_allocator/use-system-allocator`.

components/allocator/lib.rs
```diff
@@ -39,7 +39,10 @@ mod platform {
         #[cfg(target_os = "linux")]
         return libc::malloc_usable_size(ptr as *mut _);
 
-        #[cfg(not(target_os = "linux"))]
+        #[cfg(target_vendor = "apple")]
+        return libc::malloc_size(ptr);
+
+        #[cfg(not(any(target_os = "linux", target_vendor = "apple")))]
```
mrobinson, line 45: Can you also try to combine the two code paths for linux and android. The CI should catch it if it's an issue? So this would be:

```rust
        #[cfg(not(target_vendor = "apple")))]
        return libc::malloc_usable_size(ptr as *mut _);

        #[cfg(target_vendor = "apple")]
        return libc::malloc_size(ptr);
```

mrobinson (changes requested):

components/allocator/lib.rs
sagudev: I didn't know that rust can make *mut ptr to *const ptr automagically.

mrobinson (approved):
Thank you!

PR #31606, opened 2024-03-10
End ongoing IFC when inserting anonymous block-level table

### LINKED ISSUE #31603: Generated anonymous table parent is placed before preceding inline-level contents
```html
<!DOCTYPE html>
<span>foo</span>
<span style="display: table-cell">bar</span>
```

| Firefox | Servo |
| - | - |
|  |  |

Loirooriol:
So that the table appears after preceding inline-level contents. Fixes #31603.

---

- [X] These changes fix #31603


mrobinson (approved):
Looks good. Perhaps this could have the test that you wrote above?

PR #31608, opened 2024-03-10
Fix unused import left in #31564

### LINKED ISSUE #31607: unused import: `mem`, in components/net/image_cache.rs
When compiling I get

```
warning: unused import: `mem`
 --> components/net/image_cache.rs:9:11
  |
9 | use std::{mem, thread};
  |           ^^^
  |
```

#31564 used `std::mem::take(own_bytes)` instead of `mem::take(own_bytes)`. So it's a matter of either changing the consumer into `mem::take(own_bytes)`, or removing the import of `mem`.

eerii:
#31564 had an unused import issue while compiling. I changed `std::mem::take` to `mem::take` to use the import and be consistent with the rest of `std` uses. (Sorry for not noticing before the merge!)

---
- [x] These changes fix #31607
- [x] These changes do not require tests because they are just a fix for an import warning

### DIFF
--- a/components/net/image_cache.rs
+++ b/components/net/image_cache.rs
@@ -218,7 +218,7 @@ impl ImageBytes {
                 ImageBytes::InProgress(ref mut bytes) => bytes,
                 ImageBytes::Complete(_) => panic!("attempted modification of complete image bytes"),
             };
-            std::mem::take(own_bytes)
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):
Thanks!

PR #31609, opened 2024-03-10
 Update Stylo to 2023-09-01

Loirooriol:
This continues #31437

Changelog:
 - Upstream: https://github.com/servo/stylo/compare/c9275d6a3...24e88e73c
 - Servo fixups: https://github.com/servo/stylo/compare/8d29e4f...1e35d0e

---


components/layout_thread_2020/lib.rs
```diff
@@ -734,6 +734,7 @@ impl LayoutThread {
         let rayon_pool = STYLE_THREAD_POOL.lock().unwrap();
         let rayon_pool = rayon_pool.pool();
         let rayon_pool = rayon_pool.as_ref();
+        let rayon_pool = rayon_pool.filter(|p| p.current_thread_index().is_some());
```
mrobinson, line 737: Does this fix need to be extended for legacy layout as well?

So I'm still a bit confused how this works in Gecko. Does `current_thread_index()` actually return a value there for the main thread? If so, why? This approach seems okay, though we would be removing the main thread from the thread pool AFAIU. It would be nice to keep consistency with Gecko allowing the main thread to participate, as that was done for performance reasons on embedded devices, as I understand.

In either case, I think it might be nice to put a comment here explaining the situation as otherwise this is quite mysterious.

Loirooriol: Yes I forgot, legacy needs it too

Loirooriol: Fixed upstream: https://github.com/servo/stylo/commit/0c3139b534f8a950eb283261b820e7636d665639

components/script/dom/element.rs
```diff
@@ -3393,6 +3394,34 @@ impl<'a> SelectorsElement for DomRoot<Element> {
             }
         }
     }
+
+    fn add_element_unique_hashes(&self, filter: &mut BloomFilter) -> bool {
+        let mut f = |hash| filter.insert_hash(hash & BLOOM_HASH_MASK);
+
+        // We can't use style::bloom::each_relevant_element_hash(*self, f)
+        // since DomRoot<Element> doesn't have the TElement trait.
```
Loirooriol, line 3402: BTW, not worth it to do it in this PR, but should `DomRoot<Element>` implement TElement?

components/script/dom/element.rs
```diff
@@ -3393,6 +3394,34 @@ impl<'a> SelectorsElement for DomRoot<Element> {
...
+
+        for attr in self.attrs.borrow().iter() {
+            let name = style::values::GenericAtomIdent::cast(attr.local_name());
+            if !style::bloom::is_attr_name_excluded_from_filter(name) {
+                f(name.get_hash());
+            }
+        }
+
+        true
+    }
```
mrobinson, line 3424: Out of curiosity, what made you use this set of attributes to add to the hash?

components/script/dom/element.rs
mrobinson: Perhaps...Aren't we usually processing a LayoutDom node though?

mrobinson (approved):
Looks good! Just a question about the new function in `element.rs`.

components/script/dom/element.rs
Loirooriol: It's just an adaptation of `each_relevant_element_hash`:

```rs
pub fn each_relevant_element_hash<E, F>(element: E, mut f: F)
where
    E: TElement,
    F: FnMut(u32),
{
    f(element.local_name().get_hash());
    f(element.namespace().get_hash());

    if let Some(id) = element.id() {
        f(id.get_hash());
    }

    element.each_class(|class| f(class.get_hash()));

    element.each_attr_name(|name| {
        if !is_attr_name_excluded_from_filter(name) {
            f(name.get_hash())
        }
    });
}
```

In practice I believe it doesn't really matter, since this was added for `:has()` which is disabled on Servo.

mrobinson: Thanks for the explanation. BTW, you can land this whenever you are ready.

Loirooriol: So how was I supposed to land? Should I force-push Stylo's main branch to the upgrade branch?

PR #31610, opened 2024-03-10
clippy: Fix warnings in `components/constellation`

eerii:
Continuation to #31500, fixes most clippy warnings in `components/constellation`, excluding:

```rust
warning: large size difference between variants
  --> components/constellation/sandboxing.rs:31:1
   |
31 | / pub enum UnprivilegedContent {
32 | |     Pipeline(UnprivilegedPipelineContent),
   | |     ------------------------------------- the second-largest variant contains at least 0 bytes
33 | |     ServiceWorker(ServiceWorkerUnprivilegedContent),
   | |     ----------------------------------------------- the largest variant contains at least 472 bytes
34 | | }
   | |_^ the entire enum is at least 0 bytes
```

Should I apply the solution it offers (change the second member to `ServiceWorker(Box<ServiceWorkerUnprivilegedContent>)`, or simple add an ignore to it? On one hand, adding a `Box` could be detrimental for performance, but on the other side the big difference between variants can waste memory.

---

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -84,6 +84,8 @@
 //!
 //! See <https://github.com/servo/servo/issues/14704>
 
+#![allow(clippy::too_many_arguments)]
+
 use std::borrow::{Cow, ToOwned};
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet, VecDeque};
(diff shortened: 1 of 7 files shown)


components/constellation/constellation.rs
```diff
@@ -1825,15 +1827,12 @@ where
...
-                                _ => return,
-                            };
-                        },
-                        _ => (),
+                    if let MediaSessionEvent::PlaybackStateChange(ref state) = event {
+                        match state {
+                            MediaSessionPlaybackState::Playing |
+                            MediaSessionPlaybackState::Paused => (),
+                            _ => return,
+                        };
```
mrobinson, line 1835: I wonder if this could just be:

```rust
if !matches!(state,  MediaSessionPlaybackState::Playing | MediaSessionPlaybackState::Paused) {
    return;
}
```

mrobinson (approved):
Looks good, which one small suggestion:

components/constellation/constellation.rs
eerii: Yes, I think that makes sense and is consistent with other changes. Thanks!

mrobinson (approved):

PR #31611, opened 2024-03-10
clippy: Fix warnings in `components/layout_2020`

eerii:
Continuation to #31500, fixes all clippy warnings in `components/layout_2020`.

It includes a few ignores that I think are reasonable, but please give any feedback in this regard:
- `enum_variant_names`: allows it for `DisplayLayoutInternal::Table*` in `style_ext.rs` because it follows the spec names, and for `UpdatePoint` in `flow/root.rs` since moving AbsolutelyPositioned to the title seems less clear and conflicts with the base names it is wrapping.
- `module_inception`: since it would require renaming `fragment_tree/fragment_tree.rs` or the base module, and I can't think of a good enough replacement that justifies it
- `collapsible_match`: in `table/construct.rs`, this is only there until the corresponding todo is completed, and it is marked as so
- `needless_range_loop`: in `table/layout.rs` it is iterating between columns and rows of a table, and it uses the index for multiple things. the solution provided obfuscates the function.

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 #![allow(rustdoc::private_intra_doc_links)]
-#![allow(clippy::too_many_arguments)]
 
 //! Flow layout, also known as block-and-inline layout.
 
@@ -614,6 +613,7 @@ impl BlockLevelBox {
 ///
 /// - <https://drafts.csswg.org/css2/visudet.html#blockwidth>
 /// - <https://drafts.csswg.org/css2/visudet.html#normal-block>
+#[allow(clippy::too_many_arguments)]
 fn layout_in_flow_non_replaced_block_level_same_formatting_context(
     layout_context: &LayoutContext,
     positioning_context: &mut PositioningContext,
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 #![allow(rustdoc::private_intra_doc_links)]
+#![allow(clippy::too_many_arguments)]
 
 //! Flow layout, also known as block-and-inline layout.
 
@@ -856,7 +857,7 @@ impl NonReplacedFormattingContext {
             layout_context,
             positioning_context,
             &containing_block_for_children,
-            &containing_block,
+            containing_block,
         );
 
         let (block_size, inline_size) = match layout.content_inline_size_for_table {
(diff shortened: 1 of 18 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 #![allow(rustdoc::private_intra_doc_links)]
+#![allow(clippy::too_many_arguments)]
```
Taym95, line 5: can we allow `too_many_arguments ` on project level?

mrobinson: It would be nice to avoid adding functions with too many arguments in the future though.

mrobinson (approved):

components/layout_2020/flow/mod.rs
eerii: I agree with @mrobinson, I was thinking that once the clippy warning were under control we could add a clippy check to `test-tidy` or the ci, so that new code was already linted, in a similar way it does with formatting. It's hard to change previous functions but it would be nice if new ones were more manageable.

components/layout_2020/table/layout.rs
```diff
@@ -374,8 +374,7 @@ impl<'a> TableLayout<'a> {
         // >   * 100% minus the sum of the intrinsic percentage width of all prior columns in
         // >     the table (further left when direction is "ltr" (right for "rtl"))
         let mut total_intrinsic_percentage_width = 0.;
-        for column_index in 0..self.table.size.width {
-            let column_measure = &mut column_measures[column_index];
+        for column_measure in column_measures.iter_mut().take(self.table.size.width) {
```
Loirooriol, line 377: I think the `.take(self.table.size.width)` is not needed, `column_measures` shouldn't be longer than `self.table.size.width`.

components/layout_2020/flow/mod.rs
Loirooriol: Should then this be allowed for the specific functions instead of to the entire file?

eerii: Yes, I think that makes the most sense, even if it is more verbose for now. I'll change it!

components/layout_2020/table/layout.rs
eerii: I think so too, it shouldn't have more than that, we should probably take it out

components/layout_2020/table/layout.rs
```diff
@@ -971,8 +970,8 @@ impl<'a> TableLayout<'a> {
         for row_index in 0..self.table.slots.len() {
             let row = &self.table.slots[row_index];
             let mut cells_laid_out_row = Vec::new();
-            for column_index in 0..row.len() {
-                let cell = match &row[column_index] {
+            for (column_index, column_value) in row.iter().enumerate() {
```
Loirooriol, line 973: Instead of `column_value`, maybe call it `slot` or such?

Loirooriol (approved):

PR #31613, opened 2024-03-11
Take spaces into account in the max-content size of an IFC

### LINKED ISSUE #31605: Intrinsic sizing ignores spaces between atomic inlines
```html
<!DOCTYPE html>
<style>div { display: inline-block; }</style>
<div style="border: 1px solid; white-space: nowrap">
  <div>foo</div>
  <div>bar</div>
</div>
```

| Firefox | Servo |
| - | - |
|  |  |

The 2nd child overflows the container.

Loirooriol:
They were only considered in min-content sizes.
Also avoid some pointless conversions from Au to Length.

---

- [X] These changes fix #31605


mrobinson (approved):
Looks good to me!

PR #31615, opened 2024-03-11
Fix vertical alignment at the root of an IFC

### LINKED ISSUE #31604: vertical-align breaks the alignment of text at the very top of the independent FC
```html
<!DOCTYPE html>
<style>
  .wrapper { border: 1px solid; white-space: nowrap; vertical-align: bottom }
  .ib-content { display: inline-block; height: 50px; width: 50px; background: cyan }
</style>
<div class="wrapper" style="display: inline-block">
  foo
  <span>bar</span>
  <div class="ib-content"></div>
</div>
<div class="wrapper" style="display: inline-table">
  foo
  <span>bar</span>
  <div class="ib-content"></div>
</div>
```

| Firefox | Servo |
| - | - |
|  |  |

It happens for `vertical-align: top` and `bottom`, `middle`  is fine.

The text that is wrapped inside a `<span>` is also fine.

It also works well when the wrapper is not inline-level.

### LINKED ISSUE #31634: effective_vertical_align_for_inline_layout is bogus and needs to be removed
```html
<!DOCTYPE html>
<style>
.container {
  width: 50px;
  vertical-align: top;
  border: solid;
}
</style>
<div class="container" style="display: flow-root">
  <canvas width="50" height="50" style="background: cyan"></canvas>
</div>
<div class="container" style="display: inline-block">
  <canvas width="50" height="50" style="background: cyan"></canvas>
</div>
```

The 1st case is good, the 2nd is bad.

What happens is that, in the IFC established by the containers, they generate the strut using the `effective_vertical_align_for_inline_layout()` of the container! That just happens to work if the container is block-level because we get `baseline`, but if fails for inline-level containers.

We should just use `baseline` in all cases and get rid of `effective_vertical_align_for_inline_layout`.

Loirooriol:
---

- [X] These changes fix #31604 and #31634

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -76,13 +76,14 @@ use gfx::font::FontMetrics;
 use gfx::text::glyph::GlyphStore;
 use serde::Serialize;
 use servo_arc::Arc;
+use style::computed_values::line_height::T as LineHeight;
 use style::computed_values::white_space::T as WhiteSpace;
 use style::context::QuirksMode;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
 use style::values::computed::{Length, LengthPercentage};
 use style::values::generics::box_::{GenericVerticalAlign, VerticalAlignKeyword};
-use style::values::generics::text::LineHeight;
+use style::values::generics::NonNegative;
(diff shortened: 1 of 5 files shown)


Loirooriol:
This PR is not updating when I force push to the branch, I will open a new one.

PR #31194, opened 2024-01-26
Create pseudo async Create*PipelineAsync

sagudev:
Shader tests in CTS depends on CreateComputePipelineAsync.

I think having pseudo async functions is better than not having them at all.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] There are tests for these changes in WebGPU CTS

PR #31616, opened 2024-03-11
Add wpt-report output & continue if filtering failed

sagudev:
Split off from https://github.com/servo/servo/pull/31194.

Ths PR adds wptreports output in addition to other outputs. Wptreports are needed for moz-webgpu-cts, but they are also generally useful. I believe they can be feed into `./mach update-wpt` too and in comparison to raw logs they are more lightweight (raw logs have a lot of useless stuff eg.: server start, server end, ...).

WebGPU CTS failures are very spammy (a lot of printing), so we often hit "Message to long" errors in filtering phase, so no filter logs are made. Instead of failing when no filter logs are available this PR continues so all other logs still get merged.

---

### DIFF
--- a/.github/workflows/linux-wpt.yml
+++ b/.github/workflows/linux-wpt.yml
@@ -107,7 +107,7 @@ jobs:
           name: wpt-full-logs-linux-${{ inputs.wpt-layout }}
           pattern: wpt-full-logs-linux-${{ inputs.wpt-layout }}-*
           delete-merged: true
-      - name: Merge logs (full)
+      - name: Merge logs (wptreport)
         uses: actions/upload-artifact/merge@v4
         with:
           name: wptreport-linux-${{ inputs.wpt-layout }}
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-wpt.yml
```diff
@@ -87,25 +88,38 @@ jobs:
         with:
           name: wpt-full-logs-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}
           path: wpt-full-logs/*/
+      - name: Archive wptreport
+        uses: actions/upload-artifact@v4
+        if: ${{ always() }}
+        with:
+          name: wptreport-linux-${{ inputs.wpt-layout }}-${{ matrix.chunk_id }}.json
+          path: wptreport/*/
```
mrobinson, line 96: Let's incorporate this file into the full logs so that we still only have full and filtered. The reasoning here is two fold. Full/filtered/wptreport still raises the question: are the wptreport logs filtered? They aren't, so they should go with the full logs. The other thing is that I don't think it makes sense to create an artifact for every log type. We have a filtered version just because the full logs are larger and thus take longer to download. Thus it isn't a large concern to add another large file to this archive.

sagudev: In that case we need to introduce additional subfolders in zip so we can easier extract only wptreports/full logs.

sagudev:
try run: https://github.com/sagudev/servo/actions/runs/8237434998

mrobinson (approved):
Thank you.

mrobinson (approved):

Issue #31575, opened 2024-03-08
Fix as many rustdoc errors as possible


mrobinson:
1. Get Servo building and running on your system
2. Run `./mach doc`
3. Note the variety of documentation errors that show up during execution
4. Pick one or more of those errors. For instance:

```
warning: this URL is not a hyperlink
   --> components/shared/net/request.rs:189:20
    |
189 |     /// Step 12 of https://fetch.spec.whatwg.org/#concept-http-redirect-fetch
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://fetch.spec.whatwg.org/#concept-http-redirect-fetch>`
    |
    = note: bare URLs are not automatically turned into clickable links
    = note: `#[warn(rustdoc::bare_urls)]` on by default
```

In this case the line should be:

```rust
     /// Step 12 of <https://fetch.spec.whatwg.org/#concept-http-redirect-fetch>
```

servo-highfive:
Hi! If you have any questions regarding this issue, feel free to make a comment here, or ask it in [Zulip](https://servo.zulipchat.com/).

If you intend to work on this issue, then add `@servo-highfive: assign me` to your comment, and I'll assign this to you. :smile:

Aaryakhandelwal:
Hi, can be work with this issue. 
@servo-highfiv:assign me

maureenblack:
Hi, I'll love to work on this issue.
@servo-highfive: assign me

servo-highfive:
Hey @maureenblack! Thanks for your interest in working on this issue. It's now assigned to you!

Aaryakhandelwal:
I am an Outreachy applicant, can I also proceed to work with this issue?

jahielkomu:
> Yes, multiple people can work on this issue. Please just fix one or two errors and try to make sure that someone else hasn't opened a PR that fix the same ones.

Okay, thank you @mrobinson

kleenpulse:
Please, can I work on this issue.
@mrobinson

ssemanich:
Can I also work on this issue

ssemanich:
@servo-highfive: assign me

servo-highfive:
Hey @ssemanich! Thanks for your interest in working on this issue. It's now assigned to you!

oluwatobiss:
Hi,

servo-highfive: assign me

[...]

PR #31617, opened 2024-03-11
Fix rustdoc errors in `components/script/dom`

mnaibei:
Fixing rustdoc errors in components/script/dom

---

### DIFF
--- a/components/script/dom/audiobuffer.rs
+++ b/components/script/dom/audiobuffer.rs
@@ -44,7 +44,7 @@ pub struct AudioBuffer {
     #[ignore_malloc_size_of = "mozjs"]
     js_channels: DomRefCell<Vec<HeapBufferSource<Float32>>>,
     /// Aggregates the data from js_channels.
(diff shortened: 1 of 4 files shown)


mrobinson (approved):
Thank you.

mnaibei:
Hi @mrobinson do you think this merge fail is being caused because my pr is not rebased to main?

mrobinson:
> Hi @mrobinson do you think this merge fail is being caused because my pr is not rebased to main?

Looking at the failing jobs, it seems that the Mac build failed with

```
hdiutil: create failed - Resource busy
Creating Servo.app
Copying files
Swapping prefs
Finding dylibs and relinking
Adding version to Credits.rtf
Creating dmg
Packaging MacOS dmg exited with return value 1
Error: Process completed with exit code 1.
```

This is a thing that happens sometimes with GitHub Actions. Ideally, when this happens the package should wait a small number of seconds and try again. For now, we can try to land this again as-is.

PR #31619, opened 2024-03-11
layout: Properly parent table-row and table-row-group

mrobinson:
Put table cell content fragments into a hieararchy of fragments that
include their table row and table row group fragments. This ensures that
things like relative positioning and transforms set on rows and row
groups properly affect cells and cell content.

---

### DIFF
--- a/components/layout_2020/geom.rs
+++ b/components/layout_2020/geom.rs
@@ -4,7 +4,7 @@
 
 use std::convert::From;
 use std::fmt;
-use std::ops::{Add, AddAssign, Sub};
+use std::ops::{Add, AddAssign, Sub, SubAssign};
 
 use app_units::Au;
 use serde::Serialize;
@@ -26,7 +26,7 @@ pub type LengthOrAuto = AutoOr<Length>;
 pub type AuOrAuto = AutoOr<Au>;
 pub type LengthPercentageOrAuto<'a> = AutoOr<&'a LengthPercentage>;
 
-#[derive(Clone, Serialize)]
+#[derive(Clone, Copy, Serialize)]
 pub struct LogicalVec2<T> {
     pub inline: T,
     pub block: T,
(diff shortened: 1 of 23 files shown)


components/layout_2020/geom.rs
```diff
@@ -26,7 +26,7 @@ pub type LengthOrAuto = AutoOr<Length>;
 pub type AuOrAuto = AutoOr<Au>;
 pub type LengthPercentageOrAuto<'a> = AutoOr<&'a LengthPercentage>;
 
-#[derive(Clone, Serialize)]
+#[derive(Clone, Copy, Serialize)]
```
Loirooriol, line 29: Then there is a bunch of existing `.clone()` that become unnecessary. Should they be removed here, or be left for a follow-up?

components/layout_2020/table/construct.rs
```diff
@@ -663,6 +663,25 @@ where
...
+    fn push_table_row(&mut self, table_track: TableTrack) {
+        self.builder.table.rows.push(table_track);
+
+        let last_row = self.builder.table.rows.len();
+        let row_group = self
+            .current_row_group_index
+            .map(|index| &mut self.builder.table.row_groups[index]);
+        if let Some(row_group) = row_group {
+            row_group.track_range.end = last_row;
+        }
```
Loirooriol, line 684: Nit: this seems simpler as

```
        if let Some(index) = self.current_row_group_index {
            let row_group = &mut self.builder.table.row_groups[index];
            row_group.track_range.end = last_row;
        }
```

Loirooriol (approved):

components/layout_2020/geom.rs
mrobinson: Good point. In addition to the cleanup that we can do now that `AddAssign` and `SubAssign` don't need to take references for `LogicalVec`, I think this can be in a followup.

mrobinson (commented):
Thanks for the review!

mrobinson:
There are a few new failures here. At least one or two of these are due to tests expecting that a transform on `<tr>` will not transform row contents. I can't find this behavior anywhere in the specification,so I suspect it is just that no other browser supports this.

PR #31621, opened 2024-03-12
layout: make `margin` in `pbm` use app unit

atbrakhi:
Make `margin` in `pbm` use app unit

---


Loirooriol (approved):
I have added 2 patches, to allow `.auto_is(Au::zero)` instead of `.map(|t| t.auto_is(Au::zero))`, and to resolve inline margins as Au to be consistent with block margins.

PR #31623, opened 2024-03-12
Fix clippy warnings in `components/third_party`

mnaibei:
Fixing Clippy linter errors in components/third_party folder

---


mrobinson (approved):
Normally, I would say that we should be making these kind of changes in `third_party`, but blurmac is a dead project so this is very low risk and I don't even think it makes sense to create a patch for this like we do for most members of `third_party`.

PR #31341, opened 2024-02-13
layout: Add support for table rows, columns, rowgroups and colgroups

mrobinson:
This adds support for table rows, columns, rowgroups and colgroups.
There are few additions here:

1. The creation of fragments, which allows script queries and hit
   testing to work properly. These fragments are empty as all cells are
   still direct descendants of the table fragment.
2. Properly handling size information from tracks and track groups as
   well as frustrating rules about reordering rowgroups.
3. Painting a background seemlessly across track groups and groups. This
   is a thing that isn't done in legacy layout (nor WebKit)!

This causes some tests to start failing, because it exposes our lack of
support for collapse in tables.

Co-authored-by: Oriol Brufau <obrufau@igalia.com>

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] There are tests for these changes

PR #31624, opened 2024-03-12
Fix rustdoc errors in `components/layout_2020`

Aaryakhandelwal:
1)There were no function named as LineItems, thus removed the link syntax.
2) There was no variant named as Anonymous within the Fragment, thus removed the link syntax

---

- [X] These changes are part of #31575  (GitHub issue number if applicable)

- [X] These changes do not require tests because they don't effect any functionality and just fixes the documentation.

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -173,7 +173,7 @@ struct LineUnderConstruction {
     /// offset from `text-indent`.
     start_position: LogicalVec2<Length>,
 
-    /// The current inline position in the line being laid out into LineItems in this
+    /// The current inline position in the line being laid out into [`LineItem`]s in this
     /// [`InlineFormattingContext`] independent of the depth in the nesting level.
     inline_position: Length,
 
@@ -187,7 +187,7 @@ struct LineUnderConstruction {
     has_content: bool,
 
     /// Whether or not there are floats that did not fit on the current line. Before
-    /// the LineItems of this line are laid out, these floats will need to be
+    /// the [`LineItem`]s of this line are laid out, these floats will need to be
     /// placed directly below this line, but still as children of this line's Fragments.
     has_floats_waiting_to_be_placed: bool,
 
@@ -604,7 +604,7 @@ pub(super) struct InlineFormattingContextState<'a, 'b> {
     /// of the inline box is the state popped from the stack.
     inline_box_state_stack: Vec<InlineBoxContainerState>,
 
-    /// A vector of fragment that are laid out. This includes one Fragment::Anonymous
+    /// A vector of fragment that are laid out. This includes one [`Fragment::Positioning`]
     /// per line that is currently laid out plus fragments for all floats, which
     /// are currently laid out at the top-level of each [`InlineFormattingContext`].
     fragments: Vec<Fragment>,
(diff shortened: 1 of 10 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -173,7 +173,7 @@ struct LineUnderConstruction {
     /// offset from `text-indent`.
     start_position: LogicalVec2<Length>,
 
-    /// The current inline position in the line being laid out into [`LineItems`] in this
+    /// The current inline position in the line being laid out into LineItems in this
```
atbrakhi, line 176: Converting [&#96;LineItems&#96;] to `LineItems` is not the correct solution here. This should be [&#96;LineItem&#96;]s  instead. For reference you can see some examples [here](https://github.com/servo/servo/blob/da609076c32abcc3d3267cb663dbae861b2dfba7/components/layout_2020/flow/inline.rs#L47)

components/layout_2020/flow/inline.rs
```diff
@@ -187,7 +187,7 @@ struct LineUnderConstruction {
     has_content: bool,
 
     /// Whether or not there are floats that did not fit on the current line. Before
-    /// the [`LineItems`] of this line are laid out, these floats will need to be
+    /// the LineItems of this line are laid out, these floats will need to be
```
atbrakhi, line 190: same as above

components/layout_2020/flow/inline.rs
```diff
@@ -604,7 +604,7 @@ pub(super) struct InlineFormattingContextState<'a, 'b> {
     /// of the inline box is the state popped from the stack.
     inline_box_state_stack: Vec<InlineBoxContainerState>,
 
-    /// A vector of fragment that are laid out. This includes one [`Fragment::Anonymous`]
+    /// A vector of fragment that are laid out. This includes one Fragment::Anonymous
```
atbrakhi, line 607: we should not remove [&#96; and &#96;], we still want it to be a link. `Anonymous` was removed in this [PR](https://github.com/servo/servo/pull/31341) hence Rust doc is complaining,  this should be [&#96; Fragment::Positioning &#96;] now.

Aaryakhandelwal:
Should I create a new pull request with the changes

mrobinson:
> Should I create a new pull request with the changes

You should update this pull request with the new version of your changes. See https://github.com/github/hub/issues/198.

Aaryakhandelwal:
> > Should I create a new pull request with the changes
> 
> You should update this pull request with the new version of your changes. See [mislav/hub#198](https://github.com/mislav/hub/issues/198).

Yes I did, could you check once

Aaryakhandelwal:
> > Yes I did, could you check once
> 
> Your change now includes unrelated changes from other PRs. Do you mind rebasing it onto the latest from the main branch and making sure it just includes the changes you are trying to make?

I have re based it to the updated main branch. I can see the updated changes in my system locally. Could you check it once, if it's alright.

mrobinson:
> I have re based it to the updated main branch. I can see the updated changes in my system locally. Could you check it once, if it's alright.

No, it still seems to include too many changes. I recommend not doing any merges from main when creating a pull request. If you look at the "Files changed" tab in GitHub you can see that there are too many changes included in this PR due to the merge commits in your branch. The simplest thing to do now might be to redo these changes in a new PR.

Aaryakhandelwal:
Yes, I will just create a new pr. Thanks for your help.

PR #31626, opened 2024-03-12
clippy: Fix warnings in `components/net`

eerii:
Continuation to #31500, fixes all clippy warnings in `components/net` (excluding safety comments).

---

### DIFF
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -869,16 +869,13 @@ async fn scheme_fetch(
 }
 
 fn is_null_body_status(status: &Option<(StatusCode, String)>) -> bool {
-    match *status {
-        Some((status, _)) => matches!(
-            status,
-            StatusCode::SWITCHING_PROTOCOLS |
-                StatusCode::NO_CONTENT |
-                StatusCode::RESET_CONTENT |
-                StatusCode::NOT_MODIFIED
-        ),
-        _ => false,
-    }
+    matches!(
+        status,
+        Some((StatusCode::SWITCHING_PROTOCOLS, ..)) |
+            Some((StatusCode::NO_CONTENT, ..)) |
+            Some((StatusCode::RESET_CONTENT, ..)) |
+            Some((StatusCode::NOT_MODIFIED, ..))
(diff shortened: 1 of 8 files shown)


components/net/fetch/methods.rs
```diff
@@ -873,13 +870,13 @@ async fn scheme_fetch(
...
             StatusCode::SWITCHING_PROTOCOLS |
-            StatusCode::NO_CONTENT |
-            StatusCode::RESET_CONTENT |
-            StatusCode::NOT_MODIFIED => true,
-            _ => false,
-        },
+                StatusCode::NO_CONTENT |
+                StatusCode::RESET_CONTENT |
+                StatusCode::NOT_MODIFIED
+        ),
```
mrobinson, line 879: Could this just be something like this:

```
fn is_null_body_status(...) -> bool {
    matches!(
        status,
        Some(StatusCode::SWITCHING_PROTOCOLS) |
        Some(StatusCode::NO_CONTENT) |
        Some(StatusCode::RESET_CONTENT) |
        Some(StatusCode::NOT_MODIFIED)
    )
}

File: components/net/filemanager_thread.rs [hunk 2]
```diff
@@ -824,21 +826,18 @@ impl FileManagerStore {
...
+            self.insert(
+                id,
+                FileStoreEntry {
+                    origin,
+                    file_impl: FileImpl::Memory(blob_buf),
+                    refs: AtomicUsize::new(1),
+                    is_valid_url: AtomicBool::new(set_valid),
+                    outstanding_tokens: Default::default(),
+                },
+            );
```
mrobinson, line 840: Let's make this an early return.

mrobinson (changes requested):
Nice! Just a couple small comments:

components/net/filemanager_thread.rs
eerii: Done, thanks for the suggestion!

mrobinson (approved):

mrobinson:
Opened #31638 for this intermittent. Sending back to the MQ.

PR #31628, opened 2024-03-12
clippy: Fix assorted warnings in `components/`

eerii:
Continuation to #31500, fixes clippy warnings in the rest of `components/` modules (excluding all those in `script` and the ones about safety comments and result types).

It includes the following allows that seem to be reasonable:
- `upper_case_acronyms`: in the case of `WGPU` I think it looks better and more correct than `Wgpu`.
- `wrong_self_convention`: in `compositing/compositor.rs`, a function named `from_webrender` is used to turn `WebRenderPipelineId` into `PipelineId`. I can't think of a name that better represents this conversion, if anyone has ideas for this please do say them.
- `large_enum_variant`: same as in #31610

---

### DIFF
--- a/components/webdriver_server/lib.rs
+++ b/components/webdriver_server/lib.rs
@@ -66,24 +66,24 @@ use webdriver::server::{self, Session, SessionTeardownKind, WebDriverHandler};
 
 use crate::actions::{InputSourceState, PointerInputState};
 
-fn extension_routes() -> Vec<(Method, &'static str, ServoExtensionRoute)> {
-    return vec![
+fn extension_routes() -> Vec<(Method, &'static str, ServoExtensionRoutePrefs)> {
+    vec![
         (
             Method::POST,
             "/session/{sessionId}/servo/prefs/get",
-            ServoExtensionRoute::GetPrefs,
+            ServoExtensionRoutePrefs::Get,
         ),
         (
             Method::POST,
             "/session/{sessionId}/servo/prefs/set",
-            ServoExtensionRoute::SetPrefs,
+            ServoExtensionRoutePrefs::Set,
         ),
         (
             Method::POST,
             "/session/{sessionId}/servo/prefs/reset",
-            ServoExtensionRoute::ResetPrefs,
+            ServoExtensionRoutePrefs::Reset,
         ),
-    ];
+    ]
 }
 
 fn cookie_msg_to_cookie(cookie: cookie::Cookie) -> Cookie {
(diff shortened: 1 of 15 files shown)


components/webdriver_server/lib.rs
```diff
@@ -189,51 +189,51 @@ struct Handler {
...
 #[derive(Clone, Copy, Debug, PartialEq)]
-enum ServoExtensionRoute {
-    GetPrefs,
-    SetPrefs,
-    ResetPrefs,
+enum ServoExtensionRoutePrefs {
+    Get,
+    Set,
+    Reset,
 }
```
mrobinson, line 196: Hrm...I think these should probably stay. I'm not sure that `ServoExtensionRoutePrefs` makes sense...it's just that all of the ExtensionRoutes are currently pref related at the moment.

components/webdriver_server/lib.rs
```diff
@@ -189,51 +189,51 @@ struct Handler {
...
 
 #[derive(Clone, Debug, PartialEq)]
-enum ServoExtensionCommand {
-    GetPrefs(GetPrefsParameters),
-    SetPrefs(SetPrefsParameters),
-    ResetPrefs(GetPrefsParameters),
+enum ServoExtensionCommandPrefs {
+    Get(GetPrefsParameters),
+    Set(SetPrefsParameters),
+    Reset(GetPrefsParameters),
```
mrobinson, line 228: I would say the same here.

components/webgpu/lib.rs
```diff
@@ -330,6 +331,12 @@ impl WebGPU {
     }
 }
 
+type WGPUBufferMaps<'a> =
+    HashMap<id::BufferId, Rc<BufferMapInfo<'a, Option<WebGPUResponseResult>>>>;
+type WGPUPresentBufferMaps<'a> =
+    HashMap<id::BufferId, Rc<BufferMapInfo<'a, (Option<ErrorScopeId>, WebGPURequest)>>>;
```
mrobinson, line 337:
```suggestion
type WebGPUBufferMaps<'a> =
    HashMap<id::BufferId, Rc<BufferMapInfo<'a, Option<WebGPUResponseResult>>>>;
type WebGPUPresentBufferMaps<'a> =
    HashMap<id::BufferId, Rc<BufferMapInfo<'a, (Option<ErrorScopeId>, WebGPURequest)>>>;
```

Just for consistency with the rest of the code.

mrobinson (changes requested):

components/webdriver_server/lib.rs
eerii: You're right, it was cleaner originally. I reverted it back and added an allow.

mrobinson (approved):

PR #31632, opened 2024-03-12
rustdoc: Fix warnings in `components/script/dom`

azharcodeit:
Fixed unclosed HTML tag warning
Fixed unresolved link warning 

- [x] These changes are part of #31575

- [x] These changes do not require tests because they do not modify any functionality

### DIFF
--- a/components/script/dom/bindings/htmlconstructor.rs
+++ b/components/script/dom/bindings/htmlconstructor.rs
@@ -222,7 +222,7 @@ unsafe fn html_constructor(
 }
 
 /// Returns the constructor object for the element associated with the given local name.
-/// This list should only include elements marked with the [HTMLConstructor] extended attribute.
+/// This list should only include elements marked with the \[HTMLConstructor\] extended attribute.
 pub fn get_constructor_object_from_local_name(
     name: LocalName,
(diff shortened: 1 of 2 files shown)


components/script/dom/bindings/htmlconstructor.rs
```diff
@@ -222,7 +222,7 @@ unsafe fn html_constructor(
 }
 
 /// Returns the constructor object for the element associated with the given local name.
-/// This list should only include elements marked with the [HTMLConstructor] extended attribute.
+/// This list should only include elements marked with the \[HTMLConstructor\] extended attribute.
```
mrobinson, line 225: I escape the brackets isn't quite the right fix here. We either want to link to the Rust data structure, or if it's not a Rust data structure we probably want to use backticks intead.

azharcodeit: Attempted using backticks, but encountered the same warning. Setting an explicit link to [HTMLConstructor](https://html.spec.whatwg.org/multipage/dom.html#htmlconstructor) indeed resolved the issue

mrobinson (approved):

PR #31636, opened 2024-03-13
Fix vertical alignment at the root of an IFC

### LINKED ISSUE #31604: vertical-align breaks the alignment of text at the very top of the independent FC

### LINKED ISSUE #31634: effective_vertical_align_for_inline_layout is bogus and needs to be removed

Loirooriol:
At the root of an inline formatting context, we used its vertical-align in order to compute the strut. That was wrong, since vertical-align on a block container shouldn't affect the contents, it should only affect the alignment of the block container (if it's inline-level) within the parent IFC.

This was only working well if the block container was block-level, since effective_vertical_align_for_inline_layout returned `baseline` for block-level boxes.

Instead of the outer display type, this patch changes the logic to check whether we are at the root of the IFC.

---

- [X] These changes fix #31604 and #31634

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -76,12 +76,13 @@ use gfx::font::FontMetrics;
 use gfx::text::glyph::GlyphStore;
 use serde::Serialize;
 use servo_arc::Arc;
+use style::computed_values::vertical_align::T as VerticalAlign;
 use style::computed_values::white_space::T as WhiteSpace;
 use style::context::QuirksMode;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
-use style::values::computed::{Length, LengthPercentage};
-use style::values::generics::box_::{GenericVerticalAlign, VerticalAlignKeyword};
+use style::values::computed::Length;
+use style::values::generics::box_::VerticalAlignKeyword;
 use style::values::generics::text::LineHeight;
 use style::values::specified::text::{TextAlignKeyword, TextDecorationLine};
 use style::values::specified::{TextAlignLast, TextJustify};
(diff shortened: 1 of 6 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -1253,7 +1254,13 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
         let strut_size = if using_fallback_font {
             // TODO(mrobinson): This value should probably be cached somewhere.
             let container_state = self.current_inline_container_state();
-            let mut block_size = container_state.get_block_size_contribution(&font_metrics);
+            let mut block_size = container_state.get_block_size_contribution(
+                effective_vertical_align(
+                    &container_state.style,
+                    self.inline_box_state_stack.last().map(|c| &c.base),
+                ),
```
mrobinson, line 1261: I think this will be slightly easier to read if you save `vertical_align` to a temporary.

mrobinson (approved):
Thanks for the fix!

PR #31640, opened 2024-03-13
rustdoc: Fix warnings in `components/layout_2020`

Aaryakhandelwal:
1) No function named LineItems existed changed it to LineItem.
2) No variant named Anonymous present in Fragment, thus changed it to Positioning.

---

- [x] These changes do not require tests because  they don't change any functionality and only fixes in documentation being done.


mrobinson (approved):
Thanks!

PR #31641, opened 2024-03-13
layout: IFCs should not always be marked as containing floats

### LINKED ISSUE #31540: flow::BlockContainer.contains_floats() unconditionally returns true for IFCs
[flow::BlockContainer.contains_floats()](https://github.com/servo/servo/blob/096bd2977a7e322336bd885d68c7e722a47bfeaf/components/layout_2020/flow/mod.rs#L67) always returns true in the `InlineFormattingContext` case.

I asked on zulip https://servo.zulipchat.com/#narrow/stream/263398-general/topic/BlockContainer and was told this is likely a bug, so I'm opening an issue for it.

mrobinson:
Marking all IFCs as containing floats shouldn't change layout results,
but does prevent parallel layout in some cases. This change fixes an
issue where we were marking all IFCs as containing floats.

Fixes #31540.

---

- [x] These changes do not require tests because they do not result in easily observable changes.

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -971,7 +971,9 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
             match text_align {
                 TextAlign::Start => text_indent,
                 TextAlign::End => (available_space - line_length).max(text_indent),
-                TextAlign::Center => (available_space - line_length + text_indent) / 2.,
+                TextAlign::Center => {
+                    ((available_space - line_length + text_indent) / 2.).max(text_indent)
+                },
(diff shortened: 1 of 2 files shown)


Loirooriol:
It seems that `text-align: center` is centering unsafely, though this was hidden by the usage of PlacementAmongFloats.

Loirooriol:
Yeah I guess this needs `.max(text_indent)`:

https://github.com/servo/servo/blob/f5a1879bacf56200b08ac87a4a4e122d5c342f3b/components/layout_2020/flow/inline.rs#L974

Loirooriol (approved):

PR #31654, opened 2024-03-13
rustdoc: Fix errors in `components` and `ports`

komutumu:
- Corrected wrong spelling in `components/gfx/text/shaping/harfbuzz-rs`
- Corrected references in `components/layout_2020/fragment_tree`
- Corrected wrong spelling in  `ports/servoshell/minibrowser.rs`

---
- [X] These changes do not require tests because they don't change the functionality.


ports/servoshell/minibrowser.rs
```diff
@@ -295,8 +295,8 @@ impl Minibrowser {
         }
     }
 
-    /// Updates the location field from the given [BrowserManager], unless the user has started
-    /// editing it without clicking Go, returning true iff it has changed (needing an egui update).
+    /// Updates the location field from the given [Minibrowser], unless the user has started
```
atbrakhi, line 298: I think this should be `WebViewManager` instead of `BrowserManager`.  It was renamed in this [PR](https://github.com/servo/servo/pull/30842)

ports/servoshell/minibrowser.rs
```diff
@@ -295,8 +295,8 @@ impl Minibrowser {
...
+    /// editing it without clicking Go, returning true if it has changed (needing an egui update).
```
atbrakhi, line 299: this `iff` is correct, we don't need to change it to `if`. `iff` is shortened form of `if and only if`

ports/servoshell/minibrowser.rs
jahielkomu: Thank you @atbrakhi for this review.
Let me correct them immediately.

jahielkomu: Hey @atbrakhi changing it to ```WebViewManager``` causes a warning. 
I'm I missing something?

jahielkomu: I was missing something, it was the spelling. 
Thank you @atbrakhi for the review, please kindly re-review and comment

components/layout_2020/fragment_tree/fragment.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) enum Fragment {
     /// Floating content. A floated fragment is very similar to a normal
     /// [BoxFragment] but it isn't positioned using normal in block flow
     /// positioning rules (margin collapse, etc). Instead, they are laid out by
-    /// the [SequentialLayoutState] of their float containing block formatting
+    /// the [PositioningFragment] of their float containing block formatting
```
mrobinson, line 33: `SequentialLayoutState` is correct here. I think you just need to import it.

jahielkomu: Hello @mrobinson . Thank you for this review. 
I have imported ```SequentialLayoutState``` like this; ```use crate::flow::float::SequentialLayoutState; ``` I get an error message; ```warning: unused import: `crate::flow::float::SequentialLayoutState`
  --> components/layout_2020/fragment_tree/fragment.rs:26:5
   |
26 | use crate::flow::float::SequentialLayoutState;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default ```
   
   I'm missing the correct way to import? Kindly advise.

mrobinson: Sorry, I may have mislead you here. I would try to remove the import and use ``crate::flow::float::SequentialLayoutState` directly in the rustdoc comment. If that doesn't work, I would replace the brackets with backticks instead.

mrobinson (approved):
Thanks for the fixes!

components/layout_2020/fragment_tree/fragment.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) enum Fragment {
     /// Floating content. A floated fragment is very similar to a normal
     /// [BoxFragment] but it isn't positioned using normal in block flow
     /// positioning rules (margin collapse, etc). Instead, they are laid out by
-    /// the [SequentialLayoutState] of their float containing block formatting
+    /// the [crate::flow::float::SequentialLayoutState] of their float containing block formatting
     /// context.
```
Loirooriol, line 34: Please balance these lines, the former is now considerably longer than the others in this paragraph, and the second has a single word.

Loirooriol (changes requested):

components/layout_2020/fragment_tree/fragment.rs
jahielkomu: I have updated the PR @Loirooriol 
Thank you.

Loirooriol (approved):
Thanks!

PR #31657, opened 2024-03-14
servoshell: allow event tracing to be configured with RUST_LOG

delan:
To make debugging easier, servoshell logs events from winit, from servo, and to servo, but the output of these logs can be so noisy as to be unusable, unless we reconfigure which event types are logged [at compile time](https://github.com/servo/servo/blob/f3a73dbed39c38c6581fc356d3d4f2d38c906e93/ports/servoshell/app.rs#L143-L155).

We use [log](https://crates.io/crates/log) and [env_logger](https://crates.io/crates/env_logger), which allows runtime configuration with RUST_LOG, but by default, while you can filter logs by module path or filter them to only messages that match a regex, you can’t filter them to *exclude* messages by regex. Custom targets allow us to key our logs on an arbitrary string (as long as it doesn’t contain `,=/`).

This patch sets custom targets on servoshell’s event tracing logs, allowing them to be filtered by RUST_LOG. The targets are of the form “servoshell&lt;[source]@[event]” or “servoshell>[destination]@[event]”. For example:

- to trace only events from servo:<br>`RUST_LOG='servoshell<=off,servoshell>=off,servoshell<servo@'`
- to trace all events except for AxisMotion events: <br>`RUST_LOG='servoshell<,servoshell>,servoshell<winit@WindowEvent(AxisMotion)=off'`
- to trace only winit window moved events:<br>`RUST_LOG='servoshell<=off,servoshell>=off,servoshell<winit@WindowEvent(Moved)'`

You can get a very usable setup that traces most unusual events with:

RUST_LOG='warn,servoshell&lt;,servoshell>,servoshell&lt;winit@DeviceEvent=off,servoshell&lt;winit@MainEventsCleared=off,servoshell&lt;winit@NewEvents(WaitCancelled)=off,servoshell&lt;winit@RedrawEventsCleared=off,servoshell&lt;winit@RedrawRequested=off,servoshell&lt;winit@UserEvent(WakerEvent)=off,servoshell&lt;winit@WindowEvent(CursorMoved)=off,servoshell&lt;winit@WindowEvent(AxisMotion)=off,servoshell&lt;servo@EventDelivered=off,servoshell&lt;servo@ReadyToPresent=off,servoshell>servo@Idle=off,servoshell>servo@MouseWindowMoveEventClass=off'

---

- [x] These changes do not require tests because they only affect debug logging

### DIFF
--- a/ports/servoshell/tracing.rs
+++ b/ports/servoshell/tracing.rs
@@ -1,0 +1,181 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/// Get the log target for an event, as a static string.
+pub(crate) trait LogTarget {
+    fn log_target(&self) -> &'static str;
+}
+
+mod from_winit {
+    use super::LogTarget;
+    use crate::events_loop::WakerEvent;
+
+    macro_rules! target {
+        ($($name:literal)+) => {
+            concat!("servoshell<winit@", $($name),+)
+        };
+    }
+
+    impl LogTarget for winit::event::Event<'_, WakerEvent> {
+        fn log_target(&self) -> &'static str {
+            use winit::event::StartCause;
+            match self {
+                Self::NewEvents(start_cause) => match start_cause {
+                    StartCause::ResumeTimeReached { .. } => target!("NewEvents(ResumeTimeReached)"),
+                    StartCause::WaitCancelled { .. } => target!("NewEvents(WaitCancelled)"),
+                    StartCause::Poll => target!("NewEvents(Poll)"),
+                    StartCause::Init => target!("NewEvents(Init)"),
+                },
+                Self::WindowEvent { event, .. } => event.log_target(),
+                Self::DeviceEvent { .. } => target!("DeviceEvent"),
+                Self::UserEvent(WakerEvent) => target!("UserEvent(WakerEvent)"),
+                Self::Suspended => target!("Suspended"),
+                Self::Resumed => target!("Resumed"),
+                Self::MainEventsCleared => target!("MainEventsCleared"),
+                Self::RedrawRequested(_) => target!("RedrawRequested"),
+                Self::RedrawEventsCleared => target!("RedrawEventsCleared"),
+                Self::LoopDestroyed => target!("LoopDestroyed"),
+            }
+        }
+    }
+
+    impl LogTarget for winit::event::WindowEvent<'_> {
+        fn log_target(&self) -> &'static str {
+            macro_rules! target_variant {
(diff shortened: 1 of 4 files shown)


ports/servoshell/tracing.rs
```diff
@@ -0,0 +1,229 @@
...
+/// - Recommended filters when tracing is enabled:
+///   - servoshell<winit@DeviceEvent=off
+///   - servoshell<winit@MainEventsCleared=off
+///   - servoshell<winit@NewEvents(WaitCancelled)=off
+///   - servoshell<winit@RedrawEventsCleared=off
+///   - servoshell<winit@RedrawRequested=off
+///   - servoshell<winit@UserEvent(WakerEvent)=off
+///   - servoshell<winit@WindowEvent(AxisMotion)=off
+///   - servoshell<winit@WindowEvent(CursorMoved)=off
+macro_rules! trace_from_winit {
```
mrobinson, line 17: Maybe "trace_winit_event" would be a little clearer?

ports/servoshell/webview.rs
```diff
@@ -112,7 +113,7 @@ where
 
     pub fn handle_window_events(&mut self, events: Vec<EmbedderEvent>) {
         for event in events {
-            trace!("embedder <- window EmbedderEvent {:?}", event);
+            trace_to_servo!(event, "{event:?}");
```
mrobinson, line 115: "trace_embedder_event" perhaps?

ports/servoshell/webview.rs
```diff
@@ -412,11 +413,11 @@ where
         let mut need_present = false;
         let mut history_changed = false;
         for (webview_id, msg) in events {
-            trace!(
-                "embedder <- servo EmbedderMsg ({:?}, {:?})",
-                webview_id.map(|x| format!("{}", x)),
-                msg
-            );
+            if let Some(webview_id) = webview_id {
+                trace_from_servo!(msg, "{webview_id} {msg:?}");
```
mrobinson, line 416: Simillarly "trace_embedder_msg" ? Dunno. Maybe these are more complicated than they seem?

mrobinson (approved):
Seems pretty reasonable. My only doubts are about the names (typical, I know :sweat_smile:).

ports/servoshell/tracing.rs
delan: done in 198decce710d39b80ab8e02b35396515145186b5

ports/servoshell/webview.rs
delan: done in 198decce710d39b80ab8e02b35396515145186b5

ports/servoshell/webview.rs
delan: done in 198decce710d39b80ab8e02b35396515145186b5

mrobinson (approved):

PR #31660, opened 2024-03-14
Obey white-space when intrinsically sizing an IFC

### LINKED ISSUE #31652: Intrinsic sizes of block containers ignore `white-space`
```html
<!DOCTYPE html>
<style>div { display: table-cell; border: solid; white-space: pre }</style>
<div>a        </div>
```

| Servo | Firefox |
| - | - |
|  |  |

Loirooriol:
The old logic was assuming that all whitespace was a break opportunity,
and that no newlines would be preserved.

Note that text shaping considers the advance of a newline to be the same
as a space. This was problematic because if we have a segment with a
preserved space and newline, only the advance of the space should
contrinute to the size of the block container. Therefore, I'm changing
the breaker logic in other to have newline characters in their own
segment.

Then glyph_run_is_whitespace_ending_with_preserved_newline can just be
renamed to glyph_run_is_preserved_newline.

This patch is still not perfect because it doesn't check allow_wrap(),
so `nowrap` is treated like `normal`, and `pre-wrap` like `pre`.

---

- [X] These changes fix #31652
  - Some tests in `/css/css-text/tab-size/` fail because the `tab-size` is not enabled on Servo.
  - `/_mozilla/css/white-space-pre-wrap.htm` fails because of #31667

### DIFF
--- a/components/gfx/text/text_run.rs
+++ b/components/gfx/text/text_run.rs
@@ -225,6 +225,16 @@ impl<'a> TextRun {
 
         let breaker = breaker.as_mut().unwrap();
 
+        let mut push_range = |range: &std::ops::Range<usize>, options: &ShapingOptions| {
+            glyphs.push(GlyphRun {
+                glyph_store: font.shape_text(&text[range.clone()], &options),
+                range: Range::new(
+                    ByteIndex(range.start as isize),
+                    ByteIndex(range.len() as isize),
+                ),
+            });
+        };
+
         while !finished {
             let (idx, _is_hard_break) = breaker.next(text);
             if idx == text.len() {
@@ -240,9 +250,9 @@ impl<'a> TextRun {
 
             // Split off any trailing whitespace into a separate glyph run.
             let mut whitespace = slice.end..slice.end;
-            if let Some((i, _)) = word
-                .char_indices()
-                .rev()
+            let mut rev_char_indices = word.char_indices().rev().peekable();
+            let ends_with_newline = rev_char_indices.peek().map_or(false, |&(_, c)| c == '\n');
+            if let Some((i, _)) = rev_char_indices
                 .take_while(|&(_, c)| char_is_whitespace(c))
                 .last()
             {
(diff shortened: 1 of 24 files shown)


components/gfx/text/text_run.rs
```diff
@@ -254,26 +264,28 @@ impl<'a> TextRun {
...
-                        ByteIndex(whitespace.len() as isize),
-                    ),
-                });
+
+                // The breaker breaks after every newline, so either there is none,
+                // or there is exactly one at the very end. In the latter case,
+                // split it into a different run. That's because shaping considers
+                // a newline to have the same advance as a space, but during layout
+                // we want to treat the newline as having no advance.
+                if word.chars().last().unwrap() == '\n' {
```
mrobinson, line 280: Is there any way to avoid this call to `chars()` here? It has to walk the entire string again and separate characters into Unicode characters. Since this code is somewhat performance sensitive, it would be good to avoid it.

Loirooriol: Is `as_bytes()` better?

components/layout_2020/flow/inline.rs
```diff
@@ -2248,8 +2248,8 @@ struct ContentSizesComputation<'a> {
     current_line: ContentSizes,
     /// Size for whitepsace pending to be added to this line.
     pending_whitespace: Au,
-    /// Whether or not this IFC has seen any non-whitespace content.
-    had_non_whitespace_content_yet: bool,
+    /// Whether or not this IFC has seen any content, excluding discarded whitespace.
```
mrobinson, line 2251: Perhaps "collapsed" would be better here?

```suggestion
    /// Whether or not this IFC has seen any content, excluding collapsed whitespace.
```

components/layout_2020/flow/text_run.rs
```diff
@@ -442,7 +439,7 @@ impl TextRun {
             return false;
         }
 
-        let last_byte = self.text.as_bytes().get(run.range.end().to_usize() - 1);
+        let last_byte = self.text.as_bytes().get(run.range.begin().to_usize());
         last_byte == Some(&b'\n')
```
mrobinson, line 443: Maybe "last" doesn't make as much sense any longer:

```suggestion
        let byte = self.text.as_bytes().get(run.range.begin().to_usize());
        byte == Some(&b'\n')
```

mrobinson (approved):
Looks reasonable!

PR #31661, opened 2024-03-14
Cleanup UA styles for anonymous table rows and cells

Loirooriol:
These properties were being set to their initial value, this wasn't needed because they are non-inherited properties and there isn't anything else setting them.

---


mrobinson (approved):
Thanks for the clean up! Feel free to land this is the tests look good. I think this shouldn't change anything.

PR #31668, opened 2024-03-14
fonts: Add Noto Sans CJK fonts to the Linux fallback list

### LINKED ISSUE #31662: On Linux, add Noto Sans CJK fonts to fallback list
When a character isn’t supported by any of the fonts chosen by the page, some browsers search through all of the installed fonts to find one that supports text in that language (e.g. [Firefox](https://searchfox.org/mozilla-central/rev/109bb25545f0d2df31954dc0a9afbf30d900b6bb/gfx/thebes/gfxPlatformFontList.cpp#1086-1101)). Servo can’t do that yet, but we currently have hardcoded lists of “fallback fonts” that we try in those situations. There are separate lists for each platform, and on Linux, we use the list in [freetype/font_list.rs](https://github.com/servo/servo/blob/f3a73dbed39c38c6581fc356d3d4f2d38c906e93/components/gfx/platform/freetype/font_list.rs#L155-L169).

For text in Chinese, Japanese, or Korean, we add an extra list of fallback fonts, but some of these fonts haven’t been updated to support new characters in a very long time. [Noto fonts](https://fonts.google.com/noto/use) are often installed by default on Linux distros like Ubuntu. Add the following font names to the fallback list for those languages only:

- Noto Sans CJK HK
- Noto Sans CJK JP
- Noto Sans CJK KR
- Noto Sans CJK SC
- Noto Sans CJK TC
- Noto Sans HK
- Noto Sans JP
- Noto Sans KR
- Noto Sans SC
- Noto Sans TC

For testing, you can use the character [𪣻](https://charming.daz.cat/#2A8FB) (U+2A8FB), or the following test page:<br>`data:text/html;charset=utf-8,%F0%AA%A3%BB`

richarddushime:
---

- [x] These changes fix #31662 GitHub issue number

### DIFF
--- a/components/gfx/platform/freetype/font_list.rs
+++ b/components/gfx/platform/freetype/font_list.rs
@@ -162,6 +162,17 @@ pub fn fallback_font_families(codepoint: Option<char>) -> Vec<&'static str> {
             families.push("Droid Sans Fallback");
             families.push("WenQuanYi Micro Hei");
             families.push("NanumGothic");
+            families.push("Noto Sans CJK HK");
+            families.push("Noto Sans CJK JP");
+            families.push("Noto Sans CJK KR");
+            families.push("Noto Sans CJK SC");
+            families.push("Noto Sans CJK TC");
+            families.push("Noto Sans HK");
+            families.push("Noto Sans JP");
+            families.push("Noto Sans KR");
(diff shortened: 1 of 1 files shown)


delan (approved):
Thanks!

PR #31681, opened 2024-03-14
Allow pre-wrap whitespace to hang at the end of the line

### LINKED ISSUE #31667: Whitespace with `white-space: pre-wrap` at the end of the line should hang
```html
<!DOCTYPE html>
<style>
body { width: 0 }
div { display: inline-block; border: solid magenta; font: 20px/1 Ahem; white-space: pre-wrap }
</style>
<div>X   X</div>
```

| Actual | Expected |
| :-: | :-: |
|  |  |

This is in https://drafts.csswg.org/css-text-4/#white-space-phase-2

> If [white-space-collapse](https://drafts.csswg.org/css-text-4/#propdef-white-space-collapse) is [preserve](https://drafts.csswg.org/css-text-4/#valdef-white-space-collapse-preserve) and [text-wrap-mode](https://drafts.csswg.org/css-text-4/#propdef-text-wrap-mode) is not [nowrap](https://drafts.csswg.org/css-text-4/#valdef-text-wrap-mode-nowrap), the UA must (unconditionally) [hang](https://drafts.csswg.org/css-text-4/#hang) this sequence

Since we don't have selections I guess we can just trim the space (like for `white-space: normal`) instead of hanging. With a selection it should look like

Loirooriol:
---

- [X] These changes fix #31667

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -1215,7 +1215,12 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
                 .current_inline_container_state()
                 .strut_block_sizes
                 .clone();
-            self.update_unbreakable_segment_for_new_content(&strut_size, Length::zero(), false);
+            self.update_unbreakable_segment_for_new_content(
+                &strut_size,
+                Length::zero(),
+                false,
+                false,
+            );
         }
 
         self.had_inflow_content = true;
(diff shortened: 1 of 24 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -1310,13 +1314,16 @@ impl<'a, 'b> InlineFormattingContextState<'a, 'b> {
         block_sizes_of_content: &LineBlockSizes,
         inline_size: Length,
         is_collapsible_whitespace: bool,
+        is_wrappable_whitespace: bool,
```
mrobinson, line 1317: Now that there are two booleans perhaps this could become a bitflags which should make the code more readable? I'm imagining something like this:

```rust

bitflags! {
    pub struct SegmentContentFlags: u8 {
        const COLLAPSIBLE_WHITESPACE = 0b00000001;
        const WRAPPABLE_WHITESPACE = 0b00000010;
    }
}
```

mrobinson (changes requested):
Great change! This fixes so many tests. I just have one minor comment, but otherwise this looks good.

components/layout_2020/flow/inline.rs
Loirooriol: Done!

mrobinson (approved):
This looks good to me with one nit. Feel free to land this with that minor adjustment. Thanks!

PR #31684, opened 2024-03-15
Gamepad: Remove GamepadList and fix dropped connection event on startup

msub2:
This removes GamepadList, as it has been replaced in the spec IDL with `sequence<Gamepad?>`, and adjusts code where necessary.
In addition, this fixes an existing issue in servoshell where, if a gamepad is already plugged in on startup, the connection event from GilRs gets lost due to the browsing context it's targeting not yet existing.

---

- [x] These changes do not require tests because they consist of IDL changes that do not affect any existing WPTs and gamepad connection relies on physical hardware.


components/script/dom/globalscope.rs
```diff
@@ -3149,9 +3148,16 @@ impl GlobalScope {
...
-                    let gamepad_arr: [DomRoot<Gamepad>; 1] = [gamepad.clone()];
-                    gamepad_list.add_if_not_exists(&gamepad_arr);
+                    let navigator = window.Navigator();
+                    let mut gamepad_list = navigator.gamepads();
+                    if !gamepad_list
+                        .iter()
+                        .any(|g| g.as_ref().map_or(false, |g| g.gamepad_id() == gamepad.gamepad_id()))
+                    {
+                        // Ensure that the gamepad has the correct index
+                        gamepad.update_index(gamepad_list.len() as i32);
```
gterzian, line 3158: I don't think this is correct, because the index should not change. Instead, you should fill any missing indices in the list with `None`. This could happen if the focused webview changes. 

Btw, in webview, you can nest the index creation inside the `EventType::Connected` variant handling, and you can have a single `self.event_queue.push(EmbedderEvent::Gamepad(event));` for all variants.

components/script/dom/globalscope.rs
```diff
@@ -3167,16 +3173,19 @@ impl GlobalScope {
...
-                            if window.Document().is_fully_active() {
-                                gamepad.update_connected(false, gamepad.exposed());
-                                gamepad_list.remove_gamepad(index);
+                        let navigator = window.Navigator();
+                        let mut gamepad_list = navigator.gamepads();
+                        if let Some(gamepad_or_none) = gamepad_list.get(index) {
+                            if let Some(gamepad) = gamepad_or_none {
+                                if window.Document().is_fully_active() {
+                                    gamepad.update_connected(false, gamepad.exposed());
+                                    gamepad_list.remove(index);
```
gterzian, line 3182: instead of removing it, it should be set to `None`.

components/script/dom/globalscope.rs
```diff
@@ -3167,16 +3173,19 @@ impl GlobalScope {
...
+                                }
                             }
                         }
-                        for i in (0..gamepad_list.Length()).rev() {
-                            if gamepad_list.Item(i as u32).is_none() {
-                                gamepad_list.remove_gamepad(i as usize);
+                        for i in (0..gamepad_list.len()).rev() {
+                            if gamepad_list.get(i as usize).is_none() {
+                                gamepad_list.remove(i as usize);
```
gterzian, line 3188: instead of removing it, it should be set to `None`.

components/script/dom/globalscope.rs
```diff
@@ -3198,36 +3207,43 @@ impl GlobalScope {
...
-                                                &window.upcast::<GlobalScope>().task_canceller(TaskSourceName::Gamepad),
-                                            )
-                                            .expect("Failed to queue update gamepad connect task.");
+                        let navigator = window.Navigator();
+                        let gamepad_list = navigator.gamepads();
+                        if gamepad_list.is_empty() {
+                            return;
+                        }
+                        if let Some(gamepad_or_none) = gamepad_list.get(index) {
+                            if let Some(gamepad) = gamepad_or_none {
```
gterzian, line 3216: I think you can collapse and/or do an early return if `None`.

components/script/dom/navigator.rs
```diff
@@ -47,7 +47,7 @@ pub struct Navigator {
     xr: MutNullableDom<XRSystem>,
     mediadevices: MutNullableDom<MediaDevices>,
     /// <https://www.w3.org/TR/gamepad/#dfn-gamepads>
-    gamepads: MutNullableDom<GamepadList>,
+    gamepads: DomRefCell<Vec<Option<DomRoot<Gamepad>>>>,
```
gterzian, line 50: `DomRefCell<Vec<MutNullableDom<Gamepad>>>`

components/script/dom/navigator.rs
```diff
@@ -202,9 +199,7 @@ impl NavigatorMethods for Navigator {
             return Vec::new();
         }
 
-        let root = self.gamepads.or_init(|| GamepadList::new(&global, &[]));
-
-        root.list()
+        self.gamepads.borrow().to_vec()
```
gterzian, line 202: iter map into domroot using `MutNullableDom.get`

ports/servoshell/app.rs
```diff
@@ -436,7 +436,10 @@ impl App {
 
         // Catch some keyboard events, and push the rest onto the WebViewManager event queue.
         webviews.handle_window_events(embedder_events);
-        if pref!(dom.gamepad.enabled) && webviews.webview_id().is_some() {
+
+        // If the Gamepad API is enabled, handle gamepad events from GilRs.
+        // Checking for current_url_string should ensure we'll have a valid browsing context.
+        if pref!(dom.gamepad.enabled) && webviews.current_url_string().is_some() {
```
gterzian, line 442: Could you please try to remove the current_url_string condition and put the handling of gamepad events at https://github.com/servo/servo/blob/5144e337acbbff246ba4dff62e1f72d212bdd844/ports/servoshell/app.rs#L456

I think what you need is just to first have servo handle the incoming window events.

gterzian (changes requested):
Nice progress! Couple of comments...

components/script/dom/globalscope.rs
msub2: This is step 6 of the queued task though, "While navigator.[[gamepads]] is not empty and the last item of navigator.[[gamepads]] is null, remove the last item of navigator.[[gamepads]]."

gterzian: Ah, ok!

components/script/dom/globalscope.rs
msub2: I originally moved that out from the old GamepadList class, but I've actually now updated it with the more up-to-date spec steps on selecting an unused index, which will be in my next set of changes.

msub2:
Touched up everything I could, unfortunately I couldn't find a better signal for the browser context being created given what I have to work with on the embedder side. It seems pretty reliable for what it's worth, so maybe we can leave it that way for now unless someone comes up with a better solution.

components/script/dom/globalscope.rs
```diff
@@ -3198,11 +3214,15 @@ impl GlobalScope {
                 task!(update_gamepad_state: move || {
                     let global = this.root();
                     if let Some(window) = global.downcast::<Window>() {
-                        let gamepad_list = window.Navigator().gamepads();
-                        if let Some(gamepad) = gamepad_list.Item(index as u32) {
+                        let navigator = window.Navigator();
+                        let gamepad_list = navigator.gamepads();
+                        if let Some(gamepad_or_none) = gamepad_list.get(index) {
```
gterzian, line 3219: By the way are we sure the `GamepadIndex` will always match the index in the gamepad list? I guess because of the handling of a previous connected event? How would that play out if the focused webview changes?

gterzian (changes requested):
Good progress. I propose we move some of the logic to methods of the navigator, for simplicity and reuse.

msub2:
Apologies for the lack of movement on this, had a busy couple of weeks. To the point on whether the gamepad index from gilrs will always match up to the calculated index, I don't believe this is the case, and that's partially what gave me pause on adding new commits earlier. I believe we'd need an internal variable to keep track of the mapping between Navigator gamepads and gilrs gamepads, especially if exposed gamepads were to differ between webviews. I'm wondering if we can maybe use the gamepad_id member for this, as it doesn't seem to be used in the current state of the code

components/script/dom/navigator.rs
```diff
@@ -81,9 +82,52 @@ impl Navigator {
...
     }
 
-    pub fn gamepads(&self) -> DomRoot<GamepadList> {
-        self.gamepads
-            .or_init(|| GamepadList::new(&self.global(), &[]))
+    pub fn get_gamepad(&self, index: usize) -> Option<DomRoot<Gamepad>> {
+        self.gamepads.borrow().get(index).and_then(|g| g.get())
+    }
+
+    pub fn set_gamepad(&self, index: usize, gamepad: Option<DomRoot<Gamepad>>) {
```
gterzian, line 89: I think it's a good idea to:

1. pass a `&Gamepad` instead of the root as argument.
2. Break this in two funtions, one `set_gamepad` and another `remove_gamepad`, that way the logic becomes easier to follow. We can then also move the call to `shrink_gamepads_list`(which can be private then) into `remove_gamepad`. Sorry didn't think about this until I saw it...

components/script/dom/globalscope.rs
```diff
@@ -3204,26 +3190,26 @@ impl GlobalScope {
...
                                 }
                             };
-
-                            if !window.Navigator().has_gamepad_gesture() && contains_user_gesture(update_type) {
-                                window.Navigator().set_has_gamepad_gesture(true);
-                                for i in 0..gamepad_list.Length() {
-                                    if let Some(gamepad) = gamepad_list.Item(i) {
+                            if !navigator.has_gamepad_gesture() && contains_user_gesture(update_type) {
+                                navigator.set_has_gamepad_gesture(true);
+                                navigator.GetGamepads().iter().for_each(|g| {
```
gterzian, line 3195: I think you can filter on it being something, to remove the conditional below?

components/script/dom/globalscope.rs
```diff
@@ -3204,26 +3190,26 @@ impl GlobalScope {
...
+                                    if let Some(gamepad) = g {
                                         gamepad.set_exposed(true);
                                         gamepad.update_timestamp(*current_time);
-                                        let new_gamepad = Trusted::new(&*gamepad);
+                                        let new_gamepad = Trusted::new(&**gamepad);
```
gterzian, line 3199: `connected_gamepad`

gterzian (changes requested):
Thanks for the refactoring, I think it's must easier to read now. I have a couple of last changes to suggest.

I propose we deal with the gamepad index stuff in another PR. I think we need a bit more infra to track gamepad across windows, where the index would remain the same. Also see this note in the spec at the disconnect event:

> When a gamepad is disconnected from the [user agent](https://infra.spec.whatwg.org/#user-agent), if the [user agent](https://infra.spec.whatwg.org/#user-agent) has previously dispatched a [gamepadconnected](https://www.w3.org/TR/gamepad/#dfn-gamepadconnected) event for that gamepad to a [Window](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window), a [gamepaddisconnected](https://www.w3.org/TR/gamepad/#dfn-gamepaddisconnected) event MUST be dispatched to that same [Window](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window).

components/script/dom/globalscope.rs
```diff
@@ -3121,7 +3121,7 @@ impl GlobalScope {
     /// <https://www.w3.org/TR/gamepad/#dfn-gamepadconnected>
     pub fn handle_gamepad_connect(
         &self,
-        index: usize,
+        _index: usize,
```
gterzian, line 3124: Let's add a note about what we discussed re indices, and why we are ignoring it for now.

components/script/dom/navigator.rs
```diff
@@ -81,9 +82,55 @@ impl Navigator {
...
 
-    pub fn gamepads(&self) -> DomRoot<GamepadList> {
-        self.gamepads
-            .or_init(|| GamepadList::new(&self.global(), &[]))
+    pub fn get_gamepad(&self, index: usize) -> Option<DomRoot<Gamepad>> {
+        self.gamepads.borrow().get(index).and_then(|g| g.get())
+    }
+
+    pub fn set_gamepad(&self, index: usize, gamepad: &Gamepad) {
+        if index >= self.gamepads.borrow().len() {
```
gterzian, line 90: I think this needs to be moved to inside `select_gamepad_index`(in the spec that is also where you see "Append null to ...").

components/script/dom/navigator.rs
```diff
@@ -81,9 +82,55 @@ impl Navigator {
...
+            self.gamepads
+                .borrow_mut()
+                .resize_with(index + 1, Default::default);
+        }
+        if let Some(gamepad_to_set) = self.gamepads.borrow().get(index) {
+            gamepad_to_set.set(Some(gamepad));
+            if self.has_gamepad_gesture.get() {
+                gamepad.set_exposed(true);
+                if self.global().as_window().Document().is_fully_active() {
+                    gamepad.notify_event(GamepadEventType::Connected);
```
gterzian, line 100: Let's move firing the event to outside the borrow, because the JS code in the event handler could call `navigator.getGamePads`, and we'd get a double-borrow panic.

gterzian (changes requested):
Very good progress, and just a few last points.

gterzian (approved):
Awesome, thanks!

PR #31691, opened 2024-03-15
Wrap aggregate reports in code blocks

### LINKED ISSUE #30103: Test results can ping random users
For example, https://github.com/servo/servo/pull/30099#issuecomment-1678463732 has results like

```
- OK /css/css-values/viewport-units-media-queries.html
  - PASS [expected FAIL] subtest: @media(width:100vi) applies 
```

where Github treats `@media` as a ping to a user, which may not be great for that user.

This can be avoided by using `@<!---->media`, which is the shown as: @media

sagudev:
This is to prevent random ping and to prevent markdown to misinterpret symbols (webgpu test names includes `*`).

---

- [x] These changes fix #30103
- [x] These changes do not require tests because it's just CI

### DIFF
--- a/etc/ci/report_aggregated_expected_results.py
+++ b/etc/ci/report_aggregated_expected_results.py
@@ -37,7 +37,7 @@ class Item:
     def from_result(cls, result: dict, title: Optional[str] = None, print_stack=True):
         expected = result["expected"]
         actual = result["actual"]
-        title = title if title else result["path"]
+        title = title if title else f'`{result["path"]}`'
         if expected != actual:
             title = f"{actual} [expected {expected}] {title}"
         else:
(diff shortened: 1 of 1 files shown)


sagudev:
We do wrong checkout in `Process WPT results` for this PR to work.

sagudev:
Here is example run for webgpu: https://github.com/sagudev/servo/actions/runs/8292891214/job/22696435106

mrobinson (approved):
This seems fine for now, but ultimately, I think it would be better to escape the markdown.

sagudev:
Full try run: https://github.com/sagudev/servo/actions/runs/8293417609/job/22698526184

sagudev:
> This seems fine for now, but ultimately, I think it would be better to escape the markdown.

On good thing about using ``` is that we get horizontal scroll instead of text wrap.

PR #31694, opened 2024-03-15
Add RUSTC env to clippy command

### LINKED ISSUE #31692: Crown should be disabled when running cargo clippy.
Currently clippy is used like this: `RUSTC=rustc ./mach cargo-clippy`, but we can move env override directly in https://github.com/servo/servo/blob/68b82e6d6133c8213cdad09b240d7a69227cc82a/python/servo/devenv_commands.py#L136 so there will be no need for manually writing `RUSTC=rustc`.

eerii:
Adds the env override directly to the `./mach cargo-clippy` command so we can run it without `RUSTC=rustc`.

Once merged: Update the description of #31500 removing `RUSTC=rustc`.

---
- [x] These changes fix #31692
- [x] These changes do not require tests because they involve a build tool, not the actual project

### DIFF
--- a/python/servo/devenv_commands.py
+++ b/python/servo/devenv_commands.py
@@ -139,7 +139,9 @@ class MachCommands(CommandBase):
 
         self.ensure_bootstrapped()
         self.ensure_clobbered()
-        return self.run_cargo_build_like_command("clippy", params, **kwargs)
+        env = self.build_env()
+        env['RUSTC'] = 'rustc'
+        return self.run_cargo_build_like_command("clippy", params, env, **kwargs)
 
     @Command('grep',
              description='`git grep` for selected directories.',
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

python/servo/devenv_commands.py
```diff
@@ -139,7 +139,9 @@ def cargo_clippy(self, params, **kwargs):
 
         self.ensure_bootstrapped()
         self.ensure_clobbered()
-        return self.run_cargo_build_like_command("clippy", params, **kwargs)
+        env = self.build_env()
+        env['RUSTC'] = 'rustc'
+        return self.run_cargo_build_like_command("clippy", params, env, **kwargs)
```
sagudev, line 144: I think it would be safer to have `env=env`.

PR #31705, opened 2024-03-15
Pick the first or last baseline as appropriate

### LINKED ISSUE #31651: Baseline alignment of inline-tables is broken, especially if anonymous
```html
<!DOCTYPE html>
<span style="border: solid">
  a
  <span style="display: inline-table">
    <span style="display: table-row">b</span>
    <span style="display: table-row">c</span>
  </span>
</span>
<span style="border: solid">
  a
  <span style="display: table-row">b</span>
  <span style="display: table-row">c</span>
</span>
```

| Servo | Firefox |
| - | - |
|  |  |

In the former case, Servo seems to be aligning everything to the bottom. But I'm not even sure why the 2nd case, with an anonymous table, looks different.

Loirooriol:
The old logic was always picking the last baseline, but this should only happen for inline-blocks.

Since replaced elements and flex containers aren't currently setting their baselines, this is only an improvement for inline-tables.

---

- [X] These changes fix #31651

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -101,7 +101,9 @@ use crate::cell::ArcRefCell;
 use crate::context::LayoutContext;
 use crate::flow::float::{FloatBox, SequentialLayoutState};
 use crate::flow::FlowLayout;
-use crate::formatting_contexts::{Baselines, IndependentFormattingContext};
+use crate::formatting_contexts::{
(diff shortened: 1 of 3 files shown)


mrobinson:
I'm curious which tests will start passing, so I've done a rebase of this change via the GitHub interface.

Loirooriol:
The failures are due to #31722.

mrobinson (approved):

Issue #31416, opened 2024-02-23
Layout sizes for video elements are too simplistic


jdm:
https://searchfox.org/mozilla-central/source/testing/web-platform/tests/html/semantics/embedded-content/the-video-element/intrinsic_sizes.htm tests various scenarios for sizing video elements. We don't implement the layout rules related to poster elements or aspect ratios at the moment, so we fail them all. The following is a patch that allows the first test (default object size) to pass:
```diff
diff --git a/components/layout/display_list/builder.rs b/components/layout/display_list/builder.rs
index 753b7747c9..c42fed5fea 100644
--- a/components/layout/display_list/builder.rs
+++ b/components/layout/display_list/builder.rs
@@ -1879,7 +1879,7 @@ impl Fragment {
                 }
             },
             SpecificFragmentInfo::Media(ref fragment_info) => {
-                if let Some((ref image_key, _, _)) = fragment_info.current_frame {
+                if let Some(ref image_key) = fragment_info.current_frame {
                     let base = create_base_display_item(state);
                     state.add_image_item(
                         base,
diff --git a/components/layout/fragment.rs b/components/layout/fragment.rs
index f22a0edb9a..7cdb81adae 100644
--- a/components/layout/fragment.rs
+++ b/components/layout/fragment.rs
@@ -385,13 +385,17 @@ impl CanvasFragmentInfo {

 #[derive(Clone)]
 pub struct MediaFragmentInfo {
-    pub current_frame: Option<(webrender_api::ImageKey, i32, i32)>,
+    pub current_frame: Option<webrender_api::ImageKey>,
+    pub width: i32,
+    pub height: i32,
 }

 impl MediaFragmentInfo {
     pub fn new(data: HTMLMediaData) -> MediaFragmentInfo {
         MediaFragmentInfo {
             current_frame: data.current_frame,
+            width: data.width,
+            height: data.height,
         }
     }
 }
@@ -1042,11 +1046,7 @@ impl Fragment {
                 }
             },
             SpecificFragmentInfo::Media(ref info) => {
-                if let Some((_, width, _)) = info.current_frame {
-                    Au::from_px(width)
-                } else {
-                    Au(0)
-                }
+                Au::from_px(info.width)
             },
             SpecificFragmentInfo::Canvas(ref info) => info.dom_width,
             SpecificFragmentInfo::Svg(ref info) => info.dom_width,
@@ -1072,11 +1072,7 @@ impl Fragment {
                 }
             },
[...]

PR #31706, opened 2024-03-15
Implement missing ready states transition

### LINKED ISSUE #31415: Short videos dispatch loadedmetadata before the metadata is present
This is the cause of the failure in the WPT test [video_size_preserved_after_ended.html](https://searchfox.org/mozilla-central/source/testing/web-platform/tests/html/semantics/embedded-content/the-video-element/video_size_preserved_after_ended.html). The video's actual size is obtained when gstreamer sends a player event with the video metadata present (https://github.com/servo/servo/blob/4849ba901efab9304d71b316ec9e0d7e98e1993b/components/script/dom/htmlmediaelement.rs#L1704-L1705). For most videos, this causes Servo to dispatch the loadedmetadata DOM event because we change the element's ready state to HaveMetadata.

However, HTMLMediaElement has a special case when the network response EOF is reached and the element's ready state is still HaveNothing (https://github.com/servo/servo/blob/4849ba901efab9304d71b316ec9e0d7e98e1993b/components/script/dom/htmlmediaelement.rs#L2798-L2802). We change the ready state to HaveMetadata without waiting for the media player's metadata event, so the video element's properties do not reflect the actual metadata yet.

pylbrecht:
---

- [ ] These changes fix #31415

### DIFF
--- a/components/script/dom/audiotracklist.rs
+++ b/components/script/dom/audiotracklist.rs
@@ -107,6 +107,12 @@ impl AudioTrackList {
         track.add_track_list(self);
     }
 
+    pub fn remove(&self, index: usize) -> DomRoot<AudioTrack> {
+        let track: DomRoot<AudioTrack> = DomRoot::from_ref(&self.tracks.borrow_mut().remove(index));
+        track.remove_track_list();
+        track
+    }
+
     pub fn clear(&self) {
         self.tracks
             .borrow()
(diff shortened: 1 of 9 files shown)


eerii:
I tested the changes locally and they seem to be going in a great direction. At least it fixes the problem I was having in https://github.com/servo/servo/issues/31416#issuecomment-1987275602 with the `onloadedmetadata` event firing before the metadata was loaded (and thus failing the size asserts)!

pylbrecht:
Fixed in https://github.com/servo/servo/pull/32643.

PR #31708, opened 2024-03-16
rustdoc: Fix warning in `components/layout_2020`

Aaryakhandelwal:
Fixed the unresolved links present in components/layout_2020 
---

- [X] These changes are part of #31575 (GitHub issue number if applicable)
- [X] These changes do not require tests because they only change documentation and no functionality.

### DIFF
--- a/components/layout_2020/table/construct.rs
+++ b/components/layout_2020/table/construct.rs
@@ -477,7 +477,7 @@ impl TableBuilder {
 
     /// When not in the process of filling a cell, make sure any incoming rowspans are
     /// filled so that the next specified cell comes after them. Should have been called before
-    /// [`Self::handle_cell`].
+    /// [`Self::add_cell`]
     ///
     /// if `stop_at_cell_opportunity` is set, this will stop at the first slot with
     /// `incoming_rowspans` equal to zero. If not, it will insert [`TableSlot::Empty`] and


Rhea-Eve:
Sorry, I didn't see this before I added my own pr. I resolved some of the broken references differently.  See #31712.

components/layout_2020/flow/line.rs
```diff
@@ -32,7 +32,7 @@ use crate::style_ext::PaddingBorderMargin;
 use crate::ContainingBlock;
 
 pub(super) struct LineMetrics {
-    /// The block offset of the line start in the containing [`InlineFormattingContext`].
+    /// The block offset of the line start in the containing Inline Formatting Context.
```
mrobinson, line 35: I think you can just qualify the reference to InlineFormattingContext instead of removing the link.

components/layout_2020/formatting_contexts.rs
```diff
@@ -59,7 +59,7 @@ pub(crate) enum NonReplacedFormattingContextContents {
     // Other layout modes go here
 }
 
-/// The baselines of a layout or a [`BoxFragment`]. Some layout uses the first and some layout uses
+/// The baselines of a layout or [`Fragment::Box`] Some layout uses the first and some layout uses
```
mrobinson, line 62: Again, I think this just needs to be qualified. `Fragment::Box` is an enum variant, but this is actually referring to the struct stored inside the enum variant `BoxFragment`.

mrobinson (changes requested):

Rhea-Eve:
@mrobinson I think that third edit that @Aaryakhandelwal suggested seems to be accurate. She changed [`Self::handle_cell`] to [`Self::add_cell`]. After looking at it, this was my hunch too, but I wasn't entirely sure. Currently, there does not seem to be a function called handle_cell. And it seems that the comment was added at the same time as add_cell.

Aaryakhandelwal:
@mrobinson , Am I still supposed to make changes, as the next PR created by @Rhea-Eve already handles these both errors correctly.

mrobinson:
> @mrobinson , Am I still supposed to make changes, as the next PR created by @Rhea-Eve already handles these both errors correctly.

Let's land both PRs. I suggest just updating this PR to remove the first two changes.

Aaryakhandelwal:
> > @mrobinson , Am I still supposed to make changes, as the next PR created by @Rhea-Eve already handles these both errors correctly.
> 
> Let's land both PRs. I suggest just updating this PR to remove the first two changes.

Yes, I will do that

mrobinson (approved):

PR #31713, opened 2024-03-17
minibrowser: Add loading spinner 

### LINKED ISSUE #31689: servoshell: add a spinner that indicates when the page is loading
When the minibrowser is enabled, add an [egui spinner](https://docs.rs/egui/0.22.0/egui/struct.Ui.html#method.spinner) to the location bar when the page is loading.

Watch for `EmbedderMsg::LoadStart`, `EmbedderMsg::HeadParsed`, and `EmbedderMsg::LoadComplete` events, and keep track of whether the page is currently loading in `Minibrowser`.

This may be a good [newcomer issue](https://github.com/servo/servo/labels/B-newcomer) once #31688 is fixed.

image

frereit:
I've started to implement a simple spinner that indicates when a page is loading. This PR is not complete yet but I'd like to get some feedback and discuss some implementation details if that's possible.

For this "MVP", I've simply added a new bool `loading_changed` analogous to the `history_changed` field in `PumpResult::Continue`. The `WebView` then keeps track of wether or not it is loading and exposes this via `WebView::loading()`, like `WebView::shutdown_requested`. The `Minibrowser` then gets a notification from the `App` when `loading` changed and updates the spinner accordingly.

A few questions arose while implementing this:

- The comments in the code and issue indicate that `EmbedderMsg::HeadParsed` should also "influence" the loading state, but I do not understand how this should surface. In my understanding, the spinner simply shows up when the document starts loading and stops once it is complete. What is the intended "intermediary" step here?

https://github.com/servo/servo/blob/8cfc6a1898d02eb1df9c13c3c410747cb1e0b412/ports/servoshell/webview.rs#L596-L598

- It seems a bit clunky to add a `xyz_changed` for everything about the `WebView` that could be required to bubble up to the `Minibrowser`. For example, showing the Favicon might also require a `favicon_changed: bool` in the future. Perhaps it would make sense to combine all these into a `toolbar_data_changed` or `metadata_changed` or something and let the `Minibrowser` figure out what it has to do (or just refresh the entire toolbar when anything changed, that shouldn't really impact performance). Is this something I should implement in this PR?

---

- [X] These changes fix #31689 

- [X] These changes do not require tests because it is purely visual in the minibrowser.

### DIFF
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -7,7 +7,7 @@ use std::num::NonZeroU32;
 use std::sync::Arc;
 use std::time::Instant;
 
-use egui::{CentralPanel, Frame, Key, Modifiers, PaintCallback, TopBottomPanel};
+use egui::{CentralPanel, Color32, Frame, Key, Modifiers, PaintCallback, Spinner, TopBottomPanel};
 use egui_glow::CallbackFn;
 use egui_winit::EventResponse;
 use euclid::{Length, Point2D, Scale};
@@ -24,7 +24,7 @@ use crate::egui_glue::EguiGlow;
 use crate::events_loop::EventsLoop;
 use crate::geometry::winit_position_to_euclid_point;
 use crate::parser::location_bar_input_to_url;
-use crate::webview::WebViewManager;
+use crate::webview::{LoadStatus, WebViewManager};
 use crate::window_trait::WindowPortsMethods;
 
 pub struct Minibrowser {
@@ -42,6 +42,8 @@ pub struct Minibrowser {
 
     /// Whether the location has been edited by the user without clicking Go.
     location_dirty: Cell<bool>,
+
+    load_status: LoadStatus,
 }
 
 pub enum MinibrowserEvent {
@@ -83,6 +85,7 @@ impl Minibrowser {
             last_mouse_position: None,
             location: RefCell::new(initial_url.to_string()),
             location_dirty: false.into(),
+            load_status: LoadStatus::LoadComplete,
         }
     }
 
@@ -160,6 +163,16 @@ impl Minibrowser {
                                 if ui.button("go").clicked() {
                                     event_queue.borrow_mut().push(MinibrowserEvent::Go);
                                     location_dirty.set(false);
+                                }
+
+                                match self.load_status {
+                                    LoadStatus::LoadStart => {
+                                        ui.add(Spinner::new().color(Color32::GRAY));
+                                    },
+                                    LoadStatus::HeadParsed => {
+                                        ui.add(Spinner::new().color(Color32::WHITE));
+                                    },
+                                    LoadStatus::LoadComplete => { /* No Spinner */ },
                                 }
 
                                 let location_field = ui.add_sized(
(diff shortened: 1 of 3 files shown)


ports/servoshell/minibrowser.rs
```diff
@@ -314,4 +319,13 @@ impl Minibrowser {
...
         }
     }
+
+    /// Updates the spinner from the given [WebViewManager] returning true iff a redraw is needed (always).
+    pub fn update_spinner_in_toolbar(
+        &mut self,
+        browser: &mut WebViewManager<dyn WindowPortsMethods>,
+    ) -> bool {
+        self.loading = browser.loading();
+        return true;
```
frereit, line 329: Currently this always forces a `redraw`. I attempted to only redraw when `self.loading` changed like this:

```suggestion
        let need_redraw = self.loading != browser.loading();
        self.loading = browser.loading();
        return need_redraw;
```

But that resulted in some weird behavior where the spinner wasn't always being drawn. I'll investigate why this is the case but wanted to open the PR already to get feedback on the two questions in the other comments, I hope that's okay.

jdm:
>What is the intended "intermediary" step here?

One common pattern I've seen is a slow spinner in one direction while the initial network request is in flight, the the spinner changes direction or speeds up when the beginning of the response is received. This corresponds to the HeadParsed event.

delan:
> > What is the intended "intermediary" step here?
> 
> One common pattern I've seen is a slow spinner in one direction while the initial network request is in flight, then the spinner changes direction or speeds up when the beginning of the response is received. This corresponds to the HeadParsed event.

Yeah, HeadParsed happens somewhere between LoadStart and LoadComplete. It would be good to show these two periods differently, but if we can’t, then we should treat HeadParsed like LoadStart.

> It seems a bit clunky to add a `xyz_changed` for everything about the `WebView` that could be required to bubble up to the `Minibrowser`. For example, showing the Favicon might also require a `favicon_changed: bool` in the future. Perhaps it would make sense to combine all these into a `toolbar_data_changed` or `metadata_changed` or something and let the `Minibrowser` figure out what it has to do (or just refresh the entire toolbar when anything changed, that shouldn't really impact performance). Is this something I should implement in this PR?

Good point! Since we only use our history_changed flag to check if we should run an egui update (and maybe update some Minibrowser fields), I agree that it would be good to combine them.

frereit:
> > > What is the intended "intermediary" step here?
> > 
> > 
> > One common pattern I've seen is a slow spinner in one direction while the initial network request is in flight, then the spinner changes direction or speeds up when the beginning of the response is received. This corresponds to the HeadParsed event.
> 
> Yeah, HeadParsed happens somewhere between LoadStart and LoadComplete. It would be good to show these two periods differently, but if we can’t, then we should treat HeadParsed like LoadStart.

Done. During the `LoadStart` -> `HeadParsed` period the spinner is now `Color32::GRAY` and the default white otherwise.

> > It seems a bit clunky to add a `xyz_changed` for everything about the `WebView` that could be required to bubble up to the `Minibrowser`. For example, showing the Favicon might also require a `favicon_changed: bool` in the future. Perhaps it would make sense to combine all these into a `toolbar_data_changed` or `metadata_changed` or something and let the `Minibrowser` figure out what it has to do (or just refresh the entire toolbar when anything changed, that shouldn't really impact performance). Is this something I should implement in this PR?
> 
> Good point! Since we only use our history_changed flag to check if we should run an egui update (and maybe update some Minibrowser fields), I agree that it would be good to combine them.

Done.

I've also changed the `webview` to always return `need_present: true` when the spinner should be active because the spinner is animated and requires regular redraws to not be choppy. Is this the idiomatic way to achieve this or should the redraw be requested some other way?

ports/servoshell/minibrowser.rs
frereit: This is resolved by the new commit, whch implements this without problems.

ports/servoshell/minibrowser.rs
```diff
@@ -314,4 +323,28 @@ impl Minibrowser {
...
         }
     }
+
+    /// Updates the spinner from the given [WebViewManager], returning true iff it has changed
+    /// (needing an egui update).
+    pub fn update_spinner_in_toolbar(
+        &mut self,
+        browser: &mut WebViewManager<dyn WindowPortsMethods>,
+    ) -> bool {
+        let need_update = browser.load_status() == self.load_status;
```
delan, line 333: should this be `!=`?

delan (commented):
Excellent work! One question…

delan:
> I've also changed the `webview` to always return `need_present: true` when the spinner should be active because the spinner is animated and requires regular redraws to not be choppy. Is this the idiomatic way to achieve this or should the redraw be requested some other way?

This makes sense and seems good enough to me.

ports/servoshell/minibrowser.rs
frereit: You're right, thanks!

delan (approved):
Great, thanks!

frereit:
@delan Did you mean to only enable auto-merge without triggering the CI run? Is there something I need to do to finish up this PR?

frereit:
Apologies for the format errors. Fixed.

PR #31720, opened 2024-03-17
Use --locked when building in CI

Loirooriol:
Thus the build will immediately fail if a PR modifies Cargo.toml but forgets to include the changes in Cargo.lock

This was previously checked by lockfile_changed.sh after building normally, wasting resources.

---

- [X] These changes do not require tests because it's a CI change


sagudev:
Don't we have this check already?
https://github.com/servo/servo/blob/52c2b1e3a8f57af9885bbc52037fc1cad42f7dae/.github/workflows/linux.yml#L127-L128

Loirooriol:
Mm, good point. But the advantage of `--locked` is that it will immediately fail.
It takes a while to build, and the number of runners is limited. So failing fast seems better.
But I guess I can remove that `lockfile_changed.sh`.

mrobinson (approved):
Nice.

sagudev (approved):

PR #31721, opened 2024-03-17
clippy: Fix warnings in generated code

eerii:
Fixes some clippy warnings from the code generated by `CodegenRust.py`. It specifically gets rids of all _errors_ from these files.

Due to the nature of the generated code, some allowances had to be made, sometimes because of false positives, and sometimes because "fixing" the warning would be very inefficient or require changing how everything generates. The allows and their reasoning are the following:

- `approx_constant`: Values defined in `.webidl` files could be related to constants such as pi. In rust it would be better to use defined types such as `f64::consts::PI`, but because these are taken from other files they should stay as numeric values.
- `let_unit_value`: In `CGCallGenerator` the value of `result` can take the `()` type. This is intended, and while specific allows could be placed for every let binding easily, this generates too many lines so it may be better to allow it for the entire file. There is also discussion regarding this lint in https://github.com/rust-lang/rust-clippy/issues/9048.
- `needless_return`: In `CGPerSignatureCall`, the call to `wrap_return_value` can be straightforward, so a simple `true` at the end could suffice, but it could also contain nested matches, so `return true;` becomes necessary. Distinguishing between all the possible cases seems overly complicated when `return true;` works and isn't much more verbose. The other instances of returning were addressed.
- `too_many_arguments`: Like in other cases, some functions generate with too many arguments. It seems counterproductive to check every function to know if it has more than x arguments and add an allow for it automatically when it can be permitted for every file.
- `unnecessary_cast`: This is specifically for `MethodDefiner::generateArray`. There is a comment specifying that it's not easy to tell whether the methodinfo is a `JSJitInfo` or a `JSTypedMethodJitInfo`, so we cast them both to `JSJitInfo` and rely on the compiler doing the conversion. This creates many `unnecessary_cast` warnings for `JSJitInfo`, but it is documented and not trivial to fix.

---
- [x] These changes do not require tests because they do not change functionality, only fix warnings

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -420,14 +420,14 @@ class CGMethodCall(CGThing):
                 caseBody.append(CGGeneric("if %s.get().is_object() {" %
                                           (distinguishingArg)))
                 for idx, sig in enumerate(interfacesSigs):
-                    caseBody.append(CGIndenter(CGGeneric("loop {")))
+                    caseBody.append(CGIndenter(CGGeneric("'_block: {")))
                     type = sig[1][distinguishingIndex].type
 
                     # The argument at index distinguishingIndex can't possibly
                     # be unset here, because we've already checked that argc is
                     # large enough that we can examine this argument.
                     info = getJSToNativeConversionInfo(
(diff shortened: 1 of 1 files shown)


jdm (approved):
Nice simplifications!

PR #31728, opened 2024-03-18
Generated event tracing targets

delan:
---


delan:
Closing; this doesn’t eliminate the repetition that causes breakage when changing enum variants, only reduces it, so it’s not really worth the effort.

PR #31734, opened 2024-03-18
layout: Enable `text-shadow`

mrobinson:
`text-shadow` is enabled for legacy layout, so this just extends support to the modern layout system.

/css/css-pseudo/highlight-currentcolor-painting-text-shadow-001.html starts failing because Servo does not support `::highlight()`.

---


mrobinson:
After looking at some of the remaining failures, it looks like we have to make one tricky fix here. Shadows should carry with them the color of the shadow from the moment they were specified in style. Currently we resolve the color against the element we are applying them to instead.

components/layout_2020/display_list/mod.rs
```diff
@@ -414,6 +432,10 @@ impl Fragment {
             rect.size.height = Length::new(font_metrics.strikeout_size.to_nearest_pixel(dppx));
             self.build_display_list_for_text_decoration(fragment, builder, &rect, &color);
         }
+
+        if !shadows.0.is_empty() {
+            builder.wr().pop_all_shadows();
```
Loirooriol, line 437: Why is this needed?

mrobinson:
> After looking at some of the remaining failures, it looks like we have to make one tricky fix here. Shadows should carry with them the color of the shadow from the moment they were specified in style. Currently we resolve the color against the element we are applying them to instead.

It turns out I misunderstood the issue. It seems that I was observing #31736 and it's unrelated to `text-shadow`.

components/layout_2020/display_list/mod.rs
mrobinson: In WebRender you push shadows and pop them. All display items that you add to the display list between a push and a pop get a shadow and are shadowed together to properly preserve layering.

Loirooriol (approved):

mrobinson:
The failure here is #31754.

PR #31735, opened 2024-03-18
clippy: Fix some warnings in `components/script`

six-shot:
---

- [x] These changes are part of #31500  (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/identityhub.rs
+++ b/components/script/dom/identityhub.rs
@@ -64,8 +64,8 @@ pub struct Identities {
     dummy_hub: IdentityHub,
 }
 
-impl Default for Identities {
-    fn default() -> Self {
+impl Identities {
+    pub fn new() -> Self {
         Identities {
             _surface: IdentityManager::default(),
             #[cfg(any(target_os = "linux", target_os = "windows"))]
@@ -91,8 +91,8 @@ impl Default for Identities {
             #[cfg(any(target_os = "ios", target_os = "macos"))]
             Backend::Metal => &mut self.metal_hub,
             _ => &mut self.dummy_hub,
-        }
-    }
+        }}
+
 
     fn hubs(&mut self) -> Vec<(&mut IdentityHub, Backend)> {
         vec![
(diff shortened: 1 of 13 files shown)


components/script/dom/identityhub.rs
```diff
@@ -64,8 +64,8 @@ pub struct Identities {
     dummy_hub: IdentityHub,
 }
 
-impl Identities {
-    pub fn new() -> Self {
+impl Default for Identities {
+    fn default() -> Self {
```
mrobinson, line 68: You also need to update all references to `Identities::new()` to be `Identities::default()` if you change the name of the method.

mrobinson (changes requested):

mrobinson:
Does this build locally? Before putting the "x" in the box that says " ./mach build -d does not report any errors" please actually run `./mach build` locally.

six-shot:
I apologize sir I thought I did run the build for this PR I’ll fix it right away

six-shot:
@mrobinson Please review this again

components/script/dom/mediafragmentparser.rs
```diff
@@ -114,8 +114,8 @@ impl MediaFragmentParser {
             }
         } else {
             let mut iterator = fragment.split(',');
-            let start = parse_hms(iterator.next().ok_or_else(|| ())?)?;
-            let end = parse_hms(iterator.next().ok_or_else(|| ())?)?;
+            let start = parse_hms(iterator.next().ok_or(()))?;
```
eerii, line 117: I think the build error you are having originates here, since it expects an `&str` but it is receiving a `Result`. One way to solve it would be adding back the second `?`, as it will unwrap the `Result` type. You can probably also get rid of the `ok_or(())` operator since `?` should also work for `Option` types and we are not using the error type. This goes for lines 117, 118, 317, 318, 322 and 333.

six-shot: Yeah I fixed  all 6 of them  already…  I’m trying to build it before pushing to be sure there no more errors Thanks @eerii

six-shot: The PR is ready to go now

eerii: A member from the servo team will have to approve the pr but all the changes are looking good to me, thanks!

mrobinson (approved):

PR #31737, opened 2024-03-18
Attempt using version of Stylo with -x-lang enabled

### LINKED ISSUE #31723: Enable the internal -x-lang CSS property
The behavior of `text-transform` can be affected by the language of the text, e.g.

```html
<div style="text-transform: uppercase">
  <span lang="tr">i</span> <!-- İ -->
  <span lang="en">i</span> <!-- I -->
</div>
```

Therefore, we need some way to track the language.

Gecko uses an internal `-x-lang` CSS property for that, so it seems a good idea to do the same (especially since we use the same style engine).

To do that,

 1. Clone a fork of https://github.com/servo/stylo/
 1. Enable `-x-lang` not just for `gecko` but also `servo`.
   https://github.com/servo/stylo/blob/7dd884031d6b97d801d50f8fcdff69bbeb292ae0/style/properties/longhands/font.mako.rs#L257-L266
 1. Run `cargo build --features servo` to ensure it builds fine
 1. Open a PR at https://github.com/servo/stylo/
 1. Here in Servo, update `Cargo.toml` to take the various style-related crates from your PR. Compile, this will automatically update `Cargo.lock`. See 871e36f5a5f8d87cd8d3979794e8e21588f54939 for reference.
 1. In this case I don't expect anything to break, but otherwise you would need to do the changes necessary so that it still compiles fine.
 1. Open a PR with the Servo changes, so that we can run the tests and ensure everything is fine. `-x-lang` is an internal property, so I don't expect any test to complain.
 1. Then we can land the PR for Stylo
 1. Then update the PR for Servo, reverting the changes from `Cargo.toml`, but bumping `Cargo.lock` to point to the new upstream rev. See https://github.com/servo/servo/commit/0935bf52fd17e05c0ef4f0a2ef560dca32a4fa23 for reference.
 1. Finally we can land the PR for Servo

Rhea-Eve:
This is step 7 of #31723.

Note that my Stylo fork tracks stylo/main, not the branch that was previously used in Cargo.toml. If this poses issues, I can change it.

---

- [X] These changes fix #31723(GitHub issue number if applicable)


Loirooriol:
It's failing the unit tests. You can run them with `./mach test-unit`

```
error[E0063]: missing field `_x_lang` in initializer of `style::properties::style_structs::Font`
   --> components/gfx/tests/font_context.rs:106:21
    |
106 |     let mut style = FontStyleStruct {
    |                     ^^^^^^^^^^^^^^^ missing `_x_lang`
```

Rhea-Eve:
Thank you! I added a commit.

Loirooriol:
BTW, it's typically better to create branches in your fork, instead of committing to main. This way you can work on multiple PRs independently.

Loirooriol (approved):
Looks good if you revert `Cargo.toml` and update `Cargo.lock`

Rhea-Eve:
Done! Thank you for your help.

Loirooriol:
You will need to rebase, there are merge conflicts.

Loirooriol (changes requested):
Cargo.lock shouldn't modify unrelated crates

Rhea-Eve:
Yeah, something went wrong. Retrying now.

Rhea-Eve:
I now tried to resolve the merge conflict by manually editing Cargo.lock and it looks right.

Loirooriol:
Note you are not supposed to edit Cargo.lock manually.
What I do is: (there is probably a better way with `cargo update`)
1. Change Cargo.toml (e.g. patch the crates to be local, or point to a specific rev from github)
2. Compile, so that Cargo.lock gets updated
3. Undo the changes in Cargo.toml
4. Compile again.

Then Cargo.lock will point to the latest rev in the specified branch.

Loirooriol (approved):

Rhea-Eve:
> Note you are not supposed to edit Cargo.lock manually. What I do is: (there is probably a better way with `cargo update`)
> 
> 1. Change Cargo.toml (e.g. patch the crates to be local, or point to a specific rev from github)
> 2. Compile, so that Cargo.lock gets updated
> 3. Undo the changes in Cargo.toml
> 4. Compile again.
> 
> Then Cargo.lock will point to the latest rev in the specified branch.

This makes sense. Thanks for the tip!

PR #31738, opened 2024-03-18
rustdoc: fix warnings in `components/layout_2020/table`

azharcodeit:
Fix rustdoc warning about private intra-doc links

- Allowed private intra doc links by #![allow(rustdoc::private_intra_doc_links)]
- This resolves the warnings:
-- warning: public documentation for `node_id` links to private item `BaseFragmentInfo`
-- warning: public documentation for `table` links to private item `layout::TableLayout`

---

- [x] These change is part of #31575 

- [x] These changes do not require tests because they do not modify functionality

### DIFF
--- a/components/layout_2020/table/mod.rs
+++ b/components/layout_2020/table/mod.rs
@@ -46,8 +46,8 @@
 //! Fragment tree construction involves calculating the size and positioning of all table elements,
 //! given their style, content, and cell and row spans. This happens both during intrinsic inline
 //! size computation as well as layout into Fragments. In both of these cases, measurement and
-//! layout is done by [`layout::TableLayout`], though for intrinsic size computation only a partial
-//! layout is done.
+//! layout is done by the layout module located in components/layout_2020/table/layout.rs,
+//! though for intrinsic size computation only a partial layout is done.
 //!
 //! In general, we follow the following steps when laying out table content:
 //!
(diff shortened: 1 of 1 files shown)


components/layout_2020/table/mod.rs
```diff
@@ -196,7 +196,8 @@ impl TableSlotCell {
         }
     }
 
-    /// Get the node id of this cell's [`BaseFragmentInfo`]. This is used for unit tests.
```
atbrakhi, line 199: In Rust doc "[&#96;BaseFragmentInfo&#96;]" is a link, and unfortunately, if you remove that, there won't be any link, which is not what we want

azharcodeit: > In Rust doc "[`BaseFragmentInfo`]" is a link, and unfortunately, if you remove that, there won't be any link, which is not what we want

hey, just to confirm, should I keep the private link? Otherwise, I already replaced [`BaseFragmentInfo`] with its path: components/layout_2020/fragment_tree/base_fragment.rs

mrobinson: I think you should be able to replace this with something like `crate::fragment_tree::BaseFragmentInfo`.

azharcodeit: > I think you should be able to replace this with something like `crate::fragment_tree::BaseFragmentInfo`.

Any link seems to be treated as private link and cause warning. Can I allow private links instead and keep the initial link? 

#![allow(rustdoc::private_intra_doc_links)]

mrobinson: @azharcodeit Yep, you are right! The right thing to do here is to use `#![allow(rustdoc::private_intra_doc_links)]`.

mrobinson (approved):

mrobinson:
@azharcodeit Thank you!

PR #31745, opened 2024-03-18
rustdoc: Fix broken link to `FontFace` in `components/gfx`

oluwatobiss:
Add the MDN documentation link referenced in the [rustwasm](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.FontFace.html#) doc to resolve the `FontFace` broken link warning.

---

- [X] These changes are part of #31575 (GitHub issue number if applicable)
- [X] These changes do not require tests because they are documentation fixes.

### DIFF
--- a/components/gfx/platform/windows/font_template.rs
+++ b/components/gfx/platform/windows/font_template.rs
@@ -79,11 +79,8 @@ impl FontTemplateData {
         self.font_data.read().unwrap().as_ref().cloned()
     }
 
-    /// Get a [`FontFace`] for this font if it is a local font or return `None` if it's a
-    /// web font.
-    ///
-    /// [`FontFace`]: https://developer.mozilla.org/en-US/docs/Web/API/FontFace
-    ///
+    /// Get a [`Font`] for this font if it is a local font or return `None`
+    /// if it's a web font.
     pub fn get_font(&self) -> Option<Font> {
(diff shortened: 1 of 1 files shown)


components/gfx/platform/windows/font_template.rs
```diff
@@ -81,6 +81,9 @@ impl FontTemplateData {
 
     /// Get a [`FontFace`] for this font if it is a local font or return `None` if it's a
     /// web font.
+    ///
+    /// [`FontFace`]: https://developer.mozilla.org/en-US/docs/Web/API/FontFace
+    ///
```
mrobinson, line 86: This isn't the `FontFace` that's referred to here. You can just use `dwrote::FontFace`, I think.

mrobinson (changes requested):

components/gfx/platform/windows/font_template.rs
mrobinson: Actually, the issue here is that the method is returning `Font`, but I had written `FontFace` when I added this documentation because it changed while refactoring.

mrobinson (approved):

components/gfx/platform/windows/font_template.rs
oluwatobiss: Got it. Thanks for letting me know.

PR #31746, opened 2024-03-18
Improve how intrinsic sizes work for videos

### LINKED ISSUE #31416: Layout sizes for video elements are too simplistic

eerii:
Implements improved layout rules for video elements, including aspect ratios and removed poster/src elements. The base for this pr is the patch provided by @jdm in https://github.com/servo/servo/issues/31416#issue-2150465240.

The pr fixes the result of the wpt tests defined in [intrinsic_sizes.htm](https://github.com/servo/servo/blob/main/tests/wpt/tests/html/semantics/embedded-content/the-video-element/intrinsic_sizes.htm). The relevant natural width/height specification is available at https://html.spec.whatwg.org/multipage/media.html#concept-video-intrinsic-width.

|Before (main [ad8ba49](https://github.com/servo/servo/commit/ad8ba49d2cca9de554a9baa3c81188da91b7ec13))|After|
|--|--|
|||

~This change depends on #31706, since that fixes the `onloadedmetadata` event firing correctly. This was tested locally using that patch and all the tests pass. However, until that is merged, **the 4th test will fail**.~ (This is fixed)

---
- [x] These changes fix #31416


eerii:
I remembered I had this PR laying around, and it is no longer blocked by `onloadedmetadata`, so I rebased it into main and I fixed one test case that was still broken. It should be ready for review now!

eerii:
Ok, good information on the tests. After doing some testing on the unexpected results:

- `/css/css-images/object-position-svg-001p.html` and `/css/css-images/object-position-svg-002p.html` also fail for me, the poster changes must have some issue.
- `/css/css-sizing/intrinsic-size-fallback-replaced.html` is passing some subtests now which is good, since they relate to expected width and height.
- Even if `/html/rendering/replaced-elements/attributes-for-embedded-content-and-images/video-aspect-ratio.html` is passing here, it is still failing locally for me. The issue seems to be that it is not resetting the size after load with `width: auto`.
- `/html/semantics/embedded-content/the-video-element/intrinsic_sizes.htm` works for me locally, but the test fail seems to also be connected to load.
- `/html/semantics/embedded-content/the-video-element/video_content_image.htm` and `/html/semantics/embedded-content/the-video-element/video_content_text.htm` seem to be working correctly now.

jdm:
`/_mozilla/mozilla/iframe/resize_after_load.html` should be unrelated to these changes and you can ignore it.

components/layout_2020/display_list/mod.rs
```diff
@@ -291,14 +291,18 @@ impl Fragment {
...
+                            &common,
+                            rect,
+                            image_rendering,
+                            wr::AlphaType::PremultipliedAlpha,
+                            image_key,
+                            wr::ColorF::WHITE,
+                        );
+                    } else {
+                        builder.wr().push_rect(&common, rect, wr::ColorF::WHITE);
+                    }
```
eerii: The issue with the poster test came from adding a default white background if there was no poster image. Since Servo is not rendering the SVG's yet, the video background differed from the reference.

image

How it looks in Firefox:

image

eerii:
The issue with the load seems to come from `handle_player_event` in `htmlmediaelement`. When getting the `MetadataUpdated` event, it iterates through the video and audio tracks. It calls `VideoTracks.set_selected()`, which in turn calls `media_element.set_video_track`. The issue is that this function is calling `player.set_video_track`, locking the player, and it hangs indefinitely when it calls `gst_player::Player::set_video_track`. Because of this, it never reaches the `ReadyState::HaveMetadata`. I'm going to go to `gstreamer-rs` and check what could be happening.

eerii:
I made some changes and I think the tests might be fixed. There was a big problem in `LayoutHTMLVideoElementHelpers::data, since the video height and width were always being overwritten to be the named / calculated one. This wasn't a problem when there was a frame being rendered, as `get_current_frame_data()` took precedence.

However, on `PlayerEvent::MetadataUpdated` it was also modifying the video size to match the metadata. This didn't even last one frame, and then went back to the default, that's why `video-aspect-ratio` (and presumably `intrinsic_sizes`) were failing. Moving the logic from there to `set_intrinsic_size` and calling it only when attributes are mutated prevents the overwriting of values.

Additionally, `MetadataUpdated` was the only place that was emitting the `onresize` event, which [should fire](https://html.spec.whatwg.org/multipage/media.html#concept-video-intrinsic-width) every time the natural height / width changes. To avoid issues like this in the future I refactored `set_video_width` and `set_video_height` into a `resize` function that checks if the dimensions are different and, if they are, it fires the `onresize` event.

I noticed that the `resize-during-playback` test was also failing, and since I was on it, I also fixed the logic for emitting `onresize` when the media frame changes size. Since all of that is handled by `MediaFrameRenderer` and it doesn't have access to the window, I added an extra field in the `current_frame` struct tracking the resizing of the frame. Then, it is read during `PlayerEvent::VideoFrameUpdated`, which calls `HTMLVideoElement::resize` if it is necessary. This seems to make the test pass for webm, mp4 doesn't run since it returns that it is not supported.

components/script/dom/htmlvideoelement.rs
```diff
@@ -280,11 +316,32 @@ impl VirtualMethods for HTMLVideoElement {
...
+                AttributeMutation::Set(_) => {
+                    if let Some(new_value) = mutation.new_value(attr) {
+                        self.fetch_poster_frame(&new_value, CanGc::note())
+                    }
+                },
+                AttributeMutation::Removed => {
+                    self.set_intrinsic_size();
+                    self.htmlmediaelement.clear_current_frame();
+                },
+            },
```
Loirooriol, line 340: `mutation.new_value(attr)` being `Some(_)` seems equivalent to checking `AttributeMutation::Set(_)`.

```suggestion
            local_name!("poster") => {
                if let Some(new_value) = mutation.new_value(attr) {
                    self.fetch_poster_frame(&new_value, CanGc::note())
                } else {
                    self.set_intrinsic_size();
                    self.htmlmediaelement.clear_current_frame();
                }
            },
```

components/script/dom/htmlvideoelement.rs
```diff
@@ -280,11 +316,32 @@ impl VirtualMethods for HTMLVideoElement {
...
+            local_name!("src") => {
+                if matches!(mutation, AttributeMutation::Removed) {
```
Loirooriol, line 342:
```suggestion
                if mutation.is_removal() {
```

But shouldn't this also happen when `src` is set to a new value?

components/script/dom/htmlvideoelement.rs
```diff
@@ -100,20 +102,54 @@ impl HTMLVideoElement {
...
+                Some(len.to_px() as u32)
+            })
     }
 
-    pub fn set_video_height(&self, height: u32) {
-        self.video_height.set(height);
+    /// <https://html.spec.whatwg.org/multipage/#concept-video-intrinsic-width>
+    fn set_intrinsic_size(&self) {
+        let named_width = self.get_length(&local_name!("width"));
+        let named_height = self.get_length(&local_name!("height"));
```
Loirooriol, line 142: Probably better to have a `parse_plain_attribute()` with `AttrValue::from_dimension` and then to get just `.get_attr_for_layout(&ns!(), &local_name!("width")).map(AttrValue::as_dimension)`. See `components/script/dom/htmlimageelement.rs`

components/script/dom/htmlvideoelement.rs
```diff
@@ -100,20 +102,54 @@ impl HTMLVideoElement {
...
+        };
+
+        // Default height is half the width
+        let height = if let Some(x) = named_height {
+            x
+        } else if let Some(x) = named_width {
+            x / 2
+        } else {
+            DEFAULT_HEIGHT
+        };
```
Loirooriol, line 160: I tend to think this logic should actually be handled by https://github.com/servo/servo/blob/e857cdf022f319ae3fbd54e10e279ab43485995b/components/layout_2020/replaced.rs#L416

components/script/dom/htmlvideoelement.rs
```diff
@@ -416,3 +473,27 @@ impl PosterFrameFetchContext {
...
+                height: frame.height,
+            };
+        }
+
+        HTMLMediaData {
+            current_frame: None,
+            width: video.video_width.get() as i32,
+            height: video.video_height.get() as i32,
+        }
+    }
```
Loirooriol, line 508: This is mixing the natural sizes of the video with the sizes from the attributes (which should just set CSS presentational hints). Then I think this will break `object-fit`.

Loirooriol (changes requested):

components/script/dom/htmlvideoelement.rs
eerii: Ok, that makes a lot of sense, now I see it. I'll make some changes

eerii:
@Loirooriol had some great insights and it was a deeper rabbit hole than I thought, but after reading the spec and some issues I think this should be in a better state than before.

The most confusing thing was that the concept of an intrinsic ratio [was removed](https://github.com/whatwg/html/pull/6032) not too long ago, but video elements seem to have a special case. If both width and height attribtues are specified, https://html.spec.whatwg.org/#map-to-the-aspect-ratio-property comes into play and the video gets a presentational hint aspect ratio. If none of them are specified, the default size is 300x150, set here:

https://github.com/servo/servo/blob/e857cdf022f319ae3fbd54e10e279ab43485995b/components/layout_2020/replaced.rs#L416

The issue comes when one of the properties is not set. We can't set a presentational hint since we don't have both, and if we used the default video size instead, then the ratio would be wrong and it would fail `intrinsic_sizes.htm`. The discussion in https://github.com/WebKit/WebKit/pull/4509 was really insightful and it highlighted some of the problems.

To solve this, I added CSS hints for the aspect ratio when both attributes are there. This incidentally fixed some more tests to do with image aspect ratios. Canvas can be added later, it is similar but it uses a [slightly different logic](https://html.spec.whatwg.org/#attributes-for-embedded-content-and-images). Additionally, I added a field to the `HTMLMediaData` that tracks if the video has at least one of the two attributes unset. If this is the case, and there is no natural sizing (from the video metadata for example), in `replaced.rs` we overwrite the ratio with the default video size.

Additionally, I reworked a bit the logic for when posters or sources are added and removed, it should now cover more edge cases that I missed before, and video now also emits the width and height presentational attributes, unlike before that they were mixed with the video natural size.

I'm not quite sure why the WPT Export test is changing, maybe I have to rebase the changes? Aside from that, this might be a good point to rerun the tests. I tried running all of the ones that failed before locally and they seem to work.

Loirooriol:
Thanks, I will take a look tomorrow.
#33749 should have fixed the WPT Export job.

eerii:
Great to see more tests passing :)
I was able to reproduce the issue with `default object size after src is removed` when building servo in release mode. From what I can see, there can be a data race between calling load and requestAnimationFrame.

```js
v.removeAttribute("src");
v.load();
requestAnimationFrame(...);
```

```
 0:02.52 pid:300218 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.52 pid:300218 rendering frame (render)
 0:02.52 pid:300218 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:04.22 INFO Pausing until the browser exits
 0:04.22 TEST_END: Test OK. Subtests passed 5/5. Unexpected 0
```

```
 0:02.55 pid:300525 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.55 pid:300525 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:02.55 pid:300525 rendering (render)
 0:05.17 INFO Pausing until the browser exits
 0:05.18 TEST_END: Test OK. Subtests passed 4/5. Unexpected 1
```


Also, since 

I moved the clearing of the current frame directly to the attribute removal,

eerii:
Great to see more tests passing :)

I was able to reproduce the issue with `default object size after src is removed` when building servo in release mode. From what I can see, there can be a data race to the video renderer between calling `load` and `requestAnimationFrame`.

```js
v.removeAttribute("src");
v.load();
requestAnimationFrame(...);
```

```
 0:02.52 pid:300218 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.52 pid:300218 rendering frame (render)
 0:02.52 pid:300218 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:04.22 INFO Pausing until the browser exits
 0:04.22 TEST_END: Test OK. Subtests passed 5/5. Unexpected 0
```

```
 0:02.55 pid:300525 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.55 pid:300525 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:02.55 pid:300525 rendering (render)
 0:05.17 INFO Pausing until the browser exits
 0:05.18 TEST_END: Test OK. Subtests passed 4/5. Unexpected 1
```

In one of the latest changes I moved where the frame was being cleared, and the logic I added for that was incorrect. Since locally the events happened in order for me (my guess is that because the debug build is slower?) I didn't detect it before, and it passed. I moved the clearing of the current frame directly to the attribute removal, which should make it good now.

eerii:
Great to see more tests passing :)

I was able to reproduce the issue with `default object size after src is removed` when building servo in release mode. From what I can see, there can be a data race to the video renderer between calling `load` and `requestAnimationFrame`.

```js
v.removeAttribute("src");
v.load();
requestAnimationFrame(...);
```

```
 0:02.52 pid:300218 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.52 pid:300218 rendering frame (render)
 0:02.52 pid:300218 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:04.22 INFO Pausing until the browser exits
 0:04.22 TEST_END: Test OK. Subtests passed 5/5. Unexpected 0
```

```
 0:02.55 pid:300525 src is removed (attribute_mutated) <- should be cleared here, but it isn't
 0:02.55 pid:300525 calling load, set poster (media_element_load_algorithm) <- cleared here
 0:02.55 pid:300525 rendering (render)
 0:05.17 INFO Pausing until the browser exits
 0:05.18 TEST_END: Test OK. Subtests passed 4/5. Unexpected 1
```

In one of the latest changes I moved where the frame was being cleared, and the logic I added for that was incorrect. Since locally the events happened in order for me (my guess is that because the debug build is slower?) I didn't detect it before, and it passed. I moved the clearing of the current frame directly to the attribute removal, which should make it good now.

tests/wpt/meta/css/css-sizing/intrinsic-size-fallback-replaced.html.ini
```diff
@@ -1,6 +1,6 @@
 [intrinsic-size-fallback-replaced.html]
   [.test 2]
-    expected: FAIL
+    expected: PASS
```
Loirooriol, line 3: You should just be able to delete the `[.test 2]` and `[.test 6]` sections.

components/script/dom/htmlvideoelement.rs
```diff
@@ -416,3 +447,56 @@ impl PosterFrameFetchContext {
...
+        // metadata of the video is loaded).
+        let metadata = video
+            .get_video_width()
+            .zip(video.get_video_height())
+            .map(|(width, height)| MediaMetadata { width, height });
+
+        HTMLMediaData {
+            current_frame,
+            metadata,
+            has_default_size: self.width_attr().is_none() || self.height_attr().is_none(),
```
Loirooriol, line 491: I don't think this should depend on the attributes, e.g.

```html
<video width="100" height="100" style="border: solid; height: auto; aspect-ratio: auto"></video>
```

should still use the default aspect ratio of 2.

So  would just remove  `has_default_size` and handle it during layout.

Loirooriol (commented):
Sorry for the delay

components/script/dom/htmlvideoelement.rs
eerii: I think the video element has a special case for this, which is what the `video-aspect-ratio.html` test is checking. Changing this to ignore the attributes makes the test fail:

```
  ▶ TIMEOUT [expected OK] /html/rendering/replaced-elements/attributes-for-embedded-content-and-images/video-aspect-ratio.html
  │   ▶ FAIL [expected PASS] Video width and height attributes are used to infer aspect-ratio
  │   │   → assert_approx_equals: expected 2.5 +/- 0.001 but got 2
  │   │ 
  │   │ assert_ratio@http://web-platform.test:8000/html/rendering/replaced-elements/attributes-for-embedded-content-and-images/video-aspect-ratio.html:19:23
  │   │ @http://web-platform.test:8000/html/rendering/replaced-elements/attributes-for-embedded-content-and-images/video-aspect-ratio.html:30:17
  │   │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2622:25
  │   │ promise_test/tests.promise_tests</<@http://web-platform.test:8000/resources/testharness.js:738:36
  └   └ promise_test/tests.promise_tests<@http://web-platform.test:8000/resources/testharness.js:737:20
```

eerii:
So sorry about the delay, I got caught up in another work. I applied most of the suggestions (except that one) and ran all the changed tests again to check that everything still worked.

components/script/dom/htmlvideoelement.rs
Loirooriol: OK, so that happens because `width="100" height="100"` set `aspect-ratio: auto 100 / 100`, and then `auto` prefers the natural aspect ratio if available, which as per https://github.com/w3c/csswg-drafts/issues/7524#issuecomment-1204462924 it should never be from the default object size.

I added a commit to try to address this, let's see if it works

Loirooriol (approved):

Loirooriol:
It would be good to have `<video width="100" height="100" style="border: solid; height: auto; aspect-ratio: auto"></video>` in WPT, but the export is failing so it can be done in another PR.

mrobinson:
I think we could fix the WPT export job here by doing a rebase. I'll trigger that.

mrobinson:
Oh, looks like there are conflicts. In general our WPT export doesn't work well with merge commits.

eerii:
Sorry about the merge commit! :( Do you want me to properly rebase it?

mrobinson:
> Sorry about the merge commit! :( Do you want me to properly rebase it?

No worries. The WPT export should likely be improved somehow to support this usecase. If you want to add a test to this PR you will need to rebase, but otherwise I think you can land as is according to @Loirooriol's comment.

Loirooriol:
It's strange that it's complaining because this isn't modifying any test.
So yeah let's just try landing.

PR #31751, opened 2024-03-19
servoshell: draft to show a status bar tooltip

### LINKED ISSUE #31690: servoshell: show a status bar tooltip when hovering over links
When the minibrowser is enabled, add an [egui tooltip](https://docs.rs/egui/0.22.0/egui/containers/popup/fn.show_tooltip_at.html) to the bottom left corner when the page has status text.

Watch for `EmbedderMsg::Status` events, and keep track of the current status value (including `None`) in `Minibrowser`. To get the y coordinate of the bottom edge, use `ui.cursor().max.y` in a block that has `ui`, such as [the TopBottomPanel show routine](https://github.com/servo/servo/blob/68b82e6d6133c8213cdad09b240d7a69227cc82a/ports/servoshell/minibrowser.rs#L146-L190).

image

sandeepB3:
As far as I understand, three files come into play here: `webview.rs`, `app.rs`, and `minibrowser.rs`.

In `webview.rs`, the `handle_servo_events` function manages the Embedder message status and returns a `ServoEventResponse` struct updated with the status to the caller (app.rs).

The `app.rs` is the main file which manages the event loop and the windowing system of the minibrowser. Here, we add a new `status` variant to the `PumpResult` enum. Once `app.rs` calls the `handle_servo_events` function, it receives the new `ServoEventResponse` containing the status. The `status` variant can then be updated in the respective places where the `PumpResult` enum is being used.

Finally, if the `status` matches the Some variant, `app.rs` can pass this status to the `update` function inside `webview.rs` (now updated with an additional parameter `status: Option<String>`) to show the egui tooltip if there is some status.

---

- [ ] These changes fix #31690 
- [ ] There are tests for these changes: to show a status bar tooltip when hovering over links


delan:
Thanks for working on this! Sorry for the delay in reviewing it, I saw this was a draft so I assumed it wasn’t ready.

Unfortunately the status bar tooltip won’t really be visible with the current approach, because the Minibrowser will run a single egui update with the new tooltip, then immediately go back to having no tooltip in subsequent updates.

Plumbing the status through PumpResult::Continue and Minibrowser::update like this also complicates checking whether we need to run an egui update, so we end up with bugs like running two updates when `history_changed` and `status` is Some, or not running any updates when `status` goes from Some to None.

Since #32011 is more complete, and takes advantage of the new update logic introduced in #31713, I’m going to close this for now, but thanks again for your effort.

PR #31753, opened 2024-03-19
layout: Add very basic support for showing text in input boxes

mrobinson:
This is lacking many things. Notably:

1. There is no painting of selection or caret.
2. The size of the input box is wrong if it contains no text.

---


mrobinson:
I'm going to close this for now because it needs more work. I'll reopen when it's ready to go.

PR #31757, opened 2024-03-19
Update mozjs to use SM 115.9

sagudev:
Companion for https://github.com/servo/mozjs/pull/459

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,7 +329,7 @@ dependencies = [
  "bitflags 2.5.0",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.12.1",
  "lazy_static",
  "lazycell",
  "log",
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

wusyong:
There's a more clear [versioning](https://github.com/servo/mozjs/pull/456) in mozjs. While we haven't publish to crates.io yet, could it add `rev` of mozjs dependency in script's Cargo.toml?

sagudev:
> There's a more clear [versioning](https://github.com/servo/mozjs/pull/456) in mozjs. While we haven't publish to crates.io yet, could it add `rev` of mozjs dependency in script's Cargo.toml?

We could, but releases are currently versioned only for mozjs_sys, so if we get some new stuff only in mozjs we would need to go back to what we do currently (we just used `git` in Cargo.toml and then locked it to a commit using Cargo.lock).

Issue #30999, opened 2024-01-03
split varints cts.html for every subtest


sagudev:
I though I catch all `[FAIL/PASS, TIMEOUT, NOTRUN]`, but apparently not. To prevent them from happening it would be bast to have variants for every sub tests.

See: https://github.com/gpuweb/cts/issues/3233

sagudev:
I am currently waiting for https://github.com/gpuweb/cts/pull/3246 to land.

sagudev:
Currently this was resolved (somehow) in #31695, by just disabling such flaky test. But now in #31772 I observed some real timeouts that would be solved by more fine chunking.

Solution as noted in https://github.com/servo/servo/pull/31772#issuecomment-2011907601, is to chunk only problematic test per max time.

sagudev:
Times from Chrome are completely different (servo/FF: 18s vs chrome 20ms), so we would need to generate our own time data for it to work.

Moving to timeout long produced this results (from current main): https://github.com/sagudev/servo/actions/runs/8456974386/job/23169503338, so I've been thinking if we move all test that need more time in separate html file so meta long will effect only them.

PR #31772, opened 2024-03-20
webgpu: Implement onSubmittedWorkDone

sagudev:
All relevant test `webgpu:api,operation,onSubmittedWorkDone:` passes. There is also some more passes in shader tests. It serves as good example how to do safe callback (they key part is to use `move` closure).

---

### DIFF
--- a/components/script/dom/gpuqueue.rs
+++ b/components/script/dom/gpuqueue.rs
@@ -208,7 +208,7 @@ impl GPUQueueMethods for GPUQueue {
                 queue_id: self.queue.0,
             },
         )) {
-            todo!("QueueOnSubmittedWorkDone failed with {e}")
+            warn!("QueueOnSubmittedWorkDone failed with {e}")
         }
         promise
     }
--- a/components/script/dom/gpuqueue.rs
+++ b/components/script/dom/gpuqueue.rs
@@ -7,7 +7,7 @@ use std::rc::Rc;
 use dom_struct::dom_struct;
 use ipc_channel::ipc::IpcSharedMemory;
 use webgpu::identity::WebGPUOpResult;
-use webgpu::{wgt, WebGPU, WebGPUQueue, WebGPURequest, WebGPUResponse, WebGPUResponseResult};
+use webgpu::{wgt, WebGPU, WebGPUQueue, WebGPURequest, WebGPUResponse};
 
 use super::bindings::codegen::Bindings::WebGPUBinding::{GPUImageCopyTexture, GPUImageDataLayout};
 use super::gpu::{response_async, AsyncWGPUListener};
@@ -196,7 +196,7 @@ impl GPUQueueMethods for GPUQueue {
         Ok(())
     }
 
-    /// https://gpuweb.github.io/gpuweb/#dom-gpuqueue-onsubmittedworkdone
+    /// <https://gpuweb.github.io/gpuweb/#dom-gpuqueue-onsubmittedworkdone>
     fn OnSubmittedWorkDone(&self) -> Rc<Promise> {
         let global = self.global();
         let promise = Promise::new(&global);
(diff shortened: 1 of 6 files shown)


components/script/dom/gpuqueue.rs
```diff
@@ -192,4 +196,40 @@ impl GPUQueueMethods for GPUQueue {
 
         Ok(())
     }
+
+    /// https://gpuweb.github.io/gpuweb/#dom-gpuqueue-onsubmittedworkdone
```
mrobinson, line 200: To avoid a rustdoc warning:

```suggestion
    /// <https://gpuweb.github.io/gpuweb/#dom-gpuqueue-onsubmittedworkdone>
```

components/script/dom/gpuqueue.rs
```diff
@@ -192,4 +196,40 @@ impl GPUQueueMethods for GPUQueue {
...
+            },
+        )) {
+            todo!("QueueOnSubmittedWorkDone failed")
+        }
+        promise
+    }
+}
+
+impl AsyncWGPUListener for GPUQueue {
+    #[allow(unsafe_code)]
```
Taym95, line 219:
```suggestion
    
```
I don't think `#[allow(unsafe_code)]` is needed

sagudev:
I ran same test in firefox `webgpu:api,validation,image_copy,texture_related:origin_alignment:*` and it also timeout at ~96. subtest (same as servo), so we cannot really solve it (problem is inherited by wgpu), probably same problem for timeouts on: https://github.com/gpuweb/cts/blob/278383fad33d67229a7a592ce39762ae9f0addf5/src/webgpu/api/operation/onSubmittedWorkDone.spec.ts#L27

One thing that bothers me is that `--timeout-multiplier` does not help (it always hits timeout in 10s), so I guess either something is wrong with CTS timeouts, or we do not pass multiplier correctly (unlikely).

Like I was thinking in https://github.com/servo/servo/issues/30999, some tests needs to be splitted, but instead of splitting them all, we can just spit TIMEOUT tests into 2s chunks (time listings are already in CTS, although they are chromium times, so times for `webgpu:api,operation,onSubmittedWorkDone:many,serial:*` are completely wrong for us: https://github.com/gpuweb/cts/blob/278383fad33d67229a7a592ce39762ae9f0addf5/src/webgpu/listing_meta.json#L105).

sagudev:
> One thing that bothers me is that `--timeout-multiplier` does not help (it always hits timeout in 10s), so I guess either something is wrong with CTS timeouts, or we do not pass multiplier correctly (unlikely).

putting `<meta name="timeout" content="long">` on top of cts.html works (needs to be before all script tags!), so I guess timeout-multiplier does not work on harness_timeout? Although it should according to https://web-platform-tests.org/writing-tests/testharness-api.html#harness-timeout:

> Harness Timeout
>
> Execution of tests on a page is subject to a global timeout. By default this is 10s, but a test runner may set a timeout multiplier which alters the value according to the requirements of the test environment

sagudev:
All relevant test `webgpu:api,operation,onSubmittedWorkDone:` passes. There is also some more passes in shared tests. 

Try run: https://github.com/sagudev/servo/actions/runs/8512955377/job/23317378749 is as always full of flaky `webgpu:shader,execution,expression` crashes (https://github.com/servo/servo/issues/31397).

components/script/dom/gpuqueue.rs
```diff
@@ -191,4 +195,39 @@ impl GPUQueueMethods for GPUQueue {
...
+        let promise = Promise::new(&global);
+        let sender = response_async(&promise, self);
+        if let Err(e) = self.channel.0.send((
+            self.device.borrow().as_ref().unwrap().use_current_scope(),
+            WebGPURequest::QueueOnSubmittedWorkDone {
+                sender,
+                queue_id: self.queue.0,
+            },
+        )) {
+            todo!("QueueOnSubmittedWorkDone failed with {e}")
```
mrobinson, line 211: Hrm. `todo!` is an interesting choice here. Why not `warn!` which won't panic?

```suggestion
            warn!("QueueOnSubmittedWorkDone failed with {e}")
```

mrobinson (changes requested):

components/script/dom/gpuqueue.rs
sagudev: Todo as actually to-do in the future. But now that I look over at other similar code warn would really be more appropriate.

gterzian (approved):
LGTM.

mrobinson (approved):

PR #31774, opened 2024-03-20
servoshell: fix gap between minibrowser toolbar and webview

### LINKED ISSUE #31577: The minibrowser has a few pixels of unrendered space underneath the URL bar
On my system (Ubuntu), the minibrowser has a few pixels of unrendered space underneath the URL bar. This often flashes likely due to texture reuse. Unfortunately, I'm not able to do a screen recording of this behavior.

cc @delan @atbrakhi

delan:
For reasons that are unclear, checking the TopBottomPanel’s ui cursor yields a height that is one egui point too tall, leaving a gap between the minibrowser’s toolbar and the webview’s viewport.

This patch fixes that by checking the Context instead.

---

- [x] These changes fix #31577

- [x] These changes do not require tests because it’s not currently feasible to test the minibrowser frontend

### DIFF
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -142,53 +142,53 @@ impl Minibrowser {
         } = self;
         let widget_fbo = *widget_surface_fbo;
         let _duration = context.run(window, |ctx| {
-            let InnerResponse { inner: height, .. } =
-                TopBottomPanel::top("toolbar").show(ctx, |ui| {
-                    ui.allocate_ui_with_layout(
-                        ui.available_size(),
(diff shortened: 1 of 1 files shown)


mrobinson (approved):
Thank you for this fix!!

eerii:
Awesome fix, thank you so much! I think it might have introduced an unused import warning:
```
warning: unused import: `InnerResponse`
  --> ports/servoshell/minibrowser.rs:10:33
   |
10 | use egui::{CentralPanel, Frame, InnerResponse, Key, Modifiers, PaintCallback, TopBottomPanel};
   |                                 ^^^^^^^^^^^^^
   |
```

PR #31784, opened 2024-03-20
Fix some clippy warnings in `components/script/webdriver_handlers.rs`

komutumu:
Fixed some clippy warnings in components/script/webdriver_handlers.rs

---
- [X] These changes do not require tests because they do not touch functionality, they only fix lint errors.

### DIFF
--- a/components/script/webdriver_handlers.rs
+++ b/components/script/webdriver_handlers.rs
@@ -85,11 +85,11 @@ fn find_node_by_unique_id(
     }
 }
 
-fn matching_links<'a>(
-    links: &'a NodeList,
+fn matching_links(
+    links: &NodeList,
     link_text: String,
     partial: bool,
-) -> impl Iterator<Item = String> + 'a {
+) -> impl Iterator<Item = String> + '_{
     links
         .iter()
         .filter(move |node| {
@@ -251,7 +251,7 @@ pub unsafe fn jsval_to_webdriver(
                 cx,
                 object.handle(),
                 name.as_ptr(),
-                &mut HandleValueArray::new(),
+                &HandleValueArray::new(),
                 value.handle_mut(),
             ) {
                 jsval_to_webdriver(cx, global_scope, value.handle())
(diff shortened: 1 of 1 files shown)


jahielkomu:
Hello @mrobinson . I hope you are well.
Please kindly review.
Thank you

components/script/webdriver_handlers.rs
```diff
@@ -382,10 +382,7 @@ fn get_element_in_view_center_point(element: &Element) -> Option<Point2D<i64>> {
         .map(DomRoot::upcast::<Element>)
         .and_then(|body| {
             element
-                .GetClientRects()
-                .iter()
-                // Step 1
```
sagudev, line 387: This comment refers to first step in spec:
> Let rectangle be the first [element](https://dom.spec.whatwg.org/#concept-element) of the [DOMRect](https://www.w3.org/TR/geometry-1/#domrect) sequence returned by calling [getClientRects](https://www.w3.org/TR/cssom-view-1/#dom-element-getclientrects)() on [element](https://dom.spec.whatwg.org/#concept-element).

so it should be kept.

jahielkomu: Thank you @sagudev .
Your response helps a lot.

I'm working on it now.

jahielkomu:
Hello @mrobinson and @sagudev 
Please kindly review.
Thanks.

components/script/webdriver_handlers.rs
```diff
@@ -381,12 +381,9 @@ fn get_element_in_view_center_point(element: &Element) -> Option<Point2D<i64>> {
...
         .and_then(|body| {
-            element
-                .GetClientRects()
-                .iter()
-                // Step 1
-                .next()
-                .map(|rectangle| {
+           let rectangle_element = element.GetClientRects();
+           let rectangle_option = rectangle_element.iter().next();
+           rectangle_option.map(|rectangle| {
```
mrobinson, line 386: I believe that @sagudev meant that you should keep the comment. I recommend something like you had before, just restoring the original comment. We could even improve it a bit:

```
        // Step 1: Let rectangle be the first element of the DOMRect sequence 
        // returned by calling getClientRects() on element.
        element
            .GetClientRects().first()
            .map(|rectangle| {
```

mrobinson (changes requested):

jahielkomu:
@mrobinson in the comment; 
// Step 1: Let rectangle be the first element of the DOMRect sequence returned by calling getClientRects() on element, should ```element```, ```DOMRect``` and ```getClientRects()``` be links as earlier portrayed by @sagudev ?

mrobinson:
> I have updated the PR @mrobinson, please review.
> 
> @mrobinson in the comment; // Step 1: Let rectangle be the first element of the DOMRect sequence returned by calling getClientRects() on element, should `element`, `DOMRect` and `getClientRects()` be links as earlier portrayed by @sagudev ?

This is not necessary. Normal comments aren't processed into rustdoc, so it isn't always useful to make links. In addition, the link exist in the specification that is linked to. I think these links are in @sagudev's comment because they were preserved while copying and pasting from the specification.

mrobinson (approved):

sagudev (approved):

PR #31791, opened 2024-03-20
clippy: fix `result_unit_err` warnings

eerii:
This pr tackles the clippy warnings about having a function returning a `Result<_, ()>`. There are three main cases:

- The function returns a result directly from an external library (`components/url/lib.rs`). An allow is added for this case because we have no control over that.
- The `Result` type is better expressed as an `Option` (it is used to check if something is there or not, and there is no differentiation between error scenarios). The spec also says for some functions that "This will return either a TYPE, or nothing.", so using an `Option` may be more in line. I understand that for some of the cases this may be controversial, so please add a suggestion if you think a specific return should stay a `Result`.
- In cases where it made sense, added an error type to the result so it is more clear why is it failing. This was done in `components/net/cookie_storage.rs` and `components/shared/canvas/webgl_channel/mod.rs`.

---
- [x] These changes do not require tests because they only fix warnings, they don't add new functionality

### DIFF
--- a/components/canvas/canvas_paint_thread.rs
+++ b/components/canvas/canvas_paint_thread.rs
@@ -30,7 +30,7 @@ pub enum ImageUpdate {
 }
 
 pub trait WebrenderApi {
-    fn generate_key(&self) -> Result<ImageKey, ()>;
+    fn generate_key(&self) -> Option<ImageKey>;
     fn update_images(&self, updates: Vec<ImageUpdate>);
     fn clone(&self) -> Box<dyn WebrenderApi>;
 }
--- a/components/canvas/canvas_paint_thread.rs
+++ b/components/canvas/canvas_paint_thread.rs
@@ -30,6 +30,7 @@ pub enum ImageUpdate {
 }
 
 pub trait WebrenderApi {
+    /// Attempt to generate an [`ImageKey`], returning `None` in case of failure.
     fn generate_key(&self) -> Option<ImageKey>;
     fn update_images(&self, updates: Vec<ImageUpdate>);
     fn clone(&self) -> Box<dyn WebrenderApi>;
--- a/components/net/fetch/methods.rs
+++ b/components/net/fetch/methods.rs
@@ -569,17 +569,17 @@ pub enum RangeRequestBounds {
 }
 
 impl RangeRequestBounds {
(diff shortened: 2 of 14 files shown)


components/canvas/canvas_paint_thread.rs
```diff
@@ -30,7 +30,7 @@ pub enum ImageUpdate {
 }
 
 pub trait WebrenderApi {
-    fn generate_key(&self) -> Result<ImageKey, ()>;
+    fn generate_key(&self) -> Option<ImageKey>;
```
mrobinson, line 33:
```suggestion
    /// Attempt to generate an [`ImageKey`], returning `None` in case of failure.
    fn generate_key(&self) -> Option<ImageKey>;
```

components/net/fetch/methods.rs
```diff
@@ -569,17 +569,17 @@ pub enum RangeRequestBounds {
 }
 
 impl RangeRequestBounds {
-    pub fn get_final(&self, len: Option<u64>) -> Result<RelativePos, ()> {
+    pub fn get_final(&self, len: Option<u64>) -> Option<RelativePos> {
```
mrobinson, line 572: An error might help with diagnosing problems down the line. Could this just return an `Err` that's a `&'static str` explaining where in the process this failed? Even if we eventually throw those errors away they can help in debugging or be surfaced later.

components/net/fetch/methods.rs
```diff
@@ -569,17 +569,17 @@ pub enum RangeRequestBounds {
...
         match self {
             RangeRequestBounds::Final(pos) => {
                 if let Some(len) = len {
                     if pos.start <= len as i64 {
-                        return Ok(pos.clone());
+                        return Some(pos.clone());
                     }
                 }
-                Err(())
+                None
```
mrobinson, line 580: For instance this could be:

```suggestion
                Err("Tried to process RangeRequestBounds::Final without len")
```

or something along those lines.

mrobinson (changes requested):
Looks good, though maybe we can preserve context for errors in some cases:

components/net/fetch/methods.rs
eerii: That makes a lot of sense, this way the debug information is there so it can be used if needed. I changed it back and added the error description on the last commit, thanks!

mrobinson (approved):

PR #31802, opened 2024-03-21
Fix WPT reference no_red_3x3_monospace_table-ref.xht

### LINKED ISSUE #31752: Fix WPT reference no_red_3x3_monospace_table-ref.xht
Tests that are matched against this reference use `padding: 1px`, e.g. https://github.com/servo/servo/blob/06a021db552c1d2f8899484f7c1143424c387eef/tests/wpt/tests/css/CSS2/tables/table-anonymous-objects-060.xht#L13

However, the reference uses `position: absolute; top: 1px; left: 1px` https://github.com/servo/servo/blob/06a021db552c1d2f8899484f7c1143424c387eef/tests/wpt/tests/css/CSS2/tables/reference/no_red_3x3_monospace_table-ref.xht#L17-L19

This is a problem: the former shrinks the containing block by 2px (1px on the left and 1px on the right), while the latter only by 1px on the left.

This is not noticeable on other browsers because the text is small enough to fit without wrapping, but Servo defaults to a bigger `font-size` so the text needs wrapping. `./mach run -r tests/wpt/tests/css/CSS2/tables/reference/no_red_3x3_monospace_table-ref.xht  --resolution 800x600`

I guess we should make `font-family: monospace` default to `font-size: 12px` instead of `16px`, but it seems wrong to have this assumption in these tests, since they are about tables, not fonts. Also, `font-size: 12px` doesn't really guarantee that the text will not wrap, since browsers may use different monospaced fonts with varying behaviors.

So I think the steps here are:
1. Modify `no_red_3x3_monospace_table-ref.xht` by adding `right: 1px`
2. [Run all the tests](https://github.com/servo/servo/blob/main/tests/wpt/README.md#running-the-tests) in `tests/wpt/tests/css/CSS2/tables/`, [updating the expectations](https://github.com/servo/servo/blob/main/tests/wpt/README.md#updating-test-expectations)
3. Create a PR fixing the problem. It will be automatically exported to WPT, where the change will be tested on Gecko, Blink and WebKit. You will need to confirm that they won't be negatively affected.

richarddushime:
Following the Guidelines i made the necessary changes and checked all tests As requested in the issue #31752 

---

- [x] These changes fix #31752  (GitHub issue number)

### DIFF
--- a/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-059.xht.ini
+++ b/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-059.xht.ini
@@ -1,2 +1,0 @@
-[table-anonymous-objects-059.xht]
-  expected: FAIL
--- a/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-060.xht.ini
+++ b/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-060.xht.ini
@@ -1,2 +1,0 @@
-[table-anonymous-objects-060.xht]
-  expected: FAIL
--- a/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-061.xht.ini
+++ b/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-061.xht.ini
@@ -1,2 +1,0 @@
-[table-anonymous-objects-061.xht]
-  expected: FAIL
--- a/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-062.xht.ini
+++ b/tests/wpt/meta/css/CSS2/tables/table-anonymous-objects-062.xht.ini
@@ -1,2 +1,0 @@
-[table-anonymous-objects-062.xht]
-  expected: FAIL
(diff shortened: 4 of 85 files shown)


ports/servoshell/minibrowser.rs
```diff
@@ -7,7 +7,7 @@ use std::num::NonZeroU32;
 use std::sync::Arc;
 use std::time::Instant;
 
-use egui::{CentralPanel, Frame, InnerResponse, Key, Modifiers, PaintCallback, TopBottomPanel};
+use egui::{CentralPanel, Frame, Key, Modifiers, PaintCallback, TopBottomPanel};
```
mrobinson, line 10: Please remove this unrelated change.

mrobinson (changes requested):
It's amazing how many tests this fixes.

ports/servoshell/minibrowser.rs
richarddushime: on my end : current branch its not there

richarddushime: its only available on the main branch

mrobinson: You need to rebase your change onto the main branch:

`git checkout main`
`git pull`
`git checkout wptreference`
`git rebase -i main`

Do the interactive rebase and then make sure that when you do `git diff main wptreference` the change is not there.

mrobinson: Or I can simply rebase from the PR, I guess. I'll do that now.

richarddushime: Great 
Thanks

Loirooriol (approved):

mrobinson (approved):

PR #31803, opened 2024-03-21
Don't trim leading whitespace of anonymous table cells

Loirooriol:
A sequence of whitespace shouldn't generate an anonymous table row/cell, but we can't just throw away the leading whitespace, because afterwards we may encounter some other content, and then the leading whitespace should appear in the cell (noticeable with e.g. `white-space: pre`).

---

### DIFF
--- a/components/layout_2020/table/construct.rs
+++ b/components/layout_2020/table/construct.rs
@@ -56,6 +56,19 @@ pub(crate) enum AnonymousTableContent<'dom, Node> {
     },
 }
 
+impl<'dom, Node> AnonymousTableContent<'dom, Node> {
+    fn is_whitespace_only(&self) -> bool {
+        match self {
+            Self::Element { .. } => false,
+            Self::Text(_, ref text) => text.chars().all(char_is_whitespace),
+        }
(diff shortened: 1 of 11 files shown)


components/layout_2020/table/construct.rs
```diff
@@ -56,6 +56,19 @@ pub(crate) enum AnonymousTableContent<'dom, Node> {
...
 
+impl<'dom, Node> AnonymousTableContent<'dom, Node> {
+    fn is_whitespace_only(&self) -> bool {
+        match self {
+            Self::Element { .. } => false,
+            Self::Text(_, ref text) => text.chars().all(char_is_whitespace),
+        }
+    }
+
+    fn are_whitespace_only(contents: &[Self]) -> bool {
```
mrobinson, line 67: Nit: `are` is a bit awkward here because it's not super clear from reading it that plurality matches first argument.

```suggestion
    fn contents_are_whitespace_only(contents: &[Self]) -> bool {
```

mrobinson (approved):
Nice fix!

Issue #29851, opened 2023-06-07
Using rust-analyzer causes massive rebuilds.


emilio:
This is because `rust-analyzer` uses `cargo check`, which uses different features from `./mach build`, causing massive rebuilds.

In gecko we have code to generate the following in `.vscode/settings.json`:

```json
{
    "rust-analyzer.server.extraEnv": {
        "CARGO_TARGET_DIR": "/home/emilio/src/moz/gecko/obj-debug"
    },
    "rust-analyzer.cargo.buildScripts.overrideCommand": [
        "./mach",
        "--log-no-times",
        "cargo",
        "check",
        "-j",
        "10",
        "--all-crates",
        "--message-format-json"
    ],
    "rust-analyzer.check.overrideCommand": [
        "./mach",
        "--log-no-times",
        "cargo",
        "check",
        "-j",
        "10",
        "--all-crates",
        "--message-format-json"
    ]
}
```

We should figure out a way of either:

 * Making `cargo check` just work.
 * Making some mach command that does it, much like the mach command above in Gecko, and add the relevant settings to the `.vscode` directory.

Ideally we do the first, but...

emilio:
Any good reason not to check in those settings.json? We already have a .vscode directory...

mrobinson:
I think the main issue is that if a user of vscode modifies the settings, they would constantly be listed as differing. I took a little time to see if there's a way to ship a template or default version of the settings with the project, but I couldn't find anything like this. Perhaps there is a way though...

sagudev:
IIRC gecko has mach command that generates settings.json.

PR #31805, opened 2024-03-21
crown: Do not warn about crown for rustdoc or clippy

### LINKED ISSUE #31804: Missing `crown` warning issued when running `./mach doc`
When running `./mach doc` the only warning we have now is this one:

```
warning: unknown lint: `crown_is_not_used`
  --> components/script/lib.rs:11:9
   |
11 | #![deny(crown_is_not_used)]
   |         ^^^^^^^^^^^^^^^^^
   |
note: the lint level is defined here
  --> components/script/lib.rs:10:9
   |
10 | #![warn(unknown_lints)]
   |         ^^^^^^^^^^^^^
```

`./mach doc` should be able to run without warnings so that we can prevent future ones.

mrobinson:
This is important because tools like clippy and rustdoc cannot use crown
and should not issue warnings.

Fixes #31804.

---

- [x] These changes fix #31804
- [x] These changes do not require tests because they just eliminate some build warnings.

### DIFF
--- a/components/script/lib.rs
+++ b/components/script/lib.rs
@@ -5,9 +5,11 @@
 #![feature(register_tool)]
 #![deny(unsafe_code)]
 #![doc = "The script crate contains all matters DOM."]
+// Register the linter `crown`, which is the Servo-specific linter for the script
+// crate. Issue a warning if `crown` is not being used to compile, but not when
+// building rustdoc or running clippy.
 #![register_tool(crown)]
-// Issue a warning if `crown` cannot be found.
-#![warn(unknown_lints)]
+#![cfg_attr(any(doc, clippy, feature = "cargo-clippy"), allow(unknown_lints))]
 #![deny(crown_is_not_used)]
 
 // These are used a lot so let's keep them for now


mrobinson:
> > @sagudev if there is a better way to do this, please let me know.
> 
> I think we can level it down to note or help: https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-levels, so then we can deny warnings but still allow info.

I think the issue with this is that if we use a [problem matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md), this will still spam every pull request with this warning. We need to not have any spurious output at all when running `rustdoc`, I believe.

sagudev:
CI uses crown, so this shouldn't be problem, right?

But looks like info is only diagnostic level, lint levels are different. But we could do detection using build.rs.

mrobinson:
> CI uses crown, so this shouldn't be problem, right?
> 
> But looks like info is only diagnostic level, lint levels are different. But we could do detection using build.rs.

This would be for when the CI runs `rustdoc` and starts reporting warnings as build failures or surfacing them with problem matchers.

sagudev:
> > CI uses crown, so this shouldn't be problem, right?
> > But looks like info is only diagnostic level, lint levels are different. But we could do detection using build.rs.
> 
> This would be for when the CI runs `rustdoc` and starts reporting warnings as build failures or surfacing them with problem matchers.

I think there is cfg attr for rustdoc. But if we start testing using clippy, this could be problem (maybe build.rs could detect clippy and not show warning).

sagudev:
Or even better we could just pass `--force-warn unknown_lints` in mach build.

mrobinson:
> Or even better we could just pass `--force-warn unknown_lints` in mach build.

Hrm. Doesn't this mean that `RUSTFLAGS` will be forever different when run via `./mach build` and `cargo build`. This will make it harder to solve #29851. We should try to get closer to building with `cargo build` if we can.

sagudev:
Both clippy and doc have their own cfg we can use: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=468b09ceddf6d874601dc5527ae088dc

sagudev (approved):

PR #31817, opened 2024-03-21
Replace `remutex` with `parking_lot`'s `ReentrantMutex`

### LINKED ISSUE #12641: Settle on a thread id implementation
I added a `ThreadId` type to `util`, it would be nice to have just one thread id implementation. cc @emilio

mrobinson:
Many things in Servo depend on `parking_lot`, so we can replace our
homegrown remutex with `parking_lot`'s version.

Fixes #12641.

---

- [x] These changes fix #12641.


ports/servoshell/lib.rs
```diff
@@ -61,11 +61,6 @@ pub fn main() {
     // Parse the command line options and store them globally
     let args: Vec<String> = env::args().collect();
     let mut opts = Options::new();
-    opts.optflag(
```
mukilan, line 64: Is this change supposed to be part of this PR?

mrobinson: No, this is from another change. I've gone ahead and removed it from this one.

mukilan (approved):

PR #31818, opened 2024-03-22
Clippy: Fixed some clippy warnings

Aaryakhandelwal:
Fixed some clippy warnings.

---

- [X] These changes fix #31500 (GitHub issue number if applicable)

- [X] These changes do not require tests because they only fix the clippy warnings.

### DIFF
--- a/components/script/dom/htmlelement.rs
+++ b/components/script/dom/htmlelement.rs
@@ -560,11 +560,11 @@ static DATA_PREFIX: &str = "data-";
 static DATA_HYPHEN_SEPARATOR: char = '\x2d';
 
 fn is_ascii_uppercase(c: char) -> bool {
-    ('A'..='Z').contains(&c)
+    c.is_ascii_uppercase()
 }
 
 fn is_ascii_lowercase(c: char) -> bool {
-    c.is_lowercase()
+    c.is_ascii_lowercase()
 }
 
 fn to_snake_case(name: DOMString) -> DOMString {
@@ -623,7 +623,7 @@ impl HTMLElement {
             .chars()
             .skip_while(|&ch| ch != '\u{2d}')
             .nth(1)
-            .map_or(false, |ch| ch.is_lowercase())
+            .map_or(false, |ch| ch.is_ascii_lowercase())
         {
             return Err(Error::Syntax);
         }
--- a/components/script/dom/htmlelement.rs
+++ b/components/script/dom/htmlelement.rs
@@ -559,19 +559,11 @@ fn append_text_node_to_fragment(document: &Document, fragment: &DocumentFragment
 static DATA_PREFIX: &str = "data-";
 static DATA_HYPHEN_SEPARATOR: char = '\x2d';
 
-fn is_ascii_uppercase(c: char) -> bool {
-    c.is_ascii_uppercase()
-}
-
-fn is_ascii_lowercase(c: char) -> bool {
-    c.is_ascii_lowercase()
-}
-
 fn to_snake_case(name: DOMString) -> DOMString {
     let mut attr_name = String::with_capacity(name.len() + DATA_PREFIX.len());
     attr_name.push_str(DATA_PREFIX);
     for ch in name.chars() {
-        if is_ascii_uppercase(ch) {
+        if ch.is_ascii_uppercase() {
             attr_name.push(DATA_HYPHEN_SEPARATOR);
             attr_name.push(ch.to_ascii_lowercase());
         } else {
@@ -591,7 +583,7 @@ fn to_camel_case(name: &str) -> Option<DOMString> {
         return None;
     }
     let name = &name[5..];
-    let has_uppercase = name.chars().any(|curr_char| is_ascii_uppercase(curr_char));
+    let has_uppercase = name.chars().any(|curr_char| curr_char.is_ascii_uppercase());
     if has_uppercase {
         return None;
     }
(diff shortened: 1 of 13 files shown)


components/script/dom/htmlelement.rs
```diff
@@ -560,11 +560,11 @@ static DATA_PREFIX: &str = "data-";
 static DATA_HYPHEN_SEPARATOR: char = '\x2d';
 
 fn is_ascii_uppercase(c: char) -> bool {
-    'A' <= c && c <= 'Z'
+    ('A'..='Z').contains(&c)
 }
 
 fn is_ascii_lowercase(c: char) -> bool {
-    'a' <= c && c <= 'w'
+    ('a'..='w').contains(&c)
```
mukilan, line 567: It seems weird that we use `a` to `w` as the range instead of `a` to `z`. Perhaps it was a bug introduced in [this commit](https://github.com/servo/servo/commit/4ee8f56b9cdb3c47e598eb27e94f07350fb58c5d)? It seems that commit was trying to remove the use of the standard library methods `char::is_ascii_{lower,upper}case`  because they were marked unstable at that time. But looks like those methods have been stabilized since then, so perhaps we can simply remove our definitions in this file and call the `std` methods as before?

components/script/dom/htmlelement.rs
```diff
@@ -623,7 +623,7 @@ impl HTMLElement {
             .chars()
             .skip_while(|&ch| ch != '\u{2d}')
             .nth(1)
-            .map_or(false, |ch| ch >= 'a' && ch <= 'z')
+            .map_or(false, |ch| ('a'..='z').contains(&ch))
```
mukilan, line 626: I think we can use `std::char::is_ascii_lowercase` here as well.
```
.map_or(false, |ch| ch.is_ascii_lowercase())
```

components/script/dom/htmlelement.rs
jdm: That is almost certainly an Azerty keyboard layout mistake.

Aaryakhandelwal: I have made the recommended changes.

components/script/dom/htmlelement.rs
```diff
@@ -560,11 +560,11 @@ static DATA_PREFIX: &str = "data-";
 static DATA_HYPHEN_SEPARATOR: char = '\x2d';
 
 fn is_ascii_uppercase(c: char) -> bool {
-    'A' <= c && c <= 'Z'
+    ('A'..='Z').contains(&c)
 }
 
 fn is_ascii_lowercase(c: char) -> bool {
-    'a' <= c && c <= 'w'
+    c.is_lowercase()
```
mukilan, line 567: [`is_lowercase`](https://doc.rust-lang.org/std/primitive.char.html#method.is_lowercase) has different semantics than [`is_ascii_lowercase`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_lowercase). The former will return true even for lowercase non-ASCII characters such as `'δ'` whereas the latter will only check for lowercase characters in the *ASCII*  range (i.e a to z).

Since we want to preserve the semantics, in my previous comment what I was suggesting was to delete the [`fn is_ascii_lowercase` defined in this file](https://github.com/servo/servo/blob/65e0a0c076d2ebd5efb9cf6af887ac167da9594b/components/script/dom/htmlelement.rs#L566) and instead use [`is_ascii_lowercase`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_lowercase) provided by the std library.

For example, in [this line](https://github.com/servo/servo/blob/65e0a0c076d2ebd5efb9cf6af887ac167da9594b/components/script/dom/htmlelement.rs#L604):
```
  if is_ascii_lowercase(next_char) {
      result.push(next_char.to_ascii_uppercase());
  } 
```
we are calling our own definition of `is_ascii_lowercase`. We can instead replace that with a call to the std library version like so:

```
  if next_char.is_ascii_lowercase() {
      result.push(next_char.to_ascii_uppercase());
  } 
```

Similarly, the [definition of `is_ascii_uppercase`](https://github.com/servo/servo/blob/65e0a0c076d2ebd5efb9cf6af887ac167da9594b/components/script/dom/htmlelement.rs#L562) can be deleted and the calls in this file can be replaced with the calls to the std library version.

Aaryakhandelwal: Ohh, I wasn't aware of the difference between them. I will change them.

components/script/dom/htmlelement.rs
```diff
@@ -560,11 +560,11 @@ static DATA_PREFIX: &str = "data-";
 static DATA_HYPHEN_SEPARATOR: char = '\x2d';
 
 fn is_ascii_uppercase(c: char) -> bool {
-    'A' <= c && c <= 'Z'
+    c.is_ascii_uppercase()
 }
 
 fn is_ascii_lowercase(c: char) -> bool {
```
mukilan, line 566: This function is now redundant, as it is not doing any actual work itself and simply forwards the call to the standard library method. So, we can just delete this function definition from this file and simply call `.is_ascii_lowercase()`  wherever the deleted function is used. Please refer to my previous comment for an example of what needs to be done at the call site. 

Similar refactoring needs to done for `is_ascii_uppercase` as well.

mukilan (changes requested):

mukilan (approved):
Thank you for the contribution, and for iterating on this patch to incorporate the bug fix and code refactoring!

components/script/dom/htmlfontelement.rs
```diff
@@ -174,8 +174,8 @@ fn parse_size(mut input: &str) -> AttrValue {
 
     // Step 9
     match parse_mode {
-        ParseMode::RelativePlus => value = 3 + value,
-        ParseMode::RelativeMinus => value = 3 - value,
+        ParseMode::RelativePlus => value += 3,
+        ParseMode::RelativeMinus => value -= 3,
```
mukilan, line 178: Since subtraction is not commutative, this change is not equivalent to the original code i.e `value -= 3` expands to `value = value - 3` whereas we want `value = 3 - value`

I think we can simply revert the change on line 178 alone.

Aaryakhandelwal: I have reverted the changes, it should be fine now.

PR #31823, opened 2024-03-22
clippy: Fix `unnecessary_cast` warnings in `components/script`

oluwatobiss:
Remove unnecessary casts to the same type to fix the `unnecessary_cast` warnings.

**ref:** https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast

---

- [X] These changes do not require tests because they only resolve clippy warnings. They do not change functionalities.


components/script/script_runtime.rs
```diff
@@ -703,7 +703,7 @@ pub fn get_reports(cx: *mut RawJSContext, path_seg: String) -> Vec<Report> {
                 reports.push(Report {
                     path,
                     kind,
-                    size: size as usize,
+                    size: size,
```
mrobinson, line 706:
```suggestion
                    size,
```

mrobinson (changes requested):

components/script/script_runtime.rs
oluwatobiss: Got it. Thanks.

mrobinson (approved):

PR #31825, opened 2024-03-22
nix build: replace `fetchCargoTarball` with `importCargoLock`

mukilan:
[`importCargoLock`][1] allows us to use the existing Cargo.lock file. This means we no longer need to update the sha256 hash whenever the dependencies are upgraded. It also integrates with nix's rustToolchain support via `cargoSetupHooks` and automatically vendors the dependencies, allowing us to simplify the logic for `filterlock` derivation.

[1]: https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#vendoring-of-dependencies-vendoring-of-dependencies

- [x] These changes do not require tests because the only modify files related to build support for Nix.

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -96,27 +96,17 @@ stdenv.mkDerivation (androidEnvironment // rec {
     # all of Servo’s dependencies get pulled into the Nix store too, wasting over 1GB of disk space.
     # Filtering the lockfile to only the parts needed by crown saves space and builds faster.
     (let
-      vendorTarball = rustPlatform.fetchCargoTarball {
-        src = ../support/filterlock;
-        hash = "sha256-EBrL0/cEJfGIXUYqFiufRYFBAk5LSf0Cd/19+9m9JZI=";
-      };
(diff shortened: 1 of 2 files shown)


sagudev:
This is also written on provided link:

> If a Cargo.lock file is available, you can alternatively use the importCargoLock function. In contrast to fetchCargoTarball, this function does not require a hash (unless git dependencies are used)

but this might not apply to crown?

mukilan:
@sagudev, I'm not sure I follow, are you suggesting that we use `importCargoLock` for crown as well? IIUC, `buildRustPackage` internally uses `importCargoLock` , passing it the `lockFileContents`. The reason we don't directly pass the lockfile is because we don't want all Servo's dependencies ending up in nix store.

Let me know if I misunderstood your comment.

sagudev:
I was just asking, but now I see that hash was only used for `filterlock`, while I thought it was used for crown. My bad.

delan (approved):
Thanks!

PR #31833, opened 2024-03-22
layout: Allow transforming inline replaced elements

### LINKED ISSUE #31806: Inline replaced elements (`<iframe>`, `<img>`) can’t be transformed
```html
<style>
    iframe, div {
        transform: rotate(15deg);
    }
</style>
<iframe src="iframe.html"></iframe>
<div><iframe src="iframe.html"></iframe></div>
```

image

mrobinson:
This requires passing through information about whether or not the
element in question is replaced when checking to see if it's
transformable and transitively all functions that make decisions about
containing blocks. A new FragmentFlag is added to help track this -- it
will be set on both the replaced items BoxFragment container as well as
the Fragment for the replaced item itself.

Fixes #31806.

---

- [x] These changes fix #31806.

### DIFF
--- a/components/layout_2020/display_list/stacking_context.rs
+++ b/components/layout_2020/display_list/stacking_context.rs
@@ -474,7 +474,7 @@ impl StackingContext {
         if effects.filter.0.is_empty() &&
             effects.opacity == 1.0 &&
             effects.mix_blend_mode == ComputedMixBlendMode::Normal &&
-            !style.has_transform_or_perspective()
+            !style.has_transform_or_perspective(FragmentFlags::empty())
         {
             return false;
         }
@@ -896,7 +896,7 @@ struct ReferenceFrameData {
 
 impl BoxFragment {
     fn get_stacking_context_type(&self) -> Option<StackingContextType> {
-        if self.style.establishes_stacking_context() {
+        if self.style.establishes_stacking_context(self.base.flags) {
             return Some(StackingContextType::RealStackingContext);
         }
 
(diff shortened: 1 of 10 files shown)


components/layout_2020/display_list/stacking_context.rs
```diff
@@ -896,7 +896,10 @@ struct ReferenceFrameData {
 
 impl BoxFragment {
     fn get_stacking_context_type(&self) -> Option<StackingContextType> {
-        if self.style.establishes_stacking_context() {
+        if self
+            .style
+            .establishes_stacking_context(self.base.flags.contains(FragmentFlags::IS_REPLACED))
```
Loirooriol, line 901: Could pass the `self.base.flags` and check `contains(FragmentFlags::IS_REPLACED)` there. Bool parameters aren't great anyways.

components/layout_2020/style_ext.rs
```diff
@@ -420,8 +420,7 @@ impl ComputedValuesExt for ComputedValues {
         //     elements with the exception of any descendant element of text content
         //     elements."
         // https://drafts.csswg.org/css-transforms/#transformable-element
-        // FIXME(mrobinson): Properly handle tables and replaced elements here.
-        if self.get_box().display.is_inline_flow() {
+        if self.get_box().display.outside() == stylo::DisplayOutside::Inline && !is_replaced {
```
Loirooriol, line 423: I don't think this should return false for `inline-table`, `inline-block` or `inline-flex`. You should keep `is_inline_flow()`.

mrobinson:
@Loirooriol Do you mind taking another look at this one? Thank you!

Loirooriol (approved):
Sorry, I thought I had already approved this.

PR #31832, opened 2024-03-22
clippy: fix warnings in /Components/url

ektuu:
---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #_clippy warnings in /components/url

- [x] These changes do not require tests because  they just fix clippy warnings.

PR #31834, opened 2024-03-22
 clippy: Map to an error type instead of using allowing `result_unit_err` in `components/url`

ektuu:
---

- [x] These changes are related to #31832 


Loirooriol (changes requested):
Some points:
 - When you receive feedback in a PR, just update the old one, don't create a new one
 - The only difference from cd28798d7e72b4ccab4808dd111050545bcdf339 is that you avoided inserting a space before `#![deny(unsafe_code)]`, but you are still adding unnecessary newlines
 - Is there any reason to reorder the methods?
 - The template for pull requests has a list of things that you need to test, and check if they hold.
 - In particular, `./mach test-tidy` is complaining: https://github.com/servo/servo/actions/runs/8395608656/job/22995237911?pr=31834
 - What are the warnings that you are trying to address?

ektuu:
Thankyou for your feedback. I will address the issues mentioned.

In the above PR, I used a custom error type to address the warnings of functions returning 'Result<_, ()>'. 
Methods such as set_username, set_ip_host, set_password, to_file_path, and from_file_path have been modified to return a Result type with UrlError instead of ().  Previously existing method definitions have been modified or removed to align with the new error handling strategy.

ektuu:
@Loirooriol 
I've addressed the issues. can you please review?

Loirooriol (commented):
What's the reason for reordering the methods?

ektuu:
Methods have been reordered to align with the new error handling strategy. The methods which use custom error UrlError have been put together to enhance readability of the code. Would you want me to reorder them ?

components/url/lib.rs
```diff
@@ -25,6 +24,29 @@ use url::{Position, Url};
...
+pub enum UrlError {
+    SetUsername,
+    SetIpHost,
+    SetPassword,
+    ToFilePath,
+    FromFilePath,
+}
+
+impl std::fmt::Display for UrlError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
eerii, line 37: A small nit, I think it would make sense if you used the already imported `fmt::` namespace for `Display`, `Formatter` and `Result`, as most of the files in the project already use this convention. If you decide to do it, it would also be nice to update the return type of `debug_compact` to match.

```suggestion
impl fmt::Display for UrlError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```

On a more global note, I wonder if it is necessary to add this error type. This is wrapping `url`'s library functions, and they decided not to include an error type for [these functions](https://docs.rs/url/latest/url/struct.Url.html#method.set_username). As I saw it the backtrace in case of an error would be enough to determine which function was causing the issue, that's why I added the `allow`, but a case could be made that the error provides better context.

components/url/lib.rs
```diff
@@ -25,6 +24,29 @@ use url::{Position, Url};
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            UrlError::SetUsername => write!(f, "Error setting username"),
+            UrlError::SetIpHost => write!(f, "Error setting IP host"),
+            UrlError::SetPassword => write!(f, "Error setting password"),
+            UrlError::ToFilePath => write!(f, "Error converting to file path"),
+            UrlError::FromFilePath => write!(f, "Error converting from file path"),
+        }
+    }
+}
```
mrobinson, line 46: I think we can just remove this entirely as the `Debug` implementation should be enough.

components/url/lib.rs
```diff
@@ -109,16 +131,32 @@ impl ServoUrl {
...
+    pub fn to_file_path(&self) -> Result<::std::path::PathBuf, UrlError> {
+        self.0.to_file_path().map_err(|_| UrlError::ToFilePath)
     }
 
-    pub fn set_password(&mut self, pass: Option<&str>) -> Result<(), ()> {
-        self.as_mut_url().set_password(pass)
+    pub fn from_file_path<P: AsRef<Path>>(path: P) -> Result<Self, UrlError> {
+        Url::from_file_path(path)
+            .map(Self::from_url)
+            .map_err(|_| UrlError::FromFilePath)
```
mrobinson, line 159: Let's keep the methods in the same order. I'm not sure it makes sense to organize them by their return value.

mrobinson (changes requested):

components/url/lib.rs
eerii: I think `Display` is there because `std::error::Error` [requires it](https://doc.rust-lang.org/std/error/trait.Error.html), but that can probably be removed as well. Other errors in servo such as `WebGLError` or `TimeRangesError` don't implement `Error` or `Display`, just `Debug`.

mrobinson: Makes sense. Let's remove both of these.

mrobinson (approved):

mrobinson:
@ektuu Once a change has been added to the merge queue making any change to the branch will remove it from the merge queue. By clicking the "Update branch" button, you took the change off the merge queue. In general it is *not necessary* to click this button. I recommend rebasing your work locally instead of clicking this button.

mrobinson:
Looks like I was wrong here. This didn't go to the merge queue, because it is still waiting on a review from @Loirooriol, as he requested changes earlier.

Loirooriol (approved):
I share eerii's doubts about whether this is actually an improvement, but I'm fine with the change.

mrobinson:
I also agree that as it is, this isn't super useful, but it does prepare the code for better error handling in the future.

PR #31840, opened 2024-03-23
Run platform bootstrap first in bootstrap

sagudev:
So the user do not need to install llvm via choco manually. This should solve `error: linker lld-link.exe not found` error on windows.

---

- [x] These changes do not require tests because they just adjust the bootstrap scripts.

### DIFF
--- a/python/servo/platform/base.py
+++ b/python/servo/platform/base.py
@@ -55,9 +55,9 @@ class Base:
             return False
 
     def bootstrap(self, force: bool):
-        installed_something = self.install_taplo(force)
+        installed_something = self._platform_bootstrap(force)
+        installed_something |= self.install_taplo(force)
         installed_something |= self.install_crown(force)
-        installed_something |= self._platform_bootstrap(force)
         if not installed_something:
             print("Dependencies were already installed!")
 
--- a/python/servo/platform/linux.py
+++ b/python/servo/platform/linux.py
@@ -146,9 +146,11 @@ class Linux(Base):
             'void',
             'fedora linux asahi remix'
         ]:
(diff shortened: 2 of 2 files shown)


mrobinson (changes requested):
This will cause a problem on unsupported Linux distributions, as it prevents the installation of `crown` and `taplo`.

sagudev:
> This will cause a problem on unsupported Linux distributions, as it prevents the installation of `crown` and `taplo`.

We only need to do this on windows.

mrobinson:
Hrm. Thinking out loud here, maybe the order shouldn't be so important when running code in `./mach bootstrap`. What do you think about emitting a warning on Linux, but then a "Press enter to continue" prompt so that it isn't lost in the other output?

sagudev:
> Hrm. Thinking out loud here, maybe the order shouldn't be so important when running code in `./mach bootstrap`. What do you think about emitting a warning on Linux, but then a "Press enter to continue" prompt so that it isn't lost in the other output?

Warning for unsupported Linux distributions, right?

mrobinson:
> Warning for unsupported Linux distributions, right?

Yep, exactly. Perhaps this should just warn and allow continuing, rather than simply exiting and skipping the installation of taplo and crown.

jschwe:
> Yep, exactly. Perhaps this should just warn and allow continuing, rather than simply exiting and skipping the installation of taplo and crown.

It would be nice if that also happened on windows. For example, I was quite happy that the `choco` step currently happens late in the bootstrap.  I don't have choco installed, but also don't need choco to install anything, since I only care about cross-compiling.

mrobinson (approved):

mrobinson:
@jschwe I've opened #32176 for the usecase you mention.

PR #31841, opened 2024-03-23
WebIDL: Remove `JSObject` from `Document::NamedGetter`

### LINKED ISSUE #30890: WebIDL impl: remove unsafe JSObject from return value of Document::NamedGetter
Part of https://github.com/servo/servo/issues/30862

Same as https://github.com/servo/servo/issues/30889

https://github.com/servo/servo/blob/8bbcf0abafc22cd840cd03f36b5b3b7d2d815493/components/script/dom/document.rs#L4829

Should return an enum of specific types. See https://html.spec.whatwg.org/multipage/dom.html#dom-tree-accessors%3Adom-document-nameditem-filter

eerii:
Removes the direct use of `unsafe NonNull<JSObject>` in favour of an enum of specific types `NamedPropertyValue` in `Document::NamedGetter`. The possible enum types described in the [specification](https://html.spec.whatwg.org/multipage/dom.html#dom-tree-accessors%3Adom-document-nameditem-filter) are `WindowProxy`, `Element` or `HTMLCollection`.

---
- [x] These changes fix #30890
- [x] These changes do not require tests because they only change the semantics of the code, they don't change behaviour

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -4858,7 +4858,7 @@ impl DocumentMethods for Document {
     }
 
     #[allow(unsafe_code)]
-    // https://html.spec.whatwg.org/multipage/#dom-tree-accessors:dom-document-nameditem-filter
+    /// <https://html.spec.whatwg.org/multipage/#dom-tree-accessors:dom-document-nameditem-filter>
     fn NamedGetter(&self, name: DOMString) -> Option<NamedPropertyValue> {
         if name.is_empty() {
             return None;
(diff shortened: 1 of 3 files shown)


components/script/dom/document.rs
```diff
@@ -4861,7 +4859,7 @@ impl DocumentMethods for Document {
 
     #[allow(unsafe_code)]
     // https://html.spec.whatwg.org/multipage/#dom-tree-accessors:dom-document-nameditem-filter
```
mrobinson, line 4861: Do you mind updating the comment here to be valid rustdoc while you are here?

```
/// <https://html.spec.whatwg.org/multipage/#dom-tree-accessors:dom-document-nameditem-filter>
```

mrobinson (approved):

components/script/dom/document.rs
eerii: Absolutely!

mrobinson (approved):

PR #31857, opened 2024-03-25
layout: More conservatively replace `Stylist`'s `Device`

mrobinson:
Instead of replacing Stylist's device on every reflow, only replace it
when the viewport changes. In addition, preserve the root font size from
the previous reflow fixing an issue where `rem` units were not properly
computed between reflows. This fixes a bug where fonts that are sized
using `rem` units change size on reload.

In addition, this also fixes an issue where changes to the `<font>`
element's `size` and `face` attributes weren't considered changes
to presentational hints. Not constantly updating the `Device` revealed
issues where modifications to these properties were not triggering
a restyle.

---

### DIFF
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -379,7 +379,10 @@ impl LayoutThread {
             root_flow: RefCell::new(None),
             // Epoch starts at 1 because of the initial display list for epoch 0 that we send to WR
             epoch: Cell::new(Epoch(1)),
-            viewport_size: Size2D::new(Au(0), Au(0)),
+            viewport_size: Size2D::new(
+                Au::from_f32_px(window_size.initial_viewport.width),
+                Au::from_f32_px(window_size.initial_viewport.height),
+            ),
             webrender_api,
             stylist: Stylist::new(device, QuirksMode::NoQuirks),
             rw_data: Arc::new(Mutex::new(LayoutThreadData {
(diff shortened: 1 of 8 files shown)


nicoburns:
Not sure if is helpful for me to do so, but I have reviewed this and it looks good to me. The only thing that seems surprising to me is that the `root_font_size` lives on the `StylistDevice`. But given that it does, it makes sense to me to preserve it.

mrobinson:
> Not sure if is helpful for me to do so, but I have reviewed this and it looks good to me. The only thing that seems surprising to me is that the `root_font_size` lives on the `StylistDevice`. But given that it does, it makes sense to me to preserve it.

This is a bit surprising, but the reason is pretty straightforward. Stylo needs to be able to resolve `rem` units which are relative to this font size.

nicoburns:
> This is a bit surprising, but the reason is pretty straightforward. Stylo needs to be able to resolve rem units which are relative to this font size.

Yes, but the `rem` size is just the font size on the `html` element (which can be overridden in author stylesheets), so I was expecting it to be set in the browser default stylesheet and for Stylo to be able to compute it itself. But now that I think about it more, I think rem is *also* affected by the user's browser font size setting, which would be awkward to do via default stylesheet, so it would make sense this is the value that is used if it _isn't_ overridden in the author stylesheets.

mukilan (approved):

PR #31862, opened 2024-03-25
Fix table with rows but no column

Loirooriol:
We weren't generating any fragment for the rows, which meant that JS APIs like clientWidth would be 0, and also outlines weren't painted.

This aligns Servo with Blink and WebKit. Gecko is broken, it distributes twice the table height among the rows.

---

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -1432,11 +1432,7 @@ impl<'a> TableLayout<'a> {
             };
         }
 
-        let fallback_size = LogicalVec2 {
-            block: self.final_table_height,
-            inline: self.assignable_width,
-        };
-        let table_and_track_dimensions = TableAndTrackDimensions::new(&self, &fallback_size);
+        let table_and_track_dimensions = TableAndTrackDimensions::new(&self);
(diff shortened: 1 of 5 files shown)


components/layout_2020/table/layout.rs
```diff
@@ -1432,7 +1432,11 @@ impl<'a> TableLayout<'a> {
             };
         }
 
-        let table_and_track_dimensions = TableAndTrackDimensions::new(&self);
+        let fallback_size = LogicalVec2 {
+            block: self.final_table_height,
+            inline: self.assignable_width,
+        };
```
mrobinson, line 1438: I would move `fallback_size` into `TableAndTrackDimensions::new` as it seems to only be used there. It might also be nice to put a short comment before its definition that explains that this is the size used for a dimension when that dimension has no table tracks.

mrobinson (approved):
Looks good to me with one small suggestion:

components/layout_2020/table/layout.rs
Loirooriol: The `self.table.size.width == 0 && self.table.size.height == 0` case uses the same values, so I thought it would be good to keep them together. But fine either way.

PR #31867, opened 2024-03-26
clippy: Fix warnings in `components/script/dom/request.rs`

Aaryakhandelwal:
Major changes are: 
1)  changed the implementation of Into<_> to From statement as the former is included in the latter by default.
2) Dereferenced wherever necessary.

---

- [X] These changes do not require tests because they only fix clippy warnings.

### DIFF
--- a/components/script/dom/request.rs
+++ b/components/script/dom/request.rs
@@ -179,7 +179,7 @@ impl Request {
         // Step 14
         if let Some(init_referrer) = init.referrer.as_ref() {
             // Step 14.1
-            let ref referrer = init_referrer.0;
+            let referrer = &init_referrer.0;
             // Step 14.2
             if referrer.is_empty() {
                 request.referrer = NetTraitsRequestReferrer::NoReferrer;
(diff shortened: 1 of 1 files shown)


components/script/dom/request.rs
```diff
@@ -666,9 +666,9 @@ impl BodyMixin for Request {
     }
 }
 
-impl Into<NetTraitsRequestCache> for RequestCache {
-    fn into(self) -> NetTraitsRequestCache {
-        match self {
+impl From<RequestCache> for NetTraitsRequestCache {
+    fn from(val: RequestCache) -> Self {
+        match val {
```
mrobinson, line 671:
```suggestion
impl From<RequestCache> for NetTraitsRequestCache {
    fn from(value: RequestCache) -> Self {
        match value {
```

mrobinson (commented):
Looks great! I only have one nit request for these changes. Let's avoid abbreviating in this case since "value" is only two characters longer and increases clarity.

mrobinson (changes requested):
Apologies. I didn't set the proper review flag before.

components/script/dom/request.rs
mrobinson: I've gone ahead and made these changes.

mrobinson (approved):

components/script/dom/request.rs
Aaryakhandelwal: Apologies, I was preoccupied with some personal work and couldn't attend to the changes requested. 
Thank you for making the commit.

mrobinson: No worries. Thank you for the contribution!

Issue #31503, opened 2024-03-05
WPT files are not always checked


sagudev:
See more info in fixup PR: https://github.com/servo/servo/pull/31497

Tidy check skipped the check:

> Skipping WPT lint checks, because no relevant files changed.

Relevant file: https://github.com/servo/servo/blob/main/python/tidy/tidy.py

I suspect either our detection of change is wrong (works via git log) or we do not check all necessary files. The only changed WPT file in PR is: [tests/wpt/mozilla/tests/mozilla/interfaces.html](https://github.com/servo/servo/pull/31481/files#diff-979634388bbe20c74e142a0e0bfb9bfbf996e8d73fdcc2d9cd4e35b95e5550b4)

mrobinson:
The tidy script figures out what files have changed by looking backward for the first change committed by GitHub. In this case, it looks like @Loirooriol may have asked GitHub to do a rebase (a perfectly reasonable thing!), but that had the consequence of making GitHub the committer:

```
commit 06aeeeb1f3574706e6992426063306a3f10716d8 (HEAD)
Author:     Oriol Brufau <obrufau@igalia.com>
AuthorDate: Sun Mar 3 13:47:39 2024 +0100
Commit:     GitHub <noreply@github.com>
CommitDate: Sun Mar 3 12:47:39 2024 +0000
```

We need a better way of detecting which change is the "base" in order to detect which files have changed.

mrobinson:
Note that this problems existed before, but in this case we recently expanded the use of file change detection to control whether or not the manifests were linted, which means this is more likely to happen now.

mrobinson:
An option here might be to look for the first commit with GitHub as the committer *before* the top of the branch.

mrobinson:
Though then there's no way to distinguish between changes made to the working tree and the changes that are committed.

Loirooriol:
I don't recall asking Github for a rebase here. Where are you seeing that info?

https://github.com/servo/servo/commit/06aeeeb1f3574706e6992426063306a3f10716d8 says `Loirooriol committed`

https://github.com/servo/servo/commit/06aeeeb1f3574706e6992426063306a3f10716d8.patch doesn't have `Commit:     GitHub <noreply@github.com>`

mrobinson:
[...]

PR #31874, opened 2024-03-26
tidy: Make sure to run WPT lints when `--all` is passed to tidy

mrobinson:
This also fixes the checked in WPT manifest.

---

- [x] These changes do not require tests because they fix tests on the CI.


Loirooriol (approved):
This should fix #31503, right?

PR #31878, opened 2024-03-26
clippy: Fix some warnings in `components/script/timers.rs`

komutumu:
Fixed clippy warnings in components/script/timers.rs

- [X] These changes do not require tests because it doesn't change functionality

### DIFF
--- a/components/script/timers.rs
+++ b/components/script/timers.rs
@@ -418,8 +418,8 @@ enum InternalTimerCallback {
     ),
 }
 
-impl JsTimers {
-    pub fn default() -> JsTimers {
+impl Default for JsTimers {
+    pub fn default() -> Self {
         JsTimers {
             next_timer_handle: Cell::new(JsTimerHandle(1)),
             active_timers: DomRefCell::new(HashMap::new()),
--- a/components/script/timers.rs
+++ b/components/script/timers.rs
@@ -131,7 +131,7 @@ impl PartialEq for OneshotTimer {
 impl OneshotTimers {
     pub fn new(scheduler_chan: IpcSender<TimerSchedulerMsg>) -> OneshotTimers {
         OneshotTimers {
-            js_timers: JsTimers::new(),
+            js_timers: JsTimers::default(),
             timer_event_chan: DomRefCell::new(None),
             scheduler_chan,
             next_timer_handle: Cell::new(OneshotTimerHandle(1)),
@@ -418,14 +418,8 @@ enum InternalTimerCallback {
     ),
 }
 
-impl Default for JsTimers {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl JsTimers {
-    pub fn new() -> JsTimers {
+    pub fn default() -> JsTimers {
         JsTimers {
             next_timer_handle: Cell::new(JsTimerHandle(1)),
             active_timers: DomRefCell::new(HashMap::new()),
--- a/components/script/timers.rs
+++ b/components/script/timers.rs
@@ -124,7 +124,7 @@ impl PartialOrd for OneshotTimer {
 impl Eq for OneshotTimer {}
 impl PartialEq for OneshotTimer {
     fn eq(&self, other: &OneshotTimer) -> bool {
-        self as *const OneshotTimer == other as *const OneshotTimer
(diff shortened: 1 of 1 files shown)


jahielkomu:
Hello @mrobinson, I hope you are well.
Please kindly review.

components/script/timers.rs
```diff
@@ -418,6 +418,12 @@ enum InternalTimerCallback {
     ),
 }
 
+impl Default for JsTimers {
```
mrobinson, line 421: Please don't add dead code. I think you can replace all calls `JSTimers::new()` with `JSTimers::default()` everywhere and move the contents of `new()` into `default()`.

mrobinson (changes requested):

components/script/timers.rs
jahielkomu: Thank you @mrobinson 
Indeed the implementation of  ```JsTimers::default``` behavior is unused.
However won't the change and replacement of ```new()``` with ```default()``` have effects on the use of ```JsTimers::new()```

mrobinson: Yes, every usage of `JSTimers::new()` should be replaced with `JSTimers::default()`.

jahielkomu: > Thank you @mrobinson Indeed the implementation of `JsTimers::default` behavior is unused. However won't the change and replacement of `new()` with `default()` have effects on the use of `JsTimers::new()`

Apologies @mrobinson 
I have reread your review and it does make sense.

jahielkomu:
I have update the PR @mrobinson 
Please kindly review.

components/script/timers.rs
```diff
@@ -419,7 +419,7 @@ enum InternalTimerCallback {
 }
 
 impl JsTimers {
-    pub fn new() -> JsTimers {
+    pub fn default() -> JsTimers {
```
mrobinson, line 422: This method should also be the implementation of the `Default` trait. Here's an example of this from the rust documentation: https://doc.rust-lang.org/std/default/trait.Default.html#how-can-i-implement-default

mrobinson, line 422: This method should also be the implementation of the `Default` trait. Here's an example of this from the rust documentation: https://doc.rust-lang.org/std/default/trait.Default.html#how-can-i-implement-default

mrobinson (changes requested):

jahielkomu:
Hello @mrobinson.

I have update the PR, please review and revert.
I did not think that it was necessary to change the visibility of the default associated function to default it to the surrounding visibility.

components/script/timers.rs
```diff
@@ -418,8 +418,8 @@ enum InternalTimerCallback {
     ),
 }
 
-impl JsTimers {
-    pub fn default() -> JsTimers {
+impl Default for JsTimers {
```
mrobinson, line 421: You'll need to have the `default()` method inside a block like this:

```rust
impl Default for JsTimers {
 ...
}
```

but the rest of the methods will need to be inside a block that looks like this:

```rust
impl JsTimers {
 ...
}
```

mrobinson:
> I have update the PR, please review and revert. I did not think that it was necessary to change the visibility of the default associated function to default it to the surrounding visibility.

@jahielkomu I left a comment above. The build is failing because you have surrounded all JsTimer structure functions with the `Default` implementation block. The thing is that you need only put this kind of block around the functions necessary to implement the `Default` trait.

mrobinson (approved):

jahielkomu:
> > I have update the PR, please review and revert. I did not think that it was necessary to change the visibility of the default associated function to default it to the surrounding visibility.
> 
> @jahielkomu I left a comment above. The build is failing because you have surrounded all JsTimer structure functions with the `Default` implementation block. The thing is that you need only put this kind of block around the functions necessary to implement the `Default` trait.

Yes @mrobinson 
I realized I confused myself from the beginning. 

I have updated the PR and have separated the ```Default``` implementation from the rest of the methods as in 

``` 
impl Default for JsTimers {
...
}
```
then

```
impl JsTimers {
...
}
```
Instead of 
```
impl JsTimers {
impl Default for JsTimers {
...
} 
...
}
 ```

Please review and revert. 
Thank you.

PR #31879, opened 2024-03-26
fonts: Add support for WOFF2 and properly load web fonts from @imports

### LINKED ISSUE #16257: Should we permit woff2 font hints?
URL font sources are ignored if they provide a hint that [does not match](https://dxr.mozilla.org/servo/rev/5ef8a0a2f7917792105cf28f6e581a92e863c3dd/components/style/font_face.rs#120) truetype, opentype, or woff. I've seen sites using woff2 as a format; do we support that?

### LINKED ISSUE #31598: Web fonts are not loaded from stylsheets included via `@import` rules
Web fonts can be loaded from stylesheets, but when those stylesheets are updated via `@import` rules, the web fonts from the imported stylesheet are not loaded at all.

mrobinson:
This change also makes two fixes that are necessary to get WOFF2 font
tests working:

1. It adds support for loading web fonts from stylesheets included via
   @import rules.
2. It ensure that when web fonts are loaded synchronusly they invalidate
   the font cache. This led to incorrect font rendering when running
   tests before.

Fixes #31598. Fixes #16257.

---

- [x] These changes fix #31598 and fix #16257.

### DIFF
--- a/components/gfx/font_cache_thread.rs
+++ b/components/gfx/font_cache_thread.rs
@@ -559,9 +559,8 @@ impl FontCacheThread {
             // Either increment the count of loading web fonts, or wait for a synchronous load.
             if let Some(ref receiver) = receiver {
                 receiver.recv().unwrap();
-            } else {
-                number_loading += 1;
             }
+            number_loading += 1;
         });
 
         number_loading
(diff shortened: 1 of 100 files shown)


jdm (approved):
Excellent work identifying the issue with the font cache invalidation!

PR #31880, opened 2024-03-26
WebIDL: Replace `NonNull<JSObject>` in `xmlhttprequest`

eerii:
Changes the return type of `XMLHttpRequest::arraybuffer_response` from `Option<NonNull<JSObject>>` to a safe `Option<ArrayBuffer>`. Since I was already there, I also updated the function comments to proper docsring.

---
- [x] These changes are a part of #30889
- [x] These changes do not require tests because they don't add new behaviour

### DIFF
--- a/components/script/dom/xmlhttprequest.rs
+++ b/components/script/dom/xmlhttprequest.rs
@@ -5,7 +5,6 @@
 use std::borrow::ToOwned;
 use std::cell::Cell;
 use std::default::Default;
-use std::ptr::NonNull;
 use std::str::{self, FromStr};
 use std::sync::{Arc, Mutex};
 use std::{cmp, ptr, slice};
@@ -25,7 +24,7 @@ use js::jsapi::{Heap, JSObject, JS_ClearPendingException};
 use js::jsval::{JSVal, NullValue, UndefinedValue};
 use js::rust::wrappers::JS_ParseJSON;
 use js::rust::HandleObject;
-use js::typedarray::{ArrayBuffer, CreateWith};
+use js::typedarray::{ArrayBuffer, ArrayBufferU8};
 use mime::{self, Mime, Name};
 use net_traits::request::{CredentialsMode, Destination, Referrer, RequestBuilder, RequestMode};
 use net_traits::CoreResourceMsg::Fetch;
@@ -41,6 +40,7 @@ use url::Position;
 
 use crate::body::{BodySource, Extractable, ExtractedBody};
 use crate::document_loader::DocumentLoader;
+use crate::dom::bindings::buffer_source::{create_buffer_source, HeapBufferSource};
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::WindowBinding::WindowMethods;
 use crate::dom::bindings::codegen::Bindings::XMLHttpRequestBinding::{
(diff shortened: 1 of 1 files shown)


jdm (approved):
Thank you for cleaning this up!

mrobinson:
It looks like this is causing a test to start failing:

```
OK /xhr/response-data-arraybuffer.htm
    FAIL [expected PASS] subtest: XMLHttpRequest: The response attribute: ArrayBuffer data
    assert_equals: Response should be cached expected object "[object ArrayBuffer]" but got object "[object ArrayBuffer]"
```

The text is quite mysterious, but perhaps underneath these are two different types?

eerii:
I'll look into this

eerii:
Ok, the test should be fixed now. It was indeed a cache problem, I was checking if `self.response_arraybuffer` had a value but not setting it after creation. Bonus, having the `set_data` made the code even simpler.

components/script/dom/xmlhttprequest.rs
```diff
@@ -1347,29 +1347,18 @@ impl XMLHttpRequest {
...
-                .and_then(|()| {
-                    self.response_arraybuffer.set(array_buffer.get());
-                    Some(NonNull::new_unchecked(array_buffer.get()))
-                })
+        if !self.response_arraybuffer.is_initialized() {
+            // Step 2
+            let bytes = self.response.borrow();
+            self.response_arraybuffer.set_data(cx, &bytes).ok()?;
         }
+        self.response_arraybuffer.get_buffer().ok()
```
eerii, line 1358: However, now I'm not sure the step comments make much sense. Looking into the spec it says:

```
5. If this’s response type is "arraybuffer", then set this’s response object to a new ArrayBuffer object representing this’s received bytes. If this throws an exception, then set this’s response object to failure and return null. 
```

I'm not sure what the two substeps were exactly referring to. Does it make sense to leave them like this, should I move them somewhere else or remove them?

mrobinson: Yeah...this is an issue. There was a tendency to only include the step numbers in the code until now, but the spec renumbers the operations all the time. I would just try to implement what the specification says now and don't be afraid to renumber things (but this time include a description of the step with the number).

eerii: That makes a lot of sense. Would something like this last commit work?

mrobinson: Yes, that looks great. Thank you!

mrobinson: Oh, wait. I'm sorry, I spoke to soon. The step numbers and text should correspond to what is in the specification at https://xhr.spec.whatwg.org/#arraybuffer-response.

mrobinson: I'm guessing what happened here is that an older version of the specification had this separated into different sections but the latest version has it in a single one -- which is why the step numbers don't really make sense any longer.

eerii: Yes, I figured it was something like that. Should I simply remove the step numbers and leave the comments as they are then? Or change the top comment to `Step 5:` and remove the other step?

mrobinson: That makes sense to me. Thanks!

eerii: Thank you! It should be ready now

mrobinson (approved):

PR #31417, opened 2024-02-23
Initial internal support for multiple webviews

wusyong:
This patch improves support for multiple top-level browsing contexts aka “webviews” by adding several messages that allow the embedder to show, hide, move, and resize webviews in the compositor, and notify the embedder when painting order and focus changes.

This is a rebased version of #30648, while addressing review feedback given in #30767, #30840, #30841, and #30842. Notably, “browsers” have been renamed to “webviews”.

There are no user-facing changes in servoshell yet, but these changes can be used to create a multiple-document interface (#30785) or implement tabbed browsing (#31545). For the full design of this feature, see #30648.

## limitations
- desktop-style page zoom is still global
- browsers with transparent or translucent content will misbehave
- sending ShowWebView and HideWebView for one webview at a time can be inefficient
- if too many webviews are shown at once, the compositor can get stuck in a laggy state with pending epoch warnings (to reproduce, try [#31545](https://github.com/servo/servo/pull/31545) with [4.html](https://bucket.daz.cat/work/igalia/servo/4.html) and click “_blank” six or seven times)

## compositor
- WebViewManager (new) manages webview-related storage and painting order
    - it allows the compositor to store arbitrary data for each webview id
    - it keeps track of which webviews are visible, and what order to paint them in
- we now use the iframe-in-dummy-root-pipeline technique to composite the webviews together, not just when we have a non-unity pinch zoom

## servoshell
- WebViewManager (was WebView) keeps a copy of the webview rects and focus
- but for now, we always focus and show the newest webview only

## messages
embedder → compositor (no ipc)
- WindowResize (was Resize) no longer affects the viewport of individual webviews
- Zoom (existing) no longer affects the viewport of individual webviews
- ResetZoom (existing) no longer affects the viewport of individual webviews
- MoveResizeWebView (new) updates a webview’s rect, and sends new viewport to script in descendant pipelines
- ShowWebView (new) adds a webview to painting order, and notifies embedder of the new painting order
- HideWebView (new) removes a webview from painting order, and notifies embedder of the new painting order
- RaiseWebViewToTop (new) moves a webview to the top of the painting order, and notifies embedder accordingly

embedder →\* constellation
- SetWebViewThrottled (was WindowVisibility) has been reworked to be less confusing ([#31815](https://github.com/servo/servo/pull/31815), [#31816](https://github.com/servo/servo/pull/31816))

constellation → compositor
- CreateOrUpdateWebView (was SetFrameTree) now adds or updates a webview without clobbering others
[...]

PR #31892, opened 2024-03-27
compositing: Send entire scene's scroll offsets when sending WebRender display lists

### LINKED ISSUE #31807: Scroll offsets are sometimes lost when updating WebRender display list
To reproduce:

1. Create test.html as below (and iframe.html with enough text to scroll)
2. Scroll the iframe down to some non-zero y
3. Observe that the iframe appears to scroll back to the top

```html
<style>
    div {
        transform: translateY(7em) rotate(15deg);
    }
    .toggle {
        transform: translateY(7em) rotate(30deg);
    }
</style>
<div><iframe src="iframe.html"></iframe></div>
<script>
    setInterval(() => {
        document.querySelector("div").classList.toggle("toggle");
    }, 500);
</script>
```

mrobinson:
WebRender does not preserve spatial tree offsets when updating the
spatial tree. Updating the spatial tree of a pipeline can also
update the spatial tree of child pipelines. This change ensures that
WebRender always gets the scroll offsets of the entire scene when
modifying display lists in a way that may rebuild the spatial tree.

Fixes #31807.

---

- [x] These changes fix #31807.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -764,19 +764,7 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
                 let mut transaction = Transaction::new();
                 transaction
                     .set_display_list(display_list_info.epoch, (pipeline_id, built_display_list));
-
-                for node in details.scroll_tree.nodes.iter() {
-                    if let (Some(offset), Some(external_id)) = (node.offset(), node.external_id()) {
-                        let offset = LayoutVector2D::new(-offset.x, -offset.y);
-                        transaction.set_scroll_offsets(
-                            external_id,
-                            vec![SampledScrollOffset {
-                                offset,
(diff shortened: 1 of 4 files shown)


mrobinson:
Hey @delan, just let me know if you'd like me to hold off on this one until you've landed #31417.

delan:
> Hey @delan, just let me know if you'd like me to hold off on this one until you've landed #31417.

Go ahead, I probably won’t be able to land my patch until Monday anyway.

delan (commented):
The impl looks good, but can we have an automated test for this?

mrobinson:
> The impl looks good, but can we have an automated test for this?

At first I didn't think so, but after a bit of pondering it turns out that there is a way to test this. I've added that test to the latest version of the branch.

delan (approved):

PR #31894, opened 2024-03-27
clippy: Fix a variety of warnings in `components/script/dom`

ektuu:
---

- [x] These changes are  part of https://github.com/servo/servo/issues/31500
- [x] These changes do not require tests because only clippy warnings were fixed

### DIFF
--- a/components/script/dom/bindings/root.rs
+++ b/components/script/dom/bindings/root.rs
@@ -76,8 +76,16 @@ where
     }
 }
 
-/// Represents values that can be rooted through a stable address that will
+/// `StableTraceObject` represents values that can be rooted through a stable address that will
 /// not change for their whole lifetime.
+/// It is an unsafe trait that requires implementors to ensure certain safety guarantees.
+///
+/// # Safety
+///
+/// Implementors of this trait must ensure that the `trace` method correctly accounts for all
+/// owned and referenced objects, so that the garbage collector can accurately determine which
+/// objects are still in use. Failing to adhere to this contract may result in undefined behavior,
+/// such as use-after-free errors.
 pub unsafe trait StableTraceObject {
     /// Returns a stable trace object which address won't change for the whole
     /// lifetime of the value.
@@ -267,10 +275,7 @@ impl RootCollection {
     unsafe fn unroot(&self, object: *const dyn JSTraceable) {
         assert_in_script();
         let roots = &mut *self.roots.get();
-        match roots
-            .iter()
-            .rposition(|r| *r as *const () == object as *const ())
-        {
+        match roots.iter().rposition(|r| std::ptr::eq(*r, object)) {
             Some(idx) => {
                 roots.remove(idx);
             },
@@ -486,7 +491,7 @@ impl<T> Eq for Dom<T> {}
 
 impl<T> PartialEq for LayoutDom<'_, T> {
     fn eq(&self, other: &Self) -> bool {
-        self.value as *const T == other.value as *const T
+        std::ptr::eq(self.value, other.value)
     }
 }
 
--- a/components/script/dom/bindings/str.rs
+++ b/components/script/dom/bindings/str.rs
@@ -40,6 +40,11 @@ impl ByteString {
         self.0.len()
     }
 
+    /// Checks if the ByteString is empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
     /// Returns `self` with A–Z replaced by a–z.
     pub fn to_lower(&self) -> ByteString {
         ByteString::new(self.0.to_ascii_lowercase())
@@ -365,7 +370,7 @@ impl DOMString {
         let (year_int, month_int) = parse_month_component(value)?;
 
         // Step 4
-        if value.split("-").nth(2).is_some() {
+        if value.split('-').nth(2).is_some() {
             return None;
         }
         // Step 5
@@ -443,7 +448,7 @@ impl DOMString {
             if !(
                 // A valid number is the same as what rust considers to be valid,
                 // except for +1., NaN, and Infinity.
-                val.is_infinite() || val.is_nan() || input.ends_with(".") || input.starts_with('+')
+                val.is_infinite() || val.is_nan() || input.ends_with('.') || input.starts_with('+')
             ) {
                 return Some(val);
             }
@@ -677,7 +682,7 @@ fn parse_month_component(value: &str) -> Option<(i32, u32)> {
 
     // Step 4, 5
     let month_int = month.parse::<u32>().ok()?;
-    if month.len() != 2 || month_int > 12 || month_int < 1 {
+    if month.len() != 2 || !(1..=12).contains(&month_int) {
         return None;
     }
 
--- a/components/script/dom/bindings/xmlname.rs
+++ b/components/script/dom/bindings/xmlname.rs
@@ -90,8 +90,7 @@ pub enum XMLName {
 /// for details.
 pub fn xml_name_type(name: &str) -> XMLName {
     fn is_valid_start(c: char) -> bool {
-        match c {
-            ':' |
+        matches!(c, ':' |
             'A'..='Z' |
             '_' |
             'a'..='z' |
@@ -106,9 +105,7 @@ pub fn xml_name_type(name: &str) -> XMLName {
             '\u{3001}'..='\u{D7FF}' |
             '\u{F900}'..='\u{FDCF}' |
             '\u{FDF0}'..='\u{FFFD}' |
-            '\u{10000}'..='\u{EFFFF}' => true,
-            _ => false,
-        }
+            '\u{10000}'..='\u{EFFFF}')
     }
 
     fn is_valid_continuation(c: char) -> bool {
--- a/components/script/dom/htmlselectelement.rs
+++ b/components/script/dom/htmlselectelement.rs
@@ -518,11 +518,10 @@ impl Validatable for HTMLSelectElement {
         // https://html.spec.whatwg.org/multipage/#the-select-element%3Asuffering-from-being-missing
         if validate_flags.contains(ValidationFlags::VALUE_MISSING) && self.Required() {
             let placeholder = self.get_placeholder_label_option();
-            let selected_option = self
+            let is_value_missing = !self
                 .list_of_options()
-                .filter(|e| e.Selected() && placeholder.as_ref() != Some(e))
-                .next();
-            failed_flags.set(ValidationFlags::VALUE_MISSING, selected_option.is_none());
+                .any(|e| e.Selected() && placeholder != Some(e));
+            failed_flags.set(ValidationFlags::VALUE_MISSING, is_value_missing);
         }
 
         failed_flags
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -10,7 +10,7 @@ use std::default::Default;
 use std::ops::Range;
 use std::slice::from_ref;
 use std::sync::Arc as StdArc;
-use std::{cmp, iter, mem};
+use std::{cmp, iter};
 
 use app_units::Au;
 use bitflags::bitflags;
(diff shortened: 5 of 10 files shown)


components/script/dom/bindings/xmlname.rs
```diff
@@ -90,25 +90,22 @@ pub enum XMLName {
...
+    '\u{F8}'..='\u{2FF}' |
+    '\u{370}'..='\u{37D}' |
+    '\u{37F}'..='\u{1FFF}' |
+    '\u{200C}'..='\u{200D}' |
+    '\u{2070}'..='\u{218F}' |
+    '\u{2C00}'..='\u{2FEF}' |
+    '\u{3001}'..='\u{D7FF}' |
+    '\u{F900}'..='\u{FDCF}' |
+    '\u{FDF0}'..='\u{FFFD}' |
+    '\u{10000}'..='\u{EFFFF}')
```
Loirooriol, line 108: This indentation looks weird

components/script/dom/htmltextareaelement.rs
```diff
@@ -100,7 +100,7 @@ impl LayoutHTMLTextAreaElementHelpers for LayoutDom<'_, HTMLTextAreaElement> {
             // placeholder is single line, but that's an unimportant detail.
             self.placeholder()
                 .replace("\r\n", "\n")
-                .replace("\r", "\n")
+                .replace('\r', "\n")
```
Loirooriol, line 103: Is this https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern ?
But I guess `replace` accepts a str and not a char, just looking at the line above.

components/script/dom/node.rs
```diff
@@ -604,6 +604,10 @@ impl Node {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
```
Loirooriol, line 608: The point of `is_empty` is that it can be faster than `len()`. But `CharacterData`'s `Length` iterates the string (see https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#264-266).

So if you add this, you should implement this in a faster way.

components/script/dom/bindings/str.rs
```diff
@@ -40,6 +40,11 @@ impl ByteString {
         self.0.len()
     }
 
+    /// Checks if the ByteString is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
```
Loirooriol, line 45: The point of `is_empty` is that it can be faster than `len()`. In this case `Vec.is_empty()` does use `Vec.len()`, but that's an implementation detail that we shouldn't rely on. So this should be

```suggestion
        self.0.is_empty()
```

Loirooriol (changes requested):
I would recommend splitting this into different PRs, and address a single clippy warning on each one, and link to it in the description.

components/script/dom/bindings/root.rs
```diff
@@ -269,7 +279,7 @@ impl RootCollection {
         let roots = &mut *self.roots.get();
         match roots
             .iter()
-            .rposition(|r| *r as *const () == object as *const ())
+            .rposition(|r| std::ptr::eq(*r as *const _, object as *const ()))
```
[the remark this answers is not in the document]
ektuu: I apologise for the mistake.
 fixed it!

components/script/dom/bindings/xmlname.rs
ektuu: Will fix it!

components/script/dom/node.rs
ektuu: Got it!

components/script/dom/bindings/root.rs
```diff
@@ -509,17 +519,14 @@ impl<T> Clone for Dom<T> {
...
+        Dom { ptr: self.ptr }
     }
 }
 
 impl<T> Clone for LayoutDom<'_, T> {
     #[inline]
     fn clone(&self) -> Self {
-        assert_in_layout();
-        LayoutDom { value: self.value }
+        *self
```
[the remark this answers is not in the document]
ektuu: Screenshot 2024-03-27 at 19 43 46


I encountered the above warning and thus removed assert and implemented the suggested code.

components/script/dom/node.rs
```diff
@@ -1314,10 +1317,10 @@ where
 #[allow(unsafe_code)]
 pub unsafe fn from_untrusted_node_address(candidate: UntrustedNodeAddress) -> DomRoot<Node> {
     // https://github.com/servo/servo/issues/6383
-    let candidate: uintptr_t = mem::transmute(candidate.0);
+    let candidate: uintptr_t = mem::transmute(candidate.0 as usize);
     //        let object: *mut JSObject = jsfriendapi::bindgen::JS_GetAddressableObject(runtime,
     //                                                                                  candidate);
-    let object: *mut JSObject = mem::transmute(candidate);
+    let object: *mut JSObject = mem::transmute(candidate as *mut js::jsapi::JSObject);
```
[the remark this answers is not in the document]
ektuu: Screenshot 2024-03-27 at 19 49 59

I encountered the above warning.

components/script/dom/bindings/root.rs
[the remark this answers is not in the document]
mrobinson: Please copy and paste terminal output into comments. You can surround it with three backticks to preserve formatting like this:

````
```
warning: non-canonical implementation of 'clone' on a 'Copy' type components/script/dom/bindings/root.rs:
...
```
````

In this case does the error still presist if you modify the function to be what's below?

```rust
fn clone(&self) -> Self {
    assert_in_layout();
    { *self }
}
```    

The assertion is important here as it checking an invariant to using the JavaScript engine. I don't think we can remove it in this case.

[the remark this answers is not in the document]
Loirooriol: I guess the problem is that copying doesn't run the assert, so it's a bit pointless to only have it when cloning.
Possibly we should remove the Copy trait to enforce `.clone()` in order to always run the assert?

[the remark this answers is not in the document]
ektuu: > Please copy and paste terminal output into comments. You can surround it with three backticks to preserve formatting like this:
> 
> ```
> ```
> warning: non-canonical implementation of 'clone' on a 'Copy' type components/script/dom/bindings/root.rs:
> ...
> ```
> ```
> 
> In this case does the error still presist if you modify the function to be what's below?
> 
> ```rust
> fn clone(&self) -> Self {
>     assert_in_layout();
>     { *self }
> }
> ```
> 
> The assertion is important here as it checking an invariant to using the JavaScript engine. I don't think we can remove it in this case.


The error is removed after this change

components/script/dom/bindings/root.rs
```diff
@@ -78,6 +78,16 @@ where
 
 /// Represents values that can be rooted through a stable address that will
 /// not change for their whole lifetime.
+///
+///
+/// /// `StableTraceObject` is an unsafe trait that requires implementors to ensure certain safety guarantees.
```
mrobinson, line 83:
```suggestion
/// /// `StableTraceObject` is an unsafe trait that requires implementors to ensure certain safety guarantees.
```

There are two levels of rustdoc comments here.

components/script/dom/node.rs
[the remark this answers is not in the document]
Loirooriol: Then you aren't applying the suggested change. It should be

```rs
    let candidate = candidate.0 as usize;
```
```rs
    let object = candidate as *JSObject;
```

And you can stop importing `mem`.

Loirooriol (changes requested):
I still think it would be better to split this into different PRs, each one addressing a single clippy warning, and link to it in the clippy docs. This link should be in text, not in an image.

Loirooriol:
You need to rebase, there are conflicts in `components/script/dom/bindings/str.rs`

components/script/dom/node.rs
```diff
@@ -1317,10 +1317,10 @@ where
 #[allow(unsafe_code)]
 pub unsafe fn from_untrusted_node_address(candidate: UntrustedNodeAddress) -> DomRoot<Node> {
     // https://github.com/servo/servo/issues/6383
-    let candidate: uintptr_t = mem::transmute(candidate.0 as usize);
-    //        let object: *mut JSObject = jsfriendapi::bindgen::JS_GetAddressableObject(runtime,
+    let candidate = candidate.0 as usize;
+     //        let object: *mut JSObject = jsfriendapi::bindgen::JS_GetAddressableObject(runtime,
```
Loirooriol, line 1321: Don't change the indentation

components/script/dom/node.rs
```diff
@@ -604,6 +604,10 @@ impl Node {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.data.len() == 0
```
Loirooriol, line 608: `self.data` doesn't even exist??

You will need a `match self.type_id()` and do something analogous to `len()`, but without calling `CharacterData.Length()`

components/script/dom/node.rs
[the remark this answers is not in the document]
Loirooriol: You didn't stop importing `mem`. Please make sure not to introduce compile warnings.

components/script/dom/node.rs
```diff
@@ -1203,8 +1207,7 @@ impl Node {
                 match last_child.and_then(|node| {
                     node.inclusively_preceding_siblings()
                         .filter_map(DomRoot::downcast::<Element>)
-                        .filter(|elem| is_delete_type(elem))
-                        .next()
+                        .find(|elem| is_delete_type(elem))
```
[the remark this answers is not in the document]
Loirooriol: I thought this would work but it doesn't compile. Please make sure that your patch compiles even if you are only doing requested changes, reviewers can be wrong.

Loirooriol (changes requested):

components/script/dom/node.rs
ektuu: I would revert this change for now and might solve it separately

components/script/dom/htmlselectelement.rs
```diff
@@ -520,7 +520,7 @@ impl Validatable for HTMLSelectElement {
             let placeholder = self.get_placeholder_label_option();
             let is_value_missing = !self
                 .list_of_options()
-                .any(|e| e.Selected() && placeholder.as_ref() != Some(e));
+                .any(|e| e.Selected() && placeholder.as_ref() != Some(&e));
```
Loirooriol, line 523:
```suggestion
                .any(|e| e.Selected() && placeholder != Some(e));
```

ektuu:
I have made all the suggested changes ,but WPT Export keeps failing unexpectedly .What should I do?

components/script/dom/bindings/root.rs
```diff
@@ -76,8 +76,16 @@ where
...
+/// `StableTraceObject` represents values that can be rooted through a stable address that will
 /// not change for their whole lifetime.
+/// It is an unsafe trait that requires implementors to ensure certain safety guarantees.
+///
+/// # Safety
+///
+/// Implementors of this trait must ensure that the `trace` method correctly accounts for all
+/// owned and referenced objects, so that the garbage collector can accurately determine which
+/// objects are still in use. Failing to adhere to this contract may result in undefined behavior,
+/// such as use-after-free errors.
```
mrobinson, line 88: Is this an accurate description of what will happen if trace is improperly implemented? Is use-after-free possible or is the failure case a memory leak? Pinging @jdm and @sagudev who are probably the ones to best review this documentation.

mrobinson:
> I have made all the suggested changes ,but WPT Export keeps failing unexpectedly .What should I do?

This failure is unrelated to your change and should not block anything. You can try to rebase your work to see if that fixes things.

sagudev:
> > I have made all the suggested changes ,but WPT Export keeps failing unexpectedly .What should I do?
> 
> This failure is unrelated to your change and should not block anything. You can try to rebase your work to see if that fixes things.

Merge commit that include changes in WPT causes WPT Export failures.

components/script/dom/bindings/root.rs
sagudev: Sounds about right. If we do not trace js managed object SM can GC it while it is still referenced by object with bad tracer and that can cause use-after-free.

mrobinson (approved):

Loirooriol (approved):

mrobinson:
I'll rebase and squash this one before landing to clean up the commit history.

PR #31896, opened 2024-03-27
Ignore spaces before atomic inline for the min-content size

Loirooriol:
For the min-content size we should wrap lines wherever is possible,
so wrappable spaces shouldn't increase the length of the line,
they will just be removed or hang at the end of the line.

---

### DIFF
--- a/components/layout_2020/flow/inline.rs
+++ b/components/layout_2020/flow/inline.rs
@@ -2380,6 +2380,9 @@ impl<'a> ContentSizesComputation<'a> {
                     self.containing_block_writing_mode,
                 );
 
+                // For the min-content size we should wrap lines wherever is possible,
+                // so wrappable spaces shouldn't increase the length of the line,
+                // they will just be removed or hang at the end of the line.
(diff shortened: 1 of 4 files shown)


components/layout_2020/flow/inline.rs
```diff
@@ -2380,7 +2380,7 @@ impl<'a> ContentSizesComputation<'a> {
                     self.containing_block_writing_mode,
                 );
 
-                self.current_line.min_content += self.pending_whitespace + outer.min_content;
+                self.current_line.min_content += outer.min_content;
```
mrobinson, line 2383: I think this might look like a mistake without a comment and the one you made for this PR is great:

```suggestion
                // For the min-content size we should wrap lines wherever is possible,
                // so wrappable spaces shouldn't increase the length of the line,
                // they will just be removed or hang at the end of the line.
                self.current_line.min_content += outer.min_content;
```

mrobinson (approved):
Nice fix.

PR #31910, opened 2024-03-27
clippy: Fix various clippy problems in `components/scripts/dom`

six-shot:
---

- [x] These changes are part of #31500.
- [x] These changes do not require tests because they do not modify functionality.

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -2187,7 +2187,7 @@ impl GlobalScope {
 
         if let Some(index) = uncaught_rejections
             .iter()
-            .position(|promise| *promise == Heap::boxed(rejection.get()))
+            .position(|promise| *promise == *Heap::boxed(rejection.get()))
         {
             uncaught_rejections.remove(index);
         }
@@ -2200,7 +2200,7 @@ impl GlobalScope {
     pub fn add_consumed_rejection(&self, rejection: HandleObject) {
         self.consumed_rejections
             .borrow_mut()
-            .push(Heap::boxed(rejection.get()));
+            .push(*Heap::boxed(rejection.get()));
     }
 
     pub fn remove_consumed_rejection(&self, rejection: HandleObject) {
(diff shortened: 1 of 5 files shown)


components/script/dom/globalscope.rs
```diff
@@ -294,15 +294,15 @@ pub struct GlobalScope {
     ///
     /// <https://html.spec.whatwg.org/multipage/#about-to-be-notified-rejected-promises-list>
     #[ignore_malloc_size_of = "mozjs"]
-    uncaught_rejections: DomRefCell<Vec<Box<Heap<*mut JSObject>>>>,
```
mrobinson, line 297: What was the error that led you to removing the `Box` here?

six-shot: It was saying box is unnecessary 
https://rust-lang.github.io/rust-clippy/master/index.html#/vec_box

components/script/dom/globalscope.rs
```diff
@@ -2179,42 +2179,42 @@ impl GlobalScope {
     pub fn add_uncaught_rejection(&self, rejection: HandleObject) {
         self.uncaught_rejections
             .borrow_mut()
-            .push(Heap::boxed(rejection.get()));
+            .push(*Heap::boxed(rejection.get()));
```
mrobinson, line 2182: This is creating a `Box<Heap<>>` and then immediately pulling it apart. I would either keep using `Box<...>` in the vector or figure out a way to create this `Heap` without putting it in a `Box<...>`.

components/script/dom/globalscope.rs
mrobinson: I think for now you should just allow the warning for `uncaught_rejections` and `consumed_rejections` below. It seems that `Heap` always comes boxed and it's tricky to undo that. Instead put a comment like:

```rust
/// TODO: `Vec` already stores members in a `Box` so this extra box is an unnecessary layer of
/// indirection. Can it be removed?
```

cc @sagudev

mrobinson (changes requested):

mrobinson (approved):

PR #31913, opened 2024-03-27
wpt: Enable tests in `/css/css-inline/baseline-source/`

### LINKED ISSUE #31875: Implement baseline-source
After #31705 I think it should be quite straightforward to implement `baseline-source`: https://drafts.csswg.org/css-inline/#baseline-source

This will require Stylo changes, so you can follow the instructions from #31723

Basically https://github.com/servo/stylo/blob/bcdebc41b2a2222e9fa5ae6d7d1572025c32dc47/style/properties/longhands/box.mako.rs#L107 needs to change the obsolete `servo-2013` into `servo`

And then https://github.com/servo/servo/blob/d814d0553981777c01409d545e5c1c335b8eb7f4/components/layout_2020/flow/inline.rs#L2145-L2153 needs to check `BaselineSource`, picking the 1st baseline for `BaselineSource::First`, the last baseline for `BaselineSource::Last`, and keep the current logic for `BaselineSource::Auto`.

MunishMummadi:
**Description**
-Tests for `baseline-source`

- [x] These changes fix #31875 

addresses the  [issue](https://github.com/servo/servo/issues/31875)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1200,7 +1200,7 @@ dependencies = [
 [[package]]
 name = "derive_common"
 version = "0.0.1"
-source = "git+https://github.com/servo/stylo?rev=refs/pull/26/head#11bf0832b412cd359f81e8f7d0e7e2665d11ab0a"
+source = "git+https://github.com/servo/stylo?branch=2023-09-01#bcdebc41b2a2222e9fa5ae6d7d1572025c32dc47"
 dependencies = [
  "darling",
  "proc-macro2",
@@ -3450,7 +3450,7 @@ dependencies = [
 [[package]]
 name = "malloc_size_of"
 version = "0.0.1"
-source = "git+https://github.com/servo/stylo?rev=refs/pull/26/head#11bf0832b412cd359f81e8f7d0e7e2665d11ab0a"
+source = "git+https://github.com/servo/stylo?branch=2023-09-01#bcdebc41b2a2222e9fa5ae6d7d1572025c32dc47"
 dependencies = [
  "accountable-refcell",
  "app_units",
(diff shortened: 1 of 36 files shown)


Cargo.toml
Loirooriol, line 1: Don't modify this file

Cargo.lock
Loirooriol, line 1: Don't modify this file

tests/wpt/include.ini
```diff
@@ -48,7 +48,7 @@ skip: true
   [css-highlight-api]
     skip: true
   [css-inline]
-    skip: true
```
Loirooriol, line 51: We should probably keep skipping the `initial-letter` and `text-box-trim` subdirectories.

So I would go with

```
  [css-inline]
    skip: true
    [baseline-source]
      skip: false
```

or

```
  [css-inline]
    skip: false
    [initial-letter]
      skip: true
    [text-box-trim]
      skip: true
```

There's also few tests in `animation`, `parsing`, and at the root. I don't think it matters much if they are enabled or not, but if enabled you will need to update the expectations.

Loirooriol (changes requested):
Please change the title. This is not implementing baseline-source, it's just enabling tests.

tests/wpt/include.ini
MunishMummadi: I will stick with the 1st option

Loirooriol (approved):
Looks good if https://github.com/servo/servo/actions/runs/8475945097 passes.

Sometimes tests behave differently in CI (e.g. because in your system you have different fonts), so you may need some manual adjustments if something fails.

MunishMummadi:
Do u want me to update description. I will be on lookout for tests.

mrobinson (approved):

PR #31920, opened 2024-03-28
Update servo's testharnessreport

### LINKED ISSUE #31919: `--timeout-multiplier` does not affect test harness timeout
As noted in https://github.com/servo/servo/pull/31772#issuecomment-2022523518, I also got conformation on WPT's matrix chat room.

sagudev:
Update servo's testharnessreport to include new props as in [general testharnessreport](https://github.com/web-platform-tests/wpt/blob/d6c9b797f371f92fb23681150148e9ec81a63510/tools/wptrunner/wptrunner/testharnessreport.js#L73).

This makes `--timeout-multiplier` actually work.

---

- [x] These changes fix #31919 
- [x] These changes are tested manually with try run: https://github.com/sagudev/servo/actions/runs/8466221634 (longer times in WPT tests means we are waiting more time).

### DIFF
--- a/tests/wpt/tests/tools/wptrunner/wptrunner/testharnessreport-servo.js
+++ b/tests/wpt/tests/tools/wptrunner/wptrunner/testharnessreport-servo.js
@@ -1,4 +1,9 @@
-var props = {output:%(output)d, debug: %(debug)s};
+var props = {
+    output:%(output)d,
+    timeout_multiplier: %(timeout_multiplier)s,
+    explicit_timeout: %(explicit_timeout)s,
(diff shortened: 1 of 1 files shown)


jdm (approved):

mrobinson (approved):
We could also look into removing the Servo-specific version of this if it is no longer necessary.

sagudev:
> We could also look into removing the Servo-specific version of this if it is no longer necessary.

I think that requires a webdriver support as currently most important part of testharnessreport is reporting (currently this is done via specially crafted console.log messages).

PR #31934, opened 2024-03-29
Update HtmlInputElement webidl and implement setter for the files property

shanehandley:
Removes `readonly` from the `files` attribute for HtmlInputElement. It was previously `readonly` however the [webidl](https://html.spec.whatwg.org/multipage/input.html#htmlinputelement) was updated to make it mutable, with the spec defining the [setter behaviour](https://html.spec.whatwg.org/multipage/input.html#dom-input-files).

There is an associated [issue](https://github.com/whatwg/html/issues/3269) - it remains open, but I think only because of uncertainty around dispatching change events - it looks like the outcome was that events should not be fired when setting `.files` directly - I've left it out based on [this chromium issue](https://issues.chromium.org/issues/41359566) as well as firefox dispatching events [only when a flag is set](https://searchfox.org/mozilla-central/source/dom/html/HTMLInputElement.cpp#2520). 

This change does not affect more tests because they depend on copying files via a `DataTransfer` object, which has not been implemented in servo - so I didn't find any unexpected passes (I have a branch with `DataTransfer` implemented and expected the wpt to pass, but it failed due to this missing setter)

---

### DIFF
--- a/components/script/dom/htmlinputelement.rs
+++ b/components/script/dom/htmlinputelement.rs
@@ -1186,6 +1186,13 @@ impl HTMLInputElementMethods for HTMLInputElement {
         self.filelist.get().as_ref().cloned()
     }
 
+    /// <https://html.spec.whatwg.org/multipage/#dom-input-files>
+    fn SetFiles(&self, files: Option<&FileList>) {
+        if self.input_type() == InputType::File && files.is_some() {
+            self.filelist.set(files);
+        }
+    }
+
     // https://html.spec.whatwg.org/multipage/#dom-input-defaultchecked
     make_bool_getter!(DefaultChecked, "checked");
 
--- a/tests/wpt/meta/html/semantics/forms/the-input-element/files.html.ini
+++ b/tests/wpt/meta/html/semantics/forms/the-input-element/files.html.ini
@@ -1,7 +1,4 @@
 [files.html]
   type: testharness
-  [setting <input type=file>.files]
-    expected: FAIL
-
   [setting <input type=file>.files from DataTransfer]
     expected: FAIL
--- a/components/script/dom/webidls/HTMLInputElement.webidl
+++ b/components/script/dom/webidls/HTMLInputElement.webidl
@@ -23,7 +23,7 @@ interface HTMLInputElement : HTMLElement {
   [CEReactions]
            attribute boolean disabled;
   readonly attribute HTMLFormElement? form;
-  readonly attribute FileList? files;
+  attribute FileList? files;
   [CEReactions]
            attribute DOMString formAction;
   [CEReactions]
--- a/tests/wpt/meta-legacy-layout/html/dom/idlharness.https.html.ini
+++ b/tests/wpt/meta-legacy-layout/html/dom/idlharness.https.html.ini
@@ -3381,9 +3381,6 @@
   [HTMLSlotElement interface: operation assignedElements(AssignedNodesOptions)]
     expected: FAIL
 
-  [HTMLInputElement interface: attribute files]
-    expected: FAIL
-
   [Stringification of document.all]
     expected: FAIL
 
--- a/tests/wpt/meta-legacy-layout/html/semantics/forms/the-input-element/files.html.ini
+++ b/tests/wpt/meta-legacy-layout/html/semantics/forms/the-input-element/files.html.ini
@@ -1,7 +1,4 @@
 [files.html]
   type: testharness
-  [setting <input type=file>.files]
-    expected: FAIL
-
   [setting <input type=file>.files from DataTransfer]
     expected: FAIL
--- a/tests/wpt/meta/html/dom/idlharness.https.html.ini
+++ b/tests/wpt/meta/html/dom/idlharness.https.html.ini
@@ -3219,9 +3219,6 @@
   [HTMLInputElement interface: createInput("image") must inherit property "width" with the proper type]
     expected: FAIL
 
-  [HTMLInputElement interface: attribute files]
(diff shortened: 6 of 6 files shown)


components/script/dom/htmlinputelement.rs
```diff
@@ -1187,6 +1187,13 @@ impl HTMLInputElementMethods for HTMLInputElement {
         self.filelist.get().as_ref().map(|fl| fl.clone())
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-input-files
+    fn SetFiles(&self, files: Option<&FileList>) {
```
mrobinson, line 1191: Let's start using proper rustdoc for these functions:

```suggestion
    /// <https://html.spec.whatwg.org/multipage/#dom-input-files>
    fn SetFiles(&self, files: Option<&FileList>) {
```

mrobinson (changes requested):
Looks fine to me with one nit:

mrobinson (approved):

mrobinson:
It looks like this does adjust test results:

```
    OK /html/dom/idlharness.https.html?include=HTML.*
        PASS [expected FAIL] subtest: HTMLInputElement interface: attribute files
    OK /html/semantics/forms/the-input-element/files.html
        PASS [expected FAIL] subtest: setting &lt;input type=file&gt;.files
```

You'll need to update the expectations:

```
./mach test-wpt -r /html/dom/idlharness.https.html /html/semantics/forms/the-input-element/files.html --log-raw=out.log
./mach update-wpt out.log

./mach test-wpt -r /html/dom/idlharness.https.html /html/semantics/forms/the-input-element/files.html --log-raw=out.log --legacy-layout
./mach update-wpt out.log --legacy-layout
```

tests/wpt/meta/html/semantics/forms/the-input-element/files.html.ini
```diff
@@ -1,7 +1,5 @@
 [files.html]
   type: testharness
-  [setting <input type=file>.files]
-    expected: FAIL
```
mrobinson, line 4: There are two tests failing so I was expecting to see two test expectations files updated. In addition, you'll need to do the same set of updates for the legacy layout system (just pass `--legacy-layout` to all `mach` commands which should be visible above).

shanehandley: Ah apologies - I was getting timeouts when running those wpt commands earlier, so tried updating manually - but it seems to be working for me now.

mrobinson: Oh yeah, I have noticed these timeouts as well when running with the `--legacy-layout` flag. It would be nice to get to the bottom of it.

shanehandley:
I'm not knowledgable about the WPT runners - but let me know if you need me to make additional changes 👍 otherwise if this PR is problematic I can just close it 👍.

mrobinson:
Here's how you can update these tests:

1. Run the failing tests locally and save the output into a raw log:

```
$ ./mach test-wpt -r /html/dom/idlharness.https.html /html/semantics/forms/the-input-element/files.html --log-raw out.log
Running WPT tests with /home/martin/work/servo/target/release/servo
Running 4 tests in web-platform-tests

  ▶ OK /html/semantics/forms/the-input-element/files.html
  └   ▶ PASS [expected FAIL] setting <input type=file>.files

  ▶ OK /html/dom/idlharness.https.html?include=HTML.*
  │   ▶ PASS [expected FAIL] HTMLStyleElement interface: attribute disabled
  │   │ PASS [expected FAIL] HTMLStyleElement interface: document.createElement("style") must inherit property "disabled" with the proper type
  │   │ PASS [expected FAIL] HTMLPreElement interface: attribute width
  │   │ PASS [expected FAIL] HTMLPreElement interface: document.createElement("pre") must inherit property "width" with the proper type
  │   │ PASS [expected FAIL] HTMLPreElement interface: document.createElement("listing") must inherit property "width" with the proper type
  │   │ PASS [expected FAIL] HTMLPreElement interface: document.createElement("xmp") must inherit property "width" with the proper type
  └   └ PASS [expected FAIL] HTMLInputElement interface: attribute files

Ran 4 tests finished in 2.2 seconds.
  • 2 ran as expected. 0 tests skipped.
  • 2 tests had unexpected subtest results
```

2. Use the raw log to generate new results:

```
$  ./mach update-wpt out.log
```

Then for legacy layout you do the exact same two steps, but for each one you also pass `--legacy-layout`.

shanehandley:
Ah I'm very sorry Martin - I'm not sure why this happened, but when I first pushed, I had observed the tests passing for `/html/semantics/forms/the-input-element/files.html : setting &lt;input type=file&gt;.files` and just saw failures related to `DataTransfer` as per my PR comment.

When you initially asked me to update the test expectations though, after a branch switch and a rebuild, that test failed even though the code was identical. I assumed I had made a mistake, followed your instructions and pushed the updated test expectations with [Failure]s for that test.

But it seems now they are passing for you and in CI - I'm going to leave it be. But I have a better understanding of this process now - thank you.

mrobinson:
> When you initially asked me to update the test expectations though, after a branch switch and a rebuild, that test failed even though the code was identical. I assumed I had made a mistake, followed your instructions and pushed the updated test expectations with [Failure]s for that test.

No problem. Sometimes tests pass or fail locally, but do the opposite on the CI. It's just a matter of looking at the difference and using a bit of intuition to know what might be the issue. For instance, some tests inadvertently depend on fonts. Once you get the hang of updating test results, it becomes a lot easier though.

PR #29861, opened 2023-06-08
Revert "Update mozangle and cc."

mrobinson:
This reverts commit c2e90d963c389d27df7e19d9c9ef1218f191677d.

This is a temporarily revert in order to fix the build.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any error
- [x] These changes do not require tests because this is a build fix.

PR #31937, opened 2024-03-29
layout: Remove `LayoutRPC` and query layout via the `Layout` trait

mrobinson:
Instead of the tricky `LayoutRPC` interface, query layout using the
`Layout` trait. This means that now queries will requires calling layout
and then running the query. During layout an enum is used to indicate
what kind of layout is necessary.

This change also removes the mutex-locked `rw_data` from both layout
threads. It's no longer necessary since layout runs synchronously. The
one downside here is that for resolved style queries, we now have to
create two StyleContexts: one for layout and one for the query itself.
The creation of this context should not be very expensive though.

`LayoutRPC` used to be necessary because layout was run asynchronously
in a separate thread from script, but that no longer happens. With this
change, it becomes possible to safely pass nodes to layout from script
-- a cleanup that can happen in a followup change.

Fixes #29861.

---

- [x] These changes do not require tests because they should not change functionality.


jdm (approved):
Big fan of this simplification!

PR #31944, opened 2024-03-29
Clippy: add ' #Safety' section for unsafe functions

ektuu:
---

- [x] These changes are a part of https://github.com/servo/servo/issues/31500

- [x] These changes do not require tests because only docs were changed

### DIFF
--- a/components/layout/traversal.rs
+++ b/components/layout/traversal.rs
@@ -192,10 +192,11 @@ where
 ///
 /// # Safety
 ///
-/// This function is `unsafe` because it modifies the layout tree and context without any synchronization.
-///  The caller must ensure that:
-/// - There are no concurrent modifications to the layout tree.
-/// - The `LayoutContext` and `node` references are valid for the duration of the call.
+/// Marked `unsafe` because it directly modifies the layout tree and context without internal synchronization. Safe usage requires:
+/// - No concurrent access or modifications to the layout tree.
+/// - Valid `LayoutContext` and `node` references for the function's duration.
+///
+/// Non-compliance may lead to undefined behavior including data races and memory safety violations.
 pub unsafe fn construct_flows_at_ancestors<'dom>(
     context: &LayoutContext,
     mut node: impl LayoutNode<'dom>,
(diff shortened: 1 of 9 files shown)


ektuu:
I have added '#Safety' section for unsafe functions.
Please review and share your opinion on it.

mrobinson (changes requested):
If these have been generated with a large language model, we need to review them very closely to ensure that they make sense. I notice some things here that are kind of nonsense. For example for the function `fn to_layout(&self) -> LayoutDom<T>` the text reads "It is the caller's responsibility to ensure that the conversion is safe and that the resulting layout is correctly managed." This function doesn't generate a "layout" it generates a DOM node wrapped for use during layout. This is the sort of bogus documentation that one might expect to be generated by a large language model since it is using probabilities to generate language based on the text that is there "to layout" but the name doesn't capture what this function is actually doing.

ektuu:
> If these have been generated with a large language model, we need to review them very closely to ensure that they make sense. I notice some things here that are kind of nonsense. For example for the function `fn to_layout(&self) -> LayoutDom<T>` the text reads "It is the caller's responsibility to ensure that the conversion is safe and that the resulting layout is correctly managed." This function doesn't generate a "layout" it generates a DOM node wrapped for use during layout. This is the sort of bogus documentation that one might expect to be generated by a large language model since it is using probabilities to generate language based on the text that is there "to layout" but the name doesn't capture what this function is actually doing.

I apologize for the oversight in the documentation for fn to_layout(&self) -> LayoutDom<T>. Your feedback is invaluable, and I realize the description was inaccurate. I'll correct this immediately and ensure more accurate documentation moving forward.

ektuu:
hey! @mrobinson 
I have incorporated the suggested changes in this PR.
Waiting for your review:)

mrobinson (commented):
I'm going to close this one I think. I find it hard to trust documentation generated by a large language model for the Safety section of rustdoc. In addition, the new documentation replaces accurate documentation that is actually written by a human being. Thanks for the contribution, but I don't think we can accept this kind of LLM output right now. We are pending to have a discussion about how much AI-generated code should be accepted in the project. I think safety documentation is just too sensitive to even consider this though.

PR #31945, opened 2024-03-29
clippy: Fix several warnings in `components/script/dom/bindings`

azharcodeit:
Fixed warnings triggered by following checks:

- https://rust-lang.github.io/rust-clippy/master/index.html#from_raw_with_void_ptr
- https://rust-lang.github.io/rust-clippy/master/index.html#zero_ptr
- https://rust-lang.github.io/rust-clippy/master/index.html#useless_transmute
- https://rust-lang.github.io/rust-clippy/master/index.html#non_canonical_clone_impl

---

- [x] These changes do not require tests because they do not modify functionality

### DIFF
--- a/components/script/dom/bindings/buffer_source.rs
+++ b/components/script/dom/bindings/buffer_source.rs
@@ -418,7 +418,7 @@ where
     /// without causing conflicts , unexpected behavior.
     /// <https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/js/public/ArrayBuffer.h#L89>
     unsafe extern "C" fn free_func(_contents: *mut c_void, free_user_data: *mut c_void) {
-        let _ = Arc::from_raw(free_user_data as _);
+        let _ = Arc::from_raw(free_user_data as *const _);
     }
 
     unsafe {
--- a/components/script/dom/bindings/root.rs
+++ b/components/script/dom/bindings/root.rs
@@ -521,8 +521,12 @@ impl<T> Clone for Dom<T> {
 
 impl<T> Clone for LayoutDom<'_, T> {
     #[inline]
+    #[allow(clippy::non_canonical_clone_impl)]
     fn clone(&self) -> Self {
-        *self
+        assert_in_layout();
+        {
+            *self
+        }
     }
 }
 
--- a/components/script/dom/bindings/root.rs
+++ b/components/script/dom/bindings/root.rs
@@ -524,9 +524,7 @@ impl<T> Clone for LayoutDom<'_, T> {
     #[allow(clippy::non_canonical_clone_impl)]
     fn clone(&self) -> Self {
         assert_in_layout();
-        {
-            *self
-        }
+        *self
     }
 }
 
--- a/components/script/dom/bindings/root.rs
+++ b/components/script/dom/bindings/root.rs
@@ -275,7 +275,6 @@ impl RootCollection {
     unsafe fn unroot(&self, object: *const dyn JSTraceable) {
         assert_in_script();
         let roots = &mut *self.roots.get();
-        #[allow(clippy::vtable_address_comparisons)]
         match roots.iter().rposition(|r| std::ptr::eq(*r, object)) {
             Some(idx) => {
                 roots.remove(idx);
--- a/components/script/dom/bindings/root.rs
+++ b/components/script/dom/bindings/root.rs
@@ -275,6 +275,7 @@ impl RootCollection {
     unsafe fn unroot(&self, object: *const dyn JSTraceable) {
         assert_in_script();
         let roots = &mut *self.roots.get();
+        #[allow(clippy::vtable_address_comparisons)]
         match roots.iter().rposition(|r| std::ptr::eq(*r, object)) {
             Some(idx) => {
                 roots.remove(idx);
@@ -521,8 +522,7 @@ impl<T> Clone for Dom<T> {
 impl<T> Clone for LayoutDom<'_, T> {
     #[inline]
     fn clone(&self) -> Self {
-        assert_in_layout();
-        LayoutDom { value: self.value }
+        *self
     }
 }
 
--- a/components/script/dom/bindings/interface.rs
+++ b/components/script/dom/bindings/interface.rs
@@ -165,7 +165,7 @@ pub unsafe fn create_global_object(
     let private_val = PrivateValue(private);
     JS_SetReservedSlot(rval.get(), DOM_OBJECT_SLOT, &private_val);
     let proto_array: Box<ProtoOrIfaceArray> =
-        Box::new([0 as *mut JSObject; PrototypeList::PROTO_OR_IFACE_LENGTH]);
+        Box::new([ptr::null_mut::<JSObject>(); PrototypeList::PROTO_OR_IFACE_LENGTH]);
     let val = PrivateValue(Box::into_raw(proto_array) as *const libc::c_void);
     JS_SetReservedSlot(rval.get(), DOM_PROTOTYPE_SLOT, &val);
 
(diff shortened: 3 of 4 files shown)


components/script/dom/bindings/root.rs
```diff
@@ -521,8 +522,7 @@ impl<T> Clone for Dom<T> {
 impl<T> Clone for LayoutDom<'_, T> {
     #[inline]
     fn clone(&self) -> Self {
-        assert_in_layout();
-        LayoutDom { value: self.value }
+        *self
```
eerii, line 525: I think this change was talked about in https://github.com/servo/servo/pull/31894#discussion_r1541232929 but it wasn't applied in the end. In that discussion, @mrobinson proposed doing this instead to avoid removing the assert:

```suggestion
        assert_in_layout();
        { *self }
```

azharcodeit: Thanks, @eerii, very helpful! 

I double checked for this particular warning, it still persisted after proposed change. 

Since [assert_in_layout() needed to be kept](https://github.com/servo/servo/pull/31894#discussion_r1541232929), I decided to allow it. 

Considering those changes and formatting, I decided to create a new commit rather that committing your suggestion, if you don't mind :)

eerii: You're right, sorry for not realizing before! The error still persist because `Clone` has the assert but `Copy` doesn't, see https://github.com/servo/servo/pull/31894#discussion_r1541254417. Maybe the `Copy` trait should be removed like it says in the comment? Probably a good idea to wait to hear from a maintainer to see what to do, since this is a bit trickier than it seemed.

And no problem at all about the commit!

components/script/dom/bindings/root.rs
```diff
@@ -520,9 +521,12 @@ impl<T> Clone for Dom<T> {
 
 impl<T> Clone for LayoutDom<'_, T> {
     #[inline]
+    #[allow(clippy::non_canonical_clone_impl)]
     fn clone(&self) -> Self {
         assert_in_layout();
-        LayoutDom { value: self.value }
+        {
+            *self
+        }
```
Loirooriol, line 529: Is `{ *self }` any different than just `*self`?

azharcodeit: @Loirooriol, it seems like there's no difference between the two implementations. I will remove the unnecessary curly braces that were reapplied in the last commit.

components/script/dom/bindings/root.rs
mrobinson: It does sound like the appropriate thing to do here might be to remove the Copy trait implementation...but only if that's possible. Does Servo still compile with that removed?

components/script/dom/bindings/root.rs
```diff
@@ -275,6 +275,7 @@ impl RootCollection {
     unsafe fn unroot(&self, object: *const dyn JSTraceable) {
         assert_in_script();
         let roots = &mut *self.roots.get();
+        #[allow(clippy::vtable_address_comparisons)]
```
mrobinson, line 278: Based on https://github.com/rust-lang/rust-clippy/issues/6524, it seems that this lint will be removed in a new version of Rust. I don't think we want to hide this warning if it will be removed soon anyway. Maybe the thing to do here is to leave it for now and then see if upgrading our Rust version fixes it.

mrobinson (changes requested):

components/script/dom/bindings/buffer_source.rs
```diff
@@ -418,7 +418,7 @@ where
     /// without causing conflicts , unexpected behavior.
     /// <https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/js/public/ArrayBuffer.h#L89>
     unsafe extern "C" fn free_func(_contents: *mut c_void, free_user_data: *mut c_void) {
-        let _ = Arc::from_raw(free_user_data as _);
+        let _ = Arc::from_raw(free_user_data as *const _);
```
Loirooriol, line 421: What was the warning, why is `*const _` better than `_`?

components/script/dom/bindings/root.rs
azharcodeit: @mrobinson removing Copy trait implementation (`impl<T> Copy for LayoutDom<'_, T> {}`) caused following compile errors:
```
error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> components/script/layout_dom/document.rs:34:54
   |
23 |     document: LayoutDom<'dom, Document>,
   |     ----------------------------------- this field does not implement `std::marker::Copy`
...
34 | impl<'dom, LayoutDataType: LayoutDataTrait> Copy for ServoLayoutDocument<'dom, LayoutDataType> {}
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> components/script/layout_dom/element.rs:68:54
   |
55 |     element: LayoutDom<'dom, Element>,
   |     --------------------------------- this field does not implement `std::marker::Copy`
...
68 | impl<'dom, LayoutDataType: LayoutDataTrait> Copy for ServoLayoutElement<'dom, LayoutDataType> {}
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> components/script/layout_dom/node.rs:73:54
   |
51 |     pub(super) node: LayoutDom<'dom, Node>,
   |     -------------------------------------- this field does not implement `std::marker::Copy`
...
73 | impl<'dom, LayoutDataType: LayoutDataTrait> Copy for ServoLayoutNode<'dom, LayoutDataType> {}
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
  --> components/script/layout_dom/shadow_root.rs:30:54
   |
19 |     shadow_root: LayoutDom<'dom, ShadowRoot>,
   |     ---------------------------------------- this field does not implement `std::marker::Copy`
...
30 | impl<'dom, LayoutDataType: LayoutDataTrait> Copy for ServoShadowRoot<'dom, LayoutDataType> {}
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0204`.
error: could not compile `script` (lib) due to 4 previous errors

```

components/script/dom/bindings/buffer_source.rs
azharcodeit: I used *const because from_raw receives `*const `
```
pub unsafe fn from_raw(ptr: *const T) -> Self {
        unsafe { Arc::from_raw_in(ptr, Global) }
    }
```
the warning was:
```
warning: creating a `Arc` from a void raw pointer
   --> components/script/dom/bindings/buffer_source.rs:421:17
    |
421 |         let _ = Arc::from_raw(free_user_data as _);
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: cast this to a pointer of the appropriate type
   --> components/script/dom/bindings/buffer_source.rs:421:31
    |
421 |         let _ = Arc::from_raw(free_user_data as _);
    |                               ^^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#from_raw_with_void_ptr
    = note: `#[warn(clippy::from_raw_with_void_ptr)]` on by default
```

components/script/dom/bindings/root.rs
azharcodeit: @mrobinson makes sense! removed this change in the latest commit

mrobinson (approved):

Loirooriol (approved):

PR #31947, opened 2024-03-30
clippy: Fix `match_like_matches` warnings

eerii:
Fixes the remaining [`match_like_matches`](https://rust-lang.github.io/rust-clippy/master/index.html#/match_like_matches_macro) clippy warnings in `components/script`. Also updates comments near the changes to be proper rustdoc.

---
- [x] These changes do not require tests because they do not modify behaviour

### DIFF
--- a/components/script/dom/element.rs
+++ b/components/script/dom/element.rs
@@ -345,7 +345,7 @@ impl Element {
         self.is.borrow().clone()
     }
 
-    /// <https://dom.spec.whatwg.org/#concept-element-defined>
+    /// <https://dom.spec.whatwg.org/#concept-element-custom-element-state>
     pub fn set_custom_element_state(&self, state: CustomElementState) {
         // no need to inflate rare data for uncustomized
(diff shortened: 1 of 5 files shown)


components/script/dom/element.rs
```diff
@@ -345,16 +345,17 @@ impl Element {
         self.is.borrow().clone()
     }
 
+    /// <https://dom.spec.whatwg.org/#concept-element-defined>
```
mrobinson, line 348: https://dom.spec.whatwg.org/#concept-element-custom-element-state might be a better link here since `defined` is just one aspect of custom element state, but this method seems to address the general concept.

mrobinson, line 348:
```suggestion
    /// <https://dom.spec.whatwg.org/#concept-element-custom-element-state>
```

mrobinson (approved):

components/script/dom/element.rs
eerii: That makes sense, thank you for updating it!

PR #31949, opened 2024-03-30
refactor: CGClassConstructHook into Handwritten Helpers

### LINKED ISSUE #31571: Extract generated class constructor hook into handwritten functions
The output of CGClassConstructHook is [difficult to read](https://github.com/servo/servo/blob/main/components/script/dom/bindings/codegen/CodegenRust.py#L6174-L6217). By following the model of #31569 and creating two generic helper functions that cover both cases (isHTMLConstructor is true/false), it will be easier to read and modify this code in the future. 

Part of #31570.

rohansx:
- Extracted the constructor hook generated by CGClassConstructHook into two new, handwritten generic helper functions. This approach follows the model established in #31569.
- The new helper functions are designed to handle both cases of isHTMLConstructor being true or false, ensuring compatibility with the existing codebase while enhancing readability.
- Removed the Python script responsible for generating the previous, less readable Rust code, replacing it with our new, cleaner solution.

- [X] These changes fix #31571


components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::dom::bindings::root::DomRoot;
+use crate::jsapi::{JSContext, JSObject, CallArgs, GlobalScope, SafeJSContext};
+use crate::dom::bindings::import::module::callargs_is_constructing;
+use crate::dom::bindings::import::module::throw_constructor_without_new;
+use crate::dom::bindings::codegen::PrototypeList::ID;
+// Function for HTML constructor case
+pub unsafe fn construct_html_custom(global: &GlobalScope, cx: *mut JSContext, args: &CallArgs, descriptor_name: &str) -> bool {
+    let global = DomRoot::downcast::<dom::types::Global>(global).unwrap(); // Adjust type as needed
+    let prototype_id = match descriptor_name {
+        "HTMLDivElement" => PrototypeList::ID::HTMLDivElement,
```
jdm, line 15: Instead of taking string argument, we should take a PrototypeList::ID argument that we can use directly. The python code can construct the required value.

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+        &global,
+        prototype_id, // Adjust as needed
+        descriptor_name,
+    )
+}
+
+// Function for non-HTML constructor case
+pub unsafe fn construct_default_custom(global: &GlobalScope, cx: *mut JSContext, args: &CallArgs, descriptor_name: &str, ctor_name: &str) -> bool {
+    let prototype_id = match descriptor_name {
+        "HTMLDivElement" => PrototypeList::ID::HTMLDivElement,
```
jdm, line 32: Same comment about using a PrototypeList::ID argument instead of a string.

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+// Function for HTML constructor case
+pub unsafe fn construct_html_custom(global: &GlobalScope, cx: *mut JSContext, args: &CallArgs, descriptor_name: &str) -> bool {
+    let global = DomRoot::downcast::<dom::types::Global>(global).unwrap(); // Adjust type as needed
+    let prototype_id = match descriptor_name {
+        "HTMLDivElement" => PrototypeList::ID::HTMLDivElement,
+        "HTMLSpanElement" => PrototypeList::ID::HTMLSpanElement,
+        // Add cases for other descriptor names as needed
+        _ => panic!("Unsupported descriptor name: {}", descriptor_name),
+    };
+    htmlconstructor::call_html_constructor::<dom::types::Global>(
```
jdm, line 20: Instead of using `<dom::types::Global>` here, we need to add a generic type to this function (`pub unsafe fn construct_html_custom<T: DerivedFrom<Element> + DomObject>`) and use that generic type (`htmlconstructor::call_html_constructor::<T>`).

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::dom::bindings::htmlconstructor;
+use crate::dom::bindings::utils::{GetConstructorNameForReporting, MakeNativeName};
+use crate::dom::bindings::root::DomRoot;
+use crate::jsapi::{JSContext, JSObject, CallArgs, GlobalScope, SafeJSContext};
+use crate::dom::bindings::import::module::callargs_is_constructing;
+use crate::dom::bindings::import::module::throw_constructor_without_new;
+use crate::dom::bindings::codegen::PrototypeList::ID;
+// Function for HTML constructor case
+pub unsafe fn construct_html_custom(global: &GlobalScope, cx: *mut JSContext, args: &CallArgs, descriptor_name: &str) -> bool {
+    let global = DomRoot::downcast::<dom::types::Global>(global).unwrap(); // Adjust type as needed
```
jdm, line 13: I believe this needs to be a downcast to Window, not Global, for the call to call_html_constructor to succeed.

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+        "HTMLSpanElement" => PrototypeList::ID::HTMLSpanElement,
+        // Add cases for other descriptor names as needed
+        _ => panic!("Unsupported descriptor name: {}", descriptor_name),
+    };
+    htmlconstructor::call_html_constructor::<dom::types::Global>(
+        cx,
+        args,
+        &global,
+        prototype_id, // Adjust as needed
+        descriptor_name,
```
jdm, line 25: We're missing the [creator argument](https://doc.servo.org/script/dom/bindings/htmlconstructor/fn.call_html_constructor.html) here. We should accept that as an argument in `construct_html_custom` and let the python code pass the appropriate function value like it does before the changes in this PR.

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+        throw_constructor_without_new(cx, ctor_name);
+        return false;
+    }
+
+    rooted!(in(cx) let mut desired_proto = ptr::null_mut::<JSObject>());
+    let proto_result = get_desired_proto(
+        cx,
+        args,
+        prototype_id, // Adjust as needed
+        descriptor_name,
```
jdm, line 47: Same comment about the [creator argument](https://doc.servo.org/script/dom/bindings/interface/fn.get_desired_proto.html) here as earlier. This should be passed in by the python code.

components/script/dom/bindings/construct_hook.rs
```diff
@@ -0,0 +1,59 @@
...
+        desired_proto.handle_mut(),
+    );
+    assert!(proto_result.is_ok());
+    if proto_result.is_err() {
+        return false;
+    }
+
+    let nativeName = MakeNativeName(ctor_name);
```
jdm, line 55: MakeNativeName is a python function. This is going to be a bit tricky; I think we're going to need to leave the actual constructor call in the generated python code, because there's a lot of unique logic in there. What we _can_ do is something like this:
```rust
pub unsafe fn construct_default_custom(..., constructor: impl FnOnce(*mut JSContext, &CallArgs, &GlobalScope) -> bool) -> bool {
  ...
  constructor(cx, args, global)
}
```
Then modify the python code to call `f"construct_default_custom(&global, cx, &args, \"{ctorName}\", |cx, args, global| {{ {constructorInvoke} }});\n"` where constructorInvoke is the result of calling `.define()` on the constructorCall CGMethodCall value.

jdm (changes requested):
This is making good progress!

mrobinson:
@rohansx Do you plan to keep working on this one?

rohansx:
> @rohansx Do you plan to keep working on this one?

Yes.. I'll work on this issue @mrobinson

jdm:
Closing per #31571 .

PR #31950, opened 2024-03-30
Fixed some clippy warnings in components/script

komutumu:
This PR fixes some clippy warnings in components/script.

---
- [X] These changes do not require tests because they do not touch functionality


jahielkomu:
@mrobinson Hello sir.
Please kindly review and then revert

- I ask for more advice on the use of the ```// #Safety```  section. Should we just include it or try to add a few guidelines on why it is there.

- There is a clippy warning that talks about ```too many arguments``` the clippy documenation states that the upper limit for arguments is 7 and advises on separating them but how best can that be done without interfering with the code too much.

- There is a warning that talks about the use of is_none() after find() on an iterator

```
if node
           .iter()
           .find(|n| &***n as *const _ == &*node as *const _)
           .is_none()
``` 
In my attempts to fix it, I tried a lot of ways and the last one was 
```
if nodes
.iter()
.all(|n| &***n as *const _ != &*node as *const _)
```
I however, ran into a deferencing error at ```&***n``` I have researched and one of the workarounds was to use ```n``` and ```node``` directly  but I'm not sure about that. Please kindly advise me on how I can handle this.

Thank you

components/script/layout_dom/node.rs
```diff
@@ -97,7 +97,9 @@ impl<'dom, LayoutDataType: LayoutDataTrait> ServoLayoutNode<'dom, LayoutDataType
             phantom: PhantomData,
         }
     }
-
+    /// # Safety
+    ///
+    /// This function should not be called if there is no TrustedNodeAddress.
```
mrobinson, line 102: I don't think this is exactly right. Can you remove the addition of the `Safety` sections from this change? These need to be written carefully after studying the functions that they decorate.

mrobinson (changes requested):

mrobinson:
>     * There is a clippy warning that talks about `too many arguments` the clippy documenation states that the upper limit for arguments is 7 and advises on separating them but how best can that be done without interfering with the code too much.

I suspect the best way to deal with these functions is to decorate all of these functions with `#[allow(clippy::too_many_arguments)]`.

components/script/layout_dom/node.rs
jahielkomu: Okay sir

jahielkomu:
@mrobinson thank you for your guidance sir.

However, this error is persistent ```🛠 These changes could not be applied onto the latest upstream WPT. Servo's copy of the Web Platform Tests may be out of sync``` , I have pulled the latest changes locally but it persists... any suggestions sir?

mrobinson:
> @mrobinson thank you for your guidance sir.
> 
> However, this error is persistent `🛠 These changes could not be applied onto the latest upstream WPT. Servo's copy of the Web Platform Tests may be out of sync` , I have pulled the latest changes locally but it persists... any suggestions sir?

This is because you are using merge commits instead of rebasing your work and force pushing your branch. In general, using a rebase workflow will make for a cleaner commit history here and make it easier to keep track of all of you changes: https://medium.com/singlestone/a-git-workflow-using-rebase-1b1210de83e5.

mrobinson (approved):

jahielkomu:
> > @mrobinson thank you for your guidance sir.
> > However, this error is persistent `🛠 These changes could not be applied onto the latest upstream WPT. Servo's copy of the Web Platform Tests may be out of sync` , I have pulled the latest changes locally but it persists... any suggestions sir?
> 
> This is because you are using merge commits instead of rebasing your work and force pushing your branch. In general, using a rebase workflow will make for a cleaner commit history here and make it easier to keep track of all of you changes: https://medium.com/singlestone/a-git-workflow-using-rebase-1b1210de83e5.

Thank you sir for the resource. 
Let me understand it more for the next time.

PR #31952, opened 2024-03-31
Override timeout in WebGPU CTS to always be "long"

sagudev:
Try run: https://github.com/sagudev/servo/actions/runs/8496720116/job/23275056754 (as always `webgpu:shader,execution,expression` are flaky crashes).

---

- [x] These changes do not require tests because they are tests

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -741,7 +741,8 @@ tests/wpt/mozilla/tests for Servo-only tests""" % reference_path)
             filedata = file.read()
         # files are mounted differently
         filedata = filedata.replace('src=/webgpu/common/runtime/wpt.js', 'src=../webgpu/common/runtime/wpt.js')
-        # long timeout
+        # Mark all webgpu tests as long to increase their timeouts. This is needed due to wgpu's slowness.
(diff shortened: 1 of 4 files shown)


python/servo/testing_commands.py
```diff
@@ -741,6 +741,9 @@ def cts(self, repo="https://github.com/gpuweb/cts", checkout="main"):
             filedata = file.read()
         # files are mounted differently
         filedata = filedata.replace('src=/webgpu/common/runtime/wpt.js', 'src=../webgpu/common/runtime/wpt.js')
+        # long timeout
```
mrobinson, line 744: This comment hints at what happens here, but maybe it could give some motivation for why the timeout is increasing. I'm imagining something like"

```python
# Modify all tests to increase their timeouts. This is necessary because our WebGPU
# implementation is currently too slow or the CI hardware isn't powerful enough to
# run them in a reasonable time yet. See bug #12345.
```

Except the comment would explain the actual situation instead of the thing that I just made up. Without that it's hard for me to understand why this change is needed and I think that someone reading this code might be left with the same question.

mrobinson (changes requested):

mrobinson (approved):

PR #31953, opened 2024-03-31
Clippy: Fix the error of variants having the same prefix.

Aaryakhandelwal:
Renamed the variants having similar prefixes by using full paths to variants rather.

---

- [X] These changes do not require tests because they only fix clippy warnings and no functionality changes.

### DIFF
--- a/components/script/dom/htmlmediaelement.rs
+++ b/components/script/dom/htmlmediaelement.rs
@@ -417,7 +417,7 @@ impl HTMLMediaElement {
         Self {
             htmlelement: HTMLElement::new_inherited(tag_name, prefix, document),
             network_state: Cell::new(NetworkState::Empty),
-            ready_state: Cell::new(ReadyState::Nothing),
+            ready_state: Cell::new(ReadyState::HaveNothing),
             src_object: Default::default(),
             current_src: DomRefCell::new("".to_owned()),
             generation_id: Cell::new(0),
@@ -611,12 +611,12 @@ impl HTMLMediaElement {
 
         // Step 1.
         match (old_ready_state, ready_state) {
-            (ReadyState::Nothing, ReadyState::Metadata) => {
+            (ReadyState::HaveNothing, ReadyState::HaveMetadata) => {
                 task_source.queue_simple_event(self.upcast(), atom!("loadedmetadata"), &window);
                 // No other steps are applicable in this case.
                 return;
             },
-            (ReadyState::Metadata, new) if new >= ReadyState::CurrentData => {
+            (ReadyState::HaveMetadata, new) if new >= ReadyState::HaveCurrentData => {
                 if !self.fired_loadeddata_event.get() {
                     self.fired_loadeddata_event.set(true);
                     let this = Trusted::new(self);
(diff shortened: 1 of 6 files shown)


components/script/dom/htmlmediaelement.rs
```diff
@@ -405,19 +405,19 @@ pub enum NetworkState {
...
-    HaveNothing = HTMLMediaElementConstants::HAVE_NOTHING as u8,
-    HaveMetadata = HTMLMediaElementConstants::HAVE_METADATA as u8,
-    HaveCurrentData = HTMLMediaElementConstants::HAVE_CURRENT_DATA as u8,
-    HaveFutureData = HTMLMediaElementConstants::HAVE_FUTURE_DATA as u8,
-    HaveEnoughData = HTMLMediaElementConstants::HAVE_ENOUGH_DATA as u8,
+    Nothing = HTMLMediaElementConstants::HAVE_NOTHING as u8,
+    Metadata = HTMLMediaElementConstants::HAVE_METADATA as u8,
+    CurrentData = HTMLMediaElementConstants::HAVE_CURRENT_DATA as u8,
+    FutureData = HTMLMediaElementConstants::HAVE_FUTURE_DATA as u8,
+    EnoughData = HTMLMediaElementConstants::HAVE_ENOUGH_DATA as u8,
```
sagudev, line 412: I'm not sure about this one, because the names are from spec.

Aaryakhandelwal: Ohh right, I didn't notice that. I will make the changes

mrobinson: The build is now failing because you undid the rename, but didn't update the call sites. I would also add and `#![allow(...)]` directive at the head of this enum to silence the error with a comment like "# Clippy warning silenced here because these names are from the specification."

Aaryakhandelwal: Hi, I have made the suggested changes. The ./mach build --dev command is working in my local server but it is showing error in the PR.

sagudev: You need to run `./mach fmt` to fix formatting.

components/script/dom/htmlmediaelement.rs
```diff
@@ -404,6 +404,8 @@ pub enum NetworkState {
 /// <https://html.spec.whatwg.org/multipage/#dom-media-readystate>
 #[derive(Clone, Copy, Debug, JSTraceable, MallocSizeOf, PartialEq, PartialOrd)]
 #[repr(u8)]
+#[allow(clippy::enum_variant_names)]
+/// Clippy warning silenced here because these names are from the specification.
```
sagudev, line 408: This needs to be normal comment `//` not rustdoc comment `///`, then rustfmt should not complain.
```suggestion
#[allow(clippy::enum_variant_names)] // Clippy warning silenced here because these names are from the specification.
```

Aaryakhandelwal: Its alright, I just changed the formatting. It works now and no erros are detected by ./mach build --dev or ./mach test-tidy

sagudev: Rustdoc comments are used to write docs, while normal comments are used for anything else. As this is a comment that does not document stuff, but only explains why allow is needed it needs to be changed.

sagudev (approved):

PR #31963, opened 2024-04-01
Optimized code, by removing unused imports.

akwotom:
Optimized code, by removing unnecessary imports.


akwotom:
Improved code quality.

components/script/script_thread.rs
```diff
@@ -3566,11 +3566,13 @@ impl ScriptThread {
...
                 if !state_already_changed {
                     if let Some(target) = prev_mouse_over_target {
-                        if let Some(_) = target
+                        if  Some(
+                            target
                             .upcast::<Node>()
                             .inclusive_ancestors(ShadowIncluding::No)
                             .filter_map(DomRoot::downcast::<HTMLAnchorElement>)
                             .next()
+                        ).is_some()
```
fabricedesre, line 3575: This will always be true since the change tests `Some(...).is_some()` , right? If you want to get rid of of the `let` binding this should just be:

```rust
                   if target.upcast::<Node>()
                            .inclusive_ancestors(ShadowIncluding::No)
                            .filter_map(DomRoot::downcast::<HTMLAnchorElement>)
                            .next()
                            .is_some() {
                                  ....
                            }
```

akwotom: Wow, thank youuuu!

akwotom: I'll improve.

components/script/dom/bindings/callback.rs
```diff
@@ -7,6 +7,7 @@
 use std::default::Default;
 use std::ffi::CString;
 use std::mem::drop;
+// use std::ops::Deref; // Import is unused
```
sagudev, line 10: You should just remove it if not used.

akwotom: Okay, thank you.

mrobinson:
I rebased this locally and I think the issues in this PR have been addressed in other PRs.

PR #31966, opened 2024-04-01
layout: Add a `FontMetricsProvider` for resolving font-relative units

### LINKED ISSUE #14079: Style needs a way to get font metrics while computing style data.
This right now isn't implemented (see https://github.com/servo/servo/issues/7462), and I'm afraid it might be a pain to do for Stylo.

Basically, if we find an `ex` or a `ch` unit, we need to get the font metrics of the first available font for that style in order to compute the value.

This is not too common, but I think a lot of the font code is lazy in Gecko so calling it from multiple threads can be a very huge pain, and can impose a significant synchronization cost for pages that use this units.

In servo it's also a pain to implement too, since all the font code is on `gfx` and only layout has access to a `FontContext` right now.

cc @heycam @bholley @Manishearth @SimonSapin and @upsuper (I think I haven't left anyone :))

### LINKED ISSUE #31938: Stylo's Device should have access to font metrics
In Gecko, `Device` has a `document` member: https://searchfox.org/mozilla-central/rev/511d9c9e2eb0f2291dc2b2a5462ba054dfc2edfe/servo/components/style/gecko/media_queries.rs#37

From there is can get the `nsPresContext`, which is used to get the font metrics to resolve `line-height`:

 - https://searchfox.org/mozilla-central/rev/511d9c9e2eb0f2291dc2b2a5462ba054dfc2edfe/servo/components/style/gecko/media_queries.rs#125-143
- https://searchfox.org/mozilla-central/rev/511d9c9e2eb0f2291dc2b2a5462ba054dfc2edfe/layout/generic/ReflowInput.cpp#2789-2794

Servo should do something similar in order to have proper support for `lh` units: https://bugzilla.mozilla.org/show_bug.cgi?id=1310170

I think layout2020 gets the font metrics like this:

```
            crate::context::with_thread_local_font_context(layout_context, |font_context| {
                get_font_for_first_font_for_style(style, font_context)
                    .map(|font| font.borrow().metrics.clone())
            });
```

and legacy like

```
pub fn font_metrics_for_style(
    font_context: &mut LayoutFontContext,
    style: crate::ServoArc<FontStyleStruct>,
) -> FontMetrics {
    let font_group = font_context.font_group(style);
    let font = font_group.borrow_mut().first(font_context);
    let font = font.as_ref().unwrap().borrow();

    font.metrics.clone()
}
```

where the `LayoutFontContext` comes from something like

```
                with_thread_local_font_context(self.layout_context, |font_context| {
                    inline_flow
                        .minimum_line_metrics(font_context, &node.style(self.style_context()))
                });
```

So it seems we need to somehow expose the `LayoutContext` of both 2020 and 2013 (they are different) to `Device`.

mrobinson:
The only font relative unit that Servo knows how to resolve currently is
`rem` (relative to the root font size). This is because Stylo cannot do
any font queries. This adds a mechanism to allow this, exposing the
ability to properly render `ex` units in Servo.

This change only allows resolving some font size relative units thoug,
as Servo doesn't collect all the FontMetrics it needs to resolve them
all. This capability will be added in followup changes.

Some new tests fail:
 - ex-unit-001.html: This test fails because Servo does not yet have
   support for setting the weight using @font-face rules on web fonts.
 - ex-unit-004.html: This test fails because Servo does not yet have
   support for setting the Unicode range of a web font using @font-face
   rules.
 - first-available-font-001.html: This test fails because the above
   two feature are missing.

Fixes #31938. Fixes #14079.

---

- [x] These changes fix #31938 and fix #14079.

### DIFF
--- a/components/gfx/font_context.rs
+++ b/components/gfx/font_context.rs
@@ -93,12 +93,20 @@ impl<S: FontSource> FontContext<S> {
     /// Font groups are cached, so subsequent calls with the same `style` will return a reference
     /// to an existing `FontGroup`.
     pub fn font_group(&mut self, style: Arc<FontStyleStruct>) -> Rc<RefCell<FontGroup>> {
+        let font_size = style.font_size.computed_size().into();
+        self.font_group_with_size(style, font_size)
+    }
+
+    /// Like [`Self::font_group`], but overriding the size found in the [`FontStyleStruct`] with the given size
+    /// in pixels.
+    pub fn font_group_with_size(
+        &mut self,
+        style: Arc<FontStyleStruct>,
+        size: Au,
+    ) -> Rc<RefCell<FontGroup>> {
         self.expire_font_caches_if_necessary();
 
-        let cache_key = FontGroupCacheKey {
-            size: Au::from_f32_px(style.font_size.computed_size().px()),
-            style,
-        };
+        let cache_key = FontGroupCacheKey { size, style };
 
         if let Some(font_group) = self.font_group_cache.get(&cache_key) {
             return font_group.clone();
--- a/components/layout_thread_2020/lib.rs
+++ b/components/layout_thread_2020/lib.rs
@@ -43,7 +43,7 @@ use malloc_size_of::{MallocSizeOf, MallocSizeOfOps};
 use metrics::{PaintTimeMetrics, ProfilerMetadataFactory};
 use msg::constellation_msg::{BrowsingContextId, PipelineId};
 use net_traits::image_cache::{ImageCache, UsePlaceholder};
-use parking_lot::RwLock;
+use parking_lot::{ReentrantMutex, RwLock};
 use profile_traits::mem::{Report, ReportKind, ReportsChan};
 use profile_traits::path;
 use profile_traits::time::{
@@ -70,14 +70,15 @@ use style::context::{
     QuirksMode, RegisteredSpeculativePainter, RegisteredSpeculativePainters, SharedStyleContext,
 };
 use style::dom::{OpaqueNode, TElement, TNode};
-use style::driver;
 use style::error_reporting::RustLogReporter;
+use style::font_metrics::FontMetrics;
 use style::global_style_data::{GLOBAL_STYLE_DATA, STYLE_THREAD_POOL};
 use style::invalidation::element::restyle_hints::RestyleHint;
 use style::media_queries::{Device, MediaList, MediaType};
 use style::properties::style_structs::Font;
 use style::properties::PropertyId;
 use style::selector_parser::{PseudoElement, SnapshotMap};
+use style::servo::media_queries::FontMetricsProvider;
 use style::shared_lock::{SharedRwLock, SharedRwLockReadGuard, StylesheetGuards};
 use style::stylesheets::{
     DocumentStyleSheet, Origin, Stylesheet, StylesheetInDocument, UrlExtraData,
@@ -86,6 +87,8 @@ use style::stylesheets::{
 use style::stylist::Stylist;
 use style::traversal::DomTraversal;
 use style::traversal_flags::TraversalFlags;
+use style::values::computed::CSSPixelLength;
+use style::{driver, Zero};
 use style_traits::{CSSPixel, DevicePixel, SpeculativePainter};
 use url::Url;
 use webrender_api::units::LayoutPixel;
@@ -120,8 +123,9 @@ pub struct LayoutThread {
     /// Reference to the script thread image cache.
     image_cache: Arc<dyn ImageCache>,
 
-    /// Public interface to the font cache thread.
-    font_cache_thread: FontCacheThread,
+    /// Public interface to the font cache thread. This needs to be behind a [`ReentrantMutex`],
+    /// because some font cache operations can trigger others.
+    font_cache_thread: Arc<ReentrantMutex<FontCacheThread>>,
 
     /// Is this the first reflow in this LayoutThread?
     first_reflow: Cell<bool>,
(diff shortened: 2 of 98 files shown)


components/layout_thread_2020/lib.rs
```diff
@@ -1258,3 +1271,46 @@ impl RegisteredSpeculativePainters for RegisteredPaintersImpl {
...
+            let Some(servo_metrics) = font_context
+                .font_group_with_size(ServoArc::new(font.clone()), base_size.px())
+                .borrow_mut()
+                .first(font_context)
+                .map(|font| font.borrow().metrics.clone())
+            else {
+                return Default::default();
+            };
+
+            // Only use the x height of this font if it is non-zero. Some fonts return
```
Loirooriol, line 1297:
```suggestion
            // Only use the x-height of this font if it is non-zero. Some fonts return
```

components/layout_thread_2020/lib.rs
```diff
@@ -1258,3 +1271,46 @@ impl RegisteredSpeculativePainters for RegisteredPaintersImpl {
...
+            // inaccurate metrics, which shouldn't be used.
+            let x_height = if !servo_metrics.x_height.is_zero() {
+                Some(CSSPixelLength::new(servo_metrics.x_height.to_f32_px()))
+            } else {
+                None
+            };
```
Loirooriol, line 1303:
```suggestion
            let x_height = Some(servo_metrics.x_height)
                .filter(|x_height| !x_height.is_zero())
                .map(CSSPixelLength::from);
```

components/layout_thread_2020/lib.rs
```diff
@@ -1258,3 +1271,46 @@ impl RegisteredSpeculativePainters for RegisteredPaintersImpl {
...
+
+            FontMetrics {
+                x_height,
+                zero_advance_measure: None,
+                cap_height: None,
+                ic_width: None,
+                ascent: CSSPixelLength::new(servo_metrics.ascent.to_f32_px()),
```
Loirooriol, line 1310:
```suggestion
                ascent: servo_metrics.ascent.into(),
```

components/script/dom/document.rs
```diff
@@ -3473,20 +3473,14 @@ impl Document {
         have_changed
     }
 
-    /// Returns a `Device` suitable for media query evaluation.
+    /// Run the given closure using the Stylo `Device` suitable for media query evaluation.
```
Loirooriol, line 3476: I think 3rd person singular is typically used:

```suggestion
    /// Runs the given closure using the Stylo `Device` suitable for media query evaluation.
```

tests/unit/style/custom_properties.rs
```diff
@@ -9,13 +9,33 @@ use style::context::QuirksMode;
...
 };
+use style::font_metrics::FontMetrics;
 use style::media_queries::{Device, MediaType};
+use style::properties::style_structs::Font;
 use style::properties::{CustomDeclaration, CustomDeclarationValue};
 use style::rule_tree::CascadeLevel;
+use style::servo::media_queries::FontMetricsProvider;
 use style::stylesheets::layer_rule::LayerOrder;
 use style::stylist::Stylist;
+use style::values::computed::CSSPixelLength;
```
Loirooriol, line 20: `Length` is shorter (they are synonyms).

tests/unit/style/stylist.rs
```diff
@@ -8,18 +8,38 @@ use selectors::parser::{AncestorHashes, Selector};
...
 use style::selector_map::SelectorMap;
 use style::selector_parser::{SelectorImpl, SelectorParser};
+use style::servo::media_queries::FontMetricsProvider;
 use style::shared_lock::SharedRwLock;
 use style::stylesheets::StyleRule;
 use style::stylist::{
     needs_revalidation_for_testing, ContainerConditionId, LayerId, Rule, Stylist,
 };
 use style::thread_state::{self, ThreadState};
+use style::values::computed::CSSPixelLength;
```
Loirooriol, line 24: Ditto

components/gfx/font_context.rs
```diff
@@ -93,10 +93,21 @@ impl<S: FontSource> FontContext<S> {
...
+        let font_size = style.font_size.computed_size().px();
+        self.font_group_with_size(style, font_size)
+    }
+
+    /// Like [`Self::font_group`], but overriding the size found in the [`FontStyleStruct`] with the given size
+    /// in pixels.
+    pub fn font_group_with_size(
+        &mut self,
+        style: Arc<FontStyleStruct>,
+        size_in_px: f32,
```
Loirooriol, line 105: This can be a Length, to avoid the `.px()` for both callers.

components/layout_thread_2020/lib.rs
mrobinson: There's definitely a tradeoff here between being terse and readable, but I think it isn't a big deal. I've made this change.

components/layout_thread_2020/lib.rs
mrobinson: Nice catch. Fixed.

components/script/dom/document.rs
mrobinson: Sure. Done.

tests/unit/style/custom_properties.rs
mrobinson: Makes sense. I always forget if `Length` is in stylo or layout. Fixed.

tests/unit/style/stylist.rs
mrobinson: Fixed.

mrobinson:
@Loirooriol Thanks for the review. I believe I've addressed all of your comments.

Loirooriol (approved):
Thanks, so many tests passing :tada:

mrobinson:
Oops. Obviously I need to merge the change in Stylo before this one. 🤦

PR #31962, opened 2024-04-01
clippy: Mark functions taking many arguments as exceptions to clippy lints

komutumu:
Mark these functions as allowed for clippy lints. The idea is that they can be
fixed iteratively, but fixing them all now is too much work.

---
- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes are part of #31500 
- [X] These changes do not require tests because they don't touch functionality.

PR #31974, opened 2024-04-02
clippy: Allow `too_many_arguments` for existing functions

eerii:
Followup to #31962, adds allows for `clippy::too_many_arguments` for existing functions. Most of the new allows are for constructor functions (`new*`), there has been discussions [here](https://github.com/rust-lang/rust-clippy/issues/1576) on whether to make an exception for them, since it makes sense for them to have as many arguments as the struct they are creating. However, adding an explicit allow is still recommended for now.

It also changes the few global allows that were made in some files for function specific ones. This is done to get warnings if new functions with too many arguments are added in the future. Finally, it also changes the surrounding comments to proper rustdoc where necessary.

---
- [x] These changes do not require tests because they do not modify behaviour

### DIFF
--- a/components/script/canvas_state.rs
+++ b/components/script/canvas_state.rs
@@ -333,6 +333,7 @@ impl CanvasState {
     ///
     /// drawImage coordinates explained
     ///
+    /// ```
     ///  Source Image      Destination Canvas
     /// +-------------+     +-------------+
     /// |             |     |             |
@@ -345,7 +346,7 @@ impl CanvasState {
     /// |     sw      |     |     dw      |
     /// |             |     |             |
     /// +-------------+     +-------------+
-    ///
+    /// ```
     ///
     /// The rectangle (sx, sy, sw, sh) from the source image
     /// is copied on the rectangle (dx, dy, dh, dw) of the destination canvas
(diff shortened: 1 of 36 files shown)


components/script/canvas_state.rs
```diff
@@ -330,27 +330,27 @@ impl CanvasState {
...
+    /// |             |     |             |
+    /// |(sx,sy)      |     |(dx,dy)      |
+    /// |   +----+    |     |   +----+    |
+    /// |   |    |    |     |   |    |    |
+    /// |   |    |sh  |---->|   |    |dh  |
+    /// |   |    |    |     |   |    |    |
+    /// |   +----+    |     |   +----+    |
+    /// |     sw      |     |     dw      |
+    /// |             |     |             |
+    /// +-------------+     +-------------+
```
mrobinson, line 347: You'll probably want to surround the ASCII art here with three backticks to denote a code block, otherwise it might be displayed oddly in IDEs and the generated documentation.

mrobinson (changes requested):
Thanks so much for this wonderful cleanup. There's just one small nit I have below:

components/script/canvas_state.rs
eerii: Thank you for the catch!

mrobinson (approved):

PR #31979, opened 2024-04-02
script: Reduce the use of `unsafe` in LayoutDom

mrobinson:
Remove the use of unsafe code in the layout wrappers of the DOM. The
main change here is that `unsafe_get()` no longer needs to be an unsafe
method, which allows us to transitively remove or reduce unsafe blocks
from callers. The function itself is not renamed, because it's still
a bit dangerous to start removing the layers of abstraction from actual
DOM nodes.

In addition `init_style_and_opaque_layout_data` can be merged into
`initialize_data`, which removes one more unsafe method.

Finally, a "Safety" section is added to some unsafe methods.

---

- [x] These changes do not require tests because they do not modify functionality.

### DIFF
--- a/components/script/dom/canvasrenderingcontext2d.rs
+++ b/components/script/dom/canvasrenderingcontext2d.rs
@@ -127,23 +127,20 @@ impl CanvasRenderingContext2D {
 }
 
 pub trait LayoutCanvasRenderingContext2DHelpers {
-    #[allow(unsafe_code)]
-    unsafe fn get_ipc_renderer(self) -> IpcSender<CanvasMsg>;
+    fn get_ipc_renderer(self) -> IpcSender<CanvasMsg>;
     fn get_canvas_id(self) -> CanvasId;
 }
 
 impl LayoutCanvasRenderingContext2DHelpers for LayoutDom<'_, CanvasRenderingContext2D> {
-    #[allow(unsafe_code)]
-    unsafe fn get_ipc_renderer(self) -> IpcSender<CanvasMsg> {
+    fn get_ipc_renderer(self) -> IpcSender<CanvasMsg> {
         (self.unsafe_get()).canvas_state.get_ipc_renderer().clone()
     }
 
-    #[allow(unsafe_code)]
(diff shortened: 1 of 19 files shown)


components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -128,13 +128,13 @@ impl CanvasRenderingContext2D {
 
 pub trait LayoutCanvasRenderingContext2DHelpers {
     #[allow(unsafe_code)]
```
jdm, line 130:
```suggestion
```

components/script/dom/canvasrenderingcontext2d.rs
```diff
@@ -128,13 +128,13 @@ impl CanvasRenderingContext2D {
...
-    unsafe fn get_ipc_renderer(self) -> IpcSender<CanvasMsg>;
+    fn get_ipc_renderer(self) -> IpcSender<CanvasMsg>;
     fn get_canvas_id(self) -> CanvasId;
 }
 
 impl LayoutCanvasRenderingContext2DHelpers for LayoutDom<'_, CanvasRenderingContext2D> {
     #[allow(unsafe_code)]
```
jdm, line 136:
```suggestion
```

components/script/dom/htmliframeelement.rs
```diff
@@ -515,15 +515,14 @@ pub trait HTMLIFrameElementLayoutMethods {
...
 impl HTMLIFrameElementLayoutMethods for LayoutDom<'_, HTMLIFrameElement> {
     #[inline]
-    #[allow(unsafe_code)]
     fn pipeline_id(self) -> Option<PipelineId> {
-        unsafe { (self.unsafe_get()).pipeline_id.get() }
+        (self.unsafe_get()).pipeline_id.get()
     }
 
     #[inline]
     #[allow(unsafe_code)]
```
jdm, line 523:
```suggestion
```

jdm (approved):
This cleanup pleases me!

mrobinson:
Thanks for the review!

PR #31985, opened 2024-04-03
script: Split style and layout data in DOM nodes

mrobinson:
This change splits the style and layout data in DOM nodes that is
populated by style and layout passes. This makes Servo's data design
more like Gecko's. This allows:

1. Removing the various `StyleAndLayout` data structures used by layout.
2. Removing the `GetStyleAndLayoutData` and
   `GetStyleAndOpaqueLayoutData` traits. Accessing style and layout data
   are now just functions on the `LayoutNode` and `ThreadSafeLayoutNode`
   traits.
3. Styling now doesn't populate layout data. This is is postponed until
   layout itself.
4. Allows the DOM wrappers to no longer have to be generic over the
   layout data. This data was already stored using `std::any::Any` and
   the new code just makes layout responsible for downcasting. Cleaning
   up the generic type parameter in the DOM wrappers can happen in a
   followup change.

The main benefit to all of this is that we should be able to remove
unsafe creation of `ServoLayoutNode` in layout and
`TrustedLayoutNodeAddress` entirely, because `ServoLayoutNode` will be
able to be passed directly from script to layout. In addition, this
removes one more abstraction layer from the layout DOM wrappers, making
the code a lot more understandable.

Note: This increases the measured size of DOM types, but the same data
is stored. It's simply that before that data was stored behind a heap
pointer.

---

### DIFF
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -26,8 +26,8 @@ use msg::constellation_msg::{BrowsingContextId, PipelineId};
 use net_traits::image::base::{Image, ImageMetadata};
 use script_layout_interface::message::QueryMsg;
 use script_layout_interface::{
-    HTMLCanvasData, HTMLMediaData, LayoutElementType, LayoutNodeType, SVGSVGData,
-    StyleAndOpaqueLayoutData, TrustedNodeAddress,
+    GenericLayoutData, HTMLCanvasData, HTMLMediaData, LayoutElementType, LayoutNodeType,
+    SVGSVGData, StyleData, TrustedNodeAddress,
 };
 use script_traits::{DocumentActivity, UntrustedNodeAddress};
 use selectors::matching::{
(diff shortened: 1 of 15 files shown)


components/script/dom/node.rs
```diff
@@ -155,10 +155,16 @@ pub struct Node {
     /// are this node.
     ranges: WeakRangeVec,
 
-    /// Style+Layout information.
+    /// Style data for this node. This is accessed and mutated by style
+    /// passes and is used to lay out this node and populate layout data.
+    #[no_trace]
+    style_data: DomRefCell<Option<Box<StyleData>>>,
+
+    /// Layout data for this node. This is populated during layout and is can be used for
```
jdm, line 163:
```suggestion
    /// Layout data for this node. This is populated during layout and can be used for
```

jdm (approved):
This is a great cleanup!

PR #31986, opened 2024-04-03
clippy: Fix `vec_box` warnings in `components/script`

oluwatobiss:
Remove the unnecessary boxing of `Vec`'s content to fix the `vec_box` warnings.

**ref:** https://rust-lang.github.io/rust-clippy/master/index.html#vec_box

---

- [X] These changes do not require tests because they only resolve clippy warnings. They do not change functionalities.

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -294,7 +294,9 @@ pub struct GlobalScope {
     ///
     /// <https://html.spec.whatwg.org/multipage/#about-to-be-notified-rejected-promises-list>
     #[ignore_malloc_size_of = "mozjs"]
-    uncaught_rejections: DomRefCell<Vec<Heap<*mut JSObject>>>,
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
+    uncaught_rejections: DomRefCell<Vec<Box<Heap<*mut JSObject>>>>,
 
     /// Promises in this list have previously been reported as rejected
     /// (because they were in the above list), but the rejection was handled
@@ -302,7 +304,9 @@ pub struct GlobalScope {
     ///
     /// <https://html.spec.whatwg.org/multipage/#outstanding-rejected-promises-weak-set>
     #[ignore_malloc_size_of = "mozjs"]
-    consumed_rejections: DomRefCell<Vec<Heap<*mut JSObject>>>,
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
+    consumed_rejections: DomRefCell<Vec<Box<Heap<*mut JSObject>>>>,
 
     /// True if headless mode.
     is_headless: bool,
@@ -2180,7 +2184,7 @@ impl GlobalScope {
     pub fn add_uncaught_rejection(&self, rejection: HandleObject) {
         self.uncaught_rejections
             .borrow_mut()
-            .push(*Heap::boxed(rejection.get()));
+            .push(Heap::boxed(rejection.get()));
     }
 
     pub fn remove_uncaught_rejection(&self, rejection: HandleObject) {
@@ -2188,20 +2192,22 @@ impl GlobalScope {
 
         if let Some(index) = uncaught_rejections
             .iter()
-            .position(|promise| *promise == *Heap::boxed(rejection.get()))
+            .position(|promise| *promise == Heap::boxed(rejection.get()))
         {
             uncaught_rejections.remove(index);
         }
     }
 
-    pub fn get_uncaught_rejections(&self) -> &DomRefCell<Vec<Heap<*mut JSObject>>> {
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
+    pub fn get_uncaught_rejections(&self) -> &DomRefCell<Vec<Box<Heap<*mut JSObject>>>> {
         &self.uncaught_rejections
     }
 
     pub fn add_consumed_rejection(&self, rejection: HandleObject) {
         self.consumed_rejections
             .borrow_mut()
-            .push(*Heap::boxed(rejection.get()));
+            .push(Heap::boxed(rejection.get()));
     }
 
     pub fn remove_consumed_rejection(&self, rejection: HandleObject) {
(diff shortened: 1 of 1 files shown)


mrobinson (changes requested):

components/script/dom/globalscope.rs
```diff
@@ -294,15 +294,19 @@ pub struct GlobalScope {
...
 
     /// Promises in this list have previously been reported as rejected
     /// (because they were in the above list), but the rejection was handled
     /// in the last turn of the event loop.
     ///
     /// <https://html.spec.whatwg.org/multipage/#outstanding-rejected-promises-weak-set>
     #[ignore_malloc_size_of = "mozjs"]
-    consumed_rejections: DomRefCell<Vec<Heap<*mut JSObject>>>,
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
```
mrobinson, line 308: I think it makes sense to put this above the `allow` directive here and everywhere.

components/script/dom/globalscope.rs
```diff
@@ -2180,42 +2184,46 @@ impl GlobalScope {
     pub fn add_uncaught_rejection(&self, rejection: HandleObject) {
         self.uncaught_rejections
             .borrow_mut()
-            .push(*Heap::boxed(rejection.get()));
+            .push(Heap::boxed(rejection.get()));
```
mrobinson, line 2187: Are these changes still necessary?

components/script/dom/globalscope.rs
```diff
@@ -2180,42 +2184,46 @@ impl GlobalScope {
...
+            .position(|promise| *promise == Heap::boxed(rejection.get()))
         {
             uncaught_rejections.remove(index);
         }
     }
 
-    pub fn get_uncaught_rejections(&self) -> &DomRefCell<Vec<Heap<*mut JSObject>>> {
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
+    pub fn get_uncaught_rejections(&self) -> &DomRefCell<Vec<Box<Heap<*mut JSObject>>>> {
```
mrobinson, line 2203: Please move the comment above the code it comments on.

components/script/dom/globalscope.rs
```diff
@@ -2180,42 +2184,46 @@ impl GlobalScope {
...
         &self.uncaught_rejections
     }
 
     pub fn add_consumed_rejection(&self, rejection: HandleObject) {
         self.consumed_rejections
             .borrow_mut()
-            .push(*Heap::boxed(rejection.get()));
+            .push(Heap::boxed(rejection.get()));
```
mrobinson, line 2210: Can this be reverted?

components/script/dom/globalscope.rs
```diff
@@ -2180,42 +2184,46 @@ impl GlobalScope {
...
+            .position(|promise| *promise == Heap::boxed(rejection.get()))
         {
             consumed_rejections.remove(index);
         }
     }
 
-    pub fn get_consumed_rejections(&self) -> &DomRefCell<Vec<Heap<*mut JSObject>>> {
+    #[allow(clippy::vec_box)]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
+    pub fn get_consumed_rejections(&self) -> &DomRefCell<Vec<Box<Heap<*mut JSObject>>>> {
```
mrobinson, line 2226: Please move the comment above the code it comments on.

mrobinson (changes requested):

components/script/dom/globalscope.rs
oluwatobiss: Absolutely.

mrobinson: @sagudev ^^

sagudev: Am not sure I understand question. But given that we store Box<Heap<>> (and we must) we also need ` .push(Heap::boxed(rejection.get()))` (this is what we have now actually).

oluwatobiss: @sagudev To rephrase, since we are using the `allow` directive to permit a `<Vec<Box<Heap<*mut JSObject>>>>` syntax, would a `.push(Heap::boxed(rejection.get()))` be the correct code?

Or do we need to include a `*` (like so: `.push(*Heap::boxed(rejection.get()))`)?

mrobinson: Apologies, this looks fine. I was looking at the latest commit when I should have been looking at the diff from main. Please just move the comments. Otherwise this looks good.

components/script/dom/globalscope.rs
```diff
@@ -294,6 +294,8 @@ pub struct GlobalScope {
     ///
     /// <https://html.spec.whatwg.org/multipage/#about-to-be-notified-rejected-promises-list>
     #[ignore_malloc_size_of = "mozjs"]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
```
mrobinson, line 297:
```suggestion
    // `Heap` values must stay boxed, as they need semantics like `Pin`
    // (that is, they cannot be moved).
```

components/script/dom/globalscope.rs
```diff
@@ -302,6 +304,8 @@ pub struct GlobalScope {
     ///
     /// <https://html.spec.whatwg.org/multipage/#outstanding-rejected-promises-weak-set>
     #[ignore_malloc_size_of = "mozjs"]
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
```
mrobinson, line 307:
```suggestion
    // `Heap` values must stay boxed, as they need semantics like `Pin`
    // (that is, they cannot be moved).
```

components/script/dom/globalscope.rs
```diff
@@ -2194,6 +2198,8 @@ impl GlobalScope {
         }
     }
 
+    // `Heap` values must stay boxed, as they need semantics like `Pin` ie they cannot be moved
```
mrobinson, line 2201:
```suggestion
    // `Heap` values must stay boxed, as they need semantics like `Pin`
    // (that is, they cannot be moved).
```

components/script/dom/globalscope.rs
```diff
@@ -2215,6 +2221,8 @@ impl GlobalScope {
...
```
mrobinson, line 2224:
```suggestion
    // `Heap` values must stay boxed, as they need semantics like `Pin`
    // (that is, they cannot be moved).
```

mrobinson (approved):

PR #31989, opened 2024-04-03
Fixed some clippy warning by adding default implementations

komutumu:
This PR fixes clippy warnings that require creation of default implementations of structs or types by creating the default implementations.

---

### DIFF
--- a/components/script/dom/range.rs
+++ b/components/script/dom/range.rs
@@ -1083,14 +1083,17 @@ pub struct WeakRangeVec {
 }
 
 #[allow(unsafe_code)]
-impl WeakRangeVec {
+impl Default for WeakRangeVec {
     /// Create a new vector of weak references.
-    pub fn new() -> Self {
+    fn default() -> Self {
         WeakRangeVec {
             cell: UnsafeCell::new(WeakRefVec::new()),
         }
     }
+}
 
+#[allow(unsafe_code)]
+impl WeakRangeVec {
     /// Whether that vector of ranges is empty.
     pub fn is_empty(&self) -> bool {
         unsafe { (*self.cell.get()).is_empty() }
--- a/components/script/dom/range.rs
+++ b/components/script/dom/range.rs
@@ -1082,9 +1082,7 @@ pub struct WeakRangeVec {
     cell: UnsafeCell<WeakRefVec<Range>>,
 }
 
-#[allow(unsafe_code)]
 impl Default for WeakRangeVec {
-    /// Create a new vector of weak references.
     fn default() -> Self {
         WeakRangeVec {
             cell: UnsafeCell::new(WeakRefVec::new()),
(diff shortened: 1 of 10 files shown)


jahielkomu:
Hello @mrobinson I hope you are well.
Please kindly review and revert.


In the meantime there is one struct that requires a default implementation, that is, 
```
impl Documents {
    pub fn new() -> Documents {
        Documents {
            map: HashMapTracedValues::new(),
        }
    }
 ...
}
```
However when I did as,
```
impl Default for Documents {
     fn default() -> Self{
        Documents {
            map: HashMapTracedValues::new(),
        }
    }
}
```
``` Self ( ... ```  triggered an error to an unrooted type and I realise the struct has
 ``` #[crown::unrooted_must_root_lint::must_root]``` that triggered the error. I have tried to research about the error and it is concerned with the GC of Javascript and it requires a root present on the stack for it to be used safely. 

How can I go about this sir?

**Comment by jahielkomu** [comment]
This PR says unit tests are failing because of the changes therein. Let me check that again because it built okay locally then update the PR.

**Comment by mrobinson** [comment]
@jahielkomu You can run unit tests locally by executing `./mach test-unit`

**Comment by jahielkomu** [comment]
> @jahielkomu You can run unit tests locally by executing `./mach test-unit`

Thank you sir, I have updated the PR

File: components/script/dom/timeranges.rs [hunk 1]
```diff
@@ -61,11 +61,13 @@ pub struct TimeRangesContainer {
     ranges: Vec<TimeRange>,
 }
 
-impl TimeRangesContainer {
-    pub fn new() -> Self {
+impl Default for TimeRangesContainer {
+    fn default() -> Self {
         Self { ranges: Vec::new() }
     }
+}
```
fabricedesre, line 68: You should be able to just add `Default` to the `#[derive(...)]` decorator of `TimeRangesContainer`.

jahielkomu: > #[derive(...)]

Hello @fabricedesre, thank you for looking at this.

Please correct me if I get it wrong, do you mean the decorator as in 
```
#[derive(Clone, Debug, JSTraceable, MallocSizeOf)]
pub struct TimeRangesContainer {
    ranges: Vec<TimeRange>,
}

impl Default for TimeRangesContainer {
    fn default() -> Self {
        Self { ranges: Vec::new() }
    }
}
```
can be updated to ```#[derive(Default, Clone, Debug, JSTraceable, MallocSizeOf)]``` and then one wouldn't have to replace every call to ``` TimeRangesContainer::new()``` with ```TimeRangesContainer::default()``` in the test files ?

mrobinson (approved):
I've pushed some small changes to this which you can see in the latest commit.

components/script/dom/range.rs
```diff
@@ -1082,13 +1082,19 @@ pub struct WeakRangeVec {
...
+}
+
 #[allow(unsafe_code)]
 impl WeakRangeVec {
     /// Create a new vector of weak references.
     pub fn new() -> Self {
-        WeakRangeVec {
-            cell: UnsafeCell::new(WeakRefVec::new()),
-        }
+        Self::default()
```
mrobinson, line 1097: I preserved the `new()` function here because this is a type like `Vec` which has both `default()` and `new()`.

jahielkomu:
> I've pushed some small changes to this which you can see in the latest commit.

Yes I'm following closely sir, thanks.

PR #31994, opened 2024-04-04
script: Make layout DOM wrappers not generic on layout data

mrobinson:
Remove the type parameter from the layout DOM wrappers. This is possible
now that style and layout data are separate and the `Any` nature of the
layout data is exposed in the wrappers.

Removing the phantom data member of the wrappers also allows using the
default `derive` implementations for things like `Clone`, `Copy`, and
`PartialEq`.

---


jdm (approved):
Yay!

PR #31996, opened 2024-04-04
gfx: Do not apply scale to `CoreText` font metrics

mrobinson:
Since the original version of the CoreText font code, it has scaled the
metrics from CoreText by an unusual scale:

```
    let scale = px_to_pt(self.ctfont.pt_size()) / (ascent + descent);
```

It's unclear what this scale was trying to accomplish. Note that it's
passing the return value of `pt_size()` to `px_to_pt` which seems
backward. This scale seems bogus, but perhaps it's based on a
misconception about what its returned from CoreText. Unlike the return
values of `CGFont` methods, which are returned in font units, the ones
from `CTFont` are "scaled according to the point size and matrix of the
font reference."

Indeed, when just interpreting these values as pixel values, the results
more or less match Firefox and Chrome. This becomes much more obvious
now that we have support for `ex` units. Even when not using `ex`, you
can sometimes see the top parts of glyphs cut off due to this scaling.

This change removes the scaling and simply interpets the return values
of `CTFont` methods as pixels. It addresses all of the issues mentioned
above. Note that this path will eventually just be a fallback path and
metrics will come from sfnt tables in the future.

---

- [x] These changes do not require tests because MacOS layout is currently not tested so testing must be done manually.

### DIFF
--- a/components/gfx/platform/macos/font.rs
+++ b/components/gfx/platform/macos/font.rs
@@ -14,7 +14,6 @@ use core_foundation::base::CFIndex;
 use core_foundation::data::CFData;
 use core_foundation::string::UniChar;
 use core_graphics::font::CGGlyph;
-use core_graphics::geometry::CGRect;
 use core_text::font::CTFont;
 use core_text::font_descriptor::{
     kCTFontDefaultOrientation, SymbolicTraitAccessors, TraitAccessors,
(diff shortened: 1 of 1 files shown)


mrobinson:
Some screenshots comparing before and after:

## Before (left is Firefox):

Rendering this page:
```html
<!DOCTYPE html>

<style>
body {
    margin: 0;
}
</style>
<body>
<div style="font-size: 30px; width: 3ex; height: 3ex; outline: solid;">xxx xxx</div>
</body>
```

Screenshot 2024-04-04 at 16 59 22

Note the dot of the i:
Screenshot 2024-04-04 at 17 11 14

## After

Box size and line gap matches more closely:

Screenshot 2024-04-04 at 13 36 40

Dot of the i fully rendered:
Screenshot 2024-04-04 at 17 18 19

jdm (approved):

mrobinson:
GitHub is having infrastructure issues. Sending this back to the MQ.

PR #32005, opened 2024-04-05
ci: Retry `hdiutil create` with randomize backoff

mrobinson:
Regularly we see that `hdiutil creat` when creating a DMG on the MacOS
builder fails with "Resource busy." This change is an attempt to
mitigate that by retrying the command 3 time with a randomized backoff.

---

- [x] These changes do not require tests because they are just a CI fix.

### DIFF
--- a/python/servo/package_commands.py
+++ b/python/servo/package_commands.py
@@ -8,6 +8,9 @@
 # except according to those terms.
 
 from datetime import datetime
+import random
+import time
+from typing import List
 from github import Github
 
 import hashlib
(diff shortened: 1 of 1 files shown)


python/servo/package_commands.py
```diff
@@ -234,15 +255,19 @@ def package(self, build_type: BuildType, android=None, target=None, flavor=None,
...
-                subprocess.check_call(['hdiutil', 'create',
-                                       '-volname', 'Servo',
-                                       '-megabytes', '900',
-                                       dmg_path,
-                                       '-srcfolder', dir_to_dmg])
+                check_call_with_randomized_backoff(
+                    ['hdiutil', 'create', '-volname', 'Servo',
+                     '-megabytes', '900', dmg_path,
+                     '-srcfolder', dir_to_dmg],
+                    etries=3)
```
mukilan, line 266:
```suggestion
                    retries=3)
```

mukilan (approved):

python/servo/package_commands.py
mrobinson: Thanks. I'm really surprised this didn't cause an error!

PR #32007, opened 2024-04-05
Fixed some clippy warnings by replacing 'match' with 'if'

komutumu:
This PR fixes clippy warnings that arose from the use of 'match' with a suggestion to use 'if'.

---

### DIFF
--- a/components/script/dom/htmlmediaelement.rs
+++ b/components/script/dom/htmlmediaelement.rs
@@ -2766,15 +2766,12 @@ impl FetchResponseListener for HTMLMediaElementFetchListener {
             // restart the download later from where we left, we cancel
             // the current request. Otherwise, we continue the request
             // assuming that we may drop some frames.
-            match e {
-                PlayerError::EnoughData => {
-                    if let Some(ref mut current_fetch_context) =
-                        *elem.current_fetch_context.borrow_mut()
-                    {
-                        current_fetch_context.cancel(CancelReason::Backoff);
-                    }
-                },
-                _ => (),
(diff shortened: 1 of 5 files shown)


jahielkomu:
Hello @mrobinson 
Please kindly review and revert.

Please allow me ask;
```match``` provided an option that served as a default for any value that hadn't been explicitly matched by the preceding patterns using ```_ => {}``` however, in the elimination of that in favor of ``` if ``` no else block has been created since the default pattern was implementing nothing, so I ask won't this cause any kind of redundancy?

eerii:
Hi @jahielkomu! As you said the issue with the `match` statements was that they only checked for one variant and didn't do anything for the others. That's why they can safely be converted into an `if let` that is equivalent. Because there was nothing done for the other variants, no `else` is needed in this case. This actually reduces redundancy as you don't need to explicitly handle all cases like before, just the one you care about.

jahielkomu:
> Hi @jahielkomu! As you said the issue with the `match` statements was that they only checked for one variant and didn't do anything for the others. That's why they can safely be converted into an `if let` that is equivalent. Because there was nothing done for the other variants, no `else` is needed in this case. This actually reduces redundancy as you don't need to explicitly handle all cases like before, just the one you care about! :)

Yes, this makes sense...thanks @eerii. 
I think I meant unhandled cases but you rightly bring out the point that there is no need for concern.

eerii:
Yes, in this case the unhandled cases are not doing anything so they can be omitted safely.

mrobinson (approved):

PR #32008, opened 2024-04-06
webgpu: Use safe callbacks & try_recv_timeout

sagudev:
It also removes the need for HashMaps and BufferMaps struct that keep data alive until it was used by C callback.

Callbacks now do not send request to self anymore so https://sagudev.github.io/briefcase/fastgame.html actually works now.

We now do try_recv_timeout instead of running loop continuously.

---

### DIFF
--- a/components/webgpu/lib.rs
+++ b/components/webgpu/lib.rs
@@ -84,7 +84,10 @@ pub enum WebGPURequest {
         host_map: HostMap,
         map_range: std::ops::Range<u64>,
     },
-    BufferMapComplete(id::BufferId),
+    BufferMapComplete {
+        sender: IpcSender<Option<WebGPUResponseResult>>,
+        buffer_id: id::BufferId,
+    },
     CommandEncoderFinish {
         command_encoder_id: id::CommandEncoderId,
         device_id: id::DeviceId,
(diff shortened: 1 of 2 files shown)


sagudev:
Buffermapasync is not working right: https://github.com/sagudev/servo/actions/runs/8578926656/job/23513773148 yet

sagudev:
It looks like some IPC messages are dropped/missing so some promises are never resolved.

sagudev:
NOTE-TO-SELF: Ipc-channel created from `response_async` should only send one massage back (wgpu runs callback also on failures at buffer map async).

sagudev:
I solved IPC problem (more on that is written in https://servo.zulipchat.com/#narrow/stream/263398-general/topic/ipc_channel) using `Arc<Mutex<VecDeque<_>>>` (queue of messages that were send from itself). Alternative option to self_msg is to put `webrender_api`, `webrender_document` and `global` in ArcMutex, while first two are not really problematic (only used in UpdateWebRenderData) `global` is used practically everywhere so we would need a lot of locks for every operation.

It makes https://sagudev.github.io/briefcase/fastgame.html actually work instead of freeze, but no WPT result is changed: https://github.com/sagudev/servo/actions/runs/8610552319/job/23597139814

components/webgpu/lib.rs
```diff
@@ -1146,50 +1111,55 @@ impl<'a> WGPU<'a> {
...
-                                            buffer_size: info.size,
-                                            external_id: info.external_id.unwrap(),
-                                        },
-                                    )) {
-                                        warn!("Could not send UpdateWebRenderData ({})", e);
+                        let glob = Arc::clone(&self.global);
+                        let wgpu_image_map = Arc::clone(&self.wgpu_image_map);
+                        let webrender_api = Arc::clone(&self.webrender_api);
+                        let webrender_document = self.webrender_document;
+                        let callback = BufferMapCallback::from_rust(Box::from(move |result| {
```
gterzian, line 1118: Can we not just call `global.buffer_map_async` synchronously, and then handle the result and do the webrender stuff?

components/webgpu/lib.rs
```diff
@@ -431,54 +409,42 @@ impl<'a> WGPU<'a> {
...
-                                userdata as *const BufferMapInfo<Option<WebGPUResponseResult>>,
-                            );
-                            let msg = match status {
-                                BufferMapAsyncStatus::Success => {
-                                    let global = &info.global;
-                                    let (slice_pointer, range_size) = gfx_select!(info.buffer_id =>
-                                        global.buffer_get_mapped_range(info.buffer_id, 0, None))
+                        let glob = Arc::clone(&self.global);
+                        let resp_sender = sender.clone();
+                        let callback = BufferMapCallback::from_rust(Box::from(move |result| {
```
gterzian, line 414: same here: do we need the callback?

gterzian (changes requested):
So I've looked a bit into https://docs.rs/wgpu-core/0.20.0/src/wgpu_core/device/global.rs.html#2391, and I'm wondering if we really need the callbacks for the map asyncs. It seems to just makes things more complicated, see for example the lock around webrender. 

This is different from the closures of submitted work done, because those are actually called at [a later stage from when they are submitted](https://github.com/gfx-rs/wgpu/blob/08efa72a83d0acdf25177408150e4c87933c1cd7/wgpu-core/src/device/life.rs#L389). But the map asyncs callbacks are just called right then and there. 

I think https://docs.rs/wgpu/latest/wgpu/struct.BufferSlice.html#method.map_async only applies to when you are using the higher-level wgpu API. the wgpu-core `buffer_map_async` calls the callback synchronously. 

If that is correct(I think it is, but let me know if you think otherwise), I would say we should remove the callbacks, and the locks.

sagudev:
Callback is called synchronously only when result if there is immediate failure (mostly validation): https://docs.rs/wgpu-core/0.20.0/src/wgpu_core/device/global.rs.html#2388 (we currently handle this twice, see https://github.com/servo/servo/issues/32172)

I think I can see why you get confused by callbacks (I got too for few moments), but `op` (`BufferMapOperation` that contains callback) is passed to https://docs.rs/wgpu-core/0.20.0/src/wgpu_core/device/global.rs.html#2498 as described in https://github.com/gfx-rs/wgpu/blob/08efa72a83d0acdf25177408150e4c87933c1cd7/wgpu-core/src/device/life.rs#L222 and triggered on maintain (it's called by device_poll) here: https://github.com/gfx-rs/wgpu/blob/08efa72a83d0acdf25177408150e4c87933c1cd7/wgpu-core/src/device/resource.rs#L455 (later this function returns both submission and mapping closures, that are actually fired in poll functions).

So yeah I think callbacks are really needed (if they were not, they would not be in wgpu). You can always ask in wgpu matrix channel if my explanation is to messy.

gterzian (approved):
> So yeah I think callbacks are really needed 

Ok, I now see the callback is stored over at https://github.com/gfx-rs/wgpu/blob/bdf30fa45aa86b827749f7ee45ed342ce27b0fbb/wgpu-core/src/device/global.rs#L2498

LGTM!

PR #32009, opened 2024-04-06
fix: Handle table.deleteRow with no rows

shanehandley:
Fixes a case where `[table].deleteRow(-1)` is called when it has no child rows.

Spec: https://html.spec.whatwg.org/multipage/tables.html#dom-table-deleterow

WPT failure: https://wpt.fyi/results/html/semantics/tabular-data/the-table-element/remove-row.html?label=master&product=servo

---

### DIFF
--- a/components/script/dom/htmltableelement.rs
+++ b/components/script/dom/htmltableelement.rs
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
     // https://html.spec.whatwg.org/multipage/#dom-table-deleterow
     fn DeleteRow(&self, mut index: i32) -> Fallible<()> {
         let rows = self.Rows();
+
         // Step 1.
-        if index == -1 {
-            index = rows.Length() as i32 - 1;
-        }
-        // Step 2.
-        if index < 0 || index as u32 >= rows.Length() {
+        if index < -1 || index >= rows.Length() as i32 {
             return Err(Error::IndexSize);
         }
-        // Step 3.
-        DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+
+        // Step 2
+        if index == -1 && rows.Length() > 0 {
+            index = rows.Length() as i32 - 1;
+
+            // Step 3.
+            DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+        }
+
         Ok(())
     }
 
--- a/components/script/dom/htmltableelement.rs
+++ b/components/script/dom/htmltableelement.rs
@@ -395,22 +395,31 @@ impl HTMLTableElementMethods for HTMLTableElement {
         Ok(new_row)
     }
 
-    // https://html.spec.whatwg.org/multipage/#dom-table-deleterow
+    /// <https://html.spec.whatwg.org/multipage/#dom-table-deleterow>
     fn DeleteRow(&self, mut index: i32) -> Fallible<()> {
         let rows = self.Rows();
+        let num_rows = rows.Length() as i32;
 
-        // Step 1.
-        if index < -1 || index >= rows.Length() as i32 {
+        // Step 1: If index is less than −1 or greater than or equal to the number of elements
+        // in the rows collection, then throw an "IndexSizeError".
+        if !(-1..num_rows).contains(&index) {
             return Err(Error::IndexSize);
         }
 
-        // Step 2
-        if index == -1 && rows.Length() > 0 {
-            index = rows.Length() as i32 - 1;
+        let num_rows = rows.Length() as i32;
 
-            // Step 3.
-            DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+        // Step 2: If index is −1, then remove the last element in the rows collection from its
+        // parent, or do nothing if the rows collection is empty.
+        if index == -1 {
+            index = num_rows - 1;
         }
+
+        if num_rows == 0 {
+            return Ok(());
+        }
+
+        // Step 3: Otherwise, remove the indexth element in the rows collection from its parent.
+        DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


components/script/dom/htmltableelement.rs
```diff
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
     // https://html.spec.whatwg.org/multipage/#dom-table-deleterow
```
mrobinson, line 398: I think it would be good to fix the rustdoc in this function when we update it:

```suggestion
    /// <https://html.spec.whatwg.org/multipage/#dom-table-deleterow>
```

components/script/dom/htmltableelement.rs
```diff
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
...
     fn DeleteRow(&self, mut index: i32) -> Fallible<()> {
         let rows = self.Rows();
+
         // Step 1.
```
mrobinson, line 402: Traditionally just the step numbers were included in comments, but it turns out that the number changes a lot so a lot of these comments have become misleading. I think it's a good practice now to include the actual step text as it makes the code more future proof.

```suggestion
        // Step 1: If index is less than −1 or greater than or equal to the number of elements
        // in the rows collection, then throw an "IndexSizeError".
```

components/script/dom/htmltableelement.rs
```diff
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
...
-        if index == -1 {
-            index = rows.Length() as i32 - 1;
-        }
-        // Step 2.
-        if index < 0 || index as u32 >= rows.Length() {
+        if index < -1 || index >= rows.Length() as i32 {
```
mrobinson, line 403: This could potentially use a range to make things more idiomatic:

```suggestion
        let num_rows = rows.Length() as i32;
        if !(-1..num_rows).contains(index) {
```

components/script/dom/htmltableelement.rs
```diff
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
...
             return Err(Error::IndexSize);
         }
-        // Step 3.
-        DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+
+        // Step 2
```
mrobinson, line 407:
```suggestion
        // Step 2: If index is −1, then remove the last element in the rows collection from its
        // parent, or do nothing if the rows collection is empty.
```

components/script/dom/htmltableelement.rs
```diff
@@ -398,16 +398,20 @@ impl HTMLTableElementMethods for HTMLTableElement {
...
+        if index == -1 && rows.Length() > 0 {
+            index = rows.Length() as i32 - 1;
+
+            // Step 3.
+            DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+        }
+
```
mrobinson, line 414: I think this makes it so that if the index passed isn't -1 no row is removed. It might be simpler to make this more like:

```suggestion
            if num_rows == 0 {
                return Ok(());
             }
             
             if index == -1 {
                 index = num_rows - 1;
             }

             // Step 3: Otherwise, remove the indexth element in the rows collection from its parent.
             DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self()
        }
```

tests/wpt/meta/html/semantics/tabular-data/the-table-element/remove-row.html.ini
```diff
@@ -1,3 +0,0 @@
-[remove-row.html]
```
mrobinson, line 1: You'll need to also run this in legacy layout mode to update the expectation:

```
./mach run-wpt -r /html/semantics/tabular-data/the-table-element/remove-row.html --legacy-layout --log-raw out.log
./mach update-wpt ---legacy-layout out.log
```

I also recommend running the entire directory as I suspect that would have caught the issue above as well.

mrobinson (changes requested):
Nice catch. I have some comments below:

tests/wpt/meta/html/semantics/tabular-data/the-table-element/remove-row.html.ini
shanehandley: Thank you; updated - apologies I erroneously thought they were both covered by a single file.

> I also recommend running the entire directory as I suspect that would have caught the issue above as well.

This makes a lot of sense 👍 - I regularly see unexpected timeouts when running all tests in a directory, but I can see now it tends to improve after 2-3 runs.

mrobinson: Are you are MacOS by chance? There's an issue where the first time you run Servo on MacOS it needs to build the GStreamer plugin database. If you start servo with `./mach run -r` before running tests you shouldn't see so many timeouts. We're still trying to figure out how to avoid this lag on startup.

components/script/dom/htmltableelement.rs
```diff
@@ -395,19 +395,33 @@ impl HTMLTableElementMethods for HTMLTableElement {
...
             return Err(Error::IndexSize);
         }
-        // Step 3.
-        DomRoot::upcast::<Node>(rows.Item(index as u32).unwrap()).remove_self();
+
+        let num_rows = rows.Length() as i32;
+
+        // Step 2: If index is −1, then remove the last element in the rows collection from its
+        // parent, or do nothing if the rows collection is empty.
+        if (-1..num_rows).contains(&index) {
```
mrobinson, line 412: I think it makes sense to move this check above  to where you return `Err` and reverse it, as it's simply another way of saying the reverse of "if index < -1 || index >= rows.Length() as i32 {" then you can de-indent the code in this block.

shanehandley: Ah thank you - I think I misinterpreted your earlier change request. That's been updated 👍

tests/wpt/meta/html/semantics/tabular-data/the-table-element/remove-row.html.ini
shanehandley: > Are you are MacOS by chance?

I am 👍 I just did this and it works a treat, thank you.

mrobinson (approved):
Thanks! This looks good.

PR #32024, opened 2024-04-09
Fixes for HTTP header compliance.

philip-lamb:
- Fix 400 errors from nginx in response to Servo requests by implementing conformant albeit non-normative removal of whitespace from `Accept` and `Accept-Language` HTTP headers. (To match behaviour of Firefox, Safari, and Chrome) https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
- Provide `Host` header as REQUIRED by HTTP protocol https://www.rfc-editor.org/rfc/rfc9110#field.host

---

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -177,7 +177,7 @@ pub fn set_default_accept_language(headers: &mut HeaderMap) {
     // TODO(eijebong): Change this once typed headers are done
     headers.insert(
         header::ACCEPT_LANGUAGE,
-        HeaderValue::from_static("en-US, en; q=0.5"),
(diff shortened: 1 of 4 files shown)


mrobinson:
It looks like this change is failing unit tests `./mach test-unit`.

jdm (approved):

jdm:
These changes look reasonable to me. I swear hyper used to always provide the host header, but maybe that changed.

philip-lamb:
Thanks Josh.

philip-lamb:
Unfortunately, this change breaks a different set of nginx sites, so `Host`-header-related changes have been backed out in https://github.com/servo/servo/pull/32093.

PR #32025, opened 2024-04-09
Fixed some clippy warnings in components

komutumu:
This PR fixes some clippy warnings by; 

- Removing unused lifetimes
- Simplifying boolean expressions
- Removing useless use of format!
- Replacing the use Option.and_then() with .map()
- Replacing the call of Option.map().flatten() with .and_then()

---

### DIFF
--- a/components/script/dom/range.rs
+++ b/components/script/dom/range.rs
@@ -1021,7 +1021,7 @@ impl RangeMethods for Range {
 
         // Step 4.
         let ancestor = self.CommonAncestorContainer();
-        let mut iter = start_node
+        let iter = start_node
             .following_nodes(&ancestor)
             .filter_map(DomRoot::downcast::<Text>);
 
@@ -1190,7 +1190,7 @@ impl WeakRangeVec {
                 let move_end = node_is_end && range.end_offset() == offset;
 
                 let remove_from_node =
-                    !(!move_start || !move_end && node_is_end) || move_end && !node_is_start;
+                    move_start && (move_end || !node_is_end) || move_end && !node_is_start;
 
                 let already_in_child = range.start().node() == child || range.end().node() == child;
                 let push_to_child = !already_in_child && (move_start || move_end);
@@ -1252,7 +1252,7 @@ impl WeakRangeVec {
                 let move_end = node_is_end && end_offset > offset;
 
                 let remove_from_node =
-                    !(!move_start || !move_end && node_is_end) || move_end && !node_is_start;
+                    move_start && (move_end || !node_is_end) || move_end && !node_is_start;
 
                 let already_in_sibling =
                     range.start().node() == sibling || range.end().node() == sibling;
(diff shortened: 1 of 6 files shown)


jahielkomu:
Hello @mrobinson 
Please kindly review.
Thanks.

jahielkomu:
@mrobinson , a lot of the warnings talk about the use of unsafe code, should we use ```#[allow(unsafe_code)]``` ? 
Please advise.

components/script/dom/range.rs
```diff
@@ -1189,9 +1189,8 @@ impl WeakRangeVec {
                 let move_start = node_is_start && range.start_offset() == offset;
                 let move_end = node_is_end && range.end_offset() == offset;
 
-                let remove_from_node = move_start && move_end ||
-                    move_start && !node_is_end ||
-                    move_end && !node_is_start;
+                let remove_from_node =
+                    !(!move_start || !move_end && node_is_end) || move_end && !node_is_start;
```
mrobinson, line 1193: Hrm...I think this new formulation is more difficult to understand. In this case, I would add an exception for the clippy warning or rewrite the logic to be more understandable.

components/script/dom/range.rs
```diff
@@ -1252,9 +1251,8 @@ impl WeakRangeVec {
                 let move_start = node_is_start && start_offset > offset;
                 let move_end = node_is_end && end_offset > offset;
 
-                let remove_from_node = move_start && move_end ||
-                    move_start && !node_is_end ||
-                    move_end && !node_is_start;
+                let remove_from_node =
+                    !(!move_start || !move_end && node_is_end) || move_end && !node_is_start;
```
mrobinson, line 1255: I feel the same way about this new code.

mrobinson (changes requested):

components/script/dom/range.rs
jahielkomu: I absolutely agree with you @mrobinson ...I picked this straight from the clippy suggestion and felt it was more complicated but thought it was the standard. I checked that it evaluated to the same as the before and applied it.
Let me find another way for this sir.
Thank you.

components/script/dom/range.rs
jahielkomu: Let me work on it sir

jahielkomu:
I have updated the PR @mrobinson , kindly review again.

mrobinson (approved):

jahielkomu:
Hmm something seems to not be going fine sir @mrobinson .
 Is it due to the changes made in the PR?
```./mach test-unit```, ```./mach test-tidy``` and ```./mach build -d``` ran with no errors

jahielkomu:
> @mrobinson , a lot of the warnings talk about the use of unsafe code, should we use `#[allow(unsafe_code)]` ? Kindly advise.

Sir would you recommend this?

mrobinson:
> > @mrobinson , a lot of the warnings talk about the use of unsafe code, should we use `#[allow(unsafe_code)]` ? Kindly advise.


Adding `allow(unsafe_code)` shouldn't be necessary, as it's absence would cause a compilation error. I think it would be helpful to see the actual error messages that you are referring to in the clippy warnings. If they are the ones regarding the missing "# Safety" section in the documentation, the fix is to write that documentation. That requires a bit of understanding and care though, as we want the documentation to actually reflect the real safety requirements of those functions.

PR #32027, opened 2024-04-09
fix: allow form submission for input [type=image]

shanehandley:
Fixes cases where clicks on inputs of type `image` within forms were not triggering a submission.

Includes some tidying up of surrounding code for readability, some links referenced in comments were no longer valid, some small FIXME's. 

WPT failure: https://wpt.fyi/results/html/semantics/forms/form-submission-0/form-submission-algorithm.html?label=master&product=servo)

---


mrobinson (approved):
Thanks for this. It looks like a nice improvement.

mrobinson:
Looks like there are some failures now: https://github.com/servo/servo/runs/23650077411

shanehandley:
> Looks like there are some failures now: https://github.com/servo/servo/runs/23650077411

Oh apologies 🙏 they should be accounted for now.

shanehandley:
I'm a bit perplexed about these failures - it seems to bounce back and forth :/

https://github.com/servo/servo/assets/1322294/9b2ff9d2-3a32-4b7b-96ad-df48697bc515

shanehandley:
Can't work this one out - might revisit in the future.

shanehandley:
The cause of the inconsistent results appears to be timing. The WPT test:

1. Sets up a bunch of DOM interactions
2. Uses event listeners to append the side-effect of those interactions to an array
3. Makes assertions about the array

However the assertions run before the DOM interactions have completed (sometimes they have, sometimes they haven't).

A consistent 133 passes (> 10 runs) can be seen when [adjusting the step_timeout wrapper](https://github.com/web-platform-tests/wpt/blob/master/dom/events/Event-dispatch-single-activation-behavior.html#L157) with a 500ms timeout, which allows 1 & 2 to complete before running assertions in 3.

I lack the knowledge to resolve the performance issue - but thought I'd add this info if someone comes across it in the future.

mrobinson:
@shanehandley Thanks for all your work here and sorry that the PR didn't work out. This kind of timing issue is really tricky and hopefully upcoming fixes for the HTML event loop should make them less common.

Issue #32033, opened 2024-04-09
RFC: Font system redesign


mrobinson:
## Current system

Currently Servo uses a single `FontCacheThread`, which is responsible for loading the system font list and also storing the available web fonts. In addition, every layout thread stores a `FontContext` which caches all font data structures. When the data structure is not in the thread-local `FontContext` cache, IPC messages are used to request the data from the `FontCacheThread`.

### `FontCacheThread`

The `FontCacheThread` is a long-running background thread that is reponsible for reading the sytem font lists and storing data for web fonts. It also manages WebRender font primitives, sending them to the layout-thread-specific `FontContext`. Some of the data structures it manages:

- `FontIdentifier`: Unique platform-specific identifier for a source for creating a font face. For local fonts it contains either a path to a local font (or data used to find a path) and possibly a variation index. For web fonts, this is a URL. Multiple `FontTemplate`s might have the same `FontIdentifier`.
- `FontDescriptor`: Describes a font's properties (weight, stretch, style).
- `FontTemplateData`: Platform-specific data for a font. Contains `Arc<Vec<u8>>` bytes, `FontIdentifier` and in the case of MacOS a `CTFont` cache.
- `FontTemplate`: "All the information needed to create font instance handles." This includes a `FontIdentifier`, a `FontDescriptor`, and the `FontTemplateData` (both strong and weak references).

### `FontContext`

The `FontContext` is the per-layout-thread store of instantiated fonts and cached `FontTemplate`s. When a layout thread is doing font matching, it asks the `FontContext` to find templates that match a given descriptor. These requests are forwarded to the `FontCacheThread` via IPC and the responses are cached. When sending `FontTemplate`s to `FontContext` the `FontCacheThread` does not serialize font data, as that would be too expensive. Instead it serializes all of the members apart from the data into `SerializedFontTemplate` and then sends the data via a direct write into an IPC channel. The `FontContext` is also responsible for turning `FontTemplate`s into concrete `Font`s which hold platform-specific data structures via their `FontHandle` member and other instance information.

### Problems with the current system

 - **Web fonts are global**: Web fonts are associated with family names in the
   global `FontCacheThread`. Here's an example:
   ```html
    <!DOCTYPE html>
    <iframe src="font-iframe-1.html"></iframe>
    <iframe src="font-iframe-2.html"></iframe>
   ```
   `font-iframe-1.html`:
   ```html
   <!DOCTYPE html>
   <style>
   @font-face {
[...]

PR #32034, opened 2024-04-09
gfx: Remove `FontTemplateData`

mrobinson:
Now that `FontTemplateData` is more or less the same on all platforms,
it can be removed. This is a preparatory change for a full refactor of
the font system on Servo. The major changes here are:

 - Remove `FontTemplateData` and move its members into `FontTemplate`
 - Make `FontTemplate` have full interior mutability instead of only
   the `FontTemplateData` member. This is preparation for having these
   data types `Send` and `Sync` with locking.
 - Remove the strong/weak reference concept for font data. In practice,
   all font data references were strong, so this was never fully
   complete. Instead of using this approach, the new font system will
   use a central font data cache with references associated to layouts.
 - The `CTFont` cache is now a global cache, so `CTFont`s can be shared
   between threads. The cache is cleared when clearing font caches.

A benefit of this change (apart from `CTFont` sharing) is that font data
loading is platform-independent now.

This is part of #32033.

---

- [x] These changes do not require tests because this does not change observable behavior.


components/gfx/font_cache_thread.rs
```diff
@@ -74,40 +83,37 @@ impl FontTemplates {
...
 
         // If a request is made for a font family that exists,
         // pick the first valid font in the family if we failed
         // to find an exact match for the descriptor.
-        for template in &mut self.templates {
-            let maybe_template = template.get();
-            if maybe_template.is_some() {
-                return maybe_template;
+        for template in &mut self.templates.iter() {
+            if template.valid() {
```
mukilan, line 115: Do you think renaming this to `is_valid` would be more readable?

mrobinson: Sure. That seems reasonable.

components/gfx/font_cache_thread.rs
```diff
@@ -425,25 +432,20 @@ impl FontCache {
...
+        *webrender_fonts
+            .entry(template.borrow().identifier.clone())
             .or_insert_with(|| {
-                let font = match (template.bytes_if_in_memory(), template.native_font()) {
+                let template = template.borrow();
+                let font = match (template.data_if_in_memory(), template.native_font_handle()) {
                     (Some(bytes), _) => FontData::Raw((*bytes).clone()),
                     (None, Some(native_font)) => FontData::Native(native_font),
-                    (None, None) => FontData::Raw((*template.bytes()).clone()),
+                    (None, None) => unreachable!("Font should either be local of a web font."),
```
mukilan, line 445:
```suggestion
                    (None, None) => unreachable!("Font should either be local or a web font."),
```

components/gfx/font_template.rs
```diff
@@ -84,12 +90,14 @@ impl<'a> From<&'a FontStyleStruct> for FontTemplateDescriptor {
 /// font instance handles. It contains a unique
 /// FontTemplateData structure that is platform specific.
 pub struct FontTemplate {
-    identifier: FontIdentifier,
-    descriptor: Option<FontTemplateDescriptor>,
-    weak_ref: Option<Weak<FontTemplateData>>,
-    // GWTODO: Add code path to unset the strong_ref for web fonts!
-    strong_ref: Option<Arc<FontTemplateData>>,
-    is_valid: bool,
+    pub identifier: FontIdentifier,
```
mukilan, line 93: Are these fields intentionally `pub` and not `pub(crate)`?

mrobinson: Yes, I think these have to be pub because they are used by out-of-crate tests unfortunately.

components/gfx/font_template.rs
```diff
@@ -84,12 +90,14 @@ impl<'a> From<&'a FontStyleStruct> for FontTemplateDescriptor {
...
+    pub descriptor: Option<FontTemplateDescriptor>,
+    /// The data to use for this [`FontTemplate`]. For web fonts, this is always a filled, but
```
mukilan, line 95:
```suggestion
    /// The data to use for this [`FontTemplate`]. For web fonts, this is always filled, but
```

components/gfx/font_template.rs
```diff
@@ -128,95 +123,97 @@ impl FontTemplate {
...
+
+pub trait FontTemplateRefMethods {
+    /// Returns a reference to the data in this font. This may be a hugely expensive
+    /// operation (depending on the platform) which performs synchronous disk I/O
+    /// and should never be done lightly.
+    fn data(&self) -> Arc<Vec<u8>>;
+    /// Return true if this is a valid [`FontTemplate`] ie it is possible to construct a [`FontHandle`]
+    /// from its data.
+    fn is_valid(&self) -> bool;
+    fn instantiate(&self) -> Result<(), &'static str>;
```
mukilan, line 149: Would it be helpful if `instantiate` also had a docstring? For example, I find it hard to judge how it is different from the lazy loading performed by `data`.

components/gfx/platform/macos/font_template.rs
```diff
@@ -2,182 +2,69 @@
...
+                return Some(core_text_font.clone());
+            }
         }
 
-        // If you pass a zero font size to one of the Core Text APIs, it'll replace it with
-        // 12.0. We don't want that! (Issue #10492.)
-        let clamped_pt_size = pt_size.max(0.01);
+        let mut cache = cache.write();
+        let identifier_cache = cache.entry(identifier).or_insert_with(Default::default);
+        if let Some(core_text_font) = identifier_cache.get(&au_size) {
```
mukilan, line 53: Do we perform this check again here because the `identifier_cache` might been updated after the read lock is released and thus it is not guaranteed to be a "cache miss" when we acquire the write lock? If so, do you think it would be helpful to document this scenario in a comment?

mukilan (approved):
Great work!

components/gfx/font_cache_thread.rs
mrobinson: Fixed. Thanks!

components/gfx/font_template.rs
mrobinson: Fixed!

components/gfx/font_template.rs
mrobinson: Sure. I've added a docstring for this method.

mrobinson:
Thanks for the review. I've updated the branch with the changes you suggested. I'll send it to the MQ.

PR #32038, opened 2024-04-10
Remove `FontContextHandle`

mrobinson:
The `FontContextHandle` was really only used on FreeType platforms to
store the `FT_Library` handle to use for creating faces. Each
`FontContext` and `FontCacheThread` would create its own
`FontContextHandle`. This change removes this data structure in favor of
a mutex-protected shared `FontContextHandle` for an entire Servo
process. The handle is initialized using a `OnceLock` to ensure that it
only happens once and also that it stays alive for the entire process
lifetime.

In addition to greatly simplifying the code, this will make it possible
for different threads to share platform-specific `FontHandle`s, avoiding
multiple allocations for a single font.

The only downside to all of this is that memory usage of FreeType fonts
isn't measured (though the mechanism is still there). This is because
the `FontCacheThread` currently doesn't do any memory measurement.
Eventually this *will* happen though, during the font system redesign.
In exchange, this should reduce the memory usage since there is only a
single FreeType library loaded into memory now.

This is part of #32033.

---

- [x] These changes do not require tests because they do not change observable behavior.


atbrakhi (approved):
Thank you for walking me through the changes. It looks good to go.

mrobinson:
It looks like more work is necessary here. It seems FreeType is having issues being used from multiple threads at once. I need to investigate a bit.

mrobinson:
I think the issue here is that I was using `let _ = mutex.lock()` to guard a scope, but it turns out that `let _` doesn't actually create a binding! I've attempted to fix this issue by using `let _guard =`.

PR #32048, opened 2024-04-11
script: Add an implementation of `Default` for `Documents`

akwotom:
---

- [x] These changes do not require tests because no they do not change functionality.

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -437,12 +437,6 @@ pub struct Documents {
 }
 
 impl Documents {
-    pub fn new() -> Documents {
-        Documents {
-            map: HashMapTracedValues::new(),
-        }
-    }
-
     pub fn insert(&mut self, pipeline_id: PipelineId, doc: &Document) {
         self.map.insert(pipeline_id, Dom::from_ref(doc));
     }
@@ -481,6 +475,15 @@ impl Documents {
     pub fn iter(&self) -> DocumentsIter<'_> {
         DocumentsIter {
             iter: self.map.iter(),
+        }
+    }
+}
+
+impl Default for Documents {
+    #[allow(crown::unrooted_must_root)]
+    fn default() -> Self {
+        Self {
+            map: HashMapTracedValues::new(),
         }
     }
 }
@@ -1340,7 +1343,7 @@ impl ScriptThread {
         let control_port = ROUTER.route_ipc_receiver_to_new_crossbeam_receiver(state.control_port);
 
         ScriptThread {
-            documents: DomRefCell::new(Documents::new()),
+            documents: DomRefCell::new(Documents::default()),
             window_proxies: DomRefCell::new(HashMapTracedValues::new()),
             incomplete_loads: DomRefCell::new(vec![]),
             incomplete_parser_contexts: IncompleteParserContexts(RefCell::new(vec![])),


components/script/dom/bindings/callback.rs
```diff
@@ -7,6 +7,7 @@
 use std::default::Default;
 use std::ffi::CString;
 use std::mem::drop;
+// use std::ops::Deref; // Import is unused
```
mrobinson, line 10: Please don't just comment out code. Instead remove unused lines of code.

components/script/dom/window.rs
```diff
@@ -2310,7 +2310,10 @@ impl Window {
         self.Document().url()
     }
 
-    pub fn with_layout<T>(&self, call: impl FnOnce(&mut dyn Layout) -> T) -> Result<T, ()> {
+    pub fn with_layout<T>(
+        &self,
+        call: impl FnOnce(&mut dyn Layout) -> T,
+    ) -> Result<T, script_thread::LayoutNotFoundError> {
```
mrobinson, line 2316: I don't think we want to create `LayoutNotFoundError`. Instead, it would be better to ensure that a layout is always found, which should be true. I have been meaning to make this change.

components/script/script_thread.rs
```diff
@@ -485,6 +485,12 @@ impl Documents {
     }
 }
 
+impl Default for Documents {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
mrobinson, line 493: Let's focus on this issue for this PR. Can you move the contents of the `fn new()` method into `fn default()` and then remove `fn new()` altogether. This will require modifying all of the places that called `Documents::new()`.

components/script/script_thread.rs
```diff
@@ -3567,11 +3589,14 @@ impl ScriptThread {
...
-                            .filter_map(DomRoot::downcast::<HTMLAnchorElement>)
-                            .next()
+                        if Some(
+                            target
+                                .upcast::<Node>()
+                                .inclusive_ancestors(ShadowIncluding::No)
+                                .filter_map(DomRoot::downcast::<HTMLAnchorElement>)
+                                .next(),
+                        )
+                        .is_some()
```
mrobinson, line 3599: Why are you making this change? `Some(something).is_some()` always returns true, so this change breaks this code.

components/script/script_thread.rs
```diff
@@ -854,19 +860,35 @@ impl ScriptThreadFactory for ScriptThread {
...
+        Self { pipeline_id }
+    }
+}
+
+impl std::fmt::Display for LayoutNotFoundError {
+    fn fmt(&self, format: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(format, "No layout found for {}", self.pipeline_id)
+    }
+}
+
```
mrobinson, line 879: Please hold off on doing this for now. This clippy issue should be addressed in a different way.

mrobinson (changes requested):

components/script/script_thread.rs
akwotom: Okay, let met get on it.

mrobinson:
@akwotom Any updates on this one?

akwotom:
> @akwotom Any updates on this one?

Thanks for checking.
I've been having issues undoing the previous commits.

mrobinson (approved):

akwotom:
Wow, thank you.
Glad to have contributed.

akwotom:
I hear servo is in the process of switching to fixed point arithmetic.
Can you please, point me in the direction of contributing to this effort?

PR #32056, opened 2024-04-11
clippy: Fix `redundant_*` warnings

eerii:
Fixes `redundant_*` clippy warnings. Also updates the spec comments in the files changed to be propper rustdoc.

- [`redundant_file_names`](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_field_names): Removes duplicate field names in struct initialization to avoid.
- [`redundant_closure`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure): Functions can be passed directly instead of wrapping them in a closure.
- [`redundant_pattern_matching`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching): `is_ok` and `is_err` can be used instead. Here the only part I'm unsure is what to do about the steps in `components/script/dom/url.rs`, since this merges step 2.2 and 3 together.
- [`redundant_guards`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_guards): The pattern can be added directly to the `Some(...)` pattern instead of checked with an extra `if`.

---
- [x] These changes do not require tests because they don't modify behaviour

### DIFF
--- a/components/script/dom/customelementregistry.rs
+++ b/components/script/dom/customelementregistry.rs
@@ -695,9 +695,9 @@ impl CustomElementDefinition {
             observed_attributes,
             callbacks,
             construction_stack: Default::default(),
-            form_associated: form_associated,
-            disable_internals: disable_internals,
-            disable_shadow: disable_shadow,
+            form_associated,
+            disable_internals,
+            disable_shadow,
         }
     }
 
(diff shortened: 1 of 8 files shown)


jdm (approved):
Thanks!

PR #32059, opened 2024-04-11
Treat indefinite percentages as auto for margin collapse

Loirooriol:
The top and bottom margins of an element can collapse through if its height is auto or zero. Indefinite percentages behave as auto, so they shouldn't prevent the margins from collapsing.

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -13,7 +13,7 @@ use style::computed_values::clear::T as Clear;
 use style::computed_values::float::T as Float;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
-use style::values::computed::{Length, LengthOrAuto};
+use style::values::computed::{Length, LengthOrAuto, Size};
 use style::values::specified::{Display, TextAlignKeyword};
 use style::Zero;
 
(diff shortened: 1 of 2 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -69,6 +69,17 @@ impl BlockContainer {
...
+fn block_size_is_zero_or_auto(size: &Size, containing_block: &ContainingBlock) -> bool {
+    match size {
+        Size::Auto => true,
+        Size::LengthPercentage(ref lp) => {
+            // TODO: Should this resolve definite percentages? Blink does it, Gecko and WebKit don't.
+            lp.is_definitely_zero() ||
+                (lp.0.has_percentage() && containing_block.block_size.is_auto())
+        },
+    }
+}
```
mrobinson, line 81: This is a huge nit, but maybe this free function should go at the bottom of the file instead of amidst the definition and implementation of data structures?

components/layout_2020/flow/mod.rs
```diff
@@ -148,10 +158,10 @@ impl BlockLevelBox {
                     margin_inline_end)
                     .into()
             })
-            .clamp_between_extremums(min_inline_size, max_inline_size);
-
-        // The block size is irrelevant here.
-        let block_size = AuOrAuto::Auto;
```
mrobinson, line 154: Wow!

mrobinson (approved):

PR #32060, opened 2024-04-11
Element collapsing thru should collapse with its children

Loirooriol:
If the top and bottom margins of an element collapse through, then this patch treats the bottom margin as collapsing with its children, even if `height` doesn't compute to zero.

---


mrobinson (approved):
Nice! This just needs an expectations update.

PR #32064, opened 2024-04-12
layout: Resolve definite percentages when determining margin collapse

Loirooriol:
Previously we were treating a definite 0% as zero, but a definite 100% that resolved into 0px as non-zero.

That didn't make much sense, so this patch stops checking the computed values, and instead checks the layout values which have already resolved percentages. This also implies that we will now take `max-block-size` into account.

This brings Servo closer to Blink, but differs from Gecko and WebKit.

Testing: Adding new test

Stylo PR: https://github.com/servo/stylo/pull/254

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7511,7 +7511,7 @@ dependencies = [
 [[package]]
 name = "selectors"
 version = "0.32.0"
-source = "git+https://github.com/servo/stylo?branch=2025-10-01#2f19a2679ed7fd1063d5965b7c2926ef90369bbc"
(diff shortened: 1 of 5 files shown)


mrobinson:
@Loirooriol What's the status of this PR?

Loirooriol:
I can rebase and try to land. There is no CSSWG resolution about what to do, but aligning with Blink will be a net reduction of code (can also remove some methods in Stylo), and seems to make more sense.

Loirooriol:
I will probably need to add some tentative tests

Loirooriol:
Try: https://github.com/servo/servo/actions/runs/18477642921

mrobinson (approved):

PR #32032, opened 2024-04-09
Upgrade Stylo to 2023-12-01

Loirooriol:
This continues #31932

Changelog:
 - Upstream: https://github.com/servo/stylo/compare/7574a08...bf13fe5
 - Servo fixups: https://github.com/servo/stylo/compare/2783456...f6ef64e

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] There are tests for these changes

PR #32066, opened 2024-04-13
Upgrade Stylo to 2024-01-16

Loirooriol:
This continues #32032

Changelog:
 - Upstream: https://github.com/servo/stylo/compare/bf13fe5...ca91506
 - Servo fixups: https://github.com/servo/stylo/compare/f642181...e801dd1

---


components/layout/query.rs
```diff
@@ -766,10 +766,14 @@ pub fn process_resolved_style_request<'dom>(
...
-        // Firefox returns blank strings for the computed value of shorthands,
-        // so this should be web-compatible.
-        PropertyId::ShorthandAlias(..) | PropertyId::Shorthand(_) => return String::new(),
+        PropertyId::NonCustom(id) => {
+            match id.unaliased().as_longhand() {
+                Some(id) => id,
+                // Firefox returns blank strings for the computed value of shorthands,
+                // so this should be web-compatible.
+                None => return String::new(),
+            }
```
mrobinson, line 775: Could this be simply: `id.unaliased().as_longhand().unwrap_or_default()`?

components/layout/query.rs
```diff
@@ -811,10 +815,14 @@ fn process_resolved_style_request_internal<'dom>(
...
-        PropertyId::LonghandAlias(id, _) | PropertyId::Longhand(id) => id,
-        // Firefox returns blank strings for the computed value of shorthands,
-        // so this should be web-compatible.
-        PropertyId::ShorthandAlias(..) | PropertyId::Shorthand(_) => return String::new(),
+        PropertyId::NonCustom(id) => {
+            match id.unaliased().as_longhand() {
+                Some(id) => id,
+                // Firefox returns blank strings for the computed value of shorthands,
+                // so this should be web-compatible.
+                None => return String::new(),
```
mrobinson, line 823: Same comment here...

components/layout_2020/query.rs
```diff
@@ -263,9 +263,9 @@ pub fn process_resolved_style_request_for_unstyled_node<'dom>(
     );
     let style = styles.primary();
     let longhand_id = match *property {
-        PropertyId::LonghandAlias(id, _) | PropertyId::Longhand(id) => id,
-        PropertyId::ShorthandAlias(id, _) | PropertyId::Shorthand(id) => {
-            return shorthand_to_css_string(id, style);
+        PropertyId::NonCustom(id) => match id.longhand_or_shorthand() {
+            Ok(lh) => lh,
+            Err(sh) => return shorthand_to_css_string(sh, style),
```
mrobinson, line 268: I would go with `longhand` and `shorthand` here instead of `lh` and `sh`.

mrobinson (changes requested):

mrobinson (approved):
Looks good modulo a couple nits above.

components/layout/query.rs
Loirooriol: But there is no default LonghandId, the early return is needed.

components/layout_2020/query.rs
Loirooriol: Changed to `longhand_id` and `shorthand_id`.

components/layout/query.rs
mrobinson: Ah, that makes sense!

PR #32081, opened 2024-04-15
script_layout: Remove script to layout messages

mrobinson:
Instead of communicating with layout via messages, script can simply
call methods on the layout trait. This simplifies the way that script
communicates with layout and opens the path to sharing more data
structures between the two systems. This is part of a continuing 
series of cleanups after removing the layout thread.

---


components/shared/msg/constellation_msg.rs
```diff
@@ -466,7 +466,7 @@ pub enum InputMethodType {
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
-/// The equivalent of script_layout_interface::message::Msg
+/// The equivalent of script_layout_interface::Msg
```
mukilan, line 469: `Msg` is removed in this patch, so wouldn't this comment be incorrect? I guess this enum is now unused and can be deleted along with the `HangAnnotation::Layout` variant.  I'm fine if you think that can happenin a separate patch.

mukilan (approved):
Nice!

components/shared/msg/constellation_msg.rs
mrobinson: You're right. I wasn't sure if these would be used again someday, but I think we should remove them now because they are dead code.

Thanks so much for the review!

PR #32093, opened 2024-04-16
Backout `Host`-header related change from #32024.

philip-lamb:
Partial backout of #32024.

#32024 broke a different set of nginx sites, and a more sophisticated fix for the case of a missing HTTP 1.1 `Host` header is required. Until then, revert to previous behaviour.

---


mrobinson (approved):

mrobinson:
This failed to land because of a timeout in the MacOS unit tests. This seems likely to be infrastructure issues, so I'm sending this back to the MQ.

PR #32114, opened 2024-04-19
layout: Add a basic support for `list-style-position: outside`

### LINKED ISSUE #27383: Implement `list-style-position: outside` in Layout 2020
I’m looking into adding layout support for `display: list-item`. What to do for `list-style-position: outside` is not at all clear, https://drafts.csswg.org/css-lists/#list-style-position-property itself says
> This is handwavey nonsense from CSS2, and needs a real definition.

So I’m gonna skip it initially and always do as if `list-style-position: inside` was specified. This issue tracks adding support for `outside` later.

mrobinson:
This change adds very basic support for `list-style-position`.
Currently, the marker does not do any kind of baseline alignment with the rest
of the list item contents and it also doesn't force the list item to be
at least as tall as the marker.

This adds a few new failures:

- Four failures because markers do not ensure that list-items have at least the same block size as they do:
  - FAIL [expected PASS] /css/CSS2/lists/list-style-applies-to-012.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-applies-to-014.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-type-applies-to-012.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-type-applies-to-014.xht
- One failure because we don't yet support the `::marker` pseudo-selector:
  - FAIL [expected PASS] /css/css-position/position-absolute-dynamic-list-marker.html
- One failure because we don't support the list item exception for the line height quirk:
  - FAIL [expected PASS] /quirks/line-height-in-list-item.tentative.html

Fixes #27383.

---

- [x] These changes fix #27383.


mrobinson:
I took a look at these failures:

- Four failures because markers do not ensure that list-items have at least the same block size as they do:
  - FAIL [expected PASS] /css/CSS2/lists/list-style-applies-to-012.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-applies-to-014.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-type-applies-to-012.xht
  - FAIL [expected PASS] /css/CSS2/lists/list-style-type-applies-to-014.xht
- One failure because we don't yet support the `::marker` pseudo-selector:
  - FAIL [expected PASS] /css/css-position/position-absolute-dynamic-list-marker.html
- One failure because we don't support the list item exception for the line height quirk:
  - FAIL [expected PASS] /quirks/line-height-in-list-item.tentative.html

components/layout_2020/flow/mod.rs
```diff
@@ -205,6 +207,84 @@ struct FlowLayout {
...
+            Length::zero(),
+            |current_max, fragment| match fragment {
+                Fragment::Text(text) => current_max.max(text.rect.max_inline_position()),
+                Fragment::Image(image) => current_max.max(image.rect.max_inline_position()),
+                Fragment::Positioning(positioning) => {
+                    current_max.max(positioning.rect.max_inline_position())
+                },
+                //Fragment::Box(box_fragment) | //=> {
+                //    current_max.max(box_fragment.content_rect.max_inline_position())
+                //},
```
Loirooriol, line 253: Can be removed

components/layout_2020/flow/mod.rs
```diff
@@ -270,6 +350,9 @@ fn calculate_inline_content_size_for_block_level_boxes(
                 let style_box = &float_box.contents.style().get_box();
                 Some((size, style_box.float, style_box.clear))
             },
+            BlockLevelBox::OutsideMarker { .. } => {
+                Some((ContentSizes::zero(), Float::None, Clear::None))
+            },
```
Loirooriol, line 355: I think this could just use `None` together with the `OutOfFlowAbsolutelyPositionedBox`.
Since `::marker` appears at the beginning the behavior should be the same.

components/layout_2020/lists.rs
```diff
@@ -55,11 +55,11 @@ where
...
-        ListStyleType::Disc => Some("• "),
-        ListStyleType::Circle => Some("◦ "),
-        ListStyleType::Square => Some("▪ "),
-        ListStyleType::DisclosureOpen => Some("▾ "),
-        ListStyleType::DisclosureClosed => Some("‣ "),
+        ListStyleType::Disc => Some("•\u{00a0}"),
+        ListStyleType::Circle => Some("◦\u{00a0}"),
+        ListStyleType::Square => Some("▪\u{00a0}"),
+        ListStyleType::DisclosureOpen => Some("▾\u{00a0}"),
+        ListStyleType::DisclosureClosed => Some("‣\u{00a0}"),
```
Loirooriol, line 62: This seems like it would cause extra spacing for inside markers, in a way that would match Firefox but was considered bad by the CSSWG, see https://github.com/w3c/csswg-drafts/issues/4891.

But it doesn't actually seem to happen, which may be due to some white space processing bug? So I guess it's fine for now, but we should consider setting `white-space: pre` just for outside markers.

Loirooriol (approved):

components/layout_2020/flow/mod.rs
mrobinson: Removed.

components/layout_2020/flow/mod.rs
mrobinson: Oh, good point. Adjusted this.

components/layout_2020/lists.rs
Loirooriol: Not necessarily, because it should only be added to outside markers, and there is no `::marker:outside` or such.

PR #32115, opened 2024-04-19
fonts: Merge multiple methods into `PlatformFont::descriptor()`

### LINKED ISSUE #15103: Windows downloadable font support may need some fleshing out
Windows downloadable font support added in #15096 has some limitations around the family/face name that may or may not need to be fixed.  Specifically, it only pulls out Mac-format (`platform_id == 1`) English names from the font container.  I'm not sure if the font family/face is relevant at all for downloaded fonts, though, because I believe the family name has to be specified in the `@font-face` rule anyway.

mrobinson:
This combines `style()`, `boldness()`, `stretchiness()` into a
`descriptor()` method which is used when creating `FontTemplate`s for
web fonts. Eventually this method will simply read font tables using
skrifa. This is the first step.

In addition, `family_name()` and `face_name()` are removed. They were
only used for debugging and the `FontIdentifier` serves for that. On
Windows, this was adding another way in which font loading could fail,
without buying us very much. The path or URL to the font is more
important when debugging than the names in the font tables.

Closes #15103.

---

### DIFF
--- a/components/gfx/font.rs
+++ b/components/gfx/font.rs
@@ -18,7 +18,7 @@ use log::debug;
 use serde::{Deserialize, Serialize};
 use servo_atoms::{atom, Atom};
 use smallvec::SmallVec;
-use style::computed_values::{font_stretch, font_style, font_variant_caps, font_weight};
+use style::computed_values::font_variant_caps;
 use style::properties::style_structs::Font as FontStyleStruct;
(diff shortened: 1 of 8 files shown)


mrobinson:
@mukilan This is a another cleanup on top of the work we have done to streamline the font loading code.

mrobinson:
This one should be ready to review n

mukilan (approved):

PR #32120, opened 2024-04-20
fix(user-timing): fix clearing marks/measures by name

nolanlawson:
This fixes several tests in [wpt/user-timing](https://wpt.fyi/results/user-timing?label=master&product=chrome%5Bexperimental%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&product=servo&aligned) by fixing some logic errors in how marks/measures are cleared (via [`clearMarks`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/clearMarks) and [`clearMeasures`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/clearMeasures)).

There are two changes:

1. Fix the boolean logic in `clear_entries_by_name_and_type` so that, when `clearMarks('foo')` or `clearMeasures('foo')` is called, the presence of the entry name correctly filters based on existing entry names.
2. Make the `entry_name` param a `DOMString` rather than an `Option<DOMString>` since every API call has it as `Some` anyway, and I'm not aware of any [Performance APIs](https://developer.mozilla.org/en-US/docs/Web/API/Performance) where you can clear all entries regardless of type. (This is not strictly required for the fix, but I think it makes the code easier to read.)

~~I also considered adding the expected WPT results using `mach update-wpt`. But I'm not sure if you want these changes, since the expectations are currently missing (i.e. `tests/wpt/meta/user-timing` does not exist).~~ (_Update: added!_)

For the record, this PR fixes the following tests:

- `clearMarks.html.ini`
- `clearMeasures.html.ini`
- `clear_non_existent_mark.any.js.ini`
- `clear_non_existent_measure.any.js.ini`
- `clear_one_mark.any.js.ini`
- `clear_one_measure.any.js.ini`

~~In case you do want these meta files, here they are: https://github.com/nolanlawson/servo/commit/510e6146ba01bae4f4aafda8754718e24ce6e868~~

---


components/script/dom/performance.rs
```diff
@@ -93,13 +93,11 @@ impl PerformanceEntryList {
...
-        entry_type: Option<DOMString>,
+        entry_type: DOMString,
     ) {
         self.entries.retain(|e| {
-            name.as_ref().map_or(true, |name_| *e.name() != *name_) &&
-                entry_type
-                    .as_ref()
-                    .map_or(true, |type_| *e.entry_type() != *type_)
+            *e.entry_type() != *entry_type ||
+                name.as_ref().map_or(false, |name_| *e.name() != *name_)
```
nolanlawson, line 100: When the `name` is `None`, then the expectation is that all marks/measures are removed. If it's `Some`, then we need to compare the name and only remove entries that match.

jdm:
Ah, I suspect that that wpt test directory is not part of the default set of enabled directories. We should update the test config to correct that!

nolanlawson:
Ah, I see, so should `user-timing` go in this file then? [`tests/wpt/include.ini`](https://github.com/servo/servo/blob/main/tests/wpt/include.ini)

I could do that in this PR or a separate one if you'd like.

jdm:
Why don't we make a separate PR that changes that file and adds the current set if expected results, then this PR can show the change in behavior more clearly?

jdm:
Alternatively, doing that work in this PR in two separate commits sound also be ok.

nolanlawson:
OK, now that #32123 is merged, I rebased on `main` and updated the relevant tests. ^

mrobinson (approved):

Issue #17267, opened 2017-06-10
insufficient font fallback handling


imiuka:
hi there!

servo does not "properly" handle font family fallbacks: it effectively uses a single fallback font (when no font family is specified, when serif or sans-serif is specified, etc.

if no explicit font family is specified then the fallback will be used. on most platforms (macos), the default fallback family does not include cjk glyphs, emoji, or other obscure glyphs, and so they won't render:

![example](https://user-images.githubusercontent.com/975570/27006511-9ca2ea3a-4e78-11e7-8a6e-0afb34d9a943.png)

```html
<!-- fail -->                  <div>のコ Hello! ✂📋👌</div>
<!-- fail -->                  <div style="font-family: Arial, serif, sans-serif">のコ Hello! ✂📋👌</div>
<!-- ok, only if installed --> <div style="font-family: 'MS Gothic'">のコ Hello! ✂📋👌</div>
<!-- ok, only if installed --> <div style="font-family: Meiryo">のコ Hello! ✂📋👌</div>
```

- `fn last_resort_font_families` (`/components/gfx/platform/{platform-name}/font_list.rs`), returns a static list of potential fallback font families
- `FontCache::last_resort_font_template()` (its only caller) uses this list, takes the first font family that exists on the user's system, and discards the rest
- callers of `FontCache::last_resort_font_template()` then add that single fallback font to their candidate font templates

on windows, the fallback list is simply `arial`. macos uses `arial unicode ms, arial` and linux is a trio of `fira sans, dejavu sans, arial`. of these, only `arial unicode ms` (which is on the macos list) has cjk glyphs, but that family isn't installed by default, so for many, servo will use `arial`.

---

in the short term, we could improve this situation by adding a few more fallbacks to the list of fallback font families (eg, arial + one that supports cjk glyphs), and then modifying `FontContext::layout_font_group_for_style()` to add all fallbacks to the list of candidate font templates which is used in layout and rendering.

i'm not sure about the long term viability of this option, though. firefox's font fallback implementation ([windows implementation linked][1]) selects between 40(!) font families! `FontContext::layout_font_group_for_style()` returns a `SmallVec<8>` for each requested font, and simply adding all fonts would massively bloat that vec.

long term, i see servo needing a static decision table, similar to the current firefox implementation, and ways to take into account user preferences.

---

i'm happy to help work on a fix, but i'm not a servo team member and have never contributed before, so i'd like guidance on what the best path forward is (or if i'm completely wrong and off).

[...]

PR #32122, opened 2024-04-21
fonts: Fix emoji font selection on macOS

mrobinson:
This fixes two issues that were preventing emojis from being properly
selected from fonts on macOS.

1. `CTFont::get_glyphs_for_characters` takes the input characters as
   UniChar which are UTF-16 encoded characters. We need to encode the
   input `char` as UTF-16 before passing it to CoreText.
2. The font fallback list is updated with the latest logic from Gecko,
   which importantly adds "Apple Color Emoji" to the list of fallback
   fonts. Sorry for the big change, but this is just a direct port of
   the code from Gecko.

With these two changes, emojis display but in grayscale. 😅 To fix this,
another part of the font stack will need to detect when the font
supports color and pass that information to WebRender when creating the
font instance. We will likely do this in platform independent way later
that will depend on some more preliminary changes.

---

- [x] There are tests for these changes, but the macOS CI does not currently run WPT so we cannot observe the updated results.

### DIFF
--- a/components/gfx/platform/macos/font.rs
+++ b/components/gfx/platform/macos/font.rs
@@ -10,7 +10,6 @@ use std::{fmt, ptr};
 /// Implementation of Quartz (CoreGraphics) fonts.
 use app_units::Au;
 use byteorder::{BigEndian, ByteOrder};
-use core_foundation::base::CFIndex;
 use core_foundation::data::CFData;
 use core_foundation::string::UniChar;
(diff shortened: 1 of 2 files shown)


mrobinson:
Here is the rendering of the test case in #17267 after this change:

Screenshot 2024-04-21 at 14 17 46

mukilan (approved):

PR #32123, opened 2024-04-21
test: add WPT user-timing tests

nolanlawson:
As discussed in https://github.com/servo/servo/pull/32120#issuecomment-2068017033, this adds the WPT `user-timing` tests to the default list of WPT tests, and also commits the current status of the `user-timing` tests.

---

### DIFF
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -222,6 +222,8 @@ skip: true
   skip: false
 [url]
   skip: false
+[user-timing]
+  skip: false
 [wasm]
   skip: false
 [webaudio]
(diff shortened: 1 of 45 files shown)


mrobinson (changes requested):
Thanks! This looks great. One change necessary before landing: you'll also need to copy your results to `tests/wpt/meta-legacy-layout` because the same tests are run in both layout systems.

mrobinson:
I'll also send this to the CI to see if it agrees with your generated results; sometimes timing / font issues can throw things off.

nolanlawson:
Thanks for the review! Added to `meta-legacy-layout`.

mrobinson (approved):

PR #32124, opened 2024-04-21
docs: update README instructions for updating WPT

nolanlawson:
I noticed in #32123 that there are two issues with the docs for updating Web Platform Tests:

- `./mach update-wpt --sync` fails with an error (`Are you sure you don't want a patch?`) because it expects a `--patch` arg.
- The `tests/wpt/meta-legacy-layout/` test results need to be updated with a separate command.

This updates `tests/wpt/README.md` with more complete instructions.

---

- [x] These changes do not require tests because this is a docs-only change

### DIFF
--- a/tests/wpt/README.md
+++ b/tests/wpt/README.md
@@ -199,12 +199,22 @@ next synced.
 Updating the upstream tests
 ===========================
 
-In order to update the tests from upstream use the same mach update
-commands. e.g. to update the web-platform-tests:
+In order to update the tests from upstream, use the same mach update
+commands. E.g. to update the web-platform-tests:
 
-    ./mach update-wpt --sync
+    ./mach update-wpt --sync --patch
+
+Then, to update the test results:
+
     ./mach test-wpt --log-raw=update.log
     ./mach update-wpt update.log
+
+You'll also need to do the same for the legacy layout engine:
+
+    ./mach test-wpt --legacy-layout --log-raw=legacy.log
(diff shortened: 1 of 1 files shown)


tests/wpt/README.md
```diff
@@ -199,13 +199,23 @@ next synced.
...
 
+You'll also need to do the same for the legacy layout engine:
+
+    ./mach test-wpt --legacy-layout --log-raw=legacy.log
+    ./mach update-wpt --legacy-layout legacy.log
+
+Finally, commit the results from `tests/wpt/meta/` and `tests/wpt/meta-legacy-layout/`.
+
 This should create two commits in your servo repository with the
 updated tests and updated metadata.
```
mrobinson, line 221: I think this section can be removed entirely. This is about syncing the tests in the Servo repository with the ones from upstream, but this is now done by a CI job. 

At the same time the `--legacy-layout` option should be added to this section about updating test expectations: https://github.com/servo/servo/blob/1e315496f3db564ed440c9e28f61779768309013/tests/wpt/README.md#updating-test-expectations

mrobinson (changes requested):

tests/wpt/README.md
nolanlawson: Makes sense! Updated: 8d892502cb688f64e922037194f58e99685a2d4e

mrobinson (approved):
Thank you!

PR #32127, opened 2024-04-22
fonts: Rework platform font initialization

### LINKED ISSUE #13317: Support TrueType font files that contain multiple fonts
Hello,

i'd like to know how servo handles truetype font files with multiple fonts, since webrender always (hardcoded) chooses [the first font](https://github.com/servo/webrender/blob/master/webrender/src/platform/linux/font.rs#L62) in the font data given to it.
Is it some kind of preprocessing where the font data is manually split up?

Thanks in advance!

### LINKED ISSUE #24554: font-style italic with font-family doesn't work
```html
<style>
body {
  font-family: helvetica, arial, sans-serif;
}

h1 {
  font-style: italic;
}
</style>

<h1>Italic Font Style!</h1>
```

image

The `h1` should be rendered with italic style.

Built with latest master (b7b1b903d3) on macOS 10.14.6

cc @emilio

mukilan:
This change reworks the way that platform fonts are created and
descriptor data is on `FontTemplate` is initialized.

The main change here is that platform fonts for local font faces are
always initialized using the font data loaded into memory from disk.
This means that there is now only a single path for creating platform
fonts.

In addition, the font list is now responsible for getting the
`FontTemplateDescriptor` for local `FontTemplate`s. Before the font had
to be loaded into memory to get the weight, style, and width used for
the descriptor. This is what fonts lists are for though, so for every
platform we have that information before needing to load the font. In
the future, hopefully this will allow discarding fonts before needing to
load them into memory. Web fonts still get the descriptor from the
platform handle, but hopefully that can be done with skrifa in the
future.

Thsese two fixes together allow properly loading indexed font variations
on Linux machines. Before only the first variation could be
instantiated.

Fixes https://github.com/servo/servo/issues/13317.
Fixes https://github.com/servo/servo/issues/24554.

----

- [x] These changes fix #13317 and #24554 

### DIFF
--- a/components/gfx/font_cache_thread.rs
+++ b/components/gfx/font_cache_thread.rs
@@ -11,7 +11,7 @@ use std::sync::{Arc, Mutex};
 use std::{f32, fmt, mem, thread};
 
 use app_units::Au;
-use gfx_traits::{FontData, WebrenderApi};
+use gfx_traits::WebrenderApi;
 use ipc_channel::ipc::{self, IpcReceiver, IpcSender};
 use log::{debug, trace};
 use net_traits::request::{Destination, Referrer, RequestBuilder};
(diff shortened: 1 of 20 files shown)


components/gfx/font_cache_thread.rs
```diff
@@ -59,10 +60,20 @@ pub enum FontIdentifier {
     Web(ServoUrl),
 }
 
+impl FontIdentifier {
+    pub fn index(&self) -> u32 {
+        let i = match *self {
+            Self::Local(ref local_font_identifier) => local_font_identifier.index(),
+            Self::Web(_) => 0,
+        };
+        i
```
mrobinson, line 69:
```suggestion
    pub fn index(&self) -> u32 {
        match *self {
            Self::Local(ref local_font_identifier) => local_font_identifier.index(),
            Self::Web(_) => 0,
        }
```

mrobinson (approved):

components/gfx/font_cache_thread.rs
mukilan: Oops. Good catch! Fixed in recent push.

PR #32131, opened 2024-04-22
script: Fixed some clippy warnings

komutumu:
This PR fixes some clippy warnings.

- Replaces ```match``` with ```if let``` or ```if```
- Replaces ```.and_then()``` with ```.map()```
- Replaces negated comparison with ```partial_cmp()```
---
- [x] These changes do not require tests because it touches not functionality.

### DIFF
--- a/components/script/dom/customelementregistry.rs
+++ b/components/script/dom/customelementregistry.rs
@@ -484,12 +484,11 @@ impl CustomElementRegistryMethods for CustomElementRegistry {
         if form_associated {
             let _ac = JSAutoRealm::new(*cx, proto_object.get());
             unsafe {
-                match self.add_form_associated_callbacks(proto_object.handle(), &mut callbacks) {
-                    Err(error) => {
-                        self.element_definition_is_running.set(false);
-                        return Err(error);
-                    },
-                    Ok(()) => {},
+                if let Err(error) =
+                    self.add_form_associated_callbacks(proto_object.handle(), &mut callbacks)
+                {
+                    self.element_definition_is_running.set(false);
+                    return Err(error);
                 }
             }
         }
@@ -673,6 +672,7 @@ pub struct CustomElementDefinition {
 }
 
 impl CustomElementDefinition {
+    #[allow(clippy::too_many_arguments)]
     fn new(
         name: LocalName,
         local_name: LocalName,
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -743,8 +743,10 @@ impl Document {
                 .parent()
                 .and_then(|parent| parent.document())
                 .map(|document| document.base_url());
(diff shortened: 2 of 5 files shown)


jahielkomu:
Hello @mrobinson 
Kindly review.
Thanks

components/script/dom/range.rs
```diff
@@ -343,7 +343,7 @@ impl RangeMethods for Range {
         } else {
             // Step 3.
             self.set_start(node, offset);
-            if !(self.start() <= self.end()) {
+            if self.start().partial_cmp(self.end()) == Some(Ordering::Greater) {
```
mrobinson, line 346: Why not change this to `if self.start() > self.end()` ?

components/script/dom/range.rs
```diff
@@ -362,7 +362,7 @@ impl RangeMethods for Range {
         } else {
             // Step 3.
             self.set_end(node, offset);
-            if !(self.end() >= self.start()) {
+            if self.end().partial_cmp(self.start()) == Some(Ordering::Less) {
```
mrobinson, line 365: And here `self.end() < self.start()` ?

mrobinson (commented):
A couple questions about this one:

components/script/dom/range.rs
jahielkomu: Thank you @mrobinson for looking at this.
I had used ```partial_cmp``` in case self.start() or self.end() is a none integer number but looking at ```offset: u32``` , I can say it's an integer.
So yes sir, ```if self.start() > self.end() ``` is the best pick here.

Thanks again.

mrobinson (approved):

components/script/dom/range.rs
```diff
@@ -343,7 +343,7 @@ impl RangeMethods for Range {
         } else {
             // Step 3.
             self.set_start(node, offset);
-            if !(self.start() <= self.end()) {
```
mrobinson, line 346: It looks like this change needs to be reverted as it's causing a failure on the bots.

components/script/dom/range.rs
```diff
@@ -362,7 +362,7 @@ impl RangeMethods for Range {
         } else {
             // Step 3.
             self.set_end(node, offset);
-            if !(self.end() >= self.start()) {
+            if self.end() < self.start() {
```
mrobinson, line 365: Same with this one.

mrobinson (changes requested):

components/script/dom/range.rs
jahielkomu: > It looks like this change needs to be reverted as it's causing a failure on the bots.

Hello sir, let me revert it. 
Do the bots only allow this particular code?

components/script/dom/range.rs
jahielkomu: It appears to have been reverted already sir.

mrobinson (approved):

PR #32160, opened 2024-04-26
Android: load url from Intent, plus fixed some warnings

Gae24:
This seems that was already implemented, but was no longer working after some refactors

---

- [X] These changes do not require tests because it's a fix for a feature already implemented

### DIFF
--- a/ports/jniapi/src/lib.rs
+++ b/ports/jniapi/src/lib.rs
@@ -872,7 +872,7 @@ fn get_options(
     prefs.insert("js.baseline_interpreter.enabled".to_string(), false.into());
     prefs.insert("js.baseline_jit.enabled".to_string(), false.into());
     prefs.insert("js.ion.enabled".to_string(), false.into());
-    
+
     if let Some(url) = url {
         prefs.insert("shell.homepage".to_string(), url.into());
     }
--- a/ports/jniapi/src/lib.rs
+++ b/ports/jniapi/src/lib.rs
@@ -73,7 +73,7 @@ pub extern "C" fn Java_org_mozilla_servoview_JNIServo_init(
     callbacks_obj: JObject,
     surface: JObject,
 ) {
-    let (mut opts, log, log_str, _gst_debug_str) = match get_options(&env, opts, surface) {
+    let (opts, log, log_str, _gst_debug_str) = match get_options(&env, opts, surface) {
         Ok((opts, log, log_str, gst_debug_str)) => (opts, log, log_str, gst_debug_str),
         Err(err) => {
             throw(&env, &err);
@@ -873,12 +873,9 @@ fn get_options(
     prefs.insert("js.baseline_jit.enabled".to_string(), false.into());
     prefs.insert("js.ion.enabled".to_string(), false.into());
 
-    if let Some(url) = url {
-        prefs.insert("shell.homepage".to_string(), url.into());
-    }
-
     let opts = InitOptions {
         args: args.unwrap_or(vec![]),
+        url,
         coordinates,
         density,
         xr_discovery: None,
--- a/ports/jniapi/src/lib.rs
+++ b/ports/jniapi/src/lib.rs
@@ -872,6 +872,10 @@ fn get_options(
     prefs.insert("js.baseline_interpreter.enabled".to_string(), false.into());
     prefs.insert("js.baseline_jit.enabled".to_string(), false.into());
     prefs.insert("js.ion.enabled".to_string(), false.into());
+    
+    if let Some(url) = url {
+        prefs.insert("shell.homepage".to_string(), url.into());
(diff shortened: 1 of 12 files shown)


ports/jniapi/src/lib.rs
```diff
@@ -873,6 +873,10 @@ fn get_options(
     prefs.insert("js.baseline_jit.enabled".to_string(), false.into());
     prefs.insert("js.ion.enabled".to_string(), false.into());
 
+    if let Some(url) = url {
+        prefs.insert("shell.homepage".to_string(), url.into());
```
mukilan, line 877: Overriding the homepage preference to pass along the URL to load doesn't seem like the best approach here.
For example, if we always override the preference here, we will not be able to support a 'go to homepage' button in the UI to navigate away from the loaded URL.

We should probably just extend `InitOptions` to take an optional `url` member and handle it in [simpleservo::init](https://github.com/servo/servo/blob/66563ed0273c05ab9d687c883649521d14d610e8/ports/jniapi/src/simpleservo.rs#L259).

For reference, `servoshell` also has similar [fallback logic](https://github.com/servo/servo/blob/66563ed0273c05ab9d687c883649521d14d610e8/ports/servoshell/app.rs#L87) and there were some internal discussions about refactoring out the logic common to both ports, but that will be a bigger effort, so for now handling it in simpleservo::init should be fine.

mukilan (changes requested):
Thank you for the contribution! I've left a comment regarding the logic to pass along the URL to servo. The remaining changes related to the warnings look good, so I'm fine if you'd like to work on the URL logic separately and split the warning fixes into a separate PR.

ports/jniapi/src/lib.rs
Gae24: Should be fine now

support/android/apk/servoview/build.gradle
```diff
@@ -8,7 +8,6 @@ import java.util.regex.Pattern
 
 android {
     compileSdk 33
-    buildToolsVersion "33.0.2"
```
mukilan, line 11: This change breaks the build on NixOS. Although the `buildToolsVersion` is optional, [according to the documentation](https://developer.android.com/tools/releases/build-tools), it is not defaulted to the latest build tools for the SDK, but a version that the Android Gradle Plugin decides. 

```
You should always keep your Build Tools component updated by downloading the latest version using the [Android SDK Manager](https://developer.android.com/studio/intro/update). If you're using [Android plugin for Gradle 3.0.0](https://developer.android.com/studio/releases/gradle-plugin#3-0-0) or higher, your project automatically uses a default version of the build tools that the plugin specifies. To use a different version of the build tools, specify it using [buildToolsVersion](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.AppExtension.html#com.android.build.gradle.AppExtension:buildToolsVersion) in your module's build.gradle, as follows:
```
In my case, on NixOS, this means that Android Gradle Plugin will try to download and install the Build Tools version 30.0.3 into the *read-only* nix store and will fail.
```
Observed package id 'ndk;25.2.9519653' in inconsistent location '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk-bundle' (Expected '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk/25.2.9519653')
Observed package id 'ndk;25.2.9519653' in inconsistent location '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk-bundle' (Expected '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk/25.2.9519653')
Observed package id 'ndk;25.2.9519653' in inconsistent location '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk-bundle' (Expected '/nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk/25.2.9519653')
Errors during XML parse:
Additionally, the fallback loader failed to parse the XML.
Checking the license for package Android SDK Build-Tools 30.0.3 in /nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/licenses
License for package Android SDK Build-Tools 30.0.3 accepted.
Preparing "Install Android SDK Build-Tools 30.0.3 (revision: 30.0.3)".
```

Gae24: Only that it is deprecated, will put it back.

mukilan (approved):

PR #32163, opened 2024-04-27
Update WebView variants of ConstellationMsg 

### LINKED ISSUE #32162: Can't show multiple views because of compositor webviews are not in sync
**Describe the bug:**
I've been experimenting with multiview features like https://github.com/servo/servo/pull/31545 recently. For my use case, I want to display[ two views in the window](https://twitter.com/Yu_Wei_Wu/status/1783889417956123050) at the same time. I noticed the other webview usually won't show when I send `EmbedderEvent::ShowWebView` or `EmbedderEvent::FocusWebView` during `EmbedderMsg::WebViewOpened` message. It didn't always happen, but it feels like 80% of chance.

This is because constellation and compositor have their own webview managers. The other webviews aren't immediately added to the compositor's webview manager when `NewWebView` is received. So, when the compositor is handling `ShowWebView,` it might not have added the webview yet.

**To Reproduce:**
Servoshell hasn't supported multiview yet. But the whole workflow is like the following:

- Embedder sends an `EmbedderEvent::NewWebView` to create a second view.
- Constellation receives the event, [send `EmbedderMsg::WebViewOpened` back to embedder](https://github.com/servo/servo/blob/main/components/constellation/constellation.rs#L2980), but it doesn't send any event to compositor to add this webview yet.
- Embedder receives the message and sends `EmbedderEvent::FocusWebView`, constellation receives the event and sends `EmbedderEvent::WebViewFocused`, and finally embedder sends `EmbedderEvent::ShowWebView` in hope to see the webview eventually.
- Compositor received the event, which was redirected by constellation. But now, compositor's webview manager doesn't have this webview yet. Hence, it can't show the webview.
- After a few more iterations, constellation finally called [update_webview_in_compositor](https://github.com/servo/servo/blob/main/components/constellation/constellation.rs#L5417) to add this webview because it needs to traverse history or change session.

**Platform:**
I've tested on Windows and mac. It should apply to all other platforms.

wusyong:
This PR updates a few webview variants in `ConstellationMsg` enum:
- Add `ConstellationMsg::WebViewOpened` to make sure a webview is added to constellation and compositor before sending `EmbedderEvent::WebViewOpened` to embedder.
- Remove `MoveResizeWebView`, `ShowWebView`, `HideWebView`, `RaiseWebViewToTop` variants because they are not called. Servo instance will call [compositor methods](https://github.com/servo/servo/blob/main/components/servo/lib.rs#L767) directly without sending to constellation. This should make it clear that these event won't come to constellation and simplify the workflow.

---

- [x] These changes fix #32162


mrobinson (approved):

wusyong:
Sorry I forgot the first webview is a special case and it needs to send the message as well.

mrobinson (approved):

PR #32171, opened 2024-04-28
feat: Support font-relative `ch` and `ic` units

andreubotella:
After #31966, which made it possible for the first time to resolve font-relative CSS units, this change adds support for the `ch` and `ic` units.

One difference with the `ex` unit that was added in that PR is that these units must reflect the advance width of a character (the zero digit in the case of `ch`, and the CJK water radical for `ic`) as it would be rendered by the current font group. This means that the size of these units don't only depend on the first available font, in the case where that font does not contain a glyph for that character.

This is implemented by adding the advance width for these two characters as optional fields of `FontMetrics`, so the advance width computation happens in advance. Then, when the font metrics are queried as part of unit resolution, the font group is searched for the first font containing that character.

This change only implements support for these units in upright typesetting modes, since Servo does not yet have support for vertical writing modes. This means that many of the WPT tests that test for the behavior of these units with vertical writing modes do not pass.

This change also makes a number of WPT tests pass, which relied on the `ch` and `ic` units. It, however, also makes the test `/css/css-text/white-space/text-wrap-balance-overflow-002.html` fail, since it tests `text-wrap: balance`, which Servo does not yet implement, and it was only previously passing by chance due to the previous behavior of these units.

---


components/layout_thread_2020/lib.rs
```diff
@@ -1261,16 +1262,30 @@ impl FontMetricsProvider for LayoutFontMetricsProvider {
...
+                    let font = font_group
+                        .borrow_mut()
+                        .find_by_codepoint(font_context, codepoint)?;
+                    let font = font.borrow();
+                    let glyph_index = font.glyph_index(codepoint)?;
+                    // TODO: Use the vertical advance for vertical writing modes.
+                    let advance = font.glyph_h_advance(glyph_index);
+                    Some(CSSPixelLength::new(advance as f32))
+                });
+
```
mrobinson, line 1282: I think it would be better to move this to the `font.rs` and have this calculated when creating a font versus here, late in the process.

mrobinson (changes requested):

components/layout_thread_2020/lib.rs
andreubotella: For both of these units, any given font in the font group might not have a glyph for the character, and the required advance is that for the first available font which has it. (While it's unlikely for any font to not have a zero glyph, this is much more likely for the CJK water ideograph.) So these advances are a property of the font group, not of a font.

Would it make sense to compute the advances of these two characters for every font, and to resolve them when creating a `FontGroup`?

mrobinson: I think the best approach in that case is to make that field in the `FontMetrics` an `Option`. That way we only fetch this glyph once when the font is initialized rather than every time the style system needs it.

mrobinson: Here's the relevant code in Gecko for FreeType: [gfxFT2FontBase.cpp](https://searchfox.org/mozilla-central/rev/0916ef0172ce5b2a72749b659da8ad95f637ef42/gfx/thebes/gfxFT2FontBase.cpp#480).

mrobinson (approved):
This looks great, Andreu. You'll also need to run our version of rustfmt via `./mach fmt` and please squash you commits so that the upstream WPT PR is closed properly as well. Thank you. Once that's done I'll send this to the MQ.

mrobinson:
> 🤖 This change no longer contains upstreamable changes to WPT; closed existing upstream pull request ([web-platform-tests/wpt#46007](https://github.com/web-platform-tests/wpt/pull/46007)).

Oh, I spoke to soon. Our scripts are clever enough to close this upstream PR without needing to squash.

mrobinson (approved):

mrobinson:
Looks like there is one more consistently passing test:

```
    PASS [expected FAIL] /css/css-text/white-space/white-space-pre-wrap-trailing-spaces-021.html
```

PR #32173, opened 2024-04-29
webgpu: Update to wgpu 0.20

sagudev:
The only real change is Transmute trait that got removed from wgpu (and left us without queue_id->device_id) transmute so I added it back, others are actually provided by wgpu.

---

### DIFF
--- a/components/webgpu/lib.rs
+++ b/components/webgpu/lib.rs
@@ -81,7 +81,8 @@ pub enum WebGPURequest {
         buffer_id: id::BufferId,
         device_id: id::DeviceId,
         host_map: HostMap,
-        map_range: std::ops::Range<u64>,
+        offset: u64,
+        size: Option<u64>,
     },
     CommandEncoderFinish {
         command_encoder_id: id::CommandEncoderId,
@@ -407,7 +408,8 @@ impl WGPU {
                         buffer_id,
                         device_id,
                         host_map,
-                        map_range,
+                        offset,
+                        size,
                     } => {
                         let glob = Arc::clone(&self.global);
                         let resp_sender = sender.clone();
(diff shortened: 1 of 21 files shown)


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/8895986601/job/24429950022

`CRASH [expected OK] /_webgpu/webgpu/cts.https.html?q=webgpu:api,validation,encoding,beginRenderPass:depth_stencil_attachment,device_mismatch:*` is also reproducible on firefox, issue opened upstream: https://github.com/gfx-rs/wgpu/issues/5628

components/webgpu/lib.rs
```diff
@@ -1458,3 +1466,33 @@ pub struct PresentationData {
     image_desc: ImageDescriptor,
     image_data: ImageData,
 }
+
+pub trait Transmute<U: id::Marker> {
```
gterzian, line 1470: Does it not make sense to simply follow the approach of https://github.com/gfx-rs/wgpu/pull/5509/files and use the explicit method calls(such as `into_queue_id`)?

gterzian (approved):
LGTM, with one question.

components/webgpu/lib.rs
sagudev: Yes, it would, but we are missing `into_device_id`, so we need to roll our own anyway and that way we know that when `into_queue_id` is removed we need to remove queue_id->device_id transmute too.

We should be able to remove all transmute usages anyway: https://github.com/servo/servo/issues/32222.

sagudev: I changed command_buffer<->command_encoder transmutes to into_command_buffer_id, into_command_encoder_id calls.

Issue #32039, opened 2024-04-10
Servo crashes after maximizing the minibrowser


atbrakhi:
**Describe the bug:**
When you try to maximize the minibrowser, Servo crashes. It seems to be happening after https://github.com/servo/servo/pull/31417

**To Reproduce:**
Try to maximize the minibrowser

**Platform:**
This is happening on macOS, not sure about other platform. 

Error:
```
➜  servo git:(main) ./mach run
No build type specified so assuming `--dev`.
assertion `left == right` failed
  left: (1281, 36053)
 right: (0, 36053) (thread main, at /Users/atbrakhi/Documents/codespace/servo/components/compositing/compositor.rs:2268)
   0: backtrace::backtrace::libunwind::trace
             at /Users/atbrakhi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/backtrace-0.3.71/src/backtrace/libunwind.rs:105:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/atbrakhi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/backtrace-0.3.71/src/backtrace/mod.rs:66:5
   1: <servoshell::backtrace::Print as core::fmt::Debug>::fmt
             at /Users/atbrakhi/Documents/codespace/servo/ports/servoshell/backtrace.rs:54:13
   2: core::fmt::rt::Argument::fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/fmt/rt.rs:138:9
      core::fmt::write
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/fmt/mod.rs:1114:21
   3: std::io::Write::write_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/io/mod.rs:1763:15
   4: servoshell::backtrace::print
             at /Users/atbrakhi/Documents/codespace/servo/ports/servoshell/backtrace.rs:18:5
   5: servoshell::main::{{closure}}
             at /Users/atbrakhi/Documents/codespace/servo/ports/servoshell/lib.rs:139:21
   6: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/alloc/src/boxed.rs:2021:9
      std::panicking::rust_panic_with_hook
[...]

PR #32174, opened 2024-04-29
compositing: Fix `flip_rect` calculation

wusyong:
---

- [x] These changes fix the aftermath of https://github.com/servo/servo/issues/32039. `max.y` didn't flip before. I guess it's because we assume `max.y == framebuffer.y` back then.
- [x] These changes do not require tests, because they fix a bug only present when running servoshell with a GUI.


mrobinson (approved):

mrobinson:
It would be nice to at least add unit tests for this functionality.

wusyong:
> It would be nice to at least add unit tests for this functionality.

I'll add the tests later this week. I also find some subtle bugs when maximize and minimize but the flip doesn't seem like the only cause. I'll look into them further.

PR #22740, opened 2019-01-21
  Truncate debug output for long data URLs #22485

Hyperion101010:
I tried to solve the patch, here i did what you told but i am not getting the
  exact result after the build can you take a look

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] These changes fix #22485 (GitHub issue number if applicable)

- [x] There are tests for these changes OR
- [x] These changes do not require tests because ___

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/22740)

PR #32177, opened 2024-04-29
url: Only truncate data URLs for `Debug`

mrobinson:
Other types of URLs aren't so long that they need to be truncated.

---

- [x] These changes do not require tests because they just change debug output.


mrobinson:
@mukilan This is a followup after our discussion about truncation of URLs in `Debug` output.

mrobinson:
Here's the original PR for context: https://github.com/servo/servo/pull/22740

mukilan (approved):
Nice!

PR #32188, opened 2024-04-30
Disable `/fetch/api/crashtests/huge-fetch.any.js`

### LINKED ISSUE #32168: WPT Import fails
https://github.com/servo/servo/actions/runs/8865818832/job/24343044922#step:8:2419

I suspect it's from https://github.com/web-platform-tests/wpt/pull/45848

> ▶ TIMEOUT [expected OK] /fetch/api/crashtests/huge-fetch.any.html

but it only happens when doing flaky detection (could be that python does not run GC and when we run [huge-response.py](https://github.com/web-platform-tests/wpt/pull/45848/files#diff-9e54b6b9e7d378d95a21ed28dc1b545a3133fba9f73d7052ded0b7e104737eaf) for the second time we get OOM).

Temporary solution could be to just disable running this test.

sagudev:
This tests was causing the WPT import task to fail, perhaps because Python does not run GC and when we run it for the for the second time, resulting in an OOM error.

Fixes #32168.

---

- [x] These changes do not require tests because they work: https://github.com/servo/servo/actions/runs/8892777516

### DIFF
--- a/tests/wpt/meta-legacy-layout/fetch/api/crashtests/huge-fetch.any.js.ini
+++ b/tests/wpt/meta-legacy-layout/fetch/api/crashtests/huge-fetch.any.js.ini
@@ -1,11 +1,11 @@
 [huge-fetch.any.worker.html]
-  disabled: true
+  disabled: https://github.com/servo/servo/issues/32168
 
 [huge-fetch.any.sharedworker.html]
-  disabled: true
+  disabled: https://github.com/servo/servo/issues/32168
 
 [huge-fetch.any.html]
-  disabled: true
+  disabled: https://github.com/servo/servo/issues/32168
 
 [huge-fetch.any.serviceworker.html]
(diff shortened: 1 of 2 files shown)


mrobinson (approved):
Can you also add a link to this bug in the ini so that it is easier to track down? I think the syntax is "bug: ". Thanks for unblocking the WPT import!

sagudev:
> Can you also add a link to this bug in the ini so that it is easier to track down? I think the syntax is "bug: ". Thanks for unblocking the WPT import!

I did not find `bug:` but `disabled:` accepts string. See https://web-platform-tests.org/tools/wptrunner/docs/expectation.html

mrobinson:
> I did not find `bug:` but `disabled:` accepts string. See https://web-platform-tests.org/tools/wptrunner/docs/expectation.html

Perfect. Thank you!

PR #32189, opened 2024-04-30
fonts: Improve emoji support

mrobinson:
- On FreeType allow rendering bitmap fonts, which should allow rendering
  full-color emoji with Noto Color Emoji.
- Better detect situations where emoji is necessary by looking ahead one
  character while laying out. This allow processing Unicode presentation
  selectors. When detecting emoji, put emoji fonts at the front of
  fallback lists for all platforms.
- Allow sending the "EMBEDDED_BITMAPS" flag to WebRender when detecting
  fonts with bitmaps.

This enables full-color emoji on Linux and monochrome emoji on Windows.
Full-color emoji on Windows probably needs full support for processing
the COLR table and drawing separate glyph color layers.

---


mrobinson:
Closing this one for now as it needs a bit of work.

PR #32203, opened 2024-05-01
fonts: Use `FontInstanceFlags::EMBEDDED_BITMAPS` for color fonts on MacOS

mrobinson:
This flag ensures that these fonts are rendered full color in WebRender,
allowing for full color emoji.

---

- [x] There are tests for these changes, but only the manual test added here.
      This is tricky to test via WPT because color fonts are sometimes bitmap
      fonts and sometimes fonts with COLR layers. In addition, we don't run tests
      for MacOS on the bots. The hope is that there can be good automated tests for this
      for the Linux implementation.


mrobinson:
Before this change:
Screenshot 2024-05-01 at 10 27 54

After this change:
Screenshot 2024-05-01 at 10 30 18

There's more work to do here, but this is one step on the way to proper emoji support on MacOS and other platforms.

components/gfx/font_context.rs
```diff
@@ -206,16 +211,18 @@ impl<S: FontSource> FontContext<S> {
...
-            font_template,
-            font_descriptor,
-            font_instance_key,
+        let mut font = Font::new(
+            font_template.clone(),
+            font_descriptor.clone(),
             synthesized_small_caps,
-        )?)))
+        )?;
+        font.font_key = self.font_source.get_font_instance(
```
mukilan, line 219: We are first instantiating the `Font` with the default `FontInstanceKey` and then overriding it again with a valid one - is this because the a `Font` instance is needed to first get the flags to be passed to WebRender in order to get the key from it?

In the future and not necessarily in this same PR, would it be possible to get the flags from the template instead of the `Font` instance?

mukilan (approved):

components/gfx/font_context.rs
mrobinson: I think it would be hard to get this from the `FontTemplate`, but there's a possibility that we could create the `PlatformFont` first and get it from there. The other option is we could move the `FontInstanceKey` out of the font or make it an `Option`. There could conceivably be some cases where we want to keep the font cached, but also not have it in WebRender.

PR #32218, opened 2024-05-02
Servoshell: Add support for mouse thumb buttons to go back/forward

### LINKED ISSUE #31655: servoshell: make the mouse thumb buttons go back and forward
The lower thumb button should go back, and the upper thumb button should go forward. These buttons are sometimes known as “mouse4” and “mouse5”, “X1” and “X2”, or buttons 8 and 9.

The code that handles mouse input is in [Minibrowser::on_event](https://github.com/servo/servo/blob/f3a73dbed39c38c6581fc356d3d4f2d38c906e93/ports/servoshell/minibrowser.rs#L89-L111).

This may be a good [newcomer issue](https://github.com/servo/servo/labels/B-newcomer) once winit is upgraded to 0.29 (#31278).

MunishMummadi:
This patch adds support to make mouse buttons go backward and forward in servoshell's minibrowser.

[Refer to this doc](https://docs.rs/winit/latest/winit/event/enum.MouseButton.html#variant.Back)
-  And this line `macOS: Back and Forward might not work with all hardware.`
- The feature doesn't seem to work on my mac after changes.
- [x] These changes fix #31655

### DIFF
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -23,6 +23,7 @@ use servo::rendering_context::RenderingContext;
 use servo::servo_geometry::DeviceIndependentPixel;
 use servo::servo_url::ServoUrl;
 use servo::style_traits::DevicePixel;
+use winit::event::{ElementState, MouseButton, WindowEvent};
 
 use crate::egui_glue::EguiGlow;
 use crate::events_loop::EventsLoop;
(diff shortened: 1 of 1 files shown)


ports/servoshell/minibrowser.rs
```diff
@@ -107,10 +108,20 @@ impl Minibrowser {
                 self.last_mouse_position
                     .map_or(false, |p| self.is_in_browser_rect(p))
             },
-            winit::event::WindowEvent::MouseWheel { .. } |
-            winit::event::WindowEvent::MouseInput { .. } => self
-                .last_mouse_position
-                .map_or(false, |p| self.is_in_browser_rect(p)),
```
mrobinson, line 113: It looks like this change no longer checks to see if the last mouse position was in the browser rectangle? What's going on here?

mrobinson:
It's a bit unclear from your message: have you verified that this change works or does it just not work on Mac?

MunishMummadi:
I read that it doesn't work on some mac hardware(I might be completely wrong). Even though I removed `MouseWheel` event , the mouse buttons should work right?. But removing both `MouseWheel` & `MouseInput` isn't a good idea. I will change it.

ports/servoshell/minibrowser.rs
MunishMummadi: Yeah, I thought I made a way around i.e I simplified things. But it isn't the case. I will try to do some changes.

MunishMummadi:
Even though `./mach build --dev` doesn't produce any errors. I am not able to create the functionality. So, I am making this PR a draft.

mrobinson:
If this isn't functional at all, let's close this one for now to clear it out of the queue. Once you have this working just open it up again and I'll be more than happy with to take a look at it. Thanks!

MunishMummadi:
Got it!!

PR #32220, opened 2024-05-03
Do not use crown when building docs

sagudev:
---

- [x] These changes do not require tests because it's mini fix


mrobinson (approved):

mrobinson:
Eventually it might make sense to only use crown for a special lint job that's run on the CI.

PR #32230, opened 2024-05-05
Implement attributes for the `<meter>` element

shanehandley:
Implement attributes for the html `<meter>` element.

https://html.spec.whatwg.org/multipage/form-elements.html#the-meter-element

---


components/script/dom/htmlmeterelement.rs
```diff
@@ -52,4 +57,138 @@ impl HTMLMeterElement {
...
+        let max = *self.Max();
+
+        Finite::wrap(
+            self.upcast::<Element>()
+                .get_string_attribute(&local_name!("value"))
+                .parse_floating_point_number()
+                .map_or(0.0, |candidate_actual_value| {
+                    if candidate_actual_value < min || min > max {
+                        return min;
+                    }
```
sagudev, line 73: Why is this needed? Isn't clamp enough?

components/script/dom/htmlmeterelement.rs
```diff
@@ -52,4 +57,138 @@ impl HTMLMeterElement {
...
+        let max = *self.Max();
+
+        Finite::wrap(
+            self.upcast::<Element>()
+                .get_string_attribute(&local_name!("low"))
+                .parse_floating_point_number()
+                .map_or(min, |candidate_low_boundary| {
+                    if candidate_low_boundary < min || min > max {
+                        return min;
+                    }
```
sagudev, line 131: ditto

components/script/dom/htmlmeterelement.rs
```diff
@@ -52,4 +57,138 @@ impl HTMLMeterElement {
...
+        let low: f64 = *self.Low();
+
+        Finite::wrap(
+            self.upcast::<Element>()
+                .get_string_attribute(&local_name!("high"))
+                .parse_floating_point_number()
+                .map_or(max, |candidate_high_boundary| {
+                    if candidate_high_boundary < low || min > max {
+                        return low;
+                    }
```
sagudev, line 157: ditto

components/script/dom/htmlmeterelement.rs
```diff
@@ -52,4 +57,138 @@ impl HTMLMeterElement {
...
+        let min = *self.Min();
+
+        Finite::wrap(
+            self.upcast::<Element>()
+                .get_string_attribute(&local_name!("optimum"))
+                .parse_floating_point_number()
+                .map_or(max.add(min).div(2.0), |candidate_optimum_point| {
+                    if min <= max {
+                        return candidate_optimum_point.clamp(min, max);
+                    }
```
sagudev, line 182: ditto

tests/wpt/meta/html/dom/reflection-forms.html.ini
```diff
@@ -3380,132 +3380,24 @@
...
-    expected: FAIL
-
-  [meter.value: IDL set to -10000000000]
-    expected: FAIL
-
-  [meter.value: IDL set to -1]
-    expected: FAIL
-
   [meter.value: IDL set to -0]
     expected: FAIL
```
sagudev, line 3384: Apparently some setters do not work correctly (maybe wrong parsing?).

components/script/dom/htmlmeterelement.rs
shanehandley: Oh avoiding panics while working on WPT passes, I can't reproduce those cases now, so removed 👍

tests/wpt/meta/html/dom/reflection-forms.html.ini
shanehandley: Sorry for not responding to this comment yet but I went down a rabbit hole, I switched the setter implementations to `parse_floating_point_number` before setting the value, but the issue persisted.

I'm really unsure if this is something in `parse_floating_point_number` or not :/

[Spidermonkey's](https://searchfox.org/mozilla-central/source/dom/base/ResponsiveImageSelector.cpp#36) `parseFloat(-0.0)` returns `0`. But our implementation returns -0.0 - do you think this might be a factor or am I barking up the wring tree?

Thank you!

sagudev: I think that our parse_floating_point_number is wrong, but we can leave this for follow up. In firefox the corresponding function is actually https://searchfox.org/mozilla-central/source/dom/base/nsAttrValue.cpp#1892 (you can see it's only used on progress and meter element just like ours).

sagudev (approved):

sagudev:
Thank you!

Issue #32226, opened 2024-05-04
Intermittent timeout in /performance-timeline/not-restored-reasons/performance-navigation-timing-navigation-failure.tentative.window.html


jdm:
```
TIMEOUT [expected OK] /performance-timeline/not-restored-reasons/performance-navigation-timing-navigation-failure.tentative.window.html

    TIMEOUT [expected FAIL] subtest: Ensure that navigation failure blocks bfcache and gets recorded.
```

Issue #32213, opened 2024-05-02
Intermittent TIMEOUT in /performance-timeline/not-restored-reasons/performance-navigation-timing-iframes-without-attributes.tentative.window.html


syvb:
```
  ▶ TIMEOUT [expected OK] /performance-timeline/not-restored-reasons/performance-navigation-timing-iframes-without-attributes.tentative.window.html
  │   ▶ TIMEOUT [expected FAIL] RemoteContextHelper navigation using BFCache
  └   └   → Test timed out
```

TimvdLippe:
Closing as stale as this issue has not received any failure references for over a month

PR #32241, opened 2024-05-07
Skip all flaky tests in performance-timeline/not-restored-reasons

syvb:
Skips all of the `performance-timeline/not-restored-reasons` tests. There are already issues for #32226 and #32213, but it seems #31361 made all of these tests flaky.

---

- [X] These changes do not require tests because they remove tests

### DIFF
--- a/tests/wpt/include.ini
+++ b/tests/wpt/include.ini
@@ -192,6 +192,8 @@ skip: true
     skip: true
 [performance-timeline]
   skip: false
+  [not-restored-reasons]
+    skip: true
 [permissions]
   skip: false
 [quirks]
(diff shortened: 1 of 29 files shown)


sagudev:
There is also [`/fetch/fetch-later/send-on-deactivate.tentative.https.window.html`](https://github.com/servo/servo/runs/24645420377), although maybe we should just revert https://github.com/servo/servo/pull/31361 an bake it a little more (before all those test were FAIL now, they are flaky TIMEOUT/FAIL, so there isn't anything that would be missing).

mrobinson:
> There is also [`/fetch/fetch-later/send-on-deactivate.tentative.https.window.html`](https://github.com/servo/servo/runs/24645420377), although maybe we should just revert https://github.com/servo/servo/pull/31361 an bake it a little more (before all those test were FAIL now, they are flaky TIMEOUT/FAIL, so there isn't anything that would be missing).

It does sound like we should revert #31361 if it made an entire directory of tests flaky.

mukilan:
The `/html/browsers/browsing-the-web/remote-context-helper-tests/navigation-bfcache.window.html` test has also failed in the following two runs and seems related to the BFCache 
https://github.com/servo/servo/runs/24648485771
https://github.com/servo/servo/runs/24654937105

I'm in favor of reverting #31361

PR #32270, opened 2024-05-11
clippy: Fix last few warnings

Pi-Cla:
---
- [x] These changes do not require tests because this is just clippy

### DIFF
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -260,7 +260,7 @@ impl Layout for LayoutThread {
         self.handle_request(Request::FromFontCache);
     }
 
-    fn device<'a>(&'a self) -> &'a Device {
+    fn device(&self) -> &Device {
         self.stylist.device()
     }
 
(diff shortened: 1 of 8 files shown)


components/layout_thread/lib.rs
```diff
@@ -877,36 +877,34 @@ impl LayoutThread {
                     .base()
                     .restyle_damage
                     .contains(ServoRestyleDamage::REPAINT) ||
-                    self.display_list.borrow().is_none()
+                    self.display_list.borrow().is_none() && reflow_goal.needs_display_list()
```
sagudev, line 880: I think && has higher priority than || so this code isn't equal.

sagudev: Clippy gives right suggestion (note the brackets):
```
   |
14 | /     if f1() || f2() {
15 | |         if f3() {
16 | |             
17 | |         }
18 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if
   = note: `#[warn(clippy::collapsible_if)]` on by default
help: collapse nested if block
   |
14 ~     if (f1() || f2()) && f3() {
15 +         
16 +     }
   |
```

Pi-Cla: Good catch, I did not know that && has higher precedence than ||.
I've just rebased a fix.

sagudev (approved):

PR #32272, opened 2024-05-12
Implement additional logic in DOMString::set_best_representation_of_the_floating_point_number

### LINKED ISSUE #32269: parse_floating_point_number does not conform to spec
As noted in https://github.com/servo/servo/pull/32230#discussion_r1597356650

SPEC: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-floating-point-number-values

shanehandley:
This corrects some failing tests related to "-0.0" values for the `meter` element, as well as updates input number and range elements - these were not failing similar tests, it appears that the default fallback was accounting for -0.0 values.

There are also some fixes to the `progress` element where it was mishandling negative values for `max`.

See https://github.com/servo/servo/issues/32269 for additional context.

---

- [x] These changes fix #32269

### DIFF
--- a/tests/wpt/meta/html/dom/reflection-forms.html.ini
+++ b/tests/wpt/meta/html/dom/reflection-forms.html.ini
@@ -3380,24 +3380,6 @@
   [meter.tabIndex: IDL set to -2147483648]
     expected: FAIL
 
-  [meter.value: IDL set to -0]
-    expected: FAIL
-
-  [meter.min: IDL set to -0]
-    expected: FAIL
-
-  [meter.max: IDL set to -0]
-    expected: FAIL
-
-  [meter.low: IDL set to -0]
-    expected: FAIL
-
-  [meter.high: IDL set to -0]
-    expected: FAIL
-
-  [meter.optimum: IDL set to -0]
-    expected: FAIL
-
   [form.tabIndex: setAttribute() to "7\\v"]
     expected: FAIL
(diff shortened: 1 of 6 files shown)


tests/wpt/meta/html/dom/reflection-forms.html.ini
```diff
@@ -3551,18 +3533,6 @@
...
-  [progress.max: IDL set to -1]
-    expected: FAIL
-
-  [progress.max: IDL set to -0]
-    expected: FAIL
-
-  [progress.max: IDL set to 0]
-    expected: FAIL
-
   [progress.max: IDL set to 1e-10]
```
sagudev, line 3536: I think there are even more tests that we could fix with this (look for `IDL set to x`).

mrobinson (approved):

tests/wpt/meta/html/dom/reflection-forms.html.ini
shanehandley: Thank you - I did have a longer look into those. I think there's an edge case in our implementation of `is_valid_floating_point_number_string` that fails for `1.e2`. I'm pretty sure this one should be valid, but our regex enforces that a digit appears after the decimal point.

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#valid-floating-point-number

PR #32276, opened 2024-05-13
Fix RefCell borrows

sagudev:
I think this should fix https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Panic.20on.20touch.20event.20after.20.2331505, but I wrote it online so it might not work.

---


gterzian (approved):
Nice catch!

PR #32286, opened 2024-05-14
fonts: Improve font fallback

### LINKED ISSUE #12944: Myanmar text doesn't display properly on Ubuntu
On Ubuntu 16.04 with two Myanmar fonts i.e. Noto Sans Myanmar and Padauk installed, Servo Commit 94152523 cannot display Myanmar text properly as shown below:

myanmar2

However it should look somewhat like this:

myanmar1

Remark:  Myanmar text was taken from word-break-break-all-007.htm  http://file.veerkesto.net/servo_test/myanmar.html

### LINKED ISSUE #31664: On Linux, add Noto fonts to fallback list for other languages
When a character isn’t supported by any of the fonts chosen by the page, some browsers search through all of the installed fonts to find one that supports text in that language (e.g. [Firefox](https://searchfox.org/mozilla-central/rev/109bb25545f0d2df31954dc0a9afbf30d900b6bb/gfx/thebes/gfxPlatformFontList.cpp#1086-1101)). Servo can’t do that yet, but we currently have hardcoded lists of “fallback fonts” that we try in those situations. There are separate lists for each platform, and on Linux, we use the list in [freetype/font_list.rs](https://github.com/servo/servo/blob/f3a73dbed39c38c6581fc356d3d4f2d38c906e93/components/gfx/platform/freetype/font_list.rs#L155-L169).

[Noto fonts](https://fonts.google.com/noto/use) are often installed by default on Linux distros like Ubuntu. We can use [char::script()](https://docs.rs/ucd/0.1.1/ucd/trait.Codepoint.html#tymethod.script) to check what [script](https://en.wikipedia.org/wiki/Script_(Unicode)) (language) the character is in.

1. Find an enum variant in [Script](https://docs.rs/ucd/0.1.1/ucd/tables/scripts/enum.Script.html) that also has a Noto font [according to this page](https://notofonts.github.io/)
2. In [this function](https://github.com/servo/servo/blob/f3a73dbed39c38c6581fc356d3d4f2d38c906e93/components/gfx/platform/freetype/font_list.rs#L155-L169), check if `codepoint` is in that script
3. If so, add the Noto Sans font name to the fallback list

For example, for [`Script::Adlam`](https://docs.rs/ucd/0.1.1/ucd/tables/scripts/enum.Script.html#Adlam.v) we can add “Noto Sans Adlam” to the fallback list:

image

mrobinson:
- Better detect situations where emoji is necessary by looking ahead one
  character while laying out. This allow processing Unicode presentation
  selectors. When detecting emoji, put emoji fonts at the front of
  fallback lists for all platforms.

  This enables monochrome emoji on Windows. Full-color emoji on Windows
  probably needs full support for processing the COLR table and drawing
  separate glyph color layers.

- Improve the font fallback list on FreeType platforms. Ideally, Servo
  would be able to look through the entire font list to find the best
  font for a certain character, but until that time we can make sure the
  font list contains the "Noto Sans" fonts which cover most situations.

Fixes #31664.
Fixes #12944.

---

- [x] These changes are difficult to test. It's hard to test if font fallback happened and, if it
        did, what font it fell back to. I have tested these changes with https://angband.pl/doc/alluni.txt
        though and they significantly improve Servo's coverage on Linux. There's more work to do
        for macOS and Windows though.

### DIFF
--- a/components/gfx/font.rs
+++ b/components/gfx/font.rs
@@ -32,7 +32,7 @@ use crate::platform::font::{FontTable, PlatformFont};
 pub use crate::platform::font_list::fallback_font_families;
 use crate::text::glyph::{ByteIndex, GlyphData, GlyphId, GlyphStore};
 use crate::text::shaping::ShaperMethods;
-use crate::text::Shaper;
+use crate::text::{FallbackFontSelectionOptions, Shaper};
 
 #[macro_export]
 macro_rules! ot_tag {
@@ -487,9 +487,12 @@ impl FontGroup {
         &mut self,
         font_context: &FontContext<S>,
         codepoint: char,
+        next_codepoint: Option<char>,
     ) -> Option<FontRef> {
+        let options = FallbackFontSelectionOptions::new(codepoint, next_codepoint);
+
         let should_look_for_small_caps = self.descriptor.variant == font_variant_caps::T::SmallCaps &&
-            codepoint.is_ascii_lowercase();
+            options.character.is_ascii_lowercase();
         let font_or_synthesized_small_caps = |font: FontRef| {
             if should_look_for_small_caps && font.synthesized_small_caps.is_some() {
                 return font.synthesized_small_caps.clone();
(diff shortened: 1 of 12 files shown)


components/gfx/font.rs
```diff
@@ -486,20 +486,23 @@ impl FontGroup {
     pub fn find_by_codepoint<S: FontSource>(
         &mut self,
         font_context: &FontContext<S>,
-        codepoint: char,
+        character: char,
```
atbrakhi, line 489: We have renamed `codepoint` to `character` in this PR, would it make sense to rename the function as well? maybe something `find_by_character`?

atbrakhi (approved):
Looks good to me, just left a very small comment.

components/gfx/font.rs
mrobinson: Oh, great point. I've gone ahead and renamed these parameters`codepoint` and `next_codepoint`. It may make sens to rename this function later, but for now it's probably best to keep the same name.

PR #32288, opened 2024-05-15
layout: use `au` in `HoistedSharedFragment`

atbrakhi:
Use `au` in `HoistedSharedFragment` and `AbsoluteBoxOffsets`

---

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -1762,9 +1762,8 @@ impl PlacementState {
             },
             Fragment::AbsoluteOrFixedPositioned(fragment) => {
                 let offset = LogicalVec2 {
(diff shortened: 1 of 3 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -1766,7 +1766,7 @@ impl PlacementState {
                         .into(),
                     inline: Length::new(0.),
                 };
-                fragment.borrow_mut().adjust_offsets(offset);
+                fragment.borrow_mut().adjust_offsets(offset.into());
```
mrobinson, line 1769: The offset is created here by converting a `Length` into `Au`, so you could probably avoid the `.into()` here by simply creating a `LogicalVec<Au>` above.

mrobinson (approved):
Looks good to me with one change...

Issue #9222, opened 2016-01-09
Move Epoch type to gfx_traits


jdm:
It's defined in msg but used in a bunch of places that all depend on gfx_traits, and it's a gfx-related concept.

Code: `components/msg/compositor_msg.rs`, `components/gfx_traits/lib.rs`

PR #32296, opened 2024-05-15
Move non-gfx things out of `gfx_traits` and create a `base` crate

mrobinson:
For a long time, `gfx_traits` has held a lot of things unrelated to graphics
and also unrelated to the `gfx` crate (which is mostly about fonts).
This is a cleanup which does a few things:

1. Move non `gfx` crate things out of `gfx_traits`. This is important in
   order to prevent dependency cycles with a different integration between
   layout, script, and fonts.
2. Rename the `msg` crate to `base`. It didn't really contain anything
   to do with messages and instead mostly holds ids, which are used
   across many different crates in Servo. This new crate will hold the
   *rare* data types that are widely used.

Details:

 - All BackgroundHangMonitor-related things from base to a new
   `background_hang_monitor_api` crate.
 - Moved `TraversalDirection` to `script_traits`
 - Moved `Epoch`-related things from `gfx_traits` to `base`.
 - Moved `PrintTree` to base. This should be widely useful in Servo.
 - Moved `WebrenderApi` from `base` to `webrender_traits` and renamed it
   to `WebRenderFontApi`.

---

- [x] These changes do not require tests because they just move data structures to different crates and should not change behavior.


mrobinson:
Note that `Epoch` was moved to `gfx_traits` in #9222, but it's not really related to anything in the `gfx` crate and is more than just graphics related. It encapsulates both the WebRender and Layout epoch, essentially the heartbeat of a `Document`. It's definitely not font-related and `gfx` will soon become `fonts` since it's 99% fonts.

mrobinson:
This looks like a large change, but is just moving code around.

atbrakhi:
I think we also need to update `ports/jniapi/`

mrobinson:
> I think we also need to update `ports/jniapi/`

I think the Android build should be fixed now.

atbrakhi (approved):
Nice, looks good.

Issue #32297, opened 2024-05-15
Error scopes should be handled in device timeline (wgpu thread)


sagudev:
Currently they are handled in script thread, which causes problems on some tests now that we have parallel poller. In tests [webgpu:shader,execution,expression,constructor,*](https://gpuweb.github.io/cts/standalone/?q=webgpu:shader,execution,expression,constructor,*) we get very flaky results (sometimes passing test, although we shouldn't), because in [`attemptEndTestScope`](https://github.com/gpuweb/cts/blob/e7c4bf2bb6108b0fddd884e43ff68087ba3106f3/src/webgpu/util/device_pool.ts#L364) when popping validation error scope, promise does not get resolved to error (we have yet to handle validation error in `handle_msg_from_webgpu_server`).

SPEC: https://www.w3.org/TR/webgpu/#dom-gpudevice-poperrorscope

sagudev:
Given that all [gpu errors](https://www.w3.org/TR/webgpu/#gpuerror) are sent to script thread, we could just send poperrorscope to script thread to so it gets picked up from `handle_msg_from_webgpu_server` after all validation errors that needed to be in that scope are handled, but then we would be sending to same thread (I think) and we do not want that.

So if we move most error scopes handling to wgpu thread, we use ordering of messages to ensure error scopes contain all their messages. I think this way we can also reduce amount of messages send back to script thread (although https://www.w3.org/TR/webgpu/#gpuuncapturederrorevent still exists). This would also align us with specs.

Unresolved questions:
- callbacks returning gpu errors?
- validation errors from script thread (we can always just add new msg for it, but are they even needed as all validation **should** (but currently does not) happen in wgpu-core)

sagudev:
> validation errors from script thread (we can always just add new msg for it, but are they even needed as all validation should (but currently does not) happen in wgpu-core)

Validation errors should indeed be raised from wgpu-core, but JS errors (Operation, TypeError) are actually in our domain, so not all validation can be moved to wgpu-core ...

sagudev:
Done in  #32304

PR #30504, opened 2023-10-05
[WIP] Sync WebGPU specs to latest

sagudev:
When doing wgpu bump I did it with "minimal" changes to webidl, so this PR tries to catch up with specs more completely.

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #___ (GitHub issue number if applicable)

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #32304, opened 2024-05-17
webgpu: Move errorscopes to WGPU thread

sagudev:
This should lower amount of messages being set to script (no need to send ErrorScopeId and removed useless WebGPUOps::Success), simplified error scope logic and it aligned us with spec (errors scopes should happen in device timeline = WGPU thread). Also includes GPUError stuff from https://github.com/servo/servo/pull/30504 (see https://github.com/servo/servo/pull/32304/commits/dd78013176c2644b3ffe359270a2c4e2d536d51b).

https://sagudev.github.io/briefcase/WebGPUerrorScopes.html works now as expected.

Some test are now FAIL, because CTS does only check for outstanding errors in error scopes, but error scopes were empty because of bad impl.

---

- [x] These changes fix https://github.com/servo/servo/issues/32297

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -56,7 +56,7 @@ use script_traits::{
 };
 use servo_url::{ImmutableOrigin, MutableOrigin, ServoUrl};
 use uuid::Uuid;
-use webgpu::{ErrorScopeId, WebGPUDevice, WebGPUOpResult};
+use webgpu::WebGPUDevice;
 
 use super::bindings::trace::HashMapTracedValues;
 use crate::dom::bindings::cell::{DomRefCell, RefMut};
@@ -3097,17 +3097,12 @@ impl GlobalScope {
         let _ = self.gpu_devices.borrow_mut().remove(&device);
     }
 
-    pub fn handle_wgpu_msg(
-        &self,
-        device: WebGPUDevice,
-        scope: Option<ErrorScopeId>,
-        result: WebGPUOpResult,
-    ) {
+    pub fn handle_uncaptured_gpu_error(&self, device: WebGPUDevice, error: webgpu::Error) {
         self.gpu_devices
             .borrow()
             .get(&device)
             .expect("GPUDevice not found")
-            .handle_server_msg(scope, result);
+            .fire_uncaptured_error(error);
     }
 
     pub fn handle_gamepad_event(&self, gamepad_event: GamepadEvent) {
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -17,19 +17,21 @@ use webgpu::wgc::id::{BindGroupLayoutId, PipelineLayoutId};
 use webgpu::wgc::{
     binding_model as wgpu_bind, command as wgpu_com, pipeline as wgpu_pipe, resource as wgpu_res,
 };
-use webgpu::{self, wgt, ErrorScopeId, WebGPU, WebGPUOpResult, WebGPURequest};
+use webgpu::{self, wgt, PopError, WebGPU, WebGPURequest, WebGPUResponse, WebGPUResponseResult};
 
 use super::bindings::codegen::UnionTypes::GPUPipelineLayoutOrGPUAutoLayoutMode;
 use super::bindings::error::Fallible;
+use super::gpu::AsyncWGPUListener;
 use super::gpudevicelostinfo::GPUDeviceLostInfo;
 use super::gpusupportedlimits::GPUSupportedLimits;
+use super::types::GPUError;
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::EventBinding::EventInit;
 use crate::dom::bindings::codegen::Bindings::EventTargetBinding::EventTargetMethods;
 use crate::dom::bindings::codegen::Bindings::WebGPUBinding::{
     GPUBindGroupDescriptor, GPUBindGroupLayoutDescriptor, GPUBindingResource, GPUBufferBindingType,
     GPUBufferDescriptor, GPUCommandEncoderDescriptor, GPUComputePipelineDescriptor,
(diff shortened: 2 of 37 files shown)


sagudev:
We got handful of FAIL instead of PASS: https://github.com/sagudev/servo/actions/runs/9148458722/job/25151468770.

gterzian (commented):
Will review tomorrow...

components/script/dom/globalscope.rs
```diff
@@ -3097,17 +3097,12 @@ impl GlobalScope {
...
-        scope: Option<ErrorScopeId>,
-        result: WebGPUOpResult,
-    ) {
+    pub fn handle_uncaptured_gpu_error(&self, device: WebGPUDevice, error: webgpu::Error) {
         self.gpu_devices
             .borrow()
             .get(&device)
             .expect("GPUDevice not found")
-            .handle_server_msg(scope, result);
+            .fire_uncaptured_error(error);
```
gterzian, line 3105: Are we sure the device is never dropped at this point?

components/script/dom/gpuqueue.rs
```diff
@@ -81,26 +81,23 @@ impl GPUQueueMethods for GPUQueue {
                 .iter()
                 .all(|b| matches!(b.state(), GPUBufferState::Unmapped))
         });
-        let scope_id = self.device.borrow().as_ref().unwrap().use_current_scope();
         if !valid {
-            self.device.borrow().as_ref().unwrap().handle_server_msg(
-                scope_id,
-                WebGPUOpResult::ValidationError(String::from(
+            self.device
```
gterzian, line 85: I understand why you do this here, but per the spec this also belongs in the device timeline. Also, `valid` seems to be about more than just the unmapped state?

components/webgpu/gpu_error.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::wgc;
+
+/// <https://www.w3.org/TR/webgpu/#gpu-error-scope>
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub(crate) struct ErrorScope {
+    // we only store first error
+    pub errors: Option<Error>,
```
gterzian, line 17: Should we not make this a list, and just return "any" when popped, with "any" being just the first one for now?

components/script/dom/gpudevice.rs
```diff
@@ -64,33 +66,9 @@ use crate::dom::gpushadermodule::GPUShaderModule;
...
-
-#[derive(JSTraceable, MallocSizeOf)]
-struct ScopeContext {
-    error_scopes: HashMap<ErrorScopeId, ErrorScopeInfo>,
-    scope_stack: Vec<ErrorScopeMetadata>,
-    next_scope_id: ErrorScopeId,
-}
-
 #[dom_struct]
 pub struct GPUDevice {
```
gterzian, line 73: Should we not keep some additional state about the device having been lost, to properly silence any errors(while still perhaps running a few last steps), see https://www.w3.org/TR/webgpu/#errors-and-debugging.

For example, I think a ongoing map async step on the device timeline can dispatch an error even after the device is lost, and that error should be silenced(while the map async should still complete, see https://www.w3.org/TR/webgpu/#lose-the-device)

gterzian (commented):
Overall it looks good to me, but I wonder if we need the `WebGPURequest::DispatchError`, because it seems to me that those errors are always dispatched from the device timeline. 

Couple of other questions.

sagudev:
> Overall it looks good to me, but I wonder if we need the `WebGPURequest::DispatchError`, because it seems to me that those errors are always dispatched from the device timeline.
> 
> Couple of other questions.

For this PR I limited myself to only moving error scopes and respec (fresh implementation of latest spec) of GPUErrors, so `WebGPURequest::DispatchError` is a hacky way to keep old code working, but this would be removed in the future as I progressed with the respec in the future (in some parts we are completely out of spec, so full rewrite will be needed).

components/script/dom/globalscope.rs
sagudev: This is how it was done before, but in WGPU we have devices hashmap, when device is removed from it we do not dispatch any error. (we also send CleanDevice to script thread so it gets removed from gpu_devices too). But currently device is lost is not impl properly in most parts of servo (but so it's not in firefox), so this is yet something I would leave for the future work.

components/webgpu/gpu_error.rs
sagudev: That would be per spec, but why would we store multiple errors if we only use one?

components/script/dom/gpudevice.rs
sagudev: There is a lot of work to be done in servo for proper WebGPU support, but this PR is limited only to error scopes, as device lost is currently very loosely implemented (there is lost method but not much else handling), but I think I will do lost device stuff next.

components/webgpu/gpu_error.rs
gterzian: Because per the spec you should store them all but only use one anyway, the only difference is how that one is chosen. In the current implementation we could do something simple like always choosing the  last or the first, and add a comment about this at the point where this is done.

gterzian (approved):
LGTM, I would only make one change which is the one described at https://github.com/servo/servo/pull/32304#discussion_r1610123206

Also, if we don't have them already, it would be good to open issues for subsequent work, like the device lost stuff.

components/webgpu/gpu_error.rs
sagudev: I will change to vec, but generally the first error is the right one, although spec is very vague about that:
> Let error be any one of the items in scope.[[[errors]]](https://www.w3.org/TR/webgpu/#dom-gpu-error-scope-errors-slot)

last one can be from previous operation failing:
> For any two errors E1 and E2 in the list, if E2 was caused by E1, E2 should not be the one selected.

PR #32309, opened 2024-05-17
tables: Add a naive implementation of `border-collapse`

mrobinson:
This change adds a very simple implementation of `border-collapse` for
tables. No harmonization or merging is done at all for borders. Instead,
the largest border for every continuous border sets the size. Instead of
merging different border styles, they are squashed to half size -- which
isn't great, but ensures appropriate positioning.

---


components/layout_2020/table/layout.rs
```diff
@@ -1661,6 +1687,72 @@ impl<'a> TableLayout<'a> {
...
+        let Some(ref collapsed_borders) = self.collapsed_borders else {
+            return None;
+        };
+
+        let end_x = coordinates.x + cell.colspan;
+        let end_y = coordinates.y + cell.rowspan;
+        let mut result: LogicalSides<Length> = LogicalSides {
+            inline_start: collapsed_borders.inline[coordinates.x],
+            inline_end: collapsed_borders.inline[end_x],
+            block_start: collapsed_borders.block[coordinates.y + 1],
```
Loirooriol, line 1736: The `+1` seems wrong.

mrobinson: Yep. I've corrected this in the latest version.

Loirooriol (approved):

PR #32316, opened 2024-05-20
script: Have `Window` own `Layout`

mrobinson:
Have `Window` own `Layout`. This makes it impossible to have a
`Window` without `Layout`, which was true, but now the compiler checks
it. In addition, `Layout` is now released when the `Window` is,
avoiding leaking the entire `Layout`.

---

- [x] These changes do not require tests because they just fix a memory leak.

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -73,9 +73,7 @@ use parking_lot::Mutex;
 use percent_encoding::percent_decode;
 use profile_traits::mem::{self as profile_mem, OpaqueSender, ReportsChan};
 use profile_traits::time::{self as profile_time, profile, ProfilerCategory};
-use script_layout_interface::{
-    Layout, LayoutConfig, LayoutFactory, ReflowGoal, ScriptThreadFactory,
-};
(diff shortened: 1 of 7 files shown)


components/script/script_thread.rs
```diff
@@ -2856,13 +2836,11 @@ impl ScriptThread {
...
         reports.extend(unsafe { get_reports(*self.get_cx(), path_seg) });
-        SCRIPT_THREAD_ROOT.with(|root| {
-            let script_thread = unsafe { &*root.get().unwrap() };
-            let layouts = script_thread.layouts.borrow();
-            for layout in layouts.values() {
-                layout.collect_reports(&mut reports);
-            }
-        });
+
+        for document in documents.iter() {
```
jdm, line 2840:
```suggestion
        for (_, document) in documents.iter() {
```

components/script/script_thread.rs
```diff
@@ -2856,13 +2836,11 @@ impl ScriptThread {
...
+            document.1.window().layout().collect_reports(&mut reports);
```
jdm, line 2841:
```suggestion
            document.window().layout().collect_reports(&mut reports);
```

jdm (approved):

mrobinson:
Thanks for the review!

PR #32318, opened 2024-05-20
fonts: Depend directly on `freetype-sys`

mrobinson:
Instead of depending on `rust-freetype`, depend directly on
`freetype-sys` which is a transitive dependency. This provides almost
everything we need (apart from one function call). This will help us
eliminate one crate in the dependency chain.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2026,7 +2026,7 @@ dependencies = [
  "euclid",
  "fnv",
  "fontsan",
- "freetype",
+ "freetype-sys",
  "gfx_traits",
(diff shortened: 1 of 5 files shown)


mukilan (approved):

nicoburns (approved):
Seems reasonable.

There's a lot of unsafe code in this module though. Medium term we should consider switching all of the freetype code to [skrifa](https://github.com/googlefonts/fontations/tree/main/skrifa) which is pure-rust alternative which is being developed by the Google Fonts team as a replacement for freetype in Chromium.

PR #32330, opened 2024-05-21
ci: enable sccache for Rust compilation.

mukilan:
This was previously disabled in #30508 due to sccache not working well with crown. The sccache issue (mozilla/sccache#861) linked in that PR is now closed and [testing][1] on my fork also seems to indicated we should be able to turn on sccache again.

[1]: https://github.com/mukilan/servo/actions/runs/9154196647

---
- [x] These changes do not require tests because they only modify CI build configuration.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -18,6 +18,7 @@ env:
   RUST_BACKTRACE: 1
   SHELL: /bin/bash
   SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
   CCACHE: "sccache"
   CARGO_INCREMENTAL: 0
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -60,6 +60,7 @@ env:
   RUST_BACKTRACE: 1
   SHELL: /bin/bash
   SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
   CCACHE: "sccache"
   CARGO_INCREMENTAL: 0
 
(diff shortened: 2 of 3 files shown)


sagudev (approved):
I completely forgot about it (and I helped to fix sccache, but it wasn't released yet)

IIRC sccache wasn't working on windows before, but maybe it works now? We could try it in followup.

nicoburns:
Out of interest, where is the cache data stored?

sagudev:
> Out of interest, where is the cache data stored?

GitHub runners have special cached artifacts that can be used across runs: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

nicoburns:
Do we not run into this:

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB. Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the oldest caches in the repository.

Most Rust repos I'm involved in are using https://github.com/Swatinem/rust-cache (sccache may well be better for Servo's use case), but my Servo target directory very quickly exceeds 10GB and I would imagine this would happen especially quickly with multiple PRs compiling different versions of the code. Perhaps using artifacts rather than the built-in caching works around this?

mukilan:
> my Servo target directory very quickly exceeds 10GB and I would imagine this would happen especially quickly with multiple PRs compiling different versions of the code.

From what I understand, sscache caches each compilation artifact individually, as separate entries in the Action cache, so hopefully there is still some sharing even in the presence of multiple versions of code.

That said, [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) or artifact based cache is also something we can explore. rust-cache seems to support caching the cargo registry which could be complimentary to `sccache`.

PR #32332, opened 2024-05-21
fonts: Clean up messaging during web fonts loads

mrobinson:
Instead of sending a message to the script thread via IPC when a web
font loads and then sending another, just give the `FontContext` a
callback that send a single message to the script thread. This moves all
the cache invalidation internally into `FontContext` as well.

Additionally, the unused LayoutControlMessage::ExitNow enum variant is
removed.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2022,6 +2022,7 @@ dependencies = [
  "core-foundation",
  "core-graphics",
  "core-text",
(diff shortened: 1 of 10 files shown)


mukilan (approved):

mrobinson:
Intermittent failure: #32343. Sending back to the MQ.

PR #32339, opened 2024-05-22
Move webgl_channel into base crate

wusyong:
WebGL channels are interesting because they can select different channel implementations in runtime based on multiprocess or not.
There's a [Zulip discussion](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/What.20are.20other.20processes.20if.20multiprocess.20is.20enabled.3F) around multiprocess, and I feel like this is a potential improvement if other places using `ipc::channel` could also use this.

I named this crate `process` because I hope it will also add other types related to processes and threads in the future, such as moving net's async runtime to here. So we have better control over how they spawn and communicate. Any name suggestion is welcome.
 
---

- [x] There are tests for these changes in WPT. WebGL tests should be passed.


components/shared/process/channel.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub type Sender<T> = ipc_channel::ipc::IpcSender<T>;
+    pub type Receiver<T> = ipc_channel::ipc::IpcReceiver<T>;
+
+    pub fn channel<T: Serialize + for<'de> Deserialize<'de>>(
+    ) -> Result<(Sender<T>, Receiver<T>), io::Error> {
+        ipc_channel::ipc::channel()
+    }
+}
+
+mod mpmc {
```
mrobinson, line 126: You renamed this from `mpsc` to `mpmc`. :) Let's remove this module though. and use the crossbeam channel directly in the enum instead of hiding these behind a module.

components/shared/process/channel.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Debug)]
+pub struct ReceiveError;
+pub type ReceiveResult<T> = Result<T, ReceiveError>;
+
+pub enum Receiver<T>
+where
+    T: for<'de> Deserialize<'de> + Serialize,
+{
+    Ipc(ipc::Receiver<T>),
```
mrobinson, line 59: Just use ipc_channel directly:

```suggestion
    Ipc(ipc_channel::ipc::IpcReceiver<T>),
```

components/shared/process/channel.rs
```diff
@@ -0,0 +1,185 @@
...
+    Mpmc(mpmc::Receiver<T>),
```
mrobinson, line 60:
```suggestion
    CrossBeam(crossbeam_channel::Receiver<T>),
```

components/shared/process/channel.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Debug)]
+pub struct SendError;
+pub type SendResult = Result<(), SendError>;
+
+#[derive(Debug)]
+pub struct ReceiveError;
+pub type ReceiveResult<T> = Result<T, ReceiveError>;
+
+pub enum Receiver<T>
```
mrobinson, line 55: Let's call this `GenericReceiver` and `GenericSender` in order to make it clear in the code what these types are and to differentiate between `crossbeam_channel::Sender` and  `crossbeam_channel::Receiver`. Also this will avoid having to rename the type when they are imported into a file that uses both.

mrobinson (changes requested):
This is a nice start. This generic sender seems like a very basic type, unlikely to change and widely useful. Can you move this to `components/base/generic_channel.rs` instead of making a new crate? `process` is a bit odd, because crossbeam channels don't have anything to do with processes.

In addition, please use this new data structure to replace `OpaqueSender` in `components/shared/profile/mem.rs`.

Thanks!

components/shared/process/channel.rs
wusyong: Fixed in [731b64f](https://github.com/servo/servo/pull/32339/commits/731b64fa2fc597d0080905cb1a9e562c571f648b)

components/shared/process/channel.rs
wusyong: Fixed in [731b64f](https://github.com/servo/servo/pull/32339/commits/731b64fa2fc597d0080905cb1a9e562c571f648b)

components/shared/process/channel.rs
wusyong: Fixed in [731b64f](https://github.com/servo/servo/pull/32339/commits/731b64fa2fc597d0080905cb1a9e562c571f648b)

components/shared/process/channel.rs
wusyong: Fixed in [731b64f](https://github.com/servo/servo/pull/32339/commits/731b64fa2fc597d0080905cb1a9e562c571f648b)

wusyong:
@mrobinson I couldn't move it to `base` because `base` and `servo_config` would create a circular dependency.
How about rename to `runtime` or move to `profile?`

mrobinson:
> @mrobinson I couldn't move it to `base` because `base` and `servo_config` would create a circular dependency. How about rename to `runtime` or move to `profile?`

Please move `fn channel()` to the WebGL crate in this case. Even though there is a generic channel implementation it isn't always just the multiprocess option that might be controlling whether or not we want an IPC channel or a crossbeam channel.

wusyong:
> Please move `fn channel()` to the WebGL crate in this case. Even though there is a generic channel implementation it isn't always just the multiprocess option that might be controlling whether or not we want an IPC channel or a crossbeam channel.

Ah I think I can add a parameter to check instead. So we can still have a create channel function in this module.

> In addition, please use this new data structure to replace `OpaqueSender` in `components/shared/profile/mem.rs`.

Do you mean type alias `OpaqueSender` to `GenericSender`? Because ``OpaqueSender` seems to be a trait. There are other places will need to update I think.

mrobinson:
> Do you mean type alias `OpaqueSender` to `GenericSender`? Because ``OpaqueSender` seems to be a trait. There are other places will need to update I think.

Since `OpaqueSender` is solving the same sort of problem, I think you could modify the code that was using it to use `GenericSender` instead.

wusyong:
> Since `OpaqueSender` is solving the same sort of problem, I think you could modify the code that was using it to use `GenericSender` instead.

After looking into it further, I found there are more types passed in profile's [`https://github.com/servo/servo/blob/main/components/shared/profile/mem.rs#L68`](https://github.com/servo/servo/blob/main/components/shared/profile/mem.rs#L68) function. [Some types](https://github.com/servo/servo/blob/main/components/script/dom/abstractworkerglobalscope.rs#L26) are more than just a sender field. I think we still need to keep `OpaqueSender` for these types.

mrobinson (approved):

PR #32346, opened 2024-05-22
fonts: Remove web fonts when their stylesheet is removed

### LINKED ISSUE #15139: We never remove web fonts from the font cache.
This is not only a leak, but also may produce incorrect styling. Consider the following code:

```html
<style>
@font-face {
    name: foo;
    /* ... */
}
</style>
<style>
div {
    font-family: foo, arial;
}
</style>
<div class="foo">
TEXT
</div>
<script>
let s = document.querySelector('style');
s.parentNode.removeChild(s);
</script>
```

We should render the text in Arial, but as far as I know we'd render it with the `foo` font.

I remember to have seen an issue about the font cache leak, but I couldn't find it, so I'm filling this one.

mrobinson:
This is the first part of ensuring that unused fonts do not leak. This
change makes it so that when a stylesheet is removed, the corresponding
web fonts are removed from the `FontContext`.

Note: WebRender assets are still leaked, which was the situation before
for all fonts. A followup change will fix this issue.

Fixes #15139.

---

- [x] These changes fix #15139.

### DIFF
--- a/components/gfx/font_store.rs
+++ b/components/gfx/font_store.rs
@@ -8,21 +8,95 @@ use std::sync::Arc;
 use app_units::Au;
 use atomic_refcell::AtomicRefCell;
 use parking_lot::RwLock;
+use style::stylesheets::DocumentStyleSheet;
 use webrender_api::{FontInstanceFlags, FontInstanceKey, FontKey};
 
 use crate::font::FontDescriptor;
 use crate::font_cache_thread::{FontIdentifier, FontSource, LowercaseString};
+use crate::font_context::WebFontDownloadState;
 use crate::font_template::{FontTemplate, FontTemplateRef, FontTemplateRefMethods};
 
 #[derive(Default)]
 pub struct FontStore {
     pub(crate) families: HashMap<LowercaseString, FontTemplates>,
+    web_fonts_loading: Vec<(DocumentStyleSheet, usize)>,
 }
 pub(crate) type CrossThreadFontStore = Arc<RwLock<FontStore>>;
 
 impl FontStore {
     pub(crate) fn clear(&mut self) {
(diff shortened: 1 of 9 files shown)


mrobinson:
I have added a test for this change.

components/gfx/font_store.rs
```diff
@@ -8,22 +8,96 @@ use std::sync::Arc;
...
+        new_template: FontTemplate,
+    ) -> bool {
+        // Abort processing this web font if the originating stylesheet was removed.
+        if self.font_load_cancelled_for_stylesheet(&state.stylesheet) {
+            return false;
+        }
+
+        let family_name = state.css_font_face_descriptors.family_name.clone();
+        self.families
+            .entry(family_name.clone())
```
mukilan, line 91:
```suggestion
            .entry(family_name))
```

components/gfx/font_store.rs
```diff
@@ -127,4 +201,14 @@ impl FontTemplates {
         self.templates
             .push(Arc::new(AtomicRefCell::new(new_template)));
     }
+
+    pub(crate) fn remove_add_templates_for_stylesheet(
```
mukilan, line 205:
```suggestion
    pub(crate) fn remove_templates_for_stylesheet(
```

mukilan (approved):

components/gfx/font_store.rs
mrobinson: :sweat_smile: Fixed this one.

components/gfx/font_store.rs
mrobinson: Fixed this as well as a couple other places where we didn't need to pass as a reference.

PR #32350, opened 2024-05-23
fonts: Fix rendering of emoji clusters

mrobinson:
Emoji clusters, such ('🏳️‍🌈') do not render properly in Servo.
This is because `xi-unicode` is inserting a linebreak opportunity between
components of the cluster (see xi-editor/xi-editor#1322). This change
adds a workaround for this issue.

`xi-unicode` is fast, but supports an older version of the Unicode
standard than libraries like `icu4x`. In addition, `icu4x` does not
support non-contiguous segmentation which Servo currently depends on.
Finally, the currently released version of `icu4x` has the same issue
(unicode-org/icu4x#4146).

---

### DIFF
--- a/components/gfx/text/text_run.rs
+++ b/components/gfx/text/text_run.rs
@@ -13,6 +13,7 @@ use range::Range;
 use serde::{Deserialize, Serialize};
 use style::str::char_is_whitespace;
 use unicode_bidi as bidi;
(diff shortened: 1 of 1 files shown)


mrobinson:
This is a workaround, but @atbrakhi and I are working on a more comprehensive fix that drops the use of `xi-unicode` which isn't really maintained any longer.

PR #32377, opened 2024-05-24
Remove more IPC messages between script and layout

mrobinson:
Instead of bouncing messages from the compositor to script and then to
layout, just have script call methods on Layout. Additionally, we do not
need to send any followup messages to script for these messages. Instead
just execute code after calling the method on Layout.

---

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -26,8 +26,8 @@ use std::rc::Rc;
 use std::result::Result;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::thread;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
-use std::{ptr, thread};
 
 use background_hang_monitor_api::{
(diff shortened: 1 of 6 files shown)


components/script/script_thread.rs
```diff
@@ -2348,30 +2344,60 @@ impl ScriptThread {
...
-    fn handle_layout_message_from_constellation(
+    fn handle_set_scroll_states_msg(
         &self,
-        msg: LayoutControlMsg,
         pipeline_id: PipelineId,
+        scroll_states: Vec<ScrollState>,
     ) {
         let Some(window) = self.documents.borrow().find_window(pipeline_id) else {
-            warn!("Received layout message pipeline {pipeline_id} closed: {msg:?}.");
+            warn!("Received scroll stetes for closed pipeline {pipeline_id}");
```
mukilan, line 2369:
```suggestion
            warn!("Received scroll states for closed pipeline {pipeline_id}");
```

mukilan (approved):

PR #32378, opened 2024-05-24
dependabot: Group all GStreamer dependencies together for upgrades

mrobinson:
The GStreamer we use is a bracing 24 crates, which tends to flood our CI
infrastructure. Group them together in order to make things more
manageable.

---

- [x] These changes do not require tests because they just modify the dependabot configuration.


jdm (approved):
Oh thank goodness!

PR #32388, opened 2024-05-28
layout: Add support for `white-space-collapse: break-spaces`

mrobinson:
This change adds support for `white-space-collapse: break-spaces` and
adds initial parsing support for `overflow-wrap` and `word-break`. The
later two properties are not fully supported, only in their interaction
with `break-spaces`. This is a preliminary change preparing to implement
them.

In addition, `break_and_shape` is now forked and added to Layout 2020.
This function is going to change a lot soon and forking is preparation
for this. More code that is only used by Layout 2013 is moved from `gfx`
to that crate.

---


mrobinson:
I think this is passing tests, and the failure here was just an infrastructure issue. @atbrakhi if this looks good to you, I can land this and the style change as wel.

atbrakhi (approved):

PR #32410, opened 2024-05-31
webgpu: Parse and forward backend prefs to wgpu

### LINKED ISSUE #32390: Add pref for wgpu backend
WGPU has env vars: https://github.com/gfx-rs/wgpu?tab=readme-ov-file#environment-variables

Firefox has these prefs:
slika

We need at least one for backend selection.

Steps:
Add `dom.webgpu.wgpu-backend` pref to https://github.com/servo/servo/blob/42061ccafe67ac0440886e48e17a02f8b60525bc/components/config/prefs.rs#L206
Use https://docs.rs/wgpu-core/latest/wgpu_core/instance/fn.parse_backends_from_comma_list.html to parse pref value and put it in https://github.com/servo/servo/blob/42061ccafe67ac0440886e48e17a02f8b60525bc/components/webgpu/wgpu_thread.rs#L94

Testing:
Testing can be done by enabling logging in wgpu with `RUST_LOG=wgpu` and requested adapter should match backend that is set by the pref.

Here is demo webgpu site: https://demo.servo.org/experiments/webgpu-game-of-life/

`dom.webgpu.enabled` should also be enabled when testing.

hieudo-dev:
These changes enable wgpu backend selection by defining a new pref. The new pref `dom.webgpu.wgpu_backend` is parsed using `wgpu-core` and forwarded to the wgpu instance.

---

- [x] These changes fix #32390

### DIFF
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -206,6 +206,8 @@ mod gen {
                 webgpu: {
                     /// Enable WebGPU APIs.
                     enabled: bool,
+                    /// List of comma-separated backends to be used by wgpu
+                    wgpu_backend: String,
                 },
                 bluetooth: {
                     enabled: bool,
--- a/components/webgpu/wgpu_thread.rs
+++ b/components/webgpu/wgpu_thread.rs
@@ -13,7 +13,8 @@ use arrayvec::ArrayVec;
 use base::id::PipelineId;
 use euclid::default::Size2D;
 use ipc_channel::ipc::{IpcReceiver, IpcSender, IpcSharedMemory};
-use log::{error, warn};
+use log::{error, info, warn};
+use servo_config::pref;
 use webrender::{RenderApi, RenderApiSender, Transaction};
 use webrender_api::{DirtyRect, DocumentId};
 use webrender_traits::{WebrenderExternalImageRegistry, WebrenderImageHandlerType};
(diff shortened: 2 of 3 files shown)


components/config/prefs.rs
```diff
@@ -205,6 +205,7 @@ mod gen {
             dom: {
                 webgpu: {
                     /// Enable WebGPU APIs.
+                    wgpu_backend: String,
                     enabled: bool,
```
sagudev, line 209: `/// Enable WebGPU APIs.` should remain at `enabled: bool`, and `wgpu_backend: String` needs to get own doc comment.

components/webgpu/wgpu_thread.rs
```diff
@@ -91,7 +93,7 @@ impl WGPU {
         let global = Arc::new(wgc::global::Global::new(
             "wgpu-core",
             InstanceDescriptor {
-                backends: wgt::Backends::PRIMARY,
+                backends: parse_backends_from_comma_list(&pref!(dom.webgpu.wgpu_backend)),
```
sagudev, line 96: I just remembered that we need to handle empty string on our own (as it is default value), so in case of empty strung we need to return `wgt::Backends::PRIMARY`.

Se relevant firefox code for reference (they complicate with cfgs for no reason): https://searchfox.org/mozilla-central/source/gfx/wgpu_bindings/src/server.rs#109

sagudev (changes requested):

hieudo-dev:
@sagudev I have updated the PR based on your feedback. Pls let me know if you see any remaining issues.

sagudev:
I will need to do manual test on windows not on linux because we do not have gles backend enabled in wgpu :(. But otherwise it works.

sagudev (approved):

sagudev:
Thank you for all the work!

hieudo-dev:
> Thank you for all the work!

No probs! It's my pleasure hah

PR #32415, opened 2024-05-31
layout: Add a `InlineFormattingContextBuilder`

mrobinson:
The main change here is that collapsed and `text-transform`'d text is
computed as it's processed by DOM traversal. This single transformed
text is stored in the root of the `InlineFormattingContext`.

This will eventually allow performing linebreaking and shaping of the
entire inline formatting context at once. Allowing for intelligent
processing of linebreaking and also shaping across elements. This
matches more closely what LayoutNG does.

This shouldn't have any (or negligable) behavioral changes, but will
allow us to prevent linebreaking inside of clusters in a followup
change.

---


mrobinson:
I think this is ready to go after a few minor fixes. In the end, two tests have started passing -- likely due to collapsing white space more reliably.

components/layout_2020/flow/construct.rs
```diff
@@ -531,54 +469,25 @@ where
         contents: Contents,
         box_slot: BoxSlot<'dom>,
     ) {
-        // We just found a block level element, all ongoing inline level boxes
-        // need to be split around it. We iterate on the fragmented inline
-        // level box stack to take their contents and set their first_fragment
-        // field to false, for the fragmented inline level boxes that will
+        // We just found a block level element, all ongoing inline level boxes need to be split
+        // around it. We iterate on the fragmented inline level box stack to take their contents and
+        // set their first_fragment field to false, for the fragmented inline level boxes that will
```
atbrakhi, line 474: maybe we can remove this now as we don't have `fragmented_inline_boxes` and neither do we iterate on the fragment inline I think

components/layout_2020/flow/construct.rs
```diff
@@ -531,54 +469,25 @@ where
...
-                    .children
-                    .push(ArcRefCell::new(fragmented_inline));
-                fragmented_inline = InlineLevelBox::InlineBox(fragmented_parent_inline_box);
-            }
-
-            self.ongoing_inline_formatting_context
-                .inline_level_boxes
-                .push(ArcRefCell::new(fragmented_inline));
+        //
+        // After calling `split_around_block_and_finish_existing`, `self.inline_formatting_context_builder`
```
atbrakhi, line 477:
```suggestion
        // After calling `split_around_block_and_finish`, `self.inline_formatting_context_builder`
```

atbrakhi (approved):
Left some very small comments, but looks good.

PR #32416, opened 2024-05-31
Refactor(bluetooth): replace IPC channel with generic channel in BluetoothManager

### LINKED ISSUE #32411: [bluetooth] Replace IPC channel with GenericChannel
**Describe the new feature:**
With #32339, Servo now has a [generic channel](https://github.com/servo/servo/blob/main/components/shared/base/generic_channel.rs) that can select different channel implementation during runtime. This can help servo to use ipc channel if it's in multiprocess mode, and crossbeam channel if not.

There are lost of places that `ipc::channel` is used. Let's start with Bluetooth crate first since it only has two calls:

- [ ] Replace [`impl BluetoothThreadFactory for IpcSender<BluetoothRequest>`](https://github.com/servo/servo/blob/main/components/bluetooth/lib.rs#L76) with `impl BluetoothThreadFactory for GenericSender<BluetoothRequest>` instead
- [ ] Replace [`GetSelectedBluetoothDevice(Vec<String>, IpcSender<Option<String>>)`](https://github.com/servo/servo/blob/main/components/shared/embedder/lib.rs#L192) with `GetSelectedBluetoothDevice(Vec<String>, GenericSender<Option<String>>)`

Feel free to take one of them above and open the PRs. Or if you are interested in more other crates, open more issues like this. I'll also open some more in the future.

pewsheen:
From #32411

- Replace `impl BluetoothThreadFactory for IpcSender<BluetoothRequest>` with `impl BluetoothThreadFactory for GenericSender<BluetoothRequest>`
- In `ConstellationState` and `InitialConstellationState`
  - Replace member `bluetooth_ipc_sender` with `bluetooth_sender`

---

- [ ] These changes fix #32411

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,6 +464,7 @@ dependencies = [
 name = "bluetooth"
 version = "0.0.1"
 dependencies = [
+ "base",
  "bitflags 2.5.0",
  "bluetooth_traits",
  "blurdroid",
--- a/components/bluetooth/Cargo.toml
+++ b/components/bluetooth/Cargo.toml
@@ -11,6 +11,7 @@ name = "bluetooth"
 path = "lib.rs"
 
 [dependencies]
+base = { workspace = true }
 bitflags = { workspace = true }
 bluetooth_traits = { workspace = true }
 blurmock = { version = "0.1.2", optional = true }
--- a/components/bluetooth/lib.rs
+++ b/components/bluetooth/lib.rs
@@ -19,6 +19,7 @@ use std::string::String;
 use std::thread;
 use std::time::Duration;
 
+use base::generic_channel::{GenericReceiver, GenericSender};
 use bitflags::bitflags;
 use bluetooth_traits::blocklist::{uuid_is_blocklisted, Blocklist};
 use bluetooth_traits::scanfilter::{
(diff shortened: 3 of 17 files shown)


wusyong (commented):
This PR only fix first half of the issue. Perhaps the title should be replace IPC channel with generic channel in BluetoothManager.

mrobinson (commented):
I think perhaps we shouldn't use the `GenericChannel` until we need a channel to be either IPC or crossbeam. If that's not the case then it's just an abstraction that makes it a bit harder to understand the code.

mrobinson:
That said: If there are some benefits to always using crossbeam channels while in single-process mode, I would be happy to change my mind.

wusyong:
I'm okay with these. My purpose for this PR and related issues is to create beginner issues.
But I still think it's a performance loss when we use ipc_channel in single process mode.
Perhaps we should enable multiprocess mode by default to actually test it?

mrobinson:
One of the issues with the generic channel implementation is that there's a difference between what things can be sent over a crossbeam channel and what can be sent over an IPC channel. In particular, things sent over IPC channels must be serializable / deserializable. Making it generic in all cases might mean making data structures serializable when they don't actually need to be. I think we should use the generic channel in places that it makes sense -- when there's a good reason we want to make it more abstract, but not in every case. I'll close this for now.

PR #32422, opened 2024-06-01
Android: fix url resolution

### LINKED ISSUE #21948: Android's address bar is confused by localhost
I first run:

```
./mach run --android --release http://localhost:8000/examples/webgl_animation_cloth.html
```

Then I tap on the address bar and press enter.

Servo then tried to load that same URL but with domain `www.localhost.com` and fails.

Gae24:
The behavior is a bit different from before, previously any string with a . was considered a url, but I don't think is always a correct assumption.
Regarding localhost urls, only http scheme is accepted for now, should https also be accepted? 

---

- [X] These changes fix #21948 (GitHub issue number if applicable)
- [X] These changes do not require tests because no new features.

### DIFF
--- a/ports/jniapi/src/simpleservo.rs
+++ b/ports/jniapi/src/simpleservo.rs
@@ -30,6 +30,7 @@ use servo::embedder_traits::{
 };
 use servo::euclid::{Point2D, Rect, Scale, Size2D, Vector2D};
 use servo::keyboard_types::{Key, KeyState, KeyboardEvent};
+use servo::net_traits::pub_domains::is_reg_domain;
 pub use servo::script_traits::{MediaSessionActionType, MouseButton};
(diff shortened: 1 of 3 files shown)


Gae24:
Found this issue #21304, wouldn't it better to send the url field string to the rust side and parse it like in minibrowser?

ports/jniapi/src/simpleservo.rs
```diff
@@ -379,9 +380,17 @@ impl ServoGlue {
     }
 
     /// Load an URL. This needs to be a valid url.
```
mrobinson, line 382:
```suggestion
    /// Load an URL. If this is not a valid URL, try to "fix" it by adding a scheme or if all else fails,
    /// interpret the string as a search term.
```

mrobinson (approved):

PR #32423, opened 2024-06-01
fix: remove unused items in gfx

wusyong:
There are some build warnings about unused items on macOS. I suppose these are leftovers from font cache refactoring.

---

- [x] These changes do not require tests because they are unused items.


mrobinson (approved):
Looks good We do need to clear this cache, but it's likely that instead this item will be associated with the FontData and will cleared when that is freed.

PR #32424, opened 2024-06-01
base: Remove `ucd` dependency

mrobinson:
Remove the `ucd` dependency which has not been updated in 8 years. In
addition, replace it with a generated UnicodeBlock enum which reflects
the modern Unicode standard. This is generated via a Python script which
is included in the repository. The generation is not part of the build
process, because the Unicode database is hosted on the web and it does
not change the frequently.

This is done instead of bringing in the more up-to-date `unicode_blocks`
dependency. `unicode_blocks` defines each block as constant, which means
that they cannot be used in match statements -- which we do in Servo.

Co-authored-by: Lauryn Menard <lauryn.menard@gmail.com>

---


sagudev:
> This is done instead of bringing in the more up-to-date unicode_blocks
dependency. unicode_blocks defines each block as constant, which means
that they cannot be used in match statements

Maybe we can just make PR to unicode_blocks?

mukilan (approved):

mrobinson:
> > This is done instead of bringing in the more up-to-date unicode_blocks
> > dependency. unicode_blocks defines each block as constant, which means
> > that they cannot be used in match statements
> 
> Maybe we can just make PR to unicode_blocks?

I think this would require a significant change to the way that `unicode_blocks` works. In addition, this crate provides essentially a single enum.

jdm:
Our of curiosity, if it's only used by gfx why add it to base?

mrobinson:
> Our of curiosity, if it's only used by gfx why add it to base?

Good question. The hope is that with this change 'gfx' will now only contain things related to fonts and be renamed to 'fonts' (which was the majority of the crate's contents). Since calculating the Unicode block of a character is a basic text operation it will likely be used outside of the fonts crate. If we add more similar functions here related to text or Unicode, maybe we could consider a separate 'text' or 'unicode'. In the meantime, it will allow renaming 'gfx' to 'fonts' in a more natural way.

PR #32425, opened 2024-06-01
minibrowser: Disables urlbar when in fullscreen

### LINKED ISSUE #30347: minibrowser: URL bar is still visible for pages that use the Fullscreen API
A simple website like this:
```html
<!DOCTYPE html>

<p>Lorem ipsum...</p>

<script>
function toggleFullScreen() {
  if (!document.fullscreenElement) {
    document.documentElement.requestFullscreen();
  } else if (document.exitFullscreen) {
    document.exitFullscreen();
  }
}
</script>

<button onclick="toggleFullScreen();">FullScreen</button>
```

When you click the button, you want the website fullscreen without the URL bar like in any other browser.

The minibrowser still keeps the URL bar and "Go" button visible.

Nylme:
Disables the top url/toolbar when entering fullscreen via: `document.documentElement.requestFullscreen();`

---

- [x] These changes fix #30347 (GitHub issue number if applicable)

[Asking](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Adding.20a.20test.20for.20small.20pr/near/441936183) about tests.

### DIFF
--- a/ports/servoshell/minibrowser.rs
+++ b/ports/servoshell/minibrowser.rs
@@ -171,7 +171,7 @@ impl Minibrowser {
         } = self;
         let widget_fbo = *widget_surface_fbo;
         let _duration = context.run(window, |ctx| {
-            // TODO: While in fullscreen add some way to mitigate the increased phishing risk when not dispalying the URL bar:
+            // TODO: While in fullscreen add some way to mitigate the increased phishing risk when not displaying the URL bar:
             // https://github.com/servo/servo/issues/32443
             // disables drawing the toolbar in fullscreen
             if !window.fullscreen().is_some() {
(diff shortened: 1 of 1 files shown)


atbrakhi (commented):
I think this looks good.

@delan and I were discussing, and we believe hiding the URL bar without any other affordance is a phishing risk. However, we also think this should not stop us from merging this PR, at least for now, partially because we are not using servo in production right now.

I recommend opening an issue and adding it as a TODO so that we can revisit this topic later.

ports/servoshell/minibrowser.rs
```diff
@@ -171,58 +171,65 @@ impl Minibrowser {
...
-                                    location_dirty.set(false);
-                                }
-                            },
-                        );
-                    },
-                );
-            });
+            // TODO: While in fullscreen add some way to mitigate the increased phishing risk when not displaying the URL bar:
+            // https://github.com/servo/servo/issues/32443
+            // disables drawing the toolbar in fullscreen
```
atbrakhi, line 176: I think this comment is not necessary. Thanks for opening the issue and adding the TODO

atbrakhi (commented):
I think you need to run `./mach fmt` as test-tidy is failing.

atbrakhi (approved):
Thank you.

PR #32427, opened 2024-06-02
Implement outerText on HtmlElement

shanehandley:
Implements the IDL attribute / behaviour for `outerText` as well as some basic layout code.

I haven't previously made any changes to anything related to layout code, so I thought I should check to ensure that I am not going in the wrong direction.

Cheers!

---


mrobinson:
@shanehandley Any reason this was closed? I had it on my todo list to review this.

shanehandley:
Hi @mrobinson, apologies for not leaving a comment, I have been tapping away trying to improve the WPT passes for it over the past 2 weeks, but have not made much progress at all 😬.

Though if you are happy to provide feedback, I'd be very grateful 🙏 Thank you!

nicoburns:
@shanehandley General WPT results that are not "Stable unexpected results" can be ignored (certainly the numbers that you are seeing of the other categories are usual)

shanehandley:
> @shanehandley General WPT results that are not "Stable unexpected results" can be ignored (certainly the numbers that you are seeing of the other categories are usual)

Thanks Nico 👍 No worries - I was mostly worried that some of the remaining WPT failures were indicative of flaws in my implementation 😬 

I'll take another look into this soon 🙏

Melchizedek6809:
Hey there, what's the status on this PR, anything you want help with? Would love to see Servo supporting outer-/innerText

components/layout_2020/query.rs
```diff
@@ -525,8 +529,238 @@ fn is_eligible_parent(fragment: &BoxFragment) -> bool {
...
+                // Step 8: If node is a p element, then append 2 (a required line break count) at
+                // the beginning and end of items.
+                results.insert(0, InnerOrOuterTextItem::RequiredLineBreakCount(2));
+                results.push(InnerOrOuterTextItem::RequiredLineBreakCount(2));
+            },
+            _ => {},
+        }
+
+        match display {
+            Display::TableCell if iterator.peek().is_some() => {
```
shanehandley, line 717: the condition `if iterator.peek().is_some()` makes no sense here, we need a fn which determines whether it's the last cell in the row.

components/layout_2020/query.rs
```diff
@@ -525,8 +529,238 @@ fn is_eligible_parent(fragment: &BoxFragment) -> bool {
...
+                // Step 6: If node's computed value of 'display' is 'table-cell', and node's CSS box
+                // is not the last 'table-cell' box of its enclosing 'table-row' box, then append a
+                // string containing a single U+0009 TAB code point to items.
+                results.push(InnerOrOuterTextItem::Text(String::from(
+                    "\u{0009}", /* tab */
+                )));
+            },
+            Display::TableRow if iterator.peek().is_some() => {
```
shanehandley, line 725: Ditto here

shanehandley:
> Hey there, what's the status on this PR, anything you want help with? Would love to see Servo supporting outer-/innerText

Hey - I've been coming back to this every now and then, however I've not been able to improve test passes much at all. I'd be thrilled if you want to help out.

I have since spotted where I have gone wrong in a few places, but failed to find a neat solution. One example is with `display: table-cell`. The spec says:

> If node's [computed value](https://drafts.csswg.org/css-cascade/#computed-value) of ['display'](https://drafts.csswg.org/css2/#display-prop) is ['table-cell'](https://drafts.csswg.org/css-tables/#table-cell), and node's [CSS box](https://drafts.csswg.org/css-display/#css-box) is not the last ['table-cell'](https://drafts.csswg.org/css-tables/#table-cell) box of its enclosing ['table-row'](https://drafts.csswg.org/css-tables/#table-row) box, then [append](https://infra.spec.whatwg.org/#list-append) a string containing a single U+0009 TAB code point to items.

Basically in processing, it takes the content: `<tr><td>foo</td><td>bar</td></tr>` and iterates over elements returning: `\tfoo\tbar` when it should be `foo\tbar` because it just sees the element and adds the content asked by the spec. But it needs to be smarter than this.

There are likely more issues related to white space processing, but if you want to have a crack at these sorts of problems, you are most welcome 🙇

components/layout_2020/query.rs
Melchizedek6809: Maybe we can invert the logic here, instead of emitting the TAB/LF characters after each Cell/Row we instead emit one before the contents if it's not the first Cell/Row in a Table/Row? That way we should get the same results and don't have to peek. 

Might give different results with strange nesting, though I'm not sure what the standards say about the required behaviour there.

Melchizedek6809:
Alright, I've added a comment for an idea regarding the issue with TableCells/Rows that might simplify things somewhat (fingers crossed). I'll probably have some time on the weekend to look at the code more thorougly, hopefully we can get some more WPT tests to pass :)

components/layout_2020/query.rs
shanehandley: I agree, when you read the spec, it's as though the have approached things starting at a text node and moving outwards to the parents, but we approach them from parent to child.

> we instead emit one before the contents if it's not the first Cell/Row in a Table/Row?

This makes sense to me - In the specific case of cells, the logic that seems sensible in my mind:

```
if is_a_text_node() {
   append_text()
   
   if parent_node_is_display_table_cell && _not_the_last_display_table_cell_in_its_row() {
       append_tab_char()
   }
}
```

I've just pushed a very questionable implementation of this, it does get us more passes but we would have to find a way to deal with table rows.

Melchizedek6809:
Hey, started looking into the code a bit more and first off compliments! With this PR Servo seems to pass even more innerText WPT tests than Webkit.

When it comes to the remaining tests I think that iterating over a flattened tree is a limiting factor, for example there is a test where `<canvas><div>asd` needs to return an empty string which seems quite complicated with the current architecture (although we could walk upwards from the current `node`, that does seem very inefficient since we would have to do that for every node we encounter). I'll experiment a bit with that idea to (hopefully) get a working PR out of it.

Melchizedek6809:
@shanehandley Alright, got most of the WPT Tests to pass, things kinda fell into place after iterating over the tree directly. I've opened a PR in your repo, or should I open a separate PR here?

Melchizedek6809:
Hmm that PR somehow ended up in my fork, can't seem to open a PR on your repo directly, I've pushed all the changes in my fork into the implement-outertext repo: https://github.com/Melchizedek6809/servo/tree/implement-outertext

shanehandley:
> @shanehandley Alright, got most of the WPT Tests to pass, things kinda fell into place after iterating over the tree directly. I've opened a PR in your repo, or should I open a separate PR here?

Nice! Yah I think open a new PR 👍

Melchizedek6809:
Alrigh, I'll do that, there are still 2 commits from you that are not signed off, can I just squash them into the first one thats signed off?

shanehandley:
> Alright, I'll do that, there are still 2 commits from you that are not signed off, can I just squash them into the first one thats signed off?

Yeah sure 👍

PR #32429, opened 2024-06-02
Document library crates only in mach doc

### LINKED ISSUE #32412: Docs are not built for libservo

wusyong:
There's a package name conflict between libservo and servoshell, which makes libservo documentation missing.
We can just pass `--lib` to build library crates.

---

- [x] These changes fix #32412

- [x] There are tests for these changes. Docs upload CI should pass and upload successfully.


python/servo/post_build_commands.py
```diff
@@ -246,6 +246,9 @@ def doc(self, params: List[str], **kwargs):
         if not path.exists(docs):
             os.makedirs(docs)
 
+        # Document library crates to avoid package name conlict between severoshell
```
mrobinson, line 249:
```suggestion
        # Document library crates to avoid package name conflict between severoshell
```

mrobinson (approved):
Great comment.

PR #32442, opened 2024-06-05
layout: Don't default to fallback fonts for spaces

mrobinson:
Previously, when deciding the font for a space, preference was given to
the previous used font. This could means that the font chosen was a
fallback font instead of the first font that supporting the space
character in the font preference list.

This caused an issue rendering emojis surrounded by spaces with "Noto
Color Emoji" which has a space character the same size as the emoji,
leading to too much spacing between them.

---


atbrakhi (approved):
I guess this made some tests pass :)

PR #32448, opened 2024-06-06
layout: Take into account `display: table` etc in offset* queries

mrobinson:
The specification says that for deciding whether an element should be
used for offset* queries, a browser should take into account whether the
element is a table cell or table. This change makes that happen.

---

### DIFF
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -90,7 +90,8 @@ where
 
         let threadsafe_node = node.to_threadsafe();
         let mut flags = FragmentFlags::empty();
-        if let Some(element) = threadsafe_node.as_element() {
+
+        if let Some(element) = threadsafe_node.as_html_element() {
             if element.is_body_element_of_html_element_root() {
                 flags.insert(FragmentFlags::IS_BODY_ELEMENT_OF_HTML_ELEMENT_ROOT);
             }
--- a/components/layout_2020/dom_traversal.rs
+++ b/components/layout_2020/dom_traversal.rs
@@ -89,14 +89,20 @@ where
         });
 
         let threadsafe_node = node.to_threadsafe();
-        let flags = match threadsafe_node.as_element() {
-            Some(element) if element.is_body_element_of_html_element_root() => {
-                FragmentFlags::IS_BODY_ELEMENT_OF_HTML_ELEMENT_ROOT
-            },
-            Some(element) if element.get_local_name() == &local_name!("br") => {
-                FragmentFlags::IS_BR_ELEMENT
-            },
-            _ => FragmentFlags::empty(),
+        let mut flags = FragmentFlags::empty();
+        if let Some(element) = threadsafe_node.as_element() {
+            if element.is_body_element_of_html_element_root() {
+                flags.insert(FragmentFlags::IS_BODY_ELEMENT_OF_HTML_ELEMENT_ROOT);
+            }
+            match element.get_local_name() {
+                &local_name!("br") => {
+                    flags.insert(FragmentFlags::IS_BR_ELEMENT);
+                },
+                &local_name!("table") | &local_name!("th") | &local_name!("td") => {
+                    flags.insert(FragmentFlags::IS_TABLE_TH_OR_TD_ELEMENT);
+                },
+                _ => {},
+            }
         };
 
         Self {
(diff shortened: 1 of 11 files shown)


components/layout_2020/dom_traversal.rs
```diff
@@ -89,14 +89,20 @@ where
...
+            if element.is_body_element_of_html_element_root() {
+                flags.insert(FragmentFlags::IS_BODY_ELEMENT_OF_HTML_ELEMENT_ROOT);
+            }
+            match element.get_local_name() {
+                &local_name!("br") => {
+                    flags.insert(FragmentFlags::IS_BR_ELEMENT);
+                },
+                &local_name!("table") | &local_name!("th") | &local_name!("td") => {
+                    flags.insert(FragmentFlags::IS_TABLE_TH_OR_TD_ELEMENT);
+                },
```
Loirooriol, line 103: This (and possibly the br?) should only happen in the HTML namespace.

```js
<!DOCTYPE html>
<style>
*|td { display: table-cell}
</style>
<body>
<script>
var td = document.createElementNS("not-html", "td");
var child = document.createElement("div");
td.appendChild(child);
document.body.appendChild(td);
document.body.append(child.offsetParent.tagName);
</script>
</body>
```

Should say `BODY`, not `td`.

On the other hand, `var td = document.createElementNS("http://www.w3.org/1999/xhtml", "td");` should say `TD`.

mrobinson:
I've tagged these elements now only when they are in an HTML namespace.

components/script/layout_dom/node.rs
```diff
@@ -313,6 +313,14 @@ impl<'dom> ThreadSafeLayoutNode<'dom> for ServoThreadSafeLayoutNode<'dom> {
             })
     }
 
+    fn as_html_element(&self) -> Option<ServoThreadSafeLayoutElement<'dom>> {
+        let element = self.as_element();
+        if !element.map_or(false, |element| element.element.is_html_element()) {
+            return None;
+        }
+        element
```
jdm, line 321:
```suggestion
        self.as_element()
            .filter(|element| element.element.is_html_element())
```

jdm (approved):

components/script/layout_dom/node.rs
mrobinson: Thanks. I've made this change.

components/layout_2020/query.rs
```diff
@@ -531,6 +514,29 @@ fn process_offset_parent_query_inner(
     })
 }
 
+/// Returns whether or not the element with the given style and body element determination
+/// is eligible to be a parent element for offset* queries.
```
Loirooriol, line 518: This makes it sound like it's also for `offsetLeft` and such, but it's only for `offsetParent`, isn't it?

components/layout_2020/query.rs
```diff
@@ -531,6 +514,29 @@ fn process_offset_parent_query_inner(
...
+///
+/// From <https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface>:
```
Loirooriol, line 520: ... and then https://www.w3.org/TR/cssom-view-1/#dom-htmlelement-offsetparent would be a more specific link.

Loirooriol (approved):

components/layout_2020/query.rs
mrobinson: I think this is used for `offsetLeft` and friends, because they depend on the definition of `offsetParent`. For instance, the definition of `OffsetLeft()` in `htmlelement.rs` is:

```rust
    // https://drafts.csswg.org/cssom-view/#dom-htmlelement-offsetleft
    fn OffsetLeft(&self) -> i32 {
        if self.is::<HTMLBodyElement>() {
            return 0;
        }

        let node = self.upcast::<Node>();
        let window = window_from_node(self);
        let (_, rect) = window.offset_parent_query(node);

        rect.origin.x.to_nearest_px()
    }
```

components/layout_2020/query.rs
mrobinson: Yeah, even if this applies to all `offset` queries, this link is better and I've updated it.

components/layout_2020/query.rs
Loirooriol: Sorry you are correct, I think I tried the other day and it didn't seem to work, but I probably made a mistake.

PR #32457, opened 2024-06-07
Split servoshell into Desktop and common part

jschwe:
Move all desktop / winit related files into a common module in servoshell.
This allows adding support for different non-winit platforms like OpenHarmony or android into the
servoshell in a follow-up commit.

---


jschwe:
I think this MR is ready for an initial review. I had to remove the native-bluetooth from the default features, since currently (as far as I know) there is no good way to have target specific default features - and the android and openharmony ports currently both don't support bluetooth yet. if someone has an alternative idea how to deal with the bluetooth, I'm open to suggestions. 

I have a follow-up branch, which basically adds ohos support to `mach build` and CI, but without adding the ohos port to servoshell yet. Still that would already be useful, since it would allow testing that libservo for OpenHarmony builds in CI.

ports/servoshell/Cargo.toml
```diff
@@ -52,34 +52,43 @@ xr-profile = ["libservo/xr-profile"]
...
+url = { workspace = true }
+lazy_static = { workspace = true }
+
 
 [target.'cfg(not(target_os = "android"))'.dependencies]
-arboard = { version = "3" }
 backtrace = { workspace = true }
-cfg-if = { workspace = true }
+# Todo: Could argument parsing be useful on non-desktop?
+getopts = { workspace = true }
```
jschwe, line 66: I'm not really sure if this makes sense, or if the whole argument parsing should just be desktop specific. 
Probably for android / ohos, we would just make a special `about:prefs` url, that allows editing an on-disk settings file?

jschwe:
https://github.com/jschwe/servo/tree/jschwender/servoshell_winit_ohos

Is a follow-up branch which adds `ohos` target support to `mach` and tests building servoshell in CI (without ohos-servoshell containing any relevant code, but it ensures that libservo builds)

mukilan (approved):
This PR looks good to me! Thanks for this contribution. 

@mrobinson, just tagging you in case you have more thoughts on this PR.

ports/servoshell/Cargo.toml
jschwe: I removed the todo. I agree that its probably useful to keep it for now. We can always make it an optional dependency later (for non-desktop), if e.g. binary size becomes an important metric.

mukilan (approved):

PR #32458, opened 2024-06-07
Don't shrink table columns when handling cells with greater spanning

Loirooriol:
For example:
```html
<table border="1">
  <tr>  <td></td>  <td></td>  </tr>
  <tr>  <td colspan="2"></td>  </tr>
</table>
```
We should initially size the columns according to the cells in the first row since they have a span of 1. Then we handle the cell in the second row with a span of 2, this should be able to increase the size of the columns, but never decrease them.

---


mrobinson (approved):
Nice!

PR #32463, opened 2024-06-07
fonts: Merge Noto fallback lists

mrobinson:
On platforms that ship Noto, the list of fallback fonts can be shared.
This reduces code duplcation and makes it easier to keep up to date with
changes in Noto.

Note that more work is required for these fonts to be active on macOS, even
though they are in the fallback list. That work will be done in a separate PR.

---

### DIFF
--- a/components/gfx/platform/freetype/font_list.rs
+++ b/components/gfx/platform/freetype/font_list.rs
@@ -34,6 +34,7 @@ use unicode_script::Script;
 use super::c_str_to_string;
 use crate::font::map_platform_values_to_style_values;
 use crate::font_template::{FontTemplate, FontTemplateDescriptor};
+use crate::platform::add_noto_fallback_families;
 use crate::text::FallbackFontSelectionOptions;
 
 /// An identifier for a local font on systems using Freetype.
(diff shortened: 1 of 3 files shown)


components/gfx/platform/freetype/font_list.rs
```diff
@@ -207,226 +208,23 @@ pub fn fallback_font_families(options: FallbackFontSelectionOptions) -> Vec<&'st
...
-                families.push("Noto Sans CJK KR");
-            },
-            UnicodeBlock::Hanunoo => families.push("Noto Sans Hanunoo"),
-            UnicodeBlock::Hatran => families.push("Noto Sans Hatran"),
-            UnicodeBlock::Hebrew => families.push("Noto Sans Hebrew"),
+            UnicodeBlock::EnclosedIdeographicSupplement => families.push("WenQuanYi Micro Hei"),
             UnicodeBlock::Hiragana |
             UnicodeBlock::Katakana |
             UnicodeBlock::KatakanaPhoneticExtensions => {
                 families.push("TakaoPGothic");
```
mukilan, line 227: `add_noto_fallback_families` [already](https://github.com/servo/servo/pull/32463/files#diff-5ea622dafd8b31eadec41754d816ec902261b3e35293a5d8e62426eb1cbad6b0R174) adds `TakaoPGothic` to the list. Was the intention to remove that line from `add_noto_fallback_families` and add the non-Noto families here? If so, wouldn't this change the order of fallback i.e  `Noto Sans JP` and `Noto Sans CJK JP` would appear before `TakaoPGothic` in the fallback list. Is that fine?

mukilan (approved):

components/gfx/platform/freetype/font_list.rs
mrobinson: Whoops! Nice catch. These lists are so long it is easy to make a mistake. I've removed the duplicate addition. It's true that this will now add `TakaoPGothic` after the Noto fonts, but I think that's okay. The Noto fonts are high-quality and it's likely just fine if those are used as a fallback before other ones. This isn't the case for Mac, which ships with high-quality non-Noto fonts that should likely be selected first.

PR #32465, opened 2024-06-09
Add cargo-deny to mach-tidy to check license compliance.

jschwe:
Use cargo-deny to add an explicit allow-list for licenses.
The exact licenses allowed should probably be discussed, but it seems we currently don't include any dependencies with problematic licenses.

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -16,6 +16,7 @@ import os.path as path
 import shutil
 import subprocess
 import textwrap
+import json
 
 import wpt
 import wpt.manifestupdate
@@ -226,6 +227,32 @@ class MachCommands(CommandBase):
     @CommandArgument('--no-progress', default=False, action="store_true",
                      help="Don't show progress for tidy")
     def test_tidy(self, all_files, no_progress):
+        def check_dep_license_errors() -> bool:
+            print("\r ➤  Checking Licenses of Rust dependencies...")
+            result = subprocess.run(["cargo-deny", "--format=json", "check", "licenses"], encoding='utf-8',
+                                    capture_output=True)
+            if result.returncode != 0:
+                assert result.stderr is not None, "cargo deny should return error information via stderr when failing"
+                json_lines = result.stderr.splitlines()
+                error_info = [json.loads(json_struct) for json_struct in json_lines]
+                error_messages = []
+                num_license_errors = 'unknown'
+                for error in error_info:
+                    if error['type'] == 'summary':
+                        num_license_errors = error['fields']['licenses']['errors']
+                    else:
+                        crate = error['fields']['graphs'][0]['Krate']
+                        lic_name = error['fields']['notes'][0]
+                        error_msg = f'Rejected license {lic_name} . Run `cargo deny` for more details'
+                        error_messages.append(
+                            f'   | Rust dependency {crate["name"]} (version {crate["version"]}): {error_msg}')
+                print(f'    `cargo deny` reported {num_license_errors} licenses errors')
+                for msg in error_messages:
+                    print(msg)
+                return False
+            else:
+                return True
+
         tidy_failed = tidy.scan(not all_files, not no_progress)
 
         print("\r ➤  Checking formatting of rust files...")
@@ -236,7 +263,9 @@ class MachCommands(CommandBase):
         print("\r ➤  Checking formatting of toml files...")
         taplo_failed = format_toml_files_with_taplo()
 
-        tidy_failed = tidy_failed or rustfmt_failed or taplo_failed
+        cargo_deny_licenses_failed = check_dep_license_errors()
+
+        tidy_failed = tidy_failed or rustfmt_failed or taplo_failed or cargo_deny_licenses_failed
         print()
         if tidy_failed:
             print("\r ❌ test-tidy reported errors.")
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -227,9 +227,8 @@ class MachCommands(CommandBase):
                      help="Don't show progress for tidy")
     def test_tidy(self, all_files, no_progress):
         tidy_failed = tidy.scan(not all_files, not no_progress)
-        cargo_deny_licenses_failed = tidy.check_license_errors()
 
-        print("\r ➤  Checking formatting of rust files...")
+        print("\r ➤  Checking formatting of Rust files...")
         rustfmt_failed = call(["cargo", "fmt", "--", *UNSTABLE_RUSTFMT_ARGUMENTS, "--check"])
         if rustfmt_failed:
             print("Run `./mach fmt` to fix the formatting")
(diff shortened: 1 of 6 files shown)


mrobinson (changes requested):
This is a great change and something I had on my radar for a while! I have just a small suggestion here:

- Install `cargo-deny` during the `./mach bootstrap` like we do with taplo.
- Replace the existing `check_license` check in `python/tidy/tidy.py` with `cargo-deny`.

jschwe:
> Replace the existing check_license check in python/tidy/tidy.py with cargo-deny.

Just to clarify: `cargo deny` will only check the license in the Cargo.toml manifest, while the original tidy.py code ensures that each file has a license header. Would you also drop the requirement of having the license header in each file, since it can't be tested for anymore?

mrobinson:
> Just to clarify: `cargo deny` will only check the license in the Cargo.toml manifest, while the original tidy.py code ensures that each file has a license header. Would you also drop the requirement of having the license header in each file, since it can't be tested for anymore?

My apologies, I identified the wrong functionin `tidy.py` instead of `check_license`, I meant to indicate `check_dep_license_errors`.

components/servo/deny.toml
```diff
@@ -0,0 +1,89 @@
...
+    # Each entry is a crate relative path, and the (opaque) hash of its contents
+    { path = "LICENSE", hash = 0xbd0eed23 },
+]
+
+
+# This section is considered when running `cargo deny check bans`.
+# More documentation about the 'bans' section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
+[bans]
+multiple-versions = "warn"
```
sagudev, line 72: We could also use this instead of current servo-tidy multiple versions checks: https://github.com/servo/servo/blob/e6ea4a9c298c02138938c27851ccf9cca0523f67/servo-tidy.toml#L21, although this does not need to be done in this PR.

components/servo/deny.toml
sagudev, line 1: Any reason why this isn't used for whole workspace?

jschwe: I was thinking about having different rules for `libservo` (strict) vs the whole workspace, so servoshell could potentially have more relaxed rules. It seems that at least currently the difference in licenses required by servoshell is minor, so probably there is no point in having two deny.toml files.

sagudev:
> Would you want `./mach tidy` to still run the `cargo deny` checks, or would the intention be to keep the cargo deny check separate (e.g. by adding it to a pre-commit hook), as a step towards slowly moving away from mach?

Even when there is pre-commit hook, you still want to be able to run tidy manually.

components/servo/deny.toml
jschwe: I think both the multiple versions and the wildcard check could be added, but that should be done in a future PR.

jschwe:
I integrated the `cargo deny` check into `mach tidy`. since the normal `cargo deny` error output is quite large (it basically shows `cargo tree -i dependency`) due to the deep dependency trees, I just give basic error information and tell the user to run `cargo deny` themselves to get more detailed information.

Example `mach tidy output`:

```
./mach test-tidy
 ➤  Checking config file (./servo-tidy.toml)...
 ➤  Checking directories for correct file extensions...
 ➤  Checking cargo lock (/data/servo/Cargo.lock)...
 ➤  Checking files for tidiness
 ➤  Skipping WPT lint checks, because no relevant files changed.
 ➤  Checking formatting of rust files...
 ➤  Checking formatting of toml files...
 ➤  Checking Licenses of Rust dependencies...
    `cargo deny` reported 1 licenses errors
   | Rust dependency tinyfiledialogs (version 3.3.9): Rejected license zlib-acknowledgement - zlib/libpng License with Acknowledgement: . Run `cargo deny` for more details
```

python/servo/testing_commands.py
```diff
@@ -226,6 +227,32 @@ def test_content(self):
     @CommandArgument('--no-progress', default=False, action="store_true",
                      help="Don't show progress for tidy")
     def test_tidy(self, all_files, no_progress):
+        def check_dep_license_errors() -> bool:
```
mrobinson, line 230: Could this actually move to `tidy.py`? You're right that we have a long term goal of not longer relying on `mach` an the idea is that `tidy.py` can be used standalone.

python/servo/testing_commands.py
```diff
@@ -226,6 +227,32 @@ def test_content(self):
...
+            print("\r ➤  Checking Licenses of Rust dependencies...")
+            result = subprocess.run(["cargo-deny", "--format=json", "check", "licenses"], encoding='utf-8',
+                                    capture_output=True)
+            if result.returncode != 0:
```
mrobinson, line 234: I think could do an early return like this here instead

```python
    if result.returncode == 0:
        return False
```

python/servo/testing_commands.py
```diff
@@ -226,6 +227,32 @@ def test_content(self):
...
+                error_info = [json.loads(json_struct) for json_struct in json_lines]
+                error_messages = []
+                num_license_errors = 'unknown'
+                for error in error_info:
+                    if error['type'] == 'summary':
+                        num_license_errors = error['fields']['licenses']['errors']
+                    else:
+                        crate = error['fields']['graphs'][0]['Krate']
+                        lic_name = error['fields']['notes'][0]
+                        error_msg = f'Rejected license {lic_name} . Run `cargo deny` for more details'
```
mrobinson, line 246:
```suggestion
                        error_msg = f'Rejected license "{lic_name}". Run `cargo deny` for more details'
```

mrobinson (changes requested):
Looking good. Thanks for these changes. Just a few more suggestions.

python/servo/testing_commands.py
jschwe: Moved the function to tidy.py. I still kept it separate from the scan() function, since cargo deny operates on crates, not files.

mrobinson:
I've pushed a followup change which:

- Fixes a couple more nits
- Fully integrates the `cargo-deny` check into tidy instead of exposing it as a separate function. This will be important once tidy is able to emit GitHub Actions annotations (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions).
- Skip both `Cargo.lock` checks when `Cargo.lock` hasn't changed. `cargo-deny` is a bit slow locally so this should improve speed when developing. The bots always run all checks.

mrobinson (approved):

mrobinson:
Oh, in addition I've fixed an issue where an error that isn't a license error (for instance one about `Cargo.lock` being corrupted) would cause a Python exception.

jschwe:
Cool - thanks! The only thing I'm not quite convinced of is integrating the cargo deny check into the `scan` function. I don't think pointing the user to the `Cargo.lock` file is valuable when investigating a crate `cargo-deny` is complaining about (Opposed to source files, users will never be editing lock files directly). 
But this is something we could reevaluate at a later point.

mrobinson:
> You're right that adding a message to `Cargo.lock` isn't very useful, but perhaps at some point we can indicate the line in `Cargo.toml`. In addition, I think it would be a nice followup to use `cargo-deny` to implement the duplicate crate check that tidy is doing.

I should also mention that having every run from `tidy.py` has other benefits. For instance, eventually these checks can be run in parallel.

Issue #32404, opened 2024-05-29
Fix support for the DevTools protocol


eerii:
Hi! As part of the Outreachy project this year I will be working on supporting DevTools again, making them work with the current versions of Servo and Firefox. This is a meta issue that gathers all of the information in a place that can be easily referenced, and in which updates on the state of the project can be provided. Please do add anything you might consider useful or necessary to the discussion!

#### Recent (~2023-24)

These are the more relevant to restoring support for DevTools. It was broken around June 2023 and it is confirmed not to work properly from Firefox 110 onward.

#29831 : Main issue that details the broken support and previous attempts to fix it.
[Bugzilla's side](https://bugzilla.mozilla.org/show_bug.cgi?id=1836856) : Many useful links to documentation on the new interface.

#### Mid (~2020-2022)

These are from a time where DevTools worked, but these issues were never presumably solved, so if we get it up and running again it may be good to look if they are still relevant.

#26667 : Reporting unhandled promise rejections didn't work
#26368 : Errors navigating DevTools
#26323 : Unreported network events
#25119 : Make it so DevTools can be started later

#### Older (~2013-2019)

Outdated but interesting issues.

#8973 : Old experiment to create a standalone HTML UI of the Firefox DevTools
#4957 : Add support for the debug timeline

## Previous work

This section only includes relevant prs since DevTools were broken last year.

#32227 : Great pr from @oSoMoN that fixes how Servo is displayed in the `about:debugging` Firefox page and removes some warnings.
https://github.com/fabricedesre/servo/commit/0c41bf409d2489a489e136b6703075aa746bb3d1 : Work done by @fabricedesre to update the DevTools implementation to conform to the new Firefox changes. It may be a good place to start.

## Roadmap

TBD

- [ ] Determine the current state of DevTools support and update the plan of action detailing what to work on.
- [ ] ...

## Links to useful documentation

- ...

eerii:
## How to debug RDP messages

One key part of restoring the DevTools support is correctly sending the proper messages through the RDP ([remote debugging protocol](https://firefox-source-docs.mozilla.org/devtools/backend/protocol.html)). However, the message fields for each action between actors are not well documented, making it tricky to know exactly what to send.

Servo can easily show the messages it sent and the ones it received from the DevTools server. It's as easy as enabling the correct logging level when running the program.

```sh
RUST_LOG="error,devtools=debug" ./mach run --devtools=1234
```

[...]

PR #32475, opened 2024-06-11
DevTools: Display tabs and processes

eerii:
Following the work of @fabricedesre and cleaning some things a bit, this PR updates the messages sent from Servo to the DevTools server so that Tabs and Processes can be shown again in the main view.

Note that tab inspection doesn't work yet (that's the next step!), as a Watcher Actor needs to be implemented.

---
- [x] These changes are a part of #32404
- [x] These changes do not require tests because they relate to an external tool*

* I am unsure on how to write automated tests for DevTools. They can be tested manually running Servo with the propper argument: `./mach run --devtools=6080` and connecting via `about:debugging` in Firefox. If there is a way to write tests please let me know.

### DIFF
--- a/components/devtools/actors/root.rs
+++ b/components/devtools/actors/root.rs
@@ -14,7 +14,7 @@ use serde_json::{Map, Value};
 use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
 use crate::actors::device::DeviceActor;
 use crate::actors::performance::PerformanceActor;
-use crate::actors::tab::{DescriptorTraits, TabDescriptorActor, TabDescriptorActorMsg};
+use crate::actors::tab::{TabDescriptorActor, TabDescriptorActorMsg};
 use crate::actors::worker::{WorkerActor, WorkerMsg};
 use crate::protocol::{ActorDescription, JsonPacketStream};
 use crate::StreamId;
@@ -95,7 +95,12 @@ struct ListProcessesResponse {
     processes: Vec<ProcessForm>,
 }
 
-// https://searchfox.org/mozilla-central/source/devtools/server/actors/descriptors/process.js#196
+#[derive(Default, Serialize)]
+pub struct DescriptorTraits {
+    pub(crate) watcher: bool,
+    pub(crate) supportsReloadDescriptor: bool,
+}
+
 #[derive(Serialize)]
 struct ProcessForm {
     actor: String,
@@ -158,7 +163,6 @@ impl Actor for RootActor {
                 ActorMessageStatus::Processed
             },
 
-            // TODO: Unexpected message getTarget for process (when inspecting)
             "getProcess" => {
                 let reply = GetProcessResponse {
                     from: self.name(),
@@ -201,7 +205,6 @@ impl Actor for RootActor {
                         })
                         .collect(),
                 };
-                println!("LIST TABS: {:#?}", actor.tabs);
                 let _ = stream.write_json_packet(&actor);
                 ActorMessageStatus::Processed
             },
@@ -228,7 +231,6 @@ impl Actor for RootActor {
                 ActorMessageStatus::Processed
             },
 
-            // TODO: Unexpected message getWatcher for tab (when inspecting)
             "getTab" => {
                 if let Some(serde_json::Value::Number(browserId)) = msg.get("browserId") {
                     let targetTab = self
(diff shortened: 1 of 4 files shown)


components/devtools/actors/root.rs
```diff
@@ -220,13 +232,30 @@ impl Actor for RootActor {
...
-                let _ = stream.write_json_packet(&reply);
-                ActorMessageStatus::Processed
+                if let Some(serde_json::Value::Number(browserId)) = msg.get("browserId") {
+                    let targetTab = self
+                        .tabs
+                        .iter()
+                        .map(|target| {
+                            registry
+                                .find::<TabDescriptorActor>(target)
+                                .encodable(registry, true)
```
jdm, line 242: Why encode here, rather than when we find the actual tab we want to send back?

jdm (approved):
Looks like a good start!

jdm:
And you are correct—we have not prioritized any kind if automated testing for the devtools server implementation.

components/devtools/actors/root.rs
eerii: I did it this way since `TabDescriptorActorMsg` contains the `broswerId`, but `TabDescriptorActor` doesn't, and we need to filter them by that, I thought of doing the conversion beforehand, but I can absolutely try to do it other way.

jdm: Ah, I see. That's reasonable!

components/devtools/actors/root.rs
```diff
@@ -220,13 +232,30 @@ impl Actor for RootActor {
...
             "getTab" => {
-                let tab = registry.find::<TabDescriptorActor>(&self.tabs[0]);
-                let reply = GetTabReply {
-                    from: self.name(),
-                    tab: tab.encodable(registry),
-                };
-                let _ = stream.write_json_packet(&reply);
-                ActorMessageStatus::Processed
+                if let Some(serde_json::Value::Number(browserId)) = msg.get("browserId") {
+                    let targetTab = self
```
mrobinson, line 236:
```suggestion
                    let target_tab = self
```

components/devtools/actors/root.rs
```diff
@@ -220,13 +232,30 @@ impl Actor for RootActor {
...
 
             "getTab" => {
-                let tab = registry.find::<TabDescriptorActor>(&self.tabs[0]);
-                let reply = GetTabReply {
-                    from: self.name(),
-                    tab: tab.encodable(registry),
-                };
-                let _ = stream.write_json_packet(&reply);
-                ActorMessageStatus::Processed
+                if let Some(serde_json::Value::Number(browserId)) = msg.get("browserId") {
```
mrobinson, line 235:
```suggestion
                if let Some(serde_json::Value::Number(browser_id)) = msg.get("browserId") {
```

components/devtools/actors/tab.rs
```diff
@@ -9,25 +9,28 @@ use serde_json::{Map, Value};
...
-}
-
+// https://searchfox.org/mozilla-central/source/devtools/server/actors/descriptors/tab.js
 #[derive(Serialize)]
 pub struct TabDescriptorActorMsg {
     actor: String,
+    browserId: u32,
+    browsingContextId: u32,
+    isZombieTab: bool,
+    outerWindowID: u32,
```
mrobinson, line 23:
```suggestion
   #[serde(rename(serialize = "browserId"))]
    browser_id: u32,
    #[serde(rename(serialize = "browsingContextId"))]
    browsing_context_id: u32,
    #[serde(rename(serialize = "isZombieTab"))]
    is_zombie_tab: bool,
    #[serde(rename(serialize = "outerWindowId"))]
    outer_window_id: u32,
```

components/devtools/actors/root.rs
```diff
@@ -95,17 +95,25 @@ struct ListProcessesResponse {
     processes: Vec<ProcessForm>,
 }
 
+#[derive(Default, Serialize)]
+pub struct DescriptorTraits {
+    pub(crate) watcher: bool,
+    pub(crate) supportsReloadDescriptor: bool,
```
mrobinson, line 101: I think it would be good to start using serde renaming so that this file can remove `allow(non_snake_case)` one day.

```suggestion
    #[serde(rename(serialize = "supportsReloadDescriptor"))]
    pub(crate) supports_reload_descriptor: bool,
```

components/devtools/actors/root.rs
mrobinson, line 235: Would it be possible to move this code into a function in order to use an early return pattern and avoid some indentation?

mrobinson (changes requested):
Great stuff! Just a couple small nits to help gradually improve the devtools code:

components/devtools/actors/root.rs
eerii: That's awesome, I didn't know serde could do that! I found out you can even rename all the fields of a struct using camelCase [automatically](https://serde.rs/attr-rename.html), which seems more robust to avoid spelling mistakes in the future.

components/devtools/actors/root.rs
eerii: Absolutely, how about now? I used let else to further reduce indentation but I can change it if another pattern is preferred. I am also not sure about the name `get_tab_msg_by_browser_id` but I couldn't think of one that was shorter but still descriptive, any suggestion is welcome.

mrobinson (approved):

components/devtools/actors/root.rs
mrobinson: This looks great now. Thank!

components/devtools/actors/root.rs
mrobinson: Oh, great find!

PR #32476, opened 2024-06-11
clippy: fix some leftover warnings in components/net

Mikopet:
I was lurking around when I found some seems-easy-to-fix clippy warnings in `net` component.

---


mrobinson (approved):
Interesting. TIL that `clone_from` is a thing.

PR #32483, opened 2024-06-12
layout: Linebreak the entire InlineFormattingContext at once

mrobinson:
Instead of linebreaking inside each single-font text segment, linebreak
the entire inline formatting context at once. This has several benefits:

1. It allows us to use `icu_segmenter` (already in use from style),
   which is written against a newer version of the Unicode spec --
   preventing breaking emoji clusters.
2. Opens up the possibility of changing the way that linebreaking and
   shaping work -- eventually allowing shaping across inline box
   boundaries and line breaking *after* shaping.

Some tests start to fail:
 - A large portion of the unexpected results is that the linebreaker no longer seems to be putting a soft line break opportunity after the zero-width joiner. This causes the Arabic ('ع') character to render differently when it's shaped together with a zero-width joiner and no followup character. This happens in tests because the character that it is supposed to join with is in a different inline box and we do not shape across inline box boundaries. So, in summary, changes in the line breaker are causing tests that were passing by accident to start failing, when they should have been failing all along.
 - Some tests fail because we do not support `line-break: anywhere` and `word-break: manual`.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3270,6 +3270,7 @@ dependencies = [
  "gfx",
  "gfx_traits",
  "html5ever",
+ "icu_segmenter",
  "ipc-channel",
  "lazy_static",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ http = "0.2"
 hyper = "0.14"
 hyper-rustls = { version = "0.24", default-features = false, features = ["acceptor", "http1", "http2", "logging", "tls12", "webpki-tokio"] }
 hyper_serde = { path = "components/hyper_serde" }
+icu_segmenter = "1.5.0"
 image = "0.24"
 imsz = "0.2"
 indexmap = { version = "2.2.6", features = ["std"] }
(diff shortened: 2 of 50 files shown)


mrobinson:
A large portion of the unexpected results is that the linebreaker no longer seems to be putting a soft line break opportunity after the zero-width joiner. This causes the Arabic ('ع') character to render differently when it's shaped together with a zero-width joiner and no followup character. This happens in tests because the character that it is supposed to join with is in a different inline box and we do not shape across inline box boundaries. So, in summary, changes in the line breaker are causing tests that were passing by accident to start failing, when they should have been failing all along.

atbrakhi (approved):

nicoburns:
This collected inline layout would be the perfect boundary at which to break out an external library. Or indeed build upon an existing library: [cosmic-text](https://github.com/pop-os/cosmic-text) and [parley](https://github.com/linebender/parley) both already exist at this level of abstraction and already implement things like shaping before line breaking. It seems such a shame to have duplicates of all this low-level text work when everyone needs the same thing.

PR #32486, opened 2024-06-13
layout: Allocate inline box start space on segment

mrobinson:
Instead of allocating the inline padding and border space on the line,
allocate it on the segment -- which where the inline box start goes.

---


atbrakhi (approved):
interesting that nothing failed or passed. Maybe that's why we never noticed that we were allocating space on the line instead of segment.

Not sure but does it worth adding a test for this?

PR #32487, opened 2024-06-13
layout: Simplify `Contents` a little

mrobinson:
Instead of duplicating some of `NonReplacedContents` in `Contents`,
divide it into either replaced and non-replaced content, since this is
how the layout system processes `Contents` always. In addition, stop
using `TryInto` to match replaced or non-replaced contents, as it is
quite confusing to handle an `Err` as a success case.

---


atbrakhi (approved):
I think this looks good.

mrobinson:
The 2013 failures are likely due to 2013 results being out of date as these changes only affect 2020

PR #32492, opened 2024-06-14
Extend `MoveResizeWebView` to update border radius

### LINKED ISSUE #32485: Decorated Web View
**Describe the new feature:**
I tried to add decoration to webview and it seems to be [possible](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Can.20we.20add.20custom.20shaders.20to.20webrender.3F/near/443217355) by extending compositor a bit.
So now I would like to add such feature for it. At first, I thought adding config in `pref` should be enough.
However, different webviews might want different radius values depending on their sizes.
Perhaps we can extend [NewWebView](https://github.com/servo/servo/blob/0a641816bf8f402800d7ecec12d2d64505975c16/components/compositing/windowing.rs#L88) to add another field of radius value?

wusyong:
This variant seems like the right one for updating the web view's decoration.
If not set, the radius will default to zero and clip with a regular rectangle.

---

- [x] These changes fix #32485

- [x] There are tests for these changes: compositing's webview unit tests should pass.


components/compositing/compositor.rs
```diff
@@ -1222,6 +1227,10 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
                 rect_changed = rect != webview.rect;
                 size_changed = rect.size() != webview.rect.size();
                 webview.rect = rect;
+
+                if let Some(r) = radius {
+                    webview.radius = r;
+                }
```
Taym95, line 1233:
```suggestion
radius.map(|r| webview.radius = r);
```

jdm: I actually prefer `if let` here because we don't use the return value of the `map`.

mrobinson:
@wusyong I like the idea of allowing rounded display of `WebView`, but as we move toward a model where each `WebView` draws itself, maybe this could be handled by the embedding application rendering to a texture? What do you think?

wusyong:
> @wusyong I like the idea of allowing rounded display of `WebView`, but as we move toward a model where each `WebView` draws itself, maybe this could be handled by the embedding application rendering to a texture? What do you think?

I agree with this as I also did another way without this PR already. We should tell ebedders how they could do it though. IIRC we suggested blitting before.

mrobinson:
Okay. I'll close this then and then for the time being. I imagine we could add documentation for how to blit `WebView`s into the rustdoc. It would be really nice to finish off support for per-`WebView` painting.

PR #32493, opened 2024-06-14
fonts: Respect emoji variation selector when selecting fonts

mrobinson:
This uses a pretty simple heuristic to select a font likely to contain
color emoji. In the future Servo should actually check if the font also
contains a color representation of the character in question. For now
the code assumes that when a font supports color glyphs of some kind and
supports the character in question at all, it supports the color
version.

This fixes support for rendering keycap emoji clusters such as 1️⃣ .

---

- [x] These changes do not require tests because emoji rendering is quite hard to test, but we verified that this improves the "emoji-list.html" manual test.

### DIFF
--- a/components/gfx/font.rs
+++ b/components/gfx/font.rs
@@ -32,7 +32,7 @@ use crate::platform::font::{FontTable, PlatformFont};
 pub use crate::platform::font_list::fallback_font_families;
 use crate::text::glyph::{ByteIndex, GlyphData, GlyphId, GlyphStore};
 use crate::text::shaping::ShaperMethods;
-use crate::text::{FallbackFontSelectionOptions, Shaper};
+use crate::text::{EmojiPresentationPreference, FallbackFontSelectionOptions, Shaper};
 
 #[macro_export]
 macro_rules! ot_tag {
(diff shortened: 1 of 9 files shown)


mukilan (approved):

mrobinson:
Hrm...there seems to be some kind of problem with this on macOS. I'm investigating now...

mrobinson:
Now that we are selecting fonts based on the emoji presentation needs, this change revealed an issue where we were too readily choosing emoji presentation. Instead of requesting it for every possible emoji character in the absence of an explicit emoji variation selection, we should only do it for characters that have an emoji presentation themselves.

For example, "0" is classified as an "emoji character" by `unicode_properties` because it *might* be used in an emoji sequence, but it does not have an emoji presentation unless followed by the emoji variation selector. '😅' is both an emoji character and has an emoji presentation, so it doesn't need to be followed by a variation selector to require the emoji font.

mrobinson:
I've had to modify this a bit. Now this explicitly handles characters that are emoji, but without emoji presentation, such as '&#x263A;'. These kind of characters should explicitly avoid emoji-like fonts unless they are paired with an emoji variation selector.

mrobinson:
I have chosen to add a new mode `EmojiPresentationPreference::None` as well to avoid expensive lookups of tables when it shouldn't matter what the presentation is for the character.

mrobinson:
I'll go ahead and land this one. There are some changes since the last review, these are mostly minor.

mukilan (approved):

atbrakhi (approved):

Issue #32302, opened 2024-05-16
Support crown on windows without hacks


sagudev:
We need proper solution for https://github.com/servo/servo/issues/32300. Dylint has some interesting solutions: https://github.com/trailofbits/dylint/commit/c22218107853018a04e01a65a16406612e36186a

jschwe:
Ah, this explains the recent issue I had on windows.
Do we need to use `crown` on all platforms? I've been wondering if we could turn crown off by default, and only enable it in CI / ondemand.

sagudev:
> Ah, this explains the recent issue I had on windows.

I should probably add workaround in mach instead of CI file.

> Do we need to use crown on all platforms?

I think not as it only does additional pass of lints on script crate, which do not have (I think) any platform specific cfgs where lints are actually linting (dom structs).

> I've been wondering if we could turn crown off by default, and only enable it in CI / ondemand.

That's the ultimate plan, as only those who do changes in script crate needs to run it.

jschwe:
> That's the ultimate plan, as only those who do changes in script crate needs to run it.

Is there anything blocking this? Couldn't we just delete `rustc = "crown"` from the config.toml, and in CI set `CARGO_BUILD_RUSTC=crown` and add an optional parameter to mach which also sets this environment variable?

sagudev:
> Is there anything blocking this?

Not anything that I know of.

CC @jdm @mrobinson

jdm:
I'm ok with that. We may want to start by updating thevPR template to have a checklist item about running crown if any script changes are included.

sagudev:
> I'm ok with that. We may want to start by updating thevPR template to have a checklist item about running crown if any script changes are included.

CI would still run crown, so I do not think we would need this (I never really understand tidy and build on PR checklist as this is checked by PRs CI check anyway).

mrobinson:
> CI would still run crown, so I do not think we would need this (I never really understand tidy and build on PR checklist as this is checked by PRs CI check anyway).

I think this is still pretty useful as it can help reduce some back-and-forth between new contributors and reviews. In addition, sometimes I only remember to run tidy when I am filling out the PR template. :sweat_smile:

PR #32494, opened 2024-06-14
Make `crown` optional

jschwe:
This change makes `crown` optional (as previously discussed in #32302). `crown` is now only used if `--use-crown` is passed to mach or `build.use-crown` is set to `true` in `.servobuild`.

`RUSTC_BOOTSTRAP=crown,script` is now enabled conditionally on `crown` being used. During bootstrap `crown` is still build unconditionally.

To make reviewing easier, this PR is split into multiple commits, where each `crown` related lint kind was replaced in an individual commit.

In many common scenarios `cargo build` should now just work. 

---

### DIFF
--- a/docs/HACKING_QUICKSTART.md
+++ b/docs/HACKING_QUICKSTART.md
@@ -78,8 +78,10 @@ The `--` separates `mach` options from `servo` options. This is not required, bu
 This guide only covers the most important options. Be sure to look at all the available mach commands and the servo options:
 
 ```shell
-./mach --help         # mach options
-./mach run -- --help  # servo options
+./mach --help             # mach options
+./mach run -- --help      # servo options
+./mach build              # Build servo with the default configuration
+./mach build --use-crown  # Enable the crown linting tool - recommended when hacking on DOM code.
 ```
 
 ## Some basic Rust
@@ -183,26 +185,39 @@ work done by `./mach` (and vice versa).
 
 You can override this in a `.vscode/settings.json` file:
 
-```
+```json
 {
     "rust-analyzer.check.overrideCommand": [
         "./mach", "check", "--message-format=json" ],
     "rust-analyzer.cargo.buildScripts.overrideCommand": [
         "./mach", "check", "--message-format=json" ],
-    "rust-analyzer.rustfmt.overrideCommand": [ "./mach", "fmt" ],
+    "rust-analyzer.rustfmt.overrideCommand": [ "./mach", "fmt" ]
+}
+```
+
+If you are hacking on DOM code it is recommended to build servo with `./mach build --use-crown` and configure
+rust-analyzer to do the same by adding `--use-crown` to the `rust-analyzer.check.overrideCommand`:
+
+```json
+{
+    "rust-analyzer.check.overrideCommand": [
+        "./mach", "check", "--message-format=json", "--use-crown" ],
+    "rust-analyzer.cargo.buildScripts.overrideCommand": [
+        "./mach", "check", "--message-format=json", "--use-crown" ],
+    "rust-analyzer.rustfmt.overrideCommand": [ "./mach", "fmt" ]
 }
 ```
 
 If that still causes problems, then supplying a different target directory should fix this (although it will increase
 the amount of disc space used).
 
-```
+```json
 {
     "rust-analyzer.checkOnSave.overrideCommand": [
         "./mach", "check", "--message-format=json", "--target-dir", "target/lsp" ],
     "rust-analyzer.cargo.buildScripts.overrideCommand": [
         "./mach", "check", "--message-format=json", "--target-dir", "target/lsp" ],
-    "rust-analyzer.rustfmt.overrideCommand": [ "./mach", "fmt" ],
+    "rust-analyzer.rustfmt.overrideCommand": [ "./mach", "fmt" ]
 }
 ```
 
@@ -211,11 +226,11 @@ where `/nix/store/.../crown` is the output of `nix-shell etc/shell.nix --run 'co
 These settings should be enough to not need to run `code .` from within a `nix-shell etc/shell.nix`,
 but it wouldn’t hurt to try that if you still have problems.
 
-```
+```json
 {
     "rust-analyzer.server.extraEnv": {
-        "CARGO_BUILD_RUSTC": "/nix/store/.../crown",
-    },
+        "CARGO_BUILD_RUSTC": "/nix/store/.../crown"
+    }
 }
 ```
 
@@ -234,11 +249,11 @@ info: component 'llvm-tools' for target 'x86_64-unknown-linux-gnu' is up to date
 
 Then configure either your sysroot path or proc macro server path in `.vscode/settings.json`:
 
-```
+```json
 {
     "rust-analyzer.procMacro.enable": true,
     "rust-analyzer.cargo.sysroot": "[paste what you copied]",
-    "rust-analyzer.procMacro.server": "[paste what you copied]/libexec/rust-analyzer-proc-macro-srv",
+    "rust-analyzer.procMacro.server": "[paste what you copied]/libexec/rust-analyzer-proc-macro-srv"
 }
 ```
 
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -863,6 +863,12 @@ class CommandBase(object):
                     help='Build with frame pointer enabled, used by the background hang monitor.',
                 ),
                 CommandArgument('--without-wgl', group="Feature Selection", default=None, action='store_true'),
+                CommandArgument(
+                    '--use-crown',
+                    default=False,
+                    action='store_true',
+                    help='Enable servos `crown` linter tool'
+                )
             ]
 
         def decorator_function(original_function):
@@ -977,6 +983,7 @@ class CommandBase(object):
         env=None, verbose=False,
         debug_mozjs=False, with_debug_assertions=False,
         with_frame_pointer=False, without_wgl=False,
+        use_crown=False,
         target_override: Optional[str] = None,
         **_kwargs
     ):
@@ -1008,6 +1015,9 @@ class CommandBase(object):
                 assert command != 'build', "For Android / OpenHarmony `cargo rustc` must be used instead of cargo build"
                 if command == 'rustc':
                     args += ["--lib", "--crate-type=cdylib"]
+
+        if use_crown:
+            env['CARGO_BUILD_RUSTC'] = 'crown'
 
         if "-p" not in cargo_args:  # We're building specific package, that may not have features
             features = list(self.features)
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -1025,6 +1025,12 @@ class CommandBase(object):
                           'These options conflict, please specify only one of them.')
                     sys.exit(1)
             env['CARGO_BUILD_RUSTC'] = 'crown'
+            # Changing `RUSTC` or `CARGO_BUILD_RUSTC` does not cause `cargo check` to
+            # recheck files with the new compiler. `cargo build` is not affected and
+            # triggers a rebuild as expected. To also make `check` work as expected,
+            # we add a dummy `cfg` to RUSTFLAGS when using crown, so as to have different
+            # RUSTFLAGS when using `crown`, to reliably trigger re-checking.
+            env['RUSTFLAGS'] = env.get('RUSTFLAGS', "") + " --cfg=crown"
 
         if "-p" not in cargo_args:  # We're building specific package, that may not have features
             features = list(self.features)
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -1017,6 +1017,13 @@ class CommandBase(object):
                     args += ["--lib", "--crate-type=cdylib"]
 
         if use_crown:
+            if 'CARGO_BUILD_RUSTC' in env:
+                current_rustc = env['CARGO_BUILD_RUSTC']
+                if current_rustc != 'crown':
+                    print('Error: `mach` was called with `--use-crown` while `CARGO_BUILD_RUSTC` was'
+                          f'already set to `{current_rustc}` in the parent environment.\n'
+                          'These options conflict, please specify only one of them.')
+                    sys.exit(1)
             env['CARGO_BUILD_RUSTC'] = 'crown'
 
         if "-p" not in cargo_args:  # We're building specific package, that may not have features
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -867,7 +867,7 @@ class CommandBase(object):
                     '--use-crown',
                     default=False,
                     action='store_true',
-                    help='Enable servos `crown` linter tool'
+                    help="Enable Servo's `crown` linter tool"
                 )
             ]
 
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -34,4 +34,3 @@ RUSTC_BOOTSTRAP = "crown,script,style_tests"
 
 [build]
 rustdocflags = ["--document-private-items"]
-rustc = "crown"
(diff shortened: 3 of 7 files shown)


docs/HACKING_QUICKSTART.md
```diff
@@ -206,6 +206,7 @@ the amount of disc space used).
 }
 ```
 
+# Todo: A reviewer should have a look on what needs to be done for NixOS with crown now being optional
```
jschwe, line 209: I believe this section might be able to just be removed, since `crown` is not generally needed anymore - but I'm not 100% sure, so leaving this section up to reviewers.

components/dom_struct/lib.rs
```diff
@@ -13,7 +13,7 @@ pub fn dom_struct(args: TokenStream, input: TokenStream) -> TokenStream {
     }
     let attributes = quote! {
         #[derive(deny_public_fields::DenyPublicFields, domobject_derive::DomObject, JSTraceable, MallocSizeOf)]
-        #[crown::unrooted_must_root_lint::must_root]
```
sagudev, line 16: This is dangerous, why exactly do we need to cfg_attrs for crown lints, IIRC having lints that are not defined does not cause build to fail (I did non crown builds with manually setting `RUSTC` env to `rustc`)

components/script/lib.rs
```diff
@@ -2,15 +2,15 @@
...
-#![feature(register_tool)]
+#![cfg_attr(crown, feature(register_tool))]
 #![deny(unsafe_code)]
 #![doc = "The script crate contains all matters DOM."]
 // Register the linter `crown`, which is the Servo-specific linter for the script
 // crate. Issue a warning if `crown` is not being used to compile, but not when
 // building rustdoc or running clippy.
-#![register_tool(crown)]
-#![cfg_attr(any(doc, clippy, feature = "cargo-clippy"), allow(unknown_lints))]
-#![deny(crown_is_not_used)]
```
sagudev, line 13: See here we abuse unknown_lints lint to get warning if crown is not used.

components/dom_struct/lib.rs
jschwe: I wanted to get rid of `RUSTC_BOOTSTRAP=script` because it would be nice to be able to just use `cargo build` on a stable toolchain to build the project. `#![feature(register_tool)]` is unstable, and if it is removed then `crown` is not found.

sagudev: If `RUSTC_BOOTSTRAP=script` is present in .cargo/config.toml cargo build should still pick it up, right?

I am not really fan of putting crown lints behind cfg as this PR is already dangerous enough (with not running crown by default), but this can be done sometime in the future.

jschwe: > If RUSTC_BOOTSTRAP=script is present in .cargo/config.toml cargo build should still pick it up, right?

yes, that works. I would have just preferred being able to avoid `RUSTC_BOOTSTRAP` completely for release builds. But I'm fine with dropping the `cfg` parts for now. 

> this PR is already dangerous enough

We could also enable `crown` unconditionally in CI if you would prefer (by checking if the `CI` environment variable is set).

sagudev: > We could also enable crown unconditionally in CI if you would prefer (by checking if the CI environment variable is set).

`--use-crown` is good enough.

components/script/lib.rs
sagudev: Let's also keep warning if crown is not used.

components/dom_struct/lib.rs
jschwe: Just to clarify - I only added `--use-crown` to the Linux CI tests, since the lints should be entirely platform independent (based on my limited understanding of crown). Are you okay with that?

components/script/lib.rs
jschwe: done

components/dom_struct/lib.rs
jschwe: ok - done

docs/HACKING_QUICKSTART.md
sagudev: cc @mukilan

python/servo/command_base.py
```diff
@@ -303,6 +303,7 @@ def resolverelative(category, key):
         self.config["build"].setdefault("incremental", None)
         self.config["build"].setdefault("webgl-backtrace", False)
         self.config["build"].setdefault("dom-backtrace", False)
+        self.config["build"].setdefault('use-crown', False)
```
mrobinson, line 306: Can we remove this? We are gradually trying to move away from `mach` so adding as little as possible to the configuration file will make it easier to do this.

python/servo/testing_commands.py
```diff
@@ -203,6 +203,7 @@ def test_unit(self, build_type: BuildType, test_name=None, package=None, bench=F
             args += ["--", "--nocapture"]
 
         env = self.build_env()
+        env['CARGO_BUILD_RUSTC'] = 'crown'
```
mrobinson, line 206: This will mean that people running `test-unit` locally will need to rebuild all of Servo so I think this should not be unconditional.

mrobinson (changes requested):
In addition to the changes here it would be interesting to think about what a good interface for this if we were not using `mach`.

python/servo/testing_commands.py
sagudev: Do we even need this? Crown unit-tests should be independent of this and others do not need crown anyway.

docs/HACKING_QUICKSTART.md
mukilan: I am not sure about this as I don't use VSCode. I guess for developers working on the DOM code, it is useful to have failures from the lint reported in the IDE. In that case I think we should update the documentation in the previous paragraph to pass `--use-crown` to `mach check`.

python/servo/command_base.py
jschwe: Sure, I don't mind. Users hacking on DOM, could just locally `export RUSTC=crown` or edit the `.cargo/config.toml` to avoid manually passing `--use-crown`.

python/servo/testing_commands.py
jschwe: > Do we even need this? Crown unit-tests should be independent of this and others do not need crown anyway.

You are right - the crown unit tests should not depend on the globally installed crown. They do fail if rustc is not overriden though, so I'll investigate what is going wrong in the unit tests.

jschwe:
> In addition to the changes here it would be interesting to think about what a good interface for this if we were not using mach.

What exactly do you mean by "this"? Building servo without `mach` now works for me locally (and running should too - although I didn't test that yet). I believe `mach` will still be required for a couple of things like packaging (since `cargo` doesn't support that natively) and is useful (but not strictly required) for cross-compiling.

If you are referring to using `crown`, I think requiring the user to do `export RUSTC=crown` or editing the `.cargo/config.toml` is a reasonable requirement. It's a bit unfortunate that the `.cargo/config.toml` (unlike `.servobuild`) is committed, so we can't just add it to `.gitignore`. I believe as long as the `.cargo/config.toml` still needs to set `RUSTC_BOOTSTRAP=script` (see [the discussion about removing `RUSTC_BOOTSTRAP`](https://github.com/servo/servo/pull/32494#discussion_r1639731474) ) , making a `config.toml.example` and removing the `config.toml` from git is probably a bad idea.

docs/HACKING_QUICKSTART.md
jschwe: Is `crown` in `PATH` on NixOS? The current instructions tell the user to hardcode the absolute path. (But I'm wondering why this is necessary for vscode, and not in the `.cargo/config.toml`.

sagudev:
The idea I have is that we have good defaults (in `.cargo/config.toml`) that `cargo build` just works as expected (currently it does not do use media-gstreamer) then we offer `mach configure` command that would generate `.cargo/config` based on provided options (like usege of crown) that takes precedence over `.cargo/config.toml` so `cargo build` would now also work with new config. `mach build` would still exist as an alias for cargo build (maybe with some alias options for easy access).

jschwe:
I believe using `.cargo/config` now generates a deprecation warning

docs/HACKING_QUICKSTART.md
mukilan: AFAIK, `crown` will not be in `PATH`. Perhaps @delan will know why VSCode needs this extra configuration.

python/servo/command_base.py
jschwe: removed

docs/HACKING_QUICKSTART.md
jschwe: I adjusted the documentation - please review. I mentioned that `cargo build` can be used, but `mach` is still the recommended tool. From the existing documentation I understood that on NixOS `mach` would still be required.

python/servo/testing_commands.py
jschwe: Interestingly I couldn't reproduce the unit-test failure locally anymore. Lets see what CI says.

docs/HACKING_QUICKSTART.md
```diff
@@ -66,6 +66,8 @@ If you build with `-d`, run with `-d`. If you build with `-r`, run with `-r`.
 ## mach
 
 `mach` is a python utility that does plenty of things to make our life easier (build, run, run tests, update dependencies… see `./mach --help`). Beside editing files and git commands, everything else is done via `mach`.
+In most cases, you can also use `cargo` to build and run `servo`. Common cases where pure `cargo build` does not
```
mukilan, line 69: I'm not sure if we want to officially recommend using `cargo` just yet. There are other cases such as building with GStreamer support that also require using `mach`. @mrobinson, wdyt?

docs/HACKING_QUICKSTART.md
jschwe: Okay, I think I get it now - `vscode` runs outside the nix environment, so `CARGO_BUILD_RUSTC` needs to be set to an absolute path to use `crown`. 
The `mach build` will instead probably use the nix config, so `crown` should then be in path. In that case I guess no documentation updates for the nix part are required.

docs/HACKING_QUICKSTART.md
jschwe: Okay, I dropped the documentation changes for now.

docs/HACKING_QUICKSTART.md
mukilan: That does make sense to me. But I guess `CARGO_BUILD_RUSTC` will be needed only if `rust-analyzer.check.overrideCommand` is not overridden to use  `mach check` as mach will always enter the nix-shell environment.

docs/HACKING_QUICKSTART.md
mukilan: I think it is still helpful to have the text about enabling `crown` when hacking on DOM. 
With the current documentation, we recommend enabling it for NixOS, but not on other platforms where I believe we can update `rust-analyzer.check.overrideCommand`  to pass `--use-crown` to `mach check`.

jschwe: Good point - I updated the documentation to now mention `--use-crown` and how to setup rust-analyzer for vs-code to do the same.

jschwe:
One last thing to discuss: If the user sets `CARGO_BUILD_RUSTC` and passes `--use-crown` at the same time, should 
a) `CARGO_BUILD_RUSTC` always override `--use-crown`
b) `CARGO_BUILD_RUSTC` override `--use-crown` if the value of `CARGO_BUILD_RUSTC` contains  / ends with`crown` 
c) `--use-crown` should always override `CARGO_BUILD_RUSTC`

mukilan:
> One last thing to discuss: If the user sets `CARGO_BUILD_RUSTC` and passes `--use-crown` at the same time, should a) `CARGO_BUILD_RUSTC` always override `--use-crown` b) `CARGO_BUILD_RUSTC` override `--use-crown` if the value of `CARGO_BUILD_RUSTC` contains / ends with`crown` c) `--use-crown` should always override `CARGO_BUILD_RUSTC`
> 
> i think personally I would tend towards choosing a), but I don't have a strong opinion.

I guess overriding makes sense if we have a real use case. Otherwise, I would make it into an error if both are specified so there is less chance of confusion.

mukilan (approved):

python/servo/command_base.py
```diff
@@ -864,6 +864,12 @@ def common_command_arguments(build_configuration=False, build_type=False):
                     help='Build with frame pointer enabled, used by the background hang monitor.',
                 ),
                 CommandArgument('--without-wgl', group="Feature Selection", default=None, action='store_true'),
+                CommandArgument(
+                    '--use-crown',
+                    default=False,
+                    action='store_true',
+                    help='Enable servos `crown` linter tool'
```
mukilan, line 871:
```suggestion
                    help='Enable Servo's `crown` linter tool'
```

jschwe:
> I guess overriding makes sense if we have a real use case. Otherwise, I would make it into an error if both are specified so there is less chance of confusion.

@mukilan I added an error message as you suggested

mukilan:
Thanks @jschwe! I'm too am not sure if `CARGO_BUILD_RUSTC` is strictly required on NixOS, given mach will run the cargo commands inside a nix-shell. I'm am not able to validate that on my system because the rust-analyzer extension seems broken with the` vscode.fhs` package in nix.  I think overriding the variable should be fine for the time being, as we also have a warning.

I've already approved the PR, but I think we can add it to the merge queue after either @sagudev or @mrobinson also approve as they reviewed the majority of the changes.

sagudev:
I will check later today.

sagudev (changes requested):
`mach check` does not work with `--use-crown`

EDIT: but it works with `mach build`

sagudev:
also is there a reason for using CARGO_BUILD_RUSTC instead of RUSTC, I know they should be the same, but if both are configured, which one takes precedence?

jschwe:
> mach check does not work with --use-crown

What is not working for you? Can you share the error message? I can't reproduce.

sagudev:
> > mach check does not work with --use-crown
> 
> What is not working for you? Can you share the error message? I can't reproduce.

When I run with `--use-crown` I get warning that crown is not used, so something must have gone wrong.

jschwe:
> also is there a reason for using CARGO_BUILD_RUSTC instead of RUSTC

It's easier to google for `CARGO_BUILD_RUSTC` and find the [documentation](https://doc.rust-lang.org/cargo/reference/config.html). `RUSTC` does override the `CARGO_BUILD` version, which gives the user more options (a benefit in my opinion, but I guess it could also be viewed as a downside).

```
❯ RUSTC=blah ./mach check --use-crown
error: could not execute process `blah -vV` (never executed)
```

jschwe:
> When I run with mach check --use-crown I get warning that crown is not used, so something must have gone wrong.

It seems setting `CARGO_BUILD_RUSTC` does not invalidate incremental `check`s.

jschwe:
One workaround would be to pass a dummy `--cfg=crown` via `RUSTFLAGS` if `--use-crown` is passed. Different RUSTFLAGS do seem to invalidate the `cargo check` cache.

sagudev:
> One workaround would be to pass a dummy `--cfg=crown` via `RUSTFLAGS` if `--use-crown` is passed. Different RUSTFLAGS do seem to invalidate the `cargo check` cache.

Unfortunately this causes whole recompilation, but I think it worth it (ideally we would only invalidate script crate).

sagudev (approved):

jschwe:
> Unfortunately this causes whole recompilation, but I think it worth it (ideally we would only invalidate script crate).

That is unavoidable, because changing `RUSTC` also forces recompilation (for `cargo build` ). I don't think there is any way to only change the `RUSTC` driver for `script`. Maybe a `RUSTC_WRAPPER` could allow this, but that would also interfere with `sccache`.

In practice, I think that recompilation should not be an issue, since people should just choose one option and then stick with it.

mrobinson (approved):

PR #32497, opened 2024-06-14
Create send_request macro for WebGPURequest

### LINKED ISSUE #32488: Shorter way to warn on ipc send failure
Very common pattern in webgpu code is:

```rust
if let Err(e) = sender.send(WebGPURequest::ComputePassSetBindGroup { ... }) {
       warn!("Error sending ComputePassSetBindGroup with: {e:?}")
 }
```
and it drives me nuts. Instead we should have a macro that would do this automagically (we want to keep warnings in sender's error scope).

OFF-TOPIC: We should probably promote these warnings to errors, as they are ipc errors.

Taym95:
---

- [x] These changes fix #32488

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -152,7 +152,7 @@ use servo_config::{opts, pref};
 use servo_rand::{random, Rng, ServoRng, SliceRandom};
 use servo_url::{Host, ImmutableOrigin, ServoUrl};
 use style_traits::CSSPixel;
-use webgpu::{self, WebGPU, WebGPURequest};
+use webgpu::{self, send_request, WebGPU, WebGPURequest};
 use webrender::{RenderApi, RenderApiSender};
 use webrender_api::DocumentId;
 use webrender_traits::{WebRenderNetApi, WebRenderScriptApi, WebrenderExternalImageRegistry};
@@ -2079,9 +2079,7 @@ where
                         options,
                         ids,
                     };
-                    if webgpu_chan.0.send(adapter_request).is_err() {
-                        warn!("Failed to send request adapter message on WebGPU channel");
-                    }
+                    send_request!(webgpu_chan.0, adapter_request);
                 },
             },
             FromScriptMsg::GetWebGPUChan(response_sender) => {
--- a/components/script/dom/gpuadapter.rs
+++ b/components/script/dom/gpuadapter.rs
@@ -7,7 +7,9 @@ use std::rc::Rc;
 
 use dom_struct::dom_struct;
 use js::jsapi::{Heap, JSObject};
-use webgpu::{wgt, WebGPU, WebGPUAdapter, WebGPURequest, WebGPUResponse, WebGPUResponseResult};
+use webgpu::{
+    send_request, wgt, WebGPU, WebGPUAdapter, WebGPURequest, WebGPUResponse, WebGPUResponseResult,
+};
 
 use super::gpusupportedfeatures::GPUSupportedFeatures;
 use super::types::{GPUAdapterInfo, GPUSupportedLimits};
@@ -88,16 +90,7 @@ impl GPUAdapter {
 
 impl Drop for GPUAdapter {
     fn drop(&mut self) {
-        if let Err(e) = self
-            .channel
-            .0
-            .send(WebGPURequest::DropAdapter(self.adapter.0))
-        {
-            warn!(
-                "Failed to send WebGPURequest::DropAdapter({:?}) ({})",
-                self.adapter.0, e
-            );
-        };
+        send_request!(self.channel.0, WebGPURequest::DropAdapter(self.adapter.0));
     }
 }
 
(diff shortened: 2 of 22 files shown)


sagudev:
I think we can generalize this a bit more (adding optional custom error message using format syntax) and move it to ipc channel as similar pattern is everywhere not just in webgpu.

But firstly let's get some consensus on this. cc @mrobinson @wusyong @gterzian

mrobinson:
I suspect that different modules will want to handle IPC failures in different ways. For instance in the case of WebGPU, it will likely need to shut down the task and return. On the other hand  , the Constellation would probably want to restart the failed tasks.

sagudev:
> I suspect that different modules will want to handle IPC failures in different ways. For instance in the case of WebGPU, it will likely need to shut down the task and return. On the other hand , the Constellation would probably want to restart the failed tasks.

In that cases when error is actually handled we would not use macro, but do manual handling. Macro would only serves as a shorter way to write send and warn on error:
```rust
if let Err(e) = sender.send(Msg) {
    error!("Failed to send Msg due to error: {e}");
   // nothing else here
}
```

Or are you suggesting that we should actually replace all such patterns with actual error handling not just logging, so why bother with making it easier.

mrobinson:
> Or are you suggesting that we should actually replace all such patterns with actual error handling not just logging, so why bother with making it easier.

Yes, I think that at some point failed IPC messages will need to be handled via proper error handling and recovery since a failed IPC message can mean that the task the message is delivering to has disappeared. Maybe in the simplest case that means shutting the entire task down, but in some cases (the constellation for example) more will need to be done.

Taym95:
> > Or are you suggesting that we should actually replace all such patterns with actual error handling not just logging, so why bother with making it easier.
> 
> Yes, I think that at some point failed IPC messages will need to be handled via proper error handling and recovery since a failed IPC message can mean that the task the message is delivering to has disappeared. Maybe in the simplest case that means shutting the entire task down, but in some cases (the constellation for example) more will need to be done.

should I do it part of this PR?

mrobinson:
> should I do it part of this PR?

If it's possible to write a macro to halt execution of the current function, trigger the shutdown of the WebGPU task cleaning up any resources, and then terminating -- that would be excellent, I think!

mrobinson:
> > > should I do it part of this PR?
> > 
> > 
> > If it's possible to write a macro to halt execution of the current function, trigger the shutdown of the WebGPU task cleaning up any resources, and then terminate the task -- that would be excellent, I think!
> 
> What is task in this context? Because if it's just "WebGPU work that was send", that would leave page in unexpected state. Maybe we can just soft-crash the tab (just replacing current page with tab crashed page, so no actual crashing, so cleanup would be done by GC). Given that ipc fails very rarely, this would be nice enough solution.

Does the WebGPU task keep any resources for ongoing Documents? If so, these will need to be cleaned up, even if the WebGPU task continues.

sagudev:
> > > > should I do it part of this PR?
> > > 
> > > 
> > > If it's possible to write a macro to halt execution of the current function, trigger the shutdown of the WebGPU task cleaning up any resources, and then terminate the task -- that would be excellent, I think!
> > 
> > 
> > What is task in this context? Because if it's just "WebGPU work that was send", that would leave page in unexpected state. Maybe we can just soft-crash the tab (just replacing current page with tab crashed page, so no actual crashing, so cleanup would be done by GC). Given that ipc fails very rarely, this would be nice enough solution.
> 
> Does the WebGPU task keep any resources for ongoing Documents? If so, these will need to be cleaned up, even if the WebGPU task continues.

They are cleaned when Drop of associated dom objects is called (dropping GPUBuffer will send `DropBuffer` to wgpu thread), although if the thread is dead (that's probably the reason for IPC failure), everything should already be cleaned up on wgpu side.

mrobinson:
> They are cleaned when Drop of associated dom objects is called (dropping GPUBuffer will send `DropBuffer` to wgpu thread), although if the thread is dead (that's probably the reason for IPC failure), everything should already be cleaned up on wgpu side.

If the script process crashes, `Drop` implementations won't be called so what is the mechanism to ensure that resource are released on the WebGPU-side?

sagudev:
> > They are cleaned when Drop of associated dom objects is called (dropping GPUBuffer will send `DropBuffer` to wgpu thread), although if the thread is dead (that's probably the reason for IPC failure), everything should already be cleaned up on wgpu side.
> 
> If the script process crashes, `Drop` implementations won't be called so what is the mechanism to ensure that resource are released on the WebGPU-side?

There isn't such mechanism (although wgpu internally does some cleanup beforehand, so the only thing that is kept alive is ID). Similar problem might also be present in WebGL.

I think we need to have more broad discussion how we will approach erroring and recovery in servo.

mrobinson:
> I think we need to have more broad discussion how we will approach erroring and recovery in servo.

Yes, I agree. In general, Servo needs to be make much more resilient to process crashes, task panics, and IPC failures. I guess my only concern here is that we shouldn't be hiding away places where we are going to need to add real error handling in the future.

Taym95:
> > I think we need to have more broad discussion how we will approach erroring and recovery in servo.
> 
> Yes, I agree. In general, Servo needs to be make much more resilient to process crashes, task panics, and IPC failures. I guess my only concern here is that we shouldn't be hiding away places where we are going to need to add real error handling in the future.



> > I think we need to have more broad discussion how we will approach erroring and recovery in servo.
> 
> Yes, I agree. In general, Servo needs to be make much more resilient to process crashes, task panics, and IPC failures. I guess my only concern here is that we shouldn't be hiding away places where we are going to need to add real error handling in the future.

I will close

PR #32499, opened 2024-06-14
fonts: Fix loading SFNT tables on Windows

mrobinson:
DirectWrite APIs expect the bytes of table tags to be reversed when
reading them. Servo was doing this when loading font tables, but not all
of them. This led to shaping being broken on Windows. This fixes that
issue in a more comprehensive way and adds a comment to avoid this
failing in the future.

---

- [x] These changes do not require tests because, Windows currently doesn't run WPT tests.

### DIFF
--- a/components/gfx/platform/windows/font.rs
+++ b/components/gfx/platform/windows/font.rs
@@ -27,6 +27,7 @@ use crate::font::{
 };
 use crate::font_cache_thread::FontIdentifier;
 use crate::font_template::FontTemplateDescriptor;
+use crate::ot_tag;
 use crate::text::glyph::GlyphId;
 
 // 1em = 12pt = 16px, assuming 72 points per inch and 96 px per inch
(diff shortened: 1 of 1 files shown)


components/gfx/platform/windows/font.rs
```diff
@@ -142,7 +133,7 @@ impl PlatformFontMethods for PlatformFont {
         //
         // Instead, we do the parsing work using the truetype crate for raw fonts.
         // We're just extracting basic info, so this is sufficient for now.
-        let windows_metrics_bytes = self.face.get_font_table(font_tag!('O', 'S', '/', '2'));
+        let windows_metrics_bytes = self.face.get_font_table(ot_tag!('O', 'S', '/', '2'));
```
mukilan, line 136: `ot_tag!` stores the first character in the most significant byte, whereas the native [DWrite API ](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritefontface-trygetfonttable#parameters) accepts tags constructed using the[ `DWRITE_MAKE_OPENTYPE_TAG` macro](https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-dwrite_make_opentype_tag) which stores first character in the least significant byte. The `get_font_table` wrapper here [seems](https://github.com/servo/dwrote-rs/blob/main/src/font_face.rs#L187) to simply pass along the given tag value, so shouldn't we call  `u32::swap_bytes` here as well?

mrobinson: You're right! I thought falsely that this code was going through the Servo interface to access the table. I've fixed this and verified that the tables are loaded properly here.

mukilan (approved):

PR #32502, opened 2024-06-14
ci: Sign off commits by WPT importer

mrobinson:
This will be important to enforce signing off commits in the project.

---

- [x] These changes do not require tests because they only change CI behavior.

### DIFF
--- a/.github/workflows/scheduled-wpt-import.yml
+++ b/.github/workflows/scheduled-wpt-import.yml
@@ -57,7 +57,7 @@ jobs:
           ./mach update-wpt linux-layout-2013/raw/*.log --legacy-layout
           ./mach update-wpt linux-layout-2020/raw/*.log
           git add tests/wpt/meta tests/wpt/meta-legacy-layout
-          git commit -a --amend --no-edit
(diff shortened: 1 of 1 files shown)


jschwe:
Is the person behind `josh+wptsync@joshmatthews.net` still affiliated with servo? Perhaps we can get a more bot-like email, perhaps pointing to the noreply github email address of the bot?

mrobinson:
Oh, excellent question. We should likely update the address for this GitHub account perhaps to refer to a servo.org address. @jdm @mrego Do you see any issues with this? I assume we'd just use a new address like wpt@servo.org that would forward to the same mailbox as the other servo.org addresses.

mukilan:
We have @servo-bot that we already use for dependabot PR approvals and nightly releases. We could use that for import as well.  @servo-bot uses the email 'ghbot@servo.org' which forwards to infra at servo.org. We also have support for using `ghbot+<tag>@servo.org` if we need to just switch @servo-wpt-sync's email.

mukilan (approved):

jdm:
No concerns with updating the email, but we should update the email of the GitHub account as well.

PR #32515, opened 2024-06-16
ci: Use a servo.org email address for the servo-wpt-sync GitHub bot

mrobinson:
This is the new address of the WPT sync bot.

---

- [x] These changes do not require tests because this just changes CI behavior.

### DIFF
--- a/.github/workflows/scheduled-wpt-import.yml
+++ b/.github/workflows/scheduled-wpt-import.yml
@@ -42,7 +42,7 @@ jobs:
           python3 ./mach bootstrap
       - name: Add upstream remote
         run: |
-          git config --local user.email "josh+wptsync@joshmatthews.net"
+          git config --local user.email "ghbot+wpt-sync@servo.org"
           git config --local user.name "WPT Sync Bot"
           git remote add upstream https://github.com/servo/servo.git
           git fetch --unshallow upstream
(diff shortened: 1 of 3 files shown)


sagudev:
I think this bot is also used for upstream sync, so WPT might need to trust it (like it does current @servo-wpt-sync account).

Although I am not sure exactly why do we need to change the bot persona, as current one works, is trusted and has history (also has nicer icon). If it is about email, we should be able to change it without losing any history information.

mrobinson:
Maybe it's good enough to just change the bots email to something like wpt-sync@servo.org then?

sagudev:
> Maybe it's good enough to just change the bots email to something like [wpt-sync@servo.org](mailto:wpt-sync@servo.org) then?

That would be ideal IMO.

mukilan:
Do need a new `wpt-sync@servo.org` address or can we just use `ghbot+wpt-sync@servo.org`?

mrobinson:
Okay. This is ready for a review.

sagudev:
> Okay. This is ready for a review.

Just to be sure, this email is also set in bot's profile, right?

mrobinson:
> Just to be sure, this email is also set in bot's profile, right?

Yes, as of a few minutes ago, it should be added to the bot's GitHub account

sagudev (approved):

PR #32516, opened 2024-06-16
deps: Switch from `winapi` to `windows-sys` in Servo code

mrobinson:
This is part of the switch from `winapi` to `windows-sys`. `windows-sys` is
maintained by Microsoft, so is more "official." More and more crates are
switching to it.

---


wusyong:
Looks like servo doesn't use win32 API only. Perhaps it can use [windows_sys](https://docs.rs/windows-sys/0.52.0/windows_sys/index.html) instead?

mrobinson:
> Looks like servo doesn't use win32 API only. Perhaps it can use [windows_sys](https://docs.rs/windows-sys/0.52.0/windows_sys/index.html) instead?

Thanks for the advice, this change now uses `windows-sys` instead of `windows`, simplifying things greatly.

mukilan (approved):

PR #32518, opened 2024-06-17
servoshell: Reset link status when loading a new document

### LINKED ISSUE #32517: Link popup doesn't disappear when navigating
**Describe the bug:**
When hovering on top of a link, the default browsers displays the link's url in a small popup on the bottom left of the screen. It doesn't disappear if you click on the link to navigate.

**To Reproduce:**
Go to any web page with a link, and click it. You will see the popup still displaying after navigating to a new page.

**Platform:**
All platforms

webbeef:
Send an empty link status alongside `EmbedderMsg::LoadStart`.

---

- [X] These changes fix #32517
- [X] These changes do not require tests because: I could not find existing tests to update.

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -1038,6 +1038,7 @@ impl Document {
             DocumentReadyState::Loading => {
                 if self.window().is_top_level() {
                     self.send_to_embedder(EmbedderMsg::LoadStart);
+                    self.send_to_embedder(EmbedderMsg::Status(None));
                 }
                 update_with_current_time_ms(&self.dom_loading);
             },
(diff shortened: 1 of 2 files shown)


atbrakhi (commented):
I think this fixes the issue you have reported. However, there is a similar [issue](https://github.com/servo/servo/issues/32517#issuecomment-2172558496), and I am curious if you are interested in fixing that as well.

> When you hover over a link and navigate back and forth using keyboard shortcuts, you can see that the link is still visible even though we are on a different page.

atbrakhi (approved):
Nice, thank you!

PR #32532, opened 2024-06-17
Fix android build on Fedora 40.

jschwe:
* Postfix PATH with android toolchain: We only need to edit path for the linker specified in the `.cargo/config.toml` to be found. Adding the NDK clang to the end of PATH is sufficient for that. Adding the NDK clang to the front can cause problems however, since it causes the NDK `clang` to be preferred over the system clang. This can cause problems on some systems, where compiling e.g. buildscripts for HOST subsequently fails.
* Prefix target compiler and compiler flags variables with `TARGET_` so as not to influence compilation for HOST targets.
 * SET `CLANG_PATH` to avoid [bindgen #2682]

 [bindgen #2682]: https://github.com/rust-lang/rust-bindgen/issues/2682

---

- [x] These changes fix building for android in some cases

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -600,7 +600,7 @@ class CommandBase(object):
         host_cxx = env.get('HOST_CXX') or shutil.which("clang++")
 
         llvm_toolchain = path.join(env['ANDROID_NDK_ROOT'], "toolchains", "llvm", "prebuilt", host)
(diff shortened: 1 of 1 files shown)


jschwe:
Try run in my fork: https://github.com/jschwe/servo/actions/runs/9552982642/job/26330760106

mukilan (approved):
Nice! Thanks for this fix!

PR #32537, opened 2024-06-18
fix: codegen on callback

wusyong:
There are some callbacks with more complicated signatures in `UnderlyingSource` and `QueuingStrategy`.
While we haven't added them yet, I think it's better we can land these fixes first to make sure existing bindings are safe. 

---

- [x] There are tests for these changes: `tests/wpt/mozilla/tests/mozilla/binding_keyword.html`


sagudev:
For such cases we usually create testbindings and write some test for them. Sometimes there is even something in mozilla: https://searchfox.org/mozilla-central/source/dom/bindings/test

wusyong:
But these types are also blocked by SM's stream implementation. Even if we codegen them, I believe servo will still use SM's type in the tests.

jdm (approved):

PR #32540, opened 2024-06-18
mach: fix test-tidy to not skip `Cargo.lock`

### LINKED ISSUE #32530: Dependabot merging PRs that fail servo-tidy
https://github.com/servo/servo/pull/32522

This was automatically merged despite introducing a new duplicate dependency. Is something going wrong here?

mukilan:
PR https://github.com/servo/servo/pull/32465 broke the lint because it initializes FileList with a file name (./Cargo.lock). This causes it to always return an empty list when the `only_changed_files` parameter is `False` since `os.walk` requires a directory and not a file.

Fixes #32530.

---
- [x] These changes fix #32530
- [x] These changes don't have test because  `python/tidy` doesn't support testing git level operations.

### DIFF
--- a/python/tidy/tidy.py
+++ b/python/tidy/tidy.py
@@ -130,6 +130,18 @@ def progress_wrapper(iterator):
         yield thing
 
 
+def git_changes_since_last_merge(path):
+    args = ["git", "log", "-n1", "--committer", "noreply@github.com", "--format=%H"]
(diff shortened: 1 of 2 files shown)


mrobinson (approved):

mrobinson:
Thanks for the fix!

mukilan:
I've pushed a new commit that simply only changes the parameter name.

Issue #31182, opened 2024-01-26
Pass WPT tests for ResizeObserver


gterzian:
Follow-up to https://github.com/servo/servo/pull/31108

Folder is `wpt/tests/resize-observer`. 

- [ ] `calculate-depth-for-node.html` requires `attachShadow`. 
- [ ] `devicepixel.html` calculations are wrong.
- [ ] `eventloop.html` fails intermittently due to Raf delivery. 
- [ ] `fragments.html` doesn't support more than one fragment
- [ ] `notify.html` need to notify the observer when child node appended and removed. 
- [ ] `observe.html` errors due to unsupported `image.decode`
- [ ] `ordering.html` requires https://github.com/servo/servo/issues/31021
- [ ] `scrollbars-2.html` calculations are wrong
- [ ] `scrollbars.html` calculations are wrong
- [ ] `svg-with-css-box-001.html` need to observe svg
- [ ] `svg.html` need to observe svg

gterzian:
Regarding some of the layout queries to make additonal tests pass, see [the comment](https://github.com/servo/servo/pull/31108#discussion_r1470037204):

> Device_pixel_content_box should snap to device pixels. Also, `bounding_content_box_or_zero` is the wrong choice if there are multiple fragments, we don't want the bounding rect. As per the current spec it should just be the 1st fragment, with the idea of extending to an array with all fragments in the future.

gterzian:
Strategy for making all tests pass:

1. Tests that "fail in relation to rafs" can only be fixed after https://github.com/servo/servo/issues/31242
2. All other tests can probably be fixed before.

PR #32546, opened 2024-06-18
Mark some ResizeObserver tests as flaky

mrobinson:
ResizeObserver isn't completed.

---

- [x] These changes do not require tests because this just updates test expectations.

### DIFF
--- a/tests/wpt/meta/resize-observer/iframe-same-origin.html.ini
+++ b/tests/wpt/meta/resize-observer/iframe-same-origin.html.ini
@@ -1,0 +1,2 @@
+[iframe-same-origin.html]
+  expected: [PASS, FAIL]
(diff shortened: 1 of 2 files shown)


gterzian (approved):
thanks for noticing, I've added notes about these in https://github.com/servo/servo/issues/31182.

The relevant notes are:

- iframe-same-origin.html appears intermittent for an unknown reason(I haven't seen it fail locally yet, making it hard to debug so far). See https://github.com/servo/servo/pull/32546
- observe.html At least one test(test7?) appears intermittently to timeout, and when it passes it makes other test fails(as opposed to not run). See https://github.com/servo/servo/issues/32549 for how to make at least three tests pass. Others remain to un-investigated.

mrobinson:
Thanks for the review!

PR #32553, opened 2024-06-19
Implement window.HTMLDocument

### LINKED ISSUE #32536: window.HTMLDocument is not implemented
**Describe the bug:**

The `window.HTMLDocument` property is not implemented in Servo. See https://developer.mozilla.org/en-US/docs/Web/API/HTMLDocument

**To Reproduce:**

Launch servo with `./mach run https://en.wikipedia.org/`

You'll see that in the console:

```
Exception in module-execute in module ext.visualEditor.desktopArticleTarget.init:
(new ReferenceError("HTMLDocument is not defined", "https://en.wikipedia.org/w/load.php?lang=en&modules=ext.visualEditor.core.utils.parsing%7Cext.visualEditor.desktopArticleTarget.init%7Cext.visualEditor.progressBarWidget%2CsupportCheck%2CtargetLoader%2CtempWikitextEditorWidget%2Ctrack%2Cve&skin=vector-2022&version=1vn85", 36))
```

**Platform:**
all.

webbeef:
This adds the webidl and implementation needed for `window.HTMLDocument` . Behavior is similar to Gecko's one.

---

- [X] These changes fix #32536

- [X] These changes do not require tests because there is no wpt tests covering that field.

### DIFF
--- a/components/script/dom/htmldocument.rs
+++ b/components/script/dom/htmldocument.rs
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use std::ptr::NonNull;
+
+use dom_struct::dom_struct;
+use js::jsapi::JSObject;
+
+use crate::dom::bindings::codegen::Bindings::HTMLDocumentBinding::HTMLDocumentMethods;
+use crate::dom::bindings::error::{Error, Fallible};
+use crate::dom::bindings::reflector::Reflector;
+use crate::dom::bindings::root::DomRoot;
+use crate::dom::bindings::str::DOMString;
+use crate::dom::location::Location;
+use crate::script_runtime::JSContext as SafeJSContext;
+
+// https://html.spec.whatwg.org/multipage/#htmldocument
+#[dom_struct]
+pub struct HTMLDocument {
+    reflector_: Reflector,
+}
+
+impl HTMLDocumentMethods for HTMLDocument {
+    // https://html.spec.whatwg.org/multipage/#htmldocument
+    fn GetLocation(&self) -> Option<DomRoot<Location>> {
+        None
+    }
+
+    // https://html.spec.whatwg.org/multipage/#htmldocument
+    fn SupportedPropertyNames(&self) -> Vec<DOMString> {
+        Vec::new()
+    }
+
+    // https://html.spec.whatwg.org/multipage/#htmldocument
+    fn NamedGetter(
+        &self,
+        _cx: SafeJSContext,
+        _name: DOMString,
+    ) -> Fallible<Option<NonNull<JSObject>>> {
+        Err(Error::NotFound)
+    }
+}
--- a/components/script/dom/mod.rs
+++ b/components/script/dom/mod.rs
@@ -380,6 +380,7 @@ pub mod htmldialogelement;
 pub mod htmldirectoryelement;
 pub mod htmldivelement;
 pub mod htmldlistelement;
+pub mod htmldocument;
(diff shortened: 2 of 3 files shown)


components/script/dom/htmldocument.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    // https://html.spec.whatwg.org/multipage/#htmldocument
+    fn NamedGetter(
+        &self,
+        _cx: SafeJSContext,
+        _name: DOMString,
+    ) -> Fallible<Option<NonNull<JSObject>>> {
+        Err(Error::NotFound)
+    }
```
mrobinson, line 42: Could these not be implemented?

mrobinson (changes requested):

components/script/dom/htmldocument.rs
mrobinson: Is it not enough to simply ensure that window has an HTMLDocument getter that does the same thing as `fn Document(...)` in `window.rs`?

components/script/dom/htmldocument.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use crate::dom::bindings::codegen::Bindings::HTMLDocumentBinding::HTMLDocumentMethods;
+use crate::dom::bindings::error::{Error, Fallible};
+use crate::dom::bindings::reflector::Reflector;
+use crate::dom::bindings::root::DomRoot;
+use crate::dom::bindings::str::DOMString;
+use crate::dom::location::Location;
+use crate::script_runtime::JSContext as SafeJSContext;
+
+// https://html.spec.whatwg.org/multipage/#htmldocument
```
Taym95, line 18: can you follow other doc link and use this pattern: `/// <link>` for all doc link in this PR?

components/script/dom/htmldocument.rs
mrobinson: Reading through https://github.com/whatwg/html/issues/4792 this seems to be the situation:

 - The specification says that `HTMLDocument` should be an alias for `Document` ie we should not need to implement a separate document interface. The `Document` interface implements the named getter (Servo gained support for this unfortunate part of the web platform last year).
 - No browsers do this though. Instead FireFox has `window.document` implement `HTMLDocument`, but not for XML documents. This means that `responseXML` and other XML pages do not have a named getter, which seems to big the biggest compatability risk.

This change is quite strange though. It implements `HTMLDocument`, but `window.document` is still not an HTMLDocument. If we want to follow Firefox, it seems like the named getter should move to `HTMLDocument` and `HTMLDocument` should be exposed at `window.document` when the document is HTML (but not XML).

mrobinson (changes requested):

mrobinson:
> @mrobinson I think you're not talking about the same spec (the whole naming is very confusing). What is implemented here is https://html.spec.whatwg.org/multipage/nav-history-apis.html#htmldocument :

I believe this is the same specification and the part that you quoted is the bit that says that `HTMLDocument` should be an alias of `Document`. In Firefox `window.document instanceof HTMLDocument == true`.

jdm:
> > I believe this is the same specification and the part that you quoted is the bit that says that `HTMLDocument` should be an alias of `Document`. In Firefox `window.document instanceof HTMLDocument == true`.
> 
> You're right. So in Servo we would need `Document` from `components/script/dom/document.rs` to implement `HTMLDocument`? A naive approach of just doing `impl HTMLDocumentMethods for Document { ... }` doesn't work. I'll try to figure out how the instanceof check is wired up.

A couple things:
1) the use of Document::new will need to be HTMLDocument::new in appropriate cases (and the code that creates a JS reflector for HTMLDocument will need to be written)
2) the first field of HTMLDocument will need to be Document, not Reflector.

yezhizhen:
Closing due to #38433.

Issue #8371, opened 2015-11-06
Handle generic font-family names


SimonSapin:
https://drafts.csswg.org/css-fonts/#propdef-font-family
https://drafts.csswg.org/css-fonts/#generic-font-families

In the `font-family` property, the `serif`, `sans-serif`, `cursive`, `fantasy`, and `monospace` keywords represent generic families when they’re not in quotes.
- Uncomment [this code](https://github.com/servo/servo/blob/9fea6d2e46dd917f16a5f1ec0f6484e8164c7a3a/components/style/properties.mako.rs#L1734-L1738) and [this code](https://github.com/servo/servo/blob/9fea6d2e46dd917f16a5f1ec0f6484e8164c7a3a/components/style/properties.mako.rs#L1785-L1792) in `components/style/properties.mako.rs`
- Change [`get_initial_value()`](https://github.com/servo/servo/blob/9fea6d2e46dd917f16a5f1ec0f6484e8164c7a3a/components/style/properties.mako.rs#L1772) to return `FontFamily::Serif`
- Handle the new variants in `components/gfx/font_context.rs`. (I have no idea what this involves.)

SimonSapin:
See also occurrences of "generic" in `components/gfx/font_cache_task`. Since there is a fixed set of generic families, maybe the hash map in `FontCache::generic_fonts` should be replaced with a `match` expression on the style crate’s `FontFamily` enum.

SimonSapin:
CC @Ms2ger @pcwalton @glennw for winning the `git shortlog -sn components/gfx/font*` lottery.

craftytrickster:
If possible, I would like to attempt this.

eefriedman:
Okay; feel free to ask if you have any questions.

craftytrickster:
@eefriedman Thank you.
@SimonSapin Thank you for the explanation of the issue. I have made some preliminary changes which I will put in a pull request.

As for handling the variants in `components/gfx/font_context.rs`, I am not sure anything needs to be done there, since it is not directly interacting with the FontFamily enum. 

As for the modifying `components/gfx/font_cache_task`, I did replace the plain string instances with instances of the FontFamily enum, however, I did not remove the HashMap. My reasoning is because when `transform_family` is called, if we did not use the HashMap we would have to calculate the mapped family every time instead of doing it one time upfront during hashmap population.

PR #32555, opened 2024-06-19
ohos: Add fallback font for serif

jschwe:
The fallback behavior seems to have changed recently (sometime in the last 2 months). Previously we didn't need to specify any fallback for `serif`, and text requesting serif would still render. Now we need to explicitly add a fallback for `serif` otherwise no font is selected, and the text is not rendered.

Background: 

I don't see any fonts on ohos which a) don't have `Sans` in the name and b) are not some special case font, so I think we don't have much of a choice but to fallback to Harmony OS Sans. Using `Harmony OS Sans`  is probably always the preferred choice on ohos devices, except for the languages which are not supported by it (and maybe Emojis).

I'm not sure if adding an explicit fallback here is the preferred solution, or if servo should instead always try to fallback to some other font family if the requested one is not present.

### DIFF
--- a/components/fonts/platform/freetype/ohos/font_list.rs
+++ b/components/fonts/platform/freetype/ohos/font_list.rs
@@ -80,6 +80,8 @@ impl FontList {
         let alternatives = [
             ("HarmonyOS Sans", "HarmonyOS_Sans_SC_Regular.ttf"),
             ("sans-serif", "HarmonyOS_Sans_SC_Regular.ttf"),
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

mrobinson:
This looks fine to me. We are still pending to correct the way that generic families are interpreted: #8371. One would hope that this would make too much platform-specific code obsolete.

mrobinson (approved):

PR #32564, opened 2024-06-19
servoshell: Clean up keyboard event conversion a little

mrobinson:
- Use a trait to make all conversions standard
- Alphabetize all enums so they are in a consistent order
- Remove some comments about unhandled which no longer apply after the
  winit upgrade.
- Use the winit `Location` property to get the `keyboard_types`
  Location instead of relying on the physical key.

---


jdm (approved):
Nice cleanup!

PR #32572, opened 2024-06-20
Add QueuingStrategy and UnderlyingSource

wusyong:
This PR add QueuingStrategy and UnderlyingSource. The related types are also added but with `todo!()` implementations.

---

- [x] These changes do not require tests because SM stream is still enabled. This PR aims to add the implementation incrementatlly.


components/script/dom/readablebytestreamcontroller.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct ReadableByteStreamController {
+    reflector_: Reflector,
+}
+
+impl ReadableByteStreamControllerMethods for ReadableByteStreamController {
+    fn GetByobRequest(
+        &self,
+    ) -> Fallible<Option<DomRoot<super::readablestreambyobrequest::ReadableStreamBYOBRequest>>>
+    {
+        todo!()
```
mrobinson, line 36: Can you avoid using `todo()!` in DOM code because this can cause the script thread to panic?

wusyong:
@gterzian Here's the first PR to `readablestream-re-implementation`. Could you see if the implementation looks good to you?

gterzian (changes requested):
Ok just one thing that would be nice to do before merging. Also, as said we want to move the unsafe code into `dom/bindings`, but we can do that in a follow-up(cc @taym95 that may be a good one for you)

components/script/dom/readablestream.rs
```diff
@@ -59,6 +62,36 @@ pub struct ReadableStream {
 }
 
 impl ReadableStream {
+    #[allow(non_snake_case)]
+    // https://streams.spec.whatwg.org/#rs-constructor
+    pub fn Constructor(
+        cx: SafeJSContext,
+        _global: &GlobalScope,
+        _proto: Option<SafeHandleObject>,
+        underlying_source: Option<*mut JSObject>,
```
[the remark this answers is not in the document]
wusyong: But this is the signature required by the codegen.
And I think the spec also says it will convert to `UnderlyingSource` in step 2.
https://streams.spec.whatwg.org/#rs-constructor

components/script/dom/bytelengthqueuingstrategy.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[dom_struct]
+pub struct ByteLengthQueuingStrategy {
+    reflector_: Reflector,
+    high_water_mark: f64,
+}
+
+#[allow(non_snake_case)]
+impl ByteLengthQueuingStrategy {
+    // https://streams.spec.whatwg.org/#blqs-constructor
```
Taym95, line 32: better to use `/// <link>` for doc link comment

components/script/dom/countqueuingstrategy.rs
```diff
@@ -0,0 +1,158 @@
...
+            );
+            assert!(!raw_fun.is_null());
+
+            // Step 3. Set globalObject’s count queuing strategy size function to
+            // a Function that represents a reference to F,
+            // with callback context equal to globalObject’s relevant settings object.
+            let fun_obj = JS_GetFunctionObject(raw_fun);
+            let fun = Function::new(cx, fun_obj);
+            global.set_count_queuing_strategy_size(fun.clone());
+            Ok(fun)
```
Taym95, line 89: This is duplicated, but we will take care of it in another PR when we move this `unsafe_code` to `dom/bindings`.

wusyong: Yeah current `Function` creation still requires us to write unsafe c function. Maybe we can improve `Function` later.

components/script/dom/readablestream.rs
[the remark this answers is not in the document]
gterzian: Right, ok.

gterzian (approved):
Great! Let's move the unsafe code do `dom/bindings` in a subsequent step...

PR #32573, opened 2024-06-20
android: Remove serde-json build-dependency

jschwe:
There is no need to add `serde-json` as a
build-dependency (which causes serde to be built
twice when cross-compiling - once for host and once for the target)

---
- [x] These changes fix: remove an unneeded build-time dependency when cross-compiling


mrobinson (approved):
Nice cleanup.

Issue #31973, opened 2024-04-02
Fix Iframe loading


gterzian:
As part of debugging at new failing test in https://github.com/servo/servo/pull/31505(), I noted some(known?) problems with how iframes are loaded.

The test in question is `/html/browsers/browsing-the-web/navigating-across-documents/initial-empty-document/load-pageshow-events-iframe-contentWindow.html`, which asserts that **no load events are fired** for an iframe which has no navigable("load" & "pageshow" events do not fire on contentWindow of iframe that stays on the initial empty document). 

We fail all tests, meaning we fire load events, for all subtests, with [the exception of](https://github.com/servo/servo/blob/da3991c8f37251e04ba67dd7e71c3129327dddae/tests/wpt/meta/html/browsers/browsing-the-web/navigating-across-documents/initial-empty-document/load-pageshow-events-iframe-contentWindow.html.ini#L4) "load & pageshow events do not fire on contentWindow of <iframe> element created with src=''". So, for some reason we do not fire the load event, at least not within 50ms, in that case(it's not clear to me why). The test fails now also with https://github.com/servo/servo/pull/31505, hence my investigation, but I think I may want to let it fail, and instead fix all tests in one go.

Quick recap on how this relates to the spec:

1. Entry point for processing are [iframe HTML element insertion steps](https://html.spec.whatwg.org/multipage/#the-iframe-element:html-element-insertion-steps), which call into [process the iframe attributes](https://html.spec.whatwg.org/multipage/#process-the-iframe-attributes).
2. We then hit the "2 otherwise"(meaning it has a url), which then either:
3. Immediately returns if there is no src, which means not firing any load event. 
4. Immediately run the load steps for an "about:blank" iframe, which means **firing a load event on the element, not the content window**
5. Navigate the iframe otherwise, which means eventually firing the load event on the content window as per [completely-finish-loading](https://html.spec.whatwg.org/multipage/document-lifecycle.html#completely-finish-loading)

So, in other words, there are three:
- in the case of no src, no event should fire at all, and no navigation should happen. 
- in the case of "about:blank", a load event should synchronously fire on the element, but no navigation should happen. 
- Otherwise, we should navigate the iframe, and then only fire the load event when the load completes(so that's asynchronous). 

-------

What we are currently doing, and why it is wrong:

[...]

PR #32574, opened 2024-06-20
Fix ordering of documents

### LINKED ISSUE #32004: Update the rendering: fix ordering of documents
Follow-up to https://github.com/servo/servo/pull/31505

See step two of [spec](https://html.spec.whatwg.org/multipage/#update-the-rendering), and https://github.com/servo/servo/pull/31505/files#diff-2c5948584719622eef023c56cdf90e8278971f2f2eda26e7fd0485a24c32be3cR1735

I've tried doing this in https://github.com/servo/servo/pull/31505/commits/fa48454f64f01ceff17a6959c170914d46586da4, but this resulted in various unexpected test results(which I didn't note, sorry), and so it will require some further investigation. Note that our current implementation does not order docs in any way, since it usually simply handles a message immediately when received by the script-thread.

Taym95:
This PR fixes two things related to the update the rendering concept:

- Updating the rendering is now done according to the document ordering requirements of the spec(see test expectation update).
- The batching of rendering opportunities is now scoped by pipeline, to avoid missing one following closure of a pipeline([details](https://github.com/servo/servo/pull/32574#issuecomment-2275379106)). 

---

- [x] These changes fix #32004

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -126,7 +126,7 @@ use crate::dom::htmlanchorelement::HTMLAnchorElement;
 use crate::dom::htmliframeelement::HTMLIFrameElement;
 use crate::dom::identityhub::Identities;
 use crate::dom::mutationobserver::MutationObserver;
-use crate::dom::node::{document_from_node, window_from_node, Node, ShadowIncluding};
+use crate::dom::node::{window_from_node, Node, ShadowIncluding};
 use crate::dom::performanceentry::PerformanceEntry;
 use crate::dom::performancepainttiming::PerformancePaintTiming;
 use crate::dom::serviceworker::TrustedServiceWorkerAddress;
@@ -1649,25 +1649,28 @@ impl ScriptThread {
         // TODO: The specification says to filter out non-renderable documents,
         // as well as those for which a rendering update would be unnecessary,
         // but this isn't happening here.
-        let pipeline_and_docs: Vec<(PipelineId, DomRoot<Document>)> = self
+        let pipelines_to_update: Vec<PipelineId> = self
             .documents
             .borrow()
             .iter()
             .filter(|(_, document)| document.window().is_top_level())
-            .map(|(id, document)| (id, DomRoot::from_ref(&*document)))
             .flat_map(|(id, document)| {
-                let mut documents = vec![(id.clone(), document.clone())];
-                documents.extend(document.iter_iframes().filter_map(|iframe| {
-                    iframe
-                        .pipeline_id()
-                        .map(|id| (id, document_from_node(&*iframe)))
-                }));
+                let mut documents = vec![id.clone()];
+                documents.extend(
+                    document
+                        .iter_iframes()
+                        .filter_map(|iframe| iframe.pipeline_id()),
+                );
                 documents
-            })
-            .collect();
+            });
+        //println!("Second Documents: {:?}", start);
         // Note: the spec reads: "for doc in docs" at each step
         // whereas this runs all steps per doc in docs.
-        for (pipeline_id, document) in pipeline_and_docs {
+        for pipeline_id in pipelines_to_update {
+            let Some(document) = self.documents.borrow().find_document(pipeline_id) else {
+                warn!("Updating the rendering for closed pipeline {pipeline_id}.");
+                continue;
+            };
             // TODO(#32004): The rendering should be updated according parent and shadow root order
             // in the specification, but this isn't happening yet.
 
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -1662,8 +1662,8 @@ impl ScriptThread {
                         .filter_map(|iframe| iframe.pipeline_id()),
                 );
                 documents
-            });
-        //println!("Second Documents: {:?}", start);
+            })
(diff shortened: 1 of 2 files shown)


gterzian:
As per the unexpected timeouts, the re-ordering must somehow filter out iframes. Note that we don't want to filter-out any, we only want to re-order them. I think the nested `filter_map` call is perhaps to blame...

components/script/script_thread.rs
```diff
@@ -1654,7 +1654,17 @@ impl ScriptThread {
             .documents
             .borrow()
             .iter()
+            .filter(|(_, document)| document.window().is_top_level())
             .map(|(id, document)| (id, DomRoot::from_ref(&*document)))
+            .flat_map(|(id, document)| {
+                let mut documents = vec![(id.clone(), document.clone())];
+                documents.extend(document.iter_iframes().filter_map(|iframe| {
+                    iframe
+                        .pipeline_id()
```
gterzian, line 1663: Note that our implementation of iframe is broken, and that https://github.com/servo/servo/blob/66edef806579fd0b386f4ceba473e6a9f7d0ca34/components/script/dom/htmliframeelement.rs#L462 likely returns `None` when it shouldn't. 

Perhaps a solution is to put the documents with not pipeline ids as last. (you can get the docs from the `documents` above).

gterzian (changes requested):

components/script/script_thread.rs
Taym95: @gterzian found out that timeout it caused by `.filter(|(_, document)| !document.window().is_top_level())` not the filter_map

Taym95: this is result after removing `is_top_level`, not timeout but wrong result:

image

gterzian: Ok good, sounds like reversing `iter_iframes` is the way to go then!

Taym95: @gterzian did you have time to look at this?

gterzian:
So the problem is essentially https://github.com/servo/servo/issues/31973

The [test](/html/webappapis/update-rendering/child-document-raf-order.html) creates two iframes, but we actually create four documents(2 for each iframe: one "about:blank" and another), and then we remove the "about:blank" ones when the "real" ones finish navigating. I think the timeout is because the raf end-up being lost because the "about:blank" are not returned anymore as part of `iter_iframes`(while there is a time where even though they are not part of `iter_frames`, they are still inside `documents`) and then they the rafs callbacks are not called. 

So for this we actually first need to fix https://github.com/servo/servo/issues/31973

gterzian:
Ok so actually I had another look, and the problem was rather that https://github.com/servo/servo/blob/bc1c71bd4df9424e4cceed9dc8cc38efc591095b/components/script/dom/node.rs#L3151 returns the owner doc, not the doc of the iframe...

Taym95:
> Ok so actually I had another look, and the problem was rather that
> 
> https://github.com/servo/servo/blob/bc1c71bd4df9424e4cceed9dc8cc38efc591095b/components/script/dom/node.rs#L3151
> 
> returns the owner doc, not the doc of the iframe...

Thanks for checking this, what do you mean by owner doc? doc os iframe owner?

jdm:
Yeah, the document that contains the iframe element.

gterzian:
Ok I think we can merge this, perhaps @Taym95 take a look at the code I wrote, not sure if perhaps we can avoid some allocations while achieving the same result.

Taym95:
> Ok I think we can merge this, perhaps @Taym95 take a look at the code I wrote, not sure if perhaps we can avoid some allocations while achieving the same result.

Yeah I will clean it up and let's merge it.

components/script/script_thread.rs
```diff
@@ -1649,15 +1649,26 @@ impl ScriptThread {
...
         // but this isn't happening here.
-        let pipeline_and_docs: Vec<(PipelineId, DomRoot<Document>)> = self
+        let pipelines_to_update: Vec<PipelineId> = self
             .documents
             .borrow()
             .iter()
-            .map(|(id, document)| (id, DomRoot::from_ref(&*document)))
+            .filter(|(_, document)| document.window().is_top_level())
+            .flat_map(|(id, document)| {
+                std::iter::once(id.clone()).chain(
```
Taym95, line 1658: I tried to used `iter::once` here with `chain' to avoid vec allocation

Taym95, line 1658: @gterzian I tried to used `iter::once` here with `chain' to avoid vec allocation.

gterzian:
I also got the below failure, and some timeouts, but only when running the test in release mode:

gterzian:
hypothesis: we fail to run a rendering task, because the pipeline for which is was queued was closed, but after the last `note_rendering_opportunity`, which did not queue a task because one was queued for the subsequently closed pipeline.

gterzian (approved):
It seems that the [hypothesis](https://github.com/servo/servo/pull/32574#issuecomment-2275379106), tested by my latest commit which scopes the batching of rendering updates by pipeline, has been validated by the two latest test runs. The intermittency of the problem makes it practically impossible to validate through debugging locally. Let's merge this.

PR #32575, opened 2024-06-20
webgpu: Update wgpu and revamp computepass

sagudev:
wgpu's compute pass object now lives only in wgpu thread, functions return compute pass as object instead of an id, so we need to do id mapping ourself.

---


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/9602065915/job/26483847356

> Unexpected validation error occurred: At least one color attachment or depth-stencil attachment was expected, but no render target for the pipeline was specified.

is actually expected per spec: https://github.com/gfx-rs/wgpu/pull/5715

Cargo.toml
```diff
@@ -135,8 +135,8 @@ webpki-roots = "0.25"
 webrender = { git = "https://github.com/servo/webrender", branch = "0.64", features = ["capture"] }
 webrender_api = { git = "https://github.com/servo/webrender", branch = "0.64" }
 webrender_traits = { path = "components/shared/webrender" }
-wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "d0a5e48aa7e84683114c3870051cc414ae92ac03" }
-wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "d0a5e48aa7e84683114c3870051cc414ae92ac03" }
+wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "1e784c9c0af7a785b62e3b6840ed012a7477520f" }
+wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "1e784c9c0af7a785b62e3b6840ed012a7477520f" }
```
mrobinson, line 139: Is it possible to switch to a release of wgpu instead of using a git hash?

sagudev: No, unfortunately not. Releases are cut too slowly and last ones had some nasty bugs. Also I do some work on wgpu and keeping servo up to date makes it easier to squash bugs in servo and wgpu at the same time.

BTW firefox also uses git rev instead of releases updated weekly, for similar reasons.

sagudev:
More work on command buffers/encoders is left for a followup PR.

components/webgpu/wgpu_thread.rs
```diff
@@ -218,7 +223,9 @@ impl WGPU {
                             Ok(())
                         };
 
+                        // invalidate command buffer too
                         self.encoder_record_error(command_encoder_id, &result);
```
gterzian, line 227: what do you mean? Also a nit for another time: that method should also be called `maybe_`.

components/script/dom/gpudevice.rs
```diff
@@ -769,6 +771,7 @@ impl GPUDeviceMethods for GPUDevice {
             Some(wgpu_pipe::RenderPipelineDescriptor {
                 label: convert_label(&descriptor.parent.parent),
                 layout,
+                cache: None,
```
gterzian, line 774: So the idea is this will move to the backend thread as well right?

gterzian (changes requested):
Couple of things...

components/webgpu/wgpu_thread.rs
```diff
@@ -73,13 +73,19 @@ pub(crate) struct WGPU {
...
+    // (this is also reused for invalidation of command buffers)
+    error_command_encoders: HashMap<id::CommandEncoderId, String>,
     webrender_api: Arc<Mutex<RenderApi>>,
     webrender_document: DocumentId,
     external_images: Arc<Mutex<WebrenderExternalImageRegistry>>,
     wgpu_image_map: Arc<Mutex<HashMap<u64, PresentationData>>>,
     poller: Poller,
+    // We need to store passes here
+    compute_passes: HashMap<ComputePassId, Box<dyn DynComputePass>>,
+    //render_passes: HashMap<RenderPassId, Box<dyn DynRenderPass>>,
```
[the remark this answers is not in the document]
sagudev: Wgpu will also change how renderpass works, so we will need this in near future.

components/script/dom/gpudevice.rs
sagudev: Yes, when wgpu changes how renderpasses work.

components/webgpu/wgpu_thread.rs
sagudev: Specs says that we need to invalidate command encoder on error, but wgpu does not do this (yet). So we consider command encoder invalid if there is error stored in hashmap of error encoders. When command encoder is finished into command buffer (wgpu uses same id via id transmutes) we need to make sure that invalid command encoder also finishes into invalid command buffer (or if there is error when doing finish) so we reuse same error hashmap for this.

components/webgpu/wgpu_thread.rs
```diff
@@ -745,35 +751,87 @@ impl WGPU {
...
+                            let global = &self.global;
+                            let result = gfx_select!(command_encoder_id => global.render_pass_end(&render_pass));
+                            self.maybe_dispatch_wgpu_error(device_id, result.err())
                         } else {
-                            Err(String::from("Invalid RenderPass"))
-                        };
-                        self.encoder_record_error(command_encoder_id, &result);
+                            self.dispatch_error(
+                                device_id,
+                                Error::Validation("Render pass already ended".to_string()),
```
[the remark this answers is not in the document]
gterzian: Well, I guess `!valid` means ended, actually...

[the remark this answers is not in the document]
sagudev: Actually, this validation in script (content time-line) should actually be done in webgpu thread (usually it's done within wgpu). That's actually the gist of what needs to be done in servo about webgpu, we have a lot of logic in script that should be in device timeline (webgpu thread), usually already in wgpu. Maybe the specs were different when the code was written.

Here I did some ground work for changes of render pass that will come (mostly because some changes around it are already included in wgpu), but I plan to do complete refresh of RenderPass (like I did in for compute pass here), when all necessary changes land in wgpu.

components/script/dom/gpucomputepassencoder.rs
```diff
@@ -79,57 +77,79 @@ impl GPUComputePassEncoderMethods for GPUComputePassEncoder {
...
+                compute_pass_id: self.compute_pass.0,
+                device_id: self.command_encoder.device_id().0,
             })
-            .expect("Failed to send RunComputePass"); //TODO: handle error
-
-        self.command_encoder.set_state(
-            GPUCommandEncoderState::Open,
-            GPUCommandEncoderState::EncodingComputePass,
-        );
+            .expect("Failed to send EndComputePass"); //TODO: handle error
```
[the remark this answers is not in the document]
sagudev: Because is just refactored old code, but I will fix it.

components/webgpu/wgpu_thread.rs
```diff
@@ -745,35 +751,87 @@ impl WGPU {
...
                     } => {
                         let global = &self.global;
-                        let result = if let Some(pass) = compute_pass {
-                            gfx_select!(command_encoder_id => global.command_encoder_run_compute_pass(
+                        let (pass, _) = gfx_select!(
+                            command_encoder_id => global.command_encoder_create_compute_pass_dyn(
                                 command_encoder_id,
-                                &pass
-                            )).map_err(|e| format!("{:?}", e))
+                                &ComputePassDescriptor { label: None, timestamp_writes: None }
```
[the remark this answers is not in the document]
sagudev: Generally we just pass wgpu descriptors from script, but `ComputePassDescriptor` is not serializable.

components/script/dom/gpucommandencoder.rs
```diff
@@ -122,32 +105,24 @@ impl GPUCommandEncoderMethods for GPUCommandEncoder {
...
-                        .label
-                        .as_ref()
-                        .map(|l| Cow::Borrowed(&**l)),
-                    timestamp_writes: None,
-                },
-            ))
-        };
+        if let Err(e) = self.channel.0.send(WebGPURequest::BeginComputePass {
+            command_encoder_id: self.id().0,
+            compute_pass_id,
```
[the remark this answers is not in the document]
sagudev: I will only put label field as we do not support timestamps yet.

components/webgpu/wgpu_thread.rs
```diff
@@ -745,35 +751,87 @@ impl WGPU {
...
+                        ));
+                        self.compute_passes.insert(compute_pass_id, pass);
+                    },
+                    WebGPURequest::ComputePassSetPipeline {
```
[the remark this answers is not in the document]
sagudev: When we end pass we must drop it, or else it will still hold resources (and queue_submit will panic).

sagudev:
Hm, some validation still fails, apparently command encoders are not invalidated on invalid passes ending.

components/webgpu/wgpu_thread.rs
```diff
@@ -84,7 +84,7 @@ pub(crate) struct WGPU {
     wgpu_image_map: Arc<Mutex<HashMap<u64, PresentationData>>>,
     poller: Poller,
     // We need to store passes here
-    compute_passes: HashMap<ComputePassId, Box<dyn DynComputePass>>,
+    compute_passes: HashMap<ComputePassId, (Box<dyn DynComputePass>, bool)>,
```
gterzian, line 87: A doc for `valid` would be good.

gterzian (approved):
LGTM, with one nit

sagudev:
Currently `webgpu:api,validation,encoding,encoder_state:call_after_successful_finish:*` fails due to different encoder state handling in wgpu, we could probably write more logic to make it work, but I would rather work on fixing wgpu itself.

components/webgpu/wgpu_thread.rs
sagudev: Done!

components/webgpu/wgpu_thread.rs
```diff
@@ -74,16 +74,17 @@ pub(crate) struct WGPU {
...
+    /// (this is also reused for invalidation of command buffers)
     error_command_encoders: HashMap<id::CommandEncoderId, String>,
     webrender_api: Arc<Mutex<RenderApi>>,
     webrender_document: DocumentId,
     external_images: Arc<Mutex<WebrenderExternalImageRegistry>>,
     wgpu_image_map: Arc<Mutex<HashMap<u64, PresentationData>>>,
+    /// Provides access to poller thread
     poller: Poller,
-    // We need to store passes here
+    /// Store compute passes (that have not ended yet) and their validness
```
gterzian, line 87: I think `validity` is the right word.

components/webgpu/wgpu_thread.rs
```diff
@@ -1165,10 +1166,8 @@ impl WGPU {
                         };
                     },
                     WebGPURequest::DropComputePass(id) => {
-                        assert!(
-                            self.compute_passes.remove(&id).is_some(),
-                            "ComputePass should exist"
-                        );
+                        // Compute pass might already ended
```
gterzian, line 1166: needs a "have".

gterzian (approved):
LGMT, few nits again.

PR #32576, opened 2024-06-21
script: Implement the `FontFaceSet` DOM API

mukilan:
Add a skeleton implementation of FontFaceSet interface with support
for resolving the `document.fonts.ready` Promise when the loading of
web fonts is completed.

This change exposes new failures in the web platform tests.

These were ERROR before the change because `document.fonts.ready` caused
a `ReferenceError` causing the tests to be aborted and they now FAIL:
- /css/CSS2/linebox/vertical-align-top-bottom-001.html
- /css/css-flexbox/flex-one-sets-flex-basis-to-zero-px.html
- /css/css-fonts/generic-family-keywords-001.html
- /css/css-fonts/math-script-level-and-math-style/math-script-level-004.tentative.html
- /css/css-fonts/math-script-level-and-math-style/math-script-level-002.tentative.html
- /css/css-text/text-autospace/text-autospace-ligature-001.html
- /css/css-values/calc-size/calc-size-width.tentative.html

These were TIMEOUT before the change because `document.fonts.ready` was
a ReferenceError and the tests were asynchronous (reftest-wait). These now
FAIL because the assertions are now executed after fonts are loaded:
- /css/css-fonts/matching/fixed-stretch-style-over-weight.html
- /css/css-fonts/matching/range-descriptor-reversed.html
- /css/css-fonts/matching/stretch-distance-over-weight-distance.html
- /css/css-fonts/matching/style-ranges-over-weight-direction.html
- /css/css-fonts/variations/variable-box-font.html
- /css/css-fonts/variations/variable-gpos-m2b.html
- /css/css-fonts/variations/variable-gsub.html
- /css/css-fonts/variations/variable-opsz-size-adjust.html
- /css/css-position/sticky/position-sticky-change-top.html
- /css/css-position/sticky/position-sticky-fixed-ancestor.html
- /css/css-position/sticky/position-sticky-flexbox.html
- /css/css-position/sticky/position-sticky-grid.html
- /css/css-position/sticky/position-sticky-inline.html
- /css/css-position/sticky/position-sticky-rendering.html
- /css/css-position/sticky/position-sticky-stacking-context.html
- /css/css-position/sticky/position-sticky-table-td-left.html
- /css/css-position/sticky/position-sticky-table-td-right.html
- /css/css-position/sticky/position-sticky-table-tfoot-bottom.html
- /css/css-position/sticky/position-sticky-table-th-right.html
- /css/css-position/sticky/position-sticky-table-thead-top.html
- /css/css-position/sticky/position-sticky-table-tr-bottom.html
- /css/css-position/sticky/position-sticky-table-tr-top.html
- /css/css-position/sticky/position-sticky-writing-modes.html
- /css/css-pseudo/marker-intrinsic-contribution-001.html
- /css/css-text/hyphens/hyphens-character.html

These tests now PASS due to this patch:
* FAIL -> PASS
  - /html/canvas/element/text/2d.text.draw.fill.maxWidth.fontface.html
  - /html/canvas/element/text/2d.text.measure.width.empty.html
* TIMEOUT -> PASS
  - /css/css-fonts/variations/font-descriptor-range-reversed.html
  - /css/css-fonts/variations/variable-opsz.html
  - /css/css-position/sticky/position-sticky-table-th-left.html
* ERROR -> PASS
  - /css/css-fonts/generic-family-keywords-002.html
  - /css/css-fonts/generic-family-keywords-003.html
* These two tests only PASS in Layout 2020:
  - /css/CSS2/positioning/inline-static-position-001.html
  - /css/cssom-view/getBoundingClientRect-empty-inline.html

This test has subtests that PASS intermittenttly:
- /fetch/metadata/generated/css-font-face.sub.tentative.html

This test seems to have different results on the CI than locally (could
be due to available fonts):
- /css/css-fonts/generic-family-keywords-001.html

These tests are new TIMEOUTS that used to FAIL because
`documents.fonts.ready` was a ReferenceError:
- /resource-timing/TAO-match.html
- /resource-timing/content-type.html
- /resource-timing/nextHopProtocol-is-tao-protected.https.html

The failure in `/resize-observer/change-layout-in-error.html` could be
due to an issue in the ResizeObserver implementation that is now exposed
with this change, but this needs more investigation.

---


components/script/dom/window.rs
```diff
@@ -1980,6 +1980,15 @@ impl Window {
             );
         }
 
+        let document = self.Document();
+        let pending_web_fonts = self.layout.borrow().waiting_for_web_fonts_to_load();
```
gterzian, line 1984: You want to scope the borrow so that it is dropped before `font_face_set.fulfill_ready_promise_if_needed();`, because since the promise calls into JS, it is very hard to be sure there will not be another borrow of `layout`.

components/script/script_thread.rs
```diff
@@ -3310,6 +3310,12 @@ impl ScriptThread {
         // TODO: This should only dirty nodes that are waiting for a web font to finish loading!
         document.dirty_all_nodes();
         document.window().add_pending_reflow();
+
+        // TODO: This is currently required because the handlers added to the promise exposed at
+        // `document.fonts.ready` are run by the event loop only when it performs a microtask
+        // checkpoint. Without the call below, this never happens and the promise is 'stuck'
+        // waiting to be resolved until another event forces a microtask checkpoint.
+        self.rendering_opportunity(pipeline_id);
```
gterzian, line 3318: I don't think this is correct, because the mechanism that resolves the promise is not the microtask checkpoint(which by the way in our wrong implementation is called once after each iteration of the event-loop, so it is called after handling this message as well), but rather the call to `reflow`, which does happens in the "update the rendering"(not always), but I think in your case the reflow should happen after this(since a pending reflow is added above) as part of https://github.com/servo/servo/blob/66edef806579fd0b386f4ceba473e6a9f7d0ca34/components/script/script_thread.rs#L2011. Does the promise not resolve without noting a rendering opp here?

Also, see the other comment at `reflow`.

components/script/dom/window.rs
```diff
@@ -1980,6 +1980,15 @@ impl Window {
...
+        let font_face_set = document.Fonts();
+        let is_ready_state_complete = document.ReadyState() == DocumentReadyState::Complete;
+
+        if !pending_web_fonts && is_ready_state_complete {
+            font_face_set.fulfill_ready_promise_if_needed();
```
gterzian, line 1989: I don't think this is correct, instead the promise must be fulfilled as part of the [`switch-the-fontfaceset-to-loaded`](https://drafts.csswg.org/css-font-loading/#switch-the-fontfaceset-to-loaded). However the queuing of some of the task that call into this mechanism seems to originate in methods on `FontFace`(the constructor [for example](https://drafts.csswg.org/css-font-loading/#ref-for-switch-the-fontfaceset-to-loaded)), so perhaps some kind of partial implementation of that interface should be part of this PR.

gterzian (changes requested):
Couple of initial comments...

By the way, re the resizeobserver failures in layout 2013: not sure we should run them, and also why the interface is absent when we do. Is it present in 2020 by way of the [__dir__.ini` file in meta](https://github.com/servo/servo/blob/66edef806579fd0b386f4ceba473e6a9f7d0ca34/tests/wpt/meta/resize-observer/__dir__.ini)

components/script/dom/window.rs
mukilan: Isn't the lifetime of the borrow limited to the end of the statement since the `Ref` returned is a temporary and `waiting_for_web_fonts_to_load()` returns a boolean?

components/script/script_thread.rs
mukilan: > . Does the promise not resolve without noting a rendering opp here?

Yes, that is what we observed. This example exhibits the issue if `self.rendering_opportunity()` is removed.

```html
<!DOCTYPE html>
<html>
  <head>
  <script>
    document.fonts.ready.then(function() {
      console.log("Promise resolved with", arguments);
    })
  </script>
  <style>
    @font-face {
      font-family: 'FiraSans';
      src: url('https://google.com/'); /* Force font loading to fail due to sanitizer error */
      /* src: url('https://fonts.gstatic.com/s/firasans/v17/va9E4kDNxMZdWfMOD5Vvl4jL.woff2'); */
    }
    div {
      font-family: "FiraSans";
    }
  </style>
  </head>
  <body>
    <div>Test</div>
  </body>
</html>
```

However, we don't observe the issue when `src` is a valid font, so there maybe something more going on.

components/script/dom/window.rs
mukilan: I think @mrobinson and I discussed this approach during implementation but decided not to have a new task because currently we are only interested in `document.font.ready`, so implementing step 3 of `switch-the-fontfaceset-to-loaded` (Fulfill font face set’s `[[ReadyPromise]]` attribute’s value with font face set) is necessary but not step 4.

I'm not too familiar with the working of our task queue implementation or the specification, so I'll discuss this again with Martin.

components/script/dom/window.rs
gterzian: Yes, sorry I forgot about the [promise enqueuing mechanism](https://github.com/servo/servo/blob/8121c9883477cf4350e8dd8abdc3776548c11ea3/components/script/script_runtime.rs#L196), so the borrow will have been dropped by the time the microtask runs.

components/script/script_thread.rs
gterzian: Ok, thanks for clarifying. In that case, I think your fix is accidental, in that it works but doesn't address what I think is the underlying problem:

- The promise is resolved in one of the reflow calls at the end of the event-loop iteration: https://github.com/servo/servo/blob/8121c9883477cf4350e8dd8abdc3776548c11ea3/components/script/script_thread.rs#L1995
- The microtask checkpoint is done before, at https://github.com/servo/servo/blob/8121c9883477cf4350e8dd8abdc3776548c11ea3/components/script/script_thread.rs#L1974
- You are then queuing a task, accidentally the update the rendering one, but it could have been any task really,
- The task runs, followed by a micro tack checkpoint that runs the promise callback. 

So I think the more direct way of addressing the problem, is to resolve the promise right here. Even though we are not in a "task", the micro task checkpoint will run after this. (remaining question: do you need to reflow before the promise callback runs?)

components/script/dom/window.rs
gterzian: Ok I see. See my other comment then for a more direct temporary solution.

components/script/dom/window.rs
gterzian: Actually, sorry for the confusion(unrelated to any change here since you are right about the borrow), but it appears to me that when calling `resolve_native`, the promise callback is called into then and there, not in a later microtask.

gterzian:
Ok, to summarize:

- When we call `resolve_native`, the promise handler is called immediately, not via a microtask. So the issue at hand is not microtask checkpointing, or the lack thereof. 
- I had a quick look into the mechanism behind `waiting_for_web_fonts_to_load()`, and I would look further into the concurrent logic behind it as a potential source for the bug. My guess is that there is a lack of signalling for when the underlying state changes. By queueing a task you are forcing another iteration of the event-loop, by which time the underlying value has already been updated, but this is not robust(you are relying on timing being in your favor). So whatever updates the font context, should then signal the event-loop for the promise to be resolved. 
- The bug may off-course yet linger elsewhere, but that is my best guess.

components/script/script_thread.rs
gterzian: Forget about the above, we don't need the microtask checkpoint. It may still be an idea to resolve the promise already here(and see if the problem is not in the logic around it, like the check for all web fonts having been loaded already). 

Also a problem is that it is not clear that the document is fully loaded at this point, but that is a separate problem.

components/script/dom/window.rs
gterzian: EDIT: `resolve_native` does not go through the microtask stuff...

mukilan:
> When we call resolve_native, the promise handler is called immediately, not via a microtask. So the issue at hand is not microtask checkpointing, or the lack thereof.

@gterzian by 'promise handler' do you mean the callbacks registered from the JS side using `promise.then`? If so, we observed the opposite behaviour of what you've described. Using the test html shared above in the earlier comment, adding print statements before and after `resolve_native` and we can see them being printed as expected, but the `.then` callbacks is never called until the event loop received another event for example due to mouse being moved over the document.

gterzian (changes requested):
Ok so confirmed about the microtask mechanism being used, and I think loading a web-font actually makes sense as a "rendering opportunity"(even though we probably want to move to the task queuing that the spec does later). LGTM

gterzian (approved):

mrobinson (approved):

mukilan:
@gterzian Thanks for the detailed review! We've pushed a small change to get the pending font count before we force a reflow in `Window::reflow` to handle a race condition where the reflow triggers new loads.

mukilan:
@mrobinson : The `css/css-fonts/generic-family-keywords-001.html` test is also flaky (on CI and locally), so I've updated the its expectations to be `[FAIL, PASS]`

mrobinson:
I think this is causing more flakiness here: https://github.com/servo/servo/runs/38559123024

mrobinson:
Can we revert this change or fix the flakiness issues?

PR #32587, opened 2024-06-24
mach: Expose a `--skip-static-analysis` to `mach boostrap`

### LINKED ISSUE #32582: Separate build/run time deps
When running WPT tests we spend 6min on `mach bootstrap` installing build deps and utils (crown, taplo, cargo-deny, ...) that will not be used. Instead  we should provide `--runtime-only` flag that would only install runtime deps.

mrobinson:
This should speed up runners which just need to run the WPT tests.

Fixes #32582.

---

- [x] These changes fix #32582
- [x] These changes do not require tests because this just changes CI configuration.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,7 @@ jobs:
         run: |
           python3 -m pip install --upgrade pip
           sudo apt update
-          python3 ./mach bootstrap
+          python3 ./mach bootstrap --skip-lints
       - name: Set LIBCLANG_PATH # This is needed for bindgen in mozangle.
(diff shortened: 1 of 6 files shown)


sagudev (approved):
LGTM, although I am not fan of name `static-analysis` as I consider `static-analysis` to do something with code (like crown), while taplo is only toml formatted and cargo-deny is for additional rules on Cargo.toml. Maybe `lints` or `tidy-tools` or just `tidy` would more appropriate (and shorter) naming.

mrobinson:
Okay. I'll with `--skip-lints`.

PR #32594, opened 2024-06-24
Add OpenHarmony support to servoshell

### LINKED ISSUE #30541: OpenHarmony build
@mukilan has been doing an experiment to try to build Servo on OpenHarmony OS (`aarch64-linux-ohos`).

These are the current issues and workarounds:
1. socket2 needs patches for Iovlen type. [GitHub ticket](https://github.com/rust-lang/socket2/issues/451) open but no progress.
2. ipc-channel relies on shm_* functions from libc crate. ohos doesn't have those. Adding features = ['force-inprocess'] in simpleservo/api/Cargo.toml is a workaround to force single process mode like on windows.
3. Compiling freetype-sys requires CMake cross-compilation file to be used:
```
export CMAKE_TOOLCHAIN_FILE_aarch64_linux_ohos=/home/user/dev/ohos/native-linux-x64-3.2.13.5-Release/native/build/cmake/ohos.toolchain.cmake 
``` 
4. Compiling ring crate needs cross compiler to be set correctly (not sure why ~/.cargo/config.toml didn't take effect)
```
export CC_aarch64_linux_ohos=/home/user/dev/ohos/native-linux-x64-3.2.13.5-Release/native/llvm/bin/aarch64-unknown-linux-ohos-clang.sh
```
5. patch components/allocator to disable jemalloc-sys as the native library doesn't support ohos triples
6. The `nix` crate doesn't build since it uses shm_open and other methods not availale on OHOS from libc crate.

Status: Blocked on step 6 - The [nix](https://crates.io/crates/nix) crate doesn't compile because it uses features from C library (e.g shm_open) that are currently unavailable on the ohos target of [libc crate](https://github.com/rust-lang/libc/blob/0d6554f3c2115fb67e53df82a368dac166965cff/src/unix/linux_like/linux/mod.rs#L4178). Unlike other crates below, it doesn't seem this has an easy workaround this since `nix` create is used by multiple dependencies of libservo like winit etc, but needs more investigation.

jschwe:
This PR adds basic OpenHarmony support to the servoshell. There is a large commit "Add OpenHarmony support to servoshell" which adds the OpenHarmony support, while copying a lot of code from the android version. The follow-up commits then reduce the amount of duplicated code by moving shared code into common modules.

There are still quite a lot of open tasks and improvement opportunities (for which I would open seperate issues, once this initial PR is merged), but I think this may be a good starting point.

---

- [x] These changes fix #30541
- [x] There are tests for these changes: The OpenHarmony and android builds are tested in CI.


gterzian:
I don't know much about `servoshell`, and even less about Android, so I suggest we find someone else to review. 

@mrobinson @wusyong  Do you have an idea?

jschwe:
@gterzian I think you got automatically added as a reviewer due to me touching one file in `components/script/`

wusyong:
Hmm, it looks like I've been a bit behind on the recent activity of servoshell. Couldn't the android port just adopt `WindowPortsMethods` trait? And why do we need EGL module here? Shouldn't it be handled by `surfman`?

mukilan (approved):
There are a couple of code blocks that are commented out. Do you plan to have retain them for debugging? Otherwise the changes LGTM!

ports/servoshell/egl/ohos/simpleservo.rs
```diff
@@ -0,0 +1,169 @@
...
+        .inspect_err(|e| error!("Invalid initial Servo URL `{}`. Error: {e:?}", options.url))
+        .ok()
+        .unwrap_or_else(|| ServoUrl::parse("about:blank").expect("Infallible"));
+
+    let _ = servo_glue.process_event(EmbedderEvent::NewWebView(initial_url, servo.browser_id));
+
+    Ok(servo_glue)
+}
+
+struct ResourceReaderInstance;
```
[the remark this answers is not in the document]
jschwe: Thanks for catching that! Removed and imported the shared version.

ports/servoshell/egl/ohos/simpleservo.rs
```diff
@@ -0,0 +1,169 @@
...
+    let res = unsafe {
+        OH_NativeXComponent_GetXComponentSize(
+            xcomponent,
+            native_window,
+            &mut width as *mut _,
+            &mut height as *mut _,
+        )
+    };
+    assert_eq!(res, 0, "OH_NativeXComponent_GetXComponentSize failed");
+    let width: i32 = width.try_into().expect("Width to large");
```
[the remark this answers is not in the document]
jschwe: Fixed.

components/script/dom/webglrenderingcontext.rs
```diff
@@ -216,6 +216,18 @@ pub struct WebGLRenderingContext {
 }
 
 impl WebGLRenderingContext {
+    #[cfg(target_env = "ohos")]
```
jdm, line 219: I'd prefer to have a platform check in the webgl thread code, rather than changing the DOM implementation here.

jschwe:
> Hmm, it looks like I've been a bit behind on the recent activity of servoshell. Couldn't the android port just adopt WindowPortsMethods trait? 
We recently merged the previously separate android port into the `servoshell` crate, but have not started a larger effort to unify the code base yet. 

> And why do we need EGL module here? Shouldn't it be handled by surfman?

The name of the module is not ideal. A more accurate description for the "desktop" module would be "winit-based" and the  "egl" mod would be  "not-winit" . I also considered using "mobile" instead of "egl", but OpenHarmony also supports the PC scenario, so "mobile" doesn't fit that well either. 
I think we can always change the module name later if we come up with a good naming.
It is also possible that we migrate everything to be `winit` based, then the need for two different modules would also go away in the medium to long term.

components/script/dom/webglrenderingcontext.rs
jschwe: Sounds reasonable. Please check the latest commit, which reverts this change and moves it to the webgl thread.

jschwe:
> There are a couple of code blocks that are commented out. Do you plan to have retain them for debugging? Otherwise the changes LGTM!

I checked the commented code block (was there more than one?) about the window offset, and removed it, since it turned out to be unnecessary.

ports/servoshell/egl/ohos.rs
```diff
@@ -0,0 +1,626 @@
...
+                    location.file(),
+                    location.line()
+                );
+            } else {
+                let _ = error!("{} (thread {})", msg, name);
+            }
+
+            let _ = crate::backtrace::print_ohos();
+
+            // if opts::get().hard_fail && !opts::get().multiprocess {
```
mukilan, line 359: This is the other comment block I was wondering about.  It could have a `TODO:` comment attached if we plan to enable it later.

jschwe: I've removed the commented code for now. I think this panic handler is anyway a likely candidate to be merged with the corresponding desktop implementation, so we can revisit the hard exit option at that point.

mukilan (approved):

PR #32600, opened 2024-06-24
Clean up some unused dependencies

Loirooriol:
---


python/tidy/tests/Cargo.toml
```diff
@@ -4,8 +4,5 @@ version = "0.0.1"
 authors = ["The Servo Project Developers"]
 publish = false
 
-[dependencies]
-test-package = { version = "*" }
-
```
mrobinson, line 9: I guess this one might be a fake dependency for a unit test expectation.

mrobinson (changes requested):
Thanks for the clean up!

python/tidy/tests/Cargo.toml
Loirooriol: Yes I checked unit tests but not `./mach test-scripts`

mrobinson (approved):

PR #32602, opened 2024-06-25
Add ReadableStream::Constructor and its related types

wusyong:
This PR adds `ReadableStream`'s `Constructor` and all it's related types. This one is fairly large because it needs to complete the internal slots of both `ReadableStreamDefaultController` and `ReadableByteStreamController`. So I think best way to review is to verify these two functions and see if they set the internal slots correctly:
- `setup_readable_byte_stream_controller_from_underlying_source`
- `setup_readable_stream_default_controller_from_underlying_source`

The challenges I encountered including how to define controller's algorithm and how to handle the promise when we want to call these algorithm. I'll add comment to the code to indicate later

---

- [x] These changes do not require tests because SM stream is still enabled. This PR aims to add the implementation incrementatlly.

### DIFF
--- a/components/script/dom/readablebytestreamcontroller.rs
+++ b/components/script/dom/readablebytestreamcontroller.rs
@@ -2,19 +2,141 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use dom_struct::dom_struct;
-use js::rust::HandleValue as SafeHandleValue;
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::rc::Rc;
 
+use dom_struct::dom_struct;
+use js::gc::MutableHandleValue;
+use js::jsapi::{HandleValue, Heap, JSObject};
+use js::jsval::{JSVal, UndefinedValue};
+use js::rust::{HandleObject as SafeHandleObject, HandleValue as SafeHandleValue};
+
+use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::ReadableByteStreamControllerBinding::ReadableByteStreamControllerMethods;
-use crate::dom::bindings::import::module::{Error, Fallible};
-use crate::dom::bindings::reflector::Reflector;
-use crate::dom::bindings::root::DomRoot;
+use crate::dom::bindings::codegen::Bindings::UnderlyingSourceBinding::{
+    ReadableStreamController, UnderlyingSource,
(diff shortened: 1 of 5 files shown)


components/script/dom/readablebytestreamcontroller.rs
```diff
@@ -54,3 +176,327 @@ impl ReadableByteStreamControllerMethods for ReadableByteStreamController {
...
+                .is_ok());
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Algorithms for [setup_readable_stream_default_controller_from_underlying_source]
+#[derive(JSTraceable, MallocSizeOf)]
+pub enum ControllerAlgorithms {
```
wusyong, line 370: The controllers have several internal slots that are "Algorithms". I found it difficult to use generic in the controller's field, because they can be changed later. Perhaps it's better to use `dyn triat`?

components/script/dom/readablebytestreamcontroller.rs
```diff
@@ -54,3 +176,327 @@ impl ReadableByteStreamControllerMethods for ReadableByteStreamController {
...
+        ReadableStreamController::ReadableByteStreamController(controller.clone()),
+        start_result.handle_mut(),
+    )?;
+    // Step 15
+    let global = &*controller.stream.global();
+    let realm = enter_realm(&*global);
+    let comp = InRealm::Entered(&realm);
+    let start_promise = Promise::new_resolved(global, cx, start_result.handle())?;
+    // Step 16 & 17
+    start_promise.append_native_handler(
```
wusyong, line 265: Resolving a promise in the controller seems a bit complicated. Is it correct to append the handler after `new_resolved`?

components/script/dom/underlyingsource.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        };
+        Ok(rval_decl)
+    }
+}
+
+impl UnderlyingSourceStartCallback {
+    // FIXME Is this a good way to do? What are other way to call with existing js object?
+    #[allow(unsafe_code)]
+    pub fn call_with_existing_obj(
```
wusyong, line 90: This feels like a terrible way to call with existing object. The problem is that `UnderlyingSource` is a type generated by codegen without a reflector. I'm unable to get its `DomObject` and simply call `Call_`.

I think the reason is probably this is our first time encountering a dictionary with callbacks.

wusyong:
Closed it because of the discussion in https://github.com/servo/servo/issues/29088#issuecomment-2188639148

PR #32619, opened 2024-06-26
Layout: Implement CSS Grid using `taffy`

nicoburns:
## Objective

Implement CSS Grid in `layout_2020`

## Changes made

- Add dependency on the `taffy` crate
- Vendor `stylo_taffy` conversion code
- Add a new taffy-based layout mode
- Add new `Grid` box types which use the taffy-based layout mode

## WPT test results

**Overall percentage of `css-grid` passing: 38.3%** (compared to 18.6% on main)
Note: all previous test passes were either coincidental or style only

## Potential future work

- Improve the correctness (WPT score) of the implementation
- Optimise implementation to make better use of cached inline sizes
- Add a taffy-based Flexbox implementation

---

- [x] There are tests for these changes (all of WPT's `css-flexbox`)


nicoburns:
Local `css-flexbox` WPT results with the latest commit "More content_box <-> border_box conversion corrections": 

> Ran 1233 tests finished in 105.5 seconds.
  • 983 ran as expected. 0 tests skipped.
  • 119 tests failed unexpectedly
  • 1 tests timed out unexpectedly
  • 74 tests passed unexpectedly
  • 56 tests had unexpected subtest results

There's still something going very wrong as the servo.org header is completely messed up. But the bottom of the page is now rendering almost perfectly. And Google (both home page and results page are looking pretty good (although not perfect)).

nicoburns:
Fixed a typo that was using the value for `flex_grow` in place of `flex_shrink`:

> Ran 1233 tests finished in 100.7 seconds.
  • 1024 ran as expected. 0 tests skipped.
  • 75 tests failed unexpectedly
  • 1 tests timed out unexpectedly
  • 79 tests passed unexpectedly
  • 54 tests had unexpected subtest results

KnorpelSenf:
This is very cool—are there actually any plans to move to taffy, though? Is this just a proof of concept? What are the advantages, what are the disadvantages, especially w.r.t. web compat and perf?

nicoburns:
> This is very cool—are there actually any plans to move to taffy, though? Is this just a proof of concept? What are the advantages, what are the disadvantages, especially w.r.t. web compat and perf?

No such decision has been made yet. In fact there was some skepticism that such a solution will be performant and maintainable. This PR essentially serves as a proof of concept that will allow the TSC to make a decision with concrete data around things like performance and web compatibility (i.e WPT test scores).

The motivation is:
1. To go from "pretty bad flexbox support" to "pretty good flexbox and css grid support" much more quickly than we could by implementing these specs in servo directly.
2. To be able to share the implementation of Flexbox and CSS Grid (and potentially other layout modes in future) with the wider Rust UI ecosystem on an ongoing bases (and the benefits of concentrated development resources that brings).

Currently the performance is pretty poor, but that's because the full Flexbox algorithm is inherently exponential time complexity, and the current Taffy integration implements that but does not yet implement the caching layer necessary to make this reasonably performant. However, this is not something that is Taffy-specific and is something that any correct Flexbox implementation in Servo will need to contend with. The fact that this PR is surfacing that is good thing as far as I am concerned as it pushes Servo towards a layout architecture that can correctly implement all of css layout.

nicoburns:
I've squashed the history of this PR down into sensible chunks and split https://github.com/servo/servo/pull/32703 out into it's own mini PR that's hopefully uncontroversial enough to get merged quickly.

nicoburns:
Stable unexpected FAILS:

**Taffy doesn't implement safe/unsafe alignment:**
FAIL [expected PASS] /css/css-align/self-alignment/self-align-safe-unsafe-flex-003.html

**Taffy bug around resolving percentages against indefinite cross-axis heights:**
FAIL [expected PASS] 1 subtest of /css/css-flexbox/percentage-heights-001.html
FAIL [expected PASS] 3 subtests of /css/css-flexbox/percentage-heights-003.html

**Taffy doesn't implement `flex-basis: content`:**
FAIL [expected PASS] /css/css-flexbox/flexbox-flex-basis-content-001a.html
FAIL [expected PASS] /css/css-flexbox/flexbox-flex-basis-content-001b.html
FAIL [expected PASS] /css/css-flexbox/flexbox-flex-basis-content-004a.html
FAIL [expected PASS] /css/css-flexbox/flexbox-flex-basis-content-004b.html

**Taffy doesn't implement `flex-basis: fit-content`:**
FAIL [expected PASS] 1 subtest of /css/css-flexbox/flex-basis-intrinsics-001.html

**Fails to render ref (inline flex items are not blockified):**
FAIL [expected PASS] /css/css-flexbox/flexbox-gap-position-absolute.html

**Unclassified:**
FAIL [expected PASS] /css/css-flexbox/align-items-baseline-overflow-non-visible.html
FAIL [expected PASS] 3 subtests of /css/css-flexbox/alignment/flex-align-baseline-006.html
FAIL [expected PASS] /css/css-flexbox/auto-margins-001.html
FAIL [expected PASS] 1 subtest of /css/css-flexbox/fieldset-as-container-justify-center.tentative.html
FAIL [expected PASS] /css/css-flexbox/flex-aspect-ratio-img-column-012.html
FAIL [expected PASS] /css/css-flexbox/flex-aspect-ratio-img-row-007.html
FAIL [expected PASS] 4 subtests of /css/css-flexbox/flex-factor-less-than-one.html
FAIL [expected PASS] 2 subtests of /css/css-flexbox/flex-flow-013.html
FAIL [expected PASS] 3 subtests of /css/css-flexbox/flex-item-compressible-001.html
FAIL [expected PASS] /css/css-flexbox/flex-minimum-height-flex-items-014.html
FAIL [expected PASS] /css/css-flexbox/flex-minimum-width-flex-items-012.html
FAIL [expected PASS] /css/css-flexbox/flex-wrap-004.html
FAIL [expected PASS] /css/css-flexbox/flexbox-min-width-auto-001.html
FAIL [expected PASS] 2 subtests of /css/css-flexbox/multiline-min-max.html
FAIL [expected PASS] 1 subtest of /css/css-flexbox/orthogonal-writing-modes-and-intrinsic-sizing.html

Quackdoc:
Really nice playing with this, aside from performance it's been quite nice to see many websites become "usable", one site I use as a reference is https://rustytube.rs which actually can play youtube videos (invidious client). 

When using this PR, the main window will render somewhat correctly using the below command, at the very least, entry and videos are in the proper layout, so it's a great improvement even with it being slow.

`./target/production/servo --pref layout.flexbox.enabled --pref layout.columns.enabled --pref media.testing.enabled --pref layout.grid.enabled 'https://rustytube.rs'` 

however when using the below to load a video page directly, it will hang on a white screen regardless of how long it is left to load which is for sure a regressions since before it was able to not only load the page, but also play a video

`./target/production/servo --pref layout.flexbox.enabled --pref layout.columns.enabled  --pref media.testing.enabled  --pref layout.grid.enabled 'https://rustytube.rs/player?id=2oaBT0fDmW0'`

sagudev:
IIUC taffy also depends on stylo, so this might make stylo bumps even harder, right?

nicoburns:
> IIUC taffy also depends on stylo, so this might make stylo bumps even harder, right?

Taffy itself does not depend on Stylo. It's actually pretty much entirely dependency free so long as you're using `std` (we have a couple of optional deps for floating point ops on `no_std`) and you're not using the built-in node tree which depends on `slotmap` (servo wouldn't). The one other dependency is [grid](https://lib.rs/crates/grid) which is a small wrapper around `Vec` for 2D indexing/iteration. We'll probably eliminate that at some point (and just use `Vec`) but in any case it's tiny.

There is also a `taffy_stylo` integration crate in which is used in this PR. That depends on both `taffy` and `stylo` and does type conversions between them. That crate is currently living in the `taffy` repo, but for the reason you mention (easier stylo upgrades) it would probably be better if it lived either in the `stylo` repo or directly in the `servo` repo. Most of the conversions are quite trivial (see https://github.com/DioxusLabs/taffy/pull/693/files#diff-284087ee0706261aa3994b965b2702e310c857241411ba1f57a95adbba1b299eR120 for example), and Taffy has a much more slowly changing API than Stylo (and has semver versioned releases) so I think if the interop code was owned on the stylo/servo side it would be quite easy to keep up to date.

sagudev:
Ah, I was fooled by `taffy_stylo`. Thanks for explanation.

nicoburns:
@mrobinson I believe this is now ready for review:

- The crash when laying out abspos nodes in parallel has been fixed (and the workaround removed)
- And I have removed the `Clone` (and `Debug`) impl's from the Fragment tree. The current solution for this is simply to not attempt to reuse fragments and recompute them each time.
- I've copied the Taffy <-> Stylo conversion code directly into the Servo tree. This code also exists in the taffy repository as it's own crate, but I think having it here probably simplifies Stylo upgrades.
- Performance issues have not been resolved, but given that this is now behind an off-by-default feature flag I am assuming that this is acceptable and something that can be fixed in follow up PRs. I am also anticipating that the main flexbox implementation is going to run into these exact same issues once columns are implemented. And it will probably make more sense for the Taffy implementation to able to share whatever solution is devised in the main implementation rather than doing it's own thing.

Overall I'm pretty pleased with how this has turned out. The integration is much more lightweight than having an entirely separate crate (like we do for the legacy layout), and being able to run Servo with both default layout and taffy layout using the same binary is very convenient for running comparisons.

tests/wpt/meta/css/css-flexbox/__dir__.ini
```diff
@@ -1 +1 @@
-prefs: ["layout.columns.enabled:true", "layout.flexbox.enabled:true"]
+prefs: ["layout.columns.enabled:true", "layout.flexbox.enabled:true"]
```
mrobinson, line 1: It looks like this change is accidental?

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,600 @@
...
+    // container_min_cross_size: Length,
+    // container_max_cross_size: Option<Length>,
+    // flex_axis: FlexAxis,
+    // flex_direction_is_reversed: bool,
+    // flex_wrap_reverse: bool,
+    // main_start_cross_start_sides_are: MainStartCrossStart,
+    // container_definite_inner_size: FlexRelativeVec2<Option<Length>>,
+    // align_content: AlignContent,
+    // align_items: AlignItems,
+    // justify_content: JustifyContent,
```
mrobinson, line 77: Please remove commented out code.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,600 @@
...
+        let mut_ref: AtomicRefMut<'_, _> = (*self.source_child_nodes[id]).borrow_mut();
+        AtomicRefMut::map(mut_ref, |node| &mut node.taffy_layout_cache)
+    }
+
+    fn compute_child_layout(
+        &mut self,
+        node_id: taffy::NodeId,
+        inputs: taffy::LayoutInput,
+    ) -> taffy::LayoutOutput {
+        // compute_cached_layout(self, node_id, inputs, |parent, node_id, inputs| {
```
mrobinson, line 128: Ditto.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,600 @@
...
+                                &self.layout_context,
+                                &mut child_positioning_context,
+                                &content_box_size_override,
+                                &containing_block,
+                            );
+
+                            // Store layout data on child for later access
+                            child.positioning_context = child_positioning_context;
+
+                            // self.positioning_context.append(child_positioning_context);
```
mrobinson, line 250: Looks like this should be removed as well.

components/layout_2020/formatting_contexts.rs
```diff
@@ -54,6 +56,8 @@ pub(crate) struct ReplacedFormattingContext {
 pub(crate) enum NonReplacedFormattingContextContents {
     Flow(BlockFormattingContext),
     Flex(FlexContainer),
+    TaffyFlex(TaffyContainer),
```
mrobinson, line 59: Instead of adding a new formatting context type here, please simply implement a new "taffy_layout" method for `NonReplacedFormattingContextContents::Flex` that's called conditionally.

components/layout_2020/formatting_contexts.rs
```diff
@@ -230,6 +252,18 @@ impl NonReplacedFormattingContext {
...
+                positioning_context,
+                containing_block_for_children,
+                containing_block,
+            ),
+            NonReplacedFormattingContextContents::Grid(fc) => fc.layout(
+                layout_context,
+                positioning_context,
+                containing_block_for_children,
+                containing_block,
+            ),
```
mrobinson, line 266:
```suggestion
            NonReplacedFormattingContextContents::Grid(formatting_context) => formatting_context.layout(
                layout_context,
                positioning_context,
                containing_block_for_children,
                containing_block,
            ),
```

python/wpt/__init__.py
```diff
@@ -33,6 +33,8 @@ def create_parser():
                         help="Pass preferences to servo")
     parser.add_argument('--legacy-layout', '--layout-2013', '--with-layout-2013', default=False,
                         action="store_true", help="Use expected results for the legacy layout engine")
+    parser.add_argument('--taffy-layout', default=False,
+                        action="store_true", help="Use expected results for the taffy layout engine")
```
mrobinson, line 37: Instead of a command-line argument, please just use the preference. You can make it so that the preference automatically turns on grid layout and flex layout so only one needs to be passed to the command.

mrobinson (changes requested):
Just some basic comments. I have not had a chance to dig into the layout code, but I hope to do that soon. Thanks for your patience!

components/layout_2020/formatting_contexts.rs
nicoburns: That isn't going to be very easy to do I'm afraid, as the `TaffyItemBox` type has extra fields compared to the main Flexbox implementation's item representation. It would however be trivial to combine the `TaffyFlex` and `Grid` variants if that was considered desirable (you'll notice they use the same inner type already). I could also combine the contexts by adding the new fields to the main flexbox impl's context (perhaps wrapped in `Option<Box<>>`) if you want?

A third option would be to keep separate contexts but abstract the item collection logic so that it is shared. My intuition on this is that it might make sense to do in future, but I am a little wary of entangling the implementations at this point.

FYI, I suspect if/when Servo gains a non-Taffy based Grid impl that will also be able to share a context with Flexbox, as both layout modes' treatment of items is identical.

python/wpt/__init__.py
nicoburns: Hmm... I *can* do this. But I'm reluctant too as the CLI argument seems both cleaner from a code maintenance perspective and much easier to use from a user perspective (the commands for running WPT are already really long if you need to specify multiple sub-suites to run, a report output location, etc). This also seems like it would then be different to how the legacy layout engine is specified (`--legacy-layout`), so perhaps that ought to be changed to use just a pref too if I change it for taffy layout?

tests/wpt/meta/css/css-flexbox/__dir__.ini
nicoburns: Think this one is probably a rebase conflict. Will undo. What I would like, to do soon (once this PR lands) is enable the `css-grid` tests, if that is something you'd be open to.

nicoburns:
@mrobinson Are you still planning to review this? Looks like the main Flexbox implementation has pretty much caught up with this one now, but this could still be useful for CSS Grid (and I guess as an alternate Flexbox implementation seeing as we pretty much get that for free if have Taffy integration for CSS Grid).

Looks like this has conflicts due to changes in the containing block parameters, etc. But I think those will actually simplify this PR slightly, as the `IndefiniteContainingBlock` matches Taffy's semantics more closely.

mrobinson:
@nico Please accept my apologies for not reviewing this. I will do my best to review. I think there can be two changes here:

1. Remove the flexbox and only keep the grid bits as I think with the support for RTL layout flexbox in Servo is already equal or a bit more advanced than taffy.
2. Please remove the command-line argument. All other layout knobs are exposed as preferences.

Erithax:
Personally, I'd like to see Taffy flexbox stay. Taffy is a modular library used throughout the Rust ecosystem, which makes it more attractive to contribute to. I believe this modularity also aligns with the goals of the Servo project.

components/layout_2020/formatting_contexts.rs
nicoburns: This looks outdated to me. The existing code now matches the surrounding code (which I suspect it didn't when this suggestion was made). But let me know if you still want this change.

python/wpt/__init__.py
nicoburns: done!

nicoburns:
@mrobinson I believe this is now ready for re-review. I have rebased this on top of latest main, and addressed all of the review comments I could find.

Some notes:

- This still has a vendored copy of the Stylo -> Taffy type conversion code (`taffy_stylo.rs`). However, I now think it would be better if this code could live in the Stylo repo. That would allow deduplication between the copy included here, the copy in the Blitz repo, and the copy in the Taffy repo (see https://github.com/servo/stylo/pull/88).
- I have removed the Flexbox support, but I would still be keen on bringing that back in future. I think there is value in having competing implementations, and if we make the stylo->taffy type conversion crate it's own crate as above then it would likely only be ~50 LOC to enable this.
- Most of the code in `taffy/construct.rs` is duplicated from `flexbox/construct.rs`. There would be scope to deduplicating this into generic code in future if we are willing to have the `taffy` module and the `flexbox` module share code.
- I'm currently getting ~40% WPT pass rate (for `css-grid` tests) with this PR. Which doesn't sound great, but seems to be able to render a decent proportion of grids correctly in practice. This should improve over time as I execute on Taffy's [Roadmap](https://github.com/DioxusLabs/taffy/issues/345). In particular, I think missing `direction` and `writing-mode` support are causing a lot of failures.

nicoburns:
Rebased on top of https://github.com/servo/servo/pull/34090

@mrobinson Are there any outstanding issues with this PR? Maintaining this as conflicting changes land is quite time consuming, and I'm keen to get it merged so that I can move on to other tasks.

components/layout_2020/Cargo.toml
```diff
@@ -49,6 +49,8 @@ servo_geometry = { path = "../geometry" }
 servo_url = { path = "../url" }
 style = { workspace = true }
 style_traits = { workspace = true }
+taffy = { workspace = true }
+stylo_taffy = { workspace = true }
```
mrobinson, line 53: Please integrate these conversions directly into the `layout_2020` crate instead making them a separate crate.

components/layout_2020/taffy/construct.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Or pseudo-element
+    Element {
+        info: NodeAndStyleInfo<Node>,
+        display: DisplayGeneratingBox,
+        contents: Contents,
+        box_slot: BoxSlot<'dom>,
+    },
+    TextRuns(Vec<FlexTextRun<'dom, Node>>),
+}
+
```
mrobinson, line 67: This change is implementing support for grid, but this is called `FlexLevelJob`. Perhaps a better name would be `GridLevelJob`?

components/layout_2020/taffy/construct.rs
```diff
@@ -0,0 +1,240 @@
...
+    text_decoration_line: TextDecorationLine,
+    contiguous_text_runs: Vec<FlexTextRun<'dom, Node>>,
+    /// To be run in parallel with rayon in `finish`
+    jobs: Vec<FlexLevelJob<'dom, Node>>,
+    has_text_runs: bool,
+}
+
+enum FlexLevelJob<'dom, Node> {
+    /// Or pseudo-element
+    Element {
```
mrobinson, line 59: I know that this may originate from our existing code, but let's clean this up since we are touching it. I think in this case a more complete name is a better approach than the doc comment that is in conversation with the enum variant name.

```suggestion
    ElementOrPseudoElement {
```

components/layout_2020/taffy/construct.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn handle_element(
+        &mut self,
+        info: &NodeAndStyleInfo<Node>,
+        display: DisplayGeneratingBox,
+        contents: Contents,
+        box_slot: BoxSlot<'dom>,
+    ) {
+        // FIXME: are text runs considered "contiguous" if they are only separated
+        // by an out-of-flow abspos element?
+        // (That is, are they wrapped in the same anonymous flex item, or each its own?)
```
mrobinson, line 94: This comment isn't applicable any longer as it refers to flex items.

components/layout_2020/taffy/construct.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// <https://drafts.csswg.org/css-text/#white-space>
+fn is_only_document_white_space<Node>(run: &FlexTextRun<'_, Node>) -> bool {
+    // FIXME: is this the right definition? See
+    // https://github.com/w3c/csswg-drafts/issues/5146
+    // https://github.com/w3c/csswg-drafts/issues/5147
+    run.text
+        .bytes()
+        .all(|byte| matches!(byte, b' ' | b'\n' | b'\t'))
+}
```
mrobinson, line 114: I would make the version of this from the flex code public and reuse it rather than copy-and-pasting here.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+use crate::{ContainingBlock, IndefiniteContainingBlock};
+
+const DUMMY_NODE_ID: taffy::NodeId = taffy::NodeId::new(u64::MAX);
+
+fn resolve_content_size(constraint: taffy::AvailableSpace, content_sizes: ContentSizes) -> f32 {
+    match constraint {
+        taffy::AvailableSpace::Definite(limit) => {
+            let min = content_sizes.min_content.to_f32_px();
+            let max = content_sizes.max_content.to_f32_px();
+            limit.min(max).max(min)
```
mrobinson, line 38: Could this use [`clamp`](https://doc.rust-lang.org/std/primitive.f32.html#method.clamp) instead of a `min()` and then `max()` call?

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+use crate::positioned::{AbsolutelyPositionedBox, PositioningContext, PositioningContextLength};
+use crate::sizing::{ContentSizes, InlineContentSizesResult};
+use crate::style_ext::ComputedValuesExt;
+use crate::{ContainingBlock, IndefiniteContainingBlock};
+
+const DUMMY_NODE_ID: taffy::NodeId = taffy::NodeId::new(u64::MAX);
+
+fn resolve_content_size(constraint: taffy::AvailableSpace, content_sizes: ContentSizes) -> f32 {
+    match constraint {
+        taffy::AvailableSpace::Definite(limit) => {
```
mrobinson, line 35: I would import `taffy::AvailableSpace` here unless it conflicts with some other definition.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+
+/// Layout parameters and intermediate results about a flex container,
+/// grouped to avoid passing around many parameters
+struct FlexContext<'a> {
+    source_child_nodes: &'a [ArcRefCell<TaffyItemBox>],
+    layout_context: &'a LayoutContext<'a>,
+    positioning_context: &'a mut PositioningContext,
+    content_box_size_override: &'a ContainingBlock<'a>,
+    style: &'a ComputedValues,
+}
```
mrobinson, line 67:
```suggestion
/// Layout parameters and intermediate results about a grid container,
/// grouped to avoid passing around many parameters
struct GridContext<'a> {
    source_child_nodes: &'a [ArcRefCell<TaffyItemBox>],
    layout_context: &'a LayoutContext<'a>,
    positioning_context: &'a mut PositioningContext,
    content_box_size_override: &'a ContainingBlock<'a>,
    style: &'a ComputedValues,
}
```

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+        }
+
+        fn option_f32_to_size(input: Option<f32>) -> Size<Au> {
+            match input {
+                None => Size::Initial,
+                Some(length) => Size::Numeric(Au::from_f32_px(length)),
+            }
+        }
+
+        with_independant_formatting_context(
```
mrobinson, line 134: I suspect the sizing code here might need an update. @Loirooriol do you mind taking a look here as you are the expert on sizing ATM? :)

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+
+            // TODO: determine this accurately
+            //
+            // "true" is a safe default as it will prevent Servo from performing optimizations based
+            // on the assumption that the node's size does not depend on block constraints.
+            depends_on_block_constraints: true,
+        }
+    }
+
+    /// <https://drafts.csswg.org/css-flexbox/#layout-algorithm>
```
mrobinson, line 392: This link should be updated to refer to the grid specification.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+                }
+
+                fn rect_to_physical_sides<T>(rect: taffy::Rect<T>) -> PhysicalSides<T> {
+                    PhysicalSides {
+                        left: rect.left,
+                        right: rect.right,
+                        top: rect.top,
+                        bottom: rect.bottom,
+                        _unit: PhantomData,
+                    }
```
mrobinson, line 475: Please use `PhysicalSides::new()`

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,620 @@
...
+                            y: position.y,
+                            _unit: PhantomData,
+                        },
+                        size: PhysicalSize {
+                            width: size.width,
+                            height: size.height,
+                            _unit: PhantomData,
+                        },
+                    }
+                }
```
mrobinson, line 494: Please use the constructors `PhysicalRect::new()`, `PhysicalPoint::new()` and `PhysicalSize::new()`.

components/shared/stylo_taffy/src/convert.rs
```diff
@@ -0,0 +1,566 @@
+// Copyright 2013 The Servo Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
```
mrobinson, line 9: The year is wrong here.

This is your code, so you have the ability to choose a license for it and also relicense it. When contributing this to Servo, you should use the license the project uses (MPL). If you make this a separate crate you can always relicense it.

components/layout_2020/formatting_contexts.rs
```diff
@@ -125,6 +127,14 @@ impl IndependentFormattingContext {
...
                         )
                     },
+                    DisplayInside::Grid => {
+                        NonReplacedFormattingContextContents::Grid(TaffyContainer::construct(
+                            context,
+                            node_and_style_info,
+                            non_replaced_contents,
+                            propagated_text_decoration_line,
+                        ))
+                    },
```
mrobinson, line 137: Shouldn't this be dependent on the new preference?

tests/wpt/meta/css/css-grid/__dir__.ini
```diff
@@ -0,0 +1 @@
+prefs: ["layout.columns.enabled:true", "layout.flexbox.enabled:true", "layout.grid.enabled:true"]
```
mrobinson, line 1: You have added the preference, but are you using it somewhere as well? Maybe I just missed it above.

mrobinson (changes requested):

components/layout_2020/taffy/layout.rs
nicoburns: It can't because `clamp` panics if `min > max` (see docs you linked above), but that's valid in CSS (min takes precedence over max).

components/layout_2020/taffy/layout.rs
nicoburns: I've gone with `TaffyContainerContext` which matches the "taffy" naming convention of the rest of module, and reflects that almost none of the code is grid-specific (only the implementation of `taffy::LayoutGridContainer` and the calls to `taffy::compute_grid_layout`)

components/layout_2020/formatting_contexts.rs
nicoburns: This codepath can't be reached with the preference disabled anyway, because Stylo will also have the grid preference disabled which means that it won't parse `display: grid`, and you'll get whatever the fallback `display` value is. I can put in an `unreachable!()` fallback branch if you want?

tests/wpt/meta/css/css-grid/__dir__.ini
nicoburns: It sets a corresponding preference in Stylo. `display: grid` fails to parse if the Stylo pref isn't set, so none of the other code gets hit.

components/layout_2020/taffy/layout.rs
nicoburns: I'm pretty sure some of the sizing code does need updating. It would be good to fix it, but IMO it shouldn't be blocking, because it ultimately just becomes one of many layout bugs, of which there are many (but at least we get a baseline level of support).

components/layout_2020/taffy/construct.rs
nicoburns: Ok, so I've tested this, and the answer in "no" (flex runs are not considered "contiguous" if they are only separated by an out-of-flow abspos element). This is also the behaviour that Servo already implements, so I'm just going to remove this comment entirely.

components/layout_2020/taffy/construct.rs
nicoburns: Ok, so I've made this change.

However, I've also (in a subsequent commit) gone ahead and extracted the part of the construction code which is shared between the flexbox and grid code (which is almost all of it) into its own file/abstraction (`layout_2020/modern_construct.rs`). This leaves only about 15 lines of code (excluding `use` statements)  in `flexbox/construct.rs` and `taffy/construct.rs`, so I've moved that code into the `mod.rs` files and remove the `construct.rs` modules.

This is in a separate commit, so we can roll this back if you don't want to land it as part of this PR. But IMO it's a nice improvement to not have this big chunk of code duplicated. And as far as I'm aware Flexbox and Grid are specced to handle children identically (in particular the wrapping of contiguous runs of text in an anonymous block node, which is the main thing this code does), so I would not anticipate this code needing to diverge in future.

components/layout_2020/taffy/layout.rs
Loirooriol: The min-content shouldn't be greater than the max-content, but yeah I wouldn't rely on that due to bugs.
Maybe shorter as `content_sizes.shrink_to_fit(Au::from_f32_px(limit)).to_f32_px()`, but either is fine.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,607 @@
...
+                Some(length) => Size::Numeric(Au::from_f32_px(length)),
+            }
+        }
+
+        with_independant_formatting_context(
+            &mut child.taffy_level_box,
+            |independent_context| -> taffy::LayoutOutput {
+                match independent_context {
+                    IndependentFormattingContext::Replaced(replaced) => {
+                        // The containing block of a grid item is the content box of the grid container
```
Loirooriol, line 137: The containing block is actually the grid area, see https://drafts.csswg.org/css-grid/#grid-item-sizing

components/layout_2020/taffy/layout.rs
Loirooriol: I don't see blatant issues at first glance, if needed we can refine later indeed.

components/layout_2020/taffy/layout.rs
```diff
@@ -0,0 +1,607 @@
...
+                        };
+                        let size = inputs.known_dimensions.unwrap_or(computed_size);
+                        taffy::LayoutOutput {
+                            size,
+                            ..taffy::LayoutOutput::DEFAULT
+                        }
+                    },
+
+                    IndependentFormattingContext::NonReplaced(non_replaced) => {
+                        // The containing block of a grid item is the content box of the grid container
```
Loirooriol, line 199: Ditto

components/layout_2020/taffy/layout.rs
Loirooriol: Still not removed

components/layout_2020/construct_modern.rs
```diff
@@ -176,62 +175,58 @@ where
...
-                        independent_formatting_context: IndependentFormattingContext::NonReplaced(
-                            non_replaced,
-                        ),
-                        block_content_size_cache: Default::default(),
-                    })))
+                    let formatting_context =
+                        IndependentFormattingContext::NonReplaced(non_replaced);
+
+                    Some(ModernItem {
+                        kind: ModernItemKind::OutOfFlow,
```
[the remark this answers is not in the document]
nicoburns: This one was a pure typo. Well caught!

components/layout_2020/construct_modern.rs
```diff
@@ -176,62 +175,58 @@ where
...
+                            TextDecorationLine::NONE
+                        } else {
+                            self.text_decoration_line
+                        },
+                    );
+
+                    if is_abspos {
+                        Some(ModernItem {
+                            kind: ModernItemKind::OutOfFlow,
+                            order: 0,
```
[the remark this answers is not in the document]
nicoburns: This was copied over from the existing flexbox code. I haven't looked into it too deeply, but that code quotes https://www.w3.org/TR/css-flexbox-1/#order-property:

> Absolutely-positioned children of a [flex container](https://www.w3.org/TR/css-flexbox-1/#flex-container) are treated as having [order: 0](https://www.w3.org/TR/css-flexbox-1/#propdef-order) for the purpose of determining their painting order relative to [flex items](https://www.w3.org/TR/css-flexbox-1/#flex-item).

It makes sense to me that `order` would only affect paint order for abspos items, as they are laid out individually.

components/layout_2020/taffy/layout.rs
nicoburns: I believe this was true of `layout_2020`'s use of variables called "containing_block" (which seemed to mostly not actually be a containing block as the spec defines it) when this comment was written. But it looks like a lot of that has been cleaned up now. It's on my list to go through and re-evaluate the sizing constraint conversion logic with the latest layout_2020 changes.

components/layout_2020/construct_modern.rs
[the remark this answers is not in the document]
Loirooriol: Oh, it affected abspos in the past: https://www.w3.org/TR/2016/CR-css-flexbox-1-20160526/#order-property

components/layout_2020/taffy/layout.rs
Loirooriol: OK the provided `containing_block` may be wrong. But the comment seemed to refer to the containing block as defined in the spec.

Loirooriol:
You should update the expectations. Also I'm not familiar with license things so I will defer to Martin about MIT vs Apache vs MPL.

nicoburns:
@mrobinson I've inlined (and MPL licensed) the stylo/taffy type conversion code as we discussed. It was tricky to keep the feature flags for flexbox/block conversion code in the new strucutre (they'd have needed to become feature flags on the `layout_2020` crate). So I've removed that code entirely.

I have also consolidated the versions this code that I was using elsewhere, and a full version of this code (triple licensed as MIT/Apache 2.0/MPL 2.0) now lives in the Blitz repository (https://github.com/DioxusLabs/blitz/tree/main/packages/stylo_taffy) and is being used by Blitz for it's stylo/taffy interop.

components/layout_2020/flexbox/mod.rs
```diff
@@ -127,6 +162,13 @@ impl std::fmt::Debug for FlexItemBox {
 }
 
 impl FlexItemBox {
+    pub(crate) fn new(independent_formatting_context: IndependentFormattingContext) -> Self {
```
mrobinson, line 165: I don't think this has to be public.

components/layout_2020/construct_modern.rs
```diff
@@ -2,80 +2,77 @@
...
-        contents.traverse(context, info, &mut builder);
-        builder.finish()
-    }
+/// <https://drafts.csswg.org/css-text/#white-space>
+fn is_only_document_white_space(text: &str) -> bool {
+    // FIXME: is this the right definition? See
+    // https://github.com/w3c/csswg-drafts/issues/5146
+    // https://github.com/w3c/csswg-drafts/issues/5147
+    text.bytes()
+        .all(|byte| matches!(byte, b' ' | b'\n' | b'\t'))
```
mrobinson, line 29: Let's make this a function on `FlexItemTextRun`.

components/layout_2020/construct_modern.rs
```diff
@@ -103,31 +97,36 @@ where
...
+            contiguous_text_runs: Vec::new(),
+            jobs: Vec::new(),
+            has_text_runs: false,
+        }
+    }
+
     fn wrap_any_text_in_anonymous_block_container(&mut self) {
         let runs = std::mem::take(&mut self.contiguous_text_runs);
-        if runs.iter().all(is_only_document_white_space) {
+        if runs.iter().all(|r| is_only_document_white_space(&r.text)) {
```
mrobinson, line 121: Then this can just be:

```suggestion
        if runs.iter().all(FlexItemTextRun::is_only_document_white_space) {
```

components/layout_2020/formatting_contexts.rs
mrobinson: That's not necessary. I just didn't understand how this was being disabled.

components/layout_2020/taffy/stylo_taffy/convert.rs
```diff
@@ -0,0 +1,339 @@
...
+            input.inside(),
+            stylo::DisplayInside::Flow | stylo::DisplayInside::FlowRoot
+        )
+}
+
+#[inline]
+pub fn box_generation_mode(input: stylo::Display) -> taffy::BoxGenerationMode {
+    match input.inside() {
+        stylo::DisplayInside::None => taffy::BoxGenerationMode::None,
+        // stylo::DisplayInside::Contents => display = taffy::BoxGenerationMode::Contents,
```
mrobinson, line 117: Please remove this commented out code.

components/layout_2020/taffy/stylo_taffy/wrapper.rs
```diff
@@ -0,0 +1,230 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use std::ops::Deref;
+
+use style::properties::ComputedValues;
+
+use super::convert;
+
+/// A wrapper struct for anything that Deref's to a [`stylo::ComputedValues`], which implements Taffy's layout traits
+/// and can used with Taffy's layout algorithms.
```
mrobinson, line 12:
```suggestion
/// A wrapper struct for anything that Deref's to a [`stylo::ComputedValues`], which 
/// implements Taffy's layout traits and can used with Taffy's layout algorithms.
```

components/layout_2020/taffy/stylo_taffy/wrapper.rs
```diff
@@ -0,0 +1,230 @@
...
+pub struct TaffyStyloStyle<T: Deref<Target = ComputedValues>>(pub T);
+
+// Deref<stylo::ComputedValues> impl
```
mrobinson, line 15: Please remove this.

components/layout_2020/taffy/stylo_taffy/wrapper.rs
```diff
@@ -0,0 +1,230 @@
...
+impl<T: Deref<Target = ComputedValues>> From<T> for TaffyStyloStyle<T> {
+    fn from(value: T) -> Self {
+        Self(value)
+    }
+}
+
+// CoreStyle impl
```
mrobinson, line 22: I'm not sure this comment adds much. Better to just remove it.

components/layout_2020/taffy/stylo_taffy/wrapper.rs
```diff
@@ -0,0 +1,230 @@
...
+        taffy::Rect {
+            left: taffy::LengthPercentage::Length(border_styles.border_left_width.to_f32_px()),
+            right: taffy::LengthPercentage::Length(border_styles.border_right_width.to_f32_px()),
+            top: taffy::LengthPercentage::Length(border_styles.border_top_width.to_f32_px()),
+            bottom: taffy::LengthPercentage::Length(border_styles.border_bottom_width.to_f32_px()),
+        }
+    }
+}
+
+// GridContainerStyle impl
```
mrobinson, line 135: Ditto.

components/layout_2020/taffy/stylo_taffy/wrapper.rs
```diff
@@ -0,0 +1,230 @@
...
+        convert::item_alignment(self.0.get_position().align_items.0)
+    }
+
+    #[inline]
+    fn justify_items(&self) -> Option<taffy::AlignItems> {
+        convert::item_alignment(self.0.get_position().justify_items.computed.0)
+    }
+}
+
+// GridItemStyle impl
```
mrobinson, line 201: Ditto.

mrobinson (approved):
Looks good to me. I just have a few nits. Please address them before landing. Thank you!

mrobinson:
Oh, before landing please also update the PR message so that the commit message accurately reflects the new changes.

KnorpelSenf:
what a time to be alive

PR #32620, opened 2024-06-26
Create safe wrapper for JSFunctions

### LINKED ISSUE #32601: script: Safe way to create `Function`
**Describe the new feature:**
In #32572, we still need to use unsafe code to create `Funtion`. The `FunctionBinging` doesn't provide a safe way to create one. And considering many dom types don't really want `Function`. Instead, they just want the `*mut JSObject` from `*mut JSFunction`:
- https://github.com/servo/servo/blob/main/components/script/dom/bindings/interface.rs#L341
- https://github.com/servo/servo/blob/main/components/script/dom/promise.rs

We should propose a safe wrapper function(s) for `JS_NewFunction` and `JS_GetFunctionObject`. Here are the method I think we could add:

- [ ] `Function::new_native`: Create a `Rc<Function>` using `JS_NewFunction`, `JS_GetFunctionObject` and `Function::new`
- [ ] `Function::new_raw_obj`: Create a `*mut JSObject`(or handle?) using `JS_NewFunction` and `JS_GetFunctionObject`

Taym95:
Create safe wrapper for JSFunctions to hide `unsafe` code

---

- [x] These changes fix #32601

### DIFF
--- a/components/script/dom/bindings/function.rs
+++ b/components/script/dom/bindings/function.rs
@@ -31,12 +31,13 @@ macro_rules! native_raw_obj_fn {
         }
 
         unsafe {
+            let name: &std::ffi::CStr = $name;
             let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
                 *$cx,
                 Some(wrapper),
                 $nargs,
                 $flags,
-                $name.as_ptr() as *const std::ffi::c_char,
+                name.as_ptr() as *const std::ffi::c_char,
             );
             assert!(!raw_fun.is_null());
             crate::dom::bindings::import::module::jsapi::JS_GetFunctionObject(raw_fun)
--- a/components/script/dom/bindings/function.rs
+++ b/components/script/dom/bindings/function.rs
@@ -4,43 +4,38 @@
 
 #![allow(unsafe_code)]
 
-use std::ffi::c_char;
-use std::rc::Rc;
-
-use js::jsapi::{JSContext, JSObject, JS_GetFunctionObject, JS_NewFunction, Value};
-
-use crate::dom::bindings::codegen::Bindings::FunctionBinding::Function;
-use crate::dom::types::GlobalScope;
-
-type NativeFunction = unsafe extern "C" fn(*mut JSContext, u32, *mut Value) -> bool;
-
-pub struct FunctionBinding {}
-
-impl FunctionBinding {
-    pub fn new_native(call: NativeFunction, name: &[u8], nargs: u32, flags: u32) -> Rc<Function> {
-        let cx = GlobalScope::get_cx();
-        let fun_obj = Self::new_raw_obj(cx, call, name, nargs, flags);
+/// Defines a macro `native_fn!` to create a JavaScript function from a Rust function pointer.
+#[macro_export]
+macro_rules! native_fn {
+    ($call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
+        let cx = crate::dom::types::GlobalScope::get_cx();
+        let fun_obj = crate::native_raw_obj_fn!(cx, $call, $name, $nargs, $flags);
         unsafe { Function::new(cx, fun_obj) }
-    }
+    }};
+}
 
-    pub fn new_raw_obj(
-        cx: crate::script_runtime::JSContext,
-        call: NativeFunction,
-        name: &[u8],
-        nargs: u32,
-        flags: u32,
-    ) -> *mut JSObject {
+/// Defines a macro `native_raw_obj!` to create a raw JavaScript function object.
+#[macro_export]
+macro_rules! native_raw_obj_fn {
+    ($cx:expr, $call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
+        unsafe extern "C" fn wrapper(cx: *mut JSContext, argc: u32, vp: *mut JSVal) -> bool {
+            (WRAPPER_FN.unwrap())(cx, argc, vp)
+        }
+
+        static mut WRAPPER_FN: Option<unsafe fn(*mut JSContext, u32, *mut JSVal) -> bool> = None;
+
         unsafe {
-            assert_eq!(*name.last().unwrap(), b'\0');
-            let raw_fun = JS_NewFunction(
-                *cx,
-                Some(call),
-                nargs,
-                flags,
-                name.as_ptr() as *const c_char,
+            WRAPPER_FN = Some($call);
+            assert_eq!(*$name.last().unwrap(), b'\0');
+            let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
+                *$cx,
+                Some(wrapper),
+                $nargs,
+                $flags,
+                $name.as_ptr() as *const std::ffi::c_char,
             );
             assert!(!raw_fun.is_null());
-            JS_GetFunctionObject(raw_fun)
+            crate::dom::bindings::import::module::jsapi::JS_GetFunctionObject(raw_fun)
         }
-    }
+    }};
 }
--- a/components/script/dom/bindings/function.rs
+++ b/components/script/dom/bindings/function.rs
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#![allow(unsafe_code)]
-
 /// Defines a macro `native_fn!` to create a JavaScript function from a Rust function pointer.
 /// # Example
 /// ```
@@ -14,7 +12,10 @@ macro_rules! native_fn {
     ($call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
         let cx = crate::dom::types::GlobalScope::get_cx();
         let fun_obj = crate::native_raw_obj_fn!(cx, $call, $name, $nargs, $flags);
-        unsafe { Function::new(cx, fun_obj) }
+        #[allow(unsafe_code)]
+        unsafe {
+            Function::new(cx, fun_obj)
+        }
     }};
 }
 
@@ -26,10 +27,11 @@ macro_rules! native_fn {
 #[macro_export]
 macro_rules! native_raw_obj_fn {
     ($cx:expr, $call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
+        #[allow(unsafe_code)]
         unsafe extern "C" fn wrapper(cx: *mut JSContext, argc: u32, vp: *mut JSVal) -> bool {
             $call(cx, argc, vp)
         }
-
+        #[allow(unsafe_code)]
         unsafe {
             let name: &std::ffi::CStr = $name;
             let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
--- a/components/script/dom/bindings/function.rs
+++ b/components/script/dom/bindings/function.rs
@@ -5,6 +5,10 @@
 #![allow(unsafe_code)]
 
 /// Defines a macro `native_fn!` to create a JavaScript function from a Rust function pointer.
+/// # Example
+/// ```
+/// let js_function: Rc<Function> = native_fn!(my_rust_function, b"myFunction\0", 2, 0);
+/// ```
 #[macro_export]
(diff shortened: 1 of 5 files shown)


components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,45 @@
...
+use js::jsapi::{JSContext, JSObject, JS_GetFunctionObject, JS_NewFunction, Value};
+
+use crate::dom::bindings::codegen::Bindings::FunctionBinding::Function;
+use crate::dom::types::GlobalScope;
+
+type NativeFunction = unsafe extern "C" fn(*mut JSContext, u32, *mut Value) -> bool;
+
+pub struct FunctionBinding {}
+
+impl FunctionBinding {
```
Taym95, line 19: @gterzian @wusyong FYI.

components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use std::ffi::c_char;
+use std::rc::Rc;
+
+use js::jsapi::{JSContext, JSObject, JS_GetFunctionObject, JS_NewFunction, Value};
+
+use crate::dom::bindings::codegen::Bindings::FunctionBinding::Function;
+use crate::dom::types::GlobalScope;
+
+type NativeFunction = unsafe extern "C" fn(*mut JSContext, u32, *mut Value) -> bool;
```
wusyong, line 15: This will still require user to write unsafe code. I wish it could take a normal rust function and we trampoline or wrap it in the method.

I'll look if there's some possible approach.

components/script/dom/bindings/function.rs
wusyong: I was just thinking extend the Fuction type in codegen bindings. But I'm okay with any as long as the method is provided.

components/script/dom/bindings/function.rs
gterzian: Yes that sounds like a good idea, let us know what you find out.

components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,45 @@
...
+        nargs: u32,
+        flags: u32,
+    ) -> *mut JSObject {
+        unsafe {
+            let raw_fun = JS_NewFunction(
+                *cx,
+                Some(call),
+                nargs,
+                flags,
+                name.as_ptr() as *const c_char,
```
jdm, line 39: We should assert that this name ends in a null character. There's an existing assertion in one of the callers that can be moved.

components/script/dom/bytelengthqueuingstrategy.rs
```diff
@@ -70,44 +64,13 @@ impl ByteLengthQueuingStrategyMethods for ByteLengthQueuingStrategy {
...
+        // a Function that represents a reference to F,
+        // with callback context equal to globalObject’s relevant settings object.
+        global.set_byte_length_queuing_strategy_size(fun.clone());
+        Ok(fun)
     }
 }
-
-/// <https://streams.spec.whatwg.org/#byte-length-queuing-strategy-size-function>
-#[allow(unsafe_code)]
-unsafe extern "C" fn byte_length_queuing_strategy_size(
```
jdm, line 98: Why move this into another file?

Taym95: by mistake, I will put it back

components/script/dom/bindings/function.rs
sagudev: Can't we just accept Cstr instead of [u8]?

components/script/dom/bindings/function.rs
wusyong: Hmm, it seems the extern C function can't capture the rust function, and there's no way to trampoline either.
How about providing a `native_fn` macro instead of `new_native` then:

- The signature should be `native_fn!(call, name, nargs, flags)`.
- `call` is a rust function which signature should be `fn(*mut JSContext, u32, CallArgs) -> bool`
- `name` should be `String` or `&str`. We will convert it to Cstr in the macro.

So the user can get the extern C function via `let c_fn: Rc<Function> = native_fn!(rust_fn, name, nargs, flags)` after they define their rust function.

components/script/dom/bindings/function.rs
wusyong:
```suggestion
impl Function {
```

Let's just extend Function then. Creating another type could confuse people.

components/script/dom/bindings/function.rs
sagudev: > name should be String or &str. We will convert it to Cstr in the macro.

I think that would require new alloc (as CStr is null-terminated, hence one byte more). Also using C-string literals gives Cstr: https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#c-string-literals, so it should be easy enough to use.

Taym95: when I tried to call `call` from extern c wrapper I was getting:

```
can't capture dynamic environment in a fn item use the `|| { ... }` closure form instead
```

this how I could make it work :

```

#[macro_export]
macro_rules! native_fn {
    ($cx:expr, $rust_fn:expr, $name:expr, $nargs:expr, $flags:expr) => {{
        unsafe extern "C" fn wrapper(cx: *mut JSContext, argc: u32, vp: *mut JSVal) -> bool {
            (WRAPPER_FN.unwrap())(cx, argc, vp)
        }

        static mut WRAPPER_FN: Option<unsafe fn(*mut JSContext, u32, *mut JSVal) -> bool> = None;

        unsafe {
            WRAPPER_FN = Some($rust_fn);
            assert_eq!(*$name.last().unwrap(), b'\0');
            let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
                *$cx,
                Some(wrapper),
                $nargs,
                $flags,
                $name.as_ptr() as *const std::ffi::c_char,
            );
            assert!(!raw_fun.is_null());
            crate::dom::bindings::import::module::jsapi::JS_GetFunctionObject(raw_fun)
        }
    }};
}
```

Taym95: @wusyong what do you think?

wusyong: It seems fine when I test this:
```rust
#[macro_export]
macro_rules! native_raw_obj_fn {
    ($cx:expr, $call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
        unsafe extern "C" fn wrapper(cx: *mut JSContext, argc: u32, vp: *mut JSVal) -> bool {
            $call(cx, argc, vp)
        }

        unsafe {
            assert_eq!(*$name.last().unwrap(), b'\0');
            let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
                *$cx,
                Some(wrapper),
                $nargs,
                $flags,
                $name.as_ptr() as *const std::ffi::c_char,
            );
            assert!(!raw_fun.is_null());
            crate::dom::bindings::import::module::jsapi::JS_GetFunctionObject(raw_fun)
        }
    }};
}
```

Which case did you have trouble with? Count and ByteLenghtQueuing both can work in my test.

Taym95: Indeed, this works. I'm not sure what went wrong with my previous example.

components/script/dom/bindings/function.rs
Taym95: I used macro and removed this struct

wusyong:
LGTM but I don't have permission. Might need @gterzian to review.

components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![allow(unsafe_code)]
+
+/// Defines a macro `native_fn!` to create a JavaScript function from a Rust function pointer.
+#[macro_export]
```
sagudev, line 8: It would be nice to provide example of usage.

components/script/dom/bytelengthqueuingstrategy.rs
```diff
@@ -70,32 +68,20 @@ impl ByteLengthQueuingStrategyMethods for ByteLengthQueuingStrategy {
...
-            // a Function that represents a reference to F,
-            // with callback context equal to globalObject’s relevant settings object.
-            let fun_obj = JS_GetFunctionObject(raw_fun);
-            let fun = Function::new(cx, fun_obj);
-            global.set_byte_length_queuing_strategy_size(fun.clone());
-            Ok(fun)
-        }
+        // Step 2. Let F be !CreateBuiltinFunction(steps, 1, "size", « »,
+        // globalObject’s relevant Realm).
+        let fun = native_fn!(byte_length_queuing_strategy_size, b"size\0", 1, 0);
```
sagudev, line 73: Any reason why it can't be `c"size"`?

gterzian:
Are we ready to merge this one, or are there still ongoing discussions? Note this targets the wip stream branch, so we can still make changes later...

components/script/dom/bytelengthqueuingstrategy.rs
Taym95: @sagudev used  C-string literals, please check again

components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,45 @@
...
+            $call(cx, argc, vp)
+        }
+
+        unsafe {
+            let raw_fun = crate::dom::bindings::import::module::jsapi::JS_NewFunction(
+                *$cx,
+                Some(wrapper),
+                $nargs,
+                $flags,
+                $name.as_ptr() as *const std::ffi::c_char,
```
Taym95, line 39: @sagudev if I understand correctly, we don't need to `assert_eq!(*$name.last().unwrap(), b'\0');` check when we use C-string literals right?

sagudev: Indeed, but given that this is macro we cannot be sure that `$name` is really type of `CStr`, so we either need to keep that assert (runtime checking), or ensure that name is Cstr type (compiletime checking). This could be done using inner functions or `let name: &std::ffi::CStr = &name;`

wusyong (approved):

components/script/dom/bindings/function.rs
Taym95: done

components/script/dom/bindings/function.rs
```diff
@@ -0,0 +1,46 @@
...
+/// let js_function: Rc<Function> = native_fn!(my_rust_function, c"myFunction", 2, 0);
+/// ```
+#[macro_export]
+macro_rules! native_fn {
+    ($call:expr, $name:expr, $nargs:expr, $flags:expr) => {{
+        let cx = crate::dom::types::GlobalScope::get_cx();
+        let fun_obj = crate::native_raw_obj_fn!(cx, $call, $name, $nargs, $flags);
+        unsafe { Function::new(cx, fun_obj) }
+    }};
+}
```
sagudev, line 19: NIT: We could scope this with `#[allow(unsafe_code)]`, so the callers would not need this (as this is safe wrapper, right?)

sagudev (approved):

components/script/dom/bindings/function.rs
sagudev: But we could also land without this, I will let @gterzian decide.

gterzian: Ok let's do that and then land it! @Taym95

gterzian:
@Taym95 thank you!

components/script/dom/bindings/function.rs
sagudev: Just an idea, but maybe we can make replace macro with function that is generic over closure `native_fn<F: Fn>(c: F)`.

Jenifer232000:
Great improvement to the project! Wrapping JSFunctions in a safer way to hide unsafe code is a smart approach, and it's nice to see the build and tidy checks passing without issues. Clean implementation and thoughtful work on improving stability. Also, a little off-topic, but this thread reminded me of coding late while writing [geburtstagswünsche](https://geburtstagswünsche.io/) for a friend funny how dev work and real life mix sometimes. Nice contribution!

Issue #32623, opened 2024-06-26
Media: fix media resource concept for video


gterzian:
Spec: https://html.spec.whatwg.org/multipage/#media-resource

The `videoTracks` attribute of a media element must, per the [spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-videotracks):

> return a [live](https://html.spec.whatwg.org/multipage/infrastructure.html#live) [AudioTrackList](https://html.spec.whatwg.org/multipage/media.html#audiotracklist) object representing the audio tracks available in the [media element](https://html.spec.whatwg.org/multipage/media.html#media-element)'s [media resource](https://html.spec.whatwg.org/multipage/media.html#media-resource).

and the media resource](https://html.spec.whatwg.org/multipage/#media-resource) concept tells us that:

> the video data of the [media resource](https://html.spec.whatwg.org/multipage/#media-resource) is only that of the currently selected track.

Our current implementation, below, returns a `VideoTrackList` for all tracks, not just the selected one. 

https://github.com/servo/servo/blob/a730469b704878da9b484b0018f369438ce08851/components/script/dom/htmlmediaelement.rs#L2387

This makes this test assertion fail: https://searchfox.org/mozilla-central/rev/2f48061aef8c8976b73749ee845e7b85751f5f2f/testing/web-platform/tests/html/semantics/embedded-content/the-video-element/video_size_preserved_after_ended.html#26

Taym95:
@gterzian  I added this function to `VideoTrackList` to get selected tracks:

```
  pub fn selected_tracks(&self, window: &Window) -> DomRoot<VideoTrackList> {
        let selected_tracks: Vec<Dom<VideoTrack>> = self
            .tracks
            .borrow()
            .iter()
            .filter(|track| track.selected())
            .cloned()
            .collect();

        reflect_dom_object(
            Box::new(VideoTrackList {
                eventtarget: EventTarget::new_inherited(),
                tracks: DomRefCell::new(selected_tracks),
                media_element: self.media_element.clone(),
            }),
            window,
        )
    }
```
 
and use it:
```
      // https://html.spec.whatwg.org/multipage/#dom-media-videotracks
    fn VideoTracks(&self) -> DomRoot<VideoTrackList> {
        let window = window_from_node(self);
        let video_tracks_list = self
            .video_tracks_list
            .or_init(|| VideoTrackList::new(&window, &[], Some(self)));

        video_tracks_list.selected_tracks(&window)
    }
```
    
[...]

PR #32624, opened 2024-06-26
Fix html medial elem interaction between network eof and metadata playerevent

### LINKED ISSUE #31415: Short videos dispatch loadedmetadata before the metadata is present

gterzian:
Fix the interaction between `network_state`, `status`, and `ready_state`, so that the `loadedmetadata` event is fired when expected. 

The only assertion still failing is https://searchfox.org/mozilla-central/rev/2f48061aef8c8976b73749ee845e7b85751f5f2f/testing/web-platform/tests/html/semantics/embedded-content/the-video-element/video_size_preserved_after_ended.html#26. For which I have filed: https://github.com/servo/servo/issues/32623

---

- [ ] These changes fix #31415 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/htmlmediaelement.rs
+++ b/components/script/dom/htmlmediaelement.rs
@@ -1735,6 +1735,12 @@ impl HTMLMediaElement {
                 // Step 6.
                 self.change_ready_state(ReadyState::HaveMetadata);
 
+                // If we have already reached a succesful `process_response_eof`,
+                // we can move to `ReadyState::HaveEnoughData`
+                if self.network_state.get() == NetworkState::Idle && self.error.get().is_none() {
+                    self.change_ready_state(ReadyState::HaveEnoughData);
+                }
+
                 // Step 7.
                 let mut jumped = false;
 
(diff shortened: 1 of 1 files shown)


jdm (changes requested):
I spent a while thinking about this, and I'm going to open a separate PR with my own proposal. I find the logic complicated to follow when comparing against the specification; I believe @pylbrecht was correct to focus on removing any ready state changes from process_response_eof.

gterzian:
Ok thanks for looking into this @jdm, I have double-checked the the test results for `video_size_preserved_after_ended.html` in https://github.com/servo/servo/pull/32643 are the same as here(with the test still failing but only on the last assertions), so the outcome appears equivalent(at least for that specific part).

gterzian:
actually I want to double check what this does in terms of the tests apart from `/embedded-content/the-video-element/`

PR #32635, opened 2024-06-28
script: `document.visibilityState` and `document.hidden`

### LINKED ISSUE #32062: Implement Page Visibility
Spec: https://html.spec.whatwg.org/multipage/interaction.html#page-visibility

There are at least two places where we could run the [`update-the-visibility-state`](https://html.spec.whatwg.org/multipage/interaction.html#update-the-visibility-state) steps:

When the document loads: 

https://github.com/servo/servo/blob/10ec8565eaa0e9e86d3a6506ab1e83d5a41f51c8/components/script/dom/document.rs#L695

Which would match: https://html.spec.whatwg.org/multipage/browsing-the-web.html#updating-the-document:update-the-visibility-state

And when the document unloads:

https://github.com/servo/servo/blob/10ec8565eaa0e9e86d3a6506ab1e83d5a41f51c8/components/script/dom/document.rs#L2221

which would match: https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-documents:update-the-visibility-state

wusyong:
---

- [x] These changes fix #32062

- [x] There are tests for these changes: `tests/wpt/tests/html/dom/idlharness.https.html`

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -67,7 +67,6 @@ use url::Host;
 use uuid::Uuid;
 use webrender_api::units::DeviceIntRect;
 
-use super::bindings::trace::{HashMapTracedValues, NoTrace};
 use crate::animation_timeline::AnimationTimeline;
 use crate::animations::Animations;
 use crate::document_loader::{DocumentLoader, LoadType};
@@ -77,7 +76,7 @@ use crate::dom::bindings::callback::ExceptionHandling;
 use crate::dom::bindings::cell::{ref_filter_map, DomRefCell, Ref, RefMut};
 use crate::dom::bindings::codegen::Bindings::BeforeUnloadEventBinding::BeforeUnloadEvent_Binding::BeforeUnloadEventMethods;
 use crate::dom::bindings::codegen::Bindings::DocumentBinding::{
-    DocumentMethods, DocumentReadyState, NamedPropertyValue,
+    DocumentMethods, DocumentReadyState, DocumentVisibilityState, NamedPropertyValue,
 };
 use crate::dom::bindings::codegen::Bindings::EventBinding::Event_Binding::EventMethods;
 use crate::dom::bindings::codegen::Bindings::HTMLIFrameElementBinding::HTMLIFrameElement_Binding::HTMLIFrameElementMethods;
@@ -100,6 +99,7 @@ use crate::dom::bindings::refcounted::{Trusted, TrustedPromise};
 use crate::dom::bindings::reflector::{reflect_dom_object_with_proto, DomObject};
 use crate::dom::bindings::root::{Dom, DomRoot, DomSlice, LayoutDom, MutNullableDom};
 use crate::dom::bindings::str::{DOMString, USVString};
+use crate::dom::bindings::trace::{HashMapTracedValues, NoTrace};
 use crate::dom::bindings::xmlname::XMLName::Invalid;
 use crate::dom::bindings::xmlname::{
     namespace_from_domstring, validate_and_extract, xml_name_type,
@@ -153,6 +153,7 @@ use crate::dom::node::{
 use crate::dom::nodeiterator::NodeIterator;
 use crate::dom::nodelist::NodeList;
 use crate::dom::pagetransitionevent::PageTransitionEvent;
+use crate::dom::performanceentry::PerformanceEntry;
 use crate::dom::processinginstruction::ProcessingInstruction;
 use crate::dom::promise::Promise;
 use crate::dom::range::Range;
@@ -167,6 +168,7 @@ use crate::dom::touch::Touch;
 use crate::dom::touchevent::TouchEvent;
 use crate::dom::touchlist::TouchList;
 use crate::dom::treewalker::TreeWalker;
+use crate::dom::types::VisibilityStateEntry;
 use crate::dom::uievent::UIEvent;
 use crate::dom::virtualmethods::vtable_for;
 use crate::dom::webglrenderingcontext::WebGLRenderingContext;
@@ -470,6 +472,8 @@ pub struct Document {
     /// The set of all fonts loaded by this document.
     /// <https://drafts.csswg.org/css-font-loading/#font-face-source>
     fonts: MutNullableDom<FontFaceSet>,
+    /// <https://html.spec.whatwg.org/multipage/#visibility-state>
+    visibility_state: Cell<DocumentVisibilityState>,
 }
 
 #[derive(JSTraceable, MallocSizeOf)]
@@ -708,8 +712,9 @@ impl Document {
             return;
         }
 
-        // html.spec.whatwg.org/multipage/#history-traversal
-        // Step 4.6
+        // This step used to be Step 4.6 in html.spec.whatwg.org/multipage/#history-traversal
+        // But it's now Step 4 in https://html.spec.whatwg.org/multipage/#reactivate-a-document
+        // TODO: See #32687 for more information.
         let document = Trusted::new(self);
         self.window
             .task_manager()
@@ -722,9 +727,12 @@ impl Document {
                     if document.page_showing.get() {
                         return;
                     }
-                    // Step 4.6.2
+                    // Step 4.6.2 Set document's page showing flag to true.
                     document.page_showing.set(true);
-                    // Step 4.6.4
+                    // Step 4.6.3 Update the visibility state of document to "visible".
+                    document.update_visibility_state(DocumentVisibilityState::Visible);
+                    // Step 4.6.4 Fire a page transition event named pageshow at document's relevant
+                    // global object with true.
                     let event = PageTransitionEvent::new(
                         window,
                         atom!("pageshow"),
@@ -2220,9 +2228,12 @@ impl Document {
         // TODO: Step 1, increase the event loop's termination nesting level by 1.
         // Step 2
         self.incr_ignore_opens_during_unload_counter();
-        // Step 3-6
+        // Step 3-6 If oldDocument's page showing is true:
         if self.page_showing.get() {
+            // Set oldDocument's page showing to false.
             self.page_showing.set(false);
+            // Fire a page transition event named pagehide at oldDocument's relevant global object with oldDocument's
+            // salvageable state.
             let event = PageTransitionEvent::new(
                 &self.window,
                 atom!("pagehide"),
@@ -2233,7 +2244,8 @@ impl Document {
             let event = event.upcast::<Event>();
             event.set_trusted(true);
             let _ = self.window.dispatch_event_with_target_override(event);
-            // TODO Step 6, document visibility steps.
+            // Step 6 Update the visibility state of oldDocument to "hidden".
+            self.update_visibility_state(DocumentVisibilityState::Hidden);
         }
         // Step 7
         if !self.fired_unload.get() {
@@ -3297,6 +3309,7 @@ impl Document {
             mouse_move_event_index: Default::default(),
             resize_observers: Default::default(),
             fonts: Default::default(),
+            visibility_state: Cell::new(DocumentVisibilityState::Hidden),
         }
     }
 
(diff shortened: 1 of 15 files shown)


components/script/dom/document.rs
```diff
@@ -4082,6 +4090,31 @@ impl Document {
...
+            &self.global(),
+            visibility_state,
+            *self.global().performance().Now(),
+        );
+        self.window
+            .Performance()
+            .queue_entry(entry.upcast::<PerformanceEntry>());
+        // Step 4 TODO ScreenOrientation hasn't implemented yet
+        // Step 5 TODO ScreenOrientation hasn't implemented yet
+        // Step 6 ???
```
wusyong, line 4113: @gterzian I opened a draft for now because I want to know if I can skip this.
It seems Servo doesn't have the implementation in steps 4 & 5, and step 6 is so vague that it even has a comment discussing it.

components/script/dom/document.rs
```diff
@@ -4082,6 +4090,31 @@ impl Document {
...
+            &self.global(),
+            visibility_state,
+            *self.global().performance().Now(),
+        );
+        self.window
+            .Performance()
+            .queue_entry(entry.upcast::<PerformanceEntry>());
+        // Step 4 TODO ScreenOrientation hasn't implemented yet
+        // Step 5 TODO ViewTransition hasn't implemented yet
+        // Step 6 TODO This step is too vague
```
wusyong, line 4113: @gterzian, I opened a draft for now because I want to know if we could skip these.
It seems Servo hasn't implemented the parts required by steps 4 & 5. While step 6 is so vague that it even has a comment section discussing it.

gterzian: Yes on the TODO for 4 and 5, re 6 this is more like a hook for other specs to defines steps, so as those other specs are implemented they could run their visibility steps here. For example https://www.w3.org/TR/gamepad/#handling-visibility-change (even though they don't mention these steps direclty but I think that's the idea) cc @msub2

components/script/dom/document.rs
```diff
@@ -724,6 +728,8 @@ impl Document {
                     }
                     // Step 4.6.2
                     document.page_showing.set(true);
+                    // Step 4.6.3
```
mrobinson, line 731: Please add descriptive text with the step numbers as the numbers change all the time.

components/script/dom/document.rs
```diff
@@ -2233,7 +2239,8 @@ impl Document {
             let event = event.upcast::<Event>();
             event.set_trusted(true);
             let _ = self.window.dispatch_event_with_target_override(event);
-            // TODO Step 6, document visibility steps.
+            // Step 6
```
mrobinson, line 2242: The same here please.

mrobinson (changes requested):

components/script/dom/document.rs
wusyong: Does this mean I need to add the implementation of gamepad's step here?

msub2: As long as page visibility is being handled properly I can implement Gamepad-specific behavior in a future PR

wusyong: Alright, I updated the description text and I'll move this draft to PR soon.

wusyong:
I added `VisibilityStateEntry` to `mozilla/interfaces.html,` and I guess @jdm will also need to review this.
This PR adds `VisibilityStateEntry.webidl` which has `[Exposed=(Window)]`.

As for `/fetch/metadata/generated/css-font-face.https.sub.tentative.html`, I have no idea why this test is related to this PR. I saw other commits also fail with this. Is it okay to ignore it?

components/script/dom/document.rs
```diff
@@ -4082,6 +4094,35 @@ impl Document {
...
+        self.window
+            .Performance()
+            .queue_entry(entry.upcast::<PerformanceEntry>());
+        // Step 4 Run the screen orientation change steps with document.
+        // TODO ScreenOrientation hasn't implemented yet
+        // Step 5 Run the view transition page visibility change steps with document.
+        // TODO ViewTransition hasn't implemented yet
+        // Step 6 Run any page visibility change steps which may be defined in other specifications, with visibility
+        // state and document. Any other specs' visibility steps will go here.
+        // Step 7 Fire an event named visibilitychange at document, with its bubbles attribute initialized to true.
```
gterzian, line 4122: Let's add some spaces between steps please.

components/script/dom/document.rs
```diff
@@ -722,9 +726,12 @@ impl Document {
                     if document.page_showing.get() {
                         return;
                     }
-                    // Step 4.6.2
+                    // Step 4.6.2 Set document's page showing flag to true.
```
gterzian, line 729: We are here at https://html.spec.whatwg.org/multipage/#reactivate-a-document, but it's not clear if we are calling this in the right place. `#reactivate-a-document` is called from https://html.spec.whatwg.org/multipage/browsing-the-web.html#update-document-for-history-step-application, which seems to be rather at https://github.com/servo/servo/blob/ad01342f00089cbddb252d54ed55f90a20ce43da/components/script/dom/history.rs#L86

It would be good to figure out more about how Servo's implementation maps to the spec, and either move your changes to other parts, or highlight problems to be resolved as follow-ups. 

To me, the `set_activity` handler, where we are here, seems to correspond more to https://html.spec.whatwg.org/multipage/interaction.html#page-visibility:update-the-visibility-state (since we are receiving a message from the constellation telling us the system visibility has changed), but the code here seems to be #reactivate-a-document.

Answering this question requires looking at what messages the constellation sends for history updates, and for system visibility(perhaps both go through here, but we should at least clarify it).

components/script/dom/document.rs
```diff
@@ -2233,7 +2243,8 @@ impl Document {
             let event = event.upcast::<Event>();
             event.set_trusted(true);
             let _ = self.window.dispatch_event_with_target_override(event);
-            // TODO Step 6, document visibility steps.
+            // Step 6 Update the visibility state of oldDocument to "hidden".
+            self.update_visibility_state(DocumentVisibilityState::Hidden);
```
gterzian, line 2247: At least the unloading part is unambiguous.

gterzian (changes requested):
Thanks for a great start! There are some complications because it seems that Servo's implementation is out of step with the latest spec, so let's see how we can resolve these...

We should set the visibility as part of [`make-active`](https://html.spec.whatwg.org/multipage/browsing-the-web.html#make-active), which is called into by [`creating-a-new-browsing-context`](https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context), which I think corresponds to whenever a new ["about:blank" doc is created](https://github.com/servo/servo/blob/ad01342f00089cbddb252d54ed55f90a20ce43da/components/script/script_thread.rs#L2822). We should look into this, probably by filing a follow-up issue.

components/script/dom/document.rs
wusyong: I tried my best to track all changes, but I think https://html.spec.whatwg.org/multipage/#history-traversal has changed so much that it will require a big overhaul. That means not only [#reactivate-a-document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#reactivate-a-document) but also [#make-active](https://html.spec.whatwg.org/multipage/browsing-the-web.html#make-active) needs to be added. But many of their callers do not really exist yet in Servo, which brings us back to the problem of big overhaul again.

The good news is at least `set_activity` can be called correctly by pipeline and constellation when creating new context. So, the visibility state would be visible when the document is created and the history session is changed. Maybe we should add a follow up issue to track how to refactor document and history to match latest ["Populating a session history entry"](https://html.spec.whatwg.org/multipage/browsing-the-web.html#populating-a-session-history-entry) and ["Updating the document"](https://html.spec.whatwg.org/multipage/browsing-the-web.html#updating-the-document)?

gterzian: Yes opening an issue to document these problems is a good idea, also I think we should update the spec link above to be just [#make-active](https://html.spec.whatwg.org/multipage/browsing-the-web.html#make-active) (since history traversal calls into this, but it's not clear that we are really running these steps as part of history traversal).

wusyong: I opened #32687 and added a TODO comment to indicate future updates.

tests/wpt/meta/html/dom/idlharness.https.html.ini
```diff
@@ -1436,36 +1436,6 @@
   [SVGElement interface: attribute onbeforetoggle]
```
gterzian, line 1436: Besides the interface, we should also check the tests at https://github.com/servo/servo/tree/650af7db926e9cc070e136f21a56bd47d566d10a/tests/wpt/tests/page-visibility

For this they need to be unskipped at https://github.com/servo/servo/blob/650af7db926e9cc070e136f21a56bd47d566d10a/tests/wpt/include.ini

gterzian (changes requested):
Let's check the rest of the wpt tests on this.

tests/wpt/meta/html/dom/idlharness.https.html.ini
wusyong: Here's my result:
```
Ran 11 tests finished in 11.0 seconds.
  • 5 ran as expected. 2 tests skipped.
  • 4 tests timed out unexpectedly
  • 6 tests had unexpected subtest results
```

tests/wpt/meta/page-visibility/iframe-session-history.html.ini
```diff
@@ -0,0 +1,3 @@
+[iframe-session-history.html]
+  [pagehide should be called before visibilitychange, and visibilityState should be set to hidden at the right time]
```
gterzian, line 2: I think this one probably goes to the core of the issue you opened, would be good to link to it in the issue .

tests/wpt/meta/page-visibility/unload.html.ini
```diff
@@ -0,0 +1,4 @@
+[unload.html]
+  expected: TIMEOUT
+  [visibilitychange fires on unload]
+    expected: TIMEOUT
```
gterzian, line 4: I think these two unload ones should be able to pass, since the unload code is straight forward.

gterzian (changes requested):
I think for the new test failures we should spend some time figuring out why they are failing, and either file a follow-up(if it requires some other broader changes), or fix them. Seems like the one relatd to unload should not be a problem, while the iframe ones may require first https://github.com/servo/servo/issues/31973, but it would still be good to try to debug them a bit.

tests/wpt/meta/page-visibility/unload.html.ini
wusyong: I'll investigate what's happening.

wusyong: It looks like we do need `#make-active`. But perhaps we should discuss how to update all these in next week's meeting.
This PR became a feature creep IMHO. We should discuss how to deal with outdated steps of history session before adding `update_visibility_state` into them.

gterzian: I think we need an issue similar to https://github.com/servo/servo/issues/31182 with a little investigation for each tests, and also perhaps some architectural changes needed. Yes let's discuss next week.

wusyong: I opened #32742 and I'll look into #32687 next to see if we should fix the loading steps first.

components/script/dom/visibilitystateentry.rs
```diff
@@ -0,0 +1,71 @@
...
+            entry: PerformanceEntry::new_inherited(
+                name,
+                DOMString::from("visibility-state"),
+                timestamp,
+                0.,
+            ),
+        }
+    }
+
+    #[allow(crown::unrooted_must_root)]
```
gterzian, line 40: I think you can move this to `new_inherited`  if you create the entry within the call to `reflect_`

gterzian (approved):
LGTM, with one nit.

gterzian:
@mrobinson This still needs your approval for the changes you requested.

mrobinson (approved):

gterzian:
> TIMEOUT [expected PASS] /css/compositing/mix-blend-mode/mix-blend-mode-video.html

It would be good to have an idea as to why it is timing-out, and add it to https://github.com/servo/servo/issues/32742

wusyong:
@gterzian I found main branch will also timeout. After tracing the commits, I found it's caused by #32643. Looks like the PR didn't run all wpt test when merging and there are some wpt fails after that commits.

wusyong:
#32763 opened

wusyong:
> Might be intermittent, would be good to open an issue then.

#32763 opened but I don't have permission to add labels yet. Might need your help to add `I-intermittent `

PR #32636, opened 2024-06-28
webgpu: Implement proper async pipeline creation and GPUPipelineError

### LINKED ISSUE #31591: Implement WebGPU async pipeline creation
Spec: https://gpuweb.github.io/gpuweb/#async-pipeline-creation
Stubs: https://github.com/servo/servo/pull/31194

How to do it: 

1. Store the promise on the device([here](https://github.com/servo/servo/pull/31194/files#diff-57b1e4d9cc2141b425d998d66541f5441c3e1ce4955c4713462e02132ba295e3R746)). Perhaps using a map of compute pipeline id to promise.  
2. Resolve the promise in [handle_server_msg](https://github.com/servo/servo/blob/3a5ca785d3ecc1fd6cb5a519cf1a91ac61e15c8c/components/script/dom/gpudevice.rs#L177)

sagudev:
~~This will require some changes in CTS: https://github.com/gpuweb/cts/issues/3809~~ DONE

Reviewable per commit.

---

- [x] These changes fix #31591

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -220,135 +220,114 @@ impl GPUDevice {
         }
     }
 
-    fn parse_render_pipeline<'c>(
+    fn parse_render_pipeline(
         &self,
         descriptor: &GPURenderPipelineDescriptor,
     ) -> (
         Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)>,
-        Option<RenderPipelineDescriptor<'c>>,
+        RenderPipelineDescriptor<'static>,
     ) {
-        let mut valid = true;
-
         let (layout, implicit_ids, _) = self.get_pipeline_layout_data(&descriptor.parent.layout);
 
-        let desc = if valid {
-            Some(wgpu_pipe::RenderPipelineDescriptor {
-                label: convert_label(&descriptor.parent.parent),
-                layout,
-                cache: None,
-                vertex: wgpu_pipe::VertexState {
+        let desc = wgpu_pipe::RenderPipelineDescriptor {
+            label: convert_label(&descriptor.parent.parent),
+            layout,
+            cache: None,
+            vertex: wgpu_pipe::VertexState {
+                stage: wgpu_pipe::ProgrammableStageDescriptor {
+                    module: descriptor.vertex.parent.module.id().0,
+                    entry_point: Some(Cow::Owned(descriptor.vertex.parent.entryPoint.to_string())),
+                    constants: Cow::Owned(HashMap::new()),
+                    zero_initialize_workgroup_memory: true,
+                },
+                buffers: Cow::Owned(
+                    descriptor
+                        .vertex
+                        .buffers
+                        .iter()
+                        .map(|buffer| wgpu_pipe::VertexBufferLayout {
+                            array_stride: buffer.arrayStride,
+                            step_mode: match buffer.stepMode {
+                                GPUVertexStepMode::Vertex => wgt::VertexStepMode::Vertex,
+                                GPUVertexStepMode::Instance => wgt::VertexStepMode::Instance,
+                            },
+                            attributes: Cow::Owned(
+                                buffer
+                                    .attributes
+                                    .iter()
+                                    .map(|att| wgt::VertexAttribute {
+                                        format: convert_vertex_format(att.format),
+                                        offset: att.offset,
+                                        shader_location: att.shaderLocation,
+                                    })
+                                    .collect::<Vec<_>>(),
+                            ),
+                        })
+                        .collect::<Vec<_>>(),
+                ),
+            },
+            fragment: descriptor
+                .fragment
+                .as_ref()
+                .map(|stage| wgpu_pipe::FragmentState {
                     stage: wgpu_pipe::ProgrammableStageDescriptor {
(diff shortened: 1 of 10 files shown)


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/9962307544/job/27527741459

sagudev:
NOTE-TO-SELF: `webgpu:api,validation,getBindGroupLayout:index_range,auto_layout:*` also fails on firefox, open issue on wgpu.

components/script/dom/gpurenderpipeline.rs
```diff
@@ -86,13 +78,28 @@ impl GPURenderPipelineMethods for GPURenderPipeline {
...
+            .channel
+            .0
+            .send(WebGPURequest::RenderGetBindGroupLayout {
+                device_id: self.device.id().0,
+                pipeline_id: self.render_pipeline.0,
+                index,
+                id,
+            })
+        {
+            warn!("Failed to send WebGPURequest::ComputeGetBindGroupLayout {e:?}");
```
jdm, line 96: s/Compute/Render/

components/webgpu/ipc_messages/recv.rs
```diff
@@ -103,6 +103,8 @@ pub enum WebGPURequest {
         compute_pipeline_id: id::ComputePipelineId,
         descriptor: ComputePipelineDescriptor<'static>,
         implicit_ids: Option<(id::PipelineLayoutId, Vec<id::BindGroupLayoutId>)>,
+        /// present only on ASYNC versions
+        sender: Option<IpcSender<WebGPUResponse>>,
```
jdm, line 107: Should we call this async_sender, in that case?

components/webgpu/wgpu_thread.rs
```diff
@@ -448,7 +469,25 @@ impl WGPU {
                                 Some(render_pipeline_id),
                                 implicit)
                             );
-                            self.maybe_dispatch_wgpu_error(device_id, error);
+
+                            if let Some(sender) = sender {
```
jdm, line 473: It would be nice to avoid the code duplication here.

components/script/dom/gpudevice.rs
```diff
@@ -214,6 +220,139 @@ impl GPUDevice {
...
+                }),
+                multisample: wgt::MultisampleState {
+                    count: descriptor.multisample.count,
+                    mask: descriptor.multisample.mask as u64,
+                    alpha_to_coverage_enabled: descriptor.multisample.alphaToCoverageEnabled,
+                },
+                multiview: None,
+            })
+        } else {
+            /*self.dispatch_error(webgpu::Error::Validation(String::from(
```
jdm, line 348: Can we either convert this into a clearer TODO comment or remove the commented code?

components/script/dom/gpudevice.rs
```diff
@@ -214,6 +220,139 @@ impl GPUDevice {
...
+        descriptor: &GPURenderPipelineDescriptor,
+    ) -> (
+        Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)>,
+        Option<RenderPipelineDescriptor<'c>>,
+    ) {
+        let mut valid = true;
+
+        let (layout, implicit_ids, _) = self.get_pipeline_layout_data(&descriptor.parent.layout);
+
+        let desc = if valid {
```
jdm, line 234: Valid is always true here. What is the intention behind this code?

components/script/dom/gpudevice.rs
```diff
@@ -214,6 +220,139 @@ impl GPUDevice {
         }
     }
 
+    fn parse_render_pipeline<'c>(
+        &self,
+        descriptor: &GPURenderPipelineDescriptor,
+    ) -> (
+        Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)>,
+        Option<RenderPipelineDescriptor<'c>>,
```
jdm, line 228: `c is not related to any method argument. Is that intentional? What is the risk if the output lifetime is completely made up?

jdm (approved):

components/webgpu/wgpu_thread.rs
sagudev: You mean duplication between render/compute pipeline? They differentiate in `CreateRenderPipelineError` on `CreateComputePipelineError` and in responses they are sending, and I think dedup would require too much complexity (generics) for not enough profit (happens only twice).

jdm: 👍

components/script/dom/gpudevice.rs
sagudev: Yes, it's intentional to make it compile. Output lifetime is automatically determined by compiler (to be 'static, but the compiler cannot determine that from function signature), so rewrote it to static lifetime to make it more obvious.

components/script/dom/gpudevice.rs
sagudev: Ah, I planned to use https://docs.rs/wgpu/latest/wgpu/struct.ColorWrites.html#method.from_bits_retain to use validness check in wgpu-core, but I forgot.

sagudev: And it makes `webgpu:api,validation,render_pipeline,fragment_state:targets_write_mask:*` work.

jdm (approved):

PR #32642, opened 2024-06-29
webgpu: Implement ShaderCompilationInfo

sagudev:
[`webgpu:api,operation,shader_module,compilation_info:*`](https://gpuweb.github.io/cts/standalone/?runnow=1&q=webgpu:api,operation,shader_module,compilation_info:*) passes. There are many new expectations (they were crashes before).

---

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -1032,7 +1032,7 @@ impl AsyncWGPUListener for GPUDevice {
                     promise.resolve_native(&error);
                 },
             },
-            _ => unreachable!("Wrong response recived on AsyncWGPUListener for GPUDevice"),
+            _ => unreachable!("Wrong response received on AsyncWGPUListener for GPUDevice"),
         }
     }
 }
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -24,7 +24,7 @@ use super::bindings::error::Fallible;
 use super::gpu::AsyncWGPUListener;
 use super::gpudevicelostinfo::GPUDeviceLostInfo;
 use super::gpusupportedlimits::GPUSupportedLimits;
-use super::types::GPUError;
+use super::types::{GPUCompilationInfo, GPUError};
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::EventBinding::EventInit;
 use crate::dom::bindings::codegen::Bindings::EventTargetBinding::EventTargetMethods;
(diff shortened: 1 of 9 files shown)


components/script/dom/gpudevice.rs
sagudev, line 1: The promise is created and resolved here instead of in GPUCompilationInfo, because it's more easier to implement (`response_async` requires GPUObject to be already made).

gterzian: I don't think the async listener belongs on `GPUCompilationInfo`, but I do think it belongs on the shader module itself, which is accessible in `CreateShaderModule`, not on the device.

gterzian (changes requested):

components/webgpu/dom_messages.rs
```diff
@@ -25,11 +25,54 @@ use wgc::pipeline::{ComputePipelineDescriptor, RenderPipelineDescriptor};
...
+            // Based on https://searchfox.org/mozilla-central/rev/5b037d9c6ecdb0729f39ad519f0b867d80a92aad/gfx/wgpu_bindings/src/server.rs#353
+            fn len_utf16(s: &str) -> u64 {
+                s.chars().map(|c| c.len_utf16() as u64).sum()
+            }
+            let start = location.offset as usize;
+            let end = start + location.length as usize;
+            let line_start = source[0..start].rfind('\n').map(|pos| pos + 1).unwrap_or(0);
+            Self {
+                line_number: location.line_number as u64,
+                line_pos: len_utf16(&source[line_start..start]) + 1,
```
gterzian, line 62: maybe wrap overflows here and above?

gterzian (changes requested):

components/script/dom/gpudevice.rs
sagudev: Ups, my bad, I actually meant `GPUShaderModule` not `GPUCompilationInfo`.  We need to store the promise inside of `GPUShaderModule` and we need to have it constructed for `response_async`, ...

Hrm, while writing this I remembered that we can simply construct promise and GPUShaderModule before `response_async` 😅, so I did that in https://github.com/servo/servo/pull/32642/commits/26f9e698ec2e7c84d6b6eb2fc915b91c2d473eac.

components/script/dom/gpushadermodule.rs
```diff
@@ -86,6 +88,18 @@ impl GPUShaderModuleMethods for GPUShaderModule {
...
 }
 
+impl AsyncWGPUListener for GPUShaderModule {
+    fn handle_response(&self, response: Option<WebGPUResponseResult>, promise: &Rc<Promise>) {
+        match response {
+            Some(Ok(WebGPUResponse::CompilationInfo(info))) => {
+                let info = GPUCompilationInfo::from(&self.global(), info);
+                promise.resolve_native(&info);
+            },
+            _ => unreachable!("Wrong response recived on AsyncWGPUListener for GPUDevice"),
```
gterzian, line 98: received

gterzian (approved):
LGTM with one nit

components/webgpu/dom_messages.rs
sagudev: I think there would be no overflow because everything in encoded in utf-8 internally which uses more code points than utf-16, and we also get results from wgpu/naga in `u32` .

PR #32651, opened 2024-07-01
Add production-stripped cargo profile

jschwe:
Stripping the binary significantly reduces the size  (in my tests to about 1/3 of the original size). Enabling LTO and setting `codegen-units = 1` allows further size optimizations at the cost of increased compile-time.

Probably not everyone wants a stripped binary, since it makes backtraces less useful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,6 +153,12 @@ debug-assertions = true
 inherits = "release"
 debug-assertions = false
 
+[profile.production-stripped]
+inherits = "production"
+strip = true
+lto = "thin"
+codegen-units = 1
+
 [patch.crates-io]
 # If you need to temporarily test Servo with a local fork of some upstream
 # crate, add that here. Use the form:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,12 +152,12 @@ debug-assertions = true
 [profile.production]
 inherits = "release"
 debug-assertions = false
+lto = "thin"
+codegen-units = 1
 
 [profile.production-stripped]
 inherits = "production"
 strip = true
-lto = "thin"
-codegen-units = 1
 
 [patch.crates-io]
 # If you need to temporarily test Servo with a local fork of some upstream


Cargo.toml
```diff
@@ -153,6 +153,12 @@ debug-assertions = true
 inherits = "release"
 debug-assertions = false
 
+[profile.production-stripped]
+inherits = "production"
+strip = true
+lto = "thin"
+codegen-units = 1
```
mrobinson, line 160: Hrm. Perhaps it make sense to simply make the production profile have these properties? The only thing here I have doubts about is `strip`, which would make traces from nightlies a bit useless. Perhaps that can be exposed an an option somehow?

mukilan: Would it make sense to rename the current `production ` to `nightly` or `production-unstripped` so that the new `production` profile is the stripped version? In this case I think the nightly CI workflow will need to be updated.

jschwe: > Perhaps it make sense to simply make the production profile have these properties?

From my point of view it would make sense to add the `lto` and `codegen-units` options to `production`. I was just not sure if people might be concerned about increased memory usage (during link-time when using `production` )

> Perhaps that can be exposed an an option somehow?

If `production-stripped` would differ by the `stripped` option, then the name would be fitting, and it would be "like an option" I guess.

jschwe: > Would it make sense to rename the current production to nightly or production-unstripped so that the new production profile is the stripped version?

I think in many cases you probably do want the debug information to get useful backtraces. My motivation is mainly to easily build a stripped version for benchmarks, since the `.so` size influences the time it takes until the first page is rendered.

sagudev: > since the `.so` size influences the time it takes until the first page is rendered.

Interesting, do you have any measurements?

mukilan (approved):
Approving this as it seems there are no strong objections.  The discussions could perhaps be continued on Zulip.

PR #32652, opened 2024-07-01
android: fix application name used in ./mach run

mukilan:
This patch also switches the code to python format strings
so that the application name can be interpolated where
needed.

---
- [x] These changes do not require tests because they only modify mach command logic.

### DIFF
--- a/python/servo/post_build_commands.py
+++ b/python/servo/post_build_commands.py
@@ -33,6 +33,9 @@ from servo.command_base import (
 )
 
 
+ANDROID_APP_NAME = "org.servo.servoshell"
+
+
 def read_file(filename, if_exists=False):
     if if_exists and not path.exists(filename):
         return None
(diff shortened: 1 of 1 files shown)


python/servo/post_build_commands.py
```diff
@@ -115,10 +116,10 @@ def run(self, params, build_type: BuildType, android=None, debugger=False, debug
             if gst_debug:
                 extra += " -e gstdebug " + gst_debug
             script += [
-                "am start " + extra + " org.servo.servoshell/org.servo.servoshell.MainActivity",
+                f"am start {extra} {app_name}/{app_name}.MainActivity",
                 "sleep 0.5",
-                "echo Servo PID: $(pidof org.servo.servoshell)",
-                "logcat --pid=$(pidof org.servo.sevoshell)",
```
mukilan, line 121: This is the typo fixed by this patch.

python/servo/post_build_commands.py
```diff
@@ -103,8 +103,9 @@ def run(self, params, build_type: BuildType, android=None, debugger=False, debug
                 print("Android on-device debugging is not supported by mach yet. See")
                 print("https://github.com/servo/servo/wiki/Building-for-Android#debugging-on-device")
                 return
+            app_name = "org.servo.servoshell"
```
mrobinson, line 106: I think it makes sense to move this to a global constant named something like `ANDROID_APP_NAME`.

mrobinson (approved):
Thanks for this fix!

PR #32657, opened 2024-07-01
layout: Add support for table captions

mrobinson:
This adds initial support for table captions. To do this, the idea of
the table wrapper becomes a bit more concrete. Even so, the wrapper is
still reponsible for allocating space for the grid's border and padding,
as those properties are specified on the wrapper and not grid in CSS.

In order to account for this weirdness of HTML/CSS captions and grid are
now laid out and placed with a negative offset in the table wrapper
content rect.

---

### DIFF
--- a/components/layout_2020/formatting_contexts.rs
+++ b/components/layout_2020/formatting_contexts.rs
@@ -7,6 +7,7 @@ use serde::Serialize;
 use servo_arc::Arc;
 use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
+use style::selector_parser::PseudoElement;
 use style::values::specified::text::TextDecorationLine;
 
 use crate::context::LayoutContext;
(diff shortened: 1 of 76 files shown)


components/layout_2020/formatting_contexts.rs
```diff
@@ -59,12 +60,21 @@ pub(crate) enum NonReplacedFormattingContextContents {
 
 /// The baselines of a layout or a [`crate::fragment_tree::BoxFragment`]. Some layout
 /// uses the first and some layout uses the last.
-#[derive(Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize)]
```
mukilan, line 63: Maybe this struct can be marked as `Copy` as well?

components/layout_2020/table/construct.rs
```diff
@@ -825,8 +840,25 @@ where
                     ::std::mem::forget(box_slot);
                 },
                 DisplayLayoutInternal::TableCaption => {
-                    // TODO: Handle table captions.
-                    // We are doing this until we have actually set a Box for this `BoxSlot`.
```
mukilan, line 829: This comment is for the  `::std::mem::forget(box_slot);` line which is still present in this patch. Should we retain the comment? The other match arms still have it and it makes sense to comment on `mem::forget` calls since that is not seen in typical code that just lets the destructor run.

components/layout_2020/table/mod.rs
```diff
@@ -86,10 +86,24 @@ pub type TableSize = Size2D<usize, UnknownUnit>;
...
     #[serde(skip_serializing)]
     style: Arc<ComputedValues>,
 
+    /// The style of this table's grid. This is an anonymous style based on the table's style, but
+    /// eliminating all the properties handled by the "wrapper."
+    #[serde(skip_serializing)]
+    grid_style: Arc<ComputedValues>,
+
+    /// The [`BaseFragmentInfo`] for this table's grid. This is necessary so that when the
+    /// grid has a background image, it can be associatd with the table's node.
```
mukilan, line 101:
```suggestion
    /// grid has a background image, it can be associated with the table's node.
```

mukilan (approved):
Thanks for walking through the changes!

mrobinson:
@mukilan thanks for the review. I've pushed changes to address your comments and also fix some new clippy warnings.

PR #32662, opened 2024-07-02
compositor: remove cursor_pos and update step on new frame

### LINKED ISSUE #32632: Remove IOCompositor's current_pos field
**Describe the new feature:**
I was tracing a bug that caused the cursor to constantly reset to the default when in animation. And I found `current_pos` isn't set after initialized, and it will always be (0, 0) pass to hit test.

While it can be fixed by setting the position properly, this is a good opportunity to remove it. I mentioned in [previous issue](https://github.com/servo/servo/issues/32485#issuecomment-2172995527) that we want embedder that could control and handle more stuff by itself. Setting should be one of them. Unless the embedder explicitly moves the mouse or other event it prefers to, the compositor shouldn't perform a hit test.

Wdyt? Is it okay to remove it? Or we just set the position?

wusyong:
There's a bug when updating the cursor on a new frame because `cursor_pos` isn't updated at all. I would like to take this chance to remove it. It will be the embedder's responsibility to decide when it should update its cursor. In general cases, users will move their mouse, and the cursor should be updated.

---

- [x] These changes fix #32632

- [x] These changes do not require tests, but the servoshell should be built successfully.

### DIFF
--- a/components/compositing/windowing.rs
+++ b/components/compositing/windowing.rs
@@ -59,7 +59,7 @@ pub enum EmbedderEvent {
     LoadUrl(TopLevelBrowsingContextId, ServoUrl),
     /// Sent when a mouse hit test is to be performed.
     MouseWindowEventClass(MouseWindowEvent),
-    /// Sent when a mouse move.
+    /// Sent when a mouse move. This will send back a `EmbedderMsg::SetCursor` message.
     MouseWindowMoveEventClass(DevicePoint),
     /// Touch event: type, identifier, point
     Touch(TouchEventType, TouchId, DevicePoint),
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -184,9 +184,6 @@ pub struct IOCompositor<Window: WindowMethods + ?Sized> {
     /// Current mouse cursor.
     cursor: Cursor,
 
-    /// Current cursor position.
-    cursor_pos: DevicePoint,
-
     /// Offscreen framebuffer object to render our next frame to.
     /// We use this and `prev_offscreen_framebuffer` for double buffering when compositing to
     /// [`CompositeTarget::Fbo`].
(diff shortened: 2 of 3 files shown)


mrobinson (changes requested):
I don't think we should remove support for this. Perhaps it is the embedder's responsibility to manage cursor changes, but it is up to the engine to tell the embedder when a cursor change is requested. The embedder cannot manage this properly because it doesn't know anything about the WebRender display list.

Here's an example of where this is managed in WebKitGTK and I'm sure all other web engines are managed similarly: https://github.com/WebKit/WebKit/blob/d72f4bd734ae263071bcd38db5ded174974a5bd4/Source/WebKit/UIProcess/API/gtk/PageClientImpl.cpp#L174

What's the bug with `cursor_pos`? Can we fix that?

wusyong:
> What's the bug with cursor_pos? Can we fix that?

Compositor has two places to perform `update_cursor` and send `EmbedderMsg::SetCursor` to embedder. One is when it [receives mouse move event](https://github.com/servo/servo/blob/main/components/compositing/compositor.rs#L1558), the other is when there's [new webrender frame](https://github.com/servo/servo/blob/main/components/compositing/compositor.rs#L627). The problem with latter is it uses `cursor_pos` which never got updated, so it will always be `(0, 0)` and result in default cursor when calling `hit_test_at_point`.

While we can fix it by setting `cursor_pos` as well during the mouse move event, I feel like it's a good chance to remove it. So moving the mouse becomes the only way to update the cursor, which matches general cases. And if the embedder really wants to get more `EmbedderMsg::SetCursor` messages, it can just send more mouse move events manually. Besides, getting `EmbedderMsg::SetCursor` messages because of new webrender frame is confusing to embedder from my experience. If we want the cursor to get updated without moving, why not let the embedder manually request it when performing actions like `present`?

components/compositing/windowing.rs
```diff
@@ -59,7 +59,7 @@ pub enum EmbedderEvent {
     LoadUrl(TopLevelBrowsingContextId, ServoUrl),
     /// Sent when a mouse hit test is to be performed.
     MouseWindowEventClass(MouseWindowEvent),
-    /// Sent when a mouse move.
+    /// Sent when a mouse move. This will send back a `EmbedderMsg::SetCursor` message.
     MouseWindowMoveEventClass(DevicePoint),
```
wusyong, line 63: I hope the updated doc comments can clarify the goal of this issue. I think it's reasonable that only the embedder can request the cursor message.

wusyong: I think you are right. I'll close this one and set the cursor_pos instead.

PR #32663, opened 2024-07-02
Mark `test/wpt` and all subdirecdtories as vendored in `.gitattributes`

sagudev:
Instead of only marking the directory `tests`, mark `tests/wpt` and all subdirectories as vendored for language statistics. Although metadata and and `_mozilla` are not vendored, they are probably not relevant for language statistics.

---

- [x] These changes do not require tests because it only applies to GitHub lang stats

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 * text=auto eol=lf
-tests/* linguist-vendored
+tests/wpt/** linguist-vendored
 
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

sagudev:
This is much more sensible now:
52% of code is in Rust

PR #32665, opened 2024-07-02
webgpu: Update wgpu and revamp RenderPass

sagudev:
Like https://github.com/servo/servo/pull/32575 but this time revamping render pass, because of the changes landed in wgpu.

---

### DIFF
--- a/components/script/dom/gpurenderpassencoder.rs
+++ b/components/script/dom/gpurenderpassencoder.rs
@@ -3,11 +3,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use dom_struct::dom_struct;
-use webgpu::wgc::command::{render_commands as wgpu_render, RenderPass};
-use webgpu::{wgt, WebGPU, WebGPURequest};
+use webgpu::{wgt, RenderCommand, WebGPU, WebGPURenderPass, WebGPURequest};
 
 use super::bindings::codegen::Bindings::WebGPUBinding::GPUIndexFormat;
-use super::bindings::error::Fallible;
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::WebGPUBinding::{
     GPUColor, GPURenderPassEncoderMethods,
@@ -30,16 +28,15 @@ pub struct GPURenderPassEncoder {
     #[no_trace]
     channel: WebGPU,
     label: DomRefCell<USVString>,
-    #[ignore_malloc_size_of = "defined in wgpu-core"]
     #[no_trace]
-    render_pass: DomRefCell<Option<RenderPass>>,
+    render_pass: WebGPURenderPass,
     command_encoder: Dom<GPUCommandEncoder>,
 }
 
 impl GPURenderPassEncoder {
     fn new_inherited(
         channel: WebGPU,
-        render_pass: Option<RenderPass>,
+        render_pass: WebGPURenderPass,
         parent: &GPUCommandEncoder,
         label: USVString,
     ) -> Self {
@@ -47,7 +44,7 @@ impl GPURenderPassEncoder {
             channel,
             reflector_: Reflector::new(),
             label: DomRefCell::new(label),
-            render_pass: DomRefCell::new(render_pass),
+            render_pass,
             command_encoder: Dom::from_ref(parent),
         }
     }
@@ -55,7 +52,7 @@ impl GPURenderPassEncoder {
     pub fn new(
         global: &GlobalScope,
         channel: WebGPU,
-        render_pass: Option<RenderPass>,
+        render_pass: WebGPURenderPass,
         parent: &GPUCommandEncoder,
         label: USVString,
     ) -> DomRoot<Self> {
@@ -83,15 +80,17 @@ impl GPURenderPassEncoderMethods for GPURenderPassEncoder {
     }
 
     /// <https://gpuweb.github.io/gpuweb/#dom-gpuprogrammablepassencoder-setbindgroup>
-    #[allow(unsafe_code)]
-    fn SetBindGroup(&self, index: u32, bind_group: &GPUBindGroup, dynamic_offsets: Vec<u32>) {
-        if let Some(render_pass) = self.render_pass.borrow_mut().as_mut() {
-            wgpu_render::wgpu_render_pass_set_bind_group(
-                render_pass,
+    fn SetBindGroup(&self, index: u32, bind_group: &GPUBindGroup, offsets: Vec<u32>) {
+        if let Err(e) = self.channel.0.send(WebGPURequest::RenderPassCommand {
+            render_pass_id: self.render_pass.0,
+            render_command: RenderCommand::SetBindGroup {
                 index,
-                bind_group.id().0,
-                &dynamic_offsets,
-            )
+                bind_group_id: bind_group.id().0,
+                offsets,
+            },
+            device_id: self.command_encoder.device_id().0,
+        }) {
+            warn!("Error sending WebGPURequest::RenderPassCommand: {e:?}")
         }
     }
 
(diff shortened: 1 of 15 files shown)


components/webgpu/render_commands.rs
sagudev, line 1: RenderCommands were separated from WebGPURequest, because there is a lot of commands (and they all share getting of pass and erroring logic), we could also do this for ComputePass as a follow up (although it doesn't have as many commands).

components/script/dom/gpurenderpassencoder.rs
```diff
@@ -105,78 +104,99 @@ impl GPURenderPassEncoderMethods for GPURenderPassEncoder {
...
+                    r: *s[0],
+                    g: *s[1],
+                    b: *s[2],
+                    a: *s[3],
+                }
+            },
+        };
+        if let Err(e) = self.channel.0.send(WebGPURequest::RenderPassCommand {
+            render_pass_id: self.render_pass.0,
+            render_command: RenderCommand::SetBlendConstant(color),
```
gterzian, line 163: perhasp add a `send_render_command(&self , cmd: RenderCommand)` method?

components/webgpu/dom_messages.rs
```diff
@@ -255,6 +258,7 @@ pub enum WebGPURequest {
         device_id: id::DeviceId,
         pipeline_id: PipelineId,
     },
+    // COmpute pass
```
gterzian, line 261: '///'

components/webgpu/render_commands.rs
```diff
@@ -0,0 +1,151 @@
...
+//! Render pass commands
+
+use serde::{Deserialize, Serialize};
+use wgpu_core::command::{DynRenderPass, RenderPassError};
+use wgpu_core::global::Global;
+
+use crate::wgc::id;
+use crate::wgt;
+
+// https://github.com/gfx-rs/wgpu/blob/f25e07b984ab391628d9568296d5970981d79d8b/wgpu-core/src/command/render_command.rs#L17
```
gterzian, line 14: '///'

components/webgpu/render_commands.rs
```diff
@@ -0,0 +1,151 @@
...
+        offset: u64,
+    },
+    DrawIndexedIndirect {
+        buffer_id: id::BufferId,
+        offset: u64,
+    },
+    ExecuteBundles(Vec<id::RenderBundleId>),
+}
+
+pub fn do_render_command(
```
gterzian, line 75: 'apply_'?

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +872,70 @@ impl WGPU {
...
+                        // TODO: Command encoder state errors
+                        // self.maybe_dispatch_wgpu_error(device_id, error);
+                    },
+                    WebGPURequest::RenderPassCommand {
+                        render_pass_id,
+                        render_command,
+                        device_id,
+                    } => {
+                        if let Some((pass, valid)) = self.render_passes.get_mut(&render_pass_id) {
+                            *valid &= do_render_command(&self.global, pass, render_command).is_ok();
```
gterzian, line 911: why is the `&` necessary? I'm not familiar with this syntax.

components/webgpu/wgpu_thread.rs
```diff
@@ -1171,12 +1224,19 @@ impl WGPU {
...
                     WebGPURequest::DropComputePass(id) => {
-                        // Compute pass might have already ended
+                        // pass might have already ended
                         self.compute_passes.remove(&id);
                         if let Err(e) = self.script_sender.send(WebGPUMsg::FreeComputePass(id)) {
                             warn!("Unable to send FreeComputePass({:?}) ({:?})", id, e);
                         };
                     },
+                    WebGPURequest::DropRenderPass(id) => {
+                        // pass might have already ended
```
gterzian, line 1234: nit: Pass and could add a . at the end too.

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +872,70 @@ impl WGPU {
...
+                        let (pass, error) = gfx_select!(
+                            command_encoder_id => global.command_encoder_create_render_pass_dyn(
+                                command_encoder_id,
+                                desc,
+                        ));
+                        assert!(
+                            self.render_passes
+                                .insert(render_pass_id, (pass, error.is_none()))
+                                .is_none(),
+                            "RenderPass should not exist yet."
```
gterzian, line 900: But is this checked at `BeginRenderPass`?

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +872,70 @@ impl WGPU {
...
                         device_id,
+                        command_encoder_id,
                     } => {
-                        if let Some(render_pass) = render_pass {
-                            let command_encoder_id = render_pass.parent_id();
-                            let global = &self.global;
-                            let result = gfx_select!(command_encoder_id => global.render_pass_end(&render_pass));
-                            self.maybe_dispatch_wgpu_error(device_id, result.err())
+                        // TODO: Command encoder state error
+                        if let Some((mut pass, valid)) = self.render_passes.remove(&render_pass_id)
```
gterzian, line 925: Would it not be better to remove it on Drop?

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +872,70 @@ impl WGPU {
...
+                        );
+                        // TODO: Command encoder state errors
+                        // self.maybe_dispatch_wgpu_error(device_id, error);
+                    },
+                    WebGPURequest::RenderPassCommand {
+                        render_pass_id,
+                        render_command,
+                        device_id,
+                    } => {
+                        if let Some((pass, valid)) = self.render_passes.get_mut(&render_pass_id) {
```
gterzian, line 910: Why not only remove on drop, and rely on the error returned by wgpu? It should return an error if the pass has already ended. Since we are relying on wgpu for all other logic, I don't see why we should keep track of the ended status of the pass by removing it ourselves from the map.

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +872,70 @@ impl WGPU {
...
                     } => {
-                        if let Some(render_pass) = render_pass {
-                            let command_encoder_id = render_pass.parent_id();
-                            let global = &self.global;
-                            let result = gfx_select!(command_encoder_id => global.render_pass_end(&render_pass));
-                            self.maybe_dispatch_wgpu_error(device_id, result.err())
+                        // TODO: Command encoder state error
+                        if let Some((mut pass, valid)) = self.render_passes.remove(&render_pass_id)
+                        {
+                            if pass.end(&self.global).is_ok() && !valid {
```
gterzian, line 927: I think `end` cannot be ok if the pass is not valid, so we can simply check for an error?

gterzian (changes requested):
Couple of comments. 

In general I find it sometimes confusing that the `wgpu_thread` is keeping track of some state(including by removing objects from maps), or reporting errors, and relying on wgpu for other states and error. If we can remove some of this bookkeeping, or document clearly why this is necessary on top of what wgpu already does, it could clarify things. 

For example, by removing passes when they end, we sort of keep track of their ended state, which interacts with the need to also remove them when they drop. So we should either only remove them when they drop, and let wgpu handle their "ended" state entirely, or remove them when they end and then not do anything further when they drop(but I guess it means they leak if not ended). But doing a bit of both seems confusing. Same thing with the `valid` flag. If needed, let's document clearly, otherwise let's just rely on the fact that wgpu needs keep track of that internally anyway to implement the relevant device timeline steps.

sagudev:
> In general I find it sometimes confusing that the `wgpu_thread` is keeping track of some state(including by removing objects from maps), or reporting errors, and relying on wgpu for other states and error. If we can remove some of this bookkeeping, or document clearly why this is necessary on top of what wgpu already does, it could clarify things.

Ideally, this would all be done in wgpu, but currently wgpu does not match the specs (yet) about invalidation on error. So that's why we implement it ourself, but eventually this will be implemented in wgpu and we can remove most of it. I will add comments about why it's needed.
 
> For example, by removing passes when they end, we sort of keep track of their ended state, which interacts with the need to also remove them when they drop. So we should either only remove them when they drop, and let wgpu handle their "ended" state entirely, or remove them when they end and then not do anything further when they drop(but I guess it means they leak if not ended). But doing a bit of both seems confusing.

We need to drop the passes when they end, to allow destruction of resources they hold. But if the double meaning presence in hashmap is that confusing, we can maybe wrap passes with Option (`HashMap<RenderPassId, Option<(RenderPass, valid)>>`). So if it's present in hash it's not dropped and if value inside hashmap is Some the pass has not end. How does that sounds? (Although instead of using option for inner it might be better to use custom enum as hashmap methods also return option of existence).

> Same thing with the `valid` flag. If needed, let's document clearly, otherwise let's just rely on the fact that wgpu needs keep track of that internally anyway to implement the relevant device timeline steps.

Currently it's needed. I will document more why.

components/webgpu/wgpu_thread.rs
sagudev: `&` is de sugared to `valid = valid & do_render_command` where `valid` and `do_render_command` are booleans, so `&` is and operator (sets valid to true iff both valid and do_render_command are true).

components/webgpu/wgpu_thread.rs
sagudev: Actually it can, because wgpu does not store vaildness of passes. This is done the way it is, because we want to do steps 1-4 and only if we reached far enough (end returns ok), we actually check valid (step 5 in [spec](https://www.w3.org/TR/webgpu/#dom-gpurenderpassencoder-end)).

components/script/dom/gpurenderpassencoder.rs
sagudev: This is great idea!

components/webgpu/dom_messages.rs
sagudev: This isn't meant as a doccomment on BeginComputePass, but as a marker for section (next few enum variants are for compute pass).

components/webgpu/wgpu_thread.rs
sagudev: I am not sure I understand, this is in `BeginRenderPass`.

components/webgpu/wgpu_thread.rs
sagudev: Like for compute pass, we need to drop the pass when it ends.

sagudev:
All logic/reasoning is nearly identical to https://github.com/servo/servo/pull/32575.

components/webgpu/wgpu_thread.rs
gterzian: I mean: how can we make an assertion here, unless the corresponding method in `script` disallows this behavior.

components/webgpu/wgpu_thread.rs
gterzian: Ok

components/webgpu/wgpu_thread.rs
gterzian: Ok, would be good to document then.

components/webgpu/wgpu_thread.rs
```diff
@@ -86,8 +89,11 @@ pub(crate) struct WGPU {
     /// Provides access to poller thread
     poller: Poller,
     /// Store compute passes (that have not ended yet) and their validity
+    // we need to store valid field, because wgpu does not invalidate pass on error
     compute_passes: HashMap<ComputePassId, (Box<dyn DynComputePass>, bool)>,
-    //render_passes: HashMap<RenderPassId, Box<dyn DynRenderPass>>,
+    /// Store compute passes (that have not ended yet) and their validity
+    // we need to store valid field, because wgpu does not invalidate pass on error
+    render_passes: HashMap<RenderPassId, (Box<dyn DynRenderPass>, bool)>,
```
gterzian, line 96: > use custom enum

That sounds better indeed, also that way `valid` can be a field of only the variant that has not ended?

gterzian (changes requested):
Ok sounds good, let's just add some docs then and the enum sounds good to.

components/webgpu/wgpu_thread.rs
sagudev: I just followed here what we done for compute pass. In script the caller (same method) creates new id and then sends ipc message.

components/webgpu/wgpu_thread.rs
sagudev: Yes

components/webgpu/wgpu_thread.rs
gterzian: Ah ok so it would always be a new id.

components/webgpu/dom_messages.rs
```diff
@@ -291,9 +295,24 @@ pub enum WebGPURequest {
         device_id: id::DeviceId,
         command_encoder_id: id::CommandEncoderId,
     },
+    // Render Pass
```
gterzian, line 298: I think these should be tripe slash

gterzian, line 298: I think these should be tripe slash

components/webgpu/wgpu_thread.rs
```diff
@@ -851,10 +900,14 @@ impl WGPU {
...
+                        let pass = self
+                            .compute_passes
+                            .get_mut(&compute_pass_id)
+                            .expect("ComputePass should exists");
                         // TODO: Command encoder state error
-                        if let Some((mut pass, valid)) =
-                            self.compute_passes.remove(&compute_pass_id)
-                        {
+                        if let Pass::Open { mut pass, valid } = pass.take() {
+                            // We want to do steps 1-4 and only if we reached far enough (end returns ok)
```
gterzian, line 909: I usually write those like `end` to make clear its a reference to code.

also I think it would be clearer to write: "`pass.end` does step 1-4, and if it returns ok we check the validity of the pass at step 5". with a link to the spec on the first line above.

components/webgpu/wgpu_thread.rs
```diff
@@ -851,10 +900,14 @@ impl WGPU {
...
+                            // we actually check valid (step 5 in per spec: https://www.w3.org/TR/2024/WD-webgpu-20240703/#dom-gpurenderpassencoder-end).
```
gterzian, line 910: do we

components/webgpu/wgpu_thread.rs
```diff
@@ -868,21 +921,80 @@ impl WGPU {
...
-                            let global = &self.global;
-                            let result = gfx_select!(command_encoder_id => global.render_pass_end(&render_pass));
-                            self.maybe_dispatch_wgpu_error(device_id, result.err())
+                        let pass = self
+                            .render_passes
+                            .get_mut(&render_pass_id)
+                            .expect("RenderPass should exists");
+                        // TODO: Command encoder state error
+                        if let Pass::Open { mut pass, valid } = pass.take() {
+                            // We want to do steps 1-4 and only if we reached far enough (end returns ok)
```
gterzian, line 984: same here.

gterzian (approved):
LGTM with some nits.

components/webgpu/dom_messages.rs
sagudev: Same reasoning as https://github.com/servo/servo/pull/32665#discussion_r1665126864, as this comment describes section of enum not only one variant

sagudev:
This needs to be rebased after https://github.com/servo/servo/pull/32688 lands.

PR #32673, opened 2024-07-03
fonts: Add support for generic font families and font size configuration

### LINKED ISSUE #8371: Handle generic font-family names

### LINKED ISSUE #14773: Make per-document initial computed values actually work in servo
In https://bugzilla.mozilla.org/show_bug.cgi?id=1298588 I'm doing some of the work, and making them work with stylo, but there are some outstanding servo issues:

1. build_shared_layout_context needs to get the per-document structs from somewhere.  Which means they need to live somewhere.
2. All the media query bits.

These may not be urgent because it's possible that in servo right now the initial computed values can't actually change.

mrobinson:
This adds support for generic font families in Servo and allows for
configuration of them as well as their default font sizes. One
interesting fix here is that now monospace default to 13px, like it does
in other browsers.

In addition to that, this exposes a new interface in Stylo which allows
setting a default style. This is quite useful for fonts, but also for
other kinds of default style settings -- like text zoom.

Fixes #8371.
Fixes #14773.

---

- [ ] There are tests for these changes

### DIFF
--- a/components/fonts/platform/freetype/ohos/font_list.rs
+++ b/components/fonts/platform/freetype/ohos/font_list.rs
@@ -10,12 +10,15 @@ use base::text::{is_cjk, UnicodeBlock, UnicodeBlockMethod};
 use log::warn;
 use malloc_size_of_derive::MallocSizeOf;
 use serde::{Deserialize, Serialize};
+use style::values::computed::font::GenericFontFamily;
 use style::values::computed::{
     FontStretch as StyleFontStretch, FontStyle as StyleFontStyle, FontWeight as StyleFontWeight,
 };
 use style::Atom;
 
-use crate::{FallbackFontSelectionOptions, FontTemplate, FontTemplateDescriptor};
+use crate::{
+    FallbackFontSelectionOptions, FontTemplate, FontTemplateDescriptor, LowercaseFontFamilyName,
+};
 
 lazy_static::lazy_static! {
     static ref FONT_LIST: FontList = FontList::new();
@@ -181,16 +184,6 @@ where
     }
 }
 
-pub fn system_default_family(generic_name: &str) -> Option<String> {
-    if let Some(family) = FONT_LIST.find_family(&generic_name) {
-        Some(family.name.clone())
-    } else if let Some(alias) = FONT_LIST.find_alias(&generic_name) {
-        Some(alias.from.clone())
-    } else {
-        FONT_LIST.families.get(0).map(|family| family.name.clone())
-    }
-}
-
 // Based on fonts present in OpenHarmony.
 pub fn fallback_font_families(options: FallbackFontSelectionOptions) -> Vec<&'static str> {
     let mut families = vec![];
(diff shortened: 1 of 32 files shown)


components/fonts/platform/freetype/ohos/font_list.rs
```diff
@@ -238,4 +228,6 @@ pub fn fallback_font_families(options: FallbackFontSelectionOptions) -> Vec<&'st
     families
 }
 
-pub static SANS_SERIF_FONT_FAMILY: &'static str = "HarmonyOS Sans";
+pub fn default_system_generic_font_family(_generic: GenericFontFamily) -> LowercaseFontFamilyName {
+    "Harmony OS Sans".into()
```
jschwe, line 232:
```suggestion
    "HarmonyOS Sans".into()
```

Adding the space before the `OS` was probably not intentional, was it?

mukilan:
I reviewed the remaining stable results and these ones seem like genuine PASS to me:

- /css/css-fonts/generic-family-keywords-001.html
- /css/css-fonts/parsing/font-family-valid.html
- /css/css-fonts/system-ui.html

I've pushed a new commit to update the expectations for the above three:

These ones seem to unrelated to our change:
- /css/css-transforms/transform-input-017.html
- /html/rendering/widgets/input-checkbox-switch-rtl.tentative.html

and these are failing for me locally but they seem to have passed in the CI. I'm not sure if these are bad tests or just passing on CI due to the fonts installed. Should we investigate them further? 

- PASS [expected FAIL] /css/css-fonts/system-ui-ar.html
- PASS [expected FAIL] /css/css-fonts/system-ui-mixed.html
- PASS [expected FAIL] /css/css-fonts/system-ui-ur.html

Let me know if I should just update the expectations for these as well.

mrobinson:
> These ones seem to unrelated to our change:
>     * /css/css-transforms/transform-input-017.html
>     * /html/rendering/widgets/input-checkbox-switch-rtl.tentative.html

This test uses checkboxes, which according to `servo.css` have a monospace font set in the default style. Since our changes modifies the default font size of monospace, I suspect these are genuine passes.

> and these are failing for me locally but they seem to have passed in the CI. I'm not sure if these are bad tests or just passing on CI due to the fonts installed. Should we investigate them further?
>     * PASS [expected FAIL] /css/css-fonts/system-ui-ar.html
>     * PASS [expected FAIL] /css/css-fonts/system-ui-mixed.html
>     * PASS [expected FAIL] /css/css-fonts/system-ui-ur.html

It's quite likely this is due to having different fonts installed.

Unfortunately it looks like `/css/css-fonts/generic-family-keywords-001.html` is still flaky so we'll need to maintain the existing expectations update.

mukilan (approved):

PR #32674, opened 2024-07-03
clippy: Fix warnings in `shared` and `config`, `fonts`, `layout`, and `layout_2020` components

mrobinson:
---

- [x] These changes are part of #31575.

### DIFF
--- a/components/config/pref_util.rs
+++ b/components/config/pref_util.rs
@@ -3,7 +3,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::fmt;
 use std::str::FromStr;
 use std::sync::RwLock;
@@ -158,10 +157,7 @@ impl From<PrefValue> for [f64; 4] {
     fn from(other: PrefValue) -> [f64; 4] {
         match other {
             PrefValue::Array(values) if values.len() == 4 => {
-                let f = values
-                    .into_iter()
-                    .filter_map(|v| v.try_into().ok())
-                    .collect::<Vec<f64>>();
+                let f = values.into_iter().map(Into::into).collect::<Vec<f64>>();
                 if f.len() == 4 {
                     [f[0], f[1], f[2], f[3]]
                 } else {
--- a/components/fonts/platform/freetype/font.rs
+++ b/components/fonts/platform/freetype/font.rs
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use std::convert::TryInto;
 use std::os::raw::c_long;
 use std::sync::Arc;
 use std::{mem, ptr};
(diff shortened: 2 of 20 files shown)


components/config/pref_util.rs
```diff
@@ -160,7 +159,7 @@ impl From<PrefValue> for [f64; 4] {
             PrefValue::Array(values) if values.len() == 4 => {
                 let f = values
                     .into_iter()
-                    .filter_map(|v| v.try_into().ok())
+                    .filter_map(|v| v.into())
```
mukilan, line 162: I believe this is revealing an issue with the surrounding code we've overlooked. This module implements `From<PrefValue> to f64`  - `From` trait represents an infallible conversion, but our impl panics when the conversion is not possible. This also means the logic in this particular method to check if all the elements are convertible is unnecessary as the code either panics when encountering the first failure or it succeeds for all i.e we can use `map` here instead of `filter_map` and we'll never reach the else case.

Not sure if this needs to be handled in this patch, but I guess we'll need to revisit the design here and maybe just implement TryFrom instead of panicking in From implementations.

components/fonts/platform/freetype/font.rs
```diff
@@ -106,12 +105,11 @@ fn create_face(data: Arc<Vec<u8>>, face_index: u32) -> Result<FT_Face, &'static
         let library = FreeTypeLibraryHandle::get().lock();
 
         // This is to support 32bit Android where FT_Long is defined as i32.
```
mukilan, line 107: The comment notes that the conversion is required for Android. Perhaps, we just need a cfg(target_os = "android") here?

mukilan: Good point, we have both aarch64 and armv7 builds, so we'll need to include `target_pointer_width` as well in the `cfg`.

mrobinson: I've used `as FT_Long` which seems to both compile on 32-bit Android and also not cause any clippy warnings.

mrobinson:
@mukilan Thanks for the review. I've updated the change.

components/fonts/platform/freetype/font.rs
mukilan: I believe using `as` will just [reinterpret](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast) the bits of the `u32` value as `i32`, which means it is possible to have an index that silently gets interpreted a negative value.  According to the FreeType documentation, negative values [have special meaning](https://freetype.org/freetype2/docs/reference/ft2-face_creation.html#ft_open_face). I haven't fully grokked the semantics of that parameter, so I'm not sure how serious of an issue this is in practice. 

Clippy doesn't warn about this because `as` is safe in some cases but not in others, so it is [allow by default](https://rust-lang.github.io/rust-clippy/master/#/as_) and specific cases need to be enabled explicitly.

components/config/pref_util.rs
mukilan: I agree, the move to `TryFrom` can be delayed.

components/fonts/platform/freetype/font.rs
mrobinson: In this case the parameter is the `face_index` that we received from FontConfig. For it to be interpreted as a negative number the face index would need to be very large, correct. FontConfig should be getting this data from FreeType (converting it from i32 to u32). It seems like the only way this could happen on 32-bit platforms is if there is some kind of data corruption.

Additionally, on Android the `font_index` is always 0. See https://github.com/servo/servo/blob/93fdb8263d14346d0757c2192527bc8c7c577572/components/fonts/platform/freetype/android/font_list.rs#L33. This is the value that is ultimately passed to this function.

mukilan: Ah, I didn't realize it was hardcoded to `0` on Android. I'm fine with using `as` in this instance, but it seems like in general we should enable those clippy lints to catch such cases. Using `as` based on the implementation assumptions that could change (e.g on Android we could switch to different API for font enumeration in the future) seems like it could cause subtle issues in the long run.

mukilan (approved):

PR #32680, opened 2024-07-03
clippy: Fix a bunch of warnings in `script`

mrobinson:
This is just a portion of the errors that are remaining to be fixed.

---

- [x] These changes are part of #31575.


components/script/dom/headers.rs
```diff
@@ -535,6 +535,7 @@ pub fn extract_mime_type(headers: &HyperHeaders) -> Option<Vec<u8>> {
                 // Step 6.4
                 if temp_essence != essence {
                     charset = temp_charset.map(|c| c.to_string());
+                    temp_essence.clone_into(&mut essence);
```
gterzian, line 538: Should this not replace the below?

gterzian (approved):
Thank you! LGTM with onne question.

components/script/dom/headers.rs
mrobinson: Oh, nice catch. This was an oversight on my part which should be fixed now.

mrobinson:
@gterzian Thanks for the review!

PR #32681, opened 2024-07-03
Partially implement dialog.show()

lukewarlow:
---

### DIFF
--- a/components/script/dom/htmldialogelement.rs
+++ b/components/script/dom/htmldialogelement.rs
@@ -11,7 +11,7 @@ use crate::dom::bindings::codegen::Bindings::HTMLDialogElementBinding::HTMLDialo
 use crate::dom::bindings::inheritance::Castable;
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::str::DOMString;
-use crate::dom::document::Document;
+use crate::dom::document::{Document, FocusType};
 use crate::dom::element::Element;
 use crate::dom::eventtarget::EventTarget;
 use crate::dom::htmlelement::HTMLElement;
@@ -68,6 +68,31 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
     // https://html.spec.whatwg.org/multipage/#dom-dialog-returnvalue
     fn SetReturnValue(&self, return_value: DOMString) {
         *self.return_value.borrow_mut() = return_value;
+    }
+
+    /// <https://html.spec.whatwg.org/multipage/#dom-dialog-show>
+    fn Show(&self) {
+        let element = self.upcast::<Element>();
+
+        // Step 1 TODO: Check is modal flag is false
+        if element.has_attribute(&local_name!("open")) {
+            return;
+        }
+
+        // TODO: Step 2 If this has an open attribute, then throw an "InvalidStateError" DOMException.
+
+        // Step 3
+        element.set_bool_attribute(&local_name!("open"), true);
+
+        // TODO: Step 4 Set this's previously focused element to the focused element.
+
+        // TODO: Step 5 Let hideUntil be the result of running topmost popover ancestor given this, null, and false.
+
+        // TODO: Step 6 If hideUntil is null, then set hideUntil to this's node document.
+
+        // TODO: Step 7 Run hide all popovers until given hideUntil, false, and true.
+
+        // TODO(Issue #32702): Step 8 Run the dialog focusing steps given this.
     }
 
     // https://html.spec.whatwg.org/multipage/#dom-dialog-close
--- a/tests/wpt/meta/html/semantics/interactive-elements/the-dialog-element/submit-dialog-close-event.html.ini
+++ b/tests/wpt/meta/html/semantics/interactive-elements/the-dialog-element/submit-dialog-close-event.html.ini
@@ -1,3 +1,4 @@
 [submit-dialog-close-event.html]
+  expected: TIMEOUT
   [Tests submitting a dialog on a close event triggered by a previous submission.]
-    expected: FAIL
+    expected: TIMEOUT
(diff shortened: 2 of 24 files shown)


components/script/dom/htmldialogelement.rs
```diff
@@ -70,6 +70,33 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
         *self.return_value.borrow_mut() = return_value;
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-dialog-show
+    fn Show(&self) {
```
lukewarlow, line 74: This partially implements the spec, most of the TODOs are related to popover API interactions which arent' supported yet, or the focus steps which again aren't implemented yet (also not implemented for close afaict)

components/script/dom/htmldialogelement.rs
```diff
@@ -70,6 +70,31 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
         *self.return_value.borrow_mut() = return_value;
     }
 
+    // https://html.spec.whatwg.org/multipage/#dom-dialog-show
```
gterzian, line 73: `/// <...>`

components/script/dom/htmldialogelement.rs
```diff
@@ -70,6 +70,31 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
...
+    fn Show(&self) {
+        let element = self.upcast::<Element>();
+
+        // Step 1 TODO: Check is modal flag is false
+        if element.has_attribute(&local_name!("open")) {
```
gterzian, line 78: This, without the other check, seems like Step 2 to me, which could already be implemented instead.

components/script/dom/htmldialogelement.rs
```diff
@@ -70,6 +70,31 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
...
+            return;
+        }
+
+        // TODO: Step 2 If this has an open attribute, then throw an "InvalidStateError" DOMException.
+
+        // Step 3
+        element.set_bool_attribute(&local_name!("open"), true);
```
gterzian, line 85: Is the spec wrong to say "whose value is the empty string"? Seems like it should indeed be a boolean.

components/script/dom/htmldialogelement.rs
```diff
@@ -70,6 +70,31 @@ impl HTMLDialogElementMethods for HTMLDialogElement {
...
+
+        // TODO: Step 4 Set this's previously focused element to the focused element.
+
+        // TODO: Step 5 Let hideUntil be the result of running topmost popover ancestor given this, null, and false.
+
+        // TODO: Step 6 If hideUntil is null, then set hideUntil to this's node document.
+
+        // TODO: Step 7 Run hide all popovers until given hideUntil, false, and true.
+
+        // TODO: Step 8 Run the dialog focusing steps given this.
```
gterzian, line 95: There is some mechanism at https://github.com/servo/servo/blob/650af7db926e9cc070e136f21a56bd47d566d10a/components/script/dom/document.rs#L1105 which seems to be https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps

The "dialog focusing steps" appear like layer around that mechanism, so perhaps some sort of calling on the existing focusing mechanism would make sense.

gterzian (changes requested):
Looks good overall, with a couple of comments.

components/script/dom/htmldialogelement.rs
lukewarlow: It's not, (admittedly the spec isn't super clear here) step 2 throws an exception if you try and call .show() on an already open modal dialog. Step 1 is an early return for calling .show() on an already open non-modal dialog.

components/script/dom/htmldialogelement.rs
lukewarlow: The spec isn't wrong, it's just boolean attributes are odd in the spec. A boolean attribute is just about the presence of the attribute so the value is irelevant, I think an attribute has to have a value, so for boolean ones (where you typically omit a value when writing it) they're set to an empty string.

components/script/dom/htmldialogelement.rs
lukewarlow: I don't think just calling into focusing steps directly is that useful on its own because in practice it's normally going to be the wrong behaviour.

In terms of what Servo needs it':

- The [focus delegate](https://html.spec.whatwg.org/multipage/interaction.html#focus-delegate) concept
- autofocus in general seems missing too and that plays into the dialog algorithm.

Happy to make an issue and link to that from the TODO comment?

gterzian (approved):
LGTM. Thanks for filing the follow-up issue.

mrobinson:
This change seems to cause `/html/semantics/interactive-elements/the-dialog-element/submit-dialog-close-event.html` to start timing out.

tests/wpt/meta/html/semantics/interactive-elements/the-dialog-element/submit-dialog-close-event.html.ini
lukewarlow, line 1: This is the correct file right? @mrobinson

mrobinson: Yes, hrm, that seems right. Apologies that I didn't see that before. I'm a bit baffled as to why this test seems to have failed when trying to land this: https://github.com/servo/servo/runs/27162484996.

I will try again, just to verify that things are working properly.

mrobinson: It failed again. Perhaps the issue is that you added the `TIMEOUT` expectation for a subtest, but now the entire test is timing out so the `TIMEOUT` should be at the top level of the `ini` file?

lukewarlow: I've fixed that now so hopefully it'll work this time

mrobinson:
I ran

```shell
$ ./mach test-wpt -r /html/semantics/interactive-elements/the-dialog-element/submit-dialog-close-event.html --headless --log-raw out.log
$ ./mach update-wpt out.log
```

and also fixed a small compiler warning.

PR #32683, opened 2024-07-03
servoshell: Update all egui dependencies and group them for dependabot

mrobinson:
This updates egui, making the necessary changes to maintain behavior. In
addition, it groups dependencies so that they are updated in a single PR
by dependabot, avoiding manual work.

---

- [x] These changes do not require tests because this is just a dependency upgrade.

### DIFF
--- a/ports/servoshell/desktop/minibrowser.rs
+++ b/ports/servoshell/desktop/minibrowser.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use egui::{
-    pos2, CentralPanel, Color32, Frame, Key, Modifiers, PaintCallback, Pos2, Spinner,
+    pos2, CentralPanel, Color32, Frame, Key, Label, Modifiers, PaintCallback, Pos2, Spinner,
     TopBottomPanel, Vec2,
 };
 use egui_glow::CallbackFn;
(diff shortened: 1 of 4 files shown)


ports/servoshell/desktop/minibrowser.rs
```diff
@@ -282,6 +270,17 @@ impl Minibrowser {
                     let Some(servo_fbo) = servo_framebuffer_id else {
                         return;
                     };
+
+                    if let Some(status_text) = &self.status_text {
+                        egui::containers::popup::show_tooltip_at(
+                            ctx,
+                            LayerId::new(Order::Tooltip, "tooltip layer".into()),
```
mukilan, line 277: I skimmed through the egui source and I don't think we need to create the layer explicitly here. Instead we can just[ pass the parent UI's ID](https://github.com/emilk/egui/blob/8bf498d2bf849892ec0c3fa49fea4c3c504a06fe/crates/egui/src/containers/popup.rs#L43) using
```suggestion
                            ui.layer_id(),
```

egui internally creates the tooltip with the Order::Tooltip when painting the tooltip within the parent layer.

ports/servoshell/desktop/minibrowser.rs
```diff
@@ -282,6 +270,17 @@ impl Minibrowser {
...
+                            ui.id(),
```
mukilan, line 278: I believe this is passing the parent's ID as the ID of this widget, whereas we'll need to make a unique ID here. The original code did this using `tooltip_for_status_text.into()` which seems correct to me.

mukilan (changes requested):

mrobinson:
Thanks for the review @mukilan. I've made the changes you've requested.

mukilan (approved):

PR #32606, opened 2024-06-25
Upgrade stylo to 2024-06-24

Loirooriol:
This continues #32474

Changelog:
 - Upstream: https://github.com/servo/stylo/compare/07759020d8574d371bb92bca609c82e669d77566...8a13add71238c8cb0880b857a5db387ae47822c7
 - Servo fixups: https://github.com/servo/stylo/compare/f2fbf5520e9fe1f45d0fe868c3f8e2d100b6d3a8...fafff4ca2cc971acc5278977ee69d6351cfbbaca

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] There are tests for these changes

PR #32686, opened 2024-07-04
Enable grid styles, aspect ratio, gap, and use gecko's alignment representation

nicoburns:
Updates Servo to work with https://github.com/servo/stylo/pull/47

### Changes made

- Update to use new alignment style representation
- Fixes for alignment due to a fixup not being applied in stylo anymore (layout is the correct place to do this)
   - I have also fixed up layout_2013 to avoid a large number of test regressions
- Stub handling of `display: grid` to allow Servo to compile. For now we just treat it as `display: flex`, which should be acceptable (temporarily) because this code path is gated behind a `layout.grid.enabled` pref.

### Test updates

- Almost entirely newly passing enabled by:
   - The newly enabled syles
   - The better alignment implementation from Gecko
   - Better handling of the relationship between align_items and align_self
- One newly failing css-align test, but it's a CSS Grid one. So I reckon we ought not to worry about that for now.
- Layout 2013 fails are the same as https://github.com/servo/servo/pull/32606, and thus I suspect unrelated to this PR. I can update assertions is need be.

---

- [x] There are tests for these changes (WPT tests). However these changes are largely non-functional (either updating to a new style representation or enabling future implementation of functionality rather than actually implementing)


mrobinson:
Let's close this and make it part of the next Stylo upgrade.

nicoburns:
Happy to close. Just wanted to make sure that the fixes were available for the sync so as to avoid whoever does the sync (@Loirooriol ?) having to re-implement them.

mrobinson:
I've split out the alignment changes and landed those separately. We'll just have to be careful during the next stylo rebase to not introduce any extra changes.

nicoburns:
The grid and aspect ratio changes have now also landed :) It might still be useful to add the `layout.grid.enabled` feature flag and enable the CSS Grid WPT tests but that can be a separate PR.

PR #32694, opened 2024-07-04
Avoid unnecessary clones for URLs in script_module

CYBAI:
While trying to fix the issue for #31397, I noticed that there are many unnecessary clones for `url`s in the script_module file. As `url`s could be very long, it might be better to just borrow with refs instead of always cloning them.

---

- [x] These changes do not require tests because it mainly changes clones to borrows; thus, it SHOULD NOT break any tests.

### DIFF
--- a/components/script/script_module.rs
+++ b/components/script/script_module.rs
@@ -273,14 +273,16 @@ impl ModuleTree {
         self.parent_identities.borrow_mut().insert(parent_identity);
     }
 
-    pub fn insert_incomplete_fetch_url(&self, dependency: ServoUrl) {
-        self.incomplete_fetch_urls.borrow_mut().insert(dependency);
-    }
-
-    pub fn remove_incomplete_fetch_url(&self, dependency: ServoUrl) {
+    pub fn insert_incomplete_fetch_url(&self, dependency: &ServoUrl) {
         self.incomplete_fetch_urls
             .borrow_mut()
-            .shift_remove(&dependency);
+            .insert(dependency.clone());
+    }
+
+    pub fn remove_incomplete_fetch_url(&self, dependency: &ServoUrl) {
+        self.incomplete_fetch_urls
+            .borrow_mut()
+            .shift_remove(dependency);
     }
 
     /// recursively checks if all of the transitive descendants are
@@ -421,7 +423,7 @@ impl ModuleTree {
         global: &GlobalScope,
         owner: ModuleOwner,
         module_script_text: Rc<DOMString>,
-        url: ServoUrl,
+        url: &ServoUrl,
         options: ScriptFetchOptions,
     ) -> Result<ModuleObject, RethrowError> {
         let cx = GlobalScope::get_cx();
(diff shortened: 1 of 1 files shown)


gterzian (approved):
thank you!

mrobinson (commented):
FWIW, cloning a `ServoUrl` is cheap, because it just holds an `Arc` to the actual `url::Url` inside. That said, I'm not sure there is much  a problem with using more references here.

components/script/script_module.rs
```diff
@@ -706,23 +708,23 @@ impl ModuleTree {
...
+                    &self.url,
                 ),
             }
         };
 
         match specifier_urls {
             // Step 3.
             Ok(valid_specifier_urls) if valid_specifier_urls.is_empty() => {
-                debug!("Module {} doesn't have any dependencies.", self.url.clone());
+                debug!("Module {} doesn't have any dependencies.", &self.url);
```
mrobinson, line 727: I suspect you don't need to use a reference here and can just pass `self.url` directly to `debug!()`. It likely does a borrow.

components/script/script_module.rs
```diff
@@ -795,7 +797,7 @@ impl ModuleTree {
 
         self.set_status(ModuleStatus::Finished);
 
-        debug!("Going to advance and finish for: {}", self.url.clone());
+        debug!("Going to advance and finish for: {}", &self.url);
```
mrobinson, line 800: Ditto.

components/script/script_module.rs
```diff
@@ -1129,15 +1131,15 @@ impl FetchResponseListener for ModuleContext {
...
         };
 
-        module_tree.remove_incomplete_fetch_url(self.url.clone());
+        module_tree.remove_incomplete_fetch_url(&self.url);
 
         // Step 12.
         match load {
             Err(err) => {
-                error!("Failed to fetch {} with error {:?}", self.url.clone(), err);
+                error!("Failed to fetch {} with error {:?}", &self.url, err);
```
mrobinson, line 1142: Ditto.

mrobinson (changes requested):

CYBAI:
> FWIW, cloning a `ServoUrl` is cheap, because it just holds an `Arc` to the actual `url::Url` inside.

ah, thank you for pointing this out! in this case, I think I'd like to change this PR to only removing the unnecessary `.clone()` in `dbg!` macros.

> That said, I'm not sure there is much a problem with using more references here.

Sorry, I don't fully get it. Do you mean using more references here would make the implementation messier? 👀

sagudev:
> FWIW, cloning a ServoUrl is cheap, because it just holds an Arc to the actual url::Url inside

But atomic operations aren't free either, references are even cheaper.

mrobinson:
> Sorry, I don't fully get it. Do you mean using more references here would make the implementation messier? 👀

Apologies, I think this change is fine, as @sagudev says it delays the clone of atomic data structures.

components/script/script_module.rs
CYBAI: Yes, using a reference here is redundant; thanks for catching it 🙏 I've updated it in the latest commit.

mrobinson (approved):

PR #32695, opened 2024-07-04
layout: Improve layout of table captions

mrobinson:
- Instead of treating captions as a `BlockFormattingContext`, treat it as
  a `NonReplacedFormattingContext`, which allows reusing flow layout for
  captions -- fixing some issues with sizing.
- Pass in the proper size of the containing block when laying out,
  fixing margin calculation.
- Follow the unspecified rules about how various size properties on
  captions affect their size.
- Improve linebreaking around atomics, which is tested by
  caption-related tests. This fixes intrinsic size calculation regarding
  soft wrap opportunities around atomic and also makes the code making
  these actual soft wrap opportunities a bit better.
    
---

### DIFF
--- a/components/layout_2020/flow/inline/mod.rs
+++ b/components/layout_2020/flow/inline/mod.rs
@@ -104,8 +104,12 @@ use style::values::specified::box_::BaselineSource;
 use style::values::specified::text::{TextAlignKeyword, TextDecorationLine};
 use style::values::specified::{TextAlignLast, TextJustify};
 use style::Zero;
-use text_run::{add_or_get_font, get_font_for_first_font_for_style, TextRun};
+use text_run::{
+    add_or_get_font, get_font_for_first_font_for_style, TextRun, XI_LINE_BREAKING_CLASS_GL,
+    XI_LINE_BREAKING_CLASS_WJ, XI_LINE_BREAKING_CLASS_ZWJ,
+};
 use webrender_api::FontInstanceKey;
+use xi_unicode::linebreak_property;
 
 use super::float::PlacementAmongFloats;
 use crate::cell::ArcRefCell;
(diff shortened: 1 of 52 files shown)


components/layout_2020/table/layout.rs
```diff
@@ -105,7 +104,11 @@ pub(crate) struct TableLayout<'a> {
     rows: Vec<RowLayout>,
     columns: Vec<ColumnLayout>,
     cell_measures: Vec<Vec<LogicalVec2<CellOrTrackMeasure>>>,
+    /// The calculated width of the table, including space for the grid and also for any
+    /// captions.
     table_width: Au,
+    /// The table width table width minus the total horizontal border spacing (if any). This is the
```
mukilan, line 110:
```suggestion
    /// The table width minus the total horizontal border spacing (if any). This is the
```

components/layout_2020/table/layout.rs
```diff
@@ -1503,74 +1522,36 @@ impl<'a> TableLayout<'a> {
...
-            pbm.border,
-            margin,
-            None, /* clearance */
-            CollapsedBlockMargins::zero(),
-        )
+        box_fragment
     }
 
-    /// Lay out the table (grid and captions) of this [`TableLayout`] into fragments. This should
+    /// Lay out the table (grid oand captions) of this [`TableLayout`] into fragments. This should
```
mukilan, line 1554:
```suggestion
    /// Lay out the table (grid and captions) of this [`TableLayout`] into fragments. This should
```

components/layout_2020/flow/inline/mod.rs
```diff
@@ -2344,8 +2348,21 @@ impl<'a> ContentSizesComputation<'a> {
...
+                    text_run.prevent_soft_wrap_opportunity_at_end;
             },
             InlineItem::Atomic(atomic) => {
+                let soft_wrap_opportunity_prevented = mem::replace(
+                    &mut self.prevent_soft_wrap_opportunity_before_next_atomic,
+                    false,
+                );
+
+                // TODO: need to handle TextWrapMode::Nowrap.
+                if self.previous_item_was_atomic_inline && soft_wrap_opportunity_prevented {
```
mukilan, line 2362: Is the conditional here correct? Should `soft_wrap_opportunity_prevented` be negated? Currently the logic will call `line_break_opportunity` when it is `true`. For the 2nd atomic item,` previous_item_was_atomic_inline` will be `true` but `soft_wrap_opportunity_prevented` is `false` since the 1st atomic item would have reset the flag to `false`.

mukilan (approved):

PR #32696, opened 2024-07-04
layout: Add a linebreak opportunity before atomics for preferred width calculation

mrobinson:
There should be a linebreak before atomics, as long as the previous
character did not prevent it. Note that this code does not handle NoWrap
yet -- something that's common to all of preferred width calculation
currently.

---


mrobinson:
Since this alone doesn't change any test results, I'm going to roll it into #32695.

PR #32700, opened 2024-07-04
webgpu: Divide message code into separate files

sagudev:
These are two things that bugged me from https://github.com/servo/servo/pull/32255:

- dom_messages was still to big, and grouping them together in ipc-messages folder just makes sense
- I did wgc, wgt renames wrong, now they are done right in Cargo.toml so no more "wgpu-core" anywhere (that's the way it's done in firefox and [even in wgpu](https://github.com/gfx-rs/wgpu/blob/9f34acd5671987558038ba0abeb24c69220bd581/Cargo.toml#L54)).

Change is reviewable per commit

---

- [x] These changes do not require tests because rust offers fearless refactoring 💪🏾.


Cargo.toml
```diff
@@ -136,8 +136,8 @@ webpki-roots = "0.25"
 webrender = { git = "https://github.com/servo/webrender", branch = "0.64", features = ["capture"] }
 webrender_api = { git = "https://github.com/servo/webrender", branch = "0.64" }
 webrender_traits = { path = "components/shared/webrender" }
-wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
-wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
+wgc = { package = "wgpu-core", git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
+wgt = { package = "wgpu-types", git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
```
mrobinson, line 140: I think this rename actually makes it harder to understand the dependencies. Let's just stick with the real package names.

mrobinson (changes requested):

Cargo.toml
sagudev: The main reason for the rename is that code bases already uses wgc and wgt names (ex: [firefox](https://searchfox.org/mozilla-central/rev/e4a222169f9f9b1d8f2f49e4066585b6b9c4dc58/gfx/wgpu_bindings/Cargo.toml#17), [wgpu](https://github.com/gfx-rs/wgpu/blob/9f34acd5671987558038ba0abeb24c69220bd581/Cargo.toml#L54), [wgpu-native](https://github.com/gfx-rs/wgpu-native/blob/a1ede115d0d669b88b34447e58752335cb882e61/Cargo.toml#L24) and servo a little) and that's way it makes sense to follow it (given that hacking servo's webgpu usually requires looking at wgpu/firefox code) also it's less writing and we avoid dealing with `-_` problem.

mrobinson: This is a good argument, but let's stick with the full names and only rename when there is a naming conflict. We have a lot more Rust code than Firefox does with a *lot* of dependencies (a number that would be good to reduce). Any kind of obfuscation makes it harder to folks to understand what is going on when they are reading the code for the first time.

sagudev: > We have a lot more Rust code than Firefox does with a _lot_ of dependencies (a number that would be good to reduce). Any kind of obfuscation makes it harder to folks to understand what is going on when they are reading the code for the first time.

wgpu-types and wgpu-core are only used by webgpu crate that has only few direct dependencies, although they are pub reexport via wgt, wgc names. I will remove this from this PR, but I would still like to push it forward in the future (will create issue or maybe zulip channel for more wide discussion). Coincidentally, this remands me of [Wadler's Law](https://wiki.haskell.org/Wadler's_Law).

Ironically, the rename of `mozjs` to `js` bugs me (due to using `mozjs` in other projects), so similar story with reversed roles.

mrobinson: In general, using abbreviations for the names of things leads to a very tiny benefit for the person writing the abbreviation, but has a continual cost for everyone trying to understand and implement bug fixes in the code in the future. I don't think it's worth it.  I typically use the rule of thumb established by the [Google C++ guidelines](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules):

> Use names that describe the purpose or intent of the object. Do not worry about saving horizontal space as it is far more important to make your code immediately understandable by a new reader. Minimize the use of abbreviations that would likely be unknown to someone outside your project (especially acronyms and initialisms). Do not abbreviate by deleting letters within a word. As a rule of thumb, an abbreviation is probably OK if it's listed in Wikipedia. Generally speaking, descriptiveness should be proportional to the name's scope of visibility. For example, n may be a fine name within a 5-line function, but within the scope of a class, it's likely too vague.

This is particularly important in a web engine, because the codebase is so big.

components/webgpu/ipc_messages/send_dom.rs
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+//! IPC massages that are send to GPU DOM objects.
```
mrobinson, line 5: Are the DOM objects receiving these messages directly or are they somehow forwarded through the script thread?

If this is the later case:

```suggestion
//! IPC messages that are sent to the script thread and forwarded to WebGPU DOM objects.
```

Or the former:

```suggestion
//! IPC messages that are sent to the receivers on WebGPU DOM objects.
```

components/webgpu/ipc_messages/send_dom.rs
```diff
@@ -0,0 +1,77 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mrobinson, line 1: How about "to_script.rs" for this filename?

components/webgpu/ipc_messages/recv.rs
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
```
mrobinson, line 2: How about "from_script.rs" for this file name. I think it's useful to think about messages as either going to some place or coming from it.

components/webgpu/ipc_messages/recv.rs
```diff
@@ -2,7 +2,7 @@
...
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-//! IPC massages that are send/received from GPU DOM objects.
+//! IPC massages that are received from GPU DOM objects.
```
mrobinson, line 5:
```suggestion
//! IPC messages that are received from the script thread.
```

components/webgpu/lib.rs
```diff
@@ -5,7 +5,8 @@
 use log::warn;
 use webrender::RenderApiSender;
 use wgpu_thread::WGPU;
-pub use {wgpu_core as wgc, wgpu_types as wgt};
+// TODO: Minimize public reexport of wgc
+pub use {wgc, wgt};
```
mrobinson, line 9:
```suggestion
pub use wgu_core;
pub use wgpu_types;
```

mrobinson (changes requested):
This seems like a good start to a cleanup. I have a few nits below plus I think we should definitely not use the `wgc` / `wgt` names as they save 6 characters each and make the code a lot harder to understand.

components/webgpu/ipc_messages/send_dom.rs
sagudev: I think technically this is script thread (or maybe subthread of script thread, page scope; I am not so really confident in my understanding of threading in script), but what I want to say is that they are revived using https://github.com/servo/servo/blob/f29dd64a7b633e844756e6eecf9e05e0b327fc51/components/script/dom/gpu.rs#L47 which is impl on Dom objects that receives it (underneath this works by creating dom task source ...)

Then there are messages that are actually send to script_thread: https://github.com/servo/servo/blob/f29dd64a7b633e844756e6eecf9e05e0b327fc51/components/script/script_thread.rs#L2422

components/webgpu/ipc_messages/recv.rs
sagudev: The problem is that wgpu thread receives messages from multiple threads (constellation, script_thread, dom?), so the most appropriate description is receiving (in wgpu thread).

mrobinson (approved):

PR #32703, opened 2024-07-05
Implement Debug and Clone for fragment tree

nicoburns:
Derive `Debug` and `Clone` for the layout_2020 fragment tree.

- These changes are part of https://github.com/servo/servo/pull/32619 but they also work standalone so I've split them out to try to reduce the potential for conflicts in 32619 which is likely to be a long-running PR.

---

- [x] These changes do not require tests because they're just adding some derived trait impls


mrobinson:
Why does taffy need to clone the fragment tree?

nicoburns:
> Why does taffy need to clone the fragment tree?

Taffy itself doesn't. The integration of Taffy with Servo uses it (or will use it) to implement caching (incremental layout): we can skip running the actual layout algorithm for a subtree if we know that:

- The subtree hasn't been mutated since we last performed layout on this subtree
- The input constraints are the same as input constraints we have already computed a layout for.

In this case the fragment subtree from the previous layout run can be cloned which is cheaper than re-computing it from scratch. There is much more that could be done to optimise this:

1. We could have a "measure" method that doesn't produce a fragment tree at all for the case where we know a re-layout will be required (or extend the "intrinsic_inline_sizes" method to work for block sizes) (or a parameter to the "layout" method which causes it to not to produce fragments).
2. We could use `Arc` or some other mechanism (arena allocating the fragment tree?) to make cloning cheap or otherwise make the fragment tree copy-on-write.
3. (maybe?) We could somehow give layout access to the existing fragment tree such that it can directly return parts of it.

At least 1 will almost certainly need to implemented before any Taffy integration is merged, because performance will likely remain unacceptably bad until it is.

nicoburns:
Btw, part of the motivation for splitting this out into it's own PR was that another commit to `main` added one of the derives I had added and this caused a rebase conflict. However, now I've simplified the history of https://github.com/servo/servo/pull/32619 this is probably more manageable so if you'd prefer this change to land as part of that PR (if it is still necessary if/when that lands) then I can close this and go with that approach.

mrobinson:
I think the solution here isn't to allow clone the fragment tree, but to put the cached fragment in an `ArcRefCell`, which can be cloned. Internally, fragments hold their children as `ArcRefCell` so this means that you are cloning the same type that would be cloned internally. This should be faster as well, as you won't have to clone all the members of the `Fragment` you are holding. I guess the addition of `Debug` here is unrelated -- but I also wonder about how useful it is as fragments have many members and the output is likely to flood the screen.

nicoburns:
> Internally, fragments hold their children as `ArcRefCell` so this means that you are cloning the same type that would be cloned internally

Ah, so they do. I missed this because the constructors take a plain `Vec<Fragment>` and wrap them in `ArcRefCell` themselves. Perhaps I could modify the constructors (or add a new one) to take `Vec<ArcRefCell<Fragment>>` directly? That would indeed be much cheaper than deep cloning trees.

mrobinson:
> > Internally, fragments hold their children as `ArcRefCell` so this means that you are cloning the same type that would be cloned internally
> 
> Ah, so they do. I missed this because the constructors take a plain `Vec<Fragment>` and wrap them in `ArcRefCell` themselves. Perhaps I could modify the constructors (or add a new one) to take `Vec<ArcRefCell<Fragment>>` directly? That would indeed be much cheaper than deep cloning trees.

Are you trying to modify an existing `Fragment` to change its children? Taking the raw `Fragment`s as the argument to the `BoxFragment` constructor is useful because it can access the children without having to do an atomic operation and also callers don't have to construct `ArcRefCell`s for each child.

nicoburns:
> Are you trying to modify an existing `Fragment` to change its children? Taking the raw `Fragment`s as the argument to the `BoxFragment` constructor is useful because it can access the children without having to do an atomic operation and also callers don't have to construct `ArcRefCell`s for each child.

No, I'm creating new `Fragment` but reusing pre-computed children from an old fragment (where that might have been initially computed either earlier in the current layout run or in an earlier run). This means layout can be entirely skipped for unchanged subtrees (which is good for performance everywhere, but is esp. important for Flexbox and Grid as their algorithms cause subtrees to be revisited many times even within a single layout run).

I could probably also cache the `scrollable_overflow_from_children` if we're worried about the cost of computing that. I'm not sure if this would work everywhere the fragment tree is used, but we could also potentially consider switching from `Vec<ArcRefCell<Fragment>>` to `ArcRefCell<Vec<Fragment>>`

I wonder if anything actually mutates these. If not then we could potentially switch to just `Arc` rather than `ArcRefCell`.

mrobinson:
> No, I'm creating new `Fragment` but reusing pre-computed children from an old fragment (where that might have been initially computed either earlier in the current layout run or in an earlier run). This means layout can be entirely skipped for unchanged subtrees (which is good for performance everywhere, but is esp. important for Flexbox and Grid as their algorithms cause subtrees to be revisited many times even within a single layout run).

Is it possible to delay the creation of the parent `Fragment` until you have a final layout for it? In some sense, the creation of a `Fragment` is the "final" step and leans toward the immutability of the fragment tree. This could definitely change, but the current design makes this assumption, I think.

nicoburns:
> Is it possible to delay the creation of the parent `Fragment` until you have a final layout for it? In some sense, the creation of a `Fragment` is the "final" step and leans toward the immutability of the fragment tree. This could definitely change, but the current design makes this assumption, I think.

I am indeed hoping to skip creating `Fragment`s entirely for subtrees that are revisited within a single layout run. This is what I have been talking about with a separate "measure" method (or a parameter to the layout method) which essentially runs (enough of) the layout algorithm to compute the size of the node under a given set of constraints, but skips anything else (including fragment generation).

But for reusing fragments across layout runs, I think some kind of cloning is necessary, as I believe a new `Fragment` tree is created each time. So we need to be able to return fragments that have already been "used" in a previous run (but we do *not* need to be able to mutate created fragments, as we'd be creating a new fragment to hold ones being cloned). I think `Arc` would be perfect for this as it would allow for cheap, shallow clones. And would also allow for the memory to be shared keeping the memory cost of the caching down.

PR #32705, opened 2024-07-05
Implement DOM behavior relating to the autocomplete attribute

### LINKED ISSUE #25383: Implement form autocomplete, or at least the DOM behavior relating to autocomplete hints?
WHATWG describes autocompletion in terms of "hints" that user agents "sometimes have features for". I don't know whether this would be something that belongs in the main core of servo, or if it's something that servo should offer the embedder a chance to implement.

The bare minimum for technical compliance would be to parse the hints, reflect them in the DOM appropriately, and not actually autocomplete anything; WPT html/semantics/forms/the-form-element/form-autocomplete tests for this minimum.

The autocomplete IDL attribute is commented out at the IDL level in Servo's HTMLInputElement and HTMLTextAreaElement. https://html.spec.whatwg.org/#autofilling-form-controls:-the-autocomplete-attribute describes the canonical syntax of the content attribute, which the IDL attribute would reflect as a string.

alex-pricinoc:
This basic implementation only parses the autocomplete attribute according to the spec and does not autocomplete anything as described in #25383.

There are some TODOs for the missing functionality relating to the hint and scope tokens which are not used.

I've looked at how the AutofillData could be cached but I'm not sure how to detect when an element's form owner autocomplete attribute changed.

The main test for this is tests/wpt/meta/html/semantics/forms/the-form-element/form-autocomplete.html.ini but there are a few others which were affected in tests/wpt/meta/html/dom/ directory.

Let me know what else can be improved.

Thanks!

---

- [X] These changes fix #25383


components/script/dom/autofill.rs
```diff
@@ -0,0 +1,356 @@
...
+
+#[derive(Debug, PartialEq)]
+pub enum AutofillMantle {
+    Anchor,
+    Expectation,
+}
+
+#[allow(unused)]
+#[derive(Debug)]
+pub struct AutofillData {
```
gterzian, line 99: I think we should store this on the elements themselves via a `DomRefCell<Option>>`, and re-run the algorithm when necessary via the [`attribute_mutated`](https://github.com/servo/servo/blob/b243457ccc6cd5a2dab58d9c9ff8b6fee1db6a20/components/script/dom/htmlinputelement.rs#L2258) hook.

components/script/dom/htmlinputelement.rs
```diff
@@ -487,6 +488,27 @@ impl HTMLInputElement {
         )
     }
 
+    fn does_autocomplete_apply(&self) -> bool {
```
gterzian, line 491: Do you have a spec link for this?

components/script/dom/htmlformelement.rs
```diff
@@ -1689,6 +1690,19 @@ pub trait FormControl: DomObject {
         }
     }
 
+    // https://html.spec.whatwg.org/multipage/#autofill-expectation-mantle
+    fn autofill_mantle(&self) -> AutofillMantle {
```
gterzian, line 1694: I think it would be clearer if this was a function in `autofill` taking the element as argument. Also I think it could return a boolean and be renamed `is_anchor`, so that the enum can instead be used to store data.

components/script/dom/autofill.rs
```diff
@@ -0,0 +1,356 @@
...
+use style::str::split_html_space_chars;
+
+use self::AutofillCategory::*;
+use crate::dom::bindings::import::module::Castable;
+use crate::dom::bindings::str::DOMString;
+use crate::dom::element::Element;
+use crate::dom::htmlformelement::FormControl;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AutofillCategory {
```
gterzian, line 16: Spec links for all data structure please.

gterzian (changes requested):
Thanks for the intial work! Haven't reviewed everything in detail, but here are some intial comments. 


Also note https://html.spec.whatwg.org/multipage/browsing-the-web.html#history-autocomplete, which loosely matches https://github.com/servo/servo/blob/b243457ccc6cd5a2dab58d9c9ff8b6fee1db6a20/components/script/dom/document.rs#L680, we could implement this, or at least add a note and todo.

jdm:
No response from the original author.

PR #32706, opened 2024-07-05
script: Impl cloning of JSPrincipals

### LINKED ISSUE #32648: Investigate ErrorStackFrames in clones
According to MDN stack field is non-standard: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack, if this is left in all error clones are failing, but after some investigation I found out that fixing this in servo "the right way" would be a lot harder as some things are missing, so I left this out for future work.

_Originally posted by @sagudev in https://github.com/servo/mozjs/pull/408#discussion_r1330341613_
            

We want to remove https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0027-No-ErrorStackFrames-in-clones.patch eventually.

sagudev:
We were missing read/write hooks for JSPrincipals. Relevant file in mozilla: https://searchfox.org/mozilla-central/source/caps/nsJSPrincipals.cpp, although they have multiple Principals (we only have one), and they manually (de)serialize it: https://searchfox.org/mozilla-central/rev/956f21594f3178f75c3a1c278980e112a65eb19d/caps/nsJSPrincipals.cpp#112 while we are smart and use bincode for this (we already depend on it due to ipc-channel and wr).

Try run: https://github.com/sagudev/servo/actions/runs/9810258424

---

- [x] These changes fix #32648, so we can remove one patch in mozjs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -185,3 +185,6 @@ strip = true
 #
 # [patch."https://github.com/servo/<repository>"]
 # <crate> = { path = "/path/to/local/checkout" }
+
+[patch."https://github.com/servo/mozjs"]
+mozjs = { git = "https://github.com/sagudev/mozjs", branch = "no-errorstackclone-patch" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,6 +186,3 @@ strip = true
 #
 # [patch."https://github.com/servo/<repository>"]
 # <crate> = { path = "/path/to/local/checkout" }
-
-[patch."https://github.com/servo/mozjs"]
-mozjs = { git = "https://github.com/sagudev/mozjs", branch = "no-errorstackclone-patch" }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -179,3 +179,6 @@ debug-assertions = false
 #
 # [patch."https://github.com/servo/<repository>"]
 # <crate> = { path = "/path/to/local/checkout" }
+
+[patch."https://github.com/servo/mozjs"]
+mozjs = { git = "https://github.com/sagudev/mozjs", branch = "no-errorstackclone-patch" }
```
sagudev, line 184: We need to land https://github.com/servo/mozjs/pull/473 in order to remove this.

components/script/Cargo.toml
```diff
@@ -34,6 +34,7 @@ background_hang_monitor_api = { workspace = true }
 backtrace = { workspace = true }
 base = { workspace = true }
 base64 = { workspace = true }
+bincode = "1"
```
jdm, line 37: Nit: please make this a workspace dependency.

jdm (approved):
Nice! This looks as clean as can be expected from the level of the engine.

sagudev:
rebased!

PR #32716, opened 2024-07-06
 Replace null-byte terminated string literals with C-string literals

### LINKED ISSUE #32628: Replace null-byte terminated string literals with C-string literals
They are available from rust 1.77: https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#c-string-literals, most occurrences should be found by searching for `\0` in source tree.

This is generally just mechanical work, but it mostly touches unsafe code.

Does it sounds reasonable @mrobinson?

abumsoo:
I've replaced all null-byte terminated string literals that I could find  by looking for anything in the form of `b"...\0"`. This means that I did not replace strings like `b'\0'`. Reasoning being that I thought `c""` would be more confusing than `b'\0'`, but please let me know if that doesn't seem right.

Most of the byte strings that are being replaced are arguments to mozjs functions which are specifically expecting  `*const c_char` which is why I kept or added the `.as_ptr()` method to the c-strings. 

Will the mozjs functions be getting updated so that they expect `CStrings` instead?

The only byte string that I left untouched was the following:
https://github.com/servo/servo/blob/59d0f1fe1aec4bec736bf2839e43de886eaebf32/components/script/script_runtime.rs#L828-L829
because the compiler complained that `count_bytes()` was too new when I tried to replace `servo_id.len()` with `servo_id.count_bytes()` and I couldn't find a way around that.

---

- [X] These changes fix #32628 

- [X] These changes do not require tests because the type that is being used in the end even after converting the byte strings to c-strings remains the same, so there should not be any real functional changes

### DIFF
--- a/components/script/dom/bindings/interface.rs
+++ b/components/script/dom/bindings/interface.rs
@@ -69,7 +69,7 @@ impl NonCallbackInterfaceObjectClass {
     ) -> NonCallbackInterfaceObjectClass {
         NonCallbackInterfaceObjectClass {
             class: JSClass {
-                name: b"Function\0" as *const _ as *const libc::c_char,
+                name: c"Function".as_ptr(),
                 flags: 0,
                 cOps: &constructor_behavior.0,
                 spec: 0 as *const _,
@@ -344,7 +344,7 @@ pub fn create_named_constructors(
             assert!(JS_DefineProperty3(
                 *cx,
                 constructor.handle(),
-                b"prototype\0".as_ptr() as *const libc::c_char,
+                c"prototype".as_ptr(),
                 interface_prototype_object,
                 (JSPROP_PERMANENT | JSPROP_READONLY) as u32
             ));
@@ -504,7 +504,7 @@ fn define_name(cx: SafeJSContext, obj: HandleObject, name: &[u8]) {
         assert!(JS_DefineProperty4(
             *cx,
             obj,
-            b"name\0".as_ptr() as *const libc::c_char,
+            c"name".as_ptr(),
             name.handle(),
             JSPROP_READONLY as u32
         ));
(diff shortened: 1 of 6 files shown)


sagudev:
> I did not replace strings like `b'\0'`. Reasoning being that I thought `c""` would be more confusing than `b'\0'`, but please let me know if that doesn't seem right.

I agree with your reasoning here.

> Will the mozjs functions be getting updated so that they expect `CStrings` instead?

Eventually we want to have more higher level abstractions.

> because the compiler complained that `count_bytes()` was too new when I tried to replace `servo_id.len()` with `servo_id.count_bytes()` and I couldn't find a way around that.

For that we would need to be on rust 1.79, but we are currently on 1.78.

I would also like to thank you for very well written PR description.

sagudev (changes requested):
Causes a [warning](https://github.com/servo/servo/actions/runs/9820774526/job/27116170279?pr=32716#step:12:580):
```
warning: unused import: `crate::dom::bindings::utils::AsCCharPtrPtr`
  --> components/script/dom/promise.rs:37:5
   |
37 | use crate::dom::bindings::utils::AsCCharPtrPtr;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

but otherwise looks good.

abumsoo:
> Causes a [warning](https://github.com/servo/servo/actions/runs/9820774526/job/27116170279?pr=32716#step:12:580):
> 
> ```
> warning: unused import: `crate::dom::bindings::utils::AsCCharPtrPtr`
>   --> components/script/dom/promise.rs:37:5
>    |
> 37 | use crate::dom::bindings::utils::AsCCharPtrPtr;
>    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = note: `#[warn(unused_imports)]` on by default
> ```
> 
> but otherwise looks good.

Fixed!

sagudev (approved):

PR #32724, opened 2024-07-08
ci: fix security issue in try job workflow

mukilan:
The issue is that `ref/pull/{pull_request_number}` points to the latest
commit of a PR and so it could be different than the commit that was
reviewed when the try label was applied. The fix is to use the exact commit
sha at the point when the try job is triggered, which is available in
the `github` context as  `github.event.pull_request.head.sha`.

This [issue][1] was reported by GitHub user @RedYetiDev via the Security
Advisory reporting mechanism on GitHub. The fix is also based on their
proposed solution.

[1]: https://github.com/servo/servo/security/advisories/GHSA-fxqr-xgh8-3577

Co-authored-by: Aviv Keller <38299977+RedYetiDev@users.noreply.github.com>

---
- [x] These changes do not require tests because they just modify CI.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -31,15 +31,15 @@ jobs:
         arch: ['aarch64-linux-android', 'armv7-linux-androideabi', 'i686-linux-android', 'x86_64-linux-android']
     steps:
       - uses: actions/checkout@v4
-        if: github.event_name != 'issue_comment' && github.event_name != 'pull_request_target'
+        if: github.event_name != 'pull_request_target'
         with:
           fetch-depth: 2
-      # This is necessary to checkout the pull request if this run was triggered
-      # via an `issue_comment` action on a pull request.
+      # This is necessary to checkout the pull request if this run was triggered via a
+      # `pull_request_target` event.
       - uses: actions/checkout@v4
-        if: github.event_name == 'issue_comment' || github.event_name == 'pull_request_target'
+        if: github.event_name == 'pull_request_target'
         with:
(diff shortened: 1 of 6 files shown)


jschwe:
Do the `try` jobs have access to any secrets and if yes, do they really need to?

mrobinson (changes requested):
I think this should also update `try-label.yml` as well which checks out the code for the try_parser. Maybe it shouldn't do this though and simply use the try_parser from the main branch.

mukilan:
> I think this should also update try-label.yml as well which checks out the code for the try_parser. Maybe it shouldn't do this though and simply use the try_parser from the main branch.

Good catch! I've removed that one in 3233034

mrobinson (approved):

mukilan:
> Do the try jobs have access to any secrets and if yes, do they really need to?

I believe they have access because of the `secrets: inherit` in try-label.yml and dispatch-workflow.yml. Based on my understanding, I don't think they need that access currently - @sagudev or @mrobinson correct me if I'm wrong.

But apart from the secrets, the GITHUB_TOKEN passed to the workflow has read/write access to be able to comment on the PR, so this fix is still required and the usage of secrets can be reviewed and addressed in a separate PR.

mrobinson:
> I believe they have access because of the `secrets: inherit` in try-label.yml and dispatch-workflow.yml. Based on my understanding, I don't think they need that access currently - @sagudev or @mrobinson correct me if I'm wrong.

They have access to secrets so that they can interact with the intermittent dashboard currently, AFAIR.

sagudev (approved):
Interesting.

PR #32730, opened 2024-07-08
Start implementing stream internal slots and algos

gterzian:
---


components/script/dom/readablestream.rs
```diff
@@ -90,266 +118,196 @@ impl ReadableStream {
...
 
-    fn new(
-        global: &GlobalScope,
-        external_underlying_source: Option<Rc<ExternalUnderlyingSourceController>>,
-    ) -> DomRoot<ReadableStream> {
-        reflect_dom_object(
-            Box::new(ReadableStream::new_inherited(external_underlying_source)),
-            global,
-        )
+    fn new(global: &GlobalScope, controller: Controller) -> DomRoot<ReadableStream> {
```
wusyong, line 145: Step 3 in https://streams.spec.whatwg.org/#rs-constructor will need to create a `ReadableStream` but controller isn't created yet at that time. The controller is set in https://streams.spec.whatwg.org/#set-up-readable-stream-default-controller.

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -2,18 +2,124 @@
...
+            global,
+        )
+    }
+
+    pub fn set_stream(&self, stream: &ReadableStream) {
+        self.stream.set(Some(stream))
+    }
+
+    /// <https://streams.spec.whatwg.org/#ref-for-abstract-opdef-readablestreamcontroller-pullsteps>
+    pub fn perform_pull_steps(&self, read_request: ReadRequest) {
```
wusyong, line 87: My issue is more about [`pullAlgorithm`](https://streams.spec.whatwg.org/#readablestreamdefaultcontroller-pullalgorithm) and similar internal slots. Because if you search `Let pullAlgorithm`, you can see there are multiple ways to create it. Do you have any suggestions on how to implement it?

wusyong (changes requested):

components/script/dom/readablestreamdefaultcontroller.rs
jdm: Using a trait object or function pointer would be my first attempt.

components/script/dom/readablestream.rs
gterzian: Right, that step is still called from within the constructor, but if necessary we can wrap the controller in a `MutNullableDom`.

components/script/dom/readablestreamdefaultcontroller.rs
gterzian: I've added a commit that calls into the pull algorithm.

wusyong: Ah I guess that means we will need to implement the algos to other vairants.

wusyong (approved):
LGTM now, but I still don't have the permissions. Could I get some access to review and add labels?

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    #[allow(unsafe_code)]
+    pub fn new(
+        global: &GlobalScope,
+        underlying_source_type: UnderlyingSourceType,
+    ) -> DomRoot<UnderlyingSourceContainer> {
+        // Setting the prototype of the underlying source dict on the
+        // `UnderlyingSourceContainer` for later use in Call_.
+        // TODO: is this a good idea?
```
gterzian, line 74: cc @jdm

gterzian: I ended-up removing this entirely for now, it was crashing.

gterzian:
Ok, I also turned-off the old streams, and as a kind of milestone, https://github.com/servo/servo/blob/496ce717c5343984e0f3da00d223604eff03893c/tests/wpt/tests/fetch/api/basic/stream-response.any.js is now passing again(which covers part of the "native" streaming I handled in this PR). 

So I'd say let review and merge this one, so that work can resume in parallel on:
- The BYOB controller/reader part(note we were [not running tests](https://github.com/servo/servo/blob/496ce717c5343984e0f3da00d223604eff03893c/tests/wpt/include.ini#L217) for that part in the previous implementation, even though SM seems the implement it)
- Various TODOs in the default controller/reader. 
- other stuff?

components/script/dom/underlyingsourcecontainer.rs
wusyong: I still think we probably should expand the codegen to support underlying source that can call `Call`. I guess this is because it's the first dictionary that has callbacks.

jdm:
I'll try to take a look at this tonight!

components/script/dom/underlyingsourcecontainer.rs
jdm: One thing that occurs to me—I believe the DomObject bound us only needed to call wrap_call_this_object, which only uses it to call the reflector method so it can get a JS object pointer. If we add a new trait to return the JS object directly instead, we should not need to play around with dictionary prototypes, just generate an impl for the new trait instead.

sagudev:
> note we were [not running tests](https://github.com/servo/servo/blob/496ce717c5343984e0f3da00d223604eff03893c/tests/wpt/include.ini#L217) for that part in the previous implementation, even though SM seems the implement it

Doesn't that mean that readable-streams test's are run (not skipped)?

components/script/dom/readablestream.rs
```diff
@@ -182,7 +203,15 @@ impl ReadableStream {
 
     /// <https://streams.spec.whatwg.org/#readable-stream-add-read-request>
     pub fn add_read_request(&self, read_request: ReadRequest) {
-        self.reader.get().unwrap().add_read_request(read_request);
+        match self.reader {
+            ReaderType::Default(ref reader) => {
+                let Some(reader) = reader.get() else {
+                    panic!("Attempt to read stream chunk without having acquired a reader.");
```
jdm, line 221: nit: s/read stream chunk/add read request/

components/script/dom/readablestream.rs
```diff
@@ -235,39 +264,54 @@ impl ReadableStream {
...
-            panic!("Attempt to read stream chunk without having acquired a reader.");
-        };
-
-        reader.ReleaseLock();
-
-        self.reader.set(None);
+        match self.reader {
+            ReaderType::Default(ref reader) => {
+                let Some(rooted_reader) = reader.get() else {
+                    panic!("Attempt to read stream chunk without having acquired a reader.");
```
jdm, line 326: nit: s/read stream chunk/stop reading/

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -75,6 +78,7 @@ impl UnderlyingSourceContainer {
         let cx = GlobalScope::get_cx();
         rooted!(in(*cx) let mut underlying_source_dict = UndefinedValue());
         if let UnderlyingSourceType::Js(ref js_source) = underlying_source_type {
+            #[allow(unsafe_code)]
```
jdm: nit: it makes more sense to have this annotation at the method level.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -29,18 +34,47 @@ pub enum ReadRequest {
 
 impl ReadRequest {
     /// <https://streams.spec.whatwg.org/#read-request-chunk-steps>
+    #[allow(unsafe_code)]
+    #[allow(crown::unrooted_must_root)]
```
jdm, line 41: Why is this annotation required?

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -29,18 +34,47 @@ pub enum ReadRequest {
...
     pub fn close_steps(&self) {
         match self {
-            // TODO: [undefined, true]
-            ReadRequest::Read(promise) => promise.resolve_native(&()),
+            ReadRequest::Read(promise) => {
+                let cx = GlobalScope::get_cx();
+                rooted!(in(*cx) let mut rval = UndefinedValue());
+                let result = RootedTraceableBox::new(Heap::default());
+                result.set(*rval);
+                let result = ReadableStreamReadResult {
```
jdm, line 70: This value does not appear to be used.

jdm (changes requested):

gterzian:
> Doesn't that mean that readable-streams test's are run (not skipped)?

yes but the byte stream stuff is in a separate folder, at https://github.com/servo/servo/tree/main/tests/wpt/tests/streams/readable-byte-streams, so we were never running them.

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -2,18 +2,92 @@
...
+
+        // Call into underlying source if necessary.
+
+        // else, append read request to reader.
+        self.stream.get().unwrap().add_read_request(read_request);
+    }
+
+    pub fn enqueue_chunk(&self, mut chunk: Vec<u8>) {
+        let mut buffer = self.buffer.borrow_mut();
+        chunk.append(&mut buffer);
```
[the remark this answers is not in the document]
gterzian: Yes that seems to be the case. 

On this and the comment below: I looked into this and actually we should re-structure this queue concept to allow to any JS value being a chunk, a first step of which I have done in the last commit. There is still a "native" chunk variant, which I guess we could optimize.

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -6,18 +6,55 @@ use std::cell::{Cell, RefCell};
...
+    fn callback(&self, cx: JSContext, v: HandleValue, _realm: InRealm) {
+        todo!();
+    }
+}
+
+/// The rejection handler for
+/// <https://streams.spec.whatwg.org/#readable-stream-default-controller-call-pull-if-needed>
+#[derive(Clone, JSTraceable, MallocSizeOf)]
+struct PullAlgorithmRejectionHandler {
+    controller: DomRoot<ReadableStreamDefaultController>,
```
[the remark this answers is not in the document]
gterzian: How about using `Dom`? See last commit. Since `Trusted` is atomic and we don't need this here...

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -88,6 +99,27 @@ impl ReadableStreamDefaultController {
...
         *buffer = chunk;
     }
+
+    /// Does the stream have all data in memory?
+    pub fn in_memory(&self) -> bool {
+        self.underlying_source.in_memory()
+    }
+
+    /// Return bytes synchronously if the stream has all data in memory.
+    pub fn get_in_memory_bytes(&self) -> Option<Vec<u8>> {
```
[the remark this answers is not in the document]
gterzian: Yes I think we can look into something like that, this method is only used in `body.rs`.

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -28,8 +29,9 @@ use crate::script_runtime::{JSContext, JSContext as SafeJSContext};
 /// The fulfillment handler for
 /// <https://streams.spec.whatwg.org/#readable-stream-default-controller-call-pull-if-needed>
 #[derive(Clone, JSTraceable, MallocSizeOf)]
+#[allow(crown::unrooted_must_root)]
 struct PullAlgorithmFulfillmentHandler {
-    controller: DomRoot<ReadableStreamDefaultController>,
+    controller: Dom<ReadableStreamDefaultController>,
```
jdm, line 34: I don't think this is safe, since these structs aren't visible to the GC. That's why I recommend Trusted here.

jdm:
https://github.com/servo/servo/commit/b65ba8df924db6d0f2263798c6d8810dd953b64f is how I would support dictionary callback values.

components/script/dom/readablestreamdefaultcontroller.rs
gterzian: > That's why I recommend Trusted here.
I just tried it and it breaks `Callback: JSTraceable + MallocSizeOf `. 

>  these structs aren't visible to the GC

But they end-up on the `PromiseNativeHandler`, which is a dom_struct.

components/script/dom/readablestreamdefaultreader.rs
gterzian: I was able to remove it by avoiding to assign the read result to a temporary variable.

gterzian: I was able to remove it by avoiding to assign the read result to a temporary variable.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -2,25 +2,50 @@
...
+
 /// <https://streams.spec.whatwg.org/#readablestreamdefaultreader>
 #[dom_struct]
 pub struct ReadableStreamDefaultReader {
     reflector_: Reflector,
+
+    /// <https://streams.spec.whatwg.org/#readablestreamgenericreader-stream>
+    stream: Dom<ReadableStream>,
+
+    #[ignore_malloc_size_of = "Rc is hard"]
```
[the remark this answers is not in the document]
gterzian: The read request contains a `Rc<Promise>`.

components/script/dom/underlyingsourcecontainer.rs
gterzian: thank you, have included your example.

components/script/dom/readablestreamdefaultcontroller.rs
gterzian: I've strengthened this by keeping the controller rooted while the native promise handler is setup.

wusyong:
> @wusyong @Taym95 I will address the remaining points from the review, and then I think we need to merge and work in parallel, because I am starting to run into some complicated "how to do this in JS" stuff that warrant PRs in and of themselves...

I thought this PR would be merged earlier so I was waiting for it. Byte controller and reader is pretty similar to default one. So I believe once we have done the default ones, the byte part will be faster. I'll review the PR again pretty soon.

wusyong (approved):
@jdm @gterzian Could we merge this today? I believe @gterzian [mentioned ](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Re-implement.20ReadableStream/near/451703639)we could do more aggressive mergine since this is targeted to the feature branch.
There are some trivial improvements like [EnqueuedValue](https://github.com/servo/servo/pull/32730/files#diff-09888dfa911f121670b8b905131455cf548d1816bf65bdeff76fb00f9e55e743R70) and maybe QueueWithSize should share with byob controller. But I think it could wait for there's PR for such controller implementation.

Edit: I also hope we can update the feature branch to catch up main branch since there are some more commit over weeks.

gterzian:
Ok let's do it! @jdm I have addressed all of your comments, although I still had a few questions(about the rooting), but this had been noted in a TODO as needed to revisit...

PR #32731, opened 2024-07-08
ohos: Detect installed fonts

jschwe:
Improves the font fallback on OpenHarmony by scanning the systems font folder to detect installed fonts,
derive their associated font names and generate the font list based on that.
Also adds an `ohos_mock` cfg option, which allows using the ohos selection code on a non-ohos target, i.e. allowing the user to test the OpenHarmony font fallback on a Linux host, using the fonts from the OpenHarmony SDK previewer.

For now it seems that parsing the font family name from the filename seems like the most reliable option, since the `fontconfig.json` configuration file seems to be unreliable in practice..

---

### DIFF
--- a/components/fonts/platform/freetype/ohos/font_list.rs
+++ b/components/fonts/platform/freetype/ohos/font_list.rs
@@ -260,29 +260,6 @@ fn parse_font_filenames(font_files: Vec<PathBuf>) -> Vec<FontFamily> {
         Some((family_name, font))
     });
 
-    let hmos_emoji_fonts = font_files.iter().filter_map(|file_path| {
-        let hm_emojis_base_name_components = ["HMOS", "Color", "Emoji"];
-        let base_file_stem = hm_emojis_base_name_components.join("");
-        let stem = file_path.file_stem()?.to_str()?;
-        let stem_no_prefix = stem.strip_prefix(base_file_stem.as_str())?;
-        let family_variant = match stem_no_prefix {
-            "Compat" => "",
-            "Flags" => "Flags",
-            variant => {
-                warn!("Encountered unknown font variant `HMOS Color Emoji {variant}`");
-                variant
-            },
-        };
-        let mut name_components = hm_emojis_base_name_components.to_vec();
-        name_components.push(family_variant);
-        let family_name = name_components.join(" ");
-        let font = Font {
-            filepath: file_path.to_str()?.to_string(),
-            ..Default::default()
-        };
-        Some((family_name, font))
-    });
-
     let noto_fonts = font_files.iter().filter_map(|file_path| {
         let stem = file_path.file_stem()?.to_str()?;
         // Filter out non-noto fonts
@@ -324,7 +301,7 @@ fn parse_font_filenames(font_files: Vec<PathBuf>) -> Vec<FontFamily> {
         Some((family_name, font))
     });
 
-    let all_families = harmony_os_fonts.chain(hmos_emoji_fonts).chain(noto_fonts);
+    let all_families = harmony_os_fonts.chain(noto_fonts);
 
     for (family_name, font) in all_families {
         if let Some(font_list) = families.get_mut(&family_name) {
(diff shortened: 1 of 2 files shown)


components/fonts/platform/freetype/ohos/font_list.rs
```diff
@@ -65,23 +103,265 @@ struct FontList {
...
+fn detect_hos_font_style(font_modifiers: &[&str]) -> Option<String> {
+    if font_modifiers.contains(&"Italic") {
+        Some("italic".to_string())
+    } else {
+        None
+    }
+}
+
+// Note: The weights here are taken from the `alias` section of the fontconfig.json
+fn detect_hos_font_weight_alias(font_modifiers: &[&str]) -> Option<i32> {
```
jschwe, line 132: This function infers the `weight` attribute for `HarmonyOS Sans` fonts, based on suffixes such as `Medium`, `Bold` etc. The numbers here are taking from the alias section of fontconfig.json which I added below.
I'm not sure if I should use this weight here, or if the alias weight, should only be added to the `FontAlias`es.
Happy for any recommendations.

```json
{
  "fontdir": ["fonts"],
  "generic": [
    {
      "family": "HarmonyOS Sans",
      "alias": [
        {
          "HarmonyOS-Sans": 0
        },
        {
          "HarmonyOS-Sans-Light": 100
        },
        {
          "HarmonyOS-Sans-Regular": 400
        },
        {
          "HarmonyOS-Sans-Medium": 700
        },
        {
          "HarmonyOS-Sans-Bold": 900
        }
      ],
      "adjust": [
        {
          "weight": 50, "to": 100
        },
        {
          "weight": 80, "to": 400
        },
        {
          "weight": 100, "to": 700
        },
        {
          "weight": 200, "to": 900
        }
      ]
    },
    {
      "family": "HarmonyOS Sans Condensed",
      "alias": [
        {
          "HarmonyOS-Sans-Condensed": 0
        }
      ]
    },

```

components/fonts/platform/freetype/ohos/font_list.rs
```diff
@@ -68,23 +105,265 @@ struct FontList {
...
+    name_components
+}
+
+/// Parse the font file names to determine the available FontFamilies
+///
+/// In principle the `fontconfig.json` should provide information on the available fonts,
+/// but on some devices the `font_file_map` member is missing, and in general the family name to
+/// file name mapping seems to not be accurate, so we instead infer the family name from the
+/// file names of the installed fonts.
+fn parse_font_filenames(font_files: Vec<PathBuf>) -> Vec<FontFamily> {
```
mrobinson, line 215: There's a lot of code here to parse font names. How many fonts does this deal with in practice? Perhaps it would be simpler to simply have a pre-calculated list of fonts and with their weights? This parsing isn't always going to work in practice. For instance `NotoSerif[wdth,wght].ttf` likely contains multiple weights (perhaps named variants of using OpenType variations -- you can determine this with FreeType. Parsing the name is never going to give you access to this.

jschwe: > How many fonts does this deal with in practice? 

I recently updated my phone to a newer pre-release version and now there are 193 fonts in `/system/fonts`, 180 of which are Noto fonts. Previously there were a lot less Noto fonts.
The filenames of the HarmonyOS fonts also changed compared to the last version, and the available OpenHarmony versions - the previously present "Regular" suffix is now omitted in the filenames of HarmonyOS Sans fonts.

> For instance NotoSerif[wdth,wght].ttf likely contains multiple weights (perhaps named variants of using OpenType variations -- you can determine this with FreeType

How much support is there currently in servo for variable font features on other platforms? I guess I would need to manually construct the `FontTemplateDescriptor`, since the constructor assumes non-variable fonts.
Is correctly setting the weight range mandatory, or could it also be left at the default value (for now), and we would just lose the variable aspect of the fonts (but without anything breaking in wierd ways)?

The main reason I went with parsing the font names here, is that I somehow need to get the family names. On the most recent version I'm running there seems to be a newly added `font_file_map` map in the `fontconfig.json` file, which might be an option if we require OpenHarmony 5.0 and newer. But at least in my current preview build that config file is still  a bit buggy, and many font filenames in the map don't exist on the actual device.

jschwe: @mrobinson For now I at least hardcoded the paths for the HMOS color emojis, since their paths haven't changed. However, I would still like to keep the parsing logic to support the current OH 5.0 beta, at least until an update has been released, which provides the correct filenames in the `fontconfig.json` (It's already been fixed, should be in the next beta version, whenever that is released)

mrobinson: > How much support is there currently in servo for variable font features on other platforms? I guess I would need to manually construct the `FontTemplateDescriptor`, since the constructor assumes non-variable fonts.
> Is correctly setting the weight range mandatory, or could it also be left at the default value (for now), and we would just lose the variable aspect of the fonts (but without anything breaking in wierd ways)?

Apologies for the late reply: Servo has supported for named variations (which are usually stored in font tables), but not support for OpenType variations. Named variations are usually expressed as the font index. These are the kind that are typically returned via FreeType/FontConfig.

mrobinson:
I'm happy to review this if you think it's ready. I understand that the font configuration stored on OpenHarmony is in flux (hopefully it can be accurate in an upcoming version), so this code will have to adapt to the reality of the OS.

jschwe:
A review would be great. I'm also not a huge fan of parsing the font names, but I hope it is a reasonable intermediate solution.

mrobinson (approved):

PR #32741, opened 2024-07-09
script: Use the new C string literal in the DOM bindings

### LINKED ISSUE #32628: Replace null-byte terminated string literals with C-string literals

abumsoo:
Most of the changes in this PR are converting byte strings to c-strings in `CodegenRust.py`. However, some of the functions from the code that is being generated by `CodegenRust.py` that use these strings expect `*const i8` i.e. `*const c_char` and some other functions expect `&[u8]` which is a byte slice. Hence, the old helper, `str_to_const_array()`, was changed to `str_to_cstr_ptr()` and an additional helper, `str_to_byte_slice()`, was added for the conversion to `&[u8]`

This was the best solution I could think of, but I imagine there is probably a better solution. Any suggestions would be appreciated.

---

- [X] These changes fix #32628

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2084,7 +2084,7 @@ class ConstDefiner(PropertyDefiner):
             return ""
 
         def specData(const):
-            return (str_to_byte_slice(const.identifier.name),
+            return (str_to_cstr(const.identifier.name),
                     convertConstIDLValueToJSVal(const.value))
 
         return self.generateGuardedArray(
@@ -2466,12 +2466,12 @@ impl %(selfName)s {
 """ % args
 
 
+def str_to_cstr(s):
+    return "c\"%s\"" % s
+
+
 def str_to_cstr_ptr(s):
     return "c\"%s\".as_ptr()" % s
-
-
-def str_to_byte_slice(s):
-    return "c\"%s\".to_bytes_with_nul()" % s
 
 
 class CGPrototypeJSClass(CGThing):
@@ -2515,7 +2515,7 @@ class CGInterfaceObjectJSClass(CGThing):
 static NAMESPACE_OBJECT_CLASS: NamespaceObjectClass = unsafe {
     NamespaceObjectClass::new(%s)
 };
-""" % str_to_byte_slice(classString)
+""" % str_to_cstr(classString)
         if self.descriptor.interface.ctor():
             constructorBehavior = "InterfaceConstructorBehavior::call(%s)" % CONSTRUCT_HOOK_NAME
         else:
@@ -3296,7 +3296,7 @@ rooted!(in(*cx) let proto = {proto});
 assert!(!proto.is_null());
 rooted!(in(*cx) let mut namespace = ptr::null_mut::<JSObject>());
 create_namespace_object(cx, global, proto.handle(), &NAMESPACE_OBJECT_CLASS,
-                        {methods}, {constants}, {str_to_byte_slice(name)}, namespace.handle_mut());
+                        {methods}, {constants}, {str_to_cstr(name)}, namespace.handle_mut());
 assert!(!namespace.is_null());
 assert!((*cache)[PrototypeList::Constructor::{id} as usize].is_null());
 (*cache)[PrototypeList::Constructor::{id} as usize] = namespace.get();
@@ -3315,7 +3315,7 @@ assert!((*cache)[PrototypeList::Constructor::%(id)s as usize].is_null());
 <*mut JSObject>::post_barrier((*cache).as_mut_ptr().offset(PrototypeList::Constructor::%(id)s as isize),
                               ptr::null_mut(),
                               interface.get());
-""" % {"id": name, "name": str_to_byte_slice(name)})
+""" % {"id": name, "name": str_to_cstr(name)})
 
         parentName = self.descriptor.getParentName()
         if not parentName:
@@ -3386,7 +3386,7 @@ assert!((*cache)[PrototypeList::ID::%(id)s as usize].is_null());
 """ % proto_properties))
 
         if self.descriptor.interface.hasInterfaceObject():
-            properties["name"] = str_to_byte_slice(name)
+            properties["name"] = str_to_cstr(name)
             if self.descriptor.interface.ctor():
                 properties["length"] = methodLength(self.descriptor.interface.ctor())
             else:
@@ -3481,11 +3481,11 @@ assert!((*cache)[PrototypeList::Constructor::%(id)s as usize].is_null());
 
         constructors = self.descriptor.interface.legacyFactoryFunctions
         if constructors:
-            decl = "let named_constructors: [(ConstructorClassHook, &[u8], u32); %d]" % len(constructors)
+            decl = "let named_constructors: [(ConstructorClassHook, &std::ffi::CStr, u32); %d]" % len(constructors)
             specs = []
             for constructor in constructors:
                 hook = CONSTRUCT_HOOK_NAME + "_" + constructor.identifier.name
-                name = str_to_byte_slice(constructor.identifier.name)
+                name = str_to_cstr(constructor.identifier.name)
                 length = methodLength(constructor)
                 specs.append(CGGeneric("(%s as ConstructorClassHook, %s, %d)" % (hook, name, length)))
             values = CGIndenter(CGList(specs, "\n"), 4)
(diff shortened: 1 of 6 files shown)


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2511,7 +2515,7 @@ def define(self):
 static NAMESPACE_OBJECT_CLASS: NamespaceObjectClass = unsafe {
     NamespaceObjectClass::new(%s)
 };
-""" % str_to_const_array(classString)
+""" % str_to_byte_slice(classString)
```
sagudev, line 2518: Here we can change `NamespaceObjectClass::new` to accept CStr: https://doc.servo.org/src/script/dom/bindings/namespace.rs.html#25-34

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -3292,7 +3296,7 @@ def definition_body(self):
 assert!(!proto.is_null());
 rooted!(in(*cx) let mut namespace = ptr::null_mut::<JSObject>());
 create_namespace_object(cx, global, proto.handle(), &NAMESPACE_OBJECT_CLASS,
-                        {methods}, {constants}, {str_to_const_array(name)}, namespace.handle_mut());
+                        {methods}, {constants}, {str_to_byte_slice(name)}, namespace.handle_mut());
```
sagudev, line 3299: We can also change `create_namespace_object` to accept CStr.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -6476,7 +6480,7 @@ def reexportedName(name):
         if haveLegacyWindowAliases:
             cgThings.append(
                 CGList([CGGeneric("const legacy_window_aliases: &[&[u8]] = &["),
-                        CGIndenter(CGList([CGGeneric(str_to_const_array(name)) for
+                        CGIndenter(CGList([CGGeneric(str_to_byte_slice(name)) for
```
sagudev, line 6483: I think we need to change `&[&[u8]]` -> `&[&CStr]`, although this might require more changes.

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2084,7 +2084,7 @@ def generateArray(self, array, name):
             return ""
 
         def specData(const):
-            return (str_to_const_array(const.identifier.name),
+            return (str_to_byte_slice(const.identifier.name),
                     convertConstIDLValueToJSVal(const.value))
```
sagudev, line 2088: I traced usage of this to https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L1636, so this would require change in https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/constant.rs#L16-L21

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -3311,7 +3315,7 @@ def definition_body(self):
 <*mut JSObject>::post_barrier((*cache).as_mut_ptr().offset(PrototypeList::Constructor::%(id)s as isize),
                               ptr::null_mut(),
                               interface.get());
-""" % {"id": name, "name": str_to_const_array(name)})
+""" % {"id": name, "name": str_to_byte_slice(name)})
```
sagudev, line 3318: Dis would require change in https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/interface.rs#L214-L230

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -3481,7 +3485,7 @@ def defineAliasesFor(m):
             specs = []
             for constructor in constructors:
                 hook = CONSTRUCT_HOOK_NAME + "_" + constructor.identifier.name
-                name = str_to_const_array(constructor.identifier.name)
```
sagudev, line 3484: Here we need to change `create_named_constructors`: https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/interface.rs#L324 and then change https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/codegen/CodegenRust.py#L3480

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -3382,7 +3386,7 @@ def definition_body(self):
 """ % proto_properties))
 
         if self.descriptor.interface.hasInterfaceObject():
-            properties["name"] = str_to_const_array(name)
```
sagudev, line 3385: This would require change in https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/interface.rs#L278-L290

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -6457,7 +6461,7 @@ def reexportedName(name):
                 haveUnscopables = True
                 cgThings.append(
                     CGList([CGGeneric("const unscopable_names: &[&[u8]] = &["),
-                            CGIndenter(CGList([CGGeneric(str_to_const_array(name)) for
+                            CGIndenter(CGList([CGGeneric(str_to_byte_slice(name)) for
```
sagudev, line 6464: We need to change the type above and fix https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/interface.rs#L241

sagudev:
I tried to trace all `str_to_byte_slice` usages and I think we can replace them all, I write where needs to be more changes. General pattern is that a function (usually somewhere in `script::dom::bindings`) have argument `name: &[u8]` and that name is casted to `˙*const c_char` in the function.

abumsoo:
> I tried to trace all `str_to_byte_slice` usages and I think we can replace them all, I wrote where what need to be changed to remove `str_to_byte_slice`. General pattern is that a function (usually somewhere in `script::dom::bindings`) have argument `name: &[u8]` and that name is casted to `˙*const c_char` in the function.

Thank you for the detailed review! I'm going to work on those changes. I have one question. How did you know to look in those other files like `interface.rs` and `namespace.rs`? More generally, how did you know what part of `CodegenRust.py` to try to search?

sagudev:
> I have one question. How did you know to look in those other files like `interface.rs` and `namespace.rs`? More generally, how did you know what part of `CodegenRust.py` to try to search?

I looked at how result of `str_to_byte_slice` is used. For example https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L2518 is used for generated code here: https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L2516 and this is https://github.com/servo/servo/blob/a3bb8048fc09d7ce7b0f5a9bbed37b2f21ce866d/components/script/dom/bindings/namespace.rs#L25. While this example is less complex, some others are harder, but general idea is the same.

components/script/dom/bindings/codegen/CodegenRust.py
abumsoo: I saw your answer below for your process and it made a lot of sense. However I'm having  a hard time connecting the dots on this particular one. How did you go from https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L1636 to 
https://github.com/servo/servo/blob/1c6b74e1f16f6c92f80410d98608064fa4fbede2/components/script/dom/bindings/constant.rs#L16-L21
I'm not seeing anything in `generateGuardedArray` that would make me think to check inside of `constant.rs`.
The only thing I would have thought of would be to search for files where `ConstantSpec` is being used.

sagudev: https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L2086 is used here: https://github.com/abumsoo/servo/blob/8e9807fa6f3cfb82c765002d480e9d2da8f1ea4a/components/script/dom/bindings/codegen/CodegenRust.py#L2095 (passed as "fn pointer" or whatever this is called in python). And looking other arguments, and how they are used inside `generateGuardedArray`, it appears to me that `ConstantSpec` needs to be fixed. Then I searched for where `ConstantSpec` is defined.

sagudev: This one was very tricky to be honest.

abumsoo: Ohh I see. Thank you for walking me through your thought process. That was very helpful.

sagudev (approved):

PR #32761, opened 2024-07-12
Fix offsetLeft/offsetTop to match major browsers

nicoburns:
Although the spec (https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface) says that offsetTop/offsetLeft are always relative to the padding box of the offsetParent. In practice this is not true in major browsers in the case that the offsetParent is the body element and the body element is position:static. In that case offsetLeft/offsetTop are computed relative to the root nodes border box.

See: https://github.com/w3c/csswg-drafts/issues/10549

This PR updates Servo's behaviour to match that of major browsers. Which causes a number of "false fails" in the WPT suite to become passes.

---

- [x] There are tests for these changes (WPT tests rely on this behaviour when checking `data-offset-x` and `data-offset-y` properties)


atbrakhi (commented):
So after going through the [zulip](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/offsetTop.2FoffsetLeft/near/450306790) discussion as well as the PR description, am I understanding it correctly that in this PR, we are moving away from what the specification says to match the behavior of other browser engine that is not correct according to the specification?

nicoburns:
> So after going through the [zulip](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/offsetTop.2FoffsetLeft/near/450306790) discussion as well as the PR description, am I understanding it correctly that in this PR, we are moving away from what the specification says to match the behavior of other browser engine that is not correct according to the specification?

Yes, that is correct. I am assuming that if all major browser agree and WPT tests rely on a behaviour then it's likely to be the spec that ends up changing.

components/layout_2020/query.rs
```diff
@@ -412,15 +413,20 @@ fn process_offset_parent_query_inner(
 
             Some(NodeOffsetBoxInfo {
                 border_box,
-                offset_parent_node_address,
+                offset_parent_node_address: offset_parent_node.map(|node| node.0),
+                is_static_body_element: offset_parent_node.map(|node| node.1).unwrap_or(false),
```
Loirooriol, line 417: This could use `map_or`.

components/layout_2020/query.rs
```diff
@@ -446,45 +452,70 @@ fn process_offset_parent_query_inner(
...
-        // "If the offsetParent of the element is null," subtract zero in the
-        // following step.
-        .unwrap_or(Vector2D::zero());
+    let offset_parent_padding_box_corner =
+        if let Some(offset_parent_node_address) = node_offset_box.offset_parent_node_address {
+            // Although the spec (https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface)
+            // says that offsetTop/offsetLeft are always relative to the padding box of the offsetParent.
+            // In practice this is not true in major browsers in the case that the offsetParent is the body
+            // element and the body element is position:static. In that case offsetLeft/offsetTop are computed
+            // relative to the root nodes border box.
```
Loirooriol, line 461: It seems that other browsers aren't interoperable either, I'm checking

```html
<!DOCTYPE html>
<body id="body" onload="recompute()" style="border: solid">
  <div id="target" style="height: 40px; width: 40px; background: red;"></div>
  <div id="output" style="text-wrap: nowrap"></div>
  <button onclick="recompute()">recompute</button>
  <script type="text/javascript">
    function recompute() {
      let targetNode = document.getElementById("target");
      let outputNode = document.getElementById("output");
      outputNode.innerHTML = `
        offsetParent: ${targetNode.offsetParent.id}<br />
        offsetTop: ${targetNode.offsetTop}<br />
        offsetLeft: ${targetNode.offsetLeft}<br />
      `;
    }
  </script>
</body>
```

and Firefox says 8, Chromium says 11.

components/layout_2020/query.rs
```diff
@@ -446,45 +452,70 @@ fn process_offset_parent_query_inner(
...
+                            .origin
+                            .to_vector() +
+                            containing_block.origin.to_vector();
+                        border_box_corner.to_untyped()
+                    },
+                    Fragment::AbsoluteOrFixedPositioned(_) |
+                    Fragment::Text(_) |
+                    Fragment::Image(_) |
+                    Fragment::IFrame(_) |
+                    Fragment::Positioning(_) => unreachable!(),
```
Loirooriol, line 479: This is not unreachable, e.g. this crashes

```html
<!DOCTYPE html>
<html style="position: absolute">
<body id="body" onload="recompute()" style="border: solid">
  <div id="target" style="height: 40px; width: 40px; background: red;"></div>
  <div id="output" style="text-wrap: nowrap"></div>
  <button onclick="recompute()">recompute</button>
  <script type="text/javascript">
    function recompute() {
      let targetNode = document.getElementById("target");
      let outputNode = document.getElementById("output");
      outputNode.innerHTML = `
        offsetParent: ${targetNode.offsetParent.id}<br />
        offsetTop: ${targetNode.offsetTop}<br />
        offsetLeft: ${targetNode.offsetLeft}<br />
      `;
    }
  </script>
</body>
```

Loirooriol (commented):
I think the specified behavior makes more sense but since nobody follows it...

components/layout_2020/query.rs
nicoburns: Hmm... interesting, so:
1. I get 10 in Chromium (macOS) for your HTML above
2. 1px of the difference just seems to be a different default border thickness. Pinning that to 1px gives me 9px in Chromium.
3. Firefox's output of 8px looks like a bug to me. There is nothing that the node is offset 8px from.

But I guess the fact that there isn't perfect interoperability here might mean there is more scope for changing the behaviour.

Loirooriol: You getting 10 seems strange because the body should get a margin of 8px (https://html.spec.whatwg.org/multipage/rendering.html#the-page), and a border width defaults to 3px (https://drafts.csswg.org/css-backgrounds-3/#border-width). I guess that if your device has e.g. 0.5dppx, then the 3px would be 1.5 device pixels, which would be [snapped](https://drafts.csswg.org/css-values-4/#snap-a-length-as-a-border-width) to 1 device pixel, i.e. 2px. But I think 0.5dppx is weird.

Getting 8px on Firefox seems a bug indeed. This is what the offset includes:

| | Firefox | Blink, WebKit |
| - | - | - |
| `<html>` margin | ❌ | ❌ |
| `<html>` border | ✅ | ✅ |
| `<html>` padding | ✅ | ✅ |
| `<body>` margin | ✅ | ✅ |
| `<body>` border | ❌ | ✅ |
| `<body>` padding | ✅ | ✅ |

Note this happens even if the `<body>` in nested deeper in the tree (so it's not [the body element](https://html.spec.whatwg.org/multipage/dom.html#the-body-element-2)), using its parent (probably containing block?) instead of the `<html>` (if it has no parent, then the margin,border,padding of the parent are skipped).

nicoburns: > a border width defaults to 3px (https://drafts.csswg.org/css-backgrounds-3/#border-width)

So the default border width in Chrome is *weird* on my machine (Chrome 130.0.6723.59 on macOS arm64). It is affected by both dppx and zoom level. And it seems that [I am not the only one](https://stackoverflow.com/questions/75444838/html-css-default-border-with-1-5px-oder-3px).

At 100% zoom (unzoomed), the default appears to be 3 *device* pixels. Which comes out at 1.5px at my machine's default resolution which has 2dppx. If I set my machine to it's native resolution then I get a 3px default border width.

If I zoom IN then default border width is DECREASED. If I zoom OUT then default border width is INCREASED.

A dppx of 2 (haven't tested others) does NOT affect an explicitly set `medium` size, and neither does zooming in, but zooming out does (but less than the default size - seems to be a rounding thing rather than a linear scaling as it is for the default border size).

Not sure if this is a bug in Chrome (using device pixels somewhere they should be using logical pixels?), or an intentional behaviour to have a "better default" for border widths.

nicoburns: Looks like Firefox is tracking some of these issues, and that they're following an old IE model for offsetLeft/offsetTop that IE moved away from in IE8! https://bugzilla.mozilla.org/show_bug.cgi?id=255754 and https://bugzilla.mozilla.org/show_bug.cgi?id=1309427

components/layout_2020/query.rs
nicoburns: I have fixed this particular crash. Not sure if the root element can be any of the other variants.

components/layout_2020/query.rs
Loirooriol: Or `is_some_and`.

components/layout_2020/query.rs
```diff
@@ -437,9 +443,38 @@ fn process_offset_parent_query_inner(
...
+                containing_block: &euclid::Rect<Au, CSSPixel>,
+            ) -> Vector2D<Au, UnknownUnit> {
+                let (Fragment::Box(fragment) | Fragment::Float(fragment)) = fragment else {
+                    unreachable!();
+                };
+                // Again, take the *first* associated CSS layout box.
+                let border_box_corner =
+                    fragment.border_rect().origin.to_vector() + containing_block.origin.to_vector();
+                border_box_corner.to_untyped()
+            }
```
Loirooriol, line 466: Nit: no need to import `UnknownUnit` if using a lambda. And `euclid::Rect<Au, CSSPixel>` can be `PhysicalRect<Au>` .

```suggestion
            let extract_box_fragment =
                |fragment: &Fragment, containing_block: &PhysicalRect<Au>| {
                    let (Fragment::Box(fragment) | Fragment::Float(fragment)) = fragment else {
                        unreachable!();
                    };
                    // Again, take the *first* associated CSS layout box.
                    let border_box_corner = fragment.border_rect().origin.to_vector() +
                        containing_block.origin.to_vector();
                    border_box_corner.to_untyped()
                };
```

components/layout_2020/query.rs
Loirooriol: Not addressed

nicoburns:
I think I've addressed all requested changes now

components/layout_2020/query.rs
```diff
@@ -437,9 +444,36 @@ fn process_offset_parent_query_inner(
...
-        .offset_parent_node_address
-        .map(|offset_parent_node_address| {
+    let offset_parent_padding_box_corner = if let Some(offset_parent_node_address) =
+        node_offset_box.offset_parent_node_address
+    {
+        // Although the spec (https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface)
+        // says that offsetTop/offsetLeft are always relative to the padding box of the offsetParent.
+        // In practice this is not true in major browsers in the case that the offsetParent is the body
+        // element and the body element is position:static. In that case offsetLeft/offsetTop are computed
+        // relative to the root nodes border box.
```
Loirooriol, line 454:
```suggestion
        // relative to the root node's border box.
```

components/layout_2020/query.rs
```diff
@@ -437,9 +444,36 @@ fn process_offset_parent_query_inner(
...
 
-    let offset_parent_padding_box_corner = node_offset_box
-        .offset_parent_node_address
-        .map(|offset_parent_node_address| {
+    let offset_parent_padding_box_corner = if let Some(offset_parent_node_address) =
+        node_offset_box.offset_parent_node_address
+    {
+        // Although the spec (https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface)
+        // says that offsetTop/offsetLeft are always relative to the padding box of the offsetParent.
+        // In practice this is not true in major browsers in the case that the offsetParent is the body
```
Loirooriol, line 452: The 1st sentence ends abruptly.

```suggestion
        // The spec (https://www.w3.org/TR/cssom-view-1/#extensions-to-the-htmlelement-interface)
        // says that offsetTop/offsetLeft are always relative to the padding box of the offsetParent.
        // However, in practice this is not true in major browsers in the case that the offsetParent is the body
```

Loirooriol (approved):
Thanks, just some nits for the comment.

PR #32762, opened 2024-07-12
Mark flexbox-mbp-horiz-003 tests as only failing on linux

nicoburns:
These tests have been passing for me locally since I started working on Servo. Having the expectation not match the actual results is very annoying as it makes it much more difficult to update WPT results (the automated tooling is not sufficient). It seems to be a macOS vs. linux thing, and I have recently discovered that it is possible to mark tests as failing on only one platform, so that is what I have done here.

---


nicoburns:
I've also included these changes in https://github.com/servo/servo/pull/32761

mrobinson (approved):

PR #32769, opened 2024-07-12
Update mozjs (SpiderMonkey) to 128.0

sagudev:
Companion PR for https://github.com/servo/mozjs/pull/474

OffThread stuff is commented out.

Fun fact: No changes in codegen needed🎉

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ For more detailed build instructions, see the Servo book under [Setting up your 
 
 ### Android
 
+TODO: NDK 26c
+
 - Ensure that the following environment variables are set:
   - `ANDROID_SDK_ROOT`
   - `ANDROID_NDK_ROOT`: `$ANDROID_SDK_ROOT/ndk/25.2.9519653/`
--- a/README.md
+++ b/README.md
@@ -45,11 +45,9 @@ For more detailed build instructions, see the Servo book under [Setting up your 
 
 ### Android
 
-TODO: NDK 26c
-
 - Ensure that the following environment variables are set:
   - `ANDROID_SDK_ROOT`
-  - `ANDROID_NDK_ROOT`: `$ANDROID_SDK_ROOT/ndk/25.2.9519653/`
+  - `ANDROID_NDK_ROOT`: `$ANDROID_SDK_ROOT/ndk/26.2.11394342/`
  `ANDROID_SDK_ROOT` can be any directory (such as `~/android-sdk`).
   All of the Android build dependencies will be installed there.
 - Install the latest version of the [Android command-line
@@ -60,7 +58,7 @@ TODO: NDK 26c
   sudo $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install
    "build-tools;33.0.2" \
    "emulator" \
-   "ndk;25.2.9519653" \
+   "ndk;26.2.11394342" \
    "platform-tools" \
    "platforms;android-33" \
    "system-images;android-33;google_apis;x86_64"
--- a/components/script/Cargo.toml
+++ b/components/script/Cargo.toml
@@ -64,7 +64,7 @@ image = { workspace = true }
 indexmap = { workspace = true }
 ipc-channel = { workspace = true }
 itertools = { workspace = true }
-js = { package = "mozjs", git = "https://github.com/sagudev/mozjs", branch = "esr128", features = ["streams"] }
+js = { package = "mozjs", git = "https://github.com/servo/mozjs", features = ["streams"] }
 jstraceable_derive = { path = "../jstraceable_derive" }
 keyboard-types = { workspace = true }
 lazy_static = { workspace = true }
--- a/components/script/Cargo.toml
+++ b/components/script/Cargo.toml
@@ -64,7 +64,7 @@ image = { workspace = true }
 indexmap = { workspace = true }
 ipc-channel = { workspace = true }
 itertools = { workspace = true }
-js = { package = "mozjs", git = "https://github.com/servo/mozjs", features = ["streams"] }
+js = { package = "mozjs", git = "https://github.com/sagudev/mozjs", branch = "esr128", features = ["streams"] }
 jstraceable_derive = { path = "../jstraceable_derive" }
 keyboard-types = { workspace = true }
 lazy_static = { workspace = true }
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -70,7 +70,7 @@ jobs:
         uses: nttld/setup-ndk@v1
         id: setup-ndk
         with:
-          ndk-version: r25c
+          ndk-version: r26c
       - name: Setup Key Store for APK Signing
         env:
           KEYSTORE_BASE64: ${{ secrets.APK_KEYSTORE_BASE64 }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4337,7 +4337,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#d33454be74ec5b8d8faf51fab3ed477b8913898b"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#e618981f24a5105321301b33ab874632661846ed"
 dependencies = [
  "bindgen",
  "cc",
@@ -4351,7 +4351,7 @@ dependencies = [
 [[package]]
 name = "mozjs_sys"
 version = "0.128.0-0"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#d33454be74ec5b8d8faf51fab3ed477b8913898b"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#e618981f24a5105321301b33ab874632661846ed"
 dependencies = [
  "bindgen",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1385,7 +1385,7 @@ dependencies = [
  "diplomat_core",
  "proc-macro2",
  "quote",
- "syn 2.0.70",
+ "syn 2.0.72",
 ]
 
 [[package]]
@@ -1409,7 +1409,7 @@ dependencies = [
  "serde",
  "smallvec",
  "strck_ident",
- "syn 2.0.70",
+ "syn 2.0.72",
 ]
 
 [[package]]
@@ -4337,7 +4337,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#e618981f24a5105321301b33ab874632661846ed"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54"
 dependencies = [
  "bindgen",
  "cc",
@@ -4351,7 +4351,7 @@ dependencies = [
 [[package]]
 name = "mozjs_sys"
 version = "0.128.0-0"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#e618981f24a5105321301b33ab874632661846ed"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54"
 dependencies = [
  "bindgen",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4009,8 +4009,7 @@ dependencies = [
 [[package]]
 name = "libz-sys"
 version = "1.1.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c15da26e5af7e25c90b37a2d75cdbf940cf4a55316de9d84c679c9b8bfabf82e"
+source = "git+https://github.com/rust-lang/libz-sys.git?rev=f76b7a24ed0a71ef1b8bdac1895e8535cb2cb9c0#f76b7a24ed0a71ef1b8bdac1895e8535cb2cb9c0"
 dependencies = [
  "cc",
  "libc",
@@ -4337,7 +4336,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#3b7f5175d88d5f0aa078011f5e9dfe1bb2a1c22f"
 dependencies = [
  "bindgen",
  "cc",
@@ -4351,7 +4350,7 @@ dependencies = [
 [[package]]
 name = "mozjs_sys"
 version = "0.128.0-0"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54"
+source = "git+https://github.com/sagudev/mozjs?branch=esr128#3b7f5175d88d5f0aa078011f5e9dfe1bb2a1c22f"
 dependencies = [
  "bindgen",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4336,7 +4336,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#3b7f5175d88d5f0aa078011f5e9dfe1bb2a1c22f"
+source = "git+https://github.com/servo/mozjs#bf41ed080ec3c2fb84a53928c7db7691b0a0b161"
 dependencies = [
  "bindgen",
  "cc",
@@ -4350,7 +4350,7 @@ dependencies = [
 [[package]]
 name = "mozjs_sys"
 version = "0.128.0-0"
-source = "git+https://github.com/sagudev/mozjs?branch=esr128#3b7f5175d88d5f0aa078011f5e9dfe1bb2a1c22f"
+source = "git+https://github.com/servo/mozjs#bf41ed080ec3c2fb84a53928c7db7691b0a0b161"
 dependencies = [
  "bindgen",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4336,7 +4336,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/servo/mozjs#bf41ed080ec3c2fb84a53928c7db7691b0a0b161"
+source = "git+https://github.com/servo/mozjs#fb7782899503eb79801e67cf3c2fb819934ef509"
 dependencies = [
  "bindgen",
  "cc",
@@ -4349,8 +4349,8 @@ dependencies = [
 
 [[package]]
 name = "mozjs_sys"
-version = "0.128.0-0"
-source = "git+https://github.com/servo/mozjs#bf41ed080ec3c2fb84a53928c7db7691b0a0b161"
+version = "0.128.0-1"
+source = "git+https://github.com/servo/mozjs#fb7782899503eb79801e67cf3c2fb819934ef509"
 dependencies = [
  "bindgen",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,6 +589,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a12916984aab3fa6e39d655a33e09c0071eb36d6ab3aea5c2d78551f1df6d952"
 
 [[package]]
+name = "calendrical_calculations"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cec493b209a1b81fa32312d7ceca1b547d341c7b5f16a3edbf32b1d8b455bbdf"
+dependencies = [
+ "core_maths",
+ "displaydoc",
+]
+
+[[package]]
 name = "calloop"
 version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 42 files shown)


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/9911961223

sagudev:
Results are pretty good: https://github.com/sagudev/servo/actions/runs/9911961223/job/27388962921

some changes on `FinalizationRegistry`: OK [expected ERROR], but subcases fail (so no regression)

Regressions:

OK /html/semantics/scripting-1/the-script-element/module/crossorigin.html

    FAIL [expected PASS] subtest: Root module, Error in CORS-same-origin script

         assert_equals: Unexpected _log value expected "running,8-1" but got "running,8-0"

    FAIL [expected PASS] subtest: Imported module, Error in CORS-same-origin script

         assert_equals: Unexpected _log value expected "running,8-1" but got "running,8-0"

TIMEOUT [expected OK] /webmessaging/message-channels/close-event/garbage-collected.tentative.any.html

    TIMEOUT [expected FAIL] subtest: Entangled port is garbage collected, and the close event is fired.

        Test timed out

TIMEOUT [expected OK] /webmessaging/message-channels/close-event/garbage-collected.tentative.any.worker.html

    TIMEOUT [expected FAIL] subtest: Entangled port is garbage collected, and the close event is fired.

        Test timed out

I also expected https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers (apparently there are no wpt tests for it?)

sagudev:
NOTE-TO-SELF: Run debugmozjs in CI: https://github.com/servo/servo/issues/14759.

sagudev:
Android still fails: https://github.com/sagudev/servo/actions/runs/9913444136/job/27390478428#step:13:1242, but it works in mozjs, so maybe more changes are needed in mach. cc @mukilan

sagudev:
> Android still fails: https://github.com/sagudev/servo/actions/runs/9913444136/job/27390478428#step:13:1242, but it works in mozjs, so maybe more changes are needed in mach. cc @mukilan

@mukilan are you doing/willing to do android work or should somebody else do it?

mukilan:
Sorry, missed the notifications from this PR. I can look into the android part.

mukilan:
@sagudev Apologies, I accidentally pushed to your fork while trying to trigger the CI job on my fork.  Shall I force push without my commits?

sagudev:
> @sagudev Apologies, I accidentally pushed to your fork while trying to trigger the CI job on my fork. Shall I force push without my commits?

It's okay, you can push to my branch/fork as I am not currently doing any work here.

mukilan:
So [2460979](https://github.com/servo/servo/pull/32769/commits/2460979aed69565f0f160c98bf27f2a4203b2f9c) fixes the Android build on both the [CI](https://github.com/mukilan/servo/actions/runs/10088387519/job/27894081663) and my NixOS system. However, I'm not sure if I understand the underlying issue correctly. On NixOS, without the fixes for setting `LIBZ_SYS_STATIC=1` and the `cargo:rustc-link-lib=static=z` in build.rs, we get an error in the final link step about undefined references.

```
  = note: ld.lld: error: undefined hidden symbol: deflateEnd
          >>> referenced by Compression.cpp:54 (/home/mukilan/.cargo/git/checkouts/mozjs-953f58dcd23dbdae/e618981/mozjs-sys/mozjs/js/src/vm/Compression.cpp:54)
          >>>               Unified_cpp_js_src15.o:(js::Compressor::~Compressor()) in archive /home/mukilan/dev/servo.git/servo-pr/target/aarch64-linux-android/release/deps/libmozjs_sys-448477fcab6f6851.rlib

          ld.lld: error: undefined hidden symbol: deflateInit2_
          >>> referenced by Compression.cpp:75 (/home/mukilan/.cargo/git/checkouts/mozjs-953f58dcd23dbdae/e618981/mozjs-sys/mozjs/js/src/vm/Compression.cpp:75)
          >>>               Unified_cpp_js_src15.o:(js::Compressor::init()) in archive /home/mukilan/dev/servo.git/servo-pr/target/aarch64-linux-android/release/deps/libmozjs_sys-448477fcab6f6851.rlib

          ld.lld: error: undefined hidden symbol: deflate
          >>> referenced by Compression.cpp:116 (/home/mukilan/.cargo/git/checkouts/mozjs-953f58dcd23dbdae/e618981/mozjs-sys/mozjs/js/src/vm/Compression.cpp:116)
          >>>               Unified_cpp_js_src15.o:(js::Compressor::compressMore()) in archive /home/mukilan/dev/servo.git/servo-pr/target/aarch64-linux-android/release/deps/libmozjs_sys-448477fcab6f6851.rlib
```

I set `LIBZ_SYS_STATIC=1` because otherwise build scripts of `libz-sys` [adds directives](https://github.com/rust-lang/libz-sys/blob/37d0a5dd846ae72580ce4fd018b76737d446dad5/build.rs#L42) to use libz from the host, but perhaps because we are cross-compiling, the include and search lib paths are prefixed with the sysroot (e.g `/nix/store../android-sdk/ndk/sysroot/nix/store/../libz-1.3.1`, so effectively uses an invalid path.

I also needed the  `cargo:rustc-link-lib=static=z` because when rustc invokes clang for the final link step, the `-lz` added by `libz-sys` ends up towards the end of the command line. By trial and error, I found that moving it towards the beginning fixed the error. For some reason, linking it as shared library doesn't fix the undefined reference issue.

Interestingly, the CI fails without the libz fixes, but it failing during the configure step of mozjs (same error that you shared) instead of a linker error.
```
  checking whether the linker supports Identical Code Folding... no
  checking whether removing dead symbols breaks debugging... no
  checking for valid optimization flags... no
  configure: error: These compiler flags are invalid: -O3
```

I'll try to figure out the exact reason, but let me know if you are able to make sense of the above issues.

Note: The NDK does have the libz and is part of the system search paths for the linker and header includes.

sagudev:
> I set LIBZ_SYS_STATIC=1 because otherwise the build script of libz-sys [adds directives](https://github.com/rust-lang/libz-sys/blob/37d0a5dd846ae72580ce4fd018b76737d446dad5/build.rs#L42) to use libz from the host, but perhaps because we are cross-compiling, the include and search lib paths are prefixed with the sysroot (e.g /nix/store../android-sdk/ndk/sysroot/nix/store/../libz-1.3.1, so effectively uses an invalid path.

This sounds like libz-sys problem, maybe we can isolate it and fix it upstream.

> I also needed the cargo:rustc-link-lib=static=z because when rustc invokes clang for the final link step, the -lz added by libz-sys ends up towards the end of the command line. By trial and error, I found that moving it towards the beginning fixed the error.

This is very weird, maybe there is some escaping problem or something, could you share whole invocation.

Also what is weird is that android CI in mozjs works (although we only run cargo build which does not trigger linker step).

mukilan:
> This is very weird, maybe there is some escaping problem or something, could you share whole invocation.

[This](https://gist.github.com/mukilan/66f0138cf0146e8864056da5685aa724) is the invocation without the link-lib directive in servoshell/build.rs. When we add the directive, we get an additional `-lz` after the `"-lEGL" "-Wl,-Bstatic"`

mukilan:
I think the issue might in mozjs, specifically in how zlib is included and compiled.

I'm able to reproduce the issue I'm seeing with:
```
//include <zlib.h>

__attribute__ ((visibility("hidden"))) extern int inflateInit_(void*, const char*, int);
__attribute__ ((visibility("hidden"))) extern int inflateInit2_(void*, int, const char*, int);

void test() {
   inflateInit_(0, 0, 0);
   inflateInit2_(0, 0, 0, 0);
}
```

I get the same 'undefined hidden symbol' error when trying to link the above with both static and dynamic libz. I don't see any error if the visibilty is changed to 'default' or we use the zlib header directly. Although, it is curious why static linking doesn't work here even though it does in servo (i.e the rust-link-lib=static=z directive).

sagudev:
> I think the issue might in mozjs, specifically in how zlib is included and compiled.
> 
> I'm able to reproduce the issue I'm seeing with:
> 
> ```
> //include <zlib.h>
> 
> __attribute__ ((visibility("hidden"))) extern int inflateInit_(void*, const char*, int);
> __attribute__ ((visibility("hidden"))) extern int inflateInit2_(void*, int, const char*, int);
> 
> void test() {
>    inflateInit_(0, 0, 0);
>    inflateInit2_(0, 0, 0, 0);
> }
> ```
> 
> I get the same 'undefined hidden symbol' error when trying to link the above with both static and dynamic libz. I don't see any error if the visibilty is changed to 'default' or we use the zlib header directly. Although, it is curious why static linking doesn't work here even though it does in servo (i.e the rust-link-lib=static=z directive).

Where exactly is this in mozjs? https://github.com/sagudev/mozjs/tree/e618981f24a5105321301b33ab874632661846ed does not have any manual declarations by my search.

mukilan:
Sorry, I can clarify. I'm not saying we have literally declared them using the `__attribute__`.  I was just pointing out that the error 'undefined hidden symbol' seems to indicate that *somehow* during compilation of mozjs, the 'zlib' symbols in mozjs have the same effect as if they were declared with 'hidden visiblity'. This is also confirmed by looking at the contents of the libmozjs_sys.rlib that rust produces using readelf:

```
[nix-shell:~/dev]$ readelf -a libmozjs_sys-448477fcab6f6851/Unified_cpp_js_src15.o  | grep inflate
000000000050  031a0000011b R_AARCH64_CALL26  0000000000000000 inflateInit_ + 0
000000000064  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
00000000006c  031c0000011b R_AARCH64_CALL26  0000000000000000 inflateEnd + 0
000000000090  031e0000011b R_AARCH64_CALL26  0000000000000000 inflateInit2_ + 0
0000000000d8  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
000000000114  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
00000000013c  031c0000011b R_AARCH64_CALL26  0000000000000000 inflateEnd + 0
   794: 0000000000000000     0 NOTYPE  GLOBAL HIDDEN   UND inflateInit_
   795: 0000000000000000     0 NOTYPE  GLOBAL HIDDEN   UND inflate
   796: 0000000000000000     0 NOTYPE  GLOBAL HIDDEN   UND inflateEnd
   798: 0000000000000000     0 NOTYPE  GLOBAL HIDDEN   UND inflateInit2_
```

sagudev:
Try https://github.com/sagudev/mozjs/commit/ff754c5235b6711b4995b861223cff64379e9b7d of mozjs (main branch on sagudev/mozjs)

mukilan:
Yes,  [sagudev/mozjs@ff754c5](https://github.com/sagudev/mozjs/commit/ff754c5235b6711b4995b861223cff64379e9b7d) does work without the libz related hacks!


Also, I found that SM [forces](https://github.com/servo/mozjs/blob/6e4e55a002f493739c8dbe0f2d2b29b0017d5dd2/mozjs-sys/mozjs/build/moz.configure/toolchain.configure#L2170) all declarations to be [hidden by default](https://github.com/servo/mozjs/blob/6e4e55a002f493739c8dbe0f2d2b29b0017d5dd2/mozjs-sys/mozjs/config/gcc_hidden.h#L6), including [extern](https://gcc.gnu.org/wiki/Visibility) declarations.

```
#pragma GCC visibility is stronger than -fvisibility; it affects extern declarations as well. -fvisibility only affects definitions, so that existing code can be recompiled with minimal changes. This is more true for C than C++; C++ interfaces tend use classes, which are affected by -fvisibility. 
```

With `ff754c5` the mozjs rlib now contains 'default' visibility references to the zlib symbols.

```
readelf -a ./target/aarch64-linux-android/release/deps/libmozjs_sys-1280b33408e3d35a.rlib | grep inflate
000000000050  031a0000011b R_AARCH64_CALL26  0000000000000000 inflateInit_ + 0
000000000064  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
00000000006c  031c0000011b R_AARCH64_CALL26  0000000000000000 inflateEnd + 0
000000000090  031e0000011b R_AARCH64_CALL26  0000000000000000 inflateInit2_ + 0
0000000000d8  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
000000000114  031b0000011b R_AARCH64_CALL26  0000000000000000 inflate + 0
00000000013c  031c0000011b R_AARCH64_CALL26  0000000000000000 inflateEnd + 0
   794: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND inflateInit_
   795: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND inflate
   796: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND inflateEnd
   798: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND inflateInit2_
```

I still can't find documentation that explains the linker's behaviour in this scenario with hidden symbols, so I don't yet understand how it works now with `ff754c5`. But I believe it should be safe to proceed with `ff754c5` and no libz hacks in Servo, WDYT? If you agree, I can rebase and force push just the android changes after you push `ff754c5` to the esr-128 branch in mozjs?

sagudev:
> But I believe it should be safe to proceed with ff754c5 and no libz hacks in Servo, WDYT?

Sounds good to me.

> If you agree, I can rebase and force push just the android changes after you push ff754c5 to the esr-128 branch in mozjs?

Okay, I pushed changes to esr-128 branch (commit hash is different due to rebase), so you can proceed.

mukilan:
Hmm, the Android build still [fails](https://github.com/mukilan/servo/actions/runs/10111750241) on the CI due to that wierd configure error. I don't see it on my NixOS setup and I'm able to run the APK on a real device.

I guess it is unrelated to the zlib issue I was seeing locally, so I'll have to debug it. I've pushed my changes to your branch anyway.

sagudev:
Hm: https://github.com/mukilan/servo/actions/runs/10111750241/job/27964320069#step:13:1116

I think it's caused by `-I/usr/include`, and this might be from https://github.com/sagudev/mozjs/blob/15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54/mozjs-sys/build.rs#L309, but there are no such problems in mozjs CI runs.

sagudev:
> Hm: https://github.com/mukilan/servo/actions/runs/10111750241/job/27964320069#step:13:1116
> 
> I think it's caused by `-I/usr/include`, and this might be from https://github.com/sagudev/mozjs/blob/15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54/mozjs-sys/build.rs#L309, but there are no such problems in mozjs CI runs.

`-I/usr/include` indeed comes from libz-sys: https://github.com/sagudev/servo/actions/runs/10112461769/job/27966627545#step:13:4818, probably because there is zlib installed on system (but not on nix per d4727c1f555f6a8856d7b096e92d8a4b99d5ca06): https://github.com/rust-lang/libz-sys/blob/37d0a5dd846ae72580ce4fd018b76737d446dad5/build.rs#L39 (android check happens latter).

sagudev:
Although pkg-config should work also in cross too: https://docs.rs/pkg-config/latest/pkg_config/#cross-compilation (we are likely setting wrong env for it)

sagudev:
Still some errors in post build steps (maybe because I do not have keystore secret in fork): https://github.com/sagudev/servo/actions/runs/10113142896/job/27968824692#step:11:29385

mukilan:
> > Hm: https://github.com/mukilan/servo/actions/runs/10111750241/job/27964320069#step:13:1116
> > I think it's caused by `-I/usr/include`, and this might be from https://github.com/sagudev/mozjs/blob/15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54/mozjs-sys/build.rs#L309, but there are no such problems in mozjs CI runs.
> 
> `-I/usr/include` indeed comes from libz-sys: https://github.com/sagudev/servo/actions/runs/10112461769/job/27966627545#step:13:4818, probably because there is zlib installed on system (but not on nix per [d4727c1](https://github.com/servo/servo/commit/d4727c1f555f6a8856d7b096e92d8a4b99d5ca06)): https://github.com/rust-lang/libz-sys/blob/37d0a5dd846ae72580ce4fd018b76737d446dad5/build.rs#L39 (android check happens latter).

Are you referring to the `cargo:include=/usr/include` from libz-sys's build.rs? Will that directive cause cargo to modify CFLAGS env var? I couldn't find that directive in the official reference, so I don't understand how this influences mozjs-sys's configure script.

sagudev:
> Are you referring to the `cargo:include=/usr/include` from libz-sys's build.rs?

Yes

> Will that directive cause cargo to modify CFLAGS env var?

Also yes.

> I couldn't find that directive in the official reference, so I don't understand how this influences mozjs-sys's configure script.

build.rs scripts can set custom env vars for other deps to use in own build scripts per https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages: 

> Other -sys packages can take advantage of the DEP_NAME_KEY=value environment variables to better integrate with other packages. See the [“Using another sys crate”](https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate) example.

 And linked example actually shows how this is done/used in libz-sys: https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate

We do add this includes to c(pp)flags: https://github.com/servo/mozjs/pull/474/commits/15db7f6d2b4b3a7cbf7ba1a434cbb368ec409f54#diff-c7e783cd7cc2c1d816b8bd95ddd758efea220488734993e0da28dd3f41d2e5a5R308-R310

mukilan:
Oh, wow! Thats interesting. Thanks for the explanation.  I didn't realize DEP_Z_INCLUDE related to the cargo:include directive, I assumed it is there for the user to override if needed.

This also explains why setting `LIBZ_SYS_STATIC=1` helped yesterday on the [CI](https://github.com/servo/servo/pull/32769#issuecomment-2249844099), since that was also done to bypass the pkg-config stuff.

mukilan:
I did some more digging just to understand why [sagudev/mozjs@ff754c5](https://github.com/sagudev/mozjs/commit/ff754c5235b6711b4995b861223cff64379e9b7d) fixed the linker issue.

1. As noted [here](https://github.com/servo/servo/pull/32769#issuecomment-2252631806), SM build system includes gcc_hidden.h unconditionally for all translation units, so the `#pragma GCC visibility push(hidden)`  makes all declarations, including extern declaration from system headers like `zlib.h`, into `HIDDEN` symbols references in the resulting object file.
2. The SM build system also has a mechanism to [wrap system headers](https://github.com/servo/mozjs/blob/b195e1261edf029d718c728175b7686ea95bc48a/mozjs-sys/mozjs/config/make-system-wrappers.py#L8) with `#pragma GCC visibilty push(default)` & `#pragma GCC visibility pop` to force the correct DEFAULT visibility for the symbols declared within the wrapped system headers.
3. These system wrappers are [auto-generated](https://github.com/servo/mozjs/blob/b195e1261edf029d718c728175b7686ea95bc48a/mozjs-sys/mozjs/config/moz.build#L96) by the SM build system, but it seems like we need to explicitly list the headers.
4. For `zlib.h`, whether it was wrapped or not was controlled by the [MOZ_SYSTEM_ZLIB](https://github.com/servo/mozjs/blob/b195e1261edf029d718c728175b7686ea95bc48a/mozjs-sys/mozjs/config/system-headers.mozbuild#L1272C12-L1272C27) configuration variable, which  [sagudev/mozjs@ff754c5](https://github.com/sagudev/mozjs/commit/ff754c5235b6711b4995b861223cff64379e9b7d) sets to True unconditionally.

Also, found some [documentation](https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html) that could explain the linker error (HIDDEN symbol references in mozjs were not resolved by definitions in zlib.so):

> First, all of the non-default visibility attributes, when applied to a symbol reference, imply that a definition to satisfy that  reference must be provided within the current executable or shared object. If such a symbol reference has no definition  within the component being linked, then the reference must have STB_WEAK binding and is resolved to zero.

sagudev:
Interesting!

README.md
```diff
@@ -45,6 +45,8 @@ For more detailed build instructions, see the Servo book under [Setting up your
 
 ### Android
 
+TODO: NDK 26c
+
 - Ensure that the following environment variables are set:
   - `ANDROID_SDK_ROOT`
   - `ANDROID_NDK_ROOT`: `$ANDROID_SDK_ROOT/ndk/25.2.9519653/`
```
sagudev, line 52: @mukilan could you also fix readme here

components/script/Cargo.toml
```diff
@@ -64,7 +64,7 @@ image = { workspace = true }
 indexmap = { workspace = true }
 ipc-channel = { workspace = true }
 itertools = { workspace = true }
-js = { package = "mozjs", git = "https://github.com/servo/mozjs", features = ["streams"] }
+js = { package = "mozjs", git = "https://github.com/sagudev/mozjs", branch = "esr128", features = ["streams"] }
```
sagudev, line 67: before landing this needs to be reverted

README.md
mukilan: Done in [00b4235](https://github.com/servo/servo/pull/32769/commits/00b423578d8a8d5ac3715ea53d64405552061626)

sagudev: Tnx

jdm (approved):
Thank you!

sagudev:
Something is wrong with windows artifact: https://github.com/sagudev/servo/actions/runs/10134348491/job/28020479847#step:12:580
`MOZJS_FROM_SOURCE=1` fixes CI: https://github.com/sagudev/servo/actions/runs/10134353999
archive test in mozjs run successfully: https://github.com/servo/mozjs/actions/runs/10133207597/job/28018302410.

cc @wusyong

wusyong:
@sagudev I tested on my local Windows device, and everything is working. Maybe it's because Servo CI doesn't run on `windows-latest` as Mozos CI does; instead, it uses `windows-2022`. This is probably because it doesn't use the latest Visual Studio library.

sagudev:
`windows-latest` = `windows-2022` per https://github.com/actions/runner-images/tree/main

After a small research, I think the problem might be in stripping, so I am doing test run for that https://github.com/sagudev/mozjs/actions/runs/10138127855

sagudev:
> `windows-latest` = `windows-2022` per https://github.com/actions/runner-images/tree/main
> 
> After a small research, I think the problem might be in stripping, so I am doing test run for that https://github.com/sagudev/mozjs/actions/runs/10138127855

It works: https://github.com/sagudev/servo/actions/runs/10138407681/job/28030063629

sagudev:
I would like to thank everyone who helped me on this SM bump.

PR #32772, opened 2024-07-13
Use mallinfo only on target_env=gnu

ptrcnull:
mallinfo isn't available on musl, causing linking issues on build;
make sure related functions are built only for GNU Libc

---

- [X] These changes do not require tests because: this is ultimately a build system fix and a "proper" test for this would be musl CI/CD, or something


ptrcnull:
updated with the proper author name for the sign-off

jdm (approved):

PR #32777, opened 2024-07-15
Fix sizing of replaced elements with min/max sizes

### LINKED ISSUE #32776: `<img>` elements with `width` and `height` attributes have incorrect aspect ratio with `height: auto; max-height: [...]`
**Describe the bug:**

When an `` element has `width` and `height` attributes, a `height: auto` CSS property, and a `max-height: [some value]` CSS property, it *should* maintain its intrinsic aspect ratio. However, it does not in Servo:

| Firefox | Servo |
| - | - |
| image | image |

This is probably due to some [suspicious clamping](https://github.com/servo/servo/blob/e761b7d7111e4826467727f354ae8c18893ab38a/components/layout_2020/replaced.rs#L406) performed in the replaced element sizing algorithm--the [CSS spec](https://drafts.csswg.org/css2/#min-max-widths) says that if the tentative width falls outside the range of `min-width` and `max-width` (or the same for the height), the layout calculation should be redone with the clamped values. The re-calculation allows the aspect ratio to be calculated correctly. However, the current code seems to only perform the layout calculation once and then clamp the width and height afterwards.

**To Reproduce:**

<details>
<summary>Test case HTML</summary>

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <title>CSS image height: auto test</title>
    <style>
        img {
            height: auto;
            max-width: 240px;
        }
    </style>
</head>
<body>
    <img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACQAAAAYCAMAAAClZq98AAABZGlDQ1BpY2MAAHjalZEzoFdxFIC/q2y75yW7Lds2r9/f5pbGbJtzjXHM9hqX7G79snGm71ggb/LrgZjaEwLBeHT0gN4FEydNLqh4m4pUQ6M1LWbpsXCvkSOH8lt5ehEJ4FwbUYv/k+qGGdNBKgB66uFoHKSZQNdUPCx4FVBPd2YZIO0DWkffDwjSZWGf/ZHvC7YFIyM4OnZ0H5DrAQX2Nzz7G9adaADkzkCJEQgaIE/8yIbgvOCAP6EDABJQzQyOGyPsQDMGMIhhjKSA2SRw8ROnDS5BCogxmgH0/k1+E5HPSBLMxo+LTgF9CRFglsgH8YMfbxuzOnUEQKrWG7SbnveoFCouhTeLPO/VZs97swWU63Ao+DU/tAm6PwFl0VdbyUaoORf2H/tqm70SDiyARlfCs6KzAFAA2bLg4U6oPgnqnoYqU8TdvvjZegnG5mHoCVi9BspsqDn1N3tX+vZuf4z5dL93knFys+SXUE8AAAAJcEhZcwAALiMAAC4jAXilP3YAAAAHdElNRQfoBw8PKgC4t3QcAAAAGXRFWHRDb21tZW50AENyZWF0ZWQgd2l0aCBHSU1QV4EOFwAAAAlQTFRFAAAA/////wAAzV63nAAAAIBJREFUeNqF0QEKAyEQQ1H99z90KYb9aBAFbCe8MMKOdRgeMhmU4WJgS8p4Xw296mmcJS7rTi+f6+Q3gfF3Jv8rxWmCBFb+N0HIYywHRfMhSOV4kyidoKOG64LskYRxR/jARhxvKtQf23+wh5zGSLOPkjLOEsM242lgFKI7L8Nufv6zAR1Q0qgmAAAAAElFTkSuQmCC" width="480" height="480">
</body>
</html>
```
</details>

**Platform:**
Reproduced on Fedora 40 x86_64; almost certainly platform-agnostic

valadaptive:
If a (min/max)-(height/width) property is set, we still need to respect the intrinsic ratio of the element if it exists. The previous code was simply clamping the element size after doing the sizing calculations once, but this leads to an incorrect aspect ratio.

Now, we clamp to the maximum size and redo the calculations, then do the same with the minimum size.

I'm not sure if we need to do it in two steps, as the [CSS spec does](https://drafts.csswg.org/css2/#min-max-widths), or if we can just clamp between the min and max sizes in one step (possibly even before calculating the tentative size in the first place), but to avoid any unforeseen edge cases, I've opted to take the spec literally here.

---

- [x] These changes fix #32776 (GitHub issue number if applicable)

- [x] There are tests for these changes (It appears WPT has several tests that cover this behavior)

(What kind of test should I add? The [automated testing section of the book](https://book.servo.org/hacking/testing.html) mentions wpt, but am I supposed to put "home-grown" tests in there too, or is it supposed to be just for the official wpt test suite?)

### DIFF
--- a/components/layout_2020/replaced.rs
+++ b/components/layout_2020/replaced.rs
@@ -17,6 +17,7 @@ use servo_arc::Arc as ServoArc;
 use style::properties::ComputedValues;
 use style::servo::url::ComputedUrl;
 use style::values::computed::image::Image as ComputedImage;
+use style::values::generics::length::GenericLengthPercentageOrAuto;
 use style::values::CSSFloat;
 use style::Zero;
 use url::Url;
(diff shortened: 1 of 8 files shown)


jdm:
>What kind of test should I add? The [automated testing section of the book](https://book.servo.org/hacking/testing.html) mentions wpt, but am I supposed to put "home-grown" tests in there too, or is it supposed to be just for the official wpt test suite?)

If there are not already test cases covering this behavior in the official WPT suite, we should add one to it.

valadaptive:
Ah, so should I just wait for the WPT run to complete and see if anything turns green?

mrobinson:
> Ah, so should I just wait for the WPT run to complete and see if anything turns green?

Yep! Let's do that. That will also give you a pointer to which tests you can run locally to continue testing your changes, if necessary.

valadaptive:
Fixed what was causing the last set of tests to fail; let's see how it does now

valadaptive:
Looks like this fixes several WPT tests!

mrobinson (approved):

Issue #14123, opened 2016-11-08
Implement column flexbox


stshine:
To be able to have a column flex container we need to:
1. Mark flex item as in a row container or a column container
2. explicitly reflow flex item with automatic block size before others
3. make flex item have their block size set by the parent container.

Luckily most obstacles have been removed by recent development, especially #13870 and #13401 .

stshine:
Please assign this issue to me.

PR #32785, opened 2024-07-16
layout: Properly handle min/max cross container size

delan:
We need to constrain the available cross size by the min/max cross size properties, but this was not yet implemented. This patch plumbs those in from the flex container’s own containing block.

There are three new passing tests, two new passing subtests, and two subtests regressing due to flex column layout not being implemented (#14123).

Co-authored-by: Delan Azabani <dazabani@igalia.com>

---


mrobinson (approved):

delan:
Thanks for fixing the expectations, I missed that one in my commit because new files are untracked.

PR #32800, opened 2024-07-17
Implement the `aspect-ratio` property for replaced elements

valadaptive:
First step towards fully implementing `aspect-ratio`. This is probably the easy part since replaced elements are already known to have an intrinsic aspect ratio and treated as such.

This PR touches the same code as #32777 and therefore depends on it.

This will require the `aspect-ratio` property to be enabled for layout 2020 in Stylo--I currently have it pointed to my fork which does so.

---

### DIFF
--- a/components/layout_2020/style_ext.rs
+++ b/components/layout_2020/style_ext.rs
@@ -6,7 +6,7 @@ use app_units::Au;
 use style::computed_values::mix_blend_mode::T as ComputedMixBlendMode;
 use style::computed_values::position::T as ComputedPosition;
 use style::computed_values::transform_style::T as ComputedTransformStyle;
-use style::logical_geometry::WritingMode;
+use style::logical_geometry::{Direction, WritingMode};
 use style::properties::longhands::backface_visibility::computed_value::T as BackfaceVisiblity;
 use style::properties::longhands::box_sizing::computed_value::T as BoxSizing;
 use style::properties::longhands::column_span::computed_value::T as ColumnSpan;
@@ -15,7 +15,9 @@ use style::values::computed::image::Image as ComputedImageLayer;
 use style::values::computed::{Length, LengthPercentage, NonNegativeLengthPercentage, Size};
 use style::values::generics::box_::Perspective;
 use style::values::generics::length::MaxSize;
+use style::values::generics::position::{GenericAspectRatio, PreferredRatio};
 use style::values::specified::{box_ as stylo, Overflow};
+use style::values::CSSFloat;
 use style::Zero;
 use webrender_api as wr;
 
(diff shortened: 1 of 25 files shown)


valadaptive:
This still needs a bit more work--I need to enable the aspect-ratio WPT tests and fix a bug relating to max-width and border-box

valadaptive:
Turns out the "bug" was me forgetting to rebuild.

On the plus side, I did create some web platform tests (WPT tests? Has that "ATM machine" ring to it) for testing the behavior of `aspect-ratio` in combination with `box-sizing: border-box`, which seems to have been missing from the existing test suite.

Most of the existing `aspect-ratio` tests seem to be single-file ones, but there are a lot of different test cases and I don't want to create 24 individual .html files, so I used the layout test harness.

Not sure why the WPT Export step is failing--doesn't seem to be anything related to the contents of the tests themselves.

mrobinson:
> Not sure why the WPT Export step is failing--doesn't seem to be anything related to the contents of the tests themselves.

This might be related to the commit from #32803 somehow. Hopefully once you rebase it will improve, but I'll watch it.

valadaptive:
Seems to have been caused by force-pushing while the initial WPT sync action was running; I believe it tried to create the PR twice?

valadaptive:
About the regressions: replaced-element-024 and replaced-element-026 regress because we don't implement `object-fit`, and replaced-element-028 regresses because we don't treat non-available ``s with `alt` attributes properly.

components/layout_2020/style_ext.rs
```diff
@@ -535,6 +578,77 @@ impl ComputedValuesExt for ComputedValues {
...
+            ratio: mut preferred_ratio,
+        } = self.clone_aspect_ratio();
+
+        // For all cases where a ratio is specified:
+        // "If the <ratio> is degenerate, the property instead behaves as auto."
+        if let PreferredRatio::Ratio(ratio) = preferred_ratio {
+            if ratio.is_degenerate() {
+                preferred_ratio = PreferredRatio::None;
+            }
+        }
```
mrobinson, line 600:
```suggestion
        if matches!(preferred_ratio,  PreferredRatio::Ratio(ratio) if ratio.is_degenerate()) {
            preferred_ratio = PreferredRatio::None;
        }
```

components/layout_2020/style_ext.rs
```diff
@@ -535,6 +578,77 @@ impl ComputedValuesExt for ComputedValues {
         false
     }
 
+    /// Resolve the preferred aspect ratio according to the given natural aspect
+    /// ratio and the `aspect-ratio` property.
+    /// https://drafts.csswg.org/css-sizing-4/#aspect-ratio
```
mrobinson, line 583: Raw URLS in rustdoc comments lead to rustdoc warnings:

```suggestion
    /// <https://drafts.csswg.org/css-sizing-4/#aspect-ratio>
```

components/layout_2020/style_ext.rs
```diff
@@ -143,6 +145,42 @@ impl PaddingBorderMargin {
...
+/// Resolved `aspect-ratio` property with respect to a specific element. Depends
+/// on that element's `box-sizing` (and padding and border, if that `box-sizing`
+/// is `border-box`).
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct AspectRatio {
+    /// If the element that this aspect ratio belongs to uses box-sizing:
+    /// border-box, and the aspect-ratio property does not contain "auto", then
+    /// the aspect ratio is in respect to the border box. This will then contain
+    /// the summed sizes of the padding and border. Otherwise, it's 0.
+    pb_adjust: LogicalVec2<Au>,
```
mrobinson, line 157: Let's just calling this `box_sizing_adjustment`.

mrobinson (approved):
Excellent work. I have a few nits, but I'll apply them myself and try to land this.

Issue #32802, opened 2024-07-18
Implement `aspect-ratio` property


valadaptive:
**Describe the bug:**

Implement the CSS `aspect-ratio` property, described at https://drafts.csswg.org/css-sizing-4/#aspect-ratio.

Current status:
- [x] Enable the `aspect-ratio` property in Stylo

- [ ] Enable the `aspect-ratio` WPT suite
- [ ] Implement for replaced elements #32800
- [ ] Implement for non-replaced elements

freyacodes:
I have been reading through the layout component with an eye towards implementing aspect-ratio in non-replaced elements. Would you be open to me taking a stab at implementing that?

mrobinson:
Hi @freyacodes, thanks for your interest! Layout is currently undergoing a lot of architectural work, so the main folks working on layout are going to have limited to review and mentor a big project like this. I would suggest maybe picking something more in the realm of DOM / Script which has more reviewers and has a greater deal of abstraction between parts.

PR #32803, opened 2024-07-18
Bump stylo and enable `aspect-ratio` tests

valadaptive:
Bumps stylo to the latest tip of the 2024-05-31 branch to enable the `aspect-ratio` property, then enables the WPT tests for it under `css/css-sizing/aspect-ratio`.

Not sure if dependency bumps should be part of pull requests here; let me know if that's an issue.

---

- [ ] These changes are part of #32802


mrobinson (approved):

valadaptive:
Looks like I didn't run some WPTs that ended up passing just because `aspect-ratio` is now enabled. I think I got everything; I'll do another try run on my fork to make sure

valadaptive:
Looks like it's good now

PR #32812, opened 2024-07-19
Upgrade stylo to 2024-07-16

mrobinson:
This is based on https://github.com/servo/servo/pull/32606

Changelog:

   - Servo fixups: https://github.com/servo/stylo/compare/f2fbf5520e9fe1f45d0fe868c3f8e2d100b6d3a8...fafff4ca2cc971acc5278977ee69d6351cfbbaca
   - In addition, all changes are now squashed into a single commit and many unnecessary differences from upstream are eliminated.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1291,7 +1291,7 @@ dependencies = [
 [[package]]
 name = "derive_common"
 version = "0.0.1"
-source = "git+https://github.com/servo/stylo?branch=upgrade-squash#ec9d7fb1b9ac5cb617e8602087606e72f89b5ef1"
+source = "git+https://github.com/servo/stylo?branch=2024-07-16#eceb838e855e3b73a28df602b69494899f6d168b"
 dependencies = [
  "darling",
  "proc-macro2",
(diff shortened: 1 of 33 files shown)


nicoburns (approved):
Looks good to me. We'll want to rename that `dom` crate (in the `Cargo.toml`), but we want to rename a bunch of crates, so we might as well do that all at once.

mrobinson:
> Looks good to me. We'll want to rename that `dom` crate (in the `Cargo.toml`), but we want to rename a bunch of crates, so we might as well do that all at once.

The `dom` crate is already renamed to `style_dom` or did you mean something else here?

nicoburns:
@mrobinson

> The dom crate is already renamed to style_dom or did you mean something else here?

I meant that the crate itself should be renamed in the stylo repo (I'm assuming you'll want to do this upstream but it will block a crates.io release)

mrobinson:
> I meant that the crate itself should be renamed in the stylo repo (I'm assuming you'll want to do this upstream but it will block a crates.io release)

I don't think we can rename the crate upstream as that's the name that Gecko is using for all of it's DOM-related Rust code. Maybe we can convince them to use a different name, but they might go with something like `mozilla_dom` then, which would be even stranger for us. I understand that this makes things complicated for a future crates.io release though.

PR #32814, opened 2024-07-19
Remove failure breaks from webgpu thread

### LINKED ISSUE #32621: webgpu thread exits if requestDevice/requestAdapter fails
The main loop is `break`ed at:

https://github.com/servo/servo/blob/a730469b704878da9b484b0018f369438ce08851/components/webgpu/wgpu_thread.rs#L635

https://github.com/servo/servo/blob/a730469b704878da9b484b0018f369438ce08851/components/webgpu/wgpu_thread.rs#L688

sagudev:
Only breaking of main loop for thread exit is allowed.

---

- [x] These changes fix #32621

### DIFF
--- a/components/webgpu/wgpu_thread.rs
+++ b/components/webgpu/wgpu_thread.rs
@@ -694,10 +694,6 @@ impl WGPU {
                             })
                             .map_err(|e| e.to_string());
 
-                        if response.is_err() {
-                            break;
-                        }
-
                         if let Err(e) = sender.send(WebGPUResponse::Adapter(response)) {
                             warn!(
                                 "Failed to send response to WebGPURequest::RequestAdapter ({})",
@@ -733,7 +729,7 @@ impl WGPU {
                                     e
                                 )
                             }
-                            break;
+                            continue;
                         }
                         let device = WebGPUDevice(device_id);
                         let queue = WebGPUQueue(queue_id);


sagudev:
there are failures in `webgpu:api,operation,adapter,requestDevice:limit,out_of_range:*`, but before it was timeout, so it's still improvement. We need to raise TypeError in some cases of errors instead of only OperationError, but that will be done as a follow up, as we also need to get wgpu errors on js side (wgpu errors need to be serializable, there is already PR in wgpu for that).

components/webgpu/wgpu_thread.rs
```diff
@@ -694,10 +694,6 @@ impl WGPU {
                             })
                             .map_err(|e| e.to_string());
 
-                        if response.is_err() {
-                            break;
-                        }
-
```
mrobinson, line 700: Would it make sense to match this error response and `warn!()` as we do when the send immediately following it fails?

mrobinson (approved):

components/webgpu/wgpu_thread.rs
sagudev: No, because we send the error (or Ok(...)) to script (dom), where error get's handled (promise reject), while IPC errors does not get handled at all (apart from printing warning).

PR #32817, opened 2024-07-20
Enable OpenXR backend from the WebXR crate

### LINKED ISSUE #32809: Enabling WebXR OpenXR backend in servoshell
Currently servoshell uses the glwindow backend by default. Could we enable the OpenXR backend as well? I've done this locally while working on the WebXR crate so far, so the main changes would be:

- Controlling the selected backend by pref (`dom.webxr.glwindow.enabled` or `dom.webxr.openxr.enabled`), with priority still given to GlWindow if desired.
- Making the XR discovery object into an enum for glwindow or openxr

The one other consideration for it (which I haven't addressed locally) is that it would only be for Windows as long as the OpenXR backend relies on D3D11.

msub2:
This enables the OpenXR backend from the WebXR crate, allowing Servo to display WebXR content on actual headsets as opposed to just the GL window. For now it is gated by a pref and the GL window remains the default.

---

- [x] These changes fix #32809

- [x] These changes do not require tests because they involve rendering to external devices, which I have verified is working as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4450,6 +4450,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3fdb12b2476b595f9358c5161aa467c2438859caa136dec86c26fdd2efe17b92"
 
 [[package]]
+name = "openxr"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


jdm (approved):
Looks good!

msub2:
Ah think I missed a couple cfg statements for the OpenXR imports, lemme fix that first

PR #32818, opened 2024-07-20
Un-enum-ify IndependentFormattingContext

valadaptive:
This builds on top of https://github.com/servo/servo/pull/32800 because later work will probably overlap with it and I want to avoid merge conflicts.

The layout code currently has a strong distinction between replaced and non-replaced elements, most of which comes from `IndependentFormattingContext` being an enum with `Replaced` and `NonReplaced` variants. In order to implement `aspect-ratio`, we need to stop treating elements as "replaced"/"not replaced" and instead treat them as "has a preferred aspect ratio"/"does not have a preferred aspect ratio".

The CSS spec [says that calculating the automatic sizes of an element with a preferred aspect ratio involves treating it as a replaced element](https://drafts.csswg.org/css-sizing-4/#aspect-ratio-automatic), so we should be able to treat every element as replaced for the purposes of size calculation, in case they have `aspect-ratio` set.

This is a draft PR because actually implementing the "has a preferred ratio" check is left for a future commit, and I added a bunch of TODO comments so that I can remember to come back and place them in the appropriate spots. I'm opening it up for review to make sure I'm on the right track and there aren't any major concerns.

There are a few commits, but the first ("[Un-enum-ify IndependentFormattingContext](https://github.com/servo/servo/pull/32818/commits/1536cfdcbd3dd6945163a33e320e74f8492ed1af)") is probably the most useful one to review individually. You may want to hide whitespace changes because I moved some functions around scope-wise. After that is a pretty rote one that replaces all uses of the `style` and `base_fragment_info` getters with property accesses now that `IndependentFormattingContext` is not an enum, which makes a lot of one-line changes scattered throughout the code.

---

### DIFF
--- a/components/layout_2020/flexbox/construct.rs
+++ b/components/layout_2020/flexbox/construct.rs
@@ -213,7 +213,7 @@ where
 
         // https://drafts.csswg.org/css-flexbox/#order-modified-document-order
         children.sort_by_key(|child| match &*child.borrow() {
-            FlexLevelBox::FlexItem(item) => item.style().clone_order(),
+            FlexLevelBox::FlexItem(item) => item.style.clone_order(),
 
             // “Absolutely-positioned children of a flex container are treated
             //  as having order: 0 for the purpose of determining their painting order
--- a/components/layout_2020/flexbox/construct.rs
+++ b/components/layout_2020/flexbox/construct.rs
@@ -15,8 +15,7 @@ use crate::dom_traversal::{Contents, NodeAndStyleInfo, NonReplacedContents, Trav
 use crate::flow::inline::construct::InlineFormattingContextBuilder;
 use crate::flow::{BlockContainer, BlockFormattingContext};
 use crate::formatting_contexts::{
-    IndependentFormattingContext, NonReplacedFormattingContext,
-    NonReplacedFormattingContextContents,
+    IndependentFormattingContext, NonReplacedFormattingContextContents,
 };
 use crate::positioned::AbsolutelyPositionedBox;
 use crate::style_ext::DisplayGeneratingBox;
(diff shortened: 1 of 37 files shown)


valadaptive:
Hm, seems like the web platform tests I created in #32800 are being picked up here as well. Is there a good way to deal with this when creating PRs that build on each other?

jdm:
It's possible that [setting the base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) might avoid that.

valadaptive:
What should I set the base branch to? If I set it to something on my fork, the PR will appear there instead. Is there a certain branch here that's used for that purpose?

jdm:
Mmm, true. I've only used it for stacking PRs that are all part of the same fork.

mrobinson:
> Hm, seems like the web platform tests I created in #32800 are being picked up here as well. Is there a good way to deal with this when creating PRs that build on each other?

Unfortunately there isn't a good way to deal with this, as GitHub doesn't have great support for stacking PRs. The good thing is that when you rebase this branch, the WPT PR will be closed automatically.

valadaptive:
This PR may not actually be necessary for implementing `aspect-ratio`; I'll see how things shake out.

mrobinson:
@valadaptive Any updates on this. I'm happy to land this even if not necessary, if you think that it will make the code cleaner.

valadaptive:
For now, I don't think this is necessary to land. My last progress in this area was about a week ago, implementing `aspect-ratio` for `inline-block`, but I realized a lot of my work was based on a misunderstanding of the spec and would have to be redone.

If this PR does turn out to be necessary, I'll reopen it at some point in the future.

PR #32819, opened 2024-07-20
Implement URL::parse()

shanehandley:


mrobinson (approved):

components/script/dom/url.rs
```diff
@@ -133,6 +133,28 @@ impl URL {
...
+
+        // Step 2: If parsedURL is failure, then return null.
+        // Step 3: Let url be a new URL object.
+        // Step 4: Initialize url with parsedURL.
+        // Step 5: Return url.
+
+        // These steps are all handled while mapping the Result to an Option<ServoUrl>.
+        // Regarding initialization, the same condition should apply here as stated in the comments
+        // in Self::Constructor above - construct it on-demand inside `URL::SearchParams`.
+        parsed_url.ok().map(|u| URL::new(global, None, u))
```
mrobinson, line 155: I think this can be a bit simpler: `URL::new(global, None, parsed_url?)`

components/script/dom/url.rs
```diff
@@ -133,6 +133,28 @@ impl URL {
...
 
+    /// <https://url.spec.whatwg.org/#dom-url-parse>
+    pub fn Parse(
+        global: &GlobalScope,
+        url: USVString,
+        base: Option<USVString>,
+    ) -> Option<DomRoot<URL>> {
+        // Step 1: Let parsedURL be the result of running the API URL parser on url with base,
+        // if given.
+        let parsed_base = base.map(|b| ServoUrl::parse(b.0.as_str()).ok()).flatten();
```
mrobinson, line 144: Can you avoid the flatten with `and_then`?

```suggestion
        let parsed_base = base.and_then(|base| ServoUrl::parse(base.0.as_str()).ok())
```

mrobinson (changes requested):

components/script/dom/url.rs
shanehandley: Thank you, I've just tried this but the compiler was unhappy: "the `?` operator can only be used on `Option`s, not `Result`s, in a method that returns `Option`"

mrobinson: How about this: `URL::new(global, None, parsed_url.ok()?)`?

shanehandley: "the `?` operator can only be applied to values that implement `Try` the trait `Try` is not implemented for `root::Root<root::Dom<dom::url::URL>>" :(

mrobinson: Perhaps you put the `?` on the outside of the final parenthesis. I tried this locally and this seems to work fine:

```rust
Some(URL::new(global, None, parsed_url.ok()?))
```

shanehandley: Ah that's sorted it 👍

mrobinson (approved):

PR #32827, opened 2024-07-21
webgpu: Update wgpu to 22.0

sagudev:
Mostly because I need https://github.com/gfx-rs/wgpu/pull/5992.

The only real change is in https://github.com/servo/servo/pull/32827/commits/e6a710b29d288efe175dc9e3831ce31c3700ed59, that is needed because wgpu panics if id is not created.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,8 +136,8 @@ webpki-roots = "0.25"
 webrender = { git = "https://github.com/servo/webrender", branch = "0.64", features = ["capture"] }
 webrender_api = { git = "https://github.com/servo/webrender", branch = "0.64" }
 webrender_traits = { path = "components/shared/webrender" }
-wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
-wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
+wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "c0e7c1ef945a7dd61c81fb951ea554213811aee0" }
+wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "c0e7c1ef945a7dd61c81fb951ea554213811aee0" }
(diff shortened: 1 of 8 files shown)


sagudev:
Some crashes are wgpu regressions (I think, will investigate): https://github.com/sagudev/servo/actions/runs/10029291333/job/27718050321

Cargo.toml
```diff
@@ -136,8 +136,8 @@ webpki-roots = "0.25"
 webrender = { git = "https://github.com/servo/webrender", branch = "0.64", features = ["capture"] }
 webrender_api = { git = "https://github.com/servo/webrender", branch = "0.64" }
 webrender_traits = { path = "components/shared/webrender" }
-wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
-wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "f25e07b984ab391628d9568296d5970981d79d8b" }
+wgpu-core = { git = "https://github.com/gfx-rs/wgpu", rev = "c0e7c1ef945a7dd61c81fb951ea554213811aee0" }
+wgpu-types = { git = "https://github.com/gfx-rs/wgpu", rev = "c0e7c1ef945a7dd61c81fb951ea554213811aee0" }
```
mrobinson, line 140: Can we switch to a versioned release now?

sagudev: No, https://github.com/gfx-rs/wgpu/pull/5992 was committed after release and I will likely need to fix crashes upstream as they are upstream regressions.

Currently:

> We [wgpu] do a major breaking release every 12 weeks. **This happens no matter the status of various in-flight projects.**

So there is not much stability in released version (yet). 

Also Firefox follows trunk with weekly updates, and with us also tracking more regularly then releases we can detect regressions faster in wgpu (sometimes even before Firefox, as in some places we are more mature). WGPU does not run CTS in CI (and it only supports running few tests of very old CTS checkout via deno runner), so it relies on consumers on more extensive testing (mostly firefox, but servo could play more important role in the future).

gterzian (approved):
LGTM

PR #32829, opened 2024-07-22
compositor: propagate scroll events across pipelines

mukilan:
Currently we don't propagate the scroll events across pipeline and
scroll only nodes contained within the nearest pipeline that passes hit
test at a gived point. This causes iframes to "capture" the scroll and
halt the scrolling process once the iframe has been scrolled all the
way. The expected behaviour is that the parent page begins to scroll
once the iframe has been scrolled fully.

The issue is present in both desktop and mobile, but was more noticeable
on mobile, especially on the default servo.org page, because of the relative
sizes of the YouTube video's iframe wrt to the parent.

- [x] These changes might require tests in WPT if they don't already exist, but there are none as part of this change.

### DIFF
--- a/components/compositing/compositor.rs
+++ b/components/compositing/compositor.rs
@@ -1808,19 +1808,31 @@ impl<Window: WindowMethods + ?Sized> IOCompositor<Window> {
             ScrollLocation::Start | ScrollLocation::End => scroll_location,
         };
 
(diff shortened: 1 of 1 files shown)


jdm (approved):
This has bothered me for years, but I always assumed the fix would be more complicated!

Issue #32342, opened 2024-05-22
Allow mach bootstrap to run non-interactively on Windows


delan:
On Windows, mach bootstrap runs `choco install -y support/windows/chocolatey.config` in a `cmd /K`, which stays open once complete. This is good for interactive users, because it allows you to read any errors that occurred, but it makes the command annoying to use in a CI context. Perhaps we could add an option to use `cmd /C` instead.

mrobinson:
This makes sense to me. In general, we don't run choco on the CI because it is very slow and everything we need is usually installed.

PR #32836, opened 2024-07-23
bootstrap: Add `winget` fallback

jschwe:
When `choco` is not available, we can install the same packages from winget. winget is an official package manager from Microsoft, which is available by default on Windows 11.

Note: `winget` also has non-interactive installation options, but accepting license agreements should still be the responsibility of the user, so we don't add any such option for now.

---

- [x] These changes fix: Bootstrap on windows without `choco` available.

### DIFF
--- a/python/servo/platform/windows.py
+++ b/python/servo/platform/windows.py
@@ -13,6 +13,7 @@ import tempfile
 from typing import Optional
 import urllib
 import zipfile
+import shutil
 
 from servo import util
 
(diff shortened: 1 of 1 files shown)


jdm:
When is Choco not available?

jschwe:
While Choco is available by default on GitHub actions, it's not pre installed on windows. Winget on the other hand is pre installed on win 11 now.

mrobinson:
I'm some ways it is easier if there is only one way to install dependencies. It's quite common that something breaks without anyone realizing and the instructions quickly get out of date. I would be in favor of switching to winget, but only when we can ensure that the CI can also use it.

jschwe:
@mrobinson I saw that you previously [mentioned that choco is not used in CI](https://github.com/servo/servo/issues/32342#issuecomment-2124885906). Is that not the case anymore, or am I missing some context?

mrobinson:
> @mrobinson I saw that you previously [mentioned that choco is not used in CI](https://github.com/servo/servo/issues/32342#issuecomment-2124885906). Is that not the case anymore, or am I missing some context?

Oh yeah, great point. I think we should switch to *only* use winget then.

jschwe:
I'm not confident that the current PR is "production-ready". I haven't tested choco myself, so I don't know how many points also affect choco, but some things to still consider / test on fresh machines: 

- Does all the installed software actually get added to `PATH`? I faintly remember that `winget` leaves that up to each individual installer.
- Do we want / need to specify versions, or are we fine with `latest`.
- Should we pass `--no-upgrade` (e.g. LLVM has minor updates somewhat frequently
- `winget` also seems to take "ownership" of at least some applications my system, that I'm confident I didn't install via winget (e.g. Firefox) . I think we should test what happens with e,g. an existing (non winget) installation of CMake / LLVM and if winget will upgrade it, if we run `winget install`.

mrobinson:
I was experimenting with this a bit and it seems that both LLVM and WixToolset do not add themselves to the PATH by default. LLVM can be configured when running winget in interactive mode (`-i`). Unfortunately, `import` cannot be used together with `-i`.

We could have run each winget installation separately, but that would pop up a powershell window for each one. We could not pop up the powershell window, but the WiXToolset installer does not properly ask for administrator permissions when run interactively through winget. Maybe there's a way around this last issue.

jschwe:
Do we still want to support bootstrap in CI? If yes, then we can't run the llvm install in interactive mode (unless we individually check if a package is installed before invoking winget install ) . 

I guess we could also consider guessing the install location and fixing up PATH ourselves.

jschwe:
Not sure how feasible it is (to set up), but perhaps we could put up a small bug bounty to fix the upstream llvm issue? For someone who has experience with the installer tool llvm is using it might be easy to add a commline switch.

mrobinson:
> Not sure how feasible it is (to set up), but perhaps we could put up a small bug bounty to fix the upstream llvm issue? For someone who has experience with the installer tool llvm is using it might be easy to add a commline switch.

It seems that Windows Server 2025 finally has support for winger, but I think it's fine to make this transition gradually. We can add support for winget, gradually make sure that it can run non-interactively and then later remove choco support. I'm not sure how feasible it will ever be to use winget on CI as a pre-baked runner image can save a lot of build time. I think it's okay to figure that out as we go along though.

jschwe:
@mrobinson I rebased and made some adjustments to this PR. 

- Switched to `winget install` so that we can pass `--interactive`, which would be needed so that the user can tick the box to add LLVM to the PATH. 
- Directly run `winget install` with `subprocess.run`. stdout and stdin are attached to the parent shell, so we can still interact with it (e.g. if we wanted to add the `--wait`parameter). I'm not sure if there was a reason for the previous approach, but to me this seems to work fine and be preferable compared to opening a new `cmd` window.
- I'm not sure if we want to install specific versions as the current choco config does.

mrobinson:
> * Directly run `winget install` with `subprocess.run`. stdout and stdin are attached to the parent shell, so we can still interact with it (e.g. if we wanted to add the `--wait`parameter). I'm not sure if there was a reason for the previous approach, but to me this seems to work fine and be preferable compared to opening a new `cmd` window.

I think a `cmd` window was used so that `choco` could run in Administrator mode.

python/servo/platform/windows.py
```diff
@@ -28,12 +29,55 @@
...
+        # We install tools like LLVM / CMake, so we probably don't want to force-upgrade
+        # a user installed version without good reason.
+        cmd =  ['winget','install', '--interactive']
+        if force:
+            cmd.append('--force')
+        else:
+            cmd.append('--no-upgrade')
+
+        cmd.extend(WINGET_DEPENDENCIES)
+        # output is printed to the terminal that `./mach bootstrap` is running in
```
mrobinson, line 55: Nit:

```suggestion

        # The output will be printed to the terminal that `./mach bootstrap` is running in.
```

mrobinson (approved):
Thank you!

Maybe if we aren't using winget on the bots we can switch to it by default and just ensure that we always run it in interactive mode.

mrobinson (approved):
Looks good. Just needs to be rebased.

PR #32840, opened 2024-07-24
Remove googlevr feature

msub2:
Now that the googlevr backend is removed from the WebXR crate, this feature should be removed too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,12 +125,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


mrobinson (changes requested):
Please also remove references to "googlevr" in `python/servo/build_commands.py` and `python/servo/package_commands.py`.

mrobinson (changes requested):
Thanks! I look like the `vr` argument can be removed from the `change_prefs` method now too or do you think that it is still useful for some reason?

msub2:
Hmm, I guess it probably can be completely removed. The OpenXR backend isn't capable of running on Android yet, and since it's the only real WebXR backend that would be of use there we probably wouldn't need to distinguish between flavors.

mrobinson (approved):

PR #32842, opened 2024-07-24
Bump nixpkgs and add cargo-deny in shell.nix

mukilan:
We need Android NDK r26c for the SM upgrade (#32769) and the fix in nixpkgs to
make r26 NDKs function correctly (NixOS/nixpkgs#298285). Both of those are not
available in the snapshot of nixpkgs we use currently. The version of the new 
snapshot is based on `nixos-24.05`.

Also add cargo-deny to shell.nix to make `./mach test-tidy` work.

---
- [x] These changes do not require tests because they only touch nixos support code.

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -5,7 +5,7 @@
   buildAndroid ? false
 }:
 with import (builtins.fetchTarball {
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

mrobinson:
Sending to merge queue again due to #32849.

PR #32843, opened 2024-07-24
clippy: Fix warning in `components/canvas`

remysaissy:
Fixes a clippy warning in the components/canvas directory. It is part of https://github.com/servo/servo/issues/31500.

---

### DIFF
--- a/components/canvas/canvas_data.rs
+++ b/components/canvas/canvas_data.rs
@@ -251,7 +251,7 @@ impl<'a> UnshapedTextRun<'a> {
         }
     }
 
-    fn to_shaped_text_run(self) -> Option<TextRun> {
+    fn into_shaped_text_run(self) -> Option<TextRun> {
         let font = self.font?;
         if self.string.is_empty() {
             return None;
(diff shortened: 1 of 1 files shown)


mrobinson (changes requested):
What is the warning that is fixed here? If it is that the function has an incorrect name, the answer is probably to fix the name instead.

mrobinson:
Perhaps try changing the name of this function to `into_shaped_text_run`. It's actually pretty useful that this consume the `self` type.

remysaissy:
> What is the warning that is fixed here? If it is that the function has an incorrect name, the answer is probably to fix the name instead.

 _--> components/canvas/canvas_data.rs:254:27
    |
254 |     fn to_shaped_text_run(self) -> Option<TextRun> {
    |                           ^^^^
    |
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
    = note: `#[warn(clippy::wrong_self_convention)]` on by default_

remysaissy:
> Perhaps try changing the name of this function to `into_shaped_text_run`. It's actually pretty useful that this consume the `self` type.

Ok, I will update the PR to consume the self type instead + rename the function.

mrobinson (approved):

mrobinson:
Thank you!

PR #32854, opened 2024-07-25
flex: compute intrinsic sizes for flex items and flex containers

delan:
Signed-off-by: Delan Azabani <dazabani@igalia.com>

---


components/layout_2020/flexbox/mod.rs
```diff
@@ -15,10 +26,160 @@ mod layout;
...
+            .main
+            .non_auto()
+            .map(|preferred_size| preferred_size + pbm_auto_is_zero.main);
+        let outer_min_main_size = content_min_size.main + pbm_auto_is_zero.main;
+        let outer_max_main_size = content_max_size
+            .main
+            .map(|max_main_size| max_main_size + pbm_auto_is_zero.main);
+
+        // The main-size min-content contribution of a flex item is the larger of its outer
+        // min-content size and outer preferred size if that is not auto, clamped by its min/max main size.
```
delan, line 76: quote with `>`

components/layout_2020/flexbox/mod.rs
```diff
@@ -15,10 +26,162 @@ mod layout;
...
+                AuOrAuto::LengthPercentage(l) => l.min(content_size_suggestion),
+                AuOrAuto::Auto => {
+                    if let Some(l) = transferred_size_suggestion {
+                        l.min(content_size_suggestion)
+                    } else {
+                        content_size_suggestion
+                    }
+                },
+            };
+            result.clamp_below_max(max_size.inline)
```
nicoburns, line 181: This seems wrong. Min size usually takes precedence over max size. Perhaps this should only apply if the min size is a computed intrinsic size rather an actual `min-` style?

mrobinson: It's possible that is wrong, but this code comes from the automatic min-size computation that comes from before this change. This change reuses the computation, but fixing it is probably a separate issue.

mrobinson:
Changes that have been made to this branch on top of the work that @delan, @mukilan, and I did together:

- Renamed `ItemInfo` to `FlexItemBoxInlineContentSizesInfo`
- Automatic main size contribution is not duplicated and handled by `FlexItemBox::automatic_min_size()`.
- `FlexItemBox::inline_content_sizes` has been moved to `layout.rs`.
- Code to compute `FlexItemBox::inline_content_sizes` has been moved to `FlexItemBox::inline_content_size_info`.
- Made the computation of the intrinsic max-content main size generic enough to be used for that of intrinsic min-content size of single line flex box. I've done this by:
  - Making `FlexItemBoxInlineContentSizesInfo::flex_grow_or_scaled_flex_shrink_factor` and `FlexItemBoxInlineContentSizesInfo::desired_flex_fraction` part of a further enum:
    ```rust
     #[derive(Default)]
    struct DesiredFlexFractionAndGrowOrShrinkFactor {
        desired_flex_fraction: f32,
        flex_grow_or_shrink_factor: f32,
    }
   ```
  - `FlexItemBoxInlineContentSizesInfo::min_content_main_size_for_multiline_container` was added which store the per-item minimum size when the container is multi-line. The computation is this:
    ```rust
        let mut min_content_main_size_for_multiline_container =
            content_contribution_sizes.min_content;
        if style.get_position().flex_grow.is_zero() {
            min_content_main_size_for_multiline_container.min_assign(flex_base_size);
        }
        if style.get_position().flex_shrink.is_zero() {
            min_content_main_size_for_multiline_container.max_assign(flex_base_size);
        }
        min_content_main_size_for_multiline_container =
            min_content_main_size_for_multiline_container
                .clamp_between_extremums(content_min_size_no_auto.main, content_max_size.main);
   ```

mrobinson (approved):

components/layout_2020/flexbox/mod.rs
mrobinson: I have taken a deeper looker at whether or not we should be clamping here and the [specification](https://drafts.csswg.org/css-flexbox/#min-size-auto) says that we should be:

> The content-based minimum size of a [flex item](https://drafts.csswg.org/css-flexbox/#flex-item) is the smaller of its [specified size suggestion](https://drafts.csswg.org/css-flexbox/#specified-size-suggestion) and its [content size suggestion](https://drafts.csswg.org/css-flexbox/#content-size-suggestion) if its specified size suggestion exists; otherwise, the smaller of its [transferred size suggestion](https://drafts.csswg.org/css-flexbox/#transferred-size-suggestion) and its content size suggestion if the element is [replaced](https://drafts.csswg.org/css-display-4/#replaced-element) and its transferred size suggestion exists; otherwise its content size suggestion. In all cases, the size is clamped by the [maximum](https://drafts.csswg.org/css-sizing-3/#max-width) [main size](https://drafts.csswg.org/css-flexbox/#main-size) if it’s [definite](https://drafts.csswg.org/css-flexbox/#definite).

PR #32858, opened 2024-07-25
ohos/android: Redirect stdout/stderr to `log` sink

jschwe:
- Redirect stdout/stderr to the `log` sink, based on the existing android `redirect_stdout_to_logcat` implementation,
  but using the safe abstractions from `nix`.
- Using the `log` crate instead of using the OS specific raw log function

Note: I can drop the `android` commit, if directly using `__android_log_write` is preferable.

---

### DIFF
--- a/ports/servoshell/egl/log.rs
+++ b/ports/servoshell/egl/log.rs
@@ -20,7 +20,7 @@ pub(crate) enum LogRedirectError {
 pub(crate) fn redirect_stdout_and_stderr() -> Result<(), LogRedirectError> {
     fn log_raw_msg(raw_msg: &[u8]) {
         if let Ok(utf8_msg) = std::str::from_utf8(raw_msg) {
-            info!("{}", utf8_msg);
+            info!("{utf8_msg}");
         } else {
             // Note: This could happen if the message is long, and we hit the length
             // limitation in the middle of a utf-8 codepoint. We could try to handle this
--- a/ports/servoshell/egl/log.rs
+++ b/ports/servoshell/egl/log.rs
@@ -20,7 +20,7 @@ pub(crate) enum LogRedirectError {
 pub(crate) fn redirect_stdout_and_stderr() -> Result<(), LogRedirectError> {
     fn log_raw_msg(raw_msg: &[u8]) {
         if let Ok(utf8_msg) = std::str::from_utf8(raw_msg) {
-            info!("{utf8_msg}");
+            debug!("{utf8_msg}");
         } else {
             // Note: This could happen if the message is long, and we hit the length
             // limitation in the middle of a utf-8 codepoint. We could try to handle this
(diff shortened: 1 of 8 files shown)


ports/servoshell/egl/log.rs
```diff
@@ -0,0 +1,96 @@
...
+#[allow(dead_code)]
+pub(crate) enum LogRedirectError {
+    CreatePipeFailed(nix::Error),
+    RedirectToPipeFailed(nix::Error),
+}
+
+/// Redirect stdout and stderr to the logging system
+pub(crate) fn redirect_stdout_and_stderr() -> Result<(), LogRedirectError> {
+    fn log_raw_msg(raw_msg: &[u8]) {
+        if let Ok(utf8_msg) = std::str::from_utf8(raw_msg) {
```
Taym95, line 22: since this can fail, maybe better to return error from log_raw_msg?

jschwe: I put some comments on how we could handle likely errors in the `else` case, but I was thinking maybe it makes sense to first see in practice if we even hit such cases.
I still have to test this PR in practice though - since I updated my phone to the latest beta version, servo keeps crashing in some spidermonkey code and I'm still investigating that.

jschwe: I got back to this PR today, and tested it. The redirection seems to work as expected on OH. I also couldn't see any warning messages from the error case here.

mrobinson (approved):

ports/servoshell/egl/log.rs
```diff
@@ -0,0 +1,96 @@
...
+            info!("{}", utf8_msg);
```
mrobinson, line 23: This could also just be:

```suggestion
            info!("{utf8_msg}");
```

mrobinson (commented):
Are there any implications to switching from `__android_log_write` to using `info!`?

jschwe:
> Are there any implications to switching from __android_log_write to using info!?

I believe the android logger crate will also [use `__android_log_write`](https://github.com/rust-mobile/android_logger-rs/blob/master/src/lib.rs#L149) under the hood, so this adds another level of indirection in exchange for reducing code duplication. I just checked and the log priority of 3 seems to be [ANDROID_LOG_DEBUG](https://developer.android.com/ndk/reference/group/logging) and not info. Should I switch it to `debug!` instead?

mrobinson:
> > Are there any implications to switching from __android_log_write to using info!?
> 
> I believe the android logger crate will also [use `__android_log_write`](https://github.com/rust-mobile/android_logger-rs/blob/master/src/lib.rs#L149) under the hood, so this adds another level of indirection in exchange for reducing code duplication. I just checked and the log priority of 3 seems to be [ANDROID_LOG_DEBUG](https://developer.android.com/ndk/reference/group/logging) and not info. Should I switch it to `debug!` instead?

Yeah, let's do that in order to preserve the same behavior and then examine later if it makes sense.

mrobinson (approved):

PR #32860, opened 2024-07-26
Implement WebXR Gamepads Module

### LINKED ISSUE #24389: Support gamepads in WebXR
Currently our Gamepad code is tied to the VR API.

It's also not live -- the Gamepad objects need to be manually updated, which is incorrect (https://github.com/servo/servo/issues/23269)

I propose we remove support for `navigator.getGamepads()` and repurpose the existing Gamepad code for WebXR only. We can re-add support for `navigator.getGamepads()` if we ever decide to support standard (like, game controller) gamepads as opposed to XR device gamepads.

msub2:
This PR implements the WebXR Gamepads Module, which exposes a Gamepad object on XRInputSource reflecting the input from the corresponding XR controller.

---

- [x] These changes fix #24389


msub2:
This is nearly ready, just need to apply these frame updates at the correct spot instead of directly on event

components/script/dom/xrsession.rs
```diff
@@ -510,7 +518,13 @@ impl XRSession {
 
     /// <https://immersive-web.github.io/webxr/#xrframe-apply-frame-updates>
     fn apply_frame_updates(&self, _frame: &XRFrame) {
-        // TODO: add a comment about why this is empty right now!
+        // <https://www.w3.org/TR/webxr-gamepads-module-1/#xrframe-apply-gamepad-frame-updates>
+        for (id, frame) in &*self.input_frames.borrow() {
```
jdm, line 522: Should we use borrow_mut() and drain here instead? It would avoid the clone call a few lines down, and would avoid any problems where no updated input was received in a subsequent frame but we still process the previous update anyways.

jdm (approved):

PR #32861, opened 2024-07-26
dependencies: Upgrade `cookie` and rename Servo's `Cookie` to `ServoCookie`

mrobinson:
This changes updates to the new version of the `cookie` crate in Servo
which no longer uses the old `time@0.1` data types. This requires using
a new version of `time` while we transition off of the old one. This is
the first step in that process.

In addition, the overloading of the `cookie::Cookie` name was causing a
great deal of confusion, so I've renamed the Servo wrapper to
`ServoCookie` like we do with `ServoUrl`.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -983,18 +983,19 @@ dependencies = [
 
 [[package]]
 name = "cookie"
-version = "0.12.0"
+version = "0.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


atbrakhi (approved):

mrobinson:
`/html/dom/documents/resource-metadata-management/document-cookie.html` was failing because it tried to clear cookies by giving them an expiration of `Thu, 01 Jan 1970 00:00:00 UTC`. This works in most browsers, but isn't specification compliant (the timezone should be `GMT`). Since this test isn't trying to test whether the non-spec compliant behavior is supported, I have gone ahead and corrected the test.

jdm (approved):

PR #32866, opened 2024-07-27
Disable fail_fast in android matrix and remove unused install in android workflow

sagudev:
cargo deny and taplo are not used in android workflows (we do not check tidy here). In the future we could probably slim down bootstrap too.

---

- [x] These changes do not require tests because they make CI better

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -27,6 +27,7 @@ jobs:
     name: Android Build
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


jschwe:
I think the reason why we use `cargo-install` for taplo and cargo deny, is because bootstrap unconditionally installs taplo and cargo deny if they are not present (which is slow), while the github action will cache the binaries, speeding up the overall process.

sagudev:
> I think the reason why we use `cargo-install` for taplo and cargo deny, is because bootstrap unconditionally installs taplo and cargo deny if they are not present (which is slow), while the github action will cache the binaries, speeding up the overall process.

Ah, you are right, we need to make everything optional in bootstrap for this😭.

PR #32871, opened 2024-07-29
Implement AbortController

syvb:
Currently this is the same as #31361; I still need to implement fetch aborting.

---


jdm:
@syvb Could we merge the stub DOM implementation with a preference attached to it, so we can at least experiment with web content that expects the interface to exist?

arihant2math:
Since `AbortController` is implemented, I think this can be closed.

PR #32655, opened 2024-07-01
DevTools: Show HTML tree

eerii:
This PR partly restores the HTML inspector functionality for the DevTools implementation. The inspector is probably the part of the devtools that has more [associated](https://searchfox.org/mozilla-central/source/devtools/server/actors/inspector) [actors](https://firefox-source-docs.mozilla.org/devtools/backend/actor-hierarchy.html). There were quite a few API changes since the last working implementation, so making it work again requires both fixing existing actors and creating new ones.

| **Existing actors** | **New actors** |
|---------------------|----------------|
| Inspector           | Accessibility  |
| Walker              | Access. Walker |
| Highligter          | Simulator      |
| Node                | CSS Properties |
| Page Style          | Reflow         |
| Layout              |                |
| Thread              |                |
| Network Parent      |                |

Some of the new actors and messages are placeholders that don't contain correct information yet, but they are required for the flow of messages to continue, since Firefox waits until a reply is issued for some of them to emit others.

image

Right now the DOM tree outline is showing, with two caveats that will be addressed in future PRs (since this has already gotten quite big):

- Node contents don't show up, for example, the text inside a paragraph. This is because this needs to be sent as a special inline content field in the node message which is not handled yet.
- Styles and CSS are not there yet since they require quite a few more work.

A few actors and message fields also seem to be outdated, namely Emulation, Performance and Timeline. Since Emulation was only a placeholder I removed it, but I kept the other two since they had work done in them that might be useful.

---
- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes are a part of #32404
- [x] These changes do not require tests because there are no tests yet for DevTools

PR #32884, opened 2024-07-30
DevTools: Inline text and clean whitespace

eerii:
This PR addresses two issues left by #32655:

- In some cases child text nodes are small enough to show them inline in the DOM tree view. When this happens, Servo needs to append an special `inlineTextChild` to the `NodeActorMsg` with the child's contents. The conditions for this are defined [here](https://searchfox.org/mozilla-central/source/devtools/server/actors/inspector/walker.js#746):
  - The node only has one child
  - The child is a text node
  - The text is not very long (<50 characters, defined [here](https://searchfox.org/mozilla-central/source/devtools/server/actors/inspector/walker.js#180))
  - The node is not a slot or a flex item
- The spaces and line breaks in an HTML file are usually just separators that allow to format the code nicely. However, the browser treats this whitespace as an [inline anonymous box](https://developer.mozilla.org/en-US/docs/Web/CSS/Visual_formatting_model#anonymous_boxes). Usually this doesn't have any effect, but when they appear surrounded by inline elements they may have a non zero size, as seen in [this example](https://firefox-devtools.github.io/devtools-examples/whitespace-only-demo/index.html). Because of this, filtering the whitespace in the inspector tree is more tricky than it may seem, since we need to keep this inline whitespace because it can create gaps in the layout which should be inspectable. The rules for filtering are defined [here](https://firefox-source-docs.mozilla.org/devtools-user/page_inspector/how_to/examine_and_edit_html/index.html#whitespace-only-text-nodes).

|**Before**|**After**|
|---|---|
|image|image|

---
- [x] These changes are a part of #32404
- [x] These changes do not require tests because there are no tests yet for DevTools

### DIFF
--- a/components/devtools/actors/inspector/node.rs
+++ b/components/devtools/actors/inspector/node.rs
@@ -9,7 +9,9 @@ use std::collections::HashMap;
 use std::net::TcpStream;
 
 use base::id::PipelineId;
-use devtools_traits::DevtoolScriptControlMsg::{GetDocumentElement, ModifyAttribute};
+use devtools_traits::DevtoolScriptControlMsg::{
+    GetChildren, GetDocumentElement, GetNodeValue, ModifyAttribute,
+};
 use devtools_traits::{DevtoolScriptControlMsg, NodeInfo};
 use ipc_channel::ipc::{self, IpcSender};
 use serde::Serialize;
@@ -18,6 +20,8 @@ use serde_json::{self, Map, Value};
 use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
 use crate::protocol::JsonPacketStream;
 use crate::{EmptyReplyMsg, StreamId};
+
+const MAX_INLINE_LENGTH: usize = 50;
 
 #[derive(Serialize)]
 struct GetUniqueSelectorReply {
@@ -41,6 +45,8 @@ pub struct NodeActorMsg {
     container_type: Option<()>,
     pub display_name: String,
     display_type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    inline_text_child: Option<Box<NodeActorMsg>>,
     is_after_pseudo_element: bool,
     is_anonymous: bool,
     is_before_pseudo_element: bool,
@@ -55,7 +61,7 @@ pub struct NodeActorMsg {
     is_top_level_document: bool,
     node_name: String,
     node_type: u16,
-    node_value: Option<()>,
+    node_value: Option<String>,
     pub num_children: usize,
     #[serde(skip_serializing_if = "String::is_empty")]
     parent: String,
@@ -156,7 +162,7 @@ impl NodeInfoToProtocol for NodeInfo {
             let name = actors.new_name("node");
             let node_actor = NodeActor {
                 name: name.clone(),
-                script_chan,
+                script_chan: script_chan.clone(),
                 pipeline,
             };
             actors.register_script_actor(self.unique_id, name.clone());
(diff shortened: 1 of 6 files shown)


components/devtools/actors/inspector/node.rs
```diff
@@ -19,6 +19,8 @@ use crate::actor::{Actor, ActorMessageStatus, ActorRegistry};
 use crate::protocol::JsonPacketStream;
 use crate::{EmptyReplyMsg, StreamId};
 
+const MAX_INLINE_LENGTH: usize = 50;
```
mrobinson, line 22: It would be nice to have a rustdoc comment here for this constant.

components/devtools/actors/inspector/node.rs
```diff
@@ -166,13 +170,46 @@ impl NodeInfoToProtocol for NodeInfo {
             actors.script_to_actor(self.unique_id)
         };
 
+        let name = actors.actor_to_script(actor.clone());
+
+        // If a node only has a single text node as a child whith a small enough text,
+        // return it with this node as an `inlineTextChild`.
+        let inline_text_child = (|| {
+            // TODO: Also return if this node is a flex element
```
mrobinson, line 178:
```suggestion
            // TODO: Also return if this node is a flex element.
```

components/devtools/actors/inspector/node.rs
```diff
@@ -166,13 +170,46 @@ impl NodeInfoToProtocol for NodeInfo {
...
+            script_chan
+                .send(GetChildren(pipeline, name.clone(), tx))
+                .unwrap();
+            let mut children = rx.recv().ok()??;
+
+            let child = children.pop()?;
+            let msg = child.encode(actors, true, script_chan.clone(), pipeline);
+
+            // Check if it is a text node
+            if msg.node_type != 3 {
```
mrobinson, line 193: It would be better to have a constant for this magic number. A comment would be useful as well explaining that the constant is defined again to avoid a dependency on `script` where it is actually defined (`script::dom::bindings::codegen::Bindings::NodeBinding::NodeConstants`).

components/devtools/actors/inspector/node.rs
```diff
@@ -166,13 +170,46 @@ impl NodeInfoToProtocol for NodeInfo {
...
+                return None;
+            }
+
+            // Check if it is small enough
```
mrobinson, line 197:
```suggestion
            // If the text node child is too big, do not represent it inline.
```

components/script/devtools.rs
```diff
@@ -125,7 +125,44 @@ pub fn handle_get_children(
     match find_node_by_unique_id(documents, pipeline, &node_id) {
         None => reply.send(None).unwrap(),
         Some(parent) => {
-            let children = parent.children().map(|child| child.summarize()).collect();
+            let is_whitespace = |node: &NodeInfo| {
+                node.node_type == 3 &&
+                    (node.node_value.is_none() ||
+                        node.node_value.clone().unwrap().trim().is_empty())
```
mrobinson, line 131:
```suggestion
                    (node.node_value.is_none() ||
                        node.node_value.clone().unwrap().trim().is_empty())
```

Could this use `map_or_else` to avoid an `unwrap` here?

components/script/devtools.rs
```diff
@@ -125,7 +125,44 @@ pub fn handle_get_children(
...
+                    let info = child.summarize();
+                    if !is_whitespace(&info) {
+                        return Some(info);
+                    }
+
+                    (prev_inline && next_inline).then_some(info)
+                })
+                .collect();
+
+            log::warn!("{:?}", children);
```
mrobinson, line 165: I think this might have been left over from some debugging?

mrobinson (changes requested):
Looks great. Just a few minor comments:

components/script/devtools.rs
eerii: It was, sorry for the oversight

mrobinson (approved):

mrobinson:
Thanks!

PR #32886, opened 2024-07-30
layout: Fix display of new text in `textarea` elements

mrobinson:
Previously `<textarea>` was just displaying node contents, which is the
original text content, not the one updated by later typing. This change
fixes that issue.

---


mukilan (approved):
Nice!

PR #32891, opened 2024-07-30
Implement gaps in flexbox layout

Loirooriol:
As per https://drafts.csswg.org/css-align/#gaps

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -19,6 +19,7 @@ use style::properties::ComputedValues;
 use style::values::computed::length::Size;
 use style::values::computed::Length;
 use style::values::generics::flex::GenericFlexBasis as FlexBasis;
+use style::values::generics::length::LengthPercentageOrNormal;
 use style::values::specified::align::AlignFlags;
 use style::Zero;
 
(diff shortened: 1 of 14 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -299,16 +300,37 @@ impl FlexContainer {
...
             container_main_size,
             &mut flex_items,
-            |flex_context, mut line| line.layout(flex_context, container_main_size),
+            main_gap,
         );
 
-        let content_cross_size = flex_lines.iter().map(|line| line.cross_size).sum();
+        let line_count = flex_lines.len();
+        let content_cross_size = flex_lines.iter().map(|line| line.cross_size).sum::<Au>() +
+            cross_gap * (line_count as i32 - 1);
```
nicoburns, line 333: Have you tested for possible divide-by-zero here? (perhaps servo guarantees at least one line but that isn't obvious from the diff).

nicoburns (approved):
This will also need to be implemented in `compute_inline_sizes` once https://github.com/servo/servo/pull/32854 lands. But otherwise this looks good to me.

mrobinson (approved):

components/layout_2020/flexbox/layout.rs
Loirooriol: `collect_flex_lines` returns a vec with at least one line...

components/layout_2020/flexbox/layout.rs
```diff
@@ -844,9 +867,11 @@ impl FlexLine<'_> {
         &mut self,
         flex_context: &mut FlexContext,
         container_main_size: Au,
+        main_gap: Au,
     ) -> FlexLineLayoutResult {
+        let item_count = self.items.len();
         let (item_used_main_sizes, mut free_space) =
-            self.resolve_flexible_lengths(container_main_size);
+            self.resolve_flexible_lengths(container_main_size - main_gap * (item_count as i32 - 1));
```
Loirooriol, line 874: ...but here it might be possible to have zero items, then adding the gap seems wrong, but not sure if it matters.

Issue #25182, opened 2019-12-06
Implement securitypolicyviolation events


pshaughn:
I encountered this because various tests under WPT html/webappapis/scripting want to try every single global event handler to see if on* is connected to addEventListener(*) in the appropriate way; this is one of four missing ones.

Securitypolicyviolation is to be fired when a content security policy is violated as per https://w3c.github.io/webappsec-csp/#report-violation. Policy checking itself might still remain to be implemented, see issue #4577.

Issue #4577, opened 2015-01-08
Implement Content Security Policy.


Ms2ger:
Depends on #4576.

Estimated 2 months work assuming the right hooks.

ferjm:
Blocks #13942

notriddle:
https://github.com/notriddle/rust-content-security-policy

TimvdLippe:
There are quite a few test suites failing in https://wpt.fyi/results/content-security-policy/ which are because `window.open` and `iframes` don't use `fetch` navigation logic. Therefore, while most CSP logic is in place in the right places for navigation, these two codepaths don't make use of them.

As I understand it, either these two codepaths need to use `fetch` navigation logic as the spec dictates. However, if I understand previous comments, it is intentional that these codepaths are different.

Therefore, if unification is not an option, can you point me to which spots I should hook up the CSP requests? Most notably, we need to check navigation requests, but with `Destination::Frame` for example to make `child-src` tests pass.

jdm:
Both of those do use fetch logic, but #36760 is missing. I'm not sure if that's what you're referring to?

TimvdLippe:
No I don't think that's what's missing here. https://github.com/servo/servo/tree/main/tests/wpt/tests/content-security-policy/child-src are tests that are supposed to run when `Destination::Frame` is set: (https://github.com/servo/rust-content-security-policy/blob/334bfcbf0a3f503b21c90aee6aee30d4b8c9558a/src/lib.rs#L1478 coming from https://github.com/servo/rust-content-security-policy/blob/334bfcbf0a3f503b21c90aee6aee30d4b8c9558a/src/lib.rs#L1499). That's set in https://github.com/servo/servo/blob/fdb9c0af12837e5602369191f51f58f40baac01c/components/net/fetch/methods.rs#L177 which comes from https://github.com/servo/servo/blob/fdb9c0af12837e5602369191f51f58f40baac01c/components/shared/net/request.rs#L396

When I look into the callsides of that builder, I never see any codepath set it to `Destination::Frame`. I also am missing the codepath from the iframe code into this. Can you point me to the location where the iframe code eventually flows into the `RequestBuilder` code and where I would need to set the destination for it?

jdm:
[...]

PR #32893, opened 2024-07-30
script: Add support for `unsafe-eval` and `wasm-unsafe-eval`

chocolate-pie:
Add support for `unsafe-eval` and `wasm-unsafe-eval` directive and initial support for `securitypolicyviolation` event (needed to pass WPT under `content-security-policy/unsafe-eval` and `content-security-policy/wasm-unsafe-eval`).
##### WPT
content-security-policy/unsafe-eval: 10/11 tests were passed.
content-security-policy/wasm-unsafe-eval:  10/29 tests were passed.

---

- [x] These changes fix a part of https://github.com/servo/servo/issues/4577 and https://github.com/servo/servo/issues/25182


tests/wpt/meta/html/semantics/scripting-1/the-script-element/module/dynamic-import/v8-code-cache.html.ini
```diff
@@ -0,0 +1,31 @@
+[v8-code-cache.html]
+  expected: ERROR
```
jdm, line 2: Have we investigated the new timeouts and errors that this PR introduces?

components/script/security_manager.rs
```diff
@@ -0,0 +1,177 @@
...
+        // > Step 1: If url’s scheme is not an HTTP(S) scheme, then return url’s scheme.
+        if scheme != "https" && scheme != "http" {
+            return scheme.to_owned();
+        }
+        // > Step 2: Set url’s fragment to the empty string.
+        url.set_fragment(None);
+        // > Step 3: Set url’s username to the empty string.
+        let _ = url.set_password(None);
+        // > Step 4: Set url’s password to the empty string.
+        let _ = url.set_username("");
```
jdm, line 131: Nit: the calls and the comments don't match here.

components/script/dom/xmlhttprequest.rs
```diff
@@ -1527,6 +1527,7 @@ impl XMLHttpRequest {
             docloader,
             None,
             None,
+            doc.status_code(),
```
jdm, line 1530: Do we actually want to propagate the status code from the original document like this? That seems like it could be confusing, since it's a synthetic document. Should it be the status code of the xhr response instead?

components/script/dom/servoparser/mod.rs
```diff
@@ -214,6 +214,7 @@ impl ServoParser {
             loader,
             None,
             None,
+            context_document.status_code(),
```
jdm, line 217: Since this is a synthetic document, why do we want to propagate another document's status code?

components/script/dom/htmlheadelement.rs
```diff
@@ -80,6 +81,56 @@ impl HTMLHeadElement {
             }
         }
     }
+
+    /// <https://w3c.github.io/webappsec-csp/#meta-element>
```
jdm, line 85: https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-content-security-policy

components/script/dom/htmlmetaelement.rs
```diff
@@ -115,6 +117,15 @@ impl HTMLMetaElement {
         }
     }
 
+    /// <https://w3c.github.io/webappsec-csp/#meta-element>
```
jdm, line 120: https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-http-equiv-content-security-policy

components/script/dom/domparser.rs
```diff
@@ -78,6 +78,7 @@ impl DOMParserMethods for DOMParser {
                     loader,
                     None,
                     None,
+                    doc.status_code(),
```
jdm, line 81: I'm not convinced we should propagate the status code in these methods.

components/script/dom/domimplementation.rs
```diff
@@ -155,6 +155,7 @@ impl DOMImplementationMethods for DOMImplementation {
             loader,
             None,
             None,
+            self.document.status_code(),
```
jdm, line 158: I don't think we should propagate the status code.

components/script/dom/document.rs
```diff
@@ -3443,6 +3451,7 @@ impl Document {
             docloader,
             None,
             None,
+            doc.status_code(),
```
jdm, line 3454: I don't think we should propagate the status code here.

components/script/dom/document.rs
```diff
@@ -474,6 +474,8 @@ pub struct Document {
     fonts: MutNullableDom<FontFaceSet>,
     /// <https://html.spec.whatwg.org/multipage/#visibility-state>
     visibility_state: Cell<DocumentVisibilityState>,
+    /// <https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml>
+    status_code: Cell<Option<u16>>,
```
jdm, line 478: Do we ever update this value after it's initialize? If not, no need for a cell.

jdm (changes requested):
Impressive work! I've got some clarifications and I want to make sure we understand the new test results before this merges.

components/script/dom/htmlheadelement.rs
```diff
@@ -80,6 +81,56 @@ impl HTMLHeadElement {
...
+            });
+
+        for meta in candinates {
+            if let Some(ref content) = meta.get_attribute(&ns!(), &local_name!("content")) {
+                let content = content.value();
+                let content_val = content.trim();
+                if !content_val.is_empty() {
+                    match csp_list {
+                        Some(ref mut csp_list) => csp_list.append(CspList::parse(
+                            content_val,
```
jdm, line 116: Let's de-duplucate the CsoList::parse call by storing it in a local variable before the match.

jdm (approved):

PR #32897, opened 2024-07-31
Fix panic in Webrender during shutdown

### LINKED ISSUE #32867: Intermittent panic in webrender during shutdown
```
▶ CRASH [expected FAIL] /html/canvas/element/manual/imagebitmap/imageBitmapRendering-transferFromImageBitmap-webgl.html
  │ 
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ Shutting down the Constellation after generating an output file or exit flag specified
  │ called `Option::unwrap()` on a `None` value (thread WRSceneBuilder#0, at /home/runner/.cargo/git/checkouts/webrender-c3596abe1cf4f320/9d354ad/webrender/src/scene_builder_thread.rs:517)
  │    0: servoshell::backtrace::print
  │    1: servoshell::panic_hook::panic_hook
  │    3: std::panicking::begin_panic_handler::{{closure}}
  │    4: std::sys_common::backtrace::__rust_end_short_backtrace
  │    5: rust_begin_unwind
  │    6: core::panicking::panic_fmt
  │    7: core::panicking::panic
  │    8: core::option::unwrap_failed
  │    9: webrender::scene_builder_thread::SceneBuilderThread::process_transaction
  │   10: alloc::vec::in_place_collect::<impl alloc::vec::spec_from_iter::SpecFromIter<T,I> for alloc::vec::Vec<T>>::from_iter
  │   11: webrender::scene_builder_thread::SceneBuilderThread::run
  │   14: std::sys::pal::unix::thread::Thread::new::thread_start
  │   15: <unknown>
  │   16: <unknown>
  │ mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
  │ Caught signal 11thread panicked while processing panic. aborting.
  │ Redirecting call to abort() to mozalloc_abort
```

This comes from https://github.com/servo/webrender/blob/9d354ad/webrender/src/scene_builder_thread.rs#L517, so it looks like there's some missing synchronization with webrender around documents when we're trying to shutdown?

Taym95:
Fix panic in Webrender during shutdown while exiting `webgl_threads `

---

- [x] These changes fix #32867

- [X] These changes do not require tests because don't change behavior.


components/constellation/constellation.rs
```diff
@@ -2733,6 +2740,9 @@ where
         if let Err(e) = storage_ipc_receiver.recv() {
             warn!("Exit storage thread failed ({:?})", e);
         }
+        if let Err(e) = webgl_threads_receiver.recv() {
```
Taym95, line 2743: @gterzian If I understand this correctly, this will be blocking until it receives a value from `webgl_threads_sender`, which happens after `webgl_threads.exit()` has succeeded, right?

components/constellation/constellation.rs
```diff
@@ -2711,8 +2713,13 @@ where
 
         if let Some(webgl_threads) = self.webgl_threads.as_ref() {
             debug!("Exiting WebGL thread.");
-            if let Err(e) = webgl_threads.exit() {
-                warn!("Exit WebGL Thread failed ({})", e);
+            match webgl_threads.exit() {
```
gterzian, line 2716: You need to pass the sender here, and then send the reply when the exit message is handled in the webgl thread. 

See for example https://github.com/servo/servo/pull/32897/files#diff-95aa2eb890fe794918afdadfb6a89e026ec80222ceb8e5a21768d528e9ce036cR2698

gterzian (changes requested):
Thanks for taking this one on!

components/constellation/constellation.rs
gterzian: Yes but see below.

components/constellation/constellation.rs
```diff
@@ -2713,13 +2713,8 @@ where
...
             debug!("Exiting WebGL thread.");
-            match webgl_threads.exit() {
-                Ok(_) => {
-                    let _ = webgl_threads_sender.send(());
-                },
-                Err(e) => {
-                    warn!("Exit WebGL Thread failed ({})", e);
-                },
+            if let Err(e) = webgl_threads.exit(webgl_threads_sender) {
+                warn!("Exit WebGL Thread failed ({})", e);
```
mrobinson, line 2717: Small cleanup suggestion:

```suggestion
                warn!("Exit WebGL Thread failed ({e})");
```

components/shared/canvas/webgl.rs
```diff
@@ -73,9 +73,15 @@ impl WebGLThreads {
...
     /// Sends a exit message to close the WebGLThreads and release all WebGLContexts.
-    pub fn exit(&self) -> Result<(), &'static str> {
+    pub fn exit(
+        &self,
+        webgl_threads_sender: ipc_channel::ipc::IpcSender<()>,
+    ) -> Result<(), &'static str> {
         self.0
             .send(WebGLMsg::Exit)
+            .map_err(|_| "Failed to send Exit message")?;
+        webgl_threads_sender
```
gterzian, line 83: sorry I wasn't clear, this needs to go into the `Exit` message,

see for example https://github.com/servo/servo/blob/ca6169990ea8b429bbb8c11e4dcc9d8c6f59c7dd/components/net/resource_thread.rs#L375

gterzian: also we can follow the naming convention, so just `sender`.

gterzian (changes requested):

components/shared/canvas/webgl.rs
Taym95: Yes I pushed the wrong changes

components/shared/canvas/webgl.rs
```diff
@@ -106,7 +106,7 @@ pub enum WebGLMsg {
     /// request is fulfilled
     SwapBuffers(Vec<WebGLContextId>, WebGLSender<u64>, u64),
     /// Frees all resources and closes the thread.
-    Exit,
+    Exit(ipc_channel::ipc::IpcSender<()>),
```
gterzian, line 109: please import it at the top.

components/constellation/constellation.rs
```diff
@@ -2644,6 +2644,8 @@ where
             ipc::channel().expect("Failed to create IPC channel!");
         let (storage_ipc_sender, storage_ipc_receiver) =
             ipc::channel().expect("Failed to create IPC channel!");
+        let (webgl_threads_sender, webgl_threads_receiver) =
```
gterzian, line 2647: If there is a clean way to only create those if `Some(webgl_threads)`, we should do this.

components/constellation/constellation.rs
```diff
@@ -2733,6 +2735,9 @@ where
         if let Err(e) = storage_ipc_receiver.recv() {
             warn!("Exit storage thread failed ({:?})", e);
         }
+        if let Err(e) = webgl_threads_receiver.recv() {
```
gterzian, line 2738: I think this will deadlock if `self.webgl_threads == None`, so we should make this conditional to it being `Some`.

gterzian (changes requested):
One deadlock and some nits

components/constellation/constellation.rs
Taym95: only way I see is to use `Option<IpcSender<()>>` and create channel when `Some(webgl_threads)`

gterzian: Yes that is a good idea.

components/constellation/constellation.rs
Taym95: should be fixed now, ready for review again.

jdm: I don't think you pushed your changes.

Taym95: sorry my bad, pushed now.

components/constellation/constellation.rs
```diff
@@ -2733,6 +2740,13 @@ where
         if let Err(e) = storage_ipc_receiver.recv() {
             warn!("Exit storage thread failed ({:?})", e);
         }
+        if self.webgl_threads.is_some() {
+            if let Some(webgl_threads_receiver) = webgl_threads_receiver {
+                if let Err(e) = webgl_threads_receiver.recv() {
+                    warn!("Exit WebGL thread failed ({:?})", e);
+                }
+            }
+        }
```
mrobinson, line 2749: Is it impossible `webgl_threads_receiver` to be `Some(...)` when `webgl_threads` is None? If so, this should probably be an assertion instead. If not, matching a tuple is might be bit cleaner here. Either way, the error can be inline in the format statement.

```suggestion
        if let Some(webgl_threads_receiver) = webgl_threads_receiver {
            assert!(self.webgl_threads.is_some());
            if let Err(e) = webgl_threads_receiver.recv() {
                warn!("Exit WebGL thread failed ({e:?})");
            }
        }
```

```suggestion
        match (self.webgl_threads, webgl_threads_receiver)  {
            (Some(_), Some(webgl_threads_receiver)) => if let Err(e) = webgl_threads_receiver.recv() {
                    warn!("Exit WebGL thread failed ({e:?})");
             },
            _ => {},
        }
```

components/canvas/webgl_thread.rs
```diff
@@ -368,7 +368,10 @@ impl WebGLThread {
             WebGLMsg::SwapBuffers(swap_ids, sender, sent_time) => {
                 self.handle_swap_buffers(swap_ids, sender, sent_time);
             },
-            WebGLMsg::Exit => {
+            WebGLMsg::Exit(sender) => {
+                if let Err(e) = sender.send(()) {
+                    warn!("Failed to send response to WebGLMsg::Exit ({})", e)
```
mrobinson, line 373:
```suggestion
                    warn!("Failed to send response to WebGLMsg::Exit ({e})");
```

mrobinson (approved):
Looks good to me with a couple nits:

components/constellation/constellation.rs
Taym95: indeed tuple matching is better.

components/constellation/constellation.rs
```diff
@@ -2710,9 +2711,15 @@ where
         }
 
         if let Some(webgl_threads) = self.webgl_threads.as_ref() {
+            let (sender, receiver) = ipc::channel().expect("Failed to create IPC channel!");
+            _webgl_threads_sender = Some(sender);
```
gterzian, line 2715: Why the `_`?

components/constellation/constellation.rs
```diff
@@ -2644,6 +2644,7 @@ where
             ipc::channel().expect("Failed to create IPC channel!");
         let (storage_ipc_sender, storage_ipc_receiver) =
             ipc::channel().expect("Failed to create IPC channel!");
+        let (mut _webgl_threads_sender, mut webgl_threads_receiver) = (None, None);
```
gterzian, line 2647: You don't need the sender here, since it is only used inside the ` if let Some(webgl_threads)` block, so you can nest it to it, without wrapping it in an option.

components/constellation/constellation.rs
```diff
@@ -2710,9 +2711,15 @@ where
...
 
         if let Some(webgl_threads) = self.webgl_threads.as_ref() {
+            let (sender, receiver) = ipc::channel().expect("Failed to create IPC channel!");
+            _webgl_threads_sender = Some(sender);
+            webgl_threads_receiver = Some(receiver);
             debug!("Exiting WebGL thread.");
-            if let Err(e) = webgl_threads.exit() {
-                warn!("Exit WebGL Thread failed ({})", e);
+
+            if let Some(sender) = _webgl_threads_sender {
```
gterzian, line 2719: This conditional is not necessary if you follow my other comment.

components/constellation/constellation.rs
```diff
@@ -2710,9 +2711,15 @@ where
         }
 
         if let Some(webgl_threads) = self.webgl_threads.as_ref() {
```
gterzian, line 2713: I think you can assign an `Option<IPCReceiver>` here, with the `else` block returning `None`.

components/constellation/constellation.rs
```diff
@@ -2733,6 +2740,14 @@ where
         if let Err(e) = storage_ipc_receiver.recv() {
             warn!("Exit storage thread failed ({:?})", e);
         }
+        match (self.webgl_threads.take(), webgl_threads_receiver) {
```
gterzian, line 2743: I would do `if.webgl_threads.is_some()` and expect ` webgl_threads_receiver` to be some inside the block.

gterzian (approved):
Couple of small things to simplify the statements.

components/constellation/constellation.rs
Taym95: Indeed, I applied your suggestions

gterzian:
@Taym95 thanks for the work, and @mrobinson thanks for the review.

PR #32924, opened 2024-08-03
Enable WAV decoding

Melchizedek6809:
I was testing a game of mine and noticed that Servo on MacOS doesn't support WAV decoding, breaking my game. Adding gstwavparse to the gstreamer plugin list fixed that issue for me.

---

I didn't want to add new tests since for example /webaudio/the-audio-api/the-audiobuffersourcenode-interface/audiobuffersource-multi-channels.html fails on MacOS for me without these changes


jdm:
I guess the tests already pass on Linux because we don't use a limited plugin list like we do on macos and Windows.

jdm (approved):

PR #32928, opened 2024-08-04
build: Speed up first run after build on macOS

mrobinson:
GStreamer has to process plugins each time they are added when
initializing. When those files have changed, this triggers macOS
security protections which can add many seconds to access time. This
change eliminates that problem after the first packaging of libraries by
skipping packaging if everything is up-to-date and not overwriting the
dylibs everytime.

In addition, it moves a lot of the code for packaging GStreamer
libraries on macOS into the `gstreamer` module and adds type-safety and
comments to the Python.

---

- [x] These changes do not require tests because this just changes packaging behavior.

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -31,11 +31,12 @@ from mach.decorators import (
 from mach.registrar import Registrar
 
 import servo.platform
+import servo.platform.macos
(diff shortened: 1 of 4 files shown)


jdm (approved):
Very nice!

PR #32929, opened 2024-08-04
Lower `TRACKER_DASHBOARD_MAXIMUM_OUTPUT_LENGTH` limit

### LINKED ISSUE #31845: Handle HTTP Error 413: Request Entity Too Large more gracefully.
WPT import currently fails because of it: https://github.com/servo/servo/actions/runs/8405500156/job/23018617126#step:8:14419 and I've hit this when doing webgpu tests too. Ideally if too large we should split it into multiple request.

sagudev:
---

- [x] These changes fix #31845
- [x] These changes do not require tests because only limit change


mrobinson (approved):
I'm happy to try this, but 512 seems quite low already or is this being applied to every test?

sagudev:
> I'm happy to try this, but 512 seems quite low already or is this being applied to every test?

I think this is applied for every test/subtest message, but it still isn't enough: https://github.com/servo/servo/actions/runs/10235283414, so I tried to lower it to 128 (test import run: https://github.com/servo/servo/actions/runs/10235501047), but I suspect event this wouldn't be enough, due to number of subtests. We can limit number of subtest results, or simply do multiple request (one per each test + all of its subtests) in that case.

sagudev:
Still not working, but maybe we can just raise `client_max_body_size` in nginx, see https://github.com/servo/saltfs/pull/1054

mrobinson:
I think the easiest thing to do here is to simply send an empty `message` string in `make_data_from_result`. The `message` data is currently unused in the tracker, so I think it's fine for it to be blank for now.

mrobinson (changes requested):
Let's just make the `message` empty.

sagudev:
This doesn't work either: https://github.com/servo/servo/actions/runs/10236387404, as there are many subtests with very looooooooong names; ex: `CSS Animations: property <height> from [calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(calc-size(auto, size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2), size / 2 + size / 2)] to [calc-size(auto, size * 3)] at (1) should be [calc-size(auto, size * 3)]` and I think that this is causing problems for us (test that has long subtest names is 1w old, some time we have this issue: https://github.com/web-platform-tests/wpt/blob/master/css/css-values/calc-size/animation/calc-size-height-interpolation.html). I think the best way forward would still be https://github.com/servo/saltfs/pull/1054, as default 1MB limit is really small (JSON files are not really compact and we do not compress them either).

mrobinson:
Can this be closed now?

sagudev:
> Can this be closed now?

Yes.

PR #32931, opened 2024-08-04
layout: Simple port of legacy innerText getter

Melchizedek6809:
Hey there,

I've ported the functions used by the innerText getter from layout2013 to 2020, since right now it just returns "" all the time. While it's not great, now it at least returns something closer to the content.

---

There seem to be a lot of WPT tests already in html/dom/elements/the-innertext-and-outertext-properties, do these suffice, or do you prefer rust ones?

### DIFF
--- a/components/layout_2020/query.rs
+++ b/components/layout_2020/query.rs
@@ -12,10 +12,12 @@ use log::warn;
 use script_layout_interface::wrapper_traits::{
     LayoutNode, ThreadSafeLayoutElement, ThreadSafeLayoutNode,
 };
-use script_layout_interface::OffsetParentResponse;
+use script_layout_interface::{LayoutElementType, LayoutNodeType, OffsetParentResponse};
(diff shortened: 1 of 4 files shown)


jdm:
WPT tests are preferred.

Melchizedek6809:
I've just looked into the stable unexpected results, and the first one (`/css/css-conditional/container-queries/crashtests/svg-layout-root-crash.html`) seems a bit odd since it passes on my machines (tried it on a Mac and a Linux machine, both pass, though on the Mac it timed out the first time I tried it with a debug build), not sure how to handle that one.

The other 3 use innerText and probably pass because of that, I've updated the expected WPT results there and made another commit.

Melchizedek6809:
Just saw that there's already an open PR that implements outerText as well as innerText in a much more thorough manner: #32427
Hope that one gets merged soon

PR #32932, opened 2024-08-05
build: Fix a print statement when repackaging GStreamer

mrobinson:
I neglected to put make a string a formatted string, which means that
the variable won't be included in the output properly.

---

- [ ] These changes do not require tests because this fixes a print statement during the build.


mukilan (approved):

mukilan:
@mrobinson The commit is not signed.

PR #32949, opened 2024-08-06
Fix visual_studio.py to call vswhere with -utf8

hatoo:
In Japanese Windows, `./mach build` fails because `vswhere` puts cp932 string in default and fails to load json in Pyhton.
This fixes it by adding `-utf8` option

---

- [x] These changes do not require tests because it relates to `./mach build`


mrobinson (approved):

mrobinson:
@hatoo Thank you for this fix!

PR #32950, opened 2024-08-06
Add exception to rejection logic in `generic_call`

### LINKED ISSUE #32816: IDLharness test failures on operations that return promises
When there is an IDL harness test on an operation that returns a promise (such as `playEffect()` and `reset()` on GamepadHapticActuator, `getAvailability()` on Bluetooth, etc) the subtest for checking promise rejection when `this` is null/undefined fails. Relevant spots:

https://github.com/servo/servo/blob/8b3c9b744ab41f52f01ee624228e655b981824eb/tests/wpt/tests/resources/idlharness.js#L113
https://github.com/servo/servo/blob/8b3c9b744ab41f52f01ee624228e655b981824eb/tests/wpt/tests/resources/idlharness.js#L2395

sagudev:
Relevant code in mozilla uses templates (see `ExceptionPolicy`), but we can use const generic to achieve same effect.

---

- [x] These changes fix #32816

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -1870,13 +1870,17 @@ class MethodDefiner(PropertyDefiner):
             else:
                 selfHostedName = "0 as *const libc::c_char"
                 if m.get("methodInfo", True):
+                    if m.get("returnsPromise", False):
+                        exceptionToRejection = "true"
+                    else:
+                        exceptionToRejection = "false"
                     identifier = m.get("nativeName", m["name"])
                     # Go through an intermediate type here, because it's not
                     # easy to tell whether the methodinfo is a JSJitInfo or
                     # a JSTypedMethodJitInfo here.  The compiler knows, though,
                     # so let it do the work.
                     jitinfo = "&%s_methodinfo as *const _ as *const JSJitInfo" % identifier
-                    accessor = "Some(generic_method)"
+                    accessor = f"Some(generic_method::<{exceptionToRejection}>)"
                 else:
                     if m.get("returnsPromise", False):
                         jitinfo = "&%s_methodinfo" % m.get("nativeName", m["name"])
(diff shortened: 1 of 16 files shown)


sagudev:
Hah, all browser fail on `PromiseRejectionEvent interface: attribute promise`: https://wpt.fyi/results/html/dom/idlharness.worker.html?label=master&label=experimental&aligned&q=html%2Fdom%2Fidlharness.worker.html

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1870,13 +1870,17 @@ def specData(m):
             else:
                 selfHostedName = "0 as *const libc::c_char"
                 if m.get("methodInfo", True):
+                    if m.get("returnsPromise", False):
+                        exceptionToRejection = "true"
+                    else:
+                        exceptionToRejection = "false"
```
Taym95, line 1876:
```suggestion
                   exceptionToRejection  = str(m.get("returnsPromise", False))
```

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1977,10 +1981,14 @@ def getter(attr):
                 accessor = 'get_' + self.descriptor.internalNameFor(attr.identifier.name)
                 jitinfo = "0 as *const JSJitInfo"
             else:
+                if attr.type.isPromise():
+                    exceptionToRejection = "true"
+                else:
+                    exceptionToRejection = "false"
```
Taym95, line 1987: use str here? `str(attr.type.isPromise()`

sagudev: Python has uppercase `True`/`False`, we could use `str(True).lower()`, but I think that would be to obfuscated. Also ideally we would use enum here, but cons generics does not support it yet.

jdm (approved):
Nice solution!

PR #32958, opened 2024-08-07
webxr: Update hand input to match latest spec

msub2:
This updates the implementation of WebXR Hand Input in Servo to match the latest published version of the spec.

---


msub2:
The idlharness test for hand-input is the only file in the meta folders, should they be kept there (empty) or should the whole folder be removed?

gterzian:
I think it can be removed.

tests/wpt/meta/webxr/__dir__.ini
```diff
@@ -0,0 +1 @@
+prefs: [dom.webxr.hands.enabled: true]
```
jdm, line 1: Should we enable this pref by default now? Originally this was introduced because the hands spec was still very new and being iterated on rapidly.

msub2: We probably can yeah, the hands spec has pretty solid for a while now

jdm (approved):

jdm:
Oof, rebase required.

msub2:
ah whoops, shoulda seen that coming, will fix

msub2:
@jdm This should be good to go now

PR #32960, opened 2024-08-07
Initial structuredClone implementation

### LINKED ISSUE #32770: structuredClone is not available
**Describe the bug:**
Following message in servo output:
```
[script::dom::bindings::error] Error at https://www.selfcad.com/deployments/20240709030454802/modern-scripts/startup-bundle.js.gzip:17:143088 structuredClone is not defined
```
Well, it is not that kind of error like scripts not working at all, but this specific function is not available.
Function in question: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

**To Reproduce:**
I suppose that even minimal page with script like 
```
console.info(structuredClone({ some: "value", other: 1, test: true, values: [1, NaN, null] }))
```
will fail with such error.

**Platform:**
Ubuntu 24.04, DELL Inspiron PC with Nvidia GeForce GTX 16XX series.

Taym95:
---

- [x] These changes fix #32770


components/script/dom/webidls/WindowOrWorkerGlobalScope.webidl
```diff
@@ -28,6 +28,11 @@ interface mixin WindowOrWorkerGlobalScope {
   Promise<ImageBitmap> createImageBitmap(ImageBitmapSource image, optional ImageBitmapOptions options = {});
   // Promise<ImageBitmap> createImageBitmap(
   //   ImageBitmapSource image, long sx, long sy, long sw, long sh, optional ImageBitmapOptions options);
+
+  // structured cloning
+  [Throws]
+  // any structuredClone(any value);
+  any structuredClone(any value, optional StructuredSerializeOptions options = {});
```
Taym95, line 35: codegen is not working for `StructuredSerializeOptions ` I am getting:

```
   File "servo/third_party/WebIDL/WebIDL.py", line 2800, in complete
      raise WebIDLError("Unresolved type '%s'." % self.name, [self.location])
  WebIDL.WebIDLError: error: Unresolved type '<unresolved scope>::StructuredSerializeOptions'., /servo/components/script/dom/bindings/codegen/../../webidls/WindowOrWorkerGlobalScope.webidl line 35:42
    any structuredClone(any value, optional StructuredSerializeOptions options = {});
    ```
    
    I am trying to figure it out

**Comment by Taym95** [comment]
FYI @gterzian

File: components/script/dom/webidls/WindowOrWorkerGlobalScope.webidl (hunk 1 above)
Comment (gterzian) [reply components/script/dom/webidls/WindowOrWorkerGlobalScope.webidl:35]: I think the name changed, we have it at https://github.com/servo/servo/blob/3c271fb2989fa838473ab006397c7a9d8c3b4b21/components/script/dom/webidls/MessagePort.webidl#L21, so it would have to be defined as per https://html.spec.whatwg.org/multipage/web-messaging.html#structuredserializeoptions 

I think this can be done as part of this PR.

Comment (Taym95) [reply components/script/dom/webidls/WindowOrWorkerGlobalScope.webidl:35]: Indeed it changed, I will update it.

Comment (Taym95) [reply components/script/dom/webidls/WindowOrWorkerGlobalScope.webidl:34]: Renamed `PostMessageOptions` to `StructuredSerializeOptions`, ready for review

**Review by gterzian — CHANGES_REQUESTED** [review]

File: components/script/dom/workerglobalscope.rs [hunk 2]
```diff
@@ -431,6 +432,22 @@ impl WorkerGlobalScopeMethods for WorkerGlobalScope {
     fn IsSecureContext(&self) -> bool {
         self.upcast::<GlobalScope>().is_secure_context()
     }
+
+    /// <https://html.spec.whatwg.org/multipage/#dom-structuredclone>
+    fn StructuredClone(&self, cx: JSContext, value: HandleValue) -> Fallible<js::jsval::JSVal> {
```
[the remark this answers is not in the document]
Taym95: I created common StructuredClone function in globalscope

Taym95:
@mrobinson I updated WPT test and are passing in my fork, we can run it again.

components/script/dom/webidls/MessagePort.webidl
```diff
@@ -18,6 +18,7 @@ interface MessagePort : EventTarget {
   attribute EventHandler onmessageerror;
 };
 
-dictionary PostMessageOptions {
+// Used to declare which objects should be transferred.
```
gterzian, line 21: this is only true for the `transfer` part, I think the idea is that more options could added later. Anyway I don't think we need the comment here actually, because I don't see it either at https://html.spec.whatwg.org/multipage/web-messaging.html#messageport

gterzian (approved):
Awesome, thanks! I think there is just one misplaced comment that can be removed...

components/script/dom/webidls/MessagePort.webidl
Taym95: Copied it from Firefox, I will remove it.

Taym95:
```suggestion
```

PR #32966, opened 2024-08-08
webgpu: separate Queue&Device Id

### LINKED ISSUE #32222: Separate queue and device id in webgpu
Wgpu is starting to separate them, currently queue_id still equals to device_id, but it is expected that this will change in the future.

See transmute trait/method from https://github.com/servo/servo/pull/32173.

sagudev:
This will be needed for next wgpu upgrade.

---

- [x] These changes fix #32222


mrobinson (approved):
It's great to see less unsafe code.

PR #32991, opened 2024-08-09
Set the cfg properly for the production-stripped profile

webbeef:
When building with `./mach build --profile production-stripped`, the `servo_production` cfg was not set properly because the check only considered the `production` build profile. This PR set the right cfg for both `production` and any `production-xyz` build profile.

Also added the required cargo output for the ` ohos_mock` cfg in fonts.

---

- [X] These changes do not require tests because no relevant test is needed.

### DIFF
--- a/components/fonts/Cargo.toml
+++ b/components/fonts/Cargo.toml
@@ -49,6 +49,9 @@ webrender_api = { workspace = true }
 webrender_traits = { workspace = true }
 xi-unicode = { workspace = true }
 
+[lints.rust]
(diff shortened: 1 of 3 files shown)


components/fonts/build.rs
mukilan, line 1: It might be better to declare this in Cargo.toml instead of creating a build.rs

mukilan (commented):
Thank you for this fix!

mukilan (approved):

PR #33003, opened 2024-08-11
clippy: Fix various clippy warnings throughout the code 

arihant2math:
Overall formatting with clippy. However I did change one thing in 74004c8, I assumed that the values were swapped.

---

- [x] These changes fix part of #31500
- [x] These changes do not require tests because they are just formatting

### DIFF
--- a/components/canvas/webgl_thread.rs
+++ b/components/canvas/webgl_thread.rs
@@ -1085,7 +1085,8 @@ impl WebGLImpl {
                 state.restore_depth_invariant(gl);
             },
             WebGLCommand::DepthRange(near, far) => {
-gl.depth_range(near.max(0.).min(1.) as f64, far.max(0.).min(1.) as f64)            },
+                gl.depth_range(near.max(0.).min(1.) as f64, far.max(0.).min(1.) as f64)
+            },
             WebGLCommand::Disable(cap) => match cap {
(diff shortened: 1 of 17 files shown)


mukilan (changes requested):
Thank you for the contribution! The PR looks good except for the changes related to min/max code.

1. Regarding 74004c8, the original code is correct. We want the value of page_zoom to be between `MIN_ZOOM` and `MAX_ZOOM`. `result = zoom.max(MIN_ZOOM)` will ensure result is at least `MIN_ZOOM` if `zoom` is lesser than MIN_ZOOM and `result.min(MAX_ZOOM)` will ensure the final value is at most `MAX_ZOOM` if result is greater than `MAX_ZOOM`.

2. According the [documentation](https://rust-lang.github.io/rust-clippy/master/index.html#/clamp), min and max don't propagate `NaN` values while `clamp` does. Replacing min & max with clamp could therefore introduce panic or other unexpected behaviour since the final result now can be NaN.

I think we should revert the changes related to min/man -> clamp in this PR and make the change only after auditing the surrounding code where it is used to ensure the NaN handling of clamp is acceptable or desired for that code.

arihant2math:
Sure, I've just done that.

mukilan (approved):
Thank you!

PR #33004, opened 2024-08-12
Replace the lazy_static crate whth `std::sync::LazyLock` in components/script

Hmikihiro:
Replace the lazy_static crate whth `std::sync::LazyLock` in components/script.

I removed lazy_static In `script_runtime.rs` because it uses only `Mutex::new` which can be used in const statements.
This part is initialized at compile time instead of on first access.
The other parts do not change behavior.

---


mrobinson (approved):
Nice.

PR #33009, opened 2024-08-12
webgpu: Sync various parts of spec

sagudev:
From #30504, although redone.

Most changes are mechanical and reviewable per commit.

---

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -193,7 +193,8 @@ impl GPUDevice {
 
     /// <https://gpuweb.github.io/gpuweb/#abstract-opdef-validate-texture-format-required-features>
     ///
-    /// Also returns wgpu's `TextureFormat` as it already does conversion inside
+    /// Validates that the device suppports required features,
+    /// and if so returns an ok containing wgpu's `TextureFormat`
     pub fn validate_texture_format_required_features(
         &self,
         format: &GPUTextureFormat,
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -18,12 +18,13 @@ use webgpu::wgc::pipeline::RenderPipelineDescriptor;
 use webgpu::wgc::{
     binding_model as wgpu_bind, command as wgpu_com, pipeline as wgpu_pipe, resource as wgpu_res,
 };
+use webgpu::wgt::TextureFormat;
 use webgpu::{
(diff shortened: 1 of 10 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -190,6 +191,27 @@ impl GPUDevice {
         let _ = self.eventtarget.DispatchEvent(ev.event());
     }
 
+    /// <https://gpuweb.github.io/gpuweb/#abstract-opdef-validate-texture-format-required-features>
+    ///
+    /// Also returns wgpu's `TextureFormat` as it already does conversion inside
```
gterzian, line 196: I think this would be clearer: "Validates that the device suppports required features, and if so returns an ok containing wgpu's `TextureFormat`"

components/script/dom/gpudevice.rs
```diff
@@ -781,33 +846,37 @@ impl GPUDeviceMethods for GPUDevice {
...
-                format: convert_texture_format(descriptor.format),
-                usage: usg,
-                view_formats: descriptor
-                    .viewFormats
-                    .iter()
-                    .map(|tf| convert_texture_format(*tf))
-                    .collect(),
-            }
-        });
+        // TODO(sagudev): We should pass invalid bits to wgpu
```
gterzian, line 857: should or should not?

components/script/dom/gpusupportedfeatures.rs
```diff
@@ -39,43 +39,66 @@ pub struct GPUSupportedFeatures {
...
+        if features.contains(Features::RG11B10UFLOAT_RENDERABLE) {
+            set.insert(GPUFeatureName::Rg11b10ufloat_renderable);
+        }
+        if features.contains(Features::BGRA8UNORM_STORAGE) {
+            set.insert(GPUFeatureName::Bgra8unorm_storage);
+        }
+        if features.contains(Features::FLOAT32_FILTERABLE) {
+            set.insert(GPUFeatureName::Float32_filterable);
+        }
+        // TODO: clip-distances when wgpu supports it
```
gterzian, line 92: For consistency's sake, this todo could have the same format as the others added in this PR.

gterzian (approved):
LGTM with nits.

components/script/dom/gpudevice.rs
sagudev: we should, because wgpu-core will do check for invalid bits and generate proper validation errors.

components/script/dom/gpusupportedfeatures.rs
sagudev: idk what you mean (the sagudev part?)

PR #33024, opened 2024-08-12
Initial IndexedDB Support

arihant2math:
This is a partial rewrite of #25214.


mrobinson:
It seems that this PR contains many changes unrelated to IndexedDB support. Can you remove from this branch and keep only the related changes? Thank you.

arihant2math:
Seems like my branching on the fork is messed up, I'll see if I can fix it

PR #33026, opened 2024-08-13
clippy: Fix missing indentation in comments and remove on unecessary cast

DanielFrantes:
Fixing clippy warnings

---

- [X] These changes do not require tests because don't change logic of code


danik292:
> @danik292  You will need to sign your commit (use `--amend` with `-s`). See https://wiki.linuxfoundation.org/dco.

So git --amend -s  with no files in staging arena,If I understand.

mrobinson:
> So git --amend -s with no files in staging arena,If I understand.

The easiest way to do this is to squash all of your commits into a single commit and then simply `git --amend -s`. If there are files in the staging area, they are automatically committed when you commit, so they will also be committed when you amend. It's best not to have files in the staging area if you do not plan to commit them.

danik292:
@mrobinson  I singed off both my commits.

mrobinson (approved):

mrobinson:
Thanks for the contribution. I will land this now.

PR #33027, opened 2024-08-13
Add support for structured logging with tracing

delan:
Previously handled by [env_logger](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/struct.EnvFilter.html), RUST_LOG is now handled by the tracing [EnvFilter](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/struct.EnvFilter.html), whose syntax is more restrictive. In particular, the target name in a filter directive can only contain characters in the class `[\w:-]`, which breaks our [event tracing filters](https://book.servo.org/hacking/debugging.html). This patch renames those event tracing targets as follows:

- `<` → `-from-`
- `>` → `-to-`
- `@` → `:`
- `(Foo)` → `:Foo`

For example, `servoshell<winit@NewEvents(WaitCancelled)` is now `servoshell-from-winit:NewEvents:WaitCancelled`. I’ve updated the book’s [docs about this](https://book.servo.org/hacking/debugging.html#event-tracing) in servo/book#17.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4612,6 +4612,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "nu-ansi-term"
+version = "0.46.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77a8165726e8236064dbb45459242600304b42a5ea24ee2948e18e023bf7ba84"
+dependencies = [
+ "overload",
+ "winapi",
+]
+
+[[package]]
 name = "num-bigint"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


mrobinson:
Would it make sense to switch all of Servo from `log` to `tracing`?

delan:
> Would it make sense to switch all of Servo from `log` to `tracing`?

Probably! In fact, by default, `tracing` installs a logger that catches any logging via `log` and [turns it into tracing events](https://crates.io/crates/tracing-log). We’ll need to rework our inter-process crash logging, but that seems very doable.

jschwe:
> Probably! In fact, by default, tracing installs a logger that catches any logging via log and [turns it into tracing events](https://crates.io/crates/tracing-log). We’ll need to rework our inter-process crash logging, but that seems very doable.

On OpenHarmony logging messages is done via `hilog`, while tracing is done via `hitrace`. I think it is similar on Android (logcat and [system tracing](https://developer.android.com/topic/performance/tracing/custom-events-native)).
I believe it is important that we retain the ability to differentiate between what goes to the normal log and what goes to the tracing backend.

The main motivation is that logging is most likely always active, but the tracing receiver only processes messages when the developer actively enabled system tracing, otherwise the events just go to /dev/null.

delan:
Replied in [this comment](https://github.com/servo/servo/issues/33486#issuecomment-2359851826), but not longer pursuing this right now.

PR #33030, opened 2024-08-13
layout: Convert the FragmentTree to physical geometry

### LINKED ISSUE #25564: Scrollable overflow calculation should take into account containing block

mrobinson:
This converts all geometry in the FragmentTree into physical geometry,
doing conversions ahead of time instead of when traversing the fragment
tree. This is necessary to properly implement BiDi in Servo as we need
to know what side borders are on in mixed RTL and LTR contexts.

In addition, fragments are laid out in a particular context and only
that context knows its writing mode. There were issues where were using
one writing mode to lay out and another to convert to phyisical
coordinates. This isn't an issue now since we only use the default
writing mode, but starts to be an issue with BiDi text.

Closes #25564.

---

- [x] These changes do not require tests because they shouldn't change behavior
      when we don't support any non-default writing modes at all.

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -676,8 +676,10 @@ impl FlexContainer {
                     all_baselines.last = line_all_baselines.last;
                 }
 
+                let physical_line_position =
+                    flow_relative_line_position.to_physical_size(self.style.writing_mode);
                 for (fragment, _) in &mut final_line_layout.item_fragments {
-                    fragment.content_rect.start_corner += flow_relative_line_position
+                    fragment.content_rect.origin += physical_line_position;
                 }
                 final_line_layout.item_fragments
             })
@@ -705,6 +707,7 @@ impl FlexContainer {
                     let fragment = Fragment::Box(fragment);
                     child_positioning_context.adjust_static_position_of_hoisted_fragments(
                         &fragment,
+                        self.style.writing_mode,
                         PositioningContextLength::zero(),
                     );
                     positioning_context.append(child_positioning_context);
@@ -1545,15 +1548,20 @@ impl InitialFlexLineLayout<'_> {
                 let mut fragment_info = item.box_.base_fragment_info();
                 fragment_info.flags.insert(FragmentFlags::IS_FLEX_ITEM);
 
+                let container_writing_mode = flex_context.containing_block.style.writing_mode;
                 (
                     BoxFragment::new(
                         fragment_info,
                         item.box_.style().clone(),
                         item_layout_result.fragments,
-                        content_rect,
-                        flex_context.sides_to_flow_relative(item.padding),
-                        flex_context.sides_to_flow_relative(item.border),
-                        margin,
+                        content_rect.to_physical(container_writing_mode),
+                        flex_context
+                            .sides_to_flow_relative(item.padding)
+                            .to_physical(container_writing_mode),
+                        flex_context
+                            .sides_to_flow_relative(item.border)
+                            .to_physical(container_writing_mode),
+                        margin.to_physical(container_writing_mode),
                         None, /* clearance */
                         collapsed_margin,
                     ),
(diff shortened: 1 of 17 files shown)


components/layout_2020/fragment_tree/positioning_fragment.rs
```diff
@@ -6,23 +6,21 @@ use app_units::Au;
...
 /// generated by boxes.
 #[derive(Serialize)]
 pub(crate) struct PositioningFragment {
     pub base: BaseFragment,
-    pub rect: LogicalRect<Au>,
+    pub rect: PhysicalRect<Au>,
     pub children: Vec<ArcRefCell<Fragment>>,
-    pub writing_mode: WritingMode,
-
+    //pub writing_mode: WritingMode,
```
mukilan, line 23: Should this line be deleted?

components/layout_2020/geom.rs
```diff
@@ -522,3 +538,37 @@ impl From<LogicalRect<CSSPixelLength>> for LogicalRect<Au> {
         }
     }
 }
+
+pub(crate) trait ToLogical<Unit, LogicalType> {
+    fn to_logical(&self, writing_mode: WritingMode) -> LogicalType;
+}
+
+impl<Unit: Clone + Copy> ToLogical<Unit, LogicalRect<Unit>> for PhysicalRect<Unit> {
```
mukilan, line 546: I think the `Clone` bound is not required here as `Copy` implies `Clone`.

components/layout_2020/query.rs
```diff
@@ -456,31 +442,28 @@ fn process_offset_parent_query_inner(
...
-                        },
-                        Fragment::AbsoluteOrFixedPositioned(_) |
-                        Fragment::Text(_) |
-                        Fragment::Image(_) |
-                        Fragment::IFrame(_) |
-                        Fragment::Positioning(_) => None,
-                    }
-                })
-                .unwrap()
+                    .find(|fragment, _, containing_block| {
```
mukilan, line 445: Is the additional indentation here deliberate?

components/layout_2020/flexbox/layout.rs
```diff
@@ -220,7 +220,7 @@ impl FlexContainer {
         let mut item_infos = vec![];
 
         let container_is_horizontal = self.style.writing_mode.is_horizontal();
-        let flex_direction = used_flex_direction(&self.style);
+        let flex_direction = used_flex_direction(&*self.style);
```
mukilan, line 223: I think this change is not needed because the code used to compile before (due to deref coersion).

components/layout_2020/flexbox/layout.rs
```diff
@@ -1545,15 +1548,20 @@ impl InitialFlexLineLayout<'_> {
                 let mut fragment_info = item.box_.base_fragment_info();
                 fragment_info.flags.insert(FragmentFlags::IS_FLEX_ITEM);
 
+                let container_writing_mode = flex_context.containing_block.style.writing_mode;
```
mukilan, line 1551: We use the name `containing_block_writing_mode` in few other places. Does that have the same meaning as `container_writing_mode`?

components/layout_2020/fragment_tree/positioning_fragment.rs
mrobinson: Yes. It should be deleted now.

components/layout_2020/geom.rs
mrobinson: Great point. I've removed `Clone` here.

components/layout_2020/query.rs
mrobinson: No, but it's very curious that `rustfmt` did not remove it. 🤔

components/layout_2020/flexbox/layout.rs
mrobinson: Hrm. I think this was a weird merge issue. I've resolved it now.

components/layout_2020/flexbox/layout.rs
mrobinson: In this case I'm just being a bit more explicit since it is the writing mode of the flex container that we are using. "Containing block" is used liberally in other places, but it isn't exactly accurate -- though this is larger issue we need to fix. For instance, for absolutely positioned elements, they may have a different containing block further up the tree.

mrobinson:
@mukilan Thanks for the review. I think I have addressed all of your concerns.

mukilan (approved):

PR #33032, opened 2024-08-13
layout: Support `start` and `end` values for flexbox `align-self`

mrobinson:
These are similar to `flex-start` and `flex-end`, but in `wrap-reverse`
 situations, they are the  opposite.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -1554,6 +1554,7 @@ impl InitialFlexLineLayout<'_> {
                         .baseline_relative_to_margin_box
                         .unwrap_or_default(),
                     shared_alignment_baseline.unwrap_or_default(),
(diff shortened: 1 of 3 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -1874,24 +1875,34 @@ impl FlexItem<'_> {
         line_cross_size: Au,
         propagated_baseline: Au,
         max_propagated_baseline: Au,
+        wrap_reverse: bool,
     ) -> Au {
+        let ending_alignment = line_cross_size - (*used_cross_size + self.pbm_auto_is_zero.cross);
```
Loirooriol, line 1880: The parenthesis could be removed, but fine either way

```suggestion
        let ending_alignment = line_cross_size - *used_cross_size - self.pbm_auto_is_zero.cross;
```

Loirooriol (approved):

PR #33038, opened 2024-08-13
De-duplicate `syn`

### LINKED ISSUE #33034: Dedupe syn 1
The blockers for removing the syn 1 dependency have been resolved and now there are only a few crates left to update to get rid of it:

- async-recursion in the net crate should be updated or removed
- enum-iterator in the script crate should be updated or removed
- syn should be updated in the servo-media-derive crate

mrobinson:
This is the last step toward removing our use of `syn` version. 
It does three things:

1. Upgrades `async-recursion` to a newer version that uses `syn` 2.
2. Removes the use of `enum-iterator` that was only used to produce a
   trivial list of enum names. This reduces the number of crates we
   depend on by 2.
3. Upgrades `media` to a version which no longer uses `syn` 1

Fixes #33034.

---

### DIFF
--- a/components/script/task_source/mod.rs
+++ b/components/script/task_source/mod.rs
@@ -18,16 +18,16 @@ pub mod websocket;
 
 use std::result::Result;
 
-use enum_iterator::IntoEnumIterator;
-
 use crate::dom::globalscope::GlobalScope;
 use crate::task::{TaskCanceller, TaskOnce};
 
-// The names of all task sources, used to differentiate TaskCancellers.
-// Note: When adding a task source, update this enum.
-// Note: The HistoryTraversalTaskSource is not part of this,
-// because it doesn't implement TaskSource.
-#[derive(Clone, Eq, Hash, IntoEnumIterator, JSTraceable, PartialEq)]
+/// The names of all task sources, used to differentiate TaskCancellers. Note: When adding a task
+/// source, update this enum. Note: The HistoryTraversalTaskSource is not part of this, because it
+/// doesn't implement TaskSource.
+///
+/// Note: When adding or removing a [`TaskSourceName`], be sure to also update the return value of
+/// [`TaskSourceName::all`].
+#[derive(Clone, Copy, Eq, Hash, JSTraceable, PartialEq)]
 pub enum TaskSourceName {
     DOMManipulation,
     FileReading,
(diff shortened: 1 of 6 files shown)


components/script/task_source/mod.rs
```diff
@@ -18,16 +18,16 @@ pub mod websocket;
...
-// Note: The HistoryTraversalTaskSource is not part of this,
-// because it doesn't implement TaskSource.
-#[derive(Clone, Eq, Hash, IntoEnumIterator, JSTraceable, PartialEq)]
+/// The names of all task sources, used to differentiate TaskCancellers. Note: When adding a task
+/// source, update this enum. Note: The HistoryTraversalTaskSource is not part of this, because it
+/// doesn't implement TaskSource.
+///
+/// Note: When adding or removing a [`TaskSourceName`], be sure to also update the return value of
+/// [`TaskSourceName::all`].
+#[derive(Clone, Copy, Eq, Hash, JSTraceable, PartialEq)]
```
sagudev, line 30: While I am not really fan of manual notes, but I think it's okay here because it already has them and the we get one dep less.

sagudev (approved):

components/script/task_source/mod.rs
```diff
@@ -18,16 +18,16 @@ pub mod websocket;
...
 use crate::dom::globalscope::GlobalScope;
 use crate::task::{TaskCanceller, TaskOnce};
 
-// The names of all task sources, used to differentiate TaskCancellers.
-// Note: When adding a task source, update this enum.
-// Note: The HistoryTraversalTaskSource is not part of this,
-// because it doesn't implement TaskSource.
-#[derive(Clone, Eq, Hash, IntoEnumIterator, JSTraceable, PartialEq)]
+/// The names of all task sources, used to differentiate TaskCancellers. Note: When adding a task
+/// source, update this enum. Note: The HistoryTraversalTaskSource is not part of this, because it
```
mukilan, line 25: Not related to the changes in this PR, but is this comment outdated? The note says HistoryTraversalTaskSource is not part of the enum, but it does have the `HistoryTraversal` variant.

mrobinson: Interesting. Looks like the `HistoryTraversal` variant was introduced at the same time (671627e97e20ae4baf728ae6dda61ef6f857c193) as the enum so the comment may have always been wrong? @gterzian, any idea what might be up here?

PR #33043, opened 2024-08-13
Fix warnings after latest rust upgrade

mrobinson:
This fixes various unused code warnings after the recent rust upgrade.
Some of the dead code is maintained, as it is quite likely that it will
be used in future changes.

---

### DIFF
--- a/components/script/dom/mediaquerylist.rs
+++ b/components/script/dom/mediaquerylist.rs
@@ -22,8 +22,8 @@ use crate::dom::document::Document;
 use crate::dom::eventtarget::EventTarget;
 
 pub enum MediaQueryListMatchState {
-    Same(bool),
-    Changed(bool),
+    Same,
+    Changed,
 }
 
 #[dom_struct]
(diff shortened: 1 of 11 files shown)


components/script/dom/mediaquerylist.rs
```diff
@@ -22,8 +22,8 @@ use crate::dom::document::Document;
 use crate::dom::eventtarget::EventTarget;
 
 pub enum MediaQueryListMatchState {
-    Same(bool),
-    Changed(bool),
+    Same(),
+    Changed(),
```
sagudev, line 26:
```suggestion
    Same,
    Changed,
```

sagudev (approved):
LGTM, after suggestion is resolved.

components/script/dom/mediaquerylist.rs
sagudev: Will also require changes in uses of `MediaQueryListMatchState`.

Issue #32459, opened 2024-06-07
Text rendering issue in servo.org footer


atouchet:
**Describe the bug:**
Text in the servo.org footer isn't rendered correctly.

Image:
![Untitled-2](https://github.com/servo/servo/assets/26315797/00533393-90a6-4f67-97a8-fb8cf2dfb51e)

**To Reproduce:**
Load https://servo.org/ and scroll to the bottom of the page.

**Platform:**
Windows 10 x86-64

atouchet:
After some more testing I've found that the issue doesn't occur 100% of the time on new page loads. Also as far as I can tell the first build I've been able to reproduce the issue is 2024-04-23.

Regression range: https://github.com/servo/servo/compare/025a98773223412edd8a0032a44c60ee84d69eee...7ca920927c7e5f25e59557f825c4774381750e01

atouchet:
There is a font initialization PR in that range that may be related: #32127

cc @mukilan

atouchet:
I am not sure what triggers it but I just tried the latest nightly build and the text looked fine on the initial load. However if I closed Servo and relaunched it the text issues would show up on every subsequent page load.

mrobinson:
@atouchet Can you try running Servo with the fix from #32499 applied and report on your results?

mrobinson:
Update, #32499 is now merged, so maybe you can just try on the latest from main @atouchet? Thankyou!

atouchet:
I just tested the lightest nightly build (2024-06-18) which includes 982a6d33818ceb5f19d9baf94f32f0607ca979ed and unfortunately the problem doesn't appear to be fixed. It took a few page reloads to run into the issue this time but the text rendering problem is still there.

atouchet:
I've found another instance of weird text rendering on https://duckduckgo.com/ I'm not 100% sure if this is the exact same issue or not.

![Untitled-1](https://github.com/servo/servo/assets/26315797/d50d5871-7f13-4777-a3c7-f7c7e92b7335)

There appears to be problems with the subpixel rendering. Bright red and teal colours show up in certain words and letters. Strangely only some of them are displayed like this.

mukilan:
Is [this](https://github.com/servo/servo/pull/33045) PR a potential fix for this issue?

crbrz:
[...]

PR #33045, opened 2024-08-13
Use `FontInstanceFlags::SUBPIXEL_POSITION` for font instances on Windows

crbrz:
This fixes some text clipping issues on Windows. Without sub-pixel position flag the glyphs are slightly off position and depending on the font used the text can be clipped.

---

- [x] These changes partially address #32459.
- [x] These changes do not require tests because tests aren't run for Windows yet.

This is an example of text using Arial font that was clipped on the right side:
[text.webm](https://github.com/user-attachments/assets/64b3a0ef-5385-4987-ac3d-8d69aed27a7a)

### DIFF
--- a/components/fonts/platform/windows/font.rs
+++ b/components/fonts/platform/windows/font.rs
@@ -272,7 +272,7 @@ impl PlatformFontMethods for PlatformFont {
     }
 
     fn webrender_font_instance_flags(&self) -> FontInstanceFlags {
-        FontInstanceFlags::empty()
+        FontInstanceFlags::SUBPIXEL_POSITION
     }
 
     fn typographic_bounds(&self, glyph_id: GlyphId) -> Rect<f32> {


mrobinson:
It looks like this is always used in Gecko unless GDI mode is forced to be on: https://searchfox.org/mozilla-central/rev/4496b3ed9bb535832e4826f09fbcb645b559a32d/gfx/2d/ScaledFontDWrite.cpp#486

We should likely turn this on for both MacOS and Linux as well:
- https://searchfox.org/mozilla-central/rev/4496b3ed9bb535832e4826f09fbcb645b559a32d/gfx/2d/ScaledFontMac.cpp#473
- https://searchfox.org/mozilla-central/rev/4496b3ed9bb535832e4826f09fbcb645b559a32d/gfx/2d/ScaledFontFreeType.cpp#93

mrobinson (approved):

PR #33046, opened 2024-08-14
Replace the lazy_static crate with std::sync::LazyLock in components/net

Hmikihiro:
Replace the `lazy_static` crate with `std::sync::LazyLock` in components/net

---

- [x] These changes do not require tests because `lazy_static` and `LazyLock` behave in the same way.


mukilan (approved):
Thank you!

PR #33047, opened 2024-08-14
Dont use lazy static to construct mutexes

simonwuelker:
There are a few snippets of around the codebase that like this:

```rust
lazy_static::lazy_static! {
    static ref SERIAL: Mutex<()> = Mutex::new(());
}
```
Because `Mutex::new` is const (since rust `1.63.0`) we can instead initialize the Mutex at compile time like this:
```rust
static SERIAL: Mutex<()> = Mutex::new(());
```
This removes the dependency of `servoshell` and `background-hang-monitor` on `lazy-static`. There are (probably) more, but after these changes I noticed that @Hmikihiro is also working on removing `lazy-static` and I don't want to duplicate their work. 

---
- [X] These changes do not require tests because *they are trivial, atomic replacements of code patterns*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6187,7 +6187,6 @@ dependencies = [
  "ipc-channel",
  "jni",
  "keyboard-types",
- "lazy_static",
  "libc",
  "libloading",
  "libservo",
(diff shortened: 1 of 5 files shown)


mrobinson (approved):

Wuelle:
(sorry, I accidentally ran `./mach test-tidy` on the wrong branch >.> )

PR #33073, opened 2024-08-15
legacy-layout: Fix display list building after WebRender upgrade

mrobinson:
The most recent version of WebRender tracks stacking context offsets in
a different way, which broke legacy layout. It's easier just to track
the stacking context offset in Servo and apply them to the items
manually like we do in non-legacy layout.

---

- [x] There are tests for these changes, but they are not run on CI.

### DIFF
--- a/components/layout/display_list/webrender_helpers.rs
+++ b/components/layout/display_list/webrender_helpers.rs
@@ -28,6 +28,7 @@ use crate::display_list::items::{
 struct ClipScrollState<'a> {
     clip_scroll_nodes: &'a mut Vec<ClipScrollNode>,
     compositor_info: CompositorDisplayListInfo,
+    stacking_context_offset: Vec<LayoutVector2D>,
 }
 
 impl<'a> ClipScrollState<'a> {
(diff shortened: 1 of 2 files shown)


components/layout/display_list/webrender_helpers.rs
```diff
@@ -137,6 +154,8 @@ impl DisplayList {
         let mut builder = DisplayListBuilder::new(webrender_pipeline);
         builder.begin();
 
+        builder.dump_serialized_display_list();
```
mukilan, line 157: Should this be gaurded by a check for the value of `dump_display_list` debug option? Looks like setting this causes WebRender to serialize any items subsequently [added to the display list](https://github.com/servo/webrender/blob/3a951cec4bf463da545b5bbaf0b492b54ecdd7b2/webrender_api/src/display_list.rs#L1203). Is that not a performance concern, even if this is just the non-legacy layout?

mrobinson: I had left this debugging code in accidentally, but I've pushed a new version of the branch that does what you suggest. This is actually a bit useful.

mukilan (approved):

PR #33078, opened 2024-08-16
Remove lazy static

Hmikihiro:
It removes the `lazy_static` crate from dependencies because all of `lazy_static` was replaced.

`Mutex` is need not lazy initialization.
`IS_MULTIPROCESS` would be removed because it is unused constant.

---


jdm (approved):
Woohoo!

PR #33081, opened 2024-08-16
CI: use self-hosted runners for Windows build jobs

delan:
This patch makes our Windows builds use [our self-hosted runners](https://github.com/servo/ci-runners) where possible, falling back to the slower GitHub-hosted runners only when none are available, cutting Windows build jobs from over an hour to [under 18 minutes](https://github.com/servo/servo/actions/runs/10576499947/job/29302440604). Subsequent patches will extend this to self-hosted Linux runners.

This is surprisingly tricky to do in GitHub Actions, because there is no way to make a job prioritise some runners over others, and [there is no way](https://github.com/orgs/community/discussions/50926) to set a timeout for assigning a job to a runner.

The approach we ended up choosing is to prepend a “decision job” that selects the `runs-on` for a specific job (“workload”) that wants self-hosted runners. [Some impls](https://stackoverflow.com/a/78063038) of this approach are [inherently racy](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use), allowing too many jobs to queue up for not enough self-hosted runners, but we can mitigate that problem by using runner labels to “reserve” a self-hosted runner for our workload run. This comes with a few benefits…

- it’s simple to deploy, having no boilerplate to keep in sync with some underlying workflow
- it’s simple to maintain, with fewer moving parts and less shell script sludge to break
- it supports workflows with multiple jobs that want self-hosted runners
- it has the same web UI, with build logs and artifacts in the same place

…and a few caveats:

- the decision job needs its own GitHub API token, so it can write to org (or repo) self-hosted runners
- the runner could still go down between decision and workload, so we need a timeout job to check for that
- all decision jobs need to be serialised with [job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs), which could fail under contention because [there can only be](https://stackoverflow.com/a/77505633) [one pending job](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios) in a concurrency group (in addition to one running job)

**If self-hosted runners break the build after landing, please disable them by [creating a repository variable](https://github.com/servo/servo/settings/variables/actions) named NO_SELF_HOSTED_RUNNERS with any non-empty value, then let me know on Zulip.**

---

- [x] There are tests for these changes, in that the Windows builds rely on them to succeed

## More details for CI enthusiasts

In theory, a race-free approach would be to start a job or workflow for a self-hosted runner, then cancel that and start another one for a GitHub-hosted runner if no self-hosted runner was assigned in an acceptable amount of time. This could be done by [calling](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#calling-a-reusable-workflow) or [dispatching](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event) the underlying Windows workflow.

Calling the workflow would be nice because it inlines the jobs of the underlying workflow into the calling workflow, but there is [no](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28) [way](https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28) to cancel some jobs and not others with the REST API. [Job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) is the only way to cancel a job that has passed its `if` conditions and is now queued, but has not yet been assigned a runner. While cancelling a job with job concurrency lets other jobs continue running normally, it “infects” the status of the workflow, making the whole workflow look `cancelled` (an unsuccessful outcome).

We could dispatch the underlying workflow as a separate workflow run, then use the REST API to cancel it and dispatch a GitHub-hosted run if no self-hosted runners are available, monitor the surviving run until it finishes, and copy its artifacts into the original workflow. This is quite effective, but it adds some friction to the UI by hiding the actual build results behind two extra clicks.


.github/workflows/windows.yml
```diff
@@ -90,13 +117,24 @@ jobs:
...
+          "C:\Program Files (x86)\WiX Toolset v3.11\bin" >> $env:GITHUB_PATH
+
       - name: Build (${{ inputs.profile }})
         run: |
           python mach build --use-crown --locked --${{ inputs.profile }}
           cp C:\a\servo\servo\target\cargo-timings C:\a\servo\servo\target\cargo-timings-windows -Recurse
       - name: Copy resources
+        # FIXME: what does this do?
+        if: ${{ ! inputs.self-hosted-runner }}
         run: cp D:\a\servo\servo\resources C:\a\servo\servo -Recurse
```
sagudev, line 138: On github runners checkouts or done on D: drive, but because it's space is limited, we build the servo on C drive (target folder on C drive), this looks like workaround for `mach` not being able to handle relocating target folder (not sure if it's still needed though).

sagudev:
> [no way](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions) to check the status of another job in an if condition

I think this can be done as a step that obtains that using REST and then sets steps output to response. Then, next step can use this output in if condition.

delan:
> I think this can be done as a step that obtains that using REST and then sets steps output to response. Then, next step can use this output in if condition.

Good point, that’s actually what the `github-hosted` job does, the problem is more that it doesn’t help us cancel a queued job. I’ll edit the text to make that clearer.

sagudev:
Also do note that there are [workflow call nesting limits](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#nesting-reusable-workflows).

delan:
> Also do note that there are [workflow call nesting limits](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#nesting-reusable-workflows).

Yeah, we will touch the limit soon with `try` and `try-label` → `dispatch-workflow` → `windows-auto` → `window`, but for now I don’t think we’ll exceed it.

sagudev:
> [prepending a separate “decision job”](https://stackoverflow.com/a/78063038) to select the runner for a job would be [inherently racy](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

Can't we use [job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) for this?

.github/workflows/windows.yml
```diff
@@ -48,40 +56,59 @@ env:
...
+        if: ${{ ! inputs.self-hosted-runner }}
         run: |
           choco install wixtoolset
           echo "C:\\Program Files (x86)\\WiX Toolset v3.11\\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
       - uses: actions/setup-python@v5
+        if: ${{ ! inputs.self-hosted-runner }}
         with:
           python-version: "3.10"
       - name: Bootstrap
+        if: ${{ ! inputs.self-hosted-runner }}
```
sagudev, line 111: Question: How do we handle rust upgrades (usually needs now crown, and CI image will benefit of preinstalling newer rust)? We might need to add comment in rust-toolchain.toml?

Also how do we handle other changes to bootstrap? Maybe we should still run it, but make it such that it doesn't do extra work when not needed (if crown is already installed it will not be recompiled IIRC).

delan: rust upgrades are handled at runtime by mach, but not crown. baking a new image automatically updates rust, crown, and the other bootstrap tools.

in general, i’m undecided on whether we should run bootstrap at workflow runtime. i think it would be valuable to have the build fail loudly if the github-hosted environment diverges from the self-hosted environment, since it keeps the two environments aligned by requiring us to update the self-hosted environment when we land bootstrap changes.

delan:
> Can't we use [job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) for this?

That’s what we do in this patch :)

.github/workflows/windows.yml
sagudev: > baking a new image automatically updates rust, crown, and the other bootstrap tools

But this needs to be done manually, right? I think it would be nice to automate this using github workflows😁.

I am somehow worried, that diverging will be causing flaky hidden due to auto selection of runners.

.github/workflows/windows.yml
```diff
@@ -90,13 +117,24 @@ jobs:
...
       - name: Add WiX to Path
+        if: ${{ ! inputs.self-hosted-runner }}
         run: |
           "$env:WIX\bin" >> $env:GITHUB_PATH
+
+      # Configure tools in a self-hosted runner.
+      - name: Add python and wixtoolset to path
+        if: ${{ inputs.self-hosted-runner }}
+        run: |
+          "$env:LOCALAPPDATA\Programs\Python\Python312" >> $env:GITHUB_PATH
```
sagudev, line 128: This version is different than on github runners: https://github.com/servo/servo/pull/33081/files#diff-5ce5c9ff86f58b1f87b35b5227b2e84cb69f022d6741e1854f3e1e181091773cR109

sagudev:
I will do more thoroughly review by Monday.

.github/workflows/windows.yml
delan: i agree, i plan to automate that too, when i have more time.

delan:
Also to fix after the weekend, the `build-result` job [fails](https://github.com/delan/servo/actions/runs/10417638807/job/28854756911) when the `self-hosted` job is cancelled:

```yaml
      - name: Success
        if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
        run: exit 0
      - name: Failure
        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
        run: exit 1
```

delan:
It looks like a “cancelled” result in the `self-hosted` job infects the whole workflow run, even if we [fix the check](https://github.com/delan/servo/commit/8c054cac3ddfa22f04a5d1e77dc057c4140f375b#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R55) in the `main` workflow and set the job to [continue on error](https://github.com/delan/servo/commit/8c054cac3ddfa22f04a5d1e77dc057c4140f375b#diff-c0eec5cf90b2569f0f5452158c0223970e9c75ce97f5957bd5c8a30b6f025006R108). This may be a problem.

sagudev:
Maybe instead of checking for any cancelled job we can only check if workflow is canceled: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#cancelled

delan:
> So I think `contains(needs.*.result, 'failure') || cancelled()` should be enough.

Great, [that works too](https://github.com/delan/servo/commit/949758b69b5dc7caa5c9224f9575b5fa5555aec7)! But the `self-hosted` job being cancelled [still makes the whole workflow cancelled](https://github.com/delan/servo/actions/runs/10450947049/job/28936341577), and I can’t find a way around that. Do you know if it will break things like PR checks, or should we go ahead and try this approach?

Otherwise we can dispatch the workflow twice instead of calling it twice, so we can detect the workflow run being cancelled while still having full control over the result of the original workflow.

sagudev:
> Do you know if it will break things like PR checks, or should we go ahead and try this approach?

If 

> > Can't we use [job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) for this?
> 
> That’s what we do in this patch :)

I meant using [job concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) for decision task (thus avoiding race in runner selection) that would query for available self hosted runners, this way we can just set output and use that in if guards. (Although there is still some small race, but I do not think it would be problematic in practice).

delan:
> (Although there is still some small race, but I do not think it would be problematic in practice)

I’m not so sure. The consequence of a race is that the workflow wrongly selects a self-hosted runner, potentially waiting indefinitely if the runner server is down. I think we should avoid that if at all possible.

sagudev:
> > (Although there is still some small race, but I do not think it would be problematic in practice)
> 
> I’m not so sure. The consequence of a race is that the workflow wrongly selects a self-hosted runner, potentially waiting indefinitely if the runner server is down. I think we should avoid that if at all possible.

That would only be a problem if runner goes down when it's already selected by decision task, but not yet dispatched to self-hosted runner (what happens if runner goes down while doing work? Should we do fallback in this case or error?).

sagudev:
I am wrong, It could also happen for first decision task to select self-hosted runner, then before is actually dispatched for second decision task to also get self-hosted runner. We would need atomics for this (maybe a simple serverless atomic endpoint for number of available self hosted runners).

sagudev:
> But the self-hosted job being cancelled [still makes the whole workflow cancelled](https://github.com/delan/servo/actions/runs/10450947049/job/28936341577)

maybe adding result job to windows-auto (that will handle cancellation of self-hostet runner correctly) will cause for main to only pick up result of result job in windows-auto?

sagudev:
Maybe we can assign labels to unlabeled runners (this would be uniq_id) in decision task (using job concurrency to only have one job doing labeling at time), and then run windows runner on label that matches uniq_id. When run is done we remove label from runner (to mark it as free). In case we run out of unlabeled runners we fallback to github provided runner.

Endpoint for labeling: https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#set-custom-labels-for-a-self-hosted-runner-for-a-repository

sagudev:
> Already tried that in [this build](https://github.com/delan/servo/actions/runs/10446746640) ([commit](https://github.com/delan/servo/commit/8c054cac3ddfa22f04a5d1e77dc057c4140f375b)), sadly it didn’t help with the overall result of the workflow.

Ah, you are right. And apparently cancellation from `cancel-in-progress` also causes notification spam: https://github.com/orgs/community/discussions/13015#discussioncomment-4489335

delan:
> Maybe we can assign labels to unlabeled runners (this would be uniq_id) in decision task (using job concurrency to only have one job doing labeling at time), and then run windows workflow on label that matches uniq_id (runs-on uses labels). When run is done we remove label from runner (to mark it as free). In case we run out of unlabeled runners we fallback to github provided runner.

Interesting! I’ve reworked the patch to use workflow dispatch ([8a6176d3c969c](https://github.com/servo/servo/pull/33081/commits/8a6176d3c969c5e1e6065023d28935135affcd5c)), but monitoring the workflow run and copying its artifacts is complex, so I think your approach could be a lot more elegant. Let’s try it :)

delan:
I think the runner labelling approach would definitely solve the problem where two jobs get assigned to self-hosted runners when only one is available. That said, would we still be doing the fallback by setting an output in the decision job, then using that in `runs-on`?

If so, we still have the problem where if the server goes down after the decision is made, the job is stuck in `queued` forever. The job also can’t get assigned to a runner on another server, if we someday have a second server. Someone will have to notice it, cancel the run, and kick off a new run manually.

Maybe this is worth the simpler happy path? I’m not sure. But I think if we can tolerate the workflow dispatch approach, we should go with that to avoid future headaches.

sagudev:
> That said, would we still be doing the fallback by setting an output in the decision job, then using that in runs-on?

Yes

> If so, we still have the problem where if the server goes down after the decision is made, the job is stuck in queued forever

This should happen very rarely (for server to be shutdown at the right moment), we can still use timeout job for that case, but this time the job failure would be expected behavior (in such rare cases user will be notified, so someone will only need to retry it manually).

delan:
> This should happen very rarely (for server to be shutdown at the right moment), we can still use timeout job for that case, but this time the job failure would be expected behavior (in such rare cases user will be notified, so someone will only need to retry it manually).

That could work! Let me play around with that…

delan:
I got close in [this build](https://github.com/delan/servo/actions/runs/10487795220/job/29048921433) ([commit](https://github.com/delan/servo/commit/11547c3b5c0404f4d72d96cdc845cb477bb486c6))! But the magic GITHUB_TOKEN [doesn’t have permission](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication) to read or write runner labels, even if we [set `permissions` to `write-all`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token). The docs say we instead need to use a personal access token or GitHub App installation access token, but in both cases we would need a secret, which may not work in pull requests.

Could you please help me test that by opening a PR to the main branch of [delan/servo](https://github.com/delan/servo)? Then we can kick off a try build, and see if the secret is available.

sagudev:
I think secrets should be available in PRs too. See wpt-export for example: https://github.com/servo/servo/blob/8e224cb4d3730899eba2f01c26cb1bd43caa8812/.github/workflows/pull-request-wpt-export.yml#L37
but I will open PR anyway.

delan:
All done! I’m using a personal access token scoped to the org for now, which will require manual renewal once a year, but I’ll look into GitHub App installation access tokens in a separate patch.

.github/workflows/windows.yml
delan: i’ll bake a new image and fix this :)

.github/workflows/windows.yml
```diff
@@ -46,42 +46,153 @@ env:
...
+          gh api "$self_hosted_runner_scope/$(cat $runner_id)/labels" \
+            -f "labels[]=reserved-for:$unique_id" \
+            -f "labels[]=reserved-since:$reserved_since" \
+            --method POST --silent
+          echo "selected_runner_label=reserved-for:$unique_id" | tee -a $GITHUB_OUTPUT
+          echo 'is_self_hosted=true' | tee -a $GITHUB_OUTPUT
+
+  # In the unlikely event the selected runner goes down after we reserve it but
+  # before the workload starts, cancel the workflow run.
+  runner-timeout:
```
sagudev, line 117: runner-timeout must only run for self-hosted runner, as it's normal to wait a long time for github runner if there is a lot of runs (in my personal profile for example). `if: ${{ fromJSON(needs.runner-select.outputs.is-self-hosted) }}` on job (not step) should do the trick.

delan: done in ff5ad6f51f9f756143713b06121856626e09e6ee ([build](https://github.com/delan/servo/actions/runs/10505379092/job/29102818331), [build](https://github.com/delan/servo/actions/runs/10505381412/job/29102826536))

.github/workflows/windows.yml
delan: done in 9057b2074387c18486aab521924c9a4863218f4f ([ci-runners commit](https://github.com/servo/ci-runners/commit/11f14fd0c654cda5ead5997fdfc4b35ee59bed67))

.github/workflows/windows.yml
delan: one sec, the build failed because it ran on an old runner

.github/workflows/windows.yml
sagudev, line 1: In try-label we use https://github.com/actions/github-script to do such scripting which might be more appropriate (it's JS with native JSON handling and provided github.rest interface), so I think some things can be made more simpler with it, but this can be done as a follow up (or not, nothing wrong with bash but I think github-script would be nicer).

delan: thanks! i’ll be sure to play with this soon, hopefully we can simplify things further :)

.github/workflows/windows.yml
```diff
@@ -46,42 +46,154 @@ env:
...
+    steps:
+      - name: Select and reserve best available runner
+        id: select
+        # Set the variables below to your desired runner images, runner scope
+        # (org or repo), and a token with permission to write to that scope.
+        run: |
+          github_hosted_runner_label=windows-2022
+          self_hosted_image_label=self-hosted-image:windows10
+          self_hosted_runner_scope=/orgs/${{ github.repository_owner }}/actions/runners
+          export GH_TOKEN=${{ secrets.RUNNER_API_TOKEN }}
```
sagudev, line 77: This causes `GH` to fail if secrets is not provided (in personal forks). In that case we want to run on github runner.

See https://github.com/sagudev/servo/actions/runs/10506866977/job/29107421168

delan: great catch! fixed in a2535502d4f3ef2b11225d785a97fb39fa62f2dd

.github/workflows/windows.yml
```diff
@@ -46,42 +46,163 @@ env:
...
+            echo "Timeout waiting for runner assignment! Cancelling workflow run"
+            gh api "$run_url/cancel" --method POST
+          fi
+
   build:
-    name: Windows Build
-    runs-on: windows-2022
+    needs:
+      - runner-select
+    name: Windows Build [${{ needs.runner-select.outputs.unique-id }}]
```
sagudev, line 152: NIT: This is empty when `fall_back_to_github_hosted`: https://github.com/sagudev/servo/actions/runs/10507052600/job/29108001368

.github/workflows/windows.yml
sagudev: My question is how to land a change that updates rustc (and crown). Let's say we have a upgrade PR. This PR cannot land because it needs new crown (rustc upgrade is done by rustup), if we build a new image then and deploy it, no PR can land in the mean time (it will cause disturbance and requires a lot of manual intervention) and even after it lands, all other PRs needs to be rebased for try runs (I think).

TL;DR I think we still need to run `cargo install --path support/crown` (it should be quick if crown is already installed and nothing changed).

.github/workflows/windows.yml
sagudev, line 1: FUTURE-IDEA: given how many `if (!) self-hosted` it might make sense to split this into two files, but then we are repeating. [composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) exists and we should use them.

delan: interesting, thanks!

.github/workflows/windows.yml
delan: fixed in 7d0f50dd25c56b820efd6947e643a7570ef10235

sagudev:
I am not able to test this with pushing to try on servo/servo: https://github.com/servo/servo/actions/runs/10507371415/job/29108982963. Are runners and secrets set?

delan:
> I am not able to test this with pushing to try on servo/servo: https://github.com/servo/servo/actions/runs/10507371415/job/29108982963. Are runners and secrets set?

Sorry, I was in the middle of rebaking the Windows image, while also reworking that process to be easier to do while the old runners are in use. I’ll ping you when the runners are up again.

delan:
@sagudev, runners should be up now!

sagudev:
Still nothing: https://github.com/servo/servo/actions/runs/10507780176/job/29110290788, maybe because I see nothing here: https://github.com/servo/servo/actions/runners?tab=self-hosted

delan:
Hmm, they are org-scoped runners: https://github.com/organizations/servo/settings/actions/runners

sagudev:
> Hmm, they are org-scoped runners: https://github.com/organizations/servo/settings/actions/runners

I do not have permission to access that page, maybe that's the problem why I cannot trigger run on runners. Maybe you should try triggering run from your checkout `mach try win -r upstream` (where `upstream` is remote that points to servo/servo)

.github/workflows/windows.yml
```diff
@@ -46,42 +46,163 @@ env:
...
+        run: |
+          run_url=/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          export GH_TOKEN=${{ secrets.GITHUB_TOKEN }}
+
+          if [ "$(gh api "$run_url/jobs" \
+            | jq -er --arg id '${{ needs.runner-select.outputs.unique-id }}' \
+              '.jobs[] | select(.name | contains("[" + $id + "]")) | .status'
+          )" = queued ]; then
+            echo "Timeout waiting for runner assignment! Cancelling workflow run"
+            gh api "$run_url/cancel" --method POST
```
sagudev, line 146: we should fail job here to make it more obvious what is wrong with simple `exit 1` and maybe clear reservation labels. Actually are reservation labels actually cleared anywhere? If not we should add job that runs always when self hosted is selected (no matter the result of timeout or build job) that will clear up labels.

delan:
> runner did not picked it up: https://github.com/servo/servo/actions/runs/10507901060/job/29110672035

Yeah, the runners are getting labelled…

image

…but not picked up:

```
[2024-08-22T12:25:54Z INFO  monitor::runner] [249] status Reserved, age Ok(1135.717743999s), reserved for Ok(Some(31.238485391s))
[2024-08-22T12:25:54Z INFO  monitor::runner] [250] status Idle, age Ok(447.048950469s), reserved for Ok(None)
[2024-08-22T12:25:54Z INFO  monitor::runner] Unregistering runner 249 with GitHub API runner id 153
[2024-08-22T12:25:54Z INFO  monitor::profile] Destroying runner 249 with base vm name servo-windows10
```

I’ll look into it.

sagudev:
it might be due to special characters (`:`, `/`) in label names or because of long label names, maybe try simple dummy name just for testing.

delan:
We needed to [enable](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/managing-access-to-self-hosted-runners-using-groups#changing-which-repositories-can-access-a-runner-group) servo/servo (and public repos in general) to use the default runner group on [this page](https://github.com/organizations/servo/settings/actions/runner-groups/1).

I’ve done that now, but there are new failures, which I’ll have to investigate tomorrow: https://github.com/servo/servo/actions/runs/10507901060/job/29112065339

.github/workflows/windows.yml
delan: done in 4b3e3d782b5e0fcf0f36c4d2e477bd0e144c7749, though `reserved-for:` labels are already cleared by the monitor ([367f17174c872](https://github.com/servo/ci-runners/commit/367f17174c8724f6f866315cf1ca38341b206b2e)). i think that’s good enough, because we don’t really want to touch runner labels outside of job concurrency, and the runner might fail to assign or run the job for a more interesting reason that has side effects on the runner.

delan:
> I think this failure is caused by images having too new repo clone (`mach try` force pushes HEAD to try branch, and HEAD on this PR is 71 commits behind), so we need more robust repo checkout reuse.

The way we were injecting the Servo repo into the Windows image meant that the working tree was always dirty, which causes errors when switching to FETCH_HEAD. I’ve rebaked the images with fixes for that problem (see below), and added a git reset to the workflow just in case (fafd5ea8f8be50059ceb0e46e394d3d51a4aa6dd).

<details><summary>Dirty modes in working tree (fixed in <a href="https://github.com/servo/ci-runners/commit/663a24b0309e3996a8ff3aedd4350275c160adca">663a24b0309e3</a>)</summary>

image
</details>
<details><summary>Dirty symlinks in working tree (fixed in <a href="https://github.com/servo/ci-runners/commit/03460d36779ee4f3f24a7d570202b9d67632bbac">03460d36779ee</a>)</summary>

image
</details>
</details>
<details><summary>Working trees no longer dirty</summary>

image
</details>

`mach try win` run: <https://github.com/servo/servo/actions/runs/10520815018/job/29150413841>

.github/workflows/windows.yml
sagudev: I think this still needs to be addressed.

.github/workflows/windows.yml
sagudev: Oh, I didn't know we also run monitor for runners, interesting. Are labels also cleared if job is done, but timeout is not hit yet? What if runner is running longer than `SERVO_CI_MONITOR_RESERVE_TIMEOUT`, will this unreserved runner even if it's still running (being reserved)?

delan: the only time we clear labels is when a runner gets unregistered, either automatically due to its job finishing, or if the monitor notices the runner was [reserved but unassigned](https://github.com/servo/ci-runners/blob/cdc1481f9b4f8f7305c3aba976f1c51bc4c54942/monitor/src/runner.rs#L184-L192) for SERVO_CI_MONITOR_RESERVE_TIMEOUT. if the runner gets assigned, that timeout no longer applies.

sagudev: ok

delan:
I’ve added a way to quickly disable self-hosted runners if something goes wrong ([normal build](https://github.com/servo/servo/actions/runs/10558231773/job/29247296192#step:2:73), [disabled build](https://github.com/servo/servo/actions/runs/10558251907/job/29247349829#step:2:75)). This should hopefully be good to go now!

sagudev:
There is still unresolved question at https://github.com/servo/servo/pull/33081#discussion_r1726884036

delan:
> There is still unresolved question at [#33081 (comment)](https://github.com/servo/servo/pull/33081#discussion_r1726884036)

Thanks! I didn’t see that :)

.github/workflows/windows.yml
delan: always installing crown seems like a good middle ground, since it mitigates the big rustc upgrade problem while having ~zero impact on build time and complexity. it might break incremental builds, but incremental builds are currently broken either way (looking into that for a later patch). done in 4887363267395196a3d8b1968eab25baafb2d523 ([build](https://github.com/servo/servo/actions/runs/10558475197)).

.github/workflows/windows.yml
```diff
@@ -223,6 +220,12 @@ jobs:
         run: |
           "$env:WIX\bin" >> $env:GITHUB_PATH
 
+      # Always install crown, even on self-hosted runners, because it is tightly
+      # coupled to the rustc version, and we may have the wrong version if the
+      # commit we are building uses a different rustc version.
+      - name: Install crown
+        run: cargo install --path support/crown --force
```
sagudev, line 227: Why `--force`, when we do not need to replace binary if it's not updated?

delan: `--path` always behaves as if `--force` is passed, i was just making it explicit :)

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
...
+
+      - name: Check configuration for incremental builds
+        if: ${{ fromJSON(needs.runner-select.outputs.is-self-hosted) }}
+        run: |
+          cat C:\init\incremental_build_debug.txt
+          echo "`$env:LIBCLANG_PATH now = $env:LIBCLANG_PATH"
+          echo "`$env:PATH now = $env:PATH"
+
```
sagudev, line 235: This is just logging, for debugging why no incremental builds, right? In that case we might want to rename step to make this more obvious.

delan: done in 8935ae67605bb011bf434d284e1310cffdf267d7

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
   # clang_sys will search msys path before Program Files\LLVM
   # so we need to override this behaviour until we update clang-sys
   # https://github.com/KyleMayes/clang-sys/issues/150
-  LIBCLANG_PATH: C:\\Program Files\\LLVM\\bin
+  LIBCLANG_PATH: C:\Program Files\LLVM\bin
```
sagudev, line 45: I'm surprised this works, but it works!

sagudev (approved):
LGTM, but let's get another pair of eyes just to be sure.

.github/workflows/windows.yml
delan: yeah, me too, it seems `\` only functions as an escape in double-quoted strings ([spec ref yaml 1.2](https://yaml.org/spec/1.2.2/#57-escaped-characters), [yaml 1.0](https://yaml.org/spec/1.0/index.html#syntax-scalar-double))

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
...
+            echo "selected_runner_label=$github_hosted_runner_label" | tee -a $GITHUB_OUTPUT
+            echo 'is_self_hosted=false' | tee -a $GITHUB_OUTPUT
+            exit 0
+          }
+
+          # Generate a unique id that allows the workload job to find the runner
+          # we are reserving for it (via runner labels), and allows the timeout
+          # job to find the workload job run (via the job’s friendly name), even
+          # if there are multiple instances in the workflow call tree.
+          random32=$(LC_ALL=C < /dev/urandom tr -dC 0-9a-f 2> /dev/null | head -c 8)
```
mrobinson, line 90: Could this just use `uuidgen` to generate a true UUID?

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
...
+          echo "selected_runner_label=reserved-for:$unique_id" | tee -a $GITHUB_OUTPUT
+          echo 'is_self_hosted=true' | tee -a $GITHUB_OUTPUT
+
+  # In the unlikely event a self-hosted runner was selected and reserved but it
+  # goes down before the workload starts, cancel the workflow run.
+  runner-timeout:
+    needs:
+      - runner-select
+    if: ${{ fromJSON(needs.runner-select.outputs.is-self-hosted) }}
+    name: Runner timeout
```
mrobinson, line 141:
```suggestion
    name: Detect Runner Timeout
```

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
...
+        uses: actions/checkout@v4
+      - if: ${{ ! fromJSON(needs.runner-select.outputs.is-self-hosted) && github.event_name == 'pull_request_target' }}
+        uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
+
+      # Faster checkout for self-hosted runner that uses prebaked repo.
+      - if: ${{ fromJSON(needs.runner-select.outputs.is-self-hosted) && github.event_name != 'pull_request_target' }}
+        run: git fetch --depth=2 origin $env:GITHUB_SHA
```
mrobinson, line 181: I'm not sure that `--depth 2` is necessary any longer now that linting has its own job. Maybe @sagudev can confirm.

mrobinson (approved):
Looks good to me. This is very exciting. I have a few naming nits  (I know...typical) and a couple small suggestions, but otherwise this looks good.

.github/workflows/windows.yml
sagudev: Indeed, `--depth 1` should be enough.

.github/workflows/windows.yml
```diff
@@ -42,52 +42,204 @@ env:
...
+  # Automatic runner selection for job: build
+  # Runs the underlying job (“workload”) on a self-hosted runner if available,
+  # with the help of a `runner-select` job and a `runner-timeout` job.
+
+  # Selects a self-hosted runner if available, or else a GitHub-hosted runner.
+  # We generate a unique id for the workload, find an idle self-hosted runner
+  # with the given `image:` label that wasn’t already reserved by another
+  # `runner-select` job run, and reserve it with a `reserved-for:<id>` label.
+  runner-select:
+    name: Runner select
```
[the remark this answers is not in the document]
delan: no worries at all! sorry for not noticing the convention :)

.github/workflows/windows.yml
delan: great idea, i didn’t know about that command! done in 927aad49744c4984ef9d46839f252049c8eaacdf

.github/workflows/windows.yml
delan: done in bcc52e09aefc7b32424833b414e338cc5f772586

.github/workflows/windows.yml
delan: great idea, i didn’t know about that command! and since we no longer dispatch a separate workflow run, humans no longer benefit from the repo and run id. done in bcc52e09aefc7b32424833b414e338cc5f772586

.github/workflows/windows.yml
delan: done in 4f21cac163c17370bb14a5882150dadd2e678e08

delan: done in 4f21cac163c17370bb14a5882150dadd2e678e08

delan:
Try build: <https://github.com/servo/servo/actions/runs/10576499947/job/29302440604>

PR #33087, opened 2024-08-16
Test out standard-conforming about:blank iframe load changes

jdm:
This is a quick and dirty attempt to see what changes if we follow the spirit of the standard for about:blank iframes, per #31973.


jdm:
We're still missing a sync load event on the iframe element, per this testcase:
```
<script>var a = "sync";</script>
<iframe onload="console.log(a + ' load event')"></iframe>
<script>a = "async"</script>
```
which prints "async load event".

jdm:
The latest changes have broken tests that use `window.open("")` and set an onload listener on the resulting window object.

jdm:
I _think_ the problem is that we set event handlers on the window object that corresponds to the initial about:blank, and this is then overwritten by the window object for the page that is navigated to by the actual URL that was opened.

jdm:
I'm experimenting with various schemes for sharing the window object when replacing an about:blank. No success to report yet; the most correct form requires making a lot more fields mutable, which is a pain.

jdm:
I've got an implementation locally that reuses the window and also fixes bugs like about:blank documents triggering two load events. I'm now trying to figure out why the document that replaced about:blank no longer receives a load event.

jdm:
One source of multiple test failiures is that we don't have any implementation for interrupting in progress navigations.

jdm:
No point in leaving this PR open.

PR #33095, opened 2024-08-16
remove usage of legacy numeric operations in script/

simonwuelker:
These operations are deprecated and might be removed in a future rust version. Clippy's `legacy_numeric_constants` lint is also complaining about them.

For example, [`u32::max_value()`](https://doc.rust-lang.org/std/primitive.u32.html#method.max_value) should be [`u32::MAX`](https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX)

---

- [X] These changes do not require tests because they don't change behaviour


jdm (approved):
Thanks!

PR #33096, opened 2024-08-16
Make string formatting more consistent in `CodegenRust.py`

### LINKED ISSUE #31846: Consistent string formatting in `CodegenRust.py`
**Describe the new feature:**
`CodegenRust.py` currently uses all types of python [string formatting](https://docs.python.org/3/tutorial/inputoutput.html) (`+`, `%`, `.format`, `f""`...) as a result of many years of development. It would be nice to make the code consistent and have a guideline for which method to use in each case. Per the discussion [here](https://github.com/servo/servo/pull/31844#issuecomment-2016769923), both @sagudev and I think that [f strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are a good candidate since they are similar to rust and python is moving in that direction. What does everyone else think?

saito828koki:
This PR is to integrate various string formatting styles in `CodegenRust.py` into f-string.

The key changes in this PR are
- Replace string concatenations using `+` operator with f-string
- Replace `string.Template.substitute` with f-string
- Replace %formatting with f-string

This PR does **NOT** include the following changes
- Remove template strings that are used more than once
- Refactor formatting-related util functions (e.g. `fill` function, `instantiateJSToNativeConversionTemplate` function)

notes:
- I deleted an unused function `onFailureInvalidEnumValue`
- Inside f-string, we need to double curly braces `{{` `}}` to escape them (see https://docs.python.org/3/reference/lexical_analysis.html#f-strings )
- We cannot use backslash`\` inside curly braces. For example, `f"{'\n'.join(arr)}"` is invalid. (On my local Mac, `./mach test-tidy` doesn't report it. Linux build on CI fails. I feel it would be better to have a way to check tidiness as strictly as CI locally.)
- IMO, using ternary operators inside f-string sometimes makes it hard to read, so I avoid that kind of usage

---

- [x] These changes fix #31846 (GitHub issue number if applicable)

- [x] These changes do not require tests because this is a code style fix

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -661,12 +661,6 @@ def getJSToNativeConversionInfo(type, descriptorProvider, failureCode=None,
                     '%s' % (firstCap(sourceDescription), exceptionCode))),
             post="\n")
 
-    def onFailureInvalidEnumValue(failureCode, passedVarName):
-        return CGGeneric(
-            failureCode
-            or ('throw_type_error(*cx, &format!("\'{}\' is not a valid enum value for enumeration \'%s\'.", %s)); %s'
-                % (type.name, passedVarName, exceptionCode)))
-
     def onFailureNotCallable(failureCode):
         return CGGeneric(
             failureCode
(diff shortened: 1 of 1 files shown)


saito828koki:
Rebased onto the latest `main` to resolve conflicts.

jdm (approved):
Oof. It must have been a slog to update all those different usages!

PR #33098, opened 2024-08-17
Fix a floating-point error in table layout

### LINKED ISSUE #32650: Servo crashes while rendering https://html.spec.whatwg.org/
**Describe the bug:**
Servo crashes while rendering https://html.spec.whatwg.org/

```                                                                                                                                                          
  left: 776.0166666666667px
 right: 776px (thread Script(1,1), at components/layout_2020/flow/float.rs:279)
   0: <servoshell::backtrace::Print as core::fmt::Debug>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: servoshell::backtrace::print
   4: servoshell::panic_hook::panic_hook
   5: std::panicking::rust_panic_with_hook
   6: std::panicking::begin_panic_handler::{{closure}}
   7: std::sys_common::backtrace::__rust_end_short_backtrace
   8: _rust_begin_unwind
   9: core::panicking::panic_fmt
  10: core::panicking::assert_failed_inner
  11: core::panicking::assert_failed
  12: layout_2020::flow::float::PlacementAmongFloats::try_to_expand_for_auto_block_size
  13: layout_2020::flow::BlockLevelBox::layout::{{closure}}
  14: layout_2020::flow::BlockLevelBox::layout
  15: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
  16: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
  17: layout_2020::flow::BlockContainer::layout
  18: layout_2020::flow::BlockLevelBox::layout::{{closure}}
  19: layout_2020::flow::BlockLevelBox::layout
  20: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
  21: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
  22: layout_2020::flow::BlockContainer::layout
  23: layout_2020::flow::BlockFormattingContext::layout
  24: layout_2020::flow::BlockLevelBox::layout::{{closure}}
  25: layout_2020::flow::BlockLevelBox::layout
  26: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &F>::call_mut
  27: rayon::iter::plumbing::Folder::consume_iter
  28: rayon::iter::plumbing::bridge_producer_consumer::helper
  29: rayon::iter::collect::collect_with_consumer
  30: layout_2020::flow::BlockContainer::layout
  31: layout_2020::flow::BlockFormattingContext::layout
  32: layout_2020::flow::root::BoxTree::layout
  33: <layout_thread_2020::LayoutThread as script_layout_interface::Layout>::reflow
  34: script::dom::window::Window::force_reflow
  35: script::dom::window::Window::reflow
  36: script::dom::document::Document::finish_load
  37: script::dom::servoparser::ServoParser::do_parse_sync
  38: profile_traits::time::profile
  39: script::dom::servoparser::ServoParser::parse_sync
  40: script::dom::document::Document::process_pending_parsing_blocking_script
  41: script::dom::document::Document::finish_load
  42: <script::stylesheet_loader::StylesheetContext as net_traits::FetchResponseListener>::process_response_eof
  43: <script::network_listener::ListenerTask<A,Listener> as script::task::TaskOnce>::run_once
  44: <T as script::task::TaskBox>::run_box
  45: script::script_thread::ScriptThread::handle_msg_from_script
  46: script::script_thread::ScriptThread::handle_msgs
  47: script::script_thread::ScriptThread::start
  48: profile_traits::mem::ProfilerChan::run_with_memory_reporting
  49: std::sys_common::backtrace::__rust_begin_short_backtrace
  50: core::ops::function::FnOnce::call_once{{vtable.shim}}
  51: std::sys::pal::unix::thread::Thread::new::thread_start
  52: __pthread_joiner_wake
```

**To Reproduce:**
`./mach run -r https://html.spec.whatwg.org/`

**Platform:**
OSX

simonwuelker:
~~When placing a table among floating elements we can't initially know its inline size. So we do the following:~~

1) ~~Compute minimum size~~
2) ~~Find position for table given our current size guess~~
3) ~~Layout table at that position~~
4) ~~Check if the position is valid for the new size, if not then repeat steps 2-4~~

~~This happens in [`layout_2020/flow/mod.rs`](https://github.com/servo/servo/blob/main/components/layout_2020/flow/mod.rs) lines 1122-1200.~~

~~At [`layout_2020/flow/mod.rs`](https://github.com/servo/servo/blob/main/components/layout_2020/flow/mod.rs#L1166) line 1166 the loop updates the size of the placement - which is valid because it immediately validates that the size of the `placement_rect` is still larger afterwards. This is an optimization that saves us from having to take another iteration of the loop ("if table shrunk then the current position is not invalidated").~~

https://github.com/servo/servo/blob/20273b062af969152635306c3df2a3a1364ac4d1/components/layout_2020/flow/mod.rs#L1166-L1170

~~An assertion in `PlacementAmongFloats::try_to_expand_for_auto_block_size` assumed that the inline size of the placement would be the the same as the assumed size of the table. This is not true, because of the optimization explained above. It is only guaranteed that the inline size of the placed element will not be greater than the inline size of the placement rect. This commit updates the assertion accordingly~~

A floating point error in table layout caused the table to report different widths, causing a crash during float layout.

With this change, https://html.spec.whatwg.org/ loads without crashing.

**Note:** I've not worked with the float layout code before, so its entirely possible that the explanation above is just wrong. Please review carefully (: ~~

---

- [X] These changes fix #32650
- [ ] Tests would be nice but i have no idea how one would test for this :/

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -890,17 +890,45 @@ impl<'a> TableLayout<'a> {
         }
 
         let bounds = |sum_a, sum_b| self.assignable_width > sum_a && self.assignable_width < sum_b;
+
         let blend = |a: &[Au], sum_a: Au, b: &[Au], sum_b: Au| {
             // First convert the Au units to f32 in order to do floating point division.
             let weight_a =
                 (self.assignable_width - sum_b).to_f32_px() / (sum_a - sum_b).to_f32_px();
             let weight_b = 1.0 - weight_a;
-            a.iter()
+
+            let mut sum_accounting_for_floating_point_inaccuracy = Au::new(0);
+            let mut widths: Vec<Au> = a
+                .iter()
                 .zip(b.iter())
                 .map(|(guess_a, guess_b)| {
                     (guess_a.scale_by(weight_a)) + (guess_b.scale_by(weight_b))
                 })
-                .collect()
+                .inspect(|&column_width| {
+                    sum_accounting_for_floating_point_inaccuracy += column_width;
+                })
+                .collect();
+
+            if sum_accounting_for_floating_point_inaccuracy != self.assignable_width {
+                // The computations above can introduce floating-point imprecisions.
+                // Since these errors are very small (+-1Au), it's fine to simply adjust
+                // the first column such that the total width matches the assignable width
+                let difference =
+                    self.assignable_width - sum_accounting_for_floating_point_inaccuracy;
+
+                debug_assert!(
+                    difference.abs() <= Au::new(widths.len() as i32),
+                    "A deviation of more than one Au per column is unlikely to be caused by float imprecision"
+                );
+
+                // We checked if the table was empty at the top of the function, so there
+                // always is a first column
+                widths[0] += difference;
+            }
+
+            debug_assert!(widths.iter().sum::<Au>() == self.assignable_width);
+
+            widths
         };
 
         if bounds(min_content_sizes_sum, min_content_percentage_sizing_sum) {


mrobinson:
Nice work and great investigation! This looks good to me with the caveat below. I think it makes sense that @Loirooriol also looks at the change as well as he is the originator this code AFAIR.

The one concern I have is that if an table has an auto width and we try to lay it out into a certain width, I think it genuinely should consume it. From the assertion failure it almost looks like we are dealing with some floating point rounding issues though. The difference between the two values is 0,016666666666667 which happens to be exactly one `Au` (our app unit, ie 1/60th of a pixel). That seems very suspicious to me! Maybe a more conservative fix here would be to assert the equality of the values, but within a few `Au` of each other to account for rounding issues.

Wuelle:
>  The difference between the two values is 0,016666666666667 which happens to be exactly one Au. That seems very suspicious to me!

That's very suspicious indeed! I didn't know that 1Au = 1/60px.

Relaxing the assertion to check equality only within a few `Au` seems - unsatisfying. Ideally the table layout would not introduce these (presumably) rounding errors in the first place.
They also cause layout to be rather inefficient by the way. In the reduction I posted in #32650 the table layout is calculated *3 times* because each time the inline width varies by +-`1Au`.



But yeah, we should definitely wait until someone who's familiar with this code can take a look (:
Quite a fun issue!

Loirooriol:
I think this is all a bit strange because the table is before the float, so it shouldn't have much effect over it. I would like to analyze further.

Loirooriol:
Trying the testcase from https://github.com/servo/servo/issues/32650#issuecomment-2296325631, the problem seems unrelated to floats. When the table is laid out within an available space of `575px`, the `content_inline_size_for_table` becomes `575.0166666666667px`, even if the min-content contribution of the table is just `257px`, so no need to overflow.

Loirooriol:
While relaxing the assert might be fine, the problem is that table layout isn't stable: when laying out with an available space of 575px, it becomes `575.0166666666667px`. So we increase the available space to `575.0166666666667px`, but then the table becomes smaller: `575px`. This breaks the assumptions in the float code.

Wuelle:
> the problem is that table layout isn't stable

This instability seems to originate in https://github.com/servo/servo/blob/91adf39de78626df8c04c89261c4a925e8e7f689/components/layout_2020/table/layout.rs#L893-L904
(from [`TableLayout::distribute_width_to_columns`](https://github.com/servo/servo/blob/91adf39de78626df8c04c89261c4a925e8e7f689/components/layout_2020/table/layout.rs#L796))

mrobinson (approved):
This seems like a reasonable fix. I think this amount could grow to be larger if there were > 60 columns, but a table at that size is pretty unlikely and also a few pixels of difference in the first column are not going to make a big difference.

We do something like this in flexbox as well, but what we do is to calculate the remaining space and lines and recalculate the space to distribute each iteration of the loop.

PR #33100, opened 2024-08-17
Basic tab strip for the minibrowser

webbeef:
This implements a simple tab system for servoshell:
- The egui part uses the built-in SelectableLabels components and display the full tab title on hover.
- WebView structs now hold all the state for each WebView. When we need "global" state, we return the focused WebView state, eg. for the load status since it's still global in the UI.
- New keyboard shortcut: [Cmd-or-Ctrl]+[W] to close the current tab.
- New keyboard shortcut: [Cmd-or-Ctrl]+[T] to create a new tab.


msub2:
Neat! I think this would supersede this PR I remember from a little while ago also #31545

msub2:
The one thing I think would be a nice addition to this would be having the new tab URL being an `about:newtab` instead of a data URL

delan:
> Oh I did not see that one. I hope @delan won't mind and that we can push something to the finish line :)

No worries! I haven’t had time to work on that patch, and I probably won’t for a while 😅

ports/servoshell/desktop/webview.rs
```diff
@@ -393,6 +444,19 @@ where
...
             })
+            .shortcut(CMD_OR_CONTROL, 'W', || {
+                if let Some(id) = self.focused_webview_id {
+                    self.event_queue.push(EmbedderEvent::CloseWebView(id));
+                }
+            })
+            .shortcut(CMD_OR_CONTROL, 'T', || {
+                // data url encoding of: <html><head><title>New Tab</title><body>New Servo Tab!
+                let data_url =
+                    "data:text/html;base64,PGh0bWw+PGhlYWQ+PHRpdGxlPk5ldyBUYWI8L3RpdGxlPjxib2R5Pk5ldyBTZXJ2byBUYWIh";
```
atbrakhi, line 455: Just wondering if it's be better to just keep the default `ServoUrl` that is `servo.org` for the new tab as well, at least for now wdyt?

I like what m2sub suggested but as you said this could be another PR.

atbrakhi (commented):
Nice work!

ports/servoshell/desktop/webview.rs
atbrakhi: Ideally it would be really nice to have something like ecosia/duckduckgo as default, but with the current changes, I'm concerned it might make the minibrowser appear less polished or mature. While I recognize that there are still many things we're working on improving, we should try to put the best forward.

Just a thought—I’m happy to hear if others have different opinions

atbrakhi (approved):
Thanks for working on it. Looks great to me.

niutech:
Is it possible for you to use the [egui_dock](https://github.com/Adanos020/egui_dock) for tabs, so that it would also allow tiled and grid view like my [Split Browser](https://github.com/niutech/splitbrowser)? This would be a pretty unique feature among web browsers.

mrobinson:
> Is it possible for you to use the [egui_dock](https://github.com/Adanos020/egui_dock) for tabs, so that it would also allow tiled and grid view like my [Split Browser](https://github.com/niutech/splitbrowser)? This would be a pretty unique feature among web browsers.

I think for now we should keep things pretty basic. The Servo MiniBrowser is mainly used for testing the web engine. That doesn't stop anyone from making experimental browsers with Servo though.

PR #33101, opened 2024-08-17
Replace `callargs_is_constructing` util func with `is_constructing` method

### LINKED ISSUE #29821: Replace callargs_is_constructing with a method on CallArgs
I added it as a function in utils.rs for expediency, but it belongs in https://github.com/servo/mozjs/blob/b9edc816b6662a5986e190cdf53ae295049acb92/mozjs/src/jsimpls.rs#L221 instead.

saito828koki:
This PR is for refactoring `callargs_is_constructing` function.

We introduced a new method `is_constructing` on `JS::CallArgs` in https://github.com/servo/mozjs/pull/497, so now we can clean up `callargs_is_constructing` and use `is_constructing` instead.

This PR includes the following changes:

- Update `Cargo.lock` to point to the latest `mozjs` and `mozjs_sys`
- Replace `callargs_is_constructing` with [`is_constructing`](https://github.com/servo/mozjs/blob/4f0724dd3b9b58120903ff6a1043e71e7c7b9eb1/mozjs-sys/src/jsimpls.rs#L306-L309) method.
- Delete `callargs_is_constructing` since it's no longer used

---

- [x] These changes fix #29821

- [x] These changes do not require tests because this is a simple refactoring

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4319,7 +4319,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/servo/mozjs#dae748dee6fb8c1db771364bd12101db1ed34614"
+source = "git+https://github.com/servo/mozjs#4f0724dd3b9b58120903ff6a1043e71e7c7b9eb1"
 dependencies = [
  "bindgen",
  "cc",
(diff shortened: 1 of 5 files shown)


jdm (approved):
Thanks!

PR #33104, opened 2024-08-18
 make protocol handlers registrable

webbeef:
The idea is to allow embedders to implement their own protocol handlers, so that they can add support for custom / experimental ones.

The change here consist of the creation of simple registry mapping schemes to handlers. Currently the registration is done in `create_constellation()`  and the registry is passed down in the `FetchContext`.

The following protocols have been moved to this setup:
- data:
- file:
- blob:

If that approach is acceptable, I will add the hooks on the embedder api to merge the embedder protocols with the internal ones.


jdm:
This looks reasonable to me!

jdm (approved):
Thanks!

jdm:
Should this still be marked as a draft?

wusyong:
@webbeef  This is also on my todo list. Thanks for adding this feature already!

wusyong:
@webbeef I'm drafting a blog post to summarize what's next for browser and embedding stories in https://github.com/servo/servo.org/pull/171
I guess just keep pushing what you are already working on. Multiprocess mode is also one of the milestone in my mind.

Issue #30514, opened 2023-10-07
Implement support for `clip-path: polygon`


EnnuiL:
I've spotted this issue at first while attempting to correct test expectations on PR #30439 (relevant comment: https://github.com/servo/servo/pull/30439#issuecomment-1744074473);

My first assumption was "oh well, maybe this is what I deserve for running this on the Intel Arc A380 while also on Linux", but well, looking at [wpt.fyi results](https://wpt.fyi/results/css/css-masking?label=master&product=chrome%5Bexperimental%5D&product=edge%5Bexperimental%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&product=servo&aligned) show that it might be a more widespread issue, with this [comparison of `clip-path-rect-001.html`](https://wpt.fyi/analyzer?screenshot=sha1%3Afa4b31cc7352eb0ae3f0d2070499de4e26b1ee4b&screenshot=sha1%3Ab644b0ec11ed044b7d99a40fd02c0810cf860ce3) matching the ones I get:
![A cropped screenshot of a failing `clip-path-rect-001.html` test. The green box hasn't been clipped here](https://github.com/servo/servo/assets/85590273/2db67ba4-3353-4cc5-9713-7895910bf7a1)

Attempting to use `--software` while using `./mach run` didn't affect the outcome, and both `--legacy-layout` and `--layout` are affected

jdm:
According to https://github.com/servo/servo/blob/master/tests/wpt/include.ini#L40C8-L42 we only run the tests under css-masking/clip in our CI right now.

jdm:
However, running tests on directory like css-masking will run all tests underneath it, ignoring the configuration from that linked file.

EnnuiL:
Oh well, I should have paid more attention to what exactly was failing;
It turns out `clip-path` isn't implemented at all; I'll update the issue name in order to reflect that;
(maybe I could implement something...? who knows, I don't know! :p)

BudiArb:
Hi, is anyone currently working on this issue? The PR https://github.com/servo/servo/pull/33680 by @chocolate-pie was last updated almost 2 years ago, so I'm not sure if there is any plan to continue it. If no one is currently working on it, I want to assign myself

jdm:
Nobody is work on it, so you are welcome!

jdm:
https://github.com/servo/servo/pull/33680#issuecomment-2488923444 is the best path forward here, relying on vello_cpu.

PR #33107, opened 2024-08-18
layout: Add initial support for `clip-path: [<basic-shape> || <shape-box>]`

chocolate-pie:
Add initial support for `clip-path: [<basic-shape> || <shape-box>]`. This PR doesn't implement support for `polygon()` property because blob rasterizer hasn't been implemented. (and pretty difficult)

---

- [x] These changes fix a part of #30514


components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+                    parent_clip_chain_id,
+                    display_list,
+                )
+            },
+            BasicShape::Polygon(_) | BasicShape::PathOrShape(_) => None,
+        }
+    } else {
+        create_rect_clip_chain(
+            match geometry_box {
+                ShapeBox::MarginBox => compute_marginbox_radius(
```
mrobinson, line 55:
```suggestion
                ShapeBox::MarginBox => compute_margin_box_radius(
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+            parent_scroll_node_id,
+            parent_clip_chain_id,
+            display_list,
+        )
+    }
+}
+
+fn build_simple_shape(
+    shape: BasicShape,
+    layout: LayoutRect,
```
mrobinson, line 72: A layout is more than just a rectangle, so I think it's good to specify to help the reader:

```suggestion
    layout_rect: LayoutRect,
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+    match shape {
+        BasicShape::Rect(rect) => {
+            let top = rect.rect.0.resolve(Length::new(layout.height()));
+            let right = rect.rect.1.resolve(Length::new(layout.width()));
+            let bottom = rect.rect.2.resolve(Length::new(layout.height()));
+            let left = rect.rect.3.resolve(Length::new(layout.width()));
+            let x = layout.min.x + left.px();
+            let y = layout.min.y + top.px();
+            let width = layout.width() - (left + right).px();
+            let height = layout.height() - (top + bottom).px();
```
mrobinson, line 86: I think this could be simplified slightly:

```suggestion
            let insets = LayoutSideOffsets::new(
                rect.rect.0.resolve(Length::new(layout.height())).px(),
                rect.rect.1.resolve(Length::new(layout.width())).px(),
                rect.rect.2.resolve(Length::new(layout.height())).px(),
                rect.rect.3.resolve(Length::new(layout.width())).px()
            );
            let shape_rect = layout.to_rect().inner_rect(insets).to_box2d();
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+            };
+            let mut radii = webrender_api::BorderRadius {
+                top_left: corner(&rect.round.top_left),
+                top_right: corner(&rect.round.top_right),
+                bottom_left: corner(&rect.round.bottom_left),
+                bottom_right: corner(&rect.round.bottom_right),
+            };
+            let origin = LayoutPoint::new(x, y);
+            let size = LayoutSize::new(width, height);
+            let rect = LayoutRect::from_origin_and_size(origin, size);
```
mrobinson, line 104: This should now be unnecessary as `shape_rect` is defined above.

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+        },
+        BasicShape::Circle(circle) => {
+            let center = match circle.position {
+                GenericPositionOrAuto::Position(position) => position,
+                GenericPositionOrAuto::Auto => Position::center(),
+            };
+            let anchor_x = center.horizontal.resolve(Length::new(layout.width()));
+            let anchor_y = center.vertical.resolve(Length::new(layout.height()));
+            let amount = LayoutSize::new(anchor_x.px(), anchor_y.px());
+            let center = layout.min.add_size(&amount);
```
mrobinson, line 122: I would combine this to avoid the ambiguously-named `amount`.
```suggestion
            let center = layout.min.add_size(&LayoutSize::new(anchor_x.px(), anchor_y.px()));
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+        },
+        _ => None,
+    }
+}
+
+fn compute_shape_radius(
+    center: f32,
+    radius: &GenericShapeRadius<NonNegativeLengthPercentage>,
+    layout_min: f32,
+    layout_max: f32,
```
mrobinson, line 194: Nit: Perhaps "edge" here makes it a bit more obvious what these values represent. "layout" is a bit superfluous.

```suggestion
    min_edge: f32,
    max_edge: f32,
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+) -> f32 {
+    let left = (layout_min - center).abs();
+    let right = (layout_max - center).abs();
```
mrobinson, line 197: It took me a bit to understand what was going on there and these names would be more helpful, but let me know if I've misunderstood this code.

```suggestion
    let distance_from_min_edge_to_center = (layout_min - center).abs();
    let distance_from_max_edge_to_center = (layout_max - center).abs();
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+    match (radius, left.partial_cmp(&right)) {
+        (GenericShapeRadius::FarthestSide, Some(Ordering::Greater)) => left,
+        (GenericShapeRadius::FarthestSide, _) => right,
+        (_, Some(Ordering::Greater)) => right,
+        (_, _) => left,
+    }
```
mrobinson, line 203: I think this could be simplified:
```suggestion
    match radius {
        GenericShapeRadius::FarthestSide => left.max(right),
        GenericShapeRadius::ClosestSide => left.min(right),
    }
```

components/layout_2020/display_list/clip_path.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+fn create_rect_clip_chain(
+    radii: webrender_api::BorderRadius,
+    rect: LayoutRect,
+    parent_scroll_node_id: ScrollTreeNodeId,
+    parent_clip_chain_id: ClipChainId,
+    display_list: &mut super::DisplayList,
+) -> Option<ClipChainId> {
```
mrobinson, line 212: There's no way for this function to return `None`, so create the option in the caller.

```suggestion
) -> ClipChainId {
```

components/layout_2020/display_list/mod.rs
```diff
@@ -1300,3 +1297,84 @@ fn resolve_border_image_slice(
         resolve_percentage(border_image_slice.3, size.width),
     )
 }
+
+pub(crate) fn normalize_radii(rect: &units::LayoutRect, radius: &mut wr::BorderRadius) {
```
mrobinson, line 1301: Is it possible to limit the scope of the visibility here? I think this is only useful for display list building.

```suggestion
pub(super) fn normalize_radii(rect: &units::LayoutRect, radius: &mut wr::BorderRadius) {
```

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -488,7 +489,8 @@ impl StackingContext {
         if effects.filter.0.is_empty() &&
             effects.opacity == 1.0 &&
             effects.mix_blend_mode == ComputedMixBlendMode::Normal &&
-            !style.has_transform_or_perspective(FragmentFlags::empty())
+            !style.has_transform_or_perspective(FragmentFlags::empty()) &&
+            style.clone_clip_path() == ClipPath::None
```
mrobinson, line 493: Nice!

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -1059,9 +1061,15 @@ impl BoxFragment {
             );
         }
 
+        let clip_path_chain_id = self.build_clip_path_frame_if_necessary(
+            display_list,
+            containing_block.scroll_node_id,
+            containing_block.clip_chain_id,
+            &containing_block.rect,
+        );
```
mrobinson, line 1069: Why does this happen here rather than near the call to `build_clip_path_frame_if_necessary`. I think that the usage of the terminology `frame` here is left over from when clip where actually "frames" in WebRender. I would omit `frame` in the name of this method.

components/layout_2020/display_list/stacking_context.rs
```diff
@@ -1410,6 +1418,22 @@ impl BoxFragment {
         Some(sticky_node_id)
     }
 
+    fn build_clip_path_frame_if_necessary(
```
mrobinson, line 1421:
```suggestion
    fn build_clip_path_clip_chain_if_necessary(
```

mrobinson (changes requested):
Nice work! Sorry for the long review time. I just have a couple minor comments and questions, but I think this can land very soon.

mrobinson:
I've made many of the changes mentioned above as well as some other minor cleanups. In addition, I've turned on the tests for `clip-path`, since they were disabled. Now the bots should be able to test our changes there and make sure that things don't start failing.

mrobinson (approved):

PR #33112, opened 2024-08-18
Fix race in WebXR WPT test setup

msub2:
The WebXR WPT tests use a `webxr_util.js` file that handles running the tests, part of which involves appending a canvas element to the document body. Many of the webxr test pages, however, are written without explicit body elements, which in turn can cause tests to error out early from document.body being null. Wrapping the calls to `xr_promise_test` inside a listener for DOMContentLoaded ensures that these tests will now run properly, regardless of explicit body element being present.

With this, Servo's WebXR WPT results should now fully reflect the actual state of its support for the various WebXR APIs, including the WebXR Test API.

---


jdm (approved):
Thanks for doing this investigation!

PR #33114, opened 2024-08-19
mach: introduce `BuildTarget` abstraction

mukilan:
Introduce a new `BuildTarget` abstraction to centralize the code for
supporting different ways of choosing the build target (e.g --android,
--target x86_64-linux-android , --target aarch64-linux-ohos). This
is currently handled in an adhoc fashion in different commands (
mach package, install, run) leading to a proliferation of keyword
parameters for the commands and duplicated logic.

The patch introduces a new `allow_target_configuration` decorator to
do the validation and parsing of these parameters into the appropriate
`BuildTarget` subclass, which is now stored as an instance attribute
of the CommandBase class. All the code that previously relied on
`self.cross_compile_target` has been switched to use the BuildTarget.

---
- [x] These changes do not require tests because they only affect the build tooling code.

### DIFF
--- a/python/servo/gstreamer.py
+++ b/python/servo/gstreamer.py
@@ -13,6 +13,11 @@ import subprocess
 import sys
 from typing import Set
 
+# This file is called as a script from components/servo/build.rs, so
+# we need to explicitly modify the search path here.
+sys.path[0:0] = [os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))]
+from servo.platform.build_target import BuildTarget  # noqa: E402
+
 GSTREAMER_BASE_LIBS = [
     # gstreamer
     "gstbase",
@@ -242,7 +247,7 @@ def find_non_system_dependencies_with_otool(binary_path: str) -> Set[str]:
     return output
 
 
-def package_gstreamer_dylibs(binary_path: str, library_target_directory: str, cross_compilation_target: str = None):
+def package_gstreamer_dylibs(binary_path: str, library_target_directory: str, target: BuildTarget):
     """Copy all GStreamer dependencies to the "lib" subdirectory of a built version of
        Servo. Also update any transitive shared library paths so that they are relative to
        this subdirectory."""
@@ -250,7 +255,7 @@ def package_gstreamer_dylibs(binary_path: str, library_target_directory: str, cr
     # This import only works when called from `mach`.
     import servo.platform
 
-    gstreamer_root = servo.platform.get().gstreamer_root(cross_compilation_target)
+    gstreamer_root = servo.platform.get().gstreamer_root(target)
     gstreamer_version = servo.platform.macos.GSTREAMER_PLUGIN_VERSION
     gstreamer_root_libs = os.path.join(gstreamer_root, "lib")
 
--- a/python/servo/package_commands.py
+++ b/python/servo/package_commands.py
@@ -132,31 +132,21 @@ class PackageCommands(CommandBase):
                      default=None,
                      help='Package using the given Gradle flavor')
     @CommandBase.common_command_arguments(build_configuration=False, build_type=True)
-    def package(self, build_type: BuildType, android=None, target=None, flavor=None, with_asan=False):
-        if android is None:
-            android = self.config["build"]["android"]
-        if target and android:
-            print("Please specify either --target or --android.")
-            sys.exit(1)
-        if not android:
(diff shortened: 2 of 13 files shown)


python/servo/gstreamer.py
```diff
@@ -242,15 +242,15 @@ def find_non_system_dependencies_with_otool(binary_path: str) -> Set[str]:
     return output
 
 
-def package_gstreamer_dylibs(binary_path: str, library_target_directory: str, cross_compilation_target: str = None):
+def package_gstreamer_dylibs(binary_path: str, library_target_directory: str, target):
```
mrobinson, line 245: You're missing a type annotation here, I think.

python/servo/package_commands.py
```diff
@@ -132,31 +133,21 @@ class PackageCommands(CommandBase):
...
+    def package(self, build_type: BuildType, flavor=None, with_asan=False):
         env = self.build_env()
-        binary_path = self.get_binary_path(build_type, target=target, android=android, asan=with_asan)
+        binary_path = self.get_binary_path(build_type, asan=with_asan)
         dir_to_root = self.get_top_dir()
         target_dir = path.dirname(binary_path)
-        if android:
-            android_target = self.config["android"]["target"]
-            if "aarch64" in android_target:
+        if isinstance(self.target, AndroidTarget):
```
mrobinson, line 142: It might be nice to add a helper for this in the base class `def is_android(...)`.

python/servo/platform/linux.py
```diff
@@ -7,11 +7,12 @@
 # option. This file may not be copied, modified, or distributed
 # except according to those terms.
 
+import distro
 import os
 import subprocess
 from typing import Optional, Tuple
 
-import distro
+from ..build_target import BuildTarget
```
mrobinson, line 15: Would it make sense to move `BuildTarget` into `platform` to avoid having this module depend on things from the parent?

python/servo/testing_commands.py
```diff
@@ -321,12 +321,10 @@ def test_wpt_android(self, build_type: BuildType, binary_args=None, **kwargs):
         )
         return self._test_wpt(build_type=build_type, android=True, **kwargs)
 
-    def _test_wpt(self, build_type: BuildType, with_asan=False, android=False, **kwargs):
-        if not android:
-            os.environ.update(self.build_env())
```
mrobinson, line 326: Why is it possible to remove this now? This allows passing environment variables into the WPT runner.

mrobinson (changes requested):
Nice cleanup. Just a couple small comments.

python/servo/gstreamer.py
mukilan: I had originally left out the type annotation as this file is executed directly as a script from `components/servo/build.rs`.  In [6245f0b](https://github.com/servo/servo/pull/33114/commits/6245f0b30541dcff311902e33cfc44b8518139eb) I've added logic to update `sys.path` to allow imports.

python/servo/platform/linux.py
mukilan: Done in [6245f0b](https://github.com/servo/servo/pull/33114/commits/6245f0b30541dcff311902e33cfc44b8518139eb).

python/servo/testing_commands.py
mukilan: This is a stray change that should have not have been in the PR. I've restored it now, but all the tests on CI were passing on my fork even without this change. Is the use case here to pass the GStreamer related environment variables on Mac, from `build_env()` to the runner?

mrobinson: I think the idea is that if the WPT runner is reading environment variables, they will be passed properly to the runner.

I'm fine with removing it here if it isn't used on CI. If we find a situation where we need to pass environment variables, we can re-add it with a comment (and probably also for Android as well).

mukilan: The environment is not passed as a parameter to `wpt.run.run_tests` and `run_test` directly reads from the global `os.environ`, so the only impact of removing this logic should be that the additional variables added by `build_env()` will no longer be passed to the runner. If the runner was previously reading other variables from parent environment, it should still be able to do that without this logic, so I'll remove it.

mukilan:
@mrobinson Thanks for the review! I've addressed all the comments.

mrobinson (approved):

PR #33117, opened 2024-08-19
ohos: Present on vsync signal

jschwe:
Exposes `present` as a dedicated function, so it can be called synchronized with the vsync signal.
Swapping buffers based on the vsync signal makes scrolling more smooth.
Follow-up PRs will also use the vsync events to implement fling and potentially also trigger animation updates (instead of the currently hardcoded 16ms)

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4781,6 +4781,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0f2fd4c1b349c53cd06dfa959bb53076453960d793961caafd367e64c3bb5522"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


mrobinson (approved):

mukilan:
Hmm...`ohos-vsync 0.1.0` still needs `ohos-sys 0.2.2`, but main is using `ohos-sys 0.3.0`. @jschwe can a new version of `ohos-vsync` be published so we can avoid the duplicate?

jschwe:
Thanks for the ping. I released version 1.1 of ohos-vsync with a dependency bump.

PR #33121, opened 2024-08-19
Update webgl conformance tests

jdm:
We're several years behind the latest conformance test suite changes in https://github.com/KhronosGroup/WebGL/. This also includes python2->3 changes for the script to import them.


sagudev (approved):
LGTM, although it would be nice to store COMMIT file of current checkout. For webgpu we have https://github.com/servo/servo/blob/main/tests/wpt/webgpu/tests/checkout_commit.txt that is autoupdated on import.

jdm:
> LGTM, although it would be nice to store COMMIT file of current checkout. For webgpu we have https://github.com/servo/servo/blob/main/tests/wpt/webgpu/tests/checkout_commit.txt that is autoupdated on import.

Good idea! I added a commit that does this.

PR #33134, opened 2024-08-20
Update url setters test result

Taym95:
---

Maybe I am missing something, but this test does not crash


jdm (approved):
Yeah, that's a result from before it had test variants added (each ? block is a different run of the same test file).

Issue #33140, opened 2024-08-20
Create a static analysis to detect mutable borrow GC hazards


jdm:
https://github.com/servo/servo/pull/33133 contains examples of borrow hazards discovered when a GC occurs while a DOM object's field has a mutable borrow outstanding. It _should_ be possible to detect these automatically at compilation time if we can accurately answer the question "can this function call perform a GC?"

We could build up to this iteratively:
1. create a `struct CanGc(());` which can be copied freely
2. any code in Servo that calls a method from mozjs/rust-mozjs must have a `CanGc` argument passed into it, indicating that it calls a method which can perform a GC operation
3. the compiler now forces all code that calls these methods to pass a `CanGc` argument, so each caller must also receive a new `CanGc` argument
4. step 3 repeats all the way up the stack to code that is automatically called by generated bindings or the main script event loop

Once this data is available, it _should_ be possible to write an analysis pass as a new linter (similar to crown) which uses a dataflow analysis to find any function calls containing a `CanGc` argument that occur within the lifetime of a `borrow_mut()` return value. Even without this static analysis, though, the presence of the arguments will make it easier to visually determine if there are hazards present.

sagudev:
> should be possible to write an analysis pass as a new linter (similar to crown)

Actually I think this should live in crown, but as separate lints, as I do not think we can use multiple rustc drivers at once.

> any code in Servo that calls a method from mozjs/rust-mozjs must have a CanGc argument passed into it

We could enforce this by exposing methods in mozjs using CanGc argument (although I am not sure all mozjs functions can trigger GC).

jdm:
Agreed! It could incrementally work down the stack into rust-mozjs as well for better accuracy.

jdm:
https://github.com/servo/servo/issues/33926 is an example of one hazard that was found by code inspection based on the new CanGc annotations, and then verified by running with GC zeal turned up to maximum in a debug mozjs build.

sagudev:
I was wondering are we sure that can_gc argument even if not used gets optimized out, if not might not be zero cost safety check. I think making `#[crown::cangc]` attribute and adding it to the function would be more correct/easier from crown's standpoint, but it will not be helpful help until we get lint, while current can_gc arguments are already useful.

NOTE: Crown lint will be similar to https://rust-lang.github.io/rust-clippy/master/index.html#/borrow_mut.

I can write the lint if you are not already working on it.

[...]

PR #33144, opened 2024-08-20
Start marking functions that can transitively trigger a GC

jdm:
Per #33140, adding this zero-size argument to functions that can transitively GC makes it easier to recognize where hazards are present. This work can be done incrementally and iteratively as time allows.

---
- [x] These changes are part of #33140.
- [x] These changes do not require tests because there is no functional difference.

### DIFF
--- a/components/script/dom/bindings/reflector.rs
+++ b/components/script/dom/bindings/reflector.rs
@@ -15,7 +15,7 @@ use crate::dom::bindings::root::{Dom, DomRoot, Root};
 use crate::dom::bindings::trace::JSTraceable;
 use crate::dom::globalscope::GlobalScope;
 use crate::realms::AlreadyInRealm;
-use crate::script_runtime::{JSContext, CanGc};
+use crate::script_runtime::{CanGc, JSContext};
 
 /// Create the reflector for a new DOM object and yield ownership to the
 /// reflector.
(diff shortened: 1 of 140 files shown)


jdm:
This PR gets us to a point where I can file followups for less experienced contributors to help.oush it forward. There are many threads to pull!

components/script/dom/bindings/reflector.rs
```diff
@@ -25,20 +25,29 @@ where
...
+            CanGc::note(),
+        )
+    }
 }
 
 pub fn reflect_dom_object_with_proto<T, U>(
     obj: Box<T>,
     global: &U,
     proto: Option<HandleObject>,
+    can_gc: CanGc,
```
gterzian, line 43: is there a future need to pass it here, as opposed to instantiate it inside the function and hide it from the caller?

gterzian (approved):
LGTM, with one question.

components/script/dom/bindings/reflector.rs
jdm: Yes. The purpose is to infect callers so it is clear whether any function can trigger a GC.

jdm:
@last-genius Good question! A useful model for adding arguments to the generated methods is the `InRealm` arguments, which are manually annotated in Bindings.conf. there is code in CodegenRust.py that checks for these annotations and adds the argument if they're present.

PR #33145, opened 2024-08-21
webxr: Update XRView to latest spec

msub2:
This adds the recommendedViewportScale and requestViewportScale members to XRView. The actual process of dynamic viewport resizing is currently omitted, as this functionality is currently only really used in model-viewer out in the wild, and I can find no examples of it being used in stereoscopic immersive sessions. It seems the original motivation for its addition was performance considerations on monoscopic mobile AR, so this could potentially be revisited if Servo supports ARCore/ARKit-backed sessions in the future.

---

### DIFF
--- a/components/script/dom/xrview.rs
+++ b/components/script/dom/xrview.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::cell::Cell;
+
 use dom_struct::dom_struct;
 use euclid::RigidTransform3D;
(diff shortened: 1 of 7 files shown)


components/script/dom/xrview.rs
```diff
@@ -28,6 +31,9 @@ pub struct XRView {
     #[no_trace]
     view: View<ApiSpace>,
     transform: Dom<XRRigidTransform>,
+    viewport_modifiable: Cell<bool>,
+    requested_viewport_scale: Cell<f64>,
+    current_viewport_scale: Cell<f64>,
```
jdm, line 36: If this member is never read or written after the constructor, can we remove it?

jdm (approved):

PR #33146, opened 2024-08-21
webxr: Update XRPose interface to latest spec

msub2:
This adds the `linearVelocity`, `angularVelocity`, and `emulatedPosition` members on XRPose. Exposing velocity will require companion work in the WebXR crate and changes to pose handling, so that will be saved for a follow-up. `emulatedPosition` returns false by default, as there are currently no scenarios in which we would need to rely on emulation for pose reporting.

---


gterzian (approved):
Thank you

PR #33148, opened 2024-08-21
layout: Add initial support for bidirectional text (BiDi)

### LINKED ISSUE #25881: Add support for bidirectional inline layout
https://drafts.csswg.org/css-writing-modes-3/#text-direction

https://crates.io/crates/unicode-bidi implements https://www.unicode.org/reports/tr9/ and does the heavy lifting for text, but we still have to deal with a tree of potentially-nested non-replaced inline boxes whose leafs can be atomic inline-level boxes (e.g. images) as well a text.

mrobinson:
This adds supports for right-to-left text assigning bidi levels to all
line items when necessary. This includes support for the `dir` attribute
as well as corresponding CSS properties like `unicode-bidi`. It only
implements right-to-left rendering for inline layout at the moment and
doesn't include support for `dir=auto`. Because of missing features,
this causes quite a few tests to start failing, as references become
incorrect due to right-to-left rendering being active in some cases,
but not others (before it didn't exist at all).

Analysis of most of the new failures:

```
- /css/css-flexbox/gap-001-rtl.html
  /css/css-flexbox/gap-004-rtl.html
 - Require implementing BiDi in Flexbox, because the start and
   end inline margins are opposite the order of items.

- /css/CSS2/bidi-text/direction-applies-to-*.xht
  /css/CSS2/bidi-text/direction-applies-to-002.xht
  /css/CSS2/bidi-text/direction-applies-to-003.xht
  /css/CSS2/bidi-text/direction-applies-to-004.xht
  - Broken due to a bug in tables, not allocating the
    right amount of width for a column.

- /css/css-lists/inline-list.html
  - This fails because we wrongly insert a soft wrap opportunity between the
    start of an inline box and its first content.

- /css/css-text/bidi/bidi-lines-001.html
  /css/css-text/bidi/bidi-lines-002.html
  /css/CSS2/text/bidi-flag-emoji.html
  - We do not fully support unicode-bidi: plaintext

- /css/css-text/text-align/text-align-end-010.html
  /css/css-text/text-align/text-align-justify-006.html
  /css/css-text/text-align/text-align-start-010.html
  /html/dom/elements/global-attributes/*
  - We do not support dir=auto yet.

- /css/css-text/white-space/tab-bidi-001.html
  - Servo doesn't support tab stops

- /css/CSS2/positioning/abspos-block-level-001.html
  /css/css-text/word-break/word-break-normal-ar-000.html
  - Do not yet support RTL layout in block

- /css/css-text/white-space/pre-wrap-018.html
  - Even in RTL contexts, spaces at the end of the line must hang and
    not be reordered

- /css/css-text/white-space/trailing-space-and-text-alignment-rtl-002.html
  - We are letting spaces hang with white-space: pre, but they shouldn't
    hang.
```

Fixes #25881.

---


Cargo.toml
```diff
@@ -97,32 +97,32 @@ rustls = { version = "0.21.12", features = ["dangerous_configuration"] }
...
 serde_json = "1.0"
 servo-media = { git = "https://github.com/servo/media" }
 servo-media-dummy = { git = "https://github.com/servo/media" }
 servo-media-gstreamer = { git = "https://github.com/servo/media" }
-servo_arc = { git = "https://github.com/servo/stylo", branch = "2024-07-16", features = ["servo"] }
-servo_atoms = { git = "https://github.com/servo/stylo", branch = "2024-07-16" }
-size_of_test = { git = "https://github.com/servo/stylo", branch = "2024-07-16" }
+servo_arc = { git = "https://github.com/servo/stylo", rev = "refs/pull/65/head", features = ["servo"] }
+servo_atoms = { git = "https://github.com/servo/stylo", rev = "refs/pull/65/head" }
+size_of_test = { git = "https://github.com/servo/stylo", rev = "refs/pull/65/head" }
```
atbrakhi, line 109: shall we try to land this stylo change now?

atbrakhi (approved):

Cargo.toml
mrobinson: Yep. It should be ready to go now.

PR #33149, opened 2024-08-21
layout: Clean up inline layout data structures

mrobinson:
- Rename `InlineFormattingContextState` to
  `InlineFormattingContextLayout`.
- Have `InlineFormattingContextLayout` hold a reference to the
 `InlineFormattingContext`, so that it does not need to be passed
 around as an argument
- Have `LineItemLayout` hold a reference to
  `InlineFormattingContextLayout` to avoid duplicating so much data.
- Rename some members of `LineItemLayout` to make it clearer what
  they do.
- Give beter names to many lifetimes and combine some that are
  effectively the same.

---


mrobinson:
Okay. This should have proper co-authorship and be signed off now. It's ready for a review.

atbrakhi (approved):

PR #33150, opened 2024-08-21
CI: Add separate Lint&Tidy check and remove test-tidy from linux

### LINKED ISSUE #32112: Create automatic clippy tidiness tests
Now that most [clippy warnings](https://github.com/servo/servo/issues/31500) are dealt with, it would be interesting to discuss if an automatic clippy lint would make sense for the project.

**Upsides:**
- Catch bugs
- Produce cleaner code
- Avoid big cleanups in the future

**Downsides:**
- Increase the tidy check run time
- False positives
- Can be annoying at times

## Proposal

One way it could work is having it as an extra check of `test-tidy`, with a flag such as `--clippy false` to skip the clippy tests. The clippy warnings could be treated as information, printing them so you can fix (or ignore) them, but so that they don't fail the CI. clippy errors probably should still be errors and make the test fail, because they could indicate a more serious issue. The main problem I see with this approach is that `test-tidy` could take longer, specially on a first build, since it needs to compile the code to run clippy.

### Other options

Some other ideas I had but I'm not convinced with were:

- Running clippy just before the pr enters the merge queue, as a final check. This can slow down merging prs quite a bit.
- Running clippy on the pull request CI, sending a bot comment with the suggestions. This feels very spammy.

---

This feels like a difficult design decision with quite some pros and cons, and there may be much better ways of handling this, that's why I opened this issue to discuss how should we proceed.

sagudev:
Lint&Tidy runs clippy with crown (Lint) then `test-tidy` (Tidy) checks.

---

- [x] These changes partially fix #32112 (we still want to fail on warnings and do anotations)
- [x] These changes do not require tests because CI


mrobinson (changes requested):
This is looking great! I think we can now remove the installation of `cargo-deny` and `taplo` from all other workflows as well.

jschwe:
> My plan is to pass --skip-lints, but that unfortunately includes crown too (we still need it)

Is it not sufficient to just run `crown` in the lint stage?

.github/workflows/lint-tidy.yml
```diff
@@ -0,0 +1,59 @@
...
+      # `pull_request_target` event.
+      - uses: actions/checkout@v4
+        if: github.event_name == 'pull_request_target'
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 2 # This is necessary for `test-tidy`.
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.4
+      - name: Set LIBCLANG_PATH env # needed for bindgen in mozangle
+        run: echo "LIBCLANG_PATH=/usr/lib/llvm-14/lib" >> $GITHUB_ENV
```
jschwe, line 33: I would also specify `CLANG_PATH` to point to clang-14 . (Otherwise there is chance we run into https://github.com/rust-lang/rust-bindgen/issues/2682 when upgrading the ubuntu image to 24.04 and the default clang version changes. Explicitly setting CLANG_PATH, should make it clearer that both versions need to be in sync.

sagudev: There is no clang 14 in 24.04. I usually just hope for the best (if it's not broken, don't fix it), but IMO clang-sys and bindgen should fix their stuff.

sagudev:
My plan is to cache crown install after https://github.com/baptiste0928/cargo-install/issues/27 is fixed.

.github/workflows/dispatch-workflow.yml
```diff
@@ -50,6 +50,12 @@ jobs:
       unit-tests: ${{ inputs.unit-tests }}
       wpt-args: ${{ inputs.wpt-args }}
 
+  lint:
+    if: ${{ inputs.workflow == 'lint' }}
+    name: Lint & Tidy
```
mrobinson, line 55: Super nit:

```suggestion
    name: Lint
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,59 @@
...
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
+  CCACHE: "sccache"
+  CARGO_INCREMENTAL: 0
+
+jobs:
+  lint:
+    name: Lint & Tidy
```
mrobinson, line 16:
```suggestion
    name: Lint
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  workflow_call:
+  workflow_dispatch:
+
+env:
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
+  CCACHE: "sccache"
```
mrobinson, line 11: I think these aren't used currently...?

.github/workflows/main.yml
```diff
@@ -38,6 +38,11 @@ jobs:
       wpt-layout: ${{ github.event_name == 'pull_request' && 'none' || '2020' }}
     secrets: inherit
 
+  lint:
+    name: Lint & Tidy
```
mrobinson, line 42:
```suggestion
    name: Lint
```

python/servo/try_parser.py
```diff
@@ -104,6 +105,8 @@ def handle_preset(s: str) -> Optional[JobConfig]:
                          wpt_args="--processes 1 _webgpu",  # run only webgpu cts
                          profile="production",  # WebGPU works to slow with debug assert
                          unit_tests=False)  # production profile does not work with unit-tests
+    elif s in ["lint", "tidy"]:
+        return JobConfig("Lint & Tidy", Workflow.LINT)
```
mrobinson, line 109:
```suggestion
        return JobConfig("Lint", Workflow.LINT)
```

python/servo/try_parser.py
```diff
@@ -222,7 +225,14 @@ def test_empty(self):
                                   "profile": "release",
                                   "unit_tests": False,
                                   "wpt_args": ""
-                              }
+                              },
+                              {
+                                  "name": "Lint & Tidy",
```
mrobinson, line 230:
```suggestion
                                  "name": "Lint",
```

mrobinson (approved):
Looks good. I prefer the terminology "Lint" versus "Lint & Tidy" simply because the job isn't actually doing a tidy. It's just ensuring that tidy has been run.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,59 @@
+name: Lint & Tidy
```
sagudev, line 1:
```suggestion
name: Lint
```

.github/workflows/lint.yml
sagudev: I think they are for clippy: https://github.com/servo/servo/actions/runs/10508645663/job/29112990211?pr=33150#step:23:5

PR #33154, opened 2024-08-21
webgpu: Sync `GPUBuffer`

sagudev:
`create_new_external_array_buffer` is removed as we need more context to properly ensure safety (we have multiple mut nonoverlapping subslices). Some stuff from #30504.

With this PR we actually PASS more tests on `webgpu:api,validation,buffer,*` then Firefox 🎉

---


components/script/dom/promise.rs
sagudev, line 1: @jdm

jdm: The promise changes look reasonable!

components/webgpu/wgpu_thread.rs
```diff
@@ -1232,6 +1229,7 @@ impl WGPU {
                             .copy_from_slice(&array_buffer);
                         }
                         let result = gfx_select!(buffer_id => global.buffer_unmap(buffer_id));
+                        // TODO: All errors should be silenced here
```
sagudev, line 1232: I have yet to do this, relevant bug in FF: https://bugzilla.mozilla.org/show_bug.cgi?id=1842297

sagudev: DONE!

components/script/dom/gpubuffer.rs
```diff
@@ -30,31 +33,96 @@ use crate::dom::promise::Promise;
...
+    pub fn view(&mut self, range: Range<usize>) -> Result<&MappingView, ()> {
+        if self
+            .views
+            .iter()
+            .any(|view| range_overlap(&view.range, &range))
+        {
+            return Err(());
+        }
+        let cx = GlobalScope::get_cx();
+        /// `freeFunc()` must be threadsafe, should be safely callable from any thread
```
gterzian, line 100: Is there a reason why you didn't just change the signature of `create_new_external_array_buffer` to take an `Arc<Box<[u8]>>`, and store a `HeapBufferSource` as the buffer of `MappingView`? The goal of `create_new_external_array_buffer` was to move the unsafe code into `bindings`, as opposed to do it in the dom object directly. Would there be a way to achieve your goal while maintaining that boundary?

sagudev: create_new_external_array_buffer was actually unsafe, as one can create multiple mut views to same slice, for safety we must pack this in `ActiveBufferMapping` to make sure there is no overlap. We could still have `create_new_external_array_buffer` abstracted away, but I think that would make all safety invariants less clear.

gterzian: Ok I see, then perhaps move `ActiveBufferMapping` to `bindings`? It may be useful elsewhere than only gpubuffer?

sagudev: Not really, as this is from webgpu spec https://gpuweb.github.io/gpuweb/#active-buffer-mapping

gterzian: Ok thanks for pointing it out. I see that the unsafe code seems to relate to https://gpuweb.github.io/gpuweb/#active-buffer-mapping-views, which is a [list](https://infra.spec.whatwg.org/#list)<[ArrayBuffer](https://webidl.spec.whatwg.org/#idl-ArrayBuffer)> so not specific to WebGPU, So perhaps you can add a struct in bindings that encaspulates the idea of a safely managed list of views into a mapping? For example something like a `MappingManager`, with a safe `view` method returning a `MappingView`(which I think could be moved to bindings as well).

sagudev: Done in https://github.com/servo/servo/pull/33154/commits/f792055b697212776611b9d5a2da1b45dc0ee560 (with more generic names).

gterzian: thank you! (I was just gonna say this could be a good follow-up too...)

components/script/dom/promise.rs
```diff
@@ -52,6 +52,12 @@ pub struct Promise {
     permanent_js_root: Heap<JSVal>,
 }
 
+impl PartialEq for Promise {
```
gterzian, line 55: Note: I think this a candidate for something "higher-up", maybe `DomObject`?

components/script/dom/bindings/buffer_source.rs
```diff
@@ -402,44 +407,103 @@ where
...
+        unsafe { ArrayBuffer::from(self.buffer.underlying_object().get()).unwrap() }
+    }
+}
+
+impl Drop for DataView {
+    #[allow(unsafe_code)]
+    fn drop(&mut self) {
+        let cx = GlobalScope::get_cx();
+        assert!(unsafe {
+            js::jsapi::DetachArrayBuffer(*cx, self.buffer.underlying_object().handle())
```
gterzian, line 506: nit: import at top

gterzian (approved):
LGTM, with some nits and notes.

components/script/dom/bindings/buffer_source.rs
sagudev: It is already imported (although wrapped version) which we cannot use here as we have raw handle (actually we could do `Handle::from_raw()`, but I think that would be more work for less profit).

components/script/dom/promise.rs
sagudev: Opened #33191.

Issue #32898, opened 2024-07-31
ReadableStream: finish default controller and reader impl


gterzian:
Follow-up on https://github.com/servo/servo/pull/32730, and needs to target https://github.com/servo/servo/tree/readablestream-re-implementation

TODO: 

- [x] Fix use of `CanGc`. 
- [ ] Use `Dom` in native promise handlers as per https://github.com/servo/servo/issues/33604
- [x] Debug so we can get the same pass rate as now on the WPT tests.

Related to how the stream is used "natively" in Rust:
- [ ] Remove the current use of [`get_js_stream`](https://github.com/servo/servo/pull/32730/files#diff-ea8666cc30b1da7f1a13c2990763c958ef91f1d44209f02bf229d02e09fae7a1R230), which should be replaced by returning the actual `DomRoot<ReadableStream>`. This relates to the "native reading from a stream"  workflow.
- [x] Restructure the "native reading from a stream" [workflow](https://github.com/servo/servo/pull/32730/files#diff-ea8666cc30b1da7f1a13c2990763c958ef91f1d44209f02bf229d02e09fae7a1R306), used by Rust code, so that the code simply uses a `ReadableStreamDefaultReader`. The three methods to restructure are: [`read_a_chunk`](https://github.com/servo/servo/pull/32730/files#diff-ea8666cc30b1da7f1a13c2990763c958ef91f1d44209f02bf229d02e09fae7a1R307), `start_reading`, and `stop_reading`(mostly, or only, used in `components/script/body.rs`).
- [x] Optimize [`get_in_memory_bytes`](https://github.com/servo/servo/pull/32730/files#diff-09888dfa911f121670b8b905131455cf548d1816bf65bdeff76fb00f9e55e743R256)

Related to public JS methods of `ReadableStream`:
- [x] Implement [`Locked`](https://github.com/gterzian/servo/blob/79b7f153332446eba9b9f2b390f3e2f1839bd6eb/components/script/dom/readablestream.rs#L405)
- [x] Implement [`Cancel`](https://github.com/gterzian/servo/blob/79b7f153332446eba9b9f2b390f3e2f1839bd6eb/components/script/dom/readablestream.rs#L411)
- [x] Implement [tee](https://streams.spec.whatwg.org/#readable-stream-tee)
- [x] Fix rooting of composite reason when cancelling ReadableStream::tee https://github.com/servo/servo/issues/34580 fix: https://github.com/servo/servo/pull/34627

[...]

PR #33160, opened 2024-08-22
Implement ReadableStreamDefaultReader

Taym95:
* ReleaseLock
* Closed
* Cancel

Part of #32898

---

### DIFF
--- a/components/script/dom/readablestream.rs
+++ b/components/script/dom/readablestream.rs
@@ -490,32 +490,12 @@ impl ReadableStream {
         // TODO: react to sourceCancelPromise.
     }
 
-    /// <https://streams.spec.whatwg.org/#readable-stream-reader-generic-release>
-    pub fn release_lock(&self, closed_promise: &Promise) {
-        // step 1 & 2
-
-        if self.is_readable() {
-            // step 3
-            closed_promise.reject_error(Error::Type("stream state is not readable".to_owned()));
-        } else {
-            // step 4
-            closed_promise.reject_error(Error::Type(
-                "Cannot release lock due to stream state.".to_owned(),
-            ));
-        }
-
-        // step 5
-        closed_promise.set_promise_is_handled();
-
-        // step 6
-        self.perform_release_steps();
-
-        // step 2.1.8 & 2.1.9
+    pub fn set_reader(&self, new_reader: Option<&ReadableStreamDefaultReader>) {
         match self.reader {
             ReaderType::Default(ref reader) => {
-                reader.set(None);
+                reader.set(new_reader);
             },
-            _ => unreachable!("Releasing lock can only be done with a default reader."),
+            _ => unreachable!("Setting a reader can only be done on a default reader."),
         }
     }
 }
--- a/components/script/dom/readablestream.rs
+++ b/components/script/dom/readablestream.rs
@@ -8,10 +8,9 @@ use std::rc::Rc;
 
 use dom_struct::dom_struct;
 use js::jsapi::{Heap, JSObject};
-use js::jsval::{JSVal, ObjectValue, UndefinedValue};
+use js::jsval::{ObjectValue, UndefinedValue};
 use js::rust::{HandleObject as SafeHandleObject, HandleValue as SafeHandleValue};
 
-use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::QueuingStrategyBinding::QueuingStrategy;
 use crate::dom::bindings::codegen::Bindings::ReadableStreamBinding::{
     ReadableStreamGetReaderOptions, ReadableStreamMethods,
@@ -227,6 +226,14 @@ impl ReadableStream {
         stream
     }
 
+    /// Call into the release steps of the controller,
+    pub fn perform_release_steps(&self) {
+        match self.controller {
+            ControllerType::Default(ref controller) => controller.perform_release_steps(),
+            _ => todo!(),
+        }
+    }
+
     /// Call into the pull steps of the controller,
     /// as part of
     /// <https://streams.spec.whatwg.org/#readable-stream-default-reader-read>
@@ -297,7 +304,7 @@ impl ReadableStream {
 
     /// <https://streams.spec.whatwg.org/#readable-stream-error>
     /// Note: in other use cases this call happens via the controller.
-    pub fn error_native(&self, error: Error) {
+    pub fn error_native(&self, _error: Error) {
         match self.controller {
             ControllerType::Default(ref controller) => {
                 // TODO: use Error.
@@ -459,12 +466,12 @@ impl ReadableStream {
                     reader.close();
                 }
             },
-            ReaderType::BYOB(ref reader) => todo!(),
+            ReaderType::BYOB(ref _reader) => todo!(),
         }
     }
 
     /// <https://streams.spec.whatwg.org/#readable-stream-cancel>
-    fn cancel(&self, promise: &Promise, reason: SafeHandleValue) {
+    pub fn cancel(&self, promise: &Promise, _reason: SafeHandleValue) {
         self.disturbed.set(true);
 
         if self.is_closed() {
@@ -481,6 +488,35 @@ impl ReadableStream {
         // TODO: run the bytes reader steps.
 
         // TODO: react to sourceCancelPromise.
+    }
+
+    /// <https://streams.spec.whatwg.org/#readable-stream-reader-generic-release>
+    pub fn release_lock(&self, closed_promise: &Promise) {
+        // step 1 & 2
+
+        if self.is_readable() {
+            // step 3
+            closed_promise.reject_error(Error::Type("stream state is not readable".to_owned()));
+        } else {
+            // step 4
+            closed_promise.reject_error(Error::Type(
+                "Cannot release lock due to stream state.".to_owned(),
+            ));
+        }
+
+        // step 5
+        closed_promise.set_promise_is_handled();
+
+        // step 6
+        self.perform_release_steps();
+
+        // step 2.1.8 & 2.1.9
+        match self.reader {
+            ReaderType::Default(ref reader) => {
+                reader.set(None);
+            },
+            _ => unreachable!("Releasing lock can only be done with a default reader."),
+        }
     }
 }
 
--- a/components/script/dom/readablestreamdefaultreader.rs
+++ b/components/script/dom/readablestreamdefaultreader.rs
@@ -3,15 +3,15 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use std::collections::VecDeque;
+use std::mem;
 use std::rc::Rc;
-use std::{mem, ptr};
 
 use dom_struct::dom_struct;
-use js::conversions::ToJSValConvertible;
 use js::jsapi::Heap;
 use js::jsval::{JSVal, UndefinedValue};
 use js::rust::{HandleObject as SafeHandleObject, HandleValue as SafeHandleValue};
 
+use super::bindings::root::MutNullableDom;
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::ReadableStreamDefaultReaderBinding::{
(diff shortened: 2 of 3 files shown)


components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,93 @@ impl ReadableStreamDefaultReader {
...
+            stream.perform_pull_steps(ReadRequest::Read(promise.clone()));
+        } else {
+            promise.reject_error(Error::Type("stream is undefined".to_owned()));
+        }
 
         promise
     }
 
     /// <https://streams.spec.whatwg.org/#default-reader-release-lock>
     fn ReleaseLock(&self) {
```
Taym95, line 202: @gterzian @wusyong this can be reviewed.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,93 @@ impl ReadableStreamDefaultReader {
...
+        let error = Error::Type("No chunks are available because the stream is errored".to_owned());
 
-        // TODO: use TypeError.
         // <https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreadererrorreadrequests>
+        // step 3.2
+        self.empty_read_requests();
+        // step 3.3
         for request in self.read_requests.borrow_mut().drain(0..) {
-            request.error_steps();
+            request.error_steps(error.clone());
```
pewsheen, line 243: I think we should clone `self.read_requests` somewhere before clearing it. So that on the step 3.3 we can perform `error_steps()` to those `read_requests`

pewsheen: or maybe we can do `drain` only?

components/script/dom/readablestream.rs
```diff
@@ -213,6 +213,14 @@ impl ReadableStream {
         }
     }
 
+    // Call into the rellease steps of the controller,
```
gterzian, line 216: typo

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,93 @@ impl ReadableStreamDefaultReader {
...
     }
 
     /// <https://streams.spec.whatwg.org/#generic-reader-cancel>
-    fn Cancel(&self, _cx: SafeJSContext, _reason: SafeHandleValue) -> Rc<Promise> {
-        // TODO
-        Promise::new(&self.reflector_.global())
+    fn Cancel(&self, _cx: SafeJSContext, reason: SafeHandleValue) -> Rc<Promise> {
+        let promise = Promise::new(&self.reflector_.global());
+        // step 1
+        if self.stream.get().is_none() {
```
gterzian, line 256: I think it would be clearer to invert the logic, by doing `if let Some(stream) =`, and doing step 1 in the `else`.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,93 @@ impl ReadableStreamDefaultReader {
...
     /// <https://streams.spec.whatwg.org/#default-reader-release-lock>
     fn ReleaseLock(&self) {
-        if self.stream.is_readable() {
-            self.closed_promise.reject_native(&());
+        // step 1
+        if self.stream.get().is_none() {
+            return;
+        }
+
+        // <https://streams.spec.whatwg.org/#readable-stream-reader-generic-release>
```
gterzian, line 208: let's put this in its own method, it is used in many other places in the spec, and also it's clearer with the spec link. We can do the same with other spec functions.

gterzian (changes requested):
Thank you! I think a good next step would be to break-down the algos in their own methods where appropriate.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,93 @@ impl ReadableStreamDefaultReader {
...
 
-        // TODO: https://streams.spec.whatwg.org/#readable-stream-reader-generic-release
+        // step 2.1.6
+        self.closed_promise.set_promise_is_handled();
+
+        // step 2.1.7
+        stream.perform_release_steps();
+
+        // step 2.1.8 & 2.1.9
+        self.stream.set(None);
```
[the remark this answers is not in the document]
Taym95: Yeah I wanted to do that, but when we set `stream` to `None` then`reader` will be dropped so it is also undefined?

components/script/dom/readablestreamdefaultreader.rs
Taym95: indeed thanks

components/script/dom/readablestream.rs
```diff
@@ -423,6 +431,47 @@ impl ReadableStream {
...
         // TODO: react to sourceCancelPromise.
     }
+
+    /// <https://streams.spec.whatwg.org/#readable-stream-reader-generic-release>
+    pub fn release_lock(
+        &self,
+        closed_promise: &Promise,
+        read_requests: DomRefCell<VecDeque<ReadRequest>>,
+    ) {
+        // step 2.1.3
```
gterzian, line 441: I think you can start counting from 1, since you are in the generic algo.

components/script/dom/readablestream.rs
```diff
@@ -423,6 +431,47 @@ impl ReadableStream {
...
+        closed_promise.set_promise_is_handled();
+
+        // step 2.1.7
+        self.perform_release_steps();
+
+        // step 2.1.8 & 2.1.9
+        self.set(None);
+
+        // step  3.1
+        let error = Error::Type("No chunks are available because the stream is errored".to_owned());
```
gterzian, line 464: This can be moved to a method implementing https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreadererrorreadrequests on the reader.  You can use the code at https://streams.spec.whatwg.org/#readable-stream-error and move it to a new method called in both places.

components/script/dom/readablestream.rs
```diff
@@ -423,6 +431,47 @@ impl ReadableStream {
...
+
+        let read_requests_cloned = read_requests.clone();
+        // <https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreadererrorreadrequests>
+        // step 3.2
+        read_requests.borrow_mut().clear();
+        // step 3.3
+        for request in read_requests_cloned.borrow_mut().drain(0..) {
```
gterzian, line 471: I don't think you need the two steps above, and here you should do something like [here](https://github.com/servo/servo/blob/1ff141d486ddd0f28c1af45c9f873154138c9b47/components/script/dom/readablestreamdefaultreader.rs#L147), because the error steps can call back into the stream and cause a borrow error(but see above comment, you should re-use the existing code anyway).

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -74,10 +75,10 @@ impl ReadRequest {
     }
 
     /// <https://streams.spec.whatwg.org/#ref-for-read-request-close-step>
-    pub fn error_steps(&self) {
+    pub fn error_steps(&self, error: Error) {
         match self {
             // TODO: pass error type.
```
gterzian, line 80: todo can be removed.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -161,11 +161,11 @@ impl ReadableStreamDefaultReader {
     }
 
     /// <https://streams.spec.whatwg.org/#readable-stream-error>
-    pub fn error(&self, _error: Error) {
+    pub fn error(&self, error: Error) {
         self.closed_promise.reject_native(&());
         let pending_requests = mem::take(&mut *self.read_requests.borrow_mut());
```
gterzian, line 166: this part can be put in a separate method and called elsewhere(see above)

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -176,43 +176,57 @@ impl ReadableStreamDefaultReader {
             .pop_front()
             .expect("Reader must have read request when remove is called into.")
     }
+
+    /// <https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreadererrorreadrequests>
+    pub fn empty_read_requests(&self) {
```
gterzian, line 181: this seems unused, and perhaps not necessary(see other comments).

gterzian (changes requested):
Looks good, with some changes suggested.

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -169,45 +165,81 @@ impl ReadableStreamDefaultReader {
...
 
 impl ReadableStreamDefaultReaderMethods for ReadableStreamDefaultReader {
     /// <https://streams.spec.whatwg.org/#default-reader-read>
     fn Read(&self) -> Rc<Promise> {
         let promise = Promise::new(&self.reflector_.global());
 
-        self.stream
-            .perform_pull_steps(ReadRequest::Read(promise.clone()));
+        if let Some(stream) = self.stream.get() {
+            stream.perform_pull_steps(ReadRequest::Read(promise.clone()));
```
gterzian, line 180: This should call into https://streams.spec.whatwg.org/#readable-stream-default-reader-read (we don't have that method yet).

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -169,45 +165,81 @@ impl ReadableStreamDefaultReader {
...
+                rooted!(in(*cx) let mut rval = UndefinedValue());
+                unsafe {
+                    error
+                        .clone()
+                        .to_jsval(*cx, &*self.global(), rval.handle_mut())
+                };
+                request.error_steps(rval.handle());
+            }
+        } else {
+            // step 1
```
gterzian, line 220: I think we should also follow the spec more closely:

- add a conditional on `self.stream.get().is_none`().
- Call into a new method for https://streams.spec.whatwg.org/#abstract-opdef-readablestreamdefaultreaderrelease
- Move rest of algo to that new method(it can be called from other places later).

gterzian (changes requested):
Looks good, with a couple of points.

components/script/dom/readablestreamdefaultreader.rs
Taym95: I needed to changed this code because I converted stream to `MutNullableDom<ReadableStream>`, this `read` function is not implemented correctly, I think we can fix it in next PR when we add https://streams.spec.whatwg.org/#readable-stream-default-reader-read

components/script/dom/readablestream.rs
```diff
@@ -482,6 +489,35 @@ impl ReadableStream {
...
+    /// <https://streams.spec.whatwg.org/#readable-stream-reader-generic-release>
+    pub fn release_lock(&self, closed_promise: &Promise) {
+        // step 1 & 2
+
+        if self.is_readable() {
+            // step 3
+            closed_promise.reject_error(Error::Type("stream state is not readable".to_owned()));
+        } else {
+            // step 4
+            closed_promise.reject_error(Error::Type(
```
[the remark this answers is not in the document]
Taym95: indeed I missed this:

Otherwise, set reader.[[[closedPromise]]](https://streams.spec.whatwg.org/#readablestreamgenericreader-closedpromise) to [a promise rejected with](https://webidl.spec.whatwg.org/#a-promise-rejected-with) a [TypeError](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror) exception.

thanks

components/script/dom/readablestreamdefaultreader.rs
```diff
@@ -169,45 +165,135 @@ impl ReadableStreamDefaultReader {
...
-            // TODO: use TypeError.
-            let cx = GlobalScope::get_cx();
-            rooted!(in(*cx) let mut rval = UndefinedValue());
-            request.error_steps(rval.handle());
+        if self.stream.get().is_none() {
+            // step 1
+            return;
+        } else {
+            // step 2
+            self.release();
```
Taym95, line 278: @gterzian I tired to follow the spec more closely, please check again, thanks

gterzian (approved):
LGTM, thanks!

PR #33163, opened 2024-08-23
Allow prefs to be overridden from a file and set WPT-specific prefs from file

msub2:
This adds the ability to pass in a separate prefs file to override prefs.json. It also adds a new `wpt-prefs.json` file in the resources, designed to designate any prefs that should be overridden in Servo's wpt.fyi runs. 

---

### DIFF
--- a/ports/servoshell/desktop/cli.rs
+++ b/ports/servoshell/desktop/cli.rs
@@ -45,6 +45,12 @@ pub fn main() {
         "A preference to set to disable",
         "dom.webgpu.enabled=false",
     );
+    opts.optmulti(
+        "",
+        "prefs-file",
+        "Load in additional prefs from a file.",
+        "--prefs-file /path/to/prefs.json",
+    );
 
     let opts_matches;
     let content_process_token;
--- a/ports/servoshell/prefs.rs
+++ b/ports/servoshell/prefs.rs
@@ -22,14 +22,16 @@ pub fn register_user_prefs(opts_matches: &Matches) {
         .filter(|path| path.exists());
 
     let mut userprefs = if let Some(path) = user_prefs_path {
-        let mut file = File::open(path).expect("Error opening user prefs");
-        let mut txt = String::new();
-        file.read_to_string(&mut txt)
-            .expect("Can't read user prefs file");
-        prefs::read_prefs_map(&txt).expect("Can't parse user prefs file")
+        read_prefs_file(path)
     } else {
         HashMap::new()
     };
+
+    let prefs_from_files: Vec<HashMap<String, PrefValue>> = opts_matches
+        .opt_strs("prefs-file")
+        .iter()
+        .map(|path| read_prefs_file(path))
+        .collect();
 
     let argprefs: HashMap<String, PrefValue> = opts_matches
         .opt_strs("pref")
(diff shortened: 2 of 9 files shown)


mrobinson:
How would wpt.fyi specify that it needs to load the extra preferences file?

msub2:
After this there would be a companion PR to the WPT repo to ensure the Servo runner (https://github.com/web-platform-tests/wpt/blob/master/tools/wptrunner/wptrunner/executors/executorservo.py) has the flag(s) from this file applied.

sagudev:
> After this there would be a companion PR to the WPT repo to ensure the Servo runner (https://github.com/web-platform-tests/wpt/blob/master/tools/wptrunner/wptrunner/executors/executorservo.py) has the flag(s) from this file applied.

You can actually just do this in this PR and it will get exported to upstream wpt by editing https://github.com/servo/servo/blob/main/tests/wpt/tests/tools/wptrunner/wptrunner/executors/executorservo.py. Also are we sure this prefs will only be applied for wpt.fyi runs not for `mach test-wpt` runs too. I wonder how firefox does this.

jdm:
I would actually like to start specifying all directiry-level prefs for wpt test runs in the json file instead of in ini files, since it will make our test behaviour more consistent.

msub2:
I believe the current __dir__.ini overrides would be:

- dom.bluetooth.enabled
- dom.gamepad.enabled
- dom.imagebitmap.enabled
- dom.permissions.enabled
- dom.resize_observer.enabled
- dom.serviceworker.enabled
- dom.webvtt.enabled
- dom.webxr.test
- dom.webgl2.enabled

jdm:
Also:
* layout.columns.enabled
* layout.flexbox.enabled

We had a bunch of preferences enabled in meta-legacy-layout that didn't get migrated to meta, apparently:
* dom.worklet.enabled
* dom.worklet.timeout_ms
* dom.webvtt.enabled
* dom.offscreen_canvas.enabled
* dom.permissions.enabled
* dom.bluetooth.enabled
* dom.mutation_observer.enabled
* dom.serviceworker.enabled
* dom.imagebitmap.enabled

Although I now realize that this goes back to my concern that wpt.fyi results should reflect what users experience as much as possible, so I'm talking myself out of moving most of these to wpt-prefs.json.

msub2:
Yeah, I was thinking about that too. Would we want to maybe start with just adding `dom.webxr.test`? It shouldn't affect our own WPT runs since it's already part of a __dir__.ini.

jdm:
Yes, I think that makes sense.

msub2:
I've confirmed this is correctly applying the dom.webxr.test pref in local testing

jdm (approved):

jdm:
If you set the xrtest pref to false in the ini file for interfaces.html, that test should pass.

sagudev:
> ⛔ Failed to properly merge the upstream pull request ([web-platform-tests/wpt#47769](https://github.com/web-platform-tests/wpt/pull/47769)). Please address any CI issues and try to merge manually.

We should make that PR are blocked until wpt CI is happy too (although in this case it looks like WPTs flake).

msub2:
Hmmm, it seems like this just broke the most recent WPT run 😕

msub2:
I wonder if it's because it pulled a nightly for testing that doesn't have the wpt-prefs.json?

mrobinson:
> > ⛔ Failed to properly merge the upstream pull request ([web-platform-tests/wpt#47769](https://github.com/web-platform-tests/wpt/pull/47769)). Please address any CI issues and try to merge manually.
> 
> We should make that PR are blocked until wpt CI is happy too (although in this case it looks like WPTs flake).

I don't think we should block Servo PRs due to WPT CI issues. We have various WPT reviewers on the project. When things go wrong, maybe we can apply a label to the PR to signal one of the reviewers to fix the WPT usptream PR and land it manually.

mukilan:
This change is causing WPT runs to crash. I've raised #33202 as a quick fix for this.

PR #21371, opened 2018-08-09
[WIP] Generic script crate

hrvolapeter:
Separation is based on adding types to TypeHolderTrait which contains
associated types with traits that script crate actually needs. This allows
using static methods and Sized types.

Affects servo in two ways:
- generic structs do not compile in script crate but are left for later monomorphization
- dom_structs itself are moved to separate crate

TODO:
- [ ] tests
- [ ] performance
- [ ] polyshing

Fixes #1799

r? @jdm

---

- [ ] `./mach build -d` does not report any errors
- [ ] `./mach test-tidy` does not report any errors
- [ ] These changes fix #1799 (github issue number if applicable).

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because _____

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/21371)

Issue #1799, opened 2014-03-03
Split up the script crate for build speed


kmcallister:
It's by far the largest crate in terms of binary size (2.5× as big as `main`).  Building it takes almost 40 seconds and 1.8 GB resident on my machine (with debugging enabled, optimization disabled).

The various DOM types and their generated bindings mostly don't depend on each other.  We could split them among a few crates to improve compile time (especially in parallel).

jdm:
Splitting the script crate in a linear fashion may be more feasible. If we were able to move all of the generated bindings into a crate that script depends upon, that would mean that bindings and layout could build in parallel, and would reduce the working memory required to compile the rest of script. It would also speed up making changes to non-codegen parts of script.

antrik:
I'd like to point out that this issue is more serious nowadays, as not only does the `script` crate totally dominate partial rebuild time (it takes several minutes for a non-optimised build now), but also the memory usage has gotten so large that it actually makes debug builds exceed the `3 GiB` virtual memory available on 32 bit systems, as reported in https://github.com/servo/servo/issues/13045 -- and it's probably just a matter of time until non-debug builds break as well...

Ms2ger:
I don't see how that would work. The generated bindings need to be able to inline the method definitions on the concrete type, and the concrete type depends fundamentally on the wrap function, and the class / JSNative definitions with it.

jdm:
Latest idea from discussions with @asajeffrey:
* split generated bindings code into two crates - code that can be parameterized over FooTraits types (to avoid needing to import concrete DOM types), and code that needs to refer to the parameterized code using concrete types
* extract low-level types from script/ that the generated bindings code needs to refer to (e.g. converting native types to JS values) and put them in the generic bindings crate

This should give us the following crates:
js_bindings (generic codegen, generated types (eg. unions, dictioanries), common script/ types)
script (DOM API implementations and non-generic codegen)

This would mean that non-WebIDL/CodegenRust.py changes should not have to rebuild a significant portion of the generated bindings code.

jdm:
Things that will make the above idea difficult:
* concrete DOM types in generated types (eg. unions or dictionaries containing WebIDL interface types)

jdm:
At minimum, this will require collecting the list of WebIDL interfaces transitively used by a given WebIDL property (for each return value and argument), adding generic parameters for each, and propagating these generic types through layers like the union conversion methods when interacting with them.

jdm:
[...]

PR #21105, opened 2018-06-29
Assert that DOM structs have the correct first field

Manishearth:
Not having the right field can lead to fun bugs like https://github.com/ferjm/servo/pull/1 where the struct gets mis-reinterpreted as what should be its parent (but is not layout-wise)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/21105)

PR #33169, opened 2024-08-24
Use global exports from derives

sagudev:
This will make it easier to use derives from multiple crates (in preparations for #1799).

I think this is nicer solution to what was done in #21371, where you needed to manually specify base in derive invocation [`#[base = "script"]`](https://github.com/servo/servo/pull/21371#discussion_r1729789614).

---

- [x] These changes do not require tests because they are refators (rust keeps us safe)

### DIFF
--- a/components/script/dom/bindings/inheritance.rs
+++ b/components/script/dom/bindings/inheritance.rs
@@ -53,6 +53,7 @@ pub trait Castable: IDLInterface + DomObject + Sized {
     }
 }
 
+#[allow(missing_docs)]
 pub trait HasParent {
     type Parent;
     fn as_parent(&self) -> &Self::Parent;
--- a/components/script/lib.rs
+++ b/components/script/lib.rs
@@ -100,3 +100,4 @@ pub use script_runtime::JSEngineSetup;
 // pub exports for derive macros
 pub use crate::dom::bindings::trace::JSTraceable;
 pub use crate::dom::bindings::trace::CustomTraceable;
+pub use crate::dom::bindings::inheritance::HasParent;
--- a/components/script/lib.rs
+++ b/components/script/lib.rs
@@ -97,8 +97,8 @@ mod window_named_properties;
 pub use init::init;
 pub use script_runtime::JSEngineSetup;
 
+// export traits to be available for derive macros
 pub use crate::dom::bindings::inheritance::HasParent;
 pub use crate::dom::bindings::reflector::{DomObject, MutDomObject, Reflector};
 pub use crate::dom::bindings::trace::CustomTraceable;
-// pub exports for derive macros
 pub use crate::dom::bindings::trace::JSTraceable;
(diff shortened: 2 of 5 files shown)


components/script/dom/bindings/inheritance.rs
```diff
@@ -53,6 +53,7 @@ pub trait Castable: IDLInterface + DomObject + Sized {
     }
 }
 
+#[allow(missing_docs)]
```
mrobinson, line 56: It would be nice to add documentation to this trait.

components/script/lib.rs
```diff
@@ -96,3 +96,9 @@ mod window_named_properties;
 
 pub use init::init;
 pub use script_runtime::JSEngineSetup;
+
+pub use crate::dom::bindings::inheritance::HasParent;
+pub use crate::dom::bindings::reflector::{DomObject, MutDomObject, Reflector};
+pub use crate::dom::bindings::trace::CustomTraceable;
+// pub exports for derive macros
```
mrobinson, line 103: I'm not sure I understand this comment. I'm happy to try to write something here if you can give me a bit of input on what it means. Why is it labeling this export and not the ones above?

sagudev: Ah, running `mach fmt` reordered lines.

When using various derive macros (ex. `JSTraceable`) they impl traits for structs, but to find trait `JSTraceable` they used to use full path. For #1799 we need to refer to these traits from various crates, so full path is different in each crate, and to make it easier for derives will simply expect that appropriate traits will be exported in crate root.

components/script/dom/bindings/inheritance.rs
sagudev: Per https://github.com/servo/servo/pull/21105 I think this is only used for parent assertions by codegen/derive.

jdm (approved):

components/script/lib.rs
mrobinson: Thanks for the background:

```
// These derive macro exports are public, because they are used in the DOM bindings
// and it is useful that they are accessible at the root of the crate.
```

sagudev: These are not derive macros, these are traits (that have same name as derive macros, similar is in rust with `Debug` where you can manually implement `Debug` trait or use `derive(Debug)` (`Debug` derive macro) that will automatically implement `Debug` trait for you) that are used by derive macros (derive macros expand to traits impl) and that's way they need to be exported.

sagudev: I rewrote comment to: `export traits to be available for derive macros`, is that clear enough?

mrobinson: How about this:

```

// These trait exports are public, because they are used in the DOM bindings. 
// Since they are used to create derive macros, it is useful that they are accessible at the root of the crate. 
.

```

jdm:
Needs a `mach fmt`.

PR #33174, opened 2024-08-24
script: fix `querySelector` returning the root

nolanlawson:
This fixes a small bug in `Element.prototype.querySelector`:

```js
const button = document.createElement('button')

button.querySelector('*') // should return null, but returns <button> in Servo
```

This fixes a couple WPT tests, which I've also updated.

---

### DIFF
--- a/components/script/dom/node.rs
+++ b/components/script/dom/node.rs
@@ -982,8 +982,10 @@ impl Node {
                     NeedsSelectorFlags::No,
                     MatchingForInvalidation::No,
                 );
-                Ok(self
-                    .traverse_preorder(ShadowIncluding::No)
+                let mut descendants = self.traverse_preorder(ShadowIncluding::No);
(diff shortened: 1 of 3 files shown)


components/script/dom/node.rs
```diff
@@ -982,8 +982,10 @@ impl Node {
                     NeedsSelectorFlags::No,
                     MatchingForInvalidation::No,
                 );
-                Ok(self
-                    .traverse_preorder(ShadowIncluding::No)
+                let mut descendants = self.traverse_preorder(ShadowIncluding::No);
+                // Skip the root of the tree.
+                assert!(&*descendants.next().unwrap() == self);
```
nolanlawson, line 987: This is borrowed from the same check we do in `query_selector_iter` below.

mrobinson (approved):

Issue #33094, opened 2024-08-16
Text rendering issue on duckduckgo.com


atouchet:
**Describe the bug:**
Text that appears on duckduckgo.com exhibits rendering issues on some page loads. The issue does not reproduce 100% of the time.

Image:

![Untitled-1](https://github.com/user-attachments/assets/7773ef8c-3d03-4ff3-8b4c-3f603501075e)

**To Reproduce:**
1. Load duckduckgo.com.
2. Scroll down to where it says Desktop & Mobile Browser Desktop Browser Extension.
3. Notice that there are sometimes issues with the text.

**Platform:**
Windows 10

mrobinson:
@atouchet Would you mind testing this now that #33177 has landed. It's quite possibly related.

atouchet:
I tried the latest nightly build and haven't been able to reproduce the text issue so this may be fixed now. I was also having issues consistently displaying the problem in an earlier build too however so hopefully I'm not just having issues running into the right conditions. I will re-open the report if I run into it again.

PR #33177, opened 2024-08-24
Use webrender `UploadMethod::Immediate` with ANGLE

### LINKED ISSUE #33170: Rendering issue on wikipedia.org
**Describe the bug:**
Top logo/image and text at bottom left don't render correctly (main branch 2db9032e72bda6108b4eb5988eaf164141a61d64):
wikipedia

**To Reproduce:**
`servo https://wikipedia.org`

**Platform:**
Windows 11, x86_64

I tracked down the issue to `webrender`. Servo uses default texture upload method `UploadMethod::PixelBuffer`. In this mode `webrender` uses a single `TextureUploader` (https://github.com/servo/webrender/blob/c0bcdd024adac1297ceb2f34a2de46731243c970/webrender/src/renderer/upload.rs#L79-L83) to upload glyphs and images to GPU. The uploads are staged in one or more buffers before data is sent to GPU (https://github.com/servo/webrender/blob/c0bcdd024adac1297ceb2f34a2de46731243c970/webrender/src/device/gl.rs#L4561-L4598). On Windows if the staged textures are not aligned at bytes per pixel boundary in upload buffer the GPU will read the wrong data.

Using this patch (https://github.com/crbrz/webrender/commit/f88b451e77390283e7bbed9646853f44949ab6e9) for `webrender` fixes the issue on wikipedia:
wiki_good
This issue https://github.com/servo/servo/issues/33094 seems to be fixed as well.

crbrz:
Use the same webrender texture upload method as Gecko (https://searchfox.org/mozilla-central/source/gfx/webrender_bindings/src/bindings.rs#1215-1219) with ANGLE.

---

- [X] These changes fix #33170 and #33094

- [X] These changes do not require tests because they shouldn't change behavior.

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -334,6 +334,8 @@ where
                 clear_color[2] as f32,
                 clear_color[3] as f32,
             );
+            // Use same texture upload method as Gecko with ANGLE:
+            // https://searchfox.org/mozilla-central/source/gfx/webrender_bindings/src/bindings.rs#1215-1219
             let upload_method = if webrender_gl.get_string(RENDERER).starts_with("ANGLE") {
                 UploadMethod::Immediate
             } else {
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -66,6 +66,7 @@ use fonts::FontCacheThread;
 ))]
 use gaol::sandbox::{ChildSandbox, ChildSandboxMethods};
 pub use gleam::gl;
+use gleam::gl::RENDERER;
 use ipc_channel::ipc::{self, IpcSender};
 #[cfg(feature = "layout_2013")]
 pub use layout_thread_2013;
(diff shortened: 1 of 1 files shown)


components/servo/lib.rs
```diff
@@ -333,6 +334,11 @@ where
                 clear_color[2] as f32,
                 clear_color[3] as f32,
             );
+            let upload_method = if webrender_gl.get_string(RENDERER).starts_with("ANGLE") {
```
Taym95, line 337: can't we create something like Firefox [is_glcontext_angle](https://searchfox.org/mozilla-central/source/gfx/layers/wr/WebRenderBridgeParent.cpp#80) ?

components/servo/lib.rs
```diff
@@ -333,6 +334,11 @@ where
...
+                UploadMethod::Immediate
+            } else {
+                UploadMethod::PixelBuffer(ONE_TIME_USAGE_HINT)
+            };
```
Taym95, line 341: Maybe worth referring to Firefox code in comment ?

components/servo/lib.rs
crbrz: Probably with changes to `surfman` otherwise through GL bindings I didn't find another way. `webrender` uses the same check when testing for ANGLE: https://github.com/servo/webrender/blob/c0bcdd024adac1297ceb2f34a2de46731243c970/webrender/src/device/gl.rs#L1724-L1725

Taym95: Yes I see, just that we may end up with lot of `starts_with("ANGLE")` checks in code

jdm (approved):
This seems reasonable for the first time we're checking the renderer name.

PR #33186, opened 2024-08-26
layout: Enable flexbox by default on non-legacy layout

mrobinson:
Flexbox is still very much in progress, but things are working well
enough that we can enable it by default. It improves most pages that use
flexbox now.

Some tests start to pass, but these are exposing existing bugs in flexbox
or other missing features.

---


Loirooriol (approved):

wusyong:
Shoutout to all the hard work! Just tested with https://flexbox.help/ and it seems working very well!

PR #33188, opened 2024-08-26
Add initial support for `tracing` and `tracing-perfetto`

atbrakhi:
This PR introduces the initial integration of `tracing` and `tracing-perfetto`. With this in place, we should be able to land and work on adding tracing events and logging in parallel

Co-authored-by: Delan Azabani <dazabani@igalia.com>
Signed-off-by: Rakhi Sharma <atbrakhi@igalia.com>

---

### DIFF
--- a/ports/servoshell/Cargo.toml
+++ b/ports/servoshell/Cargo.toml
@@ -48,6 +48,7 @@ multiview = ["libservo/multiview"]
 native-bluetooth = ["libservo/native-bluetooth"]
 profilemozjs = ["libservo/profilemozjs"]
 refcell_backtrace = ["libservo/refcell_backtrace"]
+tracing = ["dep:tracing", "dep:tracing-subscriber", "dep:tracing-perfetto"]
 webdriver = ["libservo/webdriver"]
 webgl_backtrace = ["libservo/webgl_backtrace"]
 
@@ -110,6 +111,9 @@ raw-window-handle = "0.6"
 shellwords = "1.0.0"
 surfman = { workspace = true, features = ["sm-x11", "sm-raw-window-handle-06"] }
 tinyfiledialogs = "3.0"
+tracing = { workspace = true, optional = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"], optional = true }
+tracing-perfetto = { workspace = true, optional = true }
 webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "glwindow", "headless"] }
 winit = "0.29.10"
 
--- a/ports/servoshell/desktop/cli.rs
+++ b/ports/servoshell/desktop/cli.rs
@@ -8,12 +8,41 @@ use getopts::Options;
 use log::error;
 use servo::config::opts::{self, ArgumentParsingResult};
 use servo::servo_config::pref;
+#[cfg(feature = "tracing")]
+use tracing::subscriber::set_global_default;
+#[cfg(feature = "tracing")]
+use tracing_perfetto::PerfettoLayer;
+#[cfg(feature = "tracing")]
+use tracing_subscriber::layer::SubscriberExt;
+#[cfg(feature = "tracing")]
+use tracing_subscriber::util::SubscriberInitExt;
+#[cfg(feature = "tracing")]
+use tracing_subscriber::EnvFilter;
 
 use crate::desktop::app::App;
 use crate::panic_hook;
 
 pub fn main() {
     crate::crash_handler::install();
+
+    #[cfg(feature = "tracing")]
+    {
+        let file = std::fs::File::create("servo.pftrace").unwrap();
+        let perfetto_layer = PerfettoLayer::new(std::sync::Mutex::new(file));
+
+        // Set up a custom tracing subscriber and PerfettoLayer for performance tracing.
+        // The servo.pftrace file can be uploaded to https://ui.perfetto.dev for analysis.
+        let env_filter_layer = EnvFilter::from_default_env();
+        let subscriber = tracing_subscriber::registry()
+            .with(env_filter_layer)
+            .with(perfetto_layer);
+
+        // Set the subscriber as the global default for tracing. This means it will be used
+        // throughout the entire program unless a thread-local subscriber is set.
+        // Can only be set once; subsequent attempts will fail.
+        // <https://docs.rs/tracing/0.1.40/src/tracing/subscriber.rs.html#30>
+        set_global_default(subscriber).expect("Failed to set global default subscriber");
+    }
 
     crate::resources::init();
 
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 .coverage
 .DS_Store
 Servo.app
+servo.pftrace
 .config.mk.last
 /glfw
 webrender-captures/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,6 +783,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cmake"
+version = "0.1.51"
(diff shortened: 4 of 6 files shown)


ports/servoshell/desktop/cli.rs
```diff
@@ -8,13 +8,32 @@ use getopts::Options;
...
+    // compositor-to-constellation and script-to-constellation logging.
+    // <https://tokio.rs/tokio/topics/tracing>
+    // <https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/fmt/index.html#composing-layers>
+    let env_filter_layer = EnvFilter::from_default_env();
+    let fmt_layer = tracing_subscriber::fmt::layer();
+    tracing_subscriber::registry()
+        .with(env_filter_layer)
+        .with(fmt_layer)
+        .with(perfetto_layer)
+        .init();
```
mrobinson, line 35: Should this be turned on by a command-line argument which accepts an output filename?

atbrakhi: Oh, I have been thinking about this, but my concern comes from a different reason. This is really slowing down the build. On Mac, it takes around 5 minutes to build.

I've been discussing adding a flag with @delan and will wait for her feedback as well.

delan: yeah, there’s a smaller but still significant regression in build time on linux too (before [135s](https://github.com/user-attachments/files/16761228/cargo-timing-20240827T110216Z.html.gz), after [160s](https://github.com/user-attachments/files/16761233/cargo-timing-20240827T105843Z.html.gz)). what if we gate this behind a `tracing` or `perfetto` feature for now?

atbrakhi: sounds good to me. I have added the `tracing` feature flag in [519e954](https://github.com/servo/servo/pull/33188/commits/519e954d58413b8f9513d30a79ae694a48bc0c1f). I guess moving forward, we can modify it if needed.

ports/servoshell/desktop/cli.rs
```diff
@@ -8,13 +8,39 @@ use getopts::Options;
...
+    {
+        let file = std::fs::File::create("servo.pftrace").unwrap();
+        let perfetto_layer = PerfettoLayer::new(std::sync::Mutex::new(file));
+        // Set up a custom tracing subscriber that behaves like the default fmt
+        // subscriber, but also lets us insert our own layers for things like
+        // compositor-to-constellation and script-to-constellation logging.
+        // <https://tokio.rs/tokio/topics/tracing>
+        // <https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/fmt/index.html#composing-layers>
+        let env_filter_layer = EnvFilter::from_default_env();
+        let fmt_layer = tracing_subscriber::fmt::layer();
```
delan, line 36: i think these can be removed if we’re just using tracing for perfetto right now?

ports/servoshell/desktop/cli.rs
```diff
@@ -8,13 +8,39 @@ use getopts::Options;
...
+        tracing_subscriber::registry()
+            .with(env_filter_layer)
+            .with(fmt_layer)
```
delan, line 39: same goes for these

ports/servoshell/Cargo.toml
```diff
@@ -108,6 +109,9 @@ raw-window-handle = "0.6"
 shellwords = "1.0.0"
 surfman = { workspace = true, features = ["sm-x11", "sm-raw-window-handle-06"] }
 tinyfiledialogs = "3.0"
+tracing = { workspace = true, optional = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"], optional = true }
```
delan, line 113: same goes for the extra feature

delan (changes requested):
This panics on startup with:

```
$ ./mach build -d --features tracing
$ ./mach run -d
```

I think this is related to the [`log` compatibility](https://docs.rs/tracing/0.1.40/tracing/#consuming-log-records) that tracing [enables by default](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/index.html#optional-dependencies)? It seems to happen even if we remove the EnvFilter and fmt layers, as long as we init the registry, so we may need to disable the feature.

atbrakhi:
> This panics on startup with:
> 
> ```
> $ ./mach build -d --features tracing
> $ ./mach run -d
> ```
> 
> I think this is related to the [`log` compatibility](https://docs.rs/tracing/0.1.40/tracing/#consuming-log-records) that tracing [enables by default](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/index.html#optional-dependencies)? It seems to happen even if we remove the EnvFilter and fmt layers, as long as we init the registry, so we may need to disable the feature.

@delan as you said even removing EnvFilter and fmt layers are causing panic. How do you feel about using `set_global_default()`? It should help us avoid panic. I have pushed this change in [a298648](https://github.com/servo/servo/pull/33188/commits/a2986488412812ac9775147752e3efbedf8645f0), and have been testing this changes with #33188

delan:
> How do you feel about using `set_global_default()`? It should help us avoid panic.

Interesting! That does seem to be exactly the difference between [set_global_default](https://docs.rs/tracing/0.1.40/src/tracing/subscriber.rs.html#39-44) and [SubscriberInitExt::init](https://docs.rs/tracing-subscriber/0.3.18/src/tracing_subscriber/util.rs.html#58-92). I think that’s ok if we add a comment explaining it, though it may be simpler to disable the `tracing-log` feature. I also think we should still remove the EnvFilter and fmt layers, unless we need them for profiling.

delan:
(rebased to fix my email address; no actual changes)

atbrakhi:
> > How do you feel about using `set_global_default()`? It should help us avoid panic.
> 
> Interesting! That does seem to be exactly the difference between [set_global_default](https://docs.rs/tracing/0.1.40/src/tracing/subscriber.rs.html#39-44) and [SubscriberInitExt::init](https://docs.rs/tracing-subscriber/0.3.18/src/tracing_subscriber/util.rs.html#58-92). I think that’s ok if we add a comment explaining it, though it may be simpler to disable the `tracing-log` feature. I also think we should still remove the EnvFilter and fmt layers, unless we need them for profiling.

Sorry, should have mentioned in previous comment, even when I remove the `tracing-log` feature and also the EnvFilter and fmt layers, it still crashes. that's how i ended up with `set_global_default`

delan:
> Sorry, should have mentioned in previous comment, even when I remove the `tracing-log` feature and also the EnvFilter and fmt layers, it still crashes, that's how i ended up with `set_global_default`

No worries, in that case `set_global_default` sounds good.

delan:
I noticed that `tracing-log` still appears in `cargo tree -e features` if we specify `default-features = false` in the servoshell Cargo.toml, but disappears if we specify it in the workspace `[dependencies]`:

```toml
# Cargo.toml
tracing-subscriber = { version = "0.3.18", default-features = false }

# ports/servoshell/Cargo.toml
tracing-subscriber = { workspace = true }
```

This seems to be documented [in the Cargo book](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table):

> Specifying a workspace dependency is similar to [package dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) except: […] [features](https://doc.rust-lang.org/cargo/reference/features.html) declared in this table are additive with the features from [dependencies]

ports/servoshell/Cargo.toml
atbrakhi: we need `env-filter` later

delan (approved):

PR #33189, opened 2024-08-26
Add tracing events

atbrakhi:
Depends on #33188 

This PR adds tracing events in `Compositor`, `Constellation`, `libServo` and `Servoshell`

Some Screenshots for reference. Captured tracing events in servo.pftrace

image

image

---

Signed-off-by: Rakhi Sharma <atbrakhi@igalia.com>


components/compositing/Cargo.toml
```diff
@@ -41,6 +41,8 @@ servo_geometry = { path = "../geometry" }
 servo_url = { path = "../url" }
 style_traits = { workspace = true }
 time = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
```
delan, line 45: nit: do we need these tracing-subscriber dependencies outside of servoshell?

delan (approved):

components/compositing/Cargo.toml
atbrakhi: we don't, only `tracing` is needed. Thanks.

PR #33200, opened 2024-08-26
Correct LLVM target triple detection for Apple Silicon

### LINKED ISSUE #33199: `./mach build` broken on MacOS
I don't know much about mach but I can't currently compile the `main` branch, when running `./mach build` I get the following error:

```
No build type specified so assuming `--dev`.
error: toolchain '1.80.1-aarch64-apple-darwin' does not support target 'unknown-apple-darwin'
note: you can see a list of supported targets with `rustc --print=target-list`
note: if you are adding support for a new target to rustc itself, see https://rustc-dev-guide.rust-lang.org/building/new-target.html
Error running mach:

    ['build']

message.

subprocess.CalledProcessError: Command 'rustup target add unknown-apple-darwin' returned non-zero exit status 1.

  File "/Users/benny/src/servo/python/servo/command_base.py", line 612, in configuration_decorator
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benny/src/servo/python/servo/build_commands.py", line 80, in build
    self.ensure_bootstrapped()
  File "/Users/benny/src/servo/python/servo/command_base.py", line 807, in ensure_bootstrapped
    check_call(["rustup", "target", "add", self.target.triple()],
  File "/Users/benny/src/servo/python/servo/command_base.py", line 224, in check_call
    raise subprocess.CalledProcessError(status, ' '.join(*args))
```

It seemed to break with b6d5ac0 since the prior commit starts the build process as usual. Or are there new extra steps one needs to do before? Running `./mach build --target aarch64-apple-darwin` also doesn't work, although the error Message differs:

```
Exception: Unknown build target aarch64-apple-darwin

  File "/Users/benny/src/servo/python/servo/command_base.py", line 608, in configuration_decorator
    self.configure_build_target(kwargs)
  File "/Users/benny/src/servo/python/servo/command_base.py", line 678, in configure_build_target
    self.target = BuildTarget.from_triple(target_triple)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benny/src/servo/python/servo/platform/build_target.py", line 39, in from_triple
    raise Exception(f"Unknown build target {target_triple}")
```

Melchizedek6809:
The LLVM target triple for Apple Silicon was not detected properly.
The Python `platform` module reports the CPU architectures as
`arm64`, but the component in the LLVM triple that Rust uses should
be `aarch64`. 

---

- [X] These changes do not require tests because they are minor changes to the build scripts

### DIFF
--- a/python/servo/platform/__init__.py
+++ b/python/servo/platform/__init__.py
@@ -40,7 +40,7 @@ def host_triple():
         cpu_type = "x86_64"
     elif cpu_type == "arm":
         cpu_type = "arm"
-    elif cpu_type == "aarch64":
+    elif cpu_type in ["aarch64", "arm64"]:
         cpu_type = "aarch64"
     else:
         cpu_type = "unknown"


sagudev:
related? https://github.com/servo/servo/pull/33190

Melchizedek6809:
Seems that way, #33190 fixes the build on my machine as well

mrobinson (approved):
This is still a decent change and explains why the toolchain installation was failing. Thanks for submitting it!

PR #33204, opened 2024-08-27
layout: Add an indefinite containing block for intrinsic sizing

Loirooriol:
When computing the min-content or max-content size of an element we
need to ignore `inline-size`, `min-inline-size` and `max-inline-size`.

However, we should take the block-axis sizing properties into account.
That's because the contents could have percentages depending on them,
which can then affect their inline size via an aspect ratio.

Therefore, this patch adds `IndefiniteContainingBlock`, which is similar
to `ContainingBlock`, but it allows an indefinite inline-size. This
struct is then passed arround during intrinsic sizing.

More refinement will be needed in follow-up patches in order to fully
address the problem.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -9,7 +9,6 @@ use app_units::Au;
 use atomic_refcell::AtomicRefMut;
 use itertools::izip;
 use style::computed_values::position::T as Position;
-use style::logical_geometry::WritingMode;
 use style::properties::longhands::align_items::computed_value::T as AlignItems;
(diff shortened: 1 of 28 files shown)


mrobinson (approved):

Loirooriol:
Simplified a bit the patch: https://github.com/servo/servo/compare/78b591bc73b827659714f22edcb971631d34b456..e30fbcf454f46b2181c328180394aeffd2d9328c

mrobinson (approved):

Loirooriol:
Some fixups were needed after #33241: https://github.com/servo/servo/commit/6069b1195045358ada14406bca635119f998586e

PR #33206, opened 2024-08-27
ohos: Add FFI-APIs to navigate back and forward

jschwe:
To be useful it requires using the [latest version of the OH demo ArkTS app](https://github.com/jschwe/ServoDemo/commit/8aa467a223f7d80b78f3fcf6ed962c9c866f7670), but it is still compatible with older versions, as the newly added FFI functions will simply be unused in such a case.

---

### DIFF
--- a/ports/servoshell/egl/ohos.rs
+++ b/ports/servoshell/egl/ohos.rs
@@ -13,7 +13,6 @@ use std::time::Duration;
 
 use log::{debug, error, info, warn, LevelFilter};
 use napi_derive_ohos::{module_exports, napi};
-use napi_ohos::bindgen_prelude::Undefined;
 use napi_ohos::threadsafe_function::{
     ErrorStrategy, ThreadsafeFunction, ThreadsafeFunctionCallMode,
 };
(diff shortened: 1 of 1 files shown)


ports/servoshell/egl/ohos.rs
```diff
@@ -362,6 +366,16 @@ pub fn load_url(url: String) -> Undefined {
     call(ServoAction::LoadUrl(url)).expect("Failed to load url");
 }
 
+#[napi]
+pub fn go_back() -> Undefined {
```
Taym95, line 370: Question: why do you return `Undefined ` here?

mrobinson (approved):

ports/servoshell/egl/ohos.rs
jschwe: The `napi` macro from `napi-rs` maps `Undefined` / `()` in Rust to `undefined` in JS.
I just tested and the macro does seem to support signatures without an explicit return type, so removing it is probably best.

Taym95: Thanks. Indeed, removing it makes sense; 'Undefined' got me confused.

PR #33210, opened 2024-08-27
Remove measurement of layout query wait time

mrobinson:
Now that the script thread and the layout thread are the same the wait
time effectively zero, so there's no need to measure it. This also
removes one dependency and removes one use of legacy time.

---

- [x] These changes do not require tests because this just removes some effectively useless timing code.

### DIFF
--- a/components/layout_thread/lib.rs
+++ b/components/layout_thread/lib.rs
@@ -27,7 +27,6 @@ use fonts::{
 };
 use fonts_traits::WebFontLoadFinishedCallback;
 use fxhash::{FxHashMap, FxHashSet};
-use histogram::Histogram;
 use ipc_channel::ipc::IpcSender;
 use layout::construct::ConstructionResult;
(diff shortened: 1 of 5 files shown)


components/layout_thread/lib.rs
```diff
@@ -467,18 +463,6 @@ impl Layout for LayoutThread {
     }
 
     fn exit_now(&mut self) {
-        // Drop the root flow explicitly to avoid holding style data, such as
```
jdm, line 470: Looks like the code this comment applies to still exists, so let's revert the comment deletion.

jdm (approved):
Good find!

components/layout_thread/lib.rs
mrobinson: Oh, thanks for spotting this. Tangentially, this is something that doesn't happen in Layout 2020, so I suspect that we do not need it for legacy layout.

PR #33211, opened 2024-08-27
Remove `width` and `height` presentational hints for `<canvas>`

Loirooriol:
According to HTML, the `width` and `height` attributes should only set the natural sizes and the aspect ratio.
The `width` and `height` properties should stay as `initial` by default.

---


mrobinson (approved):
Nice. There are a few more passes.

PR #33212, opened 2024-08-27
Impl PartialEq and Eq for DomObject

### LINKED ISSUE #33191: Impl PartialEq and Eq for `DomObject`
> Note: I think this a candidate for something "higher-up", maybe `DomObject`?

_Originally posted by @gterzian in https://github.com/servo/servo/pull/33154#discussion_r1731315031_

Taym95:
---

- [x] These changes fix #33191

- [x] These changes do not require tests because there are no behavior changes.

### DIFF
--- a/components/dom_struct/lib.rs
+++ b/components/dom_struct/lib.rs
@@ -46,14 +46,6 @@ pub fn dom_struct(args: TokenStream, input: TokenStream) -> TokenStream {
                         &self.#ident
                     }
                 }
-
-                impl Eq for #name {}
-
-                impl PartialEq for #name {
-                    fn eq(&self, other: &Self) -> bool {
-                        crate::DomObject::reflector(self) == crate::DomObject::reflector(other)
-                    }
-                }
             )
             .into()
         } else {
--- a/components/dom_struct/lib.rs
+++ b/components/dom_struct/lib.rs
@@ -46,6 +46,14 @@ pub fn dom_struct(args: TokenStream, input: TokenStream) -> TokenStream {
                         &self.#ident
                     }
                 }
+
+                impl Eq for #name {}
+
+                impl PartialEq for #name {
+                    fn eq(&self, other: &Self) -> bool {
+                        crate::DomObject::reflector(self) == crate::DomObject::reflector(other)
+                    }
+                }
             )
             .into()
         } else {
--- a/components/domobject_derive/lib.rs
+++ b/components/domobject_derive/lib.rs
@@ -57,6 +57,14 @@ fn expand_dom_object(input: syn::DeriveInput) -> proc_macro2::TokenStream {
                 self.#first_field_name.init_reflector(obj);
             }
         }
+
+        impl #impl_generics Eq for #name #ty_generics #where_clause {}
+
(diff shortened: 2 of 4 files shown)


components/dom_struct/lib.rs
```diff
@@ -46,6 +46,14 @@ pub fn dom_struct(args: TokenStream, input: TokenStream) -> TokenStream {
                         &self.#ident
                     }
                 }
+
+                impl Eq for #name {}
+
+                impl PartialEq for #name {
+                    fn eq(&self, other: &Self) -> bool {
+                        self.reflector() == other.reflector()
+                    }
```
sagudev, line 55: can we just import DomObject here, so other files don't need to import it?

sagudev: also I think it would be more appropriate to do this in derive(DomObject), see https://github.com/servo/servo/blob/main/components/domobject_derive/lib.rs

Taym95: I could not import `DomObject` because it created circular dependencies

sagudev: How? proc_macro does not need actual `DomObject`, it just needs to emit `use crate::DomObject` (here crate actually refers to script crate, that calls this proc macro). This should work:

```suggestion
                    fn eq(&self, other: &Self) -> bool {
                        use crate::DomObject;
                        self.reflector() == other.reflector()
                    }
```

or we could use [fully qualified syntax](https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name):
```suggestion
                    fn eq(&self, other: &Self) -> bool {
                        crate::DomObject::reflector(self) == crate::DomObject::reflector(other)
                    }
```

Taym95: Indeed, thanks!

components/dom_struct/lib.rs
```diff
@@ -46,6 +46,14 @@ pub fn dom_struct(args: TokenStream, input: TokenStream) -> TokenStream {
...
                     }
                 }
+
+                impl Eq for #name {}
+
+                impl PartialEq for #name {
+                    fn eq(&self, other: &Self) -> bool {
+                        crate::DomObject::reflector(self) == crate::DomObject::reflector(other)
+                    }
+                }
```
sagudev, line 56: I think this should be done in domobject_derive, somewhere around here: https://github.com/servo/servo/blob/590527176e7c03648fcce797604d3e12da0c2c99/components/domobject_derive/lib.rs#L48

components/domobject_derive/lib.rs
```diff
@@ -57,6 +57,14 @@ fn expand_dom_object(input: syn::DeriveInput) -> proc_macro2::TokenStream {
                 self.#first_field_name.init_reflector(obj);
             }
         }
+
+        impl #impl_generics Eq for #name #ty_generics #where_clause {}
```
Taym95, line 61: @sagudev sorry for delay, I moved them to domobject_derive ready for review again.

sagudev (approved):

PR #33219, opened 2024-08-28
Add simple fling implementation

jschwe:
Add a simple fling implementation, which depends on a refresh tick from the embedder.
Currently this refresh tick is only implemented for OpenHarmony (using the vertical synchronization signal).
The fling implementation is very simple and will probably be improved in future PRs.

Current limitations:

- flinging code assumes that framerate is constant. The framerate could change at any time (e.g. user enables energy saving mode), but since APIs to access the current framerate are OS dependent, we would need to discuss how to access such APIs. Calculating based on time is also possible, but seems like it would lead to unwanted variations in the fling speed due to measuring inaccuracies.
- I didn't look at existing fling implementations, so probably better algorithms exist (e.g. take into account acceleration / flick).
- The constants used when calculating fling velocity should be configurable
- Maybe the embedder wants to be informed about fling starting / ending (for decisions about dynamic framerate)

Based on https://github.com/servo/servo/pull/33117

---

### DIFF
--- a/components/compositing/touch.rs
+++ b/components/compositing/touch.rs
@@ -10,7 +10,6 @@ use webrender_api::units::{DeviceIntPoint, LayoutVector2D};
 
 use self::TouchState::*;
 
-
 // TODO: All `_SCREEN_PX` units below are currently actually used as `DevicePixel`
 // without multiplying with the `hidpi_factor`. This should be fixed and the
 // constants adjusted accordingly.
@@ -114,10 +113,11 @@ impl TouchHandler {
         let Flinging {
             velocity,
             ref cursor,
-        } = &mut self.state else {
+        } = &mut self.state
+        else {
             return None;
         };
-        if velocity.length().abs() < FLING_MIN_SCREEN_PX  {
+        if velocity.length().abs() < FLING_MIN_SCREEN_PX {
             self.state = Nothing;
             return None;
         }
@@ -129,7 +129,6 @@ impl TouchHandler {
             delta: LayoutVector2D::new(velocity.x, velocity.y),
             cursor: *cursor,
         })
-
     }
 
     pub fn on_touch_move(&mut self, id: TouchId, point: Point2D<f32, DevicePixel>) -> TouchAction {
@@ -228,7 +227,7 @@ impl TouchHandler {
             Panning { .. } => {
                 self.state = Nothing;
                 TouchAction::NoAction
-            }
+            },
             Pinching => {
                 self.state = Panning {
                     velocity: Vector2D::new(0.0, 0.0),
--- a/components/compositing/touch.rs
+++ b/components/compositing/touch.rs
@@ -15,11 +15,11 @@ use self::TouchState::*;
 // constants adjusted accordingly.
 /// Minimum number of `DeviceIndependentPixel` to begin touch scrolling.
 const TOUCH_PAN_MIN_SCREEN_PX: f32 = 20.0;
-/// Factor by which the flinging velocity changes on each tick
+/// Factor by which the flinging velocity changes on each tick.
 const FLING_SCALING_FACTOR: f32 = 0.95;
-/// Minimum velocity required for transitioning to fling when panning ends
+/// Minimum velocity required for transitioning to fling when panning ends.
 const FLING_MIN_SCREEN_PX: f32 = 3.0;
-/// Maximum velocity when flinging
+/// Maximum velocity when flinging.
 const FLING_MAX_SCREEN_PX: f32 = 4000.0;
 
 pub struct TouchHandler {
--- a/components/compositing/touch.rs
+++ b/components/compositing/touch.rs
@@ -10,14 +10,18 @@ use webrender_api::units::{DeviceIntPoint, LayoutVector2D};
 
 use self::TouchState::*;
 
+
+// TODO: All `_SCREEN_PX` units below are currently actually used as `DevicePixel`
+// without multiplying with the `hidpi_factor`. This should be fixed and the
+// constants adjusted accordingly.
 /// Minimum number of `DeviceIndependentPixel` to begin touch scrolling.
 const TOUCH_PAN_MIN_SCREEN_PX: f32 = 20.0;
 /// Factor by which the flinging velocity changes on each tick
 const FLING_SCALING_FACTOR: f32 = 0.95;
 /// Minimum velocity required for transitioning to fling when panning ends
-const FLING_MIN: f32 = 3.0;
+const FLING_MIN_SCREEN_PX: f32 = 3.0;
 /// Maximum velocity when flinging
-const FLING_MAX: f32 = 4000.0;
+const FLING_MAX_SCREEN_PX: f32 = 4000.0;
 
 pub struct TouchHandler {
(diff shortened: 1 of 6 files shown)


mrobinson:
@jschwe Is this still a draft or are you looking for a review?

jschwe:
Hi @mrobinson . I believe the background why it's still a draft is that we had a call a while back and agreed it would be best to first add a generic refresh driver (fallback) for all platforms.

jschwe:
@mrobinson I updated this PR to include 2D directional fling support and made `Vsync` an EmbedderEvent, which made the changes  less invasive.

components/compositing/touch.rs
```diff
@@ -139,30 +187,53 @@ impl TouchHandler {
...
                 TouchAction::Click
             },
-            Nothing | Panning => {
+            Nothing => {
                 self.state = Nothing;
                 TouchAction::NoAction
             },
+            Panning { velocity } => {
+                if velocity.length().abs() >= FLING_MINIMUM {
+                    // Todo: point != old. Not sure which one is better to take as cursor for flinging.
```
mrobinson, line 211: Nit:

```suggestion
                    // TODO: point != old. Not sure which one is better to take as cursor for flinging.
```

components/compositing/touch.rs
```diff
@@ -3,14 +3,18 @@
...
 use style_traits::DevicePixel;
+use webrender_api::units::{DeviceIntPoint, LayoutVector2D};
 
 use self::TouchState::*;
 
 /// Minimum number of `DeviceIndependentPixel` to begin touch scrolling.
 const TOUCH_PAN_MIN_SCREEN_PX: f32 = 20.0;
+const FLING_SCALING_FACTOR: f32 = 0.95;
+const FLING_MINIMUM: f32 = 3.0;
+const FLING_MAX: f32 = 4000.0;
```
mrobinson, line 17: Do you mind adding documentation for these constants and making them consistent with `TOUCH_PAN_MIN_SCREEN_PX`?

components/compositing/touch.rs
```diff
@@ -45,7 +47,15 @@ pub enum TouchState {
     /// Contains the initial touch location.
     Touching,
     /// A single touch point is active and has started panning.
-    Panning,
+    Panning {
+        velocity: Vector2D<f32, DevicePixel>,
+    },
+    /// No active touch points, but there is still scrolling velocity
+    /// Note: For now we only support flinging along the y-axis
```
mrobinson, line 54: I think this comment can be removed now?

```suggestion
    /// Note: For now we only support flinging along the y-axis
```

components/compositing/touch.rs
```diff
@@ -81,13 +96,38 @@ impl TouchHandler {
...
             DefaultPrevented => DefaultPrevented,
             Pinching | MultiTouch => MultiTouch,
         };
     }
 
+    pub fn on_vsync(&mut self) -> Option<FlingAction> {
+        if let Flinging {
+            velocity,
+            ref cursor,
+        } = &mut self.state
```
mrobinson, line 111: It think an early return would be a bit nicer here to avoid so much horizontal identation:

```suggestion
        if let Flinging {
            velocity,
            ref cursor,
        } = &mut self.state else {
            return None;
        }
```

components/compositing/touch.rs
```diff
@@ -81,13 +96,38 @@ impl TouchHandler {
...
+                // or save a timestamp to account for a potentially changing display refresh rate.
+                *velocity *= FLING_SCALING_FACTOR;
+                debug_assert!(velocity.length() <= FLING_MAX && velocity.length() >= -FLING_MAX);
+                Some(FlingAction {
+                    delta: LayoutVector2D::new(velocity.x, velocity.y),
+                    cursor: *cursor,
+                })
+            } else {
+                self.state = Nothing;
+                None
```
mrobinson, line 124: This could also be an early return  and then we save all the meaty logic for the main body of the function.

components/compositing/touch.rs
```diff
@@ -105,16 +145,24 @@ impl TouchHandler {
...
+                    };
                     TouchAction::Scroll(delta)
                 } else {
                     TouchAction::NoAction
                 }
             },
-            Panning => {
+            Panning { ref mut velocity } => {
                 let delta = point - old_point;
+                // Todo: Probably we should track 1-3 more points and use a smarter algorithm
```
mrobinson, line 158: Nit:

```suggestion
                // TODO: Probably we should track 1-3 more points and use a smarter algorithm
```

components/compositing/touch.rs
```diff
@@ -139,30 +187,53 @@ impl TouchHandler {
...
                 self.state = Nothing;
                 TouchAction::Click
             },
-            Nothing | Panning => {
+            Nothing => {
                 self.state = Nothing;
                 TouchAction::NoAction
             },
+            Panning { velocity } => {
+                if velocity.length().abs() >= FLING_MINIMUM {
```
mrobinson, line 210:
```suggestion
            Panning { velocity } if velocity.length().abs() >= FLING_MINIMUM => {
                ....
             }
             Panning { .. } => {
                 self.state = Nothing;
                 TouchAction::NoAction
             }
```

mrobinson (approved):
Nice! Thanks for the improvements. This is very cool and it will be nice to extend it to Android.

I just have a few nits, but otherwise feel free to send it to the MQ when they are addressed.

components/compositing/touch.rs
jschwe: I added documentation and renamed `FLING_MINIMUM` to `FLING_MIN_SCREEN_PX` (not sure if that was what you had in mind). 
I also added a new todo here, since the constants are actually used with the DevicePixel unit, and would need to be multiplied with the `hidpi_factor` (but the existing touch pan constant has the same issue).

components/compositing/touch.rs
jschwe: yeah, thats much nicer

components/compositing/touch.rs
```diff
@@ -3,14 +3,24 @@
...
+// without multiplying with the `hidpi_factor`. This should be fixed and the
+// constants adjusted accordingly.
 /// Minimum number of `DeviceIndependentPixel` to begin touch scrolling.
 const TOUCH_PAN_MIN_SCREEN_PX: f32 = 20.0;
+/// Factor by which the flinging velocity changes on each tick
+const FLING_SCALING_FACTOR: f32 = 0.95;
+/// Minimum velocity required for transitioning to fling when panning ends
+const FLING_MIN_SCREEN_PX: f32 = 3.0;
+/// Maximum velocity when flinging
+const FLING_MAX_SCREEN_PX: f32 = 4000.0;
```
mrobinson, line 23:
```suggestion
/// Factor by which the flinging velocity changes on each tick.
const FLING_SCALING_FACTOR: f32 = 0.95;
/// Minimum velocity required for transitioning to fling when panning ends.
const FLING_MIN_SCREEN_PX: f32 = 3.0;
/// Maximum velocity when flinging.
const FLING_MAX_SCREEN_PX: f32 = 4000.0;
```

mrobinson (approved):

PR #33222, opened 2024-08-28
compositor: Do not parse the Cargo.lock file while building

mrobinson:
The compositor's `build.rs` script was parsing the `Cargo.lock` file in
order to tag WebRender captures with the WebRender version. The embedder
already knows what version of Servo we are using, which should be enough
to infer the WebRender revision. This changes does that and generally
does a bit of cleaning up of how captures are done.

- The name of the capture directory is now `webrender-captures`
- There is console output now when captures are done. Before it was hard
  to know if it succeeded.
- Simplify the Compositor constructor a little to avoid passing
  arguments so much.

---

- [x] These changes do not require tests because this just modifies a bit how WebRender captures are done.


mukilan (approved):
Looks good to me. I wonder if at some point we might want to allow the embedder to also allow specifying the base directory for the capture path, in addition to the version.

PR #33225, opened 2024-08-28
Pass keyboard events to WebView on Windows

crbrz:
Send keyboard events to focused `WebView` for key events not handled by `WebViewManager`.

---

- [X] These changes do not require tests because tests aren't run for Windows yet.

### DIFF
--- a/ports/servoshell/desktop/webview.rs
+++ b/ports/servoshell/desktop/webview.rs
@@ -533,7 +533,11 @@ where
     }
 
     #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


jdm:
Hmm, this fix is obviously good, but I'd like to find a way to restructure this code so that only platform-specific logic is present, to avoid similar bugs in the future.

crbrz:
I think `WebViewManager::platform_handle_key` can be removed and use `ShortcutMatcher::optional_shortcut` instead when a condition is needed.

jdm (approved):
This restructuring seems like a reasonable compromise. Thanks!

PR #33228, opened 2024-08-28
ci: Fix the WPT export job after the repository change

mrobinson:
The GitHub search API is a bit sensitive. There isn't a great way to
search for the repository organization and the branch name when looking
for open PRs. Instead use the bot username as the author name, which
should likely have been having before. This fixes the WPT export job.

---

### DIFF
--- a/python/wpt/exporter/github.py
+++ b/python/wpt/exporter/github.py
@@ -72,7 +72,9 @@ class GithubRepository:
         return GithubBranch(self, name)
 
     def get_open_pull_request_for_branch(
-        self, branch: GithubBranch
+        self,
+        github_username: str,
+        branch: GithubBranch
     ) -> Optional[PullRequest]:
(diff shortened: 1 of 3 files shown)


python/wpt/exporter/github.py
```diff
@@ -72,7 +72,9 @@ def get_branch(self, name: str) -> GithubBranch:
         return GithubBranch(self, name)
 
     def get_open_pull_request_for_branch(
-        self, branch: GithubBranch
+        self,
+        github_username: str,
```
Loirooriol, line 76: I don't get it, this parameter isn't used anywhere?

mrobinson: This was lost after a bit of debugging, but I've restored the usage now.

Loirooriol (approved):

PR #33244, opened 2024-08-29
servoshell: Add close buttons and increase interactivity of tabs

Melchizedek6809:
Hey, I've tried to improve the UI of the new tab bar a bit, mainly by adding a close button for each tab as well as another button for opening new tabs, also changed the styling so it looks more like other browsers. One can now click with the middle mouse button on a tab to close it.

Screenshot 2024-08-29 at 13 16 12

- [X] These changes do not require tests because they are for the minibrowser UI

### DIFF
--- a/ports/servoshell/desktop/minibrowser.rs
+++ b/ports/servoshell/desktop/minibrowser.rs
@@ -176,6 +176,7 @@ impl Minibrowser {
         webview_id: TopLevelBrowsingContextId,
     ) -> Option<EmbedderEvent> {
         let old_item_spacing = ui.spacing().item_spacing;
+        let old_visuals = ui.visuals().clone();
         ui.spacing_mut().item_spacing = egui::vec2(0.0, 0.0);
 
         let visuals = ui.visuals_mut();
@@ -233,6 +234,7 @@ impl Minibrowser {
 
         ui.spacing_mut().item_spacing = old_item_spacing;
         let close_button = ui.add(egui::Button::new("X").fill(fill_color));
+        *ui.visuals_mut() = old_visuals;
         if close_button.clicked() || close_button.middle_clicked() || tab.middle_clicked() {
             Some(EmbedderEvent::CloseWebView(webview_id))
         } else if !selected && tab.clicked() {
--- a/ports/servoshell/desktop/minibrowser.rs
+++ b/ports/servoshell/desktop/minibrowser.rs
@@ -19,12 +19,14 @@ use euclid::{Box2D, Length, Point2D, Scale, Size2D};
 use gleam::gl;
 use glow::NativeFramebuffer;
 use log::{trace, warn};
+use servo::base::id::WebViewId;
 use servo::compositing::windowing::EmbedderEvent;
 use servo::script_traits::TraversalDirection;
 use servo::servo_geometry::DeviceIndependentPixel;
 use servo::servo_url::ServoUrl;
 use servo::style_traits::DevicePixel;
 use servo::webrender_traits::RenderingContext;
+use servo::TopLevelBrowsingContextId;
 use winit::event::{ElementState, MouseButton};
 
 use super::egui_glue::EguiGlow;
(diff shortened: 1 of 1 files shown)


Melchizedek6809:
Just wondering whether there's anything that needs to be changed? Would love to see this merged

ports/servoshell/desktop/minibrowser.rs
```diff
@@ -267,30 +348,27 @@ impl Minibrowser {
...
                     |ui| {
                         for (webview_id, webview) in webviews.webviews().into_iter() {
-                            let msg = match (&webview.title, &webview.url) {
-                                (Some(title), _) if !title.is_empty() => title.clone(),
-                                (_, Some(url)) => url.to_string(),
-                                _ => "New Tab".to_owned(),
+                            let label = match (&webview.title, &webview.url) {
+                                (Some(title), _) => title,
+                                (None, Some(url)) => &url.to_string(),
+                                _ => "New Tab",
```
mrobinson, line 361: This could avoid the match, I think:

```suggestion
                            let label = webview.title
                                .or(webview.url.map(ServoUrl::to_string))
                                .as_ref()
                                .unwrap_or_else("New Tab");
```

ports/servoshell/desktop/minibrowser.rs
```diff
@@ -163,6 +166,84 @@ impl Minibrowser {
...
+        let old_visuals = ui.visuals().clone();
+        ui.spacing_mut().item_spacing = egui::vec2(0.0, 0.0);
+
+        let visuals = ui.visuals_mut();
+        // Remove the stroke so we don't see the border between the close button and the label
+        visuals.widgets.active.bg_stroke.width = 0.0;
+        visuals.widgets.hovered.bg_stroke.width = 0.0;
+        // Now we make sure the fill color is always the same, irrespective of state, that way
+        // we can make sure that both the label and close button have the same background color
+        visuals.widgets.noninteractive.weak_bg_fill = egui::Color32::from_gray(40);
```
mrobinson, line 188: Will this work for both light and dark themes? Is there some kind of color that works for both?

mrobinson (changes requested):
Nice work. I have a couple small comments but I don't see any big issues with this change.

ports/servoshell/desktop/minibrowser.rs
Melchizedek6809: Sadly that doesn't compile, the ownership is a bit tricky here since we need a &str, but we have to create the underlying String from the ServoUrl first. I tried rewriting it a bit but the match seems like the simplest solution so far.

ports/servoshell/desktop/minibrowser.rs
Melchizedek6809: Yeah, probably wouldn't. I've changed the code so that the colors are taken from the theme, that way it should work with most themes.

mrobinson (approved):

ports/servoshell/desktop/minibrowser.rs
```diff
@@ -267,30 +351,27 @@ impl Minibrowser {
...
-            // TODO: Add a way to close a tab eg. with a [x] control.
+            // A simple Tab header strip
             TopBottomPanel::top("tabs").show(ctx, |ui| {
                 ui.allocate_ui_with_layout(
                     ui.available_size(),
                     egui::Layout::left_to_right(egui::Align::Center),
                     |ui| {
                         for (webview_id, webview) in webviews.webviews().into_iter() {
-                            let msg = match (&webview.title, &webview.url) {
-                                (Some(title), _) if !title.is_empty() => title.clone(),
```
Wuelle, line 279: Is there a specific reason that the `if !title.is_empty()` guard was removed?
This is something I specifically added in #33229 because empty tab titles look wonky (:

Melchizedek6809: I suppose they got lost during rebasing since we were both working in the same area.

PR #33246, opened 2024-08-29
Bump xi_unicode to latest version

jschwe:
There is no changelog available, but we only use `linebreak_property` and `LineBreakLeafiter` from `xi_unicode`, so it should be safe to update,

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3900,7 +3900,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -7922,7 +7922,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


mukilan (approved):
Thanks!

mukilan:
It seems like the two test failures are due to breaking change in xi-unicode. The linebreak_property function now returns 42 instead of 40 for the codepoint 'u200D'

https://github.com/xi-editor/xi-editor/commit/0c82f2eb8c508204cf2e7560047b8e060361eaed#diff-c1bc7323d4980a7cfbd7700f185ea3a855e998af62e7cdd43e5d9985080e96e8R471

mukilan:
It looks like just a bug in xi-unicode where the numeric value of the XI_LINE_BREAKING_CLASS_ZWJ constant has not been updated to be 42 instead of 40. The `[U_LB_ZWJ](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h.html#a5d1abdf05be22cb9599f804a8506277cacf051a491922f822f16cfc66efda7f3d)` constant also has a [numeric value of 42](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h_source.html#l02428)

mrobinson:
> It looks like just ~a bug in xi-unicode~ where the numeric value of the XI_LINE_BREAKING_CLASS_ZWJ constant has not been updated to be 42 instead of 40. The [U_LB_ZWJ](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h.html#a5d1abdf05be22cb9599f804a8506277cacf051a491922f822f16cfc66efda7f3d) constant also has a [numeric value of 42](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uchar_8h_source.html#l02428)
> 
> edit: XI_LINE_BREAKING_CLASS_ZWJ is defined in servo, not xi-unicode so we can update it in this PR.

Nice investigation. We should really stop using these undocumented xi-unicode classes.

PR #33248, opened 2024-08-29
Fix automatic minimum size for column flexbox

Loirooriol:
`main_content_size_info()` was always assigning the main-axis automatic minimum size into the inline axis. But in a column flexbox, the main axis corresponds to the block axis.

---


mrobinson (approved):
Nice catch!

PR #33258, opened 2024-08-30
script: Stop using legacy `time` for `Document::reflow_timeout`

mrobinson:
Use `std::time` from the Rust standard library instead. This is one step
toward removing our use of the legacy `time` crate which has security
issues.

---

### DIFF
--- a/components/script/dom/htmlbodyelement.rs
+++ b/components/script/dom/htmlbodyelement.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::time::Duration;
+
 use dom_struct::dom_struct;
 use embedder_traits::EmbedderMsg;
 use html5ever::{local_name, namespace_url, ns, LocalName, Prefix};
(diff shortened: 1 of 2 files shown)


components/script/dom/htmlbodyelement.rs
```diff
@@ -23,9 +25,8 @@ use crate::dom::htmlelement::HTMLElement;
 use crate::dom::node::{document_from_node, window_from_node, BindContext, Node};
 use crate::dom::virtualmethods::VirtualMethods;
 
-/// How long we should wait before performing the initial reflow after `<body>` is parsed, in
-/// nanoseconds.
-const INITIAL_REFLOW_DELAY: u64 = 200_000_000;
+/// How long we should wait before performing the initial reflow after `<body>` is parsed.
+const INITIAL_REFLOW_DELAY: Duration = Duration::from_nanos(200_000_000);
```
mukilan, line 29: Can we change this to `Duration::from_millis(200)` to make it even more explicit?

mrobinson: Sure. I've done that now.

mukilan (approved):

PR #33259, opened 2024-08-30
net: Stop using legacy time in the HTTP and CORS caches

mrobinson:
This is part of switching away from using a very old version of `time`.

---

- [x] These changes are part of ##30150.

### DIFF
--- a/components/net/http_cache.rs
+++ b/components/net/http_cache.rs
@@ -11,7 +11,7 @@ use std::collections::HashMap;
 use std::ops::Bound;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
-use std::time::SystemTime;
+use std::time::{Duration, Instant, SystemTime};
 
 use headers::{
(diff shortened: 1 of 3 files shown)


components/net/http_cache.rs
```diff
@@ -75,7 +74,8 @@ struct MeasurableCachedResource {
     raw_status: Option<(u16, Vec<u8>)>,
     url_list: Vec<ServoUrl>,
     expires: Duration,
-    last_validated: Tm,
+    #[ignore_malloc_size_of = "Platform-specific Instant object isn't measured"]
```
mukilan, line 77: Do we need the explicit ignore_malloc_size_of here? It seems like [there is an implementation for the Instant type](https://github.com/servo/stylo/blob/65f8d1316a0966401bcfb9fa7dd5e3659a1605b2/malloc_size_of/lib.rs#L957).

mukilan (approved):
Nice!

components/net/http_cache.rs
mrobinson: Nope. I've removed this.

PR #33260, opened 2024-08-30
net: Stop using both versions of the `time` crate in the cookie code

mrobinson:
`std::time` is good enough for us here. `cookie` is using `time 0.3`,
but Servo can convert to standard library types when getting data from
`cookie`. This reduces our direct dependencies and removes more use of
the very old `time 0.1` series.

---


mrobinson:
I think this is ready to review. The CI failure looks like infrastructure issues.

mukilan (approved):

PR #33261, opened 2024-08-30
CI: linux-perf job for benchmarks

sagudev:
Results are uploaded to https://bencher.dev, but we can do some aggregation dashboard later (something like https://arewefastyet.com/win10/benchmarks/overview?numDays=60)

Before landing we need to create bencher account and add `BENCHER_API_TOKEN` to secrets.

Relevant chat on zulip: https://servo.zulipchat.com/#narrow/stream/416779-Performance/topic/report.20benchmark.20results

---

- [x] These changes do not require tests because they are for running benching tests

### DIFF
--- a/.github/workflows/linux-bencher.yml
+++ b/.github/workflows/linux-bencher.yml
@@ -1,0 +1,85 @@
+name: Linux WPT Tests
+on:
+  workflow_call:
+
+permissions:
+  checks: write
+  pull-requests: write
+
+env:
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  # allows overriding bencher project for pushes
+  BENCHER_PROJECT: ${{ vars.BENCHER_PROJECT || 'servo' }}
+
+jobs:
+  linux-bencher:
+    name: Bencher
+    # keep this in sync with testbed
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        if: github.event_name != 'pull_request_target'
+        with:
+          fetch-depth: 0
+      # This is necessary to checkout the pull request if this run was triggered via a
+      # `pull_request_target` event.
+      - uses: actions/checkout@v4
+        if: github.event_name == 'pull_request_target'
+        with:
+          ref: refs/pull/${{ github.event.number }}/head
+          fetch-depth: 0
+      - uses: actions/download-artifact@v4
+        with:
+          name: release-binary-linux
+          path: release-binary-linux
+      - name: unPackage binary
+        run: tar -xzf release-binary-linux/target.tar.gz
+      - name: Bootstrap dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          sudo apt update
+          sudo apt install -qy --no-install-recommends mesa-vulkan-drivers
+          python3 ./mach bootstrap --skip-lints
+      - uses: bencherdev/bencher@main
+      - name: File size
+        run: ./etc/ci/bencher.py filesize target/release/servo --bmf-output size.json
+      - name: Speedometer
+        run: |
+            python3 ./mach test-speedometer -r --bmf-output speedometer.json
+      - name: Dromaeo
+        run: |
+            python3 ./mach test-dromaeo -r dom --bmf-output dromaeo.json
+      # set options
+      - name: Set bencher opts for PRs (label try run)
+        if: github.event_name == 'pull_request_target'
+        run: |
+            echo "RUN_BENCHER_OPTIONS=--branch ${{ github.event.number }}/PR \
+            --branch-start-point ${{ github.base_ref }} \
+            --branch-start-point-hash ${{ github.event.pull_request.base.sha }} \
+            --branch-reset \
+            --github-actions ${{ secrets.GITHUB_TOKEN }}" >> "$GITHUB_ENV"
+      - name: Set bencher opts for main
+        if: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+        run: |
+            echo "RUN_BENCHER_OPTIONS=--branch main \
+            --github-actions ${{ secrets.GITHUB_TOKEN }}" >> "$GITHUB_ENV"
+      - name: Set bencher opts for try branch
+        if: ${{ github.event_name == 'push' && github.ref_name == 'try' }}
+        run: |
+            git remote add upstream https://github.com/servo/servo
+            git fetch upstream main
+            echo "RUN_BENCHER_OPTIONS=--branch try \
+            --github-actions ${{ secrets.GITHUB_TOKEN }} \
+            --hash $(git rev-parse HEAD~1) \
+            --branch-start-point main \
+            --branch-start-point-hash $(git merge-base upstream/main HEAD) \
+            --branch-reset" >> "$GITHUB_ENV"
+      # we join results and send all data once to have it all in one report
+      - name: Send results
+        continue-on-error: true
+        run: |
+          ./etc/ci/bencher.py merge size.json speedometer.json dromaeo.json --bmf-output b.json
+          bencher run --adapter json --file b.json \
+            --project ${{ env.BENCHER_PROJECT }} --token ${{ secrets.BENCHER_API_TOKEN }} --testbed ubuntu-22.04 \
+            $RUN_BENCHER_OPTIONS
--- a/etc/ci/bencher.py
+++ b/etc/ci/bencher.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+# Copyright 2024 The Servo Project Developers. See the COPYRIGHT
+# file at the top-level directory of this distribution.
+#
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+import argparse
+import json
+import os
+
+
+def size(args):
+    size = os.path.getsize(args.binary)
+    print(size)
+    with open(args.bmf_output, 'w', encoding='utf-8') as f:
+        json.dump({
+            'servo': {
+                'file-size': {
+                    'value': float(size),
+                }
+            }
+        }, f, indent=4)
+
+
+def merge(args):
+    output: dict[str, object] = dict()
+    for input_file in args.inputs:
+        with open(input_file, 'r', encoding='utf-8') as f:
+            data = json.load(f)
+            diff = set(data) & set(output)
+            if diff:
+                print("Duplicated keys:", diff)
+            output = data | output
+
+    with open(args.bmf_output, 'w', encoding='utf-8') as f:
+        json.dump(output, f, indent=4)
+
+
+parser = argparse.ArgumentParser("Helper commands for bencher")
+
+subparser = parser.add_subparsers()
+size_parser = subparser.add_parser("filesize", help="Returns BMF for filesize")
+size_parser.add_argument("binary", help="Servo binary file")
+size_parser.add_argument("--bmf-output", help="BMF JSON output file", default=None)
+size_parser.set_defaults(func=size)
+
+merge_parser = subparser.add_parser("merge", help="Merges BMF JSONs")
+merge_parser.add_argument("--bmf-output", help="BMF JSON output file")
+merge_parser.add_argument("inputs", help="BMF JSON files to merge", nargs="+")
+merge_parser.set_defaults(func=merge)
+
+args = parser.parse_args()
+args.func(args)
--- a/.github/workflows/dispatch-workflow.yml
+++ b/.github/workflows/dispatch-workflow.yml
@@ -17,6 +17,9 @@ on:
       unit-tests:
         required: true
         type: boolean
+      bencher:
+        required: true
+        type: boolean
 
 jobs:
   win:
@@ -49,6 +52,7 @@ jobs:
       wpt-layout: ${{ inputs.wpt-layout }}
       unit-tests: ${{ inputs.unit-tests }}
       wpt-args: ${{ inputs.wpt-args }}
+      bencher: ${{ inputs.bencher }}
 
   lint:
     if: ${{ inputs.workflow == 'lint' }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -32,6 +32,10 @@ on:
         required: false
         type: boolean
         default: false
+      bencher:
+        required: false
+        default: false
+        type: boolean
   workflow_dispatch:
     inputs:
       profile:
@@ -63,6 +67,10 @@ on:
         required: false
         type: boolean
         default: false
+      bencher:
+        required: false
+        default: false
+        type: boolean
 
 env:
   RUST_BACKTRACE: 1
@@ -218,4 +226,11 @@ jobs:
       profile: ${{ inputs.profile }}
       wpt-sync-from-upstream: ${{ inputs.wpt-sync-from-upstream }}
       wpt-layout: "layout-2013"
+    secrets: inherit
+
+  bencher:
+    needs: ["build"]
+    # benches must be release (we will do benches for production profile in servo/servo-nightly-builds)
+    if: ${{ inputs.bencher && inputs.profile == 'release' && github.event_name != 'workflow_dispatch' && github.event_name != 'merge_queue' }}
+    uses: ./.github/workflows/linux-bencher.yml
     secrets: inherit
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,6 +54,7 @@ jobs:
       profile: ${{ matrix.profile }}
       unit-tests: ${{ matrix.unit_tests }}
       wpt-args: ${{ matrix.wpt_args }}
+      bencher: ${{ matrix.bencher }}
 
   build-result:
     name: Result
--- a/.github/workflows/try-label.yml
+++ b/.github/workflows/try-label.yml
@@ -137,6 +137,7 @@ jobs:
       profile: ${{ matrix.profile }}
       unit-tests: ${{ matrix.unit_tests }}
       wpt-args: ${{ matrix.wpt_args }}
+      bencher: ${{ matrix.bencher }}
 
   results:
     name: Results
--- a/.github/workflows/try.yml
+++ b/.github/workflows/try.yml
@@ -21,6 +21,9 @@ on:
       unit-tests:
         required: false
         type: boolean
+      bencher:
+        required: false
+        type: boolean
 
 jobs:
   decision:
@@ -104,6 +107,7 @@ jobs:
       profile: ${{ matrix.profile }}
       unit-tests: ${{ matrix.unit_tests }}
       wpt-args: ${{ matrix.wpt_args }}
+      bencher: ${{ matrix.bencher }}
 
   build-result:
     name: Result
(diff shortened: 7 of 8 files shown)


dklassic:
A copy pasta from [zulip](https://servo.zulipchat.com/#narrow/channel/416779-Performance/topic/Benchmark.20infrastructure.20related.20issues/near/479804395) to share my current findings when testing this PR:

- Currently `servospedometer` test took around 80 seconds, so the timeout time need to be increased to maybe 100 seconds or above.
- Linux WPT profile [does not have bencher enabled](https://github.com/dklassic/servo/commit/1d99033af6a2458fd8c8bbab385cfad1c564da4d) which means `push: try in servo/servo` and `push: try in personal fork (when tokens are not set)` currently all skip `bencher` workflow.
- `main.yml` does not run `linux.yml` with `bencher: true` so bencher will not run in `push: main (PR landed)`.
- Write permission for `checks` need to be added to `bencher.yml` so GitHub Checks can be created properly by bencher (per testing for that bencher patch).
- Write permission for `pull-request` might need to be added to `bencher.yml` otherwise PR comment cannot be created by bencher (per [this attempt](https://github.com/dklassic/servo/actions/runs/11604701194/job/32314404265#step:14:968)). I'm actually not quite sure about this one because the next issue is slightly tangled with testing this one.
- Currently on a PR try run attempt it'll result in a [404 error when creating bencher report](https://github.com/dklassic/servo/actions/runs/11605715720/job/32316814144#step:14:842), I don't really understand what this error means so I'm not sure if it is an error from bencher, my bencher project setup, or somewhere else.

I'll circle back and try it cleanly again just in case.

sagudev:
> Linux WPT profile [does not have bencher enabled](https://github.com/dklassic/servo/commit/1d99033af6a2458fd8c8bbab385cfad1c564da4d) which means push: try in servo/servo and push: try in personal fork (when tokens are not set) currently all skip bencher workflow.

This is expected as perf job needs to be specified separately like `mach try perf` (will only run build then bencher job), but `mach try wpt perf` should do both WPT and perf (using same build job). I updated gist with `mach try perf`.

> Currently on a PR try run attempt it'll result in a [404 error when creating bencher report](https://github.com/dklassic/servo/actions/runs/11605715720/job/32316814144#step:14:842), I don't really understand what this error means so I'm not sure if it is an error from bencher, my bencher project setup, or somewhere else.

Hm, this looks like error from bencher server.

I will do rebase of this PR and incorporate some fixes.

.github/workflows/linux-bencher.yml
```diff
@@ -0,0 +1,83 @@
...
+
+permissions:
+  checks: write
+  pull-requests: write
+
+env:
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  # allows overriding bencher project for pushes
+  BENCHER_PROJECT: ${{ vars.BENCHER_PROJECT || 'servo' }}
```
sagudev, line 13: @dklassic use this to override `--project` parameter in your repo.

dklassic:
Good news, I've done more extensive testing and basically everything is working properly except one that I need some extra setup:
- [`push: main (PR landed)`](https://github.com/dklassic/servo/actions/runs/11641055840/job/32419449487)
- [`push: try in servo/servo`](https://github.com/dklassic/servo/actions/runs/11641073758/job/32419369661)
- [`label: try-*`](https://github.com/dklassic/servo/actions/runs/11623305915/job/32370551630)

That leaves only `push: try in personal fork (when tokens are not set)` to be tested. I'll have it tested soon.

sagudev:
https://github.com/dklassic/servo/pull/4#issuecomment-2451206394 has wierd reporting (not everything should be under same unit) and indeed it's not when checking up [full report](https://bencher.dev/perf/dklassic-servo/reports/c226feba-906c-48e4-9b05-8328274705f8?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dklassic-servo). This is bencher bug.

For push try run I do not see any reports (checks). Could you post link to it?

And for the last thing it's important to see if comparisons work, by creating push to main to create baseline report and then do try and PR to see comparison reports (these are actually cruical for try/PR runs), something like this: https://github.com/rustls/rustls/pull/2192#issuecomment-2451747582.

dklassic:
> For push try run I do not see any reports (checks). Could you post link to it?

You're right that there are no GitHub Checks, I just saw the workflow had a success run and assumed that it is working. The results are only properly uploaded to bencher but no Checks were made, weird.

dklassic:
https://github.com/dklassic/servo/actions/runs/11641073758/job/32419369661#step:14:11

Looking into the `RUN_BENCHER_OPTIONS` used, seems like a `--github-actions` is missing here in the workflow.

Lemme rerun everything with that argument again.

dklassic:
Looking again at [`linux-bencher.yml`](https://github.com/servo/servo/pull/33261/files#diff-27ee57cff0286d34c8e5d83a2fd749982cb177b7544e7be25b4494861ca831b2R67), try push does not have `--github-actions` set so it's probably expected.

I'll rerun it with that argument and see if it works.

dklassic:
So apparently the reason why GitHub Checks aren't created is just that bencher [has just released v0.4.24](https://github.com/bencherdev/bencher/releases/tag/v0.4.24) which means that change has just been made, but there's a new problem where when I rerun the whole test only lead to more confusing results:
- main push now triggers GitHub Check flow but failed to create GitHub Check
- [./mach try perf similarly died with a weird 404](https://github.com/dklassic/servo/actions/runs/11653476896/job/32446330672#step:14:884)
- [Try label worked](https://github.com/dklassic/servo/pull/7#issuecomment-2453518441) but there's no comparison report

I'm gonna have to fork bencher again to see what's holding the GitHub Check creation.

sagudev:
> @sagudev I've finally cracked the case of why GitHub Check failed to be created, it was simply because of [GitHub Check had a limit of 65535 characters while the current report had a total length of 104883](https://github.com/dklassic/servo/actions/runs/11684851456/job/32540408499#step:14:965), so multiple Checks will have to be made (is that even possible though? idk).

In that case we would need to use [github summary](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary) that has much bigger limits. I think bancher could use two switches `--report-to-github-checks` and `--report-to-github-summary`.

pewsheen:
I just saw the Github job summary was added to the action yesterday, it should be generated in the next bencher release I think
https://github.com/bencherdev/bencher/commit/fdd716f567d9ee11395b08d24e924dc5a5bd7887

dklassic:
So a slightly troubling part is that while it would be great to have try branch in personal fork also report a GitHub Check or GitHub Summary about the performance difference, it seems that bencher will simply die with error message:

> Failed to find Bencher API token, and this API endpoint requires authorization. Set the --token flag or the BENCHER_API_TOKEN environment variable.

if no token or BENCHER_API_TOKEN is provided. So if we're still providing some sort of summary in personal fork it'll had to be reported using other method.

sagudev:
> So a slightly troubling part is that while it would be great to have try branch in personal fork also report a GitHub Check or GitHub Summary about the performance difference, it seems that bencher will simply die with error message:
> 
> > Failed to find Bencher API token, and this API endpoint requires authorization. Set the --token flag or the BENCHER_API_TOKEN environment variable.
> 
> or
> 
> > a value is required for '--token ' but none was supplied
> 
> if no token or BENCHER_API_TOKEN is provided. So if we're still providing some sort of summary in personal fork it'll had to be reported using other method.

Bencher probably does not support our use case yet (basically we do not want to send results up we just want to report comparisons), so this would need to be fixed in bencher. But if we can just make it work for labels and try runs when token is present this would be enough for this PR to land, then we can deal with no token try run later.

dklassic:
> Bencher probably does not support our use case yet (basically we do not want to send results up we just want to report comparisons), so this would need to be fixed in bencher. But if we can just make it work for labels and try runs when token is present this would be enough for this PR to land, then we can deal with no token try run later.

Makes sense. I guess all what's left is to wait 

https://github.com/bencherdev/bencher/commit/fdd716f567d9ee11395b08d24e924dc5a5bd7887

to land and test if it will spit out GitHub Summary as intended in our use case then just adding all the threshold settings this PR would be good to merge.

dklassic:
https://github.com/bencherdev/bencher/releases/tag/v0.4.25

bencher 0.4.25 landed with supposedly GitHub Step Summary support. I've run some tests but for some reason the summary isn't properly written. Here's the bencher workflow [running with 0.4.25](https://github.com/dklassic/servo/actions/runs/11771497799/job/32786201077#step:7:9) but [no summary is provided in the summary page](https://github.com/dklassic/servo/actions/runs/11771497799).

I ran that old mock test and it [similarly didn't provide any summary](https://github.com/dklassic/bencher-test/actions/runs/11771593714), while a straight writing into GitHub summary environment variable does show a summary in the summary page. I'll be looking into what's happening here.

sagudev:
> https://github.com/bencherdev/bencher/releases/tag/v0.4.25
> 
> bencher 0.4.25 landed with supposedly GitHub Step Summary support. I've run some tests but for some reason the summary isn't properly written. Here's the bencher workflow [running with 0.4.25](https://github.com/dklassic/servo/actions/runs/11771497799/job/32786201077#step:7:9) but [no summary is provided in the summary page](https://github.com/dklassic/servo/actions/runs/11771497799).
> 
> I ran that old mock test and it [similarly didn't provide any summary](https://github.com/dklassic/bencher-test/actions/runs/11771561315), while a straight writing into GitHub summary environment variable [does show a summary](https://github.com/dklassic/bencher-test/actions/runs/11771600101) in the summary page. I'll be looking into what's happening here.

Oh I see what is wrong in https://github.com/bencherdev/bencher/commit/fdd716f567d9ee11395b08d24e924dc5a5bd7887. Bencher currently writes data to env var, but env var actually just contain file where data needs to be written.

sagudev:
Because when you run `echo "{markdown content}" >> $GITHUB_STEP_SUMMARY`, `$GITHUB_STEP_SUMMARY` gets expended to let's say /tmp/out.txt so the actual command that is run is `echo "{markdown content}" >> /tmp/out.txt`, which clearly writes to file not to env var.

And the only reason I know this is because I fall for it too when writing https://github.com/servo/servo/pull/33587.

dklassic:
Okay it's all here:

- bencher 0.4.26 released with a working GitHub Summary
- [Main push has a working GitHub Summary](https://github.com/dklassic/servo/actions/runs/11789776828/attempts/1#summary-32839870219)
- [So is `.\mach try perf`](https://github.com/dklassic/servo/actions/runs/11790130725/attempts/1#summary-32840571932)
- PR try label already shown working previously

So I think the testing is all done and there's probably just two actionable things left for this PR to basically be ready to merge:

- Adding extra `--github-actions` argument to `linux-bencher.yml`'s "Set bencher opts for try branch" path so that `.\mach try perf` can split out GitHub Summary as we wanted
- Currently Bencher with `--github-actions` will fail the workflow when failing to create a GitHub Check, I think we can maybe safely fail it without the aggregated result to fail

and ofc there are still some works left to do:
- Adding proper [threshold](https://bencher.dev/docs/explanation/thresholds/) setting for comparisons to show
- Moving `servospeedometer` and its hosting under servo org
- Maybe allow GitHub Check creation to fail without triggering the failure of workflow in `bencher`'s part, maybe just for when due to the character limit

sagudev:
> Currently Bencher with --github-actions will fail the workflow when failing to create a GitHub Check, I think we can maybe safely fail it without the aggregated result to fail

I do not understand what you are trying to say here.

sagudev:
Did you report
> https://github.com/dklassic/servo/pull/4#issuecomment-2451206394 has wierd reporting (not everything should be under same unit) and indeed it's not when checking up [full report](https://bencher.dev/perf/dklassic-servo/reports/c226feba-906c-48e4-9b05-8328274705f8?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dklassic-servo). This is bencher bug.

or should I?

dklassic:
> > Currently Bencher with --github-actions will fail the workflow when failing to create a GitHub Check, I think we can maybe safely fail it without the aggregated result to fail
> I do not understand what you are trying to say here

For example in this try run, the result is reported to be a failure only because the bencher part is a failure:

IMG_0030

But the reason for that bencher part failing is due to it failed to create a GitHub Check, due to the character limit of a Check. Currently bencher run with —github-action will always report a GitHub Summary, but it’ll also create Check or PR comment. Which can be safely ignored in our use case imo not a blocker.

.github/workflows/linux-bencher.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+            echo "RUN_BENCHER_OPTIONS=--branch main \
+            --github-actions ${{ secrets.GITHUB_TOKEN }}" >> "$GITHUB_ENV"
+      - name: Set bencher opts for try branch
+        if: ${{ github.event_name == 'push' && github.ref_name == 'try' }}
+        run: |
+            git remote add upstream https://github.com/servo/servo
+            git fetch upstream main
+            echo "RUN_BENCHER_OPTIONS=--branch try \
+            --github-actions \
```
dklassic, line 73:
```suggestion
            --github-actions ${{ secrets.GITHUB_TOKEN }}" \
```
Side note that GITHUB_TOKEN needs to be supplied.

sagudev:
> > > Currently Bencher with --github-actions will fail the workflow when failing to create a GitHub Check, I think we can maybe safely fail it without the aggregated result to fail
> > 
> > 
> > I do not understand what you are trying to say here
> 
> For example in [this try run](https://github.com/dklassic/servo/actions/runs/11790130725/attempts/1#summary-32840571932), the result is reported to be a failure only because the bencher part is a failure:
> 
> IMG_0030
> 
> But the reason for that bencher part failing is due to it failed to create a GitHub Check, due to the character limit of a Check. Currently bencher run with —github-action will always report a GitHub Summary, but it’ll also create Check or PR comment. Which can be safely ignored in our use case imo not a blocker.

Ah, I think we should not block on bencher failures for now. I am also hesitant to block MQ on pref job.

dklassic:
Weird thing, I tried running with the `continue-on-error` applied but it still results in failure:

https://github.com/dklassic/servo/actions/runs/11813210049

I've tried several other attempts bypassing to no avail, maybe I am missing something.

sagudev:
well in worst case we can do `|| exit 0` or simply allow build to fail if pref fails

dklassic:
https://github.com/actions/runner/issues/2347#issuecomment-1367587286

Okay digging around this seems to be a long running issue for `continue-on-error` when used on a job, the job itself will still be marked as failure. I'm running another quick attempt to see if using it specifically on the bencher step will work. But I guess we do want the whole job to be marked as success even if any part of it failed.

dklassic:
As expected, [using `continue-on-error` on a single step](https://github.com/dklassic/servo/blob/main/.github/workflows/linux-bencher.yml#L82) will work:

https://github.com/dklassic/servo/actions/runs/11819130112

Maybe we can do that for now and improve it later.

sagudev:
I rebased and squashed commits.

Thanks for all the help @dklassic, I wrote you as coauthor of squashed commit.

etc/ci/bencher.py
```diff
@@ -0,0 +1,54 @@
...
+import json
+import os
+
+
+def size(args):
+    size = os.path.getsize(args.binary)
+    print(size)
+    with open(args.bmf_output, 'w', encoding='utf-8') as f:
+        json.dump({
+            'servo': {
```
jschwe, line 22: Can we make this configurable? 
Ideally this should include the target triple, cargo features (if not default) and maybe the cargo profile.
I think we want to track the binary size for all the different platforms, and I would also want to track `--no-default-features` vs `default-features` and also the size with different profiles, e.g. a profile with `opt-level=3` vs `opt-level=z`.

This comment is definitly not blocking though, just a suggestion that we can follow-up on zulip or in an issue.

etc/ci/bencher.py
```diff
@@ -0,0 +1,54 @@
...
+                }
+            }
+        }, f, indent=4)
+
+
+def merge(args):
+    output: dict[str, object] = dict()
+    for input_file in args.inputs:
+        with open(input_file, 'r', encoding='utf-8') as f:
+            output = json.load(f) | output
```
jschwe, line 34: Could we issue a warning message when the same key is present in both dicts?

jschwe (approved):

etc/ci/bencher.py
sagudev: The problem is that we would need to do all these build in CI to actually track them for file size.

jschwe: Sure - but at least for the `main` branch we don't actually have that many new commits per day. I think it makes sense to gather more data for the main branch, since that is quite useful to at least identify large code-size increases quickly.

Spotting such issues before they are merged is of course even better, but that indeed requires quite a lot more build-time and probably would mean we would first need more build servers (or optimize caching)

wusyong (approved):
I'd also love to see this happen in action!
If this doesn't affect other CI, I think we can merge this and see how it goes.

jschwe:
Any updates here? Looks like it needs a rebase. Are you also still waiting on a review from Martin?

sagudev:
> I think the commit message is missing a bit of context here. What are we measuring exactly? Performance, binary size? I think the message should just explain all of this.

We are measuring both. This just lays ground work and provides some basic benchmarks. I think later we can also add some info from tracing like the one described here: https://github.com/servo/servo/wiki/Servo-Benchmarking-Report-%28November-2024%29#user-facing-paint-metrics

atbrakhi (commented):
@sagudev  I guess we are waiting for servo account to be created on bencher dev and the api token?

sagudev:
Actually I still didn't do:

> I will rebase and write more clear message.

because I forgot. Will do it now. But generally this is blocked on account and setting token as secret (I do not have permission to add CI secrets).

atbrakhi:
@sagudev we have an account now: https://bencher.dev/console/projects/servo-org-s-project also added the token!

epompeii:
@atbrakhi would you like a prettier URL for the project?
I think @pewsheen would be okay with letting you all having: https://bencher.dev/perf/servo
That is, your project slug would simply be `servo`. Just give the the okay, and I'll make the switch.

pewsheen:
> @atbrakhi would you like a prettier URL for the project? I think @pewsheen would be okay with letting you all have: https://bencher.dev/perf/servo That is, your project slug would simply be `servo`. Just give the the okay, and I'll make the switch.

I just changed the slug. It should be able to use `servo` now

atbrakhi:
@epompeii @pewsheen  Thank you. I had already changed the slug to "servo" yesterday but it seems like it didn’t take effect. I appreciate your help regardless :)

wusyong:
@pewsheen updated the slug today. Perhaps @atbrakhi you need to set it again to see if that takes effect?

atbrakhi:
> @pewsheen updated the slug today. Perhaps @atbrakhi you need to set it again to see if that takes effect?

I noticed that there are two slugs one on org level and one on project level. Either way it should be fine now

atbrakhi:
@sagudev now that we have the account and token, i will try to test these changes today. If you see any blocker on your side let me know! hopefully we can land it soon

atbrakhi:
> > @sagudev now that we have the account and token, i will try to test these changes today. If you see any blocker on your side let me know! hopefully we can land it soon
> 
> Did you set github secrets, if yes I can trigger try run servo/servo right now.

yes, I did. we have BENCHER_API_TOKEN in secrets now, since yday. Sorry if i wasn't clear in my previous comment.

atbrakhi:
Try run results: https://github.com/servo/servo/actions/runs/12179777680
Bencher: https://bencher.dev/perf/servo

atbrakhi (approved):

sagudev:
There are no blockers on my side.

atbrakhi:
> There are no blockers on my side.

nice, I noticed that we have some improvement comments above by you as well as @dklassic , maybe we can create a new issue and list the further improvements there, if someone else is interested they can also work on it.

We can land this PR I believe :)

PR #33262, opened 2024-08-30
script: Stop using `time` in DOM timers and `XMLHttpRequest`

mrobinson:
This switches to using `std::time` types for DOM timer operations, which
allows removing our custom time units in favor of `Duration`.

---


jdm (approved):
Looks good! I didn't realize how deeply embedded time 0.1 was everywhere.

PR #33263, opened 2024-08-30
Fix various issues with replaced elements in flex layout

Loirooriol:
In particular, this takes into account that flex items may be stretched, and if they have an aspect ratio, we ma6y need to convert the stretched size through the ratio.

---

    `row-use-cases-001.html` is now failing one subtest, but this aligns us closer to Gecko/Blink/WebKit, which fail all subtests.


components/layout_2020/flexbox/layout.rs
```diff
@@ -244,6 +243,32 @@ struct FinalFlexLineLayout {
...
+        let mapped_value = match value {
+            AlignFlags::AUTO | AlignFlags::NORMAL => self.align_items.0,
+            _ => value,
+        };
+        AlignItems(mapped_value)
+    }
+
+    // Whether an item with an `auto` preferred cross size needs to be stretched
+    // to fill the flex container.
+    // <https://drafts.csswg.org/css-flexbox/#stretched>
```
mrobinson, line 258:
```suggestion
    /// Whether an item with an `auto` preferred cross size needs to be stretched
    /// to fill the flex container.
    /// <https://drafts.csswg.org/css-flexbox/#stretched>
```

components/layout_2020/flexbox/layout.rs
```diff
@@ -1049,6 +1084,18 @@ fn cross_axis_is_item_block_axis(
     container_is_row ^ item_is_orthogonal
 }
 
+// Whether an item with an `auto` preferred cross size needs to be stretched
+// to fill the cross size of its flex line.
+// <https://drafts.csswg.org/css-flexbox/#stretched>
```
mrobinson, line 1089:
```suggestion
/// Whether an item with an `auto` preferred cross size will stretched to fill the cross size of its flex line.
/// <https://drafts.csswg.org/css-flexbox/#stretched>
```

components/layout_2020/flexbox/layout.rs
```diff
@@ -244,6 +243,32 @@ struct FinalFlexLineLayout {
...
+    fn item_with_auto_cross_size_stretches_to_container_size(
+        &self,
+        style: &ComputedValues,
+        margin: &FlexRelativeSides<AuOrAuto>,
+    ) -> bool {
+        self.container_is_single_line &&
+            item_with_auto_cross_size_stretches_to_line_size(
+                self.align_for(style.clone_align_self()),
+                margin,
+            )
```
mrobinson, line 268:
```suggestion
        item_style: &ComputedValues,
        item_margin: &FlexRelativeSides<AuOrAuto>,
    ) -> bool {
        self.container_is_single_line &&
            item_with_auto_cross_size_stretches_to_line_size(
                self.align_for(item_style.clone_align_self()),
                item_margin,
            )
```

components/layout_2020/flexbox/layout.rs
```diff
@@ -2137,14 +2192,25 @@ impl FlexItemBox {
...
         };
 
+        let cross_size =
+            if content_box_size.cross.is_auto() && auto_cross_size_stretches_to_container_size {
+                let outer_cross_size = if cross_axis_is_item_block_axis {
+                    containing_block.size.block
+                } else {
+                    containing_block.size.inline
+                };
+                outer_cross_size.map(|v| v - pbm_auto_is_zero.cross)
```
mrobinson, line 2202:
```suggestion
               if cross_axis_is_item_block_axis {
                    containing_block.size.block
                } else {
                    containing_block.size.inline
                }.map(|v| v - pbm_auto_is_zero.cross)
```

mrobinson: It would be nice to add some sort of mapping in another PR that would make this easier. Imagine something like this:

```rust
    containing_block.size
        .get_axis(self.axes.cross)
        .map(|v| v - pbm_auto_is_zero.cross)
```

where `self.axes` is `FlowRelativeVec2<Direction>`.

mrobinson (approved):
Just a couple nits below. This fixes a lot of tests!

PR #33264, opened 2024-08-30
Revert "Allow prefs to be overridden from a file and set WPT-specific prefs from file (#33163)"

mrobinson:
This reverts commit e85491b5fc0ccbedfa312b82edf37d8e1cecc780.

---

### DIFF
--- a/ports/servoshell/desktop/cli.rs
+++ b/ports/servoshell/desktop/cli.rs
@@ -45,12 +45,6 @@ pub fn main() {
         "A preference to set to disable",
         "dom.webgpu.enabled=false",
     );
-    opts.optmulti(
-        "",
-        "prefs-file",
-        "Load in additional prefs from a file.",
-        "--prefs-file /path/to/prefs.json",
-    );
 
     let opts_matches;
     let content_process_token;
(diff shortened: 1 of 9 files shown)


jdm (approved):

jdm:
I'm guessing wpt runs are still broken?

mukilan:
The upstream [WPT pr with the most recent fix wasn't merged](https://github.com/web-platform-tests/wpt/pull/47855) until earlier today, because the issue with WPT export job being broken. I think tomorrow's run will be the test for whether that worked, so maybe we can merge this PR if we find tomorrow's run is still bad?

msub2:
It looks like the 30th's run is fixed, no more flatlined numbers

mrobinson:
> It looks like the 30th's run is fixed, no more flatlined numbers

Do you have a link to these results? I still see the run from the 27th listed on wpt.fyi: https://wpt.fyi/results/?label=master&product=chrome%5Bexperimental%5D&product=edge%5Bexperimental%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&product=servo&aligned

msub2:
If you uncheck "aligned runs" you should see one that ran for the 31st

mrobinson:
Thanks. I wasn't aware of the "aligned runs" setting. The results indeed look a lot better, so I don't think we need to revert this change!

PR #33271, opened 2024-08-31
Update FakeXRDevice to support updating bounds

msub2:
This updates the IDL of FakeXRDevice to add some of the more recent additions to the spec, and implements setBoundsGeometry. It also updates boundsGeometry's underlying representation to support an arbitrary amount of points.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7783,7 +7783,7 @@ dependencies = [
 [[package]]
 name = "webxr"
 version = "0.0.1"
-source = "git+https://github.com/servo/webxr#7656508fdcda194997cdaa751f2b98797d963867"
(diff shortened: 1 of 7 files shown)


jdm (approved):

msub2:
Hmm, the test for light estimation only passes 1 argument to setViews, but the IDL doesn't mention anything about the secondary views argument being optional. I'm inclined to think that it should be since it would match the behavior of Chrome right now, where the JS polyfill has internal if-checks that prevent undefined secondaryViews from causing a problem.

msub2:
Opened a PR on the spec as well https://github.com/immersive-web/webxr-test-api/pull/84

PR #33279, opened 2024-09-01
Make all platforms use a delay during cookie tests.

### LINKED ISSUE #33278: Intermittent cookie unit test failure
```
---- cookie::test_sort_order stdout ----
thread 'cookie::test_sort_order' panicked at components/net/tests/cookie.rs:196:5:
  left: Equal
 right: Less
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   4: main::cookie::test_sort_order
```
I suspect this appeared as a result of #33260.

jdm:
- [x] These changes hopefully fix #33278.

### DIFF
--- a/components/net/tests/cookie.rs
+++ b/components/net/tests/cookie.rs
@@ -165,7 +165,6 @@ fn test_cookie_host_prefix() {
     assert!(ServoCookie::new_wrapped(cookie, url, CookieSource::HTTP).is_some());
 }
 
-#[cfg(target_os = "windows")]
 fn delay_to_ensure_different_timestamp() {
     use std::thread;
     use std::time::Duration;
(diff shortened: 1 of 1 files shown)


sagudev (approved):
LGTM, although ideal solution would be to abstract away `SystemTime::now()` so that we could mock it in tests.

mrobinson (approved):

mrobinson:
> LGTM, although ideal solution would be to abstract away `SystemTime::now()` so that we could mock it in tests.

Or perhaps somehow ensure that monotonic timestamps are used for cookie creation. I think that you can estimate a time using monotonic counters and the `timeOrigin`. I'm not sure if that's something we'd like though.

mrobinson:
BTW, the issue here is that `SystemTime::now()` is not monotonic, which means that calling it twice successively can result in the second time being before the first. I imagine this was the issue we were seeing in Windows as well.

PR #33282, opened 2024-09-02
script: Create a `CrossProcessInstant` to enable serializable monotonic time

mrobinson:
Up until now, Servo was using a very old version of time to get a
cross-process monotonic timestamp (using `time::precise_time_ns()`).
This change replaces the usage of old time with a new serializable
monotonic time called `CrossProcessInstant` and uses it where `u64`
timestamps were stored before. The standard library doesn't provide this
functionality because it isn't something you can do reliably on all
platforms. The idea is that we do our best and then fall back
gracefully.

This is a big change, because Servo was using `u64` timestamps all over
the place some as raw values taken from `time::precise_time_ns()` and
some as relative offsets from the "navigation start," which is a concept
similar to DOM's `timeOrigin` (but not exactly the same). It's very
difficult to fix this situation without fixing it everywhere as the
`Instant` concept is supposed to be opaque. The good thing is that this
change clears up all ambiguity when passing times as a `time::Duration`
is unit agnostic and a `CrossProcessInstant` represents an absolute
moment in time.

The `time` version of `Duration` is used because it can both be negative
and is also serializable.

Good things:
 - No need too pass around `time` and `time_precise` any longer.
   `CrossProcessInstant` is also precise and monotonic.
 - The distinction between a time that is unset or at `0` (at some kind
   of timer epoch) is now gone.

There still a lot of work to do to clean up timing, but this is the
first step. In general, I've tried to preserve existing behavior, even
when not spec compliant, as much as possible. I plan to submit followup
PRs fixing some of the issues I've noticed.

---


components/shared/base/cross_process_instant.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// The frequency of the value returned by `QueryPerformanceCounter` in counts per
+    /// second. This is taken from the Rust source code at:
+    /// <https://github.com/rust-lang/rust/blob/1a1cc050d8efc906ede39f444936ade1fdc9c6cb/library/std/src/sys/pal/windows/time.rs#L197>
+    #[allow(unsafe_code)]
+    fn frequency() -> i64 {
+        // Either the cached result of `QueryPerformanceFrequency` or `0` for
+        // uninitialized. Storing this as a single `AtomicU64` allows us to use
+        // `Relaxed` operations, as we are only interested in the effects on a
+        // single memory location.
+        static FREQUENCY: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
mrobinson: Can you expand on your comment? This frequency is cached across threads, but my understanding is that the actual counter value is cross-process, regardless of whether or not the frequency is fetched multiple times across different processes.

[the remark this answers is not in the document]
mrobinson: Note here the frequency is used just to convert the counter value to nanoseconds.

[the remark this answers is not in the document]
sagudev: You're right. I looked over to quickly.

components/devtools/actors/network_event.rs
```diff
@@ -300,8 +300,8 @@ impl Actor for NetworkEventActor {
                 let timings_obj = Timings {
                     blocked: 0,
                     dns: 0,
-                    connect: self.request.connect_time,
-                    send: self.request.send_time,
+                    connect: self.request.connect_time.as_nanos() as u64,
```
mukilan, line 303: Should we use `as_millis()` here? AFAICT, request.connect_time used to represent duration in milliseconds because its value is ultimately derived from https://github.com/servo/servo/blob/75c7712905c522f604a454dcb1c18b8e3f163d3b/components/net/http_loader.rs#L675

and connect_start and connect_end were intialized with

https://github.com/servo/servo/blob/75c7712905c522f604a454dcb1c18b8e3f163d3b/components/net/http_loader.rs#L617
https://github.com/servo/servo/blob/75c7712905c522f604a454dcb1c18b8e3f163d3b/components/net/http_loader.rs#L641

components/devtools/actors/network_event.rs
```diff
@@ -300,8 +300,8 @@ impl Actor for NetworkEventActor {
...
+                    send: self.request.send_time.as_nanos() as u64,
```
mukilan, line 304: Same comment as `connect_time` above.

components/shared/base/cross_process_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! In implementation a monotonic nanosecond precision timer, like [`std::time::Instant`]
```
gterzian, line 11: typo

components/devtools/actors/network_event.rs
mrobinson: You're right! I've modified this as well as all of the other uses of `CrossProcessInstant` and `Duration` in the DevTools code.

components/shared/base/cross_process_instant.rs
```diff
@@ -0,0 +1,170 @@
...
+        Self {
+            value: platform::now(),
+        }
+    }
+
+    /// Some unspecified time epoch. This is mainly useful for converting
+    /// [`CrossProcessInstant`] into a `DOMHighResolutionTimestamp`.
+    ///
+    /// TODO: This should be replaced with a proper implementaiton of `timeOrigin`. See
+    /// <https://html.spec.whatwg.org/multipage/#concept-settings-object-time-origin>
```
gterzian, line 43: The more precise spec link would be https://w3c.github.io/hr-time/#sec-time-origin.

components/script/dom/performance.rs
```diff
@@ -139,7 +136,8 @@ pub struct Performance {
     buffer: DomRefCell<PerformanceEntryList>,
     observers: DomRefCell<Vec<PerformanceObserver>>,
     pending_notification_observers_task: Cell<bool>,
-    navigation_start_precise: u64,
+    #[no_trace]
+    time_origin: CrossProcessInstant,
```
gterzian, line 140: I think here we can add https://html.spec.whatwg.org/multipage/#concept-settings-object-time-origin

The question that remains to me is: how do we share this per [`environment-settings-object`](https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object)?

components/shared/base/cross_process_instant.rs
mrobinson: Fixed. Thanks!

components/shared/base/cross_process_instant.rs
mrobinson: I've updated the spec link and improve the comment a bit to more accurately reflect the state of the code.

components/script/dom/performance.rs
mrobinson: I've added the spec link. I think this is already close to the concept of the `timeOrigin` as it is set per-`Worker` and also per `GlobalScope` in the page. Any further improvement should probably come in followups.

gterzian: Right, thanks for clarifying.

gterzian (approved):
LGTM on the script side.

components/script/dom/document.rs
```diff
@@ -5441,11 +5441,9 @@ impl DocumentMethods for Document {
     }
 }
 
-fn update_with_current_time_ms(marker: &Cell<u64>) {
-    if marker.get() == 0 {
-        let time = time::get_time();
-        let current_time_ms = time.sec * 1000 + time.nsec as i64 / 1000000;
-        marker.set(current_time_ms as u64);
+fn update_with_current_time_ms(marker: &Cell<Option<CrossProcessInstant>>) {
```
mukilan, line 5444: Perhaps the name needs to be changed as this no longer stores duration in milliseconds.

components/profile/time.rs
```diff
@@ -300,16 +299,17 @@ impl Profiler {
     }
 
     /// Get tuple (mean, median, min, max) for profiler statistics.
-    pub fn get_statistics(data: &[f64]) -> (f64, f64, f64, f64) {
-        data.iter().fold(-f64::INFINITY, |a, &b| {
+    pub fn get_statistics(data: &[Duration]) -> (Duration, Duration, Duration, Duration) {
+        data.iter().fold(0, |a, &b| {
+            let b = b.whole_nanoseconds();
```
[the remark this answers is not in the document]
mrobinson: This could be simplified even further, I think since it is only trying to make sure the data is sorted:

```rust
        debug_assert!(
            data.windows(2).all(|window| window[0] <= window[1]),
            "Data must be sorted"
        );
```

I've done this.

components/metrics/lib.rs
```diff
@@ -50,40 +55,28 @@ fn set_metric<U: ProgressiveWebMetric>(
...
-        Some(time) => time,
-        None => SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap_or_default()
-            .as_nanos() as u64,
-    };
-    let time = now - navigation_start;
-    attr.set(Some(time));
+    let metric_time = metric_time.unwrap_or_else(CrossProcessInstant::now);
+    attr.set(Some(metric_time));
```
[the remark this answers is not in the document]
mrobinson: That could lead to quite a problem in the channel that was receiving these metrics though. How would it know what the `Duration` was relative to if it was receive messages from many different pages. Using `CrossProcessInstant` eliminates this ambiguity. I think it's an improvement.

On the other side of the message this data is used two ways:

1. Written directly to JSON output as a number. I've verified that the tool for processing this data (`trace-dump.js`) normalizes all times to the minimum timestamp.
2. Used to calculate a duration (subtracting the two times). This should matter.

components/metrics/lib.rs
```diff
@@ -50,40 +55,28 @@ fn set_metric<U: ProgressiveWebMetric>(
...
+    );
 
     // Print the metric to console if the print-pwm option was given.
     if opts::get().print_pwm {
-        println!(
-            "Navigation start: {}",
-            pwm.get_navigation_start().unwrap().to_ms()
-        );
-        println!("{:?} {:?} {:?}", url, metric_type, time.to_ms());
+        println!("{:?} {:?} {:?}", url, metric_type, metric_time);
```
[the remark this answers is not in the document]
mrobinson: In this case it's an instant, so it's hard to print a very useful value here, but I can improve it.

components/metrics/lib.rs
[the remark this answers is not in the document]
mukilan: I see.  I was assuming the receiver didn't have the requirement to resolve the duration against a fixed instant in the earlier code. Do you know if it is intentional that we send use `metric_time` as both start and end time in the call to `send_profile_data` below - wouldn't that make the duration zero?

components/profile/time.rs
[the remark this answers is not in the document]
mukilan: Nice!

components/metrics/lib.rs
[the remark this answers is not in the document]
mrobinson: Yes, I think it's intentional. It's a way to mark that an event happened at a certain point in the timeline without having the duration of the event. I'm not sure how well the existing tools process these events though TBH.

mukilan (approved):
Great work!

PR #33283, opened 2024-09-02
CI: label self-hosted runners with run id to aid debugging

delan:
This patch adds `reserved-by:` labels to self-hosted runners, so that we can look at a busy runner and find the workflow run that reserved it.

For example, [in run 10666002164](https://github.com/servo/servo/actions/runs/10666002164/job/29560562028), we add the label `reserved-by:servo/servo/actions/runs/10666002164`, which we can combine with the job uuid (`reserved-for:931a2e13-72bb-4af5-9928-2aae454f2128`) to find the workload under:

- <https://github.com/servo/servo/actions/runs/10666002164>
- <https://api.github.com/repos/servo/servo/actions/runs/10666002164> `.jobs_url`
  - <https://api.github.com/repos/servo/servo/actions/runs/10666002164/jobs>

---

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -127,6 +127,7 @@ jobs:
           gh api "$self_hosted_runner_scope/$(cat $runner_id)/labels" \
             -f "labels[]=reserved-for:$unique_id" \
             -f "labels[]=reserved-since:$reserved_since" \
+            -f 'labels[]=reserved-by:${{ github.repository }}/actions/runs/${{ github.run_id }}' \
             --method POST --silent
           echo "selected_runner_label=reserved-for:$unique_id" | tee -a $GITHUB_OUTPUT
           echo 'is_self_hosted=true' | tee -a $GITHUB_OUTPUT


sagudev (approved):

sagudev:
@mrobinson Sign-off is missing but I am still allowed to put PR to MQ, I think we need to change settings for MQ to also wait for DCO check.

mrobinson:
> @mrobinson Sign-off is missing but I am still allowed to put PR to MQ, I think we need to change settings for MQ to also wait for DCO check.

I've done this.

sagudev:
> > @mrobinson Sign-off is missing but I am still allowed to put PR to MQ, I think we need to change settings for MQ to also wait for DCO check.
> 
> I've done this.

Now DCO is also marked as required so it works, might be worth checking other servo repos with DCO too (like mozjs).

sagudev:
per https://github.com/orgs/community/discussions/46757?sort=old#discussioncomment-4909336 MQ needs all required checks to also run on pushes which DCO doesn't :(

This is also known problem in DCO bot: https://github.com/dcoapp/app/issues/199, but hacky solution exists: https://github.com/dcoapp/app/issues/199#issuecomment-1626060564

PR #33287, opened 2024-09-02
Send less title changes to the embedder

### LINKED ISSUE #33286: Excessive number of title updates when loading any page
**Describe the bug:**
When loading any page (tested on https://servo.org), the `EmbedderMsg::ChangePageTitle` message is sent many times, incrementally updating the title, even though it never *actually* changes.

Example logs:
```
title: Some("")
title: Some("Servo,")
title: Some("Servo, the embeddable")
title: Some("Servo, the embeddable, independent, memor")
title: Some("Servo, the embeddable, independent, memory-safe, modular")
title: Some("Servo, the embeddable, independent, memory-safe, modular, parallel web r")
title: Some("Servo, the embeddable, independent, memory-safe, modular, parallel web rendering engine")
```

This is because the parser hands out chunks of the elements text incrementally, causing `HtmlTitleElement::children_changed` to be called repeatedly.
While not a big issue, this seems like a silly thing to do.

**To Reproduce:**
apply
```diff
diff --git a/ports/servoshell/desktop/webview.rs b/ports/servoshell/desktop/webview.rs
index a6a33f6a7a..01612f5f9b 100644
--- a/ports/servoshell/desktop/webview.rs
+++ b/ports/servoshell/desktop/webview.rs
@@ -628,6 +628,8 @@ where
                     need_update = true;
                 },
                 EmbedderMsg::ChangePageTitle(title) => {
+                    println!("title: {title:?}");
+
                     // Set the title to the target webview, and update the OS window title
                     // if this is the currently focused one.
                     if let Some(webview_id) = webview_id {

```
and load any page.

**Expected Behaviour**:
One single `EmbedderMsg::ChangePageTitle` once the entire `<title>` element has been parsed.

webbeef:
Instead of sending a title change for each incremental parsing change, we now do it when:
- the element is bound to the tree.
- the parser is done with the element (in pop())
- the title content changes after the parsing, eg. using document.title

---
- [X] These changes fix #33286 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/htmltitleelement.rs
+++ b/components/script/dom/htmltitleelement.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::cell::Cell;
+
 use dom_struct::dom_struct;
 use html5ever::{LocalName, Prefix};
 use js::rust::HandleObject;
@@ -19,6 +21,7 @@ use crate::dom::virtualmethods::VirtualMethods;
 #[dom_struct]
 pub struct HTMLTitleElement {
     htmlelement: HTMLElement,
+    popped: Cell<bool>,
 }
 
 impl HTMLTitleElement {
@@ -29,6 +32,7 @@ impl HTMLTitleElement {
     ) -> HTMLTitleElement {
         HTMLTitleElement {
             htmlelement: HTMLElement::new_inherited(local_name, prefix, document),
+            popped: Cell::new(false),
         }
     }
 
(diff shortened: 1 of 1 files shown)


components/script/dom/htmltitleelement.rs
```diff
@@ -70,9 +74,14 @@ impl VirtualMethods for HTMLTitleElement {
...
-        if node.is_in_doc() {
-            node.owner_doc().title_changed();
+
+        // Notify of title changes only after the initial full parsing
+        // of the element.
+        if self.popped.get() {
+            let node = self.upcast::<Node>();
+            if node.is_in_doc() {
+                node.owner_doc().title_changed();
+            }
```
Taym95, line 84: Maybe create a function for this to avoid duplication?

jdm:
I wonder if we should solve this in the parser instead? I haven't fully wrapped my head around the implications, but I wonder if we could build up a text buffer in the parser and flush it all at once rather than update element contents incrementally.

jdm:
I've filed #33339 about investigating the underlying parser issue. I think this is a reasonable patch over the existing behaviour until we can prioritize that investigation.

jdm (approved):

atbrakhi (commented):
@webbeef just the Signed-off is missing :)

PR #33023, opened 2024-08-12
fix: add error handling to BaseAudioContext::new_inherited

Taym95:
---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #32986

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

PR #33289, opened 2024-09-02
Return NotSupported error for AudioContext sample_rate 1 or 1000000

Taym95:
I noticed two tests in `audiocontextoptions.html.ini` specifically checking for sample_rate with values 1 and 1000000, expecting a `NotSupported` result. However, I don't see anything in the spec regarding this.

I've created a draft PR to get feedback from others. Please take a look and let me know your thoughts.

---

### DIFF
--- a/components/script/dom/baseaudiocontext.rs
+++ b/components/script/dom/baseaudiocontext.rs
@@ -124,6 +124,12 @@ impl BaseAudioContext {
             },
         };
 
+        // A sampleRate of 1 and 1,000,000 are unlikely to be supported on any browser,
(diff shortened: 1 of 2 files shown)


jdm:
We shouldn't special-case values like this. This should be addressed through the fallible audio pipeline creation in https://github.com/servo/servo/pull/33023 instead. Are you testing locally with the gstreamer backend enabled?

Taym95:
> We shouldn't special-case values like this. This should be addressed through the fallible audio pipeline creation in #33023 instead. Are you testing locally with the gstreamer backend enabled?

thanks for response, no I don't think I am testing with gstreamer backend enabled

PR #33290, opened 2024-09-02
CI: Hack around DCO not in MQ

sagudev:
See https://github.com/servo/servo/pull/33283#issuecomment-2325240423

---

- [x] These changes do not require tests because it's CI changes


mrobinson (approved):

sagudev:
@mrobinson I think we can require DCO check again, not that this landed.

PR #33295, opened 2024-09-03
openharmony: add servoshell for ohos

mukilan:
---


support/openharmony/entry/oh-package.json5
```diff
@@ -0,0 +1,12 @@
+{
+  "license": "",
```
jschwe, line 2: We should probably add the license here. Can we have keep the ArkTS app Apache-2 licensed?

support/openharmony/oh-package.json5
```diff
@@ -0,0 +1,12 @@
+{
+  "license": "MPL",
```
jschwe, line 2: Can we keep the license of the ArkTS project under Apache-2?

python/servo/command_base.py
```diff
@@ -553,6 +554,11 @@ def common_command_arguments(build_configuration=False, build_type=False, binary
                     help='Build for Android. If --target is not specified, this '
                          'will choose a default target architecture.',
                 ),
+                CommandArgument(
+                    '--ohos', default=None, action='store_true',
```
jschwe, line 558: What benefit does adding a new argument give us? In principle the target triple already tells us that we are targetting ohos.

.github/workflows/ohos.yml
```diff
@@ -52,12 +52,27 @@ jobs:
...
+        run: |
+          mkdir ~/hvigor-installation
+          cd ~/hvigor-installation
+          echo "@ohos:registry=https://repo.harmonyos.com/npm/" > .npmrc
+          npm install "@ohos/hvigor@4" "@ohos/hvigor-ohos-plugin@4"
+          echo "HVIGOR_PATH=$PWD" >> $GITHUB_ENV
       - name: Build (arch ${{ matrix.arch }} profile ${{ inputs.profile }})
         env:
           OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+          OHOS_BASE_SDK_HOME: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
```
jschwe, line 75: Is there something that expects this variable name instead of `OHOS_SDK_NATIVE`?

python/servo/command_base.py
mukilan: It is just an alias to build the default architecture for the platform. The only value is the UX for the user, which is debatable, but we have it for android, so I've added for OH as well.

.github/workflows/ohos.yml
mukilan: hvigor needs this to find the other components like toolchain and ets that it needs for building the shell. The value should actually come from the `ohos_base_sdk_home` output of the setup_sdk step. I'll fix this typo.

jschwe: Interesting. Is this perhaps only needed for hvigor 4.x? I don't think I'm setting this variable when building on the commandline locally.

mukilan: Just tried locally with hvigor 5.x:
```
Info: The OpenHarmony SDK 4.1.7.5 is targeting API-level 11
```
> hvigor ERROR: Unable to find 'sdk.dir' in 'local.properties' or 'OHOS_BASE_SDK_HOME' in the system environment path.
	 at /home/mukilan/servo-ohos/target/openharmony/aarch64-unknown-linux-ohos/release/local.properties
> hvigor ERROR: BUILD FAILED in 1 s 797 ms 
```

Do perhaps have sdk.dir in local.properties or DEVECO_SDK_HOME set?

**Comment by jschwe** [comment]
Regarding signing - I updated the servodemo with a small patch (https://github.com/jschwe/ServoDemo/pull/5)  which lets hvigor read the signing configuration from a seperate file. I also added an example file to the repo.
In principle hvigor can be extended by arbitrary typescript code, so you could do something more complex. But I think for our usecase reading the signing configuration from a file which is not in git should be sufficient.

File: support/openharmony/entry/oh-package.json5 (hunk 1 above)
Comment (mukilan) [reply support/openharmony/entry/oh-package.json5]: Done in [e0b51fb](https://github.com/servo/servo/pull/33295/commits/e0b51fb229acbabdb6d2f192cea530fa7d151538)

File: support/openharmony/oh-package.json5 (hunk 2 above)
Comment (mukilan) [reply support/openharmony/oh-package.json5]: Done in [e0b51fb](https://github.com/servo/servo/pull/33295/commits/e0b51fb229acbabdb6d2f192cea530fa7d151538)

File: .github/workflows/ohos.yml [hunk 5]
```diff
@@ -52,12 +52,34 @@ jobs:
         run: sudo apt update && python3 ./mach bootstrap --skip-lints
       - name: Setup OpenHarmony SDK
         id: setup_sdk
-        uses: openharmony-rs/setup-ohos-sdk@v0.1
+        uses: openharmony-rs/setup-ohos-sdk@v0.1 # TODO: this needs to be updated
```
mukilan, line 55: @jschwe could you publish a new version of the action for this.

jschwe: Published v0.1.2 and redirected v0.1 to point to v0.1.2

mukilan: Thanks! I've removed the todo.

support/openharmony/build-profile.json5
```diff
@@ -0,0 +1,44 @@
...
+        "signingConfig": "default",
+        "compileSdkVersion": 11,
+        "compatibleSdkVersion": 11,
+        "targetSdkVersion": 11,
+        "runtimeOS": "OpenHarmony"
+      },
+      {
+        "name": "harmonyos",
+        "signingConfig": "default",
+        "compatibleSdkVersion": "4.0.0(10)",
```
jschwe, line 15: let's change this to "4.1.0(11)" to match the API version for OpenHarmony.

support/openharmony/entry/src/main/ets/pages/Index.ets
jschwe, line 1: .gitattributes:
```
*.ets  linguist-language=ts
# or 
support/openharmony/**/*.ets linguist-language=ts
```

If we add one of the above to `.gitattributes` I think we should get syntax highlighting on github for the ets files. Could you perhaps try that out? (I guess ArkTS is not exactly TS, but it hopefully should not matter for syntax highlighting)

python/servo/command_base.py
jschwe: Considering that the long term plan is to remove mach, I'm not convinced having users get used to convenience options that "cargo build" does not have is desirable.

python/servo/package_commands.py
```diff
@@ -179,6 +188,50 @@ def package(self, build_type: BuildType, flavor=None, with_asan=False):
             except subprocess.CalledProcessError as e:
                 print("Packaging Android exited with return value %d" % e.returncode)
                 return e.returncode
+        elif self.is_openharmony():
+            # hvigor doesn't support an option to place output files in a specific directory
+            # so copy the source files into the target/openharmony directory first.
+            ohos_app_dir = path.join(self.get_top_dir(), "support", "openharmony")
+            ohos_target_dir = path.join(
```
jschwe, line 195: The source files of OH app, shouldn't be different for release / debug mode, so I don't think adding `build_type.directory_name()` is strictly necessary.

python/servo/package_commands.py
```diff
@@ -179,6 +188,50 @@ def package(self, build_type: BuildType, flavor=None, with_asan=False):
...
+                self.get_top_dir(), "target", "openharmony", self.target.triple(), build_type.directory_name())
+            if path.exists(ohos_target_dir):
+                print("Cleaning up from previous packaging")
+                delete(ohos_target_dir)
+            shutil.copytree(ohos_app_dir, ohos_target_dir)
+
+            hvigor_command = ["--no-daemon", "assembleHap", "-p", "product=default"]
```
jschwe, line 202: I guess here we have the difficulty that from the target-triple alone, we can't deduce if the user wants to build for OpenHarmony or HarmonyOS.
I guess for now It's fine to just support the default OH build.

python/servo/package_commands.py
```diff
@@ -52,6 +53,10 @@
     'mac': [
         'production/servo-tech-demo.dmg',
     ],
+    'openharmony': [
+        'openharmony/aarch64-unknown-linux-ohos/production/'
```
jschwe, line 57: Does this line mean the shared library will be archived? If yes, that would be great, since it would allow also simply downloading and packaging a prebuilt servo library from the external ServoDemo repository.

jschwe (commented):
Left some comments, but I think this PR is basically fine as is.

Just as a note, I also have a branch on my ServoDemo repo https://github.com/jschwe/ServoDemo/pull/6, which allows building servo directly via the IDE (and without requiring mach, pure cargo build).
I think both approaches can co-exist, just wanted to mention that the OH build can in principle work fully without mach already (and with only very little CMake code).

support/openharmony/build-profile.json5
mukilan: Good point. I've changed it to "4.1.0(11)".

support/openharmony/entry/src/main/ets/pages/Index.ets
mukilan: I've added `support/openharmony/**/*.ets linguist-language=ts` to .gitattributes. It seems to work when viewing the [file directly](https://github.com/servo/servo/blob/a5798ceadae9c3834db3afa03cbb477452128e73/support/openharmony/entry/src/main/ets/pages/Index.ets), but not in the diff view. Maybe that is just a limitation of GitHub's diff UI?

python/servo/package_commands.py
mukilan: The idea is to allow the HAP (which will be generated inside the source folder hierarchy) for debug/release/production profiles to co-exist like we do for other targets.

python/servo/package_commands.py
mukilan: For Android we have the `--flavor` argument to choose the variant. I was hoping to reuse that eventually, but I have not tested the HarmonyOS variant and it also requires different environment variables during build (DEVECO_SDK_HOME vs OHOS_BASE_SDK_HOME) so this currently defaults to OH.

python/servo/package_commands.py
mukilan: Nope, this is just using python's string syntax to break a single long line. I've now removed this line as we don't have a step in the workflow to upload the packages to github releases. Adding new published packages should probably happen in a separate PR and maybe we can consider prebuilt `.so` on GH release as part of that.

python/servo/command_base.py
mukilan: Fair,  but I don't see mach being replaced very soon, so I don't see the harm in having a consistent UX while we still depend upon it.

support/openharmony/entry/src/main/ets/pages/Index.ets
jschwe: > or the changes need to be in main perhaps?or the changes need to be in main perhaps?

That sounds like a likely explanation.

python/servo/package_commands.py
jschwe: Hmm, does the output folder in hvigor not differ between release / debug profiles? (To be fair I've only ever built with the default profile, which I think is debug)

python/servo/package_commands.py
jschwe: For HOS hvigor5 also seems to require the whole HOS commandline tools package, which I don't think we can currently download in CI. I'll raise that question internally.

python/servo/package_commands.py
mukilan: I'm not sure either. The output currently ends up inside `<top-level for app source>/entry/build/default/outputs/default` . Is one of the `default` in this path referring to the debug/release type? There is a [buildMode](https://github.com/servo/servo/blob/a5798ceadae9c3834db3afa03cbb477452128e73/support/openharmony/build-profile.json5#L20) declaration, but I don't follow how that translates to "default". Also, shouldn't the invocation of the `hvigor` differ for debug/release? It is currently the same for both.

mukilan: Just tried invoking hvigor with `-p buildMode=release` and while the build successfully completes successfully, it does give this warning
```
> hvigor Finished :servoshell:default@CompileResource... after 53 ms 
> hvigor WARN: Dependency libservoshell.so not found.

```
but the HAP does contain the `libservoshell.so`. The path is still the same as without `buildMode=release` i.e  `entry/build/default/outputs/default` so I guess the path doesn't seem to be affected by the buildMode.

mukilan: @jschwe the "dependency not found" error seems to originate from 'obfustcation-config-resolver.js' in the ohos-plugin module. I don't know why it says the .so is not found, but still the HAP has the library and it runs as expected. 

I've added code in [a797e9a](https://github.com/servo/servo/pull/33295/commits/a797e9a3b1d3782b269632e1731e5a1487e57ec3) . I've not removed the profile subfolder under `target/openharmony/<triple>` since the output path doesn't seem to change.

support/openharmony/entry/build-profile.json5
```diff
@@ -0,0 +1,37 @@
+{
+  "apiType": "stageMode",
+  "buildOption": {
+    "arkOptions": {
+      // "apPath": "./modules.ap"  /* Profile used for profile-guided optimization (PGO), a compiler optimization technique to improve app runtime performance. */
```
mrobinson, line 5: I guess this could be removed as well.

mrobinson (approved):

support/openharmony/entry/build-profile.json5
mukilan: I guess this is generated by the IDE. I've removed it in [ac570b1](https://github.com/servo/servo/pull/33295/commits/ac570b1037d07be31689e2819284ec4648e039ae).

python/servo/command_base.py
sagudev: mach will not be removed, but our plan is to make `mach build/check/cargo-*` commands more thin (ideally just cargo commands aliases with some options aliases).

PR #33297, opened 2024-09-03
Only handle most recent resize event in script thread

### LINKED ISSUE #33293: Sluggish resize behaviour on slow pages
**Describe the bug:**
When resizing a complex page (tested on  https://servo.zulipchat.com/, see https://html.spec.whatwg.org for a more extreme example) the resize events get queued up and the page continues to move around even after the window size stops changing.

https://github.com/user-attachments/assets/eb6ecc1d-aeaf-40cc-8b51-678335fc75d3

This happens because, in `update_the_rendering`, we process *each* resize event that occurred since the last `update_the_rendering` call:
https://github.com/servo/servo/blob/6c0394bc80963163f36dd273f8e6ed64b177dfa0/components/script/script_thread.rs#L1691-L1703

Is this necessary? From what I can see, the spec doesn't tell us to handle outdated resize events:

> 8. For each doc of docs, [run the resize steps](https://drafts.csswg.org/cssom-view/#document-run-the-resize-steps) for doc.
(https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering)

Here is what it would look like if all but the most recent resize event was thrown away:

https://github.com/user-attachments/assets/18694e53-c88e-404d-bfda-c5bf0aa1b634

I can PR this change, but I would like to first make sure that I'm not missing some important reason for why we need those events, since it seems like an intentional feature.

simonwuelker:
We don't need to bother with resize events that are outdated in the script thread.

This change removes the  resize event queue in `dom::Window` and only keeps track of the most recent one that has not been handled yet.

Refer to #33293 for an example of how can make resizing look snappier on pages where layout takes a while.

The queue was originally introduced to pass tests (https://github.com/servo/servo/pull/31505#discussion_r1592802909) but removing it does not seem to cause issues (https://github.com/Wuelle/servo/pull/1#issuecomment-2325682387)

---

- [X] These changes fix #33293

### DIFF
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -13,7 +13,7 @@ use std::rc::Rc;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
-use std::{cmp, env, mem};
+use std::{cmp, env};
 
 use app_units::Au;
 use backtrace::Backtrace;
@@ -216,9 +216,9 @@ pub struct Window {
     #[no_trace]
     devtools_marker_sender: DomRefCell<Option<IpcSender<Option<TimelineMarker>>>>,
 
-    /// Pending resize events, if any.
+    /// Most recent unhandled resize event, if any.
     #[no_trace]
-    resize_events: DomRefCell<Vec<(WindowSizeData, WindowSizeType)>>,
+    unhandled_resize_event: DomRefCell<Option<(WindowSizeData, WindowSizeType)>>,
 
     /// Parent id associated with this page, if any.
     #[no_trace]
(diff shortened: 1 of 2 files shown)


sagudev:
You can trigger try run in personal fork by running `mach try wpt-2020` instead of making a PR in own fork.

gterzian:
We cant try it a few times here as well, to check if there is anything that fails intermittently...

gterzian (approved):
Seems like a  reasonable change, because in my initial refactoring I changed the pending resize event behavior at https://github.com/servo/servo/commit/1d66ea2b2795cb7afcac787be1014f28dc7ad029#diff-a7136b19bf73401c3e2cd972eb84d50ba75421b57161eb067773597b35109c59L225  and it may not have been necessary. Thank you for looking into this. But let's run the wpt tests a few times just to make sure there is no intermittent failure that comes-up in the light of my comment in the initial PR(I wish I had noted down which test I thought was failing).

jdm:
Three successful times is giving me good vibes.

PR #33299, opened 2024-09-03
Fix intrinsic sizing of column flex containers

Loirooriol:
From https://drafts.csswg.org/css-flexbox-1/#intrinsic-cross-sizes,
> The min-content/max-content cross size of a single-line flex container
> is the largest min-content contribution/max-content contribution
> (respectively) of its flex items.

We were using the min/max-content size instead of the min/max-content contribution.

---


mrobinson (approved):
Nicely spotted!

PR #33302, opened 2024-09-03
Transform convert_* functions in gpuconvert.rs to From/TryFrom implementations

### LINKED ISSUE #33298: Transform `convert_*` functions in gpuconvert.rs to `From`/`TryFrom` implementations
See https://github.com/servo/servo/pull/33291/commits/5b582fca7355ee9dea094d6e254eab4a922e643b?diff=unified&w=1 for example

Taym95:
---

- [x] These changes fix #33298

- [x] These changes do not require tests because there is not behavior changes.

### DIFF
--- a/components/script/dom/gpuconvert.rs
+++ b/components/script/dom/gpuconvert.rs
@@ -21,375 +21,384 @@ use crate::dom::bindings::codegen::Bindings::WebGPUBinding::{
 use crate::dom::bindings::error::Fallible;
 use crate::dom::types::GPUDevice;
 
-pub fn convert_texture_format(format: GPUTextureFormat) -> wgt::TextureFormat {
-    match format {
-        GPUTextureFormat::R8unorm => wgt::TextureFormat::R8Unorm,
-        GPUTextureFormat::R8snorm => wgt::TextureFormat::R8Snorm,
-        GPUTextureFormat::R8uint => wgt::TextureFormat::R8Uint,
-        GPUTextureFormat::R8sint => wgt::TextureFormat::R8Sint,
-        GPUTextureFormat::R16uint => wgt::TextureFormat::R16Uint,
-        GPUTextureFormat::R16sint => wgt::TextureFormat::R16Sint,
-        GPUTextureFormat::R16float => wgt::TextureFormat::R16Float,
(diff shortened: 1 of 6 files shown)


components/script/dom/gpuconvert.rs
```diff
@@ -556,26 +575,32 @@ pub fn convert_bind_group_layout_entry(
...
-                })
-            }
-        },
-        GPUColor::GPUColorDict(d) => Ok(wgt::Color {
-            r: *d.r,
-            g: *d.g,
-            b: *d.b,
-            a: *d.a,
-        }),
+use std::convert::TryFrom;
```
sagudev, line 578: This use should at top of the file (I wonder why fmt does not says anything).

sagudev:
What was your reasoning when to use `::from(_)` vs `.into()`?

sagudev:
> ⚠️ Try run ([#10689902820](https://github.com/servo/servo/actions/runs/10689902820)) failed.

Only crashy flakes so this is ok.

Taym95:
> What was your reasoning when to use `::from(_)` vs `.into()`?

removed them all

components/script/dom/gpuconvert.rs
```diff
@@ -409,68 +418,78 @@ pub fn convert_store_op(op: Option<GPUStoreOp>) -> wgpu_com::StoreOp {
...
+                .transpose()?
+                .unwrap_or_default(),
+            aspect: match ic_texture.aspect {
+                GPUTextureAspect::All => wgt::TextureAspect::All,
+                GPUTextureAspect::Stencil_only => wgt::TextureAspect::StencilOnly,
+                GPUTextureAspect::Depth_only => wgt::TextureAspect::DepthOnly,
+            },
+        })
+    }
 }
```
Taym95, line 494: @sagudev could not convert some function because types are defined outside this crate, like `convert_bind_group_layout_entry`.

sagudev: I think in such cases it should be still possible to implement [`Into`](https://doc.rust-lang.org/std/convert/trait.Into.html), but we can do that in a followup.

sagudev (approved):

PR #33305, opened 2024-09-03
Fix devtool crashs after clicking `Enable connection prompt`

### LINKED ISSUE #33304: Clicking on `Enable connection prompt` in Firefox crashes DevTools.
**Describe the bug:**
Clicking on `Enable connection prompt` on Firefox crashs  Servo Devtool

**To Reproduce:**
Run Servo with Devtool `./mach run -d --devtools=6080` then open Firefox connect then click `Enable connection prompt`

image

**Platform:**
macOS

`handle_message` expects `msg.get("value")` to always return a `String`, but clicking on `Enable connection prompt` sends `"value": Bool(true)`, which causes this line to crash.

https://github.com/servo/servo/blob/e857cdf022f319ae3fbd54e10e279ab43485995b/components/devtools/actors/preference.rs#L40

Taym95:
---

- [x] These changes fix #33304
- [x] These changes do not have tests, because there are no DevTools tests.

### DIFF
--- a/components/devtools/actors/preference.rs
+++ b/components/devtools/actors/preference.rs
@@ -38,28 +38,28 @@ impl Actor for PreferenceActor {
         stream: &mut TcpStream,
         _id: StreamId,
     ) -> Result<ActorMessageStatus, ()> {
-        if let Some(mut key) = msg.get("value").unwrap().as_str() {
-            // Mapping to translate a Firefox preference name onto the corresponding Servo preference name
-            let pref_name_mapping: HashMap<&str, &str> =
-                [("dom.serviceWorkers.enabled", "dom.serviceworker.enabled")]
-                    .iter()
-                    .copied()
-                    .collect();
-            if pref_name_mapping.contains_key(key) {
-                key = pref_name_mapping.get(key).unwrap();
-            }
-
-            let pref_value = pref_map().get(key);
-            Ok(handle_preference_value(
-                pref_value,
-                self.name(),
-                msg_type,
-                stream,
-            ))
-        } else {
+        let Some(mut key) = msg.get("value").and_then(|v| v.as_str()) else {
             warn!("PreferenceActor: handle_message: value is not a string");
             return Ok(ActorMessageStatus::Ignored);
+        };
+
+        // Mapping to translate a Firefox preference name onto the corresponding Servo preference name
+        let pref_name_mapping: HashMap<&str, &str> =
+            [("dom.serviceWorkers.enabled", "dom.serviceworker.enabled")]
+                .iter()
+                .copied()
+                .collect();
+        if pref_name_mapping.contains_key(key) {
+            key = pref_name_mapping.get(key).unwrap();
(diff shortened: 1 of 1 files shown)


mrobinson:
" These changes do not require tests because no behavior change."

I think these changes do change behavior. I'll just change this to say that there aren't tests ATM for devtools.

components/devtools/actors/preference.rs
```diff
@@ -37,7 +38,13 @@ impl Actor for PreferenceActor {
...
         _id: StreamId,
     ) -> Result<ActorMessageStatus, ()> {
-        let mut key = msg.get("value").unwrap().as_str().unwrap();
+        let mut key = match msg.get("value").unwrap() {
+            Value::String(value) => value.as_str(),
+            _ => {
+                warn!("PreferenceActor: handle_message: value is not a string");
+                return Ok(ActorMessageStatus::Ignored);
+            },
+        };
```
mrobinson, line 47: Perhaps this could be a bit simpler?

```suggestion
        let mut Some(key) = match msg.get("value").unwrap().as_str() else {
            warn!("PreferenceActor: handle_message: value is not a string");
            return Ok(ActorMessageStatus::Ignored);
        };
```

Taym95: I tried it, but doesn’t it seem like the correct syntax?

jdm: Yeah, the match keyword needs to be removed.

components/devtools/actors/preference.rs
```diff
@@ -37,25 +38,28 @@ impl Actor for PreferenceActor {
...
-        // Mapping to translate a Firefox preference name onto the corresponding Servo preference name
-        let pref_name_mapping: HashMap<&str, &str> =
-            [("dom.serviceWorkers.enabled", "dom.serviceworker.enabled")]
-                .iter()
-                .copied()
-                .collect();
-        if pref_name_mapping.contains_key(key) {
-            key = pref_name_mapping.get(key).unwrap();
-        }
+        if let Some(mut key) = msg.get("value").unwrap().as_str() {
```
Taym95, line 41: @mrobinson do you think we can merge this one?

components/devtools/actors/preference.rs
```diff
@@ -37,25 +38,28 @@ impl Actor for PreferenceActor {
...
+            Ok(handle_preference_value(
+                pref_value,
+                self.name(),
+                msg_type,
+                stream,
+            ))
+        } else {
+            warn!("PreferenceActor: handle_message: value is not a string");
+            return Ok(ActorMessageStatus::Ignored);
+        }
```
mrobinson, line 62: I think it makes a bit more sense to have this as an early return rather than wrapping the entire logic of the function in the error condition check.

Taym95: done, thanks!

mrobinson (approved):
Thanks!

Issue #33292, opened 2024-09-02
jsstring_to_str should accept a NonNull argument for the js string


jdm:
It already asserts that the argument is not null, so we should force callers to handle that case by making the compiler verify it for us.

last-genius:
Going to try to take this as my first issue. To make sure I understand this (and the project's attitude to this in general) correctly - judging by the assert in `jsstring_to_str` - null pointers shouldn't really be expected there already, therefore it would be alright for the callers to use `NonNull::new_unchecked()`? Or should the `panic` shift from the assertion into an `unwrap` on `::new()` in the callers?

Thank you in advance!

jdm:
We should definitely require callers to make an explicit choice. Using unwrap in the callers as a first step would be great!

jdm:
For context, I filed this based on the stack trace in #33288 where the assertion fired due to a null pointer being passed in.

last-genius:
Is there anything I could read on the correct usage of the `rooted!` macro? A bit struggling with the following code, wondering if the diff would still work or if I should register something as a GC root. `NonNull` is a bit awkward for a simple translation here:

```diff
diff --git a/components/script/dom/console.rs b/components/script/dom/console.rs
index ae041dfa7d..dd902d26fc 100644
--- a/components/script/dom/console.rs
+++ b/components/script/dom/console.rs
@@ -71,12 +71,10 @@ where
 
 #[allow(unsafe_code)]
 unsafe fn handle_value_to_string(cx: *mut jsapi::JSContext, value: HandleValue) -> DOMString {
-    rooted!(in(cx) let mut js_string = std::ptr::null_mut::<jsapi::JSString>());
-    js_string.set(JS_ValueToSource(cx, value));
-    if js_string.is_null() {
-        return "<error converting value to string>".into();
+    match std::ptr::NonNull::new(JS_ValueToSource(cx, value)) {
+        Some(js_str) => jsstring_to_str(cx, *js_str),
+        None => "<error converting value to string>".into(),
     }
-    jsstring_to_str(cx, *js_string)
 }
```

jdm:
In that particular case, you should be able to keep the rooted! use where it is, then call set in the some branch and follow it with jsstring_to_str.

jschwe:
the associated PR is merged, so is this issue fixed now, or is there more work remaining?

jdm:
Nope, that was all.

PR #33306, opened 2024-09-03
`jsstring_to_str` should accept a `NonNull` argument for the JS string

last-genius:
Instead of asserting the raw pointer is not null, force callers to produce a NonNull pointer.

---
- [x] These changes address the suggestion in #33292
- [x] These changes do not require tests because they do not change the behavior, just move the existing verification into the callers by changing types

Please let me know if there's something else I should do or if I missed anything - this is my first PR here.


jdm (approved):
Looks good! I like that this change makes this method impossible to misuse.

PR #33312, opened 2024-09-04
Layout: Implement innerText/outerText

Melchizedek6809:
Hey, I've given the PR by @shanehandley over in #32427 the final polish, now it passes most of the Tests. I've had to diverge a bit from the algorithm as described in the whatwg spec since it only described a simplified version that would not give the same results as other browsers/WPT (it mainly left out a lot of special cases).

---

- [X] These changes do not require tests because WPT


components/layout_2020/query.rs
```diff
@@ -509,8 +514,415 @@ fn is_eligible_parent(fragment: &BoxFragment) -> bool {
...
+                    style.clone_text_transform().case(),
+                )
+                .collect::<String>();
+
+                let is_preformatted_element =
+                    white_space_collapse == WhiteSpaceCollapseValue::Preserve;
+
+                let is_final_character_whitespace = transformed_text
+                    .chars()
+                    .last()
```
Wuelle, line 716: Minor thing, but `.chars().last()` is O(n) since it needs to to iterate over all the previous characters, you could use [`next_back()`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#tymethod.next_back) for O(1) instead.

Melchizedek6809: Ah thanks! I just changed it to use `next_back` instead.

components/layout_2020/query.rs
```diff
@@ -509,8 +514,415 @@ fn is_eligible_parent(fragment: &BoxFragment) -> bool {
...
+                if transformed_text.len() > 0 {
+                    // Here we decide whether to keep or truncate the final white
+                    // space character, if there is one.
+                    if is_final_character_whitespace && !is_preformatted_element {
+                        state.may_start_with_whitespace = false;
+                        state.did_truncate_trailing_white_space = true;
+                        let truncated_text = transformed_text
+                            .chars()
+                            .take(transformed_text.len() - 1)
+                            .collect();
```
Wuelle, line 742: [`String::pop`](https://doc.rust-lang.org/std/string/struct.String.html#method.pop) also removes the last character, but without (necessarily) allocating a new string. I think this could be used here?

Melchizedek6809: Nice, works great

components/script/dom/htmlelement.rs
```diff
@@ -467,52 +469,90 @@ impl HTMLElementMethods for HTMLElement {
...
+        // Step 1: Let fragment be the rendered text fragment for value given element's node
+        // document.
+        let fragment = self.rendered_text_fragment(input);
+
+        // Step 2: Replace all with fragment within element.
+        Node::replace_all(Some(fragment.upcast()), self.upcast::<Node>());
+    }
+
+    /// <https://html.spec.whatwg.org/multipage/#dom-outertext>
+    fn GetOuterText(&self) -> Fallible<DOMString> {
```
gterzian, line 483: I think it's worth adding a method for [`get-the-text-steps`](https://html.spec.whatwg.org/multipage/dom.html#get-the-text-steps), and use it here and within `GeInnerText`.

components/script/dom/htmlelement.rs
```diff
@@ -467,52 +469,90 @@ impl HTMLElementMethods for HTMLElement {
...
+
+        window.layout_reflow(QueryMsg::ElementOuterTextQuery);
+        let text = window
+            .layout()
+            .query_element_outer_text(node.to_trusted_node_address());
+
+        Ok(DOMString::from(text))
+    }
+
+    /// <https://html.spec.whatwg.org/multipage/#the-innertext-idl-attribute%3Adom-outertext-2>
```
gterzian, line 502: https://html.spec.whatwg.org/multipage/#the-innertext-idl-attribute:dom-outertext-2

components/script/dom/htmlelement.rs
```diff
@@ -467,52 +469,90 @@ impl HTMLElementMethods for HTMLElement {
...
-                '\u{000A}' | '\u{000D}' => {
-                    if ch == '\u{000D}' && position.peek() == Some(&'\u{000A}') {
-                        // a \r\n pair should only generate one <br>,
-                        // so just skip the \r.
-                        position.next();
-                    }
+            fragment
+                .upcast::<Node>()
+                .AppendChild(text_node.upcast())
+                .unwrap();
```
gterzian, line 530: We should either `expect` with the appropriate message if this cannot fail at this point, or return the error(even though the spec is silent on that).

components/script/dom/htmlelement.rs
```diff
@@ -467,52 +469,90 @@ impl HTMLElementMethods for HTMLElement {
...
+        }
 
-                    if !text.is_empty() {
-                        append_text_node_to_fragment(&document, &fragment, text);
-                        text = String::new();
-                    }
+        // Step 6: Replace this with fragment within this's parent.
+        if let Err(error) = parent.ReplaceChild(fragment.upcast(), node) {
+            warn!("Failed to replace child node: {:?}", error);
```
gterzian, line 535: Here you can return the error, since the spec throws.

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +937,82 @@ impl HTMLElement {
...
+    fn rendered_text_fragment(&self, input: DOMString) -> DomRoot<DocumentFragment> {
+        // Step 1: Let fragment be a new DocumentFragment whose node document is document.
+        let document = document_from_node(self);
+        let fragment = DocumentFragment::new(&document);
+
+        // Step 2: Let position be a position variable for input, initially pointing at the start
+        // of input.
+        let mut position = input.chars().peekable();
+
+        // Step : Let text be the empty string.
```
gterzian, line 951: number

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +937,82 @@ impl HTMLElement {
...
+        let mut text = String::new();
+
+        // Step 4
+        while let Some(ch) = position.next() {
+            match ch {
+                // While position is not past the end of input, and the code point at position is
```
gterzian, line 957: We have some code at https://github.com/servo/servo/blob/ebed9218f2907767ba3c9dd9f27f30a6a6e9f225/components/net/fetch/headers.rs#L88

It might make sense to move it to `components/shared/net` and re-use it here.

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +937,82 @@ impl HTMLElement {
...
+        // document is document to fragment.
+        if !text.is_empty() {
+            append_text_node_to_fragment(&document, &fragment, text);
+        }
+
+        fragment
+    }
+
+    /// <https://html.spec.whatwg.org/multipage/#merge-with-the-next-text-node>
+    fn merge_with_the_next_text_node(&self, node: DomRoot<Node>) {
```
gterzian, line 992: We could make this faillible, instead of calling it inside the `node.is::<Text>()` condition(or you `expect`  below and explain why it cannot fail).

gterzian (changes requested):
Thanks for following the spec in such detail! Couple of points on the `script` side of things...

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +937,82 @@ impl HTMLElement {
...
+                    // Collect a sequence of code points that are not U+000A LF or U+000D CR from
+                    // input given position, and set text to the result.
+                    text.push(ch);
+                },
+            }
+        }
+
+        // If text is not the empty string, then append a new Text node whose data is text and node
+        // document is document to fragment.
+        if !text.is_empty() {
```
gterzian, line 984: I'm confused about the loop, do you still need this here besides what is at [966](https://github.com/servo/servo/pull/33312/files#diff-978993aad649a5f5ef459e8071ed6b982d038e71ca6d53156e9dcdf74192c702R966)?

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +937,82 @@ impl HTMLElement {
...
+
+                if let Err(error) =
+                    node_chars.ReplaceData(node_chars.Length(), 0, next_chars.Data())
+                {
+                    warn!("Failed to replace character data: {:?}", error);
+
+                    return;
+                }
+
+                // Step 4: If next's parent is non-null, then remove next.
```
gterzian, line 1009: I don't see the parent condition in the spec, do you need it?

components/script/dom/htmlelement.rs
Melchizedek6809: Makes sense, they were treated separately until they hit Layout where both ended up calling the same method, I've added a `get_inner_outer_text` method to HTMLElement which is called by both `GetInnerText` and `GetOuterText`, I didn't want call this method `get_the_text_steps` since the actual implementation is in Layout, where the method is called just that.

components/script/dom/htmlelement.rs
Melchizedek6809: Not sure about making this fallible, from the spec it doesn't sound like an operation that I'd expect to fail. It seems more like a clean up step to make sure that we don't have multiple Text nodes next to each other, which is already the case if either `node` or `next` are something other than Text nodes, or there is no `next`.

Maybe we should also call `unwrap` on the Result of the call to `ReplaceData`, since from what I can tell it only fails when fed invalid UTF-16, and I'm not sure about the architecture of Servo but invalid UTF-16 probably shouldn't make its way into the DOM at all, so we should probably fail as soon as possible in that case, or what do you think?

Melchizedek6809: I've tried to rewrite/clean it up a bit, what do you think about this version?

components/script/dom/htmlelement.rs
Melchizedek6809: Not sure whether the latest changes broke the line numbers, do you mean why there's the following code both inside the `while` loop and at the end?
```rust
if !text.is_empty() {
    append_text_node_to_fragment(&document, &fragment, text);
    text = String::new();
}
```

From what I understand that's because linefeeds should be converted to `<br>` tags, which means we might need to generate multiple text nodes. And since we might not end on a linefeed we have to put the accumulated characters into a final text node.

components/script/dom/htmlelement.rs
Melchizedek6809: Can probably be removed, since remove_self already makes sure there's a parent

components/script/dom/htmlelement.rs
Melchizedek6809: I fear then we'd have to iterate over the string twice since we also need to turn linefeeds into `<br>` elements and generate multiple Text nodes

Melchizedek6809:
@gterzian Thank you for the review! I still had some questions regarding some points in your review which I've added as comments, apart from that most things should be implemented now

mrobinson (commented):
I'll review this soon, but I have one question:

Does this need to be a layout query? It looks like it could simply run in the context of the script thread?

Melchizedek6809:
Not sure if I understand correctly but as far as I can tell the code runs on the script thread with a couple of function calls in-between to abstract over the used layout engine.  As for why it's part of the layout_2020 crate: the code is making use of both `TextTransformation` and `WhitespaceCollapse` which seem to only be available in layout_2020, although I might be mistaken since I'm not familiar with the codebase.

mrobinson:
> Not sure if I understand correctly but as far as I can tell the code runs on the script thread with a couple of function calls in-between to abstract over the used layout engine. As for why it's part of the layout_2020 crate: the code is making use of both `TextTransformation` and `WhitespaceCollapse` which seem to only be available in layout_2020, although I might be mistaken since I'm not familiar with the codebase.

Apologies, I used the wrong terminology because layout and script used to run on separate threads. They don't any longer. What I mean is that this code is just iterating through DOM nodes and doesn't need access to laid out content. It does use layout's implementation of the text transformation rules, but those are utilities used in the process of layout. 

Do you foresee a further implementation of this API to need access to the laid out content of nodes? For instance, this might include knowing which nodes aren't included in a layout tree due to `display: none` or knowing about the details of `display: content`? If not, maybe it makes sense to move the implementation of this to the script trait so that it can be shared between both old and new layout and can be run without having to wait for a pending layout. We can think about where best to move the text transformation utilities.

Melchizedek6809:
Ah, thank you for clearing that up!

We might need some additional information to properly apply `text-transform` on `:first-line` selectors, but other than that we don't need to concern ourselves too much with layout internals.

Apart from `script`, is there another crate we could move this implementation to? Just because it is already pretty big and slow to compile, so I'm a bit hesitant about moving more code in there than absolutely necessary.

Oh and do you know about the long term plans for layout_2013? I think I've read somewhere that there are plans for its removal now that layout_2020 is more mature. Just thinking out loud whether sharing the implementation between the 2 layout engines is worth the effort if the former is about to be removed anyways.

components/script/dom/htmlelement.rs
gterzian: Ok, thanks for clarifying.

gterzian (approved):
Thanks for addressing the review, only some nits remains.

components/script/dom/htmlelement.rs
```diff
@@ -897,6 +916,84 @@ impl HTMLElement {
...
+            None => return,
+        };
+
+        // Step 2: If next is not a Text node, then return.
+        if !next.is::<Text>() {
+            return;
+        }
+        // Step 3: Replace data with node, node's data's length, 0, and next's data.
+        let node_chars = node.downcast::<CharacterData>().unwrap();
+        let next_chars = next.downcast::<CharacterData>().unwrap();
```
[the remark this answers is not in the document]
Melchizedek6809: Ah alright, I've changed these accordingly

Melchizedek6809:
@gterzian Thanks for the review, those nits should now be resolved in with the newest commit.

mrobinson (approved):
I think for now this can stay in the `layout` crate, as it may be that in the future this can be implemented against the display list instead of the DOM. Alternatively, it can be moved to `script` in a followup.

mrobinson:
Apologies for taking so long to review this one!

mrobinson:
There's a couple failures here:

```
OK /html/dom/elements/the-innertext-and-outertext-properties/getter.html

    FAIL [expected PASS] subtest: Whitespace around inline-flex should not be collapsed ("&lt;div&gt;abc &lt;span style='display:inline-flex'&gt;&lt;/span&gt; def")

    assert_equals: innerText expected "abc  def" but got "abc def"

FAIL [expected PASS] subtest: Whitespace around inline-grid should not be collapsed ("&lt;div&gt;abc &lt;span style='display:inline-grid'&gt;&lt;/span&gt; def")

assert_equals: innerText expected "abc  def" but got "abc def"

FAIL [expected PASS] subtest: Leading/trailing space removal at display:inline-flex boundary ("&lt;div&gt;123&lt;span style='display:inline-flex'&gt; abc &lt;/span&gt;def")

assert_equals: innerText expected "123abcdef" but got "123 abc def"

FAIL [expected PASS] subtest: Leading/trailing space removal at display:inline-grid boundary ("&lt;div&gt;123&lt;span style='display:inline-grid'&gt; abc &lt;/span&gt;def")

assert_equals: innerText expected "123abcdef" but got "123 abc def"
```

Melchizedek6809:
> Apologies for taking so long to review this one!

No worries, thank you for the review!

> There's a couple failures here:

Alright, those seem to be the new WPT Tests I've added upstream (added a bunch in the process of writing this PR, since a couple of edge-cases weren't tested for before). Shouldn't take long to update this PR to make them pass.

Melchizedek6809:
Alright, I've added support for inlineFlex and inlineGrid and updated the WPT results. Although the inlineGrid tests still fail because it seems that grid support is still enabled (when trying to enable it servo just paniced on running the WPT suite, I suppose grid is not quite stable yet in Servo). Once grid stabilizes and the display value will actually become `Display:InlineGrid` those 2 tests should pass.

PR #33313, opened 2024-09-04
net: Add range support to `DataProtocolHandler`

### LINKED ISSUE #32645: Video element with base64 encoded data URL of valid video file contents causes a media error
This code breaks tests: https://github.com/servo/servo/blob/main/tests/wpt/tests/html/canvas/element/manual/imagebitmap/common.sub.js#L56-L78. We get an error from gstreamer that looks like this:
```
 0:03.16 pid:96913 [2024-06-29T16:00:09Z ERROR script::dom::htmlmediaelement] Player error: "Error from element /GstPlayBin:playbin/GstURIDecodeBin:uridecodebin0/GstDecodeBin:decodebin0/avdec_h264:avdec_h264-0: Could not decode stream.\nNo valid frames decoded before end of stream\n../gst-libs/gst/video/gstvideodecoder.c(1421): gst_video_decoder_sink_event_default (): /GstPlayBin:playbin/GstURIDecodeBin:uridecodebin0/GstDecodeBin:decodebin0/avdec_h264:avdec_h264-0:\nno valid frames found"
```
This is observable in `tests/wpt/tests/html/canvas/element/manual/imagebitmap/createImageBitmap-invalid-args.html`.

crbrz:
Media player can request data with an offset and will cause an error if all data is returned instead.

---

- [X] These changes fix #32645

### DIFF
--- a/components/net/protocols/data.rs
+++ b/components/net/protocols/data.rs
@@ -2,11 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-use std::future::Future;
+use std::future::{ready, Future};
 use std::pin::Pin;
 
 use data_url::forgiving_base64;
-use headers::{ContentType, HeaderMapExt};
+use headers::{ContentType, HeaderMapExt, Range};
 use http::StatusCode;
 use mime::Mime;
 use net_traits::request::Request;
@@ -16,6 +16,7 @@ use percent_encoding::percent_decode;
 use servo_url::ServoUrl;
 use url::Position;
 
+use super::{get_range_request_bounds, partial_content, range_not_satisfiable_error};
 use crate::fetch::methods::{DoneChannel, FetchContext};
 use crate::protocols::ProtocolHandler;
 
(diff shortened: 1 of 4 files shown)


jdm (approved):
The implementation looks reasonable, but I'm concerned that there do not appear to be any WPT tests for this feature (https://searchfox.org/mozilla-central/source/testing/web-platform/tests/html/canvas/element/manual/imagebitmap/createImageBitmap-colorSpaceConversion.html does not appear to involve data URLs at all). Can we add one?

jdm:
It's also odd that I see https://fetch.spec.whatwg.org/#ref-for-build-a-content-range used by the blob scheme fetch steps in https://fetch.spec.whatwg.org/#concept-scheme-fetch, but not for the data scheme fetch steps.

jdm:
Oh, I didn't look at the original issue before reading this pull request. So this must be a feature of our media stack; very interesting.

jdm:
I think we should verify what other browsers do when performing a fetch() on a data: URL with a range header present. If it doesn't cause a subset of the response to be returned, this isn't something we should expose as a feature in Servo's network stack or we introduce a new browser incompatibility.

crbrz:
This
```html
<!DOCTYPE html>
<script>
	const url = 'data:text/plain,hello%20world';
	fetch(url, { headers: { 'Range': 'bytes=6-' } }).then((response) => {
		response.text().then((text) => {
			console.log(text);
		});
	});	
</script>
```
in Chromium and Firefox prints `hello world` instead of `world`. It looks like range header is not supported for `data:` URLs.

Since we get all media data in one chunk when fetching from a `data:` URL (without this PR) an alternative would be slicing it to what the player requested in here: https://github.com/servo/servo/blob/642c25d9a73b2c3fbe6f98a32bae5c7bdb6c448f/components/script/dom/htmlmediaelement.rs#L2759-L2815
but it feels hacky.

crbrz:
Closing because these changes don't follow the spec for `data:` URLs.

PR #33316, opened 2024-09-04
minibrowser: Reset the location field when switching tabs

### LINKED ISSUE #33311: Minibrowser: Removing URL from new tab also clears urls in other tabs
**Describe the bug:**
Removing URL from new tab also clears urls in other tabs

**To Reproduce:**
- open minibrowser
- open a new tab
- delete `servo:newtab`
- check the url of previously opened tabs

**Platform:**
Tested on MacOS. Not sure if happening on other platform as well

image

webbeef:
The state of the location bar was not reset when switching tabs, hence it would stick to the currently edited value when it was being edited.

---

- [X] These changes fix #33311(GitHub issue number if applicable)

### DIFF
--- a/ports/servoshell/desktop/minibrowser.rs
+++ b/ports/servoshell/desktop/minibrowser.rs
@@ -366,6 +366,7 @@ impl Minibrowser {
                             if let Some(event) =
                                 Self::browser_tab(ui, label, webview.focused, webview_id)
                             {
+                                location_dirty.set(false);
                                 embedder_events.push(event);
                             }
                         }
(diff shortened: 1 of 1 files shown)


ports/servoshell/desktop/minibrowser.rs
```diff
@@ -366,6 +366,7 @@ impl Minibrowser {
                             if let Some(event) =
                                 Self::browser_tab(ui, label, webview.focused, webview_id)
                             {
+                                location_dirty.set(false);
```
atbrakhi, line 369: While this fixes the reported issue, I find this area a bit buggy. For example, now with below steps url is empty when we open new tab.

- open a new tab
- delete `servo:newtab`
- open new tab / `cmd + T`
- url is empty
- go to previous tab and then go back to new tab again url is not empty anymore

atbrakhi (approved):

ports/servoshell/desktop/minibrowser.rs
atbrakhi: Indeed. Would be nice to find a better solution but for now i think it's fine to land this

PR #33319, opened 2024-09-05
servoshell: add keyboard shortcuts for switching tabs

Melchizedek6809:
This PR adds Cmd/Ctrl + [1-9] for selecting a particular tab and Ctrl + PageUp/Down for selecting the previous/next tab. Thought this might be convenient since they are present in all major browsers (only safari seems to use different keys for switching to the previous/next tab)

---

- [X] These changes do not require tests because UI

### DIFF
--- a/ports/servoshell/desktop/webview.rs
+++ b/ports/servoshell/desktop/webview.rs
@@ -414,6 +414,17 @@ where
         self.shutdown_requested
     }
 
+    fn focus_webview_by_index(&self, index: usize) -> Option<EmbedderEvent> {
+        Some(EmbedderEvent::FocusWebView(self.webviews().get(index)?.0))
+    }
+
+    fn get_focused_webview_index(&self) -> Option<usize> {
+        let focused_id = self.focused_webview_id?;
+        self.webviews()
+            .iter()
+            .position(|webview| webview.0 == focused_id)
+    }
+
     /// Handle key events before sending them to Servo.
     fn handle_key_from_window(&mut self, key_event: KeyboardEvent) {
         let embedder_event = ShortcutMatcher::from_event(key_event.clone())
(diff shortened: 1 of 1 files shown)


ports/servoshell/desktop/webview.rs
```diff
@@ -414,6 +414,29 @@ where
         self.shutdown_requested
     }
 
+    fn focus_webview_by_index(&self, index: usize) -> Option<EmbedderEvent> {
+        let webviews = &self.webviews();
+        if index < webviews.len() {
```
mukilan, line 419: I think we can simplify this if/else using the `?` operator

```
Some(EmbedderEvent::FocusWebView(webviews.get(index)?.0))
```

ports/servoshell/desktop/webview.rs
```diff
@@ -414,6 +414,29 @@ where
...
+            Some(EmbedderEvent::FocusWebView(webviews[index].0))
+        } else {
+            None
+        }
+    }
+
+    fn get_focused_webview_index(&self) -> Option<usize> {
+        let focused_id = match self.focused_webview_id {
```
mukilan, line 427: The match can be simplified to
```
let focused_id = self.focused_webview_id?;
```

ports/servoshell/desktop/webview.rs
```diff
@@ -414,6 +414,29 @@ where
...
+            Some(id) => id,
+            None => return None,
+        };
+        let webviews = &self.webviews();
+        for i in 0..webviews.len() {
```
mukilan, line 432: We can replace the for loop with

```
webviews.iter().position(|webview| webview.0 == focused_id)
```

Melchizedek6809:
@mukilan Thank you for the review! Incorporated your advice and the code is quite a bit shorter/cleaner now

mukilan (approved):

PR #33320, opened 2024-09-05
webgpu: Move actual Create* implementations from `GPUDevice` to Self

sagudev:
This just makes sense and matches spec more (ex: `CreateBuffer` has written steps at `GPUBuffer` section not at `GPUDevice` section). Firefox also does the same: https://searchfox.org/mozilla-central/rev/5959ec6b84d66592a77a3e5e2d2aedc1b3e7d4c5/dom/webgpu/Device.cpp#156

Changes are mostly mechanical (just moving stuff from gpudevice.rs).

---

### DIFF
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -479,7 +479,8 @@ impl GPUDeviceMethods for GPUDevice {
         &self,
         descriptor: &GPURenderPipelineDescriptor,
     ) -> Fallible<DomRoot<GPURenderPipeline>> {
-        let render_pipeline = GPURenderPipeline::create(self, descriptor, None)?;
+        let (implicit_ids, desc) = self.parse_render_pipeline(&descriptor)?;
+        let render_pipeline = GPURenderPipeline::create(self, implicit_ids, desc, None)?;
         Ok(GPURenderPipeline::new(
             &self.global(),
             render_pipeline,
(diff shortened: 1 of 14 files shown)


components/script/dom/gpudevice.rs
```diff
@@ -866,27 +494,11 @@ impl GPUDeviceMethods for GPUDevice {
         descriptor: &GPURenderPipelineDescriptor,
         comp: InRealm,
     ) -> Fallible<Rc<Promise>> {
-        let (implicit_ids, desc) = self.parse_render_pipeline(&descriptor)?;
-
         let promise = Promise::new_in_current_realm(comp);
+        // FIXME: if ``GPURenderPipeline::create` fails we should remove IpcReceiver from set
+        // but this is currently not possible
```
sagudev, line 499: [`IpcReceiverSet`](https://docs.rs/ipc-channel/latest/ipc_channel/ipc/struct.IpcReceiverSet.html) does not support removing. In webgpu we create new receiver for every `request_async` and they are only used once (sometimes twice but that's wrong).

gterzian: Why don't you move the call to `device.parse_render_pipeline` to here so you can fail before creating the ipc route?

gterzian (approved):
One question, otherwise LGTM

components/script/dom/gpudevice.rs
sagudev: Good idea, will do but it would still be good idea to remove receivers that will not be used anymore.

PR #33321, opened 2024-09-05
CI: use self-hosted runners for Linux build jobs

delan:
This patch makes our Linux builds use [our self-hosted runners](https://github.com/servo/ci-runners) where possible, falling back to the slower GitHub-hosted runners only when none are available, cutting Linux build jobs from over half an hour to [under 8 minutes](https://github.com/servo/servo/actions/runs/10717467096/job/29717238111). Since there are now two jobs that use self-hosted runners, we extract the shared logic into workflow calls.

We also improve the monitor API error handling and tweak the monitor API timeouts a bit, with a shorter timeout for the server being completely down (5 seconds) and a longer timeout for the server to respond under contention (30 seconds). Reserving a runner can take a few seconds, so we don’t want to give up too quickly.

**If self-hosted runners break the build after landing, please disable them by [creating a repository variable](https://github.com/servo/servo/settings/variables/actions) named NO_SELF_HOSTED_RUNNERS with any non-empty value, then let me know on Zulip or mention me in a GitHub issue.** Disabling them should be more reliable now that we no longer use job concurrency (#33315).

---

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -67,8 +67,12 @@ jobs:
     uses: ./.github/workflows/self-hosted-runner-select.yml
     secrets: inherit
     with:
+      # Ubuntu 22.04 has glibc 2.34 so the binaries produced
+      # won't run on systems with older glibc e.g wpt.fyi
+      # runners which still use 20.04.
       github-hosted-runner-label: ubuntu-${{ inputs.upload && '20.04' || '22.04' }}
       self-hosted-image-name: servo-ubuntu2204
+      force-github-hosted-runner: ${{ inputs.upload }}
   runner-timeout:
     needs:
       - runner-select
@@ -82,9 +86,6 @@ jobs:
     needs:
       - runner-select
     name: Linux Build [${{ needs.runner-select.outputs.unique-id }}]
-    # Ubuntu 22.04 has glibc 2.34 so the binaries produced
-    # won't run on systems with older glibc e.g wpt.fyi
-    # runners which still use 20.04.
     runs-on: ${{ needs.runner-select.outputs.selected-runner-label }}
     steps:
       - if: ${{ ! fromJSON(needs.runner-select.outputs.is-self-hosted) }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -156,7 +156,7 @@ jobs:
         with:
           timeout_minutes: 20
           max_attempts: 2 # https://github.com/servo/servo/issues/30683
-          command: python3 ./mach test-unit --${{ inputs.profile }}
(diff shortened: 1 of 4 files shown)


.github/workflows/linux.yml
```diff
@@ -110,7 +142,7 @@ jobs:
         with:
           timeout_minutes: 20
           max_attempts: 2 # https://github.com/servo/servo/issues/30683
-          command: python ./mach test-unit --${{ inputs.profile }}
+          command: python3 ./mach test-unit --${{ inputs.profile }}
```
sagudev, line 145: SIDE-NOTE: Installing `python-is-python3` package should fix this in ubuntu.

.github/workflows/use-self-hosted.yml
sagudev, line 1: we need to make sure we do not hit workflow_call depth limit

sagudev: ☺️

.github/workflows/linux.yml
delan: done (57a3b502de74cccd51094de77df4840b339eec32) and deployed ([bab8615c411f8](https://github.com/servo/ci-runners/commit/bab8615c411f8641473a7d3f136dd215cd46df2d))

delan:
Yes, that's right. I don't think we have the capacity (yet) to run WPT jobs on self-hosted runners.

sagudev (approved):

.github/workflows/self-hosted-runner-select.yml
```diff
@@ -0,0 +1,99 @@
+name: Self-hosted runner select
```
mrobinson, line 1:
```suggestion
name: Self-hosted Runner Selection
```

.github/workflows/self-hosted-runner-select.yml
```diff
@@ -0,0 +1,99 @@
...
+
```
mrobinson, line 2:
```suggestion
```

.github/workflows/self-hosted-runner-timeout.yml
```diff
@@ -0,0 +1,40 @@
+name: Self-hosted runner timeout
```
mrobinson, line 1:
```suggestion
name: Self-hosted Runner Timeout
```

.github/workflows/self-hosted-runner-timeout.yml
```diff
@@ -0,0 +1,40 @@
...
+
```
mrobinson, line 2:
```suggestion
```

mrobinson (approved):

mrobinson (approved):

PR #33324, opened 2024-09-05
Plumb selected tracing spans into hitrace

delan:
This patch converts tracing spans to HiTrace spans on OpenHarmony, when servoshell is built with the `tracing-hitrace` feature. Only spans defined in #33189 are converted, based on the presence of a new `servo_profiling` field.

---


atbrakhi (approved):
This looks good. I think it just needs to be signed-off

PR #33336, opened 2024-09-05
layout: Lay out absolutes in atomic containing blocks

mrobinson:
When inline atomics establish containing blocks for absolute
descendants, layout should happen with those atomics as the containing
block. This ensures that the absolute descendents have the correct
containing block and Fragment parent. This wasn't happening before and
this change fixes that.

One test starts to fail because of the lack of `overflow: clip` support.

---


Loirooriol (approved):
I'm not seeing the new failure due to `overflow: clip`?

PR #33343, opened 2024-09-06
Handle all `white-space` values when intrinsically sizing an IFC

### LINKED ISSUE #33335: Wrong min-content size for `white-space: nowrap`
```html
<!DOCTYPE html>
<div style="width: 0">
  <div style="display: inline-block; border: solid; white-space: nowrap">
    A B C D E F G H I J K L M N O P Q R S T U V W X Y Z</td>
  </div>
</div>
```

| | |
| - | - |
| Expected |  |
| Actual |  |

https://github.com/servo/servo/blob/7e493ba865ef7e748ce3b14d59117315a0ec8aee/components/layout_2020/flow/inline/mod.rs#L2344

Loirooriol:
There were various cases like `text-wrap-mode: nowrap` and
`white-space-collapse: break-spaces` that weren't handled well.

Fixes #33335

flexbox_flex-formatting-interop.html fails now because we don't support
`table-layout: fixed`.

---

- [X] These changes fix #33335
- [X] There are tests for these change

### DIFF
--- a/components/fonts/glyph.rs
+++ b/components/fonts/glyph.rs
@@ -446,6 +446,11 @@ pub struct GlyphStore {
     /// Whether or not this glyph store contains only glyphs for whitespace.
     is_whitespace: bool,
 
+    /// Whether or not this glyph store ends with whitespace glyphs.
+    /// Typically whitespace glyphs are placed in a separate store,
+    /// but that may not be the case with `white-space: break-spaces`.
+    ends_with_whitespace: bool,
+
     /// Whether or not this glyph store contains only a single glyph for a single
     /// preserved newline.
     is_single_preserved_newline: bool,
@@ -460,6 +465,7 @@ impl<'a> GlyphStore {
     pub fn new(
         length: usize,
         is_whitespace: bool,
+        ends_with_whitespace: bool,
         is_single_preserved_newline: bool,
         is_rtl: bool,
     ) -> GlyphStore {
@@ -472,6 +478,7 @@ impl<'a> GlyphStore {
             total_word_separators: 0,
             has_detailed_glyphs: false,
             is_whitespace,
+            ends_with_whitespace,
             is_single_preserved_newline,
             is_rtl,
         }
@@ -490,6 +497,11 @@ impl<'a> GlyphStore {
     #[inline]
     pub fn is_whitespace(&self) -> bool {
         self.is_whitespace
+    }
+
+    #[inline]
+    pub fn ends_with_whitespace(&self) -> bool {
+        self.ends_with_whitespace
     }
 
     #[inline]
--- a/components/layout_2020/flow/inline/mod.rs
+++ b/components/layout_2020/flow/inline/mod.rs
@@ -2275,11 +2275,14 @@ struct ContentSizesComputation<'layout_data> {
     containing_block: &'layout_data IndefiniteContainingBlock<'layout_data>,
     paragraph: ContentSizes,
     current_line: ContentSizes,
-    /// Size for whitepsace pending to be added to this line.
-    pending_whitespace: Au,
+    /// Size for whitespace pending to be added to this line.
+    pending_whitespace: ContentSizes,
+    /// Whether or not the current line has seen any content (excluding collapsed whitespace),
+    /// when sizing under a min-content constraint.
+    had_content_yet_for_min_content: bool,
     /// Whether or not the current line has seen any content (excluding collapsed whitespace),
     /// when sizing under a max-content constraint.
-    had_content_yet: bool,
+    had_content_yet_for_max_content: bool,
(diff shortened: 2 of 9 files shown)


components/fonts/glyph.rs
```diff
@@ -446,6 +446,11 @@ pub struct GlyphStore {
     /// Whether or not this glyph store contains only glyphs for whitespace.
     is_whitespace: bool,
 
+    /// Whether or not this glyph store ends with whitespace glyphs.
+    /// Typically whitespace glyphs are placed on a separate store,
```
mrobinson, line 450:
```suggestion
    /// Typically whitespace glyphs are placed in a separate store,
```

components/layout_2020/flow/inline/mod.rs
```diff
@@ -2345,31 +2350,33 @@ impl<'layout_data> ContentSizesComputation<'layout_data> {
...
                                 continue;
                             }
                         }
 
                         self.commit_pending_whitespace();
                         self.add_inline_size(advance);
-                        self.had_content_yet = true;
+                        if can_wrap && run.glyph_store.ends_with_whitespace() {
+                            self.line_break_opportunity();
+                        }
```
mrobinson, line 2379: Does this only happen with `WhiteSpaceCollapse::BreakSpaces`? If so, it might be nice to leave a comment here as that is really what allows the line break opportunity after white space.

components/layout_2020/flow/inline/mod.rs
```diff
@@ -2408,25 +2414,26 @@ impl<'layout_data> ContentSizesComputation<'layout_data> {
     }
 
     fn line_break_opportunity(&mut self) {
+        self.pending_whitespace.min_content = Au::zero();
```
mrobinson, line 2417: This is white space that is hanging right? I would put a comment. Note that we don't always properly hang white space in layout yet.

components/layout_2020/flow/inline/text_run.rs
```diff
@@ -245,17 +249,17 @@ impl TextRunSegment {
...
                 }
 
                 slice.end = whitespace.start;
             }
 
             // If there's no whitespace and `word-break` is set to `keep-all`, try increasing the slice.
             // TODO: This should only happen for CJK text.
-            let can_break_anywhere = text_style.word_break == WordBreak::BreakAll ||
-                text_style.overflow_wrap == OverflowWrap::Anywhere ||
-                text_style.overflow_wrap == OverflowWrap::BreakWord;
```
mrobinson, line 257: Oof. Thanks for this cleanup,.

tests/wpt/tests/css/css-text/white-space/break-spaces-with-word-break-001.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Text Test: white-space:break-spaces + word-break:keep-all</title>
+<link rel="author" title="Oriol Brufau" href="mailto:obrufau@igalia.com">
+<link rel="help" href="https://drafts.csswg.org/css-text-4/#valdef-white-space-collapse-break-spaces">
+<link rel="help" href="https://drafts.csswg.org/css-text-4/#valdef-word-break-keep-all">
+<link rel="match" href="../../reference/ref-filled-green-100px-square.xht">
+<meta name="assert" content="break-spaces + keep-all do allow a break after space.">
```
mrobinson, line 8:
```suggestion
<meta name="assert" content="break-spaces + keep-all does allow a break after space.">
```

This is a nit obviously, but it made me consider something I never have before. The result of an expression with "plus" is singular and the result of an "and" expression is plural.

mrobinson (approved):
Nice!

components/layout_2020/flow/inline/mod.rs
Loirooriol: Yes, the condition can hold in other cases but they are handled above with a `continue`, so by this point the condition should only be true for `WhiteSpaceCollapse::BreakSpaces`.

components/layout_2020/flow/inline/mod.rs
Loirooriol: Yes, either hanging (e.g. `white-space: pre-wrap`) or removed (e.g. `white-space: normal`).
If the space needs to be included, then `commit_pending_whitespace()` should be called first.

tests/wpt/tests/css/css-text/white-space/break-spaces-with-word-break-001.html
Loirooriol: Makes sense. I just copied this from another test.

Loirooriol:
There was a windows-specific unit test that needed to pass the new parameter.

PR #33344, opened 2024-09-06
Make Crashtests with test-wait wait

### LINKED ISSUE #33342: Crashtests with test-wait don't wait
From https://web-platform-tests.org/writing-tests/crashtest.html:
```
In some cases crashtests may need to perform work after the initial page load. In this case the test may specify a class=test-wait attribute on the root element. The test will not complete until that attribute is removed from the root. At the time when the test would otherwise have ended a TestRendered event is emitted; test authors can use this event to perform modifications that are guaranteed not to be batched with the initial paint. This matches the behaviour of [reftests](https://web-platform-tests.org/writing-tests/reftests.html).
```

Servo only checks for reftest-wait right now, so any crashtest that relies on test-wait exits very quickly: https://github.com/servo/servo/blob/ebed9218f2907767ba3c9dd9f27f30a6a6e9f225/components/script/dom/window.rs#L2019-L2023

Taym95:
---

- [x] These changes fix #33342


components/script/dom/window.rs
```diff
@@ -2018,9 +2018,10 @@ impl Window {
         if self.prepare_for_screenshot && for_display {
             // Checks if the html element has reftest-wait attribute present.
             // See http://testthewebforward.org/docs/reftests.html
+            // and https://web-platform-tests.org/writing-tests/crashtest.html
             let html_element = document.GetDocumentElement();
             let reftest_wait = html_element.map_or(false, |elem| {
-                elem.has_class(&atom!("reftest-wait"), CaseSensitivity::CaseSensitive)
+                elem.has_class(&atom!("reftest-wait"), CaseSensitivity::CaseSensitive) || elem.has_class(&atom!("test-wait"), CaseSensitivity::CaseSensitive)
```
Taym95, line 2024: I could also dublicate this:
```
            let test_wait = html_element.map_or(false, |elem| {
                elem.has_class(&atom!("test-wait"), CaseSensitivity::CaseSensitive)
            });
            ```

jdm: Now that we have a prefs file that is only applied during WPT tests, could we gate the class name checks on a new preference and set that preference in the file?

jdm: Actually we can leave that for a followup issue. This seems fine to merge if we use Atom::from instead of the macro.

Taym95: > Now that we have a prefs file that is only applied during WPT tests, could we gate the class name checks on a new preference and set that preference in the file?

Could you elaborate on this? I didn't quite understand, and I would like to follow up.

jdm (approved):

jdm:
Would you.mund checking a few if those tests to understand why they are timing out? The answer might be they rely on a missing feature, but I don't want to assume that.

Taym95:
> Would you mind checking a few of those tests to understand why they are timing out? The answer might be they rely on a missing feature, but I don't want to assume that.

yes sure, I will check them

Taym95:
> Would you mind checking a few of those tests to understand why they are timing out? The answer might be they rely on a missing feature, but I don't want to assume that.

Some time out with error(missing feature) but some just timeout:

- pid:85139 [2024-09-09T12:21:11Z ERROR script::dom::bindings::error] Error at http://web-platform.test:8000/css/compositing/opacity-and-transform-animation-crash.html:18:14 document.execCommand is not a function
 0:22.65 TEST_END: TIMEOUT, expected PASS - Executor hit external timeout (this may indicate a hang)
 
- TIMEOUT /css/css-animations/crashtests/replace-keyframes-animating-filter-001.html - Executor hit external timeout (this may indicate a hang)
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1030, in _bootstrap
    self._bootstrap_inner()
    

- pid:85432 [2024-09-09T12:24:07Z ERROR script::dom::bindings::error] Error at http://web-platform.test:8000/css/css-cascade/scope-shadow-sharing.html:47:27 host.attachShadow is not a function
 0:18.73 TEST_END: TIMEOUT, expected FAIL

- pid:85611 [2024-09-09T12:25:48Z ERROR script::dom::bindings::error] Error at http://web-platform.test:8000/css/css-flexbox/animation/flex-basis-content-crash.html:8:23 target.animate is not a function
 0:04.06 pid:85611 UNSUPPORTED (log once): POSSIBLE ISSUE: unit 1 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
 0:22.87 TEST_END: TIMEOUT, expected PASS - Executor hit external timeout (this may indicate a hang)
 
- pid:85811 [2024-09-09T12:27:09Z ERROR script::dom::bindings::error] Error at https://web-platform.test:8443/css/css-paint-api/column-count-crash.https.html:34:1 CSS.paintWorklet is undefined
 0:22.53 TEST_END: TIMEOUT, expected PASS - Executor hit external timeout (this may indicate a hang)

- TIMEOUT /css/cssom-view/overflow-hidden-smooth-scroll-crash.html - Executor hit external timeout (this may indicate a hang)
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1030, in _bootstrap
    self._bootstrap_inner()

- TIMEOUT /css/filter-effects/feimage-circular-reference-foreign-object-crash.html - Executor hit external timeout (this may indicate a hang)
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1030, in _bootstrap
    self._bootstrap_inner()

jdm:
I would bet money that the ones without errors are waiting for an event that Servo does not fire. This looks good to merge!

Taym95:
> * selection-modify-line-boundary-around-empty-details.html

Done, updated test result expectations

mrobinson:
I think these tests are timing out now because `test-wait` should cause reftests to wait, only crashtests.

jdm:
> I think these tests are timing out now because `test-wait` should cause reftests to wait, only crashtests.

@mrobinson Is your comment missing a word, and possibly some context?

mrobinson:
Oh yeah, sorry my comment was indeed a bit rushed. This change caused 16 tests to starting timing out. This is because some of these tests have `test-wait` in their reference file. This `test-wait` should not apply when running a reftest. It should only apply when running a crash test.

In addition, I'm not sure the test added here effectively tests that crash tests are waiting for `test-wait` to be removed from the `<body>` element. Indeed, if you remove the wait condition from the code, `tests/wpt/mozilla/tests/mozilla/test-wait-crash.html` still passes. Curiously, this change also caused a crash test to start timing out.

When we switch to servodriver to run reftests, we can remove this code, as I believe that it allows crashtests to wait properly without causing these other tests to time out as well.

jdm:
Oh, yeah, I see now.

PR #33345, opened 2024-09-06
Initial support for `<link rel="prefetch">`

simonwuelker:
This PR adds basic support for prefetching resources using the [`rel="prefetch"`](https://html.spec.whatwg.org/multipage/#link-type-prefetch) attribute on `<link>` elements.

To allow this, I first changed the way link relations are determined. Previously, we always examined the `rel` attribute with functions like `string_is_stylesheet`. 
https://github.com/servo/servo/blob/ebed9218f2907767ba3c9dd9f27f30a6a6e9f225/components/script/dom/htmllinkelement.rs#L209-L214
This is error prone for the following reasons:
* There are synonyms for some `rel` keywords (`rel="copyright"` is the same as `rel="license"`)
* Some `rel` keywords are enabled by other attributes (`rev="made"` implies `rel="author"`)
* `rel` keywords are not mutually exclusive (`rel="stylesheet icon"` is totally valid, yet the code above doesn't handle this)

Instead, we now store the possible link relations in a bitset that can be obtained using `LinkRelations::for_element`.

---

This implementation is mostly concerned with sending the fetch request and dispatching the correct events in response. The following wpt tests now pass:
* `preload/prefetch-cache.html`
* `preload/prefetch-events.html`
* `preload/prefetch-time-to-fetch.https.html`
* `preload/prefetch-accept.html` 

The remaining tests for prefetching are mostly related to incorrect/missing headers. I'm not familiar with servos fetch api so these might be easy fixes. Is there some API in `RequestBuilder` for `sec-fetch-purpose`/`sec-fetch-dest` headers?
Failing wpt tests:

* `preload/prefetch-document.html` 
* `preload/prefetch-types.https.html` (missing `sec-fetch-*` headers)
* `preload/prefetch-headers.https.html` (missing `sec-fetch-*` headers)

I can split this PR into smaller ones if that's preferred for review.

[try run wpt-2020](https://github.com/Wuelle/servo/actions/runs/10746056589)

---

- [X] There are tests for these changes (WPT)

### DIFF
--- a/components/script/dom/htmllinkelement.rs
+++ b/components/script/dom/htmllinkelement.rs
@@ -28,7 +28,7 @@ use style::parser::ParserContext as CssParserContext;
 use style::stylesheets::{CssRuleType, Origin, Stylesheet, UrlExtraData};
 use style_traits::ParsingMode;
 
-use super::types::GlobalScope;
+use super::types::{EventTarget, GlobalScope};
 use crate::dom::attr::Attr;
 use crate::dom::bindings::cell::DomRefCell;
 use crate::dom::bindings::codegen::Bindings::DOMTokenListBinding::DOMTokenList_Binding::DOMTokenListMethods;
@@ -374,7 +374,7 @@ impl HTMLLinkElement {
 
         let fetch_context = sync::Arc::new(sync::Mutex::new(PrefetchContext {
             url: url,
-            document: Trusted::new(&document),
+            link: Trusted::new(self),
             resource_timing: ResourceFetchTiming::new(ResourceTimingType::Resource),
         }));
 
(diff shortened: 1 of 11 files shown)


components/script/dom/htmllinkelement.rs
```diff
@@ -259,6 +299,109 @@ impl VirtualMethods for HTMLLinkElement {
 }
 
 impl HTMLLinkElement {
+    /// <https://html.spec.whatwg.org/multipage/semantics.html#create-link-options-from-element>
+    fn processing_options(&self) -> LinkProcessingOptions {
+        // Step 1.
```
jdm, line 307: Let's add the spec text for each step, since the step numbering often drifts over time.

components/script/dom/htmllinkelement.rs
```diff
@@ -259,6 +299,109 @@ impl VirtualMethods for HTMLLinkElement {
...
+        // Step 6.
+        assert!(!options.href.is_empty() || options.source_set.is_some());
+
+        // Step 7.
+        options
+    }
+
+    /// The `fetch and process the linked resource` algorithm for [`rel="prefetch"`](https://html.spec.whatwg.org/multipage/links.html#link-type-prefetch)
+    fn fetch_and_process_prefetch_link(&self, href: &str) {
+        // Step 1.
```
jdm, line 352: Same note about step descriptions.

jdm (approved):
This was very readable; thanks!

PR #33346, opened 2024-09-06
layout: Implement proper absolute child position for flexbox

### LINKED ISSUE #33265: Properly position absolutes inside of flex containers

mrobinson:
This implements the requirements outlined in the [flexbox specification]
about how to position absolute children of flex containers. We must
establish a static position rectangle (to use if all insets are auto)
and also align the child into that rectangle.

[flebox specification]: https://drafts.csswg.org/css-flexbox/#abspos-items

Fixes #33265.

---


components/layout_2020/flexbox/layout.rs
```diff
@@ -264,10 +253,40 @@ impl FlexContainerConfig {
...
             )
     }
+
+    fn resolve_reversable_flex_alignment(
+        &self,
+        align_flags: AlignFlags,
+        reversed: bool,
+    ) -> AlignFlags {
+        match (align_flags, reversed) {
+            (AlignFlags::FLEX_START, false) => AlignFlags::START | align_flags.flags(),
```
Loirooriol, line 267: Should this actually match `align_flags.value()` instead?
Otherwise `align_flags.flags()` is always 0 if `align_flags` is `FLEX_START` and `FLEX_END`.

components/layout_2020/flexbox/layout.rs
```diff
@@ -863,6 +865,68 @@ impl FlexContainer {
...
+        &self,
+        absolutely_positioned_box: ArcRefCell<AbsolutelyPositionedBox>,
+        containing_block: &ContainingBlock,
+        container_size: FlexRelativeVec2<Au>,
+        positioning_context: &mut PositioningContext,
+    ) -> Fragment {
+        let alignment = {
+            let fragment = absolutely_positioned_box.borrow();
+            let make_flex_only_values_directional_for_absolutes =
+                |value: AlignFlags, reversed: bool| match (reversed, value.value()) {
```
Loirooriol, line 893: It's a bit confusing to match in the opposite order of the function parameters

```suggestion
                |value: AlignFlags, reversed: bool| match (value.value(), reversed) {
```

components/layout_2020/flexbox/layout.rs
```diff
@@ -863,6 +865,68 @@ impl FlexContainer {
         }
     }
 
+    /// Create a absolutely positioned flex child fragment, using the rules the
+    /// specification dictates. This should take into account the alignment and
+    /// justification values of the container and the child to position it within a
+    /// 'inset-modified containing block," which may be either the "static-position
```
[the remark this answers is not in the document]
mrobinson: Done.

components/layout_2020/flexbox/layout.rs
mrobinson: Yeah, I've switched this to match `align_flags.value()` after confirming that it doesn't work how we discussed it might.

components/layout_2020/flexbox/layout.rs
mrobinson: I've switched these around.

components/layout_2020/flexbox/layout.rs
```diff
@@ -707,15 +713,15 @@ impl FlexContainer {
         let mut cross_start_position_cursor = match resolved_align_content {
```
Loirooriol, line 713: I guess this could use `resolve_reversable_flex_alignment()` and remove the cases for `AlignFlags::FLEX_START`  and `AlignFlags::FLEX_END`.

Loirooriol (approved):

mrobinson:
Hrm...the new failure is a bit suspicious so I'm going to investigate a bit before landing this.

PR #33354, opened 2024-09-07
script: fixed document title being set to Some("") instead of None

Melchizedek6809:
I suppose this wasn't really an issue before but since we use the
title as a label for the tab it's a bit inconvenient, since pages
without a <title> get an empty label because of this, instead of
falling back to the URL as other browsers do.

---

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -1234,10 +1234,10 @@ impl Document {
         }
     }
 
-    /// Return the title if one is specified
-    ///
-    /// https://html.spec.whatwg.org/multipage/#document.title
-    fn get_title(&self) -> Option<DOMString> {
+    /// Determine the title of the [`Document`] according to the specification at:
+    /// <https://html.spec.whatwg.org/multipage/#document.title>. The difference
+    /// here is that when the title isn't specified `None` is returned.
+    fn title(&self) -> Option<DOMString> {
         let title = self.GetDocumentElement().and_then(|root| {
             if root.namespace() == &ns!(svg) && root.local_name() == &local_name!("svg") {
                 // Step 1.
@@ -1266,7 +1266,7 @@ impl Document {
     pub fn send_title_to_embedder(&self) {
         let window = self.window();
         if window.is_top_level() {
-            let title = self.get_title().map(String::from);
+            let title = self.title().map(String::from);
             self.send_to_embedder(EmbedderMsg::ChangePageTitle(title));
         }
     }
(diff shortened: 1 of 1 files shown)


Wuelle:
Hmm i feel like when a page explicitly sets an empty title using `<title></title>` then we should respect that and send `EmbedderMsg::ChangePageTitle(Some(""))` instead of `EmbedderMsg::ChangePageTitle(None)`. How the embedder (in this case the minibrowser) deals with this empty page title is up to them. Thats why I handled empty tab titles in https://github.com/servo/servo/pull/33229 on the minibrowser side, but the changes seem to have been lost in #33244 (see https://github.com/servo/servo/pull/33244#discussion_r1747743286).

Melchizedek6809:
Yeah you're right, that seems to be the least suprising way to handle things. I've changed the code so now we send `None` only when there's no `<title>` element. We now only need to add the check whether the title is empty back in, do you want to open a PR for that? Since it seems like a separate change from this PR.

components/script/dom/document.rs
```diff
@@ -1234,11 +1234,37 @@ impl Document {
...
+            } else {
+                // Step 2.
+                root.upcast::<Node>()
+                    .traverse_preorder(ShadowIncluding::No)
+                    .find(|node| node.is::<HTMLTitleElement>())
+            }
+        });
+
+        title.map(|title| {
+            // Steps 3-4.
```
mrobinson, line 1257: Without the spec link, these steps don't refer to anything any longer. Please add the specification link.

components/script/dom/document.rs
```diff
@@ -1234,11 +1234,39 @@ impl Document {
         }
     }
 
+    /// Return the title if one is specified
+    ///
+    /// https://html.spec.whatwg.org/multipage/#document.title
+    fn get_title(&self) -> Option<DOMString> {
```
mrobinson, line 1240: - I think this documentation comment could be a bit better and is missing a period.
- It's a rustdoc error to have a raw URL so this should be surrounded by angle brackets.
- Generally speaking, in Rust this should be named simply `title()`: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter.

```suggestion
    /// Determine the title of the [`Document`] according to the specification at:
    /// <https://html.spec.whatwg.org/multipage/#document.title>. The difference
    /// here is that when the title isn't specified `None` is returned.
    fn title(&self) -> Option<DOMString> {
```

mrobinson (changes requested):

components/script/dom/document.rs
Melchizedek6809: Makes sense, regarding the name: since `Title` is already defined wouldn't another name be more appropriate? Feels like having both `title` and `Title` could be confusing.

mrobinson (approved):

PR #33355, opened 2024-09-07
script: Use `time@0.3` for input elements and do conversion in a &str trait

mrobinson:
This changes converts all input element parsing and normalization to use
`time` instead of `chrono`. `time` is used by our dependencies, so it
makes sense to work toward removing the Servo dependency on `chrono`.

In addition, parsing and normalization also moves to a trait on &str to
prepare for the possibility of all script parsers moving to a separate
crate that can have unit tests written against it.

Code duplication is eliminated when possible and more conversion is done
using integer types. These two things together mean we pass more tests
now.

---

### DIFF
--- a/components/script/dom/bindings/str.rs
+++ b/components/script/dom/bindings/str.rs
@@ -12,13 +12,12 @@ use std::str::FromStr;
 use std::sync::LazyLock;
 use std::{fmt, ops, str};
 
-use chrono::prelude::{Utc, Weekday};
-use chrono::{Datelike, TimeZone};
 use cssparser::CowRcStr;
 use html5ever::{LocalName, Namespace};
 use num_traits::Zero;
 use regex::Regex;
 use servo_atoms::Atom;
+use time_03::{Date, Month, OffsetDateTime, Time, Weekday};
 
 /// Encapsulates the IDL `ByteString` type.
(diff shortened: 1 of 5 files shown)


components/script/dom/bindings/str.rs
```diff
@@ -731,52 +468,67 @@ fn parse_date_component(value: &str) -> Option<(i32, u32, u32)> {
...
+    if millisecond_length > 3 {
+        return None;
+    }
+    let millisecond_int = millisecond.parse::<u16>().ok()?;
+    let millisecond_int = millisecond_int * 10_u16.pow(3 - millisecond_length);
 
-    // Step 8
-    Some((hour_int, minute_int, second_float))
+    // Step 8: Return hour, minute, and second.
+    Some((hour_int, minute_int, second_int, millisecond_int))
```
gterzian, line 531: and, in our case, millis?

gterzian (approved):
LGTM

components/script/dom/bindings/str.rs
mrobinson: Good point. I've updated this comment to reflect that in our case we return milliseconds.

mrobinson:
@gterzian Thanks so much for the review.

PR #33358, opened 2024-09-07
Set empty object as `console` prototype

simonwuelker:
The console object has an empty object as its prototype, not the realm object prototype.

[Specification](https://console.spec.whatwg.org/#console-namespace):
> For historical web-compatibility reasons, the [namespace object](https://webidl.spec.whatwg.org/#dfn-namespace-object) for [console](https://console.spec.whatwg.org/#namespacedef-console) must have as its [[Prototype]] an empty object, created as if by [ObjectCreate](https://tc39.github.io/ecma262/#sec-objectcreate)([%ObjectPrototype%](https://tc39.github.io/ecma262/#sec-properties-of-the-object-prototype-object)), instead of [%ObjectPrototype%](https://tc39.github.io/ecma262/#sec-properties-of-the-object-prototype-object).

[try-run](https://github.com/Wuelle/servo/actions/runs/10754666399)

---
- [X] There are tests for these changes (wpt)

### DIFF
--- a/components/script/dom/webidls/Console.webidl
+++ b/components/script/dom/webidls/Console.webidl
@@ -5,8 +5,7 @@
 // https://console.spec.whatwg.org/
 
 [ClassString="Console",
- Exposed=(Window,Worker,Worklet),
- ProtoObjectHack]
+ Exposed=(Window,Worker,Worklet)]
 namespace console {
   // Logging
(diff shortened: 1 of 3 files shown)


jdm (approved):
Huh. That must be a spec change from when when we first implemented the Console interface.

PR #33360, opened 2024-09-08
script: Implement `TextEncoder::encodeInto()`

webbeef:
This implements https://encoding.spec.whatwg.org/#dom-textencoder-encodeinto

@mrobinson This is the same patch with your comments addressed. I switched to a new PR because I had issues rebasing some wpt manifests.

---

- [X] There are tests for these changes (wpt expectations are updated)

### DIFF
--- a/components/script/dom/textencoder.rs
+++ b/components/script/dom/textencoder.rs
@@ -98,7 +98,7 @@ impl TextEncoderMethods for TextEncoder {
         // Step 3, 4, 5, 6
         // Turn the source into a queue of scalar values.
         // Iterate over the source values.
-        for result in  source.0.chars() {
+        for result in source.0.chars() {
             let utf8_len = result.len_utf8();
             if available - written >= utf8_len {
                 // Step 6.4.1
--- a/components/script/dom/textencoder.rs
+++ b/components/script/dom/textencoder.rs
@@ -95,13 +95,10 @@ impl TextEncoderMethods for TextEncoder {
 
         let dest = unsafe { destination.as_mut_slice() };
 
-        // Step 3, 4, 5
+        // Step 3, 4, 5, 6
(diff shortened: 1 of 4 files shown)


mrobinson (changes requested):
Looking good. Just a couple more suggestions. I've filed #33366 in order to catch the issue with URLs in rustdoc.

components/script/dom/textencoder.rs
```diff
@@ -55,17 +59,71 @@ impl TextEncoder {
...
+
+        let dest = unsafe { destination.as_mut_slice() };
+
+        // Step 3, 4, 5
+        // Turn the source into a queue of scalar values.
+        let mut chars = source.0.chars();
+
+        // Step 6
+        // Iterate over the source values.
+        for result in chars {
```
mrobinson, line 104:
```suggestion
        // Step 3, 4, 5, 6
        // Turn the source into a queue of scalar values.
        // Iterate over the source values.
        for result in  source.0.chars() {
```

mrobinson (approved):

components/script/dom/textencoder.rs
```diff
@@ -55,17 +59,68 @@ impl TextEncoder {
...
+
+        let mut read = 0;
+        let mut written = 0;
+
+        let dest = unsafe { destination.as_mut_slice() };
+
+        // Step 3, 4, 5, 6
+        // Turn the source into a queue of scalar values.
+        // Iterate over the source values.
+        for result in  source.0.chars() {
```
mrobinson, line 101:
```suggestion
        for result in source.0.chars() {
```

mrobinson (commented):
For some reason the subtests marked as passing are now failing: https://github.com/servo/servo/runs/29841280313

PR #33380, opened 2024-09-09
script: Added missing spec step in `Location::SetHash`

### LINKED ISSUE #33353: `mach test-speedometer` timeout
**Describe the bug:**
Setting `window.location.hash` to same value inside `hashchange` event handler causes an infinite loop that makes `constellation` unresponsive (has to process an ever increasing history).

**To Reproduce:**
Run `mach test-speedometer -r`

**Platform:**
Windows 11, x86_64
Servo main [152e620](https://github.com/servo/servo/tree/152e62022a8c69d28c217b0fce651fd2d02dc08f)

`hashchange` infinite loop can be reproduced with:
```html
<!DOCTYPE html>
<script>
	let count = 0;
	
	window.onhashchange = () => {
		count += 1;
		console.log(count);
		
		window.location.hash = 'running';
	}
	
	window.location.hash = 'running';
</script>
```
Firefox or Chromium print `1` but Servo keeps going.

crbrz:
Added check if the new fragment is different when updating location hash. Also, fixes wrong location hash (it was set to #) when `window.location.hash` was set to an empty string.

---

- [X] These changes fix #33353

I did a [try wpt-2020 run](https://github.com/crbrz/servo/actions/runs/10772645732/job/29872675832) but no test results changed so I added new WPT tests for these changes.

### DIFF
--- a/components/script/dom/location.rs
+++ b/components/script/dom/location.rs
@@ -315,13 +315,19 @@ impl LocationMethods for Location {
             // Step 5: Set copyURL's fragment to the empty string.
             // Step 6: Basic URL parse input, with copyURL as url and fragment state as
             // state override.
-            copy_url.as_mut_url().set_fragment(match value.0.as_str() {
+            let new_fragment = match value.0.as_str() {
                 "" => Some("#"),
                 _ if value.0.starts_with('#') => Some(&value.0[1..]),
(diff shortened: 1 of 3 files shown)


jdm (approved):
Thanks for fixing this and adding tests!

PR #33384, opened 2024-09-09
Add support for `table-layout: fixed`

Loirooriol:
More details might be needed to fully support the feature, but this covers the basic functionality.

---


Loirooriol:
Some of the failures are because `overflow` should have no effect on row groups.

mrobinson (approved):

PR #33385, opened 2024-09-09
libservo: Improve finding python

jschwe:
Servo is built in a virtual environment, which sets [`VIRTUAL_ENV`](https://docs.python.org/3/library/venv.html) to the base path of the venv.
PYTHON3 is only set if mach or the user set it.

Additional changes:

- Use Path / var_os for Paths instead of strings. In General using Path APIs is preferable, since in rare cases valid paths may not be valid utf-8.
- Don't search for python 3.8 anymore, since we require a newer version anyway.
- Don't add the .exe suffix anymore, since Command::new() [will take care of that](https://doc.rust-lang.org/std/process/struct.Command.html#platform-specific-behavior).

---

- [x] These changes fix `cargo build` failing when python3.8 is installed (building without mach)

### DIFF
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -67,7 +67,7 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
 /// Tries to find a suitable python
 ///
 /// Algorithm
-/// 1. Trying to find python3/python in $VIRTUAL_ENV (this should be from servos venv)
+/// 1. Trying to find python3/python in $VIRTUAL_ENV (this should be from Servo's venv)
 /// 2. Checking PYTHON3 (set by mach)
 /// 3. Falling back to the system installation.
 ///
@@ -75,13 +75,13 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
 fn find_python() -> PathBuf {
     let mut candidates = vec![];
     if let Some(venv) = env::var_os("VIRTUAL_ENV") {
-        let base = PathBuf::from(venv);
-        let bin = base.join("bin");
-        let python = bin.join("python");
-        let python3 = bin.join("python3");
+        let bin_directory = PathBuf::from(venv).join("bin");
+
+        let python3 = bin_directory.join("python3");
         if python3.exists() {
             candidates.push(python3);
         }
+        let python = bin_directory.join("python");
         if python.exists() {
             candidates.push(python);
         }
@@ -111,7 +111,7 @@ fn find_python() -> PathBuf {
         .map(|c| c.into_os_string())
         .collect::<Vec<_>>();
     panic!(
-        "Can't find python (tried {:?})! Try enabling servos python venv, \
+        "Can't find python (tried {:?})! Try enabling Servo's Python venv, \
         setting the PYTHON3 env var or adding python3 to PATH.",
         candidates.join(", ".as_ref())
     )
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -64,26 +64,55 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
     }
 }
 
-fn find_python() -> String {
-    env::var("PYTHON3").ok().unwrap_or_else(|| {
-        let candidates = if cfg!(windows) {
-            ["python.exe", "python"]
-        } else {
-            ["python3", "python"]
-        };
-        for &name in &candidates {
-            if Command::new(name)
-                .arg("--version")
-                .output()
-                .ok()
-                .map_or(false, |out| out.status.success())
-            {
-                return name.to_owned();
-            }
+/// Tries to find a suitable python
+///
+/// Algorithm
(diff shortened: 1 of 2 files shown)


components/servo/build.rs
```diff
@@ -23,26 +23,56 @@ fn main() {
...
+///
+/// Algorithm
+/// 1. Trying to find python3/python in $VIRTUAL_ENV (this should be from servos venv)
+/// 2. Checking PYTHON3 (set by mach)
+/// 3. Falling back to the system installation.
+fn find_python() -> PathBuf {
+    let mut candidates = vec![];
+    if let Some(venv) = env::var_os("VIRTUAL_ENV") {
+        let mut base = PathBuf::from(venv);
+        base.push("bin");
```
Taym95, line 36: I think you can use `join` to avoid `push`:
```
let base = PathBuf::from(venv).join("bin");
let python = base.join("python");
```

components/servo/build.rs
```diff
@@ -23,26 +23,54 @@ fn main() {
...
+    let system_python = ["python3", "python"].map(PathBuf::from);
+    candidates.extend_from_slice(&system_python);
+
+    for name in &candidates {
+        // Command::new() allows us to omit the `.exe` suffix on windows
+        if Command::new(&name)
+            .arg("--version")
+            .output()
+            .ok()
+            .map_or(false, |out| out.status.success())
```
Wuelle, line 62: Could use `Option::is_some_and` here

sagudev:
BTW similar code is also at https://github.com/servo/servo/blob/f1ad364ec2dacca3ec7d79830ef8da9f26fbf4e2/components/script/build.rs#L67
so it would be nice to keep it in sync.

components/script/build.rs
```diff
@@ -64,26 +64,55 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
...
-                .output()
-                .ok()
-                .map_or(false, |out| out.status.success())
-            {
-                return name.to_owned();
-            }
+/// Tries to find a suitable python
+///
+/// Algorithm
+/// 1. Trying to find python3/python in $VIRTUAL_ENV (this should be from servos venv)
```
mrobinson, line 70:
```suggestion
/// 1. Trying to find python3/python in $VIRTUAL_ENV (this should be from Servo's venv)
```

components/script/build.rs
```diff
@@ -64,26 +64,55 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
...
+///
+/// Note: This function should be kept in sync with the version in `components/servo/build.rs`
+fn find_python() -> PathBuf {
+    let mut candidates = vec![];
+    if let Some(venv) = env::var_os("VIRTUAL_ENV") {
+        let base = PathBuf::from(venv);
+        let bin = base.join("bin");
+        let python = bin.join("python");
+        let python3 = bin.join("python3");
+        if python3.exists() {
```
mrobinson, line 82:
```suggestion
        let bin_directory = PathBuf::from(venv).join("bin");
        
        let python3 = bin_directory.join("python3");
        if python3.exists() {
```

components/script/build.rs
```diff
@@ -64,26 +64,55 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
...
+            candidates.push(python3);
         }
-        panic!(
-            "Can't find python (tried {})! Try fixing PATH or setting the PYTHON3 env var",
-            candidates.join(", ")
-        )
-    })
+        if python.exists() {
+            candidates.push(python);
+        }
```
mrobinson, line 87:
```suggestion
        let python = bin.join("python");
        if python.exists() {
            candidates.push(python);
        }
```

components/script/build.rs
```diff
@@ -64,26 +64,55 @@ impl<'a> phf_shared::PhfHash for Bytes<'a> {
...
+        {
+            return name.to_owned();
+        }
+    }
+    let candidates = candidates
+        .into_iter()
+        .map(|c| c.into_os_string())
+        .collect::<Vec<_>>();
+    panic!(
+        "Can't find python (tried {:?})! Try enabling servos python venv, \
```
mrobinson, line 114:
```suggestion
        "Can't find python (tried {:?})! Try enabling Servo's Python venv, \
```

mrobinson (approved):
Looks good with some nits for both versions. It would be nice to eventually share this code somehow.

jschwe:
>  It would be nice to eventually share this code somehow.

One could do something like `include!()` or 
```
#[path = ../utils/find_python.rs]
mod find_python;
```

This avoids adding another dependency to the build-script and should thus not influence compile-times.

mrobinson (approved):

jschwe:
windows try build is still running https://github.com/jschwe/servo/actions/runs/10816401959/job/30007519006, but it should pass now. The main issue was that on windows `python` is under `$VIRTUAL_ENV/Scripts` instead of in the `bin` directory as on Linux / macos.

PR #33386, opened 2024-09-09
Propagate `CanGc` from `Document::new()`

last-genius:
Following https://github.com/servo/servo/issues/33140 and #33144, more work on marking functions that can transitively trigger a GC to easier recognize hazards.

---
- [x] These changes are part of #33140  (GitHub issue number if applicable)

### DIFF
--- a/components/script/document_loader.rs
+++ b/components/script/document_loader.rs
@@ -14,6 +14,7 @@ use servo_url::ServoUrl;
 use crate::dom::bindings::root::Dom;
 use crate::dom::document::Document;
 use crate::fetch::FetchCanceller;
+use crate::script_runtime::CanGc;
 
 #[derive(Clone, Debug, JSTraceable, MallocSizeOf, PartialEq)]
 pub enum LoadType {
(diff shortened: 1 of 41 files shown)


last-genius:
There are a few autogenerated things that I couldn't figure how to modify: `VirtualX` trait methods, and `Document::Open`, which somehow autogenerates both `Open` and `Open_` methods (only one of them can GC as of now). This PR has by no means fully propagated CanGc from `Document::new`, there's still a lot of work to be done.

Another issue I wanted to ask about is whether I should make `CanGc` `Copy` or `Clone`, given that this PR has quite a few functions that call several canGc functions in turn, and currently `CanGc::note()` instead of `can_gc` has to be used on all but the first invocation (same if a function calls a canGc func in a loop, etc.), potentially complicating diffs for future changes.

jdm:
Good question about the moved value! Yes, we should derive copy and clone to make passing it around as easy as possible.

components/script/document_loader.rs
```diff
@@ -59,7 +60,7 @@ impl LoadBlocker {
 impl Drop for LoadBlocker {
     fn drop(&mut self) {
         if let Some(load) = self.load.take() {
-            self.doc.finish_load(load);
+            self.doc.finish_load(load, CanGc::note());
```
jdm, line 63: This is a really interesting result from this work! I would not have assumed that dropping a LoadBlocker value could cause a JS GC, and it's neat that this PR makes that more clear.

jdm (approved):
This is really helpful! Let's add the copy derive and remove a bunch of the CanGc::note calls, and then let's merge this!

jdm:
Actually, to reduce the risk of merge conflicts let's merge this as-is and do the derive and replacements in a followup.

PR #33394, opened 2024-09-10
Stop using `time@0.1` in Servo

mrobinson:
This removes the last few uses of `time@0.1` in Servo. There are still
dependencies from `style` and `webrender`, but they will be removed soon
as well. The uses of this version of `time` are replaced with
`std::time` types and `time@0.3` when negative `Duration` is necessary.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,8 +121,7 @@ surfman = { version = "0.9.8", features = ["chains"] }
 syn = { version = "2", default-features = false, features = ["clone-impls", "derive", "parsing"] }
 synstructure = "0.13"
 thin-vec = "0.2.13"
-time = "0.1.41"
-time_03 = { package = "time", version = "0.3", features = ["large-dates", "serde"] }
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -121,8 +121,7 @@ surfman = { version = "0.9.8", features = ["chains"] }
 syn = { version = "2", default-features = false, features = ["clone-impls", "derive", "parsing"] }
 synstructure = "0.13"
 thin-vec = "0.2.13"
-time = "0.1.41"
-time_03 = { package = "time", version = "0.3", features = ["large-dates", "serde"] }
+time_03 = { package = "time", version = "0.3", features = ["large-dates", "local-offset", "serde"] }
```
atouchet, line 124: Can this be changed to just "time" now?

mrobinson: Yes, but I plan to do this in a followup in order to reduce the size of these changes.

jdm (approved):
Woohoo!

PR #33395, opened 2024-09-10
Raf delivery: run rafs for all pipeline if tick received for any.

### LINKED ISSUE #33028: Intermittent PASS in `/html/webappapis/update-rendering/child-document-raf-order.html`
```
OK /html/webappapis/update-rendering/child-document-raf-order.html

    PASS [expected FAIL] subtest: Ordering of steps in "Update the Rendering" - child document requestAnimationFrame order
```

### LINKED ISSUE #33251: Fix Running Animation Frame Callbacks
When receiving this message

https://github.com/servo/servo/blob/46dbe4ce320eb99e851dd422f624615e18c39e15/components/script/script_thread.rs#L1875

We do two things:

- Note a rendering opp(which queues a task to "update the rendering")
- Note pending animation tick(which stores the tick on the document for later processing in the "update the rendering" task). 

When the "update the rendering" task runs, we [call into](https://github.com/servo/servo/blob/46dbe4ce320eb99e851dd422f624615e18c39e15/components/script/script_thread.rs#L1715) `tick_all_animations` for all docs, which runs the animation frame callbacks if a tick of type `AnimationTickType::REQUEST_ANIMATION_FRAME` was previously stored. 

https://github.com/servo/servo/blob/46dbe4ce320eb99e851dd422f624615e18c39e15/components/script/dom/document.rs#L3301

The spec requires rafs to be called in a given order, which is tested at https://github.com/servo/servo/blob/cbc9304c2002fb0d5c8cdade39eddfe117b12053/tests/wpt/tests/html/webappapis/update-rendering/child-document-raf-order.html#L4

Problem: the compositor sends animation ticks in no specific order, so there is a race condition on the ordering of rafs in the presence of iframes(which is not due to https://github.com/servo/servo/issues/31973), because when we "note a pending animation tick" for document, and queue the "update the rendering" task, we may not have received all ticks from the compositor by the time the task runs. 

That is, I think, why the test above fails intermittently. 

When it goes right, we hit the lucky path:

- "note a pending animation tick" for iframe
- "note a pending animation tick" for parent
- Run "update the rendering" task, which first calls the rafs of the parent. 

The unlucky one is:

- "note a pending animation tick" for iframe
- Run "update the rendering" task, which only calls the rafs of the iframe. 
- "note a pending animation tick" for parent(too late, test fails)

Fix: we need to run rafs for a document as part of "update the rendering" if one was requested, even if no ticks were received from the compositor for that document, so long as we received a tick for any pipeline in that script-thread from the compositor. 

Longer term we should consolidate the workflow between script and the compositor so that it is not pipeline-based, but rather "render opportunity" based, the current problem is a sympton of how we would do things before restructuring to match https://html.spec.whatwg.org/multipage/webappapis.html#update-the-rendering). There is a discussion over at https://servo.zulipchat.com/#narrow/stream/263398-general/topic/HTML.20event-loop.20fix

gterzian:
The first commits ensures the ordering of running rafs between parent and child docs, by running rafs for all pipeline if a tick was received for any(see #33251).

The second commits prevents a sequence of "update the rendering" tasks from running one after the other and starving other script tasks(see https://github.com/servo/servo/pull/33395#issuecomment-2343431208). 

For the first commit there is a TLA spec modelling the idea at https://gist.github.com/gterzian/1f297aa22393ea8604ad14ca6a5cfff6

---

- [ ] These changes fix #33251 and fix #33028 (GitHub issue number if applicable)

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -3335,10 +3335,17 @@ impl Document {
         self.pending_animation_ticks.borrow_mut().extend(tick_type);
     }
 
+    /// Whether this document has received an animation tick for rafs.
+    pub fn has_received_raf_tick(&self) -> bool {
+        self.pending_animation_ticks
+            .borrow()
+            .contains(AnimationTickType::REQUEST_ANIMATION_FRAME)
+    }
+
     /// As part of a `update_the_rendering` task, tick all pending animations.
-    pub fn tick_all_animations(&self) {
+    pub fn tick_all_animations(&self, should_run_rafs: bool) {
         let tick_type = mem::take(&mut *self.pending_animation_ticks.borrow_mut());
-        if tick_type.contains(AnimationTickType::REQUEST_ANIMATION_FRAME) {
+        if should_run_rafs {
             self.run_the_animation_frame_callbacks();
         }
         if tick_type.contains(AnimationTickType::CSS_ANIMATIONS_AND_TRANSITIONS) {
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -1821,6 +1821,10 @@ impl ScriptThread {
         };
         debug!("Got event.");
 
+        // Prioritize only a single update of the rendering;
+        // others will run together with the other sequential tasks.
+        let mut rendering_update_already_prioritized = false;
+
         loop {
             let pipeline_id = self.message_to_pipeline(&event);
             let _realm = pipeline_id.map(|id| {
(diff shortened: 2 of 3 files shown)


gterzian:
Interesting

> assert_array_equals: expected order of notifications lengths differ, expected array ["parent_raf", "first_child_raf", "second_child_raf"] length 3, got […, "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf", "parent_raf", "first_child_raf", "second_child_raf"] length 24

gterzian:
> Interesting

I think this can happen because the test recursively requests rafs:

https://github.com/servo/servo/blob/cbc9304c2002fb0d5c8cdade39eddfe117b12053/tests/wpt/tests/html/webappapis/update-rendering/child-document-raf-order.html#L90

and we prioritize the "update the rendering task" versus other tasks, including the [timeout](https://github.com/servo/servo/blob/cbc9304c2002fb0d5c8cdade39eddfe117b12053/tests/wpt/tests/html/webappapis/update-rendering/child-document-raf-order.html#L107) used to signal that the test is finished. 

So this can result in many runs of the prioritized update the rendering task until finally the script-thread runs the sequential timeout. 

Will try to address in a commit.

jdm (approved):
Looks like a reasonable solution!

Loirooriol:
This seems to have caused an intermittent failure: #33413

PR #33406, opened 2024-09-11
layout: Reverse `space-between` alignment for absolute children of flex containers

mrobinson:
We assume that absolutely positioned children are the only flex item, and `space-between` requires more than 1 item, so we use the fallback position, which is `safe flex-start`, and then we need to resolve `flex-start` to either `start` or `end`.

---

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -892,6 +892,8 @@ impl FlexContainer {
                         AlignFlags::END | AlignFlags::SAFE
                     },
                     (AlignFlags::STRETCH, false) => AlignFlags::START | AlignFlags::SAFE,
+                    (AlignFlags::SPACE_BETWEEN, false) => AlignFlags::START | AlignFlags::SAFE,
+                    (AlignFlags::SPACE_BETWEEN, true) => AlignFlags::END | AlignFlags::SAFE,
                     _ => value,
                 };
             let cross = make_flex_only_values_directional_for_absolutes(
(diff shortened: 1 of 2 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -892,6 +892,8 @@ impl FlexContainer {
                         AlignFlags::END | AlignFlags::SAFE
                     },
                     (AlignFlags::STRETCH, false) => AlignFlags::START | AlignFlags::SAFE,
+                    (AlignFlags::SPACE_BETWEEN, false) => AlignFlags::START,
+                    (AlignFlags::SPACE_BETWEEN, true) => AlignFlags::END,
```
Loirooriol, line 896: Presumably with `AlignFlags::SAFE`?

Loirooriol (approved):
I don't really understand the commit message.

But I think the idea is that we are assuming that the abspos is the only flex item, and `space-between` requires more than 1 item, so we use the fallback position, which is `safe flex-start`, and then we need to resolve `flex-start` to either `start` or `end`.

BTW, we broke this in https://github.com/servo/servo/commit/d169a82d2e146704db3d9eb2b5f9f49f9ef8b1da#diff-e49e880bf571ecb371556a62d96ec7a78e8ae4ec6a9fa7631385b0496e365a96L691

components/layout_2020/flexbox/layout.rs
mrobinson: Yep. I've updated teh commit message as well.

PR #33407, opened 2024-09-11
Make CanGc derive Copy and Clone

### LINKED ISSUE #33404: CanGc should derive Copy and Clone
There are a bunch of uses of CanGc::note that could be replaced with existing values in the same function if the values can be trivially copied. Most uses of CanGc::note inside of a loop or following another call that used can_gc as an argument can be replaced with can_gc.

Taym95:
---

- [x] These changes fix #33404

- [x] These changes do not require tests because there is not behavior chages.


jdm (approved):
Fantastic, thanks!

Taym95:
I solved merge conflict, can we try merge this again?

Issue #17084, opened 2017-05-29
clean-cargo-cache does not honour .servobuild settings


jdm:
If you set `cargo-home-dir` to a non-default value in a `.servobuild` file, `./mach clean-cargo-cache` continues to look in the `.cargo` directory of the current working directory.

cc @UK992

jdm:
Option no longer exists.

PR #33408, opened 2024-09-11
Fix ./mach clean-cargo-cache

Gae24:
By chance I used the command and realized that it didn't work. Also while I looked to see if it had already been reported, I stumbled in #17084, that should be closed given that the option to override cargo-home-dir was removed from servobuild.

In Python 2 the `filter` function used to return a `list`, but the behaviour changed in Python 3, wrapping the result in a list was enough. 

---

- [X] These changes do not require tests because it's a simple python fix


mrobinson:
@Gae24 Do you mind explaining a bit what the problem was in your commit description and how your change addresses it? Thank you!

jdm (approved):
Thanks!

mrobinson:
> @Gae24 Do you mind explaining a bit what the problem was in your commit description and how your change addresses it? Thank you!

This isn't a huge deal if it look okay to @jdm.

Gae24:
Basically in python 2 the filter function used to return a list, but the behaviour changed in python3, wrapping the result in a list was enough

Issue #32972, opened 2024-08-08
Panic in dom/nodes/insertion-removing-steps/blur-event.window.js ("Attempt to use script or layout while DOM not in a stable state")


jdm:
To reproduce: `./mach test-wpt tests/wpt/tests/dom/nodes/insertion-removing-steps/blur-event.window.js`

I _think_ what's happening here is that we're firing a focus event in the middle of performing DOM tree mutations, so the content JS could observe the tree in an indeterminate state if we allowed it to proceed.

Backtrace:
```
pid:10891 assertion `left == right` failed: Attempt to use script or layout while DOM not in a stable state
 0:02.96 pid:10891   left: 1
 0:02.96 pid:10891  right: 0 (thread Script(1,1), at components/script/dom/document.rs:3421)
 0:05.36 pid:10891   13: core::panicking::assert_failed
 0:05.36 pid:10891              at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:298:5
 0:05.42 pid:10891   14: script::dom::document::Document::ensure_safe_to_run_script_or_layout
 0:05.42 pid:10891              at /Users/jdm/src/servo/components/script/dom/document.rs:3421:9
 0:05.45 pid:10891   15: script::dom::bindings::callback::CallSetup::new
 0:05.45 pid:10891              at /Users/jdm/src/servo/components/script/dom/bindings/callback.rs:246:13
 0:05.48 pid:10891   16: script::dom::bindings::codegen::Bindings::EventHandlerBinding::EventHandlerNonNull::Call_
 0:05.48 pid:10891              at /Users/jdm/src/servo/target/debug/build/script-84e8aed3a23eae36/out/Bindings/EventHandlerBinding.rs:34:17
 0:05.51 pid:10891   17: script::dom::eventtarget::CompiledEventListener::call_or_handle_event
 0:05.51 pid:10891              at /Users/jdm/src/servo/components/script/dom/eventtarget.rs:249:44
 0:05.54 pid:10891   18: script::dom::event::inner_invoke
 0:05.54 pid:10891              at /Users/jdm/src/servo/components/script/dom/event.rs:726:9
 0:05.55 pid:10891   19: script::dom::event::invoke
 0:05.55 pid:10891              at /Users/jdm/src/servo/components/script/dom/event.rs:658:17
 0:05.55 pid:10891   20: script::dom::event::Event::dispatch
[...]

PR #15567, opened 2017-02-15
Replace inheritance_integrity by trait shenanigans 

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/servo/15567)

Issue #14902, opened 2017-01-06
Write a MIR lint for rooting analysis


jdm:
From the work week discussion about rooting safety, one problem that was discussed was dealing with generic types. It was proposed that a MIR lint would be safer than the current lints, because MIR lints run on monomorphized types. My understanding was that @Manishearth and @nikomatsakis were interested in making this happen.

jdm:
@Manishearth tells me that he is too busy to work on this. I would settle for any information about how to write a MIR lint so that someone else could theoretically do this work.

Manishearth:
I've never done it before. You need to [register](http://manishearth.github.io/rust-internals-docs/rustc_plugin/registry/struct.Registry.html#method.register_mir_pass) a [MIR pass](http://manishearth.github.io/rust-internals-docs/rustc/mir/transform/trait.MirMapPass.html). You create MIR passes by implementing [`MirPass`](http://manishearth.github.io/rust-internals-docs/rustc/mir/transform/trait.MirPass.html) (`MirMapPass` is auto magically implemented)

You probably want a MIR visitor (http://manishearth.github.io/rust-internals-docs/rustc/mir/visit/trait.Visitor.html) for walking the MIR.

nikomatsakis:
I started doing it and didn't get very far. It doesn't seem very hard to do though. @jdm do you plan to pursue it?

nikomatsakis:
Maybe we can chat about it on IRC :)

nox:
@nikomatsakis Think you could push what you did somewhere?

Mrowqa:
Hi. As we talked before, I will work on it as a part of Bachelor's thesis. I plan to start working on it quite soon (probably tomorrow), so I'd like to know if there are some good resources to learn about rust compiler plugins. Furthermore, is there anything else I should know? I know the general idea of handles and rooting.

jdm:
You can see the existing HIR plugin at https://github.com/servo/servo/blob/master/components/script_plugins/unrooted_must_root.rs. There are a few [unit tests](https://github.com/servo/servo/blob/master/tests/unit/script_plugins/lib.rs) and it might be good to add some more as part of this work.

jdm:
You can find documentation for the compiler internals at http://manishearth.github.io/rust-internals-docs/rustc_plugin/, and https://doc.rust-lang.org/beta/unstable-book/language-features/plugin.html may have some useful information.

jdm:
[...]

PR #33416, opened 2024-09-12
Generate JS bindings in a separate crate from hand-written implementations

### LINKED ISSUE #1799: Split up the script crate for build speed

jdm:
Based on the prior work from #21371 and the recent discussion in #1799, I have come up with a design that works for all of Servo's current WebIDL interfaces. I have yet to perform any formal benchmarks for how the build time/memory usage compares for various scenarios. I've opened this PR to make it easier to run the full testsuite and solicit feedback.

My overall goal for this PR is to replicate the work from #21371 while maintaining existing ergonomics in hand-written code in the script crate as much as possible. This largely means finding ways to minimize the use of `::<crate::DomTypeHolder>`.

If we agree on the direction of the changes, I have ideas for how to make it easier to merge in steps:
1. [x] add constructors and static functions to the generated *Methods traits (very large source of merge conflicts in this work)
2. [x] generate more Default implementations for dictionaries automatically, removing some hand-written ones
3. [x] move codegen and webidl files into a shared location outside of any single crate (eg. python/codegen and components/shared/webidls)
4. [ ] create new crate that is responsible for generating dictionaries/unions/enums, and move all hand-written impls for those types into the new crate
6. [ ] generate all bindings for interfaces and callbacks in the new crate

Known issues:
* [x] maplike key type is hardcoded
* [x] the principals.rs file in script_bindings is hacked up enough to compile and needs more work
* [x] the trace function in script_bindings/settings_stack.rs for StackEntry needs work
* [x] generic type interacts poorly with crown
* [x] iterator classes do not generate derive checks correctly
* [x] `ShouldNotImplDomObject` assertions are broken

---
- [x] These changes fix #1799


sagudev:
Hm, I think the error comes from the way we use reuse git checkout on self hosted runners. I suggest running CI in your own fork via `mach try wpt-2020` and linking result here.

jdm:
I'll see if https://github.com/jdm/servo/actions/runs/10824430930 is able to make better progress.

sagudev:
data clone errors are due to hacks in principals.rs

jdm:
https://github.com/jdm/servo/actions/runs/10887236429

sagudev:
> Only one last test result to investigate!
> 
> ```
> ▶ OK [expected CRASH] /dom/nodes/insertion-removing-steps/blur-event.window.html
>   │   ▶ FAIL [expected PASS] <input> element does not fire blur event upon DOM removal
>   │   │   → assert_false: Blur event was not fired expected false got true
>   │   │ 
>   │   │ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:8:15
>   │   │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2622:25
>   │   │ test@http://web-platform.test:8000/resources/testharness.js:633:30
>   │   └ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:1:5
>   │   ▶ FAIL [expected PASS] <button> element does not fire blur/focusout events upon DOM removal
>   │   │   → assert_false: Blur event was not fired expected false got true
>   │   │ 
>   │   │ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:25:15
>   │   │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2622:25
>   │   │ test@http://web-platform.test:8000/resources/testharness.js:633:30
>   └   └ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:11:5
> ```

Might be flake, so be sure to rule that out first.

sagudev:
> ShouldNotImplDomObject assertions are broken

I guess the problem here is that we do this check using trait shenanigans and orphan rules bit us hard, but we can port inheritance_integrity back to crown: https://github.com/servo/servo/pull/15567 or try to fix `ShouldNotImplDomObject` by using supertrait or something.

jdm:
> Only one last test result to investigate!
> 
> ```
> ▶ OK [expected CRASH] /dom/nodes/insertion-removing-steps/blur-event.window.html
>   │   ▶ FAIL [expected PASS] <input> element does not fire blur event upon DOM removal
>   │   │   → assert_false: Blur event was not fired expected false got true
>   │   │ 
>   │   │ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:8:15
>   │   │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2622:25
>   │   │ test@http://web-platform.test:8000/resources/testharness.js:633:30
>   │   └ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:1:5
>   │   ▶ FAIL [expected PASS] <button> element does not fire blur/focusout events upon DOM removal
>   │   │   → assert_false: Blur event was not fired expected false got true
>   │   │ 
>   │   │ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:25:15
>   │   │ Test.prototype.step@http://web-platform.test:8000/resources/testharness.js:2622:25
>   │   │ test@http://web-platform.test:8000/resources/testharness.js:633:30
>   └   └ @http://web-platform.test:8000/dom/nodes/insertion-removing-steps/blur-event.window.js:11:5
> ```

Aha, this is because I commented out a line in callback.rs that fails an assertion in that test (#32972): https://github.com/servo/servo/pull/33416/files#diff-ebc3f2cee053bcf3b0d2fae9c56e2a052fc1400813e65c8abfbb31a521fdc61cR248-R249

jdm:
First benchmarks for build times, using 2022 apple MacBook Air m2, 8gb, 8cpu, based on 0e56241c1be52efd52e7ba4c999fec0bf29ce2de:

Debug builds:

|     | main | branch |
| --- | --- | --- |
|clean| 3m19s | 3m13s |
|incremental, webidl + impl | 43s | 40s
|incremental, impl | 36s | 22s

Release builds:

|        | main | branch |
| --- | --- | --- |
| clean| 5m20s | 5m41s |
| incremental, webidl + impl | 1m26s | 1m32s
| incremental, impl | 1m26s | 1m07s

Testing methadology:
1. `rm -rf target/ && ./mach build [-r]`
2. open `WebGLRenderingContext.webidl` and add a new `undefined whatever();`, then add `fn Whatever(&self) {}` to the corresponding impl in `webglrenderingcontext.rs`, then build
3.  add a `fn foo(&self) {}` to the `impl Capabilities` block in `webglrenderingcontext.rs`, then build

I used the build time reported by cargo for each measurement.

sagudev:
> open WebGLRenderingContext.webidl and add a new undefined whatever();, then add fn Whatever(&self) {} to the corresponding impl in webglrenderingcontext.rs, then build

```patch
diff --git a/components/script/dom/webglrenderingcontext.rs b/components/script/dom/webglrenderingcontext.rs
index 51bde8adeb..555ffc333a 100644
--- a/components/script/dom/webglrenderingcontext.rs
+++ b/components/script/dom/webglrenderingcontext.rs
@@ -1958,6 +1958,7 @@ impl Drop for WebGLRenderingContext {
 }
 
 impl WebGLRenderingContextMethods<crate::DomTypeHolder> for WebGLRenderingContext {
+    fn Whatever(&self) {}
     // https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.1
     fn Canvas(&self) -> DomRoot<HTMLCanvasElement> {
         DomRoot::from_ref(&*self.canvas)
diff --git a/components/script_bindings/webidls/WebGLRenderingContext.webidl b/components/script_bindings/webidls/WebGLRenderingContext.webidl
index fcd3f9752a..a1e19342d6 100644
--- a/components/script_bindings/webidls/WebGLRenderingContext.webidl
+++ b/components/script_bindings/webidls/WebGLRenderingContext.webidl
@@ -683,6 +683,7 @@ interface mixin WebGLRenderingContextOverloads
 [Exposed=(Window)]
 interface WebGLRenderingContext
 {
+    undefined whatever();
 };
 WebGLRenderingContext includes WebGLRenderingContextBase;
 WebGLRenderingContext includes WebGLRenderingContextOverloads;
```

> add a fn foo(&self) {} to the impl Capabilities block in webglrenderingcontext.rs, then build

```patch
diff --git a/components/script/dom/webglrenderingcontext.rs b/components/script/dom/webglrenderingcontext.rs
index 555ffc333a..96931cf65c 100644
--- a/components/script/dom/webglrenderingcontext.rs
+++ b/components/script/dom/webglrenderingcontext.rs
@@ -4688,6 +4688,8 @@ struct Capabilities {
 }
 
 impl Capabilities {
+    fn foo(&self) {}
+    
     fn set(&self, cap: u32, set: bool) -> WebGLResult<bool> {
         let cap = CapFlags::from_enum(cap)?;
         let mut value = self.value.get();
```

Wuelle:
Fedora 40 on a Ryzen 7 3700X, 8 cores with 32gb ram, comparing https://github.com/servo/servo/commit/0e56241c1be52efd52e7ba4c999fec0bf29ce2de with https://github.com/servo/servo/pull/33416/commits/858f88d2312a91d6c0c8274c9f2cbf132309dd4b.

# Debug
| | main | branch |
| --- | --- | --- |
| clean | 03:47/03:37 | 03:55 |
| incremental, webidl + impl | 01:22 | 01:15 |
| incremental, impl | 01:05 | 00:56 |

# Release
| | main | branch |
| --- | --- | --- |
| clean | 05:36 | 04:39 |
| incremental, webidl + impl | 01:53 | 01:49 |
| incremental, impl | 01:53 | 01:12 |

(The results seem a bit noisy, I might rerun them later)

jschwe:
Probably benchmarking `production` is also interesting, since that is where the long build-time of script is really noticable, due to `codegen-units = 1`.

mukilan:
NixOS 24.05,  Intel i7-12800HX (24 threads, 8p + 8e cores), 64GB RAM 

# Release

|            | main    | branch |
|------------|---------|---------
|clean                      | 3m 37s  | 3m 27s |
|incremental, webidl + impl | 1m 19s  | 1m 15s |
|incremental, impl          | 1m 18s  | 0m 50s |


# Debug

|            | main    | branch
|------------|---------|---------
|clean                      | 2m 30s  | 2m 36s
|incremental, webidl + impl | 0m 46s  | 0m 40s
|incremental, impl          | 0m 38s  | 0m 30s


# Production

|            | main    | branch
|------------|---------|---------
|clean                      | 5m 52s | 5m 32s
|incremental, webidl + impl | 4m 10s | 3m 54s
|incremental, impl          | 4m 11s | 3m 31s

sagudev:
Ubuntu 24.04,  AMD Ryzen 7 PRO 5875U (16 threads, 8 cores), 16GB RAM (laptop)

I also unset `CCACHE` that I usually use.


# Debug

|            | main    | branch
|------------|---------|---------
|clean                      | 4m 35s  | 5m 24s
|incremental, webidl + impl | 0m 58s  | 0m 47s
|incremental, impl          | 0m 45s  | 0m 27s

# Release

|            | main    | branch |
|------------|---------|---------
|clean                      | 8m 8s  | 7m 20s |
|incremental, webidl + impl | 2m 14s  | 1m 55s |
|incremental, impl          | 2m 6s  | 1m 25s |


# Production

|            | main    | branch
|------------|---------|---------
|clean                      | 10m 46s |9m 47s
|incremental, webidl + impl | 6m 0s | 5m 1s
|incremental, impl          | 6m 7s | 4m 27s


I also stored all [cargo-timings in zip](https://github.com/user-attachments/files/17071792/cargo-timings.zip) and I must say graphs are interesting one noticeable thing we avoid is reruning build.rs of script crate (although it's quick as we have some webidl caching on python level too), but this should not run at all if we do not change any webidl file (we are missing [`rerun-if-changed`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) as by default whole crate is checked for change).

nicoburns:
This seems like a great change. I am wondering if it would be possible to get further benefits (both compile time speed ups, and modularity) by splitting the script crate "horizontally" (i.e. having each set of APIs in their own crate). I imagine there's a core set of APIs that would need to all be compiled together, but many web APIs are quite independent and naively I might expect those to be compilable independently.

sagudev:
> This seems like a great change. I am wondering if it would be possible to get further benefits (both compile time speed ups, and modularity) by splitting the script crate "horizontally" (i.e. having each set of APIs in their own crate). I imagine there's a core set of APIs that would need to all be compiled together, but many web APIs are quite independent and naively I might expect those to be compilable independently.

This should also be possible, after this lands. See #1799 and https://gist.github.com/sagudev/c4a029cfccda9553adac32749a9ba583#stage-2

jdm:
I'm planning to move ahead with this work, following the path I outlined in the initial PR description to reduce the risk of rebase conflicts and making overall changes easier to review. The first step will be making the generated Methods traits contain static methods and special operations, which should be a boring but independently useful change.

sagudev:
> I'm planning to move ahead with this work, following the path I outlined in the initial PR description to reduce the risk of rebase conflicts and making overall changes easier to review.

+1 for that, but we should also work on resolving issues before major landing (step 5).

sagudev:
I will also try to fix crown problem.

sagudev:
I tried to extract problem with `DecodeSuccessCallback` that uses `allow_unrooted_interior` but this doesn't compile (although similar code compiles in servo):

```rust
#[crown::unrooted_must_root_lint::allow_unrooted_interior]
struct Outer {
    inner: Inner2,
}

#[crown::unrooted_must_root_lint::must_root]
struct Inner2;
```

am I misunderstanding `allow_unrooted_interior`? Unfortunately there are no test/docs for `allow_unrooted_interior` that I could look at.

sagudev:
We have https://github.com/servo/servo/blob/c7a4e4f627f0bf331cedc1700ddeb92ffcba4e78/components/script/dom/bindings/mod.rs#L172-L175

sagudev:
@jdm why script crate still have build.rs and https://github.com/jdm/servo/blob/ce640c6730d1379dbd2bc3688b79875f274aaf50/components/script/dom/bindings/mod.rs#L188?

sagudev:
Crown would be problematic as currently this does not raise error:
```rust
#[crown::unrooted_must_root_lint::must_root]
struct Foo(i32);
struct Bar<TH: TypeHolderTrait>(TH::F);
//~^ ERROR: Type must be rooted, use #[crown::unrooted_must_root_lint::must_root] on the struct definition to propagate

trait TypeHolderTrait {
    type F;
    type B;
}

struct TypeHolder;

impl TypeHolderTrait for TypeHolder {
    type F = Foo;
    type B = Bar<Self>;
}

fn main() {}
```

per https://github.com/fan-tom/liquid-rust:

> We can easily get MIR of function or constexpr if we know substitution of all type variables within that fun/const. It means however that we cannot analyze generic functions in abstract, only when they are called from monomorphized function

I think it is possible to lower HIR to MIR in LatePass: https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/mir/mod.rs#L134
relevant issue that try to write MIR lint (when MirPass was still available): https://github.com/servo/servo/issues/14902

Both rooting and tracing concept are actually very similar. All (there are some exceptions) types that contain must_root types must be must_root too and same goes for Traceables, so if we can unify concepts and lints for them that would be ideal (we can model must_root with "marker" Trait).

sagudev:
I think we can only get MIR for bodies (function, closures); maybe the fastest option would be to make crown enforce something like this:
```rust
#[crown::unrooted_must_root_lint::must_root]
struct Foo(i32);
struct Bar<TH: TypeHolderTrait>(TH::F);
//~^ ERROR: Type must be rooted, use #[crown::unrooted_must_root_lint::must_root] on the struct definition to propagate

trait TypeHolderTrait {
    type F;
    type B;
}

struct TypeHolder;

impl TypeHolderTrait for TypeHolder {
    #[crown::unrooted_must_root_lint::must_root]
    type F = Foo;
    type FF = Foo;
    //~^ ERROR: Type must be rooted, use #[crown::unrooted_must_root_lint::must_root] on the type definition to propagate
    type B = Bar<Self>;
}
```

jdm:
> @jdm why script crate still have build.rs and https://github.com/jdm/servo/blob/ce640c6730d1379dbd2bc3688b79875f274aaf50/components/script/dom/bindings/mod.rs#L188?

We still generate some code via script's build script when the actual concrete DOM types are available. For example, the implementation of the DOM types trait that's generated in script_bindings.

sagudev:
> > @jdm why script crate still have build.rs and https://github.com/jdm/servo/blob/ce640c6730d1379dbd2bc3688b79875f274aaf50/components/script/dom/bindings/mod.rs#L188?
> 
> We still generate some code via script's build script when the actual concrete DOM types are available. For example, the implementation of the DOM types trait that's generated in script_bindings.

I think we should just generate this once as part of script_bindings, then include! macro to import it.

jdm:
Hmm, interesting idea. I'll think about that!

sagudev:
I tried applying https://github.com/servo/servo/pull/34163, but it didn't create any additional errors in script crate.

jdm:
> I tried applying #34163, but it didn't create any additional errors in script crate.

I'm building on #34163 and addressing the errors. I've got some more fixes for crown that will be needed.

jdm:
I'll try rebasing.

sagudev:
> I'll try rebasing.

This will be hard.

sagudev:
Btw instead of updating rust version just port https://github.com/servo/servo/pull/34257

jdm:
Mmm, looks like the test failures are triggered by the rust upgrade: https://github.com/servo/servo/pull/34270

jdm:
All the known issues now have solutions. I think we can start figuring out:
1. the name of the new crate
2. where to move the codegen and webidl files
3. how to review the changes

sagudev:
> the name of the new crate

I think `script_bindings` is a good name.

> move codegen and webidl files into a shared location outside of any single crate (eg. python/codegen and components/shared/webidls)

I am not sure how I feel about separating webidl/codegen, it would make sense to remain as part of script_*. I think we can start by creating script_bindings that has build.rs, codegen and webidls, and all of is files are simply `include!` from script.

> how to review the changes/how to make the changes easier to review

Crown changes would also make sense to split into separate PR, as for others I am not sure how easy would be to split them.

jdm:
My current plan for reviewing and merging these changes incrementally:
* crown changes (#34359)
* DomTypes and DomTypeHolder generation (#34357)
* reduce the number of hand-written impls on generated types (#34356)
* split binding setup into static initialization and runtime initialization
* create script_bindings and incrementally various hand-written pieces of components/script/dom/bindings into it, using the new DomTypes trait
* split codegen into parts that use the DomTypes types and parts that use concrete types, while still keeping it inside components/script
* move codegen into script_bindings, move all hand-written impls on generated types into script_bindings

jdm:
No real point in keeping this PR open. I'll track all future progress in #1799.

PR #33418, opened 2024-09-12
Improve spec conformance around request header validation

shanehandley:
Updates validation of forbidden request headers to catch additional cases.

Some code was de-duplicated and re-organised a little to share methods between `dom` and `net/fetch`.

---


jdm (commented):
This is much clearer. Thanks!

components/script/dom/xmlhttprequest.rs
```diff
@@ -460,41 +460,37 @@ impl XMLHttpRequestMethods for XMLHttpRequest {
...
-                    s
-                }
-            },
-            None => unreachable!(),
-        };
+
+        let name_str = name.as_str().ok_or(Error::Syntax)?;
+
+        // Step 5: If (name, value) is a forbidden request-header, then return.
+        if is_forbidden_request_header(name_str, &value.to_vec()) {
```
mukilan, line 481: Is the `.to_vec()` needed here? I think we can pass `value` to `is_forbidden_request_header` directly.

shanehandley: Thank you

mukilan (approved):

PR #33422, opened 2024-09-12
Fix prioritization of rendering updates

### LINKED ISSUE #33028: Intermittent PASS in `/html/webappapis/update-rendering/child-document-raf-order.html`

gterzian:
This PR changes how rendering tasks are prioritized, versus other script tasks, but running only one, in priority, and re-queuing any others for the next tick of the script-thread(after all sequential tasks have run in the current tick). 

---

- [ ] These changes fix #33028 (GitHub issue number if applicable)

### DIFF
--- a/components/script/script_thread.rs
+++ b/components/script/script_thread.rs
@@ -1822,8 +1822,9 @@ impl ScriptThread {
         debug!("Got event.");
 
         // Prioritize only a single update of the rendering;
-        // others will run together with the other sequential tasks.
-        let mut rendering_update_already_prioritized = false;
+        // others will be re-queued,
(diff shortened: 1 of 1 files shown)


gterzian:
Obviously this would break more than it would fix, also on second though this would also only again reduce the chance of the problem occuring, it would not remove it(because at the next tick of the loop the rendering task that is re-queued could run before a timeout, and timeout could also not be part of the sequential tasks in the current tick).

Issue #31516, opened 2024-03-06
De-duplicate phf


Loirooriol:
https://github.com/servo/stylo/commit/d0095a5904db03ee0f804b99ad56f9db0cdc1d82 updated Stylo's phf to 0.11

<details><summary>So now we have some duplication</summary>

- duplicate versions for package `phf`
    - The following packages depend on version 0.10.1 from 'crates.io':
      - markup5ever 0.11.0
      - script 0.0.1
    - The following packages depend on version 0.11.2 from 'crates.io':
      - cssparser 0.31.2
      - selectors 0.24.0
- duplicate versions for package `phf_codegen`
    - The following packages depend on version 0.10.0 from 'crates.io':
      - markup5ever 0.11.0
      - script 0.0.1
    - The following packages depend on version 0.11.2 from 'crates.io':
      - selectors 0.24.0
- duplicate versions for package `phf_generator`
    - The following packages depend on version 0.10.0 from 'crates.io':
      - phf_codegen 0.10.0
      - string_cache_codegen 0.5.2
    - The following packages depend on version 0.11.2 from 'crates.io':
      - phf_codegen 0.11.2
      - phf_macros 0.11.2
- duplicate versions for package `phf_shared`
    - The following packages depend on version 0.10.0 from 'crates.io':
      - phf 0.10.1
      - phf_codegen 0.10.0
      - phf_generator 0.10.0
      - script 0.0.1
      - string_cache 0.8.7
      - string_cache_codegen 0.5.2
    - The following packages depend on version 0.11.2 from 'crates.io':
      - phf 0.11.2
      - phf_codegen 0.11.2
      - phf_generator 0.11.2
      - phf_macros 0.11.2

</details>

To avoid the duplication we should do something like:
- [X] Update script to use 0.11
    #31537
    #33434
- [X] markup5ever was already updated to 0.11 (https://github.com/servo/html5ever/commit/b9f858c3874134afeb0dcfecc9998bae67a5f2d3), but this commit is in no released version, so we need a new release
    https://github.com/servo/html5ever/pull/537
- [X] Update Stylo to use the new html5ever/markup5ever/xml5ever
    https://github.com/servo/stylo/pull/34
- [x] Update Servo to use the new html5ever/markup5ever/xml5ever
    https://github.com/servo/servo/pull/32145
- [X] Update string_cache to phf 0.11
    https://github.com/servo/string-cache/pull/281
- [X] Release a new version of string_cache
[...]

PR #33434, opened 2024-09-13
script: update phf to 0.11

Gae24:
This one was missed when build dependencies were updated, removed entry in servo-tidy since it isn't duplicated anymore

---

- [x] These changes contribute to #31516 (GitHub issue number if applicable)

- [X] These changes do not require tests because cleanup

### DIFF
--- a/servo-tidy.toml
+++ b/servo-tidy.toml
@@ -60,8 +60,7 @@ packages = [
     # stuck on 0.8.4 with no new releases.
     "env_logger",
 
-    # Stylo uses phf 0.11, while script and other dependencies use 0.10.
-    "phf",
+    # Required until a new version of string-cache is released.
     "phf_generator",
     "phf_shared",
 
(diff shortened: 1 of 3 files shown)


servo-tidy.toml
```diff
@@ -61,7 +61,6 @@ packages = [
     "env_logger",
 
     # Stylo uses phf 0.11, while script and other dependencies use 0.10.
```
gterzian, line 63: Can this comment be removed now?

Gae24: There is still string_cache that depends on 0.10, do I simply remove the reference to script?

gterzian: Yes I think we should at least remove the reference to script, since this PR will make it untrue otherwise. If you are sure there is only one dependency left, we might also as well name it here.

Gae24: Done, checked with cargo tree.

gterzian (approved):
Thank you!

Issue #32720, opened 2024-07-08
Android: compilation fails for servoshell when LTO is enabled.


mukilan:
**Describe the bug:**
I've been trying to introduce a 'production' variant in our APK's gradle build definition to match the one in desktop binaries. This is so that logic such as signing the APK and other Android specific optimizations can be enabled for the APKs that we eventually publish to servo.org.

However, building servoshell for Android doesn't seem to work today with the `production` profile:
```
  = note: ld: error: duplicate symbol: pthread_atfork
          >>> defined at crtbegin_so.c
          >>>            /nix/store/qdy9ddp1dahja6vrk2a3zjh3q807k4hp-androidsdk/libexec/android-sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/../sysroot/usr/lib/arm-linux-androideabi/30/crtbegin_so.o:(pthread_atfork)
          >>> defined at jemalloc_sys.83be178053a84e9f-cgu.0
          >>>            /home/mukilan/dev/servo-emulator/target/armv7-linux-androideabi/production/deps/servoshell.jemalloc_sys-9b4b75f3bf1e2f30.jemalloc_sys.83be178053a84e9f-cgu.0.rcgu.o.rcgu.o:(.text.pthread_atfork+0x1)
          clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

This duplicate error is [caused by](https://github.com/tikv/jemallocator/issues/81) `jemalloc-sys` crate and upstream has already merged a [fix](https://github.com/tikv/jemallocator/pull/83), but the fix has not been published yet.

**To Reproduce:**
1. Run `./mach build --android --production`

**Platform:**
Android

PR #33435, opened 2024-09-13
android: publish nightly builds for aarch64

mukilan:
Publish the aarch64 apk and aar packages to both Github Releases
and S3 so that it can be linked from servo.org.

The focus is on getting a working version of the APK on the homepage,
so few issues are resolved with temporary solutions:
1) We publish the "release" profile instead of "production" since the
   latter will need changes in the gradle configuration (the changes
   required was previously blocked on #32720 which is now closed).
2) The scheme for the version code is simple and doesn't consider
   other factors such as API level and product variants discussed in the
   Android docs (https://developer.android.com/google/play/publishing/multiple-apks#VersionCodes)
   This should be fine for now as we don't publish to any store yet.

The change also makes it so that the 'Release nightly' workflow will
endup building all 4 variants for the Android target, but only aarch64
is uploaded. This is because GH Actions doesn't have a good way to skip
a specific job in a matrix and the additionally code complexity needed
to acheive it (either generating a JSON dynamically in a new job and using
`fromJSON` in the matrix definition or skipping each individual step
based on matrix.target and `inputs.upload`) didn't seem worth the cost
saved (this is executed only once a day).

---

- [x] These changes do not require tests because they modify just CI workflow definitions.

### DIFF
--- a/support/android/apk/build.gradle
+++ b/support/android/apk/build.gradle
@@ -1,3 +1,5 @@
+import java.text.SimpleDateFormat
+
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
     id 'com.android.application' version '8.0.1' apply false
@@ -78,4 +80,22 @@ ext.getSigningKeyInfo = { ->
     } else {
         return null
     }
+}
+
+ext {
+  // Generate unique version code based on the build date and time to support nightly
+  // builds.
+  //
+  // The version scheme is currently: yyyyMMddXX
+  // where
+  // yyyy is the 4 digit year (e.g 2024)
+  // MMdd is the 2 digit month followed by 2 digit day (e.g 0915 for September 15)
+  // XX is currently hardcoded to 00, but can be used to distinguish multiple builds within
+  // the same day.
+  //
+  // TODO: check if this interferes with caching of local builds and add option to use
+  // a static version.
+  def today = new Date()
+  def versionCodeString = new SimpleDateFormat("yyyyMMdd00").format(today)
+  generatedVersionCode = versionCodeString as int
 }
--- a/support/android/apk/servoapp/build.gradle
+++ b/support/android/apk/servoapp/build.gradle
@@ -17,8 +17,8 @@ android {
         applicationId "org.servo.servoshell"
         minSdk 30
         targetSdk 30
-        versionCode 1
-        versionName "1.0.0"
+        versionCode generatedVersionCode
+        versionName "0.0.1" // TODO: Parse Servo's TOML and add git SHA.
     }
 
     compileOptions {
--- a/support/android/apk/servoview/build.gradle
+++ b/support/android/apk/servoview/build.gradle
@@ -15,11 +15,12 @@ android {
     buildDir = rootDir.absolutePath + "/../../../target/android/gradle/servoview"
 
     ndkPath = getNdkDir()
+
     defaultConfig {
         minSdk 30
         targetSdk 30
-        versionCode 1
-        versionName "1.0"
+        versionCode generatedVersionCode
+        versionName "0.0.1" // TODO: Parse Servo's TOML and add git SHA.
     }
 
     compileOptions {
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -114,11 +114,21 @@ jobs:
           S3_UPLOAD_CREDENTIALS: ${{ secrets.S3_UPLOAD_CREDENTIALS }}
           NIGHTLY_REPO_TOKEN: ${{ secrets.NIGHTLY_REPO_TOKEN }}
           NIGHTLY_REPO: ${{ github.repository_owner }}/servo-nightly-builds
+      - name: Generate artifact attestation for APK
+        if: ${{ inputs.upload }}
+        uses: actions/attest-build-provenance@v1
+        with:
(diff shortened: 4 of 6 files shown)


jschwe:
Would you consider adding [artifact attestation](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries) for the android apk as part of this PR? I for one would feel more comfortable installing the apk from github releases, if I can verify beforehand that it was indeed built via github CI.

sagudev:
Why only upload aarch64 if we build for other targets too?

mukilan:
> Would you consider adding [artifact attestation](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries) for the android apk as part of this PR? I for one would feel more comfortable installing the apk from github releases, if I can verify beforehand that it was indeed built via github CI.

Sure, I've added it in [58e2b5d](https://github.com/servo/servo/pull/33435/commits/58e2b5d62731a114ad5bbff3e5a65f10cf03a641). I'm attesting only the artifacts created from the Nightly workflow based on [this recommendation](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#what-to-attest) from GitHub to not attest frequent CI builds used for testing.

mukilan:
> Why only upload aarch64 if we build for other targets too?

Primarily because we are planning to expose only aarch64 in servo.org's "Downloads" sections. Of the 4 targets, two are 32-bit which is not popular these days and we are not sure what level of support we will provide in the future. The remaining is x86-64 which is for emulators. For now, I don't see a good reason to add that to the download list on servo.org. Technically we can still have it on GH Releases but not link it from servo.org, but that could cause confusion, so I haven't done that in this PR, but I can change this if we think that could be useful.

support/android/apk/servoapp/build.gradle
```diff
@@ -17,8 +17,8 @@ android {
         applicationId "org.servo.servoshell"
         minSdk 30
         targetSdk 30
-        versionCode 1
-        versionName "1.0.0"
+        versionCode generatedVersionCode
+        versionName "0.0.1" // TODO: parse Servo's TOML and add git sha
```
mrobinson, line 21:
```suggestion
        versionName "0.0.1" // TODO: Parse Servo's TOML and add git SHA.
```

support/android/apk/servoview/build.gradle
```diff
@@ -15,11 +15,12 @@ android {
...
 
     ndkPath = getNdkDir()
+
     defaultConfig {
         minSdk 30
         targetSdk 30
-        versionCode 1
-        versionName "1.0"
+        versionCode generatedVersionCode
+        versionName "0.0.1" // TODO: parse Servo's TOML and add git sha
```
mrobinson, line 23:
```suggestion
        versionName "0.0.1" // TODO: Parse Servo's TOML and add git SHA.
```

support/android/apk/build.gradle
```diff
@@ -79,3 +81,22 @@ ext.getSigningKeyInfo = { ->
...
+  // HHmm is the time of the day (e.g 1720 for 5.20 pm)
+  //
+  // TODO: check if this interferes with caching of local builds and add option to use
+  // a static version.
+  def today = new Date()
+  def year = new SimpleDateFormat("yy").format(today)
+  def day = String.format("%03d", (new SimpleDateFormat("D").format(today) as int))
+  def time = new SimpleDateFormat("HHmm").format(today)
+  generatedVersionCode = (year + day + time) as int
+}
```
mrobinson, line 102: Why don't we do four digits for the year and use two digits for the version inside the day? So for instance for today: 2024091600, where the final two digits are for releases in the same day

mrobinson (approved):
Nice work!

support/android/apk/build.gradle
mukilan: I went with the current scheme to minimize the number of digits to keep the available version space large for future expansion (using potentially different schemes) and also to not rely on external state to distinguish multiple builds (like a counter). However, the cost of changing the scheme in backwards-incompatible manner is not too high, the user just needs to uninstall and reinstall the app, so I'll change the scheme as you've suggested (with 00 for the last digits for now).

mukilan: I've addressed this in [85be322](https://github.com/servo/servo/pull/33435/commits/85be322389ce71436e44f3a86f6357e9609e01f6)

mukilan:
Thank you for the review!

ChristophHaag:
> The remaining is x86-64 which is for emulators.

Apologies for butting in but it's not just for emulators. There are devices in existence that ship with android on x86, for example if/when servo ships openxr on android, you will get requests to run it on Magic Leap, which runs android on x86-64.

And there's not a big reason for it right now, but x86 android executables can also be run on ChromeOS including x86 Chroomebooks https://developer.android.com/topic/arc/device-support#overview or with waydroid on generic Linux distributions.

jdm:
>you will get requests to run it on Magic Leap, which runs android on x86-64.

That sounds like something that would still require a custom embedding (at least based on our previous magic leap support), rather than using a prebuilt nightly, even setting aside the argument about size of audience.

PR #33436, opened 2024-09-13
Add perfetto tracing events to fonts

atbrakhi:
This changes extends perfetto events to trace fonts as well

Co-authored-by:Delan Azabani <dazabani@igalia.com>
Signed-off-by: atbrakhi <atbrakhi@igalia.com>

---

### DIFF
--- a/components/fonts/font_cache_thread.rs
+++ b/components/fonts/font_cache_thread.rs
@@ -23,6 +23,7 @@ use style::values::computed::font::{
 };
 use style::values::computed::{FontStretch, FontWeight};
 use style::values::specified::FontStretch as SpecifiedFontStretch;
+use tracing::{span, Level};
 use webrender_api::{FontInstanceFlags, FontInstanceKey, FontKey};
 use webrender_traits::WebRenderFontApi;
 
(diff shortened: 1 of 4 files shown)


components/servo/lib.rs
```diff
@@ -1112,7 +1114,11 @@ impl WebRenderFontApi for WebRenderFontApiCompositorProxy {
             .send(CompositorMsg::Forwarded(ForwardedToCompositorMsg::Font(
                 FontToCompositorMsg::AddFont(sender, index, bytes_receiver),
             )));
-        let _ = bytes_sender.send(&data);
+        {
+            let span = span!(Level::TRACE, "add_font", servo_profiling = true);
```
delan, line 1118: `add_font` → `add_font send`?

components/fonts/font_cache_thread.rs
```diff
@@ -141,6 +149,9 @@ impl FontCache {
                             .font_data
                             .entry(identifier)
                             .or_insert_with(|| font_template.data());
+                        let span =
+                            span!(Level::TRACE, "FontTemplates send", servo_profiling = true);
```
delan, line 153: `FontTemplates` → `GetFontTemplates`? (oops didn’t see this one)

delan (approved):

components/fonts/font_cache_thread.rs
```diff
@@ -102,12 +103,19 @@ struct FontCache {
 }
 
 impl FontCache {
+    #[tracing::instrument(skip(self), fields(servo_profiling = true))]
```
jschwe, line 106: Sorry, this is maybe at bit late / the wrong place to ask, but what exactly does `fields(servo_profiling = true)` do? I'm asking because on OpenHarmony these events seem to also be logged via `hilog`, and it's kind of wierd seeing most log messages end with `; servo_profiling=true`. I've never seen it set to false, so I'm kind of wondering what it does.

atbrakhi: No worries! The idea behind adding the `fields(servo_profiling = true)` custom field is to help distinguish our own thread's profiling spans. Right now, we only send [our spans to hitrace](https://github.com/servo/servo/pull/33324), but if tracing ever merges spans from all threads, it could cause confusion. We could have used `name` instead of fields to differentiate them, but that would have made filtering a bit messier.

By adding the field, we can more easily separate our profiling spans from any other existing spans.

atbrakhi: I will add the PR to the MQ and if needed, we can discuss it further on Zulip

PR #33451, opened 2024-09-13
Do not remove extra columns at the end of the table

Loirooriol:
`<col>` and `<colgroup>` elements can be used to create extra columns that have no cell. We were removing these columns and column groups, but in general we shouldn't do that.

Now we will only remove them if the table has no row nor row group. matching WebKit and the expectations of some tests. But note that Gecko and Blink never remove them.

---

### DIFF
--- a/components/layout_2020/table/construct.rs
+++ b/components/layout_2020/table/construct.rs
@@ -271,8 +271,8 @@ impl TableBuilder {
     }
 
     pub fn finish(mut self) -> Table {
+        self.adjust_table_geometry_for_columns_and_colgroups();
         self.do_missing_cells_fixup();
-        self.remove_extra_columns_and_column_groups();
         self.reorder_first_thead_and_tfoot();
         self.do_final_rowspan_calculation();
         self.table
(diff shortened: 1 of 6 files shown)


components/layout_2020/table/construct.rs
```diff
@@ -287,26 +287,16 @@ impl TableBuilder {
...
-                .min(number_of_actual_table_columns);
-        }
-
-        if let Some(remove_from) = remove_from {
-            self.table.column_groups.truncate(remove_from);
+    /// than actually exist in the table. In that case, increase the size of the table.
+    ///
+    /// However, if the table has no row nor row group, remove the extra columns instead.
+    /// This matches WebKit, and some tests require it, but Gecko and Blink don't do it.
+    fn do_extra_columns_and_column_groups_fixup(&mut self) {
```
mrobinson, line 294: How about `remove_columns_and_column_groups_from _empty_table`?

components/layout_2020/table/layout.rs
```diff
@@ -799,7 +799,7 @@ impl<'a> TableLayout<'a> {
     /// Distribute width to columns, performing step 2.4 of table layout from
     /// <https://drafts.csswg.org/css-tables/#table-layout-algorithm>.
     fn distribute_width_to_columns(&self) -> Vec<Au> {
-        if self.table.slots.is_empty() {
+        if self.table.size.width.is_zero() {
```
mrobinson, line 802: It might be worth a comment here to say that a table might have columns, but no rows.

components/layout_2020/table/construct.rs
mrobinson: Or perhaps `remove_extra_columns_and_column_groups_if_table_has_no_rows`.

Loirooriol: But that's like a corner case that doesn't even match Geck and Blink. Usually what this function will do is increase the width for the extra columns:

```rs
self.table.size.width = self.table.size.width.max(self.table.columns.len());
```

mrobinson: How about `adjust_table_geometry_for_columns_and_colgroups` here where table geometry can be either the number of cells or the number of columns / colgroups?

mrobinson (approved):

PR #33459, opened 2024-09-14
ohos: Support loading resource files

jschwe:
* Support loading resource files
* Separates resource handling on OH from android.
* Assumes the OH app has the resource-files placed under `AppScope/resources/resfile`, which will cause the files to be copied to the app sandbox on the device.
* Allows us to load prefs.json from the filesystem, allowing easier editing of the prefs.
* Add a resource:// url shorthand, which allows loading local files from the resource sandbox of the application, without the user needing to remember / know the exact path to the sandbox.

Corresponding changes on the ArkTS side required, since I couldn't find a native method to determine the resource directory (besides hardcoding it): https://github.com/jschwe/ServoDemo/pull/8/files

---


jschwe:
try android+ohos: https://github.com/jschwe/servo/actions/runs/10865218669/job/30151441772 (still running)

jdm (commented):
I wonder if adding a protocol handler for resource urls would be cleaner instead of performing a manual translation as part of loading a URI.

jschwe:
> I wonder if adding a protocol handler for resource urls would be cleaner instead of performing a manual translation as part of loading a URI.

Do we currently have any protocol handlers? I guess if we want to support this in a broader context, then perhaps we should use the [`app://`](https://www.w3.org/TR/app-uri/) uri scheme, since that seems to match the use-case of accessing files packaged with an app.

jdm:
Yes, https://github.com/servo/servo/commit/1b48bd18aa855cc966869dd81530aa0da3eea4f3 added a servo: protocol for servoshell.

jschwe:
Split most of this PR into https://github.com/servo/servo/pull/33513 for bundling the resource files. This PR can then afterwards handle the `resource` protocol as suggested.

mrobinson:
@jschwe What's the status of this PR?

PR #33462, opened 2024-09-15
Properly track `rel` keywords for `<a>`/`<area>`/`<form>` elements

simonwuelker:
In #33345 I defined `LinkRelations` to track the keywords of a `"rel"` attribute on `<link>` elements. This PR makes the same change for all the other hyperlink-elements (`<a>`, `<area>` and `<form>`).

I also moved code that is shared between these hyperlink-elements (like the [`follow_the_hyperlink`](https://html.spec.whatwg.org/multipage/#following-hyperlinks-2) algorithm) into its own module, so `dom::htmlformelement` does not have to import code from `dom::htmlanchorelement`.

These changes fix some minor bugs about the case-insensitivity of `rel` keywords, for example
https://github.com/servo/servo/blob/9f2306f76095cf81d299d0c977490803f5703c75/components/script/dom/htmlanchorelement.rs#L628-L630
This made `get_element_noopener` return an incorrect result for `<a rel="nOopener">`.

[try-run wpt-2020](https://github.com/Wuelle/servo/actions/runs/10871676031)

---

### DIFF
--- a/components/script/dom/htmlanchorelement.rs
+++ b/components/script/dom/htmlanchorelement.rs
@@ -33,7 +33,7 @@ use crate::dom::mouseevent::MouseEvent;
 use crate::dom::node::{document_from_node, BindContext, Node};
 use crate::dom::urlhelper::UrlHelper;
 use crate::dom::virtualmethods::VirtualMethods;
-use crate::link_relations::{follow_hyperlink, LinkRelations};
+use crate::links::{follow_hyperlink, LinkRelations};
 
 #[dom_struct]
 pub struct HTMLAnchorElement {
--- a/components/script/dom/htmlanchorelement.rs
+++ b/components/script/dom/htmlanchorelement.rs
@@ -2,48 +2,46 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::cell::Cell;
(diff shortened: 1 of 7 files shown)


jdm (approved):
Looks great! Let's see if there are any tests for this.

jdm:
It would be nice to add some test coverage for the previously-missing functionality for form and area elements. Are you willing to do that?

Wuelle:
> It would be nice to add some test coverage for the previously-missing functionality for form and area elements. Are you willing to do that?

Yeah sure, I'll look into it. I believe if i add a wpt test here then @servo-wpt-sync will open an upstream PR right?

jdm:
That's right!

Wuelle:
Sorry about the delay on this. I'm having difficulties writing tests since servo seems to lack functionality in other areas (for example, an implicit noopener on `<area>` elements doesn't work with or without capital letters([wpt.fyi](https://wpt.fyi/results/html/semantics/links/links-created-by-a-and-area-elements/target_blank_implicit_noopener.html?label=master&product=servo)), and the entire `target_blank_implicit_noopener` test does nothing when i run it locally, even excluding `<area>`s)

Given that these changes are meant to unify the way we track `rel` attributes across elements rather than fix bugs, would it be okay to merge this without new tests?

jdm:
Yep, I'm ok with that if the testing situation is complicated!

PR #33466, opened 2024-09-16
servoshell: Update gilrs version, enable gamepad pref by default

msub2:
Now that the latest version of gilrs has released with the force feedback completion event included, we can finally unpin the commit and track regular releases. This also enables the gamepad pref by default now, as the feature should be stable going forwards.

---

- [x] These changes do not require tests because functionality should be equivalent to prior version.


jdm (approved):

msub2:
@jdm Should I try and upgrade the core-foundation deps in this PR as well or open a separate one?

msub2:
Alternatively I guess I could update servo-tidy.toml with a note about it for now

jdm:
That's going to be a long tail of other crates that depend on the older version, so go ahead and update the toml file.

msub2:
Hmm, there was a failed check but it hasn't been removed from merge queue yet for some reason. Check is from a gamepad extension interface, I'll update that in a bit

msub2:
@jdm I still need to update this one, waiting on the webxr ref space reset event one merging first since both of these modify interfaces.html

PR #33481, opened 2024-09-17
Update CustomEvent webidl interface

simonwuelker:
The [idl interface](https://dom.spec.whatwg.org/#interface-customevent) and servo's implementation had diverged.

* Extra arguments to initCustomEvent are now optional (fixes WPT `dom/events/CustomEvent.html`)
* The CustomEvent constructor is infallible
* `[Exposed=*]` (and the same for the "Event" interface since it's CustomEvent's parent.)

[try run](https://github.com/Wuelle/servo/actions/runs/10900636123)

---

- [X] There are tests for these changes (wpt)


gterzian (approved):
Thank you!

PR #33485, opened 2024-09-18
crypto: Add checks for BigInt/BigUint in getRandomValues

msub2:
`Crypto.getRandomValues()` was missing checks for BigInt64 and BigUint64. The only failing subtest now is for Float16Arrays, which I don't believe are exposed from mozjs currently.

---

### DIFF
--- a/components/script/dom/crypto.rs
+++ b/components/script/dom/crypto.rs
@@ -83,6 +83,8 @@ fn is_integer_buffer(array_type: Type) -> bool {
             Type::Uint16 |
             Type::Int16 |
             Type::Uint32 |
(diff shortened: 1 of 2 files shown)


jdm (approved):
Float16 does exist: https://doc.servo.org/mozjs/jsapi/enum.Type.html#variant.Float16

msub2:
Ah, my mistake, I'll see if I can expose it to make it constructable then

msub2:
Hmm, that might require a little more plumbing than I expected, that can probably be saved for a separate PR. This can probably be merged as is for now

PR #33487, opened 2024-09-18
dom: Append stream chunks in the correct order.

### LINKED ISSUE #33467: Investigate WPT mimesniff tests failing
While comparing WPT passes between Servo and Ladybird I noticed that Servo didn't seem to be passing any of the mimesniff tests, despite the functionality for doing so existing already. It looks like there's a JSON parse error that's tripping up Servo and preventing any of the subtests from running. Link for ref:

https://wpt.fyi/results/mimesniff/mime-types/parsing.any.html?label=master&product=servo&product=ladybird

jdm:
These changes address a logic bug when appending multiple chunks of data to a stream. This manifested intermittently if the initial chunk received over the network was large enough to encompass the full response; otherwise, subsequent chunks would be prepended instead of appended, yielding corrupted data.

---
- [x] These changes fix #33467


mrobinson (approved):
Great catch!

gterzian (approved):
Thank you!

PR #33488, opened 2024-09-18
android/ohos: Fix wrong production cfg

jschwe:
Our build script sets `servo_production` and not `production`.
See https://github.com/servo/servo/blob/632d83270498f6cb2e9d284503d86607f250b80e/ports/servoshell/build.rs#L39

---


mukilan (approved):
Good catch!

PR #33496, opened 2024-09-19
fetch: add Accept-Encoding header when the Range header exists

shanehandley:
Implements step 8.19 of the fetch spec for https://fetch.spec.whatwg.org/#http-network-or-cache-fetch.

Also some tidying of surrounding documentation an numbering fixes.

WPT results in my fork: https://github.com/shanehandley/servo/actions/runs/10937450208

---


jdm (approved):
Thanks!

Issue #31812, opened 2024-03-21
Unable to build servo as vendored dependency


happybeing:
**Describe the bug:**
I need to build offline and have been trying to build vendored versions of first the wry servo demo, then verso (which supersedes that) and now servo as I suspect the issue is really with servo.

**To Reproduce:**
I opened an issue with verso [here](https://github.com/european-browser/verso/issues/17), which has more detail than I can provide in this issue about exactly what I'm doing and how to reproduce this. To reproduce with servo, follow the steps I documented there but with servo rather than verso, but in a nutshell...

- First, clone to a cloud server, install servo dependencies (`./mach bootstrap`) although on my server (Ubuntu 20.04) some of that fails so I can't try building there.

- Use `cargo vendor` to generate vendored sources and changes for `config.toml`

- Use `rsync` to copy the servo repo and vendored sources to my laptop (Ubuntu 22.04.04 LTS).

- Edit `servo/.cargo/config.toml` to use vendored sources and prevent `cargo` from going online to update anything.

- Finally `./mach build`

With `verso` I hit and fix some different errors that are to do with servo (noted in the issue linked above), but may only be happening because I'm trying to build a different project than servo itself.

With `servo` the vendored build seems to be hitting an unrelated issue, that is currently being discussed on the zulip chat [here](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Build.20Issues/near/427922798). That reports a failure to find `stdc++`. As that isn't to do with vendoring I can't say for sure that a vendored build of servo will fail if that is fixed, but I can't test it until that and any follow on build issues are working ok. Perhaps I'll try again in a while, but for now I need to move forward and look for an alternative to using servo to make a simple browser.

servo is a great project but unwieldy and not something I can work with in my off-grid/mobile broadband connection, as with such a large repo, it proves impossible to do the remote git operations I can and do routinely everywhere else so far. BTW I did a lot of investigating to try and fix those issues before trying the vendored approach, so don't worry about suggesting ways around that!

**Platform:**
Cloud server (used only to clone and vendor): Ubuntu 20.04
Laptop (for vendored build): Ubuntu 22.04.04 LTS
cargo 1.74.1 (ecb9851af 2023-10-18)
rustc 1.74.1 (a28077b28 2023-12-04)

sagudev:
I used `cargo vendor`, then copied output in `.cargo/config.toml` and it worked normally. Maybe this happens when it's out of servo's tree (like verso), so I will test verso now.

[...]

PR #33502, opened 2024-09-20
use rerun-if-changed in script/build.rs

sagudev:
Per https://github.com/servo/servo/pull/33416#issuecomment-2363118154 I think we can avoid some reruns of script/build.rs by specifying `rerun-if-changed` for appropriate files (if not specified all crate files are checked).

---

- [x] These changes do not require tests because they are about rebuilding

### DIFF
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -23,7 +23,7 @@ fn main() {
     println!("cargo::rerun-if-changed=dom/bindings/codegen");
     println!("cargo::rerun-if-changed={}", css_properties_json.display());
     println!("cargo::rerun-if-changed=../../third_party/WebIDL/WebIDL.py");
-    // we assume no changes in third_party/ply
+    // NB: We aren't handling changes in `third_party/ply` here.
 
     let status = Command::new(find_python())
         .arg("dom/bindings/codegen/run.py")
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -16,11 +16,17 @@ fn main() {
     let start = Instant::now();
 
     let style_out_dir = PathBuf::from(env::var_os("DEP_SERVO_STYLE_CRATE_OUT_DIR").unwrap());
+    let css_properties_json = style_out_dir.join("css-properties.json");
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+
+    println!("cargo::rerun-if-changed=dom/webidls");
+    println!("cargo::rerun-if-changed=dom/bindings/codegen");
+    println!("cargo::rerun-if-changed={}", css_properties_json.display());
+    // FIXME: rerun-if-changed third_party/{ply, WebIDL}
 
     let status = Command::new(find_python())
         .arg("dom/bindings/codegen/run.py")
-        .arg(style_out_dir.join("css-properties.json"))
+        .arg(&css_properties_json)
         .arg(&out_dir)
         .status()
         .unwrap();
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -22,7 +22,8 @@ fn main() {
     println!("cargo::rerun-if-changed=dom/webidls");
     println!("cargo::rerun-if-changed=dom/bindings/codegen");
     println!("cargo::rerun-if-changed={}", css_properties_json.display());
-    // FIXME: rerun-if-changed third_party/{ply, WebIDL}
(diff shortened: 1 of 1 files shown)


components/script/build.rs
```diff
@@ -16,11 +16,17 @@ fn main() {
     let start = Instant::now();
 
     let style_out_dir = PathBuf::from(env::var_os("DEP_SERVO_STYLE_CRATE_OUT_DIR").unwrap());
+    let css_properties_json = style_out_dir.join("css-properties.json");
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
 
+    println!("cargo::rerun-if-changed=dom/webidls");
+    println!("cargo::rerun-if-changed=dom/bindings/codegen");
+    println!("cargo::rerun-if-changed={}", css_properties_json.display());
+    // FIXME: rerun-if-changed third_party/{ply, WebIDL}
```
mrobinson, line 25: Is there a reason this change cannot be made now?

sagudev: because we would need to go in parent directory and I am not sure cargo would be happy about this (this would also be change from current behavior). I think it would be more appropriate to just move ply and webidl back to script as this would also make vendoring work more properly: https://github.com/servo/servo/issues/31812.

mrobinson: 1. Moving the vendored sources is a different issue than this one and it isn't so simple as `cargo vendor` also requires us to figure out what to do with the `resources` directory. Do we event want to support `cargo vendor`?
2. Identifying a file from a different directory works fine (or else the dependency on the CSS properties file wouldn't work). For instance, I've tested this locally and this seems to work fine

```diff
diff --git a/components/script/build.rs b/components/script/build.rs
index b740c53364..3ab015504f 100644
--- a/components/script/build.rs
+++ b/components/script/build.rs
@@ -21,8 +21,8 @@ fn main() {
 
     println!("cargo::rerun-if-changed=dom/webidls");
     println!("cargo::rerun-if-changed=dom/bindings/codegen");
+    println!("cargo::rerun-if-changed=../../third_party/WebIDL/WebIDL.py");
     println!("cargo::rerun-if-changed={}", css_properties_json.display());
-    // FIXME: rerun-if-changed third_party/{ply, WebIDL}
 
     let status = Command::new(find_python())
         .arg("dom/bindings/codegen/run.py")
```

This is an improvement to what we have now and is better than adding a `TODO` which may never be solved. Note that identifying the `ply` directory here didn't work well, likely because `cargo` is picking up changes to Python compiled code caches. I think it isn't as important to identify `ply` anyway. Let's integrate this into the PR if it looks good to you.

sagudev: > Identifying a file from a different directory works fine (or else the dependency on the CSS properties file wouldn't work)

Indeed.

> Note that identifying the ply directory here didn't work well, likely because cargo is picking up changes to Python compiled code caches

there is also pycache in webidl folder, maybe we need to include only python files, will investigate this.

> I think it isn't as important to identify ply anyway

Indeed, as we do not patch it.

sagudev: > Moving the vendored sources is a different issue than this one and it isn't so simple as cargo vendor also requires us to figure out what to do with the resources directory. Do we event want to support cargo vendor?

Do we want to support cargo vendor? Probably. For the resource directory we could wrap it in crate that would provide all files, but this is indeed separate issue.

sagudev: > there is also pycache in webidl folder, maybe we need to include only python files, will investigate this.

Ah, you already only included only webidl.py.

components/script/build.rs
```diff
@@ -16,11 +16,18 @@ fn main() {
...
 
     let style_out_dir = PathBuf::from(env::var_os("DEP_SERVO_STYLE_CRATE_OUT_DIR").unwrap());
+    let css_properties_json = style_out_dir.join("css-properties.json");
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
 
+    println!("cargo::rerun-if-changed=dom/webidls");
+    println!("cargo::rerun-if-changed=dom/bindings/codegen");
+    println!("cargo::rerun-if-changed={}", css_properties_json.display());
+    println!("cargo::rerun-if-changed=../../third_party/WebIDL/WebIDL.py");
+    // we assume no changes in third_party/ply
```
mrobinson, line 26:
```suggestion
    // NB: We aren't handling changes in `third_party/ply` here.
```

mrobinson (approved):

PR #33504, opened 2024-09-20
webgpu: Support more `TextureFormat`s in `GPUCanvasContext.configure()`

sagudev:
We can present all textures formats that webrender supports.

Mappings were determined by checking how wr/wgpu formats maps to gl
https://github.com/gfx-rs/wgpu/blob/9b36a3e129da04b018257564d5129caff240cb75/wgpu-hal/src/gles/conv.rs#L10
https://github.com/servo/webrender/blob/1e73f384a7a86e413f91e9e430436a9bd83381cd/webrender/src/device/gl.rs#L4064

---


jdm (approved):

sagudev:
This is actually non spec compliant and will get "reverted" with https://github.com/servo/servo/pull/33521.

PR #33506, opened 2024-09-20
chore: Update wgpu

sagudev:
Update wgpu to https://github.com/gfx-rs/wgpu/commit/9f85f8aeea6c43c1a412bafc8fbcfb43aad0dd20

Most notable change is that ids do not have backend component anymore, so we only need one `IdentityHub` for all backends.

---

### DIFF
--- a/components/webgpu/wgpu_thread.rs
+++ b/components/webgpu/wgpu_thread.rs
@@ -636,19 +636,19 @@ impl WGPU {
                     WebGPURequest::RequestAdapter {
                         sender,
                         options,
-                        ids,
+                        adapter_id,
                     } => {
                         let global = &self.global;
                         let response = self
                             .global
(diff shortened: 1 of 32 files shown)


components/webgpu/wgpu_thread.rs
```diff
@@ -1320,3 +1312,26 @@ impl WGPU {
         }
     }
 }
+
+fn backends() -> wgt::Backends {
```
sagudev, line 1316: Maybe we can replace this with `wgt::Backends::all()`.

jdm (approved):

components/webgpu/wgpu_thread.rs
sagudev: yes we can!

PR #33508, opened 2024-09-21
bindings: Allow Guard to take multiple conditions, check for SecureContext in ConstructorEnabled

### LINKED ISSUE #33272: Conditionally exposed interface members are always accessible on visible parent objects
Discovered while investigating #27268. While these interfaces are properly prevented from being constructed/accessed directly, they are still accessible as a member on a parent object. The case I saw was `navigator.xr.test`, which returns an instance of XRTest and is gated by pref in the WebIDL, still being accessible even when `dom.webxr.test` is false.

msub2:
This updates the bindings codegen and allows Guard to take multiple conditions, as previously it was only taking one and improperly exposing some things (i.e. `XRSystem.test` was exposed even when the `dom.webxr.test` pref wasn't enabled because it passed the SecureContext check). It also adds a check for SecureContext inside the ConstructorEnabled function.

---

- [x] These changes fix #33272 


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2815,6 +2815,10 @@ def definition_body(self):
             assert isinstance(func, list) and len(func) == 1
             conditions.append(f"{func[0]}(aCx, aObj)")
 
+        secure = iface.getExtendedAttribute("SecureContext")
+        if secure:
+            conditions.append(f"unsafe {{\nlet in_realm_proof = AlreadyInRealm::assert_for_cx(aCx);\nGlobalScope::from_context(*aCx, InRealm::Already(&in_realm_proof)).is_secure_context()\n}}")
```
jdm, line 2820:
```suggestion
            conditions.append("""
unsafe {
    let in_realm_proof = AlreadyInRealm::assert_for_cx(aCx);
    GlobalScope::from_context(*aCx, InRealm::Already(&in_realm_proof))
        .is_secure_context()
}
""")
```

components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -1545,15 +1545,18 @@ def MemberCondition(pref, func, exposed, secure):
...
     if pref:
-        return f'Condition::Pref("{pref}")'
+        conditions.append(f'Condition::Pref("{pref}")')
     if func:
-        return f'Condition::Func({func})'
+        conditions.append(f'Condition::Func({func})')
     if exposed:
-        return [f"Condition::Exposed(InterfaceObjectMap::Globals::{camel_to_upper_snake(i)})" for i in exposed]
-    return "Condition::Satisfied"
+        conditions.extend([f"Condition::Exposed(InterfaceObjectMap::Globals::{camel_to_upper_snake(i)})" for i in exposed])
```
jdm, line 1556:
```suggestion
        conditions.extend([
            f"Condition::Exposed(InterfaceObjectMap::Globals::{camel_to_upper_snake(i)})"
            for i in exposed
        ])
```

jdm (approved):
I think these suggestions should fix the tidy errors.

PR #33512, opened 2024-09-21
ohos: Support product flavors

jschwe:
Support different product flavors (e.g. with different signing configurations) by supporting --flavor on ohos.
The flavor influences the package path for OH packages, so allow the flavor parameter on build + package + install.
This allows using mach to build for HarmonyOS by passing `--flavor=harmonyos` to build / package / install.

---

- [x] These changes fix packaging and installing for HarmonyOS

### DIFF
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -531,7 +531,11 @@ class CommandBase(object):
                                 help='Build in release mode without debug assertions'),
                 CommandArgument('--profile', group="Build Type",
                                 help='Build with custom Cargo profile'),
-                CommandArgument('--with-asan', action='store_true', help="Build with AddressSanitizer")
+                CommandArgument('--with-asan', action='store_true', help="Build with AddressSanitizer"),
+                CommandArgument(
+                    '--flavor', default=None,
+                    help='Product flavor to be used when packaging with Gradle/Hvigor (android/ohos).'
+                ),
             ]
 
         if build_configuration:
@@ -612,6 +616,7 @@ class CommandBase(object):
                     kwargs.pop('dev', None)
                     kwargs.pop('prod', None)
                     kwargs.pop('profile', None)
+                    self.flavor = kwargs.get('flavor', None)
 
                 if build_configuration:
                     self.configure_build_target(kwargs)
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -533,7 +533,7 @@ class CommandBase(object):
                                 help='Build with custom Cargo profile'),
                 CommandArgument('--with-asan', action='store_true', help="Build with AddressSanitizer"),
                 CommandArgument(
-                    '--flavor', default=None,
+                    '--flavor', default=None, group="Build Type",
                     help='Product flavor to be used when packaging with Gradle/Hvigor (android/ohos).'
                 ),
             ]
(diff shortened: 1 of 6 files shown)


jschwe:
CC @mukilan

mukilan:
Looks like the smoketest command is now broken due to this change:

```
TypeError: WebPlatformTestsCreator.smoketest() got an unexpected keyword argument 'flavor'
```

Perhaps the method definition needs a **kwargs to collect the unused arguments?

jschwe:
I guess the other option would be to only pass flavor if it is not None. But if adding a kwargs Parameter is acceptable, that sounds like less work on my side, so I wouldn't mind that.

jschwe:
It seems one needs to pop all kwargs in `configuration_decorator` . I'm not really that familiar with python and decorators, so I'm not really sure why or what the background here is.

python/servo/command_base.py
```diff
@@ -612,6 +616,8 @@ def configuration_decorator(self, *args, **kwargs):
                     kwargs.pop('dev', None)
                     kwargs.pop('prod', None)
                     kwargs.pop('profile', None)
+                    self.flavor = kwargs.get('flavor', None)
```
mukilan, line 619: I don't think this is the correct fix for the `smoketest` command issue. Removing the `flavor` from kwargs here will mean it never gets passed to the underlying `install`, `package` etc commands. 
i.e I suspect the below command will always behave as if flavor is None and hence the "default" flavor will be used
```
./mach install -r -ohos --flavor=hos
```

Also, we store the flavor value in `self.flavor` but we don't read that anywhere else.

mukilan (commented):
> It seems one needs to pop all kwargs in configuration_decorator . I'm not really that familiar with python and decorators, so I'm not really sure why or what the background here is.

We currently pop the 'release', 'dev', 'prod' and 'profile' keys because they are set by the `mach` CLI parsing logic, but then we collapse all the four parameters into a single `build_type` key by calling `configure_build_type` in the previous line. So when the actual decorated method (`def install` or `def package`) is called it won't see those individual options and just the 'build_type'.

> I guess the other option would be to only pass flavor if it is not None. But if adding a kwargs Parameter is acceptable, that sounds like less work on my side, so I wouldn't mind that.

I think it make sense to have `**kwargs` or even just add `flavor=None` in the smoketest method.  The latter is more explicit as it means flavor is logically supported for smoketest (we might want to test the different flavors) but currently not used.

python/servo/command_base.py
jschwe: > I suspect the below command will always behave as if flavor is None and hence the "default" flavor will be used

That indeed seems to be the case. Sorry, I should have tested that.

> Also, we store the flavor value in self.flavor but we don't read that anywhere else.

Ah, that seems to be a leftover from a different approach.

> I think it make sense to have **kwargs or even just add flavor=None in the smoketest method. The latter is more explicit as it means flavor is logically supported for smoketest (we might want to test the different flavors) but currently not used

I'm not sure if we already smoketest on android - but I guess the question would be, should `smoketest` / `run` install first (which would require knowing the flavor), or just assume that we already installed servo previously? Installing takes a bit of time, so I guess we might want to avoid installing unnecessarily.

mukilan: > I'm not sure if we already smoketest on android

It is [currently disabled](https://github.com/servo/servo/blob/a97afebdcc7a06201638c45a9c7e4b7aab25203a/.github/workflows/android.yml#L100), but the blockers have been addressed, so we should ideally do it.

> Installing takes a bit of time, so I guess we might want to avoid installing unnecessarily.

That is a good point. I guess we can have an `--install` option for such commands and not install by default.

mukilan (approved):

PR #33511, opened 2024-09-21
ohos: Add back and forward button to vendored app

jschwe:
25372340 added back and forward apis to the ohos servoshell. 157e28c5 vendored the ohos demo ArkTS app into the servo repo, but did not include the back and forward buttons, which were added upstream in the meantime.
This commit adds the missing back and forward buttons to the ArkTS code.

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors

PR #33513, opened 2024-09-21
ohos: Bundle resources into hap instead of including in the binary.

jschwe:
Bundle resource files into the .hap, so they are available as files in the application sandbox, instead of included
into the shared library.
This should slightly reduce the binary size in debug and release mode and thus loading time.

Builds ontop of #33511

---


mukilan (approved):

jschwe:
Hmm, I guess I either need to figure out how to get the resource directory on OH 4.1 or we would need to upgrade the SDK to the 5.0 beta

jschwe:
Okay, it now compiles for OH 4.1 (try ohos: https://github.com/jschwe/servo/actions/runs/10995081440/job/30525076524)

PR #33526, opened 2024-09-23
Add a `webxr` feature flag

### LINKED ISSUE #33464: Put webxr behind a feature
In order to create a servo frontend, it is necessary to create an EmbedderMethods object. To do this, you must implement a register_webxr() function which takes a webxr::MainThreadRegistry object.

https://github.com/servo/servo/blob/9f2306f76095cf81d299d0c977490803f5703c75/components/compositing/windowing.rs#L220

This requires including an entire other repo

`
webxr = { git = "https://github.com/servo/webxr" }
`

I am writing a servo frontend which cannot, even in principle, support webxr. My register_webxr function will always be blank, but I still must include the webxr crate just so I can implement the callback in which I do nothing. 

I suggest that webxr be made a default-enabled cargo "feature", and the register_webxr function elided when this feature is not enabled.

NwE0kmCE:
For now I've only added the `webxr` feature outside the `script` and `script_traits` crates, but once the `script` crate is broken up it should be pretty easy to add the feature there. The `canvas_traits` crate also can't have the feature until it's added to the `script` crate so holding off until that can be done. Also unsure if adding/which tests would be applicable here...

This is my first contribution to servo, so any feedback is much appreciated!

---

- [X] These changes fix #33464

### DIFF
--- a/components/canvas/webgl_thread.rs
+++ b/components/canvas/webgl_thread.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use std::borrow::Cow;
+#[cfg(feature = "webxr")]
 use std::collections::HashMap;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
@@ -18,8 +19,10 @@ use canvas_traits::webgl::{
     WebGLCreateContextResult, WebGLFramebufferBindingRequest, WebGLFramebufferId, WebGLMsg,
     WebGLMsgSender, WebGLProgramId, WebGLQueryId, WebGLReceiver, WebGLRenderbufferId,
     WebGLSLVersion, WebGLSamplerId, WebGLSender, WebGLShaderId, WebGLSyncId, WebGLTextureId,
-    WebGLVersion, WebGLVertexArrayId, WebXRCommand, WebXRLayerManagerId, YAxisTreatment,
+    WebGLVersion, WebGLVertexArrayId, YAxisTreatment,
 };
+#[cfg(feature = "webxr")]
+use canvas_traits::webgl::{WebXRCommand, WebXRLayerManagerId};
 use euclid::default::Size2D;
 use fnv::FnvHashMap;
 use half::f16;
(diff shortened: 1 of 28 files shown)


Wuelle:
> This is my first contribution to servo, so any feedback is much appreciated!

Hi! :wave: 

You need to sign-off your commits, you can do that by adding the `-s` flag to `git commit`.

components/canvas/webgl_thread.rs
```diff
@@ -3297,11 +3323,13 @@ impl Drop for WebXRBridgeManager {
     }
 }
 
+#[cfg(feature = "webxr")]
 struct WebXRBridgeContexts<'a> {
     contexts: &'a mut FnvHashMap<WebGLContextId, GLContextData>,
     bound_context_id: &'a mut Option<WebGLContextId>,
 }
 
+#[cfg(feature = "webxr")]
```
jdm, line 3332: My instinct is that it would be more clear to extract as many of these webxr-specific structures into a separate module that can have a single cfg applied to it, which will reduce the number of one-off cfgs needed in this file.

ports/servoshell/Cargo.toml
```diff
@@ -118,7 +119,7 @@ raw-window-handle = "0.6"
 shellwords = "1.0.0"
 surfman = { workspace = true, features = ["sm-x11", "sm-raw-window-handle-06"] }
 tinyfiledialogs = "3.0"
-webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "glwindow", "headless"] }
+webxr = { git = "https://github.com/servo/webxr", features = ["ipc", "glwindow", "headless"], optional = true }
```
jdm, line 122: I'm not convinced that making webxr an optional feature of servoshell is useful. servoshell is a simple test embedding, so nobody should be depending on it. I think servoshell should enable the libservo webxr feature unconditionally.

jschwe:
> Also unsure if adding/which tests would be applicable here...

I think perhaps we should add a job (or perhaps one job per platform) to CI which compiles `servo` with `--no-default-features`, just to see that disabling the webxr feature actually compiles.

wusyong:
@augustkline Are you still able to work on this? If you are too busy, I can help update the PR and resume the review again.

jdm:
Finished in #34241.

PR #33527, opened 2024-09-24
Use `ContentSizes::shrink_to_fit` when possible

Loirooriol:
And ensure that the minimum wins for malformed ContentSizes
whose min_content is larger than its max_content.

---

- [X] These changes do not require tests because there should be no change in behavior (assuming we never create malformed `ContentSizes`)


mrobinson (approved):

mrobinson:
Apologies for the long review wait. This looks great to me!

Issue #12054, opened 2016-07-01
HiDPI scaling on Linux


benwaffle:

metajack:
One thing you can do for now is use --device-pixel-ratio=2.0 (or 1.5 or 2.5 or whatever depending on what you want).

Do you happen to know what the normal method for doing this on Linux is? We do have support for this on Windows and Mac, although even there we could improve a bit, but mostly it is a matter of automating adjusting the device pixel ratio based on some system notification.

benwaffle:
On GNOME desktops, you can read the gsettings key `org.gnome.desktop.interface scaling-factor`

jbriggs-zvelo:
@benwaffle 
GNOME with Wayland is fancier because it does scaling per monitor. I am not up on how that's done though.

cmr:
On Wayland, you should listen to the wl_surface::enter/leave events (https://wayland.freedesktop.org/docs/html/apa.html#idp67102000), and take the scale from the wl_output::scale, and redraw with the new scale, using wl_surface::set_buffer_scale to match the output.

Note that this is only for crude scaling: you'd probably still want to set the device pixel ratio using the wl_output::geometry to get some correspondence between a CSS px (which is an angular measurement) and a display px.

`org.gnome.desktop.interface scaling-factor` isn't ever useful for an application to _read_ on Wayland, AFAIK.

benwaffle:
somehow weston gets overlapping multi-dpi monitors perfect

nox:
Is this still an issue?

Manishearth:
Yeah, a page with a bit of text and some newlines renders like this under 2x:

![image](https://user-images.githubusercontent.com/1617736/45086943-007d0d80-b122-11e8-9720-6e1ffdf8859f.png)


The recommended fix -- setting ` --device-pixel-ratio=2.0` -- makes the text larger but it's still in a corner

EyalRo:
(Wayland/Gnome) servo doesn't update the scaling factor when switching screens (e.g. if one screen has a scaling factor of 125% and the other of 100%, servo will pick one and keep it regardless of which screen it's moved to)

mukilan:
@EyalRo can you check if the below patch (mukilan/servo@321eb9c895af07d49ca0b266b337760d2ed9b162) works as expected for you:

```diff
commit 321eb9c895af07d49ca0b266b337760d2ed9b162
Author: Mukilan Thiyagarajan <mukilan@igalia.com>
[...]

PR #33529, opened 2024-09-24
servoshell: fix issues related to HiDPI

mukilan:
The current implementation has 3 main issues related to HiDPI:

1. When the window moves from a screen with scale factor of 1.5 to one
   with 1 and back to 1.5, the minibrowser toolbar actually ends up
   being scaled by a factor of 2.25 instead of 1.5. This is because we
   currently use the [set_pixels_per_point] method on egui's Context,
   but calling this with a value of `ppp` will modify egui's internal
   'zoom factor' to be:
   ```
   zoom_factor = ppp / native_points_per_pixel.
   ```
   where `native_points_per_pixel` is the window system scale factor.
   The idea is egui can calculate the final scale factor for translating
   its logical points to physical pixels as:
   ```
   points_per_pixel = zoom_factor * native_points_per_pixel
   ```
   where zoom_factor is a factor used for Ctrl+Plus, Ctrl+Minus
   behaviour. The problem is when we handle the ScaleFactorChanged winit
   event due to window moving between screens, the
   `native_points_per_pixel` still has the value of the previous
   screen's native scaling factor and not the current screen's factor.
   This seems to be the case even if we pass the ScaleFactorChanged
   event to egui before we call `set_pixels_per_point`.
2. The egui logic for handing Ctrl+Plus, Ctrl+Minus and Ctrl+0 doesn't
   interact well with servoshell's device-pixel-ratio CLI argument which
   allows the user to override the HiDPI factor. For example, Ctrl+0
   will cause egui to reset the zoom_factor to 1.0 instead of the
   override we wanted. Another issue is egui's Ctrl+Plus/Ctrl+Minus
   will scale the minibrowser in increments of 0.10 whereas
   Servo's own page zoom doesn't (it keeps multiplying by 1.1, so the
   actual increments are 0.1, 0.21. 0.33 etc)
3. The inital window size calculation on Linux currently assumes a scale
   factor of 1.0. This means the window doesn't have the expected
   default logical size of 1024*740 on HiDPI systems. On a screen with
   HiDPI factor of 1.5, the logical window size ends up being 682x493.

This change addresses all 3 issues:

For 1,  switch to the `set_zoom_factor` method of egui context
to avoid the issue with scaling by incorrect native_points_per_pixel.
To allow for the device-pixel-ratio override to work, we calculate the
actual zoom_factor as `device-pixel-ratio / window's scaling factor`.

For 2, disable egui's handling of Ctrl+Plus, Ctrl+Minus, Ctrl-0
shortcuts. It is unclear whether the current behaviour of scaling
both the toolbar and the web page was intentional, or just an accident.
This behaviour is also different from other browser where page zoom
doesn't scale the GUI, so it doesn't seem like a regression to me.

For 3, use LogicalSize type of winit which lets the physical size
calulation to be handled by winit using the windows's actual HiDPI
factor instead of hardcoded 1.0.

[set_pixels_per_point]: https://github.com/emilk/egui/blob/1603f0581882c4ca299db1d6efdba9148ebcddb6/crates/egui/src/context.rs#L1886

**EDIT: I've tested this on Linux (KDE+Wayland) and Windows 11, but not MacOS.**

---

- [x] These changes potenially fix the issue [reported in this comment](https://github.com/servo/servo/issues/12054#issuecomment-2359639134).

### DIFF
--- a/ports/servoshell/desktop/minibrowser.rs
+++ b/ports/servoshell/desktop/minibrowser.rs
@@ -88,6 +88,13 @@ impl Minibrowser {
         // Adapted from https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui_glow/examples/pure_glow.rs
         #[allow(clippy::arc_with_non_send_sync)]
         let context = EguiGlow::new(events_loop.as_winit(), Arc::new(gl), None);
+
+        // Disable the builtin egui handlers for the Ctrl+Plus, Ctrl+Minus and Ctrl+0
+        // shortcuts as they don't work well with servoshell's `device-pixel-ratio` CLI argument.
+        context
+            .egui_ctx
+            .options_mut(|options| options.zoom_with_keyboard = false);
+
         let widget_surface_fbo = match rendering_context.context_surface_info() {
             Ok(Some(info)) => NonZeroU32::new(info.framebuffer_object).map(NativeFramebuffer),
             Ok(None) => panic!("Failed to get widget surface info from surfman!"),
--- a/ports/servoshell/desktop/app.rs
+++ b/ports/servoshell/desktop/app.rs
@@ -249,15 +249,18 @@ impl App {
                         // Intercept any ScaleFactorChanged events away from EguiGlow::on_window_event, so
                         // we can use our own logic for calculating the scale factor and set egui’s
                         // scale factor to that value manually.
-                        let effective_scale_factor = window.hidpi_factor().get();
+                        let desired_scale_factor = window.hidpi_factor().get();
+                        let effective_egui_zoom_factor = desired_scale_factor / scale_factor as f32;
+
                         info!(
-                            "window scale factor changed to {}, setting scale factor to {}",
-                            scale_factor, effective_scale_factor
+                            "window scale factor changed to {}, setting egui zoom factor to {}",
+                            scale_factor, effective_egui_zoom_factor
                         );
+
                         minibrowser
                             .context
                             .egui_ctx
-                            .set_pixels_per_point(effective_scale_factor);
+                            .set_zoom_factor(effective_egui_zoom_factor);
 
                         // Request a winit redraw event, so we can recomposite, update and paint
                         // the minibrowser, and present the new frame.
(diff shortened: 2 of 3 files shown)


ports/servoshell/desktop/minibrowser.rs
```diff
@@ -88,6 +88,13 @@ impl Minibrowser {
         // Adapted from https://github.com/emilk/egui/blob/9478e50d012c5138551c38cbee16b07bc1fcf283/crates/egui_glow/examples/pure_glow.rs
         #[allow(clippy::arc_with_non_send_sync)]
         let context = EguiGlow::new(events_loop.as_winit(), Arc::new(gl), None);
+
+        // Disable the builtin egui handlers for the Ctrl+Plus, Ctrl+Minus nd Ctrl+0 shortcuts as
+        // they don't work well with servoshell's `device-pixel-ratio` CLI argument.
```
mrobinson, line 93:
```suggestion
        // Disable the builtin egui handlers for the Ctrl+Plus, Ctrl+Minus and Ctrl+0
        // shortcuts as they don't work well with servoshell's `device-pixel-ratio` CLI argument.
```

mrobinson (approved):

mukilan:
Seems like the merge queue failure is due to a network issue:
```
Root directory input is valid!
Attempt 1 of 5 failed with error: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact. Retrying request in 3000 ms...
Attempt 2 of 5 failed with error: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact. Retrying request in 4522 ms...
Attempt 3 of 5 failed with error: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact. Retrying request in 8222 ms...
Attempt 4 of 5 failed with error: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact. Retrying request in 13257 ms...
Error: Failed to CreateArtifact: Failed to make request after 5 attempts: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact
```

I'll add this back to the queue.

Issue #30849, opened 2023-12-11
Clean Shutdown of Servo


gterzian:
Requires further investigation, but some initial notes:

1. `Pipeline` doesn't wait on an ACK from script [when sending the `ExitPipeline` msg](https://github.com/servo/servo/blob/8713954a533da7f4e7ec0877b9385908596b5230/components/constellation/pipeline.rs#L409).
2. Neither does an `EventLoop` wait on an ACK from script [when sending the `ExitScriptThread` msg](https://github.com/servo/servo/blob/8713954a533da7f4e7ec0877b9385908596b5230/components/constellation/event_loop.rs#L25)
3. There is some [workflow with the BHM](https://github.com/servo/servo/blob/8713954a533da7f4e7ec0877b9385908596b5230/components/constellation/constellation.rs#L2653) involving a `closing` boolean, which is useful if the script-thread is blocking inside a script execution, however it's not good enough to ensure clean shutdown, because a script-thread could easily be already handling a message by the time it is set. The constellation does wait on an ACK from the BHM, which implies that the closing flag has been set, however that doesn't prevent script from handling one last message if past the point of checking the flag(it may be that the constellation [sends it's reply to the IPC](https://github.com/servo/servo/blob/8713954a533da7f4e7ec0877b9385908596b5230/components/constellation/constellation.rs#L1850), and then shutdowns before script is scheduled to receive it). 

Any way, requires more investigation, but looking at the panic in https://github.com/servo/servo/issues/24066 it seems clear that the problem is not that the IPC returns `None`, it's that it returns an error because the sender has already been dropped(such as `Io(Custom { kind: ConnectionReset, error: "No senders exist for this port." })` or in `Disconnected` in  https://github.com/servo/servo/issues/27146). 

The last issue linked above https://github.com/servo/servo/issues/22154 appears unrelated actually(a comment in the issue seems to correctly identify a separate problem with trying to send a message layout after it has shutdown).

_Originally posted by @gterzian in https://github.com/servo/servo/issues/27171#issuecomment-1734105018_

mrobinson:
A common problem that we have faced is that when a task panics, other tasks wait forever for a reply to an IPC message. It seems that clean shutdown should also include cases where tasks disappear.

gterzian:
> A common problem that we have faced is that when a task panics, other tasks wait forever for a reply to an IPC message.

[...]

PR #33530, opened 2024-09-24
fonts: Use `IpcSharedMemory` to send font data

### LINKED ISSUE #10230: Don't copy large Arcs over IPC
Related to https://github.com/servo/servo/issues/9555, part of the cause (fixing this is not a complete solution).

https://github.com/servo/servo/blob/b97ffffb48080a0b4769f8609a27a68145042945/components/gfx/font_cache_thread.rs#L154

We're sending an `Arc<FontTemplateData>` over IPC here, where `FontTemplateData` is a largish bucket of bytes. This is probably a holdover from the pre-IPC days.

This makes sense if we were using std's MPSC channels, but the `Arc` is redundant over IPC since a completely new `Arc` will be created on the other side and the entire bucket of bytes is copied over through the shared memory. The copying and shmem allocation here is pretty wasteful (and is probably causing the mysterious "out of buffers" issue leading to that panic.

I'm not sure if there is a solution that lets us share memory across processes safely. I think you can create an `Arc`-like abstraction with shm-backed storage, after all, shm itself is "reference counted" by the kernel and we can mark shm as readonly.

(How useful this would be depends on how the font cache thread determines when to eject an item from its cache. I haven't yet looked at this code; it fiddles with weak pointers.)

cc @pcwalton @antrik

mrobinson:
This changes modifes the way that font data is sent over IPC channels.
Instead of serializing the data or sending it via IPC byte senders, font
data is copied into shared memory and a copy of the handle is sent over
the channel.

There is also the idea of sending the file handle of the on disk data of
system fonts. This could be implemented as a further followup once there
is an abstraction in `ipc-channel` over file handles.

To accomplish this, a `FontData` abstraction is added, which also allows
caching an in-memory shared `Arc<Vec<u8>>` version of the data (neeeded
by some APIs). This could also be a place for caching font tables in the
future.

Finally, the `FontCacheThread` is renamed to the `SystemFontService`
while the proxy for this is now named `SystemFontServiceProxy`.

Fixes #10230.

---

- [x] These changes do not require tests because they should not change testable behavior.

### DIFF
--- a/components/fonts/font.rs
+++ b/components/fonts/font.rs
@@ -27,14 +27,14 @@ use style::values::computed::{FontStretch, FontStyle, FontWeight};
 use unicode_script::Script;
 use webrender_api::{FontInstanceFlags, FontInstanceKey};
 
-use crate::font_cache_thread::{FontIdentifier, FontSource};
 use crate::font_context::FontContext;
 use crate::font_template::{FontTemplateDescriptor, FontTemplateRef, FontTemplateRefMethods};
 use crate::platform::font::{FontTable, PlatformFont};
 pub use crate::platform::font_list::fallback_font_families;
+use crate::system_font_service::{FontIdentifier, SystemFontServiceProxyTrait};
 use crate::{
-    ByteIndex, EmojiPresentationPreference, FallbackFontSelectionOptions, GlyphData, GlyphId,
-    GlyphStore, Shaper,
+    ByteIndex, EmojiPresentationPreference, FallbackFontSelectionOptions, FontData, GlyphData,
+    GlyphId, GlyphStore, Shaper,
 };
 
 #[macro_export]
@@ -66,11 +66,13 @@ pub trait PlatformFontMethods: Sized {
     fn new_from_template(
         template: FontTemplateRef,
         pt_size: Option<Au>,
+        data: &Arc<Vec<u8>>,
     ) -> Result<PlatformFont, &'static str> {
-        let data = template.data();
+        let template = template.borrow();
+
         let face_index = template.identifier().index();
-        let font_identifier = template.borrow().identifier.clone();
-        Self::new_from_data(font_identifier, data, face_index, pt_size)
+        let font_identifier = template.identifier.clone();
+        Self::new_from_data(font_identifier, data.clone(), face_index, pt_size)
     }
 
     fn new_from_data(
@@ -218,6 +220,7 @@ impl malloc_size_of::MallocSizeOf for CachedShapeData {
 #[derive(Debug)]
 pub struct Font {
     pub handle: PlatformFont,
+    pub data: Arc<FontData>,
     pub template: FontTemplateRef,
     pub metrics: FontMetrics,
     pub descriptor: FontDescriptor,
@@ -251,13 +254,19 @@ impl Font {
     pub fn new(
         template: FontTemplateRef,
         descriptor: FontDescriptor,
+        data: Arc<FontData>,
         synthesized_small_caps: Option<FontRef>,
     ) -> Result<Font, &'static str> {
-        let handle = PlatformFont::new_from_template(template.clone(), Some(descriptor.pt_size))?;
+        let handle = PlatformFont::new_from_template(
+            template.clone(),
+            Some(descriptor.pt_size),
+            data.as_arc(),
+        )?;
         let metrics = handle.metrics();
 
         Ok(Font {
             handle,
+            data,
             template,
             shaper: OnceLock::new(),
             descriptor,
@@ -526,7 +535,7 @@ impl FontGroup {
     /// `codepoint`. If no such font is found, returns the first available font or fallback font
     /// (which will cause a "glyph not found" character to be rendered). If no font at all can be
     /// found, returns None.
-    pub fn find_by_codepoint<S: FontSource>(
+    pub fn find_by_codepoint<S: SystemFontServiceProxyTrait>(
         &mut self,
         font_context: &FontContext<S>,
         codepoint: char,
@@ -598,7 +607,10 @@ impl FontGroup {
     }
 
     /// Find the first available font in the group, or the first available fallback font.
-    pub fn first<S: FontSource>(&mut self, font_context: &FontContext<S>) -> Option<FontRef> {
+    pub fn first<S: SystemFontServiceProxyTrait>(
+        &mut self,
+        font_context: &FontContext<S>,
+    ) -> Option<FontRef> {
         // From https://drafts.csswg.org/css-fonts/#first-available-font:
         // > The first available font, used for example in the definition of font-relative lengths
         // > such as ex or in the definition of the line-height property, is defined to be the first
@@ -629,7 +641,7 @@ impl FontGroup {
         font_predicate: FontPredicate,
     ) -> Option<FontRef>
     where
-        S: FontSource,
+        S: SystemFontServiceProxyTrait,
         TemplatePredicate: Fn(FontTemplateRef) -> bool,
         FontPredicate: Fn(&FontRef) -> bool,
     {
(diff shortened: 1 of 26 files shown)


components/fonts/font_store.rs
```diff
@@ -3,25 +3,72 @@
...
+/// (`Arc<Vec<u8>>`), the data is lazily copied out of shared memory once per
+/// [`FontData`].
+#[derive(Debug, Deserialize, Serialize)]
+pub struct FontData {
+    /// The data of this font in shared memory. Suitable for sending across IPC channels.
+    shared_memory: Arc<IpcSharedMemory>,
+    /// A lazily-initialized copy of the data behind an [`Arc`] which can be used when
+    /// passing it to various APIs.
+    #[serde(skip)]
+    arc: OnceLock<Arc<Vec<u8>>>,
```
gterzian, line 38: I would say `Arc<OnceLock<_>>` would make more sense, which would allow the arc in `Arc<FontFata>` to be removed. 

By the way I don't understand the need for `Arc<Vec<u8>>`, why not just an `Rc<>`? (unlike the shared mem, the vec is not send or sync).

components/fonts/font_cache_thread.rs
```diff
@@ -417,52 +448,33 @@ impl From<&FontFaceRuleData> for CSSFontFaceDescriptors {
...
-
-        FontCacheThread { chan }
-    }
-
+impl SystemFontServiceProxy {
     pub fn exit(&self) {
         let (response_chan, response_port) = ipc::channel().unwrap();
-        self.chan
+        self.sender
+            .lock()
```
gterzian, line 455: What are the other threads with access to a sender? Only script right?

I think instead of the ping mechanism, we should ensure that the threads other than the constellation(which is the only calling `exit` right?) have already been shut-down before, which I think is actually the case already.

components/fonts/font_cache_thread.rs
```diff
@@ -316,20 +338,29 @@ pub trait FontSource: Clone {
...
 /// The public interface to the font cache thread, used by per-thread `FontContext` instances (via
 /// the `FontSource` trait), and also by layout.
-#[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct FontCacheThread {
-    chan: IpcSender<Command>,
+#[derive(Debug)]
+pub struct SystemFontServiceProxy {
+    sender: ReentrantMutex<IpcSender<Command>>,
+    templates: RwLock<HashMap<FontTemplateCacheKey, Vec<FontTemplateRef>>>,
+    data_cache: RwLock<HashMap<FontIdentifier, Arc<FontData>>>,
```
gterzian, line 363: I would move the arc to inside here, and drop the mutex around the sender(unless somehow necessary), and remove the arc around the proxy elsewhere.

components/fonts/font_cache_thread.rs
gterzian: Actually, depends on https://github.com/servo/servo/issues/30849

gterzian (approved):
LGTM, with a few points on the use of arc's which I think can be addressed.

components/fonts/font.rs
```diff
@@ -66,11 +66,13 @@ pub trait PlatformFontMethods: Sized {
     fn new_from_template(
         template: FontTemplateRef,
         pt_size: Option<Au>,
+        data: &Arc<Vec<u8>>,
```
gterzian, line 69: By the way I don't understand the need for `Arc<Vec<u8>>` to be ret, why not just an `Rc<>`? (unlike the shared mem, the vec is not send or sync so I see no need for an arc).

components/fonts/font_cache_thread.rs
mrobinson: > What are the other threads with access to a sender? Only script right?

I think that canvas threads also has access to this now, but probably that will need to be proxied through the `Document` / script as well ultimately.

> I think instead of the ping mechanism, we should ensure that the threads other than the constellation(which is the only calling `exit` right?) have already been shut-down before, which I think is actually the case already.

This seems like a good goal once #30849 is taken care of and the change above is made.

components/fonts/font_cache_thread.rs
mrobinson: This is necessary because the proxy needs shared ownership and interior mutability. It's owned by each per-Document `FontContext` and represents the shared cache for all threads running in the content process.

mrobinson: That said, if we can verify that only a single thread in the content process is using the content process' `SystemFontServiceProxy`, then we could probably convert this into something behind an `Rc` instead of an `Arc`.

components/fonts/font_store.rs
gterzian: Ok, thanks for the info.

components/fonts/system_font_service.rs
```diff
@@ -90,80 +89,74 @@ struct ResolvedGenericFontFamilies {
...
+        }
+    }
+}
+
+impl SystemFontService {
+    pub fn spawn(webrender_api: Box<dyn WebRenderFontApi + Send>) -> SystemFontServiceProxySender {
+        let (sender, receiver) = ipc::channel().unwrap();
+
+        thread::Builder::new()
+            .name("FontCache".to_owned())
```
mukilan, line 122: Perhaps this can be changed to `SystemFontService`.

components/fonts/font_cache_thread.rs
mrobinson: Additionally, I'll note here that putting the `RwLock` around each member is meant to reduce contention if two threads are trying to query system fonts at once.

gterzian: Ok, perhaps good for a follow-up investigation then.

components/fonts/font.rs
mrobinson: This could ultimately be `&[u8]`, I think.

components/fonts/system_font_service.rs
```diff
@@ -486,51 +498,59 @@ impl FontSource for FontCacheThread {
...
         let reply = response_port.recv();
-        if reply.is_err() {
-            let font_thread_has_closed = self.chan.send(Command::Ping).is_err();
+        let Ok(reply) = reply else {
+            let font_thread_has_closed = self.sender.lock().send(Command::Ping).is_err();
             assert!(
                 font_thread_has_closed,
                 "Failed to receive a response from live font cache"
             );
             panic!("Font cache thread has already exited.");
```
mukilan, line 533:
```suggestion
            panic!("System font service has already exited.");
```

mukilan (approved):

components/fonts/font.rs
mrobinson: Ah, I was wrong sorry. It seems the reason is that the Windows platform font handle takes an `Arc`: https://docs.rs/dwrote/latest/dwrote/struct.FontFile.html#method.new_from_data

components/fonts/system_font_service.rs
```diff
@@ -90,80 +89,74 @@ struct ResolvedGenericFontFamilies {
...
-                        let span = span!(
-                            Level::TRACE,
-                            "GetFontTemplates send",
-                            servo_profiling = true
-                        );
-                        let _span = span.enter();
-                        let _ = bytes_sender.send(data);
-                    }
+                    let _ =
+                        result_sender.send(self.get_font_templates(font_descriptor, font_family));
```
gterzian, line 152: By the way, this, combined with [this recv](https://github.com/servo/servo/pull/33530/files#diff-5345e8d47974e67c3ddf58abd80db3fb9dd8bd5b5fb948de9d3a23bd3904ea62L500), is a potential source of deadlock, because of https://github.com/servo/servo/blob/dbd1666b17b22130bcce75e6255f376e6fbe256c/components/constellation/constellation.rs#L75

components/fonts/font.rs
gterzian: Ok, thanks again for the info.

components/fonts/system_font_service.rs
gterzian: Actually, since the other thread is blocking on the receive, and hence always ready to receive whatever is sent here, it should be ok.

mrobinson:
@mukilan @gterzian Thanks for the reviews. I think I have addressed all of your comments now.

PR #33532, opened 2024-09-24
Replace IPC channel with GenericChannel in Bluetooth

### LINKED ISSUE #32411: [bluetooth] Replace IPC channel with GenericChannel

lwz23:
This PR replaces the IPC channel with `GenericChannel` in the Bluetooth component, as requested in issue #32411. 

---
Fixes #32411

- [X] These changes fix #32411

- [X] These changes do not require tests because they involve internal channel replacement and existing tests already cover the affected areas.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -462,6 +462,7 @@ dependencies = [
 name = "bluetooth"
 version = "0.0.1"
 dependencies = [
+ "base",
  "bitflags 2.6.0",
  "bluetooth_traits",
  "blurdroid",
--- a/components/bluetooth/Cargo.toml
+++ b/components/bluetooth/Cargo.toml
@@ -12,6 +12,7 @@ name = "bluetooth"
 path = "lib.rs"
 
 [dependencies]
+base = { workspace = true }
 bitflags = { workspace = true }
 bluetooth_traits = { workspace = true }
 blurmock = { version = "0.1.2", optional = true }
--- a/components/bluetooth/lib.rs
+++ b/components/bluetooth/lib.rs
@@ -19,6 +19,7 @@ use std::string::String;
 use std::thread;
 use std::time::Duration;
 
+use base::generic_channel::{GenericReceiver, GenericSender};
 use bitflags::bitflags;
 use bluetooth_traits::blocklist::{uuid_is_blocklisted, Blocklist};
 use bluetooth_traits::scanfilter::{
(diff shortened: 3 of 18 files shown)


mrobinson (commented):
I have some doubts about this one. I'm not sure it is super useful to obscure whether or not the channel goes across the process boundary. It is really useful to know that, for example, an IPC channel is being used and has a much higher overhead.

wusyong:
@lwz23 sorry for the late reply. We have a similar PR in #32416 and we decided it's probably not useful to add this.

lwz23:
@wusyong Ok, I also see this PR. And it says "This PR only fix first half of the issue. Perhaps the title should be replace IPC channel with generic channel in BluetoothManager." so I thought I complete the whole two tasks. If it can't be merged it doesn't matter, I just want to be familar with this project and try other more complex task :)

lwz23:
I've also noticed that IPC channels are used in modules other than bluetooth. Is there a need to replace this with generic channel and release PR?

mrobinson:
> I've also noticed that IPC channels are used in modules other than bluetooth. Is there a need to replace this with generic channel and release PR?

For now, if I understand correctly, the plan is to use the `GenericChannel` when a piece of code communicates with both IPC channels and crossbeam channels.

PR #33534, opened 2024-09-24
webgpu: destroy GPUTexture without erroring

sagudev:
This matches the spec: https://www.w3.org/TR/webgpu/#dom-gputexture-destroy

Splited from https://github.com/servo/servo/pull/33521 because there is a lot of new PASSes.

---

### DIFF
--- a/components/webgpu/ipc_messages/recv.rs
+++ b/components/webgpu/ipc_messages/recv.rs
@@ -151,10 +151,7 @@ pub enum WebGPURequest {
     },
     DestroyBuffer(id::BufferId),
     DestroyDevice(id::DeviceId),
-    DestroyTexture {
-        device_id: id::DeviceId,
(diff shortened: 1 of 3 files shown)


components/webgpu/ipc_messages/recv.rs
```diff
@@ -152,7 +152,6 @@ pub enum WebGPURequest {
     DestroyBuffer(id::BufferId),
     DestroyDevice(id::DeviceId),
     DestroyTexture {
-        device_id: id::DeviceId,
         texture_id: id::TextureId,
     },
```
sagudev, line 156: `DestroyTexture(id::TextureId)`

DavidYummy:
All the rizzlets in Ohio Rizz mewing fan

gterzian (approved):
LGTM

PR #33540, opened 2024-09-25
fonts: Make fast shaping determination platform-independent

mrobinson:
This makes the determination of whether or not to use fast shaping
platform independent. Previously it was less stringent for Windows,
leading to using it in cases where a font had a GSUB or GPOS table --
which broke proper shaping.

In addition, the test is made platform independent and expanded to be
more complete.

Finally, comments are added indicating that "fast shaping" will be
removed.

---


components/fonts/font.rs
```diff
@@ -237,6 +235,14 @@ pub struct Font {
     /// essentially equivalent to whether or not we use it for emoji presentation.
     /// This is cached, because getting table data is expensive.
     has_color_bitmap_or_colr_table: OnceLock<bool>,
+
+    /// Whether or not this font can do fast shaping, ie whether or not it has
+    /// a kern table, but no GSUB and GPOS tables. When this is true, Servo will
+    /// shape this text without using Harfbuzz.
```
mukilan, line 241:
```suggestion
    /// shape horizontal LTR text without using Harfbuzz.
```

mukilan (approved):

components/fonts/font.rs
mrobinson: Cool. That seems like a good clarification.

PR #33541, opened 2024-09-25
fonts: Simplify `FontContext` in two ways that affect the `FontContext` unit test

mrobinson:
This is done by no longer forwarding compositor-bound messages through
SystemFontService and making `FontContext` non-generic:

- Messages from the `FontContext` to the `Compositor` no longer need to be
  forwarded through the `SystemFontService`. Instead send these messages
  directly through the script IPC channel to the `Compositor`.

- Instead of adding a mock `SystemFontServiceProxy`, simply implement a
  mock `SystemFontService` on the other side of an IPC channel in the
  `font_context` unit test. This allows making `FontContext`
  non-generic, greatly simplifying the code. The extra complexity moves
  into the unit test.

These changes necessitate adding a new kind of `FontIdentifier`,
`FontIdentifier::Mock` due to the fact that local fonts have
platform-specific identifiers. This avoids having to pretend like the
system font service can have web fonts -- which was always a bit of a
hack.

These two changes are combined into one PR because they both require
extensive and similar chages in the `FontContext` unit test which
depended on the details of both of them.

---

### DIFF
--- a/components/fonts/system_font_service.rs
+++ b/components/fonts/system_font_service.rs
@@ -24,6 +24,7 @@ use style::values::computed::font::{
 };
 use style::values::computed::{FontStretch, FontWeight};
 use style::values::specified::FontStretch as SpecifiedFontStretch;
+use style::Atom;
 use tracing::{span, Level};
 use webrender_api::{FontInstanceFlags, FontInstanceKey, FontKey};
 use webrender_traits::WebRenderFontApi;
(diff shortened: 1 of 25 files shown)


components/fonts/system_font_service.rs
```diff
@@ -41,26 +42,27 @@ use crate::FontData;
 pub enum FontIdentifier {
     Local(LocalFontIdentifier),
     Web(ServoUrl),
+    Mock(Atom),
```
mukilan, line 45: Would it make sense to disable this variant for non-test builds?
```suggestion
    #[cfg(test)] Mock(Atom),
```

Also, instead of having the Mock variant, would it make sense to have a `fn new_mock(id: String) -> LocalFontIdentifier` method for each platform's LocalFontIdentifier impl? It does involve adding more code.

mukilan (approved):
Nice!

components/fonts/system_font_service.rs
mrobinson: I did consider that but the issue that the `LocalFontIdentifier` for Windows is currently this https://docs.rs/dwrote/latest/dwrote/struct.FontDescriptor.html which doesn't really support arbitrary local files at the moment.

mrobinson: > Would it make sense to disable this variant for non-test builds?

I went back and forth on this, but enabled it for non-test builds so that it was easier to catch issues earlier in development, rather than only when running the test build. In addition it made matches a bit easier, since we could avoid wildcards (though looking over things, I didn't go back and do this yet).

mrobinson: I feel like a decent followup to this would be to see if we can figure out how to make the `LocalFontIdentifier` for Windows a path and index.

PR #33542, opened 2024-09-25
Readablestream default controller: fulfill read requests

gterzian:
Implement fulfillment of read request when enqueuing a chunk for a stream with pending requests. Part of https://github.com/servo/servo/issues/32898

---

### DIFF
--- a/components/script/dom/readablestream.rs
+++ b/components/script/dom/readablestream.rs
@@ -27,6 +27,7 @@ use crate::dom::bindings::import::module::UnionTypes::{
 };
 use crate::dom::bindings::reflector::{reflect_dom_object, DomObject, Reflector};
 use crate::dom::bindings::root::{Dom, DomRoot, MutNullableDom};
+use crate::dom::bindings::trace::RootedTraceableBox;
 use crate::dom::bindings::utils::get_dictionary_property;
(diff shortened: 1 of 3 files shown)


gterzian:
There is quite a bit of unsafe code and it is peppered all over; at a later stage we can perhaps consolidate it somewhere.

components/script/dom/readablestream.rs
```diff
@@ -414,7 +416,11 @@ impl ReadableStream {
                     .expect("Stream must have a reader when a read request is fulfilled.");
                 let request = reader.remove_read_request();
                 if !done {
-                    request.chunk_steps(chunk);
+                    let cx = GlobalScope::get_cx();
+                    rooted!(in(*cx) let mut rval = UndefinedValue());
+                    let result = RootedTraceableBox::new(Heap::default());
+                    result.set(*chunk);
+                    request.chunk_steps(result);
```
Taym95, line 423: this logic kinda duplicated with `perform_pull_steps ` but I think we can merge it like this

wusyong (approved):

components/script/dom/readablestream.rs
gterzian: Yes I propose we do this quickly and revisit when the dust has settled.

PR #33544, opened 2024-09-25
layout: Switch `.len()` comparison to `is_empty()` in `components/layout_2020/query.rs`.

DanielFrantes:
---

- [x] These changes do not require tests because dont chenge app logic

### DIFF
--- a/components/layout_2020/query.rs
+++ b/components/layout_2020/query.rs
@@ -741,7 +741,7 @@ fn rendered_text_collection_steps<'dom>(
                     items.push(InnerOrOuterTextItem::Text(String::from(" ")));
                 };
 
-                if !transformed_text.is_empty(){
+                if !transformed_text.is_empty() {
                     // Here we decide whether to keep or truncate the final white
                     // space character, if there is one.
                     if is_final_character_whitespace && !is_preformatted_element {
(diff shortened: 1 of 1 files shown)


components/layout_2020/query.rs
```diff
@@ -741,7 +741,7 @@ fn rendered_text_collection_steps<'dom>(
                     items.push(InnerOrOuterTextItem::Text(String::from(" ")));
                 };
 
-                if transformed_text.len() > 0 {
+                if !transformed_text.is_empty() == true {
```
mrobinson, line 744: You shouldn't need to compare a boolean expression against true.

```suggestion
                if !transformed_text.is_empty() {
```

mrobinson (changes requested):

components/layout_2020/query.rs
danik292: Thank I forgot about this.🙃

danik292: Should be ok now

components/layout_2020/query.rs
```diff
@@ -741,7 +741,7 @@ fn rendered_text_collection_steps<'dom>(
                     items.push(InnerOrOuterTextItem::Text(String::from(" ")));
                 };
 
-                if transformed_text.len() > 0 {
+                if !transformed_text.is_empty(){
```
mrobinson, line 744:
```suggestion
                if !transformed_text.is_empty() {
```

mrobinson (approved):

PR #33550, opened 2024-09-26
Fix table track constrainedness

Loirooriol:
Only as size that isn't `auto` and doesn't contain percentages can constrain a table track (https://drafts.csswg.org/css-tables/#constrainedness).

However, in a bunch of cases we were only checking for `auto`.

Also, we were allowing the inline-size of a cell to constrain both its column and row. Using the block-size of the row makes more sense. The spec doesn't define constrainedness for rows, though.

---

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -371,7 +371,7 @@ impl<'a> TableLayout<'a> {
         self.columns = vec![ColumnLayout::default(); self.table.size.width];
 
         let is_length = |size: &LengthPercentageOrAuto| {
-            size.non_auto().and_then(|size| size.to_length()).is_some()
+            size.non_auto().is_some_and(|size| !size.has_percentage())
         };
 
         for column_index in 0..self.table.size.width {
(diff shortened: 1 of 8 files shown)


components/layout_2020/table/layout.rs
```diff
@@ -370,52 +370,50 @@ impl<'a> TableLayout<'a> {
         self.rows = vec![RowLayout::default(); self.table.size.height];
         self.columns = vec![ColumnLayout::default(); self.table.size.width];
 
+        let is_length = |size: &LengthPercentageOrAuto| {
+            size.non_auto().and_then(|size| size.to_length()).is_some()
```
mrobinson, line 374: I'm not sure if something like the following is clearer or not:

```suggestion
            size.non_auto().is_some_and(|size| !size.has_percentage())
```

mrobinson (approved):

components/layout_2020/table/layout.rs
Loirooriol: Done, I forgot about `has_percentage()`

PR #33551, opened 2024-09-26
Readablestream default controller: use strategy size

gterzian:
Use the strategy size algorithm in the default controller. Part of https://github.com/servo/servo/issues/32898

---

### DIFF
--- a/components/script/dom/readablestream.rs
+++ b/components/script/dom/readablestream.rs
@@ -29,7 +29,7 @@ use crate::dom::bindings::reflector::{reflect_dom_object, DomObject, Reflector};
 use crate::dom::bindings::root::{Dom, DomRoot, MutNullableDom};
 use crate::dom::bindings::trace::RootedTraceableBox;
 use crate::dom::bindings::utils::get_dictionary_property;
-use crate::dom::countqueuingstrategy::extract_high_water_mark;
+use crate::dom::countqueuingstrategy::{extract_high_water_mark, extract_size_algorithm};
 use crate::dom::globalscope::GlobalScope;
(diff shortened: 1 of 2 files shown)


components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -306,24 +324,71 @@ impl ReadableStreamDefaultController {
...
+                        error
+                            .clone()
+                            .to_jsval(*cx, &*self.global(), rval.handle_mut())
+                    };
+                    
+                    // Perform ! ReadableStreamDefaultControllerError(controller, result.[[Value]]).
+                    self.error(rval.handle());
+                    
+                    // Return result.
+                    return Err(error);
```
gterzian, line 352: Looks like there is some duplication again in the error handling, but I suggest for do these clean-ups later(who knows what other stuff will emerge as we implement the rest).

Taym95: Indeed let do clean-ups later.

components/script/dom/readablestream.rs
```diff
@@ -129,6 +129,7 @@ impl ReadableStream {
                 global,
                 UnderlyingSourceType::Js(underlying_source_dict),
                 extract_high_water_mark(strategy, 1.0)?,
+                extract_size_algorithm(strategy),
```
wusyong, line 132: Could you take this chance to expand the comments on this method?
I wanted to replace those step 1, step 2 to more useful spec comments for a while.

wusyong (approved):

Issue #32853, opened 2024-07-25
Support intrinsic sizing keywords: `min-content`, `max-content`, `fit-content`, and `fit-content()`


nicoburns:
### Describe the new feature

Servo should support the intrinsic sizing keyword from `min-content`, `max-content`, `fit-content`, and `fit-content()`.

Spec: https://drafts.csswg.org/css-sizing-3/#sizing-values

Level 4 of the css-sizing spec also defines a `stretch` keyword, but that is for extrisic sizing rather than intrinsic sizing. And it is not yet widely supported (e.g. neither Firefox nor Stylo support it although the support the similar MozAvailable) so it should probably be tracked separately.

### Motivation

These keywords are used in a number of Flexbox and particularly CSS Grid WPT tests (even when it's not what the test is supposed to be testing). So if we want to know whether we're passing those tests we'll need to support these keywords. Notably some of these tests require that block layout supports these keywords even though they are ostensibly testing flexbox/grid.

They also represent key sizing concepts within the CSS layout model which we need to be able to compute anyway, so they are a good litmus test that our layout architecture is able to correctly implement that model.

...and of course real world websites use them.

yisibl:
Will these implementations be ported to Stylo? Especially the `stretch` keyword without prefix.

yisibl:
> > Will these implementations be ported to Stylo? Especially the `stretch` keyword without prefix.
> 
> Stylo already supports all of these features due to their usage in Firefox (`stretch` is feature flagged off).

Thanks.

Loirooriol:
This is done. Some details need to be discussed in the CSSWG, but adjustments can happen in other issues.

PR #33558, opened 2024-09-26
Make ComputedValuesExt expose keywords for the sizing properties

Loirooriol:
This will allow callers to start obeying `min-content`, `max-content`, `fit-content` and `stretch` in follow-up patches.
The old functionality is kept as deprecated methods that we should eventually remove.
This patch has very little impact on the existing behavior, just some very minimal implementation of the keywords for css tables.

This also overhauls fixed-layout-2.html since:
 - It had code that wasn't doing anything
 - It had wrong expecations in prose
 - The logic seemed broken in general
 - All browsers were failing one testcase

---

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -33,8 +33,8 @@ use crate::fragment_tree::{
     PositioningFragment,
 };
 use crate::geom::{
-    AuOrAuto, LengthPercentageOrAuto, LogicalRect, LogicalSides, LogicalVec2, PhysicalPoint,
-    PhysicalRect, PhysicalSides, ToLogical, ToLogicalWithContainingBlock,
+    AuOrAuto, LogicalRect, LogicalSides, LogicalVec2, PhysicalPoint, PhysicalRect, PhysicalSides,
+    Size, SizeKeyword, ToLogical, ToLogicalWithContainingBlock,
 };
 use crate::positioned::{relative_adjustement, PositioningContext, PositioningContextLength};
 use crate::sizing::ContentSizes;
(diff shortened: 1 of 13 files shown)


components/layout_2020/table/layout.rs
```diff
@@ -245,9 +245,20 @@ impl<'a> TableLayout<'a> {
...
-            !self.table.style.box_size(writing_mode).inline.is_auto();
+            match self.table.style.box_size(writing_mode).inline {
+                Size::Numeric(_) => true,
+                Size::Keyword(keyword) => match keyword {
+                    SizeKeyword::Initial | SizeKeyword::MaxContent => false,
+                    SizeKeyword::MinContent | SizeKeyword::FitContent | SizeKeyword::Stretch => {
+                        true
+                    },
+                },
+            };
```
mrobinson, line 261: Could this be a `matches!`?

Something like this:
```suggestion
            matches!(
                self.table.style.box_size(writing_mode).inline,
                Size::Numeric(_) |
                Size::Keyword(SizeKeyword::MinContent) |
                Size::Keyword(SizeKeyword::FitContent) |
                Size::Keyword(SizeKeyword::Stretch)
             );
```

mrobinson (approved):

components/layout_2020/table/layout.rs
Loirooriol: This wouldn't cover the `fit-content()` function when we add it. So to align with Gecko & Blink, I'm going with

```rs
!matches!(
    self.table.style.box_size(writing_mode).inline,
    Size::Keyword(SizeKeyword::Initial | SizeKeyword::MaxContent)
);
```

- https://searchfox.org/mozilla-central/rev/1e8cec3727d6e09f4af41bb3d202b7a4c326ed84/layout/tables/nsTableFrame.cpp#3482
- https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/style/computed_style.h;drc=572789ee61c7e857aa288a7ac7768b2b0b4f4dcc;l=1924

Loirooriol:
I have also modified the test to check `calc-size()` on browsers that support it.

mrobinson (approved):

PR #33563, opened 2024-09-27
Fix mach run bustage

### LINKED ISSUE #33559: Can't run Servo - PostBuildCommands.run() got an unexpected keyword argument 'flavor'
```
$ ./mach run -r
Error running mach:

    ['run', '-r']

The error occurred in the implementation of the invoked mach command.

This should never occur and is likely a bug in the implementation of that
command.

message.

TypeError: PostBuildCommands.run() got an unexpected keyword argument 'flavor'

  File "/home/oriol/src/servo/python/servo/command_base.py", line 637, in configuration_decorator
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oriol/src/servo/python/servo/command_base.py", line 654, in target_configuration_decorator
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Suspect: #33512 @jschwe

jschwe:
Move `--flavor` to a new "Packaging options" group, so that we can only apply the flavor option to the commands that support it.

---
- [x] These changes fix #33559

### DIFF
--- a/python/servo/build_commands.py
+++ b/python/servo/build_commands.py
@@ -57,7 +57,7 @@ class MachCommands(CommandBase):
                      help='Print very verbose output')
     @CommandArgument('params', nargs='...',
                      help="Command-line arguments to be passed through to Cargo")
-    @CommandBase.common_command_arguments(build_configuration=True, build_type=True)
+    @CommandBase.common_command_arguments(build_configuration=True, build_type=True, package_configuration=True)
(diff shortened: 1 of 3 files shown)


mukilan (approved):
It seems like it would be more natural to handle the flavour selection as part of the `allow_target_configuration`. We can land this approach to fix the bustage.

PR #33573, opened 2024-09-27
Let protocol handlers decide if they are fetchable

### LINKED ISSUE #33564: `fetch()` calls to custom protocols don't appear to work as expected
While working on my own Servo-based browser I wanted to add the ability to change prefs at runtime. My initial attempt at this was to create an HTML page that would make fetch calls to an embedder-specific route with its own handler (in my example it was `moto:pref`) whenever a user clicked a button, but it seems the fetch calls to the embedder route just never seemed to happen. I could see in JS that the button onclick was working, and I don't recall seeing any errors when making the fetch call either.

In essence, to reproduce:
- In a custom protocol, add a route that does a println before sending the response
- Create a page that calls `fetch()` to that route
- Observe that println never occurs even when `fetch()` is called

webbeef:
This adds a 'is_fetchable()' method on the ProtocolHandler trait that is then used in the fetch code. The 'data:' protocol handler is updated to return true instead of hardcoding the scheme comparison, as well as the 'urlinfo:' handler since it's just a testing one.

---

- [X] These changes fix #33564 (GitHub issue number if applicable)

### DIFF
--- a/components/net/protocols/mod.rs
+++ b/components/net/protocols/mod.rs
@@ -29,12 +29,24 @@ use file::FileProtocolHander;
 static FORBIDDEN_SCHEMES: [&str; 4] = ["http", "https", "chrome", "about"];
 
 pub trait ProtocolHandler: Send + Sync {
+    /// Triggers the load of a resource for this protocol and returns a future
+    /// that will produce a Response. Even if the protocol is not backed by a
+    /// http endpoint, it is recommended to a least provide:
+    /// - A relevant status code.
+    /// - A Content Type.
     fn load(
         &self,
         request: &mut Request,
         done_chan: &mut DoneChannel,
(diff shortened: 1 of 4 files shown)


jdm (approved):

components/net/protocols/mod.rs
```diff
@@ -35,6 +35,10 @@ pub trait ProtocolHandler: Send + Sync {
         done_chan: &mut DoneChannel,
         context: &FetchContext,
     ) -> Pin<Box<dyn Future<Output = Response> + Send>>;
+
+    fn is_fetchable(&self) -> bool {
```
wusyong, line 39: We should add the documentation to explain how this works.

jdm (approved):

PR #33575, opened 2024-09-27
Don't recompute percentage contribution for table row

Loirooriol:
We already computed it as part of the row measure.

---


mrobinson (approved):
Nice.

Issue #30822, opened 2023-12-05
Support Xterm JS


gterzian:
https://github.com/xtermjs/xterm.js/

> Xterm.js is a front-end component written in TypeScript that lets applications bring fully-featured terminals to their users in the browser. It's used by popular projects such as VS Code, Hyper and Theia.

See also https://github.com/xtermjs/xterm.js#browser-support

gterzian:
By the way @Tyriar if you could give us a ballpark idea of some of the Web API's you rely on, it would be greatly appreciated!

gterzian:
@Tyriar thanks a lot!

gterzian:
I've added unticked boxes for all, we can tick them if we either confirm an existing support, or implement it...

gterzian:
Requires running with `--pref dom.webgl2.enabled`

Couple of errors I am seeing from `./mach run --pref dom.webgl2.enabled https://xtermjs.org/`:

```WebGL addon threw an exception during load TypeError: t.ResizeObserver is not a constructor```

```ERROR script::dom::bindings::error] Error at https://xtermjs.org/js/xterm.js:1:10901 e.composedPath is not a function```

gterzian:
@Tyriar Are you using the shadow DOM?(asking because it seems you are using `event.composedPath`, which appears to often be used in conjunction with the shadow DOM).

Tyriar:
@gterzian looks like that part is just about not messing with tooltips rendered by users of the xterm.js library. For example vscode shows a hover which is not under the primary `.xterm` element.

gterzian:
I have ticked the `InterserctionObserver` box above, although we only have stubs for that since https://github.com/servo/servo/pull/33989

Gae24:
Now when enabling `dom_resize_observer_enabled` and `dom_webgl2_enabled` xtermjs.org seems running fine, only logging:
 `task queue exceeded allotted deadline by 85ms`
While  clipboard shortcuts are the usual, others could be added, but probably should depend on the platform.
On Linux with Firefox these are the working shortcuts:
copy: `Ctrl + Insert` (`Ctrl + C` has the same behaviour of the system terminal)
paste: `Ctrl + Shift + V`  `Shift + Insert`

Tyriar:
> task queue exceeded allotted deadline by 85ms

This is a warning for performance, that likely indicates performance of 2d canvas context rendering is subpar as the webgl renderer uses that to rasterize the text.

Gae24:
[...]

Issue #10072, opened 2016-03-18
Implement FrozenArray


Ms2ger:
Filing for reference; I definitely don't expect us to implement this before Gecko (<https://bugzilla.mozilla.org/show_bug.cgi?id=1236777>).

Manishearth:
Note that we currently have `to_frozen_array` for constructing `JSVal` frozen arrays. This issue is about having them work in WebIDL.

sagudev:
I might work on this given that it would be useful for webgpu: https://gpuweb.github.io/gpuweb/#gpucompilationinfo

sagudev:
part of https://github.com/servo/servo/issues/30862

PR #33576, opened 2024-09-27
Implement Clipboard Event Api

Gae24:
Some of the things to consider:
- Before, clipboard access was only possible with keyboard shortcuts, but given that clipboard actions can also be fired in other ways. I added an EmbedderEvent for clipboard.
- Maybe some abstraction between system clipboard's contents and the ones Servo handles should be added. Currently servoshell clipboard can either provide texts or image.
- Maybe "copy", "cut" and "paste" should be added to servo_atoms.
- For now the events are only handled inside HtmlInputElement and HtmlTextAreaElement, while per spec every element that is contenteditable should handle them.
- Clipboard actions algorithm returns bool values, but it's not clear the result for what is useful, so it could be removed.
---


components/script/dom/htmlinputelement.rs
```diff
@@ -966,6 +968,74 @@ impl HTMLInputElement {
 
         failed_flags
     }
+
+    // TODO mirrored in htmltextareaelement.rs
+    // https://www.w3.org/TR/clipboard-apis/#clipboard-actions
```
mrobinson, line 973: Please use rustdoc comments when prefixing methods and functions with comments.

Gae24:
I also wanted to ask if the usage of Dom types was correct, unsure what to do, I tried to follow how was already done, in particular for the `items` field in `DataTransfer` and `DataTransferItemList`

components/script/dom/datatransferitemlist.rs
```diff
@@ -0,0 +1,245 @@
...
+use crate::dom::datatransferitem::{DataTransferItem, Kind};
+use crate::dom::file::File;
+use crate::dom::window::Window;
+use crate::script_runtime::CanGc;
+
+#[dom_struct]
+pub struct DataTransferItemList {
+    reflector_: Reflector,
+    data_transfer: MutableWeakRef<DataTransfer>,
+    items: DomRefCell<Vec<DomRoot<DataTransferItem>>>,
```
jdm, line 25:
```suggestion
    items: DomRefCell<Vec<Dom<DataTransferItem>>>,
```

components/script/dom/datatransferitem.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::dom::bindings::str::DOMString;
+use crate::dom::bindings::weakref::MutableWeakRef;
+use crate::dom::datatransfer::DataTransfer;
+use crate::dom::file::File;
+use crate::dom::globalscope::GlobalScope;
+
+#[derive(Clone, JSTraceable, MallocSizeOf)]
+pub enum Kind {
+    Text(DOMString),
+    File(DomRoot<File>),
```
jdm, line 24: It's not _wrong_ to store a `DomRoot` inside of a dom_struct type, but `Dom` is more efficient. However, this requires one of two patterns for enums like this:
1. multiple enums (RootedKind and Kind) that store `DomRoot` and `Dom` respectively
2. using the enum for the actual storage and using Dom, and exposing multiple functions that accept different arguments (`new_text`, `new_file`)

I lean towards option 1 here, with one change:
```rust
pub enum Kind<'a> {
    Text(DOMString),
    File(&'a File),
}

enum KindStorage {
    Text(DOMString),
    File(Dom<File>),
}
```
Have the `new` and `new_inherited` methods take `Kind` arguments, and convert that into the appropriate `KindStorage` value when creating the actual DataTransferItem value.

components/script/dom/datatransfer.rs
```diff
@@ -0,0 +1,149 @@
...
+            mode: Cell::new(Mode::ReadWrite),
+        }
+    }
+
+    pub fn new_with_proto(
+        window: &Window,
+        proto: Option<HandleObject>,
+        can_gc: CanGc,
+    ) -> DomRoot<DataTransfer> {
+        let item_list = DataTransferItemList::new(window, proto, can_gc);
```
jdm, line 53: Passing the `proto` value here is incorrect; it is only a valid prototype for the `DataTransfer` object.

components/script/dom/datatransferitemlist.rs
```diff
@@ -0,0 +1,245 @@
...
+        DataTransferItemList {
+            reflector_: Reflector::new(),
+            data_transfer: MutableWeakRef::new(None),
+            items: DomRefCell::new(Vec::new()),
+        }
+    }
+
+    pub fn new(
+        window: &Window,
+        proto: Option<HandleObject>,
```
jdm, line 39:
```suggestion
```

components/script/dom/datatransferitemlist.rs
```diff
@@ -0,0 +1,245 @@
...
+        can_gc: CanGc,
+    ) -> DomRoot<DataTransferItemList> {
+        reflect_dom_object_with_proto(
+            Box::new(DataTransferItemList::new_inherited()),
+            window,
+            proto,
+            can_gc,
```
jdm, line 46:
```suggestion
        reflect_dom_object(
            Box::new(DataTransferItemList::new_inherited()),
            window,
            can_gc,
```

jdm (commented):
I only looked at the members of the new DOM structs to verify that the usage made sense.

components/script/dom/datatransferitem.rs
Gae24: What should I do for `Dom<File>`error: Type must be rooted

jdm: Add this attribute to the enum: https://github.com/servo/servo/blob/163e47766841725201f683495f4506f35fc9614c/components/script/dom/raredata.rs#L24

Gae24:
I've been busy lately, but will soon resume working on this

Gae24:
Sorry for the ping @gterzian, but what do you think of the current implementation?

gterzian (commented):
I think it's going in the right direction, thank you for taking this one up! 

Any problems encountered when manually testing this?

gterzian:
> Regarding webidl interfaces, only DataTransfer's types is missing requires https://github.com/servo/servo/issues/10072.

For that we usually use https://github.com/servo/servo/blob/a501572971d7ec6761aeab1f8f5dba00158bc1a6/components/script/dom/bindings/utils.rs#L120

Gae24:
> I think it's going in the right direction, thank you for taking this one up!
> 
> Any problems encountered when manually testing this?

I haven't done much testing, many wpt tests rely on execCommand.
At the moment I just compared behaviour on paste events with firefox using this [clipboard test](https://madebyevan.com/clipboard-test/).

Gae24:
@jdm since there are many additions and commit history is a bit messy, I was thinking, while working on clipboard events here, of opening another pr which includes DataTransfer and the others api to start landing it

jdm:
@Gae24 I think that's a good idea!

gterzian:
By the way this is one of the API we need for https://github.com/servo/servo/issues/30822

jdm:
@Gae24 Is this PR waiting to be reviewed, or is there still more work needed?

Gae24:
no, I'm still working on it.

Gae24:
@jdm sorry for taking this long.
Since 1Password/arboard#163 hasn't landed yet and need to investigate why arboard doesn't provide image data when trying to paste it, for now only text is supported.

Gae24:
The unit test fail is caused by how clipboard events are now received, maybe for now I should revert it?

jdm (commented):
Partial review during my kid's nap! I'll pick this up again from htmltextareaekenent.rs.

components/script/dom/document.rs
```diff
@@ -1445,6 +1449,114 @@ impl Document {
...
+            DataTransfer::new(&self.window, Rc::new(RefCell::new(Some(drag_data_store))));
+
+        let event = ClipboardEvent::new(
+            &self.window,
+            None,
+            DOMString::from(event.as_str()),
+            EventBubbles::Bubbles,
+            EventCancelable::Cancelable,
+            None,
+            CanGc::note(),
```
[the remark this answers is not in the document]
Gae24: I added `CanGc::note()` inside `handle_clipboard_action`, but maybe is it better to pass it from `handle_event`?

components/script/dom/document.rs
```diff
@@ -1445,6 +1449,114 @@ impl Document {
...
+                // Step 7.1.2 If trusted or the implementation gives script-generated events access to the clipboard
+                if trusted {
+                    // Step 7.1.2.1 For each clipboard-part on the OS clipboard:
+
+                    // Step 7.1.2.1.1 If clipboard-part contains plain text, then
+                    let plain_string = PlainString::new(
+                        DOMString::from_string(contents.to_string()),
+                        DOMString::from("text/plain"),
+                    );
+                    let _ = drag_data_store.add(Kind::Text(plain_string));
```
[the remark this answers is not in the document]
Gae24: add returns an error to the dom if there is already some text with the same mime type. Here, I don't think we need to handle this error.

components/script/dom/document.rs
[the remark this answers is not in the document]
jdm: Yep, we want a chain as far up the stack as possible.

components/script/dom/htmltextareaelement.rs
```diff
@@ -457,6 +458,84 @@ impl HTMLTextAreaElement {
     fn selection(&self) -> TextControlSelection<Self> {
         TextControlSelection::new(self, &self.textinput)
     }
+
+    /// TODO mirrored in htmlinputelement.rs
+    /// <https://www.w3.org/TR/clipboard-apis/#clipboard-actions>
+    fn handle_clipboard_action(&self, event: &ClipboardEvent, can_gc: CanGc) -> bool {
```
jdm, line 464: Can we move the duplicated code to a helper in a new clipboard module under components/script/?

components/script/dom/htmlinputelement.rs
```diff
@@ -967,6 +968,84 @@ impl HTMLInputElement {
...
+                    // Step 3.2
+                    self.owner_document()
+                        .fire_clipboard_event(ClipboardEventType::Change, can_gc);
+                },
+                "cut" => {
+                    // Step 3.1
+                    let result = self.textinput.borrow_mut().cut_contents();
+                    // Step 3.1.3
+                    self.owner_document()
+                        .fire_clipboard_event(ClipboardEventType::Change, CanGc::note());
```
jdm, line 1002: There are more CanGc::note calls that can be replaced with the new argument.

jdm (changes requested):
This looks like about what I expected, but I'd like to see the changes to the implementation of https://www.w3.org/TR/clipboard-apis/#clipboard-actions and hopefully reduce the duplicated code between the text area and input elements.

components/script/dom/htmlinputelement.rs
```diff
@@ -967,6 +968,84 @@ impl HTMLInputElement {
...
+
+        // Step 1
+        if script_triggered && !script_may_access_clipboard {
+            return false;
+        }
+        // Step 2
+        let e = event.upcast::<Event>();
+
+        // Step 3
+        if !e.DefaultPrevented() {
```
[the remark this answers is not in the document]
Gae24: You are right, when I read [copy event](https://www.w3.org/TR/clipboard-apis/#clipboard-event-copy) I thought the event should be fired first.

components/script/dom/htmltextareaelement.rs
Gae24: would be fine moving this code inside `textinput.rs`?

jdm: That's probably fine.

components/script/dom/htmlinputelement.rs
```diff
@@ -967,6 +968,84 @@ impl HTMLInputElement {
...
+                    return result;
+                },
+                "paste" => {
+                    // Step 3.1 If there is a selection or cursor in an editable context where pasting is enabled, then
+                    if let Some(data) = event.get_clipboard_data() {
+                        let drag_data_store = data.data_store().expect("This shouldn't fail");
+                        self.textinput.borrow_mut().paste_contents(&drag_data_store);
+                    }
+                    // Step 3.1.2 Queue tasks to fire any events that should fire due to the modification.
+                    // Step 3.2 Else return false.
```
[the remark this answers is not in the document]
Gae24: I'm not sure, but I think that in cut and paste event we should check if target is contenteditable.

components/script/dom/htmlinputelement.rs
```diff
@@ -2684,10 +2606,14 @@ impl VirtualMethods for HTMLInputElement {
                 }
                 event.mark_as_handled();
             }
-        }
-
-        if let Some(clipboard_event) = event.downcast::<ClipboardEvent>() {
-            self.handle_clipboard_action(clipboard_event, CanGc::note());
+        } else if let Some(clipboard_event) = event.downcast::<ClipboardEvent>() {
+            if !event.DefaultPrevented() {
+                self.textinput.borrow_mut().handle_text_clipboard_action(
```
jdm, line 2611: We can't have a borrow_mut() active while invoking methods that can trigger a GC—this risks a double borrow panic if the GC actually happens while the borrow is oustanding. `handle_text_clipboard_action` will need to be `&self` instead, or else `handle_text_clipboard_action` can be a helper function that takes a `&RefCell<TextInput>` argument so it can scope the borrow so it is not active while calling methods with CanGc arguments.

components/script/dom/htmltextareaelement.rs
```diff
@@ -757,10 +679,14 @@ impl VirtualMethods for HTMLTextAreaElement {
                 }
                 event.mark_as_handled();
             }
-        }
-
-        if let Some(clipboard_event) = event.downcast::<ClipboardEvent>() {
-            self.handle_clipboard_action(clipboard_event, CanGc::note());
+        } else if let Some(clipboard_event) = event.downcast::<ClipboardEvent>() {
+            if !event.DefaultPrevented() {
+                self.textinput.borrow_mut().handle_text_clipboard_action(
```
jdm, line 684: Same comment here about the borrow/CanGc combo.

components/script/dom/document.rs
```diff
@@ -1448,8 +1448,73 @@ impl Document {
...
+        let event = ClipboardEvent::new(
+            &self.window,
+            None,
+            DOMString::from(action.as_str()),
+            EventBubbles::Bubbles,
+            EventCancelable::Cancelable,
+            None,
+            can_gc,
+        );
+        self.fire_clipboard_event(&event, action, can_gc);
```
jdm, line 1476: Can we add a comment that step 3 is handled by the VirtualMethods::handle_event of relevant elements that contain editable contexts?

jdm (commented):
This is getting close!

Gae24:
I tried to run this [test case](https://bug1541572.bmoattachments.org/attachment.cgi?id=9055632), but logs were empy do I need to run servo with some features or is something not yet supported?

components/script/textinput.rs
```diff
@@ -1183,3 +1128,66 @@ impl<T: ClipboardProvider> TextInput<T> {
...
+    can_gc: CanGc,
+) -> bool {
+    let e = event.upcast::<Event>();
+
+    // Step 3
+    match e.Type().str() {
+        "copy" => {
+            // Step 3.1 Copy the selected contents, if any, to the clipboard
+            if let Some(text) = text_input.borrow().get_selection_text() {
+                text_input
```
jdm, line 1145: This looks like a mutable borrow while an immutable borrow is active.

components/script/textinput.rs
```diff
@@ -1183,3 +1128,66 @@ impl<T: ClipboardProvider> TextInput<T> {
...
+            // Step 3.2 Fire a clipboard event named clipboardchange
+            owning_node
+                .owner_document()
+                .fire_clipboardchange_event(can_gc);
+        },
+        "cut" => {
+            // Step 3.1 If there is a selection in an editable context where cutting is enabled, then
+            if let Some(text) = text_input.borrow().get_selection_text() {
+                // Step 3.1.1 Copy the selected contents, if any, to the clipboard
+                text_input
```
jdm, line 1160: Same here.

jdm:
> I tried to run this [test case](https://bug1541572.bmoattachments.org/attachment.cgi?id=9055632), but logs were empty, do I need to run servo with some features or it's something not yet supported?

Interesting. It looks like Servo should support all the DOM APIs in use in that test case, so my suspicion is that either the copy action is not making it correctly to the script thread?

jdm:
Unit test is failing:
```
---- textinput::test_clipboard_paste stdout ----
thread 'textinput::test_clipboard_paste' panicked at tests/unit/script/textinput.rs:612:5:
  left: DOMString("vdefg", PhantomData<*const ()>)
 right: "abcdefg"
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   4: script_tests::textinput::test_clipboard_paste
```

Gae24:
> > I tried to run this [test case](https://bug1541572.bmoattachments.org/attachment.cgi?id=9055632), but logs were empty, do I need to run servo with some features or it's something not yet supported?
> 
> Interesting. It looks like Servo should support all the DOM APIs in use in that test case, so my suspicion is that either the copy action is not making it correctly to the script thread?

Even with paste event, which works, servo print the same: 
```
Current selection is ({})
Current activeElement is ({})
Got paste event, target is ({})
```

components/script/textinput.rs
```diff
@@ -1128,4 +1137,79 @@ impl<T: ClipboardProvider> TextInput<T> {
...
+}
+
+pub(crate) fn handle_text_clipboard_action(
+    owning_node: &impl NodeTraits,
+    textinput: &DomRefCell<TextInput<ScriptToConstellationChan>>,
+    event: &ClipboardEvent,
+    can_gc: CanGc,
+) -> bool {
+    let e = event.upcast::<Event>();
+
```
jdm, line 1157: The spec doesn't have an explicit step for this, but we need to check the IsTrusted value of the event and prevent any changes to the system clipboard if it's false.

components/script/dom/document.rs
```diff
@@ -1443,6 +1447,189 @@ impl Document {
...
+
+        // Step 3 If a script doesn't call preventDefault()
+        // the event will be handled inside target's VirtualMethods::handle_event
+
```
jdm, line 1479: The spec doesn't have an explicit step here, but we need to prevent any modification of the system clipboard if IsTrusted is false.

components/servo/lib.rs
```diff
@@ -900,6 +900,13 @@ where
             EmbedderEvent::Vsync => {
                 self.compositor.on_vsync();
             },
+            EmbedderEvent::ClipboardAction(clipboard_event) => {
+                //TODO find the best way
```
jdm, line 904: What does this mean?

resources/prefs.json
```diff
@@ -7,6 +7,7 @@
   "dom.bluetooth.testing.enabled": false,
   "dom.canvas_capture.enabled": false,
   "dom.canvas_text.enabled": true,
+  "dom.clipboard_event.enabled": false,
```
jdm, line 10: What is still missing before we will want to enable this by default?

jdm (commented):
This is really close!

jdm (approved):
Woo! ~Let's update the interfaces test file in tests/wpt/mozilla/tests and update the manifest, and we should be good to merge!~ I see you already did this.

jdm:
```
error[E0004]: non-exhaustive patterns: `EmbedderMsg::ClearClipboardContents` not covered
   --> ports/servoshell/egl/servo_glue.rs:447:19
    |
447 |             match event {
    |                   ^^^^^ pattern `EmbedderMsg::ClearClipboardContents` not covered
    |
note: `EmbedderMsg` defined here
   --> /home/runner/work/servo/servo/components/shared/embedder/lib.rs:166:1
    |
166 | pub enum EmbedderMsg {
    | ^^^^^^^^^^^^^^^^^^^^
...
196 |     ClearClipboardContents,
    |     ---------------------- not covered
    = note: the matched value is of type `EmbedderMsg`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
647 ~                 EmbedderMsg::StopGamepadHapticEffect(..) => {},
648 ~                 EmbedderMsg::ClearClipboardContents => todo!(),
    |

For more information about this error, try `rustc --explain E0004`.
error: could not compile `servoshell` (lib) due to 1 previous error
```

Gae24:
> ```
> error[E0004]: non-exhaustive patterns: `EmbedderMsg::ClearClipboardContents` not covered
>    --> ports/servoshell/egl/servo_glue.rs:447:19
>     |
> 447 |             match event {
>     |                   ^^^^^ pattern `EmbedderMsg::ClearClipboardContents` not covered
>     |
> note: `EmbedderMsg` defined here
>    --> /home/runner/work/servo/servo/components/shared/embedder/lib.rs:166:1
>     |
> 166 | pub enum EmbedderMsg {
>     | ^^^^^^^^^^^^^^^^^^^^
> ...
> 196 |     ClearClipboardContents,
>     |     ---------------------- not covered
>     = note: the matched value is of type `EmbedderMsg`
> help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
>     |
> 647 ~                 EmbedderMsg::StopGamepadHapticEffect(..) => {},
> 648 ~                 EmbedderMsg::ClearClipboardContents => todo!(),
>     |
> 
> For more information about this error, try `rustc --explain E0004`.
> error: could not compile `servoshell` (lib) due to 1 previous error
> ```

do we ignore it for now?

PR #33577, opened 2024-09-27
Simplify the computation of CAPMIN

Loirooriol:
CAPMIN is the largest min-content contribution of the table captions.

In Servo, the standard way to compute min/max-content contributions is `outer_inline_content_sizes()`, so just use that instead of reinventing the wheel.

This also fixes cyclic percentages to resolve consistently with normal block boxes.

---


mrobinson (approved):
Very nice cleanup!

PR #33579, opened 2024-09-28
Refactor sizing logic for table measures

Loirooriol:
After #33577, `get_outer_sizes_from_style()` is only used for column and cell measures. For these it's expected to ignore sizing keywords like `min-content`, so rename the function to `get_outer_sizes_for_measurement()`.

Additionally, both callers need the percentage contribution, so include it in the returned tuple.

---

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -281,10 +281,12 @@ impl<'a> TableLayout<'a> {
                     block: padding.block_sum() + border.block_sum(),
                 };
 
-                let (size, min_size, max_size, inline_size_is_auto) =
-                    get_outer_sizes_from_style(&cell.style, writing_mode, &padding_border_sums);
-                let percentage_contribution =
-                    get_size_percentage_contribution_from_style(&cell.style, writing_mode);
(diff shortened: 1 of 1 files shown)


components/layout_2020/table/layout.rs
```diff
@@ -2823,16 +2815,22 @@ fn get_size_percentage_contribution_from_style(
...
     LogicalVec2 {
-        inline: get_contribution_for_axis(size.inline, max_size.inline),
-        block: get_contribution_for_axis(size.block, max_size.block),
+        inline: get_contribution_for_axis(&size.inline, &max_size.inline),
+        block: get_contribution_for_axis(&size.block, &max_size.block),
     }
 }
 
-fn get_outer_sizes_from_style(
+fn get_outer_sizes_for_measure(
```
mrobinson, line 2823: This is getting the outer sizes for the process of measuring the table, correct? Maybe `get_outer_sizes_for_measurement` might be a bit clearer here. The former name is a bit ambiguous, because it could mean, "get the outer sizes taking a 'measure' as an argument."

```suggestion
fn get_outer_sizes_for_measure(
```

mrobinson (approved):

PR #33583, opened 2024-09-30
fix(pixels): make the unmultiply_inplace more performance

### LINKED ISSUE #33552: Move `unmultiply_inplace` to pixels crate and optimize it
https://github.com/servo/servo/blob/7fdaccde5501283708503e9a64efa578ac9ae7f7/components/canvas/webgl_thread.rs#L2841-L2848
while more performant (not tested just assumed because no fp operations) is in wrench:
https://github.com/servo/webrender/blob/bd7704a7091c51593cef9a3fe2ba86eb7e26359e/wrench/src/premultiply.rs#L21

we should move function to components/pixels, write benchmark and see if wrench version performs better.

nack098:
change the chunks to iterator and change the f32 to u16

---

- [X] These changes fix #33552  (GitHub issue number if applicable)

- [ ] These changes do not require tests because the core functionality does not change.

### DIFF
--- a/components/pixels/lib.rs
+++ b/components/pixels/lib.rs
@@ -198,11 +198,16 @@ pub fn detect_image_format(buffer: &[u8]) -> Result<ImageFormat, &str> {
 }
 
 pub fn unmultiply_inplace(pixels: &mut [u8]) {
-    for rgba in pixels.chunks_mut(4) {
-        let a = (rgba[3] as f32) / 255.0;
-        rgba[0] = (rgba[0] as f32 / a) as u8;
-        rgba[1] = (rgba[1] as f32 / a) as u8;
-        rgba[2] = (rgba[2] as f32 / a) as u8;
+    let mut iter = pixels.iter_mut();
+    while let Some(r) = iter.next() {
+        let g = iter.next().unwrap();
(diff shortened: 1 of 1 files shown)


components/pixels/lib.rs
```diff
@@ -198,11 +198,16 @@ pub fn detect_image_format(buffer: &[u8]) -> Result<ImageFormat, &str> {
...
-    for rgba in pixels.chunks_mut(4) {
-        let a = (rgba[3] as f32) / 255.0;
-        rgba[0] = (rgba[0] as f32 / a) as u8;
-        rgba[1] = (rgba[1] as f32 / a) as u8;
-        rgba[2] = (rgba[2] as f32 / a) as u8;
+    let mut iter = pixels.iter_mut();
+    while let Some(r) = iter.next() {
+        let g = iter.next().unwrap();
+        let b = iter.next().unwrap();
+        let a = iter.next().unwrap();
```
sagudev, line 205: Actually using `chunks_mut` is better than doing this manually with iterator.

sagudev:
Closing this in favor of #33582 as noted in https://github.com/servo/servo/issues/33552#issuecomment-2382168550.

PR #33586, opened 2024-09-30
Readablestream: impl default controller should pull, start algo

gterzian:
Implement the "should pull" algorithm for the default controller, and calls into the start algorithm of the underlying source. Part of https://github.com/servo/servo/issues/32898

---

### DIFF
--- a/components/script/dom/readablestreamdefaultcontroller.rs
+++ b/components/script/dom/readablestreamdefaultcontroller.rs
@@ -62,6 +62,38 @@ impl Callback for PullAlgorithmRejectionHandler {
     }
 }
 
+/// The fulfillment handler for
+/// <https://streams.spec.whatwg.org/#dom-underlyingsource-start>
+#[derive(Clone, JSTraceable, MallocSizeOf)]
+#[allow(crown::unrooted_must_root)]
+struct StartAlgorithmFulfillmentHandler {
+    // TODO: check the validity of using Dom here.
+    controller: Dom<ReadableStreamDefaultController>,
+}
+
+impl Callback for StartAlgorithmFulfillmentHandler {
+    /// Handle fufillment of pull algo promise.
+    fn callback(&self, _cx: JSContext, _v: HandleValue, _realm: InRealm) {
+        todo!();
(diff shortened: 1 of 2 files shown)


components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -200,15 +238,38 @@ impl ReadableStreamDefaultController {
...
+        if let Some(underlying_source) = rooted_default_controller.underlying_source.get() {
+            let promise = underlying_source.call_start_algorithm(
+                Controller::ReadableStreamDefaultController(rooted_default_controller.clone()),
+            );
+            let fulfillment_handler = Box::new(StartAlgorithmFulfillmentHandler {
+                controller: Dom::from_ref(&*rooted_default_controller),
+            });
+            let rejection_handler = Box::new(StartAlgorithmRejectionHandler {
+                controller: Dom::from_ref(&*rooted_default_controller),
+            });
```
wusyong, line 260: There are some stuff need to do upon fulfillment and rejection.Do they go to the `todo!()` in Callback trait implementations?

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -111,6 +112,51 @@ impl UnderlyingSourceContainer {
         None
     }
 
+    /// <https://streams.spec.whatwg.org/#dom-underlyingsource-start>
+    /// The algorithm can return any value, including a promise,
+    /// we always transform the result into a promise for convenience.
```
wusyong, line 117: We could say it's because the spec specifically asks for it in [step 10](https://streams.spec.whatwg.org/#set-up-readable-stream-default-controller).

components/script/dom/readablestreamdefaultcontroller.rs
gterzian: yes.

components/script/dom/underlyingsourcecontainer.rs
gterzian: Right.

PR #33587, opened 2024-09-30
CI: do WPT reports also as GitHub job summary

sagudev:
this is less user friendly as GitHub checks but has bigger content limit (1MB vs. 64k characters) that webgpu runs hit occasionally.

Example: https://github.com/sagudev/servo/actions/runs/11105859427/attempts/1#summary-30855004760 (`#Aggregated linux-wpt-layout-2020 report` will not be printed anymore).

---

- [x] These changes do not require tests because they are for CI

### DIFF
--- a/etc/ci/report_aggregated_expected_results.py
+++ b/etc/ci/report_aggregated_expected_results.py
@@ -184,7 +184,16 @@ def get_pr_number() -> Optional[str]:
     return None
 
 
-def create_check_run(body: str, tag: str = ""):
+def create_github_reports(body: str, tag: str = ""):
+    # GitHub will set a special environment variable which points to a file where
+    # a job summary can be appended. This is produced in addition to the GitHub
+    # check run below, as it has a larger length limit.
+    if "GITHUB_STEP_SUMMARY" in os.environ:
+        with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
+            print(body, file=f)
+
+    # Create a GitHub check run. This is more user friendly than a job summary,
+    # but has a very limited size -- thus it may fail.
     # This process is based on the documentation here:
     # https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-runs
     results = json.loads(os.environ.get("RESULTS", "{}"))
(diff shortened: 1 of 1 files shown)


etc/ci/report_aggregated_expected_results.py
```diff
@@ -184,7 +184,13 @@ def get_pr_number() -> Optional[str]:
     return None
 
 
-def create_check_run(body: str, tag: str = ""):
+def create_report_on_github(body: str, tag: str = ""):
+    # create github job summery
+    if "GITHUB_STEP_SUMMARY" in os.environ:
+        with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
+            print(body, file=f)
```
mrobinson, line 191:
```suggestion
    # GitHub will set a special environment variable which points to a file where 
    # a job summary can be appended. This is produced in addition to the GitHub 
    # check run below, as it has a larger length limit.
    if "GITHUB_STEP_SUMMARY" in os.environ:
        with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as f:
            print(body, file=f)
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -184,7 +184,13 @@ def get_pr_number() -> Optional[str]:
...
+
+    # create github check run (more user friendly but very limited report size)
```
mrobinson, line 193:
```suggestion
    # Create a GitHub check run. This is more user friendly than a job summary,
    # but has a very limited size -- thus it may fail.
```

etc/ci/report_aggregated_expected_results.py
```diff
@@ -184,7 +184,13 @@ def get_pr_number() -> Optional[str]:
     return None
 
 
-def create_check_run(body: str, tag: str = ""):
+def create_report_on_github(body: str, tag: str = ""):
```
mrobinson, line 187:
```suggestion
def create_github_reports(body: str, tag: str = ""):
```

mrobinson (approved):
Looks good to me with a few minor wording tweaks and expansions.

mrobinson (approved):

mrobinson:
Apologies, I have hit the "Update branch" button by accident.

sagudev:
> bigger content limit (1MB vs. 64k characters) that webgpu runs hit occasionally

here is one such big run: https://github.com/sagudev/servo/actions/runs/11164323520/attempts/1#summary-31036874060

PR #33588, opened 2024-09-30
ohos: Allow passing arguments to servoshell

jschwe:
Allows passing options passed via `wants` to the Ability through to servoshell.
This allows easier debugging, either by calling servoshell from a test app,
or from the commandline, e.g.
```
hdc shell aa start -a EntryAbility -b org.servo.servoshell -U "https://www.wikipedia.org" \
--pb dom.webgpu.enabled true \
--ps dom.webgpu.wgpu_backend "gl" \
--pi layout.threads 4
```

Note: While the OH `wants` API differentiates between boolean, string and integer values, we convert
everything back to strings, so we can reuse the same parsing code as the desktop servoshell.

---

### DIFF
--- a/support/openharmony/entry/src/main/ets/entryability/EntryAbility.ets
+++ b/support/openharmony/entry/src/main/ets/entryability/EntryAbility.ets
@@ -5,8 +5,29 @@ import Want from '@ohos.app.ability.Want';
 import window from '@ohos.window';
 
 export default class EntryAbility extends UIAbility {
+  init_params: LocalStorage = new LocalStorage();
   onCreate(want: Want, launchParam: AbilityConstant.LaunchParam) {
-    hilog.info(0x0000, 'testTag', '%{public}s', 'Ability onCreate');
+    let uri: string = want.uri || "https://servo.org"
+    let params: string[] = []
+    if (typeof want.parameters !== "undefined") {
+      Object.entries(want.parameters).forEach((entry) => {
+        let key = entry[0]
+        // Skip some default parameters, since servo is not interested in those
+        if (key.startsWith("ohos.")
+          || key.startsWith("component.")
+          || key === "isCallBySCB"
+          || key === "moduleName"
+        ) {
+          return
+        }
+        let value = entry[1]
+        params.push("--" + key + "=" + value.toString())
+      })
+    }
+    let servoshell_params = params.join("\u{001f}")
+    hilog.debug(0x0000, 'Servo EntryAbility', 'Servoshell parameters: %{public}s', servoshell_params);
+    this.init_params.setOrCreate('InitialURI', uri)
+    this.init_params.setOrCreate('CommandlineArgs', servoshell_params)
   }
 
   onDestroy() {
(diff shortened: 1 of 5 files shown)


support/openharmony/entry/src/main/ets/entryability/EntryAbility.ets
```diff
@@ -5,8 +5,30 @@ import Want from '@ohos.app.ability.Want';
...
+        if (key.startsWith("ohos.")
+          || key.startsWith("component.")
+          || key === "isCallBySCB"
+          || key === "moduleName"
+        ) {
+          return
+        }
+        let value = entry[1]
+        params.push("--" + key + "=" + value.toString())
+        hilog.info(0x0000, 'Servo Entryability', 'Parameters: %s=%s', key, JSON.stringify(value))
```
mukilan, line 25: Is the `{public}` modifier not needed here in `%s=%s`?

support/openharmony/entry/src/main/ets/entryability/EntryAbility.ets
```diff
@@ -17,7 +39,7 @@ export default class EntryAbility extends UIAbility {
     // Main window is created, set main page for this ability
     hilog.info(0x0000, 'testTag', '%{public}s', 'Ability onWindowStageCreate');
 
-    windowStage.loadContent('pages/Index', (err, data) => {
+    windowStage.loadContent('pages/Index',this.init_params, (err, data) => {
```
mukilan, line 42:
```suggestion
    windowStage.loadContent('pages/Index', this.init_params, (err, data) => {
```

support/openharmony/entry/src/main/ets/pages/Index.ets
```diff
@@ -48,8 +52,11 @@ struct Index {
     type: XComponentType.SURFACE,
     libraryname: 'servoshell',
   }
-  @State urlToLoad: string =  'https://servo.org'
+
   private context = getContext(this) as common.UIAbilityContext;
+  @LocalStorageProp('InitialURI') InitialURI: string = "unused"
+  @LocalStorageProp('CommandlineArgs') CommandlineArgs: string = ""
```
mukilan, line 58: Is there documentation on the `@LocalStorageProp` decorator? Why do we need this declaration and how does it relate to the code in EntryAbility's `onWindowStageCreate` method where we pass `this.init_params` 

```
windowStage.loadContent('pages/Index',this.init_params,
```

jschwe: We need to use one of the statemanagement functions to share data between the ability and the page.
Localstorage seems like the appropriate type, since we only expect to have on Ability, and don't need any cross-ability state sharing. 
We then also have the choice between `@LocalStorageProp` and  `@LocalStorageLink`, where the former is one-way synced from ability to page, and the latter provides 2-way synchronization.
We currently only need the localstorage to provide initial values to servo, so `@LocalStorageProp` is sufficient.

If we ever need / want to save state or synchronize settings across devices, then the whole settings approach would need to be revisited.

Documentation links: 
[Application state management overview](https://docs.openharmony.cn/pages/v5.0/en/application-dev/quick-start/arkts-application-state-management-overview.md), 
[Localstorage documentation](https://docs.openharmony.cn/pages/v5.0/en/application-dev/quick-start/arkts-localstorage.md), 
[App storage documentation](https://docs.openharmony.cn/pages/v5.0/en/application-dev/quick-start/arkts-appstorage.md)

support/openharmony/entry/src/main/ets/entryability/EntryAbility.ets
jschwe: Yeah, it should be added. But I removed the log statement instead, since it is overly verbose.

mukilan (approved):

PR #33597, opened 2024-10-01
Compositor: add document id to NewWebRenderFrame variant

wusyong:
This PR makes compositor proxy send Document ID as well when it's called by `new_frame_ready`.
This won't affect anything at the moment, but can help us easier to improve [`CompositingReason`](https://github.com/servo/servo/blob/main/components/compositing/compositor.rs#L249).
I also opened a [upstream ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1921922) in hope for pushing multi-document rendering.
I hope we can know which frame belongs to which document precisely, so we have better understanding on why we need to composite when render method is called.

---

### DIFF
--- a/components/shared/compositing/lib.rs
+++ b/components/shared/compositing/lib.rs
@@ -22,6 +22,7 @@ use script_traits::{
 };
 use style_traits::CSSPixel;
 use webrender_api::units::{DeviceIntPoint, DeviceIntSize, DeviceRect};
+use webrender_api::DocumentId;
 use webrender_traits::{
     CanvasToCompositorMsg, FontToCompositorMsg, NetToCompositorMsg, ScriptToCompositorMsg,
 };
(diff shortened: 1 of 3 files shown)


mrobinson (approved):

components/shared/compositing/lib.rs
```diff
@@ -93,8 +94,9 @@ pub enum CompositorMsg {
     /// Set whether to use less resources by stopping animations.
     SetThrottled(PipelineId, bool),
     /// WebRender has produced a new frame. This message informs the compositor that
-    /// the frame is ready, so that it may trigger a recomposite.
-    NewWebRenderFrameReady(bool /* composite_needed */),
+    /// the frame is ready. It contains a bool to indicate if it needs to composite and the document
+    /// ID related to this frame.
```
mrobinson, line 98:
```suggestion
    /// the frame is ready. It contains a bool to indicate if it needs to composite and the 
    /// `DocumentId` of the new frame.
```

mrobinson (approved):
This is a very minor thing, but can we keep the order of the arguments to the enum variant the same as the order of the arguments to the `new_frame_ready()` method?

mrobinson (approved):

PR #33611, opened 2024-10-01
net: Refactor `Decoder`

### LINKED ISSUE #33339: Incremental parser updates are very small
#33286 revealed an interesting result, triggered by surprisingly small parser updates to text nodes. This is worth investigating further, because I imagine it could be contributing to our slower than expected page loads.

crbrz:
Update `Decoder` with an implementation adapted from `reqwest` 0.11.27 that's fully async. With these changes `Decoder` produces chunks of `DECODER_BUFFER_SIZE` (when enough decompressed data is available) reducing the number of incremental parser updates when processing responses compressed with brotli.

There is a new pref `network.tls.ignore_unexpected_eof` that's enabled when running WPT suite. WPT server doesn't send a close_notify before closing a TLS connection and `rustls` treats this as an error: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof. The error can be ignored when we know that all data was received but WPT server doesn't always send a `Content-Length` header.

For deflate encoded content `ZlibDecoder` is used instead of `DeflateDecoder`: https://datatracker.ietf.org/doc/html/rfc7230#section-4.2.2

---

- [X] These changes fix #33339

Try [wpt-2020 run](https://github.com/crbrz/servo/actions/runs/11133093045/job/30939969832)

### DIFF
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -577,7 +577,14 @@ mod gen {
                 },
                 mime: {
                     sniff: bool,
-                }
+                },
+                tls: {
+                    /// Ignore `std::io::Error` with `ErrorKind::UnexpectedEof` received if the connection
+                    /// is closed without a close_notify.
+                    ///
+                    /// Used for tests because WPT server doesn't propely close the TLS connection.
+                    ignore_unexpected_eof: bool,
+                },
             },
             session_history: {
                 #[serde(rename = "session-history.max-length")]
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -583,6 +583,7 @@ mod gen {
                     /// is closed without a close_notify.
                     ///
                     /// Used for tests because WPT server doesn't properly close the TLS connection.
+                    // TODO: remove this when WPT server is updated to use a proper TLS implementation.
                     ignore_unexpected_eof: bool,
                 },
             },
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -579,10 +579,10 @@ mod gen {
                     sniff: bool,
                 },
                 tls: {
-                    /// Ignore `std::io::Error` with `ErrorKind::UnexpectedEof` received if the connection
+                    /// Ignore `std::io::Error` with `ErrorKind::UnexpectedEof` received when a TLS connection
                     /// is closed without a close_notify.
                     ///
-                    /// Used for tests because WPT server doesn't propely close the TLS connection.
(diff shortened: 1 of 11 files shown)


components/config/prefs.rs
```diff
@@ -577,7 +577,14 @@ mod gen {
...
                 mime: {
                     sniff: bool,
-                }
+                },
+                tls: {
+                    /// Ignore `std::io::Error` with `ErrorKind::UnexpectedEof` received if the connection
+                    /// is closed without a close_notify.
+                    ///
+                    /// Used for tests because WPT server doesn't propely close the TLS connection.
+                    ignore_unexpected_eof: bool,
```
jdm, line 586: My preference would be to fix the WPT server instead of adding more test-only behaviour differences, but I think it makes sense to handle that in a separate followup issue.

components/net/decoder.rs
```diff
@@ -161,365 +132,164 @@ impl Decoder {
...
-}
-
-impl Stream for Gzip {
-    type Item = Result<Bytes, Error>;
-
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let mut buf = self.buf.clone();
-        if let Ok(mut reader) = self.reader.lock() {
-            reader.waker = Some(cx.waker().clone());
+        let _body = std::mem::replace(&mut self.body, BodyStream::empty().peekable());
```
jdm, line 205: Since it's actually used, let's remove the _ prefix.

components/net/decoder.rs
```diff
@@ -161,365 +132,164 @@ impl Decoder {
...
-/// as a `Bytes`.
-struct Brotli {
-    inner: Box<Decompressor<Peeked<ReadableChunks<Body>>>>,
-    buf: BytesMut,
-    reader: Arc<Mutex<ReadableChunks<Body>>>,
+struct BodyStream {
+    body: Body,
+    is_secure_scheme: bool,
+    content_length: Option<ContentLength>,
+    total_read: u64,
```
jdm, line 231: Should this be usize to avoid casts?

jdm (approved):
This is really great work! Thank you for finding and fixing this issues, as well as the rustls detective work.

components/config/prefs.rs
crbrz: Yeah, fixing WPT server is preferable. I added a TODO note to remove it when that happens.

PR #33613, opened 2024-10-02
webgpu: Introduce PresentationId to ensure updates with newer presentation

### LINKED ISSUE #33602: webgpu: Use subissions indexes to make async presentation work proprely
https://github.com/servo/servo/pull/33521#discussion_r1775913701

sagudev:
We index each (re)configuration and image with sequential number and so we only replace with images with newer data (bigger number).

---

- [x] These changes fix #33602

### DIFF
--- a/components/webgpu/swapchain.rs
+++ b/components/webgpu/swapchain.rs
@@ -41,6 +41,11 @@ impl MallocSizeOf for WebGPUContextId {
 
 pub type WGPUImageMap = Arc<Mutex<HashMap<WebGPUContextId, ContextData>>>;
 
+/// Presentation id encodes current configuration and current image
+/// so that async presentation does not update context with older data
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+struct PresentationId(u64);
+
 struct GPUPresentationBuffer {
     global: Arc<Global>,
     buffer_id: id::BufferId,
@@ -189,6 +194,12 @@ pub struct ContextData {
     buffer_ids: ArrayVec<(id::BufferId, PresentationBufferState), PRESENTATION_BUFFER_COUNT>,
     /// If there is no associated swapchain the context is dummy (transparent black)
     swap_chain: Option<SwapChain>,
+    /// Next presentation id to be returned
+    next_presentation_id: PresentationId,
+    /// Current id that is presented/configured
+    ///
+    /// This value only grows
+    current_presentation_id: PresentationId,
 }
 
 impl ContextData {
@@ -214,6 +225,8 @@ impl ContextData {
                 .iter()
                 .map(|&buffer_id| (buffer_id, PresentationBufferState::Unassigned))
                 .collect(),
+            current_presentation_id: PresentationId(0),
+            next_presentation_id: PresentationId(1),
         }
     }
 
@@ -252,6 +265,7 @@ impl ContextData {
             );
             Some(buffer_id)
         } else {
+            error!("No available presentation buffer: {:?}", self.buffer_ids);
             None
         }
     }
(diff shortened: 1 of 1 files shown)


gterzian:
thanks, will review tomorrow.

gterzian (approved):
So while the code is simple, the algorithm it implements comes with some surprising properties, which I modeled using TLA+(took me 38 min). 

One surprise:

- The invariant is not "next should always be higher than current", as I thought when reading the code, it is rather "when they differ, next should always be higher". 

One thing I knew, but is confirmed:
- For a given "operation" on the ids, `update_context` should always be called after `swapchain_present`(this is ensured by the messaging order). 

You can read the model at https://gist.github.com/gterzian/aa5d96a89db280017b04917eee67f6ac 

Would be good to know what you think.

components/webgpu/swapchain.rs
```diff
@@ -309,6 +323,23 @@ impl ContextData {
...
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Returns new presentation id
+    fn next_presentation_id(&mut self) -> PresentationId {
+        let res = PresentationId(self.next_presentation_id.0);
+        self.next_presentation_id.0 += 1;
```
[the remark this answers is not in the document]
sagudev: in that case we could have special handling in check_and_update (if we have try to update very big with very small), but  given that it will not be problem in practice I do not think it's worth it.

sagudev:
> So while the code is simple, the algorithm it implements comes with some surprising properties, which I modeled using TLA+(took me 38 min).

I was always interested in formal proofing and stuff like that, but there is always more practical work waiting so I never got to it.

> * The invariant is not "next should always be higher than current", as I thought when reading the code, it is rather "when they differ, next should always be higher".

This is surprising! When can this happen?

> For a given "operation" on the ids, update_context should always be called after swapchain_present(this is ensured by the messaging order).

What? Could you explain it more, because I am not sure I understand this. I am pretty sure there can be multiple swapchain_present without update_context in between and it should also be possible to do create_context, update_context then swapchain_present.

gterzian:
> This is surprising! When can this happen?

Actually I was wrong see how fixed the priming in the latest revision(I was effectively using the new id, instead of the previous one before the increment). https://gist.github.com/gterzian/aa5d96a89db280017b04917eee67f6ac/revisions

> What? Could you explain it more, because I am not sure I understand this. I am pretty sure there can be multiple swapchain_present without update_context in between and it should also be possible to do create_context, update_context then swapchain_present.

Yes I think that is correct, what I means is that there cannot be an `update_context` without having first done a `swapchain_present`, correct?

sagudev:
> Yes I think that is correct, what I mean is that there cannot be an `update_context` without having first done a `swapchain_present`, and this is guaranteed by the messaging order from script. correct?

I think not. Actually I think this is happens when we do initial call to `configure` (swapchain_present will not happen on not configured context). It should also be possible to do update_context multiple times without swapchain_present if no texture was requested via getCurrentTexture (canvas that is configured, we do resize but we do not render anything to it).

gterzian:
Ok, I think I have now figured it out, latest spec is at https://gist.github.com/gterzian/aa5d96a89db280017b04917eee67f6ac

> It should also be possible to do update_context multiple times without swapchain_present

So the invariant is not so much that next > current(this is trivially true since update_context updates both), but rather that when there are no pending map asyncs, current is next minus one.

For example, if I remove the `IF Head(pending_ids) > current_presentation_id` condition [here](https://gist.github.com/gterzian/aa5d96a89db280017b04917eee67f6ac#file-presentation-tla-L29), I get the below failure on the invariant(while the weaker next > current invariant never fails):

Screenshot 2024-10-04 at 11 55 59 AM

> I think this happens when we do initial call to configure

Does that do a `update_context` without there being a pending map async? I think that's ok. The spec does not include any condition of there being a pending map async to update the context.

components/webgpu/swapchain.rs
```diff
@@ -350,6 +381,9 @@ impl crate::WGPU {
         let mut webgpu_contexts = self.wgpu_image_map.lock().unwrap();
         let context_data = webgpu_contexts.get_mut(&context_id).unwrap();
 
+        let presentation_id = context_data.next_presentation_id();
+        assert!(context_data.check_and_update_presentation_id(presentation_id));
```
gterzian, line 385: This is trivially true, so not useful as an assertion.

sagudev:
> So the invariant is not so much that next > current(this is trivially true since update_context updates both), but rather that when there are no pending map asyncs, current is next minus one.

Sounds right.

> Does that do a update_context without there being a pending map async? I think that's ok. The spec does not include any condition of there being a pending map async to update the context.

yes, the way you wrote it before I thought that was conclusion.

components/webgpu/swapchain.rs
sagudev: Will remove it.

PR #33614, opened 2024-10-02
refactor `CGClassConstructHook` to use handwritten constructors

### LINKED ISSUE #31571: Extract generated class constructor hook into handwritten functions

Gae24:
Following what was done in the closed pull request some preamble code was left out.
Let me know if I have to include it and if the solution to the cast is ok.

- [X] These changes fix #31571

- [X] These changes do not require tests because no behaviour changes


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -6218,51 +6218,56 @@ def definition_body(self):
...
-}}
-"""
             name = self.constructor.identifier.name
             nativeName = MakeNativeName(self.descriptor.binaryNameFor(name))
-            args = ["&global", "Some(desired_proto.handle())", "CanGc::note()"]
-            constructorCall = CGMethodCall(args, nativeName, True,
-                                           self.descriptor, self.constructor)
+
+            if len(self.exposureSet) == 1:
+                target = f"dom::types::{list(self.exposureSet)[0]}"
```
Gae24, line 6242: target declaration was split because line was longer than 120

components/script/dom/bindings/constructor_hook.rs
```diff
@@ -0,0 +1,57 @@
...
+pub unsafe fn hook_html_constructor<T: DerivedFrom<Element> + DomObject>(
+    cx: JSContext,
+    args: &CallArgs,
+    global: DomRoot<GlobalScope>,
+    proto_id: PrototypeList::ID,
+    creator: unsafe fn(JSContext, HandleObject, *mut ProtoOrIfaceArray),
+) -> bool {
+    let global = DomRoot::downcast::<Window>(global).unwrap();
+    call_html_constructor::<T>(cx, args, &global, proto_id, creator)
+}
```
sagudev, line 33: I think that this is the only call to the `call_html_constructor` so we can just change call_html_constructor instead of adding another wrapper.

sagudev: Make `call_html_constructor` accept `global: DomRoot<GlobalScope>`, then do conversion to `let global = DomRoot::downcast::<Window>(global).unwrap();` inside `call_html_constructor`.

Gae24: Given that `call_html_contructor` will simply call `html_construct`, which will again upcast it, do I pass `&GlobalScope` and do the conversion to Window inside `html_construct`?

sagudev: Yes, that would be even better.

components/script/dom/bindings/constructor_hook.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    rooted!(in(*cx) let mut desired_proto = ptr::null_mut::<JSObject>());
+    let proto_result = get_desired_proto(cx, args, proto_id, creator, desired_proto.handle_mut());
+    assert!(proto_result.is_ok());
+    if proto_result.is_err() {
+        return false;
+    }
+
+    constructor(cx, args, &global, desired_proto)
+}
```
sagudev, line 41: I think this should be part of htmlconstructor.rs file.

components/script/dom/bindings/htmlconstructor.rs
```diff
@@ -377,10 +377,10 @@ pub fn push_new_element_queue() {
     ScriptThread::push_new_element_queue();
 }
 
-pub(crate) unsafe fn call_html_constructor<T: DerivedFrom<Element> + DomObject>(
+pub unsafe fn call_html_constructor<T: DerivedFrom<Element> + DomObject>(
```
sagudev, line 380: It's weird that `pub(crate)` isn't enough.

components/script/dom/bindings/constructor_hook.rs
Gae24: Will do it, but this one is used to construct all dom types that are not html elements

Gae24: Also given that there is `assert!(proto_result.is_ok());`, isn't the check `is_err()` immediately after, pointless?

components/script/dom/bindings/htmlconstructor.rs
Gae24: I guess it's because `import.rs` tries to export it pub.

Gae24:
@sagudev Is there anything else left to do?

components/script/dom/bindings/htmlconstructor.rs
sagudev: Ah, makes sense. Although most stuff ins script create should be `pub(crate)` only (to get better dead code analysis) but we should do this after we resolve #1799.

sagudev:
> @sagudev Is there anything else left to do?

LGTM, but I will let @jdm to review it, if that's what he envisioned when he opened the issue.

jdm (approved):
This is great! Thank you both!

Issue #26878, opened 2020-06-11
Implement Crypto.subtle


jdm:
https://www.w3.org/TR/WebCryptoAPI/#dfn-SubtleCrypto

This is used in the hub.link flow for joining a hubs room on a VR device.

jdm:
Hubs relies on the deriveKey, importKey, exportKey, encrypt, and decrypt methods.

jdm:
Hubs relies [specifically](https://github.com/mozilla/hubs/blob/master/src/utils/crypto.js) on:
* ECDH
* AES-CBC
* jwk
* P-256

jdm:
I've started tinkering in https://github.com/jdm/servo/tree/subtle.

jdm:
[ring](https://docs.rs/ring/0.16.15/ring/) may be a useful basis for this work. Gecko relies on NSS (eg. https://searchfox.org/mozilla-central/rev/82c04b9cad5b98bdf682bd477f2b1e3071b004ad/dom/crypto/WebCryptoTask.cpp#2727 for the ECDH implementation for deriveKey).

https://docs.rs/ring/0.16.15/ring/agreement/index.html has APIs involving ECDH, the [P256 curve](https://docs.rs/ring/0.16.15/ring/agreement/static.ECDH_P256.html). However, it doesn't seem to have AES-CBC support. The RustCrypto group has https://docs.rs/aes/0.4.0/aes/ which looks like it might suit. Given that, we may just want to rely on the [elliptic_curve](https://docs.rs/elliptic-curve/0.4.0/elliptic_curve/) and [p256](https://docs.rs/p256/0.3.0/p256/) crates instead of ring.

msub2:
Here's a more complete overview of the algorithms and what I think pulls in the least amount of additional crates for this:

RSASSA-PKCS1-v1_5 - `ring`
RSA-PSS - `ring`
RSA-OAEP - `rsa` ([Is planned in `ring` but not yet supported](https://github.com/briansmith/ring/issues/691))
ECDSA - `ring`
ECDH - `ring`
AES-CTR - `aes` and `ctr`
AES-CBC - `aes` and `cbc`
AES-GCM - `ring`
AES-KW - `aes_kw`
HMAC - `ring`
SHA - `ring`
HKDF - `ring`
PBKDF2 - `ring`

I think it's still worth pulling in ring since it covers so many of the algorithms, unless we want to only use either ring/RustCrypto crates

PR #33628, opened 2024-10-03
crypto: Begin SubtleCrypto implementation

msub2:
This begins the implementation for SubtleCrypto by adding the generateKey function with support for AES-CBC to start. Opening as draft to get feedback on the DOM interface code first, willl update the WPT expectations when everything else is looking good. This currently lets `WebCryptoAPI/generateKey/failures_AES-CBC.https.any.html` fully pass.

---


components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+use std::rc::Rc;
+
+use aes::cipher::KeyInit;
+use aes::{Aes128, Aes192, Aes256};
```
mrobinson, line 8: I'm curious about the use of `aes`. Is it possible to use a higher-level library like `ring`?

mrobinson (changes requested):
Nice work. I know this is still a draft, but I did a quick review.

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
...
+/// The spec states to run operation steps in parallel, but DOMString does not impl Send.
+unsafe impl Send for NormalizedAlgorithm {}
+
+impl Clone for Algorithm {
+    fn clone(&self) -> Self {
+        Algorithm {
+            name: self.name.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
jdm: This is just a missing feature of the current codegen. There are manual implementations of traits like Close be and PartialEq for various codegen'd ducts and enums.

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
...
+#[derive(Clone)]
+pub enum NormalizedAlgorithm {
+    #[allow(dead_code)]
+    Algorithm(Algorithm),
+    AesKeyGenParams(AesKeyGenParams),
+}
+
+#[allow(unsafe_code)]
+/// The spec states to run operation steps in parallel, but DOMString does not impl Send.
+unsafe impl Send for NormalizedAlgorithm {}
```
jdm, line 150: This is one of those things that will bite us if we ever make use of the !Send properties of DOMString. I would prefer to explicitly convert the strings to a sendable type instead of using unsafe code here.

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
...
+                Handle::Aes192(key)
+            },
+            256 => {
+                let key = Aes256::new_from_slice(&rand).map_err(|_| Error::Operation)?;
+                Handle::Aes256(key)
+            },
+            _ => return Err(Error::Operation),
+        };
+
+        Ok(CryptoKey::new_inherited(
```
jdm, line 240: I'm surprised to see a use of new_inherited here.

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+impl SubtleCrypto {
+    /// <https://w3c.github.io/webcrypto/#aes-cbc-operations>
+    fn generate_key_aes_cbc(
+        &self,
+        usages: Vec<KeyUsage>,
+        key_gen_params: AesKeyGenParams,
+        extractable: bool,
+    ) -> Result<CryptoKey, Error> {
```
jdm, line 206: This should be `Root<CryptoKey>` as far as I can see.

components/script/dom/cryptokey.rs
```diff
@@ -0,0 +1,89 @@
...
+            extractable,
+            algorithm,
+            usages,
+            handle,
+        }
+    }
+}
+
+impl CryptoKeyMethods for CryptoKey {
+    // https://w3c.github.io/webcrypto/#cryptokey-interface-members
```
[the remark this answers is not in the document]
msub2: Ah that's my bad, they were supposed to be rustdoc yeah. Will fix

components/script/dom/subtlecrypto.rs
msub2: Potentially, according to some prior investigation by jdm though ring doesn't have support for AES-CBC https://github.com/servo/servo/issues/26878#issuecomment-658241844

components/script/dom/subtlecrypto.rs
msub2: Noted, I'll rework this then

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,250 @@
...
+use crate::dom::globalscope::GlobalScope;
+use crate::dom::promise::Promise;
+use crate::realms::InRealm;
+use crate::script_runtime::JSContext;
+use crate::task_source::TaskSource;
+
+#[dom_struct]
+pub struct SubtleCrypto {
+    reflector_: Reflector,
+    #[ignore_malloc_size_of = "Defined in rand"]
```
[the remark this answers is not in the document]
msub2: I believe the MallocSizeOf issue here comes from the external `rand` crate that's used in Servo's `rand`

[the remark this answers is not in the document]
mrobinson: I think the right thing here is to move the  `#[ignore_malloc_size_of = "Defined in rand"]` decorator to the `rand` crate so this doesn't have to be repeated everywhere.

msub2:
Alright, I believe I've addressed all the feedback so far. Let me know if there's anything major still outstanding.

msub2:
Did a little bit more work to implement exportKey (that way both the failures and successes tests can run) and fixed some assumptions I made about how the aes crate works, so the CryptoKey handle now holds the actual key data we generated (the Aes structs I had before seem to be more opaque wrappers that take in the key data to facilitate encryption/decryption, but you can't retrieve the key data itself from them).

components/script/dom/subtlecrypto.rs
```diff
@@ -0,0 +1,328 @@
...
+                let exported_key = match alg_name.as_str() {
+                    ALG_AES_CBC => subtle.export_key_aes_cbc(format, &*key),
+                    _ => Err(Error::NotSupported),
+                };
+                match exported_key {
+                    Ok(k) => {
+                        let cx = GlobalScope::get_cx();
+                        rooted!(in(*cx) let mut array_buffer_ptr = ptr::null_mut::<JSObject>());
+                        let _ = unsafe {
+                            ArrayBuffer::create(
```
jdm, line 181: Are you able to use the safe [create_buffer_source](https://doc.servo.org/script/dom/bindings/buffer_source/fn.create_buffer_source.html) instead?

components/servo/Cargo.toml
```diff
@@ -32,6 +32,7 @@ webgl_backtrace = [
 ]
 
 [dependencies]
+aes = { workspace = true }
```
jdm, line 35: Why is this added here?

jdm (approved):

components/servo/Cargo.toml
msub2: Ah my bad, must be left over from when I was still figuring things out, will remove.

tests/wpt/mozilla/tests/mozilla/interfaces.https.html
```diff
@@ -239,6 +240,7 @@
   "StorageEvent",
   "StyleSheet",
   "StyleSheetList",
+  "SubtleCrypto",
```
jdm, line 243: We should probably gate these new interfaces behind a pref until more if the functionality is present.

msub2: Ah good point, I'll add that

jdm (approved):

msub2:
@mrobinson I think you need to approve as well? Still says merging is blocked

mrobinson (approved):

PR #33630, opened 2024-10-03
Less `allow(crown::unrooted_must_root)` in bindings

sagudev:
Currently we disabled unrooted_must_root lint in all bindings, now we just disable it where we absolutely need it.

This should help with crown problem in script-split: https://github.com/servo/servo/pull/33416#issuecomment-2390725477

---

- [x] These changes do not require tests because they only limit scope of allowing lints

### DIFF
--- a/components/script/dom/bindings/codegen/CodegenRust.py
+++ b/components/script/dom/bindings/codegen/CodegenRust.py
@@ -2694,7 +2694,7 @@ class CGAbstractMethod(CGThing):
     """
     def __init__(self, descriptor, name, returnType, args, inline=False,
                  alwaysInline=False, extern=False, unsafe=False, pub=False,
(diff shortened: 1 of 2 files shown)


components/script/dom/bindings/codegen/CodegenRust.py
```diff
@@ -2694,7 +2694,7 @@ class CGAbstractMethod(CGThing):
     """
     def __init__(self, descriptor, name, returnType, args, inline=False,
                  alwaysInline=False, extern=False, unsafe=False, pub=False,
-                 templateArgs=None, docs=None, doesNotPanic=False):
+                 templateArgs=None, docs=None, doesNotPanic=False, attrs=[]):
```
mrobinson, line 2697: `attrs` is a bit ambiguous. Does `extra_deocrators` work?

mrobinson (approved):

PR #33636, opened 2024-10-03
wrapper for semaphore

sharma-shray:
These reduce code repetition for semaphore operations and add meaningful error messages.

---


mrobinson (changes requested):
Hrm. I'm not sure about these changes. Having the code inline makes it more obvious that we are not handling errors and the error messages on `expect()` could probably be improved to the point where they are all unique.

PR #33645, opened 2024-10-04
Revert "compositor: Create a single cross-process compositor API (#33619)"

mrobinson:
This reverts commit f2f5614ad64927aa82aa8937ae14a6086df49d2b.

This is causing intermittent crashes: https://github.com/servo/servo/actions/runs/11167043809/job/31044255019

---


mukilan:
I was looking into this. I suspect the issue might actually be with #33611 because I keep seeing process_response_eof in the crash backtrace.

mukilan:
I guess process_response_eof is expected as that is where the final document is ready. I'll have to investigate more.

mukilan (approved):

mrobinson:
> I was looking into this. I suspect the issue might actually be with #33611 because I keep seeing process_response_eof in the crash backtrace.

It could be... For now, let's revert this one. If we still have issues we can revert #33611 as well.

PR #33646, opened 2024-10-04
clippy: Fix `too_many_arguments` in `components/layout_thread_2020`

cashall-0:
This change adds  `#[allow(clippy::too_many_arguments)]` to ignore warnings these warnings in existing code. 

---

- [X] These changes do not require tests because they do not change functionality.


components/layout_thread_2020/lib.rs
```diff
@@ -585,7 +585,7 @@ impl LayoutThread {
             snapshot_map,
         }
     }
-
+    #[allow(clippy::too_many_arguments)]
```
mrobinson, line 588: Please re-add the blank line between the two methods here.

mrobinson (changes requested):

components/layout_thread_2020/lib.rs
cashall-0: @mrobinson  I've added the blank lines. although I'd love to ask, other than readability, is there any other reason for the spaces?

mrobinson (approved):

PR #33653, opened 2024-10-04
clippy: Fix warnings in `components/script` & `components/webgpu`

taniishkaaa:
Fixes warnings in `components/script` & `components/webgpu` as a part of #31500 

---

### DIFF
--- a/components/script/dom/htmlelement.rs
+++ b/components/script/dom/htmlelement.rs
@@ -534,8 +534,8 @@ impl HTMLElementMethods for HTMLElement {
         }
 
         // Step 8: If previous is a Text node, then merge with the next text node given previous.
-        if let Some(_previous) = previous {
-            Self::merge_with_the_next_text_node(_previous)
+        if let Some(previous) = previous {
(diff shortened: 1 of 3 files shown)


components/script/dom/htmlelement.rs
```diff
@@ -534,7 +534,9 @@ impl HTMLElementMethods for HTMLElement {
         }
 
         // Step 8: If previous is a Text node, then merge with the next text node given previous.
-        previous.map(Self::merge_with_the_next_text_node);
+        if let Some(_previous) = previous {
```
mukilan, line 537: A variable beginning with underscore usually means that it is unused. Since we can shadow variable bindings in Rust, I think we should be able to reuse the same variable name here i.e 

```
if let Some(previous) = previous {
    Self::merge_with_the_next_text_node(previous)
}
```

taniishkaaa: oh okay, got it! I'll fix it.

mrobinson (approved):

PR #33666, opened 2024-10-05
Implement keyword sizes on floats

Loirooriol:
Adds support for min-content, max-content, fit-content and stretch,
for floated elements.

---

### DIFF
--- a/components/layout_2020/flow/float.rs
+++ b/components/layout_2020/flow/float.rs
@@ -27,10 +27,10 @@ use crate::dom::NodeExt;
 use crate::dom_traversal::{Contents, NodeAndStyleInfo};
 use crate::formatting_contexts::IndependentFormattingContext;
 use crate::fragment_tree::{BoxFragment, CollapsedBlockMargins, CollapsedMargin};
-use crate::geom::{LogicalRect, LogicalVec2, PhysicalPoint, PhysicalRect, ToLogical};
+use crate::geom::{LogicalRect, LogicalVec2, PhysicalPoint, PhysicalRect, Size, ToLogical};
 use crate::positioned::{relative_adjustement, PositioningContext};
 use crate::style_ext::{Clamp, ComputedValuesExt, DisplayInside, PaddingBorderMargin};
-use crate::{ContainingBlock, IndefiniteContainingBlock};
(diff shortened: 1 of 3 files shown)


components/layout_2020/flow/float.rs
```diff
@@ -914,43 +914,61 @@ impl FloatBox {
...
-                            size.clamp_between_extremums(min_box_size.block, max_box_size.block)
-                        });
-                        let tentative_inline_size = box_size.inline.auto_is(|| {
-                            let available_size =
-                                containing_block.inline_size - pbm_sums.inline_sum();
+                        let box_size = style.content_box_size(containing_block, &pbm);
+                        let max_box_size = style.content_max_box_size(containing_block, &pbm);
+                        let min_box_size = style.content_min_box_size(containing_block, &pbm);
+                        let available_inline_size =
+                            containing_block.inline_size - pbm_sums.inline_sum();
```
mrobinson, line 921: Should this take into account other floats that might be taking up inline space or do we first possibly stretch and then place among floats?

components/layout_2020/geom.rs
```diff
@@ -761,3 +774,66 @@ impl LogicalVec2<Size<LengthPercentage>> {
...
+        return Some(match self {
+            Self::Keyword(keyword) => match keyword {
+                SizeKeyword::Initial => return None,
+                SizeKeyword::MinContent => get_content_size().min_content,
+                SizeKeyword::MaxContent => get_content_size().max_content,
+                SizeKeyword::FitContent => get_content_size().shrink_to_fit(stretch_size),
+                SizeKeyword::Stretch => stretch_size,
+            },
+            Self::Numeric(numeric) => *numeric,
+        });
```
mrobinson, line 814: I would go ahead and flatten this as it will be slightly easier to read and shorter:

```suggestion
        match self {
            Self::Keyword(SizeKeyword::Initial) => None,
            Self::Keyword(SizeKeyword::MinContent) => Some(get_content_size().min_content),
            ...
```

mrobinson (approved):

components/layout_2020/flow/float.rs
Loirooriol: No, floats receive the full containing block, just like inline-blocks, see https://github.com/w3c/csswg-drafts/issues/4028#issuecomment-2374038995

PR #33667, opened 2024-10-05
Fix `align-content` set to `start` or `end` on flexbox

Loirooriol:
We need to handle `flex-wrap: wrap-reverse`.

---


mrobinson (approved):

components/layout_2020/flexbox/layout.rs
```diff
@@ -721,12 +721,14 @@ impl FlexContainer {
         };
 
         // Implement "unsafe" alignment. "safe" alignment is handled by the fallback process above.
-        let resolved_align_content = self.config.resolve_reversable_flex_alignment(
```
Otom-obhazi, line 724: Implement "unsafe" alignment. "safe" alignment is handled by the fallback process above.

"let flex_wrap_is_reversed = flex_context.config.flex_wrap_is_reversed;"

components/layout_2020/flexbox/layout.rs
```diff
@@ -721,12 +721,14 @@ impl FlexContainer {
...
-            resolved_align_content,
```
Otom-obhazi, line 725: Resolve the flex alignment based on the configuration

PR #33673, opened 2024-10-06
webgpu: Fix HTML event loop integration

### LINKED ISSUE #33601: webgpu: Update HTML event loop integration
See https://github.com/servo/servo/pull/33521#discussion_r1780911532 and https://github.com/servo/servo/pull/33521#discussion_r1780420725

sagudev:
---

- [x] These changes fix #33601

### DIFF
--- a/components/script/dom/bindings/codegen/Bindings.conf
+++ b/components/script/dom/bindings/codegen/Bindings.conf
@@ -128,6 +128,10 @@ DOMInterfaces = {
     'weakReferenceable': True, # for usage in GlobalScope https://github.com/servo/servo/issues/32519
 },
 
+'GPUCanvasContext': {
+    'weakReferenceable': True,
+},
+
 'GamepadHapticActuator': {
     'inRealms': ['PlayEffect', 'Reset']
 },
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -69,6 +69,7 @@ use uuid::Uuid;
 use webgpu::swapchain::WebGPUContextId;
 use webrender_api::units::DeviceIntRect;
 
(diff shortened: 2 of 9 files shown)


sagudev:
Per webgpu spec:
> When updating the rendering of a WebGPU canvas (an [HTMLCanvasElement](https://html.spec.whatwg.org/multipage/canvas.html#htmlcanvaselement) or an [OffscreenCanvas](https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvas) with a [placeholder canvas element](https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvas-placeholder)) with a [GPUCanvasContext](https://www.w3.org/TR/webgpu/#gpucanvascontext) context, which occurs **before getting the canvas’s image contents**

but this is not happening in servo:
```
GetCurrentTexture
canvas_data_source
update_rendering_of_webgpu_canvas
send_swap_chain_present
GetCurrentTexture
canvas_data_source
update_rendering_of_webgpu_canvas
send_swap_chain_present
```

I think `update_rendering_of_webgpu_canvas` should happen before `canvas_data_source`. cc @gterzian

sagudev:
> > I think update_rendering_of_webgpu_canvas should happen before canvas_data_source.
> 
> As per the spec, I think we should do layout only at the end of the "update the rendering", but currently we do it in multiple places, and problably also before `update_rendering_of_webgpu_canvas` is called into. `canvas_data_source` is called as part of layout right?

I think so. We cannot really fix HTML event loop integration of webgpu until HTML event loop is fixed.

sagudev:
Closed in favor of https://github.com/servo/servo/pull/34631

PR #33677, opened 2024-10-06
Drop support for '--no-minibrowser' mode

### LINKED ISSUE #33626: Drop support for '--no-minibrowser' mode in servoshell
The `--no-minibrowser` mode was added as fallback option when we first added the mini-browser mode, to be used in case the mini-browser broke servoshell.

It is currently not very well tested and has few issues that remain open ([#33254](https://github.com/servo/servo/issues/33254) and [33255](https://github.com/servo/servo/issues/33255)). 

Supporting both modes in servoshell also makes it harder to extend and maintain. For example, one of the planned tasks for our Outreachy project is to move away from 'tinyfiledialog-rs' to something modern like [egui_modal](https://crates.io/crates/egui-modal). However, `egui_modal` requires an egui `Context` to work, which is not available in --no-minibrowser mode.

Given these reasons, it seems like a good time to drop support for `--no-minibrowser` mode from servoshell.

This will involve removing:
1. The CLI argument in https://github.com/servo/servo/blob/c7a4e4f627f0bf331cedc1700ddeb92ffcba4e78/components/config/opts.rs#L567 and config option in https://github.com/servo/servo/blob/c7a4e4f627f0bf331cedc1700ddeb92ffcba4e78/components/config/opts.rs#L133
2. Remove or update the conditional logic under `ports/servoshell/desktop` folder that uses the `opts::get().minibrowser`.

Taym95:
---

- [x] These changes fix #33626

- [x]  these changes do not require tests because changes were made to ServoShell.

### DIFF
--- a/ports/servoshell/desktop/webview.rs
+++ b/ports/servoshell/desktop/webview.rs
@@ -38,7 +38,6 @@ use tinyfiledialogs::{self, MessageBoxIcon, OkCancel, YesNo};
 use super::keyutils::{CMD_OR_ALT, CMD_OR_CONTROL};
 use super::window_trait::{WindowPortsMethods, LINE_HEIGHT};
 use crate::desktop::tracing::{trace_embedder_event, trace_embedder_msg};
-use crate::parser::location_bar_input_to_url;
 
 pub struct WebViewManager<Window: WindowPortsMethods + ?Sized> {
     status_text: Option<String>,
@@ -431,7 +430,6 @@ where
             .shortcut(CMD_OR_CONTROL, 'R', || {
                 self.focused_webview_id.map(EmbedderEvent::Reload)
             })
-            .shortcut(CMD_OR_CONTROL, 'L', || None)
             // Select the first 8 tabs via shortcuts
             .shortcut(CMD_OR_CONTROL, '1', || self.focus_webview_by_index(0))
             .shortcut(CMD_OR_CONTROL, '2', || self.focus_webview_by_index(1))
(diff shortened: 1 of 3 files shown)


ports/servoshell/desktop/webview.rs
```diff
@@ -431,29 +431,7 @@ where
...
-                        location_bar_input_to_url(&input).and_then(|url| {
-                            self.focused_webview_id
-                                .map(|id| EmbedderEvent::LoadUrl(id, url))
-                        })
-                    })
-                } else {
-                    None
-                }
-            })
+            .shortcut(CMD_OR_CONTROL, 'L', || None)
```
mukilan, line 434: I think we can simply remove this call to shortcut as it doesn't have any effect.

ports/servoshell/desktop/webview.rs
```diff
@@ -431,29 +431,7 @@ where
...
-                            .as_ref()
-                            .map(|url| url.to_string())
-                            .unwrap_or_else(String::default),
-                        None => String::default(),
-                    };
-
-                    let title = "URL or search query";
-                    let input = tinyfiledialogs::input_box(title, title, &tiny_dialog_escape(&url));
-                    input.and_then(|input| {
-                        location_bar_input_to_url(&input).and_then(|url| {
```
mukilan, line 448: This now causes "unused import" warning for `location_bar_input_to_url`. Could you remove the corresponding `use` statement as well.

mukilan (changes requested):

ports/servoshell/desktop/webview.rs
Taym95: done

ports/servoshell/desktop/webview.rs
Taym95: done

mukilan (approved):
Thank you!

PR #33685, opened 2024-10-07
clippy: Fix warning in `components/script/dom`

taniishkaaa:
Fixes warnings in `components/script/dom` as a part of #31500 

---


components/script/dom/bindings/inheritance.rs
```diff
@@ -46,7 +46,7 @@ pub trait Castable: IDLInterface + DomObject + Sized {
         T: DerivedFrom<Self>,
     {
         if self.is::<T>() {
-            Some(unsafe { mem::transmute(self) })
+            Some(unsafe { mem::transmute::<&Self, &T>(self) })
         } else {
             None
         }
```
mrobinson, line 52: What warning is fixed here?

components/script/dom/identityhub.rs
```diff
@@ -36,6 +36,12 @@ pub struct IdentityHub {
     render_passes: IdentityManager<RenderPass>,
 }
 
+impl Default for IdentityHub {
+    fn default() -> Self {
+        IdentityHub::new()
+    }
+}
+
```
mrobinson, line 44: I think this is addressed in a slightly better way in #33689, which also modifies the caller to use the `default()` method. In this change it is unused.

mrobinson (changes requested):

components/script/dom/bindings/inheritance.rs
taniishkaaa: It fixes the [missing_transmute_annotations](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_transmute_annotations) warning

components/script/dom/identityhub.rs
taniishkaaa: oh okay so i should remove these changes?

mrobinson: Yes, please remove this change.

mrobinson (approved):

PR #33688, opened 2024-10-07
layout: Avoid double negation in `CellLayout::is_empty_for_empty_cells()`

### LINKED ISSUE #33678: Avoid double negation in `is_empty_for_empty_cells()`
https://github.com/servo/servo/blob/f8933a57353aeca14a6cbc60b3cb0cf98cab6c5d/components/layout_2020/table/layout.rs#L76-L82

Could avoid the double negation by using `all()` instead of `any()`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all

Taym95:
---

- [x] These changes fix #33678 

- [x] These changes do not require tests because there is no behavior change.

### DIFF
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -78,7 +78,7 @@ impl CellLayout {
             .layout
             .fragments
             .iter()
-            .any(|fragment| !matches!(fragment, Fragment::AbsoluteOrFixedPositioned(_)))
+            .all(|fragment| matches!(fragment, Fragment::AbsoluteOrFixedPositioned(_)))
     }
 }
 
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -74,7 +74,7 @@ impl CellLayout {
 
     /// Whether the cell is considered empty for the purpose of the 'empty-cells' property.
     fn is_empty_for_empty_cells(&self) -> bool {
-        !self
+        self
             .layout
             .fragments
             .iter()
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

components/layout_2020/table/layout.rs
```diff
@@ -78,7 +78,7 @@ impl CellLayout {
             .layout
             .fragments
             .iter()
-            .any(|fragment| !matches!(fragment, Fragment::AbsoluteOrFixedPositioned(_)))
+            .all(|fragment| matches!(fragment, Fragment::AbsoluteOrFixedPositioned(_)))
```
Loirooriol, line 81: This is only removing one negation, you need to remove the one in `!self` too.
That is, checking that there isn't any of the fragments which is different than AbsoluteOrFixedPositioned is equivalent to checking that all of the fragments are AbsoluteOrFixedPositioned.

Loirooriol (changes requested):

Taym95:
> Bots caught the problem indeed: https://github.com/servo/servo/actions/runs/11216734706#summary-31178398934

Sorry I missed that, empty-cells-applies-to-014.xht is passing now

Loirooriol (approved):

mrobinson:
> Bots caught the problem indeed: https://github.com/servo/servo/actions/runs/11216734706#summary-31178398934

Whoops. Sorry. I should have caught this in my review.

Loirooriol:
Lint is failing: https://github.com/servo/servo/actions/runs/11217687511/job/31179784487?pr=33688

Issue #33683, opened 2024-10-07
Replace CanGc::note() calls with arguments passed by callers


jdm:
There are a lot of CanGc::note() calls in components/script: https://github.com/search?q=repo%3Aservo%2Fservo%20CanGc%3A%3Anote&type=code

This was introduced in #33144, but in general any use of CanGc::note() is an opportunity to make the code more precise by adding a CanGc argument that is passed by the caller instead (see #33386 for an example). This is a task that will take a very, very long time to do all at once, so it's fine to just focus on one caller at a time and propagate it upwards until it becomes too much work!

If you want to work on this task, just leave a comment linking to the specific call sites that you are going to focus on.

There is no need to write or run tests for changes related to this issue; it is strictly about making garbage collection hazards more visible when reading code, and if the code compiles then everything works!

taniishkaaa:
Hey, I'd like to start with the calls in [ParseFromString](https://github.com/servo/servo/blob/2642f3ce338f5750c5cbc8878a29f8dffe832921/components/script/dom/domparser.rs#L93) and [DOMRect](https://github.com/servo/servo/blob/2642f3ce338f5750c5cbc8878a29f8dffe832921/components/script/dom/domrect.rs#L30)

jdm:
We can ignore those for right now. One option might be to have two variants of each method, one with the argument and one without, but I would need to think more about if that's feasible.

Otom-obhazi:
@mrobinson Please I will love to work on this task

jdm:
@Otom-obhazi Ok! Please list the files that you will be addressing, like https://github.com/servo/servo/issues/33683#issuecomment-2397099840.

chickenleaf:
I'd like to go ahead w these https://github.com/servo/servo/blob/c430645606396a300b57a4d9c909db959486ca4c/components/script/dom/imagedata.rs#L57

https://github.com/servo/servo/blob/c430645606396a300b57a4d9c909db959486ca4c/components/script/dom/blob.rs#L165

jdm:
Some rules to follow when making processing each `CanGc::note()`, based on reading the surrounding code:
* if it's within a bare function or a method in a non-trait `impl` block, add the `CanGc` argument to the existing arguments and replace the `CanGc::note()` call
* if it's within a `task!` closure, keep the `CanGc::note()` call and do not add the `CanGc` argument to the existing arguments
[...]

PR #33692, opened 2024-10-07
Replace CanGc::note() calls with arguments in `components/script/dom`

taniishkaaa:
Replaces `CanGc::note()` calls with arguments passed by callers as defined in #33683 

---

### DIFF
--- a/components/script/dom/bindings/codegen/Bindings.conf
+++ b/components/script/dom/bindings/codegen/Bindings.conf
@@ -118,7 +118,7 @@ DOMInterfaces = {
 },
 
 'MediaDevices': {
-    'inRealms': ['GetUserMedia'],
+    'inRealms': ['GetUserMedia', 'GetClientRects', 'GetBoundingClientRect'],
 },
 
 'MediaQueryList': {
--- a/components/script/dom/bindings/codegen/Bindings.conf
+++ b/components/script/dom/bindings/codegen/Bindings.conf
@@ -71,7 +71,7 @@ DOMInterfaces = {
 },
 
 'Element': {
-    'canGc': ['InsertAdjacentHTML', 'SetInnerHTML', 'SetOuterHTML'],
+    'canGc': ['SetInnerHTML', 'SetOuterHTML', 'InsertAdjacentHTML', 'GetClientRects', 'GetBoundingClientRect'],
 },
 
 'EventSource': {
--- a/components/script/dom/bindings/codegen/Bindings.conf
+++ b/components/script/dom/bindings/codegen/Bindings.conf
@@ -23,7 +23,7 @@ DOMInterfaces = {
 },
 
 'BaseAudioContext': {
-    'inRealms': ['DecodeAudioData', 'Resume'],
+    'inRealms': ['DecodeAudioData', 'Resume', 'ParseFromString', 'GetBounds', 'GetClientRects'],
 },
 
 'Blob': {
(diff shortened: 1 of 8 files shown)


jdm (approved):
Looking good! Want to make one last set of changes as well?

components/script/dom/element.rs
```diff
@@ -2353,6 +2353,7 @@ impl ElementMethods for Element {
                     rect.origin.y.to_f64_px(),
                     rect.size.width.to_f64_px(),
                     rect.size.height.to_f64_px(),
+                    CanGc::note(),
```
jdm, line 2356: If you modify Bindings.conf like https://github.com/servo/servo/blob/028026bebe992431ede7c57b7e3065616d859026/components/script/dom/bindings/codegen/Bindings.conf#L43, we can add CanGc arguments to the generated trait methods that you touch in this PR as well.

taniishkaaa: sounds great! let me get on that :)

components/script/dom/bindings/codegen/Bindings.conf
```diff
@@ -40,7 +40,7 @@ DOMInterfaces = {
 
 'Range': {
     'weakReferenceable': True,
-    'canGc': ['CreateContextualFragment', 'ExtractContents', 'CloneContents', 'CloneRange', 'SurroundContents'],
+    'canGc': ['CreateContextualFragment', 'ExtractContents', 'CloneContents', 'CloneRange', 'SurroundContents', 'ParseFromString', 'GetBounds', 'GetClientRects'],
```
jdm, line 43: Sorry, I wasn't clear. Each canGc section in this file applies to a single interface name. You will need to add new sections for each interface you want to change in this way, then run a build to verify that the arguments are now present, and then add the missing arguments once the build fails.

taniishkaaa: okay, i think i understand! so the `'weakReferenceable': True,` attribute should be there for all the new sections, right?

jdm: No. Any interface that is not already present can have a new section added with only the canGc property present.

taniishkaaa: also, is this a valid section relating to the changes in `domrect.rs`?
`'DOMRect': {
    'weakReferenceable': True,
    'canGc': ['new'],
},`

jdm: Ah, no, it's only for methods that appear in the DOMRectMethods trait in this case.

taniishkaaa: Can you please check if the changes are correct this time?

components/script/dom/bindings/codegen/Bindings.conf
```diff
@@ -51,6 +51,14 @@ DOMInterfaces = {
     'weakReferenceable': True,
 },
 
+'DOMParser': {
+    'canGc': ['ParseFromString'],
+},
+
+'Element': {
+    'canGc': ['GetClientRects', 'GetBoundingClientRect'],
+},
```
jdm, line 60: Oh, I see what happened. There are already DOMParser and Element sections in this file, so you should modify those instead.

taniishkaaa: Done!

jdm (approved):
Thanks!

mrobinson:
Looks like there are some merge conflicts with this PR now.

taniishkaaa:
> Looks like there are some merge conflicts with this PR now.

I've rebased & pushed again. The conflicts should be resolved now.

PR #33689, opened 2024-10-07
clippy: Fix some clippy warnings in `components/script` and `components/devtools`

komutumu:
This PR fixes some clippy warnings by;

- Creating default implementations
- Replacing ```.filter_map ``` with ```.map```
- Collapsing ``` if let ``` ``` mach``` with ```if let```
---
- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] These changes are part of #31500
- [X] These changes do not require tests because they do not touch functionality.

PR #33693, opened 2024-10-07
clippy: Fix filter_map warning in components\devtools\actors\inspecto…

Noble14477:
…r\style_rule.rs:186:27

Got this warning: this .filter_map can be written more simply using .map. 
The warning indicates that we do not need to use .filter_map with Some(...) because there’s no actual filtering happening; every item passed into filter_map is wrapped in Some, so .map can achieve the same result.
---

- X `./mach build -d` does not report any errors
- [ ] These changes fix filter_map warning in components\devtools\actors\inspector\style_rule.rs:186:27

### DIFF
--- a/components/devtools/actors/inspector/style_rule.rs
+++ b/components/devtools/actors/inspector/style_rule.rs
@@ -185,18 +185,16 @@ impl StyleRuleActor {
             css_text: "".into(), // TODO: Specify the css text
             declarations: style
                 .into_iter()
-                .filter_map(|decl| {
-                    Some(AppliedDeclaration {
-                        colon_offsets: vec![],
-                        is_name_valid: true,
-                        is_used: IsUsed { used: true },
-                        is_valid: true,
-                        name: decl.name,
-                        offsets: vec![], // TODO: Get the source of the declaration
-                        priority: decl.priority,
-                        terminator: "".into(),
-                        value: decl.value,
-                    })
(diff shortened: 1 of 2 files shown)


components/devtools/actors/inspector/style_rule.rs
```diff
@@ -185,18 +185,16 @@ impl StyleRuleActor {
...
             declarations: style
                 .into_iter()
-                .filter_map(|decl| {
-                    Some(AppliedDeclaration {
-                        colon_offsets: vec![],
-                        is_name_valid: true,
-                        is_used: IsUsed { used: true },
-                        is_valid: true,
-                        name: decl.name,
-                        offsets: vec![], // TODO: Get the source of the declaration
```
eerii, line 195: Could you please add the "TODO" comment back? It usually means that there is some unfinished business! But otherwise looks good c:

Also, you need to sign your commits, if you go to "Checks" or [here](https://github.com/servo/servo/pull/33693/checks?check_run_id=31188045885) it has instructions on the git command to run.

Noble14477: Can you please check if its okay now...

mrobinson:
I'm fairly sure this was already done in #33689. Be sure to rebase your changes before uploading them as a PR. Thank you for the fix though!

PR #33705, opened 2024-10-07
Fix clamp-like pattern warning in font.rs

chickenleaf:
**File**: components/fonts/platform/windows/font.rs

**PR Description**
**Issue**: Clippy warned about a clamp-like pattern without using the clamp function in the stretch calculation.

---
**Change**:
Updated `let stretch = match min(9, max(1, width_val))` to `let stretch = match width_val.clamp(1, 9)` in the stretch calculation.

**Impact**:
Improved code readability.

---

- [X] ./mach build -d does not report any errors
- [X] ./mach test-tidy does not report any errors
- [X] These changes are part of Fix as many clippy problems as possible #31500
- [X] These changes do not require tests because they do not alter functionality.

### DIFF
--- a/components/fonts/platform/windows/font.rs
+++ b/components/fonts/platform/windows/font.rs
@@ -167,7 +167,7 @@ impl PlatformFontMethods for PlatformFont {
         };
 
         let weight = StyleFontWeight::from_float(weight_val as f32);
-        let stretch = match min(9, max(1, width_val)) {
+        let stretch = match width_val.clamp(1, 9) {
             1 => StyleFontStretch::ULTRA_CONDENSED,
             2 => StyleFontStretch::EXTRA_CONDENSED,
             3 => StyleFontStretch::CONDENSED,


Loirooriol (commented):
The change looks good, but

> Also prevents potential panics if max < min in future code

It's the opposite, `clamp` would panic while the original would not. But this is not a worry since `1 < 9` and I doubt this will change in a way that `max < min`.

> ☐  ./mach test-tidy does not report any errors

So it reports errors?

chickenleaf:
> The change looks good, but
> 
> > Also prevents potential panics if max < min in future code
> 
> It's the opposite, `clamp` would panic while the original would not. But this is not a worry since `1 < 9` and I doubt this will change in a way that `max < min`.
> 
> > ☐  ./mach test-tidy does not report any errors
> 
> So it reports errors?

i havent checked ./mach test-tidy the github issue only asked for the checking of ./mach build -d ? i was informed the same in the outreachy group.

chickenleaf:
> The change looks good, but
> 
> > Also prevents potential panics if max < min in future code
> 
> It's the opposite, `clamp` would panic while the original would not. But this is not a worry since `1 < 9` and I doubt this will change in a way that `max < min`.
> 
> > ☐  ./mach test-tidy does not report any errors
> 
> So it reports errors?

Hello! ./mach test-tidy reports no errors.

Loirooriol (approved):
But please fix the incorrect "prevents potential panics if max < min in future code" comment.

PR #33708, opened 2024-10-07
Refactor complex type definitions in components/layout_2020/style_ext.rs

chickenleaf:
**File**: components/layout_2020/style_ext.rs

**PR Description**

**Issue**: Clippy warned about a complex return type in `content_box_sizes_and_padding_border_margin`.

**Changes**:

1. Simplified the return type by grouping the three `LogicalVec2<Size<Au>>` values into a single tuple.
2. Updated the code to match the new return structure. (line 500 and 538)

**Impact**: Reduced complexity and improved readability without external type definitions.

**note**: Please check the block structure/indentation.

---


Loirooriol (commented):
I don't think this is simplifying anything. I would just ignore the warning, or create a struct with 3 fields for preferred, min and max sizes, and maybe another one for the PaddingBorderMargin.

Loirooriol:
So there are other places where we need to store the preferred, min and max sizes, e.g.

https://github.com/servo/servo/blob/c1b744b2b2c37cdb119a3d3f761822d22f938339/components/layout_2020/flexbox/layout.rs#L58-L60
https://github.com/servo/servo/blob/c1b744b2b2c37cdb119a3d3f761822d22f938339/components/layout_2020/table/layout.rs#L2839-L2841

But the types aren't much consistent... I guess I would leave it as-is for now, you can add `#[allow(clippy::type_complexity)]` (or whatever the complaint was) to ignore the warning. We may revisit after implementing size keywords.

mukilan:
Closing this as the lint was allowed in #33894.

PR #33709, opened 2024-10-07
Improve `Au(0)` constructions

hackerbirds:
I replaced all `Au(0)` constructions to `Au::zero()` (or other utility functions when possible), except in the legacy layout. This ended up changing some line lengths enough that I had to re-run `./mach fmt` which is the cause for the few formatting changes in the commit.

Upsides:
 - Unify all `Au(0)`s down to a single implementation
 - More readable code (imo)

---

- [X] These changes do not require tests because no logic has been modified

### DIFF
--- a/components/fonts/shaper.rs
+++ b/components/fonts/shaper.rs
@@ -26,6 +26,7 @@ use harfbuzz_sys::{
     HB_OT_LAYOUT_BASELINE_TAG_ROMAN,
 };
 use log::debug;
+use num_traits::Zero;
 
 use crate::platform::font::FontTable;
 use crate::{
@@ -107,11 +108,11 @@ impl ShapedGlyphData {
             let x_advance = Au::from_f64_px(x_advance);
             let y_advance = Au::from_f64_px(y_advance);
 
-            let offset = if x_offset == Au(0) && y_offset == Au(0) && y_advance == Au(0) {
+            let offset = if x_offset.is_zero() && y_offset.is_zero() && y_advance.is_zero() {
                 None
             } else {
                 // adjust the pen..
-                if y_advance > Au(0) {
+                if y_advance > Au::zero() {
                     *y_pos -= y_advance;
                 }
 
@@ -466,7 +467,7 @@ impl Shaper {
         let mut glyph_span = 0..0;
         let mut byte_range = 0..0;
 
-        let mut y_pos = Au(0);
+        let mut y_pos = Au::zero();
 
         // main loop over each glyph. each iteration usually processes 1 glyph and 1+ chars.
         // in cases with complex glyph-character associations, 2+ glyphs and 1+ chars can be
--- a/components/layout_2020/table/layout.rs
+++ b/components/layout_2020/table/layout.rs
@@ -1041,10 +1041,10 @@ impl<'a> TableLayout<'a> {
             |column_index: &usize| self.column_measures[*column_index].percentage.0 > 0.;
         let has_percent_zero = |column_index: &usize| !has_percent_greater_than_zero(column_index);
         let has_max_content = |column_index: &usize| {
-            self.column_measures[*column_index]
+            !self.column_measures[*column_index]
                 .content_sizes
-                .max_content !=
-                Au(0)
+                .max_content
+                .is_zero()
         };
 
         let max_content_sum =
@@ -2862,8 +2862,8 @@ fn get_outer_sizes_for_measurement(
     let min_size = style.min_box_size(writing_mode);
     let max_size = style.max_box_size(writing_mode);
     (
-        outer_size(size.map(|v| get_size_for_axis(v).unwrap_or(Au(0)))),
-        outer_size(min_size.map(|v| get_size_for_axis(v).unwrap_or(Au(0)))),
+        outer_size(size.map(|v| get_size_for_axis(v).unwrap_or_else(Au::zero))),
+        outer_size(min_size.map(|v| get_size_for_axis(v).unwrap_or_else(Au::zero))),
         outer_size(max_size.map(|v| get_size_for_axis(v).unwrap_or(MAX_AU))),
         size.inline.is_keyword(),
         get_size_percentage_contribution(&size, &max_size),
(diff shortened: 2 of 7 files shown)


mrobinson:
I think it's fine to do this for `layout_2020`, but I don't think I'd make too many unnecessary changes to legacy layout. Do you mind reverting those parts of this PR? Thanks.

hackerbirds:
> I think it's fine to do this for `layout_2020`, but I don't think I'd make too many unnecessary changes to legacy layout. Do you mind reverting those parts of this PR? Thanks.

Sure no problem!

hackerbirds:
Sorry about fucking up the rebase, it should be good now :sweat:

mrobinson (approved):

mrobinson:
@hackerbirds Looks good. Please be sure to sign your commit as well. More details at https://github.com/servo/servo/pull/33709/checks?check_run_id=31251106854.

mrobinson:
@hackerbirds Thanks. I've updated the PR description and removed the link about longer linking time. I don't see how this can possibly affect linking time.

components/fonts/shaper.rs
```diff
@@ -107,16 +108,17 @@ impl ShapedGlyphData {
...
 
-            let offset = if x_offset == Au(0) && y_offset == Au(0) && y_advance == Au(0) {
-                None
-            } else {
-                // adjust the pen..
-                if y_advance > Au(0) {
-                    *y_pos -= y_advance;
-                }
+            let offset =
+                if x_offset == Au::zero() && y_offset == Au::zero() && y_advance == Au::zero() {
```
Loirooriol, line 112:
```suggestion
                if x_offset.is_zero() && y_offset.is_zero() && y_advance.is_zero() {
```

components/layout_2020/table/layout.rs
```diff
@@ -1044,7 +1044,7 @@ impl<'a> TableLayout<'a> {
             self.column_measures[*column_index]
                 .content_sizes
                 .max_content !=
-                Au(0)
+                Au::zero()
```
Loirooriol, line 1047: Ditto

components/layout_2020/table/layout.rs
```diff
@@ -2862,8 +2862,8 @@ fn get_outer_sizes_for_measurement(
     let min_size = style.min_box_size(writing_mode);
     let max_size = style.max_box_size(writing_mode);
     (
-        outer_size(size.map(|v| get_size_for_axis(v).unwrap_or(Au(0)))),
-        outer_size(min_size.map(|v| get_size_for_axis(v).unwrap_or(Au(0)))),
+        outer_size(size.map(|v| get_size_for_axis(v).unwrap_or(Au::zero()))),
+        outer_size(min_size.map(|v| get_size_for_axis(v).unwrap_or(Au::zero()))),
```
Loirooriol, line 2866: Maybe `unwrap_or_else(Au::zero)` to reduce the number of parentheses? But either is fine.

components/layout_thread/lib.rs
```diff
@@ -875,7 +876,8 @@ impl LayoutThread {
                         root_flow.overflow.scroll.size
                     };
 
-                    let origin = Rect::new(Point2D::new(Au(0), Au(0)), root_size).to_layout();
+                    let origin =
+                        Rect::new(Point2D::new(Au::zero(), Au::zero()), root_size).to_layout();
```
Loirooriol, line 880: I guess this could use `Point2D::default()`

components/layout_2020/table/layout.rs
hackerbirds: I do prefer the `unwrap_or_else` way

components/fonts/shaper.rs
hackerbirds: This function is present in `num_traits`'s `Zero` trait, but I'm using `euclid`'s `Zero` trait which doesn't export `.is_zero()`. Should I add `num_traits`/the feature flag in `app_units` instead by modifying `Cargo.toml` or leave it as is?

hackerbirds: Perhaps I should use `style::Zero` instead?

Loirooriol: Ah, yeah, that's annoying, you can leave it as is

components/layout_2020/table/layout.rs
hackerbirds: This works here because style::Zero is used in this module, but let me know what I should do about the other case.

components/layout_thread/lib.rs
hackerbirds: `Point2D::zero()` or `Point2D::origin()` also seem to work. I think I will go with `Point2D::zero()` because it's more obvious what it does.

hackerbirds:
@Loirooriol thank you! I implemented your suggestions now

mrobinson (approved):

components/fonts/shaper.rs
mrobinson: Hrm. You should import `num_traits::Zero` here instead of the `euclid` version for this to work properly. `Au` does implement that and it includes `is_zero()`.

mrobinson (changes requested):
Looks good to me,

components/fonts/shaper.rs
hackerbirds: > Hrm. You should import `num_traits::Zero` here instead of the `euclid` version for this to work properly. `Au` does implement that and it includes `is_zero()`.

It does, but `components/fonts` doesn't have the `num-traits` dependency inside its `Cargo.toml`, so I'd need to add it before using that trait IIRC. If that's ok to add a dependency here though I can do that

mrobinson: It's okay to do that. We already depend on `num-traits` so this isn't adding else to the build chain other than perhaps modifying the crate build order (though unlikely).

hackerbirds: Ok, it's done now

mrobinson (approved):
Thank you!

PR #33717, opened 2024-10-08
Fix handling of `__Secure-` and `__Host-` Cookie prefixes

simonwuelker:
Cookie-Prefixes like "__Host-" and "__Secure-" are case insensitive
as per https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-15.html#name-storage-model. Servo previously checked for prefixes in a case-sensitive way. Additionally, the checks were doing things that were not specified in the specification.

Since the implementation and the specification had diverged a bit
 i also updated/added spec comments 
and slightly restructured the code of `ServoCookie::new_wrapped` so its easier to follow. However,
the only change in behavior is the prefix check described above.

This makes servo pass all wpt tests in `cookies/prefix/`

[try run](https://github.com/simonwuelker/servo/actions/runs/11228009029/job/31211572221)

---

- [x] There are tests for these changes (wpt)


jdm (approved):
Thank you for this cleanup!

PR #33718, opened 2024-10-08
Refactor ContextForRequestInterrupt for thread safety and fix clippy …

chickenleaf:
**File:** components/script/script_runtime.rs

**PR Description**
**Issue**: Clippy warned about ContextForRequestInterrupt having unsafe traits that could lead to potential issues.

**Change:**

1. Introduced `ThreadSafeContext` struct to wrap `*mut RawJSContext` and implement `Send` and `Sync` safely.
2. Updated `ContextForRequestInterrupt` to use `ThreadSafeContext` instead of a raw pointer.
3. Adjusted the constructor and method signatures to accommodate the new structure.

---

- [X] These changes do not require tests because they are not part of functionality.


gterzian:
Thank you, I will review this tomorrow.

components/script/script_runtime.rs
```diff
@@ -915,11 +915,19 @@ unsafe fn set_gc_zeal_options(_: *mut RawJSContext) {}
 /// enabling an interrupt to be requested
 /// from a thread other than the one running JS using that context.
 #[derive(Clone)]
-pub struct ContextForRequestInterrupt(Arc<Mutex<Option<*mut RawJSContext>>>);
```
mrobinson, line 918: Is it possible to fix this issue by simply modifying this `Arc` to be an `Rc`?

chickenleaf: > Is it possible to fix this issue by simply modifying this `Arc` to be an `Rc`?

absolutely yes, I could do that right away. I actually wasn't sure whether this context will only be used in a single-threaded scenario or if it might be accessed from multiple threads in the future. And i ended up choosing the thread safe option. Please let me know if you advice the Rc approach.

chickenleaf: > Is it possible to fix this issue by simply modifying this `Arc` to be an `Rc`?

hi I just rechecked this, the documentation explicitly states that this needs to be callable from any thread. So I guess it is best to proceed with Send Sync? I've used a custom wrapper with Send Sync here, please let me know if a direct Send Sync is preferred here.

chickenleaf:
hello! is the Rc approach preferred for this?

components/script/script_runtime.rs
```diff
@@ -915,11 +915,19 @@ unsafe fn set_gc_zeal_options(_: *mut RawJSContext) {}
...
+pub struct ContextForRequestInterrupt(Arc<Mutex<Option<ThreadSafeContext>>>);
```
sagudev, line 918: I am also not sure about this, as safety of sending JSContext is from mutex IIUC, and generally JSContext is not thread safe, so `ThreadSafeContext` is easily completely wrong. My feeling is that clippy gives bad warnings for our safe wrapper, so it might be better to just allow some of them.

chickenleaf: noted!!

chickenleaf:
> Thank you, I will review this tomorrow.

helloo @gterzian ! could you just confirm if i should just allow this clippy lint and leave it be for now?

components/script/script_runtime.rs
sagudev: cc @jdm

gterzian (changes requested):
I do not have the immediate answer for the questions raised so far, but I suggest looking into:

1. Is 'js::api::JSContext' safe to share between threads? 
2. How does Gecko use 'JS_RequestInterruptCallback'?
 
I think the context should be thread safe otherwise I don't see how the interrupt can be requested from another thread. (The mutex is necessary only for the option to work properly with shutdown, see the comment at 'request_interrupt')

If we can confirm that the raw context is send and sync, then we can add those traits to the existing safe 'JSContext', and use that inside 'ContextRequestInterrupt'.

sagudev:
> If we can confirm that the raw context is send and sync

I am pretty sure it's not: https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/JSContext.h#148

> A JSContext encapsulates the thread local state used when using the JSruntime.

sagudev:
More exactly: Generally JSContext is not thread safe (send/sync) but some parts are, [`JSContext::requestInterrupt`](https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/Runtime.cpp#457) only access/changes those parts (they are atomic) so that is thread safe.

chickenleaf:
so i guess this PR would be redundant, could ignore the clippy warning here i guess. Im so sorry i wasnt sure about JSContext not wholly being thread unsafe.

mrobinson:
@chickenleaf Apologies for sending you down the wrong path here.

PR #33721, opened 2024-10-08
clippy: Add safety documentation to CustomTraceable trait

rexbrown21:
Added safety documentation to the `CustomTraceable` trait and its implementations to comply with Clippy's requirements.

- [X] These changes are part of Fix as many clippy problems as possible Fix as many clippy problems as possible #31500

- [X] These changes do not require tests because the changes are only documentation-related.


mukilan:
As discussed in Zulip, safety documention is not a good first bug for contributors who are new to Servo, as this requires expert understanding of the domain. I recommend that instead of adding the documentation yourself, you create a github issue about the missing safety documentation and then disable the warning with an allow directive and include the link to the issue as a comment in the code.

I'll close this PR for now.

PR #33727, opened 2024-10-08
clippy: Fix `type_complexity` warning in `components/script/dom`

taniishkaaa:
Fixes `type_complexity` warning in `components/script/dom/range.rs` as a part of #31500  

---

- [X] These changes do not require tests because because they do not modify functionality, they only fix lint warnings.

### DIFF
--- a/components/script/dom/range.rs
+++ b/components/script/dom/range.rs
@@ -583,9 +583,11 @@ impl RangeMethods for Range {
         }
 
         // Steps 5-12.
-        let first_contained_child = self.contained_children()?.first_partially_contained_child;
-        let last_contained_child = self.contained_children()?.last_partially_contained_child;
-        let contained_children = self.contained_children()?.contained_children;
+        let ContainedChildren {
+            first_partially_contained_child: first_contained_child,
+            last_partially_contained_child: last_contained_child,
+            contained_children,
+        } = self.contained_children()?;
 
         if let Some(child) = first_contained_child {
             // Step 13.
@@ -698,9 +700,11 @@ impl RangeMethods for Range {
         }
 
         // Steps 5-12.
-        let first_contained_child = self.contained_children()?.first_partially_contained_child;
-        let last_contained_child = self.contained_children()?.last_partially_contained_child;
-        let contained_children = self.contained_children()?.contained_children;
+        let ContainedChildren {
+            first_partially_contained_child: first_contained_child,
+            last_partially_contained_child: last_contained_child,
+            contained_children,
+        } = self.contained_children()?;
 
         let (new_node, new_offset) = if start_node.is_inclusive_ancestor_of(&end_node) {
             // Step 13.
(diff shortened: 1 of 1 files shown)


components/script/dom/range.rs
```diff
@@ -583,8 +583,9 @@ impl RangeMethods for Range {
         }
 
         // Steps 5-12.
-        let (first_contained_child, last_contained_child, contained_children) =
-            self.contained_children()?;
+        let first_contained_child = self.contained_children()?.first_partially_contained_child;
```
jdm, line 586: Rather than calling this method three times, we can do:
```
let ContainedChildren {
  first_partially_contained_child,
  last_partially_contained_child,
  contained_children,
} = self.contained_children()?;

File: components/script/dom/range.rs [hunk 2]
```diff
@@ -697,8 +698,9 @@ impl RangeMethods for Range {
...
```
jdm, line 701: Same here.

jdm (changes requested):

components/script/dom/range.rs
taniishkaaa: hmm, makes sense

taniishkaaa: but for destructuring them in the next step, we'll have to do:
```
let ContainedChildren {
    first_partially_contained_child: first_contained_child,
    last_partially_contained_child: last_contained_child,
    contained_children,
} = self.contained_children()?;
```

jdm (approved):

components/script/dom/range.rs
```diff
@@ -52,6 +52,12 @@ pub struct Range {
     associated_selections: DomRefCell<Vec<Dom<Selection>>>,
 }
 
+pub struct ContainedChildren {
```
mukilan, line 55: I think this type doesn't have to be `pub` since the method `contained_children` is not pub itself and not used outside this file.

components/script/dom/range.rs
```diff
@@ -192,11 +192,11 @@ impl Range {
             return Err(Error::HierarchyRequest);
         }
 
-        Ok((
-            first_contained_child,
-            last_contained_child,
+        Ok(ContainedChildren {
+            first_partially_contained_child: first_contained_child,
```
mukilan, line 196: Rather than doing 

```
first_partially_contained_child: first_contained_child,
```

we can also just rename the `let` bindings `first_contained_child` and `last_contained_child` in lines 164 and 174 respectively. Same in the code where this result is being used. This way the code will closely match the specification.

taniishkaaa: you mean rename them to `first_partially_contained_child` and `last_partially_contained_child`?

taniishkaaa: okay, I'll open a follow-up PR with the changes!

taniishkaaa: Follow up at #33758

Issue #33726, opened 2024-10-08
Resolve Clippy Warning: Arc not Send and Sync in refcounted.rs


chickenleaf:
**Describe the bug:**

**Issue:**
Clippy warns about using `Arc` with `TrustedReference`:

```
warning: usage of an Arc that is not Send and Sync
   --> components/script/dom/bindings/refcounted.rs:269:36
    let refcount = Arc::new(TrustedReference::new(ptr));
```

**Question:**
Should we implement `Send` and `Sync` for `TrustedReference`, or use a `Mutex` for thread safety?

**Context:**
Implementing `Sync` requires safety documentation, needing expert input. 

**To Reproduce:**
`.\mach cargo-clippy`

--- 
**Next Steps:**
1. Disable Clippy lint temporarily.
2. Plan for `Send`/`Sync` implementation or `Mutex` usage.
3. Request expert review for best approach.

---

mrobinson:
Perhaps `Rc` is enough here.

mrobinson:
To expand, creating an `Arc` that isn't `Send` and `Sync` means that the `Arc` itself isn't `Send` and `Sync` either. Normally trying to pass an instance of such a struct across a thread boundary would cause an error. It's possible that this type no longer crosses the thread boundary or that it never did. In both of those cases an non-threadsafe `Rc` should be sufficient. It seems reasonable to try to convert the type to `Rc` and see if the compiler complains. The only other step would be to see how it is used. Is it used in datastructures that force the implementation of `Send` and `Sync` with unsafe code. If so you need to look at bit more carefully.

chickenleaf:
> To expand, creating an `Arc` that isn't `Send` and `Sync` means that the `Arc` itself isn't `Send` and `Sync` either. Normally trying to pass an instance of such a struct across a thread boundary would cause an error. It's possible that this type no longer crosses the thread boundary or that it never did. In both of those cases an non-threadsafe `Rc` should be sufficient. It seems reasonable to try to convert the type to `Rc` and see if the compiler complains. The only other step would be to see how it is used. Is it used in datastructures that force the implementation of `Send` and `Sync` with unsafe code. If so you need to look at bit more carefully.

So I tried out the Rc change and it is throwing me some 4 errors. I think Arc is needed here? I think the code is being used in a threaded context?

mrobinson:
> So I tried out the Rc change and it is throwing me some 4 errors. I think Arc is needed here? I think the code is being used in a threaded context?

[...]

PR #33741, opened 2024-10-08
Clippy/fix non send sync arc trusted reference

chickenleaf:
**File**: components/script/dom/bindings/refcounted.rs

**PR Description**

**Issue**: Clippy warned about using an Arc with TrustedReference, which is not Send and Sync.

**Changes:**

1. Replaced `Arc::new(TrustedReference::new(ptr))` with `Rc::new(TrustedReference::new(ptr))` in the file.
2. Updated the import statement from `use std::sync::{Arc, Weak};` to `use std::rc::{Rc, Weak};`.

---

- [X] These changes do not require tests because they are not part of functionality.


components/script/dom/bindings/refcounted.rs
```diff
@@ -161,7 +160,7 @@ impl TrustedPromise {
 pub struct Trusted<T: DomObject> {
     /// A pointer to the Rust DOM object of type T, but void to allow
     /// sending `Trusted<T>` between threads, regardless of T's sendability.
-    refcount: Arc<TrustedReference>,
+    refcount: Rc<TrustedReference>,
```
sagudev, line 163: Given the comment and usage I am pretty sure we need to keep `Arc` here, right @jdm?

chickenleaf: > Given the comment and usage I am pretty sure we need to keep `Arc` here, right @jdm?

hello! there was a discussion here on the changes to be done: https://github.com/servo/servo/issues/33726 please go through them and let me know if this is the right way about it.

sagudev: The problem is that we impl `Send` for `Trusted` (and `Clone` too): https://github.com/servo/servo/blob/ff6523c37e132605c0945bdc49f15b75b0566d2e/components/script/dom/bindings/refcounted.rs#L169, so if we send Trusted to another thread it will not lower ref count on drop in other thread, so we will keep it alive permanently in main thread.

jdm: You're right. Switching to Rc here looks like we risk racing modifying the refcount value between threads, so this change is not safe.

chickenleaf: hello @jdm @sagudev , could you please let me know if this warning is best ignored or should i be implementing either a Send and Sync for TrustedReference or Wrap TrustedReference in a Mutex ? alsoo i need help with deciding the implementation of a similar PR https://github.com/servo/servo/pull/33718 I am still trying to understand the larger context of their usage, any insights on that would be super helpful!!

sagudev: I think we should just ignore (allow) clippy lint (probably for whole script crate).

mrobinson: We are very close to fixing these issues in the script crate. I don't think we should ignore errors, because it will cause many more to appear.

chickenleaf: :sob:

mrobinson: > When I said ignore I meant allow as some clippy errors are simply wrong (sometimes). This is known, so that's why there are not in rustc.

Ah, that makes sense. Thanks for the clarification.

jdm: Agreed with sagudev. I'm going to close this PR, since this is no longer the desired fix for this particular warning instance. Sorry for sending you down the wrong path!

PR #33743, opened 2024-10-09
Fix clippy warnings related to safety documentation and filter_map.

rexbrown21:
Refactored the logic in `components/devtools/actors/inspector/page_style.rs` to simplify the usage of `filter_map`, addressing type mismatch errors and improving readability.

### DIFF
--- a/components/devtools/actors/inspector/page_style.rs
+++ b/components/devtools/actors/inspector/page_style.rs
@@ -172,7 +172,7 @@ impl PageStyleActor {
 
             // For each selector (plus an empty one that represents the style attribute)
             // get all of the rules associated with it.
-            let entries =
+            let entries: Vec<AppliedEntry> =
                 once(("".into(), usize::MAX))
                     .chain(selectors)
                     .filter_map(move |selector| {
@@ -206,7 +206,8 @@ impl PageStyleActor {
                             is_system: false,
                             inherited: inherited.clone(),
                         })
-                    });
+                    })
+                    .collect();
             Some(entries)
         })
         .flatten()
--- a/components/script/dom/bindings/reflector.rs
+++ b/components/script/dom/bindings/reflector.rs
@@ -123,6 +123,12 @@ impl DomObject for Reflector {
 /// A trait to initialize the `Reflector` for a DOM object.
 pub trait MutDomObject: DomObject {
     /// Initializes the Reflector
+    /// 
+    /// # Safety 
+    /// This Function is 'unsafe' because it takes a raw pointer to a 'JSObject'
+    /// The caller must ensure that the pointer is valid and properly aligned.
+    /// Undefined behavior may occur if the pointer is null or if it points to
+    /// an invalid memory location.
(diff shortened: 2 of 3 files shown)


components/script/dom/bindings/reflector.rs
```diff
@@ -120,9 +120,16 @@ impl DomObject for Reflector {
     }
 }
 
-/// A trait to initialize the `Reflector` for a DOM object.
+
+// FIXME: Add safety documentation.
+// See Github Issue Number:#33725
+#[allow(clippy::missing_safety_doc)]
```
mukilan, line 126: Is this `allow` needed for the trait? There is already an `allow` on the `init_reflector` method. Isn't that sufficient?

i.e what we probably want is just:
```
/// A trait to initialize the `Reflector` for a DOM object.
pub trait MutDomObject: DomObject {
    // FIXME: Add safety documentation.
    // See Github Issue #33725.
    #[allow(clippy::missing_safety_doc)]
    /// Initializes the `Reflector` for the DOM object.
    unsafe fn init_reflector(&self, obj: *mut JSObject);
}
```

components/script/dom/bindings/trace.rs
```diff
@@ -68,12 +68,19 @@ use crate::script_runtime::{ContextForRequestInterrupt, StreamConsumer};
...
+#[allow(clippy::missing_safety_doc)] 
+
+// Brief description of the trait
+/// This trait defines a method for tracing objects in the context of garbage collection.
 pub unsafe trait CustomTraceable {
     /// Trace `self`.
     unsafe fn trace(&self, trc: *mut JSTracer);
 }
 
+
```
mukilan, line 83: We can remove the extra blank line added here.

mukilan (changes requested):
In general, it is better to keep the PR limited to closely related changes. For example, in this PR we have changes for fixing the missing safety documentation warning and a change to page_style.rs that is not related to it. It could be better to split them into separate PRs.

For this PR, I'll just rename the title to be more generic, so we can continue to use this single PR.

rexbrown21 (commented):
ill make the changes and make another PR

mukilan:
@rexbrown21 There is another pull request https://github.com/servo/servo/pull/33748  recently created, so you might want to hold off on this PR and focus on other clippy warnings that are not safety documentation related.

components/devtools/actors/inspector/page_style.rs
```diff
@@ -206,7 +206,8 @@ impl PageStyleActor {
                             is_system: false,
                             inherited: inherited.clone(),
                         })
-                    });
+                    })
+                    .collect();
```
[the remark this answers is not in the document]
mukilan: I'm not sure if the filter_map warning is still applicable. I don't see any lint warnings for `page_style.rs` in main branch. Perhaps this issue was addressed in a different PR.

mrobinson:
I think the fix in #33748 is a better fix for the missing safety docs.

PR #33746, opened 2024-10-09
Remove most usage of ReentrantMutex in font code

jdm:
ReentrantMutex is unnecessary complexity when we're only locking it for the duration of calling `.send(..)` on channel types, because there is no way for us to accidentally invoke code on the same thread that will attempt to lock the mutex again.

I was hoping to remove all usage of ReentrantMutex, but the font code in platform/freetype looks like we would double-panic if we ever panicked while the font mutex was locked, since we would then run the Drop implementation that attempts to lock the mutex again.

---


mrobinson (approved):
This seems fine to me as long as tests are still passing!

mrobinson:
I rebased this one and I'm sending it to the MQ.

PR #33748, opened 2024-10-09
clippy: Add safety documentation and clean up unsafe methods

mrobinson:
This change:

1. Adds safety documentation where it was missing.
2. Limits the scope of unsafe code in some cases to where it is actually
   unsafe.
3. Converts some free functions to associated functions and methods,
   thereby making them more likely to be called safely.

---

- [x] These changes do not require tests because it should not change behavior.

### DIFF
--- a/components/script/layout_dom/node.rs
+++ b/components/script/layout_dom/node.rs
@@ -76,6 +76,11 @@ impl<'dom> ServoLayoutNode<'dom> {
         ServoLayoutNode { node: n }
     }
 
+    /// Create a new [`ServoLayoutNode`] for this given [`TrustedNodeAddress`].
+    ///
+    /// # Safety
+    ///
+    /// The address pointed to by `address` should point to a valid node in memory.
     pub unsafe fn new(address: &TrustedNodeAddress) -> Self {
         ServoLayoutNode::from_layout_js(LayoutDom::from_trusted_node_address(*address))
     }
--- a/components/script/script_runtime.rs
+++ b/components/script/script_runtime.rs
@@ -448,6 +448,258 @@ pub struct Runtime {
 }
 
 impl Runtime {
+    /// Create a new runtime, optionally with the given [`NetworkingTaskSource`].
+    ///
+    /// # Safety
+    ///
+    /// If panicking does not abort the program, any threads with child runtimes will continue
+    /// executing after the thread with the parent runtime panics, but they will be in an
+    /// invalid and undefined state.
+    ///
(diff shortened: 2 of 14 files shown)


components/script/script_runtime.rs
```diff
@@ -447,6 +447,260 @@ pub struct Runtime {
...
+    ///
+    /// If panicking does not abort the program, any threads with child runtimes will continue
+    /// executing after the thread with the parent runtime panics, but they will be in an
+    /// invalid and undefined state.
+    ///
+    /// This, like many calls to SpiderMoney API, is unsafe.
+    #[allow(unsafe_code)]
+    pub(crate) fn new(networking_task_source: Option<NetworkingTaskSource>) -> Runtime {
+        unsafe { Self::new_with_parent(None, networking_task_source) }
+    }
```
sagudev, line 463: safety section on safe fn, should this fn be marked as unsafe?

sagudev (approved):

components/script/script_runtime.rs
mrobinson: Sometimes we put a safety section on functions that are marked safe when there are extra conditions, but do not require unsafe code blocks. I think it's just sort of a "nice to know" thing.

components/script/layout_dom/node.rs
```diff
@@ -76,6 +76,11 @@ impl<'dom> ServoLayoutNode<'dom> {
         ServoLayoutNode { node: n }
     }
 
+    /// Create a new [`ServoLayoutNode`] for this given [`TrustedNodeAddress`].
+    ///
+    /// # Safety
+    ///
+    /// The address pointed to by `address` should point to a valid node in memory.
```
mukilan, line 83: This seems redundant given that the documenation for `TrustedNodeAddress` [claims that it represents a valid Node address](https://github.com/servo/servo/blob/c430645606396a300b57a4d9c909db959486ca4c/components/shared/script_layout/lib.rs#L141). Perhaps this is still needed since the field in TrustedNodeAddress is public, but I wonder if that should be made private?

Also, are there no other conditions that need to be highlighted in this safety section? As an example (I'm not really familiar with this code), is there anything to be said about whether or not the DOM node is already rooted?

mrobinson: Unfortunately, I don't think we can make those kind of guarantees as `TrustedNodeAddress` is sent over IPC channels. This is all very bad, of course, and we should remove `TrustedNodeAddress` entirely, but that's a much larger task.

PR #33749, opened 2024-10-09
Use `--break-system-packages` when running pip in the WPT export job

mrobinson:
This seems to be necessary now, in much the same way it is necessary on
a normal Ubuntu system. I'm not sure what changed that this is now
required.

---

- [x] These changes do not require tests because this just fixes a CI job.

### DIFF
--- a/.github/workflows/pull-request-wpt-export.yml
+++ b/.github/workflows/pull-request-wpt-export.yml
@@ -29,7 +29,7 @@ jobs:
           # See https://github.com/actions/checkout/issues/162.
           token: ${{ secrets.WPT_SYNC_TOKEN }}
       - name: Install requirements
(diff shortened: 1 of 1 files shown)


mrobinson:
Looks like this has broken because `ubuntu-latest` is now 24.04.

mukilan (approved):

Loirooriol:
Would `pip install --user` work? Might be a better option.

mrobinson:
> Would `pip install --user` work? Might be a better option.

I think either way should be fine. Changes made to the build bot aren't preserved between jobs, but maybe there are other benefits that I'm not thinking of?

Loirooriol:
It's just that `--user` sounds less scary.

PR #33751, opened 2024-10-09
Upgrade to `egui@0.29.1` and `winit@0.30.5`

Gae24:
This is the minimum amount of work necessary to upgrade to `egui@0.29.1`. Updating both `glow` and `winit` were required. Various deprecated `winit` methods are still used; the switch to the trait based API will be done in a followup PR. `egui-glow` has added a boolean on the `Painter` constructor to enable dithering, which for now is set to `false`. `epaint` has a new dependency `epaint_default_fonts` for the `default_fonts` feature.

---

- [X] These changes do not require tests because they do not change functionality.


servo-tidy.toml
```diff
@@ -52,9 +52,11 @@ packages = [
     "foreign-types",
     "foreign-types-shared",
     "metal",
-    "glow",
     "windows-core",
 
+    # wgpu-hal depends on 0.5.0
```
mrobinson, line 57:
```suggestion
    # wgpu-hal depends on 0.5.0.
```

ports/servoshell/desktop/minibrowser.rs
```diff
@@ -78,7 +78,7 @@ fn truncate_with_ellipsis(input: &str, max_length: usize) -> String {
 impl Minibrowser {
     pub fn new(
         rendering_context: &RenderingContext,
-        events_loop: &EventsLoop,
+        event_loop: &winit::event_loop::EventLoop<WakerEvent>,
```
mrobinson, line 81: It's a bit odd to only used a qualified use for one of these types, so I would also do the same for `EventLoop`.

```suggestion
        event_loop: &EventLoop<WakerEvent>,
```

ports/servoshell/desktop/headed_window.rs
```diff
@@ -72,15 +72,15 @@ impl Window {
         // #9996.
         let visible = opts.output_file.is_none() && !no_native_titlebar;
 
-        let window_builder = winit::window::WindowBuilder::new()
+        let window_attr = winit::window::Window::default_attributes()
```
mrobinson, line 75:
```suggestion
        let window_attributes = winit::window::Window::default_attributes()
```

ports/servoshell/desktop/events_loop.rs
```diff
@@ -62,7 +63,7 @@ impl EventsLoop {
         Ok(EventsLoop(if headless {
             EventLoop::Headless(Arc::new((Mutex::new(false), Condvar::new())))
         } else {
-            let mut event_loop_builder = winit::event_loop::EventLoopBuilder::with_user_event();
+            let mut event_loop_builder = winit::event_loop::EventLoop::with_user_event();
```
mrobinson, line 66: Can you also do a use `winit::event_loop::EventLoop` above and replace all the of the qualified usage in this file? It's odd to mix them. I know that the existing code does this, but we should start cleaning it up.

```suggestion
            let mut event_loop_builder = EventLoop::with_user_event();
```

mrobinson (changes requested):
Thank you! Looks good with a couple minor nits.

ports/servoshell/desktop/events_loop.rs
Gae24: I guess it was done because there's already the `EventLoop` enum in scope

mrobinson (approved):

Issue #33752, opened 2024-10-09
WPT-Exports fails on ubuntu-24.04


sagudev:
ubuntu-latest now points to `ubuntu-24.04`, that has newer python, that hates pip install:
> Run pip install -r servo/python/requirements.txt
>error: externally-managed-environment
>
>× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
>    
>    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
>    
>    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
>    
>    See /usr/share/doc/python3.12/README.venv for more information.
>
> note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.

example run: https://github.com/servo/servo/actions/runs/11252836534/job/31286826796#step:5:9

mukilan:
@mrobinson has an alternative fix for this https://github.com/servo/servo/pull/33749.

I'm also working on switching the lint job to 22.04.

sagudev:
> I'm also working on switching the lint job to 22.04.

We should hardcode  jobs to specific ubuntu versions, so there will be no more surprises anymore.

mukilan:
All our non-trivial jobs (build, docs, wpt)  already use `ubuntu-22.04`. Only the "glue" jobs like WPT export, creating Github releases etc use the "ubuntu-latest". I guess it shouldn't be a big risk if we stick with "ubuntu-latest" for the glue jobs.

PR #33753, opened 2024-10-09
Fix WPT-Export by using ubuntu-22.04 that has older python

sagudev:
temporary fix for https://github.com/servo/servo/issues/33752

I did not test changes.

---

- [x] These changes do not require tests because they are just CI


sagudev:
Whoops, I just noticed https://github.com/servo/servo/pull/33749 that has better fix.

PR #33754, opened 2024-10-09
Use `content_box_sizes_and_padding_border_margin_deprecated()` in `FlexItem::new()`

### LINKED ISSUE #33736: Use `content_box_sizes_and_padding_border_margin_deprecated()` in `FlexItem::new()`
We can replace this: https://github.com/servo/servo/blob/c1b744b2b2c37cdb119a3d3f761822d22f938339/components/layout_2020/flexbox/layout.rs#L1026-L1038

with a single `content_box_sizes_and_padding_border_margin_deprecated()` call, like we are already doing in https://github.com/servo/servo/blob/c1b744b2b2c37cdb119a3d3f761822d22f938339/components/layout_2020/flexbox/layout.rs#L2021-L2022

cashall-0:
---

- [X] These changes fix #33736 

- [x] These changes do not require tests because there is no behavior change.

### DIFF
--- a/components/layout_2020/flexbox/layout.rs
+++ b/components/layout_2020/flexbox/layout.rs
@@ -1012,6 +1012,8 @@ fn allocate_free_cross_space_for_flex_line(
 impl<'a> FlexItem<'a> {
     fn new(flex_context: &FlexContext, box_: &'a mut FlexItemBox) -> Self {
         let containing_block = flex_context.containing_block;
+        let indefinite_containing_block: IndefiniteContainingBlock =
+            IndefiniteContainingBlock::from(containing_block);
         let parent_writing_mode = containing_block.style.writing_mode;
         let item_writing_mode = box_.style().writing_mode;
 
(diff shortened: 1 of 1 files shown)


components/layout_2020/flexbox/layout.rs
```diff
@@ -1012,6 +1012,8 @@ fn allocate_free_cross_space_for_flex_line(
 impl<'a> FlexItem<'a> {
     fn new(flex_context: &FlexContext, box_: &'a mut FlexItemBox) -> Self {
         let containing_block = flex_context.containing_block;
+        let indefinite_containing_block: IndefiniteContainingBlock =
```
Taym95, line 1015:
```suggestion
        let indefinite_containing_block =
```
is type needed?

components/layout_2020/flexbox/layout.rs
```diff
@@ -1024,18 +1026,9 @@ impl<'a> FlexItem<'a> {
...
-            .map(|v| v.map(Au::from));
-        let max_size = box_
-            .style()
-            .content_max_box_size_deprecated(containing_block, &pbm)
-            .map(|v| v.map(Au::from));
-        let min_size = box_
-            .style()
-            .content_min_box_size_deprecated(containing_block, &pbm)
-            .map(|v| v.map(Au::from));
+            .content_box_sizes_and_padding_border_margin_deprecated(&indefinite_containing_block);
```
Loirooriol, line 1031: I think this would suffice, without the new variable

```suggestion
            .content_box_sizes_and_padding_border_margin_deprecated(&containing_block.into());
```

Loirooriol (approved):

Loirooriol:
And please check "These changes do not require tests because ___" and say that there is no behavior change.

components/layout_2020/flexbox/layout.rs
cashall-0: okay

cashall-0: @Loirooriol pbm was arlready defined in scope and this new method returns pbm as part of the tuple. Should I clear the previous pbm variable?

Loirooriol: Yes, please clear it.

Loirooriol:
OK, so `/css/css-flexbox/flexbox-align-self-horiz-002.xhtml` is failing because `content_box_sizes_and_padding_border_margin` uses the writing mode from the style instead of the containing block.

That seems a bug, I had noticed it but though it wasn't a big deal because we don't support vertical writing modes, however I failed to consider `direction: rtl`.

Let's address this in #33773

PR #33308, opened 2024-09-04
CI: Force github hosted runners and remove concurrency on select-runner

sagudev:
temporary fix for #33276

Either this or complete revert of #33081 is needed or MQ will progress very slowly (no PR landed in last 20 hours).

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes do not require tests because CI changes

PR #33755, opened 2024-10-09
ci: pin the ubuntu runner version in lint job.

### LINKED ISSUE #33752: WPT-Exports fails on ubuntu-24.04

mukilan:
Github has updated the `ubuntu-latest` label to use 24.04, but that
breaks the lint job. Since the build job (github hosted and self hosted
runners) still explicitly use 22.04, pin the same version for lint job
until we upgrade all the workflows together.

Try run on my fork: https://github.com/mukilan/servo/actions/runs/11252639087/job/31286207004

---
- [x] These changes fix #33752
- [x] These changes do not require tests because they change only the CI code.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
         if: github.event_name != 'pull_request_target'


sagudev (approved):

sagudev:
We need this ASAP to land anything else.

mukilan:
yep, @mrobinson merged it manually and re-enabled the status check requirements.

mrobinson:
> That went to fast! Did anybody disable MQ?

We had to disable status checks to land this as status checks will always fail until this landed.

sagudev:
> > That went to fast! Did anybody disable MQ?
> 
> We had to disable status checks to land this as status checks will always fail until this landed.

I think this busted MQ:
image

It might be something like https://github.com/servo/servo/pull/33308#issuecomment-2327919183, can anybody with permissions check that required checks are properly configured.

mrobinson:
@sagudev Thank you! That was the problem. This is really annoying because whenever you add the DCO job as a dependency it defaults to the bad configuration.

PR #33757, opened 2024-10-09
Fixed some clippy warnings in components/script/dom

komutumu:
In this PR;
Enum-variation clippy warnings have been fixed in files  components/script/dom/eventtarget.rs and  components/script/dom/domexception.rs

- [X] These changes do not require tests because they do not touch functionality

### DIFF
--- a/components/script/dom/eventtarget.rs
+++ b/components/script/dom/eventtarget.rs
@@ -58,20 +58,21 @@ use crate::realms::{enter_realm, InRealm};
 use crate::script_runtime::CanGc;
 
 #[derive(Clone, JSTraceable, MallocSizeOf, PartialEq)]
+#[allow(clippy::enum_variant_names)]
 pub enum CommonEventHandler {
     EventHandler(#[ignore_malloc_size_of = "Rc"] Rc<EventHandlerNonNull>),
 
-    ErrorEvent(#[ignore_malloc_size_of = "Rc"] Rc<OnErrorEventHandlerNonNull>),
+    ErrorEventHandler(#[ignore_malloc_size_of = "Rc"] Rc<OnErrorEventHandlerNonNull>),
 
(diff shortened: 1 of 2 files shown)


jahielkomu:
Hello @mrobinson, greetings.
Please kindly review.
Thank you.

components/script/dom/eventtarget.rs
```diff
@@ -61,17 +61,17 @@ use crate::script_runtime::CanGc;
 pub enum CommonEventHandler {
     EventHandler(#[ignore_malloc_size_of = "Rc"] Rc<EventHandlerNonNull>),
 
-    ErrorEventHandler(#[ignore_malloc_size_of = "Rc"] Rc<OnErrorEventHandlerNonNull>),
+    ErrorEvent(#[ignore_malloc_size_of = "Rc"] Rc<OnErrorEventHandlerNonNull>),
```
mrobinson, line 64: Could you simply add the `allow` directive here as well and remove the other changes? It's a bit odd to remove `Handler` from all but one variant.

mukilan (approved):

components/script/dom/eventtarget.rs
mukilan: I think it is worth discussing the specific instances in Zulip and deciding on a case-by-case basis. There could be cases where renaming makes sense instead of  `allow`.

mukilan:

mukilan:
Looks like a timeout in the windows runner selection https://github.com/servo/servo/actions/runs/11267049027/job/31331534631

Adding this back to the queue.

components/script/dom/eventtarget.rs
jahielkomu: > I think it is worth discussing the specific instances in Zulip and deciding on a case-by-case basis. There could be cases where renaming makes sense instead of `allow`.

Okay, thank you @mukilan

PR #33762, opened 2024-10-09
clippy: Rename various methods and members to conform to naming guidelines

mrobinson:
This ensure that methods named `new()` do not take `&self` or return
`Box<Self>`. In addition, method are renamed (or removed when not
necessary) to avoid being prefixed with `from_`.

---

### DIFF
--- a/components/script/dom/workerglobalscope.rs
+++ b/components/script/dom/workerglobalscope.rs
@@ -115,15 +115,14 @@ pub struct WorkerGlobalScope {
 
     #[ignore_malloc_size_of = "Defined in ipc-channel"]
     #[no_trace]
-    /// Optional `IpcSender` for sending the `DevtoolScriptControlMsg`
-    /// to the server from within the worker
-    from_devtools_sender: Option<IpcSender<DevtoolScriptControlMsg>>,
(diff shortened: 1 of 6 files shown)


jdm (approved):

components/script/dom/workerglobalscope.rs
```diff
@@ -113,17 +113,11 @@ pub struct WorkerGlobalScope {
     location: MutNullableDom<WorkerLocation>,
     navigator: MutNullableDom<WorkerNavigator>,
 
-    #[ignore_malloc_size_of = "Defined in ipc-channel"]
+    #[ignore_malloc_size_of = "Defined in crossbeam"]
     #[no_trace]
-    /// Optional `IpcSender` for sending the `DevtoolScriptControlMsg`
-    /// to the server from within the worker
-    from_devtools_sender: Option<IpcSender<DevtoolScriptControlMsg>>,
```
jdm, line 120: I think removing this field broke a ton of tests.

mrobinson: Looks like this `IpcSender` must be stored to keep the channel alive. This was really unclear from the code, so I've left a comment. This should probably be clarified a bit with another data structure that stores all of the different channels necessary for devtools communication.

PR #33769, opened 2024-10-09
use `ThreadSafeJSContext` instead of `ContextForRequestInterrupt`

### LINKED ISSUE #33759: Create Safe JSContext wrappers
based on https://github.com/servo/servo/pull/33718#issuecomment-2402134951:

I think we need to have something like this:

```rust
pub struct SafeJSContext {
    cx: *mut JSContext,
    mt_cx: Arc<Mutex<Option<*mut JSContext>>>,
}

impl SafeJSContext {
    unsafe fn cx(&self) -> *mut JSContext {
        self.cx
    }

    fn thread_safe_cx(&self) -> ThreadSafeJSContext {
         ThreadSafeJSContext {
            cx: mt_cx,
         }
    }
}

// It safe to send jscontext, but only one thread can access most fields so it basically owns jscontext
// hence it does not impl clone
unsafe impl Send for SafeJSContext {}

impl Drop for SafeJSContext {
    fn drop(&mut self) {
         self.valid.lock().unwrap() = None;
         DestroyContext(self.cx)
    }
}

pub struct ThreadSafeJSContext {
    cx: Arc<Mutex<Option<*mut JSContext>>>,
}

unsafe impl Send for ThreadSafeJSContext {}
unsafe impl Sync for ThreadSafeJSContext {}

impl ThreadSafeJSContext {
    // ... impl only thread safe operations

    fn request_int_callback(&self) {
    if let Some(cx) = self.valid.lock().unwrap() {
        JS_RequestInterruptCallback(cx)
    }
}
```

sagudev:
Companion to https://github.com/servo/mozjs/pull/512

---

- [x] These changes fix #33759
- [x] There are tests for these changes in WPT?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4519,7 +4519,7 @@ dependencies = [
 [[package]]
 name = "mozjs"
 version = "0.14.1"
-source = "git+https://github.com/sagudev/mozjs?branch=mtcx#16bffcbe82c23ac4dd8ba45a68d53881423dfd6f"
+source = "git+https://github.com/servo/mozjs#a02aaf1e11fd275f2f129d0c7ca80a9d07460036"
 dependencies = [
  "bindgen",
  "cc",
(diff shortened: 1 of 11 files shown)


sagudev:
try run: https://github.com/sagudev/servo/actions/runs/11267419609

mrobinson (approved):

PR #33770, opened 2024-10-09
Update Servo user agent strings

atouchet:
---

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -1294,26 +1294,26 @@ fn default_user_agent_string_for(agent: UserAgent) -> String {
 
     #[cfg(all(target_os = "linux", target_arch = "x86_64", not(target_env = "ohos")))]
     let desktop_ua_string =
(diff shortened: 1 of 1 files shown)


sagudev (commented):
I think last update was done in SM bump (not by me). It would make sense to continue to follow this pattern and because we are following ESR versions, websites should be happy too.

atouchet:
No I did the last version update in #29563 and it tracked the latest Firefox release at the time. I do not believe we've ever bothered to align the user agent strings in Servo with Firefox ESR.

atouchet:
For whatever it's worth I'm not entirely opposed to matching Firefox ESR, I mostly just want to use a version that won't run into website compatibility problems. As far as I'm aware the rest of the various Firefox components used in Servo (ex. Stylo, WebRender) aren't kept in sync with ESR either.

atouchet:
I suppose that using a version number later than the SpiderMonkey version in use could cause potential compatibility issues. I have pushed a commit to switch to the latest ESR version 128.

sagudev (approved):

Issue #33742, opened 2024-10-08
Android: Native Compose Library


Doomsdayrs:
**Describe the new feature:**

Currently, The Android support directory contains an example application and a view library for Servo. 

The main problem I see is that not only are they written using Java, but also using the Android native view toolkits instead of Compose.

Guidance for all new Android applications is to be written in Kotlin and using Compose.

There are a few ways this can be accomplished as I see it.

1. Conversion of the android sub-directory completely to Kotlin/Compose.
2. Creation of a Kotlin/Compose sub-module that exists separately of the current code base.
3. Extraction of Servo native interop code into a separate sub-module and perform option two.

I plan in the coming weeks to slowly send patches at least converting the current gradle files to Kotlin in order to better facilitate a migration (or no migration, I will send anyways since groovy is a mess).

msub2:
Pinging @mukilan for their thoughts

mukilan:
Moving to Kotlin is a great idea! This is something we wanted to do at some point. I'm not too familiar with Compose, so I don't have a strong opinion on that.  I guess it should be fine as long as we can still distribute `ServoView` in a way that third-party applications can embed it in their own apps and we can potentially extend it into a [WebView provider](https://chromium.googlesource.com/chromium/src/+/HEAD/android_webview/docs/webview-providers.md) in the future.

wusyong:
There's [GeckoView](https://mozilla.github.io/geckoview/) we can reference from.

sagudev:
I always wanted to see engine-servo as mozilla android component: https://github.com/mozilla-mobile/firefox-android/tree/main/android-components/components/browser/engine-system.

Doomsdayrs:
I feel like discussion of Servo as an android component for those systems is out of scope of this issue, but a separate issue for it would be wise.

wusyong:
@Doomsdayrs I think it's somewhat related a bit. Could this library evolve into something like geckoview in the future? Or do we have to rewrite everything all over again.

Doomsdayrs:
> @Doomsdayrs I think it's somewhat related a bit. Could this library evolve into something like geckoview in the future? Or do we have to rewrite everything all over again.

[...]

PR #33772, opened 2024-10-10
Conversion to Gradle KTS

Doomsdayrs:
Convert Android Groovy Script to Kotlin Script.

As to possibly facilitate #33742 and or to simply match current Android Standards.

These are history destructive commits, as git will not recognize such heavy modifications as file moves.

All Logic is preserved per each commit sequentially. This means the first commit to the last preserves logic as to allow the first commit to be merged without the following ones, but not as to allow the following commits to be merged without the prior commit. 

---

- [x] `./mach build --debug --target x86_64-linux-android` does not report any errors


support/android/apk/servoview/build.gradle.kts
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    var result = ""
+
+    ResourceGroovyMethods.eachFileRecurse(path, FileType.FILES, closureOf<File> {
+        if (this.name == filename) {
+            result = this.absolutePath
+        }
+    })
```
Doomsdayrs, line 254: This does not work, trying to find a replacement.

Doomsdayrs: Maybe it did work, the replacement is also failing to work though..

Doomsdayrs: Figured it out, `if (path.isNotBlank())` was missing.

Doomsdayrs:
Alright, with that all the build code is now in Kotlin Script.

Next steps, Kotlinifying the various functions to no longer require Groovy Closures.

Doomsdayrs:
I personally feel that I cannot improve this any further currently.

Doomsdayrs:
Just found out that github butchered verifications when rebasing.

When I get home I'll repush to fix the signing.

Doomsdayrs:
May someone enable the android CI for this PR?

mrobinson:
```
Daemon will be stopped at the end of the build 
> Task :buildSrc:checkKotlinGradlePluginConfigurationErrors
> Task :buildSrc:pluginDescriptors
> Task :buildSrc:processResources NO-SOURCE
> Task :buildSrc:compileKotlin
> Task :buildSrc:compileJava NO-SOURCE
> Task :buildSrc:compileGroovy NO-SOURCE
> Task :buildSrc:classes UP-TO-DATE

> Task :buildSrc:jar
:jar: No valid plugin descriptors were found in META-INF/gradle-plugins

> Configure project :servoapp
w: file:///home/runner/work/servo/servo/support/android/apk/servoapp/build.gradle.kts:153:69: 'capitalize(): String' is deprecated. Use replaceFirstChar instead.

WARNING: API 'splits.density' is obsolete.
FAILURE: Build failed with an exception.

* Where:
Build file '/home/runner/work/servo/servo/support/android/apk/servoapp/build.gradle.kts' line: 55

* What went wrong:
SigningConfig with name 'release' not found.
```

Doomsdayrs:
Ah, I know the solution for that.

(using "named" instead of "register" in build script)

Doomsdayrs:
May someone try the android run again?

Doomsdayrs:
Ah, Google play requires API 33 or higher. A tomorrow task, for now I sleep.

Doomsdayrs:
Request for a test again.

Doomsdayrs:
Touched the sun with that.

Doomsdayrs:
Hopefully the final attempt.

Doomsdayrs:
😮‍💨 Alright, so it wasn't because of API 34 being the target.

the solution is to correct the notification code, which I don't want to do due to maintaining scope of this PR, but seems I have to.

Doomsdayrs:
I went out of scope and fixed lint issues as well. Hopefully now an android build will work! (it did on my PC before too, but :shrug: )

Doomsdayrs:
PRAISE BE!

mukilan:
Great work!

I was testing this change locally and it crashes on my device. It seems like the APK produced doesn't contain the native shared objects (libservoshell.so and libc++_shared.so). The [artifacts generated](https://github.com/servo/servo/actions/runs/11412144557) from the recent try run also confirm this - the APK is too small and doesn't have the 'lib' folder inside it.

I can investigate the issue later this week, but if you wanted to look into it, I suspect [this logic to copy the native objects](https://github.com/servo/servo/blob/4d11b2dc84d53b90ca5f2b729ffa1a5db2da45e9/support/android/apk/servoview/build.gradle#L144) is not being triggered as expected, perhaps due to a change in the name of the task we are looking to match using the regex.

Doomsdayrs:
> Great work!
> 
> I was testing this change locally and it crashes on my device. It seems like the APK produced doesn't contain the native shared objects (libservoshell.so and libc++_shared.so). The [artifacts generated](https://github.com/servo/servo/actions/runs/11412144557) from the recent try run also confirm this - the APK is too small and doesn't have the 'lib' folder inside it.
> 
> I can investigate the issue later this week, but if you wanted to look into it, I suspect [this logic to copy the native objects](https://github.com/servo/servo/blob/4d11b2dc84d53b90ca5f2b729ffa1a5db2da45e9/support/android/apk/servoview/build.gradle#L144) is not being triggered as expected, perhaps due to a change in the name of the task we are looking to match using the regex.

Something something `tasks.forEach` is occurring before the merge task is completed. Something something now have to figure out where it goes. I am investigating this right now.

Doomsdayrs:
Maestro, another build?

support/android/apk/buildSrc/src/main/kotlin/Interop.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.io.File
+import java.util.Locale
+import java.util.Properties
+
+fun Project.getTargetDir(debug: Boolean, arch: String): String {
+    val basePath = project.rootDir.parentFile.parentFile.parentFile.absolutePath
+    return basePath + "/target/android/" + getSubTargetDir(debug, arch)
+}
+
+fun Project.getNativeTargetDir(debug: Boolean, arch: String) {
```
Doomsdayrs, line 12: This does not return a string, thus the path ends up with a `Kotlin.Unit`

Doomsdayrs:
Perchance let the CI spin again?

Doomsdayrs:
It succeeded and also had a chunky artifact size, praise!

mukilan:
Thanks! I'll review this by the end of the week.

support/android/apk/buildSrc/src/main/kotlin/Interop.kt
```diff
@@ -0,0 +1,67 @@
...
+fun Project.getNativeTargetDir(debug: Boolean, arch: String): String {
+    val basePath = project.rootDir.parentFile.parentFile.parentFile.absolutePath
+    return basePath + "/target/" + getSubTargetDir(debug, arch)
+}
+
+fun getSubTargetDir(debug: Boolean, arch: String): String {
+    return getRustTarget(arch) + "/" + if (debug) "debug" else "release"
+}
+
+fun Project.getJniLibsPath(debug: Boolean, arch: String): String =
```
mukilan, line 21: Why are some functions defined on the `Project` but not others?

mukilan, line 21: Why are some functions defined on the Project but not others?

support/android/apk/buildSrc/src/main/kotlin/VersionCode.kt
```diff
@@ -0,0 +1,20 @@
...
+// XX is currently hardcoded to 00, but can be used to distinguish multiple builds within
+// the same day.
+//
+// TODO: check if this interferes with caching of local builds and add option to use
+// a static version.
+val today = Date()
+
+val versionCodeString: String = SimpleDateFormat("yyyyMMdd00").format(today)
+
+val generatedVersionCode = versionCodeString.toInt()
```
mukilan, line 20: Does Gradle guarantee that the logic in `buildSrc` will be run only once for both servoapp and servoview. My concern is that `servoapp` and `servoview` artifacts produced by a single build might not have the same version if that is not garaunteed.

support/android/apk/servoapp/src/main/AndroidManifest.xml
```diff
@@ -26,6 +27,7 @@
               <category android:name="android.intent.category.DEFAULT" />
               <category android:name="android.intent.category.BROWSABLE" />
               <data android:scheme="http" />
+              <data android:host="*"/>
```
mukilan, line 30: What is the motivation for this change? Was it not working previously?

support/android/apk/buildSrc/src/main/kotlin/Interop.kt
Doomsdayrs: Unlike in groovy in which scope is implicit and not clear, Kotlin requires explicit scoping.

`Project.` means this function is an extension to `Project` classes, thus has access to public fields of the `Project` class, and can only be called from a project object or scope.

I shall add documentation to the top of the file in this regard.

support/android/apk/buildSrc/src/main/kotlin/VersionCode.kt
Doomsdayrs: From what I understand from experience and [documentation](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html), and [member](https://discuss.gradle.org/t/buildsrc-vs-build-logic/46708/2) of the gradle team, Yes it is only produced once.

support/android/apk/servoapp/src/main/AndroidManifest.xml
Doomsdayrs: It was a part of a series of multiple commits addressing android lint errors, which in turn caused the CI to fail.

You may experiment on a local machine by dropping the commits and running the android lint task from gradle.

support/android/apk/buildSrc/src/main/kotlin/Interop.kt
Doomsdayrs: [786edc0](https://github.com/servo/servo/pull/33772/commits/786edc07ad24676396fc98815c44e7dc70d8c879)

support/android/apk/servoapp/src/main/AndroidManifest.xml
Doomsdayrs: Checking in, Have you done the experiment yet?

support/android/apk/buildSrc/src/main/kotlin/VersionCode.kt
mukilan: Thanks for the links! It does look like Gradle guarantees that the `buildSrc` is **built once** but I am not sure if that also implies if the built code **executes** once, globally across all sub-projects. In any case, we should be safe for now as the version code only has a granularity of a day, but this might be something we need to keep in mind if we ever extend it to include minutes, seconds or some other transient state.

support/android/apk/servoapp/src/main/AndroidManifest.xml
mukilan: Sorry about the delay. I did try removing `host` locally and it didn't fail the build for me. Do you have links to the logs from CI where it failed?

support/android/apk/buildSrc/src/main/kotlin/VersionCode.kt
Doomsdayrs: because the version code is assigned to a val, we can assume it compiles into a constant int and said class is just imported by other projects.

I'll resolve this for now.

support/android/apk/servoapp/src/main/AndroidManifest.xml
mukilan: I did a [test run on the CI ](https://github.com/mukilan/servo/actions/runs/11625426495) without the host directives and it succeeded.  I'll update this PR to remove those. We can always add it back if we find it is needed.

mukilan:
Great work! Thank you for working through all the issue with the migration!

mukilan (approved):

PR #33774, opened 2024-10-10
More CanGc fixes: Range, Event, gpu error, Header

webbeef:
Apply the CanGc propagation in a few more places.

---

- [ ] These changes do not require tests because there is no behavior change


jdm (approved):
Thanks! I appreciate this effort!

Issue #33756, opened 2024-10-09
Document `embedder_traits::resources::Resource`


wusyong:
We should document this enumeration and its variants because Servo simply won't work if these resource files are not provided. Right now, most people just copy the whole resource directory to bootstrap Servo. Let's start adding the documentation and explain what they do and if they are optional:

- [ ] [Resource::](https://doc.servo.org/embedder_traits/resources/enum.Resource.html)
  - [x] Preferences
  - [ ] BluetoothBlocklist
  - [ ] DomainList
  - [x] HstsPreloadList
  - [ ] BadCertHTML
  - [ ] NetErrorHTML
  - [ ] UserAgentCSS
  - [ ] ServoCSS
  - [ ] PresentationalHintsCSS
  - [ ] QuirksModeCSS
  - [x] RippyPNG
  - [ ] MediaControlsCSS
  - [ ] MediaControlsJS
  - [ ] CrashHTML
  - [ ] DirectoryListingHTML

PR #33775, opened 2024-10-10
doc(resources): `Preferences`, `HstsPreloadList`, `RippyPNG`

wusyong:
Add documentation to `Preferences`, `HstsPreloadList`, `RippyPNG` variants of `Resource` enum.
This PR also allow `Preferences` and `HstsPreloadList` to be default instead of panic.

part of #33756

---

### DIFF
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -9,7 +9,7 @@ use std::sync::LazyLock;
 
 use embedder_traits::resources::{self, Resource};
 use gen::Prefs;
-use log::warn;
+use log::{error, warn};
 use serde_json::{self, Value};
 
 use crate::pref_util::Preferences;
@@ -18,7 +18,7 @@ pub use crate::pref_util::{PrefError, PrefValue};
 static PREFS: LazyLock<Preferences<'static, Prefs>> = LazyLock::new(|| {
     let def_prefs: Prefs = serde_json::from_str(&resources::read_string(Resource::Preferences))
         .unwrap_or_else(|_| {
-            warn!("Preference json file is invalid. Setting Preference to default values");
+            error!("Preference json file is invalid. Setting Preference to default values");
             Prefs::default()
         });
     let result = Preferences::new(def_prefs, &gen::PREF_ACCESSORS);
--- a/components/config/prefs.rs
+++ b/components/config/prefs.rs
@@ -16,8 +16,11 @@ use crate::pref_util::Preferences;
 pub use crate::pref_util::{PrefError, PrefValue};
 
 static PREFS: LazyLock<Preferences<'static, Prefs>> = LazyLock::new(|| {
-    let def_prefs: Prefs =
-        serde_json::from_str(&resources::read_string(Resource::Preferences)).unwrap_or_default();
+    let def_prefs: Prefs = serde_json::from_str(&resources::read_string(Resource::Preferences))
+        .unwrap_or_else(|_| {
+            warn!("Preference json file is invalid. Setting Preference to default values");
+            Prefs::default()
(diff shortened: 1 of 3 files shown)


components/config/prefs.rs
```diff
@@ -16,8 +16,8 @@ use crate::pref_util::Preferences;
 pub use crate::pref_util::{PrefError, PrefValue};
 
 static PREFS: LazyLock<Preferences<'static, Prefs>> = LazyLock::new(|| {
-    let def_prefs: Prefs = serde_json::from_str(&resources::read_string(Resource::Preferences))
-        .expect("Failed to initialize config preferences.");
+    let def_prefs: Prefs =
+        serde_json::from_str(&resources::read_string(Resource::Preferences)).unwrap_or_default();
```
mukilan, line 20: Should we log the error somewhere so it doesn't cause confusion when the preference doesn't match what is expected?

wusyong: I feel like we should also set sensible defaults to `Preferences`. How about adding a warning log here and opening a follow up issue to set sensible defaults?

mukilan: I think this should be an 'error' log since the resource is present, but the failure can also be due to incorrect syntax. My preference would be to make the 'no prefs.json resource, so use defaults' case explicit in the embedder API without using an empty string to represent that. I don't have a good answer for what that API should look like though. Perhaps @mrobinson has some ideas.

wusyong: Let's open an issue to discuss then. Whether to keep the json and whether to set the default in `servo_config`.
I'll add an error log here instead.

components/net/hsts.rs
```diff
@@ -88,7 +88,7 @@ impl HstsList {
 
     pub fn from_servo_preload() -> HstsList {
         let list = resources::read_string(Resource::HstsPreloadList);
-        HstsList::from_preload(&list).expect("Servo HSTS preload file is invalid")
+        HstsList::from_preload(&list).unwrap_or_default()
```
mukilan, line 91: The error here probably needs to be logged as well.

mukilan (approved):

components/shared/embedder/resources.rs
```diff
@@ -55,16 +55,26 @@ pub fn sandbox_access_files_dirs() -> Vec<PathBuf> {
 }
 
 pub enum Resource {
+    /// A json file of [`Preferences`](servo_config::pref_util::Preferences) configuration.
+    /// It can be empty but lots of features will be disabled.
     Preferences,
     BluetoothBlocklist,
     DomainList,
+    /// A preload list of HTTP Strict Transport Security. It can be an empty list and
+    /// [`HstsList::default()`](net::hsts::HstsList) will be called.
```
mrobinson, line 64:
```suggestion
    /// A preloaded list of HTTP Strict Transport Security. It can be an empty list and
    /// [`HstsList::default()`](net::hsts::HstsList) will be called.
```

components/shared/embedder/resources.rs
```diff
@@ -55,16 +55,26 @@ pub fn sandbox_access_files_dirs() -> Vec<PathBuf> {
...
     HstsPreloadList,
     BadCertHTML,
     NetErrorHTML,
     UserAgentCSS,
     ServoCSS,
     PresentationalHintsCSS,
     QuirksModeCSS,
+    /// A place holder image to display if we couldn't get the requested image.
+    ///
```
mrobinson, line 73:
```suggestion
    /// A placeholder image to display if we couldn't get the requested image.
    ///
```

mrobinson (approved):

PR #33782, opened 2024-10-10
clippy: Refactor `script/dom/globalscope.rs` to avoid overly complex types

chickenleaf:
**File**: components/script/dom/globalscope.rs

**PR Description**:

**Issue**: Clippy gave complex types warning in globalscope.rs

**Change**:

1. Replaced `Result<Vec<u8>, Error>` with `Fallible<Vec<u8>>` in callback definitions since Fallible type alias is already defined in dom/bindings/error.rs.
2. Folded `FileListenerCallback` into `FileListenerTarget` .
3. Adjusted related enum variants and their constructors to match the new type definitions.

---

- [X] These changes are part of Fix as many clippy problems as possible #31500

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -613,7 +613,7 @@ impl MessageListener {
 }
 
 /// Callback used to enqueue file chunks to streams as part of FileListener.
-fn stream_handle_incoming(stream: &ReadableStream, bytes: Result<Vec<u8>, Error>) {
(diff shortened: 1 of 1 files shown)


mukilan (commented):
Nice work! I think we can also use the `FileListenerCallback` alias in the signature of the read_file_async method in this file.

mukilan (approved):
Thank you!

PR #33790, opened 2024-10-10
clippy: Fix `type_complexity` warnings in `components/script/dom`

taniishkaaa:
Fixes `type_complexity` warnings in `components/script/dom/gpudevice.rs` as a part of #31500 

---

- [X] These changes do not require tests because they only fix lint warnings

### DIFF
--- a/components/script/dom/gpucomputepipeline.rs
+++ b/components/script/dom/gpucomputepipeline.rs
@@ -76,7 +76,12 @@ impl GPUComputePipeline {
     ) -> WebGPUComputePipeline {
         let compute_pipeline_id = device.global().wgpu_id_hub().create_compute_pipeline_id();
 
-        let (layout, implicit_ids, _) = device.get_pipeline_layout_data(&descriptor.parent.layout);
+        let layout = device
+            .get_pipeline_layout_data(&descriptor.parent.layout)
+            .explicit();
+        let implicit_ids = device
+            .get_pipeline_layout_data(&descriptor.parent.layout)
+            .implicit();
 
         let desc = ComputePipelineDescriptor {
             label: (&descriptor.parent.parent).into(),
--- a/components/script/dom/gpucomputepipeline.rs
+++ b/components/script/dom/gpucomputepipeline.rs
@@ -76,16 +76,11 @@ impl GPUComputePipeline {
     ) -> WebGPUComputePipeline {
         let compute_pipeline_id = device.global().wgpu_id_hub().create_compute_pipeline_id();
 
-        let layout = device
-            .get_pipeline_layout_data(&descriptor.parent.layout)
-            .explicit();
-        let implicit_ids = device
-            .get_pipeline_layout_data(&descriptor.parent.layout)
-            .implicit();
+        let pipeline_layout = device.get_pipeline_layout_data(&descriptor.parent.layout);
 
         let desc = ComputePipelineDescriptor {
             label: (&descriptor.parent.parent).into(),
-            layout,
+            layout: pipeline_layout.explicit(),
             stage: (&descriptor.compute).into(),
             cache: None,
         };
@@ -97,7 +92,7 @@ impl GPUComputePipeline {
                 device_id: device.id().0,
                 compute_pipeline_id,
                 descriptor: desc,
-                implicit_ids,
+                implicit_ids: pipeline_layout.implicit(),
                 async_sender,
             })
             .expect("Failed to create WebGPU ComputePipeline");
--- a/components/script/dom/gpudevice.rs
+++ b/components/script/dom/gpudevice.rs
@@ -99,10 +99,10 @@ impl PipelineLayout {
         }
     }
 
-    pub fn implicit(&self) -> Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)> {
+    pub fn implicit(self) -> Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)> {
         match self {
             PipelineLayout::Implicit((layout_id, bind_group_layout_ids)) => {
-                Some((*layout_id, bind_group_layout_ids.clone()))
+                Some((layout_id, bind_group_layout_ids))
             },
             _ => None,
         }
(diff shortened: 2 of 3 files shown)


components/script/dom/gpucomputepipeline.rs
```diff
@@ -76,7 +76,12 @@ impl GPUComputePipeline {
     ) -> WebGPUComputePipeline {
         let compute_pipeline_id = device.global().wgpu_id_hub().create_compute_pipeline_id();
 
-        let (layout, implicit_ids, _) = device.get_pipeline_layout_data(&descriptor.parent.layout);
+        let layout = device
+            .get_pipeline_layout_data(&descriptor.parent.layout)
+            .explicit();
+        let implicit_ids = device
+            .get_pipeline_layout_data(&descriptor.parent.layout)
+            .implicit();
```
sagudev, line 84:
```suggestion
        let pipeline_layout = device.get_pipeline_layout_data(&descriptor.parent.layout);
```
and then do implicit, explicit method calls where this is actually needed.

components/script/dom/gpudevice.rs
```diff
@@ -86,6 +86,29 @@ pub struct GPUDevice {
...
+    }
+
+    pub fn implicit(&self) -> Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)> {
+        match self {
+            PipelineLayout::Implicit((layout_id, bind_group_layout_ids)) => {
+                Some((*layout_id, bind_group_layout_ids.clone()))
+            },
+            _ => None,
+        }
+    }
```
sagudev, line 109:
```suggestion
    pub fn implicit(self) -> Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)> {
        match self {
            PipelineLayout::Implicit((layout_id, bind_group_layout_ids)) => {
                Some((*layout_id, bind_group_layout_ids))
            },
            _ => None,
        }
    }
```

sagudev (changes requested):

components/script/dom/gpucomputepipeline.rs
sagudev: also same for render pipeline

taniishkaaa: okay, let me fix that

components/script/dom/gpudevice.rs
```diff
@@ -86,6 +86,29 @@ pub struct GPUDevice {
     valid: Cell<bool>,
 }
 
+pub enum PipelineLayout {
+    Implicit((PipelineLayoutId, Vec<BindGroupLayoutId>)),
```
mukilan, line 90:
```suggestion
    Implicit(PipelineLayoutId, Vec<BindGroupLayoutId>),
```

The `Implict` variant is already a tuple-like structure, so having a nested tuple is redundant.

components/script/dom/gpudevice.rs
```diff
@@ -349,7 +365,10 @@ impl GPUDevice {
             },
             multiview: None,
         };
-        Ok((implicit_ids, desc))
+        Ok((
+            PipelineLayout::Implicit(pipeline_layout.implicit().unwrap()),
```
mukilan, line 369: The `unwrap` here will panic whenever the pipeline_layout is `Explicit`, whereas previously it would have simply returned `None`. I think we just need to return the `pipeline_layout` from line 260.

components/script/dom/gpudevice.rs
```diff
@@ -478,7 +497,12 @@ impl GPUDeviceMethods for GPUDevice {
         descriptor: &GPURenderPipelineDescriptor,
     ) -> Fallible<DomRoot<GPURenderPipeline>> {
         let (implicit_ids, desc) = self.parse_render_pipeline(descriptor)?;
```
mukilan, line 499:
```suggestion
        let (pipeline_layout, desc) = self.parse_render_pipeline(descriptor)?;
```

components/script/dom/gpudevice.rs
```diff
@@ -478,7 +497,12 @@ impl GPUDeviceMethods for GPUDevice {
...
-        let render_pipeline = GPURenderPipeline::create(self, implicit_ids, desc, None)?;
+        let render_pipeline = GPURenderPipeline::create(
+            self,
+            PipelineLayout::Implicit(implicit_ids.implicit().unwrap()),
```
mukilan, line 502: The unwrap here will panic whenever the `implicit_id` is `Explicit` whereas previously we simply just pass `None` to `create`. I think we just need to pass in the `pipeline_layout` from `parse_render_pipeline`.

components/script/dom/gpurenderpipeline.rs
```diff
@@ -70,7 +69,7 @@ impl GPURenderPipeline {
     /// <https://gpuweb.github.io/gpuweb/#dom-gpudevice-createrenderpipeline>
     pub fn create(
         device: &GPUDevice,
-        implicit_ids: Option<(PipelineLayoutId, Vec<BindGroupLayoutId>)>,
+        implicit_ids: PipelineLayout,
```
mukilan, line 72:
```suggestion
        pipeline_layout: PipelineLayout,
```

mukilan (approved):

mukilan (changes requested):
Sorry, accidentally approved the PR instead of 'requesting changes', so doing it again.

taniishkaaa:
I've addressed all the changes in the new commit :)

components/script/dom/gpudevice.rs
```diff
@@ -230,22 +249,19 @@ impl GPUDevice {
                 bgls.push(webgpu::WebGPUBindGroupLayout(bgl));
```
mukilan, line 249: Seems like we no longer use the `bgls`  vector in line 254 - we push elements into it, but the vector is not returned. Can we simply remove it?

mukilan (approved):
I have added a comment about an unused variable, but otherwise looks good to me!

components/script/dom/gpudevice.rs
taniishkaaa: sure, can we also remove `let mut bgls = Vec::with_capacity(max_bind_grps as usize);` in line 245?

mukilan: Right, that is the one I was referring to, just mixed up the line number.

taniishkaaa: Done 👍

mukilan (approved):
Thank you!

sagudev (approved):

PR #33791, opened 2024-10-10
Throw NotSupported error and update expected results 

### LINKED ISSUE #33714: Throw NotSupported error instead of Type in XRSession::UpdateRenderState
Code: https://github.com/servo/servo/blob/ffccc5c88b8d47efc777b0bc9d2620bed4bc84d7/components/script/dom/xrsession.rs#L619-L622
Specification: Step 1 of https://immersive-web.github.io/layers/#updaterenderstatechanges
Test: https://github.com/servo/servo/blob/main/tests/wpt/tests/webxr/layers/xrSession_updateRenderState.https.html
Expected results: https://github.com/servo/servo/blob/main/tests/wpt/meta/webxr/layers/xrSession_updateRenderState.https.html.ini

After changing the code, run `./mach test-wpt tests/wpt/tests/webxr/layers/xrSession_updateRenderState.https.html` to verify that the test now passes, then [update the expected results](https://book.servo.org/hacking/testing.html#updating-web-test-expectations).

chickenleaf:
**File**: servo/components/script/dom/xrsession.rs

**PR Description**

**Issue**: Update expected results for xrSession_updateRenderState.https.html after passing tests

**Changes**:
- changed the error type returned to a NotSupported error in XRSession::UpdateRenderState.
- tests pass with unexpected passes
- update the expected results using: 
           -  ` ./mach test-wpt tests/wpt/tests/webxr/layers/xrSession_updateRenderState.https.html --log-raw /tmp/servo.log`
           -  `./mach update-wpt /tmp/servo.log`

 

---

- [X] These changes fix #33714


components/script/dom/xrsession.rs
```diff
@@ -680,9 +680,7 @@ impl XRSessionMethods for XRSession {
             }
             // https://github.com/immersive-web/layers/issues/189
             if init.layers.is_some() {
```
Taym95, line 682:
```suggestion
      if self.has_layers_feature() || init.layers.is_some() {
          return Err(Error::NotSupported);
      }
```

components/script/dom/xrsession.rs
```diff
@@ -680,9 +680,7 @@ impl XRSessionMethods for XRSession {
             }
             // https://github.com/immersive-web/layers/issues/189
```
Taym95, line 681:
```suggestion
```
Github link can removed since issues is closed

components/script/dom/xrsession.rs
chickenleaf: ahh ! got it !! yes!

Taym95:
> i shall close this and push another PR

I think this PR was okey since passed the test.

components/script/dom/xrsession.rs
Taym95: It's better to make changes locally and apply them from here, because applying them directly often results in incorrect changes like this.

PR #33794, opened 2024-10-10
Use NaN-safe comparisons in DOMQuad::GetBounds

simonwuelker:
This fixes the behaviour of [`DOMQuad::GetBounds`](https://drafts.fxtf.org/geometry/#dom-domquad-getbounds) when some of the coordinates are `NaN`.

This results in 9 new WPT passes, mostly in `css/geometry/DOMQuad-nan.html.ini`

[try run](https://github.com/simonwuelker/servo/actions/runs/11282771048)

---


jdm (approved):
Nice!

PR #33795, opened 2024-10-10
crypto: Implement encrypt/decrypt for AES-CBC + JWK support

msub2:
This fleshes out more of the SubtleCrypto implementation by adding support for encrypt/decrypt using AES-CBC with raw or JWK input.

---


components/script/dom/subtlecrypto.rs
```diff
@@ -317,18 +623,95 @@ impl SubtleCrypto {
...
     /// <https://w3c.github.io/webcrypto/#aes-cbc-operations>
-    fn export_key_aes_cbc(&self, format: KeyFormat, key: &CryptoKey) -> Result<Vec<u8>, Error> {
+    fn import_key_aes_cbc(
+        &self,
+        format: KeyFormat,
+        data: &[u8],
+        extractable: bool,
+        usages: Vec<KeyUsage>,
+    ) -> Result<DomRoot<CryptoKey>, Error> {
+        if usages.iter().any(|usage| {
```
simonwuelker, line 633: It would be nice to add the spec steps as comments here (and elsewhere). Makes it much easier to update the code when the specification changes in the future. Something like
```rs
// Step 1. If usages contains an entry which is not one of "encrypt", "decrypt", 
// "wrapKey" or "unwrapKey", then throw a SyntaxError.
``` 
in this case.

components/script/dom/subtlecrypto.rs
```diff
@@ -115,6 +132,104 @@ impl SubtleCrypto {
 }
 
 impl SubtleCryptoMethods for SubtleCrypto {
+    /// <https://w3c.github.io/webcrypto/#SubtleCrypto-method-encrypt>
+    fn Encrypt(
```
Taym95, line 136: Encrypt and Decrypt share a lot of code. Can we create a function process_crypto_operation and call it from both Encrypt and Decrypt by passing an enum, something like this?

```
enum CryptoOperation {
    Encrypt,
    Decrypt,
}

self.process_crypto_operation(cx, algorithm, key, data, comp, CryptoOperation::Decrypt);
self.process_crypto_operation(cx, algorithm, key, data, comp, CryptoOperation::Encrypt);

```

msub2: I did consider this yeah, my only concern is the slight variations between task pair requirements (encrypt/decrypt vs importKey/exportKey) complicating making a shared function. I'll take a look today though, will see if I can figure something out

components/script/dom/subtlecrypto.rs
```diff
@@ -422,26 +441,82 @@ impl SubtleCrypto {
...
-                // TODO: Support jwk
-                Err(Error::NotSupported)
+                if !matches!(data.len() * 8, 128 | 192 | 256) {
+                    return Err(Error::Data);
+                }
+                let handle = match data.len() * 8 {
+                    128 => Handle::Aes128(data.to_vec()),
+                    192 => Handle::Aes192(data.to_vec()),
+                    256 => Handle::Aes256(data.to_vec()),
+                    _ => unreachable!(),
```
jdm: Same comment about merging this with the previous length check.

components/script/dom/subtlecrypto.rs
```diff
@@ -359,6 +500,86 @@ fn normalize_algorithm(
 }
 
 impl SubtleCrypto {
+    /// <https://w3c.github.io/webcrypto/#aes-cbc-operations>
+    fn encrypt_aes_cbc(
+        &self,
+        params: SubtleAesCbcParams,
+        key: &CryptoKey,
+        data: &[u8],
+    ) -> Result<*mut JSObject, Error> {
```
jdm, line 509: This method should take a MutHandleObject instead of returning a raw JSObject pointer.

components/script/dom/subtlecrypto.rs
```diff
@@ -359,6 +500,86 @@ fn normalize_algorithm(
...
+        Ok(array_buffer_ptr.get())
+    }
+
+    /// <https://w3c.github.io/webcrypto/#aes-cbc-operations>
+    fn decrypt_aes_cbc(
+        &self,
+        params: SubtleAesCbcParams,
+        key: &CryptoKey,
+        data: &[u8],
+    ) -> Result<*mut JSObject, Error> {
```
jdm, line 546: Same comment here about the return value.

jdm (approved):
I share Taym's note that there's a bunch of duplicates code right now. It's not clear to me if that's just because the implementation is incomplete at the moment or if there really is an opportunity to reduce duplication in the long term, so I'll leave that up to you.

msub2:
So I made the MutableHandleObject change and have been working on trying to pull out each of the middle blocks for the SubtleCrypto methods into a `perform_crypto_operation` function. Unfortunately it seems after moving the relevant code blocks around and adding the enum and its variants for the crypto operations, the code has only gotten longer and I'm not convinced it's any less verbose/repetitive than before. The only shared vars between all the operations are the task source w/ canceller, the trusted SubtleCrypto, and the trusted promise, which take 3 lines. The bulk of the logic for these functions is still within the task closures and I don't know that I can pull anything out of those. Attaching a diff of what I have so far, let me know if there's something I'm missing and/or if I should stick with this.
[Branch diff](https://github.com/user-attachments/files/17452010/patch.patch)

jdm:
That's fair. I agree that the consolidation doesn't seem like an improvement to clarity and it's not worth spending more time on it.

msub2:
Sounds good, will rebase on main and then this should be ready.

msub2:
Ah wait, one of the changes I made caused a regression in the WPT tests (I think it was an early check in one of the funcs that was removed). Going to restore that now

components/script/dom/subtlecrypto.rs
```diff
@@ -165,7 +166,14 @@ impl SubtleCryptoMethods for SubtleCrypto {
                         if !valid_usage || key_gen_params.name != key_alg {
                             Err(Error::InvalidAccess)
                         } else {
-                            subtle.encrypt_aes_cbc(key_gen_params, &key, &data)
+                            let cx = GlobalScope::get_cx();
+                            rooted!(in(*cx) let mut array_buffer_ptr = ptr::null_mut::<JSObject>());
+                            match subtle.encrypt_aes_cbc(
+                                key_gen_params, &key, &data, cx, array_buffer_ptr.handle_mut()
+                            ) {
+                                Ok(_) => Ok(array_buffer_ptr.get()),
```
jdm, line 174: It would be safer to pass around a handle to the rooted value, rather than the raw object pointer. There's a risk of the value being garbage collected when this occurs, and using a handle would make this a compile error.

The solution to this problem is to move the rooted! above the enclosing match statement so it outlives the handle that we will return from this branch.

components/script/dom/subtlecrypto.rs
```diff
@@ -214,7 +222,14 @@ impl SubtleCryptoMethods for SubtleCrypto {
                         if !valid_usage || key_gen_params.name != key_alg {
                             Err(Error::InvalidAccess)
                         } else {
-                            subtle.decrypt_aes_cbc(key_gen_params, &key, &data)
+                            let cx = GlobalScope::get_cx();
+                            rooted!(in(*cx) let mut array_buffer_ptr = ptr::null_mut::<JSObject>());
+                            match subtle.decrypt_aes_cbc(
+                                key_gen_params, &key, &data, cx, array_buffer_ptr.handle_mut()
+                            ) {
+                                Ok(_) => Ok(array_buffer_ptr.get()),
```
jdm, line 230: Same point as previous. Let's help the type system help us!

jdm (approved):

jdm:
Thanks for your patience!

PR #33800, opened 2024-10-11
Various CanGc fixes

webbeef:
A new set of CanGc propagation fixes.

---

### DIFF
--- a/components/script/dom/xrsession.rs
+++ b/components/script/dom/xrsession.rs
@@ -257,7 +257,7 @@ impl XRSession {
     //
     // This enables content that assumes all input sources are accompanied
     // by an inputsourceschange event to work properly. Without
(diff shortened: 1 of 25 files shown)


components/script/dom/xrsession.rs
```diff
@@ -256,7 +257,7 @@ impl XRSession {
     //
     // This enables content that assumes all input sources are accompanied
     // by an inputsourceschange event to work properly. Without
-    pub fn setup_initial_inputs(&self) {
+    pub fn setup_initial_inputs(&self, can_gc: CanGc) {
```
jdm, line 260: This one is only used inside of an async task, so let's remove it and use CanGc::note() instead.

jdm (approved):

PR #33801, opened 2024-10-11
ohos: publish nightly builds

mukilan:
Add OpenHarmony nightly builds so that we can make them available for download
at servo.org. As discussed on Zulip, only the OHOS variant for aarch64
architecture is published.

---

- [x] These changes do not require tests because they only add new CI jobs.

### DIFF
--- a/.github/workflows/ohos.yml
+++ b/.github/workflows/ohos.yml
@@ -6,6 +6,13 @@ on:
         required: false
         default: "release"
         type: string
+      upload:
+        required: false
+        default: false
+        type: boolean
+      github-release-id:
+        required: false
+        type: string
   workflow_dispatch:
     inputs:
       profile:
@@ -29,7 +36,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        arch: ['aarch64-unknown-linux-ohos', 'x86_64-unknown-linux-ohos']
+        target: ['aarch64-unknown-linux-ohos', 'x86_64-unknown-linux-ohos']
     steps:
       - uses: actions/checkout@v4
         if: github.event_name != 'pull_request_target'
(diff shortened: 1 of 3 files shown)


.github/workflows/ohos.yml
```diff
@@ -75,34 +82,49 @@ jobs:
...
+        if: ${{ inputs.upload && contains(matrix.target, 'aarch64') }}
+        run: |
+          python3 ./mach upload-nightly ohos \
+            --secret-from-environment \
+            --github-release-id ${{ inputs.github-release-id }}
+        env:
+          S3_UPLOAD_CREDENTIALS: ${{ secrets.S3_UPLOAD_CREDENTIALS }}
+          NIGHTLY_REPO_TOKEN: ${{ secrets.NIGHTLY_REPO_TOKEN }}
+          NIGHTLY_REPO: ${{ github.repository_owner }}/servo-nightly-builds
+      - name: Generate artifact attestation for APK
```
jschwe, line 114:
```suggestion
      - name: Generate artifact attestation for HAP
```

mukilan: Fixed in ba555ff.

.github/workflows/ohos.yml
```diff
@@ -53,7 +60,7 @@ jobs:
       - name: Setup OpenHarmony SDK
         id: setup_sdk
         # FIXME: We pin the exact patch version to the LKG version of the
-        # action to address bugs in the most recent 0.1.2 release. Once that
+        # action to address bugs in the most recent 0.1.3 release. Once that
         # has been addresssed upstream, we can revert to specifying only the
         # major.minor tag.
         uses: openharmony-rs/setup-ohos-sdk@v0.1.2
```
mrobinson, line 66: I'm curious about this change, because 0.1.2 is still indicated below.

mrobinson (approved):

.github/workflows/ohos.yml
mukilan: Sorry for the confusion. The original comment I had added was incorrect - 0.1.2 is the last known good version and 0.1.3 is the buggy one, while the comment claimed it was 0.1.2 that has the bug. The change in this PR is just correcting the comment to reflect that.

PR #33802, opened 2024-10-11
Unify logic for laying out floats and atomic inlines

Loirooriol:
---

### DIFF
--- a/components/layout_2020/flow/inline/mod.rs
+++ b/components/layout_2020/flow/inline/mod.rs
@@ -117,17 +117,18 @@ use crate::context::LayoutContext;
 use crate::flow::float::{FloatBox, SequentialLayoutState};
 use crate::flow::{CollapsibleWithParentStartMargin, FlowLayout};
 use crate::formatting_contexts::{
-    Baselines, IndependentFormattingContext, NonReplacedFormattingContextContents,
+    Baselines, IndependentFormattingContext, IndependentLayoutResult,
+    NonReplacedFormattingContextContents,
 };
 use crate::fragment_tree::{
     BoxFragment, CollapsedBlockMargins, CollapsedMargin, Fragment, FragmentFlags,
     PositioningFragment,
 };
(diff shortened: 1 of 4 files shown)


components/layout_2020/flow/inline/mod.rs
```diff
@@ -2039,58 +2028,85 @@ impl IndependentFormattingContext {
...
+            return fragment;
+        }
+
+        // Offset the content rectangle by the physical offset of the padding, border, and margin.
+        let pbm_physical_offset = pbm_sums
+            .start_offset()
+            .to_physical_size(container_writing_mode);
+        fragment.content_rect = fragment
+            .content_rect
+            .translate(pbm_physical_offset.to_vector());
```
mrobinson, line 2071: This feels like it belongs a bit closer to inline layout, could you return `pbm_sums` from this function and have it take place in inline layout?

components/layout_2020/flow/inline/mod.rs
```diff
@@ -2039,58 +2028,85 @@ impl IndependentFormattingContext {
...
+
         // Apply baselines if necessary.
-        let mut fragment = match baselines {
+        match baselines {
             Some(baselines) => fragment.with_baselines(baselines),
             None => fragment,
-        };
+        }
```
mrobinson, line 2077: You could also return `baselines` and inline layout take care of that. Perhaps just return a data structure called `IndependentFormattingContextLayout`?

mrobinson (approved):
Looks good. The only other thing than what I've mentioned below is that this should probably move to a different file such as `flow/mod.rs` or `flow/block.rs`. Not sure.

components/layout_2020/flow/inline/mod.rs
Loirooriol: Went with `IndependentLayoutResult`, since it's similar to `IndependentLayout`, but the result of that.

Loirooriol:
Moved the code to `flow/mod.rs` (there is no `flow/block.rs`)

PR #33804, opened 2024-10-11
content: Replace uses of `downcast`+`is_some`/`is_none` with `is`

### LINKED ISSUE #33797: Replace uses of downcast+is_some/is_none with is
Code like this is unnecessary:
https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/dom/dommatrix.rs#L79
We're performing a downcast operation, then throwing away the actual object and just checking if it succeeded or not. We should replace code like that with the [is](https://doc.servo.org/script/dom/bindings/inheritance/trait.Castable.html#method.is) API instead: https://github.com/servo/servo/blob/b178a2b6e2312a74f4a72781789c45a31ac164a7/components/script/dom/customelementregistry.rs#L746

Code to fix:
* https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/dom/dommatrix.rs#L79
* https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/dom/dommatrixreadonly.rs#L447
* https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/fetch.rs#L184
* https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/dom/globalscope.rs#L2929
* https://github.com/servo/servo/blob/3d4996402f86e4abbbaa0a390114a92999d147be/components/script/dom/htmlformelement.rs#L1107

No need to run any tests. If it compiles successfully, that's enough for a PR.

taniishkaaa:
Replaces uses of `downcast`+`is_some`/`is_none` with `is` in `components/script` and `components/script/dom`

---

- [X] These changes fix #33797 

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -2909,7 +2909,7 @@ impl GlobalScope {
     /// Returns a boolean indicating whether the event-loop
     /// where this global is running on can continue running JS.
     pub fn can_continue_running(&self) -> bool {
-        if self.downcast::<Window>().is_some() {
+        if !self.is::<Window>() {
             return ScriptThread::can_continue_running();
         }
         if let Some(worker) = self.downcast::<WorkerGlobalScope>() {
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -2909,7 +2909,7 @@ impl GlobalScope {
     /// Returns a boolean indicating whether the event-loop
     /// where this global is running on can continue running JS.
     pub fn can_continue_running(&self) -> bool {
-        if !self.is::<Window>() {
+        if self.is::<Window>() {
             return ScriptThread::can_continue_running();
         }
         if let Some(worker) = self.downcast::<WorkerGlobalScope>() {
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -1102,7 +1102,7 @@ impl HTMLFormElement {
             let child = child.upcast::<Node>();
 
             // Step 5.1: The field element has a datalist element ancestor.
-            if child.ancestors().any(|a| !a.is::<HTMLDataListElement>()) {
+            if child.ancestors().any(|a| a.is::<HTMLDataListElement>()) {
                 continue;
             }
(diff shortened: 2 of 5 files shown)


components/script/dom/globalscope.rs
```diff
@@ -2909,7 +2909,7 @@ impl GlobalScope {
     /// Returns a boolean indicating whether the event-loop
     /// where this global is running on can continue running JS.
     pub fn can_continue_running(&self) -> bool {
-        if self.downcast::<Window>().is_some() {
+        if !self.is::<Window>() {
```
mrobinson, line 2912: Does this reverse the logic of the original test? Previously it was true if `self` was a `Window`, but now it is false.

components/script/fetch.rs
```diff
@@ -181,7 +181,7 @@ pub fn Fetch(
 
     // Step 6. If globalObject is a ServiceWorkerGlobalScope object, then set request’s
     //         service-workers mode to "none".
-    if global.downcast::<ServiceWorkerGlobalScope>().is_some() {
+    if !global.is::<ServiceWorkerGlobalScope>() {
```
mrobinson, line 184: I think this has the same issue.

components/script/dom/htmlformelement.rs
```diff
@@ -1102,10 +1102,7 @@ impl HTMLFormElement {
             let child = child.upcast::<Node>();
 
             // Step 5.1: The field element has a datalist element ancestor.
-            if child
-                .ancestors()
-                .any(|a| DomRoot::downcast::<HTMLDataListElement>(a).is_some())
-            {
+            if child.ancestors().any(|a| !a.is::<HTMLDataListElement>()) {
```
mrobinson, line 1105: I think this has the same issue.

mrobinson (dismissed):

components/script/dom/globalscope.rs
taniishkaaa: I can simply remove the `!`(not) so it returns the right logic

jdm (approved):

PR #33806, opened 2024-10-11
Avoid recomputing inline_content_sizes() when not needed

Loirooriol:
The result of `inline_content_sizes()` may depend on the block size of the containing block, so we were always recomputing in case we got a different block size.

However, if no content has a vertical percentage or stretches vertically, then we don't need to recompute: the result will be the same anyways.

---

- [X] These changes do not require tests because there shouldn't be any behavior change

### DIFF
--- a/components/layout_2020/flow/mod.rs
+++ b/components/layout_2020/flow/mod.rs
@@ -37,7 +37,7 @@ use crate::geom::{
 };
 use crate::positioned::{AbsolutelyPositionedBox, PositioningContext, PositioningContextLength};
 use crate::replaced::ReplacedContent;
-use crate::sizing::{self, ContentSizes};
+use crate::sizing::{self, ContentSizes, InlineContentSizesResult};
 use crate::style_ext::{Clamp, ComputedValuesExt, PaddingBorderMargin};
 use crate::{ContainingBlock, IndefiniteContainingBlock};
 
@@ -234,7 +234,7 @@ impl OutsideMarker {
             &IndefiniteContainingBlock::new_for_style(&self.marker_style),
         );
         let containing_block_for_children = ContainingBlock {
-            inline_size: content_sizes.max_content,
+            inline_size: content_sizes.sizes.max_content,
             block_size: AuOrAuto::auto(),
             style: &self.marker_style,
         };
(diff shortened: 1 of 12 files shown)


components/layout_2020/flow/mod.rs
```diff
@@ -360,59 +360,78 @@ fn calculate_inline_content_size_for_block_level_boxes(
...
-                    .max(ContentSizes::zero());
+                let InlineContentSizesResult {
+                    depends_on_block_constraints,
+                    sizes,
+                } = float_box.contents.outer_inline_content_sizes(
+                    layout_context,
+                    containing_block,
+                    &LogicalVec2::zero(),
+                    false, /* auto_block_size_stretches_to_containing_block */
+                );
```
mrobinson, line 377: Maybe it'd be a little bit less code to return `InlineContentSizesResult` in the tuple here. It could be made mutable and the sizes could be adjusted with `max_assign(...)`. It looks like currently `sizes.max(ContentSizes::zero()),` is executed for every branch of the match that produces a value?

components/layout_2020/flow/mod.rs
```diff
@@ -446,7 +465,20 @@ fn calculate_inline_content_size_for_block_level_boxes(
...
+    impl Default for AccumulatedData {
+        fn default() -> Self {
+            Self {
+                max_size: ContentSizes::zero(),
+                left_floats: ContentSizes::zero(),
+                right_floats: ContentSizes::zero(),
+                depends_on_block_constraints: false,
+            }
+        }
+    }
```
mrobinson, line 477: I think if `ContentSizes` has a `Default` implementation you can just derive `Default` for `AccumulatedData`. If not perhaps you could add `Default` for `ContentSizes`.

components/layout_2020/formatting_contexts.rs
```diff
@@ -267,20 +267,22 @@ impl NonReplacedFormattingContext {
...
             containing_block_for_children.size.inline,
             AuOrAuto::Auto,
             "inline_content_sizes() got non-auto containing block inline-size",
         );
-        if let Some((previous_cb_block_size, result)) = self.content_sizes {
-            if previous_cb_block_size == containing_block_for_children.size.block {
+        if let Some((previous_cb_block_size, result)) = self.content_sizes_result {
+            if !result.depends_on_block_constraints ||
+                previous_cb_block_size == containing_block_for_children.size.block
+            {
```
mrobinson, line 279: This could possibly be a `matches!`?

components/layout_2020/style_ext.rs
```diff
@@ -249,6 +249,7 @@ pub(crate) trait ComputedValuesExt {
         LogicalVec2<Size<Au>>,
         LogicalVec2<Size<Au>>,
         PaddingBorderMargin,
+        bool, /* has_block_percentage */
```
mrobinson, line 252: clippy is really going to complain about this. Do you mind returning a new data structure. I'm not too pick about the name, but it would be nice to avoid the new clippy warning. Or perhaps this is being addressed in another PR?

components/layout_2020/style_ext.rs
```diff
@@ -513,25 +516,40 @@ impl ComputedValuesExt for ComputedValues {
...
         );
-        let box_size = self
-            .box_size(writing_mode)
-            .maybe_percentages_relative_to_basis(&containing_block_size);
+        let box_size = self.box_size(writing_mode);
+        let min_size = self.min_box_size(writing_mode);
+        let max_size = self.max_box_size(writing_mode);
+        let depends_on_block_constraints = |size: &Size<LengthPercentage>| {
+            match size {
+                // TODO: fit-content uses stretch??
```
mrobinson, line 524: I guess this comment could be expanded a bit?

mrobinson (changes requested):
Just a few things I noticed:

components/layout_2020/flow/mod.rs
Loirooriol: Done

mrobinson (approved):

Issue #30863, opened 2023-12-12
Modular JS/execution engine


gterzian:
Can we make our JS integration modular, and general? See attached document for a start of the discussion.
[Servo--modular JS engine.pdf](https://github.com/servo/servo/files/13643394/Servo--modular.JS.engine.pdf)

A checklist(of only one item for now):

- [ ] remove unsafe JS related code https://github.com/servo/servo/issues/30862

sagudev:
Just found this: https://github.com/servo/servo/pull/21371

gterzian:
> Just found this: https://github.com/servo/servo/pull/21371

Thanks for sharing, it's an interesting approach to making the script crate generic: it seems to make script generic over some it's internals, like the dom parser and tokenizer, but it does not seem to deal with how script integrates with externals such as Spidermonkey or Layout. At least that is what I make of it after a quick read. 

Like I mention in the PDF attached at the top, I think we first need to isolate how Spidermonkey is used, because obviously one cannot make script generic to all of the use of `js::jsapi` and `js::rust` that you can see all over script. But an end-goal would see script being generic over a code execution engine, in a similar way to how the linked to PR make script generic over the tokenizer and parser. 

It could be worth revisiting that PR, although it would be a separate effort from the one this issue is focused on.

gterzian:
This is an interesting example: https://github.com/chakra-core/ChakraCore/wiki/JavaScript-Runtime-%28JSRT%29-Overview

Looks a bit like the kind of API we'd have to end-up with...

darkdragon-001:
Would it be also possible to run a noscript build of Servo with an empty JS engine?

jokeyrhyme:
Just in case folks coming here haven't seen it, there's a blog post update related to this: https://servo.org/blog/2024/04/15/spidermonkey/

And here's the report it links to: https://github.com/servo/servo/wiki/Servo-and-SpiderMonkey-Report

gterzian:
> Would it be also possible to run a noscript build of Servo with an empty JS engine?

Not currently, but if we reach the stage of abstracting away the engine by way of a trait, that trait could be implemented by what could be "an empty JS engine".

PR #33811, opened 2024-10-11
Simplify structured cloning infrastructure and make DOMPointReadOnly serializable

jdm:
I started looking at what it would take to implement this missing serialization for various DOM interfaces, and the existing structured clone code does not make it easy. This PR is a series of refactorings to reduce the opportunity for errors when adding support for new interfaces, and adds serialization for DOMPointReadOnly as a demonstration of how straightforward it now is as a result.

There are a few reasons why the existing code was complex:
* lots of potential for duplicated code when extending it
* enums with multiple variants that need to be ignored, so exhaustive matches aren't possible
* an enum with Read/Write variants where only one variant was ever used for a read or write operation, leading to lots of matches with unreachable panics
* confusing control flow that stored data in temporary fields in StructuredDataHolder then pulled it out again right afterwards
* requiring data types needed by blobs in the general serializable trait interface

The refactoring attempt to address these by:
* creating an enum that only contains interfaces that are serializable, allowing exhaustive matching against it
* splitting StructuredDataHolder into two separate types for reading and writing to reduce unnecessary enum matching
* extracting interface-specific serialization logic out of the general-purpose serialization logic, hoisting it into a type-safe representation that is part of the serialization trait interface
* creating type-safe abstractions for serializing and deserializing all known serializable DOM interfaces, minimizing code changes that need to be made to the general-purpose logic when adding support for new interfaces

---

### DIFF
--- a/components/script/dom/bindings/serializable.rs
+++ b/components/script/dom/bindings/serializable.rs
@@ -9,7 +9,7 @@ use js::jsapi::{JSStructuredCloneReader, JS_ReadUint32Pair};
 
 use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::root::DomRoot;
-use crate::dom::bindings::structuredclone::{CloneableObject, StructuredDataHolder};
+use crate::dom::bindings::structuredclone::{CloneableObject, StructuredWriteDataHolder, StructuredReadDataHolder};
 use crate::dom::globalscope::GlobalScope;
 
 /// The key corresponding to the storage location
@@ -72,11 +72,11 @@ pub trait Serializable: DomObject + Sized {
     const TAG: CloneableObject;
 
     /// <https://html.spec.whatwg.org/multipage/#serialization-steps>
-    fn serialize(&self, sc_holder: &mut StructuredDataHolder) -> Result<Self::Data, ()>;
+    fn serialize(&self, sc_holder: &mut StructuredWriteDataHolder) -> Result<Self::Data, ()>;
     /// <https://html.spec.whatwg.org/multipage/#deserialization-steps>
     fn deserialize(
         owner: &GlobalScope,
-        sc_holder: &mut StructuredDataHolder,
+        sc_holder: &mut StructuredReadDataHolder,
         extra_data: Self::Data,
     ) -> Result<DomRoot<Self>, ()>;
 }
--- a/components/script/dom/bindings/serializable.rs
+++ b/components/script/dom/bindings/serializable.rs
@@ -5,41 +5,13 @@
 //! Trait representing the concept of [serializable objects]
 //! (<https://html.spec.whatwg.org/multipage/#serializable-objects>).
 
-use js::jsapi::{JSStructuredCloneReader, JS_ReadUint32Pair};
+use js::jsapi::JSStructuredCloneReader;
 
 use crate::dom::bindings::reflector::DomObject;
 use crate::dom::bindings::root::DomRoot;
 use crate::dom::bindings::structuredclone::{CloneableObject, StructuredWriteDataHolder, StructuredReadDataHolder};
 use crate::dom::globalscope::GlobalScope;
 
-/// The key corresponding to the storage location
-/// of a serialized platform object stored in a StructuredDataHolder.
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct StorageKey {
-    pub index: u32,
-    pub name_space: u32,
-}
-
-impl ToSerializeOperations for StorageKey {
-    fn to_serialize_operations(&self) -> Vec<SerializeOperation> {
-        vec![SerializeOperation::Uint32Pair(self.name_space, self.index)]
-    }
-}
-
-impl FromStructuredClone for StorageKey {
-    unsafe fn from_structured_clone(r: *mut JSStructuredCloneReader) -> StorageKey {
-        let mut key = StorageKey {
-            name_space: 0,
-            index: 0,
-        };
-        assert!(JS_ReadUint32Pair(
-            r,
-            &mut key.name_space,
-            &mut key.index,
-        ));
-        key
-     }
-}
 
 /// An interface to construct a stream of serialization operations that will be
 /// evaluated as part of serializing a DOM object.
--- a/components/script/dom/bindings/serializable.rs
+++ b/components/script/dom/bindings/serializable.rs
@@ -6,6 +6,7 @@
 //! (<https://html.spec.whatwg.org/multipage/#serializable-objects>).
 
 use crate::dom::bindings::reflector::DomObject;
+use crate::dom::bindings::root::DomRoot;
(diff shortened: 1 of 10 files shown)


jdm:
Bleah:
```

  | /home/runner/work/servo/servo/Cargo.lock:1: Rust dependency enum-iterator@2.1.0: Rejected license "0BSD - BSD Zero Clause License:". Run `cargo deny` for more details

  | /home/runner/work/servo/servo/Cargo.lock:1: Rust dependency enum-iterator-derive@1.4.0: Rejected license "0BSD - BSD Zero Clause License:". Run `cargo deny` for more details
```
I'll look for another crate I guess.

sagudev:
> Bleah:
> 
> ```
> 
>   | /home/runner/work/servo/servo/Cargo.lock:1: Rust dependency enum-iterator@2.1.0: Rejected license "0BSD - BSD Zero Clause License:". Run `cargo deny` for more details
> 
>   | /home/runner/work/servo/servo/Cargo.lock:1: Rust dependency enum-iterator-derive@1.4.0: Rejected license "0BSD - BSD Zero Clause License:". Run `cargo deny` for more details
> ```
> 
> I'll look for another crate I guess.

There is https://github.com/Peternator7/strum for alternative crate, but I think it should be ok to add 0BSD to [allowlist](https://github.com/servo/servo/blob/ded518dfd0a22d6e6b8bec8a59e8498f18124616/deny.toml#L28) per https://joinup.ec.europa.eu/licence/compatibility-check/0BSD/MPL-2.0

I do not recall where, but I know that somewhere we manually maintain list of all variants.

sagudev:
Could we use serde/bincode for this and then put all serializable objects in one enum that is also serializable (to avoid having manual tags)?

mrobinson:
I think that `strum` could also be useful for `IntoStaticStr` which would remove a lot of manually written "enum to string" code as well.

gterzian (commented):
So I don't like the proposed approach for the following reasons:

- It goes back(albeit in a nicer more structured way) to what I did initially for Blob(see [this](https://github.com/servo/servo/commit/6e8a85482c2068d4dbccb992954271f725570f91))
- By that I mean "doing the serialization by hand using SM APIs", and it broke down once I looked into a more complex object like messageport. 
- Instead I thought it was a much better idea to only use SM for a simple storage key(a `u64`), and do the rest with serde. 
- I think there is an actual conflation of concept with how the storage key is now used for blob versus how the coordinates are serialized. 
- I agree we can restructure and simplify the code, but I would like to keep the separation between "the weird JS stuff"(which is necessary for the SM callbacks) and "just serializing data with serde"(which should be the actual data we want to serialize). I think this is also important in the light of https://github.com/servo/servo/issues/30863.

components/script/dom/bindings/serializable.rs
```diff
@@ -5,27 +5,52 @@
...
+pub trait FromStructuredClone: Sized {
+    unsafe fn from_structured_clone(r: *mut JSStructuredCloneReader) -> Self;
+}
+
+/// Operations permitted as part of structured clones to serialize arbitrary values
+/// during serialization of DOM objects.
+pub enum SerializeOperation {
+    Uint32Pair(u32, u32),
+    Double(f64),
 }
```
mrobinson, line 37: This seems like a nice abstraction. The one concern I have is that it means that all data will have to be copied during serialization (apart from blobs, right?). Is this serialization performance sensitive. If it is, another alternative would be to expose matching APIs on the writer and reader like this:

```
impl Writer {
   fn write_uint32_pair(&self, a: u32, b:u32);
}

impl Reader {
   fn read_uint32_pair(&self) -> Result<(u32, u32), Err>
}
```

Then all unsafe code could be confined to the reader and the writer. Maybe there's a reason why this isn't feasible though.

mrobinson (commented):
I took a quick look at this and while I'm not very familiar with the serialization code this looks like a really nice improvement. I think we could use `strum` here instead of `enum-iterator` as it will also allow us to simplify even more code in Servo as well.

components/script/dom/blob.rs
```diff
@@ -114,13 +118,37 @@ impl Blob {
...
+            name_space: 0,
+            index: 0,
+        };
+        assert!(JS_ReadUint32Pair(r, &mut key.name_space, &mut key.index,));
+        key
+    }
+}
+
 impl Serializable for Blob {
+    type Data = StorageKey;
```
gterzian, line 146: By the way I just saw that there is an inconsistency in the current code between messageport and blob, and that we don't really need this key. We can just use the currently unused [data](https://github.com/servo/servo/blob/717baff17b651606e3e7bf53ba20dc00ae996a38/components/script/dom/bindings/structuredclone.rs#L119), just like [ports do](https://github.com/servo/servo/blob/717baff17b651606e3e7bf53ba20dc00ae996a38/components/script/dom/messageport.rs#L220)

sagudev:
> I'm curious about the performance characteristics of using `serde` serialization vs. `JS_Write` and `JS_Read` calls. As anyone done some benchmarking?

I would say it will be roughly the same (that depends on actual serializer/deserializer used, which will probably be bincode), but it will be much more safer and easier to maintain.

gterzian (commented):
My preferred alternative is at https://github.com/servo/servo/pull/33879

It is indeed very boilerplaty, but still simpler than defining the serialization by hand. and I think with a bit of refactoring the boilerplate aspect could be also improved.

jdm:
> My preferred alternative is at #33879
> 
> It is indeed very boilerplaty, but still simpler than defining the serialization by hand. and I think with a bit of refactoring the boilerplate aspect could be also improved.

I hear your concerns about hand-written serialization code. I think it would be interesting to look for a design that requires fewer changes than the existing model but still maintains the serde usage.

jdm:
After reviewing the existing code more thoroughly with the additional context, I think it's clearest to describe the model as a minimum of serialized data plus a serializable side table. It also has the interesting property that the side table is only serialized if it's sent somewhere else; if we perform a same-origin postMessage (for example) then I don't believe StructuredSerializedData is serialized at all. In contrast, this PR moves towards a model that pessimistically serializes all cloneable types, which is clearly less desirable.

gterzian:
> if we perform a same-origin postMessage

Currently all postmessage goes over IPC for simplicity, but yes this, and the serialization, could be avoided. 

A better example is the newly added [`StructuredClone`](https://github.com/servo/servo/blob/d0d02cd56cb646da1f2193e57561b78d5e4b6bdf/components/script/dom/window.rs#L1564), which in the current setup indeed avoids serde serialization. 

The main idea is that currently the "side table" data ends-up in [`StructuredSerializedData`](https://github.com/servo/servo/blob/d0d02cd56cb646da1f2193e57561b78d5e4b6bdf/components/shared/script/lib.rs#L925) which is serializable, but indeed not always actually serialized.

jdm:
Closing this, since these changes don't fit the overall vision for improving this code.

PR #33816, opened 2024-10-11
Replace CanGc::note() calls with arguments in `components/script`

taniishkaaa:
Replaces CanGc::note() calls with arguments in `components/script/fetch.rs` and `components/script/image_listener.rs` as a part of #33683 

---

### DIFF
--- a/components/script/image_listener.rs
+++ b/components/script/image_listener.rs
@@ -23,6 +23,7 @@ pub fn generate_cache_listener_for_element<
     T: ImageCacheListener + DerivedFrom<Node> + DomObject,
 >(
     elem: &T,
+    can_gc: CanGc,
 ) -> IpcSender<PendingImageResponse> {
     let trusted_node = Trusted::new(elem);
     let (responder_sender, responder_receiver) = ipc::channel().unwrap();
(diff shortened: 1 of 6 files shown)


components/script/image_listener.rs
```diff
@@ -43,7 +44,7 @@ pub fn generate_cache_listener_for_element<
                     let element = element.root();
                     // Ignore any image response for a previous request that has been discarded.
                     if generation == element.generation_id() {
-                        element.process_image_response(image, CanGc::note());
+                        element.process_image_response(image, can_gc);
```
jdm, line 47: Since this use of CanGc is within a task! macro, that means it's an async callback and we don't want to propagate it into the caller. Please revert this!

components/script/image_listener.rs
```diff
@@ -23,6 +23,7 @@ pub fn generate_cache_listener_for_element<
     T: ImageCacheListener + DerivedFrom<Node> + DomObject,
 >(
     elem: &T,
+    can_gc: CanGc
```
jdm, line 26: Since the only CanGc use is within the task macro, we can remove this argument.

jdm (changes requested):

components/script/image_listener.rs
taniishkaaa: I've pushed the updates

jdm (approved):

PR #33817, opened 2024-10-11
Fix headless mode on Windows

crbrz:
After #33660 headless mode stopped working on Windows. With these changes the messages are sent to `compositor` through `CompositorProxy` which wakes the event loop after a message is sent.

Linux and MacOS weren't fully affected because in headless mode the event loop processes messages every 5 milliseconds when not woken
https://github.com/servo/servo/blob/a86dcfc6e7e163ebc084d69fedc32a368ef7e202/ports/servoshell/desktop/events_loop.rs#L89-L138

---

- [X] There are tests for these changes but WPT is not run on Windows yet

[try run](https://github.com/crbrz/servo/actions/runs/11298073966/job/31427636494)

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -978,25 +978,25 @@ fn create_compositor_channel(
 
     let (compositor_ipc_sender, compositor_ipc_receiver) =
         ipc::channel().expect("ipc channel failure");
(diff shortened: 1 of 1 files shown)


jdm (approved):
Good catch. Thanks for fixing this!

PR #33821, opened 2024-10-12
clippy: Fix warnings in `components/script` & `components/script/dom`

taniishkaaa:
Fixes clippy warnings in `components/script` & `components/script/dom` as a part of #31500 

---

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -5306,7 +5306,6 @@ impl DocumentMethods for Document {
         url: USVString,
         target: DOMString,
         features: DOMString,
-        _can_gc: CanGc,
     ) -> Fallible<Option<DomRoot<WindowProxy>>> {
         self.browsing_context()
             .ok_or(Error::InvalidAccess)?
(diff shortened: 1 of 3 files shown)


components/script/dom/document.rs
taniishkaaa, line 1: The `can_gc` variable here is unused, so lmk if it'd be better to simply remove it?

components/script/dom/document.rs
```diff
@@ -5306,7 +5306,7 @@ impl DocumentMethods for Document {
         url: USVString,
         target: DOMString,
         features: DOMString,
-        can_gc: CanGc,
+        _can_gc: CanGc,
```
taniishkaaa, line 5309: The can_gc variable here is unused, so lmk if it'd be better to simply remove it?

jdm: Yep, please remove it!

jdm (approved):

components/script/dom/document.rs
taniishkaaa: I'm not quite sure why the build is failing even after removing it from `Bindings.conf`. Maybe we should just do: `_can_gc`?

jdm: You're right. It looks like a bug in the code generator, so using the _ prefix for now is fine.

taniishkaaa: Alright, thanks! :)

PR #33825, opened 2024-10-13
Use document encoding when parsing url in href setter

simonwuelker:
This implements the [`encoding parse a url`](https://html.spec.whatwg.org/multipage/#encoding-parsing-a-url) algorithm and uses it in `HtmlAnchorElement::set_url`. Previously we didn't specify an encoding (the default is utf-8 i believe), leading servo to fail many encoding tests.

[try run](https://github.com/simonwuelker/servo/actions/runs/11314203920/job/31464086763)

---


components/script/dom/htmlanchorelement.rs
```diff
@@ -76,15 +76,30 @@ impl HTMLAnchorElement {
...
+
+        let Some(attribute) = attribute else {
+            return;
+        };
+
+        let document = document_from_node(self);
+
+        // Step 3. Let url be the result of encoding-parsing a URL given this element's
+        // href content attribute's value, relative to this element's node document.
+        let url = document.encoding_parse_a_url(&attribute.value());
```
sagudev, line 97: nit: I think `let document = document_from_node(self);` would go under step 3, as it does `this element's node document`.

sagudev (approved):

components/script/dom/htmlanchorelement.rs
simonwuelker: right, that makes more sense

PR #33828, opened 2024-10-13
CanGc fixes originating from `gamepadevent.rs` & `stereopannernode.rs`

taniishkaaa:
Replaces CanGc::note() calls with arguments passed by callers in `components/script/dom/gamepadevent.rs` & `components/script/dom/stereopannernode.rs`

---

- [X] These changes fix are a part of #33683 

### DIFF
--- a/components/script/dom/globalscope.rs
+++ b/components/script/dom/globalscope.rs
@@ -3204,10 +3204,10 @@ impl GlobalScope {
                 );
             },
             GamepadEvent::Disconnected(index) => {
-                self.handle_gamepad_disconnect(index.0);
+                self.handle_gamepad_disconnect(index.0, can_gc);
             },
             GamepadEvent::Updated(index, update_type) => {
-                self.receive_new_gamepad_button_or_axis(index.0, update_type);
+                self.receive_new_gamepad_button_or_axis(index.0, update_type, can_gc);
             },
         };
     }
@@ -3247,7 +3247,7 @@ impl GlobalScope {
                             false,
                             can_gc,
                         );
-                        navigator.set_gamepad(selected_index as usize, &gamepad);
+                        navigator.set_gamepad(selected_index as usize, &gamepad, can_gc);
                     }
                 }),
                 &self.task_canceller(TaskSourceName::Gamepad),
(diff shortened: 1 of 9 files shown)


components/script/dom/globalscope.rs
```diff
@@ -3266,7 +3266,7 @@ impl GlobalScope {
                         let navigator = window.Navigator();
                         if let Some(gamepad) = navigator.get_gamepad(index) {
                             if window.Document().is_fully_active() {
-                                gamepad.update_connected(false, gamepad.exposed());
+                                gamepad.update_connected(false, gamepad.exposed(), can_gc);
```
jdm, line 3269: Since this is only used within a task! macro, we don't actually want to.propagarethis as an argument to handle_gamepad_disconnect.

components/script/dom/globalscope.rs
```diff
@@ -3247,7 +3247,7 @@ impl GlobalScope {
                             false,
                             can_gc,
                         );
-                        navigator.set_gamepad(selected_index as usize, &gamepad);
+                        navigator.set_gamepad(selected_index as usize, &gamepad, can_gc);
```
jdm, line 3250: This inside of a task! macro, so we don't actually want to make this change. It looks like the earlier use of can_gcis incorrect as well, so I think we want to remove the argument to handle_gamepad_connect instead.

components/script/dom/globalscope.rs
```diff
@@ -3312,7 +3317,7 @@ impl GlobalScope {
                                             window.task_manager().gamepad_task_source().queue_with_canceller(
                                                 task!(update_gamepad_connect: move || {
                                                     let gamepad = new_gamepad.root();
-                                                    gamepad.notify_event(GamepadEventType::Connected);
+                                                    gamepad.notify_event(GamepadEventType::Connected, can_gc);
```
jdm, line 3320: Let's revert this, since it's inside of a task! macro.

jdm (changes requested):

components/script/dom/globalscope.rs
taniishkaaa: sorry for the overlook, let me fix this

jdm (approved):
Thanks!

PR #33833, opened 2024-10-14
README: Add python3-setuptools to ubuntu apt install line

sagudev:
This was needed on my clean installed ubuntu 24.10.

---

- [x] These changes do not require tests because it's only docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ For more detailed build instructions, see the Servo book under [Setting up your 
 
 - Install `curl` and `python`:
   - Arch: `sudo pacman -S --needed curl python python-pip`
-  - Debian, Ubuntu: `sudo apt install curl python3-pip python3-venv`
+  - Debian, Ubuntu: `sudo apt install curl python3-pip python3-venv python3-setuptools`
   - Fedora: `sudo dnf install curl python3 python3-pip python3-devel`
   - Gentoo: `sudo emerge net-misc/curl dev-python/pip`
 - Install `rustup`: `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`


mrobinson:
Do you happen to have a copy of the error? Perhaps this is needed in `requirements.txt` instead.

sagudev:
> Do you happen to have a copy of the error? Perhaps this is needed in `requirements.txt` instead.

Unfortunately not, but given that we also install pip and venv this way I think it is right way to do as setuptools is also about installing py packages.

mrobinson:
> Unfortunately not, but given that we also install pip and venv this way I think it is right way to do as setuptools is also about installing py packages.

It's not a direct dependency of our installation process though and typically indirect dependencies are installed via `requirements.txt` AFAIK.

sagudev:
error:
```
Collecting blessings==1.7 (from -r /servo/python/requirements.txt (line 4))
  Downloading blessings-1.7-py3-none-any.whl.metadata (19 kB)
Collecting distro==1.4 (from -r /servo/python/requirements.txt (line 5))
  Downloading distro-1.4.0-py2.py3-none-any.whl.metadata (6.6 kB)
Collecting mozinfo==1.2.3 (from -r /servo/python/requirements.txt (line 6))
  Downloading mozinfo-1.2.3-py2.py3-none-any.whl.metadata (649 bytes)
Collecting mozlog==8.0.0 (from -r /servo/python/requirements.txt (line 7))
  Downloading mozlog-8.0.0.tar.gz (53 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-s9ipom8b/mozlog_9174927e8a1140948d3b8243ff162e6d/setup.py", line 5, in <module>
          from setuptools import find_packages, setup
      ImportError: cannot import name 'find_packages' from 'setuptools' (unknown location)
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

```

sagudev:
I think the problem is that mozlog expect setuptools to be already installed in its installer, but it's not installed yet (we install it by same invocation of `pip install -r`).

BTW moving setuptools in requirements.txt to be first does not help.

mrobinson:
I guess this is a decent workaround until we can get more information about what is going on here.  It's pretty likely that this should ultimately be something that `./mach bootstrap` does.

mrobinson (approved):

PR #33840, opened 2024-10-14
Fix Clippy warning: remove unneeded late initialization of string

chickenleaf:
**File**: `components/script/dom/dommatrixreadonly.rs`

**PR Description**
**Issue:** 
Clippy warning for unnecessary late initialization (`needless_late_init`).

**Changes:** 
Moved the declaration of `string` to directly initialize it based on the 2D condition, removing the need for a separate declaration.

**Impact:** 
Improves code clarity and complies with Clippy suggestions, eliminating the warning.

---

- [X] These changes are part of https://github.com/servo/servo/issues/31500.

### DIFF
--- a/components/script/dom/dommatrixreadonly.rs
+++ b/components/script/dom/dommatrixreadonly.rs
@@ -822,8 +822,6 @@ impl DOMMatrixReadOnlyMethods for DOMMatrixReadOnly {
             return Err(error::Error::InvalidState);
         }
 
-        // Step 2. Let string be the empty string.
-        // Step 3. If is 2D is true, then:
         let cx = GlobalScope::get_cx();
         let to_string = |f: f64| {
             let value = jsval::DoubleValue(f);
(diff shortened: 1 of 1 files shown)


simonwuelker:
I intentionally wrote the code this way to match the specification (https://drafts.fxtf.org/geometry/#dommatrixreadonly-stringification-behavior)

chickenleaf:
> I intentionally wrote the code this way to match the specification steps (https://drafts.fxtf.org/geometry/#dommatrixreadonly-stringification-behavior)

Oh noo I'm so sorry I did notice the extensive comments, and asked around if the warning should be fixed and did what I was recommended. Would you recommend the warning be allowed?

simonwuelker:
> Oh noo I'm so sorry I did notice the extensive comments
No worries! I *personally* would choose to silence the warning. But what you did is fine too, I just wanted to provide context for why the code is the way it is.

components/script/dom/dommatrixreadonly.rs
```diff
@@ -839,7 +837,7 @@ impl DOMMatrixReadOnlyMethods for DOMMatrixReadOnly {
                 )
             }
         };
-        if self.is2D() {
+        let string = if self.is2D() {
```
mukilan, line 840: We need to preserve the specification related comments as [discussed here](https://servo.zulipchat.com/#narrow/stream/426955-Outreachy/topic/Improve.20code.20linting.20in.20Servo's.20CI/near/476784186) i.e we move the comments about steps 2 and 3 to be just before the `let string = if ..` line.

chickenleaf: just fixed it

mukilan (approved):
Thanks!

PR #33844, opened 2024-10-15
Remove SizeKeyword, merge it into Size

Loirooriol:
It's not really useful to have it.

---


mrobinson (approved):
This makes a lot of sense!

Issue #33486, opened 2024-09-18
The use of `tracing` is flooding the log with the `Debug` output of instrumented function arguments


mrobinson:
**Describe the bug:**

`tracing` is printing arguments to instrumented functions to the log. This likely shouldn't happen as the `Debug` output of many of our argument is too large for general logging purposes. In addition, it's unclear why `tracing` is logging anything at all when the `tracing` feature is disabled during build.

**To Reproduce:**

- Run `servoshell` with `RUST_LOG=info`, for example ` RUST_LOG=info ./mach run -r`.
- Note that the output is full of serialized style data, etc, which overwhelms any other output.

![Screenshot from 2024-09-18 08-16-02](https://github.com/user-attachments/assets/67b4b46b-acb2-4166-93b4-f5d83a4be009)

delan:
Interesting! I hadn’t noticed because I usually run `RUST_LOG='warn,servoshell<,servoshell>,...'`.

I think `tracing` may be [emitting `log` records](https://docs.rs/tracing/0.1.40/tracing/#emitting-log-records) (note that only the “additional” records are trace level):

> When the “log” feature is enabled, if no tracing Subscriber is active, invoking an event macro or creating a span with fields will emit a log record. […] Additionally, log records are also generated when spans are entered, exited, and closed. Since these additional span lifecycle logs have the potential to be very verbose, and don’t include additional fields, they will always be emitted at the Trace level, rather than inheriting the level of the span that generated them. 

```
$ cargo tree -e features -i tracing | rg -A5 ' tracing feature "log"'
├── tracing feature "log"
│   └── warp v0.3.7
│       └── webdriver v0.49.0
│           ├── webdriver feature "default"
│           │   ├── script v0.0.1 (/cuffs/code/servo/components/script) (*)
│           │   ├── script_traits v0.0.1 (/cuffs/code/servo/components/shared/script)
```

delan:
So we could disable the `tracing/log` feature, or we could make sure every thread(?) has a tracing subscriber.

I’m not sure we can do the former without patching `webdriver` *and* `warp`, due to feature unification and Cargo `[patch]` not touching transitive dependencies. It shouldn’t be too hard to upstream though, I think we would just need to gate the `tracing/log` dep behind a feature?

jschwe:
I'll [copy my comment on the tracing PR](https://github.com/servo/servo/pull/33027#issuecomment-2345215452) since I believe it is relevant here:

[...]

PR #33845, opened 2024-10-15
Gate all use of `tracing` behind Cargo feature

delan:
This patch compiles out any usage of `#[tracing::instrument]` and `tracing::span!()` unless Servo is built with `--features tracing`. That way, we can at least limit the effects of #33486 to tracing builds while we fix the bug.

---

- [x] These changes partially address #33486

- [x] Manual tests
  - [x] `./mach build && RUST_LOG=info ./mach run` no longer floods the logs
  - [x] `./mach build --features tracing && RUST_LOG=info ./mach run` still floods the logs


mrobinson (approved):
This seems like an okay workaround for the moment.

PR #33849, opened 2024-10-15
clippy: fix clippy warnings in `components/script`

komutumu:
This PR contains clippy fixes for files components/script/script_runtime.rs and components/script/script_thread.rs"

---

### DIFF
--- a/components/script/script_runtime.rs
+++ b/components/script/script_runtime.rs
@@ -122,9 +122,9 @@ impl fmt::Debug for CommonScriptMsg {
 }
 
 /// A cloneable interface for communicating with an event loop.
-#[allow(clippy::result_unit_err)]
 pub trait ScriptChan: JSTraceable {
     /// Send a message to the associated event loop.
+    #[allow(clippy::result_unit_err)]
     fn send(&self, msg: CommonScriptMsg) -> Result<(), ()>;
     /// Clone this handle.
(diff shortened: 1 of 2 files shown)


components/script/script_runtime.rs
```diff
@@ -122,6 +122,7 @@ impl fmt::Debug for CommonScriptMsg {
 }
 
 /// A cloneable interface for communicating with an event loop.
+#[allow(clippy::result_unit_err)]
```
mukilan, line 125: Can we move this to be above the `fn send()` method so that we don't inadvertently `allow` new methods in this trait?

components/script/script_runtime.rs
```diff
@@ -163,6 +164,7 @@ pub enum ScriptThreadEventCategory {
 /// An interface for receiving ScriptMsg values in an event loop. Used for synchronous DOM
 /// APIs that need to abstract over multiple kinds of event loops (worker/main thread) with
 /// different Receiver interfaces.
+#[allow(clippy::result_unit_err)]
```
mukilan, line 167: Same as above. Can this be moved to before the `fn recv` method?

jahielkomu: Okay @mukilan I have updated the PR. Thank you

mukilan (approved):

PR #33853, opened 2024-10-15
clippy: Fix warnings in `components/script/dom` & `components/servo`

taniishkaaa:
Fixes warnings in `components/script/dom` & `components/servo` as a part of #31500 

---

### DIFF
--- a/components/script/dom/htmlformelement.rs
+++ b/components/script/dom/htmlformelement.rs
@@ -88,6 +88,7 @@ use crate::task_source::TaskSource;
 #[derive(Clone, Copy, JSTraceable, MallocSizeOf, PartialEq)]
 pub struct GenerationId(u32);
 
+#[allow(clippy::type_complexity)]
 #[dom_struct]
 pub struct HTMLFormElement {
     htmlelement: HTMLElement,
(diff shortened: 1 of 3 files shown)


components/script/dom/htmlformelement.rs
```diff
@@ -88,6 +88,7 @@ use crate::task_source::TaskSource;
 #[derive(Clone, Copy, JSTraceable, MallocSizeOf, PartialEq)]
 pub struct GenerationId(u32);
 
+#[allow(clippy::type_complexity)]
```
mukilan, line 91: Does this need to be at the level of the whole `struct`? Can we not have allow it just for `past_names_map`?

taniishkaaa: Of course, I can do that. I wasn't sure if it'd make for cleaner code.

mukilan: It is better to have it closer to the source of the warning, so we don't inadvertently allow new members of this struct with complex types.

mukilan: About the "cleaner code", I assume here you mean the visual noise from the `allow` directive. I guess we can just add blank lines before and after the field ([example](https://github.com/servo/servo/blob/main/components/script/dom/htmlformelement.rs#L105)) so it doesn't seem too "crammed".

taniishkaaa: okay great! I'll add the blank lines :)

taniishkaaa: but won't `./mach test-tidy` report errors after doing that?

mukilan: I tried it and it didn't.

mukilan (approved):

mukilan (approved):

PR #33865, opened 2024-10-16
Some CanGc Fixes

### LINKED ISSUE #33683: Replace CanGc::note() calls with arguments passed by callers

komutumu:
Replaces CanGc::note() calls with arguments passed by callers in components/script/do/url.rs

---
- [X] These changes fix #33683 
- [X] These changes do not require tests because do not touch functionality

### DIFF
--- a/components/script/dom/bindings/codegen/Bindings.conf
+++ b/components/script/dom/bindings/codegen/Bindings.conf
@@ -235,6 +235,7 @@ DOMInterfaces = {
 
 'URL': {
     'weakReferenceable': True,
+    'canGc': ['Parse'],
 },
 
 'VRDisplay': {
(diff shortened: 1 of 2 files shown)


jdm (approved):
Thanks!

Issue #23818, opened 2019-07-20
Wrong Type for `ROUTER.add_route` callback?


gatoWololo:
## Background
I'm messing around with a typed version of the `ipc::router::add_route` method.
Original:
```rust
pub fn add_route(&self, receiver: OpaqueIpcReceiver, callback: RouterHandler)

// where
pub type RouterHandler = Box<FnMut(OpaqueIpcMessage) + Send>;
```

Mine:
```rust
pub fn add_route<T>(&self, receiver: IpcReceiver<T>,
                           mut callback: Box<FnMut(Result<T, ipc_channel::Error>) + Send>)
```

So I call the `OpaqueIpcMessage::to` message on behalf of the user, and hide all the implementations details about `Opaque*`.

When using the typed API I found two places in Servo where I couldn't get it to type check.

## Actual Problem
From the latest master unmodified code. I'm looking at the function `add_cache_listener_for_element` in `servo/components/script/dom/htmlimageelement.rs`
The relevant parts are:
```rust
fn add_cache_listener_for_element(
            ...
            let (responder_sender, responder_receiver) = ipc::channel().unwrap();
            ...
            ROUTER.add_route(responder_receiver.to_opaque(), Box::new(move |message| {
                ...
                let image = message.to().unwrap();
                ...
                            element.process_image_response_for_environment_change(image, ...);
            ...
            }));

            image_cache.add_listener(id, ImageResponder::new(responder_sender, id));
```

Notice type inference is happening at two relevant places here: The type of `channel<T>()` and the type of `to<T>()`. I explicitly add a turbofish to `channel`:
```diff
- let (responder_sender, responder_receiver) = ipc::channel().unwrap();
+ let (responder_sender, responder_receiver) = ipc::channel::<PendingImageResponse>().unwrap();
```
I know this is the correct type from the call to `ImageResponder::new(responder_sender, id));` at the bottom which takes the `responder_sender`: `pub fn new(sender: IpcSender<PendingImageResponse>, ...) -> ...`. The **compiler agrees** with me, and type checks the program.

Now I try adding the type for the call to `.to()`:
```diff
- let image = message.to().unwrap();
+ let image = message.to::<PendingImageResponse>().unwrap();
```

The message comes from the `IpcReceiver<T>` turned `OpaqueReceiver`. Which we have established has type `PendingImageResponse`. However, the way it is being using in the callback:
```rust
element.process_image_response_for_environment_change(image, ...);
```
The compiler infers:
```
[...]

PR #33866, opened 2024-10-16
Use `ROUTER::add_typed_route` instead of `ROUTER::add_route` everywhere

simonwuelker:
This updates the `ipc-channel`, `webxr` and `servo-media` dependencies to the latest version and uses `ROUTER::add_typed_route` to make our routes more typesafe. You can read more about `ROUTER::add_typed_route` in https://github.com/servo/ipc-channel/issues/242.

This revealed *two* issues where we accidentally abuse https://github.com/servo/ipc-channel/issues/239, one of which was previously described in https://github.com/servo/servo/issues/23818.
These were fixed in https://github.com/servo/servo/pull/33882.

I also made `webxr` and `webxr-api` workspace dependencies, because every crate in `components/` used the same versions anyways, and doing so made it easier to patch them while testing.

---
- [X] These changes do not require tests because they do not change behaviour

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,7 +327,7 @@ dependencies = [
  "backtrace",
  "base",
  "crossbeam-channel",
- "ipc-channel 0.19.0",
+ "ipc-channel",
  "libc",
  "log",
  "mach2",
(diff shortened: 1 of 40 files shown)


mrobinson (approved):
Excellent!

PR #33878, opened 2024-10-17
clippy: Fix several warning about `NetworkListener::to_callback()`

lwz23:
---

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -2157,7 +2157,7 @@ impl Document {
             task_source,
             canceller: Some(canceller),
         }
-        .to_callback();
+        .into_callback();
         self.loader_mut()
             .fetch_async_with_callback(load, request, callback);
     }
@@ -2177,7 +2177,7 @@ impl Document {
             task_source,
             canceller: Some(canceller),
         }
-        .to_callback();
+        .into_callback();
         self.loader_mut()
             .fetch_async_background(request, callback, cancel_override);
     }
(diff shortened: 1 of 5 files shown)


mrobinson (changes requested):
I've done a rebase of this branch so you might need to check it out again from your fork.

components/script/dom/bindings/refcounted.rs
```diff
@@ -55,6 +55,7 @@ pub use self::dummy::LIVE_REFERENCES;
 /// A pointer to a Rust DOM object that needs to be destroyed.
 struct TrustedReference(*const libc::c_void);
 unsafe impl Send for TrustedReference {}
+unsafe impl Sync for TrustedReference {}
```
mrobinson, line 58: I would revert this change. It's unclear to me that the value stored in `TrustedReference` is actually `Sync` which means it can be used safely from more than one thread at a time. By implement `unsafe impl Sync` here you are promising the compiler that is the case. As a reviewer I would need to look much more closely at this code to understand if that's the case. Since it is unrelated to the other change which I do understand, please just revert it for now.

mrobinson (changes requested):

components/script/dom/bindings/refcounted.rs
lwz23: I delete this change, and push it to my branch. I am new to Contribute, sorry if I made some mistakes.

mrobinson (approved):

PR #33881, opened 2024-10-17
clippy: Rename `RequestId::new()` to `RequestId::next()` and fix one more issue in `servo/lib.rs`

komutumu:
This PR addresses clippy warnings in files; 
- components/servo/lib.rs 
- components/shared/net/request.rs
- components/script/fetch.rs

---
- [X] These changes are part of  #31500

### DIFF
--- a/components/servo/lib.rs
+++ b/components/servo/lib.rs
@@ -228,6 +228,7 @@ where
         feature = "tracing",
         tracing::instrument(skip(embedder, window), fields(servo_profiling = true))
     )]
+    #[allow(clippy::new_ret_no_self)]
     pub fn new(
         mut embedder: Box<dyn EmbedderMethods>,
         window: Rc<Window>,
--- a/components/shared/net/request.rs
+++ b/components/shared/net/request.rs
@@ -22,8 +22,8 @@ use crate::{ReferrerPolicy, ResourceTimingType};
 /// An id to differeniate one network request from another.
 pub struct RequestId(usize);
 
-impl Default for RequestId {
-    fn default() -> Self {
+impl RequestId {
+    pub fn next() -> Self {
         static NEXT_REQUEST_ID: AtomicUsize = AtomicUsize::new(0);
         Self(NEXT_REQUEST_ID.fetch_add(1, Ordering::Relaxed))
     }
@@ -285,7 +285,7 @@ pub struct RequestBuilder {
 impl RequestBuilder {
     pub fn new(url: ServoUrl, referrer: Referrer) -> RequestBuilder {
         RequestBuilder {
-            id: RequestId::default(),
+            id: RequestId::next(),
             method: Method::GET,
             url,
             headers: HeaderMap::new(),
--- a/components/shared/net/request.rs
+++ b/components/shared/net/request.rs
@@ -22,8 +22,8 @@ use crate::{ReferrerPolicy, ResourceTimingType};
 /// An id to differeniate one network request from another.
 pub struct RequestId(usize);
 
-impl RequestId {
-    pub fn new() -> Self {
+impl Default for RequestId {
+    fn default() -> Self {
         static NEXT_REQUEST_ID: AtomicUsize = AtomicUsize::new(0);
         Self(NEXT_REQUEST_ID.fetch_add(1, Ordering::Relaxed))
     }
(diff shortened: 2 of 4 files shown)


components/servo/lib.rs
```diff
@@ -228,6 +228,7 @@ where
         feature = "tracing",
         tracing::instrument(skip(embedder, window), fields(servo_profiling = true))
     )]
+    #[allow(clippy::new_ret_no_self)]
```
mrobinson, line 231:
```suggestion

    #[allow(clippy::new_ret_no_self)]
```

components/shared/net/request.rs
```diff
@@ -22,8 +22,8 @@ use crate::{ReferrerPolicy, ResourceTimingType};
 /// An id to differeniate one network request from another.
 pub struct RequestId(usize);
 
-impl RequestId {
-    pub fn new() -> Self {
+impl Default for RequestId {
+    fn default() -> Self {
```
mrobinson, line 26: `default()` doesn't quite capture what this is doing. Instead please call this `next()`:

```suggestion
impl RequestId {
    fn next() -> Self {
```

mrobinson (changes requested):

components/shared/net/request.rs
jahielkomu: Okay Sir, let me do that.

components/servo/lib.rs
jahielkomu: @mrobinson do you suggest I add an empty line before the ```allow``` here?

jahielkomu: When I add the empty line, ```test-tidy``` reports some issues from  there.

simonwuelker:
> These changes fix https://github.com/servo/servo/issues/31500

FYI, if you phrase your PR description like that then github will automatically close #31500 as completed when this PR is merged. Maybe rephrase it to "These changes are part of #31500".

jahielkomu:
> > These changes fix #31500
> 
> FYI, if you phrase your PR description like that then github will automatically close #31500 as completed when this PR is merged. Maybe rephrase it to "These changes are part of #31500".

The part of  "These changes fix __"  is already there on the part where the PRs are created, so I think to largely address it would be from that point.
Thank you, let me edit mine.

components/servo/lib.rs
mrobinson: Apologies. My suggestion was a mistake. Please discard it.

components/shared/net/request.rs
jahielkomu: @mrobinson when ```fn next()``` is used ```build -d``` reports errors ``` RequestId` is not an iterator``` 
Should we rename it to something else?

mrobinson: `next_id` sounds great. Thanks for checking!

jahielkomu: > `next_id` sounds great. Thanks for checking!

@mrobinson, there appears to still be an issue  with renaming it to anything else but ```default``` unless ```fn default``` is implemented in some way. Let me share a screenshot. 

Screenshot from 2024-10-17 18-31-11

simonwuelker: you need to remove the `impl Default for RequestId` block and use `impl RequestId` instead.

jahielkomu: > you need to remove the `impl Default for RequestId` block and use `impl RequestId` instead.

That is the warning we are trying to work on, the only other option would be to ```allow``` it

mukilan: The warning is triggered only because the method is currently named `new`. It will disappear if we renamed `new` to `next_id`, so there is no need to `allow` it or have `impl Default`.

jahielkomu: > The warning is triggered only because the method is currently named `new`. It will disappear if we renamed `new` to `next_id`, so there is no need to `allow` it or have `impl Default`.

You are right, looking at the suggestion provided by @mrobinson ```impl Default``` is eliminated. 
Thank you @mukilan

mrobinson (approved):

simonwuelker:
(PR title is wrong, the method is named `RequestId::next()` not `RequestId::next_id()`)

mrobinson:
> (PR title is wrong, the method is named `RequestId::next()` not `RequestId::next_id()`)

Thanks for pointing that out. I missed that clippy wasn't actually complaining about the `next()` name.

PR #33902, opened 2024-10-18
Prevent moving CanGc values between threads/tasks

jdm:
This change prevents a common misuse of CanGc. If the only use of a value is within a queued task or inside of a spawned thread, it should start a new propagated chain within that context instead of reusing a value from the initiator.

---
- [x] These changes do not require tests because they are compile-time only.


jdm:
FYI @chickenleaf @taniishkaaa

chickenleaf:
noted

chickenleaf:
so the things to look for are if the call is under `!task` or under something like this `thread::Builder:..........spawn`. in these cases we start with a new CanGc

jdm:
> so the things to look for are if the call is under `!task` or under something like this `thread::Builder:..........spawn`. in these cases we start with a new CanGc

Correct.

sagudev (approved):

Issue #33903, opened 2024-10-18
GC borrow hazard in HTMLImageElement::finish_reacting_to_environment_change


jdm:
We hold a mutable borrow of a field of HTMLImageElement at https://github.com/servo/servo/blob/main/components/script/dom/htmlimageelement.rs#L1272, then call a method that can trigger a GC at https://github.com/servo/servo/blob/main/components/script/dom/htmlimageelement.rs#L1283 while the borrow is still outstanding. We should move the abort_request call outside of the `{}` block that already exists to limit how long the borrow is active.

chickenleaf:
id like to be assigned this!

jdm:
Fixed by #33904.

PR #33904, opened 2024-10-18
Fix GC borrow hazard in HTMLImageElement::finish_reacting_to_environment_change

chickenleaf:
**File:** `components/script/dom/htmlimageelement.rs`

**PR Description**

**Issue:** 
We’re holding a field of `HTMLImageElement` at [line 1272](https://github.com/servo/servo/blob/main/components/script/dom/htmlimageelement.rs#L1272), then calling a method at [line 1283](https://github.com/servo/servo/blob/main/components/script/dom/htmlimageelement.rs#L1283) that could trigger garbage collection while still using it. To fix this, we should move the `abort_request` call outside the existing block to shorten the borrow time.

**Changes:** 
Moved the `abort_request` call outside of the existing block to reduce the duration of the mutable borrow.  

**Impact:** 
This change prevents potential issues with memory safety during garbage collection.

---

- [X] These changes fix https://github.com/servo/servo/issues/33903

### DIFF
--- a/components/script/dom/htmlimageelement.rs
+++ b/components/script/dom/htmlimageelement.rs
@@ -1272,7 +1272,7 @@ impl HTMLImageElement {
                     // Step 15.5
                     mem::swap(&mut this.current_request.borrow_mut(), &mut pending_request);
                 }
-                this.abort_request(State::Unavailable, ImageRequestPhase::Pending, can_gc);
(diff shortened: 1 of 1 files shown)


jdm (approved):

chickenleaf:
i have rebased and committed again.

PR #33919, opened 2024-10-19
Support persisting unminified external stylesheets

### LINKED ISSUE #33662: Support local tweaking of external stylesheets
https://github.com/servo/servo/pull/16348 introduced a way to cache external JS scripts, make them easier to read, and https://github.com/servo/servo/pull/26711 added support for using the cached versions instead of fetching them from the network. This makes it easier to investigate when pages are misbehaving in Servo, because we can add debugging statements or comment out problematic bits of code and see how Servo behaves.

We should add the same support for external stylesheets, to make diagnosing layout issues easier! Links to relevant pieces of code forthcoming when I am in front of a laptop.

Taym95:
---

- [x] These changes fix #33662

### DIFF
--- a/components/script/dom/window.rs
+++ b/components/script/dom/window.rs
@@ -545,14 +545,6 @@ impl Window {
         self.replace_surrogates
     }
 
-    pub fn unminify_js(&self) -> bool {
-        self.unminify_js
-    }
-
-    pub fn unminify_css(&self) -> bool {
-        self.unminify_css
-    }
-
     pub fn get_player_context(&self) -> WindowGLContext {
         self.player_context.clone()
     }
@@ -2229,19 +2221,14 @@ impl Window {
         assert!(self.document.get().is_none());
         assert!(document.window() == self);
         self.document.set(Some(document));
-        if self.unminify_js {
-            // Set a path for the document host to store unminified scripts.
-            let mut path = env::current_dir().unwrap();
-            path.push("unminified-js");
-            *self.unminified_js_dir.borrow_mut() =
-                Some(path.into_os_string().into_string().unwrap());
-        } else if self.unminify_css {
-            // Set a path for the document host to store unminified stylesheets.
-            let mut path = env::current_dir().unwrap();
-            path.push("unminified-css");
-            *self.unminified_css_dir.borrow_mut() =
-                Some(path.into_os_string().into_string().unwrap());
-        }
+
+        set_unminified_path(self.unminify_js, &self.unminified_js_dir, "unminified-js");
+
+        set_unminified_path(
+            self.unminify_css,
+            &self.unminified_css_dir,
+            "unminified-css",
+        );
     }
 
     /// Commence a new URL load which will either replace this window or scroll to a fragment.
@@ -2889,4 +2876,13 @@ fn is_named_element_with_name_attribute(elem: &Element) -> bool {
 
 fn is_named_element_with_id_attribute(elem: &Element) -> bool {
     elem.is_html_element()
+}
+
+fn set_unminified_path(option: bool, dir_ref: &DomRefCell<Option<String>>, folder_name: &str) {
+    if option {
+        // Set a path for the document host to store unminified files.
+        let mut path = env::current_dir().unwrap();
+        path.push(folder_name);
+        *dir_ref.borrow_mut() = Some(path.into_os_string().into_string().unwrap());
+    }
 }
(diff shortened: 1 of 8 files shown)


components/script/dom/window.rs
```diff
@@ -2218,13 +2229,19 @@ impl Window {
...
         self.document.set(Some(document));
-        if !self.unminify_js {
-            return;
+        if self.unminify_js {
+            // Set a path for the document host to store unminified scripts.
+            let mut path = env::current_dir().unwrap();
+            path.push("unminified-js");
+            *self.unminified_js_dir.borrow_mut() =
+                Some(path.into_os_string().into_string().unwrap());
+        } else if self.unminify_css {
```
jdm, line 2238: This makes unminifying css and js exclusive. Let's have two separate if statements instead

components/script/dom/window.rs
```diff
@@ -542,6 +549,10 @@ impl Window {
         self.unminify_js
     }
 
+    pub fn unminify_css(&self) -> bool {
```
jdm, line 552: I think we don't need this method if we return None from unminified_css_dir.

components/script/stylesheet_loader.rs
```diff
@@ -91,6 +96,94 @@ pub struct StylesheetContext {
...
 
+impl StylesheetContext {
+    fn unminify_css(&self, data: Vec<u8>, file_url: ServoUrl) -> Vec<u8> {
+        if !self.document.root().window().unminify_css() {
+            return data;
+        }
+
+        let mut style_content = data;
+
+        // Write the minified code to a temporary file and pass its path as an argument
```
jdm, line 107: Since a lot of this is duplicated with the script code, can we extract it into some common helpers in a new components/script/umminify.rs file?

jdm (changes requested):
Good start!

components/script/stylesheet_loader.rs
Taym95: Yes, this was my next step

components/script/stylesheet_loader.rs
```diff
@@ -91,6 +96,94 @@ pub struct StylesheetContext {
...
+
+        let path = if has_name {
+            // External stylesheet.
+            path.join(&file_url[url::Position::BeforeHost..])
+        } else {
+            // Inline script or url ends with '/'
+            base.join(Uuid::new_v4().to_string())
+        };
+
+        if let Err(e) = create_dir_all(base.clone()) {
```
Taym95, line 168: @jdm I was testing this, and on some websites like Google, it fails because the CSS file name is too long, which prevents creating the file. What do you think about shortening or modifying the name to make it work?

jdm: Very interesting! Let's truncate the name to some reasonable number like 100 characters, I guess?

components/script/stylesheet_loader.rs
Taym95: I removed duplicate code, I tried to not break existing code, please review carefully

components/script/stylesheet_loader.rs
```diff
@@ -98,82 +97,29 @@ pub struct StylesheetContext {
...
-                _ => {
-                    log::warn!(
-                        "Failed to execute js-beautify --type css. Will store unmodified css"
-                    );
-                },
+        if let Some((input, mut output)) = create_temp_files() {
+            if execute_js_beautify(
+                input.path(),
+                output.try_clone().unwrap(),
+                BeautifyFileType::Css,
```
jdm, line 110: Is the BeautifyFileType and the directory the only pieces in the code that is different between scripts and stylesheets? It feels like we could still remove some more duplicated logic.

Taym95: `BeautifyFileType` is used only to add `--type css` to `js-beautify` and later `html`, but the main difference is how the content is read and written. For JS, there’s s`cript.code`, while for CSS, I’ve defined `style_content`

Taym95: @jdm what do you think? should I remove more duplicated logic?

components/script/unminify.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+#[derive(Debug)]
+pub enum BeautifyFileType {
+    Css,
+    Js,
+}
+
+pub fn execute_js_beautify(input: &Path, output: File, file_type: BeautifyFileType) -> bool {
+    let mut cmd = Command::new("npx");
```
jdm, line 36: The original code invoked js-beautify directly, but now we use npx. Why?

jdm (approved):
I see what you mean; there are more differences than I first noticed. Only one question left before we merge this!

jdm:
Clippy has a note for you!

PR #33922, opened 2024-10-19
Refactor positioned layout

Loirooriol:
This unifies the size resolution into AbsoluteAxisSolver, since it needs
to know the size in order to resolve auto margins correctly anyways.
This will allow adding support for sizing keywords in a follow-up patch.

Also, this avoids doing multiple layouts due to min and max constraints,
improving performance.

Additionally, tables may end up having a custom size, different than
what we would expect by just looking at the sizing properties. This
patch ensures that we resolve margins correctly with the final size,
resulting in 2 tests now passing.

---

### DIFF
--- a/components/layout_2020/positioned.rs
+++ b/components/layout_2020/positioned.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+use std::mem;
+
 use app_units::Au;
 use rayon::iter::IntoParallelRefMutIterator;
 use rayon::prelude::{IndexedParallelIterator, ParallelIterator};
@@ -25,7 +27,8 @@ use crate::geom::{
     AuOrAuto, LengthPercentageOrAuto, LogicalRect, LogicalSides, LogicalVec2, PhysicalPoint,
     PhysicalRect, PhysicalVec, ToLogical, ToLogicalWithContainingBlock,
 };
-use crate::style_ext::{ComputedValuesExt, DisplayInside};
+use crate::sizing::ContentSizes;
+use crate::style_ext::{Clamp, ComputedValuesExt, DisplayInside};
 use crate::{ContainingBlock, DefiniteContainingBlock, IndefiniteContainingBlock};
 
 #[derive(Debug, Serialize)]
@@ -240,13 +243,11 @@ impl PositioningContext {
             style: &new_fragment.style,
         };
 
-        let take_hoisted_boxes_pending_layout = |context: &mut Self| match context
-            .for_nearest_positioned_ancestor
-            .as_mut()
-        {
-            Some(fragments) => std::mem::take(fragments),
-            None => std::mem::take(&mut context.for_nearest_containing_block_for_all_descendants),
-        };
+        let take_hoisted_boxes_pending_layout =
+            |context: &mut Self| match context.for_nearest_positioned_ancestor.as_mut() {
+                Some(fragments) => mem::take(fragments),
+                None => mem::take(&mut context.for_nearest_containing_block_for_all_descendants),
+            };
 
         // Loop because it’s possible that we discover (the static position of)
         // more absolutely-positioned boxes while doing layout for others.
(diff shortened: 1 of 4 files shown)


components/layout_2020/positioned.rs
```diff
@@ -553,149 +580,65 @@ impl HoistedAbsolutelyPositionedBox {
...
+                            .writing_mode
+                            .is_horizontal(),
+                        "Mixed horizontal and vertical writing modes are not supported yet"
+                    );
+
+                    // Clear the context since we will lay out the same descendants
+                    // more than once. Otherwise, absolute descendants will create
+                    // multiple fragments which could later lead to double-borrow
+                    // errors.
+                    positioning_context.clear();
```
Loirooriol, line 605: I guess this is no longer necessary? But I'm not sure about the implications of removing it.

mrobinson: If layout only happens once you can and should remove this as it clears all hoisted absolutes from layout.

components/layout_2020/positioned.rs
```diff
@@ -456,27 +457,32 @@ impl HoistedAbsolutelyPositionedBox {
...
+            IndependentFormattingContext::NonReplaced(non_replaced) => (
+                non_replaced
+                    .style
+                    .content_box_size_deprecated(&indefinite_containing_block, &pbm),
+                style
+                    .content_min_box_size_deprecated(&indefinite_containing_block, &pbm)
+                    .map(|t| t.map(Au::from).auto_is(Au::zero)),
+                style
+                    .content_max_box_size_deprecated(&containing_block.into(), &pbm)
+                    .map(|t| t.map(Au::from)),
```
mrobinson, line 484:
```suggestion
                    .map(|value| value.map(Au::from).auto_is(Au::zero)),
                style
                    .content_max_box_size_deprecated(&containing_block.into(), &pbm)
                    .map(|value| value.map(Au::from)),
```

components/layout_2020/positioned.rs
```diff
@@ -521,27 +530,45 @@ impl HoistedAbsolutelyPositionedBox {
...
+                IndefiniteContainingBlock::new_for_style_and_block_size(&style, block_size);
+            context
+                .inline_content_sizes(
+                    layout_context,
+                    &containing_block_for_children,
+                    &containing_block.into(),
+                )
+                .sizes
+        }));
+        let mut block_axis = block_axis_solver.solve_tentativelly();
```
mrobinson, line 565: Nit: Typo in the word "tentatively" along with the obligatory apologies for the English language.

components/layout_2020/positioned.rs
mrobinson, line 605: Yeah this can totally be removed.

components/layout_2020/positioned.rs
```diff
@@ -919,6 +878,26 @@ impl<'a> AbsoluteAxisSolver<'a> {
         }
     }
 
+    fn solve_tentativelly(&mut self) -> AxisResult {
+        self.solve(None::<fn() -> ContentSizes>)
```
mrobinson, line 882: Is the type annotation really necessary here?

mrobinson (approved):
Amazing PR, @Loirooriol.

components/layout_2020/positioned.rs
Loirooriol: Yes, I added the method to avoid having to type the annotation.

```
error[E0283]: type annotations needed
   --> components/layout_2020/positioned.rs:871:20
    |
871 |         self.solve(None)
    |              ----- ^^^^ cannot infer type of the type parameter `T` declared on the enum `Option`
    |              |
    |              required by a bound introduced by this call
    |
    = note: multiple `impl`s satisfying `_: FnOnce()` found in the following crates: `alloc`, `core`:
            - impl<A, F> FnOnce<A> for &F
              where A: std::marker::Tuple, F: Fn<A>, F: ?Sized;
            - impl<A, F> FnOnce<A> for &mut F
              where A: std::marker::Tuple, F: FnMut<A>, F: ?Sized;
            - impl<Args, F, A> FnOnce<Args> for std::boxed::Box<F, A>
              where Args: std::marker::Tuple, F: FnOnce<Args>, A: Allocator, F: ?Sized;
            - impl<F, Args> FnOnce<Args> for Exclusive<F>
              where F: FnOnce<Args>, Args: std::marker::Tuple;
            - impl<R, F> FnOnce<()> for AssertUnwindSafe<F>
              where <F as FnOnce<()>>::Output == R, F: FnOnce();
note: required by a bound in `AbsoluteAxisSolver::<'a>::solve`
   --> components/layout_2020/positioned.rs:785:51
    |
785 |     fn solve(&self, get_content_size: Option<impl FnOnce() -> ContentSizes>) -> AxisResult {
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `AbsoluteAxisSolver::<'a>::solve`
help: consider specifying the generic argument
    |
871 |         self.solve(None::<T>)
    |                        +++++
```

components/layout_2020/positioned.rs
Loirooriol: Let's see if the bots complain...

PR #33928, opened 2024-10-20
Mark promise creation methods with CanGc

jdm:
These methods call JS API methods that can trigger a GC, so we need to propagate the CanGc arguments appropriately.

---
- [x] These changes do not require tests because they are compile-time only.


chickenleaf:
thank you!! \o/

sagudev (approved):

PR #33931, opened 2024-10-20
CanGc fixes from `constantsourcenode.rs` & `window.rs`

taniishkaaa:
Replaces CanGc::note() calls with arguments passed by callers in `components/script/dom/constantsourcenode.rs` and `components/script/dom/window.rs`

---

### DIFF
--- a/components/script/dom/focusevent.rs
+++ b/components/script/dom/focusevent.rs
@@ -47,6 +47,7 @@ impl FocusEvent {
         reflect_dom_object_with_proto(Box::new(FocusEvent::new_inherited()), window, proto, can_gc)
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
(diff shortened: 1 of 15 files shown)


components/script/dom/focusevent.rs
```diff
@@ -47,6 +47,7 @@ impl FocusEvent {
         reflect_dom_object_with_proto(Box::new(FocusEvent::new_inherited()), window, proto, can_gc)
     }
 
+    #[allow(clippy::too_many_arguments)]
```
jdm, line 50: In future, can you keep clippy fixes like this to separate PRs?

jdm (approved):

components/script/dom/focusevent.rs
taniishkaaa: Alright, noted!

PR #33935, opened 2024-10-20
clippy: fix using clone() warning in components\layout_2020\display_l…

Noble14477:
…ist\mod.rs:1365:25

Clippy suggests that we remove the clone() calls for margin.left and margin.top. Instead of having this Size2D::new(margin.left.clone(), margin.top.clone()) we can have Size2D::new(margin.left, margin.top)

---


mukilan (approved):
Thank you!

Issue #33927, opened 2024-10-20
GC borrow hazard in ModuleTree::fetch_module_descendants


jdm:
We borrow `self.visited_urls` mutably, then later call `fetch_single_module_script` which might trigger a GC operation.

chickenleaf:
id like to work on this!

jdm:
Use `./mach test-wpt tests/wpt/tests/html/semantics/scripting-1/the-script-element/module` to verify that the implementation still works.

jdm:
Fixed by #33942.

PR #33942, opened 2024-10-21
Resolved potential GC borrow hazards in ModuleTree::append_handler, ModuleTree::append_dynamic_module_handler, and ModuleTree::fetch_module_descendants'

### LINKED ISSUE #33926: GC borrow hazard in ModuleTree::append_handler
We mutably borrow `self.promise` then immediately create a new promise object, which is an operation that can trigger a GC. The same hazard is present in ModuleTree::append_dynamic_module_handler as well.

chickenleaf:
**Files:** `components\script\script_module.rs`

**PR Description**

**Issue:**
Addressed GC borrow hazards in the following functions:
  1.   `ModuleTree::append_handler`
  2.   `ModuleTree::append_dynamic_module_handler`
  3.   `ModuleTree::fetch_module_descendants`

These functions previously performed mutable borrows that could lead to a GC panic under certain conditions.

**Changes:**

**Impact:**
1. Prevents GC panics by avoiding mutable borrows during operations that might trigger garbage collection.
2. Improves the stability of the module fetching and handling process.

**Verification:**
1. Verified by running the `./mach test-wpt tests/wpt/tests/html/semantics/scripting-1/the-script-element/module` suite.
2. Followed the steps to reproduce the original GC panic, no crashes, confirming the issue is resolved.

---

- [X] These changes fix #33926 and #33927 


components/script/script_module.rs
```diff
@@ -357,14 +357,14 @@ impl ModuleTree {
         let comp = InRealm::Entered(&realm);
         let _ais = AutoIncumbentScript::new(&owner.global());
 
+        let new_promise = Promise::new_in_current_realm(comp);
+
         let mut promise = self.promise.borrow_mut();
```
jdm, line 362: This borrow is still active when we call append_native_handler, so we need to reduce its scope.

components/script/script_module.rs
```diff
@@ -393,11 +393,12 @@ impl ModuleTree {
...
```
jdm, line 398: Same issue.

jdm (changes requested):

components/script/script_module.rs
jdm: Ideally we only use borrow calls in expressions/statements that need them, rather than open-ended ones like this.

components/script/script_module.rs
```diff
@@ -758,20 +759,21 @@ impl ModuleTree {
...
+                let visited_urls = self.visited_urls.borrow().clone();
+                let options = options.descendant_fetch_options();
+
+                for url in urls_to_fetch {
                     // https://html.spec.whatwg.org/multipage/#internal-module-script-graph-fetching-procedure
                     // Step 1.
-                    assert!(visited_urls.get(&url).is_some());
-
-                    let options = options.descendant_fetch_options();
+                    debug_assert!(self.visited_urls.borrow().contains(&url));
```
jdm, line 768:
```suggestion
                    assert!(self.visited_urls.borrow().contains(&url));
```

jdm (approved):
Looking good!

components/script/script_module.rs
chickenleaf: Are we using assert for all places that has a potential GC hazard? Even after making changes to ensure safety?

jdm: The original code used assert, so we shouldn't modify it in this PR. The specification this code is implementing explicitly has an assert step.

jdm (approved):

jdm:
You'll need to address the build errors caused by the changes in #33937.

chickenleaf:
> You'll need to address the build errors caused by the changes in #33937.

done

Loirooriol:
It's confusing to have the commit from https://github.com/servo/servo/pull/33951 here because it's totally unrelated,  I guess you updated your PR doing a merge instead of a rebase or something like that.

PR #33950, opened 2024-10-21
Implement keyword sizes on absolute/fixed positioned elements

Loirooriol:
---

### DIFF
--- a/components/layout_2020/geom.rs
+++ b/components/layout_2020/geom.rs
@@ -624,7 +624,7 @@ impl ToLogicalWithContainingBlock<LogicalRect<Au>> for PhysicalRect<Au> {
 
 /// The possible values accepted by the sizing properties.
 /// <https://drafts.csswg.org/css-sizing/#sizing-properties>
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub(crate) enum Size<T> {
     /// Represents an `auto` value for the preferred and minimum size properties,
     /// or `none` for the maximum size properties.
@@ -643,6 +643,8 @@ pub(crate) enum Size<T> {
     /// <https://drafts.csswg.org/css-sizing/#valdef-width-length-percentage-0>
     Numeric(T),
 }
+
+impl<T: Copy> Copy for Size<T> {}
 
 impl<T> Default for Size<T> {
     #[inline]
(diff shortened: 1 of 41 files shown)


components/layout_2020/geom.rs
```diff
@@ -624,7 +624,7 @@ impl ToLogicalWithContainingBlock<LogicalRect<Au>> for PhysicalRect<Au> {
 
 /// The possible values accepted by the sizing properties.
 /// <https://drafts.csswg.org/css-sizing/#sizing-properties>
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
```
mrobinson, line 627: It probably makes sense to add a `Copy` implementation to this type if `T` always implements `Copy` as well.

components/layout_2020/positioned.rs
```diff
@@ -775,24 +772,52 @@ impl<'a> AbsoluteAxisSolver<'a> {
...
-                    let available_size = self.containing_size -
-                        anchor.inset() -
-                        self.padding_border_sum -
-                        margin_start -
-                        margin_end;
-                    Some(content_size.shrink_to_fit(available_size))
+        let mut get_content_size = get_content_size.map(|get_content_size| {
+            let content_size = LazyCell::new(get_content_size);
+            move || *content_size
+        });
```
mrobinson, line 778: Is the idea here that you'd like to cache the results of the call to `get_content_size` so that if the return value is called more than once the inner `get_content_size` is only called once?

Loirooriol: Yes, it can be called more than once if various of `min-width`, `max-width` and `width` (or height) are intrinsic.

mrobinson (approved):

components/layout_2020/geom.rs
Loirooriol: OK, then I don't need to `clone()` for `AbsoluteAxisSolver::computed_size`.

components/layout_2020/positioned.rs
Loirooriol: Being called once is ensured by `FnOnce`, but I have added this comment:

```rs
// The provided `get_content_size` is a FnOnce but we may need its result multiple times.
// A LazyCell will only invoke it once if needed, and then reuse the result.
```

PR #27869, opened 2020-11-28
Change default referrer policy to strict-origin-when-cross-origin

CYBAI:
---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [x] These changes fix #27868

- [ ] There are tests for these changes OR
- [ ] These changes do not require tests because ___

Issue #26568, opened 2020-05-19
Use real values in WorkerScriptLoadOrigin for DOM worker constructor


jdm:
We pass None for the referrer policy and referrer urls in [this code](https://github.com/servo/servo/blob/e1cc38bea8a701108b6f2fa809f341769613b55f/components/script/dom/worker.rs#L93-L97). We should obtain the real values from the global scope instead. It probably makes sense to fix https://github.com/servo/servo/issues/11863 at the same time, so whatever API we add to GlobalScope to return a referrer policy returns a real value in workers.

PR #33977, opened 2024-10-23
Implement PolicyContainer and update the default ReferrerPolicy

### LINKED ISSUE #11863: Implement referrer policy for WorkerGlobalScope
See spec at https://w3c.github.io/webappsec-referrer-policy/. All servo referrer policy implementation thus far is just for documents. See #10311 for reference.

### LINKED ISSUE #27868: Set referrer policy default to strict-origin-when-cross-origin
Spec update in each specifications:

- `webappsec-referrer-policy`: https://github.com/w3c/webappsec-referrer-policy/pull/142
- `fetch`: https://github.com/whatwg/fetch/pull/1066
- `html`: https://github.com/whatwg/html/pull/5783

WPT: https://github.com/web-platform-tests/wpt/pull/26441

Firefox Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1589074

shanehandley:
Implements PolicyContainer and adds it where required. In the process, it changes the default ReferrerPolicy to "strict-origin-when-cross-origin" (https://github.com/servo/servo/issues/27868, which has a previous PR: https://github.com/servo/servo/pull/27869).

Policy container is then available to `WorkerGlobalScope` which helps resolve https://github.com/servo/servo/issues/11863 and https://github.com/servo/servo/issues/26568.

---

- [x] These changes fix #27868 

Some tests related to inherited referrers in iframes have started failing for unknown reasons (I wasn't able to get to the bottom of these) but may relate to previously defaulting the referrer policy to `NoReferrerWhenDowngrade`. 

[Try](https://github.com/shanehandley/servo/actions/runs/11475198929)


components/net/fetch/methods.rs
```diff
@@ -156,26 +157,33 @@ pub async fn fetch_with_cors_cache(
...
-        parser_metadata: csp::ParserMetadata::None,
-    };
-    // TODO: Instead of ignoring violations, report them.
-    request
-        .csp_list
-        .as_ref()
-        .map(|c| c.should_request_be_blocked(&csp_request).0)
-        .unwrap_or(csp::CheckResult::Allowed)
+    match &request.policy_container {
+        RequestPolicyContainer::Client => csp::CheckResult::Allowed,
```
simonwuelker, line 161: Can `RequestPolicyContainer` ever be `Client` here? If not, then it would probably be better to assert on that instead of returning `Allowed`.

```rust
let RequestPolicyContainer::PolicyContainer(container) = &request.policy_container else {
    panic!("policy container should not be \"client\"");
}
```

I haven't read the spec (will do that once I have time), but if it doesn't tell us how to handle `"client"` here then it's probably a similar issue as https://github.com/whatwg/fetch/issues/1773.

shanehandley: Thanks Simon 👍 I will need to look into this - we should (at least) be evaluating this in step 12 of https://fetch.spec.whatwg.org/#concept-fetch however that's missing atm.

gterzian:
Thank you, I will review this tomorrow(anyone else wants to review it today, feel free to do so).

components/net/fetch/methods.rs
shanehandley: After some re-reading, it says in the in the spec for [request](https://fetch.spec.whatwg.org/#concept-request) that:

> "client" is changed to a policy container during fetching...

However there are two paths to [main_fetch](https://fetch.spec.whatwg.org/#concept-main-fetch): via [fetching](https://fetch.spec.whatwg.org/#fetching) as indicated above and in that note in the request spec, and via [redirect-fetch.](https://fetch.spec.whatwg.org/#http-redirect-fetch)

It's the latter case, it's much less clear how request gets it's policy container :/

components/script/dom/document.rs
```diff
@@ -3491,11 +3495,13 @@ impl Document {
     }
 
     pub fn set_csp_list(&self, csp_list: Option<CspList>) {
-        *self.csp_list.borrow_mut() = csp_list;
+        if let Some(list) = csp_list {
```
gterzian, line 3498: This is I think a bug, and perhaps the cause of the test failures, because the parser may set this back to none?

gterzian (changes requested):
Looks good overall, thank you! 

Let's see if we can address these new test failures...

components/net/fetch/methods.rs
simonwuelker: I believe [redirect-fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch) is only called after a [main-fetch](https://fetch.spec.whatwg.org/#concept-main-fetch), so the request should have received a policy container earlier already.

shanehandley: > I believe [redirect-fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch) is only called after a [main-fetch](https://fetch.spec.whatwg.org/#concept-main-fetch), so the request should have received a policy container earlier already.

Hmm, but `main_fetch` is the final step of `redirect_fetch` is it not?

components/script/dom/document.rs
shanehandley: Thank you Gregory 👍 I was very `fetch` focused when I looked through the spec but there are many places that the spec expects referrer policy to be provided during creation of a new Document or browsing context which I have missed.

As an example: https://html.spec.whatwg.org/multipage/document-sequences.html#creating-browsing-contexts

> Step 16.2: Set document's policy container to a clone of creator's policy container

Is this the intent of `CreatorBrowsingContextInfo`?

The other potential are I saw was https://html.spec.whatwg.org/multipage/document-lifecycle.html#initialise-the-document-object:

> Step 10: Let document be a new [Document](https://html.spec.whatwg.org/multipage/dom.html#document), with ... policy container: navigationParams's [policy container](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-policy-container)

However `navigationParams` isn't something that looks to be implemented in Servo.

components/net/fetch/methods.rs
simonwuelker: Right. The redirect fetch algorithm is executed once a response received by a fetch operation has a [redirect-status](https://fetch.spec.whatwg.org/#redirect-status) (and in some other cases, but always in response to an *existing* fetch operation). Internally it involves another main fetch, which reuses the same `fetchParams` as the original fetch, so the policy container is already initialized at this point.

At least that is how I read the spec.

gterzian:
Any ideas on the three new failures starting at https://github.com/servo/servo/pull/33977/files#diff-9e83d85b2d9e88c408b8ce8760d6248d6a123c61efe5ff06b1959f1a912b33c0 ? 

I guess it wasn't related to the change I suggested above after all?

shanehandley:
> Any ideas on the three new failures starting at https://github.com/servo/servo/pull/33977/files#diff-9e83d85b2d9e88c408b8ce8760d6248d6a123c61efe5ff06b1959f1a912b33c0 ?

After some experimentation, i've found any change the to referrer provided here (in isolation of any other changes) causes these to start failing:

https://github.com/servo/servo/pull/33977/files#diff-ecffb1e0970d813b7409f2c7fc8b17814044f30b7e7c6a634cb0d1a07933a54aL243-L246

It seems they were passing previously because `NoReferrerWhenDowngrade` was used as a fallback.

gterzian (approved):
Pre-approving this because I am on holiday for the rest of the week. I think it would be good to spend a bit of time trying to fix those newly failing tests, and if it is too much, file a detailed issue for a follow-up. @jdm can I relay this one to you please?

jdm:
Will do!

jdm:
I had hoped to have time to do some investigation into what the tests are testing for and why they now fail, but this has been a very busy week for childcare. In the interests of not holding up clear overall improvements, I'm going to merge this and take in the task of filing that followup. @shanehandley If you feel motivated and have the time to dive deeper into those failures, that would be lovely, but otherwise I will consider that my responsibility.

jdm:
Well, when this is rebased I am willing to.oress the merge button.

shanehandley:
> I had hoped to have time to do some investigation into what the tests are testing for and why they now fail, but this has been a very busy week for childcare. In the interests of not holding up clear overall improvements, I'm going to merge this and take on the task of filing that followup. @shanehandley If you feel motivated and have the time to dive deeper into those failures, that would be lovely, but otherwise I will consider that my responsibility.

Thank you, no worries at all 👍 I have been poking at the new test failures, and I can make these tests pass with 1-2 changed lines, however it caused other new failures which then required more time to debug 😬 

I'll certainly try to dig deeper into it over the weekend 👍

shanehandley:
Apologies @jdm I've fixed the WPT assertions (no new failures).

Issue #30566, opened 2023-10-17
Properties with discrete animation type shouldn't be transitionable


Loirooriol:
https://drafts.csswg.org/css-transitions-1/#transitionable

> the property values are transitionable if they have an [animation type](https://drafts.csswg.org/web-animations-1/#animation-type) that is neither [not animatable](https://drafts.csswg.org/web-animations-1/#not-animatable) nor [discrete](https://drafts.csswg.org/web-animations-1/#discrete).

https://drafts.csswg.org/css-transitions-2/#transition-behavior-property

> When `normal` is specified, transitions will not be started for discrete properties, only for interpolable properties.

Servo is currently transitioning discrete properties, like should happen with `transition-behavior: allow-discrete`, but `transition-behavior` is not supported and the initial value is `normal`.

PR #33991, opened 2024-10-24
Implement `transition-behavior`

Loirooriol:
Bump Stylo to https://github.com/servo/stylo/pull/84

---


mrobinson (commented):
How does this connect to our own animation code? Is it just that the values produced by the animation-driven restyle are updated differently?

Loirooriol:
Not sure if I understand the question. The logic in Stylo used to allow transitioning discretely-animatable properties (Gecko handles this outside of Stylo), so I prevented this in https://github.com/servo/servo/issues/30566

Now I'm just restoring to the old behavior of allowing interpolation of discrete properties for `transition-behavior: allow-discrete`.

mrobinson (approved):

PR #33992, opened 2024-10-24
Use `Rc<Box<[u8]>>` for queue to optimize get_in_memory_bytes 

Taym95:
---

### DIFF
--- a/components/script/dom/readablestreamdefaultcontroller.rs
+++ b/components/script/dom/readablestreamdefaultcontroller.rs
@@ -135,7 +135,7 @@ pub struct ValueWithSize {
 #[allow(crown::unrooted_must_root)]
 pub enum EnqueuedValue {
     /// A value enqueued from Rust.
-    Native(Vec<u8>),
+    Native(Rc<Box<[u8]>>),
     /// A Js value.
     Js(ValueWithSize),
 }
@@ -194,7 +194,7 @@ impl QueueWithSizes {
                         "`get_in_memory_bytes` can only be called on a queue with native values."
                     )
                 };
-                chunk.clone()
+                chunk.iter().copied()
             })
             .collect()
     }


components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -196,7 +196,10 @@ impl QueueWithSizes {
                         "`get_in_memory_bytes` can only be called on a queue with native values."
                     )
                 };
-                chunk.clone()
+                chunk
+                    .iter()
+                    .flat_map(|v| v.iter().cloned())
+                    .collect::<Vec<u8>>()
```
jdm, line 202: I'm having trouble understanding what we're optimizing here. Each chunk is a vector of boxed slices, and it looks like we're iterating over each byte and allocating a new vector that contains them all. What's the goal here?

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -137,7 +137,7 @@ pub struct ValueWithSize {
 #[allow(crown::unrooted_must_root)]
 pub enum EnqueuedValue {
     /// A value enqueued from Rust.
-    Native(Vec<u8>),
+    Native(Vec<Rc<Box<[u8]>>>),
```
simonwuelker, line 140: Can't this be `Vec<Rc<[u8]>>`?
`Rc<T>` imposes no `Sized` restriction on `T`

components/script/dom/readablestreamdefaultcontroller.rs
Taym95: Indeed there is not optimization yet, I just switched `Vec<Rc<Box<[u8]>>>`, will optimize next

Taym95: @jdm, I updated the code, and here is a code example of my reasoning. Please correct me if I am wrong Thanks

```

pub struct QueueWithSizes {
    queue: VecDeque<Vec<u8>>, // Each chunk is owned and duplicated
}

impl QueueWithSizes {
    pub fn new() -> Self {
        QueueWithSizes {
            queue: VecDeque::new(),
        }
    }

    pub fn enqueue(&mut self, chunk: Vec<u8>) {
        self.queue.push_back(chunk);
    }

    pub fn get_in_memory_bytes(&self) -> Vec<u8> {
        self.queue.iter().flat_map(|chunk| chunk.clone()).collect()
    }
}

pub struct QueueWithSizesB {
    queue: VecDeque<Rc<Box<[u8]>>>, // Chunks are shared, no duplication
}

impl QueueWithSizesB {
    pub fn new() -> Self {
        QueueWithSizesB {
            queue: VecDeque::new(),
        }
    }

    pub fn enqueue(&mut self, chunk: Vec<u8>) {
        let boxed_chunk = Rc::new(chunk.into_boxed_slice()); 
        self.queue.push_back(boxed_chunk); 
    }

    pub fn get_in_memory_bytes(&self) -> Vec<u8> {
        self.queue.iter().flat_map(|chunk| chunk.iter().copied()).collect()
    }
}

fn main() {
    let mut queue = QueueWithSizes::new();
    let mut queue_b = QueueWithSizesB::new();
    
    let chunk1 = vec![1, 2, 3];
    let chunk2 = vec![4, 5, 6];

    // Enqueue the same chunk multiple times in QueueWithSizes
    queue.enqueue(chunk1.clone()); 
    queue.enqueue(chunk1.clone()); 
    queue.enqueue(chunk2.clone()); 

    // Enqueue the same chunk only once in QueueWithSizesB
    queue_b.enqueue(chunk1.clone());
    queue_b.enqueue(chunk1.clone()); 
    queue_b.enqueue(chunk2.clone()); 

    let bytes = queue.get_in_memory_bytes();
    println!("{:?}", bytes); // Output: [1, 2, 3, 1, 2, 3, 4, 5, 6];

    let total_memory_usage = queue.queue.iter().map(|chunk| std::mem::size_of_val(chunk)).sum::<usize>();
    println!("Memory usage (QueueWithSizes): {} bytes", total_memory_usage); // print 72 bytes

    let bytes_b = queue_b.get_in_memory_bytes();
    println!("{:?}", bytes_b); // Output: [1, 2, 3, 1, 2, 3, 4, 5, 6];

    let total_memory_usage_b = queue_b.queue.iter().map(|chunk| std::mem::size_of_val(chunk)).sum::<usize>();
    println!("Memory usage (QueueWithSizesB): {} bytes", total_memory_usage_b); // print 24 bytes
}
```

simonwuelker: But this isn't measuring the total memory usage of either queue, you're summing up the `std::mem::size_of_val(chunk)` where chunk is either  `Vec<T>` or `Rc<T>`  - this is always equal to `24`/ `8` bytes per chunk (regardless of how big of that chunk is).

Taym95: Indeed, this make more sense`std::mem::size_of_val(chunk) + chunk.len() * std::mem::size_of::<u8>()`

components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -191,12 +191,12 @@ impl QueueWithSizes {
         self.queue
             .iter()
             .flat_map(|value| {
-                let EnqueuedValue::Native(chunk) = value else {
+                let EnqueuedValue::Native(chunks) = value else {
```
jdm, line 194: Since this is now only a single chunk again, let's avoid renaming it.

jdm (approved):
Ok, I think I follow what we're optimizing here. Previously, to get the bytes in memory we would iterate over a series of Vec chunks, clone them, them allocate a new Vec big enough for all of them combined, and fill it with the contents of the cloned chunks. Now, we still need the final allocation that is big enough everything, but the intermediate clones are just an Rc so we're not duplicating each chunk temporarily.

PR #34004, opened 2024-10-24
fix(constellation): pass resolved URL to embedder

### LINKED ISSUE #33876: Redirects are not reflected in the embedder history
**Describe the bug:**
The history sent to the embedder in the `EmbedderMsg::HistoryChanged(...)` message is not correct when the loaded page was redirected.

**To Reproduce:**

Launch Servo to a redirecting url, eg. https://google.com :

```shell
❯ RUST_LOG=info ./mach run -r https://google.com 
[2024-10-16T23:55:12Z INFO  servoshell::desktop::webview] Devtools Server running on port 6222
[2024-10-16T23:55:12Z WARN  constellation::constellation] Trying to get an event-loop from an unknown browsing context group
[2024-10-16T23:55:12Z WARN  constellation::constellation] Pipeline(1,1): Visibility change for closed pipeline
[2024-10-16T23:55:12Z WARN  webrender::device::gl] Cropping texture upload Box2D((0, 0), (0, 1)) to None
[2024-10-16T23:55:12Z WARN  webrender::device::gl] Cropping texture upload Box2D((0, 0), (0, 1)) to None
[2024-10-16T23:55:12Z WARN  webrender::device::gl] Cropping texture upload Box2D((0, 0), (0, 1)) to None
[2024-10-16T23:55:12Z INFO  net::cookie_storage]  === COOKIES SENT:
[2024-10-16T23:55:12Z INFO  net::http_loader] GET request for https://google.com/
[2024-10-16T23:55:12Z INFO  net::http_loader] got 301 response for "https://google.com/"
[2024-10-16T23:55:12Z INFO  net::cookie_storage]  === COOKIES SENT:
[2024-10-16T23:55:12Z INFO  net::http_loader] GET request for https://www.google.com/
[2024-10-16T23:55:12Z INFO  net::http_loader] got 200 response for "https://google.com/"

```

Observed:

The only `HistoryChanged` message sent to the embedder is:

`HistoryChanged ["https://google.com/"] current=0`

Expected:

The embedder receives `HistoryChanged ["https://www.google.com/"] current=0`

**Platform:**
Not platform specific.

rwakulszowa:
Constellation notifies the embedder about history changes. When we hit a URL that responds with a HTTP 301 response code, we want to replace the original URL with the final location.

The change boils down to reading the URL from
`Pipeline.url`, which is kept up to date, instead of from `Pipeline.load_data.url`, which stores the original request.

---

- [x] These changes fix #33876 (GitHub issue number if applicable)

- [x] These changes do not require tests because they only manifest in browsers built on Servo, not in Servo itself.

### DIFF
--- a/components/constellation/constellation.rs
+++ b/components/constellation/constellation.rs
@@ -4695,9 +4695,9 @@ where
     )]
     fn notify_history_changed(&self, top_level_browsing_context_id: TopLevelBrowsingContextId) {
         // Send a flat projection of the history to embedder.
-        // The final vector is a concatenation of the LoadData of the past
+        // The final vector is a concatenation of the URLs of the past
         // entries, the current entry and the future entries.
-        // LoadData of inner frames are ignored and replaced with the LoadData
+        // URLs of inner frames are ignored and replaced with the URL
         // of the parent.
 
         let session_history = match self.webviews.get(top_level_browsing_context_id) {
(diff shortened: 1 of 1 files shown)


rwakulszowa:
Tested locally with https://google.com and https://wikipedia.org (+ links on the main page). The browser displays the correct url, history navigation works fine.

One questionable part is using `LoadData.url` upon encountering a `NeedsToReload::Yes`, because (IIUC), we can't use `Pipeline.url` then. My understanding is that the `Yes` will eventually turn into a `No`, where we can use `Pipeline.url` again. That being said, it'd be cool if someone could double check.

rwakulszowa:
cc @webbeef

jdm (approved):
Based on your investigation in the issue, this seems like a reasonable change! Since we don't have any existing tests for this functionality that we can build upon, I'm fine with merging it as is.

jdm:
> One questionable part is using `LoadData.url` upon encountering a `NeedsToReload::Yes`, because (IIUC), we can't use `Pipeline.url` then. My understanding is that the `Yes` will eventually turn into a `No`, where we can use `Pipeline.url` again. That being said, it'd be cool if someone could double check.

This seems fine to me. The point of NeedsToReload::Yes is that we've thrown away everything except the original data that was first used to request the page, so I don't see a better outcome here.

PR #34008, opened 2024-10-25
ServoShell: Don't request redrawing again when processing a RedrawRequested event

webbeef:
Prevents requesting a new redraw when we are already processing that same event kind.

On both Linux and MacOS, this brings cpu usage down from 15%-20% to 0% when loading pages that have no animations, like `https://example.org` or `https://google.com` .

---


jdm (approved):
Excellent!

PR #34009, opened 2024-10-25
Limit HtmlOptionsCollection.length setter to 100k elements

simonwuelker:
This is a first step towards fixing a TIMEOUT in [`html/select/options-length-too-large.html`](https://wpt.fyi/results/html/select/options-length-too-large.html?label=master&product=servo) which is eating up 60s of wpt time.

That tests tries setting `options.length` to `Number.MAX_SAFE_INTEGER`. We now (correctly) bail out and refuse to add that many options. We still time out though, because later it tries to add 100.000 elements, which we should manage to do in time, but don't.

[try run](https://github.com/simonwuelker/servo/actions/runs/11518660903#summary-32065923732)

If you want to manually verify that this is correct you can run a reduced version of the test that now passes:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta name="timeout" content="long">
    <title>select options.length too large</title>

    <script src="http://wpt.live/resources/testharness.js"></script>
    <script src="http://wpt.live/resources/testharnessreport.js"></script>
  </head>
  <body>
    <select id="test">
      <option value="1"></option>
      <option value="2"></option>
      <option value="3"></option>
    </select>

    <script>
    var mySelect = document.getElementById("test");

    test(function() {
        mySelect.options.length = 100001;
        assert_equals(mySelect.options.length, 3, "Length of <select> should remain unchanged");
    });

    test(function() {
        mySelect.options.length = Number.MAX_SAFE_INTEGER;
        assert_equals(mySelect.options.length, 3, "Length of <select> should remain unchanged");
    });
    </script>
  </body>
</html>
```
---

### DIFF
--- a/components/script/dom/htmloptionscollection.rs
+++ b/components/script/dom/htmloptionscollection.rs
@@ -136,18 +136,31 @@ impl HTMLOptionsCollectionMethods for HTMLOptionsCollection {
 
     /// <https://html.spec.whatwg.org/multipage/#dom-htmloptionscollection-length>
     fn SetLength(&self, length: u32, can_gc: CanGc) {
-        let current_length = self.upcast().Length();
-        let delta = length as i32 - current_length as i32;
-        match delta.cmp(&0) {
+        // Step 1. Let current be the number of nodes represented by the collection.
+        let current = self.upcast().Length();
+
+        match length.cmp(&current) {
+            // Step 2. If the given value is greater than current, then:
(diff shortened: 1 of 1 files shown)


jdm (approved):

jdm:
The WPT testsuite and the HTML standard are always full of such interesting surprises.

PR #34019, opened 2024-10-26
GC hazard fix in customelementregistry.rs

### LINKED ISSUE #34017: GC borrow hazard in CustomElementRegistry::WhenDefined
Mutable borrow: https://github.com/servo/servo/blob/5d4b5514aa339d115d0ac9e1131eb7bef1d99a0b/components/script/dom/customelementregistry.rs#L602
Potential GC operation while the borrow is outstanding: https://github.com/servo/servo/blob/5d4b5514aa339d115d0ac9e1131eb7bef1d99a0b/components/script/dom/customelementregistry.rs#L606

We should be able to work around this by calling `self.when_defined.borrow()` and `self.when_defined.borrow_mut()` instead of using the `map` variable.

chickenleaf:
**file:** `servo/components/script/dom/customelementregistry.rs`

**PR Description**

**issue**: borrowed data stored in 'map' variable could cause issues during garbage collection

**changes**: replaced 'map' variable with direct `borrow()` and `borrow_mut()` calls on `self.when_defined`

**impact**: safer memory handling by not storing borrowed data in a variable

---

- [X] These changes fix #34017


jdm (approved):

chickenleaf:
this seems to be failing the MacOS build?

chickenleaf:
> this seems to be failing the MacOS build?

#34022 should fix that 🫡

components/script/dom/customelementregistry.rs
```diff
@@ -598,15 +598,17 @@ impl CustomElementRegistryMethods for CustomElementRegistry {
...
-        });
+        // Steps 3, 4, 5
+        let promise = self
+            .when_defined
+            .borrow()
+            .get(&name)
+            .cloned()
+            .unwrap_or_else(|| {
+                let promise = Promise::new_in_current_realm(comp, can_gc);
+                self.when_defined.borrow_mut().insert(name, promise.clone());
```
sagudev, line 609: already borrowed: BorrowMutError (thread Script(1,1), at components/script/dom/customelementregistry.rs:609)

borrow happens at 604

jdm (approved):
Verified in play.rust-lang.org that this pattern doesn't panic, unlike the previous change. The rules around when borrows end in Rust are quite subtle sometimes!

chickenleaf:
I've been trying to learn that 😭 I have more questions if I can't figure out from the internet I shall get back to u 😭

PR #34020, opened 2024-10-26
GC hazard fix in document.rs

### LINKED ISSUE #34016: GC borrow hazard in Document::GetElementsByTagNameNS and GetElementsByClassName
https://github.com/servo/servo/blob/main/components/script/dom/document.rs#L4472
https://github.com/servo/servo/blob/main/components/script/dom/document.rs#L4485

In both cases, we're mutably borrowing a hashmap entry, then performing an operation that can trigger a GC, then updating the hashmap. We should split that into separate borrows to avoid the hazard.

chickenleaf:
**File:** `components/script/dom/bindings.rs`

**PR Description:**

**Issue:**
- Fix for GC borrow hazard in `Document::GetElementsByTagNameNS` and `Document::GetElementsByClassName` methods.

**Change:**
- Refactored the code to split mutable borrows of the `HashMap` entries into separate operations, addressing the GC borrow hazard. 
  - Replaced the `match` on `entry()` pattern with a `get` + `insert` approach.
  - First checked for the entry's existence with an immutable borrow, then created the collection (which could trigger GC), followed by a separate mutable borrow to insert the new entry.

**Impact:**
- This change ensures that mutable borrows are not held while performing operations that could trigger garbage collection, eliminating potential hazards and maintaining the original functionality.

---

- [X] These changes fix #34016

### DIFF
--- a/components/script/dom/document.rs
+++ b/components/script/dom/document.rs
@@ -4469,28 +4469,28 @@ impl DocumentMethods for Document {
         let ns = namespace_from_domstring(maybe_ns);
         let local = LocalName::from(tag_name);
         let qname = QualName::new(None, ns, local);
-        match self.tagns_map.borrow_mut().entry(qname.clone()) {
-            Occupied(entry) => DomRoot::from_ref(entry.get()),
-            Vacant(entry) => {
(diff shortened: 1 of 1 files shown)


jdm (approved):
Looks reasonable. Thanks!

PR #34022, opened 2024-10-26
ci: pin python to 3.12 for mac workflows

mukilan:
wptrunner breaks macos builds because it doesn't work well with 3.13 due
to two issues:

1. The current version (10.3.0) of the 'pillow' dependency of wptrunner
   breaks with >=3.13 and needs to be upgraded.
2. Python 3.13 has removed the 'cgi' module which was deprecated in 3.11
   and has no direct replacement. There are two files in wptrunner that
   use the cgi module and one of them is the vendored 'html5lib'.

As a quick fix, pin the Python version on MacOS runner until
web-platform-tests/wpt#48585 is addressed.

---

### DIFF
--- a/.github/workflows/mac-wpt.yml
+++ b/.github/workflows/mac-wpt.yml
@@ -41,6 +41,11 @@ jobs:
       - uses: actions/download-artifact@v4
         with:
           name: ${{ inputs.profile }}-binary-macos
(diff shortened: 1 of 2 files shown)


sagudev (approved):

mrobinson (approved):

chickenleaf:
Ahhh I was wondering what was up!!! Thank you!

PR #34026, opened 2024-10-26
Properly transfer min/max constraints on auto-sized replaced elements

Loirooriol:
We were following CSS2, which didn't handle `aspect-ratio`. This patch simplifies the logic and handles it correctly.

Unfortunately this makes 2 tests fail, but I'm pretty sure they aren't spec-compliant. I'm leaving them as-is for now since they are part of interop-2021, and Gecko, Blink and WebKit pass them (because of some non-interoperable incorrect behaviors).

I'm adding a new test that is fully passed by Servo and WebKit.

---

### DIFF
--- a/components/layout_2020/replaced.rs
+++ b/components/layout_2020/replaced.rs
@@ -5,7 +5,7 @@
 use std::fmt;
 use std::sync::{Arc, Mutex};
 
-use app_units::Au;
+use app_units::{Au, MAX_AU};
 use base::id::{BrowsingContextId, PipelineId};
 use canvas_traits::canvas::{CanvasId, CanvasMsg, FromLayoutMsg};
 use data_url::DataUrl;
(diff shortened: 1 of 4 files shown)


Loirooriol:
More specifically, the problem that this is fixing is:

```html
<!DOCTYPE html>
<style>canvas { aspect-ratio: 1; border: solid }</style>
<canvas width="100" height="50" style="min-height: 50px"></canvas>
<canvas width="100" height="50" style="min-height: 51px"></canvas>
```



It doesn't make sense for the 2nd canvas to become narrower due to a larger `min-height`, even if Blink also has this behavior.

components/layout_2020/replaced.rs
```diff
@@ -554,119 +554,34 @@ impl ReplacedContent {
...
-                    }
-                },
-            };
+            let tentative_size = get_tentative_size(box_size);
+            let max_box_size = max_box_size.map(|max_size| max_size.unwrap_or(MAX_AU));
+            // This is a simplification of the CSS2 algorithm in a way that properly handles `aspect-ratio`.
+            // We transfer min and max constraints from the other axis, and apply them in addition to
+            // non-transferred min and max constraints. In case of conflict,
+            //  - Non-transferred constraints take precedence over transferred ones.
+            //  - Min constraints take precedence over max ones from the same axis.
```
mrobinson, line 563: Do you mind dropping a link to the relevant spec issue or comment here?

mrobinson (approved):

components/layout_2020/replaced.rs
Loirooriol: Linked to https://drafts.csswg.org/css-sizing-4/#aspect-ratio-size-transfers

mrobinson: Thanks! Do you think a link to <https://github.com/w3c/csswg-drafts/issues/6071#issuecomment-2243986313> makes sense as well?

Loirooriol: Sure, done

PR #34028, opened 2024-10-26
webgpu: Move supported context format to content timeline

sagudev:
Relevant spec change: https://github.com/gpuweb/gpuweb/pull/4911 (also done by me).

draft because it needs updated expectations

---

- [x] There are tests for these changes in WebGPU CTS (actually I used this commit to create changes for CTS that we now finally got)


gterzian (approved):
LGTM

PR #34034, opened 2024-10-28
Implement `SubtleCrypto.digest`

simonwuelker:
This implements https://w3c.github.io/webcrypto/#SubtleCrypto-method-digest.

The implementation is fairly straightforward, I think the only change that may warrant discussion is the inclusion of [`ring`](https://crates.io/crates/ring) in the `script` crate. [`ring`](https://crates.io/crates/ring) is the most popular crate for cryptography [crates.io](https://crates.io/keywords/cryptography) and I think we're going to need it for future SubtleCrypto things anyways.

With this change, 100% of the web platform tests in `WebCryptoAPI/digest/` start to pass.

[try run](https://github.com/simonwuelker/servo/actions/runs/11550437751#summary-32147313879)

---


mrobinson:
Oh, I forgot to mention the thing about `ring` that makes me have doubts is its very weird license.

simonwuelker:
Yeah those are good points, I switched to using `aws-lc-rs` instead of `ring`. (Very easy because the two are compatible)

jdm (approved):
Looks good! This can merge after updating servo-tidy.toml to allow the duplicate crate version.

jdm:
Looks like our windows build setup needs to include nasm:
```
  --- stderr
  Consider installing NASM or setting `AWS_LC_SYS_PREBUILT_NASM` in the build environment.See User Guide: https://aws.github.io/aws-lc-rs/index.html
  Missing dependency: nasm
  thread 'main' panicked at C:\Users\servo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\aws-lc-sys-0.22.0\builder/main.rs:320:40:
  called `Result::unwrap()` on an `Err` value: "Required build dependency is missing. Halting build."
``|

**Comment by simonwuelker** [comment]
I'm wondering, would it be fine to set `AWS_LC_SYS_PREBUILT_NASM=1` instead of installing nasm? That would use the prebuilt binaries provided by the crate instead of rebuilding them. It's easier and would (probably?) slow down windows builds less, though the dependency is less auditable this way.

**Comment by jdm** [comment]
Mmm, build issues on both Windows and OpenHarmony. Fun. @jschwe Can you advise on the OpenHarmony error?

**Comment by jschwe** [comment]
Ugg - Now we again have a dependency that uses CMake :(. 

There are a couple issues as far as I can see:
1. We need to use the cmake version from the OpenHarmony NDK (or the latest upstream CMake). Easy to fix, I will open a PR.
2. the `aws-lc-sys`  build-script apparently sets `-Werror`, which causes the build to fail due to a warning about an unused parameter, which is set by the `Clang` Module of CMake. We can fix that by setting a CFLAG telling clang to ignore that error (ugly).
3. The CMake code of `aws-lc` apparently assumes tge `Thread` module will be found and errors out if that is not the case. I believe not just on OpenHarmony, but on most modern Linux systems linking `-lpthread` is not necessary anymore and may cause errors, so perhaps the upstream build CMake will need to be adapted. I'll have a closer look tommorow.

```
  CMake Error at /home/jschwender/tools/command-line-tools/sdk/default/openharmony/native/build-tools/cmake/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
    Could NOT find Threads (missing: Threads_FOUND)
  Call Stack (most recent call first):
    /home/jschwender/tools/command-line-tools/sdk/default/openharmony/native/build-tools/cmake/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
    /home/jschwender/tools/command-line-tools/sdk/default/openharmony/native/build-tools/cmake/share/cmake-3.28/Modules/FindThreads.cmake:226 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
    aws-lc/crypto/CMakeLists.txt:707 (find_package)
```

mrobinson:
I wasn't aware that `aws-lc-rc` used CMake to compile itself. That's unfortunate! Another option is to backtrack and use `ring` for now and just switch to `aws-lc-rc` once we have upgraded `rustls`. I don't want this choice to cause too many issues right now.

simonwuelker:
> Another option is to backtrack and use ring for now and just switch to aws-lc-rc once we have upgraded rustls.

Let's try that then. I'm not too keen on having to debug windows CI issues if I can avoid it :P

PR #34035, opened 2024-10-28
Add `medium` cargo profile for daily usage

jschwe:
- The standard `dev` mode is often too slow for daily work on servo
- Increasing the optimization level to 2 improves the performance noticably.
- Build time for opt-level = 1 is slightly higher than 2 on my machine
- Reducing debug info to `line-tables-only` improves link and build times,
  while still keeping relevant info for backtraces.

Also extend .servobuild to allow setting custom cargo profiles instead
of just release and dev.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,6 +162,14 @@ xml5ever = "0.20"
 opt-level = 3
 debug-assertions = true
 
+# A profile between `dev` and `release` which aims to offer a compromise between
+# fast incremental rebuilds and runtime speed.
+[profile.medium]
+inherits = "release"
+opt-level = 2
+incremental = true
+debug = "line-tables-only"
+
 [profile.production]
 inherits = "release"
 debug-assertions = false
--- a/servobuild.example
+++ b/servobuild.example
@@ -14,6 +14,8 @@
 [build]
 # Set "mode = dev" or use `mach build --dev` to build the project with warning.
 # or Set "mode = release" or use `mach build --release` for optimized build.
+# Use `mode = <profile>` or `mach build --profile=<profile>` to build the given
+# profile. Check the `Cargo.toml` manifest for a complete list of custom profiles.
 # Defaults to prompting before building
 #mode = "dev"
 
--- a/python/servo/command_base.py
+++ b/python/servo/command_base.py
@@ -678,6 +678,10 @@ class CommandBase(object):
             elif self.config["build"]["mode"] == "release":
                 print("No build type specified, but .servobuild specified `--release`.")
                 return BuildType.release()
+            elif self.config["build"]["mode"] != "":
+                profile = self.config["build"]["mode"]
+                print(f"No build type specified, but .servobuild specified custom profile `{profile}`.")
+                return BuildType.custom(profile)
             else:
                 print("No build type specified so assuming `--dev`.")
                 return BuildType.dev()


nicoburns (approved):
This seems useful. A couple of suggestions:

- Use `opt-level = 1` or `opt-level = 2` (default for release is 3). These (esp. 2) are generally fast enough for servo development (at least in my testing), but improve compile times.
- Use `debug = "line-tables-only"`. This will generate minimal debug info for file/line numbers when a panic occurs.

mrobinson:
I mean, if these options are useful for normal development, maybe they should be the default and then we can add a non-incremental build for when people need to prevent trouble. We already have a `production` profile and a `debug` profile.

@delan, thoughts here?

mrobinson:
I'm imagining four types of build:

1. Everyday development
2. Every development (non-incremental)
3. All debug options on
4. Production

nicoburns:
Merging this as it seems useful as-is (and doesn't affect anyone's existing workflows in it's current form). We can iterate and consider making this the default in future.

Cargo.toml
```diff
@@ -162,6 +162,14 @@ xml5ever = "0.20"
 opt-level = 3
 debug-assertions = true
 
+# Servo is slow in debug mode, so for daily testing we provide a devel profile
+# which is optimized, but has incremental building enabled for faster compile times
+[profile.devel]
```
delan, line 167: this name is too similar to cargo’s `dev` (which is already a confusing can of worms due to the whole `[profile.dev]` vs `target/debug` thing). what about something like `dev-fast`?

delan (changes requested):
Thanks! Remember to update the book’s [section about build profiles](https://book.servo.org/hacking/building-servo.html#build-profiles) :)

One change, aside from that I think we’re good to go. The rest of my comments below don’t block this patch.

> Use `debug = "line-tables-only"`. This will generate minimal debug info for file/line numbers when a panic occurs.

What’s the motivation for this compared to `full`, which is what we use in `dev`? Does it make the build faster?

> Probably the biggest incompatibility is with `sccache`, since sccache doesn't work with incremental builds. Adding a new profile is the safest option, since changing release to incremental would likely "break" users that have sccache configured (in the sense that sccache won't cache any Rust crates anymore).

When we use sccache in CI, we set `CARGO_INCREMENTAL`, which may help with that without needing an extra profile.

nicoburns:
> > Use `debug = "line-tables-only"`. This will generate minimal debug info for file/line numbers when a panic occurs.
> 
> What’s the motivation for this compared to `full`, which is what we use in `dev`? Does it make the build faster?

I haven't measured, but yes it should do. "full" can produce several hundred megabytes of debuginfo for large projects, which dwarfs the size of the binary itself, and impacts both compile and link times. The cargo team is actually considering changing the default from 2 to 1 for the built-in dev/debug profile.


> what about something like dev-fast?

I'd prefer something short and without a hyphen if possible, as I figure I'm going to be typing this a lot. Maybe `fdev`?

delan:
> I'd prefer something short and without a hyphen if possible, as I figure I'm going to be typing this a lot. Maybe `fdev`?

`fastdev`? Otherwise `fdev` works for me.

mrobinson:
I agree with @delan's idea of making this a "more optimized" version of the `dev` profile, as that makes the name of the profile more understandable. Another option could be `dev-optimized` or `dev-opt` for an optimized `dev` build.

jschwe:
>>  Use debug = "line-tables-only". This will generate minimal debug info for file/line numbers when a panic occurs.
>
> What’s the motivation for this compared to full, which is what we use in dev? Does it make the build faster?

In addition to what nico said, my common workflow for openharmony also includes:
- Potentially copying the shared library from a remote build server to my machine
- packaging the app
- pushing the app to a device over usb
- installing the app

This means binary size is quite relevant for the build + test cycle, since the rust artifact is copied multiple times in the process. The `dev` profile gives an artifact that is over 2GB large!

I would say that the close names `dev` / `devel` are perhaps not that relevant, since the usage would be 
`--dev` vs `--profile=devel` which seems distinct enough. 

If we are discussing names though, then I guess it would also make sense to revisit Martins proposal of making this the default settings for `dev` and instead provide a custom profile for debugging with gdb/lldb. 
`debug` is probably not possible since it conflicts with the target directory name of the `dev` profile, but perhaps `debugger` / `debugging` or `gdb` would work?

delan:
I think any name that only differs from `dev` or `debug` by stemming or abbreviation would be too confusing for the new profile, regardless of whether the new profile is a faster `dev` or a more debuggable `release`.

Would `fastdev` be ok?

mrobinson:
> If we are discussing names though, then I guess it would also make sense to revisit Martins proposal of making this the default settings for `dev` and instead provide a custom profile for debugging with gdb/lldb.
> `debug` is probably not possible since it conflicts with the target directory name of the `dev` profile, but perhaps `debugger` / `debugging` or `gdb` would work?

Yeah, I am also okay with turning our current `dev` profile into `dev-full` or `dev-chonky` if this new profile is good enough for most debugging / development.

jschwe:
Updated the PR to modify the standard `dev` profile instead, and provide a `dev-full` custom profile which corresponds to the standard options from cargos `dev` profile, but is intended to be used when debugging only.

> Thanks! Remember to update the book’s [section about build profiles](https://book.servo.org/hacking/building-servo.html#build-profiles) :)

I'll open a PR to the book after this has been merged, since there still might be more discussions to be had.

nicoburns:
Do we have any data on compile times with this new configuration vs. the old one? That might good to get if we're going to make it the default. I would guess the increased opt level will make it slower, but the debug info change should speed it up. Not sure what (or how big) the overall effect will be.

servobuild.example
```diff
@@ -14,6 +14,9 @@
 [build]
 # Set "mode = dev" or use `mach build --dev` to build the project with warning.
 # or Set "mode = release" or use `mach build --release` for optimized build.
+# Use `mode = devel` or `mach build --profile=devel` for an optimized build
+# with incremental compilation for faster compile times.
+# Check the `Cargo.toml` manifest for a complete list of custom profiles.
```
mrobinson, line 19:
```suggestion
# Use `mode = <profile>` or `mach build --profile=<profile>`  to build the given
# profile. Check the `Cargo.toml` manifest for a complete list of custom profiles.
```

mrobinson (approved):

jdm:
> > Do we have any data on compile times with this new configuration vs. the old one? That might good to get if we're going to make it the default.
> 
> Is it a common use case to build and run servo in `dev` mode?

99% of my builds are dev builds. My priorities are debugging symbols and build speeds.

jschwe:
I got pre-empted with other things, but now I've got a bit of time to come back to this. 

> 99% of my builds are dev builds. My priorities are debugging symbols and build speeds.

Thanks for commenting! I dropped this idea and went back to a custom profile.

Regarding performance - I tried both opt-level  1 and 2 and at least on a 5995WX Linux system binary size and build time are almost identical:

### Opt-level 1

- Size: 479MB
- Clean build: 2:06 min
- Incremental after adding some whitespace to script: 0:31

### Opt-level 2
- Size: 487 MB
- Clean build: 1:56
- Incremental: 0:29 

It was a bit surprising to me that builds are slightly faster with O2 than with O1, but I guess that probably varies by machine.

nicoburns:
This is actually slower a release build for the initial build for me, but a rebuild is quite a bit quicker. The price of incremental compilation I guess.

I would also note that a lot the problem here seems to be the script crate. Changes to `layout_2020` are much faster. Splitting up script (which should allow the compilation) probably ought to be a priority if we want faster builds.

### Release build
Clean build: 4m 03s
Incremental (script): 1m 25s
Incremental (layout_2020): 16.5s

### Medium build
Clean build: 4m 23s 
Incremental (script): 52s
Incremental (layout_2020): 6.8s

Very jealous of your 5995WX with it's 2 minute clean build times!

P.S. Some weird crate rebuilds for the release profile incremental rebuild. Can't work out why `tungstenite` would rebuild for a whitespace change to script.

delan (approved):

PR #34038, opened 2024-10-28
servoshell: Allow overriding screen resolution with a command-line argument

mrobinson:
There is a command-line argument to override the default window size,
but not one for overriding the default screen resolution. This is
important for testing pages that use screen size to have different
behavior.

In addition to adding the new option this change:

 - Renames the `--resolution` command-line argument to `--window-size`
   to remove ambiguity with the `--screen-size` argument.
 - Passes the screen size as device independent (device pixels scaled by
   HiDPI factor) to Servo internals. Not only it make it simpler to pass
   the `--window-size` override, it makes more sense. Different screens
   can have different HiDPI factors and these can be different from the
   scale of the window. This makes the screen HiDPI factor totally
   independent of the one that Servo uses for the window.

---

- [x] These changes do not require tests because this change adds an interactive configuration option to servoshell. We do not test this kind of behavior.

### DIFF
--- a/components/config/opts.rs
+++ b/components/config/opts.rs
@@ -83,6 +83,10 @@ pub struct Opts {
     /// The initial requested size of the window.
     pub initial_window_size: Size2D<u32, DeviceIndependentPixel>,
 
+    /// An override for the screen resolution. This is useful for testing behavior on different screen sizes,
+    /// such as the screen of a mobile device.
+    pub screen_size_override: Option<Size2D<u32, DeviceIndependentPixel>>,
+
     /// Whether we're running in multiprocess mode.
     pub multiprocess: bool,
 
@@ -407,6 +411,7 @@ pub fn default_opts() -> Opts {
         devtools_server_enabled: false,
         webdriver_port: None,
         initial_window_size: Size2D::new(1024, 740),
+        screen_size_override: None,
         multiprocess: false,
         background_hang_monitor: false,
         random_pipeline_closure_probability: None,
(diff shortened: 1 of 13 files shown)


nicoburns:
What's the difference between this and using the browser zoom to test such pages? I believe that also effectively acts like an additional scale factor between "user code" and the logical window size. Is the advantage here that you can specify an exact resolution?

ports/servoshell/desktop/headed_window.rs
```diff
@@ -72,11 +72,20 @@ impl Window {
         // #9996.
         let visible = opts.output_file.is_none() && !no_native_titlebar;
 
+        // Do not let the window size start larger than the screen size override.
```
mukilan, line 75: Is this constraint strictly required? Is it not valid to create a window wider than the screen, for example, to span two monitors in extended display configuration, or is this just for convenience (so we don't have to pass --window-size for simulation of mobile screens)?

mukilan (approved):

mukilan:
> What's the difference between this and using the browser zoom to test such pages? I believe that also effectively acts like an additional scale factor between "user code" and the logical window size. Is the advantage here that you can specify an exact resolution?

The goal here is to allow overriding the value returned from `window.screen.width` which some websites use to detect mobile user agents and redirect to desktop mode if that value is greater than some threshold. In other browsers like FF, the 'responsive design mode'  changes the value of `window.screen.size` to handle such cases.

ports/servoshell/desktop/headed_window.rs
mrobinson: It's not strictly required, but it makes setting the screen size a lot friendlier as you also don't have to modify the window size. I've changed this code so it only affects the default window size. This allows you to specify a window size larger than the screen, if you do it manually.

components/config/opts.rs
```diff
@@ -499,7 +504,18 @@ pub fn from_cmdline_args(mut opts: Options, args: &[String]) -> ArgumentParsingR
         "Start remote WebDriver server on port",
         "7000",
     );
-    opts.optopt("", "resolution", "Set window resolution.", "1024x740");
+    opts.optopt(
+        "",
+        "window-size",
+        "Set the initial window size to the scaled (device independent) size",
```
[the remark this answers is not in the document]
mrobinson: I've gone ahead and changed this to refer to "logical" pixels.

PR #34052, opened 2024-10-29
Enable css-grid WPT tests

nicoburns:
Changes made:

- Enable `css/css-grid` WPT tests for layout2020
- Update test assertions for these tests

Does layout2013 also need these to be enabled?

---


Loirooriol:
> Does layout2013 also need these to be enabled?

I think this is already enabling for both 2020 and 2013. We don't really check 2013 results other than when importing new tests, but if you can include them here then better.

mrobinson:
It isn't necessary to generate results for Layout 2013, as they will be updated when the weekly WPT job happens.

Loirooriol:
Dang, conflict with the tags

nicoburns:
The failing test seems to be unreliable. Sometimes it times out, sometimes it completes. What's the procedure for such tests?

mrobinson:
> The failing test seems to be unreliable. Sometimes it times out, sometimes it completes. What's the procedure for such tests?

Usually when this is the case the test is marked as having both possible results with a WPT `.ini` file. On the other hand, without incremental layout we typically don't expect this kind of non-deterministic behavior in layout. Do you have a sense for what might causing the flaky behavior?

Loirooriol:
I think `expected: [TIMEOUT, OK]` in the `.ini` does the trick, or you could file an issue with `I-intermittent`.

mrobinson:
> Usually when this is the case the test is marked as having both possible results with a WPT `.ini` file. On the other hand, without incremental layout we typically don't expect this kind of non-deterministic behavior in layout. Do you have a sense for what might causing the flaky behavior?

Oh, a case this might happen in layout is if the test case is very heavy and the algorithm used for layout is very slow.

nicoburns:
> Oh, a case this might happen in layout is if the test case is very heavy and the algorithm used for layout is very slow.

Hmm... this is a CSS Grid test, but of course Servo doesn't have grid support yet, so I'd expect this to be falling back to block layout, which shouldn't be slow? This test does have quite a few test cases (script driven) though.

nicoburns:
> I think `expected: [TIMEOUT, OK]` in the `.ini` does the trick

Thanks! That does seems to have done the trick :)

nicoburns:
@mrobinson Is anything still blocking this one? CI / WPT is passing now.

mrobinson (approved):

PR #34055, opened 2024-10-29
Import `size_of_test` from servo/stylo repo

nicoburns:
This is a small utility crate that was previous in the `servo/stylo` repo. But it is only used by crates in this repo. So moving it here. PR to remove it from `servo/stylo` is https://github.com/servo/stylo/pull/87

- [x] These changes do not require tests because it's just moving code. So if it compiles it's very unlikely that test results would change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -346,7 +346,6 @@ dependencies = [
  "malloc_size_of_derive",
  "parking_lot",
  "serde",
- "size_of_test",
  "webrender_api",
 ]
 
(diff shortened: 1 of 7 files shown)


mrobinson (changes requested):
Nice. Instead of making this a new crate though, let's remove the uses in legacy `layout` and just make this a free function in `components/shared/base/id.rs`.

mrobinson (approved):
Thank you!

PR #34063, opened 2024-10-30
servoshell: Properly convert to logical pixels when scrolling with a `PixelDelta`

### LINKED ISSUE #34051: Scrolling is slow
**Describe the bug:**
Go to any website that is has a reasonably long. For example https://en.wikipedia.org/wiki/Internet. Scrolling that page is very slow, and that makes Servo seem slow in general.

**To Reproduce:**
Scroll the page. Compare the scrolling speed with scrolling the same page in Chrome

**Platform:**
Tested on macOS 14.6.1, but I suspect this affects all platforms (although it may be particularly obvious when scrolling with a trackpad)

nicoburns:
This changes servoshell to use scroll pixel deltas received from Winit as-is without converting from Physical to Logical coordinates. This much more closely perceptually matches the scroll speed of other apps on my macOS system. I have not tested this on other platforms.

Fixes #34051.

---

- [x] These changes fix #34051

### DIFF
--- a/ports/servoshell/desktop/headed_window.rs
+++ b/ports/servoshell/desktop/headed_window.rs
@@ -24,7 +24,7 @@ use servo::webrender_api::units::{DeviceIntPoint, DeviceIntRect, DeviceIntSize};
 use servo::webrender_api::ScrollLocation;
 use servo::webrender_traits::RenderingContext;
 use surfman::{Connection, Context, Device, SurfaceType};
-use winit::dpi::{LogicalPosition, LogicalSize, PhysicalPosition, PhysicalSize};
+use winit::dpi::{LogicalSize, PhysicalPosition, PhysicalSize};
 use winit::event::{ElementState, KeyEvent, MouseButton, MouseScrollDelta, TouchPhase};
 use winit::keyboard::{Key as LogicalKey, ModifiersState, NamedKey};
 #[cfg(any(target_os = "linux", target_os = "windows"))]
@@ -423,8 +423,8 @@ impl WindowPortsMethods for Window {
                         (dx as f64, (dy * LINE_HEIGHT) as f64, WheelMode::DeltaLine)
                     },
                     MouseScrollDelta::PixelDelta(position) => {
-                        let position: LogicalPosition<f64> =
-                            position.to_logical(self.device_hidpi_factor().get() as f64);
+                        let scale_factor = self.device_hidpi_factor().inverse().get() as f64;
+                        let position = position.to_logical(scale_factor);
                         (position.x, position.y, WheelMode::DeltaPixel)
                     },
                 };


ports/servoshell/desktop/headed_window.rs
```diff
@@ -415,8 +415,6 @@ impl WindowPortsMethods for Window {
                         (dx as f64, (dy * LINE_HEIGHT) as f64, WheelMode::DeltaLine)
                     },
                     MouseScrollDelta::PixelDelta(position) => {
-                        let position: LogicalPosition<f64> =
-                            position.to_logical(self.device_hidpi_factor().get() as f64);
```
mrobinson, line 419: Okay. After a bit of investigation, I think I know what is going on here. The value expected by Servo is a value in logical pixels (`LayoutVector2D`). The value we are getting from `winit` is in physical pixels, so we should do a conversion. The issue here is that the factor returned by `device_hidpi_factor()` is the scale to convert from logical pixels to device pixels and `fn to_logical` wants the opposite (because it is dividing). I believe the correct thing to do here is to use the inverse of the factor. I've confirmed that this doesn't speed up scrolling too much and it looks a lot more natural:

```rust
                            position.to_logical(self.device_hidpi_factor().inverse().get() as f64);
```

In addition, this probably won't change behavior on Linux users with a mouse, because that goes through the `LineDelta` code path above.

mrobinson (changes requested):

mrobinson:
Nice find, by the way!

ports/servoshell/desktop/headed_window.rs
nicoburns: Ah, so scrolling would actually have been 4x too slow on my device. No wonder it felt slow! I've updated this PR to use the inverse scale factor :)

nicoburns: > In addition, this probably won't change behavior on Linux users with a mouse, because that goes through the LineDelta code path above.

Yeah, I think fancy mice with "high precision" scrolling might also use the PixelDelta, but most mice are using the LineDelta.

mrobinson (approved):

ports/servoshell/desktop/headed_window.rs
mrobinson: I suspect there is also an issue with the line delta, because when modifying the screen scaling, it scrolls different amounts even though this theoretically should be the same number of logical pixels. I think more investigation is necessary here.

nicoburns: > I suspect there is also an issue with the line delta, because when modifying the screen scaling, it scrolls different amounts even though this theoretically should be the same number of logical pixels. I think more investigation is necessary here.

Seems plausible. I don't think servoshell got much attention until recently. Unfortunately I don't have device a mouse available to test that with atm, but perhaps we could create an issue.

yezhizhen: Hi I verified a while, and believe the original code is correct, even tho it is much slower with `PixelDelta` for a mouse in high precision mode...
(which is intended)

`device_hidpi_scale_factor` is same as `devicePixelRatio` in JS console, which is 1.5 for me in all browsers.
`to_logical` needs to divide by $1.5$ to convert to Logical Pixel.

I believe we shouldn't pass the inverse after digging a while with https://github.com/servo/servo/pull/37893.
Can you guys check again? @nicoburns @mrobinson

mrobinson: Should we modify the size LINE_DELTA based on the `device_hidpi_scale_factor` then? Simply undoing this change would make servoshell scroll very slowly on macOS.

yezhizhen: I think so. Just revert and then multiply by another constant for `PixelDelta`. I cannot trigger `PixelDelta` on Windows.

Also, I think for `LineDelta`, we need to multiply `dx` by a `LINE_WIDTH`  as well. Horizontal scroll is very slow comparing to vertical. For other browsers, they are basically the same.

PR #34071, opened 2024-10-30
webgpu: Use IpcSharedMemory for mapped buffers

sagudev:
with https://github.com/servo/ipc-channel/pull/356 this allows us to reuse allocation of IpcSharedMemory.

I made webgpu map_unmap bench: https://sagudev.github.io/briefcase/webgpu_map_speed.html?write=true. Still need to do exact measurements but results are good.

---


sagudev:
I still need to do performance checking.

PR #34072, opened 2024-10-30
mach: fix test-speedometer

sagudev:
Split from https://github.com/servo/servo/pull/33261

---

- [x] These changes do not require tests because they are for tests

### DIFF
--- a/python/servo/testing_commands.py
+++ b/python/servo/testing_commands.py
@@ -529,8 +529,8 @@ class MachCommands(CommandBase):
             binary,
             "https://servospeedometer.netlify.app?headless=1",
             "--pref", "dom.allow_scripts_to_close_windows",
-            "--resolution=1100x900",
-            "--headless"], timeout=60).decode())
+            "--window-size=1100x900",
+            "--headless"], timeout=120).decode())
 
         print(f"Score: {speedometer['Score']['mean']} ± {speedometer['Score']['delta']}")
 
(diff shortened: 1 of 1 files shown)


python/servo/testing_commands.py
```diff
@@ -414,6 +414,22 @@ def test_dromaeo(self, tests, servo_binary: str, bmf_output: str | None = None):
...
+        if bmf_output:
+            with open(bmf_output, 'w', encoding='utf-8') as f:
+                json.dump({
+                    'servo': {
+                        'file-size': {
+                            'value': float(size),
+                        }
+                    }
+                }, f, indent=4)
+
```
mrobinson, line 432: Is the BMF file format defined somewhere?

python/servo/devenv_commands.py
```diff
@@ -270,3 +271,18 @@ def ndk_gdb(self, release, target):
...
+        output: dict[str, object] = dict()
+        for input_file in inputs:
+            with open(input_file, 'r', encoding='utf-8') as f:
+                output = json.load(f) | output
+
+        if bmf_output:
+            with open(bmf_output, 'w', encoding='utf-8') as f:
+                json.dump(output, f, indent=4)
+        else:
+            print(json.dumps(output, indent=4))
```
mrobinson, line 288: Where is this merge command used? It feels almost a bit too specific for a particular task and also this is a few lines of Python. It's hard to know if this is appropriate for `mach` without having the full context.

mrobinson (changes requested):
Can you remove the two new commands and just fix `speedometer_runner` in this PR?

python/servo/testing_commands.py
sagudev: https://bencher.dev/docs/reference/bencher-metric-format/

python/servo/devenv_commands.py
sagudev: TBH I was also thinking the same thing, maybe this and test-size should rather go to etc/ci a part of https://github.com/servo/servo/pull/33261

mrobinson: Yeah, that seems reasonable! That way the scripts will be available if / when we stop using `mach`.

mrobinson (approved):

PR #34073, opened 2024-10-30
chore: Update wgpu to v23

sagudev:
This version was created from trunk today, so let's use it, but we will hop back to trunk as soon as it will be needed.

---


jdm (approved):

sagudev:
This still needs new expectations.

sagudev:
Ah, apparently there are no expectations changes needed: https://github.com/sagudev/servo/actions/runs/11597312350/job/32295964422

PR #34076, opened 2024-10-30
Unify sizing logic for replaced elements

Loirooriol:
The logic varied quite a bit depending on the case, now it's unified.

This also fixes the following case where the iframe was 150px tall instead of 50px:
```html
<iframe style="min-width: 400px; max-height: 50px"></iframe>
```

This also modifies video-intrinsic-width-height.html to expect the new behavior that we share with Blink and WebKit. In fact WebKit already modified this test but forgot to export the change upstream. Firefox is different but it was already failing anyways.

---


mrobinson (commented):
What a great simplification!

mrobinson (approved):

PR #33531, opened 2024-09-24
Bring `http_network_or_cache_fetch` closer to the spec

simonwuelker:
This adds spec steps for the [ `http_network_or_cache_fetch`](https://fetch.spec.whatwg.org#http-network-or-cache-fetch) algorithm and implements the missing [`append_a_request_origin_header`](https://fetch.spec.whatwg.org/#append-a-request-origin-header) method, fixing a few WPT subtests in the process.

Draft because I would like to get https://github.com/whatwg/fetch/pull/1776 merged first and update the spec steps accordingly, but apart from that this is ready for review.

[try run](https://github.com/simonwuelker/servo/actions/runs/11013561747)

---

- [X] `./mach build -d` does not report any errors
- [X] `./mach test-tidy` does not report any errors
- [X] There are tests for these changes

PR #34081, opened 2024-10-31
net: Ensure that origin serialization is consistent

mrobinson:
A recent refactoring (#33531) made a change that resulted in the
`Origin` header including the port even when the default port for a
scheme was used. This made the serialization different from that used
for `rust-url`'s `Origin::ascii_serialization()`, breaking CORS on some
sites. This change makes it so that the serialization is consistent
again.

---

### DIFF
--- a/components/net/http_loader.rs
+++ b/components/net/http_loader.rs
@@ -125,7 +125,7 @@ impl Default for HttpState {
 }
 
 /// Step 13 of <https://fetch.spec.whatwg.org/#concept-fetch>.
-pub fn set_default_accept(request: &mut Request) {
+pub(crate) fn set_default_accept(request: &mut Request) {
     if request.headers.contains_key(header::ACCEPT) {
         return;
     }
@@ -330,7 +330,7 @@ pub fn determine_requests_referrer(
     }
 }
 
-pub fn set_request_cookies(
+fn set_request_cookies(
     url: &ServoUrl,
     headers: &mut HeaderMap,
     cookie_jar: &RwLock<CookieStorage>,
@@ -476,7 +476,7 @@ enum BodySink {
 }
 
 impl BodySink {
-    pub fn transmit_bytes(&self, bytes: Vec<u8>) {
+    fn transmit_bytes(&self, bytes: Vec<u8>) {
         match self {
             BodySink::Chunked(ref sender) => {
                 let sender = sender.clone();
@@ -490,7 +490,7 @@ impl BodySink {
         }
     }
 
-    pub fn close(&self) {
+    fn close(&self) {
         match self {
             BodySink::Chunked(_) => { /* no need to close sender */ },
             BodySink::Buffered(ref sender) => {
@@ -2264,7 +2264,7 @@ fn is_no_store_cache(headers: &HeaderMap) -> bool {
 }
 
 /// <https://fetch.spec.whatwg.org/#redirect-status>
-pub fn is_redirect_status(status: StatusCode) -> bool {
+fn is_redirect_status(status: StatusCode) -> bool {
     matches!(
         status,
         StatusCode::MOVED_PERMANENTLY |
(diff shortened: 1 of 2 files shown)


mrobinson:
cc @simonwuelker

mrobinson:
This doesn't seem to fix any WPT tests, but it does fix loading of some real world sites.

components/net/tests/http_loader.rs
```diff
@@ -1457,3 +1458,19 @@ fn test_fetch_compressed_response_update_count() {
...
+        let url = Url::parse(url_string).unwrap();
+        let origin = ImmutableOrigin::new(url.origin());
+        let serialized = format!("{}", serialize_origin(&origin));
+        assert_eq!(serialized, origin.ascii_serialization());
+    };
+
+    ensure_serialiations_match("https://example.com");
+    ensure_serialiations_match("http://example.com");
+    ensure_serialiations_match("https://example.com:1234");
+    ensure_serialiations_match("http://example.com:1234");
```
mukilan, line 1474: Should there be a case for the default port? i.e `ensure_serialiations_match("http://example.com:80");`

components/layout_2020/flexbox/layout.rs
```diff
@@ -1309,7 +1309,7 @@ fn do_initial_flex_line_layout<'items>(
 
 impl InitialFlexLineLayout<'_> {
     fn new<'items>(
-        flex_context: &mut FlexContext,
+        flex_context: &FlexContext,
```
mukilan, line 1312: Should this be part of this PR?

components/net/http_loader.rs
```diff
@@ -2320,18 +2320,30 @@ fn serialize_request_origin(request: &Request) -> headers::Origin {
...
     // Step 3. Return request’s origin, serialized.
+    serialize_origin(origin)
+}
+
+/// Step 3 of <https://fetch.spec.whatwg.org/#serializing-a-request-origin>.
+pub fn serialize_origin(origin: &ImmutableOrigin) -> headers::Origin {
     match origin {
         ImmutableOrigin::Opaque(_) => headers::Origin::NULL,
         ImmutableOrigin::Tuple(scheme, host, port) => {
+            // Note: This must be kept in sync with `Origin::ascii_serialization()``, which does not
```
mukilan, line 2331: Is this change still in conformance with the spec, or is it expected to be removed once we update the code that uses `ascii_serialization()` (or fixed in rust-url)?

components/net/tests/http_loader.rs
mrobinson: Good point. I've added this case.

components/layout_2020/flexbox/layout.rs
mrobinson: Nope. This wasn't meant to be part of this PR. I've removed it.

components/net/http_loader.rs
mrobinson: The specifications seem to differ on this. For instance, the original RFC 6454 specification says to omit the port if it's the default port (https://www.rfc-editor.org/rfc/rfc6454#section-6.2) and this is what Chromium does (https://chromium.googlesource.com/chromium/src/+/lkgr/url/scheme_host_port.cc#296). Older versions of the HTML specification had language like this as well (https://www.w3.org/TR/2011/WD-html5-20110525/origin-0.html), but the current one does not (https://html.spec.whatwg.org/#ascii-serialisation-of-an-origin).

It looks like what happens is that the data model of the URL changed such that internally the origin's port can be null when it's the default one (https://github.com/whatwg/html/pull/870). The origin in `rust-url` doesn't have support for this yet (https://github.com/servo/rust-url/issues/821). Once that is fixed this should work properly without an extra check.

For us, the most important thing is that the serialization is consistent, I think.

mukilan: That makes sense! Thanks for the additional context!

mukilan, line 2331:
```suggestion
            // Note: This must be kept in sync with `Origin::ascii_serialization()`, which does not
```

mukilan (approved):

mrobinson:
> Interesting, it's unfortunate that the specs are a bit unclear about this. Thanks for the fix ^^ Though perhaps it would be better to just fix [servo/rust-url#821](https://github.com/servo/rust-url/issues/821), so this change isn't needed anymore?

In parallel, I've also opened servo/rust-url#989 which should allow us to remove this bit of code in the future.

PR #34087, opened 2024-10-31
Use out parameter for generated methods returning JSVal

### LINKED ISSUE #6307: Use a MutableHandleValue outparam for calls returning GCThings
This is a followup to https://github.com/servo/servo/pull/6150 .

DOM methods that return JSVal or JSObjects are better served by taking a MutableHandleValue to dump the return value into.

jdm:
Returning raw JSVal values makes it easier to create GC hazards in code that calls these methods. Accepting a MutableHandle argument instead ensures that the values are rooted by the caller. As a side benefit, this also allows us to remove a bunch of awkward macro use in the WebGL code. Depends on https://github.com/servo/mozjs/pull/519.

---
- [x] These changes fix #6307

### DIFF
--- a/components/script/dom/dissimilaroriginwindow.rs
+++ b/components/script/dom/dissimilaroriginwindow.rs
@@ -5,8 +5,8 @@
 use base::id::PipelineId;
 use dom_struct::dom_struct;
 use js::jsapi::{Heap, JSObject};
-use js::jsval::{JSVal, UndefinedValue};
-use js::rust::{CustomAutoRooter, CustomAutoRooterGuard, HandleValue};
+use js::jsval::UndefinedValue;
+use js::rust::{CustomAutoRooter, CustomAutoRooterGuard, HandleValue, MutableHandleValue};
 use script_traits::{ScriptMsg, StructuredSerializedData};
 use servo_url::ServoUrl;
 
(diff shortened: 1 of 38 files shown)


components/script/dom/dissimilaroriginwindow.rs
```diff
@@ -167,9 +167,9 @@ impl DissimilarOriginWindowMethods for DissimilarOriginWindow {
     }
 
     // https://html.spec.whatwg.org/multipage/#dom-opener
-    fn Opener(&self, _: JSContext) -> JSVal {
+    fn Opener(&self, _: JSContext, mut retval: MutableHandleValue) {
         // TODO: Implement x-origin opener
-        UndefinedValue()
+        retval.set(UndefinedValue());
```
gterzian, line 172: is `retval` not undefined by default?

gterzian (approved):
LGTM

components/script/dom/history.rs
```diff
@@ -289,11 +289,12 @@ impl History {
 
 impl HistoryMethods for History {
     /// <https://html.spec.whatwg.org/multipage/#dom-history-state>
-    fn GetState(&self, _cx: JSContext) -> Fallible<JSVal> {
+    fn GetState(&self, _cx: JSContext, mut retval: MutableHandleValue) -> Fallible<()> {
```
sagudev, line 292: I understand why this needs to be mut for https://doc.servo.org/mozjs/gc/root/struct.MutableHandle.html#tymethod.set, but this is still weird as MutableHandle is clonable so `&mut` does not help a lot with safety.

components/script/dom/dissimilaroriginwindow.rs
sagudev: Not necessarily, that depends on the caller (although in this case the codegen). I think this is more correct and being more explicit is always better.

components/script/dom/history.rs
sagudev: I opened https://github.com/servo/mozjs/issues/520

sagudev:
BTW to caught clippy stuff locally you can use updated configuration for IDE: https://book.servo.org/hacking/editor-support.html#visual-studio-code

PR #34088, opened 2024-10-31
ReadableStream: Correct ReadableStream::error_native implementation and fix clippy warnings

Taym95:

I tried to finish remaining TODOs e.g `error_native` and fix clippy warnings

---

### DIFF
--- a/components/script/dom/readablestreamdefaultcontroller.rs
+++ b/components/script/dom/readablestreamdefaultcontroller.rs
@@ -135,6 +135,7 @@ pub struct ValueWithSize {
 #[allow(crown::unrooted_must_root)]
 pub enum EnqueuedValue {
     /// A value enqueued from Rust.
+    #[allow(dead_code)]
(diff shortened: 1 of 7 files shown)


components/script/dom/readablestreamdefaultcontroller.rs
```diff
@@ -135,6 +135,7 @@ pub struct ValueWithSize {
 #[allow(crown::unrooted_must_root)]
 pub enum EnqueuedValue {
     /// A value enqueued from Rust.
+    #[allow(dead_code)]
```
gterzian, line 138: Yes this is something interesting because of https://github.com/servo/servo/blob/36f6d08e45a54367f2297c740741332705a39778/components/script/dom/readablestreamdefaultcontroller.rs#L560

so we later can either not use jsvals for native values(as a kind of optimization), or remove this.

gterzian (approved):
thank you.

PR #34089, opened 2024-10-31
Fix logical window-size calculation

jschwe:
We need to divide the physical size by the hidpi factor to get the device-independent logical window size. Fixes the calculation introduced in 850e59f9.

---
- [x] These changes fix erroneous redirection to the desktop website on mobile devices with high dpi and websites that check the screen width.


mukilan:
Good catch! The issue is present in headed_window as well. I have a fix in 708b98970b52e4411b0a8a7fc67e1870c2ee7b73 which also changes egl port's ServoWindowCallbacks struct to use the Scale type instead of a raw f32 to prevent  such bugs in the future. I've triggered a [try run]( https://github.com/mukilan/servo/actions/runs/11625007713) in my fork. I'll create a new PR once that passes.

PR #34090, opened 2024-10-31
Replace ComputedValues with WritingMode on IndefiniteContainingBlock

Loirooriol:
We only need the writing mode, not the entire computed style.

---

- [X] These changes do not require tests because there should be no behavior change

### DIFF
--- a/components/layout_2020/lib.rs
+++ b/components/layout_2020/lib.rs
@@ -29,56 +29,60 @@ use app_units::Au;
 pub use flow::BoxTree;
 pub use fragment_tree::FragmentTree;
 use geom::AuOrAuto;
+use style::logical_geometry::WritingMode;
 use style::properties::ComputedValues;
 
 use crate::geom::LogicalVec2;
 
 /// A containing block useful for calculating inline content sizes, which may
 /// have inline sizes that depend on block sizes due to aspect ratio.
-pub(crate) struct IndefiniteContainingBlock<'a> {
+pub(crate) struct IndefiniteContainingBlock {
     pub size: LogicalVec2<AuOrAuto>,
-    pub style: &'a ComputedValues,
+    pub writing_mode: WritingMode,
 }
 
-impl<'a> IndefiniteContainingBlock<'a> {
-    fn new_for_style(style: &'a ComputedValues) -> Self {
-        Self::new_for_style_and_block_size(style, AuOrAuto::Auto)
+impl IndefiniteContainingBlock {
+    fn new_for_writing_mode(writing_mode: WritingMode) -> Self {
+        Self::new_for_writing_mode_and_block_size(writing_mode, AuOrAuto::Auto)
     }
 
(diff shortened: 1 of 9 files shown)


components/layout_2020/lib.rs
```diff
@@ -29,56 +29,57 @@ use app_units::Au;
...
     pub size: LogicalVec2<AuOrAuto>,
-    pub style: &'a ComputedValues,
+    pub writing_mode: WritingMode,
 }
 
-impl<'a> IndefiniteContainingBlock<'a> {
-    fn new_for_style(style: &'a ComputedValues) -> Self {
-        Self::new_for_style_and_block_size(style, AuOrAuto::Auto)
+impl IndefiniteContainingBlock {
+    fn new_for_wm(writing_mode: WritingMode) -> Self {
```
mrobinson, line 45: I think we should avoid the 'wm' abbreviation because that usually means "window manager."

```suggestion
    fn new_for_writing_mode(writing_mode: WritingMode) -> Self {
```

components/layout_2020/lib.rs
```diff
@@ -29,56 +29,57 @@ use app_units::Au;
...
+        Self::new_for_wm_and_block_size(writing_mode, AuOrAuto::Auto)
     }
 
     /// Creates an [`IndefiniteContainingBlock`] with the provided style and block size,
     /// and the inline size is set to auto.
     /// This is useful when finding the min-content or max-content size of an element,
     /// since then we ignore its 'inline-size', 'min-inline-size' and 'max-inline-size'.
-    fn new_for_style_and_block_size(style: &'a ComputedValues, block_size: AuOrAuto) -> Self {
+    fn new_for_wm_and_block_size(writing_mode: WritingMode, block_size: AuOrAuto) -> Self {
```
mrobinson, line 53:
```suggestion
    fn new_for_writing_mode_and_block_size(writing_mode: WritingMode, block_size: AuOrAuto) -> Self {
```

mrobinson (approved):

nicoburns (commented):
It seems that the reason that the `ComputedValues` (aka `style`) isn't required here is because the `FlexContainer` already has a reference to it. As that stored `ComputedValues` will contain the `WritingMode` as a field, doesn't that mean that the writing mode isn't required here either?

Loirooriol:
@nicoburns The `IndefiniteContainingBlock`s are created in places where we have the full style, including the writing mode. But then this struct is passed to code that otherwise wouldn't have access to the style nor writing mode. This other code only needs to know the writing mode, so by not passing the entire style we can avoid some style clones and having to deal with lifetimes.

mrobinson:
> For some reason CI is failing due to intermittent tests not being detected as intermittent, I will try rebasing

I think what might be going on is that we switched to a new deployment of the intermittent tracker, but the webhook that tracks newly opened issues wasn't set up. I've set that up now, so hopefully it will start tracking those issues.

PR #34068, opened 2024-10-30
CI: use new intermittent-tracker deployment

delan:
This patch switches to our new [intermittent-tracker](https://github.com/servo/intermittent-tracker) instance at [intermittent-tracker.servo.org](https://intermittent-tracker.servo.org), as deployed in servo/ci-runners [43d56bf194f81...304212f15d9b2](https://github.com/servo/ci-runners/compare/43d56bf194f81b3f88ab5d22b6339968411465a8...304212f15d9b2a46a5209ab1278e94388b3579ab).

As a result, we will also be able to test future intermittent-tracker changes by deploying to [staging.intermittent-tracker.servo.org](https://staging.intermittent-tracker.servo.org), and opening a pull request in this repo with the following changes:

- `TRACKER_API = "https://staging.intermittent-tracker.servo.org"`
- `TRACKER_DASHBOARD_SECRET_ENV_VAR = "INTERMITTENT_TRACKER_DASHBOARD_SECRET_STAGING"`

---

- [x] `./mach build -d` does not report any errors
- [x] `./mach test-tidy` does not report any errors
- [ ] ~~These changes fix #___ (GitHub issue number if applicable)~~

- [x] CI test plan
  - [x] Build with unauthenticated POST /dashboard/query (<https://github.com/servo/servo/actions/runs/11659374763/job/32459956121#step:8:3022>)
  - [x] Build with authenticated POST /dashboard/attempts (<https://github.com/delan/servo/actions/runs/11659487000/job/32461315667#step:8:2649>)
  - [x] Data stored in tracker correctly: <https://intermittent-tracker.servo.org/?build_url=https%3A%2F%2Fgithub.com%2Fdelan%2Fservo%2Factions%2Fruns%2F11659487000>

PR #34092, opened 2024-11-01
Return cached object from `CryptoKey.algorithm` getter

### LINKED ISSUE #33908: CryptoKey.algorithm should return a cached object
From https://w3c.github.io/webcrypto/#cryptokey-interface-members:
>Returns the [cached ECMAScript object](https://w3c.github.io/webcrypto/#concept-cached-object) associated with the [[[algorithm]]](https://w3c.github.io/webcrypto/#dfn-CryptoKey-slot-algorithm) internal slot.

Meanwhile, the implementation creates a new instance each time it's called: https://github.com/servo/servo/blob/main/components/script/dom/cryptokey.rs#L106-L127

simonwuelker:
Fixes the [`CryptoKey.algorithm`](https://w3c.github.io/webcrypto/#cryptokey-interface-members) getter which previously did not cache the returned object, but rather created a new one each time.

---

- [X] These changes fix #33908 

### DIFF
--- a/components/script/dom/subtlecrypto.rs
+++ b/components/script/dom/subtlecrypto.rs
@@ -11,8 +11,8 @@ use aes::cipher::{BlockDecryptMut, BlockEncryptMut, KeyIvInit, StreamCipher};
 use aes::{Aes128, Aes192, Aes256};
 use base64::prelude::*;
 use dom_struct::dom_struct;
-use js::conversions::ConversionResult;
-use js::jsapi::JSObject;
+use js::conversions::{ConversionResult, ToJSValConvertible};
+use js::jsapi::{JSObject, Value};
 use js::jsval::ObjectValue;
 use js::rust::MutableHandleObject;
 use js::typedarray::ArrayBufferU8;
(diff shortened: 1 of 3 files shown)


components/script/dom/subtlecrypto.rs
```diff
@@ -827,25 +828,41 @@ impl SubtleCrypto {
...
+
+            CryptoKey::new(
+                &self.global(),
+                KeyType::Secret,
+                extractable,
+                name,
+                algorithm_object.handle(),
+                usages,
+                handle,
+            )
```
simonwuelker, line 850: Perhaps there is an easier way to do this? I have a `AesKeyAlgorithm` and want to turn it into a `HandleObject`, but end up creating an intermediary js value that i call `to_object` on. Not very nice.

jdm: There's https://doc.servo.org/script/dom/bindings/codegen/Bindings/SubtleCryptoBinding/struct.AesKeyAlgorithm.html#method.to_jsobject, but you need to start by creating a new JSObject (JS_NewObject) that you can pass as an argument. We should probably make that less of a foot gun.

components/script/dom/subtlecrypto.rs
```diff
@@ -865,15 +882,31 @@ impl SubtleCrypto {
...
-        Ok(CryptoKey::new(
-            &self.global(),
-            KeyType::Secret,
-            extractable,
-            KeyAlgorithm { name },
-            usages,
-            handle,
-        ))
+
+        let name = DOMString::from(alg_name);
```
jdm, line 886: Can we move this into a helper function to reduce duplication?

jdm (approved):

components/script/dom/subtlecrypto.rs
simonwuelker: I see, thank you!

components/script/dom/subtlecrypto.rs
simonwuelker: I tried doing that in https://github.com/servo/servo/pull/34092/commits/42bba2d4a1fa86c1adb53b10b4f494625bb7ecb8 - there's still some duplication because I'm not sure how to return a rooted object from a function safely, so we end up with an out parameter to `AesKeyAlgorithm::from_name_and_size` and callers still need to construct the algorithm object themselves.

sagudev:
ERROR [expected TIMEOUT] /html/canvas/element/manual/imagebitmap/createImageBitmap-serializable.html
    OK [expected CRASH] /html/canvas/offscreen/canvas-host/2d.canvas.host.size.attributes.parse.minus.html

but https://github.com/servo/servo/issues/34120 and https://github.com/servo/servo/issues/34121 exists.

Might be due to https://github.com/servo/servo/pull/34068. cc @delan

PR #34096, opened 2024-11-01
servoshell: fix logical screen size calculation

mukilan:
Similar to #34089, but also modifieds the egl port's ServoWindowCallbacks
struct to use the proper Scale type instead of a raw f32 to prevent similar
bugs in new code.

---

- [x] These changes fix the issue identified by #34089


jschwe:
Does the headless_window code not need to be changed?

mukilan:
> Does the headless_window code not need to be changed?

I don't think so, as we derive all the values from the `size`  parameter to `Window::new` which is already in logical units.

mrobinson (approved):

PR #34097, opened 2024-11-01
Remove assert on controller having a stream within should call pull

gterzian:
The spec doesn't assert, so I think we were too eager to do so in that spot. 

Fixes a "Controller must have a stream when the should pull algo is called into." CRASH of tee.any.js.

The test still crashes elsewhere though with:
```called `Result::unwrap()` on an `Err` value: () (thread Script(1,1), at /Users/Gregory/Projects/servo/target/debug/build/script-5f068b888146758c/out/Bindings/UnderlyingSourceBinding.rs:168)```

To be investigated separately.

---


Taym95 (approved):

gterzian:

PR #34098, opened 2024-11-01
Fix call to to_jsobject in underlying source algos

gterzian:
This fixes quite a lot of different crashes like ```called `Result::unwrap()` on an `Err` value: () (thread Script(1,1), at /Users/Gregory/Projects/servo/target/debug/build/script-5f068b888146758c/out/Bindings/UnderlyingSourceBinding.rs:168)```

which relate to the `set_dictionary_property` call within `UnderlyingSource::to_jsobject`, which we call each time we want to call into an algo, for example [in the start algo](https://github.com/servo/servo/blob/b460d74a42333eb5b552e33a6958c3b0df1bb50d/components/script/dom/underlyingsourcecontainer.rs#L129)

The problem was a null object was passed to the call to `to_jsobject`, which is now replaced to one created using `JS_NewObject`.

Various test still fails, but later at [the actual JS call](https://github.com/servo/servo/blob/be4f79de49e3f7276137d958daf0bae474bc247e/components/script/dom/underlyingsourcecontainer.rs#L134) which returns `JS_Failed`. To be investigated separately. 

---

### DIFF
--- a/components/script/dom/underlyingsourcecontainer.rs
+++ b/components/script/dom/underlyingsourcecontainer.rs
@@ -6,7 +6,7 @@ use std::ptr;
 use std::rc::Rc;
 
 use dom_struct::dom_struct;
-use js::jsapi::{IsPromiseObject, JSObject};
+use js::jsapi::{IsPromiseObject, JSObject, JS_NewObject};
 use js::rust::IntoHandle;
 
 use crate::dom::bindings::callback::ExceptionHandling;
@@ -94,7 +94,11 @@ impl UnderlyingSourceContainer {
             if let Some(pull) = &source.pull {
                 let cx = GlobalScope::get_cx();
                 rooted!(in(*cx) let mut this_object = ptr::null_mut::<JSObject>());
+                // TODO: move this into `bindings`.
                 unsafe {
+                    let obj = JS_NewObject(*cx, ptr::null_mut());
+                    assert!(!obj.is_null());
+                    this_object.set(obj);
                     source.to_jsobject(*cx, this_object.handle_mut());
                 }
                 let this_handle = this_object.handle();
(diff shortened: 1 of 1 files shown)


jdm (approved):

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -126,6 +127,7 @@ impl UnderlyingSourceContainer {
                 let cx = GlobalScope::get_cx();
                 rooted!(in(*cx) let mut this_object = ptr::null_mut::<JSObject>());
                 unsafe {
+                    this_object.set(JS_NewObject(*cx, ptr::null_mut()));
```
simonwuelker, line 130: [`JS_NewObject`](http://devdoc.net/web/developer.mozilla.org/en-US/docs/SpiderMonkey/JSAPI_Reference/JS_NewObject.html) can fail, so we should probably add an assertion (or even fail gracefully).
```rust
let obj = JS_NewObject(*cx, ptr::null_mut());
assert!(!obj.is_null());
this_object.set(obj);
```

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -95,6 +95,7 @@ impl UnderlyingSourceContainer {
...
```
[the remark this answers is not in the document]
gterzian: yes this is somewhat annoying, and will make it harder to eventually move the unsafe part to `bindings`. Is it ok to leave it as is for now? @jdm

components/script/dom/underlyingsourcecontainer.rs
gterzian: Thanks for pointing it out.

components/script/dom/underlyingsourcecontainer.rs
```diff
@@ -125,7 +129,11 @@ impl UnderlyingSourceContainer {
             if let Some(start) = &source.start {
                 let cx = GlobalScope::get_cx();
                 rooted!(in(*cx) let mut this_object = ptr::null_mut::<JSObject>());
+                // TODO: move this into `bindings`.
                 unsafe {
+                    let obj = JS_NewObject(*cx, ptr::null_mut());
+                    assert!(!obj.is_null());
+                    this_object.set(obj);
```
gterzian, line 136: btw could we not put this a safe generated method on all `DomObject`?

components/script/dom/underlyingsourcecontainer.rs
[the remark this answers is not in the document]
jdm: Sure.

PR #34100, opened 2024-11-01
frozen array in XRInputSourcesChangeEvent, update `to_frozen_array` doc

### LINKED ISSUE #34050: XRInputSourcesChangeEvent.added/.removed should store frozen array values
Spec: https://immersive-web.github.io/webxr/#xrinputsourceschangeevent-interface
Implementation: https://github.com/servo/servo/blob/main/components/script/dom/xrinputsourceschangeevent.rs#L90-L97

Instead of converting a Rust array to an equivalent JS value, we should use the [to_frozen_array](https://doc.servo.org/script/dom/bindings/utils/fn.to_frozen_array.html) helper.

I hope we will see some new tests passing in `./mach test-wpt tests/wpt/tests/webxr/events_input_sources_change.https.html`.

chickenleaf:
PR Description
Issue:

1. `XRInputSourcesChangeEvent` was not using `to_frozen_array` for array conversion, leading to manual conversions inconsistent with WebXR spec requirements. 
2. `to_frozen_array` rustdoc comment needs to be updated.

Changes:
1. Replaced manual conversions with `to_frozen_array`
2. Used `JSContext::from_ptr(*cx)` for context creation
3. Changed `to_frozen_array` rustdoc comment to reflect broader usage
4. Add tests to ensure `XRInputSourcesChangeEvent` `.added` and `.removed` are frozen arrays, with assertions that modifications throw `TypeError`.

Test:
`tests/wpt/tests/webxr/events_input_sources_change.https.html`
---

- [X] These changes fix #34050 

- [ ] These changes do not require tests because _____

### DIFF
--- a/components/script/dom/xrinputsourceschangeevent.rs
+++ b/components/script/dom/xrinputsourceschangeevent.rs
@@ -3,9 +3,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 use dom_struct::dom_struct;
-use js::conversions::ToJSValConvertible;
 use js::jsapi::Heap;
-use js::jsval::{JSVal, UndefinedValue};
+use js::jsval::JSVal;
 use js::rust::HandleObject;
 use servo_atoms::Atom;
 
@@ -17,6 +16,7 @@ use crate::dom::bindings::inheritance::Castable;
 use crate::dom::bindings::reflector::{reflect_dom_object_with_proto, DomObject};
 use crate::dom::bindings::root::{Dom, DomRoot};
 use crate::dom::bindings::str::DOMString;
+use crate::dom::bindings::utils::to_frozen_array;
 use crate::dom::event::Event;
 use crate::dom::globalscope::GlobalScope;
 use crate::dom::window::Window;
@@ -88,12 +88,12 @@ impl XRInputSourcesChangeEvent {
         let _ac = enter_realm(global);
         let cx = GlobalScope::get_cx();
         unsafe {
-            rooted!(in(*cx) let mut added_val = UndefinedValue());
-            added.to_jsval(*cx, added_val.handle_mut());
-            changeevent.added.set(added_val.get());
-            rooted!(in(*cx) let mut removed_val = UndefinedValue());
-            removed.to_jsval(*cx, removed_val.handle_mut());
-            changeevent.removed.set(removed_val.get());
+            changeevent
+                .added
+                .set(to_frozen_array(added, JSContext::from_ptr(*cx)));
+            changeevent
+                .removed
+                .set(to_frozen_array(removed, JSContext::from_ptr(*cx)));
         }
 
         changeevent
--- a/components/script/dom/xrinputsourceschangeevent.rs
+++ b/components/script/dom/xrinputsourceschangeevent.rs
@@ -87,15 +87,8 @@ impl XRInputSourcesChangeEvent {
         }
         let _ac = enter_realm(global);
         let cx = GlobalScope::get_cx();
-        unsafe {
-            changeevent
-                .added
-                .set(to_frozen_array(added, JSContext::from_ptr(*cx)));
-            changeevent
-                .removed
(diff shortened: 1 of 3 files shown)


components/script/dom/xrinputsourceschangeevent.rs
```diff
@@ -88,12 +88,12 @@ impl XRInputSourcesChangeEvent {
...
         unsafe {
-            rooted!(in(*cx) let mut added_val = UndefinedValue());
-            added.to_jsval(*cx, added_val.handle_mut());
-            changeevent.added.set(added_val.get());
-            rooted!(in(*cx) let mut removed_val = UndefinedValue());
-            removed.to_jsval(*cx, removed_val.handle_mut());
-            changeevent.removed.set(removed_val.get());
+            changeevent
+                .added
+                .set(to_frozen_array(added, JSContext::from_ptr(*cx)));
```
jdm, line 93: Oh, we can just pass `cx` here instead of dereferencing it and recreating the JSContext type from the pointer.

jdm (approved):
This looks fine, but we can simplify it a little bit!

jdm:
>These changes do not require tests because they do not affect the functionality.

This isn't true; we're definitely changing functionality here. We could update the webxr tests to verify that the added and removed properties are frozen arrays. There's similar test code that looks like this: https://github.com/web-platform-tests/wpt/blob/e7e2de2d3ad8e8160509838ddbf37666cd69a90e/push-api/supported-encodings.https.any.js#L16

components/script/dom/xrinputsourceschangeevent.rs
chickenleaf: done!
also shall i remove this?
```
warning: unnecessary `unsafe` block
  --> components\script\dom\xrinputsourceschangeevent.rs:90:9
   |
90 |         unsafe {
   |         ^^^^^^ unnecessary `unsafe` block
   |
   = note: `#[warn(unused_unsafe)]` on by default
```

jdm: Yep!

chickenleaf:
> >These changes do not require tests because they do not affect the functionality.
> 
> This isn't true; we're definitely changing functionality here. We could update the webxr tests to verify that the added and removed properties are frozen arrays. There's similar test code that looks like this: https://github.com/web-platform-tests/wpt/blob/e7e2de2d3ad8e8160509838ddbf37666cd69a90e/push-api/supported-encodings.https.any.js#L16

Which tests need the updation?

jdm:
https://github.com/web-platform-tests/wpt/blob/e7e2de2d3ad8e8160509838ddbf37666cd69a90e/webxr/events_input_sources_change.https.html looks like a good candidate for adding the new test coverage.

chickenleaf:
> https://github.com/web-platform-tests/wpt/blob/e7e2de2d3ad8e8160509838ddbf37666cd69a90e/webxr/events_input_sources_change.https.html looks like a good candidate for adding the new test coverage.

shall i update it to do this as well? 

> tests to verify that the added and removed properties are frozen arrays

just adding two tests to see if they throw a typeerror when i try to modify the arrays.
similar to the `push-api/supported-encodings.https.any.js` L16 u sent

jdm:
Yep! That would be great!

chickenleaf:
Screenshot 2024-11-02 165259
this is good?
```
assert_throws_js(
    TypeError,
    () => { event.added[0] = "test"; },
    "added array must be frozen"
);
assert_throws_js(
    TypeError,
    () => { event.removed[0] = "test"; },
    "removed array must be frozen"
);
```
and added use strict on top. should i be adding any comments? 

also,,, is using `array.push()` better? compared to this way?

jdm:
That looks reasonable! Does the test pass?

chickenleaf:
> That looks reasonable! Does the test pass?

weirdly enough no? the whole test doesnt pass, but the subtests pass

jdm (approved):

PR #34106, opened 2024-11-02
Optimize color multiplication by replacing a * 255 with (a << 8) - a

olejaaaaaaaa:
It will work a little faster


jdm:
Have you looked at the generated code before and after? This is the kind of optimization that I would expect LLVM to do for us already.

olejaaaaaaaa:
Well, the compiler was very smart and did this optimization, I checked it and the only difference is in the Debug version

